PR #70, opened 2023-07-12
Sanitizer libraries static linking

MeatBoy106:
GCC uses `-static-libasan` and `-static-libubsan` to statically link with address and undefined behavior sanitizers while CLANG uses `-static-libsan`. This patch discriminates between the two cases and introduces static link with libubsan when using GCC as well.

Change-Id: I2441466c5c9343afd938185b8ec5047d4e95ac70


giactra (approved):
lgtm

PR #71, opened 2023-07-12
misc: Update README/README.md

BobbyRBruce:
Change-Id: Ic9174d6c3d905f85e2f7f159478d8c93ef7e4079


README.md
```diff
@@ -0,0 +1,88 @@
...
+
+To build gem5, you will need the following software: g++ or clang,
+Python (gem5 links in the Python interpreter), SCons, zlib, m4, and lastly
+protobuf if you want trace capture and playback support. Please see
+<http://www.gem5.org/documentation/general_docs/building> for more details
+concerning the minimum versions of these tools.
+
+Once you have all dependencies resolved, type
+`scons build/<CONFIG>/gem5.opt` where `<CONFIG>` is one of the options in
+build_opts like `ARM`, `NULL`, `MIPS`, `POWER`, `SPARC`, `X86`,
```
powerjg, line 30: ALL should be the default here now.

README.md
```diff
@@ -0,0 +1,88 @@
...
+
+The main source tree includes these subdirectories:
+
+* build_opts: pre-made default configurations for gem5
+* build_tools: tools used internally by gem5's build process.
+* configs: example simulation configuration scripts
+* ext: less-common external packages needed to build gem5
+* include: include files for use in other programs
+* site_scons: modular components of the build system
+* src: source code of the gem5 simulator
```
powerjg:
```suggestion
* src: source code of the gem5 simulator. The C++ source, Python wrappers, and Python standard library are found in this directory.
```

README.md
```diff
@@ -0,0 +1,88 @@
...
+bugs, requests features, or engage in community discussions. Below
+are a few of the most common we recommend using.
+
+* **gem5-users@gem5.org**: A mailing list for users of gem5 to ask questions
+or start discussions. To join the mailing list please visit
+<https://www.gem5.org/mailing_lists>.
+* **gem5-dev@gem5.org**: A mailing list for developers of gem5 to ask questions
+or start discussions. To join the mailing list please visit
+<https://www.gem5.org/mailing_lists>.
+* **GitHub Discussions**: A GitHub Discussions page. This can be used to start
```
powerjg, line 69: Reorder with this at the top? Do we want to encourage github first and mailing list last?

Similarly, maybe we should move slack up?

build_tools/infopy.py
```diff
@@ -51,7 +51,7 @@
 code = code_formatter()
 
 for source in args.files:
-    src = os.path.basename(source)
+    src = os.path.basename(source).replace(".", "_")
```
powerjg, line 58: A comment for why this is needed would be good.

src/SConscript
```diff
@@ -615,7 +615,7 @@ PySource('m5', 'python/m5/defines.py')
 
 # Generate a file that wraps the basic top level files
 gem5py_env.Command('python/m5/info.py',
-            [ File('#/COPYING'), File('#/LICENSE'), File('#/README'),
+            [ File('#/COPYING'), File('#/LICENSE'), File('#/README.md'),
```
powerjg, line 618: This is OK for now. However, I think we should just remove `info.py` completely. I can't think of any other tool that has the readme/license embedded this way.

powerjg (changes requested):

src/SConscript
BobbyRBruce: > This is OK for now. However, I think we should just remove `info.py` completely. I can't think of any other tool that has the readme/license embedded this way.

I'm not sure I know exactly why we do this at all. I assume we print this to a terminal somewhere? Any way, seems weird. I'll put it in an issue.

BobbyRBruce: Issue #89 is to remove this at some point.

README.md
BobbyRBruce: I've reordered this.

BobbyRBruce:
BTW: I didn't mean to delete that comment. It was you asking to make `ALL/gem5.opt` the default. I did so.

powerjg (approved):

PR #73, opened 2023-07-13
Unit tests miscellaneous patches

MeatBoy106:
Compiling and running unit tests with clang-16 and address sanitizer exhibits errors related to old-style memory management or permissive syntax. This pull request addresses these issues and provides opportunistic improvements along the way.

### DIFF
--- a/src/base/inifile.cc
+++ b/src/base/inifile.cc
@@ -42,17 +42,6 @@ namespace gem5
 IniFile::IniFile()
 {}
 
-IniFile::~IniFile()
-{
-    SectionTable::iterator i = table.begin();
-    SectionTable::iterator end = table.end();
-
-    while (i != end) {
-        delete (*i).second;
-        ++i;
-    }
-}
-
 bool
 IniFile::load(const std::string &file)
 {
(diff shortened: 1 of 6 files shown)


MeatBoy106:
I suggest enabling address sanitizer for unit tests as it helps catching subtle false negative, memory leaks and so. I'm not sure how to provide such patch myself though.

src/base/inifile.cc
```diff
@@ -294,32 +285,29 @@ IniFile::Section::printUnreferenced(const std::string &sectionName)
...
+IniFile::printUnreferenced() const
 {
     bool unref = false;
 
-    for (SectionTable::iterator i = table.begin();
-         i != table.end(); ++i) {
-        const std::string &sectionName = i->first;
-        Section *section = i->second;
+    for (auto& entry: table) {
+        auto& [sectionName, section] = entry;
```
powerjg, line 300: Do we know when GCC/clang started supporting this syntax? I want to make sure that all of our supported compilers support this.

powerjg (commented):
Generally, this looks good to me. There are two things that we might want to do before merging.
1. It would be great to run the compiler tests on this before merging... I think this might be possible in github now.
2. For the trie... this was implemented to have *very* high performance lookups. We should make sure that this change doesn't have an impact on gem5's performance. (On a side note, I personally and not convinced that the trie is really needed for performance, so this may be a moot issue.)

BobbyRBruce (commented):
Please address Jason's question

(Starting a review here so this comes up as being "under review" rather than "pending review")

src/base/inifile.cc
MeatBoy106: image

MeatBoy106:
Regarding performance, this new implementation guarantees the same number of memory allocation and pointer indirections than the previous implementation. The only theoretical difference I can think of should come from the unique_ptr move that implies reseting the moved pointer after it has been moved from. I would not be surprised if optimization removed both the reset and the destructor call for move unique_ptr though. My guess is: no measurable difference.

powerjg (approved):
Thanks for looking into the compilers and the other question :)

PR #77, opened 2023-07-13
base: Ostream helpers (iterable, tuple, pair, enum, pointers, optional)

MeatBoy106:
Refinements, tests and extra features for ostream helpers. Most of these imported and updated from Gerrit submission.

### DIFF
--- a/src/base/stl_helpers/ostream_helpers.test.cc
+++ b/src/base/stl_helpers/ostream_helpers.test.cc
@@ -79,3 +79,14 @@ TEST(OstreamHelpers, optional) {
     os << m;
     EXPECT_EQ(os.str(), "(-)");
 }
+
+TEST(OstreamHelpers, printer) {
+    std::string hello = "Hello";
+    std::ostringstream os;
+    os << hello;
+    EXPECT_EQ(os.str(), hello);
+
+    std::ostringstream os2;
+    os2 << gem5::stl_helpers::Printer(hello);
+    EXPECT_EQ(os2.str(), "[ H, e, l, l, o, ]");
+}
--- a/src/base/stl_helpers/ostream_helpers.test.cc
+++ b/src/base/stl_helpers/ostream_helpers.test.cc
@@ -113,6 +113,6 @@ TEST(OstreamHelpers, pointers) {
     auto uptr = std::make_unique<std::string>("Hello, World!");
     EXPECT_EQ(helped_representation(uptr), expected_representation(uptr));
 
-    auto sptr = std::make_unique<std::optional<bool>>();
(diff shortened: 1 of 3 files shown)


powerjg (dismissed):

src/base/stl_helpers/ostream_helpers.test.cc
```diff
@@ -90,3 +90,29 @@ TEST(OstreamHelpers, printer) {
     os2 << gem5::stl_helpers::Printer(hello);
     EXPECT_EQ(os2.str(), "[ H, e, l, l, o, ]");
 }
+
+
+TEST(OstreamHelpers, pointers) {
+    auto helpedRepresentation = [](const auto& val) {
```
giactra: Nit: this should be lower case instead of lower camel case being a local variable

src/base/stl_helpers/ostream_helpers.test.cc
```diff
@@ -90,3 +90,29 @@ TEST(OstreamHelpers, printer) {
...
+        std::ostringstream os;
+        os << gem5::stl_helpers::Printer(val);
+        return os.str();
+    };
+    auto expectedRepresentation = [&](const auto& ptr) {
```
giactra: Ditto

src/base/stl_helpers/ostream_helpers.test.cc
```diff
@@ -90,3 +90,29 @@ TEST(OstreamHelpers, printer) {
...
+        using gem5::stl_helpers::operator<<;
+        std::ostringstream os;
+        auto* rawPtr = &*ptr;
```
giactra: Ditto

giactra (changes requested):
Sorry to stop the merging process. Overall LGTM but I have a minor request

BobbyRBruce:
Hey @MeatBoy106 : This isn't a harder requirement but it would help us a little bit if you could add tags to the PR title as well. We're still, for now, enforcing the use of tags and github, by default, uses the PR title as the merge commit header, so it's nice to have that setup for us (when a Maintainer squashes and merges the commit they can choose to edit the header, but that means they choose what the tags are. It's nice if the contributor can do that).

MeatBoy106:
@BobbyRBruce Sure! Is it better that way?

BobbyRBruce:
> @BobbyRBruce Sure! Is it better that way?

Yes. Thank you!

BobbyRBruce (approved):

BobbyRBruce:
I'll merge this when @giactra has approved.

giactra (approved):
Thanks for the change

PR #80, opened 2023-07-14
misc: Drop older compilers and Ubuntu 18.04

### LINKED ISSUE #74: New HashHelper tests not compiling with gcc-7.5.0
I am encountering a compilation problem with the new HashHelper unit-tests [1]:

`unordered_map.h:133:55: error: 'value' is not a member of 'std::__not_<std::__and_<std::__is_fast_hash<std::hash<std::tuple<int, bool, int**, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > (*)(float)> > >, std::__detail::__is_noexcept_hash<std::tuple<int, bool, int**, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > (*)(float)>, std::hash<std::tuple<int, bool, int**, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > (*)(float)> > > > >'
       using node_type = typename _Hashtable::node_type;`

The problem doesn't seem to be present in newer versions of gcc so I am just changing the version. However we do support gcc-7 in gem5 so either we work around the problem or we deprecate gcc-7 support @BobbyRBruce @powerjg.
gcc-7 is default compiler for Ubuntu18.04 which is past its EOL anyway so I believe it is a reasonable solution

[1]: https://gem5-review.googlesource.com/c/public/gem5/+/67663

BobbyRBruce:


powerjg (approved):

giactra (approved):
LGTM. What will happen to docker images in the registry?

PR #81, opened 2023-07-14
arch-riscv: Fix clearLoadReservation merge

rogerchang23424:
The previous change
(https://gem5-review.googlesource.com/c/public/gem5/+/71818) makes the clearLoadReservation be RISC-V only.


BobbyRBruce (approved):
Thanks Roger.

PR #82, opened 2023-07-14
configs: fix GPU's default number of HW barrier/CU

KaiBatley:
AMD GCN3 and Vega GPUs assume a max of 16 WG/CU.  Any GPU WG with more than 1 WF requires a hardware barrier to allow WFs in the WG to synchronize locally.  However, currently the default gem5 GPU configuration assumes only 4 barriers per CU, which artificially prevents applications with > 4 WG/CU that could run simultaneously from running simultaneously.

This fix resolves this by updating the default number of hardware barriers per CU to 16, which mimics the support described in slide 39 here: https://www.olcf.ornl.gov/wp-content/uploads/2019/10/ORNL_Application_Readiness_Workshop-AMD_GPU_Basics.pdf

### DIFF
--- a/src/arch/generic/isa.hh
+++ b/src/arch/generic/isa.hh
@@ -70,7 +70,6 @@ class BaseISA : public SimObject
   public:
     virtual PCStateBase *newPCState(Addr new_inst_addr=0) const = 0;
     virtual void clear() {}
-    virtual void clearLoadReservation(ContextID cid) {}
 
     virtual RegVal readMiscRegNoEffect(RegIndex idx) const = 0;
     virtual RegVal readMiscReg(RegIndex idx) = 0;
--- a/src/arch/riscv/isa.hh
+++ b/src/arch/riscv/isa.hh
@@ -136,6 +136,12 @@ class ISA : public BaseISA
 
     RiscvType rvType() const { return rv_type; }
 
+    void
+    clearLoadReservation(ContextID cid)
+    {
+        Addr& load_reservation_addr = load_reservation_addrs[cid];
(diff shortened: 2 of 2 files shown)


mattsinc (approved):

mattsinc:
@KaiBatley : it looks like you forgot to add a change ID, as the second check fails with this:

None of the commits in this pull request contains a Change-ID, which we require for any changes made to gem5.  To automatically insert one, run the following:\n f=.git/hooks/commit-msg ; mkdir -p  ;  curl -Lo  https://gerrit-review.googlesource.com/tools/hooks/commit-msg ; chmod +x \n Then amend the commit with git commit --amend --no-edit, and update your pull request.

You'll have to update this before anything further can happen.

mattsinc (changes requested):
You'll also need to update the commit message to be 80 characters per line

BobbyRBruce:
The current test failing in our CI is due to the pull-request not containing a change ID. The easiest way to solve this is with the following:

```sh
n f=.git/hooks/commit-msg ; mkdir -p  ;  curl -Lo  https://gerrit-review.googlesource.com/tools/hooks/commit-msg ; chmod +x \n
git commit --amend --no-edit
```

This will add a change ID to the commit at the HEAD of the pull request. 

To avoid this going forward we strongly recommend using our pre-commit git hooks. To install execute `util/pre-commit-install.sh` in the gem5 repo. These git hooks  automatically add Change-IDs to a commit and run our style checks on code you have added or altered.

KaiBatley:
#92 Replaces this pull request to fix the issues.

PR #83, opened 2023-07-14
arch-riscv: Relation chain on RVV support 

aarmejach:
Gerrit relation chain on RVV support found here:
https://gem5-review.googlesource.com/c/public/gem5/+/68418/15


aarmejach:
The test that is failing is: `test-gem5-library-riscv-hello-restore-checkpoint-ALL-x86_64-opt`

This test is meant to fail given the changes on the vector register file done on these commits. Should we disable it?

powerjg:
We should probably fix it before we merge this change.

I think that we'll need to create a new checkpoint and update the test to use that new checkpoint. Give us a couple of days to look into this. @BobbyRBruce and @mkjost0

BobbyRBruce:
I'll work on a solution to this this afternoon. It shouldn't take too long.

These checkpoint tests are quite an overhead to maintain ...

powerjg:
@BobbyRBruce, don't forget to work on this checkpoint "fix"

BobbyRBruce:
All you need to do to get the tests passing is change this line https://github.com/gem5/gem5/blob/3c6563d6f7cf35f5e801b7a1dc9363619bb55239/configs/example/gem5_library/checkpoints/riscv-hello-restore-checkpoint.py#L93 to `checkpoint=Resource("riscv-hello-example-checkpoint")`.

Though, in doing this, I discovered the "util/cpt_upgrader.py" did not work in this case. I logged this in Issue #106. I'm not familiar with the upgrader, but we'll probably need to fix this issue as we want to give people an easy way to upgrade their checkpoints to work with this.

powerjg:
I think it would be good to implement the checkpoint upgrader #106 before merging this change. If you have questions on how to do this, let me know and I can try to help (I'm also not an expert 😄 ).

aarmejach:
Thanks Bobby, I have made the modification to the config file to use the new checkpoint restore test. Let's see if tests pass now.

I can help on the cpt_upgrader if you want, I have an idea on what needs to be done there and can take a look at it on Monday.

P.S: I had way too many develop branches on diferent remotes and made a bit of a mess on my first push, think it is fixed now.

src/arch/riscv/isa.cc
```diff
@@ -299,6 +314,8 @@ void ISA::clear()
     // mark FS is initial
     status.fs = INITIAL;
 
+    status.vs = VPUStatus::INITIAL;
```
rogerchang23424:
```suggestion
    if (rv_type == RV64) {
        status.vs = VPUStatus::INITIAL;
        misa.rvv = 1;
    }
```

vspetrov:
Hello, we've been looking into this patch and faced an issue with O3 CPU. The reproducer is quite large and depends on our internal library - we are working on making it smaller and plan to share it here. Meanwhile, i wanted to share our current view and check if the understanding of the problem is correct. 

Basically, the reproducer looks like this:
```
vsetvl (type1, vl1) //First vset vl instruction that sets custom type/vl for decode
inst_squash // any instruction that will cause a pipeline squash
vse // Any vector instruction that relies on type1,vl1 durin decode
...
vsetvl (type2, vl2) // Second vset vl with different type/vl
vse // any vector instr that uses type2/vl2
```
What happens is: O3 CPU starts executing vsetvl №1 and saves type1/vl1 on decoder structure. Then It would create inst_squash - this can be any instruction that will cause a pipeline squash in future (it can be ld/st speculation or branch for example). Then vsetvl 2 is executed BEFORE inst_squash actually triggers pipeline squash. THis saves new type/vl on decoder. This is when pipeline is actually squashed and restarted from inst_squash. However, decoding of 1st vse instruction now uses wrong type/vl values cached on decoder from 2nd vsetvl execution.

Does that view of the problem makes sense or we missing smth?

It looks like for such scenarios we might need to maintain kind of stack data structure that would be rolled back in case of pipeline squashes (smth similar to bpu history state stack). However, it will require some mechanism to pass that info from CPU to decoder. Maybe seqNum.

aarmejach:
The problem you describe makes sense to me. And the solution of rolling back decoder state on pipeline squashes is pobably a good way to fix it, and is related to ongoing discussion on how to better implement vsetvl instructions.

AFAIK the vsetvl instructions now stall younger instructions via a fake dependency, and there has been discussion to change vsetvl to behave like a control instruction (basically branch-like) to enable speculation around them. That is, speculate that the vl/vtype is the same as the previous and, if not, a rollback is done. Therefore, the described problem and the proposed solution can probably be used to rethink the vsetvl implementation and solve two problems in one shot.

To be honest i'd like to see this current PR merged ASAP, and try to make new issues describing problems we are seeing to fix them (like this vsetvl issue). This will allow much faster progress as now this PR is blocking multiple improvements. 

I'd like some thoughts on this. Of course we should go through code review, etc., but if we merge this PR it will enable faster progress.

BTW CI tests should pass now without problems.

powerjg:
My bad, @aarmejach . I'll make sure this is reviewed ASAP with a goal of getting it merged this week.

vspetrov:
@aarmejach thanks. sounds good. Just out of curiosity, do you know of any actual public RTL implementation of this vector extension and how the vsetvl states are handled there?

powerjg:
@vspetrov IIRC, the RVV standard states that `vsetvl` should be treated as a control instruction. Although, I don't have the exact citation for you.

From my understanding, if you're going to execute/schedule the vector instructions on an out-of-order core, the only option is to treat `vsetvl` as a control instruction. The performance will significantly suffer or correctness will be impacted, if you stall younger instructions after the `vsetvl`, at least with how many compilers generate code.

However, if you have a fully decoupled vector unit, then you would probably have more freedom in how you implement the `vsetvl` instruction (e.g., you could execute that instruction in the front-end). 

That said, the current implementation in gem5 is going to be using the core (out-of-order or otherwise) to execute the vector instructions. So, at least in the longer term, we will likely implement `vsetvl` as a control instruction.

In the short term, we should probably just fail (e.g., `fatal`) if trying to use the O3CPU with vector extensions enabled until we implement `vsetvl` in a more accurate way.

hnpl:
In case you're interested, the chapter 9 of the rvv spec has a note about control dependencies related to the vl register,

https://github.com/riscv/riscv-v-spec/blob/v1.0/v-spec.adoc#9-vector-memory-consistency-model

src/arch/riscv/isa.cc
```diff
@@ -234,11 +244,9 @@ namespace
 {
 
 /* Not applicable to RISCV */
-RegClass vecRegClass(VecRegClass, VecRegClassName, 1, debug::IntRegs);
-RegClass vecElemClass(VecElemClass, VecElemClassName, 2, debug::IntRegs);
-RegClass vecPredRegClass(VecPredRegClass, VecPredRegClassName, 1,
+RegClass vecPredRegClass(VecPredRegClass, VecPredRegClassName, 0,
         debug::IntRegs);
-RegClass matRegClass(MatRegClass, MatRegClassName, 1, debug::MatRegs);
+RegClass matRegClass(MatRegClass, MatRegClassName, 0, debug::MatRegs);
```
powerjg, line 250: Was this change purposeful or did it sneak in because of a rebase (the change is a 1 -> 0).

Same for `vecPredRegClass` above.

src/arch/riscv/decoder.cc
```diff
@@ -42,13 +42,21 @@ void Decoder::reset()
...
     mid = false;
+    vConfigDone = true;
     machInst = 0;
     emi = 0;
 }
 
 void
 Decoder::moreBytes(const PCStateBase &pc, Addr fetchPC)
 {
+    if (GEM5_UNLIKELY(!this->vConfigDone)) {
```
powerjg, line 63: Can we put a `TODO` here since we all agree that this is not a good long-term solution?

src/arch/riscv/isa/templates/templates.isa
powerjg, line 1: This file (and the others in this directory) are missing a LICENSE. You should be able to copy the BSD license here.

src/cpu/minor/fetch2.cc
powerjg, line 1: I think I would actually prefer to drop the minor/o3 changes and instead say that they are unsupported in this change.

We could potentially put these to changes up on a separate PR and point people to that if they insist on using minor/o3 with RVV.

powerjg (commented):
Here's an initial review. I have a few more files I'd like to go through, but I don't expect any big hurdles.

Overall, this looks great. Let's get it in and improve on it from there!

src/arch/riscv/regs/vector.hh
```diff
@@ -68,12 +69,18 @@ const std::vector<std::string> VecRegNames = {
...
 static inline TypedRegClassOps<RiscvISA::VecRegContainer> vecRegClassOps;
 
 inline constexpr RegClass vecRegClass =
     RegClass(VecRegClass, VecRegClassName, NumVecRegs, debug::VecRegs).
         ops(vecRegClassOps).
         regType<VecRegContainer>();
+inline constexpr RegClass vecElemClass =
+    RegClass(VecElemClass, VecElemClassName, NumVecRegs * NumVecElemPerVecReg,
+            debug::VecRegs).
+        ops(vecRegElemClassOps);
```
rogerchang23424: I think we should drop the vecElemClass commit because it's just solve the hanging problem for `vsetvl`, `vsetvli` and `vsetivli` in minor CPU.

src/arch/riscv/isa.cc
aarmejach: There are no specific predicate/mask registers, so `vecPredRegClass` should be empty, i.e., `0` should be ok. 

Same for `matRegClass`, until a matrix extension shows up.

src/arch/riscv/isa.cc
rogerchang23424: I suggest it because the vector only implemented for RV64 now

vmskv:
> Hello, we've been looking into this patch and faced an issue with O3 CPU. The reproducer is quite large and depends on our internal library - we are working on making it smaller and plan to share it here. Meanwhile, i wanted to share our current view and check if the understanding of the problem is correct.
> 
> Basically, the reproducer looks like this:
> 
> ```
> vsetvl (type1, vl1) //First vset vl instruction that sets custom type/vl for decode
> inst_squash // any instruction that will cause a pipeline squash
> vse // Any vector instruction that relies on type1,vl1 durin decode
> ...
> vsetvl (type2, vl2) // Second vset vl with different type/vl
> vse // any vector instr that uses type2/vl2
> ```
> 
> What happens is: O3 CPU starts executing vsetvl №1 and saves type1/vl1 on decoder structure. Then It would create inst_squash - this can be any instruction that will cause a pipeline squash in future (it can be ld/st speculation or branch for example). Then vsetvl 2 is executed BEFORE inst_squash actually triggers pipeline squash. THis saves new type/vl on decoder. This is when pipeline is actually squashed and restarted from inst_squash. However, decoding of 1st vse instruction now uses wrong type/vl values cached on decoder from 2nd vsetvl execution.
> 
> Does that view of the problem makes sense or we missing smth?
> 
> It looks like for such scenarios we might need to maintain kind of stack data structure that would be rolled back in case of pipeline squashes (smth similar to bpu history state stack). However, it will require some mechanism to pass that info from CPU to decoder. Maybe seqNum.

The shareable reproducer has been prepared:  
https://github.com/vmskv/gem5/tree/vmskv/rvv_bad_o3_vsetvl_execuion/riscv_o3_bad_vsetvl_execution_reproducer  
In out-of-order(O3CPU) mode if vsetvl is in mispredicted branch it changes the 'vl' value in decoder and the vector instructions afterwards are decoded/executed bad after the mispredicted vsetvl squashed.

src/arch/riscv/decoder.cc
aarmejach: done.

src/arch/riscv/isa/templates/templates.isa
aarmejach: done

src/cpu/minor/fetch2.cc
aarmejach: I agree. Dropped the 3 commits that were just to make minor/o3 work using the current `vset*` implementation.

I can create a PR with these 3 commits once this PR is merged. Which will be just for people that wants to use minor/o3 while we fix the `vset*` implementation.

src/arch/riscv/isa.cc
aarmejach: done, added to the switch statement.

src/arch/riscv/regs/vector.hh
aarmejach: done, i removed this one and the other 2 commits specific to minor/o3.

The drawback is that now the `restore-checkpoint` test will fail again. This has 2 implications:

1.  @BobbyRBruce could you fix the test again, basically the `regs.vector_element` field in the cpt now should be empty.
2. I will amend the PR #115 on the cpt_upgrader now to take this change into account.

aarmejach:
Thanks @vmskv. I think we should open an issue with the description of the problem and the reproducible, which will be the motivation to modify the current `vset*` implementation.

rogerchang23424:
Thanks for the update.

Cloud you rebase the branch to the top of develop tree? It has the conflict with develop branch.

Thanks

src/cpu/minor/fetch2.cc
powerjg: If you don't mind, I'm going to add a commit to this PR which throws a fatal if RVV is used with O3 or Minor CPU. (Assuming I can figure out how to do it!)

powerjg:
I went ahead and added a couple of commits to this chain. If it's OK with you, I can simply push them here. Otherwise, you can pull them from my repo: https://github.com/powerjg/gem5/commits/pr/aarmejach/83

I fixed some style problems and I added a fatal if RVV is used with o3 or minor.

src/cpu/minor/fetch2.cc
aarmejach: Sure, please go ahead!

aarmejach:
@powerjg feel free to push them here.

@BobbyRBruce a quick way to fix the restore-checkpoint test again is to run the cpt_upgrader of PR #115 over the checkpoint used in `checkpoint=Resource("riscv-hello-example-checkpoint")` that this PR is using.

Hopefully, this should get us close to merge this. I am going into vacation mode, but will monitor this to make sure we get it in asap.

src/arch/riscv/isa/decoder.isa
```diff
@@ -2012,6 +2280,2093 @@ decode QUADRANT default Unknown::unknown() {
...
+            0x7: decode BIT31 {
+                format VConfOp {
+                    0x0: vsetvli({{
+                        uint64_t rd_bits = RD;
+                        uint64_t rs1_bits = RS1;
+                        uint64_t requested_vl = Rs1_ud;
+                        uint64_t requested_vtype = zimm11;
+
+                        Rd_ud = 0;
+                    }}, VectorConfigOp, IsDirectControl, IsCondControl);
```
rogerchang23424:
```suggestion
                    }}, VectorConfigOp, IsSerializeAfter, IsNonSpeculative);
```
We should treat vset\*vl\* like csrrw instruction

src/arch/riscv/isa/decoder.isa
```diff
@@ -2012,6 +2280,2093 @@ decode QUADRANT default Unknown::unknown() {
...
+                    0x1: decode BIT30 {
+                        0x0: vsetvl({{
+                            uint64_t rd_bits = RD;
+                            uint64_t rs1_bits = RS1;
+                            uint64_t requested_vl = Rs1_ud;
+                            uint64_t requested_vtype = Rs2_ud;
+
+                            Rd_ud = 0;
+                        }}, VectorConfigOp, IsDirectControl, IsCondControl);
```
rogerchang23424: ditto

src/arch/riscv/isa/decoder.isa
```diff
@@ -2012,6 +2280,2093 @@ decode QUADRANT default Unknown::unknown() {
...
+                        0x1: vsetivli({{
+                            uint64_t rd_bits = RD;
+                            uint64_t rs1_bits = -1;
+                            uint64_t requested_vl = uimm;
+                            uint64_t requested_vtype = zimm10;
+
+                            Rd_ud = 0;
+                        }}, VectorConfigOp, IsDirectControl, IsCondControl);
```
rogerchang23424: ditto

rogerchang23424 (changes requested):

vmskv:
> Thanks @vmskv. I think we should open an issue with the description of the problem and the reproducible, which will be the motivation to modify the current `vset*` implementation.

The issue is opened https://github.com/gem5/gem5/issues/144

rogerchang23424:
Hi @powerjg,

The additional commits are LGTM. However, I think it will be better to move the `enable_rvv` option from `RiscvDecorder` to `RiscvISA` as it is one of isa configuration. There are two benefits of this.

1. Some of isa options `riscv_type` and `check_alignment` are put in `RiscvISA`. If the `enable_rvv` is in `RiscvISA`, it will be easier for user to configure.
2. The initialization of `mstatus` and `misa` CSR are depending on the `enable_rvv` option.

(nit) Riscv decoder can get `enable_rvv` option by `(dynamic_cast<ISA*>(isa))->enableRvv();` .

src/arch/riscv/isa/decoder.isa
aarmejach: Yes, in this current implementation we can treat them like `csrrw`. In future speculative implementations we can go back to the branch-like semantics.

powerjg:
Thanks for the feedback @rogerchang23424, I took your suggestions into account.

@aarmejach, I had to force push. I hope that doesn't cause a big headache for you.

There's one more comment to address from Roger and we need to figure out the checkpoint. After that, I think we're good to go.

src/arch/riscv/regs/vector.hh
BobbyRBruce: > done, i removed this one and the other 2 commits specific to minor/o3.
> 
> The drawback is that now the `restore-checkpoint` test will fail again. This has 2 implications:
> 
> 1. @BobbyRBruce could you fix the test again, basically the `regs.vector_element` field in the cpt now should be empty.
> 2. I will amend the PR [util: fix cpt upgrader for rvv changes in PR #83 #115](https://github.com/gem5/gem5/pull/115) on the cpt_upgrader now to take this change into account.

I've updated the restore checkpoint. It should work with the PR as it currently stands.

rogerchang23424 (approved):
Let Jason and Bobby do final review

aarmejach:
Sure, all comments are addressed and tests should pass.

Once a final review is done feel free to hit the button.

powerjg (approved):
Thanks @aarmejach for pushing this through!

powerjg:
Thanks for the hard work everyone!

PR #91, opened 2023-07-14
stdlib: Change resource compatibility warning

kunpai:
If the gem5 version is "develop", the warning will not be thrown.

Change-Id: Id2be1c4323c6ca06c5503c2885c1608f8d119420


src/python/gem5/resources/client_api/client_wrapper.py
```diff
@@ -289,15 +289,16 @@ def _check_resource_version_compatibility(
...
-                f"Resource {resource['id']} with version "
-                f"{resource['resource_version']} is not known to be compatible"
-                f" with gem5 version {gem5_version}. "
-                "This may cause problems with your simulation. "
-                "This resource's compatibility "
-                "with different gem5 versions can be found here: "
-                "https://resources.gem5.org"
-                f"/resources/{resource['id']}/versions"
-            )
+            if not gem5_version.upper().startswith("DEVELOP"):
```
powerjg, line 293: I would put this in the and above instead of a new line.

I.e., `if gem5_version and not gem5_version.upper().startswith("DEVELOP") and not self._...`

src/python/gem5/resources/client_api/client_wrapper.py
BobbyRBruce, line 1: If you look at line 238 of this file you'll see how the compatibility check works. I believe if you change the gem5 Version to "23.0.0.1-FOR-DEVELOP-23.1" then this will work. (`if gem5_version.startswith(version):` --- this returns true, no?

BobbyRBruce (changes requested):

src/python/gem5/resources/client_api/client_wrapper.py
kunpai: Yes, "23.0.0.1-FOR-DEVELOP-23.1" does not throw the warning. I guess the question I want to ask here is: do we want to disable this compatibility warning for the develop branch as a whole?

powerjg: I guess I'm OK with changing the develop version string... Why don't you like changing the check Bobby?

What happens when we change the version to 23.0.0.2?

powerjg:
@BobbyRBruce, please re-review.

src/python/gem5/resources/client_api/client_wrapper.py
BobbyRBruce: I don't think it'd matter as the resources with version "YY.{Major}" would still match. It'd only be a problem for resources which specified compatibility down to the hotfix.

I'm overly invested in my alternative, I just didn't like adding a hardcorded magic string here. It means that we now have to keep in mind the version on develop must contain the words "develop". Not "dev" or anything like that. But, it's a small requirement and lack of adherence to it doesn't break anything. It just produces noisy output, so it's not a big deal.

src/python/gem5/resources/client_api/client_wrapper.py
BobbyRBruce: In the interests of getting this in, I'll let this slide. Even though it's inelegant, it's still understandable.

BobbyRBruce (dismissed):

BobbyRBruce (approved):

PR #96, opened 2023-07-17
misc: Add workflow to close stale issues

powerjg:
Create a new workflow file that will hold jobs that are for managing the
repository, issues, prs, etc. This changeset then adds a job to close
issues that have been open for 30 days without a response someone marks
the issue as "needs details."

Let me know if I shouldn't have created this PR on stable. I'm still a bit confused about develop vs stable for these files :).

I haven't tested this, but I think it's straightforward. The only thing I'm not sure about is the permissions on the token.

### DIFF
--- a/.github/workflows/utils.yaml
+++ b/.github/workflows/utils.yaml
@@ -1,0 +1,19 @@
+# This workflow file contains miscellaneous tasks to manage the repository.
+name: Utils for Repository
+on:
+  schedule:
+    - cron: '30 1 * * *'
+  workflow_dispatch:
+
+jobs:
+  # This job runs the stale action to close issues that have been inactive for 30 days.
+  # It is scheduled to run every day at 1:30 AM UTC.
+  close-stale-issues:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v8.0.0
+        with:
+          close-issue-message: 'This issue is being closed because it has been inactive waiting for response for 30 days. If this is still an issue, please open a new issue and reference this one.'
+          days-before-stale: 21
+          days-before-close: 7
+          any-of-labels: 'needs details'


powerjg:
Documentation on this action can be found here: https://github.com/marketplace/actions/close-stale-issues

powerjg:
@mkjost0 any idea why the change-id test fails? I have a change id, from what I can tell.

mkjost0:
I'm actually not sure, at some point I was only checking the very last line of the commit message so I thought that might be why, but the check we're currently using should scan the entire commit message for a Change-Id, and your formatting looks right.  I'll try looking into it on my end, since even though we should scan the whole message, the only tests I ever did had the Change-Id set as the very last line.

mkjost0:
Nevermind actually, the issue is that the ci-tests.yaml is different on stable and develop, and your PR is directed towards stable.  I think since Bobby mentioned wanting to create a hotfix soon, we can move this PR to develop, but the way we iterate through commits right now on stable is out-of-date

BobbyRBruce:
Unless this is urgent (must be added in the next couple of days), can we put this on the `develop` branch instead? Though informal, I do like the following policies:

1. Stable should not be altered outside of a release as much as is possible.
2. Files within the `.github` directory can be modified between releases but the changes in the `.github` directory on `stable` should never be ahead of `develop`. Ergo, changes should first be committed to `develop` prior to being placed on `stable`.

I've got 2 or 3 other things that need to be put on stable. A hotfix request and I want to do a minor release with documentation updates and some other small non-functional things.

I'm going to get the ball rolling with that today, and hopefully have something on stable by the end of the week. Given that timeline, is it ok for this to be put on develop now?

BobbyRBruce:
@powerjg : Can you please push a patch to the develop branch?

When you do so, log it here and i'll close this PR in favor of that.

I tried editing this PR to target the develop branch but it didn't do so cleanly.

powerjg:
Hopefully this looks better. LMK if there's anything else I need to do.

BobbyRBruce (approved):
Let's give this a go.

I see nothing wrong with this but it's best tested in the wild.

Thumbs-up from me.

PR #98, opened 2023-07-18
arch-riscv: Set default check alignment True

rogerchang23424:
Raise misaligned trap if effective address if not aligned by default

Change-Id: I634aa7ddbf5282fc583316fc77ab1e37bfe415e3

### DIFF
--- a/src/arch/riscv/RiscvISA.py
+++ b/src/arch/riscv/RiscvISA.py
@@ -53,6 +53,6 @@ class RiscvISA(BaseISA):
     cxx_header = "arch/riscv/isa.hh"
 
     check_alignment = Param.Bool(
-        False, "whether to check memory access alignment"
+        True, "whether to check memory access alignment"
     )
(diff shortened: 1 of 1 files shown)


powerjg:
Any idea why this was false before? 

Also, should this really be an option? If the ISA says unaligned accesses cause faults, then we should do it and there should not be an option to incorrectly implement the ISA.

rogerchang23424:
In the RV32I section of ISA spec, the misaligned load and store handling is depending on the EEI. It can be handled invisible or generating misalignment trap to let software handle misalignment load & store. In gem5. If check_alignment is False, then we prefer to handle invisible for misaligned load & store, otherwise, we prefer to generating trap for them. I think the option need to be kept as both are match spec.

The change will generate trap for misaligned load & store by default unless users set check_alignment = False, and it can pass the misaligned load & store test of riscv-arch-test with spike and sail. They set former by default.

powerjg (approved):

PR #76, opened 2023-07-13
Find lsb set generalization and optimization

MeatBoy106:

PR #99, opened 2023-07-18
base: Added missing backup dummy __has_builtin definition

DanKouch:
Added dummy definition of __has_builtin to bitfield.hh's hasBuiltinCtz, which is already being done in popCount.

Change-Id: I4a1760a142209462bb807c6df4bc868284b6f5f3

### DIFF
--- a/src/base/bitfield.hh
+++ b/src/base/bitfield.hh
@@ -310,6 +310,13 @@ namespace {
 template<typename T>
 constexpr bool
 hasBuiltinCtz() {
+// Since the defined(__has_builtin) in the subsequent #if statement
+// won't short-circuit the macro expansion of
+// __has_builtin(__builtin_ctz), we must explicitly define it as zero
+// if it's undefined to avoid a preprocessor error.
+#ifndef __has_builtin
+#   define __has_builtin(foo) 0
+#endif
 #if defined(__has_builtin) && __has_builtin(__builtin_ctz)
     return sizeof(unsigned long long) >= sizeof(T);
 #else


DanKouch:
Without this I am unable to build for VEGA_X86 due to a preprocessor error. I believe this was introduced in #76.

The added check was taken directly from elsewhere in bitfield.hh.

mattsinc (commented):
This seems good to me, but I would like Bobby or Jason to also check before we merge this in.

BobbyRBruce (commented):
Is there an explanation for this bug? According to https://gcc.gnu.org/onlinedocs/gcc-7.5.0/cpp/Defined.html  "#if defined MACRO is precisely equivalent to #ifdef MACRO". I don't `defined` .

How can `ifndef __has_builtin == defined(__has_builtin)`? That statement must be true in your build for this bug to exist, right? Or am I missing something?

DanKouch:
@BobbyRBruce I believe the `defined` check isn't short circuiting the right hand side of the `&&` (where the macro is being used) from being expanded, which causes problems due to that macro not being defined.

BobbyRBruce:
> @BobbyRBruce I believe the `defined` check isn't short circuiting the right hand side of the `&&` (where the macro is being used) from being expanded, which causes problems due to that macro not being defined.

Ah, ok. I can believe that. To avoid anyone accidentally refactoring this away in the future (because I feel this is something I would do), can you add a comment explaining this in the code?

DanKouch:
> > @BobbyRBruce I believe the `defined` check isn't short circuiting the right hand side of the `&&` (where the macro is being used) from being expanded, which causes problems due to that macro not being defined.
> 
> Ah, ok. I can believe that. To avoid anyone accidentally refactoring this away in the future (because I feel this is something I would do), can you add a comment explaining this in the code?

Good point, I added a comment.

BobbyRBruce (approved):

PR #101, opened 2023-07-19
mem-ruby: Added support for non-system-scope atomics in VIPER

DanKouch:
Added support for performing non-SLC-set atomics in the TCC. Previously, all atomics were being passed on by the TCC to the directory. With this change, atomics will only be passed on if the SLC bit is set or if the line isn't present or available in the TCC.

If a non-SLC atomic is passed on to the directory because it is not present in the TCC, the atomic will be performed on the return path on the Data event. To accommodate the directory not performing the atomic in this case, this change also passes the SLC bit on to the directory.

The previously-named "Atomic" action has been renamed to "AtomicPassOn", with the new "Atomic" corresponding to an atomic performed directly in the TCC.

Change-Id: Ibf92f71ddceb38bd1b0da70b0a786cc4c3cf2669

### DIFF
--- a/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
@@ -61,6 +61,7 @@ machine(MachineType:TCC, "TCC Cache")
     WrVicBlkBack,           desc="L1 Write Through(dirty cache)";
     WrVicBlkEvict,          desc="L1 Write Through(dirty cache) and evict";
     Atomic,                 desc="Atomic Op";
+    AtomicPassOn,           desc="Atomic Op Passed on to Directory";
     AtomicDone,             desc="AtomicOps Complete";
     AtomicNotDone,          desc="AtomicOps not Complete";
     Data,                   desc="data messgae";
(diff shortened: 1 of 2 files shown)


src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -699,7 +728,7 @@ machine(MachineType:TCC, "TCC Cache")
   // Stalling transitions do NOT check the tag array...and if they do,
   // they can cause a resource stall deadlock!
 
-  transition(WI, {RdBlk, WrVicBlk, Atomic, WrVicBlkBack}) { //TagArrayRead} {
+  transition(WI, {RdBlk, WrVicBlk, Atomic, AtomicPassOn, WrVicBlkBack}) { //TagArrayRead} {
```
mattsinc, line 731: Do we want TagArrayRead for this?

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -717,18 +746,21 @@ machine(MachineType:TCC, "TCC Cache")
       // woken up
       st_stallAndWaitRequest;
   }
-  transition(IV, {WrVicBlk, Atomic, WrVicBlkBack}) { //TagArrayRead} {
+
+  transition(IV, {WrVicBlk, Atomic, AtomicPassOn, WrVicBlkBack}) { //TagArrayRead} {
```
mattsinc, line 750: Here too, don't we want TagArrayRead for them?

mattsinc (commented):
Two minor questions for you Daniel (although they are both technically there already before your patch, it seems like we want them).

@Matt P: I have already reviewed this internally and approved (besides the above).  So I'd prefer if you look it over before we merge it in, since I already have.

abmerop (approved):
Looks good.  I don't have a preference regarding the tag resource constraint since it is using what was there before.

mattsinc:
> Looks good. I don't have a preference regarding the tag resource constraint since it is using what was there before.

Sure, for this patch I think it's fine.  But that does seem like a (separate) bug, no?

abmerop:
I looked at this and other protocols and it seems that generally transitions that stall/stallAndWait/recycle do not have any resources constraints.  Based on that, I would say the resource constraints in this PR would be extraneous comments that could be removed.

DanKouch:
Removed those comments as extraneous per @abmerop's comment. There are additional cases where a commented-out TagArrayRead resource constraint is used, but those are on untouched transitions. Should I make another PR to remove those?

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
DanKouch: Removed as extraneous per @abmerop's comment.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
DanKouch: Removed as extraneous per @abmerop's comment.

mattsinc:
> Removed those comments as extraneous per @abmerop's comment. There are additional cases where a commented-out TagArrayRead resource constraint is used, but those are on untouched transitions. Should I make another PR to remove those?

I think having a separate commit for them is fine, yeah.  Thanks @DanKouch !

mattsinc:
> I looked at this and other protocols and it seems that generally transitions that stall/stallAndWait/recycle do not have any resources constraints. Based on that, I would say the resource constraints in this PR would be extraneous comments that could be removed.

Interesting, but good to know.

mattsinc (approved):

PR #102, opened 2023-07-19
stdlib,configs,tests: Remove deprecated Resource classes usage

BobbyRBruce:
The following resource classes are deprecated:

* `Resource`: Replaced with `obtain_resource`.
* `CustomResource`: Replaced with specializations.
* `CustomDiskImage`: Replaced with `DiskImageResource`.

There was still some usage in our codebase. In our "configs/example" scripts, in our tests, and some references, in comments, in the main gem5 source. This removes them.

This stops gem5 from regularly throwing warnings as we're using deprecated classes.

### DIFF
--- a/src/python/gem5/resources/workload.py
+++ b/src/python/gem5/resources/workload.py
@@ -52,7 +52,7 @@ class AbstractWorkload:
     workload = CustomWorkload(
         function = "set_se_binary_workload",
         parameters = {
-            "binary" : Resource("x86-print-this"),
+            "binary" : obtain_resource("x86-print-this"),
             "arguments" : ["hello", 6]
         },
     )
@@ -64,7 +64,7 @@ class AbstractWorkload:
 
     ```py
     board.set_se_binary_workload(
-        binary = Resource("x86-print-this"),
+        binary = obtain_resource("x86-print-this"),
         arguments = ["hello", 6],
     )
     ```
@@ -121,7 +121,7 @@ class CustomWorkload(AbstractWorkload):
     workload = CustomWorkload(
         function = "set_se_binary_workload",
         parameters = {
-            "binary" : Resource("x86-print-this"),
+            "binary" : obtain_resource("x86-print-this"),
             "arguments" : ["hello", 6]
         },
     )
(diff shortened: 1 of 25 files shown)


src/python/gem5/resources/workload.py
```diff
@@ -148,7 +148,8 @@ class Workload(AbstractWorkload):
 
     # Optionally we can override a parameter in the workload. In this example
     # we are going to run this workload with a difference kernel.
-    workload.set_parameter("kernel", Resource("arm64-linux-kernel-4.14.134"))
+    workload.set_parameter("kernel", \
+        obtain_resource("arm64-linux-kernel-4.14.134"))
```
powerjg: I don't think the backslash is needed. I'm not sure how this got passed black...

```suggestion
    workload.set_parameter("kernel",
        obtain_resource("arm64-linux-kernel-4.14.134")
    )
```

tests/gem5/configs/arm_boot_exit_run.py
```diff
@@ -199,15 +199,15 @@
 
 # Set the Full System workload.
 board.set_kernel_disk_workload(
```
powerjg, line 201: Does it make sense to pack this up into a workload?

tests/gem5/configs/boot_kvm_fork_run.py
```diff
@@ -179,10 +179,10 @@
 
 # Set the Full System workload.
 motherboard.set_kernel_disk_workload(
-    kernel=Resource(
+    kernel=obtain_resource(
```
powerjg, line 182: Same thing here... should this be a workload?

powerjg (commented):
A couple of small things.

src/python/gem5/resources/workload.py
BobbyRBruce: This code is in a docstring. It's an example of usage. Black probably just ignores it.

Regardless, I'll incorporate this.

tests/gem5/configs/arm_boot_exit_run.py
BobbyRBruce: Maybe, but I also like to have some "raw" set_workload examples around as well. Both are helpful.

Either way, I think I can do this in another PR if required?

tests/gem5/configs/boot_kvm_fork_run.py
BobbyRBruce: Same comment as before. It could be but I think having some set_workload examples is good too. I don't think it's in the scope of this patch to do that conversion.

powerjg (approved):

PR #103, opened 2023-07-19
misc: Merge cherry-picked commits from develop

BobbyRBruce:
This merges cherry-picked changes from the develop branch which are suitable to be part of the v23.0.1.0 minor release. In general this includes bug fixes, documentation updates, and changes that extend gem5 but do not alter and functionality a gem5 user may be depending on.


BobbyRBruce:
@mkjost0 : Is it possible to enable traditional merges for contributions to branches other than `develop`? In this case it doesn't make any sense and would be disruptive. It's only on develop we want to enforce squashes.

mkjost0:
As far as I can tell at the moment, rules regarding merges are only configurable on the repository as a whole, not by branch.

BobbyRBruce:
> As far as I can tell at the moment, rules regarding merges are only configurable on the repository as a whole, not by branch.

Yeah, seems like you're right: https://github.com/orgs/community/discussions/10809. That's rather annoying. I guess I can just quickly enable merging and put this in and turn it off. With all these branch production rules I'm surprised they make this a the one thing that's repo-wide.

powerjg:
Does #122 override this?

BobbyRBruce:
> Does #122 override this?

No. #122 only copies the .github file over to the minor release. None of that patches i've cherry-picked here modify the .github file.

BobbyRBruce:
> 

Oh wait, I see what you mean now. I messed something up somewhere... Thanks for pointing this out :)

BobbyRBruce:
@giactra : The patch has been cherry-picked here. It'll be part of the minor release.

PR #105, opened 2023-07-20
misc: Split up tests in daily-tests.yaml

mkjost0:
This splits up the gem5 library example tests by Suite UID, as right now running them together uses the runner for a long period of time.  It is important to note that doing this means additional tests from this directory will need to be manually added, such as the kvm tests.

Change-Id: Ib2a0aca08f9b51b60e9dd0528324372cf2d98c05

### DIFF
--- a/.github/workflows/daily-tests.yaml
+++ b/.github/workflows/daily-tests.yaml
@@ -160,6 +160,52 @@ jobs:
         rm -rf ~/.cache || true
     - run: echo "This job's status is ${{ job.status }}."
 
+  # split library example tests into runs based on Suite UID
+  # so that they don't hog the runners for too long
+  testlib-long-gem5_library_example_tests:
+    runs-on: [self-hosted, linux, x64, run]
+    strategy:
+      matrix:
+        test-type: [gem5-library-example-x86-ubuntu-run-ALL-x86_64-opt, gem5-library-example-riscv-ubuntu-run-ALL-x86_64-opt, lupv-example-ALL-x86_64-opt, gem5-library-example-arm-ubuntu-run-test-ALL-x86_64-opt, gem5-library-example-riscvmatched-hello-ALL-x86_64-opt]
+    container: gcr.io/gem5-test/ubuntu-22.04_all-dependencies:latest
+    needs: build-gem5
+    timeout-minutes: 1440 # 24 hours
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        # Scheduled workflows run on the default branch by default. We
+        # therefore need to explicitly checkout the develop branch.
+        ref: develop
+    - uses: actions/download-artifact@v3
+      with:
+        name: ${{needs.name-artifacts.outputs.build-name}}ALL
+        path: build/ALL
+    - run: chmod u+x build/ALL/gem5.opt
+    - name: long ${{ matrix.test-type }} gem5_library_example_tests
+      working-directory: ${{ github.workspace }}/tests
+      run: ./main.py run --uid SuiteUID:tests/gem5/gem5_library_example_tests/test_gem5_library_examples.py:test-${{ matrix.test-type }} --length=long --skip-build -vv
+    - name: create zip of results
+      if: success() || failure()
+      run: |
+        apt-get -y install zip
+        zip -r output.zip tests/testing-results
+    - name: upload zip
+      if: success() || failure()
+      uses: actions/upload-artifact@v3
+      env:
+        MY_STEP_VAR: ${{github.job}}_COMMIT.${{github.sha}}_RUN.${{github.run_id}}_ATTEMPT.${{github.run_attempt}}
+      with:
+        name: ${{ env.MY_STEP_VAR }}
+        path: output.zip
+        retention-days: 7
+    - name: Clean runner
+      if: success() || failure()
+      run:
+        rm -rf ./* || true
+        rm -rf ./.??* || true
+        rm -rf ~/.cache || true
+    - run: echo "This job's status is ${{ job.status }}."
+
   # This runs the SST-gem5 integration compilation and tests it with
   # ext/sst/sst/example.py.
   sst-test:
--- a/.github/workflows/daily-tests.yaml
+++ b/.github/workflows/daily-tests.yaml
@@ -342,7 +342,9 @@ jobs:
         rm -rf ~/.cache || true
     - run: echo "This job's status is ${{ job.status }}."
 
-  testlib-long-gem5_library_example_tests:
+  # split library example tests into runs based on Suite UID
+  # so that they don't hog the runners for too long
+  testlib-long-gem5_library_example_tests_x86_ubuntu_run_ALL_x86_64_opt:
     runs-on: [self-hosted, linux, x64, run]
     container: gcr.io/gem5-test/ubuntu-22.04_all-dependencies:latest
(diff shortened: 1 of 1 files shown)


.github/workflows/daily-tests.yaml
```diff
@@ -363,9 +365,189 @@ jobs:
...
+    - name: create zip of results
+      if: success() || failure()
+      run: |
+        apt-get -y install zip
+        zip -r output.zip tests/testing-results
+    - name: upload zip
+      if: success() || failure()
+      uses: actions/upload-artifact@v3
+      env:
+        MY_STEP_VAR: ${{github.job}}_COMMIT.${{github.sha}}_RUN.${{github.run_id}}_ATTEMPT.${{github.run_attempt}}
```
powerjg: Does this say "gem5_library_example_tests_x86_ubuntu_run_ALL_x86_64_opt" somewhere in the filename?

mkjost0: Yeah, ${{github.job}} does that, so we don't have duplicate names in the artifact output.

powerjg:
Let's clean up this file by using matrix before merging.

.github/workflows/daily-tests.yaml
```diff
@@ -172,6 +172,57 @@ jobs:
...
+        path: build/ALL
+    - run: chmod u+x build/ALL/gem5.opt
+    - uses: actions/download-artifact@v3
+      with:
+        name: ${{needs.name-artifacts.outputs.build-name}}ALL_MESI_Two_Level
+        path: build/ALL_MESI_Two_Level
+    - run: chmod u+x build/ALL_MESI_Two_Level/gem5.opt
+    - name: long ${{ matrix.test-type }} gem5_library_example_tests
+      working-directory: ${{ github.workspace }}/tests
+      run: ./main.py run --uid SuiteUID:tests/gem5/gem5_library_example_tests/test_gem5_library_examples.py:test-gem5-library-example-${{ matrix.test-type }} --length=long --skip-build -vv
```
BobbyRBruce: I don't like how brittle this is, as if we rename any of these directories or even the test, things will break, but I can't see a better way to do it, so I won't block on this. It's just something to keep in mind.

.github/workflows/daily-tests.yaml
```diff
@@ -8,404 +8,148 @@ on:
...
         with:
-          name: ${{ steps.artifact-name.outputs.name }}
-          path: build/ALL_MESI_Two_Level/gem5.opt
+          name: ${{ needs.name-artifacts.outputs.build-name }}${{ matrix.image }}
+          path: build/${{ matrix.image }}/gem5.opt
           retention-days: 5
       - run: echo "This job's status is ${{ job.status }}."
 
+  # NULL build added separately since NULL is a keyword in yaml files
   build-gem5-NULL:
```
BobbyRBruce: You explained to be earlier that you separated this out to it's own separate build stop because "NULL" is keyword. Are you sure you can't just put this as a string in the `image` matrix? E.g., `image: ["NULL", "ALL"...`.

.github/workflows/daily-tests.yaml
```diff
@@ -8,404 +8,128 @@ on:
...
-      - run: echo "This job's status is ${{ job.status }}."
+    - uses: actions/checkout@v2
+    - id: artifact-name
+      run: echo "name=$(date +"%Y-%m-%d_%H.%M.%S-")" >> $GITHUB_OUTPUT
 
-  build-gem5-VEGA_X86:
+  build-gem5:
+    strategy:
+      matrix:
+        image: [ALL, ALL_CHI, ARM, ALL_MSI, ALL_MESI_Two_Level, "NULL", NULL_MI_example, RISCV, VEGA_X86]
```
powerjg, line 24: Can you add a comment on why `"NULL"` needs quotes?

.github/workflows/daily-tests.yaml
```diff
@@ -8,404 +8,128 @@ on:
...
+        include:
```
powerjg, line 28: A comment here would be helpful. I think I get it, but someone new to this file may be confused.

.github/workflows/daily-tests.yaml
```diff
@@ -8,404 +8,128 @@ on:
...
-          name: ${{ steps.artifact-name.outputs.name }}
-          path: build/VEGA_X86/gem5.opt
+          name: ${{ needs.name-artifacts.outputs.build-name }}${{ matrix.image }}
+          path: build/${{ matrix.image }}/gem5.opt
           retention-days: 5
       - run: echo "This job's status is ${{ job.status }}."
 
-
-  # This runs the unit tests for the build/ALL/unittests.debug build.
   unittests-all-debug:
```
powerjg: This is not both debug and fast

.github/workflows/daily-tests.yaml
```diff
@@ -8,404 +8,128 @@ on:
...
+  # start running all of the long tests
+  testlib-long-tests:
+    strategy:
+      matrix:
+        test-type: [arm-boot-tests, fs, gem5_library_example_tests, gpu, insttest_se, learning_gem5, m5threads_test_atomic, memory, multi_isa, replacement-policies, riscv-boot-tests, stdlib, x86-boot-tests]
     runs-on: [self-hosted, linux, x64, run]
     container: gcr.io/gem5-test/ubuntu-22.04_all-dependencies:latest
-    needs: [build-gem5, build-gem5-ALL_CHI]
+    needs: [name-artifacts, build-gem5]
     timeout-minutes: 1440 # 24 hours
```
powerjg: Is this 24 hours *each* or total?

.github/workflows/daily-tests.yaml
```diff
@@ -8,404 +8,128 @@ on:
...
     steps:
     - uses: actions/checkout@v3
       with:
         # Scheduled workflows run on the default branch by default. We
         # therefore need to explicitly checkout the develop branch.
         ref: develop
+    # download all artifacts for each test
```
powerjg, line 88: Do you need to download all of them for all tests? If so, can you explain why in a comment?

powerjg (commented):
Generally, this looks good. It's a bit confusing, so I think it could use more comments before merging.

.github/workflows/daily-tests.yaml
mkjost0: This one I'll explain here as well, but because we're now using a matrix, none of the long tests can start until all of the versions of gem5 within the matrix have been built.  I was toying around with the idea of still only downloading the necessary artifacts for each test, but the ways I was thinking of to do that would make this file a lot less readable than if I were to just download everything, plus it doesn't take significantly longer to wait for everything else to download.  There were also marketplace actions to download multiple artifacts in one step (which I think using would allow me to only download necessary artifacts but in a readable way), but they weren't from GitHub so I opted not to use those either

BobbyRBruce:
I think you've accidently included [cpu-kvm: Make using perf when using KVM CPU optional](https://github.com/gem5/gem5/pull/105/commits/188d3390d7312a1ae399a7661c9262f8958de29f) into this PR. I'd do `git rebase -i HEAD~5` and remove this change then force push with `git push -f`.

BobbyRBruce (approved):
I think this is fine. I don't like the 'name-artifacts' job but I accept this may be the easiest way to do it.

PR #107, opened 2023-07-20
cpu-minor: Check pc valid before printing

### LINKED ISSUE #97: printing bubble instruction in Minor CPU causes segfault
It appears that printing a bubble instruction in the Minor CPU causes a segfault. 

Steps to reproduce:
1. Run gem5 with the Minor CPU and the 'Branch' or 'Fetch' debug flags.
2. Cause an interrupt so that [minor::Execute::takeInterrupt](https://github.com/gem5/gem5/blob/af72b9ba580546ac12ce05bfaac3fd53fa8699f4/src/cpu/minor/execute.cc#L422) is called.
3. Observe a segfault at [src/cpu/minor/dyn_inst.cc:116](https://github.com/gem5/gem5/blob/af72b9ba580546ac12ce05bfaac3fd53fa8699f4/src/cpu/minor/dyn_inst.cc#L116)

If the 'Branch' debug flag is used, then the segfault is caused by the DPRINTF statement at [src/cpu/minor/execute.cc:320](https://github.com/gem5/gem5/blob/af72b9ba580546ac12ce05bfaac3fd53fa8699f4/src/cpu/minor/execute.cc#L320); otherwise, if the 'Fetch' debug flag is used, then the segfault is caused by the DPRINTF statement at [src/cpu/minor/fetch1.cc:513](https://github.com/gem5/gem5/blob/af72b9ba580546ac12ce05bfaac3fd53fa8699f4/src/cpu/minor/fetch1.cc#L513).

I am using gem5 version 22.1.0.0, so I apologize if this is already fixed in the current version.

powerjg:
In https://gem5-review.googlesource.com/c/public/gem5/+/52047 inst.pc was changed from an object to a pointer. It is possible that this pointer is null (e.g., if there is an interrupt and there is a bubble). Make sure to check that it's not null before printing.

I believe that other places this pointer is dereferenced without an explicit null check are safe, but I'm not certain.

Should fix #97 

Change-Id: Idbe246cfdb62d4d75416d41b451fb3c076233bbc


src/cpu/minor/dyn_inst.cc
```diff
@@ -169,7 +174,7 @@ MinorDynInst::minorTraceInst(const Named &named_object) const
 {
     if (isFault()) {
         minorInst(named_object, "id=F;%s addr=0x%x fault=\"%s\"\n",
-            id, pc->instAddr(), fault->name());
+            id, pc ? pc->instAddr() : 0, fault->name());
```
BobbyRBruce, line 177: Shouldn't this be `pc->instAddr() ? pc->instAddr() : 0`? This change isn't just setting to 0 if `pc->instAddr` is NULL, it's setting to `pc` when it's not.

src/cpu/minor/dyn_inst.cc
```diff
@@ -209,7 +214,7 @@ MinorDynInst::minorTraceInst(const Named &named_object) const
 
         minorInst(named_object, "id=%s addr=0x%x inst=\"%s\" class=%s"
             " flags=\"%s\"%s%s\n",
-            id, pc->instAddr(),
+            id, pc ? pc->instAddr() : 0,
```
BobbyRBruce, line 217: Same here. I assume I'm missing something. I'm just asking.

src/cpu/minor/dyn_inst.cc
powerjg: No, the problem is that `pc` is was made into a `unique_ptr`, so you need to check to make sure that it's non-null before dereferencing.

The goal is to print "0" if `pc` is null (because we do not know what `pc` is) and print `pc->instAddr()` otherwise.

BobbyRBruce: Oh duh. I now get what my confusion is. I've been spending too much time as Python. I was parsing that condition as `pc if pc->instAdd() else 0`. 

Ok, This is all fine then.

BobbyRBruce (approved):

powerjg:
It was failing due to not being able to get the filelock for the checkpoint-restore test. I'm going to try to update to the latest develop to see if it works now.

```
FileLockException: Timeout occured. This is likely due to the existence of the lock file '/github/home/.cache/gem5/simpoints-se-checkpoints-v23-0-v1.lock.lock'. If there's no other process the lock file, you can manually delete the lock file and rerun the script.

At:
  src/python/gem5/utils/filelock.py(85): acquire
  src/python/gem5/utils/filelock.py(107): __enter__
  src/python/gem5/resources/downloader.py(248): get_resource
  src/python/gem5/resources/resource.py(631): obtain_resource
  /__w/gem5/gem5/configs/example/gem5_library/checkpoints/simpoints-se-restore.py(122): <module>
```

powerjg:
Any reason not to merge this?

src/cpu/minor/dyn_inst.cc
```diff
@@ -112,6 +112,11 @@ MinorDynInst::reportData(std::ostream &os) const
 std::ostream &
 operator <<(std::ostream &os, const MinorDynInst &inst)
 {
+    if (!inst.pc) {
```
andysan, line 115: Are there cases where there isn't a pc associated with a dynamic instruction but still a fault or something else we want to print? I can't think of anything obvious, but you could imagine some sort of synthetic dynamic instruction going through the pipeline to raise an interrupt or something similar.

powerjg: If there is no PC, then it prints `0x????????`. If, in the future, someone wants to change this, it should be easy to track it down.

This is a clear improvement over the current code which generates a segfault if there is no PC.

Let me know if you have a specific suggestion for improvement.

PR #108, opened 2023-07-20
util,misc: Replace MAINTAINERS.yaml with TAGS.yaml

BobbyRBruce:


TAGS.yaml
powerjg, line 1: Is there any way to link this with the tags in github? Maybe https://github.com/marketplace/actions/github-labeler

powerjg (commented):
Overall this is good with me. I'll let this marinate for a few days for others to chime in, though.

abmerop (approved):
Some kind of "auto-tagger" that Jason mentioned would be cool.  As far as this PR is concerned though LGTM.

TAGS.yaml
BobbyRBruce: It's a nice idea but i kept the tags and labels separate because the tags ar short to keep them in the header, while i expanded some of the labels out to be clearer.  So even now it's not 1-to-1. We also have some labels that'll never be applicable to a commit header such as "good-first-contribution".

Definitely worth considering though if we make an update to tags at some point.

BobbyRBruce:
@andysan : I'd appreciate your feedback on this change.

andysan:
I'm not sure I follow the purpose of this change. We presumably still want to have dedicated subsystem maintainers and we need to document that somewhere (ideally in the source repository to ensure that it is propagated in clones). Looking at the diff itself, it seems like we are just getting rid of this information.

TAGS.yaml
andysan: Maybe a better solution would be to unify tags and labels and to discourage people from listing _all_ tags affected by a change? In most cases, there is a primary target that should be the tag or a generic one (e.g., just specify arch instead of a list of arch-FOO, arch-BAR, etc.). E.g., if the intention is to change an API in sim, don't list the tags that relate to changes as a consequence of that API change. That would avoid the long list of tags at the beginning of commit messages.

We should probably also put the PR ID in a standard GIT trailer tag (similar to signed-off-by) since it isn't particularly useful in a shortlog.

powerjg:
> I'm not sure I follow the purpose of this change. We presumably still want to have dedicated subsystem maintainers and we need to document that somewhere (ideally in the source repository to ensure that it is propagated in clones). Looking at the diff itself, it seems like we are just getting rid of this information.

In practice, over the past few years, we have had few subsystem maintainers step up and review a significant portion of the changes in the subsystem they are assigned. Also, there are many people listed in this file who no longer review changes.

Can you explain the logic behind why we would keep these names around when in practice the information has not been used?

andysan:
> > I'm not sure I follow the purpose of this change. We presumably still want to have dedicated subsystem maintainers and we need to document that somewhere (ideally in the source repository to ensure that it is propagated in clones). Looking at the diff itself, it seems like we are just getting rid of this information.
> 
> In practice, over the past few years, we have had few subsystem maintainers step up and review a significant portion of the changes in the subsystem they are assigned. Also, there are many people listed in this file who no longer review changes.
> 
> Can you explain the logic behind why we would keep these names around when in practice the information has not been used?

I would suggest dropping those names in that case. The whole point of listing subsystem maintainers is that we want people to develop expertise and take responsibility for an area of gem5. That also means that people who are no longer actively reviewing changes should be encouraged to step down as maintainers. We should encourage more of people to step up rather than just giving up. IMO, this change doesn't solve a problem, it just sweeps it under the carpet.

powerjg:
We currently have no process or meaning to subsystem maintainers. At one point, we wanted to have a requirement that a subsystem maintainer sign off on the change, but other than `arm`, in practice we do not have enough maintainers to do that. 

Until we build up more maintainers and build up more expertise in subsystems, I do not see the purpose of listing people's names in this file.

Feel free to say something like "I am against merging this change" or "change line x by doing y and z". Otherwise, it sounds like it should be merged.

andysan:
I'm not going to block this change, but I think it is wrong. I would much rather see us just make the lack of subsystem maintainers explicit instead. That is, I would suggest that we keep the notion of maintainers and just drop the inactive maintainers from the file. We practically have the concept of fallback maintainers already, so we should make that (more) explicit. 

Practically speaking, at least arch-arm should be well-maintained and I'm trying to keep up with all of the kvm, virtio, and core Python changes. 

Unfortunately, this suggestion would mean that most subsystems don't have a listed maintainer. That's not necessarily a big problem though since it would mean that we can assign one in the future (fingers crossed).

BobbyRBruce:
> I'm not going to block this change, but I think it is wrong. I would much rather see us just make the lack of subsystem maintainers explicit instead. That is, I would suggest that we keep the notion of maintainers and just drop the inactive maintainers from the file. We practically have the concept of fallback maintainers already, so we should make that (more) explicit.
> 
> Practically speaking, at least arch-arm should be well-maintained and I'm trying to keep up with all of the kvm, virtio, and core Python changes.
> 
> Unfortunately, this suggestion would mean that most subsystems don't have a listed maintainer. That's not necessarily a big problem though since it would mean that we can assign one in the future (fingers crossed).

I would like people with an interest on maintaining certain subsystems to continue to do so.  I think GitHub labels on PRs can be helpful in this. With them can quickly filter PRs based on labels and allow reviewers to focus on changes they have expertise in. For now though, we can do this with commit messages tags as well. 

The reason I want to drop the MAINTAINERS file is, in our experience, explicitly stating who the maintainers are for each subsystem hasn't worked well. We've had to drop any idea that a subsystem maintainer _must_ review a patch related to that subsystem before being merged because these maintainers often didn't have the bandwidth to meet that obligation. Many subsystems are without a maintainer, and there often cases where maintainers who are not assigned to a specific subsystem still have the expertise to review a change and we see no reason in blocking them from doing so.

Right now the tags serve as notification to reviewiers about what a change is about, not about who should be in charge of it. This change reflects that.

shingarov:
>I'm not going to block this change, but I think it is wrong.

I hold the same position as Andreas.  For me, the most problematic aspect of this PR is that it effectively codifies a change to the definition of what the word "maintainer" means, shifting it towards "someone who reviews PRs".  This is definitely one valid perspective, but not the _only_ one, and our old `MAINTAINERS.yaml` captures those subtle semantics in an indirect way.  This PR seems to imply — and *codify* — that those other perspectives are non-things.

One example of such information that `MAINTAINERS.yaml` contains, is simply stating that "subsystem _S_ is actively maintained, go talk to person _P_ for more details".  And in some situations, "actively" may mean something else than "lots of PRs keep being reviewed".  It could be, say, convincing the contributor to submit the PRs in the first place which is taking 99% of the actual work (like it was in the case of getting IBM and VT to upstream their PowerGem fork of gem5 (and still *is* the case with those ~250 PRs still stuck in the fork after we reviewed the first ~150 ones)).

powerjg:
@shingarov 

I think that you're exactly right that the goal of this change is to codify the fact that maintainers are "someone who reviews PRs." This is the current state of things. If, at some point in the future, the community changes and we have many or most subsystems actively maintained by a people or if we find that most maintainers only review a certain subset of changes, then we can change the procedure again.

This change isn't meant to change the way gem5 operates. Instead, it is codifying how the community currently works. I would argue that the idea behind the current MAINTAINERS.yaml file is misleading and confusing to both newcomers and current community members (i.e., many people listed are no longer active participants in the community).

I think you have good point that having a place to go to get information on who to ask or who to talk to would be helpful. I would suggest that this file is not the right place for that. For instance, here's how Python does it: https://devguide.python.org/core-developers/experts/index.html 

One major problem with having specific maintainers (or even code experts) for subsystems in gem5 is that people come and go from the community frequently. People change roles or, more often, students graduate. Keeping up with these comings and goings is an overhead that we would like to avoid.

giactra:
I share the same views of @shingarov and @andysan.

> The reason I want to drop the MAINTAINERS file is, in our experience, explicitly stating who the maintainers are for each subsystem hasn't worked well. We've had to drop any idea that a subsystem maintainer must review a patch related to that subsystem before being merged because these maintainers often didn't have the bandwidth to meet that obligation.
Many subsystems are without a maintainer, and there often cases where maintainers who are not assigned to a specific subsystem still have the expertise to review a change and we see no reason in blocking them from doing so.
 
If there is a general maintainer that has the expertise/knowledge on maintaining a specific subsystem, shouldn't that person just be added to the subsystem list? Especially if the subsystem doesn't have a maintainer... I thought the idea was to encourage maintainers to review specifc portions of code. How can I be encouraged on taking ownership and allocating portions of my time in reviewing if I am just a generic maintainer?

> Until we build up more maintainers and build up more expertise in subsystems, I do not see the purpose of listing people's names in this file.

I would argue that explicitly stating unmaintained subsystems is the best way to entice users into becoming maintainers.. If I see (e.g.) cpu-o3 is left without a maintainer and for my own project I have developed a sufficient understaing of that part of gem5, I would be encouraged on reviewing cpu-o3 patches with the purpose of maybe becoming a maintainer once I have established a good reputation with the community.

> One example of such information that MAINTAINERS.yaml contains, is simply stating that "subsystem S is actively maintained, go talk to person P for more details"

Completely agree, and I don't understand why removing that name would make life easier for users.

Also I want to make the case of transparency. I feel users deserves to know which portion of gem5 has a dedicated maintainer... It really helps on understanding the amount of trust they can place on the code and on the likelihood there is some active development.
In general I believe with this change we are making things worse for users/developers/maintainers without achieving anything tangible. I am an arch-arm maintainer but I have maintained patches in other subsystems like cpu,arch,sim and base, and I did that regardless of the MAINTAINERS file; in other words we already have a safety net that allows maintainers to use their judgement while reviewing code outside their area.
 
I would also like to note the real problem is not the lack of maintainers in easy generic parts of gem5 (util scripts etc) but in complex orphaned subsystems like cpu-o3. We need someone with computer architecture knowledge that steps up and owns it (easier said than done I acknowledge), not someone that maintains coding style fixes (which are also important but it is not currently our problem IMHO).

BobbyRBruce:
I'm closing this pull request as this change has proven unpopular and we will try to try to work with having specific maintainers for the time being.

PR #110, opened 2023-07-21
mem-ruby,configs: Add GLC Atomic Latency VIPER Parameter

DanKouch:
Added a GLC atomic latency parameter (glc-atomic-latency) used when enqueueing response messages regarding atomics directly performed in the TCC. This latency is added in addition to the L2 response latency (TCC_latency). This represents the latency of performing an atomic within the L2.

With this change, the TCC response queue will receive enqueues with varying latencies as GLC atomic responses will have this added GLC atomic latency while data responses will not. To accommodate this in light of the queue having strict FIFO ordering (which would be violated here), this change also adds an optional parameter bypassStrictFIFO to the SLICC enqueue function which allows overriding strict FIFO requirements for individual messages on a case-by-case basis. This parameter is only being used in the TCC's atomic response enqueue call.

Change-Id: Iabd52cbd2c0cc385c1fb3fe7bcd0cc64bdb40aac

### DIFF
--- a/configs/ruby/GPU_VIPER.py
+++ b/configs/ruby/GPU_VIPER.py
@@ -489,6 +489,9 @@ def define_options(parser):
         help="Size of the mandatory queue in the GPU scalar "
         "cache controller",
     )
+    parser.add_argument(
+        "--glc-atomic-latency", type=int, default=1, help="GLC Atomic Latency"
+    )
 
 
 def construct_dirs(options, system, ruby_system, network):
@@ -875,6 +878,7 @@ def construct_tccs(options, system, ruby_system, network):
         tcc_cntrl.create(options, ruby_system, system)
         tcc_cntrl.l2_request_latency = options.gpu_to_dir_latency
         tcc_cntrl.l2_response_latency = options.TCC_latency
+        tcc_cntrl.glc_atomic_latency = options.glc_atomic_latency
         tcc_cntrl_nodes.append(tcc_cntrl)
         tcc_cntrl.WB = options.WB_L2
         tcc_cntrl.number_of_TBEs = 2560 * options.num_compute_units
--- a/src/mem/ruby/network/MessageBuffer.cc
+++ b/src/mem/ruby/network/MessageBuffer.cc
@@ -62,7 +62,8 @@ MessageBuffer::MessageBuffer(const Params &p)
     m_max_dequeue_rate(p.max_dequeue_rate), m_dequeues_this_cy(0),
     m_time_last_time_size_checked(0),
     m_time_last_time_enqueue(0), m_time_last_time_pop(0),
-    m_last_arrival_time(0), m_strict_fifo(p.ordered),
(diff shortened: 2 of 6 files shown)


mattsinc (changes requested):
As far as I can tell, your commit message needs to be updated for two things:

1.  Follow the 80 character per line limit
2.  In the second paragraph, you should explain better why you added this parameter -- i.e., specifically because the GL2 now has data requests that take less time than atomic requests, and the ignoring of ordering is intended to only happen when they interact.

mattsinc:
> As far as I can tell, your commit message needs to be updated for two things:
> 
>     1. Follow the 80 character per line limit
> 
>     2. In the second paragraph, you should explain better why you added this parameter -- i.e., specifically because the GL2 now has data requests that take less time than atomic requests, and the ignoring of ordering is intended to only happen when they interact.

@DanKouch I see you are still pushing updates.  Just let us know when you want us to review.

shingarov:
To add to Matt's list:
3. As per the "Requirements for change descriptions" section in CONTRIBUTING.md, use imperative mood in the short description.  That is: "Add GLC Atomic Latency...", not "Added..."  This has been discussed many many times already.

DanKouch:
@mattsinc @shingarov Thanks! I've made those changes and marked this as ready to review.

mattsinc (commented):
This looks good to me now, thanks!

@abmerop should also approve before we check in though.

mattsinc (approved):
Seems I literally must approve to remove the prior "request for changes" ... so doing so here.  @DanKouch please do not commit until @abmerop can review though.

abmerop (commented):
The code looks good. The commit message line lengths are too long though.  I believe commit messages are still supposed to be <80 characters per line.

mattsinc:
> The code looks good. The commit message line lengths are too long though. I believe commit messages are still supposed to be <80 characters per line.

@abmerop I was confused about this too, but I think it's GitHub playing an optical illusion on us.  If you go here: https://github.com/gem5/gem5/pull/110/commits/7170ac8f3edc4685c9121a79fe3f20f9c17043f9, you can see the commit message lines are 80 characters or less.

abmerop (approved):
Ah, I see now.  LGTM.

DanKouch:
@mattsinc @abmerop It looks like GitHub automatically reformats the message when creating the PR description per this [article](https://github.blog/changelog/2022-09-07-better-suggested-pull-request-description-from-commit-message/).

mattsinc:
> @mattsinc @abmerop It looks like GitHub automatically reformats the message when creating the PR description per this [article](https://github.blog/changelog/2022-09-07-better-suggested-pull-request-description-from-commit-message/).

Thanks @DanKouch !  It looks like there are some commits on develop now that weren't there when you pushed -- in theory it looks like GitHub will let you automatically update though via "Update branch"

DanKouch:
@mattsinc Rebased to the head of develop.

PR #113, opened 2023-07-23
arch-x86: Move CPUID values to python

abmerop:
arch-x86: Move CPUID values to python

    CPUID values for X86 are currently hard-coded in the C++ source file.
    This makes it difficult to configure the bits if needed. Move these to
    python instead. This will provide a few benefits:

    1. We can enable features for certain configurations, for example AVX
    can be enabled when the KVM CPU is used, but otherwise should not be
    enabled as gem5 does not have full AVX support.
    2. We can more accurately communicate things like cache/TLB sizes based
    on the actual gem5 configuration. The CPUID values are can be used by
    some libraries, e.g., MPI, to query system topology.
    3. Enabling some bits breaks things in certain configurations and this
    can be prevented by configuring in python. For example, enabling AVX
    seems to currently be breaking SMP, meaning gem5 can only boot one CPU
    in that configuration.

### DIFF
--- a/configs/example/gpufs/system/system.py
+++ b/configs/example/gpufs/system/system.py
@@ -231,6 +231,42 @@ def makeGpuFSSystem(args):
         clock=args.ruby_clock, voltage_domain=system.voltage_domain
     )
 
+    # If we are using KVM cpu, enable AVX. AVX is used in some ROCm libraries
+    # such as rocBLAS which is used in higher level libraries like PyTorch.
+    use_avx = False
+    if ObjectList.is_kvm_cpu(TestCPUClass):
+        # AVX also requires CR4.osxsave to be 1. These must be set together
+        # of KVM will error out.
+        system.workload.enable_osxsave = 1
+        use_avx = True
+
+    # These values are taken from a real CPU and are further explained here:
+    # https://sandpile.org/x86/cpuid.htm#level_0000_000Dh
+    avx_extended_state = [
+        0x00000007,
+        0x00000340,
+        0x00000000,
+        0x00000340,
+        0x0000000F,
+        0x00000340,
+        0x00000000,
+        0x00000000,
+        0x00000100,
+        0x00000240,
+        0x00000000,
+        0x00000040,
+        0x00000000,
+        0x00000000,
+        0x00000000,
+        0x00000000,
+    ]
+
+    # This modifies the default value for ECX only (4th in this array).
+    # See: https://sandpile.org/x86/cpuid.htm#level_0000_0001h
+    # Enables AVX, OSXSAVE, XSAVE, POPCNT, SSE4.2, SSE4.1, CMPXCHG16B,
+    # and FMA.
+    avx_cpu_features = [0x00020F51, 0x00000805, 0xEFDBFBFF, 0x1C983209]
+
     for (i, cpu) in enumerate(system.cpu):
         # Break once we reach the shader "CPU"
         if i == args.num_cpus:
@@ -247,6 +283,9 @@ def makeGpuFSSystem(args):
 
         for j in range(len(system.cpu[i].isa)):
             system.cpu[i].isa[j].vendor_string = "AuthenticAMD"
+            if use_avx:
+                system.cpu[i].isa[j].ExtendedState = avx_extended_state
+                system.cpu[i].isa[j].FamilyModelStepping = avx_cpu_features
 
     if args.host_parallel:
         # To get the KVM CPUs to run on different host CPUs, specify a
--- a/src/arch/x86/X86ISA.py
+++ b/src/arch/x86/X86ISA.py
@@ -54,3 +54,59 @@ class X86ISA(BaseISA):
     vendor_string = Param.String(
         "HygonGenuine", "Vendor string for CPUID instruction"
     )
+    name_string = Param.String(
+        "Fake gem5 x86_64 CPU", "Processor name for CPUID instruction"
+    )
+
+    # For the functions that return numerical values we use a vector of ints.
+    # The order of the values is: EAX, EBX, EDX, ECX.
+    #
+    # If the CPU function can take an index, the index value is used as an
+    # offset into the vector and four numerical values are added for each
+    # possible index value. For example, if the function accepts 3 index
+    # values, there are 12 total ints in the vector param. In addition, the
+    # last values for functions which take an index must be all zeros. All
+    # zeros indicates to the KVM cpu / OS that there are no more index values
+    # to iterate over.
+    #
+    # A good resource for these values can be found here:
+    #     https://sandpile.org/x86/cpuid.htm
+    # 0000_0001h
+    FamilyModelStepping = VectorParam.UInt32(
+        [0x00020F51, 0x00000805, 0xEFDBFBFF, 0x00000209],
+        "type/family/model/stepping and feature flags",
+    )
+    # 0000_0004h
+    CacheParams = VectorParam.UInt32(
+        [0x00000000, 0x00000000, 0x00000000, 0x00000000],
+        "cache configuration descriptors",
+    )
+    # 0000_0007h
+    ExtendedFeatures = VectorParam.UInt32(
+        [0x00000000, 0x01800000, 0x00000000, 0x00000000], "feature flags"
+    )
+    # 8000_0001h
+    FamilyModelSteppingBrandFeatures = VectorParam.UInt32(
+        [0x00020F51, 0x00000405, 0xEBD3FBFF, 0x00020001],
+        "family/model/stepping and features flags",
+    )
+    # 8000_0005h
+    L1CacheAndTLB = VectorParam.UInt32(
+        [0xFF08FF08, 0xFF20FF20, 0x40020140, 0x40020140],
+        "L1 cache and L1 TLB configuration descriptors",
+    )
+    # 8000_0006h
+    L2L3CacheAndL2TLB = VectorParam.UInt32(
+        [0x00000000, 0x42004200, 0x00000000, 0x04008140],
+        "L2/L3 cache and L2 TLB configuration descriptors",
+    )
+    # 8000_0007h
+    APMInfo = VectorParam.UInt32(
+        [0x80000018, 0x68747541, 0x69746E65, 0x444D4163],
+        "processor feedback capabilities",
+    )
+    # 8000_0008h
+    LongModeAddressSize = VectorParam.UInt32(
+        [0x00003030, 0x00000000, 0x00000000, 0x00000000],
+        "miscellaneous information",
+    )
--- a/src/arch/x86/X86ISA.py
+++ b/src/arch/x86/X86ISA.py
@@ -85,6 +85,20 @@ class X86ISA(BaseISA):
     ExtendedFeatures = VectorParam.UInt32(
         [0x00000000, 0x01800000, 0x00000000, 0x00000000], "feature flags"
     )
+    # 0000_000Dh - This uses ECX index, so the last entry must be all zeros
+    ExtendedState = VectorParam.UInt32(
+        [
+            0x00000000,
+            0x00000000,
+            0x00000000,
+            0x00000000,
+            0x00000000,
+            0x00000000,
+            0x00000000,
+            0x00000000,
+        ],
+        "extended state enumeration",
+    )
     # 8000_0001h
     FamilyModelSteppingBrandFeatures = VectorParam.UInt32(
         [0x00020F51, 0x00000405, 0xEBD3FBFF, 0x00020001],
--- a/src/arch/x86/cpuid.cc
+++ b/src/arch/x86/cpuid.cc
@@ -146,6 +146,18 @@ X86CPUID::stringToRegister(const char *str)
 bool
 X86CPUID::hasSignificantIndex(uint32_t function)
 {
+    uint16_t family = bits(function, 31, 16);
+    uint16_t funcNum = bits(function, 15, 0);
+
+    if (family == 0x0000) {
+        switch (funcNum) {
+          case ExtendedState:
+            return true;
+          default:
+            return false;
+        }
+    }
+
     return false;
 }
 
--- a/src/arch/x86/cpuid.cc
+++ b/src/arch/x86/cpuid.cc
@@ -111,9 +111,19 @@ X86CPUID::doCpuid(ThreadContext * tc, uint32_t function, uint32_t index,
         return false;
     }
 
+    int cap_offset = 0;
+
+    // Ignore index values for functions that do not take index values.
+    if (hasSignificantIndex(function)) {
+        cap_offset = index * 4;
+    }
+
+    // Ensure we have the offset and 4 dwords after it.
+    assert(capabilities[function].size() >= (cap_offset + 4));
+
     auto &cap_vec = capabilities[function];
-    result = CpuidResult(cap_vec[0], cap_vec[1],
-                         cap_vec[2], cap_vec[3]);
+    result = CpuidResult(cap_vec[cap_offset + 0], cap_vec[cap_offset + 1],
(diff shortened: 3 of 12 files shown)


configs/example/gpufs/system/system.py
```diff
@@ -231,6 +231,22 @@ def makeGpuFSSystem(args):
...
+    # If we are using KVM cpu, enable AVX. AVX is used in some ROCm libraries
+    # such as rocBLAS which is used in higher level libraries like PyTorch.
+    use_avx = False
+    if ObjectList.is_kvm_cpu(TestCPUClass):
+        # AVX also requires CR4.osxsave to be 1. These must be set together
+        # of KVM will error out.
+        system.workload.enable_osxsave = 1
+        use_avx = True
+
+    avx_extended_state = [0x00000007, 0x00000340, 0x00000000, 0x00000340,
```
mattsinc: Is it possible to add a comment explaining what these numbers mean?

src/arch/x86/cpuid.cc
```diff
@@ -31,162 +31,146 @@
...
-        FamilyModelSteppingBrandFeatures,
-        NameString1,
-        NameString2,
-        NameString3,
-        L1CacheAndTLB,
-        L2L3CacheAndL2TLB,
-        APMInfo,
-        LongModeAddressSize,
-
-        /*
```
mattsinc, line 64: seems like these are still not implemented and should have been left?

src/arch/x86/cpuid.hh
```diff
@@ -38,28 +41,41 @@ class ThreadContext;
...
-        uint64_t rcx;
-        uint64_t rdx;
 
-        // These are not in alphebetical order on purpose. The order reflects
-        // how the CPUID orders the registers when it returns results.
-        CpuidResult(uint64_t _rax, uint64_t _rbx,
-                    uint64_t _rdx, uint64_t _rcx) :
-            rax(_rax), rbx(_rbx), rcx(_rcx), rdx(_rdx)
-        {}
+struct CpuidResult
```
mattsinc, line 75: seems you removed indentation here?

mattsinc (commented):
I don't feel 100% confident in some of these changes, but have at least a couple questions for you in the below.

abmerop:
Hmm, well I have to admit I think python black makes it look worse in this particular situation, but in an effort of keeping with the style guide I will re-upload.

configs/example/gpufs/system/system.py
abmerop: Yes I can do that-  these are related to how much storage the CPU tells the CPU has for xsave.

src/arch/x86/cpuid.cc
abmerop: The reason I decided to remove these is because there are dozens of missing functions so I don't think these ones are in any way special.  In a patch in this "rel change" (if we can still call it that), it sets an enum to a value which makes the compiler skip the other values. I am guessing the original author may have added these so that the final/last enum value returns the number of functions to the CPUID call correctly

src/arch/x86/cpuid.hh
abmerop: Yes that is true -  I made an "undintation" to this PR to be more in line with the namespace changes that occurred in gem5 v22 (around that time).

src/arch/x86/cpuid.cc
mattsinc: Ok ... I don't know this code well enough to say if this is a good idea or not, but I guess at least GitHub will document it here for future reference.

src/arch/x86/X86ISA.py
```diff
@@ -54,3 +54,52 @@ class X86ISA(BaseISA):
...
+    # offset into the vector and four numerical values are added for each
+    # possible index value. For example, if the function accepts 3 index
+    # values, there are 12 total ints in the vector param. In addition, the
+    # last values for functions which take an index must be all zeros. All
+    # zeros indicates to the KVM cpu / OS that there are no more index values
+    # to iterate over.
+    #
+    # A good resource for these values can be found here:
+    #     https://sandpile.org/x86/cpuid.htm
+    # 0000_0001h
```
powerjg, line 74: As Matt mentioned above, it would be good to add comments for what these default values mean.

I'm sure you did a lot of leg work to figure out all of these details. It would be great if we didn't have to do this again :)

src/arch/x86/cpuid.hh
```diff
@@ -38,28 +41,41 @@ class ThreadContext;
...
-    };
+    CpuidResult()
+    {}
+};
 
-    uint64_t stringToRegister(const char *str);
+class X86CPUID
+{
+  public:
+    X86CPUID(const X86ISAParams &p);
```
powerjg: I don't like the idea of passing the SimObject params to a non-SimObject. I know that it will make things more verbose, but I think it would be better to explicitly pass in the parameters that you need.

I'm worried about the future if/when anything about params changes. We're coupling `X86CPUID` to `BaseParams` in a way that is unnecessary.

This isn't a big problem, though.

src/arch/x86/isa.hh
```diff
@@ -93,6 +94,8 @@ class ISA : public BaseISA
     void setThreadContext(ThreadContext *_tc) override;
 
     std::string getVendorString() const;
+
+    X86CPUID *cpuid;
```
powerjg: Probably should be `unique_ptr`

powerjg:
> Hmm, well I have to admit I think python black makes it look worse in this particular situation, but in an effort of keeping with the style guide I will re-upload.

I think that it's worth having things that don't look perfect so we don't argue about style :)

andysan:
> > Hmm, well I have to admit I think python black makes it look worse in this particular situation, but in an effort of keeping with the style guide I will re-upload.
> 
> I think that it's worth having things that don't look perfect so we don't argue about style :)

There are cases where it might be legitimate to disable black. Black has a couple of [magic comments](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) (`# fmt: off/on`) that let you do this. I sometimes use this in my own code when formatting large tables where there is some structure to the data that I want to replicate in the source code. Another trick is to leave a hanging comma which tells black to format a dicts / lists / tuples / parameters with one entry per line.

src/arch/x86/cpuid.hh
abmerop: I originally tried to make it a SimObject but I could not figure out a way (if there is one) to create a default SimObject with all the default values. Therefore, every single python script which creates an X86 CPU would have to be changed, I think, which I wanted to avoid this way. I could potentially make the X86CPUID class in the X86ISA namespace/class.

src/arch/x86/cpuid.cc
abmerop: There are dozens of CPUID functions missing and I see no reason to define them here.  Since it's possible to just assign a number to the enum name, I suggest using that rather than having these placeholder functions that aren't actually implemented just to get the number to line up.

src/arch/x86/cpuid.hh
powerjg: The class not being a SimObject isn't a problem. The (small) issue I'm raising is that I don't like that a class that is not a SimObject is taking a SimObjectParams* as the parameter to its constructor.

I would  prefer something like
```
X86CPUID::X86CPUID(std::string vendor_string, std::string name_string)
    : vendorString(vendor_string), nameString(name_string)
{...}
```
and
```
cpuid = new X86CPUID(p.vendor_string, p.name_string);
```

This way `X86CPUID` is not coupled to `X86ISAParams` but only to `X86::ISA`

abmerop: But the class is only used in X86ISA.  Doesn't your approach have the reserve problem where X86ISA depends on the enum value in X86CPUID?

powerjg: This isn't a big deal, so you can feel free to ignore.

That's exactly right. Since X86CPUID is used only in X86ISA, it's OK for X86ISA to be coupled to X86CPUID. What we don't want to couple a class that's only used in one place (X86CPUID) to a class that is outside of the same "module" (i.e., X86ISAParams).

Again, not a big deal, but I think it would be better practice to not pass around the params object. Instead of passing around the params object, I would rather see passing the explicit variable/params from that object as parameters to the class.

src/arch/x86/isa.hh
powerjg: I would like to see this change before merging.

BobbyRBruce:
The pre-commit checks failed. You can fix this locally using:

```sh
./util/pre-commit-install.sh # Ensure the hooks are installed
pre-commit run --files <files modified in this patch>
git add <files fixed by pre-commit>
git commit -m "misc: Fix pre-commit formatting issues"
```

Then push your changes.

src/arch/x86/X86ISA.py
abmerop: These are just moved from the C++ and the values have not changed, so I did not really put in any work to understand them.  I put a link to a website above that explains the values.  It would be way too much to explain what all 1024 bits in this file do

abmerop:
Not sure what I need to do for the PR list to show that this was updated, but it was updated.

src/arch/x86/isa.hh
powerjg: Any reason not to make this a `unique_ptr`?

powerjg (commented):
One more little thing. Let me know if you disagree with the suggestion. IMO, it's better to use smart pointers when possible and not use `new`/`delete`, but let me know if I'm missing something.

Thanks for the changes so far!

powerjg:
> Not sure what I need to do for the PR list to show that this was updated, but it was updated.

I think you can click the "re-review" button in the upper right by my name under "reviewers" to notify me. 

image

powerjg:
(Apologies for closing, my keyboard did something crazy...)

src/arch/x86/isa.hh
abmerop: Missed this conversation somehow (kind find them hard to see in github's dark mode).  Done now.

powerjg (dismissed):
Thanks for all of the updates!

mattsinc (dismissed):

BobbyRBruce:
The tests appear to be failing to run the `test-x86-hello64-dynamic` tests. Since the contributors to this patch are currently sleeping i'll return the tests just to make sure this isn't a bug with our runners.

abmerop:
> The tests appear to be failing to run the `test-x86-hello64-dynamic` tests. Since the contributors to this patch are currently sleeping i'll return the tests just to make sure this isn't a bug with our runners.

Seems they are still failing.  I have a few questions about this:  How can I download the artifacts(?) that show what the error is?  I can see which tests failed, how can I run *only* those tests (using ./tests/main.py?).  I wasn't sure how to do this so I ran the whole thing overnight (./main run -vv) but it appears to have either hanged or takes longer than 12 hours.

Edit: More specifically I ran ./main.py run -j60 -t60 -vv.

abmerop:
> > The tests appear to be failing to run the `test-x86-hello64-dynamic` tests. Since the contributors to this patch are currently sleeping i'll return the tests just to make sure this isn't a bug with our runners.
> 
> Seems they are still failing. I have a few questions about this: How can I download the artifacts(?) that show what the error is? I can see which tests failed, how can I run _only_ those tests (using ./tests/main.py?). I wasn't sure how to do this so I ran the whole thing overnight (./main run -vv) but it appears to have either hanged or takes longer than 12 hours.
> 
> Edit: More specifically I ran ./main.py run -j60 -t60 -vv.

Was able to find the command with -vvv.  I ran the three failing tests locally and they all pass (the amd64 hello world apps).  This makes sense to me as hello world... doesn't use CPUID probably.

powerjg:
For next time, you can download the results from the test on the actions page:
image

Then, you can search for the failing test and find the simout/simerr, which also includes the command that was run.

In this case, simerr shows the problem:
```
src/base/remote_gdb.cc:418: warn: Sockets disabled, not accepting gdb connections
src/sim/mem_state.cc:443: info: Increasing stack size by one page.
/lib/x86_64-linux-gnu/libc.so.6: CPU ISA level is lower than required
```

It looks like the changes to CPUID is causing glibc on the host (docker container running gem5 in this case) to get confused as to the CPU level.

The exact command (from the simout file) is:
```
command line: /__w/gem5/gem5/build/ALL/gem5.opt -d /tmp/gem5outx1tcuelx -re --silent-redirect /__w/gem5/gem5/tests/gem5/configs/simple_binary_run.py x86-hello64-dynamic timing --resource-directory /__w/gem5/gem5/tests/gem5/resources x86
```

Of course, you can ignore the paths, etc. when trying to reproduce.

Finally, the docker container that we are using is `gcr.io/gem5-test/ubuntu-22.04_all-dependencies:latest` as found here: https://github.com/gem5/gem5/actions/runs/5686873518/workflow?pr=113#L78

Let me know if there's more we can do to help you diagnose this in the future :smile:

powerjg:
(Side note: I want to point out how much better the situation is on github for diagnosing when tests fail! We now have access/control over every aspect of the testing! Rather than having to re-run all of the tests because one little thing failed for an unknown reason (like when we were using kokoro), we can now see the *output* of the failing test and easily reproduce!)

abmerop:
> (Side note: I want to point out how much better the situation is on github for diagnosing when tests fail! We now have access/control over every aspect of the testing! Rather than having to re-run all of the tests because one little thing failed for an unknown reason (like when we were using kokoro), we can now see the _output_ of the failing test and easily reproduce!)

Thanks Jason.  I'm sure everything in here ... somewhere.  So far I am finding the github interface is much more complex than gerrit and not really able to find anything since i've never used github much.

abmerop:
> For next time, you can download the results from the test on the actions page: image
> 
> Then, you can search for the failing test and find the simout/simerr, which also includes the command that was run.
> 
> In this case, simerr shows the problem:
> 
> ```
> src/base/statistics.hh:279: warn: One of the stats is a legacy stat. Legacy stat is a stat that does not belong to any statistics::Group. Legacy stat is deprecated.
> src/base/remote_gdb.cc:418: warn: Sockets disabled, not accepting gdb connections
> src/sim/simulate.cc:194: info: Entering event queue @ 0.  Starting simulation...
> src/sim/mem_state.cc:443: info: Increasing stack size by one page.
> src/sim/syscall_emul.cc:74: warn: ignoring syscall set_robust_list(...)
> src/sim/syscall_emul.cc:74: warn: ignoring syscall rseq(...)
> /lib/x86_64-linux-gnu/libc.so.6: CPU ISA level is lower than required
> ```
> 
> It looks like the changes to CPUID is causing glibc on the host (docker container running gem5 in this case) to get confused as to the CPU level.
> 
> The exact command (from the simout file) is:
> 
> ```
> command line: /__w/gem5/gem5/build/ALL/gem5.opt -d /tmp/gem5outx1tcuelx -re --silent-redirect /__w/gem5/gem5/tests/gem5/configs/simple_binary_run.py x86-hello64-dynamic timing --resource-directory /__w/gem5/gem5/tests/gem5/resources x86
> ```
> 
> Of course, you can ignore the paths, etc. when trying to reproduce.
> 
> Finally, the docker container that we are using is `gcr.io/gem5-test/ubuntu-22.04_all-dependencies:latest` as found here: https://github.com/gem5/gem5/actions/runs/5686873518/workflow?pr=113#L78
> 
> Let me know if there's more we can do to help you diagnose this in the future 😄

I guess I will need to run on Ubuntu 22 to see this error then.

powerjg:
You can just use docker.

```
docker run --rm -w `pwd` -v `pwd`:`pwd` -u $UID:$GID gcr.io/gem5-test/ubuntu-22.04_all-dependencies:latest build/ALL/gem5.opt tests/gem5/configs/simple_binary_run.py x86-hello64-dynamic timing x86
```

I *think* that command works (I didn't test it)

abmerop:
Fixed the tester issue

powerjg (approved):

PR #114, opened 2023-07-24
arch-arm: Hook TLBIOS instructions to the TlbiShareable obj

giactra:
FEAT_TLBIOS has been introduced by a recent patch [1] which was however missing to include the outer shareable case in the Msr disambiguation switch. Which meant the TLBIOS instructions were decoded as normal MSR instructions, with no effect whatsoever on the TLBs

[1]: https://gem5-review.googlesource.com/c/public/gem5/+/70567

Change-Id: I41665a4634fbe0ee8cc30dbc5d88d63103082ae9


andysan (approved):
LGTM

powerjg:
I've assigned you to this change, Andreas. We are going to use "assignees" as the point maintainer who is responsible for merging the change. Thanks!

PR #116, opened 2023-07-24
misc: Remove 'Change-Id' check from CI Tests

BobbyRBruce:


BobbyRBruce:
As we were moving from a 'Squash and Merge' model to a 'Merge' model we
can no longer guarantee each commit has a Change-Id. Merge commits will
always be without a Change-Id and therefore cause the CI tests to break
on every Pull Request.

This change removes this requirement from the CI Tests. From this change
onward, having a Change-Id in each commit message is no longer a
requirement.

BobbyRBruce:
Actually, I'm wrong in thinking merge will break the change-id check. Though the check is pointless as merge commits will be without a change-id and thus break the change-id requirement. I'll hold off on this PR for a bit to see what the effects of the merge strategy are.

PR #117, opened 2023-07-24
ruby: Enable all protocols in a single gem5 build

powerjg:
This PR includes the changes needed to enable all Ruby protocols in a single gem5 build.

Right now, this is just a simple rebase from the old changes on gerrit. See https://gem5-review.googlesource.com/c/public/gem5/+/59112 for details.

Things that are to do on this PR before reviewing (i.e., please ignore for now).
- [x] Implement a `Protocol` class (see comments here https://gem5-review.googlesource.com/c/public/gem5/+/59113/10)
- [x] Make sure it builds (have not tested since rebase)
- [x] Update stdlib changes
- [x] Update CHI changes
- [x] Test backwards compatible with single protocol
- [x] Test GPU FS (split) protocol
- [x] Test multi-protocol build 

To do as of 11/1
- [ ] Dig into problem with ViperBoard if rebased on this PR
- [ ] Change the `ALL` build option to build all protocols as well and update the tests to use this build


powerjg:
🥳 This is now ready for review! There's still some more testing that needs to be done, and there are likely mistakes, but it's in a state that can be reviewed now.

Let me state one thing, for the record: I spent more time fighting with Kconfig than I did updating Ruby or anything else. This is not a good system. It is not flexible in the way we need, and it's not well documented.

Also, I've done my best to ensure backwards compatibility, but I could not get the build to work perfectly. See the RELEASE-NOTES.md for details.

I hope that this addresses most of the prior reviews.

For incoming reviews, let's try to concentrate on giving feedback if there are any bugs or corner cases I may have missed. Unfortunately, there is no chance that I'll have another 15+ hours to go back and redesign this change for the third time. This is the best design I could come up with, given the prior feedback and my limited coding cycles.

powerjg:
@BobbyRBruce the tests are failing because of caching. Is there a way to clear the build cache in the runners? Alternatively, if you (or anyone else) can figure out how to enable full backwards compatibility in the Kconfig, that would fix it. [This commit](https://github.com/gem5/gem5/pull/117/commits/7d4a7cd3eabba9be521a1f725d6cd2fa5c1a23f9) is the major problem.

BobbyRBruce:
> @BobbyRBruce the tests are failing because of caching. Is there a way to clear the build cache in the runners? Alternatively, if you (or anyone else) can figure out how to enable full backwards compatibility in the Kconfig, that would fix it. [This commit](https://github.com/gem5/gem5/pull/117/commits/7d4a7cd3eabba9be521a1f725d6cd2fa5c1a23f9) is the major problem.

I just wiped the build caches. For future reference you can delete any cache here: https://github.com/gem5/gem5/actions/caches

BobbyRBruce:
Failure to build VEGA_X86 is a bug with tnhe PR. It still happened when the cache was removed.

powerjg:
> Failure to build VEGA_X86 is a bug with tnhe PR. It still happened when the cache was removed.

I didn't say it was failing *only* because of caching ;)

configs/ruby/MULTIPLE.py
```diff
@@ -0,0 +1,89 @@
+# Copyright (c) Jason Lowe-Power
```
BobbyRBruce: I'm not sure if you (a UC employee are allowed to claim this copyright 😀.

Regardless, this needs a year (I think... i don't claim to understand anything about copyright law).

BobbyRBruce (commented):
I have looked over this it broadly looks fine to me, though I think it's too complicated for me to approve just through code review review. Given how big this change is I would prefer if, in addition to the CI tests, the daily and weekly tests were run as well prior to merging.

configs/ruby/MULTIPLE.py
powerjg: I'll add years. As a professor, I (can) own copyright. UC is weird. See https://copyright.universityofcalifornia.edu/ownership/works-created-at-uc.html

powerjg:
> the daily and weekly tests were run as well prior to merging.

This is a good idea. @erin-le can you do this?

powerjg:
@abmerop not sure if you were notified of this change since I marked it as "ready for review"
cc: @giactra

src/mem/ruby/Kconfig
```diff
@@ -28,15 +28,28 @@ menu "Ruby"
         bool "Enable"
 
     if RUBY
+        menu "Ruby protocols"
+
+        config USE_MULTIPLE_PROTOCOLS
+            bool 'Use multiple protocols'
+            default y
```
abmerop, line 35: Is this overridden by build_opts? I don't quite understand the interaction between build_opts / Kconfig.

If not, do we want to make default n until multiple protocols is better tested?

src/mem/ruby/Kconfig
```diff
@@ -28,15 +28,28 @@ menu "Ruby"
...
+                Otherwise, this is ignored.
+                If this is set to anything when build with MULTIPLE,
+                it will be ignored unless using Ruby.py, in which case
+                the protocol specified here will appear as the only protocol.
+                Note: This is case sensitive.
 
-        config NEED_PARTIAL_FUNC_READS
-            bool
+        rsource "protocol/Kconfig"
+        rsource "../../learning_gem5/part3/Kconfig"
```
abmerop, line 50: Was this intended to be moved from src/Kconfig to here?

src/mem/ruby/system/RubySystem.cc
```diff
@@ -493,9 +504,17 @@ RubySystem::resetStats()
     ClockedObject::resetStats();
 }
 
-#ifndef PARTIAL_FUNC_READS
```
abmerop, line 496: Nice, this was always a source of confusion which on is being used at runtime. Should be much more clear now!

src/mem/slicc/symbols/StateMachine.py
```diff
@@ -302,38 +304,75 @@ class $py_ident(RubyController):
                 )
 
         code.dedent()
-        code.write(path, f"{py_ident}.py")
+
+        # Needed for backwards compatibility while there is only one protocol
+        # compiled. When moving to multiple protocols in the gem5 binary, this
+        # will need to change.
```
abmerop: Hmm, is this comment still correct?  Does it imply that currently the PR doesn't currently support multiple protocols?

src/mem/ruby/protocol/SConscript
```diff
@@ -34,87 +34,141 @@ from SCons.Scanner import Classic
...
-protocol_dir = None
-for path in env['PROTOCOL_DIRS']:
-    if os.path.exists(path.File("%s.slicc" % protocol).abspath):
-        protocol_dir = Dir(path)
-        break
-
-if not protocol_dir:
-    raise ValueError("Could not find {}.slicc in PROTOCOL_DIRS".format(
-        protocol))
+def slicc_emitter(target, source, env):
```
abmerop, line 72: I didn't go through the next 80 lines *super* carefully since there is a lot I don't understand here, but if it has been tested with/without multiple protocols that's good enough for me to prove it works.

abmerop (commented):
This looks good to me!  I like the approach, how the files are organized, namespacing, etc. and it looks like this required a lot of knowledge of SLICC, Kconfig, SCons, python, *and* C++. So nothing major from me, but a few questions/comments.

src/mem/ruby/Kconfig
powerjg: I'm sure there's a better way to do this. However, I spent literally half of my development time bringing this up to date on Kconfig changes, and I couldn't find a better way. Here's what I know.

- The files in `build_opts` provides defaults for the Kconfig config file.
- `USE_MULTIPLE_PROTOCOLS` is actually ignored completely in the code.
- I only use this variable to set the `PROTOCOL` variable to "MULTIPLE" by default.
- In the `build_opts`, files for backwards compatibility, each file specifies the name of the protocol in `PROTOCOL` which is the only thing that really matters.

And finally, the last thing I know is that I *do not* like Kconfig

src/mem/ruby/Kconfig
powerjg: Yes. This is another thing that caused me headaches. When moving the a "choice" to a "config" you have to specify all of the options in a single place. So, I had to move this here. I don't like it, but I also just don't like Kconfig.

src/mem/slicc/symbols/StateMachine.py
powerjg: I'll update this comment. The code is correct, but the comment is out of date.

powerjg: For rebasing reasons, this comment isn't going to be updated in the right commit, but the new text will be right after the merge.

src/mem/ruby/protocol/SConscript
powerjg: Yeah, it's been tested, though it can always be tested more!

One thing to note here is that we have to run slicc on all protocols *twice*. The first time (the emitter code code here, IIRC) we just figure out what files we will be making (which, annoyingly requires an entire slicc parsing). Then, in the action code below we actually make the files. Unfortunately, this causes gem5's build to slow down significantly in the early stages (and it's not parallelizable at all).

powerjg:
@abmerop BTW, your code to remove the static variables really helped make the rebase easier! Thanks for your work on #1349!

abmerop:
I was able to test this PR with the current GPUFS scripts and they are working for square (one of the todos).

However, I updated the viper board after applying this PR and ran into a issue with RubyNetwork. Just wondering if you have seen this? `src/mem/ruby/network/simple/routing/WeightBased.cc:127: panic: assert(msg_dsts.count() == 0) failed`

erin-le:
I tested this using test_memory_traffic_gen.py and simple_traffic_run.py from the TestLib and had to make some modifications to get the tests to pass.

TODO: Still need to change the ALL build option to use all protocols and change the TestLib tests to use the new ALL build option.

powerjg:
> However, I updated the viper board after applying this PR and ran into a issue with RubyNetwork. Just wondering if you have seen this? `src/mem/ruby/network/simple/routing/WeightBased.cc:127: panic: assert(msg_dsts.count() == 0) failed`

Hmm, no I haven't seen that. But I also haven't tried to rebase the ViperBoard PR on top of this. I wonder if it's related to the disjoint network somehow.

abmerop:
> > However, I updated the viper board after applying this PR and ran into a issue with RubyNetwork. Just wondering if you have seen this? `src/mem/ruby/network/simple/routing/WeightBased.cc:127: panic: assert(msg_dsts.count() == 0) failed`
> 
> Hmm, no I haven't seen that. But I also haven't tried to rebase the ViperBoard PR on top of this. I wonder if it's related to the disjoint network somehow.

ViperBoard is not using using multiple networks, or at least it shouldn't be. Are you planning to try and rebase it?  If not I can try and take a look this weekend

powerjg:
No, I doubt I'll have time to rebase the ViperBoard on this change. I'll be traveling at MICRO through Wednesday and then out of the office Thurs-Friday.

I'll do my best to get the DArchR changes to the ViperBoard ready to go by Monday (depending on when I finish working on my talk and if I have extra time at workshops over the weekend).

abmerop:
> No, I doubt I'll have time to rebase the ViperBoard on this change. I'll be traveling at MICRO through Wednesday and then out of the office Thurs-Friday.
> 
> I'll do my best to get the DArchR changes to the ViperBoard ready to go by Monday (depending on when I finish working on my talk and if I have extra time at workshops over the weekend).

Sounds good. On Monday I can apply additional changes to the board including rebasing if this PR goes first and the UCD ViperBoard changes get push on top of my PR. That way I think both PRs can get in to v24.1.

I have found the issue I was describing above and it is specific to VIPER and not a problem with this PR. You can probably spot the problem here: https://github.com/abmerop/gem5/blob/viper-board/src/python/gem5/prebuilt/viper/viper_network.py#L94-L95

erin-le:
I pushed the changes to add all Ruby protocols to the ALL build and update the tests to use this build where possible. I wasn't able to get the `fs/linux/arm` Testlib tests to work with any of the existing gem5 builds, so I added an ARM_X86 build opt.

configs/learning_gem5/part3/msi_caches.py
```diff
@@ -46,7 +46,7 @@
 
 class MyCacheSystem(RubySystem):
     def __init__(self):
-        if buildEnv["PROTOCOL"] != "MSI":
+        if not "RUBY_PROTOCOL_MSI" in buildEnv:
```
BobbyRBruce, line 49: so it's `in buildEnv` and not `in buildEnv["PROTOCOL"]`? It's just searching for the key to know if the protocol is there? If so, what's the value? I.e., what do I get if I go with `print(buildEnv["RUBY_PROTOCOL_MSI"])`?

BobbyRBruce (dismissed):
I want to merge this PR soon. It seems to have had ample time for feedback and it needs merged to unblock other work (inc. the v24.1 release). If you have any protests on this please speak up ASAP!


I can't see any issues here. I think this is a great contribution.

abmerop (dismissed):
My comments were addressed and it's rebased to looks good to me.

FYI, the ViperBoard issue in the PR description has also been resolved, but is still marked as a TODO.

RELEASE-NOTES.md
```diff
@@ -7,6 +7,19 @@
...
+  * **USER FACING CHANGE**: The Ruby protocols in Kconfig have changed names (they are now the same case as the SLICC file names), and in addition,  So, after this commit, your build configurations need to be updated. You can do so by running `scons menuconfig <build dir>` and selecting the right ruby options. Alternatively, if you're using a `build_opts` file, you can run `scons defconfig build/<ISA> build_opts/<ISA>` which should update your config correctly.
+  * **USER FACING CHANGE**: The the "build_opts/ALL" build spec has been updated to include all Ruby protocols . As such, gem5 compilations of the "ALL" compilation target will include all gem5 Ruby protocols (previously just MESI_Two_Level).
+  * A "build_opts/NULL_ALL_RUBY" build spec has been added to include all Ruby protocols for a "NULL ISA" build . This is useful for testing Ruby protocols without the overhead of a full ISA and is used in gem5's traffic generator tests.
+  * A "build_opts/ARM_X96" build spec has been added due to a unique restriction in the "tests/gem5/fs/linux/arm" tests which requires a compilation of gem5 with both ARM and X86 and solely the MESI_Two_Level protocol.
```
powerjg, line 22: Typo X96 -> X86

powerjg (commented):
Thanks for the help pushing this over the line! Some of the tests randomly failed, so I re-ran the failed tests. Hopefully we'll be good to go soon.

Maybe we can fix my comment when we update the release notes later in the cycle to prevent another round of CI tests?

RELEASE-NOTES.md
BobbyRBruce: I'll push a separate change after merging this :P

BobbyRBruce:
> Thanks for the help pushing this over the line! Some of the tests randomly failed, so I re-ran the failed tests. Hopefully we'll be good to go soon.
> 
> Maybe we can fix my comment when we update the release notes later in the cycle to prevent another round of CI tests?

This was just due to too many runners trying to download too much stuff at the same time... it's a problem with our runners. We can have lots running at once but if they all start at roughly the same time they tend to all need to download lots of resources at the same time. The network grinds to a halt and the downloads start to timeout.

Anyway, I'm almost certain these tests will pass.

powerjg:
🥳 Our oldest PR gets committed!!!

PR #119, opened 2023-07-25
mem: Make functional request a response when satisfied by queue

atrib:
In the memory controller, MemCtrl::MemoryPort::recvFunctional, when the functional request is satisfied by the ctrl-response queue, correctly make the packet a response.

This change mirrors AbstractMemory::functionalAccess, which uses Packet::makeResponse() after satisfying the request.

Note:
`bool trySatisfyFunctional(..)` functions return `true` or `false` based on whether the request was satisfied.
`void recvFunctional(..)` functions modify the packet to indicate successful request satisfaction.

### DIFF
--- a/src/mem/mem_ctrl.cc
+++ b/src/mem/mem_ctrl.cc
@@ -1496,6 +1496,11 @@ MemCtrl::MemoryPort::recvFunctional(PacketPtr pkt)
         // calls recvAtomic() and throws away the latency; we can save a
         // little here by just not calculating the latency.
         ctrl.recvFunctional(pkt);
+    } else {
+        // The packet's request is satisfied by the queue, but queue
+        // does not call makeResponse.
+        // Here, change the packet to the corresponding response
+        pkt->makeResponse();
     }
 
     pkt->popLabel();


src/mem/mem_ctrl.cc
```diff
@@ -1496,7 +1496,8 @@ MemCtrl::MemoryPort::recvFunctional(PacketPtr pkt)
         // calls recvAtomic() and throws away the latency; we can save a
         // little here by just not calculating the latency.
         ctrl.recvFunctional(pkt);
-    }
+    } else 
+        pkt->makeResponse();
```
powerjg:
```suggestion
    } else {
        // If the packet is found in the queue, the queue does not makeResponse
        pkt->makeResponse();
    }
```

powerjg (changes requested):
One small style thing and a suggested comment to add. Otherwise, this change looks good to me.

src/mem/mem_ctrl.cc
atrib: Small change to comment text for clarity.

```suggestion
    } else {
        // The packet's request is satisfied by the queue, but queue does not call makeResponse
        // Here, change the packet to the corresponding response
        pkt->makeResponse();
    }
```

powerjg:
As far as I can tell, nothing changed in the PR. I'm not sure if you intended to make updates or not :). Let me know when you're "done" and I'll start the CI tests.

atrib:
I only changed the comment text, so no new CI tests should be necessary. 
I am "done", as in I do not have any more changes to add.

First time submitting a pull request, so apologies if I did something incorrectly.

powerjg (dismissed):
Thanks!

atrib:
Sorry, this is now a mess. To make our lives easier, I will just create a new, clean PR

PR #120, opened 2023-07-25
mem-ruby,configs: Add GPU GLC Atomic Resource Constraints

DanKouch:
Added a resource constraint, AtomicALUOperation, to GLC atomics performed in the TCC.

The resource constraint uses a new class, ALUFreeList array. The class assumes the following:
  - There are a fixed number of atomic ALU pipelines
  - While a new cache line can be processed in each pipeline each cycle, if a cache line is currently going through a pipeline, it can't be processed again until it's finished

Two configuration parameters have been used to tune this behavior:
  - tcc-num-atomic-alus corresponds to the number of atomic ALU pipelines
  - atomic-alu-latency corresponds to the latency of atomic ALU pipelines

Change-Id: I25bdde7dafc3877590bb6536efdf57b8c540a939

### DIFF
--- a/configs/ruby/GPU_VIPER.py
+++ b/configs/ruby/GPU_VIPER.py
@@ -275,6 +275,8 @@ class TCC(RubyCache):
 
     def create(self, options):
         self.assoc = options.tcc_assoc
+        self.atomicLatency = options.atomic_alu_latency
+        self.atomicALUs = options.tcc_num_atomic_alus
         if hasattr(options, "bw_scalor") and options.bw_scalor > 0:
             s = options.num_compute_units
             tcc_size = s * 128
@@ -496,6 +498,15 @@ def define_options(parser):
     )
     parser.add_argument(
         "--glc-atomic-latency", type=int, default=1, help="GLC Atomic Latency"
+    )
+    parser.add_argument(
+        "--atomic-alu-latency", type=int, default=0, help="Atomic ALU Latency"
+    )
+    parser.add_argument(
+        "--tcc-num-atomic-alus",
+        type=int,
+        default=64,
+        help="Number of atomic ALUs in the TCC",
     )
     parser.add_argument(
         "--tcp-num-banks",
--- a/src/mem/ruby/structures/ALUFreeListArray.cc
+++ b/src/mem/ruby/structures/ALUFreeListArray.cc
@@ -1,0 +1,106 @@
+/*
+ * Copyright (c) 2023 The University of Wisconsin
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "mem/ruby/structures/ALUFreeListArray.hh"
+
+#include "base/intmath.hh"
+#include "mem/ruby/system/RubySystem.hh"
+#include "sim/cur_tick.hh"
+
+namespace gem5
+{
+
+namespace ruby
+{
+
+/*
+*
+* Models num_ALUs pipelined atomic ALUs with a depth of access_latency ticks.
+* Rather than reserving ALUs, this class assumes multiple requests can go
+* through an ALU at the same time. As such, up to numALU new requests can
+* go through at once, with the caveat that a line already being processed
+* in an ALU can't start processing again until the previous request has exited
+* the pipeline.
+*
+* ALUs aren't mapped directly to cache lines. Rather, ALUs are treated as
+* a free list.
+*
+* Behavior:
+*   Requests will go through unless one/both of the following are met:
+*       - There have been more than [numALUs] requests in the current cycle
+*       - The same line has been accessed in the past accessLatency ticks
+*/
+
+ALUFreeListArray::ALUFreeListArray(unsigned int num_ALUs, Tick access_latency)
+{
+    this->numALUs = num_ALUs;
+    this->accessLatency = access_latency;
+}
+
+bool ALUFreeListArray::tryAccess(Addr addr)
+{
+    uint32_t accesses_this_tick = 0;
+
+    // Remove requests from the tail of the queue that occured more than
+    // accessLatency ticks ago
+    Tick oldestValidRecordStart = curTick() - this->accessLatency;
+
+    while (accessQueue.size() > 0 &&
+         (accessQueue.back().startTick < oldestValidRecordStart)) {
+        accessQueue.pop_back();
+    }
+
+    for (AccessRecord& record : accessQueue) {
+        // Block access if we would be using more ALUs than we have in a
+        // single tick
+        if (record.startTick == curTick() &&
+            (++accesses_this_tick > numALUs)) {
+            return false;
+        }
+
+        // Block access if the line is already being used
+        if (record.lineAddr == makeLineAddress(addr)) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+void ALUFreeListArray::reserve(Addr addr)
+{
+    // Only called after tryAccess, so we know queue is up to date and that
+    // the access is valid
+
+    // Add record to queue
+    accessQueue.push_front(AccessRecord(makeLineAddress(addr), curTick()));
+}
+
+} // namespace ruby
+} // namespace gem5
--- a/src/mem/ruby/structures/ALUFreeListArray.hh
+++ b/src/mem/ruby/structures/ALUFreeListArray.hh
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2023 The University of Wisconsin
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __MEM_RUBY_STRUCTURES_ALUFREELISTARRAY_HH__
+#define __MEM_RUBY_STRUCTURES_ALUFREELISTARRAY_HH__
+
+#include <deque>
+
+#include "mem/ruby/common/TypeDefines.hh"
+#include "sim/cur_tick.hh"
+
+namespace gem5
+{
+
+namespace ruby
+{
+
+class ALUFreeListArray
+{
+  private:
+    unsigned int numALUs;
+    Tick accessLatency;
+
+    class AccessRecord
+    {
+      public:
+        AccessRecord(Addr line_addr, Tick start_tick) {
+          this->lineAddr = line_addr;
+          this->startTick = start_tick;
+        }
+
+        Addr lineAddr;
+        Tick startTick;
+    };
+
+    // Queue of accesses from past accessLatency cycles
+    std::deque<AccessRecord> accessQueue;
+
+  public:
+    ALUFreeListArray(unsigned int num_ALUs, Tick access_latency);
+
+    bool tryAccess(Addr addr);
+
+    void reserve(Addr addr);
+
+    Tick getLatency() const { return accessLatency; }
+};
+
+} // namespace ruby
+} // namespace gem5
+
+#endif
--- a/src/mem/ruby/structures/CacheMemory.cc
+++ b/src/mem/ruby/structures/CacheMemory.cc
@@ -73,6 +73,8 @@ CacheMemory::CacheMemory(const Params &p)
               p.start_index_bit, p.ruby_system),
     tagArray(p.tagArrayBanks, p.tagAccessLatency,
              p.start_index_bit, p.ruby_system),
+    atomicALUArray(p.atomicALUs, p.atomicLatency *
+             p.ruby_system->clockPeriod()),
     cacheMemoryStats(this)
 {
     m_cache_size = p.size;
@@ -529,6 +531,8 @@ CacheMemoryStats::CacheMemoryStats(statistics::Group *parent)
       ADD_STAT(numTagArrayWrites, "Number of tag array writes"),
       ADD_STAT(numTagArrayStalls, "Number of stalls caused by tag array"),
       ADD_STAT(numDataArrayStalls, "Number of stalls caused by data array"),
+      ADD_STAT(numAtomicALUOperations, "Number of atomic ALU operations"),
+      ADD_STAT(numAtomicALUArrayStalls, "Number of stalls caused by atomic ALU array"),
       ADD_STAT(htmTransCommitReadSet, "Read set size of a committed "
                                       "transaction"),
       ADD_STAT(htmTransCommitWriteSet, "Write set size of a committed "
@@ -562,6 +566,12 @@ CacheMemoryStats::CacheMemoryStats(statistics::Group *parent)
         .flags(statistics::nozero);
 
     numDataArrayStalls
+        .flags(statistics::nozero);
+
+    numAtomicALUOperations
+        .flags(statistics::nozero);
+
+    numAtomicALUArrayStalls
         .flags(statistics::nozero);
 
     htmTransCommitReadSet
@@ -633,6 +643,11 @@ CacheMemory::recordRequestType(CacheRequestType requestType, Addr addr)
             tagArray.reserve(addressToCacheSet(addr));
         cacheMemoryStats.numTagArrayWrites++;
         return;
+    case CacheRequestType_AtomicALUOperation:
+        if (m_resource_stalls)
+            atomicALUArray.reserve(addr);
+        cacheMemoryStats.numAtomicALUOperations++;
+        return;
     default:
         warn("CacheMemory access_type not found: %s",
              CacheRequestType_to_string(requestType));
@@ -662,6 +677,15 @@ CacheMemory::checkResourceAvailable(CacheResourceType res, Addr addr)
                     "Data array stall on addr %#x in set %d\n",
                     addr, addressToCacheSet(addr));
             cacheMemoryStats.numDataArrayStalls++;
+            return false;
+        }
+    } else if (res == CacheResourceType_AtomicALUArray) {
+        if (atomicALUArray.tryAccess(addr)) return true;
+        else {
+            DPRINTF(RubyResourceStalls,
+                    "Atomic ALU array stall on addr %#x in line address %#x\n",
+                    addr, makeLineAddress(addr));
+            cacheMemoryStats.numAtomicALUArrayStalls++;
             return false;
         }
     } else {
--- a/src/mem/ruby/structures/RubyCache.py
+++ b/src/mem/ruby/structures/RubyCache.py
@@ -44,6 +44,11 @@ class RubyCache(SimObject):
         "0B", "block size in bytes. 0 means default RubyBlockSize"
     )
 
+    # Atomic parameters only applicable to GPU atomics
+    # Zero atomic latency corresponds to instantanous atomic ALU operations
+    atomicLatency = Param.Cycles(0, "Cycles for an atomic ALU operation")
+    atomicALUs = Param.Int(64, "Number of atomic ALUs")
+
     dataArrayBanks = Param.Int(1, "Number of banks for the data array")
     tagArrayBanks = Param.Int(1, "Number of banks for the tag array")
     dataAccessLatency = Param.Cycles(1, "cycles for a data array access")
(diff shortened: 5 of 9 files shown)


mattsinc (dismissed):
This looks good to me, but @abmerop should check too before we checkin.

powerjg:
I think it would be a good idea for @tiagormk (or he can point to someone else if I've tagged the wrong person). I know that there are others working on "far atomic" implementations in Ruby. It would be good if there was a just one implementation in the `ruby/structures` directory instead of two different implementations that both have to be maintained.

Since this implementation is "first", the question is whether CHI's far atomics can use this implementation (or use this implementation with minor changes). If the two implementations are very different, we may need to have a longer discussion.

Let's hold off on merging until we hear from the CHI folks.

mattsinc:
> I think it would be a good idea for @tiagormk (or he can point to someone else if I've tagged the wrong person). I know that there are others working on "far atomic" implementations in Ruby. It would be good if there was a just one implementation in the `ruby/structures` directory instead of two different implementations that both have to be maintained.
> 
> Since this implementation is "first", the question is whether CHI's far atomics can use this implementation (or use this implementation with minor changes). If the two implementations are very different, we may need to have a longer discussion.
> 
> Let's hold off on merging until we hear from the CHI folks.

We met with those folks at your request, and the sense I got from the meeting was that the far atomics implementation was too different for there to be a single implementation that works for both -- basically their implementation would not easily work with vectors of requests like this one has to.  So we went ahead with pushing this one.  @tiagormk can of course chime in to the contrary though.

powerjg:
Ok, that sounds good. I'd like to take a pass through the new Ruby code before merging (assuming I can find the time this week).

src/mem/ruby/structures/ALUFreeListArray.cc
```diff
@@ -0,0 +1,73 @@
+#include "mem/ruby/structures/ALUFreeListArray.hh"
```
powerjg, line 30: Missing copyright

powerjg (changes requested):
With a quick glance, I notice that some files are missing copyright. I'll take a more detailed pass later this week.

tiagormk:
> > I think it would be a good idea for @tiagormk (or he can point to someone else if I've tagged the wrong person). I know that there are others working on "far atomic" implementations in Ruby. It would be good if there was a just one implementation in the `ruby/structures` directory instead of two different implementations that both have to be maintained.
> > Since this implementation is "first", the question is whether CHI's far atomics can use this implementation (or use this implementation with minor changes). If the two implementations are very different, we may need to have a longer discussion.
> > Let's hold off on merging until we hear from the CHI folks.
> 
> We met with those folks at your request, and the sense I got from the meeting was that the far atomics implementation was too different for there to be a single implementation that works for both -- basically their implementation would not easily work with vectors of requests like this one has to. So we went ahead with pushing this one. @tiagormk can of course chime in the to contrary though.

My understanding is this change only adds resource modeling to the existing atomics in the viper protocol, so I don't see any blocker in merging this.

In parallel we can keep discussing potential for modeling multi-atomic requests in CHI so support the GPU case. Unless @mattsinc and @DanKouch have already decide to go ahead and add far atomics to viper.

mattsinc:
> > > I think it would be a good idea for @tiagormk (or he can point to someone else if I've tagged the wrong person). I know that there are others working on "far atomic" implementations in Ruby. It would be good if there was a just one implementation in the `ruby/structures` directory instead of two different implementations that both have to be maintained.
> > > Since this implementation is "first", the question is whether CHI's far atomics can use this implementation (or use this implementation with minor changes). If the two implementations are very different, we may need to have a longer discussion.
> > > Let's hold off on merging until we hear from the CHI folks.
> > 
> > 
> > We met with those folks at your request, and the sense I got from the meeting was that the far atomics implementation was too different for there to be a single implementation that works for both -- basically their implementation would not easily work with vectors of requests like this one has to. So we went ahead with pushing this one. @tiagormk can of course chime in the to contrary though.
> 
> My understanding is this change only adds resource modeling to the existing atomics in the viper protocol, so I don't see any blocker in merging this.
> 
> In parallel we can keep discussing potential for modeling multi-atomic requests in CHI so support the GPU case. Unless @mattsinc and @DanKouch have already decide to go ahead and add far atomics to viper.

We did not add far atomics directly into VIPER.  However the support for our GLC atomics and your far atomics are similar (ignoring the vector part).

configs/ruby/GPU_VIPER.py
```diff
@@ -492,6 +494,15 @@ def define_options(parser):
     parser.add_argument(
         "--glc-atomic-latency", type=int, default=1, help="GLC Atomic Latency"
     )
+    parser.add_argument(
+        "--atomic-alu-latency", type=int, default=0, help="Atomic ALU Latency"
+    )
+    parser.add_argument(
+        "--tcc-num-atomic-alus",
+        type=int,
+        default=64,
```
powerjg, line 508: Why is this default so different from the one in RubyCache.py?

src/mem/ruby/structures/RubyCache.py
```diff
@@ -44,6 +44,8 @@ class RubyCache(SimObject):
         "0B", "block size in bytes. 0 means default RubyBlockSize"
     )
 
+    atomicLatency = Param.Cycles(1, "Cycles for an atomic operation")
```
powerjg: It would be helpful to add a comment here that these two parameters only apply to GPU atomics.

configs/ruby/GPU_VIPER.py
```diff
@@ -492,6 +494,15 @@ def define_options(parser):
     parser.add_argument(
         "--glc-atomic-latency", type=int, default=1, help="GLC Atomic Latency"
     )
+    parser.add_argument(
+        "--atomic-alu-latency", type=int, default=0, help="Atomic ALU Latency"
```
powerjg, line 503: Why is this default different from the default in RubyCache.py?

configs/ruby/GPU_VIPER.py
```diff
@@ -273,6 +273,8 @@ class TCC(RubyCache):
 
     def create(self, options):
         self.assoc = options.tcc_assoc
+        self.atomicLatency = options.atomic_alu_latency
+        self.atomicALUs = options.tcc_num_atomic_alus
```
powerjg, line 279: Are we guaranteed that if these options are provided that each TCC `resourceStalls` will be true?

I.e., would it be possible for the user to specify `--tcc_num_atomic_alus=8` on the command line and then it be ignored by the simulation?

src/mem/ruby/structures/ALUFreeListArray.hh
```diff
@@ -0,0 +1,51 @@
...
+        Tick startTick;
+    };
+
+    // Queue of accesses from past accessLatency cycles
+    std::deque<AccessRecord> accessQueue;
+
+  public:
+    ALUFreeListArray(unsigned int alus, Cycles accessLatency, RubySystem *rs);
+
+    bool tryAccess(int64_t idx);
```
powerjg: What is `idx` in this case? Are the ALUs used for atomic tagged based on the set index?

I think that you don't want to have any parameter here and that you want the ALUs to be, in some sense, fully associative. I could be wrong here, though.

Alternatively, if you want to model that there is 1 ALU per cache bank, you should probably pass the full address in and have the notion of bank mapping in this object.

src/mem/ruby/structures/ALUFreeListArray.hh
```diff
@@ -0,0 +1,51 @@
...
+
+    void reserve(int64_t idx);
```
powerjg: Ditto above about `idx`

src/mem/ruby/structures/ALUFreeListArray.hh
```diff
@@ -0,0 +1,51 @@
+#ifndef __MEM_RUBY_STRUCTURES_ALUFREELISTARRAY_HH__
```
powerjg, line 30: Needs a license statement, too.

powerjg (changes requested):
A few more comments/questions after a more careful review :)

src/mem/ruby/structures/ALUFreeListArray.hh
mattsinc: @DanKouch can chime in, but ultimately I believe the idx is representing a bank of the L2 (or a pseudo-bank).  I thought some of the other code in this file was doing the bank mapping -- Dan can correct me if I'm wrong here.  Yes we want the ALUs to be fully associative, but because we need to ensure a given cache line waits the full time while the ALU is allowed to be reused for other cache lines to mimic the pipeline behavior of real GPUs.  So there will need to be some parameter, but we can certainly debate what the parameter is.

configs/ruby/GPU_VIPER.py
mattsinc: I am not aware of this being possible, but I'm also not sure what proof you would want either?

configs/ruby/GPU_VIPER.py
mattsinc: Our idea was that while the GPU currently only uses these parameters, we wanted to set GPU specific parameters/values in VIPER and not set the same values by default for anyone else trying to use these parameters for other code.  But if desired we can certainly make them the same given that no one else uses it as is.

powerjg: Whatever makes sense to reduce user confusion is good for me. Usually having multiple places with different defaults adds to confusion.

configs/ruby/GPU_VIPER.py
powerjg: The code should probably look something like

```
if <user specifies `--tcc_num_atomic_alus`>:
    for cache in <tcc_sim_object_list>:
        cache.resource_stalls = True
```

Without this, I think you'll find many people misusing this code (i.e., thinking that their command line parameter changes something when it doesn't).

src/mem/ruby/structures/ALUFreeListArray.hh
powerjg: Right now, the `idx` that is passed into this function is the set number. I don't think that is the intended behavior. It should probably be the bank number, if that's the intention. The other option, which may be easier, would be to just pass in the address and do all of the bank and associativity logic in this object.

mattsinc:
@DanKouch just making sure you've seen the above comments/questions/requests?

BobbyRBruce:
Is this still actively being worked on? I don't see any movement since July 25th.

mattsinc:
yes

src/mem/ruby/structures/ALUFreeListArray.cc
DanKouch: Added, thanks!

configs/ruby/GPU_VIPER.py
DanKouch: I've updated the defaults in RubyCache.py to match

configs/ruby/GPU_VIPER.py
DanKouch: Updated RubyCache.py to match

configs/ruby/GPU_VIPER.py
DanKouch: It looks like this will happen if `--no-tcc-resource-stalls` is supplied. Since tcc-num-atomic-alus defaults to 64, doing what the above psuedocode specifies would make it so that the `--no-tcc-resource-stalls` is ignored unless `--tcc-num-atomic-alus=0` is also specified, which seems like it might be more confusing. I could have it override if tcc-num-atomic-alus isn't 64, but specifying `--no-tcc-resource-stalls` seems rather explicit.

src/mem/ruby/structures/ALUFreeListArray.hh
DanKouch: The intent here is to keep track of what cache lines are currently in an atomic ALU pipeline. Since the ALUs are intended to be fully-assosciative, we don't need the bank number. In the latest push I've updated the class to take the full address, then determine the cache line from there.

src/mem/ruby/structures/ALUFreeListArray.hh
DanKouch: Addressed above.

configs/ruby/GPU_VIPER.py
mattsinc: Personally I agree with @DanKouch here, but since @powerjg is concerned about this and is away right now, we'll probably have to wait for him to return to comment on this.

src/mem/ruby/structures/ALUFreeListArray.cc
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright (c) 2014 The University of Wisconsin
```
mattsinc: Year should probably be 2023

src/mem/ruby/structures/ALUFreeListArray.hh
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (c) 2014 The University of Wisconsin
```
mattsinc: 2023

src/mem/ruby/structures/ALUFreeListArray.cc
```diff
@@ -0,0 +1,102 @@
...
+
+    for (AccessRecord& record : accessQueue) {
+        // Block access if we would be using more ALUs than we have in a
+        // single tick
+        if (record.startTick == curTick() &&
+          (++accesses_this_tick > this->alus)) {
+            return false;
+        }
+
+        // Block access if the index is already beign used
```
mattsinc: beign --> being

src/mem/ruby/structures/ALUFreeListArray.hh
```diff
@@ -0,0 +1,79 @@
...
+namespace gem5
+{
+
+namespace ruby
+{
+
+class ALUFreeListArray
+{
+  private:
+    unsigned int alus;
```
mattsinc: we should probably call this numALUs instead

src/mem/ruby/structures/ALUFreeListArray.hh
mattsinc: See comment on updated patch

mattsinc (changes requested):
Some of these things @powerjg will have to chime in on once he is back, since what @DanKouch did seems reasonable to me but is contrary to what @powerjg asked for.  Beside that, there are a few minor changes to make in the requested changes below.  Thanks for fixing things @DanKouch !

mattsinc (dismissed):
This meets the requests I had, so I'll approve for now to kick off tests.  But since @powerjg has a different idea of how the resource stall/alu dependence should be handled, we should probably wait until he is back (unless @BobbyRBruce disagrees).

mattsinc:
@DanKouch I also see this at the bottom:

This branch is out-of-date with the base branch
Merge the latest changes from develop into this branch. 

If you see the same thing, you'll need to rebase before this could be merged in.

src/mem/ruby/structures/ALUFreeListArray.cc
```diff
@@ -0,0 +1,102 @@
...
+
+bool ALUFreeListArray::tryAccess(Addr addr)
+{
+    uint32_t accesses_this_tick = 0;
+
+    // Remove requests from the tail of the queue that occured more than
+    // accessLatency ticks ago
+    Tick oldestValidRecordStart = curTick() -
+                                ((this->accessLatency) *
+                                m_ruby_system->clockPeriod());
```
powerjg: This is likely to behave in unexpected ways. What happens if the cache (or controller) has a different clock period than the "ruby system"? I.e., you cannot assume all caches are in the same clock domain.

I suggest taking a `Tick` as the `accessLatency` and converting from cycles to ticks in the controller.

src/mem/ruby/structures/ALUFreeListArray.cc
```diff
@@ -0,0 +1,102 @@
...
+{
+
+/*
+* Behavior:
+*   Requests will go through unless one/both of the following are met:
+*       - There have been more than [numALUs] requests in the current cycle
+*       - The same line has been accessed in the past accessLatency cycles
+*/
+
+ALUFreeListArray::ALUFreeListArray(unsigned int numALUs, Cycles accessLatency,
```
powerjg: I hate to block a change on this, but the code style is not being followed in this file.

Only class member variables should be `camelCase` and all local variables (including function parameters) should be `snake_case`

src/mem/ruby/structures/ALUFreeListArray.cc
```diff
@@ -0,0 +1,102 @@
...
+    uint32_t accesses_this_tick = 0;
+
+    // Remove requests from the tail of the queue that occured more than
+    // accessLatency ticks ago
+    Tick oldestValidRecordStart = curTick() -
+                                ((this->accessLatency) *
+                                m_ruby_system->clockPeriod());
+
+    while (accessQueue.size() > 0 &&
+         (accessQueue.back().startTick < oldestValidRecordStart)){
```
powerjg: Missing a space after parens
```suggestion
         (accessQueue.back().startTick < oldestValidRecordStart)) {
```

src/mem/ruby/structures/ALUFreeListArray.cc
```diff
@@ -0,0 +1,102 @@
...
+        accessQueue.pop_back();
+    }
+
+    for (AccessRecord& record : accessQueue) {
+        // Block access if we would be using more ALUs than we have in a
+        // single tick
+        if (record.startTick == curTick() &&
+          (++accesses_this_tick > this->numALUs)) {
```
powerjg: Why can't you just use `accessQueue.size()`?

src/mem/ruby/structures/ALUFreeListArray.cc
```diff
@@ -0,0 +1,102 @@
...
+
+ALUFreeListArray::ALUFreeListArray(unsigned int numALUs, Cycles accessLatency,
+                                   RubySystem *rs)
+    : m_ruby_system(rs)
+{
+    this->numALUs = numALUs;
+    this->accessLatency = accessLatency;
+}
+
+bool ALUFreeListArray::tryAccess(Addr addr)
```
powerjg, line 66: I'm not convinced this logic actually implements the behavior you're going for. I don't understand why things stay in the queue for some amount of time, but you only check for things in the queue at exactly `curTick`.

Could you describe the behavior you're trying to simulate in this function? With this knowledge, I may be able to help you implement what you're going for.

src/mem/ruby/structures/ALUFreeListArray.hh
```diff
@@ -0,0 +1,80 @@
...
+
+namespace ruby
+{
+
+class ALUFreeListArray
+{
+  private:
+    unsigned int numALUs;
+    Cycles accessLatency;
+    RubySystem *m_ruby_system;
```
powerjg: Should be camelCase and should not have `m` at the beginning. 

Well, actually, I don't think a `RubySystem*` should be needed at all.

src/mem/ruby/structures/CacheMemory.cc
```diff
@@ -73,6 +73,8 @@ CacheMemory::CacheMemory(const Params &p)
               p.start_index_bit, p.ruby_system),
     tagArray(p.tagArrayBanks, p.tagAccessLatency,
              p.start_index_bit, p.ruby_system),
+    atomicALUArray(p.atomicALUs, p.atomicLatency,
```
powerjg: Here you should be able to use a function that's something like `cyclesToTicks(p.atomicLatency)`

src/mem/ruby/structures/RubyCache.py
```diff
@@ -44,6 +44,10 @@ class RubyCache(SimObject):
         "0B", "block size in bytes. 0 means default RubyBlockSize"
     )
 
+    # Atomic parameters only applicable to GPU atomics
+    atomicLatency = Param.Cycles(0, "Cycles for an atomic ALU operation")
```
powerjg, line 49: What does 0 latency mean, and why is that the default?

powerjg (changes requested):
Thanks for the updates. I think this is getting closer to the model we want to have, but there's still some room for improvements :)

mattsinc:
@DanKouch just making sure you saw @powerjg 's comments?

src/mem/ruby/structures/ALUFreeListArray.cc
abmerop: I think the Tick thing makes sense too.

src/mem/ruby/structures/RubyCache.py
abmerop: This may be what the behavior was prior to the patch (assume no latency / latency is part of the enqueue latency for the slicc action)

abmerop (commented):
This looks fine to me, but should address Jason's style comments.

src/mem/ruby/structures/RubyCache.py
powerjg: Ok. That information should be in a comment, then. Users will want to know what "0" means.

src/mem/ruby/structures/ALUFreeListArray.cc
DanKouch: The idea is that there are `numALUs` ALU pipelines that are `accessLatency` cycles deep. A new request can enter each pipeline each cycle, which means that each cycle `numALUs` requests can be added (hence why I'm checking for things at the queue in curTick). `accessLatency` is only being used to make sure that a request to a cache line being processed in a pipeline isn't started until the previous request to that cache line makes it all the way through (which is being done on lines 84-86).

src/mem/ruby/structures/ALUFreeListArray.cc
DanKouch: @powerjg Would the clock period not correspond to the ruby_system passed to CacheMemory (which is wrapped as RubyCache) on say, [line 158 of MOESI_AMD_BASE.py](https://github.com/gem5/gem5/blob/develop/configs/ruby/MOESI_AMD_Base.py#L158)? If it doesn't, would BankedArray, which is used by dataArray and tagArray [also seems to be using this ruby_system](https://github.com/gem5/gem5/blob/develop/src/mem/ruby/structures/BankedArray.cc#L98) to get the clock period in ticks also need to be changed?

configs/ruby/GPU_VIPER.py
DanKouch: > It looks like this will happen if `--no-tcc-resource-stalls` is supplied. Since tcc-num-atomic-alus defaults to 64, doing what the above psuedocode specifies would make it so that the `--no-tcc-resource-stalls` is ignored unless `--tcc-num-atomic-alus=0` is also specified, which seems like it might be more confusing. I could have it override if tcc-num-atomic-alus isn't 64, but specifying `--no-tcc-resource-stalls` seems rather explicit.

@powerjg Are you ok with leaving it as-is in light of this?

src/mem/ruby/structures/ALUFreeListArray.cc
DanKouch: I've updated ALUFreeListArray to take Ticks, rather than Cycles.

src/mem/ruby/structures/RubyCache.py
DanKouch: Added a comment in the latest patchset.

DanKouch:
I believe I've addressed all comments in my latest patch set. I've also rebased.

abmerop:
The changes in commit [af9beea](https://github.com/gem5/gem5/commit/af9beea6e607a21af0817f2b71bea0a2f3c007e1) look good to me.  Seems like something went wrong with the rebase though

configs/ruby/GPU_VIPER.py
powerjg: If the default is that you will have resource stalls, then this is OK. But if the default is that resources stalls are off, then this is very confusing for users other than the person who wrote the code.

src/mem/ruby/structures/ALUFreeListArray.cc
```diff
@@ -0,0 +1,95 @@
...
+
+    while (accessQueue.size() > 0 &&
+         (accessQueue.back().startTick < oldestValidRecordStart)) {
+        accessQueue.pop_back();
+    }
+
+    for (AccessRecord& record : accessQueue) {
+        // Block access if we would be using more ALUs than we have in a
+        // single tick
+        if (record.startTick == curTick() &&
```
powerjg, line 82: Shouldn't this line be `record.startTick <= curTick()`?

Otherwise, you are only checking for *a single tick* not reserving the ALU for the ALU's latency.

Furthermore, the next line is comparing this to the number of things in the queue, which will *always* be `false`. There is no way `accesses_this_tick` could be more than the number of records in `accessQueue`.

Generally, this is difficult to understand code because it's not written in a way that is directly understandable from the purpose. It would be better to have an array of N entries (where N is the number of ALUs). When you reserve and entry, you put the time that the entry will be free in the location. When you are checking for an available entry, you go through and find if there are any entries with `free time < curTick()`. 

I could be misunderstanding this code, but I would encourage you to write the code to be more like the intended behavior. E.g., if you want to "reserve an ALU for `latency` time, then write code that does that instead of using a circular buffer and counts the number of units in use.

src/mem/ruby/structures/ALUFreeListArray.cc
powerjg: I'm going to resolve this conversation, but I *still* think the code is not doing what you intend. See my comment below.

src/mem/ruby/structures/RubyCache.py
powerjg: Thanks!

src/mem/ruby/structures/ALUFreeListArray.cc
DanKouch: @powerjg 
The intent of the code isn't to reserve entire ALUs. As I'm assuming these ALUs are pipelined, multiple operations can be processed in a single ALU at once. _accessQueue doesn't keep track of ALUs_, but rather which cache lines are currently being processed.

The intent is rather to:
1. Restrict the number of ALU operations that can start in a cycle to _numALUs_
    - This is being accomplished in lines 71 to 74 (However, I was wrong in switching numALUs to accessQueue.size(), I'll switch that back as soon as I can)
2. Make sure that no ALU operation to a line already being processed can start until the previous operation has completed
    - This is accomplished in lines 77 to 79

This should model a group of numALUs pipelined ALUs that are accessLatency ticks deep, where a new operation can begin in each ALU every cycle.

Now that I think about it, however, I'd need to double-check that resource constraints for the same cycle are checked on the same tick. If they aren't I would need to change `curTick` to some sort of `curCycle`. I'm pretty sure this is the case, but I'll check again.

powerjg: I see... I misunderstood what you were trying to model. My bad :).

It would be good to update the comments to make this more clear.

configs/ruby/GPU_VIPER.py
DanKouch: Yes, the default is to have resource stalls.

src/mem/ruby/structures/ALUFreeListArray.cc
DanKouch: @powerjg I've added additional comments and made the correction I mentioned.

DanKouch:
@powerjg @mattsinc @abmerop As #367 and now #397 address underlying issues with atomics, I've marked this as a draft PR for the time being. I don't think it should be merged in until those go in.

mattsinc (dismissed):
From my perspective this is ok, although the open comments should probably be resolved by @powerjg since they reflect his concerns.

src/mem/ruby/structures/ALUFreeListArray.cc
powerjg: Thanks for the updates!

powerjg (dismissed):
lgtm. thanks for the updates!

mattsinc:
@DanKouch it looks like there is a merge conflict in this patch, based on the above:

This branch has conflicts that must be resolved
Use the [web editor](https://github.com/gem5/gem5/pull/120/conflicts) or the to resolve conflicts.
Conflicting files
configs/ruby/GPU_VIPER.py

Please resolve this soon so we can push it in :)

mattsinc:
@DanKouch : follow-up: the merge conflict is from #389 by @v-ramadas

mattsinc:
@DanKouch : now that #397 is merged in, this is next.  It looks like we need to rebase to head of develop and resolve some merge conflicts though.  Let me know if you need help with this.  I'd like this to be merged in too with v23.1 since #397 is (and both are part of a larger fix).

mattsinc (approved):
As far as I can tell, this just rebased the previous commit and resolved according merge issues, so carrying forward approval.

abmerop (approved):

mattsinc:
@abmerop : any objections to merging this in?

abmerop:
None

PR #124, opened 2023-07-26
mem: Make functional request a response when satisfied by queue

atrib:
Duplicate of #119 (remade to make life easier)

In the memory controller, MemCtrl::MemoryPort::recvFunctional, when the functional request is satisfied by the ctrl-response queue, correctly make the packet a response.

This change mirrors AbstractMemory::functionalAccess, which uses Packet::makeResponse() after satisfying the request.

Note:
bool trySatisfyFunctional(..) functions return true or false based on whether the request was satisfied.
void recvFunctional(..) functions modify the packet to indicate successful request satisfaction.


powerjg:
Thanks for cleaning it up. Next time, you can just force push and the original PR will be updated.

powerjg (approved):

PR #134, opened 2023-07-27
util: Ignore line length check for #include pragma in C/C++ files

takekoputa:
The length of the path of #include pragmas can be more than
79-character long. The length of the path of a #include pragma
can be outside of user's control.

### DIFF
--- a/src/sst/sst_responder_interface.hh
+++ b/src/sst/sst_responder_interface.hh
@@ -30,7 +30,7 @@
 
 #include <string>
 
-#include "mem/port.hh"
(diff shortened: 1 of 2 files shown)


src/sst/sst_responder_interface.hh
```diff
@@ -30,7 +30,7 @@
 
 #include <string>
 
-#include "mem/port.hh"
+#include "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaamem/port.hh"
```
BobbyRBruce: You've accidentally left this testing code here.

BobbyRBruce (changes requested):

BobbyRBruce (approved):

PR #137, opened 2023-07-27
arch-riscv: Implemented zicbom/zicboz extensions for RISC V

rogerchang23424:
Change-Id: I79d0e6059a2dbb5a0057c4f7489b999f9e803684

### DIFF
--- a/src/arch/riscv/isa/decoder.isa
+++ b/src/arch/riscv/isa/decoder.isa
@@ -677,6 +677,23 @@ decode QUADRANT default Unknown::unknown() {
                 0x1: fence_i({{
                 }}, uint64_t, IsNonSpeculative, IsSerializeAfter, No_OpClass);
             }
+
+            0x2: decode FUNCT12 {
+                format CBMOp {
+                    0x0: cbo_inval({{
+                        Mem = 0;
+                    }}, mem_flags=[INVALIDATE, DST_POC]);
+                    0x1: cbo_clean({{
+                        Mem = 0;
+                    }}, mem_flags=[CLEAN, DST_POC]);
+                    0x2: cbo_flush({{
+                        Mem = 0;
+                    }}, mem_flags=[CLEAN, INVALIDATE, DST_POC]);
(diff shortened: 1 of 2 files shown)


rogerchang23424:
This change is from: https://gem5-review.googlesource.com/c/public/gem5/+/71958

powerjg (commented):
This change is OK. I wonder if we should add some warnings because I don't think that most of the cache models (Ruby or classic) support the INVALIDATE, CLEAN, or CACHE_BLOCK_ZERO flags.

Thoughts?

rogerchang23424:
Pending until the RVV chain(PR#83) merged

powerjg:
> I wonder if we should add some warnings because I don't think that most of the cache models (Ruby or classic) support the INVALIDATE, CLEAN, or CACHE_BLOCK_ZERO flags.

Any thoughts on this?

powerjg:
The RVV chain (#83) has been merged. Can you respond to my question above before we merge this?

rogerchang23424:
> This change is OK. I wonder if we should add some warnings because I don't think that most of the cache models (Ruby or classic) support the INVALIDATE, CLEAN, or CACHE_BLOCK_ZERO flags.
> 
> Thoughts?

Hi,

The must of flags are defined in `mem/packets.hh` and CMOs are implemented in separate CPU models. Following are details about CMOs:

For the cache operation define with memory flags:
https://github.com/gem5/gem5/blob/stable/src/mem/request.hh#L1111

The Memcmd will update according to the memory flags:
https://github.com/gem5/gem5/blob/stable/src/mem/packet.hh#L993

For CACHE_BLOCK_ZERO
https://github.com/gem5/gem5/blob/stable/src/mem/request.hh#L260

I think it should warn inside the model if the model is not implemented for the specify MemCmd

powerjg:
> I think it should warn inside the model if the model is not implemented for the specify MemCmd

Sounds good. I don't think that we do a good job of this, but you're right :)

powerjg (approved):

Issue #138, opened 2023-07-27
Daily GPU tests fail to output the random seed used


mkjost0:
As mentioned in pull request #131 , the workflow for dailies needs to be altered so that the random seed used for tests is output in order to allow for us to recreate tests.  More specifics copied from the PR below:

"This test is using a random time-based seed and the tester does not print out what the value is. This could make it very difficult in the future to reproduce the test. It happens to be fine here as every seed will fail for sure. The protocol tester should print out what the seed is or the github action command should explicitly add --random-seed=."

BobbyRBruce:
Why do we have non-determinism anywhere in gem5? To me this isn't a problem with the tests, there shouldn't be anything truely random going on. The "time-based seed" should be replaced with something static.

mkjost0:
Yeah, I was thinking of just altering the files in the tests directory to use `--random-seed=` so that we don't have to make more changes to the workflow directory.  If that's okay, I can just assign this to myself and get that done.

mattsinc:
I had asked @DanKouch to do the same @mkjost0 , but if you want to get to it today, please go ahead.

@abmerop (or really, Brad) will have to comment more on @BobbyRBruce 's original question though, as I was not involved with that work.

powerjg:
Changing the command line parameter is the best thing to do in the short term (I think). 

Longer term, the problem is this line: https://github.com/gem5/gem5/blob/af72b9ba580546ac12ce05bfaac3fd53fa8699f4/src/cpu/testers/gpu_ruby_test/protocol_tester.cc#L148

We *should* be using `mt_random` here instead of `srand`. We should always use gem5's random number generator so that we have determinism by default.

Then, when we want to have multiple different seeds used, we can use (from *python*) `m5._m5.core.seedRandom(random.randint(0,100000))` (or something like this). We should expose this `seedRandom` function better, though.

Anyway, the "right" solution is a bit deeper than is worth it at this point. For now, let's just set the command line parameter.

BobbyRBruce:
> Changing the command line parameter is the best thing to do in the short term (I think).
> 
> Longer term, the problem is this line:
> 
> https://github.com/gem5/gem5/blob/af72b9ba580546ac12ce05bfaac3fd53fa8699f4/src/cpu/testers/gpu_ruby_test/protocol_tester.cc#L148
> 
> We _should_ be using `mt_random` here instead of `srand`. We should always use gem5's random number generator so that we have determinism by default.
> 
[...]

PR #140, opened 2023-07-28
gpu-compute: "<random>" -> "base/random.hh" in testers/gpu...

BobbyRBruce:
In "src/cpu/testers/gpu_ruby_test" a random number generator was used. This was using the CPP "<random>" library. This patch changes it to the gem5 random class (that declared in "base/random.hh").

In addition to this, undeterministic behavior has been removed. Via "protocol_tester.cc" the RNG is either seeded with a seed specified by the user, or goes with the gem5 default seed. This ensures reproducable runs. Prior to this patch the RNG was seeded with `time(NULL)`. This made finding faults difficult.

This, at least partially, addresses Issue #138

Change-Id: Ia8e9f7b87e91323f828e0b7f6c3906c0c5793b2c

### DIFF
--- a/src/cpu/testers/gpu_ruby_test/ProtocolTester.py
+++ b/src/cpu/testers/gpu_ruby_test/ProtocolTester.py
@@ -73,7 +73,7 @@ class ProtocolTester(ClockedObject):
     random_seed = Param.Int(
         0,
         "Random seed number. Default value (0) means \
-                                using runtime-specific value.",
+                                using base/random.hh without seed.",
     )
     log_file = Param.String("Log file's name")
     system = Param.System(Parent.any, "System we belong to")
--- a/src/cpu/testers/gpu_ruby_test/address_manager.cc
+++ b/src/cpu/testers/gpu_ruby_test/address_manager.cc
@@ -33,7 +33,6 @@
 
 #include <algorithm>
 #include <climits>
-#include <random>
 
 #include "base/intmath.hh"
 #include "base/logging.hh"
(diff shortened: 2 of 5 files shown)


abmerop (dismissed):

abmerop:
Thanks for coding this up

powerjg (commented):
Just FYI, this is going to make a significant user-facing difference when using `configs/example/gpu_random_tester.py`

Before, if you ran that script without any parameters, it would give you a different random seed. Now, you will need to specify a *different* random seed on the command line each time.

Maybe we should add a `warn` to the run script to notify the users?

(Also, thanks for taking the lead here, Bobby!!!)

BobbyRBruce:
> Just FYI, this is going to make a significant user-facing difference when using `configs/example/gpu_random_tester.py`
> 
> Before, if you ran that script without any parameters, it would give you a different random seed. Now, you will need to specify a _different_ random seed on the command line each time.
> 
> Maybe we should add a `warn` to the run script to notify the users?
> 
> (Also, thanks for taking the lead here, Bobby!!!)

I've added a warn. Please look it over and see it's saying what we want.

BobbyRBruce:
Woops. I messed up this PR. Give me a moment to fix it.

BobbyRBruce:
Ok, I fixed it :).

powerjg (commented):
This looks good to me, but I think that we should let it sit here until @mattsinc @abmerop or someone who uses this script regularly approves.

abmerop (approved):
Looks good, thanks again!

mattsinc (commented):
+1, this looks good.  @BobbyRBruce just making sure you updated the corresponding yaml scripts to pass a seed now?

PR #141, opened 2023-07-28
dev-amdgpu: Support for ROCm 5.4+ and MI200

abmerop:


src/dev/amdgpu/sdma_engine.cc
```diff
@@ -510,9 +510,12 @@ SDMAEngine::decodeHeader(SDMAQueue *q, uint32_t header)
         dmaReadVirt(q->rptr(), sizeof(sdmaAtomic), cb, dmaBuffer);
         } break;
       case SDMA_OP_CONST_FILL: {
-        q->incRptr(sizeof(sdmaConstFill));
-        warn("SDMA_OP_CONST_FILL not implemented");
-        decodeNext(q);
+        DPRINTF(SDMAEngine, "SDMA Constant fill packet\n");
+        dmaBuffer = new sdmaConstFill();
+        cb = new DmaVirtCallback<uint64_t>(
+            [ = ] (const uint64_t &)
```
BobbyRBruce, line 516: I'm confused by this... I've never seen this before. Can you explain it?

abmerop: This is creating a callback for dmaReadVirt using lambda function (which begins at the [=] part).  The callback is called when the dmaReadVirt call finishes doing all translation and all of the DMA requests have returned from memory (since the size could be arbitrarily large this may result in multiple page-size DMAs we need to wait for).  It will call the constFill function in this class when it calls back

configs/example/gpufs/runfs.py
```diff
@@ -179,10 +179,12 @@ def runGpuFSSystem(args):
...
 
-    # Verify MMIO trace is valid
-    mmio_md5 = hashlib.md5(open(args.gpu_mmio_trace, "rb").read()).hexdigest()
-    if mmio_md5 != "c4ff3326ae8a036e329b8b595c83bd6d":
-        m5.util.panic("MMIO file does not match gem5 resources")
+    # Verify MMIO trace is valid. This is only needed for Vega10 simulations.
+    if args.gpu_device == "Vega10":
+        mmio_file = open(args.gpu_mmio_trace, "rb")
+        mmio_md5 = hashlib.md5(mmio_file.read()).hexdigest()
+        if mmio_md5 != "c4ff3326ae8a036e329b8b595c83bd6d":
```
mattsinc, line 189: Can you please add a comment about what this SHA means/represents?

src/arch/amdgpu/vega/pagetable_walker.cc
```diff
@@ -239,9 +239,22 @@ Walker::WalkerState::walkStateMachine(PageTableEntry &pte, Addr &nextRead,
...
 
-    // For a four level page table block fragment size should not be needed.
-    // For now issue a panic to prevent strange behavior if it is non-zero.
-    panic_if(pde.blockFragmentSize, "PDE blockFragmentSize must be 0");
+    // Block fragment size can change the size of the pages pointed to while
+    // moving to the next PDE. A value of 0 implies native page size. A
+    // non-zero value implies the next leaf in the page table is a PTE unless
+    // the F bit is set. If we see a non-zero value, set it here and print
+    // for debugging.
+    if (pde.blockFragmentSize) {
```
mattsinc, line 247: I'm guessing this is only set for MI200/ROCm 5?  Because if we run GPUSE with ROCm still, do we want this to panic?

mattsinc (changes requested):
In addition to my comments, my guess is this only applies to GPUFS?  If so, we should probably update the commit message accordingly.  But mostly my biggest concern is what would happen with GPUSE and not ROCm 5.4 if we run this code?

src/arch/amdgpu/vega/pagetable_walker.cc
abmerop: SE mode uses X86 page table walker, so this code isn't used there.  The field starts being used around ROCm 5.4

configs/example/gpufs/runfs.py
abmerop: Added comment

mattsinc (approved):
This looks good from my perspective.  @BobbyRBruce please chime in if you feel differently.

BobbyRBruce:
I'm fine with this. Thanks for the explanation!

PR #149, opened 2023-08-01
stdlib, resources: fixed style issue in isa.hh

### LINKED ISSUE #147: Fix style in RISC-V isa file
https://github.com/gem5/gem5/blob/dceabe5fda598937b00959cc648fb7f2fbbe950a/src/arch/riscv/isa.hh#L73C19-L73C19

This should be `rvType` instead of `rv_type`.

Harshil2107:
Changed "rv_type" to "rvType".

Change-Id: I7432a87d7a37324777385707854aefba2475b98c


src/arch/riscv/isa.hh
```diff
@@ -70,7 +70,7 @@ enum FPUStatus
 class ISA : public BaseISA
 {
   protected:
-    RiscvType rv_type;
```
rogerchang23424, line 75: It has the public `rvType` for public function. Perhaps it should be `_rvType`?

rogerchang23424 (commented):
Thanks for code style fix.

src/arch/riscv/isa.hh
powerjg: +1 I think this could cause a build issue?

Harshil2107: I changed all occurrences to _rvType and just changing that line was also causing errors. The errors are in isa.cc is it ok to change rv_type to _rvType there as well?

powerjg:
I hate to ask, but could you squash these two commits into one?

Harshil2107:
I think I squashed them, please let me know if they are not.

powerjg (approved):

rogerchang23424 (approved):

PR #151, opened 2023-08-02
New function to kernel_disk_workload to allow new disk device location

leoredivo:
Added a parameter (_disk_device) to kernel_disk_workload which allows users to change the disk device location. get_disk_device() now chooses between the parameter and, if no parameter was passed, it calls a new function _get_default_disk_device() which is implemented by each board and has a default disk device according to each board, eg /dev/hda in the x86_board. The previous way of setting a disk device still exists as a default, however, with the new function users can now override this default

### DIFF
--- a/src/python/gem5/components/boards/arm_board.py
+++ b/src/python/gem5/components/boards/arm_board.py
@@ -354,7 +354,7 @@ class ArmBoard(ArmSystem, AbstractBoard, KernelDiskWorkload):
         )
 
     @overrides(KernelDiskWorkload)
-    def get_disk_device(self):
+    def _get_default_disk_device(self):
         return "/dev/vda"
 
     @overrides(KernelDiskWorkload)
--- a/src/python/gem5/components/boards/arm_board.py
+++ b/src/python/gem5/components/boards/arm_board.py
@@ -354,7 +354,7 @@ class ArmBoard(ArmSystem, AbstractBoard, KernelDiskWorkload):
         )
 
     @overrides(KernelDiskWorkload)
-    def get_default_disk_device(self):
+    def get_disk_device(self):
         return "/dev/vda"
 
     @overrides(KernelDiskWorkload)
--- a/src/python/gem5/components/boards/arm_board.py
+++ b/src/python/gem5/components/boards/arm_board.py
@@ -354,7 +354,7 @@ class ArmBoard(ArmSystem, AbstractBoard, KernelDiskWorkload):
         )
 
     @overrides(KernelDiskWorkload)
-    def _get_default_disk_device(self):
+    def get_default_disk_device(self):
         return "/dev/vda"
 
     @overrides(KernelDiskWorkload)
--- a/src/python/gem5/components/boards/arm_board.py
+++ b/src/python/gem5/components/boards/arm_board.py
@@ -386,6 +386,7 @@ class ArmBoard(ArmSystem, AbstractBoard, KernelDiskWorkload):
             "lpj=19988480",
             "norandmaps",
             "root={root_value}",
+            "disk_device={disk_device}",
             "rw",
             f"mem={self.get_memory().get_size()}",
         ]
--- a/src/python/gem5/components/boards/experimental/lupv_board.py
+++ b/src/python/gem5/components/boards/experimental/lupv_board.py
@@ -539,7 +539,7 @@ class LupvBoard(AbstractSystemBoard, KernelDiskWorkload):
         return ["console=ttyLIO0", "root={root_value}", "rw"]
 
     @overrides(KernelDiskWorkload)
-    def set_default_disk_device(self) -> str:
+    def get_default_disk_device(self) -> str:
         return "/dev/lda"
 
     @overrides(KernelDiskWorkload)
--- a/src/python/gem5/components/boards/experimental/lupv_board.py
+++ b/src/python/gem5/components/boards/experimental/lupv_board.py
@@ -539,7 +539,7 @@ class LupvBoard(AbstractSystemBoard, KernelDiskWorkload):
         return ["console=ttyLIO0", "root={root_value}", "rw"]
 
     @overrides(KernelDiskWorkload)
-    def get_disk_device(self) -> str:
+    def set_default_disk_device(self) -> str:
         return "/dev/lda"
 
     @overrides(KernelDiskWorkload)
--- a/src/python/gem5/components/boards/experimental/lupv_board.py
+++ b/src/python/gem5/components/boards/experimental/lupv_board.py
@@ -535,12 +535,17 @@ class LupvBoard(AbstractSystemBoard, KernelDiskWorkload):
         fdt.writeDtbFile(os.path.join(outdir, "device.dtb"))
 
     @overrides(KernelDiskWorkload)
-    def get_default_kernel_args(self) -> List[str]:
-        return ["console=ttyLIO0", "root={root_value}", "rw"]
-
-    @overrides(KernelDiskWorkload)
     def get_default_disk_device(self) -> str:
         return "/dev/lda"
+
+    @overrides(KernelDiskWorkload)
+    def get_default_kernel_args(self) -> List[str]:
+        return [
+            "console=ttyLIO0",
+            "root={root_value}",
+            "disk_device={disk_device}",
+            "rw",
+        ]
 
     @overrides(KernelDiskWorkload)
     def _add_disk_to_board(self, disk_image: AbstractResource) -> None:
--- a/src/python/gem5/components/boards/experimental/lupv_board.py
+++ b/src/python/gem5/components/boards/experimental/lupv_board.py
@@ -535,7 +535,7 @@ class LupvBoard(AbstractSystemBoard, KernelDiskWorkload):
         fdt.writeDtbFile(os.path.join(outdir, "device.dtb"))
 
     @overrides(KernelDiskWorkload)
-    def get_default_disk_device(self) -> str:
+    def get_disk_device(self) -> str:
         return "/dev/lda"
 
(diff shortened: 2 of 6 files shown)


mkjost0:
Hi, this test is failing both the pre-commit and Change-ID checks.

In order for the pre-commit test to pass, please ensure all code committed follows the style for gem5.  You can run `util/pre-commit-install.sh` to install pre-commit, and run `pre-commit run <hook_id>` on your commit to automatically reformat your files.  More information about pre-commit can be found in the pre-commit section here: https://www.gem5.org/documentation/general_docs/common-errors/

In order for the Change-ID test to pass, you must ensure you commit message has a Change-ID at the bottom. To amend your commit, run the following commands: 
```
f=.git/hooks/commit-msg ; mkdir -p  ;  curl -Lo  https://gerrit-review.googlesource.com/tools/hooks/commit-msg ; chmod +x
git commit --amend --no-edit
```
To learn more about Change-Ids, read the Change-ID section here: https://www.gem5.org/documentation/general_docs/common-errors/

src/python/gem5/components/boards/experimental/lupv_board.py
```diff
@@ -539,7 +539,7 @@ def get_default_kernel_args(self) -> List[str]:
         return ["console=ttyLIO0", "root={root_value}", "rw"]
 
     @overrides(KernelDiskWorkload)
-    def get_disk_device(self) -> str:
+    def set_default_disk_device(self) -> str:
```
BobbyRBruce: I think this is supposed to be `get_default_disk_device`, not `set...`.

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -139,6 +154,7 @@ def set_kernel_disk_workload(
         kernel: KernelResource,
         disk_image: DiskImageResource,
         bootloader: Optional[BootloaderResource] = None,
+        _disk_device: Optional[str] = None,
```
BobbyRBruce: `_disk_device` -> `disk_device`.

We shouldn't have a leading underscore for parameter names.

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -192,6 +211,7 @@ def set_kernel_disk_workload(
         if bootloader is not None:
             self._bootloader = [bootloader.get_local_path()]
 
+
```
BobbyRBruce: Random new line here. You should probably remove it just so our git blame doesn't get too messy.

src/python/gem5/components/boards/arm_board.py
```diff
@@ -354,7 +354,7 @@ def _add_pci_device(self, pci_device: PciVirtIO) -> None:
         )
 
     @overrides(KernelDiskWorkload)
-    def get_disk_device(self):
+    def _get_default_disk_device(self):
```
BobbyRBruce: This is a virtual function, so i think, in this case, the method shouldn't lead with an underscore.

BobbyRBruce: Ditto for other instances in this patch where the leading underscore has been applied.

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -84,11 +85,25 @@ def get_default_kernel_args(self) -> List[str]:
...
 
-    @abstractmethod
     def get_disk_device(self) -> str:
         """
         Get the disk device, e.g., "/dev/sda", where the disk image is placed.
 
+        :returns: The disk device.
+        """
+        if self._disk_device is None:
+            warn("No disk device set, ie where the disk image is located. Defaulting to board disk device")
```
BobbyRBruce: I think a warning here is probably not necessary. Up until this PR using these defaults was normal behavior. How about we change this from a `warn` to an `inform` (`from m5.util import inform`). I'd simply state, for either case, 'Disk Device set to "{VALUE HERE}"'.

BobbyRBruce (changes requested):

BobbyRBruce:
I'm aware that the the defaults in the board are problematic, so I appreciate this fix. There's a big gap in my knowledge regarding exactly how operating systems decide what the disk device is called. I don't have a good understanding of where "/dev/hda", for example,  is specified. I had assumed when I originally coded this that it was something bound tightly with the hardware and low-level OS. I.e., a disk inserted in slot was bound to "/dev/hda" in Linux, and we could rely on this being consistent. So it made sense to me to fix the disk device name to the board in a 1-to-1 manner. I realize now this 1-to-1 mapping isn't always right. 

I don't have a good understanding of how we're to know, in advance, what a computer system's disk device is called. Even when we can pass the disk device name via the `set_kernel_disk_workload` function I don't know how we're supposed to know what to pass.  If not the boards, where should we be declaring these disk device names?

Nothing I say here is a blocker for the patch. I'm just asking questions in a hope someone might have more of an understanding than I do.

BobbyRBruce:
The pre-commit checks failed. You can fix this locally using:

```sh
./util/pre-commit-install.sh # Ensure the hooks are installed
pre-commit run --files <files modified in this patch>
git add <files fixed by pre-commit>
git commit -m "misc: Fix pre-commit formatting issues"
```

src/python/gem5/components/boards/kernel_disk_workload.py
leoredivo: I changed this parameter to disk_device, however, I kept leading underscore in the member name  (self._disk_device), since it isn't a sim obj and it wasn't running when I removed it.

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -84,11 +85,25 @@ def get_default_kernel_args(self) -> List[str]:
...
 
+        :returns: The disk device.
+        """
+        if self._disk_device is None:
+            self._disk_device = get_default_disk_device()
+            inform(f"Disk Device set to {self._disk_device}")
+            return self._disk_device
+        else:
+            inform(f"Disk Device set to {self._disk_device}")
+            return self._disk_device
```
BobbyRBruce:
```suggestion
       to_return = self._disk_device if self.disk_device else get_default_disk_device()
        assert to_return is not None
        return self._disk_device
```

BobbyRBruce: I admit this suggestion may not pass our black, so download it and check...

Let's move the inform to the inform to somewhere here: https://github.com/gem5/gem5/blob/4114114beda2996817f3f0cfc353b505ed308a4b/src/python/gem5/components/boards/kernel_disk_workload.py#L182

Here we can Inform the user of the command being passed to the kernel, which will encapsulate this information along with anything else passed to the kernel. I think would be quite useful information for the user.

BobbyRBruce (changes requested):

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -182,7 +201,8 @@ def set_kernel_disk_workload(
         self.workload.command_line = (
             " ".join(kernel_args or self.get_default_kernel_args())
         ).format(
-            root_value=self.get_default_kernel_root_val(disk_image=disk_image)
+            root_value=self.get_default_kernel_root_val(disk_image=disk_image),
+            disk_device=self.get_disk_device(),
```
BobbyRBruce: Ok, we're making progress on this! How about we improve this even more. Why don't we remove `get_disk_device` function entirely and move the logic down into `set_kernel_disk_workload`.

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -39,6 +39,7 @@
 from pathlib import Path
 
 import m5
+from m5.util import inform
```
BobbyRBruce: You import this bug never use it. I did ask for an inform to be thrown when setting the kernel arguments, but this is never done. Did you forget to commit it?

BobbyRBruce (changes requested):

BobbyRBruce:
Sorry for the delayed feedback on this!

BobbyRBruce:
Hey @leoredivo : Are you still working on this?

leoredivo:
Hey @BobbyRBruce, yeah I'm still working on it, just been working on some other projects for the past few weeks. I should get an updated version to you by the end of the week.

BobbyRBruce (dismissed):

BobbyRBruce:
@leoredivo : The tests are failing with errors like this:

```sh
AttributeError: object 'ArmBoard' has no attribute 'get_disk_device'
  (C++ object is not yet constructed, so wrapped C++ methods are unavailable.)

At:
  src/python/m5/SimObject.py(851): __getattr__
  src/python/gem5/components/boards/kernel_disk_workload.py(133): get_default_kernel_root_val
  src/python/gem5/components/boards/kernel_disk_workload.py(189): set_kernel_disk_workload
  /__w/gem5/gem5/tests/gem5/arm_boot_tests/configs/arm_boot_exit_run.py(201): <module>
```

I believe this is because you've not replaced `get_default_disk_device` with `get_disk_device` in all the boards which use `kernel_disk_workload`. Can you push a fix for this?

BobbyRBruce (dismissed):

BobbyRBruce:
Failing with the following

> AttributeError: object 'ArmBoard' has no attribute 'get_default_disk_device'
  (C++ object is not yet constructed, so wrapped C++ methods are unavailable.)

BobbyRBruce:
~~You still haven't addressed the bug. The ArmBoard is still missing that function.~~

Never-mind. I wasn't paying close enough attention to the patch.

BobbyRBruce (approved):

PR #153, opened 2023-08-03
stdlib, resources: Fixed keyerror: 'is_zipped' bug

### LINKED ISSUE #127: Unable to access some gem5 resources when using Resource()
Summary: 
The error "Resource with name 'resource_name' does not exist" occurs with some gem5 resources. This includes x86-matrix-multiply-omp, x86-bubblesort, x86-floatmm, riscv-bubblesort, and arm-bubblesort. The error can be replicated by attempting to obtain a resource with Resource("resource_name") in a Python script for setting up a simulation. The fix that I found for x86-matrix-multiply-omp was to change the url for x86-matrix-multiply-omp in the resources.json file. This would not work for x86-bubblesort and x86-floatmm, as they were entirely absent from resources.json.

Detailed:
When attempting to access a number of gem5 resources with Resource(resource_name), the following error occurs:
```
Exception: Error: Resource with name 'x86-matrix-multiply-omp' does not exist

At:
  build/X86/python/gem5/resources/downloader.py(362): get_resources_json_obj
  build/X86/python/gem5/resources/resource.py(168): __init__
  materials/testing-resources/x86_matrix_multiply_omp_example.py(22): <module>
  build/X86/python/m5/main.py(597): main
```
I used x86-matrix-multiply-omp for testing, but other affected resources include x86-bubblesort, x86-floatmm, and riscv-bubblesort.

I was able to get past this error message by changing the link in _get_resources_json_uri() to 
https://raw.githubusercontent.com/gem5/gem5-resources/stable/resources.json. It now looks like this:

```
def _get_resources_json_uri() -> str:
    #return "https://resources.gem5.org/resources.json"
    return "https://raw.githubusercontent.com/gem5/gem5-resources/stable/resources.json"
```
After this change, an HTTP 404 error occurred. I attempted to access the url for x86-matrix-multiply-omp in resources.json and observed that no file was hosted there. 

To fix this, I set the environment variable for "GEM5_RESOURCE_JSON" to the file path to my local copy of resources.json. Then, I changed the url in the json to a url from the gem5 resources website. The change to the environment variable was in the function definition of _get_resources_json(), in build/X86/python/gem5/resources/downloader.py.

```
old url:
{url_base}/test-progs/matrix-multiply-omp/x86-matrix-multiply-omp-20230127

where url_base is http://dist.gem5.org/dist/v22-1

new url:
http://dist.gem5.org/dist/develop/test-progs/matrix-multiply-omp/x86-matrix-multiply-omp-20230127 

```
At this point, the script was able to obtain the resource.

I attempted to use the same fix for x86-floatmm and x86-bubblesort, but they were both absent from the json.

Harshil2107:
Change-Id: I68fffd880983ebc225ec6fc8c7f8d509759b581d

### DIFF
--- a/src/python/gem5/resources/downloader.py
+++ b/src/python/gem5/resources/downloader.py
@@ -280,17 +280,21 @@ def get_resource(
         # string-based way of doing things. It can be refactored away over
         # time:
         # https://gem5-review.googlesource.com/c/public/gem5-resources/+/51168
-        if isinstance(resource_json["is_zipped"], str):
-            run_unzip = unzip and resource_json["is_zipped"].lower() == "true"
-        elif isinstance(resource_json["is_zipped"], bool):
-            run_unzip = unzip and resource_json["is_zipped"]
-        else:
-            raise Exception(
-                "The resource.json entry for '{}' has a value for the "
-                "'is_zipped' field which is neither a string or a boolean.".format(
-                    resource_name
+        run_unzip = False
+        if "is_zipped" in resource_json:
+            if isinstance(resource_json["is_zipped"], str):
+                run_unzip = (
+                    unzip and resource_json["is_zipped"].lower() == "true"
                 )
-            )
+            elif isinstance(resource_json["is_zipped"], bool):
+                run_unzip = unzip and resource_json["is_zipped"]
+            else:
+                raise Exception(
+                    "The resource.json entry for '{}' has a value for the "
+                    "'is_zipped' field which is neither a string or a boolean.".format(
+                        resource_name
+                    )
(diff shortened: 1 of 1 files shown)


BobbyRBruce (approved):
Thanks for this!

Just for those who may come across this patch and wonder what it's fixing: the `if isinstance(resource_json["is_zipped"], str):` statement caused a bug when a resource did not have contain the `is_zipped` field. This field is optional so not specifying this in a resource entry should be supported. This patch fixes this bug by adding a check: `if "is_zipped" in resources_json:`.

PR #156, opened 2023-08-03
tests: Refactor test configs

mkjost0:
The yaml file changes made here will need to be copied to develop for our scheduled tests to run properly.  In addition, the original comments on this can be seen here:  https://gem5-review.googlesource.com/c/public/gem5/+/70340/2


tests/gem5/asmtest/README.md
```diff
@@ -0,0 +1,9 @@
+# ASM Test
+
+These tests run a set of RISCV binaries on a bare bones syscall execution.
+In addition, these test run these binaries against different CPU types.
+To run these tests by themselves, you can run the following command in the tests directory:
+
+```bash
+./main.py run gem5/asmtest --length=[length]
```
powerjg, line 8: What does `--length` mean in this case?

I guess this is copy pasted across all of these readme's, but does it make sense for this one?

I suggest listing the tests in these readme and saying whether they are quick/long/very long.   HOWEVER, I suggest doing that in another change 😄

tests/gem5/kvm_switch_tests/README.md
```diff
@@ -0,0 +1,8 @@
+# KVM Switch Tests
+
+These tests ensure that gem5 can switch processors during simulation.
+To run these tests by themselves, you can run the following command in the tests directory:
+
```
powerjg, line 5: Similar to above, in another PR, it would be good to specify the system requirements to run these tests.

tests/gem5/stdlib/configs/simple_binary_run.py
powerjg, line 1: How much of this code is duplicated in other files? 

There's a tradeoff between reducing code duplication and complexity. But I worry a bit about how much we are going to have to refactor in the tests if something changes.

powerjg (commented):
Just some comments/questions and notes for the future. This looks good to me. I'll let Bobby approve, though.

tests/gem5/stdlib/configs/simple_binary_run.py
mkjost0: Some of the code is duplicated, as every instance of simple_binary_run.py in the testing directory started from the same one file.  However, splitting it up got rid of a lot of the optional parameters that used to be at the top, and duplicating it makes it easier to see exactly what's being run in each directory.

tests/gem5/kvm_switch_tests/README.md
mkjost0: Sounds good, I can make the extra changes to the READMEs after this PR gets in

powerjg (dismissed):

BobbyRBruce (approved):

PR #166, opened 2023-08-08
Fix reporting traps (faults) to GDB in SE mode

### LINKED ISSUE #123: Reporting traps (faults) to GDB in SE mode is broken
While using gem5 in SE mode I realized that reporting traps to GDB (if connected) is broken / missing at least on 
POWER / RISC-V. 

Consider POWER: https://github.com/gem5/gem5/blob/develop/src/arch/power/faults.cc#L42-L63. `trapToGdb` returns `true` is GDB is connected *and* trap has been reported, otherwise `false`. But the code in `TrapFault::invoke` panics *if* there's a GDB connected and does nothing if there's no GDB at all! I'm pretty sure the logic should be inverse. 

On RISC-V (and possibly other architectures as well) there's no such logic at all, look  at https://github.com/gem5/gem5/blob/develop/src/arch/riscv/faults.cc, so on RISC-V one does not see software breakpoints hitting at all when connecting GDB to debug program running in gem5.

janvrany:
This addresses #123

### DIFF
--- a/src/arch/riscv/faults.cc
+++ b/src/arch/riscv/faults.cc
@@ -205,9 +205,11 @@ UnknownInstFault::invokeSE(ThreadContext *tc, const StaticInstPtr &inst)
 void
 IllegalInstFault::invokeSE(ThreadContext *tc, const StaticInstPtr &inst)
 {
-    auto *rsi = static_cast<RiscvStaticInst *>(inst.get());
-    panic("Illegal instruction 0x%08x at pc %s: %s", rsi->machInst,
-        tc->pcState(), reason.c_str());
+    if (! tc->getSystemPtr()->trapToGdb(GDBSignal::ILL, tc->contextId()) ) {
+        auto *rsi = static_cast<RiscvStaticInst *>(inst.get());
+        panic("Illegal instruction 0x%08x at pc %s: %s", rsi->machInst,
+            tc->pcState(), reason.c_str());
+    }
 }
 
 void
@@ -226,7 +228,9 @@ IllegalFrmFault::invokeSE(ThreadContext *tc, const StaticInstPtr &inst)
 void
 BreakpointFault::invokeSE(ThreadContext *tc, const StaticInstPtr &inst)
 {
-    schedRelBreak(0);
+    if (! tc->getSystemPtr()->trapToGdb(GDBSignal::TRAP, tc->contextId()) ) {
+        schedRelBreak(0);
+    }
 }
 
 void
--- a/src/arch/riscv/faults.cc
+++ b/src/arch/riscv/faults.cc
@@ -164,8 +164,6 @@ RiscvFault::invoke(ThreadContext *tc, const StaticInstPtr &inst)
         pc_state.set(addr);
         tc->pcState(pc_state);
     } else {
-        inst->advancePC(pc_state);
-        tc->pcState(pc_state);
         invokeSE(tc, inst);
     }
 }
@@ -234,6 +232,12 @@ BreakpointFault::invokeSE(ThreadContext *tc, const StaticInstPtr &inst)
 void
 SyscallFault::invokeSE(ThreadContext *tc, const StaticInstPtr &inst)
 {
+    /* Advance the PC to next instruction so - once (simulated) syscall
+       is executed - execution continues. */
+    auto pc_state = tc->pcState().as<PCState>();
+    inst->advancePC(pc_state);
+    tc->pcState(pc_state);
+
     tc->getSystemPtr()->workload->syscall(tc);
 }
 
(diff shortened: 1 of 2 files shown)


shingarov (commented):
I believe it *should* be as @janvrany says, and empirically, it happens to be so on every physical RISC-V machine I have around here.

Still, there may be some reason why we are both wrong; @gabemblack what was your rationale for adding lines 167, 168?

janvrany:
Looking at failed test, I do not really understand what went wrong... Has anyone any idea?

powerjg:
I think the VM running the tests died. Can you rebase and force push to update the branch? This will kick off the tests again.

shingarov:
Well I did a CI re-run yesterday and it didn't make any difference.  So yeah, very strange.

powerjg:
Ah, it's hitting a timeout. I'm not totally sure why... I would try running the test that it failed on `timing-cpu_2-cores_no_cache_DualChannelDDR4_2400_arm_boot_test_to-tick-ALL-x86_64-opt`

@mkjost0, do you have any suggestions on diagnosing this?

janvrany:
@powerjg I did run 
```
./build/ALL/gem5.debug -d /tmp/gem5out532w5i9e -re --silent-redirect tests/gem5/arm_boot_tests/configs/arm_boot_exit_run.py --cpu timing --num-cpus 2 --mem-system no_cache --dram-class DualChannelDDR4_2400 --resource-directory tests/gem5/resources --tick-exit 10000000000
```
on my machine and it terminated successfully after ~20 secs.

powerjg:
Let's see if it works this time. Maybe we got unlucky and ran on a really slow CI core or something.

janvrany:
@powerjg I think it got stuck again. Is there a way I can run a single testcase 

timing-cpu_2-cores_no_cache_DualChannelDDR4_2400_arm_boot_test_to-tick-ALL-x86_64-opt

on my machine? (I realized that what I run above was problably the previous one)

mkjost0:
The smallest unit of test we can run is by SuiteUID, which you can run by using `./main.py run --skip-build --uid SuiteUID:tests/gem5/arm_boot_tests/test_linux_boot.py:timing-cpu_2-cores_no_cache_DualChannelDDR4_2400_arm_boot_test_to-tick-ALL-x86_64-opt -vvv` However, it looks like the command you ran above should accomplish the same thing as running this command.

janvrany:
@mkjost0 Thanks. I tried that but still, "works for me". 

I'm not sure what else to try locally, so let me just force-push one commit at time, starting with no-change commit, to see what happens.

janvrany:
OK, so it looks like commit 7e64e7a is causing the hangup. 

In theory that change may cause infinite loop (and therefore CI timeout) for RISC-V SE workloads if that workload depends on (*incorrect*) behavior of PC being increased *before* reporting trap. 

However:
 * regardless of PC value, `invokeSE()` is going to `panic()` so it should not matter and
 * the that that hangs is ARM, so change in RISC-V should not affect it. 

So, I do not understand that's going on here and out of ideas how to debug it.

janvrany:
Ah! It is not ARM test that hangs, it is indeed RISC-V test that hangs - only the CI output is truncated! Now I can debug it. 

Note to myself: this is what fails:
```
./build/ALL/gem5.opt --debug-flags=Exec ./tests/gem5/asmtest/configs/simple_binary_run.py rv64samt-ps-sysclone_d atomic riscv --num-cores 4 --resource-directory ./tests/gem5/resources
```

powerjg:
Maybe we should make this PR be just the power change. I wonder if the gdb issue on RISC-V may be a different problem/solultion.

src/arch/riscv/faults.cc
```diff
@@ -164,8 +164,6 @@ RiscvFault::invoke(ThreadContext *tc, const StaticInstPtr &inst)
         pc_state.set(addr);
         tc->pcState(pc_state);
     } else {
-        inst->advancePC(pc_state);
-        tc->pcState(pc_state);
```
rogerchang23424, line 168: I think move the code section to `SyscallFault::invokeSE(ThreadContext *tc, const StaticInstPtr &inst)` will pass to test CI

janvrany:
@rogerchang23424 Thanks, this is exactly what I did.

janvrany:
OK,  I did what @rogerchang23424 suggested and now all checks pass. I think this is ready for review now.

rogerchang23424 (approved):

BobbyRBruce (approved):

shingarov:
Wait, what?

> [janvrany](https://github.com/janvrany) added a commit to janvrany/MachineArithmetic that referenced this pull request 16 hours ago

What does this have to do with https://github.com/janvrany/MachineArithmetic/commit/81b31f18e1a94c26b8a0ccbd8ad7698586eaf788 ??

janvrany:
> 
> What does this have to do with [janvrany/MachineArithmetic@81b31f1](https://github.com/janvrany/MachineArithmetic/commit/81b31f18e1a94c26b8a0ccbd8ad7698586eaf788) ??

Ah, sorry! Indeed this is completely unrelated, I just accidentally pasted wrong URL into commit message of MA. Fixed now.

PR #167, opened 2023-08-08
tests: Add checkpoint tests for all ISAs

### LINKED ISSUE #154: Create checkpointing tests for different ISAs, caches (Ruby/Classic), SE/FS, multicore
Checkpointing in gem5 is easy to break during development. Only have four tests currently utilizes checkpointing: 

1. [An SE mode RISC-V checkpoint save, with classic cache, 1-core](https://github.com/gem5/gem5/blob/5eda9fe2ca8dd825223b7c79766aedcaad722cc9/tests/gem5/gem5_library_example_tests/test_gem5_library_examples.py#L58)
2. [An SE mode RISC-V checkpoint restore, with classic cache, 1-core](https://github.com/gem5/gem5/blob/5eda9fe2ca8dd825223b7c79766aedcaad722cc9/tests/gem5/gem5_library_example_tests/test_gem5_library_examples.py#L58) (restored with checkpoint from 1.
3. [A checkpoint save for SimPoints, 1-core](https://github.com/gem5/gem5/blob/5eda9fe2ca8dd825223b7c79766aedcaad722cc9/tests/gem5/gem5_library_example_tests/test_gem5_library_examples.py#L58)
4. [A checkpoint restore for Simpoints, 1-core](https://github.com/gem5/gem5/blob/5eda9fe2ca8dd825223b7c79766aedcaad722cc9/tests/gem5/gem5_library_example_tests/test_gem5_library_examples.py#L58)

The Simpoint checkpoints are somewhat tangential to this discussion and can be ignored for now. The SE Mode RISC-V checkpoint tests are more meaningful. While useful, they only test checkpointing on a small subset of possible simulations. Checkpointing bugs with different ISAs, cache types, or in FS mode will not be detected.

Therefore, we should create more tests to test checkpointing in different simulations I suggest the following set of tests to be added:

1. An SE mode ARM checkpoint save, with a Ruby cache, 2-core
2. A restore of 1.
3. An SE mode X86 checkpoint save, with a different classic cache, 4-core
4. A restore of 3.
5. A FS X86 checkpoint save,  with a Ruby cache, 4 core
6. A restore of 6.

I believe these 6 tests probably cover enough ground for us right now. As a bonus, i'd add simple SE mode tests for the remaining ISAs.

Please put all the configs somewhere in the "tests/gem5" directory. Though the current configs are in "configs/example/gem5_library/checkpoints", this is because these were written as example scripts, not tests scripts. What is proposed here are pure tests.

**Note No. 1**: Use the Simulators's `to_ticks` for the FS mode tests. There's no reason these tests should run for that many ticks before restoring a checkpoint. These tests should be fast

**Note No. 2**: As you can see in the current RISC-V and Simpoint checkpointing tests, the "restore" tests download the checkpoint from gem5 Resources. This is because our testing infrastructure can't chain tests or run them in a predictable order. So a "save" test can't be run before "restore" with the "restore" test using the save. This is an unfortunate limitation which we use gem5 Resources to get around.

Harshil2107:
Added save and restore checkpoint tests for arm-hello, x86-hello, x86-fs, power-hello 
Added mips and sparc test but mips does not support checkpoint and there is a bug in sparc. 
Added test file to run the tests.

Change-Id: I2d3b96f95ee08aae921de9a885ac5be77d49f326


tests/gem5/checkpoint-tests/arm-hello-restore-checkpoint.py
```diff
@@ -0,0 +1,80 @@
+# Copyright (c) 2022 The Regents of the University of California
```
BobbyRBruce: The copyright should be 2023. Update this for all relevant files

tests/gem5/checkpoint-tests/mips-hello-restore-checkpoint.py
```diff
@@ -0,0 +1,72 @@
...
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
+This gem5 configuation script creates a simple board sharing the same
+structure as the one in
+tests/gem5/checkpoint-tests/mips-hello-save-checkpoint.py.
```
BobbyRBruce: There's no way you could've known this, but all directories in the "tests" directory should be snakecase, e.g. `checkpoint_tests` -- all lower case with underscores as spaces. The python files should be with dashes `mips-hello-save-checkpoint.y` (so keep them as they are).

The reason for this is that python can't import paths containing dashes (.e.g., `import checkpoint-tests.script` will fail). So in a python project it's best the directories camel case. For the python files themselves, you use dashes for the executable script, and underscores in module names which are supposed to be imported. This is a neat way to avoid people importing scripts designed to be directly executed.

Melissa recently pushed a patch which renamed all the directories in tests to conform to this standard.

tests/gem5/checkpoint-tests/arm-hello-restore-checkpoint.py
BobbyRBruce, line 1: So as of #156 we're doing the following directory structure:

```txt
tests/
   gem5/
      checkpoint_tests
         README.md # A md file explaining the purpose of the tests
         configs # a directory containing all the configuration scripts
              arm-hello-restore-checkpoint-.py
              ...
         test-checkpoints.py
```

Please refactor this. Remember to update all the paths you used.

tests/gem5/checkpoint-tests/test-checkpoints.py
```diff
@@ -0,0 +1,266 @@
+# Copyright (c) 2021-2023 The Regents of the University of California
```
BobbyRBruce: Why is this 2023 to 2023?. This is a new file, no?

tests/gem5/checkpoint-tests/test-checkpoints.py
```diff
@@ -0,0 +1,266 @@
...
+        "checkpoint-tests",
+        "power-hello-restore-checkpoint.py",
+    ),
+    config_args=[],
+    valid_isas=(constants.all_compiled_tag,),
+    valid_hosts=constants.supported_hosts,
+    length=constants.quick_tag,
+)
+
+# gem5_verify_config(
```
BobbyRBruce: Add a link here to the issue you created/will create to not this bug (also link the issue back to this PR).

BobbyRBruce (changes requested):
Good work! I've got some things that need changed but it's all quite minor.

tests/gem5/checkpoint-tests/test-checkpoints.py
Harshil2107: Should I remove the mips and sparc test files?

BobbyRBruce: Why would you remove those?

Harshil2107: I thought of removing mips because it does not support checkpoints. I don't know if the support will be added.

BobbyRBruce: If mips genuinely doesn't support checkpoints then you're fine to remove these tests. I want tests there for stuff that is supposed to be supported but is broken.

powerjg: +1 to remove MIPS

Harshil2107: Just removed the mips test, I will add the link to the issue when I make one.

tests/gem5/checkpoint_tests/test-checkpoints.py
```diff
@@ -0,0 +1,278 @@
...
+        "configs",
+        "power-hello-restore-checkpoint.py",
+    ),
+    config_args=[],
+    valid_isas=(constants.all_compiled_tag,),
+    valid_hosts=constants.supported_hosts,
+    length=constants.quick_tag,
+)
+
+# gem5_verify_config(
```
BobbyRBruce: We've removed MIPS, so remove the call to the tests.

tests/gem5/checkpoint_tests/test-checkpoints.py
```diff
@@ -0,0 +1,278 @@
...
+#         "configs",
+#         "mips-hello-restore-checkpoint.py",
+#     ),
+#     config_args=[],
+#     valid_isas=(constants.all_compiled_tag,),
+#     valid_hosts=constants.supported_hosts,
+#     length=constants.quick_tag,
+# )
+
+# gem5_verify_config(
```
BobbyRBruce, line 204: Add an Issue link here to a GitHub Issue page explaining this bug.

BobbyRBruce (changes requested):

BobbyRBruce (approved):

PR #169, opened 2023-08-09
cpu: Fix segment fault when using debug flags Branch

rogerchang23424:
Change-Id: I36624b93f53aa101a57d51f3b917696cb2809136

### DIFF
--- a/src/cpu/pred/bpred_unit.cc
+++ b/src/cpu/pred/bpred_unit.cc
@@ -503,12 +503,11 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
                         "return [sn:%llu] PC: %#x Restoring RAS\n", tid,
                         squashed_sn,
                         hist_it->seqNum, hist_it->pc);
-                if (hist_it->RASTarget != nullptr) {
-                    DPRINTF(Branch,
-                            "[tid:%i] [squash sn:%llu] Restoring top of RAS "
-                            "to: %i, target: %s\n", tid, squashed_sn,
-                            hist_it->RASIndex, *hist_it->RASTarget.get());
-                }
+                 DPRINTF(Branch,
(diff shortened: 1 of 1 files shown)


src/cpu/pred/bpred_unit.cc
```diff
@@ -503,10 +503,12 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
...
-                DPRINTF(Branch,
-                        "[tid:%i] [squash sn:%llu] Restoring top of RAS "
-                        "to: %i, target: %s\n", tid, squashed_sn,
-                        hist_it->RASIndex, *hist_it->RASTarget);
+                if (hist_it->RASTarget != nullptr) {
+                    DPRINTF(Branch,
+                            "[tid:%i] [squash sn:%llu] Restoring top of RAS "
+                            "to: %i, target: %s\n", tid, squashed_sn,
+                            hist_it->RASIndex, *hist_it->RASTarget.get());
+                }
```
powerjg: Out of curiosity, why not something like this? Doesn't it make sense to print even if there is no `RASTarget`?

```suggestion
                 DPRINTF(Branch,
                        "[tid:%i] [squash sn:%llu] Restoring top of RAS "
                        "to: %i, target: %s\n", tid, squashed_sn,
                        hist_it->RASIndex, 
                        hist_it->RASIndex ? *hist_it->RASTarget.get() : "no RAS");
```

powerjg (commented):
One small question

src/cpu/pred/bpred_unit.cc
rogerchang23424: Thanks for your suggestion. I'm not familiar with the part for cpu branch prediction, any changes make --debug-flags=Branch will LGTM.

powerjg (approved):
That was much more complicated than I expected! Thanks for the updated.

PR #170, opened 2023-08-09
arch-riscv: Add checking CSR condition for RVV instructions

rogerchang23424:
If status.vs == OFF, the RVV instruction should raise Illegal Instruction according to RISC-V V spec. If RVV is not implemented, all of the RVV instruction need to raise exception.


powerjg (approved):
Thanks!

PR #171, opened 2023-08-09
arch-riscv: Add dynamic VLEN and ELEN configuration support to RVV path

### LINKED ISSUE #161: Adding VLEN and ELEN dynamic configuration to RISCV V 1.0 implementation
Hi!
I want to ask if someone is working on the dynamic configuration of ELEN and VLEN constant parameters in the RISCV Vector path. At this moment, if experimenting with different values of these parameters is needed, it is mandatory to recompile gem5 each time. 

I have been working on and testing a possible implementation during the last weeks. So, I could work together on this task if you agree with that.

Also, during my testing, I found some erroneous behaviours in specific instructions (i.e. vle instructions) when using a large enough VLEN value, generally above or equal to 2^14. This can be related to the microVl selected data type for implementing these instructions, although I am not 100% sure yet.

I am also tagging Adrià since he is supervising the effort we are carrying out with RISCV Vector in our group and he could be interested in reviewing this thread in the future. @adriaarmejach

alvaromorenomv:
At this moment, VLEN and ELEN RVV parameters are set as constants that need to be modified at compile time if you want to experiment with different values. With this patch, I want to set a first point to discuss how to configure these parameters dynamically.

Also, I have modified some data types that were provoking wrong behaviour in particular instructions when using a large enough VLEN value in the considered range inside the specification.

### DIFF
--- a/src/arch/riscv/RiscvISA.py
+++ b/src/arch/riscv/RiscvISA.py
@@ -38,9 +38,35 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-from m5.params import Enum
+from m5.params import Enum, UInt32
 from m5.params import Param
 from m5.objects.BaseISA import BaseISA
+
+
+class RiscvVectorLength(UInt32):
+    min = 8
+    max = 65536
+
+    def _check(self):
+        super()._check()
+
+        # VLEN needs to be a whole power of 2. We already know value is
+        # not zero. Hence:
+        if self.value & (self.value - 1) != 0:
+            raise TypeError("VLEN is not a power of 2: %d" % self.value)
+
+
+class RiscvVectorElementLength(UInt32):
+    min = 8
+    max = 64
+
+    def _check(self):
+        super()._check()
+
+        # ELEN needs to be a whole power of 2. We already know value is
+        # not zero. Hence:
+        if self.value & (self.value - 1) != 0:
+            raise TypeError("ELEN is not a power of 2: %d" % self.value)
 
 
 class RiscvType(Enum):
@@ -58,3 +84,13 @@ class RiscvISA(BaseISA):
     riscv_type = Param.RiscvType("RV64", "RV32 or RV64")
 
     enable_rvv = Param.Bool(True, "Enable vector extension")
+    vlen = Param.RiscvVectorLength(
+        256,
+        "Length of each vector register in bits. \
+        VLEN in Ch. 2 of RISC-V vector spec",
+    )
+    elen = Param.RiscvVectorElementLength(
+        64,
+        "Length of each vector element in bits. \
+        ELEN in Ch. 2 of RISC-V vector spec",
+    )
--- a/src/arch/riscv/insts/vector.cc
+++ b/src/arch/riscv/insts/vector.cc
@@ -215,8 +215,9 @@ std::string VleMicroInst::generateDisassembly(Addr pc,
         const loader::SymbolTable *symtab) const
 {
     std::stringstream ss;
+    unsigned vlenb = vlen >> 3;
     ss << mnemonic << ' ' << registerName(destRegIdx(0)) << ", "
-       << VLENB * microIdx << '(' << registerName(srcRegIdx(0)) << ')' << ", "
+       << vlenb * microIdx << '(' << registerName(srcRegIdx(0)) << ')' << ", "
        << registerName(srcRegIdx(1));
     if (!machInst.vm) ss << ", v0.t";
     return ss.str();
@@ -226,8 +227,9 @@ std::string VlWholeMicroInst::generateDisassembly(Addr pc,
         const loader::SymbolTable *symtab) const
 {
     std::stringstream ss;
+    unsigned vlenb = vlen >> 3;
     ss << mnemonic << ' ' << registerName(destRegIdx(0)) << ", "
-       << VLENB * microIdx << '(' << registerName(srcRegIdx(0)) << ')';
+       << vlenb * microIdx << '(' << registerName(srcRegIdx(0)) << ')';
     return ss.str();
 }
 
@@ -235,8 +237,9 @@ std::string VseMicroInst::generateDisassembly(Addr pc,
         const loader::SymbolTable *symtab) const
 {
     std::stringstream ss;
+    unsigned vlenb = vlen >> 3;
     ss << mnemonic << ' ' << registerName(srcRegIdx(1)) << ", "
-       << VLENB * microIdx  << '(' << registerName(srcRegIdx(0)) << ')';
+       << vlenb * microIdx  << '(' << registerName(srcRegIdx(0)) << ')';
     if (!machInst.vm) ss << ", v0.t";
     return ss.str();
 }
@@ -245,8 +248,9 @@ std::string VsWholeMicroInst::generateDisassembly(Addr pc,
         const loader::SymbolTable *symtab) const
 {
     std::stringstream ss;
+    unsigned vlenb = vlen >> 3;
     ss << mnemonic << ' ' << registerName(srcRegIdx(1)) << ", "
-       << VLENB * microIdx << '(' << registerName(srcRegIdx(0)) << ')';
+       << vlenb * microIdx << '(' << registerName(srcRegIdx(0)) << ')';
     return ss.str();
 }
 
--- a/src/arch/riscv/insts/vector.hh
+++ b/src/arch/riscv/insts/vector.hh
@@ -32,6 +32,7 @@
 #include <string>
 
 #include "arch/riscv/insts/static_inst.hh"
+#include "arch/riscv/isa.hh"
 #include "arch/riscv/regs/misc.hh"
 #include "arch/riscv/regs/vector.hh"
 #include "arch/riscv/utility.hh"
@@ -116,11 +117,14 @@ class VectorMacroInst : public RiscvMacroInst
   protected:
     uint32_t vl;
     uint8_t vtype;
+    uint32_t vlen;
+
     VectorMacroInst(const char* mnem, ExtMachInst _machInst,
-                   OpClass __opClass)
+                   OpClass __opClass, uint32_t _vlen = 256)
         : RiscvMacroInst(mnem, _machInst, __opClass),
         vl(_machInst.vl),
-        vtype(_machInst.vtype8)
+        vtype(_machInst.vtype8),
+        vlen(_vlen)
     {
         this->flags[IsVector] = true;
     }
@@ -128,13 +132,15 @@ class VectorMacroInst : public RiscvMacroInst
 
 class VectorMicroInst : public RiscvMicroInst
 {
-  protected:
-    uint8_t microVl;
+protected:
+    uint32_t vlen;
+    uint32_t microVl;
     uint8_t microIdx;
     uint8_t vtype;
     VectorMicroInst(const char *mnem, ExtMachInst _machInst, OpClass __opClass,
-                    uint8_t _microVl, uint8_t _microIdx)
+      uint32_t _microVl, uint8_t _microIdx, uint32_t _vlen = 256)
         : RiscvMicroInst(mnem, _machInst, __opClass),
+        vlen(_vlen),
         microVl(_microVl),
         microIdx(_microIdx),
         vtype(_machInst.vtype8)
@@ -169,7 +175,7 @@ class VectorArithMicroInst : public VectorMicroInst
 {
 protected:
     VectorArithMicroInst(const char *mnem, ExtMachInst _machInst,
-                         OpClass __opClass, uint8_t _microVl,
+                         OpClass __opClass, uint32_t _microVl,
                          uint8_t _microIdx)
         : VectorMicroInst(mnem, _machInst, __opClass, _microVl, _microIdx)
     {}
@@ -182,12 +188,11 @@ class VectorArithMacroInst : public VectorMacroInst
 {
   protected:
     VectorArithMacroInst(const char* mnem, ExtMachInst _machInst,
-                         OpClass __opClass)
-        : VectorMacroInst(mnem, _machInst, __opClass)
+                         OpClass __opClass, uint32_t _vlen = 256)
+        : VectorMacroInst(mnem, _machInst, __opClass, _vlen)
     {
         this->flags[IsVector] = true;
     }
-
     std::string generateDisassembly(
             Addr pc, const loader::SymbolTable *symtab) const override;
 };
@@ -196,7 +201,7 @@ class VectorVMUNARY0MicroInst : public VectorMicroInst
 {
 protected:
     VectorVMUNARY0MicroInst(const char *mnem, ExtMachInst _machInst,
-                         OpClass __opClass, uint8_t _microVl,
+                         OpClass __opClass, uint32_t _microVl,
                          uint8_t _microIdx)
         : VectorMicroInst(mnem, _machInst, __opClass, _microVl, _microIdx)
     {}
@@ -209,8 +214,8 @@ class VectorVMUNARY0MacroInst : public VectorMacroInst
 {
   protected:
     VectorVMUNARY0MacroInst(const char* mnem, ExtMachInst _machInst,
-                         OpClass __opClass)
-        : VectorMacroInst(mnem, _machInst, __opClass)
+                         OpClass __opClass, uint32_t _vlen)
+        : VectorMacroInst(mnem, _machInst, __opClass, _vlen)
     {
         this->flags[IsVector] = true;
     }
@@ -223,8 +228,8 @@ class VectorSlideMacroInst : public VectorMacroInst
 {
   protected:
     VectorSlideMacroInst(const char* mnem, ExtMachInst _machInst,
-                         OpClass __opClass)
-        : VectorMacroInst(mnem, _machInst, __opClass)
+                         OpClass __opClass, uint32_t _vlen = 256)
+        : VectorMacroInst(mnem, _machInst, __opClass, _vlen)
     {
         this->flags[IsVector] = true;
     }
@@ -239,7 +244,7 @@ class VectorSlideMicroInst : public VectorMicroInst
     uint8_t vdIdx;
     uint8_t vs2Idx;
     VectorSlideMicroInst(const char *mnem, ExtMachInst _machInst,
-                         OpClass __opClass, uint8_t _microVl,
+                         OpClass __opClass, uint32_t _microVl,
                          uint8_t _microIdx, uint8_t _vdIdx, uint8_t _vs2Idx)
         : VectorMicroInst(mnem, _machInst, __opClass, _microVl, _microIdx)
         , vdIdx(_vdIdx), vs2Idx(_vs2Idx)
@@ -256,7 +261,7 @@ class VectorMemMicroInst : public VectorMicroInst
     Request::Flags memAccessFlags;
 
     VectorMemMicroInst(const char* mnem, ExtMachInst _machInst,
-                       OpClass __opClass, uint8_t _microVl, uint8_t _microIdx,
+                       OpClass __opClass, uint32_t _microVl, uint8_t _microIdx,
                        uint32_t _offset)
         : VectorMicroInst(mnem, _machInst, __opClass, _microVl, _microIdx)
         , offset(_offset)
@@ -268,8 +273,8 @@ class VectorMemMacroInst : public VectorMacroInst
 {
   protected:
     VectorMemMacroInst(const char* mnem, ExtMachInst _machInst,
-                       OpClass __opClass)
-        : VectorMacroInst(mnem, _machInst, __opClass)
+                        OpClass __opClass, uint32_t _vlen = 256)
+        : VectorMacroInst(mnem, _machInst, __opClass, _vlen)
     {}
 };
 
@@ -277,8 +282,8 @@ class VleMacroInst : public VectorMemMacroInst
 {
   protected:
     VleMacroInst(const char* mnem, ExtMachInst _machInst,
-                   OpClass __opClass)
-        : VectorMemMacroInst(mnem, _machInst, __opClass)
+                   OpClass __opClass, uint32_t _vlen)
+        : VectorMemMacroInst(mnem, _machInst, __opClass, _vlen)
     {}
 
     std::string generateDisassembly(
@@ -289,8 +294,8 @@ class VseMacroInst : public VectorMemMacroInst
 {
   protected:
     VseMacroInst(const char* mnem, ExtMachInst _machInst,
-                   OpClass __opClass)
-        : VectorMemMacroInst(mnem, _machInst, __opClass)
+                   OpClass __opClass, uint32_t _vlen)
+        : VectorMemMacroInst(mnem, _machInst, __opClass, _vlen)
     {}
 
     std::string generateDisassembly(
@@ -302,9 +307,10 @@ class VleMicroInst : public VectorMicroInst
   protected:
     Request::Flags memAccessFlags;
 
-    VleMicroInst(const char *mnem, ExtMachInst _machInst, OpClass __opClass,
-                 uint8_t _microVl, uint8_t _microIdx)
-        : VectorMicroInst(mnem, _machInst, __opClass, _microVl, _microIdx)
+    VleMicroInst(const char *mnem, ExtMachInst _machInst,OpClass __opClass,
+                  uint32_t _microVl, uint8_t _microIdx, uint32_t _vlen)
+        : VectorMicroInst(mnem, _machInst, __opClass, _microVl,
+                            _microIdx, _vlen)
     {
         this->flags[IsLoad] = true;
     }
@@ -319,8 +325,9 @@ class VseMicroInst : public VectorMicroInst
     Request::Flags memAccessFlags;
 
     VseMicroInst(const char *mnem, ExtMachInst _machInst, OpClass __opClass,
-                 uint8_t _microVl, uint8_t _microIdx)
-        : VectorMicroInst(mnem, _machInst, __opClass, _microVl, _microIdx)
+                  uint32_t _microVl, uint8_t _microIdx, uint32_t _vlen)
+        : VectorMicroInst(mnem, _machInst, __opClass, _microVl,
+                            _microIdx, _vlen)
     {
         this->flags[IsStore] = true;
     }
@@ -333,8 +340,8 @@ class VlWholeMacroInst : public VectorMemMacroInst
 {
   protected:
     VlWholeMacroInst(const char *mnem, ExtMachInst _machInst,
-                     OpClass __opClass)
-      : VectorMemMacroInst(mnem, _machInst, __opClass)
+                     OpClass __opClass, uint32_t _vlen)
+      : VectorMemMacroInst(mnem, _machInst, __opClass, _vlen)
     {}
 
     std::string generateDisassembly(
@@ -347,8 +354,10 @@ class VlWholeMicroInst : public VectorMicroInst
     Request::Flags memAccessFlags;
 
     VlWholeMicroInst(const char *mnem, ExtMachInst _machInst,
-                     OpClass __opClass, uint8_t _microVl, uint8_t _microIdx)
-        : VectorMicroInst(mnem, _machInst, __opClass, _microVl, _microIdx)
+          OpClass __opClass, uint32_t _microVl, uint8_t _microIdx,
+          uint32_t _vlen)
+        : VectorMicroInst(mnem, _machInst, __opClass, _microVl,
+                            _microIdx, _vlen)
     {}
 
     std::string generateDisassembly(
@@ -359,8 +368,8 @@ class VsWholeMacroInst : public VectorMemMacroInst
 {
   protected:
     VsWholeMacroInst(const char *mnem, ExtMachInst _machInst,
-                     OpClass __opClass)
-        : VectorMemMacroInst(mnem, _machInst, __opClass)
+                     OpClass __opClass, uint32_t _vlen)
+        : VectorMemMacroInst(mnem, _machInst, __opClass, _vlen)
     {}
 
     std::string generateDisassembly(
@@ -373,8 +382,10 @@ class VsWholeMicroInst : public VectorMicroInst
     Request::Flags memAccessFlags;
 
     VsWholeMicroInst(const char *mnem, ExtMachInst _machInst,
-                     OpClass __opClass, uint8_t _microVl, uint8_t _microIdx)
-        : VectorMicroInst(mnem, _machInst, __opClass, _microIdx, _microIdx)
+                      OpClass __opClass, uint32_t _microVl,
+                      uint8_t _microIdx, uint32_t _vlen)
+        : VectorMicroInst(mnem, _machInst, __opClass , _microVl,
+                          _microIdx, _vlen)
     {}
 
     std::string generateDisassembly(
@@ -385,8 +396,8 @@ class VlStrideMacroInst : public VectorMemMacroInst
 {
   protected:
     VlStrideMacroInst(const char* mnem, ExtMachInst _machInst,
-                   OpClass __opClass)
-        : VectorMemMacroInst(mnem, _machInst, __opClass)
+                   OpClass __opClass, uint32_t _vlen)
+        : VectorMemMacroInst(mnem, _machInst, __opClass, _vlen)
     {}
 
     std::string generateDisassembly(
@@ -399,7 +410,7 @@ class VlStrideMicroInst : public VectorMemMicroInst
   uint8_t regIdx;
     VlStrideMicroInst(const char *mnem, ExtMachInst _machInst,
                       OpClass __opClass, uint8_t _regIdx,
-                      uint8_t _microIdx, uint8_t _microVl)
+                      uint8_t _microIdx, uint32_t _microVl)
         : VectorMemMicroInst(mnem, _machInst, __opClass, _microVl,
                              _microIdx, 0)
         , regIdx(_regIdx)
@@ -413,8 +424,8 @@ class VsStrideMacroInst : public VectorMemMacroInst
 {
   protected:
     VsStrideMacroInst(const char* mnem, ExtMachInst _machInst,
-                   OpClass __opClass)
-        : VectorMemMacroInst(mnem, _machInst, __opClass)
+                   OpClass __opClass, uint32_t _vlen)
+        : VectorMemMacroInst(mnem, _machInst, __opClass, _vlen)
     {}
 
     std::string generateDisassembly(
@@ -427,7 +438,7 @@ class VsStrideMicroInst : public VectorMemMicroInst
     uint8_t regIdx;
     VsStrideMicroInst(const char *mnem, ExtMachInst _machInst,
                       OpClass __opClass, uint8_t _regIdx,
-                      uint8_t _microIdx, uint8_t _microVl)
+                      uint8_t _microIdx, uint32_t _microVl)
         : VectorMemMicroInst(mnem, _machInst, __opClass, _microVl,
                              _microIdx, 0)
         , regIdx(_regIdx)
@@ -441,8 +452,8 @@ class VlIndexMacroInst : public VectorMemMacroInst
 {
   protected:
     VlIndexMacroInst(const char* mnem, ExtMachInst _machInst,
-                   OpClass __opClass)
-        : VectorMemMacroInst(mnem, _machInst, __opClass)
+                   OpClass __opClass, uint32_t _vlen)
+        : VectorMemMacroInst(mnem, _machInst, __opClass, _vlen)
     {}
 
     std::string generateDisassembly(
@@ -473,8 +484,8 @@ class VsIndexMacroInst : public VectorMemMacroInst
 {
   protected:
     VsIndexMacroInst(const char* mnem, ExtMachInst _machInst,
-                   OpClass __opClass)
-        : VectorMemMacroInst(mnem, _machInst, __opClass)
+                   OpClass __opClass, uint32_t _vlen)
+        : VectorMemMacroInst(mnem, _machInst, __opClass, _vlen)
     {}
 
     std::string generateDisassembly(
@@ -516,7 +527,7 @@ class VMvWholeMicroInst : public VectorArithMicroInst
 {
   protected:
     VMvWholeMicroInst(const char *mnem, ExtMachInst _machInst,
-                         OpClass __opClass, uint8_t _microVl,
+                         OpClass __opClass, uint32_t _microVl,
                          uint8_t _microIdx)
         : VectorArithMicroInst(mnem, _machInst, __opClass, _microVl, _microIdx)
     {}
@@ -533,10 +544,12 @@ class VMaskMergeMicroInst : public VectorArithMicroInst
     RegId destRegIdxArr[1];
 
   public:
-    VMaskMergeMicroInst(ExtMachInst extMachInst, uint8_t _dstReg,
-        uint8_t _numSrcs)
+    uint32_t vlen;
+    VMaskMergeMicroInst(ExtMachInst extMachInst,
+        uint8_t _dstReg, uint8_t _numSrcs, uint32_t _vlen)
         : VectorArithMicroInst("vmask_mv_micro", extMachInst,
-                               VectorIntegerArithOp, 0, 0)
+                               VectorIntegerArithOp, 0, 0),
+          vlen(_vlen)
     {
         setRegIdxArrays(
             reinterpret_cast<RegIdArrayPtr>(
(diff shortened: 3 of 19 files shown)


alvaromorenomv:
Mentioning Adrià in case he wants to review this in the future @adriaarmejach

src/arch/riscv/RiscvSystem.py
```diff
@@ -0,0 +1,79 @@
...
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from m5.params import *
+from m5.options import *
+from m5.SimObject import *
+from m5.util.fdthelper import *
```
powerjg: I don't think this is needed.

src/arch/riscv/RiscvSystem.py
```diff
@@ -0,0 +1,79 @@
...
+
+from m5.objects.System import System
```
powerjg: This shouldn't be needed either.

src/arch/riscv/RiscvSystem.py
```diff
@@ -0,0 +1,79 @@
...
+        if self.value & (self.value - 1) != 0:
+            raise TypeError("RISCV ELEN is not a power of 2: %d" % self.value)
+
+
+class RiscvSystem(System):
+    type = "RiscvSystem"
+    cxx_header = "arch/riscv/system.hh"
+    cxx_class = "gem5::RiscvSystem"
+
+    riscv_vlen = Param.RiscvVectorLength(256, "RISC-V maximum vector length")
```
powerjg: These should probably be parameters of the ISA object, not a new system object.

src/arch/riscv/isa/decoder.isa
```diff
@@ -3070,9 +3070,11 @@ decode QUADRANT default Unknown::unknown() {
                         xc->getRegOperand(this, 2,
                             &tmp_s2);
                         auto Vs2bit = tmp_s2.as<vu>();
+                        uint32_t vlen = RiscvISA::ISA::getVecLenInBits();
```
powerjg: This should use the ISA pointer via `xc` (I think). I.e., it's theoretically possible to have cores with different length vectors.

src/arch/riscv/isa/formats/vector_mem.isa
```diff
@@ -90,7 +90,10 @@ def VMemBase(name, Name, ea_code, memacc_code, mem_flags,
 
 def format VleOp(
     memacc_code,
-    ea_code={{ EA = Rs1 + VLENB * microIdx; }},
+    ea_code={{
+        uint32_t vlenb = RiscvISA::ISA::getVecLenInBytes();
```
powerjg: Same as above, this function should not be static.

src/arch/riscv/system.hh
```diff
@@ -0,0 +1,81 @@
...
+#define __ARCH_RISCV_SYSTEM_HH__
+
+#include "params/RiscvSystem.hh"
+#include "sim/full_system.hh"
+#include "sim/sim_object.hh"
+#include "sim/system.hh"
+
+namespace gem5
+{
+class RiscvSystem : public System
```
powerjg: As mentioned above, I think this code should go in the ISA class instead of in a new class.

powerjg (commented):
Thanks for the contribution! Overall, this looks great. There's a couple of things that I don't understand (yet).

1. Why is a new `RiscvSystem` class required? (I think you can instead use the `RiscvISA` object instead.)
2. Why is the "get length" function static instead of accessed via the ISA pointer?

BTW, I think the `uint8_t` to `uint16_t` change fixes an issue that someone has brought up, but I can't find a link right now.

alvaromorenomv:
The two points that you commented on are the ones that gave me some doubts. I am going to share my insight about them:

1. At first glance, I found it more intuitive to configure the parameters via a system object at the time of instantiation of the whole system, as in the example I provided with the RiscvBoard. Although, there should be no problem with doing it directly on the ISA if it makes more sense. I am going to take a look at it.
2. I was not really conformed with using static methods because, as you commented, it is specified that Elen and Vlen should be configured per core. I agree that accessing these variables using the ISA pointer should be the way. 
But my mind is fogged with them being used in instruction methods like generateDissasembly and constructor of the classes where I need help seeing clearly how to reach an ISA object. Therefore, I would like to discuss/ask for some guidance on this implementation.

src/arch/riscv/insts/vector.cc
```diff
@@ -215,8 +215,9 @@ std::string VleMicroInst::generateDisassembly(Addr pc,
         const loader::SymbolTable *symtab) const
 {
     std::stringstream ss;
+    unsigned vlenb = RiscvISA::ISA::getVecLenInBytes();
```
alvaromorenomv: This is the concrete point that is causing me problems.

powerjg:
> The two points that you commented on are the ones that gave me some doubts. I am going to share my insight about them:
> 
> 1. At first glance, I found it more intuitive to configure the parameters via a system object at the time of instantiation of the whole system, as in the example I provided with the RiscvBoard. Although, there should be no problem with doing it directly on the ISA if it makes more sense. I am going to take a look at it.

Sounds good. 

> 2. I was not really conformed with using static methods because, as you commented, it is specified that Elen and Vlen should be configured per core. I agree that accessing these variables using the ISA pointer should be the way.
>    But my mind is fogged with them being used in instruction methods like generateDissasembly and constructor of the classes where I need help seeing clearly how to reach an ISA object. Therefore, I would like to discuss/ask for some guidance on this implementation.

Great question... it took me a bit to track it down as well. Here's what I think is similar code in Arm: 
https://github.com/gem5/gem5/blob/develop/src/arch/arm/insts/static_inst.cc#L1357

Essentially, it will look something like the code below:
```cpp
ThreadContext* tc = xc->tcBase();
auto isa = static_cast<RiscvISA::ISA *>(tc->getIsaPtr());
```

It may not be straightforward with the current interfaces to get this information in `generateDisassembly`. If you really need that info there, we may have to update the interface to make it possible to get dynamic information into those functions.

Let me know if you have any other questions!

alvaromorenomv:
Thanks for the help! 
I will change the implementation following your suggestions and let you know if I find any more problems.

BobbyRBruce:
> Thanks for the help!
> I will change the implementation following your suggestions and let you know if I find any more problems.

Hey @alvaromorenomv, do you need any help implementing this?

alvaromorenomv:
> > Thanks for the help!
> > I will change the implementation following your suggestions and let you know if I find any more problems.
> 
> Hey @alvaromorenomv, do you need any help implementing this?

Hi!
I had to take some days off.
Probably next week I will be able to push the update I was working on. Otherwise I will let you know.

alvaromorenomv:
> > > Thanks for the help!
> > > I will change the implementation following your suggestions and let you know if I find any more problems.
> > 
> > 
> > Hey @alvaromorenomv, do you need any help implementing this?
> 
> Hi! I had to take some days off. Probably next week I will be able to push the update I was working on. Otherwise I will let you know.

I have been discussing along with @adriaarmejach the problem of accessing vlen and vlenb through ISA pointer, specially on Instruction Constructors and getDissasembly methods where tc and  xc  are not accessible. As I commented in a previous post, the problem is that in these cases, getting the ISA object information is not straightforward.
We have thought of some possible solutions that I am listing below:
- Ideally we would like to include the vlen information in the Extended Machine Instruction union. This would be the best option and it makes sense as the EMI is basically the only information instruction constructors have and the vlen needs to be known. The problem is that there are only 3 free bits in the EMI union at the moment and a minimum of 4 bits would be required to codify vlen information. So, it would be necessary to extend the size of the EMI union beyond 64 bits: is there an easy way to do this? right now all unions are <=64bits.
- Another option we have considered is to include vlen as a member of RiscvStaticInst, but again we have similar problems as it is hard to access ISA object information.

Let us know your thoughts on this.

rogerchang23424:
> Ideally we would like to include the vlen information in the Extended Machine Instruction union. This would be the best option and it makes sense as the EMI is basically the only information instruction constructors have and the vlen needs to be known. The problem is that there are only 3 free bits in the EMI union at the moment and a minimum of 4 bits would be required to codify vlen information. So, it would be necessary to extend the size of the EMI union beyond 64 bits: is there an easy way to do this? right now all unions are <=64bits.

There are two methods to approach it:

1. You may need to change the type of `ExtMachInst` from `BitUnion64` to `struct`. Then, the [`bitfields.isa`](https://github.com/gem5/gem5/blob/stable/src/arch/riscv/isa/bitfields.isa) definitions need to be changed to the struct fields of name instead the bitfields of fields. Some of instruction implementations may need to update after the fields of `bitfields.isa` changed to name presentation.

2. Add the new `RiscvStaticInst` constructor with `ExtMachInst` and `vlen` parameter. The methods need to replace all of `decode_block` for RVV instructions. The vlen parameters can be get from `isa` parameter.

The both of methods are need to add `vlen` in the `RiscvStaticInst` and I suggest to do that.

src/arch/riscv/RiscvISA.py
```diff
@@ -52,9 +55,19 @@ class RiscvISA(BaseISA):
...
+
     check_alignment = Param.Bool(
         True, "whether to check memory access alignment"
     )
+
     riscv_type = Param.RiscvType("RV64", "RV32 or RV64")
 
     enable_rvv = Param.Bool(True, "Enable vector extension")
+
+    riscv_vlen_se = Param.RiscvVectorLength(
```
powerjg: I'm a bit confused on the variable name. I would have simply called it `vlen` and put in the description "Length of each vector register in bits. VLEN in Ch. 2 of RISC-V vector spec"

I know it's a bit nit-picky, but we're going to be stuck with these names forever :)

src/arch/riscv/RiscvISA.py
```diff
@@ -52,9 +55,19 @@ class RiscvISA(BaseISA):
...
+        256, "RISC-V maximum vector length"
+    )
+    riscv_elen_se = Param.RiscvVectorElementLength(
```
powerjg: Same thing as above, but for ELEN

src/arch/riscv/RiscvSystem.py
```diff
@@ -0,0 +1,79 @@
+# Copyright (c) Barcelona Supercomputing Center 2023
```
powerjg: Can we remove this file?

src/arch/riscv/isa.cc
```diff
@@ -266,6 +265,16 @@ ISA::ISA(const Params &p) :
     _regClasses.push_back(&ccRegClass);
     _regClasses.push_back(&miscRegClass);
 
+    system = dynamic_cast<RiscvSystem *>(p.system);
+
+    if (FullSystem && system){
```
powerjg: I still don't understand the need for `system` here. I think it should be removed.

Another way to look at this... these parameters `VLEN` and `ELEN` are specific to a *hart*. In gem5, the ISA object is the place where we specify the hart parameters.

src/arch/riscv/system.hh
```diff
@@ -0,0 +1,81 @@
+/*
```
powerjg: I think these files can be dropped as well.

powerjg (commented):
Looking better... we're getting there!

As far as `generateDisassembly`... maybe we shouldn't try to put the vector length there. Instead, we should print the vector length with the PC whenever we're executing vector instructions. Thoughts?

alvaromorenomv:
> Looking better... we're getting there!
> 
> As far as `generateDisassembly`... maybe we shouldn't try to put the vector length there. Instead, we should print the vector length with the PC whenever we're executing vector instructions. Thoughts?

I totally agree with your comments.
I have applied the changes in variable names and descriptions and deleted the system files and their references.
I am waiting to decide what to do with Macro Instruction constructors and generate Dissasembly to push the update.

I was trying to apply method 2, suggested by Roger, of propagating vlen to the Macro Instruction constructors that need it, so rebuilding the entire EMI definition is unnecessary. 

For generateDissasembly, vlen information is only being used in a few MicroInsts, so we could either propagate it from the MacroInst constructor or apply the idea of printing it when executing the instruction if it looks better.

powerjg:
`generateDisassembly` is only used for debugging. So, the question is, how much effort is it worth to get the "best" debugging?

If you were to get disassembly from a binary, it would *not* have the vector length encoded with each instruction. It can't because the vector length isn't encoded in the static instruction. 

As long as we can find the *current* vector length when an instruction is being executed for debugging purposes, I don't think it really matters what the implementation is. If it's possible to get it into the `ExtMachInst`, that would be fine, but if it's just in the `PcState`, I think that's OK too. Maybe we need to modify how the `PcState` object is printed?

src/arch/riscv/RiscvISA.py
```diff
@@ -52,9 +55,19 @@ class RiscvISA(BaseISA):
...
     check_alignment = Param.Bool(
         True, "whether to check memory access alignment"
     )
+
     riscv_type = Param.RiscvType("RV64", "RV32 or RV64")
 
     enable_rvv = Param.Bool(True, "Enable vector extension")
+
+    riscv_vlen_se = Param.RiscvVectorLength(
+        256, "RISC-V maximum vector length"
```
jerryzj: Should be 4096, SiFive already implemented x280 which has VLEN=512b, and Andes also has a core with VLEN=1024b

Ref: https://github.com/riscv/riscv-v-spec/issues/204

powerjg: Could you expand on why you think we should make this change? This is a default for a parameter, not something that is set in stone.

Maybe the description could be updated to say something like "the hardware length of registers, VLEN in the spec"

hnpl: @jerryzj I want to clarify that this parameter supposedly define an architecture's VLEN, not the maximum VLEN. The VLEN parameter can be changed by user.

jerryzj: @hnpl got it, I misunderstood the meaning of the variable here, thanks for the clarification.
However, I would also suggest that the default VLEN could be set to 128b as it is the default length of V extension.

alvaromorenomv:
> `generateDisassembly` is only used for debugging. So, the question is, how much effort is it worth to get the "best" debugging?
> 
> If you were to get disassembly from a binary, it would _not_ have the vector length encoded with each instruction. It can't because the vector length isn't encoded in the static instruction.
> 
> As long as we can find the _current_ vector length when an instruction is being executed for debugging purposes, I don't think it really matters what the implementation is. If it's possible to get it into the `ExtMachInst`, that would be fine, but if it's just in the `PcState`, I think that's OK too. Maybe we need to modify how the `PcState` object is printed?

Sorry for taking a bit too much time to answer. A lot of things are going on :sweat_smile:.
I have pushed an update after rebasing on top of the PcState patch from @rogerchang23424  and also including vlen as a member of Vector (RiscvStatic) instructions that need it to be split in microInsts during their construction.
 
About the generateDissasembly, I was not getting the point that it is just for debugging purposes, so sorry again.
In the update, I passed vlen to the MicroInsts through their constructor. 
I think it is not too much effort since I already needed that info in their associated MacroInstruction that creates them. 
Printing it through the PcState during execution also makes sense to me, so we don't need to modify the MicroInst constructors, although I should look deeper into how to implement it.

rogerchang23424 (commented):
Most of part are LGTM. I'm thinking if we can replace all of the `pc_ptr->as<PCState>().vlenb()` with  `VlenbBits` operand in the execute section of `*.isa` files. I think the `PCState.vlenb()` should be in `%(op_rd)s` and PCState should be declared in `%(op_desc)s` if VlenBits is exists in code.

https://github.com/gem5/gem5/blob/cfa13f9feb905cfd9853397dfc23239e1121e175/src/arch/isa_parser/operand_types.py#L592

alvaromorenomv:
> Most of part are LGTM. I'm thinking if we can replace all of the `pc_ptr->as<PCState>().vlenb()` with `VlenbBits` operand in the execute section of `*.isa` files. I think the `PCState.vlenb()` should be in `%(op_rd)s` and PCState should be declared in `%(op_desc)s` if VlenBits is exists in code.
> 
> https://github.com/gem5/gem5/blob/cfa13f9feb905cfd9853397dfc23239e1121e175/src/arch/isa_parser/operand_types.py#L592

I tried to do it through two new format methods in order to not replicate too much code in %(op_rd)s, let me know if it makes sense for you or if there is some problem that I did not notice with this implementation.

Thanks in advance!

src/arch/riscv/types.hh
```diff
@@ -178,6 +177,14 @@ BitUnion64(ExtMachInst)
 
 EndBitUnion(ExtMachInst)
 
+constexpr unsigned MaxVecLenInBits = 65536;
+static_assert(  MaxVecLenInBits >= 8 &&
+                MaxVecLenInBits <= 65536 &&
+                (MaxVecLenInBits & (MaxVecLenInBits - 1)) == 0,
+                "Unsupported max. VLEN");
```
rogerchang23424: Any reason to check this? The `MaxVecLenInBits` should not be changed in the program.

src/arch/riscv/insts/vector.hh
```diff
@@ -114,24 +118,47 @@ class VectorMacroInst : public RiscvMacroInst
...
+
+    VectorMacroInst(const char* mnem, ExtMachInst _machInst,
+                   uint32_t _vlen, OpClass __opClass)
+        : RiscvMacroInst(mnem, _machInst, __opClass),
+        vl(_machInst.vl),
+        vtype(_machInst.vtype8),
+        vlen(_vlen)
+    {
+        this->flags[IsVector] = true;
+    }
```
rogerchang23424: This part is OK to me, but I think we can just leave one constructor with default `vlen`.

rogerchang23424 (commented):
The methods of dynamic vlen is LGTM.

To approve the PR. I would like to test the RVV with different vlen and elen to verify the implementation.

src/arch/riscv/types.hh
alvaromorenomv: This check also does not make sense to me. When I started to implement the feature I was following the ARM SVE implementation and it should be a piece of code that I missed deleting after going deeper. I removed it in the new update

src/arch/riscv/insts/vector.hh
alvaromorenomv: I have changed duplicated constructors by default vlen values in the last updated. I agree that it looks better in this concrete case.

alvaromorenomv:
> The methods of dynamic vlen is LGTM.
> 
> To approve the PR. I would like to test the RVV with different vlen and elen to verify the implementation.

I have pushed an update following your comments and a few minor recommendations from @adriaarmejach.
We have been testing the patch with our vectorised benchmarks suite, including instruction tests and standard HPC-like programs. This testing has been done mainly in our FS Riscv infrastructure, which looks to perform well. However, a bit more verification would be great. 
Feel free to test it!

src/arch/riscv/RiscvISA.py
```diff
@@ -38,11 +38,37 @@
...
+class RiscvVectorLength(UInt32):
+    min = 8
+    max = 65536
+
+    def _check(self):
+        super()._check()
+
+        # VLEN needs to be a whole power of 2. We already know value is
+        # not zero. Hence:
+        if self.value & (self.value - 1) != 0:
```
powerjg, line 55: Do you want to check for `min <= self.value <= max`?

src/arch/riscv/isa.cc
```diff
@@ -267,6 +267,11 @@ ISA::ISA(const Params &p) :
     _regClasses.push_back(&ccRegClass);
     _regClasses.push_back(&miscRegClass);
 
+    panic_if( p.vlen < p.elen,
```
powerjg: This should be a `fatal` (sorry to be pedantic). `fatal` means a config error, `panic` means a gem5 bug. This is a configuration error, I believe.

powerjg (commented):
Looks great to me. Just a couple of small things.

src/arch/riscv/RiscvISA.py
alvaromorenomv: My objective with this sentence is checking that the vlen value is a power of 2 due to VLEN definition in the spec.
I believed it was the way to check it, let me know if I was wrong.

src/arch/riscv/isa.cc
alvaromorenomv: Oh, makes sense !
I am not so used to the usage of the error messages so I got fooled this time, I am changing it and updating.

rogerchang23424 (commented):
LGTM in my side

src/arch/riscv/RiscvISA.py
aarmejach: The check with min and max will be done by the `super()._check()` call

aarmejach:
@BobbyRBruce we need the checkpoint-restore test to be upgraded again using the cpt_upgrader that I just committed. Let me know if this is the right way to proceed or if you need something else from our side.

rogerchang23424 (approved):

BobbyRBruce:
I've updated the resource. The tests should pass now.

BobbyRBruce:
> @BobbyRBruce we need the checkpoint-restore test to be upgraded again using the cpt_upgrader that I just committed. Let me know if this is the right way to proceed or if you need something else from our side.

I can't seem to get this working. Attached is the checkpoint. I can't upgrade it using the upgrader.

[riscv-hello-example-checkpoint-20231009.tar.gz](https://github.com/gem5/gem5/files/12851168/riscv-hello-example-checkpoint-20231009.tar.gz)

aarmejach:
Hi @BobbyRBruce, thanks for the help. The checkpoint you sent is already upgraded and worked on my local branch of this PR!

To make it work locally I had to disable the download of the checkpoint resource of the test, otherwise it overwrites the updated checkpoint. Likely this is what is happening. Somehow we need to ensure that the checkpoint used for testing is the one you sent, not the one that is automatically downloaded if the md5 is different.

BobbyRBruce:
> Hi @BobbyRBruce, thanks for the help. The checkpoint you sent is already upgraded and worked on my local branch of this PR!
> 
> To make it work locally I had to disable the download of the checkpoint resource of the test, otherwise it overwrites the updated checkpoint. Likely this is what is happening. Somehow we need to ensure that the checkpoint used for testing is the one you sent, not the one that is automatically downloaded if the md5 is different.

Oh! Perhaps i've done something silly on my end. I'll double check this.

src/arch/riscv/insts/vector.hh
```diff
@@ -186,6 +195,12 @@ class VectorArithMacroInst : public VectorMacroInst
         this->flags[IsVector] = true;
     }
 
+    VectorArithMacroInst(const char* mnem, ExtMachInst _machInst,
+                         OpClass __opClass, uint32_t _vlen)
+        : VectorMacroInst(mnem, _machInst, __opClass, _vlen)
+    {
+        this->flags[IsVector] = true;
+    }
```
rogerchang23424: (nit) Merge the two constructor with default `vlen`

src/arch/riscv/isa/templates/vector_mem.isa
```diff
@@ -677,6 +706,7 @@ Fault
     }
     %(op_src_decl)s;
     %(op_rd)s;
+    %(set_vlenb)s;
     %(ea_code)s;
 
     Fault fault = initiateMemRead(xc, traceData, EA, Mem, memAccessFlags);
```
rogerchang23424: Please specify the size of memory for read, the method will read the memory data with maximum vector length size. (For Timing, Minor, O3 CPU)

src/arch/riscv/isa/templates/vector_mem.isa
```diff
@@ -644,14 +671,16 @@ Fault
     }
     %(op_decl)s;
     %(op_rd)s;
+    %(set_vlenb)s;
+    %(set_vlen)s;
     %(ea_code)s;
 
     Fault fault = readMemAtomicLE(xc, traceData, EA,
                                   *(vreg_t::Container*)(&Mem), memAccessFlags);
```
rogerchang23424: ditto (For Atomic CPU)

src/arch/riscv/isa/templates/vector_mem.isa
```diff
@@ -551,9 +574,11 @@ Fault
...
     %(op_decl)s;
     %(op_rd)s;
+    %(set_vlenb)s;
     %(ea_code)s;
 
-    for (size_t i = 0; i < VLENB; i++) {
+
+    for (size_t i = 0; i < vlenb; i++) {
         %(memacc_code)s;
     }
```
rogerchang23424, line 593: In the following line 585 initiateMemWrite: Please specify the size of memory for write, the method will read the memory data with maximum vector length size.

src/arch/riscv/isa/templates/vector_mem.isa
```diff
@@ -523,9 +544,11 @@ Fault
...
```
rogerchang23424, line 563: ditto

rogerchang23424 (changes requested):

src/arch/riscv/isa/templates/vector_mem.isa
alvaromorenomv: I have implemented these changes in a new commit, let me know if I correctly understood what you meant.
Thanks so much!

aarmejach:
@BobbyRBruce if you have a moment it would be great to fix the checkpoint test for this to go through as soon as possible.

I'll also start now an issue. I think it would be great if the CI testing runs the cpt_upgrader automatically before running the tests. This will allow anyone to push a PR that includes a cpt_upgrader patch and pass all tests seamlessly.

rogerchang23424:
LGTM, I have run the binary file [rvv_index.c](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/main/examples/rvv_index.c) with riscv-se.py and get the error `MemoryError: std::bad_alloc`.

Reproduce 

complie command (riscv64-unknown-elf-gcc version 13.1)
```bash 
riscv64-unknown-elf-gcc -march=rv64gcv -static rvv_index.c -o rvv_index.elf
```

`configs/example/gem5_library/riscv-se.py`

```python
from gem5.components.cachehierarchies.classic.private_l1_private_l2_cache_hierarchy import (
)
from gem5.isas import ISA
from gem5.resources.resource import BinaryResource

    l1d_size="32KiB", l1i_size="32KiB", l2_size="512KiB"
)

# Setup the system memory.
memory = SingleChannelDDR3_1600()

# Setup a single core Processor.
processor = SimpleProcessor(
    cpu_type=CPUTypes.TIMING, isa=ISA.RISCV, num_cores=1
)

# Setup the board.
board = SimpleBoard(
    clk_freq="1GHz",
    processor=processor,
    memory=memory,
)

board.set_se_binary_workload(
  BinaryResource("/path/to/rvv_index.elf")
)

simulator = Simulator(board=board, full_system=False)
simulator.run()
```

Execute command
```bash
build/ALL/gem5.opt configs/example/gem5_library/riscv-se.py
```

Gem5 compiler GCC: 13.2
Host ISA: X86-64

alvaromorenomv:
> LGTM, I have run the binary file [rvv_index.c](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/main/examples/rvv_index.c) with riscv-se.py and get the error `MemoryError: std::bad_alloc`.
> 
> Reproduce
> 
> complie command (riscv64-unknown-elf-gcc version 13.1)
> 
> ```shell
> riscv64-unknown-elf-gcc -march=rv64gcv -static rvv_index.c -o rvv_index.elf
> ```
> 
> `configs/example/gem5_library/riscv-se.py`
> 
> ```python
> from gem5.components.boards.simple_board import SimpleBoard
> from gem5.components.memory import SingleChannelDDR3_1600
> from gem5.components.processors.simple_processor import SimpleProcessor
> from gem5.components.cachehierarchies.classic.private_l1_private_l2_cache_hierarchy import (
>     PrivateL1PrivateL2CacheHierarchy,
> )
> from gem5.components.processors.cpu_types import CPUTypes
> from gem5.isas import ISA
> from gem5.utils.requires import requires
> from gem5.resources.resource import BinaryResource
> from gem5.simulate.simulator import Simulator
> 
> requires(isa_required=ISA.RISCV)
> 
> cache_hierarchy = PrivateL1PrivateL2CacheHierarchy(
>     l1d_size="32KiB", l1i_size="32KiB", l2_size="512KiB"
> )
> 
> # Setup the system memory.
> memory = SingleChannelDDR3_1600()
> 
> # Setup a single core Processor.
> processor = SimpleProcessor(
>     cpu_type=CPUTypes.TIMING, isa=ISA.RISCV, num_cores=1
> )
> 
> # Setup the board.
> board = SimpleBoard(
>     clk_freq="1GHz",
>     processor=processor,
>     memory=memory,
>     cache_hierarchy=cache_hierarchy,
> )
> 
> # Set the Full System workload.
> board.set_se_binary_workload(
>   BinaryResource("/path/to/rvv_index.elf")
> )
> 
> simulator = Simulator(board=board, full_system=False)
> print("Beginning simulation!")
> simulator.run()
> ```
> 
> Execute command
> 
> ```shell
> build/ALL/gem5.opt configs/example/gem5_library/riscv-se.py
> ```
> 
> Gem5 compiler GCC: 13.2 Host ISA: X86-64

I will try to reproduce it with exactly the same setup and update asap.
On my side, I have tested SE using a natively compiled binary that I built inside a RISCV disk image and had no problems with this concrete benchmark and Timing CPU. Also, I tried FS with Atomic and O3 before pushing the update, and I have not seen this kind of problem.

alvaromorenomv:
> > LGTM, I have run the binary file [rvv_index.c](https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/main/examples/rvv_index.c) with riscv-se.py and get the error `MemoryError: std::bad_alloc`.
> > Reproduce
> > complie command (riscv64-unknown-elf-gcc version 13.1)
> > ```
> > riscv64-unknown-elf-gcc -march=rv64gcv -static rvv_index.c -o rvv_index.elf
> > ```
> > 
> > 
> > `configs/example/gem5_library/riscv-se.py`
> > ```
> > from gem5.components.boards.simple_board import SimpleBoard
> > from gem5.components.memory import SingleChannelDDR3_1600
> > from gem5.components.processors.simple_processor import SimpleProcessor
> > from gem5.components.cachehierarchies.classic.private_l1_private_l2_cache_hierarchy import (
> >     PrivateL1PrivateL2CacheHierarchy,
> > )
> > from gem5.components.processors.cpu_types import CPUTypes
> > from gem5.isas import ISA
> > from gem5.utils.requires import requires
> > from gem5.resources.resource import BinaryResource
> > from gem5.simulate.simulator import Simulator
> > 
> > requires(isa_required=ISA.RISCV)
> > 
> > cache_hierarchy = PrivateL1PrivateL2CacheHierarchy(
> >     l1d_size="32KiB", l1i_size="32KiB", l2_size="512KiB"
> > )
> > 
> > # Setup the system memory.
> > memory = SingleChannelDDR3_1600()
> > 
> > # Setup a single core Processor.
> > processor = SimpleProcessor(
> >     cpu_type=CPUTypes.TIMING, isa=ISA.RISCV, num_cores=1
> > )
> > 
> > # Setup the board.
> > board = SimpleBoard(
> >     clk_freq="1GHz",
> >     processor=processor,
> >     memory=memory,
> >     cache_hierarchy=cache_hierarchy,
> > )
> > 
> > # Set the Full System workload.
> > board.set_se_binary_workload(
> >   BinaryResource("/path/to/rvv_index.elf")
> > )
> > 
> > simulator = Simulator(board=board, full_system=False)
> > print("Beginning simulation!")
> > simulator.run()
> > ```
> > 
> > 
> > Execute command
> > ```
> > build/ALL/gem5.opt configs/example/gem5_library/riscv-se.py
> > ```
> > 
> > 
> > Gem5 compiler GCC: 13.2 Host ISA: X86-64
> 
> I will try to reproduce it with exactly the same setup and update asap. On my side, I have tested SE using a natively compiled binary that I built inside a RISCV disk image and had no problems with this concrete benchmark and Timing CPU. Also, I tried FS with Atomic and O3 before pushing the update, and I have not seen this kind of problem.

I have been able to reproduce the error and committed, what I think is the solution.
Thanks!!

src/arch/riscv/RiscvISA.py
```diff
@@ -58,3 +84,13 @@ class RiscvISA(BaseISA):
     riscv_type = Param.RiscvType("RV64", "RV32 or RV64")
 
     enable_rvv = Param.Bool(True, "Enable vector extension")
+    vlen = Param.RiscvVectorLength(
+        256,
```
jerryzj, line 88: According to RVV spec, v extension implies VLEN=128 by default

src/arch/riscv/insts/vector.hh
```diff
@@ -114,25 +118,30 @@ class VectorMacroInst : public RiscvMacroInst
   protected:
     uint32_t vl;
     uint8_t vtype;
+    uint32_t vlen;
+
     VectorMacroInst(const char* mnem, ExtMachInst _machInst,
-                   OpClass __opClass)
+                   OpClass __opClass, uint32_t _vlen = 256)
```
jerryzj, line 126: Ditto

src/arch/riscv/insts/vector.hh
```diff
@@ -114,25 +118,30 @@ class VectorMacroInst : public RiscvMacroInst
...
-  protected:
-    uint8_t microVl;
+protected:
+    uint32_t vlen;
+    uint32_t microVl;
     uint8_t microIdx;
     uint8_t vtype;
     VectorMicroInst(const char *mnem, ExtMachInst _machInst, OpClass __opClass,
-                    uint8_t _microVl, uint8_t _microIdx)
+      uint32_t _microVl, uint8_t _microIdx, uint32_t _vlen = 256)
```
jerryzj, line 144: Ditto

src/arch/riscv/insts/vector.hh
```diff
@@ -180,12 +189,11 @@ class VectorArithMacroInst : public VectorMacroInst
 {
   protected:
     VectorArithMacroInst(const char* mnem, ExtMachInst _machInst,
-                         OpClass __opClass)
-        : VectorMacroInst(mnem, _machInst, __opClass)
+                         OpClass __opClass, uint32_t _vlen = 256)
```
jerryzj, line 194: Ditto

src/arch/riscv/insts/vector.hh
```diff
@@ -221,8 +229,8 @@ class VectorSlideMacroInst : public VectorMacroInst
 {
   protected:
     VectorSlideMacroInst(const char* mnem, ExtMachInst _machInst,
-                         OpClass __opClass)
-        : VectorMacroInst(mnem, _machInst, __opClass)
+                         OpClass __opClass, uint32_t _vlen = 256)
```
jerryzj, line 234: Ditto

src/arch/riscv/insts/vector.hh
```diff
@@ -266,17 +274,17 @@ class VectorMemMacroInst : public VectorMacroInst
 {
   protected:
     VectorMemMacroInst(const char* mnem, ExtMachInst _machInst,
-                       OpClass __opClass)
-        : VectorMacroInst(mnem, _machInst, __opClass)
+                        OpClass __opClass, uint32_t _vlen = 256)
```
jerryzj, line 279: Ditto

src/arch/riscv/pcstate.hh
```diff
@@ -62,7 +62,7 @@ class PCState : public GenericISA::UPCState<4>
 
     bool _compressed = false;
     RiscvType _rvType = RV64;
-    uint64_t _vlenb = VLENB;
+    uint64_t _vlenb = 256;
```
jerryzj: Ditto

src/arch/riscv/RiscvISA.py
alvaromorenomv: Yes, we should put 128 but that would break the CI tests that depend on checkpoints taken for VLEN=256. I think we should try to get this in asap and later we can fix it.

src/arch/riscv/pcstate.hh
alvaromorenomv: This one should also have default VLENB value, not VLEN as it has now. I am fixing it (Fixed)

src/arch/riscv/RiscvISA.py
powerjg: Can you point me to where in the spec it says 128? Also, what is the importance of this default value? You can always override this in your standard library board or from your config script.

Personally, I strongly prefer to merge this now and we can discuss changing the default later.

src/arch/riscv/insts/vector.hh
powerjg: Do we even need a default here? I think we can require this parameter now.

powerjg (commented):
Other than the default value, is there anything else we're waiting on until we can merge this? I'm ready to start using it! :)

src/arch/riscv/RiscvISA.py
hnpl: Regardless of what the spec says, I do think we should have some gem5 output saying info("RVV enabled, VLEN = %d bits") if RVV is enabled.

jerryzj: @powerjg please read the spec
https://github.com/riscv/riscv-v-spec/blob/v1.0/v-spec.adoc#183-v-vector-extension-for-application-processors

```
The V vector extension requires Zvl128b.
```

hnpl: If I'm not mistaken, `Zvl128b` means VLEN >= 128b. I think the default VLEN = 256b is sensible as long as we're clear about the VLEN we're using.

jerryzj: Nope `Zvl128b` means VLEN=128b, and the spec accepts maximum of VLEN=65536b (`zvl65536b`)

alvaromorenomv:
> Other than the default value, is there anything else we're waiting on until we can merge this? I'm ready to start using it! :)

I have updated the patch solving the conflicts with the current develop branch. Apart from this, at least to my best known, the only remaining task should be upgrading the checkpoint from the CI tests in order to get it in.

src/arch/riscv/insts/vector.hh
```diff
@@ -557,27 +573,29 @@ class VMaskMergeMicroInst : public VectorArithMicroInst
     Fault
     execute(ExecContext* xc, trace::InstRecord* traceData) const override
     {
-        vreg_t& tmp_d0 = *(vreg_t *)xc->getWritableRegOperand(this, 0);
+        vreg_t tmp_d0 = *(vreg_t *)xc->getWritableRegOperand(this, 0);
```
aarmejach: @alvaromorenomv when merging this you undid a change that is required. This line should be:
`vreg_t& tmp_d0 = *(vreg_t *)xc->getWritableRegOperand(this, 0);`

src/arch/riscv/isa.cc
```diff
@@ -267,6 +267,11 @@ ISA::ISA(const Params &p) :
     _regClasses.push_back(&ccRegClass);
     _regClasses.push_back(&miscRegClass);
 
+    fatal_if( p.vlen < p.elen,
+    "VLEN should be greater or equal",
+        "than ELEN. Ch. 2RISC-V vector spec.");
+
```
aarmejach, line 273: Maybe we can add here the message @hnpl suggested:

inform("RVV enabled, VLEN = %d bits, ELEN = %d bits", p.vlen, p.elen)

aarmejach (approved):

src/arch/riscv/insts/vector.hh
alvaromorenomv: I think it is fixed now!

src/arch/riscv/isa.cc
alvaromorenomv: I agree that it is a good place to put it. Added in the last update

rogerchang23424 (approved):
All tests are passed in my side

src/arch/riscv/RiscvISA.py
powerjg: I am *so* confused... 

First, thanks @jerryzj for the review and pointing to the spec.

Are you saying that we cannot create a machine that has a VLEN other than 128b? That doesn't make sense to me. In theory, I can build/buy a RISC-V processor with a VLEN of anything between 128 and 65536. Why does it matter what gem5 sets its default to? Anyone is able to change this value to anything they want *without rebuilding gem5.*

As far as I can tell, the spec doesn't say anything about "default" values.

Could you explain why changing the default from 256 to 128 will benefit the users of gem5?

+1 for making it clear what VLEN is, but I don't know if adding more noise to stdout is the right way to do it.

powerjg:
Looks like there's a couple of small style things :). See https://github.com/gem5/gem5/actions/runs/6546276797/job/17787430131?pr=171

alvaromorenomv:
> Looks like there's a couple of small style things :). See https://github.com/gem5/gem5/actions/runs/6546276797/job/17787430131?pr=171

I have fixed the style, lets see if it is enough now :)

Harshil2107:
I ran the checkpoint updater and uploaded the updated checkpoint  (https://resources.gem5.org/resources/riscv-hello-example-checkpoint?database=gem5-resources&version=3.0.0). I ran the `riscv-hello-restore-checkpoint.py` locally which was failing in the test, the checkpoint now seems to work. Please let me know if there are any issues regarding this.

aarmejach:
> I ran the checkpoint updater and uploaded the updated checkpoint (https://resources.gem5.org/resources/riscv-hello-example-checkpoint?database=gem5-resources&version=3.0.0). I ran the `riscv-hello-restore-checkpoint.py` locally which was failing in the test, the checkpoint now seems to work. Please let me know if there are any issues regarding this.

Thanks, test are passing. Let's have another +1 and we are ok to merge imo.

aarmejach (approved):

powerjg (approved):

PR #172, opened 2023-08-10
cpu-o3: bugfix of rename squash when SMT

HE-Wenjian:
In an SMT CPU, upon a squash, the mis-predicted(squashing) instructions can still be executing at IEW and own phys registers. If these registers are added back to the rename freelist on this Tick, the registers may be renamed to be used by other SMT thread(s). This causes register ownership hazards, which may eventually freeze the CPU. This problem seems to date back to 2014 (https://www.mail-archive.com/gem5-users@gem5.org/msg10180.html).

This patch delays the freelist update to avoid the hazard.

I tested that this patch does not cause any performance impact for my set of benchmarks on default non-SMT O3CPU.


powerjg (approved):

powerjg:
Thanks for the contribution!

PR #175, opened 2023-08-11
misc: Add continue-on-error to matrix runs

mkjost0:
This sets continue-on-error to true on any scheduled test that uses a matrix so we can have all sets of tests run regardless if one of them fails or not.

Change-Id: I8f6137ebdf62a5cecd582387316c330c8a1401ca


mkjost0:
Note: This needs to be copied to stable in order to be applied to all of our scheduled tests

BobbyRBruce (approved):
Thanks Melissa.

I'll make sure this gets on stable.

PR #177, opened 2023-08-11
mem-ruby: Add new feature far atomics in CHI

vsoriap:
Added a new feature to CHI protocol (in collaboration with @tiagormk). Here is the Jira Ticket [https://gem5.atlassian.net/browse/GEM5-1326](https://gem5.atlassian.net/browse/GEM5-1326 ). As described in CHI specs, far atomic transactions enable remote execution of Atomic Memory Operations. This pull request incorporates several changes:

* Fix Arm ISA definition of Swap instructions.  These instructions should return an operand, so their ISA definition should be Return Operation.
* Enable AMOs in Ruby Mem Test to verify that AMOs work
* Enable near and far AMO in the Cache Controler of CHI

Three configuration parameters have been used to tune this behavior:
* policy_type:  sets the atomic policy to one of the described in [our paper](https://dl.acm.org/doi/10.1145/3579371.3589065) 
* atomic_op_latency: simulates the AMO ALU operation latency
* comp_anr: configures the Atomic No return transaction to split CompDBIDResp into two different messages DBIDResp and Comp


src/mem/ruby/system/Sequencer.cc
```diff
@@ -807,6 +870,19 @@ Sequencer::makeRequest(PacketPtr pkt)
     } else if (pkt->req->isTlbiCmd()) {
         primary_type = secondary_type = tlbiCmdToRubyRequestType(pkt);
         DPRINTF(RubySequencer, "Issuing TLBI\n");
+#if defined (PROTOCOL_CHI)
```
powerjg, line 878: Is this necessary? Shouldn't all protocols have these functions/types defined?

src/mem/ruby/system/Sequencer.cc
```diff
@@ -908,6 +984,42 @@ Sequencer::issueRequest(PacketPtr pkt, RubyRequestType secondary_type)
...
+        // Create the blocksize, access mask and atomicops
+        uint32_t blockSize = RubySystem::getBlockSizeBytes();
+        std::vector<bool> accessMask(blockSize,false);
+        std::vector< std::pair<int,AtomicOpFunctor*> > atomicOps;
+        uint32_t tmpOffset = getOffset(pkt->getAddr());
+        uint32_t tmpSize = pkt->getSize();
+        std::pair<int,AtomicOpFunctor *> tmpAtomicOp(tmpOffset,
+                                                        pkt->getAtomicOp());
+        atomicOps.push_back(tmpAtomicOp);
```
powerjg: Should be able to use `emplace` and get rid of the line above with `tmpAtomicOp`

src/mem/ruby/system/Sequencer.cc
```diff
@@ -908,6 +984,42 @@ Sequencer::issueRequest(PacketPtr pkt, RubyRequestType secondary_type)
             DPRINTF(RubySequencer, "Issuing TLBI %016x\n",
                     msg->m_tlbiTransactionUid);
         }
+    } else if (pkt->isAtomicOp()){
+#if defined (PROTOCOL_CHI)
```
powerjg: If at all possible, I would like to avoid these ifdefs so that we can have a single gem5 build will all Ruby protocols. Could you explain why this ifdef is needed?

src/mem/ruby/system/Sequencer.cc
```diff
@@ -594,6 +600,57 @@ Sequencer::readCallback(Addr address, DataBlock& data,
     }
 }
 
+void
+Sequencer::atomicCallback(Addr address, DataBlock& data,
```
powerjg, line 604: I'm having trouble following the logic in this function. Could you explain a bit about what it's doing and why?

powerjg (commented):
First of all, this is great work and I really appreciate all of the effort to push this publicly!

Generally, it all looks good to me. I didn't check the details of the CHI changes, but I trust it's been tested :)

I only have a few smallish comments below. I would really like to understand better how/when the atomic functors are executed as well.

src/mem/ruby/system/Sequencer.cc
vsoriap: All other protocols use RubyRequestType_ST to encode Atomic Memory Operations, while in our implementation, we use RubyRequestType_ATOMIC_RETURN.

For example, in MESI_Three_Level, an AMO would be encoded as a Store type. Then the Data L1 will gain write permission to perform the AMO. Once the Cache Controller calls writecallback the AMO is performed, returning the previous value stored in the cache.

One option to have the same code for all protocols is to adapt them to support RubyRequestType_ATOMIC_RETURN.

src/mem/ruby/system/Sequencer.cc
vsoriap: In this case, we can remove the `ifdef`, if we always add the access mask and the atomic vector for all cases although it is empty.

src/mem/ruby/system/Sequencer.cc
vsoriap: Some operations can be coalesced with an Atomic Memory Operation. However, if the atomic was performed remotely, it is impossible to perform further atomics/writes in the same cache line. Therefore, we reissue the other requests to the ruby system.

We could make use of the already existing WriteCallback. However, we thought the proper way to address this new feature is to create a specific callback for atomic operations. This also enables future features like coalescing atomic operations as described in #200 but for CPU rather than GPU.

src/mem/ruby/system/Sequencer.cc
vsoriap: Yes, we can simplify the code

src/mem/ruby/system/Sequencer.cc
powerjg: I see. This makes sense to me now. However, I don't think the comments are doing a good job communicating the design. E.g., "perform hitCallback on first cpu request" is actually iterating through all requests and "free the ruby request form the list" is about 30 lines away from where that's actually done.

It would improve this change to update the comments in this new function.

src/mem/ruby/system/Sequencer.cc
powerjg: Got it. This makes sense, and I'll just have to figure out a way to deal with it in #117

src/mem/ruby/slicc_interface/RubyRequest.hh
```diff
@@ -113,6 +113,38 @@ class RubyRequest : public Message
         }
     }
 
+
```
tiagormk: Do we need a new constructor ? Just adding a `setWriteMask` function for the atomic requests should be cleaner.

src/mem/ruby/system/Sequencer.cc
tiagormk: For now I'd be fine with this. Each protocol would need to declare somewhere which RubyRequestType it supports so the sequencer can choose the most specific request type.

src/mem/ruby/system/Sequencer.cc
tiagormk: The earlier ifdef already selects the request type if it's CHI, so here you can avoid it and just check if both `pkt->isAtomicOp()` is true and `secondary_type`  is either `ATOMIC_RETURN` or `ATOMIC_NO_RETURN`

src/mem/ruby/system/Sequencer.cc
```diff
@@ -908,6 +984,42 @@ Sequencer::issueRequest(PacketPtr pkt, RubyRequestType secondary_type)
...
+        uint32_t tmpOffset = getOffset(pkt->getAddr());
+        uint32_t tmpSize = pkt->getSize();
+        std::pair<int,AtomicOpFunctor *> tmpAtomicOp(tmpOffset,
+                                                        pkt->getAtomicOp());
+        atomicOps.push_back(tmpAtomicOp);
+        for (int j = 0; j < tmpSize; j++) {
+            accessMask[tmpOffset + j] = true;
+        }
+
+        msg = std::make_shared<RubyRequest>(clockEdge(), pkt->getAddr(),
```
tiagormk: see my previous comment on src/mem/ruby/slicc_interface/RubyRequest.hh

tiagormk (commented):
Thanks @vsoriap for the great work!
Could you elaborate on the commit msg a bit. Mostly add the same comments you have for the policy parameter.

tiagormk (commented):
Could this be combined with the previous commit ?

src/mem/ruby/slicc_interface/RubyRequest.hh
vsoriap: We can remove it once we unify the code for creating a request

vsoriap:
Hi, I have incorporated all the changes suggested in the reviews.

powerjg (commented):
LGTM, but I have to admit I'm not an expert here. I'll let Giacomo or Tiago sign off before merging.

giactra (changes requested):
Thanks for your contribution! I will let @tiagormk approve the PR. The only comment I have is regarding the CHI_DMAController. Shall we set alloc_on_atomic = False? I believe gem5 will complain otherwise if the parameter is not set

tiagormk:
> Thanks for your contribution! I will let @tiagormk approve the PR. The only comment I have is regarding the CHI_DMAController. Shall we set alloc_on_atomic = False? I believe gem5 will complain otherwise if the parameter is not set

CHI_DMAController does also need the 'alloc_on_atomic = False'. LGTM otherwise.

vsoriap:
Hi, I added the changes in the DMA Controller

tiagormk (approved):

powerjg (approved):

giactra (approved):

yichungchen84:
This feature incurs deadlocks when configuring a cluster of cpus with a shared L3 attached to a RNF (mimicking a DSU). The same configuration works well on GEM5-23.0.1.0.

vsoriap:
Hi, which policy are you using for Atomic Memory Operations? I assume you have a L1/L2 private and a L3 Shared. If you remove the L2, it may rise problems.

yichungchen84:
@vsoriap I have L1 private, L2 private, and L3 shared on RNF, and another shared memory on HNF. Just default policy in 23.1.0.0 without any change. 
I also have the other configuration with all alloc_on_atomic off on all cache/dma controllers. I assumed this should have no atomic allocation policy. however, it still has deadlock problems.

vsoriap:
This is weird behavior because that's the exact configuration we tested. 32 ArmO3CPU with private L1 & L2 and 32x shared HNF. Could you open an issue to discuss this? Sorry for the delay, but these days I'm a bit busy.

yichungchen84:
Hi @vsoriap,
My configuration has a local shared L3 at RNF (like a cluster of CPUs attached to an RNF). After studying the difference, I found that near Atomic could be the problem. Thanks
Issue #882

PR #130, opened 2023-07-27
misc: Sync CONTRIBUTING.md with website

BobbyRBruce:
This change syncs the repo's contributing documentation with that of the website's contributing documentation:
https://www.gem5.org/contributing

From now on we'll attempt to keep the repo's CONTRIBUTING.md documentation in sync with that on the website.

Change-Id: I2c91e6dd5cd7a9b642377878b007d7da3f0ee2ad

PR #178, opened 2023-08-11
misc: Re-add header grammar guide to CONTRIBUTING.md

BobbyRBruce:
Addresses this the issue highlighted here:
https://github.com/gem5/gem5/pull/130#issuecomment-1656437005

Via https://github.com/gem5/gem5/pull/130 the grammar guide on how to write commit headers was removed. This re-adds it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -245,8 +245,9 @@ then a colon. Which tags are used depend on which components of gem5
 you have modified. **Please refer to the [MAINTAINERS.yaml](
 https://github.com/gem5/gem5/blob/stable/MAINTAINERS.yaml) for
 a comprehensive list of accepted tags**. After this colon a short description
-of the commit must be provided. **This header line must not exceed 65
-characters**.
+of the commit must be provided. This short description should be written in the imperative mood and should state what happens when the patch is applied. Keep
+it short and simple. Write it in sentence case, preferably with no punctuation.
+**This header line must not exceed 65 characters**.
 
 After this, a more detailed description of the commit can be added. This is
 inserted below the header, separated by an empty line. Including a description


shingarov (changes requested):
I do not think this commit addresses the real issues with #130.  The details of what the grammar-guide says or does not say, are of relatively low importance, compared to much larger issues such as losing version-control metadata (git-blame info, gerrit review ChangeIds), and even more worryingly, mass-editing over multiple changes which have been individually reviewed on gerrit.

I propose we do some collective brainstorming before we can decide what to do about #130.

BobbyRBruce:
> 



> I do not think this commit addresses the real issues with #130. The details of what the grammar-guide says or does not say, are of relatively low importance, compared to much larger issues such as losing version-control metadata (git-blame info, gerrit review ChangeIds), and even more worryingly, mass-editing over multiple changes which have been individually reviewed on gerrit.
> 
> I propose we do some collective brainstorming before we can decide what to do about #130.

We would welcome suggestions for improving our processes. I think the best way to do this would be to post on the gem5-dev mailing list. Specific examples of current pain-points and proposed solutions to them would be welcome to help invoke discussion.

On this specific PR, is there anything additional you need me to include in CONTRIBUTING.md?

powerjg:
@shingarov I am assigned this change, which means that I'm responsible to make sure that we're making progress towards merging it. Could you let Bobby know what changes you're looking for when you requested changes? I don't see any comments on the diff or specific suggestions. Would it be sufficient to drop "address comment on #130" statement from the commit message?

shingarov:
> what changes you're looking for?

> what changes you're looking for when you requested changes? I don't see any comments on the diff or specific suggestions

We painted ourselves into a corner where we can't fix the problem by applying a diff.  Let me explain myself.

The problem I am talking about, is not with the _content_ of `CONTRIBUTING.md`, but rather with the git history, from the perspective of trust in the repo's integrity.  Let's think about our reasons we are even bothering with following Gerrit (or now GitHub) review process.  Among many, I can think of (1) level-of-assurance (i.e. the dimension of trust which the Common Criteria contrast with the "target-of-evaluation").  For *every* line, are we absolutely sure how that line got into the codebase?  who wrote that line?  for what reasons?  was there consensus, and is the discussion leading to that consensus well-documented?  so this is legal/compliance/etc dimension without which we can't grow to attract organizations with the level of resources necessary to participate in a foundational project like ours.  Along the technical dimension, (2) avoiding Chesterton's Fence is one of the most difficult problems for software *growth*.  For this comment to be self-contained, let's recall:  If I see a fence in my system blocking my way, and I can't think of any reason why that fence is necessary, shall I just demolish it?  Probably it was erected for historical reasons which are no longer valid?  How do I find out?  So what we are enforcing with our process, is, the fence is guaranteed to have a label attached to it, saying, "this fence is erected on <this date> by Boris upon discussion with Bobby and Jason, at <this url>".  And there are tools to work with this label productively; if I am on a line in vim, I do `:G blame`, go left, tap on the sha, then down in the "git show" I click on the change-id and that opens the Gerrit review.

Now.  For `CONTRIBUTING.md` in particular, in December 2021 I wasn't sure whether the website is following some other process of its own, that's why I asked on the mailing list.  This is what Jason told me:

> Feel free to make a gerrit change request to the CONTRIBUTING.md file.

All in all, we do seem to be telling the community that the only way to change gem5 is via the review process.  In that sense I believe dcf3c4b breaks the integrity of the repo, by simply bulk-overwriting over multiple changes which were carefully separated, by their respective authors and reviewers, into logically-independent PRs, putting in change-ids, review urls, etc.

What would I propose?  As painful as it sounds, I best I can think of is `git drop dcf3c4b`.  Then we can break the residual diff into logically-cohrent chunks (I will simply agree with whatever Bobby deems "logically coherent"), make each a PR, then [at least] I (and maybe Jason?) will give them a "+2" right away (I already went through them so I know I am ok with them), we'll give them 48 hours to give others a chance to object, and merge.

powerjg:
@shingarov generally I agree with what you're saying, but we have to weigh these points with the usability/maintainability of the project. In this case, it was hurting usability/maintainability to have two different files with slightly different information and slightly different formatting. We chose one (relatively arbitrarily) to copy onto the other. 

I don't think we lost any of the git history (correct me if I'm wrong). And while blame won't work well on this file because it was reformatting, the history *is* there. 

Unless I'm fundamentally misunderstanding how git works, I don't think there is anything that breaks any integrity. The whole git history is still available, but we have made a trade off between the common case (reading CONTRIBUTING.md) and the uncommon case (looking at the git history) and chose to optimize for the common case.

shingarov:
@powerjg I highly appreciate your patience on this seemingly-trivial issue.  The reason I am being overly sensitive is because I have been through the experience where not paying attention over these tiny stupid details led to catastrophic consequences.

> We chose one (relatively arbitrarily) to copy onto the other

The crux of my objection is `copy`.  In git, a change doesn't replace "revision N of file F" with "revision N+1", because there are no files as units of information.  For example, `CONTRIBUTING.md` is a bunch of "hunks" of change, each hunk being a part of a "change-with-its-associated-review".  When you do a `cp`, you potentially overwrite hunks coming from 20 changes — that's potentially 20 authors and 20 reviewers, putting us at the risk of a contributor thinking to themselves: "why do I bother following this painstaking process, if this project's rules allow a single `cp` to summarily smash the work of 40 people?"

> I don't think we lost any of the git history (correct me if I'm wrong).

To the extent that the bits are still there, you are right, from that perspective we didn't lose anything.  At the same time, dcf3c4b does throw doubt over whether our chain-of-custody is up to our promise that "changes only go in via the review process".  This is why I was asking in #130, "did all those changes in "revision N+1" undergo any reviews?"

> we have to weigh these points with the usability/maintainability of the project

Understood, and I am happy that we are acknowledging the existence of a trade-off.  Something that got me worried in the recent conversations since the GitHub migration has started, is that I haven't heard any words relating to issues like "assurance level" or "reviewer takes responsibility for the quality of the change, as opposed to the author", etc.  This would potentially mean that either we are unaware of this whole area of problematics, or that it's not a priority for gem5.  I hope my worries are unfounded.

Now, having said these things, I think I have achieved what I wanted to achieve: whatever insight I might have is now written down for the project to use (just in case it's useful).  As long as we are explicitly aware of the issues, I am fine with whatever way we decide to go.  If we want to `git drop dcf3c4b` and "review N+1" that's one way; if you guys object, simply merging this is fine too.

shingarov:
@powerjg 
> which means that I'm responsible to make sure that we're making progress

Any objection against `git drop dcf3c4b`?

powerjg:
A couple of things:
1. I'm not sure what you mean by `git drop`. I can't find any information about this in the git man page. 
2. I *really* don't want to re-write history on a public branch. If `git drop` is like rebasing and dropping a commit, I am strongly against this. It would make it very difficult for all of the people that have cloned `gem5/develop`.
3. I think we should merge this PR and move on with development 😄

BobbyRBruce:
This contribution has been left stale for over a month. Closing.

PR #180, opened 2023-08-14
Port trace in xbar when address error

yanleeG:
When xbar encounters the address error, print out the port trace in the packet for user to debug if the port trace is enabled.

To gain the packet of the access, the parameter of findPort() function is changed from AddrRange to PacketPtr.

When running gem5 with "--debug-flags=PortTrace", we can see the full path of the unexpected access when xbar cannot find the destination of the address.

### DIFF
--- a/src/mem/coherent_xbar.cc
+++ b/src/mem/coherent_xbar.cc
@@ -159,7 +159,7 @@ CoherentXBar::recvTimingReq(PacketPtr pkt, PortID cpu_side_port_id)
     assert(is_express_snoop == cache_responding);
 
     // determine the destination based on the destination address range
-    PortID mem_side_port_id = findPort(pkt->getAddrRange());
+    PortID mem_side_port_id = findPort(pkt);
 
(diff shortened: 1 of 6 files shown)


powerjg:
Overall, this change is good with me. However, it looks like there are some other places that use the `findPort` function: https://github.com/gem5/gem5/actions/runs/5854325967/job/15870511333?pr=180#step:5:4255

yanleeG:
> Overall, this change is good with me. However, it looks like there are some other places that use the `findPort` function: https://github.com/gem5/gem5/actions/runs/5854325967/job/15870511333?pr=180#step:5:4255

Thanks for pointing this out! I did not notice that my commit does not cover all findPort() function. In order to support the original `findPort(AddrRange)` within the `CoherentXBar::recvMemBackdoorReq`, I make the packet pointer optional in `findPort()`.

powerjg (approved):
Great! I like that this is fully "backwards compatible" as well :)

BobbyRBruce (approved):

PR #183, opened 2023-08-15
fastmodel: Add option to retry licence server connection.

drinkcat:
We're seeing some occasional connection timeouts in CI, possibly when we aggressively hit the license server, so let's add a parameter to retry the connection a few times.

Also, print the time required to connect to the server to help debug issues.


drinkcat:
@gabemblack FYI

andysan (approved):

PR #184, opened 2023-08-15
gpu-compute: Change kernel-based exit location

abmerop:
The previous exit event occurs when the dispatcher sends a completion signal for a kernel, but gem5 does some kernel-based stats updates after the signal is sent. Therefore, if these exit events are used as a way to dump per-kernel stats, some of the stats for the kernel that just ended will be in the next kernel's stat dump which is misleading.

This patch moves the exit event to where the stats are updated and only exits if the dispatcher has requested a stat dump to prevent situations where stats are updated mid-kernel.

Change-Id: I74dc1cad5fc90382a2a80564764b3e7c9fb65521

### DIFF
--- a/src/gpu-compute/dispatcher.cc
+++ b/src/gpu-compute/dispatcher.cc
@@ -334,7 +334,7 @@ GPUDispatcher::notifyWgCompl(Wavefront *wf)
         DPRINTF(GPUKernelInfo, "Completed kernel %d\n", kern_id);
 
         if (kernelExitEvents) {
-            exitSimLoop("GPU Kernel Completed");
+            shader->requestKernelExitEvent();
         }
     }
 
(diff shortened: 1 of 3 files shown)


powerjg (dismissed):

abmerop:
Fixed a build issue. Moved method from protected to public scope, otherwise no code changes

powerjg (approved):

mattsinc (approved):
My only concern is about what happens if multiple streams are running -- would the flag be confused across streams?  But this is certainly an improvement regardless.

abmerop:
> My only concern is about what happens if multiple streams are running -- would the flag be confused across streams? But this is certainly an improvement regardless.

Hmm.  I haven't tried this, but I assume kernels running at the same time would be stepping on each other with or without this change

mattsinc:
> > My only concern is about what happens if multiple streams are running -- would the flag be confused across streams? But this is certainly an improvement regardless.
> 
> Hmm. I haven't tried this, but I assume kernels running at the same time would be stepping on each other with or without this change

That is probably true and that's why I support this as is.  But probably something we need to think about (an array of end signals, one per queue?) moving forward.

PR #185, opened 2023-08-15
mem: Fixing memory size type issue in port proxy.

mahyarsamani:
This patch changes the data type used for image size from int to uint64_t. Current version allows initializing AbstractMemory types with a maximum binary size of 2GiB. This will be limiting in many studies.

Change-Id: Iea3bbd525d4a67aa7cf606f6311aef66c9b4a52c


powerjg (approved):
lgtm

PR #187, opened 2023-08-16
arch-riscv: Check CSR before executing VMem instructions

rogerchang23424:
. Any instructions require vector register should check if vector is enabled. Any instructions need vtype CSR to execute them should check vill bit beforehead.

### DIFF
--- a/src/arch/riscv/isa/templates/vector_mem.isa
+++ b/src/arch/riscv/isa/templates/vector_mem.isa
@@ -177,6 +177,12 @@ Fault
     %(op_rd)s;
     %(ea_code)s;
 
+    MISA misa = xc->readMiscReg(MISCREG_ISA);
+    STATUS status = xc->readMiscReg(MISCREG_STATUS);
+    if (!misa.rvv || status.vs == VPUStatus::OFF) {
+        return std::make_shared<IllegalInstFault>(
+            "RVV is disabled or VPU is off", machInst);
+    }
     if (machInst.vill)
         return std::make_shared<IllegalInstFault>("VILL is set", machInst);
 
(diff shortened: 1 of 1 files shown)


BobbyRBruce:
@hnpl : Can you give this a review?

src/arch/riscv/isa/templates/vector_mem.isa
```diff
@@ -137,6 +137,8 @@ Fault
         return std::make_shared<IllegalInstFault>(
             "RVV is disabled or VPU is off", machInst);
     }
+    if (machInst.vill)
+        return std::make_shared<IllegalInstFault>("VILL is set", machInst);
```
powerjg, line 141: Just curious... why do you raise this exception in execute instead of decode?

src/arch/riscv/isa/templates/vector_mem.isa
```diff
@@ -175,6 +177,15 @@ Fault
     %(op_rd)s;
     %(ea_code)s;
 
+    MISA misa = xc->readMiscReg(MISCREG_ISA);
+    STATUS status = xc->readMiscReg(MISCREG_STATUS);
+    if (!misa.rvv || status.vs == VPUStatus::OFF) {
+        return std::make_shared<IllegalInstFault>(
```
powerjg, line 183: I think it makes sense to check the `VPUStatus` in execute, but I think the `misa.rvv` and `vill` should be checked at decode time.

Please correct me if I'm misunderstanding.

src/arch/riscv/isa/templates/vector_mem.isa
rogerchang23424: "Any vector instructions depended on vtype will raise illegal instruction" according to [spec](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#344-vector-type-illegal-vill). IMHO, it's OK for me to put it in either decode or execute. However, in the gem5 model, we can only raise illegal instruction in execute, just put them in execute.

Please let me know your consideration.

src/arch/riscv/isa/templates/vector_mem.isa
rogerchang23424: Same as above.

hnpl (approved):
LGTM!

src/arch/riscv/isa/templates/vector_mem.isa
powerjg: Got it. Thanks!

powerjg (approved):

PR #188, opened 2023-08-16
misc: Remove 'continue-on-error' from workload files

BobbyRBruce:
'continue-on-error' does allow for a job to run even if other jobs have failed. However, if the job fails the 'continue-on-error' setting does not set the workload as Failed if the job failed.
'if: ${{ success() || failure() }}' if the correct. This ensures a job runs even if the Workload is in a pass or success state.


mkjost0:
Using continue-on-error will mistakenly mark failed matrix runs as successful, which can make it difficult to spot failed scheduled runs.  Using success() || failure() does not stop the matrix from canceling the rest of the runs in the matrix if one job within it fails, which is not useful for our scheduled tests.  However, setting the fail-fast setting as false within the matrix settings both allows all parts of the matrix to run even if part of it fails, and also marks the workflow as failed if parts of the matrix fails, which is the behavior we are aiming for.  This change is made in this pull request: https://github.com/gem5/gem5/pull/194

BobbyRBruce:
> Using continue-on-error will mistakenly mark failed matrix runs as successful, which can make it difficult to spot failed scheduled runs. Using success() || failure() does not stop the matrix from canceling the rest of the runs in the matrix if one job within it fails, which is not useful for our scheduled tests. However, setting the fail-fast setting as false within the matrix settings both allows all parts of the matrix to run even if part of it fails, and also marks the workflow as failed if parts of the matrix fails, which is the behavior we are aiming for. This change is made in this pull request: #194

Closing this PR in favor of #194

PR #189, opened 2023-08-16
arch-x86,cpu-kvm: Fix gem5.fast due to unused variable

BobbyRBruce:
Detected via this failing workload:
https://github.com/gem5/gem5/actions/runs/5861958237

Ir caused the following compilation error to be thrown:

```
src/arch/x86/kvm/x86_cpu.cc:1462:22: error: unused variable ‘rv’ [-Werror=unused-variable]
 1462 |                 bool rv = isa->cpuid->doCpuid(tc, function, idx, cpuid);
      |                      ^~
```

`rv` is unused in the .fast compilation as it's only used in the `assert` statement immediately after. 

To fix this, the `[[maybe_unused]]` annotation is used.

### DIFF
--- a/src/arch/x86/kvm/x86_cpu.cc
+++ b/src/arch/x86/kvm/x86_cpu.cc
@@ -1459,7 +1459,8 @@ X86KvmCPU::updateCPUID()
             m5_supported.push_back(makeKvmCpuid(function, idx, cpuid));
         } else {
             while (true) {
-                bool rv = isa->cpuid->doCpuid(tc, function, idx, cpuid);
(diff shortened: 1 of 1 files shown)


abmerop (changes requested):
This will optimize the doCpuid call out completely in gem5.fast builds. It would be better to place a [[maybe_unused]] attribute on the `rv` variable

BobbyRBruce:
> This will optimize the doCpuid call out completely in gem5.fast builds. It would be better to place a [[maybe_unused]] attribute on the `rv` variable

You're right. I was too focused on the return to appreciate `doCupid` may do something other than just return a value. Fixed.

abmerop (approved):

PR #191, opened 2023-08-17
resources, stdlib: Adding 'suite' category to gem5

Harshil2107:
Change-Id: I4a3231db4ab367c6e25bd97a67b92c2499330246

### DIFF
--- a/configs/example/gem5_library/riscvmatched-microbenchmark-suite.py
+++ b/configs/example/gem5_library/riscvmatched-microbenchmark-suite.py
@@ -50,15 +50,9 @@ microbenchmarks = obtain_resource("riscv-vertical-microbenchmarks")
 # list all the microbenchmarks present in the suite
 print("Microbenchmarks present in the suite:")
 print("====================================")
-for (
-    id,
-    resource_version,
-    input_group,
-    workload,
-) in microbenchmarks.get_all_workloads():
-    print(f"Workload ID: {id}")
-    print(f"Workload Version: {resource_version}")
-    print(f"Workload Input Groups: {input_group}")
+for workload in microbenchmarks:
+    print(f"Workload ID: {workload.get_id()}")
+    print(f"Workload Version: {workload.get_resource_version()}")
     print(f"WorkloadResource Object: {workload}")
     print("====================================")
 
--- a/configs/example/gem5_library/riscvmatched-microbenchmark-suite.py
+++ b/configs/example/gem5_library/riscvmatched-microbenchmark-suite.py
@@ -1,0 +1,86 @@
+# Copyright (c) 2023 The Regents of the University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
+This script shows how to use a suite. In this example, we will use the
+RISCVMatchedBoard and the RISCV Vertical Microbenchmark Suite,
+and show the different functionalities of the suite.
+
+The print statements in the script are for illustrative purposes only,
+and are not required to run the script.
+"""
+
+from gem5.resources.resource import obtain_resource
+from gem5.simulate.simulator import Simulator
+from gem5.prebuilt.riscvmatched.riscvmatched_board import RISCVMatchedBoard
+from gem5.isas import ISA
+from gem5.utils.requires import requires
+
+requires(isa_required=ISA.RISCV)
+
+# instantiate the riscv matched board with default parameters
+board = RISCVMatchedBoard()
+
+# obtain the RISC-V Vertical Microbenchmarks
+microbenchmarks = obtain_resource("riscv-vertical-microbenchmarks")
+
+# list all the microbenchmarks present in the suite
+print("Microbenchmarks present in the suite:")
+print("====================================")
+for (
+    id,
+    resource_version,
+    input_group,
+    workload,
+) in microbenchmarks.get_all_workloads():
+    print(f"Workload ID: {id}")
+    print(f"Workload Version: {resource_version}")
+    print(f"Workload Input Groups: {input_group}")
+    print(f"WorkloadResource Object: {workload}")
+    print("====================================")
+
+# list all the WorkloadResource objects present in the suite
+for resource in microbenchmarks:
+    print(f"WorkloadResource Object: {resource}")
+
+# list all the available input groups in the suite
+print("Input groups present in the suite:")
+print(microbenchmarks.get_input_groups())
+
+# for this example, we will filter the suite
+# to run the Workload "riscv-cca-run"
+# it has the input group 'cca', which is used as the filter
+board.set_workload(list(microbenchmarks.with_input_group("cca"))[0])
+
+# run the simulation with the RISCV Matched board
+simulator = Simulator(board=board, full_system=False)
+simulator.run()
+print(
+    "Exiting @ tick {} because {}.".format(
+        simulator.get_current_tick(),
+        simulator.get_last_exit_event_cause(),
+    )
+)
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -980,7 +980,6 @@ def obtain_resource(
     # Once we know what AbstractResource subclass we are using, we create it.
     # The fields in the JSON object are assumed to map like-for-like to the
     # subclass contructor, so we can pass the resource_json map directly.
-
     return resource_class(local_path=to_path, **resource_json)
 
 
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -25,6 +25,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 from abc import ABCMeta
+import json
 import os
 from pathlib import Path
 from m5.util import warn, fatal
@@ -35,7 +36,17 @@ from .downloader import get_resource
 from .looppoint import LooppointCsvLoader, LooppointJsonLoader
 from ..isas import ISA, get_isa_from_str
 
-from typing import Optional, Dict, Union, Type, Tuple, List, Any
+from typing import (
+    Optional,
+    Dict,
+    Union,
+    Type,
+    Tuple,
+    List,
+    Any,
+    Set,
+    Generator,
+)
 
 from .client import get_resource_json_obj
 
@@ -627,6 +638,111 @@ class SimpointDirectoryResource(SimpointResource):
         return "SimpointDirectoryResource"
 
 
+class SuiteResource(AbstractResource):
+    """
+    A suite resource. This resource is used to specify a suite of workloads to
+    run on a board. It contains a list of workloads to run, along with their
+    IDs and versions.
+
+    Each workload in a suite is used to create a `WorkloadResource` object.
+    These objects are stored in a list and can be iterated over.
+    """
+
+    def __init__(
+        self,
+        workloads: Dict["WorkloadResource", Set[str]] = {},
+        resource_version: Optional[str] = None,
+        description: Optional[str] = None,
+        source: Optional[str] = None,
+        id: Optional[str] = None,
+        **kwargs,
+    ) -> None:
+        """
+        :param workloads: A list of `WorkloadResource` objects
+        created from the `_workloads` parameter.
+        :param local_path: The path on the host system where this resource is
+        located.
+        :param description: Description describing this resource. Not a
+        required parameter. By default is None.
+        :param source: The source (as in "source code") for this resource
+        on gem5-resources. Not a required parameter. By default is None.
+        :param resource_version: Version of the resource itself.
+        """
+        self._workloads = workloads
+        self._description = description
+        self._source = source
+        self._resource_version = resource_version
+
+        super().__init__(
+            id=id,
+            description=description,
+            source=source,
+            resource_version=resource_version,
+        )
+
+    def __iter__(self) -> Generator["WorkloadResource", None, None]:
+        """
+        Returns a generator that iterates over the workloads in the suite.
+
+        :yields: A generator that iterates over the workloads in the suite.
+        """
+        for workload in self._workloads.keys():
+            yield workload
+
+    def __len__(self):
+        """
+        Returns the number of workloads in the suite.
+
+        :returns: The number of workloads in the suite.
+        """
+        return len(self._workloads)
+
+    def get_category_name(cls) -> str:
+        return "SuiteResource"
+
+    def with_input_group(self, input_group: str) -> "SuiteResource":
+        """
+        Returns a new SuiteResource object with only the workloads that use the
+        specified input group.
+
+        :param input_group: The input group to filter the workloads by.
+        :returns: A new SuiteResource object with only the workloads that use
+        the specified input group.
+        """
+
+        if input_group not in self.get_input_groups():
+            raise Exception(
+                f"Input group {input_group} not found in Suite.\n"
+                f"Available input groups are {self.get_input_groups()}"
+            )
+
+        filtered_workloads = {}
+
+        for workload, input_groups in self._workloads.items():
+            if input_group in input_groups:
+                filtered_workloads[workload] = input_groups
+
+        return SuiteResource(
+            local_path=self._local_path,
+            resource_version=self._resource_version,
+            description=self._description,
+            source=self._source,
+            workloads=filtered_workloads,
+        )
+
+    def get_input_groups(self) -> Set[str]:
+        """
+        Returns a set of all input groups used by the workloads in a suite.
+
+        :returns: A set of all input groups used by the workloads in a suite.
+        """
+        return {
+            input_group
+            for input_groups in self._workloads.values()
+            for input_group in input_groups
+        }
+
+
 class WorkloadResource(AbstractResource):
     """A workload resource. This resource is used to specify a workload to run
     on a board. It contains the function to call and the parameters to pass to
@@ -657,8 +773,13 @@ class WorkloadResource(AbstractResource):
             resource_version=resource_version,
         )
 
+        self._id = id
         self._func = function
         self._params = parameters
+
+    def get_id(self) -> str:
+        """Returns the ID of the workload."""
+        return self._id
 
     def get_function_str(self) -> str:
         """
@@ -821,6 +942,21 @@ def obtain_resource(
     assert resources_category in _get_resource_json_type_map
     resource_class = _get_resource_json_type_map[resources_category]
 
+    if resources_category == "suite":
+        workloads = resource_json["workloads"]
+        workloads_obj = {}
+        for workload in workloads:
+            workloads_obj[
+                obtain_resource(
+                    workload["id"],
+                    resource_version=workload["resource_version"],
+                    resource_directory=resource_directory,
+                    clients=clients,
+                    gem5_version=gem5_version,
+                )
+            ] = set(workload["input_group"])
+        resource_json["workloads"] = workloads_obj
+
     if resources_category == "workload":
         # This parses the "resources" and "additional_params" fields of the
         # workload resource into a dictionary of AbstractResource objects and
@@ -844,6 +980,7 @@ def obtain_resource(
     # Once we know what AbstractResource subclass we are using, we create it.
     # The fields in the JSON object are assumed to map like-for-like to the
     # subclass contructor, so we can pass the resource_json map directly.
+
     return resource_class(local_path=to_path, **resource_json)
 
 
@@ -995,5 +1132,6 @@ _get_resource_json_type_map = {
     "resource": Resource,
     "looppoint-pinpoint-csv": LooppointCsvResource,
     "looppoint-json": LooppointJsonResource,
+    "suite": SuiteResource,
     "workload": WorkloadResource,
 }
--- a/tests/pyunit/stdlib/resources/pyunit_suite_checks.py
+++ b/tests/pyunit/stdlib/resources/pyunit_suite_checks.py
@@ -1,0 +1,170 @@
+# Copyright (c) 2023 The Regents of the University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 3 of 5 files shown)


src/python/gem5/resources/suite.py
```diff
@@ -0,0 +1,123 @@
...
+
+}
+class suite:
+
+impl: iter, next(maybe) , len
+
+methods: list_input_groups, with_input_group
+"""
+
+from _m5 import core
```
powerjg, line 55: This line is your problem

src/python/gem5/resources/suite.py
```diff
@@ -0,0 +1,123 @@
...
+    def __iter__(self) -> "Suite":
+        return self
+
+    def __next__(self) -> Workload:
+        try:
+            workload = self.workload_obj_list[self.index]
+        except IndexError:
+            raise StopIteration
+        self.index += 1
+        return workload
```
powerjg: I think the following would work:
```suggestion
        for workload in self.workload_obj_list:
            yield workload
```
I think that if you make `__iter__` a generator then you don't need `__next__`. I could be wrong, this is a bit confusing.

src/python/gem5/resources/suite.py
```diff
@@ -0,0 +1,123 @@
...
+from _m5 import core
+
+from typing import Dict, Any, List, Optional, Set
+
+
+class Suite:
+    def __init__(
+        self,
+        resource_id: str,
+        resource_version: str,
```
powerjg: Why does a suite require these parameters?  Resources don't require these parameters; they are optional.

Harshil2107: We did that because we need the id and version to use obtain resources. We were thinking that rather than making a `suite.py` we would add `SuiteResource` to the `resource.py` like other resources so that class constructor will not neet id and version.

src/python/gem5/resources/suite.py
powerjg: Don't forget to address this. It's still valid even though github says "outdated"

Harshil2107:
The above tests will not run properly as workload changes are not merged yet, we are waiting for the workload changes to get merged first before pushing any other changes.

configs/test.py
```diff
@@ -0,0 +1,28 @@
+from gem5.resources.resource import SuiteResource
```
powerjg: Needs license block

configs/test.py
```diff
@@ -0,0 +1,28 @@
...
+    ],
+}
+
+s = SuiteResource(
+    workloads=suite["workloads"],
+    resource_version=suite["resource_version"],
+)
+
+for workload in s.with_input_group("valorant"):
+    print(1)
```
powerjg: What is this print statement for?

configs/test.py
```diff
@@ -0,0 +1,28 @@
...
+
+suite = {
+    "id": "id",
+    "resource_version": "1.0.0",
+    "gem5_versions": ["23.1"],
+    "workloads": [
+        {
+            "id": "riscv-ubuntu-20.04-boot",
+            "resource_version": "1.0.0",
+            "input_group": ["hashiru", "valorant"],
```
powerjg: Can we choose different things here? I suggest "big" and "small"

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +556,77 @@ def _get_weights_and_simpoints_from_file(
...
+        description: Optional[str] = None,
+        source: Optional[str] = None,
+    ) -> None:
+        super().__init__(
+            local_path=local_path,
+            description=description,
+            source=source,
+            resource_version=resource_version,
+        )
+        print("in constructor")
```
powerjg: Remove debug statements.

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +556,77 @@ def _get_weights_and_simpoints_from_file(
...
+    def __iter__(self) -> "SuiteResource":
+        return self
+
+    def __next__(self) -> "Workload":
+        try:
+            workload = self.workload_obj_list[self.index]
+        except IndexError:
+            raise StopIteration
+        self.index += 1
+        return workload
```
powerjg: As previously suggested, I think you can do this:
```suggestion
    def __iter__(self) -> "SuiteResource":
        for workload in self.workload_obj_list:
            yield workload
```

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +556,77 @@ def _get_weights_and_simpoints_from_file(
...
+            description=self._description,
+            source=self._source,
+        )
+
+    def list_input_groups(self) -> Set[str]:
+        input_group_set = set()
+        for workload in self.workloads:
+            for input_group in workload["input_group"]:
+                input_group_set.add(input_group)
+        return input_group_set
```
powerjg:
```suggestion
        return {group for workload in self.workloads for group in workload["input_group"]}
```

configs/test.py
powerjg, line 1: Maybe this file isn't meant to be included?

configs/test.py
Harshil2107: This was just for checking if it gets the correct number of workloads, this is not a proper test.

src/python/gem5/resources/resource.py
Harshil2107: This change is done, we haven't pushed it yet, we checked and it works the same way.

configs/test.py
Harshil2107: yeah, this file isn't meant to be included, we will rebase everything to a single commit before converting this to a normal pr from draft.

configs/test.py
Harshil2107: This file will be removed, it was just for checking our initial implementation.

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +555,70 @@ def _get_weights_and_simpoints_from_file(
...
+class SuiteResource(AbstractResource):
+    def __init__(
+        self,
+        workloads: List[Dict[str, Any]],
+        local_path: Optional[str] = None,
+        resource_version: Optional[str] = None,
+        description: Optional[str] = None,
+        source: Optional[str] = None,
+        **kwargs,
+    ) -> None:
```
BobbyRBruce, line 659: This is missing a doc-string. The dictionary explaining the `workloads` is necessary.

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +555,70 @@ def _get_weights_and_simpoints_from_file(
...
+        super().__init__(
+            local_path=local_path,
+            description=description,
+            source=source,
+            resource_version=resource_version,
+        )
+        self.resource_version = resource_version
+        self.workloads = workloads
+        self.workload_obj_list = []
+        self.index = 0
```
BobbyRBruce: Maybe all these should be private?

```suggestion
        self._resource_version = resource_version
        self._workloads = workloads
        self._workload_obj_list = []
        self._index = 0
```

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +555,70 @@ def _get_weights_and_simpoints_from_file(
...
+        self._local_path = local_path
+        self._description = description
+        self._source = source
+        self._obtain_workloads_list()
+
+    def _obtain_workloads_list(self) -> None:
+        for workload in self.workloads:
+            self.workload_obj_list.append(
+                WorkloadResource(
```
BobbyRBruce: Is there a reason we don't use "obtain_resource" instead? This isn't getting all the information about a resource we may need.

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +555,70 @@ def _get_weights_and_simpoints_from_file(
...
+                    workload["id"],
+                    resource_version=workload["resource_version"],
+                )
+            )
+
+    def __iter__(self) -> "SuiteResource":
```
BobbyRBruce: Doesn't this return a Workload?

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +555,70 @@ def _get_weights_and_simpoints_from_file(
...
+        for workload in self.workload_obj_list:
+            yield workload
+
+    def __len__(self):
+        # return len(self.workload_obj_list)
+        return len(self.workload_obj_list)
+
+    def with_input_group(self, input_group: str) -> "SuiteResource":
```
BobbyRBruce, line 703: Doc-string for all public functions at least.

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +555,70 @@ def _get_weights_and_simpoints_from_file(
...
+            workloads=filtered_workloads,
+            description=self._description,
+            source=self._source,
+        )
+
+    def list_input_groups(self) -> Set[str]:
+        return {
+            group
+            for workload in self.workloads
+            for group in workload["input_group"]
```
BobbyRBruce: How does this work if you haven't obtained the resources?

BobbyRBruce (changes requested):

src/python/gem5/resources/resource.py
Harshil2107: Yes, we should be using `obtain_resource`. For someone who wants to create their own `SuiteResource` how do we handle the call to `obtain_resource`? I can think of 2 possible solutions:

- We ask users to create their own `resources.json` and pass it in as a client in ontain_resources, We can also pass the `clients` parameter to the `SuiteResource` constructor so that users can define which clients they want.

- We can make getters and setters for the `workload_obj_list` to allow users to directly edit the workload list. This needs some sort of way to not call `obtain_resources` when the user is using their resources. 

I am leaning towards the first one.

src/python/gem5/resources/resource.py
Harshil2107: The return type should be a `Generator[YieldType, SendType, ReturnType]` (documentation: https://docs.python.org/3/library/typing.html#typing.Generator). In our case, it would be `Generator["WorkloadResource", None, None]`.We will change it to the correct type.

kunpai:
TODO: Rebase all commits into one commit before merging

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,122 @@ def _get_weights_and_simpoints_from_file(
...
+        return SuiteResource(
+            local_path=self._local_path,
+            resource_version=self._resource_version,
+            workloads=filtered_workloads,
+            description=self._description,
+            source=self._source,
+            workload_obj_list=filtered_workload_obj_list,
+        )
+
+    def list_input_groups(self) -> Set[str]:
```
powerjg: How about "get_input_groups" since it returns a `Set` not a `List`

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,122 @@ def _get_weights_and_simpoints_from_file(
...
+        }
+
+    def get_all_workloads(
+        self,
+    ) -> Generator[Tuple[str, str, List[str], "WorkloadResource"], None, None]:
+        """
+        Returns a list of all workloads in the suite in the following format:
+        :returns: A list of all workloads in the suite.
+        """
+        for i in range(len(self._workloads)):
```
powerjg:
```suggestion
        for info, workload in zip(self._workloads, self._workload_obj_list):
```

powerjg (commented):
A couple more small things :)

src/python/gem5/resources/resource.py
```diff
@@ -672,13 +692,24 @@ def get_all_workloads(
...
         """
-        for info, workload in zip(self._workloads, self._workload_obj_list):
-            yield (
-                info["id"],
-                info["resource_version"],
-                info["input_group"],
-                workload,
+        if self._workloads == None:
+            warn(
+                "Optional paramter 'workloads' which contains resource id, version,"
```
kunpai: 'Optional parameter'

tests/pyunit/stdlib/resources/pyunit_suite_checks.py
```diff
@@ -47,6 +49,70 @@
...
+        """
+        f = io.StringIO()
+        with contextlib.redirect_stderr(f):
+            for workload, expected_workload in zip(
+                self.SuiteResource.get_all_workloads(),
+                [self.workload1, self.workload2],
+            ):
+                self.assertEqual(workload, ("", "", "", expected_workload))
+        self.assertTrue(
+            "Optional paramter 'workloads' which contains resource id, version,"
```
kunpai: "Optional parameter"

configs/example/gem5_library/riscvmatched-microbenchmark-suite.py
```diff
@@ -0,0 +1,73 @@
...
+
+# list all the microbenchmarks present in the suite
+print("Microbenchmarks present in the suite:")
+print("====================================")
+for (
+    id,
+    resource_version,
+    input_group,
+    workload,
+) in microbenchmarks.get_all_workloads():
```
BobbyRBruce: I thought we had this setup so a suite was an iterator of Workloads. Shouldn't this be `for resource in microbenchmarks:`?

configs/example/gem5_library/riscvmatched-microbenchmark-suite.py
```diff
@@ -0,0 +1,73 @@
...
+from gem5.isas import ISA
+from gem5.utils.requires import requires
+
+requires(isa_required=ISA.RISCV)
+
+# instantiate the riscv matched board with default parameters
+board = RISCVMatchedBoard()
+
+# obtain the RISC-V Vertical Microbenchmarks
+microbenchmarks = obtain_resource("riscv-vertical-microbenchmarks")
```
BobbyRBruce, line 48: This suite doesn't exist yet, right?

configs/example/gem5_library/riscvmatched-microbenchmark-suite.py
BobbyRBruce, line 1: What is the purpose of this example. I'd suggest adding a doc-string explaining.

Also, it should have a test which regularly runs to ensure it is still a functioning example, such as that in https://github.com/gem5/gem5/tree/48a40cf2f5182a82de360b7efa497d82e06b1631/tests/gem5/gem5_library_example_tests

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,153 @@ def _get_weights_and_simpoints_from_file(
...
+    IDs and versions.
+
+    Each workload in a suite is used to create a `WorkloadResource` object.
+    These objects are stored in a list and can be iterated over.
+    """
+
+    def __init__(
+        self,
+        workloads: Optional[List[Dict[str, Any]]] = None,
+        local_path: Optional[str] = None,
```
BobbyRBruce: `local_path` here confuses me. Surely suites don't have one? It's only resources that do.

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,153 @@ def _get_weights_and_simpoints_from_file(
...
+        resource_version: Optional[str] = None,
+        description: Optional[str] = None,
+        source: Optional[str] = None,
+        workload_obj_list: Optional[List["WorkloadResource"]] = None,
+        **kwargs,
+    ) -> None:
+        """
+        :param workloads: A list of dictionaries containing the workload
+        :param _workload_obj_list: A list of `WorkloadResource` objects
```
BobbyRBruce: There's a leading underscore here before `workload_obj_list`

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,153 @@ def _get_weights_and_simpoints_from_file(
...
+    """
+
+    def __init__(
+        self,
+        workloads: Optional[List[Dict[str, Any]]] = None,
+        local_path: Optional[str] = None,
+        resource_version: Optional[str] = None,
+        description: Optional[str] = None,
+        source: Optional[str] = None,
+        workload_obj_list: Optional[List["WorkloadResource"]] = None,
```
BobbyRBruce: Is there a reason this is optional? What does `None` mean which isn't the same as an empty list?

Ditto for `workloads`.

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,153 @@ def _get_weights_and_simpoints_from_file(
...
+    run on a board. It contains a list of workloads to run, along with their
+    IDs and versions.
+
+    Each workload in a suite is used to create a `WorkloadResource` object.
+    These objects are stored in a list and can be iterated over.
+    """
+
+    def __init__(
+        self,
+        workloads: Optional[List[Dict[str, Any]]] = None,
```
BobbyRBruce: Why do we have both `workloads` and `workload_obj_list`. Surely `workload_obj_list` encapsulates all the data in `workloads`? I think we should just have `workload_obj_list` and rename it as `workloads`. I don't know why the current design is like this but, as a user, how am i supposed to know which of these parameters to use to specify the workloads in a suite?

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,153 @@ def _get_weights_and_simpoints_from_file(
...
+            description=description,
+            source=source,
+            resource_version=resource_version,
+        )
+
+    def __iter__(self) -> Generator["WorkloadResource", None, None]:
+        """
+        Returns a generator that iterates over the workloads in the suite.
+
+        :returns: A generator that iterates over the workloads in the suite.
```
BobbyRBruce: Technically this doesn't return, it `yields`. So I think the syntax is `:yields: ...`

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,153 @@ def _get_weights_and_simpoints_from_file(
...
+                " and input group not found."
+            )
+            return SuiteResource(
+                local_path=self._local_path,
+                resource_version=self._resource_version,
+                workloads=self._workloads,
+                description=self._description,
+                source=self._source,
+                workload_obj_list=[],
+            )
```
BobbyRBruce: Related to what I discuss above, this optional parameter is weird. I'd add the resource id and version to the `WorkloadResource` and work from there.

Also, the Input group is in the suite, right? Not the workload itself, so we can still do this lookup.

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,153 @@ def _get_weights_and_simpoints_from_file(
...
+    def __init__(
+        self,
+        workloads: Optional[List[Dict[str, Any]]] = None,
+        local_path: Optional[str] = None,
+        resource_version: Optional[str] = None,
+        description: Optional[str] = None,
+        source: Optional[str] = None,
+        workload_obj_list: Optional[List["WorkloadResource"]] = None,
+        **kwargs,
+    ) -> None:
```
BobbyRBruce, line 659: Ok, having read through the rest of this code, I think one of the reasons you have this `workload` argument is you need a way of mapping `input_groups` to Workloads. Why not do this:

```py
workloads: Dict[ResourceWorkload, Set[str]]
```

Just a simple dictionary. The key being the Workload, and it's value being a set of input_groups. Then everything can be in one argument.

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,153 @@ def _get_weights_and_simpoints_from_file(
...
+            return set()
+        return {
+            group
+            for workload in self._workloads
+            for group in workload["input_group"]
+        }
+
+    def get_all_workloads(
+        self,
+    ) -> Generator[Tuple[str, str, List[str], "WorkloadResource"], None, None]:
```
BobbyRBruce: This isn't necessary, the class already implements `__iter__` which does exactly this.

Also, `get_all_workloads` sounds like it should return a list or set of workloads, not be a generator (in this case I'd expect it be called something like `next_workload`.)

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,153 @@ def _get_weights_and_simpoints_from_file(
...
+
+    def __len__(self):
+        """
+        Returns the number of workloads in the suite.
+
+        :returns: The number of workloads in the suite.
+        """
+        return len(self._workload_obj_list)
+
+    def with_input_group(self, input_group: str) -> "SuiteResource":
```
BobbyRBruce, line 703: I think there should be an exception thrown here if an input_group is specified which doesn't exist for this workload. Quite simple logic:

```py
if input_group not in get_input_groups():
    raise Exception("...")
```

I'd have the exception say that the value is invalid and then list the valid values (i.e., `get_input_groups()`).

BobbyRBruce (changes requested):
I didn't go over the tests very much as i had some feedback on the broader APIs and funcitonalities

configs/example/gem5_library/riscvmatched-microbenchmark-suite.py
kunpai: Yes it does not, yet

configs/example/gem5_library/riscvmatched-microbenchmark-suite.py
Harshil2107: Yes, that also works but the above function returns a tuple with the workload id, version, input groups and the WorkloadResource object.

src/python/gem5/resources/resource.py
Harshil2107: As you specified below, the `workloads` field is mainly to keep track of the id, version, and input_group and link it to the workload object.

src/python/gem5/resources/resource.py
Harshil2107: Should we return a list of tuples or just change the function name?

powerjg: I think this function serves a different purpose than `__iter__` since it also returns the id, version, and inputs. 

I also think we're doing a bit of bikeshedding here.

But, to argue about the color of the shed... One option would be to make this the implementation of `__iter__` and to have this function just return the list of workloads.

IMO, the decision here doesn't really matter as long as we pick something that is consistent and we document it.

kunpai:
> TODO: Rebase all commits into one commit before merging

Has been done :)

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,122 @@ def _get_weights_and_simpoints_from_file(
...
+    def __init__(
+        self,
+        workload_obj_list: List[Tuple["WorkloadResource", str]] = [],
+        resource_version: Optional[str] = None,
+        description: Optional[str] = None,
+        source: Optional[str] = None,
+        **kwargs,
+    ) -> None:
+        """
+        :param workloads: A list of dictionaries containing the workload
```
BobbyRBruce: The `workloads` param doc-string is no longer needed here.

src/python/gem5/resources/resource.py
```diff
@@ -554,6 +565,122 @@ def _get_weights_and_simpoints_from_file(
...
+    run on a board. It contains a list of workloads to run, along with their
+    IDs and versions.
+
+    Each workload in a suite is used to create a `WorkloadResource` object.
+    These objects are stored in a list and can be iterated over.
+    """
+
+    def __init__(
+        self,
+        workload_obj_list: List[Tuple["WorkloadResource", str]] = [],
```
BobbyRBruce: A better solution `workloads: Dict["WorkloadResource", Set[str]] = {}`.

Improvements:

1. Rename `workload_obj_list` to `workloads`.
2. Removes the confusing "List of tuples".
3. Allows for quick association of workload to group_label.

src/python/gem5/resources/resource.py
BobbyRBruce: Just get rid of this function. It's weird. If we need the id you can just do `workload.get_id() for workload in suite`. I really don't see any advantage of this function. The workloads which return via the iterator contain all the information you'd ever require.

src/python/gem5/resources/resource.py
```diff
@@ -35,10 +35,21 @@
...
+    Tuple,
+    List,
+    Any,
+    Set,
+    Generator,
+)
 
 from .client import get_resource_json_obj
 
+
```
BobbyRBruce: Remove this newline.

BobbyRBruce (changes requested):

Harshil2107:
I made an error while trying to sync the new develop branch with my working branch, I will try to fix it in the morning.

configs/example/gem5_library/riscvmatched-microbenchmark-suite.py
BobbyRBruce: This will fail now. You need to change it.

configs/example/gem5_library/riscvmatched-microbenchmark-suite.py
BobbyRBruce: You added the doc-string here but i think it should be added as a test, as I suggested.

src/python/gem5/resources/resource.py
```diff
@@ -844,6 +980,7 @@ def obtain_resource(
     # Once we know what AbstractResource subclass we are using, we create it.
     # The fields in the JSON object are assumed to map like-for-like to the
     # subclass contructor, so we can pass the resource_json map directly.
+
```
BobbyRBruce: There's a weird added line here. Remove it.

configs/example/gem5_library/riscvmatched-microbenchmark-suite.py
BobbyRBruce: Can we add it to make sure this script works as intended?

tests/pyunit/stdlib/resources/pyunit_suite_checks.py
```diff
@@ -0,0 +1,170 @@
...
+    @classmethod
+    @patch(
+        "gem5.resources.client.clientwrapper",
+        new=ClientWrapper(mock_config_json),
+    )
+    def setUpClass(cls):
+        cls.workload1 = obtain_resource("simple-workload-1")
+        cls.workload2 = obtain_resource("simple-workload-2")
+        cls.SuiteResource = SuiteResource(
+            workload_obj_list={cls.workload1: set(), cls.workload2: set()}
```
BobbyRBruce: This will fail, no? It's called `Workload` now.

BobbyRBruce (changes requested):

BobbyRBruce (approved):

PR #198, opened 2023-08-18
misc: Add DRAMSys tests to our weekly tests

mkjost0:
This adds the DRAMSys tests to our weekly-tests.yaml file

Change-Id: Ieb7903a3a7ffae6359b3de5f66e1dd65eb51fc80

### DIFF
--- a/.github/workflows/weekly-tests.yaml
+++ b/.github/workflows/weekly-tests.yaml
@@ -77,3 +77,34 @@ jobs:
         path: output.zip
         retention-days: 7
     - run: echo "This job's status is ${{ job.status }}."
+
+  dramsys-tests:
+    runs-on: [self-hosted, linux, x64, build]
+    container: gcr.io/gem5-test/ubuntu-22.04_all-dependencies:latest
+    timeout-minutes: 4320 # 3 days
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # Scheduled workflows run on the default branch by default. We
+          # therefore need to explicitly checkout the develop branch.
+          ref: develop
+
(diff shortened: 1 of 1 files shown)


BobbyRBruce:
The issue is you aren't recursively getting the submodules. The README file for DRAMSys was incorrect. I fix this in #202. Update the test to incorporate this change and it should start working.

mkjost0:
@BobbyRBruce I tested that out and the test runs successfully now, thanks! I'll make the change here then mark it as ready for review

.github/workflows/weekly-tests.yaml
```diff
@@ -77,3 +77,34 @@ jobs:
...
+        run: |
+          git clone https://github.com/tukl-msd/DRAMSys DRAMSys
+          cd DRAMSys
+          git checkout -b gem5 09f6dcbb91351e6ee7cadfc7bc8b29d97625db8f
+          git submodule update --init --recursive
+
+      # gem5 is built separately because it depends on the DRAMSys library
+      - name: Build gem5
+        working-directory: ${{ github.workspace }}
+        run: scons build/ALL/gem5.opt -j4
```
BobbyRBruce: I'm not sure how we do this elsewhere in out workflow files but shouldn't we use something like `nproc` here?

BobbyRBruce (changes requested):

.github/workflows/weekly-tests.yaml
mkjost0: Ahh, yeah we do, I can update that! I think it was something I forgot to change when I was testing

BobbyRBruce (approved):

PR #199, opened 2023-08-18
WIP: Add GPU tests to weekly.yaml file

mkjost0:
This change adds all the GPU tests to our weekly tests, though some tests aren't fully working yet, and it uses a temporary docker container that needs to be replaced.

The current issues that need to be addressed within the GPU test implementation is that the DNNMark tests currently fail due to the setting up of the cachefiles, the Pannotia tests fail to make gem5-fusion due to not finding gem5/m5ops.h, the current docker image being used is one I created locally and should be updated to be included within gem5, and we should upload all the resources used so that we don't need to clone gem5 resources within these tests.

Change-Id: I7507d2fe00b652591c4aaef4897829b19b73a182

### DIFF
--- a/.github/workflows/gpu-tests.yaml
+++ b/.github/workflows/gpu-tests.yaml
@@ -1,0 +1,82 @@
+# This workflow runs all of the very-long tests within main.py
+
+name: Weekly Tests
+
+on:
+  # Runs every Sunday from 7AM UTC
+  schedule:
+    - cron:  '00 7 * * 6'
+  # Allows us to manually start workflow for testing
+  workflow_dispatch:
+
+jobs:
+  name-artifacts:
+    runs-on: ubuntu-latest
+    outputs:
+      build-name: ${{ steps.artifact-name.outputs.name }}
+    steps:
+    - uses: actions/checkout@v2
+    - id: artifact-name
+      run: echo "name=$(date +"%Y-%m-%d_%H.%M.%S-")" >> $GITHUB_OUTPUT
+
+  build-gem5:
+    strategy:
+      matrix:
+        image: [ALL, GCN3_X86]
+        # this allows us to pass additional command line parameters
+        # the default is to add -j $(nproc), but some images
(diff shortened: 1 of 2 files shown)


mattsinc:
@mkjost0 I am away for the next few days, I will review here after Monday (and will try to do sooner if possible).

mattsinc:
@mkjost0 do you want feedback on how to get m5ops and such added?  Or do you want me to just review what is working as is?

mattsinc (commented):
I don't see anything specifically standing out as wrong here, but I am confused why the m5ops work in some cases but not others.  You mention the include failing -- is this an issue with where the tests are being run from then?  Previously we had assumed the tests were run from a place where the m5ops folder existed -- if that is not the case then maybe the relative path needs to be updated?

Likewise, for the DNNMark cachefiles, what exactly is the issue?

I am ok to approve if it improves coverage, but these seem like the bigger issues.

mkjost0:
For m5ops, I'm not sure what the issue is, but when I was having a conversation with Jason about it the other day we suspected that some of the issues might fix if we upload everything to gem5 resources and stop building things within the workflow, which is something I'm working on testing.  However, I can try double-checking the paths as well; I think the way it's currently set up should include m5ops but maybe I'm wrong about that.

For DNNMark, the issue is that files located at the root couldn't be opened while the tests were running, which I suspected had to do with mounting the cache files directory. I was a bit more lost with this problem, and we didn't have any immediate ideas as to how to approach it here.  

As for the state of these tests, we're working on getting any necessary binaries uploaded to gem5-resources to hopefully remove the need to clone the repository within these tests, and once we have a fix for the m5ops and DNNMark issues we'll move this out of the draft state where it can actually be reviewed to be run alongside our weekly tests.

mkjost0:
Note: I am closing this pull request since I opened new drafts under #256 #257 #258 and #259 that will allow for us to merge the tests as we get each section working.

PR #200, opened 2023-08-18
mem: Atomic ops to same address

BujSet:
Augmenting the DataBlock class with a change log structure to record the effects of atomic operations on a data block and service these changes if the atomic operations require return values.

Although the operations are atomic, the coalescer need not send unique memory requests for each operation. Atomic operations within a wavefront to the same address are now coalesced into a single memory request. The response of this request carries all the necessary information to provide the requesting lanes unique values as a result of their individual atomic operations. This helps reduce contention for request and response queues in simulation.

Previously, only the final value of the datablock after all atomic ops to the same address was visible to the requesting waves. This change corrects this behavior by allowing each wave to see the effect of this individual atomic op is a return value is necessary.

### DIFF
--- a/src/mem/ruby/common/DataBlock.cc
+++ b/src/mem/ruby/common/DataBlock.cc
@@ -51,9 +51,19 @@ namespace ruby
 
 DataBlock::DataBlock(const DataBlock &cp)
 {
-    m_data = new uint8_t[RubySystem::getBlockSizeBytes()];
-    memcpy(m_data, cp.m_data, RubySystem::getBlockSizeBytes());
+    uint8_t *block_update;
+    size_t block_bytes = RubySystem::getBlockSizeBytes();
+    m_data = new uint8_t[block_bytes];
+    memcpy(m_data, cp.m_data, block_bytes);
     m_alloc = true;
+    // If this data block is involved in an atomic operation, the effect
+    // of applying the atomic operations on the data block are recorded in
+    // m_atomicLog. If so, we must copy over every entry in the change log
+    for (size_t i = 0; i < cp.m_atomicLog.size(); i++) {
+        block_update = new uint8_t[block_bytes];
+        memcpy(block_update, cp.m_atomicLog[i], block_bytes);
+        m_atomicLog.push_back(block_update);
+    }
 }
 
 void
@@ -73,7 +83,20 @@ DataBlock::clear()
 bool
 DataBlock::equal(const DataBlock& obj) const
 {
-    return !memcmp(m_data, obj.m_data, RubySystem::getBlockSizeBytes());
+    size_t block_bytes = RubySystem::getBlockSizeBytes();
+    // Check that the block contents match
+    if (memcmp(m_data, obj.m_data, block_bytes)) {
+        return false;
+    }
+    if (m_atomicLog.size() != obj.m_atomicLog.size()) {
+        return false;
+    }
+    for (size_t i = 0; i < m_atomicLog.size(); i++) {
+        if (memcmp(m_atomicLog[i], obj.m_atomicLog[i], block_bytes)) {
+            return false;
+        }
+    }
+    return true;
 }
 
 void
@@ -92,7 +115,7 @@ DataBlock::atomicPartial(const DataBlock &dblk, const WriteMask &mask)
     for (int i = 0; i < RubySystem::getBlockSizeBytes(); i++) {
         m_data[i] = dblk.m_data[i];
     }
-    mask.performAtomic(m_data);
+    mask.performAtomic(m_data, m_atomicLog);
 }
 
 void
@@ -105,6 +128,28 @@ DataBlock::print(std::ostream& out) const
             << "0x" << (int)m_data[i] << " " << std::setfill(' ');
     }
     out << std::dec << "]" << std::flush;
+}
+
+int
+DataBlock::numAtomicLogEntries() const
+{
+    return m_atomicLog.size();
+}
+uint8_t*
+DataBlock::popAtomicLogEntryFront()
+{
+    assert(m_atomicLog.size() > 0);
+    auto ret = m_atomicLog.front();
+    m_atomicLog.pop_front();
+    return ret;
+}
+void
+DataBlock::clearAtomicLogEntries()
+{
+    for (auto log : m_atomicLog) {
+        delete [] log;
+    }
+    m_atomicLog.clear();
 }
 
 const uint8_t*
@@ -137,7 +182,18 @@ DataBlock::setData(PacketPtr pkt)
 DataBlock &
 DataBlock::operator=(const DataBlock & obj)
 {
-    memcpy(m_data, obj.m_data, RubySystem::getBlockSizeBytes());
+    uint8_t *block_update;
+    size_t block_bytes = RubySystem::getBlockSizeBytes();
+    // Copy entire block contents from obj to current block
+    memcpy(m_data, obj.m_data, block_bytes);
+    // If this data block is involved in an atomic operation, the effect
+    // of applying the atomic operations on the data block are recorded in
+    // m_atomicLog. If so, we must copy over every entry in the change log
+    for (size_t i = 0; i < obj.m_atomicLog.size(); i++) {
+        block_update = new uint8_t[block_bytes];
+        memcpy(block_update, obj.m_atomicLog[i], block_bytes);
+        m_atomicLog.push_back(block_update);
+    }
     return *this;
 }
 
--- a/src/mem/ruby/common/DataBlock.hh
+++ b/src/mem/ruby/common/DataBlock.hh
@@ -44,6 +44,7 @@
 #include <inttypes.h>
 
 #include <cassert>
+#include <deque>
 #include <iomanip>
 #include <iostream>
 
@@ -71,6 +72,12 @@ class DataBlock
     {
         if (m_alloc)
             delete [] m_data;
+
+        // If data block involved in atomic
+        // operations, free all meta data
+        for (auto log : m_atomicLog) {
+            delete [] log;
+        }
     }
 
     DataBlock& operator=(const DataBlock& obj);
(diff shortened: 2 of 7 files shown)


powerjg:
I know it's different, but I think that #177 is quite related. Let's make sure there are no conflicts between these changes.

BujSet:
Forgot to retain the changeId in the commit message. Added it back.

BujSet:
> I know it's different, but I think that #177 is quite related. Let's make sure there are no conflicts between these changes.

Looking through the files changed, I don't think the there should be any merge-conflict issues. I haven't tested this change with the CHI protocol, but I suspect CHI has the same issue as the VIPER protocol (where waves that request atomic updates to the same address get the same response instead of a unique responses after each atomic update). Will test this out after [#177](https://github.com/gem5/gem5/pull/177/) gets merged into develop (to prevent conflicts now).

mattsinc:
@BujSet I am away for the next few days, and while I reviewed it at AMD previously, I will review here again after Monday.

src/mem/ruby/common/DataBlock.cc
```diff
@@ -73,7 +85,17 @@ DataBlock::clear()
 bool
 DataBlock::equal(const DataBlock& obj) const
 {
-    return !memcmp(m_data, obj.m_data, RubySystem::getBlockSizeBytes());
+    size_t block_bytes = RubySystem::getBlockSizeBytes();
+    // Check that the block contents match
+    if (memcmp(m_data, obj.m_data, block_bytes))
+        return false;
```
mattsinc, line 89: minor nit: please add parentheses around these if statements.  It will bite us in the butt later if we don't.

Technically not required though.

src/mem/ruby/system/GPUCoalescer.cc
```diff
@@ -556,23 +556,46 @@ GPUCoalescer::hitCallback(CoalescedRequest* crequest,
     //
     // MUST AD DOING THIS FOR EACH REQUEST IN COALESCER
```
mattsinc: maybe not part of this commit, but AD --> ADD?

src/mem/ruby/system/GPUCoalescer.cc
```diff
@@ -556,23 +556,46 @@ GPUCoalescer::hitCallback(CoalescedRequest* crequest,
...
+                        //
+                        // Log entry contains the old value before the current
+                        // atomic operation occurred.
+                        assert(data.numAtomicLogEntries());
+                        log = data.popAtomicLogEntryFront();
+                        assert(data.numAtomicLogEntries() >= 0);
+                        pkt->setData(&log[offset]);
+                        delete [] log;
+                        log = nullptr;
+                    }
```
mattsinc: is there a case where the type is an atomic return but isAtomicOp returns false?

mattsinc (changes requested):
I don't see anything overlapping relative to the other patch, but I do have some small changes that would be helpful here.

src/mem/ruby/common/DataBlock.cc
```diff
@@ -51,9 +51,21 @@ namespace ruby
 
 DataBlock::DataBlock(const DataBlock &cp)
 {
+    uint8_t *block_update;
+    size_t block_bytes = RubySystem::getBlockSizeBytes();
+    m_data = new uint8_t[block_bytes];
+    memcpy(m_data, cp.m_data, block_bytes);
     m_data = new uint8_t[RubySystem::getBlockSizeBytes()];
     memcpy(m_data, cp.m_data, RubySystem::getBlockSizeBytes());
```
giactra: This is a duplicate

src/mem/ruby/common/DataBlock.cc
```diff
@@ -107,6 +129,30 @@ DataBlock::print(std::ostream& out) const
...
+{
+    assert(m_atomicLog.size() > 0);
+    auto ret = m_atomicLog.front();
+    m_atomicLog.pop_front();
+    return ret;
+}
+void
+DataBlock::clearAtomicLogEntries()
+{
+    if (m_atomicLog.size()) {
```
giactra: This is probably not needed as we would just skip the for loop

src/mem/ruby/common/DataBlock.hh
```diff
@@ -71,6 +72,14 @@ class DataBlock
     {
         if (m_alloc)
             delete [] m_data;
+
+        // If data block involved in atomic
+        // operations, free all meta data
+        if (m_atomicLog.size()) {
```
giactra: Ditto

src/mem/ruby/common/DataBlock.hh
```diff
@@ -94,6 +106,9 @@ class DataBlock
     void alloc();
     uint8_t *m_data;
     bool m_alloc;
+
+    // Tracks block changes when atomic ops are applied
+    std::deque<uint8_t*> m_atomicLog;
```
giactra, line 109: Wouldn't it make sense to use a unique_ptr instead?

src/mem/ruby/common/WriteMask.hh
```diff
@@ -222,26 +222,15 @@ class WriteMask
...
-            AtomicOpFunctor *fnctr = mAtomicOp[i].second;
-            (*fnctr)(p);
-        }
-    }
+    /*
+     * Performs atomic operations on the data block pointed to by p. The
+     * atomic operations to perform are in the vector mAtomicOp. The
+     * effect of each atomic operation is pushed to the atomicChangeLog
+     * so that each thread in a wavefront may see the results of their
+     * specific atomic operation.
```
giactra: This seems a GPU specific comment

src/mem/ruby/common/WriteMask.cc
```diff
@@ -55,5 +55,27 @@ WriteMask::print(std::ostream& out) const
...
+    uint8_t *block_update;
+    // Here, operations occur in FIFO order from the mAtomicOp
+    // vector. This is done to match the ordering of packets
+    // that was seen when the initial coalesced request was created.
+    for (int i = 0; i < mAtomicOp.size(); i++) {
+        // Save the old value of the data block in case a
+        // return value is needed
+        block_update = new uint8_t[mSize];
+        std::memcpy(block_update, p, mSize);
+        log.push_back(block_update);
```
giactra, line 72: I am wondering whether you could push most of the bookeping to the AtomicOpFunctor....

BujSet: I think it could be done. To do this though, the response messages would have to carry the WriteMask back up the hierarchy, since the vector of functors are held by the WriteMask itself. Since the DataBlock is already a part of the response message, I opted to put the logging structure there. Would it make more sense to use the AtomicOpFunctors? I think putting it there might make it a little clunky: in the GPUCoalescer, we would read from the return WriteMasks atomicVector of functors instead of the from the datablock.

src/mem/ruby/system/GPUCoalescer.cc
BujSet: I believe that that case should never happen, but I vaguely remember seeing this pop up on some of the tester benchmarks, but I'm failing to remember which ones. Will test this out and see if the if-check is necessary.

BujSet: Looks like I remembered wrong, and I did not see a case when the packet->isAtomicOp() needs after already seeing that the ruby request type is one of the atomic types. Put in some assert statements so that this assumption is clear moving forward.

BujSet:
@abmerop Tagging you here as well.

abmerop (commented):
This looks fine to me, but of course I have seen multiple iterations of this code.

src/mem/ruby/common/DataBlock.hh
abmerop: The DataBlock can be copied so doesn't this just make things more complex?

BobbyRBruce:
@mattsinc : Is this ok with you? You had some suggestions which I believe have been incorporated. If so I'll merge this.

mattsinc:
I was waiting for Giacomo to chime in since Giacomo was the one requesting
the changes.

mattsinc:
> 

@giactra just checking if you are ok with the changes @BujSet made?

src/mem/ruby/common/WriteMask.cc
giactra: > I think it could be done. To do this though, the response messages would have to carry the WriteMask back up the hierarchy, since the vector of functors are held by the WriteMask itself.
Since the DataBlock is already a part of the response message, I opted to put the logging structure there. Would it make more sense to use the AtomicOpFunctors? I think putting it there might make it a little clunky: in the GPUCoalescer, we would read from the return WriteMasks atomicVector of functors instead of the from the datablock.

Apologies for the late reply, I was on a small break from work. Yeah I understand, and binding the log of changes to the block itself makes a lot of sense. My original message was just meant to move the three lines to the "prologue" of the functor to keep the WriteMask cleaner

        block_update = new uint8_t[mSize];
        std::memcpy(block_update, p, mSize);
        log.push_back(block_update);

But I realise it is a little save...

giactra (approved):
LGTM

mattsinc (approved):
Thanks @giactra !  @BujSet you will need to rebase before you can merge this in, FYI.

BobbyRBruce:
> Thanks @giactra ! @BujSet you will need to rebase before you can merge this in, FYI.

Do they? It's clean merge from what I can see.

mattsinc:
> > Thanks @giactra ! @BujSet you will need to rebase before you can merge this in, FYI.
> 
> Do they? It's clean merge from what I can see.

Interesting, it is clean now but when I sent that, I saw that rebasing to the head of develop was required ...

BobbyRBruce:
> > > Thanks @giactra ! @BujSet you will need to rebase before you can merge this in, FYI.
> > 
> > 
> > Do they? It's clean merge from what I can see.
> 
> Interesting, it is clean now but when I sent that, I saw that rebasing to the head of develop was required ...

Huh, I'll look out for that. No worries, I just thought there was some reason it needed rebasing.

Since this now has 2 approvals, and I don't see any outstanding comments to address, I'll merge it,

Thank you for all great work guys!

PR #203, opened 2023-08-18
base: Make 'findLsbSetFallback' constexpr to fix gcc-8 comp

BobbyRBruce:
Compilation bug found on:
https://github.com/gem5/gem5/actions/runs/5899831222/job/16002984553

In gcc Version 8 and below the following error is received:

```
src/base/bitfield.hh: In function ‘constexpr int gem5::findLsbSet(uint64_t)’:
src/base/bitfield.hh:365:34: error: call to non-‘constexpr’ function ‘int gem5::{anonymous}::findLsbSetFallback(uint64_t)’
         return findLsbSetFallback(val);
                ~~~~~~~~~~~~~~~~~~^~~~~
scons: *** [build/ALL/kern/linux/events.o] Error 1
```

`findLsbSet` cannot be `constexr` as it calls non-constexpr function
`findLsbSetFallback`. `findLsbSetFallback`. The problematic function is the `count` on the std::bitset.

This patch changes this to a constexpr.

### DIFF
--- a/src/base/bitfield.hh
+++ b/src/base/bitfield.hh
@@ -325,33 +325,43 @@ hasBuiltinCtz() {
 }
 
 [[maybe_unused]]
-int
+constexpr int
 findLsbSetFallback(uint64_t val) {
-    // Create a mask with val's trailing zeros flipped to 1, lsb set flipped to
-    // 0 and the rest unchanged. This effectively is equivalent to doing -1.
-    // e.g.: 0101000 - 1 = 0100111
-    //          ^^^^          ^^^^
-    auto mask = val - 1;
-    // This will create a mask of ones from lsb set to last bit
-    // e.g.: 0101000 ^ 0100111 = 00001111
-    //          ^^^^                 ^^^^
-    auto masked = val ^ mask;
-    // Shift that mask to that there is 1s only where there was 0s after the
-    // lsb set before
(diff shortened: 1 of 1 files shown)


powerjg:
This was introduced in #76. It would be good to get a review from @meatboy106

MeatBoy106:
Indeed, I overlooked bitset::count not being constexpr... until c++23. This version of the fallback will not buy much over the initial findLsbSet implementation in terms of performance, though. It could actually be worse (O(N) vs O(log(N)). It will depend largely on the optimizations performed by the compiler. If the constexpr-ness is more important than performance (could break API to drop it), then I would use the original implementation as a fallback. If performance is more important, I would drop constexpr.

BobbyRBruce:
> Indeed, I overlooked bitset::count not being constexpr... until c++23. This version of the fallback will not buy much over the initial findLsbSet implementation in terms of performance, though. It could actually be worse (O(N) vs O(log(N)). It will depend largely on the optimizations performed by the compiler. If the constexpr-ness is more important than performance (could break API to drop it), then I would use the original implementation as a fallback. If performance is more important, I would drop constexpr.

I opted to change it back to the original implementation as a fallback. I think this should work.

mkjost0 (approved):

powerjg (approved):

PR #204, opened 2023-08-19
resources, stdlib: Add support for local files in obtain_resource

kunpai:
This patch allows a local JSON file to specify a local path in the JSON object of a Resource, through the "url" field.

Local paths can be entered with the prefix "file:" in the "url" field.

If the local path exists, then the Resource from there is copied into the resource directory defined in the
function earlier.

This behavior is the same as using specific Resource classes (ex. BinaryResource) and passing a local_path into the function.

But, the above class does not allow simultaneous creation of local Resources and Workloads of those local Resources.

With this patch, someone can use a local JSON, specify the location of local Resources and create a Workload from those Resources and test both together.

Change-Id: I4f845e64abe947948f1db7577c52f3b0dc068cd6

### DIFF
--- a/src/python/gem5/resources/downloader.py
+++ b/src/python/gem5/resources/downloader.py
@@ -34,6 +34,7 @@ import random
 from pathlib import Path
 import tarfile
 from urllib.error import HTTPError
+from urllib.parse import urlparse
 from typing import List, Optional, Dict
 
 from _m5 import core
@@ -310,19 +311,34 @@ def get_resource(
         if run_unzip:
             download_dest += zip_extension
 
-        # TODO: Might be nice to have some kind of download status bar here.
-        # TODO: There might be a case where this should be silenced.
-        print(
-            "Resource '{}' was not found locally. Downloading to '{}'...".format(
-                resource_name, download_dest
+        file_uri_path = _file_uri_to_path(resource_json["url"])
+        if file_uri_path:
+            if not file_uri_path.exists():
+                raise Exception(
+                    f"Could not find file at path '{file_uri_path}'"
+                )
+            print(
+                "Resource '{}' is being copied from '{}' to '{}'...".format(
+                    resource_name,
+                    urlparse(resource_json["url"]).path,
+                    download_dest,
+                )
             )
-        )
+            shutil.copy(file_uri_path, download_dest)
+        else:
+            # TODO: Might be nice to have some kind of download status bar here.
+            # TODO: There might be a case where this should be silenced.
+            print(
+                "Resource '{}' was not found locally. Downloading to '{}'...".format(
+                    resource_name, download_dest
+                )
+            )
 
-        # Get the URL.
-        url = resource_json["url"]
+            # Get the URL.
+            url = resource_json["url"]
 
-        _download(url=url, download_to=download_dest)
-        print(f"Finished downloading resource '{resource_name}'.")
+            _download(url=url, download_to=download_dest)
+            print(f"Finished downloading resource '{resource_name}'.")
 
         if run_unzip:
             print(
@@ -368,3 +384,27 @@ def get_resource(
 
                 safe_extract(f, unpack_to)
             os.remove(download_dest)
+
+
+def _file_uri_to_path(uri: str) -> Optional[Path]:
+    """
+    If the URI uses the File scheme (e.g, `file://host/path`) then
+    a Path object for the local path is returned, otherwise None.
+
+    **Note:** Only files from localhost are permitted. An exception
+    is thrown otherwise.
+
+    :param uri: The file URI to convert.
+
+    :returns: The path to the file.
+    """
+
+    if urlparse(uri).scheme == "file":
+        if urlparse(uri).netloc == "" or urlparse(uri).netloc == "localhost":
+            local_path = urlparse(uri).path
+            return Path(local_path)
+        raise Exception(
+            f"File URI '{uri}' specifies host '{urlparse(uri).netloc}'. "
+            "Only localhost is permitted."
+        )
+    return None
--- a/tests/pyunit/stdlib/resources/pyunit_local_file_path_check.py
+++ b/tests/pyunit/stdlib/resources/pyunit_local_file_path_check.py
@@ -1,0 +1,72 @@
+# Copyright (c) 2023 The Regents of the University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from gem5.resources.downloader import _file_uri_to_path
+from pathlib import Path
+import unittest
+
+
+class LocalPathTestSuite(unittest.TestCase):
+    def test_local_path_exists_single_slash(self):
+        # Test that a local path is returned as-is
+        path = "file:/test/test/file"
+        expected_path = Path("/test/test/file")
+        self.assertEqual(_file_uri_to_path(path), expected_path)
+
+    def test_non_localhost_exception(self):
+        # Test that a local path with different netloc throws an exception
+        path = "file://test/test/file"
+        # should raise Exception because netloc is not '' or 'localhost'
+        with self.assertRaises(Exception) as exception:
+            _file_uri_to_path(path)
+        self.assertEqual(
+            str(exception.exception),
+            f"File URI '{path}' specifies host 'test'. "
+            "Only localhost is permitted.",
+        )
+
+    def test_localhost_accepted(self):
+        path = "file://localhost/test/test/file"
+        # should work as expected because netloc is 'localhost'
+        expected_path = Path("/test/test/file")
+        self.assertEqual(_file_uri_to_path(path), expected_path)
+
+    def test_local_path_exists_triple_slash(self):
+        # Test that a local path is returned as-is
+        path = "file:///test/test/file"
+        expected_path = Path("/test/test/file")
+        self.assertEqual(_file_uri_to_path(path), expected_path)
+
+    def test_local_path_exists_quadruple_slash(self):
+        # Test that a local path is returned as-is
+        path = "file:////test/test/file"
+        expected_path = Path("//test/test/file")
+        self.assertEqual(_file_uri_to_path(path), expected_path)
+
+    def test_uri_not_file(self):
+        # Test that a URL returns None
+        path = "http://test/test/file"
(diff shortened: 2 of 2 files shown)


src/python/gem5/resources/resource.py
```diff
@@ -624,6 +625,15 @@ def obtain_resource(
             # `exit_ok=True` ensures no exception is thrown.
             os.makedirs(resource_directory, exist_ok=True)
 
+        # if "url" starts with "file:", then we assume the resource is
+        # already on the host system and we do not need to download it.
+        # Instead, we copy the file to the resource directory.
+        if resource_json["url"].startswith("file:"):
+            to_path = resource_json["url"].split("file:")[1]
```
BobbyRBruce: This gets immediately overwritten by the code on line 638.

src/python/gem5/resources/resource.py
```diff
@@ -624,6 +625,15 @@ def obtain_resource(
             # `exit_ok=True` ensures no exception is thrown.
             os.makedirs(resource_directory, exist_ok=True)
 
+        # if "url" starts with "file:", then we assume the resource is
+        # already on the host system and we do not need to download it.
+        # Instead, we copy the file to the resource directory.
+        if resource_json["url"].startswith("file:"):
```
BobbyRBruce: I really don't like how non-standard this is. Can we try to support the standard here https://en.wikipedia.org/wiki/File_URI_scheme ?

E.g., the following should all be valid and parsed correctly:

```
file:///home/bbruce/binary
file:/home/bbruce/binary
file://localhost/home/bbruce/binary
file:////localhost/home/bbruce/binary
```

BobbyRBruce (changes requested):
A better way to do this would be to change the downloader code to handle local files. Here is a Stackoverflow page explaining how to extend the Requests library to handle files: https://stackoverflow.com/questions/10123929/fetch-a-file-from-a-local-url-with-python-requests.

src/python/gem5/resources/resource.py
powerjg: Agreed. Specifically, you can use urllib.parse (https://docs.python.org/3/library/urllib.parse.html) and get the `scheme`

powerjg (commented):
lgtm. I'll let @BobbyRBruce approve

src/python/gem5/resources/downloader.py
```diff
@@ -253,6 +254,16 @@ def get_resource(
...
+        if (
+            "url" in resource_json
+            and resource_json["url"]
+            and (urlparse(resource_json["url"]).scheme == "file")
+        ):
+            local_path = urlparse(resource_json["url"]).path
+            if os.path.exists(local_path):
+                shutil.copy(local_path, to_path)
+                return
+
```
BobbyRBruce: I'm not sure this covers all of the specification here. I think i want to see tests I'm afraid. Could you do the following?:

1. Move this into a private function `def _file_uri_to_path(uri: str) -> Optional[Path]`. This function will return a path if the uri is a file, otherwise none.
2. Update `get_resource` to use this function.
3. Using the pyunit test framework, add unittests for this new function. I would look over https://en.wikipedia.org/wiki/File_URI_scheme to create test input data and expected outputs.

As a small example of input/output pairs to test:

|Input|Output|
|---|---|
|https://gem5.org/bla|None|
|file:///home/bbruce/binary|/home/bbruce/binary|
|file:/home/bbruce/binary|/home/bbruce/binary|
|file://localhost/home/bbruce/binary|/home/bbruce/binary|
|file:////localhost/home/bbruce/binary|/home/bbruce/binary|
|file://other-host/home/bbruce/binary| Exception("Can only obtain file from localhost") | 

Add more so we're covering every case here. Don't just copy and paste, have a think about what kind of behavior we want to verify here.

BobbyRBruce (changes requested):

src/python/gem5/resources/downloader.py
kunpai: Alright! Seems fair to me, I will also make sure to add in-line documentation to the new function as well.

src/python/gem5/resources/downloader.py
```diff
@@ -368,3 +380,19 @@ def safe_extract(
...
             os.remove(download_dest)
+
+
+def _file_uri_to_path(uri: str) -> Optional[Path]:
+    """
+    Converts a file URI to a local path.
+
+    :param uri: The file URI to convert.
+
+    :returns: The path to the file.
```
BobbyRBruce:
```suggestion
    If the URI uses the File scheme (e.g, `file://host/path`) then
    a Path object for the local path is returned, otherwise None.
    
    **Note:** Only files from localhost are permitted. An exception
    is thrown otherwise.

    :param uri: The file URI to convert.

    :returns: The path to the file.
```

src/python/gem5/resources/downloader.py
```diff
@@ -368,3 +380,19 @@ def safe_extract(
...
+    """
+    if urlparse(uri).netloc == "" or urlparse(uri).netloc == "localhost":
+        local_path = urlparse(uri).path
+        if os.path.exists(local_path):
```
BobbyRBruce: So I think there was a small misunderstanding in what I requested (please let me know if there's a reason you did otherwise).

My intent was that "None" would be returned if it wasn't conforming to the File schema (e.g., it may be a https URL, for example). The exception case was only for the case that the URI was a valid file scheme URI but for a host other than localhost support.

Basically:
- None: This isn't a File scheme URI
- Exception: Is a File scheme URI but for a feature we haven't implemented.
- Path value: just the Path specified by the file URI. It doesn't need to exist. `Path.exits()` will do that for us I believe and can be done outside of the function.

BobbyRBruce (changes requested):

src/python/gem5/resources/downloader.py
kunpai: What would be an example of a File scheme URI feature we haven't implemented?

After reading your input here, I am thinking the function I want to add should do the following:

1. sanity check if the scheme is "file".
2. if true, then checks the netloc
3. if the netloc is not localhost or blank, throws an Exception
4. if the netloc is localhost or blank, returns the Path object of the URI passed into the function
5. if the scheme being "file" is false, return None

Meanwhile, in the get_resource function:
1. if the return value of the function is a Path, check if the Path exists and if yes, do the shutil.copy function as we have been doing so far.
2. if the return value is None, throw an exception stating that the scheme is not "file" (not sure what the behavior should be here).

What do you think @BobbyRBruce?

src/python/gem5/resources/downloader.py
kunpai: I'll incorporate these

src/python/gem5/resources/downloader.py
BobbyRBruce: > What would be an example of a File scheme URI feature we haven't implemented?

So the File scheme URI allows for getting files on different hosts. For example `file://liberty/home/bbruce/some-file.txt` is valid. It's saying "The file '/home/bburce/some-file.txt' in the machine with hostname 'liberty'". This is something we're not going to implement. It's valid use of the standard but we're not going to implement it. That's the only thing, I think, we're not really supporting here. We're saying it must be the File scheme URL for localhost.

>After reading your input here, I am thinking the function I want to add should do the following:

> 1. sanity check if the scheme is "file".
> 2. if true, then checks the netloc
> 3. if the netloc is not localhost or blank, throws an Exception
> 4. if the netloc is localhost or blank, returns the Path object of the URI passed into the function
> 5. if the scheme being "file" is false, return None

That is what I wanted. I know we're really implementing two functions in one here: `is_file_scheme(uri: str)->bool` and `get_path_from_file_url(uri:str)->Path` by making `None` communicate the case where `is_file_scheme(uri) == False`, but this isn't a public API will almost certainly  just be used for this case, so let's keep it simple. Just make sure the docstring explains what each output actually means (you can use my suggested change).

>Meanwhile, in the get_resource function:

> 1. if the return value of the function is a Path, check if the Path exists and if yes, do the shutil.copy function as we have been doing so far.
> 2. if the return value is None, throw an exception stating that the scheme is not "file" (not sure what the behavior should be here).

(1.) makes sense to me. 

For (2.) I'll leave some comments in your code, it should be easier to explain there.

src/python/gem5/resources/downloader.py
```diff
@@ -253,6 +254,17 @@ def get_resource(
...
+            "url" in resource_json
+            and resource_json["url"]
+            and (urlparse(resource_json["url"]).scheme == "file")
+        ):
+            local_path = _file_uri_to_path(resource_json["url"])
+            if os.path.exists(local_path):
+                shutil.copy(local_path, to_path)
+                return
+            raise Exception(f"Could not find file")
+
```
BobbyRBruce: I suggest you move this logic down to just before https://github.com/gem5/gem5/blob/f29bfc0640c88a79eb7f94454ce31b3237ec0066/src/python/gem5/resources/downloader.py#L311

The only thing you'll really want to do is:

```py
file_uri_path = _file_uri_to_path(resource_json["url"])
if file_uri_path:
    if not file_uri_path():
        raise Exception("bla bla bla, file not found")
    # Best put a helpful message here saying your copying the file
    # from one locaiton to the other, like what we do with the
    # downloading code.
    shutil.copy(file_uri_path, download_path)
else: 
   # The code that's already there. This will end in this else-block
   # with `print(f"Finished downloading resource '{resource_name}'.")`

# Continue with the rest of the function to handle decompression, untarring, etc.
if run_unzip:
...
```

src/python/gem5/resources/downloader.py
```diff
@@ -253,6 +254,17 @@ def get_resource(
             gem5_version=gem5_version,
         )
 
+        if (
+            "url" in resource_json
+            and resource_json["url"]
+            and (urlparse(resource_json["url"]).scheme == "file")
+        ):
```
BobbyRBruce: My proposal about moving this means you don't need this if check.

src/python/gem5/resources/downloader.py
```diff
@@ -368,3 +384,24 @@ def safe_extract(
...
+    :param uri: The file URI to convert.
+
+    :returns: The path to the file.
+    """
+
+    if urlparse(uri).scheme == "file":
+        if urlparse(uri).netloc == "" or urlparse(uri).netloc == "localhost":
+            local_path = urlparse(uri).path
+            return Path(local_path)
+        raise Exception("Can obtain files only from localhost")
```
BobbyRBruce:
```suggestion
        raise Exception(
            f"File URI '{uri}' specifies host '{urlparse(uri).netloc}'. "
            "Only localhost is permitted."
        )
        
```

tests/pyunit/stdlib/resources/pyunit_local_file_path_check.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertEqual(_file_uri_to_path(path), expected_path)
+
+    def test_local_path_exists_double_slash(self):
+        # Test that a local path with different netloc throws an exception
+        path = "file://test/test/file"
+        # should raise Exception because netloc is not '' or 'localhost'
+        with self.assertRaises(Exception) as exception:
+            _file_uri_to_path(path)
+        self.assertEqual(
+            str(exception.exception), "Can obtain files only from localhost"
```
BobbyRBruce:
```suggestion
            str(exception.exception),
            f"File URI '{path}' specifies host 'test'. "
            "Only localhost is permitted."
```

tests/pyunit/stdlib/resources/pyunit_local_file_path_check.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class LocalPathTestSuite(unittest.TestCase):
+    def test_local_path_exists_single_slash(self):
+        # Test that a local path is returned as-is
+        path = "file:/test/test/file"
+        expected_path = Path("/test/test/file")
+        self.assertEqual(_file_uri_to_path(path), expected_path)
+
+    def test_local_path_exists_double_slash(self):
```
BobbyRBruce:
```suggestion
    def test_non_localhost_exception(self):
```

tests/pyunit/stdlib/resources/pyunit_local_file_path_check.py
```diff
@@ -0,0 +1,69 @@
...
+        # should raise Exception because netloc is not '' or 'localhost'
+        with self.assertRaises(Exception) as exception:
+            _file_uri_to_path(path)
+        self.assertEqual(
+            str(exception.exception), "Can obtain files only from localhost"
+        )
+        path = "file://localhost/test/test/file"
+        # should work as expected because netloc is 'localhost'
+        expected_path = Path("/test/test/file")
+        self.assertEqual(_file_uri_to_path(path), expected_path)
```
BobbyRBruce, line 55:
```suggestion
    def test_localhost_accepted(self):
        path = "file://localhost/test/test/file"
        # should work as expected because netloc is 'localhost'
        expected_path = Path("/test/test/file")
        self.assertEqual(_file_uri_to_path(path), expected_path)
```

tests/pyunit/stdlib/resources/pyunit_local_file_path_check.py
```diff
@@ -0,0 +1,69 @@
...
+        # Test that a local path is returned as-is
+        path = "file:////test/test/file"
+        expected_path = Path("//test/test/file")
+        self.assertEqual(_file_uri_to_path(path), expected_path)
+
+    def test_url_error(self):
+        # Test that a URL returns None
+        path = "http://test/test/file"
+        expected_path = None
+        self.assertEqual(_file_uri_to_path(path), expected_path)
```
BobbyRBruce:
```suggestion
    def test_uri_not_file(self):
        # Test that a URL returns None
        path = "http://test/test/file"
        self.assertIsNone(_file_uri_to_path(path))
```

BobbyRBruce (changes requested):
These are all quite minor. Generally they are no major faults.

BobbyRBruce (approved):

BobbyRBruce:
Great Work @kunpai! I'm happy with this new addition! Thanks for responding to everything promptly.

kunpai:
Thank you for your inputs @BobbyRBruce and for making this change more holistic!

PR #205, opened 2023-08-19
mem-cache: Allow clflush's uncacheable requests on classic cache

hnpl:
When a linux kernel changes a page property, it flushes the related cache lines. The kernel might change the page property before flushing the cache lines. This results in the clflush might occur in an uncacheable region.

Currently, an uncacheable request must be a read or a write. However, clflush request is neither of them.

This change aims to allow clflush requests to work on uncacheable regions. Since there is no straightforward way to check if a packet is from a clflush instruction, this change permits all Clean Invalidate Requests, which is the type of request produced by clflush, to work on uncacheable regions.

Change-Id: Ib3ec01d9281d3dfe565a0ced773ed912edb32b8f

### DIFF
--- a/src/mem/cache/cache.cc
+++ b/src/mem/cache/cache.cc
@@ -352,7 +352,7 @@ Cache::handleTimingReqMiss(PacketPtr pkt, CacheBlk *blk, Tick forward_time,
             // the clflush might occur in an uncacheable region.
             // clflush results in a CleanInvalidReq, which is neither read
             // nor write.
-            assert(pkt->req->isCleanInvalidateRequest());
+            assert(pkt->isCleanInvalidateRequest());
             allocateWriteBuffer(pkt, forward_time);
         }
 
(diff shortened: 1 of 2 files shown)


hnpl:
The if-else statement is rather silly, but I think it helps clarify the logic, and it's pretty straightforward to optimize them.

powerjg (dismissed):

hnpl:
Relevant kernel function, https://elixir.bootlin.com/linux/v5.19.17/source/arch/x86/mm/pat/set_memory.c#L1687.

powerjg:
Can you comment on how you tested this change?

hnpl:
I ran this with a setup where set_memory_uc() is called at some point.

Without this change, gem5 failed to execute the clflushopt instruction, which is part of the function above.

The setup ran without error when the above change is applied.

powerjg (approved):

PR #206, opened 2023-08-21
tests: Update asmtest script and add more test binaries

rogerchang23424:
Upload the config script to make it only for riscv asmtest and replace Resource with obtain_resourse

Change-Id: I0bab96ea352b7ce1c6838203bfa13eee795f41f9


powerjg:
Looks great to me... Thanks for the contribution! I'll let Bobby sign off.

tests/gem5/asmtest/configs/riscv_asmtest.py
```diff
@@ -170,8 +88,8 @@
 )
 
 # Set the workload
-binary = Resource(args.resource, resource_directory=args.resource_directory)
-motherboard.set_se_binary_workload(binary, arguments=args.arguments)
+binary = obtain_resource(args.resource)
```
BobbyRBruce: So, I'd quite like there to continue to be an `resource_directory=args.resource_directory` option here. When testing I store resources downloaded in "tests/gem5/resources".

tests/gem5/asmtest/tests.py
```diff
@@ -26,11 +26,6 @@
 
 from testlib import *
 
-if config.bin_path:
-    resource_path = config.bin_path
-else:
-    resource_path = joinpath(absdirpath(__file__), "..", "resources")
-
```
BobbyRBruce: I'd like this code-readded, unless there's a good argument to remove it.

BobbyRBruce (changes requested):

BobbyRBruce (approved):

PR #209, opened 2023-08-22
ext: Specialize GDBSignal MACRO to gem5

### LINKED ISSUE #94: macOS build failure due to conflicting defines in gdbremote/signals.hh
This commit breaks building on macOS https://github.com/gem5/gem5/commit/7230a3e7f0f23621d9d09df3f7420c08a08cc118 as EXC_* are already defined in exception_types.h. I'm happy to submit a PR undef'ing these, but I'm not certain if that is the desired fix?

BobbyRBruce:
The goal is to fix this issue which appears to be affects some Apple users: https://github.com/gem5/gem5/issues/94.

By specializing the `EXC_*` to gem5 we avoid the name conflicts plagiing some users.

Change-Id: I031f7110b4b4ae82677b6586903cd57b22ca2137

### DIFF
--- a/ext/gdbremote/signals.hh
+++ b/ext/gdbremote/signals.hh
@@ -168,12 +168,12 @@ namespace gem5{
     INFO = 142, //information request
     unknown = 143, //unknown signal
 
-    EXC_BAD_ACCESS = 145, //could not access memory
-    EXC_BAD_INSTRUCTION = 146, //illegal instruction/operand
-    EXC_ARITHMETIC = 147, //arithmetic exception
-    EXC_EMULATION = 148, //emulation instruction
-    EXC_SOFTWARE = 149, //software generated exception
-    EXC_BREAKPOINT = 150, //breakpoint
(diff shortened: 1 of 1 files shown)


powerjg:
If it's possible, I would prefer somehow not including `mach/exception_types.h` I'm not sure why/how that file is included. It's odd that this MacOS-unique file is included in the gem5 build.

That said, I just spent some time digging into the use of `GDBSignal` and this change should not break anything, so I would be OK with it going in as-is.

BTW, here's the commit where this was introduced: https://github.com/gem5/gem5/commit/7230a3e7f0f23621d9d09df3f7420c08a08cc118

BobbyRBruce:
@powerjg : This appears to fix the issue they are facing. Your the assignee here so i'll leave it to you, but my personal vote is to get this in. It may be inelegant but I don't see how it could introduce any problems.

powerjg (approved):

shingarov (approved):
This was affecting me as well; I have verified that the proposed diff resolves the issue for me.  Let's merge this PR.

PR #210, opened 2023-08-22
sim: provide a signal constructor with an init_state

johnnyko-google:
The current SignalSinkPort and SignalSourcePort have no ways to assign the init value of the state. Add a new constructor for them with the param init_state

Bug: 293410800
Test: boot to linux
Change-Id: Idde0a12aa0ddd0c9c599ef47059674fb12aa5d68
Reviewed-on: https://soc-sim-external-review.googlesource.com/c/gem5/gem5/+/13159
Gem5-Virtual-Platform-Presubmit-Ready: Johnny Ko <johnnyko@google.com>
Reviewed-by: Yu-hsin Wang <yuhsingw@google.com>
Perf-Presubmit-Ready: Johnny Ko <johnnyko@google.com>
Gem5-Virtual-Platform-Verified: kokoro <noreply+kokoro@google.com>
Perf-Verified: kokoro <noreply+kokoro@google.com>

### DIFF
--- a/src/sim/signal.hh
+++ b/src/sim/signal.hh
@@ -79,6 +79,7 @@ class SignalSinkPort : public Port
         _source = dynamic_cast<SignalSourcePort<State> *>(&peer);
         fatal_if(!_source, "Attempt to bind signal pin %s to "
                 "incompatible pin %s", name(), peer.name());
+        _state = _source->state();
         Port::bind(peer);
     }
     void
@@ -94,12 +95,21 @@ class SignalSourcePort : public Port
 {
   private:
     SignalSinkPort<State> *sink = nullptr;
-    State _state = {};
+    State _state;
 
   public:
-    SignalSourcePort(const std::string &_name, PortID _id=InvalidPortID) :
-        Port(_name, _id)
-    {}
+    SignalSourcePort(const std::string &_name, PortID _id = InvalidPortID)
+        : Port(_name, _id)
+    {
+        _state = {};
+    }
+
+    SignalSourcePort(const std::string &_name, PortID _id,
+                     const State &init_state)
+        : SignalSourcePort(_name, _id)
+    {
+        _state = init_state;
+    }
 
     void
     set(const State &new_state)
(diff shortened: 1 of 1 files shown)


giactra:
Hello, I am not an expert in this part of the code, but it seems to me this commit is not only allowing the state to be passed at construction time, it is also making it percolate through the ports when bound together. It looks reasonable to me, I have two questions though:

1) Would it make sense to split the binding functionality in a separate commit under the same PR? Or at least to document this change in the PR description?
2) If we pass the state in the bind method, shall we reset it in the unbind?

src/sim/signal.hh
```diff
@@ -94,12 +95,21 @@ class SignalSourcePort : public Port
...
-    SignalSourcePort(const std::string &_name, PortID _id=InvalidPortID) :
-        Port(_name, _id)
-    {}
+    SignalSourcePort(const std::string &_name, PortID _id = InvalidPortID)
+        : Port(_name, _id)
+    {
+        _state = {};
+    }
+
+    SignalSourcePort(const std::string &_name, PortID _id,
```
powerjg, line 109: Related to Giacomo's comment, it would be good to add a comment on this function explaining why a non-standard `Port` constructor is needed here.

Harshil2107:
@giactra I'm going to assign this PR to you. As a reminder, being assigned means that you are responsible to make sure that there is progress made on the PR and you can merge it when it's ready.

Thanks for the help!

src/sim/signal.hh
johnnyko-google: Hi,  I have added the description in the next commit.

johnnyko-google:
> 
> Hello, I am not an expert in this part of the code, but it seems to me this commit is not only allowing the state to be passed at construction time, it is also making it percolate through the ports when bound together. It looks reasonable to me, I have two questions though:
> 
> 1. Would it make sense to split the binding functionality in a separate commit under the same PR? Or at least to document this change in the PR description?

Hi, I have added more description in the 1st commit message.

> 2. If we pass the state in the bind method, shall we reset it in the unbind?

No needed. The unbind should not affect the value of the state. The state of source should be the same as the state of sink after bind. Before my change, it also works because both sides use default value as the init value of the state during their initialization. My change allows users to use their preferred value instead of default value.

powerjg (commented):
lgtm... I'll leave it Giacomo to approve.

giactra (approved):

PR #212, opened 2023-08-22
resources,stdlib: Add workload to resource specialization and deprecate workload.py

### LINKED ISSUE #193: Add workload to resource specialization of resources.py and deprecate workload.py
At present, there exist two pathways for resource acquisition: the first involves the utilization of the `obtain_resource()` function, while the second entails employing the `Workload` class found within `workload.py` for workloads only. Remarkably, both of these approaches leverage a shared function to access the JSON resources. Notably, a workload itself can be conceptualized as a resource. As part of an enhancement, we can create a dedicated "workload resource," thereby rendering the current `workload.py` obsolete. This transition would involve deprecating `workload.py` into a wrapper encapsulating the functionality of `obtain_resources()`. 

Successful implementation depends on issue #192  requiring updates to all workloads before pushing this change.

Harshil2107:


src/python/gem5/resources/resource.py
```diff
@@ -554,6 +554,80 @@ def _get_weights_and_simpoints_from_file(
...
+    """A workload resource. This resource is used to specify a workload to run
+    on a board. It contains the function to call and the parameters to pass to
+    that function.
+    """
+
+    def __init__(
+        self,
+        function: str = None,
+        resources: Optional[Dict[str, str]] = None,
+        additional_params: Optional[Dict[str, str]] = None,
```
BobbyRBruce: I can see why you're including these, as this reflects the schema, but this is redundant with the field `parameters`. I would remove these two fields and have a parser in the constructor to create `parameters` from the JSON entry.

src/python/gem5/resources/resource.py
```diff
@@ -640,7 +714,6 @@ def obtain_resource(
     # Obtain the type from the JSON. From this we will determine what subclass
     # of `AbstractResource` we are to create and return.
     resources_category = resource_json["category"]
-
```
BobbyRBruce: Re-add this line. No point removing it. It'll ruin `git blame`

src/python/gem5/resources/workload.py
```diff
@@ -24,219 +24,55 @@
...
-    ```py
-    workload = CustomWorkload(
-        function = "set_se_binary_workload",
-        parameters = {
-            "binary" : obtain_resource("x86-print-this"),
-            "arguments" : ["hello", 6]
-        },
+    warn(
+        "`Workload` has been deprecated. Please use the `obtain_resource` "
+        "function instead."
```
BobbyRBruce: Let's do people a favor here and construct the warn to say explicitly what call they should be making instead. How about:

```suggestion
        "`Workload` has been deprecated. Please use the `obtain_resource` "
        "function instead:\n\n"
        "```\n"
        "from gem5.resources.resource import obtain_resource\n"
        "workload = obtain_resource(\n"
        f"    resource_id=\"{workload_name}\",\n"
        f"    resource_directory=\"{resource_directory}\",\n"
        f"    gem5_version=\"{gem5_version}\",\n"
        f"    clients={clients},\n"
        f"    resource_version={resource_version},\n"
        ")\n"
        "```"
```

src/python/gem5/resources/workload.py
```diff
@@ -24,219 +24,55 @@
...
-    ```py
-    workload = CustomWorkload(
-        function = "set_se_binary_workload",
-        parameters = {
-            "binary" : obtain_resource("x86-print-this"),
-            "arguments" : ["hello", 6]
-        },
+    **Warning**: This class is deprecated and changed to a funtion and will be removed in future
+    releases of gem5. Please use the `WorkloadResource` class instead. This
+    class is merely a wrapper for it.
```
BobbyRBruce:
```suggestion
    **Warning**: This `CustomWorkload` class is deprecated. It will be removed in a
    future release of gem5. Please use the `gem5.resources.resource.WorkloadResource`
    class instead.
    
    The class has been stealthily converted to a function which wraps the
    `WorkloadResource` class.
```

src/python/gem5/resources/workload.py
```diff
@@ -24,219 +24,55 @@
...
+    """
+    warn(
+        "The `CustomWorkload` class is deprecated. Please use "
+        "`WorkloadResource` instead."
```
BobbyRBruce:
```suggestion
        "the `gem5.resources.resource.WorkloadResource` class instead."
```

src/python/gem5/resources/workload.py
```diff
@@ -24,219 +24,55 @@
...
+def Workload(
+    workload_name: str,
+    resource_directory: Optional[str] = None,
+    resource_version: Optional[str] = None,
+    clients: Optional[List] = None,
+    gem5_version: Optional[str] = core.gem5Version,
+):
     """
+    This function was created to maintain backwards compability for v23.0.0
+    and prior releases of gem5 where `Workload` was a class.
```
BobbyRBruce:
```suggestion
    **Warning**: The `Workload` class is deprecated. It will be removed in a
    future release of gem5. Please use the `gem5.resources.resource.WorkloadResource`
    class instead.
```

src/python/gem5/resources/workload.py
```diff
@@ -24,219 +24,55 @@
...
-        :param value: The value to set to the parameter.
-        """
-        self._params[parameter] = value
-
-
-class CustomWorkload(AbstractWorkload):
+    In the interests of gem5-resource specialization, the `Workload` class
+    has been dropped. Instead users are advized to use the `obtain_resource`
+    function which will return the correct `AbstractResource` implementation.
+    This function (disguised as a class) wraps this function.
```
BobbyRBruce:
```suggestion
    The class has been stealthily converted to a function which wraps the
    `WorkloadResource` class.
```

BobbyRBruce (changes requested):
Not urgent but you might as well do it now: you should remove all usage of the `Workload` class in the repo and replace it with `obtain_resources`.

src/python/gem5/resources/resource.py
Harshil2107: I have a parser in the constructor to create `parameters`, and I have the `parameters` fields to keep backward compatibility with `CustomWorkload` as it directly passed `parameter` instead of `resources` and `additional_params`.

BobbyRBruce: I don't follow. I'm suggesting we get rid of `additional_parameters` and `resources` as parameters here. This will keep backwards compatibility with `CustomWorkload`. What we'd change is when constructing this from gem5 Resources data sources. We'd parse the `resources` and `additional_params fields into `parameters`. It shouldn't be too hard to do.

Harshil2107: To confirm, do you mean adding an if in `obtain_resources` for the category `workload` and doing the parsing there instead of in the constructor of the `WorkloadResource` class?

BobbyRBruce:
You should be able to reproduce this locally with `./build/ALL/gem5.opt tests/run_pyunit.py`. I suspect some tests need updated here.

Harshil2107:
I will push another change that updates all uses of `Workload` class to `obtain_resoruce` later today.

BobbyRBruce (approved):

PR #215, opened 2023-08-23
Fix (deprecated) se.py to work with multi-ISA gem5

janvrany:
This PR fixes `se.py` to work with multi-ISA gem5. 

I know `se.py` has been deprecated and fixing it might be undesirable. However, we still use it for some of our work so we needed to fix it anyway. Since the `se.py` is not yet removed and I have a fix at hand, I decided to open the PR anyway.

### DIFF
--- a/configs/deprecated/example/se.py
+++ b/configs/deprecated/example/se.py
@@ -50,21 +50,9 @@ from m5.objects import *
 from m5.params import NULL
 from m5.util import addToPath, fatal, warn
 from gem5.isas import ISA
-from gem5.runtime import get_runtime_isa
+from gem5.runtime import get_runtime_isa, get_supported_isas
 
 addToPath("../../")
-
-from ruby import Ruby
-
-from common import Options
-from common import Simulation
-from common import CacheConfig
-from common import CpuConfig
-from common import ObjectList
-from common import MemConfig
-from common.FileSystemConfig import config_filesystem
-from common.Caches import *
-from common.cpu2000 import *
 
 
 def get_processes(args):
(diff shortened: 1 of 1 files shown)


giactra:
Thanks janvrany for this PR. I don't have strong feelings against patching deprecated code, but I don't think this is the way we should handle the issue.

From your backtrace (thanks for posting it) it seems the problem is really the import of common.cache modules (and more specifically the PageTableWalkerCache) which are NOT multi-ISA compatible. Not only rewriting the class differently would fix the problem (though other problems might arise), but this problem affects every potential client of the common module (not only se.py).

- Long term solution is to remove the common.Cache module
- Short-Medium term solution is to make it multi-ISA compatible -> means we should stop using get_runtime_isa

configs/deprecated/example/se.py
```diff
@@ -119,11 +107,92 @@ def get_processes(args):
...
+EI_E_MACHINE = 18  # e_machine first byte index
+
+
+def set_isa(args):
+    """
+    Set the architecture based on executable (if this gem5 has
+    support for multiple architectures).
+    """
+    if len(get_supported_isas()) > 1:
+        exe_path = args.cmd.split(";")[0]
```
giactra, line 144: In your code you are setting the TARGET_ISA based on the first binary passed as an argument. What happens if you have multiple binaries with different ISAs ? Basically you will be using the same data structures (like PageTableWalkerCache) for all cpus (get_runtime_isa will always return the same value), while you should make distinction between them.

As I commented earlier, I believe we should rewrite the common module to make it multi-ISA compatible (for example extracting the ISA from the CPU type or the cpu.isa object....)

janvrany:
@giactra 
> but I don't think this is the way we should handle the issue.

Fair enough. I know this is not not a "proper" fix. However, I certainly lack knowledge to do any of what you suggested myself, therefore I propose to close the PR.

configs/deprecated/example/se.py
janvrany: In that case you'd get either illegal instruction or even load error for the firts binary that's for other ISA than the first. 

Was this scenario supported back in the days se.py was not deprecated and there was no multi-ISA gem5? 

Adding a check if all binaries are for the same ISA is easy enough. I'll do it should we decide to go forward with this approach.

giactra:
> Fair enough. I know this is not not a "proper" fix. However, I certainly lack knowledge to do any of what you suggested myself, therefore I propose to close the PR.

Thanks @janvrany, it was a very reasonable fix, unfortunately there's a lot of hidden complexity in the legacy part of gem5 which doesn't make it the best candidate (and that's why we should remove it at a certain point).

While I am closing this, I will create an issue and link it to this PR...

configs/deprecated/example/se.py
giactra: > In that case you'd get either illegal instruction or even load error for the firts binary that's for other ISA than the first.

Not if the simulation really assumes multiple CPUs with different ISAs
So I can have an Arm CPU wunning an Arm workloads and a RISCV CPU running a RISCV workload in the same simulation. If things are done properly no error will arise.

> Was this scenario supported back in the days se.py was not deprecated and there was no multi-ISA gem5?

Yes without multi-ISA, the non TARGET_ISA binary will fail to be loaded by the ElfLoader class (in C++ https://github.com/gem5/gem5/blob/stable/src/base/loader/elf_object.cc#L222) when comparing the Arch with the Workloads class (https://github.com/gem5/gem5/blob/stable/src/arch/arm/linux/se_workload.cc#L67)

PR #217, opened 2023-08-23
mem-ruby: fix CHI Evict race condition

tiagormk:
When an Evict request is received from upstream for a shared line and the line is no longer cached locally (or on any other upstream cache), we need to also send an Evict downstream. In this case we need to wait until our outgoing Evict completes before completing the Evict from upstream in order be able to resolve race conditions with incoming snoops. E.g.: while our outgoing Evict is pending we may receive a snoop requesting data, but we won't be able to complete this snoop if we have already completed all upstream Evicts and we no longer have the line.

Change-Id: I23ac4f0a9c4ddd81e2425376c8d1e1c7fb66d107


mattsinc (approved):
This seems reasonably to me

PR #220, opened 2023-08-23
Improve Ruby/CHI stats for incoming/outgoing transactions 

tiagormk:
Currently we generate these stats for all defined Events in the protocol, which may generate too many stats that are never used. Though these don't appear in the stats.txt file, they unnecessarily increases simulation startup time and memory footprint.

This patch limits those stats to events with the "in_trans" and/or "out_trans" properties. SLICC compiler then checks which combinations of event+state are possible when generating the stats.

Also the possible level of detail for inTransLatHist was reduced.
Only the number of transactions for each event+initial+final state combinations is now accounted. Latency histograms are only defined per event type (similarly to outTransLatHist). This significantly reduces the final file size for generated stats.


BobbyRBruce (approved):
Looks good to me!

PR #221, opened 2023-08-23
redirect_path patch for restoring cpt

studyztp:
Modify the FDArray::unserialize function to perform a checkPathRedirect if a Process pointer is passed in.
Currently when restoring a checkpoint, it doesn't perform checkPathRedirect for files that were opened during checkpointing. This patch adds a checkPathRedirect in the FDArray::unserialize to redirect app path for restoring checkpoints.

### DIFF
--- a/src/sim/fd_array.cc
+++ b/src/sim/fd_array.cc
@@ -422,14 +422,12 @@ FDArray::unserialize(CheckpointIn &cp, Process* process_ptr) {
 
         std::string path;
 
-        if (process_ptr)
-        {
+        if (process_ptr) {
             // Check if it is needed to redirect the app path to another host
             // path
             path = process_ptr->checkPathRedirect(this_ffd->getFileName());
         }
-        else
-        {
+        else {
             path = this_ffd->getFileName();
         }
 
--- a/src/sim/fd_array.cc
+++ b/src/sim/fd_array.cc
@@ -42,6 +42,7 @@
 #include "base/output.hh"
 #include "params/Process.hh"
 #include "sim/fd_entry.hh"
+#include "sim/process.hh"
 
 namespace gem5
 {
@@ -367,7 +368,7 @@ FDArray::serialize(CheckpointOut &cp) const {
 }
 
 void
-FDArray::unserialize(CheckpointIn &cp) {
+FDArray::unserialize(CheckpointIn &cp, SimObject* process_ptr) {
     ScopedCheckpointSection sec(cp, "fdarray");
     uint64_t size;
     paramIn(cp, "size", size);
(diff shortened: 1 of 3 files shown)


src/sim/fd_array.cc
```diff
@@ -367,7 +368,7 @@ FDArray::serialize(CheckpointOut &cp) const {
 }
 
 void
-FDArray::unserialize(CheckpointIn &cp) {
+FDArray::unserialize(CheckpointIn &cp, SimObject* process_ptr) {
```
powerjg: Shouldn't this be `Process*`?

src/sim/fd_array.hh
```diff
@@ -36,9 +36,9 @@
 #include <map>
 #include <memory>
 #include <string>
-
 #include "sim/fd_entry.hh"
 #include "sim/serialize.hh"
+#include "sim/sim_object.hh"
```
powerjg: Don't include this.

Instead, inside the `gem5` namespace add the line `class Process;` This is a "forward declaration"

src/sim/fd_array.hh
```diff
@@ -117,7 +117,11 @@ class FDArray : public Serializable
      * Serialization methods for file descriptors
      */
     void serialize(CheckpointOut &cp) const override;
-    void unserialize(CheckpointIn &cp) override;
+    void unserialize(CheckpointIn &cp, SimObject* process_ptr );
```
powerjg: Should be `Process *`

powerjg (changes requested):

src/sim/fd_array.hh
```diff
@@ -36,9 +36,9 @@
 #include <map>
 #include <memory>
 #include <string>
-
```
BobbyRBruce: There should be a new line here. Please re-add it.

BobbyRBruce (changes requested):
Assuming Jason's concerns are addressed, everything else bar a small line deletion is fine with me.

src/sim/fd_array.hh
studyztp: I added it. Now I know what is "forward declaration". Thank you!

src/sim/fd_array.cc
studyztp: Fixed it.

src/sim/fd_array.hh
studyztp: Fixed it.

src/sim/fd_array.hh
studyztp: I re-added it. Thanks for reviewing this!

BobbyRBruce (dismissed):
Fine by me but I'll let @powerjg  sign off.

src/sim/fd_array.cc
```diff
@@ -418,11 +419,26 @@ FDArray::unserialize(CheckpointIn &cp) {
         setFDEntry(tgt_fd, fdep);
 
         mode_t mode = this_ffd->getFileMode();
-        std::string const& path = this_ffd->getFileName();
+
+        std::string path;
+
+        if (process_ptr)
+        {
```
powerjg: These curly braces should be on the previous line

src/sim/fd_array.cc
```diff
@@ -418,11 +419,26 @@ FDArray::unserialize(CheckpointIn &cp) {
...
+            // Check if it is needed to redirect the app path to another host
+            // path
+            path = process_ptr->checkPathRedirect(this_ffd->getFileName());
+        }
+        else
+        {
```
powerjg: Also this one

powerjg (changes requested):
One more tiny style thing

studyztp:
> One more tiny style thing

Fixed them. Thanks!

BobbyRBruce (approved):

powerjg (approved):

PR #222, opened 2023-08-24
misc: Move compiler tests to run on 'build' runners

BobbyRBruce:
This is an experiment. The runners were sometimes running out of memory building gem5. The builders have more memory so should be able to handling this. The runners have 4-cores so compilation should be faster (note the inclusion of the `-j$(nproc)`.


BobbyRBruce:
@mkjost0 I assume this is something I'll need to copy to the `stable` branch to work as intended?

mkjost0:
Yeah, since it's scheduled it runs based off of whatever's on stable

mkjost0 (approved):

PR #224, opened 2023-08-24
dev-amdgpu: Tell OS about PCIe atomic support

abmerop:
configs,dev-amdgpu: Add PCI express capability info

The ROCm stack requires PCI express atomics. Currently the first PCI
CapabilityPtr does not point to anything, which signals to the OS
(Linux) that this is an early generation PCI device. As PCI express
atomics were introduced later, the CapabilityPtr needs to point to at
least a PCI express capability structure. This capability is defined as
0x10 in Linux. We additionally set the PCI atomic based bits and
implement device specific PCI configuration space reads and writes to
the amdgpu device.

The second commit, output of simulation when loading the amdgpu
driver no longer outputs "PCIE atomics not supported". Further, an
application which uses PCIe atomics (PyTorch with a reduce_sum kernel)
now makes further progress.

First commit is a minor typo fix changing PCI capability struct to
union.

### DIFF
--- a/configs/example/gpufs/system/amdgpu.py
+++ b/configs/example/gpufs/system/amdgpu.py
@@ -185,3 +185,26 @@ def connectGPU(system, args):
         system.pc.south_bridge.gpu.DeviceID = 0x6863
     else:
         panic("Unknown GPU device: {}".format(args.gpu_device))
+
+    # Use the gem5 default of 0x280 OR'd  with 0x10 which tells Linux there is
+    # a PCI capabilities list to travse.
+    system.pc.south_bridge.gpu.Status = 0x0290
+
+    # The PCI capabilities are like a linked list. The list has a memory
+    # offset and a capability type ID read by the OS. Make the first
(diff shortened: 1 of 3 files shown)


configs/example/gpufs/system/amdgpu.py
```diff
@@ -185,3 +185,11 @@ def connectGPU(system, args):
         system.pc.south_bridge.gpu.DeviceID = 0x6863
     else:
         panic("Unknown GPU device: {}".format(args.gpu_device))
+
+    system.pc.south_bridge.gpu.Status = 0x0290
```
mattsinc, line 191: It would be nice if you could add a comment explaining these magic values here

mattsinc (changes requested):

configs/example/gpufs/system/amdgpu.py
abmerop: Added comments

mattsinc (approved):

PR #225, opened 2023-08-24
cpu-minor: Separate the reg_index of VecClassReg and VecElemReg

rogerchang23424:
In the RISC-V system, we need to VecClassReg to run RISC-V vector instruction, and VecElemReg is not applicable because the element length of vector can be resizable via vset\*vl\* instruction.

The change will seperate the reg_index for VecReg and VecElemReg to ensure that have the space for VecReg when VecElemReg is not applicable.

Change-Id: I99a82dec273baeee31df89a0ee0f5e87f3ff187c

### DIFF
--- a/src/cpu/minor/scoreboard.cc
+++ b/src/cpu/minor/scoreboard.cc
@@ -62,8 +62,11 @@ Scoreboard::findIndex(const RegId& reg, Index &scoreboard_index)
         ret = true;
         break;
       case VecRegClass:
-      case VecElemClass:
         scoreboard_index = vecRegOffset + reg.index();
+        ret = true;
+        break;
+      case VecElemClass:
+        scoreboard_index = vecRegElemOffset + reg.index();
         ret = true;
         break;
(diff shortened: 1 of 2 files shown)


seanzw (approved):
Looks good to me. Just to confirm, this ensures that scoreboard can correctly tracking `VecRegClass` when `VecElemClass` is not specified. I want to understand what does it mean when `VecElemClass` is unspecified.  Maybe `reg_classes.at(VecElemClass)->numRegs()` returns 0?

rogerchang23424:
@seanzw, changed to `applicable `.

Generally, if the register classes are not applicable. They usually put the register classes in the `src/arch/{isa}/isa.cc`. Take MIPS as example, all of not applicable register classes are in [src/arch/mips/isa.cc](https://github.com/gem5/gem5/blob/stable/src/arch/mips/isa.cc#L102). In contrast, the register classes should be found in `src/arch/{isa}/regs/` folder

Thanks

PR #228, opened 2023-08-25
arch-riscv: Enable RVV run in Minor and O3 CPU

### LINKED ISSUE #144: RISC-V RVV vset*vl* instructions can execute bad in out-of-order (O3) mode
**Describe the bug**
RVV vset\*vl\* instructions can be executed bad in out-of-order (O3) mode. If vset\*vl\* is in mis-predicted branch it changes the 'VL' value in decoder and the vector instructions afterwards are decoded/executed in correspondence with 'VL' value which was set in squashed instruction.

**Affects version**
Found in [arch-riscv: Relation chain on RVV support #83](https://github.com/gem5/gem5/pull/83)  PR.

**To Reproduce**
The reproducer is in https://github.com/vmskv/gem5/tree/vmskv/rvv_bad_o3_vsetvl_execuion/riscv_o3_bad_vsetvl_execution_reproducer 
The steps to reproduce are in [README.txt](https://github.com/vmskv/gem5/blob/vmskv/rvv_bad_o3_vsetvl_execuion/riscv_o3_bad_vsetvl_execution_reproducer/README.txt)

**Expected behavior**
If vset\*vl\* is squashed the last previous correct 'VL' value should be used to decode the vector instructions.

**Host Operating System**
Ubuntu 20.04

**Guest ISA**
RISC-V.

rogerchang23424:
Changes in the PR:

1. Change the vset\*vl\* instructions to jump/branch family, and implement the branchTarget.
2. Move the Vl and Vtype from decoder to PCState
3. get VL, VTYPE and VLENB value from PCState
4. Remove vtype checking in construction so that the minor and o3 cpu and decode the instructions after the vset\*vl\*

### DIFF
--- a/src/arch/riscv/isa.cc
+++ b/src/arch/riscv/isa.cc
@@ -40,6 +40,7 @@
 #include "arch/riscv/mmu.hh"
 #include "arch/riscv/pagetable.hh"
 #include "arch/riscv/pmp.hh"
+#include "arch/riscv/pcstate.hh"
 #include "arch/riscv/regs/float.hh"
 #include "arch/riscv/regs/int.hh"
 #include "arch/riscv/regs/misc.hh"
@@ -503,9 +504,19 @@ ISA::readMiscReg(RegIndex idx)
         }
       case MISCREG_VLENB:
         {
-            return VLENB;
+            auto rpc = tc->pcState().as<PCState>();
+            return rpc.vlenb();
         }
-        break;
+      case MISCREG_VTYPE:
+        {
+            auto rpc = tc->pcState().as<PCState>();
+            return rpc.vtype();
+        }
+      case MISCREG_VL:
+        {
+            auto rpc = tc->pcState().as<PCState>();
+            return (RegVal)rpc.vl();
+        }
       case MISCREG_VCSR:
         {
             return readMiscRegNoEffect(MISCREG_VXSAT) &
--- a/src/arch/riscv/pcstate.hh
+++ b/src/arch/riscv/pcstate.hh
@@ -56,15 +56,22 @@ constexpr enums::RiscvType RV64 = enums::RV64;
 
 class PCState : public GenericISA::UPCState<4>
 {
-  private:
+  protected:
+    typedef GenericISA::UPCState<4> Base;
+
     bool _compressed = false;
     RiscvType _rvType = RV64;
 
   public:
+    PCState(const PCState &other) : Base(other), _rvType(other._rvType)
+    {}
+    PCState &operator=(const PCState &other) = default;
     PCState() = default;
-    PCState(const PCState &other) = default;
-    PCState(Addr addr, RiscvType rvType) : UPCState(addr), _rvType(rvType)
+    explicit PCState(Addr addr) { set(addr); }
+    explicit PCState(Addr addr, RiscvType rvType)
     {
+        set(addr);
+        _rvType = rvType;
     }
 
     PCStateBase *clone() const override { return new PCState(*this); }
@@ -84,14 +91,28 @@ class PCState : public GenericISA::UPCState<4>
     void rvType(RiscvType rvType) { _rvType = rvType; }
     RiscvType rvType() const { return _rvType; }
 
+    uint64_t size() const { return _compressed ? 2 : 4; }
+
     bool
     branching() const override
     {
-        if (_compressed) {
-            return npc() != pc() + 2 || nupc() != upc() + 1;
-        } else {
-            return npc() != pc() + 4 || nupc() != upc() + 1;
-        }
+        return npc() != pc() + size() || nupc() != upc() + 1;
+    }
+
+    void
+    serialize(CheckpointOut &cp) const override
+    {
+        Base::serialize(cp);
+        SERIALIZE_SCALAR(_rvType);
+        SERIALIZE_SCALAR(_compressed);
+    }
+
+    void
+    unserialize(CheckpointIn &cp) override
+    {
+        Base::unserialize(cp);
+        UNSERIALIZE_SCALAR(_rvType);
+        UNSERIALIZE_SCALAR(_compressed);
     }
 };
 
--- a/src/arch/riscv/pcstate.hh
+++ b/src/arch/riscv/pcstate.hh
@@ -43,6 +43,7 @@
 #define __ARCH_RISCV_PCSTATE_HH__
 
 #include "arch/generic/pcstate.hh"
+#include "arch/riscv/regs/vector.hh"
 #include "enums/RiscvType.hh"
 
 namespace gem5
@@ -61,17 +62,23 @@ class PCState : public GenericISA::UPCState<4>
 
     bool _compressed = false;
     RiscvType _rvType = RV64;
+    uint64_t _vlenb = VLENB;
+    VTYPE _vtype = (1ULL << 63); // vtype.vill = 1 at initial;
+    uint32_t _vl = 0;
 
   public:
-    PCState(const PCState &other) : Base(other), _rvType(other._rvType)
+    PCState(const PCState &other) : Base(other),
+        _rvType(other._rvType), _vlenb(other._vlenb),
+        _vtype(other._vtype), _vl(other._vl)
     {}
     PCState &operator=(const PCState &other) = default;
     PCState() = default;
     explicit PCState(Addr addr) { set(addr); }
-    explicit PCState(Addr addr, RiscvType rvType)
+    explicit PCState(Addr addr, RiscvType rvType, uint64_t vlenb = VLENB)
     {
         set(addr);
         _rvType = rvType;
+        _vlenb = vlenb;
     }
 
     PCStateBase *clone() const override { return new PCState(*this); }
@@ -83,6 +90,9 @@ class PCState : public GenericISA::UPCState<4>
         auto &pcstate = other.as<PCState>();
         _compressed = pcstate._compressed;
         _rvType = pcstate._rvType;
+        _vlenb = pcstate._vlenb;
+        _vtype = pcstate._vtype;
+        _vl = pcstate._vl;
     }
 
     void compressed(bool c) { _compressed = c; }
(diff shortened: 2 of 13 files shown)


rogerchang23424:
Related issue: #144
To enable run RVV in minor, #225 is required

BobbyRBruce (dismissed):
Loving the progress we're making on RVV! Great work!

BobbyRBruce:
I'll give this another day to give others a chance to see this. Looks good to me but I'm not a RISC-V expert.

@hnpl : Can you look this over?

rogerchang23424:
cc: @vmskv @adriaarmejach @alvaromorenomv

aarmejach:
The code LGTM to enable `minor` and `o3` now. But I have a questions and a concern:

Why do we need to replicate the `vtype` and  `vl` register state in decode? I think that using the `MISCREG` architectural values will simplify things. 

My concern is that with this path the issue #144 might still trigger. Say you execute and commit a `vsetvl` instruction, changing `vtype` and `vl` state, and then the `vsetvl` instructions is squashed due to a previous branch missprediction. The newly decoded instructions after recovering from the missprediction will see the modified `vtype` and `vl` state. Maybe I am wrong, but this might still be an issue.

I am all in on enabling Minor and O3 ASAP, so if #144 is easy to fix in this patch that would be great. 

Since we are discussing, I am interested in trying to implement a proposal that does not always squash after. Making `vsetvl`instructions behave like branches that update `vtype` and `vl` state at commit time make things quite easy. If at commit time `new_vtype != previous_vtype or new_vl != previous_vl` you squashAll (method already exists), otherwise do nothing. With this approach decode can use the current architectural state for `vtype` and `vl`. This will enable loops that have the same `vsetvl` instruction, common case, to execute much faster.

rogerchang23424:
> The code LGTM to enable `minor` and `o3` now. But I have a questions and a concern:
> 
> Why do we need to replicate the `vtype` and `vl` register state in decode? I think that using the `MISCREG` architectural
> values will simplify things.

It's better to remove the `vtype` and `vl` in decoder to make it more simply. However, the `StaticInst` of vector instructions require the `vtype` and `vl` parameters to decode(construct) them.

> My concern is that with this path the issue #144 might still trigger. Say you execute and commit a `vsetvl` instruction, changing `vtype` and `vl` state, and then the `vsetvl` instructions is squashed due to a previous branch missprediction. The newly decoded instructions after recovering from the missprediction will see the modified `vtype` and `vl` state. Maybe I am wrong, but this might still be an issue.

I can't guarantee the issue #144 will be solved after the change applied, I have tested the vector with https://github.com/ksco/riscv-tests/tree/spike/isa/rv64uv, [rvv-intrinsic-doc](https://github.com/riscv-non-isa/rvv-intrinsic-doc)(all of examples are passed, except for str* since it require `vleff*` be implemented ) and [riscv_o3_bad_vsetvl_execution_reproducer](https://github.com/vmskv/gem5/tree/vmskv/rvv_bad_o3_vsetvl_execuion/riscv_o3_bad_vsetvl_execution_reproducer),all of them are passed. The example of #144 shows the correct result for both atomic and o3 cpu.

> I am all in on enabling Minor and O3 ASAP, so if #144 is easy to fix in this patch that would be great.
> 
> Since we are discussing, I am interested in trying to implement a proposal that does not always squash after. Making `vsetvl`instructions behave like branches that update `vtype` and `vl` state at commit time make things quite easy. If at commit time `new_vtype != previous_vtype or new_vl != previous_vl` you squashAll (method already exists), otherwise do nothing. With this approach decode can use the current architectural state for `vtype` and `vl`. This will enable loops that have the same `vsetvl` instruction, common case, to execute much faster.

OK, I will change the vset\*vl\* instruction to branch like instruction,  I have saw the implementation before https://github.com/hnpl/plct-gem5-1/commits/rvv-cpu.

rogerchang23424:
Rebase only

hnpl:
>> The code LGTM to enable minor and o3 now. But I have a questions and a concern:
>>
>> Why do we need to replicate the vtype and vl register state in decode? I think that using the MISCREG architectural
values will simplify things.

> It's better to remove the vtype and vl in decoder to make it more simply. However, the StaticInst of vector instructions require the vtype and vl parameters to decode(construct) them.

I think having the decoder maintaining its own state of vtype and vl regs is okay because it looks like the RISCV decoder does not have access to the architectural state of the core, i.e. the core's ThreadContext (tc) or ExecContext (xc). 

> OK, I will change the vset*vl* instruction to branch like instruction, I have saw the implementation before https://github.com/hnpl/plct-gem5-1/commits/rvv-cpu.

I want to note that this implementation takes a different direction than having the vtype and vl states in decoder.
Instead of having vtype and vl in decoder, the vtype and vl of an instruction are stored in the instruction's PCState, which is a part of every DynInst.

Having vtype and vl in PCState is necessary if the vconfig instructions are treated as branch instructions. This is because for branch instructions, the core will determine whether the branch instruction is correctly executed by comparing the PCStates. This is reflected in this change here, https://github.com/hnpl/plct-gem5-1/commit/b061650493901143ccf60585ef87e46687e2949a#diff-366ad84d19b6100d495edd74366ab39abe982fb2265e611baa839aa9d513b3d9R101.

This is how the vtype and vl are updated,
https://github.com/hnpl/plct-gem5-1/commit/3013eb0c5738bdc588d4d830cf33351d562176a7#diff-0c78d5a93acb1d90c9296beb8cfee1f57b97aec841507f8681158cb529168fa5R90

This is where the next vector instruction acquires the vtype and vl from the previous instruction,
https://github.com/hnpl/plct-gem5-1/commit/8d946c742ea093445e406e518a7e6b485e0f2a75#diff-93a5a8581aba54897a09350ea3f735d79ef1fd441f1970010c765647f6fb4489R130

rogerchang23424:
Thanks @hnpl

src/arch/riscv/isa.cc
```diff
@@ -503,9 +504,26 @@ ISA::readMiscReg(RegIndex idx)
...
+      case MISCREG_VTYPE:
+        {
+            auto rpc = tc->pcState().as<PCState>();
+            RegVal val = 0;
+            VTYPE vtype = rpc.vtype();
+            val = (RegVal)vtype;
+            if (_rvType == RV64) {
+                val &= ~(1 << 31);
+                val |= (vtype.vill << 63);
+            }
```
aarmejach: This made me realize that the current VTYPE BitUnion is defined as 32bit. I think RVV requires at least RV64, so VTYPE should be at least 64 bits. This would also simplify this code, as the cast to RegVal would be enough.

src/arch/riscv/isa.cc
```diff
@@ -503,9 +504,26 @@ ISA::readMiscReg(RegIndex idx)
...
+            return rpc.vlenb();
+        }
+      case MISCREG_VTYPE:
+        {
+            auto rpc = tc->pcState().as<PCState>();
+            RegVal val = 0;
+            VTYPE vtype = rpc.vtype();
+            val = (RegVal)vtype;
+            if (_rvType == RV64) {
+                val &= ~(1 << 31);
```
aarmejach: If we keep this line I'd change the `~(1 << 31)` for the actual mask or `~(1U << 31)`. Just to avoid issues on 32-bit systems or compilers interpreting it as a 64-bit constant.

src/arch/riscv/pcstate.hh
```diff
@@ -56,15 +57,28 @@ constexpr enums::RiscvType RV64 = enums::RV64;
 
 class PCState : public GenericISA::UPCState<4>
 {
-  private:
+  protected:
+    typedef GenericISA::UPCState<4> Base;
+
     bool _compressed = false;
     RiscvType _rvType = RV64;
```
aarmejach, line 64: Should this be initialized with `isa->rvType()`?

src/arch/riscv/pcstate.hh
```diff
@@ -56,15 +57,28 @@ constexpr enums::RiscvType RV64 = enums::RV64;
...
+    uint64_t _vlenb = 32;
+    VTYPE _vtype = 0x80000000; // vtype.vill = 1 at initial;
```
aarmejach: If VTYPE BitUnion is changed to 64 bits this needs to be updated

aarmejach (commented):
This is great! I was suggesting to go this way on a future iteration :)

Overall this LGTM, I have left a few minor coments.

src/arch/riscv/isa.cc
rogerchang23424: Change VTYPE to BitUnion64

src/arch/riscv/isa.cc
rogerchang23424: Change VTYPE to BitUnion64

src/arch/riscv/pcstate.hh
rogerchang23424: It has initialized in isa.newPCState https://github.com/gem5/gem5/blob/stable/src/arch/riscv/isa.hh#L90

rogerchang23424:
```
Logging call to command: /__w/gem5/gem5/build/ALL/gem5.opt -d /tmp/gem5outppd_thks -re --silent-redirect /__w/gem5/gem5/configs/example/gem5_library/checkpoints/riscv-hello-restore-checkpoint.py
Starting Test Suite: test-gem5-library-riscv-hello-restore-checkpoint-ALL-x86_64-opt 
Starting Test Case: test-gem5-library-riscv-hello-restore-checkpoint-ALL-x86_64-opt
Test: test-gem5-library-riscv-hello-restore-checkpoint-ALL-x86_64-opt Failed
Starting Test Case: test-gem5-library-riscv-hello-restore-checkpoint-ALL-x86_64-opt-MatchRegex
Test: test-gem5-library-riscv-hello-restore-checkpoint-ALL-x86_64-opt-MatchRegex Failed
```

@BobbyRBruce, We need to run `util/cpt_upgrader.py` to update the `riscv-hello-example-checkpoint` resource

aarmejach (approved):
LGTM but it'd be great if we can get another +1.

powerjg (approved):
Wow! Way simpler than I expected!

Thanks for putting this together!

powerjg:
@BobbyRBruce, we're just waiting on you to update the checkpoints in gem5-resources. Let me know if I can help.

BobbyRBruce (approved):
Sorry for the delay. The resource is updated and the tests should pass now.

PR #230, opened 2023-08-25
gpu-compute: Use timing DMAs for GPUFS HSA signals

abmerop:
The functional HSA signal read was a hack left in the gpu-compute code. In full system, this functional read is causing problems occasionally with the translation not yet being in the page table. The error message output by gem5 was a fatal message on the readBlob method in port proxy. Changing this to a timing DMA fixes this problem.

This commit adds the various timing DMA functions to send and receive response and clean up. A helper method "sendCompletionSignal" is added to the GPUCommandProcessor because the indentation level was getting too deep. This change applies only to FS mode. Code for SE mode is equivalent to what it was before this commit.

Change-Id: I1bfcaa0a52731cdf9532a7fd0eb06ab2f0e09d48


mattsinc (approved):
I assume SE mode can't use the same support because of something it doesn't have?

abmerop:
> I assume SE mode can't use the same support because of something it doesn't have?

Square was hanging when I tried it.  I'm not invested enough in SE mode to spend time debugging why, hence the FullSystem conditional

PR #231, opened 2023-08-26
gpu-compute: Flat scratch implementation and bug fixes

abmerop:
Add commits fixing private segment counters, flat scratch address
calculation, and implementation of flat scratch instructions.

These commits were tested using a modified version of 'square':

template <typename T>
__global__ void
scratch_square(T *C_d, T *A_d, size_t N)
{
    size_t offset = (blockIdx.x * blockDim.x + threadIdx.x);
    size_t stride = blockDim.x * gridDim.x ;

    volatile int foo; // Volatile ensures scratch / unoptimized code

    for (size_t i=offset; i<N; i+=stride) {
        foo = A_d[i];
        C_d[i] = foo * foo;
    }
}

### DIFF
--- a/src/arch/amdgpu/vega/insts/op_encodings.cc
+++ b/src/arch/amdgpu/vega/insts/op_encodings.cc
@@ -1546,6 +1546,8 @@ namespace VegaISA
         // The SEG field specifies FLAT(0) SCRATCH(1) or GLOBAL(2)
         if (iFmt->SEG == 0) {
             setFlag(Flat);
+        } else if (iFmt->SEG == 1) {
+            setFlag(FlatScratch);
         } else if (iFmt->SEG == 2) {
             setFlag(FlatGlobal);
         } else {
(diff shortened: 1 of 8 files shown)


src/arch/amdgpu/vega/insts/op_encodings.cc
```diff
@@ -1573,12 +1575,12 @@ namespace VegaISA
     Inst_FLAT::initOperandInfo()
     {
         // One of the flat subtypes should be specified via flags
-        assert(isFlat() ^ isFlatGlobal());
+        assert(isFlat() ^ isFlatGlobal() ^ isFlatScratch());
```
mattsinc, line 1578: Not relevant to this commit, but why is this an XOR (^) instead of an OR (|)?

mattsinc (approved):
Thanks, @abmerop !

src/arch/amdgpu/vega/insts/op_encodings.cc
abmerop: It must be exactly one and only one of those types, as that will determine which memory pipeline the request goes to and also how the address will be calculated

PR #236, opened 2023-08-29
util-docker: Add GitHub Action to create Docker Images

BobbyRBruce:
This is built to test the following assumptions:

1. We can trigger a GitHub action event on the changing of a file/directory.
2. We can use GitHub actions to build a docker image.
3. We can use GitHub actions to push a docker image to a container registry.
4. We can use GitHub's container registry.

Right now this will only build and push ubuntu-20.04_all-depenencies, as a test.

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -1,0 +1,42 @@
+#
+name: Docker images build and push
+
+on:
+  push:
+    branches:
+      - develop
+    paths:
+      - 'util/docker/ubuntu-20.04_all-depenencies'
+  workflow_dispatch:
+
+env:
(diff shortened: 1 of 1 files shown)


BobbyRBruce:
@mkjost0 : Is this ok with you (if so, approve it via a review)? I'm wanting to see if this works. Unfortunately there's no way for me to test this all locally so I'm having to merge it and see what happens. Though it shouldn't interfere with anything. At worst it's a broken workflow.

If this works we can automate the building of our docker images but also use GitHub's container registry.

If it works I will ask you to update some tests to use the image from the GitHub registry and ensure it works. I've found out GitHub doesn't charge for image downloads to runners (hosted or self-hosted).

mkjost0 (approved):

PR #241, opened 2023-08-30
configs,stdlib,tests: Remove get_runtime_isa()

### LINKED ISSUE #216: Port gem5 common modules to work in multi-ISA setups
As mentioned in #215, common clients (like se.py) are not multi-ISA compatible.

```
Exception: Cannot determine the the runtime ISA. Either the 'TARGET_ISA' parameter must be set or the binary only compiled to one ISA.

    At:
      src/python/gem5/runtime.py(84): get_runtime_isa
      .../gem5/configs/common/Caches.py(101): PageTableWalkerCache
      .../gem5/configs/common/Caches.py(91): <module>
      <frozen importlib._bootstrap>(241): _call_with_frames_removed
      <frozen importlib._bootstrap_external>(940): exec_module
      <frozen importlib._bootstrap>(705): _load_unlocked
      <frozen importlib._bootstrap>(1150): _find_and_load_unlocked
      <frozen importlib._bootstrap>(1176): _find_and_load
      .../gem5/configs/common/CacheConfig.py(48): <module>
      <frozen importlib._bootstrap>(241): _call_with_frames_removed
      <frozen importlib._bootstrap_external>(940): exec_module
      <frozen importlib._bootstrap>(705): _load_unlocked
      <frozen importlib._bootstrap>(1150): _find_and_load_unlocked
      <frozen importlib._bootstrap>(1176): _find_and_load
      <frozen importlib._bootstrap>(241): _call_with_frames_removed
      <frozen importlib._bootstrap>(1233): _handle_fromlist
      .../gem5/configs/deprecated/example/se.py(61): <module>
```

It is likely possible to work around the problem, but the real solution is to stop using the [get_runtime_isa](https://github.com/gem5/gem5/blob/stable/src/python/gem5/runtime.py#L55) function in those files and embrace having multiple ISAs. The ISA of a specific processing element can be retrieved by the cpu object itself rather than the TARGET_ISA environment variable.

This would probably be a mid-term solution until we eventually deprecate/remove the common library from gem5

BobbyRBruce:
`get_runtime_isa()` has been deprecated for some time. It is a leftover piece of code from when gem5 was compiled to a single ISA and that ISA used to configure the simulated system to use that ISA. Since multi-ISA compilations are possible, `get_runtime_isa()` should not be used. Unless the gem5 binary is compiled to a single ISA, a failure will occur.

The new proceedure for specify which ISA to use is by the setting of the correct `BaseCPU` implementation. E.g., `X86SimpleTimingCPU` of `ArmO3CPU`.

This patch removes the remaining `get_runtime_isa()` instances and removes the function itself. The `SimpleCore` class has been updated to allow for it's CPU factory to return a class, needed by scripts in "configs/common".

The deprecated functionality in the standard library, which allowed for the specifying of an ISA when setting up a processor and/or core has also been removed. Setting an ISA is now manditory.

Fixes #216.

### DIFF
--- a/configs/common/CacheConfig.py
+++ b/configs/common/CacheConfig.py
@@ -47,7 +47,6 @@ import m5
 from m5.objects import *
 
 from gem5.isas import ISA
-from gem5.runtime import get_runtime_isa
 
 
 def _get_hwp(hwp_option):
@@ -118,9 +117,6 @@ def config_cache(options, system):
             None,
         )
 
-        if get_runtime_isa() in [ISA.X86, ISA.RISCV]:
-            walk_cache_class = PageTableWalkerCache
-
     # Set the cache line size of the system
     system.cache_line_size = options.cacheline_size
 
@@ -151,11 +147,13 @@ def config_cache(options, system):
             icache = icache_class(**_get_cache_opts("l1i", options))
             dcache = dcache_class(**_get_cache_opts("l1d", options))
 
-            # If we have a walker cache specified, instantiate two
-            # instances here
-            if walk_cache_class:
-                iwalkcache = walk_cache_class()
-                dwalkcache = walk_cache_class()
+            # If we are using ISA.X86 or ISA.RISCV, we set walker caches.
+            if ObjectList.CPUList().get_isa(options.cpu_type) in [
+                ISA.RiscvCPU,
+                ISA.X86CPU,
+            ]:
+                iwalkcache = PageTableWalkerCache()
+                dwalkcache = PageTableWalkerCache()
             else:
                 iwalkcache = None
                 dwalkcache = None
@@ -193,7 +191,11 @@ def config_cache(options, system):
             # on these names.  For simplicity, we would advise configuring
             # it to use this naming scheme; if this isn't possible, change
             # the names below.
-            if get_runtime_isa() in [ISA.X86, ISA.ARM, ISA.RISCV]:
+            if ObjectList.CPUList().get_isa(options.cpu_type) in [
+                ISA.X86,
+                ISA.ARM,
+                ISA.RISCV,
+            ]:
                 system.cpu[i].addPrivateSplitL1Caches(
                     ExternalCache("cpu%d.icache" % i),
                     ExternalCache("cpu%d.dcache" % i),
--- a/configs/common/Caches.py
+++ b/configs/common/Caches.py
@@ -41,7 +41,6 @@ from m5.defines import buildEnv
 from m5.objects import *
 
 from gem5.isas import ISA
-from gem5.runtime import get_runtime_isa
 
 # Base implementations of L1, L2, IO and TLB-walker caches. There are
 # used in the regressions and also as base components in the
@@ -87,21 +86,3 @@ class IOCache(Cache):
     mshrs = 20
     size = "1kB"
     tgts_per_mshr = 12
-
-
-class PageTableWalkerCache(Cache):
-    assoc = 2
-    tag_latency = 2
-    data_latency = 2
-    response_latency = 2
-    mshrs = 10
-    size = "1kB"
-    tgts_per_mshr = 12
-
-    # the x86 table walker actually writes to the table-walker cache
-    if get_runtime_isa() in [ISA.X86, ISA.RISCV]:
-        is_read_only = False
-    else:
-        is_read_only = True
-        # Writeback clean lines as well
-        writeback_clean = True
--- a/configs/common/Options.py
+++ b/configs/common/Options.py
@@ -45,6 +45,8 @@ import m5
 from m5.defines import buildEnv
 from m5.objects import *
 
+from gem5.runtime import get_supported_isas
+
 vio_9p_help = """\
 Enable the Virtio 9P device and set the path to share. The default 9p path is
 m5ou5/9p/share, and it can be changed by setting VirtIO9p.root with --param. A
@@ -250,7 +252,7 @@ def addCommonOptions(parser):
     )
     parser.add_argument(
         "--cpu-type",
-        default="AtomicSimpleCPU",
+        default=list(get_supported_isas())[0].name + "AtomicSimpleCPU",
         choices=ObjectList.cpu_list.get_names(),
         help="type of cpu to run with",
     )
--- a/configs/common/Simulation.py
+++ b/configs/common/Simulation.py
@@ -81,7 +81,10 @@ def setCPUClass(options):
             TmpClass, test_mem_mode = getCPUClass(options.restore_with_cpu)
     elif options.fast_forward:
         CPUClass = TmpClass
-        TmpClass = AtomicSimpleCPU
+        TmpClass = getCPUClass(
+            ObjectList.CPUList().get_isa(options.cpu_type).name
+            + "AtomicSimpleCPU"
+        )
         test_mem_mode = "atomic"
 
     # Ruby only supports atomic accesses in noncaching mode
--- a/configs/deprecated/example/se.py
+++ b/configs/deprecated/example/se.py
@@ -55,7 +55,6 @@ from m5.util import (
 )
 
 from gem5.isas import ISA
-from gem5.runtime import get_runtime_isa
 
 addToPath("../../")
 
@@ -119,7 +118,7 @@ def get_processes(args):
         idx += 1
 
     if args.smt:
-        assert args.cpu_type == "DerivO3CPU"
+        assert isinstance(args.cpu_type, DerivO3CPU)
         return multiprocesses, idx
     else:
         return multiprocesses, 1
@@ -150,7 +149,7 @@ if args.bench:
 
     for app in apps:
         try:
-            if get_runtime_isa() == ISA.ARM:
+            if ObjectList.CPUList().get_isa(args.cpu_type) == ISA.ARM:
                 exec(
                     "workload = %s('arm_%s', 'linux', '%s')"
                     % (app, args.arm_iset, args.spec_input)
@@ -165,7 +164,7 @@ if args.bench:
             multiprocesses.append(workload.makeProcess())
         except:
             print(
-                f"Unable to find workload for {get_runtime_isa().name()}: {app}",
+                f"Unable to find workload for ISA: {app}",
                 file=sys.stderr,
             )
             sys.exit(1)
--- a/configs/example/hmc_hello.py
+++ b/configs/example/hmc_hello.py
@@ -37,17 +37,23 @@ import m5
 from m5.objects import *
 from m5.util import *
 
-from gem5.runtime import get_runtime_isa
-
 addToPath("../")
 
 from common import (
     HMC,
     MemConfig,
+    ObjectList,
 )
 
 pd = "Simple 'hello world' example using HMC as main memory"
 parser = argparse.ArgumentParser(description=pd)
+parser.add_argument(
+    "--cpu-type",
+    type=str,
+    default="X86TimingSimpleCPU",
+    choices=ObjectList.CPUList().get_names(),
+    help="CPU model to use",
+)
 HMC.add_options(parser)
 options = parser.parse_args()
 # create the system we are going to simulate
@@ -58,8 +64,8 @@ system.mem_mode = "timing"
 clk = "1GHz"
 vd = VoltageDomain(voltage="1V")
 system.clk_domain = SrcClockDomain(clock=clk, voltage_domain=vd)
-# create a simple CPU
-system.cpu = TimingSimpleCPU()
+# create a CPU
+system.cpu = ObjectList().get(options.cpu_type)()
 # config memory system
 MemConfig.config_mem(options, system)
 # hook the CPU ports up to the membus
@@ -70,10 +76,15 @@ system.cpu.createInterruptController()
 # connect special port in the system to the membus. This port is a
 # functional-only port to allow the system to read and write memory.
 system.system_port = system.membus.cpu_side_ports
-# get ISA for the binary to run.
-isa = get_runtime_isa()
 # run 'hello' and use the compiled ISA to find the binary
-binary = "tests/test-progs/hello/bin/" + isa.name.lower() + "/linux/hello"
+
+
+binary = (
+    "tests/test-progs/hello/bin/"
+    + ObjectList.CPUList().get_isa(options.cpu_type).name.lower()
+    + "/linux/hello"
+)
+
 # create a process for a simple "Hello World" application
 process = Process()
 # cmd is a list which begins with the executable (like argv)
--- a/configs/ruby/Ruby.py
+++ b/configs/ruby/Ruby.py
@@ -48,7 +48,6 @@ from m5.util import (
 )
 
 from gem5.isas import ISA
-from gem5.runtime import get_runtime_isa
 
 addToPath("../")
 
@@ -62,8 +61,8 @@ from topologies import *
 
 
 def define_options(parser):
-    # By default, ruby uses the simple timing cpu
-    parser.set_defaults(cpu_type="TimingSimpleCPU")
+    # By default, ruby uses the simple timing cpu and the X86 ISA
+    parser.set_defaults(cpu_type="X86TimingSimpleCPU")
 
     parser.add_argument(
         "--ruby-clock",
@@ -331,9 +330,9 @@ def send_evicts(options):
     # 1. The O3 model must keep the LSQ coherent with the caches
     # 2. The x86 mwait instruction is built on top of coherence invalidations
     # 3. The local exclusive monitor in ARM systems
-    if options.cpu_type == "DerivO3CPU" or get_runtime_isa() in (
-        ISA.X86,
-        ISA.ARM,
-    ):
+
+    if isinstance(
+        options.cpu_type, DerivO3CPU
+    ) or ObjectList.CPUList().get_isa(options.cpu_type) in [ISA.X86, ISA.ARM]:
         return True
     return False
--- a/configs/common/ObjectList.py
+++ b/configs/common/ObjectList.py
@@ -41,6 +41,7 @@ from textwrap import TextWrapper
 import m5.internal.params
 import m5.objects
 
+from gem5.isas import ISA
 from gem5.runtime import get_supported_isas
 
 
@@ -158,6 +159,38 @@ class CPUList(ObjectList):
                     module, self._is_obj_class
                 ):
                     self._sub_classes[name] = cls
+
+    def get_isa(self, name: str) -> ISA:
+        """For a given CPU (string representation) determine the ISA of the
+        CPU."""
+
+        cls = self.get(name)
+
+        def class_exist(className: str) -> bool:
+            """Check if a class exists."""
+            import types
+
+            result = False
+            try:
+                result = eval("type(" + className + ")") == types.ClassType
+            except NameError:
+                pass
+            return result
+
+        if class_exist(m5.objects.X86CPU) and issubclass(
+            cls, m5.objects.X86CPU
+        ):
+            return ISA.X86
+        elif class_exist(m5.objects.ArmCPU) and issubclass(
+            cls, m5.objects.ArmCPU
+        ):
+            return ISA.ARM
+        elif class_exist(m5.objects.RiscvCPU) and issubclass(
+            cls, m5.objects.RiscvCPU
+        ):
+            return ISA.RISCV
+        else:
+            raise ValueError("Unable to determine CPU ISA.")
 
 
 class EnumList(ObjectList):
--- a/configs/deprecated/example/fs.py
+++ b/configs/deprecated/example/fs.py
@@ -53,7 +53,6 @@ from m5.util import (
 from m5.util.fdthelper import *
 
 from gem5.isas import ISA
-from gem5.runtime import get_runtime_isa
 
 addToPath("../../")
 
@@ -86,9 +85,8 @@ def cmd_line_template():
     return None
 
 
-def build_test_system(np):
+def build_test_system(np, isa: ISA):
     cmdline = cmd_line_template()
-    isa = get_runtime_isa()
     if isa == ISA.MIPS:
         test_sys = makeLinuxMipsSystem(test_mem_mode, bm[0], cmdline=cmdline)
     elif isa == ISA.SPARC:
@@ -384,7 +382,8 @@ else:
 
 np = args.num_cpus
 
-test_sys = build_test_system(np)
+isa = ObjectList.CPUList.get_isa(args.cpu_type)
+test_sys = build_test_system(np, isa)
 
 if len(bm) == 2:
     drive_sys = build_drive_system(np)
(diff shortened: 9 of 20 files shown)


BobbyRBruce:
Just a draft for now. Think this solution is almost there but haven't tested it at all.

giactra:
Hi Bobby, thanks for providing a PR. Let me know when you want me to review it (I see it still marked as a draft)

configs/common/Options.py
```diff
@@ -254,6 +256,13 @@ def addCommonOptions(parser):
         choices=ObjectList.cpu_list.get_names(),
         help="type of cpu to run with",
     )
+
+    parser.add_argument(
+        "--isa",
+        choices=get_isas_str_set(),
+        help="The ISA the simulated system will use",
+    )
```
giactra: How would it work with a multi-ISA simulation? It seems you can only specify one ISA with this option

configs/common/CacheConfig.py
```diff
@@ -117,7 +116,7 @@ def config_cache(options, system):
             None,
         )
 
-        if get_runtime_isa() in [ISA.X86, ISA.RISCV]:
+        if get_isa_from_str(options.isa) in [ISA.X86, ISA.RISCV]:
             walk_cache_class = PageTableWalkerCache
```
giactra: Can't we just remove the walk_cache_class variable and the if clause?
We can handle this down below during cache instantiation:

            if walk_cache_class:
                iwalkcache = walk_cache_class()
                dwalkcache = walk_cache_class()
            else:
                iwalkcache = None
                dwalkcache = None

We can just transform the snippe above and instantiate a PageTableWalker depending on the CPU isa.
Something like (might not be accurate):

```
if isinstance(system.cpus[i], (RISCVCpu, X86Cpu)):
    iwalkcache = PageTableWalkerCache(is_read_only=False,...)
    dwalkcache = PageTableWalkerCache(is_read_only=False,...):
else:
    iwalkcache = None
    dwalkcache = None
```

configs/common/CacheConfig.py
```diff
@@ -192,7 +191,7 @@ def config_cache(options, system):
             # on these names.  For simplicity, we would advise configuring
             # it to use this naming scheme; if this isn't possible, change
             # the names below.
-            if get_runtime_isa() in [ISA.X86, ISA.ARM, ISA.RISCV]:
+            if get_isa_from_str(options.isa) in [ISA.X86, ISA.ARM, ISA.RISCV]:
```
giactra: Also here I would use the system.cpu[i] type

giactra (changes requested):
Thanks Bobby... it makes sense we provide an --isa option in order to generate the right CPU but IMHO once the CPU/ISA is generated we should rely on the SimObject itself when differentiating between different CPUs. In this way we make our code less dependant on the Options.py module which we want to deprecate anyway in the long term

configs/common/Options.py
BobbyRBruce: The option is just a single setting yes, but we could expand it anyway you want to permit more. Ultimately as long as you're compiled with the ISA, we can use it. It's just selecting the right BaseCPU implementation.

configs/common/CacheConfig.py
BobbyRBruce: Totally agree with moving it, and have done so.

I forgot we're iterating on the CPU's here, so yes, `system.cpu[i]` is the safest way.

configs/common/CacheConfig.py
BobbyRBruce: Done

configs/common/Simulation.py
```diff
@@ -52,9 +56,11 @@
 addToPath("../common")
 
 
-def getCPUClass(cpu_type):
+def getCPUClass(cpu_type: str, isa: str):
     """Returns the required cpu class and the mode of operation."""
-    cls = ObjectList.cpu_list.get(cpu_type)
+    cls = SimpleCore.cpu_class_factory(
+        cpu_type=get_cpu_type_from_str(cpu_type), isa=get_isa_from_str(isa)
+    )
```
BobbyRBruce: Note: I'm aware that this isn't all that elegant, I should really move this function out of `SimpleCore` in the stdlib, but it works for now and didn't want to take on that right now.

It's probably worth providing some more factories to allow people to select and configure cores like this.

BobbyRBruce:
> Thanks Bobby... it makes sense we provide an --isa option in order to generate the right CPU but IMHO once the CPU/ISA is generated we should rely on the SimObject itself when differentiating between different CPUs. In this way we make our code less dependant on the Options.py module which we want to deprecate anyway in the long term

I agree. We shouldn't be passing around user parameters as much as we do. Once they setup what they were supposed to setup (The CPU, for example) then they shouldn't really be needed anymore.

I've made the changes where you suggested.

My (minor) complaint is the BaseCPU classes have no direct, clean way for them to tell you what ISA their using. When used to infer it via the gem5 (lone-ISA) build, which was fine, but now it's a bit annoying. Ideally for something as noteworthy as "What ISA does this core use?", we'd have a member function called "get_isa()". However, for now, we have figure it out through the class's inheritance. It's not especially hard but it bothers how hacky it is, and I doubt it's all that obvious to someone who doesn't fully grasp our Object-Oriented design. 

I made a comment about coming up with some functions to get more information out of the BaseCPU object. Something like `get_isa(cpu: BaseCPU) -> ISA` would be a good start if adding a "getter" to the class wasn't feasible.  Anyway, not really asking for anything. Just writing down my thought on this.

BobbyRBruce:
That's for the review @giactra . I wan't completely read when I took it out of draft status but I think it's fine now. I _think_ the tests will pass now... we'll see...

configs/common/Caches.py
```diff
@@ -96,11 +95,11 @@ class PageTableWalkerCache(Cache):
...
-    else:
-        is_read_only = True
-        # Writeback clean lines as well
-        writeback_clean = True
+
+class PageTableWalkerCacheIsReadOnly(PageTableWalkerCache):
+    # Used for cases where the target ISA isn't X86 or RISCV
+    is_read_only = True
+    # Writeback clean lines as well
+    writeback_clean = True
```
giactra: This one is not needed.. The other ISA which was using a PTW cache was Arm.. However I implemented a more realistic version some time ago... So this kind of cache is currently not used by anybody

configs/common/Simulation.py
```diff
@@ -44,6 +44,10 @@
 from common import CpuConfig
 from common import ObjectList
 
+from gem5.components.processors.simple_core import SimpleCore
+from gem5.isas import get_isa_from_str
+from gem5.prcesors.cpu_types import get_cpu_type_from_str
```
giactra: Is this a typo?

configs/common/Simulation.py
giactra: Hi Bobby, it doesn't look unelegant to me, the only issue is that the ObjectList contains a wider set of cpus (from the configs subdirectory) like HPI as an example...

Couldn't we still rely on the ObjectList module and let the user choose the ISA implicitly via the CPU type? I haven't checked but I presume that when you build with ALL you will have something like

ArmAtomicCPU, ArmTmingCPU, [...], RiscvAtomicCPU, RiscvTimingCPU, [...]

Let me know if I am missing something

configs/common/Simulation.py
BobbyRBruce: Yes

configs/common/Simulation.py
BobbyRBruce: I've added code here to utilize ObjectList (which is kind of doing the same thing. My only worry here is we're hard-coding our naming scheme here. Like our CPU implementations having the name `{{ISA in Camel Case}}{{CPU Type}}`. If you're ok with this though, i'm fine with it too.

powerjg: IMO, this is OK for the "legacy" configuration scripts. I *really* want to see us focus our effort solely on the standard library because it's going to be much easier to maintain going forward. So, I think it's OK to use relatively brittle things like assuming a particular naming scheme in these legacy scripts to let them limp along a little longer.

BobbyRBruce:
The threading here seems to have gotten very confused.

@giactra : I think I've addressed all your comments.

giactra:
> The threading here seems to have gotten very confused.
> 
> @giactra : I think I've addressed all your comments.

Thanks Bobby for addressing my comments. There is one though I believe it hasn't been addressed in full, which is:

>> Couldn't we still rely on the ObjectList module and let the user choose the ISA implicitly via the CPU type? I haven't checked but I presume that when you build with ALL you will have something like ArmAtomicCPU, ArmTmingCPU, [...], RiscvAtomicCPU, RiscvTimingCPU, [...]

While the PR is now using ObjectList, we are still relying on the --isa option. The problem with the code as of now is that it still does not support the entire list of CPU types, like HPI, O3_ARM_v7a_3 that do no start with the **capitalize(isa)** prefix.

My suggestion would be to remove the --isa option and to rely on the --cpu-type option only... 

For example if I compile with a single ISA (like Arm), I could choose one of the following

`(choose from 'ArmAtomicSimpleCPU', 'ArmMinorCPU', 'ArmNonCachingSimpleCPU', 'ArmO3CPU', 'ArmTimingSimpleCPU', 'AtomicSimpleCPU', 'BaseAtomicSimpleCPU', 'BaseMinorCPU', 'BaseNonCachingSimpleCPU', 'BaseO3CPU', 'BaseTimingSimpleCPU', 'DerivO3CPU', 'MinorCPU', 'NonCachingSimpleCPU', 'O3CPU', 'TimingSimpleCPU', 'HPI', 'O3_ARM_v7a_3', 'O3_ARM_v7a_3_Etrace', 'ex5_LITTLE', 'ex5_big')`

If I compile with multi-ISA (ALL), I could choose:

`(choose from 'ArmAtomicSimpleCPU', 'ArmMinorCPU', 'ArmNonCachingSimpleCPU', 'ArmO3CPU', 'ArmTimingSimpleCPU', 'BaseAtomicSimpleCPU', 'BaseMinorCPU', 'BaseNonCachingSimpleCPU', 'BaseO3CPU', 'BaseTimingSimpleCPU', 'MipsAtomicSimpleCPU', 'MipsMinorCPU', 'MipsNonCachingSimpleCPU', 'MipsO3CPU', 'MipsTimingSimpleCPU', 'PowerAtomicSimpleCPU', 'PowerMinorCPU', 'PowerNonCachingSimpleCPU', 'PowerO3CPU', 'PowerTimingSimpleCPU', 'RiscvAtomicSimpleCPU', 'RiscvMinorCPU', 'RiscvNonCachingSimpleCPU', 'RiscvO3CPU', 'RiscvTimingSimpleCPU', 'SparcAtomicSimpleCPU', 'SparcMinorCPU', 'SparcNonCachingSimpleCPU', 'SparcO3CPU', 'SparcTimingSimpleCPU', 'X86AtomicSimpleCPU', 'X86KvmCPU', 'X86MinorCPU', 'X86NonCachingSimpleCPU', 'X86O3CPU', 'X86TimingSimpleCPU', 'HPI', 'O3_ARM_v7a_3', 'O3_ARM_v7a_3_Etrace', 'ex5_LITTLE', 'ex5_big')`

In other words the ISA choice will be tightly coupled with the CPU I use. If I want to run a multi-ISA simulation with an X86 ISA and an out of order CPU I would do:

`build/ALL/gem5.opt configs/deprecated/example/fs.py --cpu-type X86O3CPU`

This IMHO has the advantage of being more concise, it is simple and does not require any prefixing trick :)
As a default option for --cpu-type, it is AtomicSimpleCPU now, you could just prefix it with the isa as string.
```

253     parser.add_argument(
254         "--cpu-type",                                                                                                                                     
255         default=get_isas_str_set() + "AtomicSimpleCPU",
256         choices=ObjectList.cpu_list.get_names(),
257         help="type of cpu to run with",
258     ) 

```

Please let me know what you think of it.

P.S Anyway OT we should probably filter out BaseCPU types from the ObjectList output

mattsinc (commented):
This seems fine from a GPU perspective, but just making sure you checked that at least the daily tests pass with it?

configs/common/Simulation.py
```diff
@@ -78,9 +78,10 @@ def setCPUClass(options):
             CPUClass = TmpClass
             TmpClass, test_mem_mode = getCPUClass(options.restore_with_cpu)
     elif options.fast_forward:
-        CPUClass = TmpClass
-        TmpClass = AtomicSimpleCPU
-        test_mem_mode = "atomic"
+        pass
+        # CPUClass = TmpClass
+        # TmpClass = AtomicSimpleCPU
+        # test_mem_mode = "atomic"
```
BobbyRBruce: Here we need to setup an AtomicSimpleCPU for whatever the "target isa" is.

configs/deprecated/example/se.py
```diff
@@ -142,27 +140,27 @@ def get_processes(args):
...
+    #                "workload = %s(buildEnv['TARGET_ISA', 'linux', '%s')"
+    #                % (app, args.spec_input)
+    #            )
+    #        multiprocesses.append(workload.makeProcess())
+    #    except:
+    #        print(
+    ##            f"Unable to find workload for {args.isa}: {app}",
+    #           file=sys.stderr,
+    #        )
+    #        sys.exit(1)
```
BobbyRBruce: I really don't know what this code does but it behaves differently depending on whether your targeting the arm ISA or nto

configs/example/hmc_hello.py
```diff
@@ -68,10 +67,13 @@
...
 system.system_port = system.membus.cpu_side_ports
-# get ISA for the binary to run.
-isa = get_runtime_isa()
 # run 'hello' and use the compiled ISA to find the binary
-binary = "tests/test-progs/hello/bin/" + isa.name.lower() + "/linux/hello"
+
+
+# binary = (
+#    "tests/test-progs/hello/bin/" + options.isa.name.lower() + "/linux/hello"
+# )
```
BobbyRBruce: Here is a problem if we remove the target ISA.

We could change this so we pass the binary as a parameter in this case. Is that ok?

configs/ruby/Ruby.py
```diff
@@ -326,9 +324,10 @@ def send_evicts(options):
...
-    if options.cpu_type == "DerivO3CPU" or get_runtime_isa() in (
-        ISA.X86,
-        ISA.ARM,
-    ):
-        return True
+
+    # if options.cpu_type == "DerivO3CPU" or get_isa_from_str(args.isa) in (
+    #    ISA.X86,
+    #    ISA.ARM,
+    # ):
```
BobbyRBruce: We need to know here if the target ISA is X86 or ARM.

BobbyRBruce:
@giactra I've pushed a new solution to this problem.

I've dropped `--isa` as an input parameter and now just have the user set the CPU Types by passing the class name. This class name is used to construct a CPU core via `ObjectList` and can be used to infer the target ISA of the system. It does this by using the logic that a system targeting ISA Y will have CPUs that are instances of that ISA CPU's Base class. E.g.,:  `is_x86_sim = isinstanceof(ObjectList.get(arg.cpu_type), X86CPU)`. So if user passes `X86TimingSimpleCPU` the scripts would infer an `X86` ISA where needed.

Let me know if this works for you.

giactra (changes requested):
> @giactra I've pushed a new solution to this problem.
> 
> I've dropped `--isa` as an input parameter and now just have the user set the CPU Types by passing the class name. This class name is used to construct a CPU core via `ObjectList` and can be used to infer the target ISA of the system. It does this by using the logic that a system targeting ISA Y will have CPUs that are instances of that ISA CPU's Base class. E.g.,: `is_x86_sim = isinstanceof(ObjectList.get(arg.cpu_type), X86CPU)`. So if user passes `X86TimingSimpleCPU` the scripts would infer an `X86` ISA where needed.
> 
> Let me know if this works for you.

Thank you so much Bobby, this clearly looks like a step in the right direction. I believe we are almost there, there's only one improvement I would like to suggest:

Rather than always relying on the cpu type passed with the command line ->`ObjectList.get_isa(args.cpu_type)`, I would *try* to extrapolate the ISA from the instantiated cpu object itself *whenever it is possible*.

For example if every ArchCPU  (ArmCPU,RiscvCPU, etc) defined a get_isa as a classmethod, I would prefer seeing something like:

```
if system.cpu.get_isa() in ['ISA.X86', ''ISA.ARM]:
    # Do x86 and ARM stuff
```

This IMHO has the following advantages:
1) You wouldn't be required to import ObjectList if you didn't want to
2) It is more readable
3) It works better when you instantiate CPUs with multiple ISAs

Of course I would still keep the current implementation whenever the CPU object is not yet available... But I would also consider re-implementing the get_isa in ObjectList by calling the cpu_type.get_isa directly (it would be a classmethod afterall).

Please let me know if this looks reasonable for you

BobbyRBruce:
> > @giactra I've pushed a new solution to this problem.
> > I've dropped `--isa` as an input parameter and now just have the user set the CPU Types by passing the class name. This class name is used to construct a CPU core via `ObjectList` and can be used to infer the target ISA of the system. It does this by using the logic that a system targeting ISA Y will have CPUs that are instances of that ISA CPU's Base class. E.g.,: `is_x86_sim = isinstanceof(ObjectList.get(arg.cpu_type), X86CPU)`. So if user passes `X86TimingSimpleCPU` the scripts would infer an `X86` ISA where needed.
> > Let me know if this works for you.
> 
> Thank you so much Bobby, this clearly looks like a step in the right direction. I believe we are almost there, there's only one improvement I would like to suggest:
> 
> Rather than always relying on the cpu type passed with the command line ->`ObjectList.get_isa(args.cpu_type)`, I would _try_ to extrapolate the ISA from the instantiated cpu object itself _whenever it is possible_.
> 
> For example if every ArchCPU (ArmCPU,RiscvCPU, etc) defined a get_isa as a classmethod, I would prefer seeing something like:
> 
> ```
> if system.cpu.get_isa() in ['ISA.X86', ''ISA.ARM]:
>     # Do x86 and ARM stuff
> ```
> 
> This IMHO has the following advantages:
> 
> 1. You wouldn't be required to import ObjectList if you didn't want to
> 2. It is more readable
> 3. It works better when you instantiate CPUs with multiple ISAs
> 
> Of course I would still keep the current implementation whenever the CPU object is not yet available... But I would also consider re-implementing the get_isa in ObjectList by calling the cpu_type.get_isa directly (it would be a classmethod afterall).
> 
> Please let me know if this looks reasonable for you

Would it be ok with you if I log this is an enhancement to add soon and we go with this solution for now? I really do want there to be better support for determining an CPU's ISA. In fact I'd quite look forward to working on that, but it seems a little out-of-scope for this PR which is just to remove a deprecated function. A change, like the one you propose, would allow us to refactor a lot of funny logic away, which is great, but i worry this PR is starting to do too much. Right now this PR removes `get_runtime_isa` with only a little additional code in the config scripts where knowing a Core's ISA is required. The necessity for this change was only to get this scripts working in the multi-ISA complication scenario, i think we've achieved that.

configs/example/hmc_hello.py
```diff
@@ -36,16 +36,23 @@
...
+from common import ObjectList
 
 
 pd = "Simple 'hello world' example using HMC as main memory"
 parser = argparse.ArgumentParser(description=pd)
+parser.add_argument(
+    "--cpu-type",
+    type=str,
+    default="X86TimingSimpleCPU",
+    choices=CPUList.get_names(),
```
rogerchang23424:
```
NameError: name 'CPUList' is not defined
```
Please fix

BobbyRBruce: Thanks @rogerchang23424 ! Fixed!

BobbyRBruce:
> > > @giactra I've pushed a new solution to this problem.
> > > I've dropped `--isa` as an input parameter and now just have the user set the CPU Types by passing the class name. This class name is used to construct a CPU core via `ObjectList` and can be used to infer the target ISA of the system. It does this by using the logic that a system targeting ISA Y will have CPUs that are instances of that ISA CPU's Base class. E.g.,: `is_x86_sim = isinstanceof(ObjectList.get(arg.cpu_type), X86CPU)`. So if user passes `X86TimingSimpleCPU` the scripts would infer an `X86` ISA where needed.
> > > Let me know if this works for you.
> > 
> > 
> > Thank you so much Bobby, this clearly looks like a step in the right direction. I believe we are almost there, there's only one improvement I would like to suggest:
> > Rather than always relying on the cpu type passed with the command line ->`ObjectList.get_isa(args.cpu_type)`, I would _try_ to extrapolate the ISA from the instantiated cpu object itself _whenever it is possible_.
> > For example if every ArchCPU (ArmCPU,RiscvCPU, etc) defined a get_isa as a classmethod, I would prefer seeing something like:
> > ```
> > if system.cpu.get_isa() in ['ISA.X86', ''ISA.ARM]:
> >     # Do x86 and ARM stuff
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This IMHO has the following advantages:
> > 
> > 1. You wouldn't be required to import ObjectList if you didn't want to
> > 2. It is more readable
> > 3. It works better when you instantiate CPUs with multiple ISAs
> > 
> > Of course I would still keep the current implementation whenever the CPU object is not yet available... But I would also consider re-implementing the get_isa in ObjectList by calling the cpu_type.get_isa directly (it would be a classmethod afterall).
> > Please let me know if this looks reasonable for you
> 
> Would it be ok with you if I log this is an enhancement to add soon and we go with this solution for now? I really do want there to be better support for determining an CPU's ISA. In fact I'd quite look forward to working on that, but it seems a little out-of-scope for this PR which is just to remove a deprecated function. A change, like the one you propose, would allow us to refactor a lot of funny logic away, which is great, but i worry this PR is starting to do too much. Right now this PR removes `get_runtime_isa` with only a little additional code in the config scripts where knowing a Core's ISA is required. The necessity for this change was only to get this scripts working in the multi-ISA complication scenario, i think we've achieved that.

I've added the issue here: #622

giactra (approved):
Makes sense. Do you mind if I take care of the proposed improvement? Might save us some back and forth. Thanks for this fix Bobby!

powerjg:
@giactra I'll speak for Bobby and say you're welcome to add additional improvements 😄.

Do you want us to hold off merging until you push additional improvements?

giactra:
> 
> Do you want us to hold off merging until you push additional improvements?

No need, as Bobby mentioned in his comment, this PR already does its job in fixing a real bug; I will post my suggested enhancements separately targeting next release

abmerop:
Anyone else experiencing this breaking the send_evicts function in Ruby.py?  GPU SE mode is broken with this fix and I don't know what the problem is.  Wondering if it's happening for other Ruby protocols

mattsinc:
> Anyone else experiencing this breaking the send_evicts function in Ruby.py? GPU SE mode is broken with this fix and I don't know what the problem is. Wondering if it's happening for other Ruby protocols

I guess we would have seen this fail the daily GPU test then?

mattsinc:
> This seems fine from a GPU perspective, but just making sure you checked that at least the daily tests pass with it?

Sorry @abmerop , I should have followed up on this question previously, as it appears the answer was no.

PR #244, opened 2023-08-31
scons: Add an option specifying the path to mold linker binary

hnpl:
To use mold linker with gcc of version older than 12.1.0, the user has to pass the -B option to specify where the linker is. [1]

Currently in gem5, scons only looks for the mold binary at conventional places, such as /usr/libexec/mold and /usr/local/libexec/mold. There's no option to manually specify the path to the linker.

gcc-12 and mold are not widely available on older systems. Having an option to manually input the linker path allows user to build and use mold without sudo permission.

[1] https://github.com/rui314/mold#how-to-use

Change-Id: Ifb2366a0c2342bf4e7207df8db6196e14184a9d4

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -84,6 +84,7 @@ from os import mkdir, remove, environ
 from os.path import abspath, dirname, expanduser
 from os.path import isdir, isfile
 from os.path import join, split
+from pathlib import Path
 
 import logging
 logging.basicConfig()
@@ -121,6 +122,8 @@ AddOption('--ignore-style', action='store_true',
           help='Disable style checking hooks')
 AddOption('--linker', action='store', default=None, choices=linker_options,
           help=f'Select which linker to use ({", ".join(linker_options)})')
+AddOption('--mold-path', action='store', default='/usr/libexec/mold/',
+          help='Path to the binary of mold linker.')
 AddOption('--gold-linker', action='store_const', const='gold', dest='linker',
           help='Use the gold linker. Deprecated: Use --linker=gold')
 AddOption('--no-compress-debug', action='store_true',
@@ -425,19 +428,36 @@ for variant_path in variant_paths:
             conf.CheckLinkFlag('-Wl,--as-needed')
 
         linker = GetOption('linker')
+        # check for mold linker when -fuse-ld=mold fails
+        # this checks whether mold_dir has ld or ld.mold and
+        # whether -B{mold_dir} works
+        mold_check = lambda mold_dir: \
+            isdir(mold_dir) and \
+            (isfile(mold_dir/'ld') or isfile(mold_dir/'ld.mold')) and \
+            conf.CheckLinkFlag('-B' + str(mold_dir))
+        mold_check_message = \
+            "You are seeing this error because -fuse-ld=mold failed.\n" \
+            "If you did not add the folder containing mold to $PATH, you " \
+            "should do so and recompiling gem5.\n" \
+            "If that does not work, you can manually specify the path to " \
+            "the binary of mold using the --mold-path option. This will " \
+            "cause scons to look for ld or ld.mold in the same folder as " \
+            "the binary of mold."
         if linker:
             with gem5_scons.Configure(env) as conf:
                 if not conf.CheckLinkFlag(f'-fuse-ld={linker}'):
-                    # check mold support for gcc older than 12.1.0
-                    if linker == 'mold' and \
-                       (env['GCC'] and \
-                           compareVersions(env['CXXVERSION'],
-                                           "12.1.0") < 0) and \
-                       ((isdir('/usr/libexec/mold') and \
-                           conf.CheckLinkFlag('-B/usr/libexec/mold')) or \
-                       (isdir('/usr/local/libexec/mold') and \
-                           conf.CheckLinkFlag('-B/usr/local/libexec/mold'))):
-                        pass # support mold
+                    # If -fuse-ld=mold fails, we use "-B$mold_path" instead.
+                    mold_path = Path(GetOption('mold_path')).parent
+                    if linker == 'mold':
+                        if any(map(mold_check, [mold_path,
+                            Path("/usr/libexec/mold"),
+                            Path("/usr/libexec/mold")])):
+                            pass # support mold
+                        else:
+                            error(
+                                f'Linker "{linker}" is not supported. '
+                                f'{mold_check_message}'
+                            )
                     else:
                         error(f'Linker "{linker}" is not supported')
                 if linker == 'gold' and not GetOption('with_lto'):
--- a/SConstruct
+++ b/SConstruct
@@ -438,7 +438,7 @@ for variant_path in variant_paths:
         mold_check_message = \
             "You are seeing this error because -fuse-ld=mold failed.\n" \
             "If you did not add the folder containing mold to $PATH, you " \
-            "should do so and recompiling gem5.\n" \
+            "should do so and recompile gem5.\n" \
             "If that does not work, you can manually specify the path to " \
             "the binary of mold using the --mold-path option. This will " \
             "cause scons to look for ld or ld.mold in the same folder as " \


SConstruct
```diff
@@ -428,12 +432,14 @@ for variant_path in variant_paths:
...
             with gem5_scons.Configure(env) as conf:
                 if not conf.CheckLinkFlag(f'-fuse-ld={linker}'):
+                    mold_path = GetOption('mold_path')
                     # check mold support for gcc older than 12.1.0
                     if linker == 'mold' and \
                        (env['GCC'] and \
                            compareVersions(env['CXXVERSION'],
                                            "12.1.0") < 0) and \
-                       ((isdir('/usr/libexec/mold') and \
+                       ((conf.CheckLinkFlag('-B'+mold_path)) or \
```
BobbyRBruce: I think if the user specifies the mold_path and `config.CheckLinkFlag('-B'+mold_path)` fails, then we shouldn't look for alternatives, which we do via the `or`. I think if the user explicitly states path it shouldn't try alternatives. It should just fail.

BobbyRBruce:
> To use mold linker with gcc of version older than 12.1.0, the user has to pass the -B option to specify where the linker is. [1]

Why? I don't get why the user **must** pass the -B flag to enable us to use gcc versions >12.1.0. We already pass the -B flag to GCC for versions <12.1.0. What's so special about the user doing it? Mold surely installs in the same place regardless as to what GCC version it's going be linked into.

Also, this path doesn't remove the check that GCC must be <12.1.0, so this patch doesn't even enable gcc versions >12.1 with mold.

> 
> Currently in gem5, scons only looks for the mold binary at conventional places, such as /usr/libexec/mold and /usr/local/libexec/mold. There's no option to manually specify the path to the linker.

Fair... but what does this have to do with gcc versions >12.1.0?

> gcc-12 and mold are not widely available on older systems. 

What do you mean not available? If they are not available on older systems, why do we care? We can't help users who can't get mold on their system and in those cases where they can't use gcc 12, we're not fixing anything here, right? This patch is to enable mold with gcc-12. 

> Having an option to manually input the linker path allows user to build and use mold without sudo permission.

Ok I think I understand this. I just confused as I read the last part to mean "users need sudo to build mold". Is the following a fair reword?: "Having an option to manually input the mold linker path allows users to link towards built mold instances anywhere on the system, not just the default locations in `/usr` where the may not have permission to install mold install (i.e., no sudo permissions).

If so, this is fair case.

> 
> [1] https://github.com/rui314/mold#how-to-use
> 
> Change-Id: Ifb2366a0c2342bf4e7207df8db6196e14184a9d4
already pa

SConstruct
```diff
@@ -121,6 +121,10 @@ AddOption('--ignore-style', action='store_true',
           help='Disable style checking hooks')
 AddOption('--linker', action='store', default=None, choices=linker_options,
           help=f'Select which linker to use ({", ".join(linker_options)})')
+AddOption('--mold-path', action='store', default='/usr/bin/mold',
+          help='Path to the binary of mold, the linker. For GCC, the path must'
+               ' be a relative path. For clang, the path can be an absoblute'
```
BobbyRBruce: We don't support the mold linker for clang, so it doesn't matter. 

Are you sure for GCC it needs to be a relative path, because the current paths specified, such as `conf.CheckLinkFlag('-B/usr/libexec/mold'))`, appear absolute me.

SConstruct
```diff
@@ -121,6 +121,10 @@ AddOption('--ignore-style', action='store_true',
           help='Disable style checking hooks')
 AddOption('--linker', action='store', default=None, choices=linker_options,
           help=f'Select which linker to use ({", ".join(linker_options)})')
+AddOption('--mold-path', action='store', default='/usr/bin/mold',
+          help='Path to the binary of mold, the linker. For GCC, the path must'
```
BobbyRBruce: Is binary, or is it some kind of directory? If it's just a binary then how does this line work:

```py
 (isdir('/usr/libexec/mold') and \
                           conf.CheckLinkFlag('-B/usr/libexec/mold')) or \
```

This is saying "/usr/libexec/mold" is directory.

hnpl: `/usr/bin/mold` should be a binary.
I'm not very familar with how mold is installed so I keep the original code there.

BobbyRBruce: > `/usr/bin/mold` should be a binary.
> I'm not very familar with how mold is installed so I keep the original code there.

Mold appears to be installed at `/usr/libexec/mold/ld` then symlinked to `/usr/bin/mold`. The `-B` flag is a way to tell the compiler where data files are... i don't know why this helps but apparently it was necessary for GCC <12. gcc-12 and beyond you only need to do the `-fuse-ld=mold` like everything else.

SConstruct
hnpl: Yeah, the part about relative part is not true. I'll modify it.

hnpl:
>> To use mold linker with gcc of version older than 12.1.0, the user has to pass the -B option to specify where the linker is. [1]
> Why? I don't get why the user must pass the -B flag to enable us to use gcc versions >12.1.0.

The context is, when using gcc, user must pass the path to the mold binary the -B flag or -fuse-ld to gcc to use mold with gcc. Prior to 12.1.0, gcc doesn't know what mold is.

The `-B` flag adds an additional place where gcc can look for the `ld` binary. It accepts both abs/relative paths.

The `-fuse-ld=$SOME_BINARY` flag tells gcc to use `$SOME_BINARY` as a linker.  `-fuse-ld` does not accept absolute path.

> We already pass the -B flag to GCC for versions <12.1.0.

The existing -B flags don't work for all systems. Those work nicely if you have mold installed at the default location, but it doesn't work when you have the linker somewhere else.

> Fair... but what does this have to do with gcc versions >12.1.0?

Nothing. The `mold-path` option is only evaluated when the gcc version is < 12.1.0.

> Having an option to manually input the mold linker path allows users to ~link towards~ use built mold instances anywhere on the system, not just the default locations in `/usr` where [they] may not have permission to install mold (i.e., no sudo permissions).

I believe this is better. Will update.

SConstruct
hnpl: Yeah, `-B $PREFIX` will let gcc finding `$PREFIX/ld`, so it should be a directory instead of a binary. I'll update this.

hnpl:
After some testing, I found the following,

- According to https://gcc.gnu.org/onlinedocs/gcc-9.4.0/gcc/Directory-Options.html#index-B,

> For each subprogram to be run, the compiler driver first tries the -B prefix, if any. If that name is not found, or if -B is not specified, the driver tries two standard prefixes, /usr/lib/gcc/ and /usr/local/lib/gcc/. If neither of those results in a file name that is found, the unmodified program name is searched for using the directories specified in your PATH environment variable.

This basically means you can pass anything to "-B $PREFIX" and gcc will fall back to `ld` if "$PREFIX/ld" does not exist while the compilation does not fail,
So, it is possible for gcc to pass the current test for mold, `isdir(somepath) and conf.CheckLinkFlag("-B"+somepath)`, while gcc still uses ld.

- Also, `mold` will install `ld.mold` to the installation place, not `ld`. It seems that *my* version of gcc does not take `ld.mold` into account even though I added `-B/somepath` where `/somepath/ld.mold` exists.

- The above observations are for gcc-9. I tried to use -fuse-ld=mold with gcc-10 and it worked.

```
~/scr/temp/> g++-10 test.cpp -lpthread -fuse-ld=mold
~/scr/temp/> readelf -p .comment a.out

String dump of section '.comment':
  [     0]  GCC: (Ubuntu 10.5.0-1ubuntu1~20.04) 10.5.0
  [    2b]  mold 2.1.0 (d78b7486963a0a64f668ba9859c2395dca3fe3c5; compatible with GNU ld)
```

Note that mold working gcc-10 is contradictory to all official documentation [1] [2]. So I'd suggest to ignore them and drop the gcc_version check and only do the "-B" check if "-fuse-ld=mold" fails.

So, I believe the new check for `mold` should be,

```python
if not conf.CheckLinkFlag("-fuse-ld=mold"):
  if (isdir(mold_path) and isfile(mold_path+"/ld") and conf.CheckLinkFlag("-B"+mold_path):
    # OK
  else:
    # no mold support
```

[1] https://gcc.gnu.org/onlinedocs/gcc-10.5.0/gcc/Link-Options.html
[2] https://github.com/rui314/mold#how-to-use

SConstruct
```diff
@@ -425,19 +428,36 @@ for variant_path in variant_paths:
...
+        # this checks whether mold_dir has ld or ld.mold and
+        # whether -B{mold_dir} works
+        mold_check = lambda mold_dir: \
+            isdir(mold_dir) and \
+            (isfile(mold_dir/'ld') or isfile(mold_dir/'ld.mold')) and \
+            conf.CheckLinkFlag('-B' + str(mold_dir))
+        mold_check_message = \
+            "You are seeing this error because -fuse-ld=mold failed.\n" \
+            "If you did not add the folder containing mold to $PATH, you " \
+            "should do so and recompiling gem5.\n" \
```
BobbyRBruce:
```suggestion
            "should do so and recompile gem5.\n" \
```

BobbyRBruce (dismissed):

BobbyRBruce (approved):

mkjost0 (approved):

abmerop:
Could anyone explain how to actually use this with scons?  Building mold from source installs in /usr/local/bin, but if I use --mold-path=/usr/local/bin it does not work.  It used to work fine with gcc 9.4 for me, so I will revert this change locally until there is some example of how to use this new option with scons.

Edit: Nevermind, figured it out.

hnpl:
I'm surprised that it worked for g++ 9.4 without having the path to mold because, prior to this change, scons only checks for mold at 2 places,

```
- /usr/local/libexec/mold
- /usr/libexec/mold
```

Adding --mold-path=$MOLD basically adds "-B $MOLD/.." to the compilation commands, and in the above case, scons will just use "/usr/local/bin/ld".

I'm not opposed to reverting this though.

PR #250, opened 2023-09-01
misc: 'sim{out/err}' -> 'sim{out/err}.txt'

BobbyRBruce:
By default, the `--stderr-file` and `--stdout-file` arguments were directing the simulator to output files named "simerr" and "simout" respectively if an output redirect was requested.

A small annoyance is these files lack an extension meaning programs refuse to open them, or don't do so withou additional effort. On many systems they are assumed to scripts.

This patch adds the `.txt` extension to both, thus clearly indicating to other programs these are text files and can be opened and read as such.


powerjg:
This is a significant API change since many outside scripts will process these files. Let's be sure this goes through some kind of community feedback phase before merging.

BobbyRBruce:
> This is a significant API change since many outside scripts will process these files. Let's be sure this goes through some kind of community feedback phase before merging.

Yes, that's fair.

It's just a small personal annoyance, If it breaks anything for anyone I'll stall it for now.

src/arch/arm/tracers/tarmac_tracer.cc
```diff
@@ -64,9 +64,9 @@ tarmacDump(const TarmacTracerParams &p)
 {
     switch (p.outfile) {
       case TarmacDump::stdoutput:
-        return simout.findOrCreate("stdout");
+        return simout.findOrCreate("stdout.txt");
       case TarmacDump::stderror:
-        return simout.findOrCreate("stderr");
+        return simout.findOrCreate("stderr.txt");
```
giactra: I believe this is not working as you would imagine. The findOrCreate argument is matched against the stdout and stderr literals to check where to redirect the output stream:

```
OutputStream *
OutputDirectory::checkForStdio(const std::string &name)
{
    if (name == "cerr" || name == "stderr")
        return &stderr;

    if (name == "cout" || name == "stdout")
        return &stdout;

    return NULL;
}
```

https://github.com/gem5/gem5/blob/stable/src/base/output.cc#L135

By passing stdout.txt/stderr.txt comparison will fail and gem5 will always redirect to the txt files. I believe this change here should be reverted. The redirection to the txt file will be handled automatically by the rest of the code

giactra (dismissed):
Thanks Bobby for this PR, LGTM other than a minor issue in the TarmacTracer

util/plot_dram/dram_sweep_plot.py
```diff
@@ -81,7 +81,7 @@ def main():
     try:
         simout = open(sys.argv[2] + "/simout", "r")
```
andysan: This presumably needs to change as well.

BobbyRBruce: Yes. You're right

BobbyRBruce: Fixed.

src/arch/arm/tracers/tarmac_tracer.cc
BobbyRBruce: Thanks @giactra! I don't think I would've ever figured that out by myself.

This is fixed.

BobbyRBruce:
Thanks! I've incorporated the fixes.

powerjg (commented):
lgtm. I can approve if no one else wants to :)

andysan:
LGTM

powerjg (approved):

PR #256, opened 2023-09-01
misc: Add LULESH GPU tests

mkjost0:
This adds the LULESH tests, which currently run successfully, though they still use the gem5-resources directory to build the binary used as opposed to using a pre-built binary

Change-Id: I91c511fe92b7f9d11dfb027f435573f826bc6714


.github/workflows/gpu-tests.yaml
```diff
@@ -0,0 +1,86 @@
...
+    strategy:
+      matrix:
+        image: [ALL, GCN3_X86]
+        # this allows us to pass additional command line parameters
+        # the default is to add -j $(nproc), but some images
+        # require more specifications when built
+        include:
+          - command-line: -j $(nproc)
+          - image: GCN3_X86
+            command-line: -j4 --ignore-style
```
mattsinc: Same thing about j4 here as other ones

mattsinc (commented):
Same question here about docker: I don't see where that is coming from.  Maybe the container on line 56?

abmerop (commented):
Thanks for helping to move the GPU weekly tests out of the bash script!  This looks fine for GCN3.  One thing I would like though is to test the Vega ISA as well. I realize that will take more resources.  @mattsinc and I have been planning to discuss in detail if GCN3 can be deprecated. I would also like to see Vega as part of the ALL build, which would further reduce compilation for testing, I think (not sure if each yaml file is rebuilding).

I have many more detailed thoughts on moving to Vega but it is probably better for a discussions thread rather than this PR.

mkjost0:
> Thanks for helping to move the GPU weekly tests out of the bash script! This looks fine for GCN3. One thing I would like though is to test the Vega ISA as well. I realize that will take more resources. @mattsinc and I have been planning to discuss in detail if GCN3 can be deprecated. I would also like to see Vega as part of the ALL build, which would further reduce compilation for testing, I think (not sure if each yaml file is rebuilding).
> 
> I have many more detailed thoughts on moving to Vega but it is probably better for a discussions thread rather than this PR.

Yes, this is something we can look into! I think if we do that, we can put it into a separate PR so we can update all the GPU tests to use Vega at the same time after we ensure it works locally first, unless you have other thoughts on that.

.github/workflows/gpu-tests.yaml
```diff
@@ -28,6 +28,22 @@ jobs:
           retention-days: 5
       - run: echo "This job's status is ${{ job.status }}."
 
+  LULESH-tests:
+    runs-on: [self-hosted, linux, x64, build]
+    container: gcr.io/gem5-test/gcn-gpu:latest
+    needs: build-gem5
+    timeout-minutes: 600 # 10 hours
```
mattsinc: @mkjost0 similar to other apps, how did you determine this timeout?  My recollection is that LULESH sometimes runs longer than this when running the default 10 iterations -- maybe @abmerop remembers differently though...

mattsinc (dismissed):
Only one question, which may be moot if you verified.

.github/workflows/gpu-tests.yaml
abmerop: A full run takes around 15 hours on my machine.  I usually exit after the first iteration when testing now though. I am not sure how many this test is trying to run

mattsinc: Since there are no input args to the contrary, I believe it is running the full 10 iterations: https://gem5.googlesource.com/public/gem5-resources/+/refs/heads/develop/src/gpu/lulesh/src/lulesh.hip.cc#5507

Your experience squares with mine in terms of runtime.

Harshil2107: Melissa is no longer working on the project. I don't know how she got the timeout number. How long would you recommend setting the timeout?

mattsinc: I was not aware Melissa wasn't working on it anymore, sorry.

Have you tried running LULESH in your setup?  If so, how long does it take?  I would set the timeout accordingly.

abmerop: 24 hours should be sufficient... But I would like to discuss how redundant the run for this app is.  We don't gain much testing the default 10 iterations in this app vs. one, for example.  And with one it would take ~2 hours.

mattsinc: There is an input arg to control the number of iterations, so making this change is straightforward.

In terms of why multiple -- I think there is value in at least running 2 iterations to make sure inter-kernel behavior works.  But I agree 10 is a lot.

src/arch/arm/regs/misc.cc
```diff
@@ -1928,10 +1928,18 @@ faultSctlr2EL1(const MiscRegLUTEntry &entry,
     if (HaveExt(tc, ArmExtension::FEAT_SCTLR2)) {
         const SCR scr = tc->readMiscReg(MISCREG_SCR_EL3);
         const HCRX hcrx = tc->readMiscReg(MISCREG_HCRX_EL2);
-        if (auto fault = faultHcrFgtEL1<read, g_bitfield, &HFGTR::sctlrEL1>(entry, tc, inst);
-            fault != NoFault) {
+        if (
```
mattsinc: I'm not sure what these ARM changes have to do with the LULESH benchmark?

.github/workflows/gpu-tests.yaml
```diff
@@ -29,6 +31,37 @@ jobs:
...
+
+            - name: Obtain LULESH
+              working-directory: ${{ github.workspace }}/lulesh
+              # Obtains the latest LULESH compatible with this version of gem5 via
+              # gem5 Resources.
+              run: build/GCN3_X86/gem5.opt util/obtain-resource.py lulesh -p lulesh
+
+            - name: Run LULUESH tests
+              working-directory: ${{ github.workspace }}
+              run: build/GCN3_X86/gem5.opt configs/example/apu_se.py -n3 --mem-size=8GB --reg-alloc-policy=dynamic --benchmark-root="lulesh" -c lulesh
```
mattsinc: @BobbyRBruce : per @abmerop and I's conversation on the previous version, please use "0.01 2" as the input args and then reduce the timeout to 8 hours.

mattsinc (changes requested):

.github/workflows/gpu-tests.yaml
BobbyRBruce: When you say passing the input args, do you mean like this?: ` build/GCN3_X86/gem5.opt configs/example/apu_se.py -n3 --mem-size=8GB --reg-alloc-policy=dynamic --benchmark-root="lulesh" -c lulesh 0.01 2` ?

src/arch/arm/regs/misc.cc
BobbyRBruce: I'm going to separate this out, but TL;DR I needed to add this here for now to commit. "misc.cc" breaks the line-length rules and my pre-commit was complaining. I'll let you know when it's gone

.github/workflows/gpu-tests.yaml
mattsinc: Yes, that looks correct

BobbyRBruce: Done.

mattsinc (dismissed):
thanks!  Please let @abmerop confirm before you merge in though.

.github/workflows/gpu-tests.yaml
```diff
@@ -29,6 +31,39 @@ jobs:
...
+              working-directory: ${{ github.workspace }}/lulesh
+              # Obtains the latest LULESH compatible with this version of gem5 via
+              # gem5 Resources.
+              run: build/GCN3_X86/gem5.opt util/obtain-resource.py lulesh -p lulesh
+
+            - name: Run LULUESH tests
+              working-directory: ${{ github.workspace }}
+              run: |
+                  build/GCN3_X86/gem5.opt configs/example/apu_se.py -n3 --mem-size=8GB --reg-alloc-policy=dynamic --benchmark-root="lulesh" -c \
+                  lulesh 0.01 2
```
mattsinc, line 65: Actually, now that I look at this, did you test this?  I am surprised line 65 is "lulesh 0.01 2" instead of "0.01 2" -- maybe this is something with the new YAML format I'm not getting?

abmerop: It looks right to me.  I am using the left/right view on github so the original "lulesh" (no parameters) command was moved to a new line along with adding the parameters "0.01 2".  In my opinion the other view is more confusing

abmerop (dismissed):

abmerop:
> thanks! Please let @abmerop confirm before you merge in though.

done

.github/workflows/gpu-tests.yaml
mattsinc: Weird, just must be my browser's view then.

src/arch/arm/regs/misc.cc
abmerop: I think I have seen this when merging develop into my private branch. Thanks for fixing this!

BobbyRBruce: I messed up the fix in here. I'll wait for my real fix for this in #459 to be merged to develop then fix it all.

BobbyRBruce:
FYI, when this is good to go (soon, I promise), i'll do a merge commit with this one so it'll be added to develop as one commit.

Just giving a heads up before people start to notice all the touch-up commits i've been doing.

mattsinc (commented):
Sorry I could not review this faster @BobbyRBruce , I was traveling the last ~5 days.  I think the final version looks good, thanks!

Harshil2107:
Hi @mattsinc,

The Lulesh tests are encountering failures due to an apparent issue with the run command. The current command reads as follows: `build/GCN3_X86/gem5.opt configs/example/apu_se.py -n3 --mem-size=8GB --reg-alloc-policy=dynamic --benchmark-root=lulesh -c lulesh 0.01 2`. Upon attempting to run the lulesh test locally, I received an error indicating that 0.01 and 2 are invalid arguments. Despite scanning through the code, I couldn't identify any section within apu_se.py that accepts these arguments. Could you provide guidance on this matter?

abmerop:
Hi @Harshil2107 ,  I was actually just looking at this and was wondering how this command was even working. It seems it's not.  I will post a comment on what I suspect is the issue.

.github/workflows/gpu-tests.yaml
```diff
@@ -29,6 +31,39 @@ jobs:
...
```
abmerop, line 65: I think this needs to be something like:

`build/GCN3_X86/gem5.opt configs/example/apu_se.py -n3 --mem-size=8GB --reg-alloc-policy=dynamic --benchmark-root="lulesh" -c lulesh --options="0.01 2"`

I suspect otherwise the 0.01 and 2 are being passed as positional arguments to the apu_se.py script

mattsinc: Yes @abmerop is correct.  Those inputs are arguments to to the benchmark, not to apu_se.py.

PR #257, opened 2023-09-01
misc: Add DNNMark GPU tests

mkjost0:
This adds the DNNMark tests, though they currently aren't working due to the setup of cachefiles in the workflow

Change-Id: If73fc01e1e5d993c070c0d0de7ea897ff3333238

### DIFF
--- a/.github/workflows/gpu-tests.yaml
+++ b/.github/workflows/gpu-tests.yaml
@@ -1,0 +1,110 @@
+# This workflow runs all of the very-long tests within main.py
+
+name: Weekly Tests
+
+on:
+  # Runs every Sunday from 7AM UTC
+  schedule:
+    - cron:  '00 7 * * 6'
+  # Allows us to manually start workflow for testing
+  workflow_dispatch:
+
+jobs:
+  name-artifacts:
+    runs-on: ubuntu-latest
+    outputs:
+      build-name: ${{ steps.artifact-name.outputs.name }}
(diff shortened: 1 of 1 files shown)


mattsinc (commented):
I'm going to assume if the cachefile isn't working that there is a flag somewhere in here I'm not seeing that doesn't actually run DNNMark as part of the tests then?  Else I'm not sure why we want to add them?

abmerop:
That cachefiles might be having issues because there is an extra volume needed for docker.  Namely:

`--volume ${gem5_root}/gem5-resources/src/gpu/DNNMark/cachefiles:/root/.cache/miopen/2.9.0`

At least, this is what I ran into when running manually.  I don't know how to put something like this into a YAML file though, so maybe it's already there and I'm just not seeing it.

mattsinc:
> That cachefiles might be having issues because there is an extra volume needed for docker. Namely:
> 
> `--volume ${gem5_root}/gem5-resources/src/gpu/DNNMark/cachefiles:/root/.cache/miopen/2.9.0`
> 
> At least, this is what I ran into when running manually. I don't know how to put something like this into a YAML file though, so maybe it's already there and I'm just not seeing it.

This makes sense to me.

BobbyRBruce:
Closing this as it's been abandoned. An issue has been created to track the incorporation of the GPU DNNMark test into the testing infrastructure: #1421

PR #258, opened 2023-09-01
misc: Add HACC GPU tests

mkjost0:
This adds the HACC GPU tests to be run weekly

Change-Id: I77d58ee9a3d067a749bae83826266bf89bb5020f

### DIFF
--- a/.github/workflows/gpu-tests.yaml
+++ b/.github/workflows/gpu-tests.yaml
@@ -1,0 +1,60 @@
+# This workflow runs all of the very-long tests within main.py
+
+name: Weekly Tests
+
+on:
+  # Runs every Sunday from 7AM UTC
+  schedule:
+    - cron:  '00 7 * * 6'
+  # Allows us to manually start workflow for testing
+  workflow_dispatch:
+
+jobs:
+  build-gem5:
+    runs-on: [self-hosted, linux, x64, build]
+    container: gcr.io/gem5-test/gcn-gpu:latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # Scheduled workflows run on the default branch by default. We
+          # therefore need to explicitly checkout the develop branch.
+          ref: develop
+      - name: Build gem5
+        run: scons build/GCN3_X86/gem5.opt -j $(nproc) --ignore-style
+      - uses: actions/upload-artifact@v3
+        with:
+          name: weekly-test-${{ github.run_number }}-attempt-${{ github.run_attempt }}-gem5-build-gcn3
+          path: build/GCN3_X86/gem5.opt
+          retention-days: 5
+      - run: echo "This job's status is ${{ job.status }}."
+
+  HACC-tests:
+    runs-on: [self-hosted, linux, x64, build]
+    container: gcr.io/gem5-test/gcn-gpu:latest
+    needs: build-gem5
+    timeout-minutes: 120 # 2 hours
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # Scheduled workflows run on the default branch by default. We
+          # therefore need to explicitly checkout the develop branch.
+          ref: develop
+      - uses: actions/download-artifact@v3
+        with:
+          name: weekly-test-${{ github.run_number }}-attempt-${{ github.run_attempt }}-gem5-build-gcn3
+          path: build/GCN3_X86
+      - run: chmod u+x build/GCN3_X86/gem5.opt
+      - name: make hip directory
+        run: mkdir hip
+      - name: Compile m5ops and x86
+        working-directory: ${{ github.workspace }}/util/m5
+        run: |
+          export TERM=xterm-256color
+          scons build/x86/out/m5
+      - name: Download tests
+        working-directory: ${{ github.workspace }}/hip
+        run: wget http://dist.gem5.org/dist/v22-1/test-progs/halo-finder/ForceTreeTest
+      - name: Run HACC tests
+        working-directory: ${{ github.workspace }}
+        run: |
+          build/GCN3_X86/gem5.opt configs/example/apu_se.py -n3 --reg-alloc-policy=dynamic --benchmark-root=hip -c ForceTreeTest --options="0.5 0.1 64 0.1 1 N 12 rcb"


.github/workflows/gpu-tests.yaml
```diff
@@ -0,0 +1,82 @@
...
+    strategy:
+      matrix:
+        image: [ALL, GCN3_X86]
+        # this allows us to pass additional command line parameters
+        # the default is to add -j $(nproc), but some images
+        # require more specifications when built
+        include:
+          - command-line: -j $(nproc)
+          - image: GCN3_X86
+            command-line: -j4 --ignore-style
```
mattsinc: I'm not sure if your machine only has 4 cores, but if there are more I'd recommend using more here

.github/workflows/gpu-tests.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - name: make hip directory
+        run: mkdir hip
+      - name: Compile m5ops and x86
+        working-directory: ${{ github.workspace }}/util/m5
+        run: |
+          export TERM=xterm-256color
+          scons build/x86/out/m5
+      - name: Download tests
+        working-directory: ${{ github.workspace }}/hip
+        run: wget http://dist.gem5.org/dist/v22-1/test-progs/halo-finder/ForceTreeTest
```
mattsinc, line 56: Is v22-1 still the right repo folder?

mattsinc (changes requested):
The biggest thing I feel like I'm missing is HACC had a special Dockerfile it needed that was used in the weekly script.  As far as I can tell, you aren't pointing to any docker anywhere in this file, so how are you getting the equivalent information?

.github/workflows/gpu-tests.yaml
```diff
@@ -0,0 +1,82 @@
...
+        # this allows us to pass additional command line parameters
+        # the default is to add -j $(nproc), but some images
+        # require more specifications when built
+        include:
+          - command-line: -j $(nproc)
+          - image: GCN3_X86
+            command-line: -j4 --ignore-style
+    runs-on: [self-hosted, linux, x64, build]
+    needs: name-artifacts
+    container: gcr.io/gem5-test/ubuntu-22.04_all-dependencies:latest
```
BobbyRBruce:
```suggestion
    container: gcr.io/gem5-test/gcn-gpu::latest
```

I'm pretty sure GCN3 needs built in the gcn-gpu docker image for SE mode.

.github/workflows/gpu-tests.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - uses: actions/upload-artifact@v3
+        with:
+          name: ${{ needs.name-artifacts.outputs.build-name }}${{ matrix.image }}
+          path: build/${{ matrix.image }}/gem5.opt
+          retention-days: 5
+      - run: echo "This job's status is ${{ job.status }}."
+
+  HACC-tests:
+    runs-on: [self-hosted, linux, x64, build]
+    container: mkjost/testing-gpu-image:latest
```
BobbyRBruce:
```suggestion
    container: gcr.io/gem5-test/gcn-gpu::latest
```

.github/workflows/gpu-tests.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+jobs:
+  name-artifacts:
+    runs-on: ubuntu-latest
+    outputs:
+      build-name: ${{ steps.artifact-name.outputs.name }}
+    steps:
+    - uses: actions/checkout@v2
+    - id: artifact-name
+      run: echo "name=$(date +"%Y-%m-%d_%H.%M.%S-")" >> $GITHUB_OUTPUT
```
BobbyRBruce: I don't think you need this part at all. See my comments below.

.github/workflows/gpu-tests.yaml
BobbyRBruce: No, this is our cloud bucket. We're trying to refactor these tests so we don't download directory from the repo.

.github/workflows/gpu-tests.yaml
```diff
@@ -0,0 +1,82 @@
...
+    timeout-minutes: 4320 # 3 day
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # Scheduled workflows run on the default branch by default. We
+          # therefore need to explicitly checkout the develop branch.
+          ref: develop
+      - uses: actions/download-artifact@v3
+        with:
+          name: ${{needs.name-artifacts.outputs.build-name}}GCN3_X86
```
BobbyRBruce: Why not just upload and download with:

```suggestion
          name: weekly-test-${{ github.run_number }}-attempt-${{ github.run_attempt }}-gem5-build-gcn3
```

The `github.run_number` with `github.run_attempt` are unique per run of the workflow.

.github/workflows/gpu-tests.yaml
```diff
@@ -0,0 +1,82 @@
...
+      build-name: ${{ steps.artifact-name.outputs.name }}
+    steps:
+    - uses: actions/checkout@v2
+    - id: artifact-name
+      run: echo "name=$(date +"%Y-%m-%d_%H.%M.%S-")" >> $GITHUB_OUTPUT
+
+  build-gem5:
+    strategy:
+      matrix:
+        image: [ALL, GCN3_X86]
```
BobbyRBruce: We don't need `ALL` in this case.

mkjost0:
This test uses the updated docker image and works without cloning gem5-resources, so it should be ready for review and to start being run alongside weeklies.

mattsinc:
@mkjost0 just confirming you want me and @abmerop to review this now?

mkjost0:
@mattsinc Yes, this is ready for review!

.github/workflows/gpu-tests.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - name: make hip directory
+        run: mkdir hip
+      - name: Compile m5ops and x86
+        working-directory: ${{ github.workspace }}/util/m5
+        run: |
+          export TERM=xterm-256color
+          scons build/x86/out/m5
+      - name: Download tests
+        working-directory: ${{ github.workspace }}/hip
+        run: wget http://dist.gem5.org/dist/v22-1/test-progs/halo-finder/ForceTreeTest
```
BobbyRBruce, line 56: Inspired by this I created this: https://github.com/gem5/gem5/pull/317.

If we merge #317 we can replace this `wget` with `./build/GCN3_X86/gem5.opt util/obtain-resource.py force-tree-test -o ForceTreeTest`. This means if we update ForceTreeTest we can just change the entry in our database rather than update this URL. It'll also mean we always get the right version of this resources for the version of gem5 being tested.

BobbyRBruce (dismissed):
My comment is non-blocking. Just something we may want to replace at somepoint.

.github/workflows/gpu-tests.yaml
mattsinc: Looks like this is addressed, but I'll let @BobbyRBruce resolve.

.github/workflows/gpu-tests.yaml
mattsinc: same here

.github/workflows/gpu-tests.yaml
```diff
@@ -0,0 +1,60 @@
...
+          name: weekly-test-${{ github.run_number }}-attempt-${{ github.run_attempt }}-gem5-build-gcn3
+          path: build/GCN3_X86/gem5.opt
+          retention-days: 5
+      - run: echo "This job's status is ${{ job.status }}."
+
+  HACC-tests:
+    runs-on: [self-hosted, linux, x64, build]
+    container: gcr.io/gem5-test/gcn-gpu:latest
+    needs: build-gem5
+    timeout-minutes: 4320 # 3 day
```
mattsinc: Honestly if HACC takes more than a day I would be surprised -- are you seeing it take longer?  I'm fine with leaving this if it is convenient, but it does seem like a very conservative boundary.

mattsinc:
> The biggest thing I feel like I'm missing is HACC had a special Dockerfile it needed that was used in the weekly script. As far as I can tell, you aren't pointing to any docker anywhere in this file, so how are you getting the equivalent information?

@mkjost0 as far as I can tell, this is still not in the latest version and seems like the biggest open issue.

mattsinc (commented):
In addition to the one optional comment below, I'd particularly like my comment from my last review addressed:

The biggest thing I feel like I'm missing is HACC had a special Dockerfile it needed that was used in the weekly script. As far as I can tell, you aren't pointing to any docker anywhere in this file, so how are you getting the equivalent information?

.github/workflows/gpu-tests.yaml
abmerop: I agree here, I think this test actually only takes a few minutes with the given input size, so anything longer than maybe a few hours is indicative of a hang

.github/workflows/gpu-tests.yaml
abmerop: Should we wait for #317 then?  As it is, v22-1 is already out of date, however I do not know if there is a more recent binary, so that may be the newest

.github/workflows/gpu-tests.yaml
```diff
@@ -0,0 +1,60 @@
...
+        run: |
+          export TERM=xterm-256color
+          scons build/x86/out/m5
+      - name: Download tests
+        working-directory: ${{ github.workspace }}/hip
+        run: wget http://dist.gem5.org/dist/v22-1/test-progs/halo-finder/ForceTreeTest
+      - name: Run HACC tests
+        working-directory: ${{ github.workspace }}
+        run: |
+          build/GCN3_X86/gem5.opt configs/example/apu_se.py -n3 --reg-alloc-policy=dynamic --benchmark-root=hip -c ForceTreeTest --options="0.5 0.1 64 0.1 1 N 12 rcb"
```
abmerop, line 60: Just to note, `-n3` is fine for now, but newer ROCm versions are spawning more threads, so we may need to update this in the future

(Insert shameless plug for GPUFS here)

abmerop:
> In addition to the one optional comment below, I'd particularly like my comment from my last review addressed:
> 
> The biggest thing I feel like I'm missing is HACC had a special Dockerfile it needed that was used in the weekly script. As far as I can tell, you aren't pointing to any docker anywhere in this file, so how are you getting the equivalent information?

I believe this is addressed in #290 which has been merged

mattsinc:
> > In addition to the one optional comment below, I'd particularly like my comment from my last review addressed:
> > The biggest thing I feel like I'm missing is HACC had a special Dockerfile it needed that was used in the weekly script. As far as I can tell, you aren't pointing to any docker anywhere in this file, so how are you getting the equivalent information?
> 
> I believe this is addressed in #290 which has been merged

Maybe I'm just missing how this commit is using the Dockerfile in #290 then?

.github/workflows/gpu-tests.yaml
mkjost0: Ah I can update this, this value was carried over from when we were setting the timeout for all the weekly tests to be run together and was based on a conservative estimate from our weekly.sh tests, but I can decrease it since it does finish quickly

mkjost0:
> > > In addition to the one optional comment below, I'd particularly like my comment from my last review addressed:
> > > The biggest thing I feel like I'm missing is HACC had a special Dockerfile it needed that was used in the weekly script. As far as I can tell, you aren't pointing to any docker anywhere in this file, so how are you getting the equivalent information?
> > 
> > 
> > I believe this is addressed in #290 which has been merged
> 
> Maybe I'm just missing how this commit is using the Dockerfile in #290 then?



> > > In addition to the one optional comment below, I'd particularly like my comment from my last review addressed:
> > > The biggest thing I feel like I'm missing is HACC had a special Dockerfile it needed that was used in the weekly script. As far as I can tell, you aren't pointing to any docker anywhere in this file, so how are you getting the equivalent information?
> > 
> > 
> > I believe this is addressed in #290 which has been merged
> 
> Maybe I'm just missing how this commit is using the Dockerfile in #290 then?

After we pushed #290, we updated the gcn gpu Docker image to include all of the environmental variables that are set for HACC. I've tested these changes locally to make sure using this docker image instead of the special one for HACC doesn't break anything, and the tests are able to run successfully.

mattsinc:
> > > > In addition to the one optional comment below, I'd particularly like my comment from my last review addressed:
> > > > The biggest thing I feel like I'm missing is HACC had a special Dockerfile it needed that was used in the weekly script. As far as I can tell, you aren't pointing to any docker anywhere in this file, so how are you getting the equivalent information?
> > > 
> > > 
> > > I believe this is addressed in #290 which has been merged
> > 
> > 
> > Maybe I'm just missing how this commit is using the Dockerfile in #290 then?
> 
> > > > In addition to the one optional comment below, I'd particularly like my comment from my last review addressed:
> > > > The biggest thing I feel like I'm missing is HACC had a special Dockerfile it needed that was used in the weekly script. As far as I can tell, you aren't pointing to any docker anywhere in this file, so how are you getting the equivalent information?
> > > 
> > > 
> > > I believe this is addressed in #290 which has been merged
> > 
> > 
> > Maybe I'm just missing how this commit is using the Dockerfile in #290 then?
> 
> After we pushed #290, we updated the gcn gpu Docker image to include all of the environmental variables that are set for HACC. I've tested these changes locally to make sure using this docker image instead of the special one for HACC doesn't break anything, and the tests are able to run successfully.

I think I understand, thanks.  Please make the timeout change and then I'll approve (although @abmerop may want the other change mentioned above)

mattsinc (approved):
This is fine on my end, but @abmerop should also chime in about if he wants his requested change included here.

Thanks @mkjost0 !

abmerop (approved):
This all looks good to me.  My comment about the number of cpus was more of a reminder, so I'm not requesting any changes right now

.github/workflows/gpu-tests.yaml
BobbyRBruce: @mkjost0 : This has been added as a resource with the ID `hacc-force-tree`. #317 has been merged so please make this change.

BobbyRBruce (changes requested):

BobbyRBruce (approved):

PR #263, opened 2023-09-04
misc,ext,tests: Automatically split CI TestLib tests across GitHub Action jobs 

BobbyRBruce:
This PR utilizes GitHub Action's matrix's to automatically distribute the CI testlib gem5 build and test jobs across available GitHub Action Runners.

The CI tests (the `quick` testlib tests, i.e. those run with `./main.py run`) are distributed across the runners on a per directory basis --- all directories under "tests/gem5" are run as their own jobs. 

The necessary gem5 builds for each workflow are now automatically inferred via the introduction of `./main.py list`'s  `--build-targets` flag which returns the gem5 build target for a given test or collection of tests. E.g., `./main.py list --build-targets` will return the build targets for all the `quick` testlib tests and `./main.py list --build-target --uid=<id>` will return the build targets the test suite `<id>` requires.

Moving from monolithic jobs to fine-grained ones will make the locaiton of test failures more obvious. Each job has it's own artifact containing "test/testing-results" for the tests run in that job. In addition, maintenance of these files should become less burdensome due to less hardcoding.

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -46,27 +46,6 @@ jobs:
             "curl -Lo $f https://gerrit-review.googlesource.com/tools/hooks/commit-msg ; chmod +x $f\n Then amend the commit with git commit --amend --no-edit, and update your pull request."
           exit 1
 
-  build-gem5:
-    runs-on: [self-hosted, linux, x64, build]
-    if: github.event.pull_request.draft == false
-    container: ghcr.io/gem5/ubuntu-22.04_all-dependencies:latest
-    needs: [pre-commit, check-for-change-id] # only runs if pre-commit and change-id passes
-    outputs:
-      artifactname: ${{ steps.name.outputs.test }}
-    steps:
-      - uses: actions/checkout@v3
-      - id: name
-        run: echo "test=$(date +"%Y-%m-%d_%H.%M.%S")-artifact" >> $GITHUB_OUTPUT
-
-      - name: Build gem5
-        run: |
-          scons build/ALL/gem5.opt -j $(nproc)
-      - uses: actions/upload-artifact@v3
-        with:
-          name: ${{ steps.name.outputs.test }}
-          path: build/ALL/gem5.opt
-      - run: echo "This job's status is ${{ job.status }}."
-
   unittests-all-opt:
     runs-on: [self-hosted, linux, x64, run]
     if: github.event.pull_request.draft == false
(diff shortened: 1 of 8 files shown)


.github/workflows/daily-tests.yaml
```diff
@@ -71,141 +62,85 @@ jobs:
...
+        - name: Download gem5 binaries
+          uses: actions/download-artifact@v2
+          with:
+            name: "run-${{ github.run_number }}_attempt-${{ github.run_attempt }}_gem5-builds"
+
+          # GitHub's artifact action does not pereserve file permissions.
+          # Therefore, we need to make the gem5 binaries executable.
+        - name: Chmod gem5.{opt,debug,fast} to be executable
+          run:
+            find . -name "gem5.opt" | xargs chmod u+x
```
powerjg: nit, personally I'd use the following, but `xargs` works, so no need to change.

```suggestion
            find . -name "gem5.opt" -exec chmod u+x {} \;
```

.github/workflows/daily-tests.yaml
```diff
@@ -71,141 +62,85 @@ jobs:
...
+          ref: develop
+
+        # Get all the "long" Suites of tests. Used in
+      - name: Get Suite IDs for Long
+        id: suite-ids
+        run: |
+           suite_ids=$(./main.py list --length=long | grep "^SuiteUID:")
+           {
+               echo 'SUITE_IDS<<EOF'
+               ./main.py list --length=long | grep "^SuiteUID:"
```
powerjg: Would it be possible to specify the binary/variant that is needed in the runner step if we extended `main.py` to be able to output the right information? 

More generally, we might as well modify this script, or make a new one, that gives us the information we need here instead of using grep.

powerjg (commented):
This is *really cool*. Thanks for putting this together! One small nit and a suggestion below.

.github/workflows/daily-tests.yaml
BobbyRBruce: Should be in the latest push.

.github/workflows/daily-tests.yaml
BobbyRBruce: My latest push includes this. It's not a perfect solution but I think it works.

.github/workflows/ci-tests.yaml
```diff
@@ -80,39 +59,104 @@ jobs:
...
+    container: gcr.io/gem5-test/ubuntu-22.04_all-dependencies:latest
+    needs: [pre-commit, check-for-change-id, testlib-quick-matrix]
+    strategy:
+      matrix:
+        build-target: ${{ fromJson(needs.testlib-quick-matrix.outputs.build-matrix) }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build gem5
+        run: scons ${{ matrix.build-target }} -j $(nproc)
+      - run: echo "temp" > anchor.txt
```
powerjg: What is this file used for?

.github/workflows/ci-tests.yaml
```diff
@@ -80,39 +59,104 @@ jobs:
...
       - uses: actions/checkout@v3
       - uses: actions/download-artifact@v3
         with:
-          name: ${{needs.build-gem5.outputs.artifactname}}
-          path: build/ALL
-      - run: chmod u+x build/ALL/gem5.opt
-      - name: The TestLib CI Tests
+          name: ci-tests-${{ github.run_number }}-attempt-${{ github.run_attempt }}-testlib-quick-all-gem5-builds
+      - run: ls -l
+      - run: find . -name "gem5.*"
```
powerjg: For debugging?

.github/workflows/ci-tests.yaml
```diff
@@ -80,39 +59,104 @@ jobs:
...
+        # Check that the gem5.opt artifact exists and is executable.
+      - name: Chmod gem5.{opt,debug,fast} to be executable
+        run: |
+          find . -name "gem5.opt" -exec chmod u+x {} \;
+          find . -name "gem5.debug" -exec chmod u+x {} \;
+          find . -name "gem5.fast" -exec chmod u+x {} \;
+
+        # Run the testlib quick tests in the given directory.
+      - name: Run "tests/${{ matrix.test-dir }}"" TestLib quick tests
```
powerjg: Is there an extra `"` here?

.github/workflows/ci-tests.yaml
BobbyRBruce: yes. Not ready for review just yet

powerjg:
BTW, this is really cool, still! Is there anything you need from me to make progress?

BobbyRBruce:
@powerjg and @mkjost0 : this is **finally** ready for review... far too much trial and error.

You can see the latest test passing as evidence of it running well: https://github.com/gem5/gem5/actions/runs/6178661302. You can download the artifacts for any directory under "tests/gem5".

At a quick glance it seems like this is roughly 40 minutes faster. the asmtests are the weakest point, they take an hour. If we really wanted to do better we could split these up but it's good for now.

tests/gem5/fixture.py
```diff
@@ -74,7 +76,7 @@ def post_test_procedure(self, testitem):
         suiteUID = testitem.metadata.uid.suite
         testUID = testitem.metadata.name
         testing_result_folder = os.path.join(
-            config.result_path, "SuiteUID:" + suiteUID, "TestUID:" + testUID
+            config.result_path, "SuiteUID-" + suiteUID, "TestUID-" + testUID
```
powerjg, line 79: This might break how people run tests in batches. We should make sure that @giactra and @abmerop are aware and don't have any big issues.

powerjg (commented):
lgtm. I think this is a *massive* improvement. Thanks, Bobby!

tests/gem5/fixture.py
BobbyRBruce: Just to clarify, this is only changing the name of the directories in "tests/testing-results". For some reason when I created this code I used a colon. This isn't allowed in file/directory names in some OS's (MS Windows at least), and needs updated here as the `action/artifact-upload` and `action/artifact-download` actions don't allow them in file/directory names.

What has not changed is the format of the SuiteIDs (i.e., that you pass to `--uid` when running a specific job). This is just a file name fix.

powerjg (dismissed):

powerjg (approved):

PR #264, opened 2023-09-04
misc: Split testlib CI Tests to one job per dir

BobbyRBruce:
This splits the CI Tests to one job per sub-directory in "tests/gem5" via a matrix.

Advantages:
* We can utilize more runners to run the quick tests. This should mean tests run quicker.
* This approach does not require editing of the workflow as more tests are added or taken away.
* There is now an output artifact for each directory in "tests/gem5" instead of one for the entriety of every quick test in "tests".

In addition:
* The artifact retention for the test outputs has been increased to 30 days.
* The output test artifacts have been renamed to be more descriptive of the job, run, attempt, directory run, and the status.
* The 'tar' step has been removed. GitHub's 'action/artifact' can handle directories.


BobbyRBruce:
This PR was merged in #263.

Closing.

PR #270, opened 2023-09-05
misc: Fix buggy special path comparisons

### LINKED ISSUE #269: Incorrect special file comparisons in Linux kernel
**Describe the bug**
The path string comparisons in src/kern/linux/linux.cc function `Linux::openSpecialFile()` do not check for string equality but rather equality of prefixes.
This causes `"/proc/self/maps0123456789"` to compare "equal" to `"/proc/self/maps"`, so a program accessing that invalid file path will actually access `"/proc/self/maps"`.

**Affects version**
develop @ 48a40cf

**gem5 Modifications**
None

**To Reproduce**
1. Compile the following program `bug.c` with the command `cc -static -o bug bug.c`:
```c
#include <stdio.h>
#include <stdlib.h>

int main() {
  FILE *f = fopen("/proc/self/maps0123456789", "r");
  if (f == NULL) {
    // correct path
    perror("fopen");
    return EXIT_FAILURE;
  } else {
    // actual path
    char line[256];
    while (fgets(line, sizeof line, f))
      printf("%s", line);
  }
}
```
1. Compile gem5 with command `scons build/X86/gem5.opt`
2. Execute the simulation with `./build/X86/gem5.opt configs/deprecated/example/se.py -c ./bug`
3. Observe it happily print out the contents of "/proc/self/maps", rather than printing an error.

**Expected behavior**
The above program should print "fopen: No such file or directory", not print out the contents of a different file, "/proc/self/maps".

**Host Operating System**
Ubuntu 20.04

**Host ISA**
X86

**Compiler used**
gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0

nmosier:
This patch fixes the buggy special path comparisons in src/kern/linux/linux.cc Linux::openSpecialFile(), which only checked for equality of path prefixes, but not equality of the paths themselves. This patch replaces those buggy comparisons with regular std::string::operator== string equality comparisons.

GitHub issue: https://github.com/gem5/gem5/issues/269

Change-Id: I216ff8019b9a6a3e87e364c2e197d9b991959ec1

### DIFF
--- a/src/kern/linux/linux.cc
+++ b/src/kern/linux/linux.cc
@@ -58,19 +58,19 @@ Linux::openSpecialFile(std::string path, Process *process,
     bool matched = false;
     std::string data;
 
-    if (path.compare(0, 13, "/proc/meminfo") == 0) {
+    if (path == "/proc/meminfo") {
         data = Linux::procMeminfo(process, tc);
         matched = true;
-    } else if (path.compare(0, 11, "/etc/passwd") == 0) {
+    } else if (path == "/etc/passwd") {
         data = Linux::etcPasswd(process, tc);
         matched = true;
-    } else if (path.compare(0, 15, "/proc/self/maps") == 0) {
+    } else if (path == "/proc/self/maps") {
         data = Linux::procSelfMaps(process, tc);
         matched = true;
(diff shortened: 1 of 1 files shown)


powerjg (approved):
Thanks for the contribution! Looks good to me.

PR #272, opened 2023-09-05
misc,tests: Add "commits-since-last-run" to Daily tests

BobbyRBruce:
This commits since the last time the daily test was run. This is helpful to more quickly lcoate the cause of build failures.

### DIFF
--- a/.github/workflows/daily-tests.yaml
+++ b/.github/workflows/daily-tests.yaml
@@ -8,6 +8,44 @@ on:
     - cron:  '0 7 * * *'
 
 jobs:
+  commits-since-last-run:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


mkjost0 (approved):

mkjost0:
The current issue with this change is that artifacts can only be accessed per workflow, not across multiple workflows.  Every time you run this, it is unable to find the last artifact, which means we can't find the difference in commit history.  The fix to this would be to use an action that allows you to download artifacts from past workflows instead of the default download artifact action.

BobbyRBruce:
Marked this as "low-priority". It needs tested but I don't have the bandwidth to do that right now. Closing for a later date.

PR #274, opened 2023-09-06
mem-cache: Fix bug in classic cache while clflush

hnpl:
This change, https://github.com/gem5/gem5/pull/205, mistakenly allocates write buffer for clflush instruction when there's a cache miss. However, clflush in gem5 is not a write instruction. Thus, the cache should allocate miss buffer in this case.

Change-Id: I9c1c9b841159c4420567e9c929e71e4aa27d5c28

### DIFF
--- a/src/mem/cache/cache.cc
+++ b/src/mem/cache/cache.cc
@@ -338,22 +338,27 @@ Cache::handleTimingReqMiss(PacketPtr pkt, CacheBlk *blk, Tick forward_time,
 
         if (pkt->isWrite()) {
             allocateWriteBuffer(pkt, forward_time);
-        } else if (pkt->isRead()) {
+        } else {
             // uncacheable accesses always allocate a new MSHR
 
             // Here we are using forward_time, modelling the latency of
             // a miss (outbound) just as forwardLatency, neglecting the
             // lookupLatency component.
+
+            // Here we allow allocating miss buffer for read requests
+            // and x86's clflush requests. A clflush request should be
+            // propagate through all levels of the cache system.
+
+            // Doing clflush in uncacheable regions might sound contradictory;
+            // however, it is entirely possible due to how the Linux kernel
+            // handle page property changes. When a linux kernel wants to
+            // change a page property, it flushes the related cache lines. The
+            // kernel might change the page property before flushing the cache
+            // lines. This results in the clflush might occur in an uncacheable
+            // region, where the kernel marks a region uncacheable before
+            // flushing. clflush results in a CleanInvalidReq.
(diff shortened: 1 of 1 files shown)


hnpl:
Tested with setting a page uncacheable in full system mode.

The change in #205 results in an assertion error when calling allocatesWriteBuffer.

#205 + #274 works.

BobbyRBruce (dismissed):

powerjg:
Before we merge... How do we know this doesn't break arm cache management instructions?

powerjg (changes requested):
Blocking until we are sure this doesn't affect Arm instructions

giactra:
Being summoned by the Arm mention =P

Unless I am missing something, it seems to me this commit is sort of reverting https://github.com/gem5/gem5/pull/205
Before #205, the code was:

        if (pkt->isWrite()) {
            allocateWriteBuffer(pkt, forward_time);
        } else {
            allocateMissBuffer(pkt, forward_time);
        }

With this patch we go back to:

        if (pkt->isWrite()) {
            allocateWriteBuffer(pkt, forward_time);
        } else if (pkt->isRead()) {
            allocateMissBuffer(pkt, forward_time);
        } else {
            assert(pkt->isCleanInvalidateRequest());
            allocateMissBuffer(pkt, forward_time);
        }

Which is essentially the same other than the extra assert. IMHO this extra clause is not really worth it. Shall we just revert #205 ?

hnpl:
I agree that the if/else statements here are rather silly.

However, the problem before #205 was that, in the else statement, the code asserts that the packet is a read. Essentially, within an uncacheable region, classic cache only serves either read or write requests.

```
        if (pkt->isWrite()) {
            allocateWriteBuffer(pkt, forward_time);
        } else {
            assert(pkt->isRead());
            allocateMissBuffer(pkt, forward_time);
        }
```

#205 and #274 basically loosen this restriction to serving read/write/clflush on uncacheable regions.

The comments in #205 show that clflush can happen on uncacheable regions.

giactra (changes requested):
Yeah apologies I missed that assert.
In that case I believe the code should actually be:

        if (pkt->isWrite()) {
            allocateWriteBuffer(pkt, forward_time);
        } else {
            assert(pkt->isRead() || pkt->isCleanInvalidateRequest());
            allocateMissBuffer(pkt, forward_time);
        }

hnpl:
I reworded the code comments and change the assert as suggested.

giactra (approved):
Thanks for the changes!

BobbyRBruce (approved):

powerjg (approved):

PR #279, opened 2023-09-07
sim: add bypass_on_change to the set() of a signal

johnnyko-google:
When reset a port, we don't want to trigger a onChange(). Offer an option to bypass it and update state only.

Change-Id: Ia53b7a76d2a320ea67101096cdbfe2eafaf440d2


giactra (approved):
LGTM

PR #288, opened 2023-09-08
mem-ruby: start using txnid and DBID identifiers in CHI transactions

giactra:
With this PR our CHI implementation starts making use of the txnid and DBID identifiers.
Note: we were already making use of the txnId for DVM messages to convey the DVM address. This is still the case.
In the future we should realign the DVM logic so that the txnId is solely used as a transaction identifier.

### DIFF
--- a/src/mem/slicc/symbols/StateMachine.py
+++ b/src/mem/slicc/symbols/StateMachine.py
@@ -64,6 +64,7 @@ python_class_map = {
     "DMASequencer": "DMASequencer",
     "RubyPrefetcher": "RubyPrefetcher",
     "Cycles": "Cycles",
+    "Addr": "Addr",
 }
 
 
--- a/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
+++ b/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
@@ -63,6 +63,7 @@ action(AllocateTBE_Request, desc="") {
         out_msg.usesTxnId := false;
         out_msg.event := Event:SendRetryAck;
         out_msg.retryDest := in_msg.requestor;
+        out_msg.txnId := in_msg.txnId;
         retryQueue.emplace(in_msg.addr,false,in_msg.requestor);
       }
     }
@@ -145,6 +146,7 @@ action(AllocateTBE_SeqRequest, desc="") {
       assert(in_msg.Prefetch == PrefetchBit:No);
       out_msg.is_local_pf := false;
       out_msg.is_remote_pf := false;
+      out_msg.txnId := max_outstanding_transactions;
 
       if ((in_msg.Type == RubyRequestType:LD) ||
           (in_msg.Type == RubyRequestType:IFETCH)) {
@@ -2508,6 +2510,7 @@ action(Send_Data, desc="") {
   enqueue(datOutPort, CHIDataMsg, data_latency) {
     out_msg.addr := tbe.addr;
     out_msg.type := tbe.snd_msgType;
+    out_msg.txnId := tbe.txnId;
 
     int offset := tbe.snd_pendBytes.firstBitSet(true);
     assert(offset < blockSize);
@@ -2564,6 +2567,7 @@ action(Send_CompUC, desc="") {
     out_msg.type := CHIResponseType:Comp_UC;
     out_msg.responder := machineID;
     out_msg.Destination.add(tbe.requestor);
+    out_msg.txnId := tbe.txnId;
   }
 }
 
@@ -2574,6 +2578,7 @@ action(Send_CompUC_Stale, desc="") {
     out_msg.type := CHIResponseType:Comp_UC;
     out_msg.responder := machineID;
     out_msg.Destination.add(tbe.requestor);
+    out_msg.txnId := tbe.txnId;
     // We don't know if this is a stale clean unique or a bug, so flag the
     // reponse so the requestor can make further checks
     out_msg.stale := true;
@@ -2587,6 +2592,7 @@ action(Send_CompAck, desc="") {
     out_msg.type := CHIResponseType:CompAck;
     out_msg.responder := machineID;
     out_msg.Destination.add(mapAddressToDownstreamMachine(tbe.addr));
+    out_msg.txnId := tbe.txnId;
   }
 }
 
@@ -2610,6 +2616,7 @@ action(Send_CompDBIDResp, desc="") {
     out_msg.type := CHIResponseType:CompDBIDResp;
     out_msg.responder := machineID;
     out_msg.Destination.add(tbe.requestor);
+    out_msg.txnId := tbe.txnId;
   }
 }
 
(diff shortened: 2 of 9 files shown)


src/mem/slicc/symbols/StateMachine.py
```diff
@@ -64,6 +64,7 @@
     "DMASequencer": "DMASequencer",
     "RubyPrefetcher": "RubyPrefetcher",
     "Cycles": "Cycles",
+    "Addr": "Addr",
```
powerjg, line 67: Could you explain how this doesn't break things since `Addr` is already a type that SLICC understands? Or, could you explain why this change is needed?

What worries me is that there are a number of places with code like `self.symtab.find("Addr", Type)`. Does specifying this in the `python_class_map` not have any effect on how SLICC understands this type?

powerjg (commented):
I reviewed the first 3 changes which affect all of Ruby/SLICC. I haven't reviewed the CHI changes.

> Rather than pushing the burden on us, we should always
allow a developer to use a static_cast in slicc and let the C++ compiler
complain if the generated code is not compatible

I disagree with this sentiment. Errors in the C++ code that is generated by SLICC are very confusing and difficult to track down. If at all possible, we should have errors in *SLICC* not in the generated C++ code.

A better solution to this problem would be to introduce a new keyword in SLICC (e.g., `integer_cast` that would only be used for compatible integer types (e.g., as defined in `InfixOperatorExprAST`).

An alternative which would keep some error checking in SLICC, but not require the headache of adding a new operator would be to add a check here for integer type compatibility.

I'm not going to push super hard for this because SLICC is already awful and this change doesn't make it *that* much worse. I'm happy to hear pushback :smile:

giactra:
Thanks for reviewing this PR Jason

> I reviewed the first 3 changes which affect all of Ruby/SLICC. I haven't reviewed the CHI changes.
> 
> > Rather than pushing the burden on us, we should always
> > allow a developer to use a static_cast in slicc and let the C++ compiler
> > complain if the generated code is not compatible
> 
> I disagree with this sentiment. Errors in the C++ code that is generated by SLICC are very confusing and difficult to track down. If at all possible, we should have errors in _SLICC_ not in the generated C++ code.

I haven't tweaked ruby extensively so I am not super familiar with SLICC error reporting. IMHO re-implementing a type checking system in a DSL which is directly mapped to C++ feels a bit like re-inventing the wheel. Especially as the motivation for keeping this is SLICC is an inherent flaw in the way the error gets reported and not theoretical reasons (like "we want to be type safer than C++").

> 
> A better solution to this problem would be to introduce a new keyword in SLICC (e.g., `integer_cast` that would only be used for compatible integer types (e.g., as defined in `InfixOperatorExprAST`).
> 
> An alternative which would keep some error checking in SLICC, but not require the headache of adding a new operator would be to add a check here for integer type compatibility.

I will try this alternative you suggested

> 
> I'm not going to push super hard for this because SLICC is already awful and this change doesn't make it _that_ much worse. I'm happy to hear pushback 😄

The real issue is that it took me 1 day to write a simple function which was taking an address, incrementing it by one (an integer) and then return the new value...

src/mem/slicc/symbols/StateMachine.py
giactra: > Could you explain how this doesn't break things since `Addr` is already a type that SLICC understands? Or, could you explain why this change is needed?
>

As the commit suggests this is just needed to be able to use the Addr type as a python Param for the CacheController. Even if SLICC recognises Addr as a type, you can't currently use it as a parameter. See below for further info
 
> What worries me is that there are a number of places with code like `self.symtab.find("Addr", Type)`. Does specifying this in the `python_class_map` not have any effect on how SLICC understands this type?

That was my understanding. The python_class_map is only used for type checking the Param type for the CacheController

```
 270         code(
 271             """
 272 from m5.params import *
 273 from m5.SimObject import SimObject
 274 from m5.objects.Controller import RubyController
 275  
 276 class $py_ident(RubyController):
 277     type = '$py_ident'
 278     cxx_header = 'mem/ruby/protocol/${c_ident}.hh'
 279     cxx_class = 'gem5::ruby::$py_ident'
 280 """
 281         )
 282         code.indent()
 283         for param in self.config_parameters:  
 284             dflt_str = ""
 285  
 286             if param.rvalue is not None:      
 287                 dflt_str = str(param.rvalue.inline()) + ", "
 288                                                                                                                                                                                                                                                                               
 289             if param.type_ast.type.c_ident in python_class_map:
 290                 python_type = python_class_map[param.type_ast.type.c_ident]
 291                 code(
 292                     '${{param.ident}} = Param.${{python_type}}(${dflt_str}"")'
 293                 )
 294  
 295             else:
 296                 self.error(
 297                     "Unknown c++ to python class conversion for c++ "
 298                     "type: '%s'. Please update the python_class_map "
 299                     "in StateMachine.py",     
 300                     param.type_ast.type.c_ident,
 301                 )

```

giactra:
Hi Jason, I played a bit with my code and I have to push back a bit on delegating the type checking logic to the C++ compiler.

I tried to introduce a casting error in my code and the error log was not obscure at all!
```

CHI-dvm-misc-node-actions.sm:217: Warning: Non-void return ignored, return type is 'Tick'
CHI-dvm-misc-node-actions.sm:221: Warning: Non-void return ignored, return type is 'Tick'
 [VER TAGS]  -> ARM/sim/tags.cc
 [     CXX] ARM/mem/ruby/protocol/MiscNode_Controller.py.cc -> .o
 [     CXX] ARM/mem/ruby/protocol/MiscNode_Controller.py.cc -> .pyo
 [     CXX] ARM/mem/ruby/protocol/Memory_Controller.py.cc -> .pyo
 [     CXX] ARM/mem/ruby/protocol/Memory_Controller.py.cc -> .o
 [     CXX] ARM/mem/ruby/protocol/Cache_Controller.cc -> .o
build/ARM/mem/ruby/protocol/Cache_Controller.cc: In member function ‘gem5::ruby::Cache_TBE* gem5::ruby::Cache_Controller::allocateRequestTBE(const Addr&, const gem5::ruby::CHIRequestMsg&)’:
build/ARM/mem/ruby/protocol/Cache_Controller.cc:9690:26: error: invalid ‘static_cast’ from type ‘gem5::ruby::Cache_TBE*’ to type ‘gem5::Addr’ {aka ‘long unsigned int’}
 9690 |         (*tbe).m_txnId = static_cast<Addr >(tbe)
      |                          ^~~~~~~~~~~~~~~~~~~~~~~
scons: *** [build/ARM/mem/ruby/protocol/Cache_Controller.o] Error 1
```

The only disadvantage of it is that it does not reference the protocol file/line directly so a developer would have to infer it from the name of the function/method. It doesn't really seem to me a big deal. Considering that you are happy to allow casting between integer types, the chance of someone making mistakes when casting the few custom types usable in SLICC is really minimal... Please let me know your thoughts :)

powerjg (commented):
This looks good to me. Thanks for the updates.

We should update the documentation on Ruby here: https://www.gem5.org/documentation/general_docs/ruby/slicc/#other-compiler-features

@Harshil2107  will you please add a bullet point about using "value" for `static_cast`?

powerjg (approved):

PR #291, opened 2023-09-08
misc,tests: Fix GPU Daily Tests

BobbyRBruce:
Fixes the following error:
https://github.com/gem5/gem5/actions/runs/6106624129/job/16572039810

In GitHub Actions certain actions require Docker. You can therefore not run them within a container. The 'gpu-tests' job runs in the 'gcr.io/gem5-test/gcn-gpu:latest' container. 'wei/wget' depends on docker and therefore fails.

This is fixed by separating the running of `wei/wget` to a different jobs, with the test-progs uploaded as artifacts for use in 'gpu-tests'.


powerjg (approved):
lgtm. We should really make square a `Resource`... maybe a task for someone else, though

mattsinc (approved):

BobbyRBruce:
I was incorrect. This wasn't the source of the problem. Closing.

PR #295, opened 2023-09-09
configs: Add missing gpufs option

abmerop:
Change-Id: Ia9368ea693171ed0cde26042454f4132c4401d65

### DIFF
--- a/configs/example/gpufs/runfs.py
+++ b/configs/example/gpufs/runfs.py
@@ -137,6 +137,13 @@ def addRunFSOptions(parser):
         "MI200 (gfx90a)",
     )
 
+    parser.add_argument(
+        "--exit-at-gpu-kernel",
+        type=int,
+        default=-1,
+        help="Exit simulation after running this many kernels",
+    )
+
 
 def runGpuFSSystem(args):
     """


abmerop:
Have been getting many emails asking why copying and pasting the instructions to run gpufs does not work on stable, this fixes it.

mattsinc (approved):

abmerop:
Please leave this open for Bobby or Jason to check.  I don't know how the whole hotfix versioning things need to be accounted for.

BobbyRBruce:
This is just adding something to a config script example, what's the justification for this being a hotfix? 

`configs/example` isn't an interface for gem5. We really need to move the GPU code out of here and into the standard library if we're dependent on these scripts to run GPU simulations.

abmerop:
> This is just adding something to a config script example, what's the justification for this being a hotfix?
> 
> `configs/example` isn't an interface for gem5. We really need to move the GPU code out of here and into the standard library if we're dependent on these scripts to run GPU simulations.

1) Check out stable
2) Follow the instructions to run GPUFS
3) See "config option doesn't exist"

Until there is a way for GPUFS to use stdlib, I don't think we really have a option not to fix these

powerjg:
I could be mistaken, but it sounds like the problem is a documentation improvement, not a bug which rises to level of requiring a hotfix (which is a pretty large overhead).

I would suggest updating the documentation to say "use `--exit-at-gpu-kernel` only on gem5's develop branch or after 23.1 is released. `--exit-at-gpu-kernel` does not work with gem5 23.0.`

If this was such a high priority issue, I would have expected that it would have been raised less than 3 months after a release.

I'm going to close this change because as it is it cannot be merged as it doesn't follow our hotfix guidelines: https://github.com/gem5/gem5/blob/stable/CONTRIBUTING.md#hotfixes

If you want to continue discussing this issue, I suggest opening an issue and we can discuss there.

abmerop:
> I could be mistaken, but it sounds like the problem is a documentation improvement, not a bug which rises to level of requiring a hotfix (which is a pretty large overhead).
> 
> I would suggest updating the documentation to say "use `--exit-at-gpu-kernel` only on gem5's develop branch or after 23.1 is released. `--exit-at-gpu-kernel` does not work with gem5 23.0.`
> 
> If this was such a high priority issue, I would have expected that it would have been raised less than 3 months after a release.


This will not work.  It will still complain about the missing option.  The issue is the patches went out while the v23.0 branch was being created and this change was missed. I can point you to the gerrit relation chain to show you if you want.

Also there have been people complaining but they email me directly.  There are not that many users for GPUFS right now, sure.  But we finally have it working and it's disappointing you seem to not care that people cannot follow the documentation we have and get things to work on stable.

abmerop:
For the folks I have linked this PR to in order to get stable to work, you can add the change as is.  This change:

- Does not require retesting of stable whatsoever
- It does not require changing the documentation. The documentation is already correct minus this minor fix
- This is fixed on develop will be released whenever the next release is

BobbyRBruce:
@mattsinc or @abmerop  : Can you point me to towards this change on develop? I'll create a hotfix over the next day for stable.

powerjg:
@BobbyRBruce (sorry, this is in the wrong place, but I don't want to forget)... We may also need to hotfix the URL for mongodb. Let's talk about that before we create this hotfix.

abmerop:
Yes, it is here:  https://gem5-review.googlesource.com/c/public/gem5/+/71418

And this is the staging patch which uses it: https://gem5-review.googlesource.com/c/public/gem5/+/72138  .  I can't recall why these were not in the same commit, but that was a mistake on my part.

This was submitted during the 2(?)-ish week staging period for v23, so I think it was just something we missed:
image

I don't think you all need to really do much here... Unless I am missing something this would just be a commit.  It's not a 'true hotfix' more like updating a README file

PR #300, opened 2023-09-11
misc,config,tests: Add stats output compare to compiler tests

BobbyRBruce:


BobbyRBruce:
@mkjost0 : can you test this out to make sure it works? I can't test it locally.

BobbyRBruce:
This needs checked to see if it works before incorporation. It's not a huge improvement over our current tests. Labeling as low-priority and closing for now.

PR #305, opened 2023-09-12
util: Added cli tool for gem5 resources

Harshil2107:
- create_resource is functional but need polishing

- Validate and Update resource are work in progress.

Change-Id: I6ec65ccef72257dd9ea6e5e83019306834be7367


Harshil2107:
Note: Cli will not work without the URI as an env variable

BobbyRBruce:
I think the direction you were heading in your old design was too haphazrd and was going to make it hard for you to maintain, expand, test, and improve going forward. 

I pushed some patches which I hope will  show a slightly better approach. Nothing I've submitted is complete, or even what you may need to fully complete the task, but it just serves as an example/tutorial of a better way to do things.

I'd advise going through each commit one-by-one to see the development.

As a very high-level note I'll say that you should think about creating a design with is easy to extend without modification. If i want to create a new datasource I should be able to create a new class which implements an API, then just plug it in. If modification is necessary then the modification should only affect that one thing, in one place. If the schema changes I want to have just one place I change it in this tool, nothing more.

Harshil2107:
Thank you for the feedback. I will look into the suggested changes.

BobbyRBruce:
I'm not sure exactly what you were doing but I fixed the bugs Iw as running into with this workflow:

```sh
pip install -e . # Installs the tool.
gem5-resources-manager # Run the tool.
```

util/gem5-resources-manager-v2/resource_demo.json
BobbyRBruce, line 1: Format this file with new-lines and intends.

util/gem5-resources-manager-v2/src/gem5_resources_manager/creator/creator.py
```diff
@@ -0,0 +1,154 @@
...
+            "--ignore-db-check",
+            "-d",
+            action="store_true",
+            help="ignore the check that the entered resource does not exist in the database.",
+        )
+        parser.add_argument(
+            "--handle-url",
+            "-u",
+            action="store_true",
+            help="automatically fill in the is_zipped, is_tarred, and md5sum fields if the url field has a link to the source repository.",
```
BobbyRBruce, line 49: Remember to break up your strings if they get too long.

util/gem5-resources-manager-v2/test-scritp.py
BobbyRBruce, line 1: I really really really REAALY want tests for this, but can you write some Pyunit tests instead?

This is not fun to read and non-standard.

util/gem5-resources-manager-v2/src/gem5_resources_manager/updater/updater.py
```diff
@@ -0,0 +1,126 @@
...
+            for field in optional_fields.keys():
+                is_update = input(f"Do you want to add the {field} field? (Y/N): ")
+                if is_update.lower() == "y":
+                    print("Current value:", optional_fields[field])
+                    new_value = input("Enter new value: ")
+                    updated_resource[field] = new_value
+
+        print("Updated resource:", json.dumps(updated_resource, indent=4))
+
+    def get_dependent_resources(self, resource, dependent_resources):
```
BobbyRBruce: There's a loy of functions in this PR that don't have typing information. This include return type info.

Pleaes try to get into the habit of adding these. In this case it's hard to know what `dependent_resources` is supposed to be. A set of resources, i'd guess?

util/gem5-resources-manager-v2/src/gem5_resources_manager/resource_cli.py
```diff
@@ -0,0 +1,30 @@
...
+        subtool.get_arg_parser(subparser)
+
+    args = parser.parse_args()
+
+    datasource = AtlasDataSource(
+        mongo_uri="", database_name="gem5-vision", collection_name="versions_test"
+    )
+
+    datasource.open()
+    subtool.execute(args=args, data_source=datasource)
```
BobbyRBruce, line 29: This is cause of your bug. At least in part. What were you trying to do here? `subtool` is just the variable that holds the value in the for iteration. Once that's over it's just going to be the last thing set.

You've changed this code a lot, when i handed it to you it looked like this:

```py
subparserer = parser.add_subparsers()

     parser.add_argument("obj")
     for subtool in subtools:
         subtool_parser = subtool.__class__.get_arg_parser()
         subtool_parser.set_defaults(obj=subtool)
         subparserer.add_parser(subtool.__class__.get_arg_parser())

     args = parser.parse_args()

     subtool = args.obj
```

The key here is we add the argument `obj` to the parser, then for each sub parser we override `obj`  with the `AbstractSubtool` that `subtool_parser` is associated with. Given we only ever use one subtool, which select at the start, this `obj` value will only be overriden by the subtool you choose. (Note `obj` is a bad name. Here is is clearly a class).

After this we can do something like `subtool().execute()` or whatever it can do. You should probably try changing this back.

util/gem5-resources-manager-v2/src/gem5_resources_manager/abstract_subtool.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+from argparse import ArgumentParser, Namespace
+from .data_source.abstract_data_source import AbstractDataSource
+
+
+class AbstractSubtool(ABC):
+    def get_arg_parser() -> ArgumentParser:
```
BobbyRBruce, line 7: Due to your redesign, this is no longer the signature of this function.

util/gem5-resources-manager-v2/src/gem5_resources_manager/abstract_subtool.py
```diff
@@ -0,0 +1,11 @@
...
+        raise NotImplementedError
+
+    def execute(args: Namespace, data_source: AbstractDataSource):
```
BobbyRBruce, line 10: Same story here: this isn't a member function, it's a class function.

util/gem5-resources-manager-v2/src/gem5_resources_manager/creator/creator.py
```diff
@@ -0,0 +1,154 @@
...
+            "-f",
+            type=str,
+            help="json file dictionary containing field entries",
+            default="{}",
+        )
+
+    def execute(self, args: Namespace, data_source: AbstractDataSource):
+        self.data_source = data_source
+        print("Creator.execute()", args.category)
+        self.create_resource_json(args)
```
BobbyRBruce, line 74: This is messy. Why are we calling execute, setting the member variable `data_source` then calling the function `create_resource_json` with `args`. Then in `create_resource_json` using both `args` and `self.data_source`. If you think it through, why not just move `create_resource_json` into `execute`? And why have member variables, or even have these as member functions?

Do we need a `Creator` Object to hold ata and state information? Isn't this a better case for functional programming?

util/gem5-resources-manager-v2/src/gem5_resources_manager/creator/creator.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def create_resource_json(self, args: Namespace):
+        resource = {}
+        required, optional = self.data_source.get_fields(args.category)
+        populated_fields = ast.literal_eval(args.field_entries)
+        resource["category"] = args.category
+        del required["category"]
```
BobbyRBruce, line 81: Here you are just setting then deleting a variable

util/gem5-resources-manager-v2/src/gem5_resources_manager/creator/creator.py
```diff
@@ -0,0 +1,154 @@
...
+            resource_id=resource["id"]
+        ):
+            # throw runtime exception
+            raise Exception("Resource already exists in database")
+
+        if not args.ignore_schema_validation:
+            self.data_source.validate_resource(resource)
+        self.data_source.create_new_entry(resource)
+
+    def enter_fields(self, fields, resource, populated_fields, args, is_optional=False):
```
BobbyRBruce, line 99: There's so much confusion here. What information do we want? I think

* The list of fiends for each category.
* For each field a to know if ti's optonal.
* The "type" fo the field (its "schema").
* If we're running in `verbose` mode.
* If `is_optional`.

So, how about you create a new class called `FieldType` which encapulates all the schema information about a field. This woulde include info on whether its optinal, andprovide the field's accepted value types as well as a validator. We can then have a function which returns a List of `Field` objects for that category

Also, let's rename `is_optional` to `ask_on_optionmal_param`. `is_optional` doesn't have any meaning to me.

 Here's the basic pseudo-coee:



```py
field_types = get_field_types(category="binary")

for field_type in field_types:
    if field_type.is_optional():
        if not ask_on_optional_param:
            continue # No inclusion of optional parameters.
        # Here you do "choice logic" asking if they want to 
        # set the parameter. If "no" `continue` called.
    
    val = None
    while not val:
        print(f"The field {field_type.get_name()} expects a value of {field_type.get_value_type_as_str()}")
        val = input('bla bla')
        if not field_type.is_valid_value(val):
            val = None
            print("Incorrect value type!")
 
    # Set the value to the resource here
```

And we've discussed thia many, many times now, and I've still got no movement on it: Can we please cleanup this default value/None value/field missing mess. I really think it'd help a lot if we implemented the following rule:

1. A missing field is consider to have a value of `None`.
2. In line with 1. an "Optional" field is a field that can be `None` or a value.
3. We drop this default value thing. If a field is optional and not set by the user, it's stored as None. If it's not optional then the user is promted for a value which must be given, with this value being valid in accordance to the schema.

util/gem5-resources-manager-v2/src/gem5_resources_manager/creator/creator.py
```diff
@@ -0,0 +1,154 @@
...
+    def handle_url(self, url, populated_fields):
+        urllib.request.urlretrieve(url, "temp")
+        is_zipped = zipfile.is_zipfile("temp")
+        is_tarred = tarfile.is_tarfile("temp")
+        # if url ends with .tar or .tar.gz, then it is tarred
+        if url.endswith(".tar"):
+            is_tarred = True
+        if url.endswith(".tar.gz"):
+            is_tarred = True
+            is_zipped = True
```
BobbyRBruce, line 140: Don't do it this way.  You can't infer file type by extension. Use Python's tar and gzip library

util/gem5-resources-manager-v2/src/gem5_resources_manager/creator/creator.py
```diff
@@ -0,0 +1,154 @@
...
+
+        def file_as_bytes(file):
+            with file:
+                return file.read()
+
+        populated_fields["is_zipped"] = is_zipped
+        populated_fields["is_tar_archive"] = is_tarred
+        populated_fields["md5sum"] = hashlib.md5(
+            file_as_bytes(open("temp", "rb"))
```
BobbyRBruce, line 149: This won't work for directories. Import the code from md5.py

util/gem5-resources-manager-v2/src/gem5_resources_manager/creator/creator.py
```diff
@@ -0,0 +1,154 @@
...
+                or fields[field]["type"] == "object"
+            ):
+                resource[field] = ast.literal_eval(user_input)
+            else:
+                resource[field] = user_input
+            if field == "url" and args.handle_url:
+                self.handle_url(resource[field], populated_fields)
+
+    def handle_url(self, url, populated_fields):
+        urllib.request.urlretrieve(url, "temp")
```
BobbyRBruce, line 132: use "pathutil" to automatically create a tempfile

util/gem5-resources-manager-v2/src/gem5_resources_manager/data_source/atlas_data_source.py
```diff
@@ -0,0 +1,170 @@
...
+            client.close()
+            raise ConnectionFailure("Collection does not exist")
+
+        self.collection = collection
+
+    def close(self):
+        pass
+
+    def resource_exists(
+        self, resource_id: str, resource_version: Optional[str] = "1.0.0"
```
BobbyRBruce, line 56: Why does the `resource_version` have a default?

util/gem5-resources-manager-v2/src/gem5_resources_manager/data_source/atlas_data_source.py
```diff
@@ -0,0 +1,170 @@
...
+            )
+            .sort("resource_version", -1)
+            .limit(1)
+        )
+        json_resource = json_util.dumps(resource)
+        res = json.loads(json_resource)
+        if res == []:
+            return False
+        return True
+
```
BobbyRBruce, line 74:
```suggestion
        return bool(res)

```

util/gem5-resources-manager-v2/src/gem5_resources_manager/data_source/atlas_data_source.py
```diff
@@ -0,0 +1,170 @@
...
+        self, resource_id: str, resource_version: str
+    ) -> ResourceCategory:
+        resource = (
+            self.collection.find(
+                {
+                    "id": resource_id,
+                    "resource_version": resource_version,
+                },
+                {"_id": 0},
+            )
```
BobbyRBruce, line 85: You're writing this over and over. Make a private function calkled "get_resource" so you just write it the once.

util/gem5-resources-manager-v2/src/gem5_resources_manager/data_source/atlas_data_source.py
```diff
@@ -0,0 +1,170 @@
...
+            raise Gem5DataSourceEntryNotFound(resource_id)
+        return res[0]
+
+    def get_resource_from_file(self, file):
+        # append resource to resources.json
+        resources = None
+        if os.path.exists(file):
+            with open(file) as infile:
+                resources = json.load(infile)
+        return resources
```
BobbyRBruce, line 133: WHy is this here? This has nothing to do with the Atlas DB.

util/gem5-resources-manager-v2/src/gem5_resources_manager/data_source/dummy_database.py
BobbyRBruce, line 1: I didn't bother to review this file RN.

util/gem5-resources-manager-v2/src/gem5_resources_manager/data_source/json_data_source.py
BobbyRBruce, line 1: I didn't bother to review this file RN.

util/gem5-resources-manager-v2/src/gem5_resources_manager/schemas/json_validator.py
BobbyRBruce, line 1: I won't review this on this opas but i'll say you **need** to add comments here to explain your logic. This code smells. Too much hard coding, if statements, random deletion of fields. You need to comminicate the flow of logic here.

BobbyRBruce (changes requested):

util/gem5-resources-manager-v2/src/gem5_resources_manager/creator/creator.py
Harshil2107: Here is my interpretation of this:
- If a field is missing that is the key does not exist then it means that the resource doesnot have the field
- Currently the fields cannot be None or Null but if they are lets say "" or [] it means that the resource has the field but the value is an empty string or empty array.

BobbyRBruce:
@Harshil2107 : I'll let you reopen this if you feel differently but I think we'll never get this in and should just close this and call it a dead end project (we can always change our mind later). If we didn't need it in the last 6 months, I doubt we'll ever need it. 

You may want to see if there's anything here worth salvaging given your work on gem5-resources in the past few months. Probably not of all of it, but maybe some? It's up to you. If you still feel like this may be of use at some point but you won't work on it anytime soon, create an issue explaining the idea a bit and link to this so someone else could theoretically have a go

Harshil2107:
I moved away from needing a cli, as I feel that I only have 2 use cases that are prettry far away.

1: Just add of update one (or a few very different ones) resource for that the current resource manager works

2: To add or update a bunch of resources and in most cases these are unique among each change so I think just writing a script to create the JSON is faster.

I did push some helpful scripts to gem5/util if I was using the same ones a decent amount.

PR #313, opened 2023-09-13
scons: Revert "Add an option specifying the path to mold linker binary"

### LINKED ISSUE #312: Specifying mold linker does not actually use mold linker
**Describe the bug**
Specifying mold linker does not actually use mold linker

**Affects version**
develop, starting after #244 

**gem5 Modifications**
No modifications

**To Reproduce**
scons --linker=mold  -j24 build/VEGA_X86/gem5.opt -k  --mold-path=/usr/local/bin/mold

See notes about Ubuntu 20 below.

If writing code, or a terminal command, use code blocks. Either an inline code block, \`scons build/ALL/gem5.opt\` (enclosed in two \`) or a multi-line codeblock:

**Expected behavior**
This:

image

Not this:

image

**Host Operating System**
Ubuntu 20.04

**Host ISA**
amd64

**Compiler used**
gcc 9.4.0

**Additional information**
Ubuntu 20.04 (in contrast to 22.04) does not have the mold linker as part of the apt / snap.  Therefore, most people built from source.  Like most linux configure && make && make install, this goes in /usr/local.  This worked just fine util #244 . Specifying the direct path should just work in my opinion and requesting users to 'just upgrade to 22.04' is not a reasonable option.

BobbyRBruce:
Reverts gem5/gem5#244

Fixes #312

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -84,7 +84,6 @@ from os import mkdir, remove, environ
 from os.path import abspath, dirname, expanduser
 from os.path import isdir, isfile
 from os.path import join, split
-from pathlib import Path
 
 import logging
 logging.basicConfig()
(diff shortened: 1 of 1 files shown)


powerjg:
@hnpl

BobbyRBruce:
Arg. The need for Gerrit IDs strikes again.

BobbyRBruce:
Tests should pass now.

hnpl:
The expected behavior of the option is explained here, https://github.com/gem5/gem5/pull/244#issuecomment-1702246735.

There's no easy way to determine whether mold is available as you'd have to check the elf of the compiled test binary in scons.

I'm not opposed to reverting this change though I'd rather keep this as an option.

abmerop:
Hi Hoa, 

I first noticed the issue on the 'thunderbird' machine.  Since the stars aligned and you happen to have access, you could test there if you like.

hnpl:
Thanks for the info! Let's revert #244 until I figure out a fix.

PR #316, opened 2023-09-13
mem-ruby: patch fixes a protocol error in MOESI_CMP_Directory

gPatsWat:
When there is race between FwdGetX
and PUTX on owner. Owner in this case hands off
ownership to GetX requestor and PUTX still goes
through. But since owner has changed, state should go back to M and PUTX is essentially trashed.
An Unblock to the Directory in this case will give an undefined transition. I have added transitions which indicate that when an Unblock is served to the Directory, it means that some kind of ownership transfer has happened while a PUTX/PUTO was in progress.

Change-Id: I37439b5a363417096030a0875a51c605bd34c127

### DIFF
--- a/src/mem/ruby/protocol/MOESI_CMP_directory-dir.sm
+++ b/src/mem/ruby/protocol/MOESI_CMP_directory-dir.sm
@@ -1014,6 +1014,31 @@ machine(MachineType:Directory, "Directory protocol")
     i_popIncomingRequestQueue;
   }
 
+  //this happens when there is race between FwdGetX
+  //and PUTX on owner. Owner in this case hands off
+  //ownership to GetX requestor and PUTX still goes
+  //through. But since owner has changed, state should
+  //go back to M and PUTX is essentially trashed.
+  transition(MI, Unblock, M) {
+    w_deallocateTBE;
+    j_popIncomingUnblockQueue;
+  }
+
+  transition(MIS, Unblock, M) {
+    w_deallocateTBE;
+    j_popIncomingUnblockQueue;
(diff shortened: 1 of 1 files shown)


gPatsWat:
reproducing this is hard as the errors were found in ruby_random_tester after a lot of loads. but please let me know if the race condition requires more explanation

mattsinc (commented):
Can you please comment more on how you were using the ruby random tester for this?  For example, you mention in your commit message you added O transitions, and I can see why that could be a problem too, but if we can't test it, I'm worried we can't really verify if it is or isn't correct...

gPatsWat:
Before I mention the ruby tester parameters I want to note that I used a modified version of MOESI_CMP_Directory to support explicit clusters (as opposed to using numa bits). For the ruby tester, I used --maxloads=1000000, num cpus = 8, num clusters= 4.

The Ruby random has to be run for a long time, which I could do by increasing maxloads (don't know any other way). Though I think these errors can be reproduced for a smaller system.

gPatsWat:
fwd_getx_putx_race
This explains one of the races which occur, numbered in order of events accordingly.
C1, C2, C3, C4 are the 4 clusters, each with 2 cores

powerjg (approved):
Seems good to me. Thanks for the fix and the detailed explanation!

BobbyRBruce:
@mattsinc : Does this explanation satisfy?

mattsinc (approved):
agreed, thanks!

gPatsWat:
hi, whom do I request to merge this? : -)

PR #317, opened 2023-09-14
util,resources,stdlib: Add 'obtain-resource.py' utility to easily obtain resources from the CLI

BobbyRBruce:
This allows users to obtain resources via the CLI instead of having to write a python script to do so. It is essentially a nice CLI wrapper for "gem5.resources.resource.obtain_resource"

## Usage

```sh
> scons build/ALL/gem5.opt -j `nproc`
> ./build/ALL/gem5.opt util/obtain-resource.py --help

usage: obtain-resource.py [-h] [-p PATH] [-q] id

positional arguments:
  id                    The resource id to download.

options:
  -h, --help            show this help message and exit
  -p PATH, --path PATH  The path the resource is to be downloaded to. If not specified, the resource will be downloaded to the default
                        location in the gem5 local cache of resources
  -q, --quiet           Suppress output.
```

E.g.:

```sh
./build/ALL/gem5.opt util/obtain-resource.py arm-hello64-static -p arm-hello
```

Will download the resource with ID `arm-hello64-static` to `arm-hello` in the CWD.

### DIFF
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -621,6 +621,7 @@ def obtain_resource(
     resource_version: Optional[str] = None,
     clients: Optional[List] = None,
     gem5_version=core.gem5Version,
+    to_path: Optional[str] = None,
     quiet: bool = False,
 ) -> AbstractResource:
     """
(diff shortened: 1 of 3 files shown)


src/python/gem5/resources/resource.py
```diff
@@ -654,47 +662,56 @@ def obtain_resource(
...
+
+        # Here we ensure the directory in which the resource is to be stored
+        # is created.
+        #
+        # `exist_ok=True` here as, occasionally, if multiple instance of gem5
+        # are started simultaneously, a race condition can exist to create the
+        # resource directory. Without `exit_ok=True`, threads which lose this
+        # race will thrown a `FileExistsError` exception. `exit_ok=True`
+        # ensures no exception is thrown.
+        Path(to_path).parent.mkdir(parents=True, exist_ok=True)
```
powerjg: You may want to wrap this in a try/except and print a nice error if it fails. E.g., if the user specifies a `to_path` that they don't have permission for it would be nice to not have a backtrace but instead have a `fatal`. 

This is a minor nit. Feel free to ignore.

BobbyRBruce: I've added a helpful `fatal` here.

powerjg (approved):

PR #323, opened 2023-09-15
stdlib, resources: Added pretty printing resource

Harshil2107:
- Implemented a __str__  for AbstractResource __str__ prints resource category, id and version.
link to resources website is also printed.

Change-Id: Iad5825ff7d8d505ceb236e00dc49bb56055fc8f0

### DIFF
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -70,7 +70,6 @@ class AbstractResource:
 
     def __init__(
         self,
-        category: Optional[str] = None,
         id: Optional[str] = None,
         resource_version: Optional[str] = None,
         local_path: Optional[str] = None,
@@ -94,14 +93,22 @@ class AbstractResource:
                 "exist."
             )
         self._id = id
-        self._category = category
         self._local_path = local_path
         self._description = description
         self._source = source
         self._version = resource_version
 
+    def get_category_name(cls) -> str:
+        raise NotImplementedError
+
     def __str__(self):
-        return f"{self._category} resource({self._id}, {self._version})\nFor more information, please visit https://resources.gem5.org/resources/{self._id}?version={self._version}"
+        message = (
+            f"{self.get_category_name()}({self._id}, {self._version})\n"
+            "For more information, please visit "
+            f"https://resources.gem5.org/resources/{self._id}?"
+            f"version={self._version}"
+        )
+        return message
 
     def get_resource_version(self) -> str:
         """Returns the version of the resource."""
@@ -129,7 +136,6 @@ class FileResource(AbstractResource):
         self,
         local_path: str,
         id: Optional[str] = None,
-        category: Optional[str] = None,
         resource_version: Optional[str] = None,
         description: Optional[str] = None,
         source: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


src/python/gem5/resources/resource.py
```diff
@@ -25,6 +25,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 from abc import ABCMeta
+import json
```
powerjg: Needed?

src/python/gem5/resources/resource.py
```diff
@@ -741,6 +800,7 @@ def obtain_resource(
     # Once we know what AbstractResource subclass we are using, we create it.
     # The fields in the JSON object are assumed to map like-for-like to the
     # subclass contructor, so we can pass the resource_json map directly.
+
```
powerjg: Extra newline?

powerjg (commented):
lgtm. Two tiny things

src/python/gem5/resources/resource.py
```diff
@@ -91,12 +93,16 @@ def __init__(
...
-
+        self._id = id
+        self._category = category
         self._local_path = local_path
         self._description = description
         self._source = source
         self._version = resource_version
 
+    def __str__(self):
+        return f"{self._category} resource({self._id}, {self._version})\nFor more information, please visit https://resources.gem5.org/resources/{self._id}?version={self._version}"
```
BobbyRBruce: I don't think black cares about strings breaking line length rules so i typically do it manually.

```suggestion
        return f"{self._category} resource({self._id}, {self._version})\n"
                "For more information, please visit "
               f"https://resources.gem5.org/resources/{self._id}?"
               f"version{self._version}"
```

In python

```py
a_string = "hello this is an example of a string that may be a bit too long"
```

is the same as

```py
a_string = "hello "
           "this is an"
           " example of a string that may be "
           "a bi"
           "t too "
           "long"
```

src/python/gem5/resources/resource.py
```diff
@@ -122,6 +128,8 @@ class FileResource(AbstractResource):
     def __init__(
         self,
         local_path: str,
+        id: Optional[str] = None,
+        category: Optional[str] = None,
```
BobbyRBruce: Instead of having this as a parameter why not have the `AbstractResource` class enforce the creation of a `def get_category_name(cls) -> str` function. Then each of the categories has this implemented.

BobbyRBruce (changes requested):

src/python/gem5/resources/resource.py
Harshil2107: The second commit removes the unneeded import, I will rebase everything into one commit after all requested changes are done.

src/python/gem5/resources/resource.py
Harshil2107: I removed this extra line in the second commit.

powerjg (approved):

BobbyRBruce (approved):
Not sure if having `get_category_name` return the class name is my favorite choice here but it's fine for now and can be fixed later. Approving.

PR #325, opened 2023-09-18
arch-riscv: Fix inst flags for jal and jalr

rogerchang23424:
The jal and jalr share the same instruction format JumpConstructor, which sets the IsCall and IsReturn flags by the register ID. However, it may cause wrong instruction flags set for jal because the section "handle the 'Jalr' instruction" misses the opcode checking. The PR fix the issue to ensure the IsReturn can be only set in Jalr.

Change-Id: I9ad867a389256f9253988552e6567d2b505a6901


powerjg (approved):
I'm not sure why the tests are failing... We'll look into it.

PR #327, opened 2023-09-18
cpu,mem,sim: Optimize code using MAGPIE

BobbyRBruce:
This patch was created using the Machine Automated General Performance Improvement via Evolution of Software (MAGPIE) framework (https://github.com/bloa/magpie).

It made the following optimizations:

* In "src/mem/snoop_filter.{cc/hh}": Moves the body of `finishRequest` inline from .cc to .hh; Adds const to args in `finishRequest`.
* In "src/mem/eventq.{cc/hh}": Inlines `aquire` and `release` from .cc .hh.
* In "src/cpu/pc_event.hh": Adds `empty` to check return `pcMap.empty()`
* In "src/cpu/simple/base.cc": Adds check in `checkPcEventQueue` to skip checks if `pcEventQueue.empty()`.
* In "src/cpu/simple/timing.{cc/hh}: `updatecycleCounts()` inlined from .cc to .hh.

MAGPIE Bibtex:

```
https://github.com/Article{blot:2022:magpie,
  author    = {Aymeric Blot and
               Justyna Petke},
  title     = {{MAGPIE:} {M}achine Automated General Performance Improvement via Evolution of Software},
  journal   = {Computing Research Repository},
  volume    = {abs/2208.02811},
  url       = {https://arxiv.org/abs/2208.02811},
  doi       = {10.48550/arXiv.2208.02811},
  year      = {2022},
}
```

Co-authored-by: Americ Blot <a.blot@cs.ucl.ac.uk>
Co-authored-by: Justyna Petke <j.petke@ucl.ac.uk>
Co-authored-by: William B. Langdon <w.langdon@cs.ucl.ac.uk>

### DIFF
--- a/src/cpu/simple/base.cc
+++ b/src/cpu/simple/base.cc
@@ -122,6 +122,9 @@ BaseSimpleCPU::BaseSimpleCPU(const BaseSimpleCPUParams &p)
 void
 BaseSimpleCPU::checkPcEventQueue()
 {
+    //if empty now then no whole PCEventQueue is empty so no need to check
+    if (threadInfo[curThread]->thread->pcEventQueue.empty()) return;
+
     Addr oldpc, pc = threadInfo[curThread]->thread->pcState().instAddr();
     do {
         oldpc = pc;
--- a/src/sim/eventq.cc
+++ b/src/sim/eventq.cc
@@ -109,20 +109,6 @@ Event::insertBefore(Event *event, Event *curr)
 }
 
 void
-Event::acquire()
-{
-    if (flags.isSet(Event::Managed))
-        acquireImpl();
-}
-
-void
-Event::release()
-{
-    if (flags.isSet(Event::Managed))
-        releaseImpl();
-}
-
-void
 Event::acquireImpl()
 {
 }
--- a/src/cpu/pc_event.hh
+++ b/src/cpu/pc_event.hh
@@ -121,6 +121,7 @@ class PCEventQueue : public PCEventScope
 
         return doService(pc, tc);
     }
+    bool empty() const { return pcMap.empty(); }
 
     range_t equal_range(Addr pc);
     range_t equal_range(PCEvent *event) { return equal_range(event->pc()); }
--- a/src/cpu/simple/timing.cc
+++ b/src/cpu/simple/timing.cc
@@ -1077,16 +1077,6 @@ TimingSimpleCPU::completeDataAccess(PacketPtr pkt)
 }
 
 void
-TimingSimpleCPU::updateCycleCounts()
-{
-    const Cycles delta(curCycle() - previousCycle);
-
-    baseStats.numCycles += delta;
-
-    previousCycle = curCycle();
-}
-
-void
 TimingSimpleCPU::DcachePort::recvTimingSnoopReq(PacketPtr pkt)
 {
     for (ThreadID tid = 0; tid < cpu->numThreads; tid++) {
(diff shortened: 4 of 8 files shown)


src/cpu/simple/base.cc
```diff
@@ -383,8 +386,10 @@ BaseSimpleCPU::preExecute()
     //If we decoded an instruction this "tick", record information about it.
     if (curStaticInst) {
 #if TRACING_ON
+      if (debug::ExecEnable) { //getInstRecord does nothing unless ExecEnable
```
giactra: This looks wrong to me. The getInstRecord is called irrespective of the tracer (the method is part of the interface). The Exec tracer (enebled by ExecEnable) is only one of them...

src/sim/eventq.cc
```diff
@@ -223,7 +209,6 @@ EventQueue::remove(Event *event)
 Event *
 EventQueue::serviceOne()
 {
-    std::lock_guard<EventQueue> lock(*this);
```
andysan: This is definitely incorrect. Removing this line removes the mutex that protects the event queue in multi-threaded simulation.

andysan (changes requested):
The changes that move implementations to header files in this change look correct, but I doubt it makes any differences in the case when the method isn't virtual and the caller is in the same source file. Such cases should already be inlined.

The change to the mutex is definitely incorrect and will introduce a race condition when using multiple event queues.

src/cpu/simple/base.cc
BobbyRBruce: The `getInstRecord` function body is:

```cpp
if (!debug::ExecEnable)
        return NULL;

    return new InstPBTraceRecord(*this, when, tc, si, pc, mi);
```

My unease at this change is more that it couples the logic of this and the logic of `getInstRecord` if we were to update `getInstRecord` this could break without us knowing.

Essentially, the optimization here is just saving a function call which this tool seems to think is worth doing. 

I think i'd probably vote to remove this change too but mostly due to it being a bit brittle.

BobbyRBruce:
> The changes that move implementations to header files in this change look correct, but I doubt it makes any differences in the case when the method isn't virtual and the caller is in the same source file. Such cases should already be inlined.
> 
> The change to the mutex is definitely incorrect and will introduce a race condition when using multiple event queues.

I've yet to update the PR description here but, yeah, this is a pretty niche case. The researcher who did this, Bill Langdon (UCL), was mostly interested in optimizing simulation in SE mode, timing CPU, running "RNAFold". The tool mentioned works by optimising ASTs based on some target run.

I uploaded this pretty blindly on his behalf, but I agree that the lock removal should be reverted. It's not an optimization in all cases. the inlines seem pretty harmless and probably do save a little time.

I'll let ya all know when this is ready for review. still talking with Bill.

src/sim/eventq.cc
BobbyRBruce: Agreed, will revert

src/cpu/simple/base.cc
giactra: The getInstRecord is a virtual method implemented by every tracer we support in gem5.

See for example:

https://github.com/gem5/gem5/blob/stable/src/arch/arm/tracers/tarmac_tracer.cc#L100

The ExecEnable is just specific to ONE of them (the ExecTracer), even though I see now some other tracers are relying on it like the InstPBTracer... This is a bit misleading.

Anyway with this change basically we would break all existing tracers not using the ExecEnable flag like the example posted above

BobbyRBruce:
Closing this as the researcher's working on this are still deciding if they want to try further optimizations/improvements. This PR can be re-opened as needed.

PR #328, opened 2023-09-19
dev-amdgpu: Handle GPU atomics on host memory addresses

abmerop:
It is possible to execute a GPU atomic instruction using a memory address that is in the host memory space (e.g, HMM, __managed__, hipHostMalloc'd address). Since these are in host memory they are passed to the SystemHub DmaDevice. However, this currently executes as a write packet without modifying data. This leads to hangs in applications that use atomics for forward progress (e.g., HeteroSync).

It is not clear where these are handled on a real GPU, but they are certainly not handled by the software stack nor driver, so they must be handled in hardware and therefore implemented in gem5. Handling for atomics in the SystemHub makes the most sense.

To make atomics work a few extra changes need to be made to the SystemHub. (1) The atomic is implemented as a host memory read, followed by calling the AtomicOpFunctor, followed by a write. This requires a second event to handle read response, performing atomic, and issuing a write. (2) Atomics must be serialized otherwise two atomics might return the same value which is incorrect. This patch adds serialization logic for all request types to the same address to handle this. (3) With the added complexity of the SystemHub, a new debug flag explicitly for SystemHub is added.

Testing done: The heterosync application with input "sleepMutex 10 16 4" previously hung before this patch. It passes with the patch applied. This application tests both (1) and (2) above, as it allocates locks with hipHostMalloc and has multiple workgroups sending an atomic request in the same Tick, verifying the serialization mechanism.

Change-Id: Ife84b30037d1447dd384340cfeb06fdfd472fff9

### DIFF
--- a/src/dev/amdgpu/system_hub.cc
+++ b/src/dev/amdgpu/system_hub.cc
@@ -31,6 +31,8 @@
 
 #include "dev/amdgpu/system_hub.hh"
 
+#include "debug/AMDGPUSystemHub.hh"
+#include "mem/packet_access.hh"
 #include "mem/port.hh"
 
 namespace gem5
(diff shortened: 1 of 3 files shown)


src/dev/amdgpu/system_hub.cc
```diff
@@ -39,16 +41,92 @@ namespace gem5
 void
 AMDGPUSystemHub::sendRequest(PacketPtr pkt, Event *callback)
 {
-    ResponseEvent *dmaRespEvent = new ResponseEvent(callback);
+    // Some requests, in particular atomics, need to be sent in order
+    // to received the correct values. If there is an atomic in progress
```
mattsinc: minor: received --> receive

mattsinc (changes requested):

mattsinc:
Also I think this only works for GPUFS, right?

abmerop:
> Also I think this only works for GPUFS, right?

Yes, that is correct.  In general anything in src/dev/amdgpu is full system only

mattsinc (approved):

PR #329, opened 2023-09-19
cpu: Explicitly define cache_line_size -> 64-bit unsigned int

hnpl:
While it's plausible to define the cache_line_size as a 32-bit unsigned int, the use of cache_line_size is way out of its original scope.

cache_line_size has been used to produce an address mask, which masking out the offset bits from an address. For example, [1], [2], [3], and [4]. However, since the cache_line_size is an "unsigned int", the type of the value is not guaranteed to be 64-bit long. Subsequently, the bit twiddling hacks in [1], [2], [3], and [4] produce 32-bit mask, i.e., 0x00000000FFFFFFC0.

This behavior at least caused a problem in LLSC in RISC-V [5], where the load reservation (LR) relies on the mask to produce the cache block address. Two distinct 64-bit addresses can be mapped to the same cache block using the above mask.

This patch explicitly defines cache_line_size as a 64-bit unsigned int so the cache block mask can be produced correctly for 64-bit addresses.

[1] https://github.com/gem5/gem5/blob/3bdcfd6f7abdf0c9125ec37018df76a29998a055/src/cpu/simple/atomic.hh#L147
[2] https://github.com/gem5/gem5/blob/3bdcfd6f7abdf0c9125ec37018df76a29998a055/src/cpu/simple/timing.hh#L224
[3] https://github.com/gem5/gem5/blob/3bdcfd6f7abdf0c9125ec37018df76a29998a055/src/cpu/o3/lsq_unit.cc#L241
[4] https://github.com/gem5/gem5/blob/3bdcfd6f7abdf0c9125ec37018df76a29998a055/src/cpu/minor/lsq.cc#L1425
[5] https://github.com/gem5/gem5/blob/3bdcfd6f7abdf0c9125ec37018df76a29998a055/src/arch/riscv/isa.cc#L787

Change-Id: I29abc7aaab266a37326846bbf7a82219071c4ffe

### DIFF
--- a/src/cpu/base.cc
+++ b/src/cpu/base.cc
@@ -257,8 +257,8 @@ BaseCPU::mwait(ThreadID tid, PacketPtr pkt)
     AddressMonitor &monitor = addressMonitor[tid];
 
     if (!monitor.gotWakeup) {
-        int block_size = cacheLineSize();
-        uint64_t mask = ~((uint64_t)(block_size - 1));
+        Addr block_size = cacheLineSize();
+        Addr mask = ~(block_size - 1);
 
         assert(pkt->req->hasPaddr());
         monitor.pAddr = pkt->getAddr() & mask;
@@ -282,8 +282,8 @@ BaseCPU::mwaitAtomic(ThreadID tid, ThreadContext *tc, BaseMMU *mmu)
     RequestPtr req = std::make_shared<Request>();
 
     Addr addr = monitor.vAddr;
-    int block_size = cacheLineSize();
-    uint64_t mask = ~((uint64_t)(block_size - 1));
+    Addr block_size = cacheLineSize();
+    Addr mask = ~(block_size - 1);
     int size = block_size;
 
     //The address of the next line if it crosses a cache line boundary.
--- a/src/cpu/base.hh
+++ b/src/cpu/base.hh
@@ -143,7 +143,7 @@ class BaseCPU : public ClockedObject
     bool _switchedOut;
 
     /** Cache the cache line size that we get from the system */
-    const unsigned int _cacheLineSize;
+    const uint64_t _cacheLineSize;
 
     /** Global CPU statistics that are merged into the Root object. */
     struct GlobalStats : public statistics::Group
@@ -394,7 +394,7 @@ class BaseCPU : public ClockedObject
     /**
      * Get the cache line size of the system.
      */
-    inline unsigned int cacheLineSize() const { return _cacheLineSize; }
+    inline uint64_t cacheLineSize() const { return _cacheLineSize; }
 
     /**
      * Serialize this object to the given output stream.
--- a/src/cpu/base.hh
+++ b/src/cpu/base.hh
@@ -143,7 +143,7 @@ class BaseCPU : public ClockedObject
     bool _switchedOut;
 
     /** Cache the cache line size that we get from the system */
-    const uint64_t _cacheLineSize;
+    const Addr _cacheLineSize;
 
     /** Global CPU statistics that are merged into the Root object. */
     struct GlobalStats : public statistics::Group
@@ -394,7 +394,7 @@ class BaseCPU : public ClockedObject
     /**
      * Get the cache line size of the system.
      */
-    inline uint64_t cacheLineSize() const { return _cacheLineSize; }
+    inline Addr cacheLineSize() const { return _cacheLineSize; }
 
     /**
      * Serialize this object to the given output stream.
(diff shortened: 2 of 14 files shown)


giactra:
Thanks Hoa for this patch. uint64_t seems like a good solution; may I propose to use Addr as an alternative?

Addr is already a typedef of uint64_t and in my opinion it makes conceptually sense as the cacheline size is usually used in arithmetic/bitwise operations with addresses.

I also noticed from a quick grep there are other classes storing the cacheLineSize as an integer. The system class and the mem_proxy are two notable examples. As you are already undertaking the effort of replacing machine dependent types to explicitly sized ones, could you apply the same change to those classes? It would be great if we could unify the cacheLineSize representation.

hnpl:
> Thanks Hoa for this patch. uint64_t seems like a good solution; may I propose to use Addr as an alternative?
> 
> Addr is already a typedef of uint64_t and in my opinion it makes conceptually sense as the cacheline size is usually used in arithmetic/bitwise operations with addresses.
> 
> I also noticed from a quick grep there are other classes storing the cacheLineSize as an integer. The system class and the mem_proxy are two notable examples. As you are already undertaking the effort of replacing machine dependent types to explicitly sized ones, could you apply the same change to those classes? It would be great if we could unify the cacheLineSize representation.

I'll update all variables using cache_line_size / cacheLineSize to Addr.

I'm still quite hesitant using Addr for the cache line size variables because semantically, when I see the type Addr, I would expect the variable to contain an address.

hnpl:
It looks like this is failing, `atomic-cpu_1-cores_classic_SingleChannelDDR3_1600_riscv-boot-test_to-tick-ALL-x86_64-opt`.

BobbyRBruce:
> It looks like this is failing, `atomic-cpu_1-cores_classic_SingleChannelDDR3_1600_riscv-boot-test_to-tick-ALL-x86_64-opt`.

Double check the `testlib-quick...` artifact here: https://github.com/gem5/gem5/actions/runs/6254190584?pr=329. It contains the "tests/testing-results" directory which has the stdout and stderr.

I've been having some issues with resources downloading. It may just be that.

hnpl:
> > It looks like this is failing, `atomic-cpu_1-cores_classic_SingleChannelDDR3_1600_riscv-boot-test_to-tick-ALL-x86_64-opt`.
> 
> Double check the `testlib-quick...` artifact here: https://github.com/gem5/gem5/actions/runs/6254190584?pr=329. It contains the "tests/testing-results" directory which has the stdout and stderr.
> 
> I've been having some issues with resources downloading. It may just be that.

Is there a way to obtain the artifact other than downloading from GitHub? The artifact is > 1GiB and the download speed from GitHub is ~20KiB/s on my side.

BobbyRBruce:
> > > It looks like this is failing, `atomic-cpu_1-cores_classic_SingleChannelDDR3_1600_riscv-boot-test_to-tick-ALL-x86_64-opt`.
> > 
> > 
> > Double check the `testlib-quick...` artifact here: https://github.com/gem5/gem5/actions/runs/6254190584?pr=329. It contains the "tests/testing-results" directory which has the stdout and stderr.
> > I've been having some issues with resources downloading. It may just be that.
> 
> Is there a way to obtain the artifact other than downloading from GitHub? The artifact is > 1GiB and the download speed from GitHub is ~20KiB/s on my side.

No, the artifacts need to exist on GitHub servers. However, PR #263 will split the artifacts up on a per-test-directory basis so you can pull the artifact you want instead of all of them.

Edit: Also... I don't think we can accommodate you either way if you're experiencing below dial-up level internet speeds.

hnpl:
I don't think the internet speed is on my side. The Google internet speed test shows my workstation has ~1Gbps downloading/uploading w.r.t traffic from/to the Bay Area.

However, the GitHub download speed is truly abysmal on my side.

Edit: my ethernet UC Davis workstation

BobbyRBruce:
> I don't think the internet speed is on my side. The Google internet speed test shows my workstation has ~1Gbps downloading/uploading w.r.t traffic from/to the Bay Area.
> 
> However, the GitHub download speed is truly abysmal on my side.

I got 11MB/s when downloading the artifact on my UC Davis workstation. I don't know what's going on exactly but I don't think the problem is on GitHub's end.

hnpl:
It is what it is then. I'd let someone with a better bandwidth take over this.

BobbyRBruce (approved):
This looks fine to me. @giactra : Is it all ok with you?

giactra:
Apologies for missing these messages...

PR #341, opened 2023-09-21
arch,arch-riscv: Remove setRegOperand in VecRegOperand

rogerchang23424:
The RISC-V vector instructions still work without setRegOperand.
We should fix the register statistic issue by https://github.com/gem5/gem5/pull/360 to avoid duplicate statistic register write count

Change-Id: Ib6a52935e00c3e557b366abfcf60450dca05614d

### DIFF
--- a/src/arch/isa_parser/operand_types.py
+++ b/src/arch/isa_parser/operand_types.py
@@ -372,7 +372,6 @@ class VecRegOperand(RegOperand):
 
     def makeWrite(self):
         return f"""
-        xc->setRegOperand(this, {self.dest_reg_idx}, &tmp_d{self.dest_reg_idx});
         if (traceData) {{
             traceData->setData({self.reg_class}, &tmp_d{self.dest_reg_idx});
         }}
--- a/src/arch/riscv/insts/vector.hh
+++ b/src/arch/riscv/insts/vector.hh
@@ -579,7 +579,6 @@ class VMaskMergeMicroInst : public VectorArithMicroInst
                 memcpy(Vd + i * byte_offset, s + i * byte_offset, byte_offset);
             }
         }
-        xc->setRegOperand(this, 0, &tmp_d0);
         if (traceData)
             traceData->setData(vecRegClass, &tmp_d0);
         return NoFault;


hnpl:
To elaborate this change, the generated vector instructions modify the vector register directly making the memcpy in setRegOperand redundant.

```C++
Vadd_vvMicro<ElemType>::execute()
{
   ...
   auto &tmp_d0 = *(RiscvISAInst::VecRegContainer *)xc->getWritableRegOperand(this, 0);
   auto Vd = tmp_d0.as<vu>();
   RiscvISAInst::VecRegContainer tmp_s0;
   xc->getRegOperand(this, 0, &tmp_s0);
   auto Vs1 = tmp_s0.as<vu>();
   RiscvISAInst::VecRegContainer tmp_s1;
   xc->getRegOperand(this, 1, &tmp_s1);
   auto Vs2 = tmp_s1.as<vu>();
   [[maybe_unused]] RiscvISA::vreg_t tmp_v0;
   [[maybe_unused]] uint8_t* v0;
   if(!machInst.vm) {
       xc->getRegOperand(this, _numSrcRegs-1, &tmp_v0);
       v0 = tmp_v0.as<uint8_t>();
    }
    COPY_OLD_VD(2);;

            for (uint32_t i = 0; i < this->microVl; i++) {

            uint32_t ei = i + vtype_VLMAX(vtype, true) * this->microIdx;
            if (this->vm || elem_mask(v0, ei)) {

                        Vd[i] = Vs2[i] + Vs1[i];
                    }

            }
        ;

        xc->setRegOperand(this, 0, &tmp_d0);

}
```

However, as suggested in the original review, using setRegOperand seems to provide better statistics. https://gem5-review.googlesource.com/c/public/gem5/+/67292/comments/c48339ff_10c1c101.

As an example, setRegOperand will eventually call setReg in O3 CPU, which will record stastistics,
https://github.com/gem5/gem5/blob/f5a255c68d1ce0318da0d46f744ee7513e0acda9/src/cpu/o3/cpu.cc#L1004.

I'd favor making a new function acting similarly to setRegOperand but instead of recording the statistics and do memcpy, the new function just records the statistics. I think having the more precise statistics is important.

rogerchang23424:
If the operand is dest. The ISA parser will generate the code that writes data back to the regfile. For vector and matrix register, they write to the regfile by calling the `getWritableRegOperand` to get the pointer of `VecRegContainer` rather than `setRegOperand` to copy results to regfile, and `getWriteRegOperand` only be called if the operand is dest, so I think the `getWriteRegOperand` is enough for instruction statistics.

https://github.com/gem5/gem5/blob/f5a255c68d1ce0318da0d46f744ee7513e0acda9/src/arch/isa_parser/operand_types.py#L274

getWritableRegOperand will call the getWritableReg in O3 CPU. However, it records the dest vector register as read statistics. I think it should record as write statistics and remove `setRegOperand` to avoid duplicate statistics.

https://github.com/gem5/gem5/blob/f5a255c68d1ce0318da0d46f744ee7513e0acda9/src/cpu/o3/cpu.cc#L988

It's weird for me to trace the instruction read/write only in O3 CPU, should we also need to do read/write statistic in other CPU models?

hnpl:
> If the operand is dest. The ISA parser will generate the code that writes data back to the regfile. For vector and matrix register, they write to the regfile by calling the `getWritableRegOperand` to get the pointer of `VecRegContainer` rather than `setRegOperand` to copy results to regfile, and `getWriteRegOperand` only be called if the operand is dest, so I think the `getWriteRegOperand` is enough for instruction statistics.


https://github.com/gem5/gem5/blob/f5a255c68d1ce0318da0d46f744ee7513e0acda9/src/arch/isa_parser/operand_types.py#L274

I see. Thanks for pointing this out. I totally missed the statistics recording in `getWriteRegOperand()`.

> 
> getWritableRegOperand will call the getWritableReg in O3 CPU. However, it records the dest vector register as read statistics. I think it should record as write statistics and remove `setRegOperand` to avoid duplicate statistics.
> 
> https://github.com/gem5/gem5/blob/f5a255c68d1ce0318da0d46f744ee7513e0acda9/src/cpu/o3/cpu.cc#L988

There's probably no overcounting as `O3::getWritableReg()` records `numVecRegReads`, while `O3::setReg()` records `numVecRegWrites`.

> It's weird for me to trace the instruction read/write only in O3 CPU, should we also need to do read/write statistic in other CPU models?

Yeah, I was biased towards O3 CPU. Now I looked more into statistics recording, it seems that we have a bigger problem. I'll explain this in the next comment.

hnpl:
Currently, in statistics recording of the _destination_ register of RVV instructions, we are using the CPU's `getWritableRegOperand()` function and the `setRegOperand()` function. Here is how the statistics are recorded for each CPU type.

In O3CPU,

```
O3CPU::getWritableRegOperand()
    |----> O3CPU::getWritableReg() -> stats::numVecRegReads++
O3CPU::setRegOperand()
    |----> O3CPU::setReg() -> stats::numVecRegWrites++
```

In Simple CPU types (Atomic/Timing),

```
SimpleCPU::getWritableRegOperand() -> stats::numRegWrites[regClass]++
    |----> SimpleThread::getWritableReg()
SimpleCPU::setRegOperand() -> stats::numRegWrites[regClass]++
    |----> SimpleThread::setReg()
```

In Minor CPU,

```
MinorCPU::getWritableRegOperand()
    |----> MinorThread(SimpleThread)::getWritableReg()
MinorCPU::setRegOperand()
    |----> MinorThread(SimpleThread)::setReg()
```

It looks like,
- [1] There's a stat recording inconsistency between SimpleCPU and O3CPU. This inconsistency causes problem with or without calling `setRegOperand()`.
- [2] I'm not sure how MinorCPU counts each instruction type, but this counting is probably not triggered by register accesses.

Regarding [1], I think we can change `O3CPU::getWritableReg()` to record `numVecRegWrites` instead, and remove `setRegOperand()`. Another option is to change `SimpleThread::getWritableRegOperand()` to record `numRegReads[regClass]`, and keep `setRegOperand()`.

I'm not familiar with the implementation details of vector/matrix register to say which one is more realistic, but I'd prefer the latter one. For example, when you update a register with a mask, you probably have to read out the register file, update the unmasked elements, then write it back to the register file.

rogerchang23424:
Hi @hnpl,

Thanks you for elaborate the issue of instruction statistic. I think we should create the new github issue about inconsistent instruction statistics behavior of Simple CPU and O3 CPU.

----

@giactra do you have any concern for this. MatrixReg and VecPredReg only used for ARM instructions currently

hnpl:
I uploaded the patch for this #360, which should resolve the issue blocking this patch.

giactra:
> Hi @hnpl,
> 
> Thanks you for elaborate the issue of instruction statistic. I think we should create the new github issue about inconsistent instruction statistics behavior of Simple CPU and O3 CPU.
> 
> @giactra do you have any concern for this. MatrixReg and VecPredReg only used for ARM instructions currently

Hi all,

I believe #360 is definitely needed for consistency across CPUs... Needless to say handling statistics in the prologue of the instruction execution is not ideal as a fault might arise in the body of the instruction before the dest register gets written. Anyway this "corner" case is probably not interesting from a stats point of view (at least for Arm where FP overflow,underflow etc exception generation is not implemented)

giactra (approved):

hnpl:
Lgtm

hnpl (approved):

PR #343, opened 2023-09-21
sim-se: zero out memory allocated via brk()

### LINKED ISSUE #342: Syscall emulation of brk does not zero-initialize allocated memory
**Describe the bug**
The syscall emulation of the `brk` system call does not zero-initialize the memory that it allocates. In most cases, the newly allocated memory will contain zeroes anyway, but in some cases it contains non-zero values, e.g. when there was a prior call to `sbrk()` with a negative size to deallocate memory.

This appears to cause random crashes in glibc's `free()` for malloc-heavy workloads (like 502.gcc_r in the SPEC CPU2017 benchmarks). It seems that glibc's `malloc`/`free` implementation implictly relies on new memory allocations obtained via `brk()` being zero-initialized. I wasn't able to come up with a small proof-of-concept to demonstrate a malloc/free crash caused by this bug, but you can try compiling 502.gcc_r from the SPEC CPU2017 benchmarks with Clang 16.0.2 if you would like to reproduce a crash.

Here is a simple example directly using the `brk(2)` system call (instead of indirectly via `malloc(3)/free(3)`) to demonstrate how gem5 doesn't zero-initialize program break allocations:
```c
#include <unistd.h>
#include <stdio.h>
#include <string.h>
#define N 32
int main() {
  char *buf = (char *) sbrk(N);
  memset(buf, 0xAB, N);
  sbrk(-N);
  char *buf2 = (char *) sbrk(N);
  for (size_t i = 0; i < N; ++i) {
    printf("%02hhx", buf2[i]);
  }
  printf("\n");
}
```

**Affects version**
develop @ [958eda6](https://github.com/gem5/gem5/commit/958eda6961bc4e2b3b9bb52089c491eaf59353d6)

**gem5 Modifications**
None

**To Reproduce**
1. Compile the above POC program with `cc -static -o brk brk.c`
2. Compile gem5: `scons build/X86/gem5.opt`
3. Run the program on the host: `./brk`
4. Run the program using syscall emulation in gem5: `./build/X86/gem5.opt configs/deprecated/example/se.py --cmd=./brk`

**Terminal Output**
The gem5 SE simulation prints:
```

gem5 version DEVELOP-FOR-23.1
gem5 compiled Sep 19 2023 05:44:26
gem5 started Sep 21 2023 01:15:23
gem5 executing on cafe-cet, pid 3612799
command line: /home/nmosier/gem5-develop/gem5/build/X86/gem5.opt /home/nmosier/gem5-develop/gem5/configs/deprecated/example/se.py --cmd ./brk

warn: The `get_runtime_isa` function is deprecated. Please migrate away from using this function.
warn: The `get_runtime_isa` function is deprecated. Please migrate away from using this function.
**** REAL SIMULATION ****
src/sim/mem_state.cc:443: info: Increasing stack size by one page.
src/sim/syscall_emul.cc:74: warn: ignoring syscall mprotect(...)
abababababababababababababababababababababababababababababababab
Exiting @ tick 34591000 because exiting with last active thread context
```
which is incorrect (specifically, the `ababab...ab`).

**Expected behavior**
The host prints
```
0000000000000000000000000000000000000000000000000000000000000000
```
which is correct.

**Host Operating System**
Ubuntu 22.04

**Host ISA**
X86

**Compiler used**
Clang 16.0.4
**Additional information**
I'll submit a PR shortly.

nmosier:
The syscall emulation of brk() incorrectly did not ensure that newly allocated memory was zero-initialized, which Linux guarantees and which seems to be the expectation of glibc's malloc() and free() implementation. This patch fixes the incorrect behavior by zero- initalizing all memory allocations via brk().

GitHub issue: https://github.com/gem5/gem5/issues/342

Change-Id: I53cf29d6f3f83285c8e813e18c06c2e9a69d7cc2

### DIFF
--- a/src/sim/mem_state.cc
+++ b/src/sim/mem_state.cc
@@ -49,7 +49,7 @@ MemState::MemState(Process *owner, Addr brk_point, Addr stack_base,
       _stackBase(stack_base), _stackSize(max_stack_size),
       _maxStackSize(max_stack_size), _stackMin(stack_base - max_stack_size),
       _nextThreadStackBase(next_thread_stack_base),
-      _mmapEnd(mmap_end), _endBrkPoint(brk_point)
+      _mmapEnd(mmap_end)
 {
 }
 
@@ -67,7 +67,6 @@ MemState::operator=(const MemState &in)
     _stackMin = in._stackMin;
     _nextThreadStackBase = in._nextThreadStackBase;
     _mmapEnd = in._mmapEnd;
-    _endBrkPoint = in._endBrkPoint;
     _vmaList = in._vmaList; /* This assignment does a deep copy. */
 
     return *this;
@@ -108,25 +107,33 @@ void
 MemState::updateBrkRegion(Addr old_brk, Addr new_brk)
 {
     /**
-     * To make this simple, avoid reducing the heap memory area if the
-     * new_brk point is less than the old_brk; this occurs when the heap is
-     * receding because the application has given back memory. The brk point
-     * is still tracked in the MemState class as an independent field so that
-     * it can be returned to the application; we just do not update the
-     * region unless we expand it out.
-     */
-    if (new_brk < old_brk) {
-        _brkPoint = new_brk;
-        return;
-    }
-
-    /**
      * The regions must be page aligned but the break point can be set on
      * byte boundaries. Ensure that the restriction is maintained here by
      * extending the request out to the end of the page. (The roundUp
      * function will not round up an already aligned page.)
      */
-    auto page_aligned_brk = roundUp(new_brk, _pageBytes);
+    auto page_aligned_new_brk = roundUp(new_brk, _pageBytes);
+    auto page_aligned_old_brk = roundUp(old_brk, _pageBytes);
+
+    /**
+     * Reduce the heap memory area if the new_brk point is less than
+     * the old_brk; this occurs when the heap is receding because the
+     * application has given back memory. This may involve unmapping
+     * heap pages, if new_brk rounds to a lower-address page. The
+     * previous behavior was to leave such pages mapped for simplicity;
+     * however, that was not what Linux does in practice and may
+     * violate the assumptions of applications like glibc malloc,
+     * whose default configuration for Linux requires all pages
+     * allocated via brk(2) to be zero-filled (specifically,
+     * by setting MORECORE_CLEARS to 2).
+     */
+    if (new_brk < old_brk) {
+        const auto length = page_aligned_old_brk - page_aligned_new_brk;
+        if (length > 0)
+            unmapRegion(page_aligned_new_brk, length);
+        _brkPoint = new_brk;
+        return;
+    }
 
     /**
      * Create a new mapping for the heap region. We only create a mapping
@@ -135,17 +142,16 @@ MemState::updateBrkRegion(Addr old_brk, Addr new_brk)
      *
      * Since we do not track the type of the region and we also do not
      * coalesce the regions together, we can create a fragmented set of
-     * heap regions. To resolve this, we keep the furthest point ever mapped
-     * by the _endBrkPoint field.
+     * heap regions.
      */
-    if (page_aligned_brk > _endBrkPoint) {
-        auto length = page_aligned_brk - _endBrkPoint;
+    if (page_aligned_new_brk > page_aligned_old_brk) {
+        auto length = page_aligned_new_brk - page_aligned_old_brk;
         /**
          * Check if existing mappings impede the expansion of brk expansion.
          * If brk cannot expand, it must return the original, unmodified brk
          * address and should not modify the mappings here.
          */
-        if (!isUnmapped(_endBrkPoint, length)) {
+        if (!isUnmapped(page_aligned_old_brk, length)) {
             return;
         }
 
@@ -156,8 +162,7 @@ MemState::updateBrkRegion(Addr old_brk, Addr new_brk)
          * implemented if it actually becomes necessary; probably only
          * necessary if the list becomes too long to walk.
          */
-        mapRegion(_endBrkPoint, length, "heap");
-        _endBrkPoint = page_aligned_brk;
+        mapRegion(page_aligned_old_brk, length, "heap");
     }
 
     _brkPoint = new_brk;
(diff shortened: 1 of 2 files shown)


src/sim/syscall_emul.cc
```diff
@@ -270,6 +270,13 @@ brkFunc(SyscallDesc *desc, ThreadContext *tc, VPtr<> new_brk)
 
     mem_state->updateBrkRegion(brk_point, new_brk);
 
+    if (new_brk > brk_point) {
```
giactra: Nit: Could you add a comment here just describing the zeroing?

giactra (changes requested):
That's interesting. I looked at the brk(2) man pages [1][2] and it does not reference any sort of zeroing happening when the program break gets increased. I know this is automatically done when the allocator ask the OS memory with mmap+/dev/zero but I don't see any zeroing being referenced by the linux kernel with the program break.

Could you point me to some documentation describing this behaviour?

[1]: https://man7.org/linux/man-pages/man2/brk.2.html
[2]: https://linux.die.net/man/2/brk

nmosier:
I look into the issue a bit further, and it looks like the real bug is due to the fact that the `gem5::MemState::updateBrkRegion()` method does not unmap deallocated heap pages when the program break recedes. While I can't point to documentation yet showing that pages newly allocated with brk() should be zero-filled, I was able to find exactly where in glibc malloc's source that they encode this assumption: [malloc/malloc.c, line 517](https://sourceware.org/git?p=glibc.git;a=blob;f=malloc/malloc.c;h=d0bbbf371048ee8aa8a30c03b189cb268b8ad9e4;hb=refs/heads/master#l517):
```c
/*  MORECORE_CLEARS           (default 1)                                                                                                                                                                                                     
     The degree to which the routine mapped to MORECORE zeroes out                                                                                                                                                                            
     memory: never (0), only for newly allocated space (1) or always                                                                                                                                                                          
     (2).  The distinction between (1) and (2) is necessary because on                                                                                                                                                                        
     some systems, if the application first decrements and then                                                                                                                                                                               
     increments the break value, the contents of the reallocated space                                                                                                                                                                        
     are unspecified.                                                                                                                                                                                                                         
 */
```
and in [sysdeps/unix/sysv/linux/Makefile](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/Makefile;h=063719bae645b24a4474729640453fa4b5a63056;hb=refs/heads/master) they set the default for Linux to be 2, i.e., memory returned by MORECORE (which defaults to sbrk()) is always zero-filled:
```
ifeq ($(subdir),malloc)
CFLAGS-malloc.c += -DMORECORE_CLEARS=2
endif
```

What I had before, zero-filling newly allocated brk() memory directly, would also work, but the Linux kernel doesn't appear to do that. It seems to only unmap freed pages when the program break recedes.

nmosier:
I'm wondering if the test that failed is flaky, since all tests passed locally on my machine.

powerjg:
Yeah, we're having intermittent network issues with some of our runners where the test binaries are failing to download. Hopefully it will work when we re-run it.

giactra:
> I look into the issue a bit further, and it looks like the real bug is due to the fact that the `gem5::MemState::updateBrkRegion()` method does not unmap deallocated heap pages when the program break recedes. While I can't point to documentation yet showing that pages newly allocated with brk() should be zero-filled, I was able to find exactly where in glibc malloc's source that they encode this assumption: [malloc/malloc.c, line 517](https://sourceware.org/git?p=glibc.git;a=blob;f=malloc/malloc.c;h=d0bbbf371048ee8aa8a30c03b189cb268b8ad9e4;hb=refs/heads/master#l517):
> 
> ```c
> /*  MORECORE_CLEARS           (default 1)                                                                                                                                                                                                     
>      The degree to which the routine mapped to MORECORE zeroes out                                                                                                                                                                            
>      memory: never (0), only for newly allocated space (1) or always                                                                                                                                                                          
>      (2).  The distinction between (1) and (2) is necessary because on                                                                                                                                                                        
>      some systems, if the application first decrements and then                                                                                                                                                                               
>      increments the break value, the contents of the reallocated space                                                                                                                                                                        
>      are unspecified.                                                                                                                                                                                                                         
>  */
> ```
> 
> and in [sysdeps/unix/sysv/linux/Makefile](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/Makefile;h=063719bae645b24a4474729640453fa4b5a63056;hb=refs/heads/master) they set the default for Linux to be 2, i.e., memory returned by MORECORE (which defaults to sbrk()) is always zero-filled:
> 
> ```
> ifeq ($(subdir),malloc)
> CFLAGS-malloc.c += -DMORECORE_CLEARS=2
> endif
> ```
> 
> What I had before, zero-filling newly allocated brk() memory directly, would also work, but the Linux kernel doesn't appear to do that. It seems to only unmap freed pages when the program break recedes.

Thanks @nmosier for taking the time to look into the documentation. Pointing to the glibc implementation really helps on clearing out some doubts!

src/sim/mem_state.cc
```diff
@@ -107,26 +106,34 @@ MemState::isUnmapped(Addr start_addr, Addr length)
...
+     * Reduce the heap memory area if the new_brk point is less than
+     * the * old_brk; this occurs when the heap is receding because the
+     * application has given back memory. This may involve unmapping
+     * heap pages, if new_brk rounds to a lower-address page. The
+     * previous, buggy behavior was to leave such pages mapped for
+     * simplicity. However, this introduced a bug where, if the heap
+     * was later expanded out, this routine reuses previously mapped
+     * heap pages that were not zero-filled (and instead were filled
+     * with stale application data), violating the assumptions of
+     * glibc malloc.
```
giactra: Nit: I would reword this part of the comment as IMHO it's not entirely accurate to label previous behaviour as intrinsically buggy.

From what I understood not zeroing out the old heap is technically legal and it does not violate any glibc assumption as long as glibc is informed by such behaviour via the MORECORE_CLEAR value. We should just note that Linux expect the zeroing and unmapping pages anyway regardless of the MORECORE_CLEAR is the safest option that works for every system.

giactra (changes requested):
Code looks great I just have a minor comment

powerjg (approved):
lgtm! Thanks for the fix

giactra (approved):

PR #351, opened 2023-09-22
arch: Enable customized decoder class name

rogerchang23424:
Developers can make the own ISADesc action in the SConscript with their decoder class name.

Change-Id: I011cf059642e178913e1f62df4e5c02401cc132e


giactra (approved):
LGTM

PR #352, opened 2023-09-23
WIP: GPUFS stdlib board

abmerop:
Still working on this but stuck on this error:

`Error in unproxying param 'memories' of board`

Hoping y'all stlib experts can help me out here because I don't see anything called 'memories' anywhere.  Maybe part of what would be the 'system' object in legacy configs/?.  Using the following:

`cat viper.py`
```from src.python.gem5.prebuilt.viper.board import X86ViperBoard


# Here we setup the board. The prebuilt X86DemoBoard allows for Full-System X86
# simulation.
board = X86ViperBoard("Vega10")

# We then set the workload. Here we use the "x86-ubuntu-18.04-boot" workload.
# This boots Ubuntu 18.04 with Linux 5.4.49. If the required resources are not
# found locally, they will be downloaded.
board.set_workload(obtain_resource("x86-ubuntu-18.04-boot"))

simulator.run()```

And then:

`build/VEGA_X86/gem5.opt viper.py`


abmerop:
No feedback in about a year and #1349 would require a different solution

PR #353, opened 2023-09-24
cpu: Fetch Directed Instruction Prefetching (FDIP)

bgodala:
Implementing a decoupled frontend aka FDIP. This patch is first of many optimizations to improve performance of CPU front-end. This patch adds a Fetch Target Queue (FTQ) which is used by prefetcher to issue prefetch requests. This version works only with ARM ISA.

Change-Id: Ibcea9e32edc471f7e61c9b5a8f519fab2ebd2fa5

### DIFF
--- a/configs/common/Options.py
+++ b/configs/common/Options.py
@@ -190,6 +190,7 @@ def addNoISAOptions(parser):
     parser.add_argument("--l2_assoc", type=int, default=8)
     parser.add_argument("--l3_assoc", type=int, default=16)
     parser.add_argument("--cacheline_size", type=int, default=64)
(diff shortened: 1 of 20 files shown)


mattsinc (commented):
Just a question more than a review: is there a fundamental reason why this only works with the ARM ISA?  Looking quickly through your changes, it does not appear to make any changes (except maybe to PCState?) that only work for ARM.  And it would be great if we can find a way to generalize this to work for all ISAs given the importance of FDIP to modern OOO CPU design.

(not requesting that for this patch, just more of a general commentary)

PR #355, opened 2023-09-24
cpu: Fetch Directed Instruction Prefetching (FDIP)

bgodala:
Implementing a decoupled frontend aka FDIP. This patch is first of many optimizations to improve performance of CPU front-end. This patch adds a Fetch Target Queue (FTQ) which is used by prefetcher to issue prefetch requests. This version works only with ARM ISA.

Change-Id: Ibcea9e32edc471f7e61c9b5a8f519fab2ebd2fa5

### DIFF
--- a/configs/common/Caches.py
+++ b/configs/common/Caches.py
@@ -53,8 +53,8 @@ class L1Cache(Cache):
     assoc = 2
     tag_latency = 2
     data_latency = 2
-    response_latency = 2
-    mshrs = 4
+    response_latency = 0
+    mshrs = 64
     tgts_per_mshr = 20
 
 
@@ -62,6 +62,7 @@ class L1_ICache(L1Cache):
     is_read_only = True
     # Writeback clean lines as well
     writeback_clean = True
+    # write_buffers = 32
 
 
 class L1_DCache(L1Cache):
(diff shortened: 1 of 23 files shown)


mattsinc (commented):
(reposting here since other pull request was closed)



Just a question more than a review: is there a fundamental reason why this only works with the ARM ISA? Looking quickly through your changes, it does not appear to make any changes (except maybe to PCState?) that only work for ARM. And it would be great if we can find a way to generalize this to work for all ISAs given the importance of FDIP to modern OOO CPU design.

(not requesting that for this patch, just more of a general commentary)

andysan:
@mattsinc I think the main problem is the fact that x86 is variable length. I know that @dhschall spent a lot of time dealing with that when he implemented his FDIP patches (which support Arm, x86) that he is in the process of posting. Since there are now two changes adding support for FDIP, I think we need to carefully review both changesets for features and decide which one to merge based on feature completeness. There is a draft PR with his FDIP changes here #359.

bgodala:
@mattsinc As @andysan pointed out variable length makes things tricky. Another problem is that in the current implementation, every branch that was in the pipeline must be present in the branch history of branch predictor. If a new branch is discovered later in the pipeline then would cause problems in BPU . In case of ARM, I avoid this problem by pre-decoding branch instructions in fetch stage (inserted in BTB speculatively, also another problem) and reset FDIP related state. The same can be done for X86 but it may not be fair assumption. There are several other issues with X86 like micro branches (branching to another microop within the same instructions’s microop stream), exception handling, in syscalls somehow branch instruction was not the last instruction which again breaks many things. I planned to push my changes in phases. In the next phase I plan to add support to handle newly discovered branches in the pipeline and also prefect recovery of branch predictor state after squash. After that I can add X86 support.

mattsinc:
@Harshil2107 I know almost nothing about the gem5 CPU models and should not have been assigned as a reviewer for this.  I am certainly curious in the model, hence my question, but that should not be mistaken for expertise.

erin-le:
@bgodala Hello, this PR is being closed due to inactivity. If you would like to continue working on it, please feel free to reopen it. Thank you!

PR #356, opened 2023-09-25
sim: Probe listener template with lambda

dhschall:
Adds a new probe listener template which can be used to instantiate with a lambda function that is called by notify(). It is similar to ProbeListenerArg with class but provides more flexibility. I.e. the can be another object than the one instantiating the lambda which allows to listen to any object. Furthermore additional parameters can be passed in easily.

Change-Id: Iba451357182caf25097b9ae201cd5c647aff3a4f


andysan (approved):
LGTM

powerjg (approved):
This seems quite useful! @kunpai, we should make sure to note this in the release notes.

PR #357, opened 2023-09-25
arch: Add instruction size and PC set methods

dhschall:
Add the instruction size of a static instruction. x86 and arm decoders add now the instruction size to the macro instruction. However, microops are still handled by the fetch stage which is not nice.
Furthermore, we add a set method to the PC state. It allows setting a PC state to acertain address.
Both methods are required for the decoupled front-end.

Change-Id: I311fe3f637e867c42dee7781f5373ea2e69e2072

### DIFF
--- a/src/arch/generic/pcstate.hh
+++ b/src/arch/generic/pcstate.hh
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2020 ARM Limited
+ * Copyright (c) 2023 The University of Edinburgh
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
@@ -123,6 +124,13 @@ class PCStateBase : public Serializable
     virtual void
     uReset()
     {
+        _upc = 0;
+    }
+
+    virtual void
+    set(Addr val)
+    {
+        _pc = val;
         _upc = 0;
     }
 
@@ -310,6 +318,14 @@ class PCStateWithNext : public PCStateBase
     }
 
     void
+    set(Addr val) override
+    {
+        PCStateBase::set(val);
+        _npc = 0;
+        _nupc = 1;
+    }
+
+    void
     serialize(CheckpointOut &cp) const override
     {
         PCStateBase::serialize(cp);
@@ -359,9 +375,9 @@ class SimplePCState : public PCStateWithNext
      * @param val The value to set the PC to.
      */
     void
-    set(Addr val)
+    set(Addr val) override
     {
-        this->pc(val);
+        Base::set(val);
         this->npc(val + InstWidth);
     };
 
@@ -402,7 +418,7 @@ class UPCState : public SimplePCState<InstWidth>
     }
 
     void
-    set(Addr val)
+    set(Addr val) override
     {
         Base::set(val);
         this->upc(0);
--- a/src/cpu/static_inst.hh
+++ b/src/cpu/static_inst.hh
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2017, 2020 ARM Limited
+ * Copyright (c) 2022-2023 The University of Edinburgh
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
@@ -251,6 +252,11 @@ class StaticInst : public RefCounted, public StaticInstFlags
     }
 
     /**
+     * Instruction size in bytes. Necessary for dynamic instruction sizes
+     */
+    size_t _size = 0;
+
+    /**
      * Base mnemonic (e.g., "add").  Used by generateDisassembly()
      * methods.  Also useful to readily identify instructions from
      * within the debugger when #cachedDisassembly has not been
@@ -306,6 +312,17 @@ class StaticInst : public RefCounted, public StaticInstFlags
     {
         panic("buildRetPC not defined!");
     }
+
+    size_t size() const
+    {
+        if (_size == 0) fatal(
+            "Instruction size for this instruction not set! It's size is "
+            "required for the decoupled front-end. Either use the standard "
+            "front-end or this ISA needs to be extended with the instruction "
+            "size. Refer to the X86, Arm or RiscV decoders for an example.");
+        return _size;
+    }
+    virtual void size(size_t newSize) { _size = newSize; }
 
     /**
      * Return the microop that goes with a particular micropc. This should
--- a/src/arch/arm/decoder.hh
+++ b/src/arch/arm/decoder.hh
@@ -138,6 +138,7 @@ class Decoder : public InstDecoder
         StaticInstPtr si = defaultCache.decode(this, mach_inst, addr);
         DPRINTF(Decode, "Decode: Decoded %s instruction: %#x\n",
                 si->getName(), mach_inst);
+        si->size((!emi.thumb || emi.bigThumb) ? 4 : 2);
         return si;
     }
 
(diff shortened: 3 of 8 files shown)


powerjg (commented):
Thanks for this change! Generally, it makes sense to me to have the size of the instruction in the static inst.

Two questions:
1. Why not set the size in the `decode` function? Why do we need to have a separate function call to set the size?
2. Is this going to have any impact on other ISAs? Specifically, if SPARC does not implement the `size` function and the size of all instructions is the default (0), will anything break?

dhschall:
Thanks, @powerjg for looking into that.

First of all, I have to say I am myself not very happy with the way it's implemented. It was more of a hack as I need the size for the decoupled front-end.
In particular I don't like the way I pass the size from the macroops to the microops (The changes in `fetch.cc`).

I hope someone can give me feedback on the idiomatic way I should do it.

1. To make sure I understand correct. I.e. in for RISC-V you mean moving the set function a few lines up into
`Decoder::decode(ExtMachInst mach_inst, Addr addr)` starting at 678? If so that's a good idea, I can do that.
2. Good point as well. I only tested the decoupled front end with those three ISA's that's why I only changed it there. However, I'll do it for the other ISA's as well. Therefore, question back: If I understand the code correctly have all other ISA's fixed size instructions of size 4 (All implement some kind of `...PCState<4>`). Is that correct?

powerjg:
> Thanks, @powerjg for looking into that.
> 
> First of all, I have to say I am myself not very happy with the way it's implemented. It was more of a hack as I need the size for the decoupled front-end. In particular I don't like the way I pass the size from the macroops to the microops (The changes in `fetch.cc`).

I *think* you can do this in `decode` as well. I believe all of the micro-ops are generated when the macro-op `StaticInst` is created. You may have to add an optional `size` parameter and change the decoders a bit more.

> 
> I hope someone can give me feedback on the idiomatic way I should do it.
> 
> 1. To make sure I understand correct. I.e. in for RISC-V you mean moving the set function a few lines up into
>    `Decoder::decode(ExtMachInst mach_inst, Addr addr)` starting at 678? If so that's a good idea, I can do that.

That was exactly what I was thinking. 👍 

> 2. Good point as well. I only tested the decoupled front end with those three ISA's that's why I only changed it there. However, I'll do it for the other ISA's as well. Therefore, question back: If I understand the code correctly have all other ISA's fixed size instructions of size 4 (All implement some kind of `...PCState<4>`). Is that correct?

I don't have enough confidence to say that's right 😄. Personally, I suggest keeping the size 0 for the instructions if we're not confident in what the size should be. Then, we just need to make sure there isn't a terrible failure condition when the size is 0 (e.g., we should `fatal` and say that FDIP isn't compatible with an ISA that doesn't implement the size).

dhschall:
Ok I fixed now 1. and 2. according your suggestion.

The micro-ops bit I need to check. I guess I need a bit a better understanding.

src/cpu/o3/fetch.cc
```diff
@@ -1245,6 +1245,14 @@ Fetch::fetch(bool &status_change)
                     staticInst = curMacroop->fetchMicroop(this_pc.microPC());
                 }
                 newMacro |= staticInst->isLastMicroop();
+
+                /// TODO: Make it proper
+                // This is a hack to make sure that the size of the macroop
+                // is propergated to the microops. This is needed for
+                // for the branch predictor to compute the return address
+                // in the decoupled frontend.
+                if (curMacroop)
```
andysan: This seems slightly ugly since it would mean that adding up the size of the micro ops within a macro op would result in a size that is larger than the macro op. I guess we could use a convention where the macro op has a size and the micro ops reflect that size (which is what you are implementing), but that seems like it might cause issues down the line.

Is there any chance we could track this in the branch predictor instead? And should we even need perform predictions on micro ops?

dhschall: I am open to any other suggestions on how to handle that.
The problem is the following: You have a branch instruction that gets broken into let's say the address calc and the actual jump microop. (All have the same instruction address but different microPCs) The BPU will be queried only for the jump micro-op.
When the prediction is fall-through the BPU computes the next InstAddr. For this, it needs the size of the overall macroop.
At the moment this is done by advancing the PCState. This state knows when and where the instruction ends.
For a decoupled fetch PC this is not available as you search instruction addresses until you hit in the BTB and you don't have a valid PC state.

This is easy for RISC-V or ARM but we also want to support x86

dhschall: IMHO, reflecting the instruction size in each micro-op is not wrong since they are just conceptual and have no size. They are microarchitecture-dependent

powerjg: I like the solution @dhschall proposes here. It's not ideal, but I don't think there are any other solutions that I can see.

powerjg: Oops, spoke too soon. It would still be nice if this was in the constructor of the micro-op instead of having to be replicated in all CPU models.

src/cpu/static_inst.hh
```diff
@@ -307,6 +313,13 @@ class StaticInst : public RefCounted, public StaticInstFlags
         panic("buildRetPC not defined!");
     }
 
+    size_t size() const
+    {
+      if (_size == 0) panic("Instruction size not set!");
+      return _size;
```
powerjg: Two more spaces :)

```suggestion
        if (_size == 0) panic("Instruction size not set!");
        return _size;
```

src/cpu/static_inst.hh
```diff
@@ -307,6 +313,13 @@ class StaticInst : public RefCounted, public StaticInstFlags
         panic("buildRetPC not defined!");
     }
 
+    size_t size() const
+    {
+      if (_size == 0) panic("Instruction size not set!");
```
powerjg: This could use a more detailed error message. Maybe "Instruction size for this ISA is not set for branch predictor that needs instruction size. Either use a different branch predictor or this ISA needs to be extended with the instruction size."

Also, I think this should be a `fatal` as it is a configuration error, not a bug.

src/arch/generic/pcstate.hh
```diff
@@ -126,6 +127,13 @@ class PCStateBase : public Serializable
         _upc = 0;
     }
 
+    virtual void
```
powerjg, line 130: Can you explain the need for this change? (I think I'm missing something)

src/cpu/o3/fetch.cc
andysan: That sounds like a good reason to keep the size in the micro-op. It's still slightly weird for micro-code sequences where the with a branch that isn't the last micro-op. I don't think that's very common though, so we probably don't need to worry about that at the moment.

I agree with Jason that this should be handled in the constructor rather than the CPU model unless there is a really good reason.

dhschall: Thanks for the suggestions. I think I found a way to easily pass the instruction size of all micro-ops.
Let me know if this is ok.

src/arch/generic/pcstate.hh
dhschall: For the decoupled front-end I need a method to set the PC state to the start of a specific instruction address.
This should set the PC to the address passed in and reset the uop pc to 0.
Same is for `PCStateWithNext`. Here, I reset the next up pc to its initial value 1 (line 262)

src/cpu/o3/fetch.cc
andysan: Thanks, this looks like a better solution to me.

powerjg (approved):
This looks good to me! Thanks for all of the updates.

andysan (approved):

dhschall:
> This looks good to me! Thanks for all of the updates.

I have to say thank you @powerjg and @andysan for all the suggestions and help. Now I also like the solution much more

PR #362, opened 2023-09-26
misc: Add Clang format to pre-commit

trollet91:
In this MR you will find:
- clang-format based on Jason work with option only up to clang-format v10 (latest supported in ubuntu20).
- fix for the BitUnion macro not being properly indented. The syntaxt in the clang format was wrong and we have to change EndBitUnion to BitUnionEnd because of how clang-format match strings.
-  WIP commit with how we could  fix the include/forward declatarion automatically. This needs more work.

Discussion in #333


.clang-format
```diff
@@ -0,0 +1,456 @@
...
+AlignConsecutiveMacros: false
+
+# AlignEscapedNewlines (EscapedNewlineAlignmentStyle) clang-format 5
+AlignEscapedNewlines: DontAlign
+
+# AlignOperands (Boolean) clang-format 3.5
+AlignOperands:   true
+
+# AlignTrailingComments (TrailingCommentsAlignmentStyle) clang-format 3.7
+AlignTrailingComments: false
```
AntonTimofeev:
```suggestion
AlignTrailingComments: true
```

.clang-format
```diff
@@ -0,0 +1,456 @@
...
+# AlignConsecutiveBitFields (AlignConsecutiveStyle) clang-format 11
+
+# AlignConsecutiveDeclarations (Boolean) clang-format 3.8
+AlignConsecutiveDeclarations: false
+
+# AlignConsecutiveMacros (Boolean) clang-format 9
+AlignConsecutiveMacros: false
+
+# AlignEscapedNewlines (EscapedNewlineAlignmentStyle) clang-format 5
+AlignEscapedNewlines: DontAlign
```
AntonTimofeev:
```suggestion
AlignEscapedNewlines: Left
```

.clang-format
```diff
@@ -0,0 +1,456 @@
...
+# AllowShortBlocksOnASingleLine (ShortBlockStyle) clang-format 3.5
+AllowShortBlocksOnASingleLine: Empty
+
+# AllowShortCaseLabelsOnASingleLine (Boolean) clang-format 3.6
+AllowShortCaseLabelsOnASingleLine: false
+
+# AllowShortEnumsOnASingleLine (Boolean) clang-format 11
+
+# AllowShortFunctionsOnASingleLine (ShortFunctionStyle) clang-format 3.5
+AllowShortFunctionsOnASingleLine: None
```
AntonTimofeev:
```suggestion
AllowShortFunctionsOnASingleLine: InlineOnly
```

.clang-format
```diff
@@ -0,0 +1,456 @@
...
+AllowShortLoopsOnASingleLine: false
+
+# AlwaysBreakAfterDefinitionReturnType (DefinitionReturnTypeBreakingStyle) clang-format 3.7
+# Deprecated
+
+# AlwaysBreakAfterReturnType (ReturnTypeBreakingStyle) clang-format 3.8
+AlwaysBreakAfterReturnType: AllDefinitions
+
+# AlwaysBreakBeforeMultilineStrings (Boolean) clang-format 3.4
+AlwaysBreakBeforeMultilineStrings: false
```
AntonTimofeev, line 72:
```suggestion
AlwaysBreakBeforeMultilineStrings: true
```

.clang-format
```diff
@@ -0,0 +1,456 @@
...
+# BinPackParameters (Boolean) clang-format 3.7
+BinPackParameters: true
+
+# BitFieldColonSpacing (BitFieldColonSpacingStyle) clang-format 12
+
+# BraceWrapping (BraceWrappingFlags) clang-format 3.8
+BraceWrapping:
+  AfterCaseLabel:  false
+  AfterClass:      true
+  AfterControlStatement: Never
```
AntonTimofeev:
```suggestion
  AfterControlStatement: MultiLine
```

.clang-format
```diff
@@ -0,0 +1,456 @@
...
+CompactNamespaces: false
+
+# ConstructorInitializerAllOnOneLineOrOnePerLine (Boolean) clang-format 3.7
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+
+# ConstructorInitializerIndentWidth (Unsigned) clang-format 3.7
+ConstructorInitializerIndentWidth: 4
+
+# ContinuationIndentWidth (Unsigned) clang-format 3.7
+ContinuationIndentWidth: 4
```
AntonTimofeev, line 153:
```suggestion
ContinuationIndentWidth: 8
```

.clang-format
```diff
@@ -0,0 +1,456 @@
...
+# ObjCBreakBeforeNestedBlockParam (Boolean) clang-format 11
+
+# ObjCSpaceAfterProperty (Boolean) clang-format 3.7
+
+# ObjCSpaceBeforeProtocolList (Boolean) clang-format 3.7
+
+# PPIndentWidth (Integer) clang-format 13
+
+# PackConstructorInitializers (PackConstructorInitializersStyle) clang-format 14
+
```
AntonTimofeev, line 281:
```suggestion
PackConstructorInitializers: Never

```

.clang-format
```diff
@@ -0,0 +1,456 @@
...
+# IndentCaseLabels (Boolean) clang-format 3.3
+IndentCaseLabels: false
+
+# IndentExternBlock (IndentExternBlockStyle) clang-format 11
+
+# IndentGotoLabels (Boolean) clang-format 10
+IndentGotoLabels: true
+
+# IndentPPDirectives (PPDirectiveIndentStyle) clang-format 6
+IndentPPDirectives: AfterHash
```
AntonTimofeev:
```suggestion
IndentPPDirectives: None
```

.clang-format
```diff
@@ -0,0 +1,456 @@
...
+
+# IndentAccessModifiers (Boolean) clang-format 13
+
+# IndentCaseBlocks (Boolean) clang-format 11
+
+# IndentCaseLabels (Boolean) clang-format 3.3
+IndentCaseLabels: false
+
+# IndentExternBlock (IndentExternBlockStyle) clang-format 11
+
```
AntonTimofeev, line 211:
```suggestion
IndentExternBlock: NoIndent

```

.clang-format
```diff
@@ -0,0 +1,456 @@
...
+# SpaceBeforeAssignmentOperators (Boolean) clang-format 3.7
+SpaceBeforeAssignmentOperators: true
+
+# SpaceBeforeCaseColon (Boolean) clang-format 12
+
+# SpaceBeforeCpp11BracedList (Boolean) clang-format 7
+SpaceBeforeCpp11BracedList: false
+
+# SpaceBeforeCtorInitializerColon (Boolean) clang-format 7
+SpaceBeforeCtorInitializerColon: true
```
AntonTimofeev, line 389:
```suggestion
SpaceBeforeCtorInitializerColon: false
```

.clang-format
```diff
@@ -0,0 +1,456 @@
...
+# Standard (LanguageStandard) clang-format 3.7
+Standard:        Auto
+
+# StatementAttributeLikeMacros (List of Strings) clang-format 12
+
+# StatementMacros (List of Strings) clang-format 8
+StatementMacros:
+
+# TabWidth (Unsigned) clang-format 3.7
+TabWidth:        8
```
AntonTimofeev:
```suggestion
TabWidth:        4
```

AntonTimofeev (commented):
This looks better for me

trollet91:
@AntonTimofeev
We cannot choose format options like we prefer. As discussed before, maintainers want the new config to match as close as possible the current format of the gem5 codebase. The goal is to minimize the impact of adding .clang-format.
We also cannot use options more recent than clang10 (newest supported in ubuntu20).

AntonTimofeev:
Some of suggestions are taken from: https://www.gem5.org/documentation/general_docs/development/coding_style/

AlignTrailingComments
```
    int _fooVariable;   // starts with '_' because it has public accessor functions
    int barVariable;    // no '_' since it's internal use only
```

AllowShortFunctionsOnASingleLine
```
    // short inline methods can go all on one line
    int fooVariable() const { return _fooVariable; }
```

AfterControlStatement
```
// exception case
for (...;
     ...;
     ...) // brace could be up here
{ // but this is optionally OK *only* when the 'for' spans multiple lines
    ...
}
```

I'll re-check if suggested parameters are supported by clang10.

.clang-format
```diff
@@ -0,0 +1,456 @@
...
+
+# KeepEmptyLinesAtTheStartOfBlocks (Boolean) clang-format 3.7
+KeepEmptyLinesAtTheStartOfBlocks: false
+
+# LambdaBodyIndentation (LambdaBodyIndentationKind) clang-format 13
+
+# LineEnding (LineEndingStyle) clang-format 16
+
+# MacroBlockBegin (String) clang-format 3.7
+MacroBlockBegin: "^\
```
andysan: I suspect these regular expressions aren't matching what you expect due to operator precedence. That's probably why you had to rename EndBitUnion. I got this to work as expected:

```
MacroBlockBegin: "^(\
BitUnion(8|16|32|64)|\
SubBitUnion\
)$"

MacroBlockEnd: "^(\
EndBitUnion|\
EndSubBitUnion\
)$"
```
With the change above, you should be able to avoid renaming EndBitUnion.

src/arch/amdgpu/vega/pagetable.hh
```diff
@@ -63,7 +63,7 @@ BitUnion64(PageTableEntry)
     Bitfield<2> c;
     Bitfield<1> s;
     Bitfield<0> v;
-EndBitUnion(PageTableEntry)
+ButUnionEnd(PageTableEntry)
```
odanrc: The sed cmd had a typo

trollet91: Andreas Sandberg has found a mistake in the clang format. The sed is not needed anymore. 
Thanks anyway!

.clang-format
trollet91: Fixed, thanks!

trollet91:
The next steps are:
- Run clang-format v10 and check that nothing breaks (ex: src/systemc/tests/systemc/compliance_1666/test001/test001.cpp seems to be broken after running clang-format)
- Fix incompatibilities between clang-format and gem5 style checker during pre-commit hooks
- Commit all formated changes

@BobbyRBruce I will not have time to push this forward anymore this week, feel free to take ownership of this patch.

BobbyRBruce:
> The next steps are:
> 
> * Run clang-format v10 and check that nothing breaks (ex: src/systemc/tests/systemc/compliance_1666/test001/test001.cpp seems to be broken after running clang-format)
> * Fix incompatibilities between clang-format and gem5 style checker during pre-commit hooks
> * Commit all formated changes
> 
> @BobbyRBruce I will not have time to push this forward anymore this week, feel free to take ownership of this patch.

I was probably being too optimistic/demanding to say we want this in that quickly. As long as this is being actively worked on to a reasonable extent, I'm happy.

Am I correct in thinking we can get rid of the gem5 style checker pre-commit hook when we add clang-format? I would very much like to. If there's an irreconcilable conflict I'd go with clang-format over the hook.

powerjg:
> I was probably being too optimistic/demanding to say we want this in that quickly. As long as this is being actively worked on to a reasonable extent, I'm happy.
> 
> Am I correct in thinking we can get rid of the gem5 style checker pre-commit hook when we add clang-format? I would very much like to. If there's an irreconcilable conflict I'd go with clang-format over the hook.

I don't think you were being too optimistic. I think this is really just a few hours of work. The question is who will make the time 😄.

I also think we shouldn't worry about the gem5 style checker vs clang-format. Let's drop the style-checker.

andysan:
> > The next steps are:
> 
> > 
> 
> > * Run clang-format v10 and check that nothing breaks (ex: src/systemc/tests/systemc/compliance_1666/test001/test001.cpp seems to be broken after running clang-format)
> 
> > * Fix incompatibilities between clang-format and gem5 style checker during pre-commit hooks
> 
> > * Commit all formated changes
> 
> > 
> 
> > @BobbyRBruce I will not have time to push this forward anymore this week, feel free to take ownership of this patch.
> 
> 
> 
> I was probably being too optimistic/demanding to say we want this in that quickly. As long as this is being actively worked on to a reasonable extent, I'm happy.
> 
> 
> 
> Am I correct in thinking we can get rid of the gem5 style checker pre-commit hook when we add clang-format? I would very much like to. If there's an irreconcilable conflict I'd go with clang-format over the hook.

I think the only missing bit might be include block order. It's be nice to keep that, especially since some includes (IIRC, Python) are sensitive to order. It's also desirable to keep the "primary" (hh corresponding to the current cc) header first to ensure the header is independent. 

IIUC, this should be possible in clang-format.

powerjg:
We are planning to merge this once we get most of the PRs listed in #558 merged. So, anyone that has an open PR right now that is not in the list for #558, you should expect some headaches in merging/rebasing after we create the staging branch for 23.1 (i.e., when you rebase to be included in develop-for-24.0).

andysan:
> We are planning to merge this once we get most of the PRs listed in #558 merged. So, anyone that has an open PR right now that is not in the list for #558, you should expect some headaches in merging/rebasing after we create the staging branch for 23.1 (i.e., when you rebase to be included in develop-for-24.0).

I'm supportive of merging this one sooner rather than later, but could we please create a PR with the updated configuration that includes the updates to the codebase as well? The isort PR had some weird unexpected behaviour that I caught when reviewing the changes it made to the codebase and I wouldn't be surprised if the same happens here. Setting a merge deadline Monday/Tuesday next week seems reasonable.

trollet91:
As said in my message from October 13, this PR needs more work.
Sadly I dont have time to work on gem upstream right now.
If we want to merge this soon, someone needs to take the ownership of this PR and fix the formating issues.

powerjg:
Thanks @trollet91 for pushing on this! @BobbyRBruce is taking it over (apologies for not making that clear in this thread).

@andysan, we'll have 2 weeks of testing the staging branch to work out any kinks. Monday is a good target for merging this after we merge the other changes for the staging branch. One goal is to make sure that we are the least disruptive to current PRs as we can be. I think that doing it "atomically" between when the staging branch is *ready to be* created and when it's created is the best way.

andysan:
> Thanks @trollet91 for pushing on this! @BobbyRBruce is taking it over (apologies for not making that clear in this thread).
> 
> @andysan, we'll have 2 weeks of testing the staging branch to work out any kinks. Monday is a good target for merging this after we merge the other changes for the staging branch. One goal is to make sure that we are the least disruptive to current PRs as we can be. I think that doing it "atomically" between when the staging branch is _ready to be_ created and when it's created is the best way.

OK, that sounds good. The quiet period while the staging branch stabilises sounds like the ideal time to merge this. Depending on how intrusive this is, we should consider pushing out the merge a few more days.

Looking forward to this landing in the code base! :)

BobbyRBruce:
I've taken this change and applied it to the codebase. It obviously touches a lot of files and is a significant change to the codebase.

include/gem5/m5ops.h
```diff
@@ -30,50 +30,51 @@
 #define __GEM5_M5OP_H__
 
 #ifdef __cplusplus
-extern "C" {
+extern "C"
```
andysan: This is probably not a common problem, but I suspect we don't want to indent in these cases. I consider this to be similar to namespaces. "IndentExternBlock: false" would fix this.

src/arch/amdgpu/common/gpu_translation_state.hh
```diff
@@ -86,19 +85,30 @@ struct GpuTranslationState : public Packet::SenderState
...
 
-    GpuTranslationState(BaseMMU::Mode tlb_mode,
-                       bool _prefetch=false,
-                       Packet::SenderState *_saved=nullptr)
-        : tlbMode(tlb_mode), tc(nullptr), deviceId(0), pasId(0),
-          tlbEntry(nullptr), isPrefetch(_prefetch), issueTime(0), hitLevel(0),
-          saved(_saved)
-    { }
+    GpuTranslationState(BaseMMU::Mode tlb_mode, bool _prefetch = false,
+        Packet::SenderState *_saved = nullptr) :
```
andysan: This is diverts from the current style guide. The colon should be on the next line together with the first member. That said, I think it makes sense like this since it makes diffs smaller if you add some member variables. Setting "BreakConstructorInitializers: BeforeColon" would fix this.

include/gem5/asm/generic/m5ops.h
```diff
@@ -41,73 +41,72 @@
...
+#define M5OP_WORK_END 0x5b
 
-#define M5OP_DIST_TOGGLE_SYNC   0x62
+#define M5OP_DIST_TOGGLE_SYNC 0x62
 
-#define M5OP_WORKLOAD           0x70
+#define M5OP_WORKLOAD 0x70
 
-
-#define M5OP_FOREACH                                            \
```
andysan, line 84: Didn't we use to align backslash in these cases? Is this something we want to keep? "AlignEscapedNewlines: Right" would fix this.

src/arch/arm/aapcs32.hh
```diff
@@ -96,14 +96,16 @@ constexpr bool IsAapcs32CompositeV = IsAapcs32Composite<T>::value;
...
 // we approximate that by detecting only arrays with that property.
 
-template <typename T, std::size_t count, typename Enabled=void>
+template <typename T, std::size_t count, typename Enabled = void>
 using Aapcs32HomogeneousAggregate = T[count];
 
 template <typename T>
-struct IsAapcs32HomogeneousAggregate : public std::false_type {};
+struct IsAapcs32HomogeneousAggregate : public std::false_type
+{};
```
andysan: This looks a bit weird. I think we used to have the curlies on on separate lines in these cases.

src/arch/arm/fastmodel/iris/thread_context.cc
```diff
@@ -659,46 +656,46 @@ ThreadContext::getReg(const RegId &reg, void *val) const
     const RegClassType type = reg.classValue();
     if (flat) {
         switch (type) {
-          case IntRegClass:
+        case IntRegClass:
```
andysan, line 672: This is inconsistent with the current style guide, but I don't think there is any way to make clang-format adhere to the half indentation style we use. We either have to indent a full step or not at all, so I think this is the right compromise.

src/arch/arm/fastmodel/iris/thread_context.hh
```diff
@@ -124,9 +120,21 @@ class ThreadContext : public gem5::ThreadContext
...
 
-        bool empty() const { return events->empty(); }
-        bool validIds() const { return !ids.empty(); }
-        void clearIds() { ids.clear(); }
+        bool
+        empty() const
+        {
+            return events->empty();
+        }
+        bool
```
andysan, line 132: I would prefer a newline between function implementations. Not sure what the current style guide says though.

src/arch/arm/freebsd/freebsd.hh
```diff
@@ -218,28 +216,28 @@ class ArmFreebsd64 : public ArmFreebsd, public OpenFlagTable<ArmFreebsd64>
...
-    static constexpr int TGT_O_CREAT     = 0x00000200; //!< O_CREAT
-    static constexpr int TGT_O_EXCL      = 0x00000800; //!< O_EXCL
-    static constexpr int TGT_O_NOCTTY    = 0x00008000; //!< O_NOCTTY
-    static constexpr int TGT_O_TRUNC     = 0x00000400; //!< O_TRUNC
-    static constexpr int TGT_O_APPEND    = 0x00000008; //!< O_APPEND
-    static constexpr int TGT_O_NONBLOCK  = 0x00000004; //!< O_NONBLOCK
-    static constexpr int TGT_O_SYNC      = 0x00000080; //!< O_SYNC
-    static constexpr int TGT_FASYNC      = 0x00000040; //!< FASYNC
-    static constexpr int TGT_O_DIRECT    = 0x00010000; //!< O_DIRECT
+    static constexpr int TGT_O_RDONLY = 0x00000000;    //!< O_RDONLY
```
andysan, line 220: This is a bit annoying since it makes it harder to read bit masks. We probably don't want to generally mandate that values are aligned though. "AlignConsecutiveAssignments: true" would fix that, but I think it makes other parts of the code a bit messier.  We might want to consider enabling that locally for these types of use cases.

src/arch/arm/insts/crypto.cc
```diff
@@ -43,135 +43,117 @@
...
-};
-
-const uint8_t
-Crypto::aesINVSHIFT[16] = {
-    0, 13, 10, 7, 4, 1, 14, 11,
-    8, 5, 2, 15, 12, 9, 6, 3
-};
+namespace ArmISA
+{
+const uint8_t Crypto::aesSBOX[256] = {0x63, 0x7c, 0x77, 0x7b, 0xf2, 0x6b, 0x6f,
```
andysan: This looks a bit weird. I would prefer it if initialisation lists start on a new line.

src/arch/arm/insts/crypto.cc
```diff
@@ -43,135 +43,117 @@
...
+    0x89, 0x80, 0x9b, 0xb6, 0xc1, 0x58, 0xe8, 0x23, 0x65, 0xaf, 0xea, 0x25,
+    0x6f, 0xb1, 0xc8, 0x43, 0xc5, 0x54, 0xfc, 0x1f, 0x21, 0x63, 0xa5, 0xf4,
+    0x07, 0x09, 0x1b, 0x2d, 0x77, 0x99, 0xb0, 0xcb, 0x46, 0xca, 0x45, 0xcf,
+    0x4a, 0xde, 0x79, 0x8b, 0x86, 0x91, 0xa8, 0xe3, 0x3e, 0x42, 0xc6, 0x51,
+    0xf3, 0x0e, 0x12, 0x36, 0x5a, 0xee, 0x29, 0x7b, 0x8d, 0x8c, 0x8f, 0x8a,
+    0x85, 0x94, 0xa7, 0xf2, 0x0d, 0x17, 0x39, 0x4b, 0xdd, 0x7c, 0x84, 0x97,
+    0xa2, 0xfd, 0x1c, 0x24, 0x6c, 0xb4, 0xc7, 0x52, 0xf6, 0x01};
+
+const uint8_t Crypto::aesSHIFT[16] = {
+    0, 5, 10, 15, 4, 9, 14, 3, 8, 13, 2, 7, 12, 1, 6, 11};
```
andysan: Also looks weird. The closing curly should be on a separate line.

src/arch/arm/insts/data64.hh
```diff
@@ -43,24 +43,24 @@
...
-
 class DataXImmOp : public ArmStaticInst
 {
   protected:
     RegIndex dest, op1;
     uint64_t imm;
 
     DataXImmOp(const char *mnem, ExtMachInst _machInst, OpClass __opClass,
-               RegIndex _dest, RegIndex _op1, uint64_t _imm) :
+        RegIndex _dest, RegIndex _op1, uint64_t _imm) :
```
andysan: It seems like the old style consistently aligned argument lists. We should probably continue to do that to avoid a lot of unnecessary changes.

src/arch/arm/insts/fplib.cc
```diff
@@ -4367,32 +4223,20 @@ template <>
...
-            0xb800,
-            0x293a,
-            0x0000,
-            0x0000,
-            0x0000,
-            0x0000,
-            0x0000
-        }
-    };
+    static uint16_t coeff[2][8] = {{
```
andysan: This just looks ugly and hard to follow.

src/arch/arm/linux/se_workload.cc
```diff
@@ -146,350 +143,241 @@ setTLSFunc64(SyscallDesc *desc, ThreadContext *tc, uint32_t tlsPtr)
...
-        { base + 134, "bdflush" },
-        { base + 135, "sysfs" },
-        { base + 136, "personality" },
-        { base + 137, "reserved#138" },
-        { base + 138, "setfsuid" },
-        { base + 139, "setfsgid" },
-        { base + 140, "llseek", _llseekFunc },
+    SyscallTable32(int base) :
+        SyscallDescTable<EmuLinux::SyscallABI32>({
+            {base + 0, "syscall"}, {base + 1, "exit", exitFunc},
```
andysan: This type of formatting is massively unhelpful for readability. We should consider one element per line in some cases. Alternatively, we should consider telling clang-format to not change the way list elements are formatted since that sometimes reflects the underlying structure of the dataset.

src/arch/arm/mmu.hh
```diff
@@ -52,9 +52,8 @@
 
 namespace gem5
 {
-
```
andysan, line 55: This seems like a pretty consistent change. Should we consider mandating a newline instead? I find that it could help readability a bit since we aren't indenting namespaces.

src/arch/arm/pagetable.cc
```diff
@@ -61,39 +59,51 @@ bool
 V7LPageTableOps::isValid(pte_t pte, unsigned level) const
 {
     switch (level) {
-        case 1: return  pte & 0x1;
```
andysan, line 64: Do we want to allow this construct? It's pretty common as far as I can tell.  AllowShortCaseLabelsOnASingleLine controls that.

src/arch/arm/regs/misc.cc
```diff
@@ -726,566 +724,565 @@ unflattenMiscReg(int reg)
...
-    { MiscRegNum64(3, 6, 12, 0, 0), MISCREG_VBAR_EL3 },
-    { MiscRegNum64(3, 6, 12, 0, 1), MISCREG_RVBAR_EL3 },
-    { MiscRegNum64(3, 6, 12, 0, 2), MISCREG_RMR_EL3 },
-    { MiscRegNum64(3, 6, 12, 12, 4), MISCREG_ICC_CTLR_EL3 },
-    { MiscRegNum64(3, 6, 12, 12, 5), MISCREG_ICC_SRE_EL3 },
-    { MiscRegNum64(3, 6, 12, 12, 7), MISCREG_ICC_IGRPEN1_EL3 },
-    { MiscRegNum64(3, 6, 13, 0, 2), MISCREG_TPIDR_EL3 },
-    { MiscRegNum64(3, 7, 14, 2, 0), MISCREG_CNTPS_TVAL_EL1 },
-    { MiscRegNum64(3, 7, 14, 2, 1), MISCREG_CNTPS_CTL_EL1 },
-    { MiscRegNum64(3, 7, 14, 2, 2), MISCREG_CNTPS_CVAL_EL1 }
```
andysan: I have seen quite a few instances of this type of change where a white space after a curly in an initialisation list is removed. We should check which style is more common and go with that. I haven't found any obvious option to enforce that though.

src/arch/arm/regs/misc.cc
```diff
@@ -2735,508 +2721,418 @@ ISA::initializeMiscRegMetadata()
...
-      .allPrivileges();
-    InitReg(MISCREG_ELR_HYP)
-      .allPrivileges();
-    InitReg(MISCREG_FPSID)
-      .reset(p.fpsid)
-      .allPrivileges();
-    InitReg(MISCREG_FPSCR)
-      .res0(mask(14, 13) | mask(6, 5))
-      .allPrivileges();
+    InitReg(MISCREG_CPSR).reset(resetCPSR(system)).allPrivileges();
```
andysan: There doesn't seem to be any logic to how this has been reformatted.  The formatter sometimes applies more aggressive line breaks and sometimes removes existing line breaks seemingly at random.

.clang-format
```diff
@@ -0,0 +1,455 @@
...
+
+# BreakBeforeConceptDeclarations (BreakBeforeConceptDeclarationsStyle) clang-format 12
+
+# BreakBeforeInlineASMColon (BreakBeforeInlineASMColonStyle) clang-format 16
+
+# BreakBeforeTernaryOperators (Boolean) clang-format 3.7
+BreakBeforeTernaryOperators: false
+
+# BreakConstructorInitializers (BreakConstructorInitializersStyle) clang-format 5
+BreakConstructorInitializers: AfterColon
```
andysan: This should be 'BeforeColon' to be consistent with current use in gem5. I can see the value in AfterColon, but let's not change that now.

.clang-format
```diff
@@ -0,0 +1,455 @@
...
+Language:        Cpp
+
+# Based on clang-10 (latest version supported on Ubuntu 20)
+# No BasedOnStyle
+
+# AccessModifierOffset (Integer) clang-format 3.3
+AccessModifierOffset: -2
+
+# AlignAfterOpenBracket (BracketAlignmentStyle) clang-format 3.8
+AlignAfterOpenBracket: DontAlign
```
andysan: It seems like most code is consistent with the 'Align' option.

.clang-format
```diff
@@ -0,0 +1,455 @@
...
+# AllowShortBlocksOnASingleLine (ShortBlockStyle) clang-format 3.5
+AllowShortBlocksOnASingleLine: Empty
+
+# AllowShortCaseLabelsOnASingleLine (Boolean) clang-format 3.6
+AllowShortCaseLabelsOnASingleLine: false
+
+# AllowShortEnumsOnASingleLine (Boolean) clang-format 11
+
+# AllowShortFunctionsOnASingleLine (ShortFunctionStyle) clang-format 3.5
+AllowShortFunctionsOnASingleLine: All
```
andysan, line 54: This should be 'None' to be consistent with the current style guide. Another option would be to set it to 'Inline' and set AlwaysBreakAfterReturnType to 'TopLevelDefinitions'.

.clang-format
```diff
@@ -0,0 +1,455 @@
...
+  AfterStruct:     true
+  AfterUnion:      true
+  AfterExternBlock: true
+  BeforeCatch:     false
+  BeforeElse:      false
+  # Not clang-format 10
+  # BeforeLambdaBody: false
+  # BeforeWhile:     false
+  IndentBraces:    false
+  SplitEmptyFunction: false
```
andysan: All of the split options (maybe with the exception of SplitEmptyRecord) should be true to replicate the current style guide.

.clang-format
andysan: It seems like 'Right' is a pretty common pattern in the code base. I'd suggest setting this one to 'Right' if that produces a smaller diff.

.clang-format
andysan: I haven't seen any cases of the AfterHash style in the code base. We should consider keeping this as None unless we it makes the diff larger.

.clang-format
```diff
@@ -0,0 +1,455 @@
...
+AllowAllConstructorInitializersOnNextLine: true
+
+# AllowAllParametersOfDeclarationOnNextLine (Boolean) clang-format 3.3
+AllowAllParametersOfDeclarationOnNextLine: true
+
+# AllowShortBlocksOnASingleLine (ShortBlockStyle) clang-format 3.5
+AllowShortBlocksOnASingleLine: Empty
+
+# AllowShortCaseLabelsOnASingleLine (Boolean) clang-format 3.6
+AllowShortCaseLabelsOnASingleLine: false
```
andysan, line 49: Lots of the code base does this, we should consider setting this to true.

.clang-format
```diff
@@ -0,0 +1,455 @@
...
+# SeparateDefinitionBlocks (SeparateDefinitionStyle) clang-format 14
+
+# ShortNamespaceLines (Unsigned) clang-format 13
+
+# SortIncludes (SortIncludesOptions) clang-format 3.8
+# TODO: we cannot sort includes as it breaks the compilation
+# One way to fix this would be to use IWYU (include what you use)
+# But this adds complexity as some header are generated during the compilation,
+# hence cannot be run during pre-commit.
+SortIncludes:    false
```
andysan: This should be set to `CaseInsensitive` if we are retiring the old script. I think I have a working config (see separate comment).

.clang-format
```diff
@@ -0,0 +1,455 @@
...
+ForEachMacros:
+  - GEM5_FOR_EACH_IN_PACK
+
+# IfMacros (List of Strings) clang-format 13
+
+# IncludeBlocks (IncludeBlocksStyle) clang-format 6
+IncludeBlocks:   Preserve
+
+# IncludeCategories (List of IncludeCategories) clang-format 3.8
+IncludeCategories:
```
andysan, line 186: The following seems to (mostly) replicate what the old scripts enforce:

```
# IncludeBlocks (IncludeBlocksStyle) clang-format 6
IncludeBlocks: Regroup

# IncludeCategories (List of IncludeCategories) clang-format 3.8
IncludeCategories:
    - Regex: ^<Python.h>$
      Priority: 0
      SortPriority: -1
    - Regex: ^"pybind11/.*"
      Priority: 1
      SortPriority: -1
    - Regex: ^<.*\.h>
      Priority: 1
    - Regex: ^<[^.]*>$
      Priority: 2
    - Regex: ^<.*
      Priority: 3
    - Regex: .*
      Priority: 4
```

.clang-format
```diff
@@ -0,0 +1,455 @@
...
+# AccessModifierOffset (Integer) clang-format 3.3
+AccessModifierOffset: -2
+
+# AlignAfterOpenBracket (BracketAlignmentStyle) clang-format 3.8
+AlignAfterOpenBracket: DontAlign
+
+# AlignArrayOfStructures (ArrayInitializerAlignmentStyle) clang-format 13
+
+# AlignConsecutiveAssignments (Boolean) clang-format 3.8
+AlignConsecutiveAssignments: false
```
andysan, line 17: We might want to consider enforcing this depending on the size of the resulting diff. We use this type of alignment in many cases where it is useful to compare values (e.g., bit fields).

.clang-format
```diff
@@ -0,0 +1,455 @@
...
+IncludeIsMainSourceRegex: ''
+
+# IndentAccessModifiers (Boolean) clang-format 13
+
+# IndentCaseBlocks (Boolean) clang-format 11
+
+# IndentCaseLabels (Boolean) clang-format 3.3
+IndentCaseLabels: false
+
+# IndentExternBlock (IndentExternBlockStyle) clang-format 11
```
andysan, line 215: This should be set to false to be consistent with the current code base (depending on which clang version we require).

andysan (changes requested):
OK, so this is a massive diff. As I mentioned earlier, I'm generally supportive of switching to clang-format. I have played around with the configuration locally and suggested some changes which seem to more closely mirror what we currently implement.

There are some cases where the code gets a lot less readable because of this change. This is particularly the case for structured data where indentation and line breaks are used to make the underlying structure more obvious. It would be nice if we could actually review the change set and ensure that the relevant `// clang-format off` and `// clang-format on` tags.

powerjg:
@andysan Some of these points are important and need to be taken into account. However, we've agreed in the past that the goal is to get *close* to the current style, but not to try to mimic it *perfectly*. Reducing the diff would be nice, but we should weight that with how much time/effort it takes to try different clang format styles. Furthermore, we should limit the cases of "my preference would be" and concentrate on what is best for the project as a whole.

If you could reply to your review comments specifying what is *required* and what is *optional* it will make it much smoother to try to merge this change.

andysan:
> @andysan Some of these points are important and need to be taken into account. However, we've agreed in the past that the goal is to get _close_ to the current style, but not to try to mimic it _perfectly_. Reducing the diff would be nice, but we should weight that with how much time/effort it takes to try different clang format styles. Furthermore, we should limit the cases of "my preference would be" and concentrate on what is best for the project as a whole.

Given the size of the diff, I strongly believe that the best for the project as a whole is to make it as small as possible. Massive changes like these make it really painful to port changes between revisions and rebase existing patches, so we should try to minimise them where possible There are unavoidable changes like the way we indent switch statements that we will have to live with, but we shouldn't avoid tuning the configuration where there are avoidable changes. Honestly, it's not that much effort I just went over a few files and I have managed to reduce it quite substantially for those files. If we really care about what's best for the project, we should let this take the time it needs instead of rushing to meet some arbitrary deadline.

> If you could reply to your review comments specifying what is _required_ and what is _optional_ it will make it much smoother to try to merge this change.

Consider all of the suggested changes to the clang-format file as required unless they specifically say otherwise. They were designed to minimize the diff on the files I have looked at and were suggested to solve the issues I pointed out in the rest of the review.

powerjg:
> Honestly, it's not that much effort I just went over a few files and I have managed to reduce it quite substantially for those files. 

If you've already gone through the effort to make the changes, it would be incredibly helpful for you to push those changes here. You should have permission to update the PR.

andysan:
What I would suggest that we do, which I think _is_ the best thing for the project is to update this PR with the changes I suggested, we then let maintainers review the change and find some of the structures where this would be very disruptive and make the code harder to follow and annotate them in separate PRs. For example, we have various structures where we define properties of registers which get a lot harder to read after this change. I would rather review them now and add the necessary comments to disable clang-format than selectively revert this change in the future. Concretely, that probably means that this change won't go into this release. :(

Note that I'm not saying that we shouldn't include this change. We absolutely should. I just want to minimise the impact on the  community. The more intrusive this change is the more unlikely it is that people rebase their changes and submit them upstream.

andysan:
> > Honestly, it's not that much effort I just went over a few files and I have managed to reduce it quite substantially for those files.
> 
> If you've already gone through the effort to make the changes, it would be incredibly helpful for you to push those changes here. You should have permission to update the PR.

I'm just on my way out and will be busy with other projects the next couple of days. :(

powerjg:
I don't disagree with anything you're saying @andysan. However, what I'm seeing is that we have been working on merging clang-format for *over 2 years*. See https://gem5-review.googlesource.com/c/public/gem5/+/49063 for the beginning of the conversation.

Over and over again, someone has come in at the "last second" (i.e., when someone says "ok, we're getting ready to merge") and makes suggestions. It's not that the suggestions aren't helpful--they are---but the problem is that the suggestions are coming late. 

As far as I can tell, the only way this change will ever be merged is if we have a concerted push with some forcing function (e.g., a release). I'm fine with pushing this until the next release, but I foresee the same thing is going to happen again. We'll wait until a few days before creating the staging branch for 24.0 and then say "we can't merge this because it's not perfect."

I'll also point out that many people including me, Tom, Bobby, Giacomo, Gabe, Andreas, and others, have spent a non-insignificant amount of time and effort on this change. It is disheartening that we cannot get it over the finish line.

Finally, I want to point out that while this may cause some short-term pain with updating changes, it should significantly decrease the pain of contributing afterwards. I would guess that at least 25-50% of review comments are on style, this will reduce the workload of both reviewers and contributors. It is a *very important* change to get in.

andysan:
> I don't disagree with anything you're saying @andysan. However, what I'm seeing is that we have been working on merging clang-format for _over 2 years_. See https://gem5-review.googlesource.com/c/public/gem5/+/49063 for the beginning of the conversation.
> 
> Over and over again, someone has come in at the "last second" (i.e., when someone says "ok, we're getting ready to merge") and makes suggestions. It's not that the suggestions aren't helpful--they are---but the problem is that the suggestions are coming late.

I don't think it is fair to say that this is last second. The PR had been sitting idle for a month and a half until Bobby picked it up. The updated patch with the changes applied to the code base had been available for less than 24h when I posted my reviews. The original author even described it as "this PR needs more work" 4 days ago. There is no way that can be described as last second.

We need to rethink the whole release process if it becomes a justification to merge half-baked changes. That's approach is simply not good enough.

> As far as I can tell, the only way this change will ever be merged is if we have a concerted push with some forcing function (e.g., a release). I'm fine with pushing this until the next release, but I foresee the same thing is going to happen again. We'll wait until a few days before creating the staging branch for 24.0 and then say "we can't merge this because it's not perfect."
> 
> I'll also point out that many people including me, Tom, Bobby, Giacomo, Gabe, Andreas, and others, have spent a non-insignificant amount of time and effort on this change. It is disheartening that we cannot get it over the finish line.

I'm absolutely not saying that we can't get this over the finishing line. That's what I'm trying to achieve here. Every time we rush a big change like this we break things. Every time that happens, I need to explain to my colleagues why we should continue to use gem5 or why we shouldn't fork it and just go our own way. My life needs less of this. Can we please make sure I that I can at least justify to my colleagues that we did our best to reduce breakage?

> Finally, I want to point out that while this may cause some short-term pain with updating changes, it should significantly decrease the pain of contributing afterwards. I would guess that at least 25-50% of review comments are on style, this will reduce the workload of both reviewers and contributors. It is a _very important_ change to get in.

Completely agree, we need this change. But at the same time, we want to encourage people to submit their code. If they can't rebase their code because we changed _everything_, it'll just end up bit rotting.

andysan:
Before I forget: We'll almost certainly need to change `util/emacs/*` and `util/vi/*` to match the updated style.

pranith:
Is there any interest in increasing the line with up from 80 characters? I ask because many other major open source code bases have moved on from 80 characters because of the wide usage of high density displays. Especially for C++, with wide namespace and long identifiers, increasing the width makes the code more readable.

andysan:
> Is there any interest in increasing the line with up from 80 characters? I ask because many other major open source code bases have moved on from 80 characters because of the wide usage of high density displays. Especially for C++, with wide namespace and long identifiers, increasing the width makes the code more readable.

I think this is something that is worth considering. IIRC, black enforces 88 characters for Python (unless we changed that). I would be reluctant to go much longer since that tends to decrease readability.

powerjg:
> I think this is something that is worth considering. IIRC, black enforces 88 characters for Python (unless we changed that). I would be reluctant to go much longer since that tends to decrease readability.

We use 79 characters for python/Black: https://github.com/gem5/gem5/blob/stable/pyproject.toml#L7

giactra:
Hi @BobbyRBruce, I have some time this week to dedicate at this PR. Do you mind if I steal it from you? It would be nice if we could get some good progress and get closer to getting this merged

BobbyRBruce:
> Hi @BobbyRBruce, I have some time this week to dedicate at this PR. Do you mind if I steal it from you? It would be nice if we could get some good progress and get closer to getting this merged

Steal all you won't I wan't planning on getting back to it anytime soon.

src/arch/arm/fastmodel/iris/thread_context.cc
giactra: Agreed

src/arch/arm/fastmodel/iris/thread_context.hh
giactra: So https://clang.llvm.org/docs/ClangFormatStyleOptions.html#separatedefinitionblocks could be a solution. The downside is that it will always enforce it if enabled with SDS_Always.
However in gem5 we have this unofficial coding style of having a newline in general between function definitions, but not if they are short (on a single line).

So I usually see this:

```
// Notice there is no newline
int func1() { return 1; }
int funct2() { return 2; }
```

and

```
// Notice the newline
int
func3()
{
    return 3;
}

int
func4()
{
    return 4
}
```

So I would keep it as it is

src/arch/arm/insts/crypto.cc
giactra: Yeah, I am digging but to be fair the only solution to this problem seems to be the BAS_BlockIndent option in https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignafteropenbracket. However replacing the current BAS_Align with BAS_BlockIndent will massively change most of our code which is usually written with BAS_Align in mind

src/arch/arm/insts/crypto.cc
giactra: Ditto

src/arch/arm/fastmodel/iris/thread_context.hh
andysan: So you're suggesting that we set this to SDS_Leave?

src/arch/arm/insts/crypto.cc
andysan: I think for some of these data structures where here is some structure in the data or it's just a very large block of values, we should consider just suppressing clang-format for that region. This might be one of these cases.

powerjg: +1

I think there's probably quite a few places we want to suppress clang format for this reason.

src/arch/arm/fastmodel/iris/thread_context.hh
odanrc: Is there a drawback (other than slightly increasing file length) in adding this new line in these scenarios? IMHO it is better to add this new line, as it reduces the likelihood of merge conflicts between PRs that touch functions that are close to each other.

src/arch/amdgpu/common/tlb.hh
```diff
@@ -55,336 +55,353 @@
...
+        virtual void markDelayed() = 0;
 
         /**
-         * if true, then this is not the last level TLB
+         * The memory for this object may be dynamically allocated, and it
+         * may be responsible for cleaning itslef up which will happen in
+         * this function. Once it's called the object is no longer valid.
          */
-        bool hasMemSidePort;
+        virtual void finish(Fault fault, const RequestPtr &req,
```
odanrc, line 97: Is aligning the args to the left parenthesis the rule that led to the least amount of immediate changes in the codebase?

I ask this because since the alignment depends on the length of the function´s name/return type, it becomes variable, possibly requiring re-alignment in the future in case the function´s name/return type changes. I am not a fan of this approach, and I prefer only adding one level of indentation to the continuation lines, but this is just my opinion, so you can disregard it :)

odanrc (approved):

rogerchang23424:
The PR isn't on the ToT of develop branch, can we rebase it?

.clang-format
giactra: I checked and while I agree it is useful in certain circumstances, it makes the diff much bigger. I propose for users to turn clang-format off in case they believe an alignment will improve readability

.clang-format
giactra: This is already handled by BraceWrapping.AfterExternBlock

src/arch/arm/freebsd/freebsd.hh
giactra: Agree!

giactra:
> The PR isn't on the ToT of develop branch, can we rebase it?

I just rebased it

src/arch/arm/fastmodel/iris/thread_context.hh
giactra: I think @odanrc  makes a good point. I used the SDS_Always and it looks like the price we pay for it it is relatively ok (the PR adds more lines but also avoids lines removal) considering it will make it more likely to avoid merge conflicts which is ultimately the major concern with the introduction of clang-format

src/arch/arm/regs/misc.cc
giactra: This has been addressed with [Cpp11BracedListStyle=false](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#cpp11bracedliststyle) and it creates a smaller diff

src/arch/amdgpu/common/tlb.hh
giactra: > Is aligning the args to the left parenthesis the rule that led to the least amount of immediate changes in the codebase?
> 

Yes as most of the code was written in this way

> I ask this because since the alignment depends on the length of the function´s name/return type, it becomes variable, possibly requiring re-alignment in the future in case the function´s name/return type changes. I am not a fan of this approach, and I prefer only adding one level of indentation to the continuation lines, but this is just my opinion, so you can disregard it :)

Yes I agree that aligning is not the best for the reason you mentioned (on top of which I would also add merge/rebase conflicts). However at the moment changing the option would make the diff bigger so I would keep it as it is if you don't mind

odanrc: If the plan is to keep it as is and not change to the new-line rule, then that is the sensible approach; however, if the plan is to later create a PR to change to use the new-line rule, then that means that we will generate merge conflicts on the same spots twice, which does not seem to be a good idea.

BobbyRBruce (dismissed):
I pushed a merge commit so this PR is up-to-date with the latest changes on develop. As part of that merge commit I re-ran the pre-commit on all the files.

I'm personally happy with this and therefore happy to approve it as-is.

rogerchang23424 (commented):
Something require changed

src/arch/riscv/insts/vector.cc
```diff
@@ -429,16 +468,16 @@ VMaskMergeMicroInst::VMaskMergeMicroInst(ExtMachInst extMachInst,
...
+    for (uint8_t i = 0; i < _numSrcs; i++) {
         setSrcRegIdx(_numSrcRegs++, vecRegClass[VecMemInternalReg0 + i]);
     }
 }
 
 Fault
-VMaskMergeMicroInst::execute(ExecContext* xc,
-    trace::InstRecord* traceData) const
+VMaskMergeMicroInst::execute(ExecContext *xc,
+                             trace::InstRecord *traceData) const
```
rogerchang23424, line 478: Can we clang-format off in `VMaskMergeMicroInst::execute` section? It's like generated from `isa_parser.py`

src/arch/riscv/insts/vector.cc
```diff
@@ -575,51 +614,53 @@ VlFFTrimVlMicroOp::branchTarget(ThreadContext *tc) const
...
-    : VectorArithMicroInst("vlseg_deintrlv_micro", extMachInst,
-                            SimdAddOp, 0, 0),
-        vlen(_vlen)
+VlSegDeIntrlvMicroInst::VlSegDeIntrlvMicroInst(
+    ExtMachInst extMachInst, uint32_t _micro_vl, uint32_t _dstReg,
+    uint32_t _numSrcs, uint32_t _microIdx, uint32_t _numMicroops,
+    uint32_t _field, uint32_t _vlen, uint32_t _sizeOfElement)
+    : VectorArithMicroInst("vlseg_deintrlv_micro", extMachInst, SimdAddOp, 0,
+                           0),
+      vlen(_vlen)
```
rogerchang23424, line 663: Can we clang-format off in `VlSegDeIntrlvMicroInst::VlSegDeIntrlvMicroInst` section?

src/arch/riscv/insts/vector.cc
```diff
@@ -631,37 +672,37 @@ VlSegDeIntrlvMicroInst::VlSegDeIntrlvMicroInst(ExtMachInst extMachInst, uint32_t
...
+    for (uint32_t i = 0; i < _numSrcs; i++) {
         uint32_t index = VecMemInternalReg0 + i + (microIdx * _numSrcs);
         setSrcRegIdx(_numSrcRegs++, vecRegClass[index]);
     }
 }
 
 Fault
-VlSegDeIntrlvMicroInst::execute(ExecContext* xc, trace::InstRecord* traceData) const
+VlSegDeIntrlvMicroInst::execute(ExecContext *xc,
+                                trace::InstRecord *traceData) const
```
rogerchang23424, line 690: Can we clang-format off in `VlSegDeIntrlvMicroInst::execute` section ?

rogerchang23424 (changes requested):
Few thing I would like to not apply clang-format

BobbyRBruce:
I appear to have messed up that merge. I'm going to reattempt it.

src/arch/riscv/insts/vector.cc
```diff
@@ -484,37 +523,37 @@ VMaskMergeMicroInst::generateDisassembly(Addr pc,
 }
 
 Fault
-VxsatMicroInst::execute(ExecContext* xc, trace::InstRecord* traceData) const
+VxsatMicroInst::execute(ExecContext *xc, trace::InstRecord *traceData) const
```
rogerchang23424, line 526: Can we clang-format off for the section?

src/arch/riscv/insts/vector.cc
```diff
@@ -684,39 +725,41 @@ VlSegDeIntrlvMicroInst::generateDisassembly(Addr pc, const loader::SymbolTable *
...
-    : VectorArithMicroInst("vsseg_reintrlv_micro", extMachInst,
-                            SimdAddOp, 0, 0),
-        vlen(_vlen)
+VsSegIntrlvMicroInst::VsSegIntrlvMicroInst(
+    ExtMachInst extMachInst, uint32_t _micro_vl, uint32_t _dstReg,
+    uint32_t _numSrcs, uint32_t _microIdx, uint32_t _numMicroops,
+    uint32_t _field, uint32_t _vlen, uint32_t _sizeOfElement)
+    : VectorArithMicroInst("vsseg_reintrlv_micro", extMachInst, SimdAddOp, 0,
+                           0),
+      vlen(_vlen)
```
rogerchang23424: Can we clang-format off for the section?

src/arch/riscv/insts/vector.cc
```diff
@@ -728,27 +771,28 @@ VsSegIntrlvMicroInst::VsSegIntrlvMicroInst(ExtMachInst extMachInst, uint32_t _mi
...
+        setSrcRegIdx(_numSrcRegs++,
+                     vecRegClass[_dstReg + (i * numMicroops) + (microIdx)]);
     }
 }
 
 Fault
-VsSegIntrlvMicroInst::execute(ExecContext* xc,
-    trace::InstRecord* traceData) const
+VsSegIntrlvMicroInst::execute(ExecContext *xc,
+                              trace::InstRecord *traceData) const
```
rogerchang23424, line 792: Can we clang-format off the section?

src/arch/riscv/regs/float.hh
```diff
@@ -90,22 +90,79 @@ unboxF32(uint64_t v)
...
+boxF16(uint16_t v)
+{
+    return mask(63, 16) | v;
+}
+
+static constexpr uint64_t
+boxF32(uint32_t v)
+{
+    return mask(63, 32) | v;
+}
```
rogerchang23424:
```suggestion
/* clang-format off */
static constexpr uint64_t boxF16(uint16_t v) { return mask(63, 16) | v; }
static constexpr uint64_t boxF32(uint32_t v) { return mask(63, 32) | v; }
/* clang-format on */
```

src/arch/riscv/regs/float.hh
```diff
@@ -90,22 +90,79 @@ unboxF32(uint64_t v)
...
+f32(freg_t r)
+{
+    return { unboxF32(r.v) };
+}
+
+static constexpr float64_t
+f64(freg_t r)
+{
+    return r;
+}
```
rogerchang23424:
```suggestion
/* clang-format off */
static constexpr float16_t f16(uint16_t v) { return {v}; }
static constexpr float32_t f32(uint32_t v) { return {v}; }
static constexpr float64_t f64(uint64_t v) { return {v}; }
static constexpr float16_t f16(freg_t r) { return {unboxF16(r.v)}; }
static constexpr float32_t f32(freg_t r) { return {unboxF32(r.v)}; }
static constexpr float64_t f64(freg_t r) { return r; }
/* clang-format on */
```

src/arch/riscv/regs/float.hh
```diff
@@ -90,22 +90,79 @@ unboxF32(uint64_t v)
...
+freg(float64_t f)
+{
+    return f;
+}
+
+static constexpr freg_t
+freg(uint_fast64_t f)
+{
+    return { f };
+}
```
rogerchang23424:
```suggestion
/* clang-format off */
static constexpr freg_t freg(float16_t f) { return {boxF16(f.v)}; }
static constexpr freg_t freg(float32_t f) { return {boxF32(f.v)}; }
static constexpr freg_t freg(float64_t f) { return f; }
static constexpr freg_t freg(uint_fast64_t f) { return {f}; }
/* clang-format on */
```

src/arch/riscv/regs/float.hh
```diff
@@ -155,73 +212,60 @@ enum : RegIndex
...
+    Fa4 = floatRegClass[_Fa4Idx], Fa5 = floatRegClass[_Fa5Idx],
+    Fa6 = floatRegClass[_Fa6Idx], Fa7 = floatRegClass[_Fa7Idx],
+
+    Fs2 = floatRegClass[_Fs2Idx], Fs3 = floatRegClass[_Fs3Idx],
+    Fs4 = floatRegClass[_Fs4Idx], Fs5 = floatRegClass[_Fs5Idx],
+    Fs6 = floatRegClass[_Fs6Idx], Fs7 = floatRegClass[_Fs7Idx],
+    Fs8 = floatRegClass[_Fs8Idx], Fs9 = floatRegClass[_Fs9Idx],
+    Fs10 = floatRegClass[_Fs10Idx], Fs11 = floatRegClass[_Fs11Idx],
+
+    Ft8 = floatRegClass[_Ft8Idx], Ft9 = floatRegClass[_Ft9Idx],
```
rogerchang23424: These part should not apply clang-format

src/arch/riscv/regs/float.hh
```diff
@@ -155,73 +212,60 @@ enum : RegIndex
...
-    "fa2", "fa3", "fa4", "fa5",
-    "fa6", "fa7", "fs2", "fs3",
-    "fs4", "fs5", "fs6", "fs7",
-    "fs8", "fs9", "fs10", "fs11",
-    "ft8", "ft9", "ft10", "ft11"
+    "ft0", "ft1", "ft2",  "ft3",  "ft4", "ft5", "ft6",  "ft7",
+    "fs0", "fs1", "fa0",  "fa1",  "fa2", "fa3", "fa4",  "fa5",
+    "fa6", "fa7", "fs2",  "fs3",  "fs4", "fs5", "fs6",  "fs7",
+    "fs8", "fs9", "fs10", "fs11", "ft8", "ft9", "ft10", "ft11"
 };
```
rogerchang23424: These part should not apply clang-format

src/arch/riscv/rvk.hh
```diff
@@ -121,53 +121,89 @@ const uint8_t _rvk_emu_sm4_sbox[256] = {
...
 
-inline int32_t _rvk_emu_sll_32(int32_t rs1, int32_t rs2)
-    { return rs1 << (rs2 & 31); }
-inline int32_t _rvk_emu_srl_32(int32_t rs1, int32_t rs2)
-    { return (uint32_t)rs1 >> (rs2 & 31); }
-inline int64_t _rvk_emu_sll_64(int64_t rs1, int64_t rs2)
-    { return rs1 << (rs2 & 63); }
-inline int64_t _rvk_emu_srl_64(int64_t rs1, int64_t rs2)
-    { return (uint64_t)rs1 >> (rs2 & 63); }
+inline int32_t
```
rogerchang23424, line 124: The file is reference from [rvkrypto-fips](https://github.com/rvkrypto/rvkrypto-fips/tree/main). It should not apply clang-format

src/arch/riscv/utility.hh
```diff
@@ -65,59 +65,120 @@ namespace gem5
...
+struct double_widthf<uint32_t>
+{
+    using type = float64_t;
+};
+
+template <>
+struct double_widthf<int32_t>
+{
+    using type = float64_t;
+};
```
rogerchang23424, line 138: I would like to keep original format, the change makes it less readable.

```suggestion
/* clang-format off */
template<typename Type> struct double_width;
template<> struct double_width<uint8_t>     { using type = uint16_t;};
template<> struct double_width<uint16_t>    { using type = uint32_t;};
template<> struct double_width<uint32_t>    { using type = uint64_t;};
template<> struct double_width<uint64_t>    { using type = __uint128_t;};
template<> struct double_width<int8_t>      { using type = int16_t; };
template<> struct double_width<int16_t>     { using type = int32_t; };
template<> struct double_width<int32_t>     { using type = int64_t; };
template<> struct double_width<int64_t>     { using type = __int128_t; };
template<> struct double_width<float32_t>   { using type = float64_t;};

template<typename Type> struct double_widthf;
template<> struct double_widthf<uint32_t>    { using type = float64_t;};
template<> struct double_widthf<int32_t>     { using type = float64_t;};
/* clang-format on */
```

src/arch/riscv/regs/int.hh
```diff
@@ -82,72 +106,49 @@ enum : RegIndex
...
+                       A4 = intRegClass[_A4Idx], A5 = intRegClass[_A5Idx],
+                       A6 = intRegClass[_A6Idx], A7 = intRegClass[_A7Idx],
+                       S2 = intRegClass[_S2Idx], S3 = intRegClass[_S3Idx],
+                       S4 = intRegClass[_S4Idx], S5 = intRegClass[_S5Idx],
+                       S6 = intRegClass[_S6Idx], S7 = intRegClass[_S7Idx],
+                       S8 = intRegClass[_S8Idx], S9 = intRegClass[_S9Idx],
+                       S10 = intRegClass[_S10Idx], S11 = intRegClass[_S11Idx],
+                       T3 = intRegClass[_T3Idx], T4 = intRegClass[_T4Idx],
+                       T5 = intRegClass[_T5Idx], T6 = intRegClass[_T6Idx],
+                       Ureg0 = intRegClass[_Ureg0Idx];
```
rogerchang23424, line 130: Can we clang-fromat off ?

```suggestion
/* clang-format off */
inline constexpr RegId
    Zero = intRegClass[_ZeroIdx],
    Ra = intRegClass[_RaIdx],
    Sp = intRegClass[_SpIdx],
    Gp = intRegClass[_GpIdx],
    Tp = intRegClass[_TpIdx],
    T0 = intRegClass[_T0Idx],
    T1 = intRegClass[_T1Idx],
    T2 = intRegClass[_T2Idx],
    S0 = intRegClass[_S0Idx],
    S1 = intRegClass[_S1Idx],
    A0 = intRegClass[_A0Idx],
    A1 = intRegClass[_A1Idx],
    A2 = intRegClass[_A2Idx],
    A3 = intRegClass[_A3Idx],
    A4 = intRegClass[_A4Idx],
    A5 = intRegClass[_A5Idx],
    A6 = intRegClass[_A6Idx],
    A7 = intRegClass[_A7Idx],
    S2 = intRegClass[_S2Idx],
    S3 = intRegClass[_S3Idx],
    S4 = intRegClass[_S4Idx],
    S5 = intRegClass[_S5Idx],
    S6 = intRegClass[_S6Idx],
    S7 = intRegClass[_S7Idx],
    S8 = intRegClass[_S8Idx],
    S9 = intRegClass[_S9Idx],
    S10 = intRegClass[_S10Idx],
    S11 = intRegClass[_S11Idx],
    T3 = intRegClass[_T3Idx],
    T4 = intRegClass[_T4Idx],
    T5 = intRegClass[_T5Idx],
    T6 = intRegClass[_T6Idx],
    Ureg0 = intRegClass[_Ureg0Idx];
 /* clang-format on */
```

src/arch/riscv/insts/vector.cc
BobbyRBruce: Is there a good argument against this style? I fear we're going to go back and forth forever on stuff like this. I personally don't see much wrong with it.

.clang-format
rogerchang23424: I prefer `Inline` in this option and set AlwaysBreakAfterReturnType to 'TopLevelDefinitions'

src/arch/riscv/insts/vector.cc
rogerchang23424: Sorry for lately reply. I have hide and deleted some commit not required update. 

> Is there a good argument against this style? I fear we're going to go back and forth forever on stuff like this. I personally don't see much wrong with it.

Some of format option required are updated here: https://github.com/gem5/gem5/pull/362#discussion_r1413750610. 

For the part of MicroStaticInst implement. I feel it will be little harder to compare the StaticInst generated from `isa_parser.py`.

Feel free to leave comments, thanks.

.clang-format
giactra: Hey @rogerchang23424 I tried out your option and I diffed current develop with both setups:

1) As of now:
2144 files changed, 194403 insertions(+), 219474 deletions(-)

2) With your changes:
2168 files changed, 198226 insertions(+), 194691 deletions(-)

As you can see you have more insertions and files changed but far less deletions. Is this the right thing to do? Shall we add and subtract and base our strategy over an absolute number? If that's the case we should go for your solution.

However let me just add I have the feeling we might achieve a better score by using [RTBS_ExceptShortType](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakafterreturntype). However that option requires clang-format v19 which is currently unavailable in the pre-commit mirror:

https://github.com/pre-commit/mirrors-clang-format

rogerchang23424: Thanks for @giactra points out, I think the best way is only set AlwaysBreakAfterReturnType to 'TopLevelDefinitions'.

Current: https://github.com/rogerchang23424/gem5/commit/5c3279fa139527c8ae6d4f2af82bd193d83bee65
Only TopLevelDefinition: https://github.com/rogerchang23424/gem5/commit/7f15d7adec1a1ab1e61580bbc8296688586506cf
My change: https://github.com/rogerchang23424/gem5/commit/7f5453a534a5bbd1c00a3658aae02c58a338e673 (more line additions and deletions)

src/arch/riscv/regs/float.hh
```diff
@@ -90,79 +90,25 @@ unboxF32(uint64_t v)
...
-    return mask(63, 16) | v;
-}
-
-static constexpr uint64_t
-boxF32(uint32_t v)
-{
-    return mask(63, 32) | v;
-}
+/* clang-format off */
+static constexpr uint64_t boxF16(uint16_t v) { return mask(63, 16) | v; }
```
arichardson, line 94: Shouldn't this formatting be possible with the AllowShortFucntions flag?

abmerop:
v24.1?  Should we rip the band aid off?

BobbyRBruce:
> v24.1? Should we rip the band aid off?

At this stage it's anyone's task to do if they want it. I've found it hard to stop this task getting lost in bikeshedding discussions about the CPP code style so I've been putting it off. Anyone who wants to clean this up and go for another round of reviewing would have my full support.

arichardson:
I just tried using https://github.com/alandefreitas/clang-unformat to generate a minimal diff clang-format config file. I used `src/base/` as the input files to check (since using all of src/ takes too long). Not sure if there is a directory that is considered to be "mostly conforming to the current style" that would be a better choice?

The resulting .clang-format was the following:
```
# .clang-format inferred with clang-unformat (https://www.github.com/alandefreitas/clang-unformat)
Language: Cpp                                      # single option
BasedOnStyle: Microsoft                            # edit distance 24197
AccessModifierOffset: -2                           # edit distance 24197
AlignAfterOpenBracket: DontAlign                   # edit distance 23291
AlignArrayOfStructures: Right                      # edit distance 20277
AlignConsecutiveAssignments: None                  # edit distance 20277
AlignConsecutiveBitFields: None                    # inherited from prefix Align
AlignConsecutiveDeclarations: None                 # edit distance 20277
AlignConsecutiveMacros: None                       # inherited from prefix Align
AlignEscapedNewlines: Left                         # edit distance 19643
AlignOperands: DontAlign                           # edit distance 19259
AlignTrailingComments: false                       # edit distance 18843
AllowAllArgumentsOnNextLine: false                 # edit distance 18785
AllowAllParametersOfDeclarationOnNextLine: true    # did not affect the output
AllowShortBlocksOnASingleLine: Never               # Using default value Never
AllowShortCaseLabelsOnASingleLine: false           # edit distance 18785
AllowShortEnumsOnASingleLine: true                 # did not affect the output
AllowShortFunctionsOnASingleLine: InlineOnly       # edit distance 18504
AllowShortIfStatementsOnASingleLine: Never         # edit distance 18504
AllowShortLambdasOnASingleLine: None               # edit distance 18429
AllowShortLoopsOnASingleLine: false                # edit distance 18429
AlwaysBreakAfterDefinitionReturnType: TopLevel     # edit distance 17970
AlwaysBreakAfterReturnType: None                   # edit distance 17970
AlwaysBreakBeforeMultilineStrings: true            # edit distance 17894
AlwaysBreakTemplateDeclarations: MultiLine         # edit distance 17894
BinPackArguments: true                             # edit distance 17894
BinPackParameters: true                            # edit distance 17894
BitFieldColonSpacing: Both                         # did not affect the output
BreakBeforeBraces: Custom                          # edit distance 14810
BraceWrapping:
  AfterCaseLabel: true                             # inherited from prefix BraceWrapping
  AfterClass: true                                 # edit distance 17894
  AfterControlStatement: Never                     # edit distance 15196
  AfterEnum: true                                  # edit distance 15196
  AfterFunction: true                              # edit distance 15196
  AfterNamespace: true                             # edit distance 15196
  AfterObjCDeclaration: true                       # inherited from prefix BraceWrapping
  AfterStruct: true                                # edit distance 15196
  AfterUnion: true                                 # inherited from prefix BraceWrapping
  AfterExternBlock: true                           # edit distance 15196
  BeforeCatch: false                               # edit distance 15135
  BeforeElse: false                                # edit distance 14790
  BeforeLambdaBody: false                          # edit distance 14790
  BeforeWhile: false                               # edit distance 14790
  IndentBraces: false                              # edit distance 14790
  SplitEmptyFunction: true                         # edit distance 14790
  SplitEmptyRecord: true                           # inherited from prefix BraceWrapping
  SplitEmptyNamespace: true                        # inherited from prefix BraceWrapping
BreakAfterJavaFieldAnnotations: true               # did not affect the output
BreakBeforeBinaryOperators: None                   # edit distance 14790
BreakBeforeConceptDeclarations: true               # did not affect the output
BreakBeforeTernaryOperators: true                  # did not affect the output
BreakConstructorInitializers: BeforeColon          # edit distance 14790
BreakInheritanceList: AfterComma                   # edit distance 14757
BreakStringLiterals: true                          # did not affect the output
ColumnLimit: 80                                    # edit distance 12432
CompactNamespaces: true                            # did not affect the output
ConstructorInitializerIndentWidth: 4               # edit distance 12432
ContinuationIndentWidth: 4                         # edit distance 12432
Cpp11BracedListStyle: true                         # edit distance 12432
DeriveLineEnding: true                             # inherited from prefix Derive
DerivePointerAlignment: true                       # edit distance 12301
EmptyLineAfterAccessModifier: Leave                # edit distance 12300
EmptyLineBeforeAccessModifier: Leave               # edit distance 12299
ExperimentalAutoDetectBinPacking: true             # edit distance 12284
FixNamespaceComments: false                        # edit distance 12258
IncludeBlocks: Preserve                            # edit distance 12258
IndentAccessModifiers: false                       # edit distance 12258
IndentCaseBlocks: false                            # inherited from prefix Indent
IndentCaseLabels: false                            # edit distance 12258
IndentExternBlock: AfterExternBlock                # did not affect the output
IndentGotoLabels: false                            # inherited from prefix Indent
IndentPPDirectives: None                           # edit distance 12258
IndentRequires: false                              # inherited from prefix Indent
IndentWidth: 4                                     # edit distance 12258
IndentWrappedFunctionNames: false                  # edit distance 12258
InsertTrailingCommas: None                         # did not affect the output
KeepEmptyLinesAtTheStartOfBlocks: true             # edit distance 12258
LambdaBodyIndentation: OuterScope                  # edit distance 12230
MaxEmptyLinesToKeep: 2                             # edit distance 12188
NamespaceIndentation: None                         # edit distance 12188
PackConstructorInitializers: BinPack               # edit distance 12188
PenaltyBreakAssignment: 2                          # edit distance 12188
PenaltyBreakBeforeFirstCallParameter: 64           # edit distance 12175
PenaltyBreakComment: 2                             # inherited from prefix PenaltyBreak
PenaltyBreakFirstLessLess: 2                       # inherited from prefix PenaltyBreak
PenaltyBreakOpenParenthesis: 16                    # edit distance 12152
PenaltyBreakString: 128                            # edit distance 12152
PenaltyBreakTemplateDeclaration: 2                 # inherited from prefix PenaltyBreak
PenaltyExcessCharacter: 64                         # edit distance 12118
PenaltyIndentedWhitespace: 4                       # edit distance 12208
PenaltyReturnTypeOnItsOwnLine: 32                  # edit distance 12208
PointerAlignment: Left                             # edit distance 12207
QualifierAlignment: Leave                          # edit distance 12207
ReferenceAlignment: Pointer                        # did not affect the output
RemoveBracesLLVM: false                            # edit distance 12207
ReflowComments: false                              # edit distance 12129
RequiresClausePosition: OwnLine                    # did not affect the output
SeparateDefinitionBlocks: Leave                    # edit distance 12129
ShortNamespaceLines: 0                             # did not affect the output
SortIncludes: Never                                # edit distance 12119
SortUsingDeclarations: true                        # did not affect the output
SpaceAfterCStyleCast: false                        # edit distance 12119
SpaceAfterLogicalNot: false                        # edit distance 12119
SpaceAfterTemplateKeyword: false                   # edit distance 12117
SpaceAroundPointerQualifiers: Default              # did not affect the output
SpaceBeforeAssignmentOperators: true               # edit distance 12117
SpaceBeforeCaseColon: false                        # edit distance 12117
SpaceBeforeCpp11BracedList: false                  # edit distance 12117
SpaceBeforeCtorInitializerColon: true              # edit distance 12117
SpaceBeforeInheritanceColon: true                  # edit distance 12117
SpaceBeforeParens: Custom                          # edit distance 12117
SpaceBeforeParensOptions:
  AfterControlStatements: true                     # edit distance 12117
  AfterForeachMacros: false                        # inherited from prefix SpaceBefore
  AfterFunctionDeclarationName: false              # edit distance 12117
  AfterFunctionDefinitionName: false               # edit distance 12117
  AfterIfMacros: false                             # inherited from prefix SpaceBefore
  AfterOverloadedOperator: false                   # edit distance 12117
  AfterRequiresInClause: false                     # inherited from prefix SpaceBefore
  AfterRequiresInExpression: false                 # inherited from prefix SpaceBefore
  BeforeNonEmptyParentheses: false                 # edit distance 12117
SpaceBeforeRangeBasedForLoopColon: true            # edit distance 12117
SpaceBeforeSquareBrackets: false                   # edit distance 12117
SpaceInEmptyBlock: false                           # edit distance 12117
SpaceInEmptyParentheses: false                     # edit distance 12117
SpacesBeforeTrailingComments: 1                    # edit distance 12117
SpacesInAngles: Leave                              # edit distance 12109
SpacesInCStyleCastParentheses: false               # edit distance 12109
SpacesInConditionalStatement: false                # edit distance 12109
SpacesInContainerLiterals: true                    # did not affect the output
SpacesInParentheses: false                         # edit distance 12109
Standard: c++11                                    # edit distance 12109
TabWidth: 0                                        # did not affect the output
UseCRLF: true                                      # did not affect the output
UseTab: Never                                      # edit distance 12109
```

erin-le:
This PR is being closed for inactivity. If someone would like to pick this up in the future, please reopen it or ping someone who has the permissions to reopen it.

PR #364, opened 2023-09-27
arch-arm: Remove Jazelle state + ThumbEE support

giactra:
This PR removes Jazelle state (while still keeping a "Trivial Jazelle implementation",
see Arm Architecture Reference Manual) and ThumbEE support

### DIFF
--- a/src/arch/arm/pcstate.hh
+++ b/src/arch/arm/pcstate.hh
@@ -75,7 +75,6 @@ class PCState : public GenericISA::UPCState<4>
     enum FlagBits
     {
         ThumbBit = (1 << 0),
-        JazelleBit = (1 << 1),
         AArch64Bit = (1 << 2)
     };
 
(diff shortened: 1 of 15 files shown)


andysan (dismissed):
No idea what's happening to the tests, but the changes look good to me.

giactra:
The following test is failing:

`Test: test-checkpoint-arm-hello-restore-checkpoint-ALL-x86_64-opt Failed`

I presume changing the PCState will break checkpoint restore for a previously taken one...

src/arch/arm/pcstate.hh
```diff
@@ -75,8 +75,7 @@ class PCState : public GenericISA::UPCState<4>
     enum FlagBits
     {
         ThumbBit = (1 << 0),
-        JazelleBit = (1 << 1),
-        AArch64Bit = (1 << 2)
+        AArch64Bit = (1 << 1)
```
andysan: This change is probably breaking checkpointing.

andysan:
> The following test is failing:
> 
> `Test: test-checkpoint-arm-hello-restore-checkpoint-ALL-x86_64-opt Failed`
> 
> I presume changing the PCState will break checkpoint restore for a previously taken one...

That makes sense. I would expect that the change of the aarch64 flag is the culprit here.

giactra:
Hi @Harshil2107, I have been trying to locate https://resources.gem5.org/resources/arm-hello-test-checkpoint/raw?database=gem5-resources&version=1.0.0 in the gem5-resources project but I couldn't find the JSON entry (https://github.com/gem5/gem5-resources/blob/stable/resources.json). How is that possible?

Harshil2107:
The JSON entry is on the MongoDB database. The `resources.json` is the old way to store resources, it is no longer updated and we keep it there so it works with older versions of gem5. New resources are pushed to the MongoDB database.

Harshil2107:
To update the checkpoint, should the checkpoint updater (util/cpt_upgraders)  be changed?

giactra:
> To update the checkpoint, should the checkpoint updater (util/cpt_upgraders) be changed?

Thanks. At the end I decided to modify my PR to make it compatible with previous checkpoints (basically I am not shifting the AArch64Bit in the PCState... )

andysan (approved):

PR #365, opened 2023-09-27
misc: Update gem5 to use clang-15 and clang-16

### LINKED ISSUE #292: Add clang 15 and clang 16 to the compiler tests
Currently, our compiler tests don't check versions 15 and 16 of clang, so we need to create new Docker images for this and make sure they're added to our tests.  In addition, once these are added, we can remove older versions of clang as to not overrun our testing infrastructure.

mkjost0:
This introduces the changes necessary for clang-15 and clang-16 to run within gem5, and adds them to the compiler tests.

This also updates the dockerfiles for ubuntu 22.04 to include the steps necessary to compile clang-15 and clang-16.

### DIFF
--- a/src/cpu/o3/rename.cc
+++ b/src/cpu/o3/rename.cc
@@ -286,7 +286,7 @@ Rename::setActiveThreads(std::list<ThreadID> *at_ptr)
 
 
 void
-Rename::setRenameMap(UnifiedRenameMap rm_ptr[])
+Rename::setRenameMap(UnifiedRenameMap rm_ptr[MaxThreads])
 {
     for (ThreadID tid = 0; tid < numThreads; tid++)
         renameMap[tid] = &rm_ptr[tid];
--- a/.github/workflows/compiler-tests.yaml
+++ b/.github/workflows/compiler-tests.yaml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image: [gcc-version-12, gcc-version-11, gcc-version-10, gcc-version-9, gcc-version-8, clang-version-14, clang-version-13, clang-version-12, clang-version-11, clang-version-10, clang-version-9, clang-version-8, clang-version-7, ubuntu-20.04_all-dependencies, ubuntu-22.04_all-dependencies, ubuntu-22.04_min-dependencies]
+        image: [gcc-version-12, gcc-version-11, gcc-version-10, gcc-version-9, gcc-version-8, clang-version-16, clang-version-15, clang-version-14, clang-version-13, clang-version-12, clang-version-11, clang-version-10, clang-version-9, clang-version-8, clang-version-7, ubuntu-20.04_all-dependencies, ubuntu-22.04_all-dependencies, ubuntu-22.04_min-dependencies]
         opts: [.opt, .fast]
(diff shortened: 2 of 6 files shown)


mkjost0:
Note: This should not be merged until the new docker images for clang-15 and clang-16 have been created.

util/dockerfiles/ubuntu-22.04_clang-version/Dockerfile
```diff
@@ -27,6 +27,7 @@ FROM ubuntu:22.04
 
 # Valid version values:
 # 13
+# 15
```
BobbyRBruce, line 30: Might as well add `#14` here.

util/dockerfiles/ubuntu-22.04_clang-version/clang-16/Dockerfile
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) 2022 The Regents of the University of California
```
BobbyRBruce: Update to 2023

util/dockerfiles/ubuntu-22.04_clang-version/clang-16/Dockerfile
BobbyRBruce, line 1: I take it this is necessary as you can't install clang 16 in 22.04?

If this is the case can we put this in something like "util/dockerfiles/ubuntu-22.04_clang-16"?

BobbyRBruce (changes requested):
You need to update [util/dockerfiles/docker-compose.yaml](https://github.com/gem5/gem5/blob/62d34ef37460e9e8cbbbaeb50605572a861565b9/util/dockerfiles/docker-compose.yaml) for these new Docker image targets.

src/cpu/o3/rename.cc
```diff
@@ -286,7 +286,7 @@ Rename::setActiveThreads(std::list<ThreadID> *at_ptr)
 
 
 void
-Rename::setRenameMap(UnifiedRenameMap rm_ptr[])
+Rename::setRenameMap(UnifiedRenameMap rm_ptr[MaxThreads])
```
andysan, line 289: It might be cleaner to just introduce a `PerThreadUnifiedRenameMap` type (or something along those lines) that is a `std::array<UnifiedRenameMap, MaxThreads>`. That would make it behave as a normal C++ container but with the performance of a plain array. I realise this is a slightly bigger change.

util/dockerfiles/ubuntu-22.04_clang-version/clang-16/Dockerfile
mkjost0: Yeah I can update this, I couldn't get clang 16 to work following the steps that we used for past versions.

src/cpu/o3/rename.cc
mkjost0: Right now, we want to prioritize getting this change in so we can test the new compilers in our weekly tests, but I introduced this in an issue (#373 ).  I don't have the time to make this change right now, but this way it can still be addressed.

BobbyRBruce (approved):
I added a commit to this PR fixing some things. 

I've uploaded the two new images. They should work.

powerjg (approved):

Harshil2107 (approved):

PR #133, opened 2023-07-27
cpu: Set SLC bit for GPU tester

abmerop:
This fixes issue #131 by reverting to the old behavior of performing all atomics at the system level. To do this the SLC bit needs to be set for all atomic requests.

Change-Id: I63f4e449be1b02c933832d09700237f8c8026f4c

PR #367, opened 2023-09-28
mem-ruby: Always pass on GPU atomics to dir in write-through TCC

DanKouch:
Added checks to ensure that atomics are not performed in the TCC when it is configured as a write-through cache. Also added SLC bit overwrite to ensure directory preforms atomics when there is a write-through TCC.

Change-Id: I4514e6c8022aeb7785f2c59871cd9acec8161ed8

### DIFF
--- a/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
@@ -299,16 +299,22 @@ machine(MachineType:TCC, "TCC Cache")
         // checked when the read response is received.
         if (in_msg.Type == CoherenceResponseType:NBSysWBAck) {
           trigger(Event:WBAck, in_msg.addr, cache_entry, tbe);
-        } else if (is_slc_set) {
-            // If the SLC bit is set, the response needs to bypass the cache
-            // and should not be allocated an entry.
+        } else if(in_msg.Type == CoherenceResponseType:NBSysResp) {
+          // If the SLC bit is set or the cache is write-through and
+          // we're receiving modified data (such as from an atomic),
+          // the response needs to bypass the cache and should not be
+          // allocated an entry.
+          if(is_slc_set || (!WB && in_msg.State == CoherenceState:Modified)) {
             trigger(Event:Bypass, in_msg.addr, cache_entry, tbe);
-        } else if (in_msg.Type == CoherenceResponseType:NBSysResp) {
-          if(presentOrAvail(in_msg.addr)) {
-            trigger(Event:Data, in_msg.addr, cache_entry, tbe);
           } else {
-            Addr victim :=  L2cache.cacheProbe(in_msg.addr);
-            trigger(Event:L2_Repl, victim, getCacheEntry(victim), TBEs.lookup(victim));
+            if(presentOrAvail(in_msg.addr)) {
+              // Responses with atomic data will only reach here if the
+              // SLC bit isn't set and the cache is WB
+              trigger(Event:Data, in_msg.addr, cache_entry, tbe);
+            } else {
+              Addr victim :=  L2cache.cacheProbe(in_msg.addr);
+              trigger(Event:L2_Repl, victim, getCacheEntry(victim), TBEs.lookup(victim));
+            }
           }
         } else {
           error("Unexpected Response Message to Core");
@@ -357,13 +363,13 @@ machine(MachineType:TCC, "TCC Cache")
                 trigger(Event:WrVicBlk, in_msg.addr, cache_entry, tbe);
             }
         } else if (in_msg.Type == CoherenceRequestType:Atomic) {
-          // If the request is system-level, or if the address isn't in the cache,
-          // then send the request to the directory. Since non-SLC atomics won't be
-          // performed by the directory, TCC will perform the atomic on the return path
-          // on Event:Data.
+          // If the request is system-level, if the address isn't in the cache,
+          // or if this cache is write-through, then send the request to the
+          // directory. Since non-SLC atomics won't be performed by the directory,
+          // TCC will perform the atomic on the return path on Event:Data.
           // The action will invalidate the cache line if SLC is set and the address is
           // in the cache.
-          if(in_msg.isSLCSet || !presentOrAvail(in_msg.addr)) {
+          if(in_msg.isSLCSet || !WB || !presentOrAvail(in_msg.addr)) {
             trigger(Event:AtomicPassOn, in_msg.addr, cache_entry, tbe);
           } else {
             trigger(Event:Atomic, in_msg.addr, cache_entry, tbe);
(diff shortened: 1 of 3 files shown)


DanKouch:
@mattsinc @abmerop This should address the atomic bug exposed with Pannotia BC/PageRank.

DanKouch:
Had to make some quick changes to fix a potential case where atomics could be performed twice, but this PR should be ready to review now.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -299,16 +299,22 @@ machine(MachineType:TCC, "TCC Cache")
         // checked when the read response is received.
         if (in_msg.Type == CoherenceResponseType:NBSysWBAck) {
           trigger(Event:WBAck, in_msg.addr, cache_entry, tbe);
-        } else if (is_slc_set) {
-            // If the SLC bit is set, the response needs to bypass the cache
-            // and should not be allocated an entry.
+        } else if(in_msg.Type == CoherenceResponseType:NBSysResp) {
```
mattsinc, line 302: I am not sure why this is correct -- now we only check the SLC bit if it's a response from the directory, why is that correct?  I can understand why on a response we'd need to check if it's a WT atomic, but not why SLC can only be checked inside it.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -640,7 +646,9 @@ machine(MachineType:TCC, "TCC Cache")
         out_msg.Dirty := true;
         out_msg.writeMask.orMask(in_msg.writeMask);
         out_msg.isGLCSet := in_msg.isGLCSet;
-        out_msg.isSLCSet := in_msg.isSLCSet;
+        // Overwrite the SLC bit to 1 if the cache is write-through so that the
```
mattsinc: I can see why you did this, because if we're using WT then SLC and WT+atomic are treated the same.  But I suspect it could lead to some unintended behavior down the line when that may not be the case.  I suppose the alternative is adding a field to the Network packet for WT?

mattsinc (commented):
Also, have you tested this with the Pannotia apps, or with the Ruby Random tester with WT caches?

DanKouch:
> Also, have you tested this with the Pannotia apps, or with the Ruby Random tester with WT caches?

Yes, I've tested with Pannotia BC and the Ruby Random Tester with WT.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
DanKouch: The SLC bit isn't being set in the response path, so I don't think there would be any impact currently, but I agree it's not the cleanest solution. I suppose another solution would be to just pass WB_L2 to the directory like how WB is being passed to the TCC in its configuration.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -357,13 +363,13 @@ machine(MachineType:TCC, "TCC Cache")
...
-          // performed by the directory, TCC will perform the atomic on the return path
-          // on Event:Data.
+          // If the request is system-level, if the address isn't in the cache,
+          // or if this cache is write-through, then send the request to the
+          // directory. Since non-SLC atomics won't be performed by the directory,
+          // TCC will perform the atomic on the return path on Event:Data.
           // The action will invalidate the cache line if SLC is set and the address is
           // in the cache.
-          if(in_msg.isSLCSet || !presentOrAvail(in_msg.addr)) {
+          if(in_msg.isSLCSet || !WB || !presentOrAvail(in_msg.addr)) {
```
abmerop, line 372: May have asked this before, but this still confuses me.  If it's not in the cache we send to directory regardless?  Wouldn't that mean *everything* goes to the directory unless something did a load first, even if it's a GLC atomic?

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -299,16 +299,22 @@ machine(MachineType:TCC, "TCC Cache")
...
             trigger(Event:Bypass, in_msg.addr, cache_entry, tbe);
-        } else if (in_msg.Type == CoherenceResponseType:NBSysResp) {
-          if(presentOrAvail(in_msg.addr)) {
-            trigger(Event:Data, in_msg.addr, cache_entry, tbe);
           } else {
-            Addr victim :=  L2cache.cacheProbe(in_msg.addr);
-            trigger(Event:L2_Repl, victim, getCacheEntry(victim), TBEs.lookup(victim));
+            if(presentOrAvail(in_msg.addr)) {
+              // Responses with atomic data will only reach here if the
+              // SLC bit isn't set and the cache is WB
```
abmerop, line 312: If they can only reach here in those situations, should add an assert for both cases

abmerop (commented):
I have tested this with GPUFS and it does work with BC (gets to 1024 iterations / converges as expected) and pagerank. However, I have a few questions below on if it's working for the right reasons

Regarding the protocol tester, have you tested with #133 reverted?  Currently the tester is forcing SLC too, which is known to work.  We'd really like to test GLC

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
mattsinc: this is a good point

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
mattsinc: it looks like this change remains?

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
DanKouch: @abmerop Yes, I have tested this on the Ruby Random Tester with #133 reverted on WT caches. I'm currently working on a patch that will allow us to revert #133 for WB caches.

For your second question, we'll pass WB GLC atomic requests to the directory, but the directory won't perform the atomic itself; the TCC will perform it on the return path in transition(A, Data, M).

abmerop: Ah, thanks I see now.  This patch looks good to me then.  If you aren't planning to combine it with the patch you are working on I will approve

DanKouch:
Updated to gather WT L2 behavior in the directory from passed parameter, rather than by overwriting the SLC bit of atomic instructions in the TCC.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
mattsinc: @abmerop do you still want this?

mattsinc (approved):
From my end this is fine, although @abmerop still has one request pending that he'll need to ok/request further changes for

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
abmerop: Yes, please add this.  It will help later if the protocol has further changes and doesn't consider the comment

mattsinc: Sounds good, @DanKouch let us know when this is done

abmerop: I'll drop this as the if conditional above checks it.  Restarted the tests because I don't see how modifying a gpu ruby protocol caused arm hello world tests to fail. Should be able to submit after that

PR #369, opened 2023-09-28
python: Add importer to standalone gem5py_m5

powerjg:
I believe the point of this binary was to allow people to use the m5 objects without the entire gem5 binary. However, without adding the importer call, this did not work. Unfortunately, with the importer call there is a circular dependence on the original gem5py.cc file. Therefore, this change creates a new file that has the importer call.

Now, with the `gem5py_m5` binary you can run python code that references modules in `src/python`. Note that `_m5` is not available, so anything that depends on the gem5 SimObjects' implementation will not work. However, thic can still be useful for things like getting Resources, processing stats, etc.

Change-Id: I5c0e5d1a669fe5ce491458df916f2049c81292eb

### DIFF
--- a/src/python/gem5py_m5.cc
+++ b/src/python/gem5py_m5.cc
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2019 ARM Limited
+ * All rights reserved
+ *
+ * The license below extends only to copyright in the software and shall
+ * not be construed as granting a license to any other intellectual
+ * property including but not limited to intellectual property relating
+ * to a hardware implementation of the functionality of the software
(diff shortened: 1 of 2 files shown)


src/python/gem5py_m5.cc
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (c) 2019 ARM Limited
+ * All rights reserved
+ *
```
BobbyRBruce, line 4: Not a blocker but you're not giving yourself credit here. If based on "gem5py.cc" you've had to hack quite a bit of this to get it working.

BobbyRBruce (approved):

PR #370, opened 2023-09-29
arch-riscv: Update FS bits when doing floating point loads

### LINKED ISSUE #349: gem5 RISCV Full System Simulation w/ Ubuntu 22.04 Failed to Execute STREAM+OpenMP+clang-14/gcc-12 properly
**Describe the bug**
gem5 RISCV Full System Simulation w/ Ubuntu 22.04 Failed to Execute STREAM+OpenMP+clang14 properly using STREAM with 1 thread and gem5 in atomic memory mode.

**Affects version**
develop for v23.1
https://github.com/gem5/gem5/tree/f5a255c68d1ce0318da0d46f744ee7513e0acda9

**gem5 Modifications**
No

**To Reproduce**
Step 1. Create the Ubuntu 22.04 disk image as showed here, https://github.com/hnpl/hpc-disk-image#4-building-the-rv64gc-disk-image.
Step 2. In the guest system, install gcc-12 or clang-14 and the related tools.
Step 3. Recompiling the STREAM benchmark in the guest `/home/ubuntu/simple-vectorizable-microbenchmarks/stream` with the following command,
```sh
(sudo?) make -f makefiles/Makefile-hw M5_BUILD_PATH=/home/ubuntu/gem5/util/m5/build/riscv/ M5OPS_HEADER_PATH=/home/ubuntu/gem5/include/
```
Optionally, if you're using clang, you'd need to modify the file `/home/ubuntu/simple-vectorizable-microbenchmarks/stream/makefiles/Makefile-hw`.
Step 4. Run the following command,
```sh
build/RISCV/gem5.opt -re --outdir=atomic-500000 riscv-fs-test.py --command "/home/ubuntu/simple-vectorizable-microbenchmarks/stream/stream.rv64gc.m5 500000"
```

gem5 Python configuration,

```Python
import m5

from gem5.components.cachehierarchies.classic.private_l1_private_l2_cache_hierarchy import (
)
from gem5.isas import ISA
from gem5.utils.override import overrides
from gem5.resources.resource import Resource, CustomResource, DiskImageResource, FileResource

# Run a check to ensure the right version of gem5 is being used.

# Setup the cache hierarchy.
# For classic, PrivateL1PrivateL2 and NoCache have been tested.
# For Ruby, MESI_Two_Level and MI_example have been tested.
    l1d_size="32KiB", l1i_size="32KiB", l2_size="512KiB"
)

# Setup the system memory.
memory = SingleChannelDDR3_1600()

# Setup a single core Processor.
processor = SimpleProcessor(
    cpu_type=CPUTypes.ATOMIC, isa=ISA.RISCV, num_cores=1
)

class ModifiedRiscvBoard(RiscvBoard):
    @overrides(RiscvBoard)
    def get_default_kernel_args(self):
        return [
            "earlyprintk=ttyS0",
            "console=ttyS0",
            "lpj=7999923",
            "root=/dev/vda1",
            "init=/root/gem5-init.sh",
            "rw",
        ]

# Setup the board.
board = ModifiedRiscvBoard(
    clk_freq="1GHz",
    processor=processor,
    memory=memory,
)


import argparse
parser = argparse.ArgumentParser()
parser.add_argument("--command", type=str)
args = parser.parse_args()

command = args.command

    kernel=obtain_resource("riscv-bootloader-vmlinux-5.10"),
    disk_image=DiskImageResource("/scr/hn/rv64gc-hpc-2204.img"),
    readfile_contents=f"{command}"
)

#m5.debug.flags["ExecAll"].enable()
#m5.debug.flags["LLSC"].enable()
simulator.run()
print("exit 1")
m5.debug.flags["ExecAll"].enable()
m5.debug.flags["LLSC"].enable()
#m5.debug.flags["RiscvMiscFS"].enable()
simulator.run()
print("exit 2")
#simulator.run()
#print("exit 3")

```

**Terminal Output**
N/A

**Expected behavior**
[Expection] The `board.platform.terminal` file should show,

```
Number of threads: 1
...
array_a error_count: 0, max_error: 0.000000
array_b error_count: 0, max_error: 0.000000
array_c error_count: 0, max_error: 0.000000

```

[Reality] The `board.platform.terminal` file shows non-zero error_counts, e.g.,
```
Number of threads: 1
...
array_a error_count: 352070, max_error: 15.000000
array_b error_count: 352070, max_error: 3.000000
array_c error_count: 352070, max_error: 4.000000

```

**Host Operating System**
Ubuntu 22.04

**Host ISA**
X86

**Compiler used**
N/A

hnpl:
This problem is similar to the problem described in [1]. This problem produces symptoms as described in [2].

In short, the Linux kernel relies on the CSR_STATUS's FS bits to decide whether to save the floating point registers. If the FS bits are set to DIRTY, the floating point registers will be saved during context switching / task switching.

Currently, with the patch in [1], we only change the FS bits upon every floating arithmetic instruction. However, since floating load instructions also mutate the state of floating point registers, the FS bits should be updated to DIRTY.

The problem in [2] arose when the program populates the content of one floating register to an array by repeatedly using `fld fa5, EA`. A context switch occured upon a page fault, and while handling that page fault, the kernel might have to handle an interrupt. This caused the kernel to task switch between handling page fault and handling interrupt. This caused __switch_to() to be called, which will save the floating point registers only if the SD (indirectly set by FS) bits are set to DIRTY, while restoring the floating point registers to the switch-to task [3]. This caused the floating point registers to be zeroed out when it was restored as it was never saved before.

[1] https://gem5-review.googlesource.com/c/public/gem5/+/65272
[2] https://github.com/gem5/gem5/issues/349
[3] https://github.com/torvalds/linux/blob/v6.5/arch/riscv/include/asm/switch_to.h#L56

Change-Id: Ia5656da5a589a8e29fb699d2ee12885b8f3fa2d2

### DIFF
--- a/src/arch/riscv/isa/decoder.isa
+++ b/src/arch/riscv/isa/decoder.isa
@@ -61,6 +61,11 @@ decode QUADRANT default Unknown::unknown() {
                     return std::make_shared<IllegalInstFault>("FPU is off",
                                                                machInst);
 
+                // Mutating any floating point register changes the FS bit
+                // of the STATUS CSR.
+                status.fs = FPUStatus::DIRTY;
+                xc->setMiscReg(MISCREG_STATUS, status);
+
                 Fp2_bits = Mem;
             }}, {{
                 EA = rvZext(Rp1 + offset);
@@ -312,6 +317,11 @@ decode QUADRANT default Unknown::unknown() {
                     return std::make_shared<IllegalInstFault>("FPU is off",
                                                                machInst);
 
+                // Mutating any floating point register changes the FS bit
+                // of the STATUS CSR.
+                status.fs = FPUStatus::DIRTY;
+                xc->setMiscReg(MISCREG_STATUS, status);
+
                 Fc1_bits = Mem;
             }}, {{
                 EA = rvZext(sp + offset);
@@ -495,6 +505,12 @@ decode QUADRANT default Unknown::unknown() {
                     if (status.fs == FPUStatus::OFF)
                         return std::make_shared<IllegalInstFault>(
                                     "FPU is off", machInst);
+
+                    // Mutating any floating point register changes the FS bit
+                    // of the STATUS CSR.
+                    status.fs = FPUStatus::DIRTY;
+                    xc->setMiscReg(MISCREG_STATUS, status);
+
                     freg_t fd;
                     fd = freg(f64(Mem));
                     Fd_bits = fd.v;


aarmejach (approved):
LGTM

BobbyRBruce (approved):

rogerchang23424:
Should we also need to consider the instruction `flw`, `flh`, `c.flw` and `c.flwsp`?

hnpl:
Definitely! Thanks for pointing this out. I'll make a patch accordingly.

PR #376, opened 2023-09-29
arch-riscv: Change to VS bits to DIRTY for rvv insts changing vregs

hnpl:
This is similar to [1] and [2].

Essentially, the VS bits of STATUS CSR keep track of the state of
the vector registers. (VS bits == DIRTY) means the content of vector
registers have been updated since the last time the VS bits were updated.

This chain of changes is supposed to change the VS bits to DIRTY for if any
vector register is potentially updated.

[1] https://gem5-review.googlesource.com/c/public/gem5/+/65272
[2] https://github.com/gem5/gem5/pull/370

Change-Id: I0427890dadc63b74a470d7405807dcfcad18005b

### DIFF
--- a/src/arch/riscv/isa/templates/vector_arith.isa
+++ b/src/arch/riscv/isa/templates/vector_arith.isa
@@ -168,6 +168,9 @@ Fault
     if (machInst.vill)
         return std::make_shared<IllegalInstFault>("VILL is set", machInst);
 
+    status.vs = VPUStatus::DIRTY;
+    xc->setMiscReg(MISCREG_STATUS, status);
+
     %(op_decl)s;
     %(op_rd)s;
     %(vm_decl_rd)s;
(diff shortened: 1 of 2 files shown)


aarmejach (commented):
LGTM. This will now enable to test kernel v6.5 with RVV support! :)

aarmejach (approved):

powerjg (approved):

powerjg:
@hnpl, any reason not to remove the "draft" status?

hnpl:
I have not tested this.

rogerchang23424 (approved):

hnpl:
I tested this with the STREAM benchmark with array size of 5 x 10^6 in FS mode and 1 CPU. I didn't see any errors. Since there's only one CPU, task switching should happen frequently, and it appears that the VS bits were correctly set.

PR #69, opened 2023-07-12
scons: Change to Kconfig build system

rogerchang23424:
The PR contains the following changes:
- Move all of the config options(`env["CONF"]`) from SConsopt to Kconfig files
- Update `build_opts` files to Kconfig option formats
- The Ruby Protocol files are only built if `RUBY=y`
- Remove the default-default build target
- Kconfig commands are included in the PR:
    - defconfig
    - setconfig
    - meunconfig
    - guiconfig
    - listnewconfig
    - savedefconfig
    - oldconfig
    - olddefconfig
- Add the `python3-tk` package dependencies
 
Jira issue: https://gem5.atlassian.net/browse/GEM5-1211

PR #377, opened 2023-09-29
dev-amdgpu,mem-ruby: Add support to checkpoint and restore between kernels in GPUFS

v-ramadas:
Earlier, GPU checkpointing was working only if a checkpoint was created before the first kernel execution. This pull request adds support to checkpoint in-between any two kernel calls. It does so by doing the following.

- Adds flush support in the GPU_VIPER protocol
- Adds flush support in the GPUCoalescer
- Updates cache recorder to use the GPUCoalescer during simulation cooldown and cache warmup times.

### DIFF
--- a/src/dev/amdgpu/pm4_packet_processor.cc
+++ b/src/dev/amdgpu/pm4_packet_processor.cc
@@ -1044,6 +1044,7 @@ PM4PacketProcessor::serialize(CheckpointOut &cp) const
     int num_queues = queues.size();
     Addr id[num_queues];
     Addr mqd_base[num_queues];
+    uint64_t mqd_read_index[num_queues];
     Addr base[num_queues];
     Addr rptr[num_queues];
     Addr wptr[num_queues];
@@ -1060,6 +1061,7 @@ PM4PacketProcessor::serialize(CheckpointOut &cp) const
     uint32_t hqd_active[num_queues];
     uint32_t hqd_vmid[num_queues];
     Addr aql_rptr[num_queues];
+    uint32_t aql[num_queues];
     uint32_t doorbell[num_queues];
     uint32_t hqd_pq_control[num_queues];
 
@@ -1068,9 +1070,10 @@ PM4PacketProcessor::serialize(CheckpointOut &cp) const
         PM4Queue *q = iter.second;
         id[i] = q->id();
         mqd_base[i] = q->mqdBase();
+        mqd_read_index[i] = q->getMQD()->mqdReadIndex;
         bool cur_state = q->ib();
         q->ib(false);
-        base[i] = q->base() >> 8;
+        base[i] = q->base();
         rptr[i] = q->getRptr();
         wptr[i] = q->getWptr();
         q->ib(true);
@@ -1088,6 +1091,7 @@ PM4PacketProcessor::serialize(CheckpointOut &cp) const
         hqd_active[i] = q->getMQD()->hqd_active;
         hqd_vmid[i] = q->getMQD()->hqd_vmid;
         aql_rptr[i] = q->getMQD()->aqlRptr;
+        aql[i] = q->getMQD()->aql;
         doorbell[i] = q->getMQD()->doorbell;
         hqd_pq_control[i] = q->getMQD()->hqd_pq_control;
         i++;
@@ -1096,6 +1100,7 @@ PM4PacketProcessor::serialize(CheckpointOut &cp) const
     SERIALIZE_SCALAR(num_queues);
     SERIALIZE_ARRAY(id, num_queues);
     SERIALIZE_ARRAY(mqd_base, num_queues);
+    SERIALIZE_ARRAY(mqd_read_index, num_queues);
     SERIALIZE_ARRAY(base, num_queues);
     SERIALIZE_ARRAY(rptr, num_queues);
     SERIALIZE_ARRAY(wptr, num_queues);
@@ -1112,6 +1117,7 @@ PM4PacketProcessor::serialize(CheckpointOut &cp) const
     SERIALIZE_ARRAY(hqd_active, num_queues);
     SERIALIZE_ARRAY(hqd_vmid, num_queues);
     SERIALIZE_ARRAY(aql_rptr, num_queues);
+    SERIALIZE_ARRAY(aql, num_queues);
     SERIALIZE_ARRAY(doorbell, num_queues);
     SERIALIZE_ARRAY(hqd_pq_control, num_queues);
 }
@@ -1127,6 +1133,7 @@ PM4PacketProcessor::unserialize(CheckpointIn &cp)
 
     Addr id[num_queues];
     Addr mqd_base[num_queues];
+    uint64_t mqd_read_index[num_queues];
     Addr base[num_queues];
     Addr rptr[num_queues];
     Addr wptr[num_queues];
@@ -1143,11 +1150,13 @@ PM4PacketProcessor::unserialize(CheckpointIn &cp)
     uint32_t hqd_active[num_queues];
     uint32_t hqd_vmid[num_queues];
     Addr aql_rptr[num_queues];
+    uint32_t aql[num_queues];
     uint32_t doorbell[num_queues];
     uint32_t hqd_pq_control[num_queues];
 
     UNSERIALIZE_ARRAY(id, num_queues);
     UNSERIALIZE_ARRAY(mqd_base, num_queues);
+    UNSERIALIZE_ARRAY(mqd_read_index, num_queues);
     UNSERIALIZE_ARRAY(base, num_queues);
     UNSERIALIZE_ARRAY(rptr, num_queues);
     UNSERIALIZE_ARRAY(wptr, num_queues);
@@ -1164,6 +1173,7 @@ PM4PacketProcessor::unserialize(CheckpointIn &cp)
     UNSERIALIZE_ARRAY(hqd_active, num_queues);
     UNSERIALIZE_ARRAY(hqd_vmid, num_queues);
     UNSERIALIZE_ARRAY(aql_rptr, num_queues);
+    UNSERIALIZE_ARRAY(aql, num_queues);
     UNSERIALIZE_ARRAY(doorbell, num_queues);
     UNSERIALIZE_ARRAY(hqd_pq_control, num_queues);
 
@@ -1172,19 +1182,20 @@ PM4PacketProcessor::unserialize(CheckpointIn &cp)
         memset(mqd, 0, sizeof(QueueDesc));
 
         mqd->mqdBase = mqd_base[i] >> 8;
-        mqd->base = base[i];
-        mqd->rptr = rptr[i];
-        mqd->ibBase = ib_base[i];
-        mqd->ibRptr = ib_rptr[i];
+        mqd->mqdReadIndex = mqd_read_index[i];
+        mqd->base = base[i] >> 8;
+        mqd->aql = aql[i];
 
         PM4MapQueues* pkt = new PM4MapQueues;
         memset(pkt, 0, sizeof(PM4MapQueues));
         newQueue(mqd, offset[i], pkt, id[i]);
 
         queues[id[i]]->ib(false);
+        queues[id[i]]->rptr(rptr[i]);
         queues[id[i]]->wptr(wptr[i]);
         queues[id[i]]->ib(true);
         queues[id[i]]->wptr(ib_wptr[i]);
+        queues[id[i]]->rptr(ib_rptr[i]);
         queues[id[i]]->offset(offset[i]);
         queues[id[i]]->processing(processing[i]);
         queues[id[i]]->ib(ib[i]);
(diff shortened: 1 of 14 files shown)


src/dev/amdgpu/pm4_packet_processor.cc
```diff
@@ -1164,27 +1173,29 @@ PM4PacketProcessor::unserialize(CheckpointIn &cp)
...
         PM4MapQueues* pkt = new PM4MapQueues;
         memset(pkt, 0, sizeof(PM4MapQueues));
         newQueue(mqd, offset[i], pkt, id[i]);
 
         queues[id[i]]->ib(false);
+        queues[id[i]]->rptr(rptr[i]);
         queues[id[i]]->wptr(wptr[i]);
         queues[id[i]]->ib(true);
         queues[id[i]]->wptr(ib_wptr[i]);
+        queues[id[i]]->rptr(ib_rptr[i]);
```
abmerop: This whole thing seems kind of weird.  All three of these get called twice, which one is correct?  I will also check after this review... I think we need to be seeing if it's an IB (indirect buffer) or not first and using the appropriate read/write pointers and ib boolean

src/dev/amdgpu/pm4_packet_processor.cc
```diff
@@ -1164,27 +1173,29 @@ PM4PacketProcessor::unserialize(CheckpointIn &cp)
     UNSERIALIZE_ARRAY(hqd_active, num_queues);
     UNSERIALIZE_ARRAY(hqd_vmid, num_queues);
     UNSERIALIZE_ARRAY(aql_rptr, num_queues);
+    UNSERIALIZE_ARRAY(aql, num_queues);
     UNSERIALIZE_ARRAY(doorbell, num_queues);
     UNSERIALIZE_ARRAY(hqd_pq_control, num_queues);
 
-    for (int i = 0; i < num_queues; i++) {
+    for (int i = num_queues - 1; i >= 0; i--) {
```
abmerop: Any reason why this is reversed?  It it necessary?

src/dev/amdgpu/pm4_packet_processor.cc
```diff
@@ -1044,6 +1044,7 @@ PM4PacketProcessor::serialize(CheckpointOut &cp) const
     int num_queues = queues.size();
     Addr id[num_queues];
     Addr mqd_base[num_queues];
+    uint64_t mqd_read_index[num_queues];
```
abmerop, line 1047: This is fine for now -  I originally added these and you are following the style around it, I understand.  But at some point we should remove these VLAs (variable length arrays) because I don't think they are standard C++.  It happens to be OK with gcc/clang which gem5 supports but I think other compilers like MSVC will error out

src/mem/ruby/system/GPUCoalescer.cc
```diff
@@ -129,8 +137,16 @@ UncoalescedTable::updateResources()
             instPktsRemaining.erase(seq_num);
 
             // Release the token
-            DPRINTF(GPUCoalescer, "Returning token seqNum %d\n", seq_num);
-            coalescer->getGMTokenPort().sendTokens(1);
+            if (!RubySystem::getWarmupEnabled()
+                    && !RubySystem::getCooldownEnabled()) {
```
abmerop, line 145: For this and other places where we are checking if Ruby is in warmup, we should probably comment why the code is conditioned out

abmerop (commented):
Thanks, this is great work so far!  It will be a huge feature to be able to checkpoint/restore between GPU kernels.  Everything looks correct to me, I just have a few recommendations below

src/mem/ruby/system/GPUCoalescer.cc
v-ramadas: I added a few comments wherever Ruby checks if it is in warmup or cooldown.

src/dev/amdgpu/pm4_packet_processor.cc
v-ramadas: I updated this to write the appropriate read/write pointers to the queue based on the ib field value

src/dev/amdgpu/pm4_packet_processor.cc
v-ramadas: I ran into a tlb related issue when restoring, and wanted to check if it was because of the order in which I restore. Since this diagnosis was incorrect, this change is not needed. I have reverted it to the original version

abmerop:
Thanks for the updates, this looks good to me.  I guess we are just keeping it a draft while deciding how to handle the AVX save/restore issue?

v-ramadas:
I was thinking of either reverting the commit https://github.com/gem5/gem5/pull/113/commits/9acfc5a751c4df408b2fa9f73aa4512392c48c02 or updating the GPUFS README to ask the user to disable these fields before checkpointing until xsave/xrstor instructions are implemented (and the problem is fixed). What would you suggest?

abmerop:
I would just add a command line option to enable/disable AVX, then use that to set use_avx in system.py.  Please don't revert that commit as it will break DNNMark.

v-ramadas:
Thanks for the suggestion! I've pushed a new commit to this PR that adds an option to disable avx. I've also removed the WIP tag and marked the PR ready for review

abmerop (dismissed):
Looks good to me, thanks for all the changes!

mattsinc:
@v-ramadas : this is the error:

src/gpu-compute/gpu_dyn_inst.hh:42:10: fatal error: debug/GPUMem.hh: No such file or directory
   42 | #include "debug/GPUMem.hh"
   
Do things compile for you locally?

src/mem/ruby/protocol/GPU_VIPER-TCP.sm
```diff
@@ -469,6 +474,24 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
...
+      out_msg.Requestor := machineID;
+      assert(is_valid(cache_entry));
+      out_msg.DataBlk := cache_entry.DataBlk;
+      out_msg.writeMask.clear();
+      out_msg.writeMask.orMask(cache_entry.writeMask);
+      out_msg.Destination.add(mapAddressToRange(address,MachineType:TCC,
+                              TCC_select_low_bit, TCC_select_num_bits));
+      out_msg.MessageSize := MessageSizeType:Data;
+      out_msg.Type := CoherenceRequestType:WriteFlush;
+      out_msg.InitialRequestTime := curCycle();
```
mattsinc, line 488: does this need to be curTick?

mattsinc (commented):
We'll have to figure out why the error is happening, otherwise one small question

abmerop:
> We'll have to figure out why the error is happening, otherwise one small question

I think it is due to the ALL build not building the GPU, and the DebugFlags for GPUMem, etc. are guarded in src/gpu-compute/SConscript are guarded by the configuration variable BUILD_GPU.

I assume this means that all GPU patches will now fail tests if that's the case

src/mem/ruby/protocol/GPU_VIPER-TCP.sm
v-ramadas: I set it to curCycle() since other actions that set out_msg.InitialRequestTime also set it to curCycle(). I think this field is used to calculate the latency after the request completes, which might not be important during flush.

v-ramadas:
> @v-ramadas : this is the error:
> 
> src/gpu-compute/gpu_dyn_inst.hh:42:10: fatal error: debug/GPUMem.hh: No such file or directory 42 | #include "debug/GPUMem.hh"
> 
> Do things compile for you locally?

Yes, I am able to compile and execute locally.

src/mem/ruby/system/CacheRecorder.cc
```diff
@@ -30,8 +30,11 @@
 #include "mem/ruby/system/CacheRecorder.hh"
 
 #include "debug/RubyCacheTrace.hh"
+#include "mem/packet.hh"
+#include "mem/ruby/system/GPUCoalescer.hh"
```
abmerop, line 34: To fix the ALL test, I think we need to wrap the GPUCoalescer.hh header with an `#ifdef BUILD_GPU` / `#endif`. 

For example: https://github.com/gem5/gem5/blob/stable/src/mem/ruby/profiler/Profiler.cc#L74

Hopefully this won't be needed if we go forward with #306

v-ramadas: Thank you for pointing this out! I've added similar guards and I am able to compile both VEGA_X86 and X86 locally. Also verified that this does not break checkpointing.

src/mem/ruby/system/CacheRecorder.cc
```diff
@@ -143,13 +180,27 @@ CacheRecorder::enqueueNextFetchRequest()
...
 
             Sequencer* m_sequencer_ptr = m_seq_map[traceRecord->m_cntrl_id];
+#if BUILD_GPU
+            GPUCoalescer* m_coal_ptr;
+            m_coal_ptr = m_coalescer_map[traceRecord->m_cntrl_id];
+#endif
             assert(m_sequencer_ptr != NULL);
+#if BUILD_GPU
+            if (m_coal_ptr == NULL)
+                m_sequencer_ptr->makeRequest(pkt);
```
powerjg, line 190:
```suggestion
            if (m_coal_ptr == NULL) {
                m_sequencer_ptr->makeRequest(pkt);
            }
```

abmerop (approved):

src/mem/ruby/system/CacheRecorder.cc
```diff
@@ -54,14 +57,29 @@ CacheRecorder::CacheRecorder()
 {
 }
 
+#if BUILD_GPU
```
mattsinc, line 60: shouldn't this be #ifdef instead of #if?

src/mem/ruby/system/CacheRecorder.cc
```diff
@@ -81,6 +99,9 @@ CacheRecorder::~CacheRecorder()
         m_uncompressed_trace = NULL;
     }
     m_seq_map.clear();
+#if BUILD_GPU
```
mattsinc, line 102: same here

mattsinc (commented):
If if/ifdef behave the same, maybe fine.  But one question remains.

src/mem/ruby/system/CacheRecorder.cc
v-ramadas: At the start of compile time, configs/build_gpu.hh is generated with a value of either 0 or 1 (if VEGA_X86) for  BUILD_GPU. #ifdef always returns true since BUILD_GPU is defined in either case. I used #if to differentiate between its possible values 0 and 1.

mattsinc: I am not sure I understand the distinction, but ok.

powerjg: I'm unconvinced this was a safe thing to do. It is bad practice to rely on magic values like this! 

In fact, I'm pretty sure that #69 will break this change.

I'm quite concerned about adding more build-time guard flags. We are trying to move away from configuring gem5 at build time (see change for multi-ISA build (already merged) and multi-Ruby build (not yet merged)).

powerjg:
I had half reviewed this, but got distracted by other things. I'm quite concerned about the `#if` change.

src/mem/ruby/system/CacheRecorder.cc
mattsinc: Do you have a recommendation about an alternative then?

mattsinc:
> I had half reviewed this, but got distracted by other things. I'm quite concerned about the `#if` change.

We can revert the commit to give you more time to review -- I am just not sure how to revert.

src/mem/ruby/system/CacheRecorder.cc
powerjg: There's potential problems with this code that I would like to see explained:
1. It is bad practice to use magic numbers like the fact that BUILD_GPU is 0 when not VEGA_X86 (or something like that). If we use a compile-time define, we must use something that is explicit and not implicit.
2. We want to avoid compile-time defines. I would need to spend time understanding the problem and trying to come up with a different solution. It would be great if @v-ramadas would try to come up with a different solution to the problem.

What *is* the problem this is `#if` is trying to solve?

powerjg:
image

You should be able to click the revert button. However, I would rather not add more commit spam... If we can (quickly) resolve this with a follow up change, that would be best.

src/mem/ruby/system/CacheRecorder.cc
mattsinc: I am not sure I understand the implicit/explicit part, because BUILD_GPU is explicitly set to 1 in the case you described.  I assume you mean it is not explicitly set to 0 in other builds?

Ultimately from @abmerop above, this is the problem:

"I think it is due to the ALL build not building the GPU, and the DebugFlags for GPUMem, etc. are guarded in src/gpu-compute/SConscript are guarded by the configuration variable BUILD_GPU.

I assume this means that all GPU patches will now fail tests if that's the case."

The only options I can think of are:

1. Don't guard the GPU DebugFlags with BUILD_GPU -- I guess there is no harm in them being accessible in a build that is not for the GPU?
2. Have ALL build the GPU -- not sure if there is a Board/stdlib related reason for this though?
3. Use something like ifdef's and explicitly set BUILD_GPU to false in things that aren't GCN3_X86/VEGA_X86.

mattsinc:
> > We'll have to figure out why the error is happening, otherwise one small question
> 
> I think it is due to the ALL build not building the GPU, and the DebugFlags for GPUMem, etc. are guarded in src/gpu-compute/SConscript are guarded by the configuration variable BUILD_GPU.
> 
> I assume this means that all GPU patches will now fail tests if that's the case

@powerjg : this is the error the #if was fixing -- the compilation failed because the added debug flag is guarded by BUILD_GPU.

abmerop:
Implementing #403 should get rid of most (all?) of the new ones in the patch.  We could remove the other ones in various places of the code if we go forward with #306.

mattsinc:
> Implementing #403 should get rid of most (all?) of the new ones in the patch. We could remove the other ones in various places of the code if we go forward with #306.

You're right @abmerop , #403 is likely the solution here.  My above description is also not precise enough -- the real issue is this separate GPUCoalescer map we need without #403.

@powerjg I guess the choice then revert this or let it be in/merged until #403 is done.  @v-ramadas is prioritizing #403 as the next fix, but it will still be a few days.

v-ramadas:
I've created a PR (#448) to address #403 and remove the #if guards

PR #384, opened 2023-10-02
stdlib: Remove duplicated FU in RISCVMatched

kunpai:
This patch removes the duplicate "U74IntFU()"
in the U74FUPool in the RISCVMatched Core.

Change-Id: Idc65166d3c0bf739f2c6a0866267d2ecb963cb9e


powerjg:
Maybe I have a misunderstanding, but I thought that having two `U74IntFU()` meant that two integer instructions could execute at the same time.

My questions are:
- Does this change the performance of the RISCVMatchedBoard?
- If so, does this change improve the accuracy of the board?

kunpai:
I did not know that two IntFU() meant two executions of integer instructions at the same time.

It does not seem that having one improves the performance of the board. It actually hurts it.

So, no, it does not improve the accuracy of the board.

Closing this PR for that reason.

PR #386, opened 2023-10-03
dev: add debug flag in register bank.

hungweihsuG:
Print extra logs for the full/partial read/write access to the registers through the register bank. The debug flag is empty by default and would not print anything.

Test: run unittest of dev/reg_bank.test.xml to check the behavior would not affect the original functionality.
run gem5 with debug flags and use m5term to poke on registers.

Change-Id: I6af4b79c50c36acbbde763c1b3020d7243088261

### DIFF
--- a/src/dev/reg_bank.hh
+++ b/src/dev/reg_bank.hh
@@ -42,7 +42,9 @@
 #include <utility>
 
 #include "base/bitfield.hh"
+#include "base/debug.hh"
 #include "base/logging.hh"
+#include "base/trace.hh"
 #include "base/types.hh"
 #include "sim/byteswap.hh"
 #include "sim/serialize_handlers.hh"
@@ -861,9 +863,27 @@ class RegisterBank : public RegisterBankBase
         void reset() override { _resetter(*this); }
     };
 
+    // Allow gem5 models to set a debug flag to the register bank for logging
+    // all full/partial read/write access to the registers. The register bank
+    // would not log if the flag is not set.
+    //
+    // The debug flag is the one declared in the SConscript
+    //
+    // DebugFlag('HelloExample')
+    //
+    // Then the flag can be set in the register bank with:
+    //
+    // setDebugFlag(::gem5::debug::HelloExample)
+    void
+    setDebugFlag(const ::gem5::debug::SimpleFlag& flag)
+    {
+        _debug_flag = &flag;
+    }
+
   private:
     std::map<Addr, std::reference_wrapper<RegisterBase>> _offsetMap;
 
+    const ::gem5::debug::SimpleFlag* _debug_flag = nullptr;
     Addr _base = 0;
     Addr _size = 0;
     const std::string _name;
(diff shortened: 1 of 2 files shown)


src/dev/reg_bank.hh
```diff
@@ -861,9 +863,26 @@ class RegisterBank : public RegisterBankBase
...
+    // would not log if the flag is not set.
+    //
+    // The debug flag is the one declared in the SConscript
+    //
+    // DebugFlag('HelloExample')
+    //
+    // Then the flag can be set in the register bank with:
+    //
+    // setDebugFlag(::gem5::debug::HelloExample)
+    void setDebugFlag(const ::gem5::debug::SimpleFlag& flag)
```
giactra: Nit: void should be on a separate line:

```
void
setDebugFlags
```

src/dev/reg_bank.hh
```diff
@@ -956,45 +975,34 @@ class RegisterBank : public RegisterBankBase
...
-                it++;
-            } else {
-                // Skip the end of the register, and then we're done.
-                reg.read(ptr + done, 0, remaining);
-                return;
-            }
+        if (_debug_flag) {
+            ::gem5::trace::getDebugLogger()->dprintf_flag(
+                curTick(), name(), _debug_flag->name(), "%s", ss.str());
         }
```
giactra, line 1007: I would consider using DPRINTFV which has been introduced exactly for this reason..

https://github.com/gem5/gem5/blob/stable/src/base/trace.hh#L189

hungweihsuG: Hi! DPRINTFV expects a flag instance however the _debug_flag here is a pointer that might be nullptr. So here we still need an extra null check for this use case. Do you think we should add yet another macro to deal with this use case to include it within GEM5_UNLIKELY?

giactra: Yeah you are right. I don't want to stall this any further... Either we use references instead of pointers or we add the pointer version as you suggested. This can be done with another PR (but only if you feel like)

giactra (commented):
There is only one minor comment and then we are ready to merge this

src/dev/reg_bank.hh
giactra: Bump

giactra:
Hey @hungweihsuG there's only a minor comment to address: https://github.com/gem5/gem5/pull/386#discussion_r1362200057 and then we can merge this...

src/dev/reg_bank.hh
hungweihsuG: Sorry for the late reply. I have fixed this part.

src/dev/reg_bank.hh
hungweihsuG: I decided not to change this in this PR.

giactra (approved):
Thanks for your changes

Harshil2107 (approved):

hungweihsuG:
Hi @powerjg !

I am not familiar with the details of these CI Tests and I can't see the error messages about why they failed.
Could you help me with this issue?

BobbyRBruce:
> Hi @powerjg !
> 
> I am not familiar with the details of these CI Tests and I can't see the error messages about why they failed. Could you help me with this issue?

WE've been having some trouble with our runners (the services that run these tests). I'm pretty sure the tests failed because they failed. I've restarted your tests for you (note: for the next day or two tests may take a while to complete, we're working behind the scenes to setup more runners, but right now we've not got much capacity)

hungweihsuG:
Thanks for helping to merge this change!

PR #389, opened 2023-10-03
configs: Add configurable GPU L1,L2 num banks and L2 latencies

v-ramadas:
Previously, the L1, L2 number of banks and L2 latencies were not configurable through command line arguments. This commit adds support to configure them through the arguments '--tcp-num-banks' for number of banks in L1, '--tcc-num-banks' for number of banks in L2, and '--tcc-tag-access-latency', and '--tcc-data-access-latency'

Change-Id: Ie3b713ead16865fd7120e2d809ebfa56b69bc4a1

### DIFF
--- a/configs/ruby/GPU_VIPER.py
+++ b/configs/ruby/GPU_VIPER.py
@@ -154,6 +154,8 @@ class TCPCntrl(TCP_Controller, CntrlBase):
             dataAccessLatency=options.TCP_latency,
         )
         self.L1cache.resourceStalls = options.no_resource_stalls
+        self.L1cache.dataArrayBanks = options.tcp_num_banks
+        self.L1cache.tagArrayBanks = options.tcp_num_banks
         self.L1cache.create(options)
         self.issue_latency = 1
         # TCP_Controller inherits this from RubyController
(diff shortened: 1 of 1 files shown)


mattsinc (approved):
This looks good to me, but I'm not sure if @powerjg has checked in his ViperBoard.  If so, we'll need to work with him to integrate these into that.

abmerop (approved):

configs/ruby/GPU_VIPER.py
```diff
@@ -298,7 +300,10 @@ def create(self, options):
 class TCCCntrl(TCC_Controller, CntrlBase):
     def create(self, options, ruby_system, system):
         self.version = self.versionCount()
-        self.L2cache = TCC()
+        self.L2cache = TCC(
+            tagAccessLatency=options.tcc_tag_access_latency,
+            dataAccessLatency=options.tcc_data_access_latency,
+        )
         self.L2cache.create(options)
         self.L2cache.resourceStalls = options.no_tcc_resource_stalls
```
powerjg, line 308: Isn't this logic backwards?

powerjg (commented):
I'm generally against adding command line options. I think they are easy to misuse and easy to make mistakes.

That said, I won't block this change.

powerjg:
> This looks good to me, but I'm not sure if @powerjg has checked in his ViperBoard. If so, we'll need to work with him to integrate these into that.

We haven't gotten that in yet.

mattsinc:
> > This looks good to me, but I'm not sure if @powerjg has checked in his ViperBoard. If so, we'll need to work with him to integrate these into that.
> 
> We haven't gotten that in yet.

Until that is merged in, I'm not sure what the alternative would be?  I guess just enabling the underlying code added here but not having a command line arg?

configs/ruby/GPU_VIPER.py
mattsinc: Sorry, I missed that there was an important, pending question here.  I'm not sure what the issue is though, both options are assigned to the corresponding one?

PR #396, opened 2023-10-05
configs,ext: Updated the gem5 SST Bridge to use SST 13.0.0

kaustav-goswami:
This change updates the gem5 SST Bridge to use SST 13.0.0. Changes are made to replace SimpleMem class to StandardMem class as SimpleMem will be deprecated in SST 14 and above. In addition, the translator.hh is updated to translate more types of gem5 packets. A new parameter `ports` was added on SST's side when invoking the gem5 component which does not require recompiling the gem5 component whenever a new outgoing bridge is added in a gem5 config.

Change-Id: I45f0013bc35d088df0aa5a71951422cabab4d7f7

### DIFF
--- a/configs/example/sst/riscv_fs.py
+++ b/configs/example/sst/riscv_fs.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2021 The Regents of the University of California
+# Copyright (c) 2021-2023 The Regents of the University of California
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
@@ -104,15 +104,8 @@ def createHiFivePlatform(system):
 
     system.platform.pci_host.pio = system.membus.mem_side_ports
 
-    system.platform.rtc = RiscvRTC(frequency=Frequency("100MHz"))
+    system.platform.rtc = RiscvRTC(frequency=Frequency("10MHz"))
     system.platform.clint.int_pin = system.platform.rtc.int_pin
-
-    system.pma_checker = PMAChecker(
-        uncacheable=[
-            *system.platform._on_chip_ranges(),
-            *system.platform._off_chip_ranges(),
-        ]
-    )
 
     system.iobus = IOXBar()
     system.bridge = Bridge(delay="50ns")
@@ -121,6 +114,15 @@ def createHiFivePlatform(system):
     system.bridge.ranges = system.platform._off_chip_ranges()
 
     system.platform.setNumCores(1)
+
+    for cpu in system.cpu:
+        # pma_checker has to be added for each of the system cpus.
+        cpu.mmu.pma_checker = PMAChecker(
+            uncacheable=[
+                *system.platform._on_chip_ranges(),
+                *system.platform._off_chip_ranges(),
+            ]
+        )
 
     system.platform.attachOnChipIO(system.membus)
     system.platform.attachOffChipIO(system.iobus)
--- a/ext/sst/INSTALL.md
+++ b/ext/sst/INSTALL.md
@@ -1,8 +1,8 @@
 # Installing SST
 
-The links to download SST source code are available here
-[http://sst-simulator.org/SSTPages/SSTMainDownloads/].
-This guide is using the most recent SST version (11.0.0) as of September 2021.
+The links to download SST source code are available at
+<http://sst-simulator.org/SSTPages/SSTMainDownloads/>.
+This guide is using the most recent SST version (13.0.0) as of September 2023.
 The following guide assumes `$SST_CORE_HOME` as the location where SST will be
 installed.
 
@@ -11,14 +11,14 @@ installed.
 ### Downloading the SST-Core Source Code
 
 ```sh
-wget https://github.com/sstsimulator/sst-core/releases/download/v11.1.0_Final/sstcore-11.1.0.tar.gz
-tar xf sstcore-11.1.0.tar.gz
+wget https://github.com/sstsimulator/sst-core/releases/download/v13.0.0_Final/sstcore-13.0.0.tar.gz
+tar xzf sstcore-13.0.0.tar.gz
 ```
 
 ### Installing SST-Core
 
 ```sh
-cd sstcore-11.1.0
+cd sstcore-13.0.0
 ./configure --prefix=$SST_CORE_HOME --with-python=/usr/bin/python3-config \
             --disable-mpi # optional, used when MPI is not available.
 make all -j$(nproc)
@@ -36,14 +36,14 @@ export PATH=$SST_CORE_HOME/bin:$PATH
 ### Downloading the SST-Elements Source Code
 
 ```sh
-wget https://github.com/sstsimulator/sst-elements/releases/download/v11.1.0_Final/sstelements-11.1.0.tar.gz
-tar xf sstelements-11.1.0.tar.gz
+wget https://github.com/sstsimulator/sst-elements/releases/download/v13.0.0_Final/sstelements-13.0.0.tar.gz
+tar xzf sstelements-13.0.0.tar.gz
 ```
 
 ### Installing SST-Elements
 
 ```sh
-cd sst-elements-library-11.1.0
+cd sst-elements-library-13.0.0
 ./configure --prefix=$SST_CORE_HOME --with-python=/usr/bin/python3-config \
             --with-sst-core=$SST_CORE_HOME
 make all -j$(nproc)
@@ -58,24 +58,36 @@ echo "export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$SST_CORE_HOME/lib/pkgconfig/" >> 
 
 ### Building gem5 library
 
-At the root of gem5 folder,
-
+At the root of the gem5 folder, you need to compile gem5 as a library. This
+varies  dependent on which OS you are using. If you're using Linux, then
+execute the following:
 ```sh
 scons build/RISCV/libgem5_opt.so -j $(nproc) --without-tcmalloc --duplicate-sources
 ```
+In case you're using Mac, then type the following:
+```sh
+scons build/RISCV/libgem5_opt.dylib -j $(nproc) --without-tcmalloc --duplicate-sources
+```
 
-**Note:** `--without-tcmalloc` is required to avoid a conflict with SST's malloc.
-`--duplicate-sources` is required as the compilation of SST depends on sources to be present in the "build" directory.
+**Note:**
+* `--without-tcmalloc` is required to avoid a conflict with SST's malloc.
+* `--duplicate-sources` is required as the compilation of SST depends on sources to be present in the "build" directory.
+* The Mac version was tested on a Macbook Air with M2 processor.
 
 ### Compiling the SST integration
 
-At the root of gem5 folder,
-
+Go to the SST directory in the gem5 repo.
 ```sh
 cd ext/sst
-make
 ```
-
+Depending on your OS, you need to copy the correct `Makefile.xxx` file to
+`Makefile`.
+```sh
+cp Makefile.xxx Makefile    # linux or mac
+make -j4
+```
+The make file is hardcoded to RISC-V. IN the case you wish to compile to ARM,
+edit the Makefile or pass `ARCH=RISCV` to `ARCH=ARM` while compiling.
 ### Running an example simulation
 
 See `README.md`
--- a/ext/sst/sst/arm_example.py
+++ b/ext/sst/sst/arm_example.py
@@ -10,7 +10,7 @@
 # unmodified and in its entirety in all distributions of the software,
 # modified or unmodified, in source code or in binary form.
 #
-# Copyright (c) 2021 The Regents of the University of California
+# Copyright (c) 2021-2023 The Regents of the University of California
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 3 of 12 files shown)


configs/example/sst/riscv_fs.py
```diff
@@ -104,7 +104,7 @@ def createHiFivePlatform(system):
 
     system.platform.pci_host.pio = system.membus.mem_side_ports
 
-    system.platform.rtc = RiscvRTC(frequency=Frequency("100MHz"))
+    system.platform.rtc = RiscvRTC(frequency=Frequency("10MHz"))
```
BobbyRBruce, line 107: Why are we doing this?

ext/sst/sst/arm_example.py
```diff
@@ -35,6 +23,18 @@
...
+# All rights reserved.
+#
+# The license below extends only to copyright in the software and shall
+# not be construed as granting a license to any other intellectual
+# property including but not limited to intellectual property relating
+# to a hardware implementation of the functionality of the software
+# licensed hereunder.  You may use the software subject to the license
+# terms below provided that you ensure that this notice is replicated
+# unmodified and in its entirety in all distributions of the software,
+# modified or unmodified, in source code or in binary form.
```
BobbyRBruce: Why did you move this? "The license below extends..." makes no sense if it's at the bottom.

ext/sst/sst_responder_subcomponent.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 The Regents of the University of California
+// Copyright (c) 2023 The Regents of the University of California
```
BobbyRBruce: You've done this all over but you want to do `2021-2023`. What you're saying here is you don't claim copyright for anything prior to 2023.

Fix this in the places you've done this.

BobbyRBruce (commented):
It's not a _huge_ problem but this is one of those cases where it's nicer to break up your commits. You made a lot of formatting changes here in addition to just updating the SST bridge. It'd be nice to have these two things segregated so I don't lose focus reviewing. That's not a request for change, just an FYI for the future. You can have more than one commit in a PR.

I trust you've actually run the sst example? I believe that's the only test we have.

BobbyRBruce:
@hnpl : Can you give this a lookover. You're the closest we have to an expect in this.

ext/sst/sst/arm_example.py
kaustav-goswami: Oh, I thought the latest commit goes on the top. I'll fix it in the next patch.

configs/example/sst/riscv_fs.py
kaustav-goswami: The clocks between the simulators are correctly synchronized. However, when the kernel boots and clocksource changes to the RiscvRTC, the frequency no longer is synchronized, which becomes 10x slower.

ext/sst/sst_responder_subcomponent.cc
kaustav-goswami: I'll fix this in the next patch.

kaustav-goswami:
> It's not a _huge_ problem but this is one of those cases where it's nicer to break up your commits. You made a lot of formatting changes here in addition to just updating the SST bridge. It'd be nice to have these two things segregated so I don't lose focus reviewing. That's not a request for change, just an FYI for the future. You can have more than one commit in a PR.
> 
> I trust you've actually run the sst example? I believe that's the only test we have.

The reason that I have put this file within this commit as the file is also related to the SST setup that is needed to run the example entire example.

I can actually split this up into two commits. This example has one more correction to be made. I can revert this file and push another change as a separate commit with `configs` tag.

ext/sst/INSTALL.md
```diff
@@ -11,14 +11,14 @@ installed.
 ### Downloading the SST-Core Source Code
 
 ```sh
-wget https://github.com/sstsimulator/sst-core/releases/download/v11.1.0_Final/sstcore-11.1.0.tar.gz
-tar xf sstcore-11.1.0.tar.gz
+wget https://github.com/sstsimulator/sst-core/releases/download/v13.0.0_Final/sstcore-13.0.0.tar.gz
+tar xvf sstcore-13.0.0.tar.gz
```
BobbyRBruce:
```suggestion
tar xzf sstcore-13.0.0.tar.gz
```

Minor nits here:

1. `v` is just `verbose`. This is an uncesssary addition/
2. I like to be explicit and add `z` which means to also unzip (though it does do this automatically).

ext/sst/INSTALL.md
```diff
@@ -36,14 +36,14 @@ export PATH=$SST_CORE_HOME/bin:$PATH
 ### Downloading the SST-Elements Source Code
 
 ```sh
-wget https://github.com/sstsimulator/sst-elements/releases/download/v11.1.0_Final/sstelements-11.1.0.tar.gz
-tar xf sstelements-11.1.0.tar.gz
+wget https://github.com/sstsimulator/sst-elements/releases/download/v13.0.0_Final/sstelements-13.0.0.tar.gz
+tar xvf sstelements-13.0.0.tar.gz
```
BobbyRBruce:
```suggestion
tar xzf sstelements-13.0.0.tar.gz
```

ext/sst/INSTALL.md
```diff
@@ -58,24 +58,34 @@ echo "export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$SST_CORE_HOME/lib/pkgconfig/" >>
 
 ### Building gem5 library
 
-At the root of gem5 folder,
-
+At the root of the gem5 folder, you need to compile gem5 as a library. This
+varies which OS you use. If you're using Linux, then type the following:
```
BobbyRBruce:
```suggestion
varies  dependent on which OS you are useing If you're using Linux, then execute the following:
```

ext/sst/INSTALL.md
```diff
@@ -58,24 +58,34 @@ echo "export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$SST_CORE_HOME/lib/pkgconfig/" >>
...
 ```sh
 scons build/RISCV/libgem5_opt.so -j $(nproc) --without-tcmalloc --duplicate-sources
 ```
+In case you're using Mac, then type the following:
```
BobbyRBruce, line 67:
```suggestion
If you are using Mac OS X, then execute the following:
```

ext/sst/INSTALL.md
```diff
@@ -58,24 +58,34 @@ echo "export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$SST_CORE_HOME/lib/pkgconfig/" >>
...
 
-At the root of gem5 folder,
-
+Go to the SST directory in the gem5 repo.
 ```sh
 cd ext/sst
-make
 ```
-
+According to the OS that you're using, you need to rename the `Makefile.xxx` to `Makefile`.
```
BobbyRBruce:
```suggestion
Depending on your OS, you need to copy the correct `Makefile.xxx` file to `Makefile`.
```

ext/sst/INSTALL.md
```diff
@@ -58,24 +58,34 @@ echo "export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$SST_CORE_HOME/lib/pkgconfig/" >>
...
+```sh
+cp Makefile.xxx Makefile    # linux or mac
+make -j4
+```
+Change `ARCH=RISCV` to `ARCH=ARM` in the `Makefile` in case you're compiling
+for ARM.
```
BobbyRBruce:
```suggestion
The make file is hardcoded to RISC-V. IN the case you wish to compile to ARM, edit the Makefile by changing `ARCH=RISCV` to `ARCH=ARM`.
```

configs/example/sst/riscv_fs.py
BobbyRBruce: I don't understand this. In this file the clock speed is set via an argument. How can we guarantee these are in-sync?

BobbyRBruce (changes requested):

configs/example/sst/riscv_fs.py
kaustav-goswami: I looked at the kernel boot time and the simulator's simulation time. There was a difference of 10x. However, I'd still say that we need better tests to verify this. The kernel switches to the RiscvRTC here:

[    0.007518] clocksource: Switched to clocksource riscv_clocksource

Booting the same kernel in a gem5 only setup and gem5 + SST shows the same 10x difference.

ext/sst/INSTALL.md
kaustav-goswami: Sure, I'll update this.

ext/sst/INSTALL.md
kaustav-goswami: Updated in the next patch.

ext/sst/INSTALL.md
kaustav-goswami: done.

configs/example/sst/riscv_fs.py
powerjg: What do you mean "the frequency becomes 10x slower" What frequency? 

You need to justify why this needs to be different for SST compared to the rest of gem5.

@kunpai had to change this 100MHz to get boot to work. See https://gem5-review.googlesource.com/c/public/gem5/+/64013

I wonder if Kunal remembers why he chose 100MHz

kaustav-goswami: 1. The clock read by the kernel is 10x slower if the frequency is set to 100MHz. This time is read from the `printk` statements during the kernel boot.
2. The simulation ends ~300 ms (as per stats.txt and also SST), but the `printk` statements in the kernel print the kernel time ~3 sec when the kernel is simulated until an m5 exit is encountered.
3. I tried using 1MHz, which, in my case works (exists with m5 exit from early kernel space) but the kernel clock becomes 10x faster. The default frequency is 1MHz in the RiscvRTC simobject (`src/dev/riscv/HiFive.py`)
```
    # The default 1MHz setting is taken from SiFive's U54MC
    # core complex. Set to other frequencies if necessary.
```
5. The clocks between gem5 and SST are in sync. I tested this by looking at every memory packet that gem5 is sending to SST, and, whether SST is receiving the same packet at the exact same time.

kunpai: > What do you mean "the frequency becomes 10x slower" What frequency?
> 
> You need to justify why this needs to be different for SST compared to the rest of gem5.
> 
> @kunpai had to change this 100MHz to get boot to work. See https://gem5-review.googlesource.com/c/public/gem5/+/64013
> 
> I wonder if Kunal remembers why he chose 100MHz

I believe I chose 100 MHz because:
1. Our original chosen value of 1 MHz did not complete an FS run even after a week of running.
2. The RISCV Board provided in `src/python/gem5/components/boards/riscv_board.py`, initializes the RTC to be 100 MHz as well [https://github.com/gem5/gem5/blob/48a40cf2f5182a82de360b7efa497d82e06b1631/src/python/gem5/components/boards/riscv_board.py#L115C65-L115C65](https://github.com/gem5/gem5/blob/48a40cf2f5182a82de360b7efa497d82e06b1631/src/python/gem5/components/boards/riscv_board.py#L115C65-L115C65)

Comparing IPC values for 100 MHz vs 1 MHz for SE mode did not give much of a difference if I remember correctly.

powerjg: Why is SST different than without SST? 

Is it wrong throughout all of gem5?

kaustav-goswami: I don't know why this is different in SST. Also, this only happens in RISCV.  I have not extensively tested ARM version of SST, but from what I remember, this issue is absent in the `arm_example.py`.

I'll verify the time reported via `stats.txt` and the `printk` statements for both gem5 and gem5 + SST setup. I'll use a similar system in both cases. I'll get back with the results.

kaustav-goswami: I did an experiment with a minimal system with 1 Timing CPU clocked at 1 GHz and 1 GiB DRAM without caches. I executed `riscv-boot-exit-nodisk` which executes `m5 exit` as the kernel init script. Here are the results: https://docs.google.com/spreadsheets/d/1wT6_dCyDIwVWuKAXJb617DeE8E5rDKZ3wcgOjh1S_l8/edit?usp=sharing

powerjg: Here's the results for those of you who don't want to open a google doc

<html>
<body>
<google-sheets-html-origin><style type="text/css"></style>

Sl. | Experiment | Gem5 | gem5 + SST
-- | -- | -- | --
stats.txt | printk | SST clock | stats.txt | printk
1 | 1MHz | 29.52111 | 2.709169 | 36.4687 | 36.468664 | 3.283364
2 | 10MHz | 30.541303 | 28.106481 | 38.1729 | 38.172885 | 34.528904
3 | 100MHz | 48.81478 | 462.931175 | 75.5352 | 75.535206 | 716.822168


</body>
</html>

So, we should change all of these 100MHz to 10MHz, then :)

Can you create a new PR with that change? You don't have to update this PR.

powerjg (approved):

hnpl (approved):
LGTM

hnpl:
This is outside of the scope of this change, but since we updated to SST 13.0.0, we should update the test as well.

Here is the test https://github.com/gem5/gem5/blob/develop/tests/deprecated/nightly.sh#L154. I'm not sure why it says deprecated but not ported to supposedly new tests.

Here is the Dockerfile used for testing https://github.com/gem5/gem5/tree/develop/util/dockerfiles/sst-11.1.0.

configs/example/sst/riscv_fs.py
kaustav-goswami: Okay, before opening that PR, I just want to test 100MHz with the RISCV board just to be double sure. I'll update you soon with the results.

BobbyRBruce (approved):

powerjg:
Hmm... is this going to break the daily tests?

See https://github.com/gem5/gem5/blob/stable/.github/workflows/daily-tests.yaml#L214 and #436

Issue #311, opened 2023-09-13
GPU GL2 atomics failing


abmerop:
**Describe the bug**
I am creating this issue because I want to make sure it is tracked and fixed before the next stable release.  Desc: GPU local atomics (in the L2 caches of a GPU) are currently failing two of the now deprecated weekly tests.

**Affects version**
1bebf6a3cc (upstream/develop, origin/develop, github-develop, develop) sim-se: Use tgt_stat64 instead of tgt_stat in newfstatatFunc (#283)

**gem5 Modifications**
No modifications

**To Reproduce**
Steps to reproduce the behavior. Please assume starting from a clean repository:
1. Compile gem5 with command:   scons --linker=mold  -j126 build/VEGA_X86/gem5.opt -k
2. build/VEGA_X86/gem5.opt configs/example/gpufs/vega10_kvm.py --disk-image <path>/ubuntu20-rocm54 --kernel <path>/vmlinux-5.8.0-ubuntu-config --gpu-mmio-trace <path>/gem5-resources/src/gpu-fs/vega_mmio.log  --app  <path>/gem5-resources/src/gpu/pannotia/bc/bin/bc_hip --opts="/home/gem5/1k_128k.gr"

The disk image does not matter. It happens in any ROCm version due to bug being in the TCC / SLICC code. Alternate SE mode:

1a. docker run -u $UID:$GID -w <gem5_cwd> gcr.io/gem5-test/gcn-gpu:latest scons -j126 build/VEGA_X86/gem5.opt -k
2a. docker run -u $UID:$GID -w <gem5_cwd> gcr.io/gem5-test/gcn-gpu:latest build/VEGA_X86/gem5.opt configs/example/apu_se.py -n8  -c  <path>/gem5-resources/src/gpu/pannotia/bc/bin/bc.gem5  --options="1k_128k.gr"  --dgpu --gfx-version=gfx900

**Terminal Output**
simerr output: 
build/VEGA_X86/mem/ruby/protocol/TCC_Transitions.cc:768: panic: Invalid transition                                                                                                                                                                                                  system.ruby.tcc_cntrl0 time: 36221253233 addr: 0xc020c0 event: WrVicBlk state: M

**Expected behavior**
Expect application to run.  It works on stable v23.

**Host Operating System**
Two: Ubuntu 20.04 and Ubuntu 22.04

**Host ISA**
amd64

**Compiler used**
gcc 9.4.0

**Additional information**
Most likely caused by PR #101

[...]

Issue #131, opened 2023-07-27
GPU Failure in the Daily Tests


mkjost0:
Currently, the long gpu tests are failing in our dailies.  This failure can be seen here (https://github.com/gem5/gem5/actions/runs/5653802586/job/15319989256). In order to recreate this, you can run the following code below, though all 4 suites in the tests fail.  These fail after the inclusion of pull request 101 (https://github.com/gem5/gem5/pull/101)

```
scons build/VEGA_X86/gem5.opt
build/VEGA_X86/gem5.opt configs/example/ruby_gpu_random_test.py --test-length 50000 --num-dmas 0
```

When you run this, we receive the following output, followed by a backtrace output:

```
src/mem/dram_interface.cc:690: warn: DRAM device capacity (8192 Mbytes) does not match the address range assigned (1024 Mbytes)
src/cpu/testers/gpu_ruby_test/protocol_tester.cc:140: warn: Data race check is not enabled
src/base/statistics.hh:279: warn: One of the stats is a legacy stat. Legacy stat is a stat that does not belong to any statistics::Group. Legacy stat is deprecated.
src/sim/simulate.cc:194: info: Entering event queue @ 0.  Starting simulation...
      1: global: Episode 0
src/cpu/testers/gpu_ruby_test/address_manager.cc:391: warn: Expected return values are:
         0
src/cpu/testers/gpu_ruby_test/tester_thread.cc:388: warn: GpuWavefront(TesterThread ID = 0, CU ID = 0): Atomic Op returned unexpected value
        Episode 0
        Lane ID 0
        Address [0x31884, line 0x31880]
        Atomic Op's return value 1

src/cpu/testers/gpu_ruby_test/protocol_tester.cc:345: panic: GPU Ruby Tester: Failed!
Memory Usage: 1358852 KBytes
Program aborted at tick 1537
```

abmerop:
Seem to be a few things going on here. This test changed entirely because now all atomics are being done at TCC instead of SD like there were before the PR.

1) wdb_writeDirtyBytes is reseting cache line value to 0 every time.  The data value in an atomic request is not valid I believe. This should be replaced with a new action which only OR's the writeMask.  Setting the mask is needed as it is used by atomicPartial.
2) The value returned is the value *after* the atomic. GPUs return the value *before* the atomic. This can be fixed easily by swapping ar_sendAtomicResponse and pa_performAtomic.
3) There is some other bug after making those changes related to AtomicDone in M state transition missing.

[...]

PR #397, opened 2023-10-05
mem-ruby: SLICC Fixes to GLC Atomics in WB L2

DanKouch:
Made the following changes to fix the behavior of GLC atomics in a WB L2:
   - Stored atomic write mask in TBE For GLC atomics on an invalid line that bypass to the directory, but have their atomics performed on the return path.
   - Replaced !presentOrAvail() check for bypassing atomics to directory (which will then be performed on return path), with check for invalid line state.
   - Replaced wdb_writeDirtyBytes action used when performing atomics with owm_orWriteMask action that doesn't write from invalid atomic request data block
   - Fixed atomic return path actions

Change-Id: I6a406c313d2f9c88cd75bfe39187ef94ce84098f

### DIFF
--- a/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
@@ -106,16 +106,17 @@ machine(MachineType:TCC, "TCC Cache")
   }
 
   structure(TBE, desc="...") {
-    State TBEState,     desc="Transient state";
-    DataBlock DataBlk,  desc="data for the block";
-    bool Dirty,         desc="Is the data dirty?";
-    bool Shared,        desc="Victim hit by shared probe";
-    MachineID From,     desc="Waiting for writeback from...";
-    NetDest Destination, desc="Data destination";
-    int numAtomics,     desc="number remaining atomics";
-    int atomicDoneCnt,  desc="number AtomicDones triggered";
-    bool isGLCSet,      desc="Bypass L1 Cache";
-    bool isSLCSet,      desc="Bypass L1 and L2 Cache";
+    State TBEState,                  desc="Transient state";
+    DataBlock DataBlk,               desc="data for the block";
+    bool Dirty,                      desc="Is the data dirty?";
+    bool Shared,                     desc="Victim hit by shared probe";
+    MachineID From,                  desc="Waiting for writeback from...";
+    NetDest Destination,             desc="Data destination";
+    int numPendingDirectoryAtomics,  desc="number of pending atomics to be performed in directory";
+    int atomicDoneCnt,               desc="number AtomicDones triggered";
+    bool isGLCSet,                   desc="Bypass L1 Cache";
+    bool isSLCSet,                   desc="Bypass L1 and L2 Cache";
+    WriteMask atomicWriteMask,       desc="Atomic write mask";
   }
 
   structure(TBETable, external="yes") {
@@ -265,13 +266,15 @@ machine(MachineType:TCC, "TCC Cache")
         TBE tbe := TBEs.lookup(in_msg.addr);
         Entry cache_entry := getCacheEntry(in_msg.addr);
 
+        // The trigger queue applies only to atomics performed in the directory.
+
         // There is a possible race where multiple AtomicDone triggers can be
         // sent if another Atomic to the same address is issued after the
         // AtomicDone is triggered but before the message arrives here. For
         // that case we count the number of AtomicDones in flight for this
         // address and only call AtomicDone to deallocate the TBE when it is
         // the last in flight message.
-        if (tbe.numAtomics == 0 && tbe.atomicDoneCnt == 1) {
+        if (tbe.numPendingDirectoryAtomics == 0 && tbe.atomicDoneCnt == 1) {
             trigger(Event:AtomicDone, in_msg.addr, cache_entry, tbe);
         } else {
             trigger(Event:AtomicNotDone, in_msg.addr, cache_entry, tbe);
@@ -370,7 +373,7 @@ machine(MachineType:TCC, "TCC Cache")
           // TCC will perform the atomic on the return path on Event:Data.
           // The action will invalidate the cache line if SLC is set and the address is
           // in the cache.
-          if(in_msg.isSLCSet || !WB || !presentOrAvail(in_msg.addr)) {
+          if(in_msg.isSLCSet || !WB) {
             trigger(Event:AtomicPassOn, in_msg.addr, cache_entry, tbe);
           } else {
             trigger(Event:Atomic, in_msg.addr, cache_entry, tbe);
@@ -549,7 +552,23 @@ machine(MachineType:TCC, "TCC Cache")
     cache_entry.DataBlk.clearAtomicLogEntries();
   }
 
-  action(bar_sendBypassedAtomicResponse, "bar", desc="send bypassed Atomic Ack") {
+  action(baplr_sendBypassedAtomicPerformedLocallyResponse, "barplr", desc="send locally-performed bypassed Atomic Ack") {
+    peek(responseFromNB_in, ResponseMsg) {
+        enqueue(responseToCore_out, ResponseMsg, l2_response_latency) {
+          out_msg.addr := address;
+          out_msg.Type := CoherenceResponseType:TDSysResp;
+          out_msg.Destination.add(in_msg.WTRequestor);
+          out_msg.Sender := machineID;
+          out_msg.MessageSize := in_msg.MessageSize;
+          out_msg.DataBlk := cache_entry.DataBlk;
+          out_msg.isGLCSet := tbe.isGLCSet;
+          out_msg.isSLCSet := tbe.isSLCSet;
+        }
+    }
+    cache_entry.DataBlk.clearAtomicLogEntries();
+  }
+
+  action(bapdr_sendBypassedAtomicPerformedInDirectoryResponse, "bapdr", desc="send bypassed Atomic Ack") {
     peek(responseFromNB_in, ResponseMsg) {
         enqueue(responseToCore_out, ResponseMsg, l2_response_latency) {
           out_msg.addr := address;
@@ -585,7 +604,7 @@ machine(MachineType:TCC, "TCC Cache")
       TBEs.allocate(address);
       set_tbe(TBEs.lookup(address));
       tbe.Destination.clear();
-      tbe.numAtomics := 0;
+      tbe.numPendingDirectoryAtomics := 0;
       tbe.atomicDoneCnt := 0;
     }
     if (coreRequestNetwork_in.isReady(clockEdge())) {
@@ -595,6 +614,10 @@ machine(MachineType:TCC, "TCC Cache")
         }
         tbe.isGLCSet := in_msg.isGLCSet;
         tbe.isSLCSet := in_msg.isSLCSet;
+        if(in_msg.Type == CoherenceRequestType:Atomic){
+          tbe.atomicWriteMask.clear();
+          tbe.atomicWriteMask.orMask(in_msg.writeMask);
+        }
       }
     }
   }
@@ -617,6 +640,20 @@ machine(MachineType:TCC, "TCC Cache")
       cache_entry.DataBlk.copyPartial(in_msg.DataBlk,in_msg.writeMask);
       cache_entry.writeMask.orMask(in_msg.writeMask);
       DPRINTF(RubySlicc, "Writing to TCC: %s\n", in_msg);
+    }
+  }
+
+  action(wardb_writeAtomicResponseDirtyBytes, "wardb", desc="write data to TCC") {
+    peek(responseFromNB_in, ResponseMsg) {
+      cache_entry.DataBlk := in_msg.DataBlk;
+      cache_entry.writeMask.orMask(tbe.atomicWriteMask);
+      DPRINTF(RubySlicc, "Writing to TCC: %s\n", in_msg);
+    }
+  }
+
+  action(owm_orWriteMask, "owm", desc="or TCCs write mask") {
+    peek(coreRequestNetwork_in, CPURequestMsg) {
+      cache_entry.writeMask.orMask(in_msg.writeMask);
     }
   }
 
@@ -726,14 +763,20 @@ machine(MachineType:TCC, "TCC Cache")
   }
 
 
-  action(ina_incrementNumAtomics, "ina", desc="inc num atomics") {
-    tbe.numAtomics := tbe.numAtomics + 1;
+  action(inpa_incrementNumPendingDirectoryAtomics, "inpa", desc="inc num atomics") {
+    // Only increment number of atomics if they will actually be performed in directory
+    // That is, if the SLC bit is set or if the cache is write through
+    peek(coreRequestNetwork_in, CPURequestMsg) {
+      if (in_msg.isSLCSet || !WB) {
+        tbe.numPendingDirectoryAtomics := tbe.numPendingDirectoryAtomics + 1;
+      }
+    }
   }
 
 
-  action(dna_decrementNumAtomics, "dna", desc="inc num atomics") {
-    tbe.numAtomics := tbe.numAtomics - 1;
-    if (tbe.numAtomics==0) {
+  action(dnpa_decrementNumPendingDirectoryAtomics, "dnpa", desc="dec num atomics") {
+    tbe.numPendingDirectoryAtomics := tbe.numPendingDirectoryAtomics - 1;
+    if (tbe.numPendingDirectoryAtomics==0) {
       enqueue(triggerQueue_out, TriggerMsg, 1) {
         tbe.atomicDoneCnt := tbe.atomicDoneCnt + 1;
         out_msg.addr := address;
@@ -876,17 +919,8 @@ machine(MachineType:TCC, "TCC Cache")
 
   transition(V, Atomic, M) {TagArrayRead, DataArrayWrite} {
     p_profileHit;
-    wdb_writeDirtyBytes;
-    pa_performAtomic;
-    ar_sendAtomicResponse;
-    p_popRequestQueue;
-  }
-
-  transition(I, Atomic, M) {TagArrayRead, TagArrayWrite, DataArrayWrite} {
-    p_profileMiss;
-    a_allocateBlock;
     ut_updateTag;
-    wdb_writeDirtyBytes;
+    owm_orWriteMask;
     pa_performAtomic;
     ar_sendAtomicResponse;
     p_popRequestQueue;
@@ -900,9 +934,19 @@ machine(MachineType:TCC, "TCC Cache")
     st_stallAndWaitRequest;
   }
 
-  transition({M, W}, Atomic) {TagArrayRead, TagArrayWrite, DataArrayWrite} {
+  transition(W, Atomic, WI) {
+    t_allocateTBE;
+    wb_writeBack;
+    // need to try this request again after writing back the current entry -- to
+    // do so, put it with other stalled requests in a buffer to reduce resource
+    // contention since they won't try again every cycle and will instead only
+    // try again once woken up
+    st_stallAndWaitRequest;
+  }
+
+  transition(M, Atomic) {TagArrayRead, TagArrayWrite, DataArrayWrite} {
     p_profileHit;
-    wdb_writeDirtyBytes;
+    owm_orWriteMask;
     pa_performAtomic;
     ar_sendAtomicResponse;
     p_popRequestQueue;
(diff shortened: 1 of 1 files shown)


DanKouch:
This, together with #120 should address the underlying cause of #131.

With this patch, ruby random tester works with a WB L2 and #133 reverted. As #120 addresses issues with non-WB L2, after both of these go in, I'll make another PR to revert #133.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -105,16 +105,17 @@ machine(MachineType:TCC, "TCC Cache")
...
-    DataBlock DataBlk,  desc="data for the block";
-    bool Dirty,         desc="Is the data dirty?";
-    bool Shared,        desc="Victim hit by shared probe";
-    MachineID From,     desc="Waiting for writeback from...";
-    NetDest Destination, desc="Data destination";
-    int numAtomics,     desc="number remaining atomics";
-    int atomicDoneCnt,  desc="number AtomicDones triggered";
-    bool isGLCSet,      desc="Bypass L1 Cache";
-    bool isSLCSet,      desc="Bypass L1 and L2 Cache";
+    State TBEState,             desc="Transient state";
```
mattsinc: Why are all of these changed?  Did you mess around with spacing or something?  Seems like only 118 should be changed ...

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -707,6 +739,10 @@ machine(MachineType:TCC, "TCC Cache")
     }
   }
 
+  action(dnant_decrementNumAtomicsNoTrigger, "dnant", desc="inc num atomics") {
```
mattsinc: What is the "no trigger" part about?

mattsinc (changes requested):

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -837,17 +873,8 @@ machine(MachineType:TCC, "TCC Cache")
 
   transition(V, Atomic, M) {TagArrayRead, DataArrayWrite} {
     p_profileHit;
-    wdb_writeDirtyBytes;
-    pa_performAtomic;
-    ar_sendAtomicResponse;
-    p_popRequestQueue;
-  }
-
-  transition(I, Atomic, M) {TagArrayRead, TagArrayWrite, DataArrayWrite} {
```
abmerop, line 885: Is I>M gone now for Atomic?

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -837,17 +873,8 @@ machine(MachineType:TCC, "TCC Cache")
...
-    p_profileMiss;
-    a_allocateBlock;
     ut_updateTag;
-    wdb_writeDirtyBytes;
+    owm_orWriteMask;
     pa_performAtomic;
     ar_sendAtomicResponse;
```
abmerop, line 925: This will send the modified data *after* the atomic, wouldn't it?  Atomics must return the previous data, before an atomic was performed.   I am still thinking we need to save the old data somewhere.  This applies for all transitions that call the ar_sendAtomicResponse action

abmerop (changes requested):
Not quite done with review, but I think the wrong data may be being returned

abmerop:
Before submitting this series, I want to test a GPU application which performs GLC atomics to the same address in the same wavefront.  This will make sure this and Bujji's patches (atomic log support) still all work together. That would probably be the most 'advanced' atomic test we could do right now

mattsinc:
> Before submitting this series, I want to test a GPU application which performs GLC atomics to the same address in the same wavefront. This will make sure this and Bujji's patches (atomic log support) still all work together. That would probably be the most 'advanced' atomic test we could do right now

@DanKouch I believe we have such a test (microbenchmark) internally.  Have you tested it on this commit?  If so what happens?  Also, assuming we have one, can you please point @abmerop to where it is in our internal repo (maybe via email, not here)?

mattsinc:
> > Before submitting this series, I want to test a GPU application which performs GLC atomics to the same address in the same wavefront. This will make sure this and Bujji's patches (atomic log support) still all work together. That would probably be the most 'advanced' atomic test we could do right now
> 
> @DanKouch I believe we have such a test (microbenchmark) internally. Have you tested it on this commit? If so what happens? Also, assuming we have one, can you please point @abmerop to where it is in our internal repo (maybe via email, not here)?

@DanKouch ping on this.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
DanKouch: The longer name of the added entry pushed back the indentation of the description, so I further indented the rest to match. Should I change this back?

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
DanKouch: The trigger queue appears to be in place to account for race conditions where an atomic request that gets performed in the directory is sent to the directory while another atomic request that gets sent to the directory is being processed by the directory.

In cases where the atomic is performed in the directory I don't believe this race condition exists since the directory is only reading data; the behavior would be similar to reads that go to the directory, which don't have a trigger queue. Because of this, and because when the trigger queue is emptied the atomic is transitioned to invalid, I've circumvented the trigger queue by adding a way to directly decrement the atomic count without sending a trigger. This is only used in a WB L2 with atomic requests that are performed locally on the return path.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
DanKouch: The previous data is stored in the atomic log, and read back from the atomic log in the coalescer per [this change](https://github.com/gem5/gem5/pull/200/files#diff-9e61b05abc867f6a980b3e43684b5e477ebd2b896a7e79aba0c573d5fa9a964bR591) in #200 .

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
mattsinc: I see, thanks -- I guess this is ok

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
mattsinc: I don't think we should circumvent the trigger queue.  Doing this in the past is how we inadvertently introduced races into this protocol (e.g., because of replays when the queue was full).

mattsinc (changes requested):

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
abmerop: Ah, right.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
abmerop: The trigger queue in the previous case was used in a transition that isn't guaranteed to change states, so it used the trigger queue to change the state iif the atomic count was zero.  I think this case is different because we will always transition from A>M on data and then I believe further atomics will stay M>M?  Will let Dan confirm though.

abmerop (commented):
This mostly looks good, but I need to think about the trigger queue comment.

I checked and this is fixing #311 and the underlying cause of #131 which should allow #133 to be reverted.  So once this is checked in #311 can be marked fixed and #133 can be reverted to test GLC atomics

mattsinc:
I can't seem to comment under @abmerop 's last comment, but my mental model after the prior data races with trigger queue is that we should not try to elide it or create some sort of "side channel".  It is unclear why M-->M obviates the need for this?

abmerop:
Looking at this again, I don't think there is anyway to implement this action that doesn't use the trigger queue.  If there are two or more atomics to the same line, we need to decrement counters and only move to the next state when the counter is zero.  There is no mechanism in SLICC to do this as far as I know besides the trigger queue.  Some other protocols have a "nextState" value in the TBE and transition based on that, but it's not in VIPER, would be a lot of work to add, and it probably not even a better solution than trigger queue

abmerop:
Ignore that last comment, I see the different now.  Would be is more acceptable to just decrement the counter in `barplr_sendBypassedAtomicPerformedLocallyResponse`?

mattsinc:
> Ignore that last comment, I see the different now. Would be is more acceptable to just decrement the counter in `barplr_sendBypassedAtomicPerformedLocallyResponse`?

I know @DanKouch is having some computer problems, so let me at least reply here with what we discussed.  It sounds like you see the same thing, but with this path, the TBEs/MSHRs hold all subsequent atomics to the same address/line, so there should not be the race we had with the trigger queue.

In terms of fixes, I asked @DanKouch to change the naming scheme to avoid confusion (along with some other smaller things).

I suppose we could do the decrement in barplr_sendBypassedAtomicPerformedLocallyResponse, but I feel like keeping the actions separate makes more sense?

In any event, @DanKouch this has been dragging on for a long time so please update soon.

mattsinc:
> > > Before submitting this series, I want to test a GPU application which performs GLC atomics to the same address in the same wavefront. This will make sure this and Bujji's patches (atomic log support) still all work together. That would probably be the most 'advanced' atomic test we could do right now
> > 
> > 
> > @DanKouch I believe we have such a test (microbenchmark) internally. Have you tested it on this commit? If so what happens? Also, assuming we have one, can you please point @abmerop to where it is in our internal repo (maybe via email, not here)?
> 
> @DanKouch ping on this.

This too, @DanKouch

abmerop:
> > Ignore that last comment, I see the different now. Would be is more acceptable to just decrement the counter in `barplr_sendBypassedAtomicPerformedLocallyResponse`?
> 
> I know @DanKouch is having some computer problems, so let me at least reply here with what we discussed. It sounds like you see the same thing, but with this path, the TBEs/MSHRs hold all subsequent atomics to the same address/line, so there should not be the race we had with the trigger queue.
> 
> In terms of fixes, I asked @DanKouch to change the naming scheme to avoid confusion (along with some other smaller things).
> 
> I suppose we could do the decrement in barplr_sendBypassedAtomicPerformedLocallyResponse, but I feel like keeping the actions separate makes more sense?
> 
> In any event, @DanKouch this has been dragging on for a long time so please update soon.

OK, sounds good.  I am fine with the rename and keeping separate actions

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
mattsinc: @DanKouch looks like this name still needs to be updated?

mattsinc (changes requested):

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
DanKouch: @mattsinc Yep, just rebased for now. I have the changes ready but I'm running into some trouble with non-returning atomics. I'll push as soon as I get that sorted.

mattsinc:
> > > > Before submitting this series, I want to test a GPU application which performs GLC atomics to the same address in the same wavefront. This will make sure this and Bujji's patches (atomic log support) still all work together. That would probably be the most 'advanced' atomic test we could do right now
> > > 
> > > 
> > > @DanKouch I believe we have such a test (microbenchmark) internally. Have you tested it on this commit? If so what happens? Also, assuming we have one, can you please point @abmerop to where it is in our internal repo (maybe via email, not here)?
> > 
> > 
> > @DanKouch ping on this.
> 
> This too, @DanKouch

Making sure you saw this, @DanKouch ?

DanKouch:
> > > > > Before submitting this series, I want to test a GPU application which performs GLC atomics to the same address in the same wavefront. This will make sure this and Bujji's patches (atomic log support) still all work together. That would probably be the most 'advanced' atomic test we could do right now
> > > > 
> > > > 
> > > > @DanKouch I believe we have such a test (microbenchmark) internally. Have you tested it on this commit? If so what happens? Also, assuming we have one, can you please point @abmerop to where it is in our internal repo (maybe via email, not here)?
> > > 
> > > 
> > > @DanKouch ping on this.
> > 
> > 
> > This too, @DanKouch
> 
> Making sure you saw this, @DanKouch ?

@mattsinc 

Yep, this is where I'm running into problems with non-returning atomics. From what I can tell we're not differentiating between them in the protocol, and the coalescer assumes the atomic log (that gets filled whenever an atomic is performed) won't be populated for non-returning atomics. Clearing the atomic log in the coalescer if we're doing a non-returning atomic would be a quick change, but I'm not sure if that should be handled in the protocol itself.

mattsinc:
> > > > > > Before submitting this series, I want to test a GPU application which performs GLC atomics to the same address in the same wavefront. This will make sure this and Bujji's patches (atomic log support) still all work together. That would probably be the most 'advanced' atomic test we could do right now
> > > > > 
> > > > > 
> > > > > @DanKouch I believe we have such a test (microbenchmark) internally. Have you tested it on this commit? If so what happens? Also, assuming we have one, can you please point @abmerop to where it is in our internal repo (maybe via email, not here)?
> > > > 
> > > > 
> > > > @DanKouch ping on this.
> > > 
> > > 
> > > This too, @DanKouch
> > 
> > 
> > Making sure you saw this, @DanKouch ?
> 
> @mattsinc
> 
> Yep, this is where I'm running into problems with non-returning atomics. From what I can tell we're not differentiating between them in the protocol, and the coalescer assumes the atomic log (that gets filled whenever an atomic is performed) won't be populated for non-returning atomics. Clearing the atomic log in the coalescer if we're doing a non-returning atomic would be a quick change, but I'm not sure if that should be handled in the protocol itself.

@abmerop , @DanKouch : do we want to separate this issue out and fix in the suggested fix by @mysoreanoop in #508?

DanKouch:
@mattsinc @abmerop 
I've pushed a new patchset which I think clarifies numAtomics by renaming it to numPendingDirectoryAtomics, and using it only to keep track of the number of atomics that the TCC is waiting on the directory to perform. Since this is only incremented for atomics the directory will perform, there's no longer any need for a "no trigger" decrement action, which I've removed. I've also added a comment to the trigger queue to clarify that it's only used when the directory will be performing the atomic.

The patch doesn't address non-return atomics.

mysoreanoop:
> The patch doesn't address non-return atomics.

I'll send in the PR for clearing atomic log on NO_RETURN in GPUCoalescer then.

Also, I'm seeing this error:
```
build/GCN3_X86/mem/ruby/protocol/TCC_Transitions.cc:801: panic: Invalid transition
system.ruby.tcc_cntrl0 time: 324722691 addr: 0x10807c0 event: AtomicDone state: M 
``` 
Haven't fully analyzed it, but could it be related to this PR? If not, I can start a new issue.
PS: This was encountered before applying this PR.

mattsinc:
> > The patch doesn't address non-return atomics.
> 
> I'll send in the PR for clearing atomic log on NO_RETURN in GPUCoalescer then.
> 
> Also, I'm seeing this error:
> 
> ```
> build/GCN3_X86/mem/ruby/protocol/TCC_Transitions.cc:801: panic: Invalid transition
> system.ruby.tcc_cntrl0 time: 324722691 addr: 0x10807c0 event: AtomicDone state: M 
> ```
> 
> Haven't fully analyzed it, but could it be related to this PR? If not, I can start a new issue. PS: This was encountered before applying this PR.

@mysoreanoop : without more information it's very difficult to say.  Ultimately the problem is that there is a transition needed that isn't there.  But how are you getting into that case?  That's what we'd need to know.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -370,7 +373,7 @@ machine(MachineType:TCC, "TCC Cache")
           // TCC will perform the atomic on the return path on Event:Data.
           // The action will invalidate the cache line if SLC is set and the address is
           // in the cache.
-          if(in_msg.isSLCSet || !WB || !presentOrAvail(in_msg.addr)) {
```
mattsinc, line 373: Remind me why this is removed?

mattsinc (commented):
I think the change seems sane, although I would appreciate @abmerop chiming in about the non-return atomics and if we want them incorporated here or not.  One other question.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
DanKouch: The presentOrAvail check was intended to check if the cache line was invalid, but it wasn't quite doing that. To account for invalid, lines, I've added "Atomic" to the AtomicPassOn transition for going from I to A [here](https://github.com/gem5/gem5/pull/397/files#diff-70af2194466dfb4563c3a46e40715dfcffbf43f262eb4eac9da661c67e240996R968).

mattsinc (approved):
Thanks @DanKouch !  This makes sense to me now, but we should probably let @abmerop check too.

abmerop:
> > The patch doesn't address non-return atomics.
> 
> I'll send in the PR for clearing atomic log on NO_RETURN in GPUCoalescer then.
> 
> Also, I'm seeing this error:
> 
> ```
> build/GCN3_X86/mem/ruby/protocol/TCC_Transitions.cc:801: panic: Invalid transition
> system.ruby.tcc_cntrl0 time: 324722691 addr: 0x10807c0 event: AtomicDone state: M 
> ```
> 
> Haven't fully analyzed it, but could it be related to this PR? If not, I can start a new issue. PS: This was encountered before applying this PR.

This PR fixes this invalid transition

mattsinc:
> Thanks @DanKouch ! This makes sense to me now, but we should probably let @abmerop check too.

@abmerop : this one is waiting on you -- let us know if you need anything further.  We'll merge #120 and @mysoreanoop 's once this one is in.

abmerop (approved):

mattsinc:
> Made the following changes to fix the behavior of GLC atomics in a WB L2:
> 
>     * Stored atomic write mask in TBE For GLC atomics on an invalid line that bypass to the directory, but have their atomics performed on the return path.
> 
>     * Replaced !presentOrAvail() check for bypassing atomics to directory (which will then be performed on return path), with check for invalid line state.
> 
>     * Replaced wdb_writeDirtyBytes action used when performing atomics with owm_orWriteMask action that doesn't write from invalid atomic request data block
> 
>     * Fixed atomic return path actions
> 
> 
> Change-Id: I6a406c313d2f9c88cd75bfe39187ef94ce84098f

@DanKouch : I changed "TLB" to "TBE" here, as I'm pretty sure you meant TBE?

mattsinc:
> > Made the following changes to fix the behavior of GLC atomics in a WB L2:
> > ```
> > * Stored atomic write mask in TBE For GLC atomics on an invalid line that bypass to the directory, but have their atomics performed on the return path.
> > 
> > * Replaced !presentOrAvail() check for bypassing atomics to directory (which will then be performed on return path), with check for invalid line state.
> > 
> > * Replaced wdb_writeDirtyBytes action used when performing atomics with owm_orWriteMask action that doesn't write from invalid atomic request data block
> > 
> > * Fixed atomic return path actions
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Change-Id: I6a406c313d2f9c88cd75bfe39187ef94ce84098f
> 
> @DanKouch : I changed "TLB" to "TBE" here, as I'm pretty sure you meant TBE?

@DanKouch : are you ok with this change?  If so we should merge this.

DanKouch:
> > > Made the following changes to fix the behavior of GLC atomics in a WB L2:
> > > ```
> > > * Stored atomic write mask in TBE For GLC atomics on an invalid line that bypass to the directory, but have their atomics performed on the return path.
> > > 
> > > * Replaced !presentOrAvail() check for bypassing atomics to directory (which will then be performed on return path), with check for invalid line state.
> > > 
> > > * Replaced wdb_writeDirtyBytes action used when performing atomics with owm_orWriteMask action that doesn't write from invalid atomic request data block
> > > 
> > > * Fixed atomic return path actions
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Change-Id: I6a406c313d2f9c88cd75bfe39187ef94ce84098f
> > 
> > 
> > @DanKouch : I changed "TLB" to "TBE" here, as I'm pretty sure you meant TBE?
> 
> @DanKouch : are you ok with this change? If so we should merge this.

@mattsinc Yes, thank you for fixing that.

mattsinc:
> > > > Made the following changes to fix the behavior of GLC atomics in a WB L2:
> > > > ```
> > > > * Stored atomic write mask in TBE For GLC atomics on an invalid line that bypass to the directory, but have their atomics performed on the return path.
> > > > 
> > > > * Replaced !presentOrAvail() check for bypassing atomics to directory (which will then be performed on return path), with check for invalid line state.
> > > > 
> > > > * Replaced wdb_writeDirtyBytes action used when performing atomics with owm_orWriteMask action that doesn't write from invalid atomic request data block
> > > > 
> > > > * Fixed atomic return path actions
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Change-Id: I6a406c313d2f9c88cd75bfe39187ef94ce84098f
> > > 
> > > 
> > > @DanKouch : I changed "TLB" to "TBE" here, as I'm pretty sure you meant TBE?
> > 
> > 
> > @DanKouch : are you ok with this change? If so we should merge this.
> 
> @mattsinc Yes, thank you for fixing that.

Ok, please go ahead and click "Merge pull request" at the bottom then -- and congrats on merging this!

powerjg:
FYI, only maintainers (like me and you) can merge things, Matt.

mattsinc:
> FYI, only maintainers (like me and you) can merge things, Matt.

Good to know, thanks.

PR #392, opened 2023-10-04
stdlib: Added retrying when connection issue to db

Harshil2107:
Change-Id: I1b4e8405c310efebabc39b270135236e62aa65db

PR #404, opened 2023-10-05
stdlib: Improve handing of errors in Atlas request failures

### LINKED ISSUE #340: Retry, rather than immediately fail when `urlopen` Exception raised in "client_wrapper.py"
Fairly frequently tests will fail to download a resource. If you look into the stderr you see errors like the following:

```sh
warn: Error getting resources from client gem5-resources: <urlopen error [Errno -2] Name or service not known>
Exception: Resource with ID 'rv32ud-ps-fmin' not found.

At:
  src/python/gem5/resources/client_api/client_wrapper.py(151): get_resource_json_obj_from_client
  src/python/gem5/resources/client.py(152): get_resource_json_obj
  src/python/gem5/resources/resource.py(658): obtain_resource
  /__w/gem5/gem5/tests/gem5/asmtest/configs/riscv_asmtest.py(108): <module>
```

Based on the "Name or service not known", I believe in this case it's due to DNS lookup being unsuccessful. I don't know why this happens but I'd guess it's just a bad networking issue on one of the services we host our GitHub Runners (I suspect Amarillo... I've had problems with its networking before).

Regardless, in this case, and in others regarding networking issues, it'd be good if this code retried a few times before failing. This would make our tests considerably less flakey. This error has nothing to do with errors in our codebase, only errors in our networking (I _think_).

I'd suggest implementing a back-off algorithm that retries a few times with some idle time between each attempt. A niave approach could be : retry immediately; retried after 10 seconds; retried after 20 seconds; retry after 30 second. After this we can fail. `urlopen` may provide some functions to do this for you, I haven't bothered to look.

BobbyRBruce:
Now:

* The Atlas Client will attempt a connection 4 times, using an exponential backoff approach between attempts.
* When a failure does arise a rich output is given so problems can be easily diagnosed.

Addresses: #340

### DIFF
--- a/src/python/gem5/resources/client_api/atlasclient.py
+++ b/src/python/gem5/resources/client_api/atlasclient.py
@@ -30,7 +30,7 @@ import json
 import time
 import itertools
 from .abstract_client import AbstractClient
-
+from urllib.error import HTTPError
 from m5.util import warn
 
 
@@ -77,7 +77,7 @@ class AtlasClient(AbstractClient):
 
     def get_token(self):
         return self._atlas_http_json_req(
-            self.url,
+            self.authUrl,
             data_json={"key": self.apiKey},
             headers={"Content-Type": "application/json"},
             purpose_of_request="Get Access Token with API key",
@@ -122,7 +122,7 @@ class AtlasClient(AbstractClient):
             try:
                 response = request.urlopen(req)
                 break
-            except Exception as e:
+            except HTTPError as e:
                 if attempt >= max_failed_attempts:
                     raise AtlasClientHttpJsonRequestError(
                         client=self,
@@ -167,7 +167,10 @@ class AtlasClient(AbstractClient):
         }
 
         resources = self._atlas_http_json_req(
-            url, data=data, headers=headers, purpose_of_request="Get Resources"
+            url,
+            data_json=data,
+            headers=headers,
+            purpose_of_request="Get Resources",
         )["documents"]
 
         # I do this as a lazy post-processing step because I can't figure out
--- a/src/python/gem5/resources/client_api/atlasclient.py
+++ b/src/python/gem5/resources/client_api/atlasclient.py
@@ -25,10 +25,39 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 from urllib import request, parse
-from urllib.error import HTTPError, URLError
 from typing import Optional, Dict, Union, Type, Tuple, List, Any
 import json
+import time
+import itertools
 from .abstract_client import AbstractClient
+
+from m5.util import warn
+
+
+class AtlasClientHttpJsonRequestError(Exception):
+    def __init__(
+        self,
+        client: "AtlasClient",
+        data: Dict[str, Any],
+        purpose_of_request: Optional[str],
+    ):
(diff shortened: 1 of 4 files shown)


BobbyRBruce:
@Harshil2107 : I was looking over #392 and had a few suggestions. This then spiraled into me just doing it. Open for feedback here. I haven't tested this yet (so pseudo-drafty), but I think the design idea is sound.

BobbyRBruce:
I don't know what's going on, but the tests are failing with the 404 error, so at least we can see what the error would look like to the user:

```sh
warn: `Resource` has been deprecated. Please use the `obtain_resource` function instead.
warn: Attempt 1 of Atlas HTTP Request failed.
Purpose of Request: Get Access Token with API key.

Failed with Exception:
HTTP Error 404: Not Found

Retrying after 2 seconds...
warn: Attempt 2 of Atlas HTTP Request failed.
Purpose of Request: Get Access Token with API key.

Failed with Exception:
HTTP Error 404: Not Found

Retrying after 4 seconds...
warn: Attempt 3 of Atlas HTTP Request failed.
Purpose of Request: Get Access Token with API key.

Failed with Exception:
HTTP Error 404: Not Found

Retrying after 8 seconds...
Exception: Error getting resources from client gem5-resources

At:
  src/python/gem5/resources/client_api/client_wrapper.py(117): get_all_resources_by_id
  src/python/gem5/resources/client_api/client_wrapper.py(150): get_resource_json_obj_from_client
  src/python/gem5/resources/client.py(152): get_resource_json_obj
  src/python/gem5/resources/resource.py(736): obtain_resource
  src/python/gem5/resources/resource.py(975): Resource
  /__w/gem5/gem5/tests/gem5/m5_util/configs/simple_binary_run.py(92): <module>
  src/python/m5/main.py(642): main
```

This somehow isn't giving the errors information from the `AtlasClient` like i thought they'd be. This is a bug.

BobbyRBruce:
Think i've fixed the issue partially in the latest push... we'll see.

src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -47,22 +77,70 @@ def __init__(self, config: Dict[str, str]):
...
+            data_json={"key": self.apiKey},
+            headers={"Content-Type": "application/json"},
+            purpose_of_request="Get Access Token with API key",
+        )["access_token"]
+
+    def _atlas_http_json_req(
+        self,
+        url: str,
+        data_json: Dict[str, Any],
+        headers: dict[str, str],
```
Harshil2107: Shouldn't this be `Dict[str, str]`?

src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -47,22 +77,70 @@ def __init__(self, config: Dict[str, str]):
...
+        for attempt in itertools.count(start=1):
+            try:
+                response = request.urlopen(req)
+                break
+            except Exception as e:
+                if attempt >= max_failed_attempts:
+                    raise AtlasClientHttpJsonRequestError(
+                        client=self,
+                        data=data_json,
+                        headers=headers,
```
Harshil2107: `header` is being passed as a parameter to `AtlasClientHttpJsonRequestError` but it does not have a parameter named `header`.

powerjg (approved):
lgtm

PR #406, opened 2023-10-06
python,misc: Add more hooks to pre-commit and update those we have

BobbyRBruce:
**This is a draft**. I'm playing around with these hooks to see which are of value are not (one commit per hook).


powerjg (commented):
These are all OK with me except for the disallow submodules. 

However, I think this should be split into different PRs for each change.

BobbyRBruce:
Closing this PR. It has been broken apart into the following PR: 

#419 
#420 
#421 
#422 
#423
#424

PR #407, opened 2023-10-06
mem-ruby: Far atomics fix

giactra:
The PR is fixing the CHI fromSequencer helper function which is making
use of the undefined tbe entry.

This has been broken by #177

Change-Id: I52feff4b5ab2faf0aa91edd6572e3e767c88e257


mattsinc (approved):
This seems reasonable/correct to me.

PR #408, opened 2023-10-06
gpu-compute: Update tokens for flat global/scratch

abmerop:
Memory instructions acquire coalescer tokens in the schedule stage.
Currently this is only done for buffer and flat instructions, but not
flat global or flat scratch. This change now acquires tokens for flat
global and flat scratch instructions. This provides back-pressure to the
CUs and helps to avoid deadlocks in Ruby.

The change also handles returning tokens for buffer, flat global, and
flat scratch instructions. This was previously only being done for
normal flat instructions leading to deadlocks in some applications when
the tokens were exhausted.

To simplify the logic, added a needsToken() method to GPUDynInst which
return if the instruction is buffer or any flat segment.

The waitcnts were also incorrect for flat global and flat scratch. We
should always decrement vmem and exp count for stores and only normal
flat instructions should decrement lgkm. Currently vmem/exp are not
decremented for flat global and flat scratch which can lead to deadlock.
This change set fixes this by always decrementing vmem/exp and lgkm only
for normal flat instructions.

Change-Id: I673f4ac6121e4b5a5e8491bc9130c6d825d95fc5

### DIFF
--- a/src/arch/amdgpu/vega/insts/instructions.cc
+++ b/src/arch/amdgpu/vega/insts/instructions.cc
@@ -43894,9 +43894,11 @@ namespace VegaISA
     {
         Wavefront *wf = gpuDynInst->wavefront();
 
-        if (gpuDynInst->exec_mask.none() && isFlat()) {
+        if (gpuDynInst->exec_mask.none()) {
             wf->decVMemInstsIssued();
-            wf->decLGKMInstsIssued();
+            if (isFlat()) {
+                wf->decLGKMInstsIssued();
+            }
             return;
         }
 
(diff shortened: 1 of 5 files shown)


mattsinc:
@abmerop : are the waitcnt's incorrect before this patch, after this patch, or both?  I'm not sure what this wording means:

The waitcnts are also incorrect for flat global and flat scratch. We should always decrement vmem and exp count for stores and only normal flat instructions should decrement lgkm. Currently vmem/exp are not decremented for flat global and flat scratch which can lead to deadlock.

I assume it means you fixed it in this patch?

mattsinc (approved):
In regard to my question about the commit message, my reading after viewing the code is you are saying it was a problem before this commit, and is fixed by this commit.  Maybe worth updating commit message to make this clear?

abmerop:
Yes, that is fixed by this commit.  I'll update the commit message (might be a while as my gem5 machine is down)

abmerop:
> In regard to my question about the commit message, my reading after viewing the code is you are saying it was a problem before this commit, and is fixed by this commit. Maybe worth updating commit message to make this clear?

Updated commit message

mattsinc (approved):

PR #410, opened 2023-10-06
dev-amdgpu,gpu-compute: Implement GPU and HSA timestamps

abmerop:
This PR adds two commit to handle timestamps in the ROCm runtime. ROCr uses a mix of GPU timestamp reads and HSA packet timestamps to output profiling information for a task dispatch.

The first patch added timestamps to the HSA completion signal indicating when the task started and ended and require changing the flow of completion signal DMAs to ensure the DMA of the timestamp values completed before writing the completion signal value.

Second commit adds MMIOs for reading the GPU's timestamp counter. This MMIO resides in the GFX MMIO space so a new class is added to handle MMIOs in that address range.

### DIFF
--- a/src/dev/amdgpu/amdgpu_gfx.cc
+++ b/src/dev/amdgpu/amdgpu_gfx.cc
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2023 Advanced Micro Devices, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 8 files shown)


src/dev/amdgpu/amdgpu_gfx.cc
```diff
@@ -0,0 +1,73 @@
+/*
```
mattsinc, line 1: I am not an expert at reading the legalese at the top, but just verifying you made sure this doesn't include whatever problematic phrases caused trouble a while back?

mattsinc (approved):

src/dev/amdgpu/amdgpu_gfx.cc
powerjg: This looks like the classic 3-clause BSD to me. But I'll let Matt P. confirm.

abmerop: Yea it's the standard 3 clause BSD... the problematic line is not there

mattsinc: thanks both

PR #411, opened 2023-10-06
cpu-kvm, arch-x86: flush TLB after syscalls

### LINKED ISSUE #409: x86 KVM in SE doesn't flush TLB when pages are unmapped
**Describe the bug**
The X86KvmCPU model doesn't invalidate TLB entries for pages that have been unmapped. For example, if the application calls the `munmap()` system call, the application will still be able to access the unmapped pages:
```c
#include <sys/mman.h>                                                                   
#include <stdlib.h>                                                                     
#include <stdio.h>                                                                      
int main() {                                                                            
  char *mem = mmap(NULL, 0x1000, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE, -1, 0);
  *mem = 0x42;                                                                          
  munmap(mem, 0x1000);                                                                  
  printf("%02hhx\n", *mem); // should crash, but doesn't in X86KvmCPU
}                                                                                       
```
This can also cause reads to access the wrong physical page if we have remapped a new physical page to the virtual address:
```c
#include <sys/mman.h>                                                                  
#include <stdlib.h>                                                                    
#include <stdio.h>                                                                     
int main() {                                                                           
  char *mem = (char *) 0x12340000;                                                     
  mmap(mem, 0x1000, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE | MAP_FIXED, -1, 0);
  *mem = 0x42;                                                                         
  munmap(mem, 0x1000);                                                                 
  mmap(mem, 0x1000, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE | MAP_FIXED, -1, 0);
  printf("%02hhx\n", *mem); // should print "00", but prints "42" on X86KvmCPU         
} 
```

**Affects version**
develop @ [ae104cc](https://github.com/gem5/gem5/commit/ae104cc431c065fbf53ab003cbf4aba23de12ac1)

**gem5 Modifications**
None

**To Reproduce**
1. Compile the second program above: `cc -static mremap.c -o mremap`
2. Compile gem5 as usual: `scons build/X86/gem5.opt`
3. Run `mremap` in SE mode using X86KvmCPU: `./build/X86/gem5.opt configs/deprecated/example/se.py --cmd ./mremap --cpu-type=X86KvmCPU`

**Terminal Output**
Simulation prints:
```
42
```

**Expected behavior**
Host prints
```
00
```

**Host Operating System**
Ubuntu 22.04

**Host ISA**
X86

**Compiler used**
GCC 11.4.0
**Additional information**
i'll submit a PR to fix this soon

nmosier:
Modified the x86 KVM-in-SE syscall handler to flush the TLB following each syscall, in case the page table has been modified. This is done by reloading the value in %cr3. Doing this requires an intermediate GPR, which we store in a new scratch buffer following the syscall code at address `syscallDataBuf`.

GitHub issue: https://github.com/gem5/gem5/issues/409

Change-Id: Ibc20018c97ebb1794fa31a0c71e0857d661c7c9d

### DIFF
--- a/src/arch/x86/process.cc
+++ b/src/arch/x86/process.cc
@@ -483,13 +483,35 @@ X86_64Process::initState()
         physProxy.writeBlob(idtPhysAddr + 0xE0, &PFGate, sizeof(PFGate));
 
         /* System call handler */
+        // First, we write to the MMIO m5ops range (0xffffc90000007000)
+        // to trap out of the VM back into gem5 to emulate the system
+        // call. Upon re-entering the VM, we need to flush the TLB in
+        // case the system call modified existing page mappings (e.g.,
+        // munmap, mremap, brk). To do this, we can simply read/write
+        // cr3; however, doing so requires saving the value to an
+        // intermediate GPR (%rax, in this case). We save/restore the
+        // value of %rax in the scratch region syscallDataBuf.
+        const Addr syscallDataBuf = syscallCodeVirtAddr + 0x100;
         uint8_t syscallBlob[] = {
             // mov    %rax, (0xffffc90000007000)
             0x48, 0xa3, 0x00, 0x70, 0x00,
             0x00, 0x00, 0xc9, 0xff, 0xff,
+            // mov    %rax, (syscallDataBuf)
+            0x48, 0xa3, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00,
+            // mov    %cr3, %rax
+            0x0f, 0x20, 0xd8,
+            // mov    %rax, %cr3
+            0x0f, 0x22, 0xd8,
+            // mov    (syscallDataBuf), %rax
+            0x48, 0xa1, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00,
             // sysret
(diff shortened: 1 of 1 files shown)


powerjg (approved):
Thanks for another great fix!

abmerop (approved):
Thank you for another great fix and bug description!

You had another related to XSave so now I am curious to know if you happen to be working with KVM+AVX.  If so, we should sync up along with @mattsinc and @v-ramadas as we are trying to fix something in that area

nmosier:
> You had another related to XSave so now I am curious to know if you happen to be working with KVM+AVX. If so, we should sync up along with @mattsinc and @v-ramadas as we are trying to fix something in that area

I haven't yet encountered any AVX-related crashes in the KVM workloads I'm running, but I think AVX should be broken in the KVM cpu for the same reason that SSE was. The fix should be pretty similar to the SSE fix, I think -- I'd be happy to take a look. Is there a GitHub issue you can point me to?

mattsinc:
@v-ramadas : do you want to post the failure here (with AVX)?

powerjg:
@mattsinc @v-ramadas you should probably open a separate issue or PR instead of following up on this one.

PR #358, opened 2023-09-25
BPU refactoring

dhschall:
This set of patches combines multiple enhancements in the branch predictor unit. Specifically it:
- Adds a new BranchType class useful for better statistics and debugging.
- Creates Separate SimObjects for BTB, RAS and Indirect branch predictor to allow different implementation of each.
- Moves the logics log RAS and indirect branch predictor management out of the main BPU to simplify its design.
- Improves statistics

The changes to the branch predictor are required to implement the decoupled front-end

PR #412, opened 2023-10-07
cpu: Restructure BTB

dhschall:
This is the first PR in a series of enhancements to the BPU proposed in #358.
However, I think putting everything into one PR is not nice to review and prone to oversee I might did.

This PR restructures the BTB:
- A new abstract BTB class is created to enable different BTB implementations. The new BTB class gets its own parameter and stats.
- An enum is added to differentiate branch instruction types. This enum is used to enhance statistics and BPU management.
- The existing BTB is moved into `simple_btb` as default.
- An additional function is added to store the static instruction in the BTB. This function is used for the decoupled front-end.
- Update configs to match new BTB parameters.

### DIFF
--- a/src/cpu/pred/btb.hh
+++ b/src/cpu/pred/btb.hh
@@ -1,4 +1,16 @@
 /*
+ * Copyright (c) 2022-2023 The University of Edinburgh
+ * All rights reserved
+ *
+ * The license below extends only to copyright in the software and shall
+ * not be construed as granting a license to any other intellectual
+ * property including but not limited to intellectual property relating
+ * to a hardware implementation of the functionality of the software
+ * licensed hereunder.  You may use the software subject to the license
+ * terms below provided that you ensure that this notice is replicated
+ * unmodified and in its entirety in all distributions of the software,
+ * modified or unmodified, in source code or in binary form.
+ *
  * Copyright (c) 2004-2005 The Regents of The University of Michigan
  * All rights reserved.
  *
@@ -29,9 +41,13 @@
 #ifndef __CPU_PRED_BTB_HH__
 #define __CPU_PRED_BTB_HH__
 
+
 #include "arch/generic/pcstate.hh"
-#include "base/logging.hh"
-#include "base/types.hh"
+#include "base/statistics.hh"
+#include "cpu/pred/branch_type.hh"
+#include "cpu/static_inst.hh"
+#include "params/BranchTargetBuffer.hh"
+#include "sim/clocked_object.hh"
 
 namespace gem5
 {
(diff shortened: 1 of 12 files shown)


src/cpu/pred/btb.hh
```diff
@@ -29,103 +41,87 @@
...
+    virtual void memInvalidate() override = 0;
 
-  public:
-    /** Creates a BTB with the given number of entries, number of bits per
-     *  tag, and instruction offset amount.
-     *  @param numEntries Number of entries for the BTB.
-     *  @param tagBits Number of bits for each tag in the BTB.
-     *  @param instShiftAmt Offset amount for instructions to ignore alignment.
+    /** Looks up an address in the BTB. Upon hit returns the target of the
+     * branch. Returns nullptr.
```
andysan: You can probably drop the description of the return value here, that's already described below. If you decide to keep it, update the description of a miss.

src/cpu/pred/btb.hh
```diff
@@ -29,103 +41,87 @@
...
 
-    /** Checks if a branch is in the BTB.
+    /** Checks if a branch address is in the BTB. Intended as a quick check
+     *  before calling lookup. Does not update statistics.
      *  @param inst_PC The address of the branch to look up.
-     *  @param tid The thread id.
      *  @return Whether or not the branch exists in the BTB.
      */
-    bool valid(Addr instPC, ThreadID tid);
+    virtual bool valid(ThreadID tid, Addr instPC);
```
andysan: Should this be purely virtual? I.e., is there any reason for a BTB to not implement this?

src/cpu/pred/btb.hh
```diff
@@ -29,103 +41,87 @@
...
-
-    /** Looks up an address in the BTB. Must call valid() first on the address.
+    /** Looks up an address in the BTB and return the instruction
+     * information if existant. May not be supported in all BTBs.
      *  @param inst_PC The address of the branch to look up.
-     *  @param tid The thread id.
      *  @return Returns the target of the branch.
      */
-    const PCStateBase *lookup(Addr instPC, ThreadID tid);
+    virtual const StaticInstPtr lookupInst(ThreadID tid, Addr instPC);
```
andysan: Should this be purely virtual? Is there any reason you might not want to implement it?

src/cpu/pred/btb.hh
```diff
@@ -29,103 +41,87 @@
...
      */
-    DefaultBTB(unsigned numEntries, unsigned tagBits,
-               unsigned instShiftAmt, unsigned numThreads);
+    virtual const PCStateBase *lookup(ThreadID tid, Addr instPC,
+                            BranchType type = BranchType::NoBranch) = 0;
 
-    void reset();
-
-    /** Looks up an address in the BTB. Must call valid() first on the address.
+    /** Looks up an address in the BTB and return the instruction
```
andysan, line 84: It might be worth renaming this to `getInst` or `lookupInstNoStats` to indicate that this doesn't actually update any stats. You should also document this behaviour.

src/cpu/pred/simple_btb.cc
```diff
@@ -0,0 +1,181 @@
...
+        && btb[btb_idx].tid == tid) {
+        return btb[btb_idx].target.get();
+    } else {
+        stats.misses[type]++;
+        return nullptr;
+    }
+}
+
+const StaticInstPtr
+SimpleBTB::lookupInst(ThreadID tid, Addr instPC)
```
andysan: There is a lot of duplication between the two lookup methods. I'd suggest refactoring this so you have a `getBTBEntry` (or something along those lines) method that returns a `BTBEntry` or `nullptr` that you use in both.

andysan (commented):
A couple of comments, but looks good overall.

src/cpu/pred/btb.hh
dhschall: Its new that we are also storing the static instruction in the BTB thats why I did not make it pure.
But I can change that

andysan: I think we might want to make this purely virtual. All of the implementations in gem5 will implement this interface and the CPUs will presumably depend on it at some point.

src/cpu/pred/simple_btb.cc
dhschall: Great suggestion! Thanks for that.
I combined now all lookup functions in an internal function.

andysan (approved):
Thanks for the updated version, this looks good to me.

PR #415, opened 2023-10-09
configs: Add an example elastic trace generation script

giactra:
Current [TraceCPU documentation](https://www.gem5.org/documentation/general_docs/cpu_models/TraceCPU) still references the deprecated **se.py/fs.py** scripts for elastic trace generation (script paths are also outdated).

With this PR we provide a simpler Arm based elastic trace generation script that can
be used out of the box by a user or that can be extended as needed.

### DIFF
--- a/configs/common/cores/arm/O3_ARM_Etrace.py
+++ b/configs/common/cores/arm/O3_ARM_Etrace.py
@@ -1,0 +1,58 @@
+# Copyright (c) 2012, 2017-2018, 2023 Arm Limited
+# All rights reserved.
+#
+# The license below extends only to copyright in the software and shall
+# not be construed as granting a license to any other intellectual
+# property including but not limited to intellectual property relating
+# to a hardware implementation of the functionality of the software
+# licensed hereunder.  You may use the software subject to the license
+# terms below provided that you ensure that this notice is replicated
+# unmodified and in its entirety in all distributions of the software,
+# modified or unmodified, in source code or in binary form.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


powerjg (approved):
This looks good to me. It would be nice if we could focus our efforts around the stdlib, but I like that this change doesn't use common/Options.py!

giactra:
> This looks good to me. It would be nice if we could focus our efforts around the stdlib, but I like that this change doesn't use common/Options.py!

Thanks Jason; I considered using stdlib at the beginning but I realised it would have required me to add/amend the existing APIs (of SimpleProcessor IIRC). It looked painful/ugly so I decided to use the existing models in configs/example/arm as the logic required is minimal and it does not force me to change/break any API

powerjg:
> I considered using stdlib at the beginning but I realised it would have required me to add/amend the existing APIs (of SimpleProcessor IIRC). It looked painful/ugly so I decided to use the existing models in configs/example/arm as the logic required is minimal and it does not force me to change/break any API

Looping in @BobbyRBruce 

Yeah, I'm not surprised that it wasn't straightforward. The APIs Bobby developed for the stdlib were the minimum to get things working, and we have always planned on having to update them as we port old (or new) use cases over the to stdlib. In general, we are hoping to encourage people to use those APIs instead of the "raw" m5 library. 

It would be good for us to understand better what API changes are necessary to get things like the TraceCPU working in the stdlib. Right now we don't have much bandwidth to look into it here, though.

rjc-arch (approved):
Looks good to me, as long as everyone else is happy. Thanks for the improvements, Giacomo.

PR #416, opened 2023-10-09
arch-arm: Implement FEAT_TCR2 and FEAT_SCTLR2

giactra:
This is simply adding the new Armv8.9 registers defined in the related features:

- FEAT_TCR2
- FEAT_SCTLR2


BobbyRBruce (approved):
Not am arm expert so will leave it to you, but this looks fine to me.

andysan (approved):

PR #421, opened 2023-10-09
python,misc: Add `reorder-python-imports` hook to pre-commit

BobbyRBruce:
This:

1. Adds `reorder-python-imports` to pre-commit.
2. Applies this to all files in the repo.
3. Adds (2.) to ".git-blame-ignore-revs".


andysan:
I'd suggest that we use isort instead which seems more popular on GitHub and it is known to work well with black. I usually use that for my own projects. I personally like this isort configuration (we would probably add `m5` as a first-party library to ensure correct ordering`):
```
[tool.isort]
profile = "black"
multi_line_output = 3
force_grid_wrap = 2
```

Which works with this pre-commit configuration:
```
- repo: https://github.com/PyCQA/isort
  rev: 5.11.5
  hooks:
  - id: isort
```

This ensures that multiple imports from the same module end up in a list instead of as separate import statements. I generally find that more readable. For example, we currently get this:
```
from foo import bar
from foo import baz
```

With the isort configuration above, you get this instead:
```
from foo import (
    bar,
    baz,
)
```

andysan:
I have had a quick look at the way this changes Python files. There is some weirdness that we need to resolve if we decide to use this tool (I would prefer isort):

- Some comment blocks seem to be modified for no good reason.
- Local includes and `m5` includes are mixed. That seems a bit weird.

We probably don't need to remove this from the blame history. It's a pretty localised change, so I it would cause a lot of issues if it appears in blame. In fact, that might be desirable since changes to the import order could potentially come with unforeseen side-effects.

powerjg:
+1 for isort and that local includes and `m5` should be separate.

Thanks for pointing these things out, Andreas!

BobbyRBruce:
Good points. Seems like `isort` is the way to go. I'm closing this PR in favor of #431

PR #424, opened 2023-10-09
misc,python: Add `pyupgrade` to pre-commit

BobbyRBruce:
This adds the [pyupgrade](https://github.com/asottile/pyupgrade) hook to pre-commit.

This hook automatically upgrades the syntax to the recommended standards for the newer version of the language.


andysan:
This looks good to me, but I would suggest not adding a git blame ignore for this one.

BobbyRBruce:
> This looks good to me, but I would suggest not adding a git blame ignore for this one.

Removed.

I'd like a second pair of eyes on this commit: https://github.com/gem5/gem5/pull/424/commits/c4156b06fb6e393ce9d253eb2af6e042267e84f0

When compiling gem5 without it there was an out of bounds exception. The tuple `base` was empty. I _suspect_ it's because `pyupgrade` changes instances of `class Foo(object)` to `class Foo` and this meant `base` could be empty. I'm not 100% sure on this but the patch in that commit does appear to fix the error.

andysan:
> > This looks good to me, but I would suggest not adding a git blame ignore for this one.
> 
> Removed.
> 
> I'd like a second pair of eyes on this commit: [c4156b0](https://github.com/gem5/gem5/commit/c4156b06fb6e393ce9d253eb2af6e042267e84f0)
> 
> When compiling gem5 without it there was an out of bounds exception. The tuple `base` was empty. I _suspect_ it's because `pyupgrade` changes instances of `class Foo(object)` to `class Foo` and this meant `base` could be empty. I'm not 100% sure on this but the patch in that commit does appear to fix the error.

That sounds like a plausible explanation to me.

andysan (approved):
LGTM

PR #425, opened 2023-10-09
configs,tests: Remove `mkdir` in simpoint-se-checkpoint.py

BobbyRBruce:
This `mkdir` is problematic as it doesn't create the directory recursively. This casues errors if `dir` is `X/Y/Z` and both `Y` and `Z` has not been created. An error will be returned (`No such file or directory`).

This issue was fixed with: https://github.com/gem5/gem5/pull/263. The checkpointing code already recursively creates directories as needed. Ergo was can remove this `mkdir` statement.


BobbyRBruce:
This bug causes some errors when running the CI tests. I'm therefore quite eager to get this. @powerjg  Can you give this approval?

powerjg (approved):

PR #426, opened 2023-10-09
misc: Add Workflow to give merging advice

BobbyRBruce:
In response to discussions in
https://github.com/orgs/gem5/discussions/261, `Squash Merges` have been enabled.

This adds a GitHub Action to add a comment to a PR when the commit is in a mergable state explaining which strategy `Merge` or `Squash Merge` is preferable.

### DIFF
--- a/.github/workflows/when-pr-mergeable.yaml
+++ b/.github/workflows/when-pr-mergeable.yaml
@@ -1,0 +1,35 @@
+name: Write Comment on PR Update
+
+on:
+    pull_request:
+        types: [opened, edited, synchronize]
+
+jobs:
+    write-single-commit-commment:
+        # Only run when the pull request is mergeable and has one commit
+        if: github.event.pull_request.mergeable == true && github.event.pull_request.comments == 1
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/github-script@v6
+              with:
+                script: |
+                    github.rest.issues.createComment({
+                        issue_number: context.issue.number,
+                        owner: context.repo.owner,
+                        repo: context.repo.repo,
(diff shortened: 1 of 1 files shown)


.github/workflows/when-pr-mergeable.yaml
```diff
@@ -0,0 +1,35 @@
+name: Write Comment on PR Update
+
+on:
+    pull_request:
+        types: [opened, edited, synchronize]
+
+jobs:
+    write-single-commit-commment:
+        # Only run when the pull request is mergeable and has one commit
+        if: github.event.pull_request.mergeable == true && github.event.pull_request.comments == 1
```
powerjg, line 10: Do *all* single commit changes have a single comment? What happens if I comment on a PR, wouldn't that mean there would be >1 comment?

How about "commits"? See https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request

.github/workflows/when-pr-mergeable.yaml
```diff
@@ -0,0 +1,35 @@
...
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/github-script@v6
+              with:
+                script: |
+                    github.rest.issues.createComment({
+                        issue_number: context.issue.number,
+                        owner: context.repo.owner,
+                        repo: context.repo.repo,
+                        body: 'This PR is now in a mergeable state. As it as a multi-commit PR, we recommend using the `Merge`. Please ensure the title and description of the merge commit (defaulted to the PR title and description) is sufficient to describe the collection of patches. `Squash` may be used if you wish to squash all the commits into a single commit.'
```
powerjg, line 34: You may want to add some information about when someone should squash vs merge.

PR #428, opened 2023-10-10
cpu: Restructure RAS

dhschall:
This is the second patch set to restructure the branch predictor unit

The return address stack (RAS) is restructured to be a separate SimObject. 
The intention is to better separate functionality to make the main predictor
simpler. Furthermore, it allows disabling the RAS altogether or developing
new RAS features like corruption detection more easily.

Change-Id: I8aacf7d4c8e308165d0e7e15bc5a5d0df77f8192

### DIFF
--- a/src/cpu/pred/BranchPredictor.py
+++ b/src/cpu/pred/BranchPredictor.py
@@ -57,6 +57,15 @@ class BranchType(Enum):
     ]
 
 
+class ReturnAddrStack(SimObject):
+    type = "ReturnAddrStack"
+    cxx_class = "gem5::branch_prediction::ReturnAddrStack"
+    cxx_header = "cpu/pred/ras.hh"
+
+    numThreads = Param.Unsigned(Parent.numThreads, "Number of threads")
+    numEntries = Param.Unsigned(16, "Number of RAS entries")
+
+
 class BranchTargetBuffer(ClockedObject):
     type = "BranchTargetBuffer"
     cxx_class = "gem5::branch_prediction::BranchTargetBuffer"
@@ -120,10 +129,10 @@ class BranchPredictor(SimObject):
     numThreads = Param.Unsigned(Parent.numThreads, "Number of threads")
     instShiftAmt = Param.Unsigned(2, "Number of bits to shift instructions by")
 
-    RASSize = Param.Unsigned(16, "RAS size")
-
     btb = Param.BranchTargetBuffer(SimpleBTB(), "Branch target buffer (BTB)")
-
+    ras = Param.ReturnAddrStack(
+        ReturnAddrStack(), "Return address stack, set to NULL to disable RAS."
+    )
     indirectBranchPred = Param.IndirectPredictor(
         SimpleIndirectPredictor(),
         "Indirect branch predictor, set to NULL to disable "
--- a/src/cpu/pred/bpred_unit.cc
+++ b/src/cpu/pred/bpred_unit.cc
@@ -60,13 +60,11 @@ BPredUnit::BPredUnit(const Params &params)
       numThreads(params.numThreads),
       predHist(numThreads),
       btb(params.btb),
-      RAS(numThreads),
+      ras(params.ras),
       iPred(params.indirectBranchPred),
       stats(this),
       instShiftAmt(params.instShiftAmt)
 {
-    for (auto& r : RAS)
-        r.init(params.RASSize);
 }
 
 BPredUnit::BPredUnitStats::BPredUnitStats(statistics::Group *parent)
@@ -172,19 +170,12 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
         //       support coroutines.
         if (inst->isReturn()) {
             ++stats.RASUsed;
-            predict_record.wasReturn = true;
-            // If it's a function return call, then look up the address
-            // in the RAS.
-            const PCStateBase *ras_top = RAS[tid].top();
-            if (ras_top)
-                set(target, inst->buildRetPC(pc, *ras_top));
-
-            // Record the top entry of the RAS, and its index.
-            predict_record.usedRAS = true;
-            predict_record.RASIndex = RAS[tid].topIdx();
-            set(predict_record.RASTarget, ras_top);
-
-            RAS[tid].pop();
+            // If it's a return from a function call, then look up the
+            // RETURN address in the RAS.
+            const PCStateBase *return_addr = ras->pop(tid,
+                                                predict_record.rasHistory);
+            if (return_addr)
+                set(target, return_addr);
 
             DPRINTF(Branch, "[tid:%i] [sn:%llu] Instruction %s is a return, "
                     "RAS predicted target: %s, RAS index: %i\n",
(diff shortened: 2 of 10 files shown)


src/cpu/pred/BranchPredictor.py
```diff
@@ -116,7 +125,9 @@ class BranchPredictor(SimObject):
     RASSize = Param.Unsigned(16, "RAS size")
```
andysan: Is this still needed? If not, just move the default to the new `ReturnAddrStack` object.

src/cpu/pred/ras.cc
```diff
@@ -83,5 +114,210 @@ ReturnAddrStack::restore(unsigned top_entry_idx, const PCStateBase *restored)
     }
 }
 
+std::string
+ReturnAddrStack::AddrStack::print(int n)
```
andysan: This should probably be called something else as it isn't exactly printing anything. Maybe `toStr`?

src/cpu/pred/bpred_unit.hh
```diff
@@ -218,13 +218,15 @@ class BPredUnit : public SimObject
                          void *indirect_history, ThreadID _tid,
                          const StaticInstPtr & inst)
             : seqNum(seq_num), pc(instPC), bpHistory(bp_history),
-              indirectHistory(indirect_history), tid(_tid),
+              indirectHistory(indirect_history), rasHistory(nullptr),
+              tid(_tid),
               predTaken(pred_taken), inst(inst)
         {}
 
         PredictorHistory(const PredictorHistory &other) :
```
andysan, line 226: Are `wasReturn` and `wasCall` still used in this struct? Probably a good idea to remove them if not.

andysan (commented):
This look reasonable to me. When looking at the change, it seems like we add a new behaviour (restoring the stack) when there is an incorrectly predicted return? You need to mention that in the commit message if that is indeed a new behaviour since it fixes a bug.

dhschall:
I debugged why the test fail for RISC-V https://github.com/gem5/gem5/actions/runs/6466291455/job/17561077071?pr=428
This is due to a strange instruction in RISC-V. `jalr` or `c_jalr`. Its flags for call and return are both set as it should push and pop the ras.
https://github.com/gem5/gem5/blob/48a40cf2f5182a82de360b7efa497d82e06b1631/src/arch/riscv/isa/formats/standard.isa#L281

This behavior seems to be compliant with the specs https://riscv.org/wp-content/uploads/2017/05/riscv-spec-v2.2.pdf page 17.

The reason my code breaks is that I assumed an instruction can be either call or return and therefore I added an assert to catch bugs. Seems this is not a bug but a feature 🙈

src/cpu/pred/ras.cc
dhschall: Is `showRAS` more explicite?

src/cpu/pred/bpred_unit.cc
```diff
@@ -176,37 +174,30 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
...
+            // RETURN address in the RAS.
+            const PCStateBase *return_addr = ras->pop(tid,
+                                                predict_record.rasHistory);
+            if (return_addr)
+                set(target, return_addr);
 
             DPRINTF(Branch, "[tid:%i] [sn:%llu] Instruction %s is a return, "
                     "RAS predicted target: %s, RAS index: %i\n",
                     tid, seqNum, pc, *target, predict_record.RASIndex);
         }
```
dhschall, line 183: To fix the not bug I could simply make this to `} else if (inst->isCall()) {`. But I am not sure if that's what we want. Theoretically, It should just skip all stack operations.

src/cpu/pred/bpred_unit.cc
```diff
@@ -340,30 +333,15 @@ BPredUnit::squash(const InstSeqNum &squashed_sn, ThreadID tid)
...
-             RAS[tid].pop();
-        }
-        if (pred_hist.front().usedRAS) {
-            if (pred_hist.front().RASTarget != nullptr) {
-                DPRINTF(Branch, "[tid:%i] [squash sn:%llu]"
-                        " Restoring top of RAS to: %i,"
-                        " target: %s\n", tid, squashed_sn,
-                        pred_hist.front().RASIndex,
-                        *pred_hist.front().RASTarget);
-            }
```
dhschall, line 353: The restoring of the stack was already done before. I just tried to move all this complexity away from the BPU into the RAS.
A simple simple call to `squash` should resolve any incorrect stack operation

src/cpu/pred/bpred_unit.cc
```diff
@@ -172,37 +170,30 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
         //       support coroutines.
```
dhschall, line 170: Ohhhh. Is it this note?
To support coroutines I should allow both, push and pop, in one instruction?

andysan (dismissed):
This looks good to me. @rjc-arch could you confirm that this doesn't introduce any unexpected performance difference?

src/cpu/pred/ras.cc
```diff
@@ -72,9 +100,9 @@ ReturnAddrStack::pop()
 }
 
 void
-ReturnAddrStack::restore(unsigned top_entry_idx, const PCStateBase *restored)
+ReturnAddrStack::AddrStack::restore(unsigned _tos,const PCStateBase *restored)
```
rjc-arch: Nit: there should be a space after the comma.

src/cpu/pred/ras.cc
```diff
@@ -83,5 +111,219 @@ ReturnAddrStack::restore(unsigned top_entry_idx, const PCStateBase *restored)
     }
 }
 
+std::string
+ReturnAddrStack::AddrStack::showRAS(int n)
```
rjc-arch: Nit: The name `showRAS` sounds like it prints the RAS. Would a name like `toString` or similar be more descriptive?

src/cpu/pred/ras.cc
```diff
@@ -83,5 +111,219 @@ ReturnAddrStack::restore(unsigned top_entry_idx, const PCStateBase *restored)
...
+{
+    std::stringstream ss;
+    ss << "";
```
rjc-arch: Nit: Is this redundant?

src/cpu/pred/ras.cc
```diff
@@ -83,5 +111,219 @@ ReturnAddrStack::restore(unsigned top_entry_idx, const PCStateBase *restored)
...
+    for (int i = 0; i<n; i++) {
```
rjc-arch: Nit: there should be spaces between binary operators according to the gem5 coding guidelines.

src/cpu/pred/ras.cc
```diff
@@ -83,5 +111,219 @@ ReturnAddrStack::restore(unsigned top_entry_idx, const PCStateBase *restored)
...
+        int idx = int(tos)-i;
+        if (idx < 0 || addrStack[idx] == nullptr) {
+            break;
+        }
+        ss << std::dec << idx << ":0x" << std::setfill('0') << std::setw(8)
```
rjc-arch: Should this be `setw(16)` to properly print 64-bit addresses?

src/cpu/pred/ras.hh
```diff
@@ -41,70 +58,168 @@ namespace branch_prediction
...
+       *  @param top_of_stack the index saved at the time of the prediction.
+       *  @param restored The new target address of the new top of the RAS.
+       */
+      void restore(unsigned top_of_stack, const PCStateBase *restored);
+
+      bool empty() { return usedEntries == 0; }
+
+      bool full() { return usedEntries >= numEntries; }
+
+      /** print the n top entries of the stack */
```
rjc-arch: Nit: technically it returns them as a string, but doesn't print anything.

src/cpu/pred/ras.hh
```diff
@@ -41,70 +58,168 @@ namespace branch_prediction
...
-
-    /** The number of used entries in the RAS. */
-    unsigned usedEntries;
-
-    /** The top of stack index. */
-    unsigned tos;
+    /** The number of threads */
+    unsigned numThreads;
+    /** Fallback to BTB in case a stack corruption was detected. */
+    unsigned useBtbFallback;
```
rjc-arch: This looks like it's unused.

rjc-arch (commented):
Hi David - this is all looking very good! Thanks for all your hard work.
I have left a few small comments below, but I couldn't see any significant problems.

rjc-arch:
I can confirm I have run performance regressions on this patch using SPEC 2017 SimPoints, and this patch does not affect the reported IPC.

rjc-arch (approved):
Thanks for making the changes, David.
Looks good to me now!

andysan (approved):

PR #429, opened 2023-10-10
cpu: Refactor indirect predictor

dhschall:
This is the third patch set to restructure the branch predictor unit

Simplify indirect predictor interface. Several of the existing functions where merged together into four clear once. Those four are similar to the main direction predictor interface. 'lookup', 'update', 'squash' and 'commit'. This makes the interface much more clear, allows better functionality isolation and makes it simpler to develop new predictor models.

Change-Id: I6d6b43965b2986ef959953a64c428e50bc68d38e

### DIFF
--- a/src/cpu/pred/BranchPredictor.py
+++ b/src/cpu/pred/BranchPredictor.py
@@ -100,6 +100,13 @@ class SimpleIndirectPredictor(IndirectPredictor):
     indirectPathLength = Param.Unsigned(
         3, "Previous indirect targets to use for path history"
     )
+    speculativePathLength = Param.Unsigned(
+        256,
+        "Additional buffer space to store speculative path history. "
+        "If there are more speculative branches in flight the history cannot "
+        "be recovered. Set this to an appropriate value respective the CPU"
+        "pipeline depth or a high value e.g. 256 to make it 'unlimited'.",
+    )
     indirectGHRBits = Param.Unsigned(13, "Indirect GHR number of bits")
     instShiftAmt = Param.Unsigned(2, "Number of bits to shift instructions by")
 
@@ -119,7 +126,8 @@ class BranchPredictor(SimObject):
 
     indirectBranchPred = Param.IndirectPredictor(
         SimpleIndirectPredictor(),
-        "Indirect branch predictor, set to NULL to disable indirect predictions",
+        "Indirect branch predictor, set to NULL to disable "
+        "indirect predictions",
     )
 
 
--- a/src/cpu/pred/indirect.hh
+++ b/src/cpu/pred/indirect.hh
@@ -1,6 +1,16 @@
 /*
  * Copyright (c) 2014 ARM Limited
- * All rights reserved.
+ * Copyright (c) 2023 The University of Edinburgh
+ * All rights reserved
+ *
+ * The license below extends only to copyright in the software and shall
+ * not be construed as granting a license to any other intellectual
+ * property including but not limited to intellectual property relating
+ * to a hardware implementation of the functionality of the software
+ * licensed hereunder.  You may use the software subject to the license
+ * terms below provided that you ensure that this notice is replicated
+ * unmodified and in its entirety in all distributions of the software,
+ * modified or unmodified, in source code or in binary form.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are
@@ -26,11 +36,16 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+/* @file
+ * Indirect target predictor interface
+ */
+
 #ifndef __CPU_PRED_INDIRECT_BASE_HH__
 #define __CPU_PRED_INDIRECT_BASE_HH__
 
 #include "arch/generic/pcstate.hh"
 #include "cpu/inst_seq.hh"
+#include "cpu/pred/branch_type.hh"
 #include "params/IndirectPredictor.hh"
 #include "sim/sim_object.hh"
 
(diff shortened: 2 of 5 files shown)


src/cpu/pred/indirect.hh
```diff
@@ -51,21 +66,55 @@ class IndirectPredictor : public SimObject
...
-                                           void * indirect_history,
-                                           bool actually_taken) = 0;
+    virtual void reset() {};
+
+    /**
+     * Predicts the indirect target of an indirect branch.
+     * @param tid Thread ID of the branch.
+     * @param sn The sequence number of the branch.
+     * @param PC The branch PC address.
+     * @param iHistory The pointer to the history object.
```
andysan: This should be `i_history` to comply with the style guide.

src/cpu/pred/indirect.hh
```diff
@@ -51,21 +66,55 @@ class IndirectPredictor : public SimObject
...
+     * Is called right after the prediction which updates the state
+     * speculatively. In case the branch was mispredicted the function
+     * is called again with the corrected information.
+     * The function is called for ALL branches as some predictors incooperate
+     * all branches in their history.
+     * @param PC The branch PC address.
+     * @param squash Whether the update is called at a misprediction
+     * @param taken Whether a conditional branch was taken
+     * @param target The target address if this branch.
+     * @param brType The branch instruction type.
```
andysan: This should be `br_type` to comply with the style guide.

src/cpu/pred/indirect.hh
```diff
@@ -51,21 +66,55 @@ class IndirectPredictor : public SimObject
...
+                                      Addr pc, void * &iHistory) = 0;
+
+    /**
+     * Updates the indirect predictor with history information of a branch.
+     * Is called right after the prediction which updates the state
+     * speculatively. In case the branch was mispredicted the function
+     * is called again with the corrected information.
+     * The function is called for ALL branches as some predictors incooperate
+     * all branches in their history.
+     * @param PC The branch PC address.
```
andysan: Inconsistent. The parameter is (correctly) `pc`.

src/cpu/pred/indirect.hh
```diff
@@ -51,21 +66,55 @@ class IndirectPredictor : public SimObject
...
+     * @param squash Whether the update is called at a misprediction
+     * @param taken Whether a conditional branch was taken
+     * @param target The target address if this branch.
+     * @param brType The branch instruction type.
+     * @param iHistory The pointer to the history object.
```
andysan: Should be `i_history`.

src/cpu/pred/indirect.hh
```diff
@@ -51,21 +66,55 @@ class IndirectPredictor : public SimObject
...
+    virtual void update(ThreadID tid, InstSeqNum sn, Addr pc, bool squash,
+                        bool taken, const PCStateBase& target,
+                        BranchType brType, void * &iHistory) = 0;
+
+    /**
+     * Squashes a branch. If the branch modified the history
+     * reverts the modification.
+     * @param tid Thread ID
+     * @param sn The sequence number of the branch.
+     * @param iHistory The pointer to the history object.
```
andysan: `i_history`

src/cpu/pred/simple_indirect.cc
andysan, line 1: Could you reorder the method implementations in this file to match the original order where reasonable? GitHub's diff tool gets really confused which makes it hard to review the changes.

andysan (commented):
This looks mostly OK, but it's hard to review for equivalence since some methods have been reordered. Could you reorder `simple_indirect.cc` to match the old order where there is a mostly equivalent method to make the diff smaller?

src/cpu/pred/BranchPredictor.py
```diff
@@ -100,6 +100,12 @@ class SimpleIndirectPredictor(IndirectPredictor):
     indirectPathLength = Param.Unsigned(
         3, "Previous indirect targets to use for path history"
     )
+    speculativePathLength = Param.Unsigned(
```
andysan, line 103: How does this interact with the O3 CPU? Would it make sense to make this unbounded instead (i.e., assuming it is sized to match the speculative capabilities of the CPU)?

src/cpu/pred/simple_indirect.cc
```diff
@@ -64,172 +76,288 @@ SimpleIndirectPredictor::SimpleIndirectPredictor(
...
+                    "ghr:%#x, pathHist sz:%#x\n",
+                    sn, history->pcAddr, history->was_indirect,
+                    history->ghr,
+                    threadInfo[tid].pathHist.size());
 
-    if (t_info.pathHist.empty()) return;
 
-    if (t_info.headHistEntry < t_info.pathHist.size() &&
-        t_info.pathHist[t_info.headHistEntry].seqNum <= seq_num) {
-        if (t_info.headHistEntry >= pathLength) {
```
dhschall, line 143: Before, there was no speculative path length only path length plus and additional pointer to the last element.
However, the pointer is only incremented and never decreased. IIRC this caused problems if you speculate.
Looking at the code now it seems like a bug to me??
The `headHistEntry` entry is set initially to 0. 
Once the path grows you increment the pointer.
Once the pointer is greater than 3 you will pop this an entry every time you commit a branch.

src/cpu/pred/BranchPredictor.py
dhschall: See my comments in the code where its used

src/cpu/pred/simple_indirect.cc
```diff
@@ -64,172 +76,292 @@ SimpleIndirectPredictor::SimpleIndirectPredictor(
...
+    history->pcAddr = pc;
+    history->was_indirect = true;
+
+    /** Do the prediction for indirect branches (no returns) */
+    PCStateBase* target = nullptr;
+    history->hit = lookup(tid, pc, target, history);
+    return target;
 }
 
 bool
```
dhschall, line 140: I think I should merge these two functions into one.
If there is just one caller there is no need or not?

andysan: Keeping them separate seems ok to me.

andysan (commented):
This looks OK to me. I'll approve this once @rjc-arch is has run some performance validation.

dhschall:
Agree, lets wait for the tests.
Thank you @andysan

rjc-arch:
We have now finished running performance regressions using SPEC 2017 SimPoints, and can confirm that the IPCs reported by gem5 are not changed by this refactoring.

andysan (approved):

PR #431, opened 2023-10-10
misc,python: Add `isort` hook to pre-commit

BobbyRBruce:


andysan:
There are a couple of places where isort seems to be mixing modules in the local directory and in gem5 itself. That seems wrong to me. We might want to try to add this to the configuration:
```
known_first_party = m5
```

Alternatively, we could do something like this (I think):
```
known_m5 = m5
known_gem5 = gem5
sections=FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,M5,GEM5,LOCALFOLDER
```

I can't remember what the style guide says off the top of my head, but it should be something like the above.

andysan:
Also, I would be fine not including this in the blame ignore file. The change is pretty pretty limited in scope and it won't mess with git blame that much.

BobbyRBruce:
> There are a couple of places where isort seems to be mixing modules in the local directory and in gem5 itself. That seems wrong to me. We might want to try to add this to the configuration:
> 
> ```
> known_first_party = m5
> ```
> 
> Alternatively, we could do something like this (I think):
> 
> ```
> known_m5 = m5
> known_gem5 = gem5
> sections=FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,M5,GEM5,LOCALFOLDER
> ```
> 
> I can't remember what the style guide says off the top of my head, but it should be something like the above.

I've incorporated this. See if it works for you.

andysan:
> > There are a couple of places where isort seems to be mixing modules in the local directory and in gem5 itself. That seems wrong to me. We might want to try to add this to the configuration:
> > ```
> > known_first_party = m5
> > ```
> > 
> > 
> > Alternatively, we could do something like this (I think):
> > ```
> > known_m5 = m5
> > known_gem5 = gem5
> > sections=FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,M5,GEM5,LOCALFOLDER
> > ```
> > 
> > 
> > I can't remember what the style guide says off the top of my head, but it should be something like the above.
> 
> I've incorporated this. See if it works for you.

There is something weird going on in `configs/`. It seems like `common` isn't detected as a local folder for some reason. Not sure what to do about that though. We might need to use relative imports or do something creative here if we want it to be consistent.

Here are some more thoughts:

- Do we want to order `m5.params` and `m5.proxy` with respect to `m5.objects`? Most existing files seem to have params and proxy imports before object imports. Treating them as just another m5 thing works for me.
- Do we want to put `_m5`  in its own block? It's very much internal plumbing, so it might make sense, but the current config seems OK as well.
- We should probably explicitly tell isort what to do about `slicc`. Should it be in its own group, m5, or firstparty?

ivanaamit:
@andysan

> Do we want to put _m5 in its own block? It's very much internal plumbing, so it might make sense, but the current config seems OK as well.

I made this change. Please take a look to see if that is what you had in mind.


> We should probably explicitly tell isort what to do about slicc. Should it be in its own group, m5, or first-party?

Do you have a suggestion for which solution would be the best?

andysan:
> @andysan
> 
> > Do we want to put _m5 in its own block? It's very much internal plumbing, so it might make sense, but the current config seems OK as well.
> 
> I made this change. Please take a look to see if that is what you had in mind.

This is exactly what I had in mind, thanks! I think this makes sense.
 
> > We should probably explicitly tell isort what to do about slicc. Should it be in its own group, m5, or first-party?
> 
> Do you have a suggestion for which solution would be the best?

Maybe we should just treat slicc as if it belongs to _m5? It might even make sense to move it into that namespace at some if it isn't used by external scripts.

BobbyRBruce:
> > @andysan
> > > Do we want to put _m5 in its own block? It's very much internal plumbing, so it might make sense, but the current config seems OK as well.
> > 
> > 
> > I made this change. Please take a look to see if that is what you had in mind.
> 
> This is exactly what I had in mind, thanks! I think this makes sense.
> 
> > > We should probably explicitly tell isort what to do about slicc. Should it be in its own group, m5, or first-party?
> > 
> > 
> > Do you have a suggestion for which solution would be the best?
> 
> Maybe we should just treat slicc as if it belongs to _m5? It might even make sense to move it into that namespace at some if it isn't used by external scripts.

I've added `slicc` to be bundled with `_m5`.

configs/common/Options.py
```diff
@@ -38,13 +38,13 @@
 
 import argparse
 
+from common import ObjectList
```
andysan, line 41: There is something weird happening here. It seems like isort isn't detecting this as a local import. It might be possible to fix this by doing something like `from . import ObjectList` instead. IIRC, there was a reason for including the full path here and it had to do with Python 2.7 compatibility.

There are similar issues in multiple example configuration files where local imports get confused.

I think this can be solved using relative imports. That should tell isort that it is a local import and it would avoid the ugly import path fixup.

andysan (approved):
There is still some weirdness in the example configurations. However, I don't think that needs to block this change. I would suggest we merge this as is and fix up the examples later.

andysan:
This PR has been sitting around for a while without much input from anyone other than Bobby and Ivana. I'm going to go ahead and merge this since I think it is good enough and we'll likely end up with annoying conflicts if we wait.

PR #432, opened 2023-10-11
tests,misc: Add "build/ALL/gem5.fast" Clang compilation to CI

BobbyRBruce:
While we do run compiler tests weekly, 9/10 the issue is a strict check in clang we did not check before incorporating code into the codebase.

Therefore, running a clang compilation as part of our CI would help us catch errors quicker.

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -89,6 +89,22 @@ jobs:
             build-matrix: ${{ steps.build-matrix.outputs.build-matrix }}
             test-dirs-matrix: ${{ steps.dir-matrix.outputs.test-dirs-matrix }}
 
+    clang-fast-compilation:
+        # gem5 binaries built in `quick-gem5-builds` always use GCC.
+        # Clang is more strict than GCC. This job checks that gem5 compiles
+        # with Clang. It compiles build/ALL/gem5.fast to maximize the change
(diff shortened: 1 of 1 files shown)


BobbyRBruce:
We need to hold off on this until our compiler tests are passing again...

.github/workflows/ci-tests.yaml
```diff
@@ -89,6 +89,22 @@ jobs:
...
             test-dirs-matrix: ${{ steps.dir-matrix.outputs.test-dirs-matrix }}
 
+    clang-fast-compilation:
+        # gem5 binaries built in `quick-gem5-builds` always use GCC.
+        # Clang is more strict than GCC. This job checks that gem5 compiles
+        # with Clang. It compiles build/ALL/gem5.fast to maximize the change
+        # for compilation error to be exposed.
+        runs-on: [self-hosted, linux, x64]
+        if: github.event.pull_request.draft == false
+        container: ghcr.io/gem5/clang-version-11:latest
```
powerjg: Why not do clang 16 (or the highest we support)? That seems to be where we get the most errors.

BobbyRBruce: Just wanted to make sure 16 was working first before doing this. It is so have updated.

ivanaamit (approved):

PR #433, opened 2023-10-11
configs: GPUFS option to disable KVM perf counters

abmerop:
Add a --no-kvm-perf option to disable KVM perf counters for GPUFS scripts. This is useful for users who have KVM enabled but configured with more restrictive settings, which seems to be the default in newer Linux distros.

Change-Id: I7508113d0f7c74deb21ea7b2770522885a0ec822


mattsinc (approved):
This violates Jason's request not to add more Option based parameters, but given the lack of GPU board, I don't see an alternative in the short-term.

PR #439, opened 2023-10-11
arch-vega: Implement buffer_atomic_cmpswap

abmerop:
This is a standard compare and swap but implemented on vector memory buffer instructions (i.e., it is the same as FLAT_ATOMIC_CMPSWAP with MUBUF's special address calculation).

This was tested using a Tensile kernel, a backend for rocBLAS, which is used by PyTorch and Tensorflow. Prior to this patch both ML frameworks crashed. With this patch they both make forward progress.

Change-Id: Ie76447a72d210f81624e01e1fa374e41c2c21e06

### DIFF
--- a/src/arch/amdgpu/vega/insts/instructions.cc
+++ b/src/arch/amdgpu/vega/insts/instructions.cc
@@ -40581,8 +40581,87 @@ namespace VegaISA
     void
     Inst_MUBUF__BUFFER_ATOMIC_CMPSWAP::execute(GPUDynInstPtr gpuDynInst)
     {
-        panicUnimplemented();
+        Wavefront *wf = gpuDynInst->wavefront();
+
+        if (gpuDynInst->exec_mask.none()) {
+            wf->decVMemInstsIssued();
+            return;
+        }
+
+        gpuDynInst->execUnitId = wf->execUnitId;
+        gpuDynInst->latency.init(gpuDynInst->computeUnit());
+        gpuDynInst->latency.set(gpuDynInst->computeUnit()->clockPeriod());
+
+        ConstVecOperandU32 addr0(gpuDynInst, extData.VADDR);
+        ConstVecOperandU32 addr1(gpuDynInst, extData.VADDR + 1);
+        ConstScalarOperandU128 rsrcDesc(gpuDynInst, extData.SRSRC * 4);
+        ConstScalarOperandU32 offset(gpuDynInst, extData.SOFFSET);
+        ConstVecOperandU32 src(gpuDynInst, extData.VDATA);
+        ConstVecOperandU32 cmp(gpuDynInst, extData.VDATA + 1);
+
+        rsrcDesc.read();
+        offset.read();
+        src.read();
+        cmp.read();
+
+        int inst_offset = instData.OFFSET;
+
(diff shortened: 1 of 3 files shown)


src/arch/amdgpu/vega/insts/instructions.cc
```diff
@@ -40581,8 +40581,87 @@ namespace VegaISA
...
+        Wavefront *wf = gpuDynInst->wavefront();
+
+        if (gpuDynInst->exec_mask.none()) {
+            wf->decVMemInstsIssued();
+            return;
+        }
+
+        gpuDynInst->execUnitId = wf->execUnitId;
+        gpuDynInst->latency.init(gpuDynInst->computeUnit());
+        gpuDynInst->latency.set(gpuDynInst->computeUnit()->clockPeriod());
```
mattsinc, line 40593: I can't figure out why this line is way off -- indenting problem?

src/arch/amdgpu/vega/insts/instructions.cc
```diff
@@ -40581,8 +40581,87 @@ namespace VegaISA
...
+
+        ConstVecOperandU32 addr0(gpuDynInst, extData.VADDR);
+        ConstVecOperandU32 addr1(gpuDynInst, extData.VADDR + 1);
+        ConstScalarOperandU128 rsrcDesc(gpuDynInst, extData.SRSRC * 4);
```
mattsinc, line 40597: is the * 4 just for the 128B?

src/arch/amdgpu/vega/insts/instructions.cc
```diff
@@ -40581,8 +40581,87 @@ namespace VegaISA
...
+        for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
+            if (gpuDynInst->exec_mask[lane]) {
+                (reinterpret_cast<VecElemU32*>(gpuDynInst->x_data))[lane]
+                    = src[lane];
+                (reinterpret_cast<VecElemU32*>(gpuDynInst->a_data))[lane]
+                    = cmp[lane];
+            }
+        }
+
+        gpuDynInst->computeUnit()->globalMemoryPipe.issueRequest(gpuDynInst);
```
mattsinc, line 40640: indenting problem?

mattsinc (changes requested):

src/arch/amdgpu/vega/insts/instructions.cc
abmerop: Hmm, not seeing any indention issues on either the github view nor in my text editor

src/arch/amdgpu/vega/insts/instructions.cc
abmerop: The resource descriptor is always 4 SGPRs and must be dword aligned.  The field itself is 5 bits wide so I assume this is done to use few bits in the instruction encoding.  This is described in sections 8.1.2 and 12.14 in the Vega ISA manual

src/arch/amdgpu/vega/insts/instructions.cc
mattsinc: Weird ... going to approve then.

mattsinc (approved):

PR #443, opened 2023-10-12
arch-riscv: Fix write back register issue of vmask_mv_micro

rogerchang23424:
After removing the setRegOperand in VecRegOperand
https://github.com/gem5/gem5/pull/341. The vmask_vm_micro will not write back to register because tmp_d0 is not the reference type. The PR will make tmp_d0 as reference of regFile.

Change-Id: I2a934ad28045ac63950d4e2ed3eecc4a7d137919

### DIFF
--- a/src/arch/riscv/insts/vector.hh
+++ b/src/arch/riscv/insts/vector.hh
@@ -555,7 +555,7 @@ class VMaskMergeMicroInst : public VectorArithMicroInst
     Fault
     execute(ExecContext* xc, trace::InstRecord* traceData) const override
     {
-        vreg_t tmp_d0 = *(vreg_t *)xc->getWritableRegOperand(this, 0);
+        vreg_t& tmp_d0 = *(vreg_t *)xc->getWritableRegOperand(this, 0);
         auto Vd = tmp_d0.as<uint8_t>();
         constexpr uint8_t elems_per_vreg = VLENB / sizeof(ElemType);
         size_t bit_cnt = elems_per_vreg;


hnpl (approved):
lgtm

Out of my curiosity, can you provide some information about how the rvv instructions are tested?

rogerchang23424:
Just simply compile the `rvv_branch.c` and `rvv_reduce.c` and with simple RISC-V SE script config.

rvv_branch.c:
https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/3f0ac66eea7799a072d47e33efed28b3c000d291/examples/rvv_branch.c

rvv_reduce.c:
https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/3f0ac66eea7799a072d47e33efed28b3c000d291/examples/rvv_reduce.c

They contain the vector mask instruction `vmfne.vf` to verify the implementation of micro instruction `vmask.mv`.

Compile command: (Require gcc 13 to complie vector instructinos)
```bash
riscv64-unknown-elf-gcc -march=rv64gcv -static rvv_branch.c -o rvv_branch.elf
```

Run simulation:
```bash
build/ALL/gem5.opt /path/to/riscv-se-config.py
```

The test is passed if the output screen print "pass". Ohterwise, print "fail".

hnpl (approved):
lgtm

powerjg (approved):

hnpl:
> riscv64-unknown-elf-gcc

Is this cross compiler available on any linux distribution?

rogerchang23424:
I don't install them from linux distribution. I build the compuler
from［riscv toolchain(https://github.com/riscv-collab/riscv-gnu-toolchain),
it should work for all linux platform.

Hoa Nguyen ***@***.***> 於 2023年10月15日 週日 06:00 寫道：

hnpl:
Thanks for the info. Can you elaborate more on how to compile the cross compiler? I used to use the `rvv-next` branch from the above repo but the branch has been deleted.

I compiled the `master` branch using the following commands after cloning the above repo but the compiler does not perform vectorization on my workload, which does not have vector intrinsics.

```sh
RUN git clone --recursive https://github.com/riscv/riscv-gnu-toolchain -j`nproc`

WORKDIR /riscv/riscv-gnu-toolchain/
RUN ./configure --prefix=/riscv/_install --enable-multilib --with-arch=rv64gcv
RUN make linux -j`nproc`
RUN make install
```

I used the following command to compile the binary but the binary does not contain any vector instructions.

```sh
/riscv/_install/bin/riscv64-unknown-linux-gnu-gcc -O2 -fopenmp -march=rv64gcv -ftree-vectorize -fopt-info-vec-all -fsimd-cost-model=unlimited -fvect-cost-model=unlimited -c stream_kernels.c -o ./stream_kernels.rvv.o
```

rogerchang23424:
Sorry for lately reply. The gcc-13 only support `vector intrinsic` feature and it doesn't support `auto-vectorization` because the patch is submitted to late [link](https://gcc.gnu.org/pipermail/gcc-patches/2023-March/613260.html). I hope the feature will support in GCC 14.

hnpl:
Hey Roger, thanks for the information. I'm looking forward to the next release too!

PR #444, opened 2023-10-12
misc,tests: Add dummy jobs to workflows for status checks

BobbyRBruce:
With this change we can block PRs if the compiler, daily, and/or weekly tests are failing by setting these dummy jobs as require status checks.


powerjg (commented):
Probably shouldn't run this on self hosted

BobbyRBruce:
> Probably shouldn't run this on self hosted

Fixed.

powerjg (approved):

PR #445, opened 2023-10-12
stdlib: Extend SE Binary Workload std file params to accept strings

kunpai:
This patch improves the stdin_file, stdout_file, stderr_file processing in SE Binary Workload.

Earlier, it assumed that Paths would be passed into the function.

Therefore, these parameters could not be defined
in the Resource JSON file as a string to use as is, since the processing functions like as_posix() would fail.

This is a current pain point of trying to make a Suite JSON of SPEC Resources, since they use the stdin, stdout and stderr files.

This patch converts the string of the Path in the JSON file into a Path object, while also retaining the old functionality of allowing Paths to be passed in.

The processing for both remains the same.

Change-Id: I8fd7d385ff301549a7db2166c9ad0174ec0c201a

### DIFF
--- a/src/python/gem5/components/boards/se_binary_workload.py
+++ b/src/python/gem5/components/boards/se_binary_workload.py
@@ -111,10 +111,8 @@ class SEBinaryWorkload:
             else:  # stdin_file is a string
                 process.input = FileResource(Path(stdin_file)).get_local_path()
         if stdout_file is not None:
-            if isinstance(stdout_file, Path):
-                process.output = stdout_file.as_posix()
-            else:  # stdout_file is a string
-                process.output = Path(stdout_file).as_posix()
+            assert isinstance(stdout_file, Path) or isinstance(stdout_file, str)
+            process.output = str(stdout_file)
         if stderr_file is not None:
             if isinstance(stderr_file, Path):
                 process.errout = stderr_file.as_posix()
--- a/src/python/gem5/components/boards/se_binary_workload.py
+++ b/src/python/gem5/components/boards/se_binary_workload.py
@@ -39,6 +39,7 @@ from ..processors.switchable_processor import SwitchableProcessor
 
 from gem5.resources.elfie import ELFieInfo
 from gem5.resources.looppoint import Looppoint
+from gem5.resources.resource import FileResource
 
 from m5.objects import SEWorkload, Process
 
@@ -65,9 +66,9 @@ class SEBinaryWorkload:
         self,
         binary: BinaryResource,
         exit_on_work_items: bool = True,
-        stdin_file: Optional[FileResource] = None,
-        stdout_file: Optional[Path] = None,
-        stderr_file: Optional[Path] = None,
+        stdin_file: Optional[Union[FileResource, str]] = None,
+        stdout_file: Optional[Union[Path, str]] = None,
+        stderr_file: Optional[Union[Path, str]] = None,
         env_list: Optional[List[str]] = None,
         arguments: List[str] = [],
         checkpoint: Optional[Union[Path, CheckpointResource]] = None,
@@ -105,13 +106,20 @@ class SEBinaryWorkload:
         process.executable = binary_path
         process.cmd = [binary_path] + arguments
         if stdin_file is not None:
-            process.input = stdin_file.get_local_path()
+            if isinstance(stdin_file, Path):
+                process.input = FileResource(stdin_file).get_local_path()
+            else:  # stdin_file is a string
+                process.input = FileResource(Path(stdin_file)).get_local_path()
         if stdout_file is not None:
-            process.output = stdout_file.as_posix()
+            if isinstance(stdout_file, Path):
+                process.output = stdout_file.as_posix()
+            else:  # stdout_file is a string
+                process.output = Path(stdout_file).as_posix()
         if stderr_file is not None:
-            process.errout = stderr_file.as_posix()
-        if env_list is not None:
-            process.env = env_list
+            if isinstance(stderr_file, Path):
+                process.errout = stderr_file.as_posix()
+            else:  # stderr_file is a string
+                process.errout = Path(stderr_file).as_posix()
 
         if isinstance(self.get_processor(), SwitchableProcessor):
             # This is a hack to get switchable processors working correctly in


src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -105,13 +106,17 @@ def set_se_binary_workload(
         process.executable = binary_path
         process.cmd = [binary_path] + arguments
         if stdin_file is not None:
-            process.input = stdin_file.get_local_path()
+            if isinstance(stdin_file, Path):
+                process.input = FileResource(stdin_file).get_local_path()
+            process.input = FileResource(Path(stdin_file)).get_local_path()
```
BobbyRBruce: I don't understand this.

The construction of this function still says `stdin_file` is a `Optional[FileResource]` defaulted to `None`, so if someone is passing `stdin_file` of type `Path` they are just wrong. 



I get the feeling here there's implication this is either a `str` or `Path` and this handles the conversion. Though in the case where `stdin_file` is a Path you're setting `process.input` on line 110 then immediately overriding it on line 111.

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -105,13 +106,17 @@ def set_se_binary_workload(
...
         if stdout_file is not None:
-            process.output = stdout_file.as_posix()
+            if isinstance(stdout_file, Path):
+                process.output = stdout_file.as_posix()
+            process.output = Path(stdout_file).as_posix()
```
BobbyRBruce: Almost exactly the same comments here. This should've always been a Path because that's its type.

Also you're overriding the value you set on 114 on 115.

BobbyRBruce (changes requested):
First of all, this PR title is really confusing. It took me a while to know what you mean by "JSON support". You mean in the very niche case we are setting up  resources (inc. workloads) by reading a JSON files.

It also took me a while to get the problem here generally but, essentially, what you're saying is this: the path is defined in the JSON file as a string and you need to cast it to a `Path` before se_binary_workload as it only accepted `Path` objects.

I don't really see where the problem is here. All this PR attempted to do is do the casing inside the Workload functiomn instead of before. Why don't you just update the code you're wriging to handle the parsing of this JSON file?

While I agree we have **terrible** enforcement of using `Path` over passing `str`s when it comes to filepaths, I do think having a `Path` type should be the standard and I don't really want to start jumping in and trying to support strings anymore than we do.



I don't unerstand the code much at all I'm afraid. You're changing `stdin_file` but it was always of type `Optional[FileResource]` anyway, so what was the problem with `Path`?

kunpai:
> First of all, this PR title is really confusing. It took me a while to know what you mean by "JSON support". You mean in the very niche case we are setting up resources (inc. workloads) by reading a JSON files.
> 
> It also took me a while to get the problem here generally but, essentially, what you're saying is this: the path is defined in the JSON file as a string and you need to cast it to a `Path` before se_binary_workload as it only accepted `Path` objects.
> 
> I don't really see where the problem is here. All this PR attempted to do is do the casing inside the Workload functiomn instead of before. Why don't you just update the code you're wriging to handle the parsing of this JSON file?
> 
> While I agree we have **terrible** enforcement of using `Path` over passing `str`s when it comes to filepaths, I do think having a `Path` type should be the standard and I don't really want to start jumping in and trying to support strings anymore than we do.
> 
> I don't unerstand the code much at all I'm afraid. You're changing `stdin_file` but it was always of type `Optional[FileResource]` anyway, so what was the problem with `Path`?

Hey Bobby,

Thanks for the feedback! I understand that the commit message was not descriptive, and I will work on it.

As for why this change is necessary, I think the exact use case for it could help:

Since Suites are now a part of gem5, we would like to have a Suite of SPEC2006. This introduces certain pain points, and those are:

1. We cannot distribute SPEC and yet, everyone on our network should be able to use it. Therefore, we require to use the local JSON support provided by gem5.
2. While there is a script to create CustomWorkload resources from the JSONs, the way that script works is by using `Path` and converting those strs of stdin_file into `Path` objects. While useful, it cannot be used with `obtain_resource` or through any of the JSON env variables we provide: `GEM5_RESOURCE_JSON_APPEND` and `GEM5_JSON_APPEND`, IIRC since they do not follow the schema. They can only be referenced through CustomWorkload, as far as I have been testing.

As a way to standardize those WorkloadResources to follow the schema, I found that since these file paths are strings, and the se_binary_workload expects them to be a Path object. The CustomWorkloads I mentioned above passed these file paths in as parameters converted into Path objects already, which is why they worked. I want to eliminate this need of creating a custom Workload to begin with, so anyone can add the JSON to their gem5 data sources, mention `obtain_resource("spec2006-gcc")` or something, and it would work flawlessly.

The easiest solution I could think of was to add support for strings, which is what I have done here.

But, I also understand that there may be a more elegant way of doing it. Do you have an idea that could help me in this use case?

src/python/gem5/components/boards/se_binary_workload.py
BobbyRBruce: @kunpai You still haven't addressed this.

src/python/gem5/components/boards/se_binary_workload.py
BobbyRBruce: @kunpai : Ditto here.

BobbyRBruce:
> > First of all, this PR title is really confusing. It took me a while to know what you mean by "JSON support". You mean in the very niche case we are setting up resources (inc. workloads) by reading a JSON files.
> > It also took me a while to get the problem here generally but, essentially, what you're saying is this: the path is defined in the JSON file as a string and you need to cast it to a `Path` before se_binary_workload as it only accepted `Path` objects.
> > I don't really see where the problem is here. All this PR attempted to do is do the casing inside the Workload functiomn instead of before. Why don't you just update the code you're wriging to handle the parsing of this JSON file?
> > While I agree we have **terrible** enforcement of using `Path` over passing `str`s when it comes to filepaths, I do think having a `Path` type should be the standard and I don't really want to start jumping in and trying to support strings anymore than we do.
> > I don't unerstand the code much at all I'm afraid. You're changing `stdin_file` but it was always of type `Optional[FileResource]` anyway, so what was the problem with `Path`?
> 
> Hey Bobby,
> 
> Thanks for the feedback! I understand that the commit message was not descriptive, and I will work on it.

Please do :).

> 
> As for why this change is necessary, I think the exact use case for it could help:
> 
> Since Suites are now a part of gem5, we would like to have a Suite of SPEC2006. This introduces certain pain points, and those are:
> 
> 1. We cannot distribute SPEC and yet, everyone on our network should be able to use it. Therefore, we require to use the local JSON support provided by gem5.
> 2. While there is a script to create CustomWorkload resources from the JSONs, the way that script works is by using `Path` and converting those strs of stdin_file into `Path` objects. While useful, it cannot be used with `obtain_resource` or through any of the JSON env variables we provide: `GEM5_RESOURCE_JSON_APPEND` and `GEM5_JSON_APPEND`, IIRC since they do not follow the schema. They can only be referenced through CustomWorkload, as far as I have been testing.
> 
> As a way to standardize those WorkloadResources to follow the schema, I found that since these file paths are strings, and the se_binary_workload expects them to be a Path object. The CustomWorkloads I mentioned above passed these file paths in as parameters converted into Path objects already, which is why they worked. I want to eliminate this need of creating a custom Workload to begin with, so anyone can add the JSON to their gem5 data sources, mention `obtain_resource("spec2006-gcc")` or something, and it would work flawlessly.

Ok, I still don't completely understand but let me throw some stuff out there which might help clear things up. This is how I (partially) understand things:

We define a Workload, essentially, in a JSON object (whether in an actual JSON file or as part of a MongoDB entry, it doesn't matter). This object gives the "set_workload" function to call and the parameters we pass. So roughly something like this:

```json
{
    "function": "set_kernal_disK_workload",
    "parameters" : {
         "kernel": {
              "function": "obtain_resource",
              "parameters" : {
                   "id" : "some_kernel",
                   "version" : "1"
              }
          }
          "disk_image": {
              "function": "obtain_resource",
              "parameters" : {
                   "id" : "an_image",
                   "version" : "2"
              }
          },
          "readfile_contents" : "echo 'hello world'"
     }
}
```

In the above example we want to call "set_kernal_disk_workload" and set the `kernel` parameter to a function call to `obtain_resources`, and the same for the disk image. The `readfile_contents` is just a string so no function call required. I realize this probably isn't _exactly_ what we do. I can't remember right now the specific details, but the main point is a parameter can either be a json primary (int, str, etc), or a function call (note: an object is constructed via function call, so we can pass objects). There's probably some corner cases we can't cover with this, but it'll allow us to define most function calls to "set_workload" functions we care about.


Where you see a problem is the [`set_se_binary_workload` function](https://github.com/gem5/gem5/blob/0442c9a88c5e0031d766470c76d1cd8138b313cb/src/python/gem5/components/boards/se_binary_workload.py#L64) where the parameters are as follows:

```py
def set_se_binary_workload(
    self,
    binary: BinaryResource,
    exit_on_work_items: bool = True,
    stdin_file: Optional[FileResource] = None,
    stdout_file: Optional[Path] = None,
    stderr_file: Optional[Path] = None,
    env_list: Optional[List[str]] = None,
    arguments: List[str] = [],
    checkpoint: Optional[Union[Path, CheckpointResource]] = None,
) -> None:
```

We have these `Optional[Path]` parameters. I _think_ you're saying we have some translation code to take strings from the Workload's JSON object and convert them to `Path`s. so something like this:

```json
{
    "function": "set_se_binary_workload",
    "parameters" : {
         "stdout_file": "/home/bbruce/stdoutfile",
          "other parames..." : "bla... bla blla"
     }
}
```

`stdout_file` is a string but we converted to a `Path` object before actually calling `set_se_binary_workload`.

Your problem is this transation code isn't available when we import JSON files via this environment variable (i don't know why this would be the case, but i'll go with it).

So your solution is these parameters can be either `Path` objects or strings, and we just convert in the `set_se_binary_workload` function as needed.


> 
> The easiest solution I could think of was to add support for strings, which is what I have done here.
> 
> But, I also understand that there may be a more elegant way of doing it. Do you have an idea that could help me in this use case?

If what i wrote above is correct, I don't like that there's some magic hardcoding which translates strings to `Path` objects arbitrarily for workloads. it ruins the flexibility of the `Workload` JSON objects which is pretty de-coupled from the workload functions. Ideally we'd accept one of two solutions:

1. The `set_se_binary_workload` should accept these values as strings. This means our JSON representation uses strings also.
2. The `set_se_binary_workload` function accepts `Path` objects and the Workload JSON entry is updated to reflect this and we remove the string to Path translation translation.

(1.) isn't great because we really should be using `Path`s instead of strings (though we pass strings around a lot already so 🤷‍♂️), and it changes the API. (2.) is better for me but it'd mean updating all the Workload JSON objects that use the `set_se_binary_workload` function to construct `Path` objects.

For now, it _may_ be tollerable to just have these parameters accept `Path` objects or strings but this should be documented in the codebase on why we're doing this. It's messy and confusing. Also, please see my code-review comments because I think your conversion is broken.

The part of your solution i really don't understand is why you state the user gem5-resources JSON files (those imported via ENV variables) supposedly conform to a different schema than the database. That should not be the case. If it is, that's worrying. That means we have to maintain two seperate schemas to repersent the same data. Can you check what's going on here?

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -105,13 +106,20 @@ def set_se_binary_workload(
         process.executable = binary_path
         process.cmd = [binary_path] + arguments
         if stdin_file is not None:
-            process.input = stdin_file.get_local_path()
+            if isinstance(stdin_file, Path):
```
BobbyRBruce, line 109: `stdin_file` is never of type `Path` it's either `FileResource` or `str`.

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -105,13 +106,20 @@ def set_se_binary_workload(
...
+                process.input = FileResource(stdin_file).get_local_path()
+            else:  # stdin_file is a string
+                process.input = FileResource(Path(stdin_file)).get_local_path()
         if stdout_file is not None:
-            process.output = stdout_file.as_posix()
+            if isinstance(stdout_file, Path):
+                process.output = stdout_file.as_posix()
+            else:  # stdout_file is a string
+                process.output = Path(stdout_file).as_posix()
```
BobbyRBruce: Do the folllowing:

```suggestion
            assert isinstance(stdout_file, Path) or isinstance(stdout_file, str)
            process.output = str(stdout_file)
```

Reasons:

1. If you want to get the Path as a string then you  just need to cast it to a string.
2. `Path(stdout_file).as_posix()` is just converting a string to a Path then returning a string.
3. `as_posix` is not what we want to do here. Its job is to convert non-unix-style paths (for example, Windows paths which contain backslashes) to unix-style ("posix") paths. Someone introduced this to the codebase at some point and while it doesn't do any harm, it doesn't do any good either.

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -105,13 +106,20 @@ def set_se_binary_workload(
...
         if stderr_file is not None:
-            process.errout = stderr_file.as_posix()
-        if env_list is not None:
-            process.env = env_list
+            if isinstance(stderr_file, Path):
+                process.errout = stderr_file.as_posix()
+            else:  # stderr_file is a string
+                process.errout = Path(stderr_file).as_posix()
```
BobbyRBruce, line 120: Do the same thing here.

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -105,13 +106,20 @@ def set_se_binary_workload(
         process.executable = binary_path
         process.cmd = [binary_path] + arguments
         if stdin_file is not None:
-            process.input = stdin_file.get_local_path()
+            if isinstance(stdin_file, Path):
+                process.input = FileResource(stdin_file).get_local_path()
+            else:  # stdin_file is a string
+                process.input = FileResource(Path(stdin_file)).get_local_path()
```
BobbyRBruce, line 112: This is circular. Why not just `process.input = stdin_file` ?

BobbyRBruce (changes requested):

Issue #403, opened 2023-10-05
Make cache recorder future proof and more elegant


v-ramadas:
The cache recorder is used during cache cooldown and warmup to either issue flushes to memory or populate cache. It uses a sequencer map to track all the L1 caches (https://github.com/gem5/gem5/blob/stable/src/mem/ruby/system/RubySystem.cc#L180) and calls their makeRequest() (https://github.com/gem5/gem5/blob/stable/src/mem/ruby/system/CacheRecorder.cc#L100) function during cooldown/warmup. Pull request #377 extends this support to GPU coalescers by adding a coalescer map to the cache recorder. It does so by duplicating code within the cache recorder to both populate the coalescer map and issue requests to the coalescer. If some other component requires flush support in the future, this approach would lead to more code duplication.

Would it be a good idea to maintain a list of RubyPort instances instead of separate lists for sequencers and gpu coalescers; i.e. use `std::vector<RubyPort>` instead of `std::vector<Sequencer>` at https://github.com/gem5/gem5/blob/stable/src/mem/ruby/system/RubySystem.cc#L180? Since makeRequest() is virtual in RubyPort and overriden by derived classes Sequencer and GPUCoalescer, the RubyPort map can be populated with pointers to either the sequencer or coalescer (or any similar component), and no code duplication is needed when issuing the makeRequest() call. Would this be a good (more elegant) alternative?

powerjg:
Sounds like a good idea to me!

BobbyRBruce:
Is it safe to close this issue?

mattsinc:
@v-ramadas should confirm, but since #448 was merged in, yes I believe this can be closed.

v-ramadas:
Yes, will close the issue.

PR #448, opened 2023-10-12
mem-ruby: Update cache recorder to use RubyPort and remove BUILD_GPU guards

v-ramadas:
This PR updates cache recorder to use a vector of RubyPorts for cache cooldown and warmup instead of Sequencer or GPUCoalescer vectors (refer to issue #403 for more details). It also removes the extra guards that were added in #377 to prevent compile-time failures in non-GPU builds.


powerjg (approved):
This looks great! Thanks for the quick fix!

abmerop (approved):

mattsinc (approved):

PR #455, opened 2023-10-14
cpu: Branch Predictor Refactoring

dhschall:
Major refactoring of the branch predictor unit.
- Clearer control flow of the main branch predictor
- Remove `uncondBranch` and `btbUpdate` functions in favor
  of a common `historyUpdate` function. There is now only
  one lookup function for conditional branches and the new
  `historyUpdate` for speculative history update.
- Added a new *target provider* class.
- More expressive statistics depending on the different branch
  types.
- Cleanup the branch history management

### DIFF
--- a/src/cpu/pred/BranchPredictor.py
+++ b/src/cpu/pred/BranchPredictor.py
@@ -57,6 +57,15 @@ class BranchType(Enum):
     ]
 
 
+class TargetProvider(Enum):
+    vals = [
+        "NoTarget",
+        "BTB",
+        "RAS",
+        "Indirect",
+    ]
+
+
 class ReturnAddrStack(SimObject):
     type = "ReturnAddrStack"
     cxx_class = "gem5::branch_prediction::ReturnAddrStack"
--- a/src/cpu/pred/bi_mode.cc
+++ b/src/cpu/pred/bi_mode.cc
@@ -1,4 +1,16 @@
 /*
+ * Copyright (c) 2022-2023 The University of Edinburgh
+ * All rights reserved
+ *
+ * The license below extends only to copyright in the software and shall
+ * not be construed as granting a license to any other intellectual
+ * property including but not limited to intellectual property relating
+ * to a hardware implementation of the functionality of the software
+ * licensed hereunder.  You may use the software subject to the license
+ * terms below provided that you ensure that this notice is replicated
+ * unmodified and in its entirety in all distributions of the software,
+ * modified or unmodified, in source code or in binary form.
+ *
  * Copyright (c) 2014 The Regents of The University of Michigan
  * All rights reserved.
  *
@@ -73,7 +85,7 @@ BiModeBP::BiModeBP(const BiModeBPParams &params)
  * chooses the taken array and the taken array predicts taken.
  */
 void
-BiModeBP::uncondBranch(ThreadID tid, Addr pc, void * &bpHistory)
+BiModeBP::uncondBranch(ThreadID tid, Addr pc, void * &bp_history)
 {
     BPHistory *history = new BPHistory;
     history->globalHistoryReg = globalHistoryReg[tid];
@@ -81,17 +93,29 @@ BiModeBP::uncondBranch(ThreadID tid, Addr pc, void * &bpHistory)
     history->takenPred = true;
     history->notTakenPred = true;
     history->finalPred = true;
-    bpHistory = static_cast<void*>(history);
-    updateGlobalHistReg(tid, true);
+    bp_history = static_cast<void*>(history);
 }
 
 void
-BiModeBP::squash(ThreadID tid, void *bpHistory)
+BiModeBP::updateHistories(ThreadID tid, Addr pc, bool uncond,
+                         bool taken, Addr target, void * &bp_history)
 {
-    BPHistory *history = static_cast<BPHistory*>(bpHistory);
+    assert(uncond || bp_history);
+    if (uncond) {
+        uncondBranch(tid, pc, bp_history);
+    }
+    updateGlobalHistReg(tid, taken);
+}
+
+
+void
+BiModeBP::squash(ThreadID tid, void * &bp_history)
+{
+    BPHistory *history = static_cast<BPHistory*>(bp_history);
     globalHistoryReg[tid] = history->globalHistoryReg;
 
     delete history;
+    bp_history = nullptr;
 }
 
 /*
@@ -104,7 +128,7 @@ BiModeBP::squash(ThreadID tid, void *bpHistory)
  * direction predictors for the final branch prediction.
  */
 bool
-BiModeBP::lookup(ThreadID tid, Addr branchAddr, void * &bpHistory)
+BiModeBP::lookup(ThreadID tid, Addr branchAddr, void * &bp_history)
 {
     unsigned choiceHistoryIdx = ((branchAddr >> instShiftAmt)
                                 & choiceHistoryMask);
@@ -136,17 +160,11 @@ BiModeBP::lookup(ThreadID tid, Addr branchAddr, void * &bpHistory)
     }
 
     history->finalPred = finalPrediction;
-    bpHistory = static_cast<void*>(history);
-    updateGlobalHistReg(tid, finalPrediction);
+    bp_history = static_cast<void*>(history);
 
     return finalPrediction;
 }
 
-void
-BiModeBP::btbUpdate(ThreadID tid, Addr branchAddr, void * &bpHistory)
-{
-    globalHistoryReg[tid] &= (historyRegisterMask & ~1ULL);
-}
 
 /* Only the selected direction predictor will be updated with the final
  * outcome; the status of the unselected one will not be altered. The choice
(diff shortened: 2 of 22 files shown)


src/cpu/pred/bi_mode.cc
```diff
@@ -221,7 +239,7 @@ BiModeBP::update(ThreadID tid, Addr branchAddr, bool taken, void *bpHistory,
         }
     }
 
-    delete history;
+    delete history; bp_history = nullptr;
```
andysan: Separate line for the nulllptr.

src/cpu/pred/tage_base.cc
```diff
@@ -1,4 +1,16 @@
 /*
+ * Copyright (c) 2022-2023 The University of Edinburgh
```
andysan: This file doesn't have any other changes, so don't update the copyright header here.

src/cpu/pred/tage_base.hh
```diff
@@ -1,4 +1,16 @@
 /*
+ * Copyright (c) 2022-2023 The University of Edinburgh
```
andysan: This file doesn't have any other changes, so don't update the copyright header here.

src/cpu/pred/tage_sc_l.cc
```diff
@@ -444,25 +456,25 @@ TAGE_SC_L::update(ThreadID tid, Addr branch_pc, bool taken, void *bp_history,
...
-                                              bi->scBranchInfo, corrTarget);
+                                              bi->scBranchInfo, target);
 
         tage->updateHistories(tid, branch_pc, taken, bi->tageBranchInfo, false,
-                              inst, corrTarget);
+                              inst, target);
     }
 
-    delete bi;
+    delete bi; bp_history = nullptr;
```
andysan: New line.

src/cpu/pred/tournament.cc
```diff
@@ -346,7 +315,7 @@ TournamentBP::squash(ThreadID tid, void *bp_history)
     }
 
     // Delete this BPHistory now that we're done with it.
-    delete history;
+    delete history; bp_history = nullptr;
```
andysan: New line

src/cpu/pred/tournament.cc
```diff
@@ -329,11 +298,11 @@ TournamentBP::update(ThreadID tid, Addr branch_addr, bool taken,
     }
 
     // We're done with this history, now delete it.
-    delete history;
+    delete history; bp_history = nullptr;
```
andysan: New line.

src/cpu/pred/bpred_unit.cc
```diff
@@ -123,246 +94,380 @@ BPredUnit::drainSanityCheck() const
...
+}
+
+
+
+
+bool
+BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
+                   PCStateBase &pc, ThreadID tid, PredictorHistory* &hist)
+{
+    assert(hist==nullptr);
```
andysan: There should be white spaces on either side of the operator here.

src/cpu/pred/bpred_unit.cc
```diff
@@ -123,246 +94,380 @@ BPredUnit::drainSanityCheck() const
...
+    }
+
+    DPRINTF(Branch, "[tid:%i, sn:%llu] PC:%#x BTB:%s\n",
+            tid, seqNum, hist->pc,  (hist->btbHit) ? "hit" : "miss");
 
+
+    // If we model a decoupled front-end the BP requires a BTB hit
+    // otherwise it cannot detect a branch.
+    const bool allow_btb_override =
+                    (hist->btbHit || !requiresBTBHit) ? true : false;
```
andysan: No need to use a trinary operator here, the condition is already boolean.

src/cpu/pred/bpred_unit.cc
```diff
@@ -123,246 +94,380 @@ BPredUnit::drainSanityCheck() const
...
+     * For indirect branches/calls check the indirect
+     * predictor if one is available. Not for returns.
+     * Note that depending on the implementation a
+     * indirect predictor might only return a target
+     * for an indirect branch with a changing target.
+     * As most indirect branches have a static target
+     * using the target from the BTB is the optimal
+     * to save space in the indirect preditor itself.
+     */
     if (iPred) {
```
andysan: I'd suggest merging the condition on this line and the next.

andysan (changes requested):
I'll need some more time to understand what's happening in brped_unit.cc. In the meantime, here are some minor style nits.

dhschall:
Sure! I think this needs time.Maybe @ rjc-arch can also have a look if he spots issues.
Also I tried to explain more what's going on in the code and why things are happen. If something is not obvious please let me know so I can improve the comments.

src/cpu/pred/bi_mode.cc
```diff
@@ -73,25 +85,37 @@ BiModeBP::BiModeBP(const BiModeBPParams &params)
...
+
+
+void
+BiModeBP::squash(ThreadID tid, void * &bp_history)
+{
+    BPHistory *history = static_cast<BPHistory*>(bp_history);
     globalHistoryReg[tid] = history->globalHistoryReg;
 
     delete history;
+    bp_history = nullptr;
```
dhschall, line 118: The reason I added in all predictors the explicit nulling of the `bp_history` is I added a sanity check in the deconstructor of the `PredictorHistory`.  It fails when the history is not null and is intended to prevent memory leaks

src/cpu/pred/BranchPredictor.py
```diff
@@ -128,6 +137,11 @@ class BranchPredictor(SimObject):
 
     numThreads = Param.Unsigned(Parent.numThreads, "Number of threads")
     instShiftAmt = Param.Unsigned(2, "Number of bits to shift instructions by")
+    requiresBTBHit = Param.Bool(
+        False,
```
dhschall: TODO: Improve the description of this new feature.

Its actually not a feature but a more realistic modelling of the BPU behavior for high-end servers.

src/cpu/pred/bpred_unit.cc
```diff
@@ -419,105 +534,206 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
...
+      ADD_STAT(earlyResteers, statistics::units::Count::get(),
+              "Number of branches that got squashed after decode."),
+
+      ADD_STAT(condPredicted, statistics::units::Count::get(),
+               "Number of conditional branches predicted"),
+      ADD_STAT(condPredictedTaken, statistics::units::Count::get(),
+               "Number of conditional branches predicted as taken"),
+      ADD_STAT(condIncorrect, statistics::units::Count::get(),
+               "Number of conditional branches incorrect"),
+      ADD_STAT(incorrect, statistics::units::Count::get(),
```
dhschall: TODO: cleanup stats.
incorrect is the same as condIncorrect.
Also we should maybe remove some stats.

Should we disable the `total` flag to prevent printing `0` values

src/cpu/pred/bi_mode.cc
andysan: I figured it was something along those lines. That makes sense.

andysan:
I just talked to @rjc-arch about the SPEC2k17 stats and how they are affected by `requiresBTBHit`. The current version of the patch has a behaviour that is very similar, but not identical, to the old version. I _think_ this might be caused by subtle changes in the order that branch predictors are consulted / updated, but that's just a theory.

When enabling the flag, there is a drop in IPC for a few SimPoints. I think this is expected as there will be some cases where there is a BTB miss but there could've been a hit in other predictors. The old version of the branch predictor is technically cheating because it relies on the decoded instruction to make decision. Requiring a BTB hit emulates a behaviour where such information is cached in the BTB which is a reasonable way to do it in hardware.

I would suggest that we remove the requiresBTBHit param from this change and introduce it in a separate PR. I think there are three reasons for this:

- It introduces a new behaviour, so it's not really appropriate for a refactor patch.
- It should be on by default since it more accurately reflects how hardware works, which requires wider awareness and more careful review.
- We might need to have an independent control for the indirect predictor. It seems plausible that a hit in the indirect predictor could supply a new target even if there is a BTB miss.

dhschall:
> I would suggest that we remove the requiresBTBHit param from this change and introduce it in a separate PR. I think there are three reasons for this:

That's a good idea. I'll do that.
I just want to add a comment on the indirect branches. 
The biggest difference I observed when I implemented this feature was for returns. Indirect branches are relatively rare as compared to returns.
However, the big question is how the branch predictor knows if an instruction is a return without a BTB hit?

The behavior for direct branches is a bit less cheating because gem5 still models the BTB miss.

andysan:
> That's a good idea. I'll do that. I just want to add a comment on the indirect branches. The biggest difference I observed when I implemented this feature was for returns. Indirect branches are relatively rare as compared to returns. However, the big question is how the branch predictor knows if an instruction is a return without a BTB hit?

That makes sense, I didn't consider that. I can't see how the BP would know if something is a return unless you have the instruction word (which you wouldn't have at this stage) or you have a BTB hit.

dhschall:
@andysan I moved the `requireBTBhit` behaviour into #493.
This one should now be mostly refactoring and should not change performance

dhschall:
I have applied all suggestions from @rjc-arch  in #499, which belongs to the BPU refactoring.
@rjc-arch is this ok for you and do you saw any performance differences?

rjc-arch (approved):
Hi David. Thanks for making the changes. 
Based on a diff against the previous version, I'm happy you've addressed all the issues I mentioned in my review [here](https://github.com/gem5/gem5/pull/493#pullrequestreview-1689665772). Looks good to me.

andysan (approved):

Issue #456, opened 2023-10-14
There are no GPUFS tests


abmerop:
There are no GPUFS tests despite being usable for several years.  All of the old resources are available so we could do something like the code below.  The last release is broken on stable and apparently fixing on stable is impossible (see #295) so testing is necessary. 

Something like this should work and takes 1-2 minutes with KVM:

```  gpufs-square-test:
    runs-on: [self-hosted, linux, x64, build]
    container: # whatever the default is ?
    needs: build-gem5  # don't know ?
    timeout-minutes: 20
    steps:
      # Is the VEGA_X86 build already available ??
      # Need gem5-resources as well for vega mmio trace file
      - name: Get disk image
        working-directory: ${{ github.workspace }}  # Is this already available somewhere?
        run:  build/VEGA_X86/gem5.opt util/obtain-resource.py x86-gpu-fs-img -p x86-gpu-fs-img
      - name: Get kernel
        working-directory: ${{ github.workspace }}  # Some cache folder??
        run:  build/VEGA_X86/gem5.opt util/obtain-resource.py x86-linux-kernel-5.4.0-105-generic -p x86-linux-kernel-5.4.0-105-generic
      - name: Get square
        working-directory: ${{ github.workspace }}  # Some cache folder??
        run:  build/VEGA_X86/gem5.opt util/obtain-resource.py x86-gpu-square -p x86-gpu-square
      - name: Run GPUFS square test
        working-directory: ${{ github.workspace }}
        run: build/VEGA_X86/gem5.opt configs/example/gpufs/vega10_kvm.py --disk-image ./x86-gpu-fs-img --kernel ./x86-linux-kernel-5.4.0-105-generic --app ./x86-gpu-square --gpu-mmio-trace ./gem5-resources/src/gpu-fs/vega_mmio.log
      # Ideally would check for "PASSED!" in output too```

mattsinc:
@v-ramadas : can you please check if this works with your stuff?

v-ramadas:
Running the square command gives me the following error : 
`OSError: Can't find file 'disks/x86-gpu-fs-img' on M5_PATH`.
If I specify the disk image path as `--disk-image ../gem5/x86-gpu-fs-img` instead of `--disk-image x86-gpu-fs-img`, it moves past this.

[...]

PR #457, opened 2023-10-14
configs: Fix missing param exchange for GPUFS

abmerop:
PR #367 adds an option to configs/ruby/GPU_VIPER.py that was not added to the corresponding dGPU equal for GPUFS and thus all GPUFS runs are failing. Fixed in this patch.

Change-Id: I773ff567c94f8e8515e343611fa1d7b66a7187e6


abmerop:
Perhaps coincidence, but this caught the *exact* problems I want to address in #456. 

Caught this with my own personal tester 1 hour back:

image

BobbyRBruce (approved):

mattsinc (approved):
good catch!

PR #462, opened 2023-10-16
misc: Add missing RISCV valid ISA option to README.md

rogerchang23424:
The list of valid ISA options should be same as the website:
https://www.gem5.org/documentation/general_docs/building

Change-Id: Id5ace5b0356ec35634caec5b11159551801c0615


shingarov (changes requested):
I would use a more informative commit message.  For example: "misc: Add RISCV to the list of valid ISA options".

As an aside: when I look at an area of code, it more often than not results in discovering other problems, so I usually try to fix them while I am at it.  In this case of line 41, there are two problems: the missing RISCV, and there is no period at the end.  I don't know whether you meant to fix the period in a separate PR?

shingarov (approved):

PR #464, opened 2023-10-16
arch-riscv: Dynamically add V extension to device tree

hnpl:
Currently, we are hardcoding the ISA string in the device tree
generator. The ISA string from the device tree affects which
ISA extensions will be used by the bootloader/kernel.

This function allows generating the ISA string from the gem5's
ISA object rather than using hardcoded values.

This series of changes also correct a couple of hardcoded
RISC-V ISA strings in the standard library, as well as not
enable RVV instructions for the U74 core model.

Signed-off-by: Hoa Nguyen <hn@hnpl.org>

### DIFF
--- a/src/arch/riscv/RiscvISA.py
+++ b/src/arch/riscv/RiscvISA.py
@@ -13,6 +13,7 @@
 #
 # Copyright (c) 2016 RISC-V Foundation
 # Copyright (c) 2016 The University of Virginia
+# Copyright (c) 2023 The Regents of the University of California
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
@@ -94,3 +95,18 @@ class RiscvISA(BaseISA):
         "Length of each vector element in bits. \
         ELEN in Ch. 2 of RISC-V vector spec",
     )
+
+    def get_isa_string(self):
+        isa_extensions = []
+        # check for the base ISA type
+        if self.riscv_type.value == "RV32":
+            isa_extensions.append("rv32")
+        elif self.riscv_type.value == "RV64":
+            isa_extensions.append("rv64")
+        # use imafdc by default
+        isa_extensions.extend(["i", "m", "a", "f", "d", "c"])
+        # check for the vector extension
+        if self.enable_rvv.value == True:
+            isa_extensions.append("v")
+        isa_string = "".join(isa_extensions)
+        return isa_string
--- a/src/dev/riscv/HiFive.py
+++ b/src/dev/riscv/HiFive.py
@@ -251,7 +251,7 @@ class HiFive(HiFiveBase):
     def annotateCpuDeviceNode(self, cpu, state):
         cpu.append(FdtPropertyStrings("mmu-type", "riscv,sv48"))
         cpu.append(FdtPropertyStrings("status", "okay"))
-        cpu.append(FdtPropertyStrings("riscv,isa", "rv64imafdcsu"))
+        cpu.append(FdtPropertyStrings("riscv,isa", "rv64imafdc"))
         cpu.appendCompatible(["riscv"])
 
         int_node = FdtNode("interrupt-controller")
--- a/src/python/gem5/components/boards/experimental/lupv_board.py
+++ b/src/python/gem5/components/boards/experimental/lupv_board.py
@@ -316,7 +316,7 @@ class LupvBoard(AbstractSystemBoard, KernelDiskWorkload):
             node.append(FdtPropertyWords("reg", state.CPUAddrCells(i)))
             node.append(FdtPropertyStrings("mmu-type", "riscv,sv48"))
             node.append(FdtPropertyStrings("status", "okay"))
-            node.append(FdtPropertyStrings("riscv,isa", "rv64imafdcsu"))
+            node.append(FdtPropertyStrings("riscv,isa", "rv64imafdc"))
             # TODO: Should probably get this from the core.
             freq = self.clk_domain.clock[0].frequency
             node.appendCompatible(["riscv"])
--- a/src/python/gem5/components/boards/riscv_board.py
+++ b/src/python/gem5/components/boards/riscv_board.py
@@ -280,7 +280,11 @@ class RiscvBoard(AbstractSystemBoard, KernelDiskWorkload):
             node.append(FdtPropertyWords("reg", state.CPUAddrCells(i)))
             node.append(FdtPropertyStrings("mmu-type", "riscv,sv48"))
             node.append(FdtPropertyStrings("status", "okay"))
-            node.append(FdtPropertyStrings("riscv,isa", "rv64imafdc"))
+            node.append(
+                FdtPropertyStrings(
+                    "riscv,isa", core.core.isa[0].get_isa_string()
+                )
+            )
             # TODO: Should probably get this from the core.
             freq = self.clk_domain.clock[0].frequency
             node.append(FdtPropertyWords("clock-frequency", freq))
--- a/src/python/gem5/prebuilt/riscvmatched/riscvmatched_core.py
+++ b/src/python/gem5/prebuilt/riscvmatched/riscvmatched_core.py
@@ -214,3 +214,4 @@ class U74Core(BaseCPUCore):
         core_id,
     ):
         super().__init__(core=U74CPU(cpu_id=core_id), isa=ISA.RISCV)
+        self.core.isa[0].enable_rvv = False


src/arch/riscv/RiscvISA.py
```diff
@@ -58,3 +59,18 @@ class RiscvISA(BaseISA):
...
+            isa_extensions.append("rv32")
+        elif self.riscv_type.value == "RV64":
+            isa_extensions.append("rv64")
+        # use imafdc by default
+        isa_extensions.extend(["i", "m", "a", "f", "d", "c"])
+        # check for the vector extension
+        if self.enable_rvv.value == True:
+            isa_extensions.append("v")
+        isa_string = "".join(isa_extensions)
+        return isa_string
```
BobbyRBruce, line 112: This code is strange to me... I have a few suggestions (though I believe your code is functionally correct):


```suggestion
    def get_isa_string(self) ->:
        # 32 or 64 bit.
        to_return = "rv32" if self.riscv_type.value == "RV32" else "rv64"
        assert to_return = "rv32 || self.riscv_type.value = "RV64"
        
        # use imafdc by default.
        to_return += "imafdc"
        
        # check for the vector extension
        if self.enable_RVV.value:
            to_return += "v"
        
        return to_return
```

High level:

1. Vector of strings to concatenate strings and characters 🤢.
2. If the case that neither "rv32" and "rv64" are specified is possible, i'd like an error thrown. I think that must be the case here?
3. Nit: `(self.enable_rvv.value == True) == (self.enable_rvv.value)`

I'm guessing you did this `isa_extensions` vector with some grander idea of having the isa be generated automatically from whatever extensions are there, but for now I think i'd prefer the code to be cleaner.

hnpl: Agreed.

Nit

> to_return += "imafdc"

Since each letter represents an extension, I'd like to keep them separately and concat them latter on. I believe this is cleaner than having a hardcoded string.

powerjg: Here's a different suggestion. 

Do we also need to include the Z extension here?

```suggestion
    def get_isa_string(self):
        isa_extensions = []
        # check for the base ISA type
        if self.riscv_type.value == "RV32":
            isa_extensions.append("rv32")
        elif self.riscv_type.value == "RV64":
            isa_extensions.append("rv64")
        # use imafdc by default
        isa_extensions.extend(
            ["i", # integer instructions
            "m", # multiply/divide instructions
            "a", # atomic instructions
            "f", # floating point 
            "d", # double precision floating point
            "c", # compressed instructions
            ]
        )
        # check for the vector extension
        if self.enable_rvv.value == True:
            isa_extensions.append("v") # vector instructions
        isa_string = "".join(isa_extensions)
        return isa_string
```

hnpl: We should include the Z extensions we are supporting. However, I don't think Z extensions that gem5 has need kernel support to be executed correctly.

Also, the base extensions (IMAFDC) are pretty unique in this sense as we need to tell the kernel explicitly.

Having said that, do we have a list of Z extensions that were implemented?

hnpl: Here are the Z exts recognized by the kernel

https://elixir.bootlin.com/linux/v6.5.7/source/arch/riscv/kernel/cpu.c#L213

powerjg: @rogerchang23424 may know the Z extensions we support.

rogerchang23424: gem5 support following Z extensions:
CMO: Zicbom(develop), Zicboz(develop)
Zi*: Zicntr, Zicsr, Zifencei, Zihpm
Zc*: Zca, Zcf, Zcd(includedi in the c exts), Zcb(develop),
Zf*: Zfh
Zb*: Zba, Zbb, Zbc, Zbkb, Zbkc, Zbkx, Zbs
Zk*: Zknd, Zkne, Zknh, Zksed, Zksh

The kernel can recognize the Z exts in v6.5.7 includes:
Zicbom, Zicboz, Zicntr, Zicsr, Zifencei, Zihpm, Zba, Zbb and Zbs

Thanks

powerjg:
@BobbyRBruce, we need your input on whether the original commit, my proposed change, or something else is required before you will approve.

@hnpl, please add the z extensions as well, assuming it makes sense to :)

BobbyRBruce:
> @BobbyRBruce, we need your input on whether the original commit, my proposed change, or something else is required before you will approve.
> 
> @hnpl, please add the z extensions as well, assuming it makes sense to :)

Maybe there's a way to make our design here a bit crisper and more easy to extend. If we're to add more optional extensions going forward (this seems inevitable), maybe we want a better way to encapsulate this information via the SImObjects.

I was playing around with it and came up with this as a draft idea:

```py
from m5.params import Enum
from m5.params import Param
from m5.objects.BaseISA import BaseISA

from typing import List


class RiscvType(Enum):
    vals = ["RV32", "RV64"]


class RiscvExtension:
    def __init__(self, id: str, name: str):
        # Honestly not sure what the word for this is, but I mean the letter
        # that represents the extension (???) 'i', 'm', etc.
        self._id = id 

        # The name of the extension (human readable).
        self._name = name

    def get_name(self) -> str:
        return self._name
    
    def get_id(self) -> id:
        return self._id
    
    def __eq__(self, __value: object) -> bool:
        return isinstance(__value, RiscvExtension) \
                    and __value.get_id() == self._id
    
    def __str__(self):
        return "RiscvExtension: " + self._id + " | " + self._name

    
class RiscvISA(BaseISA):
    type = "RiscvISA"
    cxx_class = "gem5::RiscvISA::ISA"
    cxx_header = "arch/riscv/isa.hh"

    check_alignment = Param.Bool(
        True, "whether to check memory access alignment"
    )
    riscv_type = Param.RiscvType("RV64", "RV32 or RV64")

    enable_rvv = Param.Bool(True, "Enable vector extension")

    # I'd prefer this to be a set. A set implies there is an order and that
    # duplications are permitted. This is the case with extensions. However
    # I supect order is necessary for creating the isa string, so we need 
    # this (???). Should each RiscvExtension object have a "order priority"?
    # of some kind? I don't know how this order is determined.
    #
    # This is the base, non-configurable extensions. They come with the gem5
    # Riscv ISA whether you like it or not.
    _base_extensions = [ 
        RiscvExtension("i", "Integer Instructions"),
        RiscvExtension("m", "Multiply/Divide Instructions"),
        RiscvExtension("a", "Atomic Instructions"),
        RiscvExtension("f", "Single-Precision Floating-Point Instructions"),
        RiscvExtension("d", "Double-Precision Floating-Point Instructions"),
        RiscvExtension("c", "Compressed Instructions"),
    ]

    def get_extensions(self) -> List[RiscvExtension]:
        extensions = [].extend(self._base_extensions)
        if self.enable_rvv.value:
            extensions.append(RiscvExtension("v", "Vector Instructions"))
        return self._base_extensions
    
    def get_isa_string(self) -> str:
        return self.riscv_type.value.lower() + "".join(
            [extension.get_id() for extension in self.get_extensions()]
        )
```

It organizes the ISA extensions in a neater way. If we need to expose an extension at the Python level, this gives us a starting point.

hnpl:
@powerjg: I added the Z extensions as follows in my test, not in this change,

```sh
rv64imafdcv_Zicbom_Zicboz_Zicntr_Zicsr_Zifencei_Zihpm_Zba_Zbb_Zbs
```

It looks like the kernel and the bootloader recognize those extensions. However, for the cache maintenance extensions Zicbom and Zicboz, we'll have to modify the device tree and test them with classic/Ruby cache, and it's ways out of the scope of making vector instructions work in full system. So I'd say let's not adding the Z extensions for now until we do more testing.

@BobbyRBruce: I like the idea of having a separate risc-v extensions class. Let's do that in another patch.

BobbyRBruce:
> @powerjg: I added the Z extensions as follows,
> 
> ```shell
> rv64imafdcv_Zicbom_Zicboz_Zicntr_Zicsr_Zifencei_Zihpm_Zba_Zbb_Zbs
> ```
> 
> It looks like the kernel and the bootloader recognize those extensions. However, for the cache maintenance extensions Zicbom and Zicboz, we'll have to modify the device tree and test them with classic/Ruby cache, and it's ways out of the scope of making vector instructions work in full system. So I'd say let's not adding the Z extensions for now until we do more testing.
> 
> @BobbyRBruce: I like the idea of having a separate risc-v extensions class. Let's do that in another patch.



> @powerjg: I added the Z extensions as follows,
> 
> ```shell
> rv64imafdcv_Zicbom_Zicboz_Zicntr_Zicsr_Zifencei_Zihpm_Zba_Zbb_Zbs
> ```
> 
> It looks like the kernel and the bootloader recognize those extensions. However, for the cache maintenance extensions Zicbom and Zicboz, we'll have to modify the device tree and test them with classic/Ruby cache, and it's ways out of the scope of making vector instructions work in full system. So I'd say let's not adding the Z extensions for now until we do more testing.
> 
> @BobbyRBruce: I like the idea of having a separate risc-v extensions class. Let's do that in another patch.

Sure. I think focusing on this too much is just bikeshedding. It's a thumbs-up from me.

BobbyRBruce (dismissed):

powerjg:
How are the z extensions added in this PR? I think I'm missing something.

hnpl:
Oops. I'll remove the z extensions.

hnpl:
I just checked the PR again. This change doesn't add Z extensions to the ISA string.

rogerchang23424 (approved):

powerjg (approved):

PR #465, opened 2023-10-16
cpu, arch-arm: Add IsPseudo tag for gem5 pseudo instructions

giactra:
This only applies to pseudo instructions with their own encoding (m5 ops)... In other
words memory mapped m5 operations are not supported. This make sense as they should
rather be treated as device accesses... Though it is something to take into consideration
when relying on the flag


BobbyRBruce (approved):
Looks fine to me. Thanks for @giactra !

giactra:
Thanks Bobby!

PR #470, opened 2023-10-17
util: Improve GitHub Action runners: Enable KVM; Better Cleanup; Better Tooling

BobbyRBruce:
This PR adds the following the GitHub Actions runners:

1. Enables KVM to be run within docker containers within the VMs, if permitted. Now, any Docker containers wanting to use KVM must create containers with the `--device /dev/kvm` argument. This may make it hard or impossible to utilize with GitHub Actions. Nonetheless it is enabled.
2. Improves the docker prune step (a cleanup step carried out after each job) so it now removes all the Docker images in the VM.
3. Adds the "halt-helper.sh" script which automatically, and safely, halts (shutsdown) all the VMs so maintenance tasks can be undertaken.


powerjg:
You've got some typos in your description :)

powerjg (approved):
lgtm

PR #471, opened 2023-10-17
arch-riscv: Fatal if specify RVV in RV32 mode

rogerchang23424:
Currently we only support RVV in RV64 mode, if users want to run RISC-V
in RV32 mode, the should disable RVV manually.

Change-Id: I1f224d2b927d576d13991aa440c9e2c2c3969f6a


rogerchang23424:
Not intent to be merged

PR #476, opened 2023-10-17
sim: Make SIGINT work in more cases

powerjg:
Currently, if you hit "ctrl-c" in some cases it doesn't actually exit the simulator.
This change makes it so that the simulator loop will exit even if it's executing
events on an event queue that is not the "main" event queue. An example of
where this happens is when restoring from checkpoint in Ruby.

Change-Id: Ia58c6745f286e2049f1a71664f339440aebd5d4d

### DIFF
--- a/src/sim/simulate.cc
+++ b/src/sim/simulate.cc
@@ -293,22 +293,30 @@ doSimLoop(EventQueue *eventq)
         assert(curTick() <= eventq->nextTick() &&
                "event scheduled in the past");
 
-        if (mainQueue && async_event) {
+        if (async_event) {
             async_event = false;
-            // Take the event queue lock in case any of the service
-            // routines want to schedule new events.
-            std::lock_guard<EventQueue> lock(*eventq);
-            if (async_statdump || async_statreset) {
-                statistics::schedStatEvent(async_statdump, async_statreset);
-                async_statdump = false;
-                async_statreset = false;
+
+            if (mainQueue) {
+                // Take the event queue lock in case any of the service
+                // routines want to schedule new events.
+                std::lock_guard<EventQueue> lock(*eventq);
+                if (async_statdump || async_statreset) {
+                    statistics::schedStatEvent(async_statdump,
+                                               async_statreset);
+                    async_statdump = false;
+                    async_statreset = false;
+                }
+
+                if (async_io) {
+                    async_io = false;
+                    pollQueue.service();
+                }
             }
 
-            if (async_io) {
-                async_io = false;
-                pollQueue.service();
-            }
-
+            // Whether or not we're executing events on the main queue, we
+            // should exit on ctrl-c or an exception. For instance, it is
+            // required to have the check separate when restoring from a
+            // checkpoint in Ruby.
             if (async_exit) {
                 async_exit = false;
                 exitSimLoop("user interrupt received");


powerjg:
This works when hitting "ctrl-c" in the main event queue. I have not tested it to see if this fixes the problem with restoring Ruby. I probably won't have a chance to test this for a while.

src/sim/simulate.cc
```diff
@@ -293,22 +293,30 @@ doSimLoop(EventQueue *eventq)
         assert(curTick() <= eventq->nextTick() &&
                "event scheduled in the past");
 
-        if (mainQueue && async_event) {
+        if (async_event) {
             async_event = false;
```
andysan, line 297: There is a risk that you will be losing IO and stat events after this patch because async_event is cleared even if there are pending IO or stat signals. This notification interface is a bit of a mess and should probably be replaced with an std::atomic_uint32_t bit field instead of separate variables.

powerjg: I'm not totally sure how to fix this as I don't quite understand the situation in which it happens. Could you provide a theoretical trace?

On the other hand... is it better or worse to potentially lose a stat signal (which, btw, I don't think is a commonly used interface) or where ctrl-c simply doesn't do anything for seconds to minutes at a time?

I'm not going to have time to continue working on this beyond small updates. Our choices are (1) keep the current broken behavior or (2) go with something like this change. The best option, someone else picking this up and converting to some thread-safe implementation, seems quite unlikely in the short term. 

Do you think this change, on the whole, improves gem5, makes it worse, or makes no difference? If it's the latter two, then I'll drop the change. If it's the former, we should commit this and make a bigger change later.

src/sim/simulate.cc
```diff
@@ -293,22 +293,30 @@ doSimLoop(EventQueue *eventq)
...
-            // Take the event queue lock in case any of the service
-            // routines want to schedule new events.
-            std::lock_guard<EventQueue> lock(*eventq);
```
andysan, line 300: You're actually changing the locking behaviour of the `async_exit` case here. It's probably benign though.

src/sim/simulate.cc
andysan: I'm not so concerned about the missing a stat signal, I'm more concerned about SIGIO. This change introduces a large the window where things can go wrong. Focusing on the SIGIO case, this chain of events would cause SIGIO to get lost:

1. Start processing non-main queue
2. SIGIO handler sets `async_event` and `async_io`.
3. `doSimLoop` clears `async_event` but doesn't service the SIGIO since it isn't in the main event loop.

After this sequence of events, `async_event` is false and `async_io` is true. Because of this, the SIGIO won't be processed until another asynchronous event happens.

Looking at the change again, I think it is also fundamentally broken when performing a multi-threaded simulation. The problem there is that each thread will end up posting its own exit event.

While I think it makes sense to think about this problem and maybe fix it in the future, I think this change is causing more trouble than it is worth. A user can still quit gem5 while it is performing initialisation (e.g., not just when using a private event queue) by sending a SIGQUIT. You can do this from the terminal with ctrl-\.

andysan: I _think_ the reason for the mainQueue check is to ensure that just one queue handles asynchronous signals. I don't think it is strictly necessary to handle them from queue 0.

If this is the case, we could replace it with an atomic which would let us read and clear the pending events in one atomic operation. Combine that with a global async event lock or appropriate locking in the individual handlers and you could (probably) remove the mainQueue check altogether.

I think something along these lines might work:
```
enum EventType :: uint32_t {
    IO = 1,
    ...
};
std::atomic<std::uint32_t]> asyncEvents;
...


void
raiseAsyncEvent(EventType ev)
{
    asyncEvents.fetch_or(std::static_cast<uint32_t>(ev), std::memory_order_release);
    getEventQueue(0)->wakeup();
}

void
doSimloop()
{
    ...
    // Atomically get and clear pending async events
    uint32_t pending_flags = asyncEvents.fetch_and(0, std::memory_order_acquire);

    if (pending) {
        std::lock_guard<EventQueue> lock(*eventq);
        if (pending & EventTypes::IO) {
           ...
        }
    }
   ...
}

```

powerjg:
Unfortunately, I don't have the bandwidth to try to understand/incorporate these changes. I'll close this PR and someone else can create a new PR with these suggestions.

PR #478, opened 2023-10-18
[DO NOT MERGE][Test-only] test kconfig daily tests

rogerchang23424:


rogerchang23424:
Since the PR changes the methods to specify config options, I have created the other PR #478 to ensure the daily tests can work well after kconfig changed.

rogerchang23424:
The PR is just test if the daily test will break after kconfig changes. Close the PR we don't need to run test anymore.

PR #479, opened 2023-10-18
arch-riscv: Copy Misc Regs when swiching cpus

### LINKED ISSUE #451: The core gets stall after switching the CPU on full system under the riscv architecture
**Describe the bug**
I use SimpleSwitchableProcessor for simulation.：
processor = SimpleSwitchableProcessor(
    starting_core_type=CPUTypes.ATOMIC,
    switch_core_type=CPUTypes.TIMING,
    isa=ISA.RISCV,
    num_cores=2,
)

After the OS starts up, I  log in and execute m5 exit, which triggers the CPU to switch to TIMING cpu, but the following information is printed at this time：
675404961000: board.processor.switch.core.fetch: Running stage
675404961000: board.processor.switch.core.fetch: Attemupting to fetch from [tid:0]
675404961000: board.processor.switch.core.fetch: [tid:0] Attempting to translate and read instruction, starting at PC(0x17322=>0x17326).(0=>1)
675404961000: board.processor.switch.core.fetch: [tid:0] Fetching cache line 0x17300 for addr 0x17320ttempting to traanslate
src/cpu/03/fetch.cc:604: warn: Address 0x17300 is outside of physical memory, stopping fetch
675404961000: board.processor.switch.core: Schedulingnext tick!
From the log, we can see that after switching the CPU, we have been fetching instructions from address 0x17300, but this address is is outside of physical memory, resulting in no instructions being fetched, and the CPU stalled. 
The analysis found that the status of the CSR register was wrong after switching the CPU, causing the virtual address 0x17300 to not be converted by TLB.

**Affects version**
v23.0.1.0

**gem5 Modifications**
Modify the src/arch/riscv/isa.cc to add a copy of csr to fix this bug：

void
ISA::copyRegsFrom(ThreadContext *src)
{
          // First loop through the integer registers.
         for (auto &id: intRegClass)
                tc->setReg(id, src->getReg(id));
  
          // Second loop through the float registers.
         for (auto &id: floatRegClass)
               tc->setReg(id, src->getReg(id));
  
          //copy PC/NPC
          tc->pcState(src->pcState());
  
          //Copy csr
          for(int id = 0; id < NUM_MISCREGS; id ++) {
             tc->setMiscRegNoEffect(id, src->readMiscReg(id))
          }
}

**To Reproduce**
1 .Change SimpleProcessor in  configs/example/gem5_library/riscv-fs.py  to SimpleSwitchableProcessor，and make some other modifications：riscv-disk-img needs to be made by yourself, including the m5 package
2. Compile gem5 with command scons ./build/RISCV/gem5.opt -j 32
3. Execute the simulation with ./build/RISCV/gem5.opt configs/example/gem5_library/riscv-fs.py
4. After logging in to the OS, execute command： m5 exit

**Host Operating System**
Ubuntu 22.04,

**Host ISA**
X86

hnpl:
Misc Regs might contain rather important information about the state of a core, e.g., information in CSR registers.

This patch enforces copying the CSR registers when switching cpus. The bug and the proposed fix are reported here [1].

[1] https://github.com/gem5/gem5/issues/451

Change-Id: I611782e6e3bcd5530ddac346342a9e0e44b0f757

### DIFF
--- a/src/arch/riscv/isa.cc
+++ b/src/arch/riscv/isa.cc
@@ -294,6 +294,10 @@ ISA::copyRegsFrom(ThreadContext *src)
         tc->setReg(id, &vc);
     }
 
+    // Copying Misc Regs
+    for (int i = 0; i < NUM_MISCREGS; i++)
+        tc->setMiscRegNoEffect(i, src->readMiscRegNoEffect(i));
+
     // Lastly copy PC/NPC
     tc->pcState(src->pcState());
 }


hnpl:
The fix is proposed by [qqli579](https://github.com/qqli579).

Tested by switching the CPU type right after booting the Linux kernel in two scenarios,
- Case 1: switching from ATOMIC to TIMING.
- Case 2: switching from ATOMIC to O3.

Without this change, both cases will result in invalid address translation. The TIMING cpu panicked, while the O3 cpu hung.

With this change, both cases returns correct result. However, the O3 cpu will warn about fetching from an invalid address.

qqli579:
it's ok

BobbyRBruce (approved):

PR #486, opened 2023-10-19
arch-arm: Fix KVM Failed to set register (0x603000000013808c)

giactra:
Some debug registers were incorrectly tagged
(e.g. as being writeable). This was causing a bug in some gem5-KVM runs where gem5 was trying to initialize the state of those registers (OSLSR_EL1) [1] but KVM was returning an error (as the registers were RO).

    src/arch/arm/kvm/armv8_cpu.cc#L408

Change-Id: Iaa8d94e0e4b6fa34c39b30f0f12fcd7fe8a8da4d

### DIFF
--- a/src/arch/arm/regs/misc.cc
+++ b/src/arch/arm/regs/misc.cc
@@ -3916,18 +3916,18 @@ ISA::initializeMiscRegMetadata()
     InitReg(MISCREG_MDCCINT_EL1)
       .fault(EL1, faultMdccsrEL1)
       .fault(EL2, faultMdccsrEL2)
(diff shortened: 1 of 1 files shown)


powerjg (approved):

giactra:
Mmmm.. It seems like there is some checkpoint restore riscv test failing

powerjg:
We updated that checkpoint this morning. You need to rebase to include this change so the tests pass: https://github.com/gem5/gem5/commit/73c48a482831ac87b176751dd713d02d378ea7be

PR #493, opened 2023-10-20
cpu: Require BTB hit to detect branches.

dhschall:
In a high performance CPU there is no other way than a BTB hit
to know about a branch instruction and its type. For low-end CPU's
pre-decoding might sit in from of the BPU to provide this information.
Currently, the BPU models only low-end behavior and updates the
RAS and the indirect branch prediction even without a BTB hit.
This patch adds three things to model the correct behavior for high-end
CPUs.
1. A check before the RAS and indirect predictor wheather there was
a BTB hit or not. Only for BTB hits the BPU will consolidate RAS, and
indirect predictor.
2. Since, this check requires a BTB hit for indirect branches they must
also be installed into the BTB. For returns this was already done.
3. Finally, the BTB update previously happened at squash (decode
or commit). Since this can be out-of-order that means branches from
the false path can get installed without ever been retired.

### DIFF
--- a/src/cpu/pred/BranchPredictor.py
+++ b/src/cpu/pred/BranchPredictor.py
@@ -137,6 +137,15 @@ class BranchPredictor(SimObject):
 
     numThreads = Param.Unsigned(Parent.numThreads, "Number of threads")
     instShiftAmt = Param.Unsigned(2, "Number of bits to shift instructions by")
+    requiresBTBHit = Param.Bool(
+        False,
+        "Requires the BTB to hit for returns and indirect branches. For an"
+        "advanced front-end there is no other way than a BTB hit to know "
+        "that the branch exists in the first place. Furthermore, the BPU "
+        "needs to know the branch type to make the correct RAS operations. "
+        "This info is only available from the BTB. "
+        "Low-end CPUs predecoding might be used to identify branches. ",
+    )
 
     btb = Param.BranchTargetBuffer(SimpleBTB(), "Branch target buffer (BTB)")
     ras = Param.ReturnAddrStack(
--- a/src/cpu/pred/bpred_unit.cc
+++ b/src/cpu/pred/bpred_unit.cc
@@ -58,6 +58,7 @@ namespace branch_prediction
 BPredUnit::BPredUnit(const Params &params)
     : SimObject(params),
       numThreads(params.numThreads),
+      requiresBTBHit(params.requiresBTBHit),
       instShiftAmt(params.instShiftAmt),
       predHist(numThreads),
       btb(params.btb),
@@ -169,7 +170,11 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
     /* -----------------------------------------------
      * Branch Target Buffer (BTB)
      * -----------------------------------------------
-     * The BTB will be checked for all branches.
+     * First check for a BTB hit. This will be done
+     * regardless of whether the RAS or the indirect
+     * predictor provide the final target. That is
+     * necessary as modern front-end does not have a
+     * chance to detect a branch without a BTB hit.
      */
     stats.BTBLookups++;
     const PCStateBase * btb_target = btb->lookup(tid, pc.instAddr(), brType);
@@ -187,6 +192,13 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
             tid, seqNum, hist->pc,  (hist->btbHit) ? "hit" : "miss");
 
 
+    // In a high performance CPU there is no other way than a BTB hit
+    // to know about a branch instruction. In that case consolidate
+    // indirect and RAS predictor only if there was a BTB it.
+    // For low end CPUs predecoding might be used to identify branches.
+    const bool branch_detected = (hist->btbHit || !requiresBTBHit);
+
+
     /* -----------------------------------------------
      * Return Address Stack (RAS)
      * -----------------------------------------------
@@ -196,7 +208,7 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
      * Return: pop the the return address from the
      *    top of the RAS.
      */
-    if (ras) {
+    if (ras && branch_detected) {
         if (inst->isCall()) {
             // In case of a call build the return address and
             // push it to the RAS.
@@ -239,7 +251,7 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
      * using the target from the BTB is the optimal
      * to save space in the indirect predictor itself.
      */
-    if (iPred && hist->predTaken &&
+    if (iPred && hist->predTaken && branch_detected &&
         inst->isIndirectCtrl() && !inst->isReturn()) {
 
         ++stats.indirectLookups;
@@ -362,7 +374,7 @@ BPredUnit::commitBranch(ThreadID tid, PredictorHistory* &hist)
                 hist->inst,
                 hist->target->instAddr());
 
-    // Commite also Indirect predictor and RAS
+    // Commit also Indirect predictor and RAS
     if (iPred) {
         iPred->commit(tid, hist->seqNum,
                            hist->indirectHistory);
@@ -561,27 +573,29 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
         }
 
         // Correct BTB ---------------------------------------------------
-        // Check if the misprediction happened was because of a BTB miss
-        // or incorrect indirect predictor
-        if (actually_taken) {
-            if (hist->inst->isIndirectCtrl() && !hist->inst->isReturn()) {
-                ++stats.indirectMispredicted;
-            } else {
+        // Update the BTB for all mispredicted taken branches.
+        // Always if `requiresBTBHit` is true otherwise only if the
+        // branch was direct or no indirect predictor is available.
+        if (actually_taken &&
+            (requiresBTBHit || hist->inst->isDirectCtrl() ||
+            (!iPred && !hist->inst->isReturn()))) {
 
-                ++stats.BTBUpdates;
-                btb->update(tid, hist->pc, corr_target,
-                            getBranchType(hist->inst));
-
+            if (!hist->btbHit) {
                 ++stats.BTBMispredicted;
                 if (hist->condPred)
                     ++stats.predTakenBTBMiss;
-
-                btb->incorrectTarget(hist->pc, hist->type);
-
-                DPRINTF(Branch,"[tid:%i] [squash sn:%llu] "
-                    "BTB miss PC %#x %s \n", tid, squashed_sn,
-                    hist->pc, toString(hist->type));
             }
+
+            DPRINTF(Branch,"[tid:%i] BTB Update called for [sn:%llu] "
+                        "PC %#x -> T: %#x\n", tid,
+                        hist->seqNum, hist->pc, hist->target->instAddr());
+
+            stats.BTBUpdates++;
+            btb->update(tid, hist->pc,
+                            *hist->target,
+                            hist->type,
+                            hist->inst);
+            btb->incorrectTarget(hist->pc, hist->type);
         }
 
     } else {
--- a/src/cpu/pred/bpred_unit.hh
+++ b/src/cpu/pred/bpred_unit.hh
@@ -368,6 +368,14 @@ class BPredUnit : public SimObject
     /** Number of the threads for which the branch history is maintained. */
     const unsigned numThreads;
 
+    /** Requires the BTB to hit for returns and indirect branches. For an
+     * advanced front-end there is no other way than a BTB hit to know
+     * that the branch exists in the first place. Furthermore, the BPU needs
+     * to know the branch type to make the correct RAS operations.
+     * This info is only available from the BTB.
+     * Low-end CPUs predecoding might be used to identify branches. */
+    const bool requiresBTBHit;
+
     /** Number of bits to shift instructions by for predictor addresses. */
     const unsigned instShiftAmt;
 


dhschall:
This patch was moved out of the BPU refactoring #455 to get more opinions on what should be the correct behavior.

Some questions:
- [ ] Is it correct to say a BTB hit is required to update the RAS
- [ ] Is it also correct to do this for indirect branches? I would argue to that most of the indirect branches have static targets. For the BPU such behaviour is no different to direct branches. To save space in the indirect predictor I would say use the indirect predictor only for branches with changing targets. For example, in Seznec's [ITTAGE](https://jilp.org/jwac-2/program/cbp3_07_seznec.pdf) this is also the case. The BTB is used as base predictor. Samsung in doing the same in their [Exynos](https://ieeexplore.ieee.org/document/9138988) CPU.
- [ ] When do we install branches into the BTB. Once a BTB miss is detected with the risk that its a branch from the false path and we trash useful stuff or at commit?

andysan:
@BobbyRBruce I think we need a wider discussion about this change. We know that requiring a BTB hit will change performance for SPEC2k17 (on average, there is a small IPC decrease). I think that is expected.

Looking at the code, we are currently "cheating" in a few places where we assume that instructions have been decoded when they are passed to the BPU. This is clearly wrong. Requiring a BTB hit makes this more realistic since the BTB stores branch metadata, so you could argue that all of the information we get from the static instruction would have been cached in the BTB in that case.

In summary, I think requiring a BTB hit is making the model more realistic, but we might need to thin about where we require a BTB hit. This will need to simmer a bit before we merge it since it introduces a non-trivial performance change. We need to make sure that all relevant stakeholders have a chance to provide input before we merge it.

src/cpu/pred/BranchPredictor.py
```diff
@@ -137,6 +137,15 @@ class BranchPredictor(SimObject):
 
     numThreads = Param.Unsigned(Parent.numThreads, "Number of threads")
     instShiftAmt = Param.Unsigned(2, "Number of bits to shift instructions by")
+    requiresBTBHit = Param.Bool(
+        False,
+        "Requires the BTB to hit for returns and indirect branches. For an"
+        "advances front-end there is no other way than a BTB hit to know "
```
rjc-arch: Typo: should be 'advanced'.

src/cpu/pred/bpred_unit.cc
```diff
@@ -566,29 +589,20 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
...
-        if (actually_taken) {
-            if (hist->inst->isIndirectCtrl() && !hist->inst->isReturn()) {
-                ++stats.indirectMispredicted;
-            } else {
-
-                ++stats.BTBUpdates;
-                btb->update(tid, hist->pc, corr_target,
-                            getBranchType(hist->inst));
+        // Check if the misprediction happened was because of a BTB miss.
+        if (actually_taken &&!hist->btbHit) {
```
rjc-arch: Nit: Needs an extra space here according to the coding guidelines.

src/cpu/pred/bpred_unit.cc
```diff
@@ -566,29 +589,20 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
         }
 
         // Correct BTB ---------------------------------------------------
-        // Check if the misprediction happened was because of a BTB miss
-        // or incorrect indirect predictor
-        if (actually_taken) {
-            if (hist->inst->isIndirectCtrl() && !hist->inst->isReturn()) {
```
rjc-arch, line 567: Do we need to take account of the indirect mispredictions in the case that `requiredBTBHit` is false? In that case, the behaviour shouldn't be changed? Likewise for the removed update in the else clause.

I'm not sure if this is important or not, I just wanted to flag it to make sure it's an intended difference in behaviour. If it's what you intended, please ignore this comment.

src/cpu/pred/bpred_unit.hh
```diff
@@ -118,35 +122,70 @@ class BPredUnit : public SimObject
...
      * @param actually_taken The correct branch direction.
      * @param tid The thread id.
+     * @param inst The static instruction that caused the misprediction
+     * @param from_commit Indicate whether the squash is comming from commit
+     *              or from decode. Its optional and used for statistics.
      */
     void squash(const InstSeqNum &squashed_sn,
                 const PCStateBase &corr_target,
-                bool actually_taken, ThreadID tid);
+                bool actually_taken, ThreadID tid, bool from_commit=true);
```
rjc-arch: Nit: this could go on a separate line for consistency.

src/cpu/pred/bpred_unit.hh
```diff
@@ -118,35 +122,70 @@ class BPredUnit : public SimObject
      * @param corr_target The correct branch target.
      * @param actually_taken The correct branch direction.
      * @param tid The thread id.
+     * @param inst The static instruction that caused the misprediction
```
rjc-arch: Looks like the `inst` param has been removed since you wrote the comment.

src/cpu/pred/bpred_unit.hh
```diff
@@ -118,35 +122,70 @@ class BPredUnit : public SimObject
...
 
-     /**
-     * If a branch is not taken, because the BTB address is invalid or missing,
-     * this function sets the appropriate counter in the global and local
-     * predictors to not taken.
-     * @param inst_PC The PC to look up the local predictor.
+    /**
+     * Once done with the prediction this function updates the
+     * path and global history. All branches call this function
+     * including unconditional once.
```
rjc-arch: Typo: 'once' -> 'ones'?

src/cpu/pred/bpred_unit.hh
```diff
@@ -118,35 +122,70 @@ class BPredUnit : public SimObject
...
+     * @param PC The branch's PC that will be updated.
```
rjc-arch: Nit: missing documentation for `tid`.

src/cpu/pred/bpred_unit.hh
```diff
@@ -118,35 +122,70 @@ class BPredUnit : public SimObject
...
+    /**
+     * @param bp_history Pointer to the history object.  The predictor
+     * will need to update any state and delete the object.
+     */
+    virtual void squash(ThreadID tid, void * &bp_history) = 0;
+
+
+    /**
+     * Updates the BP with taken/not taken information.
+     * @param PC The branch's PC that will be updated.
```
rjc-arch: Nit: missing documentation for `tid`.

src/cpu/pred/bpred_unit.hh
```diff
@@ -176,21 +211,21 @@ class BPredUnit : public SimObject
...
-     * @param corrTarget The resolved target of the branch (only needed
-     * for squashed branches)
-     * @todo Make this update flexible enough to handle a global predictor.
+     * Looks up a given PC in the BTB to get current static instruction
+     * information. This is necessary in a decoupled frontend as
+     * the information does not usually exist at that this point.
+     * Only for instructions (branches) that hit in the BTB this information
+     * is available as the BTB stores them together with the target.
+     *
+     * @param inst_PC The PC to look up.
```
rjc-arch: Nit: missing documentation for tid.

src/cpu/pred/bpred_unit.hh
```diff
@@ -213,88 +248,133 @@ class BPredUnit : public SimObject
...
+        void *indirectHistory = nullptr;
+
+        void *rasHistory = nullptr;
 
-        /** Whether this branch was mispredicted */
-        bool mispredict = false;
     };
 
-    typedef std::deque<PredictorHistory> History;
+    typedef std::deque<PredictorHistory*> History;
```
rjc-arch: RFC: I wonder if this should be named something like `HistoryQueue`?
 Or alternatively (perhaps better) should `PredictorHistory` be `PredictorHistoryItem` or `PredictorHistoryEntry`? 

The formenr looses some of the brevity that the typedef is trying to achieve, but at the moment it looks like a `History` is a queue of Histories.

I haven't read forward in the code yet, but I'm wondering if this could be confusing for a future reader.

src/cpu/pred/bpred_unit.hh
```diff
@@ -213,88 +248,133 @@ class BPredUnit : public SimObject
...
+    void commitBranch(ThreadID tid, PredictorHistory* &bpu_history);
+
+
+
+  protected:
     /** Number of the threads for which the branch history is maintained. */
     const unsigned numThreads;
 
+    /** Requires the BTB to hit for returns and indirect branches. For an
+     * advances front-end there is no other way than a BTB hit to know
```
rjc-arch: Typo: 'advances' -> 'advanced'?

src/cpu/pred/bpred_unit.hh
```diff
@@ -312,42 +392,47 @@ class BPredUnit : public SimObject
     /** The indirect target predictor. */
     IndirectPredictor * iPred;
 
+    /** Statistics */
     struct BPredUnitStats : public statistics::Group
     {
-        BPredUnitStats(statistics::Group *parent);
+        BPredUnitStats(statistics::Group *parent, BPredUnit *bp);
```
rjc-arch: Would the `parent` ever not be the `bp`? In `bpred_unit.cc` it's called with `(this, this)`. If the pointers will always be the same, they could be combined.

src/cpu/pred/bpred_unit.cc
```diff
@@ -123,246 +94,377 @@ BPredUnit::drainSanityCheck() const
...
+    hist->targetProvider = TargetProvider::NoTarget;
+
+    /* -----------------------------------------------
+     * Branch Target Buffer (BTB)
+     * -----------------------------------------------
+     * First check for a BTB hit. This will be done
+     * regardless of whether the RAS or the indirect
+     * predictor provide the final target.
+     */
+    stats.BTBLookups++;
```
rjc-arch, line 179: Is this stat (and the dependent ones below) meaningful if it is always incremented even when the RAS or indirect predictor provide the result?
(No change necessary if this is the way you intended.)

src/cpu/pred/bpred_unit.cc
```diff
@@ -123,246 +94,377 @@ BPredUnit::drainSanityCheck() const
...
+
+
+    if (iPred) {
+        // Update the indirect predictor with the direction prediction
+        iPred->update(tid, seqNum, hist->pc, false, hist->predTaken,
+                      *hist->target, brType, hist->indirectHistory);
+    }
 
-    return pred_taken;
+    // dump();
```
rjc-arch: Left over debug code?

src/cpu/pred/bpred_unit.cc
```diff
@@ -380,36 +482,48 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
...
     ppMisses->notify(1);
 
-    DPRINTF(Branch, "[tid:%i] Squashing from sequence number %i, "
-            "setting target to %s\n", tid, squashed_sn, corr_target);
+
+    DPRINTF(Branch, "[tid:%i] Squash from %s start from sequence number %i, "
+            "setting target to %s\n", tid, from_commit ? "commit" : "decode",
+            squashed_sn, corr_target);
+
+    // dump();
```
rjc-arch: Left over debugging code?

src/cpu/pred/bpred_unit.cc
```diff
@@ -380,36 +482,48 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
...
 
     // If there's a squash due to a syscall, there may not be an entry
     // corresponding to the squash.  In that case, don't bother trying to
     // fix up the entry.
     if (!pred_hist.empty()) {
 
-        auto hist_it = pred_hist.begin();
-        //HistoryIt hist_it = find(pred_hist.begin(), pred_hist.end(),
-        //                       squashed_sn);
+        PredictorHistory* const hist = *(pred_hist.begin());
```
rjc-arch: You could use `front()` here for a slightly cleaner syntax.

src/cpu/pred/bpred_unit.cc
```diff
@@ -419,105 +533,210 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
...
+              "Branches get corrected by decode or after execute. Also a "
+              "branch misprediction can be detected out-of-order. Therefore, "
+              "a corrected branch might not end up beeing committed in case "
+              "an even earlier branch was mispredicted"),
+      ADD_STAT(earlyResteers, statistics::units::Count::get(),
+              "Number of branches that got redirected after decode."),
+      ADD_STAT(committed, statistics::units::Count::get(),
+              "Number of branches finally committed "),
+      ADD_STAT(mispredicted, statistics::units::Count::get(),
+              "Number of committed branches that where mispredicted."),
```
rjc-arch: Typo: 'where' -> 'were'

src/cpu/pred/bpred_unit.cc
```diff
@@ -419,105 +533,210 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
...
+      ADD_STAT(targetProvider, statistics::units::Count::get(),
+              "The componend providing the target for taken branches"),
```
rjc-arch: Typo: 'componend' -> 'component'

src/cpu/pred/bpred_unit.cc
```diff
@@ -419,105 +533,210 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
...
+              "Number of branches where the target was incorrect or not "
+              "available at prediction time."),
+      ADD_STAT(condPredicted, statistics::units::Count::get(),
+               "Number of conditional branches predicted"),
+      ADD_STAT(condPredictedTaken, statistics::units::Count::get(),
+               "Number of conditional branches predicted as taken"),
+      ADD_STAT(condIncorrect, statistics::units::Count::get(),
+               "Number of conditional branches incorrect"),
+      ADD_STAT(predTakenBTBMiss, statistics::units::Count::get(),
+               "Number of branches predicted taken but miss in BTB"),
```
rjc-arch: Typo: 'miss' -> 'missed'

src/cpu/pred/bpred_unit.cc
```diff
@@ -419,105 +533,210 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
...
+      ADD_STAT(NotTakenMispredicted, statistics::units::Count::get(),
+               "Number branches predicted 'not taken' but turn out "
```
rjc-arch: Typo: 'turn' -> 'turned'

src/cpu/pred/bpred_unit.cc
```diff
@@ -419,105 +533,210 @@ BPredUnit::squash(const InstSeqNum &squashed_sn,
...
+               "to be taken"),
+      ADD_STAT(TakenMispredicted, statistics::units::Count::get(),
+               "Number branches predicted taken but are actually not taken"),
+      ADD_STAT(BTBLookups, statistics::units::Count::get(),
+               "Number of BTB lookups"),
+      ADD_STAT(BTBUpdates, statistics::units::Count::get(),
+               "Number of BTB lookups"),
```
rjc-arch: Typo: 'lookups' -> 'updates'

src/cpu/pred/loop_predictor.hh
```diff
@@ -1,6 +1,17 @@
...
+ * The license below extends only to copyright in the software and shall
+ * not be construed as granting a license to any other intellectual
+ * property including but not limited to intellectual property relating
+ * to a hardware implementation of the functionality of the software
+ * licensed hereunder.  You may use the software subject to the license
+ * terms below provided that you ensure that this notice is replicated
+ * unmodified and in its entirety in all distributions of the software,
+ * modified or unmodified, in source code or in binary form.
  *
+ * Copyright (c) 2014 The University of Wisconsin
```
rjc-arch: Was this line added by accident?

src/cpu/pred/ltage.hh
```diff
@@ -97,7 +109,7 @@ class LTAGE : public TAGE
 
         virtual ~LTageBranchInfo()
         {
-            delete lpBranchInfo;
+            delete lpBranchInfo; lpBranchInfo = nullptr;
```
rjc-arch: Should be on separate lines.

src/cpu/pred/ltage.cc
```diff
@@ -94,8 +106,8 @@ LTAGE::predict(ThreadID tid, Addr branch_pc, bool cond_branch, void* &b)
 
 // PREDICTOR UPDATE
 void
-LTAGE::update(ThreadID tid, Addr branch_pc, bool taken, void* bp_history,
-              bool squashed, const StaticInstPtr & inst, Addr corrTarget)
+LTAGE::update(ThreadID tid, Addr branch_pc, bool taken, void* &bp_history,
```
rjc-arch: Nit: the parameter names don't match between the declaration in the header and the definition here.

src/cpu/pred/multiperspective_perceptron.cc
```diff
@@ -614,9 +635,9 @@ MultiperspectivePerceptron::lookup(ThreadID tid, Addr instPC,
 
 void
 MultiperspectivePerceptron::update(ThreadID tid, Addr instPC, bool taken,
-                                   void *bp_history, bool squashed,
+                                   void * &bp_history, bool squashed,
```
rjc-arch: Nit: the parameter names don't match between the declaration in the header and the definition here.

src/cpu/pred/multiperspective_perceptron_tage.cc
```diff
@@ -589,9 +601,9 @@ MPP_StatisticalCorrector::condBranchUpdate(ThreadID tid, Addr branch_pc,
 
 void
 MultiperspectivePerceptronTAGE::update(ThreadID tid, Addr instPC, bool taken,
-                                   void *bp_history, bool squashed,
+                                   void * &bp_history, bool squashed,
```
rjc-arch: Nit: the parameter names don't match between the declaration in the header and the definition here.

src/cpu/pred/tage.cc
```diff
@@ -56,8 +68,8 @@ TAGE::TAGE(const TAGEParams &params) : BPredUnit(params), tage(params.tage)
 
 // PREDICTOR UPDATE
 void
-TAGE::update(ThreadID tid, Addr branch_pc, bool taken, void* bp_history,
-              bool squashed, const StaticInstPtr & inst, Addr corrTarget)
+TAGE::update(ThreadID tid, Addr branch_pc, bool taken, void * &bp_history,
```
rjc-arch: Nit: the parameter names don't match between the declaration in the header and the definition here.

src/cpu/pred/tage.cc
```diff
@@ -77,21 +89,22 @@ TAGE::update(ThreadID tid, Addr branch_pc, bool taken, void* bp_history,
...
         tage->updateStats(taken, bi->tageBranchInfo);
         tage->condBranchUpdate(tid, branch_pc, taken, tage_bi, nrand,
-                               corrTarget, bi->tageBranchInfo->tagePred);
+                               target, bi->tageBranchInfo->tagePred);
     }
 
     // optional non speculative update of the histories
-    tage->updateHistories(tid, branch_pc, taken, tage_bi, false, inst,
-                          corrTarget);
+    tage->updateHistories(tid, branch_pc, taken, tage_bi, false, inst,target);
```
rjc-arch: Nit: missing space after the comma, according to the coding guidelines.

src/cpu/pred/tage_sc_l.cc
```diff
@@ -410,8 +422,8 @@ TAGE_SC_L::predict(ThreadID tid, Addr branch_pc, bool cond_branch, void* &b)
 }
 
 void
-TAGE_SC_L::update(ThreadID tid, Addr branch_pc, bool taken, void *bp_history,
-        bool squashed, const StaticInstPtr & inst, Addr corrTarget)
+TAGE_SC_L::update(ThreadID tid, Addr branch_pc, bool taken, void *&bp_history,
```
rjc-arch: Nit: the parameter names don't match between the declaration in the header and the definition here.

src/cpu/pred/tournament.cc
```diff
@@ -134,46 +135,18 @@ TournamentBP::calcLocHistIdx(Addr &branch_addr)
 
 inline
 void
-TournamentBP::updateGlobalHistTaken(ThreadID tid)
+TournamentBP::updateGlobalHist(ThreadID tid, bool taken)
 {
-    globalHistory[tid] = (globalHistory[tid] << 1) | 1;
+    globalHistory[tid] = (globalHistory[tid] << 1) | (taken) ? 1 : 0;
```
rjc-arch: I think the parens need to go around the whole `(taken ? 1 : 0)` instead of the `(taken)`. [Or `(globalHistory[tid] << 1) | taken` would also work.]

The `|` binds more tightly than the `?:`, which I'm guessing is not what you intended. The current code sets the global history to `taken`; i.e. a history lenght of one.

src/cpu/pred/tournament.cc
```diff
@@ -134,46 +135,18 @@ TournamentBP::calcLocHistIdx(Addr &branch_addr)
...
-void
-TournamentBP::btbUpdate(ThreadID tid, Addr branch_addr, void * &bp_history)
-{
-    unsigned local_history_idx = calcLocHistIdx(branch_addr);
-    //Update Global History to Not Taken (clear LSB)
-    globalHistory[tid] &= (historyRegisterMask & ~1ULL);
-    //Update Local History to Not Taken
-    localHistoryTable[local_history_idx] =
-       localHistoryTable[local_history_idx] & (localPredictorMask & ~1ULL);
+        (localHistoryTable[local_history_idx] << 1) | (taken) ? 1 : 0;
```
rjc-arch: Same problem here.

src/cpu/pred/tournament.cc
```diff
@@ -212,51 +185,47 @@ TournamentBP::lookup(ThreadID tid, Addr branch_addr, void * &bp_history)
...
+        updateLocalHist(history->localHistoryIdx, taken);
+    }
 }
 
+
 void
 TournamentBP::update(ThreadID tid, Addr branch_addr, bool taken,
-                     void *bp_history, bool squashed,
-                     const StaticInstPtr & inst, Addr corrTarget)
+                     void * &bp_history, bool squashed,
```
rjc-arch: Nit: the parameter names don't match between the declaration in the header and the definition here.

rjc-arch (commented):
Hi David. This is all looking good! My comments below are mostly nitpicks about the coding guidelines and typos in the comments. Sorry to be pedantic, but they should be easy fixes, and it would be a shame not to fix them now.

The only real bug that I think you need to address (unless I'm mistaken) is the handling of the global and local histories in `tournament.cc`.

src/cpu/pred/bpred_unit.hh
dhschall: I just kept it as it was but since I'm anyway refactoring we can consider renaming.

The `PredictorHistory` basically captures the state at prediction time to be able to roll back.
And the `History` is a buffer of outstanding not yet committed predictions

Considering that I would say it should be `PredictionInfo` or `PredictInfo`.
The history "queue" could be `PredictQueue` or `PredictionBuffer`.
However, if I'm correct Intel calls it Branch-order-buffer` and I have seen that terminology at some point as well.

If that is ok then I'll change the naming to: 
`PredictorHistory` -> `PredictInfo`.
The `predHist` -> `bob` for branch-order-buffer.

Also, then I should change all`hist` to `pi` for an instance of `PredictInfo`

src/cpu/pred/bpred_unit.cc
dhschall: That was intentionally kept from before to let users know that this function exists.
Should I remove?

src/cpu/pred/tournament.cc
dhschall: Thanks for catching that!

src/cpu/pred/bpred_unit.cc
dhschall: In the end it should be like that. However, since this feature should be separate I'll move it out of the BPU refactoring

src/cpu/pred/bpred_unit.hh
rjc-arch: Ah, sorry, I didn't realize that was an existing name from the previous code.
We can leave it as it is if you like. but I did find it a little confusing having a `History` being a collection of `PredictorHistory`s. I see what you mean, though - the `PredictorHistory` also contains histories.

Considering the alternatives, perhaps it's better to leave things as they are. @andysan do you have an opinion on this?

andysan: I would suggest not changing it in this PR given that it is the old name. No need to make this unnecessarily complicated, I would rather merge this as soon as possible and focus on the new functionality.

src/cpu/pred/bpred_unit.cc
rjc-arch: Ack. If it's as you intended then there's no need to move it unless you prefer to.

src/cpu/pred/bpred_unit.cc
rjc-arch: Personally I'd either remove it, or leave it uncommented and guarded with a debug flag (if that's possible). A developer should know about its existence from the declaration/definition.

But it would be interesting to hear other opinions on this. @andysan do you have an opinion?

andysan: I would vote to remove it. These types of comments have a tendency to go stale and it the purpose of this function should be pretty obvious from the header file.

src/cpu/pred/bpred_unit.cc
dhschall: I removed this stat as it is captured in the new mispredictions per branch stat

Harshil2107:
Hello @andysan ,
Would it be alright if I assign this PR to you? Please be aware that being assigned to a PR implies that you are responsible for ensuring that the PR is reviewed and merged.

Another note, the above point of requiring BTB hits was brought up the gem5 community meeting and everyone agreed that we should require BTB hits to make gem5 more realistic and accurate.

src/cpu/pred/bpred_unit.cc
```diff
@@ -187,6 +192,13 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
             tid, seqNum, hist->pc,  (hist->btbHit) ? "hit" : "miss");
 
 
+    // In a high performance CPU there is no other way than a BTB hit
+    // to know about a branch instruction. In that case consolidate
+    // indirect and RAS predictor only if there was a BTB it.
+    // For low end CPUs predecoding might be used to identify branches.
+    const bool allow_btb_override = (hist->btbHit || !requiresBTBHit);
```
rjc-arch: Optional: could this be named better? It doesn't really signal overriding the BTB (in fact, when true it uses the BTB result). Might something like `use_ras_and_indirect`, or `branch_type_info_available` be clearer to future readers?

rjc-arch (approved):
Hi @dhschall. I have one optional comment about the clarity of a variable name, but overall this looks good to me as-is.

rjc-arch:
For information, I have run this patch using SPEC 2017 SimPoints, and this patch does not noticeably affect the reported IPC when the `requiresBTBHit` is disabled.
The IPC drops for some SimPoints when `requiresBTBHit` is enabled, which is expected. This would be considered more realistic behaviour for high performance front-ends.

src/cpu/pred/bpred_unit.cc
trollet91: Agree on the confusing name. Why not using something like `branch_detected` as variable name?

dhschall: Thank you for the suggestions. I renamed the variable to `branch_detected`

powerjg:
This looks good to me. Does anyone have a reason not to merge?

BobbyRBruce (approved):

andysan:
I'd like to wait for @rjc-arch to ack the change before merging. I believe he has run some more performance regressions.

Edit: Looking at the PR log, I think we can go ahead and merge this. The version reviewed by Richard is basically the same as this one.

andysan (approved):

PR #494, opened 2023-10-20
misc: Integrate a Capstone Disassembler in gem5

giactra:
Capstone is an open source disassembler [1] already used by
other projects (like QEMU).

gem5 is already capable of disassembling instructions.  Every StaticInst
is supposed to define a generateDisassembly method which returns the
instruction mnemonic (opcode + operand list) as a string.

This "distributed" implementation of a disassembler relies
on the developer to properly populate the metadata fields
of the base instruction class.
The growing complexity of the ISA code and the massive reuse
of base classes beyond their intended use has led to a
disassembling logic which contains several bugs.

By allowing a tracer to rely on a third party disassembler, we fill the
instruction trace with a more trustworthy instruction stream.

This will make any trace parsing tool to work better and it will
also allow us to spot/fix our own bugs by comparing instruction
traces with native vs custom disassembler

[1]: http://www.capstone-engine.org/

### DIFF
--- a/src/arch/arm/tracers/ArmCapstone.py
+++ b/src/arch/arm/tracers/ArmCapstone.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2018, 2023 Arm Limited
+# Copyright (c) 2023 Arm Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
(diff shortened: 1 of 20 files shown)


shingarov (approved):
This is most definitely the way to go!

I am working on a synthetic ISA backend inferred from formal specifications (so far we experimented with Sail, LibreSOC project's ISACaller, and executing Valgrind VEX symbolically) but that work is progressing very slowly.  In the meantime, having at least disassembly functioning in an ISA-agnostic manner is a great step forward.

powerjg:
Whoa... this was a bit quick to merge such a large changeset! This also touches a number of different subsystems (cpu, sim, arch-arm), and it would be good to give the community some time to check out the change.

powerjg:
@giactra Do you have some documentation on how to use this support? This seems like a cool new feature, but I worry that most people won't be able to figure out how to use it without some documentation/support.

giactra:
Hi @powerjg, the usage is pretty straightforward: you need to instantiate a CapstoneDisassembler SimObject as a child of the current tracer [1]. Needless to say it will only have an effect if the tracer is tracing via the disassembler object.

With this PR I converted both Arm TarmacTracer and the ISA agnostic ExeTracer (the default tracer in gem5) to use the disassembler.

So to enable it for Arm, the code would look like:

```
for cpu in cluster:
    cpu.tracer.disassembler = ArmCapstoneDisassembler()
```

To use it with other ISAs, someone needs to extend the base Capstone disassembler to the specific ISA. If you have a look at the Arm one, that's really few lines of code.

Let me know if you want me to write any documentation, whether a README file or whether write a paragraph for next release.

[1]: https://github.com/gem5/gem5/blob/develop/src/sim/InstTracer.py#L55

shingarov:
I have framed a chunk of time to dedicate to gem5 after the Argentinian FAST conference is over; so I can spend some time helping with this.

powerjg:
That would be great Boris!

Giacomo, I don't know the best/right way to convey the information. I agree that's it's pretty simple. However, if someone wasn't following every single PR on github or reading every single commit message they would have no way of knowing this support now exists. So, we need to write down somewhere, probably on the website, what the capstone disassembler is, why/when someone would use it, and how to use it with gem5.

I would have requested this before merging, but this change was merged before many people could see it. Less than 12 hours from posting to commit is not enough time for review for such a large change. Please give others a chance to chime in next time, Boris.

giactra:
> Giacomo, I don't know the best/right way to convey the information. I agree that's it's pretty simple. However, if someone wasn't following every single PR on github or reading every single commit message they would have no way of knowing this support now exists. So, we need to write down somewhere, probably on the website, what the capstone disassembler is, why/when someone would use it, and how to use it with gem5.
> 

Shall I add a paragraph to https://www.gem5.org/documentation/general_docs/debugging_and_testing/debugging/trace_based_debugging and maybe reference it in the next release notes?

powerjg:
That sounds great to me... Thanks for the quick response!

PR #498, opened 2023-10-21
gpu-compute,dev-hsa: ROCm 5.5+ support

abmerop:
ROCm 5.5 support including:
- Vendor packet completion signals
- Queue remapping race condition fix
- Backwards compatible GPR allocation
- Fix transient readBlob fatal reading kernel descriptor

Change-Id: Id0841407bec564c84f590c943f0609b17e01e14c

### DIFF
--- a/src/dev/amdgpu/pm4_packet_processor.cc
+++ b/src/dev/amdgpu/pm4_packet_processor.cc
@@ -384,7 +384,10 @@ PM4PacketProcessor::mapQueues(PM4Queue *q, PM4MapQueues *pkt)
                 "Mapping mqd from %p %p (vmid %d - last vmid %d).\n",
                 addr, pkt->mqdAddr, pkt->vmid, gpuDevice->lastVMID());
 
-        gpuDevice->mapDoorbellToVMID(pkt->doorbellOffset,
+        // The doorbellOffset is a dword address. We shift by two / multiply
+        // by four to get the byte address to match doorbell addresses in
+        // the GPU device.
+        gpuDevice->mapDoorbellToVMID(pkt->doorbellOffset << 2,
                                      gpuDevice->lastVMID());
 
         QueueDesc *mqd = new QueueDesc();
@@ -444,6 +447,8 @@ PM4PacketProcessor::processMQD(PM4MapQueues *pkt, PM4Queue *q, Addr addr,
 
     DPRINTF(PM4PacketProcessor, "PM4 mqd read completed, base %p, mqd %p, "
             "hqdAQL %d.\n", mqd->base, mqd->mqdBase, mqd->aql);
+
+    gpuDevice->processPendingDoorbells(offset);
 }
 
 void
@@ -472,6 +477,8 @@ PM4PacketProcessor::processSDMAMQD(PM4MapQueues *pkt, PM4Queue *q, Addr addr,
     // Register doorbell with GPU device
     gpuDevice->setSDMAEngine(pkt->doorbellOffset << 2, sdma_eng);
     gpuDevice->setDoorbellType(pkt->doorbellOffset << 2, RLC);
+
+    gpuDevice->processPendingDoorbells(pkt->doorbellOffset << 2);
 }
 
 void
@@ -576,6 +583,7 @@ PM4PacketProcessor::unmapQueues(PM4Queue *q, PM4UnmapQueues *pkt)
         gpuDevice->deallocatePasid(pkt->pasid);
         break;
       case 2:
+        panic("Unmapping queue selection 2 unimplemented\n");
         break;
       case 3: {
         auto &hsa_pp = gpuDevice->CP()->hsaPacketProc();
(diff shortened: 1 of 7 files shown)


src/dev/hsa/hsa_packet.hh
```diff
@@ -100,6 +100,13 @@ struct _hsa_barrier_or_packet_t
     uint64_t completion_signal;
 };
 
+struct _hsa_generic_vendor_pkt
+{
+    uint32_t padding[14];
+    uint64_t completion_signal;
+};
+static_assert(sizeof(_hsa_generic_vendor_pkt) == 64);
```
mattsinc, line 109: Do we need to document why 64?

src/gpu-compute/gpu_command_processor.cc
```diff
@@ -473,18 +473,27 @@ GPUCommandProcessor::driver()
...
- *       the proper handling of any required vendor-specific packets.
- *       In the version of ROCm that is currently supported (1.6)
- *       the runtime will send packets that direct the CP to
- *       invalidate the GPUs caches. We do this automatically on
- *       each kernel launch in the CU, so this is safe for now.
+ * TODO: For now we simply tell the HSAPP to finish the packet and write a
+ * completion signal, if any. However, in the future proper handing may be
+ * required for vendor specific packets.
+ *
+ * In the version of ROCm that is currently supported the runtime will send
```
mattsinc, line 513: Be explicit about version?

mattsinc (changes requested):

src/dev/hsa/hsa_packet.hh
abmerop: I can change to Addr which should make it implicitly clear why.

src/gpu-compute/gpu_command_processor.cc
abmerop: FS mode will work on anything 4.0-5.4 right now.  I kind of don't want to have to keep updating a comment. The one I replaced was still referencing ROCm 1.6.  IMO user can check documentation for what "supported" versions means

mattsinc: Maybe we should just add a pointer to the documentation here then?

abmerop:
Adding a few more commits to this as they are somewhat related to support for latest ROCm.

src/gpu-compute/gpu_command_processor.cc
abmerop: OK.  Doesn't seem to be any documentation though, so not sure what to point to right now

src/dev/amdgpu/pm4_packet_processor.cc
```diff
@@ -384,7 +384,7 @@ PM4PacketProcessor::mapQueues(PM4Queue *q, PM4MapQueues *pkt)
                 "Mapping mqd from %p %p (vmid %d - last vmid %d).\n",
                 addr, pkt->mqdAddr, pkt->vmid, gpuDevice->lastVMID());
 
-        gpuDevice->mapDoorbellToVMID(pkt->doorbellOffset,
+        gpuDevice->mapDoorbellToVMID(pkt->doorbellOffset << 2,
```
mattsinc, line 390: If possible, it would be great to document why << 2

src/dev/hsa/hsa_packet.hh
```diff
@@ -100,6 +100,13 @@ struct _hsa_barrier_or_packet_t
     uint64_t completion_signal;
 };
 
+struct _hsa_generic_vendor_pkt
+{
+    uint32_t padding[14];
+    Addr completion_signal;
+};
+static_assert(sizeof(_hsa_generic_vendor_pkt) == 64);
```
mattsinc, line 109: Same thing here, it would be great if we could explain why 64.

src/gpu-compute/gpu_command_processor.cc
mattsinc: I think the update is fine then, thanks.

mattsinc (dismissed):
approving to at least run tests.  Some commenting/documentation questions though.

powerjg:
@mattsinc you don't have to approve to run the tests. They run as long as (1) it's not a draft and (2) the submitter has contributed to gem5 before.

By "approving" the PR you are allowing the PR to be merged *as is* without any changes.

mattsinc:
> @mattsinc you don't have to approve to run the tests. They run as long as (1) it's not a draft and (2) the submitter has contributed to gem5 before.
> 
> By "approving" the PR you are allowing the PR to be merged _as is_ without any changes.

I did not mean to press it, sorry.  I was trying to scroll and click on "Quote Reply", but my web browser was lagging and clicked this instead ... will try to avoid moving forward.  I did not see a way to undo once I'd clicked it.

powerjg:
np, that's why I asked :)

You can go back and select "request changes" to prevent merging, if you want.

image

src/dev/amdgpu/pm4_packet_processor.cc
abmerop: Added comment

src/dev/hsa/hsa_packet.hh
abmerop: Added comment

mattsinc (approved):

PR #502, opened 2023-10-24
Support for classic prefetchers in Ruby

tiagormk:
This patch  adds supports for using the "classic" prefetchers with ruby cache controllers.

This pull request includes a few commits making the changes in this order:
- Refactor decouples the classic cache and prefetchers interfaces
- Extras probes for later integration with ruby
- General ruby-side support
- Adds support for the CHI protocol

Commit [mem-ruby: support prefetcher in CHI protocol](https://github.com/gem5/gem5/commit/2bdb65653bf81b85de8b40c425b1a9614f7e5feb) may be used as example on how to add support for other protocols.

JIRA issues that may be related to this pull request:
    https://gem5.atlassian.net/browse/GEM5-457
    https://gem5.atlassian.net/browse/GEM5-1112

### DIFF
--- a/src/mem/cache/base.hh
+++ b/src/mem/cache/base.hh
@@ -327,6 +327,10 @@ class BaseCache : public ClockedObject
         bool hasBeenPrefetched(Addr addr, bool is_secure) const override
         { return cache.hasBeenPrefetched(addr, is_secure); }
 
+        bool hasBeenPrefetched(Addr addr, bool is_secure,
+                               RequestorID requestor) const override
+        { return cache.hasBeenPrefetched(addr, is_secure, requestor); }
+
         bool inMissQueue(Addr addr, bool is_secure) const override
         { return cache.inMissQueue(addr, is_secure); }
 
@@ -1277,11 +1281,14 @@ class BaseCache : public ClockedObject
 
     bool hasBeenPrefetched(Addr addr, bool is_secure) const {
         CacheBlk *block = tags->findBlock(addr, is_secure);
-        if (block) {
-            return block->wasPrefetched();
-        } else {
-            return false;
-        }
+        return block && block->wasPrefetched();
+    }
+
+    bool hasBeenPrefetched(Addr addr, bool is_secure,
+                           RequestorID requestor) const {
+        CacheBlk *block = tags->findBlock(addr, is_secure);
+        return block && block->wasPrefetched() &&
+               (block->getSrcRequestorId() == requestor);
     }
 
     bool inMissQueue(Addr addr, bool is_secure) const {
--- a/src/mem/cache/base.hh
+++ b/src/mem/cache/base.hh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012-2013, 2015-2016, 2018-2019 ARM Limited
+ * Copyright (c) 2012-2013, 2015-2016, 2018-2019, 2023 ARM Limited
  * All rights reserved.
  *
  * The license below extends only to copyright in the software and shall
@@ -59,6 +59,7 @@
 #include "debug/CachePort.hh"
 #include "enums/Clusivity.hh"
 #include "mem/cache/cache_blk.hh"
+#include "mem/cache/cache_probe_arg.hh"
 #include "mem/cache/compressors/base.hh"
 #include "mem/cache/mshr_queue.hh"
 #include "mem/cache/tags/base.hh"
@@ -113,28 +114,6 @@ class BaseCache : public ClockedObject
         Blocked_NoWBBuffers = MSHRQueue_WriteBuffer,
         Blocked_NoTargets,
         NUM_BLOCKED_CAUSES
-    };
-
-    /**
-     * A data contents update is composed of the updated block's address,
-     * the old contents, and the new contents.
-     * @sa ppDataUpdate
-     */
-    struct DataUpdate
-    {
-        /** The updated block's address. */
-        Addr addr;
-        /** Whether the block belongs to the secure address space. */
-        bool isSecure;
-        /** The stale data contents. If zero-sized this update is a fill. */
-        std::vector<uint64_t> oldData;
-        /** The new data contents. If zero-sized this is an invalidation. */
-        std::vector<uint64_t> newData;
-
-        DataUpdate(Addr _addr, bool is_secure)
-          : addr(_addr), isSecure(is_secure), oldData(), newData()
-        {
-        }
     };
 
   protected:
@@ -336,6 +315,26 @@ class BaseCache : public ClockedObject
 
   protected:
 
+    struct CacheAccessorImpl : CacheAccessor
+    {
+        BaseCache &cache;
+
+        CacheAccessorImpl(BaseCache &_cache) :cache(_cache) {}
+
+        bool inCache(Addr addr, bool is_secure) const override
+        { return cache.inCache(addr, is_secure); }
+
+        bool hasBeenPrefetched(Addr addr, bool is_secure) const override
+        { return cache.hasBeenPrefetched(addr, is_secure); }
+
+        bool inMissQueue(Addr addr, bool is_secure) const override
+        { return cache.inMissQueue(addr, is_secure); }
+
+        bool coalesce() const override
+        { return cache.coalesce(); }
+
+    } accessor;
+
     /** Miss status registers */
     MSHRQueue mshrQueue;
 
@@ -352,20 +351,20 @@ class BaseCache : public ClockedObject
     prefetch::Base *prefetcher;
 
     /** To probe when a cache hit occurs */
-    ProbePointArg<PacketPtr> *ppHit;
+    ProbePointArg<CacheAccessProbeArg> *ppHit;
 
     /** To probe when a cache miss occurs */
-    ProbePointArg<PacketPtr> *ppMiss;
+    ProbePointArg<CacheAccessProbeArg> *ppMiss;
 
     /** To probe when a cache fill occurs */
-    ProbePointArg<PacketPtr> *ppFill;
+    ProbePointArg<CacheAccessProbeArg> *ppFill;
 
     /**
      * To probe when the contents of a block are updated. Content updates
      * include data fills, overwrites, and invalidations, which means that
      * this probe partially overlaps with other probes.
      */
-    ProbePointArg<DataUpdate> *ppDataUpdate;
+    ProbePointArg<CacheDataUpdateProbeArg> *ppDataUpdate;
 
     /**
      * The writeAllocator drive optimizations for streaming writes.
--- a/src/mem/cache/cache_probe_arg.hh
+++ b/src/mem/cache/cache_probe_arg.hh
@@ -59,6 +59,10 @@ struct CacheAccessor
     /** Determine if address has been prefetched */
     virtual bool hasBeenPrefetched(Addr addr, bool is_secure) const = 0;
 
+    /** Determine if address has been prefetched by the requestor */
+    virtual bool hasBeenPrefetched(Addr addr, bool is_secure,
+                                   RequestorID requestor) const = 0;
+
     /** Determine if address is in cache miss queue */
     virtual bool inMissQueue(Addr addr, bool is_secure) const = 0;
 
--- a/src/mem/cache/cache_probe_arg.hh
+++ b/src/mem/cache/cache_probe_arg.hh
@@ -95,17 +95,23 @@ struct CacheDataUpdateProbeArg
     Addr addr;
     /** Whether the block belongs to the secure address space. */
     bool isSecure;
+    /** Block original requestor */
+    const RequestorID requestorID;
     /** The stale data contents. If zero-sized this update is a fill. */
     std::vector<uint64_t> oldData;
     /** The new data contents. If zero-sized this is an invalidation. */
     std::vector<uint64_t> newData;
+    /** Set if the update is from a prefetch or evicting a prefetched
+    *  block that was never used. */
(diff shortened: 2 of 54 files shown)


src/mem/cache/cache_probe_arg.hh
```diff
@@ -0,0 +1,104 @@
...
+ * @sa ppDataUpdate
+ */
+struct CacheDataUpdateProbeArg
+{
+    /** The updated block's address. */
+    Addr addr;
+    /** Whether the block belongs to the secure address space. */
+    bool isSecure;
+    /** The stale data contents. If zero-sized this update is a fill. */
+    std::vector<uint64_t> oldData;
```
powerjg, line 105: Does this make more since to be `std::vector<uint8_t>`?

src/mem/cache/prefetch/base.hh
```diff
@@ -426,7 +424,7 @@ class Base : public ClockedObject
      * @param pkt The memory request causing the event
```
powerjg: Quite a bit of this documentation is out of date now. Given the change in the design, it would be great to better document the parameters etc.

src/mem/cache/prefetch/base.cc
```diff
@@ -279,8 +267,8 @@ Base::regProbeListeners()
      * parent cache using the probe "Miss". Also connect to "Hit", if the
      * cache is configured to prefetch on accesses.
      */
-    if (listeners.empty() && cache != nullptr) {
-        ProbeManager *pm(cache->getProbeManager());
+    if (listeners.empty() && probeManager != nullptr) {
+        ProbeManager *pm(probeManager);
```
powerjg: Nit: This local variable could be dropped. It's not a big deal now, but I'm thinking about future maintenance and this could become confusing.

src/mem/cache/cache_probe_arg.hh
```diff
@@ -84,17 +84,23 @@ struct CacheDataUpdateProbeArg
     Addr addr;
     /** Whether the block belongs to the secure address space. */
     bool isSecure;
+    /** Block original requestor */
+    const RequestorID requestorID;
```
powerjg, line 103: Feel free to say "no", but would it make sense to store the thread context here instead of in the probe?

src/mem/ruby/structures/RubyPrefetcherProxy.cc
```diff
@@ -0,0 +1,229 @@
...
+                          MessageBuffer *_pf_queue)
+    :Named(_parent->name()),
+    prefetcher(_prefetcher),
+    cacheCntrl(_parent),
+    pfQueue(_pf_queue),
+    pfEvent([this]{ issuePrefetch(); }, name()),
+    ppHit(nullptr), ppMiss(nullptr),
+    ppFill(nullptr), ppDataUpdate(nullptr)
+{
+    if (prefetcher) {
```
powerjg, line 62: Why would this ever be null? Maybe we should have an assert?

Similar for `cacheCntrl` and `pfQueue`, we should `fatal_if` these are null.

powerjg (commented):
Just a few small things. Overall, everything looks great to me. I'm excited about merging this change!

src/mem/cache/cache_probe_arg.hh
tiagormk: It's indeed a bit odd. I'm assuming 64bit chunks make things easier for src/mem/cache/compressors/* which are the consumers of this vector.

src/mem/cache/prefetch/base.hh
tiagormk: do you mean add more documentation for the refactored classes in src/mem/cache/cache_probe_arg.hh and src/mem/ruby/structures/RubyPrefetcherProxy.hh ?

src/mem/cache/prefetch/base.cc
tiagormk: ack

src/mem/cache/cache_probe_arg.hh
tiagormk: This should be ID of either the cache object or the prefetcher object that brought in the line. I'm not sure how we could get this from the thread context here.

src/mem/ruby/structures/RubyPrefetcherProxy.cc
tiagormk: The idea is that the proxy may not be attached to any prefetcher. This make thinks a bit cleaner on the SLICC and config side to support a config both with or without prefetcher (see the follow-up patch that adds the CHI support). The alternative would be to either conditionally create the proxy in SLICC, or let it be another simobj defined by the python script.

cacheCntrl and pfQueue I'd agree should always be non-null and will add the check.

src/mem/cache/prefetch/base.hh
powerjg: I mean for the refactored functions/classes. E.g., this line says `param pkt`, but it's now `param arg` (or `cache`... github doesn't show me the next line 😞)

src/mem/cache/cache_probe_arg.hh
powerjg: Ack. Thanks 😄

src/mem/cache/prefetch/base.hh
tiagormk: ack, updated the docs with mismatch

powerjg (commented):
Everything looks great to me! @giactra, @mattsinc, or @abmerop, does anyone else want to review before we merge?

mattsinc:
> Everything looks great to me! @giactra, @mattsinc, or @abmerop, does anyone else want to review before we merge?

I am not an expert at this part of the codebase (prefetching), so if you feel it's good to go, it's fine with me.

powerjg:
> I am not an expert at this part of the codebase (prefetching), so if you feel it's good to go, it's fine with me.

@mattsinc you *are* a Ruby expert and there are a number of changes to generic Ruby interfaces as well, so I thought you might want to take a look.

src/mem/ruby/structures/RubyPrefetcherProxy.hh
```diff
@@ -0,0 +1,182 @@
...
+
+    bool hasBeenPrefetched(Addr addr, bool is_secure,
+                            RequestorID requestor) const override
+    {
+        return const_cast<AbstractController*>(cacheCntrl)->
+            hasBeenPrefetched(addr, is_secure, requestor);
+    }
+
+    bool inMissQueue(Addr addr, bool is_secure) const override
+    {
```
giactra: Are these const casts really needed?

src/mem/cache/base.hh
```diff
@@ -115,28 +116,6 @@ class BaseCache : public ClockedObject
         NUM_BLOCKED_CAUSES
     };
 
-    /**
```
mattsinc, line 118: The answer must be no, since tests are passing, but how does this change not affect the replacement policy support?

jjuninho:
Hi @tiagormk,
`RUSC` state is missing from the initial state list for the `Prefetch` event.
```
transition({UD,UD_T,SD,UC,SC,RU,RSC,RSD,RUSD,SC_RSC,SD_RSC,SD_RSD,UC_RSC,UC_RU,UD_RU,UD_RSD,UD_RSC}, Prefetch) {
  Callback_ExpressPrefetchHit;
  Pop_ReqRdyQueue;
}
```
This causes "Invalid transition" panic when `Prefetch` is requested to `RUSC` line.

src/mem/ruby/structures/RubyPrefetcherProxy.hh
powerjg: @tiagormk can you address this comment this week so we can merge this change before the staging branch for version 23.1 is created on Friday? Thanks!

src/mem/cache/base.hh
powerjg: I think this is not a problem. From my understanding, the replacement policies and the prefetchers are completely orthogonal and do not share any base code. So, any changes to prefetchers should not impact replacement policies. Let me know if I'm mistaken.

tiagormk: I don't think there is any dependency between the prefetchers and the replacement policies.

src/mem/ruby/structures/RubyPrefetcherProxy.hh
tiagormk: thanks @giactra, indeed not necessary. I will be pushing an updated patch shortly.

tiagormk:
Added new commit fixing the issue mentioned by @jjuninho

powerjg (approved):
Looks good to me!

yichungchen84:
Hi teams, does this new feature support cluster CPUs? I have attached multiple CPUs to a CHI_RNF node and it seems to have the following issues:

1. src/mem/cache/prefecth/base.cc, line 120 assert(!system && !probeManager): since multiple CPUs point to the same RNF, the system are the same and therefore it leads to this assertion failure.
2. CHI-cache-func.sm, line 428 assert (in_msg.txnId == max_outstanding_transactions): I haven't figured out what could possibly cause this assertion. need a correct TBE allocation in the cluster?
3. CHI-cache-actions.sm, line 1382 assert(is_HN == false): because of the cluster? 

Thanks for the help.

tiagormk:
@yichungchen84 the first error must be related to assigning the same prefetcher object to multiple caches. I believe that's not supported yet. I'm not sure about the others.

Could you share more details on your config (e.g. your config.json) ?

yichungchen84:
@tiagormk thanks for the help. I fixed the first issue. The second issue is new in 23.1.0.0. Do you happen to know anyone can help with this?

tiagormk:
@yichungchen84

CHI-cache-func.sm, line 428 assert (in_msg.txnId == max_outstanding_transactions): this should be fixed by https://github.com/gem5/gem5/pull/734
Can you give it a try ?

CHI-cache-actions.sm, line 1382 assert(is_HN == false): this could be an intermediate cache controller (e.g.: a shared L3 cache) misconfigured as a home node. Only the last-level caches can be the home nodes. Could you share more details on your config (e.g. your config.json) ? And also run with --debug-flags=RubyGenerated so we can see which transaction/cache controller is triggering the assert.

yichungchen84:
@tiagormk 
Thanks. it fixed the issue.

CHI-cache-actions.sm, line 1382 <= I figured out the root cause of the issue. LLC (in this case SLC) at HNF does not support prefetcher. (It makes sense since the LLC is close to main memory. Someone may need it in the future?)

A new issue: 
CHI-cache-actions.sm, line 3563/3590 assert(isWriteReqType(tbe.reqType)) <= any idea how I can fix this one? 

I would like to share the config.json. Unfortunately, the file is secured by IT. It may take a while to export it.

tiagormk:
Hi @yichungchen84, if you can't share the full config you can try running with --debug-flags=RubyGenerated and let us know which transaction triggers that assert.

yichungchen84:
@tiagormk I'm on it. With the flag, it may need 10X time. (I'm running full system mode)

tiagormk:
@yichungchen84 if logs are too big you can use --debug-start to start print closer to the failure point. Gemn5 prints out the sim tick when the error happens, e.g. if the error happens at tick 10000000, you could use --debug-start=9000000 to start logging 1million ticks before that.

yichungchen84:
@tiagormk Thanks for the help. 

Here is the error message (only the last step):
50000: AllocSeqRequest
50000: executing AllocateTBE_SeqRequest
50000: next_state: UD
50000: AtomicLoad
50000: Initiate_Request
50000: Initiate_Atomic_UD
50000: Profile_Hit
error: CHI-cache-actions.sm:3590 assert failure
(system.terminal => Run /sbin/init as init process)

(I probably can send you files by email due to many restrictions set by IT. Let me know if you need more information. )

yichungchen84:
@tiagormk 
Hi, I found out that it is probably not a prefetcher-related issue. It relates to a CHI atomic operation that is probably caused by 23.1.0.0 update with the Far atomic implement in CHI. My current solution is to add AtomicStore to isWriteReqType and AtomicLoad to isReadReqType (CHI-cache-func.sm). Please let me know if this is a way to solve the issue.

tiagormk:
Thanks @yichungchen84 for the debugging ! I think the correct fix is to include both AtomicLoad and AtomicStore in isWriteReqType (since all atomics are technically stores). Can you give that a shot and push the fix ?

yichungchen84:
@tiagormk  Thanks for the help. I confirm that the solution works.
I encountered other issues
either 
1. 
exe: SnpInvPipe
exe: ProcessNextState
next_state: BUSY_BLKD
state: BUSY_BLKD, event: SendInvSnpResp
exe: Pop_triggerQueue
exe: Send_InvSnpResp
Runtime Error at CHI=cache-actions.sm:1525 => assert(tbe.dataMaybeDirtyUpstream == false)
2.
state: BUSY_BLKD, even: Final
exe: Pop_TriggerQueue
exe: Finalize_UpdateCacheFromTBE
exe: Finalize_UpdateDirectoryFromTBE
exe: Finalize_DeallocateRequest
next_state:SC
panic: Possible Deadlock detected

any ideas?
My design has a cluster of CPUs sharing a L3. The cluster attached to a RNF, If I take out the L3, it works without above issues.

---
New info: I tried the same configuration with 23.1.0.0 and 23.0.1.0. Only 23.1.0.0  has the error 2. I was wondering if there is an CHI update causing the issue. (Both CHI_config files disable atomic and prefetch settings.)

---
New info: After rebuilding GEM5 with only CHI-prefetcher related files from 23.0.1.0, I found that the atomic feature is the problem.

PR #503, opened 2023-10-25
[DO NOT MERGE][Test-only] test kconfig weekly tests and gpu tests

rogerchang23424:


powerjg:
I'm not sure this is the best way to test things... we don't have *free* compute resources. By running the weekly tests like this you're taking valuable CI testing resources from others in the community.

In this case, we don't need to ensure that the tests completely pass, only that the *build* works. Would it be possible to simply manually inspect the files instead of using 3-5 days of testing resources?

powerjg:
Apologies. I need to convert this to a draft so it stops taking all of our CI testing resources.

rogerchang23424:
OK, I have removed the very-long tests in the weekly and gpu tests and only leave the tests need to build

rogerchang23424:
The PR is just test if the weekly test will break after kconfig changes. Close the PR we don't need to run test anymore.

PR #507, opened 2023-10-25
configs,ext: gem5 SST bridge calls m5.instantiate() in gem5

kaustav-goswami:
This change updates the gem5 SST bridge to call m5.instantiate() in the gem5 config script instead of in the SST component. This allows more flexibility for the gem5-SST setup, as we can now write traffic generators using the bridge.

Change-Id: I510a8c15f8fb00bdbdd60dafa2d9f5ad011e48f2

### DIFF
--- a/ext/sst/gem5.hh
+++ b/ext/sst/gem5.hh
@@ -80,7 +80,6 @@
 #include <sst/core/component.h>
 
 #include <sst/core/interfaces/stringEvent.h>
-#include <sst/core/interfaces/simpleMem.h>
 
 #include <sim/simulate.hh>
 
--- a/configs/example/sst/arm_fs.py
+++ b/configs/example/sst/arm_fs.py
@@ -185,3 +185,4 @@ for cpu in system.cpu:
     cpu.createInterruptController()
 
 root = Root(full_system=True, system=system)
+m5.instantiate()
--- a/configs/example/sst/riscv_fs.py
+++ b/configs/example/sst/riscv_fs.py
@@ -175,3 +175,4 @@ for cpu in system.cpu:
     cpu.createInterruptController()
 
 root = Root(full_system=True, system=system)
+m5.instantiate()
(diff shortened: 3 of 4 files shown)


ext/sst/gem5.hh
```diff
@@ -80,7 +80,6 @@
 #include <sst/core/component.h>
 
 #include <sst/core/interfaces/stringEvent.h>
-#include <sst/core/interfaces/simpleMem.h>
```
powerjg, line 83: Why this change?

powerjg (commented):
Seems OK to me. One problem I foresee is that this isn't going to be compatible with how the stdlib does startup. This problem doesn't need to be solve now, but I would like to see the SST stuff be compatible with the stdlib in the long-term.

One small thing below to address.

ext/sst/gem5.hh
kaustav-goswami: My mistake here. This change was supposed to be in https://github.com/gem5/gem5/pull/396 to remove all deprecation warning messages. Do I revert this and create a different PR for this change? Or do I update the commit note?

kaustav-goswami:
> Seems OK to me. One problem I foresee is that this isn't going to be compatible with how the stdlib does startup. This problem doesn't need to be solve now, but I would like to see the SST stuff be compatible with the stdlib in the long-term.
> 
> One small thing below to address.

I agree that we cannot the simulator module with SST-based boards. I think to solve this issue, we need to make changes in the simulator module and add an (Mem) Interface that is based on the OutgoingRequestBridge.

ext/sst/gem5.hh
powerjg: You don't have to make a new PR, but if you could split this out into its on commit it would be better.

kaustav-goswami: I have split this into two commits in the latest patch.

powerjg:
Can you rebase on current develop?

powerjg (approved):

PR #510, opened 2023-10-26
util: Added script to copy resources from mongodb

### LINKED ISSUE #505: Provide a way to replicate tet suite resources storage setup.
Enchencement feature:

In our lab, we don't have access to external network everywhere.
Before gem5 v23, we used to duplicate the resources into an internal database.
This way we could run the test suite without accessing outside network. 

Since v23, the resource setup became much much more complex and I didn't find a way to replicate this setup internaly.

Providing an automated way to replicate the resource storage would allow:
- Using gem5 test suite in environment with no external network with miniml effort.
- Orgs can use their own servers to store the resources. Reduce the demand on UCDavis servers.

Harshil2107:
- This script copies all resources from a mongodb database locally The script creates a resources.json and downloads all the resources. It also updates the resources.json to point to these local downloads instead of the cloud bucket.

Change-Id: I15480c4ba82bbf245425205c9c1ab7c0f3501cc3


util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,181 @@
...
+    )
+    parser.add_argument(
+        "--output_dir",
+        type=str,
+        help="Output directory absolute path, default is the cwd. The resources.json and all resources will be saved in this directory",
+    )
+    args = parser.parse_args()
+
+    if args.output_dir == None:
+        output_path = pathlib.Path(pathlib.Path.cwd())
```
powerjg:
```suggestion
        output_path = pathlib.Path.cwd()
```

I think this is equivalent

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,181 @@
...
+    )
+    parser.add_argument(
+        "--collection_name",
+        type=str,
+        default="resources",
+        help="Name of the collection",
+    )
+    parser.add_argument(
+        "--auth_url",
+        default="https://realm.mongodb.com/api/client/v2.0/app/data-ejhjf/auth/providers/api-key/login",
```
powerjg: \> 80 characters

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,181 @@
...
+    data = json.dumps(data).encode("utf-8")
+    req = request.Request(
+        auth_url, data=data, headers={"content-type": "application/json"}
+    )
+    try:
+        response = request.urlopen(req)
+    except Exception as e:
+        print(e)
+        print("Failed to get token")
+        exit(1)
```
powerjg: Don't print exceptions, raise them or handle them.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,181 @@
...
+        "Content-Type": "application/json",
+    }
+
+    req = request.Request(url, data=data, headers=headers)
+    try:
+        response = request.urlopen(req)
+    except Exception as e:
+        print("Failed to get resources")
+        print(e)
+        exit(1)
```
powerjg: Don't print exceptions. Raise them or handle them.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def save_resources_to_file(resources, output):
+    path = output.joinpath("resources.json")
+
+    with open(path, "w") as f:
+        json.dump(resources, f, indent=4)
+
+
+def download_resources(output):
```
powerjg: Generally, types would be helpful. It would also be good to have docstrings for these functions.

Another point: Doesn't "download_resource.py" or something like that already have this function?

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,181 @@
...
+    if not path.exists():
+        path.mkdir()
+
+    with open(output.joinpath("resources.json")) as f:
+        resources = json.load(f)
+
+    for resource in resources:
+        if "url" in resource.keys():
+            url = resource["url"]
+            filename = url.split("/")[-1]
```
powerjg: What about `Path(urllib.parse(url).path).name` or something like that. I don't love the idea of using string splitting. But this is a small nit. Whatever you think is more readable.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,181 @@
...
+            filepath = pathlib.Path(path).joinpath(filename)
+            if not filepath.exists():
+                request.urlretrieve(url, filepath)
+                resource["url"] = "file://" + str(filepath)
```
powerjg: What about `filepath.as_uri()`?

https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_uri

util/offline_db/get_resources_from_db.py
Harshil2107: I added typing and docstrings.

Yes, in `downloader.py` the `_download()` function does download. It is a more complex function and the majority of the above function will remain the same. The only change will be to call `_download()` function instead of `urlretrieve` as I will still need to parse which resources can be downloaded and get the filepath.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,214 @@
...
+
+    try:
+        response = request.urlopen(req)
+    except Exception as e:
+        raise Exception(f"Failed to get resources: {e}")
+
+    return json.loads(response.read().decode("utf-8"))["documents"]
+
+
+def save_resources_to_file(resources: List, output: "pathlib.Path"):
```
powerjg:
```suggestion
def save_resources_to_file(resources: List, output: pathlib.Path):
```
You shouldn't need quotes

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,214 @@
...
+    with open(output.joinpath("resources.json")) as f:
+        resources = json.load(f)
+
+    for resource in resources:
+        if "url" in resource.keys():
+            url = resource["url"]
+            filename = pathlib.Path(parse.urlsplit(url).path).name
+            filepath = pathlib.Path(path).joinpath(filename)
+
+            if not filepath.exists():
```
powerjg: if it *does* exist, you should probably double check the md5sum in case it's a new version.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,214 @@
...
+                request.urlretrieve(url, filepath)
+
+                resource["url"] = "file://" + filepath.as_uri()
```
powerjg: I think this will have an extra `file://`

```suggestion
                resource["url"] = filepath.as_uri()
```

util/offline_db/get_resources_from_db.py
powerjg: I was actually referring to the `util/obtain-resource.py` script.

I'm ok with keeping this script as it is and not trying to combine with `util/obtain-resource.py`

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,214 @@
...
+
+    with open(output.joinpath("resources.json"), "w") as f:
+        json.dump(resources, f, indent=4)
+
+
+if __name__ == "__main__":
```
powerjg:
```suggestion
if __name__ == "__main__" or __name__ == "__m5_main__":
```

This way it's compatible with gem5 as well as python.

powerjg (commented):
A few more little things.

Harshil2107:
I am trying to get the download progress working, I am having some issues trying to fix import errors when I try to use `src/python/gem5/utils/progress_bar.py`.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,239 @@
...
+            ):
+                with tqdm(
+                    unit="B",
+                    unit_scale=True,
+                    unit_divisor=1024,
+                    miniters=1,
+                    desc=f"Downloading {url}\n",
+                ) as t:
+                    request.urlretrieve(url, filepath, progress_hook(t))
+                resource["url"] = filepath.as_uri()
```
trollet91: Is this line needed? If yes needs absolute() before as_uri, otherwise crash in my system.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,239 @@
...
+                != resource["md5sum"]
+            ):
+                with tqdm(
+                    unit="B",
+                    unit_scale=True,
+                    unit_divisor=1024,
+                    miniters=1,
+                    desc=f"Downloading {url}\n",
+                ) as t:
+                    request.urlretrieve(url, filepath, progress_hook(t))
```
trollet91: The tqdm display is broken in my system. It does not override the current line and creates a new line at each update. If this is not fixable, a print print(f"Download {url} -> {filepath}") could be enough.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,239 @@
...
+    if not path.exists():
+        path.mkdir()
+
+    with open(output.joinpath("resources.json")) as f:
+        resources = json.load(f)
+
+    for resource in resources:
+        if "url" in resource.keys():
+            url = resource["url"]
+            filename = pathlib.Path(parse.urlsplit(url).path).name
```
trollet91: This line induces a bug in the case multiple resources have the same filename.
Ex: binaries for each isa can have same name but different path in url.
https://storage.googleapis.com/dist.gem5.org/develop/test-progs/hello/bin/arm/linux/hello32-static
https://storage.googleapis.com/dist.gem5.org/develop/test-progs/hello/bin/x86/linux/hello32-static
The current behavior is to download all the resources in the same folder using the last part of the url (here hello32-static). Each new download will overwrite previously downloaded hello32-static file.

Solution:
Keep the directory structure in the output directory.
This can be done by removing the following prefix 'storage.googleapis.com/dist.gem5.org/dist/develop'.
This is a bit hacky. A proper solution would have to set this path information in the json that is downloaded. This way won't have to do string manipulation on the url

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,239 @@
...
+            filepath = pathlib.Path(path).joinpath(filename)
+
+            if (
+                not filepath.exists()
+                or hashlib.md5(filepath.read_bytes()).hexdigest()
```
trollet91: This check does not work for me.
Even after downloading a file, the md5 does not match the md5 in the json file.
Are you sure that the md5 in the json file are correct?

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,239 @@
...
+        help="Output directory absolute path, default is the cwd."
+        "The resources.json and all resources will be saved in this directory",
+    )
+    args = parser.parse_args()
+
+    if args.output_dir == None:
+        output_path = pathlib.Path.cwd()
+    else:
+        output_path = pathlib.Path(args.output_dir)
+
```
trollet91: You can add an assert that either output path does not exist or that it's a directory.
In the case it does not exist, call os.makedirs(output_path)

trollet91 (changes requested):

powerjg:
@trollet91 Thanks for the detailed review! This is really helpful!

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,239 @@
...
+    with open(output.joinpath("resources.json"), "w") as f:
+        json.dump(resources, f, indent=4)
+
+
+if __name__ == "__main__" or __name__ == "__m5_main__":
+    parser = argparse.ArgumentParser(
+        description="Get resources from the database"
+    )
+    parser.add_argument(
```
trollet91: Another suggestion would be to replace all these options (excpet --output_dir) by an option that takes the path to the gem5_deafult_config.py file. We could then parse the config dict inside this script to get the different clients used by Gem5 test suite.

This will prevent this script to lack behind each time gem5_deafult_config.py is updated.

util/offline_db/get_resources_from_db.py
Harshil2107: I noticed that the md5sum for resources that are tarred or zipped were taken before tarring and zipping, that is why they are different. 

There are two solutions that I can think of:

- Untar and unzip the file and check the md5sum with the JSON.
- Check the md5 sum of the downloaded file with the new download but this already downloads the file again so it seems redundant.

Harshil2107:
There is another bug that I found, the resources with new versions with the same URL as older versions will not update to point to the local version. I will push another commit with all the requested changes and this bug fixed.

Harshil2107:
There is still an error with md5sum, it is because the md5 on the JSON is that of the decompressed file. 

Potential solution:
- add a field to the JSON that refers to the compressed md5 as well.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,223 @@
...
+        default=pathlib.Path.cwd(),
+        help="Output directory absolute path, default is the cwd."
+        "The resources.json and all resources will be saved in this directory",
+    )
+    args = parser.parse_args()
+
+    output_path = pathlib.Path(args.output_dir)
+
+    # Get the gem5 config from the stable branch of gem5
+    gem5_config = request.urlopen(
```
trollet91: As a user I want this script to download the resources needed from my gem5 repo. This repo might be a few version behind upstream release or be based on latest develop.
In your current solution, you always get the resources from upstream stable.
A better way IMO would be to have an option:

```
parser.add_argument(
    "--config-file",
   required=True,
   help="Path to the config file used to get list of clients providing resources. Default one is src/python/gem5_default_config.py")
```

Related question: why is this file a .py and not .json?

Harshil2107: The config file is .py because we wanted to build gem5 with the default config so we can import it during initialization in `client.py` and having a JSON file was causing issues with building with scons so we switched it to a .py.

I am working on the documentation to update config files, the current version can be found here: https://github.com/gem5/website/pull/90

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+
+if __name__ == "__main__" or __name__ == "__m5_main__":
+    parser = argparse.ArgumentParser(
+        description="Get resources from the database"
+    )
+
+    parser.add_argument(
+        "--config-file-url",
+        type=str,
+        default="https://raw.githubusercontent.com/gem5/gem5/stable/"
```
trollet91: I still think that this should be a path to a local file and not an URL.
We want this script to download the resources currently used in the user setup.

util/offline_db/get_resources_from_db.py
trollet91: For simplicity, I think it's fine to drop the md5 check untill we have the compressed md5 in the json.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+    req = request.Request(
+        auth_url, data=data, headers={"content-type": "application/json"}
+    )
+
+    try:
+        response = request.urlopen(req)
+    except Exception as e:
+        raise Exception(f"Failed to get token: {e}")
+
+    token = json.loads(response.read().decode("utf-8"))["access_token"]
```
BobbyRBruce: Not all that important, if it's easy, it'd be good if we could improve the error handling here with the following.

```suggestion
    json_response = resource.read().decode("utf-8")
    assert "access_token" in json_response, ("'access_token' not key in" 
                                             "json dictionary. JSON: \n"
                                             f"{json_respose}")
    token = json_response["access_token"]
```

This make it a little easier to know if something went wrong at the read/decode step or there json output doesn't have the right key and, if the latter, helpfully prints the JSON it received via the assert.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+    data = json.dumps(data).encode("utf-8")
+
+    req = request.Request(
+        auth_url, data=data, headers={"content-type": "application/json"}
+    )
+
+    try:
+        response = request.urlopen(req)
+    except Exception as e:
+        raise Exception(f"Failed to get token: {e}")
```
BobbyRBruce: Nitpik:

```suggestion
        raise Exception("Failed to obtain token via url request. "
                        f"Exception thrown:\n{e}")
```

1. I prefer printing the exception received on a new line.
2. Reading this, at least in code makes me think the token request is going to proceed the `:`. `Failed to get token: APT_TOKEN_1`. This is probably just me, but my phrasing increases the clarity by a tiny amount..

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+        "Authorization": f"Bearer {token}",
+        "Content-Type": "application/json",
+    }
+
+    req = request.Request(url, data=data, headers=headers)
+
+    try:
+        response = request.urlopen(req)
+    except Exception as e:
+        raise Exception(f"Failed to get resources: {e}")
```
BobbyRBruce:
```suggestion
        raise Exception("Failed to obtain resources via url request. "
                        f"Exception thrown:\n{e}")
```

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+
+    return json.loads(response.read().decode("utf-8"))["documents"]
```
BobbyRBruce: Same thing here with the slightly better error detection. (also not really all that important).

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+def save_resources_to_file(resources: List, output: pathlib.Path):
```
BobbyRBruce: In cases where it's possible to know can you state, via typing the type of objects the `List` type will consist of. In this case `List[Dict]`.

Also, in this case I think `pathlib.Path` as the type here is overly specific. I'd do `TextIO`. it can be imported `from typing import TextIO`. `TextIO` is a "any text IO", which is what `open` can handle in this case. `Path` of this type so it's fine.

Finally, i think it's a good habit to add a return type. In this case the return type is `None`.


```suggestion
def save_resources_to_file(resources: List[Dict], output: TextIO) -> None:
```

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+    :param output: Output directory absolute path
+    """
+
+    path = output.joinpath("resources.json")
+
+    with open(path, "w") as f:
+        json.dump(resources, f, indent=4)
+
+
+def progress_hook(t):
```
BobbyRBruce: Missing typing here but also missing documentation. really can't tell what what this function does.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+        if tsize not in (None, -1):
+            t.total = tsize
+        displayed = t.update((b - last_b[0]) * bsize)
+        last_b[0] = b
+        return displayed
+
+    return update_to
+
+
+def download_resources(output: pathlib.Path):
```
BobbyRBruce: Add a return typing here.

Also, let's change the import to `from pathlib import Path` and change all there instances from `pathlib.Path` to just `Path`. It's just neater.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+    updates the url field to point to the local download of the resource.
+    :param output: Output directory absolute path
+    """
+
+    path = output.joinpath("offline_resources")
+    if not path.exists():
+        path.mkdir()
+
+    with open(output.joinpath("resources.json")) as f:
+        resources = json.load(f)
```
BobbyRBruce: First of all, can we split this into two functions `download_resources` and `update_urls`? This feels to me to be two seperate tasks and decoupling them would make this code a little easier to follow.

In addition, can we take the reading and writing of the JSON files out. Can we just have the signature of this function be:

```py
def download_resources(resources: List[Dict], offline_resources_dir: Path) -> None:
```

it takes the resources input and download them all to the specified output directory. To make things easier when creating the URLs after this you can state in the docstring you download all resources to files/directories named "[id]-[version]".

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+
+    for resource in resources:
+        if "url" in resource.keys():
+            url = resource["url"]
+            filename = pathlib.Path(
+                *pathlib.Path(parse.urlsplit(url).path).parts[3:]
+            )
```
BobbyRBruce: This is really confusing, and if we end up keeping this I'd split it up into a line per function/step with comments explaining what each is doing. we're taking the URL, splitting it (?), getting the path, then casting it as a path, then getting the third part? ( din't know what part is and why "3" is significant). Is this just assuming our document is always 3-levels down a directory heirarchy? If so, please remember that the resource URL can be ANYTHING. It could be 65 directories down. There's no standard here.

util/offline_db/get_resources_from_db.py
BobbyRBruce: This doesn't make much sense that you're using "md5sum" here as a placeholder because we know that's wrong and to add the other "md5sum" we'll need to create another field.

Why don't you check to see if `url_md5sum` exists. if not, this check is skipped. The resource doesn't have a record of the md5sum as it prior to processing after a download. If `url_md5sum` does exist the resource is checked after downloading.

This doesn't break things and allows you to "add" the feature just by adding the data to the resource entires.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+            filepath = pathlib.Path(path).joinpath(filename)
+            filepath.parent.mkdir(parents=True, exist_ok=True)
+
+            if (
+                not filepath.exists()
+                or hashlib.md5(filepath.read_bytes()).hexdigest()
+                != resource["md5sum"]
+            ):
+                with tqdm(
```
BobbyRBruce: I know this isn't to be compiled into the gem5 binary, but I'd still prefer not having dependencies like this if we don't need this. This is purely aethetic.

Furthermore, _if_ you're going to require third-party libraries like this, ensure you include, at a minimum, a "requirement.txt" file and a README which explains the installation and usage.

util/offline_db/get_resources_from_db.py
BobbyRBruce: Agreed. 

So the script doesn't get bogged down with relative paths ( we should be able to move the script around and not have some hardcoded paths break), perhaps we can use a hardlink here so this file exists both in "util/offline_db" and "src/python/". Then we just have this default set to the directory the script is in plus "gem5_default_confg.py" (I don't actually know if git supports hardlinks ... you should check that before proceeding).

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+        default="https://raw.githubusercontent.com/gem5/gem5/stable/"
+        "src/python/gem5_default_config.py",
+        help="URL to the gem5 config file",
+    )
+
+    parser.add_argument(
+        "--output_dir",
+        type=str,
+        default=pathlib.Path.cwd(),
+        help="Output directory absolute path, default is the cwd."
```
BobbyRBruce: Nit: I'm pretty sure absolute paths are not a requirement here. Relative paths will be handled when the `Path` object is constructed.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+    config_file_suffix = pathlib.Path(
+        parse.urlsplit(args.config_file_url).path
+    ).suffix
+
+    # Get the gem5 config file from the url
+    if config_file_suffix == ".py":
+        gem5_config = request.urlopen(args.config_file_url)
+        gem5_config = gem5_config.read().decode("utf-8").split("=")[-1]
+        gem5_config = eval(gem5_config)
+        gem5_config = gem5_config["sources"]["gem5-resources"]
```
BobbyRBruce: So I can understand this can we improve this by having each line set a new variable, each of which describes it's value. E.g., first can be `request_returned`, the secon can be `decoded_request` etc. 

Also, please explain what `.split("="][-1]` is doing.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+    elif config_file_suffix == ".json":
+        gem5_config = request.urlopen(args.config_file_url)
+        gem5_config = json.loads(gem5_config.read().decode("utf-8"))
+        gem5_config = gem5_config["sources"]["gem5-resources"]
```
BobbyRBruce: Same here, but this is more understandable.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,238 @@
...
+    args = parser.parse_args()
+
+    output_path = pathlib.Path(args.output_dir)
+
+    config_file_suffix = pathlib.Path(
+        parse.urlsplit(args.config_file_url).path
+    ).suffix
+
+    # Get the gem5 config file from the url
+    if config_file_suffix == ".py":
```
BobbyRBruce: Let's not use file extensions to determine file formats. It's never a good idea.

Why don't you try to load it as json and if an exception is raised you quietly catch it and try to parse it as Python. If that fails you raise an exception explaining to the user the process and that it failed. Not the neatest thing ever but it doesn't make assumptions about the file extension.

BobbyRBruce (changes requested):
In additionL This is a tool so I'd expect a README.md included outlining what this is for and how to use it. I mentioned in one of my comments if this has dependencies that need installed, include a "requirements.txt" and explain in the readme how to install these dependencies.

util/offline_db/get_resources_from_db.py
Harshil2107: This function I copied from stdlib (https://github.com/gem5/gem5/blob/d0d3c74ce00f6d852a39060f3b7ce083887b1cfd/src/python/gem5/utils/progress_bar.py), this function just gives the download progress.

BobbyRBruce: I still think typing and documentation is necessary to help anyone maintaining this code to understand it. Out of context this function is hard to decipher.

util/offline_db/get_resources_from_db.py
Harshil2107: The default config python file follows the structure
```
gem5-config = {
    sources:{

    }
}
```
so I am just grabing the dictionary part of the file contents.

Harshil2107:
I could not find a way to get hard links to work with git. I just put the default config in the offline_db folder. For using the config files I switched it to python as the config file that we have in gem5 is a .py file.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,262 @@
...
+    # will be converted to  develop/checkpoints/...
+    filename = Path(
+        # This line combines the parts of the path
+        *Path(
+            # This line gets the path after https://
+            parse.urlsplit(resource_url).path
+        )
+        # This line splits the path into parts
+        # and gets the parts after the dist folder
+        .parts[3:]
```
BobbyRBruce: You're assuming all the URLs start with "https://storage.googleapi.com/dist.gem5.org/" This is not the case. We could have a URL "https://bobbybruce.net/assets/resource1". The URL to a resource can be literally anywhere.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,262 @@
...
+                or "url-md5sum" not in resource.keys()
+                or hashlib.md5(filepath.read_bytes()).hexdigest()
+                != resource["url-md5sum"]
+            ):
+                print(f"Downloading {url} to {filepath}")
+                request.urlretrieve(url, filepath)
+            resource["url"] = filepath.absolute().as_uri()
+
+
+if __name__ == "__main__" or __name__ == "__m5_main__":
```
BobbyRBruce: I assume this is just vanilla Python? We don't need gem5 to run this. Ergo:

```suggestion
if __name__ == "__main__":
```

should be fine.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,262 @@
...
+        if "url" in resource.keys():
+            url = resource["url"]
+            filepath = get_resource_local_filepath(url, path)
+
+            if (
+                not filepath.exists()
+                or "url-md5sum" not in resource.keys()
+                or hashlib.md5(filepath.read_bytes()).hexdigest()
+                != resource["url-md5sum"]
+            ):
```
BobbyRBruce: I think the logic here is slightly incorrect. In this logic the file will be redownloaded everytime if `url-md5sum` does not exist e. I think the following is what you want:

```suggestion
            if (
                not filepath.exists() # If the file does not exist
                or ("url-md5sum" in resource.keys() # If it exist but the md5sum is wrong.
                and hashlib.md5(filepath.read_bytes()).hexdigest()
                != resource["url-md5sum"])
            ):
```

BobbyRBruce (changes requested):
We still need a README.md for this to explain what this is for, how it's used, what the output should be.

util/offline_db/README.md
```diff
@@ -0,0 +1,22 @@
+
+# Offline Database Utility
+
+This Python script facilitates downloading all the resources from a database, storing them locally for offline usage. It operates by accessing an API, retrieving resources, and updating references within a JSON file to point to these local copies.
+
+### Running the Script
```
BobbyRBruce: I think this should be `##` (i.e., 2 `#`s, not 3).

util/offline_db/README.md
```diff
@@ -0,0 +1,22 @@
+
+# Offline Database Utility
+
+This Python script facilitates downloading all the resources from a database, storing them locally for offline usage. It operates by accessing an API, retrieving resources, and updating references within a JSON file to point to these local copies.
```
BobbyRBruce: There's no standardization to conform to here, so this is just my opinion (i.e., you don't really need to listen to me or do anything), but I do think in text documents where you don't have a clear concept of "lines" it's best to have system to split paragraphs up so they don't exist as one giant line.
GitHub wraps for us but not all tools will when reading the text file.

I personally advocate splitting paragraphs into a list of sentences.
So, in this case you'd get:



```suggestion
This Python script facilitates downloading all the resources from a database, storing them locally for offline usage.
It operates by accessing an API, retrieving resources, and updating references within a JSON file to point to these local copies.
```

(( where it really makes this cleaner is when you fix a typo or grammar mistake in a Markdown document.
The diff is more precise, showing a line changing rather than a whole paragraph  )).

util/offline_db/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Running the Script
+
+#### Arguments
+- `--config-file-path`: Filepath to the gem5 config file.
+- `--output_dir`: Output directory path (default: current working directory).
```
BobbyRBruce: Can we change this to `--output-dir`? I like command line arguments to use '-'s (I can't rationalize it, I just think they look better), and we're already using style for `--config-file-path` so let's be consistent.

util/offline_db/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+
+```bash
+python3 ./get_resources_from_db.py [--config-file-path CONFIG_FILE_PATH] [--output_dir OUTPUT_DIR]
```
BobbyRBruce: Again, another nit-pick about `-`s, but can we call this script `get-resource-from-db.py`? 🙏

In this case there is actually a small reason why we should call use `-`s for Python code we don't want to import.
You can't easily import files/directories into Python that have `-`s in their names, so having this file with `-` stops anyone misusing this script.
We don't want users to (easily) import this to something else and cause all sorts of bugs and complications.
This script is meant to be run with `python <script>` and having the file name containing `-`'s solidifies that.

util/offline_db/README.md
```diff
@@ -0,0 +1,22 @@
...
+```
+
+## Functionality
+
+### What does it do?
+- Retrieves resources from specified MongoDB Atlas database using an API.
```
BobbyRBruce:
```suggestion
- Retrieves resources from a specified MongoDB Atlas database using an API.
```

util/offline_db/gem5_default_config.py
```diff
@@ -0,0 +1,39 @@
...
+            "dataSource": "gem5-vision",
+            "database": "gem5-vision",
+            "collection": "resources",
+            "url": "https://data.mongodb-api.com/app/data-ejhjf/endpoint/data/v1",
+            "authUrl": "https://realm.mongodb.com/api/client/v2.0/app/data-ejhjf/auth/providers/api-key/login",
+            "apiKey": "OIi5bAP7xxIGK782t8ZoiD2BkBGEzMdX3upChf9zdCxHSnMoiTnjI22Yw5kOSgy9",
+            "isMongo": True,
+        }
+    }
+}
```
BobbyRBruce: I know you had an explanation here, but I really have forgot it, so please re-explain.

Why isn't this just a `.json` file?
The only reason this `gem5_default_config.py` exists as a Python file in "src" is because we can't add JSON files easily to the gem5 build.
Here we do not have this restriction.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,249 @@
...
+                print(f"Downloading {url} to {filepath}")
+                request.urlretrieve(url, filepath)
+            resource["url"] = filepath.absolute().as_uri()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Get resources from the database"
```
BobbyRBruce: This comment applies across a lot of the documentation/comments in this PR, but I'll bring it up here as this is a good example:

We are overloading the  "resource" and I fear it's going to confuse people. I've mentioned it before but we use it to mean 3 different parts of the infrastructure:

1. The entry in the database/JSON that we reference to find the disk images, binaries, etc. we need. (Small note here: from my understanding MongoDB uses the term "documents" to refer to what I think of as "entries").
2. The file that's obtained from the URL in the database entry (two small caveats here: (A) some of these are compressed or archived directories and therefore need further processing to be used in a simulation, is a resource just the downloaded file or the file after the decompressing and unarchiving operations?; and (B) some resources don't even have a file that needs downloaded. All the data they need is in the database entry. `SimpointResource` works like this. What would it mean to "download a SimpointResource", for example?
3. The `object` constructed from the `Resource` classes. I.e., the object returned by the `obtain_resource` function.



I'm open to alternative viewpoints here but I think a "gem5 Resource" makes most sense being the Python object. That's what's ultimately passed to the simulation configuration. This object is constructed from the database entry/document. The resource object may reference a "remote file" (or files -- we shouldn't limit ourself here) but this is just a design quirk. A resource is still a resource whether or not it a "remote file" it needs to download for the simulator.

How about:

* The gem5 Resource object: A Python object passed to the simulation to configure what simulated hardware runs, and how it is run.
* The gem5 Resource data-source entry: The enerty in the database (in most cases gem5's MongoDB database) which defines the object and is referenced by the Standard Library to construct a gem5 Resource object.
* A gem5 resource remote file: A remote file referenced by a gem5 Resource object which may be downloaded and used in a simulation. For example, the `x86-ubuntu-22.04-img` resource is dependent on the a remote disk image containing X86 Ubuntu 22.04 but this alone is not the "resource" as information on the Root partition is required by the simulator. It is the gem5 Resources object which encapsulates everything which is needed.

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,249 @@
...
+
+    output_path = Path(args.output_dir)
+
+    # Get the gem5 config file from the  file
+
+    with open(args.config_file_path) as f:
+        gem5_config_file_contents = f.read()
+    gem5_config_str = gem5_config_file_contents.split("=")[-1]
+    gem5_config_json = eval(gem5_config_str)
+    gem5_config_gem5_resources = gem5_config_json["sources"]["gem5-resources"]
```
BobbyRBruce: If we're using the Python file to contain this data. Why don't we just `from .gem5_default_config import config`?

((Though I am advocating for this to be a JSON file rather than a Python file)).

util/offline_db/get_resources_from_db.py
```diff
@@ -0,0 +1,249 @@
...
+    # Parse the gem5 config
+    db_url = gem5_config_gem5_resources["url"]
+    data_source = gem5_config_gem5_resources["dataSource"]
+    collection_name = gem5_config_gem5_resources["collection"]
+    db_name = gem5_config_gem5_resources["database"]
+    auth_url = gem5_config_gem5_resources["authUrl"]
+    api_key = gem5_config_gem5_resources["apiKey"]
+
+    if not output_path.exists():
+        output_path = output_path.mkdir()
```
BobbyRBruce: This is a bit dangerous. This will only create one directory the last directory in the path, and none of that directory's parents. So, if the value of `output_path` is "x/y/z" and only an empty directory "x" exists, `output_path.exits()` returns `False` but `output_oath.mkdir()` will raise an exception because "z" cannot be created without "y" first being created.

Then solution to this is to use `os.makedirs(<path>)` instead. `os.makedirs` recursively creates the directories.

BobbyRBruce (changes requested):

util/offline_db/README.md
powerjg: I strongly agree with Bobby's style suggestion. One sentence per line makes diffs much easier to review.

util/offline_db/get_resources_from_db.py
powerjg: I agree with all of this. I'll just add that, colloquially, when I say "we need a gem5 resource" or "get the resource" I am referring to the artifact that is used with gem5. E.g., a disk image is a resource. 

However, the documentation should be more precise, and I like how Bobby describes it here.

util/offline_db/gem5_default_config.py
Harshil2107: My assumption here was that the use will be using a gem5-config from a version of gem5 which will be python file.

util/offline_db/get_resources_from_db.py
Harshil2107: I can change this to `output_path.mkdir(parent=True, exist_ok= True)` and this will create all the parents.

Quoting from the pathlib documentation: "If parents is true, any missing parents of this path are created as needed; they are created with the default permissions without taking mode into account (mimicking the POSIX mkdir -p command)."

link: https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

util/offline_db/get_resources_from_db.py
Harshil2107: I am just doing this here as the user might point it to a different path.

util/offline_db/gem5_default_config.py
BobbyRBruce: Given we are providing this file alongside the tool, maybe we should just make it a JSON file? It keeps the tool simple with less conversion between Python and JSON.

util/offline_db/get_resources_from_db.py
BobbyRBruce: If we were to keep this as a Python file _I think_ there's a way to dynamically import the file.

Though this code is massively simplified if we just have the file in a JSON format.

util/offline_db/README.md
```diff
@@ -0,0 +1,23 @@
+
+# Offline Database Utility
+
+This Python script facilitates downloading all the resources from a database, storing them locally for offline usage.
+It operates by accessing an API, retrieving resources, and updating references within a JSON file to point to these local copies.
```
BobbyRBruce:
```suggestion
This Python script makes local copies of all the gem5 Resource's external files and data.
This is useful for cases where an internet connection is not permitted.
With this script users can cache a local copy which can be used by gem5.

Two interconnected parts of the gem5 infrastructure are cached using this script.
They are:

1. **The gem5 Resources database**: When the gem5 Standard Library attempts to contruct a gem5 Resource it will reference that resource's entry in the gem5 MongoDB database.
It then uses this data to construct the resource.
With this script we download this database as a JSON file for gem5 to use in-place of the database.
2. **gem5 Resource Remote Files**: Resources can (and often do) require files to be downloaded for them to function correctly in gem5.
Once the gem5 Resource is constructed from the database entry the gem5 Resource may download the remote files it requires.
As a simple example, a `DiskImageResource` will contain disk image partitioning information necessary for a gem5 simulation to be setup correctly. The disk image file is a remote file which is downloaded as this information cannot be stored directly in the database.
The location of these file is stored within the resource's database entry as a URL.
To create a local cache of these remote files we download all the remote files and, when creating the JSON file from the database, update the URLs to remote files to File URIs to their local equivalents.

Once this script is finished running the user will have a directory of all the remote gem5 Resources files and a JSON file representation of the database, with the remote files' URLs updated to local path URIs.
```

util/offline_db/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+```bash
+python3 ./get-resources-from-db.py [--config-file-path CONFIG_FILE_PATH] [--output_dir OUTPUT_DIR]
+```
+
+## Functionality
+
+### What does it do?
+- Retrieves resources a from specified MongoDB Atlas database using an API.
+- Downloads these resources to a local directory.
```
BobbyRBruce: Remote this and replace it with a short section explaining the config file. I think we should explain briefly, what these values mean and how a user may set them themselves.

util/offline_db/get-resources-from-db.py
```diff
@@ -0,0 +1,249 @@
...
+                != resource["url-md5sum"]
+            ):
+                print(f"Downloading {url} to {filepath}")
+                request.urlretrieve(url, filepath)
+            resource["url"] = filepath.absolute().as_uri()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Get gem5 resource objects from the database"
```
BobbyRBruce:
```suggestion
        description="Download all the remote gem5 Resources data and files "
                     "to be cached and used locally."
```

BobbyRBruce (changes requested):
Got some documentation suggestions here.

My only comment on top of this is please convert the config into a JSON file.

BobbyRBruce (approved):

PR #511, opened 2023-10-29
arch-riscv: Correct BootloaderKernelWorkload symbol table

hnpl:
Currently, the kernel's symbols are shifted by `kernel_paddr_offset`, which is where the kernel is located in the physcial address space. However, the symbols are mapped to virtual addresses, which stay the same even though the physical address space is shifted.

This patch removes the offset for the kernel's symbols virtual addresses.

Change-Id: I7c35f925777220f56bd8c69bba14c267d2048ade


rogerchang23424 (approved):
LGTM

powerjg (approved):

PR #512, opened 2023-10-29
base,sim: Add the SymbolType field to the Symbol object

hnpl:
Symbol type is part of the info provided by an ELF object's symtab.
It indicates whether a symbol is a file symbol, or a function symbol, etc.

This chain of commits introduces a way to only load function symbols
to the gem5's symbol table. The RISC-V BootloaderKernelWorkload now
loads only function symbols from the bootloader and the kernel binaries
by default.


Harshil2107 (approved):
I am not an expert on the ins and outs of this code, but the changes look good to me.

PR #514, opened 2023-10-29
mem-ruby, stdlib: Far atomics fix

### LINKED ISSUE #449: Arm timing 2-core CHI Boot test failing with `fatal: ... alloc_on_atomic without default or user set value`
**Describe the bug**

The [gem5 Daily tests are currently failing](https://github.com/gem5/gem5/actions/runs/6492500986/job/17634711866).

It is the `SuiteUID:tests/gem5/arm_boot_tests/test_linux_boot.py:timing-cpu_2-cores_chi_DualChannelDDR4_2400_arm_boot_test_m5-exit-ALL-x86_64-opt-CHI` Test suite is failing with `fatal: board.cache_hierarchy.core_clusters0.dcache.alloc_on_atomic without default or user set value`.

I suspect given the error is coming from the `cache_hierarchy` that this is a bug with CHI, but I have not investigated this issue beyond just reporting the error.

**Affects version**

At the time of writing it affects the `develop` branch. This report was based on revision [f7ad8fe4350be7d4299b22d9030385e3f261c7d8](https://github.com/gem5/gem5/commit/f7ad8fe4350be7d4299b22d9030385e3f261c7d8).

**To Reproduce**

The failing test can be run with:

```sh
cd tests
./main.py run --length=long --uid timing-cpu_2-cores_chi_DualChannelDDR4_2400_arm_boot_test_m5-exit-ALL-x86_64-opt-CHI -j`nproc`
```

The code this tests runs is the following:

```sh
# Build gem5 to the CHI Protocol
scons build/ALL_CHI/gem5.opt -j `nproc`

# Execute the following.
.build/ALL_CHI/gem5.opt tests/gem5/arm_boot_tests/configs/arm_boot_exit_run.py --cpu timing --num-cpus 2 --mem-system chi --dram-class DualChannelDDR4_2400 --resource-directory tests/gem5/resources
```

**Terminal Output**

```sh
fatal: board.cache_hierarchy.core_clusters0.dcache.alloc_on_atomic without default or user set value
```

**Host Operating System**

Ubuntu 22.04

**Host ISA**

X86

**Compiler used**

GCC

giactra:
This PR is fixing https://github.com/gem5/gem5/issues/449 by applying the following changes

1) Setting up alloc_on_atomic=False in the stdlib
This is directly related to the error message reported by the Issue #449

2) Disabling far atomics in stdlib with policy type = 0
There is an invalid transaction error, likely caused by the fact the current implementation
is expecting a 2 level cache hierarchy whereas the stdlib example only allocates one
level of caches (L1). This needs further investigation

3) Explicitly clearing the atomic log
Even by disabling far atomics, the execution of atomicPartial was populating
the atomic log queue without ever clearing it. This caused the OOM killer in Linux
to detect the leak and to kill it when the physical resources of the machine no longer
sufficed. IMHO the atomic log interface should be revamped as atomic users should
be allocating the atomic log only if explicitly needed

### DIFF
--- a/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
+++ b/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
@@ -2299,6 +2299,7 @@ action(UpdateDataState_FromADataResp, desc="") {
 
     tbe.oldDataBlk := tbe.dataBlk;
     tbe.dataBlk.atomicPartial(tbe.dataBlk, tbe.atomic_op);
+    tbe.dataBlk.clearAtomicLogEntries();
     tbe.dataDirty := true;
 
     DPRINTF(RubySlicc, "Atomic after %s\n", tbe.dataBlk);
@@ -3424,6 +3425,7 @@ action(Callback_AtomicHit, desc="") {
   DataBlock oldDataBlk;
   oldDataBlk := tbe.dataBlk;
   tbe.dataBlk.atomicPartial(tbe.dataBlk, tbe.atomic_op);
+  tbe.dataBlk.clearAtomicLogEntries();
 
   sequencer.atomicCallback(tbe.addr, oldDataBlk, false);
   DPRINTF(RubySlicc, "Atomic after %s\n", tbe.dataBlk);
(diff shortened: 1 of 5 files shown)


aarmejach (approved):
LGTM

Was just looking into the memory leak now :)

powerjg:
> There is an invalid transaction error, likely caused by the fact the current implementation
> is expecting a 2 level cache hierarchy whereas the stdlib example only allocates one
> level of caches (L1). This needs further investigation

Ah, this makes sense. We should definitely create a more realistic CHI-based `CacheHierarchy` for the standard library. It's important, but it hasn't quite made it to the top of the todo list, yet.

powerjg (approved):

PR #517, opened 2023-10-30
arch-riscv: Fixing CMO instructions and allowing using CMO instructions in FS mode

### LINKED ISSUE #516: RISC-V's `cbo_zero` instruction does not work correctly
**Describe the bug**
RISC-V's `cbo_zero` instruction does not work correctly in gem5.

**Affects version**
Current develop branch (https://github.com/gem5/gem5/commit/e9da8d67bdbb23fbd0578a379f08f42bce50121d)

**gem5 Modifications**
Add Zicbom and Zicboz extension to the RiscvBoard device tree.

**To Reproduce**
1. Clone current develop branch.
2. Add _Zicbom and _Zicboz to the RiscvBoard device tree.
3. Using riscv-fs.py and some modification to use Riscv's BootloaderKernelWorkload, run Full System simulation with Linux kernel 6.5.5, and with the command "m5 exit;" after booting.

**Terminal Output**
Guest terminal output,

```shell
[    0.126620] Run /root/gem5-init.sh as init process
[    0.147407] _swap_info_get: Bad swap file entry 440ffffffc704050
[    0.147425] BUG: Bad page map in process m5  pte:ffffffc7040508c0 pmd:4043b001
[    0.147447] addr:0000000000011000 vm_flags:00000075 anon_vma:0000000000000000 mapping:ffffffd8804931b8 index:1
[    0.147476] file:m5 fault:filemap_fault mmap:ext4_file_mmap read_folio:ext4_read_folio
[    0.147515] CPU: 0 PID: 40 Comm: m5 Not tainted 6.5.5 #1
[    0.147536] Hardware name: riscv-virtio (DT)
[    0.147550] Call Trace:
[    0.147561] [<ffffffff80005b1a>] dump_backtrace+0x1c/0x24
[    0.147585] [<ffffffff8088c570>] show_stack+0x2c/0x38
[    0.147608] [<ffffffff80897d86>] dump_stack_lvl+0x3c/0x54
[    0.147632] [<ffffffff80897db2>] dump_stack+0x14/0x1c
[    0.147656] [<ffffffff80152c68>] print_bad_pte+0x1cc/0x23a
[    0.147678] [<ffffffff8015480c>] unmap_page_range+0x344/0x9ee
[    0.147701] [<ffffffff80154f86>] unmap_vmas+0xd0/0x16e
[    0.147722] [<ffffffff8015fa38>] exit_mmap+0xd4/0x23c
[    0.147747] [<ffffffff8000c4ba>] mmput+0x56/0xf0
[    0.147768] [<ffffffff800140ac>] do_exit+0x1d0/0x758
[    0.147793] [<ffffffff800147a8>] do_group_exit+0x24/0x70
[    0.147820] [<ffffffff8001f23e>] get_signal+0x696/0x6e8
[    0.147846] [<ffffffff80004a90>] arch_do_signal_or_restart+0x5a/0x58a
[    0.147872] [<ffffffff8007b6d8>] exit_to_user_mode_prepare+0x6a/0x9c
[    0.147898] [<ffffffff80898a76>] syscall_exit_to_user_mode+0x24/0x38
[    0.147925] [<ffffffff80898734>] do_trap_ecall_u+0x4c/0x144
[    0.147951] [<ffffffff8000336c>] ret_from_exception+0x0/0x64 
```

**Expected behavior**

**Host Operating System**
Ubuntu 22.04

**Host ISA**
ARM

**Compiler used**
gcc-11

**Additional information**
Linux kernel uses `cbo_zero` to clear a page in a page table. Apparently the page was not zeroed out correctly, and it caused the wrong page table entry problem above.
Also, gem5 panicked when using `--debug-flags=ExecAll` while executing the `cbo_zero` instruction.

hnpl:
arch-riscv: Fix implementation of CMO extension instructions

This change introduces a template for store instruction's mem access.
The new template is called CacheBlockBasedStore.

The reasons for not reusing the current Store's mem access template
are as follows,
- The CMO extension instructions operate on cache block size granularity,
while regular load/store instructions operate on data of size 64 bits or
fewer.
- The writeMemAtomicLE/writeMemTimingLE interfaces do not allow passing
nullptr as data. However, CPUs in gem5 rely on (data == NULL) to detect
CACHE_BLOCK_ZERO instructions. Setting `Mem = 0;` to `uint64_t Mem;`
does not solve the problem as the reference is allocated and thus,
it's always true that `&Mem != NULL`. This change uses the
writeMemAtomic/writeMemTiming interfaces instead.
- Per CMO v1.0.1, the instructions in the spec do not generate
address misaligned faults.
- The CMO extension instructions do not use IMM.

---

arch-riscv: Fix generateDisassembly for Store with 1 source reg

Currently, store instructions are assumed to have two source registers.
However, since we are supporting the RISC-V CMO instructions, which
are Store instructions in gem5 but they only have one source register.
This change allows printing disassembly of Store instructions with
one source register.

---

arch-riscv: Make Zicbom/Zicboz extensions optional in FS mode

Currently, we're enable Zicbom/Zicboz by default. Since those
extensions might be buggy as they are not well-tested, making
those entensions optional allows running simulation where
the performance implication of the instructions do not matter.

Effectively, by turning off the extensions, we simply remove
those extensions from the device tree, so the OS would not
use them. It doesn't prohibit the userspace application to
use those instructions, however.

---

arch-riscv: Add all supporting Z extensions to RISC-V isa string

### DIFF
--- a/src/arch/riscv/RiscvISA.py
+++ b/src/arch/riscv/RiscvISA.py
@@ -96,6 +96,9 @@ class RiscvISA(BaseISA):
         ELEN in Ch. 2 of RISC-V vector spec",
     )
 
+    enable_Zicbom_fs = Param.Bool(True, "Enable Zicbom extension in FS mode")
+    enable_Zicboz_fs = Param.Bool(True, "Enable Zicboz extension in FS mode")
+
     def get_isa_string(self):
         isa_extensions = []
         # check for the base ISA type
@@ -110,8 +113,10 @@ class RiscvISA(BaseISA):
             isa_extensions.append("v")
         isa_string = "".join(isa_extensions)
 
-        isa_string += "_Zicbom"  # Cache-block Management Instructions
-        isa_string += "_Zicboz"  # Cache-block Zero Instruction
+        if self.enable_Zicbom_fs.value:
+            isa_string += "_Zicbom"  # Cache-block Management Instructions
+        if self.enable_Zicboz_fs.value:
+            isa_string += "_Zicboz"  # Cache-block Zero Instruction
         isa_string += "_Zicntr"  # Performance Couter Spec
         isa_string += "_Zicsr"  # RMW CSR Instructions (Privileged Spec)
         isa_string += "_Zifencei"  # FENCE.I Instruction (Unprivileged Spec)
--- a/src/arch/riscv/RiscvISA.py
+++ b/src/arch/riscv/RiscvISA.py
@@ -109,4 +109,15 @@ class RiscvISA(BaseISA):
         if self.enable_rvv.value == True:
             isa_extensions.append("v")
         isa_string = "".join(isa_extensions)
+
+        isa_string += "_Zicbom"  # Cache-block Management Instructions
+        isa_string += "_Zicboz"  # Cache-block Zero Instruction
+        isa_string += "_Zicntr"  # Performance Couter Spec
+        isa_string += "_Zicsr"  # RMW CSR Instructions (Privileged Spec)
+        isa_string += "_Zifencei"  # FENCE.I Instruction (Unprivileged Spec)
+        isa_string += "_Zihpm"  # Performance Couter Spec
+        isa_string += "_Zba"  # Address Generation
+        isa_string += "_Zbb"  # Basic Bit Manipulation
+        isa_string += "_Zbs"  # Single-bit Instructions
+
         return isa_string
--- a/src/arch/riscv/insts/mem.cc
+++ b/src/arch/riscv/insts/mem.cc
@@ -55,8 +55,13 @@ std::string
 Store::generateDisassembly(Addr pc, const loader::SymbolTable *symtab) const
 {
     std::stringstream ss;
-    ss << mnemonic << ' ' << registerName(srcRegIdx(1)) << ", " <<
-        offset << '(' << registerName(srcRegIdx(0)) << ')';
+    if (_numSrcRegs == 1) {
+        ss << mnemonic << ' ' << offset << '(' << registerName(srcRegIdx(0))
+           << ")";
+    } else {
+        ss << mnemonic << ' ' << registerName(srcRegIdx(1)) << ", " <<
+            offset << '(' << registerName(srcRegIdx(0)) << ')';
+    }
     return ss.str();
 }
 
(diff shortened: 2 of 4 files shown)


hnpl:
Fix #516 

Should be merged after #464

src/arch/riscv/RiscvISA.py
```diff
@@ -94,3 +94,37 @@ class RiscvISA(BaseISA):
...
+            isa_extensions.append("v")
+
+        isa_string = "".join(isa_extensions)
+
+        if self.enable_Zicbom_fs.value:
+            isa_string += "_Zicbom"  # Cache-block Management Instructions
+        if self.enable_Zicboz_fs.value:
+            isa_string += "_Zicboz"  # Cache-block Zero Instruction
+        isa_string += "_Zicbom"  # Cache-block Management Instructions
+        isa_string += "_Zicboz"  # Cache-block Zero Instruction
```
rogerchang23424: Repeat isa_string of `zicboz` and `zicbom`

hnpl: Fixed!

src/arch/riscv/insts/mem.cc
```diff
@@ -55,8 +55,12 @@ std::string
...
 {
     std::stringstream ss;
-    ss << mnemonic << ' ' << registerName(srcRegIdx(1)) << ", " <<
-        offset << '(' << registerName(srcRegIdx(0)) << ')';
+    if (_numSrcRegs == 1)
+        ss << mnemonic << ' ' << offset << '(' << registerName(srcRegIdx(0))
+           << ")";
+    else
+        ss << mnemonic << ' ' << registerName(srcRegIdx(1)) << ", " <<
+            offset << '(' << registerName(srcRegIdx(0)) << ')';
```
powerjg:
```suggestion
    if (_numSrcRegs == 1) {
        ss << mnemonic << ' ' << offset << '(' << registerName(srcRegIdx(0))
           << ")";
    } else {
        ss << mnemonic << ' ' << registerName(srcRegIdx(1)) << ", " <<
            offset << '(' << registerName(srcRegIdx(0)) << ')';
   }
```

Usually it's just a polite request to use `{` around single-line if statements, but this one is required, IMO. Having multiple lines without `{` is *incredibly* likely to generate bugs in the future.

powerjg (dismissed):

src/arch/riscv/insts/mem.cc
hnpl: Done

rogerchang23424 (approved):

BobbyRBruce (approved):

hnpl:
Exquisite

PR #518, opened 2023-10-30
gpu-compute: Fix for failing assertion in GPUCoalescer

### LINKED ISSUE #508: m_atomicLog not getting cleared
**Describe the bug**
An assertion in [GPUCoalescer.cc ](https://github.com/gem5/gem5/blob/b6ce2d0db891a2dee0bace10222dec06cb4c175b/src/mem/ruby/system/GPUCoalescer.cc#L635) fails with the below debug log (also attached):
```
152221825000: system.ruby.tcp_cntrl0.coalescer: Got hitCallback for 0x1025F00
152221825000: system.ruby.tcp_cntrl0.coalescer: Responding to 4 packets for addr 0x1025F00; type: ATOMIC_NO_RETURN
152221825000: system.ruby.tcp_cntrl0.coalescer: numAtomicLogEntries() = 4  (This is the failing assertion variable)
```
Maybe related to updates from this commit: https://github.com/gem5/gem5/commit/f6a453362fbc330bda69d187ac3ffed991bee1ef

This is not seen with the latest master/stable version of gem5

**Affects version**
Develop branch: 6ddf8c94ee1368f8a573fc7f782dcf6a8534293b

**gem5 Modifications**
None

**To Reproduce**
1. Compile gem5 with command `scons build/GCN3_X86/gem5.opt`
2. Download [hsti.zip](https://github.com/gem5/gem5/files/13171331/hsti.zip)
3. Run with: `docker run -u $UID:$GID --volume $(pwd):$(pwd) -w $(pwd) gcr.io/gem5-test/gcn-gpu:v22-1  gem5/build/GCN3_X86/gem5.opt -d chai/hsti --debug-file=debug --debug-flags=GPUCoalescer --debug-start=152000000000 gem5/configs/example/apu_se.py --cpu-type=DerivO3CPU --num-cpus=12 -c gem5-resources/src/gpu/chai/HIP-U-gem5/HSTI/bin/hsti.gem5 1>chai/hsti/simout 2>chai/hsti/simerr`

**Terminal Output**
[logs.zip](https://github.com/gem5/gem5/files/13172103/logs.zip)

**Host Operating System**
Ubuntu 20.04, Docker (same OS)

**Additional Information**
Seems like adding a pop for every ATOMIC_NO_RETURN pkt [here](https://github.com/gem5/gem5/blob/f6a453362fbc330bda69d187ac3ffed991bee1ef/src/mem/ruby/system/GPUCoalescer.cc#L573) similar to the case of ATOMIC_RETURN solves the issue for the benchmark I'm running, but perhaps the cleaner solution would be to not record m_atomicLog entry when processing ATOMIC_NO_RETURN.

mysoreanoop:
Commit #200 introduces colaesced atomics to same addr from within a WF; atomicLog was part of it to record the changes on datablock sequentially; However, for non-returning atomics, this log isn't cleared, which manifests as a failing assertion.

Fixes issue #508, adds to PR #397 

Change-Id: Ib2aa470120163552488c3e69913d87b70b293c6e

### DIFF
--- a/src/mem/ruby/system/GPUCoalescer.cc
+++ b/src/mem/ruby/system/GPUCoalescer.cc
@@ -599,6 +599,8 @@ GPUCoalescer::hitCallback(CoalescedRequest* crequest,
                 // data response is not needed.
                 case RubyRequestType_ATOMIC_NO_RETURN:
                     assert(pkt->isAtomicOp());
+                    data.popAtomicLogEntryFront();
+                    break;
                 case RubyRequestType_ST:
                     data.setData(pkt->getPtr<uint8_t>(), offset, pkt_size);
                     break;
--- a/src/mem/ruby/system/GPUCoalescer.cc
+++ b/src/mem/ruby/system/GPUCoalescer.cc
@@ -599,7 +599,8 @@ GPUCoalescer::hitCallback(CoalescedRequest* crequest,
                 // data response is not needed.
                 case RubyRequestType_ATOMIC_NO_RETURN:
                     assert(pkt->isAtomicOp());
-                    data.popAtomicLogEntryFront();
+                    log = data.popAtomicLogEntryFront();
+                    delete [] log;
                     data.setData(pkt->getPtr<uint8_t>(), offset, pkt_size);
                     break;
                 case RubyRequestType_ST:
--- a/src/mem/ruby/system/GPUCoalescer.cc
+++ b/src/mem/ruby/system/GPUCoalescer.cc
@@ -600,6 +600,7 @@ GPUCoalescer::hitCallback(CoalescedRequest* crequest,
                 case RubyRequestType_ATOMIC_NO_RETURN:
                     assert(pkt->isAtomicOp());
                     data.popAtomicLogEntryFront();
+                    data.setData(pkt->getPtr<uint8_t>(), offset, pkt_size);
                     break;
                 case RubyRequestType_ST:
                     data.setData(pkt->getPtr<uint8_t>(), offset, pkt_size);


src/mem/ruby/system/GPUCoalescer.cc
```diff
@@ -599,6 +599,8 @@ GPUCoalescer::hitCallback(CoalescedRequest* crequest,
                 // data response is not needed.
                 case RubyRequestType_ATOMIC_NO_RETURN:
                     assert(pkt->isAtomicOp());
+                    data.popAtomicLogEntryFront();
+                    break;
```
mysoreanoop, line 605: PS: added a break because I think there is no alternative fall through case when handling atomic packets -- correct me if I'm wrong.

mysoreanoop: Ignore that, _ST case applies still, didn't notice the comment

mysoreanoop:
Due to the fallthrough case in the change, with my compiler, I get the implicit-fallthrough warning as an error -- which can be fixed by modifying the SConstruct file like so:
```
diff --git a/SConstruct b/SConstruct
index e7fa4b53b7..47bed3014b 100755
--- a/SConstruct
+++ b/SConstruct
@@ -452,6 +452,7 @@ for variant_path in variant_paths:
         env.Append(CCFLAGS=['-Werror',
                              '-Wno-error=deprecated-declarations',
                              '-Wno-error=deprecated',
+                             '-Wno-error=implicit-fallthrough',
                             ])
```
Should I add this change as well, or replicate the statements within the cases?

src/mem/ruby/system/GPUCoalescer.cc
mattsinc: It does look like there needs to be a break here, thanks for catching.

mattsinc (dismissed):
This seems ok, but would appreciate @abmerop checking.

mattsinc:
@dankouch just making sure you see this.

powerjg:
There should only be 1 commit. Whoever merges should use *squash* merge!

powerjg:
I assume, but I want to double check... This is *unrelated* to the memory leak discussed in #514 (https://github.com/gem5/gem5/pull/514#issue-1967109979). Please let me know one way or the other.

powerjg (changes requested):
I think this is also related to #449. Should the same fix be applied to `Sequencer.cc`?

src/mem/ruby/system/GPUCoalescer.cc
```diff
@@ -599,6 +599,9 @@ GPUCoalescer::hitCallback(CoalescedRequest* crequest,
                 // data response is not needed.
                 case RubyRequestType_ATOMIC_NO_RETURN:
                     assert(pkt->isAtomicOp());
+                    data.popAtomicLogEntryFront();
```
abmerop: We need to explicitly free the memory here as it does in the ATOMIC_RETURN case.  Something like this would prevent a memory leak:

`log = data.popAtomicLogEntryFront();
delete [] log;`

abmerop (changes requested):
We should fix a memory leak while modifying this code.  Thanks for the contribution!

DanKouch:
Would this conflict with CHI now clearing the atomic log within the protocol [here](https://github.com/gem5/gem5/pull/514/files#diff-ef6b9d16329f3bdb4467703b887640a7bd8d3436311de27e0c5e341ecd6a8353R2302) in #514?

mattsinc:
> I think this is also related to #449. Should the same fix be applied to `Sequencer.cc`?

Since this issue is specifically related to GPU atomics, there could be a similar memory leak in the CPU, but it would have to be for something other than GPU atomics.  Where do you want us to look in the Sequencer?

mattsinc:
> Would this conflict with CHI now clearing the atomic log within the protocol [here](https://github.com/gem5/gem5/pull/514/files#diff-ef6b9d16329f3bdb4467703b887640a7bd8d3436311de27e0c5e341ecd6a8353R2302) in #514?

In theory since this commit is about GPU atomics and CHI does not support GPUs, it should be similar concept but unrelated codewise?

abmerop:
> I assume, but I want to double check... This is _unrelated_ to the memory leak discussed in #514 ([#514 (comment)](https://github.com/gem5/gem5/pull/514#issue-1967109979)). Please let me know one way or the other.

I believe it is the same memory leak, so it would have to be handled for other atomic no return requests (I guess in Sequencer?)

Maybe another fix would be to have an additional boolean parameter for DataBlock and WriteMask atomic methods which indicates whether or not to populate the log at all and default to false. The protocols would have to explicitly set to true for atomics with return though which may be a lot of changes.  I don't know how many protocols actual use atomics though.

src/mem/ruby/system/GPUCoalescer.cc
```diff
@@ -599,6 +599,10 @@ GPUCoalescer::hitCallback(CoalescedRequest* crequest,
                 // data response is not needed.
                 case RubyRequestType_ATOMIC_NO_RETURN:
                     assert(pkt->isAtomicOp());
+                    log = data.popAtomicLogEntryFront();
+                    delete [] log;
```
abmerop, line 603: Sorry to suggest yet another change, especially since I suggest this.  However, I noticed that clearAtomicLogEntries() will delete *all* of the entries and clear the log.  The code I suggested assumes that the log only has one entry which I think is not necessarily true:

`data.clearAtomicLogEntries()`

mysoreanoop: This is already within a for-loop -- wouldn't that take care of all the log entries?

abmerop: Oh I see.  The relation between DataBlock and packet was a bit unclear until I read the comment for atomic return.  This look fine to me then

abmerop (approved):

mattsinc (approved):
@mysoreanoop: this looks good to me, but just making sure you saw @powerjg 's request that you also update Sequencer.cc?  I believe that is blocking this patch.

mysoreanoop:
@powerjg -- I'm not sure what's needed in Sequencer.cc
https://github.com/gem5/gem5/blob/8b63299ad61f6ea5b3ae50cc830e632b17d8430e/src/mem/ruby/system/Sequencer.cc#L749 -- this is where we could clear the data->m_atomicLog -- but I'm not sure. Do you have any insight on this?

powerjg:
@abmerop said the following.

> I believe it is the same memory leak, so it would have to be handled for other atomic no return requests (I guess in Sequencer?)

If there is a memory leak that is in both Sequencer and in GPUSequencer, it should be fixed in both. 

I am not an expert on this code. @bujset contributed the original atomic log code. If this introduced a memory leak, we need to fix it. #200 

Also, it would be great if you could answer this specific question: Is the memory leak resolved in the change the same as the memory leak mentioned in #514. If you need more information about #514, then talk to Giacomo in that PR.

mattsinc:
> @abmerop said the following.
> 
> > I believe it is the same memory leak, so it would have to be handled for other atomic no return requests (I guess in Sequencer?)
> 
> If there is a memory leak that is in both Sequencer and in GPUSequencer, it should be fixed in both.
> 
> I am not an expert on this code. @BujSet contributed the original atomic log code. If this introduced a memory leak, we need to fix it. #200
> 
> Also, it would be great if you could answer this specific question: Is the memory leak resolved in the change the same as the memory leak mentioned in #514. If you need more information about #514, then talk to Giacomo in that PR.

This specific problem will not be fixed by #514 because 514 only updates the CHI protocol, which is not used in GPUs.  So there is a separate fix needed here.

I am not sure where exactly @abmerop is envisioning a fix being done in Sequencer.cc as the atomic log -- I assume it's related to https://github.com/gem5/gem5/blob/8b63299ad61f6ea5b3ae50cc830e632b17d8430e/src/mem/ruby/system/Sequencer.cc#L721?

BujSet:
When I wrote the changes to GPUCoalescer, I had noted my assumption that Store and AtomicNoReturn requests follow the same path. However, looking through the VIPER protocol files, it seems that AtomicNoReturn and AtomicReturn requests are treated the same (so logs are created and populated even on AtomicNoReturn requests). And I'm guessing this assumption doesn't hold in other protocols as well. 

I think the correct fix will require changes to the cache protocols to treat AtomicReturns and AtomicNoReturns differently, rather than changing the GPUCoalescer. As discussed in #508, the RubyRequestType already has the information needed to differentiate these, so I think you would just need to have separate action in the protocols themselves. I think not create the logs if you don't need them is a cleaner solution than creating logs in both cases, and deleting them when not needed.

abmerop:
My suggestion was to modify the WriteMask atomic method to add an additional parameter whether or not the log should be used, with false as the default.  The log only seems to be necessary for GPU, so we would set the parameter to true in VIPER. Then there are no changes to CHI, Sequencer, nor GPUCoalescer to fix the leak

mysoreanoop:
Seems like at the cache closest to the core (TCP), the accepted request has a RubyRequestType, and when it passes on to the next cache (TCC), it switches to CoherenceRequestType which does not distinguish between RETURN and NO_RETURN. However, interestingly, it has 2 types: Atomic (used ubiquitously) and AtomicWriteBack -- which is unused. We could rename this for those atomics needing a return (=log), else don't use log. How does that sound?

abmerop:
The GLC bit determines if it is return or no return and I believe that is already propagated, so we could use that

mysoreanoop:
Alright then, I can add in the changes to `performAtomic` and related functions.

src/mem/ruby/system/GPUCoalescer.cc
```diff
@@ -599,6 +599,10 @@ GPUCoalescer::hitCallback(CoalescedRequest* crequest,
...
+                    data.setData(pkt->getPtr<uint8_t>(), offset, pkt_size);
```
mysoreanoop, line 604: @BujSet is this necessary for ATOMIC_NO_RETURN?

BujSet: I think this would solve the assertion error if you need results quickly, however, this still only addresses the problem after the fact. I started a patch that will fix this issue by not creating the log when it's not needed (i.e. when the request is an AtomicNoReturn). I'm still working on this though since it touches quite a few different files (key issue is that the VIPER messages reuse the CPU message structure which ignores the type of atomic in the incoming RubyRequest, and in the event that the request is a miss the atomic action happens on the response path, so the delineation of what type of atomic request is issued must be preserved throughout the entire trip of the request). Will provide an update when I have a working solution.

mysoreanoop: Got it, I'll wait for your update and take down this PR then. 
I can also help make those changes if you need extra hands; let me know

BujSet: See my PR here #546. @abmerop, When I was looking through the protocol code for the directories, I noticed that only the SLC bit is set and read. Rather than implementing checks for the GLC bit, I just added some extra types to distinguish the atomic request type. Do you think this would be fine, or would it be preferable to add in the support for the GLC bit set and checks?

abmerop: Since you've already done the work, I think it is cleaner not to pass the GLC bit beyond TCC as it isn't needed otherwise.  I thought it was already passed up hence my earlier comment

mysoreanoop: Cool, I will close this PR for now. @BujSet I meant to ask if line 604 above is necessary for the NO_RETURN case; if not, you could add a break statement.

mysoreanoop:
Closing in favor of https://github.com/gem5/gem5/pull/546.

src/mem/ruby/system/GPUCoalescer.cc
BujSet: Oh. I'm not sure. In line with the comment, I just followed the same format as what happens for a store. But thinking about more, maybe setData doesn't need to be called for either store or atomic no return requests. Will look into this.

PR #521, opened 2023-10-31
python: Handle unicode characters in config files

caizixian:
Previously, opening a config file (such as `configs/example/hmc_hello.py`) containing non-ASCII characters causes UnicodeDecodeError.
Also switch to use more an more idiomatic context manager for handling files.

Change-Id: Ia39cbe2c420e9c94f3a84af459b7e5f4d9718d14

### DIFF
--- a/src/python/m5/main.py
+++ b/src/python/m5/main.py
@@ -632,7 +632,9 @@ def main():
             if not options.P:
                 sys.path = [os.path.dirname(sys.argv[0])] + sys.path
             filename = sys.argv[0]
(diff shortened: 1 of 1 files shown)


powerjg (approved):
Thanks for the fix!

caizixian:
The test results look fine. The only failed run is due to DNS resolution.
```
warn: `Resource` has been deprecated. Please use the `obtain_resource` function instead.
URLError: <urlopen error [Errno -3] Temporary failure in name resolution>
```

PR #526, opened 2023-11-03

mattsinc:
Print was not properly ending in a newline, which caused confusion when looking a trace with GPUTLB enabled.  This fixes that.


mattsinc:
Apparently I am still getting used to Github and additional commits that should be separate were added to this thread.  So closing this and will reopen shortly.

PR #527, opened 2023-11-03

mattsinc:
Print was not properly ending in a newline, which caused confusion when looking a trace with GPUTLB enabled. This fixes that.


mattsinc:
It appears my change over also lost my ChangeID info, and updating the commits adds the annoying merge branch commits on top of each branch, so closing one more time to resolve this and opening separately.

PR #528, opened 2023-11-03
mem-ruby, gpu-compute: update GPU L1I$ MRU info

mattsinc:
Previously the GPU L1 I$ (SQC) was not updating the MRU information on hits in the SQC.  This commit resolves that by adding support to the appropriate Ruby transition.


mattsinc:
It appears my change over also lost my ChangeID info, and updating the commits adds the annoying merge branch commits on top of each branch, so closing one more time to resolve this and opening separately.

PR #529, opened 2023-11-03

mattsinc:

Print was not properly ending in a newline, which caused confusion when looking a trace with GPUTLB enabled. This fixes that.

Change-Id: I1e0d93e2f555bfbeb7c02b281bdce4ffe5398e31

### DIFF
--- a/src/arch/amdgpu/common/tlb_coalescer.cc
+++ b/src/arch/amdgpu/common/tlb_coalescer.cc
@@ -482,7 +482,7 @@ TLBCoalescer::processProbeTLBEvent()
                     stats.localqueuingCycles += (curTick() * pkt_cnt);
                 }
 
-                DPRINTF(GPUTLB, "Successfully sent TLB request for page %#x",
+                DPRINTF(GPUTLB, "Successfully sent TLB request for page %#x\n",
                        virt_page_addr);
 
                 //copy coalescedReq to issuedTranslationsTable


abmerop (approved):

mattsinc:
@powerjg before I rerun, just checking if there is something I'm missing here -- when I click on details for the failed CI tests, I just see "This job failed" -- it seems unlikely adding a newline character would cause a failure on these tests, but nonetheless wanted to check before investing tester resources.

abmerop:
The actions don't seem to be running at all for the past 3 days except for some stuff on stable which is odd since it hasn't been updated in ~4 months?

mattsinc:
> The actions don't seem to be running at all for the past 3 days except for some stuff on stable which is odd since it hasn't been updated in ~4 months?

I know @powerjg and I have been working with the IT folks at Madison to resolve some issues with the runners.  I had assumed those issues meant the tests just wouldn't run though (as opposed to failing like in this case), but perhaps I'm wrong.

powerjg:
The tests will fail if they are queued for more than 24 hours without any runners picking them up.

The daily/weekly/compiler tests officially run on "stable", but they actually check out develop before running anything.

mattsinc:
> The tests will fail if they are queued for more than 24 hours without any runners picking them up.
> 
> The daily/weekly/compiler tests officially run on "stable", but they actually check out develop before running anything.

Thanks @powerjg !  How do we restart/kickoff the tests?

powerjg:
It looks like it's queued. Hopefully it will run soon

mattsinc:
> It looks like it's queued. Hopefully it will run soon

This is another one that is purely formatting/cosmetic and shouldn't affect anything, if @BobbyRBruce wants to merge it and save test bandwidth.  But I'll let it up to him to decide.

Issue #474, opened 2023-10-17
Generalize support for exit events


powerjg:
Note: This issue is a brain dump and will likely be updated.

## Problem

Today, the exit events in gem5 are not very flexible. Most of the information that moves from the C++ event queue to Python is implemented via strings (e.g., "user interrupt received"). Some other places for improvements

- `SIGUSR1` is hardcoded to dump stats
- `SIGUSR2` is hardcoded to dump and reset stats
- `SIGINT` does not always actually exit the simulation

The other main problem we are trying to solve is enabling users to extend the exit event types.
For instance, exiting for SimPoints, LoopPoints, or checking on the current state of the simulation.

## Some steps

- [ ] Fix SIGINT to work when executing on the non-main event queue
  - #476 
- [ ] Refactor `SIGUSR1/2` to exit to python and then execute the dump/reset stats in Python
- [ ] Create a different class for user-defined exit events
- [ ] Also allow for return python-implemented exit events
- [ ] Implement a new signal (maybe using `SIGRT`?) so that Python can send a signal to stop the simulation

andysan:
Looking at the code, it's pretty clear that there are race conditions in the way asynchronous events are notified. I think we might need to rewrite that code to use a flag field based on an std::atomic_uint32_t to ensure we can atomically read and clear multiple flags.

andysan:
I have been thinking a bit about this and replacing the strings with proper objects would be a big improvement. Ideally, I would like to see something looking a bit like exceptions on the Python side. We currently have two big problems. It's very easy to lose exit events by incorrectly ignoring the return value of `m5.simulate()`. The other problem is drain where we need to run the simulator, but we actually ignore exit events. Treating them as exceptions would be cleaner.

I think there is a path that maintains backwards compatibility here. We could implement `__str__` in the exit event wrapper to maintain backwards compatibility. Exit event handling could be handled internally by a global callback (that can be overridden as a keyword argument to `simulate`?)  that either returns the event, returns None (ignore the event), or throws the event as an exception. We could provide two implementation, one that always throws the event and one that always returns the event. The default would be to return the event to maintain backwards compatibility.

We should probably avoid using a signal to stop the simulator from Python, at least SIGRT which is used by KVM. A better option would probably be to use a socket (which could use a SIGIO in gem5) which would give us a bit more flexibility to control the simulator from the outside.

trollet91:
> The other problem is drain where we need to run the simulator, but we actually ignore exit events

[...]

PR #531, opened 2023-11-03
sim,python: Restore sigint handler in python

powerjg:
Currently, if you try to use ctrl-c while python code is running nothing
happens. This is not ideal. This change enables users to use ctrl-c
while python is running (e.g., when a large disk image is downloading).
To do this, we moved the `initSignals` function in gem5 from `main` to
the simulate loop. Thus, every time the simulate loop starts (i.e., is
called from python) gem5 will install its signal handlers. Also, when
the control is returned to python, we put python's default SIGINT
handler back.

Change-Id: I14490e48d931eb316e8c641217bf8d8ddaa340ed

### DIFF
--- a/src/sim/init_signals.cc
+++ b/src/sim/init_signals.cc
@@ -84,7 +84,8 @@ setupAltStack()
 
 static void
 installSignalHandler(int signal, void (*handler)(int sigtype),
-                     int flags = SA_RESTART)
+                     int flags = SA_RESTART,
+                     struct sigaction *old_sa = NULL)
 {
     struct sigaction sa;
 
@@ -93,7 +94,7 @@ installSignalHandler(int signal, void (*handler)(int sigtype),
     sa.sa_handler = handler;
     sa.sa_flags = flags;
 
-    if (sigaction(signal, &sa, NULL) == -1)
+    if (sigaction(signal, &sa, old_sa) == -1)
         panic("Failed to setup handler for signal %i\n", signal);
 }
 
@@ -196,9 +197,6 @@ initSignals()
     // Dump intermediate stats and reset them
     installSignalHandler(SIGUSR2, dumprstStatsHandler);
 
-    // Exit cleanly on Interrupt (Ctrl-C)
-    installSignalHandler(SIGINT, exitNowHandler);
-
     // Print the current cycle number and a backtrace on abort. Make
     // sure the signal is unmasked and the handler reset when a signal
     // is delivered to be able to invoke the default handler.
@@ -217,5 +215,20 @@ initSignals()
     // PollQueue class.
     installSignalHandler(SIGIO, ioHandler);
 }
+
+struct sigaction old_int_sa;
+
+void initSigInt()
+{
+    // Exit cleanly on Interrupt (Ctrl-C)
+    installSignalHandler(SIGINT, exitNowHandler, SA_RESTART, &old_int_sa);
+}
+
+void restoreSigInt()
+{
+    // Restore the old SIGINT handler
+    sigaction(SIGINT, &old_int_sa, NULL);
+}
+
 
 } // namespace gem5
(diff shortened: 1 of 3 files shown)


powerjg:
@andysan here's another attempt at the signal handler problem. Let me know what you think.

andysan:
At a high-level, I think this sort of makes sense. It is slightly weird that you can't interrupt the Python code when you are not executing the event loop. From an implementation perspective, I think it would be better to save the old SIGINT handler using `sigaction` when you install gem5's SIGINT handler in simulate, and then restore it when returning. That avoids the fixup code in Python and makes the behaviour consistent if built without Python. That should make the code easier to follow since the intent is clearer.

If you want to make this really Python friendly, we should consider throwing a `KeyboardInterrupt` from `m5.simulate` if the simulator was interrupted. This would definitely break some scripts, so we need to be careful if we do this.

powerjg:
> I think it would be better to save the old SIGINT handler using sigaction when you install gem5's SIGINT handler in simulate, and then restore it when returning. 

Great idea... I don't know why I didn't do it that way...

> If you want to make this really Python friendly, we should consider throwing a KeyboardInterrupt from m5.simulate if the simulator was interrupted. This would definitely break some scripts, so we need to be careful if we do this.

I think this is a bigger change than what we want to do right now. Personally, I think that this should be tied to a better interface for m5 exit events. This is on our short-term to do list, so hopefully we'll tackle this soon. See #474

powerjg:
Thanks for the feedback, Andreas. I've implemented this as you suggested. I've tested it by hitting "ctrl-c" while gem5 is running the sim loop and during the python execution before and after the sim loop. That now works as expected. I have not tested other signals, but I don't have any reason they should behave differently.

andysan:
Thanks, this looks a lot better! However, I'm a bit worried that restoring the other signal handlers might lead to weird behaviour. For example, if you restore SIGUSRx, you could end up in a situation where a stat dump isn't triggered if you happen to be servicing an event in Python. That seems wrong to me. Similarly, we probably want to print a backtrace if a call from Python triggers an abort or segfault in the C++ code. For the problem you're solving, I think it would be sufficient to just save and restore SIGINT.

BobbyRBruce:
@powerjg : Are you wanting this to get in for v23.1? If so, do you need help moving this forward?

powerjg:
There's little chance I'll have time until after the ASPLOS deadline. I looked at Andreas's suggestion and it's not totally straightforward.

andysan:
> There's little chance I'll have time until after the ASPLOS deadline. I looked at Andreas's suggestion and it's not totally straightforward.

As far as I can tell, you'd get the desired behaviour if you just save and restore SIGINT. It seems to me like the other signal handlers can be left in place when executing Python code, That way, we can still intercept SIGUSRx for stat dumps, and SIGSEGV and SIGABRT for backtraces.

powerjg:
I understand your suggestion, but I don't have time to implement it right now, unfortunately. I appreciate the review, but often times things that looks like a 5 minute job to the reviewer take significantly more effort.

The problem is that now we need two separate functions to initialize the signals with your suggestion. It's not hard, but it's not a 5 minute job. I estimate 30-60 minutes + testing. I just don't have the time right now with deadlines, etc.

So, it looks like this is going to languish unless someone else picks it up.

powerjg:
Ok, I've tried a different approach. Let me know if this is what you were thinking.

That said, I don't like that we're treating SIGINT uniquely. This code could easily get out of hand.

andysan:
> Ok, I've tried a different approach. Let me know if this is what you were thinking.
> 
> That said, I don't like that we're treating SIGINT uniquely. This code could easily get out of hand.

I think this this is fine. The other signals need to be handled inside gem5 itself, SIGINT is the only one that needs to hand over control to the simulation script.

andysan (approved):

PR #532, opened 2023-11-03
stdlib, resources: Update JSON data in workload

### LINKED ISSUE #192: Workloads lacks specifying constituent Resource version: suggest change to gem5 Resources JSON schema.
The current workloads lack the specification of resource versions for the resources they contain. This issue arises when attempting to retrieve a specified resource from the workload. In this scenario, the system automatically resorts to using the latest resource version that is compatible with the installed gem5 version. However, this behavior is not the intended one. Ideally, a workload should explicitly define the precise version of the resource it requires.

This situation can also lead to complications when multiple data sources contain resources required by a workload. Instead of adhering to the resource version specified in the workload, the "obtain resource" function retrieves the latest compatible version from all available data sources. This behavior diverges from the intended outcome.

The current workload structure is as follows:

```
{
    "category": "workload",
    "id": "x86-matrix-multiply-omp-100-8-looppoint-region-7",
    "description": "A workload which loads a Looppoint JSON file for the 'x86-matrix-multiply-omp' binary with inputs '100' and '8' and runs region 7 via a checkpoint.",
    "function": "set_se_looppoint_workload",
    "resources": {
        "binary": "x86-matrix-multiply-omp",
        "looppoint": "x86-matrix-multiply-omp-100-8-looppoint",
        "checkpoint": "x86-matrix-multiply-omp-100-8-looppoint-checkpoint-region-7"
    },
    "additional_params": {
        "region_id": "7"
    },
    "architecture": "X86",
    "size": 0,
    "tags": [],
    "code_examples": [],
    "license": "",
    "author": [],
    "source_url": "",
    "resource_version": "1.0.0",
    "gem5_versions": [
        "23.0"
    ],
    "example_usage": "Workload(\"x86-matrix-multiply-omp-100-8-looppoint-region-7\")"
}
```

The new change would be to change the `resources` field to `obtain_resources`. The `obtain_resources` field would be a dictionary that has a category of resource for example `"binary"` as key and a dictionary containing resource id and version as value.

The above resource will look as follows with the new changes:
```
{
    "category": "workload",
    "id": "x86-matrix-multiply-omp-100-8-looppoint-region-7",
    "description": "A workload which loads a Looppoint JSON file for the 'x86-matrix-multiply-omp' binary with inputs '100' and '8' and runs region 7 via a checkpoint.",
    "function": "set_se_looppoint_workload",
    "obtain_resources": {
        "binary": {
             "id": "x86-matrix-multiply-omp",
             "resource_version": "1.0.0"
         }
        "looppoint":{
             "id": "x86-matrix-multiply-omp-100-8-looppoint",
             "resource_version": "1.0.0"
         }
        "checkpoint": {
             "id": "x86-matrix-multiply-omp-100-8-looppoint-checkpoint-region-7",
             "resource_version": "1.0.0"
         }
    },
    "additional_params": {
        "region_id": "7"
    },
    "architecture": "X86",
    "size": 0,
    "tags": [],
    "code_examples": [],
    "license": "",
    "author": [],
    "source_url": "",
    "resource_version": "2.0.0",
    "gem5_versions": [
        "23.1"
    ],
    "example_usage": "Workload(\"x86-matrix-multiply-omp-100-8-looppoint-region-7\")"
}
```
This change will add a new version for all resources and this new version will be compatible with gem5 version 23.1 and onwards.

Incorporating this new change would necessitate an update to the resources schema. As modifications accumulate within resources, maintaining a universally applicable schema becomes progressively intricate. To address this challenge, a pragmatic approach could involve maintaining distinct schemas tailored to different gem5 versions. Subsequently, a validation script could be developed, leveraging the `gem5_versions` attribute of each resource to ensure alignment with the appropriate schema. This approach would contribute to a more streamlined and manageable system as gem5 evolves.

Harshil2107:
- resources field in workload now supports a dict with resources id and version.

- Older workload JSON are still supported but added a deprecation waring

Change-Id: I137dbb99799a5294e84ce7d5d914f05e4cfe9e00

### DIFF
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -964,10 +964,22 @@ def obtain_resource(
             for key in resource_json["resources"].keys():
                 assert isinstance(key, str)
                 value = resource_json["resources"][key]
-                assert isinstance(value, str)
-                params[key] = obtain_resource(
-                    value,
-                )
+
+                if isinstance(value, str):
+                    warn(
+                        "Deprecation warning: resources field in workloads has changed"
+                        "from { category: id } to"
+                        "{ category: { id: id, resource_version: resource_version } }"
+                        "Please update your resource.json file to reflect this change."
+                    )
+                    params[key] = obtain_resource(
+                        value,
+                    )
+                elif isinstance(value, dict):
+                    params[key] = obtain_resource(
+                        value["id"],
+                        resource_version=value["resource_version"],
+                    )
         if "additional_params" in resource_json:
             for key in resource_json["additional_params"].keys():
                 assert isinstance(key, str)
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -970,7 +970,8 @@ def obtain_resource(
                         "Deprecation warning: resources field in workloads has changed"
                         "from { category: id } to"
                         "{ category: { id: id, resource_version: resource_version } }"
-                        "Please update your resource.json file to reflect this change."
+                        "The current develop branch of gem5 supports both formats"
+                        "but this will be removed in the 23.1 release."
                     )
                     params[key] = obtain_resource(
                         value,


src/python/gem5/resources/resource.py
```diff
@@ -964,10 +964,22 @@ def obtain_resource(
...
-                params[key] = obtain_resource(
-                    value,
-                )
+
+                if isinstance(value, str):
+                    warn(
+                        "Deprecation warning: resources field in workloads has changed"
+                        "from { category: id } to"
+                        "{ category: { id: id, resource_version: resource_version } }"
+                        "Please update your resource.json file to reflect this change."
```
powerjg: This could be printed if someone is using the database, right? How would that person update their `resources.json`?

Harshil2107: Yes, should I change it to "Please update the database" or "Please update the workload JSON objects"?

powerjg: You should give a warning that will be helpful to the user. If you said "Please update the database" what could the user possibly do about this? How would they interpret that message?

I don't think you should tell the user to do anything if they cannot take any actions.

*Possibly* if the user is using a local version of a database (e.g., a specialized json file), *only then* should you suggest a change, since only then they could actually take an action.

BobbyRBruce (commented):
At some point we had a very nice scheme to change our schema for workloads to give use more flexbility. Is that in the pipeline?

I'm really uncomfortable with all this hard-coding this creeping into our design. This new schema is just as inflexible as the previous one (which is why we needed to update it in the first place) For example, what if we want to pass another parameter to "obtain_resource" here? We'd have to change it again.

I remember at one point we had a design on the whiteboard that looked something like this:

```json
{
    "function": "set_kernal_disK_workload",
    "parameters" : {
         "kernel": {
              "function": "obtain_resource",
              "parameters" : {
                   "id" : "some_kernel",
                   "version" : "1"
              }
          }
          "disk_image": {
              "function": "obtain_resource",
              "parameters" : {
                   "id" : "an_image",
                   "version" : "2"
              }
          },
          "readfile_contents" : "echo 'hello world'"
     }
}
```

Which i like a lot better

src/python/gem5/resources/resource.py
BobbyRBruce: I don't think any warning should be given. This patch is to keep this version of gem5 compatible with both versions the old and the newer version the schema. The warning should be to us that we need to do something about this and keep track of it.

I would hope the schema versioning functionality is implemented soon so we can more easily keep track of these changes in a logical way.

Harshil2107: The if condition change is to keep develop working with both versions of workload, I think we will remove it before we release 23.1 as when we release a new version the older workload will work with 23.0 and the new one will work with 23.1.

BobbyRBruce: > The if condition change is to keep develop working with both versions of workload, I think we will remove it before we release 23.1 as when we release a new version the older workload will work with 23.0 and the new one will work with 23.1.

I would like that but we are running low on time.

BobbyRBruce (approved):

PR #534, opened 2023-11-05
Fix calculation of compressed size in bytes

odanrc:
An integer division in the compression:Base:getSize() was being done, which led to rounding down instead of up.

### DIFF
--- a/src/mem/cache/compressors/base.cc
+++ b/src/mem/cache/compressors/base.cc
@@ -78,7 +78,7 @@ Base::CompressionData::getSizeBits() const
 std::size_t
 Base::CompressionData::getSize() const
 {
-    return std::ceil(_size/8);
+    return std::ceil(_size/(float)CHAR_BIT);
 }
 
 Base::Base(const Params &p)


odanrc:
I just wanted to point out that it was incredibly much more difficult to make my first contribution under GitHub than under Gerrit (when I had little git experience). The process was much more cumbersome. There should probably be a troubleshooting section here (https://www.gem5.org/contributing). Two things I had to fix while making this contribution:

1 - Forked without the branches (through GitHub´s UI), which I fixed with:

git remote add gem5 https://github.com/gem5/gem5.git
git fetch gem5

2 - Had this message pop up when pushing: ¨remote: Support for password authentication was removed on August 13, 2021.¨ (yeah, I hadn´t used this computer to code in a while). This was fixed by generating a token in https://github.com/settings/tokens and then running

git remote set-url origin https://<token>@github.com/<username>/gem5.git

powerjg (approved):
thanks for the feedback. We'll put that on the website!

PR #540, opened 2023-11-06
mem-ruby, gpu-compute: fix SQC/TCP requests to same line

mattsinc:
Currently, the GPU SQC (L1I$) and TCP (L1D$) have a performance bug
where they do not behave correctly when multiple requests to the same
cache line overlap one another.  The intended behavior is that if the
first request that arrives at the Ruby code for the SQC/TCP misses, it
should send a request to the GPU TCC (L2$).  If any requests to the
same cache line occur while this first request is pending, they should
wait locally at the L1 in the MSHRs (TBEs) until the first request has
returned.  At that point they can be serviced, and assuming the line
has not been evicted, they should hit.

For example, in the following test (on 1 GPU thread, in 1 WG):

load Arr[0]
load Arr[1]
load Arr[2]

The expected behavior (confirmed via profiling on real GPUs) is that
we should get 1 miss (Arr[0]) and 2 hits (Arr[1], Arr[2]) for such a
program.

However, the current support in the VIPER SQC/TCP code does not model
this correctly.  Instead it lets all 3 concurrent requests go straight
through to the TCC instead of stopping the Arr[1] and Arr[2] requests
locally while Arr[0] is serviced.  This causes all 3 requests to be
classified as misses.

To resolve this, this patch adds support into the SQC/TCP code to
prevent subsequent, concurrent requests to a pending cache line from being
sent in parallel with the original one.  To do this, we add an
additional transient state (IV) to indicate that a load is pending to
this cache line.  If a subsequent request of any kind to the same cache
line occurs while this load is pending, the requests are put on the
local wait buffer and woken up when the first request returns to the
SQC/TCP.  Likewise, when the first load is returned to the SQC/TCP, it
transitions from IV --> V.

As part of this support, additional transitions were also added to
account for corner cases such as what happens when the line is evicted
by another request that maps to the same set index while the first load
is pending (the line is immediately given to the new request, and when
the load returns it completes, wakes up any pending requests to the same
line, but does not attempt to change the state of the line) and how GPU
bypassing loads and stores should interact with the pending requests
(they are forced to wait if they reach the L1 after the pending,
non-bypassing load; but if they reach the L1 before the non-bypassing
load then they make sure not to change the state of the line from IV if
they return before the non-bypassing load).

As part of this change, we also move the MSHR behavior from internally
in the GPUCoalescer for loads to the Ruby code (like all other
requests).  This is important to get correct hits and misses in stats
and other prints, since the GPUCoalescer MSHR behavior assumes all
requests serviced out of its MSHR also miss if the original request to
that line missed.

Although the SQC does not support stores, the TCP does.  Thus,
we could have applied a similar change to the GPU stores at the TCP.
However, since the TCP support assumes write-through caches and does not
attempt to allocate space in the TCP, we elected not to add this support
since it seems to run contrary to the intended behavior (i.e., the
intended behavior seems to be that writes just bypass the TCP and thus
should not need to wait for another write to the same cache line to
complete).

Additionally, making these changes introduced issues with deadlocks at
the TCC.  Specifically, some Pannotia applications have accesses to the
same cache line where some of the accesses are GLC (i.e., they bypass
the GPU L1 cache) and others are non-GLC (i.e., they want to be cached
in the GPU L1 cache). We have support already per CU in the above code.
However, the problem here is that these requests are coming from
different CUs and happening concurrently (seemingly because different
WGs are at different points in the kernel around the same time).
This causes a problem because our support at the TCC for the TBEs
overwrites the information about the GPU bypassing bits (SLC, GLC) every
time. The problem is when the second (non-GLC) load reaches the TCC, it
overwrites the SLC/GLC information for the first (GLC) load. Thus, when
the the first load returns from the directory/memory, it no longer has
the GLC bit set, which causes an assert failure at the TCP.

After talking with other developers, it was decided the best way handle
this and attempt to model real hardware more closely was to move the
point at which requests are put to sleep on the wakeup buffer from the
TCC to the directory. Accordingly, this patch includes support for that
-- now when multiple loads (bypassing or non-bypassing) from different
CUs reach the directory, all but the first one will be forced to wait
there until the first one completes, then will be woken up and
performed.  This required updating the WTRequestor information at the
TCC to pass the information about what CU performed the original request
for loads as well (otherwise since the TBE can be updated by multiple
pending loads, we can't tell where to send the final result to).  Thus,
I changed the field to be named CURequestor instead of WTRequestor since
it is now used for more than stores.  Moreover, I also updated the
directory to take this new field and the GLC information from incoming
TCC requests and then pass that information back to the TCC on the
response -- without doing this, because the TBE can be updated by
multiple pending, concurrent requests we cannot determine if this memory
request was a bypassing or non-bypassing request.  Finally, these
changes introduced a lot of additional contention and protocol stalls at
the directory, so this patch converted all directory uses of z_stall to
instead put requests on the wakeup buffer (and wake them up when the
current request completes) instead. Without this, protocol stalls cause
many applications to deadlock at the directory.

However, this exposed another issue at the TCC: other applications
(e.g., HACC) have a mix of atomics and non-atomics to the same cache
line in the same kernel.  Since the TCC transitions to the A state when
an atomic arrives. For example, after the first pending load returns to
the TCC from the directory, which causes the TCC state to become V, but
when there are still other pending loads at the TCC. This causes invalid
transition errors at the TCC when those pending loads return, because
the A state thinks they are atomics and decrements the pending atomic
count (plus the loads are never sent to the TCP as returning loads).
This patch fixes this by changing the TCC TBEs to model the number of
pending requests, and not allowing atomics to be issued from the TCC
until all prior, pending non-atomic requests have returned.

Change-Id: I37f8bda9f8277f2355bca5ef3610f6b63ce93563

### DIFF
--- a/src/mem/ruby/protocol/GPU_VIPER-SQC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-SQC.sm
@@ -48,6 +48,9 @@ machine(MachineType:SQC, "GPU SQC (L1 I Cache)")
 {
   state_declaration(State, desc="SQC Cache States", default="SQC_State_I") {
     I, AccessPermission:Invalid, desc="Invalid";
+    // Note: currently IV in the TCP is only for pending loads to a given cache
+    // line.  Since the SQC is read only, there are no stores.
+    IV, AccessPermission:Invalid, desc="Going from I to V, waiting on TCC data";
     V, AccessPermission:Read_Only, desc="Valid";
   }
 
@@ -98,6 +101,7 @@ machine(MachineType:SQC, "GPU SQC (L1 I Cache)")
   void unset_tbe();
   void wakeUpAllBuffers();
   void wakeUpBuffers(Addr a);
+  void wakeUpAllBuffers(Addr a);
   Cycles curCycle();
 
   // Internal functions
@@ -270,12 +274,31 @@ machine(MachineType:SQC, "GPU SQC (L1 I Cache)")
     }
   }
 
+  action(t_allocateTBE, "t", desc="allocate TBE Entry") {
+    check_allocate(TBEs);
+    TBEs.allocate(address);
+    set_tbe(TBEs.lookup(address));
+  }
+
+  action(d_deallocateTBE, "d", desc="Deallocate TBE") {
+    TBEs.deallocate(address);
+    unset_tbe();
+  }
+
+  action(st_stallAndWaitRequest, "st", desc="Stall and wait on the address") {
+    stall_and_wait(mandatoryQueue_in, address);
+  }
+
   action(p_popMandatoryQueue, "pm", desc="Pop Mandatory Queue") {
     mandatoryQueue_in.dequeue(clockEdge());
   }
 
   action(pr_popResponseQueue, "pr", desc="Pop Response Queue") {
     responseToSQC_in.dequeue(clockEdge());
+  }
+
+  action(wada_wakeUpAllDependentsAddr, "wada", desc="Wake up any requests waiting for this address") {
+    wakeUpAllBuffers(address);
   }
 
   action(l_loadDoneHit, "ldh", desc="local load done (hits in SQC)") {
@@ -313,22 +336,52 @@ machine(MachineType:SQC, "GPU SQC (L1 I Cache)")
 
   // Transitions
 
+  // if another request arrives for the same cache line that has a pending
+  // load, put it on the wakeup buffer.  This reduced resource contention since
+  // they won't try again every cycle and will instead only try again once woken
+  // up
+  transition(IV, {Fetch}) {
+      st_stallAndWaitRequest;
+  }
+
   // transitions from base
-  transition({I, V}, Repl, I) {TagArrayRead, TagArrayWrite} {
+  transition({I, IV, V}, Repl, I) {TagArrayRead, TagArrayWrite} {
     // since we're evicting something, don't bother classifying as hit/miss
     ic_invCache;
   }
 
-  transition(I, Data, V) {TagArrayRead, TagArrayWrite, DataArrayRead} {
+  // if we got a response for a load where the line is in I, then
+  // another request must have come in that replaced the line in question in
+  // the cache.  Thus, complete this request without allocating the line, but
+  // still deallocate TBE and wakeup any dependent addresses.
+  transition(I, Data) {TagArrayRead, TagArrayWrite, DataArrayRead} {
+    // don't profile this as a hit/miss since it's a reponse from L2,
+    // so we already counted it
+    l_loadDoneMiss;
+    wada_wakeUpAllDependentsAddr;
+    d_deallocateTBE;
+    pr_popResponseQueue;
+  }
+
+  // if line is currently in IV, then Data is returning the data for a
+  // pending load, so transition to V, deallocate TBE, and wakeup any dependent
+  // requests so they will be replayed now that this request has returned.
+  transition(IV, Data, V) {TagArrayRead, TagArrayWrite, DataArrayRead} {
     a_allocate;
     // don't profile this as a hit/miss since it's a reponse from L2,
     // so we already counted it
     w_writeCache;
     l_loadDoneMiss;
+    wada_wakeUpAllDependentsAddr;
+    d_deallocateTBE;
     pr_popResponseQueue;
   }
 
-  transition(I, Fetch) {TagArrayRead, TagArrayWrite} {
+  // if we have a load that misses, allocate TBE entry and transition to IV
+  // to prevent subsequent requests to same cache line from also going to TCC
+  // while this request is pending
+  transition(I, Fetch, IV) {TagArrayRead, TagArrayWrite} {
+    t_allocateTBE;
     nS_issueRdBlkS;
     uu_profileDataMiss; // since line wasn't in SQC, we missed
     p_popMandatoryQueue;
--- a/src/mem/ruby/protocol/GPU_VIPER-TCP.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCP.sm
@@ -292,10 +292,13 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
         TBE tbe := TBEs.lookup(in_msg.LineAddress);
         DPRINTF(RubySlicc, "%s\n", in_msg);
         if (in_msg.Type == RubyRequestType:LD) {
-          if ((in_msg.isGLCSet || in_msg.isSLCSet) && is_valid(cache_entry)) {
-            // Read requests with GLC or SLC bit set should not cache in the L1.
-            // They need to bypass the L1 and go to the L2. If an entry exists
-            // in the L1, it needs to be evicted
+          // Read requests with GLC or SLC bit set should not cache in the L1.
+          // They need to bypass the L1 and go to the L2.  If an entry exists in
+          // the L1, it needs to be evicted, and if no entry or invalid entry in
+          // the L1, still need to bypass.  The LoadBypassEvict Event handles
+          // both cases in its transitions below, so call LoadBypassEvict for
+          // both.
+          if ((in_msg.isGLCSet || in_msg.isSLCSet)) {
             trigger(Event:LoadBypassEvict, in_msg.LineAddress, cache_entry, tbe);
           }
           else {
--- a/src/mem/ruby/protocol/GPU_VIPER-TCP.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCP.sm
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2011-2015 Advanced Micro Devices, Inc.
+ * Copyright (c) 2023 Matthew D. Sinclair
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -53,10 +54,14 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
 {
   state_declaration(State, desc="TCP Cache States", default="TCP_State_I") {
     I, AccessPermission:Invalid,   desc="Invalid";
+    // Note: currently IV in the TCP is only for pending loads to a given cache
+    // line. Since the TCP is write through, stores should be allowed to pass
+    // through without requiring them to wait.
+    IV, AccessPermission:Invalid,  desc="Going from I to V, waiting on TCC data";
     V, AccessPermission:Read_Only, desc="Valid";
     A, AccessPermission:Invalid,   desc="Waiting on Atomic";
 
-    F, AccessPermission:Invalid, desc="Flushing; Waiting for Ack";
+    F, AccessPermission:Invalid,   desc="Flushing; Waiting for Ack";
   }
 
   enumeration(Event, desc="TCP Events") {
@@ -102,6 +107,8 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
     bool Dirty,        desc="Is the data dirty (different than memory)?";
     int NumPendingMsgs,desc="Number of acks/data messages that this processor is waiting for";
     bool Shared,       desc="Victim hit by shared probe";
+    bool isGLCSet,     desc="Bypass L1 Cache";
+    bool isSLCSet,     desc="Bypass L1 and L2 Cache";
    }
 
   structure(TBETable, external="yes") {
@@ -123,6 +130,7 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
   void unset_tbe();
   void wakeUpAllBuffers();
   void wakeUpBuffers(Addr a);
+  void wakeUpAllBuffers(Addr a);
   Cycles curCycle();
 
   // Internal functions
@@ -472,6 +480,15 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
     check_allocate(TBEs);
     TBEs.allocate(address);
     set_tbe(TBEs.lookup(address));
+
+    // pass GLC/SLC information along
+    if (mandatoryQueue_in.isReady(clockEdge())) {
+      peek(mandatoryQueue_in, RubyRequest) {
+        DPRINTF(RubySlicc, "Address: %p\n", address);
+        tbe.isGLCSet := in_msg.isGLCSet;
+        tbe.isSLCSet := in_msg.isSLCSet;
+      }
+    }
   }
 
   action(d_deallocateTBE, "d", desc="Deallocate TBE") {
@@ -510,6 +527,10 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
     responseToTCP_in.dequeue(clockEdge());
   }
 
+  action(st_stallAndWaitRequest, "st", desc="Stall and wait on the address") {
+    stall_and_wait(mandatoryQueue_in, address);
+  }
+
   action(l_loadDoneHit, "ldh", desc="local load done (hits in TCP)") {
     assert(is_valid(cache_entry));
     if (use_seq_not_coal) {
@@ -525,6 +546,20 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
       sequencer.readCallback(address, cache_entry.DataBlk, false, MachineType:L1Cache);
     } else {
       coalescer.readCallback(address, MachineType:L1Cache, cache_entry.DataBlk);
+    }
+  }
+
+  action(ldmi_loadDoneMissInv, "ldmi",
+         desc="local load done (misses in TCP and line was evicted)") {
+    // since line was evicted, can't rely on data from cache entry, so use from
+    // the response message
+    peek(responseToTCP_in, ResponseMsg) {
+      DataBlock tmp:= in_msg.DataBlk;
+      if (use_seq_not_coal) {
+        sequencer.readCallback(address, tmp, false, MachineType:L1Cache);
+      } else {
+        coalescer.readCallback(address, MachineType:L1Cache, tmp);
+      }
     }
   }
 
@@ -604,6 +639,10 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
     L1cache.setMRU(address);
   }
 
+  action(wada_wakeUpAllDependentsAddr, "wada", desc="Wake up any requests waiting for this address") {
+    wakeUpAllBuffers(address);
+  }
+
 //  action(zz_recycleMandatoryQueue, "\z", desc="recycle mandatory queue") {
 //    mandatoryQueue_in.recycle(clockEdge(), cyclesToTicks(recycle_latency));
 //  }
@@ -632,11 +671,19 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
   // Stalling transitions do NOT check the tag array...and if they do,
   // they can cause a resource stall deadlock!
 
-  transition({A}, {Load, Atomic, StoreThrough}) { //TagArrayRead} {
-      z_stall;
+  // if another request arrives for the same cache line that has a pending
+  // atomic or load, put it on the wakeup buffer instead of z_stall'ing it.  By
+  // doing so we reduce resource contention since they won't try again every cycle
+  // and will instead only try again once woken up
+  transition({A, IV}, {Load, LoadBypassEvict, Atomic, Store, StoreThrough, Flush}) {
+      st_stallAndWaitRequest;
   }
 
-  transition(I, Load) {TagArrayRead} {
+  // if we have a load that misses, allocate TBE entry and transition to IV
+  // to prevent subsequent requests to same cache line from also going to TCC
+  // while this request is pending
+  transition(I, Load, IV) {TagArrayRead} {
+    t_allocateTBE;
     n_issueRdBlk;
     uu_profileDataMiss;
     p_popMandatoryQueue;
@@ -694,14 +741,38 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
     p_popMandatoryQueue;
   }
 
-  transition(I, TCC_Ack, V) {TagArrayRead, TagArrayWrite, DataArrayRead, DataArrayWrite} {
-    a_allocate;
-    w_writeCache;
-    l_loadDoneMiss;
+  // if we got a response for a load where the line is in I, then
+  // another request must have come in that replaced the line in question in
+  // the cache.  Thus, complete this request without allocating the line, but
+  // still deallocate TBE and wakeup any dependent addresses.
+  // (Note: this assumes TCC_AckWB is what stores use)
+  transition(I, TCC_Ack) {TagArrayRead, TagArrayWrite} {
+    wada_wakeUpAllDependentsAddr;
+    // NOTE: Because we invalidated the cache line, the assert in l_loadDoneMiss
+    // will fail -- unlike atomics that automatically go to I when the line returns
+    // loads do not automatically go to I.  Resolve this by passing data from
+    // message.
+    ldmi_loadDoneMissInv;
+    d_deallocateTBE;
     pr_popResponseQueue;
   }
 
-  transition(I, Bypass, I) {
+  // if line is currently in IV, then TCC_Ack is returning the data for a
+  // pending load, so transition to V, deallocate TBE, and wakeup any dependent
+  // requests so they will be replayed now that this request has returned.
+  transition(IV, TCC_Ack, V) {TagArrayRead, TagArrayWrite, DataArrayRead, DataArrayWrite} {
+    a_allocate;
+    w_writeCache;
+    wada_wakeUpAllDependentsAddr;
+    l_loadDoneMiss;
+    d_deallocateTBE;
+    pr_popResponseQueue;
+  }
+
+  // if a bypass request arrives back at the TCP, regardless of whether the line
+  // is in I (from the bypass request) or IV (from a subsequent non-bypassing
+  // load), retain the current state and complete the bypassing request.
+  transition({I, IV}, Bypass) {
     rb_bypassDone;
     pr_popResponseQueue;
   }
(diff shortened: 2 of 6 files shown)


powerjg:
Can you rebase this change to have only your changes? Let me or @Harshil2107 know if you have any problems or need any help.

src/mem/ruby/protocol/GPU_VIPER-SQC.sm
```diff
@@ -308,22 +331,52 @@ machine(MachineType:SQC, "GPU SQC (L1 I Cache)")
...
+  // if line is currently in IV, then Data is returning the data for a
+  // pending load, so transition to V, deallocate TBE, and wakeup any dependent
+  // requests so they will be replayed now that this request has returned.
+  transition(IV, Data, V) {TagArrayRead, TagArrayWrite, DataArrayRead} {
     a_allocate;
     // don't profile this as a hit/miss since it's a reponse from L2,
     // so we already counted it
     w_writeCache;
     l_loadDoneMiss;
+    wada_wakeUpAllDependentsAddr;
```
powerjg, line 375: Do you need to call `wakeupAllBuffers()` (without the address parameter)? I believe that if you have a conflict miss, the miss will wait on the "conflict" address (the probe address), but you will wake up on the original address.

powerjg (commented):
One small thing below.

src/mem/ruby/protocol/GPU_VIPER-TCP.sm
```diff
@@ -520,6 +544,20 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
     }
   }
 
+  action(ldmi_loadDoneMissInv, "ldmi",
```
mattsinc, line 552: I added this action/the updated transition below because BC in Pannotia from the weekly tests fails from the assert on line 539 in l_loadDoneMiss when another request evicted this line before the response returned.  However, although this update resolves that problem, there is now a segfault with BC that I am still chasing down.

All per checkin and daily GPU tests pass with this update, as well as a number of microbenchmarks that we are running to isolate latency/bandwidth/etc. for various GPU components.  So it seems like the issues are isolated to the "pre-emptive eviction" case that the larger scale of Pannotia introduces.

src/mem/ruby/protocol/GPU_VIPER-SQC.sm
mattsinc: I have been exclusively using wakeupAllBuffers(addr) in response to the variety of issues @KyleRoarty and I had to fix previously (in commits on Googlesource) where not using wakeupAllBuffers(addr) was leading to deadlocks.  I could be wrong though and can try this.

Regardless, I believe we want to wake up on the original address, since that is the one that the other pending requests were waiting on?

mattsinc:
> Can you rebase this change to have only your changes? Let me or @Harshil2107 know if you have any problems or need any help.

This is done, although as I noted a couple minutes ago, when testing the weekly tests, some of them fail with a segfault when the line with a pending load is evicted by another memory request before the response is returned -- so I am still digging into that.  All per commit and daily GPU tests pass, so this is something exclusive to larger workloads with more contention.  I have attempted to model the support in the new ldmi action directly on how the GPU atomics and stores handle this same case in the TCP (the cache line potentially not being present in the cache on a return), but thus far running with gdb has been extremely difficult as the source of the segfault seems to keep changing when gdb and/or gem5.debug is used ...

Hoping @abmerop may spot something wrong with the ldmi action I added while I dig further.

src/mem/ruby/protocol/GPU_VIPER-SQC.sm
```diff
@@ -308,22 +331,52 @@ machine(MachineType:SQC, "GPU SQC (L1 I Cache)")
...
+  // the cache.  Thus, complete this request without allocating the line, but
+  // still deallocate TBE and wakeup any dependent addresses.
+  transition(I, Data) {TagArrayRead, TagArrayWrite, DataArrayRead} {
+    // don't profile this as a hit/miss since it's a reponse from L2,
+    // so we already counted it
+    l_loadDoneMiss;
+    wada_wakeUpAllDependentsAddr;
+    d_deallocateTBE;
+    pr_popResponseQueue;
+  }
```
abmerop, line 364: I'm not sure this is possible to happen.  There is no transition(IV, Repl, I) case that I can see at least.

src/mem/ruby/protocol/GPU_VIPER-TCP.sm
```diff
@@ -687,14 +737,38 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
...
+  transition(I, TCC_Ack) {TagArrayRead, TagArrayWrite} {
+    pr_popResponseQueue;
+    wada_wakeUpAllDependentsAddr;
+    // NOTE: Because we invalidated the cache line, the assert in l_loadDoneMiss
+    // will fail -- unlike atomics that automatically go to I when the line returns
+    // loads do not automatically go to I.  Resolve this by passing data from
+    // message.
+    ldmi_loadDoneMissInv;
+    d_deallocateTBE;
+  }
```
abmerop: Perhaps the seg fault is if you have two TCC_Ack responses or something?  I don't really have any other ideas.

I don't see a good way around this, but a line in a transient state shouldn't be selected as a victim IMO.  It doesn't appear that cacheProbe considers the "AccessPermission" of a line though. Maybe it just works out that the replacement policy only picks addresses in base cases because they are the oldest.

Really it seems like if there are *no* lines in a base state that can be replaced, we should stall the incoming requests until there is one.  TCP looks like it is not allocating until a response is received, so I guess you would have to count the number of in-flight requests to a cache set and stall further requests if that is >= the associativity.  That's a big change though, so probably finding the seg fault is going to be easier.

mattsinc:
@abmerop , @powerjg : I have a fix now for the Pannotia applications, but it presents a new challenge that requires discussion.

The issue is: in some of the kernels in the Pannotia applications have accesses to the same cache line where some of the accesses are GLC (i.e., they bypass the GPU L1 cache) and others are non-GLC (i.e., they want to be cached in the GPU L1 cache).  We have support already per CU in the above code.  However, the problem here is that these requests are coming from different CUs and happening concurrently (seemingly because different WGs are at different points in the kernel around the same time).

Here is the issue: our support at the L2 for the TBEs overwrites the information about the GPU bypassing bits (SLC, GLC) every time: https://gem5.googlesource.com/public/gem5/+/refs/heads/develop/src/mem/ruby/protocol/GPU_VIPER-TCC.sm#738 --> https://gem5.googlesource.com/public/gem5/+/refs/heads/develop/src/mem/ruby/protocol/GPU_VIPER-TCC.sm#542.  The problem is when the second (non-GLC) load reaches the TCC, it overwrites the SLC/GLC information for the first (GLC) load.  Thus, when the the first load returns from the directory/memory, it no longer has the GLC bit set, which causes an assert failure at the TCP.

I see four possible fixes, but need feedback:

1.  We change the pending requests at the TCC/L2 to not allow additional requests for a cache line to pass through while another request to the same line is pending.  Basically this: https://gem5.googlesource.com/public/gem5/+/refs/heads/develop/src/mem/ruby/protocol/GPU_VIPER-TCC.sm#737, would do a st_stallAndWaitRequest instead.  I have implemented this locally and verified that the Pannotia apps work with this change.  However, it is a big change to the L2 as it now (correctly?) blocks multiple requests from the same cache line from going beyond the L2 in parallel.

2.  We change the way messages flow through Ruby such that we send the GLC/SLC bits all the way to the final destination (e.g., Directory, Main Memory) such that we can pass them all the way back -- this way the information can never be overwritten by another concurrent request.  But it requires changes to components that are not solely used by the GPU.  Originally @v-ramadas had made this change: https://gem5-review.googlesource.com/c/public/gem5/+/66991/5/src/mem/ruby/protocol/RubySlicc_MemControl.sm#77, but @powerjg 's feedback was to use the TBE at the L2 instead, which @v-ramadas subsequently did: https://gem5-review.googlesource.com/c/public/gem5/+/67191.  Given the new access pattern we didn't envision at the time, we could revert to Vishnu's old version and resolve this problem without blocking requests at the L2 like above in option 1?

3.  Have L2 TBEs track multiple pending requests to the same line, which would then require a) changes to the TBE to hold multiple pending requests and b) need a unique identifier to differentiate these requests once the request returns.  This is probably the most invasive change, but I suspect probably more realistic relative to real hardware?

4.  Require bypassing accesses to create a TBE at the L1/TCP, which we can use to get the SLC/GLC information when it returns.  This would prevent bypassing and non-bypassing requests from proceeding in parallel from the same CU, ensurnig that the TBE information has the correct SLC/GLC information.  However, this would not allow even multiple non-bypassing requests from the same CU to proceed in parallel, which seems like the opposite of what hardware would do.

Please let me know what you think.  I can push the updates to this accordingly.

src/mem/ruby/protocol/GPU_VIPER-SQC.sm
mattsinc: This hits on something else that has been bothering me about the current support.  

Note: this example is for the TCP, even though you asked about the SQC (apparently I can't add a comment on the line in question here since it's not a line I modified in this patch).

When a load returns from the L2 (https://github.com/gem5/gem5/pull/540/files#diff-c3f37e59aaafb17f88001c748f5d290395705147dcbfc4a064566431dc31e71cR278), assuming it's not a GLC/SLC load, the load either does a TCC_Ack if the line is valid or evicts something else to make space.  I'm pretty sure this means we'll never get to the transition(I, TCC_Ack...) transition because we'll force someone else out.  I have a local change I'm testing to resolve this, which basically checks if there is a TBE entry but the line is invalid and if so assumes the line got evicted and treats the load like a bypassing load's response.  This seems right to me -- if we evicted the line, we don't want to put it right back in.  But would appreciate any insight or feedback on if this sounds reasonable or is overoptimizing.

Regarding your SQC comment: it looks like I need to update 343 (https://github.com/gem5/gem5/pull/540/files#diff-1bbe1d71ce75a763a3c6ec0ed1ea0945249815209fc73edbb3ddebfc8590c139R343) to include IV?

mattsinc:
> @abmerop , @powerjg : I have a fix now for the Pannotia applications, but it presents a new challenge that requires discussion.
> 
> The issue is: in some of the kernels in the Pannotia applications have accesses to the same cache line where some of the accesses are GLC (i.e., they bypass the GPU L1 cache) and others are non-GLC (i.e., they want to be cached in the GPU L1 cache). We have support already per CU in the above code. However, the problem here is that these requests are coming from different CUs and happening concurrently (seemingly because different WGs are at different points in the kernel around the same time).
> 
> Here is the issue: our support at the L2 for the TBEs overwrites the information about the GPU bypassing bits (SLC, GLC) every time: https://gem5.googlesource.com/public/gem5/+/refs/heads/develop/src/mem/ruby/protocol/GPU_VIPER-TCC.sm#738 --> https://gem5.googlesource.com/public/gem5/+/refs/heads/develop/src/mem/ruby/protocol/GPU_VIPER-TCC.sm#542. The problem is when the second (non-GLC) load reaches the TCC, it overwrites the SLC/GLC information for the first (GLC) load. Thus, when the the first load returns from the directory/memory, it no longer has the GLC bit set, which causes an assert failure at the TCP.
> 
> I see four possible fixes, but need feedback:
> 
>     1. We change the pending requests at the TCC/L2 to not allow additional requests for a cache line to pass through while another request to the same line is pending.  Basically this: https://gem5.googlesource.com/public/gem5/+/refs/heads/develop/src/mem/ruby/protocol/GPU_VIPER-TCC.sm#737, would do a st_stallAndWaitRequest instead.  I have implemented this locally and verified that the Pannotia apps work with this change.  However, it is a big change to the L2 as it now (correctly?) blocks multiple requests from the same cache line from going beyond the L2 in parallel.
> 
>     2. We change the way messages flow through Ruby such that we send the GLC/SLC bits all the way to the final destination (e.g., Directory, Main Memory) such that we can pass them all the way back -- this way the information can never be overwritten by another concurrent request.  But it requires changes to components that are not solely used by the GPU.  Originally @v-ramadas had made this change: https://gem5-review.googlesource.com/c/public/gem5/+/66991/5/src/mem/ruby/protocol/RubySlicc_MemControl.sm#77, but @powerjg 's feedback was to use the TBE at the L2 instead, which @v-ramadas subsequently did: https://gem5-review.googlesource.com/c/public/gem5/+/67191.  Given the new access pattern we didn't envision at the time, we could revert to Vishnu's old version and resolve this problem without blocking requests at the L2 like above in option 1?
> 
>     3. Have L2 TBEs track multiple pending requests to the same line, which would then require a) changes to the TBE to hold multiple pending requests and b) need a unique identifier to differentiate these requests once the request returns.  This is probably the most invasive change, but I suspect probably more realistic relative to real hardware?
> 
>     4. Require bypassing accesses to create a TBE at the L1/TCP, which we can use to get the SLC/GLC information when it returns.  This would prevent bypassing and non-bypassing requests from proceeding in parallel from the same CU, ensurnig that the TBE information has the correct SLC/GLC information.  However, this would not allow even multiple non-bypassing requests from the same CU to proceed in parallel, which seems like the opposite of what hardware would do.
> 
> 
> Please let me know what you think. I can push the updates to this accordingly.

After talking with Brad and @abmerop , we decided the best way to model this was option 5: have the directory be the point at which we do not allow additional requests for a cache line to pass through while another request to the same line is pending.  Brad felt this was more realistic to support.  Thus, in the version I just pushed, I added that support.  Doing so required fixing several new corner cases though, which I have updated the comments in the commit and the commit message to explain.

I have tested this on all nightly/daily and weekly tests, and it passes for them.

powerjg (commented):
Seems OK to me. I'll let @abmerop or Brad approve.

mattsinc:
> Seems OK to me. I'll let @abmerop or Brad approve.

Thanks @powerjg , I just updated the commit message as I realized I did not explain/justify one case.  Done now.

I don't know if Brad has a Github ID, so @abmerop will have to stand in for him :)

mattsinc:
Rebased to head of develop

abmerop (dismissed):
This looks fine to me.  It seems to me that the TBE SLC/GLC bit is only used for local atomic now, is that right?

I'd like to test all of the atomic changes together but right now there are 3 different patches that all conflict on the TCC, so whoever submits first doesn't have to rebase I guess..

mattsinc:
> This looks fine to me. It seems to me that the TBE SLC/GLC bit is only used for local atomic now, is that right?
> 
> I'd like to test all of the atomic changes together but right now there are 3 different patches that all conflict on the TCC, so whoever submits first doesn't have to rebase I guess..

At some level I think the TBE GLC/SLC is not used directly anymore, even for local atomics -- right?

Which other patches?  @DanKouch merged in one, which I've integrated when I rebased develop.  My guess is you mean Anoop's and #120?

mattsinc:
@abmerop : my previous commit had some sort of merge conflict for TCP, which I attempted to resolve and push again.  However I see see this above:

This branch has conflicts that must be resolved
Use the [web editor](https://github.com/gem5/gem5/pull/540/conflicts) or the to resolve conflicts.
Conflicting files
src/mem/ruby/protocol/GPU_VIPER-TCP.sm

Not sure what more to do ...

mattsinc:
> @abmerop : my previous commit had some sort of merge conflict for TCP, which I attempted to resolve and push again. However I see see this above:
> 
> This branch has conflicts that must be resolved Use the [web editor](https://github.com/gem5/gem5/pull/540/conflicts) or the to resolve conflicts. Conflicting files src/mem/ruby/protocol/GPU_VIPER-TCP.sm
> 
> Not sure what more to do ...

I tried manually editing in Github instead this time and that seems to have worked ... so I guess this is ready for review again.

Let me know what tests you want me to run, or if you want me to pull their patches locally and test things.

abmerop:
The 3 patches I meant haven't been committed yet.  Was referring to this one, #120 and #546 all touch TCC in similar places.  Shouldn't be a difficult merge just a heads up..

mattsinc:
> The 3 patches I meant haven't been committed yet. Was referring to this one, #120 and #546 all touch TCC in similar places. Shouldn't be a difficult merge just a heads up..

Thanks.  Once my current tests are done, I will re-run will these two (and @DanKouch's #397) all together to see what happens and report back.

mattsinc:
> > The 3 patches I meant haven't been committed yet. Was referring to this one, #120 and #546 all touch TCC in similar places. Shouldn't be a difficult merge just a heads up..
> 
> Thanks. Once my current tests are done, I will re-run will these two (and @DanKouch's #397) all together to see what happens and report back.

There is one compilation issue after merging #397 into my branch, but I'll hold off on pushing until the tests complete so to avoid multiple pushes here as much as possible.

mattsinc:
> > > The 3 patches I meant haven't been committed yet. Was referring to this one, #120 and #546 all touch TCC in similar places. Shouldn't be a difficult merge just a heads up..
> > 
> > 
> > Thanks. Once my current tests are done, I will re-run will these two (and @DanKouch's #397) all together to see what happens and report back.
> 
> There is one compilation issue after merging #397 into my branch, but I'll hold off on pushing until the tests complete so to avoid multiple pushes here as much as possible.

All tests in nightly and weekly except LULESH (which already had a bug so I didn't test it) and FW (which fails to compile ... looking into) pass with this patch and #397 together.

There are also compilation issues when integrating #120 and #546 into this patch + #397.  Will just push all compilation updates as one once done testing.

abmerop:
What's the LULESH bug this time? ...

mattsinc:
> What's the LULESH bug this time? ...

src/gpu-compute/gpu_dyn_inst.cc:787: fatal: flat access at addr 0x2000000000000800 has a memory violation

abmerop:
> > What's the LULESH bug this time? ...
> 
> src/gpu-compute/gpu_dyn_inst.cc:787: fatal: flat access at addr 0x2000000000000800 has a memory violation

OK, we should discuss that elsewhere.  For the record, I just tested LULESH on GPUFS with develop tip (PR 563) and it completes successfully with the `0.01 2` options and ~75 minutes and matches hardware output.

mattsinc:
> > > What's the LULESH bug this time? ...
> > 
> > 
> > src/gpu-compute/gpu_dyn_inst.cc:787: fatal: flat access at addr 0x2000000000000800 has a memory violation
> 
> OK, we should discuss that elsewhere. For the record, I just tested LULESH on GPUFS with develop tip (PR 563) and it completes successfully with the `0.01 2` options and ~75 minutes and matches hardware output.

Digging deeper, it appears the problem is that with the old weekly.sh script (which I'm using to run tests locally), when one specifies the VEGA_X86 build, it does not specify --gfx-version=gfx900 or --dgpu.  So it seems like it must be running with the gfx8 ISA on the VEGA architecture and this causes problems.  When I manually add those flags and run locally, it does not have the above failure (it is still running since I didn't set to 2 iterations, but it is past the prior failure point).  Probably I should update locally for tests though ...

Since we are going to be moving to the new testing system, I'm not sure if it's worth fixing this or not.

abmerop:
> > > > What's the LULESH bug this time? ...
> > > 
> > > 
> > > src/gpu-compute/gpu_dyn_inst.cc:787: fatal: flat access at addr 0x2000000000000800 has a memory violation
> > 
> > 
> > OK, we should discuss that elsewhere. For the record, I just tested LULESH on GPUFS with develop tip (PR 563) and it completes successfully with the `0.01 2` options and ~75 minutes and matches hardware output.
> 
> Digging deeper, it appears the problem is that with the old weekly.sh script (which I'm using to run tests locally), when one specifies the VEGA_X86 build, it does not specify --gfx-version=gfx900 or --dgpu. So it seems like it must be running with the gfx8 ISA on the VEGA architecture and this causes problems. When I manually add those flags and run locally, it does not have the above failure (it is still running since I didn't set to 2 iterations, but it is past the prior failure point). Probably I should update locally for tests though ...
> 
> Since we are going to be moving to the new testing system, I'm not sure if it's worth fixing this or not.

Yea, probably not worth looking into a fix.  Running gfx803 (likely) on Vega build is well into undefined behavior territory.

mattsinc:
> > > > > What's the LULESH bug this time? ...
> > > > 
> > > > 
> > > > src/gpu-compute/gpu_dyn_inst.cc:787: fatal: flat access at addr 0x2000000000000800 has a memory violation
> > > 
> > > 
> > > OK, we should discuss that elsewhere. For the record, I just tested LULESH on GPUFS with develop tip (PR 563) and it completes successfully with the `0.01 2` options and ~75 minutes and matches hardware output.
> > 
> > 
> > Digging deeper, it appears the problem is that with the old weekly.sh script (which I'm using to run tests locally), when one specifies the VEGA_X86 build, it does not specify --gfx-version=gfx900 or --dgpu. So it seems like it must be running with the gfx8 ISA on the VEGA architecture and this causes problems. When I manually add those flags and run locally, it does not have the above failure (it is still running since I didn't set to 2 iterations, but it is past the prior failure point). Probably I should update locally for tests though ...
> > Since we are going to be moving to the new testing system, I'm not sure if it's worth fixing this or not.
> 
> Yea, probably not worth looking into a fix. Running gfx803 (likely) on Vega build is well into undefined behavior territory.

So far with all of #120, #397, and #540 integrated into this patch, all daily/nightly tests and all weekly tests through PageRank have passed (except LULESH, which is running separately with the above change).  I suspect the others will pass too given so many other large applications are passing.  However, this is not exhaustive for GPU atomic tests.  @abmerop  any chance you can run the internal atomic test you mentioned on #540 with this?  And @DanKouch can you please run the additional atomic microbenchmarks on this to validate?

abmerop:
I could.. but this patch needs rebasing.  I could also share the test

mattsinc:
@abmerop : updated to head of develop, pushed commit that integrates all of minor changes to integrate all of #120, #397, and #540.  All GPU tests pass for daily/nightly and weekly (except FW, which doesn't compile and which I'm looking into) when tested for VEGA_X86.

@DanKouch please test this on our internal GPU atomic microbenchmarks, @abmerop please test on Aaron's GPU atomic microbenchmark AMD has internally.  I think those should be the last tests needed?

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
```diff
@@ -527,9 +533,9 @@ machine(MachineType:Directory, "AMD Baseline protocol")
         out_msg.ForwardRequestTime := tbe.ForwardRequestTime;
         out_msg.ProbeRequestStartTime := tbe.ProbeRequestStartTime;
         out_msg.OriginalResponder := tbe.LastSender;
-        if(tbe.atomicData){
-          out_msg.WTRequestor := tbe.WTRequestor;
-        }
+        out_msg.isGLCSet := tbe.isGLCSet;
+        out_msg.isSLCSet := tbe.isSLCSet;
+        out_msg.CURequestor := tbe.CURequestor;
```
mattsinc: @BujSet: this integration with your patch looks wrong ... but also passes all tests.  Am I missing something?

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
```diff
@@ -565,13 +573,15 @@ machine(MachineType:Directory, "AMD Baseline protocol")
         out_msg.addr := address;
         out_msg.Type := CoherenceResponseType:NBSysWBAck;
         out_msg.Destination.add(in_msg.Requestor);
-        out_msg.WTRequestor := in_msg.WTRequestor;
+        out_msg.CURequestor := in_msg.CURequestor;
```
mattsinc, line 578: This is the change I made for integration with #397

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
mattsinc: I went ahead and updated it in the latest patch, but would appreciate it @BujSet if you can confirm this looks right on your end.  Also rerunning all tests.

abmerop:
The atomic benchmark is passing for both gfx900, gfx90a in GPUFS.

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
```diff
@@ -527,9 +533,11 @@ machine(MachineType:Directory, "AMD Baseline protocol")
         out_msg.ForwardRequestTime := tbe.ForwardRequestTime;
         out_msg.ProbeRequestStartTime := tbe.ProbeRequestStartTime;
         out_msg.OriginalResponder := tbe.LastSender;
-        if(tbe.atomicData){
-          out_msg.WTRequestor := tbe.WTRequestor;
-        }
+        out_msg.isGLCSet := tbe.isGLCSet;
+        out_msg.isSLCSet := tbe.isSLCSet;
+	if(tbe.atomicData){
```
BujSet, line 538: @mattsinc is this the part that you wanted review on? I think this should be fine, tbe.atomicData flag is set on an atomic request, irrespective of whether return data is needed. Though, looking through this, I imagine that we would not need tbe.atomicDataReturn and tbe.atomicDataNoReturn flags anymore since we can just use the GLC bits (since I believe they're now being propagated through all cache levels).

mattsinc: Yes, this is it.  I don't know if I agree about GLC/SLC vs. atomicDataReturn/NoReturn though -- there can be GLC atomics that either return or don't return, for example.

In any event I don't think this involves a decision for this patch though.

mattsinc:
> The atomic benchmark is passing for both gfx900, gfx90a in GPUFS.

Great, I think we are just waiting on @DanKouch's checks then.

DanKouch:
> > The atomic benchmark is passing for both gfx900, gfx90a in GPUFS.
> 
> Great, I think we are just waiting on @DanKouch's checks then.

@mattsinc All of our internal atomic benchmarks run to completion, and all of my correctness-checking test programs pass as well.

abmerop (approved):

mattsinc:
Thanks @DanKouch , @abmerop , @BujSet & co. for helping with this!

mattsinc:
> > > > > What's the LULESH bug this time? ...
> > > > 
> > > > 
> > > > src/gpu-compute/gpu_dyn_inst.cc:787: fatal: flat access at addr 0x2000000000000800 has a memory violation
> > > 
> > > 
> > > OK, we should discuss that elsewhere. For the record, I just tested LULESH on GPUFS with develop tip (PR 563) and it completes successfully with the `0.01 2` options and ~75 minutes and matches hardware output.
> > 
> > 
> > Digging deeper, it appears the problem is that with the old weekly.sh script (which I'm using to run tests locally), when one specifies the VEGA_X86 build, it does not specify --gfx-version=gfx900 or --dgpu. So it seems like it must be running with the gfx8 ISA on the VEGA architecture and this causes problems. When I manually add those flags and run locally, it does not have the above failure (it is still running since I didn't set to 2 iterations, but it is past the prior failure point). Probably I should update locally for tests though ...
> > Since we are going to be moving to the new testing system, I'm not sure if it's worth fixing this or not.
> 
> Yea, probably not worth looking into a fix. Running gfx803 (likely) on Vega build is well into undefined behavior territory.

@Harshil2107 the error you are seeing is what was reported here.

PR #546, opened 2023-11-09
mem-ruby: Fix for not creating log entries on atomic no return requests

BujSet:
Augmenting Datablock and WriteMask to support optional arg to distinguish between return and no return. In the case of atomic no return requests, log entries should not be created when performing the atomic.

Change-Id: Ic3112834742f4058a7aa155d25ccc4c014b60199a


src/mem/ruby/common/DataBlock.hh
```diff
@@ -96,7 +96,8 @@ class DataBlock
     void setData(PacketPtr pkt);
     void copyPartial(const DataBlock &dblk, int offset, int len);
     void copyPartial(const DataBlock &dblk, const WriteMask &mask);
-    void atomicPartial(const DataBlock & dblk, const WriteMask & mask);
+    void atomicPartial(const DataBlock & dblk, const WriteMask & mask,
+            bool isAtomicNoReturn=false);
```
abmerop: I would actually suggest the opposite for the default.  In other words, do not use atomic log by default. I think that will prevent needing to change any of the other protocols with use atomics besides VIPER.  That might be a bit confusing but if we rename the parameter to "noAtomicLog" or similar that would be clear.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -755,7 +760,16 @@ machine(MachineType:TCC, "TCC Cache")
   }
 
   action(pa_performAtomic, "pa", desc="Perform atomic") {
-    cache_entry.DataBlk.atomicPartial(cache_entry.DataBlk, cache_entry.writeMask);
+    peek(coreRequestNetwork_in, CPURequestMsg) {
+      if (in_msg.Type == CoherenceRequestType:AtomicReturn) {
+        cache_entry.DataBlk.atomicPartial(cache_entry.DataBlk, cache_entry.writeMask);
```
abmerop: if you agree with my prior comment, this would have false as the 3rd parameter

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -755,7 +760,16 @@ machine(MachineType:TCC, "TCC Cache")
...
+      } else {
+        // Set the isAtomicNoReturn flag to ensure that logs are not
+        // generated erroneously
+        assert(in_msg.Type == CoherenceRequestType:AtomicNoReturn);
+        cache_entry.DataBlk.atomicPartial(cache_entry.DataBlk, cache_entry.writeMask, true);
```
abmerop, line 813: and this true could stay but would be implied

src/mem/ruby/common/WriteMask.hh
```diff
@@ -230,7 +230,8 @@ class WriteMask
      * specific atomic operation.
      */
     void performAtomic(uint8_t * p,
-            std::deque<uint8_t*>& atomicChangeLog) const;
+            std::deque<uint8_t*>& atomicChangeLog,
+            bool isAtomicNoReturn=false) const;
```
abmerop: same thing here

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -363,7 +363,9 @@ machine(MachineType:TCC, "TCC Cache")
             } else {
                 trigger(Event:WrVicBlk, in_msg.addr, cache_entry, tbe);
             }
-        } else if (in_msg.Type == CoherenceRequestType:Atomic) {
+        } else if ((in_msg.Type == CoherenceRequestType:Atomic) ||
```
mysoreanoop: Could we perhaps get rid of the use of `CoherenceRequestType:Atomic` in VIPER files entirely? The Return or NoReturn would complete all cases, right?

src/mem/ruby/common/DataBlock.hh
BujSet: Oh, yeah that makes more sense. Will make the change.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
BujSet: I thought about that as well, and I agree that it would make more sense to remove. However, I though it might make future developments more problematic because now the CPU protocols that use the CPURequestMsg structure use the Atomic type, while the GPU ones use either AtomicReturn or AtomicReturn. I'm not sure what would be a better option. In my opinion, I feel like the GPU protocols should use their own GPURequest and Response messages that are different from the CPU ones, but this would be a pretty significant refactor, and has a large code duplication.

BujSet:
Not sure if the flow has changed, but is there anything I need to do to have the unit tests begin running on this PR? Seems like they're stalled, waiting on some approval before beginning (been stalled in this state since last night). From the "[gem5: A guide to contributing](https://www.gem5.org/contributing)" it mentions that "Once you have created a pull request, the gem5 Continuous Integration (CI) tests will run."

powerjg:
> Not sure if the flow has changed, but is there anything I need to do to have the unit tests begin running on this PR?

We're just having significant flakiness on our unit test runners right now. Things may take a while to run.

BujSet:
Is there a way to see more details on why certain tests are failing? When I click the "Details" option on the "CI Tests / unittests-all-opt (pull request)" tab, it takes me to a page that only says "This job failed." I tested my PR locally with `scons build/NULL/unittests.opt`, and all the tests that were enabled passed. Just trying to see if this is a runner issue (e.g. runner failed to launch tests and that's the 'failure'), or if this is a true error, and how I can reproduce it locally to debug.

mattsinc:
> Is there a way to see more details on why certain tests are failing? When I click the "Details" option on the "CI Tests / unittests-all-opt (pull request)" tab, it takes me to a page that only says "This job failed." I tested my PR locally with `scons build/NULL/unittests.opt`, and all the tests that were enabled passed. Just trying to see if this is a runner issue (e.g. runner failed to launch tests and that's the 'failure'), or if this is a true error, and how I can reproduce it locally to debug.

I don't know if it's the issue here or not, but the unit tester has been flaky and failing randomly the last couple days.  Not sure if we want to try rerunning first (don't know the answer to your question though).

abmerop (changes requested):
This isn't passing our atomic test (one written by Aaron if you recall).  I'll try and give you more details soon

abmerop:
I can't seem to add a comment outside the range of a hunk, but the new Atomic CoherenceRequestTypes need to be added to the directories.  This fixes the atomic test:

```--- a/src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
+++ b/src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
@@ -402,7 +402,9 @@ machine(MachineType:Directory, "AMD Baseline protocol")
           trigger(Event:RdBlkM, in_msg.addr, entry, tbe);
         } else if (in_msg.Type == CoherenceRequestType:WriteThrough) {
           trigger(Event:WriteThrough, in_msg.addr, entry, tbe);
-        } else if (in_msg.Type == CoherenceRequestType:Atomic) {
+        } else if (in_msg.Type == CoherenceRequestType:Atomic ||
+                   in_msg.Type == CoherenceRequestType:AtomicReturn ||
+                   in_msg.Type == CoherenceRequestType:AtomicNoReturn) {
           trigger(Event:Atomic, in_msg.addr, entry, tbe);
         } else if (in_msg.Type == CoherenceRequestType:VicDirty) {
           if (getDirectoryEntry(in_msg.addr).VicDirtyIgnore.isElement(in_msg.Requestor)) {
@@ -746,7 +748,9 @@ machine(MachineType:Directory, "AMD Baseline protocol")

         // CPU + GPU or GPU only system
         if ((in_msg.Type != CoherenceRequestType:WriteThrough &&
-             in_msg.Type != CoherenceRequestType:Atomic) ||
+             in_msg.Type != CoherenceRequestType:Atomic &&
+             in_msg.Type != CoherenceRequestType:AtomicReturn &&
+             in_msg.Type != CoherenceRequestType:AtomicNoReturn) ||
              !in_msg.NoWriteConflict) {
           if (noTCCdir) {
             probe_dests.add(mapAddressToRange(address, MachineType:TCC,
```


I don't think the other directories are even used but wouldn't hurt to make the same change there too.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -609,7 +611,10 @@ machine(MachineType:TCC, "TCC Cache")
...
     if (coreRequestNetwork_in.isReady(clockEdge())) {
       peek(coreRequestNetwork_in, CPURequestMsg) {
-        if(in_msg.Type == CoherenceRequestType:RdBlk || in_msg.Type == CoherenceRequestType:Atomic){
+        if(in_msg.Type == CoherenceRequestType:RdBlk ||
+           in_msg.Type == CoherenceRequestType:Atomic ||
+           in_msg.Type == CoherenceRequestType:AtomicReturn ||
+           in_msg.Type == CoherenceRequestType:AtomicNoReturn){
           tbe.Destination.add(in_msg.Requestor);
         }
         tbe.isGLCSet := in_msg.isGLCSet;
```
abmerop, line 618: (2 lines down)  Do we need to check for AtomicReturn/AtomicNoReturn there too?

Edit: To be clear, line 622

abmerop (commented):
Thanks for all the updates!  I think this should mostly cover all the new types, but have one potential spot missing commented on

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
abmerop: Thinking about it more, I believe it is needed there as TCP can now generate two request types and that handles Atomics to same address.  I didn't find any others missing though, so that would be the last change

mattsinc: I agree we need to check there.

abmerop (approved):
LGTM

src/mem/ruby/protocol/MOESI_AMD_Base-RegionBuffer.sm
```diff
@@ -458,7 +458,9 @@ machine(MachineType:RegionBuffer, "Region Buffer for AMD_Base-like protocol")
             trigger(Event:CPUWrite, in_msg.addr, cache_entry, tbe);
           } else if (in_msg.Type == CoherenceRequestType:WriteThrough ) {
             trigger(Event:CPUWrite, in_msg.addr, cache_entry, tbe);
-          } else if (in_msg.Type == CoherenceRequestType:Atomic ) {
+          } else if (in_msg.Type == CoherenceRequestType:Atomic ||
```
mattsinc, line 461: I think it's fine to change this file too, although I don't think it's tested?

src/mem/ruby/protocol/MOESI_AMD_Base-Region-dir.sm
```diff
@@ -436,7 +439,9 @@ machine(MachineType:Directory, "AMD_Base-like protocol")
           trigger(Event:RdBlkS, in_msg.addr, entry, tbe);
         } else if (in_msg.Type == CoherenceRequestType:RdBlkM) {
           trigger(Event:RdBlkM, in_msg.addr, entry, tbe);
-        } else if (in_msg.Type == CoherenceRequestType:Atomic) {
+        } else if (in_msg.Type == CoherenceRequestType:Atomic ||
```
mattsinc, line 442: It's fine to change this file too, but I don't think it's tested?

mattsinc (approved):

PR #552, opened 2023-11-13
cpu: Remove SLC bit restraint for GPU tester

DanKouch:
This reverts gem5#133, the temporary work-around for gem5#131, allowing both SLC and GLC atomic requests to be made in the GPU tester.

The underlying issues behind gem5#131 have been resolved by gem5#367 and gem5#397.

Change-Id: Iab8817192a4aaae6a691e8ac442acaf1844f1aec

### DIFF
--- a/src/cpu/testers/gpu_ruby_test/gpu_wavefront.cc
+++ b/src/cpu/testers/gpu_ruby_test/gpu_wavefront.cc
@@ -189,7 +189,6 @@ GpuWavefront::issueAtomicOps()
                                              AtomicOpFunctorPtr(amo_op));
         req->setPaddr(address);
         req->setReqInstSeqNum(tester->getActionSeqNum());
-        req->setCacheCoherenceFlags(Request::SLC_BIT);
         // set protocol-specific flags
         setExtraRequestFlags(req);
 


DanKouch:
@mattsinc @abmerop The issues which were causing the ruby random tester to fail have been resolved, so #133 can be reverted. I've run the random tester for both write-back and write-through L2s.

mattsinc (approved):
I'm not sure CPU is the proper tag for this, but it is what @abmerop used in #133, so I guess leave it be to be consistent.

abmerop (approved):
The tag is mainly due to it being in the cpu directory. I think eventually only the github tags will be used but that hasn't been discussed yet

abmerop:
Looks like you need a change ID... If you did not make any local git add's you can do `git commit --amend`, close editor, and it should add one

DanKouch:
@abmerop

Added, thanks!.

It looks like the `pre-commit install` command on the contributing page doesn't install hooks for commit-msg (which is what the Change-ID hook uses).

mattsinc:
> @abmerop
> 
> Added, thanks!.
> 
> It looks like the `pre-commit install` command on the contributing page doesn't install hooks for commit-msg (which is what the Change-ID hook uses).

Yeah, unfortunately one has to install it separately.

mattsinc:
> This reverts #133, the temporary work-around for #131, allowing both SLC and GLC atomic requests to be made in the GPU tester.
> 
> The underlying issues behind #131 have been resolved by #367 and #397.

@DanKouch you will also need to update this "message" box to include the commit ID as I believe this is what shows up in the git commit log.

mattsinc:
Seems like the tests were run before adding the ChangeID?  If so, updating the branch to head of develop may kick them off again?

mattsinc:
> Seems like the tests were run before adding the ChangeID? If so, updating the branch to head of develop may kick them off again?

Nevermind, with the updates I see now the ChangeID failure is gone (but @DanKouch will need to update to head of develop eventually).

PR #553, opened 2023-11-13
arch-arm: Only build ArmCapstoneDisassembler when ISA is arm

hnpl:
Currently, if the Capstone header file is found in the host system, scons will try to build the ArmCapstoneDisassembler regardless of the gem5 target ISA. This is causing problem when the host has Capstone, but the gem5 target ISA is not arm. Compiling gem5 in this case will cause errors, e.g., ArmISA and ArmSystem is not found.

This change aims to prevent building the ArmCapstoneDisassembler when the gem5 target ISA is not arm.

Ref:
[1] The Arm Capstone PR https://github.com/gem5/gem5/pull/494

Change-Id: I1e714d34aec8fe2a2af8cd351536951053a4d8a5

### DIFF
--- a/src/arch/arm/tracers/SConscript
+++ b/src/arch/arm/tracers/SConscript
@@ -43,7 +43,7 @@ Source('tarmac_tracer.cc', tags='arm isa')
 Source('tarmac_record.cc', tags='arm isa')
 Source('tarmac_record_v8.cc', tags='arm isa')
 
-if env['CONF']['HAVE_CAPSTONE']:
+if env['CONF']['USE_CAPSTONE']:
     SimObject('ArmCapstone.py', sim_objects=['ArmCapstoneDisassembler'],
-              tags=['capstone', 'arm isa'])
-    Source('capstone.cc', tags=['capstone', 'arm isa'])
+              tags=['arm isa'])
+    Source('capstone.cc', tags=['arm isa'])
--- a/src/cpu/Kconfig
+++ b/src/cpu/Kconfig
@@ -27,3 +27,9 @@ config HAVE_CAPSTONE
     def_bool $(HAVE_CAPSTONE)
 
 rsource "kvm/Kconfig"
+
+config USE_CAPSTONE
+    depends on HAVE_CAPSTONE
+    depends on USE_ARM_ISA
+    bool "Use CapstoneDisassembler"
+    default y
--- a/src/cpu/SConscript
+++ b/src/cpu/SConscript
@@ -115,9 +115,11 @@ Source('simple_thread.cc')
 Source('thread_context.cc')
 Source('thread_state.cc')
 Source('timing_expr.cc')
-SourceLib('capstone', tags='capstone')
-Source('capstone.cc', tags='capstone')
-SimObject('Capstone.py', sim_objects=['CapstoneDisassembler'], tags='capstone')
+
+if env['CONF']['USE_CAPSTONE']:
+    SourceLib('capstone')
+    Source('capstone.cc')
+    SimObject('Capstone.py', sim_objects=['CapstoneDisassembler'])
 
 SimObject('DummyChecker.py', sim_objects=['DummyChecker'])
 Source('checker/cpu.cc')


hnpl:
I'm not sure if this is the best fix for this problem as having to check `USE_ARM_ISA` of a file under `src/arch/arm` is rather silly.

I'm confused by the SCons tags. When SCons if the Capstone library is available [1], it will imply the `capstone` tag if the library exists in the host. However, the `ArmCapstoneDisassembler` has 2 tags: `arm isa` and `capstone` [2]. Since I'm not trying to build the arm ISA, I'm quite certain that the `arm isa` is not implied. It seems that SCons tried to build an object if at least one of the tags is enabled.

[1] https://github.com/gem5/gem5/blob/develop/src/cpu/SConsopts#L46
[2] https://github.com/gem5/gem5/blob/develop/src/arch/arm/tracers/SConscript#L48

giactra (dismissed):
I am going to approve this, but I agree it is really confusing.. I thought the presence of the arm isa tag would imply this would only get compiled when building for Arm...

BobbyRBruce:
> I am going to approve this, but I agree it is really confusing.. I thought the presence of the arm isa tag would imply this would only get compiled when building for Arm.

I'm not 100% sure on this but my understanding of how tagging works is like this (apologies for my primitive pseudocode Scons):

```py
SimObject('x.py', sim_objects=['X']m tags=['A', 'B'])

if condition_a_met():
   conf.env.TagImplies('A', 'gem5 lib')

if condition_b_met():
   conf.env.TagImplies('B', 'gem5 lib')

# Somewhere in Scons something like this exists:

for simobject in simobject_list:
    if simobject.get_tags() is None or simobject.contains_tag('gem5 lib'):
        compile(simobject)
```

Therefore in this example SimObject `X` is compiled if either `condition_a_met` or `condition_b_met` returns true. It's an OR not an AND. 

If i'm right we can only safely tags to filter compilation with one tag per SimObject using this approach.

If you really wanted to continue with tagging you could make a new tag called `arm and capstone`:

```py
if env['CONF']['HAVE_CAPSTONE'] and env['CONF']['USE_ARM_ISA']:
    conf.env.TagImplies('arm with capstone', 'gem5 lib')

...

SimObject('ArmCapstone.py', sim_objects=['ArmCapstoneDisassembler'],
              tags='arm with capstone')
    Source('capstone.cc', tags='arm with capstone')

```

This is probably more in-keeping with how this is supposed to work. I believe the `conf.env.TagImplies('X', 'gem5 lib')` is indirectly stating "`X` is a requirement for compiling the gem5 library".

I'd honestly prefer we these tags and just use `if env['CONF']['USE_ARM_ISA']` where needed as it's more explicit.

hnpl:
Thanks Giacomo for reviewing this.

Thanks for explaining this Bobby, it makes sense now.

I think the right solution would be to move the capstone check to `src/` instead of `src/cpu/` because I believe that `src/cpu/` does not know about the target ISA. In `src/`, we only `tagImplies('capstone')` when capstone is available and the target architecture is arm. I'll spend some time updating this.

BobbyRBruce (changes requested):
Giving a "Request changes" to blow the change as Hoa isn't ready for this to be merged just yet.

hnpl:
This turns out to be way trickier than I thought. Essentially, if there are more than one ISAs supporting capstone then having the `capstone` tag in both ISAs wouldn't work as scons will try to build objects specific to both ISAs. That basically means that we should not have `capstone` as a tag.

So, we must determine whether we use capstone or not somewhere that we have access to both target ISA list and capstone check. I guess the best place is in `src/SConscript`. The implementation defines `env['CONF']['HAVE_CAPSTONE']` as whether capstone is available, and `env['CONF']['USE_CAPSTONE']` as whether capstone is available and the ISA supports capstone. This way we don't have to worry about multi-ISA support.

BobbyRBruce:
I'm not entirely use it does, but make sure any changes you do don't class too much with the upcoming KConfig chances (#69). They should be merged soon.

src/SConscript
```diff
@@ -61,6 +61,13 @@ Import('*')
 if env['CONF']['USE_EFENCE']:
     env.Append(LIBS=['efence'])
 
+# Check for capstone dependency. Currently, only arm ISA supports capstone
+# disassembler, we only allow building capstone-related object if arm is one
+# of the simulated ISAs.
+
+env['CONF']['USE_CAPSTONE'] = \
+    env['CONF']['HAVE_CAPSTONE'] and env['USE_ARM_ISA']
+
```
rogerchang23424: Note: The options will need to be in the Kconfig file after #69 merged.

BobbyRBruce:
@hnpl : The KConfig changes have been merged. Do you have time to get this patch refactored to work with it in time for the v23.1 staging branch creation (Dec 1st)?

hnpl:
Is there a quick way to test whether this change works with KConfig?

hnpl:
Ping @rogerchang23424

src/SConscript
rogerchang23424: The option `USE_CAPSTONE` is depends on the config option `HAVE_CAPSTONE` and `USE_ARM_ISA`. It should be set in the `src/cpu/Kconfig` rather than in the set in `SConscript` directly. 

https://github.com/gem5/gem5/blob/d94d6017b051d1b33d8950bddcbb89662fbec503/src/cpu/Kconfig#L26

The script in Kconfig should be like following:

```
config USE_CAPSTONE
      depends on HAVE_CAPSTONE
      depends on USE_ARM_ISA
      bool "Use CapstoneDisassembler"
      default y
```
The USE_CAPSTONE will set to True by default if ARM ISA is build and capstone is detected. The USE_CAPSTONE option can be set via `setconfig` or `menuconfig` command.

For the config options dependency in the Kconfig, please see:

https://docs.kernel.org/kbuild/kconfig-language.html#menu-attributes

powerjg:
@hnpl can you rebase on develop and try to see if the fix still works?

BobbyRBruce (approved):

PR #522, opened 2023-10-31
arch-riscv: Support combination of privilege modes configuration

rogerchang23424:
The user can select privilege modes witch is included in the system, not always enable the user and supervisor privilege modes.

Change-Id: I6978753ac28455199dc235c25e3bb809b4f0310b

PR #554, opened 2023-11-13
arch-riscv: Move fault handler addr logic to ISA

wmin0:
mtvec.mode is extended in the new riscv proposal, like fast interrupt. This change moves that part from Fault class to ISA class for extendable.

Ref: https://github.com/riscv/riscv-fast-interrupt

Change-Id: Ifcfd6f4ce011149b5982dabfd2421f48c9a10b75


wmin0:
CC @rogerchang23424

rogerchang23424:
Wait for #522 merged to avoid conflict

rogerchang23424 (approved):

BobbyRBruce (approved):

PR #564, opened 2023-11-14
mem-cache: Prefetchers Improvements

rjc-arch:
This pull request contains a set of small patches which fix some bugs in the gem5 prefetchers, and aligns out-of-the box prefetcher performance more closely with that which a typical user would expect.

The performance patches have been tested with an out-of-the-box (untuned) Stride prefetcher configuration against a set of SPEC 2017 SimPoints, and show a modest IPC uplift across the board, with no IPC degradation.

The new defaults were identified as part of work on gem5 prefetchers undertaken by Nikolaos Kyparissas while on internship at Arm.

### DIFF
--- a/src/mem/cache/prefetch/Prefetcher.py
+++ b/src/mem/cache/prefetch/Prefetcher.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2012, 2014, 2019 ARM Limited
+# Copyright (c) 2012, 2014, 2019, 2022 Arm Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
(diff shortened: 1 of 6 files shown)


BobbyRBruce (dismissed):
Though I have little knowledge of the prefetches, I can't find anything wrong here. Nothing here appears capable of breaking anything for current users.

So it's a thumbs-up from me but i'll let it sit a while so others can have a look.

src/mem/cache/prefetch/Prefetcher.py
```diff
@@ -192,6 +192,7 @@ class StridePrefetcher(QueuedPrefetcher):
     use_requestor_id = Param.Bool(True, "Use requestor id based history")
 
     degree = Param.Int(4, "Number of prefetches to generate")
+    distance = Param.Int(0, "How far ahead of the demand stream to prefetch")
```
powerjg: Could you add a sentence here about what "0" means? Does 0 mean the *next* block or the current block?

If "0" means the current block, then we should probably set the default to "1", right?

powerjg (commented):
Everything looks fine to me. Just one small comment/question below.

rjc-arch:
Thanks for the comments @BobbyRBruce & @powerjg! I've updated the comment on the `distance` `Param` as requested, and added some more detail to the commit message.
I've also changed the `Param` to a `Param.Unsigned`, as the distance should be zero or positive.
I've re-tested the updated set of patches on a sample of SimPoints and the changes continue to work as expected.

powerjg (approved):
Thanks for the updates!

odanrc:
A bit late to the party, but LGTM

PR #565, opened 2023-11-14
gpu-compute: Minor edits for atomic no returns and stores

BujSet:
Since returned data is not needed for AtomicNoReturn and Store memory requests, the coalescer need not spend time writing in dummy data for packets of these types.

Change-Id: Ie669e8c2a3bf44b5b0c290f62c49c5d4876a9a6a

### DIFF
--- a/src/mem/ruby/system/GPUCoalescer.cc
+++ b/src/mem/ruby/system/GPUCoalescer.cc
@@ -600,8 +600,8 @@ GPUCoalescer::hitCallback(CoalescedRequest* crequest,
                 // data response is not needed.
                 case RubyRequestType_ATOMIC_NO_RETURN:
                     assert(pkt->isAtomicOp());
+                    break;
                 case RubyRequestType_ST:
-                    data.setData(pkt->getPtr<uint8_t>(), offset, pkt_size);
                     break;
                 case RubyRequestType_LD:
                     pkt->setData(data.getData(offset, pkt_size));


BujSet:
@abmerop Note sure who else would need to review this, but just following up on one of the discussions about atomic no return requests. I believe that this operation is not needed (it's probably not a huge deal if it needs to remain). Just might make the code easier to follow if it wasn't there.

src/mem/ruby/system/GPUCoalescer.cc
```diff
@@ -601,7 +601,6 @@ GPUCoalescer::hitCallback(CoalescedRequest* crequest,
                 case RubyRequestType_ATOMIC_NO_RETURN:
                     assert(pkt->isAtomicOp());
```
abmerop, line 602: Surprised there no compiler error about implicit fall-through here.  You could optionally add a `break;` here while you are touching this code

abmerop (commented):
It looks like this was there before the atomic log.  It seems unintuitive that a store would need to set data on the return path, but I am not sure why it was there before.  I suppose if none of the tests break it is fine.

src/mem/ruby/system/GPUCoalescer.cc
mattsinc: agree, this seems really bad and we should fix ASAP

mattsinc (approved):

mattsinc:
> It looks like this was there before the atomic log. It seems unintuitive that a store would need to set data on the return path, but I am not sure why it was there before. I suppose if none of the tests break it is fine.

Maybe a leftover from the backing store?

abmerop (approved):

abmerop:
> > It looks like this was there before the atomic log. It seems unintuitive that a store would need to set data on the return path, but I am not sure why it was there before. I suppose if none of the tests break it is fine.
> 
> Maybe a leftover from the backing store?

That'd be my guess too.

PR #570, opened 2023-11-16
dev-amdgpu: Add VMID map to checkpoint

v-ramadas:
When restoring checkpoints for certain applications, gem5 tries to create new doorbells with a pre-existing queue ID and simulation crashes shortly after. This commit adds existing IDs to the GPU device's used VMID map so that new doorbells are aware of existing queue IDs and use a new ID. This ensures that queue IDs are unique after checkpoint restoration

Change-Id: I9bf89a2769db26ceab4441634ff2da936eea6d6f

### DIFF
--- a/src/dev/amdgpu/amdgpu_device.cc
+++ b/src/dev/amdgpu/amdgpu_device.cc
@@ -678,10 +678,13 @@ AMDGPUDevice::serialize(CheckpointOut &cp) const
     uint64_t regs_size = regs.size();
     uint64_t doorbells_size = doorbells.size();
     uint64_t sdma_engs_size = sdmaEngs.size();
+    uint64_t used_vmid_map_size = usedVMIDs.size();
 
     SERIALIZE_SCALAR(regs_size);
     SERIALIZE_SCALAR(doorbells_size);
     SERIALIZE_SCALAR(sdma_engs_size);
+    // Save the number of vmids used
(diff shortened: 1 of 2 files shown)


powerjg (commented):
+1 from me, but I'll let the GPU experts sign off.

abmerop (commented):
Could we instead checkpoint the usedVMIDs map in amdgpu_device?  This works for now because we only support 1 VMID, but it would be more future-proof to checkpoint the map in the amdgpu_device::serialize / unserialize in case some day we add support for multiple VMIDs (essentially multiple GPU processes)

v-ramadas:
I will try that out. Thanks for the idea @abmerop!

mattsinc:
> Could we instead checkpoint the usedVMIDs map in amdgpu_device? This works for now because we only support 1 VMID, but it would be more future-proof to checkpoint the map in the amdgpu_device::serialize / unserialize in case some day we add support for multiple VMIDs (essentially multiple GPU processes)

This is a good idea

v-ramadas:
Hi @abmerop, @mattsinc, I added the VMID map to the checkpoint and committed this change. I tested this out with Lulesh and FW successfully. Please let me know if the changes look good.

abmerop (approved):

mattsinc (commented):
Sorry for the delayed reply, I am slowly going through my inbox after the ISCA deadline.  Looks good, thanks @v-ramadas !

PR #571, opened 2023-11-17
util: Bump GPUFS build docker to 5.4.2

abmerop:
This dockerfile is used to *build* applications (e.g., from gem5-resources) which can be run using full system mode in a GPU build. The next releases disk image will use ROCm 5.4.2, therefore bump the version from 4.2 to that version.

Again this is used to *build* input applications only and is not needed to run or compile gem5 with GPUFS. For example:

$ docker build -t rocm54-build .
/some/gem5-resources/src/gpu/lulesh$ docker run --rm -u $UID:$GID -v \
    ${PWD}:${PWD} -w ${PWD} rocm54-build make

Change-Id: If169c8d433afb3044f9b88e883ff3bb2f4bc70d2

### DIFF
--- a/util/dockerfiles/gpu-fs/Dockerfile
+++ b/util/dockerfiles/gpu-fs/Dockerfile
@@ -44,12 +44,12 @@ RUN apt -y install wget gnupg2 rpm
 # Get the radeon gpg key for apt repository
 RUN wget -q -O - https://repo.radeon.com/rocm/rocm.gpg.key | apt-key add -
 
-# Modify apt sources to pull from ROCm 4.2 repository only
-RUN echo 'deb [arch=amd64] https://repo.radeon.com/rocm/apt/4.2/ ubuntu main' | tee /etc/apt/sources.list.d/rocm.list
+# Modify apt sources to pull from ROCm 5.4.2 repository only
+RUN echo 'deb [arch=amd64] https://repo.radeon.com/rocm/apt/5.4.2/ ubuntu main' | tee /etc/apt/sources.list.d/rocm.list
 
 RUN apt-get update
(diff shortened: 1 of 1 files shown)


util/dockerfiles/gpu-fs/Dockerfile
```diff
@@ -27,7 +27,7 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-FROM --platform=${BUILDPLATFORM} ubuntu:20.04
+FROM ubuntu:20.04
```
powerjg: Why this change? The old version would, in theory, allow this to build Arm binaries as well as x86 binaries.

I know that it's unlikely that someone would want to build Arm binaries for the AMD GPU, but it doesn't hurt to have the old version, from what I understand.

abmerop: Ah ok.  I removed it because I didn't know what it does, but I'll add it back

powerjg (commented):
lgtm. Thanks for the update. I'll let Bobby sign off so he can rebuild and push the new image.

mattsinc (approved):

abmerop:
Thanks Jason. FYI will be my last GPUFS related patch before v23.1.  Although I may write some documentation stuff on the other repos :)

powerjg:
@BobbyRBruce want to make sure you see this to rebuild the image

BobbyRBruce:
> @BobbyRBruce want to make sure you see this to rebuild the image

It's rebuilt and uploaded.

PR #573, opened 2023-11-19
arch-riscv: Add PCEvent for RISCV FS Workload kernel panic/oops

### LINKED ISSUE #550: gem5's Linux printk helpers no longer work when using Linux kernel v5.18+ in the guest system
**Describe the bug**
Mismatch between the gem5's DmesgRingBuffer's `struct metadata` and the corresponding `struct prb_desc_ring` in Linux kernel v5.18+.

**Affects version**
Current develop (https://github.com/gem5/gem5/tree/1204267fd8dc2076b4f3b6288a0040116017193d).

**gem5 Modifications**
N/A

**To Reproduce**
Induce a kernel panic where a PCEvent is added for the `panic()` function in the guest.

**Terminal Output**

```shell
458840683000: global: PC based event serviced at 0xffffffff8088c8b4: Kernel panic in simulated kernel                            
src/kern/linux/events.cc:74: info: Dumping kernel dmesg buffer to board.workload.dmesg...                                        
src/mem/port_proxy.hh:185: fatal: readBlob(0x11, ...) failed  
```

**Additional information**
gem5's DmesgRingBuffer's `struct metadata` is used to interpret the `struct prb_desc_ring` in Linux kernel.

gem5's DmesgRingBuffer's `struct metadata` has 5 fields and is 40-byte long. [1]
In Linux kernel v5.10-v5.17, the corresponding `struct prb_desc_ring` in Linux kernel has 5 fields and is 40-byte long. [2]
However, since v5.18, the struct has a new field, which makes the struct 48-byte long. [3]

Adding a new field to the metadata struct in gem5 fix the problem for new kernel versions.

[1] https://github.com/gem5/gem5/blob/1204267fd8dc2076b4f3b6288a0040116017193d/src/kern/linux/helpers.cc#L288-L296
[2] https://elixir.bootlin.com/linux/v5.17.15/source/kernel/printk/printk_ringbuffer.h#L71
[3] https://elixir.bootlin.com/linux/v5.18.19/source/kernel/printk/printk_ringbuffer.h#L71

hnpl:
Inspired by the similar feature in ARM's full system workload, this change adds
an option to halt gem5 simulation if the guest system encounter kernel panic
or kernel oops.

On RiscvISA::BootloaderKernelWorkload, by default, the simulation
will exit upon kernel panic, while kernel oops will not induce simulation halt.
This is because the system will essentially do nop after a kernel panic, while the
system might be still functional after a kernel oops.

Dumping kernel's dmesg is useful for diagonizing the cause of kernel panic, so
ideally, we want to dump the guest's dmesg to the host. However, due to a bug
described in [1], kernel v5.18+ dmesg might not be dumped properly. Hence, the
dmesg will not be dumped to the host.

On RiscvISA::FsLinux, this feature is turned off by default as the symbols from the
official RISC-V kernel resource are stripped from the binary. However, if this feature
is enable, the dmesg will be dumped to the host system.

[1] https://github.com/gem5/gem5/issues/550

Change-Id: I8f52257727a3a789ebf99fdd4dffe5b3d89f1ebf

### DIFF
--- a/src/arch/riscv/RiscvFsWorkload.py
+++ b/src/arch/riscv/RiscvFsWorkload.py
@@ -55,6 +55,16 @@ class RiscvLinux(KernelWorkload):
     )
     dtb_addr = Param.Addr(0x87E00000, "DTB address")
 
+    # gem5 event upon guest's kernel panic
+    # Default to false because when the kernel is compiled into the bootloader
+    # it will not have symbols
+    exit_on_kernel_panic = Param.Bool(
+        False, "Generate gem5 panic upon the guest's kernel panic."
+    )
+    exit_on_kernel_oops = Param.Bool(
+        False, "Generate gem5 panic upon the guest's kernel oops."
+    )
+
 
 class RiscvBootloaderKernelWorkload(Workload):
     type = "RiscvBootloaderKernelWorkload"
@@ -84,5 +94,28 @@ class RiscvBootloaderKernelWorkload(Workload):
 
     # booting parameters
     command_line = Param.String(
-        "", "Booting arguments, to be passed to the kernel"
+        "", "Booting arguments, to be passed to the kernel."
+    )
+
+    # gem5 event upon guest's kernel panic
+    # Note that if the kernel doesn't have symbols there will be a warning and
+    # gem5 will not exit
+    exit_on_kernel_panic = Param.Bool(
+        True, "Generate gem5 exit upon the guest's kernel panic."
+    )
+    exit_on_kernel_oops = Param.Bool(
+        False, "Generate gem5 exit upon the guest's kernel oops."
+    )
+
+    # Note: Duplicated from KernelWorkload for now
+    on_panic = Param.KernelPanicOopsBehaviour(
+        "DumpDmesgAndExit",
+        "Define how gem5 should behave after a Linux Kernel Panic. "
+        "Handler might not be implemented for all architectures.",
+    )
+
+    on_oops = Param.KernelPanicOopsBehaviour(
+        "DumpDmesgAndExit",
+        "Define how gem5 should behave after a Linux Kernel Oops. "
+        "Handler might not be implemented for all architectures.",
     )
--- a/src/arch/riscv/linux/fs_workload.cc
+++ b/src/arch/riscv/linux/fs_workload.cc
@@ -32,7 +32,10 @@
 #include "base/loader/dtb_file.hh"
 #include "base/loader/object_file.hh"
 #include "base/loader/symtab.hh"
+#include "cpu/pc_event.hh"
+#include "kern/linux/events.hh"
 #include "sim/kernel_workload.hh"
+#include "sim/sim_exit.hh"
 #include "sim/system.hh"
 
 namespace gem5
@@ -72,6 +75,42 @@ FsLinux::initState()
     for (auto *tc: system->threads) {
         RiscvISA::Reset().invoke(tc);
         tc->activate();
+    }
+}
+
+void
+FsLinux::startup()
+{
+    KernelWorkload::startup();
+
+    addExitOnKernelPanicEvent();
+    addExitOnKernelOopsEvent();
+}
+
+void
+FsLinux::addExitOnKernelPanicEvent()
+{
+    const std::string dmesg_output = name() + ".dmesg";
+    if (params().exit_on_kernel_panic) {
+        kernelPanicPcEvent = addKernelFuncEvent<linux::PanicOrOopsEvent>(
+            "panic", "Kernel panic in simulated system.",
+            dmesg_output, params().on_panic
+        );
+        warn_if(!kernelPanicPcEvent, "Failed to find kernel symbol 'panic'");
+    }
+}
+
+void
+FsLinux::addExitOnKernelOopsEvent()
+{
+    const std::string dmesg_output = name() + ".dmesg";
+    if (params().exit_on_kernel_oops) {
+        kernelOopsPcEvent = addKernelFuncEvent<linux::PanicOrOopsEvent>(
+            "oops_exit", "Kernel oops in simulated system.",
+            dmesg_output, params().on_oops
+        );
+        warn_if(!kernelOopsPcEvent,
+                "Failed to find kernel symbol 'oops_exit'");
     }
 }
 
@@ -170,6 +209,30 @@ BootloaderKernelWorkload::loadDtb()
 }
 
 void
+BootloaderKernelWorkload::addExitOnKernelPanicEvent()
+{
+    const std::string dmesg_output = name() + ".dmesg";
+    if (params().exit_on_kernel_panic) {
+        kernelPanicPcEvent = addFuncEvent<linux::PanicOrOopsEvent>(
+            kernelSymbolTable, "panic", "Kernel panic in simulated system.",
+            dmesg_output, params().on_panic
+        );
+    }
+}
+
+void
+BootloaderKernelWorkload::addExitOnKernelOopsEvent()
+{
+    const std::string dmesg_output = name() + ".dmesg";
+    if (params().exit_on_kernel_oops) {
+        kernelOopsPcEvent = addFuncEvent<linux::PanicOrOopsEvent>(
+            kernelSymbolTable, "oops_exit", "Kernel oops in simulated system.",
+            dmesg_output, params().on_oops
+        );
+    }
+}
+
+void
 BootloaderKernelWorkload::initState()
 {
     loadBootloader();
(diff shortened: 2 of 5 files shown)


BobbyRBruce (dismissed):
Giving a soft-approval. I'd like someone with a bit more RISC-V Knowledge to have a look at it.

BobbyRBruce:
@rjc-arch @rogerchang23424 : Could one of you see if this looks right to you?

rogerchang23424 (commented):
It's OK to me.

src/arch/riscv/linux/fs_workload.cc
```diff
@@ -196,5 +260,11 @@ BootloaderKernelWorkload::unserialize(CheckpointIn &checkpoint)
     kernelSymbolTable.unserialize("kernel_symbol_table", checkpoint);
 }
 
+void
+Gem5Exit::process(ThreadContext *tc)
+{
+    fatal(descr());
```
powerjg: I think this should be a `panic` since it's not an incorrect configuration. 

Really, I'm not sure if it should be a `panic` *or* a `fatal`. What is this purpose of this event?

src/arch/riscv/linux/fs_workload.hh
```diff
@@ -44,11 +44,32 @@ namespace RiscvISA
...
+    ~FsLinux()
+    {
+        if (kernelPanicPcEvent != nullptr)
+        {
+            delete kernelPanicPcEvent;
+        }
+        if (kernelOopsPcEvent != nullptr)
+        {
+            delete kernelOopsPcEvent;
+        }
```
powerjg: This is not to style.

```suggestion
        if (kernelPanicPcEvent != nullptr) {
            delete kernelPanicPcEvent;
        }
        if (kernelOopsPcEvent != nullptr) {
            delete kernelOopsPcEvent;
        }
```

src/arch/riscv/linux/fs_workload.hh
```diff
@@ -87,7 +117,20 @@ class BootloaderKernelWorkload: public Workload
...
+    ~BootloaderKernelWorkload()
+    {
+        if (kernelPanicPcEvent != nullptr)
+        {
+            delete kernelPanicPcEvent;
+        }
+        if (kernelOopsPcEvent != nullptr)
+        {
+            delete kernelOopsPcEvent;
+        }
```
powerjg: Also not to style.

```suggestion
        if (kernelPanicPcEvent != nullptr) {
            delete kernelPanicPcEvent;
        }
        if (kernelOopsPcEvent != nullptr) {
            delete kernelOopsPcEvent;
        }
```

src/arch/riscv/linux/fs_workload.cc
hnpl: I think the best way to describe this event is to exit gem5 when there's a kernel panic.

I leaned towards using `fatal` because `panic` implies there's something with gem5. A kernel panic could also be a product of a wrong configuration, e.g., using a wrong boot parameter.

Another reason is that `panic` dumps the backtrace while `fatal` does not. The backtrace might be misleading as the exit message is pushed further upward.

Anyway, if `fatal` means there __must__ be something wrong with the configuration then I'll switch to use `panic`.

powerjg: If the goal is to exit the gem5 simulation loop, then you should call `exitSimLoop` (this isn't the exact name of the function, but it's something like that). I agree that neither `panic` nor `fatal` is right in this case.

rjc-arch (commented):
Hi @hnpl. The generic parts of this code look good as far as I can see.

src/arch/riscv/linux/fs_workload.cc
rjc-arch: +1 it's `exitSimLoop` from `sim/sim_exit.hh`.

hnpl: Thanks for the suggestions. I updated the exit event accordingly.

I want to note that, for ArmFsWorkloads, guest's kernel panic will cause a gem5 panic, so we have different kernel panic exit behaviors for workloads of different ISAs. This might be confusing to users.

src/arch/riscv/RiscvFsWorkload.py
```diff
@@ -84,5 +92,13 @@ class RiscvBootloaderKernelWorkload(Workload):
 
     # booting parameters
     boot_args = Param.String(
-        "", "Booting arguments, to be passed to the kernel"
+        "", "Booting arguments, to be passed to the kernel."
+    )
+
+    # gem5 event upon guest's kernel panic
+    exit_on_kernel_panic = Param.Bool(
+        True, "Generate gem5 exit upon the guest's kernel panic."
```
powerjg, line 104: Why is this True when it is False in the RiscvFsWorkload.py

powerjg (dismissed):

hnpl:
Current linux kernel from gem5 resources does not have symbol. This exit on
kernel panic behavior relies on being able to find the panic function
symbol. Default this to true on RiscvFsWorkload will crash the simulation.

powerjg:
Hmm. I think maybe we should always default to false and only set it to true if we know the resource has symbols. I'm not sure how to do that since the resource may not have that information easily available or have the API to make that update.

src/arch/riscv/linux/fs_workload.cc
```diff
@@ -75,6 +78,39 @@ FsLinux::initState()
...
+    addExitOnKernelPanicEvent();
+    addExitOnKernelOopsEvent();
+}
+
+void
+FsLinux::addExitOnKernelPanicEvent()
+{
+    const std::string dmesg_output = name() + ".dmesg";
+    if (params().exit_on_kernel_panic) {
+        kernelPanicPcEvent = addFuncEventOrPanic<linux::KernelPanic>(
```
powerjg: Use a different version of this function so that it doesn't panic if it can't find the symbol. Instead, we want to have a warning.

rjc-arch:
From a quick glance at the code it looks like we could use `addFuncEvent` instead of `addFuncEventOrPanic` to add the events, and then manually print a warning instead of panicking if `nullptr` is returned (in the case that the `panic`/`oops_exit` symbols are not found).

I'm making some changes to the Arm Panic/Oops handling at the moment related to #550; I've just tried making the same change there, and it seems to work. (I'm not sure if you want to wait to see what I have done before merging this change?)

(@powerjg You just got in ahead of me!)

powerjg:
I am going to take over this patch. Yeah, `addFuncEvent` is what I was thinking. When you push a patch for #550 I'll follow the same format.

@rjc-arch, do you think you'll have a patch for #550 by tomorrow?

rjc-arch:
I hope so. I'm just responding to some feedback from @andysan before pushing.

rjc-arch:
Hi @powerjg. I've just pushed #639 if you'd like to take a look. I'm not sure how well it fits in with what you were planning. I suspect we can hoist the exit events fully into the common code now that adding the events fails gracefully.

powerjg:
Yes, that looks great @rjc-arch! 

I think this change is going to need a bit more time, though. I need to understand why `BootloaderKernelWorkload` is not a subclass of `KernelWorkload`, which is preventing us from hoisting the duplicate code right now.

After #639 is merged, this should be rebased on top of that change (possibly with hoisting common code).

src/arch/riscv/linux/fs_workload.hh
```diff
@@ -87,7 +115,18 @@ class BootloaderKernelWorkload: public Workload
...
     }
 
+    ~BootloaderKernelWorkload()
+    {
+        if (kernelPanicPcEvent != nullptr) {
+            delete kernelPanicPcEvent;
+        }
+        if (kernelOopsPcEvent != nullptr) {
+            delete kernelOopsPcEvent;
+        }
```
BobbyRBruce, line 125: Fyi, you can safely "delete" `nullptr`. It's part of the CPP standard. It does nothing.

(Not a request for change. This is fine, just overly-verbose).

BobbyRBruce (commented):
I'm not sure what the intended affect is here but as currently written the exit events will do nothing by default (just continue the simulation loop). If you want a different default behavior (such as exit the simulation like on the `ExitEvent.EXIT`) you need to update this Dictionary: https://github.com/gem5/gem5/blob/569e21f798f61c22a91e8f6bad083a82fdf43034/src/python/gem5/simulate/simulator.py#L275C14-L275C35

powerjg:
Thanks, Bobby. I've updated the default dictionary.

powerjg:
@BobbyRBruce I have verified this works. Can you please re-review?

BobbyRBruce (approved):

PR #530, opened 2023-11-03
mem-ruby, gpu-compute: update GPU L1I$ MRU info

mattsinc:
Previously the GPU L1 I$ (SQC) was not updating the MRU information on hits in the SQC.  This commit resolves that by adding support to the appropriate Ruby transition.

Change-Id: I2be0995a8666ed0d4425074a1cb066451a1cc6e6

PR #581, opened 2023-11-20
mem-cache: Revert "Prefetchers Improvements"

### LINKED ISSUE #580: `Assertion 'flags.isSet(STATIC_DATA|DYNAMIC_DATA)' failed` encountered when running "configs/example/gem5_library/riscv-ubuntu-run.py" [gem5 Library Example Daily Test Failure]
**Describe the bug**

The Daily gem5 Library RISCV Ubuntu Run test is failing: https://github.com/gem5/gem5/actions/runs/6901014917/job/18783090892

This test runs gem5 using the [configs/example/gem5_library/riscv-ubuntu-run.py](https://github.com/gem5/gem5/blob/3896673ddc943cdf2157651d090920da466bc6a7/configs/example/gem5_library/riscv-ubuntu-run.py) configuration file. It produces the following error:

```sh
gem5.opt: src/mem/packet.hh:1236: const T* gem5::Packet::getConstPtr() const [with T = unsigned char]: Assertion `flags.isSet(STATIC_DATA|DYNAMIC_DATA)' failed.
```

**Revisions affected and candidate breaking commits**

This was seen on the develop branch. The earliest known commit to have this bug is db6a8697864564aab1a11081edc05152c0478f83, found in this failing Daily Test: https://github.com/gem5/gem5/actions/runs/6901014917/job/18783090892.

The last commit known to not to contain this bug is ceabe86b311e3bdd7603b4e4631d2ed18365a5bd.

Based on a basic `git rev-log` of this information (`git rev-list --oneline ceabe86..db6a869`) the candidate breaking commits are:

* db6a869786 mem-cache: Prefetchers Improvements (#564)
* 4ca2efac16 mem-ruby: AtomicNoReturn should check comp_anr instead of comp_wu (#545)
* 4965367724 mem-ruby, gpu-compute: fix SQC/TCP requests to same line (#540)
* bfe899e48e stdlib, resources: Update JSON data in workload (#532)
* 047a494c2b mem-cache: Optimize strided prefetcher address generation
* 2abd65c270 mem: added distance parameter to stride prefetcher
* c3326c78e6 mem-ruby, gpu-compute: fix SQC/TCP requests to same line
* 065ddf759f mem-ruby, gpu-compute: fix bug with GPU bypassing loads
* 38045d7a25 mem-cache: Added clean eviction check for prefetchers.
* 6416304e07 mem-cache: Update default prefetch options.
* 8598764a03 mem-cache: Squash prefetch queue entries by block address.
* 50c9cbf613 stdlib, resources: Fixed deprecation warning

**To Reproduce**

```sh
scons build/ALL/gem5.opt -j`nproc`
./build/ALL/gem5.opt configs/example/gem5_library/riscv-ubuntu-run.py
```

**Terminal Output**

```sh
gem5.opt: src/mem/packet.hh:1236: const T* gem5::Packet::getConstPtr() const [with T = unsigned char]: Assertion `flags.isSet(STATIC_DATA|DYNAMIC_DATA)' failed.
Program aborted at tick 5261728243428
--- BEGIN LIBC BACKTRACE ---
/__w/gem5/gem5/build/ALL/gem5.opt(_ZN4gem515print_backtraceEv+0x30)[0x56239b912010]
/__w/gem5/gem5/build/ALL/gem5.opt(_ZN4gem512abortHandlerEi+0x4c)[0x56239b936afc]
/lib/x86_64-linux-gnu/libc.so.6(+0x42520)[0x7f207d30b520]
/lib/x86_64-linux-gnu/libc.so.6(pthread_kill+0x12c)[0x7f207d35fa7c]
/lib/x86_64-linux-gnu/libc.so.6(raise+0x16)[0x7f207d30b476]
/lib/x86_64-linux-gnu/libc.so.6(abort+0xd3)[0x7f207d2f17f3]
/lib/x86_64-linux-gnu/libc.so.6(+0x2871b)[0x7f207d2f171b]
/lib/x86_64-linux-gnu/libc.so.6(+0x39e96)[0x7f207d302e96]
/__w/gem5/gem5/build/ALL/gem5.opt(_ZN4gem58prefetch4Base12PrefetchInfoC1EPNS_6PacketEmb+0x17b)[0x56239bd2263b]
/__w/gem5/gem5/build/ALL/gem5.opt(_ZN4gem58prefetch4Base16PrefetchListener6notifyERKPNS_6PacketE+0x19a)[0x56239bd22dfa]
/__w/gem5/gem5/build/ALL/gem5.opt(_ZN4gem59BaseCache13recvTimingReqEPNS_6PacketE+0x1f1)[0x56239bc91161]
/__w/gem5/gem5/build/ALL/gem5.opt(_ZThn72_N4gem59BaseCache11CpuSidePort13recvTimingReqEPNS_6PacketE+0x8a)[0x56239bc8c0fa]
/__w/gem5/gem5/build/ALL/gem5.opt(_ZN4gem512CoherentXBar13recvTimingReqEPNS_6PacketEs+0x4a4)[0x56239bb564a4]
/__w/gem5/gem5/build/ALL/gem5.opt(_ZN4gem59BaseCache19sendMSHRQueuePacketEPNS_4MSHRE+0x202)[0x56239bca22e2]
/__w/gem5/gem5/build/ALL/gem5.opt(_ZN4gem59BaseCache19CacheReqPacketQueue18sendDeferredPacketEv+0x101)[0x56239bc90cc1]
/__w/gem5/gem5/build/ALL/gem5.opt(_ZN4gem510EventQueue10serviceOneEv+0xc2)[0x56239b926dc2]
/__w/gem5/gem5/build/ALL/gem5.opt(_ZN4gem59doSimLoopEPNS_10EventQueueE+0x68)[0x56239b94e208]
/__w/gem5/gem5/build/ALL/gem5.opt(_ZN4gem58simulateEm+0x27b)[0x56239b94e7fb]
/__w/gem5/gem5/build/ALL/gem5.opt(+0x2616830)[0x562399704830]
/__w/gem5/gem5/build/ALL/gem5.opt(+0x1db2764)[0x562398ea0764]
/lib/x86_64-linux-gnu/libpython3.10.so.1.0(+0x12b6d3)[0x7f207e2516d3]
/lib/x86_64-linux-gnu/libpython3.10.so.1.0(_PyObject_Call+0x5c)[0x7f207e20d22c]
/lib/x86_64-linux-gnu/libpython3.10.so.1.0(_PyEval_EvalFrameDefault+0x4b26)[0x7f207e19c766]
/lib/x86_64-linux-gnu/libpython3.10.so.1.0(+0x1c681f)[0x7f207e2ec81f]
/lib/x86_64-linux-gnu/libpython3.10.so.1.0(_PyEval_EvalFrameDefault+0x9d78)[0x7f207e1a19b8]
/lib/x86_64-linux-gnu/libpython3.10.so.1.0(+0x1c681f)[0x7f207e2ec81f]
/lib/x86_64-linux-gnu/libpython3.10.so.1.0(_PyEval_EvalFrameDefault+0x82ae)[0x7f207e19feee]
/lib/x86_64-linux-gnu/libpython3.10.so.1.0(+0x1c681f)[0x7f207e2ec81f]
/lib/x86_64-linux-gnu/libpython3.10.so.1.0(PyEval_EvalCode+0xbe)[0x7f207e2e794e]
/lib/x86_64-linux-gnu/libpython3.10.so.1.0(+0x1c1edd)[0x7f207e2e7edd]
/lib/x86_64-linux-gnu/libpython3.10.so.1.0(+0x12eb13)[0x7f207e254b13]
/lib/x86_64-linux-gnu/libpython3.10.so.1.0(_PyEval_EvalFrameDefault+0x69ee)[0x7f207e19e62e]
--- END LIBC BACKTRACE ---
For more info on how to address this issue, please visit https://www.gem5.org/documentation/general_docs/common-errors/ 
```

**Host Operating System**

This was run in a docker image ([ghcr.io/gem5/ubuntu-22.04_all-dependencies:latest](ghcr.io/gem5/ubuntu-22.04_all-dependencies:latest)) on top of an Ubuntu 22.04 OS.

**Host ISA**

X86

BobbyRBruce:
Reverts gem5/gem5#564 to fix #580.

Exactly what in #530 is causing the bug is unknown.

### DIFF
--- a/src/mem/cache/Cache.py
+++ b/src/mem/cache/Cache.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2012-2013, 2015, 2018, 2022 Arm Limited
+# Copyright (c) 2012-2013, 2015, 2018 ARM Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
@@ -112,7 +112,7 @@ class BaseCache(ClockedObject):
         "Notify the hardware prefetcher on every access (not just misses)",
     )
     prefetch_on_pf_hit = Param.Bool(
-        True, "Notify the hardware prefetcher on hit on prefetched lines"
+        False, "Notify the hardware prefetcher on hit on prefetched lines"
     )
 
     tags = Param.BaseTags(BaseSetAssoc(), "Tag store")
--- a/src/mem/cache/prefetch/Prefetcher.py
+++ b/src/mem/cache/prefetch/Prefetcher.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2012, 2014, 2019, 2022 Arm Limited
+# Copyright (c) 2012, 2014, 2019 ARM Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
@@ -192,13 +192,6 @@ class StridePrefetcher(QueuedPrefetcher):
     use_requestor_id = Param.Bool(True, "Use requestor id based history")
 
     degree = Param.Int(4, "Number of prefetches to generate")
-    distance = Param.Unsigned(
-        0,
-        "How far ahead of the demand stream to start prefetching. "
-        "Skip this number of strides ahead of the first identified prefetch, "
-        "then generate `degree` prefetches at `stride` intervals. "
-        "A value of zero indicates no skip.",
-    )
 
     table_assoc = Param.Int(4, "Associativity of the PC table")
     table_entries = Param.MemorySize("64", "Number of entries of the PC table")
--- a/src/mem/cache/prefetch/base.cc
+++ b/src/mem/cache/prefetch/base.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2014, 2022 Arm Limited
+ * Copyright (c) 2013-2014 ARM Limited
  * All rights reserved.
  *
  * The license below extends only to copyright in the software and shall
@@ -245,7 +245,6 @@ Base::probeNotify(const PacketPtr &pkt, bool miss)
     // operations or for writes that we are coaslescing.
     if (pkt->cmd.isSWPrefetch()) return;
     if (pkt->req->isCacheMaintenance()) return;
-    if (pkt->isCleanEviction()) return;
     if (pkt->isWrite() && cache != nullptr && cache->coalesce()) return;
     if (!pkt->req->hasPaddr()) {
         panic("Request must have a physical address");
--- a/src/mem/cache/prefetch/queued.cc
+++ b/src/mem/cache/prefetch/queued.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014-2015, 2022 Arm Limited
+ * Copyright (c) 2014-2015 ARM Limited
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
@@ -178,7 +178,7 @@ Queued::notify(const PacketPtr &pkt, const PrefetchInfo &pfi)
     if (queueSquash) {
         auto itr = pfq.begin();
         while (itr != pfq.end()) {
-            if (blockAddress(itr->pfInfo.getAddr()) == blk_addr &&
+            if (itr->pfInfo.getAddr() == blk_addr &&
                 itr->pfInfo.isSecure() == is_secure) {
                 DPRINTF(HWPrefetch, "Removing pf candidate addr: %#x "
                         "(cl: %#x), demand request going to the same addr\n",
--- a/src/mem/cache/prefetch/stride.cc
+++ b/src/mem/cache/prefetch/stride.cc
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2018 Inria
- * Copyright (c) 2012-2013, 2015, 2022-2023 Arm Limited
+ * Copyright (c) 2012-2013, 2015 ARM Limited
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
@@ -84,7 +84,6 @@ Stride::Stride(const StridePrefetcherParams &p)
     threshConf(p.confidence_threshold/100.0),
     useRequestorId(p.use_requestor_id),
     degree(p.degree),
-    distance(p.distance),
     pcTableInfo(p.table_assoc, p.table_entries, p.table_indexing_policy,
         p.table_replacement_policy)
 {
@@ -168,16 +167,16 @@ Stride::calculatePrefetch(const PrefetchInfo &pfi,
             return;
         }
 
-        // Round strides up to atleast 1 cacheline
-        int prefetch_stride = new_stride;
-        if (abs(new_stride) < blkSize) {
-            prefetch_stride = (new_stride < 0) ? -blkSize : blkSize;
-        }
-
-        Addr new_addr = pf_addr + distance * prefetch_stride;
         // Generate up to degree prefetches
         for (int d = 1; d <= degree; d++) {
-            addresses.push_back(AddrPriority(new_addr += prefetch_stride, 0));
+            // Round strides up to atleast 1 cacheline
+            int prefetch_stride = new_stride;
+            if (abs(new_stride) < blkSize) {
+                prefetch_stride = (new_stride < 0) ? -blkSize : blkSize;
+            }
+
+            Addr new_addr = pf_addr + d * prefetch_stride;
+            addresses.push_back(AddrPriority(new_addr, 0));
         }
     } else {
         // Miss in table
--- a/src/mem/cache/prefetch/stride.hh
+++ b/src/mem/cache/prefetch/stride.hh
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2018 Inria
- * Copyright (c) 2012-2013, 2015, 2022 Arm Limited
+ * Copyright (c) 2012-2013, 2015 ARM Limited
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
@@ -104,8 +104,6 @@ class Stride : public Queued
     const bool useRequestorId;
 
     const int degree;
-
-    const int distance;
 
     /**
      * Information used to create a new PC table. All of them behave equally.


BobbyRBruce:
~~**Note**: Currently testing to see if this revert fixes #580 which I believe was caused by #564.~~

~~Not ready for review.~~

BobbyRBruce:
I've confirmed this revert fixes the issue described in #580. I won't rush in (i.e., force it through over the next day) to give us time to come up with a better fix. However, we do need #580 resolved soon for our v23.1 release. I also don't want our Daily tests failing for very long. They has a tendency to get away from us when we let them fail for too long.

src/mem/cache/Cache.py
```diff
@@ -112,7 +112,7 @@ class BaseCache(ClockedObject):
         "Notify the hardware prefetcher on every access (not just misses)",
     )
     prefetch_on_pf_hit = Param.Bool(
-        True, "Notify the hardware prefetcher on hit on prefetched lines"
+        False, "Notify the hardware prefetcher on hit on prefetched lines"
```
rogerchang23424, line 115: I have test the `riscv-ubuntu-run.py` in the local machine and it works well by change the default value of `prefetch_on_pf_hit` to `False`. Cloud you just simply change it?

rogerchang23424 (changes requested):

src/mem/cache/Cache.py
andysan: This sounds plausible. @rjc-arch is looking into this and trying to understand what's going on here. Our current theory is that there is  bug triggered by `prefetch_on_pf_hit` related to which packets the prefetcher processes. It seems like the problem might be caused by a request received through the `ppHit` listener which the prefetcher thinks should have data but doesn't. My guess is that the hardware prefetcher doesn't filter out some cache maintenance (or similar) operation correctly.

I would suggest that we either revert the specific change that enables `prefetch_on_pf_hit` or wait for @rjc-arch to produce a patch.

rjc-arch:
Hi All - my apologies that this change has caused problems!

I agree it is likely that it's the `prefetch_on_pf_hit` that is causing the problem. Unfortunately, this is the setting that makes the most difference for prefetcher efficacy; without this setting the prefetcher will trigger, generate a sequence of prefetches, then lose the pattern again as the demand stream hits the prefetches. The other rationale for having this parameter switched on by default is that it is located in the cache model, and therefore easy to overlook when working with prefetchers.

Having said that, if this default is causing problems for some architectures/workloads then I would certainly support reverting, at least until we understand the cause. The other patches in the PR should be fine to keep, so I would suggest following @rogerchang23424's recommendation.

src/mem/cache/Cache.py
powerjg: I *don't* like the idea of just changing `prefetch_on_pf_hit` to false as  then we'll have code that is broken in gem5. I would prefer to go ahead and revert this whole patch if we can't get a fix quickly.

rjc-arch: Hi @powerjg. Just to clarify, changing `prefetch_on_pf_hit` to False _would_ be reverting the whole commit (but preserving the other commits in the pull request, which are on the same theme but separate).

Having said that, we do suspect the parameter `prefetch_on_pf_hit = True` is exposing a separate bug or edge case in the prefetcher code, rather than being a bug itself, so I agree we need to look into the underlying cause as well.

rogerchang23424 (commented):
Change is not required now

src/mem/cache/Cache.py
andysan: Or put differently, reverting all of the changes in the PR won't remove underlying broken code, it'll just disable it by setting `prefetch_on_pf_hit` to false again.

rogerchang23424: Hi, I have trace the `riscv-ubuntu-run.py` with `--debug-flags=Cache`, the following is the output:
```
5261728243428: board.cache_hierarchy.l1dcaches0: sendMSHRQueuePacket: MSHR WriteReq [1078fb2e0:1078fb2e7]
5261728243428: board.cache_hierarchy.l1dcaches0: createMissPacket: created UpgradeReq [1078fb2c0:1078fb2ff] from WriteReq [1078fb2e0:1078fb2e7]
5261728243428: board.cache_hierarchy.dptw_caches0: handleSnoop: snoop hit for UpgradeReq [1078fb2c0:1078fb2ff], old state is state: 6e616324 (S) writable: 0 readable: 1 dirty: 0 prefetched: 0 | tag: 0x20f1f6 secure: 0 valid: 1 | set: 0xb way: 0
5261728243428: board.cache_hierarchy.dptw_caches0: new state is state: 6e616320 (I) writable: 0 readable: 0 dirty: 0 prefetched: 0 | tag: 0xffffffffffffffff secure: 0 valid: 0 | set: 0xb way: 0
5261728243428: board.cache_hierarchy.l2caches0: access for UpgradeReq [1078fb2c0:1078fb2ff] hit state: 6 (E) writable: 1 readable: 1 dirty: 0 prefetched: 1 | tag: 0x41e3e secure: 0 valid: 1 | set: 0xcb way: 0x5
```

The gem5 crashed on the `ppHit->notify` because of the `UpgradeReq` and the cache block is *prefetched*.

BobbyRBruce:
> Hi All - my apologies that this change has caused problems!
> 
> I agree it is likely that it's the `prefetch_on_pf_hit` that is causing the problem. Unfortunately, this is the setting that makes the most difference for prefetcher efficacy; without this setting the prefetcher will trigger, generate a sequence of prefetches, then lose the pattern again as the demand stream hits the prefetches. The other rationale for having this parameter switched on by default is that it is located in the cache model, and therefore easy to overlook when working with prefetchers.
> 
> Having said that, if this default is causing problems for some architectures/workloads then I would certainly support reverting, at least until we understand the cause. The other patches in the PR should be fine to keep, so I would suggest following @rogerchang23424's recommendation.

I will merge this revert PR as it doesn't seem like there's a better fix coming soon. It should be pretty straight-forward to un-revert this when needed. We need this to get our tests passing again and our codebase in a good state for the staging branch creation on the 1st.

src/mem/cache/Cache.py
rjc-arch: Thanks for the extra information, @rogerchang23424. I've only just managed to re-produce this problem now - it looks like the gem5 resource manager was picking up an old disk image which doesn't exhibit the problem from somewhere on my system. After purging all the copies I could find it's now failing with the assertion.

I'm running it up in a debugger now, then I'll take a proper look.

PR #582, opened 2023-11-21
dev: Fix `std::min` type mismatch in reg_bank.hh

BobbyRBruce:
https://github.com/gem5/gem5/pull/386 included two cases in "src/dev/reg_bank.hh" where `std:: min` was used to compare a an integer of type `size_t` and another of type `Addr`. This causes an error on my Apple Silicon Mac as the comparison between an "unsigned long" and an "unsigned long long" is not permitted. To fix this issue this patch changes `reg_size`  from `size_t` to `Addr`, as well as it the types of the values it was derived from and the variable used to hold the return from the `std::min` calls. While not completely correct typing from a labelling perspective (`reg_bytes` is not an address), functions in "src/dev/reg_bank.hh" already abuse `Addr` in this way frequently (for example,  `bytes` in the `write` function).

### DIFF
--- a/src/dev/reg_bank.hh
+++ b/src/dev/reg_bank.hh
@@ -979,9 +979,9 @@ class RegisterBank : public RegisterBankBase
         std::ostringstream ss;
         while (done != bytes) {
           RegisterBase &reg = it->second.get();
-          const size_t reg_off = addr - it->first;
-          const size_t reg_size = reg.size() - reg_off;
-          const size_t reg_bytes = std::min(reg_size, bytes - done);
+          const Addr reg_off = addr - it->first;
+          const Addr reg_size = reg.size() - reg_off;
+          const Addr reg_bytes = std::min(reg_size, bytes - done);
 
           if (reg_bytes != reg.size()) {
               if (_debug_flag) {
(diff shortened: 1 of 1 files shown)


BobbyRBruce:
For the curious, it seems like The C++ stdlib provided by MaxOSX.sdk enforces demands equal type/bitwidth for the integers passed to `std:min`:

```sh
In file included from build/ALL/python/_m5/param_IdeDisk.cc:17:
In file included from src/dev/storage/ide_disk.hh:52:
In file included from src/dev/storage/ide_ctrl.hh:40:
src/dev/reg_bank.hh:984:36: error: no matching function for call to 'min'
          const size_t reg_bytes = std::min(reg_size, bytes - done);
                                   ^~~~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/__algorithm/min.h:40:1: note: candidate template ignored: deduced conflicting types for parameter '_Tp' ('size_t' (aka 'unsigned long') vs. 'Addr' (aka 'unsigned long long'))
min(const _Tp& __a, const _Tp& __b)
```

I'm surprised it doesn't upcast in these cases.

powerjg (approved):
I don't like to use `Addr` this way, but given the rest of the code does use `Addr` as a size, I think this is fine.

PR #584, opened 2023-11-21
arch-arm: Fix Virtual Interrupt logic in secure mode

giactra:
This PR is fixing remaining issues in the ArmISA::Interrupt class; more specifically it is enabling
virtual interrupts in secure mode (when FEAT_SEL2 is present). Previous version was assuming no
virtual interrupt was possible in secure mode. We fix this assumption by replacing the security check
with the EL2Enabled helper which closely matches the Arm pseudocode


rjc-arch (approved):
Looks good to me.

powerjg (approved):

PR #587, opened 2023-11-21
misc: update x86-npb-benchmarks.py to use suites

Harshil2107:
- updated the x86-npb-benchmarks.py to use npb workloads and suites.

The suites and workloads are not in the database are also waiting feedback. I am attaching the JSON file here.
[npb_workloads_suite.json](https://github.com/gem5/gem5/files/13431116/npb_workloads_suite.json)

To run the x86-npb-benchmarks.py script use the GEM5_RESOURCE_JSON_APPEND env variable. The full command is:
```
GEM5_RESOURCE_JSON_APPEND=[path to npb_workloads_suite.json] ./build/X86/gem5.opt configs/example/gem5_library/x86-npb-benchmarks.py --benchmark [benchmark]
```
Change-Id: I248e6452ea4122e9260e34e4368847660edae577

### DIFF
--- a/configs/example/gem5_library/x86-npb-benchmarks.py
+++ b/configs/example/gem5_library/x86-npb-benchmarks.py
@@ -75,8 +75,6 @@ requires(
 
 # Following are the list of benchmark programs for npb.
 
-benchmark_choices = ["bt", "cg", "ep", "ft", "is", "lu", "mg", "sp"]
-
 # We are restricting classes of NPB to A, B and C as the other classes (D and
 # F) require main memory size of more than 3 GB. The X86Board is currently
 # limited to 3 GB of memory. This limitation is explained later in line 136.
@@ -85,12 +83,11 @@ benchmark_choices = ["bt", "cg", "ep", "ft", "is", "lu", "mg", "sp"]
 # works with class D in the current configuration. More information on the
 # memory footprint for NPB is available at https://arxiv.org/abs/2010.13216
 
-size_choices = ["A", "B", "C"]
-
 parser = argparse.ArgumentParser(
     description="An example configuration script to run the npb benchmarks."
 )
 
+npb_suite = obtain_resource("npb-benchmark-suite")
 # The only positional argument accepted is the benchmark name in this script.
 
 parser.add_argument(
@@ -98,15 +95,7 @@ parser.add_argument(
     type=str,
     required=True,
     help="Input the benchmark program to execute.",
-    choices=benchmark_choices,
-)
-
-parser.add_argument(
-    "--size",
-    type=str,
-    required=True,
-    help="Input the class of the program to simulate.",
-    choices=size_choices,
+    choices=[workload.get_id() for workload in npb_suite],
 )
 
 parser.add_argument(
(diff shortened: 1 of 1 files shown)


Harshil2107:
I think we can also directly call `obtain_resource` on the workload id that is provided instead of getting all the workloads from the suite and then select the benchmark from the suite.

configs/example/gem5_library/x86-npb-benchmarks.py
```diff
@@ -98,15 +94,32 @@
...
+        "npb-lu-A",
+        "npb-lu-B",
+        "npb-lu-C",
+        "npb-mg-A",
+        "npb-mg-B",
+        "npb-mg-C",
+        "npb-sp-A",
+        "npb-sp-B",
+        "npb-sp-C",
+    ],
```
powerjg: Can we get this from the suite instead of writing them all out here?

Harshil2107: We can, I did this at the start, but when we initialize the suite it downloads all the resources that are required so it will also download everything when we do `-h` to get benchmark choices.

powerjg:
That sounds like a bug/missing feature. If I run `obtain_resource("x86-npb-suite")` it should really just download the information needed for the suite and lazily download the data needed for workloads._

configs/example/gem5_library/x86-npb-benchmarks.py
powerjg: That sounds like a bug/missing feature. If I run obtain_resource("x86-npb-suite") it should really just download the information needed for the suite and lazily download the data needed for workloads.

Harshil2107: I will push a patch that would download the workload and its dependent resources when we first call them in the iterator. I will also add a function that just returns the metadata for a workload only.

Harshil2107: PR #609 has changes to lazily download workloads.

Harshil2107:
Note: This change is made on the assumption that #625 get in.

configs/example/gem5_library/x86-npb-benchmarks.py
```diff
@@ -85,28 +83,24 @@
...
 parser = argparse.ArgumentParser(
     description="An example configuration script to run the npb benchmarks."
 )
 
+
+benchmark_options = {}
+npb_suite = obtain_resource("npb-benchmark-suite")
+
+for workload in npb_suite:
+    benchmark_options[workload.get_id()] = workload
```
BobbyRBruce:
```suggestion
npb_suite = obtain_resource("npb-benchmark-suite")
```

configs/example/gem5_library/x86-npb-benchmarks.py
```diff
@@ -85,28 +83,24 @@
...
-    choices=benchmark_choices,
-)
-
-parser.add_argument(
-    "--size",
-    type=str,
-    required=True,
-    help="Input the class of the program to simulate.",
-    choices=size_choices,
+    choices=list(benchmark_options.keys()),
```
BobbyRBruce:
```suggestion
    choices=[workload.get_id() for workload in npb_suite],
```

configs/example/gem5_library/x86-npb-benchmarks.py
```diff
@@ -193,23 +188,7 @@
...
-    # The x86 linux kernel will be automatically downloaded to the
-    # `~/.cache/gem5` directory if not already present.
-    # npb benchamarks was tested with kernel version 4.19.83
-    kernel=obtain_resource("x86-linux-kernel-4.19.83"),
-    # The x86-npb image will be automatically downloaded to the
-    # `~/.cache/gem5` directory if not already present.
-    disk_image=obtain_resource("x86-npb"),
-    readfile_contents=command,
-)
+board.set_workload(benchmark_options[args.benchmark])
```
BobbyRBruce:
```suggestion
board.set_workload(obtain_resource(args.benchmark))
```

BobbyRBruce (changes requested):
Just do this. I think it's a bit clearer.

BobbyRBruce (approved):

PR #589, opened 2023-11-22
cpu,arch-arm,arch-riscv: adding new instruction types to RISC-V

ivanfv:
This commit adds more detailed instruction types for RISC-V Vector. Concretely, it substitutes VectorIntegerArith, VectorFloatArith, VectorIntegerReduce and VectorFloatReduce with more specific types related to the operation that each instruction (e.g., VectorIntegerAdd or VectorIntegerMult).

Additionaly, fixes two RISC-V instruction types (VectorXXX) that were used in ARM SVE, placing the proper SimdXXX ones.

Change-Id: I31774fa6a7cd249abfffec68d11d3d77f08ad70b

CC @adriaarmejach


src/cpu/minor/BaseMinorCPU.py
powerjg, line 1: I think these changes would need to be reflected in O3CPU too, right?

src/arch/arm/isa/insts/sve.isa
```diff
@@ -4249,19 +4249,19 @@ let {{
     sveTerInst('mls', 'Mls', 'SimdMultAccOp', signedTypes, mlsCode)
     # ADCLT
     adcltCode = 'res = srcElem1 + srcElem2 + carryIn;'
-    sveTerInstUnpred('adclt', 'Adclt', 'VectorIntegerArithOp', unsignedTypes,
+    sveTerInstUnpred('adclt', 'Adclt', 'SimdAddOp', unsignedTypes,
```
powerjg, line 4429: I think SVE is considered "vector" since it's variable length. But I'll let @giactra chime in here with his opinion.

powerjg (commented):
Overall, this looks OK to me. I'm all for more fine-grained information.

Let's wait for Arm people to comment on the SVE changes before merging.

src/cpu/minor/BaseMinorCPU.py
ivanfv: Yes, that's right. I did not add them before because the prior types (e.g., VectorIntegerArith) were not present in the O3 CPU FU. I fixed it and added a new commit.

src/arch/arm/isa/insts/sve.isa
powerjg: @giactra any opinions here?

powerjg: @giactra are you going to be able to look at this before the staging branch is created on Friday?

andysan (dismissed):
Can we please split this into two pull requests? The change to the O3 CPU needs some further discussion. Getting more fine-grained information is a good idea, but why do we need two different types of vector units with seemingly the same functionality? That doesn't seem right to me. This needs to be justified in the commit message if there is a good reason for that change. The commit message also incorrectly states that this is a RISC-V change when it is in fact an O3 CPU change.

@giactra can you take a look at the Arm changes please?

ivanfv:
Thanks for reviewing this, @andysan .

The fact of having two different types of vector units in O3 mimics what is done in the Minor one, where they are separated. I guess that if we follow a "single vector unit approach", we will need to add the Vector types into the Simd one, and I am not sure if this is appropriate. This will also require to modify the Minor CPU accordingly.

We understand that, at this point, having two similar vector units seem duplication, but also believe that this will make sense once the Vector unit is enhanced more functionalities with respect to the Simd one.

In any case, if you still feel I should proceed to split the PR and further discuss this, please let me know.

andysan:
> Thanks for reviewing this, @andysan .
> 
> The fact of having two different types of vector units in O3 mimics what is done in the Minor one, where they are separated. I guess that if we follow a "single vector unit approach", we will need to add the Vector types into the Simd one, and I am not sure if this is appropriate. This will also require to modify the Minor CPU accordingly.
> 
> We understand that, at this point, having two similar vector units seem duplication, but also believe that this will make sense once the Vector unit is enhanced more functionalities with respect to the Simd one.
> 
> In any case, if you still feel I should proceed to split the PR and further discuss this, please let me know.

Yes. please do that. I think it's worth making a conscious decision about the way we model vector units.

Looking at the minor CPU, it seems like the SIMD unit is what handles "normal" SIMD instructions, and `MinorDefaultVecFU` handles irregular instructions like gather/scatter and horizontal operations. Having different functional units with subtly different feature sets isn't that unusual, but exactly what you handle in which unit is very microarchitectural. It seems like the original intention was that the `Vector` prefix was used for irregular operations and `SIMD` for regular operations. This isn't very helpful though since these are usual seen as synonymous. It might make sense to just use the `Vector` (or maybe `SIMD` if that change is less intrusive) prefix for all instructions and `VectorIrregFoo` for operations that can't be performed independently per element.

aarmejach:
I think the `Vector` prefix was introduced with the addition of the RISC-V Vector ISA extension. All Arm ISA definitions employ `SIMD*`, while all RVV ISA definitions employ `Vector*`. The other difference is that RVV loads/stores are also tagged with specific opclasses if they originate from a "Vector" instruction.

I think that today there is no fundamental difference in the way they are modeled and we can probably make `SIMD*` and non-memory `Vector*` opclasses go through the SIMD functional unit by default. Which does not modify O3, beyond accepting more opclasses on certain FUs, and we can revert the Minor Vector FU and coalesce it with the SIMD one.

However, I think the main reason to tag RVV instructions as `Vector*` is that many RVV architectures will implement concepts from classic Vector designs, i.e., large vector registers but a narrower amount of vector lanes. Which likely means different modelling for `Vector*` and `SIMD*` instructions will be needed at the FU level in the future. In fact, we are planning to make a Vector FU model that can have a configurable number of lanes, and implement things like chaining, etc. Thus, leaving `SIMD*` opclasses for more traditional SVE/AVX `register width == FU width` and `Vector*` for more segmented execution style instructions. 

If this makes sense we can keep the `Vector*` opclasses and for now make them go through the SIMD FUs both in Minor and O3.

powerjg:
At this point, let's not do major refactoring of how we have named things.

I think we need to have a separation of SIMD and Vector instruction tags so we can use different functional units for them. This would be best to happen *before* the 23.1 release so that during development of 24.0 we can concentrate on creating core/processor stdlib components that have the "right" mix of units.

However, this change must be compatible with the O3CPU. People are currently using O3CPU+RVV, so we can't fundamentally break what they are doing.

So, a question to @ivanfv and @aarmejach, does this change affect running RVV apps on the O3CPU?

aarmejach:
To me the sensitive thing to do is to make `Vector*Load` and `Vector*Store` opclasses go to the existing `MinorDefaultMemFU` for Minor, and for O3 to put `Vector*Load` to the `ReadPort` and `RdWrPort` FUs and `Vector*Store` to the `WritePort` and `RdWrPort` FUs. 

The rest of the `Vector*` opclasses can go into the existing `MinorDefaultFloatSimdFU` for Minor and `SIMD_Unit` for O3, and remove the current `MinorDefaultVecFU`.

This does not affect O3 or Minor. In fact, currently `Vector*` opclasses do not have an assigned FU in O3 by default.

If we agree this is the way to go, @ivanfv can work on this, but I don't know if it is wise to rush this today.

powerjg:
I'm happy with this suggestion @aarmejach! 

> This does not affect O3 or Minor. In fact, currently Vector* opclasses do not have an assigned FU in O3 by default.

I would like to see this "fixed" before the 23.1 release so we can say "improved support for RVV in Minor and O3CPUs" in the release notes :). This would be a great milestone.

> I don't know if it is wise to rush this today.

If we can get the changes that add the new op types and update the decoder for RVV done before the staging branch, that would be good. This way, while we're testing the staging branch we can make sure to test these changes.

For the changes on functional units (both Minor and O3), we can put that on the staging branch before the release. Those changes shouldn't have to be tested as rigorously.

I suggest removing the Arm changes. The Arm maintainers @andysan and @giactra can have the responsibility of correctly defining the instruction types for SVE once we make changes to the functional units in develop-for-24.0. As far as I can tell, if we remove those changes, add the new vector types, and use the "SIMD" functional units for the vector types, the Arm SVE instructions will be completely unchanged. (Let me know if I'm wrong here, though.)

Finally, we're pushing to see how much we can finish today, but I'm aware that it's the end of the day on Friday in Europe, so if we can't get everything done, we'll pick it back up Monday morning. I *will* warn that we are likely to merge #362 at the end of the day today (or over the weekend), which is going to make merging/rebasing a headache.

andysan:
> I'm happy with this suggestion @aarmejach!
> 
> > This does not affect O3 or Minor. In fact, currently Vector* opclasses do not have an assigned FU in O3 by default.
> 
> I would like to see this "fixed" before the 23.1 release so we can say "improved support for RVV in Minor and O3CPUs" in the release notes :). This would be a great milestone.
> 
> > I don't know if it is wise to rush this today.
> 
> If we can get the changes that add the new op types and update the decoder for RVV done before the staging branch, that would be good. This way, while we're testing the staging branch we can make sure to test these changes.
> 
> For the changes on functional units (both Minor and O3), we can put that on the staging branch before the release. Those changes shouldn't have to be tested as rigorously.
> 
> I suggest removing the Arm changes. The Arm maintainers @andysan and @giactra can have the responsibility of correctly defining the instruction types for SVE once we make changes to the functional units in develop-for-24.0. As far as I can tell, if we remove those changes, add the new vector types, and use the "SIMD" functional units for the vector types, the Arm SVE instructions will be completely unchanged. (Let me know if I'm wrong here, though.)

That would limit the impact to the Arm side, but it still makes significant changes to the way we classify instructions. My concern is that we don't fully understand how these instruction types are supposed to be used and it is certainly not documented anywhere. In fact, they seem redundant. The main difference that @aarmejach seems more like a microarchitectural difference which would apply to SVE (I believe there are implementations already where FU width != register widht).

I would be more in favour of merging the SIMD and Vector concepts since it seems like all of the differences are down to microarchitecture and how the timing models are configured. We would of course merge these changes as is to make the release, but I worry that it will just introduce additional work in the future if/when we decide go back to one vector instruction namespace.

ivanfv:
> To me the sensitive thing to do is to make `Vector*Load` and `Vector*Store` opclasses go to the existing `MinorDefaultMemFU` for Minor, and for O3 to put `Vector*Load` to the `ReadPort` and `RdWrPort` FUs and `Vector*Store` to the `WritePort` and `RdWrPort` FUs.
> 
> The rest of the `Vector*` opclasses can go into the existing `MinorDefaultFloatSimdFU` for Minor and `SIMD_Unit` for O3, and remove the current `MinorDefaultVecFU`.
> 
> This does not affect O3 or Minor. In fact, currently `Vector*` opclasses do not have an assigned FU in O3 by default.
> 
> If we agree this is the way to go, @ivanfv can work on this, but I don't know if it is wise to rush this today.

Applying this + removing the Arm changes is something I see feasible to do before the end of today.

However, merging Simd and Vector requires more effort and I guess it might require further discussion.

giactra:
From what I understood, one of the reasons why the Vector* op classes were introduced (https://gem5-review.googlesource.com/c/public/gem5/+/67293) was to allow a fine grained tuning for pure vector instructions latency; more specifically to differentiate them from SIMD implementations like Neon.

If this was the main motivation, I would suggest to merge the two categories and to provide maybe a better way to define per-instruction latency. It seems to me the only way this is possible now is for the MinorCPU only, via the MinorFUTiming... I don't see anything like it for the O3CPU

camel-cdr:
I'm not sure if this is the correct place to write this, but it seems related:

I think something similar should be done for the permutation instructions, specifically `vrgather` and `vcompress` are performance outliers, which isn't currently reflected in the model.

[here](https://camel-cdr.github.io/rvv-bench-results/index.html) are a few measurements, but to summarize, these are some throughput measurements:

`vcompress.vm`:

|         | VLEN | e8m1 | e8m2 | e8m4 | e8m8  |
| ---     | ---  | ---  | ---  | ---  | ---   |
| c906    | 128  | 4    | 10   | 32   | 136   |
| c908    | 128  | 4    | 10   | 32   | 139.4 |
| c920    | 128  | 0.5  | 2.4  | 5.4  | 20.0  |
| bobcat* | 256  | 32   | 64   | 132  | 260   |
| x280*   | 512  | 65   | 129  | 257  | 513   |

`vrgather.vv`:

|         | VLEN | e8m1 | e8m2 | e8m4 | e8m8  |
| ---     | ---  | ---  | ---  | ---  | ---   |
| c906    | 128  | 4    | 16   | 64   | 256   |
| c908    | 128  | 4    | 16   | 64.9 | 261.1 |
| c920    | 128  | 0.5  | 2.4  | 8.0  | 32.0  |
| bobcat* | 256  | 68   | 132  | 260  | 516   |
| x280*   | 512  | 65   | 129  | 257  | 513   |

*bobcat: note that it was explicitly stated, that they didn't optimize the permutation instructions

*x280: the numbers are from llvm-mca, but I was told they match reality. There is also supposed to be a `vrgather` fast path for `vl<=256`. I think they didn't have much incentive to make this fast, as the x280 mostly targets AI.

I think that the C920 results are the most representative for what to expect of future desktop CPUs.
Personally, I suspect we'll see `vrgather.vv` perform well for any `SEW` under `LMUL=1`, and then grow exponential per element with higher `LMUL` in the best case, as an all to all mapping is quite expensive to scale.
`vcompress.vm` should be better scalable than `vrgather.vv`, since the work is subdividable, and I think we might see a range of implementations from similar to `vrgather.vv` to almost linear growth with `LMUL`.

ivanfv:
Hi All,

Just wanted to let you know that we are currently refactoring this pull request to combine Simd* and Vector* instruction namespaces, so we remove the redundancies. I will update the pull request as soon as we are done.

BR,
Ivan

ivanaamit:
Hi @camel-cdr, could you please create a new issue for the changes you proposed? Thank you.

ivanfv:
Hi, 

I have been modifying this pull request in a way Simd and Vector instruction types are now combined (keeping only the Simd ones). I also modified minor and o3 CPU accordingly. Let me know your thoughts.

Ivan

src/arch/riscv/insts/vector.cc
```diff
@@ -414,7 +414,7 @@ VMvWholeMicroInst::generateDisassembly(Addr pc,
 VMaskMergeMicroInst::VMaskMergeMicroInst(ExtMachInst extMachInst,
     uint8_t _dstReg, uint8_t _numSrcs, uint32_t _vlen, size_t _elemSize)
     : VectorArithMicroInst("vmask_mv_micro", extMachInst,
-                            VectorIntegerArithOp, 0, 0),
+                            SimdMiscOp, 0, 0),
```
aarmejach: I think this performs an addition maybe use: `SimdAddOp`

aarmejach:
```suggestion
                            SimdAddOp, 0, 0),
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2629,7 +2629,7 @@ decode QUADRANT default Unknown::unknown() {
                         }
 
                         Vd_vu[i + offset] = (vu)res;
-                    }}, OPIVV, VectorIntegerArithOp);
+                    }}, OPIVV, SimdConvertOp);
```
aarmejach: Use the existing `SimdCvtOp`

```suggestion
                    }}, OPIVV, SimdCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2648,7 +2648,7 @@ decode QUADRANT default Unknown::unknown() {
                         }
 
                         Vd_vi[i + offset] = (vi)res;
-                    }}, OPIVV, VectorIntegerArithOp);
+                    }}, OPIVV, SimdConvertOp);
```
aarmejach:
```suggestion
                    }}, OPIVV, SimdCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2657,57 +2657,57 @@ decode QUADRANT default Unknown::unknown() {
...
                         return fmax<et>(ftype<et>(src1), ftype<et>(src2));
                     }, Vs1_vu, Vs2_vu);
-                }}, OPFVV, VectorFloatReduceOp);
+                }}, OPFVV, SimdFloatReduceCmpOp);
                 0x08: VectorFloatFormat::vfsgnj_vv({{
                     Vd_vu[i] = fsgnj<et>(ftype<et>(Vs2_vu[i]),
                                          ftype<et>(Vs1_vu[i]),
                                          false, false).v;
-                }}, OPFVV, VectorFloatArithOp);
+                }}, OPFVV, SimdFloatAluOp);
```
aarmejach: Sign-injection insts should be cvt maybe 
```suggestion
                }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2657,57 +2657,57 @@ decode QUADRANT default Unknown::unknown() {
...
                 0x09: VectorFloatFormat::vfsgnjn_vv({{
                     Vd_vu[i] = fsgnj<et>(ftype<et>(Vs2_vu[i]),
                                          ftype<et>(Vs1_vu[i]),
                                          true, false).v;
-                }}, OPFVV, VectorFloatArithOp);
+                }}, OPFVV, SimdFloatAluOp);
```
aarmejach:
```suggestion
                }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2657,57 +2657,57 @@ decode QUADRANT default Unknown::unknown() {
...
                 0x0a: VectorFloatFormat::vfsgnjx_vv({{
                     Vd_vu[i] = fsgnj<et>(ftype<et>(Vs2_vu[i]),
                                          ftype<et>(Vs1_vu[i]),
                                          false, true).v;
-                }}, OPFVV, VectorFloatArithOp);
+                }}, OPFVV, SimdFloatAluOp);
```
aarmejach:
```suggestion
                }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
+                        }}, OPFVV, SimdMiscOp);
                     }
                 }
                 0x12: decode VS1 {
                     format VectorFloatCvtFormat {
                         0x00: vfcvt_xu_f_v({{
                             Vd_vu[i] = f_to_ui<et>(ftype<et>(Vs2_vu[i]),
                                                    softfloat_roundingMode);
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x01: vfcvt_x_f_v({{
                             Vd_vu[i] = f_to_i<et>(ftype<et>(Vs2_vu[i]),
                                                   softfloat_roundingMode);
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x02: vfcvt_f_xu_v({{
                             auto fd = ui_to_f<et>(Vs2_vu[i]);
                             Vd_vu[i] = fd.v;
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x03: vfcvt_f_x_v({{
                             auto fd = i_to_f<et>(Vs2_vu[i]);
                             Vd_vu[i] = fd.v;
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x06: vfcvt_rtz_xu_f_v({{
                             Vd_vu[i] = f_to_ui<et>(ftype<et>(Vs2_vu[i]),
                                                    softfloat_round_minMag);
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x07: vfcvt_rtz_x_f_v({{
                             Vd_vu[i] = f_to_i<et>(ftype<et>(Vs2_vu[i]),
                                                   softfloat_round_minMag);
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                     }
                     format VectorFloatWideningCvtFormat {
                         0x08: vfwcvt_xu_f_v({{
                             Vd_vwu[i] = f_to_wui<et>(
                                 ftype<et>(Vs2_vu[i + offset]),
                                 softfloat_roundingMode);
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x09: vfwcvt_x_f_v({{
                             Vd_vwu[i] = f_to_wi<et>(
                                 ftype<et>(Vs2_vu[i + offset]),
                                 softfloat_roundingMode);
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x0a: vfwcvt_f_xu_v({{
                             auto fd = ui_to_wf<vu>(Vs2_vu[i + offset]);
                             Vd_vwu[i] = fd.v;
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x0b: vfwcvt_f_x_v({{
                             auto fd = i_to_wf<vu>(Vs2_vu[i + offset]);
                             Vd_vwu[i] = fd.v;
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x0c: vfwcvt_f_f_v({{
                             auto fd = f_to_wf<et>(
                                 ftype<et>(Vs2_vu[i + offset]));
                             Vd_vwu[i] = fd.v;
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x0e: vfwcvt_rtz_xu_f_v({{
                             Vd_vwu[i] = f_to_wui<et>(
                                 ftype<et>(Vs2_vu[i + offset]),
                                 softfloat_round_minMag);
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x0f: vfwcvt_rtz_x_f_v({{
                             Vd_vwu[i] = f_to_wi<et>(
                                 ftype<et>(Vs2_vu[i + offset]),
                                 softfloat_round_minMag);
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                     }
                     format VectorFloatNarrowingCvtFormat {
                         0x10: vfncvt_xu_f_w({{
                             Vd_vu[i + offset] = f_to_nui<vu>(
                                 ftype<ewt>(Vs2_vwu[i]),
                                 softfloat_roundingMode);
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x11: vfncvt_x_f_w({{
                             Vd_vu[i + offset] = f_to_ni<vu>(
                                 ftype<ewt>(Vs2_vwu[i]),
                                 softfloat_roundingMode);
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x12: vfncvt_f_xu_w({{
                             auto fd = ui_to_nf<et>(Vs2_vwu[i]);
                             Vd_vu[i + offset] = fd.v;
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x13: vfncvt_f_x_w({{
                             auto fd = i_to_nf<et>(Vs2_vwu[i]);
                             Vd_vu[i + offset] = fd.v;
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x14: vfncvt_f_f_w({{
                             auto fd = f_to_nf<et>(ftype<ewt>(Vs2_vwu[i]));
                             Vd_vu[i + offset] = fd.v;
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x15: vfncvt_rod_f_f_w({{
                             softfloat_roundingMode = softfloat_round_odd;
                             auto fd = f_to_nf<et>(ftype<ewt>(Vs2_vwu[i]));
                             Vd_vu[i + offset] = fd.v;
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x16: vfncvt_rtz_xu_f_w({{
                             Vd_vu[i + offset] = f_to_nui<vu>(
                                 ftype<ewt>(Vs2_vwu[i]),
                                 softfloat_round_minMag);
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -2716,129 +2716,129 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x17: vfncvt_rtz_x_f_w({{
                             Vd_vu[i + offset] = f_to_ni<vu>(
                                 ftype<ewt>(Vs2_vwu[i]),
                                 softfloat_round_minMag);
-                        }}, OPFVV, VectorFloatConvertOp);
+                        }}, OPFVV, SimdFloatConvertOp);
```
aarmejach:
```suggestion
                        }}, OPFVV, SimdFloatCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3106,40 +3106,40 @@ decode QUADRANT default Unknown::unknown() {
                             }
                         }
                         Rd_vu = pos;
-                    }}, OPMVV, VectorMiscOp);
+                    }}, OPMVV, SimdAluOp);
```
aarmejach:
```suggestion
                    }}, OPMVV, SimdMiscOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3106,40 +3106,40 @@ decode QUADRANT default Unknown::unknown() {
...
                 }
                 0x12: decode VS1 {
                     format VectorIntExtFormat {
                         0x02: vzext_vf8({{
                             auto offset = (vlen / SEW) * index;
 
                             Vd_vu[i] = Vs2_vextu[i + offset];
-                        }}, OPMVV, VectorIntegerExtensionOp);
+                        }}, OPMVV, SimdExtensionOp);
```
aarmejach: To follow convention maybe i'd go with `SimdExt`
```suggestion
                        }}, OPMVV, SimdExtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3106,40 +3106,40 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x03: vsext_vf8({{
                             auto offset = (vlen / SEW) * index;
 
                             Vd_vi[i] = Vs2_vext[i + offset];
-                        }}, OPMVV, VectorIntegerExtensionOp);
+                        }}, OPMVV, SimdExtensionOp);
```
aarmejach:
```suggestion
                        }}, OPMVV, SimdExtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3106,40 +3106,40 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x04: vzext_vf4({{
                             auto offset = (vlen / SEW) * index;
 
                             Vd_vu[i] = Vs2_vextu[i + offset];
-                        }}, OPMVV, VectorIntegerExtensionOp);
+                        }}, OPMVV, SimdExtensionOp);
```
aarmejach:
```suggestion
                        }}, OPMVV, SimdExtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3106,40 +3106,40 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x05: vsext_vf4({{
                             auto offset = (vlen / SEW) * index;
 
                             Vd_vi[i] = Vs2_vext[i + offset];
-                        }}, OPMVV, VectorIntegerExtensionOp);
+                        }}, OPMVV, SimdExtensionOp);
```
aarmejach:
```suggestion
                        }}, OPMVV, SimdExtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3106,40 +3106,40 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x06: vzext_vf2({{
                             auto offset = (vlen / SEW) * index;
 
                             Vd_vu[i] = Vs2_vextu[i + offset];
-                        }}, OPMVV, VectorIntegerExtensionOp);
+                        }}, OPMVV, SimdExtensionOp);
```
aarmejach:
```suggestion
                        }}, OPMVV, SimdExtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3106,40 +3106,40 @@ decode QUADRANT default Unknown::unknown() {
...
                         0x07: vsext_vf2({{
                             auto offset = (vlen / SEW) * index;
 
                             Vd_vi[i] = Vs2_vext[i + offset];
-                        }}, OPMVV, VectorIntegerExtensionOp);
+                        }}, OPMVV, SimdExtensionOp);
```
aarmejach:
```suggestion
                        }}, OPMVV, SimdExtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3213,159 +3213,159 @@ decode QUADRANT default Unknown::unknown() {
                                 *cnt = *cnt+1;
                             }
                         }
-                    }}, OPMVV, VectorMiscOp);
+                    }}, OPMVV, SimdAluOp);
                     0x11: VectorIntFormat::vid_v({{
                         Vd_vu[i] = ei;
-                    }}, OPMVV, VectorMiscOp);
+                    }}, OPMVV, SimdAluOp);
```
aarmejach:
```suggestion
                    }}, OPMVV, SimdMiscOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3404,7 +3404,7 @@ decode QUADRANT default Unknown::unknown() {
                             }
                         }
                     }
-                }}, OPIVI, VectorMiscOp);
+                }}, OPIVI, SimdAluOp);
```
aarmejach: I'd mark all register to register "move" insts as Misc for now. 

```suggestion
                }}, OPIVI, SimdMiscOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3443,43 +3443,43 @@ decode QUADRANT default Unknown::unknown() {
...
```
aarmejach:
```suggestion
                }}, OPIVI, SimdMiscOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3581,7 +3581,7 @@ decode QUADRANT default Unknown::unknown() {
                         }
 
                         Vd_vu[i + offset] = (vu)res;
-                    }}, OPIVI, VectorIntegerArithOp);
+                    }}, OPIVI, SimdConvertOp);
```
aarmejach:
```suggestion
                    }}, OPIVI, SimdCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3600,42 +3600,42 @@ decode QUADRANT default Unknown::unknown() {
                         }
 
                         Vd_vi[i + offset] = (vi)res;
-                    }}, OPIVI, VectorIntegerArithOp);
+                    }}, OPIVI, SimdConvertOp);
```
aarmejach:
```suggestion
                    }}, OPIVI, SimdCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3662,7 +3662,7 @@ decode QUADRANT default Unknown::unknown() {
                             }
                         }
                     }
-                }}, OPIVX, VectorMiscOp);
+                }}, OPIVX, SimdAluOp);
```
aarmejach:
```suggestion
                }}, OPIVX, SimdMiscOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3701,7 +3701,7 @@ decode QUADRANT default Unknown::unknown() {
...
```
aarmejach:
```suggestion
                }}, OPIVX, SimdMiscOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3823,7 +3823,7 @@ decode QUADRANT default Unknown::unknown() {
                         }
 
                         Vd_vu[i + offset] = (vu)res;
-                    }}, OPIVX, VectorIntegerArithOp);
+                    }}, OPIVX, SimdConvertOp);
```
aarmejach:
```suggestion
                    }}, OPIVX, SimdCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3842,7 +3842,7 @@ decode QUADRANT default Unknown::unknown() {
                         }
 
                         Vd_vi[i + offset] = (vi)res;
-                    }}, OPIVX, VectorIntegerArithOp);
+                    }}, OPIVX, SimdConvertOp);
```
aarmejach:
```suggestion
                    }}, OPIVX, SimdCvtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3909,37 +3909,37 @@ decode QUADRANT default Unknown::unknown() {
...
                                            ftype_freg<et>(freg(Fs1_bits)));
                             Vd_vu[i] = fd.v;
-                    }}, OPFVF, VectorFloatArithOp);
+                    }}, OPFVF, SimdFloatCmpOp);
                     0x08: vfsgnj_vf({{
                         Vd_vu[i] = fsgnj<et>(ftype<et>(Vs2_vu[i]),
                                              ftype_freg<et>(freg(Fs1_bits)),
                                              false, false).v;
-                    }}, OPFVF, VectorFloatArithOp);
+                    }}, OPFVF, SimdFloatAluOp);
```
aarmejach: Need to add a new opclass `SimdFloatExtOp`
```suggestion
                    }}, OPFVF, SimdFloatExtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3909,37 +3909,37 @@ decode QUADRANT default Unknown::unknown() {
...
                     0x09: vfsgnjn_vf({{
                         Vd_vu[i] = fsgnj<et>(ftype<et>(Vs2_vu[i]),
                                              ftype_freg<et>(freg(Fs1_bits)),
                                              true, false).v;
-                    }}, OPFVF, VectorFloatArithOp);
+                    }}, OPFVF, SimdFloatAluOp);
```
aarmejach:
```suggestion
                    }}, OPFVF, SimdFloatExtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3909,37 +3909,37 @@ decode QUADRANT default Unknown::unknown() {
...
                     0x0a: vfsgnjx_vf({{
                         Vd_vu[i] = fsgnj<et>(ftype<et>(Vs2_vu[i]),
                                              ftype_freg<et>(freg(Fs1_bits)),
                                              false, true).v;
-                    }}, OPFVF, VectorFloatArithOp);
+                    }}, OPFVF, SimdFloatAluOp);
```
aarmejach:
```suggestion
                    }}, OPFVF, SimdFloatExtOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -3971,7 +3971,7 @@ decode QUADRANT default Unknown::unknown() {
                             tmp_d0.as<vu>()[0] = Rs1_vu;
                         }
                     }
-                }}, OPFVF, VectorMiscOp);
+                }}, OPFVF, SimdAluOp);
```
aarmejach:
```suggestion
                }}, OPFVF, SimdMiscOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -4012,7 +4012,7 @@ decode QUADRANT default Unknown::unknown() {
                             }
                         }
                     }
-                }}, OPFVF, VectorMiscOp);
+                }}, OPFVF, SimdAluOp);
```
aarmejach:
```suggestion
                }}, OPFVF, SimdMiscOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -4250,7 +4250,7 @@ decode QUADRANT default Unknown::unknown() {
                             tmp_d0.as<vu>()[0] = Rs1_vu;
                         }
                     }
-                }}, OPIVX, VectorMiscOp);
+                }}, OPIVX, SimdAluOp);
```
aarmejach:
```suggestion
                }}, OPIVX, SimdMiscOp);
```

src/arch/riscv/isa/decoder.isa
```diff
@@ -4291,7 +4291,7 @@ decode QUADRANT default Unknown::unknown() {
                             }
                         }
                     }
-                }}, OPIVX, VectorMiscOp);
+                }}, OPIVX, SimdAluOp);
```
aarmejach:
```suggestion
                }}, OPIVX, SimdMiscOp);
```

src/cpu/FuncUnit.py
```diff
@@ -100,25 +100,21 @@ class OpClass(Enum):
...
+        "SimdUnitStrideMaskStore",
+        "SimdStridedLoad",
+        "SimdStridedStore",
+        "SimdIndexedLoad",
+        "SimdIndexedStore",
+        "SimdUnitStrideFaultOnlyFirstLoad",
+        "SimdWholeRegisterLoad",
+        "SimdWholeRegisterStore",
+        "SimdExtension",
+        "SimdConvert",
```
aarmejach: remove

src/cpu/FuncUnit.py
```diff
@@ -100,25 +100,21 @@ class OpClass(Enum):
...
+        "SimdFloatConvert",
```
aarmejach: remove, using exisiting `SimdFloatCvt`

src/cpu/FuncUnit.py
```diff
@@ -100,25 +100,21 @@ class OpClass(Enum):
...
+        "SimdUnitStrideMaskLoad",
+        "SimdUnitStrideMaskStore",
+        "SimdStridedLoad",
+        "SimdStridedStore",
+        "SimdIndexedLoad",
+        "SimdIndexedStore",
+        "SimdUnitStrideFaultOnlyFirstLoad",
+        "SimdWholeRegisterLoad",
+        "SimdWholeRegisterStore",
+        "SimdExtension",
```
aarmejach:
```suggestion
        "SimdExt",
```

src/cpu/FuncUnit.py
```diff
@@ -100,25 +100,21 @@ class OpClass(Enum):
...
+        "SimdConvert",
+        "SimdFloatConvert",
+        "SimdConfig",
```
aarmejach, line 118: need to add `SimdFloatExt`

src/cpu/minor/BaseMinorCPU.py
```diff
@@ -219,6 +219,10 @@ class MinorDefaultFloatSimdFU(MinorFU):
             "Matrix",
             "MatrixMov",
             "MatrixOP",
+            "SimdExtension",
```
aarmejach:
```suggestion
            "SimdExt",
```

src/cpu/minor/BaseMinorCPU.py
```diff
@@ -219,6 +219,10 @@ class MinorDefaultFloatSimdFU(MinorFU):
...
+            "SimdConvert",
```
aarmejach: remove

src/cpu/minor/BaseMinorCPU.py
```diff
@@ -219,6 +219,10 @@ class MinorDefaultFloatSimdFU(MinorFU):
...
+            "SimdFloatConvert",
```
aarmejach: remove

src/cpu/minor/BaseMinorCPU.py
```diff
@@ -219,6 +219,10 @@ class MinorDefaultFloatSimdFU(MinorFU):
...
+            "SimdConfig",
```
aarmejach, line 225: add `SimdFloatExt`

src/cpu/minor/BaseMinorCPU.py
```diff
@@ -249,31 +269,7 @@ class MinorDefaultMiscFU(MinorFU):
...
-            "VectorFloatConvert",
-            "VectorIntegerReduce",
-            "VectorFloatReduce",
-            "VectorMisc",
-            "VectorIntegerExtension",
-            "VectorConfig",
-        ]
-    )
-    opLat = 1
+MinorDefaultVecFU = MinorDefaultFloatSimdFU
```
aarmejach: I'd remove it completely, by default we make Float and Simd opClasses go to the same FU. It also needs to be removed from the `MinorDefaultFUPool`.

src/cpu/o3/FuncUnitConfig.py
```diff
@@ -106,6 +106,10 @@ class SIMD_Unit(FUDesc):
         OpDesc(opClass="SimdReduceCmp"),
         OpDesc(opClass="SimdFloatReduceAdd"),
         OpDesc(opClass="SimdFloatReduceCmp"),
+        OpDesc(opClass="SimdExtension"),
```
aarmejach:
```suggestion
        OpDesc(opClass="SimdExt"),
```

src/cpu/o3/FuncUnitConfig.py
```diff
@@ -106,6 +106,10 @@ class SIMD_Unit(FUDesc):
...
+        OpDesc(opClass="SimdConvert"),
+        OpDesc(opClass="SimdFloatConvert"),
```
aarmejach: remove

src/cpu/o3/FuncUnitConfig.py
```diff
@@ -106,6 +106,10 @@ class SIMD_Unit(FUDesc):
...
+        OpDesc(opClass="SimdConfig"),
```
aarmejach, line 111: Add `SimdFloatExt`

src/cpu/op_class.hh
```diff
@@ -108,32 +108,28 @@ static const OpClass MemReadOp = enums::MemRead;
...
-             = enums::VectorWholeRegisterStore;
-static const OpClass VectorIntegerArithOp = enums::VectorIntegerArith;
-static const OpClass VectorFloatArithOp = enums::VectorFloatArith;
-static const OpClass VectorFloatConvertOp = enums::VectorFloatConvert;
-static const OpClass VectorIntegerReduceOp = enums::VectorIntegerReduce;
-static const OpClass VectorFloatReduceOp = enums::VectorFloatReduce;
-static const OpClass VectorMiscOp = enums::VectorMisc;
-static const OpClass VectorIntegerExtensionOp = enums::VectorIntegerExtension;
-static const OpClass VectorConfigOp = enums::VectorConfig;
+static const OpClass SimdConvertOp = enums::SimdConvert;
```
aarmejach: remove

src/cpu/op_class.hh
```diff
@@ -108,32 +108,28 @@ static const OpClass MemReadOp = enums::MemRead;
...
+static const OpClass SimdFloatConvertOp = enums::SimdFloatConvert;
```
aarmejach: remove

src/cpu/op_class.hh
```diff
@@ -108,32 +108,28 @@ static const OpClass MemReadOp = enums::MemRead;
...
+static const OpClass SimdExtensionOp = enums::SimdExtension;
```
aarmejach:
```suggestion
static const OpClass SimdExtOp = enums::SimdExt;
```

src/cpu/op_class.hh
```diff
@@ -108,32 +108,28 @@ static const OpClass MemReadOp = enums::MemRead;
...
+static const OpClass SimdConfigOp = enums::SimdConfig;
```
aarmejach, line 134: Add `SimdFloatExtOp`

aarmejach (changes requested):
Thanks @ivanfv. Made a few change requests.

ivanfv:
Thanks @adriaarmejach , I think I addressed your change requests and updated the commit properly. 

However, if I run the tests locally there is at least one (suite-riscv-vertical-microbenchmarks_riscv_suite_test_to-tick-ALL-x86_64-opt) that fails with the following error:

```
NameError: name 'MinorDefaultVecFU' is not defined

At:
  src/python/gem5/prebuilt/riscvmatched/riscvmatched_core.py(79): <module>
  <string>(40): exec_module
  <frozen importlib._bootstrap>(703): _load_unlocked
  <frozen importlib._bootstrap>(1006): _find_and_load_unlocked
  <frozen importlib._bootstrap>(1027): _find_and_load
  src/python/gem5/prebuilt/riscvmatched/riscvmatched_processor.py(35): <module>
  <string>(40): exec_module
  <frozen importlib._bootstrap>(703): _load_unlocked
  <frozen importlib._bootstrap>(1006): _find_and_load_unlocked
  <frozen importlib._bootstrap>(1027): _find_and_load
  src/python/gem5/prebuilt/riscvmatched/riscvmatched_board.py(73): <module>
  <string>(40): exec_module
  <frozen importlib._bootstrap>(703): _load_unlocked
  <frozen importlib._bootstrap>(1006): _find_and_load_unlocked
  <frozen importlib._bootstrap>(1027): _find_and_load
  /home/ifernand/projects/gem5-insts-types/tests/gem5/suite_tests/configs/suite_run_workload.py(111): <module>
  src/python/m5/main.py(669): main
```
I need to check if removing _MinorDefaultVecFU_ and keeping only the Simd one has other effects or if this is only a matter of fixing the test properly. Actually, this failing test is the reason I was keeping the _MinorDefaultVecFU_ defined to be the same as the _MinorDefaultFloatSimdFU_  (and included in the FUPool) in the prior version of the commit.

Feedback from other folks would be also very appreciated.

Ivan

powerjg:
@Harshil2107 I think you added this test. Could you take a look?

Harshil2107:
I will take a look at the suite test.

Additionally, please install pre-commit (https://github.com/gem5/gem5/blob/stable/CONTRIBUTING.md#using-pre-commit) as the pre-commit tests are failing.

ivanfv:
Thanks, @Harshil2107. I think I solved the pre-commit issue.

src/cpu/minor/BaseMinorCPU.py
```diff
@@ -249,33 +269,6 @@ class MinorDefaultMiscFU(MinorFU):
     opLat = 1
 
 
-class MinorDefaultVecFU(MinorFU):
```
Harshil2107, line 252: The suite test is failing as the RISCV microbenchmark suite uses the riscvmatched board which uses these instructions https://github.com/gem5/gem5/blob/c890e6b113a59caeec3eea9bafea4a03a471f89d/src/python/gem5/prebuilt/riscvmatched/riscvmatched_core.py#L79

We can either just remove this from the riscvmatched board or we keep the vector instructions. I am not an expert at this so I am not sure what the best approach is.

ivanfv: Thanks for taking a look, @Harshil2107. I believe we can remove that from the riscvmatched board since we are using the Simd types, and they are already in the MinorDefaultFloatSimdFU. I fixed the commit accordingly, but feedback from other folks would be appreciated.

powerjg: This is OK with me :)

ivanfv:
Hi,

I believe I implemented the feedback so far and checks are passing. Is there anything else we want to modify in this pull request?

Ivan

giactra:
> Hi,
> 
> I believe I implemented the feedback so far and checks are passing. Is there anything else we want to modify in this pull request?
> 
> Ivan

Hi Ivan, thanks for your effort.

I see this PR and its commit are mislabelled as they are tagged as **arch-riscv** whereas they touch **arch-arm** and **cpu** subsystems as well.

I would propose not only to rename the PR, but also to split this PR in multiple commits. From a logical point of view I believe we should have

1) Removing VectorIntegerArithOp and other non load/store op classes in favour of SIMD op classes (this should be tagged as cpu,arch-arm,arch-risv (or misc) alternatively)
2) Renaming Vector(.\*) into Simd(.\*) (this should be tagged as arch-riscv,cpu)
3) Other FU changes (tagged as cpu) <- Still need to review this

ivanfv:
Hi @giactra, thanks for the feedback.

I already divided the PR into three commits:

1. Changing a few supposedly wrong ARM's that were set as Vector to Simd (arch-arm).
3. Renaming RISC-V's Vector types to Simd (arch-riscv).
4. Modifying FUs accordingly (cpu).

Let me know if this matches what you have in mind or if I misunderstood something.

Ivan

ivanfv:
Hi,

Just wanted to let you know that I realized that this pull request somewhat depends on PR #851 and PR #913, as they are adding new vector instructions with the "VectorXXX" type. As a consequence, ideally it would be better if they are merged before this one.

Once they are merged, I will update this PR accordingly to set those new vector segmented instructions to follow the Simd convention.

BR,
Ivan

aarmejach (approved):

ivanaamit:
All dependent PRs have been merged.

@ivanfv, could you please resolve the conflicts? Thanks.

@giactra, could you please review this PR? Thanks.

ivanfv:
Hi, I think I solved the conflicts

aarmejach:
We would like to make progress on this one to unlock some other internal PRs that depend on this. 

I think the concerns @andysan had are resolved by adopting the `Simd` namespace everywhere, and it is clear to me that the sve instructions should have been marked as `Simd` from the start.

andysan:
> We would like to make progress on this one to unlock some other internal PRs that depend on this. 
> 
> 
> 
> I think the concerns @andysan had are resolved by adopting the `Simd` namespace everywhere, and it is clear to me that the sve instructions should have been marked as `Simd` from the start. 

I won't have time to look at this the next couple of week. @giactra could you have a look and approve if this looks ok for us?

giactra (approved):
LGTM as I think it cleans up the FUs/OpDescs caos. In the long term I believe we should provide to the users the capability to fine tune the latency of single instructions without the need of introducing new shared OpClasses.

I believe the current approach as it is now has a high risk of duplication

ivanaamit:
Hi @andysan and @ivanfv, Andreas's [requested change](https://github.com/gem5/gem5/pull/589#pullrequestreview-1759409178) is still pending. I just wanted to double-check if it has been resolved before merging this PR. Thank you.

aarmejach:
@ivanaamit i think with the aproval of @giactra it is good to go

PR #593, opened 2023-11-23
arch-x86: Fix misc registers in mov instructions

OdnetninI:
MOV instructions 8C and 8E can be prefixed with a REX prefix to extend the source/destination register. 
However, the R bit in REX will be applied to the segment register.  
The decoder file checks for valid segment registers, checking the MODRM_REG only, however, later this will be extended with the REX_R when adding the register to the sources/destinations of the instruction.  
This will trigger an assert.

Additionally, MOV instructions of various miscelaneous registers are also not check for being valid when taking into account the REX_R bit.

This patch checks that the REX_R is not set, otherwise, UD2 will be generated.


OdnetninI:
There is a very similar issue with MOV from misc registers.
Updating the patch soon.

powerjg (approved):
lgtm! Thanks for the fix!

PR #596, opened 2023-11-24
arch-arm: add Sve mla and mls indexed

Nitesh8998:
This contains the implementation of mla and MLS index version instructions from ARM SVE2 ISA specification.

### DIFF
--- a/src/arch/arm/insts/sve.hh
+++ b/src/arch/arm/insts/sve.hh
@@ -498,27 +498,6 @@ class SveTerPredOp : public ArmStaticInst
             Addr pc, const loader::SymbolTable *symtab) const override;
 };
 
-/// Ternary, destructive, unpredicated , !INDEXED! SVE Instruction
-class SveTerIndexedOp : public ArmStaticInst
-{
-  protected:
-    RegIndex dest, op1, op2;
-    uint16_t imm;
-    uint8_t esize;
-
-    SveTerIndexedOp(const char* mnem, ExtMachInst _machInst,
-                    OpClass __opClass, RegIndex _dest,
-                    RegIndex _op1, RegIndex _op2, uint16_t _imm) :
-        ArmStaticInst(mnem, _machInst, __opClass),
-        dest(_dest), op1(_op1), op2(_op2) , imm(_imm)
-    {}
-
-    std::string generateDisassembly(
-            Addr pc, const loader::SymbolTable *symtab) const override;
-};
-
-
-
 /// Ternary, destructive, unpredicated SVE instruction.
 class SveTerUnpredOp : public ArmStaticInst
 {
--- a/src/arch/arm/insts/sve.hh
+++ b/src/arch/arm/insts/sve.hh
@@ -498,6 +498,27 @@ class SveTerPredOp : public ArmStaticInst
             Addr pc, const loader::SymbolTable *symtab) const override;
 };
 
+/// Ternary, destructive, unpredicated , !INDEXED! SVE Instruction
+class SveTerIndexedOp : public ArmStaticInst
+{
+  protected:
+    RegIndex dest, op1, op2;
+    uint16_t imm;
+    uint8_t esize;
+
+    SveTerIndexedOp(const char* mnem, ExtMachInst _machInst,
+                    OpClass __opClass, RegIndex _dest,
+                    RegIndex _op1, RegIndex _op2, uint16_t _imm) :
+        ArmStaticInst(mnem, _machInst, __opClass),
+        dest(_dest), op1(_op1), op2(_op2) , imm(_imm)
+    {}
+
+    std::string generateDisassembly(
+            Addr pc, const loader::SymbolTable *symtab) const override;
+};
+
+
+
 /// Ternary, destructive, unpredicated SVE instruction.
 class SveTerUnpredOp : public ArmStaticInst
 {
--- a/src/arch/arm/isa/formats/sve_2nd_level.isa
+++ b/src/arch/arm/isa/formats/sve_2nd_level.isa
@@ -246,6 +246,59 @@ namespace Aarch64
     }  // decodeSveIntMulAdd
 
     StaticInstPtr
+    decodeSveMultiplyAccIndexed(ExtMachInst machInst)
+    {
+        RegIndex zda = (RegIndex) (uint8_t) bits(machInst, 4, 0);
+        RegIndex zn = (RegIndex) (uint8_t) bits(machInst, 9, 5);
+        RegIndex zm ;
+
+        //= (RegIndex) (uint8_t) bits(machInst, 19, 16);
+
+        uint8_t size = bits(machInst, 23, 22);
+        uint16_t imm;
+        uint8_t opc = (bits(machInst, 10));
+
+       switch(size) {
+            case 0b00:
+            case 0b01:
+                zm = (RegIndex)(uint8_t)bits(machInst, 18, 16);
+                imm = (uint16_t)(bits(machInst, 22) << 2)
+                | bits(machInst, 20, 19);
+                switch(opc) {
+                    case 0x0: return new Sve2Mlai<int16_t>(
+                        machInst, zda, zn, zm, imm);
+                    case 0x1: return new Sve2Mlsi<int16_t>(
+                        machInst, zda, zn, zm, imm);
+                }
+                break;
+
+            case 0b10:
+                zm = (RegIndex)(uint8_t)bits(machInst, 18, 16);
+                imm = (uint16_t)bits(machInst, 20, 19);
+                switch(opc) {
+                    case 0x0: return new Sve2Mlai<int32_t>(
+                        machInst, zda, zn, zm, imm);
+                    case 0x1: return new Sve2Mlsi<int32_t>(
+                        machInst, zda, zn, zm, imm);
+                }
+                break;
+
+            case 0b11:
+                zm = (RegIndex)(uint8_t)bits(machInst, 19, 16);
+                imm = (uint16_t)bits(machInst, 20);
+                switch(opc) {
+                    case 0x0: return new Sve2Mlai<int64_t>(
+                        machInst, zda, zn, zm, imm);
+                    case 0x1: return new Sve2Mlsi<int64_t>(
+                        machInst, zda, zn, zm, imm);
+                }
+            break;
+        }
+        return new Unknown64(machInst);
+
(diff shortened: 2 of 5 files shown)


powerjg:
@giactra are you going to be able to review this before the staging branch is created on Friday?

andysan:
> @giactra are you going to be able to review this before the staging branch is created on Friday?

We'll need to run some arch compliance tests on this. Unfortunately, I don't think we'll be able to do that before the weekend.

powerjg:
Moving this to *not* in 23.1, then.

powerjg:
@andysan now that we have more testing bandwidth, can we make these compliance tests public? It would improve the workflow for people if they could see test results and we could run them.

andysan:
> @andysan now that we have more testing bandwidth, can we make these compliance tests public? It would improve the workflow for people if they could see test results and we could run them.

Sorry, that's unfortunately not possible. They are produced by our internal test teams and not suitable for public consumption.

giactra:
Just came back from a long vacation; will find some time today to review the PR

src/arch/arm/insts/sve.hh
```diff
@@ -498,6 +498,27 @@ class SveTerPredOp : public ArmStaticInst
             Addr pc, const loader::SymbolTable *symtab) const override;
 };
 
+/// Ternary, destructive, unpredicated , !INDEXED! SVE Instruction
+class SveTerIndexedOp : public ArmStaticInst
+{
+  protected:
+    RegIndex dest, op1, op2;
+    uint16_t imm;
+    uint8_t esize;
```
giactra: Other than esize (which is not necessary), this is the same as SveBinIdxUnpredOp. I would suggest implementing MLA/S (idx) on top of SveBinIdxUnpredOp

src/arch/arm/isa/templates/sve.isa
```diff
@@ -568,6 +606,8 @@ class %(class_name)s : public %(base_class)s
 };
 }};
 
+
+
```
giactra: Please remove these newlines

src/arch/arm/isa/insts/sve.isa
```diff
@@ -3094,6 +3122,9 @@ let {{
             substDict = {'targs': type, 'class_name': 'Sve' + Name}
             exec_output += SveOpExecDeclare.subst(substDict)
 
+
+
+
```
giactra: Please remove these newlines

src/arch/arm/isa/formats/sve_2nd_level.isa
```diff
@@ -245,6 +245,59 @@ namespace Aarch64
...
 
+    StaticInstPtr
+    decodeSveMultiplyAccIndexed(ExtMachInst machInst)
+    {
+        RegIndex zda = (RegIndex) (uint8_t) bits(machInst, 4, 0);
+        RegIndex zn = (RegIndex) (uint8_t) bits(machInst, 9, 5);
+        RegIndex zm ;
+
+        //= (RegIndex) (uint8_t) bits(machInst, 19, 16);
+
```
giactra: You could remove this

src/arch/arm/isa/formats/sve_2nd_level.isa
```diff
@@ -245,6 +245,59 @@ namespace Aarch64
...
+        uint8_t size = bits(machInst, 23, 22);
+        uint16_t imm;
+        uint8_t opc = (bits(machInst, 10));
+
+       switch(size) {
+            case 0b00:
+            case 0b01:
+                zm = (RegIndex)(uint8_t)bits(machInst, 18, 16);
+                imm = (uint16_t)(bits(machInst, 22) << 2)
```
giactra: I undertand zm and imm depend on the element type; may I suggest you to define them within the switch case?
Something like:

       switch(size) {
            case 0b00:
            case 0b01: {
                RegIndex zm = (RegIndex)(uint8_t)bits(machInst, 18, 16);
                RegIndex imm = (uint16_t)(bits(machInst, 22) << 2)

                [...]
            }



I would like to avoid having some code using uninitialized variables by accident

src/arch/arm/isa/templates/sve.isa
```diff
@@ -515,6 +515,44 @@ class %(class_name)s : public %(base_class)s
...
+        %(set_reg_idx_arr)s;
+        %(constructor)s;
+        esize = sizeof(Element);
+    }
+
+    Fault execute(ExecContext *, trace::InstRecord *) const override;
+};
+}};
+
+
```
giactra: If we use sveBinIdxInst this won't be needed as the python generator will automatically pick SveBinIdxUnpredOpDeclare which does its job

src/arch/arm/isa/insts/sve.isa
```diff
@@ -2096,6 +2096,34 @@ let {{
...
+        }'''
+        iop = ArmInstObjParams(name, 'Sve2' + Name+ 'i', 'SveTerIndexedOp',
+                               {'code': code, 'op_class': opClass}, [])
+        header_output += SveTerIndexedOpDeclare.subst(iop)
+        exec_output += SveOpExecute.subst(iop)
+        for type in types:
+            substDict = {'targs' : type,
+                         'class_name' : 'Sve2' + Name + "i"}
+            exec_output += SveOpExecDeclare.subst(substDict)
+
```
giactra: This is effectively the same as sveBinIdxInst.. I believe we should reuse existing code

giactra (changes requested):
First of all thanks for your contribution and apologies if it took me some time to review it.

I believe it looks good from a functional perspective but there is a lot of duplication with existing code. This is not your fault; the template library has grown chaotically and the chance of duplicating classes and functions is extremely high. Please check my comments to see what can be rewritten with existing logic

src/arch/arm/isa/insts/sve.isa
Nitesh8998: Hi,

Thank you for your suggestions, while I make the changes I have a few observations I would like to bring to you.
the sveBinIdxInst function, here would might not work as it does the following 

```
Element destElem = 0;
```
Instead of 
```
Element destElem = AA64FpDestMerge_x[i];
```

which is needed for mla/s indexed. However, as you mentioned the template library does have some repeated code and I see a  ```sveTerIdxInst``` function that existed before and solves the purpose I shall use that.


Also, I have another observation and would like to know more about why the implementation is done this way. For both ```sveTerIdxInst``` and ```sveBinIdxInst``` there is a temporary  ```VegRegContainer``` created as ```auxDest``` ( seems like auxiliary destination), which then is written into ``` AA64FpDest_x[i] ``` using a separate for loop as below:

```
    Element destElem = AA64FpDestMerge_x[i];
        '''
        code += '''
            %(op)s
            auxDest[i] = destElem;
        }

        for (unsigned i = 0; i < eCount; i++) {
            AA64FpDest_x[i] = auxDest[i];
        }''' % {'op': op}
```

Is there a reason why this is done instead of using ```AA64FpDest_x[i]``` in the previous for-loop (like other instructions that use the ```sveTerInst``` function?

Let me know.

Regards

giactra: > Hi,
> 
> Thank you for your suggestions, while I make the changes I have a few observations I would like to bring to you. the sveBinIdxInst function, here would might not work as it does the following
> 
> ```
> Element destElem = 0;
> ```
> 
> Instead of
> 
> ```
> Element destElem = AA64FpDestMerge_x[i];
> ```
> 

Oh right I missed that!

> which is needed for mla/s indexed. However, as you mentioned the template library does have some repeated code and I see a `sveTerIdxInst` function that existed before and solves the purpose I shall use that.
> 

That's great!

> Also, I have another observation and would like to know more about why the implementation is done this way. For both `sveTerIdxInst` and `sveBinIdxInst` there is a temporary `VegRegContainer` created as `auxDest` ( seems like auxiliary destination), which then is written into `AA64FpDest_x[i]` using a separate for loop as below:
> 
> ```
>     Element destElem = AA64FpDestMerge_x[i];
>         '''
>         code += '''
>             %(op)s
>             auxDest[i] = destElem;
>         }
> 
>         for (unsigned i = 0; i < eCount; i++) {
>             AA64FpDest_x[i] = auxDest[i];
>         }''' % {'op': op}
> ```
> 
> Is there a reason why this is done instead of using `AA64FpDest_x[i]` in the previous for-loop (like other instructions that use the `sveTerInst` function?
> 
> Let me know.
> 
> Regards

This could potentially be an unnecessary copy. I have two possible explanations:

1) To handle the case where the source = destination register, meaning

`AA64FpDest_x[i] = destElem`

might change the source register value while still iterating. This is not the case anyway, as while writing to the destination register changes the physical register, the source operands are copied anyway into temporaries tmp_s* with getRegOperand

2) To handle FP exceptions thrown while iterating over the vectors. In case of an exception we don't want to have an inconsistent register state.. FP exceptions are not implemented in Arm anyway so again, there seem to be non need for that.

I will run some tests and see if we can safely remove that auxiliary vector.. Anyway it should be indipendent from this PR. Thanks for flagging it though!

src/arch/arm/isa/templates/sve.isa
```diff
@@ -515,6 +515,10 @@ class %(class_name)s : public %(base_class)s
 };
 }};
 
+
+
+
+
```
giactra: Please remove newlines

src/arch/arm/isa/templates/sve.isa
```diff
@@ -568,6 +572,8 @@ class %(class_name)s : public %(base_class)s
 };
 }};
 
+
+
```
giactra: Ditto

src/arch/arm/isa/insts/sve.isa
```diff
@@ -3094,6 +3094,9 @@ let {{
             substDict = {'targs': type, 'class_name': 'Sve' + Name}
             exec_output += SveOpExecDeclare.subst(substDict)
 
+
+
+
```
giactra: Please remove newlines

giactra (changes requested):
Thanks Nitesh, it looks good to me, some minor coding style issues to fix (basically removing newlines addition) and then we are free to merge this

src/arch/arm/isa/insts/sve.isa
Nitesh8998: Thank you for clearing the doubts, do let me know if I can do anything regarding this. I would like to be involved with it.

giactra (approved):

PR #597, opened 2023-11-24
dev-amdgpu: Writeback PM4 queue rptr when empty

abmerop:
The GPU device keeps a local copy of each ring buffers read pointer (rptr) to avoid constant DMAs to/from host memory. This means it needs to be periodically updated on the host side as the driver uses this to determine how much space is left in the queue and may hang if it believe the queue is full. For user-mode queues, this already happens when queues are unmapped. For kernel mode queues (e.g., HIQ, KIQ) the rptr is never updated leading to a hang.

In this patch the rptr for *all* queues is reported back to the kernel whenever the queue reaches an empty state (rptr == wptr). Additionally to handle PM4 queue wrap-around, the queue processing function checks if the queue is not empty instead of rptr < wptr. This is state because the driver fills PM4 queues with NOP packets on initialization and when wrap around occurs.

Change-Id: Ie13a4354f82999208a75bb1eaec70513039ff30f

### DIFF
--- a/src/dev/amdgpu/pm4_packet_processor.cc
+++ b/src/dev/amdgpu/pm4_packet_processor.cc
@@ -168,7 +168,7 @@ PM4PacketProcessor::decodeNext(PM4Queue *q)
     DPRINTF(PM4PacketProcessor, "PM4 decode queue %d rptr %p, wptr %p\n",
             q->id(), q->rptr(), q->wptr());
 
-    if (q->rptr() < q->wptr()) {
+    if (q->rptr() != q->wptr()) {
         /* Additional braces here are needed due to a clang compilation bug
            falsely throwing a "suggest braces around initialization of
            subject" error. More info on this bug is available here:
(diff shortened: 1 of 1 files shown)


src/dev/amdgpu/pm4_packet_processor.cc
```diff
@@ -181,11 +181,28 @@ PM4PacketProcessor::decodeNext(PM4Queue *q)
...
         q->processing(false);
         if (q->ib()) {
             q->ib(false);
             decodeNext(q);
         }
+
+        // Write back rptr when the queue is empty. For static queues which
+        // are not unmapped, this is how the driver knows there is enough
+        // space in the queue to continue writing packets to the ring buffer.
+        if (q->getMQD()->aqlRptr) {
```
mattsinc, line 196: I'm not sure how this check confirms that the queue is empty?

abmerop: It will be empty it rptr == wptr by HSA spec. -- This is the != change above.  Basically, this is just a nullptr check.  The HSA spec basically says "it's up to the device" when to report back the rptr.  As a simulator there isn't much downside to report it as much as possible.

abmerop: Most apps do not see this case. I used the bug label after running an app that uses (abuses?) 64 HSA queues

mattsinc (approved):

PR #598, opened 2023-11-24
mem-ruby: Unused L3CacheCntrl freed

mysoreanoop:
Seems like the MOESI_AMD_Base-L3Cache.sm file is unused in the VIPER protocol. It's confusing to have it in the GPU_VIPER.slicc file.

### DIFF
--- a/configs/ruby/GPU_VIPER.py
+++ b/configs/ruby/GPU_VIPER.py
@@ -338,6 +338,7 @@ class L3Cache(RubyCache):
         self.replacement_policy = TreePLRURP()
 
 
+# unused in GPU_VIPER; see git blame for discussion
 class L3Cntrl(L3Cache_Controller, CntrlBase):
     def create(self, options, ruby_system, system):
         self.version = self.versionCount()
--- a/src/mem/ruby/protocol/GPU_VIPER.slicc
+++ b/src/mem/ruby/protocol/GPU_VIPER.slicc
@@ -8,4 +8,4 @@ include "GPU_VIPER-msg.sm";
 include "GPU_VIPER-TCP.sm";
 include "GPU_VIPER-SQC.sm";
 include "GPU_VIPER-TCC.sm";
-include "MOESI_AMD_Base-L3cache.sm";
+include "MOESI_AMD_Base-L3cache.sm"; // unused in GPU_VIPER; see git blame for discussion


mattsinc (commented):
I don't know how to reject a commit on Github, but if I could I would.  I do not think this should be merged in for several reasons:

- while coderot is important, removing code because "it seems like it should go away" is not a good reason.
- I believe some of AMD's internal gem5 protocols do use this code, thus removing it breaks their compatibility
- If users in the future want to change the setup, they would then have to add this code back in before they could start doing so
- if a user understands this code well enough to reason about why it should be removed, it contradicts the claim in the commit message that it is confusing.

Happy to be convinced otherwise, but this seems unnecessary and I am not in favor of it being changed.  Others welcome to disagree though.

mysoreanoop:
Closing, with a few clarifications:
1. My intention was to improve readability/save a second or two for new developers.
2. No comment
3. No comment
4. Understanding the code base well enough usually takes time, and having more files to look into, I thought, perhaps lengthens it. It's confusing for new users is what I meant; maybe I'm wrong.

I understand your concerns. Thanks for taking the time to review this.

mysoreanoop:
Sorry for reopening, but would you be okay with just an addition of a comment in the slicc file indicating that it is unused?
It would keep all the parties involved happy, and also additionally help new users.

Plus, I'm not sure if having an instantiation of the L3 controller in the config file would lead to any wasted memory, but it seems like it perhaps wouldn't if there's any lazy loading thing in place -- perhaps somebody can look into it?

powerjg:
Just FYI, this *is* used in AMD_MOESI_base: https://github.com/gem5/gem5/blob/783bae7c39339a168dc6d871895a4dd89aa42c78/src/mem/ruby/protocol/MOESI_AMD_Base.slicc#L5

State machines in SLICC can be used in more than one protocol. So, just because a state machine doesn't appear in a particular protocol doesn't mean it's unused.

mysoreanoop:
But that's a different protocol altogether, right? 
It is unused in GPU_VIPER.slicc is what I meant. And the changes proposed were still within GPU_VIPER. Please correct me if I'm wrong.

powerjg:
Hmm, I see what you mean @mysoreanoop. 

@mattsinc why not delete the unused code? If someone is using this internally, they should have their own config script that actually uses it. The change in `GPU_VIPER.py` shouldn't hurt anything.

I agree with *not* removing the state machine from the `GPU_VIPER.slicc` file.

mattsinc:
> Hmm, I see what you mean @mysoreanoop.
> 
> @mattsinc why not delete the unused code? If someone is using this internally, they should have their own config script that actually uses it. The change in `GPU_VIPER.py` shouldn't hurt anything.
> 
> I agree with _not_ removing the state machine from the `GPU_VIPER.slicc` file.

@abmerop do you need this kept around for internal AMD stuff?

If not, and if others feel strongly this is a worthwhile change, I won't stand in the way.

abmerop:
I am not using VIPER, so I doubt anyone would notice changes to VIPER

configs/ruby/GPU_VIPER.py
```diff
@@ -338,6 +338,7 @@ def create(self, options, ruby_system, system):
         self.replacement_policy = TreePLRURP()
 
 
+# unused in GPU_VIPER; see git blame for discussion
```
mysoreanoop, line 341: Shall I remove this (instead of commenting as unused) then? @abmerop

abmerop: I don't have a strong preference, and it seems @mattsinc is strongly against removal.  I think adding the comment (what you have now) is a good compromise.  Will give Matt one more opportunity to review but it looks fine to me

mysoreanoop: Cool, then it's ready to merge, I think

abmerop: Agreed, I will let Matt comment one more time (likely tomorrow).  In the meantime it looks like one of the commits are missing a Change-ID.  I'd also recommend squashing the commits as it looks like one is essentially undoing the other. Thanks,

mattsinc: If others don't think it hurts their protocols and think this is fine, then we should just merge it -- once the issue @abmerop raised is fixed.

Harshil2107:
Hi @mysoreanoop ,

```
f=.git/hooks/commit-msg
mkdir -p 
chmod +x $f/commit-msg
```

ivanaamit:
@mysoreanoop, are you going to be able to make these changes before the staging branch is created tomorrow?

mysoreanoop:
Yes, I'll be able to do that in a couple of hours.

mysoreanoop:
@Harshil2107 is this okay?

configs/ruby/GPU_VIPER.py
mysoreanoop: Added the change ID; not sure if the squash is something I should do -- seems like one of the maintainers should? Otherwise please let me know.

ivanaamit: @mysoreanoop, you should squash the commits. Please let me know if you have trouble with squashing.

https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

ivanaamit (approved):

PR #599, opened 2023-11-26
arch-vega,arch-gcn3: Bugfix V_PERM_B32 and V_OR3_B32

abmerop:
The V_PERM_B32 instruction is selecting the correct byte, but is shifting into place moving by bits instead of bytes. The V_OR3_B32 instruction is calling the wrong instruction implementation in the decoder.

This patch fixes both issues plus a bonus fix for GCN3's V_PERM_B32. (GCN3 does not have V_OR3_B32).

Change-Id: Ied66c43981bc4236f680db42a9868f760becc284


abmerop:
no gpu expertise required for this review...

mattsinc (approved):

PR #601, opened 2023-11-27
scons: Add an option to reduce memory usage of ld

hnpl:
Linking the gem5 binary consists of linking numerous object files. By default, ld optimizes for speed over for memory consumption [1]. This leads to the huge memory consumption of gem5 at the linking stage.

This patch adds an option to add the `--no-keep-memory` flag to ld. According to the documentation [1], this flag optimizes the memory usage of ld.

[1] https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_chapter/ld_2.html#IDX133

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -131,6 +131,8 @@ AddOption('--with-systemc-tests', action='store_true',
           help='Build systemc tests')
 AddOption('--install-hooks', action='store_true',
           help='Install revision control hooks non-interactively')
+AddOption('--limit-ld-memory-usage', action='store_true',
+          help='Tell ld, the linker, to reduce memory usage.')
 AddOption('--gprof', action='store_true',
           help='Enable support for the gprof profiler')
 AddOption('--pprof', action='store_true',
@@ -587,6 +589,14 @@ for variant_path in variant_paths:
                     conf.CheckLinkFlag('-Wl,--threads')
                     conf.CheckLinkFlag(
                             '-Wl,--thread-count=%d' % GetOption('num_jobs'))
+
+        with gem5_scons.Configure(env) as conf:
+            ld_optimize_memory_usage = GetOption('limit_ld_memory_usage')
+            if ld_optimize_memory_usage:
+                if conf.CheckLinkFlag('-Wl,--no-keep-memory'):
+                    env.Append(LINKFLAGS=['-Wl,--no-keep-memory'])
+                else:
+                    error("Unable to use --no-keep-memory with the linker")
 
         # Treat warnings as errors but white list some warnings that we
         # want to allow (e.g., deprecation warnings).


hnpl:
When this flag was set, compiling gem5 worked on my docker container, which had 6GiB of RAM. Without the flag, compiling gem5 failed. 

Note: The build process still occasionally failed due to out of memory, but one can run scons again with fewer CPUs to fix this problem. This patch primarily aims to fix the memory usage of the linking stage at the end of the compilation process.

shingarov:
Yeah, this is one of those `--which-way-do-you-want-me-to-suck` options on the part of `ld`, where there isn't really a universally satisfactory solution.  If one wanted to "be even more helpful"™, they could go even further and choose the default based on parsing the `MemFree:` line of `/proc/meminfo` — and quickly reach the peak of absurd hairiness.

Hmmm.  Why is `--limit-ld-memory-usage` a better flag name than `--no-keep-memory`?  If the user is actively looking for a way to tell `ld` to moderate its RAM appetite, they most probably know it's exactly `ld` that is running out of memory, and possibly already know about its `--no-keep-memory` flag (or at least it's not unreasonable to expect they'd `man ld`).  If so, perhaps a more general solution would be to give the user a way to pass arbitrary extra flags to `ld`, something like `--extra-ld-flags`?  And then in the `help=` mention your scenario as one common usecase?  I am not saying I like this way better than your diff, I am just trying to explore/brainstorm different options.

hnpl:
I appreciate your opinion on this. I agree that this is far from an ideal solution. Generally, I think a compilation flag should be "utilize this tool" rather than "utilize this tool with this specific behavior". However, I'd argue for this case as currently, compiling gem5 has a hard memory capacity constraint, and this is a way to alleviate that problem.

Regarding the naming of the flag, I don't think`--limit-ld-memory-usage` is the best name as it's just too specific to `ld`. I also don't know if linkers other than `ld` and `gold` support the `--no-keep-memory` flag. Though, I do think the `--no-keep-memory` name is too ambiguous for users of gem5 that do not use linker a lot. To be honest, it took me way time more than I expected to figure out the name that flag so I don't know if the `--no-keep-memory` is that well-known.

Opening a way to add to linker flags from `scons` is interesting. I don't have a strong opinion about this. I can see it adds a lot of flexibility to compile gem5, but it also opens a lot of ways to compile gem5 that we don't want to support.

Harshil2107:
Hi @hnpl ,

```
f=.git/hooks/commit-msg
mkdir -p 
chmod +x $f/commit-msg
```

hnpl:
Thanks @Harshil2107, I updated the commit with the change id.

powerjg (approved):

powerjg:
@BobbyRBruce with this change, if we add this option when compiling with gcc 9 we should be able to add that back to the supported compilers and test it again. I'm not sure if we care to do that or not, though.

BobbyRBruce (approved):

PR #609, opened 2023-11-27
resources,stdlib: update suites to lazy download

Harshil2107:
- suites will lazily download all the resources when the __iter__ method is called instead of download all of them upon initialization.

Change-Id: I46a2fe00712818cda6282a71c1470f596cb1f06c

### DIFF
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -650,11 +650,15 @@ class SuiteResource(AbstractResource):
 
     def __init__(
         self,
-        workloads: Dict["WorkloadResource", Set[str]] = {},
+        workloads: Dict[Tuple[str, str], "WorkloadResource"] = {},
+        workloads_id_map: Dict[Tuple[str, str], Set[str]] = {},
         resource_version: Optional[str] = None,
         description: Optional[str] = None,
         source: Optional[str] = None,
         id: Optional[str] = None,
+        resource_directory: Optional[str] = None,
+        clients: Optional[List] = None,
+        gem5_version: Optional[str] = core.gem5Version,
         **kwargs,
     ) -> None:
         """
@@ -672,6 +676,10 @@ class SuiteResource(AbstractResource):
         self._description = description
         self._source = source
         self._resource_version = resource_version
+        self._workloads_id_map = workloads_id_map
+        self._resource_directory = resource_directory
+        self._clients = clients
+        self._gem5_version = gem5_version
 
         super().__init__(
             id=id,
@@ -680,13 +688,28 @@ class SuiteResource(AbstractResource):
             resource_version=resource_version,
         )
 
+    def _download_workloads(self) -> None:
+        """
+        Download all the workloads in the suite.
+        """
+        for workload_id_ver in self._workloads_id_map.keys():
+            if self._workloads[workload_id_ver] is None:
+                self._workloads[workload_id_ver] = obtain_resource(
+                    workload_id_ver[0],
+                    resource_version=workload_id_ver[1],
+                    resource_directory=self._resource_directory,
+                    clients=self._clients,
+                    gem5_version=self._gem5_version,
+                )
+
     def __iter__(self) -> Generator["WorkloadResource", None, None]:
         """
         Returns a generator that iterates over the workloads in the suite.
 
         :yields: A generator that iterates over the workloads in the suite.
         """
-        yield from self._workloads.keys()
+        self._download_workloads()
+        yield from self._workloads.values()
 
     def __len__(self):
         """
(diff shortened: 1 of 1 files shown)


powerjg:
Put this on the list for 23.1!

src/python/gem5/resources/resource.py
```diff
@@ -680,25 +688,50 @@ def __init__(
...
 
         :returns: The number of workloads in the suite.
         """
-        return len(self._workloads)
+        return len(self._workloads_id_map)
 
     def get_category_name(cls) -> str:
         return "SuiteResource"
 
+    def get_constituent_workloads(self) -> List[Tuple[str, str]]:
```
powerjg: How about `get_workload_ids`

src/python/gem5/resources/resource.py
```diff
@@ -650,11 +650,15 @@ class SuiteResource(AbstractResource):
 
     def __init__(
         self,
-        workloads: Dict["WorkloadResource", Set[str]] = {},
+        workloads: Dict[Tuple[str, str], "WorkloadResource"] = {},
+        workloads_id_map: Dict[Tuple[str, str], Set[str]] = {},
```
powerjg, line 654: **MUST** update the documentation below. This new parameter is not obvious at all. Please update for all new parameters :)

src/python/gem5/resources/resource.py
```diff
@@ -650,16 +650,22 @@ class SuiteResource(AbstractResource):
...
         resource_version: Optional[str] = None,
         description: Optional[str] = None,
         source: Optional[str] = None,
         id: Optional[str] = None,
+        resource_directory: Optional[str] = None,
+        clients: Optional[List] = None,
+        gem5_version: Optional[str] = core.gem5Version,
```
powerjg, line 661: These new parameters are also not included in the documentation. Please update.

powerjg (changes requested):

powerjg (approved):

BobbyRBruce (changes requested):
I don't see the point in this. With this change we're downloading all the workloads in a suite the first time we ever iterate through it. Why? This is no better than downloading all the workloads on construction and adds   lot of confusing logic to get there.

Harshil2107:
The main reason is that we download all the workloads for a suite that we might not need and this takes a lot of time. For example the micro benchmarks suite has 40+ workloads with different binaries. With this change we can filter the suite with input groups and get information about the workloads in the suite without downloading the resources. We will only download when we call the `__iter__` that is when they actually want to use the suite.

An example of this would be that in our configs/example/gem5_library/riscvmatched-microbenchmark-suite.py example the line:
`board.set_workload(list(microbenchmarks.with_input_group("cca"))[0])`

will download all the resources for all the workloads when we only want one workload with the "cca" input group. With this change on the filtered workloads will be downloaded that is just one workload with the input group "cca"

BobbyRBruce:
> The main reason is that we download all the workloads for a suite that we might not need and this takes a lot of time. For example the micro benchmarks suite has 40+ workloads with different binaries. With this change we can filter the suite with input groups and get information about the workloads in the suite without downloading the resources. We will only download when we call the `__iter__` that is when they actually want to use the suite.
> 
> An example of this would be that in our configs/example/gem5_library/riscvmatched-microbenchmark-suite.py example the line: `board.set_workload(list(microbenchmarks.with_input_group("cca"))[0])`
> 
> will download all the resources for all the workloads when we only want one workload with the "cca" input group. With this change on the filtered workloads will be downloaded that is just one workload with the input group "cca"

Is there not a way to just download the data upon construction but only download the actual resource (Disk Images, etc.) when we actually need to load it into the simulation?

Harshil2107:
> > The main reason is that we download all the workloads for a suite that we might not need and this takes a lot of time. For example the micro benchmarks suite has 40+ workloads with different binaries. With this change we can filter the suite with input groups and get information about the workloads in the suite without downloading the resources. We will only download when we call the `__iter__` that is when they actually want to use the suite.
> > An example of this would be that in our configs/example/gem5_library/riscvmatched-microbenchmark-suite.py example the line: `board.set_workload(list(microbenchmarks.with_input_group("cca"))[0])`
> > will download all the resources for all the workloads when we only want one workload with the "cca" input group. With this change on the filtered workloads will be downloaded that is just one workload with the input group "cca"
> 
> Is there not a way to just download the data upon construction but only download the actual resource (Disk Images, etc.) when we actually need to load it into the simulation?

This would be a much bigger  change for all the resources, I assume you mean that we only download the resource when a first call to it is actually made?

powerjg:
> Is there not a way to just download the data upon construction but only download the actual resource (Disk Images, etc.) when we actually need to load it into the simulation?

This is exactly the feedback that I had 😆

BobbyRBruce:
I've created an alternative approach here: https://github.com/gem5/gem5/pull/625. Let me know what you think and if it meets the requirements.

BobbyRBruce:
Superseded by #625

PR #611, opened 2023-11-27
stdlib, resources: removed  deprecated if statement  in obtain_resource for workload resources

Harshil2107:
- The resources field in workload now changed to a dict of id and version from a string with just the id.
There was an if statement added to support both versions in develop. Removing the if statement so that 23.1 supports the new changes only.

Change-Id: Id8dc3f932f53a156e4fb609a215db7d85bd81a44


Harshil2107:
Note: This is a draft PR as I am waiting to update the database with new versions of workload.

BobbyRBruce (approved):

PR #617, opened 2023-11-29
scons: Move CPPPATH systemc_home to "src/systemc" folder

rogerchang23424:
Files under src/systemc require the include path of systemc_home

Change-Id: Ibcbac2762259a0b997ac444b2c63a218c27af9ee


rogerchang23424:
It cloud not be correct solution

BobbyRBruce (approved):

PR #620, opened 2023-11-29
tests: switch lulesh to use vega_x86

Harshil2107:
Change-Id: Ifbf0fdfd7d8c2bbaad0b6094090acecd1cb8055c

### DIFF
--- a/.github/workflows/gpu-tests.yaml
+++ b/.github/workflows/gpu-tests.yaml
@@ -23,24 +23,6 @@ jobs:
           # therefore need to explicitly checkout the develop branch.
                   ref: develop
             - name: Build gem5
-              run: scons build/GCN3_X86/gem5.opt -j $(nproc) --ignore-style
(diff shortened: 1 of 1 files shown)


BobbyRBruce (changes requested):
Can the HACC tests use VEGA? In which case we could just update the `build-gem5` job?

abmerop:
> Can the HACC tests use VEGA? In which case we could just update the `build-gem5` job?

Anything working in GCN3 works in Vega the last time I tested, so I don't see why not

BobbyRBruce (approved):

mattsinc:
> > Can the HACC tests use VEGA? In which case we could just update the `build-gem5` job?
> 
> Anything working in GCN3 works in Vega the last time I tested, so I don't see why not

Sorry I am just seeing this after the ASPLOS deadline.  Looks good and yes in theory we should be able to move fully to VEGA.

PR #625, opened 2023-11-30
stdlib: Mv resource download to `get_local_path` and add `ShadowResource`

BobbyRBruce:
This change decouple's the downloading of a resource from it's data. With this change the `obtain_resource` function returns the `AbstractResource` implementation which contains the data. The resource itself (e.g., the actual disk image, binary, file, etc.) is only downloaded to the host system, if not already present, upon the `get_local_path` call.

`get_local_path` is the function used by gem5 to ultimately load the resource into a simulation, therefore this change ensures we only download resources when they are loaded into a simulation.

This change is not ideal and comes with the following caveats:

1. The `downloader` function is created in `obtain_workload` and passed to the `AbstractResource` implementation for later use. This function comes with the following requirements:
    * The function will download the resource to `local_path`.
    * The function will not re-download the resources if already present
as this function is called _everytime_ `get_local_path` is called.
2. The directories needed to store `local_path` are created in `obtain_workload` regardless. Ergo even if the resource is not used and `get_local_path` is never called these directories are still created.

In keeping with this efficiency `ShadowResource` is introduced to allow the storing of just the resource ID and Version of a resource with additional information only obtained when requested.

### DIFF
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -27,6 +27,7 @@
 import json
 import os
 from abc import ABCMeta
+from functools import partial
 from pathlib import Path
 from typing import (
     Any,
@@ -94,6 +95,7 @@ class AbstractResource:
         local_path: Optional[str] = None,
         description: Optional[str] = None,
         source: Optional[str] = None,
+        downloader: Optional[partial] = None,
     ):
         """
         :param local_path: The path on the host system where this resource is
@@ -104,18 +106,17 @@ class AbstractResource:
         string should navigate users to where the source for this resource
         may be found. Not a required parameter. By default is None.
         :param resource_version: Version of the resource itself.
+        :param downloader: A partial function which is used to download the
+        resource. If set, this is called if the resource is not present at the
+        specified `local_path`.
         """
 
-        if local_path and not os.path.exists(local_path):
-            raise Exception(
-                f"Local path specified for resource, '{local_path}', does not "
-                "exist."
-            )
         self._id = id
         self._local_path = local_path
         self._description = description
         self._source = source
         self._version = resource_version
+        self._downloader = downloader
 
     def get_category_name(cls) -> str:
         raise NotImplementedError
@@ -134,7 +135,19 @@ class AbstractResource:
         return self._version
 
     def get_local_path(self) -> Optional[str]:
-        """Returns the local path of the resource."""
+        """Returns the local path of the resource.
+
+        If specified the `downloader` partial function is called to download
+        the resource if it is not present or up-to-date at the specified
+        `local_path`.
+        """
+        if self._downloader:
+            self._downloader()
+        if self._local_path and not os.path.exists(self._local_path):
+            raise Exception(
+                f"Local path specified for resource, '{self._local_path}', "
+                "does not exist."
+            )
         return self._local_path
 
     def get_description(self) -> Optional[str]:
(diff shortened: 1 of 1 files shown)


powerjg:
The approach seems sound to me. 

My one concern is the utilities that we use to download resources. I assume they are unaffected?

Also, it looks like you need to rebase :)

src/python/gem5/resources/resource.py
```diff
@@ -750,6 +807,52 @@ def get_input_groups(self) -> Set[str]:
...
+        """
+        Used to set or override a workload parameter
+
+        :param parameter: The parameter of the function to set.
+        :param value: The value to set to the parameter.
+        """
+        self._params[parameter] = value
+
+    def get_category_name(cls) -> str:
+        return "WorkloadResource"
```
powerjg: if this is only used for lazy workloads, then instead of `ShadowResource` this class should be something like `Lazy/ShadowWorkloadResource`

src/python/gem5/resources/resource.py
```diff
@@ -750,6 +807,52 @@ def get_input_groups(self) -> Set[str]:
...
+        return getattr(self._workload, attr)
+
+    def set_parameter(self, parameter: str, value: Any) -> None:
+        """
+        Used to set or override a workload parameter
+
+        :param parameter: The parameter of the function to set.
+        :param value: The value to set to the parameter.
+        """
+        self._params[parameter] = value
```
powerjg: Why do we need to override this function?

src/python/gem5/resources/resource.py
```diff
@@ -750,6 +807,52 @@ def get_input_groups(self) -> Set[str]:
...
+        """if getting the id or resource version, we keep the object in the
+        "shdow state" where the `obtain_resource` function has not been called.
+        When more information is needed by calling another attribute, we call
+        the `obtain_resource` function and store the result in the `_workload`.
+        """
+        if attr in {"get_id", "get_resource_version"}:
+            return getattr(super(), attr)
+        if not self._workload:
+            self._workload = self._obtain_resource_call()
+        return getattr(self._workload, attr)
```
powerjg, line 841: Could we put this code in the Resource class?

powerjg (commented):
I fear this implementation is now too complicated to maintain easily. However, I'll be happy to hear you out for why the complexity is needed.

Why do we need a new resource class?

src/python/gem5/resources/resource.py
BobbyRBruce: No, that was a mistake. Didn't mean to include it.

src/python/gem5/resources/resource.py
BobbyRBruce: I didn't mean to leave in these functions. Removing.

src/python/gem5/resources/resource.py
BobbyRBruce: I think we _could_ but i'm not sure we _should_. It does add a little more complexity and it's only use is in the `Suite` and, maybe, `Workload`s if we want to use it there too (even in this case I don't see much value in it)

BobbyRBruce:
> I fear this implementation is now too complicated to maintain easily. However, I'll be happy to hear you out for why the complexity is needed.
> 
> Why do we need a new resource class?

It's the least complex solution available to us. I don't particularly like it but the cost of having to get all workloads in a suite is too high. The only alternative to this would be a massive refactoring of our Client interaction to bundle resource requests.

In this design a Suite's workloads only store the ID and Version until that workload is accessed in any way other than to get the ID/Version. When this happens it is obtained via the `obtain_resource` function.

I think this is superior to #609 as when the Workloads are downloaded is more sensible (i.e., not just the first time the `Suite` is iterated through) and **only** the Workloads needed are ultimately downloaded. With this you can iterate through a suite and see the workloads at a surface level and only download their data when more detailed information is required.

src/python/gem5/resources/resource.py
powerjg: Ack

powerjg (approved):

PR #627, opened 2023-11-30
util-docker: Enforce cmake version >=3.24 for DRAMSys

BobbyRBruce:
DRAMSys requires cmake 3.24 or greater. By default neither Ubuntu 22.04 or 20.04 delevery this by APT.

In both cases wget is required. In 20.04 OpenSSL is required.


powerjg (commented):
This is *terribly* annoying, but I don't have a better solution.

We should also update the README in ext/DRAMSys to make this clear.

powerjg (approved):

powerjg:
I updated the README for you. As soon as you build the new images, this is good to go.

BobbyRBruce:
> I updated the README for you. As soon as you build the new images, this is good to go.

Thanks. The images are already rebuilt.

PR #630, opened 2023-11-30
stdlib: Integrate BootloaderKernelWorkload

hnpl:
This change does the following,

- Change the name of several python parameter names of the RiscvBootloaderKernelWorkload. This is done to conform the expectation from the stdlib, e.g., the kernel path must be `object_file`, and the boot parameter must be `command_line`.
- Use RiscvBootloaderKernelWorkload by default for all full system RISC-V simulations. RiscvBootloaderKernelWorkload is a superset of RiscvFsWorkload.

### DIFF
--- a/src/arch/riscv/RiscvFsWorkload.py
+++ b/src/arch/riscv/RiscvFsWorkload.py
@@ -67,9 +67,7 @@ class RiscvBootloaderKernelWorkload(Workload):
     bootloader_addr = Param.Addr(
         0x0, "Where to place the bootloader in memory."
     )
-    kernel_filename = Param.String(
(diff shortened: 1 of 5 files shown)


hnpl:
~This is a rough sketch of integrating the new workload type to the RISC-V board. It still has a bug where the base `configure_workload` function was called while it was `configure_workload` of the child class that should have been called.~

Edit: the change has been revised and this is no longer a problem.

powerjg (approved):
Looks good to me. More testing would be nice. I suppose we can do that on the staging branch, though.

BobbyRBruce (approved):

PR #390, opened 2023-10-04
arch-riscv: Add bootloader+kernel workload

hnpl:
Aims to boot OpenSBI + Linux kernel.

Change-Id: I9ee93cc367e8c06bdd0c7ddf43335d32965be14d

PR #635, opened 2023-12-01
arch-riscv: correctly pass arguments to kernel with new bootloader+kernel

aryanitish:
The [PR](https://github.com/gem5/gem5/pull/390) adds support for new bootloader and linux kernel. However after applying the changes the arguments are not passed correctly to the kernel resulting in kernel panic during simulations. This commit fixes the issue.

### DIFF
--- a/src/python/gem5/components/boards/riscv_board.py
+++ b/src/python/gem5/components/boards/riscv_board.py
@@ -256,7 +256,7 @@ class RiscvBoard(AbstractSystemBoard, KernelDiskWorkload):
             root.append(node)
 
         node = FdtNode(f"chosen")
-        bootargs = " ".join(self.get_default_kernel_args())
+        bootargs = self.workload.command_line
         node.append(FdtPropertyStrings("bootargs", [bootargs]))
         node.append(FdtPropertyStrings("stdout-path", ["/uart@10000000"]))
         root.append(node)


powerjg:
Related to #630

src/python/gem5/components/boards/riscv_board.py
```diff
@@ -256,7 +256,7 @@ def generate_device_tree(self, outdir: str) -> None:
             root.append(node)
 
         node = FdtNode(f"chosen")
-        bootargs = " ".join(self.get_default_kernel_args())
+        bootargs = self.workload.boot_args
```
powerjg: After #630 is merged, I think this should be
```suggestion
        bootargs = self.workload.command_line
```

We should double check with the `ArmBoard` to see what it uses.

hnpl: In #630, workload.boot_args was renamed to workload.command_line, so we should use workload.command_line here as well.

https://github.com/gem5/gem5/pull/630/files#diff-f50c1a48b07134c503b466730301cb8f8aa3b1c1f86fba977f2eecf350cd37b8L88

```suggestion
        bootargs = self.workload.command_line
```

hnpl (commented):
Thanks for the fix. There's a small detail related to #630 that needs to be updated, but this looks good.

powerjg (approved):

rogerchang23424 (approved):

powerjg (approved):

Harshil2107 (approved):

PR #636, opened 2023-12-01
Fix for gem5 Issue #550

rjc-arch:
This Pull-Request addresses gem5 Issue #550. The code that dumps the Dmesg buffer is now templated on the two variants of the `Metadata` structure, and the correct one is chosen based on the detected Kernel version.

To support this functionality, the pull request also adds Symbol Size data to the loader Symbol Table, and adds a method to query the Kernel Version from the image in guest memory. The new attributes in the Symbol class are de-serialized speculatively, so no checkpoint upgrader is required to support this change.

### DIFF
--- a/src/arch/arm/insts/static_inst.cc
+++ b/src/arch/arm/insts/static_inst.cc
@@ -401,8 +401,8 @@ ArmStaticInst::printTarget(std::ostream &os, Addr target,
     if (symtab) {
         auto it = symtab->findNearest(target);
         if (it != symtab->end()) {
-            ccprintf(os, "<%s", it->name);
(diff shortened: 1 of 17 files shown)


hnpl:
Hi @rjc-arch, this looks great. I guess the only missing part is to update the unit tests.

rjc-arch:
Hi @hnpl, I've updated the existing tests and it looks like they are passing now. I mistakenly thought the unit tests were run automatically with the regressions tests, so I didn't notice them failing.

However I'm afraid I don't think I'll have time to add new tests for the size member until next week.

powerjg (approved):

PR #647, opened 2023-12-03
misc: Merge Weekly GPU tests into Weekly Tests

BobbyRBruce:
This seperation was only for convenience while GPU tests were under development and rapidly changing. This test merges the GPU tests into the weekly tests where they belong.


.github/workflows/weekly-tests.yaml
```diff
@@ -11,6 +11,87 @@ on:
     workflow_dispatch:
 
 jobs:
+    build-gcn-gpu-gem5:
```
mattsinc, line 14: GCN is sort of not relevant anymore if we are moving completely to vega.  But this isn't an important enough distinction to require wholesale changes in my opinion.

.github/workflows/weekly-tests.yaml
```diff
@@ -11,6 +11,87 @@ on:
...
+
+    HACC-tests:
+        runs-on: [self-hosted, linux, x64]
+        container: ghcr.io/gem5/gcn-gpu:latest
+        needs: build-gcn-gpu-gem5
+        timeout-minutes: 120 # 2 hours
+        steps:
+            - uses: actions/checkout@v3
+              with:
+          # Scheduled workflows run on the default branch by default. We
```
mattsinc: Comments not indented properly?

.github/workflows/weekly-tests.yaml
```diff
@@ -11,6 +11,87 @@ on:
...
+            - name: make hip directory
+              run: mkdir hip
+            - name: Compile m5ops and x86
+              working-directory: ${{ github.workspace }}/util/m5
+              run: |
+                  export TERM=xterm-256color
+                  scons build/x86/out/m5
+            - name: Download tests
+              working-directory: ${{ github.workspace }}/hip
+              run: wget http://dist.gem5.org/dist/v22-1/test-progs/halo-finder/ForceTreeTest
```
mattsinc, line 89: Is v22-1 still correct?

.github/workflows/weekly-tests.yaml
```diff
@@ -11,6 +11,87 @@ on:
...
+              # gem5 Resources.
+              run: build/VEGA_X86/gem5.opt util/obtain-resource.py lulesh -p lulesh
+
+            - name: Run LULUESH tests
+              working-directory: ${{ github.workspace }}
+              run: |
+                  build/VEGA_X86/gem5.opt configs/example/apu_se.py -n3 --mem-size=8GB --reg-alloc-policy=dynamic --dgpu --gfx-version=gfx900 -c \
+                  lulesh --options="0.01 2"
+
+    HACC-tests:
```
mattsinc, line 64: How are you handling the previously separate Dockerfile for HACC?  I don't see anything here setting up the environment properly for it.

mattsinc (changes requested):
The name is somewhat of a misnomer because as far as I can tell this only merges HACC and LULESH in, instead of all the other GPU weekly tests.  Maybe they are already included?

.github/workflows/weekly-tests.yaml
abmerop: This might just be referring to the name of the docker image.  It seems to actually be building and using vega.  I've seen vega be called "GCN5" in various places so it's not a complete misnomer.

.github/workflows/weekly-tests.yaml
abmerop: These were added to the gcn-gpu docker image a while back, so the second Dockerfile is no longer needed/used.

.github/workflows/weekly-tests.yaml
abmerop: Also wondering about this.  I doubt the binary itself has changed since 22.1, but I don't know if each version has duplicate copies on dist.gem5.org.  If not, changing the version would 404.

.github/workflows/weekly-tests.yaml
BobbyRBruce: This is just a copy and paste job from the gpu-tests.yaml. Comment were not properly indented in there either. This is an artifact of our pre-commit YAML formatter which, annoyingly, doesn't care about comments. 

Might be worth fixing these comments in another PR.

.github/workflows/weekly-tests.yaml
BobbyRBruce: So, the situation with the URLs is my old versioning system here (i.e., in this case `dist/<version>/test-progs` has basically broken down. I don't create a new sub-directory per-release. Each resources now has a and ID and a Version and we have a database which maps these (among other things, e.g., a link to documentation and source) to a URL.

So, as long as this version works for you it's fine. What we eventually need to do is add `ForceTreeTest` to our gem5 Resources database to make this mapping explicit. In this case we can use the "util/obtain-resource.py` tool to automatically get this resources (as we do with lulush).

@Harshil2107 : Is there any reason not to do this for ForceTreeTest? Can you do it quickly so we can just include it in this change?

BobbyRBruce:
> The name is somewhat of a misnomer because as far as I can tell this only merges HACC and LULESH in, instead of all the other GPU weekly tests. Maybe they are already included?

Somewhat, yes. The other GPU tests are a bit stalled but can be included here directly. If I remember correctly the only reason for this separation is because Melissa was going over rapid iterations of these tests and didn't like having to deal with merge conflicts as the Weekly tests file was also doing the same. Things have stabilized a bit.

All-in-all this isn't a massively important change. It's just a bit of a cleanup.

.github/workflows/weekly-tests.yaml
mattsinc: I'm not sure why we wouldn't just fix the spacing for lines added in this commit, as part of this commit?

.github/workflows/weekly-tests.yaml
mattsinc: If this is where the binaries are, then I think this is fine and doesn't need to be immediately addressed.  But up to you @BobbyRBruce

mattsinc:
> > The name is somewhat of a misnomer because as far as I can tell this only merges HACC and LULESH in, instead of all the other GPU weekly tests. Maybe they are already included?
> 
> Somewhat, yes. The other GPU tests are a bit stalled but can be included here directly. If I remember correctly the only reason for this separation is because Melissa was going over rapid iterations of these tests and didn't like having to deal with merge conflicts as the Weekly tests file was also doing the same. Things have stabilized a bit.
> 
> All-in-all this isn't a massively important change. It's just a bit of a cleanup.

My comment was more that the commit message is incorrect rather than asking for more tests to be added.  If this commit "only" adds HACC and LULESH that is fine, but the commit message should reflect this.

.github/workflows/weekly-tests.yaml
Harshil2107: I can create a resource for the binary used in ForceTreeTest. Just to confirm the category would be binary?

.github/workflows/weekly-tests.yaml
BobbyRBruce: I've added a commit to this PR which fixes the comment indentation.

Harshil2107 (approved):
LGTM

mattsinc (approved):

PR #648, opened 2023-12-03
arch-x86: Revert Fix misc registers in mov instructions

BobbyRBruce:
Reverts gem5/gem5#593


BobbyRBruce:
I only created this PR to see if reverting this changed fixed a bug I'm observing with the X86 boot tests. This does not appear to be the case. I'm therefore closing.

PR #651, opened 2023-12-04
configs: Fix apu_se.py CPU type checks

abmerop:
The current checks do not work. Correct the CPU type names

Change-Id: I81778873df0567c4a8dabbbe659c4c7a39326f98

### DIFF
--- a/configs/example/apu_se.py
+++ b/configs/example/apu_se.py
@@ -398,8 +398,8 @@ if buildEnv["PROTOCOL"] == "None":
     fatal("GPU model requires ruby")
 
 # Currently the gpu model requires only timing or detailed CPU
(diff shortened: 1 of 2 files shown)


abmerop:
Fixes issue caused by #241 

FYI silly errors like this could be found immediately with a simple test which does not require docker or a new build since I think tests already build vega: `build/VEGA_X86/gem5.opt configs/example/apu_se.py -c /bin/ls`

We had a nearly identical issue on the last release so trying to prevent that happening again.

mattsinc (approved):
From my perspective this seems ok, but @BobbyRBruce should probably sign off on.

BobbyRBruce (approved):
Good catch! Thanks for the prompt fix.

PR #652, opened 2023-12-05
configs: Fix issues after get_runtime_isa() #241 removed

rogerchang23424:
1. Fix the wrong ISA detect of get_isa function
2. Fix the typo ObjectLIst.cpu_list
3. Fix missing PageTableWalkerCache
4. Fix the invalid default cpu_type paramter

Change-Id: I217ea8da8a6d8e712743a5b32c4c0669216ce6c4

### DIFF
--- a/configs/common/Caches.py
+++ b/configs/common/Caches.py
@@ -86,3 +86,14 @@ class IOCache(Cache):
     mshrs = 20
     size = "1kB"
     tgts_per_mshr = 12
+
+
+class PageTableWalkerCache(Cache):
+    assoc = 2
+    tag_latency = 2
+    data_latency = 2
+    response_latency = 2
+    mshrs = 10
+    size = "1kB"
+    tgts_per_mshr = 12
+    is_read_only = False
--- a/configs/ruby/Ruby.py
+++ b/configs/ruby/Ruby.py
@@ -331,8 +331,8 @@ def send_evicts(options):
     # 2. The x86 mwait instruction is built on top of coherence invalidations
     # 3. The local exclusive monitor in ARM systems
 
-    if isinstance(
-        options.cpu_type, DerivO3CPU
-    ) or ObjectList.CPUList().get_isa(options.cpu_type) in [ISA.X86, ISA.ARM]:
+    if isinstance(options.cpu_type, DerivO3CPU) or ObjectList.cpu_list.get_isa(
+        options.cpu_type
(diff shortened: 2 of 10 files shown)


giactra (dismissed):

configs/ruby/Ruby.py
```diff
@@ -331,8 +331,8 @@ def send_evicts(options):
     # 2. The x86 mwait instruction is built on top of coherence invalidations
     # 3. The local exclusive monitor in ARM systems
 
-    if isinstance(
-        options.cpu_type, DerivO3CPU
-    ) or ObjectList.CPUList().get_isa(options.cpu_type) in [ISA.X86, ISA.ARM]:
+    if isinstance(options.cpu_type, DerivO3CPU) or ObjectList.cpu_list.get_isa(
```
mattsinc, line 334: @abmerop just making sure you saw this, not sure why it didn't affect the GPU stuff too?

abmerop: I modified GPU_VIPER.py to basically do this instead of calling this function... Looks like this will fix the issue though. Admittedly I put that patch together as quickly as possible since release is imminent so it would be worth reverting that particular hunk at some point

configs/common/Caches.py
```diff
@@ -86,3 +86,14 @@ class IOCache(Cache):
...
+
+class PageTableWalkerCache(Cache):
+    assoc = 2
+    tag_latency = 2
+    data_latency = 2
+    response_latency = 2
+    mshrs = 10
+    size = "1kB"
+    tgts_per_mshr = 12
+    is_read_only = False
```
abmerop, line 99: Is this needed for the commit?

abmerop (commented):
Just one comment, otherwise looks good to me.  We should get this in for sure for 23.1.

configs/common/Caches.py
abmerop: Oops... Reread the commit message and understand now

abmerop (approved):

configs/ruby/Ruby.py
mattsinc: Do we want to remove the VIPER change then?  At some level if feels fine to let it be ...

abmerop: My thought process is I am hoping to switch to a VIPER stdlib board for v24.0 and therefore this won't be used after that

PR #655, opened 2023-12-06
configs: Make riscv/fs_linux work in build/ALL/gem5.opt

rogerchang23424:
Change-Id: If9add7dc5e9c5600f769d27817da41466158942b

### DIFF
--- a/configs/example/riscv/fs_linux.py
+++ b/configs/example/riscv/fs_linux.py
@@ -52,6 +52,8 @@ from m5.util import (
 )
 from m5.util.fdthelper import *
 
+from gem5.utils.requires import requires
+
 addToPath("../../")
 
 from common import (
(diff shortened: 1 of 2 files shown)


rogerchang23424:
Blocked by #652

configs/example/riscv/fs_linux.py
```diff
@@ -135,20 +140,8 @@ def generateDtb(system):
 
 # ----------------------------- Add Options ---------------------------- #
 parser = argparse.ArgumentParser()
-Options.addCommonOptions(parser)
+Options.addCommonOptions(parser, ISA.RISCV)
```
powerjg, line 143: No need to make any changes, but since this script uses `common/Options.py` it should probably be moved to "deprecated." 

The maintainers are explicitly not maintaining anything that uses `common/Options.py`. That said, I appreciate the community continuing to make these scripts work.

rogerchang23424: I decide to keep the parameter. In the multi-isa system, the default of cpu_type should be the base type of RISC-V CPU.

BobbyRBruce (approved):

powerjg:
@BobbyRBruce This needs to be cherry-picked to staging

PR #656, opened 2023-12-06
mem: Add a flag on AbstractMemory to control statistics collection

wmin0:
The stats initialization in the AbstractMemory allocates the space according to the max requestors of the System. This may cause issues in multiple system simulation.
Given there are two system A and B. A has one requestor and a memory, while B has two requestors. When the requestor with requestor id 2 sending requests to the meomry in A, the simulator would crash because requestor id 2 is out of the allocated space.

Current solution is adding a SysBridge between across A and B which would rewrite the requestor id to a valid one. This solution works but it needs to the bridge at the correct boundary which may not easy. In addition, the stats would record a mapped data which may not accurate.

To reduce the complexity, we add an flag to AbstractMemory to control the stats. If users don't want the statistics and want to solve the cross system issue simply, users can disable the statistics collection. We also makes the flag by default True to not disturb current users.

Change-Id: Ibb46a63d216d4f310b3e920815a295073496ea6e

### DIFF
--- a/src/mem/AbstractMemory.py
+++ b/src/mem/AbstractMemory.py
@@ -77,4 +77,11 @@ class AbstractMemory(ClockedObject):
 
     writeable = Param.Bool(True, "Allow writes to this memory")
 
-    collect_stats = Param.Bool(True, "Collect traffic statistics")
+    collect_stats = Param.Bool(
+        True,
+        "Collect statistics per requestor for "
+        "each type of access. Set this to `False` if "
+        "requestors may be unknown or when running "
+        "with multiple `System` objects without a "
+        "`SysBridge`.",
+    )
--- a/src/mem/AbstractMemory.py
+++ b/src/mem/AbstractMemory.py
@@ -76,3 +76,5 @@ class AbstractMemory(ClockedObject):
     )
 
     writeable = Param.Bool(True, "Allow writes to this memory")
+
+    collect_stats = Param.Bool(True, "Collect traffic statistics")
--- a/src/mem/abstract_mem.cc
+++ b/src/mem/abstract_mem.cc
@@ -433,8 +433,9 @@ AbstractMemory::access(PacketPtr pkt)
 
             assert(!pkt->req->isInstFetch());
             TRACE_PACKET("Read/Write");
-            if (collectStats)
+            if (collectStats) {
                 stats.numOther[pkt->req->requestorId()]++;
+            }
         }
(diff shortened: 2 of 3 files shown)


wmin0:
cc @yanleeG

src/mem/AbstractMemory.py
```diff
@@ -76,3 +76,5 @@ class AbstractMemory(ClockedObject):
     )
 
     writeable = Param.Bool(True, "Allow writes to this memory")
+
+    collect_stats = Param.Bool(True, "Collect traffic statistics")
```
powerjg:
```suggestion
    collect_stats = Param.Bool(True, "Collect statistics per requestor for "
                               "each type of access. Set this to `False` if "
                               "requestors may be unknown or when running "
                               "with multiple `System` objects without a "
                               "`SysBridge`.")
```

src/mem/abstract_mem.cc
```diff
@@ -447,10 +448,12 @@ AbstractMemory::access(PacketPtr pkt)
...
         TRACE_PACKET(pkt->req->isInstFetch() ? "IFetch" : "Read");
-        stats.numReads[pkt->req->requestorId()]++;
-        stats.bytesRead[pkt->req->requestorId()] += pkt->getSize();
-        if (pkt->req->isInstFetch())
-            stats.bytesInstRead[pkt->req->requestorId()] += pkt->getSize();
+        if (collectStats) {
+            stats.numReads[pkt->req->requestorId()]++;
+            stats.bytesRead[pkt->req->requestorId()] += pkt->getSize();
+            if (pkt->req->isInstFetch())
+                stats.bytesInstRead[pkt->req->requestorId()] += pkt->getSize();
```
powerjg: It will reduce the likelihood of future bugs to use curly braces around this if statement (i.e., it's not clear that the curly brace on line 456 is for the outer if statement and not this one).

powerjg (changes requested):
Look OK to me, but the added parameter needs more explanation. See suggestion below

src/mem/AbstractMemory.py
wmin0: Thanks for the suggestion. I updated the commit. Please check :)

src/mem/abstract_mem.cc
wmin0: Done

powerjg (approved):

BobbyRBruce (approved):

PR #658, opened 2023-12-06
stdlib: Fix the chi protocol of arm boot tests

rogerchang23424:
Change-Id: I63f17a73b2e16bc26d9b41babc63439a6040791f


giactra (approved):
This looks good to me; I am wondering if we shouldn't replicate configs/ruby setup which is binding together use_prefetcher and prefetcher with 

self.use_prefetcher = self.prefetcher != NULL

In the long term we should remove use_prefetcher anyway and just check if prefetcher holds a valid ptr

powerjg (approved):

BobbyRBruce (approved):

PR #663, opened 2023-12-07
arch-x86: Revert Fix misc registers in mov instructions

BobbyRBruce:
Reverts gem5/gem5#593


BobbyRBruce:
This changed introduced by #664.

BobbyRBruce:
I've found a more precise fix here: #666. Closing this fix. There's no need to revert the whole change.

PR #469, opened 2023-10-16
Improving stdlib

mahyarsamani:

PR #665, opened 2023-12-08
cpu,stdlib: Updating StridedGen functionality.

BobbyRBruce:
This change updates the functionality of StridedGen to allow for more flexible configuration of the generator. This change makes LinearGen obsolete.

This was cherry-picked from #469.

Change-Id: I0165da1f94b9bc390da1acbb53833f42b7c37f96

### DIFF
--- a/src/cpu/testers/traffic_gen/base.cc
+++ b/src/cpu/testers/traffic_gen/base.cc
@@ -527,19 +527,20 @@ BaseTrafficGen::createNvm(Tick duration,
 }
 
 std::shared_ptr<BaseGen>
-BaseTrafficGen::createStrided(Tick duration,
-                             Addr start_addr, Addr end_addr, Addr blocksize,
-                             Addr stride_size, int gen_id,
-                             Tick min_period, Tick max_period,
-                             uint8_t read_percent, Addr data_limit)
+BaseTrafficGen::createStrided(
+        Tick duration,
+        Addr start_addr, Addr end_addr, Addr offset,
+        Addr block_size, Addr superblock_size, Addr stride_size,
+        Tick min_period, Tick max_period,
+        uint8_t read_percent, Addr data_limit)
 {
-    return std::shared_ptr<BaseGen>(new StridedGen(*this, requestorId,
-                                                  duration, start_addr,
-                                                  end_addr, blocksize,
-                                                  system->cacheLineSize(),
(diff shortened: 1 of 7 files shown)


powerjg:
Ping: @mahyarsamani 

How has this been tested?

mahyarsamani:
> Ping: @mahyarsamani
> 
> How has this been tested?

Regarding the pattern this generates: I have manually tested it (used it in my experiments). I don't have a better idea that focuses on testing traffic generators though.

BobbyRBruce:
> > Ping: @mahyarsamani
> > How has this been tested?
> 
> Regarding the pattern this generates: I have manually tested it (used it in my experiments). I don't have a better idea that focuses on testing traffic generators though.

For something like this can't you write a test which just does a little simulation which utilizes them. At a minimum this would make sure they don't crash. Is having a ref of the output also possible? I assume we can fix any RNG involved here so we can verify nothing has changed.

mahyarsamani:
> > > Ping: @mahyarsamani
> > > How has this been tested?
> > 
> > 
> > Regarding the pattern this generates: I have manually tested it (used it in my experiments). I don't have a better idea that focuses on testing traffic generators though.
> 
> For something like this can't you write a test which just does a little simulation which utilizes them. At a minimum this would make sure they don't crash. Is having a ref of the output also possible? I assume we can fix any RNG involved here so we can verify nothing has changed.

Should I push that as another change on top of this pull request? Can I even? :)

BobbyRBruce:
> > > > Ping: @mahyarsamani
> > > > How has this been tested?
> > > 
> > > 
> > > Regarding the pattern this generates: I have manually tested it (used it in my experiments). I don't have a better idea that focuses on testing traffic generators though.
> > 
> > 
> > For something like this can't you write a test which just does a little simulation which utilizes them. At a minimum this would make sure they don't crash. Is having a ref of the output also possible? I assume we can fix any RNG involved here so we can verify nothing has changed.
> 
> Should I push that as another change on top of this pull request? Can I even? :)

Yes you should, and yes you can even.

mahyarsamani:
I couldn't figure out how to push to the same pull request. Therefore, I have created a new pr here: https://github.com/gem5/gem5/pull/762

BobbyRBruce:
Closed in favor of #762

PR #672, opened 2023-12-11
arch-riscv: squash walks with tlb hits in startWalkWrapper

saul44203:
Because each vector load is fragmented into 64 byte cache-aligned chunks, and one page-table walk is issued per fragment on tlb miss, walks start to accumulate on a pending queue, which is processed in a blocking way (no pending walks can be issued while one is being processed). This adds noticeable latency on vector loads when VLEN is sufficiently large.

This commit fixes the issue by allowing walks to be squashed if a TLB lookup hits just before starting the walk on `startWalkWrapper`. This idea was taken from the ARM walker.

### DIFF
--- a/src/arch/riscv/pagetable_walker.cc
+++ b/src/arch/riscv/pagetable_walker.cc
@@ -200,8 +200,19 @@ Walker::startWalkWrapper()
 {
     unsigned num_squashed = 0;
     WalkerState *currState = currStates.front();
+
+    // check if we get a tlb hit to skip the walk
+    Addr vaddr = Addr(sext<VADDR_BITS>(currState->req->getVaddr()));
+    TlbEntry *e = tlb->lookup(vaddr, currState->satp.asid, currState->mode,
+                              true);
+    Fault fault = NoFault;
+    if (e) {
+       fault = tlb->checkPermissions(currState->status, currState->pmode,
+                                     vaddr, currState->mode, e->pte);
+    }
+
     while ((num_squashed < numSquashable) && currState &&
-        currState->translation->squashed()) {
+           (currState->translation->squashed() || (e && fault == NoFault))) {
         currStates.pop_front();
         num_squashed++;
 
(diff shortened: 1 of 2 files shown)


src/arch/riscv/pagetable_walker.cc
```diff
@@ -200,18 +200,32 @@ Walker::startWalkWrapper()
...
-    while ((num_squashed < numSquashable) && currState &&
-        currState->translation->squashed()) {
+
+    // check if we get a tlb hit to skip the walk
+    Addr vaddr = Addr(sext<VADDR_BITS>(currState->req->getVaddr()));
+    TlbEntry *e = tlb->lookup(vaddr, currState->satp.asid, currState->mode, true);
+    Fault fault = NoFault;
+    if (e)
+       fault = tlb->checkPermissions(currState->status, currState->pmode, vaddr,
+                                     currState->mode, e->pte);
```
powerjg: Please use `{ }` around multi-line if statements. We've had too many bugs crop up because of this :)

```suggestion
    if (e) {
       fault = tlb->checkPermissions(currState->status, currState->pmode, vaddr,
                                     currState->mode, e->pte);
    }
```

powerjg (commented):
One small thing below. Otherwise, this looks good to me.

BobbyRBruce:
The current test failing in our CI is due to the pull-request not containing a change ID. The easiest way to solve this is with the following:

n f=.git/hooks/commit-msg ; mkdir -p  ;  curl -Lo  https://gerrit-review.googlesource.com/tools/hooks/commit-msg ; chmod +x \n
git commit --amend --no-edit
This will add a change ID to the commit at the HEAD of the pull request.

To avoid this going forward we strongly recommend using our pre-commit git hooks. To install execute util/pre-commit-install.sh in the gem5 repo. These git hooks automatically add Change-IDs to a commit and run our style checks on code you have added or altered.

saul44203:
Thanks. Should be good now. Tell me otherwise if something is missing or has to be changed.

BobbyRBruce (approved):

BobbyRBruce:
> Thanks. Should be good now. Tell me otherwise if something is missing or has to be changed.

THanks for the contribution. This appears to be working fine. I'll merge it.

PR #673, opened 2023-12-12
arch-riscv: Move trap handler status to RiscvISA

rogerchang23424:
The change will enable customize the xpie, xie and xpp bits, privilege mode configuration and delegation. Some of CPUs handle xpie, xie and xpp bits in the other CSR. For example, if the CPU support CLIC mode, the xpie and xpp will handle in the xcause rather than xstatus.

Reference: https://github.com/riscv/riscv-fast-interrupt
Change-Id: I03159a4b5093dc0aa1b36df8729e9610b1d3bd7f

### DIFF
--- a/src/arch/riscv/faults.cc
+++ b/src/arch/riscv/faults.cc
@@ -65,40 +65,10 @@ RiscvFault::invoke(ThreadContext *tc, const StaticInstPtr &inst)
              name(), pc_state);
 
     if (FullSystem) {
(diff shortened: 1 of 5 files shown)


Harshil2107:
@hnpl can you also take a look at this?

hnpl (approved):
Thanks Roger! I think it makes a lot of sense to move the trap handler to the ISA object. There's a small nit-pick but overall, this looks good to me.

rogerchang23424:
The PR is not required changed anymore, close the bug.

PR #675, opened 2023-12-12
stdlib,resources: Fix obtaining gem5 Looppoint resources

BobbyRBruce:
There were two small bugs preventing gem5 from obtaining Looppoint resources.

1. When obtained via a `WorkloadResource` there was an assert which assumed the values in the resource's DB entry's `additional_parameter` field were of type string. This is not the case. For Looppoint resources there are additional parameters which are arrays.
2. Due to changes introduced in https://github.com/gem5/gem5/pull/625, the Looppoint CSV and JSON files were not being downloaded when needed. This was fixed by replacing access to the `local_path` variable with a call to `get_local_path()`.


powerjg (approved):
That was an easy fix!

PR #678, opened 2023-12-13
configs: Add options for RISC-V ISA configuration

rogerchang23424:
The change will make users configure the RVType, VLEN, ELEN and PrivilegeMode combination more easily.

Change-Id: I7687f3a2a44e0bcff4588768be2954adad8dceb0

### DIFF
--- a/configs/common/Options.py
+++ b/configs/common/Options.py
@@ -689,6 +689,31 @@ def addCommonOptions(parser, default_isa: Optional[ISA] = None):
         help="Override vendor string returned by CPUID instruction in X86.",
     )
 
+    # RISC-V ISA options
+    parser.add_argument(
+        "--riscv-type",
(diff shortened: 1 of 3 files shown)


powerjg (commented):
I won't block this patch, but I want to let you know that it's my medium-term goal to delete `common/Options.py`. This file is totally out of hand and basically impossible to maintain.

I am a strong proponent of using *python scripts* as the interface to gem5 *not* exposing every possible option as command line parameters.

rogerchang23424:
The PR is make user to configure the options of RiscvISA `riscv_type`, `vlen` and `elen` easier. I have no idea how to configure the RiscvISA options in the stdlib.`RiscvProcessor` or `RiscvBoard`. I would like to collect the ideas from community, Thanks.

PR #679, opened 2023-12-13
tests: Silence Clang 16 warnings

arichardson:
I was trying to build with clang 16 and ran into these -Werror warnings

Change-Id: I9207990fcfe9c1a5485945294969f21d1d812a7c

### DIFF
--- a/src/base/compiler.hh
+++ b/src/base/compiler.hh
@@ -90,7 +90,7 @@
 // This version is for macros which are statement-like, which frequently use
 // "do {} while (0)" to make their syntax look more like normal c++ statements.
 #  define GEM5_DEPRECATED_MACRO_STMT(name, definition, message) \
-     do {{definition;} GEM5_DEPRECATED_MACRO(name, {}, message);} while (0)
+     do {{definition;} GEM5_DEPRECATED_MACRO(name, ({}), message);} while (0)
 
 // To mark a class as deprecated in favor of a new name, add a respective
 // instance of this macro to the file that used to declare the old name.
--- a/src/base/bitunion.test.cc
+++ b/src/base/bitunion.test.cc
@@ -129,7 +129,7 @@ containingFunc(uint64_t init_val, uint64_t fieldVal)
 // Declare these as global so g++ doesn't ignore them. Initialize them in
 // various ways.
 EmptySixtyFour emptySixtyFour = 0;
-EmptyThirtyTwo emptyThirtyTwo{};
+[[maybe_unused]] EmptyThirtyTwo emptyThirtyTwo{};
 [[maybe_unused]] EmptySixteen emptySixteen;
 EmptyEight emptyEight(0);
 


SConstruct
```diff
@@ -670,6 +670,8 @@ for variant_path in variant_paths:
         with gem5_scons.Configure(env) as conf:
             conf.CheckCxxFlag('-Wno-c99-designator')
             conf.CheckCxxFlag('-Wno-defaulted-function-deleted')
+            # Silence warning for DPRINTF_UNCONDITIONAL
+            conf.CheckCxxFlag('-Wno-compound-token-split-by-macro')
```
powerjg: I would prefer make (minor) changes to `DPRINTF_UNCONDITIONAL` instead of quieting a warning. How hard do you think that would be? What's the warning?

arichardson: That would definitely be better, let me check:

```
src/base/trace.test.cc:645:5: error: '}' and ')' tokens terminating statement expression appear in different macro expansion contexts [-Werror,-Wcompound-token-split-by-macro]
    DPRINTF_UNCONDITIONAL(TraceTestDebugFlag, "Test message");
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/base/trace.hh:253:5: note: expanded from macro 'DPRINTF_UNCONDITIONAL'
    GEM5_DEPRECATED_MACRO_STMT(DPRINTF_UNCONDITIONAL,      \
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/base/compiler.hh:93:53: note: expanded from macro 'GEM5_DEPRECATED_MACRO_STMT'
     do {{definition;} GEM5_DEPRECATED_MACRO(name, {}, message);} while (0)
                                                    ^
src/base/compiler.hh:89:66: note: expanded from macro 'GEM5_DEPRECATED_MACRO'
     ([](){[[deprecated(message)]] int name{}; return name;}(), (definition))
                                                                 ^~~~~~~~~~
src/base/trace.test.cc:645:5: note: ')' token is here
    DPRINTF_UNCONDITIONAL(TraceTestDebugFlag, "Test message");
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/base/trace.hh:253:5: note: expanded from macro 'DPRINTF_UNCONDITIONAL'
    GEM5_DEPRECATED_MACRO_STMT(DPRINTF_UNCONDITIONAL,      \
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/base/compiler.hh:93:24: note: expanded from macro 'GEM5_DEPRECATED_MACRO_STMT'
     do {{definition;} GEM5_DEPRECATED_MACRO(name, {}, message);} while (0)
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/base/compiler.hh:89:76: note: expanded from macro 'GEM5_DEPRECATED_MACRO'
     ([](){[[deprecated(message)]] int name{}; return name;}(), (definition))
```

arichardson: Turns out it was a lot easier to fix than expected so I've done that now.

powerjg: Awesome, thanks!

src/base/compiler.hh
```diff
@@ -90,7 +90,7 @@
 // This version is for macros which are statement-like, which frequently use
 // "do {} while (0)" to make their syntax look more like normal c++ statements.
 #  define GEM5_DEPRECATED_MACRO_STMT(name, definition, message) \
-     do {{definition;} GEM5_DEPRECATED_MACRO(name, {}, message);} while (0)
```
arichardson, line 93: Looks like GCC is not happy with this change, will try to fix (hopefully `(void)0` works).

arichardson: Fixed now.

powerjg (approved):

powerjg:
@Harshil2107 we might want to run the compiler tests on this change to make sure that it doesn't break any of our supported compilers. Assuming we can now compile with clang-16, we may want to add that compiler to our tests as well.

BobbyRBruce:
It's strange these haven't been seen in our compiler tests. We do test with clang 16 (perhaps a slightly older version of clang 16?)

BobbyRBruce (approved):

BobbyRBruce:
I'm gonna merge this. I'm planning to rerun the compiler tests soon (once I fix the current compilation failure)

PR #681, opened 2023-12-13
Add support for RISC-V semihosting

arichardson:
See https://github.com/riscv-software-src/riscv-semihosting for the current specification. Almost all code is shared with the Arm implementation.

Tested by running some binaries built with [picolibc](https://github.com/picolibc/picolibc).


powerjg:
Wow! This is a big contribution!

We're (the maintainers) currently focused on the v23.1 release, but after the holidays we'll definitely help out with getting this merged! You may want to join the developer meeting tomorrow (#661), if that's possible for you, to discuss this and other RISC-V improvements.

src/arch/arm/ArmSemihosting.py
```diff
@@ -33,13 +33,14 @@
...
 
+from m5.objects.BaseSemihosting import BaseSemihosting
 from m5.objects.Serial import SerialDevice
 from m5.objects.Terminal import Terminal
 from m5.params import *
 from m5.SimObject import *
 
 
-class ArmSemihosting(SimObject):
+class ArmSemihosting(BaseSemihosting):
```
rogerchang23424, line 39: Please remove the parameters that are specified in the `BaseSemihosting`

arichardson: Good catch, will update

rogerchang23424:
Another

rogerchang23424:
The debug flag `Semihosting` is defined in the `src/arch/arm`. It need to move to `src/arch/generic`.

https://github.com/gem5/gem5/blob/c66862f6e3fbad5b0066c847aea8c5aef62b367e/src/arch/arm/SConscript#L135

andysan:
@giactra: Could you have a look at this? It looks like quite a lot of Arm specific stuff has ended up in the generic file.

arichardson:
> @giactra: Could you have a look at this? It looks like quite a lot of Arm specific stuff has ended up in the generic file.

Could you give me an example? RISC-V semihosting reuses the same semihosting ops as ARM (https://github.com/ARM-software/abi-aa/blob/2023Q3/semihosting/semihosting.rst#semihosting-operations) so I believe all of this needs to be in the generic file?

src/arch/generic/SConscript
```diff
@@ -40,6 +40,10 @@ Import('*')
 
 Source('htm.cc')
 Source('mmu.cc')
+if env['CONF']['USE_ARM_ISA'] or env['CONF']['USE_RISCV_ISA']:
+     Source('semihosting.cc')
+     SimObject('BaseSemihosting.py', sim_objects=['BaseSemihosting'])
+     DebugFlag('Semihosting', tags='arm isa')
```
rogerchang23424: The tags `arm isa` required to be removed, the flag will not build if the `USE_ARM_ISA` is not set.

arichardson: Oh yes stupid copy paste error. Will fix tomorrow

configs/example/riscv/fs_linux.py
```diff
@@ -168,11 +178,17 @@ def generateDtb(system):
 system.mem_mode = mem_mode
 system.mem_ranges = [AddrRange(start=0x80000000, size=mdesc.mem())]
 
+semihosting = None
```
rogerchang23424: The parameter is `NULL` to make it happy.

src/arch/riscv/semihosting.cc
```diff
@@ -0,0 +1,191 @@
...
+{
+    // Check if the surrounding bytes match the semihosting magic sequence.
+    PortProxy& proxy = portProxyImpl(xc->tcBase());
+    Addr PrevInstAddr = xc->pcState().instAddr() - 4;
+    Addr NextInstAddr = xc->pcState().instAddr() + 4;
+    if ((PrevInstAddr & ~Addr(15)) != (NextInstAddr & ~Addr(15))) {
+        DPRINTF(Semihosting,
+                "Ebreak cannot be a semihosting ebreak since previous "
+                "and next instruction are not sufficiently aligned\n");
+        return false;
```
rogerchang23424: The spec doesn't required the sequence of instruction to be in the same instruction block(16 bytes), right?

andysan:
> > @giactra: Could you have a look at this? It looks like quite a lot of Arm specific stuff has ended up in the generic file.
> 
> Could you give me an example? RISC-V semihosting reuses the same semihosting ops as ARM (https://github.com/ARM-software/abi-aa/blob/2023Q3/semihosting/semihosting.rst#semihosting-operations) so I believe all of this needs to be in the generic file?

The feature bytes and the exit reasons both seem somewhat Arm specific. I would imagine that there will be a point where the RISC-V implementation will need to modify these and add RV-specific features. The fact that the function tables are architecture specific, also implies that the feature structure needs to be architecture specific.

giactra:
> Could you give me an example? RISC-V semihosting reuses the same semihosting ops as ARM (https://github.com/ARM-software/abi-aa/blob/2023Q3/semihosting/semihosting.rst#semihosting-operations) so I believe all of this needs to be in the generic file?

Hey Alex, I tried to have a look at the link you mentioned in the first PR commit [1] and while I see the specification of the intruction trap sequence and register definition, I don't see any reference to the semihosting ops nor I find a line that describes them as being the same as Arm. Could you point me to the relevant documentation?

Thanks!

[1]: https://github.com/riscv-software-src/riscv-semihosting/blob/main/riscv-semihosting-spec.adoc#register

src/arch/generic/semihosting.hh
```diff
@@ -0,0 +1,551 @@
...
+#include "sim/sim_object.hh"
+
+namespace gem5
+{
+
+struct BaseSemihostingParams;
+class SerialDevice;
+
+/**
+ * Semihosting for AArch32, AArch64, RISCV-32 and RISCV-64
```
andysan: To avoid confusion, could you add a link to the normative specification (https://github.com/ARM-software/abi-aa/blob/main/semihosting/semihosting.rst) here? There used to be a link in the R-V specification, but it was dropped for some reason. It's a bit weird that R-V implements some of the quirks we needed for backwards compatibility and pre-Armv8 features, so I think it might be worth clarifying that it is literally implementing the Arm spec.

src/arch/riscv/semihosting.cc
arichardson: Oh indeed that is too strict, they just need to be on the same page (if paging is enabled).

> If paging is in use in the current mode, this sequence must not cross a page boundary as the semihosting system must be able to check for the semihosting sequence without needing data from potentially missing pages.

src/arch/generic/semihosting.hh
arichardson: I believe the intention behind the RISC-V semihosting spec was to be able to reuse as much as possible from existing ARM implementations (both emulators as well as libc support). It's not a final spec, so I believe it's still possible to make some changes to it to remove old quirks (although QEMU has implemented the current version since 6.0 so not sure if there is much that can be changed at this point).

CC: @keith-packard

configs/example/riscv/fs_linux.py
arichardson: I still need to initialize this variable so I can pass it to the constructor further down.

I guess the alternative would be to use something like

```
workload_args = dict()
if args.semihosting:
    workload_args["semihosting"] = RiscvSemihosting(
        files_root_dir=args.semihosting_root,
        cmd_line=args.kernel,
    )
if args.bare_metal:
    system.workload = RiscvBareMetal(**workload_args)
    system.workload.bootloader = args.kernel
else:
    system.workload = RiscvLinux(**workload_args)
    system.workload.object_file = args.kernel
```

but I'm not convinced that is cleaner.

src/arch/generic/semihosting.hh
keith-packard: Yup, the goal was to share the same spec and implementations to avoid fragmentation. I just inherited the bits from the actual authors (mostly Palmer and Megan, I believe?) and got them published so I could upstream the qemu support and use that for picolibc testing.

configs/example/riscv/fs_linux.py
rogerchang23424: > I still need to initialize this variable so I can pass it to the constructor further down.
> 
> I guess the alternative would be to use something like
> 
> ```
> workload_args = dict()
> if args.semihosting:
>     workload_args["semihosting"] = RiscvSemihosting(
>         files_root_dir=args.semihosting_root,
>         cmd_line=args.kernel,
>     )
> if args.bare_metal:
>     system.workload = RiscvBareMetal(**workload_args)
>     system.workload.bootloader = args.kernel
> else:
>     system.workload = RiscvLinux(**workload_args)
>     system.workload.object_file = args.kernel
> ```
> 
> but I'm not convinced that is cleaner.

LGTM, works in my side.

src/arch/riscv/semihosting.cc
rogerchang23424: Ok, I would like to check if the instruction sequence is in the same 4K Bytes page block. It will reduce the strict checking.

arichardson:
Add the latest updates as fixup commits to the end - will rebase to create a sensible history once reviewed.

configs/example/riscv/fs_linux.py
rogerchang23424: > ```
> workload_args = dict()
> if args.semihosting:
>     workload_args["semihosting"] = RiscvSemihosting(
>         files_root_dir=args.semihosting_root,
>         cmd_line=args.kernel,
>     )
> if args.bare_metal:
>     system.workload = RiscvBareMetal(**workload_args)
>     system.workload.bootloader = args.kernel
> else:
>     system.workload = RiscvLinux(**workload_args)
>     system.workload.object_file = args.kernel
> ```

Cloud you apply the code?

rogerchang23424: bump

powerjg:
@andysan and @giactra, do you see any changes that are needed before this is merged?

src/arch/generic/BaseSemihosting.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) 2018, 2019 Riscv Limited
```
giactra: This code is basically copied/extracted from ArmSemihosting.py so it should bear the original copyright holder (Arm Limited), whereas here you replaced Arm with Riscv

giactra (changes requested):
Thanks @powerjg I believe there is still an unaddressed request from Andreas to include a reference to the semihosting documentation https://github.com/gem5/gem5/pull/681/files#r1427809702, also I spotted a copyright issue (see below in review).

Otherwise LGTM, though I am running some tests to make sure nothing is broken... We should probably add some semihosting unit tests to our regressions in the long term

ivanaamit:
Hi @arichardson, as it has been three weeks without updates on this PR, I wanted to check in and see if you are currently working on incorporating the requested changes. Thank you.

arichardson:
> Hi @arichardson, as it has been three weeks without updates on this PR, I wanted to check in and see if you are currently working on incorporating the requested changes. Thank you.

Apologies, I missed these comments in the flood of other GitHub notifications. Will rebase and update shortly

ivanaamit:
Hi @arichardson, another friendly reminder to check if you are working on updating this PR. Thank you.

arichardson:
> Hi @arichardson, another friendly reminder to check if you are working on updating this PR. Thank you.

I am planning to get to this today or tomorrow. Thank you for the reminder!

arichardson:
Sorry for the delay, I've now rebased and added two new fixup commits that should address the outstanding comments. Please let me know if this looks good and I'll squash the commits to a sensible history.

src/arch/generic/BaseSemihosting.py
arichardson: I failed to exclude this line from my search+replace :( Should be fixed now.

rogerchang23424 (approved):
LGTM on RISC-V part, Thanks

giactra:
Thanks @arichardson for updating the header! Though there is still the remaining request from Andreas to be addressed  https://github.com/gem5/gem5/pull/681/files#r1427809702 and then this is probably good to go

arichardson:
Sorry for the delay. Updated the PR to add pointers to the documentation and rebased and edited the history into two commits.

giactra (dismissed):
Thanks for the patience, looks good from my side, waiting for @andysan to provide his feedback as well

src/arch/generic/BaseSemihosting.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2018, 2019 Riscv Limited
```
andysan: This should presumably have been in the previous commit.

src/arch/generic/BaseSemihosting.py
andysan: It seems like this was fixed in the wrong commit.

src/sim/workload.hh
```diff
@@ -103,6 +103,8 @@ class Workload : public SimObject
     virtual Addr getEntry() const = 0;
     virtual ByteOrder byteOrder() const = 0;
     virtual loader::Arch getArch() const = 0;
+    /// Returns the semihosting interface if supported by the current workload.
+    virtual SimObject* getSemihosting() const { return nullptr; }
```
andysan, line 107: I'd suggest not adding this here unless it's used by other architectures.

andysan (dismissed):

ivanaamit:
@arichardson, could you please address Andreas's comments? Thank you.

arichardson:
> @arichardson, could you please address Andreas's comments? Thank you.

Sorry I missed these since I get way too many GitHub notifications. Will try to address ASAP

ivanaamit:
@arichardson, we would appreciate it if you could address the pending comments. Additionally, you should resolve conflicting files for this PR to be merged. Thank you.

ivanaamit:
@arichardson, it would be great if you could simply rebase, and then we can merge this PR. Andreas's comments are not blocking the merge; you can address them if you have time. Thank you.

src/sim/workload.hh
rogerchang23424: We get the semihost from workload `src/arch/riscv/insts/standard.cc`. https://github.com/gem5/gem5/pull/681/commits/e48114c6581d9493030b155717e4d93ec1600957#diff-e0c1e51647bad6c3927cecd92b966ef5f523489faaf39bc4647b1b712a0e39b8

There are three types of workloads used in RISC-V system, `RiscvBareMetal`, `RiscvLinux` and `RiscvBootloaderKernelWorkload`. I think adding the getSimihosting in generic `Workload` is make sense to me

BobbyRBruce:
I've resolved the merge conflicts via a merge commit from develop to this branch. Given @andysan's comments about things being in the wrong commits anyway, I'll submit this as a squash-merge once the tests pass.

arichardson:
Sorry for the delay here and thanks for taking over!

src/arch/generic/semihosting.hh
```diff
@@ -0,0 +1,557 @@
...
+ *
+ * This class implements the Arm semihosting interface. This interface
+ * allows baremetal code access service, such as IO, from the
+ * simulator. It is conceptually a simplified version of gem5's more
+ * general syscall emulation mode.
+ *
+ * Note: While the current (as of Q1 2024) RISC-V semihosting specificion does
+ * not explicitly define the calls that are required to be implemented
+ * (https://github.com/riscv-non-isa/riscv-semihosting), the intention of the
+ * spec is to reuse the Arm interfaces (which is what QEMU also does).
```
arichardson:
```suggestion
 * Note: The RISC-V semihosting specification reuses the Arm semihosting
 * interfaces (https://github.com/riscv-non-isa/riscv-semihosting).
```

arichardson: This has now been fixed in the latest RISC-V semihosting spec.
There is a 0.3 release that says the following:

> The RISC-V semihosting specification adapts the design of publicly available and open source
> semihosting specification [1] to minimize the development effort. Only the mechanism of invoke a
> semihosting service (aka semihosting binary interface) is architecture specific whereas all defined
> semihosting services [1] are architecture independent and resemble system calls of an unix-style
> operating system.

With [1] pointing to the ARM semihosting spec.

src/arch/riscv/RiscvFsWorkload.py
```diff
@@ -43,10 +44,12 @@ class RiscvBareMetal(Workload):
...
     reset_vect = Param.Addr(0x0, "Reset vector")
+    semihosting = Param.RiscvSemihosting(
+        NULL,
+        "Enable support for RISC-V semihosting by settings this parameter",
+    )
     auto_reset_vect = Param.Bool(
-        True,
-        "Use bootloader entry point as reset vector. "
-        "If `auto_reset_vect` is true, the `reset_vect` parameter is ignored.",
+        True, "Use bootloader entry point as reset vector. "
```
rogerchang23424: This part should not be changed in the PR
```suggestion
        True,
        "Use bootloader entry point as reset vector. "
        "If `auto_reset_vect` is true, the `reset_vect` parameter is ignored.",
```

powerjg (approved):

BobbyRBruce (approved):

PR #684, opened 2023-12-14
configs: Fix SMT cpu type checking

giactra:
The args.cpu_type is not a type but a string so the isinstance checking will always fail and an assertion will always be thrown

Change-Id: I6a88d1a514bb323c517949632f4e76be40e87e8c


powerjg (commented):
lgtm. I'll let Bobby merge this so it's in sync between staging and develop

PR #690, opened 2023-12-17
Remove replacement_policy namespace

pranith:
When compared to indexing policies, replacement policies have an extra namespace layer making things a bit more verbose. I tried removing it and I feel it looks more consistent with the rest of the cache code-base. Please let me know if you have any comments. Thanks!

### DIFF
--- a/src/mem/cache/prefetch/associative_set.hh
+++ b/src/mem/cache/prefetch/associative_set.hh
@@ -58,7 +58,7 @@ class AssociativeSet
     /** Pointer to the indexing policy */
     BaseIndexingPolicy* const indexingPolicy;
     /** Pointer to the replacement policy */
-    replacement_policy::Base* const replacementPolicy;
+    BaseReplacementPolicy* const replacementPolicy;
     /** Vector containing the entries of the container */
     std::vector<Entry> entries;
 
@@ -73,7 +73,7 @@ class AssociativeSet
      * @param init_val initial value of the elements of the set
      */
     AssociativeSet(int assoc, int num_entries, BaseIndexingPolicy *idx_policy,
-        replacement_policy::Base *rpl_policy, Entry const &init_val = Entry());
+        BaseReplacementPolicy *rpl_policy, Entry const &init_val = Entry());
 
     /**
      * Find an entry within the set
--- a/src/mem/cache/prefetch/associative_set_impl.hh
+++ b/src/mem/cache/prefetch/associative_set_impl.hh
@@ -37,7 +37,7 @@ namespace gem5
 
 template<class Entry>
 AssociativeSet<Entry>::AssociativeSet(int assoc, int num_entries,
-        BaseIndexingPolicy *idx_policy, replacement_policy::Base *rpl_policy,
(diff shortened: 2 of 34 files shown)


odanrc:
Hello, and thanks for the contribution!

I started introducing namespaces to the (cache) code because gem5 is (also) used by researchers to implement new algorithms, and these algorithms' names sometimes overlap (The Feedback Directed Prefetcher and another recent FDP is a pair of example), so we end up having to add suffixes to the classes. What do the suffixes do? They end up namespacing the class. Therefore, the namespace is necessary, whether it is a C++ namespace, or a "workaround namespace". The problem with the workaround namespaces is that they do not always work well: most replacement policies end up with the "RP" suffix for short, but then the base replacement policy ends up needing to be way more verbose because "BaseRP" would likely clash with another name eventually. The C++ namespace ends up being a constantly functional solution.

Now, one of the problems brought up is that "replacement_policy" is too long. I agree it is, and I wouldn't object to a proposal to renaming it to, for example, "repl_policy". Due to gem5 having a steep learning curve, and the high number of first-time coders using it, I end up favoring verbose/longer names, though (i.e., I prefer keeping replacement_policy, as it is not even needed to be written out that frequently).

Regarding the namespace indexing_policy not existing, I thought I had added it, but I guess the PR got lost when I changed computers, so I ended up never pushing for it. Ditto for the cache namespace. The idea when I started adding namespaces was to eventually add more namespaces in other cache-related code (and hopefully in more places in the future), but the cache code needed some cleanup and I was starting a new job back then, so I ended up not being able to continue. I actually think we should be going on the opposite direction of removing namespaces, and resume adding them!

Finally, if a rename is eventually pursued, there is a deprecation process that must be followed, so that users have 2 versions to convert from the old namespace to the new namespace without breaking their code.

pranith:
Thanks for the explanation Daniel! It makes sense to keep the namespace. I don't mind long names in general, but the 80 character line width limit in gem5 makes it icky to use verbose names.

What do you think about renaming the 'Base' class to 'BaseReplPolicy' or to 'BaseReplacementPolicy'? That way, we can 'use namespace replacement_policy' at the beginning of the file and use the class name within the file without any confusion?

odanrc:
I am against using "using namespace" in general, but it is a C++ feature, so I won't oppose it if there is support from the maintainers for that. Unless it is being added to a header file; then it is definitely a blocker, as it would break the purpose of using namespaces.

Anyway, since your goal of using "using namespace" is just to have a shorter way to declare replacement_policy::Base (i.e., a single class in that namespace), wouldn't it make more sense to simply alias the Base class in the file using it? e.g., "typedef replacement_policy::Base BaseReplPol".

Regarding renaming Base to something (much) longer (BaseReplPolicy or BaseReplacementPolicy), IMHO it would break the purpose of reducing the size. Also, it would entail a 2-version deprecation process anyway. Honestly, given the effort/dedication the deprecation process requires, the inconvenience that these changes will create for users, and the fact that this change would just apply to a handful of instances, I would say it is not worth it.

pranith:
Thanks Daniel! I like the 'typedef ...' idea better. Will use that going forward. Closing this PR.

PR #696, opened 2023-12-19
misc: Turn off 'maybe-uninitialized' warn for regex include

BobbyRBruce:
https://github.com/gem5/gem5/pull/636 triggered a bug with the GCC
compiler and its interaction with the CPP stdlib regex library, outlined here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105562.

This was causing the gem5 Compiler tests to fail for GCC-12: https://github.com/gem5/gem5/actions/runs/7219055796

This fix turns off the 'maybe-unitialized' warning when we include the regex headers in "src/kern/linux/helpers.cc".

### DIFF
--- a/src/kern/linux/helpers.cc
+++ b/src/kern/linux/helpers.cc
@@ -37,7 +37,27 @@
 
 #include "kern/linux/helpers.hh"
 
+/* The following pragmas are to fix a bug in GCC \w the CPP standard library,
+ * outlined here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105562.
+ * This ignores the 'maybe-unitialized' warning when importing regex for
+ * GCC 12.1.
+ */
+
+// ignore 'maybe-uniitialized' warnings for GCC 12.1.
+#if __GNUC__ &&  __GNUC__ == 12 && __GNUC_MINOR__ == 1
+    #if __has_warning( "-Wmaybe-uninitialized" )
+        #define SUPPRESSING_MAYBE_UNINITIALIZED_WARNING
+        // save diagnostic state.
+        #pragma GCC diagnostic push
+        #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+    #endif
+#endif
 #include <regex>
+#ifdef SUPPRESSING_MAYBE_UNINITIALIZED_WARNING
+    // restore the diagnostic state.
+    #pragma GCC diagnostic pop
+#endif
+
 #include <string>
 #include <type_traits>
 #include <vector>


powerjg (dismissed):

src/kern/linux/helpers.cc
```diff
@@ -37,7 +37,22 @@
...
 
+/* The following pragmas are to fix a bug in GCC \w the CPP standard library,
+ * outlined here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105562.
+ * This ignores the 'maybe-unitialized' warning when importing regex.
+ */
+
+// save diagnostic state
+#pragma GCC diagnostic push
+// ignore 'maybe-uniitialized' warnings
+#if !defined(__has_warning) || __has_warning("-Wmaybe-uninitialized")
```
powerjg: You may want to check for GCC version 12.1 here, too. No need to disable this warning for other GCCs.

I *think* `(__GNUC__ == 12 && __GNUC_MINOR__ >= 1)` will work, but tbh chatGPT gave me that code, so YMMV.

powerjg (commented):
How sure are we that push/pop will do nothing if there aren't any new diagnostic pragmas?

src/kern/linux/helpers.cc
```diff
@@ -37,7 +37,22 @@
...
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
+// include regex
 #include <regex>
```
rogerchang23424, line 55: (nit) The compile only failed in gcc 12.1.0 version. The gcc 12.3.0 and 13.2.0 can work without the warning ignore.

BobbyRBruce:
> How sure are we that push/pop will do nothing if there aren't any new diagnostic pragmas?

I'm pretty sure, that my understanding of how this functionality works. It's designed so you can have different warning levels for different sections of code.

src/kern/linux/helpers.cc
BobbyRBruce: Thanks! I've set this to only disable this diagnostic when it's GCC 12.1.0.

BobbyRBruce:
> > How sure are we that push/pop will do nothing if there aren't any new diagnostic pragmas?
> 
> I'm pretty sure, that my understanding of how this functionality works. It's designed so you can have different warning levels for different sections of code.

I've made some changes to this which I think make this hypothetical impossible.

powerjg (dismissed):
Thanks for the updates, Bobby. This is a very annoying bug to get around. We should seriously consider supporting fewer compilers!

powerjg (approved):

PR #698, opened 2023-12-19
configs: Fix SMT cpu type checking

BobbyRBruce:
The args.cpu_type is not a type but a string so the isinstance checking will always fail and an assertion will always be thrown

A cherry-pick of #684 to develop


giactra (approved):

giactra:
Hey Bobby shall I merge this?

PR #701, opened 2023-12-20
mem-ruby: allow comparison of int and Addr in SLICC

arteen1000:
allow easy isolation of specific addresses in coherence protocols. useful for debugging.

Change-Id: I93e07956b8e29837219d328dacfbd5c6067c1a62

### DIFF
--- a/src/mem/slicc/ast/OperatorExprAST.py
+++ b/src/mem/slicc/ast/OperatorExprAST.py
@@ -62,7 +62,15 @@ class InfixOperatorExprAST(ExprAST):
         # Figure out what the input and output types should be
         if self.op in ("==", "!=", ">=", "<=", ">", "<"):
             output = "bool"
-            if ltype != rtype:
+
+            if (
+                str(ltype) == "Addr"
+                and str(rtype) == "int"
+                or str(ltype) == "int"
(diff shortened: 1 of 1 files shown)


src/mem/slicc/ast/OperatorExprAST.py
```diff
@@ -62,7 +62,15 @@ def generate(self, code, **kwargs):
         # Figure out what the input and output types should be
         if self.op in ("==", "!=", ">=", "<=", ">", "<"):
             output = "bool"
-            if ltype != rtype:
+
+            if (
```
mattsinc, line 66: I don't think the formatting of this if matches the formatting elsewhere in the file

mattsinc (changes requested):
Beyond the one minor change, I agree with the spirit of this change.  @powerjg should view too though -- I think it's fine to allow int and Addr in SLICC, but I guess it's worth us discussing what the purpose of allow these return types is from SLICC?

powerjg (approved):
Fine with me :)

powerjg:
@mattsinc can you re-review since you "requested changes" to allow this to merge?

mattsinc (approved):

PR #715, opened 2023-12-27
arch-riscv: Fix issue when vl=0 in VectorIntMaskMacroConstructor

QQeg:
I’ve been working on a fix for the issue #759 where ‘vd’ incorrectly stores all zeros when ‘vl’ is set to 0 in VectorIntMaskMacroConstructor. My solution seems to work, but it behaves differently from other macros when ‘vl’ = 0. Instead of pushing a ‘nop’ to ‘microops’, it pushes a micro operation that remains ineffective due to ‘vl’ being 0. I’m wondering if anyone has thoughts on this? Is there a better way to tackle this problem?
Thanks in advance for your insights!

### DIFF
--- a/src/arch/riscv/isa/templates/vector_arith.isa
+++ b/src/arch/riscv/isa/templates/vector_arith.isa
@@ -1097,11 +1097,7 @@ template<typename ElemType>
     int32_t micro_vl = std::min(tmp_vl, micro_vlmax);
     StaticInstPtr microop;
 
-    if (micro_vl == 0) {
-        microop = new VectorNopMicroInst(_machInst);
-        this->microops.push_back(microop);
-    }
-    for (int i = 0; i < num_microops && micro_vl > 0; ++i) {
+    for (int i = 0; i < num_microops && micro_vl >= 0; ++i) {
         microop = new %(class_name)sMicro<ElemType>(_machInst, micro_vl, i);
         microop->setDelayedCommit();
         this->microops.push_back(microop);


powerjg:
Seems OK to me. 

I wonder if @adriaarmejach, @rogerchang23424, or @hnpl have any opinions.

aarmejach:
I think this should be fine for LMUL >= 1, but I do not know if it works for LMUL<1. Could it be tested?

rogerchang23424:
I'm busy working on my internal project. I will find another day to test the implementation

QQeg:
I’ve conducted tests using [intrinsic_mf8.c](https://github.com/QQeg/gem5/blob/VectorIntMask_issue/reproduce/intrinsic_mf8.c) as my template. For every vmadc instruction with LMUL<1, I set vl to 0, 2, and vlmax. All instructions functioned as expected. For example, here are the outputs (gem5 and spike) of **vmadc_vx_i8mf8_b64**:

vl = 0,
``` REAL SIMULATION 
First four elements of Vs1 = 1 1 1 1 
First four elements of Vd  = 1 1 1 1 
Exiting @ tick 4701500 because exiting with last active thread context

----SPIKE----
bbl loader
First four elements of Vs1 = 1 1 1 1 
First four elements of Vd  = 1 1 1 1 
```

vl = 2,
```  REAL SIMULATION 
First four elements of Vs1 = 1 1 1 1 
First four elements of Vd  = 3 1 1 1 
Exiting @ tick 4701500 because exiting with last active thread context

----SPIKE----
bbl loader
First four elements of Vs1 = 1 1 1 1 
First four elements of Vd  = 3 1 1 1
```

vl = 4 (vlmax)
```  REAL SIMULATION 
First four elements of Vs1 = 1 1 1 1 
First four elements of Vd  = 15 1 1 1 
Exiting @ tick 4727000 because exiting with last active thread context

----SPIKE----
bbl loader
First four elements of Vs1 = 1 1 1 1 
First four elements of Vd  = 15 1 1 1
```

Is it enough, or should we have a more intensive test?

aarmejach (approved):
LGTM

powerjg (approved):

PR #728, opened 2024-01-01
misc: Add a generic cache template

pranith:
This allows us to use a shorter tag than otherwise possible.

Change-Id: Ib339efd97bf33b706f93eb7f0eb5851ceb51afb6

### DIFF
--- a/src/mem/cache/tags/indexing_policies/IndexingPolicies.py
+++ b/src/mem/cache/tags/indexing_policies/IndexingPolicies.py
@@ -44,6 +44,9 @@ class BaseIndexingPolicy(SimObject):
     # Get the associativity
     assoc = Param.Int(Parent.assoc, "associativity")
 
+    # number of bits in the tag
+    tag_bits = Param.Unsigned(64, "number of bits in the tag")
+
 
 class SetAssociative(BaseIndexingPolicy):
     type = "SetAssociative"
--- a/src/mem/cache/tags/indexing_policies/base.hh
+++ b/src/mem/cache/tags/indexing_policies/base.hh
@@ -96,6 +96,11 @@ class BaseIndexingPolicy : public SimObject
      */
     const int tagShift;
 
+    /**
+     * The tag mask.
+     */
+    const uint64_t tagMask;
+
   public:
     /**
      * Convenience typedef.
--- a/src/base/associative_cache.hh
+++ b/src/base/associative_cache.hh
@@ -1,0 +1,347 @@
+/*
+ * Copyright (c) 2024 Pranith Kumar
+ * All rights reserved
+ *
+ * The license below extends only to copyright in the software and shall
+ * not be construed as granting a license to any other intellectual
+ * property including but not limited to intellectual property relating
+ * to a hardware implementation of the functionality of the software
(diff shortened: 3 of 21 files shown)


src/mem/cache/tags/indexing_policies/base.hh
```diff
@@ -96,6 +96,16 @@ class BaseIndexingPolicy : public SimObject
      */
     const int tagShift;
 
+    /**
+     * The number of tag bits per entry
+     */
+    const int tagBits;
+
+    /**
+     * The tag mask.
```
powerjg, line 100: Can you add more information here about how this is used?

src/mem/cache/tags/indexing_policies/IndexingPolicies.py
```diff
@@ -44,6 +44,9 @@ class BaseIndexingPolicy(SimObject):
     # Get the associativity
     assoc = Param.Int(Parent.assoc, "associativity")
 
+    # number of bits in the tag
+    tag_bits = Param.Unsigned(64, "number of bits in the tag")
```
powerjg, line 48: Can you improve the help string here and/or improve the name of the variable? There are a few confusions possible here.
- The number of bits in the tag would really be 64 - index bits - offset bits, so 64 bits for the tag is misleading.
- What would this parameter be used for? Why would you set the number of tag bits to not be *all* of the remaining bits? Adding this explanation to the help string will enable others to understand and use this feature.

powerjg (commented):
The code seems OK, but before we merge this I would like to understand the use case. When would you set the tag to *not* be the entire address for an indexing policy?

If you could update the commit message with (1) the problem that you are solving and (2) a detailed explanation of why this is the solution to the problem, it will make it easier to review the code.

src/mem/cache/tags/indexing_policies/base.hh
```diff
@@ -96,6 +96,16 @@ class BaseIndexingPolicy : public SimObject
      */
     const int tagShift;
 
+    /**
+     * The number of tag bits per entry
+     */
```
odanrc: Nit: This variable is not really needed

pranith: Removed the unused variable.

src/mem/cache/tags/indexing_policies/IndexingPolicies.py
pranith: I used 64 so that I don't change the current semantics of the code. I am open to suggestions on what to use here, so please let me know.

There are cases where the tag is shortened either by hashing or dropping the upper bits that do not change often to save storage. BTB is one example which uses the reduced tags. I added a patch showing how the reduced tag is used in BTB in the series.

pranith:
> The code seems OK, but before we merge this I would like to understand the use case. When would you set the tag to _not_ be the entire address for an indexing policy?
> 
> If you could update the commit message with (1) the problem that you are solving and (2) a detailed explanation of why this is the solution to the problem, it will make it easier to review the code.

Thanks for the review. I updated the commit message and added a few other patches where I use this tag mask (the BTB patch). Thanks!

pranith:
I'll send a new PR.

PR #729, opened 2024-01-02
mem-cache: Add default values for args in tagged entry

pranith:
Add a default function arg so that we can easily invoke these functions from code that does not care about the 'is_secure' field for its entries.

Change-Id: Ib0fb985bffd52c3b2187e2f6eb90135871fd8ff5

### DIFF
--- a/src/mem/cache/tags/tagged_entry.hh
+++ b/src/mem/cache/tags/tagged_entry.hh
@@ -78,7 +78,7 @@ class TaggedEntry : public ReplaceableEntry
      * @return True if the tag information match this entry's.
      */
     virtual bool
-    matchTag(Addr tag, bool is_secure) const
+    matchTag(Addr tag, bool is_secure = false) const
     {
         return isValid() && (getTag() == tag) && (isSecure() == is_secure);
     }
@@ -90,7 +90,7 @@ class TaggedEntry : public ReplaceableEntry
      * @param tag The tag value.
      */
     virtual void
-    insert(const Addr tag, const bool is_secure)
+    insert(const Addr tag, const bool is_secure = false)
     {
         setValid();
         setTag(tag);


powerjg (commented):
Why is this change needed? Simply saving 7 characters from a call site is probably not a good enough reason. Is this changing solving a specific problem for you?

Since this does not change any code from any of the call sites in upstream gem5, I am hesitant to merge it unless it's solving some specific problem for you.

Improving the commit message do describe the problem this solves would make it more likely to be merged.

odanrc:
I agree with Jason. One major problem is that this address-secure bit pair is somewhat the standard API of the cache/memory systems. Changing the signature here without changing it everywhere else would break this standard. Now, this is not a request to start defaulting this arg everywhere because this could lead to misuse or bugs being added (it'd be a major change that brings little benefit).

I understand, though, that indexing policies can (and are) used for internal caches that don't need the secure address bits, even if the ISA as a whole does need these bits. So it would probably make sense to change the API to make it non-mandatory, but probably not by simply defaulting to false, as it would be a remedy, not a fix.

Some context on the topic: the secure argument was added because it is needed by some ISAs. Other ISAs could need to add other "specialization to their space addressing", so I think possible solutions would be to make the Addr type a base interface or a struct declared per ISA. These changes, however, would require a huge effort, and one of them has already been opposed in the past (e.g., penalizing ISAs without addressing specializations with virtualization is bad). This is the kind of discussion that would probably require discussion in the mailing list so that the interested parties could chime in.

pranith:
The motivation for this change was to use the tagged entry in more generic structures that are not memory system specific. As @odanrc suggested, it makes sense to create a generic base class and then use that for the tagged entry. That work is on-going and I will post those patches soon. Closing this PR.

PR #730, opened 2024-01-03
gpu-compute: Added register file cache support

KaiBatley:
The RFC is defaulted to a size of 0 which removes it completely. To use the RFC set the --register-file-cache-size to a non-zero multiple of two. In addition, rfc_pipe_length may be altered to increase or decrease RFC latency benefit.


mattsinc:
@KaiBatley you will need to add a ChangeID to this too

src/gpu-compute/GPU.py
```diff
@@ -96,6 +96,14 @@ class VectorRegisterFile(RegisterFile):
     cxx_header = "gpu-compute/vector_register_file.hh"
 
 
+class RegisterFileCache(SimObject):
+    type = "RegisterFileCache"
+    cxx_class = "gem5::RegisterFileCache"
+    cxx_header = "gpu-compute/register_file_cache.hh"
+    simd_id = Param.Int(-1, "SIMD ID associated with this Register File Cache")
```
mattsinc: Just curious why -1 for the default here?

src/gpu-compute/GPU.py
```diff
@@ -96,6 +96,14 @@ class VectorRegisterFile(RegisterFile):
...
+    cache_size = Param.Int(0, "capacity of rfc")
```
mattsinc: capacity means size (number of entries), right?  If so maybe we should update description.

src/gpu-compute/GPU.py
```diff
@@ -150,6 +158,11 @@ class ComputeUnit(ClockedObject):
     dpbypass_pipe_length = Param.Int(
         4, "vector ALU Double Precision bypass latency"
     )
+
+    rfc_pipe_length = Param.Int(
+        2, "vector ALU Double Precision bypass latency"
```
mattsinc: I'm not sure the description is right here -- seems a copy-paste error from above.

src/gpu-compute/register_file.cc
```diff
@@ -194,6 +194,10 @@ RegisterFile::RegisterFileStats::RegisterFileStats(statistics::Group *parent)
     : statistics::Group(parent),
       ADD_STAT(registerReads,
               "Total number of DWORDs read from register file"),
+      ADD_STAT(cache_read_hits,
```
mattsinc: To avoid confusion, I recommend calling these "rfc_..." instead, otherwise folks will likely think cache refers to the L1/L2/etc.

src/gpu-compute/register_file_cache.cc
```diff
@@ -0,0 +1,172 @@
+/*
```
mattsinc, line 1: Copyright should be different here since we wrote it, not AMD.  See this file for an example: https://github.com/gem5/gem5/blob/develop/src/mem/ruby/structures/ALUFreeListArray.cc#L2

src/gpu-compute/register_file_cache.hh
```diff
@@ -0,0 +1,115 @@
+/*
+ * Copyright (c) 2015-2017 Advanced Micro Devices, Inc.
```
mattsinc: Same here

src/gpu-compute/register_file_cache.cc
```diff
@@ -0,0 +1,172 @@
...
+void
+RegisterFileCache::setParent(ComputeUnit *_computeUnit)
+{
+    computeUnit = _computeUnit;
+}
+
+bool
+RegisterFileCache::inRFC(int regIdx)
+{
+    return (lruHash.find(regIdx) != lruHash.end());
```
mattsinc, line 66: How does this work if the RFC is disabled?  I guess it just returns false since end is always what the search finds?

mattsinc (changes requested):

src/gpu-compute/register_file_cache.cc
```diff
@@ -0,0 +1,172 @@
+/*
+ * Copyright (c) 2015-2017 Advanced Micro Devices, Inc.
```
abmerop: 2023 / 2024 and probably UW Madison?

configs/example/apu_se.py
```diff
@@ -533,10 +540,16 @@
                 simd_id=j, wf_size=args.wf_size, num_regs=args.sreg_file_size
             )
         )
+        rfcs.append(
+            RegisterFileCache(
+                simd_id=j, cache_size=args.register_file_cache_size
+            )
+        )
```
abmerop, line 547: Should this be added to the FS mode config too? (configs/example/gpufs/system/amdgpu.py I think).

abmerop (commented):
Looks good, just some minor comments

src/gpu-compute/register_file_cache.cc
mattsinc: Right -- I mentioned it in the other file but forgot to put "here too" here, thanks.

configs/example/apu_se.py
mattsinc: Good point.  @KaiBatley can you please do this?

src/gpu-compute/GPU.py
KaiBatley: I just copied from the RegisterFile above which sets it to -1.

src/gpu-compute/register_file_cache.cc
KaiBatley: Yes, you are correct. The lruHash will always be empty (find will return lruHash.end) and the function will return false.

src/gpu-compute/GPU.py
powerjg: IMO, either you should have *no* default, which means the user must set the value, or you should have a sensible default.

If you need to have an id that's unique then you should do this in `startup` (e.g., with a `register` function that calls to the parent object) not expect the user to set the correct ids.

abmerop (dismissed):

src/gpu-compute/GPU.py
mattsinc: I suppose this is a point of debate then -- if @KaiBatley is simply using the same approach that already exists for the GPU, asking him to change it is tricky because then the code itself wouldn't be consistent across the pieces.  But at the same time I can understand not wanting to perpetuate a design one thinks is problematic.

@KaiBatley do you feel like you know what needs to be done to address @powerjg 's request?

powerjg: > simply using the same approach that already exists for the GPU, asking him to change it is tricky because then the code itself wouldn't be consistent across the pieces. But at the same time I can understand not wanting to perpetuate a design one thinks is problematic.

I agree! 

There are two different points/issues here:
1. Default argument `-1` vs. no default. If it is *required* that the user specify something other than `-1`, then I would advocate for making it have no default so there is an error if the user doesn't set the value. If, on the other hand, `-1` is meaningful, then keeping that the default is OK.
2. Counting "ids" should be done in C++ (or python) not *by the user*. This is a broader design problem that is pervasive across not just the GPU but also Ruby. This is something that should probably be addressed in a separate commit, and it not particularly important.

My comment was more about (1) than (2). (2) was more for @KaiBatley (and others') benefit in the future.

KaiBatley: I have removed the default of -1 for the simd_id as it has no meaningful value.

powerjg: Thanks!

powerjg (approved):
Adding back approval since Matt P. approved and the only change is the default value of simd_id

mattsinc (approved):
Thanks @powerjg !  One small request: can you confirm that the license added here (register_file_cache.[cc. hh]) is the "correct" one and doesn't have the problematic clause that caused us issues previously?  It looks correct to me, but I always worry about this now.

powerjg:
> Thanks @powerjg ! One small request: can you confirm that the license added here (register_file_cache.[cc. hh]) is the "correct" one and doesn't have the problematic clause that caused us issues previously? It looks correct to me, but I always worry about this now.

Looks fine to me. For the record, https://github.com/gem5/gem5/blob/stable/LICENSE is the canonical license for gem5.

PR #731, opened 2024-01-03
gpu-compute: WAX dependency detection

KaiBatley:
WAX Dependencies would be missed if a RAW Dependency also existed.

### DIFF
--- a/src/gpu-compute/vector_register_file.cc
+++ b/src/gpu-compute/vector_register_file.cc
@@ -58,14 +58,19 @@ VectorRegisterFile::VectorRegisterFile(const VectorRegisterFileParams &p)
 bool
 VectorRegisterFile::operandsReady(Wavefront *w, GPUDynInstPtr ii) const
 {
+    bool src_ready = true, dst_ready=true;
     for (const auto& srcVecOp : ii->srcVecRegOperands()) {
         for (const auto& physIdx : srcVecOp.physIndices()) {
             if (regBusy(physIdx)) {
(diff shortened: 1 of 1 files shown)


ivanaamit:
Hi @KaiBatley,

```
f=.git/hooks/commit-msg
mkdir -p $f
chmod +x $f/commit-msg
```

mattsinc:
@KaiBatley let me know if you need help with the hooks.

mattsinc (commented):
This looks good except for the ChangeID

abmerop (dismissed):
Looks good to me.  I assume the intention is to print *all* of the dependencies since it will only break the inner loop?

mattsinc:
> Looks good to me. I assume the intention is to print _all_ of the dependencies since it will only break the inner loop?

Good point.  That seems reasonable to me too, but we could break out of both loops after finding the first RAW/WAX -- that way we would ensure that at least one of them are found, but not overwhelm the counters with many RAWs/WAXs being found in the same cycle.  Maybe that would be better?

KaiBatley:
> > Looks good to me. I assume the intention is to print _all_ of the dependencies since it will only break the inner loop?
> 
> Good point. That seems reasonable to me too, but we could break out of both loops after finding the first RAW/WAX -- that way we would ensure that at least one of them are found, but not overwhelm the counters with many RAWs/WAXs being found in the same cycle. Maybe that would be better?

I've modified it to work this way.

abmerop (approved):

mattsinc (approved):
I debated if it was cleaner to add a second break or put the !ready checks above in the next loop iteration's checks.  I guess this is cleaner, thanks @KaiBatley !

PR #733, opened 2024-01-03
gpu-compute: Support for MI200 GPU model

abmerop:


configs/example/gpufs/mi200.py
```diff
@@ -0,0 +1,159 @@
...
+def addDemoOptions(parser):
+    parser.add_argument(
+        "-a", "--app", default=None, help="GPU application to run"
+    )
+    parser.add_argument(
+        "-o", "--opts", default="", help="GPU application arguments"
+    )
+
+
+def runMI100GPUFS(cpu_type):
```
mattsinc: MI200 instead of MI100?

configs/example/gpufs/mi200.py
```diff
@@ -0,0 +1,159 @@
...
+    args.script = tempRunscript
+    args.dgpu_xor_low_bit = 0
+    args.gpu_device = "MI200"
+
+    # Run gem5
+    runfs.runGpuFSSystem(args)
+
+
+if __name__ == "__m5_main__":
+    runMI100GPUFS("X86KvmCPU")
```
mattsinc: MI200 instead of MI100 here too?

src/gpu-compute/wavefront.cc
```diff
@@ -341,48 +340,6 @@ Wavefront::initRegState(HSAQueueEntry *task, int wgSizeInWorkItems)
                         wfSlotId, wfDynId, physSgprIdx,
                         task->privMemPerItem());
                 break;
-              case GridWorkgroupCountX:
```
mattsinc, line 344: Are these removed in gfx90a?  Else why are you removing them?

mattsinc (changes requested):
Few small things, thanks for doing this!

configs/example/gpufs/mi200.py
abmerop: Whoops, will change these two

src/gpu-compute/wavefront.cc
abmerop: These might have been removed in GFX7 (?).  The bits have been repurposed it seems.  See bits 457:455 here: https://llvm.org/docs/AMDGPUUsage.html#code-object-v3-kernel-descriptor .

These must now always be zero, therefore the code will never be used.  I didn't want to leave the code there as it could cause the SGPR offsets to be incorrect

mattsinc (approved):
I only have one question remaining, maybe naive: where in this is the code that sets up the device?  E.g., number of CUs, SQC, etc.?  I can't seem to find it.

I see the end of mi200.py sets memory size and a couple other things, but the rest seems to be completely absent unless I'm missing something ...

abmerop:
> I only have one question remaining, maybe naive: where in this is the code that sets up the device? E.g., number of CUs, SQC, etc.? I can't seem to find it.
> 
> I see the end of mi200.py sets memory size and a couple other things, but the rest seems to be completely absent unless I'm missing something ...

The topology is a separate config file and is mostly independent of the device type

mattsinc:
> > I only have one question remaining, maybe naive: where in this is the code that sets up the device? E.g., number of CUs, SQC, etc.? I can't seem to find it.
> > I see the end of mi200.py sets memory size and a couple other things, but the rest seems to be completely absent unless I'm missing something ...
> 
> The topology is a separate config file and is mostly independent of the device type

Right but doesn't it have to be called somewhere from mi200.py?

abmerop:
> > > I only have one question remaining, maybe naive: where in this is the code that sets up the device? E.g., number of CUs, SQC, etc.? I can't seem to find it.
> > > I see the end of mi200.py sets memory size and a couple other things, but the rest seems to be completely absent unless I'm missing something ...
> > 
> > 
> > The topology is a separate config file and is mostly independent of the device type
> 
> Right but doesn't it have to be called somewhere from mi200.py?

It is currently a command line option (--gpu-topology I think) and corresponding command line stuff to set the number of CUs... I was planning to work on some topology stuff soon and probably will revisit the VIPER board, so soon the configuration will be somewhat easier

mattsinc:
> > > > I only have one question remaining, maybe naive: where in this is the code that sets up the device? E.g., number of CUs, SQC, etc.? I can't seem to find it.
> > > > I see the end of mi200.py sets memory size and a couple other things, but the rest seems to be completely absent unless I'm missing something ...
> > > 
> > > 
> > > The topology is a separate config file and is mostly independent of the device type
> > 
> > 
> > Right but doesn't it have to be called somewhere from mi200.py?
> 
> It is currently a command line option (--gpu-topology I think) and corresponding command line stuff to set the number of CUs... I was planning to work on some topology stuff soon and probably will revisit the VIPER board, so soon the configuration will be somewhat easier

Ok, I think I understand.

abmerop:
@andysan @giactra is this commit in particular ok with ARM folks?  https://github.com/gem5/gem5/pull/733/commits/4903fe2db1d9052a44c078730a8217361e879a16

abmerop:
Buried in this are some ARM changes.  I want to make sure @andysan and @giactra are okay with pulling out ARM's fplib from 2013, and if possible let me know if I am using it correctly as I am basically ignoring the FPSCR register (weird github glitch?)

giactra:
> Buried in this are some ARM changes. I want to make sure @andysan and @giactra are okay with pulling out ARM's fplib from 2013, and if possible let me know if I am using it correctly as I am basically ignoring the FPSCR register (weird github glitch?)

About the FPSCR, it does mainly two things

1) Encodes rounding mode and denormal flushing switches -> most of the times it is not really important for your simulation to get accurate FP results  (does 2.34 instead of 2.35 as a result constitute a problem?)

2) Carries floating point flags set during the operation -> This could be more important if the hardware exposes floating point exceptions to the software (an action might be taken if we have a NaN result, or an overflow/underflow has happened).

IMHO you need to figure out if point 2 is relevant to your use case. If not, you can probably ignore it.

Giacomo

abmerop:
> > Buried in this are some ARM changes. I want to make sure @andysan and @giactra are okay with pulling out ARM's fplib from 2013, and if possible let me know if I am using it correctly as I am basically ignoring the FPSCR register (weird github glitch?)
> 
> About the FPSCR, it does mainly two things
> 
> 1. Encodes rounding mode and denormal flushing switches -> most of the times it is not really important for your simulation to get accurate FP results  (does 2.34 instead of 2.35 as a result constitute a problem?)
> 2. Carries floating point flags set during the operation -> This could be more important if the hardware exposes floating point exceptions to the software (an action might be taken if we have a NaN result, or an overflow/underflow has happened).
> 
> IMHO you need to figure out if point 2 is relevant to your use case. If not, you can probably ignore it.
> 
> Giacomo

Thanks Giacomo.  When I looked at it is seems the default(?) rounding mode / denormal flushing are the same as what hardware is doing so it seems safe to ignore.  For the carry and other flags, I'm currently doing only one operation at a time so I think that should be OK.  Similar with exceptions it seems the default is to not raise any. So I think this is good as-is.

PR #735, opened 2024-01-04
fastmodel: Fix the Fastmodel RemoteGDB initial

rogerchang23424:
Change-Id: Iec9ef145ccac353b8a41f501dd76bf53288dd478


giactra (approved):
This is beyond the scope of the patch but I find it misleading there are two regCache64 members in this class (one is inherited from the ArmISA::RemoteGDB base). I think we should rename this to avoid any confusion

PR #739, opened 2024-01-05
misc: Update .mailmap file

ivanaamit:
The .mailmap file is designed to maintain a record of unique contributors, aiming for a single identifier for each person. What is included in this file does not impact or alter commits; rather, it just merges the counts for all commits by one person under a single name.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,8 +1,11 @@
 Abdul Mutaal Ahmad <abdul.mutaal@gmail.com>
 adarshpatil <adarshpatil123@gmail.com>
+Aditya K Kamath <a_kamath@hotmail.com> aditya <a_kamath@hotmail.com>
 Adrià Armejach <adria.armejach@bsc.es> Adrià Armejach <adria.armejach@gmail.com>
+Adrià Armejach <adria.armejach@bsc.es> Adrià Armejach <66964292+aarmejach@users.noreply.github.com>
 Adrian Herrera <adrian.herrera@arm.com>
 Adrien Pesle <adrien.pesle@arm.com>
+Adwaith R Krishna <adwaithrk19@gmail.com>
 Akash Bagdia <akash.bagdia@ARM.com> Akash Bagdia <akash.bagdia@arm.com>
 Alec Roelke <alec.roelke@gmail.com> Alec Roelke <ar4jc@virginia.edu>
 Alexander Klimov <Alexander.Klimov@arm.com>
@@ -10,21 +13,19 @@ Alexandru Dutu <alexandru.dutu@amd.com> Alexandru <alexandru.dutu@amd.com>
 Alex Richardson <alexrichardson@google.com>
 Ali Jafri <ali.jafri@arm.com>
 Ali Saidi <Ali.Saidi@arm.com> Ali Saidi <ali.saidi@arm.com>
-Ali Saidi <Ali.Saidi@arm.com> Ali Saidi <Ali.Saidi@ARM.com>
 Ali Saidi <Ali.Saidi@arm.com> Ali Saidi <saidi@eecs.umich.edu>
 Alistair Delva <adelva@google.com>
+Alvaro Moreno <alvaro.moreno@bsc.es>
 Amin Farmahini <aminfar@gmail.com>
 Anders Handler <s052838@student.dtu.dk>
 Andrea Mondelli <andrea.mondelli@huawei.com> Andrea Mondelli <andrea.mondelli@ucf.edu>
-Andrea Mondelli <andrea.mondelli@huawei.com> Andrea Mondelli <Andrea.Mondelli@ucf.edu>
 Andrea Pellegrini <andrea.pellegrini@gmail.com>
 Andreas Hansson <andreas.hanson@arm.com> Andreas Hansson <andreas.hansson>
 Andreas Hansson <andreas.hanson@arm.com> Andreas Hansson <andreas.hansson@arm.com>
-Andreas Hansson <andreas.hanson@arm.com> Andreas Hansson <Andreas.Hansson@ARM.com>
 Andreas Hansson <andreas.hanson@arm.com> Andreas Hansson <andreas.hansson@armm.com>
 Andreas Sandberg <Andreas.Sandberg@arm.com> Andreas Sandberg <andreas.sandberg@arm.com>
-Andreas Sandberg <Andreas.Sandberg@arm.com> Andreas Sandberg <Andreas.Sandberg@ARM.com>
 Andreas Sandberg <Andreas.Sandberg@arm.com> Andreas Sandberg <andreas@sandberg.pp.se>
+Andreas Sandberg <Andreas.Sandberg@arm.com> Andreas Sandberg <andreas@sandberg.uk>
 Andrew Bardsley <Andrew.Bardsley@arm.com> Andrew Bardsley <Andreas.Bardsley@arm.com>
 Andrew Lukefahr <lukefahr@umich.edu>
 Andrew Schultz <alschult@umich.edu>
@@ -32,11 +33,14 @@ Andriani Mappoura <andriani.mappoura@arm.com>
 Angie Lee <peiyinglee@google.com>
 Anis Peysieux <anis.peysieux@inria.fr>
 Ani Udipi <ani.udipi@arm.com>
+anoop <mysanoop@gmail.com>
 Anouk Van Laer <anouk.vanlaer@arm.com>
 ARM gem5 Developers <none@none>
 Arthur Perais <Arthur.Perais@univ-grenoble-alpes.fr> Arthur Perais <arthur.perais@inria.fr>
 Arun Rodrigues <afrodri@gmail.com>
 Ashkan Tousi <ashkan.tousimojarad@arm.com>
+atrah22 <atul.rahman@outlook.com>
+Atri Bhattacharyya <atri.bhattacharyya@epfl.ch>
 Austin Harris <austinharris@utexas.edu> Austin Harris <mail@austin-harris.com>
 Avishai Tvila <avishai.tvila@gmail.com>
 Ayaz Akram <yazakram@ucdavis.edu>
@@ -48,6 +52,7 @@ Bjoern A. Zeeb <baz21@cam.ac.uk>
 Blake Hechtman <bah13@duke.edu> Blake Hechtman <blake.hechtman@amd.com>
 Blake Hechtman <bah13@duke.edu> Blake Hechtman ext:(%2C%20Nilay%20Vaish%20%3Cnilay%40cs.wisc.edu%3E) <bah13@duke.edu>
 Bobby R. Bruce <bbruce@ucdavis.edu> Bobby Bruce <bbruce@amarillo.cs.ucdavis.edu>
+Bobby R. Bruce <bbruce@ucdavis.edu> Bobby Bruce <bbruce@ucdavis.edu>
 Boris Shingarov <shingarov@gmail.com> Boris Shingarov <shingarov@labware.com>
 Brad Beckmann <brad.beckmann@amd.com> Brad Beckmann <Brad.Beckmann@amd.com>
 Brad Beckmann <brad.beckmann@amd.com> Brad Beckmann ext:(%2C%20Nilay%20Vaish%20%3Cnilay%40cs.wisc.edu%3E) <Brad.Beckmann@amd.com>
@@ -60,15 +65,13 @@ Brian Grayson <b.grayson@samsung.com>
 Cagdas Dirik <cdirik@micron.com> cdirik <cdirik@micron.com>
 Carlos Falquez <c.falquez@fz-juelich.de>
 Chander Sudanthi <chander.sudanthi@arm.com> Chander Sudanthi <Chander.Sudanthi@arm.com>
-Chander Sudanthi <chander.sudanthi@arm.com> Chander Sudanthi <Chander.Sudanthi@ARM.com>
 Charles Jamieson <cjamieson2@wisc.edu>
-CHEN Meng <tundriolaxy@gmail.com>
+Chen Meng <tundriolaxy@gmail.com>
 Chen Zou <chenzou@uchicago.edu>
 Chia-You Chen <hortune@google.com>
-Chow, Marcus <marcus.chow@amd.com>
+Marcus Chow <marcus.chow@amd.com>
 Chris Adeniyi-Jones <Chris.Adeniyi-Jones@arm.com>
 Chris Emmons <chris.emmons@arm.com> Chris Emmons <Chris.Emmons@arm.com>
-Chris Emmons <chris.emmons@arm.com> Chris Emmons <Chris.Emmons@ARM.com>
 Chris January <chris.january@arm.com>
 Christian Menard <christian.menard@tu-dresden.de> Christian Menard <Christian.Menard@tu-dresden.de>
 Christopher Torng <clt67@cornell.edu>
@@ -83,17 +86,19 @@ Daecheol You <daecheol.you@samsung.com>
 Dam Sunwoo <dam.sunwoo@arm.com>
 Dan Gibson <gibson@cs.wisc.edu>
 Daniel Carvalho <odanrc@yahoo.com.br> Daniel <odanrc@yahoo.com.br>
+Daniel Carvalho <odanrc@yahoo.com.br> Daniel Carvalho <odanrc@users.noreply.github.com>
 Daniel Carvalho <odanrc@yahoo.com.br> Daniel R. Carvalho <odanrc@yahoo.com.br>
 Daniel Gerzhoy <daniel.gerzhoy@gmail.com>
 Daniel Johnson <daniel.johnson@arm.com>
+Daniel Kouchekinia <DanKouch@users.noreply.github.com>
 Daniel Sanchez <sanchezd@stanford.edu>
 Davide Basilio Bartolini <davide.basilio.bartolini@huawei.com>
 David Guillen-Fandos <david.guillen@arm.com> David Guillen <david.guillen@arm.com>
 David Guillen-Fandos <david.guillen@arm.com> David Guillen Fandos <david.guillen@arm.com>
 David Hashe <david.hashe@amd.com> David Hashe <david.j.hashe@gmail.com>
 David Oehmke <doehmke@umich.edu>
-David Schall <david.schall2@arm.com>
-Derek Christ <dchrist@rhrk.uni-kl.de>
+David Schall <david.schall@ed.ac.uk> David Schall <david.schall2@arm.com>
+Derek Christ <dchrist@rhrk.uni-kl.de> Derek Christ <44267643+derchr@users.noreply.github.com>
 Derek Hower <drh5@cs.wisc.edu>
 Deyaun Guo <guodeyuan@tsinghua.org.cn> Deyuan Guo ext:(%2C%20Nilay%20Vaish%20%3Cnilay%40cs.wisc.edu%3E) <guodeyuan@tsinghua.org.cn>
 Deyaun Guo <guodeyuan@tsinghua.org.cn> Deyuan Guo <guodeyuan@tsinghua.org.cn>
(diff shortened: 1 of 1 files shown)


.mailmap
```diff
@@ -140,10 +154,14 @@ GWDx <gwdx@mail.ustc.edu.cn>
 Hamid Reza Khaleghzadeh <khaleghzadeh@gmail.com> Hamid Reza Khaleghzadeh ext:(%2C%20Lluc%20Alvarez%20%3Clluc.alvarez%40bsc.es%3E%2C%20Nilay%20Vaish%20%3Cnilay%40cs.wisc.edu%3E) <khaleghzadeh@gmail.com>
 handsomeliu <handsomeliu@google.com>
 Hanhwi Jang <jang.hanhwi@gmail.com>
-Hoa Nguyen <hoanguyen@ucdavis.edu>
+Harshil Patel <hpppatel@ucdavis.edu> Harshil Patel <harshilp2107@gmail.com>
+Harshil Patel <hpppatel@ucdavis.edu> Harshil Patel <91860903+Harshil2107@users.noreply.github.com>
+He, Wenjian <wheac@connect.ust.hk>
```
powerjg: Probably want to remove the comma and swap the name order. I could be wrong, though.

.mailmap
```diff
@@ -372,22 +418,25 @@ Steve Reinhardt <stever@gmail.com> Steve Reinhardt <stever@eecs.umich.edu>
 Steve Reinhardt <stever@gmail.com> Steve Reinhardt <steve.reinhardt@amd.com>
 Steve Reinhardt <stever@gmail.com> Steve Reinhardt <Steve.Reinhardt@amd.com>
 Stian Hvatum <stian@dream-web.no>
+studyztp <studyztp@gmail.com>
```
powerjg: This is Zhantong Qiu (Who may also have a UCD email somewhere...)

.mailmap
```diff
@@ -233,20 +265,21 @@ Matteo Andreozzi <matteo.andreozzi@arm.com> Matteo Andreozzi <Matteo.Andreozzi@a
...
-Matthias Jung <jungma@eit.uni-kl.de>
-Matthias Jung <matthias.jung@iese.fraunhofer.de>
+Matthias Jung <matthias.jung@iese.fraunhofer.de> Matthias Jung <jungma@eit.uni-kl.de>
 Matt Horsnell <matt.horsnell@arm.com> Matt Horsnell <matt.horsnell@ARM.com>
 Matt Horsnell <matt.horsnell@arm.com> Matt Horsnell <Matt.Horsnell@arm.com>
 Matt Horsnell <matt.horsnell@arm.com>Matt Horsnell <Matt.Horsnell@ARM.com>
 Matt Poremba <matthew.poremba@amd.com> Matt Poremba <Matthew.Poremba@amd.com>
-Matt Sinclair <mattdsinclair@gmail.com> Matthew Sinclair <matthew.sinclair@amd.com>
+Matt Sinclair <mattdsinclair.wisc@gmail.com> Matt Sinclair <mattdsinclair@gmail.com>
+Matt Sinclair <mattdsinclair.wisc@gmail.com> Matthew Sinclair <matthew.sinclair@amd.com>
```
mattsinc, line 267: I think 276 and 277 are the same except for capitalization?

ivanaamit: Fixed it. Thanks.

.mailmap
```diff
@@ -48,6 +52,7 @@ Bjoern A. Zeeb <baz21@cam.ac.uk>
 Blake Hechtman <bah13@duke.edu> Blake Hechtman <blake.hechtman@amd.com>
 Blake Hechtman <bah13@duke.edu> Blake Hechtman ext:(%2C%20Nilay%20Vaish%20%3Cnilay%40cs.wisc.edu%3E) <bah13@duke.edu>
 Bobby R. Bruce <bbruce@ucdavis.edu> Bobby Bruce <bbruce@amarillo.cs.ucdavis.edu>
+Bobby R. Bruce <bbruce@ucdavis.edu> Bobby Bruce <bbruce@ucdavis.edu>
 Boris Shingarov <shingarov@gmail.com> Boris Shingarov <shingarov@labware.com>
 Brad Beckmann <brad.beckmann@amd.com> Brad Beckmann <Brad.Beckmann@amd.com>
 Brad Beckmann <brad.beckmann@amd.com> Brad Beckmann ext:(%2C%20Nilay%20Vaish%20%3Cnilay%40cs.wisc.edu%3E) <Brad.Beckmann@amd.com>
```
mattsinc, line 58: Probably we want line 58 to be fixed as Nilay's email address is embedded in Brad's info?

Also I think 57 and 58 are identical except for this issue.

.mailmap
```diff
@@ -83,17 +86,19 @@ Daecheol You <daecheol.you@samsung.com>
...
 Dan Gibson <gibson@cs.wisc.edu>
 Daniel Carvalho <odanrc@yahoo.com.br> Daniel <odanrc@yahoo.com.br>
+Daniel Carvalho <odanrc@yahoo.com.br> Daniel Carvalho <odanrc@users.noreply.github.com>
 Daniel Carvalho <odanrc@yahoo.com.br> Daniel R. Carvalho <odanrc@yahoo.com.br>
 Daniel Gerzhoy <daniel.gerzhoy@gmail.com>
 Daniel Johnson <daniel.johnson@arm.com>
+Daniel Kouchekinia <DanKouch@users.noreply.github.com>
 Daniel Sanchez <sanchezd@stanford.edu>
 Davide Basilio Bartolini <davide.basilio.bartolini@huawei.com>
 David Guillen-Fandos <david.guillen@arm.com> David Guillen <david.guillen@arm.com>
```
mattsinc, line 96: There are some minor differences in last name (hyphen vs no hyphen) but otherwise 96 and 97 appear to be the same.

.mailmap
```diff
@@ -107,11 +112,12 @@ Earl Ou <shunhsingou@google.com>
 eavivi <eavivi@ucdavis.edu>
 Éder F. Zulian <zulian@eit.uni-kl.de>
 Edmund Grimley Evans <Edmund.Grimley-Evans@arm.com>
-Eduardo José Gómez Hernández <eduardojose.gomez@um.es>
+Eduardo José Gómez Hernández <eduardojose.gomez@um.es> Eduardo José Gómez Hernández <git@edujgh.net>
 Eliot Moss <moss@cs.umass.edu>
 Emilio Castillo <castilloe@unican.es> Emilio Castillo <ecastill@bsc.es>
 Emilio Castillo <castilloe@unican.es> Emilio Castillo ext:(%2C%20Nilay%20Vaish%20%3Cnilay%40cs.wisc.edu%3E) <castilloe@unican.es>
```
mattsinc, line 118: Nilay's info embedded in 118 again.

After fixing, 117 and 118 seem the same.

.mailmap
```diff
@@ -206,11 +224,16 @@ Koan-Sin Tan <koansin.tan@gmail.com>
...
 Krishnendra Nathella <Krishnendra.Nathella@arm.com> Krishnendra Nathella <krinat01@arm.com>
 ksco <numbksco@gmail.com>
-kunpai <kunpai@ucdavis.edu>
+Kunal Pai <kunpai@ucdavis.edu> Kunal Pai <62979320+kunpai@users.noreply.github.com>
+Kunal Pai <kunpai@ucdavis.edu> kunpai <kunpai@ucdavis.edu>
+Kunal Pai <kunpai@ucdavis.edu> paikunal <kunpai@ucdavis.edu>
+Kunal Pai <kunpai@ucdavis.edu> KUNAL PAI <kunpai@ucdavis.edu>
 Kyle Roarty <kyleroarty1716@gmail.com> Kyle Roarty <Kyle.Roarty@amd.com>
 Laura Hinman <llhinman@ucdavis.edu>
 Lena Olson <leolson@google.com> Lena Olson <lena@cs.wisc,edu>
```
mattsinc, line 233: 233 and 234 are identical

.mailmap
mattsinc: I am not sure I fully understand the need to have two very similar versions of my info here, but if this satisfies the Davis folks, fine with me.

.mailmap
```diff
@@ -279,21 +301,24 @@ Nathan Binkert <nate@binkert.org> Nathan Binkert <binkertn@umich.edu>
...
 Neil Natekar <nanatekar@ucdavis.edu>
-Nicholas Lindsay <nicholas.lindsay@arm.com>
+Nicholas Lindsay <nicholas.lindsay@arm.com> Nicholas Lindsay <Nicholas.Lindsey@arm.com>
+Nicholas Mosier <nmosier@stanford.edu> Nicholas Mosier <nh.mosier@gmail.com>
 Nicolas Boichat <drinkcat@google.com>
 Nicolas Derumigny <nderumigny@gmail.com>
 Nicolas Zea <nicolas.zea@gmail.com>
+Nikolaos Kyparissas <nikolaos.kyparissas@arm.com>
 Nikos Nikoleris <nikos.nikoleris@arm.com> Nikos Nikoleris <nikos.nikoleris@gmail.com>
 Nilay Vaish ext:(%2C%20Timothy%20Jones%20%3Ctimothy.jones%40cl.cam.ac.uk%3E) <nilay@cs.wisc.edu>
```
mattsinc, line 311: Nilay's info has Tim Jones' embedded into it.

.mailmap
```diff
@@ -308,29 +333,39 @@ Po-Hao Su <supohaosu@gmail.com>
 Polina Dudnik <pdudnik@cs.wisc.edu> Polina Dudnik <pdudnik@gmail.com>
 Polydoros Petrakis <ppetrak@ics.forth.gr>
 Pouya Fotouhi <pfotouhi@ucdavis.edu> Pouya Fotouhi <Pouya.Fotouhi@amd.com>
+Prajwal Hegde <prhegde@wisc.edu>
 Prakash Ramrakhyani <prakash.ramrakhyani@arm.com> Prakash Ramrakhani <Prakash.Ramrakhani@arm.com>
```
mattsinc, line 337: 337 and 338 identical.

.mailmap
```diff
@@ -308,29 +333,39 @@ Po-Hao Su <supohaosu@gmail.com>
...
+Pu (Luke) Yi <lukeyi@stanford.edu>
 Quentin Forcioli <quentin.forcioli@telecom-paris.fr>
 Radhika Jagtap <radhika.jagtap@arm.com> Radhika Jagtap <radhika.jagtap@ARM.com>
 Rahul Thakur <rjthakur@google.com>
-Reiley Jeapaul <Reiley.Jeyapaul@arm.com>
+Rajarshi Das <drajarsh@gmail.com>
+Ranganath (Bujji) Selagamsetty <bujji.selagamsetty@amd.com> BujSet <ranganath1000@gmail.com>
+Razeza <borisov.dn@phystech.edu>
+Reiley Jeapaul <reiley.jeyapaul@arm.com> Reiley Jeapaul <Reiley.Jeyapaul@arm.com>
 Rekai Gonzalez-Alberquilla <rekai.gonzalezalberquilla@arm.com> Rekai Gonzalez Alberquilla <rekai.gonzalezalberquilla@arm.com>
```
mattsinc, line 348: 348 and 349 identical.

.mailmap
ivanaamit: They are not identical. One is ",edu" and the other one is ".edu".

.mailmap
powerjg: I think there may be some confusion about the purpose/syntax of this file. Here's link to the documentation: https://git-scm.com/docs/gitmailmap

Basically, each line is `[correct info] [incorrect info that should be the info on the left]`.

So, in this case, Brad made a change, but Nilay committed on behalf of brad (because this was how it worked with mercurial before we moved to git). So, the commit was authored by Brad, not Nilay. Nilay was just the one that happened to push upstream (again, mercurial was different from the way git works and the conversion from hg to git wasn't perfect).

So, this is correct. Anytime there is a commit that says "Brad Beckmann ext:(..." we should interpret this as "Brad Beckmann <brad.beckmann@amd.com>".

The main point of this file is so we can track community stats. So, we want to make sure that we have unique (and preferably correct) names on the left hand side of each line. This will allow us to answer questions like "how many people have contributed to the gem5 codebase" or "what is the number of committers with more that 10 commits in the past 2 years."

Finally, I'll point out this *doesn't change the commit info at all*. So, all of the details of which institution is still available. I believe that `git shortlog` and a couple of other commands like that are the only times the mailmap file is used.

.mailmap
powerjg: Right, that's the whole point of this file. This is the same person. So we want `David Guillen <david.guillen@arm.com>`, `David Guillen Fandos <david.guillen@arm.com>`, and `David Guillen-Fandos <david.guillen@arm.com>` to show up as David Guillen-Fandos <david.guillen@arm.com>`. 

We *do not want* these three "different" committers to show up as three different individuals when we run `git shortlog`. We want this to show up as a single entry. This file explains how to relate different committer details to the same human.

.mailmap
mattsinc: The ",edu" one is a typo then, as ",edu" is not a valid email address.

.mailmap
mattsinc: ok, as long as the ones I pointed out that seem duplicated are intentional, then that's fine

.mailmap
powerjg: Right... again, that's the point of this file is to fix these "typos" in the commits. We can't change the commit information, but we can give git a way to know that the email `lena@cs.wisc.edu` and `lena@cs.wisc,edu` are actually the same person, which is also the same as `leolson@google.com`

mattsinc: I think our messages are just flying past one another, but just acknowledging you answered this with your previous post.

powerjg: No worries. Thanks for the careful review!

.mailmap
```diff
@@ -1,42 +1,46 @@
...
 Anders Handler <s052838@student.dtu.dk>
 Andrea Mondelli <andrea.mondelli@huawei.com> Andrea Mondelli <andrea.mondelli@ucf.edu>
-Andrea Mondelli <andrea.mondelli@huawei.com> Andrea Mondelli <Andrea.Mondelli@ucf.edu>
 Andrea Pellegrini <andrea.pellegrini@gmail.com>
 Andreas Hansson <andreas.hanson@arm.com> Andreas Hansson <andreas.hansson>
 Andreas Hansson <andreas.hanson@arm.com> Andreas Hansson <andreas.hansson@arm.com>
-Andreas Hansson <andreas.hanson@arm.com> Andreas Hansson <Andreas.Hansson@ARM.com>
 Andreas Hansson <andreas.hanson@arm.com> Andreas Hansson <andreas.hansson@armm.com>
 Andreas Sandberg <Andreas.Sandberg@arm.com> Andreas Sandberg <andreas.sandberg@arm.com>
-Andreas Sandberg <Andreas.Sandberg@arm.com> Andreas Sandberg <Andreas.Sandberg@ARM.com>
```
BobbyRBruce, line 26: I _think_ there were reasons we kept these capitalizations. Some commits had them and we needed to include these entries to match them (there was some case-sensitivity in the way the author search works in git, maybe (??)). 

Have we proven they are unnecessary?

BobbyRBruce (commented):
I just want some clarification about the `ARM.come` deletions. If they are fine, then it's a thumbs-up from me.

.mailmap
ivanaamit: It looks like the mailmap file works case-insensitively. You can check it out [here](https://git-scm.com/docs/gitmailmap).

andysan: It's worth checking if this actually works the way it's specified in the documentation. I tried removing the different capitalisations locally at one point and it seemed like the `git-shortlog` didn't merge entries correctly at the time.

andysan:
The mailmap file was briefly discussed in PMC last year. However, it seems like the conclusions weren't implemented at the time.

The short summary is that we currently map the tuple name and email to an arbitrary constant mail/email tuple. In some cases, even an outdated email address. This can sometimes be problematic. For example, I have contributed to gem5 in a personal capacity where I personally own the copyright of the commits that get mapped to my Arm email making it seem like these should be copyright Arm. Similarly, there are others who have left Arm and now work on other architectures, but their contributions still get attributed to Arm. What's worse is that their new commits get mapped to an outdated email address they no longer have access to.

It was suggested that we should apply a variant of these rules at the time:
I think we need to come up with some policy for how we manage the mailmap file since this is potentially quite sensitive. In my opinion, we should have rules along these lines:

- Anyone can submit "obvious" fixes where obvious is minor changes such as consistent use of uppercase/lowercase and obvious misspellings.
- Obvious email remapping, such as spelling and short/long form but obviously belonging to the same institution, should be acceptable.
- Changes to a domain name should only be done if requested by the owner of the email address. For example, my old contributions using my personal email address shouldn't be attributed to Arm.
- Name changes (e.g., after getting married) should normally only be done if requested by the affected person. Adding/removing a middle name, spelling, or replacing nicknames would be exceptions to this rule.

We should probably document this somewhere better at some point. Maybe as a discussion or issue?

.mailmap
ivanaamit: We did check before pushing the commit that the file works case-insensitively. Maybe it didn't work that way before.

powerjg:
@andysan, I'd like to kindly remind you that the mailmap file is not a legal document, nor does it alter the commit history when the institution for commits is changed. This point is crucial and, as you may notice from previous discussions in this thread, has been thoroughly clarified: **The file is solely used for statistical purposes**. For instance, it helps in reporting to funders about the growth of our community. The **exclusive purpose** of this file is to track statistics, ensuring that committer information, which remains unaltered, is accurately associated with real individuals.

andysan:
> @andysan, I'd like to kindly remind you that the mailmap file is not a legal document, nor does it alter the commit history when the institution for commits is changed. This point is crucial and, as you may notice from previous discussions in this thread, has been thoroughly clarified: **The file is solely used for statistical purposes**. For instance, it helps in reporting to funders about the growth of our community. The **exclusive purpose** of this file is to track statistics, ensuring that committer information, which remains unaltered, is accurately associated with real individuals.

I understand this. The problem is that it affects how the normal `git-log` command displays commits by default. Unless you actively disable mailmap, it looks like the commit history has changed. It's made worse by the fact that there are known cases where the commit log doesn't show the most up-to-date email address even though the original commit does. This is clearly wrong.

In most cases, there is no need to remap email addresses across domains. The git-shortlog command is able to merge commits from the same author using multiple email addresses (I believe this is the default behaviour) and it supports showing them as separate committers. In fact, I would argue that mapping commits to across institutions is detrimental for statistics since this incorrectly inflates/deflates an institution's contributions to gem5. For example, multiple RISC-V commits are incorrectly being attributed to Arm because of this. This is also clearly wrong.

powerjg:
> > @andysan, I'd like to kindly remind you that the mailmap file is not a legal document, nor does it alter the commit history when the institution for commits is changed. This point is crucial and, as you may notice from previous discussions in this thread, has been thoroughly clarified: **The file is solely used for statistical purposes**. For instance, it helps in reporting to funders about the growth of our community. The **exclusive purpose** of this file is to track statistics, ensuring that committer information, which remains unaltered, is accurately associated with real individuals.
> 
> I understand this. The problem is that it affects how the normal `git-log` command displays commits by default. Unless you actively disable mailmap, it looks like the commit history has changed. It's made worse by the fact that there are known cases where the commit log doesn't show the most up-to-date email address even though the original commit does. This is clearly wrong.
> 
> In most cases, there is no need to remap email addresses across domains. The git-shortlog command is able to merge commits from the same author using multiple email addresses (I believe this is the default behaviour) and it supports showing them as separate committers. In fact, I would argue that mapping commits to across institutions is detrimental for statistics since this incorrectly inflates/deflates an institution's contributions to gem5. For example, multiple RISC-V commits are incorrectly being attributed to Arm because of this. This is also clearly wrong.

I'm pretty sure it's a local option to use the mailmap for git-log. See https://stackoverflow.com/questions/13777171/configuring-git-log-to-use-mailmap-by-default and https://git-scm.com/docs/git-log#Documentation/git-log.txt---no-mailmap

BobbyRBruce (approved):

powerjg:
I went ahead and merged as-is, because I don't think this changes anything with respect to legal status compared to before it was merged. 

If someone else wants to make updates to this file (or delete it altogether), feel free to create a PR so we can discuss.

andysan:
> > > @andysan, I'd like to kindly remind you that the mailmap file is not a legal document, nor does it alter the commit history when the institution for commits is changed. This point is crucial and, as you may notice from previous discussions in this thread, has been thoroughly clarified: **The file is solely used for statistical purposes**. For instance, it helps in reporting to funders about the growth of our community. The **exclusive purpose** of this file is to track statistics, ensuring that committer information, which remains unaltered, is accurately associated with real individuals.
> > 
> > 
> > I understand this. The problem is that it affects how the normal `git-log` command displays commits by default. Unless you actively disable mailmap, it looks like the commit history has changed. It's made worse by the fact that there are known cases where the commit log doesn't show the most up-to-date email address even though the original commit does. This is clearly wrong.
> > In most cases, there is no need to remap email addresses across domains. The git-shortlog command is able to merge commits from the same author using multiple email addresses (I believe this is the default behaviour) and it supports showing them as separate committers. In fact, I would argue that mapping commits to across institutions is detrimental for statistics since this incorrectly inflates/deflates an institution's contributions to gem5. For example, multiple RISC-V commits are incorrectly being attributed to Arm because of this. This is also clearly wrong.
> 
> I'm pretty sure it's a local option to use the mailmap for git-log. See https://stackoverflow.com/questions/13777171/configuring-git-log-to-use-mailmap-by-default and https://git-scm.com/docs/git-log#Documentation/git-log.txt---no-mailmap

Yes, there is and it seems to use mailmap by default on Ubuntu 22.04.

I have looked a bit at how this is done in other projects. Linux makes extensive use of mailmap and Zephyr seems to use it as well. In both of these projects, it seems like the assumption is that emails follow the author (the sign-off line is what is used to identify copyright) most of the time. I.e., mailmap maps to the most recent valid email address. I think that's what we are trying to achieve here. However, there are multiple cases where we map correct email addresses to incorrect email addresses that are no longer in use which seems wrong.

It also seems like mailmap is almost exclusively updated by the people affected by the mapping. That's probably not feasible in gem5 though.

PR #760, opened 2024-01-11
misc: Move indexing and replacement policies to base

pranith:
I tried moving the generic policies to base as suggested in this patch.

Change-Id: I22b4e18e002feaee681ea0fa91a050d274345c15


powerjg:
lgtm. I'm happy with this change. We may want to give everyone else a bit of time to take a look, though.

odanrc:
Should we create a new dir in base called cache to put these and the new cache class? Can you think of a scenario where the IP or the RP could be used in a context other than caches? I can´t think of any.

pranith:
Agree, moving them to a cache subdir in base will keep it cleaner. Please let me know if that is preferred and I will create a new PR for all the cache related commits with this new directory structure.

powerjg:
+1 I like the idea!

pranith:
Updated and pushed to the generic cache patch.

PR #762, opened 2024-01-12
Updating strided generator

mahyarsamani:
This change improves the functionality of strided generator to create trace with better flexibility. 
It allows the user to manually set offset and stride size instead of calculating it based on a "gen_id".
This way different patterns could be created with the same SimObject.
In addition, this change adds stdlib components for strided generator.


giactra:
Hey Mahyar, it seems like you have added by mistake some sim.output/stats files to the commits/PR (unless this was intentional)

ivanaamit:
Hi @mahyarsamani, you can install pre-commit to fix the first test.

https://github.com/gem5/gem5/blob/stable/CONTRIBUTING.md#using-pre-commit

mahyarsamani:
> Hey Mahyar, it seems like you have added by mistake some sim.output/stats files to the commits/PR (unless this was intentional)
Hi Giacomo. Yes, they are added intentionally. Basically their purpose is to track changes in stats generated by the memory (trusted_stats.json) and the trace of addresses generated by the traffic generators (trusted_simout.txt).

giactra:
> > Hey Mahyar, it seems like you have added by mistake some sim.output/stats files to the commits/PR (unless this was intentional)
> > Hi Giacomo. Yes, they are added intentionally. Basically their purpose is to track changes in stats generated by the memory (trusted_stats.json) and the trace of addresses generated by the traffic generators (trusted_simout.txt).

Thanks Mahyar...
I appreciate the effort you put into this, but I have to say this looks weird to me for the following reasons

1) We are cluttering simulator with lots of output/stats files
2) I am not sure what is the simout guaranteeing:
Just posting an example:

> Global frequency set at 1000000000000 ticks per second
Beginning simulation!
Exiting @ tick 9576748 because system.processor.cores1.generator is finished updating the memory.
.

How is this proving that the strided generator works as intended?

3) gem5 tried in the past to do stats based regression checking and didn't go very well... Even by having a reasonable infostructure in place for visualizing and reporting stats differences (which we don't), some stats can inevitably change for other reasons. Take request latency as an example. A reasonable change in the memory controller or in the interconnect would modify those numbers and invalidate most of the tests...

IMHO if you want to make sure the strided generator is properly tested, you should come up with with some sort of unit-tests for the generator... Something like a simple config file and a generator that writes a specific pattern in memory.... Later memory should be inspected for that specific value according to the memory access pattern....

powerjg:
> gem5 tried in the past to do stats based regression checking and didn't go very well... Even by having a reasonable infostructure in place for visualizing and reporting stats differences (which we don't), some stats can inevitably change for other reasons.

@giactra this sounds like you're arguing against doing any statistics-based testing in gem5. I think it's really important that we not only check the functional correctness of gem5 but also ensure that commits don't make unintended changes to the timing of the simulation.

**We need a way to test the timing/statistics output of gem5.** This is  a specific feature request which I have heard from many current and prospective gem5 users.

I agree with you that this change is probably not doing a very good job of testing a single specific component of gem5, but I don't think we should completely dismiss testing the statistics.

**Do you have a specific suggestion for how to do this in a different way?**

What if we instead had 1 test for each traffic generator which was connected to *no cache* and *simple memory*. This would give us a check that the traffic generators aren't changing their timing.

Separately, we could take a few traffic patterns and connect them with *no cache* to some DRAM models to check that the DRAM models are not changing their timing unexpectedly.

Finally, we could take a few traffic patterns and connect them to different cache models and then to the *simple memory* to check that the caches aren't unexpectedly changing their timing.

We also probably need some scripts to do statistics updating in case a change *expectedly* causes differences in timing.

@mahyarsamani, can you split this PR up into two PRs: one for "updating the strided generator" and on for "adding statistics tests to the traffic generators" so we can incorporate the good work here while continuing to discuss the testing changes?

giactra:
> > gem5 tried in the past to do stats based regression checking and didn't go very well... Even by having a reasonable infostructure in place for visualizing and reporting stats differences (which we don't), some stats can inevitably change for other reasons.
> 
> @giactra this sounds like you're arguing against doing any statistics-based testing in gem5. I think it's really important that we not only check the functional correctness of gem5 but also ensure that commits don't make unintended changes to the timing of the simulation.

Let me backtrack here. I didn't mean to say any stats based regression testing (performance) is doomed to fail and it is not important. I believe it is utterly important but I don't think simple text based reference matching will give us what we want.

In general (not related to this change) I feel in order for stats based regression to be useful the following things should be in place

1) Conditionally allowing small variations in the stats output
2) Visualization tool for tracking stats variation, with capability to pinpoint outliers to PRs/nightly runs



> 
> **Do you have a specific suggestion for how to do this in a different way?**
> 

I proposed a functional validation example in my previous message

> What if we instead had 1 test for each traffic generator which was connected to _no cache_ and _simple memory_. This would give us a check that the traffic generators aren't changing their timing.
> 

This looks better IMHO. Also I believe we could use Monitors and query them from python (if it is not possible already, shouldn't be hard to implement) to extract relevant statistics like average latency/bandwidth.
In this way we won't have to store 800 text files in gem5, and we could encode the expected latency/bandwidth in the pythonfile (or pass it via cmdline)

mahyarsamani:
Hello all,

Sorry for my late reply and sorry for this mess and thank you for your feedback. To correct this, I will do the following:

1- I will keep this PR open, but I will remove the commits that relate to testing (very poorly) the traffic generators. This mean that this PR will update the c++ source code for strided generator and add stdlib components for it.
2- I will create a new PR that adds tests that only test out the functionality of the traffic generators ({LinearGenerator,RandomGenerator,StridedGenerator} + NoCache + SimpleMemory).

Let me know if you do not agree with this plan.

src/python/gem5/components/processors/strided_generator.py
```diff
@@ -0,0 +1,98 @@
+# Copyright (c) 2021 The Regents of the University of California
```
BobbyRBruce: Shouldn't this be 2023?

src/python/gem5/components/processors/strided_generator.py
```diff
@@ -0,0 +1,98 @@
...
+
+class StridedGenerator(AbstractGenerator):
+    def __init__(
+        self,
+        num_cores: int = 1,
+        duration: str = "1ms",
+        rate: str = "100GB/s",
+        block_size: int = 64,
+        superblock_size: int = 64,
+        stride_size: int = None,
```
BobbyRBruce: The typing here is wrong. It should be an optional integer if it can be `None`.

```suggestion
        stride_size: Optional[int] = None,
```

src/python/gem5/components/processors/strided_generator_core.py
```diff
@@ -0,0 +1,106 @@
+# Copyright (c) 2021 The Regents of the University of California
```
BobbyRBruce: Shouldn't this be 2023?

BobbyRBruce (changes requested):
This looks mostly correct to me. I just have some minor comments.

powerjg (commented):
Looks OK to me. It would be nice to have a test for this, but that's not 100% required.

BobbyRBruce (approved):

PR #767, opened 2024-01-12
misc: Added dependabot config file

Harshil2107:
- Added a yaml file to make dependabot target develop instead of stable.

Change-Id: I5b28c06960c5a346b40e2af8f9284b11d9cc07cd

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,17 @@
+---
+version: 2
+updates:
+    - package-ecosystem: pip
+      directory: /
+      schedule:
+          interval: monthly
+      assignees:
+          - Harshil2107
+      commit-message:
+          prefix: 'misc: '
+    # Raise pull requests for version updates
+    # to pip against the `develop` branch
+      target-branch: develop
+    # Labels on pull requests for version updates only
+      labels:
+          - misc


.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+    - package-ecosystem: pip
+      directory: /
+      schedule:
+          interval: monthly
+    # Raise pull requests for version updates
+    # to pip against the `develop` branch
+      target-branch: develop
+    # Labels on pull requests for version updates only
+      labels:
+          - pip dependencies
```
powerjg: Is this label right? It won't always be `pip` and will often be `npm`.

I think the right label is `misc`

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+---
+version: 2
+updates:
+    - package-ecosystem: pip
```
powerjg, line 4: I think `npm` is the one this has been targeting the most, right?

powerjg (commented):
While you're doing this why don't you also update some other options:
- `[assignees](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#assignees)` Put yourself as the assignee
- `[commit-message](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message)` with the right `prefix`. I.e., `misc:`

.github/workflows/dependabot.yml
Harshil2107: I can add npm but I dont think we use react/ node.js anywhere in gem5. I think so far all dependabot PRs have been python library updates.

.github/workflows/dependabot.yml
Harshil2107: I will update label to misc

.github/workflows/dependabot.yml
powerjg: My bad. You're right. I thought that the gem5-resources-manager was written with node.

powerjg (approved):

PR #745, opened 2024-01-07
Add a generic cache template library

pranith:
Add a generic cache template to construct internal storage structures. Also add some example use cases by converting the prefetcher tables to use this new library.

PR #769, opened 2024-01-14
mem-cache: Add a tag mask to indexing policies

pranith:
In some internal structures, a shorter tag is preferred to save storage. This can be achieved by techniques like hashing the full tag to generate the shortened tag or just ignoring the upper bits of the tag.

Adding a tag mask allows us to implement such techniques in gem5.

Change-Id: Ib339efd97bf33b706f93eb7f0eb5851ceb51afb6

### DIFF
--- a/src/mem/cache/tags/indexing_policies/base.hh
+++ b/src/mem/cache/tags/indexing_policies/base.hh
@@ -96,6 +96,12 @@ class BaseIndexingPolicy : public SimObject
      */
     const int tagShift;
 
+    /**
+     * The mask to be applied to the address to generate a tag.
+     * If this masks out unique bits, we can have aliases.
+     */
+    const uint64_t tagMask;
+
   public:
     /**
      * Convenience typedef.
--- a/src/mem/cache/tags/indexing_policies/IndexingPolicies.py
+++ b/src/mem/cache/tags/indexing_policies/IndexingPolicies.py
@@ -44,6 +44,11 @@ class BaseIndexingPolicy(SimObject):
     # Get the associativity
     assoc = Param.Int(Parent.assoc, "associativity")
 
+    # number of bits in the tag above the index and offset bit
+    # If this is less than 64, the other bits are ignored
+    # ignored_bits: (64 - tag_bits - index_bits - offset_bits)
+    tag_bits = Param.Unsigned(64, "number of bits in the tag")
+
 
 class SetAssociative(BaseIndexingPolicy):
     type = "SetAssociative"
--- a/src/mem/cache/tags/indexing_policies/base.cc
+++ b/src/mem/cache/tags/indexing_policies/base.cc
@@ -59,7 +59,8 @@ BaseIndexingPolicy::BaseIndexingPolicy(const Params &p)
     : SimObject(p), assoc(p.assoc),
       numSets(p.size / (p.entry_size * assoc)),
       setShift(floorLog2(p.entry_size)), setMask(numSets - 1), sets(numSets),
-      tagShift(setShift + floorLog2(numSets))
+      tagShift(setShift + floorLog2(numSets)),
+      tagMask(mask(p.tag_bits))
 {
     fatal_if(!isPowerOf2(numSets), "# of sets must be non-zero and a power " \
              "of 2");
@@ -98,7 +99,7 @@ BaseIndexingPolicy::setEntry(ReplaceableEntry* entry, const uint64_t index)
 Addr
 BaseIndexingPolicy::extractTag(const Addr addr) const
 {
-    return (addr >> tagShift);
+    return (addr >> tagShift) & tagMask;
 }
 
 } // namespace gem5


andysan:
Is this actually desirable? We don't really save any space in the simulator by masking out bits here, so the only reason you'd do it is to exclude some top bits of the PA in the tag match. Is there a good reason for simulating this? Effectively, you would end up aliasing multiple PAs if this feature is ever used.

powerjg:
> Is this actually desirable? We don't really save any space in the simulator by masking out bits here, so the only reason you'd do it is to exclude some top bits of the PA in the tag match. Is there a good reason for simulating this? Effectively, you would end up aliasing multiple PAs if this feature is ever used.

IIRC, from the original discussion, this is exactly why the feature is needed. The BTB for instance allows aliasing of addresses.

andysan:
> > Is this actually desirable? We don't really save any space in the simulator by masking out bits here, so the only reason you'd do it is to exclude some top bits of the PA in the tag match. Is there a good reason for simulating this? Effectively, you would end up aliasing multiple PAs if this feature is ever used.
> 
> IIRC, from the original discussion, this is exactly why the feature is needed. The BTB for instance allows aliasing of addresses.

Hmm, OK, I highly doubt this is the case in modern BTBs though because of all of the security considerations post Spectre/Meltdown. That said, I think we need to generalise the way we represent tags and match tags to make it work for more cases. Specifically for structures like the BTB and TLB need more information than just an address in the tag (you'd normally include address space identifiers among other things).

pranith:
> Is this actually desirable? We don't really save any space in the simulator by masking out bits here, so the only reason you'd do it is to exclude some top bits of the PA in the tag match. Is there a good reason for simulating this? Effectively, you would end up aliasing multiple PAs if this feature is ever used.

The 'tagMask' field added here can be used to mask out any tag that the cache chooses to use and is not limited to PA. For example, in store sets, we can hash the PC and use only part of the hashed PC as the tag. Similarly, in way predictor, we can hash the VA and use part of the hashed VA as the tag.

pranith:
> 
> Hmm, OK, I highly doubt this is the case in modern BTBs though because of all of the security considerations post Spectre/Meltdown. That said, I think we need to generalise the way we represent tags and match tags to make it work for more cases. Specifically for structures like the BTB and TLB need more information than just an address in the tag (you'd normally include address space identifiers among other things).

Yes, agreed that this might not be enough for BTBs. We want to provide the flexibility for users to extend this and implement what is appropriate for their use case. The users are expected to override the 'setTag()/getTag()/extractSet()' etc., methods. An example for this can be found in how the BTBs/StoreSets are structured in PR #745.

andysan:
> > Hmm, OK, I highly doubt this is the case in modern BTBs though because of all of the security considerations post Spectre/Meltdown. That said, I think we need to generalise the way we represent tags and match tags to make it work for more cases. Specifically for structures like the BTB and TLB need more information than just an address in the tag (you'd normally include address space identifiers among other things).
> 
> Yes, agreed that this might not be enough for BTBs. We want to provide the flexibility for users to extend this and implement what is appropriate for their use case. The users are expected to override the 'setTag()/getTag()/extractSet()' etc., methods. An example for this can be found in how the BTBs/StoreSets are structured in PR #745.

I still think this change is very problematic. It's going to be hard for users to understand how to set this parameter correctly and it won't be a long-term solution either. In order to generalise the caches, we will need tags that are longer than standard integer types in C++ (this applies to at least TLBs and BTBs). This implies that masking won't work once that won't be well-defined when applied to a struct (I think this is what we'd end up with to generalise). In practice, I think we need a way to specify a custom tag type instead with its own equality operator. This would make it possible to implement the type of aliasing you're trying to emulate here without the limitations and it would make it clear to users that this is a very special mode of operation that is likely to break things.

pranith:
> 
> I still think this change is very problematic. It's going to be hard for users to understand how to set this parameter correctly and it won't be a long-term solution either. In order to generalise the caches, we will need tags that are longer than standard integer types in C++ (this applies to at least TLBs and BTBs). This implies that masking won't work once that won't be well-defined when applied to a struct (I think this is what we'd end up with to generalise). In practice, I think we need a way to specify a custom tag type instead with its own equality operator. This would make it possible to implement the type of aliasing you're trying to emulate here without the limitations and it would make it clear to users that this is a very special mode of operation that is likely to break things.

Agreed that we can further improve how the tag is structured. This particular implementation supports only upto 'Addr' width of tags. Adding a 'tagMask' as in this patch allows us to use tags that are smaller than the 'Addr' width. This is sufficient for certain use cases as evidenced in the store set and BTB patches.

Does it cover all use cases? No. We have to work on implementing that in the future.

pranith:
Hi, just wanted to know if there is any other review feedback for this change?

powerjg (dismissed):
This change is OK with me. I don't see it hurting anything, and if it's useful for someone it seems reasonable to merge.

src/mem/cache/tags/indexing_policies/base.hh
```diff
@@ -96,6 +96,11 @@ class BaseIndexingPolicy : public SimObject
      */
     const int tagShift;
 
+    /**
+     * The tag mask.
```
odanrc: Since this is probably not that common, I would improve the documentation a bit here (as ¨the¨ implies that the reader knows what this is). I would write something along the lines of ¨Mask to be applied to the address in order to generate a tag. If the mask masks out a couple of bits from the address some addresses will generate aliased tags.¨.

odanrc:
I tend to agree with @andysan here. Overall, by using 64 we would still be backwards compatible; however, for the very common case of tag extraction we now have an extra noop masking operation. Have you verified if there is an effect on performance? If so, how worse does it get?

Since it won´t currently be used by any component other than the BTB´s indexing policy and there is some controversy here, in order to get the wheel running why don´t we simply add this parameter to that policy?

src/mem/cache/tags/indexing_policies/base.hh
pranith: Updated the comment for the variable.

pranith:
> I tend to agree with @andysan here. Overall, by using 64 we would still be backwards compatible; however, for the very common case of tag extraction we now have an extra noop masking operation. Have you verified if there is an effect on performance? If so, how worse does it get?

We are adding one AND instruction to extract the tag compared to the original. I don't think it will add any measurable performance impact.

> 
> Since it won´t currently be used by any component other than the BTB´s indexing policy and there is some controversy here, in order to get the wheel running why don´t we simply add this parameter to that policy?

I understand your point, but I still think this is generic and simple enough to go into the base policy. We are also using in the store set tag extraction and having it implemented multiple times in each policy duplicates the code.

odanrc:
> > I tend to agree with @andysan here. Overall, by using 64 we would still be backwards compatible; however, for the very common case of tag extraction we now have an extra noop masking operation. Have you verified if there is an effect on performance? If so, how worse does it get?
> 
> We are adding one AND instruction to extract the tag compared to the original. I don't think it will add any measurable performance impact.

Yeah, likely not a big impact. The virtual extract tag call required from the workaround I suggested would likely be more costly than the extra AND instruction.

> > Since it won´t currently be used by any component other than the BTB´s indexing policy and there is some controversy here, in order to get the wheel running why don´t we simply add this parameter to that policy?
> 
> I understand your point, but I still think this is generic and simple enough to go into the base policy. We are also using in the store set tag extraction and having it implemented multiple times in each policy duplicates the code.

This brings us back to the point of isolating the changes needed by each PR. If we had a PR introducing the generic cache using the minimal API along with a use case (i.e., the prefetchers), and then another PR making the BTB use the generic cache that also modifies the generic cache´s API minimally to support that BTB change, then it would have been easier to understand the purpose and suitability of this solution as a reviewer. Right now, even though the changes are in separate PRs, the reviewers have to review them all together.

In any case, I won´t block this change because it is simple and is backwards compatible. In the future we could work on generalizing the caches to separate the secure bit and other design-specific info appropriately. It would be nice to wait a bit for @andysan´s input here before merging, though.

andysan:
> > > I tend to agree with @andysan here. Overall, by using 64 we would still be backwards compatible; however, for the very common case of tag extraction we now have an extra noop masking operation. Have you verified if there is an effect on performance? If so, how worse does it get?
> > 
> > 
> > We are adding one AND instruction to extract the tag compared to the original. I don't think it will add any measurable performance impact.
> 
> Yeah, likely not a big impact. The virtual extract tag call required from the workaround I suggested would likely be more costly than the extra AND instruction.
> 
> > > Since it won´t currently be used by any component other than the BTB´s indexing policy and there is some controversy here, in order to get the wheel running why don´t we simply add this parameter to that policy?
> > 
> > 
> > I understand your point, but I still think this is generic and simple enough to go into the base policy. We are also using in the store set tag extraction and having it implemented multiple times in each policy duplicates the code.
> 
> This brings us back to the point of isolating the changes needed by each PR. If we had a PR introducing the generic cache using the minimal API along with a use case (i.e., the prefetchers), and then another PR making the BTB use the generic cache that also modifies the generic cache´s API minimally to support that BTB change, then it would have been easier to understand the purpose and suitability of this solution as a reviewer. Right now, even though the changes are in separate PRs, the reviewers have to review them all together.
> 
> In any case, I won´t block this change because it is simple and is backwards compatible. In the future we could work on generalizing the caches to separate the secure bit and other design-specific info appropriately. It would be nice to wait a bit for @andysan´s input here before merging, though.

My take on this particular change is that it is going to become problematic and will need to be dropped once the generic cache supports more complex tags (structs). At that point, masking won't be a well-defined operation so we will need to drop this change or at least redefine it to work on the address only (which may not be what you want in those cases).

In my opinion, tags aliasing onto each other in this way is a weird corner case (why do you have a tag if you allow aliases?). Including that support in generic code will lead to confusing and unexpected behaviour when users set the tag mask incorrectly. I suspect most people will assume that tag mask refers to the operation where you extract the tag from the address rather than chopping bits off the tag.

I would prefer to just have a specialised "imprecise" indexing policy in the cases where this is needed. That would avoid any confusion and would work with complex tags once that's supported.

Anyway, I won't block this, but I don't think it is the right thing to do if we want to avoid confusion and have a well-thought-out API that works long term.

pranith:
Thank you @andysan for your review and comments.

 
> My take on this particular change is that it is going to become problematic and will need to be dropped once the generic cache supports more complex tags (structs). At that point, masking won't be a well-defined operation so we will need to drop this change or at least redefine it to work on the address only (which may not be what you want in those cases).

I don't see a problem with this :). Progress is iterative. If and when we come up with a better tag generation scheme, this change will not be relevant anymore. But, until then, I think it will suffice for the cases where the tag width does not exceed the address width.

>
> In my opinion, tags aliasing onto each other in this way is a weird corner case (why do you have a tag if you allow aliases?). 

Allowing tags to alias is a real-world use case employed throughout the processor pipeline. Many structures make a trade-off to reduce the tag size for area considerations or use the freed up space to store more tags. This is especially true in areas where you don't need to be 100% accurate like prefetchers for example. BTB is another example where more tags are preferred rather than accurate tags since there are other ways to ensure accuracy by the way of Branch Predictors. The case is similar for store sets and with way predictors, all of which can be implemented with generic indexing policies.

> Including that support in generic code will lead to confusing and unexpected behaviour when users set the tag mask incorrectly. I suspect most people will assume that tag mask refers to the operation where you extract the tag from the address rather than chopping bits off the tag.

We provide way more many ways for the users to mess things up if they change knobs that they shouldn't. In the default case, since the tag bits are set to 64, it should work as expected.

andysan:
> Thank you @andysan for your review and comments.
> 
> > My take on this particular change is that it is going to become problematic and will need to be dropped once the generic cache supports more complex tags (structs). At that point, masking won't be a well-defined operation so we will need to drop this change or at least redefine it to work on the address only (which may not be what you want in those cases).
> 
> I don't see a problem with this :). Progress is iterative. If and when we come up with a better tag generation scheme, this change will not be relevant anymore. But, until then, I think it will suffice for the cases where the tag width does not exceed the address width.

My problem with this change is that it introduces a public API (a param) that we know will need to be removed within the next 1-2 releases because we know it won't work with more general tags. We shouldn't do that, it's bad for usability, especially when the use case is somewhat niche. Furthermore, the more assumptions we make about tags being plain integers, the harder it will be to implement more general tag support.

> > In my opinion, tags aliasing onto each other in this way is a weird corner case (why do you have a tag if you allow aliases?).
> 
> Allowing tags to alias is a real-world use case employed throughout the processor pipeline. Many structures make a trade-off to reduce the tag size for area considerations or use the freed up space to store more tags. This is especially true in areas where you don't need to be 100% accurate like prefetchers for example. BTB is another example where more tags are preferred rather than accurate tags since there are other ways to ensure accuracy by the way of Branch Predictors. The case is similar for store sets and with way predictors, all of which can be implemented with generic indexing policies.

This is not true for BTBs any more. The  BTB is not using truncated tags on modern microarchitecture. This was the case in the distant past, but gem5 is a research tool and should be forward looking. In practice, BTB tags need to be exact because the microarchitecture needs to protect against poisoning attacks.

For other predictive structures, such as prefetchers, it seems like a better option would be to perform the truncation when generating the lookup key. Many of them would typically not use a plain address anyway and the meaning of the truncation is likely component-specific. Conceptually, I would like to think of the generic cache as an associative container (a bit like `std::map`) that doesn't alias. If my component needs to simulate aliasing for some reasons, that should be done in the component when generating the key.

IMO, having support for complex tags (which breaks this change) is much more valuable most precise structures effectively require this nowadays. For example, we should really be thinking of addresses as a tuple of address and address space specification. This is true both for VAs (ASID & VMID specify the address space) and PAs (secure/non-secure/realm/etc.). A flat 64b integer will never be able to support that properly.

pranith:
> My problem with this change is that it introduces a public API (a param) that we know will need to be removed within the next 1-2 releases because we know it won't work with more general tags. We shouldn't do that, it's bad for usability, especially when the use case is somewhat niche. Furthermore, the more assumptions we make about tags being plain integers, the harder it will be to implement more general tag support.

Nothing fundamentally changes from the state tags are in today. We currently use tags that fit in 64-bits. This change allows us to use tags less than that. If the use-case dictates smaller tags, this change is helpful. If in the future, we have better tag generation scheme, users will adopt that scheme. Not allowing smaller tags to be used today just so that they can use a better scheme way down the line does not seem appropriate to me.

> This is not true for BTBs any more. The BTB is not using truncated tags on modern microarchitecture. This was the case in the distant past, but gem5 is a research tool and should be forward looking. In practice, BTB tags need to be exact because the microarchitecture needs to protect against poisoning attacks.

I disagree with the opinion that shorter tags are a niche. There are other ways to protect against poisoning attacks. In this particular case, you can use multi-level BTBs. The backing full-tag BTB can be accessed later to verify the predictions coming from the fast access truncated tag BTB.

I still think this is besides the point. How do we enable having shorter tags today?

> IMO, having support for complex tags (which breaks this change) is much more valuable most precise structures effectively require this nowadays. For example, we should really be thinking of addresses as a tuple of address and address space specification. This is true both for VAs (ASID & VMID specify the address space) and PAs (secure/non-secure/realm/etc.). A flat 64b integer will never be able to support that properly.

That is a welcome change when it is ready. No question about that.

andysan:
> > My problem with this change is that it introduces a public API (a param) that we know will need to be removed within the next 1-2 releases because we know it won't work with more general tags. We shouldn't do that, it's bad for usability, especially when the use case is somewhat niche. Furthermore, the more assumptions we make about tags being plain integers, the harder it will be to implement more general tag support.
> 
> Nothing fundamentally changes from the state tags are in today. We currently use tags that fit in 64-bits. This change allows us to use tags less than that. If the use-case dictates smaller tags, this change is helpful. If in the future, we have better tag generation scheme, users will adopt that scheme. Not allowing smaller tags to be used today just so that they can use a better scheme way down the line does not seem appropriate to me.

The problem is that this change is extremely short-sighted. We know that we will need to use structs to represent tags in the future to support important use cases like TLBs and caches with address space tagging. Merging something that is fundamentally incompatible with that seems extremely short-sighted, especially if we know there are better solutions like handling it in the caller.

> > This is not true for BTBs any more. The BTB is not using truncated tags on modern microarchitecture. This was the case in the distant past, but gem5 is a research tool and should be forward looking. In practice, BTB tags need to be exact because the microarchitecture needs to protect against poisoning attacks.
> 
> I disagree with the opinion that shorter tags are a niche. There are other ways to protect against poisoning attacks. In this particular case, you can use multi-level BTBs. The backing full-tag BTB can be accessed later to verify the predictions coming from the fast access truncated tag BTB.
> 
> I still think this is besides the point. How do we enable having shorter tags today?

Wouldn't masking the address in the caller have the exact same effect? This would have the benefit of making the aliasing behaviour component-specific which in my opinion would be more obvious to users. This is especially true when the key isn't an address (e.g., a history vector).

pranith:
> The problem is that this change is extremely short-sighted. We know that we will need to use structs to represent tags in the future to support important use cases like TLBs and caches with address space tagging. Merging something that is fundamentally incompatible with that seems extremely short-sighted, especially if we know there are better solutions like handling it in the caller.

If handling it in the caller is the right way, then may be TLBs and caches with address space tagging should handle it in the caller? :-)

> Wouldn't masking the address in the caller have the exact same effect? This would have the benefit of making the aliasing behaviour component-specific which in my opinion would be more obvious to users. This is especially true when the key isn't an address (e.g., a history vector).

You have to adapt it to your specific use-case by overriding the generic functions according to your needs.

Again, this patch does not change anything compared to the current state of things other than add support for shorter tags. If and when we have the better tag struct mechanism, we will adapt the users to that mechanism. I don't think it is appropriate to wait for a better implementation to come around rather than supporting the use-cases today.

BobbyRBruce:
I'm going to close this PR. It appears the discussion reached a point of stalemate and there's been no activity in months. If anyone wishes they may reopen and continue this work (or the discussion) if they see fit.

PR #770, opened 2024-01-14
Implement AssociativeSet from AssociativeCache

pranith:
This is for review purposes, does not build.


pranith:
@odanrc Updated with the review comments.

This PR is a subset of PR #745. This does not build as we need to make changes to the prefetch code but it is skipped for review purposes.

odanrc:
I don't understand. This PR was much simpler yesterday, and it looked like it would compile. Now it touches way more files to the point that it's looking like the "main" PR again. The idea of splitting the main PR was to create simpler PRs that would still be able to build. What's the error when we create a PR that only creates the new cache classes and updates only the prefetcher to use it?

pranith:
The problem is that I couldn't find a way to make one PR dependent on a previous PR. I tried splitting as:

PR 1: Move policies to base/cache
PR 2: Add cache template library (dependent on PR 1)
PR 3: AssociativeSet from cache (dependent on PR 2)
...

But all I can do is base the PR after the develop branch. Do you know how I could achieve the above? This was automatic on gerrit.

odanrc:
I am still unfamiliar with GitHub´s interface, so I don´t know how to request stacked PRs. @powerjg do you know how to do that?

In any case this PR is the first in the chain, so the changes made for the later PRs shouldn´t be a problem here.

pranith:
> I am still unfamiliar with GitHub´s interface, so I don´t know how to request stacked PRs. @powerjg do you know how to do that?
> 
> In any case this PR is the first in the chain, so the changes made for the later PRs shouldn´t be a problem here.

I think github has an interface to review each individual patch in a PR.

You need to click the hash of the first commit in the PR.

odanrc:
Alright, let´s give it a try. You can delete this PR, since it is now mostly a duplicate of #745 . I still strongly think that the tag shift change should be in its own PR, so it should remain like that.

PR #774, opened 2024-01-15
arch-vega: Fix upsize cast error in newer compilers

abmerop:
Newer compilers error on -Warray-length in the recent MI200 patches due to casting from a 32-bit data type to a 64-bit type. Change it to cast the 32-bit integer first then 64-bit integer latter to remove the warning.

Rerun of validation tests on the three instructions passed.

Change-Id: I0309e5f7b5b8cc8ce1651660ddddb120fa6e7666


abmerop:
(Mini-)compiler test found this. Seems to be the only issue

powerjg (approved):

mattsinc (approved):

PR #779, opened 2024-01-16
stdlib: Refactor stdlib such that `obtain_resources` can request multiple resources from databases at a time

### LINKED ISSUE #644: Enable bundling of `obtain_workload` requests in gem5 Resources stdlib component
At present the `obtain_resource` function is used to get Resources from the gem5 standard library but runs on a "single-request-per-resource" model to our clients (in almost all cases the MongoDB database where we store our gem5 Resources data).

This is very inefficient in cases where we need to get many resources. With the introduction of `Suite` resource objects we have a single `Suite` resource containing many `Workload` resources which, in turn` contain several resources. A "single-request-per-resource" model is very costly in this case.

#625 was added to fix this case but is _very hacky_ and not an ideal long term solution.

I suggest extended the functionality of the `obtain_resource` to allow the specification of 1..N resources and, if multiple are specified, these requests are bundled.

Harshil2107:
To-do list for this refactor
- [x] Update MongoDB query to request multiple resources at once
- [x] Update JSON query to handle multiple resources
- [x] Update client code to work with multiple resources instead on one 
- [x] Update handling of multiple datasources and make sure they work
- [x] Update error handling
- [x] refactor obtain_resources to request for multiple resources

Change-Id: Ibee1cdfe480f5fd9bb2695ab6d86c1b2d81f4bf3


src/python/gem5/resources/client_api/abstract_client.py
```diff
@@ -55,8 +55,7 @@ def _url_validator(self, url: str) -> bool:
     @abstractmethod
     def get_resources(
         self,
-        resource_id: Optional[str] = None,
-        resource_version: Optional[str] = None,
+        resource_info: List[Dict[str, str]],
         gem5_version: Optional[str] = None,
     ) -> List[Dict[str, Any]]:
         """
```
BobbyRBruce: So this isn't a bad start at but I have a couple of suggestions which i think could make this better:

1. Let's create a new function to handle "multiple id requests" (or whatever you want to call them), and leave the `get_resources` signature unchanged and backwards compatible). With the design as is, it'd be something like:

```py
   def get_resources(
        self,
        resource_id: Optional[str] = None,
        gem5_version: Optional[str] = None,
    ) -> List[Dict[str, Any]]:
        return get_multi_resources(
            resource_info = [ {resource_id, resource_version} ],
            gem5_version = gem5_version,
        )

   # I didn't think very long about this function name... consider it a placeholder.
   def get_multi_resources( 
        self,
        resource_info: List[Dict[str, str]],
        gem5_version: Optional[str] = None,
    ) -> List[Dict[str, Any]]:
        pass #TODO.
```

2. The original `get_resources` function didn't query on _just_ ID and version, as the function signature shows it also queries by `gem5_version`. This design fixes the `gem5_version` across all queries which, at this level of the API, is not a guarantee. We should support cases such as "I want Resource 'X', v2, for gem5 v24, and Resource 'Y', v3, for gem5 v23". So, for our multi-query function we should have something like:

```py
def get_multi_resources( 
        self,
        get_resources_query : List[Dict[str, Any]]
    ) -> List[Dict[str, Any]]:
        # pseudocode here to show how to access this data.
        for query in get_resources_query:
            print(f"Here i'm getting a query for resource {query['resource']} "
                  f"Version {query['version']}, for gem5 {query['gem5_version]}")
```

Here the `get_resources_query` parameter would essentially be a list of `**kwards` of the `get_resources` function so, we'd change `get_resources` to:

```py
   def get_resources(
        self,
        resource_id: Optional[str] = None,
        gem5_version: Optional[str] = None,
    ) -> List[Dict[str, Any]]:
        return get_multi_resources(
            get_resources_query = [
               {
                   "resource_id" : resource_id,
                   "resource_version: resource_version,
                   "gem5_version": gem5_version,
               }
            ],
        )
```

or I think the following would also work:

```py
   def get_resources(
        self,
        resource_id: Optional[str] = None,
        gem5_version: Optional[str] = None,
    ) -> List[Dict[str, Any]]:
        return get_multi_resources(
            get_resources_query = [
               locals()
            ],
        )
```

From anything using this API this casts the "multi id/version" scenario as a bundling of many `get_resource` calls. What is returned is the set of all resources from running each of these queries.

src/python/gem5/resources/client_api/client_wrapper.py
BobbyRBruce, line 1: I think the `ClientWrapper` class is a mistake and that it'd be good to take the opportunity to refactor it a way as part of this change. Its purpose is a bit confused. I believe it's original purpose was to have a single object which contained, setup and handled the set of clients, however, it's grown a bit unwieldily and seems to have a lot of code that could easily be part of just `gem5.resources.client`.

I don't claim to have thought this all the way through but looking at the code I think the functionality provided by `ClientWrapper` could be moved down into "gem5.resources.client".

As an example, ["gem5.resources.client"'s `get_resource_json_obj` function calls `get_resource_json_obj_from_client` from `ClientWrapper`](https://github.com/gem5/gem5/blob/bae34876780dfb2bc22b9151bfda1d39ee80cfb1/src/python/gem5/resources/client.py#L167). I think it'd be good to just flatten this and put the logic of `get_resource_json_object_from_client` inside the `get_resource_json_obj` function.

I think we could iron out and get rid of a lot of bad code her, as well as side-step the refactoring nightmare that they necessitate, if we did this.

src/python/gem5/resources/client.py
BobbyRBruce, line 1: **This comment will make more sense when you read my comment in the `AbstractClient` class. It's essentially the same argument bubbled up to "client.py"**.

I think we should add a function here to handle the special "multi get resource json" case and leave the current API unchanges. I.e., we expand but we don't modify the API.

BobbyRBruce (changes requested):
This comment is more just to clarify how i've come to think of this code having looked over it for the few time in many months. 

It seems to me the core API to access the client data is contained in "gem5.resoures.client", and it's just two functions. One is `list_resources` which is pretty basic and just returns all resources (I'm not even sure if this is used anywhere) and the other is `get_resource_json_obj` which is by far the most important. This is what's called from `obtain_resource` and returns the JSON object in which all the resources are ultimately build. Everything else that is called beyond that are "sub-APIs" which support this (some, which I've mentioned in my reviews could probably be removed to avoid unnecessary levels of abstraction we have here).


The key to how this all will work is deciding how we incorporate getting multiple resource JSON objects from "gem5.resources.client". From there it doesn't seem too difficult to filter it backwards through to the client (especially if we remove some of the abstractions that have crept in and confuse things). Once this is done I think it'll be really easy to "filter it forwards" to the `Resource` classes.


I think there are several ways to do this, but I've made some suggestions here that are my preference. I touch on this in my review comments but I think we should expand the API rather than modify it. We don't need to break the current API. Most resource requests don't need to be bundled, we only need bundling in exceptional cases and we should be able to flip between these options as we see fit when engineering the resource object code. I also think the API should make obtaining multiple resource json object look like bundling many small `get_resource_json_obj` queries.


**Most importantly**: The solution you came up with is good. I hope it doesn't seem like i'm criticizing it. It helped me understand the problem more clearly.

src/python/gem5/resources/client_api/abstract_client.py
Harshil2107: If we are just changing the main api in the client.py file to have one for both single and multiple resource, we can keep t he parameters and return type the same for single resource and change the other abstracted file to query for multiple resources as query for multiple resources will also get a single resource but as everything is abstracted we dont need to maintain two sets for JSON and MongoDB queries.

src/python/gem5/resources/client.py
```diff
@@ -163,7 +163,36 @@ def get_resource_json_obj(
                          current build. If ``None``, filtering based on compatibility
                          is not performed.
     """
+    # this will be removed when obtain resources is refactored
```
powerjg: Is this comment still valid?

src/python/gem5/resources/resource.py
```diff
@@ -722,16 +725,24 @@ class SuiteResource(AbstractResource):
 
     def __init__(
         self,
-        workloads: Dict["WorkloadResource", Set[str]] = {},
+        workloads: Dict[Tuple[str, str], Set[str]] = {},
```
powerjg: Why is the key not a `WorkloadResource` object? 

Wouldn't `WorkloadResource` also have the same information (id and version)?

src/python/gem5/resources/resource.py
```diff
@@ -722,16 +725,24 @@ class SuiteResource(AbstractResource):
...
         resource_version: Optional[str] = None,
         description: Optional[str] = None,
         source: Optional[str] = None,
         id: Optional[str] = None,
+        resource_directory: Optional[str] = None,
+        download_md5_mismatch: Optional[bool] = True,
+        clients: Optional[List] = None,
+        gem5_version: Optional[str] = None,
+        quiet: Optional[bool] = False,
+        local_path: Optional[str] = None,
```
powerjg: You added parameters. You need to add documentation for each of these.

src/python/gem5/resources/resource.py
```diff
@@ -753,13 +770,43 @@ def __init__(
             resource_version=resource_version,
         )
 
+    def _get_all_workload_json(self) -> List[Dict[str, str]]:
+        """
+        This function gets all the resources for the workload.
+
+        :returns: A list of all the resources for the workload.
```
powerjg: Can you define what is the key and what is the value in the dictionary return value?

src/python/gem5/resources/resource.py
```diff
@@ -753,13 +770,43 @@ def __init__(
...
+            # We create a WorkloadResource object for each workload
+            # only when it is requested.
+            yield WorkloadResource(
+                local_path=self._local_path,
+                resource_directory=self._resource_directory,
+                download_md5_mismatch=self._download_md5_mismatch,
+                clients=self._clients,
+                gem5_version=self._gem5_version,
+                quiet=self._quiet,
+                **workload,
```
powerjg: I don't understand this line of code. I guess I don't know what's in the `workload` dictionary.

This is clever, but maybe not readable. A comment could help, or maybe the code could be be more verbose.

src/python/gem5/resources/resource.py
```diff
@@ -865,6 +912,12 @@ def __init__(
         source: Optional[str] = None,
         local_path: Optional[str] = None,
         parameters: Optional[Dict[str, Any]] = {},
+        resources: Optional[Dict[str, Dict[str, str]]] = {},
```
powerjg: Whoa... please define the keys and the values here.

src/python/gem5/resources/resource.py
```diff
@@ -865,6 +912,12 @@ def __init__(
...
+        resource_directory: Optional[str] = None,
+        download_md5_mismatch: Optional[bool] = True,
+        clients: Optional[List] = None,
+        gem5_version: Optional[str] = None,
+        quiet: Optional[bool] = False,
```
powerjg: All of these need to have documentation.

src/python/gem5/resources/resource.py
```diff
@@ -884,6 +937,144 @@ def __init__(
...
+        clients: List[str],
+        gem5_version: str,
+        quiet: bool,
+    ) -> None:
+        """
+        This function gets all the resources for the workload.
+
+        :param resources: A dictionary of resources to get. The keys are the
+                            parameter names and the values are the resource
+                            id and version.
```
powerjg: Ok, this is better!

What is a "parameter name" and why is the resource id/version a dictionary? Why is the id a key and the version a value?

An example would go a long way here.

src/python/gem5/resources/resource.py
```diff
@@ -884,6 +937,144 @@ def __init__(
...
+        :param to_path: The path to which the resource is to be downloaded. If
+                        ``None``, the resource will be downloaded to the resource
+                        directory with the file/directory name equal to the ID of
+                        the resource.
+        :param resource_directory: The location of the directory in which the
+                                   resource is to be stored. If this parameter is
+                                   not set, it will set to the environment variable
+                                   ``GEM5_RESOURCE_DIR``. If the environment is not
+                                   set it will default to ``~/.cache/gem5`` if
```
powerjg: Is this the only place this default is specified? If not, then there should be a more general function for this path.

src/python/gem5/resources/resource.py
```diff
@@ -884,6 +937,144 @@ def __init__(
...
+                                   available, otherwise the CWD. *Note*: This
+                                   argument is ignored if the ``to_path`` parameter
+                                   is specified.
+        :param download_md5_mismatch: If the resource is present, but does not
+                                      have the correct md5 value, the resource will
+                                      be deleted and re-downloaded if this value is
+                                      ``True``. Otherwise an exception will be thrown.
+                                      ``True`` by default.
```
powerjg: This is not "true" ;). The parameter does not have a default.

src/python/gem5/resources/resource.py
```diff
@@ -884,6 +937,144 @@ def __init__(
...
+        :param clients: A list of clients to search for the resource. If this
```
powerjg: What is a client?

src/python/gem5/resources/client_api/abstract_client.py
powerjg: Why did you not take Bobby's suggestion here?

powerjg (commented):
A few comments :)

powerjg (commented):
A few comments :)

src/python/gem5/resources/client.py
```diff
@@ -163,7 +163,36 @@ def get_resource_json_obj(
...
 
+def get_multiple_resource_json_obj(
+    resource_info: List[Dict[str, str]],
+    clients: Optional[List[str]] = None,
+    gem5_version: Optional[str] = core.gem5Version,
+) -> List[Dict]:
+    """
+    Get the resource json object from the clients wrapper.
+
+    :param resource_info: The resource info.
```
BobbyRBruce: Definitely explain here the  schema this List of dictionaries need to be abide by.

src/python/gem5/resources/client.py
```diff
@@ -163,7 +163,36 @@ def get_resource_json_obj(
...
+    # resources
+    return _get_clientwrapper().get_resource_json_obj_from_client(
+        resource_info, clients, gem5_version
+    )[0]
+
 
+def get_multiple_resource_json_obj(
+    resource_info: List[Dict[str, str]],
+    clients: Optional[List[str]] = None,
+    gem5_version: Optional[str] = core.gem5Version,
```
BobbyRBruce: It won't be my hill to die on but could this not be part of `resource_info` be part of `resource_info`? I know behind-the-scenes `gem5_version` is used in a post-processing stop to filter resources, to get the most compatible version, but from the point-of-view of the API this looks like a query to me and the gem5_version we want the resource to work with is part of that query.

src/python/gem5/resources/client_api/client_wrapper.py
BobbyRBruce, line 1: I've said this before but is there any way we can take this opportunity to refactor away "client_wrapper.py" and maybe even "cient.py"? Can't these functions be moved into `AbtractClient` or "resource.py"?

I find them confusing and really appear to only be helper classes. I see things working this way ideally:

1. **resources.py**: Contains the user-facing API.
2. **client_api**: Contains `AbstractClient` and its subclasses . This `AbstractClient` defines the API used by "resources.py" to get the resources it needs from the various clients. This is a softer API as it's not user-facing. It's fine to change but does mandate updating the functions in "resources.py". The API exists so we can add new clients in the future by simply inheriting from `AbstractClient` and implementing all the abstract functions.

The logic to me seems simple: The clients are instantiated (whatever they may be `JsonClient`, `AtlasClient`: many or just one) and these are used by the functions in "resource.py" to get the resources. Everything else just feels a bit supurfluous and confuses the system.

src/python/gem5/resources/resource.py
```diff
@@ -884,6 +937,144 @@ def __init__(
...
+            resource_directory=resource_directory,
+            download_md5_mismatch=download_md5_mismatch,
+            clients=clients,
+            gem5_version=gem5_version,
+            quiet=quiet,
+        )
+
+    def get_all_resources(
+        self,
+        resources: Dict[str, Dict[str, str]],
```
BobbyRBruce: I don't understand this. You're saying when the Workload is instantiated you get all its constituent resources (downloads and all). Isn't reintroducing a problem here?  Getting everything immediately?

A suite contains a set of Workload objects. The problem we had was this meant we needed to get all the resources (downloads and all) in the suite to just have a functioning suite object. This was of course incredibly inefficient when you only wanted to use one of these workloads.

The solution I implemented had the Workload objects constructed with a partial function which were essentially defining the function call needed to download all its components in our bucket. This partial function was only called when someone tried to accesses these resources via `get_local_path`. Ergo, you had a Suite of workloads, you could iterate through the workloads and inspect ids and versions as you wanted to, and only when you called `get_local_path` did it download what it needed for that workload.

I appreciate this solution is imperfect, but it saves us the bandwidth. 

While your solution in this PR solves pinging the database to death, does this not reintroduce the other bandwidth problem we were facing?

src/python/gem5/resources/resource.py
BobbyRBruce: I believe the thinking there is the entry for the Suite in the database only stores the resource_info in a simple JSON form. something like:

```json
workloads: { id: "hello-world", version: "v2.0.1"}
```

By having the signature of the `Suite`'s constructer mirror that of the databse enter you can pass the entry directly to the constructor.

At present we have some special code for suites to inspect the database entry `Workload` objects constructed the Workload objects, then pass it to the constructor. To me this makes logical sense but I can see why it's a pain to translate the database entry to the object.


My argument for keeping this as an object is as follows: This design assumes construction of a suite solely through the `obtain_resource` function. However, i think this should be allowed (and frankly, encouraged in some circumstances):

```py
workload_1 = Workload(...)
workload_2 = Workload(...)

my_suite = Suite({workload_1, workload_2})
```

For suites i'll also saw that allowing that above is conceptually much simpler than something like:

```py
workload_1 = Workload(id=1, version=2)
... # Some logic here to load `workload_1` into a suite

my_suite = Suite({id=1, version=2 ...})
```

I don't personally envision users being happy with a design that mandates they made an entry in a JSON file and load that JSON file and use `obtain_resources` to construct it.

What I see the `obtain_resource` function as doing (and the "clients" infastructure it depends on) is handle the translation between sources and the resource objects.

The trade off here is between having a constructor that represents our schema, or a constructor that an end user can use intuitively. The former has benefits. For example, it means if we update the schema we have less code that needs updated. However, I think the value in intuitive constructors that work in more scenarios is important.

BobbyRBruce (commented):
As I stated in a meeting last week: I'm taking a backseat to managing the development of gem5 resources instrastructure. So I only review this as I would any other PR and i will respect Harshill's design decisions assuming they do what he intended and pass our usual checks and inspections. In short, I'll no longer push for "my vision": I simply don't have the bandwidth to do so and I was blocking people by not making my mind up about stuff. Also, I think it's right that Harshel has "ownership" over this code and it's future development given he's not the most familiar with it.

Ergo my comments here can be read in a "are you sure there isn't a  better way?" tone. I won't fight too hard as long as nothing jumps out at me and everything else in the project continues to work with the changes

src/python/gem5/resources/resource.py
Harshil2107: This function also created the same partial as passes it to the constructor of the resource, so it will only download the required resource when `get_local_path` is called.

I am going to refactor the Workload and Suite resource so that we dont have to pass all the `obtain_resource` arguments like clients, resource_directory, etc to the constructors.

src/python/gem5/resources/resource.py
Harshil2107: This is an exact copy of the field in JSON. In JSON we have a resoruces field which looks like this:
```
resources: {
    "binary":{
        "id":"id",
        "resource_version":"1.0.0"
    },
    "readin_file":{
        "id":"id2",
        "resource_version":"1.0.0"
    },
}
```

so the key is the name of the paramter that is used in the set workload functions and the value is the resource that we have to pass it.

I am going to refactor this so that I dont have to pass this many variables around into the worklaod and directly just give it the objects.

src/python/gem5/resources/resource.py
Harshil2107: Client is basically our list of datasoruces. By default it is only our MongoDB but if someone adds another MongoDB or JSON as a datasource, then they can specify a subset of these clients to search resources from. By default we go through all.

src/python/gem5/resources/resource.py
Harshil2107: I changed it back so that we make all the workload resource objects so now the workloads_obj is Dict[WorkloadResource, Set[str]] like it used to be before. I am jsut getting all the workloads with a single call to the database and parsing that to create thins dictionary. This also only downloads the smaller resoruces(binaries, etc) only when required that is when `get_local_path` is called. 

The only thing to note with this solution is that I have the specific logic for parsing dependent workloads outside the workload class but I think this is better than passing a lot of unecessay variables in the SuitResource cinstructor.

Harshil2107:
I have moved the code that fetched all the dependent resources for both workloads and suites outside the class. This solves the issue of passing all the parameters to the class constructors.

I have also removed the ShadowResource class as it is not used anymore.

The SuiteResource class now directly takes in WorkloadResource objects. Even though these objects are created the actual resource files are only downloaded when required that is when the `get_local_path` variable is called.

src/python/gem5/resources/client_api/client_wrapper.py
```diff
@@ -102,56 +103,72 @@ def list_resources(
...
+        resource_info: str,
         clients: Optional[List[str]] = None,
-    ) -> List[Dict]:
+    ) -> Dict[str, Any]:
         """
         This function returns all the resources with the given id from all the
         sources.
 
-        :param resource_id: The id of the resource to search for.
+        :param resource_info: List of dictionaries containing information about the
```
powerjg: This doesn't match the type above.

src/python/gem5/resources/client_api/jsonclient.py
```diff
@@ -75,25 +75,40 @@ def get_resources_json(self) -> List[Dict[str, Any]]:
...
-                    resource
-                    for resource in filter
-                    if resource["resource_version"] == resource_version
-                ]
+    ) -> Dict[str, Any]:
+        filtered_resources = []  # Unfiltered.
+
+        for resource in resource_info:
+            if "resource_version" in resource.keys():
+                filtered_resources.append(
```
powerjg: I think using the `filter` function would be more readable here.

src/python/gem5/resources/client_api/abstract_client.py
Harshil2107: I just kept it as one function as this function is not user facing so it will be called by our sub apis which users can not interact with.

src/python/gem5/resources/client_api/abstract_client.py
```diff
@@ -72,6 +71,40 @@ def get_resources(
...
         raise NotImplementedError
 
+    def sort_resources(resources: List) -> List:
```
BobbyRBruce: This could have better typing. I don't know if this is a List of `Resource` or some dictionary structure. If it's a dictionary then lay out the types expected and expand the description to be clear about what is expected here.

src/python/gem5/resources/client_api/abstract_client.py
```diff
@@ -111,10 +144,12 @@ def filter_incompatible_resources(
...
 
-    def get_resources_by_id(self, resource_id: str) -> List[Dict[str, Any]]:
+    def get_resources_by_id(
+        self, resource_info: List[Dict[str, str]]
+    ) -> List[Dict[str, Any]]:
         """
         :param resource_id: The ID of the Resource.
 
         :return: A list of all the Resources with the given ID.
         """
```
BobbyRBruce: This documentation needs updated. We need to know the format of `resource_info` via this documentation. it's a list of dictionaries mapping strings to strings, but what are these strings? What's the schema of this?

src/python/gem5/resources/resource.py
```diff
@@ -730,8 +733,10 @@ def __init__(
         **kwargs,
     ) -> None:
         """
-        :param workloads: A list of ``WorkloadResource`` objects
-                          created from the ``_workloads`` parameter.
+        :param workloads: A Dict of Tuples containing the workload ID and
+                          version as the key and a set of input groups as the
+                          value. This Dict is created from the ``_workloads``
+                          parameter.
```
BobbyRBruce: The typing in the function suggests this is a dictionary mapping WorkloadResources to strings.

src/python/gem5/resources/client.py
```diff
@@ -163,7 +169,218 @@ def get_resource_json_obj(
...
+    else:
+        resource_info = [{"id": resource_id, "gem5_version": gem5_version}]
+
+    # We will return a list when we refactor ontain_resources to handle multiple
+    # resources
+    return _get_resource_json_obj_from_client(resource_info, clients)[0]
+
+
+def get_multiple_resource_json_obj(
+    resource_info: List[Dict[str, str]],
```
BobbyRBruce: I don't like how much we're passing around `Dict[str,str]`. Why don't we create a class to hold the Query information?:

```py
class ClientQuery:
    def __init__(self, resource_id: str, version: str, gem5_version: Optional[str] = None):
       self._id = resource_id
       self._version = version
       self._gem5_version = gem5_version

    def get_id(self) -> str
        return self._id

    def get_version(self) -> str
        return self._version

    def get_gem5_version(self) -> Optional[str]
        return self._gem5_version

    def sortable_tuple(self) -> Tuple:
        to_return = (self.get_id().lower(),)
        for val in self.get_version().split("."):
            if val.isdigit():
                to_return += (int(val),)
            else:
                to_return += (str(val).lower(),)
        return to_return
```

BobbyRBruce (changes requested):

src/python/gem5/resources/client.py
Harshil2107: If we are creating this class, cant we just make the class encompass the query itself.
```
class ClientQuery:
    def __inti__(self, query_dicts: List[Dict[str,str]]):
        self.query = query_dicts

    def add_resource_to_query(self, resource_id:str, gem5_version:str,  resoruce_version:str):
        self.query.append(
            "id":resource_id, "gem5_version": gem5_version, "resource_version":resoruce_version)
```

Essentially just encapsulate the list into a class, we can also have a function that returns the query in MongoDB format if we want as well.

src/python/gem5/resources/client_api/client_query.py
```diff
@@ -0,0 +1,22 @@
+from typing import Optional
```
powerjg, line 27: This file has to have a copyright statement.

src/python/gem5/resources/client_api/abstract_client.py
```diff
@@ -55,13 +58,15 @@ def _url_validator(self, url: str) -> bool:
...
     ) -> List[Dict[str, Any]]:
         """
-        :param resource_id: The ID of the Resource. Optional, if not set, all
-                            resources will be returned.
+        :param resource_info: A list of dictionaries containing the
+                                information to query the resources. Each
+                                ClientQuery object can contain the following keys:
+                                - resource_id: The ID of the Resource.
+                                - resource_version: The version of the `Resource`.
+                                - gem5_version: The version of gem5.
```
BobbyRBruce: This needs updated. The parameter is a List of Client queries which specify the query data needed to get the resources.

src/python/gem5/resources/client_api/abstract_client.py
```diff
@@ -55,13 +58,15 @@ def _url_validator(self, url: str) -> bool:
     @abstractmethod
     def get_resources(
         self,
-        resource_id: Optional[str] = None,
-        resource_version: Optional[str] = None,
-        gem5_version: Optional[str] = None,
+        resource_info: List[ClientQuery],
```
BobbyRBruce: I really like this `ClientQuery` improvement. My nit-pic here is, can we rename `resource_info` to `client_queries`?

This commend applies to other occurrences of this in this change.

src/python/gem5/resources/client_api/abstract_client.py
```diff
@@ -111,10 +150,17 @@ def filter_incompatible_resources(
...
+        self, resource_info: List[ClientQuery]
+    ) -> List[Dict[str, Any]]:
         """
-        :param resource_id: The ID of the Resource.
+        :param resource_info: A list of ClientQuery objects containing the
+                            information to query the resources. Each
+                            ClientQuery object can contain the following keys:
+                            - resource_id: The ID of the Resource.
+                            - resource_version: The version of the `Resource`.
+                            - gem5_version: The version of gem5.
```
BobbyRBruce: Ditto, this documentation needs updated.

src/python/gem5/resources/client_api/client_query.py
```diff
@@ -0,0 +1,48 @@
...
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from typing import Optional
+
```
BobbyRBruce, line 28: Can you add a docstring here just explaining what this class is for. I.e., that encapsulates a query to be processed by our clients to get a resource.

src/python/gem5/resources/client_api/jsonclient.py
```diff
@@ -75,25 +76,49 @@ def get_resources_json(self) -> List[Dict[str, Any]]:
...
+        def filter_resource(resource, resource_info):
+            for resource_query in resource_info:
+                gem5_version_match = False
+                resource_version_match = False
+
+                if (
+                    resource_query.get_gem5_version() is not None
+                    and not resource_query["gem5_version"].startswith(
+                        "DEVELOP"
+                    )
```
BobbyRBruce: This hasn't been updated to use the `ClientQuery`.

src/python/gem5/resources/client_api/abstract_client.py
```diff
@@ -72,6 +77,40 @@ def get_resources(
...
+                else:
+                    to_return += (str(val).lower(),)
+            return to_return
+
+        return sorted(
+            resources,
+            key=lambda resource: sort_tuple(resource),
+            reverse=True,
+        )
+
```
BobbyRBruce, line 113: I won't ask for a change in this PR as I don't think there's an easy elegant solution here, but I'm not a fan of this function as it requires the construction of this dictionary in accordance to this arbitrary schema we don't properly define anywhere (i.e., the dictionary contains two keys: `id`, `resources_version`).

Though, just more-or-less a comment as "food-for-thought". 

I think we've got into the habit of passing around dictionaries way too much when we should really be passing objects.

src/python/gem5/resources/resource.py
```diff
@@ -972,6 +938,233 @@ def obtain_resource(
...
+    # Once we know what AbstractResource subclass we are using, we create it.
+    # The fields in the JSON object are assumed to map like-for-like to the
+    # subclass contructor, so we can pass the resource_json map directly.
+    return resource_class(
+        local_path=to_path, downloader=downloader, **resource_json
+    )
+
+
+def _get_suite(
+    suite: Dict[str, str],
```
BobbyRBruce: This isn't a Dictionary of strings to strings.

src/python/gem5/resources/resource.py
```diff
@@ -972,6 +938,233 @@ def obtain_resource(
...
+    suite["workloads"] = workload_input_group_dict
+    return SuiteResource(
+        local_path=local_path,
+        downloader=None,
+        **suite,
+    )
+
+
+def _get_workload(
+    workload: Dict[str, str],
```
BobbyRBruce: This isn't a Dictionary of strings to strings.

src/python/gem5/resources/resource.py
```diff
@@ -972,6 +938,233 @@ def obtain_resource(
...
+                         version.
+    :param quiet: If ``True``, suppress output. ``False`` by default.
+    """
+    params = {}
+    resources = workload["resources"]
+    # Mapping IDs to parameter names
+    id_to_param_map = {
+        resource_info["id"]: param
+        for param, resource_info in resources.items()
+    }
```
BobbyRBruce: You're going to need to rethink this. `resources` maps the `set_workload` function's parameters to resources. This is fine as a function can't have a two parameters of the same name. However, you're creating a map here which maps resources IDs to parameter names. What if a `set_workload` function class two parameters set to the same resource?

In addition, what about versioning here? This information is not used here. A resource id is not enough to specify which version you're wanting.

src/python/gem5/resources/resource.py
```diff
@@ -972,6 +938,233 @@ def obtain_resource(
...
+    # Fetching resources as a list of dicts
+    resource_details_list = get_multiple_resource_json_obj(db_query, clients)
+
+    # Preparing the final mapping of parameters to their complete JSON objects
+    param_to_resource_json = {}
+    for resource in resource_details_list:
+        resource_id = resource["id"]
+        if resource_id in id_to_param_map:
+            param = id_to_param_map[resource_id]
+            param_to_resource_json[param] = resource
```
BobbyRBruce: So, why not `workload["resources"]`
```suggestion
    for param_name, param_resource in workload["resources"].items():
    
        # This could probably be a fancy one-liner thing but I can't be bothered figuring it out.
        resource_match = None
        for resource in resource_details_list:
            if param_resource["id"] == resource["id"] && param_resource["version"] == resource["version"]:
                resource_match = resource
                break
        if not resource_match:
             raise Exception("error...")
        param_to_resource_json[param_name]=resource_match
```

BobbyRBruce: I think could also merge this into the for loop below and just have one for loop constructing these resources.

BobbyRBruce (changes requested):

src/python/gem5/resources/client_api/abstract_client.py
```diff
@@ -55,13 +58,15 @@ def _url_validator(self, url: str) -> bool:
...
     ) -> List[Dict[str, Any]]:
         """
-        :param resource_id: The ID of the Resource. Optional, if not set, all
-                            resources will be returned.
+        :param client_queries: A list of client queries containing the
+                                information to query the resources. Each
+                                ClientQuery object can contain the following keys:
+                                - resource_id: The ID of the Resource.
+                                - resource_version: The version of the `Resource`.
+                                - gem5_version: The version of gem5.
```
BobbyRBruce: The client queries don't contain any keys. The definition of the `ClientQuery` is the class itself.

src/python/gem5/resources/client_api/abstract_client.py
```diff
@@ -111,10 +150,17 @@ def filter_incompatible_resources(
...
+        self, client_queries: List[ClientQuery]
+    ) -> List[Dict[str, Any]]:
         """
-        :param resource_id: The ID of the Resource.
+        :param client_queries: A list of ClientQuery objects containing the
+                            information to query the resources. Each
+                            ClientQuery object can contain the following keys:
+                            - resource_id: The ID of the Resource.
+                            - resource_version: The version of the `Resource`.
+                            - gem5_version: The version of gem5.
```
BobbyRBruce: Ditto

src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -155,24 +156,39 @@ def _atlas_http_json_req(
...
 
-        if filter:
-            data["filter"] = filter
+        search_conditions = []
+        for resource in client_queries:
+            condition = {
+                "id": resource.get_resource_id(),
+            }
+
+            if not resource.get_gem5_version().startswith("DEVELOP"):
```
BobbyRBruce, line 175: It'd be neater here to add a function in `ClientQuery` like `is_develop_version(self) -> bool`. First of all we do this comparison a few times, so best to have the logic in one place. Secondly, whether you're querying for develop or an official version, is a property of the query so the logic should be put there.

src/python/gem5/resources/client_api/client_query.py
BobbyRBruce: @Harshil2107 , you resolved this conversation without addressing the comment.

src/python/gem5/resources/resource.py
BobbyRBruce: This comment was marked as resolved but not addressed. The typing here, `workloads: Dict["WorkloadResource", Set[str]] = {},`, does not match the what the docstring is saying.

src/python/gem5/resources/client_api/atlasclient.py
BobbyRBruce: Also, maybe I'm wrong but doesn't this break if `get_gem5_version()` returns `None`?

src/python/gem5/resources/resource.py
```diff
@@ -972,6 +938,234 @@ def obtain_resource(
...
+        # Adding the additional parameters to the workload parameters
+        if workload["additional_params"]:
+            for key in workload["additional_params"].keys():
+                assert isinstance(key, str)
+                value = workload["additional_params"][key]
+                params[key] = value
+
+    return WorkloadResource(
+        local_path=local_path,
+        downloader=None,
```
BobbyRBruce, line 1151: Are we not passing the downloader?

BobbyRBruce (changes requested):

src/python/gem5/resources/resource.py
Harshil2107: No, as the workload resource itself has no files to download, it just has resources that we need to download and those will be downloaded when we create those specific resource objects.

BobbyRBruce: Ack.

src/python/gem5/resources/client_api/atlasclient.py
Harshil2107: `gem5_version` should never be `None` as by default it is the current version. I will change the default in client query class to point to the current gem5 version.

BobbyRBruce (approved):

Issue #703, opened 2023-12-20
Deprecate GCN3 and move exclusively to Vega / gfx9


BobbyRBruce:
### Discussed in https://github.com/orgs/gem5/discussions/306

<div type='discussions-op-text'>

<sup>Originally posted by **abmerop** September 12, 2023</sup>
I'd like to propose deprecating the GCN3 ISA and moving exclusively to Vega.  Vega is essentially a superset ISA of GCN3, meaning it contains everything GCN3 has and more.  As a result of this, there is an enormous amount of redundant code. gem5 also now supports an APU model with Vega as well as a dGPU so users can simulate either one, just like GCN3. 

The plan would be as follows:
- Update all GPU gem5-resources Makefiles to build the gfx902 target (Vega APU).  gfx900 (dGPU) is already being built.
- Ensure that all GPU gem5-resources tests work as well or better with Vega compared to GCN3.
- Update documentation for removal references to GCN3.
- Change the default --gfx-version to gfx902 or gfx900.
   - I think if we choose gfx900, we should also implicitly set the --dgpu flag as this is no gfx900 APU.
- Delete GCN3 related files.

The benefits would be:
- Removes 90k lines of code from the gem5 source code.
- Less applications to test and one less build target to test.
- A single GPU ISA means GPU can be built into build/ALL/gem5.* further reducing testing (I think - Ruby protocol might be an issue).
- gfx9 is the current ISA used in Radeon™ Instinct™ products up to and including the most recent hardware - So we would be simulating *current* products.

Will add more to plan/benefits if I think of anything else, but wanted to get the discussion started.</div>

abmerop:
I was talking to @mattsinc about this and i think the hang up was DNNMark. However, we discussed that it is working on gfx902 so I think this plan is a go.

mattsinc:
This is fine with me, if @abmerop confirmed that.

abmerop:
Alright, I will put together a PR soon

abmerop:
Added PR #781 .  As far as the list of items goes (I am not sure how to make one of those checklists):  gem5-resources have been updated to build gfx902.  We have confirmed everything is working on gfx9 that was working on gfx8.  In that PR, I change the default gfx version to gfx902 and remove the files.  If we all agree on that last point the last step is to update documentations.

powerjg:
```
- [ ] This is how you make a checkbox
- [ ]
```

abmerop:
Believe this is done

PR #781, opened 2024-01-17
arch-gcn3: Remove gcn3

abmerop:
Related to issue #703 , this PR removes GCN3 related files and updates source code, documentation, and tests to switch over to Vega is that was not done already. Highlights are:

 - Remove all src/arch/amdgpu/gcn3 files and update Kconfigs.
 - Remove references to GCN3 and replace with Vega where applicable.
 - Update the build targets in the gcn-gpu Docker. This will need to be rebuilt but not urgently.
 - Remove the GCN3 tag in testlib. Most tests seem to be using Vega already, so that commit is small.


abmerop:
I believe I got everything that was conditioned off by GCN3 in the source.  I grep'd for GCN3, gfx8, carrizo, and fiji essentially.  MattS:  Let me know if you can think of some other things that could be removed.

Bobby: Let me know if the testlib changes are going to break anything.

BobbyRBruce:
> I believe I got everything that was conditioned off by GCN3 in the source. I grep'd for GCN3, gfx8, carrizo, and fiji essentially. MattS: Let me know if you can think of some other things that could be removed.
> 
> Bobby: Let me know if the testlib changes are going to break anything.

I don't believe this breaks anything. Given the CI tests have passed part-way proves it. So it's a thumbs-up from me.

BobbyRBruce (approved):

configs/example/apu_se.py
```diff
@@ -375,7 +375,7 @@
 parser.add_argument(
     "--gfx-version",
     type=str,
-    default="gfx801",
+    default="gfx902",
```
mattsinc, line 378: Should 900 or 902 be default?  Probably 900?

mattsinc (approved):

mattsinc:
We will need a corresponding commit in gem5-resources too, right?

configs/example/apu_se.py
abmerop: I chose 902 as it's also APU and to avoid needing to change the --dgpu option.  I think for 900 --dgpu would need to be set by default and there would need to be a new option to specify APU

abmerop:
> We will need a corresponding commit in gem5-resources too, right?

Yea.  I started doing this but there are a ton of READMEs that are very redundant.  Someone else mentioned this before too.  I am not sure if we should start with that work first.  I could replace all of the GCN3 instances with VEGA and that should be a decent start though.

configs/example/apu_se.py
mattsinc: Sure, fine.

mattsinc:
> > We will need a corresponding commit in gem5-resources too, right?
> 
> Yea. I started doing this but there are a ton of READMEs that are very redundant. Someone else mentioned this before too. I am not sure if we should start with that work first. I could replace all of the GCN3 instances with VEGA and that should be a decent start though.

Yeah, Bobby mentioned that.  The tricky thing is expectations -- if users are expecting each application to have its own README, then it leads to the duplication.  If instead we can reorg the website to have a single README for an entire benchmark suite and they fill in the appropriate inputs, then it would be less duplication but maybe less clear of a website org...

PR #732, opened 2024-01-03
arch-arm: Handle invalid case for encodeAArch64SysReg

giactra:
This patch is amending encodeAArch64SysReg so that it covers the case where there are no arch numbers available for the misc index passed as an argument.

This could happen if the register ID is a gem5 pseudo register which is not associated with any architected op1/op2/crn/crm tuple.

Rather than panicking we return a nullopt.

Change-Id: I7ab70467105ef93c0c78ac4e999c7dc8e5e09925

PR #784, opened 2024-01-18
arch-arm: Fix compile error in kvm

powerjg:
The addition of std::optional in #732 caused a compile error. This change fixes the error by checking to see if the value is present and panicing otherwise.

Change-Id: I46c3fb76eb0e14ba7bede7c336293fbe9add8c84


giactra (approved):
Ops, thanks for fixing this!

PR #786, opened 2024-01-18
base: Fix Integer overflow in AddrRange

### LINKED ISSUE #240: Integer overflow in AddrRange subset check
**Describe the bug**
If AddrRange `A` ends at the last byte of a 64-bit address space, it will be treated as a subset of any AddrRange `B` beginning at address 0. This is because `A._end` rolls over to zero, and so the check `A._start >= B._start && A._end <= B._end` becomes true.

**Affects version**
23.0.1.0

**gem5 Modifications**
I am working on modeling a novel memory device that requires specific allocation of the address space (which is why I encountered this error). But I have not made any changes to the `AddrRange` code, and the reproducible example below does not involve my device.

**To Reproduce**
Run this function and check the result:
```
bool checkAddrRangeBug() {
    AddrRange first_four_bytes = RangeSize(0x0, 4);
    AddrRange last_four_bytes = RangeSize(0xfffffffffffffffc, 4);
    return last_four_bytes.isSubset(first_four_bytes);
}
```
It returns `true` for me.

**Expected behavior**
The two ranges should be treated as distinct.

**Host Operating System**
Ubuntu 22.04 (kernel 5.15.90-wsl)

**Host ISA**
x86-64

**Compiler used**
gcc 11.3.0

Harshil2107:
This PR fixes the bug mentioned in #240.

Change-Id: I32139fa37b741185ba766d49dfa8cdab34f19010

### DIFF
--- a/src/base/addr_range.test.cc
+++ b/src/base/addr_range.test.cc
@@ -1544,6 +1544,19 @@ TEST(AddrRangeTest, SubtractionAssignmentOfRangeListFromRangeList)
                 expected_range3, expected_range4));
 }
 
+TEST(AddrRangeTest, isNotSubsetLastByte)
+{
+    /* An issue raised in https://github.com/gem5/gem5/issues/240 where if an
+     * address range ends at the last byte of a 64 bit address space, it will
+     * be considered a subset of any other address range that starts at the
+     * first byte of the range.
+     */
+
+    AddrRange first_four_bytes = RangeSize(0x0, 4);
+    AddrRange last_four_bytes = RangeSize(0xfffffffffffffffc, 4);
+
+    EXPECT_FALSE(last_four_bytes.isSubset(first_four_bytes));
+}
 /*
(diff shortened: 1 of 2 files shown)


src/base/addr_range.test.cc
```diff
@@ -1544,6 +1544,19 @@ TEST(AddrRangeTest, SubtractionAssignmentOfRangeListFromRangeList)
                 expected_range3, expected_range4));
 }
 
+TEST(AddrRangeTest, isNotSubsetLastByte)
+{
+    /* An issue raised in https://github.com/gem5/gem5/issues/240 where if an
```
pranith, line 1549: This comment can be copied into the commit message to make the motivation clearer.

pranith (changes requested):

pranith (approved):

powerjg (approved):

PR #789, opened 2024-01-19
arch-vega: Reorganize inst and misc files

abmerop:
This PR reorganizes the instructions.cc into multiple files and renames some files which do not match their corresponding header file names. The intention is to make iterating on development of these files faster.


mattsinc (approved):
I sort of liked having everything in one file, but if you think this is cleaner, fine with me.

abmerop:
> I sort of liked having everything in one file, but if you think this is cleaner, fine with me.

My main reasoning is not really that it's cleaner.  The description of the first commit basically:

"The Vega instructions.cc file is 47k lines long which results in both
large compilation times whenever it is modified and long style check
times. This makes iterating over more complex instruction
implementations very time consuming.

This commit moves the instruction definitions to multiple files based on
the instruction encoding (SOP2, VOP2, FLAT, DS, etc.). The resulting
files are much smaller (max is 8k lines) and compilation and style check
times are much more reasonable."

After a while it gets very annoying waiting several minutes for one translation unit to compile and link.  For complex instructions like MFMA that aren't straightforward to "get right" on the first try it becomes very painful.

mattsinc:
> > I sort of liked having everything in one file, but if you think this is cleaner, fine with me.
> 
> My main reasoning is not really that it's cleaner. The description of the first commit basically:
> 
> "The Vega instructions.cc file is 47k lines long which results in both large compilation times whenever it is modified and long style check times. This makes iterating over more complex instruction implementations very time consuming.
> 
> This commit moves the instruction definitions to multiple files based on the instruction encoding (SOP2, VOP2, FLAT, DS, etc.). The resulting files are much smaller (max is 8k lines) and compilation and style check times are much more reasonable."
> 
> After a while it gets very annoying waiting several minutes for one translation unit to compile and link. For complex instructions like MFMA that aren't straightforward to "get right" on the first try it becomes very painful.

I'm assuming you mean that only the files you change are recompiled?  If so, then this is fine.

PR #791, opened 2024-01-21
mem-ruby: Fix possible dirty line loss in CHI when ReadShared hit on UD line

jjuninho:
In case ReadShared hit on a UD line and there's no sharers, this chage makes the downstream passes Dirty to the requestor whenever possible even though it doesn't deallocate the line. This will make the requestor to SD and the downstream to UD_RSD.
In the previous implementation, loosely exclusive intermediate cache can cause loss of dirty data. Example error condition is as below.
   
Configurations
L2 cache: Roughly inclusive to L1 without back-invalidation
- dealloc_on_* = false
- dealloc_backinv_* = false
L3 cache: Roughly exclusive to L2 without back-invalidation
- alloc_on_readshared = tue
- alloc_on_readunique = false
- dealloc_on_shared = false
- dealloc_on_unique = true
- dealloc_backinv_* = false
- is_HN = false
LLC: Same clusivity as L3 except is_HN = true
For all caches, allow_SD = true and fwd_unique_on_readshared = false
    
Example problem sequence:
1. L1 sends ReadUnique then becomes UD. L2 is UC_RU. L3 and LLC are RU.
2. L1 evicts the line to L2 by WriteBackFull (UD_PD). L2 becomes UD.
3. L2 evicts the line to L3 using WriteBackFull (UD_PD). L3 becomes UD.
4. L1 reads the line with ReadShared which misses on L2.
5. L2 reads the line with ReadShared which hits on L3. L3 becomes UD_RSC because it doesn't deallocate the line (dataToBeInvalid=false)
6. L3 evicts the line to LLC by WriteCleanFull (UD_PD) because L3 doesn't back-invalidate and still has sharer. The local cache line is invalidated by Deallocate_CacheBlock. L3 becomes RUSC and LLC becomes UD_RU.
7. When UD_RU is evicted at LLC, the UD_RU line is dropped expecting the upstream to writeback, causing loss of dirty data

### DIFF
--- a/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
+++ b/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
@@ -2600,11 +2600,18 @@ action(Send_CompData, desc="") {
   bool is_rd_nsd := tbe.reqType == CHIRequestType:ReadNotSharedDirty;
   bool is_rd_unique := tbe.reqType == CHIRequestType:ReadUnique;
 
-  // if the config allows (or not caching the data) and line has no sharers
-  bool snd_unique_on_rs := (fwd_unique_on_readshared || tbe.dataToBeInvalid)
+  // if the request type allows SD
+  bool snd_dirty_on_rs := is_rd_shared && !is_rd_nsd;
+
+  // Send UC/UD on ReadShared or ReadNotSharedDirty if the line has no sharers
+  // and one of the followings are met
+  // 1) the config allows or
+  // 2) local cache won't have the line or
+  // 3) dirty will be passed
+  bool snd_unique_on_rs := (fwd_unique_on_readshared ||
+                            tbe.dataToBeInvalid ||
+                            snd_dirty_on_rs)
                           && tbe.dataUnique && tbe.dir_sharers.isEmpty();
-  // if the request type allows and we won't be caching the data
-  bool snd_dirty_on_rs := is_rd_shared && !is_rd_nsd && tbe.dataToBeInvalid;
 
   if (is_rd_once) {
     tbe.snd_msgType := CHIDataType:CompData_I;
--- a/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
+++ b/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
@@ -2600,18 +2600,18 @@ action(Send_CompData, desc="") {
   bool is_rd_nsd := tbe.reqType == CHIRequestType:ReadNotSharedDirty;
   bool is_rd_unique := tbe.reqType == CHIRequestType:ReadUnique;
 
-  // if the request type allows SD
-  bool snd_dirty_on_rs := is_rd_shared && !is_rd_nsd;
-
   // Send UC/UD on ReadShared or ReadNotSharedDirty if the line has no sharers
   // and one of the followings are met
   // 1) the config allows or
   // 2) local cache won't have the line or
-  // 3) dirty will be passed
+  // 3) Dirty will be passed but the request doesn't allow SD
   bool snd_unique_on_rs := (fwd_unique_on_readshared ||
                             tbe.dataToBeInvalid ||
-                            snd_dirty_on_rs)
+                            (tbe.dataDirty && is_rd_nsd))
                           && tbe.dataUnique && tbe.dir_sharers.isEmpty();
+  // if the request type allows and we won't be caching the data
+  bool snd_dirty_on_rs := (is_rd_shared && !is_rd_nsd) &&
+                          !tbe.dir_ownerExists;
 
   if (is_rd_once) {
     tbe.snd_msgType := CHIDataType:CompData_I;
--- a/src/mem/ruby/protocol/chi/CHI-cache-funcs.sm
+++ b/src/mem/ruby/protocol/chi/CHI-cache-funcs.sm
@@ -892,10 +892,10 @@ void copyCacheAndDir(CacheEntry cache_entry, DirEntry dir_entry,
   if (is_valid(cache_entry) &&
       ((initialState == State:UD) || (initialState == State:SD) ||
         (initialState == State:UC) || (initialState == State:SC) ||
-        (initialState == State:UD_RSC) || (initialState == State:SD_RSC) ||
-        (initialState == State:UC_RSC) || (initialState == State:SC_RSC) ||
-        (initialState == State:UD_RSD) || (initialState == State:SD_RSD) ||
-        (initialState == State:UD_T))) {
+        (initialState == State:UD_RU) || (initialState == State:UD_RSC) ||
+        (initialState == State:SD_RSC) || (initialState == State:UC_RSC) ||
+        (initialState == State:SC_RSC) || (initialState == State:UD_RSD) ||
+        (initialState == State:SD_RSD) || (initialState == State:UD_T))) {
     tbe.dataBlk := cache_entry.DataBlk;
     tbe.dataBlkValid.fillMask();
     tbe.dataValid := true;
--- a/src/mem/ruby/protocol/chi/CHI-cache-funcs.sm
+++ b/src/mem/ruby/protocol/chi/CHI-cache-funcs.sm
@@ -892,10 +892,10 @@ void copyCacheAndDir(CacheEntry cache_entry, DirEntry dir_entry,
   if (is_valid(cache_entry) &&
       ((initialState == State:UD) || (initialState == State:SD) ||
         (initialState == State:UC) || (initialState == State:SC) ||
-        (initialState == State:UD_RU) || (initialState == State:UD_RSC) ||
-        (initialState == State:SD_RSC) || (initialState == State:UC_RSC) ||
-        (initialState == State:SC_RSC) || (initialState == State:UD_RSD) ||
-        (initialState == State:SD_RSD) || (initialState == State:UD_T))) {
+        (initialState == State:UD_RSC) || (initialState == State:SD_RSC) ||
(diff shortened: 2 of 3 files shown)


ivanaamit:
Hi @tiagormk, could you take a look at this PR? Thank you.

src/mem/ruby/protocol/chi/CHI-cache-funcs.sm
```diff
@@ -892,10 +892,10 @@ void copyCacheAndDir(CacheEntry cache_entry, DirEntry dir_entry,
   if (is_valid(cache_entry) &&
       ((initialState == State:UD) || (initialState == State:SD) ||
         (initialState == State:UC) || (initialState == State:SC) ||
-        (initialState == State:UD_RSC) || (initialState == State:SD_RSC) ||
-        (initialState == State:UC_RSC) || (initialState == State:SC_RSC) ||
-        (initialState == State:UD_RSD) || (initialState == State:SD_RSD) ||
-        (initialState == State:UD_T))) {
+        (initialState == State:UD_RU) || (initialState == State:UD_RSC) ||
```
tiagormk: We don't set dataValid=true if any upstream cache has the data in unique state to prevent the local copy from being used (because it may have been modified upstream). This changes allows dataValid=true when the state UD_RU. I think this can generate inconsistency when handling other transactions.

tiagormk (commented):
Hi @jjuninho could you elaborate a bit on which events could lead to the loss of a dirty data block ?

If the local cache has the data in UD state and an upstream cache requests it as unique (or as shared, but the fwd_unique_on_readshared param is set) , the data should be always sent upstream as UD (see https://github.com/gem5/gem5/blob/stable/src/mem/ruby/protocol/chi/CHI-cache-actions.sm#L2588C9-L2592C6), thus we should be able to always drop our block since it's the upstream responsibility to writeback the dirty block.

jjuninho:
@tiagormk , Thank you very much for the comments. My understanding on UD_RU state was wrong.

Btw, here's the sequence that led to the loss of dirty data. I apologize that I have lost the full traces I used for debugging couple of months ago, and writing this based on my memory.

Configs:
- L2 caches: (roguhly inclusive to L1 without back invalidation)
  - dealloc_on_*=false
  - dealloc_backinv_*=false
- L3 caches: (rouhgly exclusive without back invalidation)
  - alloc_on_readshared=true
  - alloc_on_readunique=false
  - dealloc_on_shared=false
  - dealloc_on_unique=true
  - dealloc_backinv_*=false
  - is_HN=false
- Last-level cache: (same clusivity as L3)
  - is_HN=true
- allow_SD=true for all caches
- fwd_unique_on_readshared=false for all caches

Sequence:
1. L1 sends ReadUnique then becomes UD. L2 is UC_RU. L3 and LLC are RU.
2. L1 evicts the line to L2 by WriteBack_UD_PD. L2 becomes UD.
3. L2 evicts the line to L3 using WriteBackFull_UD_PD. L3 becomes UD.
4. L1 reads the line with ReadShared. L2 reads the line with ReadShared. L2 becomes SC_RSC. L3 becomes UD_RSC.
5. L3 evicts the line to LLC.
    - At this time, `Initiate_Replacement_WB` sends WriteClean because L3 doesn't back-invalidate and still has sharer. https://github.com/gem5/gem5/blob/d1fca18eb37d53b1e6fcb7802b35f4be87fd81ec/src/mem/ruby/protocol/chi/CHI-cache-actions.sm#L1380C1-L1391C4
    - The cache line in L3 eventually goes Invalid. https://github.com/gem5/gem5/blob/d1fca18eb37d53b1e6fcb7802b35f4be87fd81ec/src/mem/ruby/protocol/chi/CHI-cache-transitions.sm#L765
    - L3 becomes RUSC and LLC becomes UD_RU.
6. UD_RU line is just dropped at LLC.

Although my memory on the sequence is correct, I should've changed the other part to keep the meaning of UD_RU. It'll be grategul if you give any advice on that.

tiagormk:
Thanks for input @jjuninho. I think what you said make sense.  The solution for this is a bit tricky though. If we maintain the meaning of UD_RU then the Writeback in your patch will fail because the data your are writing is maked as invalid.

IMO it makes more sense to change Send_CompData (https://github.com/gem5/gem5/blob/stable/src/mem/ruby/protocol/chi/CHI-cache-actions.sm#L2568) to always propagate the dirty state in those conditions you described. E.g.:

Change:
  // if the config allows (or not caching the data) and line has no sharers
  bool snd_unique_on_rs := (fwd_unique_on_readshared || tbe.dataToBeInvalid)
                          && tbe.dataUnique && tbe.dir_sharers.isEmpty();
  // if the request type allows and we won't be caching the data
  bool snd_dirty_on_rs := is_rd_shared && !is_rd_nsd && tbe.dataToBeInvalid;

To:
  // check if the request type allows sending SD
  bool snd_dirty_on_rs := is_rd_shared && !is_rd_nsd;

  // send UC/UD if line has no sharers and config allows, or not caching the data, 
  // or we'd send SD but can't
  bool snd_unique_on_rs := (fwd_unique_on_readshared || tbe.dataToBeInvalid || (tbe.dataDirty && !snd_dirty_on_rs))
                          && tbe.dataUnique && tbe.dir_sharers.isEmpty();

src/mem/ruby/protocol/chi/CHI-cache-actions.sm
```diff
@@ -2600,11 +2600,18 @@ action(Send_CompData, desc="") {
...
+  bool snd_dirty_on_rs := is_rd_shared && !is_rd_nsd;
+
+  // Send UC/UD on ReadShared or ReadNotSharedDirty if the line has no sharers
+  // and one of the followings are met
+  // 1) the config allows or
+  // 2) local cache won't have the line or
+  // 3) dirty will be passed
+  bool snd_unique_on_rs := (fwd_unique_on_readshared ||
+                            tbe.dataToBeInvalid ||
+                            snd_dirty_on_rs)
```
tiagormk: could we change this line from `snd_dirty_on_rs` to `(tbe.dataDirty && !snd_dirty_on_rs)` ?

I think we still want to send CompData_SD when `fwd_unique_on_readshared==false`. Only if the request doesn't allow it (`snd_dirty_on_rs==false`) we would ignore fwd_unique_on_readshared and send CompData_UD.

jjuninho: @tiagormk What about changing as below.
```
  // if the config allows (or not caching the data) and line has no sharers
  // Send UC/UD on ReadShared or ReadNotSharedDirty if the line has no sharers
  // and one of the followings are met
  // 1) the config allows or
  // 2) local cache won't have the line or
  // 3) Dirty will be passed but the request doesn't allow SD
  bool snd_unique_on_rs := (fwd_unique_on_readshared ||
                            tbe.dataToBeInvalid ||
                            (tbe.dataDirty && is_rd_nsd && !tbe.dir_ownerExists))
                           && tbe.dataUnique && tbe.dir_sharers.isEmpty();

  // if the request type allows SD
  bool snd_dirty_on_rs := ((is_rd_shared && !is_rd_nsd) ||
                           (is_rd_nsd && snd_unique_on_rs)) &&
                          !tbe.dir_ownerExists;
```

Baseline assumption is 'ditry' is passed wherever possible,
- `tbe.dataDirty && is_rd_nsd && !tbe.dir_ownerExists` is to specify the condition for passing dirty to ReadNotSharedDirty. If dirty will be passed for ReadNotSharedDirty, unique should be sent.
- Make it more straightforward for passing dirty condition; If owner exists upstream, then don't pass dirty. Otherwise, if the request is ReadShared, then pass dirty. If the request is ReadNotSharedDitry, send dirty only if Unique will be sent. This implies dirty is passed wherever possible.

tiagormk: Thanks for the updates ! Made a few tweaks in your code below:
```
  // if the config allows (or not caching the data) and line has no sharers
  // Send UC/UD on ReadShared or ReadNotSharedDirty if the line has no sharers
  // and one of the followings are met
  // 1) the config allows or
  // 2) local cache won't have the line or
  // 3) Dirty will be passed but the request doesn't allow SD
  bool snd_unique_on_rs := (fwd_unique_on_readshared ||
                            tbe.dataToBeInvalid ||
                            (tbe.dataDirty && is_rd_nsd))
                           && tbe.dataUnique && tbe.dir_sharers.isEmpty();

  // if the request type allows SD
  bool snd_dirty_on_rs := (is_rd_shared && !is_rd_nsd) &&
                                            !tbe.dir_ownerExists;
```
- removed `!tbe.dir_ownerExists` from `snd_unique_on_rs`. The ` tbe.dir_sharers.isEmpty()`  check already implies `!tbe.dir_ownerExists`.
- removed `(is_rd_nsd && snd_unique_on_rs)` from `snd_dirty_on_rs`, since `snd_dirty_on_rs` is only actually checked if snd_unique_on_rs==false.

ivanaamit:
Hi @jjuninho,

```
f=.git/hooks/commit-msg
mkdir -p $f
chmod +x $f
```

jjuninho:
@ivanaamit , Thank you for the guide. I'm aware of that this pull request is not satisfying the prerequisites for merge. I'm working behind the heavy company firewalls and haven't found the way to make the pre-commit work. I'll apply your guide out side of my office and push the commits again.

BobbyRBruce (approved):

src/mem/ruby/protocol/chi/CHI-cache-actions.sm
jjuninho: @tiagormk , I've updated the codes accordingly. Please check the changes.

PR #794, opened 2024-01-23
arch-riscv: add unit-stride fault-only-first loads (i.e. vle*ff)

saul44203:
This patch provides unit-stride fault-only-first loads (i.e. vle*ff) for the RISC-V architecture.

They are implemented within the regular unit-stride load (i.e.  vle*). A snippet named `fault_code` is inserted with templating to change their behaviour to fault-only-first.

A part from this, a new micro based on the vset\*vl\* instructions (VlFFTrimVlMicroOp) is inserted as the last micro in the macro constructor to trim the VL to it's corresponding length based on the faulting index.

This trimming micro waits for the load micros to finish (via data dependency) and has a reference to the other micros to check whether they faulted or not. The new VL is calculated with the VL of each micro, stopping on the first faulting one (if there's such a fault).

I've tested this with VLEN=128,256,...,16384 and all the corresponding SEW+LMUL configurations.

Change-Id: I7b937f6bcb396725461bba4912d2667f3b22f955

### DIFF
--- a/src/arch/riscv/insts/vector.hh
+++ b/src/arch/riscv/insts/vector.hh
@@ -31,6 +31,7 @@
 
 #include <string>
 
+#include "arch/riscv/faults.hh"
 #include "arch/riscv/insts/static_inst.hh"
 #include "arch/riscv/isa.hh"
 #include "arch/riscv/regs/misc.hh"
@@ -306,6 +307,10 @@ class VseMacroInst : public VectorMemMacroInst
 
 class VleMicroInst : public VectorMicroInst
 {
+  public:
+    mutable bool trimVl;
+    mutable uint32_t faultIdx;
+
   protected:
     Request::Flags memAccessFlags;
 
@@ -313,6 +318,7 @@ class VleMicroInst : public VectorMicroInst
                   uint32_t _microVl, uint32_t _microIdx, uint32_t _vlen)
         : VectorMicroInst(mnem, _machInst, __opClass, _microVl,
                             _microIdx, _vlen)
+        , trimVl(false), faultIdx(_microVl)
     {
         this->flags[IsLoad] = true;
     }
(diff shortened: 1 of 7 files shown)


src/arch/riscv/insts/vector.hh
```diff
@@ -572,6 +578,101 @@ class VxsatMicroInst : public VectorArithMicroInst
...
+                          _microVl, _microIdx, _vlen),
+          microops(_microops)
+    {
+        setRegIdxArrays(
+            reinterpret_cast<RegIdArrayPtr>(
+                &std::remove_pointer_t<decltype(this)>::srcRegIdxArr),
+            nullptr
+        );
+
+        // Create data dependency with load micros
```
rogerchang23424: Thanks for the contribution. Cloud you move the implementation to `vector.cc`?

src/arch/riscv/isa/formats/vector_mem.isa
```diff
@@ -42,6 +42,18 @@ def declareVMemTemplate(class_name):
     template class {class_name}<uint64_t>;
     '''
 
+fault_code = '''
```
rogerchang23424: I would like to align the format with others

```suggestion
def getFaultCode():
   return '''
```

rogerchang23424 (changes requested):

rogerchang23424:
The PR failed the `rvv_strcpy` and `rvv_strlen` tests of [rvv-intrinsic-doc](https://github.com/riscv-non-isa/rvv-intrinsic-doc) in SE mode. Please fix it.

saul44203:
Hi @rogerchang23424, thanks for the feedback.

saul44203:
Hi @rogerchang23424, thanks for the feedback.

With respect to the rvv_strlen issue, the error was on the vlm.v and vsm.v instructions. The vl was not right and the mask was not being loaded/stored properly. I added another commit to fix this.

For rvv_strcpy, apart from the vlm.v/vsm.v thing, there was also the vmsif issue that was fixed in #814. I pulled the changes and now it works.

To test the fault code I also used the rvv_strlen and modified it to make it produce an invalid page fault. Tell me if you want me to share it.

rogerchang23424:
Thanks for your update. The tests work well on my side

src/arch/riscv/isa/templates/vector_mem.isa
```diff
@@ -452,8 +466,7 @@ def template VlmConstructor {{
     %(set_reg_idx_arr)s;
     %(constructor)s;
 
-    const uint32_t micro_vlmax = vlen / width_EEW(_machInst.width);
-    int32_t micro_vl = (std::min(this->vl, micro_vlmax) + 7) / 8;
+    int32_t micro_vl = (this->vl + 7) / 8;
```
rogerchang23424: I think we can separate the `fix vlm.v/vsm.v` into different PR. How do you think

saul44203: Done #830.

BobbyRBruce (approved):

PR #795, opened 2024-01-23
systemc: Reduce unnecessary backdoor request in atomic transaction

wmin0:
The backdoor request in b_transport is only used for hinting the dmi capability. Since most of traffic patterns are continous, we can cache the previous backdoor request result to spare the backdoor inspect of next request.

Change-Id: I53c47226f949dd0be19d52cad0650fcfd62eebbc

### DIFF
--- a/src/systemc/tlm_bridge/tlm_to_gem5.cc
+++ b/src/systemc/tlm_bridge/tlm_to_gem5.cc
@@ -301,6 +301,24 @@ TlmToGem5Bridge<BITWIDTH>::destroyPacket(PacketPtr pkt)
 
 template <unsigned int BITWIDTH>
 void
+TlmToGem5Bridge<BITWIDTH>::cacheBackdoor(gem5::MemBackdoorPtr backdoor)
+{
+    if (backdoor == nullptr) return;
+
+    // We only need to register the callback at the first time.
+    if (requestedBackdoors.find(backdoor) == requestedBackdoors.end()) {
+        backdoor->addInvalidationCallback(
+            [this](const MemBackdoor &backdoor)
+            {
+                invalidateDmi(backdoor);
+            }
+        );
+        requestedBackdoors.emplace(backdoor);
+    }
+}
+
+template <unsigned int BITWIDTH>
+void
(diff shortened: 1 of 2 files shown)


powerjg (commented):
Since this is in the TLM code and not the core gem5, I'm happy to just approve. However, could you explain what the benefit of this change is? It seems to be making the code significantly more complex, and the benefit of this complexity isn't clear from the commit message.

wmin0:
Hi Jason,

The benefit depends on what the complexity of memory side model is. If the memory model is a SimpleMemory. Since handling the backdoor requests almost zero cost in SimpleMemory, the benefit is small. If the memory model models permission check, the model may need to calculate the correct allowed range for each backdoor request. In that case, reducing repeated backdoor request would speed up the whole simulation.

I'm not sure if it's clear for you about the intention of this change. I can update the commit message as well if needed.

Thanks.

powerjg:
So the intention is to increase performance. That makes sense. Could you give us an estimate for how much performance you get from this? 

There's a tradeoff between the maintainability of code and the performance. If this is getting 10% improvement, then it's totally worth it. If it gets 0.1% improvement, it's probably not worth it. 

I would say this particular change cause a pretty significant decrease in maintainability (e.g., if I think about coming back to this code in 5 years I foresee having no idea how to modify it). However, it's also not in a part of the codebase that is likely to need maintenance. So, I would say if it provides some tangible performance benefit, it's probably worth merging, but it would be nice to know what that tangible performance benefit is.

wmin0:
Hi Jsaon,

I only tested this change with our internal memory model stacks. I got around 5% improvement. But changing to SimpleMemory, I got nothing. It's hard to show a reproducible experiment to you since it highly depends on how complicated of the memory model stacks. I'll try to add more comments in code to make it more clear. Hoping it would make you feel better. If it still not works to you, I would consider withdrawal this change. I totally understand your concern since there's not so much users actively using gem5 + systemc plaform...

Thanks.

wmin0:
Change updated.

powerjg (approved):
Thanks for the info. 5% is not nothing!

wmin0:
Thanks for your help :)

ivanaamit (approved):

PR #797, opened 2024-01-23
Increased packets sanity check limit to 1024

amatabsc:
For some simulations with big values for VLEN (e.g. 8k and 16k) there were more packets created on the fly and, as a consequence, failing the simulations. The sanity check has been increased in order to solve this high VLEN cases.

Supervised by [@aarmejach](https://github.com/aarmejach)
Change-Id: I137b0f3113687b3fc9c4154d19ca5e8017e6e992

### DIFF
--- a/src/mem/packet_queue.cc
+++ b/src/mem/packet_queue.cc
@@ -118,8 +118,8 @@ PacketQueue::schedSendTiming(PacketPtr pkt, Tick when)
 
     // add a very basic sanity check on the port to ensure the
     // invisible buffer is not growing beyond reasonable limits
-    if (!_disableSanityCheck && transmitList.size() > 128) {
-        panic("Packet queue %s has grown beyond 128 packets\n",
+    if (!_disableSanityCheck && transmitList.size() > 1024) {
+        panic("Packet queue %s has grown beyond 1024 packets\n",
               name());
     }
 


hnpl (approved):
This looks good to me.

Side note: I'm not very familiar with how PacketQueue objects are used, but it seems that the capacity of the PacketQueue is something that should be parameterized.

andysan:
It's worth noting that the packet queue exploding like this is potentially a big problem in terms of simulation accuracy. 1024 packets holding 64B cache lines would be equivalent to 64KiB of magic cache (behaving a lot like a low-latency L1 cache) that lives in non-intuitive places. That's quite a lot of memory that users won't know about. The reason this limit was set pretty low (I think it used to be lower than 128) was to avoid artificially increasing the amount of "invisible" caching in the system.

The "right" way of dealing with this would be for components that use the `PacketQueue` to apply throttling when packets when packets are getting backed up. Effectively, the panic triggering is indicative of a modelling bug.

powerjg:
This is a change that people have been requesting for a long time. I think we should merge it, despite the potential issues.

>  The reason this limit was set pretty low (I think it used to be lower than 128) was to avoid artificially increasing the amount of "invisible" caching in the system.

I **strongly** agree with this statement! However, in many cases the packet queue is used where there actually already is modeling of the buffers/caching. For instance, in the memory controllers, there is a packet queue on the response, but it is correctly modeled in the read/write queues on the request side.

I think the *better* solution would be to remove the packet queue from gem5 completely and put correct backpressure modeling throughout the system, but that's unlikely to be completed anytime soon. In the meantime, there's many use cases (vectors, GPUs, etc) where we need to support up to 1TB/s or more of bandwidth, and the packet queue limit is a perennial pain in these common use cases.

andysan:
> This is a change that people have been requesting for a long time. I think we should merge it, despite the potential issues.

I'm a bit worried that this will just hide the problem and lead to weird unforeseen consequences, but we need some sort of short-term hack that avoids panics. Would it be better to use the current level (128 entries) as a warning? We could also consider adding a panic with a higher, configurable, threshold that the component owning the queue can set. 

> > The reason this limit was set pretty low (I think it used to be lower than 128) was to avoid artificially increasing the amount of "invisible" caching in the system.
> 
> I **strongly** agree with this statement! However, in many cases the packet queue is used where there actually already is modeling of the buffers/caching. For instance, in the memory controllers, there is a packet queue on the response, but it is correctly modeled in the read/write queues on the request side.

A short term fix would be to suppress the sanity check in such cases. It'd be interesting to understand how common this pattern is. It seems like a deeper queue wouldn't necessarily affect the modelling accuracy in these cases.

> I think the _better_ solution would be to remove the packet queue from gem5 completely and put correct backpressure modeling throughout the system, but that's unlikely to be completed anytime soon. In the meantime, there's many use cases (vectors, GPUs, etc) where we need to support up to 1TB/s or more of bandwidth, and the packet queue limit is a perennial pain in these common use cases.

Would it make sense to make the check configurable using a (global?) option to support these use cases? Increasing the limit to something big enough to cover this use case seems like it defeats the purpose of the check for the majority of users.

abmerop:
I don't have a strong preference on this either way, but I have a few comments that might be relevant here.  For reference, the comments on this change when it came up previously (2018) was:  why not just disable the sanity check for your specific simulation?

Regarding parameterizing it, I vaguely recall attempting to do this and it was a bit difficult because the PacketQueue is not a SimObject.  I think trying to add the plumbing to pass down the value was not straightforward too.

I think the packet queue is only used in the Queued*Ports.  In the GPU model at least, we do not use these. I did a quick grep and couldn't find them in the configs/ directory either.  I basically agree that we should prefer modeling true back pressure and not have magic buffers.  But that would be a much larger change.

andysan:
> I don't have a strong preference on this either way, but I have a few comments that might be relevant here. For reference, the comments on this change when it came up previously (2018) was: why not just disable the sanity check for your specific simulation?
> 
> Regarding parameterizing it, I vaguely recall attempting to do this and it was a bit difficult because the PacketQueue is not a SimObject. I think trying to add the plumbing to pass down the value was not straightforward too.

Yes, this is definitely an issue. One option we should consider is making this a global simulation option. It's a bit of a hack, but we could add a param in the system or root object. The former would require a change in the API since there is currently no way of accessing the system from `PacketQueue`. It might be worth making this change for other reasons like checking the memory mode. We should probably make the default limit configurable as well if we have the ability to pass options to the queue.

powerjg (approved):
I'm going to approve this change. I think the pros outweigh the cons here. If there is a true bug in someone's implementation, a packet queue of 1000 should also catch the issue just as well as a queue of 128, though it may take longer. Furthermore, it's important for a wide variety of people's research to enable higher bandwidth connections that can (correctly) overwhelm a queue of 128. 

The best solution would be to remove the packet queue, but we currently don't have the resources or a volunteer to do this. Until that happens, I don't think we'll be causing any one's research to break with this change, and we will be enabling people to have fewer local changes as well.

PR #804, opened 2024-01-24
util: Update gcn-gpu to remove GCN3 add gfx902

abmerop:
This removes the gfx803 and gfx801 targets for building applications as GCN3 will be removed from gem5. It also removes the copy/paste bug from the HACC docker which is clobbering the HCC targets and removing gfx902.

Change-Id: I9a0d7fda437e797baf0f743a0a450948b9260b07


abmerop:
Ideally this would be labelled gcn-gpu:v24-0 and not override the v23-1

mattsinc (dismissed):

BobbyRBruce:
Was there some rebase error here. Where is the adding of gfx902?

abmerop:
It is on line 107 which was already there. This is just removing the second definition of HCC_AMDGPU_TARGET which is clobbering the correct value with gfx902 in the list

BobbyRBruce (approved):

BobbyRBruce:
> It is on line 107 which was already there. This is just removing the second definition of HCC_AMDGPU_TARGET which is clobbering the correct value with gfx902 in the list

Ah, I see. Thanks for the explanation.

PR #809, opened 2024-01-26
arch-riscv: adding support for local interrupts

robhau:
Besides the standard RISC-V interrupts software, timer, and external interrupt, the RISC-V specification also offers the possibility to implement local interrupts. With this patch, we contribute an extension of RiscvInterrupts that enables connecting interrupt sources to the local interrupt controller. We assigned the local interrupts to machine-level and gave them the highest priority. If two local interrupts are pending, there exception code will be the tie-breaker (higher ID > lower ID). 32 Bit systems only recognize the local interrupts 16 to 31, 64 Bit systems 16 to 63.

Change-Id: Iff8d34e740b925dce351c0c6f54f4bd37a647e0c


robhau:
I forgot to select the develop branch. Hence, I close this pull request.

PR #810, opened 2024-01-26
arch-arm: Replace CRYPTO extension with canonical names

giactra:
These are:

FEAT_AES,
FEAT_PMULL,
FEAT_SHA256,
FEAT_SHA1,
FEAT_CRC32

With this patch we are also enabling them by default by adding them to the Armv8 release object. Some of them are mandatory anyway since Armv8.1

Change-Id: I221ae8646d91151fdfaf97a4815168a4fe3d8c5a


powerjg (approved):
lgtm

PR #811, opened 2024-01-26
arch-riscv: adding support for local interrupts

robhau:
Besides the standard RISC-V interrupts software, timer, and external interrupt, the RISC-V specification also offers the possibility to implement local interrupts. With this patch, we contribute an extension of RiscvInterrupts that enables connecting interrupt sources to the local interrupt controller. We assigned the local interrupts to machine-level and gave them the highest priority. If two local interrupts are pending, there exception code will be the tie-breaker (higher ID > lower ID). 32 Bit systems only recognize the local interrupts 16 to 31, 64 Bit systems 16 to 63.

Change-Id: Iff8d34e740b925dce351c0c6f54f4bd37a647e0c

PR #813, opened 2024-01-26
arch-riscv: adding support for local interrupts

robhau:
Besides the standard RISC-V interrupts software, timer, and external interrupt, the RISC-V specification also offers the possibility to implement local interrupts. With this patch, we contribute an extension of RiscvInterrupts that enables connecting interrupt sources to the local interrupt controller. We assigned the local interrupts to machine-level and gave them the highest priority. If two local interrupts are pending, there exception code will be the tie-breaker (higher ID > lower ID). 32 Bit systems only recognize the local interrupts 16 to 31, 64 Bit systems 16 to 63.

I was not able to reopen the pull request https://github.com/gem5/gem5/pull/811 after my force push. Therefore, i opened a new pull request again. I hope that it is not problematic 😅 .

Change-Id: Iff8d34e740b925dce351c0c6f54f4bd37a647e0c


src/arch/riscv/interrupts.hh
```diff
@@ -137,83 +82,32 @@ class Interrupts : public BaseInterrupts
...
-        unsigned long ip_ulong;
-        unsigned long ie_ulong;
-        UNSERIALIZE_SCALAR(ip_ulong);
-        ip = ip_ulong;
-        UNSERIALIZE_SCALAR(ie_ulong);
-        ie = ie_ulong;
-    }
+    void unserialize(CheckpointIn &cp) override;
+
+    Port &getPort(const std::string &if_name, PortID idx);
```
rogerchang23424: missing override here

src/arch/riscv/faults.hh
```diff
@@ -93,6 +94,54 @@ enum ExceptionCode : uint64_t
     INT_EXT_USER = 8,
     INT_EXT_SUPER = 9,
     INT_EXT_MACHINE = 11,
+    INT_LOCAL_16 = 16,
```
rogerchang23424: I would like to start with 0 like `INT_LOCAL_0, INT_LOCAL_1 ...`

Ref: https://five-embeddev.com/quickref/interrupts.html

rogerchang23424 (changes requested):

src/arch/riscv/RiscvInterrupts.py
```diff
@@ -27,10 +28,35 @@
...
+from m5.objects.IntPin import VectorIntSinkPin
+from m5.params import VectorParam
 
 
 class RiscvInterrupts(BaseInterrupts):
     type = "RiscvInterrupts"
     cxx_class = "gem5::RiscvISA::Interrupts"
     cxx_header = "arch/riscv/interrupts.hh"
+
+    local_interrupt_pins = VectorIntSinkPin("Pins for local interrupts")
```
rogerchang23424, line 42: Just curious. Why we need to add `intSinkPin` in the `RiscvInterrupts`.

Don't you think post interrupt in the model without local_interrupt_pins connection  
`tc->getCpuPtr()->postInterrupt(tc->threadId(), <int_local_num>, 0)` is good?

src/arch/riscv/faults.hh
robhau: Do you mean `INT_LOCAL_0` to `INT_LOCAL_47`?

I do not understand your reference. In https://five-embeddev.com/quickref/interrupts.html, the local interrupts also start with 16 (`platform_defined16, platform_defined17 ...`)

src/arch/riscv/RiscvInterrupts.py
robhau: I think it would be better to modularize this, as it is more flexible. If you, for example, want to connect your device to an other hart, you just have to do a simple change in the configuration switch. If you want to post the interrupt directly from the external device model, you would have to count your hardware threads in order to obtain the correct thread.

rogerchang23424: Ack

src/arch/riscv/faults.hh
rogerchang23424: > Do you mean `INT_LOCAL_0` to `INT_LOCAL_47`?

Yes

> I do not understand your reference. In https://five-embeddev.com/quickref/interrupts.html, the local interrupts also start with 16 (`platform_defined16, platform_defined17 ...`)

It means the interrupt ID 16 - X is defined by platform. It comment `Optional platform defined interrupt source 0.` in the description field for the interrupt ID 16. So I think use `INT_LOCAL_0` will make it clear to know the interrupt source it from 0.

Another example for using local interrupt in SiFive E31:

 
clint interrupt

robhau: I think it is better to keep `INT_LOCAL_16 ...` to highlight that, e.g., `INT_LOCAL_i` corresponds to the i-th bit in `mie` and `mip`.

robhau: I forgot to refresh my browser before writing the comment above. Thank you for the examples. I will change this.

src/arch/riscv/regs/misc.hh
```diff
@@ -1438,6 +1440,54 @@ USTATUS_MASKS[enums::Num_RiscvType][enums::Num_PrivilegeModeSet] = {
...
+const RegVal LOCAL_25_MASK = 1ULL << 25;
+const RegVal LOCAL_24_MASK = 1ULL << 24;
+const RegVal LOCAL_23_MASK = 1ULL << 23;
+const RegVal LOCAL_22_MASK = 1ULL << 22;
+const RegVal LOCAL_21_MASK = 1ULL << 21;
+const RegVal LOCAL_20_MASK = 1ULL << 20;
+const RegVal LOCAL_19_MASK = 1ULL << 19;
+const RegVal LOCAL_18_MASK = 1ULL << 18;
+const RegVal LOCAL_17_MASK = 1ULL << 17;
+const RegVal LOCAL_16_MASK = 1ULL << 16;
```
rogerchang23424: Please align the number `INT_LOCAL_0`, `INT_LOCAL_1` ...

src/arch/riscv/regs/misc.hh
```diff
@@ -1438,6 +1440,54 @@ USTATUS_MASKS[enums::Num_RiscvType][enums::Num_PrivilegeModeSet] = {
...
+const RegVal LOCAL_50_MASK = 1ULL << 50;
+const RegVal LOCAL_49_MASK = 1ULL << 49;
+const RegVal LOCAL_48_MASK = 1ULL << 48;
+const RegVal LOCAL_47_MASK = 1ULL << 47;
+const RegVal LOCAL_46_MASK = 1ULL << 46;
+const RegVal LOCAL_45_MASK = 1ULL << 45;
+const RegVal LOCAL_44_MASK = 1ULL << 44;
+const RegVal LOCAL_43_MASK = 1ULL << 43;
+const RegVal LOCAL_42_MASK = 1ULL << 42;
+const RegVal LOCAL_41_MASK = 1ULL << 41;
```
rogerchang23424: ditto

rogerchang23424 (changes requested):

src/arch/riscv/regs/misc.hh
robhau: Thank you for your hint. I already replaced this individual bit masks for every local interrupt with `mask(63,16)` from `bitfield.hh`, but I forgot to delete the mentioned variables 😅 .

rogerchang23424 (approved):
LGTM

BobbyRBruce:
@robhau : Do you mind if we "Squash merge" this PR into the develop branch?

robhau:
@BobbyRBruce I do not mind ☺️

powerjg (approved):

src/arch/riscv/interrupts.hh
```diff
@@ -137,83 +82,32 @@ class Interrupts : public BaseInterrupts
...
-        ip = ip_ulong;
-        UNSERIALIZE_SCALAR(ie_ulong);
-        ie = ie_ulong;
-    }
+    void unserialize(CheckpointIn &cp) override;
+
+    Port &getPort(const std::string &if_name, PortID idx) override;
+
+    void raiseInterruptPin(uint32_t num);
+    void lowerInterruptPin(uint32_t num) {};
```
rogerchang23424, line 110: I think the device cloud not clear the interrupt pending without handling the `lowerInterruptPin`, is right?

robhau: Yes, the device can not clear the interrupt, it has to be done in software. If i understand interrupt pins correctly, two consecutive raise() calls on a `IntSourcePin` do not lead to two `raiseInterruptPin()` calls at the `IntSinkPin` device. Therefore, an interrupt source that does not support interrupt clearing is still enforced to hold the interrupt pin until the ISR is entered. With an empty `lowerInterruptPin()` function, the interrupt source just calls `raise()` followed by `lower()`.

What do you think about adding an additional parameter in the configuration script that contains the IDs of interrupts that can be cleared by `lower()`? Then, in `lowerInterruptPin()`, interrupts are cleared if `num` is in this list of IDs.

rogerchang23424: I have no idea why you need to raise interrupt pins twice. IMHO, the gem5 will check interrupts after each instructions is commit. So we can just call `raise()` once in the interrupt source and the interrupt pending will be set. For cleaning the interrupt pending, it should be the process of ISR execution.

Just curious the interrupt pending will be clear before jumping the ISR?

robhau: Please correct me if I understood your question wrong:

At the moment, local interrupt pending bits can only be cleared by writing to `mip`.

```asm
csrrc zero, mip, (1 << 16)
```
To enable consecutive interrupts from devices with periodic interrupts, e.g., a sensor, `lowerInterruptPin()` has to be called before the next `raiseInterruptPin()`.

rogerchang23424: Ok, I will check the issue again after long Chinese New Year holiday.

PR #826, opened 2024-01-31
python: Adding fatal statement to notify user mistakes.

mahyarsamani:
This change adds a fatal statement to check all params for all
SimObjects have been unproxied before C++ object are created.
The fatal statement notifies the user of a mistake that could
possibly lead to a SimObject to not have its params unproxied.
This mistake could be made by adding a child SimObject with a
name that starts with an underscore.


src/python/m5/SimObject.py
```diff
@@ -1199,6 +1199,24 @@ def getCCParams(self):
...
+                    "when SimObject::unproxyParams is called on the "
+                    "decendents the root Simobject. This at the time of "
+                    "adding this error message is in simulate.py at lines "
+                    "103-104 (current commit hash: f56459470a). "
+                    "One reason why this might happen is if you have "
+                    "mistakenly added a child SimObject as an attr and not a "
+                    "child by giving it a name that starts with an underscore "
+                    "('_'). Look at SimObject::__setattr__. To check if this "
+                    f"is the case look at where you instantiate {self.path()} "
+                    "and make sure it is not an orphan."
```
powerjg: This is a bit too verbose, I think. If you really think it needs more information, just add a comment here. Someone can grep for this error and find your comment.

```suggestion
                    f"Param {param} for {self._name} has value = {value}. "
                    "This value is a not a valid value. This could be caused "
                    f"by {param} not having been unproxied correctly. "
                    "One reason why this might happen is if you have "
                    "mistakenly added a child SimObject as an attr and not a "
                    "child by giving it a name that starts with an underscore "
                    f"`_`. {self.path()} should not say 'orphan.'"
```

powerjg: @mahyarsamani are you planning to take this into account, or do you have a compelling reason that you shouldn't simplify?

powerjg (approved):

Issue #668, opened 2023-12-10
SE with SMT hangs on RISCV and fails on ARM


cnpapado:
**Describe the bug**
SE SMT simulation hangs on RISCV and fails with `Assertion 'params.numPhysVecPredRegs >= numThreads * regClasses.at(VecPredRegClass)->numRegs()' failed.` (found on `src/cpu/o3/cpu.cc:202`) on ARM.

I've been reading that SMT should work in SE mode, that's why I am opening the issue. However if that's not the case, does anyone know a version were it works at least for ARM? 

**Affects version**
v23.0.1 
In v22.1 simulation completes successfully on RISCV but the same assertion fails on ARM (that part is true also for earlier versions) 

**To Reproduce**
1. Compile with `scons build/RISCV/gem5.opt -jX` and `scons build/ARM/gem5.opt -j48`
2. Execute with 
```
build/RISCV/gem5.opt configs/deprecated/example/se.py '--cmd=tests/test-progs/hello/bin/riscv/linux/hello;tests/test-progs/hello/bin/riscv/linux/hello' --caches --l1d_size=32kB --l1d_assoc=8 --l1i_size=32kB --l1i_assoc=8 --l2cache --cpu-type=DerivO3CPU --mem-type=DDR4_2400_16x4 --mem-size=64GB --smt
```
and 
```
build/ARM/gem5.opt configs/deprecated/example/se.py '--cmd=tests/test-progs/hello/bin/arm/linux/hello;tests/test-progs/hello/bin/arm/linux/hello' --caches --l1d_size=32kB --l1d_assoc=8 --l1i_size=32kB --l1i_assoc=8 --l2cache --cpu-type=DerivO3CPU --mem-type=DDR4_2400_16x4 --mem-size=64GB --smt
```
I know that `se.py` has been deprecated but I suspect it is not a bug in the config script, hence I'm still using it to reproduce. However I'm happy to try with some other configuration file as well.

**Expected behavior**
Successfully complete the simulation.

**Host Operating System**
Ubuntu 22.04

**Host ISA**
X86

**Compiler used**
g++ v10.2.1

Let me know if you need more information. Since I need this especially for ARM, I am happy trying to fix it, but I'll need some guidance since I'm new to gem5. Thanks a lot.

giactra:
Hi @cnpapado, thanks for reporting this. The issue for Arm is that the base O3CPU doesn't allocate enough physical predicate registers by default to implement the two architectural reg files required for the 2 HW threads.

[...]

PR #828, opened 2024-02-01
cpu-o3, arch: Fix SMT bug arising from v23.0 and make gem5 more robust with SMT

giactra:
This PR is fixing https://github.com/gem5/gem5/issues/668. It fixes it for all ISAs other than Arm with the first commit, which is setting the number of architectural Matrix registers to 0 for those ISA which are not using them.

It then partly fixes it for Arm as well with the 2nd commit: by removing RenameMap::numFreeEntries we don't stall renaming unless a matrix instruction is encountered... This means most binaries will run with SMT as long as they don't use FEAT_SME instructions. Please note: this is not simply a SMT fix, it will generally address a shortcoming in the way we were renaming instructions.

If an Arm binary wants to use SMT with FEAT_SME, the 4th commit will make sure the lack of physical registers is notified explicitly at the beginning of simulation, rather than silently blocking renaming

### DIFF
--- a/src/cpu/o3/rename.cc
+++ b/src/cpu/o3/rename.cc
@@ -1260,7 +1260,7 @@ Rename::readFreeEntries(ThreadID tid)
             freeEntries[tid].robEntries,
             freeEntries[tid].lqEntries,
             freeEntries[tid].sqEntries,
-            renameMap[tid]->numFreeEntries(),
+            renameMap[tid]->minFreeEntries(),
             renameMap[tid]->numFreeEntries(IntRegClass),
             renameMap[tid]->numFreeEntries(FloatRegClass),
             renameMap[tid]->numFreeEntries(VecRegClass),
--- a/src/cpu/o3/rename.cc
+++ b/src/cpu/o3/rename.cc
@@ -1225,9 +1225,6 @@ Rename::checkStall(ThreadID tid)
     } else if (calcFreeLQEntries(tid) <= 0 && calcFreeSQEntries(tid) <= 0) {
         DPRINTF(Rename,"[tid:%i] Stall: LSQ has 0 free entries.\n", tid);
         ret_val = true;
-    } else if (renameMap[tid]->numFreeEntries() <= 0) {
-        DPRINTF(Rename,"[tid:%i] Stall: RenameMap has 0 free entries.\n", tid);
-        ret_val = true;
     } else if (renameStatus[tid] == SerializeStall &&
                (!emptyROB[tid] || instsInProgress[tid])) {
         DPRINTF(Rename,"[tid:%i] Stall: Serialize stall and ROB is not "
--- a/src/arch/mips/isa.cc
+++ b/src/arch/mips/isa.cc
@@ -105,7 +105,7 @@ constexpr RegClass vecElemClass(VecElemClass, VecElemClassName, 2,
         debug::IntRegs);
 constexpr RegClass vecPredRegClass(VecPredRegClass, VecPredRegClassName, 1,
         debug::IntRegs);
-constexpr RegClass matRegClass(MatRegClass, MatRegClassName, 1, debug::MatRegs);
+constexpr RegClass matRegClass(MatRegClass, MatRegClassName, 0, debug::MatRegs);
 constexpr RegClass ccRegClass(CCRegClass, CCRegClassName, 0, debug::IntRegs);
 
 } // anonymous namespace
--- a/src/arch/power/isa.cc
+++ b/src/arch/power/isa.cc
@@ -57,7 +57,7 @@ RegClass vecRegClass(VecRegClass, VecRegClassName, 1, debug::IntRegs);
 RegClass vecElemClass(VecElemClass, VecElemClassName, 2, debug::IntRegs);
 RegClass vecPredRegClass(VecPredRegClass, VecPredRegClassName, 1,
         debug::IntRegs);
-RegClass matRegClass(MatRegClass, MatRegClassName, 1, debug::MatRegs);
+RegClass matRegClass(MatRegClass, MatRegClassName, 0, debug::MatRegs);
 RegClass ccRegClass(CCRegClass, CCRegClassName, 0, debug::IntRegs);
 
 } // anonymous namespace
--- a/src/arch/sparc/isa.cc
+++ b/src/arch/sparc/isa.cc
@@ -74,7 +74,7 @@ RegClass vecRegClass(VecRegClass, VecRegClassName, 1, debug::IntRegs);
 RegClass vecElemClass(VecElemClass, VecElemClassName, 2, debug::IntRegs);
 RegClass vecPredRegClass(VecPredRegClass, VecPredRegClassName, 1,
         debug::IntRegs);
-RegClass matRegClass(MatRegClass, MatRegClassName, 1, debug::MatRegs);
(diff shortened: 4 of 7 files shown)


src/cpu/o3/rename.cc
```diff
@@ -1225,9 +1225,6 @@ Rename::checkStall(ThreadID tid)
     } else if (calcFreeLQEntries(tid) <= 0 && calcFreeSQEntries(tid) <= 0) {
         DPRINTF(Rename,"[tid:%i] Stall: LSQ has 0 free entries.\n", tid);
         ret_val = true;
-    } else if (renameMap[tid]->numFreeEntries() <= 0) {
```
powerjg, line 1228: Could you explain why this stall case is no longer needed? I assume that the original intention of this check (something like `renameMaps[<inst type>].numFreeEntries()`) is checked somewhere else, but I'm not immediately seeing it.

powerjg (commented):
Looks good to me. Just one question.

src/cpu/o3/rename.cc
giactra: Hey Jason, I describe it in the commit message: https://github.com/gem5/gem5/pull/828/commits/f53a2bf9d3be38dd60c90373de6891265ad97f95. If it is still not clear I can expand, please let me know and thanks for the review

giactra: In short numFreeEntries is a bit misleading as it does not differentiate according to inst type. It just returns the *minimum* number of freeList registers across all classes. For example, if I have

IntRegs -> 200 free regs
FloatRegs -> 100 free regs
EsoticRegClass -> 0 free regs.

numFreeEntries will return 0!

Then rename will allways stall even if there is no instruction in the pipeline making use of EsoticRegClass registers! This is overly conservative, rename should be stopped only if instructions within the pipeline can't be renamed (and this happen in renameInsts: https://github.com/gem5/gem5/blob/stable/src/cpu/o3/rename.cc#L662

powerjg: Ah! Got it. This makes sense.

powerjg (dismissed):

powerjg (approved):

giactra:
I can see we are failing checkpoint restore tests.. I presume this happens cause some ISAs are passing from 1 MatRegClass to 0 MatRegClass... Is there a way to update the checkpoints?

powerjg:
I guess this means we need to re take the checkpoints or manually fix them? I don't think we can use the checkpoint upgrader in this case. 

We should go ahead and make a note in the release notes that checkpoints will be broken.

giactra:
> I guess this means we need to re take the checkpoints or manually fix them? I don't think we can use the checkpoint upgrader in this case.

Thanks Jason; I think over time I lost track on what needs to be done in order to update a checkpoint. I can see in the gem5 repo the scripts for generating a checkpoint. The issue is I don't know what to do once the checkpoint is generated. Should it be stored in the gem5-resources repo? I can't find any reference of them in https://github.com/gem5/gem5-resources/blob/develop/resources.json.

> 
> We should go ahead and make a note in the release notes that checkpoints will be broken.

You might remember I have always felt strongly against storing a fixed checkpoint somewhere and to use it as a regression. My argument is that it only makes life harder for a developer; especially as I can't find documentation on what needs to be done once a checkpoint test fails. If the purpose is just to flag checkpoint compatibility has been broken, shouldn't we just have checkpoint compliance job producing a warning (and not a failure?).

powerjg:
@BobbyRBruce or @Harshil2107 can you re take the checkpoint and fix this failing test? I'm with @giactra and I'm not sure what the procedure is.

While doing this, we should document the procedure.

BobbyRBruce:
Shouldn't the checkpoint updater be updated to solve this? Presumably checkpoints in v23.0 and v23.1 need updated to work with this change.

I'd say the procedure is, if your break a checkpoint test then the PR should include an checkpoint updater change then we can use that to update the checkpoint and upload a new resource.

powerjg:
In this case, I'm not sure the checkpoint updater will work. I could be wrong, but this is changing the value of something in the checkpoint, not adding a new thing.

Generally, checkpoints are not meant to be backwards or forwards compatible. They fundamentally can't be. There will be times (e.g., at releases) when we have to say "you need to take a new checkpoint." One point of these checkpoint-based tests are to know when checkpoints break.

@giactra the main reason to have these checkpoint tests is to make sure that checkpointing in general doesn't break. If you have a solution to test whether or not checkpointing works that doesn't depend on a particular checkpoint, then I'm open to the idea. However, I am generally against having "warnings" in tests, as they will be ignored by developers and the community. Tests should either pass or fail.

giactra:
> I'd say the procedure is, if your break a checkpoint test then the PR should include an checkpoint updater change then we can use that to update the checkpoint and upload a new resource.

I would argue that this is not always feasible and it does not always make sense to artificially patch a checkpoint. Say for example I took a checkpoint on v21.2, while using a specific vector extension. So I take a checkpoint in the middle of userspace before a ROI. Then after 2 years that feature gets amended/removed. What shall the upgrader do? Whatever modification I do to the checkpoint will likely bring me to a region of code that might possibly cause a fault.

In the opposite scenario (new feature being added) the checkpoint might not make use of the new feature so even in this case it might make sense to retake the checkpoint.


> Generally, checkpoints are not meant to be backwards or forwards compatible. They fundamentally can't be. There will be times (e.g., at releases) when we have to say "you need to take a new checkpoint." One point of these checkpoint-based tests are to know when checkpoints break.
> 
> @giactra the main reason to have these checkpoint tests is to make sure that checkpointing in general doesn't break. If you have a solution to test whether or not checkpointing works that doesn't depend on a particular checkpoint, then I'm open to the idea. However, I am generally against having "warnings" in tests, as they will be ignored by developers and the community. Tests should either pass or fail.

We are already testing checkpointing functionality in Arm, and what we do is we boot Linux while continuously taking and restoring checkpoints [1]. In this way we test checkpoint save/restore without the need to take a snapshot of the simulator which is bound to go stale in case something changes...

I think giving a user the capability to automatically update their checkpoint is extremely useful especially for hard/long to 
craft simpoints. However I believe the cpt_upgrader methodology has the following flaws

1) There is no policy for deprecating/removing cpt upgraders (we still have very old upgraders I doubt anyone is using)
2) Are checkpoint upgraders being tested?

For example now that I am crafting the upgrader for this PR I got the idea most of our upgraders are actually broken. They will probably poison a user checkpoint. For example:

With multi-ISA we don't store root.isa anymore, so all upgraders checking for it are bound to fail [2] [3]. Some people maybe understood this check was broken so it was decided to skip it altogether (See https://github.com/gem5/gem5/pull/171/commits/bfb295ac3f047301d5557a75c7b93cdb7179cd2a which is patching every checkpoint regardless of the ISA, effectively destroying a fully working non RISCV checkpoint).

[1]: https://github.com/gem5/gem5/blob/stable/tests/gem5/fs/linux/arm/test.py#L50
[2]: https://github.com/gem5/gem5/blob/stable/util/cpt_upgraders/arm-sve.py#L10
[3]: https://github.com/gem5/gem5/blob/stable/util/cpt_upgraders/x86-add-tlb.py

Harshil2107:
I will retake the checkpoints and upload them as a new version for the existing ones, I will tag the gem5_version as 24.0.

Harshil2107:
I have updated the checkpoints and reran the tests.

giactra:
Thanks @Harshil2107 !

PR #830, opened 2024-02-01
arch-riscv: fix vl in mask load/store (i.e vlm.v/vsm.v)

saul44203:
The vlm.v and vsm.v unit-stride mask load/store instructions are constructed with an incorrect VL when the current one is larger than than VLEN/EEW  (i.e. when LMUL > 1). This commit fixes the issue for both instructions.

Change-Id: I05e29d327e52fa2a225c6b83b927b01f84b9d2db

### DIFF
--- a/src/arch/riscv/isa/templates/vector_mem.isa
+++ b/src/arch/riscv/isa/templates/vector_mem.isa
@@ -452,8 +452,7 @@ def template VlmConstructor {{
     %(set_reg_idx_arr)s;
     %(constructor)s;
 
-    const uint32_t micro_vlmax = vlen / width_EEW(_machInst.width);
-    int32_t micro_vl = (std::min(this->vl, micro_vlmax) + 7) / 8;
+    int32_t micro_vl = (this->vl + 7) / 8;
     StaticInstPtr microop;
 
     if (micro_vl == 0) {
@@ -479,8 +478,7 @@ def template VsmConstructor {{
     %(set_reg_idx_arr)s;
     %(constructor)s;
 
-    const uint32_t micro_vlmax = vlen / width_EEW(_machInst.width);
-    int32_t micro_vl = (std::min(this->vl, micro_vlmax) + 7) / 8;
+    int32_t micro_vl = (this->vl + 7) / 8;
 
     StaticInstPtr microop;
     if (micro_vl == 0) {


rogerchang23424 (approved):

powerjg (dismissed):

QQeg:
I believe the problem lies in the incorrect calculation of 'micro_vlmax'.
And I think it's better to keep the guard?
```
std::min(this->vl, micro_vlmax)
```

hnpl (commented):
Thanks for the contribution.

I'm not very familiar with RVV masking, so please correct if I'm wrong. Is the problem of the original code is that `micro_vlmax <= this->vl`, and thus for `LMUL = k`, there are `k` micro-ops that write to v0?

Also, do we really need to micro-code the vlm and vsm instructions? I ask this because I believe we can load all the necessary masks using one vle8.v even if LMUL = 8.

Side node: doing `ceil(int, int)` seems to be a very common pattern; this function maybe a good candidate to be implemented in `src/base/bitfield.hh`.

QQeg:
> Is the problem of the original code is that micro_vlmax <= this->vl

Yes, and it's because `micro_vlmax` is fixed to `vlen / width_EEW(_machInst.width)`. For example, if `vlen` is set to `256` then `micro_vlmax` is fixed to `32` (`EEW` will always be `8` for these two instructions). Therefore, if `this->vl` exceeds `32`, it will cause the register to not be properly loaded/stored.

> thus for LMUL = k, there are k micro-ops that write to v0?

There's only one micro-op regardless of LMUL. 
Below is the code snippet from VlmConstructor
```c++
if (micro_vl == 0) {
        microop = new VectorNopMicroInst(_machInst);
    } else {
        microop = new Vle8_vMicro(_machInst, micro_vl, 0, vlen);
        microop->setDelayedCommit();
        microop->setFlag(IsLoad);
    }
```

>  I believe we can load all the necessary masks using one vle8.v even if LMUL = 8.

This is true. Actually, I've already addressed the issue by revising the calculation of `micro_vlmax` which takes `LMUL` into account. So, `micro_vlmax` is not fixed to a number, but I haven't made a PR yet. The solution in this PR seems to work, I just have a slight concern about the removal of the guard ```std::min(this->vl, micro_vlmax) ```.

About the side note, could you please provide me with some references or information about the function `ceil(int, int)`? 
I don't know what it does when `ceil` have two input arguments.

saul44203:
> And I think it's better to keep the guard?

@QQeg I think the guard is not necessary because `ceil(vl/8) <= micro_vlmax` anyways. This is the case because 
`vl <= VLMAX == VLEN/SEW*LMUL == micro_vlmax*LMUL`, and because `LMUL <= 8` you can see that `vl/8 <= micro_vlmax`.

QQeg:
Hi, @saul44203 
First, let me bring the original code here,
```c++
const uint32_t micro_vlmax = vlen / width_EEW(_machInst.width);
int32_t micro_vl = (std::min(this->vl, micro_vlmax) + 7) / 8;
```

The purpose of the guard is to ensure that, like the expression you gave, 
`vl <= VLMAX == VLEN/SEW*LMUL == micro_vlmax*LMUL` is always true.

However, in the original code, the value of `micro_vlmax` is used instead of `VLMAX` in the guard. This is why I pointed out that the issue lies in `micro_vlmax`. We can simply resolve this problem by changing the value of `micro_vlmax` from `VLEN/EEW` to `VLEN/EEW*LMUL`(which is `VLMAX`), while ensuring that `vl <= VLMAX == VLEN/SEW*LMUL == micro_vlmax*LMUL`.

rogerchang23424:
> ```c++
> const uint32_t micro_vlmax = vlen / width_EEW(_machInst.width);
> int32_t micro_vl = (std::min(this->vl, micro_vlmax) + 7) / 8;
> ```

I think we cloud replace the `vlen /width_EEW(_machInst.width);` with `vlen / 8`. As the size of element for vlm.v and vsm.v are 8.

The `this->vl` is configure by `vset*vl*` instruction and it cloud be larger than the vlen for each registers accoring to the section 6.3 of RISC-V vector spec.

Compared the implementation of vle8.v and vlm.v from spike

diff vle8.v vlm.v

```diff
< // vle8_v
---
> //vlm_v
4c4
< const reg_t vl = false ? (((*p).VU.vl->read() + 7) / 8) : (*p).VU.vl->read();
---
> const reg_t vl = true ? (((*p).VU.vl->read() + 7) / 8) : (*p).VU.vl->read();
26,27c26,27
< reg_t veew = false ? 1 : sizeof(int8_t) * 8;
< float vemul = false ? 1 : ((float)veew / (*p).VU.vsew * (*p).VU.vflmul);
---
> reg_t veew = true ? 1 : sizeof(int8_t) * 8;
> float vemul = true ? 1 : ((float)veew / (*p).VU.vsew * (*p).VU.vflmul);
```

I think the parameters `vl`, `veew` and `vemul` will be critical part for handling load instructions.

Thanks for all

rogerchang23424:
> ```c++
> const uint32_t micro_vlmax = vlen / width_EEW(_machInst.width);
> int32_t micro_vl = (std::min(this->vl, micro_vlmax) + 7) / 8;
> ```

We cloud change to `std::min((this->vl + 7) / 8, micro_vlmax);` to have the `this->vl` guard, but I think is not required because of `this->vl <= vlen`.

I have tests the sample binary `rvv_strlen.c` and works.

Compared the implementation of vle8.v and vlm.v from spike

diff vle8.v vlm.v

```diff
< // vle8_v
---
> //vlm_v
4c4
< const reg_t vl = false ? (((*p).VU.vl->read() + 7) / 8) : (*p).VU.vl->read();
---
> const reg_t vl = true ? (((*p).VU.vl->read() + 7) / 8) : (*p).VU.vl->read();
26,27c26,27
< reg_t veew = false ? 1 : sizeof(int8_t) * 8;
< float vemul = false ? 1 : ((float)veew / (*p).VU.vsew * (*p).VU.vflmul);
---
> reg_t veew = true ? 1 : sizeof(int8_t) * 8;
> float vemul = true ? 1 : ((float)veew / (*p).VU.vsew * (*p).VU.vflmul);
```

I think the parameters `vl`, `veew` and `vemul` will be critical part for handling load instructions.

Thanks for all

powerjg:
I assume we should wait to merge until this change has been made. Let me know if I'm wrong.

hnpl:
My understanding is that the vlm.v and vsm.v instructions only use `vl`, and both do not use `EEW` directly (`vl` is derived from `EEW` anyway). So, I believe that using `EEW` in the implementation of those instructions is very misleading.

Having said that, I'm not opposed to keeping `micro_vlmax`, but I'd prefer removing `micro_vlmax` from the instructions.

BobbyRBruce (approved):

PR #831, opened 2024-02-01
misc: Merge develop .github dir into stable

BobbyRBruce:


powerjg:
Is there an easy way to add a description when you make merges like this which includes the information as to what's in the merge?

BobbyRBruce:
> Is there an easy way to add a description when you make merges like this which includes the information as to what's in the merge?

Not really, no. I just copy the directory. Literally `cp -r .github` from one branch to another. I don't think there's a better way. It's an unconventional way to use git and github, but we do development on our non-main branch which is just, annoyingly, unsupported in github when it comes to testing so we need to do it this way.

PR #835, opened 2024-02-01
sim: Updating Process::Map

mahyarsamani:
Changing size from int to int64_t to allow for mapping regions bigger than 2GB.

### DIFF
--- a/src/sim/process.cc
+++ b/src/sim/process.cc
@@ -401,7 +401,7 @@ Process::unserialize(CheckpointIn &cp)
 }
 
 bool
-Process::map(Addr vaddr, Addr paddr, int size, bool cacheable)
+Process::map(Addr vaddr, Addr paddr, int64_t size, bool cacheable)
 {
     pTable->map(vaddr, paddr, size,
                 cacheable ? EmulationPageTable::MappingFlags(0) :
--- a/src/sim/process.hh
+++ b/src/sim/process.hh
@@ -158,7 +158,7 @@ class Process : public SimObject
      * @return True if the map operation was successful.  (At this
      *           point in time, the map operation always succeeds.)
      */
-    bool map(Addr vaddr, Addr paddr, int size, bool cacheable = true);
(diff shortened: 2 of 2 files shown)


BobbyRBruce:
There's some horrific rebase erroring here as you're targeting stable instead of develop. Also, can you update the PR title to include tags?

hnpl (approved):
good catch!

src/sim/process.cc
```diff
@@ -401,7 +401,7 @@ Process::unserialize(CheckpointIn &cp)
 }
 
 bool
-Process::map(Addr vaddr, Addr paddr, int size, bool cacheable)
+Process::map(Addr vaddr, Addr paddr, int64_t size, bool cacheable)
```
hnpl, line 404:
```suggestion
Process::map(Addr vaddr, Addr paddr, Addr size, bool cacheable)
```

I believe it's safe to assume that the size should be unsigned, so we should use an unsigned integer here.

In other part of the code base, we are using Addr for the size of things, we can do it here.

src/sim/process.hh
```diff
@@ -158,7 +158,7 @@ class Process : public SimObject
      * @return True if the map operation was successful.  (At this
      *           point in time, the map operation always succeeds.)
      */
-    bool map(Addr vaddr, Addr paddr, int size, bool cacheable = true);
+    bool map(Addr vaddr, Addr paddr, int64_t size, bool cacheable = true);
```
hnpl, line 161:
```suggestion
    bool map(Addr vaddr, Addr paddr, Addr size, bool cacheable = true);
```

Ditto

hnpl (commented):
Good catch! There's a small suggestion about the type, but this lgtm.

src/sim/process.cc
powerjg: Seems reasonable, but if you look at the function that `Process::map` calls, it uses an `int64_t`, not `uint64_t`. Unless we change that function, I think this is safer.

hnpl: Agreed.

powerjg (approved):

PR #846, opened 2024-02-05
arch-riscv: Add BasePMAChecker to support customized PMA

rogerchang23424:
The RISC-V privilege spec don't specify the implementation of PMA(physical memory attribute), which is addressed in the previous CL[1].

This CL creates the BasePMAChecker to support customized PMA so that we can only focus on the features wanted in the study. The CL also leaves the common methods `check` and `takeOverFrom` to make MMU easy to interact with PMA.

[1] https://gem5-review.googlesource.com/c/public/gem5/+/40596

Change-Id: I9725e3a8f7f9276e41f0d06988259456149d2a77

### DIFF
--- a/src/arch/riscv/RiscvMMU.py
+++ b/src/arch/riscv/RiscvMMU.py
@@ -49,7 +49,7 @@ class RiscvMMU(BaseMMU):
 
     itb = RiscvTLB(entry_type="instruction")
     dtb = RiscvTLB(entry_type="data")
-    pma_checker = Param.PMAChecker(PMAChecker(), "PMA Checker")
+    pma_checker = Param.BasePMAChecker(PMAChecker(), "PMA Checker")
     pmp = Param.PMP(PMP(), "Physical Memory Protection Unit")
 
     @classmethod
--- a/src/arch/riscv/pma_checker.cc
+++ b/src/arch/riscv/pma_checker.cc
@@ -51,7 +51,7 @@ namespace RiscvISA
 {
 
 PMAChecker::PMAChecker(const Params &params) :
-SimObject(params),
+BasePMAChecker(params),
 uncacheable(params.uncacheable.begin(), params.uncacheable.end())
 {
 }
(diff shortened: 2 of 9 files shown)


src/arch/riscv/RiscvMMU.py
```diff
@@ -49,7 +49,7 @@ class RiscvMMU(BaseMMU):
 
     itb = RiscvTLB(entry_type="instruction")
     dtb = RiscvTLB(entry_type="data")
-    pma_checker = Param.PMAChecker(PMAChecker(), "PMA Checker")
+    pma_checker = Param.BasePMAChecker(PMAChecker(), "PMA Checker")
     pmp = Param.PMP(PMP(), "Physical Memory Protection Unit")
```
powerjg, line 53: General note, not a specific comment on the PR. There's no need to take any action on this.

I think we need to move away from the design where we instantiate SimObjects in these SimObject descriptions. It's super error prone and causes significant confusion when trying to find the "default" values for SimObject parameters.

src/arch/riscv/pma_checker.cc
```diff
@@ -50,8 +50,25 @@ namespace gem5
...
 
+BasePMAChecker::BasePMAChecker(const BasePMACheckerParams &params) :
+SimObject(params)
+{
+}
+
+void
+BasePMAChecker::check(const RequestPtr &req)
+{
+    panic("BasePMAChecker::check is not implemented");
```
powerjg: Any reasons not to make these functions pure virtual?

powerjg (commented):
One small question below. Otherwise, it looks good to me.

src/arch/riscv/pma_checker.cc
rogerchang23424: Changed to the pure virtual

src/arch/riscv/RiscvMMU.py
rogerchang23424: I think I will keep the default here to make the config script backward compatible, although the board specify the `pma_checker` of mmu with `PMAChecker` instance.

powerjg: Yeah, this make sense. At some point in the future, I would like to start fixing these things, though :)

powerjg (approved):

PR #848, opened 2024-02-05
arch-arm: Crypto instruction execution requires SIMD to be enabled

giactra:
Crypto instructions will cause an undefined instruction when executed with SIMD disabled. The PR is also
refactoring their implementation by checking the release object instead of the ID register field. This is improving
readability


andysan (approved):

powerjg (approved):

mattsinc:
@giactra : trying to do some crypto mining in gem5? ;)

(I know you probably meant cryptography)

PR #852, opened 2024-02-05
arch-vega,gpu-compute,mem-ruby: SQC Invalidation Support

v-ramadas:
This PR adds support for SQC (GPU I-cache) invalidation to the GPU model. It does this by updating the GPU-VIPER-SQC protocol to support flushes, the sequencer model to send out invalidates and the gpu compute model to send invalidates and handle responses. It also adds support for S_ICACHE_INV, a VEGA ISA instruction that invalidates the entire GPU I-cache. Additionally, the PR modifies the kernel start behavior to invalidate the I-cache too. It previously invalidated only the L1 D-cache.

### DIFF
--- a/src/gpu-compute/compute_unit.cc
+++ b/src/gpu-compute/compute_unit.cc
@@ -1085,11 +1085,9 @@ ComputeUnit::SQCPort::MemReqEvent::process()
     SenderState *sender_state = safe_cast<SenderState*>(pkt->senderState);
     [[maybe_unused]] ComputeUnit *compute_unit = sqcPort.computeUnit;
 
-    if (pkt->req->systemReq()) {
-        assert(compute_unit->shader->systemHub);
-        SystemHubEvent *resp_event = new SystemHubEvent(pkt, &sqcPort);
-        compute_unit->shader->systemHub->sendRequest(pkt, resp_event);
-    } else if (!(sqcPort.sendTimingReq(pkt))) {
+    assert(!pkt->req->systemReq());
+
+    if (!(sqcPort.sendTimingReq(pkt))) {
         sqcPort.retries.push_back(std::pair<PacketPtr, Wavefront*>
                 (pkt, sender_state->wavefront));
     }
--- a/src/gpu-compute/compute_unit.cc
+++ b/src/gpu-compute/compute_unit.cc
@@ -397,9 +397,9 @@ ComputeUnit::startWavefront(Wavefront *w, int waveId, LdsChunk *ldsChunk,
 }
 
 /**
- * trigger invalidate operation in the cu
+ * trigger invalidate operation in the CU
  *
- * req: request initialized in shader, carrying the invlidate flags
+ * req: request initialized in shader, carrying the invalidate flags
  */
 void
 ComputeUnit::doInvalidate(RequestPtr req, int kernId){
@@ -423,6 +423,26 @@ ComputeUnit::doInvalidate(RequestPtr req, int kernId){
 void
 ComputeUnit::doFlush(GPUDynInstPtr gpuDynInst) {
     injectGlobalMemFence(gpuDynInst, true);
+}
+
+/**
+ * trigger SQCinvalidate operation in the CU
+ *
+ * req: request initialized in shader, carrying the invalidate flags
+ */
+void
+ComputeUnit::doSQCInvalidate(RequestPtr req, int kernId){
+    GPUDynInstPtr gpuDynInst
+        = std::make_shared<GPUDynInst>(this, nullptr,
+            new KernelLaunchStaticInst(), getAndIncSeqNum());
+
+    // kern_id will be used in inv responses
+    gpuDynInst->kern_id = kernId;
+    // update contextId field
+    req->setContext(gpuDynInst->wfDynId);
+
+    gpuDynInst->staticInstruction()->setFlag(GPUStaticInst::Scalar);
+    scalarMemoryPipe.injectScalarMemFence(gpuDynInst, true, req);
 }
 
 // reseting SIMD register pools
@@ -1012,7 +1032,14 @@ ComputeUnit::DataPort::recvReqRetry()
 bool
 ComputeUnit::SQCPort::recvTimingResp(PacketPtr pkt)
 {
-    computeUnit->handleSQCReturn(pkt);
+    SenderState *sender_state = safe_cast<SenderState*>(pkt->senderState);
+    /** Process the response only if there is a wavefront associated with it.
+     * Otherwise, it is from SQC invalidate that was issued at kernel start
+     * and doesn't have a wavefront or instruction associated with it.
+     */
+    if (sender_state->wavefront != nullptr) {
+        computeUnit->handleSQCReturn(pkt);
+    }
 
     return true;
 }
--- a/src/gpu-compute/compute_unit.cc
+++ b/src/gpu-compute/compute_unit.cc
@@ -1046,6 +1046,28 @@ ComputeUnit::SQCPort::recvReqRetry()
     }
 }
 
+const char*
+ComputeUnit::SQCPort::MemReqEvent::description() const
+{
+    return "ComputeUnit SQC memory request event";
+}
+
+void
+ComputeUnit::SQCPort::MemReqEvent::process()
+{
+    SenderState *sender_state = safe_cast<SenderState*>(pkt->senderState);
+    [[maybe_unused]] ComputeUnit *compute_unit = sqcPort.computeUnit;
+
+    if (pkt->req->systemReq()) {
+        assert(compute_unit->shader->systemHub);
+        SystemHubEvent *resp_event = new SystemHubEvent(pkt, &sqcPort);
+        compute_unit->shader->systemHub->sendRequest(pkt, resp_event);
+    } else if (!(sqcPort.sendTimingReq(pkt))) {
+        sqcPort.retries.push_back(std::pair<PacketPtr, Wavefront*>
+                (pkt, sender_state->wavefront));
+    }
+}
+
 void
 ComputeUnit::sendRequest(GPUDynInstPtr gpuDynInst, PortID index, PacketPtr pkt)
 {
--- a/src/gpu-compute/compute_unit.hh
+++ b/src/gpu-compute/compute_unit.hh
@@ -680,6 +680,41 @@ class ComputeUnit : public ClockedObject
                 kernId(_kernId){ }
         };
 
+        class MemReqEvent : public Event
+        {
+          private:
+            SQCPort &sqcPort;
+            PacketPtr pkt;
+
+          public:
+            MemReqEvent(SQCPort &_sqc_port, PacketPtr _pkt)
+                : Event(), sqcPort(_sqc_port), pkt(_pkt)
+            {
+              setFlags(Event::AutoDelete);
+            }
+
+            void process();
+            const char *description() const;
+        };
+
+        class SystemHubEvent : public Event
+        {
+          SQCPort *sqcPort;
+          PacketPtr reqPkt;
+
+          public:
+            SystemHubEvent(PacketPtr pkt, SQCPort *_sqcPort)
+                : sqcPort(_sqcPort), reqPkt(pkt)
+            {
+                setFlags(Event::AutoDelete);
+            }
+
+            void
+            process()
+            {
+            }
+        };
+
         std::deque<std::pair<PacketPtr, Wavefront*>> retries;
 
       protected:
--- a/src/gpu-compute/compute_unit.hh
+++ b/src/gpu-compute/compute_unit.hh
@@ -698,24 +698,6 @@ class ComputeUnit : public ClockedObject
             const char *description() const;
         };
 
-        class SystemHubEvent : public Event
-        {
-          SQCPort *sqcPort;
-          PacketPtr reqPkt;
-
-          public:
-            SystemHubEvent(PacketPtr pkt, SQCPort *_sqcPort)
-                : sqcPort(_sqcPort), reqPkt(pkt)
-            {
-                setFlags(Event::AutoDelete);
-            }
-
-            void
-            process()
-            {
-            }
-        };
-
         std::deque<std::pair<PacketPtr, Wavefront*>> retries;
 
       protected:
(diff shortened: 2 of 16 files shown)


src/arch/amdgpu/vega/insts/instructions.cc
```diff
@@ -4818,6 +4818,10 @@ namespace VegaISA
     Inst_SOPP__S_ICACHE_INV::Inst_SOPP__S_ICACHE_INV(InFmt_SOPP *iFmt)
         : Inst_SOPP(iFmt, "s_icache_inv")
     {
+        setFlag(MemBarrier);
+        setFlag(GPUStaticInst::MemSync);
+        setFlag(MemSync);
+
```
mattsinc: extra line of space, remove

src/gpu-compute/fetch_unit.cc
```diff
@@ -388,6 +388,29 @@ FetchUnit::FetchBufDesc::flushBuf()
...
+    readPtr = bufStart;
+
+    for (int i = 0; i < fetchDepth; ++i) {
+        freeList.push_back(bufStart + i * cacheLineSize);
+    }
+
+    DPRINTF(GPUFetch, "WF[%d][%d]: Id%d Fetch dropped, flushing fetch "
+            "buffer\n", wavefront->simdId, wavefront->wfSlotId,
+            wavefront->wfDynId);
+
```
mattsinc: extra line of space, remove

src/gpu-compute/fetch_unit.cc
```diff
@@ -471,6 +494,13 @@ FetchUnit::FetchBufDesc::reserveBuf(Addr vaddr)
 void
 FetchUnit::FetchBufDesc::fetchDone(Addr vaddr)
 {
+    if (vaddr == 0) {
```
mattsinc: you should probably add a comment here explaining what vaddr == 0 represents.

A bigger debate is if we can really guarantee vaddr == 0 only for s_icache_inv, but for now probably ok.

src/gpu-compute/scalar_memory_pipeline.cc
```diff
@@ -160,4 +160,55 @@ ScalarMemPipeline::issueRequest(GPUDynInstPtr gpuDynInst)
...
+                0, 0, 0, computeUnit.requestorId(), 0, gpuDynInst->wfDynId);
+    } else {
+        req->requestorId(computeUnit.requestorId());
+    }
+
+    req->setPaddr(0);
+
+    PacketPtr pkt = nullptr;
+
+    if (kernelMemSync) {
```
mattsinc, line 188: add comment here explaining if/else

src/gpu-compute/scalar_memory_pipeline.cc
```diff
@@ -160,4 +160,55 @@ ScalarMemPipeline::issueRequest(GPUDynInstPtr gpuDynInst)
...
+    assert(gpuDynInst->isScalar());
+
+    if (!req) {
+        req = std::make_shared<Request>(
+                0, 0, 0, computeUnit.requestorId(), 0, gpuDynInst->wfDynId);
+    } else {
+        req->requestorId(computeUnit.requestorId());
+    }
+
+    req->setPaddr(0);
```
mattsinc, line 181: add comment explaining this

src/gpu-compute/scalar_memory_pipeline.cc
```diff
@@ -160,4 +160,55 @@ ScalarMemPipeline::issueRequest(GPUDynInstPtr gpuDynInst)
...
+        pkt = new Packet(req, MemCmd::MemSyncReq);
+        pkt->pushSenderState(
+                new ComputeUnit::SQCPort::SenderState(
+                    gpuDynInst->wavefront(), nullptr));
+
+        ComputeUnit::SQCPort::MemReqEvent *sqc_event =
+                new ComputeUnit::SQCPort::MemReqEvent
+                (computeUnit.sqcPort, pkt);
+
+        computeUnit.schedule(
```
mattsinc: probably the schedule can be after the if-else, since it's the same in both?  And maybe the sqc_event creation too?

src/gpu-compute/shader.cc
```diff
@@ -221,6 +221,11 @@ Shader::prepareInvalidate(HSAQueueEntry *task) {
         // all necessary INV flags are all set now, call cu to execute
         cuList[i_cu]->doInvalidate(req, task->dispatchId());
 
+
+        if ((i_cu % 4) == 0) {
```
mattsinc: Do we have a variable in the Python for this?  4 is a magic number because of how many CUs share an SQC.  To make this code more resilient we should really add/use such a variable to avoid problems later.

mattsinc: also add comment here explaining logic

src/mem/ruby/protocol/GPU_VIPER-SQC.sm
```diff
@@ -313,6 +320,11 @@ machine(MachineType:SQC, "GPU SQC (L1 I Cache)")
     APPEND_TRANSITION_COMMENT(cache_entry.DataBlk);
   }
 
+  action(inv_invDone, "inv", desc="local inv done") {
+    sequencer.invL1Callback();
+  }
+
```
mattsinc, line 326: extra line of space, remove

src/mem/ruby/system/Sequencer.cc
```diff
@@ -348,6 +350,10 @@ Sequencer::insertRequest(PacketPtr pkt, RubyRequestType primary_type,
         return RequestStatus_Ready;
     }
 
+    if (pkt->cmd == MemCmd::MemSyncReq) {
+        return RequestStatus_Aliased;
```
mattsinc, line 359: What is this for?

mattsinc (changes requested):
In general looks good, but some questions/comments/changes.  Nice job!

src/arch/amdgpu/vega/insts/instructions.cc
v-ramadas: Done

src/mem/ruby/protocol/GPU_VIPER-SQC.sm
v-ramadas: Done

src/gpu-compute/fetch_unit.cc
v-ramadas: Added comment

src/gpu-compute/scalar_memory_pipeline.cc
v-ramadas: Done

src/gpu-compute/scalar_memory_pipeline.cc
v-ramadas: Done

src/gpu-compute/scalar_memory_pipeline.cc
v-ramadas: Done

src/gpu-compute/shader.cc
v-ramadas: We have one in the top level config script but did not in the Shader/Compute Unit models. I added one and use it in this logic instead. It gets its value from the top level config script. I wanted to do this in my initial commit but completely forgot about it. Thanks for pointing it out.

v-ramadas: Done

src/mem/ruby/system/Sequencer.cc
v-ramadas: Added a comment explaining it

v-ramadas:
@mattsinc, thanks for reviewing! I have made the updates you suggested and added comments accordingly.

src/mem/ruby/protocol/MOESI_AMD_Base-CorePair.sm
```diff
@@ -575,9 +581,12 @@ machine(MachineType:CorePair, "CP-like Core Coherence")
                   if (in_msg.Type == RubyRequestType:LD) {
                     trigger(Event:C1_Load_L1miss, in_msg.LineAddress,
                             cache_entry, tbe);
-                  } else {
+                  } else if ((in_msg.Type == RubyRequestType:ST) || (in_msg.Type == RubyRequestType:ATOMIC)) {
```
mattsinc: What does this have to do with the SQC commit?

src/mem/ruby/protocol/MOESI_AMD_Base-CorePair.sm
```diff
@@ -613,9 +625,12 @@ machine(MachineType:CorePair, "CP-like Core Coherence")
                   if (in_msg.Type == RubyRequestType:LD) {
                     trigger(Event:C0_Load_L1miss, in_msg.LineAddress,
                         cache_entry, tbe);
-                  } else {
+                  } else if ((in_msg.Type == RubyRequestType:ST) || (in_msg.Type == RubyRequestType:ATOMIC)) {
```
mattsinc: same here

src/gpu-compute/shader.cc
mattsinc: Probably we'd need to update that script too to reflect this change, but it shouldn't be part of this patch

src/gpu-compute/compute_unit.cc
```diff
@@ -1042,6 +1069,28 @@ ComputeUnit::SQCPort::recvReqRetry()
...
+void
+ComputeUnit::SQCPort::MemReqEvent::process()
+{
+    SenderState *sender_state = safe_cast<SenderState*>(pkt->senderState);
+    [[maybe_unused]] ComputeUnit *compute_unit = sqcPort.computeUnit;
+
+    if (pkt->req->systemReq()) {
+        assert(compute_unit->shader->systemHub);
+        SystemHubEvent *resp_event = new SystemHubEvent(pkt, &sqcPort);
+        compute_unit->shader->systemHub->sendRequest(pkt, resp_event);
```
abmerop: I don't think this path is possible.  If it is, since the SystemHubEvent process() does nothing we will have requests silently dropped.  I would instead make this an assert that it is not a system request.  If we see the assert in the future we'll know exactly where to go back and fix it

src/gpu-compute/compute_unit.hh
```diff
@@ -674,6 +675,41 @@ class ComputeUnit : public ClockedObject
...
+            {
+                setFlags(Event::AutoDelete);
+            }
+
+            void
+            process()
+            {
+            }
+        };
+
```
abmerop: Probably not needed due to above comment

src/gpu-compute/fetch_unit.cc
```diff
@@ -388,6 +388,28 @@ FetchUnit::FetchBufDesc::flushBuf()
...
+
+    for (int i = 0; i < fetchDepth; ++i) {
+        freeList.push_back(bufStart + i * cacheLineSize);
+    }
+
+    DPRINTF(GPUFetch, "WF[%d][%d]: Id%d Fetch dropped, flushing fetch "
+            "buffer\n", wavefront->simdId, wavefront->wfSlotId,
+            wavefront->wfDynId);
+}
+
```
abmerop: This seems to be exactly the same as flushBuf(), is there any reason not to use that?  I realize the terms have different meanings, but for a read-only buffer they are functionally the same

src/gpu-compute/fetch_unit.cc
```diff
@@ -471,6 +493,18 @@ FetchUnit::FetchBufDesc::reserveBuf(Addr vaddr)
 void
 FetchUnit::FetchBufDesc::fetchDone(Addr vaddr)
 {
+    // If the return vaddr is 0, then it belongs to an SQC invalidation
+    // request. This request calls incLGKMInstsIssued() function in its
+    // execution path. Since there is no valid memory return response
+    // associated with this instruction, decLGKMInstsIssued() is not
+    // executed. Do this here to decrement the counter and invalidate
+    // all buffers
+    if (vaddr == 0) {
```
abmerop: This method seems to be called in one place only which has access to the packet.  I think it would be safer to change the method to pass the packet, then here it could check if it is a MemSyncResp (or whatever SQC invalidate response is)

src/gpu-compute/compute_unit.cc
v-ramadas: Thanks for letting me know. I was not sure about this part earlier. I have now changed this to an assert and the code still works

src/gpu-compute/compute_unit.hh
v-ramadas: Removed

src/gpu-compute/fetch_unit.cc
v-ramadas: I updated the function to use the packet command instead. Thanks for the suggestion!

src/mem/ruby/protocol/MOESI_AMD_Base-CorePair.sm
v-ramadas: I incorrectly added this file. These changes were in my stash and I applied it wrongly added it when making review changes. Removed this from the commit now.

These changes are from when I added Flush support for L1,L2. Since GPUFS uses KVM instead of the CorePair, I didn't get to test this. Haven't made a commit for these changes because of this reason

abmerop (dismissed):

src/gpu-compute/fetch_unit.cc
mattsinc: good call, @abmerop !

mattsinc (dismissed):

src/arch/amdgpu/vega/insts/instructions.cc
abmerop, line 1: Forgot this file no longer exists.  It would now go in the sopp.cc file

mattsinc: good point @abmerop , @v-ramadas you must not have updated to the head of develop before making your changes?

v-ramadas: I was at the head of develop when I first made the branch. This was before the commit to break up instructions.cc into multiple files was merged though. I will update the PR with this change now

v-ramadas:
@mattsinc @abmerop, I rebased my branch to the top of develop and wrote the S_ICACHE_INV behavior in sopp.cc. I also removed the older commit that was causing the merge issue. I've pushed the changes to this PR after verifying my test still works.

abmerop (approved):

mattsinc (approved):

PR #857, opened 2024-02-06
mem-cache: Fix circular dependency in QoS mem

hristobelchev:
This PR removes a circular dependency between `QoSMemSinkCtrl` and `QoSMemSinkInterface` that prevented the `controller()` function of `QoSMemSinkInterface` from being used by removing the default value for `QoSMemSinkCtrl.interface`.

Change-Id: I4ecc39b974e239be1a2e9285e1f6f8ea873c018d

### DIFF
--- a/src/mem/qos/QoSMemSinkCtrl.py
+++ b/src/mem/qos/QoSMemSinkCtrl.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2018-2020 ARM Limited
+# Copyright (c) 2018-2020,2024 ARM Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
@@ -36,7 +36,6 @@
 # Author: Matteo Andreozzi
 
 from m5.objects.QoSMemCtrl import *
-from m5.objects.QoSMemSinkInterface import *
 from m5.params import *
 
 
@@ -46,9 +45,7 @@ class QoSMemSinkCtrl(QoSMemCtrl):
     cxx_class = "gem5::memory::qos::MemSinkCtrl"
     port = ResponsePort("Response ports")
 
-    interface = Param.QoSMemSinkInterface(
-        QoSMemSinkInterface(), "Interface to memory"
-    )
+    interface = Param.QoSMemSinkInterface("Interface to memory")
 
     # the basic configuration of the controller architecture, note
     # that each entry corresponds to a burst for the specific DRAM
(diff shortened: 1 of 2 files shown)


src/mem/qos/QoSMemSinkCtrl.py
```diff
@@ -46,9 +45,9 @@ class QoSMemSinkCtrl(QoSMemCtrl):
     cxx_class = "gem5::memory::qos::MemSinkCtrl"
     port = ResponsePort("Response ports")
 
-    interface = Param.QoSMemSinkInterface(
-        QoSMemSinkInterface(), "Interface to memory"
-    )
```
powerjg, line 51: Correct me if I'm wrong, but I'm pretty sure the code below would work.

```
interface = Param.QoSMemSinkInterface("Interface to memory")
```

This is not a circular dependency because it is in `params` not `m5.objects.QoSMemSinkInterface`.

Generally, problems like this are why we should not instantiate SimObjects as default values for parameters in the SimObject description files.

Let me know if I'm wrong, though.

powerjg (changes requested):

src/mem/qos/QoSMemSinkCtrl.py
hristobelchev: Hi @powerjg, I tested your suggestion and it indeed worked - I've updated the commit to feature it. 

I thought it was originally a circular dependency problem as:

1. `QoSMemSinkInterface.controller()` was instantiating `QoSMemSinkCtrl` (so `QoSMemSinkInterface.py` needed to import `QoSMemSinkCtrl.py`)
2. `QoSMemSinkCtrl()` was instantiating `QoSMemSinkInterface` as a default param (so `QoSMemSinkCtrl.py` needed to import `QoSMemSinkInterface.py`)

I agree with the reasoning that SimObjects as default values for parameters can be problematic, especially when they are coupled like in this case.

src/mem/qos/mem_sink.hh
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2020 ARM Limited
+ * Copyright (c) 2018-2020,2024 ARM Limited
```
powerjg: You can remove this change to the copyright

src/mem/qos/mem_sink.cc
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2020 ARM Limited
+ * Copyright (c) 2018-2020,2024 ARM Limited
```
powerjg: You can remove this change to the copyright

powerjg (commented):
I think you don't need the copyright updates as noted below. Otherwise, it looks good to me.

powerjg (approved):

odanrc (approved):

PR #859, opened 2024-02-07
arch-riscv: Fix load and store to use EEW instead of SEW

QQeg:
Vector unit-stride instructions have an EEW encoded directly in the instruction, We should use that instead of SEW in vtype.

Change-Id: I282041ce8ed57fbcca899f7497ef6c6fb2dfcf85

Ref:
https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#73-vector-loadstore-width-encoding

### DIFF
--- a/src/arch/riscv/isa/templates/vector_mem.isa
+++ b/src/arch/riscv/isa/templates/vector_mem.isa
@@ -172,11 +172,10 @@ Fault
         return fault;
 
     const size_t micro_vlmax = vlen / width_EEW(machInst.width);
-    const size_t micro_elems = vlen / width_EEW(machInst.width);
 
     size_t ei;
 
-    for (size_t i = 0; i < micro_elems; i++) {
+    for (size_t i = 0; i < micro_vlmax; i++) {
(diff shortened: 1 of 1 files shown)


src/arch/riscv/isa/templates/vector_mem.isa
```diff
@@ -171,7 +171,7 @@ Fault
     if (fault != NoFault)
         return fault;
 
-    const size_t micro_vlmax = vtype_VLMAX(machInst.vtype8, vlen, true);
+    const size_t micro_vlmax = vlen / width_EEW(machInst.width);
     const size_t micro_elems = vlen / width_EEW(machInst.width);
```
rogerchang23424: I think we can remove the `micro_elems` and only leave `micro_vlmax`

QQeg: Sure, I've pushed a new commit.

rogerchang23424 (approved):

BobbyRBruce (approved):

PR #861, opened 2024-02-09
util: update list_changes.py to support multiple Change-Ids

rjc-arch:
The original version of `list_changes.py` assumed no more than one `Change-Id` tag per commit. However, since transitioning to GitHub, the repository now contains some merge commits containing multiple `Change-Id`s.

This patch updates `list_changes.py` to support commits with any number of `Change-Id` tags.

### DIFF
--- a/util/maint/list_changes.py
+++ b/util/maint/list_changes.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 #
-# Copyright (c) 2017-2018 Arm Limited
+# Copyright (c) 2017-2018, 2024 Arm Limited
 # All rights reserved
 #
 # The license below extends only to copyright in the software and shall
@@ -88,20 +88,20 @@ class Commit:
         return self._tags
 
     @property
-    def change_id(self):
-        """Get the Change-Id tag from the commit
+    def change_ids(self):
+        """Get the Change-Id tag(s) from the commit
 
-        Returns: A change ID or None if no change ID has been
-        specified.
+        Returns: A list of change IDs. May be empty if no change IDs
+        have been specified.
 
         """
+
         try:
             cids = self.tags["Change-Id"]
         except KeyError:
-            return None
+            return list()
 
-        assert len(cids) == 1
-        return cids[0]
+        return cids[:]
 
     def __str__(self):
         return f"{self.rev[0:8]}: {self.log[0]}"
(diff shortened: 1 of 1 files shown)


andysan (approved):

powerjg (approved):
How much would it hurt your workflow if we dropped the *requirement* for change ids?

rjc-arch:
Hi @powerjg. We could work around dropping the requirement, although the Change-Ids are nice to have. We will continue adding them on our side at least. 

My understanding was that we'd already agreed to drop that requirement, and @BobbyRBruce was going to look into adding an optional-but-recommended alternative tag as a pre-commit step that we could use instead (with the understanding that not having the tag would not block a commit). Having said that, I think we could still work around having untagged commits.

BobbyRBruce:
> Hi @powerjg. We could work around dropping the requirement, although the Change-Ids are nice to have. We will continue adding them on our side at least.
> 
> My understanding was that we'd already agreed to drop that requirement, and @BobbyRBruce was going to look into adding an optional-but-recommended alternative tag as a pre-commit step that we could use instead (with the understanding that not having the tag would not block a commit). Having said that, I think we could still work around having untagged commits.

I feel like we _kind of_ agreed to drop the requirement but there some kickback as people still had a desire to map commits. A reasonably good example of this is when commits are cherry-picked from one branch to another. The commit hash changes and there's no official way to say where or even if a commit was cherry-picked from elsewhere.

I have thought about an alternative tag but I can't find a solution that wouldn't be messy. What I'd love is if commit messages could automatically have the PR ID appended, something like `gem5 Github #861 , Commit 2` (meaning the commit was added in GitHub PR #861 and was the 2nd commit in the chain). The issue here is there's no real way to automate this. 

Right now i find change IDs to be quite irritating to manage and offer rather low benefit. It gets even more complicated now that we allow squash-merges as well as traditional merges. As such the de-facto policy is that you can add them, we provide tooling for you to include them, but we don't really 'support' beyond that.

rjc-arch:
@BobbyRBruce it would be great if we had a way to link the commits back to a PR (and more importantly the associated discussion), but I understand that would be very hard to automate.

The use case you described (cherry-picking) is the main one I am interested in, but I think we can make it work with Change-Ids only in our own commits.

Issue #827, opened 2024-02-01
RISCV RVV vfredmax.vs and vfredmin.vs implementation is incorrect


debjyoti0891:
**Describe the bug**
The implementation of RISCV RVV vfredmax.vs and vfredmin.vs is incorrect. 
- The expected behavior for min(3, NaN) = 3.  Similarly, max(3,NaN) = 3. For vfredmax and vfredmin, if any of the elements in the two source vectors is not-a-NaN, the output should be the max or min of not-a-NaN values.  
- This behaviour is demonstrated by spike as well as qemu. In gem5, this is not true. 

**Affects version**
Branch: develop Commit: ce32d7c523ae5c9612dfd6242092a3630e301898

**To Reproduce**
Test code is shown below. 
- The code loads a quiet NaN (0x7fc00000) into `v9[0]` and splats a constant 1.0 into vector register `v8` and the does a max reduction.
- The result of the reduction in `ft0` is checked as a NaN using ```feq.s   a1,ft0,ft0 \n\```. For gem5, this evaluates to True, *which is an error*.
- For the second part of the code, constant 34.5 is loaded in `ft0` and a max is performed using `fmax.s` between  `ft1` and `ft0`. this results correctly in 34.5 as the value stored in `ft2`.

```
int main(){
    float ret = 123.345,misc = 234.456;
    __asm__ volatile("lui a1,0x7fc00 \n\
        vsetivli    zero,8,e32,m1,ta,ma \n\
        vmv.s.x v9,a1 \n\
        lui a1,0x3f800 \n\
        vmv.v.x v8,a1 \n\
        vfredmax.vs v9,v8,v9 \n\
        vfmv.f.s    ft0,v9 \n\
        feq.s   a1,ft0,ft0 \n\
        beqz a1,lab1 \n\
        lui  a1,0x42090 \n\
        vmv.s.x  v9,a1 \n\
        vfredmax.vs  v8,v8,v9 \n\
        vfmv.f.s ft0,v8 \n\
        j    lab2 \n\
    lab1:\n\
        lui  a1,0x7fc00 \n\
        fmv.w.x  ft0,a1 \n\
    lab2:\n\
        lui  a1,0x420A0 \n\
        fmv.w.x  ft1,a1\n\
        fmax.s  ft2,ft1,ft0\n\
        fmv.s  %0,ft0\n\
        fmv.s  %1,ft2\n\
    " : "=f"(ret),
        "=f"(misc));
[...]

PR #863, opened 2024-02-12
arch-riscv: This commit fixes bug in vfmv.f.s impl. in riscv

debjyoti0891:
The existing implementation of vfmv instruction did not type cast the first element of the source vector, which caused the "freg" to interpret the result as a NaN.

With the type cast to f32, the value is correctly recognized as float and sign extended to be stored in the fd register.

Git issue: https://github.com/gem5/gem5/issues/827

Change-Id: Ibe9873910827594c0ec11cb51ac0438428c3b54e


hnpl (commented):
Hi, thanks for the fix.

I wonder if we should always cast the value to float32. My understanding is that the values in a vector register could be a set of float16 values or a set of float64 values too. Maybe we need to obtain the type from VTYPE.

~However, my understanding of vfmv.v.s is that it should be floating point type agnostic as we are merely copying bits from 1 floating point reg to a vector reg.~

According to https://github.com/riscv/riscv-v-spec/issues/339, we'll need to check for NaN if SEW != FLEN.

debjyoti0891:
Hi @hnpl ,

Thanks for the feedback. Indeed, the cast should be based on the SEW. I am curious if I could use the [spike reference implementation](https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/insns/vfmv_s_f.h) as a baseline.

For NaN box, I am curious about the interpretation. 
- Do I check if the value is NaN in element 0 for each datatype (fp8, fp16, fp32,fp64)? 
- If this is true, the result stored in the floating point register is a canonical nan (7fc00000).

src/arch/riscv/isa/decoder.isa
```diff
@@ -2746,7 +2746,7 @@ decode QUADRANT default Unknown::unknown() {
                         // The encodings corresponding to the masked versions
                         // (vm=0) of vfmv.f.s are reserved
                         0x1: VectorNonSplitFormat::vfmv_f_s({{
-                            freg_t fd = freg(Vs2_vu[0]);
+                            freg_t fd = freg(f32(Vs2_vu[0]));
```
rogerchang23424: From `template/vector_aarith.isa`

```
def template VectorFloatNonSplitDecodeBlock {{

switch(machInst.vtype8.vsew) {
case 0b010: return new %(class_name)s<float32_t>(machInst);
case 0b011: return new %(class_name)s<float64_t>(machInst);
default: GEM5_UNREACHABLE;
}
```

I think we also need to check the type of `vu` before casting the `f32` and `f64`. We can assume the `FLEN = 64` as we don't have `FLEN` configuration.

ivanaamit:
@debjyoti0891, can you address Roger's comment (https://github.com/gem5/gem5/pull/863#discussion_r1488792013)? We would like to merge this PR. Thank you.

debjyoti0891:
> @debjyoti0891, can you address Roger's comment ([#863 (comment)](https://github.com/gem5/gem5/pull/863#discussion_r1488792013))? We would like to merge this PR. Thank you.

surely, I plan to wrap this up in the next week.

ivanaamit:
@debjyoti0891, a friendly reminder to address pending comment. Thanks.

rogerchang23424 (approved):
Thanks for updating

powerjg (approved):

debjyoti0891:
@ivanaamit , a friendly reminder to merge the PR

PR #864, opened 2024-02-12
stdlib: add socks proxy to atlas client

Harshil2107:
Change-Id: I683e329b4a915fa8e767ee7585eb70c5bf8d1439

### DIFF
--- a/src/python/SConscript
+++ b/src/python/SConscript
@@ -314,7 +314,6 @@ PySource('gem5.utils.multiprocessing',
 PySource('gem5.utils.multiprocessing',
     'gem5/utils/multiprocessing/popen_spawn_gem5.py')
 
-
 PySource('', 'importer.py')
 PySource('m5', 'm5/__init__.py')
 PySource('m5', 'm5/SimObject.py')
--- a/src/python/SConscript
+++ b/src/python/SConscript
@@ -303,6 +303,8 @@ PySource('gem5.utils', 'gem5/utils/filelock.py')
 PySource('gem5.utils', 'gem5/utils/override.py')
 PySource('gem5.utils', 'gem5/utils/progress_bar.py')
 PySource('gem5.utils', 'gem5/utils/requires.py')
+PySource('gem5.utils',
+         'gem5/utils/socks_ssl_context.py')
 PySource('gem5.utils.multiprocessing',
     'gem5/utils/multiprocessing/__init__.py')
 PySource('gem5.utils.multiprocessing',
@@ -311,6 +313,7 @@ PySource('gem5.utils.multiprocessing',
     'gem5/utils/multiprocessing/context.py')
 PySource('gem5.utils.multiprocessing',
     'gem5/utils/multiprocessing/popen_spawn_gem5.py')
+
 
 PySource('', 'importer.py')
 PySource('m5', 'm5/__init__.py')
--- a/src/python/gem5/resources/client.py
+++ b/src/python/gem5/resources/client.py
@@ -152,7 +152,6 @@ def get_resource_json_obj(
     resource_version: Optional[str] = None,
     clients: Optional[List[str]] = None,
     gem5_version: Optional[str] = core.gem5Version,
-    proxy_context: Optional[ssl.SSLContext] = None,
 ) -> Dict:
     """
     Get the resource json object from the clients wrapper.
@@ -164,10 +163,8 @@ def get_resource_json_obj(
                          compatibility. By default, it is the gem5 version of the
                          current build. If ``None``, filtering based on compatibility
                          is not performed.
-    :param proxy_context: The SOCKS proxy context to use for the request.
-                          'None' if no proxy is to be used.
     """
 
     return _get_clientwrapper().get_resource_json_obj_from_client(
-        resource_id, resource_version, clients, gem5_version, proxy_context
+        resource_id, resource_version, clients, gem5_version
     )
--- a/src/python/gem5/resources/client.py
+++ b/src/python/gem5/resources/client.py
@@ -26,7 +26,6 @@
 
 import json
 import os
-import ssl
 from pathlib import Path
 from typing import (
     Dict,
--- a/src/python/gem5/resources/client.py
+++ b/src/python/gem5/resources/client.py
@@ -26,6 +26,7 @@
 
 import json
 import os
+import ssl
 from pathlib import Path
 from typing import (
     Dict,
@@ -151,6 +152,7 @@ def get_resource_json_obj(
     resource_version: Optional[str] = None,
     clients: Optional[List[str]] = None,
     gem5_version: Optional[str] = core.gem5Version,
+    proxy_context: Optional[ssl.SSLContext] = None,
 ) -> Dict:
     """
     Get the resource json object from the clients wrapper.
@@ -162,8 +164,10 @@ def get_resource_json_obj(
                          compatibility. By default, it is the gem5 version of the
                          current build. If ``None``, filtering based on compatibility
                          is not performed.
+    :param proxy_context: The SOCKS proxy context to use for the request.
+                          'None' if no proxy is to be used.
     """
 
     return _get_clientwrapper().get_resource_json_obj_from_client(
-        resource_id, resource_version, clients, gem5_version
+        resource_id, resource_version, clients, gem5_version, proxy_context
     )
--- a/src/python/gem5/resources/client_api/abstract_client.py
+++ b/src/python/gem5/resources/client_api/abstract_client.py
@@ -24,6 +24,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+import ssl
 import urllib.parse
 from abc import (
     ABC,
@@ -58,6 +59,7 @@ class AbstractClient(ABC):
         resource_id: Optional[str] = None,
         resource_version: Optional[str] = None,
         gem5_version: Optional[str] = None,
+        proxy_context: Optional[ssl.SSLContext] = None,
     ) -> List[Dict[str, Any]]:
         """
         :param resource_id: The ID of the Resource. Optional, if not set, all
@@ -68,6 +70,8 @@ class AbstractClient(ABC):
                                  parameter will be ignored.
         :param gem5_version: The version of gem5. Optional, if not set, all
                              versions will be returned.
+        :param proxy_context: The proxy context to be used for the request.
+                              'None' if no proxy is to be used.
         :return: A list of all the Resources with the given ID.
         """
         raise NotImplementedError
@@ -111,10 +115,15 @@ class AbstractClient(ABC):
                     filtered_resources.append(resource)
         return filtered_resources
 
-    def get_resources_by_id(self, resource_id: str) -> List[Dict[str, Any]]:
+    def get_resources_by_id(
+        self, resource_id: str, proxy_context: object
+    ) -> List[Dict[str, Any]]:
         """
         :param resource_id: The ID of the Resource.
-
+        :param proxy_context: The proxy context to be used for the request.
+                              'None' if no proxy is to be used.
         :return: A list of all the Resources with the given ID.
         """
-        return self.get_resources(resource_id=resource_id)
+        return self.get_resources(
+            resource_id=resource_id, proxy_context=proxy_context
+        )
--- a/src/python/gem5/resources/client_api/abstract_client.py
+++ b/src/python/gem5/resources/client_api/abstract_client.py
@@ -70,8 +70,6 @@ class AbstractClient(ABC):
                                  parameter will be ignored.
         :param gem5_version: The version of gem5. Optional, if not set, all
                              versions will be returned.
-        :param proxy_context: The proxy context to be used for the request.
-                              'None' if no proxy is to be used.
         :return: A list of all the Resources with the given ID.
         """
         raise NotImplementedError
@@ -115,15 +113,11 @@ class AbstractClient(ABC):
                     filtered_resources.append(resource)
         return filtered_resources
 
-    def get_resources_by_id(
-        self, resource_id: str, proxy_context: object
-    ) -> List[Dict[str, Any]]:
+    def get_resources_by_id(self, resource_id: str) -> List[Dict[str, Any]]:
         """
         :param resource_id: The ID of the Resource.
-        :param proxy_context: The proxy context to be used for the request.
-                              'None' if no proxy is to be used.
         :return: A list of all the Resources with the given ID.
         """
         return self.get_resources(
-            resource_id=resource_id, proxy_context=proxy_context
+            resource_id=resource_id,
         )
--- a/src/python/gem5/resources/client_api/abstract_client.py
+++ b/src/python/gem5/resources/client_api/abstract_client.py
@@ -114,6 +114,7 @@ class AbstractClient(ABC):
     def get_resources_by_id(self, resource_id: str) -> List[Dict[str, Any]]:
         """
         :param resource_id: The ID of the Resource.
+
         :return: A list of all the Resources with the given ID.
         """
         return self.get_resources(resource_id=resource_id)
--- a/src/python/gem5/resources/client_api/abstract_client.py
+++ b/src/python/gem5/resources/client_api/abstract_client.py
@@ -24,7 +24,6 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-import ssl
 import urllib.parse
 from abc import (
     ABC,
@@ -59,7 +58,6 @@ class AbstractClient(ABC):
         resource_id: Optional[str] = None,
         resource_version: Optional[str] = None,
         gem5_version: Optional[str] = None,
-        proxy_context: Optional[ssl.SSLContext] = None,
     ) -> List[Dict[str, Any]]:
         """
         :param resource_id: The ID of the Resource. Optional, if not set, all
@@ -118,6 +116,4 @@ class AbstractClient(ABC):
         :param resource_id: The ID of the Resource.
         :return: A list of all the Resources with the given ID.
         """
-        return self.get_resources(
-            resource_id=resource_id,
-        )
+        return self.get_resources(resource_id=resource_id)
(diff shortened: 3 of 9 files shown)


src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -125,6 +126,27 @@ def _atlas_http_json_req(
...
+            # If the "use_proxy" variable is specified we setup a socks5
+            # connection.
+
+            import socket
+            import ssl
+
+            import socks
+
+            IP_ADDR, host_port = use_proxy.split(":")
+            PORT = int(host_port)
```
powerjg: Why is `IP_ADDR` and `PORT` capitalized?

src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -125,6 +126,27 @@ def _atlas_http_json_req(
         """
         data = json.dumps(data_json).encode("utf-8")
 
+        # check to see if user requests a proxy connection
+        use_proxy = os.getenv("GEM5_USE_PROXY")
+        if use_proxy:
+            # If the "use_proxy" variable is specified we setup a socks5
+            # connection.
```
powerjg: Does it make sense to split this out into a function? Is this duplicated code from the other place you check proxies?

Harshil2107: Yes, I was thinking of moving this to a function that just returns the context as we use this context twice first in atlasclient and then in downloader.py.

src/python/gem5/resources/client_api/atlasclient.py
Harshil2107: It was capitalized in the original code so I did not change it.

src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -133,7 +136,10 @@ def _atlas_http_json_req(
 
         for attempt in itertools.count(start=1):
             try:
-                response = request.urlopen(req)
+                if proxy_context is not None:
+                    response = request.urlopen(req, context=proxy_context)
+                else:
+                    response = request.urlopen(req)
```
powerjg: This if statement is unneeded. You can pass `None` as the context.

```suggestion
                response = request.urlopen(req, context=proxy_context)
```

src/python/gem5/resources/downloader.py
```diff
@@ -61,7 +61,12 @@
 """
 
 
-def _download(url: str, download_to: str, max_attempts: int = 6) -> None:
+def _download(
+    url: str,
+    download_to: str,
+    max_attempts: int = 6,
+    proxy_context: object = None,
```
powerjg: Is that type right?

I think it's `Optional[ssl.SSLContext]`

src/python/gem5/resources/downloader.py
```diff
@@ -181,6 +172,27 @@ def _download(url: str, download_to: str, max_attempts: int = 6) -> None:
...
+def get_proxy_context():
+    use_proxy = os.getenv("GEM5_USE_PROXY")
+    if use_proxy:
+        import socket
+        import ssl
+
+        import socks
+
+        IP_ADDR, host_port = use_proxy.split(":")
+        PORT = int(host_port)
```
powerjg: Why are these all caps?

src/python/gem5/resources/downloader.py
```diff
@@ -215,6 +227,7 @@ def get_resource(
     clients: Optional[List] = None,
     gem5_version: Optional[str] = core.gem5Version,
     quiet: bool = False,
+    proxy_context: Optional[object] = None,
```
powerjg: I'm pretty sure the type is https://docs.python.org/3/library/ssl.html#ssl.SSLContext

src/python/gem5/resources/client_api/jsonclient.py
```diff
@@ -78,6 +78,7 @@ def get_resources(
         resource_id: Optional[str] = None,
         resource_version: Optional[str] = None,
         gem5_version: Optional[str] = None,
+        proxy_context: Optional[object] = None,
```
powerjg: Type is not object

src/python/gem5/resources/client_api/client_wrapper.py
```diff
@@ -104,6 +104,7 @@ def get_all_resources_by_id(
         self,
         resource_id: str,
         clients: Optional[List[str]] = None,
+        proxy_context: Optional[object] = None,
```
powerjg: Type is not object

src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -158,6 +164,7 @@ def get_resources(
         resource_id: Optional[str] = None,
         resource_version: Optional[str] = None,
         gem5_version: Optional[str] = None,
+        proxy_context: Optional[object] = None,
```
powerjg: Type is not object

src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -102,6 +103,7 @@ def _atlas_http_json_req(
         data_json: Dict[str, Any],
         headers: Dict[str, str],
         purpose_of_request: Optional[str],
+        proxy_context: Optional[object] = None,
```
powerjg: Type should be https://docs.python.org/3/library/ssl.html#ssl.SSLContext, I think

src/python/gem5/resources/client_api/abstract_client.py
```diff
@@ -111,10 +114,15 @@ def filter_incompatible_resources(
                     filtered_resources.append(resource)
         return filtered_resources
 
-    def get_resources_by_id(self, resource_id: str) -> List[Dict[str, Any]]:
+    def get_resources_by_id(
+        self, resource_id: str, proxy_context: object
```
powerjg: I don't think the type is correct

src/python/gem5/resources/client_api/abstract_client.py
```diff
@@ -58,6 +58,7 @@ def get_resources(
         resource_id: Optional[str] = None,
         resource_version: Optional[str] = None,
         gem5_version: Optional[str] = None,
+        proxy_context: Optional[object] = None,
```
powerjg: Type is not correct

src/python/gem5/resources/client.py
```diff
@@ -151,6 +151,7 @@ def get_resource_json_obj(
     resource_version: Optional[str] = None,
     clients: Optional[List[str]] = None,
     gem5_version: Optional[str] = core.gem5Version,
+    proxy_context: Optional[object] = None,
```
powerjg: I don't think the type is correct. I believe it should be https://docs.python.org/3/library/ssl.html#ssl.SSLContext

src/python/gem5/resources/downloader.py
Harshil2107: They were caps in the original code, I will change them to lower case.

src/python/gem5/resources/resource.py
```diff
@@ -963,13 +966,14 @@ def obtain_resource(
                     ``resource_directory`` parameter.
     :param quiet: If ``True``, suppress output. ``False`` by default.
     """
-
+    proxy_context = get_proxy_context()
```
BobbyRBruce: Why do we need to get the proxy_context here? Can't we get this in the `get_resource_json` function? Can't we just call this `get_proxy_context` function when it's needed? This approach requires changing quite a lot of function signatures and passing data around.

BobbyRBruce (changes requested):

src/python/gem5/resources/resource.py
Harshil2107: I did this because we only want to change the context once and then pas the same context around.

src/python/gem5/resources/downloader.py
```diff
@@ -181,6 +173,26 @@ def _download(url: str, download_to: str, max_attempts: int = 6) -> None:
             )
 
 
+def get_proxy_context():
```
BobbyRBruce: Please put a return type here

src/python/gem5/resources/resource.py
BobbyRBruce: If you only want to ever run this function once to get the data then I think there's an easier way. What about the following?

```py
# In downloader.py

_gem5_ssl_context = None
def get_proxy_context() -> TYPE_HERE:
    global _gem5_ssl_context
    use_proxy = os.getenv("GEM5_USE_PROXY")
    if use_proxy and not _gem5_ssl_context:
        import socket

        import socks

        ip_addr, host_port = use_proxy.split(":")
        port = int(host_port)
        socks.set_default_proxy(socks.SOCKS5, ip_addr, port)
        socket.socket = socks.socksocket

        # base SSL context for https connection
        _gem5_ssl_context = ssl.create_default_context()
        ctx.check_hostname = False
        ctx.verify_mode = ssl.CERT_NONE
    return gem5_ssl_context
```

With this we'd only need to call `get_proxy_context` at the point of creating the connection. The setting of the default proxy has once and we store the ssl_context to be returned everytime.

BobbyRBruce (changes requested):

src/python/gem5/resources/resource.py
Harshil2107: This causes an import error due to circular imports, as the downloader.py needs to access a function from client.py and client.py imports the AltasClient which in turn import get_proxy_context() from downloader.py.

Harshil2107:
I implemented the suggestion you mentioned, and it causes circular imports. I think we can maybe move this context function to abstract_client.py and that should solve this issue.

Harshil2107:
This should fix the circular import error.

src/python/gem5/utils/socks_ssl_context.py
BobbyRBruce, line 1: You need to add copyright info.

src/python/gem5/utils/socks_ssl_context.py
```diff
@@ -0,0 +1,25 @@
+import os
+import ssl
+from typing import Optional
+
+_gem5_ssl_context = None
+
+
+def get_proxy_context() -> Optional[ssl.SSLContext]:
+    global _gem5_ssl_context
```
BobbyRBruce, line 42: Can you add some docstring here?

src/python/gem5/resources/downloader.py
```diff
@@ -181,6 +170,29 @@ def _download(url: str, download_to: str, max_attempts: int = 6) -> None:
...
+        socks.set_default_proxy(socks.SOCKS5, ip_addr, port)
+        socket.socket = socks.socksocket
+
+        # base SSL context for https connection
+        _gem5_ssl_context = ssl.create_default_context()
+        _gem5_ssl_context.check_hostname = False
+        _gem5_ssl_context.verify_mode = ssl.CERT_NONE
+    return _gem5_ssl_context
+
+
```
BobbyRBruce: This is no longer needed.

src/python/gem5/resources/downloader.py
```diff
@@ -79,38 +85,21 @@ def _download(url: str, download_to: str, max_attempts: int = 6) -> None:
...
-                # base SSL context for https connection
-                ctx = ssl.create_default_context()
-                ctx.check_hostname = False
-                ctx.verify_mode = ssl.CERT_NONE
-
                 # get the file as a bytes blob
                 request = urllib.request.Request(url)
-                with urllib.request.urlopen(request, context=ctx) as fr:
+                with urllib.request.urlopen(
+                    request, context=proxy_context
```
BobbyRBruce, line 96:
```suggestion
                    request, context=get_proxy_context()
```

src/python/gem5/resources/client_api/jsonclient.py
```diff
@@ -25,6 +25,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import json
+import ssl
```
BobbyRBruce: I wouldn't think this import is needed.

src/python/gem5/resources/client_api/client_wrapper.py
```diff
@@ -25,6 +25,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import itertools
+import ssl
```
BobbyRBruce: I wouldn't think this import is needed.

src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -158,6 +161,7 @@ def get_resources(
         resource_id: Optional[str] = None,
         resource_version: Optional[str] = None,
         gem5_version: Optional[str] = None,
+        proxy_context: Optional[ssl.SSLContext] = None,
```
BobbyRBruce: Remove this. No longer required.

src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -133,7 +136,7 @@ def _atlas_http_json_req(
 
         for attempt in itertools.count(start=1):
             try:
-                response = request.urlopen(req)
+                response = request.urlopen(req, context=proxy_context)
```
BobbyRBruce:
```suggestion
                response = request.urlopen(req, context=get_proxy_context())
```

src/python/gem5/resources/client_api/abstract_client.py
BobbyRBruce, line 1: Nothing in this file should be changed.

src/python/gem5/resources/client.py
BobbyRBruce, line 1: Nothing in this file should be changed.

src/python/SConscript
```diff
@@ -312,6 +314,7 @@ PySource('gem5.utils.multiprocessing',
 PySource('gem5.utils.multiprocessing',
     'gem5/utils/multiprocessing/popen_spawn_gem5.py')
 
+
```
BobbyRBruce: Accidental new line here. Remove.

src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -26,6 +26,8 @@
 
 import itertools
 import json
+import os
+import ssl
```
BobbyRBruce: These imports are no longer needed.

src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -123,8 +126,8 @@ def _atlas_http_json_req(
 
         **Warning**: This function assumes a JSON response.
         """
+        proxy_context = get_proxy_context()
```
BobbyRBruce: Remove this.

src/python/gem5/resources/client_api/atlasclient.py
```diff
@@ -123,8 +126,8 @@ def _atlas_http_json_req(
...
         data = json.dumps(data_json).encode("utf-8")
-
```
BobbyRBruce: Re-add the newline here.

src/python/gem5/resources/downloader.py
```diff
@@ -28,6 +28,7 @@
 import os
 import random
 import shutil
+import ssl
```
BobbyRBruce: I wouldn't think this import is needed.

src/python/gem5/resources/downloader.py
```diff
@@ -61,7 +63,11 @@
 """
 
 
-def _download(url: str, download_to: str, max_attempts: int = 6) -> None:
+def _download(
+    url: str,
+    download_to: str,
+    max_attempts: int = 6,
+) -> None:
```
BobbyRBruce: I don't think this needs reformatted.

src/python/gem5/resources/downloader.py
```diff
@@ -79,38 +85,21 @@ def _download(url: str, download_to: str, max_attempts: int = 6) -> None:
 
     # TODO: This whole setup will only work for single files we can get via
     # wget. We also need to support git clones going forward.
-
+    proxy_context = get_proxy_context()
```
BobbyRBruce: Remove this.

BobbyRBruce (changes requested):

src/python/gem5/resources/client_api/abstract_client.py
BobbyRBruce: @Harshil2107 : Please don't "Resolve conversation" when you didn't resolve the issue. There's still changes in this file.

src/python/gem5/resources/downloader.py
```diff
@@ -79,38 +80,20 @@ def _download(url: str, download_to: str, max_attempts: int = 6) -> None:
 
     # TODO: This whole setup will only work for single files we can get via
     # wget. We also need to support git clones going forward.
-
```
BobbyRBruce: Re-add this line to keep git clean.

src/python/gem5/resources/downloader.py
```diff
@@ -79,38 +80,20 @@ def _download(url: str, download_to: str, max_attempts: int = 6) -> None:
...
-                ctx = ssl.create_default_context()
-                ctx.check_hostname = False
-                ctx.verify_mode = ssl.CERT_NONE
-
                 # get the file as a bytes blob
                 request = urllib.request.Request(url)
-                with urllib.request.urlopen(request, context=ctx) as fr:
+                with urllib.request.urlopen(
+                    request, context=get_proxy_context()
+                ) as fr:
```
BobbyRBruce: You could be a little neater here. Checking the environment variable isn't necessary:

```python
# check to see if user requests a proxy connection.
# `get_proxy_context() returns the proxy context if the user requested use of one, else `None`.
proxy_context = get_proxy_context()
if proxy_context:
    # get the file as a bytes blob
    request = urllib.request.Request(url)
    with urllib.request.urlopen(request, context=proxy_context()) as fr:
    # ...
```

src/python/gem5/resources/downloader.py
```diff
@@ -79,38 +80,20 @@ def _download(url: str, download_to: str, max_attempts: int = 6) -> None:
...
                     with tqdm.wrapattr(
                         open(download_to, "wb"),
                         "write",
```
BobbyRBruce, line 100: I'm curious as to why we've got two different ways to download depending on whether a proxy was specified or not. The only difference appears to be that if proxy is specified `urlopen` is used while if it's not `urlretrieve` is used. If we used the `urlopen` approach for both, would that not work?

**This is not a blocker for this PR. Nor am I seriously asking for you to investigate. This isn't your code. I just think it's an odd design and I don't know the reason**.

BobbyRBruce (changes requested):

src/python/gem5/resources/downloader.py
powerjg: This is my code, IIRC. I don't remember why. Hopefully I wrote something down in the commit message 😆

src/python/gem5/resources/downloader.py
BobbyRBruce: Or, `with urllib.request.urlopen(request, context=proxy_context) as fr:` rather.

BobbyRBruce (approved):

PR #871, opened 2024-02-15
mem-cache: Fix possible crash in base prefetcher

rjc-arch:
When processing memory Packets for prefetch, the `PrefetchInfo` class constructor will attempt to copy the `Packet` data. In cases where the `Packet` under consideration does not contain data, an assertion will be triggered in the Packet's `getConstPtr` method, causing the simulation to crash.

This problem was first exposed by Bug #580 when processing an `UpgradeReq` memory packet.

This patch addresses the problem by suppressing the copying of the `Packet` data during construction of a `PrefetchInfo` object in cases where the `Packet` has no data.

This patch addresses Bug #580 [1], which was exposed by PR #564 [2], subsequently reverted by PR #581 [3]

[1] https://github.com/gem5/gem5/issues/580
[2] https://github.com/gem5/gem5/pull/564
[3] https://github.com/gem5/gem5/pull/581

Change-Id: Ic1e828c0887f4003441b61647440c8e912bf0fbc

### DIFF
--- a/src/mem/cache/prefetch/base.cc
+++ b/src/mem/cache/prefetch/base.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2014, 2023 ARM Limited
+ * Copyright (c) 2013-2014, 2023-2024 Arm Limited
  * All rights reserved.
  *
  * The license below extends only to copyright in the software and shall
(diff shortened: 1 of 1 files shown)


rjc-arch:
Since this bug was first reported the disk image has been upgraded, and the new version does not seem to trigger the bug. I have tested this patch, back-ported to gem5-23.0, against the old v1.0.0 disk image, and it seems to fix the problem.

powerjg (approved):

odanrc (approved):

PR #872, opened 2024-02-15
mem-cache: Prefetchers Improvements

rjc-arch:
This pull request contains a set of small patches which fix some bugs in the gem5 prefetchers, and aligns out-of-the box prefetcher performance more closely with that which a typical user would expect.

The performance patches have been tested with an out-of-the-box (untuned) Stride prefetcher configuration against a set of SPEC 2017 SimPoints, and show a small-to-modest IPC uplift across the about half the benchmarks, with no significant IPC degradation.

The new defaults were identified as part of work on gem5 prefetchers undertaken by Nikolaos Kyparissas while on internship at Arm.

This PR is an updated version of PR #564, which was reverted due to Bug #580. Bug #580 was fixed in PR #871. This PR updates #564 to the latest state of the develop branch, and should be applied after PR #871.

### DIFF
--- a/src/mem/cache/prefetch/base.cc
+++ b/src/mem/cache/prefetch/base.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2014, 2023-2024 Arm Limited
+ * Copyright (c) 2013-2014, 2022-2024 Arm Limited
  * All rights reserved.
  *
  * The license below extends only to copyright in the software and shall
@@ -238,6 +238,7 @@ Base::probeNotify(const CacheAccessProbeArg &acc, bool miss)
     // operations or for writes that we are coaslescing.
     if (pkt->cmd.isSWPrefetch()) return;
     if (pkt->req->isCacheMaintenance()) return;
+    if (pkt->isCleanEviction()) return;
     if (pkt->isWrite() && cache.coalesce()) return;
     if (!pkt->req->hasPaddr()) {
         panic("Request must have a physical address");
--- a/src/mem/cache/prefetch/stride.hh
+++ b/src/mem/cache/prefetch/stride.hh
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2018 Inria
- * Copyright (c) 2012-2013, 2015 ARM Limited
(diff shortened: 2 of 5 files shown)


src/mem/cache/prefetch/base.cc
```diff
@@ -238,6 +238,7 @@ Base::probeNotify(const CacheAccessProbeArg &acc, bool miss)
     // operations or for writes that we are coaslescing.
```
odanrc, line 238: Side comment - can be ignored: I would say that the new line should have been added here, but the problem is that this comment, as it is, does not add any relevant information. Ideally this comment should inform why the prefetcher is not notified in these scenarios. It would be nice if you could add this why info to the new line you are adding in this PR. It would be extra nice if you could improve the whole comment to explain why the prefetcher is not notified in each of these scenarios.

src/mem/cache/prefetch/stride.hh
```diff
@@ -105,6 +105,8 @@ class Stride : public Queued
 
     const int degree;
 
+    const int distance;
```
odanrc, line 114: Nit: Could you document what the distance means, so that it is easier for students to understand this prefetcher? Feel free to copy what you wrote in the python file.

odanrc (approved):

src/mem/cache/prefetch/stride.hh
rjc-arch: Done.

src/mem/cache/prefetch/base.cc
rjc-arch: Hi Daniel. That's a fair comment. It looks like this section has become a collection of edge cases that have caused problems in the past. 
I'm not personally aware of the exact reasons for most of these early returns. Perhaps we could leave improving this documentation for a future commit?

odanrc: Sure, no problem!

odanrc (approved):

powerjg (dismissed):

ivanaamit:
Hi @rjc-arch, for tests to be run on this PR, you should first resolve the conflict. Thank you.

powerjg (approved):

PR #886, opened 2024-02-21
arch-riscv,dev: Update the PLIC implementation

rogerchang23424:
Update the PLIC based on the [riscv-plic-spec](https://github.com/riscv/riscv-plic-spec) in the PR:
- Support customized PLIC hardID and privilege mode configuration
- Backward compatable with the n_contexts parameter, will generate the config like {0,M}, {0,S}, {1,M} ...

Change-Id: Ibff736827edb7c97921e01fa27f503574a27a562

### DIFF
--- a/src/dev/riscv/Plic.py
+++ b/src/dev/riscv/Plic.py
@@ -58,10 +58,8 @@ class PlicBase(BasicPioDevice):
 class Plic(PlicBase):
     """
     This implementation of PLIC is based on
-    the SiFive U54MC datasheet:
-    https://sifive.cdn.prismic.io/sifive/fab000f6-
-    0e07-48d0-9602-e437d5367806_sifive_U54MC_rtl_
-    full_20G1.03.00_manual.pdf
+    the riscv-plic-spec repository:
+    https://github.com/riscv/riscv-plic-spec/releases/tag/1.0.0
     """
 
     type = "Plic"
@@ -69,9 +67,20 @@ class Plic(PlicBase):
     cxx_class = "gem5::Plic"
     pio_size = 0x4000000
     n_src = Param.Int("Number of interrupt sources")
+    # Ref: https://github.com/qemu/qemu/blob/760b4dc/hw/intc/sifive_plic.c#L285
+    hart_config = Param.String(
+        "",
+        "String represent for PLIC hart/pmode config like QEMU plic"
+        "Ex."
+        "'M'              1 hart with M mode"
+        "'MS,MS'          2 harts, 0-1 with M and S mode"
+        "'M,MS,MS,MS,MS'  5 harts, 0 with M mode, 1-5 with M and S mode",
+    )
     n_contexts = Param.Int(
+        0,
+        "Deprecated, use `hart_config` instead. "
         "Number of interrupt contexts. Usually the number "
-        "of threads * 2. One for M mode, one for S mode"
+        "of threads * 2. One for M mode, one for S mode",
     )
 
     def generateDeviceTree(self, state):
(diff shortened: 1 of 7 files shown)


src/dev/riscv/Plic.py
```diff
@@ -58,19 +58,26 @@ class PlicBase(BasicPioDevice):
...
+        "",
+        "String represent for PLIC hart/pmode config like QEMU plic"
+        "Ex."
+        "'M'              1 hart with M mode"
+        "'MS,MS'          2 harts, 0-1 with M and S mode"
+        "'M,MS,MS,MS,MS'  5 harts, 0 with M mode, 1-5 with M and S mode",
+    )
     n_contexts = Param.Int(
-        "Number of interrupt contexts. Usually the number "
+        "Deprecated, use `hard_config` instead. Usually the number "
```
powerjg:
```suggestion
        "Deprecated, use `hart_config` instead. Usually the number "
```

src/dev/riscv/plic.cc
```diff
@@ -421,6 +429,37 @@ Plic::propagateOutput()
...
+          hart_id++;
+          break;
+        case 'M':
+          contextConfigs.emplace_back(hart_id, ExceptionCode::INT_EXT_MACHINE);
+          break;
+        case 'S':
+          contextConfigs.emplace_back(hart_id, ExceptionCode::INT_EXT_SUPER);
+          break;
+        default:
+          panic("hart_config should not contains the value: %c", c);
```
powerjg:
```suggestion
          fatal("hart_config should not contains the value: %c", c);
```

This is a config issue, not a bug in gem5.

src/dev/riscv/Plic.py
```diff
@@ -89,12 +96,26 @@ def generateDeviceTree(self, state):
...
-            int_extended.append(phandle)
-            int_extended.append(0x9)
+        if self.n_contexts != 0:
+            for cpu in cpus:
+                phandle = int_state.phandle(cpu)
+                int_extended.append(phandle)
+                int_extended.append(0xB)
+                int_extended.append(phandle)
+                int_extended.append(0x9)
+        elif self.hart_config != "":
```
powerjg, line 108: This needs to be updated in src/python/gem5/components/boards/riscv_board.py and src/python/gem5/prebuilt/riscvmatched/riscvmatched_board.py as well.

The duplicated code is definitely a problem (apologies for asking you to make the updates). Unfortunately, I don't think that having the code for producing the device tree in the SimObject configuration files can work in general.

src/dev/riscv/plic.cc
```diff
@@ -57,10 +57,18 @@ Plic::Plic(const Params &params) :
     PlicBase(params),
     system(params.system),
     nSrc(params.n_src),
-    nContext(params.n_contexts),
     registers(params.name, pioAddr, this),
     update([this]{updateOutput();}, name() + ".update")
 {
+    panic_if(params.hart_config != "" && params.n_contexts != 0,
```
powerjg:
```suggestion
    fatal_if(params.hart_config != "" && params.n_contexts != 0,
```

This is a config issue, not a bug in gem5.

powerjg (commented):
As an aside, I'm getting more and more concerned about all of the *code* in the SimObject description file. This code should be in the stdlib or in configuration files.

The problem is that when the code is in the SimObject description file, it's *very* difficult for users to know where default values come from.

src/dev/riscv/Plic.py
rogerchang23424: > This needs to be updated in src/python/gem5/components/boards/riscv_board.py and src/python/gem5/prebuilt/riscvmatched/riscvmatched_board.py as well.

Done

> The duplicated code is definitely a problem (apologies for asking you to make the updates). Unfortunately, I don't think that having the code for producing the device tree in the SimObject configuration files can work in general.

Yes, I think it will be use only in the `configs/example/riscv/fs_linux.py`, right?

powerjg: Honestly, I'm not sure how/where this code is used, which is another problem with having code in these files :)

powerjg (approved):
Thanks for the updates!

PR #887, opened 2024-02-21
sim-se: Implement statx system call for Linux x86-64

nmosier:
Implement the statx Linux-specific system call for x86-64. statx is used by LLVM's libc.

Change-Id: Ic000a36a5e5c1399996f520fa357b9354c73c864

### DIFF
--- a/src/sim/syscall_emul.hh
+++ b/src/sim/syscall_emul.hh
@@ -682,14 +682,7 @@ template <typename OS, typename TgtStatPtr, typename HostStatPtr>
 void
 copyOutStatxBuf(TgtStatPtr tgt, HostStatPtr host, bool fakeTTY = false)
 {
-#if !defined(__linux__)
-    static_assert(false, "statx is a Linux-specific system call.");
-#endif
-
-    const auto bo = [] (auto& value) {
-        constexpr ByteOrder bo = OS::byteOrder;
-        return value = htog(value, bo);
-    };
+    constexpr ByteOrder bo = OS::byteOrder;
 
     if (fakeTTY) {
         tgt->stx_dev_major = MAJOR(0xA);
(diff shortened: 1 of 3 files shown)


powerjg (commented):
looks OK to me. I didn't look closely, but if no one else has the time, I can approve.

src/sim/syscall_emul.hh
```diff
@@ -677,6 +678,68 @@ copyOutStatfsBuf(TgtStatPtr tgt, HostStatPtr host)
 #endif
 }
 
+template <typename OS, typename TgtStatPtr, typename HostStatPtr>
+void
+copyOutStatxBuf(TgtStatPtr tgt, HostStatPtr host, bool fakeTTY = false)
+{
+#if !defined(__linux__)
+    static_assert(false, "statx is a Linux-specific system call.");
+#endif
```
giactra: Can I ask you why you are checking if we are in a linux host?

This is checked in other stat* implementations (like ststfs) when the *host* syscall is used to mimic the guest (linux) syscall. In this case it doesn't seem like the backend of statxFunc is using statx so I don't see the reason why we should check this, especially since the static_assert will fail at compile time making it impossible to build on non-Linux hosts (like macOS)

src/sim/syscall_emul.hh
```diff
@@ -677,6 +678,68 @@ copyOutStatfsBuf(TgtStatPtr tgt, HostStatPtr host)
...
+    tgt->stx_mode = host->stx_mode;
+    if (fakeTTY) {
+      // Claim to be character device.
+      tgt->stx_mode &= ~S_IFMT;
+      tgt->stx_mode |= S_IFCHR;
+    }
+    bo(tgt->stx_mode);
+    bo(tgt->stx_nlink = host->stx_nlink);
+    bo(tgt->stx_uid = host->stx_uid);
+    bo(tgt->stx_gid = host->stx_gid);
```
giactra: I honestly doubt

    bo(tgt->stx_nlink = host->stx_nlink)

is more readable and intuitive than

    tgt->stx_nlink = htog(host->stx_nlink, bo);

Especially as it relies on some obscure C assignment rule on what an assignment expression evaluates to. I'd vote to keep this simple

src/sim/syscall_emul.hh
```diff
@@ -1456,6 +1519,47 @@ stat64Func(SyscallDesc *desc, ThreadContext *tc,
...
+    std::string path;
+
+    if (!SETranslatingPortProxy(tc).tryReadString(path, pathname))
+        return -EFAULT;
+
+    if (path.empty() && !(flags & OS::TGT_AT_EMPTY_PATH))
+        return -ENOENT;
+    flags = flags & ~OS::TGT_AT_EMPTY_PATH;
+
+    warn_if(flags != 0, "newfstatat: Flag bits %#x not supported.", flags);
```
giactra: statx instead of newfstatat?

giactra (changes requested):

src/sim/syscall_emul.hh
nmosier: I understand, thanks for explaining. I removed this check in the latest commit.

src/sim/syscall_emul.hh
nmosier: I changed it in the latest commit to use the method used in other stat* functions.

src/sim/syscall_emul.hh
nmosier: Fixed in the latest commit

src/sim/syscall_emul.hh
```diff
@@ -677,6 +678,73 @@ copyOutStatfsBuf(TgtStatPtr tgt, HostStatPtr host)
...
+        tgt->stx_dev_major = MAJOR(0xA);
+        tgt->stx_dev_minor = MINOR(0xA);
+    } else {
+        tgt->stx_dev_major = host->stx_dev_major;
+        tgt->stx_dev_minor = host->stx_dev_minor;
+    }
+    tgt->stx_dev_major = htog(tgt->stx_dev_major, bo);
+    tgt->stx_dev_minor = htog(tgt->stx_dev_minor, bo);
+    tgt->stx_ino = host->stx_ino;
+    tgt->stx_ino = htog(tgt->stx_ino, bo);
```
giactra, line 697: I checked other copyOut helpers and I see why you are doing this in two assignments (for consistency).
I honestly believe we should break our past habit and write this in one :) :

    tgt->stx_ino = htog(host->stx_ino, bo);


Same applies to the rest of the file

giactra (changes requested):
Thanks for your changes, there is one issue I highlighted

ivanaamit:
@nmosier, could you please address Giacomo's comment? Thank you.

src/sim/syscall_emul.hh
nmosier: I think writing it as one statement in that way could lead to buggy results if the host and target have fields of different sizes. E.g., say the host has a 64-bit stx_ino field but the target has a 32-bit stx_ino field. When writing it as two statements, it (1) narrows the integer to 32 bits, then (2) possibly reverses the byte order.
But if writing it as one statement as above, those operations happen in the reverse order: (1) reverse the byte order, then (2) narrow integer to 32 bits.

I think this is a pathological case, since statx is only implemented on Linux currently I believe, so we'd expect the statx member sizes to always match from host to target. Let me know if the above isn't a concern and I can rewrite all 2-statement assignments into 1-statement assignments.

giactra: Yeah you are absolutely right, I missed this case

giactra (approved):

PR #892, opened 2024-02-24
sim-se, arch-x86: initialize max stack size from parameter

nmosier:
Initialize x86 process' max stack size to the value given in the process params, rather than hard-coding it to 8 MB, which made it impossible to run x86 programs requiring more than 8 MB of stack.

Change-Id: I0b17fe60b016b1e4a82d704ef7ad367974ea6a08


powerjg (approved):
lgtm. I re-ran the tests as I don't think that it was anything you did.

PR #894, opened 2024-02-26
mem-ruby: Add missing transition for SLC writes to VIPER TCC

DanKouch:
Bypassed write though requests on invalid lines in the TCC should be written though to the directory. This transition was previously missing.

Change-Id: I16b117c4e085ce6be0ed5297aa0129d52cd35a51

### DIFF
--- a/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
@@ -1106,7 +1106,7 @@ machine(MachineType:TCC, "TCC Cache")
     st_stallAndWaitRequest;
   }
 
-  transition(I, WrVicBlk) {TagArrayRead} {
+  transition(I, {WrVicBlk, WrVicBlkEvict}) {TagArrayRead} {
     p_profileMiss;
     wt_writeThrough;
     p_popRequestQueue;


mattsinc (approved):

abmerop:
Is this a race condition?  Otherwise I don't follow how a WrVicBlkEvict can be triggered while in I state as it checks `if (presentOrAvail(...))` to trigger it

DanKouch:
> Is this a race condition? Otherwise I don't follow how a WrVicBlkEvict can be triggered while in I state as it checks `if (presentOrAvail(...))` to trigger it

From my understanding presentOrAvail(...) will return true for invalid lines since the line is available for use, so this should happen any time an SLC write goes through the TCC when the associated cache line is unused.

abmerop:
> > Is this a race condition? Otherwise I don't follow how a WrVicBlkEvict can be triggered while in I state as it checks `if (presentOrAvail(...))` to trigger it
> 
> From my understanding presentOrAvail(...) will return true for invalid lines since the line is available for use, so this should happen any time an SLC write goes through the TCC when the associated cache line is unused.

Ah yes, that make sense, thanks

abmerop (approved):

mattsinc (approved):

PR #895, opened 2024-02-26
arch-vega: Implement accumulation offset

abmerop:
This PR implements a few changes related to the accumulation offset which is new in MI200. Previously MI100 contained two vector register files: the architectural and accumulation register files. These have now been unified and the architectural register file is twice the size. As a result of this the dispatch packet set an offset into the unified vector register file for where the former accumulation registers would go. The changes are:

- Calculate the accumulation offset from dispatch packet and store in HSA task.
- Update the accumulation move instructions (v_accvgpr_read/write) to use it.
- Update the current MFMA instructions to use it.
- Make the MFMA examples more clean.

### DIFF
--- a/src/arch/amdgpu/vega/insts/vop3p.cc
+++ b/src/arch/amdgpu/vega/insts/vop3p.cc
@@ -596,10 +596,10 @@ void Inst_VOP3P__V_DOT8_U32_U4::execute(GPUDynInstPtr gpuDynInst)
 
 void Inst_VOP3P__V_ACCVGPR_READ::execute(GPUDynInstPtr gpuDynInst)
 {
-    // The Acc register file is not supported in gem5 and has been removed
-    // in MI200. Therefore this instruction becomes a mov.
     Wavefront *wf = gpuDynInst->wavefront();
-    ConstVecOperandU32 src(gpuDynInst, extData.SRC0);
+    unsigned accum_offset = wf->accumOffset;
+
+    ConstVecOperandU32 src(gpuDynInst, extData.SRC0+accum_offset);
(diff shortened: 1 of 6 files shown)


src/arch/amdgpu/vega/gpu_registers.hh
```diff
@@ -129,7 +129,7 @@ namespace VegaISA
         REG_LDS_DIRECT = 254,
         REG_SRC_LITERAL = 255,
         REG_VGPR_MIN = 256,
-        REG_VGPR_MAX = 511
+        REG_VGPR_MAX = 767
```
mattsinc: Ideally it would be good to know where these magic numbers come from

mattsinc (dismissed):

mattsinc:
@mjkpolo just making sure you see this

src/arch/amdgpu/vega/gpu_registers.hh
mjkpolo: I'm wondering about this too since Table 22 in the mi300 spec (and Table 20 in the mi200 spec) says the VGPR value is from 256-511, and `REG_VGPR_MIN` is 256 but `REG_VGPR_MAX` is not 767 instead of 511

abmerop: Dropped this commit.  It seems there isn't an index for AV0-AV255 so this change might not be correct.

mattsinc: ok, I will approve again then.  Thanks @mjkpolo for spotting this!

mattsinc (approved):

PR #899, opened 2024-02-28
mem-ruby: Add categorization of bypassed atomics in TCC

DanKouch:
Adds categorization of bypassed atomics in TCC to the TBE as either return or no-return, which gets consumed in pa_performAtomic to determine if atomic logs should be stored.

Reestablishes TCC bypassed atomics after #546.

Change-Id: Ibc1fa2b795ef1c47c3893a0b1911fa7993522d38


DanKouch:
@mattsinc Atomics bypassed from the TCC are broken without this as of #546. Previously, when pa_performAtomic was called on the return path from the directory on bypassed atomics, the in_msg.Type was neither AtomicNoReturn nor AtomicReturn. This change stores whether the atomic returns or does not return in the TBE, which is consumed in pa_performAtomic for bypassed atomics.

mattsinc (approved):
I was not aware it had broken after that patch (or I had forgotten), but this patch seems fine.  We are adding a lot of stuff to the TBEs lately, but for now I don't see a better way.

abmerop (approved):

Issue #881, opened 2024-02-20
arch-riscv: return values of m5 pseudo instructions in 32 bit systems


robhau:
**Return value of m5 instructions**

M5 pseudo instructions return a `64 bit` value.

Example:

https://github.com/gem5/gem5/blob/8759131df3e4af1080fb96c67d8708869c9d74c0/src/sim/pseudo_inst.cc#L153-L158

**Current State** 

At the moment, for `32 bit` systems, the result of a m5 pseudo instruction is shrunk down to `32 bit` and stored only in `a0`

https://github.com/gem5/gem5/blob/8759131df3e4af1080fb96c67d8708869c9d74c0/src/arch/riscv/isa/formats/m5ops.isa#L38-L47

**Problem**

In `32 bit` systems, `64 bit` return values are split into two `32 bit` values and stored in `a0` and `a1`. Software still asumes that `a1` is part of the return value. This leads to wrong values (`a1` can hold old values before m5 function call).
State which version of gem5 this bug was found in. If on the develop branch state the Commit revision ID you are working.

**Gem5 Modification**

In `src/arch/isa/formats/m5ops.isa`, the upper 32 bit of the result have to be stored in `a1` for `RV32`. That could be look like

```C++
def format M5Op() {{
    iop = InstObjParams(name, Name, 'PseudoOp', '''
            uint64_t result;
            if (machInst.rv_type == RV32) {
                pseudo_inst::pseudoInst<RegABI32>(xc->tcBase(), M5FUNC, result);
                a0 = mbits(result, 31, 0);
                a1 = mbits(result, 63, 32);
            } else {
                pseudo_inst::pseudoInst<RegABI64>(xc->tcBase(), M5FUNC, result);
                a0 = rvSext(result);
            }''',
            ['IsNonSpeculative', 'IsSerializeAfter'])
    header_output = BasicDeclare.subst(iop)
    decoder_output = BasicConstructor.subst(iop)
    decode_block = BasicDecode.subst(iop)
    exec_output = BasicExecute.subst(iop)
}};
```

However, when I try to compile it, I receive that a1 is not declared. 

I added `a1` in `src/arch/riscv/isa/operands.isa`, but when i look in the execution trace, the value is not written to `a1`.

rogerchang23424:
Hi @robhau,

We may need to update the implementation in the part to fix the issue

[...]

PR #900, opened 2024-02-28
arch-riscv,sim: m5ops argument / return fix for 32 bit RISC-V

robhau:
M5Ops C / C++ functions partially use 64 bit arguments and return value. In general, 64 bit arguments and return values are possible for 32 bit RISC-V systems as well, since the arguments and the return value is split into two registers. However, at the moment, this does not work for 32 bit RISC-V systems on the simulator side, since there is a one to one mapping between argument registers and m5op function parameters.

To solve this problem, the get() function of the RISC-V reg_abi is updated. It now will merge two registers if there is a 64 bit argument. For this, the function code has to be passed to the get() function. The default value of this function code is set to 0xF00, since 0x00 is already used for M5_ARM. The parameter list of other get() functions for argument return is also extended by this function code parameter with the keyword [[maybe_unused]].

To enable a return value of size 64 bit, a0 is assigned with the lower 32 bit and a1 with the higher 32 bit.

Related Issue: https://github.com/gem5/gem5/issues/881

Change-Id: Id7162cd569879e708fa5e73a2741a2a60ad23b71


giactra:
Hi @robhau, I don't think this is the right way to handle it. Unless there's something I am missing, there is a way to automatically merge two 32bit regs to provide a single function call argument. You just need to specialize the Argument accordingly. See

https://github.com/gem5/gem5/blob/stable/src/arch/arm/reg_abi.hh#L62

In other words there is no need to change the signature, you just need the template instantiation logic to provide a specialization if IsWideV<Args> is encountered

robhau:
Hi @giactra,
thank you for your answer😄 . I do not have much time to look in detail today, but here are some of my first thoughts:

Just `IsWideV` would only signal that two registers have to be merged. But with only knowing the argument index `state`, it is not possible to determine the merged arguments.

Example:
Assume, that `IsWideV` is passed an `state` equals zero.

If the actual function is `m5_exit(uint64_t ns)`, we have to merge `a0` and `a1`.

But if it is `m5_write_file(void *buffer, uint64_t len, uint64_t offset, const char *filename)`, we have to return that value of `a0` only, because in 32 bit RISC-V systems, the pointer size is 32 bit ([RISC-V Calling Conventions](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#cc-type-details))

giactra:
> Hi @giactra, thank you for your answer😄 . I do not have much time to look in detail today, but here are some of my first thoughts:
> 
> Just `IsWideV` would only signal that two registers have to be merged. But with only knowing the argument index `state`, it is not possible to determine the merged arguments.

I am not sure I understand this point

> 
> Example: Assume, that `IsWideV` is passed an `state` equals zero.
> 
> If the actual function is `m5_exit(uint64_t ns)`, we have to merge `a0` and `a1`.
> 
> But if it is `m5_write_file(void *buffer, uint64_t len, uint64_t offset, const char *filename)`, we have to return that value of `a0` only, because in 32 bit RISC-V systems, the pointer size is 32 bit ([RISC-V Calling Conventions](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#cc-type-details))

Regarding your example, it should return a0 only, as if you check https://github.com/gem5/gem5/blob/stable/src/arch/arm/reg_abi.hh#L61 `std::is_integral_v<void*>` will evaluate to false and non merging arg retrieval should be used...

robhau:
Thank you for your answer ☺️ .


> > Hi @giactra, thank you for your answer😄 . I do not have much time to look in detail today, but here are some of my first thoughts:
> > 
> > Just IsWideV would only signal that two registers have to be merged. But with only knowing the argument index state, it is not possible to determine the merged arguments.
> 
> I am not sure I understand this point
> 

`IsWideV` should be true if the argument wide is larger than register wide.

I think that `IsWideV` is always true because it takes the arguments from `src/sim/pseudo_inst.hh` and these arguments are all 64 bit wide. (Please correct me if I am wrong. I am not experienced with C++ templates)

> > Example: Assume, that IsWideV is passed an state equals zero.
> > 
> > If the actual function is m5_exit(uint64_t ns), we have to merge a0 and a1.
> > 
> > But if it is m5_write_file(void *buffer, uint64_t len, uint64_t offset, const char *filename), we have to return that value of a0 only, because in 32 bit RISC-V systems, the pointer size is 32 bit ([RISC-V Calling Conventions](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#cc-type-details))
>
> Regarding your example, it should return a0 only, as if you check https://github.com/gem5/gem5/blob/stable/src/arch/arm/reg_abi.hh#L61 std::is_integral_v<void*> will evaluate to false and non merging arg retrieval should be used...


I am not sure if this will work.

Another example:
functions:
1. `m5_write_file(void *buffer, uint64_t len, uint64_t offset, const char *filename)`
2. `void m5_reset_stats(uint64_t ns_delay, uint64_t ns_period)`

`state` equals 1, i.e., we have to return the second argument.

For function 1, we have to merge `a1` and `a2`.
For function 2, we have to merge `a2` and `a3`.

How can we decide which argument register have to be merged without knowing the actual function?

giactra:
> Thank you for your answer ☺️ .
> 

Thanks to you for trying to fix this!

> > > Hi @giactra, thank you for your answer😄 . I do not have much time to look in detail today, but here are some of my first thoughts:
> > > Just IsWideV would only signal that two registers have to be merged. But with only knowing the argument index state, it is not possible to determine the merged arguments.
> > 
> > 
> > I am not sure I understand this point
> 
> `IsWideV` should be true if the argument wide is larger than register wide.
> 
> I think that `IsWideV` is always true because it takes the arguments from `src/sim/pseudo_inst.hh` and these arguments are all 64 bit wide. (Please correct me if I am wrong. I am not experienced with C++ templates)
> 


Well not always (we have pointer parameters as well), but most of the times yes for a 32-bit ABI


> > > Example: Assume, that IsWideV is passed an state equals zero.
> > > If the actual function is m5_exit(uint64_t ns), we have to merge a0 and a1.
> > > But if it is m5_write_file(void *buffer, uint64_t len, uint64_t offset, const char *filename), we have to return that value of a0 only, because in 32 bit RISC-V systems, the pointer size is 32 bit ([RISC-V Calling Conventions](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#cc-type-details))
> > 
> > 
> > Regarding your example, it should return a0 only, as if you check https://github.com/gem5/gem5/blob/stable/src/arch/arm/reg_abi.hh#L61 std::is_integral_v<void*> will evaluate to false and non merging arg retrieval should be used...
> 
> I am not sure if this will work.
> 
> Another example: functions:
> 
>     1. `m5_write_file(void *buffer, uint64_t len, uint64_t offset, const char *filename)`
> 
>     2. `void m5_reset_stats(uint64_t ns_delay, uint64_t ns_period)`
> 
> 
> `state` equals 1, i.e., we have to return the second argument.
> 
> For function 1, we have to merge `a1` and `a2`. For function 2, we have to merge `a2` and `a3`.
> 
> How can we decide which argument register have to be merged without knowing the actual function?

I see what you mean now. In principle this should be automatically handled when the variadic templates are recursively instantiated. The m5_op signature is scanned argument by argument and the value is retrieved by calling a sequence of Argument<>::get() until the argument tuple is collected.

So example 2:

>     2. `void m5_reset_stats(uint64_t ns_delay, uint64_t ns_period)`

you would call:

```
    static Arg
    get(ThreadContext *tc, typename ABI::State &state)
    {
        panic_if(state >= ABI::ArgumentRegs.size(),
                "Ran out of syscall argument registers.");
        return bits(tc->getReg(ABI::ArgumentRegs[state++]), 31, 0);
    }
```

(from https://github.com/giactra/gem5/blob/develop/src/arch/riscv/reg_abi.hh#L67)

On uint64_t ns_delay:  State will be updated from 0 to 1 (see state++) -> using a0
On uint64_t ns_period. State will be updated from 1 to 2 (see state++) -> using a1

You can see what is the issue here: you are using a0 and a1 rather than a0..a3. This just means you need to fix the RISCV implementation accordingly in the IsWideV = true case :)

Hope this helps

robhau:
Thank you for your answer ☺️ 

> I see what you mean now. In principle this should be automatically handled when the variadic templates are recursively instantiated. The m5_op signature is scanned argument by argument and the value is retrieved by calling a sequence of Argument<>::get() until the argument tuple is collected.

I do not think that the m5_op signature is scanned that is used inside workload code, but the function signature from the functions in `pseudo_inst.hh`.

These are the functions that are defined inside the gem 5 simulator and executed to generate the specific results.
https://github.com/gem5/gem5/blob/c890e6b113a59caeec3eea9bafea4a03a471f89d/src/sim/pseudo_inst.hh#L67-L82

These function are given as argument for the `invokeSimcall` function call etc. pp.
https://github.com/gem5/gem5/blob/c890e6b113a59caeec3eea9bafea4a03a471f89d/src/sim/pseudo_inst.hh#L111-L131

Function signature of `invokeSimCall`

https://github.com/gem5/gem5/blob/c890e6b113a59caeec3eea9bafea4a03a471f89d/src/sim/guest_abi.hh#L48-L59

The parameter types of the internal m5 functions are all 64 bit wide (`uint64_t`).

> On uint64_t ns_delay: State will be updated from 0 to 1 (see state++) -> using a0
> On uint64_t ns_period. State will be updated from 1 to 2 (see state++) -> using a1
>
> You can see what is the issue here: you are using a0 and a1 rather than a0..a3. This just means you need to fix the RISCV > implementation accordingly in the IsWideV = true case :)

IsWideV = true is the case for both `m5_write_file` and `m5_reset_stats` (second argument), but for the first we would merge `a1` with `a2`, for the latter `a2` and `a3`.

giactra:
> Thank you for your answer ☺️
> 
> > I see what you mean now. In principle this should be automatically handled when the variadic templates are recursively instantiated. The m5_op signature is scanned argument by argument and the value is retrieved by calling a sequence of Argument<>::get() until the argument tuple is collected.
> 
> I do not think that the m5_op signature is scanned that is used inside workload code, but the function signature from the functions in `pseudo_inst.hh`.

You are actually right, but the trick is that the two signatures should match. So scanning the (e.g) pseudo_inst::writefile and the m5_write_file should be the same. If the two signatures were different, all this guest ABI behaviour wouldn't work.

> IsWideV = true is the case for both `m5_write_file` and `m5_reset_stats` (second argument), but for the first we would merge `a1` with `a2`, for the latter `a2` and `a3`.

Again, I really feel the issue is that RISCV is not doing any merging, unless I completely misunderstood the behaviour of the GuestABI

Please copy the merge behaviour of AArch32 (it's 4 lines of code) and let me know if it works.

We shouldn't add any unnecessary argument to differentiate between pseudo-ops. I believe it defeats the purpose of using GuestABI in the first place.

robhau:
> > Thank you for your answer ☺️
> > 
> > > I see what you mean now. In principle this should be automatically handled when the variadic templates are recursively instantiated. The m5_op signature is scanned argument by argument and the value is retrieved by calling a sequence of Argument<>::get() until the argument tuple is collected.
> > 
> > I do not think that the m5_op signature is scanned that is used inside workload code, but the function signature from the functions in pseudo_inst.hh.
> 
> You are actually right, but the trick is that the two signatures should match. So scanning the (e.g) pseudo_inst::writefile and the m5_write_file should be the same. If the two signatures were different, all this guest ABI behaviour wouldn't work.

The function signature are not the same.

`gem5::pseudo_inst::writefile`:
https://github.com/gem5/gem5/blob/c890e6b113a59caeec3eea9bafea4a03a471f89d/src/sim/pseudo_inst.hh#L75-L76

`Addr` and `Tick` are `uint64_t` (8 byte large).

`m5_write_file`:

https://github.com/gem5/gem5/blob/c890e6b113a59caeec3eea9bafea4a03a471f89d/include/gem5/m5ops.h#L59-L60

Here, `void*` and `const char*` should be only 4 byte large for 32 bit RISC-V systems.

> Please copy the merge behaviour of AArch32 (it's 4 lines of code) and let me know if it works.

I copied the content of `get()` from `src/arch/arm/reg_abi.hh`.

M5ops were included in my RISC-V code as inline assembly.
```C
uint64_t __attribute__((naked)) m5_write_file(void *buffer, uint64_t len, uint64_t offset,
                       const char *filename)
{
    asm (".insn	4, 0x9e00007b");
    asm ("ret");
}
```
I printed out `state` on every function call of `get()`.

simout:
```sh
415950000: system.cpu: T0 : 0x12c @main+74    : addi a5, zero, 384         : IntAlu :  D=0x0000000000000180
416000000: system.cpu: T0 : 0x130 @main+78    : c_li a3, 0                 : IntAlu :  D=0x0000000000000000
416000000: system.cpu: T0 : 0x132 @main+80    : c_li a4, 0                 : IntAlu :  D=0x0000000000000000
416050000: system.cpu: T0 : 0x134 @main+82    : c_li a1, 16                : IntAlu :  D=0x0000000000000010
416050000: system.cpu: T0 : 0x136 @main+84    : c_li a2, 0                 : IntAlu :  D=0x0000000000000000
416100000: system.cpu: T0 : 0x138 @main+86    : c_addi4spn a0, sp, 16      : IntAlu :  D=0x000000000005ffe0
416100000: system.cpu: T0 : 0x13a @main+88    : c_jal -296                 : IntAlu :  D=0x000000000000013c
417250000: global: pseudo_inst::pseudoInst(79)
state: 0
state: 2
state: 4
state: 6
src/arch/riscv/reg_abi.hh:76: panic: panic condition state + 1 >= ABI::ArgumentRegs.size() occurred: Ran out of syscall argument registers.
```
As you can see, register merging is already done for the first argument, even if it is not appropriate.

> We shouldn't add any unnecessary argument to differentiate between pseudo-ops. I believe it defeats the purpose of using GuestABI in the first place.

I am not 100 % happy with my solution too 😅 . But at the moment, the m5 opcode has to be respected somehow.

giactra:
Gotcha, that’s the second part of the issue.
void* IS equivalent to Addr. Same applies to any pointer like char*. In the sense that Addr is the gem5 view of a guest address (pointer). Only that in gem5 we always represent Addr as uint64. On top of the merging behaviour for IsWideV, you need to specialize the template for Addr when in 32-bit ABI to update the state once and read one register argument only. (Treating it like a 32bit variable even if it is not)

robhau:
> Gotcha, that’s the second part of the issue.
void* IS equivalent to Addr. Same applies to any pointer like char*. In the sense that Addr is the gem5 view of a guest address (pointer). Only that in gem5 we always represent Addr as uint64. On top of the merging behaviour for IsWideV, you need to specialize the template for Addr when in 32-bit ABI to update the state once and read one register argument only. (Treating it like a 32bit variable even if it is not)

I came to the same conclusion 😄. However, I am not sure how, for example, `gem5::Addr` can be distinguished from `uint64_t`, since `Addr` is `uint64_t`.

```C++
typedef uint64_t Addr;
```

giactra:
> > Gotcha, that’s the second part of the issue.
> > void* IS equivalent to Addr. Same applies to any pointer like char*. In the sense that Addr is the gem5 view of a guest address (pointer). Only that in gem5 we always represent Addr as uint64. On top of the merging behaviour for IsWideV, you need to specialize the template for Addr when in 32-bit ABI to update the state once and read one register argument only. (Treating it like a 32bit variable even if it is not)
> 
> I came to the same conclusion 😄. However, I am not sure how, for example, `gem5::Addr` can be distinguished from `uint64_t`, since `Addr` is `uint64_t`.
> 
> ```c++
> typedef uint64_t Addr;
> ```

Oh right! That is unfortunate. 

I believe the solutions are the following

1) Redefine Addr as

`struct Addr { uint64_t value }; `

Even if we overload some operators to make it look like a primitive type (as it is now), this is something that requires some testing in terms of performance analysis; the Addr is one of the most important data types in gem5.

2) Retrofit VPtr (used in syscall emulation) for the pseudo inst ABI. The problem with this approach is that the Vptr is assuming a SE proxy while we should be able to use this in both SE and FS.

3) Add a temporary

`struct GuestAddr { Addr value };`

To be used by pseudo instructions only. Then with GuestAddr we would be able to specialize the ABI accordingly.

IMHO the cleanest solution would be number two, but I presume this will require some work. To provide a quick but not so dirty fix we can rely on number three for the moment.. Please let me know your thoughts

robhau:
Hi,
thank you again for your help! 😄 

> I believe the solutions are the following
> 
> 1. Redefine Addr as
> 
> struct Addr { uint64_t value }; 
> 
> Even if we overload some operators to make it look like a primitive type (as it is now), this is something that requires some testing in terms of performance analysis; the Addr is one of the most important data types in gem5.

My colleague also suggested me to wrap `Addr` inside a struct (I only asked how to distinguish between original type and typedef), but I do not like this solution since `Addr` appears almost everywhere.




> 2. Retrofit VPtr (used in syscall emulation) for the pseudo inst ABI. The problem with this approach is that the Vptr is assuming a SE proxy while we should be able to use this in both SE and FS.
> 
> 3. Add a temporary
> 
> `struct GuestAddr { Addr value };`
> 
> To be used by pseudo instructions only. Then with GuestAddr we would be able to specialize the ABI accordingly.
> 
> IMHO the cleanest solution would be number two, but I presume this will require some work. To provide a quick but not so dirty fix we can rely on number three for the moment.. Please let me know your thoughts

I also prefer solution 3 at the moment and would implement this now.

robhau:
At the moment, the 32 bit RISC-V asm tests are failing. @rogerchang23424 mentioned the need for a change in https://github.com/gem5/gem5/issues/881.

rogerchang23424:
Thanks for @robhau fixing, it looks like that the [macros/mt](https://github.com/gem5/gem5-resources/tree/stable/src/asmtest/isa/macros/mt) need to be updated

mmapFunc has the type `VPtr<>`
https://github.com/gem5/gem5/blob/c57a6b0d595cbeccd468e356e1e1d8a388c71c33/src/sim/syscall_emul.hh#L1852-L1867

src/sim/pseudo_inst.cc
```diff
@@ -368,10 +368,10 @@ m5checkpoint(ThreadContext *tc, Tick delay, Tick period)
 }
 
 uint64_t
-readfile(ThreadContext *tc, Addr vaddr, uint64_t len, uint64_t offset)
+readfile(ThreadContext *tc, GuestAddr vaddr, uint64_t len, uint64_t offset)
```
rogerchang23424, line 371: I feel like the we can simple change `Addr` to `VPtr<>`

```suggestion
readfile(ThreadContext *tc, VPtr<> vaddr, uint64_t len, uint64_t offset)
```

Details of `VPtr<>`:

The `VPtr<>` is same as `ProxyPtr<void, SETranslatingPortProxy>`
https://github.com/gem5/gem5/blob/f35815cd48a0716452a88594563c2c5131d69424/src/sim/proxy_ptr.hh#L400

`ProxyPtr<T, Proxy>` declaration
https://github.com/gem5/gem5/blob/f35815cd48a0716452a88594563c2c5131d69424/src/sim/proxy_ptr.hh#L237-L327

get arguments fro `ProxyPtr`, use `ABI::UIntPtr` type. Usually the type of `UIntPtr` is `uint32_t` if the syscall ABI is 32bits, otherwise the size of `UIntPtr` is `uint64_t`

https://github.com/gem5/gem5/blob/f35815cd48a0716452a88594563c2c5131d69424/src/sim/proxy_ptr.hh#L363-L372

Feel free to leave the message if any concern

rogerchang23424 (changes requested):

src/sim/pseudo_inst.cc
giactra: Hi @rogerchang23424, using VPtr would be the best solution, but as I mentioned in one of my earlier comments it assumes a SETranslatingPortProxy which is SE specific and makes it not usable in FS mode. And we want to be able to call m5ops in FS mode

rogerchang23424: @giactra Sorry for missing the comments above. Currently we can't pass the github CI because of the SegmentFault of the binary `rv32uamt-ps-amoxxx_w` in SE mode. I need to find the way to solve it.

Thanks!

robhau: Hi @rogerchang23424,
I will take a look on it tomorrow.

rogerchang23424:
> Thanks for @robhau fixing, it looks like that the [macros/mt](https://github.com/gem5/gem5-resources/tree/stable/src/asmtest/isa/macros/mt) need to be updated
> 
> mmapFunc has the type `VPtr<>`
> 
> https://github.com/gem5/gem5/blob/c57a6b0d595cbeccd468e356e1e1d8a388c71c33/src/sim/syscall_emul.hh#L1852-L1867

The reason caused the asm test failed because the sizeof `size_t` and `off_t` are 64 bits.

There are two solution to pass the tests

1. Change the sizeof `size_t` and `off_t` to 32bits
2. Update the system call of [macros/mt](https://github.com/gem5/gem5-resources/tree/stable/src/asmtest/isa/macros/mt)

@robhau, which would you prefer?

Update: the type of `time_t` and `off_t` are changed to 64bits [link](https://patchwork.ozlabs.org/project/glibc/patch/4e95f95966d8d7c6a8339160dc62d81c1f6a1bfb.1578824547.git.alistair.francis@wdc.com/). The syscall of [macros/mt](https://github.com/gem5/gem5-resources/tree/stable/src/asmtest/isa/macros/mt) need to be updated

robhau:
> > Thanks for @robhau fixing, it looks like that the [macros/mt](https://github.com/gem5/gem5-resources/tree/stable/src/asmtest/isa/macros/mt) need to be updated
> > mmapFunc has the type `VPtr<>`
> > https://github.com/gem5/gem5/blob/c57a6b0d595cbeccd468e356e1e1d8a388c71c33/src/sim/syscall_emul.hh#L1852-L1867
> 
> The reason caused the asm test failed because the sizeof `size_t` and `off_t` are 64 bits.
> 
> There are two solution to pass the tests
> 
>     1. Change the sizeof `size_t` and `off_t` to 32bits
> 
>     2. Update the system call of [macros/mt](https://github.com/gem5/gem5-resources/tree/stable/src/asmtest/isa/macros/mt)
> 
> 
> @robhau, which would you prefer?

Hi,
if I understand correct, each RISC-V register corresponds to a syscall argument, right?

Wouldn't it be possible to adapt the SPARC ISA, i.e., define a `get()` function for `RiscvISA::SEWorkload::SyscallABI32`?

In `src/arch/riscv/se_workload.hh`:
```C++
namespace gem5
{
namespace RiscvISA
{
...
class SEWorkload : public gem5::SEWorkload
{
...
    struct SyscallABI32 : public RegABI32
    {
        using UintPtr = uint32_t;
    };
};

} // namespace RiscvISA

namespace guest_abi
{
template <typename ABI, typename Arg>
struct Argument<ABI, Arg,
    typename std::enable_if_t<std::is_base_of_v<RiscvISA::SEWorkload::SyscallABI32, ABI>>>
{

    static Arg
    get(ThreadContext *tc, typename ABI::State &state)
    {
        return bits(tc->getReg(ABI::ArgumentRegs[state++]), 31, 0);
    }
};
...
} // namespace guest_abi

} // namespace guest_abi

```

Then, we would not have to change types or the assembler code.

rogerchang23424:
It makes sense to me. If the type is `uint64_t` or `int64_t`, are you tent to only get the lower bits of register?

robhau:
I thought that there were only 32 bit arguments in RV32. I checked a few syscalls in [mt](https://github.com/gem5/gem5-resources/blob/stable/src/asmtest/isa/macros/mt/test_macros_mt_ecall.h), and every argument register corresponds to one function argument.

1. example
https://github.com/gem5/gem5/blob/bbde68c08c3e1b3e0ffb7f884f8488efb9d301d1/src/sim/syscall_emul.hh#L1602-L1606

```C++
// clone a new thread
  li      a0, CLONE_FLAGS
  li      s2, MEM_SIZE
  add     a1, t1, s2        // pointer to the high address of the new stack
  mv      a2, sp            // ptid
  mv      a3, t2            // pointer to the low address of the new TLS,
                            // assuming TLS grows upward
  mv      a4, sp            // ctid
  li      a7, SYSCALL_CLONE // clone syscall number
```

`a0` --> `RegVal flags`
`a1` --> `RegVal newStack`
`a2` --> `VPtr<> ptidPtr`
`a4` --> `VPtr<> ctidPtr`
`a3` --> `VPtr<> tlsPtr`

side question: why is `tlsPtr` stored in `a3` and `ctidPtr` in `a4`? `tlsPtr` appears after `ctidPtr` in the function signature of `doClone`.

2. example
https://github.com/gem5/gem5/blob/bbde68c08c3e1b3e0ffb7f884f8488efb9d301d1/src/sim/syscall_emul.hh#L1850-L1855
```C++
_alloc_mem:
  li      a0, 0
  li      a1, MEM_SIZE
  li      a2, MMAP_PROT_FLAGS
  li      a3, MMAP_MAP_FLAGS
  li      a4, -1
  li      a5, 0
  li      a7, SYSCALL_MMAP
  ecall
  ret
```
`a0` --> `VPtr<> start`
`a1` --> `typename OS::size_t length`
`a2` --> `int prot`
`a3` --> `int tgt_flags`
`a4` --> `int tgt_fd`
`a5` --> `OS::off_t offset`

giactra:
Thanks @robhau  for your changes. Can I ask you to split the PR in multiple commits? For example I believe there should be at least:

Commit 1: Introducing GuestAddr
Commit 2: Any RISCV fixes

I believe this will make it easier for gerriters like myself to review the PR. You could also have two separate PRs if you want but it is really up to you (this might make it easier to provide a descriptive title in the PR)

robhau:
> Thanks @robhau for your changes. Can I ask you to split the PR in multiple commits? For example I believe there should be at least:
> 
> Commit 1: Introducing GuestAddr Commit 2: Any RISCV fixes
> 
> I believe this will make it easier for gerriters like myself to review the PR. You could also have two separate PRs if you want but it is really up to you (this might make it easier to provide a descriptive title in the PR)

Yes I can do that ☺️.

rogerchang23424 (approved):

rogerchang23424:
Hi,

I have create the PR in gem5-resources to update the m5 operation for the asmtest write_host method to avoid failed after the PR merged
https://github.com/gem5/gem5-resources/pull/25/commits/3114c821e71c7d41be0f92b84262c370bcf4e43e

ivanaamit:
@giactra, could you please re-review this PR? Thank you.

src/arch/arm/reg_abi.hh
```diff
@@ -54,11 +55,13 @@ struct RegABI64 : public GenericSyscallABI64
...
 {
 
+using namespace pseudo_inst;
+
 template <typename ABI, typename Arg>
 struct Argument<ABI, Arg,
     typename std::enable_if_t<
         std::is_base_of_v<ArmISA::RegABI32, ABI> &&
-        std::is_integral_v<Arg> &&
+        (std::is_integral_v<Arg> || std::is_same<Arg,GuestAddr>::value) &&
```
giactra: I believe this is wrong.
Here the behaviour between is_integral_v and GuestAddr should be different/opposite.

In the first case, we really want an uint64_t and this is achieved by using 2 32bit args register. In the GuestAddr case, as the Addr is 32 bit only, we want the LSB only and only one reg should be used...

src/sim/pseudo_inst.hh
```diff
@@ -64,18 +64,29 @@ decodeAddrOffset(Addr offset, uint8_t &func)
     func = bits(offset, 15, 8);
 }
 
+struct GuestAddr
+{
+  uint64_t addr;
```
giactra: 4 whitespace needed

src/sim/syscall_abi.hh
```diff
@@ -78,19 +79,22 @@ struct GenericSyscallABI32 : public GenericSyscallABI
...
+        (std::is_integral_v<Arg> || std::is_same<Arg,GuestAddr>::value)>>
 {
     static Arg
     get(ThreadContext *tc, typename ABI::State &state)
     {
         panic_if(state >= ABI::ArgumentRegs.size(),
                 "Ran out of syscall argument registers.");
-        return tc->getReg(ABI::ArgumentRegs[state++]);
+        auto arg = tc->getReg(ABI::ArgumentRegs[state++]);
+        return *reinterpret_cast<Arg*>(&arg);
```
giactra: This is ok, but I would honestly pay the price of duplication and treat is_integral and GuestAddr separately, so that you are not forced to use the reinterpret_cast trick

giactra (changes requested):

src/arch/riscv/se_workload.hh
```diff
@@ -61,14 +62,33 @@ class SEWorkload : public gem5::SEWorkload
...
+        std::is_integral_v<Arg> &&
+        ABI::template IsWideV<Arg>>>
+{
+    static Arg
+    get(ThreadContext *tc, typename ABI::State &state)
+    {
+        return (Arg) bits(tc->getReg(ABI::ArgumentRegs[state++]), 31, 0);
+    }
+};
+
```
rogerchang23424, line 91: Sorry I need to request you to change the RISC-V system ABI call. I think it's not a good way to handle the `uint64_t` and `int64_t` in RV32 SE mode.

I have check the type of `size_t` and `off_t` in ARM and X86 Linux Call

ARMLinux32
https://github.com/gem5/gem5/blob/0ec8cf8d05bb76b16d41d799d99323867c50c885/src/arch/arm/linux/linux.hh#L115-L120

X86Linux32
https://github.com/gem5/gem5/blob/0ec8cf8d05bb76b16d41d799d99323867c50c885/src/arch/x86/linux/linux.hh#L345-L349

Can we just declare the size_t and off_t as `uint32_t` in 'RiscvLinux32` instead of change `RiscvISA::SEWorkload::SyscallABI32`?

rogerchang23424 (changes requested):

src/arch/arm/reg_abi.hh
robhau: Hi,

I added this since this function was also called before my modifications because of `ABI::template IsWideV<Arg>` and `typedef uint64_t Addr`. (or do I miss something?).

However, you are right, the pointer size should be only 32 bit large.

src/sim/syscall_abi.hh
robhau: The reinterpret_cast would be still necessary in the get function for GuestAddr, because it is not possible to cast uint64_t to GuestAddr. If you are OK with it, i would like to keep both cases in one function.

src/arch/riscv/reg_abi.hh
```diff
@@ -68,7 +70,29 @@ struct Argument<ABI, Arg,
...
+        ABI::template IsWideV<Arg>>>
+{
+    static Arg
+    get(ThreadContext *tc, typename ABI::State &state)
+    {
+        panic_if(state >= ABI::ArgumentRegs.size(),
+                "Ran out of syscall argument registers.");
+
+        auto arg = bits(tc->getReg(ABI::ArgumentRegs[state++]), 31, 0);
+        return *reinterpret_cast<Arg*>(&arg);
```
rogerchang23424: Sorry for the wrong comment for `off_t`. Please ignore latest one.

the type of `time_t` and `off_t` are changed to 64bits [link](https://patchwork.ozlabs.org/project/glibc/patch/4e95f95966d8d7c6a8339160dc62d81c1f6a1bfb.1578824547.git.alistair.francis@wdc.com/). The syscall of [macros/mt](https://github.com/gem5/gem5-resources/tree/stable/src/asmtest/isa/macros/mt) need to be updated

The type of size_t still 32 bits

rogerchang23424 (changes requested):

rogerchang23424 (approved):
Thanks @robhau. Let @giactra do final review

ivanaamit:
@giactra, could you please review this PR? Thanks.

src/sim/syscall_abi.hh
giactra: I don't have super strong objection against it, I am just trying to avoid reinterpret casting as much as possible in cases where this is not needed.

Unless I am missing something you don't need casting for GuestAddr if you do:

        auto arg = tc->getReg(ABI::ArgumentRegs[state++]);
        return GuestAddr{arg};

Alternatively you can avoid casting by providing conversion operators(ish) to the GuestAddr. I didn't check but I believe that if you do

        struct GuestAddr
        {
            GuestAddr(Addr _addr) : addr(_addr) {}
            Addr addr;     
        };

Doing the following should just work:

       retrurn tc->getReg(ABI::ArgumentRegs[state++]);

src/arch/arm/reg_abi.hh
```diff
@@ -54,6 +55,8 @@ struct RegABI64 : public GenericSyscallABI64
 namespace guest_abi
 {
 
+using namespace pseudo_inst;
```
giactra: Relying on "using namespace" in header files is discouraged as it can lead to all sorts of issues with chain inclusion. Please remove it here in in other parts of the PR

src/sim/pseudo_inst.hh
```diff
@@ -64,18 +64,29 @@ decodeAddrOffset(Addr offset, uint8_t &func)
     func = bits(offset, 15, 8);
 }
 
+struct GuestAddr
+{
+    uint64_t addr;
+};
```
giactra: As we agreed this is a temporary patch to the problem, maybe we should add a comment here detailing why this Addr wrapper is needed

giactra (changes requested):
I have 3 comments from my side but I believe we are almost there. As this is touching other ISAs as well, it would be nice to have a review from other ISA maintainers though.

powerjg (commented):
It looks good to me, but I'll let Giacomo sign off

giactra (approved):
Thanks for your changes and patience! LGTM

PR #901, opened 2024-02-28
tests: Update tests to use specific resource versions

Harshil2107:
This update modifies the test configuration to specify the versions of resources used, rather than automatically using the latest versions. 
Previously, if a resource was updated for a change, it could potentially cause tests to fail if those tests were incompatible with the new version of the resource. 
Now, with this change, tests are tied to specific versions of resources, ensuring that any updates to resources will require corresponding updates to the tests to maintain compatibility.

Change-Id: I9633b1749f6c6c82af6aa6697b7e7656020f62fa

### DIFF
--- a/tests/gem5/checkpoint_tests/configs/power-hello-restore-checkpoint.py
+++ b/tests/gem5/checkpoint_tests/configs/power-hello-restore-checkpoint.py
@@ -61,7 +61,7 @@ board = SimpleBoard(
     cache_hierarchy=cache_hierarchy,
 )
 board.set_se_binary_workload(
-    obtain_resource("power-hello"),
+    obtain_resource("power-hello", resource_version="1.0.0"),
     checkpoint=obtain_resource(
         "power-hello-test-checkpoint-v24-0", resource_version="2.0.0"
     ),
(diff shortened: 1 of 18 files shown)


tests/gem5/checkpoint_tests/configs/x86-hello-restore-checkpoint.py
```diff
@@ -62,7 +62,9 @@
 )
 board.set_se_binary_workload(
     obtain_resource("x86-hello64-static"),
```
powerjg: Why not add a version to this one?

tests/gem5/checkpoint_tests/configs/x86-fs-restore-checkpoint.py
```diff
@@ -76,7 +76,9 @@
 board.set_kernel_disk_workload(
     kernel=obtain_resource("x86-linux-kernel-5.4.49"),
     disk_image=obtain_resource("x86-ubuntu-18.04-img"),
```
powerjg: Shouldn't these have versions too?

tests/gem5/checkpoint_tests/configs/power-hello-restore-checkpoint.py
```diff
@@ -62,7 +62,9 @@
 )
 board.set_se_binary_workload(
     obtain_resource("power-hello"),
```
powerjg: Shouldn't this have a version?

powerjg (commented):
I pointed out some below, but there are many `obtain_resources` calls that don't have versions. I suggest in the **tests** we should have explicit versions everywhere.

ivanaamit (approved):
lgtm

PR #902, opened 2024-02-28
dev: RegisterBank addRegistersAt for fragmented reg banks

giactra:
One of the limitations of the RegBank class is that it does not allow you to pass a non-contiguous set of registers.  Its simplest form will just accept an initializer list of registers and it will store them in sequence.

A more refined version [1] will optionally accept an offset value to be passed alongside the register reference. This is not meant to be used by the register bank to store the register at the provided offset.

It is rather used by the bank to sanity check the register sits exactly at the provided range.

The way to work around this for a fragemented register space is to explicitly allocate RAZ/RAO blocks as registers and to pass them to addRegisters together with the others.  (See the SysSecCtrl [2] as an example)

This makes it a bit tedious to model a register bank with gaps between its registers. First, the exact number and position of the gaps needs to be extraced from a spec. These sometimes report only implemented registers and their offset, and omit to document gaps/reserved space. So a developer needs to manually add register offset and size to check if all registers are contiguous.  Second, these reserved register blocks need to be instantiated in the bank adding boilerplate code and affecting readibility.

For these reasons we add a new registration method, called addRegisters*At*.  It reuses the RegisterAdder class but this time the offset field is really used to instruct the bank where the register should be mapped. The method is templated and the template parameter tells the bank which register type should be used to fill the remaining space. We make the RegBank the owner of this filler space (registers are generated internally within addRegistersAt).

[1]: https://github.com/gem5/gem5/blob/stable/src/dev/reg_bank.hh#L106
[2]: https://github.com/gem5/gem5/blob/stable/src/dev/arm/ssc.cc#L48

Change-Id: I614ae6e9eeb40b365ac9b6dd8b75abbfdb9cb687

### DIFF
--- a/src/dev/reg_bank.hh
+++ b/src/dev/reg_bank.hh
@@ -1,4 +1,16 @@
 /*
+ * Copyright (c) 2024 Arm Limited
+ * All rights reserved
+ *
+ * The license below extends only to copyright in the software and shall
+ * not be construed as granting a license to any other intellectual
+ * property including but not limited to intellectual property relating
+ * to a hardware implementation of the functionality of the software
+ * licensed hereunder.  You may use the software subject to the license
+ * terms below provided that you ensure that this notice is replicated
+ * unmodified and in its entirety in all distributions of the software,
+ * modified or unmodified, in source code or in binary form.
+ *
  * Copyright 2020 Google, Inc.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -41,6 +53,7 @@
 #include <sstream>
 #include <utility>
 
+#include "base/addr_range.hh"
 #include "base/bitfield.hh"
 #include "base/debug.hh"
 #include "base/logging.hh"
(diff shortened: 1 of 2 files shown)


src/dev/reg_bank.hh
```diff
@@ -955,6 +978,47 @@ class RegisterBank : public RegisterBankBase
...
+
+        std::vector<RegisterAdder> vec{adders};
+        std::sort(vec.begin(), vec.end(),
+            [] (const auto& first, const auto& second) {
+                return first.offset.value() < second.offset.value();
+            }
+        );
+
+        for (auto &adder: vec) {
+            assert(adder.offset && adder.reg);
```
powerjg, line 996: ~~Does this mean the offset cannot be 0?~~

~~I *think* that's what the code says, if this isn't intended you may want to take a closer look.~~

Ok, I see this is a `std::optional`. The code is OK, but I think it's confusing. It is not clear from this code (without spending the time to dig into the definition of `RegisterAdder` that calling `adder.offset` will give you whether or not an offset exists. Someone outside coming to this code would expect to find the `offset` value.

I suppose there's no action here, but I'll leave the note for other reviewers.

powerjg (approved):

src/dev/reg_bank.hh
giactra: Yeah I agree. To be fair, I didn't remove the old addRegisters flavour, but IMHO by adding this new utility, passing the { offset, reg } pair to addRegisters loses most of its meaning... If we removed it, there won't be any need to use optionals. The best way to make sure your offsets are respected is... to use addRegistersAt! Anyway I would defer this discussion until someone decides to deprecate it.

PR #908, opened 2024-03-04
misc: Tag checkpoints with the ISA of the CPUs

giactra:
With the introduction of multi-ISA gem5, we don't store the TARGET_ISA
anymore as a string in the root section of the checkpoint [1].  There is
therefore no way at the moment to asses the ISA of a CPU/ThreadContext.
This is a problem when it comes to checkpoint updates which are ISA
specific.

By explicitly serializing the ISA as a string under the cpu.isa section
we avoid this problem and we let cpt_upgraders be aware of the ISA in
use.

[1]: https://gem5-review.googlesource.com/c/public/gem5/+/48884

### DIFF
--- a/src/arch/arm/isa.cc
+++ b/src/arch/arm/isa.cc
@@ -80,7 +80,7 @@ RegClass floatRegClass(FloatRegClass, FloatRegClassName, 0, debug::FloatRegs);
 
 } // anonymous namespace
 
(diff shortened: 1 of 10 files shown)


powerjg (approved):
Looks good to me. We'll have to take a bit of care with the current checkpoints used for testing when merging this, but I'm happy for it to go in as is.

Issue #382, opened 2023-10-02
Adding Vector Segmented Loads/Stores to RISC-V V 1.0 implementation


ivanfv:
Hello,

I am opening this issue to ask if someone is working on the Vector Segmented Loads/Stores to the RISC-V Vector 1.0 implementation (e.g., vlseg, vsseg), which, if I am not wrong, are not supported by gem5 yet.

For the moment being, I am checking 1) how LD2, LD3 and LD4 Arm's instructions are implemented and 2) how vector register gather could help.

I will report the advances over here once I have significant progress.

Best,
Ivan

CC @adriaarmejach

ivanaamit:
Hi @ivanfv, I'm just checking to see the status of this issue. Are you currently working on it? Thanks.

ivanfv:
Hi @ivanaamit , yes, I am working on this. Currently, I am validating the vector segmented loads implementation and I plan to create a pull request about them very soon. Thanks for checking this!

ivanaamit:
Since #851 and #913 have been merged, I am closing this issue. @ivanfv, please reopen it if I am mistaken. Thanks.

PR #913, opened 2024-03-05
arch-riscv: adding vector unit-stride segment stores to RISC-V

ivanfv:
This commit adds support for vector unit-stride segment store operations for RISC-V (vssegXeXX). This implementation is based in two types of microops:
- VsSegIntrlv microops that properly interleave source registers into structs.
- VsSeg microops that store data in memory as contiguous structs of several fields.

Change-Id: Id80dd4e781743a60eb76c18b6a28061f8e9f723d

Gem5 issue: https://github.com/gem5/gem5/issues/382

### DIFF
--- a/src/arch/riscv/utility.hh
+++ b/src/arch/riscv/utility.hh
@@ -308,7 +308,7 @@ elem_mask(const T* vs, const int index)
 
 template<typename T>
 inline int
-elem_mask_vlseg(const T* vs, const int elem, const int num_fields)
+elem_mask_vseg(const T* vs, const int elem, const int num_fields)
 {
     int index = floor(elem / num_fields);
     static_assert(std::is_integral_v<T>);
--- a/src/arch/riscv/insts/vector.cc
+++ b/src/arch/riscv/insts/vector.cc
@@ -501,7 +501,6 @@ VxsatMicroInst::generateDisassembly(Addr pc,
     return ss.str();
 }
 
-
 VlFFTrimVlMicroOp::VlFFTrimVlMicroOp(ExtMachInst _machInst, uint32_t _microVl,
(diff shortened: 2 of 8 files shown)


powerjg:
@hnpl can you take a look at this?

src/arch/riscv/utility.hh
```diff
@@ -308,7 +308,7 @@ elem_mask(const T* vs, const int index)
 
 template<typename T>
 inline int
-elem_mask_vlseg(const T* vs, const int elem, const int num_fields)
+elem_mask_vseg(const T* vs, const int elem, const int num_fields)
```
rogerchang23424, line 311: (nit) Why you need to check the function name?

ivanfv: The thing is that I created `elem_mask_vlseg` when implementing the load operations (vlseg -> vector load segmented) but the store operations (vsseg) also use the same function, so I wanted to make the name more generic and cover both (vseg -> vector segmented). 

In any case, let me know if you would prefer another naming for it.

rogerchang23424: Thanks for replying. No further change required for me. I just want to know the reason of naming change.

rogerchang23424 (approved):

aarmejach:
I'd be great if we could make progress on this one as #589 depends on this one. And we have some additional PRs that in turn depend on #589 :)

powerjg:
It looks OK to me, but I haven't had time to look closely. I'm hoping that @hnpl can take a look today. Either way, we can merge on Monday if there are no other issues.

hnpl (approved):
lgtm!

powerjg (approved):

PR #914, opened 2024-03-05
arch-riscv: Move alignment check to Physical Memory Attribute(PMA)

rogerchang23424:
In the RISC-V unprivileged spec[1], the misaligned load/store support is depend on the EEI.
    
In the RISC-V privileged spec ver1.12[2], the PMA specify wether the misaligned access is support for each data width and the memory region.
    
In the [3] of `mcause` spec, we cloud directly raise misalign exception if there is no memory region misalignment support. If the part of memory region support misaligned-access, we need to translate the `vaddr` to `paddr` first then check the `paddr` later. The page-fault or access-fault is rose before misalign-fault.
    
The benefit of moving check_alignment option from ISA option to PMA option is we can specify the part region of memory support misalign load/store.

MMU will check alignment with virtual addresss if there is no misaligned memory region specified. If there are some misaligned memory region supported, translate address first and check alignment at final.
    
[1] https://github.com/riscv/riscv-isa-manual/blob/main/src/rv32.adoc#base-instruction-formats
[2] https://github.com/riscv/riscv-isa-manual/blob/main/src/machine.adoc#physical-memory-attributes
[3] https://github.com/riscv/riscv-isa-manual/blob/main/src/machine.adoc#machine-cause-register-mcause

### DIFF
--- a/src/arch/riscv/PMAChecker.py
+++ b/src/arch/riscv/PMAChecker.py
@@ -53,3 +53,7 @@ class PMAChecker(BasePMAChecker):
     cxx_class = "gem5::RiscvISA::PMAChecker"
 
     uncacheable = VectorParam.AddrRange([], "Uncacheable address ranges")
+    misaligned = VectorParam.AddrRange(
+        [],
+        "Address ranges support misaligned load/store to memory",
+    )
--- a/src/arch/riscv/pma_checker.cc
+++ b/src/arch/riscv/pma_checker.cc
@@ -37,6 +37,8 @@
 
 #include "arch/riscv/pma_checker.hh"
 
+#include "arch/riscv/faults.hh"
+#include "arch/riscv/mmu.hh"
 #include "base/addr_range.hh"
 #include "base/types.hh"
 #include "mem/packet.hh"
@@ -54,14 +56,41 @@ PMAChecker::PMAChecker(const Params &params) :
 BasePMAChecker(params),
 uncacheable(params.uncacheable.begin(), params.uncacheable.end())
 {
+    for (auto& range: params.misaligned) {
+        misaligned.insert(range, true);
+    }
 }
 
-void
-PMAChecker::check(const RequestPtr &req)
+Fault
+PMAChecker::check(const RequestPtr &req, BaseMMU::Mode mode, Addr vaddr)
 {
     if (isUncacheable(req->getPaddr(), req->getSize())) {
         req->setFlags(Request::UNCACHEABLE | Request::STRICT_ORDER);
     }
+
+    return hasMisaligned() ? checkPAddrAlignment(req, mode, vaddr) : NoFault;
+}
+
+Fault
+PMAChecker::checkVAddrAlignment(
+    const RequestPtr &req, BaseMMU::Mode mode)
+{
+    // We need to translate address before alignment check
+    // if there are some memory ranges support misaligned load/store
+    if (hasMisaligned()) {
+        return NoFault;
+    }
+
+    // Ingore alignment check for instruction fetching
+    if (mode == BaseMMU::Execute) {
+        return NoFault;
+    }
+    assert(req->hasVaddr());
+    Addr alignSize = mask(req->getArchFlags() & MMU::AlignmentMask) + 1;
+    if (addressAlign(req->getVaddr(), alignSize)) {
+        return NoFault;
+    }
+    return createMisalignFault(req->getVaddr(), mode);
 }
 
 bool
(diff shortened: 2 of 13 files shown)


src/arch/riscv/PMAChecker.py
```diff
@@ -53,3 +53,7 @@ class PMAChecker(BasePMAChecker):
     cxx_class = "gem5::RiscvISA::PMAChecker"
 
     uncacheable = VectorParam.AddrRange([], "Uncacheable address ranges")
+    misaligned = VectorParam.AddrRange(
+        [],
+        "Misaligned load/store address ranges",
```
powerjg: I think this help string should be expanded a bit. For instance, it's unclear to me without reading the rest of this change whether these ranges will *allow* or *disallow* unaligned load/store

src/arch/riscv/pma_checker.cc
```diff
@@ -54,14 +55,19 @@ PMAChecker::PMAChecker(const Params &params) :
...
 {
+    for (auto& range: params.misaligned) {
+        misaligned.insert(range, true);
+    }
 }
 
-void
-PMAChecker::check(const RequestPtr &req)
+Fault
+PMAChecker::check(const RequestPtr &req, BaseMMU::Mode mode, Addr vaddr)
```
powerjg, line 65: Why do we need the `vaddr` argument? Isn't this in the `req`?

If this *is* needed for some reason that I can't see here, please add some detailed comments to the appropriate docstring in the header file.

src/python/gem5/components/boards/experimental/lupv_board.py
```diff
@@ -87,6 +87,7 @@ def __init__(
         processor: AbstractProcessor,
         memory: AbstractMemorySystem,
         cache_hierarchy: AbstractCacheHierarchy,
+        misaligned_ranges: List[AddrRange] = [],
```
powerjg: I don't think this should be an argument to all boards. The board should specify in its specialization what the allowed ranges for misalignment should be. I don't see any case that the *user* of this board will want to change this.

Another way to look at the stdlib, this board is the "motherboard" you would buy from the store. The motherboard defines things like the system information, it doesn't allow people to change it.

src/python/gem5/components/boards/riscv_board.py
```diff
@@ -85,6 +85,7 @@ def __init__(
         processor: AbstractProcessor,
         memory: AbstractMemorySystem,
         cache_hierarchy: AbstractCacheHierarchy,
+        misaligned_ranges: List[AddrRange] = [],
```
powerjg: Ditto above. This should not be an argument.

src/arch/riscv/PMAChecker.py
rogerchang23424: Changed to `Address ranges support misaligned load/store to memory`

src/arch/riscv/pma_checker.cc
rogerchang23424: In the `setupWalk` step of `PageTableWalker`, we don't intialize the virtual address of `read` packet request.

https://github.com/gem5/gem5/blob/f35815cd48a0716452a88594563c2c5131d69424/src/arch/riscv/pagetable_walker.cc#L494-L518

We need to specify the `vaddr` manually in the page table walker like `pmpCheck`

https://gem5-review.googlesource.com/c/public/gem5/+/46279

src/arch/riscv/pma_checker.cc
```diff
@@ -94,6 +118,42 @@ PMAChecker::takeOverFrom(BasePMAChecker *old)
...
+    const RequestPtr &req, BaseMMU::Mode mode, Addr vaddr)
+{
+    Addr addr = 0;
+    if (mode != BaseMMU::Execute) {
+      assert(req->hasPaddr());
+      addr = req->getPaddr();
+      Addr alignSize = mask(req->getArchFlags() & 0x7) + 1;
+      auto it = misaligned.contains(RangeSize(addr, req->getSize()));
+      if (it == misaligned.end()) {
+          return checkAlignmentInternal(
```
powerjg: I think this is not correct.

From my understanding around the discussion on [the previous change](https://gem5-review.googlesource.com/c/public/gem5/+/46279/comments/084355fe_df1c3950), the `vaddr` is only used as the value that is returned in the fault. It should be be used as the address we're checking.

For instance, on the PTW, we need to check for alignment on the physical address, but raise a fault which has the virtual address.

I *think* this code is actually checking the `vaddr` for alignment, not the physical address that the PTW is using.

src/arch/riscv/pma_checker.cc
powerjg: Thanks for pointing me to that review. I see I asked the same question there, which shows this is confusing!

However, I am not convinced this code is performing the correct actions. Can you see my comment below?

src/arch/riscv/pma_checker.cc
rogerchang23424: Thanks @powerjg point out the issue of checking physicall address alignment. I have updated the patch to fix the issue.

powerjg (dismissed):
Thanks for the updates!

src/arch/riscv/pma_checker.cc
```diff
@@ -94,6 +123,61 @@ PMAChecker::takeOverFrom(BasePMAChecker *old)
...
+      addr = req->getPaddr();
+      Addr alignSize = mask(req->getArchFlags() & MMU::AlignmentMask) + 1;
+      if (
+          !addressAlign(addr, alignSize) &&
+          !misAlignedSupport(RangeSize(addr, req->getSize()))
+      ) {
+          return createMisalignFault(
+              (req->hasVaddr() ? req->getVaddr() : vaddr), mode);
+      }
+    }
```
powerjg: Sorry I didn't catch this sooner! We're missing spaces here.

rogerchang23424: Hmmm... I have added some comments on the `pma_checker.hh`. Let me know if you have any consideration

powerjg (approved):

PR #920, opened 2024-03-08
dev-arm: Handle translation aborts and add IRQ support to the SMMU

giactra:
At the moment the SMMU is not handling translation errors gracefully the way it is described by the SMMUv3 spec: whenever a translation fault arises, simulation aborts as a whole. With this PR we add minimal support for
translation fault handling, which means:

1) Not aborting simulation, but rather:
2) Writing an event entry to the SMMU_EVENTQ (event queue)
3) Signaling the PE an error arose and there is an event entry to be consumed. The signaling is achieved
with the addition of the eventq SPI. Using an MSI is also possible though it is currently disabled by the SMMU_IDR0.MSI being set to zero.

The PR is addressing issues reported by https://github.com/orgs/gem5/discussions/898

### DIFF
--- a/src/dev/arm/RealView.py
+++ b/src/dev/arm/RealView.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2009-2022 Arm Limited
+# Copyright (c) 2009-2022, 2024 Arm Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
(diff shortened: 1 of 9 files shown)


andysan (dismissed):

andysan (approved):

powerjg (approved):
lgtm

PR #925, opened 2024-03-11
python,stdlib: Support added for multiple disk boards

amatabsc:
This commit introduces the initial support for using multiple disks on different gem5 boards. Currently, only the necessary infrastructure has been added to support multiple disks on the board without implementing full functionality.
Procedure for adding additional functionality:
```python
#Your _add_disk_to_board and _finalize_workload from your_board.py should be something like this:
    def _add_disk_to_board(self, disk_image: AbstractResource, is_root: bool): 
        image = CowDiskImage(
            child=RawDiskImage(read_only=True), read_only=False
        )

        image.child.image_file = disk_image.get_local_path()

        if is_root:
            self.disk.vio.image = image

        else:
            idx = len(self._disks)
            new_disk_pio_addr = 0x10009000 + 0x1000 * len(self._disks)
            new_disk = RiscvMmioVirtIO(
                vio=VirtIOBlock(),
                interrupt_id=0x8,
                pio_size=0x1000,
                pio_addr=new_disk_pio_addr,
            )

            new_disk.vio.image = image

            setattr(self, f"disk{idx}", new_disk)
            disk = getattr(self, f"disk{idx}")

            self._disks.append(disk)

    def _finalize_workload(self): #You should create this new method
        # Connect the additional disks added to the board
        self._off_chip_devices += self._disks
        self._setup_io_devices()
        self._setup_pma()

        self.generate_device_tree(m5.options.outdir)
        self.workload.dtb_filename = os.path.join(
            m5.options.outdir, "device.dtb"
        )

        self.workload.dtb_addr = 0x87E00000
```
cc: @aarmejach 
Change-Id: I21d30660bb4dc126ce418f3252f0feb65b41ee0c


src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -110,6 +110,17 @@ def _add_disk_to_board(self, disk_image: DiskImageResource) -> None:
...
+    def _finalize_workload(self) -> None:
+        """
+        Finalizes the board setup dependent on the workload.
+        This function connects the IO devices to the board, generates the
+        device tree and expose the location of the dtb to the gem5's workload
+        object. This function should be called after all IO devices are set up
+        (e.g. disk images, PIO devices.)
+        """
+        raise NotImplementedError
+
```
BobbyRBruce: I'm not sure if this is necessary as I believe you could change where the disk images are added inside the board itself, though I may be wrong.

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -145,6 +156,7 @@ def set_kernel_disk_workload(
         disk_image: DiskImageResource,
         bootloader: Optional[BootloaderResource] = None,
         disk_device: Optional[str] = None,
+        additional_disks: List[DiskImageResource] = [],
```
BobbyRBruce: The docstring needs updated here. It should be added that the `disk_device` is assumed to contain the image with the root to be mounted. `additional_disks` needs documentation too.

BobbyRBruce (commented):
Am I correct in thinking this is still a WIP/Draft? I'm wondering how much feedback you want here for now.

I'm in strong support of this endevour. In general I think there's a lot we can do to improve the `KernelDiskWorkload` class. Notably how the device path is determined (I really have no good answers for this, it's a mess right now).

src/python/gem5/components/boards/kernel_disk_workload.py
amatabsc: I have implemented that as you recommended. Thank you for the advice.

src/python/gem5/components/boards/kernel_disk_workload.py
amatabsc: Done, I have updated the description of the disk_image and additional_disks to address the new functionality.

amatabsc:
> Am I correct in thinking this is still a WIP/Draft? I'm wondering how much feedback you want here for now.
> 
> I'm in strong support of this endevour. In general I think there's a lot we can do to improve the `KernelDiskWorkload` class. Notably how the device path is determined (I really have no good answers for this, it's a mess right now).

This board changes are fully supported in riscv. We are already using this changes locally, but I was not sure if you wanted to modify any existing board to use it. Although, I have added the changes in the riscv_matched board as an example. Let me know if I should proceed in a different way.

src/python/gem5/prebuilt/riscvmatched/riscvmatched_board.py
```diff
@@ -149,9 +152,7 @@ def _setup_board(self) -> None:
             # Contains a CLINT, PLIC, UART, and some functions for the dtb, etc.
             self.platform = HiFive()
             # Note: This only works with single threaded cores.
-            self.platform.plic.hart_config = ",".join(
-                ["MS" for _ in range(self.processor.get_num_cores())]
-            )
+            self.platform.plic.n_contexts = self.processor.get_num_cores() * 2
```
BobbyRBruce: I think you included this by accident

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -145,6 +147,7 @@ def set_kernel_disk_workload(
...
```
BobbyRBruce: Sorry to change my mind here, it's a small suggestion

```suggestion
        additional_disk_images: List[DiskImageResource] = [],
```

It's just a bit clearer and makes the link between it and `disk_image` more apparent.

Also, you should add this to the end of the parameters. If someone has been calling this function with positional arguments then their scripts may break (i.e., it'll look like their trying to pass the `readfile` as `additional_disks`.

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -66,8 +66,6 @@ def get_default_kernel_args(self) -> List[str]:
                     "root={root_value}",
                 ]
             ...
-
-
```
BobbyRBruce: Re-add these lines just to cleanup the diff.

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -97,7 +95,11 @@ def get_disk_device(self) -> str:
         raise NotImplementedError
 
     @abstractmethod
-    def _add_disk_to_board(self, disk_image: DiskImageResource) -> None:
```
BobbyRBruce, line 100: Let's do this a bit different. Here's my worry: By doing this we're forcing change in all boards that inherit from `KernelDiskWorkload` to support multiple disks. This may not always be wanted and certainly won't be the case when this is first implemented.

How about:

```py

def _add_disk_to_board(self, disk_image: Disk ImageResource) -> None:
    """
    Sets the configuration needed to add the disk image to the board.
    .. note::
        This will be executed at the end of the
        ``set_kernel_disk_workload`` function.
    :param disk_image: The disk image to add to the system.
    """
    raise NotImplementedError

def _supports_multiple_images(cls) -> bool:
   """Returns True if the board supports adding multiple additional
   disk images beyond just the root disk image.
   """
   return False

def _add_additional_disks_to_board(
    self,
    additional_disk_images: List[DiskImageResource],
) -> None:
    """<TODO SOME DOCSTRING HERE>"""
    assert self. _supports_multiple_images(), \
        "`_add_additional_disks_to_board` called when `_supports_multiple_images` returned0 False."
    raise NotImplementedError


# Then, down in the `set_kernel_disk_workload` function:

self._add_disk_to_board(disk_image=disk_image)
if self._supports_multiple_images():
    self._add_additional_disk_images_to_board(additional_images = additional_images)
elif len(additional_images) > 0:
    warn("This board does not support multiple disk images. Additional images will not be added to the board.")
```

With this, for boards that support multiple disk images, you'd just need to override the `_supports_multiple_disk_images` function to return `True`. Nothing would need to change in any boards we don't need or want to implement this feature yet.

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -97,7 +95,11 @@ def get_disk_device(self) -> str:
...
+    def _add_disk_to_board(
+        self,
+        disk_image: DiskImageResource,
+        bench_image: Optional[DiskImageResource],
```
BobbyRBruce: The typing here is a little wrong, I believe it should be a list of `DiskImageResources`. You can see my other comment on this but I think the name should be `additional_disk_images`, and the docstring in this function should be updated to explain things should also be updated.

BobbyRBruce (changes requested):

BobbyRBruce:
> > Am I correct in thinking this is still a WIP/Draft? I'm wondering how much feedback you want here for now.
> > I'm in strong support of this endevour. In general I think there's a lot we can do to improve the `KernelDiskWorkload` class. Notably how the device path is determined (I really have no good answers for this, it's a mess right now).
> 
> This board changes are fully supported in riscv. We are already using this changes locally, but I was not sure if you wanted to modify any existing board to use it. Although, I have added the changes in the riscv_matched board as an example. Let me know if I should proceed in a different way.



> > Am I correct in thinking this is still a WIP/Draft? I'm wondering how much feedback you want here for now.
> > I'm in strong support of this endevour. In general I think there's a lot we can do to improve the `KernelDiskWorkload` class. Notably how the device path is determined (I really have no good answers for this, it's a mess right now).
> 
> This board changes are fully supported in riscv. We are already using this changes locally, but I was not sure if you wanted to modify any existing board to use it. Although, I have added the changes in the riscv_matched board as an example. Let me know if I should proceed in a different way.

I've made some suggested changes on the PR via my review.I think it'd be good to split this change up into 1 commit which makes the necessary changes to the the `KernelDiskWorkload` class and one additional commit for each board updated to utilize this multiple disk image feature. I don't think updating every board to use this is necessary for this PR to be merged (see my review comments for more context on how to implement this). 

I would ask for some tests to be included to verify the multiple disk image feature for the boards you submit actually work. I can help you with that: you'd need to provide me with a config script which utilizes multiple disk images on a board and some way to verify that the images have been loaded. This can wait a bit until all the changes have been implemented but should be thought about.

src/python/gem5/components/boards/kernel_disk_workload.py
amatabsc: I think your idea is very interesting, I've already applied it.

src/python/gem5/components/boards/kernel_disk_workload.py
amatabsc: Done!

src/python/gem5/prebuilt/riscvmatched/riscvmatched_board.py
amatabsc: Sorry, my bad

src/python/gem5/components/boards/kernel_disk_workload.py
amatabsc: Done!

amatabsc:
> > > Am I correct in thinking this is still a WIP/Draft? I'm wondering how much feedback you want here for now.
> > > I'm in strong support of this endevour. In general I think there's a lot we can do to improve the `KernelDiskWorkload` class. Notably how the device path is determined (I really have no good answers for this, it's a mess right now).
> > 
> > 
> > This board changes are fully supported in riscv. We are already using this changes locally, but I was not sure if you wanted to modify any existing board to use it. Although, I have added the changes in the riscv_matched board as an example. Let me know if I should proceed in a different way.
> 
> > > Am I correct in thinking this is still a WIP/Draft? I'm wondering how much feedback you want here for now.
> > > I'm in strong support of this endevour. In general I think there's a lot we can do to improve the `KernelDiskWorkload` class. Notably how the device path is determined (I really have no good answers for this, it's a mess right now).
> > 
> > 
> > This board changes are fully supported in riscv. We are already using this changes locally, but I was not sure if you wanted to modify any existing board to use it. Although, I have added the changes in the riscv_matched board as an example. Let me know if I should proceed in a different way.
> 
> I've made some suggested changes on the PR via my review.I think it'd be good to split this change up into 1 commit which makes the necessary changes to the the `KernelDiskWorkload` class and one additional commit for each board updated to utilize this multiple disk image feature. I don't think updating every board to use this is necessary for this PR to be merged (see my review comments for more context on how to implement this).
> 
> I would ask for some tests to be included to verify the multiple disk image feature for the boards you submit actually work. I can help you with that: you'd need to provide me with a config script which utilizes multiple disk images on a board and some way to verify that the images have been loaded. This can wait a bit until all the changes have been implemented but should be thought about.

I've prepared the following script, have in mind that the rootfs.img and the benchmarks.img should have your own route.
```python
import argparse

from gem5.isas import ISA
from gem5.resources.resource import (
    obtain_resource,
    DiskImageResource,
)

parser = argparse.ArgumentParser(
    description="A script which uses the RISCVMatchedBoard in FS mode."
)

parser.add_argument(
    "-i",
    "--to-init",
    action="store_true",
    help="Exit the simulation after the Linux Kernel boot.",
)

args = parser.parse_args()

# instantiate the riscv matched board with default parameters
board = RISCVMatchedBoard(
    clk_freq="1.2GHz",
    l2_size="2MB",
    is_fs=True,
)
additional_disk_images = ["rootfs.img",
                          "benchmarks.img"
                         ]
disk_array = []
for i in range(len(additional_disk_images)):
    if i == 0:
        disk_array.append(
            DiskImageResource(
                local_path=additional_disk_images[i], root_partition="1"
            )
        )
    else:
        disk_array.append(
            DiskImageResource(
                local_path=additional_disk_images[i],
            )
        )
kernel=obtain_resource("riscv-bootloader-vmlinux-5.10")
    kernel=kernel,
    disk_image=disk_array[0],
    additional_disk_images=disk_array[1:],
)

simulator.run()

BobbyRBruce:
@amatabsc : I've added two commits to this PR. The first is just minor formatting changes (nothing functional). The other adds a test for booting multiple disks on the RISCVMatched board. This test is simple. The same disk image is mounted twice and when the system is finished booting, `lsblk` is run and the output checked to ensure both disks are present.

I'm going to rerun this test to double-check this works (RISCVMatched takes a long time to boot, so bare with me...)/

BobbyRBruce:
I've found getting test to verify this features is harder than I expected. I'm going to play around a bit with this configuration manually just to confirm this works but I may have to submit this without a test for now.

erin-le:
@amatabsc, @BobbyRBruce: This PR is being closed for inactivity. If anyone would like to continue working on this, please feel free to reopen it. Thank you!

PR #926, opened 2024-03-11
dev-amdgpu: Support for ROCm 6.0

abmerop:
Implement several features new in ROCm 6.0 and features required for future devices. Includes the following:

- Support for multiple command processors
- Improve handling of unknown register addresses
- Use AddrRange for MMIO address regions
- Handle GART writes through SDMA copy
- Implement PCIe indirect reads and writes
- Improve PM4 write to check dword count
- Implement common MI300X instruction


abmerop:
FYI, We will have to retest the checkpointing support.

configs/example/gpufs/system/system.py
```diff
@@ -180,9 +180,15 @@ def makeGpuFSSystem(args):
...
     system.pc.south_bridge.gpu.sdmas = sdma_engines
 
-    # Setup PM4 packet processor
-    pm4_pkt_proc = PM4PacketProcessor()
-    system.pc.south_bridge.gpu.pm4_pkt_proc = pm4_pkt_proc
+    # Setup PM4 packet processors
+    pm4_procs = []
+    pm4_procs.append(
+        PM4PacketProcessor(
+            ip_id=0, mmio_range=AddrRange(start=0xC000, end=0xD000)
```
mattsinc, line 195: is it ok for the address range to be the same for each of them?

src/dev/amdgpu/amdgpu_defines.hh
```diff
@@ -61,36 +71,11 @@ constexpr int MMIO_BAR = 5;
 constexpr uint32_t VGA_ROM_DEFAULT = 0xc0000;
 constexpr uint32_t ROM_SIZE = 0x20000;        // 128kB
 
-/* SDMA base, size, mmio offset shift. */
```
mattsinc, line 64: I'm assuming these are all just removed in ROCm 6?  If so, does removing them break GPUSE?

src/dev/amdgpu/amdgpu_nbio.cc
```diff
@@ -53,22 +53,38 @@ AMDGPUNbio::setGPUDevice(AMDGPUDevice *gpu_device)
...
+    // For Vega10 we rely on the golden values in an MMIO trace. Return
+    // immediately as to not clobber those values.
+    if (gpuDevice->getGfxVersion() == GfxVersion::gfx900) {
+        if (offset == AMDGPU_PCIE_DATA || offset == AMDGPU_PCIE_DATA2) {
+            return;
+        }
+    }
+
     switch (offset) {
-      // This is a PCIe status register. At some point during driver init
```
mattsinc, line 57: any reason we can't update comment for new name?

src/dev/amdgpu/amdgpu_nbio.cc
```diff
@@ -123,6 +139,18 @@ AMDGPUNbio::writeMMIO(PacketPtr pkt, Addr offset)
         DPRINTF(AMDGPUDevice, "MM write to reg %#lx data %#lx\n",
                 mm_index_reg, pkt->getLE<uint32_t>());
         gpuDevice->setRegVal(AMDGPU_MM_DATA, pkt->getLE<uint32_t>());
+    } else if (offset == AMDGPU_PCIE_INDEX) {
```
mattsinc, line 154: ideally we should comment what each of these represent

src/dev/amdgpu/pm4_packet_processor.cc
```diff
@@ -350,21 +352,46 @@ PM4PacketProcessor::decodeHeader(PM4Queue *q, PM4Header header)
...
-        [ = ](const uint32_t &) { writeDataDone(q, pkt, addr); });
-    //TODO: the specs indicate that pkt->data holds the number of dword that
-    //need to be written.
-    dmaWriteVirt(addr, sizeof(uint32_t), cb, &pkt->data);
+    DPRINTF(PM4PacketProcessor, "PM4 write addr: %p data: %p destSel: %d "
+            "addrIncr: %d resume: %d writeConfirm: %d cachePolicy: %d\n",
+            pkt->destAddr, pkt->data, pkt->destSel, pkt->addrIncr,
+            pkt->resume, pkt->writeConfirm, pkt->cachePolicy);
+
+    if (pkt->destSel == 5) {
```
mattsinc, line 364: it would be good if we explained what the magic number 5 represents

configs/example/gpufs/system/system.py
abmerop: Yes, these are the same for Vega10 and MI200 (what gem5 currently supports).  There are multiple PM4 packet processors in upcoming MI300 patch

src/dev/amdgpu/amdgpu_defines.hh
abmerop: They aren't removed, but I am getting rid of the hard-coded values here in favor of adding AddrRanges for each structure's MMIO range. Since they will be different in a future patch the values will need to be set on the python side


Nothing in src/dev/amdgpu will break SE mode as it's unused there

src/dev/amdgpu/amdgpu_nbio.cc
abmerop: This comment was actually completely incorrect.  The driver will do an indirect write to update the value properly.  This was from a very early patch to get things working.  The new name is to match the name in the driver code

src/dev/amdgpu/pm4_packet_processor.cc
abmerop: There is a comment below, but this isn't well documented so I don't have anything to point to.  One can infer where the write should go based on the address value though and/or looking at the driver code and searching for "WRITE_DATA_DST_SEL." For example 5 is very obviously a VRAM address

src/dev/amdgpu/amdgpu_nbio.cc
abmerop: Added comment in updated commits

src/dev/amdgpu/pm4_packet_processor.cc
mattsinc: It wasn't obvious to me, but if you feel like it becoming a GART request after is clear, then fine.

mattsinc (approved):

PR #927, opened 2024-03-11
Adding schedule_next_max_ticks to Simulator

mahyarsamani:
This commit extends the Simulator class to allow the user to set a max_tick with Simulator::run which will be the total sum of simulated ticks (through exiting due to exit events and re-enterin), and schedule next_max_ticks which allows the user to schedule simulation for next_max_ticks ticks from curTick.

Change-Id: I17334ac74d0a0c540cb67199f02ccdbc1209be32

### DIFF
--- a/src/python/gem5/simulate/simulator.py
+++ b/src/python/gem5/simulate/simulator.py
@@ -347,6 +347,10 @@ class Simulator:
         self._last_exit_event = None
         self._exit_event_count = 0
 
+        self._original_max_ticks = 0
+        self._remaining_max_tick = 0
+        self._next_max_tick = -1
+
         if checkpoint_path:
             warn(
                 "Setting the checkpoint path via the Simulator constructor is "
@@ -386,6 +390,20 @@ class Simulator:
         """
         for core in self._board.get_processor().get_cores():
             core._set_inst_stop_any_thread(inst, self._instantiated)
+
+    def schedule_next_max_ticks(self, ticks: int) -> None:
+        """
+        Schedule an ``MAX_TICK`` after running the simulation for certain
+        number of ticks from curTick.
+
+        :param tikcs: Number of ticks to simulate.
+        """
+        if self._next_max_tick != -1:
+            raise RuntimeError(
+                f"Already have a next_max_tick scheduled "
+                f"for {self._next_max_tick} ticks."
+            )
+        self._next_max_tick = ticks
 
     def get_stats(self) -> Dict:
         """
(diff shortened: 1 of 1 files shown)


src/python/gem5/simulate/simulator.py
```diff
@@ -568,11 +588,10 @@ def run(self, max_ticks: int = m5.MaxTick) -> None:
...
         events accordingly.
 
-        :param max_ticks: The maximum number of ticks to execute per simulation
-                          run. If this ``max_ticks`` value is met, a ``MAX_TICK``
+        :param max_ticks: The maximum number of ticks to simulate overall.
+                          If this ``max_ticks`` value is met, a ``MAX_TICK``
                           exit event is received, if another simulation exit
-                          event is met the tick count is reset. This is the
-                          **maximum number of ticks per simulation run**.
+                          event is met the tick count is reset.
```
powerjg, line 594: Is this changing the behavior? If so, it concerns me. Either way, the description *still* isn't clear.

mahyarsamani: Yes, it's changing the behavior. I am not hellbent on getting this merged, if you don't see any value in this I can close this.

powerjg: The problem with changing behavior is that there are many scripts out there that expect the old behavior. We could add a new parameter which has this behavior and deprecate `max_ticks`, but we can't just break people's current scripts.

mahyarsamani: I see, I can make that change.

BobbyRBruce:
I think it's fine to add this feature but i don't think you should interfere with max_ticks current behavior. Remember we want to extent, not modify, when adding a new feature.

I think we over-use `max_ticks` as, for a while, it was the more straight-forward way to schedule an exit based on ticks. Last year I added more functions to the `m5.simulate` module: https://github.com/gem5/gem5/commit/8479a691aa57c8f5763b71d26d579697ae96d007. I believe the `scheduleTickExitAbsolute` function there is what you want to use. Though the part you need to implement (and for which I'd welcome a PR), is extending the stdlib's `Simulator` API to utilize these functions.

powerjg:
@mahyarsamani I think this is taken care of in #1167 (https://github.com/gem5/gem5/pull/1167/files#diff-5f953e360b63f100c0d5b74c02d3d36e5c04ad79be1b5b09e0b981ab6b900ca1R432)

If so, can you close this PR?

mahyarsamani:
> @mahyarsamani I think this is taken care of in #1167 (https://github.com/gem5/gem5/pull/1167/files#diff-5f953e360b63f100c0d5b74c02d3d36e5c04ad79be1b5b09e0b981ab6b900ca1R432)
> 
> If so, can you close this PR?

Yep, I think so. Will do

PR #935, opened 2024-03-14
stdlib, tests, configs: Add a new PrivateL1PrivateL2WalkCache hierarchy

giactra:
From [1] The PrivateL1PrivateL2Cache hierarchy has been amended with an MMUCache, which is basically a small cache in front of the page table walker.

Not every ISA makes use of it.
Arm for example already implements caching of page table walks, via the partial_levels parameter in the ArmTLB.
With this patch we define a new module which explicitly makes use of the WalkCache. Configurations that do not require
another cache in the first level of the memsys (for the ptw) can use the PrivateL1PrivateL2CacheHierarchy
    
[1]: https://gem5-review.googlesource.com/c/public/gem5/+/49364

### DIFF
--- a/configs/example/gem5_library/checkpoints/simpoints-se-restore.py
+++ b/configs/example/gem5_library/checkpoints/simpoints-se-restore.py
@@ -60,8 +60,8 @@ from m5.stats import (
 )
 
 from gem5.components.boards.simple_board import SimpleBoard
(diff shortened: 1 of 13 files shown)


BobbyRBruce (approved):
Thanks! This is a great improvement!

giactra:
Thanks for your review Bobby!

PR #940, opened 2024-03-15
gpu-compute: Add support for skipping GPU kernels

mwboyer:
gpu-compute: Add support for skipping GPU kernels

This commit adds two new command-line options:

--skip-until-gpu-kernel N
Skips (non-blit) GPU kernels until the target kernel is reached.
Execution continues normally from there. Blit kernels are not skipped
because they are responsible for copying the kernel code and metadata
for the non-blit kernels. Note that skipping kernels can impact
correctness; this feature is only useful if the kernel of interest has
no data-dependent behavior, or its data-dependent behavior is not based
on data generated by the skipped kernels.

--exit-after-gpu-kernel N
Ends the simulation after completing (non-blit) GPU kernel N.

This commit also renames two existing command-line options:
--debug-at-gpu-kernel -> --debug-at-gpu-task
--exit-at-gpu-kernel  -> --exit-at-gpu-task

These were renamed because they count GPU tasks, which include both
kernels launched by the application as well as blit kernels.

Change-Id: If250b3fd2db05c1222e369e9e3f779c4422074bc

### DIFF
--- a/src/gpu-compute/GPU.py
+++ b/src/gpu-compute/GPU.py
@@ -360,6 +360,10 @@ class GPUCommandProcessor(DmaVirtDevice):
     walker = Param.VegaPagetableWalker(
         VegaPagetableWalker(), "Page table walker"
     )
+    target_non_blit_kernel_id = Param.Int(
+        0,
+        "Skip kernels until reaching this kernel (counting only non-blit kernels)",
+    )
 
 
 class StorageClassType(Enum):
--- a/src/gpu-compute/gpu_command_processor.cc
+++ b/src/gpu-compute/gpu_command_processor.cc
@@ -36,6 +36,7 @@
 #include "arch/amdgpu/vega/pagetable_walker.hh"
 #include "base/chunk_generator.hh"
 #include "debug/GPUCommandProc.hh"
+#include "debug/GPUDisp.hh"
 #include "debug/GPUInitAbi.hh"
 #include "debug/GPUKernelInfo.hh"
 #include "dev/amdgpu/amdgpu_device.hh"
@@ -48,6 +49,7 @@
 #include "sim/full_system.hh"
 #include "sim/process.hh"
 #include "sim/proxy_ptr.hh"
+#include "sim/sim_exit.hh"
 #include "sim/syscall_emul_buf.hh"
 
 namespace gem5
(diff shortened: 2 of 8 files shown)


src/gpu-compute/GPU.py
```diff
@@ -360,6 +360,9 @@ class GPUCommandProcessor(DmaVirtDevice):
     walker = Param.VegaPagetableWalker(
         VegaPagetableWalker(), "Page table walker"
     )
+    target_user_kernel_id = Param.Int(
```
mattsinc: should this be tasks instead of kernels?

mattsinc (commented):
Besides the one comment, a broader question: what does it mean to skip them and maintain correctness?  Wouldn't you need KVM or checkpoints to do so?

src/gpu-compute/GPU.py
mwboyer: The target kernel ID referenced here is the kernel number not counting blit kernels. I have been using the word "task" to refer to both blit and non-blit kernels, and the word "kernel" to refer only to non-blit kernels. There's probably less confusing terminology that we could use here. (Actually, this variable specifically refers to "user kernels", which I was using to further emphasize that this counts only non-blit kernels. If you can think of a better phrase than "user kernel", I'm very open to rewording this to avoid confusion.)

mwboyer:
> Besides the one comment, a broader question: what does it mean to skip them and maintain correctness? Wouldn't you need KVM or checkpoints to do so?

No correctness guarantees are provided when skipping kernels. This feature is only useful for skipping ahead to a kernel that does not have data-dependent behavior. I will reword the commit message and the help text associated with the command-line argument to make this clear.

src/gpu-compute/GPU.py
mattsinc: Probably the only thing to do then is update the comment on the next line to reflect what you said here, thanks for the clarification.

mwboyer:
@mattsinc I have updated the patch to address your comments. I have also replaced "user_kernel" with "non_blit_kernel" in the code for clarity

src/gpu-compute/gpu_command_processor.cc
```diff
@@ -288,10 +326,7 @@ GPUCommandProcessor::dispatchKernelObject(AMDKernelCode *akc, void *raw_pkt,
 
     initABI(task);
     ++dynamic_task_id;
-
-    // The driver expects the start time to be in ns
-    Tick start_ts = curTick() / sim_clock::as_int::ns;
-    dispatchStartTime.insert({disp_pkt->completion_signal, start_ts});
```
mattsinc, line 294: why is this being removed?

mwboyer: It is being moved, not removed, to above the if statement that checks whether a kernel should be skipped. See line 282 in the updated file

mattsinc (approved):

abmerop (approved):

PR #942, opened 2024-03-18
stdlib: Add multiprocessing simulation class

Harshil2107:
This PR is my first stab at adding a mutli processing module to stdlib.

- There are some extra files that are used for testing but will be removed when I undraft the PR

Change-Id: I98b03c52eb769731a04d3d2aab8a79eb8ac91d15


Harshil2107:
I am not to sure about what name the class should be. I have tentatively gone with MultiSim. I will also push a test that runs the example script.

tests/gem5/multiprocessing_tests/README.md
```diff
@@ -0,0 +1,9 @@
+# Learning gem5
```
BobbyRBruce, line 1: This isn't "Learning gem5".

tests/gem5/multiprocessing_tests/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+This set of tests ensure that the multisim example scripts.
```
BobbyRBruce, line 3: This sentence is not complete.

tests/gem5/multiprocessing_tests/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+To run these tests by themselves, you can run the following command in the tests directory:
+
+```bash
+./main.py run gem5/multiprocessing_tests --length=[length]
```
BobbyRBruce, line 8: No need for `--length=[length]` here. They are all "quick" which is `length`'s default value.

src/python/gem5/simulate/multi_sim.py
```diff
@@ -0,0 +1,113 @@
...
+):
+    sim_callables = []
+
+    for config, resource in itertools.product(configs, resources):
+        sim_callables.append(partial(config, resource))
+
+    return sim_callables
+
+
+def run_simulator(
```
BobbyRBruce: Remember to specify private functions like this by adding an underscore: `def _run_simulator(...`.

src/python/gem5/simulate/multi_sim.py
```diff
@@ -0,0 +1,113 @@
...
+    def __init__(self, sim_callables: List[Callable[[], Simulator]]):
+        """
+        :param sim_callables: A list of callables that return a Simulator object
+        """
+        self.sim_callables = sim_callables
+
+    def run_all(self, num_cpus: int):
+        """
+        Run all the simulations in parallel
+        :param num_cpus: The number of simulations to run in parallel at a time
```
BobbyRBruce: I thought we agreed to call this parameter `num_processes` (or `num_proc`). This really has nothing to do with CPUs.

configs/example/gem5_library/multisim/example_callable.py
BobbyRBruce, line 1: Is there a good reason this is segregated into its own Python file? I'll review from now on assuming there is a reason but it seems odd to me.

configs/example/gem5_library/multisim/example_callable.py
```diff
@@ -0,0 +1,59 @@
...
+    # instantiate the riscv matched board with default parameters
+    board = RISCVMatchedBoard(
+        clk_freq="2.2GHz",
+    )
+
+    board.set_workload(resource)
+
+    # run the simulation with the RISCV Matched board
+    simulator = Simulator(board=board, full_system=False)
+    return simulator
```
BobbyRBruce: This is all a bit clunky. I'd suggest returning partial functions. In this example you're returning Simulator _objects_, not a function which can be called. I'd suggest a change like this (haven't actually ran this but I'm pretty sure it's the right way to do things):

```suggestion
from gem5.isas import ISA
from gem5.prebuilt.riscvmatched.riscvmatched_board import RISCVMatchedBoard
from gem5.simulate.simulator import Simulator
from gem5.utils.requires import requires
from typing import List, Callable, Collection
from functools import partial

requires(isa_required=ISA.RISCV)

def get_simulator_partials(workloads: Collection[WorkloadResource]) -> List[Callable[[], Simulator]:
    to_return = []
    for clk_freq in ["2.2GHz", "3.6GHz", "4GHz"]:
        for workload in workloads:
            board = RISCVMatchedBoard(clk_freq=clk_freq)
            board.set_workload(workload)
            to_return.append(partial(Simulator, board))
    return to_return
```

Here we have a function which takes in a collection of workloads. For each workload the a simulation with the RISCV Board is setup 3 times, each with a different clock frequency.

configs/example/gem5_library/multisim/multi-sim-example.py
```diff
@@ -0,0 +1,66 @@
...
+# sim_factory.append(
+#     partial(run_riscvmathed_workload, workload_2)
+# )
+
+sim_factory = get_cross_product_sim_callables(
+    [run_riscvmathed_workload, run_riscvmatched_worklaod_diff_clocks],
+    [workload_1, workload_2],
+)
+
+multi_sim = MultiSim(sim_factory)
```
BobbyRBruce: Keeping in mind the changes I made to 'example_callable.py', I'd change this to:

```suggestion
from .example_callable import get_simulator_partials
multi_sim = MultiSim(get_simulator_partials(workloads=[workload1, workload2]))
```

src/python/gem5/simulate/multi_sim.py
```diff
@@ -0,0 +1,113 @@
...
+    simulations that can be run in parallel is limited to the value of the num_cpus parameter
+    """
+
+    def __init__(self, sim_callables: List[Callable[[], Simulator]]):
+        """
+        :param sim_callables: A list of callables that return a Simulator object
+        """
+        self.sim_callables = sim_callables
+
+    def run_all(self, num_cpus: int):
```
BobbyRBruce: I think `run` is better than `run_all`.

src/python/gem5/simulate/multi_sim.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        Run all the simulations in parallel
```
BobbyRBruce: Be careful with the phrasing here. You are not always running all the simulations in parallel. If `num_cpus` (`num_proc`/`num_processes`) were set to 1 this statement would not be true, for example.

src/python/gem5/simulate/multi_sim.py
```diff
@@ -0,0 +1,113 @@
...
+            processes.append(process)
+
+        while processes:
+            for process in processes:
+                if not process.is_alive():
+                    processes.remove(process)
+            sleep(1)
+
+        print("All simulations have finished")
+
```
BobbyRBruce: You _really_ should've read up on multiprocess `pools` before this. This functionality you provide already exists to handle the scheduling of new processes. The following should be roughly what you need to do:

```py
# Specify the number of processes the pool is allowed to use.
# Note: If `multiprocessing.Pool()` then all available processes are used.
pool = multiprocessing.Pool(processes=num_proc)

# Create a process for each element in `self._sim_callables` passed to function `_run_simulator`. I.e.:
# Process 1 : _run_simulator(self.sim_callables[0])
# Process 2 : _run_simulator(self.sim_callables[1])
# ... and so on.
pool.map(_run_simulator, self.sim_callables) 

# The `pool.map` function will only return when all the processes have finished executing.

def _run_simulator(sim_callable: Callable[[], Simulator]) -> None:
    sim_callable().run()
```

src/python/gem5/simulate/multi_sim.py
```diff
@@ -0,0 +1,113 @@
...
+def get_cross_product_sim_callables(
+    configs: List[Any], resources: List[AbstractResource]
+):
+    sim_callables = []
+
+    for config, resource in itertools.product(configs, resources):
+        sim_callables.append(partial(config, resource))
+
+    return sim_callables
```
BobbyRBruce: I think it's best to leave it up to the user to define their partial functions. I suspect if you try to handle every use case like this we'll end up with far too many functions to properly maintain.

So i'd just remove this helper function.

tests/gem5/multiprocessing_tests/test-multiprocesing.py
```diff
@@ -0,0 +1,61 @@
...
+from testlib import *
+
+if config.bin_path:
+    resource_path = config.bin_path
+else:
+    resource_path = joinpath(absdirpath(__file__), "..", "resources")
+
+all_simulations_completed_verifier = verifier.MatchRegex(
+    re.compile(r"All simulations have finished")
+)
```
BobbyRBruce, line 43: Two comments on this test and the wider change:

1. I'm happy with "exit code == 0" for this.
2. I don't think the `MultiSim` object should be printing this message. I'd assume if the `run` (or `run_all`) function returns without complaint then all the simulations have finished.

BobbyRBruce (changes requested):
I think the big gap in your knowledge if you haven't properly dived into the `Multiprocessing` module beyond the `Process` class. I'd suggest reaching through a tutorial on some of the more "advanced" features. `multiprocessing.pool` is very powerful and solves a lot of the issues you tried to fix here (see my in-line comments).

configs/example/gem5_library/multisim/example_callable.py
BobbyRBruce: Just to clarify I feel that "example_callable.py" and "multi-sim-example.py" could exist in the same file.

tests/gem5/multiprocessing_tests/test-multiprocesing.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+import os
+import re
+
+from testlib import *
+
+if config.bin_path:
+    resource_path = config.bin_path
+else:
+    resource_path = joinpath(absdirpath(__file__), "..", "resources")
```
BobbyRBruce, line 39: You've copy and pasted this from other test scripts but the `resource_path` is never used anywhere here.

tests/gem5/multiprocessing_tests/test-multiprocesing.py
```diff
@@ -0,0 +1,61 @@
...
+    verifiers=(all_simulations_completed_verifier,),
+    config=joinpath(
+        config.base_dir,
+        "configs",
+        "example",
+        "gem5_library",
+        "multisim",
+        "multi-sim-example.py",
+    ),
+    config_args=[],
```
BobbyRBruce, line 57: Thought: Why don't you have the single argument of this config to be the number of threads? It doesn't make much sense for it to be hard-coded in the script.

BobbyRBruce:
I created a little toy example for an approach I think is better. Despite being a simple bare-bones proof-of-concept I hope it's clear how it'd all fit together if built upon: https://github.com/BobbyRBruce/gem5/commit/599e76f65df39ecc156ddfea93fc8e7c9e7acaa1.

You can run the example with `./build/ALL/gem5.opt multisim/config.py`.

src/python/gem5/simulate/multi_sim.py
powerjg: Strong +1 from me

src/python/gem5/simulate/multi_sim.py
```diff
@@ -0,0 +1,113 @@
...
+        :param num_cpus: The number of simulations to run in parallel at a time
+        """
+        processes = []
+
+        for sim_callable in self.sim_callables:
+            if len(processes) > num_cpus:
+                for process in processes:
+                    if not process.is_alive():
+                        processes.remove(process)
+                    os.kill(process.pid, signal.SIGTERM)
```
powerjg: This was just some testing code from me. You should remove this.

src/python/gem5/simulate/multi_sim.py
```diff
@@ -0,0 +1,113 @@
...
+                sleep(1)
+            process_name = (
+                f"{sim_callable.func.__name__}_{sim_callable.args[0].get_id()}"
+            )
+            process = Process(
+                target=run_simulator,
+                args=(sim_callable,),
+                name=process_name,
+            )
+            print(f"Starting process {process_name}")
```
powerjg: We shouldn't have any `print` statements in this code. There's a good argument to need *logging* but for that, we should use something better. E.g., `info()` if we can or `logging` in python. That said, we need to tread lightly here with changes. My only suggestion is to remove the print statement or change it to an `info`

src/python/gem5/simulate/multi_sim.py
```diff
@@ -0,0 +1,113 @@
...
+            process.start()
+            processes.append(process)
+
+        while processes:
+            for process in processes:
+                if not process.is_alive():
+                    processes.remove(process)
+            sleep(1)
+
+        print("All simulations have finished")
```
powerjg: Again, no print statements.

src/python/gem5/simulate/multi_sim.py
powerjg: Yeah, I had a good reason not to use `multiprocessing.Pool`. It *does* work, but it's going to be limiting our features.

The reason is that I am looking forward to a new feature that wouldn't be possible with `Pool`. I want to be able to track the processes ourselves. Specifically, I want to send a special signal periodically and ask the running simulation to tell me what it's current status is. This is related to the better support for exit events.

I can provide more detail later, but I have a plan!

src/python/gem5/simulate/multi_sim.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def get_cross_product_sim_callables(
+    configs: List[Any], resources: List[AbstractResource]
+):
+    sim_callables = []
+
+    for config, resource in itertools.product(configs, resources):
+        sim_callables.append(partial(config, resource))
+
```
powerjg:
```suggestion

    sim_callables = [partial(config, resource) for config, resource in itertools.product(configs, resources)]

```

src/python/gem5/simulate/multi_sim.py
powerjg: I *think* I follow what you're saying Bobby, but I'm not totally certain. Are you suggesting deleting `get_cross_product_sim_callable`?

It could be useful to have some "common" examples here, but I agree that we can't be everything to everyone. I'm also not sure how to use this particular function since it has no documentation.

Harshil, either remove the function or add documentation.

src/python/gem5/simulate/multi_sim.py
powerjg: Yeah, you're also not running *all* in parallel. You're running all of the simulations with the parallelism specified with `num_cpus` (should be `num_processes`)

src/python/gem5/simulate/multi_sim.py
```diff
@@ -0,0 +1,113 @@
+# Copyright (c) 2024 The Regents of the University of California
```
powerjg: I suggest this file should be in `gem5/utils/multiprocessing`

configs/example/gem5_library/multisim/example_callable.py
```diff
@@ -0,0 +1,59 @@
...
+    board = RISCVMatchedBoard()
+
+    board.set_workload(resource)
+
+    # run the simulation with the RISCV Matched board
+    simulator = Simulator(board=board, full_system=False)
+    return simulator
+
+
+def run_riscvmatched_worklaod_diff_clocks(resource) -> Simulator:
```
powerjg: Spelling


```suggestion
def run_riscvmatched_workload_diff_clocks(resource) -> Simulator:
```

configs/example/gem5_library/multisim/example_callable.py
powerjg: If this works +1.

`to_return.append(lambda: Simulator(board))` may work as well if you don't want to use `partial`. I can't decide which I like more.

configs/example/gem5_library/multisim/multi-sim-example.py
```diff
@@ -0,0 +1,66 @@
...
+from gem5.isas import ISA
+from gem5.resources.resource import obtain_resource
+from gem5.simulate.multi_sim import (
+    MultiSim,
+    get_cross_product_sim_callables,
+)
+
+# getting the resources we want to run
+workload_1 = obtain_resource("riscv-gapbs-tc-run")
+workload_2 = obtain_resource("riscv-npb-is-size-s-run")
```
powerjg, line 41: Can you use the example suite?

configs/example/gem5_library/multisim/example_callable.py
powerjg: I *think* Bobby is right here. Can you test if you can put this in a single file?

I think the only thing that needs to be in another file is what's in `multi_sim.py`

powerjg (commented):
It's looking good. A few comments below.

configs/example/gem5_library/multisim/multi-sim-example.py
Harshil2107: I cannot currently as suites dont work with multiprocessing. After the refactor obtain_resource PR is merged, we can change this.

src/python/gem5/simulate/multi_sim.py
Harshil2107: I removed the function and added the code to create cross product in the example script.

configs/example/gem5_library/multisim/example_callable.py
Harshil2107: I tried putting them in same file, it gives a PicklingError saying `import of module '__m5_main__' failed`.

BobbyRBruce:
This is replaced by the work done in #1167

PR #945, opened 2024-03-18
mem-ruby: Copyback UD_RU line when evicted in CHI protocol

jjuninho:
This is a followed up fix to #791 mem-ruby: Fix possible dirty line loss in CHI when ReadShared hit on UD line.
UD_RU line may have stale data since the upstream could have updated the line, so its local cache line data is treated as invalid (dataValid=false). But when the line is evicted, it must be written back to downstream because the upstream may have the line in clean state (UC). This change fixes it by performing copy back the UD_RU line while keeping its dataValid as false.

Example error case:
- L3 was in UD_RSC and being evicted without back-invalidation. LLC (HN) was in RU state.
- Because there's still upstream sharer, L3 sends WriteClean.
- Because the data state was unique and dirty, L3 sends CBWrData_UD_PD.
- LLC becomes UD_RU.
- When the line is evicted from LLC (LocalHN_Eviction), the line is just dropped, causing the loss of the dirty copy

### DIFF
--- a/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
+++ b/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
@@ -401,6 +401,12 @@ action(Initiate_Replacement, desc="") {
   }
   copyCacheAndDir(cache_entry, getDirEntry(address), tbe, initial);
 
+  // UD_RU line needs to be written back because the upstream may be in UC state.
+  if (initial == State:UD_RU) {
+    tbe.dataBlk := cache_entry.DataBlk;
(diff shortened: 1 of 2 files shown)


ivanaamit:
@tiagormk

ivanaamit:
@tiagormk, this is a friendly reminder to look at this PR when you have time. The changes are related to the CHI protocol. Thank you.

tiagormk (approved):
Thanks @jjuninho 
These changes make sense to me.

powerjg (approved):

PR #947, opened 2024-03-19
python,stdlib: Improve exit event return

### LINKED ISSUE #751: Modify gem5 to pass an object upon an exit event instead of a string
I've resurrected this issue from our old Jira service: https://gem5.atlassian.net/browse/GEM5-1128

This task is mostly related in how the gem5 simulator returns an exit event event. At present the only meaningful information communicated is the exit event cause string. In this Simulator package we parse this and classify exit events into broad categories. gem5 should be modified to return an object containing information on how an exit event was triggered, at what tick and instruction, and what kind of exit event was triggered (user interrupt, tick/instruction count reached, CPU switch request, etc).
 
This change will require considerable refactoring of the Simulation package.

> **Commend on this issue by @andysan** : https://gem5.atlassian.net/browse/GEM5-1128?focusedCommentId=13376
>
>My original long-term plan when rewriting the wrappers in PyBind was to to use a mechanism resembling exceptions where each exit is an object derived from GlobalSimLoopExitEvent. This would have a few nice benefits:
>
> * It’d make it possible to add custom data to exits (e.g., a work item ID).
> * It makes it possible to use the exit object as an exception in Python. This could be useful * when handling an unexpected simulator exit in m5.drain (these are currently dropped silently).
> * It makes exit code matching much less error prone since you end up doing isinstance(event, MyExitEvent) checks instead of string matching.
>* Properly exposing global events (not just exit events) to Python would make it possible to build custom global events implemented in Python. We currently only support this for local events. This could, for example, be used to schedule stat dumps without exiting from the simulation loop.
>
>The main reason I didn't export the constructor of GlobalSimLoopExitEvent is that the original SWIG code had issues with a memory leak. I think the main issue was that GlobalSimLoopExitEvent objects are normally owned by the C++ side and not returned as smart pointers.
>
>The first step would be to fix the memory leak. Should be reasonably easy without changing the Python API. I think we'd need the following changes:
>
>* C++: Change/add `GlobalSimLoopExitEvent that allows event creation without scheduling
>* C++: Remove the exitSimLoop helper function and replace it with a makeExitEvent helper that returns a shared pointer.
>* PyBind: Export more GlobalSimLoopExitEvent members for scheduling etc.
>* PyBind: Update the GlobalSimLoopExitEvent bindings to use reference counting
>
>Generalising this further would require Python bindings for GlobalEvent. I think the easiest way to make this work reliably would be to disallow instantiation of GlobalEvent from Python and create a Python-specific base class (we use this trick for normal events, see PyEvent). The unfortunate consequence of this trick is that exit events from C++ would have GlobalSimLoopExitEvent as their base, but that wouldn’t be the case for exit events implemented in Python. It might make sense to create a separate PyGlobalSimLoopExitEvent class that patches up that discrepancy

Modifying exit events to return objects instead of strings was also discussed here: https://gem5-review.googlesource.com/c/public/gem5/+/66231

BobbyRBruce:


Harshil2107:
@powerjg @BobbyRBruce : I got the new hyper call events to work with this last commit. I will clean up the files and push the final version but I tested this by replacing the m5 exit in the disk images to `m5 hypercall` followed by 1,2 and 3 for the three exits we have.

We dont need to define a handler explicitly anymore.

src/python/gem5/simulate/exit_handler.py
```diff
@@ -0,0 +1,207 @@
...
+    @overrides(ExitHandler)
+    def _exit_simulation(self) -> bool:
+        return False
+
+
+class AfterBootExitHandler(ExitHandler):
+    @overrides(ExitHandler)
+    def _process(self, simulator: "Simulator") -> None:
+        print("Second exit: Started `after_boot.sh` script")
+        print("Switching to Timing CPU")
```
powerjg: This line is not generic and should be removed.

src/python/gem5/simulate/exit_handler.py
```diff
@@ -0,0 +1,207 @@
...
+
+    @overrides(ExitHandler)
+    def _exit_simulation(self) -> bool:
+        return True
+
+
+class KernelBootedExitHandler(ExitHandler):
+    @overrides(ExitHandler)
+    def _process(self, simulator: "Simulator") -> None:
+        print("First exit: kernel booted")
```
powerjg: "First" and "Second" (below) should probably be removed. We can't guarantee that there won't be some new event added before this.

Also, we should probably not use "print" but some kind of logging. Maybe this is a change for another patch, though.

src/python/gem5/simulate/simulator.py
```diff
@@ -91,6 +97,32 @@ class Simulator:
...
+            3: AfterBootScriptExitHandler,
+            # The default exit handler for scheduled exit event, such as those
+            # scheduled by the user via `scheduleTickExitAbsolute` or
+            # `scheduleTickExitFromCurrent`.
+            4: ScheduledExitEventHandler,
+        }
+        assert all(
+            i >= 0 for i in default_map.keys()
+        ), "Exit handler mapped to ID <0 in default map."
+        return default_map
```
powerjg: I still think this can automatically be generated (and extended) without having to modify the stdlib code (and rebuild gem5). I can mock this up on Thursday.

powerjg (commented):
A couple of small things from a quick skim of the first few files. Also, *many files are missing copyright*.

src/sim/init_signals.hh
```diff
@@ -40,6 +42,9 @@ void initSignals();
 
 // separate out sigint handler so that we can restore the python one
 void initSigInt();
+void initSigRtmin();
+std::string extractStringFromJSON(std::string& full_str, std::string start_str,
```
powerjg, line 46: Is this really a full "json" parser or more restricted? I think it's more restricted with just {'key':'value'}.

Assuming I'm right, you should change the name to make that more clear. Also, adding some comments in the function to make the restriction clear is important as well.

src/sim/sim_events.cc
```diff
@@ -54,19 +54,38 @@ namespace gem5
 
 GlobalSimLoopExitEvent::GlobalSimLoopExitEvent(Tick when,
                                                const std::string &_cause,
-                                               int c, Tick r)
+                                               int c, Tick r,
+                                               uint64_t hypercall_id,
+        std::map<std::string, std::string> payload)
```
powerjg, line 59: Why not create a new type for hypercalls that doesn't have the extraneous parameters. I don't think we'll ever need the code, repeat, or cause parameters for the hypercall. IMO, we should specialize more and reuse old code less.

In fact, I feel like this could be the other way around where we use the hypercall for the "old" way with the payload being `{'cause': <whatever>}`

src/sim/sim_exit.hh
```diff
@@ -55,6 +55,21 @@ void exitSimLoop(const std::string &message, int exit_code = 0,
 void exitSimLoopNow(const std::string &message, int exit_code = 0,
                     Tick repeat = 0, bool serialize = false);
 
+void exitSimLoopWithHypercall(const std::string &message, int exit_code,
```
powerjg, line 58: As said above, I don't think we need these extra parameters here. But, let me know if I'm mistaken.

tests/gem5/stdlib/configs/hypercall-exit-workload/include/gem5/asm/generic/m5ops.h
powerjg, line 1: These files shouldn't be here. Just use the files in gem5/include

util/hypercall_external_signal/transmitter.py
powerjg, line 1: copyright...

util/m5/src/command/hypercall.cc
powerjg, line 1: copyright

powerjg (commented):
I don't like the default dict because it separates the definition of the class from the hypercall number. I mocked up a way to improve this. I haven't tested it, but I don't know why it wouldn't work. This will also essentially automatically override the exit handler if you specify your own implementation. If you want to use one handler for multiple different exit events, I'm sure we could figure out how to do that. Let me know what you think.

https://github.com/powerjg/gem5/commit/73b3ae63438c32fbe8dc175c337c7b84fd2f7c6c

src/python/gem5/simulate/simulator.py
BobbyRBruce: This can be extended and modified without rebuilding gem5. This is just the default allocation but it can be extended/modified via the configuration script as needed..

BobbyRBruce:
Note on this: As there is an urgency to have _a_ version of this available for work at UC Davis I have merged it prematurely. The comments about things to be improved here are still valid and will be addressed

BobbyRBruce:
This PR has been cherry-picked to the v24.1.1.0 staging branch: https://github.com/gem5/gem5/tree/release-staging-v24-1-1-0

PR #950, opened 2024-03-20
arch-vega: Various vega fixes to enable nanogpt

abmerop:
This PR fixes some issues observed that were needed to get nanogpt working.


src/arch/amdgpu/vega/insts/vop3p.cc
```diff
@@ -675,9 +678,15 @@ Inst_VOP3P__V_PK_FMA_F32::execute(GPUDynInstPtr gpuDynInst)
...
                                        : bits(src2[lane], 31, 0);
 
-            float dword1 = std::fma(*reinterpret_cast<float*>(&s0l),
-                                    *reinterpret_cast<float*>(&s1l),
-                                    *reinterpret_cast<float*>(&s2l));
+            float s0lf = *reinterpret_cast<float*>(&s0l);
+            float s1lf = *reinterpret_cast<float*>(&s1l);
+            float s2lf = *reinterpret_cast<float*>(&s2l);
+
+            if (neg & 1) s0lf = -s0lf;
```
mattsinc, line 685: can you add a comment explaining these neg if's?

src/arch/amdgpu/vega/insts/vop3p.cc
```diff
@@ -686,9 +695,15 @@ Inst_VOP3P__V_PK_FMA_F32::execute(GPUDynInstPtr gpuDynInst)
...
                                           : bits(src2[lane], 31, 0);
 
-            float dword2 = std::fma(*reinterpret_cast<float*>(&s0h),
-                                    *reinterpret_cast<float*>(&s1h),
-                                    *reinterpret_cast<float*>(&s2h));
+            float s0hf = *reinterpret_cast<float*>(&s0h);
+            float s1hf = *reinterpret_cast<float*>(&s1h);
+            float s2hf = *reinterpret_cast<float*>(&s2h);
+
+            if (neg_hi & 1) s0hf = -s0hf;
```
mattsinc, line 702: same here

src/arch/amdgpu/vega/insts/vop3p.cc
```diff
@@ -731,23 +746,36 @@ Inst_VOP3P__V_PK_MUL_F32::execute(GPUDynInstPtr gpuDynInst)
     int opsel = instData.OPSEL;
     int opsel_hi = extData.OPSEL_HI;
 
+    int neg = extData.NEG;
```
mattsinc, line 749: same here

mattsinc:
I also tried checking if GCN3 has the same problems, but XNOR does not appear to be there and it was not clear how to validate the off by 1 decoder error.  Was hoping one of these helped with issues there too.

src/arch/amdgpu/vega/insts/vop3p.cc
abmerop: I could add something somewhere, but the input modifiers like neg/neg_hi are common to all vop3p instructions (also regular vop3) so to reduce clutter in the code maybe it makes more sense at the top of the file?

abmerop:
> I also tried checking if GCN3 has the same problems, but XNOR does not appear to be there and it was not clear how to validate the off by 1 decoder error. Was hoping one of these helped with issues there too.

This is a new instruction starting in Vega20 I believe. See page 3 for new instructions: https://www.amd.com/content/dam/amd/en/documents/radeon-tech-docs/instruction-set-architectures/vega-7nm-shader-instruction-set-architecture.pdf (PDF page 11)

There were 5 instructions affected by the of-by-one fix starting at opcode 668.  These are also new in Vega10, GCN3 only uses up to opcode 658.

src/arch/amdgpu/vega/insts/vop3p.cc
mattsinc: yeah, I'm fine with a separate commit for this then.

mattsinc (approved):

PR #964, opened 2024-03-26
arch-riscv: Fix the RiscvBareMetal parameter reset_vect

rogerchang23424:
The `reset_vect` has exist for a long time and `reset_vect` will not effect if the user gonna to use customized reset_vect. The CL added the `auto_reset_vect` to let the config determine the `reset_vect` from workload entry point or user-specified

Ref: https://gem5-review.googlesource.com/c/public/gem5/+/42053

Change-Id: I928c0dc42aaa85ceabf8d75f9654486496e0ffee

### DIFF
--- a/src/arch/riscv/RiscvFsWorkload.py
+++ b/src/arch/riscv/RiscvFsWorkload.py
@@ -43,6 +43,11 @@ class RiscvBareMetal(Workload):
     bootloader = Param.String("File, that contains the bootloader code")
     bare_metal = Param.Bool(True, "Using Bare Metal Application?")
     reset_vect = Param.Addr(0x0, "Reset vector")
+    auto_reset_vect = Param.Bool(
+        True,
+        "Use bootloader entry point as reset vector. "
+        "If `auto_reset_vect` is true, the `reset_vect` parameter is ignored.",
+    )
 
 
 class RiscvLinux(KernelWorkload):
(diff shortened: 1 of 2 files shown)


src/arch/riscv/bare_metal/fs_workload.hh
```diff
@@ -45,9 +45,9 @@ class BareMetal : public Workload
     // checker for bare metal application
     bool _isBareMetal;
     // entry point for simulation
-    Addr _resetVect;
     loader::ObjectFile *bootloader;
     loader::SymbolTable bootloaderSymtab;
+    Addr _resetVect;
```
powerjg: Is this change necessary? I guess it doesn't matter...

src/arch/riscv/RiscvFsWorkload.py
```diff
@@ -43,6 +43,9 @@ class RiscvBareMetal(Workload):
     bootloader = Param.String("File, that contains the bootloader code")
     bare_metal = Param.Bool(True, "Using Bare Metal Application?")
     reset_vect = Param.Addr(0x0, "Reset vector")
+    auto_reset_vect = Param.Bool(
+        True, "Use bootloader entry point as reset vector"
```
powerjg: Could you add here "If this is true, the `reset_vect` parameter is ignored"

powerjg (commented):
Two nits

hnpl (approved):

src/arch/riscv/RiscvFsWorkload.py
rogerchang23424: Sure

hnpl (approved):

powerjg (approved):

PR #966, opened 2024-03-26
mem-cache, configs, arch-arm: Handle partitioning policies through a PartitionManager

giactra:
This PR is offloading some of the partitioning logic to the partitioning manager, effectively changing
the partitioning interface. Rather than always relying on the PartitionFieldExtention data structure to
convey partition IDs, we make it implementation defined by introducing the partitioning manager abstraction.
We want user to be able to extract the partitionId more flexibly and this requires using a SimObject.

Users can extend the PartitioningManager, overriding the readPacketPartitionId, therefore providing their
own mean of injecting/extracting partitioning data from a packet


src/dev/arm/mpam.hh
```diff
@@ -0,0 +1,62 @@
...
+
+#ifndef __DEV_ARM_MPAM_HH__
+#define __DEV_ARM_MPAM_HH__
+
+#include "mem/cache/tags/partitioning_policies/partition_manager.hh"
+
+namespace gem5
+{
+
+class MpamMSC : public partitioning_policy::PartitionManager
```
odanrc: Could you write down as a comment in the file what MpamMSC stands for? I had to google it to find out

odanrc (approved):
I would probably have implemented it as a PartitionPolicy::Multi (see multi compressor and multi prefetcher), since it only forwards the calls to its sub-partition policies, but I am ok with this approach too.

src/dev/arm/mpam.hh
giactra: Ack

giactra:
> I would probably have implemented it as a PartitionPolicy::Multi (see multi compressor and multi prefetcher), since it only forwards the calls to its sub-partition policies, but I am ok with this approach too.

Yeah, so at the moment it could look like a simple dispatcher but you could actually have more
complex logic within it. For example in the MPAM MSC controller you can dynamically adjust partitions by programming it so it is effectively a partition controller… Also nothing stops you from having a single partitioning policy and I believe the “multi” starts losing a bit of sense (but I have not looked into the prefetcher code to be fair)

src/dev/arm/mpam.hh
odanrc: Also, shouldnt the class be named MPAMMSC since this is how it is referred to in the docs?

giactra: I would prefer as it is for readability if you don't mind

odanrc: I think we shouldn´t break the naming conventions, other than major exceptions (e.g., the name is the identity of the product - for example, let´s say a company created a product called eXplorer - then the class name should probably follow this name). I don´t think MpamMSC is the case for a major exception, but just a lack of break down into namespaces and classes.

I was reading a bit on MPAM. This [url](https://developer.arm.com/documentation/107768/0100/How-does-MPAM-work-/MPAM-information-bundle), in particular, seems to describe the following namespace:
```
namespace MPAM {
class Requester;
class InformationBundle;
class Partition;
class PartitionMonitor; // or Manager
class PMG;
class MSC; // So this is a PartitionMonitor, right?
}
```

Do you think it would make sense to create this namespace and remove this _Mpam_ prefix?

src/dev/arm/Mpam.py
```diff
@@ -0,0 +1,43 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from m5.objects.PartitioningPolicies import PartitionManager
+
+
+class MpamMSC(PartitionManager):
+    type = "MpamMSC"
+    abstract = True
+    cxx_header = "dev/arm/mpam.hh"
```
odanrc, line 42: Should the file be renamed, since it only contains MSC code, or do you plan to put all MPAM related stuff there?

odanrc (approved):

src/dev/arm/Mpam.py
giactra: Thanks @odanrc; Yessss, I plan to push more MPAM stuff in this file in the near future (currently under internal review)

BobbyRBruce (approved):

PR #967, opened 2024-03-26
stdlib: Add tree structure to the AbstractCacheHierarchy

giactra:
One of things we miss in gem5 is the capability to neatly compose the
cache hierarchy of CPUs and clusters of CPUs.  The BaseCPU
addPrivateSplitL1Caches and addTwoLevelCacheHierarchy APIs have
historically been used to bind cache levels together.

These APIs have been superseded by the introduction of the Cache
hierarchy abstraction in the standard library. The standard library
makes it cleaner for a user to quickly instantiate a hierarchy of caches
with few lines of code.  While this removes a lot of complexity for a
user, the Hierarchy objects still have little information about their
internal topology.

To address this problem, this patch adds a tree data structure to the
AbstractCacheHierarchy class, where every node of the tree represent
a cache in the hierarchy. In this way we will expose APIs for traversing
and querying the tree.

For example a 2 CPUs system with private L1, private L2 and shared L3
will contain the following tree:

         [root]
           |
          [L3]
           /\
          /  \
        [L2] [L2]
         |    |
        [L1] [L1]

### DIFF
--- a/src/python/gem5/components/cachehierarchies/abstract_cache_hierarchy.py
+++ b/src/python/gem5/components/cachehierarchies/abstract_cache_hierarchy.py
@@ -1,3 +1,15 @@
+# Copyright (c) 2024 Arm Limited
+# All rights reserved.
+#
+# The license below extends only to copyright in the software and shall
+# not be construed as granting a license to any other intellectual
+# property including but not limited to intellectual property relating
+# to a hardware implementation of the functionality of the software
+# licensed hereunder.  You may use the software subject to the license
+# terms below provided that you ensure that this notice is replicated
+# unmodified and in its entirety in all distributions of the software,
+# modified or unmodified, in source code or in binary form.
+#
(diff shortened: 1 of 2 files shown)


src/python/gem5/components/cachehierarchies/abstract_cache_hierarchy.py
```diff
@@ -28,17 +40,72 @@
...
 
 from ..boards.abstract_board import AbstractBoard
 
 
+class CacheNode:
+    def __init__(
+        self,
+        name: str,
+        cache: SimObject,
+        next_level: int,
```
BobbyRBruce: Shouldn't this be : 
```suggestion
        next_level: "CacheNode",
```

Down in the `add_child`, `self` is passed.

BobbyRBruce (approved):
Great changes! Thanks @giactra

PR #973, opened 2024-03-27
stdlib,tests: Add StatTester SimObject and Scalar tests

BobbyRBruce:
This SimObject can be used to quickly test the statistics are functioning correctly. The SimObject schedules a single event which sets the statistics to values dependent on the SimObject params.

With this commit the "Scalar" stats have a StatTester subclass that can be used for testing. More can be added as required.

Tests are included to check our Scalar SimStat functionality.


src/test_objects/Kconfig
```diff
@@ -0,0 +1,29 @@
...
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+config USE_TEST_OBJECTS
+    bool "Object for testing"
```
powerjg: I think it would be a good idea to expand this help information. Think about it as if you're someone looking at the menuconfig. Will you know if you should or shouldn't enable this option?

src/test_objects/stat_tester.hh
```diff
@@ -0,0 +1,119 @@
...
+ *     from m5.objects import (
+ *         Root,
+ *         ScalarStatTester,
+ *     )
+ *
+ *     stat_tester = ScalarStatTester()
+ *
+ *     stat_tester.value = 42
+ *     stat_tester.name = "example"
+ *     root = Root(full_system=False, system=stat_tester)
```
powerjg:
```suggestion
 *     root = Root(full_system=False, tester=stat_tester)
```

There's no need to call it "system"

In fact, this code may be more clear:

```suggestion
 *     root = Root(full_system=False)
 *     root.stat_tester = stat_tester
```

src/test_objects/stat_tester.hh
```diff
@@ -0,0 +1,119 @@
...
+ *
+ *     ---------- Begin Simulation Statistics ----------
+ *     simSeconds                               18446744.073710
+ *     simTicks                                 18446744073709551616
+ *     finalTick                                18446744073709551616
+ *     simFreq                                  1000000000000
+ *     hostSeconds                                      0.00
+ *     hostTickRate                             111124964299455114248192
+ *     hostMemory                                  404481920
+ *     system.example                                     42
```
powerjg: If you take my advice above, this will change.

src/test_objects/stat_tester.hh
```diff
@@ -0,0 +1,119 @@
...
+
+  protected:
+    void setStats() override;
+    struct ScalarStatTesterStats : public statistics::Group
+    {
+        ScalarStatTesterStats(
+            statistics::Group *parent,
+            const ScalarStatTesterParams &params
+        );
+        statistics::Scalar scalar;
```
powerjg, line 113: Not in this change, but in the future, does it make sense to test all of the stat types here?

BobbyRBruce: I'm not sure precisely what you mean but you can see what it looks like when I expand this in this patch where I add tests for Vector stats (yet to be incorporated PR, waiting on this to be merged): https://github.com/BobbyRBruce/gem5/commit/dff74a9950c2eae056b4ee77760c062952133cd4

src/test_objects/Kconfig
BobbyRBruce: I expanded this. Let me know if it's more suitable.

src/test_objects/stat_tester.hh
powerjg: Sorry for the confusion. I understand what you did in that commit (I guess I missed the PR). That's a reasonable approach. 

I would have instead added a line here that adds a Vector stat, but your way works fine too!

powerjg (approved):

PR #976, opened 2024-03-28
dev: Remove duplicate virtio files

giactra:
Remove the following files:
* src/dev/virtio/rng 2.cc
* src/dev/virtio/rng 2.hh

Which were a copy of rng.hh and rng.cc. Probably added to the repository by accident. They were not compiled by scons

Change-Id: I9d1da19cc243c513ab7af887b1b6260d8e361b57


andysan (approved):
Seems like these got added in a release-related merge.

PR #725, opened 2023-12-29
arm,stdlib: added kvm support to the ARM board

kaustav-goswami:
This change adds support to use KVM cores on the ARM board. The board simulates gic to enable KVM, similar to the gem5 ARM FS configs. The limitation is that it only supports VExpress_GEM5_V1.

PR #977, opened 2024-03-28
util-m5: Add default M5OP_ADDR to arm64

hnpl:
As pointed out here [1], the expected M5OP_ADDR for arm64 arch is 0x10010000. This change reflects that.

[1] https://github.com/gem5/gem5/pull/725

Change-Id: I7e72f5ea20d4aacf3115a485ba7cd664d33d037e


BobbyRBruce (dismissed):

powerjg (commented):
Is this going to break `semi` or the instruction-based ops?

hnpl:
I think it won't break instruction-based m5 ops as x86 does the same thing. I'm not sure about semi, though it doesn't look like semi uses m5_mem.

https://github.com/gem5/gem5/blob/9207458fd771e402cc98f67806dc83f5a0f79a0c/util/m5/src/abi/x86/SConsopts#L31

giactra (approved):

PR #990, opened 2024-04-03
util: Fixed gem5img.py script

elatalhm:
Made the script more robust to different names.

### DIFF
--- a/util/gem5img.py
+++ b/util/gem5img.py
@@ -194,7 +194,8 @@ def findPartOffset(devFile, fileName, partition):
         r"\s*:\s*"  # Separator
         r"start=\s*(?P<start>\d+),\s*"  # Partition start record
         r"size=\s*(?P<size>\d+),\s*"  # Partition size record
-        r"type=(?P<type>\d+)"  # Partition type record
+        r"type=(?P<type>.+?),*?"  # Partition type record
+        r".*"  # anything else, e.g., name field
         r"\s*$"  # End of line
     )
     lines = out.splitlines()


util/gem5img.py
```diff
@@ -194,7 +194,8 @@ def findPartOffset(devFile, fileName, partition):
         r"\s*:\s*"  # Separator
         r"start=\s*(?P<start>\d+),\s*"  # Partition start record
         r"size=\s*(?P<size>\d+),\s*"  # Partition size record
-        r"type=(?P<type>\d+)"  # Partition type record
+        r"type=(?P<type>.+?),*?"  # Partition type record
+        r".*"  # anything else, e.g., name field
```
giactra, line 198: The name field should be already covered (see https://github.com/gem5/gem5/pull/990/commits/bfbab75346561ec9479925ea43fb9578a0929fb5#diff-4cde24986c38a36e85628238a3e49f0732987ef0c7ec70c942ef179f5af6c578R193) unless you are telling me the sfdisk output sometimes rearranges its position.

Just to clarify, I am not oppsing the regex per se, I am wondering whether the comment is a bit misleading. For example I can see by playing a bit with sfdisk that I have an uuid entry so maybe that should be used as an example?

util/gem5img.py
```diff
@@ -194,7 +194,8 @@ def findPartOffset(devFile, fileName, partition):
         r"\s*:\s*"  # Separator
         r"start=\s*(?P<start>\d+),\s*"  # Partition start record
         r"size=\s*(?P<size>\d+),\s*"  # Partition size record
-        r"type=(?P<type>\d+)"  # Partition type record
+        r"type=(?P<type>.+?),*?"  # Partition type record
```
giactra, line 197: Could you specify on the PR/commit what's the problem with

`(?P<type>\d+)`

I presume you encountered some non digit types; I think it would be great if these example were documented in the PR

giactra (commented):
Hi @elatalhm first of all thanks for this contribution. I have two comments (see below) which  are mainly requesting to better document in the PR what is the problem and the solution you are providing

ivanaamit:
Hi @elatalhm,

```
f=.git/hooks/commit-msg
mkdir -p $f
chmod +x $f
```

ivanaamit:
@elatalhm, could you please address Giacomo's comments and incorporate the requested changes? Thank you.

util/gem5img.py
samueltphd: This was in response to an issue I was having with the x86 disk image from the gem5 resource page. I was getting the following output from calling ``sudo python3 util/gem5img.py mount x86-ubuntu.py mnt''

%> /usr/sbin/losetup /dev/loop20 x86-ubuntu.img                                                                                                                                                            
No partition description was found in sfdisk output:                                                                                                                                                        
  label: dos                                                                                                                                                                                                
  label-id: 0xeb7b110b
  device: /dev/loop20
  unit: sectors
  sector-size: 512
   
  /dev/loop20p1 : start=        2048, size=     4190208, type=83, bootable
Could not determine size of first partition.

elatalhm: The type field can produce hexadecimal numbers. This can be verified just by running `sfdisk -T` to list the supported types.

util/gem5img.py
elatalhm: I get this output on the x86-ubuntu.img file from gem5 resources:

```
$ sfdisk -d x86-ubuntu.img
label: dos
label-id: 0xd0c30ca2
device: x86-ubuntu.img
unit: sectors

x86-ubuntu.img1 : start=        2048, size=     4190208, type=83, bootable
```

The problem is caused by the `, bootable` at the end. The end of the regex has a `\s*$` so it expects the line to end after the type field.

BobbyRBruce (approved):

Issue #882, opened 2024-02-20
CHI (Arm) near/far atomic operation causes deadlock


yichungchen84:
GEM5 23.1.0.0:
CHI near/far operation causes deadlocks when CPU is configured with L1 private, L2 private, and L3 shared on a RNF, and another shared memory on HNF. (FS mode, Arm)

System configuration:
  CPU: 4 ArmO3CPUs share a RNF (L1 private, L2 private, and L3 shared ), total 2 RNF. 2 HNFs with distributed SLC.
  RAM: DDR4
  FS OS: ubuntu 20.04
(The configuration has no deadlock with 23.0.1.0.)

Current solution:
By removing near/far atomic operation feature in 23.1.0.0, FS mode runs with no deadlock.

Further study:
When the system configured without shared L3, it has no deadlock. Therefore, an initial guess is that near atomic operation might be the issue.

andysan:
Ping @tiagormk.

vsoriap:
Hi @yichungchen84, could you attach the config.ini and config.json you use.

Note that the far AMO feature is intended to be used only with three levels of cache. However, near AMO should work as the old version 23.0.0.

yichungchen84:
Hi @vsoriap 

Sorry that due to the IT restriction, I cannot share the config,ini/json files. I will see if I can share the configuration files (CHI_config.py, etc).

My configuration has private L1, private L2, shared L3, and SLC. I try to set all cache controllers "alloc_on_atomic = false". I assume it should turn off AMO feature (am I correct?), but it still has deadlocks with "alloc_on_atomic = false".

Would you mind to let me know where I can look into source code of far AMO feature/policy, so I can configure it to support my cache hierarchy? Thanks.

vsoriap:
Okay, I see that you have 4 levels of cache. We never tested that configuration, and you may encounter some problems. How L3 and SLC have the flags is_HN configured?

`alloc_on_atomic` is not a flag that turns off AMO. AMOs are instructions supported in the RISC-V and ARM architecture. You cannot switch them off. What you can do is remove them from your binary.

What `alloc_on_atomic` really encodes is if the HNF needs to allocate a cache line or not if an Atomic Transaction arrives at the HNF and the cache line is not present. If true, the HNF requests the cache line to the SNF, performs the operation, and stores it in the local cache associated with the HNF (in your case, I assume it's the SLC). If false, the HNF requests the cache line to the SNF, performs the atomic operation, and then writes the updated value back to SNF.

[...]

PR #994, opened 2024-04-04
mem-ruby: Implement no_alloc Far Atomics in CHI

vsoriap:
This PR introduces a missing pice of far atomic implementation.  This pull request incorporates several changes:

- Enable 2-level and 4-level (and N-level) cache hierarchies, removing Atomic_NoWait transactions
- Fix Unique Near policy implementation that raised abort
- Add support for alloc_on_atomic == False. Enables Far Atomics on systems where the HNF does not allocate evicted lines at LLC (Like in WriteUpdate).

### DIFF
--- a/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
+++ b/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
@@ -1058,13 +1058,29 @@ action(Initiate_AtomicReturn_LocalWrite, desc="") {
   tbe.actions.push(Event:SendDBIDResp_AR);
   tbe.actions.pushNB(Event:WriteFEPipe);
   tbe.actions.push(Event:SendCompData_AR);
-  tbe.actions.push(Event:WriteFEPipe);
-  tbe.actions.push(Event:CheckCacheFill);
   tbe.actions.push(Event:DelayAtomic);
+  tbe.actions.push(Event:CheckCacheFill);
   tbe.actions.push(Event:WriteBEPipe);
   tbe.actions.push(Event:TagArrayWrite);
 }
 
+action(Initiate_AtomicReturn_WriteBack, desc="") {
+  if ((tbe.dir_sharers.count() > 0) && tbe.dataMaybeDirtyUpstream) {
+    tbe.actions.push(Event:SendSnpUnique);
+  } else if (tbe.dir_sharers.count() > 0){
+    // no one will send us data unless we explicitly ask
+    tbe.actions.push(Event:SendSnpUniqueRetToSrc);
+  }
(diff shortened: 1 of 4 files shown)


vsoriap:
This PR should fix the problems experienced by @yichungchen84 in the issue #882

ivanaamit:
@tiagormk, can you please review this PR? It fixes the issue #882. Thanks.

yichungchen84:
@vsoriap Thanks, I will have a quick check over the weekend.

fschaetzle:
@vsoriap Thanks for the solution. I run two different configurations:
1. core clusters with priv. L1/L2, shared L3 and HN without SLC.
2. cores with priv. L1/L2 and HN incl. SLC.

By adding RUSC state to AtomicReturn and AtomicNoReturn transition the first config run successfully.

However, the second config run into assert(tbe.expected_req_resp.hasExpected()) from Receive_RespSepDataFromCompData action. Looks like the assertion is thrown by L2. After receiving DBIDResp it also receives CompData_I from HN before sending NCBWrData. Maybe a race condition?

vsoriap:
Hi, @fschaetzle. 

Could you describe a little bit more the second config. I would like to know how many CPUs and the cache controller flags you set in that config (e.g., alloc_on_atomic, policy_type, allow_SD, etc)

fschaetzle:
@vsoriap I tested small configurations with 4 CPUs, allow_SD=true, policy_type=1 and only HN allocates on atomic.
I think [this](https://github.com/fschaetzle/gem5/commit/7bf08d73d5b1995f888bbd40ef7123b697e0e986) is a possible solution for the issues I encountered (at least for small systems and I will verify bigger systems over the weekend).
It adds:
- RUSC states in AtomicReturn and AtomicNoReturn transition.
- CheckCacheFill action
- HN (alloc_on_atomic=true) waits for NCBWrData before sending CompData. Otherwise we run into the assertion mentioned above.

vsoriap:
You are right, @fschaetzle !! I accidentally removed the CheckCacheFill and forgot to remove the NoBlock clause in the CompData. I've updated the latest change to add those changes and added you as a co-author.

I've also changed how the WriteFE/BE latency is defined to only have one of each per transaction, as in WriteUpdate.

fschaetzle:
@vsoriap thank you. From my side everything is fine for the PR.

ivanaamit:
@tiagormk, could you please review this PR? The changes are related to the CHI protocol. Thank you.

tiagormk (approved):
LGTM! Thanks @vsoriap for the fixes and @fschaetzle for testing.

ivanaamit (approved):

PR #996, opened 2024-04-05
stdlib: Improve gem5 PyStats

BobbyRBruce:
This PR incorporates numerous improvements and fixes to the gem5 PyStats. This includes:

* PyStats now support SimObject Vectors. The PyStats representing them are subscribable and therefore acceptable by accessing an index: e.g.,: `simobjectvec[0]`. (This replaces the `Vector` group PyStat)
* Adds the `SparseHist` PyStats.
* Adds the `Vector2d` to PyStats.
* The `Distribution` PyStats is fixed to be a vector of Scalars.
* Tests added for the PyStat's Vector and bugs fixed.


BobbyRBruce:
I believe I've fixed the failing tests. This should be ready for review.

src/python/m5/ext/pystats/group.py
```diff
@@ -68,17 +65,60 @@ def __init__(
...
+    def __iter__(self):
+        return iter(self.value)
+
+    def __len__(self):
+        return len(self.value)
+
+    def get_all_stats_of_name(self, name: str) -> List[AbstractStat]:
+        """
+        Get all the stats in the vector of that name. Useful for performing
+        operations on all the stats of the same name in a vector.
```
powerjg: Could you add an example of this? Reading the code and looking at the description, I don't know how I would use this function.

src/python/m5/stats/gem5stats.py
```diff
@@ -244,38 +374,35 @@ def get_simstat(
...
+        time_converstion = (
+            None  # TODO https://gem5.atlassian.net/browse/GEM5-846
+        )
+        final_tick = Root.getInstance().resolveStat("finalTick").value
+        sim_ticks = Root.getInstance().resolveStat("simTicks").value
+        simulated_begin_time = int(final_tick - sim_ticks)
+        simulated_end_time = int(final_tick)
+
+        if prepare_stats:
+            _m5.stats.processDumpQueue()
```
powerjg: Why do this for every object in the list of inputs and not just once?

powerjg (commented):
Two small things, but overall it looks great! Thanks for the hard work!!

src/python/m5/ext/pystats/group.py
BobbyRBruce: (jotting this down while it's fresh in my mind --- I will add a test later to verify it works like I think).

I can't really remember, but i think the reason for this function is for ases you have a simobject tree which makes stats look like this:


```py
processor = {
  cpus = [
      { stats1 = { stat2 = 5.6 } },
      { stats_a = { stat_a = { stat2 = 4.3 } } },
      { stat2 = 3.6 }
  ]
}
```

With your output to stats.txt looking something like:

```
processor.cpu0.stats1.stat2 = 5.2
processor.cpu1.stats_a.stats1.stat2 = 4.3
processor.cpu2.stat2 = 3.6
```

The argument someone had (I think) was the JSON output is worse if you just want to get all the `stat2` stats from cores in a the `processor` simobject. For these big tree like structures, there was a desire to have a way to just iterate through all the child nodes recursively and get all the stats with the name a specific name. Therefore, this function was created and would return `{stat2 =5.2, stat2=4.3,  stat2 = 3.6}` if you run something like `processor.get_all_stats_of_name("stat2")`.

I think the typicall usecase here would be you've got a multicore system and you want to get an average of all the stats called `X` across all cores.

This function also works for the case where you want _just_ the `stats1.stat2` stats. So, in our example, just `{stats1.stat2 = 5.2, stats1.stat2=4.3}`. TL;DR: it's a way of getting all the stats that share a particular name basically.

I don't know if that's in anyway useful but I'm pretty sure that's why I added this.

src/python/m5/stats/gem5stats.py
BobbyRBruce: Good point. I've moved this code so it's only done once in commit 8f0ed4606184bb8de2647063508a40c9276f19ef.

src/python/m5/stats/gem5stats.py
```diff
@@ -243,39 +373,37 @@ def get_simstat(
...
-        elif isinstance(r, SimObject):
-            if prepare_stats:
+        creation_time = datetime.now()
+        time_converstion = (
+            None  # TODO https://gem5.atlassian.net/browse/GEM5-846
+        )
+        final_tick = Root.getInstance().resolveStat("finalTick").value
+        sim_ticks = Root.getInstance().resolveStat("simTicks").value
+        simulated_begin_time = int(final_tick - sim_ticks)
+        simulated_end_time = int(final_tick)
```
powerjg: I think all of this code can go outside the for loop as well.

src/python/m5/ext/pystats/group.py
powerjg: Got it! I think it makes sense to include. Can you add a code snippet example to the docstring?

powerjg (changes requested):

src/python/m5/ext/pystats/group.py
BobbyRBruce: I did write a test but I realized this code is incredibly broken. I tried for a while to fix it but couldn't quite figure it out. I've therefore opted to remove it (if this is ok with you). The "find" feature does a similar thing: a recursive search through the tree, it just doesn't handle the old text-based stats names (e.g., find("gem5") will search for a stat called "gem5" and return those it finds, but 'root.gem5' won't return the gem5 stat in the 'root' object).

The PR has this function removed now.

powerjg: Sounds good to me.

src/python/m5/ext/pystats/__init__.py
```diff
@@ -24,22 +24,11 @@
...
-    "JsonLoader",
-]
+from .abstract_stat import *
+from .group import *
+from .jsonloader import *
+from .serializable_stat import *
+from .simstat import *
+from .statistic import *
+from .storagetype import *
+from .timeconversion import *
```
powerjg: I really don't like this. From [PEP008](https://peps.python.org/pep-0008/#imports)

> Wildcard imports (from <module> import *) should be avoided, as they make it unclear which names are present in the namespace, confusing both readers and many automated tools.

BobbyRBruce:
@powerjg : I think this is ready to go. Feel free to review it.

powerjg (approved):
I have to admit, I have totally lost context for this in the past couple of weeks. I think the only thing that changed was the `import *` and that's all I had asked for? If so, I guess this is good based on my prior reviews.

PR #1009, opened 2024-04-09
dev: Fix interrupt logic in uart8250

bgasiorzewski:
Hi, we've noticed some issues with the Uart8250 device when using it as the Linux console. Sometimes the Uart interrupt would remain constantly posted, so Linux would continue to try and handle it, effectively resulting in an infinite loop. With this patch, I'm no longer seeing any issues, but my testing has been limited to configurations and workloads we're interested in at Imagination, so please let me know if there's some other tests I should run or if you notice any other issues.

This patch fixes several issues with interrupt posting and clearing in the uart8250 device.

The "status" member variable and the console interrupt should be kept in sync. However, in one code path in readIir, the interrupt bit was being cleared in the status variable but not in the platform controller.

Additionally, in some code paths, the interrupts would be cleared in the status variable and in the interrupt controller, but a future interrupt would remain scheduled, causing a spurious interrupt and setting a bit in status to 1.

These issues can confuse the kernel and result in an ininite interrupt handling loop.

Another issue is related to the fact that there are two interrupt causes (TX and RX) and both of them can be valid at the same time. When one of them becomes no longer valid, we should check the status of the other one before clearing the interrupt.

This patch addresses the issues listed above and refactors the interrupt clearing logic to reduce repetition.

### DIFF
--- a/src/dev/serial/uart8250.cc
+++ b/src/dev/serial/uart8250.cc
@@ -57,7 +57,6 @@ Uart8250::processIntrEvent(int intrBit)
     } else {
        DPRINTF(Uart, "UART InterEvent, not interrupting\n");
     }
-
 }
 
 /* The linux serial driver (8250.c about line 1182) loops reading from
@@ -78,12 +77,26 @@ Uart8250::scheduleIntr(Event *event)
     static const Tick interval = 225 * sim_clock::as_int::ns;
     DPRINTF(Uart, "Scheduling IER interrupt for %s, at cycle %lld\n",
             event->name(), curTick() + interval);
-    if (!event->scheduled())
+    if (!event->scheduled()) {
         schedule(event, curTick() + interval);
-    else
+    } else {
         reschedule(event, curTick() + interval);
+    }
 }
 
+void
+Uart8250::clearIntr(int intrBit)
+{
+    if ((status & intrBit) == 0) {
+        return;
+    }
+    status &= ~intrBit;
+
+    // Clear interrupt at the controller if neither TX nor RX is pending now
+    if (!status) {
+        platform->clearConsoleInt();
+    }
+}
 
 Uart8250::Uart8250(const Params &p)
     : Uart(p, p.pio_size), registers(this, name() + ".registers"),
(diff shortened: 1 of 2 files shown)


src/dev/serial/uart8250.cc
```diff
@@ -137,11 +147,14 @@ Uart8250::readRbr(Register8 &reg)
...
-    status &= ~RX_INT;
-    platform->clearConsoleInt();
+    clearIntr(RX_INT);
 
     if (device->dataAvailable() && registers.ier.get().rdi)
         scheduleIntr(&rxIntrEvent);
+    else if (rxIntrEvent.scheduled())
+        // Must deschedule a future interrupt if there is no more data to read,
+        // otherwise this would confuse the kernel (iir.id=rx but lsr.rdr=0)
+        deschedule(rxIntrEvent);
```
powerjg: Please use curly braces around multiline if statements. There have been too many bugs that were caused by trying to optimize for 2 characters :)

```suggestion
    else if (rxIntrEvent.scheduled()) {
        // Must deschedule a future interrupt if there is no more data to read,
        // otherwise this would confuse the kernel (iir.id=rx but lsr.rdr=0)
        deschedule(rxIntrEvent);
    }
```

powerjg (commented):
lgtm. One small style thing below that I'd like to see changed (you can amend the commit and force push). As a small nit, it would be better to use curly braces everywhere, but only required where I mentioned below. To give you context on this, it's something that I push for after spending ~days tracking down a bug where someone meant to add a statement inside the `if`, but there were no curly braces.

We'll (@Harshil2107) will run some tests to make sure this works with RISC-V, Arm, and x86 (I think we use this device with all of those systems...)

bgasiorzewski:
> lgtm. One small style thing below that I'd like to see changed (you can amend the commit and force push). As a small nit, it would be better to use curly braces everywhere, but only required where I mentioned below. To give you context on this, it's something that I push for after spending ~days tracking down a bug where someone meant to add a statement inside the `if`, but there were no curly braces.

Thanks for the quick review. I went ahead and added braces everywhere :)

> We'll (@Harshil2107) will run some tests to make sure this works with RISC-V, Arm, and x86 (I think we use this device with all of those systems...)

Great, thanks, let me know if there's any issues.

Harshil2107:
I ran the changes in this PR with RISCV and X86. The terminal connect work, the console shows and I can type to the console.

ivanaamit (approved):

PR #1011, opened 2024-04-10
 mem,gpu-compute: Implement GPU TCC directed invalidate

abmerop:
The GPU device currently supports large BAR which means that the driver
can write directly to GPU memory over the PCI bus without using SDMA or
PM4 packets. The gem5 PCI interface only provides an atomic interface
for BAR reads/writes, which means the values cannot go through timing
mode Ruby caches. This causes bugs as the TCC cache is allowed to keep
clean data between kernels for performance reasons. If there is a BAR
write directly to memory bypassing the cache, the value in the cache is
stale and must be invalidated.

In this commit a TCC invalidate is generated for all writes over PCI
that go directly to GPU memory. This will also invalidate TCP along the
way if necessary. This currently relies on the driver synchonization
which only allows BAR writes in between kernels. Therefore, the cache
should only be in I or V state.

To handle a race condition between invalidates and launching the next
kernel, the invalidates return a response and the GPU command processor
will wait for all TCC invalidates to be complete before launching the
next kernel.

This fixes issues with stale data in nanoGPT and possibly PENNANT.

### DIFF
--- a/src/gpu-compute/compute_unit.cc
+++ b/src/gpu-compute/compute_unit.cc
@@ -1701,16 +1701,20 @@ ComputeUnit::DataPort::processMemReqEvent(PacketPtr pkt)
     } else if (!(sendTimingReq(pkt))) {
         retries.push_back(std::make_pair(pkt, gpuDynInst));
 
-        DPRINTF(GPUPort,
-                "CU%d: WF[%d][%d]: index %d, addr %#x data req failed!\n",
-                compute_unit->cu_id, gpuDynInst->simdId, gpuDynInst->wfSlotId,
-                id, pkt->req->getPaddr());
+        if (gpuDynInst) {
+            DPRINTF(GPUPort,
+                    "CU%d: WF[%d][%d]: index %d, addr %#x data req failed!\n",
+                    compute_unit->cu_id, gpuDynInst->simdId,
+                    gpuDynInst->wfSlotId, id, pkt->req->getPaddr());
+        }
     } else {
-        DPRINTF(GPUPort,
-                "CU%d: WF[%d][%d]: gpuDynInst: %d, index %d, addr %#x data "
-                "req sent!\n", compute_unit->cu_id, gpuDynInst->simdId,
-                gpuDynInst->wfSlotId, gpuDynInst->seqNum(), id,
-                pkt->req->getPaddr());
+        if (gpuDynInst) {
+            DPRINTF(GPUPort,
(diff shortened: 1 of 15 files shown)


abmerop:
@v-ramadas  should review too but not able to add as reviewer.

src/gpu-compute/compute_unit.cc
```diff
@@ -1701,16 +1737,20 @@ ComputeUnit::DataPort::processMemReqEvent(PacketPtr pkt)
     } else if (!(sendTimingReq(pkt))) {
         retries.push_back(std::make_pair(pkt, gpuDynInst));
 
-        DPRINTF(GPUPort,
-                "CU%d: WF[%d][%d]: index %d, addr %#x data req failed!\n",
-                compute_unit->cu_id, gpuDynInst->simdId, gpuDynInst->wfSlotId,
-                id, pkt->req->getPaddr());
+        if (gpuDynInst) {
```
mattsinc, line 1740: Do we want else's here that print/panic if there is no gpuDynInst?

src/mem/ruby/system/VIPERCoalescer.cc
```diff
@@ -106,6 +107,10 @@ VIPERCoalescer::makeRequest(PacketPtr pkt)
         invTCP();
     }
 
+    if (pkt->req->isInvL2()) {
+        invTCC(pkt);
+    }
+
```
mattsinc, line 113: probably should remove extra line here

mattsinc (approved):

v-ramadas:
Looks good to me

src/mem/ruby/system/VIPERCoalescer.cc
abmerop: I was keeping the previous format of having an empty line before return statement (unless github is showing you something I am not seeing)

src/gpu-compute/compute_unit.cc
abmerop: The invalidates generated by the device won't have any associated instruction, so panic wouldn't work here. For printing there is not much information to add here as most are related to the instruction.  I don't want to assume no instruction means L2 invalidate as there may be other non-instruction memory requests in the future.  There are DPRINTFs added under the GPUCoalescer debug flag for when invalidates are sent due to BAR writes that should make up for the missing print here

mattsinc: ok ... I am worried about tracking this down then, but we can let it through and see.

Issue #798, opened 2024-01-23
Drop support for Ubuntu 20.04


ivanaamit:
We plan to drop support for Ubuntu 20.04 in the upcoming v24.0.

pranith:
Shouldn't there be a period of deprecation before we drop support?

andysan:
[Ubuntu 20.04 LTS](https://ubuntu.com/about/release-cycle) gets standard security updates until April 2025 and extended support until 2029(!). I think we have phased out 20.04 by now, but I wouldn't be surprised if it still used by a large-ish minority lot of gem5 users still use it. We clearly need to phase out support at some point though. However, unless there is a really compelling reasons, I would suggest that we wait at least a few months after the standard support expires in April before dropping it.

powerjg:
Yeah, the problem is that we support *too many compilers and libraries*. We officially support at least [14 major versions](https://github.com/gem5/gem5/blob/stable/.github/workflows/compiler-tests.yaml#L19). This is a maintenance overhead.

If we distributed gem5 as a binary, and we only had to compile gem5 for Ubuntu 20.04 once, it would be no problem at all to continue support. The difficulty is how often the compiler tests break and the maintenance overhead of making it possible for everyone to compile gem5 on all systems. #555 is a good example of the painfulness of this process. We've also recently had to add annoying special cases in the code for specific versions of compilers (e.g., #696 which was caused by #636 which was caused by #555)

Basically, there's only so much time in the day to work on development of gem5. The question is: should we spend the time ensuring that 20.04 continues to work for another year, or should we spend that time on something else. Given the 53 open issues at this moment, I believe there are higher priorities than upstream gem5 continuing to support 20.04.

Now, dropping support for 20.04 doesn't mean that gem5 will stop working overnight on machine with 20.04. It just means that we may stop doing tests on 20.04 and end up requiring a compiler or library version that is not default on 20.04. Given the prevalence of docker now, it seems pretty straightforward to use a new ubuntu distribution on an older distribution via docker containers.

It's a tradeoff with no perfect solution.

andysan:
I can totally see how it would simplify lots of things if we could just ditch older platforms more quickly. From a purely selfish perspective, I wouldn't mind dropping support for 20.04. I don't think any of my colleagues relies on that version of Ubuntu. I'm more worried about breaking other peoples' flows.

[...]

PR #1020, opened 2024-04-14
tests,docker: Drop support for GCC <v11 and Clang <v14

BobbyRBruce:
`apt install gcc` in Ubuntu 22.04 gives version 11, and `apt install clang` gives version 14. If we are to remove support for Ubuntu 20.04 (#798), there is little justification in maintaining support for compiler versions before then.


powerjg (commented):
The idea seems fine to me. I didn't look in detail at the PR, though

BobbyRBruce:
This is superceded by #1292

PR #1022, opened 2024-04-14
github: Update 'ubuntu-22.04' to 'ubuntu-latest'

BobbyRBruce:
There was some inconsistency in the GitHub Workflow files on using 'ubuntu-latest' (which gets the latest Ubuntu version) or 'ubuntu-22.04'. To keep things consistent 'ubuntu-latest' is now used in all cases. This also saves us updating workloads upon release of a new Ubuntu version.

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -14,7 +14,7 @@ concurrency:
 jobs:
     pre-commit:
     # runs on github hosted runner
-        runs-on: ubuntu-22.04
+        runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


ivanaamit (approved):

powerjg:
I would have pushed back on this change if I had seen it in time.

The problem with doing this is that all of our tests could start failing because canonical updates something. We don't want to be dependent on something like that, IMO. Instead, I think we should be adding *more* specificity to the images we use.

I would prefer this is reverted. We should use specific versions everywhere (even if this means updating things once every 2 years)

PR #1023, opened 2024-04-14
arch-vega: Fix output warnings, gem5.fast

abmerop:
Fix gem5.fast build not building when using gpu model.

Removes very spammy stat distribution bucket size prints when running gpu model.

### DIFF
--- a/src/arch/amdgpu/vega/operand.hh
+++ b/src/arch/amdgpu/vega/operand.hh
@@ -517,12 +517,23 @@ namespace VegaISA
             switch(_opIdx) {
               case REG_EXEC_LO:
                 {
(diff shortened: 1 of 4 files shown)


mattsinc (approved):
I'm not sure what the operand.hh changes have to do with gem5.fast, but this seems fine.

abmerop:
> I'm not sure what the operand.hh changes have to do with gem5.fast, but this seems fine.

The template class is going to instantiate the readSpecialVal function for all operand types.  Some of the special values are 64-bit, so reading into a 32-bit register (NumDwords == 1) doesn't make much sense.  This was causing a warning with memcpy trying to copy a 64-bit value into a 32-bit variable and warnings are of course errors in gem5.  Why this only happened for gem5.fast I don't know.

PR #1025, opened 2024-04-14
util-docker: Bump gpu-fs build docker to ROCm 6.0.2

abmerop:
This bumps the docker image used to build GPU applications for input to GPUFS simulations from ROCm 5.4.2 to ROCm 6.0.2 and Ubuntu from 20.04 to 22.04. This matches the versions in gem5-resources#29 .

Several notes were added to the Dockerfile to describe where the RUN commands come from. A README.md is also added to clarify that this is not a disk image for GPUFS and is only used to build applications.

Change-Id: I9ada99e2ed1854cb7adb76f2a1fa662bab398f86


mattsinc (approved):
From my perspective this is good.

PR #1031, opened 2024-04-15
tests: Add the 'tests-se-across-os-version' tests

BobbyRBruce:
These tests ensure gem5's SE-mode is able to interpret a simple suite of Syscalls across different versions of the same OS.


BobbyRBruce:
[cross-os-version-bins.zip](https://github.com/gem5/gem5/files/14977996/cross-os-version-bins.zip)


WIP commit. The code uploaded thus far would be dependent on these resources being uploaded to gem5 Resources. They were compiled from the `matrix-multiply` and `print-this` resources but compiled on 22.04 and 24.04 Ubuntu hosts specifically for the purposes of this test.

BobbyRBruce:
Closing: I don't think these tests are of much use.

Issue #1033, opened 2024-04-15
"fatal: ... pio is already connected to interrupt_out_port" occurring when multiple interrupts present


BobbyRBruce:
**Describe the bug**
The Daily Tests are failing, at least in part, due to PR #1007 the initializing all interrupt objects, one-per-thread. connecting these multiple interrupts to PIO ports does not work correctly and fails with errors such as:

```shell
fatal: Port <orphan System>.cpu0.interrupts.pio is already connected to <orphan System>.caches.sequencers0.interrupt_out_port[0], cannot connect <orphan System>.caches.sequencers0.interrupt_out_port[2]
```

From my understanding the issue arises from this code: https://github.com/gem5/gem5/blob/630f3822b86093f869b3411a0b1d5d70358d32f2/src/cpu/BaseCPU.py#L178, which worked fine for single interrupts but fails with multiple

**Affects version**
The current develop branch. Error introduced with PR #1007.

**To Reproduce**
Run `./build/ALL_MSI/gem5.opt configs/learning_gem5/part3/simple_ruby.py`

To create ALL_MSI I added "build_opts/ALL_MSI" with following contents:
 
```
RUBY=y                                                                         
BUILD_ISA=y                                                                    
USE_ARM_ISA=y                                                                  
USE_MIPS_ISA=y                                                                 
USE_POWER_ISA=y                                                                
USE_RISCV_ISA=y                                                                
USE_SPARC_ISA=y                                                                
USE_X86_ISA=y                                                                  
USE_TEST_OBJECTS=y                                                             
RUBY_PROTOCOL_MSI=y
```

**Terminal Output**
```shell
fatal: Port <orphan System>.cpu0.interrupts.pio is already connected to <orphan System>.caches.sequencers0.interrupt_out_port[0], cannot connect <orphan System>.caches.sequencers0.interrupt_out_port[2]
```

**Host Operating System**
Ubuntu 22.04

**Host ISA**
X86

**Additional information**
Add any other information which does not fit in the previous sections but may be of use in fixing this bug.

BobbyRBruce:
PR #1034 reverts #1007. This is just a stop-gap measure until a better solution can be found (if merged, #1034 can be later unreverted).

@nmosier : Do you have any quick fix for this?

PR #1034, opened 2024-04-15
cpu-o3, arch-x86: Revert "initialize interrupts for all SMT ..."

BobbyRBruce:
This reverts commit bc39283451c6b563f8db6660ce694e838a3b38b0.

Stops #1033 (but not a true fix for the issue).

This revert can be unreverted when a solution to #1033 is found.


powerjg (changes requested):
Should be closed in favor of #1035

Issue #1004, opened 2024-04-08
sim-se, cpu-o3, arch-x86: SMT fails under SE due to disconnected interrupt request port


nmosier:
**Describe the bug**
When trying to run more than one workload under syscall emulation with SMT enabled on the O3 CPU, gem5 panics with the following error:
```
...
src/arch/x86/interrupts.cc:330: panic: panic condition !intRequestPort.isConnected() occurred: Int port not connected to anything!
Memory Usage: 649572 KBytes
Program aborted at tick 0
--- BEGIN LIBC BACKTRACE ---
...
```

**Affects version**
develop @ [c650712](https://github.com/gem5/gem5/commit/c65071282d007b3bde1778ef126af444b33c23ac)

**gem5 Modifications**
None

**To Reproduce**
1. Compile gem5: `scons build/X86/gem5.opt`
2. `./build/X86/gem5.opt configs/deprecated/example/se.py --cpu-type=DerivO3CPU --caches --smt --cmd='./tests/test-progs/hello/bin/x86/linux/hello;./tests/test-progs/hello/bin/x86/linux/hello'`

**Terminal Output**
```
nmosier@cafe-cet:~/gem5-develop/gem5$ ./build/X86/gem5.opt configs/deprecated/example/se.py --cpu-type=DerivO3CPU --caches --smt -c './hello;./hello'                                                                                       
gem5 Simulator System.  https://www.gem5.org
gem5 is copyrighted software; use the --copyright option for details.

gem5 version DEVELOP-FOR-24.0
gem5 compiled Apr  5 2024 14:57:31
gem5 started Apr  8 2024 01:52:24
gem5 executing on cafe-cet, pid 138658
command line: ./build/X86/gem5.opt configs/deprecated/example/se.py --cpu-type=DerivO3CPU --caches --smt -c './hello;./hello'

warn: The se.py script is deprecated. It will be removed in future releases of  gem5.
Global frequency set at 1000000000000 ticks per second
warn: No dot file generated. Please install pydot to generate the dot file and pdf.
src/mem/dram_interface.cc:690: warn: DRAM device capacity (8192 Mbytes) does not match the address range assigned (512 Mbytes)
src/base/statistics.hh:279: warn: One of the stats is a legacy stat. Legacy stat is a stat that does not belong to any statistics::Group. Legacy stat is deprecated.
system.remote_gdb: Listening for connections on port 7000
[...]

PR #1035, opened 2024-04-16
cpu: Fix Ruby/x86 pio port connections

### LINKED ISSUE #1033: "fatal: ... pio is already connected to interrupt_out_port" occurring when multiple interrupts present

powerjg:
Fixes #1033

In the BaseCPU object _uncached_interrupt_response_ports is a class variable, not an instance variable. #1004 changed the explicit self._uncached_interrupt_response_ports to use extend. This caused the list of ports to be extended *for all cores*, which caused problems when using a system with more than 1 core.

This reverts the `extend` part of the change, but keeps the rest.

Change-Id: I6dc7d6da6763048d82960229d34933a3a2ac36e0

### DIFF
--- a/src/cpu/BaseCPU.py
+++ b/src/cpu/BaseCPU.py
@@ -302,20 +302,25 @@ class BaseCPU(ClockedObject):
 
         # Practically speaking, these ports will exist on the x86 interrupt
         # controller class.
+        # Note: _uncached_interrupt_response_ports is a *class* variable
+        # not a *instance* variable. Thus, we cannot use `extend` here.
         if "pio" in self.ArchInterrupts._ports:
-            self._uncached_interrupt_response_ports.extend(
-                [f"interrupts[{i}].pio" for i in range(self.numThreads)]
+            self._uncached_interrupt_response_ports = (
+                self._uncached_interrupt_response_ports
+                + [f"interrupts[{i}].pio" for i in range(self.numThreads)]
             )
         if "int_responder" in self.ArchInterrupts._ports:
-            self._uncached_interrupt_response_ports.extend(
-                [
+            self._uncached_interrupt_response_ports = (
+                self._uncached_interrupt_response_ports
+                + [
                     f"interrupts[{i}].int_responder"
                     for i in range(self.numThreads)
                 ]
             )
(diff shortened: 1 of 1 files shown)


powerjg:
Just FYI @nmosier. This took a bit to track down.

BobbyRBruce (approved):

abmerop (approved):
Fix works for me

nmosier:
Thanks for fixing this! Sorry I was MIA (submission deadline…) and for breaking it in the first place.

Issue #958, opened 2024-03-23
AVX KVM Xsave issue


abmerop:
**Describe the bug**
The AVX / YMM register state is not saved or restored in gem5 with the X86KvmCPU leading to crashes on checkpoint restoration when AVX is enabled in CPUID.

**Affects version**
develop @ 0c684f2d331e47570f47e980307977284666582e

**gem5 Modifications**
No modification

**To Reproduce**
This is easiest to reproduce using a full system GPU configuration as it enables AVX by default and supports checkpoint/restore.  This requires the VEGA_X86 build. The application doesn't really matter here, so for the application one can simply use a blank shell script.

1. scons build/VEGA_X86/gem5.opt -j`nproc`
2. touch hello.sh
3. build/VEGA_X86/gem5.opt util/obtain-resource.py x86-gpu-fs-img -p x86-gpu-fs-img
4. build/VEGA_X86/gem5.opt util/obtain-resource.py x86-linux-kernel-5.4.0-105-generic -p vmlinux-5.4.0-105-generic
5. build/VEGA_X86/gem5.opt configs/example/gpufs/vega10_kvm.py --disk-image ./x86-gpu-fs-img --kernel ./vmlinux-5.4.0-105-generic  --gpu-mmio-trace gem5-resources/src/gpu-fs/vega_mmio.log --app hello.sh --checkpoint-dir hello_cpt
6. build/VEGA_X86/gem5.opt configs/example/gpufs/vega10_kvm.py --disk-image ./x86-gpu-fs-img --kernel ./vmlinux-5.4.0-105-generic  --gpu-mmio-trace gem5-resources/src/gpu-fs/vega_mmio.log --app hello.sh --restore-dir hello_cpt

Note:  There is currently a workaround command line option that disables AVX to avoid this issue.  Adding the `--disable-avx` command line option should *not* see this error.

**Terminal Output**

```
[    8.736510] ------------[ cut here ]------------
[    8.736510] Bad FPU state detected at switch_fpu_return+0x7d/0x120, reinitializing FPU registers.
[    8.736510] WARNING: CPU: 0 PID: 461 at /build/linux-hwe-5.4-utjlqf/linux-hwe-5.4-5.4.0/arch/x86/mm/extable.c:114 ex_handler_fprestore+0x65/0x70
[...]

PR #1040, opened 2024-04-19
arch-x86: Add XCR0 register and add to X86KvmCPU

abmerop:
The extended control registers were not being updated in the KVM thread context nor updated in the KVM state. This was causing issues when checkpointing since the XCR0 value was reverting to the default value rather than what it was previously before the checkpoint. THis was causing multiple applications to crash due to executing instructions which are now illegal instructions due to XCR0 being incorrect.

This commit adds the XCR0 as a misc register similar to the exiting x86 control registers and adds all of the helper functions to access and set the register value. It also adds support for updating the KVM CPU's state with the register value and updating the thread context's misc reg value so that it is checkpointed along with the other misc regs.

Note that this does *not* add support for XSAVE of the AVX state (i.e., the upper 128 bits of YMM registers). It does however fix the immediate problem in issue #958 .

Change-Id: I97456c8b57cbc7b381bd4be94944ce6567a43c76

### DIFF
--- a/src/arch/x86/fs_workload.cc
+++ b/src/arch/x86/fs_workload.cc
@@ -302,6 +302,11 @@ FsWorkload::initState()
     // Point to the page tables.
     tc->setMiscReg(misc_reg::Cr3, PageMapLevel4);
 
+    // Only used if cr4.osxsave is set
+    XCR0 xcr0 = tc->readMiscRegNoEffect(misc_reg::Xcr0);
+    xcr0.x87 = 1; // Must be 1 according to x86 specification
+    tc->setMiscReg(misc_reg::Xcr0, xcr0);
+
     Efer efer = tc->readMiscRegNoEffect(misc_reg::Efer);
     // Enable long mode.
     efer.lme = 1;
--- a/src/arch/x86/regs/misc.hh
+++ b/src/arch/x86/regs/misc.hh
@@ -405,6 +405,9 @@ enum : RegIndex
     // "Fake" MSRs for internally implemented devices
     PciConfigAddress,
 
+    XcrBase,
+    Xcr0 = XcrBase,
+
     NumRegs
 };
 
@@ -422,6 +425,13 @@ cr(int index)
 {
     assert(index >= 0 && index < NumCRegs);
     return CrBase + index;
+}
+
+static inline RegIndex
+xcr(int index)
+{
+    assert(index >= 0 && index < NumXCRegs);
+    return XcrBase + index;
 }
 
 static inline RegIndex
@@ -648,6 +658,24 @@ EndBitUnion(CR4)
 BitUnion64(CR8)
     Bitfield<3, 0> tpr; // Task Priority Register
 EndBitUnion(CR8)
+
+BitUnion64(XCR0)
+    Bitfield<0> x87; // x87 FPU/MMX support (must be 1)
+    Bitfield<1> sse; // XSAVE support for MXCSR and XMM registers
+    Bitfield<2> avx; // AVX enabled and XSAVE support for upper halves of YMM
+                     // registers
+    Bitfield<3> bndreg; // MPX enabled and XSAVE support for BND0-BND3
+                        // registers
+    Bitfield<4> bndsrc; // MPX enabled and XSAVE support for BNDCFGU and
+                        // BNDSTATUS registers
+    Bitfield<5> opmask; // AVX-512 enabled and XSAVE support for opmask
+                        // registers k0-k7
+    Bitfield<6> zmm_hi256; // AVX-512 enabled and XSAVE support for upper
+                           // halves of lower ZMM registers
+    Bitfield<7> hi16_zmm; // AVX-512 enabled and XSAVE support for upper ZMM
+                          // registers
+    Bitfield<9> pkru; // XSAVE support for PKRU register
+EndBitUnion(XCR0)
 
 BitUnion64(DR6)
     Bitfield<0> b0;
--- a/configs/example/gem5_library/checkpoints/simpoints-se-restore.py
+++ b/configs/example/gem5_library/checkpoints/simpoints-se-restore.py
@@ -125,7 +125,7 @@ board.set_se_simpoint_workload(
         weight_list=[0.1, 0.2, 0.4, 0.3],
         warmup_interval=1000000,
     ),
-    checkpoint=obtain_resource("simpoints-se-checkpoints-v24-0"),
+    checkpoint=obtain_resource("simpoints-se-checkpoints"),
 )
 
 
--- a/src/arch/x86/isa.cc
+++ b/src/arch/x86/isa.cc
@@ -338,6 +338,8 @@ ISA::setMiscReg(RegIndex idx, RegVal val)
         break;
       case misc_reg::Cr8:
         break;
+      case misc_reg::Xcr0:
+        break;
       case misc_reg::Rflags:
         {
             RFLAGS rflags = val;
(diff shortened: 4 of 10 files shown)


abmerop:
@v-ramadas I stumbled upon a solution to this kernel panic while working on something slightly related, so created a PR for it.  We should be able to test checkpointing GPU ML apps now

v-ramadas:
Thanks! This is great. I will test it out when I get some time.

src/arch/x86/fs_workload.cc
```diff
@@ -302,6 +302,10 @@ FsWorkload::initState()
     // Point to the page tables.
     tc->setMiscReg(misc_reg::Cr3, PageMapLevel4);
 
+    XCR0 xcr0 = tc->readMiscRegNoEffect(misc_reg::Xcr0);
+    xcr0.x87 = 1; // Must be 1
```
mattsinc: Ideally it would be good to explain why it must be 1

src/arch/x86/regs/misc.hh
```diff
@@ -405,6 +405,9 @@ enum : RegIndex
     // "Fake" MSRs for internally implemented devices
     PciConfigAddress,
 
+    XcrBase,
```
mattsinc, line 408: it's a little weird that you declare XcrBase, then set it to the Xcr0 in the next line, but I guess if it works ok

mattsinc (changes requested):
Amazing!  Just one small thing,

src/arch/x86/regs/misc.hh
abmerop: Just following the style above with the other groups of registers.  There just happens to only be one XCR register

src/arch/x86/fs_workload.cc
abmerop: Added comments

abmerop:
@BobbyRBruce @Harshil2107  The failing tests are checkpoint restore checks.  The checkpoint appears to be part of a resource, how does that get updated?  This change implicitly adds another element to an array param

mattsinc (dismissed):

abmerop:
.... Do I need to write a checkpoint upgrader?

Harshil2107:
If the checkpoint upgrader needs to be updated, then I can run the updated checkpoint upgrader on the existing checkpoints and push a new version. If the checkpoint upgrader doesnt need to be update I can just retake the checkpoints and update the version.

Can you tell me if you need to update the checkpoint upgrader?

abmerop:
Well if you take new checkpoint *with this change applied* then that would also fix the failing tests.

I suppose writing an upgrader is more correct as other might need to use it.  Is the upgrader run as part of the tests?

abmerop:
> If the checkpoint upgrader needs to be updated, then I can run the updated checkpoint upgrader on the existing checkpoints and push a new version. If the checkpoint upgrader doesnt need to be update I can just retake the checkpoints and update the version.
> 
> Can you tell me if you need to update the checkpoint upgrader?

Added a checkpoint upgrader.  I guess this would have to be applied to the resources?

Harshil2107:
I will run and update the failing checkpoints.

abmerop:
> I will run and update the failing checkpoints.

Thank you!

Harshil2107:
I just reran the tests with updated checkpoints.

Harshil2107:
To get the tests to pass you would need to update the resource version to 3.0.0 in the x86-fs-restore-checkpoint.py, x86-hello-restore-checkpoint.py. I can also push a commit with this changes if you want me to.

abmerop:
> To get the tests to pass you would need to update the resource version to 3.0.0 in the x86-fs-restore-checkpoint.py, x86-hello-restore-checkpoint.py. I can also push a commit with this changes if you want me to.

Sure... I don't know where those files are.

Harshil2107:
The checkpoint upgrader doesnt seem to fix the checkpoints, should I take new checkpoints or do we need to update the updater again?

abmerop:
It looks like the testlib tests have a different name for the ini config section, so I will have to rewrite the upgrader

abmerop:
Updated the checkpoint upgrader.  You might need to manually edit the m5.cpt file to remove the x86-add-xcr0 tag from the file before running the upgrader again.  Or simply use the 2.0.0 files

abmerop:
If I could make a suggestion.  Is it possible to have testlib run the checkpoint upgrader before running a test? Then you could "officially" recreate during the release freeze? That way it wouldn't require someone with access to the checkpoint files to post a PR requiring checkpoint upgrades.  Although that might be rare enough it doesn't make sense.

Harshil2107:
The updated simpoint is causing other PRs to fail, I will create create a version specific resource for this PR and so that the tests in other PRs dont fail.

Harshil2107:
> If I could make a suggestion. Is it possible to have testlib run the checkpoint upgrader before running a test? Then you could "officially" recreate during the release freeze? That way it wouldn't require someone with access to the checkpoint files to post a PR requiring checkpoint upgrades. Although that might be rare enough it doesn't make sense.

The reason for not doing that is, that the checkpoints on gem5 resources become outdated and need to be updated all at once. As we have to keep the checkpoints in sync with gem5 resources, I feel it is better to keep on updating them as the changes are made. 

I will push a change that defines the resource versions explicitly instead of grabbing the latest version in another PR. In this way the tests for other PRs with older code will not fail, when we update the checkpoints.

Harshil2107:
I have created a new resource for this change, now the tests in other PRs should not be affected. I will sync the resources when I push the PR I mentioned above (this should not affect tests for other PRs).

abmerop:
> > If I could make a suggestion. Is it possible to have testlib run the checkpoint upgrader before running a test? Then you could "officially" recreate during the release freeze? That way it wouldn't require someone with access to the checkpoint files to post a PR requiring checkpoint upgrades. Although that might be rare enough it doesn't make sense.
> 
> The reason for not doing that is, that the checkpoints on gem5 resources become outdated and need to be updated all at once. As we have to keep the checkpoints in sync with gem5 resources, I feel it is better to keep on updating them as the changes are made.
> 
> I will push a change that defines the resource versions explicitly instead of grabbing the latest version in another PR. In this way the tests for other PRs with older code will not fail, when we update the checkpoints.

I see. I was attempting to find a solution that didn't require a single entity (UC Davis) to have to be involved.

powerjg:
> I see. I was attempting to find a solution that didn't require a single entity (UC Davis) to have to be involved.

Thanks! I agree that would be great. The main problem here is that checkpoints are deeply tied to the code of gem5 (which is a problem). Really, checkpoints should be more of an API that can't be changed willy-nilly, but that's not how it's architected right now.

There's a tradeoff in the work to do per-commit and per-release. When possible, I'd like to do the work per-commit to make release time easier. Updating checkpoints should be rare enough (a few times every release) that it's better to do it per-commit than waiting until the release.

You're not the first to suggest auto-running the checkpoint upgrader (or removing checkpointing from testing). But, one of the main arguments against this is that it would mask when checkpoints need to be updated for the next release (putting more onus on us at release time).

mattsinc:
@abmerop it doesn't look like you need me for anything here at this point, since you are still working with @Harshil2107 on the checkpoint changes.  Let me know if I'm missing something though.  I added Harshil as a review to hopefully help with this.

abmerop:
> @abmerop it doesn't look like you need me for anything here at this point, since you are still working with @Harshil2107 on the checkpoint changes. Let me know if I'm missing something though. I added Harshil as a review to hopefully help with this.

The checks has passed now, so I just need a reviewer to approve

mattsinc (approved):
Probably @Harshil2107 or @powerjg should approve too?

Harshil2107 (approved):
lgtm

PR #1045, opened 2024-04-20
mem-ruby: Fix functional reads for MESI Three-Level messages

### LINKED ISSUE #1044: mem-ruby: Ruby functional read failures with MESI Three-Level caches
**Describe the bug**
When running the O3 processor under various configurations with MESI Three-Level Ruby caches, I've encountered frequent Ruby functional read errors, causing a panic [here](https://github.com/gem5/gem5/blob/e578f83739e54ed0fdbb7a68dfce688a4080760e/src/mem/ruby/system/RubyPort.cc#L431).

**Affects version**
develop @ [e578f83](https://github.com/gem5/gem5/commit/e578f83739e54ed0fdbb7a68dfce688a4080760e)

**gem5 Modifications**
None

**To Reproduce**
One way to easily reproduce it, at least on my system, is to run two copies of a simple program on the O3 processor as SMT threads. It appears that the ruby function read is failing somewhere in libc startup code.
1. Create a build options file for X86 MESI Three-Level: `printf 'RUBY=y\nNUMBER_OF_BITS_PER_SET=128\nRUBY_PROTOCOL_MESI_THREE_LEVEL=y\nBUILD_ISA=y\nUSE_X86_ISA=y\n' > build_opts/X86_MESI_Three_Level`
1. Compile gem5: `scons build/X86_MESI_Three_Level/gem5.opt`
2. Compile a simple program: `echo 'int main() {}' > test.c && cc test.c -o test`
3. Run two copies of the program on O3 under SE with SMT: `./build/X86_MESI_Three_Level/gem5.opt configs/deprecated/example/se.py --cpu-type=X86O3CPU --smt --ruby -c './test;./test'`

**Terminal Output**
```
$ ./build/X86_MESI_Three_Level/gem5.opt configs/deprecated/example/se.py --cpu-type=X86O3CPU --ruby --smt -c './test;./test' --interp-dir=/usr

gem5 version DEVELOP-FOR-24.0
gem5 compiled Apr 19 2024 23:18:59
gem5 started Apr 20 2024 15:06:02
gem5 executing on cafe-cet, pid 1286658
command line: ./build/X86_MESI_Three_Level/gem5.opt configs/deprecated/example/se.py --cpu-type=X86O3CPU --ruby --smt -c './test;./test' --interp-dir=/usr

**** REAL SIMULATION ****
src/mem/ruby/system/Sequencer.cc:680: warn: Replacement policy updates recently became the responsibility of SLICC state machines. Make sure to setMRU() near callbacks in .sm files!
src/sim/syscall_emul.cc:74: warn: ignoring syscall mprotect(...)
src/sim/syscall_emul.cc:74: warn: ignoring syscall mprotect(...)
src/sim/syscall_emul.cc:74: warn: ignoring syscall mprotect(...)
src/sim/syscall_emul.cc:74: warn: ignoring syscall mprotect(...)
src/sim/syscall_emul.cc:74: warn: ignoring syscall mprotect(...)
src/sim/syscall_emul.cc:74: warn: ignoring syscall mprotect(...)
src/sim/syscall_emul.cc:74: warn: ignoring syscall mprotect(...)
src/sim/syscall_emul.cc:74: warn: ignoring syscall mprotect(...)
src/mem/ruby/system/RubyPort.cc:431: fatal: Ruby functional read failed for address 0x40a10
Memory Usage: 791320 KBytes
```

**Expected behavior**
The two copies of `test` should exit cleanly.

**Host Operating System**
Ubuntu 22.04

**Host ISA**
X86

**Compiler used**
GCC 11.4.0

**Additional information**
I will submit a patch shortly to fix this

nmosier:
Fix #1044. This patch adds checks for message types (PUTX_COPY, DATA, DATA_EXCLUSIVE) that contain data blocks but were missing from the original `functionalRead` method in MESI Three-Level messages.

Change-Id: I0cedc314166c9cc037bf20f5b7fef5552dd1253c

### DIFF
--- a/src/mem/ruby/protocol/MESI_Three_Level-msg.sm
+++ b/src/mem/ruby/protocol/MESI_Three_Level-msg.sm
@@ -81,8 +81,9 @@ structure(CoherenceMsg, desc="...", interface="Message") {
   PrefetchBit Prefetch,         desc="Is this a prefetch request";
 
   bool functionalRead(Packet *pkt) {
-    // Only PUTX messages contains the data block
-    if (Class == CoherenceClass:PUTX) {
+    // Valid data block is only present in message with following types
+    if (Class == CoherenceClass:PUTX ||
+        Class == CoherenceClass:DATA_EXCLUSIVE) {
         return testAndRead(addr, DataBlk, pkt);
     }
 


ivanaamit (dismissed):

src/mem/ruby/protocol/MESI_Three_Level-msg.sm
```diff
@@ -81,8 +81,11 @@ structure(CoherenceMsg, desc="...", interface="Message") {
   PrefetchBit Prefetch,         desc="Is this a prefetch request";
 
   bool functionalRead(Packet *pkt) {
-    // Only PUTX messages contains the data block
-    if (Class == CoherenceClass:PUTX) {
+    // Valid data block is only present in message with following types
+    if (Class == CoherenceClass:PUTX ||
+        Class == CoherenceClass:PUTX_COPY ||
+	Class == CoherenceClass:DATA ||
```
powerjg: Is there a guarantee that this data isn't stale?

nmosier: I don't have a thorough understanding of this MESI Three-Level protocol, so I'm not sure.

I added checks for `CoherenceClass:DATA` and `CoherenceClass:DATA_EXCLUSIVE` because they're similarly checked in MESI_Two_Level-msg.sm's [ResponseMsg::functionalRead()](https://github.com/gem5/gem5/blob/97a05304527f0d1112ff7aee20e22a9de11760f9/src/mem/ruby/protocol/MESI_Two_Level-msg.sm#L101), and I assumed the situation here would be analogous (although MESI_Three_Level doesn't have a separate class for requests or responses, if I'm not mistaken).

The only difference in what message types are considered from MESI Two-Level to MESI Three-Level is `CoherenceClass:PUTX_COPY`. I added that because I assumed it's similar to `CoherenceClass:PUTX`.

The message type that appears to cause the bug in #1044 is DATA_EXCLUSIVE in particular. I could revise the patch to only additionally check for that message type, rather than the other two (PUTX_COPY and DATA).

powerjg: Even if the response message's type of `DATA` implies that this data is the most up to date version in the system, I wouldn't immediately assume that it's true for the coherence message (L1->L2 message). For instance, there could be another cache that is in the process of upgrading to "M" state and already completed its store. I don't know if this implementation of MESI allows the store to complete before receiving all of the invalidation acks. It *could* depending on the consistency model, I believe.

Anyway, most of this is theoretical. I suppose if this fixes an issue and results in the correct execution it's *probably* OK. Looking at the Ruby `functionalRead` code (in `RubySystem.cc`), it seems that we need an idea of "maybe stale" in the network as well as the caches, but that doesn't exist right now.

I feel more comfortable with `DATA_EXCLUSIVE` returning the functional read than just `DATA` without a lot more digging into the details of the protocol. `PUTX` I also feel OK about. From my guess, I would say that `PUTX_COPY` is explicitly stale as the L0 cache still has the data in `M` state (according to the comment) so it could have been updated. `DATA` is *probably* ok since there is a separate `STALE_DATA` message, but that also implies that the L1 can give data to the L0 that could have a more up to date copy somewhere else, so, it's unclear if that's safe or not.

nmosier:
I updated the patch to only return functional reads for DATA_EXCLUSIVE and PUTX types in response to @powerjg's insight.

ivanaamit (approved):

PR #1052, opened 2024-04-21
cpu-o3: Clear thread state in time buffers on thread exit

### LINKED ISSUE #1049: cpu-o3: executing MFENCE in two parallel SMT threads causes hang
**Describe the bug**
Executing an x86 `mfence` instruction in two sibling SMT threads on the O3 CPU causes the CPU to get stuck. Specifically, it appears that one of the `mfence`s reaches the head of the ROB but never becomes ready to execute.

This simple assembly proof of concept (`mfence.asm`) triggers the bug:
```
        global _start

_start:
        mov dword [rsp], eax
        mfence

        mov eax, 60
        mov edi, 0
        syscall
```

**Affects version**
develop @ [c54039d](https://github.com/gem5/gem5/commit/c54039da5bfe9d2101bfa3889252a15230053b11)

**gem5 Modifications**
None

**To Reproduce**
2. Assemble the POC: `nasm -felf64 -o mfence.o mfence.asm && ld -o mfence mfence.o`
3. Run it under SMT: `./build/X86/gem5.opt configs/deprecated/example/se.py --cpu-type=X86O3CPU --smt --caches -c './mfence;./mfence'`

**Terminal Output**
```
$ timeout -s INT 10 ./build/X86/gem5.opt configs/deprecated/example/se.py --cpu-type=X86O3CPU -c './mfence;./mfence' --smt --caches 

gem5 version DEVELOP-FOR-24.0
gem5 compiled Apr 20 2024 20:38:46
gem5 started Apr 20 2024 21:17:07
gem5 executing on cafe-cet, pid 1395698
command line: ./build/X86/gem5.opt configs/deprecated/example/se.py --cpu-type=X86O3CPU -c './mfence;./mfence' --smt --caches

**** REAL SIMULATION ****
Exiting @ tick 15376263500 because user interrupt received
```
As you can see, the simulation times out after 10 seconds without both threads exiting.

**Expected behavior**
Both threads should exit immediately.

**Host Operating System**
Ubuntu 22.04

**Host ISA**
X86

**Compiler used**
GCC 11.4.0

**Additional information**
None yet

nmosier:
Fix #1049. Clear only thread-specific state from the O3 CPU time buffers, rather than clearing *all* state for *all* threads.

Specifically, this patch adds a `clearStates()` method for all O3 time buffer structs in src/cpu/o3/comm.hh. Upon thread exit, `CPU::removeThread()` now invokes this method for all structs in each time buffer, rather than flushing out the time buffers (which nukes the states for all threads, not just the exiting one).

Change-Id: I48cd50e39b3cd5e0068ddfc19a03d9bbd3f31bd5

### DIFF
--- a/src/cpu/o3/comm.cc
+++ b/src/cpu/o3/comm.cc
@@ -1,0 +1,74 @@
+#include "cpu/o3/comm.hh"
+
+#include <algorithm>
+
+#include "cpu/o3/dyn_inst.hh"
+
+namespace gem5
+{
+
+namespace o3
+{
+
+/**
+ * Remove instructions belonging to given thread from plain
+ * instruction array. Automatically updates the array size.
+ */
+static void
+removeThreadInsts(ThreadID tid, DynInstPtr *insts, int &size)
+{
+    auto has_tid = [tid] (const DynInstPtr &inst) -> bool {
+        return inst->threadNumber == tid;
+    };
+    DynInstPtr *last = std::remove_if(insts, insts + size, has_tid);
+    size = last - insts;
+}
+
+void
+FetchStruct::clearStates(ThreadID tid)
+{
+    removeThreadInsts(tid, insts, size);
+}
+
+void
+DecodeStruct::clearStates(ThreadID tid)
+{
+    removeThreadInsts(tid, insts, size);
+}
+
+void
+RenameStruct::clearStates(ThreadID tid)
+{
+    removeThreadInsts(tid, insts, size);
+}
+
+void
+IEWStruct::clearStates(ThreadID tid)
+{
+    removeThreadInsts(tid, insts, size);
+    mispredictInst[tid] = nullptr;
+    pc[tid] = nullptr;
+    squash[tid] = false;
+    branchMispredict[tid] = false;
+}
+
+void
+TimeStruct::clearStates(ThreadID tid)
+{
+    // Reset the thread's decode, rename, IEW, and commit structs to
+    // their default values.
+    decodeInfo[tid] = DecodeComm();
+    renameInfo[tid] = RenameComm();
+    iewInfo[tid] = IewComm();
+    commitInfo[tid] = CommitComm();
+
+    decodeBlock[tid] = false;
+    decodeUnblock[tid] = false;
+    renameBlock[tid] = false;
+    renameUnblock[tid] = false;
+    iewBlock[tid] = false;
+    iewUnblock[tid] = false;
+}
+
+}
+}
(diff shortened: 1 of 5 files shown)


ivanaamit:
@andysan, we would appreciate it if you could review this PR. Thank you.

andysan:
I don't think I'm the right person for O3 SMT issues. Maybe @giactra or @rjc-arch can help?

src/cpu/o3/comm.cc
```diff
@@ -0,0 +1,74 @@
...
+ * Remove instructions belonging to given thread from plain
+ * instruction array. Automatically updates the array size.
+ */
+static void
+removeThreadInsts(ThreadID tid, DynInstPtr *insts, int &size)
+{
+    auto has_tid = [tid] (const DynInstPtr &inst) -> bool {
+        return inst->threadNumber == tid;
+    };
+    DynInstPtr *last = std::remove_if(insts, insts + size, has_tid);
```
giactra, line 23: How is this working? insts is actually pointing to a C-style array of DynInstPtr. What will remove_if actually do? I though it was supposed to be used with mutable types like std::vector. I have the gut feeling this might be wrong

ivanaamit:
@nmosier, could you address Giacomo's questions? Thank you.

src/cpu/o3/comm.cc
nmosier: Sorry for the slow response. Correct me if I'm wrong, but I think you might be thinking of [std::erase_if](https://en.cppreference.com/w/cpp/container/vector/erase2), which operates on C++ containers like std::vector and std::map? My understanding is that [std::remove_if](https://en.cppreference.com/w/cpp/algorithm/remove) works on arbitrary ranges, including C-style arrays. How std::remove_if works is for each element to be removed, it moves (specifically using a move assignment) an element from the end of the array to the index of the remove element.

giactra: You are right. However I was wondering whether pushing at the end of the array is enough. Shouldn't we de-allocate the dynamic instruction as well to avoid memory leaks?

src/cpu/o3/comm.cc
```diff
@@ -0,0 +1,74 @@
+#include "cpu/o3/comm.hh"
```
giactra, line 1: You should add a license header for the new file

giactra (changes requested):
Thanks for clearing my doubts, there are two issues that need to be addressed

giactra:
Bump @nmosier

ivanaamit:
@nmosier, can you address Giacomo's comments? Thanks.

ivanaamit:
@nmosier We're closing this PR due to over three months of inactivity. Feel free to reopen it if you resume work on it. Thanks!

src/cpu/o3/comm.cc
nmosier: I fixed these issues in a new pull request, #1681.

nmosier:
I fixed the issues raised here and reopened a new pull request, #1681. (I don't have permissions to reopen this pull request.)

PR #1054, opened 2024-04-21
arch-vega: Pass s_memtime through smem pipe

mjkpolo:
Previously, the cycle count was written with no
latency. Now, the cycle count is written when
the scalar memory pipeline executes s_memtime,
because a s_memtime should be treated like a
s_load_dwordx2. Through microbenchmarks, the
s_memtime latency has been set to 41 clock cycles.

Change-Id: I8860ab03e52d6dc9c80072cea99d6e0913201c45

### DIFF
--- a/src/mem/ruby/system/Sequencer.cc
+++ b/src/mem/ruby/system/Sequencer.cc
@@ -52,6 +52,7 @@
 #include "debug/RubyHitMiss.hh"
 #include "debug/RubySequencer.hh"
 #include "debug/RubyStats.hh"
+#include "gpu-compute/gpu_dyn_inst.hh"
 #include "mem/packet.hh"
 #include "mem/ruby/profiler/Profiler.hh"
 #include "mem/ruby/protocol/PrefetchBit.hh"
@@ -583,12 +584,14 @@ Sequencer::readCallback(Addr address, DataBlock& data,
         if (ruby_request) {
             assert((seq_req.m_type == RubyRequestType_LD) ||
                    (seq_req.m_type == RubyRequestType_Load_Linked) ||
+                   (seq_req.m_type == RubyRequestType_Memtime) ||
                    (seq_req.m_type == RubyRequestType_IFETCH));
         }
         if ((seq_req.m_type != RubyRequestType_LD) &&
             (seq_req.m_type != RubyRequestType_Load_Linked) &&
             (seq_req.m_type != RubyRequestType_IFETCH) &&
-            (seq_req.m_type != RubyRequestType_REPLACEMENT)) {
+            (seq_req.m_type != RubyRequestType_REPLACEMENT) &&
+            (seq_req.m_type != RubyRequestType_Memtime)) {
             // Write request: reissue request to the cache hierarchy
             issueRequest(seq_req.pkt, seq_req.m_second_type);
             break;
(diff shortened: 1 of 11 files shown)


mjkpolo:
@mattsinc something weird is happening in SCons. When I build for VEGA_X86, no problems, but when I build for ALL I can replicate the failure above. I fixed the above issue by including compute_unit.hh instead of gpu_dyn_inst.hh in Sequencer.cc, but now I have
```
src/gpu-compute/compute_unit.hh:46:10: fatal error: enums/PrefetchType.hh: No such file or directory
   46 | #include "enums/PrefetchType.hh"
      |          ^~~~~~~~~~~~~~~~~~~~~~~

```
I believe enums are generated by scons like DebugFlags:
https://github.com/gem5/gem5/blob/115322319c709a6e2a12e8b5c5299aa813d94633/src/SConscript#L214-L238
I'm assuming the issue is that the GPU.py SimObject:
https://github.com/gem5/gem5/blob/115322319c709a6e2a12e8b5c5299aa813d94633/src/gpu-compute/SConscript#L37-L42
 isn't added to SCons before the Sequencer
https://github.com/gem5/gem5/blob/115322319c709a6e2a12e8b5c5299aa813d94633/src/mem/ruby/system/SConscript#L68
Do you know of a way to add a dependence from `Sequencer.cc` to `SimObject('GPU.py'...`

src/mem/ruby/system/Sequencer.cc
```diff
@@ -52,6 +52,7 @@
 #include "debug/RubyHitMiss.hh"
 #include "debug/RubySequencer.hh"
 #include "debug/RubyStats.hh"
+#include "gpu-compute/gpu_dyn_inst.hh"
```
abmerop, line 55: This include is going to cause all kinds of problems because the GPU model is not built by default

abmerop (changes requested):
This seems overly intrusive and I'm not sure about this.  The instruction seems like it should just read the shader cycle value, why does it need to go to the cache?

mjkpolo:
> This seems overly intrusive and I'm not sure about this. The instruction seems like it should just read the shader cycle value, why does it need to go to the cache?

I agree, it definitely feels wrong. The reason I sent it to the SQC is because a memtime should be a memory request, which you waitcnt on to continue after. I still don't think gem5 *requires* a waitcnt after memtime with this change, but it matches the accuracy of the waitcnt following a memtime.
```
old version:

memtime -> waitcnt -> s_nop -> s_nop -> ... -> memtime -> waitcnt -> done

new version:
         /     41 cycles      \ 
        /                      \ 
       /   go to SQC            \                                  / 41 cycles \
memtime -> waiting for memtime... -> waitcnt -> s_nop -> s_nop -> memtime -> waitcnt -> done


```
We only see the 41 cycles of the first memtime, because there is no critical section left after the second memtime, so it doesn't delay execution by 41 cycles again.

mjkpolo:
Note: I will be closing this, making changes, and merging first with Matt P. before opening another pull request on the official repo.

src/mem/ruby/system/Sequencer.cc
mattsinc: This would definitely need to be wrapped in BUILD_GPU macro if we keep this

mattsinc:
@mjkpolo let's discuss further in meeting tomorrow.

mattsinc:
After discussing with @mjkpolo and @abmerop , consensus was that we should create a ScalarSequencer class for the GPU scalar and SQC caches and implement something like memtimeCallback there -- right now SQC/scalar are using Sequencer.cc instead of something like a GPUSequencer because they don't use vectors.  That is causing the root problem here because Sequencer is also used by non-GPU code.  So adding ScalarSequencer and moving the functionality Marco added to Sequencer.cc in current version to ScalarSequencer should resolve main concerns.

PR #1059, opened 2024-04-22
systemc: remove if clause in Gem5ToTlmBridgeBase

robhau:
In the payload event queue in Gem5ToTlmBridgeBase, the phase is checked twice for BEGIN_RESP. This commit removes the second if clause since it is unnecessary.

Duplicate if clause in line 234 & line 256

https://github.com/gem5/gem5/blob/dd2689905fcb77a65f190705ba5f2b793535e28c/src/systemc/tlm_bridge/gem5_to_tlm.cc#L234-L267

please correct me if I am missing something important

### DIFF
--- a/src/systemc/tlm_bridge/gem5_to_tlm.cc
+++ b/src/systemc/tlm_bridge/gem5_to_tlm.cc
@@ -253,15 +253,13 @@ Gem5ToTlmBridge<BITWIDTH>::pec(
         if (need_retry) {
             blockingResponse = &trans;
         } else {
-            if (phase == tlm::BEGIN_RESP) {
-                // Send END_RESP and we're finished:
-                tlm::tlm_phase fw_phase = tlm::END_RESP;
-                sc_core::sc_time delay = sc_core::SC_ZERO_TIME;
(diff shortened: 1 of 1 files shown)


ivanaamit (approved):

robhau:
@ivanaamit
Thank you for reviewing the PR 😄 .
I saw that the message of the merged commit contains the pull request message and not original message. Is there a particular reason for this?

powerjg:
> I saw that the message of the merged commit contains the pull request message and not original message. Is there a particular reason for this?

That's just how github works. We try to do squash and merge instead of merge most of the time, but the UI isn't very good and it's easy to pick the "wrong" one. I put "wrong" in quotes because we don't really have any official policy for how to choose.

robhau:
> > I saw that the message of the merged commit contains the pull request message and not original message. Is there a particular reason for this?
> 
> That's just how github works. We try to do squash and merge instead of merge most of the time, but the UI isn't very good and it's easy to pick the "wrong" one. I put "wrong" in quotes because we don't really have any official policy for how to choose.

Thanks for answering 😄.

Good to know! Then I will now keep the commit message in the first PR comment.

PR #1026, opened 2024-04-15
github,tests: Add Pyunit tests to CI GitHub Action Workflow

BobbyRBruce:
Due to an oversight, the PyUnit tests were not being run as part of the gem5 CI tests. This was because they are located in "tests/pyunit" instead of "tests/gem5", where the CI GitHub Action workflow searched for tests to run and where all other tests reside.

This adds the Pyunit tests as a seperate job in the CI GitHub Action's workflow.

PR #1060, opened 2024-04-22
arch-arm: Refactor PTW

giactra:
This PR is refactoring the Arm PageTableWalker in the following way:

1) Simplifying the currState handling logic (mainly the tear down)
2) Amending the TlbTestInterface APIs to use a RequestPtr reference
3) Use finalizePhysical even when MMU is off, which means allowing memory mapped m5ops to work also in that circumstance

### DIFF
--- a/src/arch/arm/mmu.cc
+++ b/src/arch/arm/mmu.cc
@@ -200,6 +200,26 @@ MMU::invalidateMiscReg()
 }
 
 Fault
+MMU::testAndFinalize(const RequestPtr &req,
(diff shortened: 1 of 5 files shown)


andysan (approved):

giactra:
Hi @BobbyRBruce there seems to be an issue with Pyunit tests after #1026

ivanaamit:
@giactra, please rebase on top of the current develop and the tests should pass. Thanks.

PR #1061, opened 2024-04-22
mem-cache: Remove power-of-2 requirement for TreePLRU num leaves

nmosier:
Remove the requirement in TreePLRU's implementation that the number of leaves (i.e., the number of cache ways) be a power of two. Firstly, on some recent processors, this is not the case---for example, Intel Golden Cove's L1D has 12 ways. Secondly, The implementation of TreePLRU appears to work just fine as-is with a way count that's not a power of two.

Change-Id: If2a27dc5bbe7a8e96684f79ce791df5c0b582230

### DIFF
--- a/src/mem/cache/replacement_policies/tree_plru_rp.cc
+++ b/src/mem/cache/replacement_policies/tree_plru_rp.cc
@@ -104,6 +104,7 @@ TreePLRU::TreePLRUReplData::TreePLRUReplData(
 TreePLRU::TreePLRU(const Params &p)
   : Base(p), numLeaves(p.num_leaves), count(0), treeInstance(nullptr)
 {
+    fatal_if(numLeaves < 2, "Number of leaves must be two or greater");
 }
 
 void
--- a/src/mem/cache/replacement_policies/tree_plru_rp.cc
+++ b/src/mem/cache/replacement_policies/tree_plru_rp.cc
@@ -104,8 +104,6 @@ TreePLRU::TreePLRUReplData::TreePLRUReplData(
 TreePLRU::TreePLRU(const Params &p)
   : Base(p), numLeaves(p.num_leaves), count(0), treeInstance(nullptr)
 {
-    fatal_if(!isPowerOf2(numLeaves),
-             "Number of leaves must be non-zero and a power of 2");
 }
 
 void


BobbyRBruce (dismissed):

src/mem/cache/replacement_policies/tree_plru_rp.cc
```diff
@@ -104,8 +104,6 @@ TreePLRU::TreePLRUReplData::TreePLRUReplData(
 TreePLRU::TreePLRU(const Params &p)
   : Base(p), numLeaves(p.num_leaves), count(0), treeInstance(nullptr)
 {
-    fatal_if(!isPowerOf2(numLeaves),
```
odanrc, line 107: It must be greater than 0 because of the modulus operator

odanrc (changes requested):

src/mem/cache/replacement_policies/tree_plru_rp.cc
odanrc: There is also a bug when num_leaves == 1: lines 203 and 208 initialize index as 0, and line 129 does 0-1/2

nmosier: I added a check that the number of leaves is >= 2 for now. I confirmed that it crashes if numLeaves == 1.

odanrc (approved):

giactra (commented):
This has nothing to do with the PR, but while I was reviewing the code (for the first time) I noticed the following:

```
    TreePLRUReplData* treePLRUReplData = new TreePLRUReplData(
        (count % numLeaves) + numLeaves - 1,
        std::shared_ptr<PLRUTree>(treeInstance));
```

There's something I don't understand in the index assignment in the constructor. Say that we have 4 leaves. That would imply a structure like the following

```
    1 
   / \
  0    0 
 / \  / \
A  B  C  D
```

Shouldn't the index of A be 3 (traversing the tree in BFS)? Why is it set to

(count % numLeaves) + numLeaves - 1 = 7 instead?

odanrc:
> This has nothing to do with the PR, but while I was reviewing the code (for the first time) I noticed the following:
> 
> ```
>     TreePLRUReplData* treePLRUReplData = new TreePLRUReplData(
>         (count % numLeaves) + numLeaves - 1,
>         std::shared_ptr<PLRUTree>(treeInstance));
> ```
> 
> There's something I don't understand in the index assignment in the constructor. Say that we have 4 leaves. That would imply a structure like the following
> 
> ```
>     1 
>    / \
>   0    0 
>  / \  / \
> A  B  C  D
> ```
> 
> Shouldn't the index of A be 3 (traversing the tree in BFS)? Why is it set to
> 
> (count % numLeaves) + numLeaves - 1 = 7 instead?

In your example numNodes=7 and numLeaves=4 (leaves are A, B, C and D). When instantiating A count is 0, so A´s index is 0+4-1=3. B´s index is 1+4-1=4, C´s index is 2+4-1=5, and D´s index is 3+4-1=6.

giactra:
> In your example numNodes=7 and numLeaves=4 (leaves are A, B, C and D). When instantiating A count is 0, so A´s index is 0+4-1=3. B´s index is 1+4-1=4, C´s index is 2+4-1=5, and D´s index is 3+4-1=6.

Yep for some weird math I was assuming 0 % 4 = 4. I need a break

giactra (approved):

PR #1065, opened 2024-04-23
cpu-kvm: Support perf counters on hybrid host architectures

### LINKED ISSUE #1064: cpu-kvm: KVM CPU reports zero instructions simulated when running on Intel Alder Lake host
**Describe the bug**
When running the X86KvmCPU on an E-core on an Intel Alder Lake host, the Kvm CPU reports zero instructions simulated. This  makes it impossible to schedule exits after N instructions to take a checkpoint, for example. 

The problem is that on hybrid architectures like Intel Alder Lake, there are multiple types of cores, each of which have their own set of supported hardware performance counters (i.e., performance counters of type PERF_TYPE_HARDWARE). Linux supports only creating a hardware performance counter for either a P-core or E-core, but not both. Furthermore, by default, all hardware performance counters created are bound to P-cores, not E-cores. Since the X86KvmCPU is not aware of this, it binds its performance counters implicitly to P-cores. If gem5 runs on an E-core, then these P-core counters are never enabled, thus always returning zero when read from.

**Affects version**
develop @ [1153223](https://github.com/gem5/gem5/commit/115322319c709a6e2a12e8b5c5299aa813d94633)

**gem5 Modifications**
None

**To Reproduce**
1. Compile gem5: `scons build/X86/gem5.opt`
2. Run the built-in hello world on an E-core of an Intel Alder Lake or Raptor Lake system (where CPU 23 is an E-core): `taskset -c 23 ./build/X86/gem5.opt configs/deprecated/example/se.py --cpu-type=X86KvmCPU -c tests/test-progs/hello/bin/x86/linux/hello`
3. Check the number of simInsts: `grep simInsts m5out/stats.txt`

**Terminal Output**
```
$ grep simInsts m5out/stats.txt
```
That is, simInsts isn't even present in m5out/stats.txt (since it's zero and thus omitted).

**Expected behavior**
The number of simInsts should be nonzero.

**Host Operating System**
Ubuntu 22.04

**Host ISA**
X86, Intel Alder Lake

**Compiler used**
GCC 11.4.0

**Additional information**
I will submit a patch for this shortly

nmosier:
Fix #1064 by adding support for hardware performance counters on hybrid architectures like Intel Alder Lake.

Hybrid architectures have multiple types of cores, each of which require the instantiation of a separate performance counter. The KVM CPU's PerfKvmCounter class was not aware of this, any only instantiated a single performance counter, implicitly bound to the P-core only. This meant that if gem5 ever ran on an E-core, the various hardware performance counters would not get updated properly, in some cases always zero (e.g., for the number of instructions executed).

This patch adds support for hybrid host architectures as follows. First, we convert PerfKvmCounter into an abstract class, which has two concrete implementations: SimplePerfKvmCounter and HybridPerfKvmCounter. The former is used for non-hybrid architectures or for non-hardware performance counters and is functionally equivalent to the prior implementation of PerfKvmCounter. The latter is used for instantiating hardware performance counters (i.e., of type PERF_TYPE_HARDWARE) on hybrid host architectures. It does so by internally instantiating two SimplePerfKvmCounters, one for a P-core and one for an E-core. Upon read, it sums the results of reading the two internal counters.

Change-Id: If64fcb0e2fcc1b3a6a37d77455c2b21e1fc81150

### DIFF
--- a/src/cpu/kvm/perfevent.cc
+++ b/src/cpu/kvm/perfevent.cc
@@ -73,7 +73,7 @@ PerfKvmCounter::PerfKvmCounter()
 }
 
 PerfKvmCounter *
-PerfKvmCounter::create()
+PerfKvmCounter::create(bool allow_hybrid)
 {
     // Check if we're running on a hybrid host architecture. Linux exposes
     // this via sysfs. If the directory /sys/devices/cpu exists, then we are
@@ -82,22 +82,25 @@ PerfKvmCounter::create()
     // existence of /sys/devices/cpu_atom to indicate a hybrid host
     // architecture.
     const char *atom_path = "/sys/devices/cpu_atom";
-    if (DIR *atom_dir = opendir(atom_path)) {
-        closedir(atom_dir);
+    if (allow_hybrid) {
+        if (DIR *atom_dir = opendir(atom_path)) {
+            closedir(atom_dir);
 
-        // Since we're running on a hybrid architecture, use a hybrid
-        // performance counter. This uses two 'physical' performance counters
-        // to implement a 'logical' one which is the sum of the two.
-        return new HybridPerfKvmCounter();
-    } else {
+            // Since we're running on a hybrid architecture, use a hybrid
+            // performance counter. This uses two 'physical' performance
+            // counters to implement a 'logical' one which is the sum of the
+            // two.
+            return new HybridPerfKvmCounter();
+        }
+
         if (errno != ENOENT)
             warn("Unexpected error code from opendir(%s): %s\n",
                  atom_path, std::strerror(errno));
-
-        // We're running on a regular architecture, so use a regular
-        // performance counter.
-        return new SimplePerfKvmCounter();
     }
+
+    // We're running on a regular architecture, so use a regular
+    // performance counter.
+    return new SimplePerfKvmCounter();
 }
 
 SimplePerfKvmCounter::SimplePerfKvmCounter()
@@ -178,7 +181,7 @@ SimplePerfKvmCounter::enableSignals(pid_t tid, int signal)
 }
 
 void
-SimplePerfKvmCounter::attach(PerfKvmCounterConfig &config,
+SimplePerfKvmCounter::attach(const PerfKvmCounterConfig &config,
                              pid_t tid, const PerfKvmCounter *parent)
 {
     assert(!attached());
@@ -288,8 +291,19 @@ SimplePerfKvmCounter::read(void *buf, size_t size) const
     } while (_size);
 }
 
+PerfKvmCounterConfig
+HybridPerfKvmCounter::fixupConfig(const PerfKvmCounterConfig &in,
+                                  ConfigSubtype config_subtype)
+{
+    PerfKvmCounterConfig out = in;
+    out.attr.config |= config_subtype;
+    if (out.attr.sample_period > 1)
+        out.attr.sample_period /= 2;
+    return out;
+}
+
 void
-HybridPerfKvmCounter::attach(PerfKvmCounterConfig &config, pid_t tid,
+HybridPerfKvmCounter::attach(const PerfKvmCounterConfig &config, pid_t tid,
                              const PerfKvmCounter *parent)
 {
     // We should only be using hybrid performance counters for hardware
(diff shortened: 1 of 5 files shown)


andysan:
Thanks for fixing this! I have only looked at this very briefly and I can't claim to understand all the intricacies of PMUs Intel's hybrid architecture.  I find the way perf seems to treat the P/E cores a bit odd, it's very different from how it deals with big/LITTLE cores in Arm systems where events are just migrated as the process migrates between cores. Anyway, that seems to be the way it works, so we need to handle that.

A potential issue I spotted when looking at this is how it interacts with thread migration. It seems like that would make it hard/impossible to reliably track overflows. For example, say that you have an instruction counter that has been programmed to stop the VM after 10M instructions and there is a thread migration after 5M instructions. Would the VM exit happen after 15M instructions in that case?

powerjg (dismissed):
Looks good to me. Thread migration is important, as Andreas mentioned, but can be left to another change.

nmosier:
Good point about the thread migrations. I'm working on a (surprisingly simple) solution for that but can submit it as a separate patch.

powerjg:
It's up to you. Since this is blocked on an unrelated testing failure, if you have a fix, feel free to amend this PR :)

nmosier:
Okay, I just submitted an additional commit which adds support for overflows when migrating threads across hybrid cores. Basically, all it does it halve the requested sample period N. This guarantees an overflow will always trigger before more than N events occur (but may trigger when as few as N/2 events occur). But it appears gem5 already knows to continue execution if we've seen fewer events (e.g., instructions or cycles) than expected and continues the thread execution until a total of N events have occurred.

I've tested this out locally on my Alder Lake machine by pinning two gem5 KVM instances to a P-core (0) and E-core (23):
```sh
#!/bin/bash

rm -r m5out.{1,2}

args="$@"

run() {
    taskset -c 0,23 ./build/X86/gem5.opt -d "$1" --debug-flag=KvmAll --debug-file=dbgout configs/deprecated/example/se.py --cpu-type=X86KvmCPU -c ./loop $args
}

run m5out.1 &
run m5out.2 &

wait

grep 'simInsts' m5out.{1,2}/stats.txt
```

`loop` is built from `loop.asm` with (`nasm -felf64 -o loop.o loop.asm && ld -o loop loop.o`), which has a very predictable instruction execution count:
```
        global _start

_start:
        mov rcx, 10000000000
loop:
        loop loop
end:
        mov eax, 60
        mov edi, 0
        syscall
```

Which, when run with a 5 billion instruction limit, prints out the following:
```
$ bash test.sh --max-insts=5000000000
...
Exiting @ tick 6577256850500 because a thread reached the max instruction count
src/cpu/kvm/base.cc:593: hack: Pretending totalOps is equivalent to totalInsts()
Exiting @ tick 9091946082000 because a thread reached the max instruction count
src/cpu/kvm/base.cc:593: hack: Pretending totalOps is equivalent to totalInsts()
m5out.1/stats.txt:simInsts                                   5000000002                       # Number of instructions simulated (Count)
m5out.2/stats.txt:simInsts                                   5000000001                       # Number of instructions simulated (Count)
```
The reason thread 1 takes much longer than thread 2 is that thread 1 is first scheduled on the slower E-core, but is rescheduled on the faster P-core once thread 2 finishes on the P-core.

If run to completion, it prints out the following:
```
$ bash test.sh
...
Exiting @ tick 18198550282500 because exiting with last active thread context
src/cpu/kvm/base.cc:593: hack: Pretending totalOps is equivalent to totalInsts()
m5out.1/stats.txt:simInsts                                  10000001937                       # Number of instructions simulated (Count)
m5out.2/stats.txt:simInsts                                  10000000004                       # Number of instructions simulated (Count)
```
I'm not sure why there's an extra 1933 instructions added when migrating thread 1 from an E-core to a P-core. Probably some Linux kernel oddity?

src/cpu/kvm/perfevent.cc
```diff
@@ -67,33 +68,54 @@ PerfKvmCounterConfig::~PerfKvmCounterConfig()
...
+PerfKvmCounter::create(bool allow_hybrid)
 {
-    attach(config, tid, parent);
+    // Check if we're running on a hybrid host architecture. Linux exposes
+    // this via sysfs. If the directory /sys/devices/cpu exists, then we are
+    // running on a regular architecture. Otherwise, /sys/devices/cpu_core and
+    // /sys/devices/cpu_atom should exist. For simplicity, we use the
+    // existence of /sys/devices/cpu_atom to indicate a hybrid host
+    // architecture.
+    const char *atom_path = "/sys/devices/cpu_atom";
```
andysan, line 84: It'd be nice if we could use std::filesystem instead but that requires C++17, so it might not be an option yet.

andysan (approved):
Thanks for this change and the thorough testing with thread migration. This looks good to me. Not sure about the additional instructions though. Could be some weird interaction between how we configure event filtering and the thread migration code.

PR #1071, opened 2024-04-25
cpu,arch: Add IsInvalid flag to Unknown insts

hnpl:
The IsInvalid flag indicates that the static instruction is not part of the executing ISA and not part of m5's pseudo-instructions. This flag provides a way to recognize an illegal instruction at the decode stage.

### DIFF
--- a/src/arch/arm/isa/insts/misc.isa
+++ b/src/arch/arm/isa/insts/misc.isa
@@ -848,7 +848,8 @@ let {{
     '''
     unknownIop = ArmInstObjParams("unknown", "Unknown", "UnknownOp", \
                                   { "code": unknownCode,
-                                    "predicate_test": predicateTest })
+                                    "predicate_test": predicateTest },
+                                  ['IsInvalid'])
     header_output += BasicDeclare.subst(unknownIop)
     decoder_output += BasicConstructor.subst(unknownIop)
     exec_output += PredOpExecute.subst(unknownIop)
--- a/src/arch/arm/isa/insts/misc64.isa
+++ b/src/arch/arm/isa/insts/misc64.isa
@@ -183,7 +183,7 @@ let {{
             return std::make_shared<UndefinedInstruction>(machInst, true);
     '''
     unknown64Iop = ArmInstObjParams("unknown", "Unknown64", "UnknownOp64",
-                                    unknownCode)
+                                    unknownCode, ['IsInvalid'])
     header_output += BasicDeclare.subst(unknown64Iop)
     decoder_output += BasicConstructor64.subst(unknown64Iop)
     exec_output += BasicExecute.subst(unknown64Iop)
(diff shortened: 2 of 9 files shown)


giactra (commented):
Hi, this makes sense to me. Is there a reason why Arm hasn't been included?

hnpl:
Hi Giacomo, I haven't figured out how to update the flag for Unknown/Unknown64 instructions in the Arm ISA yet.
If you don't mind, can you suggest a way to change the IsInvalid flag for the instructions?

giactra:
> Hi Giacomo, I haven't figured out how to update the flag for Unknown/Unknown64 instructions in the Arm ISA yet. If you don't mind, can you suggest a way to change the IsInvalid flag for the instructions?

Hi @hnpl , you just need to add the flag [after the unknownCode](https://github.com/gem5/gem5/blob/stable/src/arch/arm/isa/insts/misc64.isa#L185) for Unknown64 (and do something similar for the 32bit version)

hnpl:
@giactra Thanks for the suggestion. I think the problem is that, for the flag to be available to the decoder, the flag must be set in the StaticInst constructor. Since `Unknown64` and `Unknown` are using `BasicConstructor64` and `BasicConstructor`, there's currently no way to inject the flag-related code to the constructors.

In terms of solution, I can modify the `BasicConstructor64` and `BasicConstructor` templates; however, I think it would be a scope-creep change to those template. I think a better way would be defining separate templates for `Unknown64`'s and `Unknown`'s constructors.

hnpl:
@giactra You are correct that we can pass the flag to the ArmInstObjParams. I modified the code as you suggested.

giactra (approved):
This LGTM, one comment though which you don't have to address with this PR:

Do you have a rough idea on how this will be used? I am asking this as Arm has multiple definitions of invalid instructions:

1) Unknown/Unknown64 for instructions unspecified in the ISA
2) FailUnimplemented for instructions specified in the ISA but that we don't implement
3) DecoderFault for instructions which have some other issues (like alignment faults)

So depending on the usage of IsInvalid we might consider adding at least point two as well. Anyway this can be done by later PR once it is clear what will be the usage of the flag

hnpl:
@giactra Those are good points. I have a specific use case when you can speculatively decode a whole cache block without knowing the start of 1 instruction within the block. In the RISC-V case, it is a mixture of 16-bit and 32-bit instructions and this flag allows you to halt the speculative decoding when you encounter an instruction that is not in the instruction set. I believe you can do similar things in ARM too with the presence of Thumb instructions.

My definition of the Invalid flag is that, a static instruction is invalid if it is not specified by the instruction set. In gem5 case, an instruction is illegal if it is not in the target ISA and not one of the m5 pseudo-instructions.

In the point two you mention, it is an instruction that is not supported by the microarchitecture but is specified by the ISA; so the instruction in point two does not fall within the definition of an invalid instruction. I believe it should trigger an Illegal instruction fault.

However, now that I think about it, an instruction that is not implemented is as good as an unknown instruction. I'll update the flag for those instructions in point two to have the invalid flag in a different PR.

PR #1073, opened 2024-04-25
mem-cache: Revert "Remove power-of-2 requirement for..."

ivanaamit:
Reverts gem5/gem5#1061. Caused error https://github.com/gem5/gem5/actions/runs/8828536310/job/24242396632#step:23:42.

```
src/mem/dram_interface.cc:690: warn: DRAM device capacity (16384 Mbytes) does not match the address range assigned (2048 Mbytes)
src/mem/dram_interface.cc:690: warn: DRAM device capacity (16384 Mbytes) does not match the address range assigned (2048 Mbytes)
src/sim/kernel_workload.cc:46: info: kernel located at: /__w/gem5/gem5/tests/gem5/resources/arm64-linux-kernel-5.4.49
src/base/loader/symtab.cc:95: warn: Cannot insert a new symbol table due to name collisions. Adding prefix to each symbol's name can resolve this issue.
src/mem/cache/replacement_policies/tree_plru_rp.cc:107: fatal: fatal condition numLeaves < 2 occurred: Number of leaves must be two or greater
Memory Usage: 4934480 KBytes
```

### DIFF
--- a/src/mem/cache/replacement_policies/tree_plru_rp.cc
+++ b/src/mem/cache/replacement_policies/tree_plru_rp.cc
@@ -104,7 +104,8 @@ TreePLRU::TreePLRUReplData::TreePLRUReplData(
 TreePLRU::TreePLRU(const Params &p)
   : Base(p), numLeaves(p.num_leaves), count(0), treeInstance(nullptr)
 {
-    fatal_if(numLeaves < 2, "Number of leaves must be two or greater");
+    fatal_if(!isPowerOf2(numLeaves),
+             "Number of leaves must be non-zero and a power of 2");
 }
 
 void


ivanaamit:
@nmosier @odanrc, we are reverting this change because the Daily tests are failing.

Harshil2107:
The test is still failing
```
Logging call to command: /home/harshilp/worktrees/fix-m5/build/ALL/gem5.opt -d /tmp/gem5out0zxhoirz -re --silent-redirect /home/harshilp/worktrees/fix-m5/tests/gem5/arm_boot_tests/configs/arm_boot_exit_run.py --cpu atomic --num-cpus 4 --mem-system no_cache --dram-class HBM2Stack --resource-directory /home/harshilp/worktrees/fix-m5/tests/gem5/resources
Starting Test Suite: atomic-cpu_4-cores_no_cache_HBM2Stack_arm_boot_test_m5-exit-ALL-x86_64-opt 
Starting Test Case: atomic-cpu_4-cores_no_cache_HBM2Stack_arm_boot_test_m5-exit-ALL-x86_64-opt
Test: atomic-cpu_4-cores_no_cache_HBM2Stack_arm_boot_test_m5-exit-ALL-x86_64-opt Passed
Logging call to command: /home/harshilp/worktrees/fix-m5/build/ALL_CHI/gem5.opt -d /tmp/gem5outqx136m4m -re --silent-redirect /home/harshilp/worktrees/fix-m5/tests/gem5/arm_boot_tests/configs/arm_boot_exit_run.py --cpu timing --num-cpus 2 --mem-system chi --dram-class DualChannelDDR4_2400 --resource-directory /home/harshilp/worktrees/fix-m5/tests/gem5/resources
Starting Test Suite: timing-cpu_2-cores_chi_DualChannelDDR4_2400_arm_boot_test_m5-exit-ALL-x86_64-opt-CHI 
Starting Test Case: timing-cpu_2-cores_chi_DualChannelDDR4_2400_arm_boot_test_m5-exit-ALL-x86_64-opt-CHI
Test: timing-cpu_2-cores_chi_DualChannelDDR4_2400_arm_boot_test_m5-exit-ALL-x86_64-opt-CHI Failed
================================================== Results: 1 Passed, 1 Failed in 3.1e+02 seconds  =====================================================
```

Harshil2107:
never mind, the tests pass, it was a build issue, they were breaking with --skip-build as I only had ALL build and not every build the tests require.

odanrc:
@ivanaamit , this means that there might be a test with an out-of-bounds access. Reverting it would not be a proper solution, as it would just hide that a bug exists.

The associativity (param name: assoc) of whichever cache is using this RP is set to 1. Since 1 is a valid associativity, the RP should work with 1. We may have been extra cautious with this restriction, though, so what I recommend as a first step to fix it is to change line 194 from
`return candidates[tree_index - (numLeaves - 1)];`
to
`return candidates.at(tree_index - (numLeaves - 1));`
and to change that assertion from < 2 to < 1 (numLeaves should never be 0, as caches´ sizes must always be greater than 0). If nothing breaks after this change, then we are all good, and that assertion was protecting from an assumed error that couldn´t really happen. I *think* that this should work, and that we were just being overzealous. Could you give this a try?

If that does not work, then the bug really exists, so the solution would be to rework TreeLRUPR::instantiateEntry to either have a special scenario for when numLeaves is 1, or to shift all indices by 1.

powerjg:
> Reverting it would not be a proper solution, as it would just hide that a bug exists

I'm not sure I agree with this statement. Until we have a proper fix, I think we should revert this commit.

The problem is that with the commit merged, we are **MASKING** other possible errors in the daily tests since there is a failure because of #1061. Before #1061 there were no failures. It is a big problem to have tests failing.

We'll try your quick fix, but if we can't get it to pass by end of day today, we're going to merge the revert.

ivanaamit:
@odanrc, your proposed change is passing the tests. I will push a PR shortly. Thank you!

PR #1078, opened 2024-04-28
arch-riscv: Add Integer Conditional operations extension (Zicond) instructions

cyyself:
This PR added RISC-V Integer Conditional Operations Extension, which is in the RVA23U64 Profile Mandatory Base. And the performance of conditional move instructions in micro-architecture is an interesting point to explore.

Zicond instructions added: czero.eqz, czero.nez

Changes based on spec:
https://github.com/riscvarchive/riscv-zicond/releases/download/v1.0.1/riscv-zicond_1.0.1.pdf


hnpl (approved):
Thanks for the contribution!

rogerchang23424 (approved):

ivanaamit (approved):

PR #1079, opened 2024-04-28
arch-riscv: Fix narrowing/widening type-convert instructions

QQeg:
Correct ei calculation under VectorFloatWideningCvtFormat and VectorFloatNarrowingCvtFormat.

Change-Id: I08699ffe3b9f8a7d4543023437626cc054344053

### DIFF
--- a/src/arch/riscv/isa/formats/vector_arith.isa
+++ b/src/arch/riscv/isa/formats/vector_arith.isa
@@ -801,7 +801,7 @@ def format VectorFloatWideningCvtFormat(code, category, *flags) {{
     set_src_reg_idx += setSrcWrapper(src3_reg_id)
     set_src_reg_idx += setSrcVm()
     code = maskCondWrapper(code)
-    code = eiDeclarePrefix(code)
+    code = eiDeclarePrefix(code, widening=True)
     code = loopWrapper(code)
     code = fflags_wrapper(code)
 
@@ -858,7 +858,7 @@ def format VectorFloatNarrowingCvtFormat(code, category, *flags) {{
     set_src_reg_idx += setSrcWrapper(src3_reg_id)
     set_src_reg_idx += setSrcVm()
     code = maskCondWrapper(code)
-    code = eiDeclarePrefix(code)
+    code = eiDeclarePrefix(code, widening=True)
     code = loopWrapper(code)
     code = fflags_wrapper(code)
     code = narrowingOpRegisterConstraintChecks(code)


src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -858,7 +858,7 @@ def format VectorFloatNarrowingCvtFormat(code, category, *flags) {{
     set_src_reg_idx += setSrcWrapper(src3_reg_id)
     set_src_reg_idx += setSrcVm()
     code = maskCondWrapper(code)
-    code = eiDeclarePrefix(code)
+    code = eiDeclarePrefix(code, widening=True)
```
hnpl, line 861: This is a narrowing instruction so having the parameter "widening=True" is quite odd.

hnpl (commented):
Thanks for addressing this issue. I have a couple of questions about how the instructions are implemented.

I'm assuming that this change addresses the instructions in 13.18 and 13.19 of the rvv 1.0 spec. The instructions double/halve the size of the resulting data type.

I do not completely understand how the narrowing/widening instructions are implemented here, so please correct me if I'm wrong. The eiDeclarePrefix function chooses how `ei` are determined, i.e.,
- If widening/narrowing instructions,
```
uint32_t ei = i + micro_vlmax * this->microIdx;
```
- Otherwise,
```
```

Here, micro_vlmax are determined as follows for both widening/narrowing instructions,

```
    const int64_t vlmul = vtype_vlmul(machInst.vtype8);
    const int32_t t_micro_vlmax = vtype_VLMAX(machInst.vtype8, vlen, true);
    const int32_t micro_vlmax = vlmul < 0 ? t_micro_vlmax : t_micro_vlmax / 2;
```

So, the `micro_vlmax` is always `SEW * LMUL / 2` if `LMUL > 1`, which means the `emul` (effective lmul?) of the destination regs are always halved. Isn't this an incorrect behavior for the widening instructions?

QQeg:
Hi @hnpl,

For the code you mentioned:
```
    const int64_t vlmul = vtype_vlmul(machInst.vtype8);
    const int32_t t_micro_vlmax = vtype_VLMAX(machInst.vtype8, vlen, true);
    const int32_t micro_vlmax = vlmul < 0 ? t_micro_vlmax : t_micro_vlmax / 2;
```

Firstly, from `utility.hh`, we can see that the encoding of LMUL to lmul is as follow:
https://github.com/gem5/gem5/blob/1bb5d3b99e74da0a144e14d7eaab7cfd04764a50/src/arch/riscv/utility.hh#L234-L245

Secondly, `t_micro_vlmax` is equal to `VLEN / SEW` if `LMUL >= 1`, since it is in the scope of a single microop, not the entire instruction. 
You can see the implementation of `vtype_VLMAX` here: 
(Note that the argument `per_reg` is set to be `true` when calculating `t_micro_vlmax`.)
https://github.com/gem5/gem5/blob/1bb5d3b99e74da0a144e14d7eaab7cfd04764a50/src/arch/riscv/utility.hh#L253-L261

As a result, `micro_vlmax` is always `(VLEN / SEW) / 2` if `LMUL >= 1`.

And the reason why `micro_vlmax` is halved even when `LMUL == 1` is that the widening/narrowing instructions use `VectorIntWideningMacroConstructor` as their template for decoder_output.
The following code is from template `VectorIntWideningMacroConstructor`:
https://github.com/gem5/gem5/blob/1bb5d3b99e74da0a144e14d7eaab7cfd04764a50/src/arch/riscv/isa/templates/vector_arith.isa#L366-L370

As we can see from line 370, the widening/narrowing instructions will be turned into `2 * LMUL` microops when `LMUL >= 1`, which differs from other instructions.

This is exacly the reason why we should change the calculation of `ei` from,
```
```
to
```
uint32_t ei = i + micro_vlmax * this->microIdx;

// Actually this is the same as saying
// ei = i + (vtype_VLMAX(vtype, vlen, true) / 2) * this->microIdx;
// in this case
```
or else the index will be out of bounds.

src/arch/riscv/isa/formats/vector_arith.isa
QQeg: I agree that it's a bit absurd.
I just followed the pattern that has been done before.
Perhaps we could introduce another parameter called `narrowing` to reduce confusion?

hnpl (approved):
Thanks for the explanation!

hnpl:
This is beyond the scope of this fix, but I'm not clear on why there are `2*LMUL` micro-ops for narrowing instruction. Is that because each micro-op uses 1 source register, does the operation, then writes to a half of a destination register? If that's the case then I don't think we need 2 micro-ops when `LMUL==1`.

QQeg:
For a single narrowing instruction, let's say `vfncvt.f.f.w`, with `vlen = 256` and `LMUL=1`, it's actually narrowing a `512-bit` number to a `256-bit` number.
While in a single microop, we only deal with a maximum of `vlen-bit` number. In this case, that means converting a `256-bit` number from source reg to half(`128-bit`) of the destination reg, thus requiring a total of 2 microops.
 I believe this is why there are `2*LMUL` micro-operations for a narrowing instruction.

hnpl:
Ah I got it. Thanks!

For the context, the narrowing instructions convert elements of size 2\*SEW to elements of size SEW. The input consists of 2\*LMUL vector registers, and the output consists of 1\*LMUL vector registers. So, the SEW and LMUL values we obtained from the vtype are those of the output, not the input.

ivanaamit (approved):

Issue #862, opened 2024-02-09
CLFLUSH causes TLB assertion error


hnpl:
**Describe the bug**

Executing `clflush` instruction in FS mode caused this assertion error,

```
src/sim/simulate.cc:199: info: Entering event queue @ 14392355269750.  Starting simulation...                                                                                            
src/sim/simulate.cc:199: info: Entering event queue @ 15473386459500.  Starting simulation...                                                                      
src/sim/power_state.cc:105: warn: PowerState: Already in the requested power state, request ignored                                                                
src/sim/simulate.cc:199: info: Entering event queue @ 15474400558750.  Starting simulation...                                                                      
src/mem/ruby/system/Sequencer.cc:668: warn: Replacement policy updates recently became the responsibility of SLICC state machines. Make sure to setMRU() near callba
cks in .sm files!                        
gem5.opt: src/cpu/translation.hh:257: void gem5::DataTranslation<ExecContextPtr>::finish(const Fault&, const RequestPtr&, gem5::ThreadContext*, gem5::BaseMMU::Mode)
 [with ExecContextPtr = gem5::TimingSimpleCPU*; gem5::Fault = std::shared_ptr<gem5::FaultBase>; gem5::RequestPtr = std::shared_ptr<gem5::Request>]: Assertion `mode
== state->mode' failed.                  
Program aborted at tick 15478398983501
```

The problem seems to be that,
- In the gem5's X86 ISA implementation, `clflush*` is a store instruction.
- In TimingSimpleCPU, the translation of a store instruction will set the TLB to `BaseMMU::Write` mode.
- While translation is happening, the state of the translation is changed to `BaseMMU::Read` as the mem request caused by `clflush` is a `isCacheClean` request. [https://github.com/gem5/gem5/blob/v23.1.0.0/src/arch/x86/tlb.cc#L556-L564](https://github.com/gem5/gem5/blob/v23.1.0.0/src/arch/x86/tlb.cc#L556-L564)
- This resulted in a state mismatch, and subsequently caused the mention assertion error, which requires the state of the TLB and the state of translation to be the same.

**Affects version**
v23.1

**gem5 Modifications**
N/A

**Additional information**
Related issue: #226 
Related change: #592 

Reverting #592 fixes the error on my side, but it's not ideal due to issue #226.
I think we can make the `clflush*` instructions to be load instructions, but I'm not sure if that change is functionally correct.

AKKamath:
A possible fix is to change line 257 to:
```
[...]

PR #1080, opened 2024-04-28
arch-x86: Fix TLB Assertion Error on CFLUSH

Lukas-Zenick:
Fixed the assertion statement in the cpu's translation.hh file so that it doesn't fail the assertion if the cache is clean.

I compile this c code to `test`
```c
#include <stdio.h>

static inline void clflush(volatile void *p) {
    __asm__ volatile ("clflush (%0)" : : "r"(p) : "memory");
}

int main() {
    int data = 42;  // Example variable

    printf("Value before clflush: %d\n", data);

    clflush(&data);

    printf("Value after clflush: %d\n", data);

    return 0;
}
```
And run it with this script
`./build/X86/gem5.opt configs/learning_gem5/part1/two_level.py ./test`
In order to verify that it no longer fails the assertion check.

GitHub Issue: #862 
Change-Id: I6004662e7c99f637ba0ddb07d205d1657708e99f

### DIFF
--- a/src/cpu/translation.hh
+++ b/src/cpu/translation.hh
@@ -254,7 +254,7 @@ class DataTranslation : public BaseMMU::Translation
            BaseMMU::Mode mode)
     {
         assert(state);
-        assert(mode == state->mode);
+        assert(mode == state->mode || req->isCacheClean());
         if (state->finish(fault, index)) {
             if (state->getFault() == NoFault) {
                 // Don't access the request if faulted (due to squash)


hnpl:
I still think we need to have the assertion `mode == state->mode` without the `||`. However, I'm okay with this if it means this change allows clflush to work.

src/cpu/translation.hh
```diff
@@ -254,7 +254,7 @@ class DataTranslation : public BaseMMU::Translation
            BaseMMU::Mode mode)
     {
         assert(state);
-        assert(mode == state->mode);
+        assert(mode == state->mode || req->isCacheClean());
```
powerjg, line 257: I feel a little uncomfortable with this change. However, if it is what makes the most sense (e.g., this can't be fixed by changing the original `Mode`, then I'm good with this change.

Thanks for the contribution!

hnpl (approved):

hnpl:
Is there a plan to merge this to v24.0 release?

ivanaamit (approved):

giactra:
Hi, I got to see this PR randomly as it is tagged as x86 but it actually affects every ISA as it touches the generic src/cpu/translation.hh code.

I believe the current solution, while it fixes the problem for X86 is a hack for all other ISAs and I don't see the reason why this couldn't be fixed properly in the x86 code. As far as I understood from a quick look at the issue, the problem arises as CFLUSH is treated as a read for permission checking:

> // CLFLUSHOPT/WB/FLUSH is treated as read for protection checks but modelled as write
> assert(mode == state->mode || req->isCacheClean());

Well it shouldn't be hard to amend the permission checking logic in the TLB to treat the CacheClean as a corner case for BaseMMU::Write requests. I believe we do something similar in the ArmMMU code. I propose to revert this PR until things are done properly

ivanaamit:
@giactra, we apologize for the oversight. I have created the revert PR.

giactra:
> @giactra, we apologize for the oversight. I have created the revert PR.

No worries, thanks for handling this so quickly!

PR #1089, opened 2024-04-30
arch-generic: Fix reading from special :semihosting-features file

arichardson:
The implementation of SYS_FLEN was missing, which caused picolibc to treat this file as not implemented. Additionally, there was a bug in the SYS_READ call that was comparing the wrong variable against the passed buffer length. It was comparing the current file position against the buffer length instead of the number of written bytes. Finally, pos was unititialized which could result in spurious errors.

Change-Id: I8b487a79df5970a5001d3fef08d5579bb4aa0dd0

### DIFF
--- a/src/arch/generic/semihosting.cc
+++ b/src/arch/generic/semihosting.cc
@@ -94,7 +94,7 @@ const std::map<uint64_t, const char *> BaseSemihosting::exitCodes{
         {0x20029, "semi:ADP_Stopped_DivisionByZero"},
 };
 
-const std::vector<uint8_t> BaseSemihosting::features{
+const std::array<uint8_t, 5> BaseSemihosting::features{
         0x53, 0x48, 0x46, 0x42, // Magic
         0x3, // EXT_EXIT_EXTENDED, EXT_STDOUT_STDERR
 };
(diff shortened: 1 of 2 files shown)


src/arch/generic/semihosting.cc
```diff
@@ -649,18 +649,26 @@ BaseSemihosting::FileBase::flen()
     return -EINVAL;
 }
 
+static const std::vector<uint8_t> features;
```
giactra: I am a bit confused. How is this used?

arichardson: Oops that accidentally slipped into the uploaded patch
```suggestion
```

src/arch/generic/semihosting.cc
```diff
@@ -649,18 +649,26 @@ BaseSemihosting::FileBase::flen()
...
+
```
arichardson:
```suggestion
```

giactra (approved):

giactra:
@Harshil2107 @BobbyRBruce for some reason we are failing a simpoints-se-restore test. This is happening for every new PR. Was the stored checkpoint changed recentely?

Harshil2107:
yes, It was updated in 1040, I will update the database to get the tests passing and I will also update the simpoint example so future updates dont break the tests for everyone.

PR #1090, opened 2024-04-30
arch-generic: More reliable special file name handling in semihosting

arichardson:
Currently, the filesRootDir is prepended for all paths that do not start with '/'. However, we should not be doing this for the special files :tt and :semihosting-features. Noticed this while testing semihosting with a non-empty filesRootDir.

Change-Id: I156c8b680cb71cdc88788be3b0e93fc1d52e11e5


giactra (approved):
Looks good to me, thanks for this change!

PR #1093, opened 2024-05-02
gpu-compute: Invalidate Scalar cache when SQC invalidates

abmerop:
The scalar cache is not being invalidated which causes stale data to be left in the scalar cache between GPU kernels. This commit sends invalidates to the scalar cache when the SQC is invalidated. This is a sufficient baseline for simulation.

Since the number of invalidates might be larger than the mandatory queue can hold and no flash invalidate mechanism exists in the VIPER protocol, the command line option for the mandatory queue size is removed, which is the same behavior as the SQC.

Change-Id: I1723f224711b04caa4c88beccfa8fb73ccf56572

### DIFF
--- a/configs/ruby/GPU_VIPER.py
+++ b/configs/ruby/GPU_VIPER.py
@@ -498,13 +498,6 @@ def define_options(parser):
         "--noL1", action="store_true", default=False, help="bypassL1"
     )
     parser.add_argument(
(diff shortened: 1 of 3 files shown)


mattsinc (approved):

mattsinc:
@v-ramadas FYI

PR #1097, opened 2024-05-02
util: Bump gpu-fs docker to ROCm 6.1

abmerop:
This version matches the disk image on gem5-resources.

Change-Id: I69a45ef290f0fdf2167ead4d67d4d789d30e0e91


mattsinc (dismissed):
From my perspective this is good, but @BobbyRBruce should probably chime in too

mattsinc (approved):
From my perspective this is good, but @BobbyRBruce should probably chime in too

abmerop:
Updated README

abmerop:
I don't think Bobby needs to review, but I think he updates the ghcr images which should be done for the release

BobbyRBruce:
Sorry for the delay on this, but this image on our docker registry has been updated to incorporate this change as of this morning.

PR #1099, opened 2024-05-03
arch-riscv: Fix CSR instruction behavior 2nd attempts

rogerchang23424:
Quote from change[1]

> The RISC-V spec clarifies the CSR instruction operation, some of them
shall not read or write CSR by the hints of RD/RS1/uimm, but the
original version use the 'data != oldData' condition to determine
whether write or not, and always read CSR first.
See CSR instruction in spec:
Section 9.1 Page 56 of https://github.com/riscv/riscv-isa-manual/releases/download/Ratified-IMAFDQC/riscv-spec-20191213.pdf

|||Register operand|||
|--- |--- |--- |--- |--- |
|Instruction|rd is x0|rs1 is x0|Reads CSR|Writes CSR|
|CSRRW|Yes|-|No|Yes|
|CSRRW|No|-|Yes|Yes|
|CSRRS/CSRRC|-|Yes|Yes|No|
|CSRRS/CSRRC|-|No|Yes|Yes|
|||Immediate operand|||
|Instruction|rd is x0|uimm = 0|Reads CSR|Writes
CSR|
|CSRRWI|Yes|-|No|Yes|
|CSRRWI|No|-|Yes|Yes|
|CSRRSI/CSRRCI|-|Yes|Yes|No|
|CSRRSI/CSRRCI|-|No|Yes|Yes|

The issue cause the ubuntu hanging because we shared the same status CSR with `mstatus`, `sstatus` and `ustatus` and interrupt enabling CSR with mip, sip and uip. We may need to read origin CSR without effect of unmask bits to avoid override the bits of other CSR. Now the ubuntu can work after the patch merged.

[1] https://gem5-review.googlesource.com/c/public/gem5/+/67717

### DIFF
--- a/src/arch/riscv/isa.cc
+++ b/src/arch/riscv/isa.cc
@@ -576,6 +576,11 @@ ISA::readMiscReg(RegIndex idx)
         {
             return readMiscRegNoEffect(MISCREG_FFLAGS) & FFLAGS_MASK;
         }
+      case MISCREG_FCSR:
+        {
+            return readMiscRegNoEffect(MISCREG_FFLAGS) |
+                  (readMiscRegNoEffect(MISCREG_FRM) << FRM_OFFSET);
+        }
       default:
         // Try reading HPM counters
         // As a placeholder, all HPM counters are just cycle counters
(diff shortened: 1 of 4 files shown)


powerjg:
Can you expand a little more on "The issue cause the ubuntu hanging"

As far as I know, all of our ubuntu boot tests are passing. Is this a problem with 24.04? A new kernel?

rogerchang23424:
Hi @powerjg,

Previous there is a change try to fix read/write CSR behavior
1. https://gem5-review.googlesource.com/c/public/gem5/+/67717

They are got reverted [CL](https://gem5-review.googlesource.com/c/public/gem5/+/68637) because the ubuntu hangs. Because some of CSR registers is physically shared. For example, the register `sstatus` and `mstatus` shared the same physical space. If we execute the `csrrw zero, sstatus, t0`. We get the all of unmask data zero and write the t0 to same physical space. It leads to overwrite some bits of `mstatus` to zero and the ubuntu can't work well.

The PR also split the some of CSRs to different miscreg index and handle them individually to avoid overwriting issue. I have test with `ubuntu-20.04` image and it works well. So I think it will work in ubuntu 24.04? I haven't test it.

powerjg (approved):
Thanks for the info!

PR #1100, opened 2024-05-03
mem-ruby: Implement NS bit for CHI transactions

giactra:
This patch is adding the NS bit to CHI requests to make sure they are properly tagged according to their security

Change-Id: I33d3610edefbb5a05a6090e9125c35d4fb8bca58
Reviewed-by: Tiago Muck <tiago.muck@arm.com>


ivanaamit (approved):
I am not an expert in this area, but the code looks good to me.

PR #1101, opened 2024-05-03
mem-ruby: Implement MakeReadUnique in CHI

giactra:
Change-Id: I64cd3c62804cca184d68287fc099534e9205f2b8
Reviewed-by: Tiago Muck <tiago.muck@arm.com>


powerjg (approved):
Seems fine to me.

I assume that you have internal tests for this. It would be super useful for the community to have those tests available for others to use.

vsoriap:
Hi, nice to have this feature in CHI. However, I see that the event type MakeReadUnique is never used for sequencer requests (such as load, store, or AMO. Would you plan to develop the condition to select whether ReadUnique or MakeReadUnique are issued?

PR #1103, opened 2024-05-03
gpu: Consolidated fixes for v24.0

abmerop:
Includes fixes for several bugs reported via email, self found, and internal reports.  Also includes runs through Valgrind and UBsan. See individual commits for more details.

### DIFF
--- a/src/arch/amdgpu/vega/operand.hh
+++ b/src/arch/amdgpu/vega/operand.hh
@@ -490,7 +490,7 @@ namespace VegaISA
         typename std::enable_if<Condition, void>::type
         setBit(int bit, int bit_val)
         {
-            DataType &sgpr = *((DataType*)srfData.data());
+            GEM5_ALIGNED(8) DataType &sgpr = *((DataType*)srfData.data());
             replaceBits(sgpr, bit, bit_val);
         }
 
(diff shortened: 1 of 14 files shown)


src/arch/amdgpu/vega/operand.hh
```diff
@@ -490,7 +490,7 @@ namespace VegaISA
         typename std::enable_if<Condition, void>::type
         setBit(int bit, int bit_val)
         {
-            DataType &sgpr = *((DataType*)srfData.data());
+            GEM5_ALIGNED(8) DataType &sgpr = *((DataType*)srfData.data());
```
mattsinc, line 493: I'm not sure I understand this datatype, but I get the concept

src/gpu-compute/compute_unit.cc
```diff
@@ -1772,7 +1772,7 @@ ComputeUnit::ScalarDataPort::MemReqEvent::process()
         SystemHubEvent *resp_event = new SystemHubEvent(pkt, &scalarDataPort);
         compute_unit->shader->systemHub->sendRequest(pkt, resp_event);
     } else if (!(scalarDataPort.sendTimingReq(pkt))) {
-        scalarDataPort.retries.push_back(pkt);
+        scalarDataPort.retries.emplace_back(pkt);
```
mattsinc, line 1775: From a memory leak perspective, what is the difference between emplace_back and push_back?

mattsinc (approved):

Issue #591, opened 2023-11-22
Can not build tlm with ARM


Harshil2107:
**Description**
I am making this issue based on an email from gem5 dev mailing list. The original mail is https://www.mail-archive.com/gem5-dev@gem5.org/msg44952.html. The email states that they can not build gem5 with ARM and systemC support using tlm. They can build gem5 and can also build gem5 as a library as stated in the README in util/tlm/README but the `scons` command fails. I have also been able to reproduce this issue.

I think this is due to scons missing a dependency that is required to build.

**Affects version**
23.0

**To Reproduce**
Steps to reproduce the behavior. Please assume starting from a clean repository:

1. Compile gem5 with command: `scons build/ARM/gem5.opt -j {cpus}`
2. Follow the steps highlighted in util/tlm/README.

```sh
scons build/ARM/gem5.opt
scons --with-cxx-config --without-python --without-tcmalloc USE_SYSTEMC=0 build/ARM/libgem5_opt.so
cd util/tlm
scons
```

For Mac OSX this command should be used:

```sh
scons build/ARM/gem5.opt
scons --with-cxx-config --without-python --without-tcmalloc USE_SYSTEMC=0 build/ARM/libgem5_opt.dylib
cd util/tlm
scons
```

**Terminal Output**
`AttributeError: 'Values' object has no attribute 'duplicate_sources'`.

BobbyRBruce:
I'm not getting this error when I reproduce. I'm running into this though when I try to build "libgem5_opt.so" though:

```sh
build/ARM/cxx_config/Gem5ToTlmBridge64.cc:149:24: error: cannot convert ‘sc_gem5::Gem5ToTlmBridge<64>*’ to ‘gem5::SimObject*’ in return
  149 |     return this->create();
      |            ~~~~~~~~~~~~^~
      |                        |
      |                        sc_gem5::Gem5ToTlmBridge<64>*
scons: *** [build/ARM/cxx_config/Gem5ToTlmBridge32.os] Error 1
scons: *** [build/ARM/cxx_config/Gem5ToTlmBridge64.os] Error 1
```

rogerchang23424:
Hi @BobbyRBruce,

If you are building it on the top of develop branch, steps of build gem5 so have changed to

```bash
scons defconfig build/ARM build_opts/ARM
scons setconfig build/ARM USE_SYSTEMC=n
scons --with-cxx-config --without-python --without-tcmalloc build/ARM/libgem5_opt.so
[...]

PR #1105, opened 2024-05-05
util, ext: Fix building TLM

Lukas-Zenick:
Fixed the issue that did not allow building TLM.

Build commands:
```bash
scons build/ARM/gem5.opt
scons setconfig build/ARM USE_SYSTEMC=n
scons --with-cxx-config --without-python --without-tcmalloc build/ARM/libgem5_opt.so
cd util/tlm
scons
```
Following this README, I tested it succeeds with the simple examples:
https://gem5.googlesource.com/public/gem5/+/master/util/tlm/README

GitHub Issue: #591 
Change-Id: If07fae2eb20ad62627e733573f61bc42d594f970

### DIFF
--- a/util/tlm/SConstruct
+++ b/util/tlm/SConstruct
@@ -44,6 +44,7 @@ env = Environment()
 
 #Make the gem5 root available in SConscripts
 env['GEM5_ROOT'] = gem5_root
+env['GEM5BUILD'] = os.path.join(gem5_root, 'build')
 
 shlibsuffix = env['SHLIBSUFFIX']
 
@@ -52,6 +53,7 @@ env.Append(CPPPATH=[gem5_root + '/build/' + gem5_arch,
                     gem5_root + '/util/systemc/gem5_within_systemc',
                     gem5_root + '/ext/systemc/src',
                     gem5_root + '/ext',
+                    gem5_root + '/src',
                     '#src',
                     '#examples/common',
                     ])
@@ -63,6 +65,12 @@ env.Append(CXXFLAGS=['-std=c++17',
 
 if gem5_variant == 'debug':
     env.Append(CXXFLAGS=['-g', '-DGEM5_DEBUG'])
+
+AddOption('--duplicate-sources', action='store_true', default=False,
+          dest='duplicate_sources',
+          help='Create symlinks to sources in the build directory')
+AddOption('--verbose', action='store_true',
+          help='Print full tool command lines')
 
 deps = [] # keep track of all dependencies required for building the binaries
 
--- a/util/tlm/SConstruct
+++ b/util/tlm/SConstruct
@@ -44,6 +44,8 @@ env = Environment()
 
 #Make the gem5 root available in SConscripts
 env['GEM5_ROOT'] = gem5_root
+#This won't work with kconfig-based builds, where the build directory is
+# not necessarily named "build".
 env['GEM5BUILD'] = os.path.join(gem5_root, 'build')
 
 shlibsuffix = env['SHLIBSUFFIX']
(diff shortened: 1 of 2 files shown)


giactra:
Hi @Lukas-Zenick, first of all thank you so much for the fix. I’d like to point out though the title of the issue you addressed is probably not correct. I believe the compilation error happens for non Arm simulations as well, it’s just that lots of the TLM doc has been written with Arm as an example.
For this reason I would consider changing the tags in the PR and commit title from arch-arm to util,ext

Lukas-Zenick:
@ivanaamit Would it be possible to get these changes merged into develop by tomorrow? I have a project deadline and it's worth more for my grade if it gets merged in. If it still needs more intensive review then no worries.

Harshil2107 (dismissed):
I am not a expert at scons stuff but this lgtm.

util/tlm/SConstruct
```diff
@@ -64,6 +66,12 @@ env.Append(CXXFLAGS=['-std=c++17',
 if gem5_variant == 'debug':
     env.Append(CXXFLAGS=['-g', '-DGEM5_DEBUG'])
 
+AddOption('--duplicate-sources', action='store_true', default=False,
+          dest='duplicate_sources',
+          help='Create symlinks to sources in the build directory')
```
giactra: You don't need this one as the option is already present below (see --no-duplicate-sources). You just need to bump the option over here before it is used

util/tlm/SConstruct
```diff
@@ -44,6 +44,7 @@ env = Environment()
 
 #Make the gem5 root available in SConscripts
 env['GEM5_ROOT'] = gem5_root
+env['GEM5BUILD'] = os.path.join(gem5_root, 'build')
```
giactra, line 49: As a note: this won't work with kconfig based builds where the build directory is not necessarily in "build" and can have different names (see https://www.gem5.org/documentation/general_docs/kconfig_build_system/ ). Anyway I am ok on merging this for now as long as we document this issue

giactra (changes requested):
Hi @Lukas-Zenick we can merge it quickly but there's one actionable item on my side

util/tlm/SConstruct
Lukas-Zenick: What do you mean by "bump the option over here before it is used"? What I was referencing for this code was the main gem5 SConstruct file which has the following on line 145:
```python
AddOption('--duplicate-sources', action='store_true', default=False,
          dest='duplicate_sources',
          help='Create symlinks to sources in the build directory')
AddOption('--no-duplicate-sources', action='store_false',
          dest='duplicate_sources',
          help='Do not create symlinks to sources in the build directory')
```
so I just assumed that you would always need to define both flags

util/tlm/SConstruct
Lukas-Zenick: Gotcha, do I need to do anything specific in terms of documentation aside from having a code comment that specifies that it is not compatible with kconfig builds?

util/tlm/SConstruct
giactra: You don't need to define both options. You just need to move the following code:

```
AddOption('--no-duplicate-sources', action='store_false',
          dest='duplicate_sources',
          help='Do not create symlinks to sources in the build directory')
```

To line 69.

ivanaamit:
Hi @giactra, could you please re-review this PR? I have applied the changes you requested. Let me know if you would like me to create an issue about the code not being compatible with Kconfig builds. Thanks.

giactra:
> Hi @giactra, could you please re-review this PR? I have applied the changes you requested. Let me know if you would like me to create an issue about the code not being compatible with Kconfig builds. Thanks.

Thanks @ivanaamit .
I think the remaining issue is that both commit are wrongly tagged as arch-arm. They should rather be util or ext or util,ext depending on the affected subdirectories

ivanaamit:
I am getting a "permission denied" message when trying to rebase.

@giactra, if you could approve this PR, I will then do a "squash and merge" and change the commit message directly on GitHub.

giactra (approved):

PR #1107, opened 2024-05-06
cpu-minor: Integrate executeStats with int/fp/vec ALU Accesses

poal023:
Although stats for int instructions are already tracked in Minor, might be better to be more consistent and include `executeStats` in tracking statistics, so whenever users are searching for # of int ALU ops, they would find it under execute rather than under the commit stats for Minor. Moreover, this gives a bit more granularity to the stats, as users can see per thread ALU accesses alongside the total.

Perhaps others have some differing opinions about tracking such stats, and prefer leaving `intInstructions` as is?

### DIFF
--- a/src/cpu/minor/execute.cc
+++ b/src/cpu/minor/execute.cc
@@ -876,6 +876,15 @@ Execute::doInstCommitAccounting(MinorDynInstPtr inst)
     cpu.commitStats[inst->id.threadId]->numOps++;
     cpu.commitStats[inst->id.threadId]
         ->committedInstType[inst->staticInst->opClass()]++;
+    if (inst->isInst()) {
+        if (inst->staticInst->isVector()) {
+            cpu.executeStats[inst->id.threadId]->numVecAluAccesses++;
+        } else if (inst->staticInst->isFloating()) {
+            cpu.executeStats[inst->id.threadId]->numFpAluAccesses++;
+        } else if (inst->staticInst->isInteger()) {
+            cpu.executeStats[inst->id.threadId]->numIntAluAccesses++;
+        }
+    }
 
     /* Set the CP SeqNum to the numOps commit number */
     if (inst->traceData)


src/cpu/minor/execute.cc
```diff
@@ -876,6 +876,17 @@ Execute::doInstCommitAccounting(MinorDynInstPtr inst)
     cpu.commitStats[inst->id.threadId]->numOps++;
     cpu.commitStats[inst->id.threadId]
         ->committedInstType[inst->staticInst->opClass()]++;
+     if (inst->isInst())
+    {
+            if (inst->staticInst->isVector())
+            {
+                    cpu.executeStats[inst->id.threadId]->numVecAluAccesses++;
```
mattsinc: It does not look like this code addition conforms to the spacing and formatting rules.  Did you run the commit hook for this?  Did you run the style.py script before pushing?

mattsinc (changes requested):
Besides the formatting and spacing issues, I don't have any big issues with this.  However, I suspect the commit message will need to be updated to be more clear about what this patch is, instead of talking about what folks may or may not want in stats

Thanks again for adding this @poal023 !

src/cpu/minor/execute.cc
poal023: I think I did, but I double-checked and it does look like it's not conforming to the style guide, will fix. Thanks for catching that!

poal023: Update: Adjusted the spaces so they align with the style guide, indentation should conform now!

mattsinc (approved):
@powerjg from my perspective this is ok, but I'm not a CPU/stats expert.  Let me know if there is something I'm missing.

mattsinc (changes requested):
I just noticed you are trying to merge this into stable -- it should be merged into develop instead.  I'm guessing it will be easier to close this pull request and open a new one that does so?

poal023:
See #1108, that should push to develop (and have all the necessary changes). 

Sorry for the silly mistake!

PR #1117, opened 2024-05-08
mem-ruby: Reduce handshaking between CorePair and dir

NSurawar:
Currently when data is downgraded by MOESI_AMD_Base-CorePair (e.g. due to a replacement) this requires a 4-way handshake between the CorePair and the dir. Specifically, the CorePair send a message telling the dir it'd like to downgrade then, the dir sends an ACK back and then, the CorePair writes the data back, and finally, the dir ACKs the writeback.
This is very inefficient and not representative of how modern protocols downgrade a request. Accordingly, this commits updates the downgrade support such that the CorePair writes back the data immediately and then the dir ACKs it.
Thus, this approach requires only a 2-way handshake.

Change-Id: I7ebc85bb03e8ce46a8847e3240fc170120e9fcd6

### DIFF
--- a/src/mem/ruby/protocol/MOESI_AMD_Base-CorePair.sm
+++ b/src/mem/ruby/protocol/MOESI_AMD_Base-CorePair.sm
@@ -730,7 +730,7 @@ machine(MachineType:CorePair, "CP-like Core Coherence")
       out_msg.DataBlk := cache_entry.DataBlk;
       assert(cache_entry.Dirty);
       out_msg.Destination.add(mapAddressToMachine(address, MachineType:Directory));
-      out_msg.MessageSize := MessageSizeType:Request_Control;
+      out_msg.MessageSize := MessageSizeType:Writeback_Data;
       out_msg.Type := CoherenceRequestType:VicDirty;
       out_msg.InitialRequestTime := curCycle();
       if (cache_entry.CacheState == State:O) {
@@ -1109,27 +1109,6 @@ machine(MachineType:CorePair, "CP-like Core Coherence")
         out_msg.Sender := machineID;
         out_msg.Destination.add(mapAddressToMachine(address, MachineType:Directory));
         out_msg.MessageSize := MessageSizeType:Response_Control;
-        DPRINTF(RubySlicc, "%s\n", out_msg);
-      }
-    }
-  }
-
-  action(wb_data, "wb", desc="write back data") {
-    peek(responseToCore_in, ResponseMsg) {
-      enqueue(responseNetwork_out, ResponseMsg, issue_latency) {
-        out_msg.addr := address;
-        out_msg.Type := CoherenceResponseType:CPUData;
-        out_msg.Sender := machineID;
-        out_msg.Destination.add(mapAddressToMachine(address, MachineType:Directory));
-        out_msg.DataBlk := tbe.DataBlk;
-        out_msg.Dirty := tbe.Dirty;
-        if (tbe.Shared) {
-          out_msg.NbReqShared := true;
-        } else {
-          out_msg.NbReqShared := false;
-        }
-        out_msg.State := CoherenceState:Shared; // faux info
-        out_msg.MessageSize := MessageSizeType:Writeback_Data;
         DPRINTF(RubySlicc, "%s\n", out_msg);
       }
     }
@@ -2427,19 +2406,16 @@ machine(MachineType:CorePair, "CP-like Core Coherence")
   }
 
   transition(MO_I, NB_AckWB, I) {L2TagArrayWrite} {
-    wb_data;
     d_deallocateTBE;
     pr_popResponseQueue;
   }
 
   transition(ES_I, NB_AckWB, I) {L2TagArrayWrite} {
-    wb_data;
     d_deallocateTBE;
     pr_popResponseQueue;
   }
 
   transition(MO_S0, NB_AckWB, S0) {L2TagArrayWrite} {
-    wb_data;
     i2_invL2;
     a2_allocateL2;
     d_deallocateTBE; // FOO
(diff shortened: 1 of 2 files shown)


src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
```diff
@@ -1397,22 +1400,38 @@ machine(MachineType:Directory, "AMD Baseline protocol")
...
+    wada_wakeUpAllDependentsAddr;
+    dt_deallocateTBE;
     p_popRequestQueue;
   }
 
   transition(BL, {VicDirty, VicClean}) {
     zz_recycleRequestQueue;
   }
 
+  //CPUData is now deprecated
```
mattsinc: I don't think this is correct -- since right below this you have BL, CPUData, U -- meaning you have two different transitions for CPUData when it's in BL.  So only one of them should be possible, right?

mattsinc (changes requested):
In addition to the changes we discussed earlier, one more.

Otherwise this looks good -- nice job!

abmerop:
FYI, I may not get to this until next week

mattsinc:
> FYI, I may not get to this until next week

@NSurawar has changes to make from my suggestions already, so no worries.

mattsinc:
@NSurawar can you please wrap the changes up this week?

mattsinc (dismissed):
Ok from my end, @abmerop let us know

mattsinc:
@NSurawar : you'll also need to update this branch to the head of develop before this can be committed.

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
```diff
@@ -362,6 +362,9 @@ machine(MachineType:Directory, "AMD Baseline protocol")
         if (in_msg.Type == CoherenceResponseType:CPUPrbResp) {
           trigger(Event:CPUPrbResp, in_msg.addr, entry, tbe);
         } else if (in_msg.Type == CoherenceResponseType:CPUData) {
+          //Deprecated - WriteBack data (wb_data) is now sent with VicDirty itself
+          //Leaving this in for reference
+          assert(0);
```
abmerop: I would just remove this. The error below should tell you which message type is missing

src/mem/ruby/protocol/MOESI_AMD_Base-CorePair.sm
```diff
@@ -2427,19 +2426,16 @@ machine(MachineType:CorePair, "CP-like Core Coherence")
   }
 
   transition(MO_I, NB_AckWB, I) {L2TagArrayWrite} {
-    wb_data;
```
abmerop, line 2430: Remove the wb_data action since these are the only 4 cases it is used

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
```diff
@@ -1397,22 +1400,33 @@ machine(MachineType:Directory, "AMD Baseline protocol")
...
+    wada_wakeUpAllDependentsAddr;
+    dt_deallocateTBE;
     p_popRequestQueue;
   }
 
   transition(BL, {VicDirty, VicClean}) {
     zz_recycleRequestQueue;
   }
 
+  //CPUData is now deprecated
```
abmerop: Should remove the CPUData event, in_port triggers, and all CPUData transitions if that is the case

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
```diff
@@ -1397,22 +1400,33 @@ machine(MachineType:Directory, "AMD Baseline protocol")
     p_popRequestQueue;
   }
 
-  transition(U, VicDirty, BL) {L3TagArrayRead} {
+  transition(U, VicDirty, U) {L3TagArrayRead, L3DataArrayWrite} {
```
abmerop: The 3rd transition `, U` is not necessary here (it will remain in the same state)

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
```diff
@@ -1397,22 +1400,33 @@ machine(MachineType:Directory, "AMD Baseline protocol")
...
+    d_writeDataToMemory;
+    al_allocateL3Block;
+    pr_profileL3HitMiss; //Must come after al_allocateL3Block and before dt_deallocateTBE
+    wada_wakeUpAllDependentsAddr;
+    dt_deallocateTBE;
     p_popRequestQueue;
   }
 
-  transition(U, VicClean, BL) {L3TagArrayRead} {
+  transition(U, VicClean, U) {L3TagArrayRead, L3DataArrayWrite} {
```
abmerop: Also do not need 3rd parameter here

abmerop (commented):
This seems reasonable but I'd recommending going through and removing all of the dead code introduced by this change (CPUData can go, and wd_data action).

mattsinc:
@NSurawar are you able to make @abmerop 's requested changes this weekend?

mattsinc:
> @NSurawar are you able to make @abmerop 's requested changes this weekend?

@NSurawar please let me know -- we are trying to release a new version of gem5 and I will make these changes if needed so this doesn't linger.

NSurawar:
I'm sorry for the delay. I was caught up in some other work.
I'm looking into it now.

NSurawar:
I've removed the stale code now. Please let me know if any additional changes are required.

src/mem/ruby/protocol/MOESI_AMD_Base-CorePair.sm
BobbyRBruce: @NSurawar : has this been addressed?

BobbyRBruce: Never mind, I see it now. Resolving this conversation.

BobbyRBruce (approved):
Looks fine to me but i'll wait for @mattsinc or @abmerop to confirm and squash-merge it.

abmerop (approved):

mattsinc:
Thanks all for getting this across the line while I was away

PR #1118, opened 2024-05-08
mem-ruby: Fix NullPointerException in RubyRequest

ivanaamit:
This PR includes a check for `m_pkt` being null and appropriately handles that case. This issue was causing the Daily tests to fail.

Change-Id: I87142ca14ca4ab3d8306153a1cf34c2629a119ba


ivanaamit:
Hi @giactra, just wanted to let you know that we are pushing the patch to fix the failed Daily tests.

abmerop (approved):

PR #1120, opened 2024-05-09
arch-vega: Fix SDWA dst select

abmerop:
The destination select should take a value of the selection size (dword, word, or byte) starting at bit 0, move that to the selected destination, and then apply the unused constraint (DST_U) to the remaining word or bytes. Currently the code is selecting the word/byte currently being iterated over, rather than the least significant word/byte. As a result, any selection that is not word 0 or byte 0 will be replaced with the original destination value at those bits. This results in the wrong value.

This commit changes the orig bits to be the original dest value at the lowest word / byte location. Tested with the mfma_i32_16x16x16i8 example which uses an SDWA V_OR_B32 to pack i8 values into VGPRs for the MFMA.

Change-Id: I54ed819479a25fa9276d29a8f14f0fea7fd71afe

### DIFF
--- a/src/arch/amdgpu/vega/insts/inst_util.hh
+++ b/src/arch/amdgpu/vega/insts/inst_util.hh
@@ -701,7 +701,7 @@ namespace VegaISA
         if (sel < SDWA_WORD_0) { // we are selecting 1 byte
             // if we sign extended depends on upper-most bit of byte 0
             signExt = (signExt &&
-                       (bits(currDstVal, VegaISA::MSB_PER_WORD, 0) & 0x80));
+                       (bits(currDstVal, VegaISA::MSB_PER_BYTE, 0) & 0x80));
 
             for (int byte = 0; byte < 4; ++byte) {
                 low_bit = byte * VegaISA::BITS_PER_BYTE;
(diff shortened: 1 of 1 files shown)


mattsinc (approved):
I wonder if this changed since GCN3 -- this is my code you're fixing, and I thought I had tested it on non-"start at bit 0" cases at the time, but perhaps not ...

mattsinc:
@mjkpolo , @v-ramadas FYI

abmerop:
> I wonder if this changed since GCN3 -- this is my code you're fixing, and I thought I had tested it on non-"start at bit 0" cases at the time, but perhaps not ...

I don't think so, but the documentation isn't super clear on how some fields are handled, even the latest cdna3 spec.  This is how I interpreted it combined with observations from testing

mattsinc:
> > I wonder if this changed since GCN3 -- this is my code you're fixing, and I thought I had tested it on non-"start at bit 0" cases at the time, but perhaps not ...
> 
> I don't think so, but the documentation isn't super clear on how some fields are handled, even the latest cdna3 spec. This is how I interpreted it combined with observations from testing

I must not have tested it properly then ...

Issue #1126, opened 2024-05-13
kvm full system workloads fail to run on x86


Harshil2107:
**Describe the bug**
This bug was introduced in #1065 
KVM full system workload fail with the following error:
```
build/X86/gem5.opt configs/example/gem5_library/test_script.py 
gem5 Simulator System.  https://www.gem5.org/
gem5 is copyrighted software; use the --copyright option for details.
 
gem5 version DEVELOP-FOR-24.0
gem5 compiled May 10 2024 11:06:09
gem5 started May 10 2024 11:10:58
gem5 executing on COE-CS-sterling, pid 65871
command line: build/X86/gem5.opt configs/example/gem5_library/test_script.py
 
info: Using default config
Resource 'x86-ubuntu-22.04-img' was not found locally. Downloading to '/home/harshilp/.cache/gem5/x86-ubuntu-22.04-img.gz'...
Finished downloading resource 'x86-ubuntu-22.04-img'.
Decompressing resource 'x86-ubuntu-22.04-img' ('/home/harshilp/.cache/gem5/x86-ubuntu-22.04-img.gz')...
Finished decompressing resource 'x86-ubuntu-22.04-img'.
warn: <orphan X86Board>.kvm_vm already has parent not resetting parent.
        Note: kvm_vm is not a parameter of X86Board
warn: (Previously declared as <orphan X86Board>.processor.kvm_vm
Global frequency set at 1000000000000 ticks per second
warn: No dot file generated. Please install pydot to generate the dot file and pdf.
src/mem/dram_interface.cc:690: warn: DRAM device capacity (8192 Mbytes) does not match the address range assigned (4096 Mbytes)
src/sim/kernel_workload.cc:46: info: kernel located at: /home/harshilp/.cache/gem5/x86-linux-kernel-5.4.0-105-generic
src/base/statistics.hh:279: warn: One of the stats is a legacy stat. Legacy stat is a stat that does not belong to any statistics::Group. Legacy stat is deprecated.
      0: board.pc.south_bridge.cmos.rtc: Real-time clock set to Sun Jan  1 00:00:00 2012
board.pc.com_1.device: Listening for connections on port 3456
src/base/statistics.hh:279: warn: One of the stats is a legacy stat. Legacy stat is a stat that does not belong to any statistics::Group. Legacy stat is deprecated.
src/dev/intel_8254_timer.cc:128: warn: Reading current count from inactive timer.
board.remote_gdb: Listening for connections on port 7000
[...]

Issue #1119, opened 2024-05-09
Build ROCm disk image


ivanaamit:
Build ROCm disk image following these instructions: https://github.com/gem5/gem5-resources/pull/29.

Harshil2107:
@abmerop I am trying to run the new gpu diskimage with the example script mi200.py as shows in the readme but it leads to the following error. Does this look like a kvm bug or a config issue by me?

```
./build/VEGA_X86/gem5.opt configs/example/gpufs/mi200.py --disk-image /home/harshilp/gem5-resources-worktrees/gem5-resources-repo/src/x86-ubuntu-gpu-ml/disk-image/x86-ubuntu-gpu-ml --kernel  /home/harshilp/gem5-resources-worktrees/gem5-resources-repo/src/x86-ubuntu-gpu-ml/vmlinux-gpu-ml --app ./pytorch_test.py
gem5 Simulator System.  https://www.gem5.org
gem5 is copyrighted software; use the --copyright option for details.

gem5 version DEVELOP-FOR-24.0
gem5 compiled May 23 2024 13:08:48
gem5 started May 24 2024 14:15:07
gem5 executing on COE-CS-sterling, pid 1157351
command line: ./build/VEGA_X86/gem5.opt configs/example/gpufs/mi200.py --disk-image /home/harshilp/gem5-resources-worktrees/gem5-resources-repo/src/x86-ubuntu-gpu-ml/disk-image/x86-ubuntu-gpu-ml --kernel /home/harshilp/gem5-resources-worktrees/gem5-resources-repo/src/x86-ubuntu-gpu-ml/vmlinux-gpu-ml --app ./pytorch_test.py

warn: Physical memory size specified is 8GB which is greater than 3GB.  Twice the number of memory controllers would be created.
Global frequency set at 1000000000000 ticks per second
warn: system.workload.acpi_description_table_pointer.rsdt adopting orphan SimObject param 'entries'
warn: No dot file generated. Please install pydot to generate the dot file and pdf.
src/mem/dram_interface.cc:690: warn: DRAM device capacity (8192 Mbytes) does not match the address range assigned (4096 Mbytes)
src/sim/kernel_workload.cc:46: info: kernel located at: /home/harshilp/gem5-resources-worktrees/gem5-resources-repo/src/x86-ubuntu-gpu-ml/vmlinux-gpu-ml
[...]

PR #1127, opened 2024-05-13
Revert "cpu-kvm: Support perf counters on hybrid host architectures"

Harshil2107:
Reverts gem5/gem5#1065

Reverting this change because this PR breaks X86 kvm as mentioned in the issue #1126.

### DIFF
--- a/src/cpu/kvm/BaseKvmCPU.py
+++ b/src/cpu/kvm/BaseKvmCPU.py
@@ -72,12 +72,6 @@ class BaseKvmCPU(BaseCPU):
     usePerfOverflow = Param.Bool(
         False, "Use perf event overflow counters (EXPERIMENTAL)"
     )
-    allowHybridPerf = Param.Bool(
-        True,
-        "Enable hybrid performance counters if hybrid host architecture "
-        "detected. Required for accurate stats if gem5 may run on E-core on a "
-        "hybrid host architecture (uncommon).",
-    )
     alwaysSyncTC = Param.Bool(
         False, "Always sync thread contexts on entry/exit"
     )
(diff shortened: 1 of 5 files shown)


Harshil2107:
@nmosier @andysan ping to inform you of this.

andysan:
Wouldn't it be better to just disable this feature by default instead? That way, we can treat the feature as experimental and we don't end up reverting a large amount of code (less disruptive). It's presumably working on some systems when enabled.

powerjg:
Related: #1138

ivanaamit:
@nmosier , @andysan : We will merge this on Monday if there is no new patch that fixes this issue.

ivanaamit (approved):

nmosier:
Sorry for the late reply.

I agree with @andysan — I think it would be better to simply disable hybrid performance counters by default and treat them as an experimental feature rather than just revert the entire patch. Reverting this patch has once again made the KVM CPU non-functional on the E-cores of my Alder Lake system.

Can I just resubmit this patch with [this line](https://github.com/gem5/gem5/blob/6f4ba0b422c2e2e77c5ae37be5c619cc00967cce/src/cpu/kvm/BaseKvmCPU.py#L76) changed from `True` to `False`, disabling hybrid performance counters by default? Unfortunately, I don’t think I’ll have time in the forseeable future to debug issue #1126. However, in the meantime, it would be nice to be able to use the KVM CPU on my Alder Lake E-cores.

powerjg:
I agree that it would be good to have support for hybrid performance counters. However, we were not able to get #1065 to work on our systems (see discussion here: https://github.com/gem5/gem5/issues/1126#issuecomment-2113033296 apologies for the information being spread around).

I am positive that there are some problems with the current KVM support. We are also having issues here: https://github.com/gem5/gem5/issues/1119#issuecomment-2130387680

I'm also wary of adding parameters to *SimObjects* that are required for different hosts. I don't like the idea of tying simulation parameters to a host.

That said, if we can 1) document the problems and which hosts are causing the issue and 2) explain the needed configurations for different hosts, then we can resubmit the patch. However, we need to do a lot more testing this time before committing 🙂

PR #1135, opened 2024-05-14
arch-riscv: add agnostic option to vector tail/mask policy for mem and arith instructions

### LINKED ISSUE #997: arch-riscv: vector tail/mask policy dependencies
It has come to @adriaarmejach's and my attention that the destination register, in some vector instructions, is set as an additional source register as a mean to implement the tail/mask undisturbed policy (which in gem5 is assumed by default, as the agnostic policy allows the same behavior) so as to copy old inactive (i.e. in tail) or masked-off contents when renaming is in place.

The issue with this is that it can potentially create very odd data dependencies which can affect performance. This is very clear with loads that want to write to the whole register (but also other instructions that may not care/overwrite), as they will have to wait for any prior vector instruction that has the same destination.

To solve this we propose to also implement the option of an agnostic policy which can write 1s to the masked-off/inactive elements. This would circumvent the need of using the destination register as source (as there's no need of copying) when both tail and mask are agnostic (i.e. `vtype.vta=1` and `vtype.vma=1`).

saul44203:
These two commits add agnostic capability for both tail/mask policies, for vector memory and arithmetic instructions respectively. The common policy for instructions is to act as undisturbed if one is (i.e. tail or mask), or write all 1s if none.

For those instructions in which multiple micro instructions are instantiated to write to the same register (`VlStride` and `VlIndex` for memory, and `VectorGather`, `VectorSlideUp` and `VectorSlideDown` for arithmetic), a (new) micro instruction named `VPinVdCpyVsMicroInst` has been used to pin the destination register so that there's no need to copy the partial results between them. This idea is similar to what's on ARM's  SVE code.  This micro also implements the tail/mask policy for this cases.

Finally, it's worth noting that while now using an agnostic policy for both tail/mask should remove all dependencies with old destination registers, there's an exception with `VectorSlideUp`. The `vslideup_{vx,vi}` instructions need the elements in the offset to be unchanged. The current implementation overrides the current vta/vma and makes them act as undisturbed, since they require the old destination register anyways. There's a minor issue with this though, as `v{,f}slide1up` variants do not need this, but since they share the same constructor, will act all the same.

Related issue #997.

### DIFF
--- a/src/arch/generic/vec_reg.hh
+++ b/src/arch/generic/vec_reg.hh
@@ -140,8 +140,11 @@ class VecRegContainer
     VecRegContainer() {}
     VecRegContainer(const VecRegContainer &) = default;
 
+    /** Set the container. */
+    void set(uint8_t val) { memset(container.data(), val, SIZE); }
+
     /** Zero the container. */
-    void zero() { memset(container.data(), 0, SIZE); }
+    void zero() { set(0); }
 
     /** Assignment operators. */
     /** @{ */
--- a/src/arch/riscv/insts/vector.cc
+++ b/src/arch/riscv/insts/vector.cc
@@ -444,14 +444,14 @@ VMaskMergeMicroInst::execute(ExecContext* xc,
     uint32_t vlenb = pc_ptr->as<PCState>().vlenb();
     const uint32_t elems_per_vreg = vlenb / elemSize;
     size_t bit_cnt = elems_per_vreg;
+
+    // mask tails are always treated as agnostic: writting 1s
+    tmp_d0.set(0xff);
+
     vreg_t tmp_s;
-    xc->getRegOperand(this, 0, &tmp_s);
-    auto s = tmp_s.as<uint8_t>();
-    // cp the first result and tail
-    memcpy(Vd, s, vlenb);
-    for (uint8_t i = 1; i < this->_numSrcRegs; i++) {
+    for (uint8_t i = 0; i < this->_numSrcRegs; i++) {
         xc->getRegOperand(this, i, &tmp_s);
-        s = tmp_s.as<uint8_t>();
+        auto s = tmp_s.as<uint8_t>();
         if (elems_per_vreg < 8) {
             const uint32_t m = (1 << elems_per_vreg) - 1;
             const uint32_t mask = m << (i * elems_per_vreg % 8);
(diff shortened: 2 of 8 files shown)


ivanaamit:
@hnpl, if you have time, could you please review this PR? Thank you.

hnpl (approved):
The code looks good to me. I agree with the use of pinned writes as it allows parallel writes and avoids register renaming. This is probably more realistic than the current implementation.

I don't have tests at my disposal to test the new implementation of the affected instructions. Please let me know if you've tested those instructions!

saul44203:
Hi. I have at least tested one instruction from each of the arithmetic instruction formats, and on the mem side I've tested them all in theory.

hnpl:
That's great!

saul44203:
Fixed conflicts with vrgather (#1134) and viota (#1137) changes.

aarmejach:
On our long-running simulations we have hit a bug. @saul44203 has identified the issue. I would hold the merge until fixed.

ivanaamit:
@saul44203 , @aarmejach : Since you are still working on this, should we remove it from the goals for this release?

aarmejach:
Yes, we are probably close, but since time is tight I think we should not rush and drop it.

saul44203:
Hi. In theory this PR should be ready.

One bug we fixed was related to fused instructions (e.g. `vfmacc`, `vmadd`, etc)  needing a bypass (similar to the `vslideup` case) to always get the old destination register, as they are "destructive" and use the destination register as another operand.

The other issue was related to the destination vector register counter (` _numTypedDestRegs[VecRegClass]`) not being set correctly. That would (sometimes) produce the `VPinVdCpyVsMicroInst` to go through rename while the free list didn't have enough registers (i.e. skip the blocking), eventually getting to an assertion error.  

Aside from that the `VPinVdCpyVsMicroInst` class has been split into two: `{VPinVd,VCpyVs}MicroInst` respectively, as it seems cleaner and more reusable.

powerjg (approved):
Assuming @saul44203 says this is ready to merge, I'm good with getting it in for v24.0

aarmejach:
We have done a good amount of testing and seen no issues.

aarmejach:
It'd be great to try to merge this one.

We have done a lot of coverage tests on this and seen no issues.

PR #1136, opened 2024-05-15
Adding Spatter

mahyarsamani:
This PR adds source code for C++ implementation of SpatterGen as well as SpatterKernel. SpatterGen uses a PyBindMethod to add kernels to the backend code. This way the process of processing json files could be offloaded to python. In addition it adds standard library components for SpatterGenCore and SpatterGen. These two components follow the same structure as AbstractCore and AbstractProcessor. In addition spatter_kernel.py adds a definition for SpatterKernel in python to make adding kernels to C++ easier. Also it adds utility functions for parsing dictionaries read from json as well as partitioning traces for multicore setups.

### DIFF
--- a/src/cpu/testers/spatter_gen/SpatterGen.py
+++ b/src/cpu/testers/spatter_gen/SpatterGen.py
@@ -1,0 +1,115 @@
+# Copyright (c) 2024 The Regents of The University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from m5.citations import add_citation
+from m5.objects.ClockedObject import ClockedObject
+from m5.params import *
+from m5.proxy import *
+from m5.util.pybind import PyBindMethod
+
+
+class SpatterKernelType(Enum):
+    vals = ["scatter", "gather"]
+
+
+class SpatterProcessingMode(Enum):
+    vals = ["synchronous", "asynchronous"]
+
+
+class SpatterGen(ClockedObject):
+    type = "SpatterGen"
+    cxx_header = "cpu/testers/spatter_gen/spatter_gen.hh"
+    cxx_class = "gem5::SpatterGen"
+
+    system = Param.System(Parent.any, "System this SpatterGen is a part of.")
+
+    processing_mode = Param.SpatterProcessingMode(
+        "How to process kernels accross multiple SpatterGen cores. "
+        "Whether to synchronize on kernel boundaries or not."
+    )
+
+    port = RequestPort("Port to send memory requests.")
+
+    int_regfile_size = Param.Int("Size of the integer register file.")
+    fp_regfile_size = Param.Int("Size of the floating point register file.")
+    request_gen_latency = Param.Cycles(
+        "Number of cycles to spend for creating a request."
+    )
+    request_gen_rate = Param.Int("Number of requests generate per cycle.")
+    request_buffer_entries = Param.Int("Size of the request buffer.")
+    send_rate = Param.Int(
+        "Number of requests to send in parallel."
+        "Emulates the number of dcache ports."
+    )
+
+    cxx_exports = [
+        PyBindMethod("addKernel"),
+        PyBindMethod("proceedPastSyncPoint"),
+    ]
+
+
+add_citation(
+    SpatterGen,
+    """@inproceedings{10.1145/3422575.3422794,
+author = {Lavin, Patrick and Young, Jeffrey and Vuduc, Richard and Riedy,
+Jason and Vose, Aaron and Ernst, Daniel},
+title = {Evaluating Gather and Scatter Performance on CPUs and GPUs},
+year = {2021},
+isbn = {9781450388993},
+publisher = {Association for Computing Machinery},
+address = {New York, NY, USA},
+url = {https://doi.org/10.1145/3422575.3422794},
+doi = {10.1145/3422575.3422794},
+abstract = {This paper describes a new benchmark tool,
+Spatter, for assessing memory system architectures in the context of a
+specific category of indexed accesses known as gather and scatter.
+These types of operations are increasingly used to express sparse and
+irregular data access patterns, and they have widespread utility in many
+modern HPC applications including scientific simulations, data mining and
+analysis computations, and graph processing. However, many traditional
+benchmarking tools like STREAM, STRIDE, and GUPS focus on characterizing
+only uniform stride or fully random accesses despite evidence that modern
+applications use varied sets of more complex access patterns. Spatter is an
+open-source benchmark that provides a tunable and configurable framework to
+benchmark a variety of indexed access patterns, including variations of gather
+/ scatter that are seen in HPC mini-apps evaluated in this work. The design of
+Spatter includes backends for OpenMP and CUDA, and experiments show how it can
+be used to evaluate 1) uniform access patterns for CPU and GPU, 2) prefetching
+regimes for gather / scatter, 3) compiler implementations of vectorization for
+gather / scatter, and 4) trace-driven “proxy patterns” that reflect the
+patterns found in multiple applications. The results from Spatter experiments
+show, for instance, that GPUs typically outperform CPUs for these operations
+in absolute bandwidth but not fraction of peak bandwidth, and that Spatter can
+better represent the performance of some cache-dependent mini-apps than
+traditional STREAM bandwidth measurements.},
+booktitle = {Proceedings of the International Symposium on Memory Systems},
+pages = {209–222},
+numpages = {14},
+location = {Washington, DC, USA},
+series = {MEMSYS '20}
+}
+""",
+)
--- a/src/cpu/testers/spatter_gen/spatter_gen.cc
+++ b/src/cpu/testers/spatter_gen/spatter_gen.cc
@@ -1,0 +1,582 @@
+/*
+* Copyright (c) 2024 The Regents of The University of California
+* All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without
+* modification, are permitted provided that the following conditions are
+* met: redistributions of source code must retain the above copyright
+* notice, this list of conditions and the following disclaimer;
+* redistributions in binary form must reproduce the above copyright
+* notice, this list of conditions and the following disclaimer in the
+* documentation and/or other materials provided with the distribution;
+* neither the name of the copyright holders nor the names of its
+* contributors may be used to endorse or promote products derived from
+* this software without specific prior written permission.
+*
+* THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+* "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+* LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+* A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+* OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+* SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+* LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+* DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+* THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+* OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#include "cpu/testers/spatter_gen/spatter_gen.hh"
+
+#include "base/cprintf.hh"
+#include "debug/SpatterGen.hh"
+#include "debug/SpatterKernel.hh"
+#include "enums/SpatterKernelType.hh"
+#include "enums/SpatterProcessingMode.hh"
+#include "mem/packet.hh"
+#include "sim/sim_exit.hh"
+#include "sim/system.hh"
+
+namespace gem5
+{
+
+using enums::SpatterKernelTypeStrings;
+using enums::SpatterProcessingMode;
+
+SpatterGen::SpatterGen(const Params& params):
+    ClockedObject(params),
+    state(SpatterGenState::RUNNING),
+    requestorId(params.system->getRequestorId(this)),
+    numPendingMemRequests(0),
+    stats(this),
+    mode(params.processing_mode),
+    port(this, name() + ".port"),
+    intRegFileSize(params.int_regfile_size), intRegUsed(0),
+    fpRegFileSize(params.fp_regfile_size), fpRegUsed(0),
+    requestGenLatency(params.request_gen_latency),
+    requestGenRate(params.request_gen_rate),
+    firstGeneratorAvailableTime(0),
+    nextGenEvent([this](){ processNextGenEvent(); }, name() + ".GenEvent"),
+    requestBufferEntries(params.request_buffer_entries),
+    requestBuffer(clockPeriod()),
+    sendRate(params.send_rate),
+    firstPortAvailableTime(0),
+    nextSendEvent([this](){ processNextSendEvent(); }, name() + ".SendEvent"),
+    receiveBuffer(clockPeriod())
+{
+    fatal_if(fpRegFileSize < requestBufferEntries,
+            "fp_regfile_size should be >= request_buffer_entries."
+            "if request_buffer_entries is bigger than fp_regfile_size,"
+            "it may result in inaccuracies in your simulation."
+            "Ideally: fp_regfile_size >> request_buffer_entries."
+    );
+    generatorBusyUntil.resize(requestGenRate, 0);
+    portBusyUntil.resize(sendRate, 0);
+}
+
+Port&
+SpatterGen::getPort(const std::string& if_name, PortID idx)
+{
+    if (if_name == "port") {
+        return port;
+    } else {
+        return ClockedObject::getPort(if_name, idx);
+    }
+}
+
+void
+SpatterGen::startup()
+{
+    scheduleNextGenEvent(curTick());
+}
+
+void
+SpatterGen::SpatterGenPort::sendPacket(PacketPtr pkt)
+{
+    panic_if(blocked(), "Should never try to send if port is blocked.");
+    if (!sendTimingReq(pkt)) {
+        blockedPacket = pkt;
+        DPRINTF(
+            SpatterGen,
+            "%s: Port blocked when sending %s.\n",
+            __func__, pkt->print()
+        );
+    }
+}
+
+void
+SpatterGen::SpatterGenPort::recvReqRetry()
+{
+    DPRINTF(SpatterGen, "%s: Port received a ReqRetry.\n", __func__);
+    panic_if(
+            blockedPacket == nullptr,
+            "Received reqRetry with no blocked packet."
+            );
+    if (!sendTimingReq(blockedPacket)) {
+        DPRINTF(
+            SpatterGen,
+            "%s: Port blocked when sending %s.\n",
+            __func__, blockedPacket->print()
+        );
+    } else {
+        blockedPacket = nullptr;
+        owner->recvReqRetry();
+    }
+}
+
+void
+SpatterGen::recvReqRetry()
+{
+    if (nextSendEvent.pending()) {
+        nextSendEvent.wake();
+        scheduleNextSendEvent(nextCycle());
+    }
+}
+
+bool
+SpatterGen::SpatterGenPort::recvTimingResp(PacketPtr pkt) {
+    return owner->recvTimingResp(pkt);
+}
+
+bool
+SpatterGen::recvTimingResp(PacketPtr pkt)
+{
+    DPRINTF(SpatterGen, "%s: Received pkt: %s.\n", __func__, pkt->print());
+    assert(pkt->isResponse());
+
+    // record trip time.
+    SpatterAccess* spatter_access = pkt->findNextSenderState<SpatterAccess>();
+    Tick trip_time = (curTick() - requestDepartureTime[pkt->req]);
+    requestDepartureTime.erase(pkt->req);
+    spatter_access->recordTripTime(trip_time);
+
+    int trips_left = spatter_access->tripsLeft();
+    assert(trips_left >= 0);
+    if (trips_left > 0) {
+        stats.numIndexReads++;
+        stats.indexBytesRead += pkt->getSize();
+        stats.totalIndexReadLatency += trip_time;
+
+        stats.indexAccessLatency.sample(trip_time);
+        receiveBuffer.push(spatter_access, curTick());
+    } else {
+        stats.valueAccessLatency.sample(trip_time);
+        stats.totalIndirectAccessLatency.sample(
+                                            spatter_access->tripTimeSoFar()
+                                            );
+        if (spatter_access->type() == SpatterKernelType::gather) {
+            stats.numValueReads++;
+            stats.valueBytesRead += pkt->getSize();
+            stats.totalValueReadLatency += trip_time;
+        } else if (spatter_access->type() == SpatterKernelType::scatter) {
+            stats.numValueWrites++;
+            stats.valueBytesWritten += pkt->getSize();
+            stats.totalValueWriteLatency += trip_time;
+        } else {
+            panic("Unknown kernel type.");
+        }
+        // CAUTION: We're going to decrement fpRegUsed here,
+        // it could cause inaccuracies if processNextGenEvent
+        // is called after recvTimingResp on the same tick.
+        // i.e. we might end up releasing a register on the same
+        // cycle that we are allocating it.
+        // it's probably not going to ever be an issue since
+        // fpRegFileSize is probably >> requestBufferEntries
+        // i.e. the chances of running out of fp registers is low because
+        // we do not simulate parts of the pipeline that back things up into
+        // fp registers, e.g. functional units of ALU.
+        fpRegUsed--;
+        delete spatter_access;
+    }
+
+    // delete the pkt since we don't need it anymore.
+    delete pkt;
+
+    if (!nextGenEvent.pending()) {
+        scheduleNextGenEvent(nextCycle());
+    }
+
+    numPendingMemRequests--;
+    checkForSimExit();
+    return true;
+}
+
+void
+SpatterGen::addKernel(
+    uint32_t id, uint32_t delta, uint32_t count,
+    SpatterKernelType type,
+    size_t index_size, Addr base_index_addr,
+    size_t value_size, Addr base_value_addr,
+    const std::vector<uint32_t>& indices
+)
+{
+    DPRINTF(
+        SpatterGen,
+        "%s: Adding kernel with id: %d, delta: %d, count: %d, type: %s.\n",
+        __func__, id, delta, count, SpatterKernelTypeStrings[type]
+    );
+    SpatterKernel new_kernel(
+                            requestorId,
+                            id, delta, count, type,
+                            index_size, base_index_addr,
+                            value_size, base_value_addr
+                            );
+    new_kernel.setIndices(indices);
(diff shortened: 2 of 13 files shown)


powerjg:
I know this is a draft, but a few quick comments:
1. You need to add copyright blocks
2. The names should be something more descriptive than "model"
3. The commit message and the PR description need to have details

src/base/stats/units.hh
```diff
@@ -174,6 +174,23 @@ class Second : public Base
     std::string getUnitString() const override { return Second::toString(); }
 };
 
+class NanoSecond : public Base
```
mahyarsamani: @powerjg If I want to express latency stats in seconds, due to precision limitations they will all be zero. I added this new unit to express the latencies in nanoseconds. I am not really sure if this is what the right way to do this is. Let me know if I have to change this to Tick.

powerjg: Yeah, I'm against this. Just use `Ticks` if you need high precision.

src/cpu/testers/spatter_gen/SpatterGen.py
```diff
@@ -0,0 +1,55 @@
...
+    int_regfile_size = Param.Int("Size of the integer register file.")
+    fp_regfile_size = Param.Int("Size of the floating point register file.")
+    request_gen_latency = Param.Int(
+        "Number of cycles to spend for creating a request."
+    )
+    request_gen_bandwidth = Param.Int("Number of requests generate per cycle.")
+    request_buffer_size = Param.Int("Size of the request buffer.")
+    send_bandwidth = Param.Int("Number of requests to send in parallel.")
+
+    cxx_exports = [PyBindMethod("addKernel")]
```
powerjg: Can you add a reference to the spatter paper for this class? Call `add_citation(SpatterGen, <bibtex text>)` at the end of this file.

src/cpu/testers/spatter_gen/SpatterGen.py
```diff
@@ -0,0 +1,55 @@
...
+    cxx_header = "cpu/testers/spatter_gen/spatter_gen.hh"
+    cxx_class = "gem5::SpatterGen"
+
+    system = Param.System(Parent.any, "System this SpatterGen is a part of.")
+
+    port = RequestPort("Port to send memory requests.")
+
+    int_regfile_size = Param.Int("Size of the integer register file.")
+    fp_regfile_size = Param.Int("Size of the floating point register file.")
+    request_gen_latency = Param.Int(
```
powerjg: Should this be a `Param.Cycles` if it's cycles?

src/cpu/testers/spatter_gen/SpatterGen.py
```diff
@@ -0,0 +1,55 @@
...
+        "Number of cycles to spend for creating a request."
+    )
+    request_gen_bandwidth = Param.Int("Number of requests generate per cycle.")
```
powerjg:
```suggestion
    request_gen_rate = Param.Int("Number of requests generate per cycle.")
```
Bandwidth isn't exactly the best term here.

src/cpu/testers/spatter_gen/SpatterGen.py
```diff
@@ -0,0 +1,55 @@
...
+    request_buffer_size = Param.Int("Size of the request buffer.")
```
powerjg:
```suggestion
    request_buffer_entries = Param.Int("Size of the request buffer.")
```
To me, `size` implies it is in `Bytes`

src/cpu/testers/spatter_gen/SpatterGen.py
```diff
@@ -0,0 +1,55 @@
...
+    send_bandwidth = Param.Int("Number of requests to send in parallel.")
```
powerjg: I don't understand how this is different from `request_gen_bandwidth`. The name and/or help text needs to be updated to make it more clear.

src/cpu/testers/spatter_gen/enums.cc
```diff
@@ -0,0 +1,44 @@
...
+    "INDEX",
+    "VALUE"
+};
+
+const char* sleepyStateStrings[NUM_SLEEPY_STATE] = {
+    "AWAKE",
+    "ASLEEP"
+};
+
+} // namespace gem5
```
powerjg: You shouldn't need this file at all. See the code here for how it works: https://github.com/gem5/gem5/blob/stable/src/base/stl_helpers/ostream_helpers.hh#L81

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,227 @@
...
+namespace gem5
+{
+
+using enums::SpatterKernelType;
+
+class SpatterGen: public ClockedObject
+{
+  private:
+
+    class SleepyEvent : public EventFunctionWrapper
```
powerjg: Can you use a less generic name, like "SpatterGenEvent"? I think I see how you want this to be generic, but I don't think it makes sense to make it generic here and now 🙂

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,227 @@
...
+    {
+      // TODO: split pending into pendingInput and pendingOutput
+      private:
+        SleepyState _state;
```
powerjg:
```suggestion
        SpatterGen::State _state;
```

I would make this private to the `SpatterGen` class.

src/cpu/testers/spatter_gen/enums.hh
```diff
@@ -0,0 +1,53 @@
...
+};
+extern const char* accessTypeStrings[NUM_ACCESS_TYPE];
+
+enum SleepyState
+{
+    AWAKE,
+    ASLEEP,
+    NUM_SLEEPY_STATE
+};
+extern const char* sleepyStateStrings[NUM_SLEEPY_STATE];
```
powerjg: I would make this private to the `SpatterGen` class

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,227 @@
...
+    Tick firstGeneratorAvailableTime;
+    // tracking the busy state of our so called "AGU"s.
+    std::vector<Tick> generatorBusyUntil;
+    SleepyEvent nextGenEvent;
+    void processNextGenEvent();
+    // put requests to the cache in the request buffer.
+    int requestBufferSize;
+    std::queue<PacketPtr> requestBuffer;
+    // same explanation as receiveBufferInsertionTime
+    std::queue<Tick> requestBufferInsertionTime;
```
powerjg: I don't understand this comment

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,227 @@
...
+    // if nextGenEvent has to be schedule at tick when then schedule it.
+    // this function should only be called when nextGenEvent is not pending.
+    void scheduleNextGenEvent(Tick when);
+
+    // bandwidth to issue memory requests to cache,
+    // this is supposed to model the number of cache ports
+    // we will assume it takes 1 cycle to issue memory requests
+    int sendBandwidth;
+    Tick firstPortAvailableTime;
+    std::vector<Tick> portBusyUntil;
```
powerjg, line 213: Why is this a vector?

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,227 @@
...
+    void processNextSendEvent();
+    // if nextSendEvent has to be schedule at tick when then schedule it.
+    // this function should only be called when nextSendEvent is not pending.
+    void scheduleNextSendEvent(Tick when);
+
+    // put the memory responses here.
+    // no need to limit the size of this buffer.
+    // it's a response buffer and it will automatically
+    // be limited by requestBufferSize, intRegFileSize, fpRegFileSize
+    std::queue<PacketPtr> receiveBuffer;
```
powerjg: I'm concerned these are queues. Do we have guarantees on ordering?

It also seems like there are pairs of `PacketPtr` and `Tick`. Why not use one datastructure?

src/python/gem5/components/processors/spatter_generator.py
```diff
@@ -0,0 +1,103 @@
...
+                send_bandwidth,
+            )
+        )
+
+        if not clk_freq is None:
+            clock_domain = SrcClockDomain(
+                clock=clk_freq, voltage_domain=VoltageDomain()
+            )
+            for generator in self.cores:
+                generator.clk_domain = clock_domain
```
powerjg: It would be good to add a comment here about why you don't need an `else` statement.

src/python/gem5/components/processors/spatter_generator_core.py
```diff
@@ -0,0 +1,70 @@
...
+    @overrides(AbstractCore)
+    def connect_dcache(self, port: Port) -> None:
+        self.generator.port = port
+
+    def add_kernel(self, kernel: SpatterKernel) -> None:
+        self._kernels.append(kernel)
+
+    def start_traffic(self) -> None:
+        for kernel in self._kernels:
+            print(f"Adding kernel: {kernel}")
```
powerjg: Probably should drop this print for upstream.

src/python/gem5/components/processors/spatter_kernel.py
```diff
@@ -0,0 +1,103 @@
...
+    for i in range(num_leaves):
+        lower_bound = i * interleave_size
+        upper_bound = min(lower_bound + interleave_size, len(original_trace))
+        partitions[i % num_partitions] += original_trace[
+            lower_bound:upper_bound
+        ]
+    return partitions
+
+
+class SpatterKernel:
```
powerjg: This class should really come with some help comments explaining how it's used. Pointers to spatter documentation is great.

src/python/gem5/components/processors/spatter_generator.py
```diff
@@ -0,0 +1,103 @@
+# Copyright (c) 2024 The Regents of the University of California
```
powerjg: I would put these files in a subdirectory. You can expose them in the `processors` module via `__init__.py` if you want, but this directory is getting out of hand.

src/cpu/testers/spatter_gen/spatter_gen.cc
```diff
@@ -0,0 +1,584 @@
...
+    bool no_index = receiveBuffer.empty();
+    bool no_kernel = kernels.empty();
+    if (no_pending_mem && no_queued_req && no_index && no_kernel)
+    {
+        exitSimLoop(
+            csprintf("SpatterGen %s done generating requests.", name()),
+            0,
+            nextCycle()
+        );
+    }
```
powerjg: Is this how the other generators work? I thought if they had nothing to do, they would just not schedule any events and you would get a `MaxTick` exit. I would like all generators to have the same behavior.

src/cpu/testers/spatter_gen/spatter_gen.cc
```diff
@@ -0,0 +1,584 @@
...
+}
+
+void
+SpatterGen::SpatterGenPort::sendPacket(PacketPtr pkt) {
+    panic_if(blocked(), "Should never try to send if port is blocked.");
```
powerjg:
```suggestion
void
SpatterGen::SpatterGenPort::sendPacket(PacketPtr pkt)
{
    panic_if(blocked(), "Should never try to send if port is blocked.");
```

src/cpu/testers/spatter_gen/spatter_gen.cc
```diff
@@ -0,0 +1,584 @@
...
+        stats.indexerAccessLatency.sample(trip_time);
+        // TODO: record travel time for indexer accesses.
+        receiveBuffer.push(pkt);
+        receiveBufferInsertionTime.push(curTick());
+        // REMEMBER: might need to wake nextGenEvent up?
+        // IDEA: Separate pending to pendingInput and pendingOutput?
+        // I think for now we should not wake up nextGenEvent,
+        // because nextGenEvent does not sleep if there are not
+        // enough registers available, which is something that
+        // this scope of the if does not influence.
```
powerjg: Is this a future todo?

src/cpu/testers/spatter_gen/spatter_gen.cc
```diff
@@ -0,0 +1,584 @@
...
+            stats.totalValuesWriteLatency += trip_time;
+        } else {
+            panic("Unknown access type.");
+        }
+        // CAUTION: We're going to decrement fpRegUsed here,
+        // it could cause inaccuracies if processNextGenEvent
+        // is called after recvTimingResp.
+        // it may be fine though since we don't simulate the rest,
+        // it's probably not going to ever be an issue since
+        // fpRegFileSize is probably >> requestBufferSize
```
powerjg: Would it make sense to put a panic/assert here?

src/cpu/testers/spatter_gen/spatter_gen.cc
```diff
@@ -0,0 +1,584 @@
...
+            if (front.done()) {
+                exitSimLoop(
+                    csprintf(
+                    "SpatterGen %s done generating "
+                    "requests from SpatterKernel %d.",
+                    name(), front.id()
+                    ),
+                    0,
+                    nextCycle()
+                );
```
powerjg: Why is this not done in `checkSimDone`?

src/cpu/testers/spatter_gen/spatter_gen.cc
```diff
@@ -0,0 +1,584 @@
...
+                "%s: Pushed pkt: %s to requestBuffer. Access to VALUES.\n",
+                __func__, pkt->print()
+            );
+            // now deallocate resources for reading the index
+            delete idx_info;
+            delete idx_pkt;
+            int_reg_used_delta--;
+            receiveBuffer.pop();
+            receiveBufferInsertionTime.pop();
+        } else if (have_kernel && have_int_reg) {
```
powerjg: I would split these two `if`/`else if` clauses into their own functions. This function is too big.

powerjg (commented):
Overall, this looks great. I didn't check the logic carefully in `spatter_gen.cc`, but I trust you've tested this and compared with upstream spatter.

Most of my comments are style. The only "big" thing that I would like to see changed is moving the state information into the `SpatterGen` class to make it private. 

One thing to keep in mind when pushing code upstream is that you should try to make things as self-contained as possible. Don't expose things publicly unless absolutely required.

Finally, there are some places where you need to add more documentation. Think about how someone would discover how to use this if they don't know anything about Spatter and they are not involved in our project.

src/cpu/testers/spatter_gen/spatter_gen.cc
mahyarsamani: checkSimDone is for overall exit. This exit is per kernel. We're not completely done here. I think it is useful to try dump stats per kernel.

src/cpu/testers/spatter_gen/spatter_gen.cc
mahyarsamani: I agree. I would argue for not changing it right now. I will update this after we have a more definitive picture of what we want the SpatterKernel class to look like.

src/cpu/testers/spatter_gen/spatter_gen.cc
mahyarsamani: I will put a panic telling user to make fpRegFileSize bigger than request_buffer_entries.

src/cpu/testers/spatter_gen/spatter_gen.hh
mahyarsamani: Answer to second question is, it's a preference.
About your first question, do you mean that responses do not arrive at the same order requests were issued? If so, I can ensure that all responses before an expected response have arrived before another access is made based on the resp.

src/cpu/testers/spatter_gen/spatter_gen.hh
mahyarsamani: I need to access it with index.

mahyarsamani:
@powerjg @giactra 
Still working on addressing some comments. Will fix by tomorrow.

src/cpu/testers/spatter_gen/spatter_gen.cc
mahyarsamani: Yes

src/python/gem5/components/processors/spatter_gen/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .spatter_generator import SpatterGenerator
```
powerjg, line 28: Copyright...

src/cpu/testers/spatter_gen/spatter_gen.cc
```diff
@@ -0,0 +1,595 @@
...
+    panic_if(fpRegFileSize < requestBufferEntries,
+            "fp_regfile_size should be >= request_buffer_entries."
+            "if request_buffer_entries is bigger than fp_regfile_size,"
+            "it may result in inaccuracies in your simulation."
+            "Ideally: fp_regfile_size >> request_buffer_entries."
+    );
+    generatorBusyUntil.resize(requestGenRate);
+    for (int i = 0; i < requestGenRate; i++) {
+        generatorBusyUntil[i] = 0;
+    }
```
powerjg: You don't need these for loops. You can set the value to use when calling `resize`. See https://en.cppreference.com/w/cpp/container/vector/resize

src/cpu/testers/spatter_gen/spatter_gen.cc
powerjg: What do you think about this?

src/cpu/testers/spatter_gen/spatter_gen.cc
powerjg: I would argue that if we don't know what we want the kernel to look like, then it's not ready to be merged upstream. We should go ahead and do it the right way the first time. Once something is merged, the activation energy to change it is very high.

src/cpu/testers/spatter_gen/spatter_gen.hh
powerjg: Don't forget about this.

src/python/gem5/components/processors/spatter_gen/spatter_kernel.py
```diff
@@ -0,0 +1,181 @@
...
+            lower_bound:upper_bound
+        ]
+    return partitions
+
+
+class SpatterKernel:
+    """This class encapsulates one kernel in a spatter trace.
+    A spatter trace is represented with a json file.
+    An example of a spatter trace can be found here:
+        bit.ly/3X2IijM
```
powerjg: Can you point to the repo instead? I don't trust that bitly will be around in 20 years.

src/cpu/testers/spatter_gen/spatter_gen.cc
mahyarsamani: You are right, this behavior is different. However, I remember having much trouble with using bandwidth stats with that behavior.

src/cpu/testers/spatter_gen/spatter_gen.cc
```diff
@@ -0,0 +1,591 @@
...
+    requestGenLatency(params.request_gen_latency),
+    requestGenRate(params.request_gen_rate),
+    firstGeneratorAvailableTime(0),
+    nextGenEvent([this](){ processNextGenEvent(); }, name() + ".GenEvent"),
+    requestBufferEntries(params.request_buffer_entries),
+    sendRate(params.send_rate),
+    firstPortAvailableTime(0),
+    nextSendEvent([this](){ processNextSendEvent(); }, name() + ".SendEvent")
+{
+    panic_if(fpRegFileSize < requestBufferEntries,
```
powerjg: `fatal`? If this is a user configuration error it should be fatal. Panic is for bugs (like an assert)

src/cpu/testers/spatter_gen/spatter_gen.cc
```diff
@@ -0,0 +1,591 @@
...
+            "%s: Port blocked when sending %s.\n",
+            __func__, pkt->print()
+        );
+    }
+}
+
+void
+SpatterGen::SpatterGenPort::recvReqRetry()
+{
+    DPRINTF(SpatterGen, "%s: Port received a ReqRetry.\n", __func__);
```
powerjg, line 110: I would add an `assert(blockedPacket);` here

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,239 @@
...
+#include "sim/clocked_object.hh"
+#include "sim/eventq.hh"
+
+namespace gem5
+{
+
+typedef std::tuple<PacketPtr, Tick> TimedPacket;
+
+PacketPtr entryPacket(const TimedPacket& timed_pkt);
+Tick entryTick(const TimedPacket& timed_pkt);
```
powerjg: 1. Where are these defined?
2. Why are they not in the `SpatterGen` class? Do you have any uses `TimedPacket` outside of the class members?

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,239 @@
...
+    {
+      // TODO: split pending into pendingInput and pendingOutput
+      private:
+        enum SleepState
+        {
+            AWAKE,
+            ASLEEP
+        };
+
+        SleepState _state;
```
powerjg: Seems like this could just be a boolean `isPending` (or `isWaiting` or something like that)

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,239 @@
...
+
+      public:
+        SpatterGenEvent(const std::function<void(void)> &callback,
+                    const std::string &name):
+            EventFunctionWrapper(callback, name), _state(SleepState::AWAKE)
+        {}
+        // a SpatterGenEvent will only be asleep if it is pending output
+        bool pending() { return _state == SleepState::ASLEEP; }
+        void sleep() { _state = SleepState::ASLEEP; }
```
powerjg: What is the difference between `pending()` and `sleep()`?

powerjg (commented):
A few more little things

src/cpu/testers/spatter_gen/spatter_gen.hh
mahyarsamani: pending is a getter and sleep is a setter.

src/cpu/testers/spatter_gen/spatter_gen.hh
mahyarsamani: I like to keep it an enum, I know right now the state is a boolean like value. In future I want to extend this so that it can encode more states, e.g. ASLEEP_ON_INPUT, ASLEEP_ON_OUTPUT, ...

src/cpu/testers/spatter_gen/spatter_gen.hh
mahyarsamani: They are defined in spatter_gen.cc. It's a matter of preference for me to not define it in the SpatterGen class. I can move it inside if there is a right/wrong.

src/cpu/testers/spatter_gen/spatter_gen.cc
mahyarsamani: Will do

src/cpu/testers/spatter_gen/spatter_gen.cc
mahyarsamani: Will do

src/cpu/testers/spatter_gen/spatter_gen.hh
powerjg: They are polluting the gem5 namespace is the problem. I would put them in the class until they are used outside of it.

mahyarsamani: I see, will do.

powerjg (dismissed):
Looks good to me. Thanks for the updates!

mahyarsamani:
@giactra This is ready for review.

powerjg (changes requested):
There's a merge/rebase issue here.

mahyarsamani:
> There's a merge/rebase issue here.

I think it should be fixed now.

powerjg (dismissed):

powerjg:
@giactra let me know if you want us to hold off merging until you have time to review.

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,224 @@
...
+#include "sim/clocked_object.hh"
+#include "sim/eventq.hh"
+
+namespace gem5
+{
+
+using enums::SpatterKernelType;
+using enums::SpatterProcessingMode;
+
+class SpatterGen: public ClockedObject
```
giactra, line 78: I appreciate there is a reference to the publication in the SpatterGen.py file, but I believe it would be beneficial for a user to have a brief description of Spatter classes and what they are actually doing, so that a user is not "forced" to read the paper, especially if the paper requires a subscription

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,224 @@
...
+#include "mem/port.hh"
+#include "params/SpatterGen.hh"
+#include "sim/clocked_object.hh"
+#include "sim/eventq.hh"
+
+namespace gem5
+{
+
+using enums::SpatterKernelType;
+using enums::SpatterProcessingMode;
```
giactra: Is there a reason to make them available like this in the gem5 namespace? Shall we restrict the visibility to the Spatter class?

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,224 @@
...
+
+class SpatterGen: public ClockedObject
+{
+  private:
+    class SpatterGenEvent : public EventFunctionWrapper
+    {
+      private:
+        // TODO: split pending into pendingInput and pendingOutput
+        enum SleepState
```
giactra: I would use scoped enum here

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,224 @@
...
+
+        SleepState _state;
+
+      public:
+        SpatterGenEvent(const std::function<void(void)> &callback,
+                    const std::string &name):
+            EventFunctionWrapper(callback, name), _state(SleepState::AWAKE)
+        {}
+        // a SpatterGenEvent will only be asleep if it is pending output
+        bool pending() { return _state == SleepState::ASLEEP; }
```
giactra: Nitpick: this could be marked as const

src/cpu/testers/spatter_gen/spatter_gen.hh
```diff
@@ -0,0 +1,224 @@
...
+        statistics::Histogram indexAccessLatency;
+        statistics::Histogram valueAccessLatency;
+        statistics::Histogram totalIndirectAccessLatency;
+
+        virtual void regStats() override;
+
+        SpatterGenStats(SpatterGen* spatter_gen);
+    };
+
+    enum SpatterGenState
```
giactra: Ditto

src/cpu/testers/spatter_gen/utility_structs.hh
```diff
@@ -0,0 +1,240 @@
...
+
+#include "base/random.hh"
+#include "base/types.hh"
+#include "enums/SpatterKernelType.hh"
+#include "mem/packet.hh"
+
+namespace gem5
+{
+
+using enums::SpatterKernelType;
```
giactra: Ditto

src/cpu/testers/spatter_gen/utility_structs.hh
```diff
@@ -0,0 +1,240 @@
...
+    bool hasReady(Tick current_time)
+    {
+        if (empty()) {
+            return false;
+        }
+        return (current_time - insertionTimes.front()) >= latency;
+    }
+};
+
+typedef std::tuple<Addr, size_t> AccessPair;
```
giactra: I would move this within the SpatterAccess class

src/cpu/testers/spatter_gen/utility_structs.hh
```diff
@@ -0,0 +1,240 @@
...
+* SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+* LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+* DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+* THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+* OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#ifndef __CPU_TESTERS_SPATTER_GEN_SPATTER_KERNEL_HH__
+#define __CPU_TESTERS_SPATTER_GEN_SPATTER_KERNEL_HH__
```
giactra: This is not matching the header file

src/cpu/testers/spatter_gen/utility_structs.hh
```diff
@@ -0,0 +1,240 @@
...
+        items.pop();
+        insertionTimes.pop();
+        return item;
+    }
+
+    T front() { return items.front(); }
+
+    bool empty() { return items.empty(); }
+
+    size_t size() { return items.size(); }
```
giactra: Could be marked as const

src/cpu/testers/spatter_gen/utility_structs.hh
```diff
@@ -0,0 +1,240 @@
...
+{
+    RequestorID requestorId;
+    SpatterKernelType _kernelType;
+    Tick accTripTime;
+    std::queue<AccessPair> accessPairs;
+
+    SpatterAccess(
+        RequestorID requestor_id,
+        SpatterKernelType kernel_type,
+        std::queue<AccessPair> access_pairs
```
giactra: I would pass by const reference here as you are making a copy

src/cpu/testers/spatter_gen/utility_structs.hh
```diff
@@ -0,0 +1,240 @@
...
+    ):
+        requestorId(requestor_id), _kernelType(kernel_type),
+        accTripTime(0), accessPairs(access_pairs)
+    {}
+
+    SpatterKernelType type() { return _kernelType; }
+
+    int tripsLeft() { return accessPairs.size(); }
```
giactra: could be const

src/cpu/testers/spatter_gen/utility_structs.hh
```diff
@@ -0,0 +1,240 @@
...
+
+    void recordTripTime(Tick trip_time) { accTripTime += trip_time; }
+
+    Tick tripTimeSoFar() { return accTripTime; }
```
giactra: could be const

src/cpu/testers/spatter_gen/utility_structs.hh
```diff
@@ -0,0 +1,240 @@
...
+
+    void setIndices(std::vector<uint32_t> pattern)
+    {
+        indices.assign(pattern.begin(), pattern.end());
+        remRolls = indices.size();
+    }
+
+    SpatterKernelType type() { return _type; }
+
+    bool done() { return iteration == count; }
```
giactra: Could be const

src/cpu/testers/spatter_gen/utility_structs.hh
```diff
@@ -0,0 +1,240 @@
...
+    ):
+        requestorId(requestor_id),
+        _id(id), delta(delta), count(count),
+        _type(type),
+        indexSize(index_size), baseIndexAddr(base_index_addr),
+        valueSize(value_size), baseValueAddr(base_value_addr),
+        index(0), iteration(0), remRolls(0)
+    {}
+
+    uint32_t id() { return _id; }
```
giactra: Could be const

giactra (dismissed):
Thanks @powerjg, I am not going to do a full functionality check assuming it has been tested and works properly already. I only wanted to make sure the code was properly documented (that's the main comment I have if you see below). Also some minor comments on the code from a SW engineering point of view

src/cpu/testers/spatter_gen/spatter_gen.hh
mahyarsamani: Would the documentation in spatter_kernel.py suffice? I am now looking at writing it up and wonder if I should extend what is already there.

BobbyRBruce:
@powerjg : Can you sign off on this so we can get it in for the v24.0 staging?

BobbyRBruce (approved):

PR #1138, opened 2024-05-15
arch-x86: Improve KVM set XCR

abmerop:
This adds two failsafes which may cause a panic on some machines. First, check the host machine has the KVM XCR capability before calling getXCRs or setXCRs. Second, ensure the x87 bit, which must always be one, will always return at least one by modifying the return value in readMiscReg.

Change-Id: I5e778acc926a47443ef6cef29fabd84eb69bb9ba

### DIFF
--- a/src/arch/x86/isa.cc
+++ b/src/arch/x86/isa.cc
@@ -244,6 +244,10 @@ ISA::readMiscReg(RegIndex idx)
         return base;
     }
 
+    if (idx == misc_reg::Xcr0) {
+        return regVal[idx] | 1;
+    }
+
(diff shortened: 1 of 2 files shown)


mattsinc (approved):
@v-ramadas : FYI

powerjg (dismissed):
I am blocking this change until we reconcile it with https://github.com/gem5/gem5/pull/1127

There have been multiple KVM changes and some combination has broken KVM on some machines. We need to sort this out before we make more changes to KVM.

If this fixes #1126, that would be great. Please let us know.

cc: @andysan @nmosier @Harshil2107

Issue #555, opened 2023-11-13
Compiling gem5 with GCC Version 9 appears to nearly double object file size


BobbyRBruce:
**Describe the bug**

When compiling gem5 with GCC Version 9, the object files appear to be nearly double in size compared to other compilers and GCC versions. This results in a linking process which causes out-of-memory errors in systems with <=16GB of memory.

When using a GCC-9:

```shell
> du -su build/*

12G	build/ALL

> du -su build/ALL/*

2.6G	build/ALL/arch
65M	build/ALL/base
56K	build/ALL/config
455M	build/ALL/cpu
29M	build/ALL/debug
1.1G	build/ALL/dev
288M	build/ALL/enums
7.5M	build/ALL/ext
11M	build/ALL/gem5.build
1.3G	build/ALL/gem5.opt
724K	build/ALL/gem5py
5.6M	build/ALL/gem5py_m5
8.2M	build/ALL/kern
29M	build/ALL/learning_gem5
670M	build/ALL/mem
2.2M	build/ALL/params
4.3M	build/ALL/proto
4.7G	build/ALL/python
200M	build/ALL/sim
6.6M	build/ALL/sst
117M	build/ALL/systemc
```

Compared to GCC-10:

```shell
> du -su build/*

7.2G	build/ALL

> du -su build/ALL/*

1.3G	build/ALL/arch
43M	build/ALL/base
56K	build/ALL/config
330M	build/ALL/cpu
12M	build/ALL/debug
474M	build/ALL/dev
195M	build/ALL/enums
7.4M	build/ALL/ext
11M	build/ALL/gem5.build
1.2G	build/ALL/gem5.opt
744K	build/ALL/gem5py
5.8M	build/ALL/gem5py_m5
4.5M	build/ALL/kern
22M	build/ALL/learning_gem5
484M	build/ALL/mem
2.2M	build/ALL/params
3.0M	build/ALL/proto
3.0G	build/ALL/python
144M	build/ALL/sim
5.0M	build/ALL/sst
82M	build/ALL/systemc
```

**To Reproduce**

```shell
cd gem5
docker run -u $UID:$GID -v $(pwd):/work -w /work ghcr.io/gem5/gcc-version-9:v23-0 /usr/bin/env python3 /usr/bin/scons --ignore-style build/ALL/gem5.opt -j$(nproc)
```

We found on systems with <=16GB of memory, the above will fail with an out-of-memory error during the linking process

**Host Operating System**

[...]

PR #1145, opened 2024-05-20
misc: Remove gcc 8 support, gem5 support GCC >= v10

BobbyRBruce:
note: Due to #556 / #555, we don't support GCC 9. This PR removes gcc-8 which means gem5 would support GCC >= version 10.

The reason for removing gcc-8:

1. We already dropped support for gcc-9. I don't see any good reason to support anything <9 as a result.
2. GCC is relatively old, and we're probably supporting a bit too many compiler versions anyway. In Ubuntu 22.04, gcc-11 is downloaded by default with `apt`. It doesn't seem many system are still using gcc.
3. There is a weird compiler bug in gcc-8 which is causes failure when compiling gem5 since the inclusion of #1123. The error received is as follows:

```sh
In file included from src/arch/riscv/tlb.hh:42,
                 from src/arch/riscv/mmu.hh:45,
                 from build/ALL/arch/riscv/generated/exec-g.cc.inc:14,
                 from build/ALL/arch/riscv/generated/generic_cpu_exec.cc:5:
src/arch/riscv/utility.hh: In instantiation of ‘FloatType gem5::RiscvISA::ftype(IntType) [with FloatType = float8_t; IntType = unsigned char]’:
build/ALL/arch/riscv/generated/exec-ns.cc.inc:38839:42:   required from ‘gem5::Fault gem5::RiscvISAInst::Vfwcvt_xu_f_vMicro<ElemType>::execute(gem5::ExecContext*, gem5::trace::InstRecord*) const [with ElemType = float8_t; gem5::Fault = std::shared_ptr<gem5::FaultBase>]’
build/ALL/arch/riscv/generated/exec-ns.cc.inc:38856:16:   required from here
src/arch/riscv/utility.hh:327:15: error: parameter ‘a’ set but not used [-Werror=unused-but-set-parameter]
 ftype(IntType a) -> FloatType
       ~~~~~~~~^
src/arch/riscv/utility.hh: In instantiation of ‘IntType gem5::RiscvISA::f_to_wui(FloatType, uint_fast8_t) [with FloatType = float8_t; IntType = short unsigned int; uint_fast8_t = unsigned char]’:
build/ALL/arch/riscv/generated/exec-ns.cc.inc:38838:49:   required from ‘gem5::Fault gem5::RiscvISAInst::Vfwcvt_xu_f_vMicro<ElemType>::execute(gem5::ExecContext*, gem5::trace::InstRecord*) const [with ElemType = float8_t; gem5::Fault = std::shared_ptr<gem5::FaultBase>]’
build/ALL/arch/riscv/generated/exec-ns.cc.inc:38856:16:   required from here
src/arch/riscv/utility.hh:570:20: error: parameter ‘a’ set but not used [-Werror=unused-but-set-parameter]
 f_to_wui(FloatType a, uint_fast8_t mode)
```

Note: This is currently causing our SST Daily tests to fail, and our compiler tests to fail.

### DIFF
--- a/.github/workflows/compiler-tests.yaml
+++ b/.github/workflows/compiler-tests.yaml
@@ -16,8 +16,8 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                image: [gcc-version-13, gcc-version-12, gcc-version-11, gcc-version-10, gcc-version-8, clang-version-16, clang-version-15, clang-version-14,
-                    clang-version-13, clang-version-12, clang-version-11, clang-version-10, clang-version-9, clang-version-8, clang-version-7, ubuntu-20.04_all-dependencies,
+                image: [gcc-version-13, gcc-version-12, gcc-version-11, gcc-version-10, clang-version-16, clang-version-15, clang-version-14, clang-version-13,
+                    clang-version-12, clang-version-11, clang-version-10, clang-version-9, clang-version-8, clang-version-7, ubuntu-20.04_all-dependencies,
                     ubuntu-22.04_all-dependencies, ubuntu-24.04_all-dependencies, ubuntu-24.04_min-dependencies]
                 opts: [.opt, .fast]
         runs-on: [self-hosted, linux, x64]
--- a/SConstruct
+++ b/SConstruct
@@ -631,8 +631,8 @@ for variant_path in variant_paths:
               "src/SConscript to support that compiler.")))
 
     if env['GCC']:
-        if compareVersions(env['CXXVERSION'], "7") < 0:
-            error('gcc version 7 or newer required.\n'
+        if compareVersions(env['CXXVERSION'], "10") < 0:
+            error('gcc version 10 or newer required.\n'
                   'Installed version:', env['CXXVERSION'])
 
         # Add the appropriate Link-Time Optimization (LTO) flags if
@@ -655,17 +655,6 @@ for variant_path in variant_paths:
         env.Append(TCMALLOC_CCFLAGS=[
             '-fno-builtin-malloc', '-fno-builtin-calloc',
             '-fno-builtin-realloc', '-fno-builtin-free'])
-
-        if compareVersions(env['CXXVERSION'], "9") < 0:
-            # `libstdc++fs`` must be explicitly linked for `std::filesystem``
-            # in GCC version 8. As of GCC version 9, this is not required.
-            #
-            # In GCC 7 the `libstdc++fs`` library explicit linkage is also
-            # required but the `std::filesystem` is under the `experimental`
-            # namespace(`std::experimental::filesystem`).
-            #
-            # Note: gem5 does not support GCC versions < 7.
-            env.Append(LIBS=['stdc++fs'])
 
     elif env['CLANG']:
         if compareVersions(env['CXXVERSION'], "6") < 0:
--- a/util/dockerfiles/docker-compose.yaml
+++ b/util/dockerfiles/docker-compose.yaml
@@ -42,13 +42,6 @@ services:
             context: ubuntu-24.04_min-dependencies
             dockerfile: Dockerfile
         image: ghcr.io/gem5/ubuntu-24.04_min-dependencies:latest
-    gcc-8:
-        build:
-            context: ubuntu-20.04_gcc-version
-            dockerfile: Dockerfile
-            args:
-                - version=8
-        image: ghcr.io/gem5/gcc-version-8:latest
     gcc-10:
         build:
             context: ubuntu-20.04_gcc-version
--- a/util/dockerfiles/sst/Dockerfile
+++ b/util/dockerfiles/sst/Dockerfile
@@ -36,6 +36,17 @@ RUN apt -y update && apt -y upgrade && \
 
 ENV SST_CORE_HOME="/sst/"
 
+# Use GCC 10
+RUN apt update && apt -y install gcc-10 g++-10
+RUN update-alternatives --install \
+    /usr/bin/g++ g++ /usr/bin/g++-10 100
+RUN update-alternatives --install \
+    /usr/bin/gcc gcc /usr/bin/gcc-10 100
+RUN update-alternatives --install \
+    /usr/bin/c++ c++ /usr/bin/g++-10 100
+RUN update-alternatives --install \
+    /usr/bin/cc cc /usr/bin/gcc-10 100
+
 # SST Stuff
 RUN mkdir /sst
 


powerjg:
+1 from me. Probably need to update the website too

ivanaamit (approved):

giactra:
Hi, I appreciate the effort put into making things work in our CI system. However It would be nice if other maintainers would be tagged in these important changes which are affecting the usability of gem5 for everyone. In the past we used to ask people in the gem5-dev mailing list if there was any concern in dropping support for an existing compiler.

For example in this case I would have suggested to simply warn if gcc < 10, rather than preventing compilation to happen. Dropping support means we don't test a specific compiler version but it seems a bit to much to throw an error for the default compiler that gets shipped with a currently suported LTS release (Ubuntu 20.04)..

powerjg:
There's two issues here:
1. **The technical issue**: Without significantly more maintainer resources, we cannot support every compiler on every OS across multiple years. It's simply impossible from a project management point of view. gem5 developers often use compiler options that are not backwards compatible. Fixing these issues are often very time consuming.
2. **The communication issue**: It's difficult to know who wants to be tagged on every change. If you watch the github project, you should receive an email for every PR. It's a lot, so it's inevitable that anyone will miss something. We also do a pretty good job of tagging things. As always, this is a collaborative open-source project, so you should feel free to suggest a better solution.

giactra:
> There's two issues here:
> 
>     1. **The technical issue**: Without significantly more maintainer resources, we cannot support every compiler on every OS across multiple years. It's simply impossible from a project management point of view. gem5 developers often use compiler options that are not backwards compatible. Fixing these issues are often very time consuming.
> 

I am 100% not disagreeing with this. But there is a difference between *dropping support* for gcc9 and *preventing any gcc9 build to happen*. IMHO it would have sufficed to *warn* rather than *panic*. Let me stress again it is the default compiler shipped with a living LTS release and we are not saying to a user "we don't support you if you decide to build gem5 with it", we are saying "we are not allowing you to build gem5 with it". Again, big difference.

>     2. **The communication issue**: It's difficult to know who wants to be tagged on every change. If you watch the github project, you should receive an email for every PR. It's a lot, so it's inevitable that anyone will miss something. We also do a pretty good job of tagging things. As always, this is a collaborative open-source project, so you should feel free to suggest a better solution.

I am against setting up lots of bureaucratic policies for something that can be handled with common sense... Similarly we don't impose a fixed minimum amount of time a patch has to be under review before it can be merged, though we are vocal when something big gets merged with a single review and with little time for reviewers to chime in (and that's fine!)
At the end of the day the message I sent was just a polite reminder there are other maintainers outside of UC Davis that would like to be included whenever there is a meaningful change like this one.
I think you would have similarly asked to be included if I had merged that patch with only Andreas review

abmerop:
This change completely breaks building GPUSE since the docker image has an old GCC.  Why is this not a warning instead?

powerjg:
I understand the idea of using a warning, but this is not going to work for the people who have the responsibility of maintaining the codebase.

The underlying problem is that gem5 is, in my point of view, an *incredibly unique* project that we distribute source code and not a binary. Given this odd distribution method, it is a massive burden on the maintainers to ensure compatibility with a huge number of compilers and systems. I've been saying for **years** that this burden is not sustainable. This fact, that the maintainers cannot keep up with both changing compilers and people using newer and newer compiler features, is not new.

There are two options going forward here:

## Option 1: Keep GCC 8 as a supported compiler

This will require a fix for the issue described above which was introduced by #1123. We can either revert #1123 or spend a not-insignificant amount of time trying to fix the problem.
Looking forward, I have major concerns that keeping GCC 8 as a supported compiler will continue to be a high maintenance burden. GCC does not support backwards compatibility well, and we continue to use newer and newer C++ features.

## Option 2: Drop GCC 8 as a supported compiler

If we do this, then we will also have to update the GPU-SE docker container to use a more recent version of GCC (cc: @mattsinc). I don't know the feasibility of this option.

## *NON-OPTION* Make GCC 8 "unsupported" but allow it to compile with a warning

This is a firm non-option from a maintenance point of view. If we did this, then we will *not notice* when changes (like #1123) cause compiler errors. Instead, they will manifest by either issues opened or by other tests (e.g., the GPU-SE tests) failing.

I'm open to other suggestions, as long as they do not increase the testing and maintenance burden.

abmerop:
> ## Option 2: Drop GCC 8 as a supported compiler
> If we do this, then we will also have to update the GPU-SE docker container to use a more recent version of GCC (cc: @mattsinc). I don't know the feasibility of this option.

I am working on this option. It looks like GCC 10 is available in the default 20.04 apt sources (i.e., it does not need some extra ppa repository).  Will post a PR later.

powerjg:
Thanks, Matt!

giactra:
> ## _NON-OPTION_ Make GCC 8 "unsupported" but allow it to compile with a warning
> 
> This is a firm non-option from a maintenance point of view. If we did this, then we will _not notice_ when changes (like #1123) cause compiler errors. 

Oddly enough, the opposite is true. First of all, we won't notice anyway if we disable a legitimate compiler to build gem5. Secondly, I am quite sure the issue reported by @BobbyRBruce  **IS NOT** a compiler bug, and we are removing the only compiler which is properly flagging the issue.

```
In file included from src/arch/riscv/tlb.hh:42,
                 from src/arch/riscv/mmu.hh:45,
                 from build/ALL/arch/riscv/generated/exec-g.cc.inc:14,
                 from build/ALL/arch/riscv/generated/generic_cpu_exec.cc:5:
src/arch/riscv/utility.hh: In instantiation of ‘FloatType gem5::RiscvISA::ftype(IntType) [with FloatType = float8_t; IntType = unsigned char]’:
build/ALL/arch/riscv/generated/exec-ns.cc.inc:38839:42:   required from ‘gem5::Fault gem5::RiscvISAInst::Vfwcvt_xu_f_vMicro<ElemType>::execute(gem5::ExecContext*, gem5::trace::InstRecord*) const [with ElemType = float8_t; gem5::Fault = std::shared_ptr<gem5::FaultBase>]’
build/ALL/arch/riscv/generated/exec-ns.cc.inc:38856:16:   required from here
src/arch/riscv/utility.hh:327:15: error: parameter ‘a’ set but not used [-Werror=unused-but-set-parameter]
 ftype(IntType a) -> FloatType
       ~~~~~~~~^
src/arch/riscv/utility.hh: In instantiation of ‘IntType gem5::RiscvISA::f_to_wui(FloatType, uint_fast8_t) [with FloatType = float8_t; IntType = short unsigned int; uint_fast8_t = unsigned char]’:
build/ALL/arch/riscv/generated/exec-ns.cc.inc:38838:49:   required from ‘gem5::Fault gem5::RiscvISAInst::Vfwcvt_xu_f_vMicro<ElemType>::execute(gem5::ExecContext*, gem5::trace::InstRecord*) const [with ElemType = float8_t; gem5::Fault = std::shared_ptr<gem5::FaultBase>]’
build/ALL/arch/riscv/generated/exec-ns.cc.inc:38856:16:   required from here
src/arch/riscv/utility.hh:570:20: error: parameter ‘a’ set but not used [-Werror=unused-but-set-parameter]
 f_to_wui(FloatType a, uint_fast8_t mode)
```

The compiler here is complaining we are not using a. Why?
We are calling [ftype](https://github.com/gem5/gem5/blob/develop/src/arch/riscv/utility.hh#L327) with [FloatType = float8_t; IntType = unsigned char].

Which means the resulting function is

```
ftype(IntType a) -> FloatType
{
    GEM5_UNREACHABLE;
}
```

So the issue is that we are not handling the unsigned char (uint8_t) case. This code as of now is generating a broken function with undefined behaviour at runtime. For whatever reason other compilers are not complaining for this. Not only the uint8_t case should be handled, but IMHO GEM5_UNREACHABLE shouldn't be used in this helper and a static_assertion should be preferred instead.

So we wouldn't have been able to catch this problem if it wasn't for gcc-8.

Technical point aside, I really just want to encourage other maintainers to ask for feedback for changes like this before merging things too quickly. Please don't take it as a personal attack; I just don't know a better way to convey my thoughts. I honestly believe the project would only benefit from having multiple people looking into things in a collaborative manner. I am happy to continue this discussion on a different channel though as it is not strictly related to the PR

PR #1156, opened 2024-05-21
mem-ruby: Remove VIPER StoreThrough temp cache storage

abmerop:
StoreThrough in VIPER when the TCP is disabled, GLC bit is set, or SLC bit is set will bypass the TCP, but will temporarily allocate a cache entry seemingly to handle write coalescing with valid blocks. It does not attempt to evict a block if the set is full and the address is invalid. This causes a panic if the set is full as there is no spare cache entry to use temporarily to use for DataBlk manipulation. However, a cache block is not required for this.

This commit removes using a cache block for StoreThrough with invalid blocks as there is no existing data to coalesce with. It creates no allocate variants of the actions needed in StoreThrough and pulls the DataBlk information from the in_msg instead. Non-invalid blocks do not have this panic as they have a cache entry already.

Fixes issues with StoreThroughs on more aggressive architectures like MI300.

Change-Id: Id8687eccb991e967bb5292068cbe7686e0930d7d

### DIFF
--- a/src/mem/ruby/protocol/GPU_VIPER-TCP.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCP.sm
@@ -449,6 +449,28 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
     }
   }
 
+  action(wtna_writeThroughNoAlloc, "wtna", desc="Write through without allocation") {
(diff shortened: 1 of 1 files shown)


mattsinc (approved):

mattsinc:
@v-ramadas FYI

PR #1157, opened 2024-05-22
dev-amdgpu: Fix pending PCI RLC doorbell

abmerop:
SDMA RLC queues do not currently remove their doorbell mapping. This can cause issues re-registering the queue and prevents the pending doorbells feature from working. In addition the data value of the doorbell (the ring buffer rptr) is not saved, leading to UB when this workaround is used.

This commit removes the doorbell mapping from the gpu device when the SDMA engine unmaps an RLC queue and copies the next doorbell value to the pending packet as was originally intended.

Change-Id: Ifd551450f439c065579afcf916f8ff192e7598ab

### DIFF
--- a/src/dev/amdgpu/amdgpu_device.cc
+++ b/src/dev/amdgpu/amdgpu_device.cc
@@ -517,10 +517,13 @@ AMDGPUDevice::writeDoorbell(PacketPtr pkt, Addr offset)
              offset);
 
         // We have to ACK the PCI packet immediately, so create a copy of the
-        // packet here to send again.
+        // packet here to send again. The packet data contains the value of
+        // the doorbell to write so we need to copy that as the original
+        // packet gets deleted after the PCI write() method returns.
         RequestPtr pending_req(pkt->req);
         PacketPtr pending_pkt = Packet::createWrite(pending_req);
         uint8_t *pending_data = new uint8_t[pkt->getSize()];
+        memcpy(pending_data, pkt->getPtr<uint8_t>(), pkt->getSize());
         pending_pkt->dataDynamic(pending_data);
 
         pendingDoorbellPkts.emplace(offset, pending_pkt);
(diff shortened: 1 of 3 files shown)


abmerop:
MI300+Llama3 gets pretty far with this

abmerop:
Marking a milestone for attention

mattsinc:
> Marking a milestone for attention

Sorry, I am massively behind in emails and tonight is the SC rebuttal deadline.  I will review this one now but rest will have to wait until tomorrow.

src/dev/amdgpu/amdgpu_device.cc
```diff
@@ -521,6 +521,7 @@ AMDGPUDevice::writeDoorbell(PacketPtr pkt, Addr offset)
         RequestPtr pending_req(pkt->req);
         PacketPtr pending_pkt = Packet::createWrite(pending_req);
         uint8_t *pending_data = new uint8_t[pkt->getSize()];
+        memcpy(pending_data, pkt->getPtr<uint8_t>(), pkt->getSize());
```
mattsinc, line 526: Can you please add a comment explaining this?  I'm assuming it's copying doorbell information into the packet?  But why is it needed?

mattsinc (changes requested):

src/dev/amdgpu/amdgpu_device.cc
abmerop: As it is now, the packet data is uninitialized and the random value read when replaying the packet won't write to the correct doorbell.  This causes the queue to not be woken up, followed by the application deadlocking

Edit: Slight correction, the packet data contains the circular buffer pointer and the packet address is the doorbell ID.  The application deadlock occurs since it begins reading a random amount past the write pointer value

mattsinc (approved):

PR #1162, opened 2024-05-24
arch-vega: Fix GCC 13 build errors

abmerop:
The new static analysis in GCC 13 finds issues with operand.hh. This commit fixes the error so that gem5 compiles when BUILD_GPU is true.

Change-Id: I6f4b0d350f0cabb6e356de20a46e1ca65fd0da55

### DIFF
--- a/src/arch/amdgpu/vega/operand.hh
+++ b/src/arch/amdgpu/vega/operand.hh
@@ -514,6 +514,29 @@ namespace VegaISA
         {
             assert(NumDwords == 1 || NumDwords == 2);
 
+            if (_opIdx >= REG_INT_CONST_POS_MIN &&
+                _opIdx <= REG_INT_CONST_NEG_MAX) {
+                assert(sizeof(DataType) <= sizeof(srfData));
+                DataType misc_val(0);
+                assert(isConstVal(_opIdx));
+                misc_val = (DataType)_gpuDynInst
+                    ->readConstVal<DataType>(_opIdx);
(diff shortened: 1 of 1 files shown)


src/arch/amdgpu/vega/operand.hh
```diff
@@ -514,6 +514,29 @@ namespace VegaISA
         {
             assert(NumDwords == 1 || NumDwords == 2);
 
+            if (_opIdx >= REG_INT_CONST_POS_MIN &&
```
mattsinc, line 517: Please just add a comment explaining what this is doing -- I think I get it, but not 100% sure.

mattsinc (dismissed):

src/arch/amdgpu/vega/operand.hh
abmerop: I will add a comment saying something along the lines of "there are ~80 inline constants" so using a switch statement is painful.  The second check I put in was too avoid too much code duplication. 

On a more technical note, it seems GCC 13's static analyzer is testing for all possible values of this `_opIdx` even though this method would never be called for values not checked here (panic in the default just in case).  I think making `isConstVal` constexpr would work, but I was concerned that would break older compiler versions that are still (for now) supported officially by gem5

mattsinc: Sounds great, thanks!

BobbyRBruce (approved):

Issue #981, opened 2024-04-01
c.jr/c.jalr semantic wrong


Ye-Jinhong:
**Describe the bug**
As RISC-V unpriviledge isa spec says, `c.jr/c.jalr` expands to `jalr x0/x1 0(rs1)`. Since `jalr` should force the bit 0 to **zero**, I think `c.jr/c.jalr` should act the same way.

hnpl:
Hi @Ye-Jinhong, can you elaborate the issue more? (i.e., what was the unexpected output for c.jal/c.jalr?)

Ye-Jinhong:
Hi @hnpl , it is specified that `jalr` should set the bit 0 to **zero** in RISC-V unpriviledge isa spec.

> The indirect jump instruction JALR (jump and link register) uses the I-type encoding. The target address is obtained by adding the sign-extended 12-bit I-immediate to the register rs1, then setting the least-significant bit of the result to zero. 

And in "C" extension, `c.jr/c.jalr` expands to `jalr x0/x1 0(rs1)`

>C.JR (jump register) performs an unconditional control transfer to the address in register rs1. C.JR expands to jalr x0, 0(rs1).

>C.JALR (jump and link register) performs the same operation as C.JR, but additionally writes the address of the instruction following the jump (pc+2) to the link register, x1. C.JALR expands to jalr x1, 0(rs1).

So  the result of `c.jr/c.jalr rs1` which is the next fetch PC should equals to 0 + rs1 of `jalr`, setting the bit 0 to **zero**. This means the result is the value `(rs1 + 0) & ~0x1`.

In [decoder.isa](https://github.com/gem5/gem5/blob/stable/src/arch/riscv/isa/decoder.isa), line 450 to 456
```
                0x0: Jump::c_jr({{
                    if (RC1 == 0) {
                        return std::make_shared<IllegalInstFault>(
                                "source reg x0", machInst);
                    }
                    NPC = rvZext(Rc1);
                }}, IsIndirectControl, IsUncondControl);
``` 
and line 468 to 471
```
                    default: Jump::c_jalr({{
                        ra = rvSext(NPC);
                        NPC = rvZext(Rc1);
                    }}, IsIndirectControl, IsUncondControl, IsCall);
```
The NPC is not correct due to not masking the bit 0 like `NPC = rvZext(Rc1) & ~0x1`.

[...]

PR #1163, opened 2024-05-24
arch-riscv: Fix c.jalr and c.jr instruction

rogerchang23424:
The bit 0 of register should be 0 for jump address. Wrong handling the jump address may cause infinite run or segment fault.

gem5 issue: https://github.com/gem5/gem5/issues/981
Change-Id: I605de64bd3cece70288ffcfaf1f4263c17ab2971

### DIFF
--- a/src/arch/riscv/isa/formats/compressed.isa
+++ b/src/arch/riscv/isa/formats/compressed.isa
@@ -157,6 +157,36 @@ def template CBasicExecute {{
     }
 }};
 
+def template CJumpExecute {{
+    Fault
+    %(class_name)s::execute(
+        ExecContext *xc, trace::InstRecord *traceData) const
+    {
+        %(op_decl)s;
+        %(op_rd)s;
+        %(code)s;
+        %(op_wb)s;
+        return NoFault;
+    }
+
+    std::unique_ptr<PCStateBase>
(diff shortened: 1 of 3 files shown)


src/arch/riscv/isa/formats/compressed.isa
```diff
@@ -111,6 +111,8 @@ def format CompressedStore(sdisp_code, memacc_code,
         inst_flags, 'Store', exec_template_base='Store')
 }};
 
+
+
```
hnpl: Nit: these extra empty lines should be removed.

hnpl (approved):
The code looks good to me.

I'm not sure if the riscv-related tests we have contain those instructions. Can you let me know if the code has been tested?

rogerchang23424:
I have tested them with riscof. However, the case doesn't included in the riscof, so I add them to test case https://github.com/riscv-non-isa/riscv-arch-test/pull/466.

Since we can't dump the memory signature directly with gem5. Instead, I compared the signature with customized plugin (https://github.com/gem5/gem5-resources/pull/7) with other simulators.

hnpl:
That's great!!

BobbyRBruce (approved):

PR #1167, opened 2024-05-25
stdlib,configs,tests: Add gem5 MultiSim (MultiProcessing for gem5)

BobbyRBruce:
This allows for multiple gem5 simulations to be spawned from a single parent gem5 process, as defined in a simgle gem5 configuration. In this design _all_ the `Simulator`s are defined in the simulation script and then added to the mutlisim module. For example:

```py
from gem5.simulate.Simulator import Simulator
import gem5.utils.multisim as multisim

# Construct the board[0] and board[1] as you wish here...

simulator1 = Simulator(board=board[0], id="board-1")
simulator2 = Simulator(board=board[1], id="board-2")

multisim.add_simulator(simulator1)
multisim.add_simulator(simulator2)
```

This specifies that two simulations are to be run in parallel in seperate threads: one specified by `simulator1` and another by `simulator2`. They are then added to MultiSim via the `multisim.add_simulator` function. The user can specify an id via the Simulator constructor. This is used to give each process a unique id and output directory name. Given this, the id should be a helpful name describing the simulation being specified. If not specified one is automatically given.

To run these simulators we use `<gem5 binary> -m gem5.utils.multisim <script> -p <num_processes>`. Note: multisim is an executable module in gem5. This is the same module we input into our scripts to add the simulators. This is an intentionally modular encapsulated design. When the module processes a script it will schedule multiple gem5 jobs and, dependent on the number of processes specified, will create child gem5 processes to processes tjese jobs (jobs are just gem5 simulations in this case). The `--processes` (`-p`) argument is optional and if not specified the max number of processes which can be run concurrently will be the number of available threads on the host system.

The id for each process is used to create a subdirectory inside the `outputdor` (`m5out`) of that id name. E.g, in the example above the ID's are `board-1` and `board-2`. Therefore the m5 out directory will look as follows:

```sh
- m5out
    - board-1
        - stats.txt
        - config.ini
        - config.json
        - terminal.out
    - board-2
        - stats.txt
        - config.ini
        - config.json
        - terminal.out
```

Each simulations output is encapsulated inside the subdirectory of the id name.

If the multisim configuation script is passed directly to gem5 (like a traditional gem5 configuraiton script, i.e.: `<gem5 binary> <script>`), the user may run a single simulation specified in that script by passing its id as an argument. E.g. `<gem5 binary> <script> board-1` will run the `board-1` simulation specified in `script`. If no argument is passed an Exception is raised asking the user to either specify or use the MultiSim module if multiprocessing is needed.

If the user desires a list of ids of the simulations specified in a given MultiSim script, they can do so by passing the `--list` (`-l`) parameter to the config script. I.e., `<gem5 binary> <script> --list` will list all the IDs for all the simulations specified in`script`.

This change comes with two new example scripts found in 'configs/example/gem5_library/multsim" to demonstrate multisim in both an SE and FS mode simulation. Tests have been added which run these scripts as part of gem5' Daily suite of tests.

Notes
=====

* **Bug fixed**: The `NoCache` classic cache hierarchy has been modified so the Xbar is no longet set with a `__func__` call. This interfered with MultiProcessing as this structure is not serializable via Pickle. This was quite bad design anyway so should be changed

* **Change**: `readfile_contents` parameter previously wrote its value to a file called "readfile" in the output dorectory. This has been changed to write to a file called "readfile_{hash}" with "{hash}" being a hash of the `readfile_contents`. This ensures that, during multisim running, this file is not overwritten by other processes.

* **Removal note**: This implementation supercedes the functionality outlined in 'src/python/gem5/utils/multiprocessing'. As such, this code has been removed.

Limitations/Things to Fix/Improve
=================================

* Though each Simulator process has its own output directory (a subdirectory within m5out, with an ID set by the user unique to that Simulator), the stdout and stderr are still output to the terminal, not the output directory. This results in: 1. stdout and stderr data lost and not recorded for these runs. 2. An incredibly noisy terminal output.
* Each process uses the same cached resources. While there are locks on resources when downloading, each processes will hash the resources they require to ensure they are valid. This is very inefficient in cases where resources are common between processes (e.g., you may have 10 processes each using the same disk image with each processes hashing the disk images independently to give the same result to validate the resources).

Change-Id: Ief5a3b765070c622d1f0de53ebd545c85a3f0eee


powerjg:
> Passing a list of Simulator objects means the MultiSim module calls run. Ergo users can't use simulator.run(<max_ticks>). However, this should be available via a set function in the simulator. This may require some refactoring in some cases.

I believe there is already an interface for this. If not @mbabaie and I were talking about it last week.

> Though each Simulator process has its own output directory (a subdirectory within m5out, with an ID set by the user unique to that Simulator), the stdout and stderr are still output to the terminal, not the output directory. This results in: 1. stdout and stderr data lost and not recorded for these runs. 2. An incredibly noisy terminal output.

This is annoying, but I would say it's ok.

> If a 'readfile' is created the readfile is set to the parent root directory, not the individual run. if the readfile is different for each simulation, this can cause bugs as the file is overwritten constantly.

It seems like this limitation should be solved before merging.

>Each process uses the same cached resources. If not present many threads can end up downloading the same resources at the same time. In a similar vein, processes often try to obtain resources at the same time causing the a lot of slowdown.

Does the file lock not work for this? I think this limitation is also high priority.

> All available threads are taken. The user cannot specify the number of processes. There's no real orchastration framework for these processes. running many in parallel can be difficult to manage.

I don't quite understand what this means, but I think it's OK for now.

configs/example/gem5_library/multisim/multisim-fs-x86-npb.py
```diff
@@ -0,0 +1,130 @@
...
+# Here we imagine an examperiment wanting to run each NPB benchmark on the same
+# system twice: once with 1 core and once with 2 cores.
+for benchmark in obtain_resource("npb-benchmark-suite"):
+    for num_cores in [1, 2]:
+        # Limit the number of simulators, so we don't run a massive amount of
+        # them in the proof-of-concept.
+        if multisim.num_simulators() >= 6:
+            break
+
+        cache_hierarchy = NoCache()
```
powerjg: Can we do the MESI_Two_level (it's required above...)

src/python/gem5/utils/multisim/__init__.py
```diff
@@ -0,0 +1 @@
+from .multisim import *
```
powerjg: 1. this file needs a copyright statement.
2. It's a bit of a pain, but it would be better to specify what you want to import instead of using `*`

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -0,0 +1,299 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""This module contains the gem5 MultiSim framework. The gem5 MultiSim
+work functions by allowing a user to specify multiple simulations to run from
+a single gem5 config script. The MuliSim framework will then run these
+simulations in parallel using the Python multiprocessing library.
+
+The framework works by having the user add the simulators in a configuration
+via the `add_simulator`function, each with a unique, user specified, id. This
```
powerjg:
```suggestion
via the `add_simulator` function, each with a unique, user specified, id. This
```

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -0,0 +1,299 @@
...
+MultiSim framework then uses the Python multiprocessing library to run the
+simulations in parallel by loading the config script as a module in each child
+process and then selecting the simulation to run via the iid in the global set
+of scheduled simulators jobs to run.
+The only difference between the child processes is the id of the simulator.
+
+Important notes
+---------------
+
+1. You cannot load/instnatiate the simulators in the main process. You cannot
```
powerjg:
```suggestion
1. You cannot load/instantiate the simulators in the main process. You cannot
```

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -0,0 +1,299 @@
...
+even load the config script (i.e., `import config_script`). This means the
+config script is passed as a string referencing the config script as a modeule.
```
powerjg:
```suggestion
config script is passed as a string referencing the config script as a module.
```

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -0,0 +1,299 @@
...
+        """This function redirects the child process output to a subdirectory
+        in the directory specified by the `outdir` option.
+
+        This subdirectory is named after the id assigned to the simulator.
+
+        Note: This function is not yet complete. There are two missing
+        features:
+
+        1. The stdout and stderr of the child process are not redirected to the
+           subdirectory. The appear in the terminal.
```
powerjg: This feature worked with the previous implementation of multiprocessing. The *right* thing to do is to respect the `-re` parameters of the gem5 binary that is used to spawn the multiprocessing script.

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -0,0 +1,299 @@
...
+    """Returns the number of simulators added to the MultiSim."""
+    return len(_multi_sim)
+
+
+def add_simulator(simulator: "Simulator", id: str) -> None:
+    """Add a single simulator to the Multisim. Doing so informs the simulators
+    to run this simulator via multiprocessing.
+
+    **Note:** If this function is not run using the MultiSim module then the
+    user will be promted to eitehr do so if they desire multiple gem5
```
powerjg:
```suggestion
    user will be prompted to either do so if they desire multiple gem5
```

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -0,0 +1,299 @@
...
+    processes or to pass the id of the simulator to run. This function will
+    attempt to run the simulation with the id passed as an argument. If
+    such simulatoon exists the simulation will end without failure (or any
```
powerjg:
```suggestion
    such simulation exists the simulation will end without failure (or any
```

powerjg (commented):
Mostly just spelling mistakes 😆. You should install a spell checker!

I am still concerned about the two limitations mentioned in my comment above. But otherwise, it all looks great! Thanks for all of the improvements over my first stab at it!

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -0,0 +1,299 @@
...
+    Set,
+)
+
+# A global variable which __main__.py flips to `True` when multisim is run as
+# an executable module.
+module_run = False
+
+
+class SimulatorJob:
+    def __init__(self, simulator: "Simulator", id: str):
```
Harshil2107: I think it would be good to have a default for the ids, maybe somthing like `workload-id-version`?

I think the easiest way to do this would be to have a `get_id` for `Simulater` class which would default value to id and version of the resource it is suppossed to run (we can get those as all resource clas have a `get_id` and `get_version` functions). We can set the default id to be the `SimulaterJob` id.

while typing this out, I also realised that this would not be as straight forward if we are not using workloads as:

- What if we are using local resources, generally we users wont give and id and version for these.
- What if we are using something the is not `set_workload`, but lets say `set_kernel_disk_worklaod` then how would this work?

This is just a small thing that I feel would be convinent, it is not necessay.

Harshil2107:
I think this should be more of a future feature thing, but I wanted to ask if this would be possible to implement easily with the current implementation (to me it looks like multiprocessing implementation wouldnt need a lot of changes).

With the introduction of suites, it would be ncie if we could just run the suite directly, and it would spawn multiple processes to run the workloads in the suite automatically. Something along the lines of:

```
resource = obtain_resource("suite")

#some function that allows us to pass a board and a suite
run_suite = add_workload_to_board(board, resoruce, num_process=4)
```

The above could be directly a part of the multiprocessing module like a helper class or a sub class. This also relies on the fact that we have some default for the output directories. 

I think it would be nice to integrate suites more heavliy with the multiprocessing module as it is essentially how we want it to run (not sure on this?), something like I want to run this suite on this board.

src/python/gem5/simulate/simulator.py
```diff
@@ -105,6 +106,8 @@ def __init__(
         ] = None,
         expected_execution_order: Optional[List[ExitEvent]] = None,
         checkpoint_path: Optional[Path] = None,
+        max_ticks: Optional[int] = m5.MaxTick,
```
powerjg, line 109: I'm against this change (apologies if I didn't notice until now).

The max ticks should not be part of the simulator object. It's a function of a particular run. A normal simulation procedure would be run multiple times with different max ticks.

src/python/gem5/simulate/simulator.py
```diff
@@ -360,6 +384,46 @@ def function_generator(func: Callable):
...
+
+        if not self._instantiated:
+            raise Exception(
+                "If not set via the Simulator constructor, or via `set_id`, "
+                "the simulator/simulation ID cannot be obtained "
+                "prior to initialization as it is based, partially, "
+                "on a hash of the instiaite simulated system."
+            )
+
+        return self.hash().hex().lower()
```
powerjg: This is a string, right? (https://docs.python.org/3/library/functions.html#hex) The function return says its an `int`. Which one is wrong?

src/python/gem5/simulate/simulator.py
```diff
@@ -360,6 +384,46 @@ def function_generator(func: Callable):
...
+
+    def set_max_ticks(self, max_tick: int) -> None:
```
powerjg, line 432: I like this function. Can you add a docstring that specifies whether this is a delta or absolute tick (I assume absolute, but it's important to make that clear).

src/python/gem5/simulate/simulator.py
```diff
@@ -649,3 +728,32 @@ def save_checkpoint(self, checkpoint_dir: Path) -> None:
...
+            hash(attr)
+            for attr in [
+                self._full_system,
+                self._on_exit_event,
+                self._expected_execution_order,
+                self._checkpoint_path,
+            ]
+        )
+
+    def hash(self) -> int:
```
powerjg: This should probably be `__hash__` if you want it to work with dictionaries, sets, and the builtin `hash` function.

src/python/gem5/simulate/simulator.py
```diff
@@ -649,3 +789,32 @@ def save_checkpoint(self, checkpoint_dir: Path) -> None:
...
         """
         m5.checkpoint(str(checkpoint_dir))
+
+    def _get_simobject_tree_hash(self) -> int:
+        if self._instantiated:
+            with StringIO() as f:
+                self._root.print_ini(ini_file=f)
+                return hash(f.read())
+
+        return self.hash(self.board)
```
powerjg: This seems like it's going to be a big overhead. I see two problems
1. I think it's a bad idea to write a file and then read it back to do this. We should make the `ini` generation just be a string (or develop a better hash).
2. This is going to be high overhead no matter what, We should cache this value.

BobbyRBruce: Thank you for your comments but this is still a draft RN, I'm playing around with these changes and just Github to store my progress. I'll let you know when I arrive on a solution here and for your other points. I've not executed any of this new code yet so would not be surprised if it's buggy.

src/python/gem5/simulate/simulator.py
BobbyRBruce: It should be string. Fixed.

src/python/gem5/simulate/simulator.py
BobbyRBruce: I think this warrants further discussion because I don't fully understand it. Setting of the max ticks has been part of the simulator for some time so not exactly sure what you're asking to be changed here.

src/python/gem5/utils/multisim/multisim.py
BobbyRBruce: I'm honestly not sure exactly how your code worked, but the design right now does have the parent processes parameters apply to the child process. I.e, `-re` will redirect the terminal output to the m5out directory. Is this how you intended this to work?

src/python/gem5/utils/multisim/multisim.py
BobbyRBruce: I've kind of went with this. I've added IDs to Simulators but left it up to the multisim module to add a default if none were provided. I spent too long experimenting with a "hash as ID" idea but it just provided too complex to work.

BobbyRBruce:
> I think this should be more of a future feature thing, but I wanted to ask if this would be possible to implement easily with the current implementation (to me it looks like multiprocessing implementation wouldnt need a lot of changes).
> 
> With the introduction of suites, it would be ncie if we could just run the suite directly, and it would spawn multiple processes to run the workloads in the suite automatically. Something along the lines of:
> 
> ```
> resource = obtain_resource("suite")
> 
> #some function that allows us to pass a board and a suite
> run_suite = add_workload_to_board(board, resoruce, num_process=4)
> ```
> 
> The above could be directly a part of the multiprocessing module like a helper class or a sub class. This also relies on the fact that we have some default for the output directories.
> 
> I think it would be nice to integrate suites more heavliy with the multiprocessing module as it is essentially how we want it to run (not sure on this?), something like I want to run this suite on this board.

Definitely more of a future feature for now. With this I think you could easily write such a helper function.

configs/example/gem5_library/checkpoints/riscv-hello-save-checkpoint.py
```diff
@@ -107,8 +107,8 @@
 
 # Lastly we run the simulation.
 max_ticks = 10**6
-simulator = Simulator(board=board, full_system=False)
-simulator.run(max_ticks=max_ticks)
+simulator = Simulator(board=board, full_system=False, max_ticks=max_ticks)
+simulator.run()
```
powerjg, line 111: Why this change? I think it makes a lot more sense to have the `max_ticks` in `run`, personally. See my comment below in the `run` function for more detail.

configs/example/gem5_library/multisim/multisim-fs-x86-npb.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+# Here we imagine an examperiment wanting to run each NPB benchmark on the same
+# system twice: once with 1 core and once with 2 cores.
+for benchmark in obtain_resource("npb-benchmark-suite"):
+    for num_cores in [1, 2]:
+        # Limit the number of simulators, so we don't run a massive amount of
+        # them in the proof-of-concept.
+        if multisim.num_simulators() >= 6:
+            break
```
powerjg: Does this line mean that this script will just fail on a system with more than 6 cores? I would suggest at least printing why you're exiting and not doing any work.

configs/example/gem5_library/multisim/multisim-fs-x86-npb.py
```diff
@@ -0,0 +1,140 @@
...
+    processor.switch()
+    yield False
+
+
+def handle_workend():
+    m5.stats.dump()
+    yield True
+
+
+# Here we imagine an examperiment wanting to run each NPB benchmark on the same
```
powerjg:
```suggestion
# Here we imagine an experiment wanting to run each NPB benchmark on the same
```

configs/example/gem5_library/multisim/multisim-fs-x86-npb.py
```diff
@@ -0,0 +1,140 @@
...
+            board=board,
+            on_exit_event={
+                ExitEvent.WORKBEGIN: handle_workbegin(),
+                ExitEvent.WORKEND: handle_workend(),
+            },
+        )
+
+        simulator.set_id(f"{benchmark.get_id()}_cores-{num_cores}")
+
+        multisim.add_simulator(simulator)
```
powerjg, line 138: I think it would be better to explicitly run the simulations. E.g.,

```suggestion
        multisim.add_simulator(simulator)
        
multisim.run_all(num_processes=4)
```

src/python/gem5/simulate/simulator.py
powerjg: Ok, I've come around to this. But, please see below with my comment on the `run` function.

src/python/gem5/simulate/simulator.py
```diff
@@ -505,6 +589,38 @@ def get_roi_ticks(self) -> List[int]:
 
         return to_return
 
+    def override_outdir(self, new_outdir: Path) -> None:
```
powerjg, line 592: Cool! This is a good solution

src/python/gem5/simulate/simulator.py
```diff
@@ -574,8 +690,23 @@ def run(self, max_ticks: int = m5.MaxTick) -> None:
                           exit event is received, if another simulation exit
                           event is met the tick count is reset. This is the
                           **maximum number of ticks per simulation run**.
+                          **WARNING**: This this parameter is deprecated and
+                          will be removed in future releases of gem5. Please
+                          set the max ticks via the `set_max_ticks` function.
+                          or via the Simulator constructor.
```
powerjg: I disagree with deprecating this API. 

I suggest keeping this API and doing a `warn` if the `_max_ticks` is *changed* when calling `run` from when the simulator was originally instantiated or if set via `set_max_ticks`. 

I think the *most common* use case will be to set the max ticks in the `run` function.

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -0,0 +1,245 @@
...
+from typing import (
+    Optional,
+    Set,
+)
+
+# A global variable which __main__.py flips to `True` when multisim is run as
+# an executable module.
+module_run = False
+
+_multi_sim: Set["Simulator"] = set()
```
powerjg, line 70: Just curious (and related to the previous version with the `hash` implementation)... is it possible to have two simulators that are *exactly* the same and run both simulations? I actually want this feature, because I may have some pseudo random number which makes the simulations behave differently.

src/python/m5/params.py
```diff
@@ -882,9 +882,6 @@ def __str__(self):
     def __eq__(self, other):
         return self.pc == other.get_pc() and self.count == other.get_count()
 
-    def __hash__(self):
-        return hash((int(self.pc), int(self.count)))
-
```
powerjg: Why this change?

src/python/m5/proxy.py
```diff
@@ -167,6 +167,9 @@ def getindex(obj, index):
     def set_param_desc(self, pdesc):
         self._pdesc = pdesc
 
+    def __hash__(self) -> int:
+        return simobject_hash(str(self))
+
```
powerjg: Why this change?

src/python/gem5/utils/multisim/__main__.py
```diff
@@ -0,0 +1,69 @@
...
+    parser = argparse.ArgumentParser(
+        description="Pass the config script specifying the simulations to run "
+        "using multisim."
+    )
+    parser.add_argument(
+        "config",
+        type=str,
+        help="The path to the config script specifying the simulations to run using multisim.",
+    )
+    parser.add_argument(
```
powerjg: Should this have the following?
```suggestion
    parser.add_argument(
        default=None,
```

configs/example/gem5_library/multisim/multisim-fs-x86-npb.py
powerjg: Ok, I have thought about it a bit more, and I'm not sure if I agree with myself... We need *something* here, but I don't know what. Maybe something like

```suggestion
        if __name__ == "__m5_main__" or if __name__ == "__main__": 
            print(f"""This script is meant to be run with the `multisim` module
You can run it with `<gem5 binary> -m gem5.multiprocessing.multisim -p <processes> {__file__}
See ??? for more information"""
           
```

configs/example/gem5_library/multisim/multisim-fs-x86-npb.py
BobbyRBruce: Yeah, this is just debugging code I left in by minstake.

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -90,7 +90,8 @@ def get_simulator_ids_child_process(
 
     _load_module(module_path)
     global _multi_sim
-
+    print(_multi_sim)
+    print([sim.get_id() for sim in _multi_sim])
```
powerjg: Debugging?

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -114,6 +115,7 @@ def get_simulator_ids(config_module_path: Path) -> Set[str]:
     )
     p.start()
     p.join()
+    print(dict_return)
```
powerjg: Ditto

powerjg (commented):
Apologies if this wasn't ready for another review.

src/python/gem5/utils/multisim/multisim.py
BobbyRBruce: yeah, I should have this marked as draft. I still like these debug prints there.

src/python/m5/params.py
BobbyRBruce: Arg... this was the cause of a bug I've been trying to figure out for 20 minutes. Yeah, I didn't mean to remove this :).

src/python/gem5/utils/multisim/multisim.py
BobbyRBruce: It would still be possible. Simulator objects can be exactly the same. The Python hash and equals operator defaults to the object pointer, not anything to do with the contents.

configs/example/gem5_library/multisim/multisim-fs-x86-npb.py
BobbyRBruce: I'm not really sure what's actionable here. There's already code which gives this error in the multisim module: that is if a user uses `multisim.add_simulator` outsi@de of the `gem5.multiprocessing.multisim` module, so i don't see the use in adding this to the configs.

BobbyRBruce:
@powerjg : This is ready to review. I'll do a 'squash and merge' for this so you can ignore all the minor commits in this PR.

src/python/gem5/simulate/simulator.py
```diff
@@ -689,21 +689,14 @@ def run(self, max_ticks: Optional[int] = None) -> None:
...
-                          **maximum number of ticks per simulation run**.
-                          **WARNING**: This this parameter is deprecated and
-                          will be removed in future releases of gem5. Please
-                          set the max ticks via the `set_max_ticks` function.
-                          or via the Simulator constructor.
+                          **maximum number of ticks per simulation run.
         """
 
-        if max_ticks:
+        if hasattr(self, "_max_ticks") and self._max_ticks and max_ticks:
```
powerjg: I'm not sure this is going to work as expected. You set `_max_ticks` in the constructor, so `hasattr(self, "_max_ticks")` will always be true. https://github.com/gem5/gem5/blob/9674ca1f45d171e01349d0ae871a52a329fd0667/src/python/gem5/simulate/simulator.py#L293

```suggestion
        if max_ticks and max_ticks != self._max_ticks:
```

src/python/gem5/utils/multisim/multisim.py
powerjg: Thanks!

src/python/m5/params.py
```diff
@@ -882,6 +882,9 @@ def __str__(self):
     def __eq__(self, other):
         return self.pc == other.get_pc() and self.count == other.get_count()
 
+    def __hash__(self):
+        return hash((int(self.pc), int(self.count)))
+
```
powerjg: Can you confirm that this change is needed?

src/python/m5/proxy.py
```diff
@@ -167,9 +167,6 @@ def getindex(obj, index):
     def set_param_desc(self, pdesc):
         self._pdesc = pdesc
 
-    def __hash__(self) -> int:
-        return simobject_hash(str(self))
-
```
powerjg: Can you confirm that this change is needed?

powerjg (commented):
A few final things below. I'm going to check this out now and test it. I'll let you know if I find any other potential improvements.

src/python/m5/params.py
BobbyRBruce: Your commenting on a commit, not the PR. This isn't a change, it's reverting a change in an earlier commit I think.

src/python/m5/proxy.py
BobbyRBruce: your commenting on a commit, not the PR. This change is just reverting the PR back to the code as it currently is on develop

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -0,0 +1,299 @@
...
+        parser.add_argument(
+            "-l",
+            "--list",
+            help="List the ids of the simulators to run.",
+            action="store_true",
+        )
+        args = parser.parse_args()
+        if args.list:
+            print(simulator.get_id())
+        elif not args.id:
```
powerjg, line 283: If you specify an id, but it doesn't match, nothing happens. I think we should explicitly have an error in that case.

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -0,0 +1,299 @@
...
+            raise Exception(
+                "If running this script directly as a configuration script "
+                "then a single argument must be specified: the id of the "
+                "simulator to run. This will run the simulation associated "
+                "with that id and no other. If thgfe intent is instead to run "
```
powerjg:
```suggestion
                "with that id and no other. If the intent is instead to run "
```

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -0,0 +1,299 @@
...
+    **Note:** If this function is not run using the MultiSim module then the
+    user will be prompted to either do so if they desire multiple gem5
+    processes or to pass the id of the simulator to run. This function will
+    attempt to run the simulation with the id passed as an argument. If
+    such simulation exists the simulation will end without failure (or any
+    simulations having been run).
+
+    :param simulator: The simulator to add to the multisim.
+    :param id: The id of the simulator. This is used to reference the
+    simululation. This is particularly important when referencing the correct
```
powerjg:
```suggestion
    simulation. This is particularly important when referencing the correct
```

powerjg (commented):
These are my last comments!

src/python/m5/proxy.py
powerjg: Oh, my bad!

src/python/gem5/utils/multisim/multisim.py
BobbyRBruce: This may be hard so I'm gonna drop it for now. Warning to users: Without specifying a valid ID this will finish without running anything and without error.

powerjg (approved):

Adawntoremember (commented):
[]()

PR #1161, opened 2024-05-23
util: Bump GPUSE docker to GCC 10

abmerop:
The docker image uses GCC 8 by default which is no longer supported in gem5. Bump the version to GCC 10.

Change-Id: I0e46e0f82d881fe73eccef75e2f343b41073f6fe

PR #1170, opened 2024-05-27
util-docker,gpu,gpu-compute: Improve GCN-GPU Dockerfile

BobbyRBruce:
* The GCC used in the GCN-GPU images was increase from version 8 to 
  version 10. This was necessary due to PR #1145 which made GCC require
  GCC >=10. This patch was previously part of #1161 but has been merged into
  this PR.
* A patch has been applied to ROCm-OpenCL-Runtime to fix a linking error
  in which there were multiple definitions of `ret_val`. This issue is
  highlighted here: https://github.com/ROCm/ROCm-OpenCL-Runtime/issues/113.
  This was previously part #1161 but has been moved into this PR.
* The Dockerfile's `RUN` command (built to layers in the Docker image)
  have been refactored so sources and built objects are deleted in the
  same RUN command as where they were built and installed. This reduces
  the size of the image substantially: from 16.3GB down to 6.6GB.
* The `apt upgrade` has been removed. This step (previously at the start
  of the file) did nothing of importance. Removing it saves both time
  building the image and reduces the size of the image by a small amount.
* `--depth=1` is used when cloning repositories so the entire commit tree
  tree is not pulled each time. This saves some time when building the
  image.
* `apt -y update` has been added  where `apt -y install` is used so
  CACHED image layers do not become an issue in the future if the image
  were to be rebuilt.

### DIFF
--- a/util/dockerfiles/gcn-gpu/Dockerfile
+++ b/util/dockerfiles/gcn-gpu/Dockerfile
@@ -23,108 +23,160 @@
 # THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
 FROM --platform=${BUILDPLATFORM} ubuntu:20.04
 ENV DEBIAN_FRONTEND=noninteractive
-RUN apt -y update && apt -y upgrade && \
-    apt -y install build-essential git m4 scons zlib1g zlib1g-dev \
+
+RUN apt -y update && \
+    apt -y install \
+    # General gem5 requirements when building/running in Ubuntu 20.04, also
+    # some utilities needed to compile and obtain the right packages that
+    # make up this Docker image.
+    build-essential git m4 scons zlib1g zlib1g-dev \
     libprotobuf-dev protobuf-compiler libprotoc-dev libgoogle-perftools-dev \
     python3-dev python-is-python3 doxygen libboost-all-dev \
-    libhdf5-serial-dev python3-pydot libpng-dev libelf-dev pkg-config gdb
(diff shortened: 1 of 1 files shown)


BobbyRBruce:
@mattsinc @abmerop : This is something I was playing around with. I wanted to see if I get this image a bit smaller so we could distribute it just a bit easier. This PR reduces the image size by over 50%

However, the big space saving thing in this PR is removing the sources and built code for  `ROCR-Runtime`, `HIP`, `ROCclr`, `ROCm-OpenCL-Runtime`, and `rocBLAS` after they are installed and/or used in setting up the GCN-GPU Docker environment. I don't know if this is ok. Is there a reason the sources and build dirs are kept in this image? Is it safe to remove them from the image? I grepped our repositories and couldn't see any evidence we need these sources to run anything, but I may have missed something.

I built the Docker images this PR creates (at the time of writing) and uploaded it to `ghcr.io/gem5/gcn-gpu-refactor` if you wish to pull it and see it still works as it should (`docker pull ghcr.io/gem5/gcn-gpu-refactor:latest`).

I don't consider this a very important change. Just if it's safe to do, it would be nice to merge it and distribute smaller docker images.

abmerop:
Thanks Bobby!  I believe deleting the source is fine and likely was not done before as the original author might not have been thinking about the image size.  I am testing this now and I'll let you know in a few hours.

Should I plan to close #1161 after that?

abmerop (approved):
The SE mode tests that I have on my end all passed, so approving this

BobbyRBruce:
> The SE mode tests that I have on my end all passed, so approving this

Great! I'm gonna just merge this once the test pass (for some reason when I change the PR title we need to rerun the tests...). I've updated the PR description to point back to #1161 as the original source of these changes, but I see no point in having two PRs instead of one, so let's just merge this and close the other.

mattsinc:
I am away and not able to run tests right now, but if @abmerop is seeing all tests are passing for now, that's fine.  I thought the sources were included for the various DNNMark-type tests when it came to creating the cache files and debugging, but I could be wrong.

PR #1171, opened 2024-05-27
arch-x86: set AF=0 when logical instructions execute

### LINKED ISSUE #1168: arch-x86: logical instructions execute in-order due to flags input dependency on Zaps CCReg
**Describe the bug**
I noticed in the O3 pipeline viewer that logical x86 instructions like AND, OR, XOR, and TEST unexpectedly had a flags register (i.e., a CCReg) as a source operand (i.e., input dependency). This was surprising, because none of these instructions depend on prior flags; they only set flags as an output.

As a result, in the O3 processor, logical instructions _do not execute until the most recent instruction to update flags writes back_. Given that a large fraction of instructions update flags, including other logical instructions, this means that _**logical instructions effectively execute in-order**_ with respect to each other and other flag updates.

The following x86 assembly program demonstrates this behavior:
```
        global _start          
                               
_start:                        
        mov ecx, 1024          
loop:                          
        times 64 test rax, rax 
        dec ecx                
        jnz loop               
        mov eax, 60
        mov edi, 0
        syscall
```
Running this on the O3 processor and viewing the results in the O3 pipeline viewer confirm that the `TEST` instructions all execute in-order, despite there being no true architectural dependencies between subsequent TEST instructions. Also, the IPC reported in m5out/stats.txt is abysmal, only 1.0.

**Affects version**
develop @ [4f6fdbf](https://github.com/gem5/gem5/commit/4f6fdbf8bfc38c7d8f3a6281d676d73b9b8a097f)

**gem5 Modifications**
None

**To Reproduce**
2. Compile the x86 assembly program above: `nasm -felf64 -o test.o test.asm && ld -o test test.o`
3. Run the program with O3 under SE: `./build/X86/gem5.opt configs/deprecated/example/se.py --cpu-type=X86O3CPU --caches -c ./test`
4. View the IPC: `grep ipc m5out/stats.txt`

**Terminal Output**
```sh
$ ./build/X86/gem5.opt configs/deprecated/example/se.py --cpu-type=X86O3CPU --caches -c ./test
...
Exiting @ tick 33675000 because exiting with last active thread context
$ grep ipc m5out/stats.txt
system.cpu.ipc                               1.004989                       # IPC: instructions per cycle (core level) ((Count/Cycle))
system.cpu.commitStats0.ipc                  1.004989                       # IPC: instructions per cycle (thread level) ((Count/Cycle))
```

**Expected behavior**
Given the default configuration of gem5's O3 processor model, we would expect an IPC much much higher than 1 op/cycle.

**Host Operating System**
Ubuntu 22.04

**Host ISA**
X86

**Compiler used**
GCC 11.4.0

**Additional information**
I did some digging, and it looks like the unexpected source flag register of these logical instructions is due to the fact that the micro-ops implementing these macro-ops do not set the AF (adjust flag). For example, the [micro-asm implementation of TEST_R_R](https://github.com/gem5/gem5/blob/4f6fdbf8bfc38c7d8f3a6281d676d73b9b8a097f/src/arch/x86/isa/insts/general_purpose/compare_and_test/test.py#L52)—which consists of a single `and` micro-op—only writes the OF, SF, ZF, PF, and CF flags, but _not_ the AF flag. As a result, the micro-op implicitly depends on the flags register containing the AF flag—specifically, the [Zaps CC register](https://github.com/gem5/gem5/blob/4f6fdbf8bfc38c7d8f3a6281d676d73b9b8a097f/src/arch/x86/regs/ccr.hh#L53).

Note that the value of the [AF flag is architecturally undefined](https://www.felixcloutier.com/x86/test) following a logical instruction like TEST. That's probably why gem5's X86 logical micro-ops never modify the AF register.

However, I ran some experiments on my Intel Alder Lake hardware and it appears that logical instructions simply set the AF register to zero. I will submit a patch shortly that will implement this behavior. This will eliminate logical instructions' dependencies on the ZAPS register.

nmosier:
Fix #1168. Prevent logical instructions like AND, OR, and TEST from having input dependencies on the previous value of the Zaps register (ZF+AF+PF+SF) by having them set AF=0, rather than not modifying AF.

### DIFF
--- a/src/arch/x86/isa/insts/general_purpose/compare_and_test/test.py
+++ b/src/arch/x86/isa/insts/general_purpose/compare_and_test/test.py
@@ -37,26 +37,26 @@ microcode = """
 def macroop TEST_M_R
 {
     ld t1, seg, sib, disp
-    and t0, t1, reg, flags=(OF, SF, ZF, PF, CF)
+    and t0, t1, reg, flags=(OF, SF, ZF, PF, CF, AF)
 };
 
 def macroop TEST_P_R
 {
     rdip t7
     ld t1, seg, riprel, disp
-    and t0, t1, reg, flags=(OF, SF, ZF, PF, CF)
+    and t0, t1, reg, flags=(OF, SF, ZF, PF, CF, AF)
 };
 
 def macroop TEST_R_R
 {
-    and t0, reg, regm, flags=(OF, SF, ZF, PF, CF)
+    and t0, reg, regm, flags=(OF, SF, ZF, PF, CF, AF)
 };
 
 def macroop TEST_M_I
 {
     ld t1, seg, sib, disp
     limm t2, imm
-    and t0, t1, t2, flags=(OF, SF, ZF, PF, CF)
+    and t0, t1, t2, flags=(OF, SF, ZF, PF, CF, AF)
 };
 
 def macroop TEST_P_I
@@ -64,12 +64,12 @@ def macroop TEST_P_I
     rdip t7
     ld t1, seg, riprel, disp
     limm t2, imm
-    and t0, t1, t2, flags=(OF, SF, ZF, PF, CF)
+    and t0, t1, t2, flags=(OF, SF, ZF, PF, CF, AF)
 };
 
 def macroop TEST_R_I
 {
     limm t1, imm
-    and t0, reg, t1, flags=(OF, SF, ZF, PF, CF)
+    and t0, reg, t1, flags=(OF, SF, ZF, PF, CF, AF)
 };
 """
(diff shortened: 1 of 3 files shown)


powerjg:
Looks good to me!

Can you provide the new results of your microbenchmark after this patch?

nmosier:
Sure — the microbenchmark now gets an ipc of 4.3 (versus 1.0 before). I also ran the SPEC2017 intspeed benchmarks ("reference" input size), and saw an average IPC improvement of 2.8% (best case: 641.leela_s, with +6.4%).

powerjg (approved):
Thanks for the info!

I am slightly worried that this is going to cause surprising regressions, but let's give it a try :)

PR #1173, opened 2024-05-28
arch-riscv: Add rvZext to BranchTarget

rogerchang23424:
Ensure the upper xlen bits are all zeros

Change-Id: Id81330eced907d21320bc1af85ad38fb6e95f6b1

### DIFF
--- a/src/arch/riscv/isa/formats/compressed.isa
+++ b/src/arch/riscv/isa/formats/compressed.isa
@@ -173,7 +173,7 @@ def template CJumpExecute {{
     %(class_name)s::branchTarget(ThreadContext *tc) const
     {
         PCStateBase *pc_ptr = tc->pcState().clone();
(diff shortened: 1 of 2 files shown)


powerjg (approved):
This seems fine to me, but I would like to understand the bug that it is fixing. Why do you need to zero extend the PC? Shouldn't the register already be zero extended?

hnpl (approved):

rogerchang23424:
Hi @powerjg,

The type of `pc` and `npc` is `Addr`(`uint64_t)
https://github.com/gem5/gem5/blob/7fa0342a7c7982967c948b346ba7077dce54efd2/src/arch/generic/pcstate.hh#L257-L282

Most of write-back values to register are sign-ext. We should zero-ext the address computation results to ensure the upper xlen bits are zeros.

PR #1179, opened 2024-05-29
mem-cache: Fix maybe-uninitialized warning

ylldummy:
When compiler tries to inline a vector construction with a default value as default constructed ReplaceableEntry. It can complain about the uninitialized member.

Let's provide basic initialization to the members.

Example codepath:
 SignaturePathV2 constructor
 -> GlobalHistoryEntry() as init_value to AssociativeSet
 -> AssociativeSet initialize vector<Entry> with init_value

### DIFF
--- a/src/mem/cache/replacement_policies/replaceable_entry.hh
+++ b/src/mem/cache/replacement_policies/replaceable_entry.hh
@@ -73,7 +73,7 @@ class ReplaceableEntry
     uint32_t _way;
 
   public:
-    ReplaceableEntry() = default;
+    ReplaceableEntry() : _set(0), _way(0) {}
     virtual ~ReplaceableEntry() = default;
 
     /**


src/mem/cache/replacement_policies/replaceable_entry.hh
```diff
@@ -65,12 +65,12 @@ class ReplaceableEntry
     /**
      * Set to which this entry belongs.
      */
-    uint32_t _set;
+    uint32_t _set = 0;
```
odanrc: Nit: use list initialization instead of assignment

src/mem/cache/replacement_policies/replaceable_entry.hh
```diff
@@ -65,12 +65,12 @@ class ReplaceableEntry
...
 
     /**
      * Way (relative position within the set) to which this entry belongs.
      */
-    uint32_t _way;
+    uint32_t _way = 0;
```
odanrc: Nit: ditto

odanrc (approved):

src/mem/cache/replacement_policies/replaceable_entry.hh
ylldummy: Fixed in the new push.

src/mem/cache/replacement_policies/replaceable_entry.hh
ylldummy: Fixed in the new push.

BobbyRBruce (approved):

PR #1182, opened 2024-05-30
cpu-o3: Do not set Executed on load instruction to be replayed

jjuninho:
A load instruction can be replayed when
1) it's strictly ordered or
2) it falls into load-store forwarding mismatch.

Case 1 was considered in executeLoad function but the case 2 wasn't. It causes the case-2 replayed load instruction to violate the assertion condition "assert(!load_inst->isExecuted())" in LSQUnit::read. This commit fixes the problem by adding consideration of the case 2 in LSQUnit::executeLoad.

### DIFF
--- a/src/cpu/o3/lsq_unit.cc
+++ b/src/cpu/o3/lsq_unit.cc
@@ -611,6 +611,12 @@ LSQUnit::executeLoad(const DynInstPtr &inst)
     if (inst->isTranslationDelayed() && load_fault == NoFault)
         return load_fault;
 
+    // Partial Store-to-Load Forwarding condition marks the load to be
+    // reissued during LSQUnit::read(). In this case we shouldn't notify
+    // iewStage that the instruction is ready for commit.
+    if (!inst->isIssued() && !inst->effAddrValid())
+        return load_fault;
+
     if (load_fault != NoFault && inst->translationCompleted() &&
             inst->savedRequest->isPartialFault()
             && !inst->savedRequest->isComplete()) {


powerjg:
This makes sense to me. Could you describe how you tested it?

jjuninho:
Hi @powerjg,
Below is the test description.
- Workload: Geekbench6 Clang Single-Core (aarch64)
- Simulation mode: Syscall emulation
- HW config: Proprietary SoC which has 
    - 6 ArmO3CPUs and 4 ArmMinorCPUs, 
    - LPDDR5 DRAM
    - Modified Ruby CHI network
- Symptom: `assert(!load_inst->isExecuted())` assertion failure.
- Clue: `LSQUnit` debug trace showed that the assertion failure occurs right after "Load-store forwarding mis-match" and the load instruction is rescheduled in that case.

With this fix, the assertion failure is gone.

powerjg:
@giactra This change makes sense to me, but I'm not an expert at the O3CPU. Could you take a look just to get a second pair of eyes on the change?

@jjuninho thanks for the info! I'm being a little more careful than normal with this change because, if wrong, it could cause difficult to track down bugs in other simulations.

pranith:
@jjuninho Can you please explain what you mean by "load-store forwarding mis-match"? There are quite a few loads that will not get forwarding data from the stores. I am trying to understand why you are hitting this assert now. Are there any other changes that are causing this assert to fire? Can you please post the logs and stack trace?

jjuninho:
@pranith 
1. Load-store forwarding mis-match.
I meant the line below. The condition is when the address range of the load instruction partially covered by a store in the store queue.
https://github.com/gem5/gem5/blob/3b7307182f272756bd805ca2d5f4af79efa639a0/src/cpu/o3/lsq_unit.cc#L1558C17-L1560C72


2. Debug trace log
Below is the `LSQUnit` debug trace.
```
20408206907: system.chips.clusters.cpu_groups0.cpus.iew.lsq: received Load req vaddr: 0x4902ff8, paddr: 0x8494bff8, pc: 842b64
20408206907: system.chips.clusters.cpu_groups0.cpus.iew.lsq.thread0: Read called, load idx: 19669349, store idx: 17846627, storeHead: 17846590 addr: 0 split
20408206907: system.chips.clusters.cpu_groups0.cpus.iew.lsq.thread0: Load-store forwarding mis-match. Store idx 17846609 to load addr 0x4902ff8
20408206907: system.chips.clusters.cpu_groups0.cpus.iew.lsq.thread0: Load [sn:116166214] not executed from fault
20408206907: system.chips.clusters.cpu_groups0.cpus.iew.lsq.thread0: executeLoad setExecuted 2 PC (0x842b64=>0x842b68).(1=>2), [sn:116166214], hasRequest: 1, strictlyOrdered: 0, atCommit: 0, fault=0x5aac500e44b0
```
In the log above, the last line is what I added to my local to check at which point the "Executed" flag is set. It was set at
https://github.com/gem5/gem5/blob/3b7307182f272756bd805ca2d5f4af79efa639a0/src/cpu/o3/lsq_unit.cc#L639C1-L639C33

The log below is the last debug print before the assertion failure.
```
20408209098: system.chips.clusters.cpu_groups0.cpus.iew.lsq: received Load req vaddr: 0x4902ff8, paddr: 0x8494bff8, pc: 842b64
```
And below is the assertion failure message.
```
gem5.opt: src/cpu/o3/lsq_unit.cc:1344: gem5::Fault gem5::o3::LSQUnit::read(gem5::o3::LSQUnit::LSQRequest*, ssize_t): Assertion `!load_inst->isExecuted()' failed.
```
The line number is different from the mainline develop branch since I made a change to LSQUnit for assigning QoS value for demand access packet. I'm pretty sure that it doesn't affect this problem.

3. gdb call stack at the assertion point
```
#6  0x00007ffff6a39e96 in __GI___assert_fail (assertion=0x55555941895f "!load_inst->isExecuted()", file=0x55555941723a "src/cpu/o3/lsq_unit.cc", line=1333, 
    function=0x555559418908 "gem5::Fault gem5::o3::LSQUnit::read(gem5::o3::LSQUnit::LSQRequest*, ssize_t)") at ./assert/assert.c:101
#7  0x000055555899d3d5 in gem5::o3::LSQUnit::read (this=0x555563723900, request=0x55567491cd80, load_idx=19669338) at src/cpu/o3/lsq_unit.cc:1333
#8  0x000055555898b98e in gem5::o3::LSQ::read (this=0x55555e679880, request=0x55567491cd80, load_idx=19669338) at src/cpu/o3/lsq.cc:1554
#9  0x00005555589870a1 in gem5::o3::LSQ::pushRequest (this=0x55555e679880, inst=..., isLoad=true, data=0x0, size=16, addr=76558328, flags=..., res=0x0, 
    amo_op=std::unique_ptr<gem5::AtomicOpFunctor> = {...}, byte_enable=std::vector<bool> of length 16, capacity 64 = {...}) at src/cpu/o3/lsq.cc:862
#10 0x000055555894ec62 in gem5::o3::CPU::pushRequest (this=0x55555e676000, inst=..., isLoad=true, data=0x0, size=16, addr=76558328, flags=..., res=0x0, 
    amo_op=std::unique_ptr<gem5::AtomicOpFunctor> = {...}, byte_enable=std::vector<bool> of length 16, capacity 64 = {...}) at src/cpu/o3/cpu.hh:556
#11 0x000055555894db77 in gem5::o3::DynInst::initiateMemRead (this=0x55566428c500, addr=76558328, size=16, flags=..., 
    byte_enable=std::vector<bool> of length 16, capacity 64 = {...}) at src/cpu/o3/dyn_inst.cc:423
#12 0x00005555573f007d in gem5::initiateMemRead<gem5::ExecContext> (xc=0x55566428c500, addr=76558328, size=16, flags=..., 
    byte_enable=std::vector<bool> of length 16, capacity 64 = {...}) at src/arch/generic/memhelpers.hh:59
#13 0x00005555573ed0a8 in gem5::initiateMemRead<gem5::ExecContext, std::array<unsigned long, 2ul> > (xc=0x55566428c500, traceData=0x0, addr=76558328, mem=..., flags=...)
    at src/arch/generic/memhelpers.hh:72
#14 0x00005555573298d2 in gem5::ArmISAInst::MicroLdPairUop::initiateAcc (this=0x555670cd5a00, xc=0x55566428c500, traceData=0x0)
    at build/ARM/arch/arm/generated/exec-ns.cc.inc:68702
#15 0x000055555894d8b6 in gem5::o3::DynInst::initiateAcc (this=0x55566428c500) at src/cpu/o3/dyn_inst.cc:379
#16 0x00005555589977b0 in gem5::o3::LSQUnit::executeLoad (this=0x555563723900, inst=...) at src/cpu/o3/lsq_unit.cc:605
#17 0x0000555558984243 in gem5::o3::LSQ::executeLoad (this=0x55555e679880, inst=...) at src/cpu/o3/lsq.cc:242
#18 0x000055555896d5b4 in gem5::o3::IEW::executeInsts (this=0x55555e6777e8) at src/cpu/o3/iew.cc:1176
#19 0x000055555896f224 in gem5::o3::IEW::tick (this=0x55555e6777e8) at src/cpu/o3/iew.cc:1422
#20 0x000055555892ae1d in gem5::o3::CPU::tick (this=0x55555e676000) at src/cpu/o3/cpu.cc:363
#21 0x0000555558928311 in operator() (__closure=0x55555e676470) at src/cpu/o3/cpu.cc:75
#22 0x000055555893208c in std::__invoke_impl<void, gem5::o3::CPU::CPU(const gem5::BaseO3CPUParams&)::<lambda()>&>(std::__invoke_other, struct {...} &) (__f=...)
    at /usr/include/c++/11/bits/invoke.h:61
#23 0x0000555558931e1e in std::__invoke_r<void, gem5::o3::CPU::CPU(const gem5::BaseO3CPUParams&)::<lambda()>&>(struct {...} &) (__fn=...) at /usr/include/c++/11/bits/invoke.h:111
#24 0x0000555558931c18 in std::_Function_handler<void(), gem5::o3::CPU::CPU(const gem5::BaseO3CPUParams&)::<lambda()> >::_M_invoke(const std::_Any_data &) (__functor=...)
    at /usr/include/c++/11/bits/std_function.h:290
#25 0x0000555555d749f0 in std::function<void ()>::operator()() const (this=0x55555e676470) at /usr/include/c++/11/bits/std_function.h:590
#26 0x0000555555dff6b4 in gem5::EventFunctionWrapper::process (this=0x55555e676438) at src/sim/eventq.hh:1162
#27 0x0000555557d5aab1 in gem5::EventQueue::serviceOne (this=0x55555b2c4a20) at src/sim/eventq.cc:249
#28 0x0000555557d83535 in gem5::doSimLoop (eventq=0x55555b2c4a20) at src/sim/simulate.cc:323
#29 0x0000555557d82fd1 in gem5::simulate (num_cycles=18446744073709551615) at src/sim/simulate.cc:229
```

pranith:
Thanks for the logs and traces @jjuninho. I will go through them.

Just to be sure, can you please try to reproduce this assertion without any local changes? This is a common scenario as you described and we should have hit the assert long before if what you describe is what is happening.

jjuninho:
@pranith 
I reproduced the problem with the mainline gem5 code and pushed a **temporary commit** to to reproduce it. **I will remove the temporary commit after review.**

I added a test program which causes this problem. What the program does is,
1) Perform 16B store to the end of 4KiB page.
2) Make a 16B load with +8 byte offset with the preceding store. This will cause "load-store forwarding mismatch" and the page fault as well.

I also modified starter_se.py to use ex5_big CPU config instead of O3_ARM_v7a.

How to reproduce:
1) Build the `test_arm_lsq.c` to aarch64 target: go to tests/test-progs/test-arm-lsq then `make`
2) Revert the fix 6c0c6b3d5f1c8a3e726cf81854859a778649bdd2
3) scons build/ARM/gem5.opt
4) `build/ARM/gem5.opt configs/example/arm/starter_se.py --cpu o3 --cpu-freq 2GHz --num-cores 1 --mem-type LPDDR5_6400_1x16_BG_BL32 --mem-channels 4 --mem-ranks 2 tests/test-progs/test-arm-lsq/test_arm_lsq`


The problem is that I see an assertion from other place even with 6c0c6b3d5f1c8a3e726cf81854859a778649bdd2 from commit.cc. So I'm changing this PR to draft until I solve the other assertion.

pranith:
Thanks @jjuninho. I will take a look this week.

src/cpu/o3/lsq_unit.cc
```diff
@@ -626,15 +626,21 @@ LSQUnit::executeLoad(const DynInstPtr &inst)
...
             inst->forwardOldRegs();
         DPRINTF(LSQUnit, "Load [sn:%lli] not executed from %s\n",
                 inst->seqNum,
                 (load_fault != NoFault ? "fault" : "predication"));
-        if (!(inst->hasRequest() && inst->strictlyOrdered()) ||
+        // LSQUnit clears Issued flag after calling rescheduleMemInst and
+        // CanIssue flag is cleared in the rescheduleMemInst.
+        // If either of them are set then it isn't rescheduled one.
+        if (((inst->isIssued() || inst->readyToIssue()) &&
+             !(inst->hasRequest() && inst->strictlyOrdered())) ||
```
pranith: We shouldn't be notifying the iewStage that this instruction is ready for commit since it will be reissued from the IQ.

pranith:
This seems to be a partial STLF case where some of the bytes of the load are coming from a previous store and the rest of the bytes are coming from the cache. 

For these cases, we don't want to notify iewStage that this load is ready for commit yet. Something like the following should work instead.

```
diff --git a/src/cpu/o3/lsq_unit.cc b/src/cpu/o3/lsq_unit.cc
index 68fd464627..6cd874e90f 100644
--- a/src/cpu/o3/lsq_unit.cc
+++ b/src/cpu/o3/lsq_unit.cc
@@ -611,6 +611,11 @@ LSQUnit::executeLoad(const DynInstPtr &inst)
     if (inst->isTranslationDelayed() && load_fault == NoFault)
         return load_fault;
 
+    // A partial STLF condition will mark the load to be reissued
+    if (!inst->isIssued() && !inst->effAddrValid()) {
+        return load_fault;
+    }
+
     if (load_fault != NoFault && inst->translationCompleted() &&
             inst->savedRequest->isPartialFault()
             && !inst->savedRequest->isComplete()) {
```

jjuninho:
@pranith Thank you! Now I can see my original fix was very wrong. I was struggling to fix all the assertions at the commit stage.
I created a commit according to your comments.

pranith (approved):

powerjg (approved):
Thanks for the help @pranith!

PR #1187, opened 2024-05-30
arch-riscv: correctly set dynamic VLEN for all arith instructions 

saul44203:
Some arithmetic instructions of the riscv vector extension where still using the default VLEN=256 instead of the dynamic one through the inherited `vlen` attribute. Most of them only use this to calculate the effective index for the mask element like so:

```
if (this->vm || elem_mask(v0, ei)) {
...
```

This means that instructions will wrongly compute the mask index in the second and subsequent micro instructions (`microIdx` > 0). This commit fixes this by adding the corresponding `set_vlen` snippet to the affected instruction formats.

Change-Id: Ib041de972d6938490741a9fb4c214a6a5172c34e

### DIFF
--- a/src/arch/riscv/isa/formats/vector_arith.isa
+++ b/src/arch/riscv/isa/formats/vector_arith.isa
@@ -209,6 +209,7 @@ def format VectorIntFormat(code, category, *flags) {{
         vm_decl_rd = vmDeclAndReadData()
 
     set_vlenb = setVlenb()
+    set_vlen = setVlen() if need_elem_idx else ""
 
     microiop = InstObjParams(name + "_micro",
         Name + "Micro",
@@ -217,6 +218,7 @@ def format VectorIntFormat(code, category, *flags) {{
          'set_dest_reg_idx': set_dest_reg_idx,
          'set_src_reg_idx': set_src_reg_idx,
          'set_vlenb' : set_vlenb,
+         'set_vlen' : set_vlen,
          'vm_decl_rd': vm_decl_rd,
          'copy_old_vd': copyOldVd(old_vd_idx),
          'declare_varith_template': declareVArithTemplate(Name + "Micro")},
@@ -484,6 +486,7 @@ def format VectorIntMaskFormat(code, category, *flags) {{
         vm_decl_rd = vmDeclAndReadData()
 
     set_vlenb = setVlenb()
+    set_vlen = setVlen() if need_elem_idx else ""
 
     microiop = InstObjParams(name + "_micro",
         Name + "Micro",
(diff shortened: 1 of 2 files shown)


powerjg (dismissed):

hnpl:
@saul44203 Thanks for the fix!

Along with the above change, I think we should remove this default value for `vlen` in the `VectorMicroInst` constructor.

https://github.com/gem5/gem5/blob/3b7307182f272756bd805ca2d5f4af79efa639a0/src/arch/riscv/insts/vector.hh#L143-L149

saul44203:
> @saul44203 Thanks for the fix!
> 
> Along with the above change, I think we should remove this default value for `vlen` in the `VectorMicroInst` constructor.
> 
> https://github.com/gem5/gem5/blob/3b7307182f272756bd805ca2d5f4af79efa639a0/src/arch/riscv/insts/vector.hh#L143-L149

I think it makes sense to remove it but there seems to be child classes that do not pass the `vlen` explicitly to the parent constructor and they also do not include it as a parameter in their own.

What should we do?

hnpl:
> > @saul44203 Thanks for the fix!
> > Along with the above change, I think we should remove this default value for `vlen` in the `VectorMicroInst` constructor.
> > https://github.com/gem5/gem5/blob/3b7307182f272756bd805ca2d5f4af79efa639a0/src/arch/riscv/insts/vector.hh#L143-L149
> 
> I think it makes sense to remove it but there seems to be child classes that do not pass the `vlen` explicitly to the parent constructor and they also do not include it as a parameter in their own.
> 
> What should we do?

I see. I think we should merge this change and make an issue to refactor the codebase to remove the default vlen.

saul44203:
> > > @saul44203 Thanks for the fix!
> > > Along with the above change, I think we should remove this default value for `vlen` in the `VectorMicroInst` constructor.
> > > https://github.com/gem5/gem5/blob/3b7307182f272756bd805ca2d5f4af79efa639a0/src/arch/riscv/insts/vector.hh#L143-L149
> > 
> > 
> > I think it makes sense to remove it but there seems to be child classes that do not pass the `vlen` explicitly to the parent constructor and they also do not include it as a parameter in their own.
> > What should we do?
> 
> I see. I think we should merge this change and make an issue to refactor the codebase to remove the default vlen.

Done. Now it won't compile though. Should I set this as a draft until the refactor is done?

hnpl:
I think it's more important to merge the change before removing `vlen=256`, so I think we can merge this first and have the "removing `vlen=256`" change as a separate PR.

Thanks for fixing this!

ivanaamit (approved):

PR #1198, opened 2024-06-04
arch-arm: avoid using an uninitialized variable use in MMU walks

arichardson:
While running a simple Arm32 binary, I noticed that all memory transactions were being marked as NS instead of S once I turn on the MMU (even though the page tables have the NS bit set to zero). The result of this was that semihosting calls were failing since they were using functional accesses with the SECURE flag set, but the caches only contained NS tagged entries so these accesses always read stale values from DRAM.

Digging through the Arm MMU code it appears that the NS bit lookup was being keyed of the `secureLookup` flag which is only used for long descriptors. I believe 0c28712f51f5b0748c3afd5b287969ceb615ea8d should have used isSecure instead of secureLookup. To avoid using these uninitialized values in the future I wrapped the LPAE state in a std::optional to ensure that it is only accessed once initialized.

Change-Id: Ibc406ed3f4cfa768f470e34a5eca3c1a2bf45cd8

### DIFF
--- a/src/arch/arm/table_walker.hh
+++ b/src/arch/arm/table_walker.hh
@@ -890,17 +890,24 @@ class TableWalker : public ClockedObject
 
         /** If the access comes from the secure state. */
         bool isSecure;
+        /** Whether lookups should be treated as using the secure state.
+         * This is usually the same as isSecure, but can be set to false by the
+         * long descriptor table attributes. */
+        bool secureLookup = false;
 
         /** True if table walks are uncacheable (for table descriptors) */
         bool isUncacheable;
 
         /** Helper variables used to implement hierarchical access permissions
-         * when the long-desc. format is used (LPAE only) */
-        bool secureLookup;
-        bool rwTable;
-        bool userTable;
-        bool xnTable;
-        bool pxnTable;
+         * when the long-desc. format is used. */
+        struct LongDescData
+        {
+            bool rwTable = false;
+            bool userTable = false;
+            bool xnTable = false;
+            bool pxnTable = false;
+        };
+        std::optional<LongDescData> longDescData;
 
         /** Hierarchical access permission disable */
         bool hpd;
--- a/src/arch/arm/table_walker.cc
+++ b/src/arch/arm/table_walker.cc
@@ -132,9 +132,8 @@ TableWalker::WalkerState::WalkerState() :
     sctlr(0), scr(0), cpsr(0), tcr(0),
     htcr(0), hcr(0), vtcr(0),
     isWrite(false), isFetch(false), isSecure(false),
-    isUncacheable(false),
-    secureLookup(false), rwTable(false), userTable(false), xnTable(false),
-    pxnTable(false), hpd(false), sh(0), irgn(0), orgn(0), stage2Req(false),
+    isUncacheable(false), longDescData(std::nullopt),
+    hpd(false), sh(0), irgn(0), orgn(0), stage2Req(false),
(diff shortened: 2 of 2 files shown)


giactra:
Thanks @arichardson for spotting a problem. Just to be clear, this error should arise only when short-descriptors are in use, so I presume you are purposely running an AArch32 trusted OS without LPAE (just checking this is the intended configuration).

The patch introducing the bug https://github.com/gem5/gem5/commit/0c28712f51f5b0748c3afd5b287969ceb615ea8d is not necessarily using the wrong variable; it was wrongly assuming secureLookup was initialized for short descriptor as well.
While your fix makes perfect sense, I like the idea of using the same variable wrt to long descriptors and to decouple the security state of the CPU from the security state of the access. So what I am proposing is to keep using secureLookup for short descriptors and to initialize it accordingly with

` currState->secureLookup = currState->isSecure`

This also means using secureLookup [here](https://github.com/gem5/gem5/blob/stable/src/arch/arm/table_walker.cc#L675)

With respect of the std::optional addition, it looks like a good idea to me, though I would rename the instance from lpae (which is VMSAv8-32 only concept) to longDescData (or something similar). Please let me know what you think about this

arichardson:
> Thanks @arichardson for spotting a problem. Just to be clear, this error should arise only when short-descriptors are in use, so I presume you are purposely running an AArch32 trusted OS without LPAE (just checking this is the intended configuration).

Yes exactly, I was running a baremetal picolibc binary on AArch32, where the python configuration starts up in secure mode.

> 
> The patch introducing the bug [0c28712](https://github.com/gem5/gem5/commit/0c28712f51f5b0748c3afd5b287969ceb615ea8d) is not necessarily using the wrong variable; it was wrongly assuming secureLookup was initialized for short descriptor as well. While your fix makes perfect sense, I like the idea of using the same variable wrt to long descriptors and to decouple the security state of the CPU from the security state of the access. So what I am proposing is to keep using secureLookup for short descriptors and to initialize it accordingly with
> 
> ` currState->secureLookup = currState->isSecure`
> 
> This also means using secureLookup [here](https://github.com/gem5/gem5/blob/stable/src/arch/arm/table_walker.cc#L675)
> 

Ah that makes a lot more sense - I just looked at the comment above the variable and assumed it was used incorrectly. Will update the PR to apply your suggestions.

> With respect of the std::optional addition, it looks like a good idea to me, though I would rename the instance from lpae (which is VMSAv8-32 only concept) to longDescData (or something similar). Please let me know what you think about this

That sounds good to me, will make that change.

arichardson:
@giactra Thanks for the review, the new version should hopefully address your comments. I decided to always use the secureLookup flag for the Request::Flags since it is now the same in all these cases. Hopefully that is correct and I didn't miss something.

src/arch/arm/table_walker.hh
```diff
@@ -264,7 +264,7 @@ class TableWalker : public ClockedObject
         bool
         secure(bool have_security, WalkerState *currState) const override
         {
-            if (have_security && currState->secureLookup) {
```
giactra: Nit: this should be kept as secureLookup for consistency

giactra (commented):
Thanks @arichardson for your changes, LGTM, only one minor thing and then we are ready to merge this

src/arch/arm/table_walker.hh
arichardson: Done, somehow missed that in the last update.

giactra (approved):
Thanks for your changes!

PR #1200, opened 2024-06-05
arch-arm,mem: Don't hardcode secure mode accesses for semihosting

arichardson:
When accessing memory using functionalAccess(), the MMU could tell us to use a nonsecure access even though the CPU is operating in secure mode. I noticed this while trying to run a simple semihosting hello world with the MMU+caches enabled and the semihosting calls ended up reading from memory instead of the caches due to an S/NS mismatch.

See also https://github.com/gem5/gem5/pull/1198 which happens to also mask the issue I saw, but I believe both changes are needed.

Change-Id: I9e6b9839b194fbd41938e2225449c74701ea7fee

### DIFF
--- a/src/arch/arm/mmu.cc
+++ b/src/arch/arm/mmu.cc
@@ -813,7 +813,8 @@ MMU::translateMmuOff(ThreadContext *tc, const RequestPtr &req, Mode mode,
     // security state of the processor
     if (state.isSecure)
         req->setFlags(Request::SECURE);
-
+    else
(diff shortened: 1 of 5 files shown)


giactra:
Thanks for these fixes! It looks good to me, I would only amend the commit title (and PR) to include the "mem" tag as well (so: arch-arm,mem) as it touches the translation port proxy which is used by every ISA

giactra (approved):

ivanaamit (approved):

PR #1213, opened 2024-06-10
gpu-compute: Added functions to choose replacement policies for GPU

yuxiaojia:
Adding RP_choose functions to change replacement policies among
TreePLRU, LRU, FIFO, LFU, LIP, MRU, NRU, RRIP, SecondChance AND ShiPMem replacement 
policies  for TCC, TCP and SQC caches for GPU

### DIFF
--- a/configs/example/apu_se.py
+++ b/configs/example/apu_se.py
@@ -380,6 +380,16 @@ parser.add_argument(
     help="Gfx version for gpuNote: gfx902 is not fully supported by ROCm",
 )
 
+parser.add_argument("--tcp-rp", type=str, default="TreePLRURP", help="cache replacement policy"
+                    "policy for tcp")
+
+parser.add_argument("--tcc-rp", type=str, default="TreePLRURP", help="cache replacement policy"
+                    "policy for tcc")
+
+# sqc rp both changes sqc rp and scalar cache rp
+parser.add_argument("--sqc-rp", type=str, default="TreePLRURP", help="cache replacement policy"
+                    "policy for sqc")
+
 Ruby.define_options(parser)
 
 # add TLB options to the parser
--- a/configs/example/apu_se.py
+++ b/configs/example/apu_se.py
@@ -892,9 +892,9 @@ gpu_port_idx = gpu_port_idx - args.num_cp * 2
 token_port_idx = 0
 for i in range(len(system.ruby._cpu_ports)):
     if isinstance(system.ruby._cpu_ports[i], VIPERCoalescer):
-        system.cpu[shader_idx].CUs[token_port_idx].gmTokenPort = (
-            system.ruby._cpu_ports[i].gmTokenPort
-        )
+        system.cpu[shader_idx].CUs[
+            token_port_idx
+        ].gmTokenPort = system.ruby._cpu_ports[i].gmTokenPort
         token_port_idx += 1
 
 wavefront_size = args.wf_size
--- a/configs/ruby/GPU_VIPER.py
+++ b/configs/ruby/GPU_VIPER.py
@@ -1132,8 +1132,6 @@ def RP_choose(test_name):
         replacement_policy = LRURP()
     elif test_name == 'FIFORP':
         replacement_policy = FIFORP()
-    elif test_name == 'LRURP':
-        replacement_policy = LRURP()
     elif test_name == 'LFURP':
         replacement_policy = LFURP()
     elif test_name == 'LIPRP':
--- a/configs/ruby/GPU_VIPER.py
+++ b/configs/ruby/GPU_VIPER.py
@@ -149,7 +149,7 @@ class TCPCache(RubyCache):
         self.size = MemorySize(options.tcp_size)
         self.assoc = options.tcp_assoc
         self.resourceStalls = options.no_tcc_resource_stalls
-        self.replacement_policy = TreePLRURP()
+        self.replacement_policy = RP_choose(options.tcp_rp)
 
 
 class TCPCntrl(TCP_Controller, CntrlBase):
@@ -241,7 +241,7 @@ class SQCCache(RubyCache):
     def create(self, options):
         self.size = MemorySize(options.sqc_size)
         self.assoc = options.sqc_assoc
-        self.replacement_policy = TreePLRURP()
+        self.replacement_policy = RP_choose(options.tcp_rp)
 
 
 class SQCCntrl(SQC_Controller, CntrlBase):
@@ -303,7 +303,7 @@ class TCC(RubyCache):
         self.start_index_bit = math.log(options.cacheline_size, 2) + math.log(
             options.num_tccs, 2
         )
-        self.replacement_policy = TreePLRURP()
+        self.replacement_policy = RP_choose(options.tcp_rp)
 
 
 class TCCCntrl(TCC_Controller, CntrlBase):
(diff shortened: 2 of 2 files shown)


mattsinc:
@yuxiaojia : you will need to update the commit message to use the appropriate format and have a ChangeID before this can be merged in

See https://www.gem5.org/contributing "Contributing" for details.

mattsinc (commented):
In terms of code, this seems fine -- but like I said you won't be able to merge until you fix commit message and add a ChangeID

yuxiaojia:
> In terms of code, this seems fine -- but like I said you won't be able to merge until you fix commit message and add a ChangeID

I will try to fix it right now

mattsinc (approved):
@abmerop would be great if you can check to, but maybe not needed

abmerop:
You need to change the GPUFS file too.  This will (did) break GPUFS.

mattsinc:
Thanks @abmerop -- @yuxiaojia can you please make the equivalent changes there?

mattsinc:
Shoot, I just realize that I merged this already -- @yuxiaojia please make a separate commit updating the GPUFS file then.  Do this as your first priority since it is breaking GPUFS.

abmerop:
I put a commit here: https://github.com/gem5/gem5/pull/1230

configs/example/apu_se.py
```diff
@@ -380,6 +380,28 @@
     help="Gfx version for gpuNote: gfx902 is not fully supported by ROCm",
 )
 
+parser.add_argument(
+    "--tcp-rp",
+    type=str,
+    default="TreePLRURP",
+    help="cache replacement policy" "policy for tcp",
```
odanrc, line 387: I am very late to the party, but this is how the PR should have been implemented to avoid hardcoding:

Add this line to each of these new arguments:
        choices=ObjectList.rp_list.get_names(),

configs/ruby/GPU_VIPER.py
```diff
@@ -149,7 +149,7 @@ def create(self, options):
         self.size = MemorySize(options.tcp_size)
         self.assoc = options.tcp_assoc
         self.resourceStalls = options.no_tcc_resource_stalls
-        self.replacement_policy = TreePLRURP()
+        self.replacement_policy = RP_choose(options.tcp_rp)
```
odanrc, line 152: Replace these lines using RP_choose by:
self.replacement_policy = ObjectList.rp_list.get(options.tcp_rp)

configs/ruby/GPU_VIPER.py
```diff
@@ -1124,3 +1124,28 @@ def create_system(
     ruby_system.network.number_of_virtual_networks = 11
 
     return (cpu_sequencers, dir_cntrl_nodes, mainCluster)
+
+
+def RP_choose(test_name):
```
odanrc, line 1129: This is a hardcoded version of gem5/configs/ObjectList.py::rp_list

odanrc:
I can´t really run gem5 in my machine to test a commit, so I won´t risk breaking stuff uploading a PR with the changes I propose, but if you can easily test this here are my suggestions on how to make a PR to improve this code by removing code duplication and hardcoded strings.

mattsinc:
@yuxiaojia : can you please make a new commit that incorporates @odanrc 's suggestions (and of course test them first)?

yuxiaojia:
> @yuxiaojia : can you please make a new commit that incorporates @odanrc 's suggestions (and of course test them first)?

Hi Profess Sinclair, I tried to make the change as suggested but seems like it caused an error that "could not create sytem for ruby protocol GPU_VIPER." I am currently looking into this problem.

mattsinc:
> > @yuxiaojia : can you please make a new commit that incorporates @odanrc 's suggestions (and of course test them first)?
> 
> Hi Profess Sinclair, I tried to make the change as suggested but seems like it caused an error that "could not create sytem for ruby protocol GPU_VIPER." I am currently looking into this problem.

Ok, feel free to message me or post here if you need help.

configs/ruby/GPU_VIPER.py
yuxiaojia: > This is a hardcoded version of gem5/configs/ObjectList.py::rp_list

Hi, from what my understanding now, the change does not work since, for example, if we want to make self.replacement_policy = TreePLRURP(), however the self.replacement_policy = ObjectList.rp_list.get(options.tcp_rp) will not make the same change to call the class TreePLRURP(), since it would generate an error that "could not create sytem for ruby protocol GPU_VIPER". Please let me know if it makes sense, or I understand the get() function incorrectly.

odanrc: I may not have understood your question. Are you saying that you wanted to use ObjectList.rp_list.get() to replace the assignments ´self.replacement_policy = TreePLRURP()´? If so, then you would probably have to do it like ´ObjectList.rp_list.get("TreePLRURP")´, IIRC, but I have never tried doing that. Regardless, this line is not the major focus of my suggestion for deduplication, so it does not need to be changed to use the object list. Are you having trouble making the line ´self.replacement_policy = ObjectList.rp_list.get(options.tcp_rp)´ work?

yuxiaojia: Yes, I am not able to make the line  ´self.replacement_policy = ObjectList.rp_list.get(options.tcp_rp)´  work now since it will not create the ruby system as error said, and that TreePLRURP is an example I had

odanrc: Could you give me a command line to run a test case so that I can try creating the PR?

yuxiaojia: Sorry for the late response. You could run an example command line to both use RRIP for TCP and TCC:

`docker run --volume $(pwd):$(pwd) -w $(pwd) docker.io/library/gccimage build/VEGA_X86/gem5.opt  configs/example/apu_se.py -n 3 --dgpu --gfx-version=gfx900 --num-compute-units=4 --cu-per-sa=4 --num-gpu-complex=1 --reg-alloc-policy=dynamic --num-tccs=8 --num-dirs=64 --mem-size=16GB --mem-type=HBM_2000_4H_1x64 --vreg-file-size=16384 --sreg-file-size=800 --tcc-size=4MB --gpu-clock=1801MHz --ruby-clock=1000MHz --vrf_lm_bus_latency=6 --mem-req-latency=69 --mem-resp-latency=69 --mandatory_queue_latency=1 --max-cu-tokens=160 --max-coalesces-per-cycle=10 --sqc-size=16kB --tcp-size=4MB --scalar-mem-req-latency=28 --TCP_latency=4 --tcp-assoc=16 --tcp-num-banks=16 --TCC_latency=121 --tcc-assoc=16 --tcc-tag-access-latency=1 --tcc-data-access-latency=1 --tcp-rp=RRIPRP --WB_L2 --tcc-rp=RRIPRP -c TESTNAME`

The docker.io/library/gccimage can be built by 
```
cd util/dockerfiles/gcn-gpu/
docker build -t gccimage .
```

odanrc: Thanks for the command. I could not run the docker image at all (could not find shared object for python3.10), but when I ran the command manually I found the issue.

Where I said to put:
 self.replacement_policy = ObjectList.rp_list.get(options.tcp_rp)
Should have been (the last 2 characters were missing):
 self.replacement_policy = ObjectList.rp_list.get(options.tcp_rp)()

I created a PR to fix this. In theory I have tested it, but I hadn´t worked in gem5 since gerrit, so if you could double check that I tested this appropriately, it wold be great: https://github.com/gem5/gem5/pull/1530

mattsinc: Thanks @odanrc !  I'm not sure which docker @yuxiaojia is using, but in theory you should be able to test yours using the same instructions here: https://github.com/gem5/gem5-resources/blob/develop/src/gpu/square/README.md

PR #1216, opened 2024-06-11
mem-ruby: Fix deadlock in GPU_VIPER when issuing atomic requests

v-ramadas:
When a compute unit issues several requests to the same line,
the requests wait in the L2 if it is a writeback cache. If the line is
invalid initially and the first request is atomic in nature, the L2
cache issues a request to main memory. On data return, the cache line
transitions to M but doesn't wake up the other requests, resulting in
a deadlock. This commit adds a wakeup call on data return for atomics
and fixes potential deadlocks.

### DIFF
--- a/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
@@ -606,6 +606,7 @@ machine(MachineType:TCC, "TCC Cache")
           out_msg.Destination.add(in_msg.Requestor);
           out_msg.Sender := machineID;
           out_msg.MessageSize := MessageSizeType:Response_Data;
+          out_msg.instSeqNum := in_msg.instSeqNum;
           out_msg.DataBlk := cache_entry.DataBlk;
           out_msg.isGLCSet := in_msg.isGLCSet;
           out_msg.isSLCSet := in_msg.isSLCSet;
--- a/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
@@ -1299,6 +1299,7 @@ machine(MachineType:TCC, "TCC Cache")
     wardb_writeAtomicResponseDirtyBytes;
     pa_performAtomic;
     baplr_sendBypassedAtomicPerformedLocallyResponse;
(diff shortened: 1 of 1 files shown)


mattsinc:
@v-ramadas : it looks like you pulled a lot of commits other than your own.  You'll need to fix that before this can go in.

mattsinc:
@v-ramadas : I think part of the problem is you tried to push to stable instead of develop

mattsinc:
Finally, it looks like your commit message is missing a ChangeID?

v-ramadas:
@mattsinc: Oops. I didn't realize that I was merging with stable. I changed it to develop now.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -606,6 +606,7 @@ machine(MachineType:TCC, "TCC Cache")
           out_msg.Destination.add(in_msg.Requestor);
           out_msg.Sender := machineID;
           out_msg.MessageSize := MessageSizeType:Response_Data;
+          out_msg.instSeqNum := in_msg.instSeqNum;
```
mattsinc, line 609: Is this needed for this commit?  It doesn't seem to affect the wada part.

v-ramadas: This helped me debug the issue and might be useful in fixing future issues. Without this, I was seeing a sequence number of 0 in all the messages exiting the TCC and it made it hard to track requests.

mattsinc: So ultimately the issue is: we want commits to be self-contained, and only contain what is needed to fix a specific bug or add a specific feature, etc.  In this case, it seems like you've combined two things: one to add more information and one to fix a bug.  I understand they are related, but if this commit were rolled back/reverted in the future, then both would be lost.  So I think the right solution would be to have two separate commits.

v-ramadas: I've split it across two commits and updated the pull request

mattsinc (approved):

mattsinc:
Thanks @v-ramadas !

abmerop (approved):

PR #1217, opened 2024-06-11
gpu-compute, util-m5: add GPU kernel exit events

yuxiaojia:
The GPUFS scripts include support for dumping and resetting
stats at kernel boundaries by identifying specific GPU kernel 
exit events. This commit extends that support to work with 
GPU SE-mode support.

Change-Id: I662233ae71e2987d90af1fd0100e29036b2ef1c6

### DIFF
--- a/configs/example/apu_se.py
+++ b/configs/example/apu_se.py
@@ -894,9 +894,9 @@ gpu_port_idx = gpu_port_idx - args.num_cp * 2
 token_port_idx = 0
 for i in range(len(system.ruby._cpu_ports)):
     if isinstance(system.ruby._cpu_ports[i], VIPERCoalescer):
-        system.cpu[shader_idx].CUs[token_port_idx].gmTokenPort = (
-            system.ruby._cpu_ports[i].gmTokenPort
-        )
+        system.cpu[shader_idx].CUs[
+            token_port_idx
+        ].gmTokenPort = system.ruby._cpu_ports[i].gmTokenPort
         token_port_idx += 1
 
 wavefront_size = args.wf_size
--- a/configs/example/apu_se.py
+++ b/configs/example/apu_se.py
@@ -894,9 +894,9 @@ gpu_port_idx = gpu_port_idx - args.num_cp * 2
 token_port_idx = 0
 for i in range(len(system.ruby._cpu_ports)):
     if isinstance(system.ruby._cpu_ports[i], VIPERCoalescer):
-        system.cpu[shader_idx].CUs[
-            token_port_idx
-        ].gmTokenPort = system.ruby._cpu_ports[i].gmTokenPort
+        system.cpu[shader_idx].CUs[token_port_idx].gmTokenPort = (
+            system.ruby._cpu_ports[i].gmTokenPort
+        )
         token_port_idx += 1
 
 wavefront_size = args.wf_size
@@ -1015,7 +1015,7 @@ while True:
         exit_event.getCause() == "m5_exit instruction encountered"
         or exit_event.getCause() == "user interrupt received"
         or exit_event.getCause() == "simulate() limit reached"
-        or  "exiting with last active thread context" in exit_event.getCause()
+        or "exiting with last active thread context" in exit_event.getCause()
     ):
         print(f"breaking loop due to: {exit_event.getCause()}.")
         break
@@ -1045,9 +1045,7 @@ while True:
         m5.stats.dump()
         m5.stats.reset()
     else:
-        print(
-            f"Unknown exit event: {exit_event.getCause()}. Continuing..."
-        )
+        print(f"Unknown exit event: {exit_event.getCause()}. Continuing...")
 
     exit_event = m5.simulate(maxtick - m5.curTick())
 
--- a/configs/example/apu_se.py
+++ b/configs/example/apu_se.py
@@ -1032,10 +1032,6 @@ while True:
         print("GPU Blit Kernel Completed dump and reset")
         m5.stats.dump()
         m5.stats.reset()
-    elif "Skipping GPU Kernel" in exit_event.getCause():
-        print("Skipping GPU Kernel dump and reset")
-        m5.stats.dump()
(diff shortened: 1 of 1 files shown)


mattsinc:
@yuxiaojia : this should go onto develop, not stable.  Because you tried to add to stable, it grabbed ~400 commits other than yours.

mattsinc:
Also I think we need to use "util-m5" in addition to "gpu-compute" in the commit message.

yuxiaojia:
> Also I think we need to use "util-m5" in addition to "gpu-compute" in the commit message.



> Also I think we need to use "util-m5" in addition to "gpu-compute" in the commit message.

Thanks for pointing that out, I will create a new pull request after my previous PR got merged

mattsinc:
@yuxiaojia : you don't need to wait for #1213 to be merged, you can just redo what you did here but with develop as the target instead

mattsinc:
@yuxiaojia : getting closer.  However it looks like this commit now includes ~8 commits, including other ones part of #1213.  I recommend putting these commits on a separate branch from #1213 to avoid this problem.

mattsinc:
@yuxiaojia : also please update the commit message to be:

gpu-compute: util-m5: add GPU kernel exit events

The GPUFS scripts include support for dump'ing and reset'ing stats at kernel 
boundaries by identifying specific GPU kernel exit events.  This commit
extends that support to work with GPU SE-mode support.

(you may need to update the above to fit in the character limit per line)

yuxiaojia:
> @yuxiaojia : getting closer. However it looks like this commit now includes ~8 commits, including other ones part of #1213. I recommend putting these commits on a separate branch from #1213 to avoid this problem.

I think I made a mistake that I did not create a separate branch for #1213, so I might need to wait for 1213 to get approved then rebase this commit to resolve this problem

mattsinc:
> > @yuxiaojia : getting closer. However it looks like this commit now includes ~8 commits, including other ones part of #1213. I recommend putting these commits on a separate branch from #1213 to avoid this problem.
> 
> I think I made a mistake that I did not create a separate branch for #1213, so I might need to wait for 1213 to get approved then rebase this commit to resolve this problem

I believe if you create a new branch in your repo, remove the commits from #1213, and then force push the branch it will create a new branch without those commits and update this gem5 pull request.  But if this is too complex I can help you with it tomorrow.

mattsinc (dismissed):
@abmerop please check too

configs/example/apu_se.py
```diff
@@ -1008,6 +1010,45 @@ def find_file(base_list, rel_path):
...
+        m5.stats.dump()
+        m5.stats.reset()
+    elif "GPU Blit Kernel Completed" in exit_event.getCause():
+        print("GPU Blit Kernel Completed dump and reset")
+        m5.stats.dump()
+        m5.stats.reset()
+    elif "Skipping GPU Kernel" in exit_event.getCause():
+        print("Skipping GPU Kernel dump and reset")
+        m5.stats.dump()
+        m5.stats.reset()
```
abmerop: You should never see this event since the kernel to skip to is not set in the GPUCommandProcessor above.  I would probably also not dump stats here and just continue as no kernel was simulated, however I don't know the exact use case you are going for.

abmerop (dismissed):

configs/example/apu_se.py
mattsinc: Yeah, @yuxiaojia this should probably be removed then -- @abmerop did the kernel skip only get added for GPUFS?

abmerop: Yes, it looks like it was only added for GPUFS

.github/workflows/daily-tests.yaml
```diff
@@ -125,7 +125,7 @@ jobs:
             # download all artifacts for each test. Thoguh this is inelegant,
             # it's simpler than figuring otu which long tests requires which
             # binary.
-            - uses: actions/download-artifact@v4
+            - uses: actions/download-artifact@v4.0.0
```
mattsinc: @yuxiaojia are these changes required as part of your update?

src/base/loader/symtab.test.cc
```diff
@@ -720,7 +720,7 @@ TEST(LoaderSymtabTest, FindNearestRoundWithNext)
     EXPECT_TRUE(symtab.insert(symbols[0]));
     EXPECT_TRUE(symtab.insert(symbols[1]));
 
-    Addr next_addr;
+    Addr next_addr = 0;
```
mattsinc: Same thing -- this doesn't seem part of your change

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -654,10 +655,12 @@ machine(MachineType:TCC, "TCC Cache")
 
   action(p_profileMiss, "pm", desc="Profile cache miss") {
       L2cache.profileDemandMiss();
+      DPRINTF(RubyHitMiss, "in TCC miss at %p \n", address);
```
mattsinc: This is part of a separate commit

mattsinc (changes requested):
It looks like you merged in other changes when updating -- please fix.

configs/example/apu_se.py
```diff
@@ -892,9 +894,9 @@ def find_file(base_list, rel_path):
 token_port_idx = 0
 for i in range(len(system.ruby._cpu_ports)):
     if isinstance(system.ruby._cpu_ports[i], VIPERCoalescer):
-        system.cpu[shader_idx].CUs[
-            token_port_idx
-        ].gmTokenPort = system.ruby._cpu_ports[i].gmTokenPort
+        system.cpu[shader_idx].CUs[token_port_idx].gmTokenPort = (
```
mattsinc: Looks like this still needs to be removed

mattsinc (changes requested):

mattsinc (approved):

abmerop (approved):

PR #1226, opened 2024-06-12
gpu-compute,mem-ruby: Add RubyHitMiss flag for TCP and TCC cache

yuxiaojia:
Add hit and miss print for TCP and TCC cache with RubyHitMiss debug flag

Change-Id: I4430532b901811e03d9b077b61e2eca4557b34e1


mattsinc (changes requested):
In terms of code changes everything looks good to me.  However there are at least two things that need to be changed elsewhere:

- your commit message needs to include the "mem-ruby" tag too -- as this is largely a mem-ruby change.
- It looks like you have ~6 commits for this change -- can you please merge those into a single commit?

yuxiaojia:
> In terms of code changes everything looks good to me. However there are at least two things that need to be changed elsewhere:
> 
> * your commit message needs to include the "mem-ruby" tag too -- as this is largely a mem-ruby change.
> * It looks like you have ~6 commits for this change -- can you please merge those into a single commit?

I tried to squash these 6 commits into one locally, but it seems like the remote one does not work

mattsinc:
> > In terms of code changes everything looks good to me. However there are at least two things that need to be changed elsewhere:
> > 
> > * your commit message needs to include the "mem-ruby" tag too -- as this is largely a mem-ruby change.
> > * It looks like you have ~6 commits for this change -- can you please merge those into a single commit?
> 
> I tried to squash these 6 commits into one locally, but it seems like the remote one does not work

What do you mean the remote one does not work?

yuxiaojia:
> > > In terms of code changes everything looks good to me. However there are at least two things that need to be changed elsewhere:
> > > 
> > > * your commit message needs to include the "mem-ruby" tag too -- as this is largely a mem-ruby change.
> > > * It looks like you have ~6 commits for this change -- can you please merge those into a single commit?
> > 
> > 
> > I tried to squash these 6 commits into one locally, but it seems like the remote one does not work
> 
> What do you mean the remote one does not work?

I can squash them locally, but when I push the change here, it seems like it will not modify here

mattsinc:
> > > > In terms of code changes everything looks good to me. However there are at least two things that need to be changed elsewhere:
> > > > 
> > > > * your commit message needs to include the "mem-ruby" tag too -- as this is largely a mem-ruby change.
> > > > * It looks like you have ~6 commits for this change -- can you please merge those into a single commit?
> > > 
> > > 
> > > I tried to squash these 6 commits into one locally, but it seems like the remote one does not work
> > 
> > 
> > What do you mean the remote one does not work?
> 
> I can squash them locally, but when I push the change here, it seems like it will not modify here

How are you trying to push it?

yuxiaojia:
> > > > > In terms of code changes everything looks good to me. However there are at least two things that need to be changed elsewhere:
> > > > > 
> > > > > * your commit message needs to include the "mem-ruby" tag too -- as this is largely a mem-ruby change.
> > > > > * It looks like you have ~6 commits for this change -- can you please merge those into a single commit?
> > > > 
> > > > 
> > > > I tried to squash these 6 commits into one locally, but it seems like the remote one does not work
> > > 
> > > 
> > > What do you mean the remote one does not work?
> > 
> > 
> > I can squash them locally, but when I push the change here, it seems like it will not modify here
> 
> How are you trying to push it?



> > > > > In terms of code changes everything looks good to me. However there are at least two things that need to be changed elsewhere:
> > > > > 
> > > > > * your commit message needs to include the "mem-ruby" tag too -- as this is largely a mem-ruby change.
> > > > > * It looks like you have ~6 commits for this change -- can you please merge those into a single commit?
> > > > 
> > > > 
> > > > I tried to squash these 6 commits into one locally, but it seems like the remote one does not work
> > > 
> > > 
> > > What do you mean the remote one does not work?
> > 
> > 
> > I can squash them locally, but when I push the change here, it seems like it will not modify here
> 
> How are you trying to push it?

Sorry for the late reply, just found a way to do it. Thanks!

configs/example/apu_se.py
```diff
@@ -892,9 +892,9 @@ def find_file(base_list, rel_path):
 token_port_idx = 0
 for i in range(len(system.ruby._cpu_ports)):
     if isinstance(system.ruby._cpu_ports[i], VIPERCoalescer):
-        system.cpu[shader_idx].CUs[token_port_idx].gmTokenPort = (
-            system.ruby._cpu_ports[i].gmTokenPort
-        )
+        system.cpu[shader_idx].CUs[
```
mattsinc: This will need to be removed

mattsinc (changes requested):

configs/example/apu_se.py
yuxiaojia: We do not want this change? I think the reason it warns me about this is its black format issue.

mattsinc: I don't think this change has anything to do with the rest of this commit, right?

mattsinc: Assuming that's the case, then if this change is needed, we should make a separate pull request for it

yuxiaojia: It is irrelevant to our change, and I can make a separate pull request for it

mattsinc: Yes, please do -- thanks!

yuxiaojia: I just found out that the latest develop already solved that issue, and I have already merged develop into this branch

mattsinc (dismissed):
Thanks!

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -655,10 +655,12 @@ machine(MachineType:TCC, "TCC Cache")
 
   action(p_profileMiss, "pm", desc="Profile cache miss") {
       L2cache.profileDemandMiss();
+      DPRINTF(RubyHitMiss, "in TCC miss at %p \n", address);
```
abmerop: A while (years) back we had some changes to make Ruby address prints be consistent.  I believe we decided on `%#lx`.  It makes it easier to search through ProtocolTrace output

src/mem/ruby/system/GPUCoalescer.hh
```diff
@@ -291,22 +291,25 @@ class GPUCoalescer : public RubyPort
 
     void readCallback(Addr address,
                       MachineType mach,
-                      DataBlock& data);
+                      DataBlock& data,
+                      bool externalHit);
```
abmerop: Have you tested other protocols?  Adding a new parameter without a default seems like it's going to break *any* other protocol which does not add the new parameter.

abmerop (changes requested):

src/mem/ruby/system/GPUCoalescer.hh
mattsinc: This is a good point ... how do we test this easily?  I had thought the daily/weekly tests would be enough, but maybe not?

abmerop: Oh, scratch that. I didn't notice with was only in GPUCoalescer. A default would still be nice to avoid breaking other GPU protocols.  I don't know what "externalHit" means so I don't know what that should be but either true/false.

mattsinc: I think defaulting to false makes sense, as it would at least not make people think something hit when it (maybe) didn't?

abmerop: Sure that makes sense

abmerop (approved):

mattsinc (approved):
Thanks @yuxiaojia !

BobbyRBruce:
@mattsinc @abmerop @yuxiaojia : VEGA_X86 fails to compile due to these changes: https://github.com/gem5/gem5/actions/runs/9553546579/job/26332597675

abmerop:
Build fails on develop (too slow, see above comment).

I know I pushed on the default parameter, but the easiest solution is to just remove it.  I was attempting to save myself some work later but this problem is bigger.

mattsinc:
Strange, I thought @yuxiaojia had tested this and I thought the above tests would compile VEGA -- hence I was comfortable merging it in.  Seems that is not the case?  Or maybe I'm missing the above comment by @abmerop ...

In any event @yuxiaojia is traveling today (I think), so we should revert until he can fix this if it is a legit issue.

PR #1230, opened 2024-06-12
configs: Add replacement policy options for GPUFS

abmerop:
GPU_VIPER.py was modified to use these options but they did not exist, breaking GPUFS.  This commit adds them to fix the issue.

Change-Id: I0095f400ea606c4e8d91a41870ef208465cef803

### DIFF
--- a/configs/example/gpufs/runfs.py
+++ b/configs/example/gpufs/runfs.py
@@ -195,6 +195,28 @@ def addRunFSOptions(parser):
         help="Disable KVM perf counters (use this with LSF / ETX)",
     )
 
+    parser.add_argument(
+        "--tcp-rp",
+        type=str,
+        default="TreePLRURP",
+        help="cache replacement policy" "policy for tcp",
+    )
+
+    parser.add_argument(
+        "--tcc-rp",
+        type=str,
+        default="TreePLRURP",
+        help="cache replacement policy" "policy for tcc",
+    )
+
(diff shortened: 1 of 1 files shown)


mattsinc (approved):
@yuxiaojia : FYI

mattsinc (commented):
@abmerop : while we're at it, do we want to update the way that GPUFS picks the replacement policy?  Or I guess @yuxiaojia could make a separate commit for that?

abmerop:
> @abmerop : while we're at it, do we want to update the way that GPUFS picks the replacement policy? Or I guess @yuxiaojia could make a separate commit for that?

Currently both GPUFS and GPU-SE are using the common ruby config to set the replacement policy, is modifying that what you were thinking?

mattsinc:
> > @abmerop : while we're at it, do we want to update the way that GPUFS picks the replacement policy? Or I guess @yuxiaojia could make a separate commit for that?
> 
> Currently both GPUFS and GPU-SE are using the common ruby config to set the replacement policy, is modifying that what you were thinking?

I think?  I think you're saying by adding these flags here, a user can pick other replacement policies now in both GPUFS and GPUSE?

abmerop:
> I think? I think you're saying by adding these flags here, a user can pick other replacement policies now in both GPUFS and GPUSE?

Yes, they can do that now, and the command line parameter will be the same.

PR #1234, opened 2024-06-13
arch,cpu,sim: Add mechanism to partially print vector regs

hnpl:
Currently, gem5's inst tracer prints the whole vector register container by default. The size of vector register containers in gem5 is the maximum size allowed by the ISA. For vector-length agnostic (VLA) vector registers, this means ARM SVE vector container is 2048 bits long, and RISC-V vector container is 65535 bits long. Note that VLA implementation in gem5 allows the vector length to be varied within the limit specified by the ISAs.

However, in most use cases of gem5, the vector length is much less than 65535 bits. This causes two issues: (1) the vector container requires allocating and moving around a large amount of unused data while only a fraction of it is used, and (2) printing the execution trace of a vector register results in a wall of text with a small amount of useful data.

This change addresses the problem (2) by providing a mechanism to limit the amount data printed by the instruction tracer. This is done by adding a function printing the first X bits of a vector register container, where X is the vector length determined at runtime, as opposed to the vector container size, which is determined at compilation time.

Change-Id: I815fa5aa738373510afcfb0d544a5b19c40dc0c7

### DIFF
--- a/src/cpu/reg_class.hh
+++ b/src/cpu/reg_class.hh
@@ -45,6 +45,7 @@
 #include <iterator>
 #include <string>
 
+#include "arch/generic/vec_reg.hh"
 #include "base/cprintf.hh"
 #include "base/debug.hh"
 #include "base/intmath.hh"
@@ -170,7 +171,7 @@ class RegClassOps
     /** Print the name of the register specified in id. */
     virtual std::string regName(const RegId &id) const;
     /** Print the value of a register pointed to by val of size size. */
-    virtual std::string valString(const void *val, size_t size) const;
+    virtual std::string valString(const void *val, const size_t& size) const;
     /** Flatten register id id using information in the ISA object isa. */
     virtual RegId
     flatten(const BaseISA &isa, const RegId &id) const
@@ -245,6 +246,11 @@ class RegClass
     valString(const void *val) const
     {
         return _ops->valString(val, regBytes());
+    }
+    std::string
+    valString(const void *val, const uint64_t& num_bytes) const
+    {
+        return _ops->valString(val, std::min(regBytes(), num_bytes));
     }
     RegId
     flatten(const BaseISA &isa, const RegId &id) const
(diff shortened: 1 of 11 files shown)


hnpl:
I tested this change set by producing the traces from a simple RVV binary with and without this change set.

The desired outcome is to only have differences in the amount of vector register content printed by the tracer. It looks like the output matches the expected outcome.

image


I'll test this change with SVE binaries and probably some x86 binaries with SSE* instructions.

I also tested this change set by compiling `build/ALL/gem5.opt` and it worked.

src/cpu/reg_class.hh
```diff
@@ -354,15 +360,30 @@ RegClass::operator[](RegIndex idx) const
...
-        assert(size == sizeof(ValueType));
-        return csprintf("%s", *(const ValueType *)val);
+        if constexpr (is_vec_reg_container<ValueType>::value) {
+            if (size == sizeof(ValueType)) {
+                return csprintf("%s", *(const ValueType *)val);
+            } else {
+                return ((const ValueType *)val)->getString(size);
+            }
+        } else {
+            assert(size == sizeof(ValueType));
```
powerjg, line 384: Should this be a static assert?

powerjg (commented):
Generally, it looks good to me. One small thing, but if you disagree, then it's good to go from my point of view.

src/cpu/reg_class.hh
hnpl: Yeah, agreed. The value passed to size is regBytes(), which is a constexpr.

hnpl: I just figure out that making this assert being a static_assert does not work because of this change.

Previously, there is only one place calling this function, and the value passed to `size` is always a constexpr, so the static_assert would make sense.

Now that we call this function is called at two different places, thus `size` is no longer guaranteed to be determined at compilation time.

hnpl:
It seems to work properly with SVE instructions.

image

src/sim/insttracer.hh
```diff
@@ -232,14 +232,38 @@ class InstRecord
     setData(const RegClass &reg_class, RegVal val)
     {
         new(&data.asReg) InstResult(reg_class, val);
-        dataStatus = DataReg;
+        switch (reg_class.type()) {
+            case IntRegClass:
+            case MiscRegClass:
```
hnpl, line 237: I'm not really confident about this part. I'm not entirely sure that MiscRegs and CCRegs are always not larger than 64 bits.

hnpl:
I just realized that the x86 instructions modifying xmm registers are broken down into micro-ops and do not need the vector container printing method. So the SSE* instructions will have the same trace as before.

BobbyRBruce (approved):

PR #1251, opened 2024-06-17
Revert "cpu-o3: Do not set Executed on load instruction to be replayed"

giactra:
Reverts gem5/gem5#1182

This is breaking O3 execution. Investigating the matter

### DIFF
--- a/src/cpu/o3/lsq_unit.cc
+++ b/src/cpu/o3/lsq_unit.cc
@@ -611,12 +611,6 @@ LSQUnit::executeLoad(const DynInstPtr &inst)
     if (inst->isTranslationDelayed() && load_fault == NoFault)
         return load_fault;
 
-    // Partial Store-to-Load Forwarding condition marks the load to be
-    // reissued during LSQUnit::read(). In this case we shouldn't notify
-    // iewStage that the instruction is ready for commit.
-    if (!inst->isIssued() && !inst->effAddrValid())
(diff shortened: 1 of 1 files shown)


pranith:
Can you share an example program that is failing @giactra?

giactra:
> example

I will try to come up with a minimal example, but FYI I am quite sure long regressions are broken anyway (o3 linux boot is probably stuck as it has been running for quite some time on my machine with no output being generated)

powerjg:
Thanks for pushing this, Giacomo!

On a side note, do you think it would be reasonable to (1) drop the requirement for change-ids on "revert" commits or (2) drop the requirement from all changes?

ivanaamit:
Hi @giactra, thanks for pushing this.

However, the revert button on GitHub doesn't create a proper commit message and doesn't include the change ID. Could you please update the PR with those changes so it is ready to be merged? Thanks.

giactra:
> Hi @giactra, thanks for pushing this.
> 
> However, the revert button on GitHub doesn't create a proper commit message and doesn't include the change ID. Could you please update the PR with those changes so it is ready to be merged? Thanks.

Is it ok now?

ivanaamit:
@giactra, the commit message should start with a tag: 

cpu-o3: Revert "Do not set Executed on load instruction to..."

BobbyRBruce (approved):

PR #1254, opened 2024-06-17
gpu-compute,mem-ruby: Revert "Add RubyHitMiss flag for TCP and TCC cache"

BobbyRBruce:
Reverts gem5/gem5#1226

### DIFF
--- a/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
@@ -655,12 +655,10 @@ machine(MachineType:TCC, "TCC Cache")
 
   action(p_profileMiss, "pm", desc="Profile cache miss") {
       L2cache.profileDemandMiss();
-      DPRINTF(RubyHitMiss, "in TCC miss at %#lx\n", address);
   }
 
   action(p_profileHit, "ph", desc="Profile cache hit") {
       L2cache.profileDemandHit();
-      DPRINTF(RubyHitMiss, "in TCC hit at %#lx\n", address);
   }
 
   action(t_allocateTBE, "t", desc="allocate TBE Entry") {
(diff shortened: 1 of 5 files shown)


mattsinc (approved):
I am a little confused why the compilation that happens with each commit didn't catch this, but for now approving until @yuxiaojia can fix locally.

abmerop:
> I am a little confused why the compilation that happens with each commit didn't catch this, but for now approving until @yuxiaojia can fix locally.

The short answer is build/ALL/gem5.opt built by the CI test doesn't build the GPU files (BUILD_GPU=False).

abmerop (approved):

mattsinc:
> > I am a little confused why the compilation that happens with each commit didn't catch this, but for now approving until @yuxiaojia can fix locally.
> 
> The short answer is build/ALL/gem5.opt built by the CI test doesn't build the GPU files (BUILD_GPU=False).

That's probably not good if GPU builds are not being included in CI per-checkin tests?  I'm sure there is a good reason I'm just forgetting though.

BobbyRBruce:
> > > I am a little confused why the compilation that happens with each commit didn't catch this, but for now approving until @yuxiaojia can fix locally.
> > 
> > 
> > The short answer is build/ALL/gem5.opt built by the CI test doesn't build the GPU files (BUILD_GPU=False).
> 
> That's probably not good if GPU builds are not being included in CI per-checkin tests? I'm sure there is a good reason I'm just forgetting though.

We just can't afford to do a lot of compilations for the CI tests. We compile ALL and run tests which use ALL and save varying different compilations for Daily or Weekly tests.

abmerop:
> > > > I am a little confused why the compilation that happens with each commit didn't catch this, but for now approving until @yuxiaojia can fix locally.
> > > 
> > > 
> > > The short answer is build/ALL/gem5.opt built by the CI test doesn't build the GPU files (BUILD_GPU=False).
> > 
> > 
> > That's probably not good if GPU builds are not being included in CI per-checkin tests? I'm sure there is a good reason I'm just forgetting though.
> 
> We just can't afford to do a lot of compilations for the CI tests. We compile ALL and run tests which use ALL and save varying different compilations for Daily or Weekly tests.

I did a quick test adding BUILD_GPU=True to build_opts/ALL and it *does* compile.  I don't know how much more time that adds though.  Plus you cannot run anything with it anyway.  That would require building all ruby protocols too.

PR #1261, opened 2024-06-19
configs: Check before use replacement policy options

abmerop:
Rather than adding the options to *every* config that might be using GPU_VIPER.py, just change the Ruby config to check if the option is available before trying to use it. Otherwise, reverts to what was the default on stable.

Change-Id: Ia6f1d0827d489ee2a35c598b644461cbff59e247

### DIFF
--- a/configs/ruby/GPU_VIPER.py
+++ b/configs/ruby/GPU_VIPER.py
@@ -149,7 +149,8 @@ class TCPCache(RubyCache):
         self.size = MemorySize(options.tcp_size)
         self.assoc = options.tcp_assoc
         self.resourceStalls = options.no_tcc_resource_stalls
-        self.replacement_policy = RP_choose(options.tcp_rp)
+        if hasattr(options, "tcp_rp"):
+            self.replacement_policy = RP_choose(options.tcp_rp)
 
 
 class TCPCntrl(TCP_Controller, CntrlBase):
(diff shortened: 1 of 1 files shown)


abmerop:
Should fix daily test failure.  https://github.com/gem5/gem5/actions/runs/9568053451/job/26400126071

mattsinc:
@abmerop I assume we'd need the same for the other replacement policy options like tcc-rp too?

@yuxiaojia : FYI

powerjg (dismissed):

configs/ruby/GPU_VIPER.py
```diff
@@ -303,7 +305,8 @@ def create(self, options):
         self.start_index_bit = math.log(options.cacheline_size, 2) + math.log(
             options.num_tccs, 2
         )
-        self.replacement_policy = RP_choose(options.tcp_rp)
+        if hasattr(options, "tcp_rp"):
```
mattsinc: I think this needs to be tcc_rp?

configs/ruby/GPU_VIPER.py
```diff
@@ -241,7 +242,8 @@ class SQCCache(RubyCache):
     def create(self, options):
         self.size = MemorySize(options.sqc_size)
         self.assoc = options.sqc_assoc
-        self.replacement_policy = RP_choose(options.tcp_rp)
+        if hasattr(options, "tcp_rp"):
```
mattsinc: I think this needs to be sqc_rp?

mattsinc (changes requested):

configs/ruby/GPU_VIPER.py
abmerop: Yea, I think this was incorrect before

mattsinc: Just to make sure I'm not going crazy: did you push a change to update this?  Or just add a comment?  If the former, I do not see it ...

abmerop: Still working on it, should be up now

mattsinc (approved):
Thanks, @abmerop !  We'll have to be more careful about compiling many different things locally moving forward since I mistakenly though CI was testing GPU builds.

PR #1264, opened 2024-06-20
arch-riscv: Fix TLB lookup with vaddrs

### LINKED ISSUE #1235: The riscv virtual address translation process may not work under non 4KiB page sizes
**Describe the bug**
The process of converting virtual addresses into physical addresses under the riscv architecture is as follows:
```
1. Let a be satp:ppn × PAGESIZE, and let i = LEVELS − 1. (For Sv32, PAGESIZE=2^12 and
LEVELS=2.)
2. Let pte be the value of the PTE at address a+va:vpn[i]×PTESIZE. (For Sv32, PTESIZE=4.)
If accessing pte violates a PMA or PMP check, raise an access exception.
3. If pte:v = 0, or if pte:r = 0 and pte:w = 1, stop and raise a page-fault exception.
4. Otherwise, the PTE is valid. If pte:r = 1 or pte:x = 1, go to step 5. Otherwise, this PTE is a
pointer to the next level of the page table. Let i = i − 1. If i < 0, stop and raise a page-fault
exception. Otherwise, let a = pte:ppn × PAGESIZE and go to step 2.
5. A leaf PTE has been found. Determine if the requested memory access is allowed by the
pte:r, pte:w, pte:x, and pte:u bits, given the current privilege mode and the value of the SUM
and MXR fields of the mstatus register. If not, stop and raise a page-fault exception.
6. If i > 0 and pa:ppn[i − 1 : 0] 6= 0, this is a misaligned superpage; stop and raise a page-fault
exception.
7. If pte:a = 0, or if the memory access is a store and pte:d = 0, either raise a page-fault
exception or:
• Set pte:a to 1 and, if the memory access is a store, also set pte:d to 1.
• If this access violates a PMA or PMP check, raise an access exception.
• This update and the loading of pte in step 2 must be atomic; in particular, no intervening
store to the PTE may be perceived to have occurred in-between.
8. The translation is successful. The translated physical address is given as follows:
• pa.pgoff = va.pgoff.
• If i > 0, then this is a superpage translation and pa:ppn[i − 1 : 0] = va:vpn[i − 1 : 0].
• pa:ppn[LEVELS − 1 : i] = pte:ppn[LEVELS − 1 : i]
```
In step 8, if i > 0, then：
```
 If i > 0, then this is a superpage translation and pa:ppn[i − 1 : 0] = va:vpn[i − 1 : 0].
```
But the implementation in the doTranslate function in src/arch/riscv/tlb.cc of gem5 is like this:
```
Addr paddr = e->paddr << PageShift | (vaddr & mask(e->logBytes));
```
There is no problem when the page size is 4KiB, but if the page size is not 4KiB, this implementation will have problems.

**Affects version**
gem5 version 23.1.0.0

**gem5 Modifications**
There are two places in gem5 that need to be modified.One is 
```
Addr paddr = e->paddr << PageShift | (vaddr & mask(e->logBytes));
```
in the doTranslate() function in the src/arch/riscv/tlb.cc file, which should be modified as follows:
```
Addr paddr = ((e->paddr >> (e->logBytes - PageShift)) << e->logBytes) | (vaddr & mask(e->logBytes));
```
Another place is that the translateWithTLB function in the file needs to be modified as follows:
```
Addr
TLB::translateWithTLB(Addr vaddr, uint16_t asid, BaseMMU::Mode mode)
{
    TlbEntry *e = lookup(vaddr, asid, mode, false);
    assert(e != nullptr);
    return ((e->paddr >> (e->logBytes - PageShift)) << e->logBytes) | (vaddr & mask(e->logBytes));
}
```

**Host Operating System**
Ubuntu 22.04,

**Host ISA**
X86,

**Compiler used**
GCC 11.4.0

**Additional information**

powerjg:
Previously, all of the TLB lookup/insert functions were using the full virtual addresses even though the variables in the functions said "vpn." This change explicitly converts the virtual address to the VPN without any least significant zeros for the offset. I.e., vpn >> page_size.

The main bug solved in this changeset is the asid was |'d with the upper bits of the virtual address, but sometimes there were all 1's. Therefore, you could get a TLB hit even if the ASID was different. Interestingly, the page that seemed to cause these issues was a 1 GiB page.

This change also starts refactoring some of the page table details to support sv46 and sv57 page table formats.

In my testing, the Linux kernel boot uses large pages (even OpenSBI uses large pages), so it seems that large pages also work. However, this seems like magic to me, so I'm not sure if it's correct.

This change also updates some asserts, and debug statements with more useful debugging information.

Partially fixes #1235. More testing needs to be done to be confident.

Change-Id: If832fe8ae89874cf48d7367f5694c5adba2bd6d7

### DIFF
--- a/src/arch/riscv/tlb.cc
+++ b/src/arch/riscv/tlb.cc
@@ -67,6 +67,12 @@ using namespace RiscvISA;
 static Addr
 buildKey(Addr vpn, uint16_t asid)
 {
+    // Note ASID is 16 bits
+    // The VPN in sv39 is up to 39-12=27 bits
+    // The VPN in sv48 is up to 48-12=36 bits
+    // The VPN in sv57 is up to 57-12=45 bits
+    // So, shifting the ASID into the top 16 bits is safe.
+    assert(bits(vpn, 63, 48) == 0);
     return (static_cast<Addr>(asid) << 48) | vpn;
 }
 
@@ -110,6 +116,12 @@ TLB::lookup(Addr vpn, uint16_t asid, BaseMMU::Mode mode, bool hidden)
 {
     TlbEntry *entry = trie.lookup(buildKey(vpn, asid));
 
+    DPRINTF(TLBVerbose, "lookup(vpn=%#x, asid=%#x, key=%#x): "
+                        "%s ppn=%#x (%#x) %s\n",
+            vpn, asid, buildKey(vpn, asid), entry ? "hit" : "miss",
+            entry ? entry->paddr : 0, entry ? entry->size() : 0,
+            hidden ? "hidden" : "");
+
     if (!hidden) {
         if (entry)
             entry->lruSeq = nextSeq();
@@ -131,9 +143,6 @@ TLB::lookup(Addr vpn, uint16_t asid, BaseMMU::Mode mode, bool hidden)
             else
                 stats.readHits++;
         }
-
-        DPRINTF(TLBVerbose, "lookup(vpn=%#x, asid=%#x): %s ppn %#x\n",
-                vpn, asid, entry ? "hit" : "miss", entry ? entry->paddr : 0);
     }
 
     return entry;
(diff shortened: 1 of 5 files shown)


powerjg:
Tests are still running, but I'm hopeful. I'm setting this as ready for review so the tests can start running

src/arch/riscv/tlb.cc
```diff
@@ -267,9 +295,10 @@ TLB::createPagefault(Addr vaddr, BaseMMU::Mode mode)
...
 
 Addr
-TLB::translateWithTLB(Addr vaddr, uint16_t asid, BaseMMU::Mode mode)
+TLB::translateWithTLB(Addr vaddr, uint16_t asid, Addr xmode,
+                      BaseMMU::Mode mode)
 {
-    TlbEntry *e = lookup(vaddr, asid, mode, false);
+    TlbEntry *e = lookup(getVPNFromVAddr(vaddr, xmode), asid, mode, false);
     assert(e != nullptr);
     return e->paddr << PageShift | (vaddr & mask(e->logBytes));
```
hnpl, line 303: This still is very odd to me. It looks like we only shift the pfn by 12 bits, while the number of bits we keep from vaddr is >= 12 bits.

Note that, in step 8, what we did was,

```C++
entry.paddr = pte.ppn;
```

Perhaps we should have a simple setup to do unit testing on these functions.

powerjg: We *should* test them. I can go update the documentation, but from all of my tests `paddr` is the PPN. For huge pages, the PPN will be the aligned 4KiB page number, not the "super page" number.

So, we need to get the offset from the `vaddr` (based on the size of the page), and then *or* that with the upper bits of the address from the PPN. 

Does this make sense, or do you think there's still something wrong.

Overall, this code is *incredibly* confusing with many different (and contradictory) names for the same thing.

hnpl: Step 6 of the the translation in the spec says that if a page is a leaf page then the unused ppn parts must be 0. So I think you're correct.

hnpl (approved):
lgtm

PR #1266, opened 2024-06-20
cpu: Fix `std::min` type mismatch in reg_class.hh

BobbyRBruce:
Introduced in #1234, this caused compilation to faill in Apple Silicon systems. This bug is the same as #582 where a more detailed explanation is provided

### DIFF
--- a/src/cpu/reg_class.hh
+++ b/src/cpu/reg_class.hh
@@ -248,7 +248,7 @@ class RegClass
         return _ops->valString(val, regBytes());
     }
     std::string
-    valString(const void *val, const uint64_t& num_bytes) const
+    valString(const void *val, const size_t& num_bytes) const
     {
         return _ops->valString(val, std::min(regBytes(), num_bytes));
     }


BobbyRBruce:
@hnpl : Not sure if this is the best solution but it is _a_ solution. I don't know this code well and what the best types are for these variables (or if it really matters at all). Let me know what you think of this chance. I'm happy to accept an alternative fix.

hnpl:
@BobbyRBruce: the code lgtm. Thanks for fixing this! It seems that `size_t` is problematic when porting to Mac.

`regBytes()` and `num_bytes` both indicate a register size in bytes. I think using either of 32-bit integer or 64-bit integer type is okay.

giactra:
Hi Bobby, can't we just change the signature of the method to accept size_t instead of uint64_t? In this way we won't have to cast (but I understand we might be pushing the issue somewhere else)

BobbyRBruce:
> Hi Bobby, can't we just change the signature of the method to accept size_t instead of uint64_t? In this way we won't have to cast (but I understand we might be pushing the issue somewhere else)

Turns out that does work. I was mistaken in thinking there was a function already with that signature, but that's in `RegClassOpts`.

I just forced push this suggestion,

giactra (approved):
Thanks for this fix!

hnpl:
I'm not against this as using 32-bit or 64-bit integer types wouldn't matter much in this case. However, for other cases, would it cause problems in future with this implicit downward type casting?

powerjg:
IMO, this is OK for now. However, we should probably go through the code carefully and look for these system-specific integer types (size_t, long, etc.) and try to remove them where appropriate (I just found another this morning when working on a Windows build...)

PR #1270, opened 2024-06-20
gpu-compute,tests: Move GPU tests to testlib

BobbyRBruce:
A new host tag `gcn_gpu` has been added. This allows for selection of those GPU tests which depend upon the gcn-gpu docker image to run.

In addition to this, the square GPU tests has been moved to the CI tests. This ensures some GPU code is compiled and run on every PR.

Change-Id: Ic1a3e3a5c14566c5a53aeb74ab6d0725a1310fff

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -221,6 +221,32 @@ jobs:
                   path: tests/testing-results
                   retention-days: 30
 
+    gpu-tests:
+        runs-on: [self-hosted, linux, x64]
+        container: ghcr.io/gem5/gcn-gpu:latest
+        timeout-minutes: 180
+        needs: [pre-commit, check-for-change-id]
+        steps:
+            - uses: actions/checkout@v4
+
+            # Build the VEGA_X86/gem5.opt binary.
+            - name: Build VEGA_X86/gem5.opt
+              run: scons build/VEGA_X86/gem5.opt -j`nproc`
+
+            # Run the GPU tests.
+            - name: Run Testlib GPU Tests
+              working-directory: ${{ github.workspace }}/tests
+              run: ./main.py run  --skip-build -vvv -t $(nproc) --host gcn_gpu gem5/gpu
+
+            # Upload the tests/testing-results directory as an artifact.
+            - name: Upload results
+              if: success() || failure()
+              uses: actions/upload-artifact@v4
+              with:
+                  name: ci-tests-run-${{ github.run_number }}-attempt-${{ github.run_attempt }}-gpu-status-${{ steps.run-tests.outcome }}-output
+                  path: tests/testing-results
+                  retention-days: 30
+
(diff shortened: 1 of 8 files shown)


BobbyRBruce:
I'm not sure if this is really any better, but just what I came up with trying to get the GPU tests into test lib. Due to the docker container requirement they are still a special case.

.github/workflows/ci-tests.yaml
```diff
@@ -221,6 +221,32 @@ jobs:
...
+        timeout-minutes: 180
+        needs: [pre-commit, check-for-change-id]
+        steps:
+            - uses: actions/checkout@v4
+
+            # Build the VEGA_X86/gem5.opt binary.
+            - name: Build VEGA_X86/gem5.opt
+              run: scons build/VEGA_X86/gem5.opt -j`nproc`
+
+            # Run the GPU tests.
```
mattsinc, line 236: How do I understand what tests this runs?  It does not seem to correspond to any of the other file names in this commit ...

.github/workflows/daily-tests.yaml
```diff
@@ -80,6 +80,30 @@ jobs:
...
+        timeout-minutes: 300
+
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  ref: develop
+            - name: Build VEGA_X86/gem5.opt
+              run: scons build/VEGA_X86/gem5.opt -j`nproc`
+
+            - name: Run Testlib GPU Tests
```
mattsinc, line 95: Same question here

.github/workflows/ci-tests.yaml
BobbyRBruce: The code that follows comment this calls our TestLib framework:

```shell
./main.py run  --skip-build -vvv -t $(nproc) --host gcn_gpu gem5/gpu
```

The important part here is the "gem5/gpu" at the end. That means it's running the tests defined in the "tests/gem5/gpu" directory. For example: https://github.com/gem5/gem5/blob/fe8b2b24eb9b08c07badd6f08fdeedc32db5a5c1/tests/gem5/gpu/test_gpu_apu_se.py. This script show what script gem5 is running and with what parameters using the Vega binary we built in the previous step.

The only snag with this implementation is needing to pass `--host gcn_gpu` to testlib to run the tests. This is because we need to compile VEGA and run the tests in the special docker container and I use this flag to ensure this happens.

.github/workflows/daily-tests.yaml
BobbyRBruce: My answer here is similar to the above but the only difference here is passing `--length=long` to state we want to run the Daily/Long tests found in the "tests/gem5/gpu" directory.

mattsinc (approved):
Thanks @BobbyRBruce , I don't feel like I have a strong grasp on this code, but based on your reply seems ok ..

PR #1275, opened 2024-06-21
mem: Change long in src/mem/physical.cc to int64_t

erin-le:
This changes `long`s in src/mem/physical.cc, which are 32 bits or more, to `uint64_t`s, which are exactly 64 bits.

Change-Id: I64e089a2ac087bcf58b9c3c918c59dc5ff75d010

### DIFF
--- a/src/mem/physical.cc
+++ b/src/mem/physical.cc
@@ -367,7 +367,7 @@ PhysicalMemory::serializeStore(CheckpointOut &cp, unsigned int store_id,
     // memories that are not part of the address map can overlap
     std::string filename =
         name() + ".store" + std::to_string(store_id) + ".pmem";
-    int64_t range_size = range.size();
+    Addr range_size = range.size();
 
     DPRINTF(Checkpoint, "Serializing physical memory %s with size %d\n",
             filename, range_size);
@@ -452,7 +452,7 @@ PhysicalMemory::unserializeStore(CheckpointIn &cp)
     uint8_t* pmem = backingStore[store_id].pmem;
     AddrRange range = backingStore[store_id].range;
 
-    int64_t range_size;
+    Addr range_size;
     UNSERIALIZE_SCALAR(range_size);
 
     DPRINTF(Checkpoint, "Unserializing physical memory %s with size %d\n",
@@ -463,8 +463,8 @@ PhysicalMemory::unserializeStore(CheckpointIn &cp)
               range_size, range.size());
 
     uint64_t curr_size = 0;
-    int64_t* temp_page = new int64_t[chunk_size];
-    int64_t* pmem_current;
+    uint8_t* temp_page = new uint8_t[chunk_size];
+    uint8_t* pmem_current;
     uint32_t bytes_read;
     while (curr_size < range.size()) {
         bytes_read = gzread(compressed_mem, temp_page, chunk_size);
@@ -473,11 +473,12 @@ PhysicalMemory::unserializeStore(CheckpointIn &cp)
 
         assert(bytes_read % sizeof(int64_t) == 0);
 
-        for (uint32_t x = 0; x < bytes_read / sizeof(int64_t); x++) {
+        for (uint32_t x = 0; x < bytes_read / sizeof(uint8_t); x++) {
             // Only copy bytes that are non-zero, so we don't give
             // the VM system hell
             if (*(temp_page + x) != 0) {
-                pmem_current = (int64_t*)(pmem + curr_size + x * sizeof(int64_t));
+                pmem_current = (uint8_t*)(pmem + curr_size + x *
+                sizeof(uint8_t));
                 *pmem_current = *(temp_page + x);
             }
         }
--- a/src/mem/physical.cc
+++ b/src/mem/physical.cc
@@ -367,7 +367,7 @@ PhysicalMemory::serializeStore(CheckpointOut &cp, unsigned int store_id,
     // memories that are not part of the address map can overlap
     std::string filename =
         name() + ".store" + std::to_string(store_id) + ".pmem";
-    long range_size = range.size();
+    int64_t range_size = range.size();
 
     DPRINTF(Checkpoint, "Serializing physical memory %s with size %d\n",
             filename, range_size);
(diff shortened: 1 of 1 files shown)


giactra:
I would use Addr instead, which is still 64 bit. This is what we have been commonly using to describe memory (addresses)

src/mem/physical.cc
```diff
@@ -463,21 +463,21 @@ PhysicalMemory::unserializeStore(CheckpointIn &cp)
               range_size, range.size());
 
     uint64_t curr_size = 0;
-    long* temp_page = new long[chunk_size];
-    long* pmem_current;
+    int64_t* temp_page = new int64_t[chunk_size];
```
powerjg: @giactra do you think it makes sense to use Addr here?

src/mem/physical.cc
```diff
@@ -463,21 +463,21 @@ PhysicalMemory::unserializeStore(CheckpointIn &cp)
...
     while (curr_size < range.size()) {
         bytes_read = gzread(compressed_mem, temp_page, chunk_size);
         if (bytes_read == 0)
             break;
 
-        assert(bytes_read % sizeof(long) == 0);
+        assert(bytes_read % sizeof(int64_t) == 0);
 
-        for (uint32_t x = 0; x < bytes_read / sizeof(long); x++) {
+        for (uint32_t x = 0; x < bytes_read / sizeof(int64_t); x++) {
```
powerjg: Same here.

I would say uint64_t makes the list sense, but I don't think it matters.

src/mem/physical.cc
```diff
@@ -367,7 +367,7 @@ PhysicalMemory::serializeStore(CheckpointOut &cp, unsigned int store_id,
     // memories that are not part of the address map can overlap
     std::string filename =
         name() + ".store" + std::to_string(store_id) + ".pmem";
-    long range_size = range.size();
+    int64_t range_size = range.size();
```
powerjg: Addr makes sense for range size

powerjg (commented):
The best option is a mix of Addr and uint64_t IMO. However, I worry slightly that the previous implementation was signed

src/mem/physical.cc
giactra: Honestly, I don't know why we are not allocating with bytes here (using uint8_t)

powerjg: Yes! That makes more sense. To answer your question, it's because this code was committed in 2006, lol

src/mem/physical.cc
erin-le: What do you mean by "list sense"?

powerjg: See Giacomo's comment above. It was a typo since I was on my phone.

giactra (changes requested):
Hi all, I gave a second look at the code, and I think I figured out the motivation behind using long here when allocating an intermediate page/buffer. The reason is *in theory* efficiency; the code down below implements a custom "optimized" memcopy. 

        for (uint32_t x = 0; x < bytes_read / sizeof(long); x++) {
            // Only copy bytes that are non-zero, so we don't give
            // the VM system hell
            if (*(temp_page + x) != 0) {
                pmem_current = (long*)(pmem + curr_size + x * sizeof(long));
                *pmem_current = *(temp_page + x);
            }

The optimization comes from the elimination of zero copies (see comment) and from the use of word sized stores to copy data (wider support than bytes sized stores) : data is copied word by word and a word is skipped if it is zero.

It is my personal opinion we should amend this code to make it more maintainable. 
We might think about replacing the for loop with a memcpy (these routines are anyway heavily optimized according to the running host), but I believe we could get rid of the intermediate copy straight away and gzread directly on pmem, updating the pointer according to the bytes_read. Please let me know your thoughts

powerjg:
That's great with me! We should double check the performance of restoring a checkpoint isn't significantly different though.

src/mem/physical.cc
```diff
@@ -463,29 +463,17 @@ PhysicalMemory::unserializeStore(CheckpointIn &cp)
...
-
-        for (uint32_t x = 0; x < bytes_read / sizeof(long); x++) {
-            // Only copy bytes that are non-zero, so we don't give
-            // the VM system hell
-            if (*(temp_page + x) != 0) {
-                pmem_current = (long*)(pmem + curr_size + x * sizeof(long));
-                *pmem_current = *(temp_page + x);
-            }
-        }
+        assert(bytes_read % sizeof(int64_t) == 0);
```
giactra: I don't think this is needed anymore

giactra (changes requested):
Thanks for your changes, this looks much cleaner! I only have 2 minor comments. One is below; the other one is that we can remove chunk_size as it is not needed anymore.

As a final request, I just wanted to make sure you ran some checkpoint tests to verify cp restore still works as expected (I don't see an issue in the code per se but it's better to test this before it gets merged).

Many thanks!

src/mem/physical.cc
```diff
@@ -463,29 +463,17 @@ PhysicalMemory::unserializeStore(CheckpointIn &cp)
               range_size, range.size());
 
     uint64_t curr_size = 0;
-    long* temp_page = new long[chunk_size];
-    long* pmem_current;
     uint32_t bytes_read;
     while (curr_size < range.size()) {
-        bytes_read = gzread(compressed_mem, temp_page, chunk_size);
+        bytes_read = gzread(compressed_mem, pmem, chunk_size);
```
erin-le, line 468: It looks like chunk_size is still used here. Since this is the only place it's used, should I put the value of chunk_size directly into gzread and remove the declaration, or would it be better to leave it as is?

erin-le:
> As a final request, I just wanted to make sure you ran some checkpoint tests to verify cp restore still works as expected (I don't see an issue in the code per se but it's better to test this before it gets merged).

For testing, I used `riscv-hello-restore-checkpoint.py` and `simpoints-se-restore.py` from gem5_library and ran them using both gem5 stable and the build for this branch. The diff for stats.txt only showed slight differences with stats such as hostTickRate, hostMemory, hostSeconds, etc., and the diff for standard out only showed that I was running different gem5 versions. Based on this, I believe it should still work properly.

giactra (approved):

PR #1285, opened 2024-06-25
cpu: Add cpuIdlePins to indicate the threadContext of CPU is idle

rogerchang23424:
If the threacContext of CPU enters the suspend mode, raise the threadID of threadContext cpu_idle_pins with the high signal to target. If the threadContext of CPU enters the activate mode, lower the threadID of thread cpu_idle_pins with low signal to target.

### DIFF
--- a/src/cpu/BaseCPU.py
+++ b/src/cpu/BaseCPU.py
@@ -45,6 +45,7 @@ from m5.objects.ClockDomain import *
 from m5.objects.ClockedObject import ClockedObject
 from m5.objects.CPUTracers import ExeTracer
 from m5.objects.InstTracer import InstTracer
+from m5.objects.IntPin import VectorIntSourcePin
 from m5.objects.Platform import Platform
 from m5.objects.ResetPort import ResetResponsePort
 from m5.objects.SubSystem import SubSystem
@@ -154,6 +155,13 @@ class BaseCPU(ClockedObject):
     )
 
     model_reset = ResetResponsePort("Generic reset for the CPU")
+
+    cpu_idle_pins = VectorIntSourcePin(
+        "This signal indicates the thread context of CPU is idle. Should be "
+        "equal to the numThreads or keep empty to ignore idle signals. If the "
+        "user specifies cpu_idle_pins less than numThreads. The ContextId "
+        "greater than the size of cpu_idle_pins will be ignored."
+    )
 
     tracer = Param.InstTracer(default_tracer, "Instruction tracer")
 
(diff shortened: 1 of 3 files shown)


src/cpu/base.cc
```diff
@@ -537,6 +545,11 @@ BaseCPU::activateContext(ThreadID thread_num)
 
     DPRINTF(Thread, "activate contextId %d\n",
             threadContexts[thread_num]->contextId());
+
+    if (thread_num < cpuIdlePins.size()) {
+      cpuIdlePins[thread_num]->lower();
```
giactra: 4 whitespaces needed

src/cpu/base.cc
```diff
@@ -551,6 +564,11 @@ BaseCPU::suspendContext(ThreadID thread_num)
 {
     DPRINTF(Thread, "suspend contextId %d\n",
             threadContexts[thread_num]->contextId());
+
+    if (thread_num < cpuIdlePins.size()) {
+      cpuIdlePins[thread_num]->raise();
```
giactra: Ditto

src/cpu/BaseCPU.py
```diff
@@ -155,6 +156,10 @@ def takeOverFrom(self, old_cpu):
 
     model_reset = ResetResponsePort("Generic reset for the CPU")
 
+    cpu_idle_pins = VectorIntSourcePin(
+        "This signal indicates the thread context of CPU is idle"
```
giactra: It would be nice if we could document this further. For example, from an initial look at the way it is used it seems like there should be a 1:1 relation between pins and number of HW threads. However it looks like we allow for "silent" threads:


    if (thread_num < cpuIdlePins.size()) {
      cpuIdlePins[thread_num]->lower();
      
Means we can generate less pins and related threads won't generate any idle output... I would like this to be specified in the Param help string

src/cpu/base.cc
```diff
@@ -191,6 +191,12 @@ BaseCPU::BaseCPU(const Params &p, bool is_checker)
     modelResetPort.onChange([this](const bool &new_val) {
         setReset(new_val);
     });
+
+    for (int i = 0; i < params().port_cpu_idle_pins_connection_count; i++) {
+      cpuIdlePins.emplace_back(new IntSourcePin<BaseCPU>(
```
giactra: 4 whitespace needed

giactra (changes requested):

src/cpu/BaseCPU.py
rogerchang23424: Please check the Param helping description, Thanks

giactra (approved):
Thanks for your changes

PR #1292, opened 2024-06-27
util-docker: Cleanup, refactor, better document Dockerfiles

BobbyRBruce:
* Removes the "docker-compose.yaml" in favor of "docker-bake.hcl". This uses the `docker buildx` tool which has the advantage of enabling multi-platformm builds where desired. By default all images are built targeting `linux/arm64`, `linux/amd64` and `linux/riscv64` as targets with the exception of the GPU images where only `linux/amd64` makes sense.
* Remove unused/older Docker build targets (these can easily be re-added but they were not regularly built or have any current usage).
* Update "README.md" to better describe these Dockerfiles and how they are built.
* Simplify GCC and Clang compiler images. Each uses the Ubuntu 24.04 All Deps image as a base then specialized the compiler on top.
   * To simply things, all compiler versions are built from 24.04. This means **narrowing the supported versions from GCC v10 to v14 and Clang v14 to v18**.
*  Fix some bugs in the "docker-bake.hcl" thus ensuring all targets may be built from it.
* Cleanup the systemc and sst images: reducing their size and building them off the common 24.04 ubuntu base image.


util/dockerfiles/docker-bake.hcl
```diff
@@ -35,59 +35,147 @@ variable "TAG" {
   default = "latest"
 }
 
+# Common attributes across all targets. Note: these can be overwritten.
+target "common" {
+  # Here we are enabling multi-platform builds. We are compiling to both ARM
+  # amd X86.
```
powerjg: This comment doesn't match the next line which also has RISC-V.

Does it make sense to build these dockerfile images for RISC-V? Right now, I don't know of anyone with RISC-V hardware fast enough to run and test gem5.

util/dockerfiles/docker-bake.hcl
```diff
@@ -35,59 +35,147 @@ variable "TAG" {
...
+  }
+  context = "gcc-compiler"
+  tags = ["${IMAGE_URI}/gcc-version-13:${TAG}"]
+}
+
+group "ubuntu-releases" {
+  targets=[
+    "ubuntu-24-04_all-dependencies",
+    "ubuntu-22-04_all-dependencies",
+    "ubuntu-20-04_all-dependencies",
```
powerjg: Is this still supported now that we don't support gcc 8? Or, do we have something special in the dockerfile to upgrade GCC?

util/dockerfiles/docker-bake.hcl
```diff
@@ -126,79 +206,40 @@ target "ubuntu-24-04_min-dependencies" {
...
-    version = ver
-  }
-  tags = ["${IMAGE_URI}/clang-version-${ver}:${TAG}"]
+  context = "gcn-gpu"
+  tags = ["${IMAGE_URI}/gcn-gpu:${TAG}"]
 }
 
-target "clang-compilers-base-22-04" {
-  name = "clang-compilers-${replace(ver, ".", "-")}"
+target "gpu-fs" {
```
powerjg, line 209: What is this docker image used for?

powerjg (commented):
A couple of small things.

util/dockerfiles/docker-bake.hcl
BobbyRBruce: I think i've been hesitant to drop this because i'm not really sure if I should be randomly deciding what we support, but effectively we don't support 20_04. I'll push a patch to this chain to just get rid of it.

util/dockerfiles/docker-bake.hcl
BobbyRBruce: Huh... I just I don't know.

@mattsinc @abmerop : I was, at some point, given this docker to add to our system. Is this still used?

util/dockerfiles/docker-bake.hcl
BobbyRBruce: Done.

util/dockerfiles/docker-bake.hcl
BobbyRBruce: I included it because it really costs us next-to-nothing. It's not a big extra step and I quite like supporting all three in theory. Yes, not high utility for now, but's there.

BobbyRBruce: I updated the comment.

util/dockerfiles/docker-bake.hcl
mattsinc: I believe 22.04 is still used, but @abmerop can confirm

util/dockerfiles/docker-bake.hcl
mattsinc: I suspect, but don't know for sure, that GPUSE will not work anymore if 20.04 is removed -- because the versions of gcc in newer versions don't work with ROCm 4.  But I guess we can run this test and see.

util/dockerfiles/docker-bake.hcl
BobbyRBruce: @mattsinc : I'm pretty sure Jason is asking about the "gpu-fs" docker image. The 22-04 image still exists.

BobbyRBruce: Actually, i can't even build this image. I get `Unable to locate package rocm-dev`. @mattsinc : you last updated the "gpu-fs/Dockerfile", I think you may have broken it in the process?

BobbyRBruce: Nevermnind, it started building for some reason... I don't know why...

abmerop: This is used to build GPU applications. It contains the GPU compiler and libraries. You can think of it like a "cross compiler" docker.

BobbyRBruce: As long as you still use it, that's good enough for me.

util/dockerfiles/README.md
```diff
@@ -1,3 +1,75 @@
...
+All the currently supported Docker images, stored in the registery, are built using these Dockerfiles.
+
+### Docker buildx
+
+The Dockerfiles are built using the Docker buildx feature. The buildx feature is used to build multi-architecture images. The buildx feature is available in Docker 19.03 and later versions.
+
+For more information on the Docker buildx feature, refer to the [Docker documentation](https://docs.docker.com/buildx/working-with-buildx/).
+
+In this setup we store the buildx configurations in the "docker-bake.hcl" file.
+It is worth contsulting this files and noting the "targerts" and "groups" these can be passed to the buildx command to build that target image or group of images.
```
Harshil2107:
```suggestion
It is worth consulting these files and noting the "targets" and "groups", these can be passed to the buildx command to build that target image or group of images.
```

Harshil2107 (dismissed):
Found a spelling mistake in the readme, apart from that LGTM.

ivanaamit (dismissed):

util/dockerfiles/sst/Dockerfile
```diff
@@ -24,52 +24,45 @@
...
-    tar xf sstelements-13.0.0.tar.gz;
-WORKDIR /sst/sst-elements-library-13.0.0/
-RUN ./configure --prefix=$SST_CORE_HOME --with-python=/usr/bin/python3-config \
-                                        --with-sst-core=$SST_CORE_HOME; \
-    make all -j $(nproc); \
-    make install;
-
-# Setting the environmental variables
-ENV PATH=$PATH:$SST_CORE_HOME/bin
-ENV PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$SST_CORE_HOME/lib/pkgconfig/
```
Harshil2107, line 75: Removing this set up, is most likely the cause of the failing daily test with this new docker image.

ivanaamit (approved):

PR #1304, opened 2024-07-01
arch-arm: support 64-bit PMCCNTR from AArch32

arichardson:
For ARMv8 CPUs this register allows reading a 64-bit cycle counter in from 32-bit execution state.

Change-Id: I7cd9e2711ada5156920440cc3c89e7a74ca54a49


giactra (approved):
LGTM, Thanks for this fix!

PR #1305, opened 2024-07-01
misc, tests: Fix missing 's' in GPU tests

BobbyRBruce:
This caused the weekly tests to fail. It's 'tests' not 'test'.

Change-Id: Ib01d2c4cc3aa51c137d13453ef8c8628f57c9adf


mattsinc (commented):
Sorry, just getting to this now -- thanks for fixing

PR #1312, opened 2024-07-02
arch-riscv: add rv32 option to FS Linux config file

cyyself:
Since we have supported RISC-V 32, add this option to allow the RISC-V 32 full system to run easily.


rogerchang23424 (approved):

hnpl (approved):
LGTM.

powerjg (approved):
I'll approve this, but I don't like extending *example* scripts to have lots of parameters. IMO, if you want to build a different system, you should create *a new script* and specialize the standard library board. (E.g., maybe we should have a `RISCVBoardRV32` in `gem5.components.boards`. I would strongly support a PR to add a board (and workloads!) for 32 bit RISC-V systems.)

I want to emphasize that python scripts in the `configs/examples` directory are *examples* of how you can write scripts to use gem5. These files are **not** the interface to gem5.

PR #1316, opened 2024-07-02
arch-riscv: use sign-extend for all address generation

cyyself:
In gem5, we use the same code base for RISC-V 32 and 64.

However, if we need to allow modifiable XLEN control on CSR.mstatus in the future, we should follow the RISC-V ISA manual to sign-extend all the register results, including PC and GPR. If this feature implemented, the simulator needs to handle user-mode in RV32 but CSR.SATP sets to Sv39. In this case, 0x80000000 and 0xffffffff80000000 are different addresses in the 64-bit S-Mode perspective, but they are the same in the 32-bit U-Mode perspective. We should avoid this wrong behavior happening before we implement this feature.

Thus, we need to sign-extend the results of all the addresses, including the PC and memory addresses, which currently use zero-extend. As specified in the RISC-V ISA manual, we use zero-extend in narrow XLEN mode for the physical address implemented in TLB.

Changes based on spec:
1. Sign-extend narrow XLEN: https://github.com/riscv/riscv-isa-manual/blob/riscv-isa-release-b7a445a-2024-07-02/src/machine.adoc?plain=1#L567
2. Zero-extend physical address: https://github.com/riscv/riscv-isa-manual/blob/riscv-isa-release-b7a445a-2024-07-02/src/supervisor.adoc?plain=1#L1670

### DIFF
--- a/src/arch/riscv/tlb.cc
+++ b/src/arch/riscv/tlb.cc
@@ -385,9 +385,16 @@ TLB::translate(const RequestPtr &req, ThreadContext *tc,
         if (fault == NoFault) {
             if (req->getFlags() & Request::PHYSICAL) {
                 /**
-                 * we simply set the virtual address to physical address
+                 * we simply set the virtual address to physical address.
+                 *
+                 * For RV32, we follow what the specification said:
+                 * When mapping between narrower and wider addresses,
+                 * RISC-V zero-extends a narrower physical address to a
+                 * wider size.
                  */
-                req->setPaddr(req->getVaddr());
+                req->setPaddr(((ISA*) tc->getIsaPtr())->rvType() == RV32 ?
+                              bits(req->getVaddr(), 31, 0) :
+                              req->getVaddr());
             } else {
                 fault = doTranslate(req, tc, translation, mode, delayed);
             }
@@ -418,9 +425,20 @@ TLB::translate(const RequestPtr &req, ThreadContext *tc,
 
         Process * p = tc->getProcessPtr();
 
-        Fault fault = p->pTable->translate(req);
-        if (fault != NoFault)
-            return fault;
+        /*
+         * In RV32 Linux, as vaddr >= 0x80000000 is legal in userspace
+         * (except for COMPAT mode for RV32 Userspace in RV64 Linux), we
+         * need to ignore the upper bits beyond 32 bits.
+         */
+        Addr vaddr = ((ISA*) tc->getIsaPtr())->rvType() == RV32 ?
+                      bits(req->getVaddr(), 31, 0) :
+                      req->getVaddr();
+        Addr paddr;
+
+        if (!p->pTable->translate(vaddr, paddr))
+            return std::make_shared<GenericPageTableFault>(req->getVaddr());
+
+        req->setPaddr(paddr);
 
         return NoFault;
     }
--- a/src/arch/riscv/isa/decoder.isa
+++ b/src/arch/riscv/isa/decoder.isa
@@ -69,7 +69,7 @@ decode QUADRANT default Unknown::unknown() {
 
                 Fp2_bits = Mem;
             }}, {{
-                EA = rvZext(Rp1 + offset);
+                EA = rvSext(Rp1 + offset);
             }});
             0x2: c_lw({{
                 offset = CIMM2<1:1> << 2 |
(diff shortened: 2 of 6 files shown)


hnpl:
Thanks for the contribution.

I think, for every instruction produces an effective address (EA), we'd need a check whether we are in the machine mode to decide whether we want to zero extend or sign extend the EA. My assumption is that the machine mode in RISC-V works directly on physical address space.

Since we're not modifying XLEN during a simulation in gem5, sign extension and zero extension are effectively the same. However, this is a quite big change so I think it needs more testing. @Harshil2107, if you have time, can you test this on RISC-V FS/SE modes using rv32/rv64?

src/arch/riscv/tlb.cc
```diff
@@ -385,9 +385,16 @@ TLB::translate(const RequestPtr &req, ThreadContext *tc,
...
+                 * we simply set the virtual address to physical address.
+                 *
+                 * For RV32, we follow what the specification said:
+                 * When mapping between narrower and wider addresses,
+                 * RISC-V zero-extends a narrower physical address to a
+                 * wider size.
                  */
-                req->setPaddr(req->getVaddr());
+                req->setPaddr(((ISA*) tc->getIsaPtr())->rvType() == RV32 ?
+                              req->getVaddr() & 0xffffffffLL :
```
powerjg: If we're going to do this, let's create a function in case there are other places where the logic needs to be replicated. Also, please use `bits(31,0)` or something similar instead of manually masking.

src/arch/riscv/tlb.cc
```diff
@@ -418,9 +425,20 @@ TLB::translate(const RequestPtr &req, ThreadContext *tc,
...
-        if (fault != NoFault)
-            return fault;
+        /*
+         * In RV32, as vaddr >= 0x80000000 is legal in userspace (except for
+         * COMPAT mode for RV32 Userspace in RV64 Linux), we need to ignore
+         * the upper bits beyond 32 bits.
+         */
+        Addr vaddr = ((ISA*) tc->getIsaPtr())->rvType() == RV32 ?
+                      req->getVaddr() & 0xffffffffLL :
+                      req->getVaddr();
```
powerjg: Should use the function mentioned above.

src/arch/riscv/tlb.cc
```diff
@@ -418,9 +425,20 @@ TLB::translate(const RequestPtr &req, ThreadContext *tc,
 
         Process * p = tc->getProcessPtr();
 
-        Fault fault = p->pTable->translate(req);
-        if (fault != NoFault)
-            return fault;
+        /*
+         * In RV32, as vaddr >= 0x80000000 is legal in userspace (except for
```
powerjg: Is this part of the ISA spec or is it Linux specific?

powerjg (commented):
I have a small software engineering comment below. However, I'm a bit concerned with thes change in general. It should definitely be tested in FS and SE mode. I just spent a bunch of time fixing a similar issue...

src/arch/riscv/tlb.cc
cyyself: Oh. Indeed, it's Linux-specific. I will fix it soon.

src/arch/riscv/tlb.cc
cyyself: Fixed.

rogerchang23424:
Since we changed the all generated address to sign-extend. Should we need to sign-ext the address for the RV32 mode only instruction? `c.jal`, `c.flw`, `c.fsw`, `c.flwsp`, `c.fswsp`.

https://github.com/gem5/gem5/blob/ebfb8999cb2c6a3ef87146515dd2cc5d097bc7bc/src/arch/riscv/isa/decoder.isa#L214-L228

https://github.com/gem5/gem5/blob/ebfb8999cb2c6a3ef87146515dd2cc5d097bc7bc/src/arch/riscv/isa/decoder.isa#L83-L101

cyyself:
> Since we changed the all generated address to sign-extend. Should we need to sign-ext the address for the RV32 mode only instruction? `c.jal`, `c.flw`, `c.fsw`, `c.flwsp`, `c.fswsp`.
> 

Sure. I forgot this. I will fix them soon.

rogerchang23424 (approved):
LGTM

cyyself:
Kindly ping

powerjg:
Thanks for the reminder. It may be ~10 days before we (at UC Davis) have time to triage and go through PRs/issues. We're busy prepping for the [bootcamp](https://www.gem5.org/events/bootcamp-2024)

BobbyRBruce (approved):
All the requested changes appear to have been addressed. This is a thumbs-up from me. I'll leave it to @powerjg to merge.

PR #1319, opened 2024-07-02
arch-riscv: Fix implicit int-to-float conversion in .isa files

arichardson:
Explicitly convert to float/double to fix compiler warnings that I have turned on locally.

Change-Id: I09e1aca0048bc990c49906778e81276c0c4ebb80


hnpl (approved):
LGTM.

`1 << vlmul` outputs a 32-bit result while `vlmul` is a 64-bit type, though I don't think it matters in this case as correct values of `vlmul` and `numFields` should not result in integer overflow.

BobbyRBruce (approved):

PR #1324, opened 2024-07-03
arch-arm: Fix decoding of unconditional AArch32 SIMD instructions

arichardson:
While running some AArch32 code, I noticed that the output did not match QEMU or real hardware whenever I made the compiler target 32-bit ARMv8 (compiling for ARMv7 yielded correct results). After spending a long time comparing gem5 vs QEMU logfiles I noticed that we were decoding vselect.f32 as VCMLA and therefore giving an incorrect result that only affected control flow in some specific configurations. It turns out ae69a12b3bd40bc86997c96cac9239f2f973aa95 changed the way ARMv8 SIMD/FP instructions were handled and caused incorrect decoding of instructions such as vrintp/veselect/etc.

See https://developer.arm.com/documentation/ddi0597/2024-03/A32-Instructions-by-Encoding/System-register-access--Advanced-SIMD--floating-point--and-Supervisor-call?lang=en#advsimdext

Fixes: ae69a12b3bd40bc86997c96cac9239f2f973aa95
Change-Id: Iee839fc3a343b2d102bc774fe541059c865cc700

### DIFF
--- a/src/arch/arm/isa/formats/fp.isa
+++ b/src/arch/arm/isa/formats/fp.isa
@@ -98,7 +98,10 @@ let {{
     decodeNeonData(ExtMachInst machInst);
 
     StaticInstPtr
-    decodeAdvancedSIMD(ExtMachInst machInst);
+    decodeUncondAdvancedSIMD(ExtMachInst machInst);
+
+    StaticInstPtr
+    decodeFloatingPointDataProcessing(ExtMachInst machInst);
     '''
 
     decoder_output = '''
@@ -338,11 +341,8 @@ let {{
     '''
     decoder_output += '''
     StaticInstPtr
-    decodeAdvancedSIMD(ExtMachInst machInst)
+    decodeUncondAdvancedSIMD3Reg(ExtMachInst machInst)
     {
-        uint8_t op_code = (bits(machInst, 25) << 1)
-                          | bits(machInst, 21);
-
         RegIndex vd = (RegIndex)(2 * (bits(machInst, 15, 12) |
                                (bits(machInst, 22) << 4)));
         RegIndex vn = (RegIndex)(2 * (bits(machInst, 19, 16) |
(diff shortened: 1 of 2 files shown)


ivanaamit:
@giactra, could you review this PR? You reviewed a couple of others, and this one might have slipped through. Thanks.

giactra:
> @giactra, could you review this PR? You reviewed a couple of others, and this one might have slipped through. Thanks.

Hi Ivana, I didn't comment but I actually did an internal review, in the sense that I tried this patch with some internal tests and I can see there are some new failures. My guess is that while it fixes part of the decode tree, it breaks another. Which is not surprising considering how convoluted the AAch32 decode tree is. I will try to debug the new failures and hopefully will come back to @arichardson with a fix

ivanaamit:
@giactra, Alexander created a separate PR with a fix. Is that the one you were referring to? Can we go ahead and merge this PR? Thanks.

arichardson:
Alternatively if you have an example of a new decoding failure I can address it here

ivanaamit:
@giactra Could you please update us on the status of this PR? Can it be merged, or are there any changes needed? Thanks.

arichardson:
ping? Do you have an example encoding that is incorrect with this patch?

ivanaamit:
@giactra, any feedback would be greatly appreciated. Thanks!

giactra:
Hi all, apologies for my late reply. I will try to come up with some examples/information sometime next week

erin-le:
@giactra Hello, could you provide another update on the status of this on your end? Thank you!

giactra:
Apologies, it took me some time to extract some data. Basically this PR is breaking the decoding of some VCMLAs. For example 0xfe2f78ae would not be recognised as a valid instruction with this PR

arichardson:
> Apologies, it took me some time to extract some data. Basically this PR is breaking the decoding of some VCMLAs. For example 0xfe2f78ae would not be recognised as a valid instruction with this PR

Thanks for the example, let me dig into this next week

erin-le:
@arichardson Hello, thank you for your contribution. Could you give an update on what the status of this is on your end?

arichardson:
I am currently extremely busy with other things but I hope to get back to this soon. Sorry!

erin-le:
@arichardson This PR is being closed for inactivity, but can be reopened if you would like to continue working on it. Thank you!

arichardson:
> 0xfe2f78ae

@giactra I just tried to reproduce this - but LLVM does not recognize this as a valid instruction:

`fe2f 78ae     <unknown>`

PR #1325, opened 2024-07-03
arch-arm: Fix incorrect operation of VRINT* instructions

arichardson:
After a lot of debugging and comparing traces I noticed that vrintp was giving different results from QEMU. An input of 0x3f800000 (1.0) was being passed to the fplib helpers as (uint32_t)1 which has a completely different floating-point interpretation and the result was therefore completely wrong.

I've fixed this as well as all remaining implicit float-to-int conversions in the ARM instruction execution. There are more -W(implicit-)float-conversion warnings in the other executors, but for now this fixes the issue I was seeing.

Change-Id: Ifdeee745ca155d7f4504ac4c54235ac431acdeb9

### DIFF
--- a/src/arch/arm/isa/insts/fp.isa
+++ b/src/arch/arm/isa/insts/fp.isa
@@ -469,7 +469,9 @@ let {{
 
     singleSimpleCode = vfpEnabledCheckCode + '''
         [[maybe_unused]] FPSCR fpscr = (FPSCR) FpscrExc;
-        FpDest = %(op)s;
+        static_assert(std::is_same_v<decltype(%(op)s), uint32_t>,
+                      "operation triggers invalid implicit conversion");
+        FpDest_uw = %(op)s;
     '''
     singleCode = singleSimpleCode + '''
         FpscrExc = fpscr;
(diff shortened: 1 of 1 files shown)


src/arch/arm/isa/insts/fp.isa
```diff
@@ -485,14 +487,19 @@ let {{
...
+                  "%(func)s, fpscr.fz, fpscr.dn, fpscr.rMode))"
+    singleUnaryOp = "floatToBits32(unaryOp(fpscr, FpOp1, " + \
+                    "%(func)s, fpscr.fz, fpscr.rMode))"
     doubleCode = vfpEnabledCheckCode + '''
         [[maybe_unused]] FPSCR fpscr = (FPSCR) FpscrExc;
-        double dest = %(op)s;
-        FpDestP0_uw = dblLow(dest);
-        FpDestP1_uw = dblHi(dest);
+        [[maybe_unused]] uint64_t cOp1 = ((uint64_t)FpOp1P0_uw |
+                                          ((uint64_t)FpOp1P1_uw << 32));
```
giactra: I understand adding this here avoids some repetition, but we need to be very careful about introducing unnecessary reads. This is not only a simulator performance problem; when we use these magic operands, we inform the CPU the instruction has a Read dependency over that register. This will affect the dependency graph and it will pollute dynamic scheduling with some unnecessary serialization (If a RAW is present).

Only code making use of FpOp1P0 and FpOp1P1 should read them and the maybe_unused scares me a bit.

giactra (commented):
Thanks for fixing this!

src/arch/arm/isa/insts/fp.isa
arichardson: I will update this PR to avoid maybe_unused by adding this call to the vrint() helpers.

arichardson:
Sorry for the delay, I was travelling. Updated now.

arichardson:
Ping?

ivanaamit:
@giactra, could you please take another look at this PR? It seems like it might be ready for merging. Thanks!

giactra (approved):

PR #1326, opened 2024-07-03
arch-arm: Fix implicit int-to-float conversion in VCMP

arichardson:
Explicitly convert to float/double to fix compiler warnings that I have turned on locally. It might make sense to make use of fplib functions to be portable across different host float formats but something as simple as comparison against zero should be safe.

Change-Id: I96c6ee7c5497fece11be07234ff80ff86e7555e2

### DIFF
--- a/src/arch/arm/isa/insts/fp.isa
+++ b/src/arch/arm/isa/insts/fp.isa
@@ -1360,13 +1360,12 @@ let {{
     vcmpZeroSCode = vfpEnabledCheckCode + '''
         FPSCR fpscr = (FPSCR) FpscrExc;
         vfpFlushToZero(fpscr, FpDest);
-        // This only handles imm == 0 for now.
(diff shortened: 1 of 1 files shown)


giactra (changes requested):
Hi @arichardson thanks for this contribution. 
I believe one of the reasons why the code was written that way (using an imm instead of just comparing to zero) was to be able to reuse the snippet in other scenarios (with imm != 0). Can we just compare directly with zero? Ideally we should also change the instruction definition to not use an immediate but it is not worth the effort

ivanaamit:
@arichardson, could you please address Giacomo's comment? Thank you.

ivanaamit:
@arichardson, just a reminder to address Giacomo's review. Thanks!

arichardson:
Sorry for the delay, updated now.

giactra (approved):

PR #1327, opened 2024-07-03
arch-arm: Fix remaining implicit float conversion warnings in .isa

arichardson:
This fixes the remaining implicit int/float conversions and enables the float conversion warnings for clang when building the Arm instruction execution logic. This depends on the previous fixes.

Change-Id: I51aac94644a483175842c36da2d49d308aaceb49


giactra:
This looks good to me, I would only ask you if you could update the commit title to be "arch,arch-arm" as you are also changing generic code (the isa parser)

ivanaamit (approved):

PR #1329, opened 2024-07-03
sim: Add error message for kernel exceeding memory size

erin-le:
This commit adds an error message to src/sim/kernel_workload.cc to tell the user when the end address of the kernel is greater than the size of memory. The error message also specifies the minimum memory size needed to fit the kernel.                                                            

Change-Id: I7d8f50889ed8172f64b84f98301a35e5f2f352d3

### DIFF
--- a/src/sim/kernel_workload.cc
+++ b/src/sim/kernel_workload.cc
@@ -104,12 +104,22 @@ KernelWorkload::initState()
     };
     if (params().object_file != "")  {
         if (params().addr_check) {
+            auto memory_size = system->memSize();
+            auto kernel_size = _end - _start;
             // Validate kernel mapping before loading binary
+            fatal_if((kernel_size > memory_size),
+                "Kernel is mapped to invalid location (not memory). start"
+                " (%#lx) - end (%#lx) %#lx:%#lx\n"
+                "The kernel size (%ld bytes, %ld MB) appears to be larger "
+                "than total system memory (%ld bytes, %ld MB). Try increasing"
+                " system memory.", _start, _end, mapper(_start), mapper(_end),
+                kernel_size, kernel_size>>20, memory_size, memory_size>>20);
             fatal_if(!system->isMemAddr(mapper(_start)) ||
-                    !system->isMemAddr(mapper(_end)),
-                    "Kernel is mapped to invalid location (not memory). "
-                    "start (%#x) - end (%#x) %#x:%#x\n",
-                    _start, _end, mapper(_start), mapper(_end));
+                !system->isMemAddr(mapper(_end)),
+                "Kernel is mapped to invalid location (not memory). start"
+                " (%#lx) - end (%#lx) %#lx:%#lx\n", _start, _end,
+                mapper(_start), mapper(_end));
+
         }
         // Load program sections into memory
         image.write(phys_mem);


src/sim/kernel_workload.cc
```diff
@@ -102,12 +102,29 @@ KernelWorkload::initState()
...
+                    inform("Increase your memory size to at least %dMB to fit"
+                    " the kernel.\n", (_end>>20) + 1);
+                }
+                else{
+                    inform("Increase your memory size to at least %dMB to fit"
+                    " the kernel.\n", _end>>20);
+                }
+            }
             // Validate kernel mapping before loading binary
             fatal_if(!system->isMemAddr(mapper(_start)) ||
```
giactra, line 117: Unless there's a specific case I am missing, most of the code you added is already covered here... When a fatal is encountered a user can use this info to increase the memory size. This feels a bit like duplication IMHO. I understand "Kernel is mapped to invalid location (not memory)" is not particularly descriptive so maybe we should work more on making things clearer for a programmer

src/sim/kernel_workload.cc
```diff
@@ -102,12 +102,29 @@ KernelWorkload::initState()
     auto mapper = [this](Addr a) {
         return (a & _loadAddrMask) + _loadAddrOffset;
     };
+    auto memory_size = system->memSize();
     if (params().object_file != "")  {
         if (params().addr_check) {
+            // Check that memory size is large enough to contain the kernel.
+            if (memory_size < _end){
```
giactra: My gut feeling is that there might be something wrong here in some cases.... for example what happens if the image starts at an address different than 0? then you might end up with an _end bigger than memory size even if you have allocated enough memory..

giactra (commented):
Hi, please see my comments below

src/sim/kernel_workload.cc
BobbyRBruce: I was the one of actually asked @erin-le to do this. My main complaint with the current error is while it correctly reports the error, a subset of this error is undoubtably the case where there simply isn't enough memory to load the kernel and in this case it'd be useful to have this communicated directly. It's a simple mistake for a beginner to make and the error returned might be hard for some to fully understand. I think determining this specific bug and specializing the error returned is easily doable and worthwhile.

Would you prefer if the more specific error was appended to the "Kernel is mapped to invalid location"? I've suggested this in my other comment. It could keep things more concise.

src/sim/kernel_workload.cc
BobbyRBruce: I think this is fair, would it be reasonable to quickly determine the image size somewhere (taking into account start and end --- we can assume kernel memory is contiguous?). Then down in the fatal-if we do:

```cpp
!system->isMemAddr(mapper(_end)),
    "Kernel is mapped to invalid location (not memory). " +
    kernel_size > system->memSize() ? " The kernel size (" +   
    kenel_size + " blocks) appears to be larger than total system memory " + system->memSize() + " blocks). "
    "Try increasing system memory" : ""
```

(I'm sure this isn't real CPP so please consider if Pseudocode).

The rest of the error we can remove and keep the error in one line.

giactra: This looks like a good solution to me!

BobbyRBruce:
@erin-le : Here's a quick basic breakdown of what I think needs done to satisfy the reviwers here:

1. Just have one error (the current error message), and design it so there is an additional line added in the case the kernel size is bigger than the system memory therefore suggesting this is probably the issue.
2. Calculate the kernel size using the start and end addresses (not just the end). 

I think general logic should flow like this:

1. If `!system->isMemAddr(mapper(_start)`, print the corrent error.
2. If, in addition, the kernel size (based on start and end address) appears bigger than total system memory append to the string a hint that it appears the kernel is bigger than total system memory. Include in this string the sizes involved, if possible.

src/sim/kernel_workload.cc
```diff
@@ -104,12 +104,28 @@ KernelWorkload::initState()
     };
     if (params().object_file != "")  {
         if (params().addr_check) {
+            auto memory_size = system->memSize();
+            auto kernel_size = _end - _start;
             // Validate kernel mapping before loading binary
+            char error_msg_buffer [500];
```
giactra: I don't see the reason for shooting ourselves in the foot with a fixed size array

src/sim/kernel_workload.cc
```diff
@@ -104,12 +104,28 @@ KernelWorkload::initState()
...
+                "The kernel size (%ld bytes, %ld MB) appears to be larger "
+                "than total system memory (%ld bytes, %ld MB). Try increasing"
+                " system memory.", _start, _end, mapper(_start), mapper(_end),
+                kernel_size, kernel_size>>20, memory_size, memory_size>>20);
+            }
+            else{
+                snprintf(error_msg_buffer, 500, "Kernel is mapped to invalid"
+                "location (not memory). start (%#lx) - end (%#lx) %#lx:%#lx\n",
+                    _start, _end, mapper(_start), mapper(_end));
+            }
```
giactra: This seems to me a bit too convoluted... First we populate the error message and then we print it but only if the fatal condition is met? This means the second snprintf will happen even if there is no issue whatsoever. I prefer Bobby solution of handling everything with ternary operator... Alternatively I guess am fine on having 2 fatal_ifs

giactra (changes requested):
Thanks @erin-le for your changes! I believe the solution should be reworked a bit I am afraid

src/sim/kernel_workload.cc
erin-le: I wasn't able to get the ternary operator to work (I couldn't vary the number of inputs to the format string and having too many caused <bad format> to be printed), so I used 2 fatal_ifs.

src/sim/kernel_workload.cc
```diff
@@ -104,12 +104,28 @@ KernelWorkload::initState()
     };
     if (params().object_file != "")  {
         if (params().addr_check) {
+            auto memory_size = system->memSize();
+            auto kernel_size = _end - _start;
             // Validate kernel mapping before loading binary
-            fatal_if(!system->isMemAddr(mapper(_start)) ||
+            if (kernel_size > memory_size){
+                fatal_if(!system->isMemAddr(mapper(_start)) ||
                     !system->isMemAddr(mapper(_end)),
```
giactra: I think there is a misunderstanding. The kernel_size > memory size should be the fatal_if condition. No need to have an extra if-else. So the code should be something like:

fatal_if(kernel_size > memory_size, [...]);
fatal_if(system->isMemAddr[...]);

giactra (changes requested):

src/sim/kernel_workload.cc
```diff
@@ -104,12 +104,24 @@ KernelWorkload::initState()
...
+            auto kernel_size = _end - _start;
             // Validate kernel mapping before loading binary
+            fatal_if((kernel_size > memory_size),
+            "Kernel is mapped to invalid location (not memory). start"
+            " (%#lx) - end (%#lx) %#lx:%#lx\n"
+            "The kernel size (%ld bytes, %ld MB) appears to be larger "
+            "than total system memory (%ld bytes, %ld MB). Try increasing"
+            " system memory.",_start, _end, mapper(_start), mapper(_end),
+            kernel_size, kernel_size>>20, memory_size, memory_size>>20
+            );
```
giactra: This shouldn't be on a newline

src/sim/kernel_workload.cc
```diff
@@ -104,12 +104,24 @@ KernelWorkload::initState()
...
+            auto memory_size = system->memSize();
+            auto kernel_size = _end - _start;
             // Validate kernel mapping before loading binary
+            fatal_if((kernel_size > memory_size),
+            "Kernel is mapped to invalid location (not memory). start"
+            " (%#lx) - end (%#lx) %#lx:%#lx\n"
+            "The kernel size (%ld bytes, %ld MB) appears to be larger "
+            "than total system memory (%ld bytes, %ld MB). Try increasing"
+            " system memory.",_start, _end, mapper(_start), mapper(_end),
+            kernel_size, kernel_size>>20, memory_size, memory_size>>20
```
giactra: This should have 4 white spaces indentation wrt the fatal_if

src/sim/kernel_workload.cc
```diff
@@ -104,12 +104,24 @@ KernelWorkload::initState()
...
         if (params().addr_check) {
+            auto memory_size = system->memSize();
+            auto kernel_size = _end - _start;
             // Validate kernel mapping before loading binary
+            fatal_if((kernel_size > memory_size),
+            "Kernel is mapped to invalid location (not memory). start"
+            " (%#lx) - end (%#lx) %#lx:%#lx\n"
+            "The kernel size (%ld bytes, %ld MB) appears to be larger "
+            "than total system memory (%ld bytes, %ld MB). Try increasing"
+            " system memory.",_start, _end, mapper(_start), mapper(_end),
```
giactra: Please add a whitespace after the comma

src/sim/kernel_workload.cc
```diff
@@ -104,12 +104,24 @@ KernelWorkload::initState()
...
             fatal_if(!system->isMemAddr(mapper(_start)) ||
-                    !system->isMemAddr(mapper(_end)),
-                    "Kernel is mapped to invalid location (not memory). "
-                    "start (%#x) - end (%#x) %#x:%#x\n",
-                    _start, _end, mapper(_start), mapper(_end));
+                !system->isMemAddr(mapper(_end)),
+            "Kernel is mapped to invalid location (not memory). start"
+            " (%#lx) - end (%#lx) %#lx:%#lx\n", _start, _end,
+            mapper(_start), mapper(_end)
+            );
```
giactra: Ditto

giactra (changes requested):
We are almost ready to merge this. Can I ask you to squash the 4 commits into 1 and force push the PR?

giactra (approved):
Thanks for your changes

giactra:
Hi @erin-le 

I was about to merge your PR but I noticed it would be under

122753811+erin-le@users.noreply.github.com 

I believe you should amend some config so that it gets authored with your real name/surname/email. @powerjg do you know how to do that?

powerjg:
🤷 If you `git commit --amend` on your local machine, it *should* show up correctly, but I'm not certian.

We really haven't been trying very hard to get canonical emails when committing. My view is that it doesn't really matter. If our goal is tracking things, we can fix it in the .mailmerge file. If our goal is anything to do with copyright, that ship sailed years and years ago.

giactra:
> 🤷 If you `git commit --amend` on your local machine, it _should_ show up correctly, but I'm not certian.
> 
> We really haven't been trying very hard to get canonical emails when committing. My view is that it doesn't really matter. If our goal is tracking things, we can fix it in the .mailmerge file. If our goal is anything to do with copyright, that ship sailed years and years ago.

Unfortunately github overrides the commit info if the “Keep my email private” is set in:

https://github.com/settings/emails

@erin-le would you mind checking your configuration by following the link?

erin-le:
I checked my settings and made my email public. If that didn't fix it, I'm fine with using the noreply.github.com email, since it has the name I go by anyways.

PR #1331, opened 2024-07-04
arch-riscv: Improve widening/narrowing vectors overlap check

### LINKED ISSUE #442: RISC-V Vector Extension float32_t bugs/unsupported widening instructions
**Describe the bug**
Hello, tried running some RISC-V RVV code using syscall mode and ran into a couple of unsupported instructions/bugs.

**Affects version**
Develop branch https://github.com/gem5/gem5/commit/486916b5d4339ea68141b11bafdb167176cf2fb4

**gem5 Modifications**
No modifications.

**To Reproduce**
```
scons build/RISCV/gem5.fast -j8
$GEM5/build/RISCV/gem5.fast $GEM5/configs/example/gem5_library/riscv-se.py 
```
Script used for SE mode using RISCVMatchedBoard:
```
from gem5.isas import ISA
from gem5.resources.resource import obtain_resource, BinaryResource

# Run a check to ensure the right version of gem5 is being used.

# Setup the board.
board = RISCVMatchedBoard()
binary = BinaryResource("/path/to/gem5_example")
board.set_se_binary_workload(binary)

simulator = Simulator(board=board, full_system=False)
simulator.run()
print("Exiting @ tick {} because {}.".format(simulator.get_current_tick(), simulator.get_last_exit_event_cause(),))
```
Using QEMU(v8.0) user-mode side-by-side to show expected results.

## 1. Some widening instructions cause panic for LMUL=8: 
Using a dot product example for _vwredsum_, normally most widening instructions don't work with LMUL=8 as you can't widen to LMUL=16, but some of them such as _vwredsum_ and _vwmacc_ should work correctly.
```
#include "riscv_vector.h"
#include "stdio.h"
#include "stdlib.h"
void rvv_dot_prod(int16_t *pSrcA, int16_t *pSrcB, uint32_t n, int64_t *result)
{
    size_t vl;
    vint16m4_t vSrcA, vSrcB;
    vint64m1_t vSum = __riscv_vmv_s_x_i64m1(0, 1);
    while (n > 0) {
        vl = __riscv_vsetvl_e16m4(n);
        vSrcA = __riscv_vle16_v_i16m4(pSrcA, vl);
        vSrcB = __riscv_vle16_v_i16m4(pSrcB, vl);
        vSum = __riscv_vwredsum_vs_i32m8_i64m1(__riscv_vwmul_vv_i32m8(vSrcA, vSrcB, vl), vSum, vl);
        pSrcA += vl;
        pSrcB += vl;
        n -= vl;
    }
    *result = __riscv_vmv_x_s_i64m1_i64(vSum);
}

void scalar_dot_prod(int16_t *pSrcA, int16_t *pSrcB, uint32_t n, int64_t *result)
{
    int64_t sum = 0;
    while (n > 0) {
        sum += ((int64_t)*pSrcA++ * *pSrcB++);
        n--;
    }
    *result = sum;
}

int main(int argc, char **argv)
{
    size_t vlen = 31;
    int16_t *input = (int16_t *)malloc(vlen * sizeof(int16_t));
    int16_t initial_value = 0;
    for (size_t i = 0; i < vlen; ++i) {
        input[i] = initial_value++;
    }
    int64_t scalar_result, vector_result;
    rvv_dot_prod(input, input, vlen, &vector_result);
    scalar_dot_prod(input, input, vlen, &scalar_result);
    printf("RVV: %ld\nScalar: %ld\n", vector_result, scalar_result);
    free(input);
    return 0;
}
```
```
$GEM5/build/RISCV/gem5.fast $GEM5/configs/example/gem5_library/riscv-se.py 

gem5 version DEVELOP-FOR-23.1
gem5 compiled Oct 10 2023 08:30:21
gem5 started Oct 12 2023 10:20:16

Beginning simulation!
src/arch/riscv/linux/se_workload.cc:60: warn: Unknown operating system; assuming Linux.
build/RISCV/arch/riscv/generated/decoder-ns.hh.inc:40614: panic: panic condition vlmul == 3 occurred: LMUL=8 is illegal for widening inst
Memory Usage: 16917496 KBytes
Program aborted at tick 11200518
--- BEGIN LIBC BACKTRACE ---
/home//gem5//build/RISCV/gem5.fast(_ZN4gem515print_backtraceEv+0x30)[0x55e3aa721d90]
/home//gem5//build/RISCV/gem5.fast(_ZN4gem512abortHandlerEi+0x4e)[0x55e3aa73b37e]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x14420)[0x7f4117551420]
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0xcb)[0x7f411671d00b]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x12b)[0x7f41166fc859]
/home//gem5//build/RISCV/gem5.fast(+0xb05885)[0x55e3a9f8d885]
/home//gem5//build/RISCV/gem5.fast(_ZN4gem512RiscvISAInst8Vwmul_vvIjEC2ENS_16bitfield_backend17BitUnionOperatorsINS_8RiscvISA36BitfieldUnderlyingClassesExtMachInstEEE+0x52d)[0x55e3aa3c3bfd]
/home//gem5//build/RISCV/gem5.fast(_ZN4gem58RiscvISA7Decoder10decodeInstENS_16bitfield_backend17BitUnionOperatorsINS0_36BitfieldUnderlyingClassesExtMachInstEEE+0x106a7)[0x55e3aa0fb4c7]
/home//gem5//build/RISCV/gem5.fast(_ZN4gem58RiscvISA7Decoder6decodeENS_16bitfield_backend17BitUnionOperatorsINS0_36BitfieldUnderlyingClassesExtMachInstEEEm+0xa7)[0x55e3a9fe6f37]
/home//gem5//build/RISCV/gem5.fast(_ZN4gem58RiscvISA7Decoder6decodeERNS_11PCStateBaseE+0x11b)[0x55e3a9fe708b]
/home//gem5//build/RISCV/gem5.fast(_ZN4gem55minor6Fetch28evaluateEv+0xda5)[0x55e3aad37715]
/home//gem5//build/RISCV/gem5.fast(_ZN4gem55minor8Pipeline8evaluateEv+0x1d2)[0x55e3aad49222]
/home//gem5//build/RISCV/gem5.fast(+0x12c6567)[0x55e3aa74e567]
/home//gem5//build/RISCV/gem5.fast(_ZN4gem510EventQueue10serviceOneEv+0xa8)[0x55e3aa72e9f8]
/home//gem5//build/RISCV/gem5.fast(_ZN4gem59doSimLoopEPNS_10EventQueueE+0x70)[0x55e3aa751970]
/home//gem5//build/RISCV/gem5.fast(_ZN4gem58simulateEm+0x263)[0x55e3aa751f03]
/home//gem5//build/RISCV/gem5.fast(+0xb347f2)[0x55e3a9fbc7f2]
/home//gem5//build/RISCV/gem5.fast(+0xad2203)[0x55e3a9f5a203]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(+0x26f0c3)[0x7f41177d10c3]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(_PyObject_Call+0x60)[0x7f4117818180]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(_PyEval_EvalFrameDefault+0x5a52)[0x7f41175d9c12]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(+0x1d4424)[0x7f4117736424]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(_PyFunction_Vectorcall+0x9e)[0x7f4117817cbe]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(+0x71d5d)[0x7f41175d3d5d]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(_PyEval_EvalFrameDefault+0x8401)[0x7f41175dc5c1]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(+0x1d4424)[0x7f4117736424]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(_PyFunction_Vectorcall+0x9e)[0x7f4117817cbe]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(+0x71d5d)[0x7f41175d3d5d]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(_PyEval_EvalFrameDefault+0x4198)[0x7f41175d8358]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(+0x1d4424)[0x7f4117736424]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(_PyEval_EvalCodeWithName+0x52)[0x7f4117736772]
/lib/x86_64-linux-gnu/libpython3.9.so.1.0(PyEval_EvalCodeEx+0x42)[0x7f41177367c2]
--- END LIBC BACKTRACE ---
For more info on how to address this issue, please visit https://www.gem5.org/documentation/general_docs/common-errors/ 

Aborted (core dumped)

$ qemu-riscv64 -cpu rv64,v=true,vext_spec=v1.0,vlen=256,elen=64 gem5_example 
RVV: 9455
Scalar: 9455
```

## 2. Some float32 vector instructions return NaNs
Vector sum example using _vfredusum_, also noted similar behavior _vfredmax_. The float64 variants seem to work correctly.
```
#include "riscv_vector.h"
#include "stdio.h"

void vector_sum(float *pSrc, uint32_t n, float *pResult)
{
    size_t vl;
    vfloat32m8_t vSrc;
    vfloat32m1_t vSum = __riscv_vfmv_s_f_f32m1(0.0f, 1);
    while (n > 0) {
        vl = __riscv_vsetvl_e32m8(n);
        vSrc = __riscv_vle32_v_f32m8(pSrc, vl);
        vSum = __riscv_vfredusum_vs_f32m8_f32m1(vSrc, vSum, vl);
        pSrc += vl;
        n -= vl;
    }
    *pResult = __riscv_vfmv_f_s_f32m1_f32(vSum);
}

void scalar_sum(float *pSrc, uint32_t n, float *pResult)
{
    float sum = 0.0f;
    while (n > 0U) {
        sum += *pSrc++;
        n--;
    }
    *pResult = sum;
}

int main(int argc, char **argv)
{
    float input[] = {1.0f, 2.0f, 3.0f, 10.0f, 42.0f};
    size_t vlen = 5;
    float scalar_result, vector_result;
    vector_sum(input, vlen, &vector_result);
    scalar_sum(input, vlen, &scalar_result);
    printf("RVV: %f \nScalar: %f\n", vector_result, scalar_result);
    return 0;
}
``` 
```
$GEM5/build/RISCV/gem5.fast $GEM5/configs/example/gem5_library/riscv-se.py 

gem5 version DEVELOP-FOR-23.1
gem5 compiled Oct 10 2023 08:30:21
gem5 started Oct 12 2023 10:34:58

Beginning simulation!
src/arch/riscv/linux/se_workload.cc:60: warn: Unknown operating system; assuming Linux.
RVV: nan 
Scalar: 58.000000
Exiting @ tick 18640874 because exiting with last active thread context.

$ qemu-riscv64 -cpu rv64,v=true,vext_spec=v1.0,vlen=256,elen=64 gem5_example 
RVV: 58.000000 
Scalar: 58.000000
```

**Host Operating System**
Ubuntu 20.04

**Host ISA**
X86

**Compiler used**
gcc 9.4.0 to build gem5
Clang 16 for RVV code
```
clang -O3 -march=rv64gcv --target=riscv64-unknown-linux-musl -static -o gem5_example main.c
```

tommarin:
This PR improves the vector register groups overlap check in widening/narrowing 
instructions.

- Fix wrong illegal overlap condition between VS2 and VD vector register 
groups.
- Also check VS1 vector register group for overlap with VD in vector-vector 
instructions.
- Parametrize widening/narrowing factors in overlap check function to potentially 
handle more cases.

Fixes issue #442.

### DIFF
--- a/src/arch/riscv/isa/formats/vector_arith.isa
+++ b/src/arch/riscv/isa/formats/vector_arith.isa
@@ -553,7 +553,6 @@ def format VectorGatherFormat(code, category, *flags) {{
     else:
         error("not supported category for VectorIntFormat: %s" % category)
     src2_reg_id = "vecRegClass[_machInst.vs2 + vs2_idx]"
-    src2_reg_id = "vecRegClass[_machInst.vs2 + vs2_idx]"
 
     # vtmp0 as dummy src reg to create dependency with pin vd micro
     src3_reg_id = "vecRegClass[VecMemInternalReg0 + vd_idx]"
--- a/src/arch/riscv/isa/formats/vector_arith.isa
+++ b/src/arch/riscv/isa/formats/vector_arith.isa
@@ -79,38 +79,45 @@ let {{
             uint32_t ei = i + vtype_VLMAX(vtype, vlen, true) * this->microIdx;
             ''' + code
 
-    def wideningOpRegisterConstraintChecks(code):
-        return '''
-            const uint32_t num_microops = 1 << std::max<int64_t>(0, vtype_vlmul(machInst.vtype8) + 1);
+    def wideningOpRegisterConstraintChecks(code, src2_dw):
+        constraint_checks = '''
+            const uint32_t num_microops = 1 << std::max<int64_t>(0, vlmul + 1);
             if ((machInst.vd % alignToPowerOfTwo(num_microops)) != 0) {
                 std::string error =
                     csprintf("Unaligned Vd group in Widening op");
                 return std::make_shared<IllegalInstFault>(error, machInst);
             }
-            if ((machInst.vs2 <= machInst.vd) && (machInst.vd < (machInst.vs2 + num_microops - 1))) {
-                // A destination vector register group can overlap a source vector
-                // register group if The destination EEW is greater than the source
-                // EEW, the source EMUL is at least 1, and the overlap is in the
-                // highest- numbered part of the destination register group.
+            '''
+        if not src2_dw:
+            constraint_checks += '''
+            if (((vlmul < 0) && (VS2 == VD)) ||
+                ((vlmul >= 0) && (VS2 < VD + num_microops - (1 << vlmul)) &&
+                 (VD < VS2 + (1 << vlmul)))) {
+                // A destination vector register group can overlap a source
+                // vector register group if the destination EEW is greater than
+                // the source EEW, the source EMUL is at least 1, and the
+                // overlap is in the highest- numbered part of the destination
+                // register group.
                 std::string error =
                     csprintf("Unsupported overlap in Vs2 and Vd for Widening op");
                 return std::make_shared<IllegalInstFault>(error, machInst);
             }
-            ''' + code
+            '''
+        return constraint_checks + code
 
     def narrowingOpRegisterConstraintChecks(code):
         return '''
-            const uint32_t num_microops = 1 << std::max<int64_t>(0, vtype_vlmul(machInst.vtype8) + 1);
+            const uint32_t num_microops = 1 << std::max<int64_t>(0, vlmul + 1);
             if ((machInst.vs2 % alignToPowerOfTwo(num_microops)) != 0) {
                 std::string error =
                     csprintf("Unaligned VS2 group in Narrowing op");
                 return std::make_shared<IllegalInstFault>(error, machInst);
             }
-            if ((machInst.vs2 < machInst.vd) && (machInst.vd <= (VS2 + num_microops - 1))) {
-                // A destination vector register group can overlap a source vector
-                // register group The destination EEW is smaller than the source EEW
-                // and the overlap is in the lowest-numbered part of the source
-                // register group
+            if ((VS2 < VD) && (VD <= (VS2 + num_microops - 1))) {
+                // A destination vector register group can overlap a source
+                // vector register group if the destination EEW is smaller than
+                // the source EEW and the overlap is in the lowest-numbered
+                // part of the source register group
                 std::string error =
                     csprintf("Unsupported overlap in Vs2 and Vd for Narrowing op");
                 return std::make_shared<IllegalInstFault>(error, machInst);
@@ -329,10 +336,12 @@ def format VectorIntWideningFormat(code, category, *flags) {{
     else:
         error("not supported category for VectorIntFormat: %s" % category)
     src2_reg_id = ""
+    src2_dw = False
     if inst_suffix in ["vv", "vx"]:
         src2_reg_id = "vecRegClass[_machInst.vs2 + _microIdx / 2]"
     elif inst_suffix in ["wv", "wx"]:
         src2_reg_id = "vecRegClass[_machInst.vs2 + _microIdx]"
+        src2_dw = True
 
     set_dest_reg_idx = setDestWrapper(dest_reg_id)
 
@@ -355,7 +364,7 @@ def format VectorIntWideningFormat(code, category, *flags) {{
         code = eiDeclarePrefix(code, widening=True)
     code = loopWrapper(code)
 
-    code = wideningOpRegisterConstraintChecks(code)
+    code = wideningOpRegisterConstraintChecks(code, src2_dw)
 
     vm_decl_rd = ""
     if v0_required:
@@ -746,10 +755,12 @@ def format VectorFloatWideningFormat(code, category, *flags) {{
     else:
         error("not supported category for VectorFloatFormat: %s" % category)
     src2_reg_id = ""
+    src2_dw = False
     if inst_suffix in ["vv", "vf"]:
         src2_reg_id = "vecRegClass[_machInst.vs2 + _microIdx / 2]"
     elif inst_suffix in ["wv", "wf"]:
         src2_reg_id = "vecRegClass[_machInst.vs2 + _microIdx]"
+        src2_dw = True
 
     set_dest_reg_idx = setDestWrapper(dest_reg_id)
 
@@ -773,7 +784,7 @@ def format VectorFloatWideningFormat(code, category, *flags) {{
     code = loopWrapper(code)
     code = fflags_wrapper(code)
 
-    code = wideningOpRegisterConstraintChecks(code)
+    code = wideningOpRegisterConstraintChecks(code, src2_dw)
 
     vm_decl_rd = ""
     if v0_required:
--- a/src/arch/riscv/isa/formats/vector_arith.isa
+++ b/src/arch/riscv/isa/formats/vector_arith.isa
@@ -79,7 +79,28 @@ let {{
             uint32_t ei = i + vtype_VLMAX(vtype, vlen, true) * this->microIdx;
             ''' + code
 
-    def wideningOpRegisterConstraintChecks(code, src2_sew_mul, dest_sew_mul):
+    def wideningOpRegisterConstraintChecks(code, src2_sew_mul, dest_sew_mul,
+        src1_is_vec):
+        def checkOverlap(vreg_name, vreg_emul):
+            check_code = '''
+            if ((({vreg_emul} < 0) && ({vreg_name} == VD)) ||
+                (({vreg_emul} >= 0) &&
+                 ({vreg_name} < VD + num_microops - (1 << {vreg_emul})) &&
+                 (VD < {vreg_name} + (1 << {vreg_emul})))) {
+                // A destination vector register group can overlap a source
+                // vector register group if the destination EEW is greater than
+                // the source EEW, the source EMUL is at least 1, and the
+                // overlap is in the highest- numbered part of the destination
+                // register group.
+                std::string error =
+                    csprintf("Unsupported overlap in {vreg_name} and VD for "
+                             "Widening op");
+                return std::make_shared<IllegalInstFault>(error, machInst);
+            }
+            '''
+            check_code = check_code.replace("{vreg_name}", vreg_name)
+            check_code = check_code.replace("{vreg_emul}", vreg_emul)
+            return check_code
         src2_sew_mul_bits = src2_sew_mul.bit_length() - 1
         dest_sew_mul_bits = dest_sew_mul.bit_length() - 1
         constraint_checks = '''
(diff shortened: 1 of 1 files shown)


tommarin:
Considering the alignment to LMUL, the condition check in the case SEW->2\*SEW could be greatly simplified: the only illegal case is Vs2 == Vd. Still, it probably makes sense to keep it generic like in the original idea, to support for potential future cases (e.g. SEW->4\*SEW, as mentioned here: https://riscv.org/wp-content/uploads/2019/06/17.40-Vector_RISCV-20190611-Vectors.pdf). But in that case the calculation of num_microops will have to be made generic as well.

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -79,38 +79,45 @@ let {{
             uint32_t ei = i + vtype_VLMAX(vtype, vlen, true) * this->microIdx;
             ''' + code
 
-    def wideningOpRegisterConstraintChecks(code):
-        return '''
-            const uint32_t num_microops = 1 << std::max<int64_t>(0, vtype_vlmul(machInst.vtype8) + 1);
+    def wideningOpRegisterConstraintChecks(code, src2_dw):
+        constraint_checks = '''
+            const uint32_t num_microops = 1 << std::max<int64_t>(0, vlmul + 1);
```
hnpl: Is this specific to the case where `SEW` is widened to `2*SEW`?

hnpl (commented):
Thanks for the contribution. I believe the issue in #442 would be fixed by this; however, I still have a couple of concerns.

- If I understand correctly, the `constraint_checks` code checks if the end of VD group overlaps with the start of VS group and if the end of VS2 group overlaps with the start of VD group. This totally makes sense to me. However, I'm not sure if `VD + num_microops - (1 << vlmul))` indicates the end of the VD group. Does this refer to the ["highest-numbered part" in the spec](https://github.com/riscv/riscv-v-spec/blob/2f68ef7256d6ec53e4d2bd7cb12862f406d64e34/v-spec.adoc?plain=1#L997)?
- How do you test the widening instructions with overlapping vector registers?

tommarin:
> Thanks for the contribution. I believe the issue in #442 would be fixed by this; however, I still have a couple of concerns.
> 
>     * If I understand correctly, the `constraint_checks` code checks if the end of VD group overlaps with the start of VS group and if the end of VS2 group overlaps with the start of VD group. This totally makes sense to me. However, I'm not sure if `VD + num_microops - (1 << vlmul))` indicates the end of the VD group. Does this refer to the ["highest-numbered part" in the spec](https://github.com/riscv/riscv-v-spec/blob/2f68ef7256d6ec53e4d2bd7cb12862f406d64e34/v-spec.adoc?plain=1#L997)?
> 
>     * How do you test the widening instructions with overlapping vector registers?

Thank you for reviewing the PR!
* Yes, that is the logic behind that boundary. We have to leave a space for overlap in the highest part, which depends on the size of the source vector group.
* I have checked that the commit solves the issue in two distinct cases, one being the example mentioned in the issue #442. I have also checked a custom-made example where the overlap happens (vd == vs2): this is properly detected.

I will try to generalize this function a little bit, e.g. de-hardcoding the widening factors.

src/arch/riscv/isa/formats/vector_arith.isa
tommarin: Yes, indeed. As of now, no instruction widens to more than 2*SEW, but let me try to generalize it.

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -79,38 +79,45 @@ let {{
...
-            if ((machInst.vs2 <= machInst.vd) && (machInst.vd < (machInst.vs2 + num_microops - 1))) {
-                // A destination vector register group can overlap a source vector
-                // register group if The destination EEW is greater than the source
-                // EEW, the source EMUL is at least 1, and the overlap is in the
-                // highest- numbered part of the destination register group.
+            '''
+        if not src2_dw:
+            constraint_checks += '''
+            if (((vlmul < 0) && (VS2 == VD)) ||
+                ((vlmul >= 0) && (VS2 < VD + num_microops - (1 << vlmul)) &&
```
hnpl: Please correct me if I'm wrong. I think it should be an `or` rather than an `and` for the boundary check, i.e.,

```python
((vlmul >= 0) && 
    ((VS2 < VD + num_microops - (1 << vlmul)) || (VD < VS2 + (1 << vlmul))))
```

This is because even if one of the start and the end of the VD group "overly" overlaps with VS2, it violates the boundary constraint.

tommarin: To derive that condition, I started from the opposite one, which is where there is no (illegal) overlap.
- Any source vector higher than the top of the VD group (minus some overlap, sized as the VS2 group) is acceptable.
- Any source vector minor than the bottom of the VD group (minus the VS2 vector group size, to ensure no overlap at all) is acceptable.

If we negate this `or` conditions, we get an `and`: anything in the middle should be discarded.

A counterexample is, with LMUL=4, `vwmul.vv  v8, v0, v4`. `(VS2 < VD + 8 - 4)` is `true`, but this case is completely legal, as `(VD < VS2 + 4)` is `false`.

I tried to draw some of the cases, one real (LMUL=4, SEW->2\*SEW) and one hypothetical (LMUL=2, SEW->4\*SEW).
The red/orange area is the one where VS2 cannot sit because overlap is illegal. Orange area is also impossible because of alignment constraints.
In yellow I represented the area where both conditions are true in the original implementation, with boxes representing the register groups. In purple, the behavior of the new one (correct).

Please tell me if it makes sense. I will push a new commit with the generalization when ready.

image

hnpl: Thanks for the detailed explanation! I believe the condition you have is right.

I arrived at the same conclusion as yours when I think of this problem as the finding overlapping 1D segments problem. Basically we want VS2 to be either entirely on the left of VD (VS2.end < VD.start), or mostly to the right of VD (VS2.end > VD.start - num_microops - LMUL). The negation of this condition is exactly the same as yours.

saul44203: I believe this condition is right as well.  Apart from this, shouldn't we also check the vs1 vector register group for the vector-vector instructions (i.e. `.vv` and `.wv`)?

tommarin: You're right. I've added another commit to fix that too.

hnpl (approved):

ivanaamit (approved):

PR #1340, opened 2024-07-09
arch-riscv: fix initialization for some vector reduction insts

saul44203:
Vector reduce float (widening and non-widening) and integer (widening) instructions initialize the reduce loop operation with the first element of the destination register (i.e. `Vd[0]`). 

Since all reductions per spec seem to be `Vd[0] = Vs1[0] + Vs2[*]` (where `+` is an arbitrary binary op and `*` indicates all active elements) gem5 will calculate this incorrectly if `Vd[0]` and/or `Vs1[0]` are non-neutral for the operation (the later case being because it's not taken into account at all).

To solve this we just have to initialize the reduction loop to `Vs1[0]` (the non-widening integer reduction already does this).

Change-Id: I2da260e9ff15fe087d6981660ee202130d5ea15a

### DIFF
--- a/src/arch/riscv/isa/templates/vector_arith.isa
+++ b/src/arch/riscv/isa/templates/vector_arith.isa
@@ -1826,7 +1826,7 @@ Fault
 
     auto reduce_loop =
         [&, this](const auto& f, const auto* _, const auto* vs2) {
-            vu tmp_val = Vd[0];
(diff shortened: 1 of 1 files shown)


hnpl (approved):
Good catch!

BobbyRBruce (approved):

PR #1342, opened 2024-07-11
arch,cpu: Implement generic reset method for MMU

rogerchang23424:
Implementing generic reset method for MMU allows each ISA implementing their own reset methods. The default reset MMU method is flush all TLB entries. For example, The RISC-V needs to do PMP reset when received the reset signal, but the TLBs don't require to be flushed.

Change-Id: I158261570fb6e5216ec105fbdc53460f83f88d15

### DIFF
--- a/src/arch/generic/mmu.cc
+++ b/src/arch/generic/mmu.cc
@@ -94,6 +94,13 @@ BaseMMU::flushAll()
 }
 
 void
+BaseMMU::reset()
+{
+    // flush the TLBs by defaults
+    flushAll();
+}
+
+void
 BaseMMU::demapPage(Addr vaddr, uint64_t asn)
 {
     itb->demapPage(vaddr, asn);
--- a/src/arch/generic/mmu.hh
+++ b/src/arch/generic/mmu.hh
@@ -109,6 +109,8 @@ class BaseMMU : public SimObject
 
     virtual void flushAll();
 
+    virtual void reset();
+
     void demapPage(Addr vaddr, uint64_t asn);
 
     virtual Fault
--- a/src/cpu/base.cc
+++ b/src/cpu/base.cc
@@ -694,8 +694,8 @@ BaseCPU::setReset(bool state)
             tc->getIsaPtr()->resetThread();
             // reset the decoder in case it had partially decoded something,
             tc->getDecoderPtr()->reset();
-            // flush the TLBs,
-            tc->getMMUPtr()->flushAll();
+            // reset MMU,
+            tc->getMMUPtr()->reset();
             // Clear any interrupts,
             interrupts[tc->threadId()]->clearAll();
             // and finally reenable execution.


giactra (commented):
What is the difference with flushAll?
Could you document the method on how reset is supposed to be used?

rogerchang23424:
Sorry for replying the issue late

It's because I want to implement generic MMU reset methods. For example, do PMP reset in MMU reset method in RiscvISA.

Since the `flushAll` invalidate all of entry of ITLB and DTLB and no other side effect. I think we can add some extra MMU reset under `BaseMMU::flushAll`

No sure if my understand is correct. Feel free to leave your consideration

giactra:
> Sorry for replying the issue late
> 
> It's because I want to implement generic MMU reset methods. For example, do PMP reset in MMU reset method in RiscvISA.
> 
> Since the `flushAll` invalidate all of entry of ITLB and DTLB and no other side effect. I think we can add some extra MMU reset under `BaseMMU::flushAll`
> 
> No sure if my understand is correct. Feel free to leave your consideration

I see; I checked the existing code for PMP reset:

https://github.com/gem5/gem5/blob/stable/src/arch/riscv/faults.cc#L197

Is it necessary to add a virtual method to the generic MMU if it is for a single ISA? It probably isn't unless you need to add a reset call in the shared uarch code.
Please let me know what's the long term plan for this.

rogerchang23424:
I think we don't need to add the extra MMU reset for specific arch currently. I will close the PR.

rogerchang23424:
I have discuss the issue with my fork. We need to implement a generic MMU `reset`. They handles MMU is different way with `flushAll`

- `BaseMMU::flushAll`: Invalidate/Flush all of instruction and data TLB entries. They can be called by specific instruction. For example: In RISC-V, `flushAll` methods will be called by `sfence.vma`. However, they should not clean the PMPCfg register.
- `BaseMMU::reset`: Do reset when CPU received the reset signal. For example, In RISC-V privilege spec 3.4, clean the A and L bit in PMP register, but the spec doesn't require to clean all of the TLB entries.

Reset in Spec:
```
Upon reset, a hart’s privilege mode is set to M. The mstatus fields MIE and MPRV are reset to 0. If little-endian memory accesses are supported, the mstatus/mstatush field MBE is reset to 0. The misa register is reset to enable the maximal set of supported extensions, as described in [Section 3.1.1](https://gchips-sw-tools.teams.x20web.corp.google.com/sim-tools/RISC-V%20spec/riscv-isa-release-87edab7-2024-05-04/priv-isa-asciidoc.html#misa). For implementations with the "A" standard extension, there is no valid load reservation. The pc is set to an implementation-defined reset vector. The mcause register is set to a value indicating the cause of the reset. Writable PMP registers’ A and L fields are set to 0, unless the platform mandates a different reset value for some PMP registers’ A and L fields. If the hypervisor extension is implemented, the hgatp.MODE and vsatp.MODE fields are reset to 0. If the Smrnmi extension is implemented, the mnstatus.NMIE field is reset to 0. No WARL field contains an illegal value. All other hart state is UNSPECIFIED.
```

giactra (approved):

PR #1344, opened 2024-07-11
stdlib: Adding Modification support in stdlib.

mahyarsamani:
This implements a capability in AbstractBoard to allow running
    user defined code between calling _pre_instantiate and
    m5.instantiate while using the Simulator pacakge. The simulator
    object will call a secondary function from the board called
    `_apply_modifiers` to allow user provided function to be
    executed after everything is connected.

    This change adds the following:
    - A step in the instantiation process in the Simulator class that
    calls the `_apply_modifiers` method from the board passed to the
    Simulator object.
    - `_apply_modifier` method to the AbstractBoard that does not
    really require further extension that calls on the modifiers
    attached to the board to apply their modification.
    - `_modifiers` attribute to the AbstractBoard to store all
    the modifiers attached to the board.
    - `add_modifier` method to add a modifier object to the
    `_modifiers` attribute.
    - Example modifier for modifying the branch predictor in an
    O3CPU core.

    This is useful since it makes fast and reusable changes to the
    parameters of a component in the standard library. An important
    use case for this would be to explore the parameter space of
    different SimObjects in gem5 for building reliable prebuilts.

### DIFF
--- a/src/python/SConscript
+++ b/src/python/SConscript
@@ -37,6 +37,7 @@ PySource("gem5.simulate", "gem5/simulate/simulator.py")
 PySource("gem5.simulate", "gem5/simulate/exit_event.py")
 PySource("gem5.simulate", "gem5/simulate/exit_event_generators.py")
 PySource("gem5.components", "gem5/components/__init__.py")
+PySource("gem5.components", "gem5/components/modifier.py")
 PySource("gem5.components.boards", "gem5/components/boards/__init__.py")
 PySource("gem5.components.boards", "gem5/components/boards/abstract_board.py")
 PySource(
@@ -473,6 +474,22 @@ PySource(
 PySource(
     "gem5.components.processors",
     "gem5/components/processors/traffic_generator.py",
+)
+PySource(
+    "gem5.components.processors.modifiers",
+    "gem5/components/processors/modifiers/__init__.py",
+)
+PySource(
+    "gem5.components.processors.modifiers.o3_modifiers",
+    "gem5/components/processors/modifiers/o3_modifiers/__init__.py",
+)
+PySource(
+    "gem5.components.processors.modifiers.o3_modifiers",
+    "gem5/components/processors/modifiers/o3_modifiers/o3_cpu_modifier.py",
+)
+PySource(
+    "gem5.components.processors.modifiers.o3_modifiers",
+    "gem5/components/processors/modifiers/o3_modifiers/branchpred_modifier.py",
 )
 PySource("gem5.prebuilt", "gem5/prebuilt/__init__.py")
 PySource("gem5.prebuilt.demo", "gem5/prebuilt/demo/__init__.py")
--- a/src/python/SConscript
+++ b/src/python/SConscript
@@ -26,362 +26,574 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-Import('*')
+Import("*")
 
-PySource('gem5', 'gem5/__init__.py')
-PySource('gem5', 'gem5/coherence_protocol.py')
-PySource('gem5', 'gem5/isas.py')
-PySource('gem5', 'gem5/runtime.py')
-PySource('gem5.simulate', 'gem5/simulate/__init__.py')
-PySource('gem5.simulate', 'gem5/simulate/simulator.py')
-PySource('gem5.simulate', 'gem5/simulate/exit_event.py')
-PySource('gem5.simulate', 'gem5/simulate/exit_event_generators.py')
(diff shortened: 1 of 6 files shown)


mahyarsamani:
@powerjg @BobbyRBruce

src/python/SConscript
```diff
@@ -15,7 +15,7 @@
 # this software without specific prior written permission.
 #
 # THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
-# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# 'AS IS' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
```
powerjg: lol, this line shouldn't change.

src/python/SConscript
powerjg, line 1: Can you either revert the style changes in this file or create two commits: One that runs black on the file and one that makes your change?

src/python/gem5/components/modifier.py
```diff
@@ -0,0 +1,71 @@
...
+    board. It will be useful when searching parameter space for SimObjects used
+    by components in the standard library without having to go through that
+    exposes that parameter to the user through the constructor.
+
+    Each modifier needs to implement the following methods:
+    - _get_simobjects_from_board: This method should return a list of SimObjects
+    (not components, e.g. O3CPU and not SimpleProcessor) that it will modify.
+    - _do_modification: This method should actually do the modification it is
+    supposed to do. E.g. it changes the BranchPredictor of a cpu core.
+    """
```
powerjg, line 51: Can you show an example use case here?

powerjg (commented):
A couple of small things below.

You should also add some tests for this.

mahyarsamani:
Comments are addressed.

BobbyRBruce:
> Comments are addressed.

You still haven't added tests for this.

mahyarsamani:
> > Comments are addressed.
> 
> You still haven't added tests for this.

How do you envision the tests for this? Should they test that they don't cause errors or actually test that they accomplish the modification?

BobbyRBruce:
> > > Comments are addressed.
> > 
> > 
> > You still haven't added tests for this.
> 
> How do you envision the tests for this? Should they test that they don't cause errors or actually test that they accomplish the modification?

Both these sound good. I'd do both if possible

BobbyRBruce:
I don't mean this to sound snide. I'm genuinely trying to understand this, but could you maybe explain a use-case where this feature is necessary? The thing that I'm stuck on is why do we want to modify things in this way and require doing so at pre-instantiate? If we want to modify any SimObject, shouldn't we just expose them and modify them before calling `Simulator.run`?

There's probably some very silly thing I'm not thinking of, but I could use a little help in understanding this.

hnpl:
Correct me if I'm wrong. This change is meant for common changes to the board without having to inherit too many classes to override the _pre_instantiate function. I have an example of using arm SVE and RISC-V vector extension, where the vector length is set by modifying ISA object or system object, which is quite tedious to do it for every gem5 config. [Example](https://github.com/hnpl/gem5_components/blob/62f5715341c956e69e9dd72fd436c4cc1c936ed5/vector_cores/vector_cores.py#L33).

Having said that, I don't have a strong opinion on this. Having modifiers is useful if you have to apply the same modifications multiple times. However, if we have to repeatedly apply the same modifications to the objects provided by stdlib, it probably signals that something is missing from the stdlib. Also, ideally, if we modify any stdlib object, the object should be inherited to signal that it was not the same object as defined in the stdlib. Having modifiers allows you to bypass this.

mahyarsamani:
The main objective this change is hoping to accomplish is allow for running code (modifiers) after things are "connected" and before they are instantiated. I think Hoa's interpretation is correct. However, there is value in making modifications structured. My use case is that I want to tweak some SimObject parameters for making reliable stdlib components. It becomes very tiresome to try to expose every param that I want to expose through the constructor (specially since they might be buried deep in the hierarchical nature of stdlib). In addition to the ease, it helps me track the different versions of my components that I had for later reproduction and data gathering. Imagine a scenario when you have to go back and fourth between different values for a certain parameter, in this scenario storing the modifiers is a more robust choice that "remembering" the values. I have examples in this repo [here](https://github.com/darchr/auto-eval-and-accuracy/blob/main/scripts/mods/noc_mods.py) (is private).

BobbyRBruce:
I create this PR to fixup the pre_instantiation code and see if this will work for you: https://github.com/gem5/gem5/pull/1497

As part of this PR I created this commit as a demo of how someone might add this branch_prediction setting "modification" using the pre-instantiation functions: https://github.com/gem5/gem5/pull/1497/commits/4c608292407fa6ef6dbb0d4fdbf7afbead2eb420. Please let me know if this is suitable. I'm not aware of all the requirements.

BobbyRBruce:
Closing. This is no longer require due to #1497

PR #1350, opened 2024-07-14
arch-vega: Pass s_memtime through smem pipe

mjkpolo:
The Vega ISA's s_memtime instruction is used to obtain a cycle value from the GPU.  Previously, this was implemented to obtain the cycle count when the memtime instruction reached the execute stage of the GPU pipeline.  However, from microbenchmarking we have found that this under reports the latency for memtime instructions relative to real hardware.  Thus, we changed its behavior to go through the scalar memory pipeline and obtain a latency value from the the SQC (L1 I$).  This mirrors the suggestion of the AMD Vega ISA manual that s_memtime should be treated like a s_load_dwordx2.

The default latency was set based on microbenchmarking.

Change-Id: I5e251dde28c06fe1c492aea4abf9f34f05784420


src/arch/amdgpu/common/tlb.cc
```diff
@@ -379,7 +379,11 @@ namespace X86ISA
 
         assert(seg != segment_idx::Ms);
         Addr vaddr = req->getVaddr();
-        DPRINTF(GPUTLB, "TLB Lookup for vaddr %#x.\n", vaddr);
+        if (req->hasNoAddr()) {
```
mattsinc, line 382: Probably the commit message needs to discuss the generic "has no addr" support since that is one of the key pieces.

src/mem/request.hh
```diff
@@ -255,7 +255,8 @@ class Request : public Extensible<Request>
          * These flags are *not* cleared when a Request object is
          * reused (assigned a new address).
          */
-        STICKY_FLAGS = INST_FETCH
+        STICKY_FLAGS = INST_FETCH,
```
mattsinc, line 258: how did we come up with this address?

src/mem/ruby/system/VIPERSequencer.cc
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright (c) 2013-2015 Advanced Micro Devices, Inc.
```
mattsinc: Copyright should reflect UW adding this file, not AMD, since you are the author ... I guess you can include me if you want to, but definitely you and UW

src/mem/ruby/system/VIPERSequencer.hh
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2013-2015 Advanced Micro Devices, Inc.
```
mattsinc: ditto

src/mem/ruby/system/VIPERSequencer.py
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) 2015 Advanced Micro Devices, Inc.
```
mattsinc: ditto

mattsinc (changes requested):

src/mem/request.hh
mjkpolo: `TLBI_EXT_SYNC_COMP` seems to be the largest value of FlagsType, so `HAS_NO_ADDR`'s value is that << 1

src/gpu-compute/GPUStaticInstFlags.py
```diff
@@ -108,4 +108,5 @@ class GPUStaticInstFlags(Enum):
         "MAC",  # MAC
         "MAD",  # MAD
         "MFMA",  # MFMA
+        "HAS_NO_ADDR",  # Request has no address but goes to SQC for timing
```
abmerop: Nit: This doesn't match TheStyle of the names above

abmerop (commented):
One minor comment otherwise looks fine to me.

src/gpu-compute/GPUStaticInstFlags.py
mjkpolo: I had a problem where if it was `hasNoAddr` vs `HAS_NO_ADDR`, the compiler thought I was referring to the function `hasNoAddr`. I can try using the `::` scope to refer to the flag if I spell it camel case like you're suggesting and see if that works.

mattsinc: I think @abmerop is suggesting HASNOADDR to be consistent with the other ones.  Are you saying that didn't work @mjkpolo ?

src/arch/amdgpu/common/tlb.cc
mattsinc: I have updated it

src/mem/request.hh
mattsinc: Please add a comment explaining this

src/mem/ruby/system/VIPERSequencer.cc
```diff
@@ -0,0 +1,109 @@
+/*
+ * Copyright (c) 2024 UW Madison
+ *
```
mattsinc, line 3: These should be "Copyright (c) 2024 The University of Wisconsin"

mattsinc (changes requested):

src/gpu-compute/GPUStaticInstFlags.py
mjkpolo: oh sure I can do that instead, I think either `HasNoAddr` or `HASNOADDR` would work. @abmerop where you implying `HasNoAddr` by

> This doesn't match **TheStyle**

or `HASNOADDR`

abmerop: `NoAddr`

src/mem/ruby/system/VIPERSequencer.cc
mattsinc: @mjkpolo just making sure you saw this

mjkpolo: ope I forgot about this let me make that change quick

mjkpolo:
These changes behave differently in GPU FS mode compared to SE mode. Initially we thought this was ok, due to `mi200.py` being different, and not because of FS mode. After testing on real mi200 (gfx90a) hardware, I've determined that `mi200.py` should behave the same as `apu_se.py`. I think we should freeze this pull request until this is resolved.

mattsinc:
> These changes behave differently in GPU FS mode compared to SE mode. Initially we thought this was ok, due to `mi200.py` being different, and not because of FS mode. After testing on real mi200 (gfx90a) hardware, I've determined that `mi200.py` should behave the same as `apu_se.py`. I think we should freeze this pull request until this is resolved.

I'm not sure why you expect apu_se.py (which models a Vega10) and mi200.py (which models an MI200) to behave the same?  I guess I would expect the two of them to have different latencies even in a situation where both MI200 and Vega10 were configured to run in apu_se.py.  If it's as simple as the latencies for the two configurations are different, then probably it's a sign we just need to tune the configurations differently?

@abmerop thoughts?

mattsinc (approved):
From a code perspective I believe this has resolved all the issues @abmerop and I highlighted, so I'm approving it so the tests can be run.

Let me know @abmerop , @mjkpolo about the configuration issue though.  If it's as simple as a Vega10 and MI200 need different scalar cache latencies, I am inclined to not block this patch.

abmerop:
There *is* still a vega10 config for FS mode, I have not removed it yet. Note that you need to use the MMIO trace for this one, but you can try with that.

It sounds like the mi200 config would need different latency values?

ivanaamit:
@mattsinc, @abmerop, @mjkpolo Just checking if this PR is ready to be merged. Thanks.

mattsinc:
> @mattsinc, @abmerop, @mjkpolo Just checking if this PR is ready to be merged. Thanks.

@mjkpolo should chime in, but I believe the answer is "yes", this can be merged.  @mjkpolo should confirm though.

mjkpolo:
As long as we are ok with the timing being correct for apu_se.py, and timing for FS mode slightly more correct now, eventually correct once changes from Nagendra and others in your group are merged, then it should be good to merge

mattsinc:
> As long as we are ok with the timing being correct for apu_se.py, and timing for FS mode slightly more correct now, eventually correct once changes from Nagendra and others in your group are merged, then it should be good to merge

@mjkpolo : as I understand it, the issue now is about correlation with real hardware in FS vs. SE mode -- which is technically separate from this patch, which is focused on adding support for the feature itself.  Assuming you agree with this statement, then I think we could include updates for GPUFS correlation as part of a separate, future patch by @v-ramadas or someone else.  Seem reasonable?  Am I missing something?

src/arch/amdgpu/vega/insts/smem.cc
```diff
@@ -950,10 +951,26 @@ namespace VegaISA
     void
     Inst_SMEM__S_MEMTIME::execute(GPUDynInstPtr gpuDynInst)
     {
+        Wavefront *wf = gpuDynInst->wavefront();
+        gpuDynInst->execUnitId = wf->execUnitId;
+        gpuDynInst->latency.init(gpuDynInst->computeUnit());
+        gpuDynInst->latency.set(gpuDynInst->computeUnit()->clockPeriod()*41);
```
mattsinc: Please add an input arg to make this configurable, instead of using 41 as a magic number.

mattsinc (changes requested):

mattsinc (approved):
Once the tests have passed again, this should be good to go, nice job @mjkpolo !

mattsinc:
@abmerop let us know if you want to review one more time or if I should merge

src/arch/amdgpu/vega/tlb.cc
```diff
@@ -342,20 +345,27 @@ GpuTLB::issueTLBLookup(PacketPtr pkt)
...
+            // host memory for a memtime request
+            pkt->req->setSystemReq(false);
+        } else {
+            VegaTlbEntry *entry = lookup(virt_page_addr, false);
+            assert(entry);
 
-        // Set if this is a system request
-        pkt->req->setSystemReq(entry->pte.s);
+            // Set if this is a system request
+            pkt->req->setSystemReq(entry->pte.s);
```
abmerop, line 362: nit: some tabs here

abmerop (approved):
Looks fine, minor nitpick

src/arch/amdgpu/vega/tlb.cc
abmerop: Whoops- misunderstood the highlighting

Issue #1355, opened 2024-07-17
arch-riscv: Change wfi behaviour


h0lyalg0rithm:
**Describe the bug**
WFI instruction implementation in gem5 does not match the specification.

**Affects version**
Commit ID in develop - 9b8c84cb5d2f71b5af4329f9ff6cdad380a6225c

**Expected behavior**
According to the spec https://github.com/riscv/riscv-isa-manual/releases/tag/20240411 on page 51 and 52 it mentions the following

> The WFI instruction can also be executed when interrupts are disabled. The operation of WFI must be unaffected by the global interrupt bits in mstatus (MIE and SIE) and the delegation register mideleg (i.e., the hart must resume if a locally enabled interrupt becomes pending, even if it has been delegated to a less-privileged mode), but should honor the individual interrupt enables (e.g, MTIE) (i.e.,
implementations should avoid resuming the hart if the interrupt is pending but not individually enabled). WFI is also required to resume execution for locally enabled interrupts pending at any privilege level, regardless of the global interrupt enable at each privilege level.

**Current Implementation**
In the code here https://github.com/gem5/gem5/blob/9b8c84cb5d2f71b5af4329f9ff6cdad380a6225c/src/arch/riscv/isa/decoder.isa#L4883
I checks for pending interrupt  and masked interrupt but we dont check if the interrupt is enabled.This results in the hart being resumed instead.

4883   │                             if (ic->readIP() == 0
4884   │                                 && xc->readMiscReg(MISCREG_NMIP) == 0) {
4885   │                                 tc->quiesce();
4886   │                             }

**Suggested fix**
Add new method to the https://github.com/gem5/gem5/blob/9b8c84cb5d2f71b5af4329f9ff6cdad380a6225c/src/arch/riscv/interrupts.hh 
```
bool checkInterruptsWfi() const {
     return checkNonMaskableInterrupt() || (ip & ie).any());
}
```
The method is similar to the existing `checkInterrupts` method but it doesnt take the globalMask() into account, as mentioned in the spec.

If someone who has more knowledge about the spec could validate if this approach is correct.

**Host Operating System**
Ubuntu 22.04, Mac OS X, etc.

**Host ISA**
RISC-V

cc @adriaarmejach

robhau:
Hi @h0lyalg0rithm,

> The WFI instruction can also be executed when interrupts are disabled.

I think that means that we can execute `wfi` if interrupts are enabled **or** disabled.

[...]

PR #1356, opened 2024-07-17
arch-riscv: Change wfi behaviour

h0lyalg0rithm:
Previously we were only checking if there are any pending interrupts regardless of whether its enabled or not.This resulted in the simulator spending lots of cpu cycles simulating the wfi loop which results in poor performance of the simulator

Related to issue #1355 
Change-Id: Ifb6214509e65de7f6bb7db91af30c43bd582c873


h0lyalg0rithm:
Duplicate #1364

PR #1364, opened 2024-07-19
arch-riscv: Extend wfi behavior

### LINKED ISSUE #1355: arch-riscv: Change wfi behaviour

robhau:
At the moment, a hart does not halt if there are pending interrupts. However, an implementation can also consider the enable status of the individual interrupts, i.e., a halted hart would only resume if there are locally enabled pending interrupts. This commit introduces this behavior. The wfi behavior is controlled by the new configuration variable wfi_pending_resume of RiscvISA.

Change-Id: I316239f9732c6e73e6ad692491bca08d773dd995

### DIFF
--- a/src/arch/riscv/RiscvISA.py
+++ b/src/arch/riscv/RiscvISA.py
@@ -116,7 +116,7 @@ class RiscvISA(BaseISA):
     enable_Zicboz_fs = Param.Bool(True, "Enable Zicboz extension in FS mode")
 
     wfi_resume_on_pending = Param.Bool(
-        True,
+        False,
         "If wfi_resume_on_pending is set to True, the hart will resume "
         "execution when interrupt becomes pending. The local enabled status "
         "is not considered.\n"
--- a/src/arch/riscv/RiscvISA.py
+++ b/src/arch/riscv/RiscvISA.py
@@ -14,6 +14,7 @@
 # Copyright (c) 2016 RISC-V Foundation
 # Copyright (c) 2016 The University of Virginia
 # Copyright (c) 2023 The Regents of the University of California
+# Copyright (c) 2024 University of Rostock
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
@@ -113,6 +114,12 @@ class RiscvISA(BaseISA):
 
     enable_Zicbom_fs = Param.Bool(True, "Enable Zicbom extension in FS mode")
     enable_Zicboz_fs = Param.Bool(True, "Enable Zicboz extension in FS mode")
+
+    wfi_pending_resume = Param.Bool(
+        True,
+        "Set true to specify that only a pending interrupt is required "
+        "to resume after wfi instruction",
+    )
 
     def get_isa_string(self):
         isa_extensions = []
--- a/src/arch/riscv/RiscvISA.py
+++ b/src/arch/riscv/RiscvISA.py
@@ -115,10 +115,14 @@ class RiscvISA(BaseISA):
     enable_Zicbom_fs = Param.Bool(True, "Enable Zicbom extension in FS mode")
     enable_Zicboz_fs = Param.Bool(True, "Enable Zicboz extension in FS mode")
 
-    wfi_pending_resume = Param.Bool(
+    wfi_resume_on_pending = Param.Bool(
         True,
-        "Set true to specify that only a pending interrupt is required "
-        "to resume after wfi instruction",
+        "If wfi_resume_on_pending is set to True, the hart will resume "
+        "execution when interrupt becomes pending. The local enabled status "
+        "is not considered.\n"
+        "If wfi_resume_on_pending is set to False, the hart will only "
+        "resume the execution when an locally enabled interrupt becomes "
+        "pending.",
     )
 
     def get_isa_string(self):
--- a/src/arch/riscv/isa.cc
+++ b/src/arch/riscv/isa.cc
@@ -260,7 +260,7 @@ RegClass ccRegClass(CCRegClass, CCRegClassName, 0, debug::IntRegs);
 ISA::ISA(const Params &p) : BaseISA(p, "riscv"),
     _rvType(p.riscv_type), enableRvv(p.enable_rvv), vlen(p.vlen), elen(p.elen),
     _privilegeModeSet(p.privilege_mode_set),
-    _wfiPendingResume(p.wfi_pending_resume)
+    _wfiResumeOnPending(p.wfi_resume_on_pending)
 {
     _regClasses.push_back(&intRegClass);
     _regClasses.push_back(&floatRegClass);
--- a/src/arch/riscv/isa.cc
+++ b/src/arch/riscv/isa.cc
@@ -3,6 +3,7 @@
  * Copyright (c) 2016 The University of Virginia
  * Copyright (c) 2020 Barkhausen Institut
  * Copyright (c) 2022 Google LLC
+ * Copyright (c) 2024 University of Rostock
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -258,7 +259,8 @@ RegClass ccRegClass(CCRegClass, CCRegClassName, 0, debug::IntRegs);
 
 ISA::ISA(const Params &p) : BaseISA(p, "riscv"),
     _rvType(p.riscv_type), enableRvv(p.enable_rvv), vlen(p.vlen), elen(p.elen),
-    _privilegeModeSet(p.privilege_mode_set)
+    _privilegeModeSet(p.privilege_mode_set),
+    _wfiPendingResume(p.wfi_pending_resume)
 {
     _regClasses.push_back(&intRegClass);
     _regClasses.push_back(&floatRegClass);
--- a/src/arch/riscv/isa.hh
+++ b/src/arch/riscv/isa.hh
@@ -103,10 +103,10 @@ class ISA : public BaseISA
      * The WFI instruction can halt the execution of a hart.
      * If this variable is set true, the execution resumes if
      * an interrupt becomes pending. If this variable is set
-     * to false, the execution only resumes if an local enabled
+     * to false, the execution only resumes if an locally enabled
      * interrupt becomes pending.
     */
-    const bool _wfiPendingResume;
+    const bool _wfiResumeOnPending;
 
   public:
     using Params = RiscvISAParams;
@@ -180,7 +180,7 @@ class ISA : public BaseISA
 
     PrivilegeModeSet getPrivilegeModeSet() { return _privilegeModeSet; }
 
-    bool getWfiPendingResume() { return _wfiPendingResume; }
+    bool resumeOnPending() { return _wfiResumeOnPending; }
 
     virtual Addr getFaultHandlerAddr(
         RegIndex idx, uint64_t cause, bool intr) const;
--- a/src/arch/riscv/isa.hh
+++ b/src/arch/riscv/isa.hh
@@ -5,6 +5,7 @@
  * Copyright (c) 2016 RISC-V Foundation
  * Copyright (c) 2016 The University of Virginia
  * Copyright (c) 2020 Barkhausen Institut
+ * Coypright (c) 2024 University of Rostock
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -98,6 +99,15 @@ class ISA : public BaseISA
      */
     PrivilegeModeSet _privilegeModeSet;
 
+    /**
+     * The WFI instruction can halt the execution of a hart.
+     * If this variable is set true, the execution resumes if
+     * an interrupt becomes pending. If this variable is set
+     * to false, the execution only resumes if an local enabled
+     * interrupt becomes pending.
+    */
+    const bool _wfiPendingResume;
+
   public:
     using Params = RiscvISAParams;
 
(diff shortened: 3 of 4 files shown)


robhau:
Related to https://github.com/gem5/gem5/issues/1355 and https://github.com/gem5/gem5/pull/1356.

src/arch/riscv/RiscvISA.py
```diff
@@ -114,6 +115,12 @@ class RiscvISA(BaseISA):
     enable_Zicbom_fs = Param.Bool(True, "Enable Zicbom extension in FS mode")
     enable_Zicboz_fs = Param.Bool(True, "Enable Zicboz extension in FS mode")
 
+    wfi_pending_resume = Param.Bool(
+        True,
+        "Set true to specify that only a pending interrupt is required "
+        "to resume after wfi instruction",
```
powerjg: Apologies, I'm not super familiar with the RISC-V spec for interrupts... Could you expand on this documentation? Specifically, if you set this to `False` what is the behavior? It's non-obvious, at least to me.

powerjg (commented):
lgtm except that a little more information in the help string would be nice.

src/arch/riscv/RiscvISA.py
robhau: Hi @powerjg,

please tell me if the following documentation is sufficient. Then, I will add them as comments.

From the privileged RISC-V Spec, p. 51 - 52.

> If an enabled interrupt is present or later becomes present while the hart is stalled, the interrupt trap
> will be taken on the following instruction, i.e., execution resumes in the trap handler and mepc = pc +
> 4.

An enabled pending interrupt will resume the execution of a hart.

> The WFI instruction can also be executed when interrupts are disabled. The operation of WFI must be
> unaffected by the global interrupt bits in mstatus (MIE and SIE) and the delegation register mideleg
> (i.e., the hart must resume if a locally enabled interrupt becomes pending, even if it has been delegated
> to a less-privileged mode), but should honor the individual interrupt enables (e.g, MTIE) (i.e.,
> implementations should avoid resuming the hart if the interrupt is pending but not individually
> enabled). WFI is also required to resume execution for locally enabled interrupts pending at any
> privilege level, regardless of the global interrupt enable at each privilege level.
> If the event that causes the hart to resume execution does not cause an interrupt to be taken, execution
> will resume at pc + 4, and software must determine what action to take, including looping back to
> repeat the WFI if there was no actionable event.

A Hart can also resume if an interrupt is pending that is not locally enabled. If tasks do not need to be handled
in ISRs, the context switch is avoided.

*I am no native english speaker, but because of* can *and* should *I interpret it as optional. Therefore, the behavior is implementation specific.*

If `wfi_pending_resume` is set to True, the execution will resume when an interrupt becomes pending without considering the local enabled status.

If `wfi_pending_resume` is set to False, a pending interrupt will not be sufficient to resume execution. It has to be locally enabled too.

@h0lyalg0rithm Could you please check whether the documentation is correct?

powerjg: > I am no native english speaker, but because of can and should I interpret it as optional. Therefore, the behavior is implementation specific.

I *am* a native english speaking, but I have little confidence I understand the meaning of the spec here! The "can" and "should" are quite confusing in this context.

That said, your explanation at the end (presumably what you propose to add to the help string for the parameter) makes sense to me. Whether that's the intention of the spec... I'll let someone else chime in 🙂.

I vote for setting the default to whatever is faster when booting Linux (`False`?)

h0lyalg0rithm: Not a native english speaker, but a suggestion on the parameter
If `wfi_pending_only` is set to True, the hart will resume execution only when it receives a local pending interrupt
If `wfi_pending_only` is set to False, the hart will resume execution on any local pending interrupt.

robhau: Thank you @powerjg and @h0lyalg0rithm for your feedback. I extended the documentation in the .py file and also slightly renamed the member variable in the C++ source files as well as in the .py file.

I tried to run the riscvmatched-fs.py configuration script for `wfi_resume_on_pending` set to True and False to see changes in the simulation time. The goal was to see what configuration required less simulation time. However, at some point, I receive an "Aborted (core dumped)" in both simulations.

> ...
> src/arch/riscv/isa.cc:954: warn: 1637075195385: context 1: 40000 consecutive SC failures.
> src/arch/riscv/isa.cc:954: warn: 1755334591864: context 3: 30000 consecutive SC failures.
> src/arch/riscv/isa.cc:954: warn: 1815331475174: context 1: 50000 consecutive SC failures.
> src/cpu/minor/lsq.cc:1604: panic: Do not expect cross-cache-line atomic memory request
> Memory Usage: 18078880 KBytes
> Program aborted at tick 1860298736905
> --- BEGIN LIBC BACKTRACE ---
> build/RISCV/gem5.fast(_ZN4gem515print_backtraceEv+0x30)[0x556488aea470]
> ...

I will test the configuration script on the development branch were i started the branch for the pull request.

robhau: I tested the configuration script on the development branch. It seems that there is an bug, but independent from this pull request

Related issue: https://github.com/gem5/gem5/issues/1393

I will try to run full system simulations with other configuration scripts this week.

robhau: I have run `configs/example/gem5_library/riscv-ubuntu-run.py` over night. I slightly modified the configuration script to use the multisim framework.


```python
# Copyright (c) 2021 The Regents of the University of California
# All rights reserved.
#
# Redistribution and use in source and binary forms, with or without
# modification, are permitted provided that the following conditions are
# met: redistributions of source code must retain the above copyright
# notice, this list of conditions and the following disclaimer;
# redistributions in binary form must reproduce the above copyright
# notice, this list of conditions and the following disclaimer in the
# documentation and/or other materials provided with the distribution;
# neither the name of the copyright holders nor the names of its
# contributors may be used to endorse or promote products derived from
#
# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT

"""
This script shows an example of running a full system RISCV Ubuntu boot
simulation using the gem5 library. This simulation boots Ubuntu 20.04 using
2 TIMING CPU cores. The simulation ends when the startup is completed
successfully.

Usage
-----

``
scons build/RISCV/gem5.opt
./build/RISCV/gem5.opt \
    configs/example/gem5_library/riscv-ubuntu-run.py
``
"""

import m5
from m5.objects import Root

from gem5.components.memory import DualChannelDDR4_2400
from gem5.isas import ISA
import gem5.utils.multisim as multisim

for wfi in [True, False]:
    # This runs a check to ensure the gem5 binary is compiled for RISCV.

    # With RISCV, we use simple caches.
    from gem5.components.cachehierarchies.classic.private_l1_private_l2_walk_cache_hierarchy import (
        PrivateL1PrivateL2WalkCacheHierarchy,
    )

    # Here we setup the parameters of the l1 and l2 caches.
    cache_hierarchy = PrivateL1PrivateL2WalkCacheHierarchy(
        l1d_size="16kB", l1i_size="16kB", l2_size="256kB"
    )

    # Memory: Dual Channel DDR4 2400 DRAM device.

    memory = DualChannelDDR4_2400(size="3GB")

    # Here we setup the processor. We use a simple processor.
    processor = SimpleProcessor(
        cpu_type=CPUTypes.TIMING, isa=ISA.RISCV, num_cores=2
    )

    # Here we setup the board. The RiscvBoard allows for Full-System RISCV
    # simulations.
    board = RiscvBoard(
        clk_freq="3GHz",
        processor=processor,
        memory=memory,
    )

    cores = processor.get_cores()
    for core in cores:
        for isa in core.core.isa:
            isa.wfi_resume_on_pending = wfi

    # Here we a full system workload: "riscv-ubuntu-20.04-boot" which boots
    # Ubuntu 20.04. Once the system successfully boots it encounters an `m5_exit`
    # instruction which stops the simulation. When the simulation has ended you may
    # inspect `m5out/system.pc.com_1.device` to see the stdout.
    board.set_workload(
        obtain_resource("riscv-ubuntu-20.04-boot", resource_version="3.0.0")
    )

    simulator = Simulator(board=board, id=f"wfi_{wfi}")
    multisim.add_simulator(simulator)
#simulator.run()
```

Simulation performance for True:
```txt
simSeconds                                   6.981280                       # Number of seconds simulated (Second)
simTicks                                 6981279948783                       # Number of ticks simulated (Tick)
finalTick                                6981279948783                       # Number of ticks from beginning of simulation (restored from checkpoints and never reset) (Tick)
simFreq                                  1000000000000                       # The number of ticks per simulated second ((Tick/Second))
hostSeconds                                  15454.50                       # Real time elapsed on the host (Second)
hostTickRate                                451731162                       # The number of ticks simulated per host second (ticks/s) ((Tick/Second))
hostMemory                                    3434208                       # Number of bytes of host memory used (Byte)
simInsts                                  12160758620                       # Number of instructions simulated (Count)
simOps                                    12174871730                       # Number of ops (including micro ops) simulated (Count)
hostInstRate                                   786875                       # Simulator instruction rate (inst/s) ((Count/Second))
hostOpRate                                     787788                       # Simulator op (including micro ops) rate (op/s) ((Count/Second))
```

Simulation performance for False:
```txt
simSeconds                                   6.981280                       # Number of seconds simulated (Second)
simTicks                                 6981279948783                       # Number of ticks simulated (Tick)
finalTick                                6981279948783                       # Number of ticks from beginning of simulation (restored from checkpoints and never reset) (Tick)
simFreq                                  1000000000000                       # The number of ticks per simulated second ((Tick/Second))
hostSeconds                                  15395.96                       # Real time elapsed on the host (Second)
hostTickRate                                453448749                       # The number of ticks simulated per host second (ticks/s) ((Tick/Second))
hostMemory                                    3434208                       # Number of bytes of host memory used (Byte)
simInsts                                  12078727756                       # Number of instructions simulated (Count)
simOps                                    12092840866                       # Number of ops (including micro ops) simulated (Count)
hostInstRate                                   784539                       # Simulator instruction rate (inst/s) ((Count/Second))
hostOpRate                                     785455                       # Simulator op (including micro ops) rate (op/s) ((Count/Second))
```

robhau: @powerjg

With `False`, Ubuntu is booted about 1 minute faster than with `True`.
With `True`, about $82\cdot10^6$ more instructions are simulated compared to `False`.

Therefore, I will set the default value to false.

powerjg: A 1.0038x speedup!

This looks good to me. Thanks for all of your hard work :)

powerjg (approved):

Issue #1371, opened 2024-07-23
mem: StridePrefetcherHashedSetAssociative doesn't work


erin-le:
**Describe the bug**
When the prefetcher of a cache is set to the StridePrefetcherHashedSetAssociative prefetcher, it causes the simulation to crash.

**Affects version**
gem5 develop for 24.1, 97f6f3c4dafec239258be82a0e8d6eb23b741ec7

**gem5 Modifications**
To more easily change between prefetchers in the configuration script, I added a cache hierarchy based on `src/python/gem5/components/cachehierarchies/classic/private_l1_cache_hierarchy.py` and modified the Sconscript file in src/python to include it.

**To Reproduce**
Steps to reproduce the behavior. Please assume starting from a clean repository:

1. Add the following cache hierarchy to `src/python/gem5/components/cachehierarchies/classic`:

```python
# Copyright (c) 2021 The Regents of the University of California
# All rights reserved.
#
# Redistribution and use in source and binary forms, with or without
# modification, are permitted provided that the following conditions are
# met: redistributions of source code must retain the above copyright
# notice, this list of conditions and the following disclaimer;
# redistributions in binary form must reproduce the above copyright
# notice, this list of conditions and the following disclaimer in the
# documentation and/or other materials provided with the distribution;
# neither the name of the copyright holders nor the names of its
# contributors may be used to endorse or promote products derived from
# this software without specific prior written permission.
#
# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
[...]

PR #1374, opened 2024-07-23
mem: Fix "Need is_secure arg" prefetcher crash

erin-le:
This PR fixes the "Need is_secure arg" crash that occurs when using IndirectMemoryPrefetcher, SignaturePathPrefetcher, SignaturePathPrefetcherV2, STeMSPrefetcher, and PIFPrefetcher. This was done by changing some variables to have the type AssociativeSet<...> instead of AssociativeCache<...> and adding in "false" or an existing value for the value of the secure bit in some function calls. Further changes may be needed to move away from hard-coding values.

### DIFF
--- a/src/mem/cache/prefetch/associative_set.hh
+++ b/src/mem/cache/prefetch/associative_set.hh
@@ -90,7 +90,7 @@ class AssociativeSet : public AssociativeCache<Entry>
     using AssociativeCache<Entry>::accessEntryByAddr;
     using AssociativeCache<Entry>::findEntry;
     using AssociativeCache<Entry>::insertEntry;
-    using AssociativeCache<Entry>::getPossibleEntries;
+    // using AssociativeCache<Entry>::getPossibleEntries;
 
     using AssociativeCache<Entry>::replPolicy;
     using AssociativeCache<Entry>::indexingPolicy;
--- a/src/mem/cache/prefetch/associative_set.hh
+++ b/src/mem/cache/prefetch/associative_set.hh
@@ -90,8 +90,6 @@ class AssociativeSet : public AssociativeCache<Entry>
     using AssociativeCache<Entry>::accessEntryByAddr;
     using AssociativeCache<Entry>::findEntry;
     using AssociativeCache<Entry>::insertEntry;
-    // using AssociativeCache<Entry>::getPossibleEntries;
-
     using AssociativeCache<Entry>::replPolicy;
     using AssociativeCache<Entry>::indexingPolicy;
 };
--- a/src/mem/cache/prefetch/pif.cc
+++ b/src/mem/cache/prefetch/pif.cc
@@ -176,15 +176,15 @@ PIF::notifyRetiredInst(const Addr pc)
                 // Insert the spatial entry into the history buffer and update
                 // the 'iterator' table to point to the new entry
                 historyBuffer.push_back(spatialCompactor);
-
+                bool is_secure = false;
                 auto idx_entry = index.findEntry(spatialCompactor.trigger,
-                false);
+                is_secure);
                 if (idx_entry != nullptr) {
                     index.accessEntry(idx_entry);
                 } else {
                     idx_entry = index.findVictim(spatialCompactor.trigger);
                     assert(idx_entry != nullptr);
-                    index.insertEntry(spatialCompactor.trigger, false,
+                    index.insertEntry(spatialCompactor.trigger, is_secure,
                     idx_entry);
                 }
                 idx_entry->historyIt =
@@ -207,6 +207,7 @@ PIF::calculatePrefetch(const PrefetchInfo &pfi,
     }
 
     const Addr pc = pfi.getPC();
+    bool is_secure = pfi.isSecure();
 
     // First check if the access has been prefetched, this is done by
     // comparing the access against the active Stream Address Buffers
@@ -221,7 +222,7 @@ PIF::calculatePrefetch(const PrefetchInfo &pfi,
 
     // Check if a valid entry in the 'index' table is found and allocate a new
     // active prediction stream
-    IndexEntry *idx_entry = index.findEntry(pc, false);
+    IndexEntry *idx_entry = index.findEntry(pc, is_secure);
 
     if (idx_entry != nullptr) {
         index.accessEntry(idx_entry);
--- a/src/mem/cache/prefetch/pif.cc
+++ b/src/mem/cache/prefetch/pif.cc
@@ -177,13 +177,15 @@ PIF::notifyRetiredInst(const Addr pc)
                 // the 'iterator' table to point to the new entry
                 historyBuffer.push_back(spatialCompactor);
 
-                auto idx_entry = index.findEntry(spatialCompactor.trigger);
+                auto idx_entry = index.findEntry(spatialCompactor.trigger,
+                false);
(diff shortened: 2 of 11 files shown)


giactra:
Yeah, I have a patchset which is fixing this by allowing the AssociativeCache to use generic tags... Anyway it makes sense to patch this for now.. This should definitely be backported into v24.0 I believe @BobbyRBruce

src/mem/cache/prefetch/associative_set.hh
```diff
@@ -90,7 +90,7 @@ class AssociativeSet : public AssociativeCache<Entry>
     using AssociativeCache<Entry>::accessEntryByAddr;
     using AssociativeCache<Entry>::findEntry;
     using AssociativeCache<Entry>::insertEntry;
-    using AssociativeCache<Entry>::getPossibleEntries;
+    // using AssociativeCache<Entry>::getPossibleEntries;
```
hnpl: Can you either remove this line, or provide comments in the code on why this line is commented out?

src/mem/cache/prefetch/pif.cc
```diff
@@ -176,14 +176,16 @@ PIF::notifyRetiredInst(const Addr pc)
                 // Insert the spatial entry into the history buffer and update
                 // the 'iterator' table to point to the new entry
                 historyBuffer.push_back(spatialCompactor);
-
-                auto idx_entry = index.findEntry(spatialCompactor.trigger);
+                bool is_secure = false;
```
hnpl: Since the `is_secure` variable is not changed afterwards, can you change this to `constexpr bool is_secure = false;`?

src/mem/cache/prefetch/signature_path.cc
```diff
@@ -192,7 +192,8 @@ SignaturePath::getSignatureEntry(Addr ppn, bool is_secure,
 SignaturePath::PatternEntry &
 SignaturePath::getPatternEntry(Addr signature)
 {
-    PatternEntry* pattern_entry = patternTable.findEntry(signature);
+    bool is_secure = false;
```
hnpl: Since the `is_secure` variable is not changed afterwards, can you change this to `constexpr bool is_secure = false;`?

src/mem/cache/prefetch/signature_path_v2.cc
```diff
@@ -125,7 +125,8 @@ SignaturePathV2::handlePageCrossingLookahead(signature_t signature,
     GlobalHistoryEntry *gh_entry = globalHistoryRegister.findVictim(0);
     assert(gh_entry != nullptr);
     // Any address value works, as it is never used
-    globalHistoryRegister.insertEntry(0, gh_entry);
+    bool is_secure = false;
```
hnpl: Since the `is_secure` variable is not changed afterwards, can you change this to `constexpr bool is_secure = false;`?

src/mem/cache/prefetch/signature_path_v2.cc
```diff
@@ -125,7 +125,8 @@ SignaturePathV2::handlePageCrossingLookahead(signature_t signature,
...
+    globalHistoryRegister.insertEntry(0, is_secure, gh_entry); // false,
```
hnpl: Can you remove this comment?

hnpl (commented):
There are a couple of nitpicking comments, but other than that, the code looks good to me.

powerjg:
@erin-le can you make the updates suggested by Hoa?

odanrc:
I don´t really understand what is the bug from the PR description, so I have a couple of questions.

I see that this PR mostly reverts https://github.com/gem5/gem5/pull/745 within the prefetcher realm. Doesn´t the bug also affect src/mem/cache/prefetch/delta_correlating_prediction_tables.hh and other places that use AssociativeCache? Why is this only a partial - although close to full - reversal? i.e., why don´t we fully revert PR 745?

hnpl:
The `Need is_secure arg` panic refers to a gem5 panic related to `TaggedEntry`.
https://github.com/gem5/gem5/pull/745/files#diff-eccdc055cc61dd6193ad97da8fa8f08756cf86d59e9d0c3bad8596267741709aR115

I'm not sure if fully reverting #745 for now makes more sense. Given that @giactra already has some patchset for this bug, this PR is more of a temporary fix.

@erin-le Thanks! There is a comment that was not addressed. Also, can you test other prefetchers that were affected by #745, such as `SlimAMPM`?

erin-le:
Sorry about that, it should be removed now. Regarding testing, I went through the following prefetchers

```
MultiPrefetcher StridePrefetcherHashedSetAssociative StridePrefetcher TaggedPrefetcher 
IndirectMemoryPrefetcher SignaturePathPrefetcher SignaturePathPrefetcherV2 
AMPMPrefetcher DCPTPrefetcher IrregularStreamBufferPrefetcher SlimAMPMPrefetcher 
BOPPrefetcher SBOOEPrefetcher STeMSPrefetcher PIFPrefetcher
```
which should be all of the ones in `src/mem/cache/prefetch/Prefetcher.py`.

After the changes, all of them except for StridePrefetcherHashedSetAssociative work. The error message is as follows.

```
TypeError: BasePrefetcher.__init__() takes 1 positional argument but 2 were given
Error setting param L1ICache.prefetcher to <orphan StridePrefetcherHashedSetAssociative>
```
This didn't look like it was related to the crashes that were patched in this PR, so I opened a separate issue for it [here](https://github.com/gem5/gem5/issues/1371).

src/mem/cache/prefetch/spatio_temporal_memory_streaming.cc
```diff
@@ -219,9 +222,11 @@ STeMS::reconstructSequence(
 
     // Now query the PST with the PC of each RMOB entry
     idx = 0;
+    bool is_secure = false;
```
hnpl: Ditto

hnpl (approved):
Thanks for looking into this @erin-le. There are a couple of nitpicks, but other than that, the code looks good to me!

src/mem/cache/prefetch/spatio_temporal_memory_streaming.cc
```diff
@@ -92,12 +92,15 @@ STeMS::checkForActiveGenerationsEnd(const CacheAccessor &cache)
             }
             if (generation_ended) {
                 // PST is indexed using the PC (secure bit is unused)
-                auto pst_entry = patternSequenceTable.findEntry(pst_addr);
+                bool is_secure = false;
```
hnpl: Can you change this to `constexpr bool is_secure = false;`?

hnpl (approved):

BobbyRBruce (approved):

BobbyRBruce:
@giactra @erin-le @hnpl : I'm merging this as it Erin appears to have addressed all the concerns and it looks good to me. However, i don't have a good grasp of the importance of this. Is this something that should be hotfixed to stable? It's fixing a bug impacting users? I know Giacomo said this 2 weeks back but just triple checking this is the case.

hnpl:
@BobbyRBruce I think this is a candidate for a hotfix. The affected prefetchers are unusable without this fix.

BobbyRBruce:
Ok. I'll get that done this week.

BobbyRBruce:
Hotfix created here: https://github.com/gem5/gem5/pull/1425

PR #1378, opened 2024-07-24
arch-vega: Improve SDWA, SDWAB, and DPP

abmerop:
This PR has four components:

- Implement a helper method for SDWAB similar to SDWA helpers. SDWAB is used for VOPC instructions only (vector compares).
- Update two instructions commonly using SDWAB to use helper (v_cmp_ne_u16 and v_cmp_eq_u16).
- Add panics to *all* VOP1 and VOP2 instructions which do not implement SDWA or DPP if they use an SDWA or DPP register.
- Add panics to *all* VOPC instructions which do not implement SDWAB or DPP if they are an SDWA or DPP register.

Only VOP1, VOP2, and VOPC may use SDWA, SDWAB, or DPP. The panics should therefore cover all instructions which have missing implementations for these modes. The intent is to exit gem5 instead of continuing simulation will data that is likely incorrect. Continuing simulation only makes debugging gem5 more difficult.

### DIFF
--- a/src/arch/amdgpu/vega/insts/op_encodings.hh
+++ b/src/arch/amdgpu/vega/insts/op_encodings.hh
@@ -35,6 +35,7 @@
 #include "arch/amdgpu/vega/gpu_decoder.hh"
 #include "arch/amdgpu/vega/gpu_mem_helpers.hh"
 #include "arch/amdgpu/vega/insts/gpu_static_inst.hh"
+#include "arch/amdgpu/vega/insts/inst_util.hh"
 #include "arch/amdgpu/vega/operand.hh"
 #include "debug/GPUExec.hh"
 #include "debug/VEGA.hh"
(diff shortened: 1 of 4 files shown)


src/arch/amdgpu/vega/insts/op_encodings.hh
```diff
@@ -421,6 +422,145 @@ namespace VegaISA
...
 
+        template<typename T>
+        uint32_t
+        sdwabSelect(uint32_t dword, const SDWASelVals sel,
+                    bool sign_ext, bool neg, bool abs)
+        {
+            int low_bit = 0, high_bit = 0;
+            uint32_t rv = dword;
+
+            if (sel < SDWA_WORD_0) {
```
mattsinc, line 437: Ideally we should add comments explaining what each of these are doing.

src/arch/amdgpu/vega/insts/op_encodings.hh
```diff
@@ -421,6 +422,145 @@ namespace VegaISA
...
+                high_bit = low_bit + VegaISA::MSB_PER_WORD - 1;
+            } else {
+                low_bit = 0;
+                high_bit = sizeof(uint32_t) * VegaISA::BITS_PER_BYTE - 1;
+                assert(sel == SDWA_DWORD);
+            }
+
+            rv = bits(dword, high_bit, low_bit);
+
+            // Apply ABS, then NEG, then SEXT.
```
mattsinc: looks like this comment should go before line 455?

src/arch/amdgpu/vega/insts/op_encodings.hh
```diff
@@ -421,6 +422,145 @@ namespace VegaISA
...
+            }
+
+            if (sign_ext) {
+                if (std::is_integral_v<T>) {
+                    if (rv & sign_bit) {
+                        rv = sext(rv, high_bit + 1) & 0xFFFFFFFF;
+                    }
+                } else {
+                    // It is not entirely clear what to do here. Literal
+                    // extensions for FP operands appends zeros to mantissa
```
mattsinc: appends --> append

src/arch/amdgpu/vega/insts/vopc.cc
```diff
@@ -3874,17 +4089,27 @@ namespace VegaISA
     Inst_VOPC__V_CMP_NE_U16::execute(GPUDynInstPtr gpuDynInst)
     {
         Wavefront *wf = gpuDynInst->wavefront();
-        ConstVecOperandU16 src0(gpuDynInst, instData.SRC0);
-        ConstVecOperandU16 src1(gpuDynInst, instData.VSRC1);
+        ConstVecOperandU32 src0(gpuDynInst, instData.SRC0);
```
mattsinc: why is U32 correct for a U16 instruction?  Maybe because SDWA needs an extra bit internally?

mattsinc (changes requested):

src/arch/amdgpu/vega/insts/vopc.cc
abmerop: GPU always reads 32-bits into a register.  In order to be able to select upper or lower 16-bits, we need to read the full value which the U16 types will not do.  This will become uint16_t when the lambda function is called (upper 16-bits get cropped).

src/arch/amdgpu/vega/insts/op_encodings.hh
abmerop: Done

src/arch/amdgpu/vega/insts/vopc.cc
abmerop: Actually, it seems this doesn't matter here.  The `sdwabHelper` will read the 32-bit value if it is SDWAB. Otherwise it reads (correctly) the lower 16-bits.  

I noticed I did not change this in the other compare so I can either revert this change here or apply the change to the other instruction making it U32.  Leaning towards the former.

mattsinc: I think reverting is fine then.

abmerop: Done

mattsinc (approved):

PR #1379, opened 2024-07-24
arch-vega: Fix unconditional clamps in VOP3

abmerop:
Some instructions are clamping floating point outputs unconditionally, leading to incorrect results. This commit finds instructions with this issue and checks the clamp bit before applying clamp.

Change-Id: Ibc6de3813d81fd4f9d2c98dd497d19dd34cf6bde

### DIFF
--- a/src/arch/amdgpu/vega/insts/vop3.cc
+++ b/src/arch/amdgpu/vega/insts/vop3.cc
@@ -2464,7 +2464,9 @@ namespace VegaISA
             if (wf->execMask(lane)) {
                 float out = std::fma(src0[lane], src1[lane], vdst[lane]);
                 out = omodModifier(out, extData.OMOD);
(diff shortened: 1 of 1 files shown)


mattsinc (approved):
Is there a good way to verify this is really all that are affected?

abmerop:
> Is there a good way to verify this is really all that are affected?

These are the only three instruction which check the bit.  I think VOP3 will probably need to add some panics that modifiers are ignored similar to the other PR I posted today.

mattsinc:
> > Is there a good way to verify this is really all that are affected?
> 
> These are the only three instruction which check the bit. I think VOP3 will probably need to add some panics that modifiers are ignored similar to the other PR I posted today.

Ok, I guess we can handle that separately

PR #868, opened 2024-02-14
arch-riscv: Fix fflags behavior of float inst. in O3 CPU

rogerchang23424:
According to the RISC-V spec [1]. Any float-point instructions accumulate FFLAGS register rather than write it to reflect the CSR behavior.

In the previous implementation. We read the FFLAGS, set the exception flags, and write the result back to the FFLAGS. This works in the gem5 simple and minor CPU model as they are actually written to `regFile` after executing the instructions. However, in the gem5 O3 CPU model, it will record in the `destMiscReg` buffer until the commit stage when writing to the `miscReg` in the execution stage. The next instruction will get the old FFLAGS and cause the incorrect result.

The CL introduced the `MISCREG_FFLAGS_EXE` and used the same size of `miscRegFile` because the `MISCREG_FFLAGS_EXE` and `MISCREG_FFLAGS` shared the same space. When executing the float-pointing instruction, any exception flags should be updated via `MISCREG_FFLAGS_EXE` to accumulate the FFLAGS in `setMiscReg` method. For the MISCREG_FFLAGS, it should only be called in the CSROp.

[1] Syntactic Dependencies: Appendix A
https://github.com/riscv/riscv-isa-manual/blob/c80ecada1ccc3120e2c8003bd7ca8e91c2e91bf9/src/mm-eplan.adoc#syntactic-dependencies-rules-9-11

gem5 issue: https://github.com/gem5/gem5/issues/755

Change-Id: Ib7f13d95b8a921c37766a54a217a5a4b1ef17c6f

Issue #755, opened 2024-01-10
Failed to pass the float riscv-tests in O3CPU


rogerchang23424:
**Describe the bug**
Get the error when running the `rv64uf-ps-fadd` in SE mode

**Affects version**
v22.0.0.2 v22.1.0.0 v23.0.1.0 v23.1.0.0

**To Reproduce**

Build gem5
```bash
scons -j$(nproc) build/RISCV/gem5.opt
```

Run the command
For v22.0 run
```bash
build/RISCV/gem5.opt --debug-flags=Exec configs/example/se.py --cmd tests/gem5/resources/rv64uf-ps-fadd/rv64uf-ps-fadd --cpu-type DerivO3CPU --caches
```

For v22.1 and v23.0, run 
```bash
build/RISCV/gem5.opt --debug-flags=Exec tests/gem5/configs/simple_binary_run.py rv64uf-ps-fadd o3 riscv --num-cores 4 --resource-directory tests/gem5/resources
```

For v23.1
```bash
build/RISCV/gem5.opt --debug-flags=Exec tests/gem5/asmtest/configs/riscv_asmtest.py rv64uf-ps-fadd o3 --num-cores 4 --resource-directory tests/gem5/resources
```

**Terminal Output**

```bash
gem5 version 22.0.0.2
gem5 compiled Jan 10 2024 00:06:11
gem5 started Jan 10 2024 00:28:21
gem5 executing on 4aafa3c79b5b, pid 2715
command line: build/RISCV/gem5.opt --debug-flags=Exec configs/example/se.py --cmd tests/gem5/resources/rv64uf-ps-fadd/rv64uf-ps-fadd --cpu-type DerivO3CPU --caches

Global frequency set at 1000000000000 ticks per second
build/RISCV/mem/dram_interface.cc:690: warn: DRAM device capacity (8192 Mbytes) does not match the address range assigned (512 Mbytes)
build/RISCV/arch/riscv/linux/se_workload.cc:60: warn: Unknown operating system; assuming Linux.
0: system.remote_gdb: listening for remote gdb on port 7000
**** REAL SIMULATION ****
build/RISCV/sim/simulate.cc:194: info: Entering event queue @ 0.  Starting simulation...
  78500: system.cpu: T0 : 0x80000000 @_start    : auipc t0, 0                : IntAlu :  D=0x0000000080000000
[...]

PR #1387, opened 2024-07-26
Hypervisor (H) extension for RISCV

Navyera:
This set of commits provides an initial implementation of the H-extension for RISCV cores in gem5.

The implementation seems functional for the AtomicCPU, however more changes are needed for O3CPU to function correctly in H-extension scenarios. We plan to add these changes in a follow-up PR possibly.

In summary the implemented features include:

- New CSRs mandated by the H-extension
- CSR write mask support (CSRExecute rework)
- Additional fault handling & delegation
- Extensions for existing instruction implementations
- G-Stage translation support
- Two-stage translation support
- New instructions mandated by the H-extension


src/arch/riscv/isa/formats/fp.isa
```diff
@@ -36,20 +36,20 @@ def template FloatExecute {{
...
         %(op_rd)s;
 
+        RegVal FFLAGS = xc->readMiscReg(MISCREG_FFLAGS);
         std::feclearexcept(FE_ALL_EXCEPT);
         %(code)s;
-        xc->setMiscReg(MISCREG_FFLAGS_EXE, softfloat_exceptionFlags);
+
+        FFLAGS |= softfloat_exceptionFlags;
         softfloat_exceptionFlags = 0;
+        xc->setMiscReg(MISCREG_FFLAGS, FFLAGS & FFLAGS_MASK);
```
rogerchang23424: If you do the change again. It will cause fflags issue in O3CPU https://github.com/gem5/gem5/issues/755

BobbyRBruce:
The RISCV tests are failing with:

```
src/arch/riscv/pagetable_walker.cc:454: panic: panic condition misa.rvh occurred: Timing walks are not supported with h extension
```

src/arch/riscv/RiscvISA.py
```diff
@@ -106,7 +107,9 @@ class RiscvISA(BaseISA):
         ELEN in Ch. 2 of RISC-V vector spec",
     )
     privilege_mode_set = Param.PrivilegeModeSet(
-        "MSU",
+        "MHSU",  # MHSU to enable hypervisor (H-extension)
```
hnpl: The MHSU mode should not be the default if the timing cpu/timing mode is not supported.

powerjg:
Thanks for the contribution! I'm excited to dig in to review the change, and I'm excited for this feature to be merged!

src/arch/riscv/isa/formats/fp.isa
Navyera: Hey, thanks for the comment, we had doubts about that part. Could you briefly explain what the bug is here? I read #755  but from the traces given I do not yet understand the "logic" behind the bug. Should we set FFLAGS instead of OR'ing?

src/arch/riscv/RiscvISA.py
Navyera: Yeah that is totally reasonable, I just wanted to see how many of the tests were passing with H on. If H-extension is disabled a lot of logic is guarded out. 

I am going to set the default to MSU and push, hopefully tests will pass :)

src/arch/riscv/isa/formats/fp.isa
rogerchang23424: It means we will get the wrong fflags result in O3CPU compared with AtomicCPU, you can test via command

```bash
build/RISCV/gem5.opt --debug-flags=Exec tests/gem5/configs/simple_binary_run.py rv64uf-ps-fadd o3 riscv --num-cores 4 --resource-directory tests/gem5/resources
```

For atomic CPU
```bash
build/RISCV/gem5.opt --debug-flags=Exec tests/gem5/configs/simple_binary_run.py rv64uf-ps-fadd atomic riscv --num-cores 4 --resource-directory tests/gem5/resources
```
> Should we set FFLAGS instead of OR'ing?

No, we accumulate the FFLAGS in each float arithmetic operation.

It's because the O3CPU[3] handles setMiscReg in different way compare to other CPU models, SimpleCPU[1] and MinorCPU[2]. The O3CPU wouldn't write the value back to MiscRegFile until the instruction is committed, we may get the old FFLAGS value in the next float instruction, and cause the riscv-tests failed. So that PR https://github.com/gem5/gem5/pull/868  use eatra misc_reg index `MISCREG_FFLAGS_EXE` to accumulate the FFLAGS to `MISCREG_FFLAGS` `miscRegFile`, just like what is implemented in the ARM[4]. ARM handles each float arithmetic exception via `MISCREG_FPSCR_EXC` rather than read/write `MISCREG_FPSCR` directly.

[1] SimpleCPU
https://github.com/gem5/gem5/blob/d2c8754ab3b72e52a07d62662de0120b941f1ae5/src/cpu/simple/exec_context.hh#L259-L263

[2] MinorCPU
https://github.com/gem5/gem5/blob/d2c8754ab3b72e52a07d62662de0120b941f1ae5/src/cpu/minor/exec_context.hh#L263-L267

[3] O3CPU
https://github.com/gem5/gem5/blob/d2c8754ab3b72e52a07d62662de0120b941f1ae5/src/cpu/o3/dyn_inst.hh#L1016-L1032

[4] 
https://github.com/gem5/gem5/blob/47c4dad8694448c40435875309a8399abebd7fc2/src/arch/arm/isa.cc#L890-L896

I seperated it with `NUM_PHYS_MISCREGS` and `NUM_MISCREGS` just because there is no reason to upgrade the checkpoint and avoid the checkpoint breakage.

src/arch/riscv/regs/misc.hh
```diff
@@ -200,6 +201,42 @@ enum MiscRegIndex
...
+    MISCREG_VSCAUSE,
+    MISCREG_VSTVAL,
+    MISCREG_VSATP,
+
+    // This register stores the V-bit.
+    // It is not mandated by the specification,
+    // however it is useful for the implementation.
+    MISCREG_VIRT,
+
+    // H-extension registers end here
```
rogerchang23424, line 240: You may need to update the checkpoint upgrade `util/cpt_upgraders`

src/arch/riscv/isa.cc
```diff
@@ -984,6 +1003,250 @@ ISA::resetThread()
...
+    auto new_reg_data_all = (reg_data_all & ~write_mask)
+                          | (writeData & write_mask);
+
+
+    switch (csr) {
+        case CSR_FCSR: {
+            setMiscReg(MISCREG_FFLAGS, bits(writeDataMasked, 4, 0));
+            setMiscReg(MISCREG_FRM, bits(writeDataMasked, 7, 5));
+            break;
+        }
```
rogerchang23424: (nit) I hope we can align the implement of `MISCREG_VCSR`. They are do similar things.

https://github.com/gem5/gem5/blob/d2c8754ab3b72e52a07d62662de0120b941f1ae5/src/arch/riscv/isa.cc#L595-L599

src/arch/riscv/isa/formats/standard.isa
```diff
@@ -339,7 +339,7 @@ def template CSRExecute {{
         }
 
         RegIndex midx = csr_data_it->second.physIndex;
-        const std::string& csrName = csr_data_it->second.name;
+        std::string csrName = csr_data_it->second.name;
```
rogerchang23424, line 342: (nit) Is any reason to change it to copy the string from csr_data to csrName?

src/arch/riscv/isa/formats/standard.isa
```diff
@@ -394,7 +433,7 @@ def template CSRExecute {{
             }
             DPRINTF(RiscvMisc, "Writing %#x to CSR %s.\n",
                     data, csrName);
-            xc->setMiscReg(midx, data);
+            isa->writeCSR(csr, data);
```
rogerchang23424: It may affect the statistics of MiscReg if the `writeCSR` doesn't set miscReg value via `xc->setMiscReg`

https://github.com/gem5/gem5/blob/d2c8754ab3b72e52a07d62662de0120b941f1ae5/src/cpu/simple/exec_context.hh#L259-L263

rogerchang23424 (commented):
I don't look into the details of hypervisor spec. I just do review briefly

src/arch/riscv/isa.cc
```diff
@@ -984,6 +1003,250 @@ ISA::resetThread()
...
+    return readval;
+}
+
+RegVal
+ISA::readCSR(uint64_t csr)
+{
+    auto csr_it = CSRData.find(csr);
+
+    // panic if the method was used with bad csr idx
+    panic_if(csr_it == CSRData.end(), "Illegal CSR passed to readCSR");
```
rogerchang23424: Raise `IllegalInst` instead of `panic_if`?

ivanaamit:
Hi @Navyera, could you please address Roger's comments? Thank you.

Navyera:
Hi @ivanaamit , @rogerchang23424. I am sorry for the late reply, I was on summer vacation, but I'm now back and looking into the comments and test issues. I'll provide an update as soon as possible.

src/arch/riscv/isa/formats/standard.isa
Navyera: It can no longer be const or reference due to possible remapping (reassignment) by swapToVirtCSR method.

src/arch/riscv/isa.cc
Navyera: I think it has been checked at the start of CSR execute (standard.isa:335).

```
auto csr_data_it = csr_data.find(csr);
if (csr_data_it == csr_data.end()) {
    return std::make_shared<IllegalInstFault>(
        csprintf("Illegal CSR index %#x\n", csr), machInst);
}
```

So the panic_if in readCSR is just defensive programming.

src/arch/riscv/isa/formats/standard.isa
Navyera: I think all paths of writeCSR result in setMiscReg being called

src/arch/riscv/regs/misc.hh
Navyera: We have locally updated the cpt_upgrader but we have this failing test: test-gem5-library-riscv-hello-restore-checkpoint-ALL-x86_64-opt.

The code tries to restore some of the new MISCREGs for H-extension and crashes.

Do we have to upgrade the checkpoint beforehand? Would the code automatically try to do that?

Navyera:
Hi @rogerchang23424 @ivanaamit , we have addressed all of the comments hopefully.

Currently the only failing test is `test-gem5-library-riscv-hello-restore-checkpoint-ALL-x86_64-opt` which probably uses a precomputed checkpoint as a resource.

How should we go about updating that resource to include default values for the new H-extension registers we have added?

Harshil2107:
Hi, to update the checkpoint you would probably need to update the `checkpoint_upgrader` in `util/cpt-upgrader` and `util/cpt-upgrader.py`. I can then update the checkpoint and upload the new one to gem5 resources.

Navyera:
Hi @Harshil2107 , I have updated the checkpoint upgrader and in my local testing, checkpoints seem to be succesfully upgraded. Let me know if you encounter any weirdness with the `riscv-hext.py` upgrader!

Navyera:
Hi @Harshil2107 @ivanaamit , the last failed test run has the same issue with the checkpoint that is not upgraded for the H-extension.

We have already added the relevant upgrader, but the resource does not seem to be updated yet.

On local runs I am getting the following:

```
command line: ./build/ALL/gem5.opt ./configs/example/gem5_library/checkpoints/riscv-hello-restore-checkpoint.py

info: Using default config
Resource 'riscv-hello-example-checkpoint' was not found locally. Downloading to '/home/nick/.cache/gem5/riscv-hello-example-checkpoint.tar'...
Finished downloading resource 'riscv-hello-example-checkpoint'.
Unpacking the the resource 'riscv-hello-example-checkpoint' ('/home/nick/.cache/gem5/riscv-hello-example-checkpoint.tar')
src/mem/dram_interface.cc:690: warn: DRAM device capacity (8192 Mbytes) does not match the address range assigned (32 Mbytes)
src/arch/riscv/isa.cc:311: info: RVV enabled, VLEN = 256 bits, ELEN = 64 bits
src/sim/globals.cc:91: warn: **********************************************************
src/sim/globals.cc:92: warn: !!! Checkpoint is missing the following version tags:
src/sim/globals.cc:96: warn:   riscv-hext
src/sim/globals.cc:96: warn:   x86-add-xcr0
src/sim/globals.cc:100: warn: You might experience some issues when restoring and should run the checkpoint upgrader (util/cpt_upgrader.py) on your checkpoint
src/sim/globals.cc:103: warn: **********************************************************
src/sim/serialize.hh:379: fatal: fatal condition !paramInImpl(cp, name, param) occurred: Can't unserialize 'board.processor.cores.core.xc.0:hvip_ulong'
Memory Usage: 278596 KBytes

```

Harshil2107:
Hi @Navyera ,

Sorry for the delay, I didn't get the notification of the original ping. I will run the upgrader on the checkpoints on resources and update them.

Harshil2107:
I am trying to restore the following checkpoint https://resources.gem5.org/resources/riscv-hello-example-checkpoint/raw?database=gem5-resources&version=3.0.0 but it seems like it is missing some keys. I just wanted to verify if this was the checkpoint you were able to upgrade locally. This is the output I get when trying to restore the checkpoint.

I just downloaded the checkpoint from the `url`and untarred it, then ran the upgrader.
```bash
python3 ./util/cpt_upgrader.py -v  ./riscv-checkpoint-v4/
Processing file ./riscv-checkpoint-v4/m5.cpt.... 

has tags riscv-vext memory-per-range arm-sysreg-mapping-ns armv8 riscv-dyn-vlen mempool-to-seworkload multiple-event-queues smt-interrupts arm-ccregs dvfs-perflevel arm-sve cpu-pid arm-gicv2-banked-regs x86-add-tlb etherswitch globals-to-root riscv-pcstate arm-miscreg-teehbr register-files ide-dma-abort arm-hdlcd-upgrade ruby-block-size-bytes process-fdmap-rename mempool-sections arm-contextidr-el2 isa-is-simobject remove-arm-cpsr-mode-miscreg arm-gem5-gic-ext 

Traceback (most recent call last):
  File "/usr/lib/python3.10/configparser.py", line 791, in get
    value = d[option]
  File "/usr/lib/python3.10/collections/__init__.py", line 986, in __getitem__
    return self.__missing__(key)            # support subclasses that define __missing__
  File "/usr/lib/python3.10/collections/__init__.py", line 978, in __missing__
    raise KeyError(key)
KeyError: 'regVal'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/harshilp/worktrees/gem5/./util/cpt_upgrader.py", line 363, in <module>
    process_file(cpt_file, **vars(args))
  File "/home/harshilp/worktrees/gem5/./util/cpt_upgrader.py", line 266, in process_file
    Upgrader.get(tag).update(cpt, tags)
  File "/home/harshilp/worktrees/gem5/./util/cpt_upgrader.py", line 152, in update
    self.upgrader(cpt)
  File "<string>", line 50, in upgrader
  File "/usr/lib/python3.10/configparser.py", line 794, in get
    raise NoOptionError(option, section)
configparser.NoOptionError: No option 'regVal' in section: 'board.processor.cores.core.isa'
```
I wanted to verify this as it also could be that the checkpoint it self is missing some stuff.

Navyera:
@Harshil2107 That is the behaviour I noticed locally as well... I filled in the missing keys by hand and then I was able to upgrade the checkpoint successfully. I thought that they were missing on purpose to make the checkpoint generic but maybe it is a problem...

Harshil2107:
do you think it is possible to add the missing field in the hypervisor upgrader? I think completely re taking the checkpoint will pass the tests but not having a upgrader might break checkpoints that other users might have.

Navyera:
Unless I am misunderstanding what you mean, I do not think that hardcoding the addition of the missing keys in the hypervisor upgrader is a robust solution. 

For example, one of the missing keys is only requested by an x86 upgrader. It would make no sense to add these keys in a RISCV upgrader imo...

```
nick@tommy:~/CAL/hypervisor/gem5$ grep -R "regVal" ./util/cpt_upgraders/ 
./util/cpt_upgraders/isa-is-simobject.py:        "x86": ("regVal"),
./util/cpt_upgraders/x86-add-xcr0.py:                regVals = cpt.get(sec, "regVal")
./util/cpt_upgraders/x86-add-xcr0.py:                regVals = f"{regVals} 1"
./util/cpt_upgraders/x86-add-xcr0.py:                cpt.set(sec, "regVal", regVals)
./util/cpt_upgraders/x86-add-xcr0.py:            regVals = cpt.get(sec, "regVal")
./util/cpt_upgraders/x86-add-xcr0.py:            regVals = f"{regVals} 1"
./util/cpt_upgraders/x86-add-xcr0.py:            cpt.set(sec, "regVal", regVals)

```

Harshil2107:
It looks like this is the issue, I will go through the existing upgraders and update them so only riscv ones are run to update the riscv checkpoints.

Harshil2107:
It looks like the main issue is that the `isaName` field is missing from the checkpoint. This is an issue on multiple checkpoints in gem5 resources. I will see if I can write another updater that can automatically add this field correctly to the checkpoints.

src/arch/riscv/isa/decoder.isa
```diff
@@ -4835,41 +4804,93 @@ decode QUADRANT default Unknown::unknown() {
...
+                                if (misa.rvh) {
+                                    if (hstatus.spv) {
+                                        setV(xc);
+                                    }
+                                    hstatus.spv = 0;
+                                    xc->setMiscReg(MISCREG_HSTATUS, hstatus);
+                                }
+
+                                status.mprv = 0; // 20211203
                             }
```
rogerchang23424: The part should handle before set status CSR

src/arch/riscv/isa.cc
```diff
@@ -985,6 +1018,250 @@ ISA::resetThread()
...
+            (readMiscReg(MISCREG_FRM) << FRM_OFFSET));
+    }
+
+    return readval;
+}
+
+RegVal
+ISA::readCSR(uint64_t csr)
+{
+    auto csr_it = CSRData.find(csr);
```
rogerchang23424: `getCSRDataMap()`? So that we can get customized CSR without panic

Ref: https://gem5-review.googlesource.com/c/public/gem5/+/62891

src/arch/riscv/isa.cc
```diff
@@ -985,6 +1018,250 @@ ISA::resetThread()
...
+        default: break;
+    }
+
+    return readval;
+}
+
+void
+ISA::writeCSR(uint64_t csr, RegVal writeData)
+{
+    auto csr_it = CSRData.find(csr);
```
rogerchang23424: ditto

rogerchang23424 (commented):
Some cooment is the PR. Most of part is LGTM.

src/arch/riscv/fault_codes.hh
powerjg, line 1: Needs a copyright statement.

src/arch/riscv/RiscvISA.py
```diff
@@ -107,7 +108,10 @@ class RiscvISA(BaseISA):
         ELEN in Ch. 2 of RISC-V vector spec",
     )
     privilege_mode_set = Param.PrivilegeModeSet(
-        "MSU",
+        "MSU",  # set MHSU to enable hypervisor (H-extension)
+        # O3 CPU is not supported in MHSU currently
```
powerjg: Why does O3CPU not work? This is not a problem, but we need to have some detailed documentation.

src/arch/riscv/isa.cc
```diff
@@ -242,7 +275,7 @@ namespace RiscvISA
     [MISCREG_HPMCOUNTER30H]  = "HPMCOUNTER30H",
     [MISCREG_HPMCOUNTER31H]  = "HPMCOUNTER31H",
 
-    [MISCREG_FFLAGS_EXE]    = "FFLAGS_EXE",
+    [HELPER_FFLAGS_EXE] = "FFLAGS_EXE",
```
powerjg: Why the change from MISCREG->HELPER?

src/arch/riscv/isa.cc
```diff
@@ -307,7 +339,7 @@ ISA::copyRegsFrom(ThreadContext *src)
     }
 
     // Copying Misc Regs
-    for (int i = 0; i < NUM_PHYS_MISCREGS; i++)
+    for (int i = 0; i < NUM_MISCREGS; i++)
```
powerjg: Why this change?

src/arch/riscv/isa.cc
```diff
@@ -705,7 +728,7 @@ ISA::setMiscReg(RegIndex idx, RegVal val)
                         res |= (old_val & (0xFF << (8*i)));
                     }
                 }
-
+                tc->getMMUPtr()->flushAll();
```
powerjg, line 765: I assume it is in the spec that you have to flush the TLB when the PMP is updated?

src/arch/riscv/isa.cc
```diff
@@ -985,6 +1018,250 @@ ISA::resetThread()
     Reset().invoke(tc);
 }
 
+bool
+ISA::isV() const
```
powerjg: Can we name this something more descriptive? Maybe "isInVirtualMode" or something like that?

src/arch/riscv/memflags.hh
```diff
@@ -0,0 +1,26 @@
...
+    // We can only utilize the lower 8 bits of a
+    // 64-bit value to encode these.
+    // see src/mem/request.hh ARCH_BITS
+    // Lower 3 bits are already used in mmu.hh
+    // for alignment flags
+    enum XlateFlags
+    {
+        HLVX = 1ULL << 3,
+        FORCE_VIRT = 1ULL << 4,
+        LR = 1ULL << 5,
```
powerjg: Can you add a comment to each of these which explains what "HLVX", "FORCE_VIRT", and "LR" means

src/arch/riscv/mmu.hh
```diff
@@ -76,10 +76,11 @@ class MMU : public BaseMMU
                 PageBytes, start, size, tc, this, mode, flags));
     }
 
-    PrivilegeMode
-    getMemPriv(ThreadContext *tc, BaseMMU::Mode mode)
+    MemAccessInfo
+    getMemAccessInfo(ThreadContext *tc, BaseMMU::Mode mode)
     {
-        return static_cast<TLB*>(dtb)->getMemPriv(tc, mode);
```
powerjg, line 91: Is this function no longer needed?

src/arch/riscv/regs/misc.hh
```diff
@@ -246,22 +283,11 @@ enum MiscRegIndex
...
-    MISCREG_MIE = MISCREG_IE,
-    // This CSR shared the same space with MISCREG_FFLAGS
-    MISCREG_FFLAGS_EXE = NUM_PHYS_MISCREGS,
-    MISCREG_FCSR,
-    MISCREG_USTATUS,
-    MISCREG_UIP,
-    MISCREG_UIE,
-    MISCREG_SSTATUS,
-    MISCREG_SIP,
-    MISCREG_SIE,
```
powerjg: Why is it OK to delete these?

src/arch/riscv/tlb.cc
```diff
@@ -310,56 +340,137 @@ TLB::doTranslate(const RequestPtr &req, ThreadContext *tc,
...
-                delayed = true;
-                return fault;
+    Fault fault;
+    // No check on special_access
+    if (special_access) {
+        fault = NoFault;
+    }
+    else {
+
+
```
powerjg: Extra newlines

src/arch/riscv/tlb.cc
```diff
@@ -310,56 +340,137 @@ TLB::doTranslate(const RequestPtr &req, ThreadContext *tc,
...
     delayed = false;
 
-    Addr vaddr = Addr(sext<VADDR_BITS>(req->getVaddr()));
-    SATP satp = tc->readMiscReg(MISCREG_SATP);
+    MemAccessInfo memaccess = getMemAccessInfo(tc, mode, req->getArchFlags());
+    Addr vaddr = req->getVaddr();
+
+    MISA misa = tc->readMiscReg(MISCREG_ISA);
+
+    bool special_access = false
```
powerjg, line 353: What does "special" mean in this context?

src/arch/riscv/tlb.cc
```diff
@@ -310,56 +340,137 @@ TLB::doTranslate(const RequestPtr &req, ThreadContext *tc,
...
+        fault = walker->start(tc, translation, req, mode);
+        if (translation != nullptr || fault != NoFault) {
+            delayed = true;
             return fault;
+        }
     }
 
-    Addr paddr = e->paddr << PageShift | (vaddr & mask(e->logBytes));
+    if (fault != NoFault)
+        return fault;
```
powerjg, line 428: could `e` leak if this is hit? I.e., you don't `delete e`

src/arch/riscv/tlb.cc
```diff
@@ -310,56 +340,137 @@ TLB::doTranslate(const RequestPtr &req, ThreadContext *tc,
...
+    }
+
+    // if we want to write and it isn't writable, do a page table walk
+    // again to update the dirty flag.
+    if (e && (mode == BaseMMU::Write) && !e->pte.w) {
+        DPRINTF(TLB, "Dirty bit not set, repeating PT walk\n");
+        fault = walker->start(tc, translation, req, mode);
+        if (translation != nullptr || fault != NoFault) {
+            delayed = true;
             return fault;
```
powerjg, line 421: could `e` leak if this is hit? I.e., you don't `delete e`

util/cpt_upgraders/riscv-hext.py
powerjg, line 1: This file needs a copyright statement.

src/arch/riscv/pagetable_walker.cc
```diff
@@ -189,10 +197,44 @@ Walker::WalkerState::initState(ThreadContext * _tc,
...
+
+    // If functional entry.vaddr will be set
+    // in start functional (req == NULL)
+    entry.vaddr = functional ?
+        (Addr)0 :
+        req->getVaddr();
+
+    entry.asid = satp.asid;
+    if (debug_condition()) {
+        DPRINTF(PageTableWalker, "======================================\n");
```
powerjg: Is this absolutely necessary? I think prints like this end up spamming the output.

src/arch/riscv/pagetable_walker.cc
```diff
@@ -670,10 +1071,29 @@ Walker::WalkerState::retry()
...
+            "[%s] Raising page fault. vaddr: %#x gvaddr: %#x\n",
+            curstage == FIRST_STAGE ? "1stSTAGE" : "GSTAGE",
+            entry.vaddr, gresult.vaddr);
+    }
+    return walker->tlb->createPagefault(entry.vaddr, mode,
+                        gresult.vaddr, curstage == GSTAGE, memaccess.virt);
+}
+
+bool
+Walker::WalkerState::debug_condition() {
```
powerjg: This should be removed. If you want to print things in separately from the current debug flags, add a new one. You can also add compound debug flags. 

I can understand wanting to do this for a subset of addresses, but I don't think a one-off implementation in the ptw is the right way to do this.

src/arch/riscv/pagetable_walker.cc
```diff
@@ -242,91 +284,355 @@ Walker::startWalkWrapper()
...
+        read = nullptr;
+    }
+
+    Addr pte_addr = setupWalk(guest_paddr);
+    read = createReqPacket(pte_addr, MemCmd::ReadReq, sizeof(PTESv39));
+    glevel = SV39_LEVELS - 1;
+
+    // TODO THE TIMING PATH IS UNTESTED
+    if (timing)
+    {
```
powerjg: The style is not right in this block of code. E.g.,

```suggestion
    if (timing) {
```

src/arch/riscv/pagetable_walker.cc
```diff
@@ -242,91 +284,355 @@ Walker::startWalkWrapper()
...
+    // it takes care to do the right thing
+    // when functional is true
+    return walk();
 }
 
 Fault
 Walker::WalkerState::stepWalk(PacketPtr &write)
 {
-    assert(state != Ready && state != Waiting);
+    int &curlevel = curstage == GSTAGE ? glevel : level;
```
powerjg: Why is this a reference?

src/arch/riscv/pagetable_walker.cc
```diff
@@ -399,21 +715,55 @@ Walker::WalkerState::stepWalk(PacketPtr &write)
...
-                        // to update the dirty flag.
-                        if (!pte.d && mode != BaseMMU::Write)
-                            entry.pte.w = 0;
-                        doTLBInsert = true;
+                        // TLB inserts are OK for single stage walks
+                        // For two-stage, FIRST_STAGE will reach here just once
+                        // (GStage reaches here multiple times)
+                        if (walkType == OneStage || walkType == GstageOnly ||
+                            (walkType == TwoStage && curstage == FIRST_STAGE))
+                        {
```
powerjg: Style comment: This is *way* too deep of nesting. I suggest adding a function.

src/arch/riscv/pagetable_walker.cc
```diff
@@ -485,8 +857,10 @@ Walker::WalkerState::stepWalk(PacketPtr &write)
         read = new Packet(request, MemCmd::ReadReq);
         read->allocate();
 
-        DPRINTF(PageTableWalker,
-                "Loading level%d PTE from %#x\n", level, nextRead);
+        if (debug_condition()) DPRINTF(PageTableWalker,
+                "[%s] Get level%d PTE from %#x\n",
+                curstage == GSTAGE ? "GSTAGE" : "1stSTAGE",
+                curlevel, nextRead);
```
powerjg: this really needs curly braces. it's quite confusing to read and could lead to bugs.

powerjg (changes requested):
It would be great if you could add the tests that you used (e.g., comparing this to spike) to the gem5 test suite. 

Also, we need to update the RELEASE_NOTES and make it very, very clear that this only works for atomic mode and not for any timing CPU. (Not just O3, but also it doesn't work for timing or minor, right?)

Most of my comments below are about style. The page table walker file needs a lot of improvement to come up to our style standards.

Finally, with the staging branch being created next week, unfortunately, I think it's quite unlikely we can get this merged before we do the 24.1 release.

Harshil2107:
I will retake the checkpoints after the changes are approved.

src/arch/riscv/RiscvISA.py
Navyera: Added some additional explanation

src/arch/riscv/isa.cc
Navyera: imo it helps to have the invariance: MISCREG = PHYSREG, and the previous change breaks this.

The fact that OutOfOrder execution needs to buffer intermediate values before actually writing them to the physical register, does not mean that an actual implementation would necessarily add additional physical registers.

Hence, I designated these extra buffer registers as HELPER.

src/arch/riscv/isa.cc
Navyera: See above reply

src/arch/riscv/isa.cc
Navyera: yes, spike code does this as well

src/arch/riscv/isa.cc
Navyera: Renamed to virtualizationEnabled

src/arch/riscv/memflags.hh
Navyera: done

src/arch/riscv/mmu.hh
Navyera: correct

src/arch/riscv/tlb.cc
Navyera: added explanation in the code. special_access = bypass the TLB and don't insert

src/arch/riscv/tlb.cc
Navyera: Nice catch, added the delete

src/arch/riscv/tlb.cc
Navyera: Nice catch, added the delete

src/arch/riscv/pagetable_walker.cc
Navyera: removed this completely, as well as all usages.

src/arch/riscv/pagetable_walker.cc
Navyera: To change the actual object members when we write : curlevel--;

src/arch/riscv/pagetable_walker.cc
Navyera: Completely agree, but I did not want to extract out the logic, especially since all the previous steps are numbered inline with comments.

Still, we can refactor this if it is a problem.

src/arch/riscv/pagetable_walker.cc
Navyera: was removed anyway

Navyera:
Hi @powerjg , I just rebased on top of current develop with changes addressing your comments. 
By the way, let me know if you prefer an alternative workflow. I am most familiar and confident with rebasing, but I can do it some other way, if there are benefits to alternative approaches.

@Harshil2107 could you please retake the checkpoints, so that the restore tests stop complaining?

src/arch/riscv/pagetable_walker.cc
powerjg: I think this syntax is confusing. What I remind my students often is that the purpose of code isn't for humans to communicate to computers, but for humans to communicate to other humans. The *intention* of this code is not clear to future humans who will come back and look at this code and try to understand it. 

Another way to look at it is that compilers optimize human-focused code so computers can understand it. Thus, the code you feed to the compiler should be optimized for readability, not to reduce the total number of characters.

src/arch/riscv/RiscvISA.py
```diff
@@ -107,7 +108,11 @@ class RiscvISA(BaseISA):
         ELEN in Ch. 2 of RISC-V vector spec",
     )
     privilege_mode_set = Param.PrivilegeModeSet(
-        "MSU",
+        "MSU",  # set MHSU to enable hypervisor (H-extension)
+        # All timing CPUs are NOT supported in MHSU currently
```
powerjg:
```suggestion
        # No timing CPUs are supported in MHSU currently
        # *ONLY THE ATOMIC CPU / ATOMIC MEMORY* is supported
```

Let me know if that is a misinterpretation. The comment needs to be a bit more explicit than it is now.

src/arch/riscv/faults.cc
```diff
@@ -137,6 +165,71 @@ RiscvFault::invoke(ThreadContext *tc, const StaticInstPtr &inst)
...
+                        _code = INT_TIMER_SUPER;
+                        break;
+                    case INT_EXT_VIRTUAL_SUPER:
+                        _code = INT_EXT_SUPER;
+                        break;
+                    default:
+                        break;
+                }
+            }
+            else {
```
powerjg:
```suggestion
            } else {
```

src/arch/riscv/faults.cc
```diff
@@ -137,6 +165,71 @@ RiscvFault::invoke(ThreadContext *tc, const StaticInstPtr &inst)
...
+                    hstatus.gva = mustSetGva();
+                    // Paragraph 8.5.2 RISCV Privileged Spec 20211203
+                    if (isGuestPageFault()) {
+                        tc->setMiscReg(MISCREG_HTVAL, trap_value2());
+                    }
+                }
+                // Write changes to hstatus
+                tc->setMiscReg(MISCREG_HSTATUS, hstatus);
+            }
+            else if (prv == PRV_S &&
```
powerjg:
```suggestion
            } else if (prv == PRV_S &&
```

src/arch/riscv/faults.cc
```diff
@@ -137,6 +165,71 @@ RiscvFault::invoke(ThreadContext *tc, const StaticInstPtr &inst)
...
+                status.mpv = pv;
+                status.gva = mustSetGva();
+                // Paragraph 8.5.2 RISCV Privileged Spec 20211203
+                if (isGuestPageFault()) {
+                    tc->setMiscReg(MISCREG_MTVAL2, trap_value() >> 2);
+                }
+                // Going to M-mode for handling, disable V if it's on
+                if (virtualizationEnabled(tc)) { resetV(tc); }
+            }
+            else if (prv == PRV_S &&
```
powerjg:
```suggestion
            } else if (prv == PRV_S &&
```

src/arch/riscv/interrupts.cc
```diff
@@ -215,20 +286,22 @@ void
...
     unsigned long ip_ulong;
+    unsigned long hvip_ulong;
     unsigned long ie_ulong;
-    UNSERIALIZE_SCALAR(ip_ulong);
-    ip = ip_ulong;
-    UNSERIALIZE_SCALAR(ie_ulong);
-    ie = ie_ulong;
+    UNSERIALIZE_SCALAR(ip_ulong);   ip = ip_ulong;
+    UNSERIALIZE_SCALAR(hvip_ulong); hvip = hvip_ulong;
+    UNSERIALIZE_SCALAR(ie_ulong);   ie = ie_ulong;
```
powerjg:
```suggestion
    UNSERIALIZE_SCALAR(ip_ulong);   
    ip = ip_ulong;
    UNSERIALIZE_SCALAR(hvip_ulong); 
    hvip = hvip_ulong;
    UNSERIALIZE_SCALAR(ie_ulong);   
    ie = ie_ulong;
```

src/arch/riscv/isa.cc
powerjg: I don't know if I understand the need to change this variable name. I see your point, but it's changing the *meaning* of something that's used in other places in the code.

If you are strongly for keeping this change, then it's necessary to (1) add comments explaining the change in the code and (2) put this change in a separate commit so that the blame/log explains the change in semantics.

I know this seems like a lot to ask for a little thing, but I am thinking about the future when people come back to this code and wonder "what does "helper" mean, and why is this not MISCREG"

src/arch/riscv/isa.cc
```diff
@@ -565,20 +604,22 @@ ISA::readMiscReg(RegIndex idx)
...
-            setMiscRegNoEffect(idx, status);
+            if ((virtualizationEnabled() && idx == MISCREG_VSSTATUS) ||
+                         (idx == MISCREG_STATUS)) {
+                RegVal bits_of_interest = STATUS_MPP_MASK |
+                                          STATUS_MPRV_MASK |
+                                          STATUS_MXR_MASK |
+                                          STATUS_SUM_MASK;
+
+                // If at least one of these changed, flush
+                if ((oldstatus ^ status) & bits_of_interest) {
```
powerjg, line 623: If each of these is a boolean variable, can you change this to use boolean logic instead of bitwise logic? (e.g., `&&` instead of `&`)

src/arch/riscv/isa.cc
```diff
@@ -985,6 +1018,250 @@ ISA::resetThread()
...
+{
+    if (csr == CSR_SATP) {
+        STATUS status = readMiscReg(MISCREG_STATUS);
+        if (pm != PRV_M && status.tvm == 1) {
+            return std::make_shared<IllegalInstFault>(
+                    "SATP access with TVM enabled\n",
+                    machInst);
+        }
+    }
+    else if (csr == CSR_VSATP) {
```
powerjg:
```suggestion
    } else if (csr == CSR_VSATP) {
```

src/arch/riscv/isa.cc
```diff
@@ -985,6 +1018,250 @@ ISA::resetThread()
...
+        if (virtualizationEnabled()) {
+        HSTATUS hstatus = readMiscReg(MISCREG_HSTATUS);
+        if (hstatus.vtvm == 1)
+            return std::make_shared<VirtualInstFault>(
+                "VSATP access with hstatus.vtvm enabled",
+                machInst);
+        }
+    }
+    else if (csr == CSR_HGATP) {
```
powerjg:
```suggestion
    } else if (csr == CSR_HGATP) {
```

src/arch/riscv/isa.cc
```diff
@@ -985,6 +1018,250 @@ ISA::resetThread()
...
+                    machInst);
+        }
+    }
+
+
+    return NoFault;
+}
+
+RegVal
+ISA::backdoorReadCSRAllBits(uint64_t csr) {
```
powerjg:
```suggestion
RegVal
ISA::backdoorReadCSRAllBits(uint64_t csr) 
{
```

src/arch/riscv/isa.cc
```diff
@@ -995,6 +1272,93 @@ ISA::getFaultHandlerAddr(RegIndex idx, uint64_t cause, bool intr) const
...
+void resetV(ThreadContext *tc) {
+    assert(virtualizationEnabled(tc));
+    tc->setMiscReg(MISCREG_VIRT, 0);
+}
+
+// FPU status update function
+Fault updateFPUStatus(ExecContext *xc, ExtMachInst machInst, bool set_dirty) {
+
+
+    MISA misa = xc->readMiscReg(MISCREG_ISA);
```
powerjg:
```suggestion
Fault 
updateFPUStatus(ExecContext *xc, ExtMachInst machInst, bool set_dirty) 
{
    MISA misa = xc->readMiscReg(MISCREG_ISA);
```

src/arch/riscv/isa.cc
```diff
@@ -995,6 +1272,93 @@ ISA::getFaultHandlerAddr(RegIndex idx, uint64_t cause, bool intr) const
...
+            xc->setMiscReg(MISCREG_VSSTATUS, vsstatus);
+        }
+    }
+
+    return NoFault;
+}
+
+// VPU status update function
+Fault updateVPUStatus(
+    ExecContext *xc, ExtMachInst machInst, bool set_dirty, bool check_vill) {
```
powerjg: Style

src/arch/riscv/isa.cc
```diff
@@ -995,6 +1272,93 @@ ISA::getFaultHandlerAddr(RegIndex idx, uint64_t cause, bool intr) const
...
+    MISA misa = xc->readMiscReg(MISCREG_ISA);
+    STATUS status = xc->readMiscReg(MISCREG_STATUS);
+    STATUS vsstatus = misa.rvh && virtualizationEnabled(xc) ?
+        xc->readMiscReg(MISCREG_VSSTATUS) : 0;
+
+    if (!misa.rvv || status.vs == VPUStatus::OFF ||
+        (misa.rvh && virtualizationEnabled(xc) &&
+         vsstatus.vs == VPUStatus::OFF))
+        return std::make_shared<IllegalInstFault>(
+            "RVV is disabled or VPU is off", machInst);
```
powerjg: Must have curly braces since the inner block is multiple lines. However, IMO all `if` statements should always have curly braces. Again, I have spent many hours tracking down bugs because someone tried to save 2 characters. It ends up being not worth it in the end :)

src/arch/riscv/isa/formats/standard.isa
powerjg: How can you be sure of this since this line changes from `setMiscReg` to `writeCSR`?

src/arch/riscv/tlb.cc
```diff
@@ -248,50 +249,79 @@ TLB::remove(size_t idx)
...
+        code = gpf ? ExceptionCode::STORE_GUEST_PAGE :
+                     ExceptionCode::STORE_PAGE;
     else
-        code = ExceptionCode::INST_PAGE;
-    return std::make_shared<AddressFault>(vaddr, code);
+        code = gpf ? ExceptionCode::INST_GUEST_PAGE :
+                     ExceptionCode::INST_PAGE;
+
+    if (gpf) { assert(virt); }
+    return std::make_shared<AddressFault>(vaddr, code, gvaddr, virt);
```
powerjg: Curly braces!

src/arch/riscv/tlb.cc
```diff
@@ -400,6 +536,20 @@ TLB::translate(const RequestPtr &req, ThreadContext *tc,
             }
         }
 
+        // according to the RISC-V tests, negative physical addresses trigger
+        // an illegal address exception.
+        // TODO where is that written in the manual?
```
powerjg, line 537: Do you want to fill in this "todo" before committing?

util/cpt_upgraders/riscv-hext.py
```diff
@@ -0,0 +1,106 @@
...
+            re.search(r".*processor.*\.core.*\.isa$", sec)
+            and cpt.get(sec, "isaName") == "riscv"
+        ):
+            # Updating RVH misc registers (dummy values)
+            mr = cpt.get(sec, "miscRegFile").split()
+            if len(mr) == 187:
+                print(
+                    "MISCREG_* RVH registers already seem " "to be inserted."
+                )
+            else:
```
powerjg, line 59: Do we need to be worried that we change the name from MISCREG_FFLAGS to HELPER_...?

src/arch/riscv/pagetable_walker.cc
```diff
@@ -242,91 +278,284 @@ Walker::startWalkWrapper()
...
+    {
+        MISA misa = tc->readMiscReg(MISCREG_ISA);
+        panic_if(misa.rvh, "Timing walks are not supported with h extension");
+        nextState = state;
+        state = Waiting;
+        timingFault = NoFault;
+        // DO GSTAGE HERE IF NEEDED AND THEN DO PACKETS FOR *PTE
+        sendPackets();
+    }
+    else
```
powerjg: Is it possible to split this out into another function. The `startWalk` function is now *very* complicated. (And it's renamed? Unfortunately, I no longer understand what the walker's logical algorithm is.)

src/arch/riscv/pagetable_walker.cc
powerjg: I think this does need be refactored, along with other parts of this file. The problem is that if someone new comes to the code (or someone like me who is coming back to this code and is only familiar with the old code), they can't understand what the logical progression is. Since things are done in very long functions with deeply nested if/while statements, it is very hard for humans to understand the code. And, that's what this should be all about. Helping future humans understand the code. Also, I think that refactoring this code will be absolutely required when implementing timing accesses anyway.

src/arch/riscv/pagetable_walker.cc
powerjg: Also, this should be `cur_level` since it is a local variable. Also, from reading the code more closely in the rest of my review, it is even more confusing that we are using a reference.

src/arch/riscv/isa/templates/vector_arith.isa
powerjg, line 1: The changes in these two files make me feel uncomfortable because we don't have any tests for them. There could be some subtle differences in the code blocks, but it looks like all of the changes were just copy-paste. How did you test that these changes have no effects on vector code?

powerjg (commented):
It would be good if we could produce a `fatal` at initialization if we try to create a system with a timing-based CPU when enabling H-mode. However, it may be that we can only do it at runtime. Either way, the user should be informed with a very clear message that only atomic is supported when using the h-extension.

Lots of little things below. I think the biggest comment is that the changes to the page table walker are so complex that they really need to be refactored so future people can understand them (and to prepare them to be used in timing mode). Otherwise, there are lots of little style things (which, tbh, I didn't comment on every one of them). 

@rogerchang23424 I resolved some of your comments where it was clear that the authors made the updates, but it would be good if you could re-review. Also, please feel free to disagree with my comments. You're more of an expert on this code than I am.

Sorry @Navyera for taking so long to get back to this. My excuses are (1) This is a *HUGE* amount of code to review, so it takes having a large chunk of time to concentrate. (2) The release. (3) The holidays. Thanks for pinging me on it, and I'll do my best to help you get things merged as soon as we can.

rogerchang23424:
I will review again once the PR is on the ToT

src/arch/riscv/isa.cc
Navyera: The values are whole registers in this case so the bitwise operation makes sense (if I am not mistaken)

src/arch/riscv/isa/templates/vector_arith.isa
Navyera: Well, since there are no tests as you said I would consider the change untested indeed. We are not familiar with the vector arithmetic stuff so I do not think we can adequately test this code ourselves. 

The change was done to mirror the changes for fpuStatus (updateVPUStatus is similar to the pattern for updateFPUStatus). We simply saw a chance to avoid some duplication and improve readability. We can undo the change if it is deemed unsafe.

Navyera:
Hi @rogerchang23424  @powerjg , I just rebased on top of the current develop.

I have hopefully addressed most of the style comments brought up by @powerjg and also attempted a refactor of the PTW logic to clearly separate GStage logic and not overcomplicate the old code (where possible). We are open to comments on the refactor.

src/arch/riscv/isa/formats/standard.isa
powerjg: To test this, you can run something like Linux boot (riscv-ubuntu-boot workload) on top of the "base" develop and with your changes and compare the read/write miscregs numbers in the stats.

src/arch/riscv/interrupts.cc
```diff
@@ -72,17 +73,48 @@ Interrupts::globalMask() const
     PrivilegeMode prv = (PrivilegeMode)tc->readMiscReg(MISCREG_PRV);
     switch (prv) {
         case PRV_U:
+            if (misa.rvh && virtualizationEnabled(tc)) {
+                STATUS vsstatus = tc->readMiscReg(MISCREG_VSSTATUS);
+                INTERRUPT hideleg = tc->readMiscReg(MISCREG_HIDELEG);
+
+                mask.local = ~hideleg.local | ~mideleg.local;
+                if (status.uie)
+                    mask.local = mask.local | (hideleg.local & mideleg.local);
```
powerjg: Use curly braces

src/arch/riscv/interrupts.cc
```diff
@@ -91,24 +123,61 @@ Interrupts::globalMask() const
...
-            if (status.sie) {
-                mask.sei = mask.sti = mask.ssi = 1;
-                mask.local = mask.local | mideleg.local;
+            if (misa.rvh && virtualizationEnabled(tc)) {
+                STATUS vsstatus = tc->readMiscReg(MISCREG_VSSTATUS);
+                INTERRUPT hideleg = tc->readMiscReg(MISCREG_HIDELEG);
+
+                mask.local = ~hideleg.local | ~mideleg.local;
+                if (status.sie)
+                    mask.local = mask.local | (hideleg.local & mideleg.local);
```
powerjg: Curly brace

src/arch/riscv/isa.cc
```diff
@@ -804,9 +810,15 @@ ISA::setMiscReg(RegIndex idx, RegVal val)
                 if (new_val.mode != AddrXlateMode::BARE &&
                     new_val.mode != AddrXlateMode::SV39)
                     new_val.mode = cur_val.mode;
+
+                // Flush TLB if satp changed as Spike does
+                if (new_val != cur_val) {
```
powerjg: Let's look a little deeper here. If the current implementation is correct (i.e., if it functionally works without flushing the TLB), then we can probably elide this. It's also possible that the original gem5 code was wrong and we could have read stale translations.

src/arch/riscv/isa.cc
```diff
@@ -827,6 +839,20 @@ ISA::setMiscReg(RegIndex idx, RegVal val)
                 setMiscRegNoEffect(idx, val & ~wpri_mask);
             }
             break;
+          case MISCREG_HGATP:
+            {
+                tc->getMMUPtr()->flushAll();
+                SATP cur_val = readMiscRegNoEffect(idx);
+                SATP new_val = val;
+
+                if (new_val.mode != AddrXlateMode::BARE &&
```
powerjg, line 891: Curly braces, please :)

powerjg (commented):
Let's do one more quick round. Double check all of the comments, rebase (again, I'm sorry 😢) and then Roger can take one final pass.

erin-le:
@Navyera: Hello, thank you for your contribution! Could you respond to the new set of comments when you have time?

erin-le:
@Navyera Hello, just checking in again. Could you give an update on the status of this PR on your end? Thank you!

Navyera:
Hi @erin-le, I am sorry for the late reply on this. We are currently rebasing and re-testing everything on top of the current develop branch. Hoping to sync with the reviewers so once I push the rebased commits the final review pass can take place.

We had paused work on this for a while, but we're now actively working to get it ready for merge as soon as possible.

Navyera:
@powerjg @erin-le @rogerchang23424 
Just completed and pushed the rebase. Hoping we can finalize things and put this on the staging branch.

src/arch/riscv/isa/formats/standard.isa
Navyera: @rogerchang23424  was totally right on this one, we changed readCSR/writeCSR to use xc->read/write so that miscReg  stats are updated

powerjg:
@erin-le can you create a place to track things that we want on the staging branch? This should be one of those things :)

BobbyRBruce:
@Navyera: Are you still working on this? We can't merge it with these tests failing

Navyera:
Hi @BobbyRBruce , the failing tests are due to checkpoints that must be updated with the cpt-upgrader we had provided.
There was a discussion about this starting [here](https://github.com/gem5/gem5/pull/1387#issuecomment-2402509187) and continuing in the following comments, but it never got resolved.

Harshil2107:
I have updated the checkpoint using the new  cpt upgrader in this PR. 

Here is the link to the updated checkpoint: https://resources.gem5.org/resources/riscv-hello-example-checkpoint?version=6.0.0

I have also pushed a commit to update the tests to use this new checkpoint.

erin-le:
@Navyera: Hello, it unfortunately looks like this PR is causing the [SST tests](https://github.com/gem5/gem5/actions/runs/15308939906/job/43068592271) to fail. The specific error message is as follows:
```
build/RISCV/arch/riscv/linux/fs_workload.cc:53: info: Loading DTB file: m5out/device.dtb at address 0x87e00000
gem5Component-gem5_node-> init phase: 1
gem5Component-gem5_node-> init phase: 2
gem5Component-gem5_node-> init phase: 3
gem5Component-gem5_node-> init phase: 4
gem5Component-gem5_node-> init phase: 5
gem5Component-gem5_node->Component is being setup.
gem5.cc:295: info: Entering event queue @ 0.  Starting simulation...
sst: build/RISCV/arch/riscv/tlb.cc:335: gem5::Addr gem5::RiscvISA::TLB::hiddenTranslateWithTLB(gem5::Addr, uint16_t, gem5::Addr, gem5::BaseMMU::Mode): Assertion `e != nullptr' failed.
Program aborted at tick 327201500305
--- BEGIN LIBC BACKTRACE ---
```
To run the SST tests locally, I did the following:
1. Ran `docker run -u $UID:$GID --volume ./:/gem5 --rm -it ghcr.io/gem5/sst-env:latest` in my gem5 directory
2. Ran `scons build/RISCV/libgem5_opt.so --without-tcmalloc --duplicate-sources --ignore-style -j $(nproc)` in the `/gem5` directory in the Docker container
3. In the `/gem5/ext/sst` directory in the Docker container, ran `mv Makefile.linux Makefile`, `make -j $(nproc)`, and `sst --add-lib-path=./ sst/example.py`

The specific commands run by the SST tests can also be found in `gem5/.github/workflows/daily-tests.yaml`. Could you take a look and see if you can fix this failure? Thank you!

powerjg:
Note that the normal RISC-V boot also fails. The following command has the same error:

```sh
build/ALL/gem5.opt configs/example/gem5_library/riscv-fs.py
```

@Navyera We are aiming to make a release in ~1.5 weeks. We need to resolve this bug ASAP.

powerjg:
FYI, the problem occurs for load-reserved accesses. Why do we not lookup the TLB for load-reserved? I'm testing now to see if it works if I remove `memaccess.lr` from the check for `special_access` in `doTranslate`

powerjg:
I believe https://github.com/gem5/gem5/pull/2331 fixes this issue. @Navyera I would appreciate it if you could review.

Navyera:
@powerjg Thanks for identifying and fixing this so quickly!

We had not tested the boot with timing CPUs. The code expects to find that address in the TLB, but since we forced a TLB bypass the translation is missing, hence the error.

The reason we had modified the code in that way (forced the TLB bypass for LR) is the following code from the Spike ISA simulator:

```C
struct xlate_flags_t {
  const bool forced_virt : 1 {false};
  const bool hlvx : 1 {false};
  const bool lr : 1 {false};
  const bool ss_access : 1 {false};
  const bool clean_inval : 1 {false};

  bool is_special_access() const {
    return forced_virt || hlvx || lr || ss_access || clean_inval;
  }
};
...

...
    if (likely(!xlate_flags.is_special_access() && aligned && tlb_hit)) {
      res = *(target_endian<T>*)(tlb_data[vpn % TLB_ENTRIES].host_offset + addr);
    } else {
      load_slow_path(addr, sizeof(T), (uint8_t*)&res, xlate_flags);
    }
...

```

It is not yet clear to me why and in which cases LR would need a TLB bypass and I am not really familiar with the LR/SC pattern and its details in RISCV. We simply "trusted" Spike on this one.

We can look into the details of this further, but I think your fix addresses this for now.

powerjg:
@Navyera FYI, there's still a bug. Running a simple RISC-V boot sometimes hangs. If you run the gem5 script `configs/example/gem5_library/riscvmatched-fs.py --to-init` it never exits with this PR. Before this PR, it should exit after booting Linux.

I'm not sure what's wrong. I'm guessing it might have to do with the pseudo ops, but I don't have much evidence for that.

Note that booting ubuntu 24.04 seems to work as expected.

If we can't resolve this issue before the release next week I'm afraid we're going to have to revert this PR.

powerjg:
Looks like the problem is MinorCPU+20.04, btw. I'll test with MinorCPU+24.04 now.

Navyera:
Hi @powerjg , confirming the behaviour you're describing, we're also looking into possible causes/fixes. Did MinorCPU+24.04 work? I have only tested with the 20.04 image so far. That boot takes quite a while unfortunately...

powerjg:
> Did MinorCPU+24.04 work? 

Not with this PR applied. Without this PR, it works.

> That boot takes quite a while unfortunately...

Minor+20.04 only takes less than 10 minutes on my M4 Macbook.

Navyera:
@powerjg  I think we got it. #2355 should fix the problem. Let me know if the change works on your end as well.

P.S. I am really tempted to get an M4 MacBook 😄 The i9-14900K system I am using does that boot in ~24mins !!!

PR #1388, opened 2024-07-26
arch-arm: Add support for AArch32 PMEVCNTR*/PMEVTYPER*/PMCCFILTR

arichardson:
These registers were only handled in AArch64 mode but are also accessible as a c14 registers for AArch32.

Change-Id: I62fe54427e96265df0589308afa1b5d665dbf210


giactra (approved):
LGTM. The trapping of the register access is not implemented, but on the other hand, neither the AArch64 version is

PR #1389, opened 2024-07-26
arch: Dump semihosting write buffer in debug output

arichardson:
This makes it easier to debug unexpected semihosting outputs (in my case a wrong buffer argument was being passed).

Change-Id: I342610a92fb8efe121d030f7b9ea3307efc4fec3

### DIFF
--- a/src/arch/generic/semihosting.cc
+++ b/src/arch/generic/semihosting.cc
@@ -250,6 +250,7 @@ BaseSemihosting::callWrite0(ThreadContext *tc, InPlaceArg arg)
     PortProxy &proxy = portProxy(tc);
     std::string str;
     proxy.readString(str, arg.addr);
+    DDUMP(Semihosting, str.data(), str.size());
     std::cout.write(str.c_str(), str.size());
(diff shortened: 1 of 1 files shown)


giactra:
This looks like a good idea to me. Just to nitpick (as usual), we have specific tags for the commit/PR message.
We don't use "arch-generic", but we have "arch" instead. The list of tags is covered by: https://github.com/gem5/gem5/blob/stable/MAINTAINERS.yaml#L43
Could you amend the commit/PR title plz?

arichardson:
> This looks like a good idea to me. Just to nitpick (as usual), we have specific tags for the commit/PR message. We don't use "arch-generic", but we have "arch" instead. The list of tags is covered by: https://github.com/gem5/gem5/blob/stable/MAINTAINERS.yaml#L43 Could you amend the commit/PR title plz?

Done, surprised this wasn't caught by the pre-commit checks. Will make sure to pick the right tags in the future.

giactra (approved):

PR #1395, opened 2024-07-29
misc: Remove GCN3 from maintainers

abmerop:
These files no longer exist. This was missed when GCN3 support was removed in the last release.

Change-Id: I5d2aab1952c37e64da5c362a6201aa6750531c1b


powerjg (approved):

mattsinc (approved):
There are other changes we should make to make sure both of us are listed as maintainers for the GPU pieces beyond this, but I'll try to do that in a later commit.

PR #1401, opened 2024-07-31
arch-riscv: adding support for uncacheable address ranges in SE mode

iliasslasri:
This PR implements the support for uncacheable ranges in the riscv arch
in SE mode.
In the RISCV MMU, we changed the BasePMAChecker to PMAChecker which
inherents from BasePMAChecker, this class has an attribute
uncacheable_range that we can init in the python script, and implement
functions like ```isUncaheable```that we used to condition the
turning of the flag *UNCHACHEABLE* of the request when translating
the address.

In the mmu, pagetable_walker and tlb, we change the attribute type
BasePMAChcker to PMAChecker (so as in The wrappers RiscvMMU.py and
RiscvTLB.py), then we use this attribute (i.e, pma) to call the function
isUncacheable and condition the setting of the flag
```Request::UNCACHEBLE```.

Change-Id: I68f6ed6f15fda8ca5878e66b581863ec1abcd79a

Signed-off-by: Iliass Lasri <lasri.iliass@huawei.com>
Signed-off-by: Marcelo Ruaro <marcelo.ruaro@huawei.com>

### CODE REVIEW

File: src/arch/riscv/tlb.cc [hunk 1]
```diff
@@ -422,6 +428,14 @@ TLB::translate(const RequestPtr &req, ThreadContext *tc,
...
             return fault;
 
+        // look if the physical address is uncacheable
+        // and set the flag accordingly
+        Addr paddr = req->getPaddr();
+        if(pma->isUncacheable(paddr, req->getSize())){
+            req->setFlags(Request::UNCACHEABLE);
+            DPRINTF(TLB, "Access to address %#x turned to uncacheable.", paddr);
+        }
+
```
rogerchang23424: I found there some common code section to add the `Request::UNCACHEABLE` flags to Request Packet. 

I think we can have the abstract method `checkSE` in `BasePMAChecker` and implement it in the `PMAChecker::checkSE` like

```cpp
void
PMAChecker::checkSE(const RequestPtr &req)
{
    if (isUncacheable(req->getPaddr(), req->getSize())) {
            req->setFlags(Request::UNCACHEABLE);
            DPRINTF(TLB, "Access to address %#x turned to uncacheable.", req->getPaddr());
    }
}
```

Ref: https://gem5-review.googlesource.com/c/public/gem5/+/40596

It won't require to change the `BasePMAChecker` to `PMAChecker`. How do you think?

ivanaamit:
Hi @iliasslasri,

```
f=.git/hooks/commit-msg
mkdir -p $f
chmod +x $f
```

hnpl (commented):
Thanks for the contribution.

I saw that the pma_checker parameter is now of type `PMAChecker` instead of its base class. Is there a particular reason for this change?

iliasslasri:
Hi @hnpl,
Indeed, I changed the type of the pma_checker from ```BasePMAChecker``` to ```PMAChecker```, to use the extra attribute ```AddrRangeList uncacheable;```, to add the support of uncacheable ranges without adding attributes or function to any class. The PMAChecker inherits from BasePMAChecker, so there will be no compatibility issue.

src/arch/riscv/tlb.cc
rogerchang23424: @iliasslasri,

Please reference the pervious PR https://github.com/gem5/gem5/pull/846, I feel that this PR revert the BasePMAChecker changes.

src/arch/riscv/RiscvMMU.py
```diff
@@ -49,7 +49,7 @@ class RiscvMMU(BaseMMU):
 
     itb = RiscvTLB(entry_type="instruction")
     dtb = RiscvTLB(entry_type="data")
-    pma_checker = Param.BasePMAChecker(PMAChecker(), "PMA Checker")
+    pma_checker = Param.PMAChecker("PMA Checker")
```
powerjg, line 52: I strongly agree with this change to remove an instance of a SimObject as the default value for a parameter.

However, I worry that there are configuration scripts that depend on this. We need to test the stdlib boards and the config scripts in `configs/examples` before we can merge this change.

In fact, it looks like this is likely why the CI tests are failing. See https://github.com/gem5/gem5/actions/runs/10286529082/job/28476297182?pr=1401

ivanaamit:
Hi @iliasslasri, you need to fix the failing tests in this PR to get it closer to being merged. Thanks.

ivanaamit:
Hi @iliasslasri, please let me know if you're still working on this. If not, I'll close the PR, but you can reopen it anytime if you decide to continue. Thanks!

erin-le:
Closing due to inactivity. This PR can be reopened if someone wants to work on it in the future.

PR #1403, opened 2024-07-31
Fixed implementation of BOP (#1402)

### LINKED ISSUE #1402: The implementation of the Best Offset Prefetcher in Gem5 is not accurate to what the paper describes
The implementation of the Pierre Michaud's Best Offset Prefetcher (BOP) has some inconsistencies with respect to the [paper](http://www.irisa.fr/alf/downloads/michaud/dpc2_michaud.pdf) because of which the performance estimated of BOP are pessimistic. In particular
1. The hash computation which is used to index the RR Table results in a significant number of cache conflicts. This in turns hurts the performance due to the contention at the RR Table.
2. In the tag computation, which is done when an address is stored in the RR Table, the tag is computed incorrectly. In particular, the current version of code uses the following formula to compute the tag: `(addr >> blkSize) & tagMask`. For cache line sizes of 64 bytes, the `blkSize` variable has the value 64. However, if we right shift `addr` by 64, we end up either zeroing it out (if the compiler uses logical right shift), or not changing the value at all (if the compiler uses logical right shift with rotate). Note that the compiler is free to choose between the two types of right shift operations because we end up bit-wise ANDing the result anyways.
This, of course, is not what we want. Instead of `(addr >> blkSize) & tagMask`, the correct formula is `(addr >> lBlkSize) & tagMask` which performs the right shift not by the $block\ size$, but rather by the $log_2(block\ size)$. In our example, this would mean that we right shift by 6 instead of 64 which is the correct value since the BOP operates at cache line granularity (as opposed to word address granularity). 
3. The best offset learning algorithm which is implemented differs from what the paper describes. The pseudo-code for the current implementation is as follows:
```
if all offsets have been visited
	increment the round count
	if its time to choose a new offset
		choose the new offset
		enable prefetching
		reset learning structures
	else if the best score is less than the threshold
		disable prefetching
	endif
endif
```
In the current implementation, BOP will only choose a new offset if all offsets have been visited. However, if the first offset in the list of offsets is indeed the best offset (its score exceeds the prefetching threshold), then BOP should not wait until it tests all offsets to choose the first offset. To fix this issue, the psuedocode needs to be be broken into two independent parts: the first part which checks if all offsets have been visited and updates the data structures, and the second part which performs the best offset selection. If we couple these two distinct pieces together, then BOP would have to wait until all offsets have been evaluated before it can select the new best offset. This, in turn, reduces the reaction time of BOP as it would not be able to react quickly to the changes in workload's memory access patterns. If the workload's access offset changes from $o_1$ to $o_2$, then until all offsets have been evaluated, BOP would not update the best offset from $o_1$ to $o_2$. This would not be a problem for workloads with fairly long intervals of constant offsets, but for workloads which frequently change their access pattern, the current of implementation of BOP would not be able to provide good coverage.

My colleague and I have fixed the code and we want to contribute our implementation to upstream. I'm creating this issue so that I can open a pull request.

Setu-Gupta:
This PR fixes the issues with the implementation of the Best Offset Prefetcher described in issue #1402

On branch bop
Changes to be committed:
	modified:   src/mem/cache/prefetch/bop.cc
	modified:   src/mem/cache/prefetch/bop.hh

### DIFF
--- a/src/mem/cache/prefetch/bop.cc
+++ b/src/mem/cache/prefetch/bop.cc
@@ -1,5 +1,6 @@
 /**
  * Copyright (c) 2018 Metempsy Technology Consulting
+ * Copyright (c) 2024 Samsung Electronics
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -36,7 +37,6 @@ namespace gem5
 
 namespace prefetch
 {
-
 BOP::BOP(const BOPPrefetcherParams &p)
     : Queued(p),
       scoreMax(p.score_max), roundMax(p.round_max),
@@ -47,17 +47,13 @@ BOP::BOP(const BOPPrefetcherParams &p)
       delayTicks(cyclesToTicks(p.delay_queue_cycles)),
       delayQueueEvent([this]{ delayQueueEventWrapper(); }, name()),
       issuePrefetchRequests(false), bestOffset(1), phaseBestOffset(0),
-      bestScore(0), round(0)
+      bestScore(0), round(0), degree(p.degree)
 {
     if (!isPowerOf2(rrEntries)) {
         fatal("%s: number of RR entries is not power of 2\n", name());
     }
     if (!isPowerOf2(blkSize)) {
         fatal("%s: cache line size is not power of 2\n", name());
-    }
-    if (!(p.negative_offsets_enable && (p.offset_list_size % 2 == 0))) {
-        fatal("%s: negative offsets enabled with odd offset list size\n",
-              name());
     }
 
     rrLeft.resize(rrEntries);
@@ -103,7 +99,8 @@ BOP::delayQueueEventWrapper()
             delayQueue.front().processTick <= curTick())
     {
         Addr addr_x = delayQueue.front().baseAddr;
-        insertIntoRR(addr_x, RRWay::Left);
+        Addr addr_tag = tag(addr_x);
+        insertIntoRR(addr_x, addr_tag, RRWay::Left);
         delayQueue.pop_front();
     }
 
@@ -114,22 +111,24 @@ BOP::delayQueueEventWrapper()
 }
 
 unsigned int
-BOP::hash(Addr addr, unsigned int way) const
+BOP::index(Addr addr, unsigned int way) const
 {
-    Addr hash1 = addr >> way;
-    Addr hash2 = hash1 >> floorLog2(rrEntries);
-    return (hash1 ^ hash2) & (Addr)(rrEntries - 1);
+    Addr lrrEntries = floorLog2(rrEntries);
+    Addr lineaddr = addr >> lBlkSize;
+    Addr hash = lineaddr ^ (lineaddr >> (lrrEntries << way));
+    hash &= ((1ULL << lrrEntries) - 1);
+    return hash % rrEntries;
 }
 
 void
-BOP::insertIntoRR(Addr addr, unsigned int way)
+BOP::insertIntoRR(Addr addr, Addr tag, unsigned int way)
 {
     switch (way) {
         case RRWay::Left:
-            rrLeft[hash(addr, RRWay::Left)] = addr;
+            rrLeft[index(addr, RRWay::Left)] = tag;
             break;
         case RRWay::Right:
-            rrRight[hash(addr, RRWay::Right)] = addr;
+            rrRight[index(addr, RRWay::Right)] = tag;
             break;
     }
 }
@@ -163,7 +162,7 @@ BOP::resetScores()
 inline Addr
 BOP::tag(Addr addr) const
 {
-    return (addr >> blkSize) & tagMask;
+    return (addr >> lBlkSize) & tagMask;
 }
 
 bool
@@ -185,14 +184,21 @@ BOP::testRR(Addr addr) const
 }
 
 void
-BOP::bestOffsetLearning(Addr x)
+BOP::bestOffsetLearning(Addr addr)
 {
-    Addr offset_addr = (*offsetsListIterator).first;
-    Addr lookup_addr = x - offset_addr;
+    Addr offset_tag = (*offsetsListIterator).first;
+
+    // Compute the lookup tag for the RR table. addr is a tag value, not an
+    // address. Therefore, subtracting the offset from addr may result in
+    // integer underflow. Therefore, we first convert the tag back to address
+    // by right shifting it, and then subtract the offset. This gives us a
+    // new lookup address which we use to compute the lookup tag
+    Addr lookup_tag = tag((addr << lBlkSize) - (offset_tag << lBlkSize));
 
     // There was a hit in the RR table, increment the score for this offset
-    if (testRR(lookup_addr)) {
-        DPRINTF(HWPrefetch, "Address %#lx found in the RR table\n", x);
+    if (testRR(lookup_tag)) {
+        DPRINTF(HWPrefetch, "Address %#lx found in the RR table\n",
+                lookup_tag);
         (*offsetsListIterator).second++;
         if ((*offsetsListIterator).second > bestScore) {
             bestScore = (*offsetsListIterator).second;
@@ -201,6 +207,7 @@ BOP::bestOffsetLearning(Addr x)
         }
     }
 
+    // Move the offset iterator forward to prepare for the next time
     offsetsListIterator++;
 
     // All the offsets in the list were visited meaning that a learning
@@ -208,20 +215,26 @@ BOP::bestOffsetLearning(Addr x)
     if (offsetsListIterator == offsetsList.end()) {
         offsetsListIterator = offsetsList.begin();
         round++;
+    }
 
-        // Check if the best offset must be updated if:
-        // (1) One of the scores equals SCORE_MAX
-        // (2) The number of rounds equals ROUND_MAX
-        if ((bestScore >= scoreMax) || (round == roundMax)) {
-            bestOffset = phaseBestOffset;
+    // Check if its time to re-calculate the best offset
+    if ((bestScore >= scoreMax) || (round >= roundMax))
+    {
             round = 0;
+            if (bestScore > badScore)
+            {
+                bestOffset = phaseBestOffset;
+                round = 0;
+                bestScore = 0;
+                phaseBestOffset = 0;
+                resetScores();
+                issuePrefetchRequests = true;
+            }
+            else
+                issuePrefetchRequests = false;
+            resetScores();
             bestScore = 0;
             phaseBestOffset = 0;
-            resetScores();
-            issuePrefetchRequests = true;
-        } else if (bestScore <= badScore) {
-            issuePrefetchRequests = false;
-        }
     }
 }
 
@@ -234,21 +247,21 @@ BOP::calculatePrefetch(const PrefetchInfo &pfi,
     Addr tag_x = tag(addr);
 
     if (delayQueueEnabled) {
-        insertIntoDelayQueue(tag_x);
+        insertIntoDelayQueue(addr);
     } else {
-        insertIntoRR(tag_x, RRWay::Left);
+        insertIntoRR(addr, tag_x, RRWay::Left);
     }
 
     // Go through the nth offset and update the score, the best score and the
     // current best offset if a better one is found
-    bestOffsetLearning(tag_x);
+    bestOffsetLearning(addr);
 
-    // This prefetcher is a degree 1 prefetch, so it will only generate one
-    // prefetch at most per access
     if (issuePrefetchRequests) {
-        Addr prefetch_addr = addr + (bestOffset << lBlkSize);
-        addresses.push_back(AddrPriority(prefetch_addr, 0));
-        DPRINTF(HWPrefetch, "Generated prefetch %#lx\n", prefetch_addr);
+        for (int d = 1; d <= degree; d++) {
+            Addr prefetch_addr = addr + ((d * bestOffset) << lBlkSize);
+            addresses.push_back(AddrPriority(prefetch_addr, 0));
+            DPRINTF(HWPrefetch, "Generated prefetch %#lx\n", prefetch_addr);
+        }
     }
 }
 
@@ -260,10 +273,11 @@ BOP::notifyFill(const CacheAccessProbeArg &arg)
     // Only insert into the RR right way if it's the pkt is a HWP
     if (!pkt->cmd.isHWPrefetch()) return;
 
-    Addr tag_y = tag(pkt->getAddr());
+    Addr addr = pkt->getAddr();
+    Addr tag_y = tag(addr);
 
     if (issuePrefetchRequests) {
-        insertIntoRR(tag_y - bestOffset, RRWay::Right);
+        insertIntoRR(addr, tag_y - bestOffset, RRWay::Right);
     }
 }
 
--- a/src/mem/cache/prefetch/bop.cc
+++ b/src/mem/cache/prefetch/bop.cc
@@ -67,8 +67,10 @@ BOP::BOP(const BOPPrefetcherParams &p)
     rrLeft.resize(rrEntries);
     rrRight.resize(rrEntries);
 
-    // Following the paper implementation, a list with the specified number
-    // of offsets which are of the form 2^i * 3^j * 5^k with i,j,k >= 0
+    /*
+     * Following the paper implementation, a list with the specified number
+     * of offsets which are of the form 2^i * 3^j * 5^k with i,j,k >= 0
+     */
     const int factors[] = { 2, 3, 5 };
     unsigned int i = 0;
     int64_t offset_i = 1;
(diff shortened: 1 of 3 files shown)


ivanaamit:
Hi @Setu-Gupta,

```
f=.git/hooks/commit-msg
mkdir -p $f
chmod +x $f
```

Additionally, we have a specific style that every commit message must follow. You can find the guidelines here: https://www.gem5.org/contributing (search for "Committing").

Setu-Gupta:
Hi @ivanaamit 

Thanks for letting me know to fix the PR. I've amended the commit message. I've also added the information of the co-author (@AbhishekUoR) of this commit.

Thanks

src/mem/cache/prefetch/bop.cc
```diff
@@ -47,18 +47,14 @@ BOP::BOP(const BOPPrefetcherParams &p)
...
     if (!isPowerOf2(rrEntries)) {
         fatal("%s: number of RR entries is not power of 2\n", name());
     }
     if (!isPowerOf2(blkSize)) {
         fatal("%s: cache line size is not power of 2\n", name());
     }
-    if (!(p.negative_offsets_enable && (p.offset_list_size % 2 == 0))) {
-        fatal("%s: negative offsets enabled with odd offset list size\n",
-              name());
-    }
```
odanrc, line 66: Add fatal_if(degree <= 0, ...)

src/mem/cache/prefetch/bop.hh
```diff
@@ -151,6 +153,7 @@ class BOP : public Queued
         void notifyFill(const CacheAccessProbeArg &arg) override;
 
     public:
+    	unsigned int degree;
```
odanrc: Add doxygen documentation

src/mem/cache/prefetch/bop.cc
```diff
@@ -114,22 +111,24 @@ BOP::delayQueueEventWrapper()
 }
 
 unsigned int
-BOP::hash(Addr addr, unsigned int way) const
+BOP::index(Addr addr, unsigned int way) const
 {
-    Addr hash1 = addr >> way;
-    Addr hash2 = hash1 >> floorLog2(rrEntries);
-    return (hash1 ^ hash2) & (Addr)(rrEntries - 1);
+    Addr lrrEntries = floorLog2(rrEntries);
```
odanrc: Main: Change lrrEntries -> lrr_entries
Nit: Change lrr_entries to log_rr_entries, because this significantly improves code readability (I only understood what ´l´ meant when I saw the fix in the tag function).

From gem5´s coding style (https://www.gem5.org/documentation/general_docs/development/coding_style/):
Local variables are lower case, with underscores separating words (e.g., local_variable).

src/mem/cache/prefetch/bop.cc
```diff
@@ -114,22 +111,24 @@ BOP::delayQueueEventWrapper()
...
+    Addr lineaddr = addr >> lBlkSize;
```
odanrc: Ditto line_addr

src/mem/cache/prefetch/bop.cc
```diff
@@ -201,27 +207,34 @@ BOP::bestOffsetLearning(Addr x)
...
+    }
 
-        // Check if the best offset must be updated if:
-        // (1) One of the scores equals SCORE_MAX
-        // (2) The number of rounds equals ROUND_MAX
-        if ((bestScore >= scoreMax) || (round == roundMax)) {
-            bestOffset = phaseBestOffset;
+    // Check if its time to re-calculate the best offset
+    if ((bestScore >= scoreMax) || (round >= roundMax))
+    {
```
odanrc: Check the Braces section in the coding style:
For control blocks (if, while, etc.), opening braces must be on the same line as the control keyword with a space between the closing parenthesis and the opening brace.

src/mem/cache/prefetch/bop.cc
```diff
@@ -114,22 +111,24 @@ BOP::delayQueueEventWrapper()
 }
 
 unsigned int
-BOP::hash(Addr addr, unsigned int way) const
+BOP::index(Addr addr, unsigned int way) const
```
odanrc, line 126: Nit: in general it is better to avoid these modifications that don´t really change anything, as this could needlessly create merge conflicts for other users. There is no need to revert, though, as it is quite unlikely that someone is modifying the prefetcher.

src/mem/cache/prefetch/bop.cc
```diff
@@ -185,14 +184,21 @@ BOP::testRR(Addr addr) const
 }
 
 void
-BOP::bestOffsetLearning(Addr x)
+BOP::bestOffsetLearning(Addr addr)
```
odanrc: Since addr is not an address but a tag, rename it to addr_tag or something similar. Then update the comment below to remove the outdated information

src/mem/cache/prefetch/bop.cc
```diff
@@ -201,27 +207,34 @@ BOP::bestOffsetLearning(Addr x)
...
+            if (bestScore > badScore)
+            {
+                bestOffset = phaseBestOffset;
+                round = 0;
+                bestScore = 0;
+                phaseBestOffset = 0;
+                resetScores();
+                issuePrefetchRequests = true;
+            }
+            else
```
odanrc: nit: I suggest adding braces to the else too, in order to make it less easy to add bugs in the future

src/mem/cache/prefetch/bop.cc
```diff
@@ -201,27 +207,34 @@ BOP::bestOffsetLearning(Addr x)
...
-        if ((bestScore >= scoreMax) || (round == roundMax)) {
-            bestOffset = phaseBestOffset;
+    // Check if its time to re-calculate the best offset
+    if ((bestScore >= scoreMax) || (round >= roundMax))
+    {
             round = 0;
+            if (bestScore > badScore)
+            {
+                bestOffset = phaseBestOffset;
+                round = 0;
```
odanrc: We already assign 0 to round, bestScore and phaseBestOffset, and call resetScores outside this conditional block, so something is looking fishy here. Is it a bug or is it duplicated code?

src/mem/cache/prefetch/bop.cc
```diff
@@ -234,21 +247,21 @@ BOP::calculatePrefetch(const PrefetchInfo &pfi,
...
-    bestOffsetLearning(tag_x);
+    bestOffsetLearning(addr);
 
-    // This prefetcher is a degree 1 prefetch, so it will only generate one
-    // prefetch at most per access
     if (issuePrefetchRequests) {
-        Addr prefetch_addr = addr + (bestOffset << lBlkSize);
-        addresses.push_back(AddrPriority(prefetch_addr, 0));
-        DPRINTF(HWPrefetch, "Generated prefetch %#lx\n", prefetch_addr);
+        for (int d = 1; d <= degree; d++) {
```
odanrc: Nit: in the paper he uses d as offset. I suggest renaming this variable to avoid confusing

src/mem/cache/prefetch/bop.cc
```diff
@@ -201,27 +207,34 @@ BOP::bestOffsetLearning(Addr x)
...
-        // Check if the best offset must be updated if:
-        // (1) One of the scores equals SCORE_MAX
-        // (2) The number of rounds equals ROUND_MAX
-        if ((bestScore >= scoreMax) || (round == roundMax)) {
-            bestOffset = phaseBestOffset;
+    // Check if its time to re-calculate the best offset
+    if ((bestScore >= scoreMax) || (round >= roundMax))
+    {
             round = 0;
+            if (bestScore > badScore)
```
odanrc: Could you document why this is checked? It has been years since I reviewed this code/properly read the paper, so I don´t recall how it should work anymore.

src/mem/cache/prefetch/bop.cc
```diff
@@ -114,22 +111,24 @@ BOP::delayQueueEventWrapper()
...
 unsigned int
-BOP::hash(Addr addr, unsigned int way) const
+BOP::index(Addr addr, unsigned int way) const
 {
-    Addr hash1 = addr >> way;
-    Addr hash2 = hash1 >> floorLog2(rrEntries);
-    return (hash1 ^ hash2) & (Addr)(rrEntries - 1);
+    Addr lrrEntries = floorLog2(rrEntries);
+    Addr lineaddr = addr >> lBlkSize;
+    Addr hash = lineaddr ^ (lineaddr >> (lrrEntries << way));
```
odanrc: ¨The RR table is accessed through a simple hash function.
For instance, for a 256-entry RR table, we XOR the 8 least
significant line address bits with the next 8 bits to obtain the
table index. For 12-bit tags, we skip the 8 least significant
line address bits and extract the next 12 bits.¨

- Here we use (64-8)-bit tags, right?
- The paper does not mention using the way here, and I don´t recall what is its used for. Could you document why?

odanrc (changes requested):
I had to resort to the paper (I just skimmed across) to try to understand what is being fixed. My points of issue with the [Issue](https://github.com/gem5/gem5/issues/1402) were the use of ¨incorrectly¨ in bullet 2 without any description of what is the expected correct methodology. Similarly, in bullet 3 it would be nice to have a summary of how it was implemented, and how it was supposed to be implemented. Could you fill this information in the Issue so that the next review is more focused on the algorithm?

src/mem/cache/prefetch/bop.cc
Setu-Gupta: Since the current version of BOP was not working as expected, i.e. it wasn't providing the expected performance on Spec workloads, I started looking for other implementations of BOP to see what could be the problem. While examining the prefetcher, I found that it was experiencing a lot of hash collisions while indexing the RR Table. To figure out if this was intended behavior or not, I started searching for other implementations of BOP online and I found the original code submitted by the author for DPC-2 [here](https://comparch-conf.gatech.edu/dpc2/final_program.html).

If you look at the source code, you'll two functions `rr_index_left` and `rr_index_right` which are used to generate the index which would be used to index the RR table. This logic is implemented in the `index` function in this PR. In the author's code you'll see that he multiplies the RRINDEX by 2 for indexing the right side, whereas he doesn't do the same for the left side. In this PR the `index` function is called with 0 for left, whereas it is called with 1 for right  (refer to the enum `RRWay` defined at line 57 of `bop.hh`). Therefore, if we right shift with the parameter `way`, then when `way` corresponds to left, we'll right shift with 0, and when it corresponds to right, we'll right shift by 1, i.e. multiply by 2. Thus, this code is equivalent to what the author submitted for DPC-2.

src/mem/cache/prefetch/bop.cc
Setu-Gupta: Fixed in the latest commit

src/mem/cache/prefetch/bop.cc
Setu-Gupta: Fixed in the latest commit

src/mem/cache/prefetch/bop.cc
Setu-Gupta: Fixed in the latest commit

src/mem/cache/prefetch/bop.cc
Setu-Gupta: Fixed in the latest commit

src/mem/cache/prefetch/bop.cc
Setu-Gupta: Fixed in the latest commit

src/mem/cache/prefetch/bop.cc
Setu-Gupta: The reason why I changed the function name to index is because the original author's code which was submitted for DPC-2 used the word index instead of hash. Also, while I was reading this code, the first time I came across this function, the name `hash` confused me as this function was actually being used to index the RR Table. That's why I thought it would be better to change the name to something more meaningful

src/mem/cache/prefetch/bop.cc
Setu-Gupta: Fixed in the latest commit

src/mem/cache/prefetch/bop.cc
Setu-Gupta: Fixed in the latest commit

src/mem/cache/prefetch/bop.cc
Setu-Gupta: Fixed in the latest commit

src/mem/cache/prefetch/bop.hh
Setu-Gupta: Fixed in the latest commit

src/mem/cache/prefetch/bop.cc
Setu-Gupta: Fixed in the latest commit. I also fixed the check that I had removed.

src/mem/cache/prefetch/bop.cc
Setu-Gupta: > Here we use (64-8)-bit tags, right?

We will use 8 bit tags if there are 256 entries in the RR Table. While this is the default size of the RR Table, it is not a constant. One can in fact change the size if they wish to do so.

> The paper does not mention using the way here, and I don´t recall what is its used for. Could you document why?

Since the reason why I made this change is not mentioned in the paper, should I still add a comment explaining the logic?

Setu-Gupta:
> I had to resort to the paper (I just skimmed across) to try to understand what is being fixed. My points of issue with the [Issue](https://github.com/gem5/gem5/issues/1402) were the use of ¨incorrectly¨ in bullet 2 without any description of what is the expected correct methodology. Similarly, in bullet 3 it would be nice to have a summary of how it was implemented, and how it was supposed to be implemented. Could you fill this information in the Issue so that the next review is more focused on the algorithm?

I've updated the issue. I hope now the reviewers would be able to understand the changes.

src/mem/cache/prefetch/bop.cc
```diff
@@ -201,27 +215,37 @@ BOP::bestOffsetLearning(Addr x)
...
+        if (bestScore > badScore)  {
             bestOffset = phaseBestOffset;
             round = 0;
             bestScore = 0;
             phaseBestOffset = 0;
             resetScores();
             issuePrefetchRequests = true;
-        } else if (bestScore <= badScore) {
+        }
+        else {
```
odanrc: According to the guidelines the else must be in the same line as the }

src/mem/cache/prefetch/bop.cc
odanrc: It does not seem to have changed in the latest commit

odanrc:
> > I had to resort to the paper (I just skimmed across) to try to understand what is being fixed. My points of issue with the [Issue](https://github.com/gem5/gem5/issues/1402) were the use of ¨incorrectly¨ in bullet 2 without any description of what is the expected correct methodology. Similarly, in bullet 3 it would be nice to have a summary of how it was implemented, and how it was supposed to be implemented. Could you fill this information in the Issue so that the next review is more focused on the algorithm?
> 
> I've updated the issue. I hope now the reviewers would be able to understand the changes.

That update was great! Thanks!

src/mem/cache/prefetch/bop.cc
odanrc: Yes, please add a comment explaining this logic! Your explanation was great, and will make it much easier to understand the code.

src/mem/cache/prefetch/bop.cc
odanrc: This one was not updated either

Setu-Gupta: Fixed now

src/mem/cache/prefetch/bop.cc
Setu-Gupta: Fixed now. It was indeed duplicate code. Thanks for pointing it out. I've removed the redundant code

src/mem/cache/prefetch/bop.cc
Setu-Gupta: I've added a comment explaining the calculation

src/mem/cache/prefetch/bop.cc
Setu-Gupta: Fixed. By the way, is there a clang format file which outlines all the rules for Gem5 somewhere? I'm asking as right now I'm fixing the formatting manually because of which I sometimes miss some changes. With a clang-format file, all this work can be automated. Moreover, we can also add it to the commit hooks so that it automatically lints the code before committing

powerjg:
>  By the way, is there a clang format file which outlines all the rules for Gem5 somewhere? I'm asking as right now I'm fixing the formatting manually because of which I sometimes miss some changes. With a clang-format file, all this work can be automated. Moreover, we can also add it to the commit hooks so that it automatically lints the code before committing

Great idea! https://github.com/orgs/gem5/discussions/333

odanrc (approved):
Thanks for the changes!

ivanaamit (approved):

PR #1404, opened 2024-07-31
mem-ruby,sim-se: Clear LL/SC locks after functional writes

hexengraf:
Functional writes atomically update all copies of a data block, so they should invalidate any pending LL/SC locks, just like a conventional write would.

Change-Id: Ic79d2d8d24901f1b6a2ce81dc0e2decc84c0ebbc

### DIFF
--- a/src/mem/ruby/system/Sequencer.cc
+++ b/src/mem/ruby/system/Sequencer.cc
@@ -269,6 +269,9 @@ Sequencer::functionalWrite(Packet *func_pkt)
                 ++num_written;
         }
     }
+    // Functional writes to addresses being monitored
+    // will fail (remove) the monitor entry.
+    llscClearMonitor(makeLineAddress(func_pkt->getAddr()));
 
     return num_written;
 }


hnpl (approved):
Thanks for the contribution.

powerjg:
Functional accesses should be only *debugging* accesses (e.g., for a SE mode system call). I don't know if I follow why they should clear LL/SC if they are not part of the "functionality" of the hardware we're simulating.

I'll note that the name "functional" is quite confusing and "debugging access" would be a better name.

Could you explain what the situation is that the functional access needs to clear the LL/SC?

hexengraf:
Unless my understanding of Syscall Emulation mode is completely off-basis, I don't see functional reads/writes as "debugging access", since they're still performing expected functionality of the workload (in an atomic, non-simulated way).

Emulated syscalls still interact with the same address space which is used by the simulation. So it is possible to have, for instance, a syscall called in a thread interacting with a shared location which is the target of a simulated LL/SC in another thread.

I don't have a concrete example of the lack of this code causing trouble, but I think the logic tracks.

What I've seen in practice is emulated mutex calls _reading_ the same address that is the target of simulated LL/SC operations.
If these emulated calls were _writing_ to the address, then they should clear the LL/SC locks (because that would be the expected behavior if the same syscall was being simulated instead of emulated).

Imagine you have the following implementation of CAS using LL/SC (taken from RISC-V's specification):
```
cas:
  lr.w t0, (a0) # Load original value.
  bne t0, a1, fail # Doesn't match, so fail.
  sc.w t0, a2, (a0) # Try to update.
  bnez t0, cas # Retry if store-conditional failed.
  li a0, 0 # Set return to success.
  jr ra # Return.
```
If an emulated syscall happens to write to `(a0)` between `bne t0, a1, fail` and `sc.w t0, a2, (a0)`, this CAS may result in a false success.

powerjg:
I see what you're saying... This is a tough one.

Functional accesses are not meant to cause side effects. They should be side effect-free. E.g., if using the remote GDB feature to read memory/registers you don't want that to have any effect on the system state.

I can see how with mutexes in SE mode you would need the SE mode accesses to have side effects. I think my *preference* would be to say "use FS mode for this". 

Alternatively, in the mutex implementation in SE mode we could clear the LL/SC. I think I would prefer a specialized fix for a particular SE mode thing than a generic change to the functional accesses.

The higher-level point is that this generic change to functional accesses scares me :). It's a big change that could have some follow-on effects that aren't obvious.

hexengraf:
Ok, let me see if I correctly understood your point.

Functional accesses are expected to be auxiliary functions to be used inside debug tools like GDB?
If the answer is yes, I would say one of the problems here is conflating two different usages (SE syscalls and debug tools) into a single API (functional accesses).

I can see how functional _reads_ are side-effect free, but functional _writes_ as well? A write, by definition, is a side-effect on the system state (at least from the workload's perspective). Sounds weird to me having functional write as a debug operation (maybe I'm just failing to see the application).

In the specific case of SE mode as a whole, I would argue that functional accesses having side-effects is _expected_. We emulate the execution of a syscall outside the simulation, but we still want it to take effect on the simulation itself. Handling this case-by-case doesn't sound like a scalable solution for SE.

Personally, I don't see any reasons to fear unexpected implications from this change (as far as SE mode is concerned), but if there are other uses of functional writes outside SE mode and you judge it to be unsafe doing this change, I'm okay with that as well.

If that is the case, my suggestion is to plan a way to isolate SE mode to an independent API, or change the API to indicate when it is being used by syscalls.

powerjg:
> If the answer is yes, I would say one of the problems here is conflating two different usages (SE syscalls and debug tools) into a single API (functional accesses).

Yes, this is definitely a problem.

I think I understand the issue better now, and I've come around to your arguments. Thanks for working with me on this :)

powerjg (approved):

PR #1409, opened 2024-08-02
python,misc,stdlib: addressing mypy lints

AnthonyMichaelTDM:
This pr includes various type fixes, small refactors, import updates, etc. that aim to reduce the number of errors emitted by

```sh
# /workspace/gem5/
mypy ./src/python
```

without introducing breaking changes or causing any of the tests run by

```sh
# /workspace/gem5/tests/
./main.py run -j $(nproc) -t $(nproc)
```
to fail

I've reduced the number of errors from 706 on current stable, down to 48.

Reducing them further is certainly possible, and something I'm willing to explore (so that we can incorporate mypy into the ci system), however it may require breaking changes, or modifying things I don't understand well enough to be comfortable refactoring.

Many of the TODO's added here are more questions that I have, it may be worth addressing them before committing to merging this.

### DIFF
--- a/src/python/gem5/components/boards/abstract_board.py
+++ b/src/python/gem5/components/boards/abstract_board.py
@@ -173,12 +173,18 @@ class AbstractBoard:
         """
         return self._cache_hierarchy
 
-    def get_cache_line_size(self) -> int | None:
+    def get_cache_line_size(self) -> int:
         """Get the size of the cache line.
 
         :returns: The size of the cache line size.
         """
-        return getattr(self, "cache_line_size", None)
+        if (line_size := getattr(self, "cache_line_size", None)) is not None:
+            return line_size
+        else:
+            # TODO: is it better to just return 0 (or some other default value) if the attribute is not set?
+            raise Exception(
+                "The cache line size has not been set for this board."
+            )
 
     def connect_system_port(self, port: Port) -> None:
         self.system_port = port
@@ -218,7 +224,7 @@ class AbstractBoard:
         This function is used by the Simulator module to setup the simulation
         correctly.
         """
-        if self._is_fs == None:
+        if self._is_fs is None:
             raise Exception(
                 "The workload for this board not yet to be set. "
                 "Whether the board is to be executed in FS or SE "
--- a/src/python/gem5/components/boards/abstract_board.py
+++ b/src/python/gem5/components/boards/abstract_board.py
@@ -29,21 +29,25 @@ from abc import (
     ABCMeta,
     abstractmethod,
 )
+from pathlib import Path
 from typing import (
+    TYPE_CHECKING,
     List,
     Optional,
     Sequence,
     Tuple,
 )
 
-from m5.objects import (
-    AddrRange,
+from m5.objects.ClockDomain import (
     ClockDomain,
-    IOXBar,
-    Port,
     SrcClockDomain,
-    System,
-    VoltageDomain,
+)
+from m5.objects.System import System
+from m5.objects.VoltageDomain import VoltageDomain
+from m5.objects.XBar import IOXBar
+from m5.params import (
+    AddrRange,
+    Port,
 )
 
 from ...resources.resource import WorkloadResource
@@ -51,6 +55,13 @@ from .mem_mode import (
     MemMode,
     mem_mode_to_string,
 )
+
+if TYPE_CHECKING:
+    from ..cachehierarchies.abstract_cache_hierarchy import (
+        AbstractCacheHierarchy,
+    )
+    from ..memory.abstract_memory_system import AbstractMemorySystem
+    from ..processors.abstract_processor import AbstractProcessor
 
 
 class AbstractBoard:
@@ -101,20 +112,22 @@ class AbstractBoard:
         # Set the processor, memory, and cache hierarchy.
         self.processor = processor
         self.memory = memory
-        self._cache_hierarchy = cache_hierarchy
+        self._cache_hierarchy: Optional[
+            "AbstractCacheHierarchy"
+        ] = cache_hierarchy
         if cache_hierarchy is not None:
-            self.cache_hierarchy = cache_hierarchy
+            self.cache_hierarchy: "AbstractCacheHierarchy" = cache_hierarchy
 
         # This variable determines whether the board is to be executed in
         # full-system or syscall-emulation mode. This is set when the workload
         # is defined. Whether or not the board is to be run in FS mode is
         # determined by which kind of workload is set.
-        self._is_fs = None
+        self._is_fs: bool | None = None
 
         # This variable is used to record the checkpoint directory which is
         # set when declaring the board's workload and then used by the
         # Simulator module.
-        self._checkpoint = None
+        self._checkpoint: Path | None = None
 
         # Setup the board and memory system's memory ranges.
         self._setup_memory_ranges()
@@ -133,7 +146,10 @@ class AbstractBoard:
         """
         return self.processor
 
-    def get_memory(self) -> "AbstractMemory":
+    # NOTE: Am I sure that this should be a AbstractMemorySystem instead of a m5.objects.AbstractMemory?
+    # the type being assigned to this previously would imply that, but I'd have to ensure that that's how it's used as well
+    # if it's used as a m5.objects.AbstractMemory then I need either refactor the constructor or convert the values to the correct type
+    def get_memory(self) -> "AbstractMemorySystem":
         """Get the memory (RAM) connected to the board.
 
         :returns: The memory system.
@@ -157,12 +173,12 @@ class AbstractBoard:
         """
         return self._cache_hierarchy
 
-    def get_cache_line_size(self) -> int:
+    def get_cache_line_size(self) -> int | None:
         """Get the size of the cache line.
 
         :returns: The size of the cache line size.
         """
-        return self.cache_line_size
+        return getattr(self, "cache_line_size", None)
 
     def connect_system_port(self, port: Port) -> None:
         self.system_port = port
@@ -265,7 +281,7 @@ class AbstractBoard:
         raise NotImplementedError
 
     @abstractmethod
-    def get_dma_ports(self) -> List[Port]:
+    def get_dma_ports(self) -> List[Port] | None:
         """Get the board's Direct Memory Access ports.
         This abstract method must be implemented within the subclasses if they
         support DMA and/or full system simulation.
@@ -369,8 +385,8 @@ class AbstractBoard:
         self.get_memory().incorporate_memory(self)
 
         # Incorporate the cache hierarchy for the motherboard.
-        if self.get_cache_hierarchy():
-            self.get_cache_hierarchy().incorporate_cache(self)
+        if cache_hierarchy := self.get_cache_hierarchy():
+            cache_hierarchy.incorporate_cache(self)
 
         # Incorporate the processor into the motherboard.
         self.get_processor().incorporate_processor(self)
@@ -380,8 +396,8 @@ class AbstractBoard:
     def _post_instantiate(self):
         """Called to set up anything needed after ``m5.instantiate``."""
         self.get_processor()._post_instantiate()
-        if self.get_cache_hierarchy():
-            self.get_cache_hierarchy()._post_instantiate()
+        if cache_hierarchy := self.get_cache_hierarchy():
+            cache_hierarchy._post_instantiate()
         self.get_memory()._post_instantiate()
 
     def _pre_instantiate(self):
--- a/src/python/gem5/components/boards/abstract_board.py
+++ b/src/python/gem5/components/boards/abstract_board.py
@@ -112,11 +112,9 @@ class AbstractBoard:
         # Set the processor, memory, and cache hierarchy.
         self.processor = processor
         self.memory = memory
-        self._cache_hierarchy: Optional[
-            "AbstractCacheHierarchy"
-        ] = cache_hierarchy
+        self._cache_hierarchy = cache_hierarchy
         if cache_hierarchy is not None:
-            self.cache_hierarchy: "AbstractCacheHierarchy" = cache_hierarchy
+            self.cache_hierarchy = cache_hierarchy
 
         # This variable determines whether the board is to be executed in
         # full-system or syscall-emulation mode. This is set when the workload
--- a/src/python/gem5/components/boards/abstract_board.py
+++ b/src/python/gem5/components/boards/abstract_board.py
@@ -334,6 +334,16 @@ class AbstractBoard:
         """
         raise NotImplementedError
 
+    def get_mem_ranges(self) -> List[AddrRange]:
+        if not hasattr(self, "mem_ranges"):
+            raise Exception(
+                "The board does not have any memory ranges. This is likely "
+                "due to the memory ranges not being set in the `_setup_memory_ranges` "
+                "function."
+            )
+
+        return getattr(self, "mem_ranges")
+
     @abstractmethod
     def _setup_memory_ranges(self) -> None:
         """
--- a/src/python/gem5/components/boards/abstract_board.py
+++ b/src/python/gem5/components/boards/abstract_board.py
@@ -144,9 +144,6 @@ class AbstractBoard:
         """
         return self.processor
 
-    # NOTE: Am I sure that this should be a AbstractMemorySystem instead of a m5.objects.AbstractMemory?
-    # the type being assigned to this previously would imply that, but I'd have to ensure that that's how it's used as well
-    # if it's used as a m5.objects.AbstractMemory then I need either refactor the constructor or convert the values to the correct type
     def get_memory(self) -> "AbstractMemorySystem":
         """Get the memory (RAM) connected to the board.
 
--- a/src/python/gem5/components/boards/abstract_system_board.py
+++ b/src/python/gem5/components/boards/abstract_system_board.py
@@ -25,18 +25,26 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 from abc import ABCMeta
-
-from m5.objects import (
-    SimObject,
-    System,
+from typing import (
+    TYPE_CHECKING,
+    Optional,
 )
+
+from m5.objects.System import System
+from m5.SimObject import SimObject
 
 from ...utils.override import overrides
 from .abstract_board import AbstractBoard
 
+if TYPE_CHECKING:
+    from ..cachehierarchies.abstract_cache_hierarchy import (
+        AbstractCacheHierarchy,
+    )
+    from ..memory.abstract_memory_system import AbstractMemorySystem
+    from ..processors.abstract_processor import AbstractProcessor
+
 
 class AbstractSystemBoard(System, AbstractBoard):
-
     """
     An abstract board for cases where boards should inherit from System.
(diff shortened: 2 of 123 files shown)


AnthonyMichaelTDM:
it seems that while rebasing it modified one of the CI workflows? If needed I can rebase again or something

AnthonyMichaelTDM:
yeah, I'm gonna rebase that commit out of the tree

src/python/gem5/components/boards/abstract_board.py
```diff
@@ -29,21 +29,25 @@
...
-    System,
-    VoltageDomain,
+)
+from m5.objects.System import System
+from m5.objects.VoltageDomain import VoltageDomain
+from m5.objects.XBar import IOXBar
+from m5.params import (
+    AddrRange,
+    Port,
 )
```
powerjg, line 51: I would rather not make these kinds of changes. It's a bit different from the traditional gem5 style.

AnthonyMichaelTDM:
huh, I'll look into a fix for that rq

src/python/gem5/components/boards/abstract_board.py
```diff
@@ -101,20 +112,22 @@ def __init__(
         # Set the processor, memory, and cache hierarchy.
         self.processor = processor
         self.memory = memory
-        self._cache_hierarchy = cache_hierarchy
+        self._cache_hierarchy: Optional[
+            "AbstractCacheHierarchy"
+        ] = cache_hierarchy
         if cache_hierarchy is not None:
-            self.cache_hierarchy = cache_hierarchy
+            self.cache_hierarchy: "AbstractCacheHierarchy" = cache_hierarchy
```
powerjg: This code seems correct, but it's really ugly. I am going to push back a bit on making changes that make the code more difficult to read.

In general, the type checking in python should improve the code's readability, not make it worse.

src/python/gem5/components/boards/abstract_board.py
```diff
@@ -133,7 +146,10 @@ def get_processor(self) -> "AbstractProcessor":
         """
         return self.processor
 
-    def get_memory(self) -> "AbstractMemory":
+    # NOTE: Am I sure that this should be a AbstractMemorySystem instead of a m5.objects.AbstractMemory?
```
powerjg: You're correct. It should be AbstractMemorySystem

src/python/gem5/components/boards/abstract_board.py
```diff
@@ -162,7 +178,13 @@ def get_cache_line_size(self) -> int:
 
         :returns: The size of the cache line size.
         """
-        return self.cache_line_size
+        if (line_size := getattr(self, "cache_line_size", None)) is not None:
+            return line_size
+        else:
+            # TODO: is it better to just return 0 (or some other default value) if the attribute is not set?
+            raise Exception(
```
powerjg, line 180: `cache_line_size` is a parameter of `System`, so maybe this function should be in the `AbstractSystemBoard`

src/python/gem5/components/boards/abstract_board.py
```diff
@@ -318,6 +340,16 @@ def get_mem_side_coherent_io_port(self):
         """
         raise NotImplementedError
 
+    def get_mem_ranges(self) -> List[AddrRange]:
+        if not hasattr(self, "mem_ranges"):
```
powerjg, line 339: Same, this is a parameter of `System`

src/python/gem5/components/boards/abstract_board.py
```diff
@@ -369,8 +401,8 @@ def _connect_things(self) -> None:
         self.get_memory().incorporate_memory(self)
 
         # Incorporate the cache hierarchy for the motherboard.
-        if self.get_cache_hierarchy():
-            self.get_cache_hierarchy().incorporate_cache(self)
+        if cache_hierarchy := self.get_cache_hierarchy():
```
powerjg, line 399: I don't like the idea of the walrus operator... call me old fashioned...

IMO, the tradeoff of slightly harder to read to save 9 characters isn't worth it.

src/python/gem5/components/boards/abstract_system_board.py
```diff
@@ -48,7 +56,7 @@ def __init__(
         clk_freq: str,
         processor: "AbstractProcessor",
         memory: "AbstractMemorySystem",
-        cache_hierarchy: "AbstractCacheHierarchy",
+        cache_hierarchy: Optional["AbstractCacheHierarchy"] = None,
```
powerjg, line 59: This optional should probably be documented why it's optional somewhere...

powerjg (commented):
Wow! This is a ton of work! Thanks for the effort.

I am making some comments on the PR, though don't take these comments as action items. We should chat a bit about the right way forward.

A question for you: Do you think there's a way to do this piecemeal instead of all at once? A PR with 2000 lines changed is difficult to review.

src/python/gem5/components/boards/abstract_board.py
AnthonyMichaelTDM: That's fair, but without them, mypy sometimes can't tell that the items are in the objects module (or if it does, it recognizes them as modules rather than classes), the alternative would be to add `# type: ignore` to a bunch (but not all) of m5 imports

undoing these types of changes would greatly reduce the lines changes by this pr though, which itself could be reason enough to be honest.

let me know

src/python/gem5/components/boards/abstract_board.py
AnthonyMichaelTDM: good to know, thanks

src/python/gem5/components/boards/abstract_board.py
AnthonyMichaelTDM: I'll see if there's a more readable way to write this block

src/python/gem5/components/boards/abstract_board.py
AnthonyMichaelTDM: Although I disagree, that's fair.
The alternative is to throw `self.get_cache_hierarchy()` into a temp variable and then check for None.
The point isn't to save 9 characters, it's to ensure that we're referring to the same  "cache_hierarchy" throughout the code block, the current approach (before my changes) assumes that the functions it's calling are pure functions without side-effects, and while that tends to be true it can't be statically guaranteed and could change, which would cause the code to not work as expected.

Something like

```python
cache_hierarchy = self.get_cache_hierarchy()
if cache_hierarchy:
    ...
```

src/python/gem5/components/boards/abstract_board.py
AnthonyMichaelTDM: That makes sense, should I do that?
The reason I avoided that approach was to prevent api changes.

src/python/gem5/components/boards/abstract_system_board.py
AnthonyMichaelTDM: It's optional because AbstractBoard's is optional, and some subclasses of AbstractSystemBoard want it optional (and treat it like it is)

AnthonyMichaelTDM:
> A question for you: Do you think there's a way to do this piecemeal instead of all at once? A PR with 2000 lines changed is difficult to review.

Yes, many of these commits aren't dependent on each other and can be arbitrarily reordered (so, can be put into their own branch and merged separately), but I'd rather avoid that for now as it makes things harder on my end.

src/python/gem5/components/boards/abstract_board.py
AnthonyMichaelTDM: yeah turns out the annotations aren't necessary in this case, will remove

AnthonyMichaelTDM: fixed by [the most recent commit](https://github.com/gem5/gem5/pull/1409/commits/b73fe85e54ab679124848013b91444a7ca3aaf37)

src/python/gem5/components/boards/abstract_system_board.py
AnthonyMichaelTDM: should I add that note to a comment above the argument?

src/python/gem5/components/boards/abstract_board.py
AnthonyMichaelTDM: This is mostly due to how the type stubs are generated (I think), maybe that could be something to look at more closely?

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -240,9 +248,13 @@ def set_kernel_disk_workload(
...
-                self._checkpoint = checkpoint
-            elif isinstance(checkpoint, CheckpointResource):
-                self._checkpoint = Path(checkpoint.get_local_path())
+                self._checkpoint: Path | None = checkpoint
+            elif (
+                isinstance(checkpoint, CheckpointResource)
+                and (checkpoint_path := checkpoint.get_local_path())
+                is not None
+            ):
+                self._checkpoint = Path(checkpoint_path)
```
AnthonyMichaelTDM, line 257: there are also cases like this where the walrus operator provides a lot more convenience

src/python/gem5/components/boards/abstract_board.py
AnthonyMichaelTDM: I'll probably sneak removing that comment into a future commit

src/python/gem5/components/boards/abstract_board.py
AnthonyMichaelTDM: speaking of imports, are relative or absolute imports prefered?

src/python/gem5/components/processors/simple_core.py
```diff
@@ -43,6 +42,8 @@ class SimpleCore(BaseCPUCore):
     def __init__(self, cpu_type: CPUTypes, core_id: int, isa: ISA):
         requires(isa_required=isa)
         super().__init__(
+            # WARN: we're initializing this with the wrong type, that, or the constructor expects the wrong type
+            # figure out which is the case and fix it
             core=SimpleCore.cpu_simobject_factory(
                 isa=isa, cpu_type=cpu_type, core_id=core_id
             ),
```
AnthonyMichaelTDM, line 49: can you provide some insights on this?

src/python/gem5/resources/resource.py
```diff
@@ -1256,6 +1288,11 @@ def _get_to_path_and_downloader_partial(
             gem5_version=gem5_version,
             quiet=quiet,
         )
+
+    # WARN: This assertion seems to not be valid (causing some tesets to fail), investigate the cause
+    # in the meaning, we only assert when TYPE_CHECKING is True so that mypy treats this as a type hint
+    if TYPE_CHECKING:
+        assert to_path
     return to_path, downloader
```
AnthonyMichaelTDM, line 1296: can you provide some insight as to what was going on here? I would think that to_path would be a required argument and should not be None, but when I had this assertion it would cause some tests to fail

src/python/gem5/simulate/simulator.py
```diff
@@ -374,9 +377,10 @@ def function_generator(func: Callable):
         self._board = board
         self._full_system = full_system
         self._expected_execution_order = expected_execution_order
-        self._tick_stopwatch = []
+        self._tick_stopwatch: List[Tuple[ExitEvent, int]] = []
 
-        self._last_exit_event = None
+        # The type is actually the GlobalSimLoopExitEvent class, but that's defined in the C++ code and there are no bindings for it so I'm just using Any here.
+        self._last_exit_event: None | Any = None
```
AnthonyMichaelTDM, line 383: this can be updated once/if we have better type stubs

src/python/gem5/simulate/simulator.py
```diff
@@ -560,13 +564,22 @@ def get_last_exit_event_cause(self) -> str:
...
         """
         Returns the last exit event status code
         """
-        return self._last_exit_event.getCode()
+        # TODO: determine sane default
+        return (
+            last_exit_event.getCode()
+            if (last_exit_event := self._last_exit_event)
+            else -1
+        )
```
AnthonyMichaelTDM, line 582: Does -1 work as a sane default, or should something else be used?

Raising an exception is another option too because as we discussed, this shouldn't happen if people are doing things correctly

src/python/m5/SimObject.py
```diff
@@ -220,47 +232,44 @@ def __init__(cls, name, bases, dict):
...
-            cls._deprecated_params.parent = base._deprecated_params
-            cls._values.parent = base._values
-            cls._hr_values.parent = base._hr_values
-            cls._children.parent = base._children
-            cls._port_refs.parent = base._port_refs
-            # mark base as having been subclassed
-            base._instantiated = True
-        else:
-            cls._base = None
+        # TODO: "cls._values_dict" doesn't seem to exist, need to check if this is a typo
```
AnthonyMichaelTDM, line 265: is _values_dict a type defined in the cpp code? if so we can blame this on the stubs and do something like
```python
if TYPE_CHECKING:
    cls._values_dict: Dict[str, Any] = getattr(cls, "_values_dict")
```
in the meantime so mypy knows what's going on

src/python/m5/SimObject.py
```diff
@@ -504,7 +513,8 @@ def py_call(self, *args, **kwargs):
             return func(self, *args, **kwargs)
 
         f = py_call if override else cxx_call
-        f.__pybind = PyBindMethod(
+        # TODO: Figure out where the __pybind attribute is coming from, and adjust type hints accordingly.
+        f.__pybind = PyBindMethod(  # type: ignore
             name,
```
AnthonyMichaelTDM, line 518: just commenting here so we don't forget to address this

src/python/m5/ext/pystats/simstat.py
```diff
@@ -48,12 +48,14 @@ def __init__(
         time_conversion: Optional[TimeConversion] = None,
         simulated_begin_time: Optional[Union[int, float]] = None,
         simulated_end_time: Optional[Union[int, float]] = None,
-        **kwargs: Dict[str, Union[Group, Statistic, List[Group]]]
+        **kwargs: Dict[
+            str, Union[Group, Statistic, List[Group], List[Statistic]]
+        ],
     ):
```
AnthonyMichaelTDM, line 54: note here, I don't think this is how kwargs is supposed to be typed generally, but it's iffy, this file contains a decent chunk of the remaining mypy errors

src/python/m5/params.py
```diff
@@ -587,19 +595,22 @@ def __init__(cls, name, bases, dict):
...
+            if (size := getattr(self, "size", None)) is not None:
+                # TODO: Maybe only override min and max if they are not already set
+                # (rather than overriding both if either are unset)
+                if getattr(self, "unsigned", False):
+                    self.min = 0
+                    self.max = 2**size - 1
+                else:
+                    self.min = -(2 ** (size - 1))
+                    self.max = (2 ** (size - 1)) - 1
+            else:
```
AnthonyMichaelTDM, line 608: Adding a comment so we can discuss this TODO, does that sound like a good idea?
It seems right, but it changes behavior, so I held off doing it

src/python/m5/params.py
```diff
@@ -897,71 +908,70 @@ def pybind_predecls(cls, code):
...
+            xor_high_bit = 0
+            if xorHighBit is not None:
+                xor_high_bit = int(xorHighBit)
+            for i in range(0, self.intlvBits):
+                bit1 = intlv_high_bit - i
+                mask = 1 << bit1
+                if xor_high_bit != 0:
+                    bit2 = xor_high_bit - i
+                    mask |= 1 << bit2
+                self.masks[self.intlvBits - i - 1] = mask
```
AnthonyMichaelTDM, line 975: I'm pretty highly confident this is a correct implementation, but I would appreciate another set of eyes

AnthonyMichaelTDM:
just added some comments of my own to indicate some "points of interest" for future discussion

src/python/gem5/components/boards/abstract_board.py
AnthonyMichaelTDM: yeah, I did some testing (process: move the function to AbstractSystemBoard, run mypy to see what breaks, fix some of the new errors, run mypy again, ...) and ... there's a lot of things that would have to be refactored (so much so I'd probably put it in a new pr). It's definitely possible, if it makes sense it's probably a good idea, and I'm willing to do that work, but it would likely be a breaking change.

AnthonyMichaelTDM: TL;DR, I think this change is worth doing, but I'd suggest we wait until after this PR is figured out and do it in a separate pr, should I open an issue?

erin-le:
Adding a checklist so the gem5 dev team can review this PR commit by commit:

- [x] [misc: fix typo in licensing info](https://github.com/gem5/gem5/pull/1409/commits/c704e1b0d3ede9b34d9bc9d71bef207e7bd9140e)
- [x] [python: remove python2 artifacts](https://github.com/gem5/gem5/pull/1409/commits/6ab1e15aa72e5a5cc370c030076a69e7083443c3)
- [x] [python: refactor CheckedIntType __init__](https://github.com/gem5/gem5/pull/1409/commits/a1d33274ee7e3cb9d75e046c3b872cb07c10cf9b)
    - comment from author about TODO message that still needs to be addressed
- [ ] [python: refactor AddrRange __init__](https://github.com/gem5/gem5/pull/1409/commits/12fbf13976ea666b642611730855e81fd542d249)
- [ ] [python: refactor multidict.py](https://github.com/gem5/gem5/pull/1409/commits/294e07d5abaefc06ba122a9a05041e29a86f8bd7)
- [ ] [python: refactor MetaSimObject __init__](https://github.com/gem5/gem5/pull/1409/commits/ada5a8484c22121a99dc9ec3021ce3fa57fc09c7)
- [ ] [python: explicitly handle null cases](https://github.com/gem5/gem5/pull/1409/commits/f28b706f7a833cc64988bd90618b586517d0ff80)
- [ ] [misc: add mypy configuration to ignore missing imports](https://github.com/gem5/gem5/pull/1409/commits/d3d62c5d054ad25e8a3f87d407594b7c764d105d)
- [ ] [python: update imports, type hints, and fix typos](https://github.com/gem5/gem5/pull/1409/commits/d0d91782a2180923e380a9a5893fed62c6c715d6)
- [ ] [python: add get_mem_ranges function to AbstractBoard](https://github.com/gem5/gem5/pull/1409/commits/37d2f498d5d7835e06d7df4ebe567a622cc0c462)
- [ ] [python: further improve null handling](https://github.com/gem5/gem5/pull/1409/commits/9a5d64ad655dbf8ca39ac931f5601da961e1af5f)
- [ ] [python: update imports, type hints, and fix typos. Part 2](https://github.com/gem5/gem5/pull/1409/commits/978d996cfbb62c7a994e248d45224ae0da90ff49)
- [ ] [python: patch regressions](https://github.com/gem5/gem5/pull/1409/commits/d2c6162a928de41b6825fe0c24e058448927dbe4)
- [ ] [python: fix compilation error](https://github.com/gem5/gem5/pull/1409/commits/9530079ab2a92fd3513525ccae63446a069e66e0)
- [ ] [python: remove ugly annotations in AbstractBoard.__init__](https://github.com/gem5/gem5/pull/1409/commits/b73fe85e54ab679124848013b91444a7ca3aaf37)

powerjg:
Thanks for this contribution. However, given the lack of activity, I am going to close this PR. If you (or anyone) wants to improve our type checking, I strongly suggest breaking the PR into smaller pieces to make reviewing easier. It may also be a good idea to add some python tests before making big refactoring changes.

PR #1419, opened 2024-08-06
cpu: LoopPoint analysis object

studyztp:
Adding SimObject and ProbeListener objects for gem5 to perform LoopPoint analysis.
The original paper for the LoopPoint paper: https://ieeexplore.ieee.org/document/9773236
LoopPoint is a sampling methodology that uses the program counter of a set of selected loops and their number of iteration to mark the execution points in a macro-benchmark for the defined regions.

Therefore, the LoopPont analysis object in gem5 will 
1. Define regions with the number of user-level instruction executed
2. Collect loop iterations for every region
3. Collect user-level basic block vector for every region
4. Exclude specific memory ranges to avoid analyzing multi-threaded libraries
5. Raise an exit event at the end of every region

The LoopPoint analysis object in gem5 is only tested with full-system simulation.

### DIFF
--- a/src/cpu/simple/probes/LooppointAnalysis.py
+++ b/src/cpu/simple/probes/LooppointAnalysis.py
@@ -84,6 +84,7 @@ class LooppointAnalysisManager(SimObject):
     cxx_exports = [
         PyBindMethod("getGlobalBBV"),
         PyBindMethod("clearGlobalBBV"),
+        PyBindMethod("getBBInstMap"),
         PyBindMethod("getGlobalInstCounter"),
         PyBindMethod("clearGlobalInstCounter"),
         PyBindMethod("getBackwardBranchCounter"),
--- a/src/cpu/simple/probes/LooppointAnalysis.py
+++ b/src/cpu/simple/probes/LooppointAnalysis.py
@@ -1,0 +1,25 @@
+# Copyright (c) 2024 The Regents of the University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
--- a/src/cpu/simple/probes/LooppointAnalysis.py
+++ b/src/cpu/simple/probes/LooppointAnalysis.py
@@ -24,6 +24,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+from m5.citations import add_citation
 from m5.objects import SimObject
 from m5.objects.Probe import ProbeListenerObject
 from m5.params import *
@@ -54,16 +55,19 @@ class LooppointAnalysis(ProbeListenerObject):
         "the LooppointAnalysis manager"
     )
     bb_valid_addr_range = Param.AddrRange(
-        AddrRange(start=0, end=0), "the valid address range for basic blocks"
+        "the valid address range for basic blocks. If the address start and"
+        "end are both 0 (AddrRange(start=0, end=0)), it means every address is"
+        "valid."
     )
     marker_valid_addr_range = Param.AddrRange(
-        AddrRange(start=0, end=0), "the valid address range for markers"
+        "the valid address range for markers. If the address start and end are"
+        "both 0 (AddrRange(start=0, end=0)), it means every address is valid."
     )
     bb_excluded_addr_ranges = VectorParam.AddrRange(
         [], "the excluded address ranges for basic blocks"
     )
     if_listening = Param.Bool(
-        True, "if the LooppointAnalysis is listening to " "the probe point"
+        True, "if the LooppointAnalysis is listening to the probe point"
     )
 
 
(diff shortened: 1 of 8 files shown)


src/cpu/simple/probes/LooppointAnalysis.py
```diff
@@ -0,0 +1,90 @@
...
+        PyBindMethod("getGlobalBBV"),
+        PyBindMethod("clearGlobalBBV"),
+        PyBindMethod("getGlobalInstCounter"),
+        PyBindMethod("clearGlobalInstCounter"),
+        PyBindMethod("getBackwardBranchCounter"),
+        PyBindMethod("getMostRecentBackwardBranchPC"),
+        PyBindMethod("getMostRecentBackwardBranchCount"),
+    ]
+
+    region_length = Param.Int(100000000, "the length of the region")
```
powerjg: Can you add a pointer to the loop point paper here? Use the `add_citation` function, IIRC.

You may also want to write up a short "paper" on how to use loop point and sim point in gem5, put it on arxiv and add that as a citation. You've done lots of work here, and you should get some credit :)

src/cpu/simple/probes/LooppointAnalysis.py
```diff
@@ -0,0 +1,90 @@
...
+
+    looppoint_analysis_manager = Param.LooppointAnalysisManager(
+        "the LooppointAnalysis manager"
+    )
+    bb_valid_addr_range = Param.AddrRange(
+        AddrRange(start=0, end=0), "the valid address range for basic blocks"
+    )
+    marker_valid_addr_range = Param.AddrRange(
+        AddrRange(start=0, end=0), "the valid address range for markers"
+    )
```
powerjg: Does it make sense to have defaults here?

src/cpu/simple/probes/LooppointAnalysis.py
```diff
@@ -0,0 +1,90 @@
...
+    bb_excluded_addr_ranges = VectorParam.AddrRange(
+        [], "the excluded address ranges for basic blocks"
+    )
+    if_listening = Param.Bool(
+        True, "if the LooppointAnalysis is listening to " "the probe point"
```
powerjg:
```suggestion
        True, "if the LooppointAnalysis is listening to the probe point"
```

src/cpu/simple/probes/LooppointAnalysis.py
powerjg:
```suggestion
    region_length = Param.Int(100_000_000, "the length of the region")
```

src/cpu/simple/probes/looppoint_analysis.cc
```diff
@@ -0,0 +1,280 @@
...
+      bbValidAddrRange(params.bb_valid_addr_range),
+      markerValidAddrRange(params.marker_valid_addr_range),
+      ifListening(params.if_listening),
+      bbInstCounter(0)
+{
+    DPRINTF(LooppointAnalysis, "Start listening from the beginning of the "
+                            "simulation? %s\n", ifListening ? "Yes" : "No");
+
+    for (int i = 0; i < params.bb_excluded_addr_ranges.size(); i++)
+    {
```
powerjg:
```suggestion
    for (int i = 0; i < params.bb_excluded_addr_ranges.size(); i++) {
```

src/cpu/simple/probes/looppoint_analysis.cc
```diff
@@ -0,0 +1,280 @@
...
+LooppointAnalysis::updateLocalBBV(const Addr pc)
+{
+    if (localBBV.find(pc) == localBBV.end())
+    {
+        localBBV.insert(std::make_pair(pc, 1));
+    }
+    else
+    {
+        localBBV.find(pc)->second++;
+    }
```
powerjg: I'm not going to do this for the whole file, but please change this style

```suggestion
    if (localBBV.find(pc) == localBBV.end()) {
        localBBV.insert(std::make_pair(pc, 1));
    } else {
        localBBV.find(pc)->second++;
    }
```

src/cpu/simple/probes/looppoint_analysis.cc
```diff
@@ -0,0 +1,280 @@
...
+}
+
+void
+LooppointAnalysis::checkPc(const std::pair<SimpleThread*,
+                                                     StaticInstPtr>& inst_pair)
```
powerjg: Align with the `(`

src/cpu/simple/probes/looppoint_analysis.cc
```diff
@@ -0,0 +1,280 @@
...
+    {
+        // ignore this if it is not in the valid address range
+        return;
+    }
+
+    if (bbExcludedAddrRanges.size() > 0)
+    {
+        for (int i = 0; i < bbExcludedAddrRanges.size(); i++)
+        {
+            if (pc >= bbExcludedAddrRanges[i].start() &&
```
powerjg: Can you use `contains` (https://github.com/gem5/gem5/blob/stable/src/base/addr_range.hh#L489)?

src/cpu/simple/probes/looppoint_analysis.cc
```diff
@@ -0,0 +1,280 @@
...
+        // if it is a control instruction, we see it as the end of a basic
+        // block
+        lpaManager->updateValidControl(pc);
+        lpaManager->updateBBInstMap(pc, bbInstCounter);
+        updateLocalBBV(pc);
+        lpaManager->updateGlobalBBV(pc);
+        bbInstCounter = 0;
+
+        if (markerValidAddrRange.end() > 0 &&
+         (pc < markerValidAddrRange.start() || pc > markerValidAddrRange.end())
```
powerjg: `contains`?

src/cpu/simple/probes/looppoint_analysis.cc
```diff
@@ -0,0 +1,280 @@
...
+    }
+}
+
+void
+LooppointAnalysis::regProbeListeners()
+{
+    if (ifListening)
+    {
+        if (listeners.size() == 0) {
+            listeners.push_back(new looppointAnalysisListener(this,
```
powerjg, line 176: See #1439 to make sure this doesn't conflict

src/cpu/simple/probes/looppoint_analysis.hh
```diff
@@ -0,0 +1,331 @@
...
+     * This variable stores the Program Counter address of the most recent
+     * valid backward branch that we consider as a marker.
+     */
+    Addr mostRecentBackwardBranchPC;
+
+    /**
+     * This set stores the Program Counter addresses of the valid backward
+     * branches.
+     */
+    std::unordered_set<Addr> backwardBranchPC;
```
powerjg: Is there a reason you can't use `backwardBranchCounter` for this?

src/cpu/simple/probes/looppoint_analysis.hh
```diff
@@ -0,0 +1,331 @@
...
+    {
+        globalInstCounter++;
+    };
+
+    Addr getMostRecentBackwardBranchPC() const
+    {
+        return mostRecentBackwardBranchPC;
+    };
+
+    std::unordered_map<Addr, uint64_t> getBackwardBranchCounter() const
```
powerjg: We should be careful with this function. This map could be very large and this function will likely return a copy. Where is it called? Should it return a copy? should it return a reference?

powerjg (commented):
Mostly small comments below. Great job!

src/cpu/simple/probes/LooppointAnalysis.py
studyztp: I added the citation with the `add_citation` function. 

Thanks! I'll prepare a write-up on how to use them :)

src/cpu/simple/probes/looppoint_analysis.cc
studyztp: It does conflict because I am using the constructor of the ProbeListener object to add the listeners to the probe point and remove them using its destructor too; but I think I can push a patch to fix it later after we decide what is the best way to add the listeners in that pull request.

src/cpu/simple/probes/looppoint_analysis.hh
studyztp: It is a great catch! I removed `backwardBranchPC`and started using `backwardBranchCounter` instead now.

src/cpu/simple/probes/looppoint_analysis.hh
studyztp: It can be called at every end of the region. In the Python script, the map will be stored as a dictionary. We don't expect to change anything in the map. 
I think by default, Pybind converts it to dictionary that copied the map's data, so it doesn't return a reference, but since we don't expect to change anything in the map. returning a copy should be fine.

src/cpu/simple/probes/LooppointAnalysis.py
studyztp: You are right. I removed the default and hint on inputting `AddrRange(start=0, end=0)` in the parameter descriptions.

studyztp:
> Mostly small comments below. Great job!

Thank you for the detailed feedback! I addressed all of them and pushed the changes. I am really sorry for being late.

studyztp:
@BobbyRBruce Here is the PR for the LoopPoint Analysis. It is independent of the InstTracker. I noticed that this PR does not show up when I search for my username in the PRs.

src/cpu/simple/probes/LooppointAnalysis.py
```diff
@@ -0,0 +1,134 @@
...
+        PyBindMethod("clearLocalBBV"),
+    ]
+
+    looppoint_analysis_manager = Param.LooppointAnalysisManager(
+        "the LooppointAnalysis manager"
+    )
+    bb_valid_addr_range = Param.AddrRange(
+        "the valid address range for basic blocks. If the address start and"
+        "end are both 0 (AddrRange(start=0, end=0)), it means every address is"
+        "valid."
```
BobbyRBruce: When you have literal strings separated by whitespace (inc. newlines) in Python (and CPP too) they concatenated. So in this case the string passed to `Param.AddrRange`  will be "the valid address range for basic blocks. If the address start andend are both 0 (AddrRange(start=0, end=0)), it means every address isvalid." (note the lack of space between "and" and "end" and "is" and "valid").

Please add spaces where needed and in the other places in this PR this occurs.

BobbyRBruce (changes requested):
One tiny thing to fix, otherwise I'm fine with this and am happy to have it merged.

src/cpu/simple/probes/LooppointAnalysis.py
studyztp: Thanks for catching this. I fixed all places where this issue might occur.

studyztp:
> One tiny thing to fix, otherwise I'm fine with this and am happy to have it merged.

I fixed the params' python string spacing issue. I thin

BobbyRBruce (approved):

PR #1420, opened 2024-08-06
arch-arm: Fix incorrect behaviour of VFNMS and VFNMA

arichardson:
This was found while comparing a diverging execution against QEMU traces
and checking for the first mismatched program counter. Fortunately this was
caused by a branch shortly after this incorrect computation but still took
a long time to track down.

There are two issues here: the decoder had inverted the cases for *S and *A,
and the sign bit was wrong for VFN*.

### DIFF
--- a/src/arch/arm/isa/formats/fp.isa
+++ b/src/arch/arm/isa/formats/fp.isa
@@ -2725,7 +2725,7 @@ let {{
             }
             break;
           case 0x9:
-            if ((opc3 & 0x1) == 0) {
+            if (bits(machInst, 6) == 1) {
                 if (single) {
                     return decodeVfpRegRegRegOp<VfnmaS>(
                             machInst, vd, vn, vm, false);


arichardson:
@giactra This was really difficult to track down since it was just a flipped sign bit - I wonder if it would be possible to run the Arm test suite against gem5? I believe all that's needed is a simple trickbox device?

src/arch/arm/isa/formats/fp.isa
```diff
@@ -2725,7 +2725,7 @@ let {{
             }
             break;
           case 0x9:
-            if ((opc3 & 0x1) == 0) {
+            if (bits(machInst, 6) == 1) {
                 if (single) {
```
arichardson, line 2729: I noticed all the 16-bit operations are missing here and will decode as the double versions, probably also worth fixing but maybe in a follow-up commit?

giactra: Yeah, I have a FEAT_FP16 implementation I plan to upstream sometime but I believe it is not covering AArch32

src/arch/arm/isa/insts/fp.isa
```diff
@@ -541,13 +541,17 @@ let {{
...
+                     "fpMulAdd<float>", "fpMulAdd<double>",
+                      "-cOp1, cOp2,  cOp3" )
     buildTernaryFpOp("Vfnma", "FpRegRegRegOp", "SimdFloatMultAccOp",
-                     "fpMulAdd<float>", "fpMulAdd<double>", "-cOp1, cOp2, -cOp3" )
+                     "fpMulAdd<float>", "fpMulAdd<double>",
+                      " cOp1, cOp2, -cOp3" )
     buildTernaryFpOp("Vfnms", "FpRegRegRegOp", "SimdFloatMultAccOp",
-                     "fpMulAdd<float>", "fpMulAdd<double>", " cOp1, cOp2, -cOp3" )
+                     "fpMulAdd<float>", "fpMulAdd<double>",
+                      "-cOp1, cOp2, -cOp3" )
```
giactra: Unless I am missing something, this does not look right to me.

According to the reference manual, you only negate the first operand in the VFNMA case, where the 6th bit is 1:

```
op1_neg = (op ==1)
constant op32 = if op1_neg then FPNeg(S[n], fpcr) else S[n];
S[d] = FPMulAdd(FPNeg(S[d], fpcr), op32, S[m], fpcr);
```

I am surprised this is matching QEMU. Could it be the bug is somewhere else?

arichardson: For VFNMS, we are subtracting the product, i.e. negating the first input operand as well as negating the destination operation, so -cop1 and -cop3 looks correct to me.
For VFNMA, we are adding the negated destination, so only cop3 is negated there.

But I just looked at the latest ARM ARM and it does seem to appear that the op1_neg does not match the encoding of VNFMA/VFNMS, the subtract version should negate the input operand not the other way around. I wonder if this is an artifict of the way the reference manual is generated? QEMU and LLVM agree that bit 6 == 0 is the subtract version and the instruction overview (https://developer.arm.com/documentation/ddi0597/2024-06/A32-Instructions-by-Encoding/System-register-access--Advanced-SIMD--floating-point--and-Supervisor-call?lang=en#fpdp3reg) also shows the subtract as having bit 6 as zero.

arichardson: It also seems wrong here: https://developer.arm.com/documentation/ddi0597/2024-06/SIMD-FP-Instructions/VFNMA--Vector-Fused-Negate-Multiply-Accumulate-?lang=en#VFNMA_A1_H. The name and the semantics definitely don't match up... And running my test code on real hardware matches QEMU rather than GEM5 so I'd be inclined to say that the reference manual is wrong. Where would be the right place to file this bug?

giactra: Hey @arichardson I can easily file this internally if you want me to.

However when you say the Arm reference manual does not match the semantics, I would point you to the compiler definition which seems to me the most intuitive:

https://developer.arm.com/documentation/dui0802/b/Advanced-SIMD-and-Floating-point-Programming--32-bit-/VFNMA--VFNMS?lang=en

> VFNMA multiplies the values in the operand registers, adds the value in the destination register, and places the final result in the destination register. The result of the multiply is not rounded before the accumulation. The final result is negated.

So the negation only happens in the final result... Meaning for the compiler guide this are the addition and subtraction:

a) S[d] = - [S[d] + (S[n] * S[m])] => Addition
b) S[d] = - [S[d] - (S[n] * S[m])] => Subtraction

So doing some transformations for the VFNMA:

1) S[d] = - [S[d] + (S[n] * S[m])]
2) S[d] = - S[d] - (S[n] * S[m])
3) S[d] = -S[d] + (-S[n] * S[m])

Brings us to the reference manual result

arichardson: Thanks for checking this - it looks like I was looking at VFMA in the manual instead of VFNMA. Sorry about the noise.

arichardson: Dropped the incorrect second commit now.

giactra (approved):

PR #1427, opened 2024-08-07
sim: Enable debug tracing configuration with M5Op

robhau:
Adds a new pseudo instruction m5_configure_tracing. It can be used to turn on or turn off the debug tracing from the workload, e.g., to skip the booting.

Change-Id: I7b86762b8e4fa918df6cf697fa7abe9e359a88b1

### DIFF
--- a/include/gem5/asm/generic/m5ops.h
+++ b/include/gem5/asm/generic/m5ops.h
@@ -12,6 +12,7 @@
  * modified or unmodified, in source code or in binary form.
  *
  * Copyright (c) 2003-2006 The Regents of The University of Michigan
+ * Copyright (c) 2024 University of Rostock
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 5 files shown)


giactra:
I can see how this is useful, but I am wondering if we can't piggyback on existing logic to achieve a similar behaviour. For example reusing workbegin/workend m5 ops which are used anyway to annotate ROIs.

abmerop:
> I can see how this is useful, but I am wondering if we can't piggyback on existing logic to achieve a similar behaviour. For example reusing workbegin/workend m5 ops which are used anyway to annotate ROIs.

For reference: The GPU model does something similar (catching an exit event and turning on debugging): https://github.com/gem5/gem5/blob/stable/configs/example/gpufs/runfs.py#L309

powerjg:
As others have said, I think this can already be done. Also, we are working to refactor the exit events in general to be more flexible. See https://github.com/gem5/gem5/issues/751 for more details.

Also, another example: https://github.com/gem5bootcamp/2024/blob/main/materials/02-Using-gem5/03-running-in-gem5/03-run-x86-SE/complete/03-run-x86-SE.py#L79 and slide number 23 here: https://bootcamp.gem5.org/#02-Using-gem5/03-running-in-gem5

robhau:
Thank you for you feedback 😄. I was not aware that this already possible. Therefore, I close this PR..

Also thank you for the example, I can then start using it for my simulations 😄.

robhau:
Thank you for you feedback 😄. I was not aware that this already possible. Therefore, I close this PR.

Also thank you for the example, I can then start using it for my simulations 😄.

powerjg:
No worries. Thanks for being part of the community and contributing back!

PR #1430, opened 2024-08-07
dev-amdgpu: Fix issues found by address sanitizer

abmerop:
These commits primarily fix the SDMA engine which was (1) using pointer arithmetic on a variable returned by new and then attempting to free the modified pointer and (2) using a buffer after it was freed due to the DMA device calling completion event before Ruby actually completed.

Some minor fixes are included: Stop using uninitialized value as packet context and using same request pointer for two separate packets for GPU invalidations.

### DIFF
--- a/src/dev/amdgpu/sdma_engine.cc
+++ b/src/dev/amdgpu/sdma_engine.cc
@@ -653,7 +653,7 @@ SDMAEngine::writeDone(SDMAQueue *q, sdmaWrite *pkt, uint32_t *dmaBuffer)
 {
     DPRINTF(SDMAEngine, "Write packet completed to %p, %d dwords\n",
             pkt->dest, pkt->count);
-    delete []dmaBuffer;
+    delete [] dmaBuffer;
     delete pkt;
     decodeNext(q);
 }
@@ -686,6 +686,7 @@ SDMAEngine::copy(SDMAQueue *q, sdmaCopy *pkt)
         // Copy the minimum page size at a time in case the physical addresses
         // are not contiguous.
         ChunkGenerator gen(pkt->source, pkt->count, AMDGPU_MMHUB_PAGE_SIZE);
+        uint8_t *buffer_ptr = dmaBuffer;
         for (; !gen.done(); gen.next()) {
             Addr chunk_addr = getDeviceAddress(gen.addr());
             assert(chunk_addr);
@@ -693,10 +694,10 @@ SDMAEngine::copy(SDMAQueue *q, sdmaCopy *pkt)
             DPRINTF(SDMAEngine, "Copying chunk of %d bytes from %#lx (%#lx)\n",
                     gen.size(), gen.addr(), chunk_addr);
 
-            gpuDevice->getMemMgr()->readRequest(chunk_addr, dmaBuffer,
+            gpuDevice->getMemMgr()->readRequest(chunk_addr, buffer_ptr,
                                                 gen.size(), 0,
                                                 gen.last() ? cb : nullptr);
-            dmaBuffer += gen.size();
+            buffer_ptr += gen.size();
         }
     } else {
         auto cb = new DmaVirtCallback<uint64_t>(
@@ -731,6 +732,7 @@ SDMAEngine::copyReadData(SDMAQueue *q, sdmaCopy *pkt, uint8_t *dmaBuffer)
         // Copy the minimum page size at a time in case the physical addresses
         // are not contiguous.
         ChunkGenerator gen(pkt->dest, pkt->count, AMDGPU_MMHUB_PAGE_SIZE);
+        uint8_t *buffer_ptr = dmaBuffer;
         for (; !gen.done(); gen.next()) {
             Addr chunk_addr = getDeviceAddress(gen.addr());
             assert(chunk_addr);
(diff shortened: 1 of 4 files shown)


src/dev/amdgpu/sdma_engine.cc
```diff
@@ -653,11 +654,27 @@ SDMAEngine::writeDone(SDMAQueue *q, sdmaWrite *pkt, uint32_t *dmaBuffer)
...
             pkt->dest, pkt->count);
-    delete []dmaBuffer;
+
+    auto cleanup_cb = new EventFunctionWrapper(
+        [ = ]{ writeCleanup(dmaBuffer); }, name());
+
+    auto system_ptr = gpuDevice->CP()->system();
+    if (!system_ptr->isAtomicMode()) {
+        warn_once("SDMA cleanup assumes 2000 tick timing for completion."
+                " This has not been testing in timing mode\n");
```
mattsinc: testing --> tested

src/dev/amdgpu/sdma_engine.cc
```diff
@@ -770,11 +790,27 @@ SDMAEngine::copyDone(SDMAQueue *q, sdmaCopy *pkt, uint8_t *dmaBuffer)
...
             pkt->dest, pkt->count);
-    delete []dmaBuffer;
+
+    auto cleanup_cb = new EventFunctionWrapper(
+        [ = ]{ copyCleanup(dmaBuffer); }, name());
+
+    auto system_ptr = gpuDevice->CP()->system();
+    if (!system_ptr->isAtomicMode()) {
+        warn_once("SDMA cleanup assumes 2000 tick timing for completion."
+                " This has not been testing in timing mode\n");
```
mattsinc: same here

src/dev/amdgpu/sdma_engine.cc
```diff
@@ -770,11 +790,27 @@ SDMAEngine::copyDone(SDMAQueue *q, sdmaCopy *pkt, uint8_t *dmaBuffer)
...
+    }
+    schedule(cleanup_cb, curTick() + 10000);
```
mattsinc, line 806: above print says 2000 ticks, but seems like 10K here?

src/dev/amdgpu/sdma_engine.cc
```diff
@@ -653,11 +654,27 @@ SDMAEngine::writeDone(SDMAQueue *q, sdmaWrite *pkt, uint32_t *dmaBuffer)
...
+
+    auto cleanup_cb = new EventFunctionWrapper(
+        [ = ]{ writeCleanup(dmaBuffer); }, name());
+
+    auto system_ptr = gpuDevice->CP()->system();
+    if (!system_ptr->isAtomicMode()) {
+        warn_once("SDMA cleanup assumes 2000 tick timing for completion."
+                " This has not been testing in timing mode\n");
+    }
+    schedule(cleanup_cb, curTick() + 10000);
```
mattsinc, line 668: above print says 2K ticks, but seems like 10K here?

src/dev/amdgpu/sdma_engine.cc
```diff
@@ -1018,11 +1054,26 @@ SDMAEngine::ptePdeDone(SDMAQueue *q, sdmaPtePde *pkt, uint64_t *dmaBuffer)
...
-    delete []dmaBuffer;
+    auto cleanup_cb = new EventFunctionWrapper(
+        [ = ]{ ptePdeCleanup(dmaBuffer); }, name());
+
+    auto system_ptr = gpuDevice->CP()->system();
+    if (!system_ptr->isAtomicMode()) {
+        warn_once("SDMA cleanup assumes 2000 tick timing for completion."
+                " This has not been testing in timing mode\n");
+    }
+    schedule(cleanup_cb, curTick() + 10000);
```
mattsinc, line 1071: same here

src/dev/amdgpu/sdma_engine.cc
```diff
@@ -1018,11 +1054,26 @@ SDMAEngine::ptePdeDone(SDMAQueue *q, sdmaPtePde *pkt, uint64_t *dmaBuffer)
...
             pkt->dest, pkt->count);
 
-    delete []dmaBuffer;
+    auto cleanup_cb = new EventFunctionWrapper(
+        [ = ]{ ptePdeCleanup(dmaBuffer); }, name());
+
+    auto system_ptr = gpuDevice->CP()->system();
+    if (!system_ptr->isAtomicMode()) {
+        warn_once("SDMA cleanup assumes 2000 tick timing for completion."
+                " This has not been testing in timing mode\n");
```
mattsinc: same here

src/gpu-compute/compute_unit.cc
```diff
@@ -409,8 +409,6 @@ ComputeUnit::doInvalidate(RequestPtr req, int kernId){
 
     // kern_id will be used in inv responses
     gpuDynInst->kern_id = kernId;
-    // update contextId field
-    req->setContext(gpuDynInst->wfDynId);
```
mattsinc, line 413: Does this get set somewhere else now?

mattsinc (changes requested):

src/gpu-compute/compute_unit.cc
abmerop: No. As far as I can tell this is not needed.

src/dev/amdgpu/sdma_engine.cc
abmerop: Added a comment here. 2K is the minimum required wait time, but I've added more in case this changes (e.g., if you change the GPU frequency it could)

src/gpu-compute/compute_unit.cc
mattsinc: I'm not sure why it goes with this commit then?  But I guess ok ...

mattsinc (approved):

src/gpu-compute/compute_unit.cc
abmerop: Things found with asan/ubsan

mattsinc: My point was more should it be a separate commit/pull request.  But small enough that probably does not matter.

abmerop: It is it's own commit, there are 3 in this PR

mattsinc: gotcha ... this is something I find much harder to conceptualize in Github vs. Gerrit

Issue #1087, opened 2024-04-30
global instruction counter


studyztp:
Currently, if we want to schedule an exit event after a certain number of instructions have been executed, we need to use either [BaseCPU::scheduleInstStopAnyThread](https://github.com/gem5/gem5/blob/stable/src/cpu/base.cc#L818) or [this](https://github.com/gem5/gem5/blob/stable/src/cpu/base.cc#L326C18-L326C40). Both methods require scheduling the exit event in advance. However, the current method has two major disadvantages:

1. It is difficult to achieve a precise global instruction count-based exit for multi-core simulations because we can only schedule these events per core. As a result, the instruction counting is not done as a whole. [related mailing list thread](https://www.mail-archive.com/gem5-users@gem5.org/msg21842.html)
2. The checkpoint might capture the future exit event and cause an incorrect exit event when restoring the checkpoint. Related mailing list thread. [related mailing list thread](https://www.mail-archive.com/gem5-users@gem5.org/msg22383.html)

Therefore, I believe it is necessary to have a global instruction counter that can trigger an exit event based on actual instruction counting, and the counting should be across multiple cores. This is easily achievable with the probing feature in gem5.  I can create a pull request for it, but I still want to discuss about if it is needed since there could also be risks associated with continually adding features.

powerjg:
If you do this with an optional probe, I think it's a great idea. Please open a PR.

BobbyRBruce:
@studyztp L was this ever done?

studyztp:
@powerjg @BobbyRBruce Sorry, I don't know why I missed the message. I'll create a PR today!

studyztp:
pull request created https://github.com/gem5/gem5/pull/1433

PR #1433, opened 2024-08-08
cpu: add GlobalInstTracker and LocalInstTracker

studyztp:
The GlobalInstTracker manages the global instruction counter and responsible for triggering an exit event when the global instruction counter reaches the defined threshold.

The LocalInstTracker listens to one core's retiredInsts probe point and updates the GlobalInstTracker every time there is an instruction committed.

The purpose of this instruction tracker is to raise an instruction executed exit event with multi-core simulation.

Related discussion can be found:
https://github.com/gem5/gem5/issues/1087

Change-Id: Iab6fec57f14f28e590b035506282130ba8662706

### DIFF
--- a/configs/example/gem5_library/x86-global-inst-tracker.py
+++ b/configs/example/gem5_library/x86-global-inst-tracker.py
@@ -24,6 +24,33 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+"""
+
+This script demonstrates how to use global and local instruction trackers to
+monitor and control the simulation based on the number of instructions
+committed.
+
+This script will create a global instruction tracker to manage each local
+instruction tracker and connects each local instruction tracker to a core.
+The global instruction tracker will raise an event when the number of
+instructions committed across all cores reaches a certain threshold.
+
+In this script, we expect to start monitorning the instruction committed from
+the start of the simulation and raise a SIMPOINT_BEGIN exit event when all
+cores in combination have committed 100,000,000 instructions. Then, we will
+change the threshold to 20,000 instructions and raise another SIMPOINT_BEGIN
+event when the new threshold is reached. Finally, we will stop listening to
+instructions.
+
+Usage:
+------
+
+scons build/X86/gem5.opt
+./build/X86/gem5.opt [--debug-flags=InstTracker] \
+    configs/example/gem5_library/x86-global-inst-tracker.py
+
+"""
+
 import m5
 from m5.objects import (
     GlobalInstTracker,
@@ -41,14 +68,6 @@ from gem5.isas import ISA
 from gem5.resources.resource import obtain_resource
 from gem5.simulate.exit_event import ExitEvent
 from gem5.simulate.simulator import Simulator
-
-"""
-
-Usage:
-
-gem5.opt configs/example/gem5_library/x86-global-inst-tracker.py
-
-"""
 
 cache_hierarchy = PrivateL1CacheHierarchy(
     l1d_size="64kB",
@@ -118,13 +137,10 @@ def max_inst_handler():
     # we can stop listening to instructions
     for tracker in all_trackers:
         tracker.stopListening()
-    """
-    similarly, we can start listening to instructions again by calling:
-
-    for tracker in all_trackers:
-        tracker.startListening()
-
-    """
+    # similarly, we can start listening to instructions again by calling:
+    #
+    # for tracker in all_trackers:
+    #     tracker.startListening()
     m5.stats.dump()
     m5.stats.reset()
     yield False
--- a/configs/example/gem5_library/x86-global-inst-tracker.py
+++ b/configs/example/gem5_library/x86-global-inst-tracker.py
@@ -1,0 +1,143 @@
+# Copyright (c) 2024 The Regents of the University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import m5
+from m5.objects import (
+    GlobalInstTracker,
+    LocalInstTracker,
+)
+
+from gem5.components.boards.simple_board import SimpleBoard
+from gem5.components.cachehierarchies.classic.private_l1_cache_hierarchy import (
+    PrivateL1CacheHierarchy,
+)
+from gem5.components.memory.single_channel import SingleChannelDDR4_2400
+from gem5.components.processors.cpu_types import CPUTypes
+from gem5.components.processors.simple_processor import SimpleProcessor
+from gem5.isas import ISA
+from gem5.resources.resource import obtain_resource
+from gem5.simulate.exit_event import ExitEvent
+from gem5.simulate.simulator import Simulator
+
+"""
+
+Usage:
+
+gem5.opt configs/example/gem5_library/x86-global-inst-tracker.py
+
+"""
+
+cache_hierarchy = PrivateL1CacheHierarchy(
+    l1d_size="64kB",
+    l1i_size="64kB",
+)
+
+memory = SingleChannelDDR4_2400("1GB")
+
+processor = SimpleProcessor(cpu_type=CPUTypes.TIMING, num_cores=9, isa=ISA.X86)
+
+# setup instruction tracker
+
+# first, we need to create a global instruction tracker
+global_inst_tracker = GlobalInstTracker(
+    # threshold to trigger the event
+    inst_threshold=100_000_000
+)
+
+# then, we create a local instruction tracker for each core
+# and store them in a list
+all_trackers = []
+
+for core in processor.get_cores():
+    tracker = LocalInstTracker(
+        # we pass in the glboal instruction tracker to the local one
+        global_inst_tracker=global_inst_tracker,
+        # this parameter tells the tracker to start listening to instructions
+        # from the beginning of the simulation. If set to False, the tracker
+        # will
+        start_listening=True,
+    )
+    # we attach the tracker to the core
+    core.core.probeListener = tracker
+    # then store the tracker in the list
+    all_trackers.append(tracker)
+
+
+board = SimpleBoard(
+    clk_freq="1GHz",
+    processor=processor,
+    memory=memory,
+    cache_hierarchy=cache_hierarchy,
+)
+
+board.set_se_binary_workload(
+    binary=obtain_resource(resource_id="x86-matrix-multiply-omp"),
+    arguments=[100, 8],
+)
+
+
+def max_inst_handler():
+    # this handler tests the instruction tracker
+    # when it reached this function, it means that it successfully raised
+    # the ExitEvent.MAX_INSTS event
+    print("Reached MAX_INSTS with 100000000 instructions")
+    print("Changing threshold to 20000")
+    # we can change the threshold of the global instruction tracker
+    global_inst_tracker.changeThreshold(20000)
+    # we need to reset the counter of the global instruction tracker
+    # the counter does not reset automatically
+    global_inst_tracker.resetCounter()
+    m5.stats.dump()
+    m5.stats.reset()
+    yield False
+    print("Reached MAX_INSTS with 20000 instructions")
+    print("Stop listening to instructions")
+    # we can stop listening to instructions
+    for tracker in all_trackers:
+        tracker.stopListening()
+    """
+    similarly, we can start listening to instructions again by calling:
+
+    for tracker in all_trackers:
+        tracker.startListening()
+
+    """
+    m5.stats.dump()
+    m5.stats.reset()
+    yield False
+
+
+simulator = Simulator(
+    board=board,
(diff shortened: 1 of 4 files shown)


configs/example/gem5_library/x86-global-inst-tracker.py
```diff
@@ -0,0 +1,143 @@
...
+    # we can stop listening to instructions
+    for tracker in all_trackers:
+        tracker.stopListening()
+    """
+    similarly, we can start listening to instructions again by calling:
+
+    for tracker in all_trackers:
+        tracker.startListening()
+
+    """
```
powerjg: Don't do multiline comments with `"""`. `"""` should only be used for doc string.

configs/example/gem5_library/x86-global-inst-tracker.py
```diff
@@ -0,0 +1,143 @@
...
+from gem5.simulate.exit_event import ExitEvent
+from gem5.simulate.simulator import Simulator
+
+"""
+
+Usage:
+
+gem5.opt configs/example/gem5_library/x86-global-inst-tracker.py
+
+"""
```
powerjg: I think this needs to be at the top of the file, before the imports. https://realpython.com/documenting-python-code/#:~:text=Module%20docstrings%20are%20placed%20at%20the%20top%20of%20the%20file%20even%20before%20any%20imports.

Also, could you explain what this fine *does* here?

src/cpu/probes/inst_tracker.cc
```diff
@@ -0,0 +1,100 @@
...
+{
+    DPRINTF(InstTracker, "ifListening = %s\n", ifListening ? "true" : "false");
+}
+
+void
+LocalInstTracker::regProbeListeners()
+{
+    if (ifListening) {
+        if (listeners.empty())
+        {
```
powerjg:
```suggestion
        if (listeners.empty()) {
```

src/cpu/probes/inst_tracker.hh
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * this function is called when the ProbePoint "RetiredInsts" is notified
+     *
+     * @param inst the number of retired instructions. It is usually 1.
+     */
+    void retiredInstsHandler(const uint64_t& inst);
+
+  private:
+    typedef ProbeListenerArg<LocalInstTracker, uint64_t>
+                                                    LocalInstTrackerListener;
```
powerjg:
```suggestion
    typedef ProbeListenerArg<LocalInstTracker, uint64_t>
        LocalInstTrackerListener;
```

src/cpu/probes/inst_tracker.hh
```diff
@@ -0,0 +1,132 @@
...
+  public:
+    /** stop listening to the ProbePoints */
+    void stopListening();
+
+    /** start listening to the ProbePoints */
+    void startListening()
+    {
+      ifListening = true;
+      regProbeListeners();
+    }
```
powerjg:
```suggestion
    void
    startListening()
    {
        ifListening = true;
        regProbeListeners();
    }
```

src/cpu/probes/inst_tracker.hh
```diff
@@ -0,0 +1,132 @@
...
+      * threshold across all the cores.
+     */
+    void updateAndCheckInstCount(const uint64_t& inst);
+
+  private:
+    /** the number of instructions that have been executed across all the cores
+    */
+    uint64_t instCount;
+
+
```
powerjg:
```suggestion

```

src/cpu/probes/inst_tracker.hh
```diff
@@ -0,0 +1,132 @@
...
+    {
+      instCount = 0;
+      DPRINTF(InstTracker, "Resetting the instruction counter\n"
+                                              "instCount = %lu\n", instCount);
+    };
+
+    uint64_t getThreshold() const
+    {
+      return instThreshold;
+    };
```
powerjg:
```suggestion
    void
    changeThreshold(uint64_t new_threshold)
    {
        instThreshold = new_threshold;
        DPRINTF(InstTracker, "Changing the instruction threshold\n"
                        "instThreshold = %lu\n", instThreshold);
    }

    void
    resetCounter()
    {
        instCount = 0;
        DPRINTF(InstTracker, "Resetting the instruction counter\n"
                       "instCount = %lu\n", instCount);
    }

    uint64_t
    getThreshold() const
    {
        return instThreshold;
    }
```

powerjg (commented):
A few small style things and some documentation requests. Overall, it looks good.

ivanaamit:
@studyztp, just a reminder to incorporate the changes Jason requested. Thanks!

edit: I noticed you pushed 3 commits yesterday. If you've addressed Jason's comments, please mark them as resolved. This will make it easier for someone to review.

studyztp:
> @studyztp, just a reminder to incorporate the changes Jason requested. Thanks!
> 
> edit: I noticed you pushed 3 commits yesterday. If you've addressed Jason's comments, please mark them as resolved. This will make it easier for someone to review.

Thank you for the reminder. The pushes were to address Jason's comments. I'll resolve them right now.

configs/example/gem5_library/x86-global-inst-tracker.py
studyztp: I fixed the style and added explanation of the script.

studyztp:
> A few small style things and some documentation requests. Overall, it looks good.

Thank you for the feedback! I addressed all of them and pushed the changes. I am really sorry for taking so long.

configs/example/gem5_library/x86-global-inst-tracker.py
```diff
@@ -0,0 +1,159 @@
...
+from gem5.simulate.simulator import Simulator
+
+cache_hierarchy = PrivateL1CacheHierarchy(
+    l1d_size="64kB",
+    l1i_size="64kB",
+)
+
+memory = SingleChannelDDR4_2400("1GB")
+
+processor = SimpleProcessor(cpu_type=CPUTypes.TIMING, num_cores=9, isa=ISA.X86)
```
BobbyRBruce, line 81: Having 9 cores seems odd to me. If theres a reason for this I think there should be a comment explaining it. If there isn't a reason can we do something more realistic, like 8 cores?

configs/example/gem5_library/x86-global-inst-tracker.py
```diff
@@ -0,0 +1,159 @@
...
+    # we can stop listening to instructions
+    for tracker in all_trackers:
+        tracker.stopListening()
+    # similarly, we can start listening to instructions again by calling:
+    #
+    # for tracker in all_trackers:
+    #     tracker.startListening()
+    m5.stats.dump()
+    m5.stats.reset()
+    yield False
```
BobbyRBruce: More of a comment/suggestion/idea for improvement.


This code highlights how scheduling exit events at particular instruction counts is hard. It looks a lot like how MAX_TICK was a few years ago with scripts needing to reset MAX_TICK to a new value when the old one was reached. In response to that problem I added new "to tick exits" functions : https://github.com/BobbyRBruce/gem5/commit/8479a691aa57c8f5763b71d26d579697ae96d007. This allowed users to specify a set of ticks to exit from in advance, not requiring it the exit to reset "MAX_TICK"/

It'd be a good contribution to do the same for instructions. Instead of resetting counters and setting thresholds there should be a `scheduleInstExitAtInstsFromCurrent(<int>)` allowing for the scheduling of an exist `<int>` instruction executions from the point of calling it and `scheduleInstExitAtAbsoluteInst(<int>)` allowing for the scheduling of an exit at the `<int>` instruciton executed since simulation began.

Most importantly this would allow for multiple exits to be scheduled at once allowing for something like:


```
scheduleInstExitAtAbsoluteInst(100000000) # First exit
scheduleInstExitAtAbsoluteInst(100000000 + 20000) # Second exit
scheduleInstExitAtAbsoluteInst(100000000 + 20000 + 20000) # Third exit

def inst_exit_handler():
    print("Reached 100000000 instructions executed")
    m5.stats.dump()
    m5.stats.reset()
    yield False

    print("Progressed a further 20000 instructions executed")
    m5.stats.dump()
    m5.stats.reset()
    yield False

    print("Progressed a further 20000 instructions executed. Time to finish")
    m5.stats.dump()
    m5.stats.reset()
    yield True
```

I think it'd make all this significantly cleaner.

configs/example/gem5_library/x86-global-inst-tracker.py
```diff
@@ -0,0 +1,159 @@
...
+        global_inst_tracker=global_inst_tracker,
+        # this parameter tells the tracker to start listening to instructions
+        # from the beginning of the simulation. If set to False, the tracker
+        # will
+        start_listening=True,
+    )
+    # we attach the tracker to the core
+    core.core.probeListener = tracker
+    # then store the tracker in the list
+    all_trackers.append(tracker)
```
BobbyRBruce, line 107: Could we just make this a property of all cores/processor in the stdlib? Would it be costly or unwise? With that we could just have a `get_inst_count` method for each core (i.e., for the number of instructions executed by that core) and a `get_inst_count` method in the processor (i.e, the "global" count across all its cores). Likewise both the cores and the processor can each have a "set_exit_at_inst" method which exits when their `get_inst_count` values reach a certain value.

configs/example/gem5_library/x86-global-inst-tracker.py
BobbyRBruce: Relating this back to my other comment regarding putting the tracker into the stdlib: you could have a function `add_exit_at_inst` instead of `set_exit_at_inst` to add an exit, not just change the value of the next exit.

I.e.:

```
processor.add_exit_at_inst(100) 
processor.add_exit_at_inst(200)
```

This would exit at both 100 and 200.

BobbyRBruce (changes requested):
Really great contribution here. I'm not sure if my comments here are wise so they are very much a "what do you think of this?" sort of thing rather than a direct request for changes. I think they are good ideas but I also accept there may be good reasons you don't do things this way. Let me know what you think.

configs/example/gem5_library/x86-global-inst-tracker.py
powerjg: You have to have (number of threads) + 1 cores in SE mode. So, if the workload has 8 threads, then you have to have 9 cores.

studyztp: If I record correctly, using openmp for SE mode needs a core to handle the scheduler, that's why it is using 9 cores instead of 8 cores here. Adding comments is a good idea. I'll add comments to explain this here.

configs/example/gem5_library/x86-global-inst-tracker.py
studyztp: I understand the importance of having the exit events to be scheduled easily. I'm not against adding `scheduleInstExitAtInstsFromCurrent(<int>)` and `scheduleInstExitAtAbsoluteInst(<int>)`. But just a personal design decision, I prefer to keep the C++ side robust and let the Python side to handle all the decisions. I have a crazy idea. Since scheduling a list of exits is a common use case, is it possible to wrap it inside the standard library in `simulator.py`(https://github.com/gem5/gem5/blob/stable/src/python/gem5/simulate/simulator.py)? 
For example, we can have a functions `scheduleInstExitAtInstsFromCurrent(<int>)` and `scheduleInstExitAtAbsoluteInst(<int>)`  that take in the exit list from the user, and add it to the `_schedule_exits` dict to be called before the exit event generator/functions.

In config script:
```
scheduleInstExitAtAbsoluteInst(100000000) # First exit
scheduleInstExitAtAbsoluteInst(100000000 + 20000) # Second exit
scheduleInstExitAtAbsoluteInst(100000000 + 20000 + 20000) # Third exit
```

In `simulator.py`:
```
def rescheduleGlbalInstTracker(inst):
    self._global_inst_tracker.changeThreshold(inst)

def scheduleInstExitAtInstsFromCurrent(inst):
    if not self._global_inst_tracker:
            raise.ValueError("global inst tracker not setup")
    self.scheduleInstExitAtAbsoluteInst(inst + self._global_inst_tracker.getCurrentInst())

def scheduleInstExitAtAbsoluteInst(inst):
    if self._schedule_exits[MAX_INSTS]:
        self._schedule_exits[MAX_INSTS].add(inst)
        self._schedule_exits[MAX_INSTS].sort()
    else:
        if not self._global_inst_tracker:
            raise.ValueError("global inst tracker not setup")
        self._schedule_exits[MAX_INSTS] = {
            "param_list" = set(inst),
            "function" = rescheduleGlbalInstTracker
        }

# in the while loop in the run() function check for if need to reschedule (https://github.com/gem5/gem5/blob/stable/src/python/gem5/simulate/simulator.py#L732)

if self.__schedule_exits[exit_enum]:
    function = self._schedule_exits[exit_enum]["function"]
    param = self._schedule_exits[exit_enum]["param"].pop()
    function(param)
    if len(self._schedule_exits[exit_enum]["param"]) == 0:
        del self._schedule_exits[exit_enum]
```

The downside of this is that it's completely rely on the ExitEvent enum. It's a crazy idea. I'm open to any opinions.

configs/example/gem5_library/x86-global-inst-tracker.py
studyztp: For one core, it has an existing function to check the instruction counts (https://github.com/gem5/gem5/blob/stable/src/cpu/BaseCPU.py#L73) and a way to set an exit event at instructions (https://github.com/gem5/gem5/blob/stable/src/cpu/BaseCPU.py#L72). 
However, there isn't a way to schedule an instruction exit event for multiple cores. This PR is to provide a way for the multiple cores' scenario. We can have `set_exit_at_inst` method exits in both core level and processor level if they call different functions to schedule the exit event. 
I don't have a strong opinion on this.

BobbyRBruce: I think the approach I just outlined is better. I think logically asking the processor how many instructions have been executed, I would expect the sum over all the cores within it. Likewise, having the cores and the processor be the ones that trigger the exit events when some "max inst" condition is met feels cleaner to me for the user. 

My only question is if this somehow adds a big overhead to simulation time. I feel like it doesn't, but I don't know for sure.

configs/example/gem5_library/x86-global-inst-tracker.py
BobbyRBruce: Oh, thank you. I forgot about this restriction. Yes a small comment to remind people like me would be appreciated. Thank you.

configs/example/gem5_library/x86-global-inst-tracker.py
BobbyRBruce: It's not a bad idea but I am against it because I don't think the Simulator should be responsible for setting up/scheduling exit events. We've done this before but I don't think it was a good idea. Eventually, somehow, to do this, you need the Simulator to inspect the simulated system and modify or "do something" to get it working. IN your example, it needs to know where to attach the probe listener, or at which cores you want to be counting, what sum over which cores you want to your global count, etc.

I think the Simulator module should be as agnostic as possible to the system it's simulating and should only be responsible for running the simulated system and handling the exit events accordingly. So, in this case I want it capable of receiving a "to instruction" exit events but I don't want it to be responsible for setting them or ordering them. 

Finally, I think logically it should be the cores that keep track of the instructions they have executed and where the user goes to state they want exit events to occur after X and Y instructions are executed. This means a core can be placed anywhere in the system, in any configuration, in any way, and it doesn't "break" the functionality of this exit events.

BobbyRBruce (commented):
I'm coming more around to my idea of having this exit events scheduling functionality imbedded in the AbstractCore/AbstractProcessor

configs/example/gem5_library/x86-global-inst-tracker.py
studyztp: I see. I changed the input type of the GlobalInstTracker from a single threshold to a list of thresholds. I also added a new setter function to add new thresholds to the tracker. This will allow us to implement the scheduling functions in the stdlib. Additionally, I updated the config script to showcase the changes.

configs/example/gem5_library/x86-global-inst-tracker.py
studyztp: I think it can add significant overhead. In general, it adds more executions to every instruction committed. I believe there is a balanced solution for this interface, but we should move this discussion to the future PR for its stdlib interface.

studyztp:
> I'm coming more around to my idea of having this exit events scheduling functionality imbedded in the AbstractCore/AbstractProcessor

I think it's a good idea. I've updated the backend of the GlobalInstTracker. This should allow us to implement the functionalities in the stdlib. We can discuss more details about how the functionalities should work and interact in the future PR for its stdlib interface.

configs/example/gem5_library/x86-global-inst-tracker.py
BobbyRBruce: Ok, I can live with this now as it doesn't change or break and APIs, and we can do it my way (or some other alt) later without any issue.

BobbyRBruce (approved):
I'm fine for this to go.

PR #1437, opened 2024-08-08
DO_NOT_MERGE_YET: arch-arm: fix decoding of additional SIMD instructions in T32 mode 

arichardson:
This also includes the change from https://github.com/gem5/gem5/pull/1324, so please merge that one first.

I noticed that in T32 mode the instruction `vselge.f32 s2, s0, s2` was
being decoded as `vmuls.w f2, f0, f2`, resulting in completely wrong
results for the floating point computation. It appears the lower part
of the T32 decoder table is just a copy-paste of the one above that does
not match the tables in the Arm reference. I spent a lot of time looking
at https://developer.arm.com/documentation/ddi0597/2024-06/T32-Instructions-by-Encoding/32-bit?lang=en
and I believe that in this lower section of the decoding table (with
bits 15-9 all set to 1), the only remaining encodings that are viable
here are in the "System register access, Advanced SIMD, and floating-
point" category, more specifically "Additional Advanced SIMD and
floating-point instructions".

Change-Id: Ifa98cf24c75a2c2831ec6e77c2879bd18638d204

### DIFF
--- a/src/arch/arm/isa/bitfields.isa
+++ b/src/arch/arm/isa/bitfields.isa
@@ -161,6 +161,7 @@ def bitfield HTS            hts;
 
 def bitfield LTOPCODE_15    ltopcode15;
 def bitfield LTOPCODE_11_8  ltopcode11_8;
+def bitfield LTOPCODE_11    ltopcode11;
 def bitfield LTOPCODE_7_6   ltopcode7_6;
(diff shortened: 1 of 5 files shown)


erin-le:
This PR is being closed for inactivity, but feel free to reopen it if you would like to work on it further!

PR #1438, opened 2024-08-08
arch-arm: downgrade a warning to a DPRINTF

arichardson:
Programming an event ID while counters are disabled is perfectly fine, so we should just log this using DPRINTF instead of printing a warn() every time it happens.

Change-Id: Ib9499857271033ef941f74a7f012d8694328eaf3

### DIFF
--- a/src/arch/arm/pmu.cc
+++ b/src/arch/arm/pmu.cc
@@ -543,8 +543,9 @@ PMU::CounterState::detach()
         sourceEvent->detachEvent(this);
         sourceEvent = nullptr;
     } else {
-        debugCounter("detaching event not currently attached"
-            " to any event\n");
+        DPRINTFS(PMUVerbose, &pmu,
+                 "detaching event %d not currently attached to any event"
+                 " [counterId = %d]\n", eventId, counterId);
     }
 }
 


src/arch/arm/pmu.cc
```diff
@@ -543,8 +543,9 @@ PMU::CounterState::detach()
         sourceEvent->detachEvent(this);
         sourceEvent = nullptr;
     } else {
-        debugCounter("detaching event not currently attached"
-            " to any event\n");
+      DPRINTFS(PMUVerbose, &pmu,
+          "detaching event %d not currently attached to any event [counterId "
+          "= %d, sourceEvent = %#x]\n", eventId, counterId, sourceEvent);
```
giactra: Not a big deal, but why are you printing sourceEvent if it is equal to nullptr?

giactra (commented):
This LGTM, one minor thing

src/arch/arm/pmu.cc
arichardson: Oh yeah that does not make any sense - I just copied the previous format without thinking about that. Will fix.

ivanaamit: @arichardson can you push this fix? Thanks.

arichardson: sorry for the delay, fixed now.

giactra (approved):

PR #1436, opened 2024-08-08
Added implementation of IPCP prefetcher

Setu-Gupta:
This PR adds the IPCP prefetcher described in [this ](https://ieeexplore.ieee.org/document/9138971) paper.

**NOTE: This contribution is made by @AbhishekUoR and @xmlizhao**

On branch ipcp
Changes to be committed:
	modified:   src/mem/cache/prefetch/Prefetcher.py
	modified:   src/mem/cache/prefetch/SConscript
	new file:   src/mem/cache/prefetch/ipcp.cc
	new file:   src/mem/cache/prefetch/ipcp.hh

Change-Id: Ie52a9602719773248aa17e9bc3d98203d9b1e23b

PR #1441, opened 2024-08-08
Added implementation of IMP

### LINKED ISSUE #1050: Indirect Memory Prefetcher only performs a single indirect prefetch even if the prefetch distance is bigger than 1

Setu-Gupta:
This PR adds a new implementation for the [Indirect Memory Prefetcher](https://pages.cs.wisc.edu/~yxy/pubs/imp.pdf). The older version of IMP does not work because it has two major flaws:
1. It's stream detection logic is incorrect
2. It can only perform prefetches of the form $$A[B[i]]$$ if it knows the _value_ of $$B[i]$$. However, if it does not know this value, it cannot issue indirect prefetches. This also means that it cannot prefetch for the future accesses of the type $$A[B[i+delta]]$$ with $$delta > 0$$. Therefore, it suffers from timeliness issues.

This PR introduces a new implementation of IMP which solves both these issues and fixes various other bugs. This PR will also fix Issue #1050 

**Note: This PR is the result of a joint collaboration with @AbhishekUoR and @xmlizhao**

Change-Id: Id6c66bc73bc1ff78087f61f47ec3ee237c332b85


powerjg:
How are you testing this code? Having some way to test the code and see that its behaving correctly would be helpful in reviewing the code.

Setu-Gupta:
> How are you testing this code? Having some way to test the code and see that its behaving correctly would be helpful in reviewing the code.

I used the following micro-benchmark to test the code. On this workload, both the Stride and the Tagged prefetchers show a performance detriment over the baseline of no prefetcher.  The older version of IMP also shows a performance detriment. On the other hand I see a slight improvement with IMPv2.

```
/*
 * Copyright (c) 2024 Samsung Electronics
 * All rights reserved.
 *
 * Redistribution and use in source and binary forms, with or without
 * modification, are permitted provided that the following conditions are
 * met: redistributions of source code must retain the above copyright
 * notice, this list of conditions and the following disclaimer;
 * redistributions in binary form must reproduce the above copyright
 * notice, this list of conditions and the following disclaimer in the
 * documentation and/or other materials provided with the distribution;
 * neither the name of the copyright holders nor the names of its
 * contributors may be used to endorse or promote products derived from
 * this software without specific prior written permission.
 *
 * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
 * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
 * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
 * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
 * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
 * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
 * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
 * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
 * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
#include <stddef.h>
#include <stdint.h>
#include <stdlib.h>
#include <stdio.h>
#include <time.h>

#define SIZE (1ULL << 26)
#define ITERS (1ULL << 16)

typedef struct
{
        uint32_t x;
        volatile uint32_t padding[15];
} value_t;

int main()
{
        static size_t* index_arr;
        index_arr = (size_t*)malloc(ITERS * sizeof(size_t));

        static volatile value_t* array;
        array = (value_t*)malloc(SIZE * sizeof(value_t));

        srand(time(NULL));
        for(size_t i = 0; i < ITERS; i++)
                index_arr[i] = rand() % SIZE;

        for(size_t i = 0; i < SIZE; i++)
                array[i].x = i;

        uint32_t accumulator = array[0].x;
        for(size_t i = 0; i < ITERS; i++)
        {
                asm volatile("dmb sy" ::: "memory");
                accumulator += array[index_arr[i]].x;
        }

        free((void*)index_arr);
        free((void*)array);

        return 0;
}
```

Note that this was evaluated on the ARM model. The `dmb sy` instruction is a [memory fence](https://developer.arm.com/documentation/dui0489/c/arm-and-thumb-instructions/miscellaneous-instructions/dmb--dsb--and-isb) which prevents the memory accesses from being reordered. Without the fence we observe multiple accesses to the `index_array` before an access to the outer indirect array, `array`. Since the [IMP algorithm](https://pages.cs.wisc.edu/~yxy/pubs/imp.pdf) relies on the ordering of the index accesses and indirect accesses to be maintained, without the memory fence, IMP cannot learn anything.

powerjg:
> On this workload, both the Stride and the Tagged prefetchers show a performance detriment over the baseline of no prefetcher. The older version of IMP also shows a performance detriment. On the other hand I see a slight improvement with IMPv2.

The microbenchmark is great! However, it is still pretty complex to figure out what the expected result should be.

What is the expected performance improvement? How do we know that the IMPv2 is working as expected?

If you could say something like "with IMPv2 there should be about 10,000 prefetch requests and most of them should hit after accounting for cold misses" it would be quite helpful.

Setu-Gupta:
> What is the expected performance improvement? How do we know that the IMPv2 is working as expected?

We expect to see high prefetch accuracy with IMPv2, which we do indeed (around 99%). This indicates that the indirect pattern detection logic (i.e. the base address calculation logic) is functioning as expected.

> If you could say something like "with IMPv2 there should be about 10,000 prefetch requests and most of them should hit after accounting for cold misses" it would be quite helpful.

It's a bit difficult to make any comments about how many prefetches made by IMP would actually turn out to be useful. The issue is that IMP, due to its algorithm, suffers from a timeliness problem. When we access memory by using constructs like `A[B[i]]`, we end up performing two reads. The problem is that these two reads happen very close in time. Since IMP needs the _value_ of `B[i]` to calculate the _address_ of `A[B[i]]`, it cannot issue prefetches in advance for the array `A`. The paper also doesn't say a lot about how to handle this problem. The authors just say that instead of issuing prefetches for `A[B[i]]`, IMP will issue prefetches for `A[B[i+delta]]` for some delta > 0. However, the paper does not mention exactly how IMP is supposed to the read the _value_ `B[i+delta]`. The way we handle this in our PR is by using the `notifyFill` method. We store all the context required to issue the prefetches for `A[B[i+delta]]` and issue a prefetch for `B[i+delta]`. When we get a cache fill for this prefetch, we then compute the _address_ for `A[B[i+delta]]` in `notifyFill`. To be honest with you, I'm not entirely sure if this implementation matches the intent of the authors or not, but this was the cleanest way that we could think of implementing this logic.

AbhishekUoR:
@ivanaamit, Is there a problem with this PR? How long should we expect it to be reviewed?

src/mem/cache/prefetch/indirect_memory_v2.hh
```diff
@@ -0,0 +1,662 @@
...
+#include <utility>
+#include <vector>
+
+#include "mem/cache/prefetch/queued.hh"
+#include "mem/packet.hh"
+#include "mem/packet_access.hh"
+#include "params/IMPv2Prefetcher.hh"
+#include "sim/system.hh"
+
+namespace gem5
```
odanrc, line 80: The code needs to be fixed to follow gem5´s coding style: https://www.gem5.org/documentation/general_docs/development/coding_style/ . There are many things that I see right off the bat, such as (but not only):
- “namespaces should not increase the indentation“;
- “Class member names (method and variables, including const variables) are mixed case, start with a lowercase letter, and do not contain underscores“;
- “For control blocks (if, while, etc.), opening braces must be on the same line as the control keyword with a space between the closing parenthesis and the opening brace.“

There used to be a pre-commit around that fixed some stuff, but I don´t know if it is still there.

This comment applies to all PRs.

src/mem/cache/prefetch/indirect_memory_v2.hh
```diff
@@ -0,0 +1,662 @@
...
+{
+    namespace prefetch
+    {
+        namespace IMPv2Internals
+        {
+            template<typename T>
```
odanrc, line 86: Use src/base/sat_counter.hh instead

src/mem/cache/prefetch/indirect_memory_v2.hh
```diff
@@ -0,0 +1,662 @@
...
+                        IPDEntry()
+                        {
+                            index1 = std::nullopt;
+                            index2 = std::nullopt;
+                            recorded_misses = 0;
+                            found_params = false;
+                            base_address_candidates.clear();
+                        }
+
+                        ~IPDEntry()
```
odanrc, line 364: Remove this destructor declaration, as what it does is already done automatically: https://stackoverflow.com/a/12068989

src/mem/cache/prefetch/indirect_memory_v2.hh
```diff
@@ -0,0 +1,662 @@
...
+#include "mem/packet.hh"
+#include "mem/packet_access.hh"
+#include "params/IMPv2Prefetcher.hh"
+#include "sim/system.hh"
+
+namespace gem5
+{
+    namespace prefetch
+    {
+        namespace IMPv2Internals
```
odanrc, line 84: These belong in the IMP class. Do they really need to be exposed publicly? If not, then move them inside the IMP class.

src/mem/cache/prefetch/indirect_memory_v2.hh
```diff
@@ -0,0 +1,662 @@
...
+                         */
+                        std::vector<std::vector<Addr>> base_address_candidates;
+
+                        IPDEntry()
+                        {
+                            index1 = std::nullopt;
+                            index2 = std::nullopt;
+                            recorded_misses = 0;
+                            found_params = false;
+                            base_address_candidates.clear();
```
odanrc, line 361: nit: no need to do this clear, as this is what the default constructor does

src/mem/cache/prefetch/indirect_memory_v2.hh
```diff
@@ -0,0 +1,662 @@
...
+                         * observed after index1 was set, until index2 got set.
+                         * The number of tracked cache misses, i.e. the range
+                         * of the inner iterator is equal to
+                         * num_miss_after_index
+                         */
+                        std::vector<std::vector<Addr>> base_address_candidates;
+
+                        IPDEntry()
+                        {
+                            index1 = std::nullopt;
```
odanrc, line 357: nit: in general it is better to use initializer lists or in-class initialization, for both performance and versatility: https://stackoverflow.com/questions/9903248/initializing-fields-in-constructor-initializer-list-vs-constructor-body

src/mem/cache/prefetch/indirect_memory_v2.hh
```diff
@@ -0,0 +1,662 @@
...
+        } // namespace IMPv2Internals
+
+        class IMP : public Queued
+        {
+            private:
+                IMPv2Internals::PrefetchTable prefetch_table;
+                IMPv2Internals::IndirectPatternDetector ipd;
+
+                /*
+                 * The degree for streaming and indirect prefetches
```
odanrc, line 452: nit: ´degree´ does not give much information for people that not very familiar with prefetchers. Do you mind changing the documentation to something more informative? It still makes sense to call the variable a degree, but I like how the documentation was declared in old IMP:  /** Number of prefetches generated when using the streaming prefetcher */

src/mem/cache/prefetch/indirect_memory_v2.hh
```diff
@@ -0,0 +1,662 @@
...
+            }
+
+            template<typename T>
+            void saturating_decrement(T& old_value)
+            {
+                if (old_value > 0)
+                    old_value -= 1;
+            }
+
+            /*
```
odanrc, line 100: Could you change the documentation to Doxygen? For reference: https://www.doxygen.nl/manual/docblocks.html
i.e., it should start with /**, not /*, and parameters are declared as:
@param name_of_param description
@return The description of the thing being returned

This comment applies to all PRs.

src/mem/cache/prefetch/indirect_memory_v2.hh
```diff
@@ -0,0 +1,662 @@
+/**
```
odanrc, line 1: I think that this copyright will not suffice. Please check LICENSE for a template of the license we use in gem5. Some companies require adding some extra information (for example, check src/arch/arm/decoder.hh), so you may want to check with the people responsible for the copyright info at Samsung.

odanrc (changes requested):
That is some great amount of documentation. Thanks for taking the time to write all that! I haven´t even started reviewing the algorithm, as I still need to re-read the paper to remember how it was supposed to work, but the review feedback I give here is also useful to the other PRs you have uploaded, as they mostly refer to gem5´s coding style, as well as some C++ recommended practices.

odanrc:
> @ivanaamit, Is there a problem with this PR? How long should we expect it to be reviewed?

I would say that since these PRs are extremely specific, and require both knowledge of the prefetching framework, and the paper. I wouldn´t expect for these PRs to be reviewed quickly. Each one of them will require the reviewer to read the papers, and then understand and review the code, a task that will take many hours.

Could you tell me why did you opt to implement from scratch instead of fixing the original implementation? Was it unsalvageable with respect to the bugs?

giactra:
> Could you tell me why did you opt to implement from scratch instead of fixing the original implementation? Was it unsalvageable with respect to the bugs?

That's the first thing I thought when I saw the PR.
Anyway if the initial implementation of IMP is broken we should probably remove it

powerjg:
> Anyway if the initial implementation of IMP is broken we should probably remove it

After reviewing is mostly done, I would suggest adding a new commit which deletes the old implementation and renames this implementation.

> Each one of them will require the reviewer to read the papers, and then understand and review the code, a task that will take many hours.

Yes, exactly. It's difficult for (mostly "volunteer") reviewers to find the time and build the expertise to review. If there's anything @Setu-Gupta can do to make the reviewing easier, it would be greatly appreciated!

Setu-Gupta:
> Could you tell me why did you opt to implement from scratch instead of fixing the original implementation? Was it unsalvageable with respect to the bugs?

Yes, I did feel that the older code was more or less unsalvageable. There were many fundamental issues with the older implementation which would have required complete re-writes of various parts of the prefetcher. Since to fix the code I anyways would have had to re-write everything, I felt that it would be more easier and cleaner to start from scratch. This way I could structure the code in a way which I believe is easier to understand, and closer to _my_ interpretation of the author's intents.

> Anyway if the initial implementation of IMP is broken we should probably remove it

Yes, I agree. If this PR gets accepted, I believe the old version should be removed to avoid any confusion.

> After reviewing is mostly done, I would suggest adding a new commit which deletes the old implementation and renames this implementation.

Yes, I'll do that once the reviews are done. Also, I would like, if possible, for one or more of the original authors to go over the code. I know that this is probably impossible since they are now professors and they probably won't have the time, but if possible, I would like someone who is more knowledgeable than me to go over this implementation. The reason I say this is because the paper, at least in my opinion, doesn't go into _all_ the details about how IMP works. For instance, it says that IMP reads the index value, but it never says how. Due to these missing pieces in the paper, I had to _assume_ what the authors' intent was when they wrote the paper, However, my assumption(s) could be wrong. That's why I would feel more comfortable if someone who is more knowledgeable in IMP can have a quick look at my code.

> Yes, exactly. It's difficult for (mostly "volunteer") reviewers to find the time and build the expertise to review. If there's anything @Setu-Gupta can do to make the reviewing easier, it would be greatly appreciated!

I understand this issue. After all Gem5, just like most open source software, is maintained by volunteers and I'd like to thank you all for this! I can try to answer questions that people have about IMP based on _my_ interpretation of the paper. Also, I tried to write meaningful comments which explain my code so that the people who may not be aware of IMP can figure out how the code works. A good place to start would be the explanation of the algorithm in the [`src/mem/cache/prefetch/indirect_memory_v2.hh`](https://github.com/gem5/gem5/pull/1441/files#diff-9a7bd2464cb1d4c30c9b062ea470a6f9062601a8f1d49acdb00db4a76509afe4) file. Of course, I would be happy to answer questions about this implementation. Moreover, if people want to talk to me over Zoom or something regarding this implementation, I'm also open to that. They can email me and we can set up a time to discuss the implementation in detail. **Please let me know what else I can do to ensure that people don't face challenges in understanding this implementation!**

src/mem/cache/prefetch/indirect_memory_v2.hh
```diff
@@ -0,0 +1,662 @@
...
+                        streaming_threshold(threshold), stream_table_size(size)
+                    {}
+
+                    /*
+                     * Add a load address, addr, to the stream table.
+                     * The address was generated by a instruction which the IP
+                     * points to. Return the address delta if the confidence
+                     * is high enough for the stream associated with IP
+                     */
+                    std::optional<Addr> addToStreamTable(const Addr IP,
```
odanrc, line 148: This function adds an address to the table. I don´t think it makes sense for it to return an optional delta, as this is not related to the insertion itself. Instead, I think that this function should return void, and that inStreamTable´s return value should be changed so that it returns the entry.

src/mem/cache/prefetch/indirect_memory_v2.hh
```diff
@@ -0,0 +1,662 @@
...
+            }
+
+            /*
+             * The StreamDetector performs two functions:
+             *  1. It detects stream accesses to the indexing array, B
+             *  2. If it is confident about a stream, it returns the predicted
+             *     delta for that stream so that IMP may issue stream
+             *     prefetches using the delta
+             */
+            class StreamDetector
```
odanrc, line 107: Can you reuse the Stride prefetcher instead of implementing this class? IIRC, a stream prefetcher is a stride prefetcher where the degree and the distance are 1, right? Then to get the outcome of inStreamTable you would instead call calculatePrefetch.

src/mem/cache/prefetch/indirect_memory_v2.hh
```diff
@@ -0,0 +1,662 @@
...
+                     * The confidence threshold above which a stream is
+                     * classified as detected
+                     */
+                    const uint64_t streaming_threshold;
+
+                    /*
+                     * The stream table is indexed using the IP of the load
+                     * instruction
+                     */
+                    std::map<Addr, StreamTableEntry> stream_table;
```
odanrc, line 134: nit: ideally this would use the associative cache implementation, as it would allow users to change the replacement and indexing policies as desired. The other prefetchers take this approach, so it would be nice to see it here too. I am not pushing for it, though, as I think this class as a whole should be replaced by the stride prefetcher (either in this PR, or in the future).

odanrc (changes requested):

src/mem/cache/prefetch/indirect_memory_v2.hh
Setu-Gupta: The saturating counter implemented in `src/base/sat_counter.hh` would not work in this case since that counter's maximum value must be of the form $2^n - 1$ where $n$ is the number of bits in the counter. In this implementation, however, the maximum values of the counter variables are not _always_ of the form $2^n -1$. For instance the value of the `streaming_threshold`, which serves as the upper bound on the value of `StreamTableEntry::confidence`, is actually 5 (look at `src/mem/cache/Prefetcher.py:306`).

src/mem/cache/prefetch/indirect_memory_v2.hh
Setu-Gupta: The reason why I have kept these outside the IMP class is because these functions are not specific to just `IMP`. These are general functions which are used to modify saturating counters which are used in various structures of IMP. Making these functions a part of the IMP class gives the false impression of these functions being associated with an IMP object, which they are not. These are just helper functions.

src/mem/cache/prefetch/indirect_memory_v2.hh
Setu-Gupta: The `addToStreamTable` function performs two jobs:
1. It adds an entry to the stream table
2. It checks if the added entry's confidence was high enough, and if it was, it returns the associated stride delta.
However, I do agree that this function can be renamed to indicate the two things that it does. Unfortunately I can't think of any concise name for this function which would encapsulate both these tasks, Do you have any suggestions?

The `inStreamTable` function does not modify the state of the Stream Detector. All that this function does is to answer whether or not a given address was predicted by the Stream Detector. Since we don't want any other information about the Stream Table entries from this function, I chose not to return a pointer to the entry, but rather simply return a boolean. This function is then used by the main `IMP` class to classify accesses as index accesses and indirect accesses. Index accesses, i.e. access to the array `B` if we are tracking `A[B[i]]`, should get the `true` from this function, whereas accesses to the indirect array `A` should get `false`.

src/mem/cache/prefetch/indirect_memory_v2.hh
Setu-Gupta: Yes, you are correct that a stream prefetcher is nothing by a stride prefetcher with delta = 1. In fact what we have in IMP is not a stream detector, but rather a stride detector. The reason why I named the class `StreamDetector` was because that's the terminology that the paper uses.

The stride prefetcher implementation in Gem5 does a bit more than this simplistic stride detector. That's the reason why I chose to go with a custom implementation.

src/mem/cache/prefetch/indirect_memory_v2.hh
Setu-Gupta: The prefetch table (which is divided into two distinct parts, the `stream_table` and the `indirect_table`) is a direct mapped structure, which is indexed by the IP, as per the paper's description (see Figure 5 of the [paper](https://pages.cs.wisc.edu/~yxy/pubs/imp.pdf)). Since this structure is a direct mapped structure, converting it into an associative cache would lead to more confusion. So, I think we should leave it as it is. What do you think?

src/mem/cache/prefetch/indirect_memory_v2.hh
odanrc: I think it would be a simple fix to make the sat counter more generic. The only place where we enforce being a power of 2 is the fact that the constructor uses the number of bits to initialize the max val. We can simply create another constructor that accepts a max val. The only issue is that the signatures will be conflicting, but that should be easy to solve with dummy params. It is better to add functionality that can be easily unit tested than reimplementing it from the scratch due to such a small change.

src/mem/cache/prefetch/indirect_memory_v2.hh
odanrc: I have a strong opinion against duplicating code, specially given that I spent years deduplicating stuff in the cache directory, so I will double down on this request.

I understand that it does more stuff, but isn´t there a combination of parameters that ends up generating the addresses that you want? I am leaning towards thinking that the extra stuff will just slightly degrade performance, but won´t do something irreversible to the state, or end up generating different addresses.

src/mem/cache/prefetch/indirect_memory_v2.hh
odanrc: I tend to think that since gem5 is used by researchers, we want to make their lives easier in case they want to expand existing state-of-the-art algorithms. As such, I think that using a generic cache will be better, even if the original paper uses a specific configuration. That said, as I said this request is is a nitpick, and can be ignored without a problem.

src/mem/cache/prefetch/indirect_memory_v2.hh
odanrc: For me that is exactly the issue: it performs two jobs. An alternative would be to add one or two more functions to do the other jobs (e.g., one to check if the entry with the given addr has a high-enough confidence, and the other to get the entry´s delta). The goal is to make the code more easily understandable, at the cost of slightly worse performance (due to having to do an extra lookup).

odanrc (changes requested):

Harshil2107:
Hi @AbhishekUoR and @Setu-Gupta,

Thank you for your contribution! Could you please update us on the status of this PR? I noticed that PRs #1461, #1454, and #1441 seem to be related and add prefetchers. It would be helpful if you could create a discussion post to track the status of these PRs and inform the community about the changes. Additionally, we have a gem5 developer's meeting next Thursday, October 10th. It would be great if you could join us to discuss these updates.

Setu-Gupta:
Hi Harshil

Thanks a lot for the invitation to the developers' meeting but
unfortunately both Abhishek and I won't be able to make it due to time
constraints.

Thanks and Regards
Setu

erin-le:
@Setu-Gupta Hello, I noticed that you force-pushed this PR last month. Are you still actively working on this change? If not, would you be okay with closing this PR and reopening it when you want to continue working on it?

Setu-Gupta:
Hi Erin

Yes, I think that'll be better since I'm not sure when I will be able to
find time to work on it.

Thanks
Setu

AbhishekUoR:
@Setu-Gupta could you please let me know what changes are required and I will make those happen in upcoming weeks?

Setu-Gupta:
Hi @AbhishekUoR 

Here's the list of stuff that we need to do:
1. We need to re-format and re-verify the implementation to make it conform to Gem5's coding style.
2. We need to replace our saturating counter implementation with the one which is already in gem5, i.e. the one in `src/base/sat_counter.hh`. Note that since our implementation does _not_ necessarily cap the counter values at $2^n -1$ for some $n$, we'll have to modify Gem5's implementation instead.
3. We need to rename and/or remove some functions to clean up the implementation. Look at @odanrc's comments for more details.

Thanks
Setu

erin-le:
@AbhishekUoR Hello, just wanted to ask about the current status of this PR. Thank you!

AbhishekUoR:
Will try to circle back on this within two weeks

erin-le:
@AbhishekUoR: Hello, could you give another update on the status of this PR? Thank you!

AbhishekUoR:
Yes will do this weekend
Thank you for reminding me

Thanks,
Abhishek Singh


Remember "Chance favors the prepared mind."(Pasteur)

erin-le:
@AbhishekUoR: Hello, could you give an update on the status of this PR? Thank you!

AbhishekUoR:
I need to take a look at this on what is remaining. Give me a week

erin-le:
@AbhishekUoR: Hello, could you give another update on the status of this PR?

erin-le:
@AbhishekUoR: Hello, do you have any new updates on the status of this PR?

AbhishekUoR:
Not yet. I will require 2-3 weeks to work on it as I have other deadlines

Thanks,
Abhishek Singh


Remember "Chance favors the prepared mind."(Pasteur)

erin-le:
@AbhishekUoR: Hello, since it looks like you don't have much time to work on this PR, would you be okay with closing this PR, https://github.com/gem5/gem5/pull/1436, and https://github.com/gem5/gem5/pull/1461? They can be reopened later if you want to continue working on them.

erin-le:
This PR is being closed for inactivity. If you'd like to continue working on it, however, please feel free to reopen it. Thank you!

PR #1445, opened 2024-08-09
arch-vega: Swizzle multi-dword scratch requests

abmerop:
Scratch memory requests that are larger than one dword are using a different memory layout than global instructions. Rather than being placed contiguously, each dword is interleaved 64 lanes * 4 bytes away as described in Section 9.1.5.2. "Swizzled Buffer Addressing" in the MI300 specification. This was verified by comparing MI300 output (which uses scratch_ instructions) with MI200 (which uses buffer instructions). MI300 FashionMNIST bs=1 now matches CPU reference.

This requires several changes to the instruction implementations:
 - For stores, data in the GPUDynInst can be swizzled before the data is written to memory. This is easy to do using a helper method. This is done in the template<int N> variant of initMemWrite. To use this x2 stores are changed to use template<int N> rather than loading a U64. The swizzle function is renamed to swizzleAddr to avoid confusion with swizzleData.
 - For loads, data is unswizzled in completeAcc when writing register values. This is not as easy to implement as a helper and is thus implemented for the three load instructions that load more than one dword.
 - Accessing swizzled data requires at least one packet per dword. A new GPU memory helper is added to create these packets for scratch requests specifically. This is called in the template<int N> variant of initMemRead / initMemWrite. Loads and stores of x2 are changed to use this variant instead of accessing a U64.

The GPUDynInst status vector restrictions are increased to allow for swizzled x4 accesses. For simplicity this does not currently support misaligned swizzled accesses and will panic upon seeing such a case.

Change-Id: Ic686c51e28e0af029a043d5a5b3d4069f2cb94f9

### DIFF
--- a/src/arch/amdgpu/vega/gpu_mem_helpers.hh
+++ b/src/arch/amdgpu/vega/gpu_mem_helpers.hh
@@ -129,6 +129,55 @@ initMemReqHelper(GPUDynInstPtr gpuDynInst, MemCmd mem_req_type,
     }
 }
 
+template<int N>
+inline void
+initScratchReqHelper(GPUDynInstPtr gpuDynInst, MemCmd mem_req_type)
+{
+    int req_size = N * sizeof(VegaISA::VecElemU32);
+    int block_size = gpuDynInst->computeUnit()->cacheLineSize();
+
+    gpuDynInst->resetEntireStatusVector();
+    for (int lane = 0; lane < VegaISA::NumVecElemPerVecReg; ++lane) {
+        if (gpuDynInst->exec_mask[lane]) {
+            Addr vaddr[N];
+
+            for (int dword = 0; dword < N; ++dword) {
+                int stride = VegaISA::NumVecElemPerVecReg
+                           * sizeof(VegaISA::VecElemU32);
+                vaddr[dword] = gpuDynInst->addr[lane] + dword * stride;
+
+                // Do not allow misaligned for simplicity for now.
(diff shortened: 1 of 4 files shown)


src/arch/amdgpu/vega/gpu_mem_helpers.hh
```diff
@@ -129,6 +129,55 @@ initMemReqHelper(GPUDynInstPtr gpuDynInst, MemCmd mem_req_type,
...
+inline void
+initScratchReqHelper(GPUDynInstPtr gpuDynInst, MemCmd mem_req_type)
+{
+    int req_size = N * sizeof(VegaISA::VecElemU32);
+    int block_size = gpuDynInst->computeUnit()->cacheLineSize();
+
+    gpuDynInst->resetEntireStatusVector();
+    for (int lane = 0; lane < VegaISA::NumVecElemPerVecReg; ++lane) {
+        if (gpuDynInst->exec_mask[lane]) {
+            Addr vaddr[N];
```
mattsinc, line 142: Do we need to check in here if a lane is active?

abmerop: Yes, It is checking exec mask here. For swizzled accesses there will be multiple `vaddr`s per-lane so it looks a bit different than global accesses

mattsinc: I must have missed it, thanks.

mattsinc (approved):

Issue #1429, opened 2024-08-07
Further generalize the AssociativeCache


giactra:
**Generic Cache library**
The AssociativeCache introduced in https://github.com/gem5/gem5/pull/745 models a base associative storage managed by our existing indexing and replacement policies. At the moment the AssociativeCache is mainly used by several caching agents in the memory subsystem (e.g. prefetchers). One of the known limitations (as addressed in the aforementioned PR) is the lack of support for non Addr (uint64_t) tags.
In other words the data structure APIs assume the entry template parameter is tagged with a single unsigned integer. This restricts the usage for other caching agents which for example use more complex tags (like TLB entries which are tagged based on a set of individual fields, like address, ASID etc).

@andysan suggested the following:

> It would be a huge improvement if you could make the tag and lookup keys template parameters, but I don't know if this would be feasible in this PR though. It might have a large knock-on effect on the indexing policies.
> 
> I think what we need to truly generalise this is a way to specify the following:
> 
> - A lookup type. This could be a struct or Addr
>  - A hash function that works on the lookup type. Modern caches don't always use the text-book shift and mask approach. This may need to be a parameter.
> - A comparison operator. In practice, we should probably store the entire lookup type as a tag instead of throwing away bits to save space. This would effectively require an operator== for the lookup type which would just work for many types.
> 
> As a concrete example, this would let you define a cache using the following pseudo-code:
> 
> struct CacheKey {
>     uint16_t vmid;
>     uint16_t asid;
>     bool ns;
>     Addr pa;
> };
> 
> uint64_t
> my_hash(const CacheKey &key)
> {
>     return (key.pa >> bits) & mask_bits;
> }
> 
> AssociativeCache<CacheKey, my_hash> my_cache;
> 
> You'd then perform lookups using instances of CacheKey.

As has already been noted, one of the difficulties of implementing this is not in the cache per se, but in the interaction with the indexing policy SimObject , which is extracting the set number based on an unsigned integer [1]. We would either have to deal with templated SimObjects or trying to find some other workaround

[1]: https://github.com/gem5/gem5/blob/stable/src/mem/cache/tags/indexing_policies/base.hh#L149

PR #1446, opened 2024-08-09
base, mem-cache: Make the AssociativeCache more generic

giactra:
This PR implements #1429. It mainly achieve so with the following changes

1) The IndexingPolicy is now a templated SimObject to make its APIs work with different data types.
As an example, look at the getPossibleEntries, which is requiring an Addr object whereas we want to be able to call the method with different keys depending on the Tag
2) The AssociativeCache extracts type information from the Entry template parameter.
This means any AssociativeCache entry will have to define the following types:

KeyType = This is the data type used for lookups (in its simplest case, it is Addr)
IndexingPolicy = This is the base indexing policy SimObject

As an example, the PR is also reworking the TaggedEntry to be AssociativeCache compliant. This
ultimately allows us to remove the weird overloading of cache querying methods with the secure flag, and to
remove the AssociativeSet which was providing such weird interface.
As mentioned in the [base, mem-cache: Rewrite TaggedEntry code](https://github.com/gem5/gem5/commit/7ee97904646a9e3b6a03c95a5a098d836a113336) commit, further cleanup is needed. TaggedEntry
is really a misleading name as its sole difference with the CacheEntry (which is also tagged) is the presence of
the secure bit. A better name should be chosen.


giactra:
@odanrc

src/mem/cache/tags/indexing_policies/base.hh
```diff
@@ -49,23 +49,28 @@
...
-class ReplaceableEntry;
-
 /**
  * A common base class for indexing table locations. Classes that inherit
  * from it determine hash functions that should be applied based on the set
  * and way. These functions are then applied to re-map the original values.
  * @sa  \ref gem5MemorySystem "gem5 Memory System"
  */
-class BaseIndexingPolicy : public SimObject
+template <class Types>
```
odanrc, line 71: nit (but pending towards a request, since this class will likely be used by many): add @tparam Types description to the class documentation

src/mem/cache/tags/indexing_policies/base.hh
```diff
@@ -156,10 +195,19 @@ class BaseIndexingPolicy : public SimObject
...
+struct AddrTypes
+{
+    using KeyType = Addr;
+    using Params = BaseIndexingPolicyParams;
+};
+
+using BaseIndexingPolicy = IndexingPolicyTemplate<AddrTypes>;
+template class IndexingPolicyTemplate<AddrTypes>;
+
 } // namespace gem5
```
odanrc, line 231: nit: Do you have something against instead of deleting the base.cc file renaming it to an _impl file and including it here?

src/mem/cache/tags/indexing_policies/set_associative.cc
```diff
@@ -63,14 +63,14 @@ SetAssociative::extractSet(const Addr addr) const
...
-                                                                        const
+SetAssociative::regenerateAddr(const Addr &tag,
+                               const ReplaceableEntry* entry) const
 {
     return (tag << tagShift) | (entry->getSet() << setShift);
 }
 
 std::vector<ReplaceableEntry*>
-SetAssociative::getPossibleEntries(const Addr addr) const
+SetAssociative::getPossibleEntries(const Addr &addr) const
```
odanrc, line 73: My understanding is that the secure bit is a part of the address - i.e., the address (A,0) is not the same as the address (A,1). As such, the secure bit should be used by the indexing policy too - i.e., these different addresses should be mapped to different locations. I always thought not having the secure bit in the IdxPol´s API was a bug. Am I misunderstanding something?

src/mem/cache/cache_blk.hh
```diff
@@ -545,10 +547,10 @@ class TempCacheBlk final : public CacheBlk
     }
 
     void
-    insert(const Addr addr, const bool is_secure) override
+    insert(const KeyType &tag) override
     {
-        CacheBlk::insert(addr, is_secure);
-        _addr = addr;
+        CacheBlk::insert(tag);
+        _addr = tag.address;
```
odanrc, line 552: Ditto: shouldn´t we store both the address and the secure bit for the TempCacheBlk?

src/base/cache/cache_entry.hh
```diff
@@ -127,6 +152,8 @@ class CacheEntry : public ReplaceableEntry
      */
     bool valid{false};
 
+    BaseIndexingPolicy *indexingPolicy{nullptr};
```
odanrc: I don´t like this dependency. I think it would be better to store an std::function extractTag instead - we don´t need access to any of the other indexing policy members here.

src/base/cache/associative_cache.hh
```diff
@@ -48,12 +60,14 @@ namespace gem5
 template <typename Entry>
 class AssociativeCache : public Named
 {
-    static_assert(std::is_base_of_v<CacheEntry, Entry>,
+  protected:
+
+    static_assert(std::is_base_of_v<ReplaceableEntry, Entry>,
                   "Entry should be derived from CacheEntry");
```
odanrc: Update error message to ReplaceableEntry

src/mem/cache/prefetch/indirect_memory.cc
```diff
@@ -172,7 +174,8 @@ IndirectMemory::allocateOrUpdateIPDEntry(
 {
     // The address of the pt_entry is used to index the IPD
     Addr ipd_entry_addr = (Addr) pt_entry;
-    IndirectPatternDetectorEntry *ipd_entry = ipd.findEntry(ipd_entry_addr);
+    const IndirectPatternDetectorEntry::KeyType key{ipd_entry_addr, false};
```
odanrc, line 177: Why do we use false for the secure bit here? I think we should use pt_entry->secure, but I am not sure

src/mem/cache/prefetch/pif.cc
```diff
@@ -176,16 +176,15 @@ PIF::notifyRetiredInst(const Addr pc)
                 // Insert the spatial entry into the history buffer and update
                 // the 'iterator' table to point to the new entry
                 historyBuffer.push_back(spatialCompactor);
-                constexpr bool is_secure = false;
-                auto idx_entry = index.findEntry(spatialCompactor.trigger,
-                is_secure);
+
+                const IndexEntry::KeyType key{spatialCompactor.trigger, false};
```
odanrc, line 180: This looks like a bug in the code: the CompactorEntry should keep track of the secure bit, as it is part of the address

src/mem/cache/prefetch/signature_path.cc
```diff
@@ -192,17 +194,17 @@ SignaturePath::getSignatureEntry(Addr ppn, bool is_secure,
 SignaturePath::PatternEntry &
 SignaturePath::getPatternEntry(Addr signature)
 {
-    constexpr bool is_secure = false;
-    PatternEntry* pattern_entry = patternTable.findEntry(signature, is_secure);
+    const PatternEntry::KeyType key{signature, false};
```
odanrc, line 197: Ditto looks like a bug: I think getPatternEntry and updatePatternTable should be passed a KeyType as an argument instead of Addr, and then in SignatureEntry we would also have a secure bit stored, so that we can use it in calculatePrefetch.

src/mem/cache/prefetch/stride.cc
```diff
@@ -107,19 +107,18 @@ Stride::allocateNewContext(int context)
...
     // Create new table
-    auto ins_result = pcTables.emplace(std::piecewise_construct,
-                           std::forward_as_tuple(context),
-                           std::forward_as_tuple(table_name.c_str(),
-                                                 pcTableInfo.numEntries,
-                                                 pcTableInfo.assoc,
-                                                 pcTableInfo.replacementPolicy,
-                                                 pcTableInfo.indexingPolicy,
-                                                 StrideEntry(initConfidence)));
+    pcTables[context].reset(new PCTable(
```
odanrc, line 112: nit: use make_unique instead of new

src/mem/cache/prefetch/stride.hh
```diff
@@ -136,16 +136,17 @@ class Stride : public Queued
...
         void invalidate() override;
 
         Addr lastAddr;
         int stride;
         SatCounter8 confidence;
     };
-    typedef AssociativeSet<StrideEntry> PCTable;
-    std::unordered_map<int, PCTable> pcTables;
+    using PCTable = AssociativeCache<StrideEntry>;
+    std::unordered_map<int, std::unique_ptr<PCTable>> pcTables;
```
odanrc, line 148: nit: what do you think about using shared_ptr and then updating the signatures of allocateNewContext and findTable to return shared_ptrs (or weak_ptr)? On one hand we would not be passing a non-const ptr arond, but on the other hand we don´t really need to give ownership to the caller (that is why in theory the weak_ptr would be preferable)

src/mem/cache/prefetch/spatio_temporal_memory_streaming.cc
```diff
@@ -91,16 +92,14 @@ STeMS::checkForActiveGenerationsEnd(const CacheAccessor &cache)
                 }
             }
             if (generation_ended) {
+                const ActiveGenerationTableEntry::KeyType key{pst_addr, false};
```
odanrc, line 95: Looks like another prefetcher with incorrect use of the secure bit. I haven´t taken a look at the fix, but if it is as trivial as the others could you fix this one too?

src/mem/cache/tags/SConscript
```diff
@@ -29,7 +29,7 @@
 Import('*')
 
 SimObject('Tags.py', sim_objects=[
-    'BaseTags', 'BaseSetAssoc', 'SectorTags', 'CompressedTags', 'FALRU'])
+    'BaseTags', 'BaseSetAssoc', 'SectorTags', 'CompressedTags', 'FALRU', 'TaggedIndexingPolicy', 'TaggedSetAssociative'])
```
odanrc: nit: break the line

src/mem/cache/tags/tagged_entry.hh
```diff
@@ -31,44 +43,113 @@
...
 #include "base/types.hh"
 #include "mem/cache/replacement_policies/replaceable_entry.hh"
+#include "mem/cache/tags/indexing_policies/base.hh"
+#include "params/TaggedIndexingPolicy.hh"
+#include "params/TaggedSetAssociative.hh"
 
 namespace gem5
 {
 
+struct TaggedTypes
```
odanrc: Since these are new classes, I suggest replacing Tagged by SecureAddr in TaggedTypes, TaggedIndexingPolicy and TaggedSetAssociative. Tagged doesn´t really convey its meaning.

I don´t think we should rename TaggedEntry for now, to avoid adding more lines to an already extensive PR

src/mem/cache/tags/tagged_entry.hh
```diff
@@ -31,44 +43,113 @@
...
+ * It extracts the address but discards the secure bit which
+ * is used for tagging only
+ */
+class TaggedSetAssociative : public TaggedIndexingPolicy
+{
+  protected:
+    virtual uint32_t
+    extractSet(const KeyType &key) const
+    {
+        return (key.address >> setShift) & setMask;
```
odanrc, line 82: For the future: Since this is an specialization to use SecureAddr, it should take the secure bit into account. Then regenerateAddr should also return KeyType instead of Addr

src/mem/cache/tags/sector_tags.cc
```diff
@@ -360,7 +358,7 @@ SectorTags::regenerateBlkAddr(const CacheBlk* blk) const
     const SectorSubBlk* blk_cast = static_cast<const SectorSubBlk*>(blk);
     const SectorBlk* sec_blk = blk_cast->getSectorBlock();
     const Addr sec_addr =
-        indexingPolicy->regenerateAddr(blk->getTag(), sec_blk);
+        indexingPolicy->regenerateAddr({blk->getTag(), false}, sec_blk);
```
odanrc: This also looks like a bug. Why is it not using blk->isSecure()?

odanrc (commented):
Thanks for taking the courage to tackle this monster, Giacomo. I see this PR as a first step towards making the addressing system more versatile and less error prone (I have fixed the lack of is-secure many times before). Do you plan to continue the work, or do you plan to stop after this PR?

I envision the KeyType being used in many more places (close to all) where Addr is currently being used, for the reason mentioned in a few comments: the secure bit is part of the address.

giactra:
First of all thanks for your review, I have already spotted some valuable comments. Wrt to your question, I have spent lots of time reworking this PR until I found an interface that I considered more or less acceptable (the initial version didn't use templates and was relaying on std::any instead)

> Thanks for taking the courage to tackle this monster, Giacomo. I see this PR as a first step towards making the addressing system more versatile and less error prone (I have fixed the lack of is-secure many times before). Do you plan to continue the work, or do you plan to stop after this PR?
> 

 I don't plan to spend more time on this in the near future; the only changes I already have internally and I will push at a certain point (before v24.1 I hope) is the use of the AssociativeCache with the Arm TlbEntry. I 'd be happy if someone decided to take over and cleaned up existing code even further.

> I envision the KeyType being used in many more places (close to all) where Addr is currently being used, for the reason mentioned in a few comments: the secure bit is part of the address.

Makes sense, and I see you left some comments in the prefetcher code where the security of the access is not considered when searching/inserting an entry:

https://github.com/gem5/gem5/pull/1446#discussion_r1712665316

So refactoring the code exposed this issue (which was otherwise sometimes hidden). The reason why I didn't address this and fixed existing code is because I intend this PR to be a refactoring only PR.
I don't want these changes to affect the timing in any way; I believe aforementioned problems (setting is_secure=false) should be addressed by a separate PR.

src/mem/cache/tags/tagged_entry.hh
giactra: I totally agree with you, but on the other hand SecureIndexingPolicy/SecureSetAssociative might sounds like something specific to the Secure address space (whereas it should handle Non-Secure address space as well)

Maybe more than Secure we should find a word which describes the presence of multiple address spaces?

Or maybe we could say that multiple address spaces are the norm, and we could rename



TaggedIndexingPolicy -> IndexingPolicy
TaggedSetAssociative -> SetAssociative

**Option1**:

BaseIndexingPolicy -> same
SetAssociative -> BaseSetAssociave (to match BaseIndexingPolicy)

**Option2**:

BaseIndexingPolicy -> GlobalIndexingPolicy
SetAssociative -> GlobalSetAssociative

Where the Global kw refers to the fact that it does not make distinction across address spaces...

What do you think?

src/mem/cache/tags/indexing_policies/base.hh
giactra: So you mean keeping the method implementation in the impl file instead of having them inlined in the header?

src/mem/cache/tags/indexing_policies/set_associative.cc
giactra: While I get your point, I believe the bottom line is that at the moment the BaseIndexingPolicy is behaving like a global address space policy (check https://github.com/gem5/gem5/pull/1446#discussion_r1713370364). If the object resource is to be partitioned by address spaces, they just have the indexed cache to stop using it.
Of course this means we would have to take into consideration the security bit in the TaggedIndexingPolicy as well; I believe this change should be part of a different PR

powerjg:
@giactra thanks for all of your hard work here!

> So refactoring the code exposed this issue (which was otherwise sometimes hidden). The reason why I didn't address this and fixed existing code is because I intend this PR to be a refactoring only PR.
> I don't want these changes to affect the timing in any way; I believe aforementioned problems (setting is_secure=false) should be addressed by a separate PR.

I totally agree that this should be in a separate PR, and there's no expectation that you should take this on. However, I would greatly appreciate it if you could create a new issue while it's fresh in your mind so we don't forget to work on this at some point.

src/mem/cache/cache_blk.hh
giactra: We do store the secure bit.. If you see the tag type is KeyType which carries the secure bit. CacheBlk::insert ultimately calls the TaggedEntry::insert which stores the secure bit in it.

src/mem/cache/prefetch/indirect_memory.cc
giactra: If you see https://github.com/gem5/gem5/blob/v23.1.0.0/src/mem/cache/prefetch/indirect_memory.cc#L173C35-L173C80 it has always been like this.

As I mentioned in my earlier comment, this is supposed to be a refactoring PR and fixing this behaviour it is beyond its scope. We should certainly address this with a later PR

src/mem/cache/prefetch/pif.cc
giactra: Ditto: https://github.com/gem5/gem5/blob/v23.1.0.0/src/mem/cache/prefetch/pif.cc#L180

src/mem/cache/prefetch/signature_path.cc
giactra: Ditto

src/mem/cache/prefetch/spatio_temporal_memory_streaming.cc
giactra: Ditto

src/mem/cache/tags/sector_tags.cc
giactra: Ack. At the moment no regenerateAddr is making use of the secure bit, so passing the blk security or false does not make a difference. But I agree we should nevertheless pass it

src/mem/cache/tags/tagged_entry.hh
giactra: Makes sense

src/mem/cache/prefetch/stride.hh
giactra: Yes it makes sense... Alternatively we could make them return a reference to the table... It seems to me the best way to express ownership. Let me know what you think of returning a reference or if you still prefer weak_ptr. I will update the PR accordingly

src/mem/cache/tags/sector_tags.cc
giactra: Done

src/base/cache/cache_entry.hh
giactra: Yeah, this change does not convince me either.

I can implement your suggestion, but in general I feel part of the problem comes from the fact that tagging information (we need the IP to extract the Tag) shouldn't come from the IndexingPolicy anyway (though I understand why we are currently doing it).

I moved the IP to the cache entry because in the previous iteration the IP was not templatized. Now that we can extract multiple data types from the Types tparam, there is also the opportunity to keep the extractTag method within the associative cache if we wanted to...

odanrc: How do you think we should proceed here? I am ok with leaving as is and fixing in the future, but it definitely yells ´fix me!' :)

src/mem/cache/cache_blk.hh
odanrc: Indeed. Resolved.

side note: I have no idea how to mark things as resolved in GitHub - can one only do that as a maintainer?

src/mem/cache/prefetch/indirect_memory.cc
odanrc: Resolved (created jira task)

src/mem/cache/tags/indexing_policies/base.hh
odanrc: Yes. It makes it easier to go through documentation, as well as keeps things standardized to facilitate always opening the desired file.

src/mem/cache/tags/tagged_entry.hh
odanrc: Out of the options you provided, I am leaning towards the Global option, but just weakly - it may work now, but if we end up having someone experimenting with more address space possibilities, this may introduce confusion.

I still think that although I recognize that the SecureAddr prefix is ambiguous, it should be ok to add it as a part of gem5´s learning curve, as long as we document it well enough to make it explicit that it can also contain addresses outside the secure space. I guess SecurableAddr resolves that ambiguity, but it may be a daunting name - nonetheless, with enough documentation it should be ok. Thoughts on SecurableAddr?

src/mem/cache/tags/tagged_entry.hh
odanrc: Could you also add a Jira task for this, just so that it does not get lost?

src/mem/cache/prefetch/spatio_temporal_memory_streaming.cc
odanrc: Resolved (created jira task)

src/mem/cache/prefetch/pif.cc
odanrc: Resolved (created jira task)

src/mem/cache/prefetch/signature_path.cc
odanrc: Resolved (created jira task)

src/mem/cache/tags/indexing_policies/set_associative.cc
odanrc: Could you also add a Jira task for this?

odanrc (commented):
Sorry for the delay continuing the review here. I am ok with most of this - I just think we should have jira tasks for the bugs we found here.

src/base/cache/cache_entry.hh
```diff
@@ -82,10 +101,10 @@ class CacheEntry : public ReplaceableEntry
      * @param tag The tag value.
```
pranith: this @param needs to be updated.

src/base/cache/cache_entry.hh
pranith: I don't quite follow why we need the indexing policy in the cache entry. Isn't it a property of the cache object? The cache object would have access to the indexing policy which should give you the tag directly instead asking the cache entry for the tag. What am I missing?

src/mem/cache/cache_blk.hh
giactra: I don't know tbf. Probably. This is something I don't like about GitHub, together with the fact that I cannot add a non-maintainer as reviewer

src/base/cache/cache_entry.hh
giactra: Ack

src/base/cache/cache_entry.hh
giactra: Hey @pranith, the justification is more or less described in one of the commits:

> Exposing the tag of a cache entry through the associative
> cache APIs makes it hard to generalize the cache for
> structured tags. Ultimately the tag should be a property
> of the cache entry and any tag extraction logic (if needed)
> should reside there. In this we can reuse the associative
> cache for different Entry params, each one bearing a different
> representation of a tag

I agree the indexing policy should be a property of the cache. The issue is at the moment the AssociativeCache is also using it to extract the Tag of an entry. Which shouldn't ideally happen as tagging and indexing are supposed to be "orthogonal" concepts. But anyway the tag is stored in the Entry so my reasoning was that considering the current design the Entry should be responsible for extracting the Tag.

To address more directly your question, if we keep the getTag method within the cache, then we need to sort out an extractTag argument and return types

```
virtual Addr
getTag(const Addr addr) const
{
    return indexingPolicy->extractTag(addr);
}
```

This change came from a previous iteration of the PR, which was not using templates. And templatizing a SimObject (IndexingPolicy) was the real pain I wanted to avoid. Moving the getTag within the Entry solved the problem and afterall it made sense as we store the tag in the entry as I said. Of course the price to pay was moving the IndexingPolicy within the entry which is far from ideal. 

I don't know which is the best solution.. We could for example store a pointer to the AssociativeCache within the Entry... Alternatively an extractTag functor. I am open to suggestions so let me know what you think and if you have abetter solution in mind

src/mem/cache/tags/tagged_entry.hh
giactra: > Out of the options you provided, I am leaning towards the Global option, but just weakly - it may work now, but if we end up having someone experimenting with more address space possibilities, this may introduce confusion.
> 

You think so? I actually believe the opposite. Let me explain:
IMHO we shouldn't instantiate different templates per different address spaces. The current Tagged classes should be used for "entries tagged according to the address space they belong to".
So whenever we introduce new address spaces we will have to transform the secure boolean into a PASpace
(physical address space) identifier. But we would still be using the same "Tagged" classes.

This is something that will definitely happen at a certain point. In Arm since FEAT_RME new PASpaces have been added, so on top of Secure/Non-Secure (from TrustZone) we now have Real and Root address spaces. We don't implement FEAT_RME for now in gem5, but I have already started replacing the secure/non-secure boolean with an enum in some parts of gem5.

That is why I am pushing back against SecurableAddr; it works for now but we will likely change it in the future. I really believe the underlying principle behind the TaggedEntry is: "tag which contains the PASpace within it", so if we don't want to simply call those classes IndexingPolicy and SetAssociative, IMHO we should name them after the PASpace tagging.

src/base/cache/cache_entry.hh
giactra: Note: I posted an extractTag implementation, feel free to review it @pranith @odanrc

src/mem/cache/tags/indexing_policies/base.hh
giactra: @odanrc I tried it and it doesn't work particularly well.
There is a conflict with the explicit instantiation above if we try to add the include at the bottom. I feel adding a impl file might be a bit of an overkill with such a simple class like this one.

src/base/cache/cache_entry.hh
pranith: The updated functor implementation is way better, thanks for fixing it @giactra!

src/base/cache/associative_cache.hh
```diff
@@ -48,12 +60,14 @@ namespace gem5
 template <typename Entry>
 class AssociativeCache : public Named
 {
-    static_assert(std::is_base_of_v<CacheEntry, Entry>,
-                  "Entry should be derived from CacheEntry");
+  protected:
 
-    typedef replacement_policy::Base BaseReplacementPolicy;
+    static_assert(std::is_base_of_v<ReplaceableEntry, Entry>,
```
pranith, line 65: Does this work if Entry is directly derived from ReplaceableEntry instead of CacheEntry?

giactra: Yes, as long as it complies with the template interface… In fact the TaggedEntry inherits directly from ReplaceableEntry now!

giactra: In fact I believe some of the CacheEntry methods might be devirtualised now. They were probably made virtual to accomodate for the TaggedEntry

pranith (approved):

src/base/cache/associative_cache.hh
pranith: thanks, the changes LGTM.

src/base/cache/cache_entry.hh
```diff
@@ -46,9 +59,25 @@ namespace gem5
 class CacheEntry : public ReplaceableEntry
 {
   public:
+    using IndexingPolicy = BaseIndexingPolicy;
+    using KeyType = Addr;
+    using TagExtractor = std::function<Addr(Addr)>;
+
     CacheEntry() = default;
```
odanrc: Can this class even work without a tag extractor? I think we may not need the register function, and instead we should always provide the extractor upon construction. By always registering the tag extractor here we guarantee that the classes that inherit from this one do not forget to register the extractor.

src/mem/cache/tags/indexing_policies/base.hh
```diff
@@ -156,10 +199,26 @@ class BaseIndexingPolicy : public SimObject
...
+  public:
+    using KeyType = Addr;
+    using Params = BaseIndexingPolicyParams;
 };
 
+using BaseIndexingPolicy = IndexingPolicyTemplate<AddrTypes>;
+template class IndexingPolicyTemplate<AddrTypes>;
+
+static constexpr auto
+genTagExtractor(BaseIndexingPolicy *ip)
```
odanrc, line 225: I am curious to why this static function is global, and not a static function in the base indexing policy. Is it due to the increased verbosity of doing the latter approach, or is there some kind of other benefit that makes this a better coding practice?

src/mem/cache/tags/tagged_entry.hh
```diff
@@ -102,38 +191,55 @@ class TaggedEntry : public CacheEntry
...
     bool _secure;
 
-    /** Clear secure bit. Should be only used by the invalidation function. */
-    void clearSecure() { _secure = false; }
-
-    /** Do not use API without is_secure flag. */
-    using CacheEntry::matchTag;
-    using CacheEntry::insert;
+    /** The entry's tag. */
+    Addr _tag;
```
odanrc, line 234: Note for the future: we can remove all secure-related functions from this class, and store a KeyType instead. Then, if needed, the user would call the KeyType´s getter and extract the secure bit directly from the KeyType.

src/mem/cache/tags/tagged_entry.hh
odanrc: Alright, you seem to have a better vision of what changes will be needed in the future to support those, so feel free to proceed naming it how you see fit :)

src/mem/cache/cache_blk.hh
odanrc: Gonna vent a bit too ;)

Although I understand why we adopted GitHub, the move from gerrit was probably the main reason that made me decide to step out of the maintainer role. The UI here is extremely confusing to me, and I keep losing track of the requests I made. Besides, finding what changed between reviews is usually a struggle; right now, for example, if I click to ´show changes since my last review´ I get a ´We went looking everywhere, but couldn’t find those commits.´.

src/mem/cache/tags/indexing_policies/base.hh
giactra: Yeah, ideally we should place it there. the issue of keeping in the templated class is the following:

As of now, even in the TaggedEntry, the extractTag is expecting a Addr data type. So we cannot simply add the following

```
static constexpr auto                    
genTagExtractor(IndexingPolicyTemplate *ip)                                                                                                                    
{                                      
    return [ip] (const KeyType &key) { return ip->extractTag(key); };
}  
```

So I went for this quick and dirty option because placing

```
static constexpr auto                    
genTagExtractor(IndexingPolicyTemplate *ip)                                                                                                                    
{                                      
    return [ip] (Addr addr) { return ip->extractTag(addr); };
}  
```

in the template felt equally wrong (it will be defined for non-Addr tag policies as well). But anyway we already have an Addr based extractTag in the template so if you want I can go for the second option.

I believe the cleanest solution would be to add an extra level in the hierarchy and move tag extraction logic over there so that it does not reside in the template. In lots of cases you don't really need to extract the Tag and you shouldn't be doing it for structured tags

odanrc: Nah, it is fine to leave as is; I was just wondering about the thought process.

src/base/cache/cache_entry.hh
giactra: Yes makes sense and I amended the PR accordingly. As a note: while this has been done for the CacheEntry, the TaggedEntry is still using the register method. The reason being is that some CacheBlk child classes are being allocated with the default constructor (example: https://github.com/gem5/gem5/blob/stable/src/mem/cache/tags/fa_lru.cc#L81)

They are anyway using their own tag extraction logic (https://github.com/gem5/gem5/blob/stable/src/mem/cache/tags/fa_lru.hh#L244) and not only that...

odanrc: FALRU always being the naysayer...

odanrc (approved):
Thanks for the changes. Just make sure not to forget creating the jira tasks, if you haven´t done so already. LGTM!

src/mem/cache/tags/tagged_entry.hh
giactra: https://github.com/gem5/gem5/issues/1494

src/mem/cache/tags/tagged_entry.hh
giactra: https://github.com/gem5/gem5/issues/1494

giactra:
Thanks @odanrc and @pranith for your reviews, they made this PR much cleaner

src/mem/cache/tags/tagged_entry.hh
```diff
@@ -102,38 +191,55 @@ class TaggedEntry : public CacheEntry
...
 };
 
+static constexpr auto
+genTagExtractor(TaggedIndexingPolicy *ip)
```
powerjg, line 246: Can you add a couple of lines of documentation here? I think this means "generic" tag extractor, but I'm not certain.

Also, this is magic c++ syntax (at least to me). Could you add a line explaining what the syntax is doing?

powerjg (commented):
This looks good to me. I have one small comment below about adding some documentation. 

My only (minor) concern is the templating magic. I worry a little about how someone new to the code will know how to add a new indexing policy. But, I guess there's enough examples that they should be able to figure it out.

src/mem/cache/tags/tagged_entry.hh
giactra: It actually stands for "generate tag extractor" as it returns a lambda. I will document this better

powerjg (approved):

giactra:
Thanks @powerjg for your review!

PR #1449, opened 2024-08-09
mem: Stride Prefetcher Fix (#1448)

### LINKED ISSUE #1448: The implementation of the Stride prefetcher has a few bugs
**Describe the bug**
The current implementation of the Stride Prefetcher has the following three issues:
1. The stride prefetcher works at the granularity of cache lines. Therefore, for computing the stride, the prefetcher should use cache line aligned addresses, and not word aligned addresses
2. Since the prefetcher would be notified on every word access, if  a workload reads multiple bytes from a single cache line (which happens a lot), then when the prefetcher observes access to words $w_1$ and $w_2$ such that both of the words lie in the same cache line, the prefetcher would end up computing the stride to be 0. This, by itself is not an issue, but since the prefetcher updates its confidence counters based on the computed and the last observed stride, this computation would reduce the confidence value, thereby reducing the effectiveness of the prefetcher.
3. If the current version of the prefetcher finds that the new stride does not match the old stride, it still issues prefetches using the new stride even though the new stride has not crossed the confidence threshold yet. This, in turn, can result in a whole lot of useless prefetches which hurt the accuracy and put unnecessary pressure on the memory subsystem.

**Affects version**
The version of Gem5 on the develop branch

My colleague (@AbhishekUoR) and I have fixed the issue and we are opening a PR for it. This issue is just there to ensure that everyone knows about why the PR is being opened in the first place

Setu-Gupta:
This PR fixes the issues mentioned in #1448.

**Note that this contribution is the result of a joint collaboration with @AbhishekUoR**

This PR introduces the following 4 changes:
1. It changes the addresses which are used to compute the stride to cache line aligned addresses (the current version uses word aligned addresses)
2. It correctly returns if the stride does not match (as opposed to issuing prefetches using the new stride incorrectly)
3. It returns if the new stride is 0, indicating multiple reads from the same cache line.
4. It removes code which is no longer necessary after the addition of changes number 1 and 3.

Change-Id: Ic346d0e15df6d07e2b93289c8d6b89b4c2f45a34

### DIFF
--- a/src/mem/cache/prefetch/stride.cc
+++ b/src/mem/cache/prefetch/stride.cc
@@ -184,7 +184,7 @@ Stride::calculatePrefetch(const PrefetchInfo &pfi,
         // Round strides up to at least 1 cacheline
         int prefetch_stride = entry->stride;
         if (abs(prefetch_stride) < blkSize) {
-            prefetch_stride = (new_stride < 0) ? -blkSize : blkSize;
+            prefetch_stride = (prefetch_stride < 0) ? -blkSize : blkSize;
         }
 
         Addr new_addr = pf_addr + distance * prefetch_stride;
--- a/src/mem/cache/prefetch/stride.cc
+++ b/src/mem/cache/prefetch/stride.cc
@@ -86,7 +86,8 @@ Stride::Stride(const StridePrefetcherParams &p)
     degree(p.degree),
     distance(p.distance),
     pcTableInfo(p.table_assoc, p.table_entries, p.table_indexing_policy,
-                p.table_replacement_policy)
+                p.table_replacement_policy),
+    useCachelineAddr(p.use_cache_line_address)
 {
 }
 
@@ -133,7 +134,8 @@ Stride::calculatePrefetch(const PrefetchInfo &pfi,
     }
 
     // Get required packet info
-    Addr pf_addr = blockAddress(pfi.getAddr());
+    Addr pf_addr = useCachelineAddr ? blockAddress(pfi.getAddr())
+                                    : pfi.getAddr();
     Addr pc = pfi.getPC();
     bool is_secure = pfi.isSecure();
     RequestorID requestor_id = useRequestorId ? pfi.getRequestorId() : 0;
@@ -151,13 +153,13 @@ Stride::calculatePrefetch(const PrefetchInfo &pfi,
         int new_stride = pf_addr - entry->lastAddr;
 
         // Do nothing on repeated memory access
-        if (new_stride == 0)
+        if (useCachelineAddr && new_stride == 0)
             return;
 
         bool stride_match = (new_stride == entry->stride);
 
         // Adjust confidence for stride entry
-        if (stride_match && new_stride != 0) {
+        if (stride_match) {
             entry->confidence++;
         } else {
             entry->confidence--;
@@ -174,13 +176,15 @@ Stride::calculatePrefetch(const PrefetchInfo &pfi,
 
         entry->lastAddr = pf_addr;
 
-        // Return if we don't see a stride match
-        if (!stride_match)
-            return;
-
         // Abort prefetch generation if below confidence threshold
         if (entry->confidence.calcSaturation() < threshConf) {
             return;
+        }
+
+        // Round strides up to at least 1 cacheline
+        int prefetch_stride = entry->stride;
+        if (abs(prefetch_stride) < blkSize) {
+            prefetch_stride = (new_stride < 0) ? -blkSize : blkSize;
         }
 
         Addr new_addr = pf_addr + distance * prefetch_stride;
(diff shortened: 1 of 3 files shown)


src/mem/cache/prefetch/stride.cc
```diff
@@ -149,6 +149,11 @@ Stride::calculatePrefetch(const PrefetchInfo &pfi,
 
         // Hit in table
         int new_stride = pf_addr - entry->lastAddr;
+
+        // Do nothing on repeated memory access
```
odanrc, line 155: Why is this the correct approach? My initial thought is that deciding if 0-strides should not modify the confidence should at the very least be a configuration parameter.

If we conclude that 0 strides should not update the confidence the code in line 157 needs to be updated.

src/mem/cache/prefetch/stride.cc
```diff
@@ -169,17 +174,15 @@ Stride::calculatePrefetch(const PrefetchInfo &pfi,
 
         entry->lastAddr = pf_addr;
 
+        // Return if we don't see a stride match
```
odanrc: I think this would be a bug. The point of having the confidence threshold is exactly to decide when the predicted stride is relevant or not. If we have 99% confidence that an access to X will have a stride S, but this time we got a stride of S2, then the confidence indicates that S2 was just an outlier, and that we will probably see S again in the next access.

odanrc (changes requested):

giactra:
@rjc-arch

src/mem/cache/prefetch/stride.cc
Setu-Gupta: I agree. I've modified the code and now the users can choose between the granularity of cache line or word address. Only if the granularity is set to cache lines, we will ignore zero strides

src/mem/cache/prefetch/stride.cc
Setu-Gupta: I agree. I've changed the code to reflect this

src/mem/cache/prefetch/stride.cc
odanrc: Alright, I agree with/understand the change to line 162; however, I don´t understand why the zero-stride check only applies to the cache-line scenario. Could you clarify?

src/mem/cache/prefetch/stride.cc
```diff
@@ -174,9 +181,9 @@ Stride::calculatePrefetch(const PrefetchInfo &pfi,
             return;
         }
 
-        // Round strides up to atleast 1 cacheline
-        int prefetch_stride = new_stride;
-        if (abs(new_stride) < blkSize) {
+        // Round strides up to at least 1 cacheline
+        int prefetch_stride = entry->stride;
+        if (abs(prefetch_stride) < blkSize) {
             prefetch_stride = (new_stride < 0) ? -blkSize : blkSize;
```
odanrc: I think new_stride should have been replaced by prefetch_stride here too

src/mem/cache/prefetch/stride.cc
```diff
@@ -133,7 +134,8 @@ Stride::calculatePrefetch(const PrefetchInfo &pfi,
     }
 
     // Get required packet info
-    Addr pf_addr = pfi.getAddr();
+    Addr pf_addr = useCachelineAddr ? blockAddress(pfi.getAddr())
```
odanrc, line 137: Is there a performance benefit on not using block addresses given that we end up rounding the prefetch strides to block addresses anyway?

src/mem/cache/prefetch/stride.cc
Setu-Gupta: The reason why I was treating the the zero stride case separately was because I was trying to guard against multiple accesses to the same cache line. This can only happen if the prefetcher is operating at the granularity of the cache lines. If it operating at the granularity of word addresses, then we don't care even if we observe zero strides

src/mem/cache/prefetch/stride.cc
Setu-Gupta: In a micro-benchmark that I wrote I saw a very small improvement in IPC when I was operating at the cache line granularity. This mainly happens because when we operate on word granularity, we may issue multiple prefetches for the same cache line for different words. Although these prefetches can be coalesced into one at the prefetch queue, we would still be exhausting the bandwidth of the prefetcher if we issue a lot of redundant prefetches. That's the reason why I've set the default parameters of the prefetcher such that it operates on cache line granularity

src/mem/cache/prefetch/stride.cc
Setu-Gupta: Yes, you are correct. Thanks for catching this error. I've fixed this

odanrc (approved):
The changes seem reasonable, but since I imagine the stride prefetcher is one of the most used prefetchers by gem5 users, and these changes effectively change its behavior, we will probably need to add an entry to the release notes @BobbyRBruce .

BobbyRBruce:
> The changes seem reasonable, but since I imagine the stride prefetcher is one of the most used prefetchers by gem5 users, and these changes effectively change its behavior, we will probably need to add an entry to the release notes @BobbyRBruce .

Added here: #1460

BobbyRBruce (approved):

Issue #1349, opened 2024-07-13
Multiple RubySystem objects


abmerop:
Allow for multiple RubySystem objects. Not to be confused with multiple Ruby **protocols**.  This has been a topic in the past for having one System object with two nodes communicating, but has not made it into gem5. Now, we have a PCI device with a unique Ruby protocol (amdgpu). This would allow for more modular GPUs in gem5 and much easier integration with stdlib. For example, there could be an X86PciBoard (~name) and you could "plug in" any PCI device, e.g., an AMD gpu. 

## Challenges

RubySystem has multiple static members - i.e., is a singleton.  These would have to be removed. The biggest challenge would be the makeLineAddress function. This requires either passing a pointer to RubySystem or defining it elsewhere. The last developer meeting proposes the latter.

## Opportunity

The AMD GPU object right now uses a "disjoint" Ruby network which works for dGPU simulation.  There is a design issue with Ruby's network that causes problems with GPUFS because both share the same directory SLICC machine.  This causes bad NoC routes and immediately deadlocks simulation if a timing CPU is used (O3, Minor, Simple).  Mutiple RubySystems would remove this problem and would allow for **APU** simulations in GPUFS.

abmerop:
Had some time to look into this today. Here is what I am suggesting:

- AbstractController gets a RubySystem pointer. This will require changing some configs to make everything inheriting AbstractController set a RubySystem pointer.
   - Alternate: The sequencer already has a RubySystem poiner and most SLICC state machines (files ending with .sm) have a sequencer. Not all do which will require changing those SMs, so this is less ideal. However, all config files except stdlib set the ruby_system pointer when creating a RubySequencer SimObject

- Modify SLICC parser to set sizes for DataBlock and WriteMask. These use RubySystem static members to alloc memory. This is going to be very difficult to modify in a non-intrusive way otherwise (i.e., would probably requiring changing every single protocol to allocate data blocks manually).
   - Currently I have modified the SLICC parser to find any SLICC object defined using `structure(...)` which contains a `DataBlock` member. I removed the default constructor from DataBlock and forced a constructor containing the block size to be used.
[...]

PR #1453, opened 2024-08-12
mem-ruby: Remove static methods from RubySystem

abmerop:
There are several parts to this PR to work towards #1349 .

(1) Make RubySystem::getBlockSizeBytes non-static by providing ways to
access the block size or passing the block size explicitly to classes.

The main changes are:
 - DataBlocks must be explicitly allocated. A default ctor still exists
   to avoid needing to heavily modify SLICC. The size can be set using a
   realloc function, operator=, or copy ctor. This is handled completely
   transparently meaning no protocol or config changes are required.
 - WriteMask now requires block size to be set. This is also handled
   transparently by modifying the SLICC parser to identify WriteMask
   types and call setBlockSize().
 - AbstractCacheEntry and TBE classes now require block size to be set.
   This is handled transparently by modifying the SLICC parser to
   identify these classes and call initBlockSize() which calls
   setBlockSize() for any DataBlock or WriteMask.
 - All AbstractControllers now have a pointer to RubySystem. This is
   assigned in SLICC generated code and requires no changes to protocol
   or configs.
 - The Ruby Message class now requires block size in all constructors.
   This is added to the argument list automatically by the SLICC parser.
   
(2) Relax dependence on common functions in src/mem/ruby/common/Address.hh
so that RubySystem::getBlockSizeBits is no longer static. Many classes
already have a way to get block size from the previous commit, so they
simply multiple by 8 to get the number of bits. For handling SLICC and
reducing the number of changes, define makeCacheLine, getOffset, etc. in
RubyPort and AbstractController. The only protocol changes required are
to change any "RubySystem::foo()" calls with "m_ruby_system->foo()".

For classes which do not have a way to get access to block size but
still used makeLineAddress, getOffset, etc., the block size must be
passed to that class. This requires some changes to the SimObject
interface for two commonly used classes: DirectoryMemory and
RubyPrefecther, resulting in user-facing API changes

User-facing API changes:
 - DirectoryMemory and RubyPrefetcher now require the cache line size as
   a non-optional argument.
 - RubySequencer SimObjects now require RubySystem as a non-optional
   argument.
 - TesterThread in the GPU ruby tester now requires the cache line size
   as a non-optional argument.

(3) Removes static member variables in RubySystem which control
randomization, cooldown, and warmup. These are mostly used by the Ruby
Network. The network classes are modified to take these former static
variables as parameters which are passed to the corresponding method
(e.g., enqueue, delayHead, etc.) rather than needing a RubySystem object
at all.

Change-Id: Ia63c2ad5cf0bf9d1cbdffba5d3a679bb4d3b1220

(4) There are two major SLICC generated static methods: getNumControllers()
on each cache controller which returns the number of controllers created
by the configs at run time and the functions which access this method,
which are MachineType_base_count and MachineType_base_number. These need
to be removed to create multiple RubySystem objects otherwise NetDest,
version value, and other objects are incorrect.

To remove the static requirement, MachineType_base_count and
MachineType_base_number are moved to RubySystem. Any class which needs
to call these methods must now have a pointer to a RubySystem. To enable
that, several changes are made:
 - RubyRequest and Message now require a RubySystem pointer in the
   constructor. The pointer is passed to fields in the Message class
   which require a RubySystem pointer (e.g., NetDest). SLICC is modified
   to do this automatically.
 - SLICC structures may now optionally take an "implicit constructor"
   which can be used to call a non-default constructor for locally
   defined variables (e.g., temporary variables within SLICC actions). A
   statement such as "NetDest bcast_dest;" in SLICC will implicitly
   append a call to the NetDest constructor taking RubySystem, for
   example.
 - RubySystem gets passed to Ruby network objects (Network, Topology).


abmerop:
@powerjg I was just talking to Brandon, who had implemented this originally about 7 years ago or more, and we came to the conclusion that we might have to modify *every* protocol to have DataBlock (and WriteMask) ctors that take the size as an argument. I wasn't able to make the last dev meeting but I think this would be a good agenda item for the next one.

abmerop:
> @powerjg I was just talking to Brandon, who had implemented this originally about 7 years ago or more, and we came to the conclusion that we might have to modify _every_ protocol to have DataBlock (and WriteMask) ctors that take the size as an argument. I wasn't able to make the last dev meeting but I think this would be a good agenda item for the next one.

Current commits require no changes to protocol files. Minor changes to configs.

src/mem/ruby/structures/CacheMemory.cc
```diff
@@ -89,11 +86,21 @@ CacheMemory::CacheMemory(const Params &p)
 }
 
 void
-CacheMemory::init()
+CacheMemory::setRubySystem(RubySystem* rs)
 {
+    dataArray.setClockPeriod(rs->clockPeriod());
+    tagArray.setClockPeriod(rs->clockPeriod());
+    atomicALUArray.setClockPeriod(rs->clockPeriod());
```
powerjg, line 93: Just a note: we could take this change and extend it to allow different caches to have different clocks.

This is not something we should do in this PR, but it is something we can now do more easily!

src/mem/ruby/structures/DirectoryMemory.cc
```diff
@@ -69,7 +69,7 @@ DirectoryMemory::DirectoryMemory(const Params &p)
 void
 DirectoryMemory::init()
 {
-    m_num_entries = m_size_bytes / RubySystem::getBlockSizeBytes();
+    m_num_entries = m_size_bytes / 64 /* TODO */;
```
powerjg: Still a "todo" here

configs/ruby/GPU_VIPER.py
```diff
@@ -688,7 +688,7 @@ def construct_gpudirs(options, system, ruby_system, network):
         dir_cntrl.addr_ranges = dram_intf.range
 
         # Append
-        exec("system.ruby.gpu_dir_cntrl%d = dir_cntrl" % i)
+        exec("ruby_system.gpu_dir_cntrl%d = dir_cntrl" % i)
```
powerjg, line 691: Really, we should try to remove all `exec` calls, but this is OK for now.

src/mem/ruby/common/NetDest.cc
```diff
@@ -38,19 +40,26 @@ namespace ruby
 
 NetDest::NetDest()
 {
-  resize();
+}
+
+NetDest::NetDest(RubySystem *ruby_system)
+{
+    m_ruby_system = ruby_system;
```
powerjg: Should this be like below? Maybe I'm missing something.

```suggestion
NetDest::NetDest(RubySystem *ruby_system) : m_ruby_system(ruby_system)
{
```

powerjg (changes requested):
This looks fantastic. I only have two actionable comments below. Otherwise, this is ready from my point of view.

I'll note here that there is significant overlap with #117. Rebasing that is going to be a PITA, but I'll take that on. I think using this PR's implementation is a better way forward than some of the choices I made in #117.

Another note (no action needed). I think we could use `Parent.any` or some other proxy in many cases (as long as we made it a rule that all Ruby objects are children of their RubySystem). However, my philosophy is that explicit is better than implicit, so I'm very happy with the approach of this PR.

Thanks for all of the hard work!

configs/ruby/GPU_VIPER.py
abmerop: Agreed

src/mem/ruby/common/NetDest.cc
abmerop: Made this change

src/mem/ruby/structures/DirectoryMemory.cc
abmerop: Whoops, fixed now

powerjg (dismissed):

src/cpu/testers/gpu_ruby_test/address_manager.cc
```diff
@@ -64,7 +64,9 @@ AddressManager::AddressManager(int n_atomic_locs, int n_normal_locs_per_atomic)
     std::shuffle(
         randAddressMap.begin(),
         randAddressMap.end(),
-        std::default_random_engine(random_mt.random<unsigned>(0,UINT_MAX))
+        // TODO: This is a bug unrelated to this draft PR but the GPU tester is
+        // useful for testing this PR.
+        std::default_random_engine(random_mt.random<unsigned>(0,UINT_MAX-1))
```
mattsinc, line 69: maybe spacing off on this line?

mattsinc (approved):
One minor thing -- thanks for all your hard work on this!

src/cpu/testers/gpu_ruby_test/address_manager.cc
abmerop: Do you want this removed?

mattsinc: no ... it looks fine in this format but in the review format it looks like line 69 had incorrect spacing

powerjg:
@abmerop should we do a merge commit or a squash? I.e., do these commits make sense on their own?

abmerop:
I guess squash.  The commits are mostly self contained changes, but there might be one or two fixes of earlier patches later in the series.

mattsinc:
Sorry I missed the end of this, but seems just a squash/merge since I last reviewed.

src/mem/ruby/network/MessageBuffer.hh
```diff
@@ -124,6 +125,7 @@ class MessageBuffer : public SimObject
     const MsgPtr &peekMsgPtr() const { return m_prio_heap.front(); }
 
     void enqueue(MsgPtr message, Tick curTime, Tick delta,
+                bool ruby_is_random, bool ruby_warmup,
```
giactra, line 128: Can we have some documentation on what there two new parameters mean?

abmerop: Does #1735 suffice?  These are basically the former `RubySystem::getRandomization()` and `RubySystem::getWarmupEnabled()`

PR #1454, opened 2024-08-12
mem-cache: Implementation of SMS prefetcher

AbhishekUoR:
This PR adds the SMS prefetcher described in [this ](https://web.eecs.umich.edu/~twenisch/papers/isca06.pdf) paper.
This work was done in collaboration with @Setu-Gupta, and @xmlizhao

On branch sms
Changes to be committed:
modified: src/mem/cache/prefetch/Prefetcher.py
modified: src/mem/cache/prefetch/SConscript
new file: src/mem/cache/prefetch/sms.cc
new file: src/mem/cache/prefetch/sms.hh

Change-Id: I68d3bb6cf07385177d0f776fb958f652cfc41489

### DIFF
--- a/src/mem/cache/prefetch/Prefetcher.py
+++ b/src/mem/cache/prefetch/Prefetcher.py
@@ -599,6 +599,22 @@ class BOPPrefetcher(QueuedPrefetcher):
     on_inst = False
 
 
+class SmsPrefetcher(QueuedPrefetcher):
+    # Paper: https://web.eecs.umich.edu/~twenisch/papers/isca06.pdf
+    type = "SmsPrefetcher"
+    cxx_class = "gem5::prefetch::Sms"
+    cxx_header = "mem/cache/prefetch/sms.hh"
+    ft_size = Param.Unsigned(64, "Size of Filter and Active generation table")
+    pht_size = Param.Unsigned(16384, "Size of pattern history table")
+    region_size = Param.Unsigned(4096, "Spatial region size")
+
+    queue_squash = True
+    queue_filter = True
+    cache_snoop = True
+    prefetch_on_access = True
+    on_inst = False
+
+
 class SBOOEPrefetcher(QueuedPrefetcher):
     type = "SBOOEPrefetcher"
     cxx_class = "gem5::prefetch::SBOOE"
--- a/src/mem/cache/prefetch/sms.cc
+++ b/src/mem/cache/prefetch/sms.cc
@@ -1,0 +1,165 @@
+/*
+ * Copyright (c) 2024 Samsung Electronics
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ * Describes a SMS prefetcher based on template policies.
+ */
+
+#include "mem/cache/prefetch/sms.hh"
+
+#include "debug/HWPrefetch.hh"
+#include "params/SmsPrefetcher.hh"
+
+namespace gem5
+{
+
+namespace prefetch
+{
+
+Sms::Sms(const SmsPrefetcherParams &p)
+    : Queued(p), Max_Contexts(p.ft_size), MAX_PHTSize(p.pht_size),
+      Region_Size(p.region_size)
+{
+        AGT.clear();
+        AGTPC.clear();
+        FT.clear();
+        PHT.clear();
+        fifoFT.clear();
+        lruAGT.clear();
+        lruPHT.clear();
+
+}
+void
+Sms::notifyEvict(const EvictionInfo &info)
+{
+    //Check if any active generation has ended
+    Addr regionBase = roundDown(info.addr, Region_Size);
+    std::pair <Addr,Addr> pc_offset = AGTPC[regionBase];
+    if (AGT.find(regionBase) != AGT.end()) {
+        //remove old recording
+        if (PHT.find(pc_offset) != PHT.end()) {
+            PHT[pc_offset].clear();
+        }
+        //Move from AGT to PHT
+        for (std::set<Addr>::iterator it = AGT[regionBase].begin();
+         it != AGT[regionBase].end(); it ++) {
+            PHT[pc_offset].insert(*it);
+        }
+        lruPHT.push_front(pc_offset);
+    }
+
+    while (PHT.size() > MAX_PHTSize) {
+        PHT.erase(lruPHT.back());
+        lruPHT.pop_back();
+    }
+
+    AGTPC.erase(regionBase);
+    AGT.erase(regionBase);
+
+
+}
+void
+Sms::calculatePrefetch(const PrefetchInfo &pfi,
+    std::vector<AddrPriority> &addresses,
+    const CacheAccessor &cache)
+{
+
+    if (!pfi.hasPC()) {
+        DPRINTF(HWPrefetch, "Ignoring request with no PC.\n");
+        return;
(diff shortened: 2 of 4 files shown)


ivanaamit:
@AbhishekUoR, the tests aren't running for this PR. We believe there's a bug with GitHub that occurs when a PR is initially created on the stable branch and then later changed to develop—GitHub doesn't seem to handle that well. I'll try closing and reopening the PR, but if that doesn't trigger the tests, I suggest you close this PR and open a new one with the same changes, targeting the develop branch instead of stable. Thank you for understanding.

edit: The tests are running now, which is great. To avoid this issue in the future, please try to target the develop branch from the start.

src/mem/cache/prefetch/sms.cc
```diff
@@ -0,0 +1,165 @@
...
+{
+
+namespace prefetch
+{
+
+Sms::Sms(const SmsPrefetcherParams &p)
+    : Queued(p), Max_Contexts(p.ft_size), MAX_PHTSize(p.pht_size),
+      Region_Size(p.region_size)
+{
+        AGT.clear();
```
giactra: Nit: 4 whitespaces instead of 8

src/mem/cache/prefetch/sms.cc
```diff
@@ -0,0 +1,165 @@
...
+        fifoFT.clear();
+        lruAGT.clear();
+        lruPHT.clear();
+
+}
+void
+Sms::notifyEvict(const EvictionInfo &info)
+{
+    //Check if any active generation has ended
+    Addr regionBase = roundDown(info.addr, Region_Size);
```
giactra: Local variables use snake case (region_base)

src/mem/cache/prefetch/sms.cc
```diff
@@ -0,0 +1,165 @@
...
+    while (PHT.size() > MAX_PHTSize) {
+        PHT.erase(lruPHT.back());
+        lruPHT.pop_back();
+    }
+
+    AGTPC.erase(regionBase);
+    AGT.erase(regionBase);
+
+
+}
```
giactra, line 84: Nit, newline between two methods

src/mem/cache/prefetch/sms.cc
```diff
@@ -0,0 +1,165 @@
...
+{
+
+    if (!pfi.hasPC()) {
+        DPRINTF(HWPrefetch, "Ignoring request with no PC.\n");
+        return;
+    }
+
+    Addr blk_addr = blockAddress(pfi.getAddr());
+    Addr pc = pfi.getPC();
+    Addr regionBase = roundDown(blk_addr, Region_Size);
```
giactra: Ditto

src/mem/cache/prefetch/sms.cc
```diff
@@ -0,0 +1,165 @@
...
+            fifoFT.pop_back();
+        }
+    }
+
+    //Prediction
+    std::pair <Addr, Addr> pc_offset = std::make_pair(pc,offset);
+    if (PHT.find(pc_offset) != PHT.end()) {
+        for (std::set<Addr>::iterator it = PHT[pc_offset].begin();
+         it != PHT[pc_offset].end(); it ++) {
+            Addr prefAddr = blockAddress(regionBase + (*it));
```
giactra: Ditto

src/mem/cache/prefetch/sms.cc
```diff
@@ -0,0 +1,165 @@
...
+    //Check if any active generation has ended
+    Addr regionBase = roundDown(info.addr, Region_Size);
+    std::pair <Addr,Addr> pc_offset = AGTPC[regionBase];
+    if (AGT.find(regionBase) != AGT.end()) {
+        //remove old recording
+        if (PHT.find(pc_offset) != PHT.end()) {
+            PHT[pc_offset].clear();
+        }
+        //Move from AGT to PHT
+        for (std::set<Addr>::iterator it = AGT[regionBase].begin();
```
giactra: Nit, consider using auto

src/mem/cache/prefetch/sms.cc
```diff
@@ -0,0 +1,165 @@
...
+        for (std::deque <Addr>::iterator lit = lruAGT.begin();
+         lit != lruAGT.end(); lit ++) {
+            if ((*lit) == regionBase) {
+                lruAGT.erase(lit);
+                lruAGT.push_front(regionBase);
+                break;
+            }
+        }
+    }
+    else if (FT.find(regionBase) != FT.end()) {
```
giactra: This should be

    } else if () {

src/mem/cache/prefetch/sms.cc
```diff
@@ -0,0 +1,165 @@
...
+        //Recycle FT entry
+        FT.erase(regionBase);
+        //Make space for next entry
+        while (AGT.size() > Max_Contexts) {
+            AGT.erase(lruAGT.back());
+            AGTPC.erase(lruAGT.back());
+            lruAGT.pop_back();
+        }
+    }
+    else {
```
giactra: Ditto

giactra (changes requested):
Hi @AbhishekUoR thanks for this contribution!

So far I have coding style comments you can start address if you want. I will review the functionality sometime this week

giactra:
Hi @AbhishekUoR, thanks for addressing my comments When reviewing the functionality I noticed we actually already have a spatial prefetcher implementation which is the [Spatio-Temporal Memory Streaming Prefetcher (STeMS)](https://github.com/gem5/gem5/blob/stable/src/mem/cache/prefetch/spatio_temporal_memory_streaming.hh). Which I presume is building up on SMS adding temporal streaming.
Did you notice that prefetcher implementation in gem5 when you developed SMS?

I believe the two prefetchers should try to reuse as much code as possible. For example the AGT entry definition in the STeMS.

Another comment I have is that I see you hardcoded an LRU replacement policy with a deque. It would also be nice if we could make it configurable with the ReplacementPolicy Param, and to use the AssociativeCache as well.

Please let me know your thoughts

AbhishekUoR:
Hello @giactra, I initially tried using STeMS as SMS by disabling the TEMS component, but I found that they differ in how they learn patterns. Another reason for keeping SMS as a separate prefetcher is that ARM cores specifically use SMS and GSMS (for larger pages). 

I chose not to use the STeMS infrastructure to develop independent code that I can test without worrying about issues related to associative set implementations, such as those mentioned here: https://gem5-review.googlesource.com/c/public/gem5/+/37597. 

I agree with the idea of using a replacement policy parameter, which I may consider for future changes to SMS. Currently, I am focusing on implementing other prefetcher-related changes and bug fixes in gem5.

src/mem/cache/prefetch/Prefetcher.py
```diff
@@ -599,6 +599,22 @@ class BOPPrefetcher(QueuedPrefetcher):
     on_inst = False
 
 
+class SmsPrefetcher(QueuedPrefetcher):
+    # Paper: https://web.eecs.umich.edu/~twenisch/papers/isca06.pdf
+    type = "SmsPrefetcher"
+    cxx_class = "gem5::prefetch::Sms"
+    cxx_header = "mem/cache/prefetch/sms.hh"
+    ft_size = Param.Unsigned(64, "Size of Filter and Active generation table")
```
giactra, line 607: From the paper it looks like the AGT is split into a filter table and an accumulation table.
Shall we just call it agt_size? And amen the comment accordingly:

"Size of Filter and Accumulation Table"

src/mem/cache/prefetch/sms.hh
```diff
@@ -0,0 +1,82 @@
...
+class Sms : public Queued
+{
+
+  private:
+    const int Max_Contexts; //= 64;
+    const uint64_t MAX_PHTSize; //= 512;
+    const Addr Region_Size; //= 4096;
+
+    std::map< Addr, std::set<Addr> > AGT;
+    std::map< Addr, std::pair<Addr,Addr> > AGTPC;
```
giactra, line 60: Isn't this the Accumulation Table? (Which is part of the AGT together with the FT?)
Also I would consider having a single map like

```
struct AccumulationTableEntry
{
    Addr pc;
    Addr offset;
    std::set<Addr> pattern;
};
std::map<Addr, AccumulationTableEntry> accumulationTable;
```

Instead of having two different maps with the need of storing the same key twice

src/mem/cache/prefetch/sms.hh
```diff
@@ -0,0 +1,82 @@
...
+{
+
+
+class Sms : public Queued
+{
+
+  private:
+    const int Max_Contexts; //= 64;
+    const uint64_t MAX_PHTSize; //= 512;
+    const Addr Region_Size; //= 4096;
```
giactra, line 57: I would remove the comment here which will change if someone tweaks the parameters.
Also from a coding style perspective I would consider:
```

    const int MaxContexts;
    const uint64_t MaxPHTSize;
    const Addr RegionSize;
```

src/mem/cache/prefetch/sms.cc
giactra: This still needs some resolving

giactra:
> Hello @giactra, I initially tried using STeMS as SMS by disabling the TEMS component, but I found that they differ in how they learn patterns.

I didn't look into STeMS, but I was not proposing for you to tweak it to make it work like SMS, rather to extract the common code and use it as a base for both (if it is not the algorithm I believe there *might* still be chance to reuse some data structures). Anyway it is not a big deal and the amount of lines we save is probably trivial

> 
> I chose not to use the STeMS infrastructure to develop independent code that I can test without worrying about issues related to associative set implementations, such as those mentioned here: https://gem5-review.googlesource.com/c/public/gem5/+/37597.
> 

I looked into the commit. I don't see anything functionally wrong in the indexing policy implementation as it is. Rather I agree the current interface makes its usage very weird for tables (instead of caches) because of the size/entry_size different meaning. But setting the (entry_size=1) and (size = assoc) effectively makes it work as a fully associative indexing policy which is what you are hardcoding in this PR
You can see this in the BaseIndexingPolicy constructor

https://github.com/gem5/gem5/blob/stable/src/mem/cache/tags/indexing_policies/base.cc#L60C1-L60C48

      numSets(p.size / (p.entry_size * assoc)),
      setShift(floorLog2(p.entry_size)), setMask(numSets - 1), sets(numSets),

Becomes

      numSets(1),
      setShift(0), setMask(0), sets(1),


Anyway I don't want this to be a blocker and I'd rather have this contribution merged rather than forcing you to reuse existing code.
As you are likely running some experiments to check the performance implementation of these prefetchers, you are probably the best person to make the proposed changes while ensuring there is no performance regression

Harshil2107:
Hi @AbhishekUoR and @Setu-Gupta,

Thank you for your contribution! Could you please update us on the status of this PR? I noticed that PRs #1461, #1454, and #1441 seem to be related and add prefetchers. It would be helpful if you could create a discussion post to track the status of these PRs and inform the community about the changes. Additionally, we have a gem5 developer's meeting next Thursday, October 10th. It would be great if you could join us to discuss these updates.

giactra:
Hi @AbhishekUoR, do you think you will have some time to review my change requests?
If not, as I believe changes are minimal, we can consider merging this as is and we will update the changes ourselves with a separate PR. Let me know what you think about it

AbhishekUoR:
Hello,

Thank you for making the changes. I reviewed them and they look good.

Thanks,
Abhishek Singh


Remember "Chance favors the prepared mind."(Pasteur)

giactra (approved):

PR #1464, opened 2024-08-15
arch-riscv: Fix GDB breakpoint addr issue after #1316

### LINKED ISSUE #1463: GDB breakpoint not work when firmware address begins 0x80000000 in RV32 mode
**Describe the bug**
gdb continue working when set breakpoint 

**Affects version**
develop branch

**gem5 Modifications**
Add `system.workload.wait_for_remote_gdb = args.wait_gdb` before `root = Root(...)` in `fs_linux.py`

**To Reproduce**
1. Create `rot13.c` and `rot13.ld` anywhere

`rot13.c`
```c
char text[] = "Vafgehpgvba frgf jnag gb or serr!";

// Don't use the stack, because sp isn't set up.
volatile int wait = 1;

int main()
{
    while (wait)
        ;

    // Doesn't actually go on the stack, because there are lots of GPRs.
    int i = 0;
    while (text[i]) {
        char lower = text[i] | 32;
        if (lower >= 'a' && lower <= 'm')
            text[i] += 13;
        else if (lower > 'm' && lower <= 'z')
            text[i] -= 13;
        i++;
    }

done:
    while (!wait)
        ;
}
```
`rot13.ld`
```
OUTPUT_ARCH( "riscv" )
ENTRY(main)

SECTIONS
{
  . = 0x80000000;
  .text : { *(.text) }
  .data : { *(.data) }
}
```

2. Compile the `rot13.c`

```bash
riscv64-unknown-elf-gcc -static -nostdlib -g -Og -nostartfiles -ffreestanding -march=rv32ic_zicsr -mabi=ilp32 -Wl,-Trot13.ld rot13.c -o rot13.elf
```

3. Run script

```bash
build/ALL/gem5.opt configs/example/riscv/fs_linux.py --bare-metal --kernel rot13.elf --riscv-32bits --wait-gdb
```

You need to open GDB to let gem5 continue
4. Open GDB

```bash
riscv64-unknown-elf-gdb rot13.elf
```

5. Run commands in GDB

```bash
target remote :7000
si
si
si
si
set wait = 0
b *0x8000000c
c
```

**Expected behavior**
gdb should stop at the breakpoint

**Host ISA**
X86

**Compiler used**
gcc 13.2

rogerchang23424:
The PR https://github.com/gem5/gem5/pull/1316 changes to set pc with sign-extend address in order to support run RV32 program in RV64 architecture. However, it will let GDB with inconsistent address with gem5 PCState (0x80001000 in GDB and 0xffffffff80001000 in gem5 PCState).

In the PR, we will start define the difference rvType in PCState or ISA and machInst.rv_type. The rvType should define which XLEN is used for architecture, and machInst.rv_type is used to define XLEN from current privilege model (xstatus.sxl in supervisor, and xstatus.uxl in user).

gem5 issue: https://github.com/gem5/gem5/issues/1463
Change-Id: Iac56b47b89e2402f1af3a12583c08d5078c7b677

### DIFF
--- a/src/arch/riscv/pcstate.hh
+++ b/src/arch/riscv/pcstate.hh
@@ -75,7 +75,7 @@ class PCState : public GenericISA::UPCState<4>
         _vtype(other._vtype), _vl(other._vl)
     {}
     PCState &operator=(const PCState &other) = default;
-    PCState() = default;
+    PCState() { _rvType = RV64; };
     explicit PCState(Addr addr) { set(addr); }
     explicit PCState(Addr addr, RiscvType rvType, uint64_t vlenb)
     {
@@ -91,11 +91,25 @@ class PCState : public GenericISA::UPCState<4>
     {
         Base::update(other);
         auto &pcstate = other.as<PCState>();
+        _pc = zext(pcstate._pc);
+        _npc = zext(pcstate._npc);
         _compressed = pcstate._compressed;
         _rvType = pcstate._rvType;
         _vlenb = pcstate._vlenb;
         _vtype = pcstate._vtype;
         _vl = pcstate._vl;
+    }
+
+    void
+    set(Addr val) override
+    {
+        Base::set(zext(val));
+    }
+
+    void
+    setNPC(Addr val)
+    {
+        Base::setNPC(zext(val));
     }
 
     void compressed(bool c) { _compressed = c; }
@@ -114,6 +128,13 @@ class PCState : public GenericISA::UPCState<4>
     uint32_t vl() const { return _vl; }
 
     uint64_t size() const { return _compressed ? 2 : 4; }
+
+    Addr zext(Addr val) const {
+        if (_rvType == RV32) {
+          val = bits(val, 31, 0);
+        }
+        return val;
+    }
 
     bool
     branching() const override


cyyself:
Sorry for this trouble caused. For me, I would like to sign-extend everything, including `PC`, to reduce the design complexity of modifiable XLEN. When setting breakpoints, we should also use the address that is being sign-extended.

src/arch/riscv/pcstate.hh
```diff
@@ -75,7 +75,7 @@ class PCState : public GenericISA::UPCState<4>
         _vtype(other._vtype), _vl(other._vl)
     {}
     PCState &operator=(const PCState &other) = default;
-    PCState() = default;
+    PCState() { _rvType = RV64; };
```
powerjg, line 78: I don't understand the need for this change. On line 67, we set `_rvType` to 64 by default. Could you explain why this line is needed?

powerjg (commented):
One small thing below. Otherwise, it looks good to me.

cyyself:
I think fixing in this way would be better to match the ISA specification: https://github.com/gem5/gem5/pull/1470

PR #1467, opened 2024-08-15
stdlib: Give user's disk_device priority when setting root val

nkrim:
In `get_default_kernel_root_val()`, now prioiritizes the explicit disk_device passed from the user over the default implemented by the board.

Also adjusts syntax for selecting this value in `set_kernel_disk_workload()` for consistency. 

It seems that the common use case for setting `disk_device` is that there is a mismatch between where the disk image is mounted and where the board expects it by default. In this case, it also seems common that the root partition will be on this explicit device as well. 

In cases where this is not true, explicit kernel arguments can be used to define the distinct disk device apart from the root. However, this seems less common than the above so, in that case, it would be easier to tie these together.


hnpl (approved):
LGTM. Thanks for the contribution.

BobbyRBruce (approved):
This is a nice quick-fix to this problem. I do think we need to rework this code quite a bit and come up with something more user friendly, but this is a step in the right direction.

PR #1470, opened 2024-08-16
arch-riscv: fix GDB breakpoint issue for RV32

cyyself:
Since PR #1316, we use sign-extend for all address generation, including PC, to match the ISA specification for modifiable XLEN. However, when we set a breakpoint using remote GDB, our address is not sign-extended. This causes the breakpoint to be set at the wrong address, as specified in Issue #1463. This PR fixes the issue by sign-extending the address when setting a breakpoint. This also matches the RISC-V ISA Specification that "must sign-extend results to fill the entire widest supported XLEN in the destination register."

Change-Id: I9b493bf8ad5b1ef45a9728bb40fc5e38250fe9c3

### DIFF
--- a/src/arch/riscv/remote_gdb.cc
+++ b/src/arch/riscv/remote_gdb.cc
@@ -234,6 +234,20 @@ RemoteGDB::acc(Addr va, size_t len)
 }
 
 void
+RemoteGDB::insertHardBreak(Addr addr, size_t kind)
+{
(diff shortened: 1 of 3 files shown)


rogerchang23424:
Thanks for fixing the issue. The method makes easier to solve the RV32 GDB issue without changing the rvType behavior in PCState and Decoder. I will close the issue https://github.com/gem5/gem5/pull/1464.

powerjg (approved):

rogerchang23424 (approved):

PR #1474, opened 2024-08-17
sim-se: Replaces std::list with std::multiset for _vmaList

### LINKED ISSUE #1468: Making the implementation of mmap/munmap/mremap/fixupFault more time-efficient
The time complexity of several methods in SE mode related to virtual memory page management is currently O(n), where n is the amount of pages. This is because gem5 keeps track of all currently mapped virtual memory pages in a simple linked list `std::list<VMA> gem5::MemState::_vmaList`, so whenever gem5 needs to find a page, it needs to linearly iterate over all `VMA` objects in the linked list. This operation is O(n) in time complexity and can happen in system calls including `mmap`, `munmap` and `mremap`. (It also happens when `MemState::fixupFault` is called for page faults in SE mode.) Therefore, for executables that map plenty of memory pages, the total overhead is unbearable (O(nn)).

I have worked out a solution to bring the time complexity down to O(logn) for this operation. The changes are confined to just a few files and are minimally disruptive. Would the community be interested in reviewing a PR for this fix?

FJShen:
Several operations (mapRegion/UnmapRegion/fixupFault) involving the _vmaList had time complexity O(n), where n is the number of currently mapped virtual memory pages. This was because the list of VMA objects had to be linearly iterated for each search.

These operations could be sped up to O(logn) using a full-fledged Interval Tree, but std::multiset also does the job. This is because for any VMA object in _vmaList, (1) its AddrRange is never interleaved (AddrRange::_masks.size()==0) and (2) its AddrRange never overlaps with that of any other VMA object in the list (MemState::mapRegion() avoids creating a region that has preexisting mappings). This means a strict weak ordering of VMA objects, which is necessary for std::multiset to behave properly, can be established by simply comparing their start and end addresses in the current use case of _vmaList.

This PR assumes MemState::remapRegion() never remaps a region to a preexisting mapping (similar to MemState::mapRegion()).

If VMA objects in the _vmaList have an AddrRange that is interleaved or overlaps with each other's intervals, then this PR becomes incorrect.

github issue: #1468

### DIFF
--- a/src/sim/mem_state.cc
+++ b/src/sim/mem_state.cc
@@ -78,14 +78,45 @@ MemState::resetOwner(Process *owner)
     _ownerProcess = owner;
 }
 
+std::pair<MemState::OrderedVMAList::iterator,
+        MemState::OrderedVMAList::iterator>
+MemState::getListOfOvalappingVMA(Addr start_addr, Addr length){
+    Addr end_addr = start_addr + length;
+    const AddrRange range(start_addr, end_addr);
+
+    // We want to find a range of all VMA objects that have a chance of
+    // intersecting with the queried range
+    VMA dummy_vma{range, _pageBytes};
+    auto lb = _vmaList.lower_bound(dummy_vma);
+    auto ub = _vmaList.upper_bound(dummy_vma);
+
+    // The lower/upper-bounds are not sufficient when we use
+    // VMA::StrictWeakOrder to compare them. We have to grow the range.
+    while (lb!=_vmaList.begin() && lb->end() > range.start()) {
+        lb--;
+    }
+
+    while (ub!=_vmaList.end() && ub->start() < range.end()) {
+        ub++;
+    }
+
+    return {lb, ub};
+}
+
 bool
 MemState::isUnmapped(Addr start_addr, Addr length)
 {
     Addr end_addr = start_addr + length;
     const AddrRange range(start_addr, end_addr);
-    for (const auto &vma : _vmaList) {
-        if (vma.intersects(range))
-            return false;
+
+    // We want to find a range of all VMA objects that have a chance of
+    // intersecting with the queried range
+    auto p = getListOfOvalappingVMA(start_addr, length);
+    auto lb = p.first;
+    auto ub = p.second;
+
+    for (auto vma_it = lb; vma_it != ub; vma_it++) {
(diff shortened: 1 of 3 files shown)


ivanaamit:
Hi @FJShen ,

```
f=.git/hooks/commit-msg
mkdir -p $f
chmod +x $f
```

FJShen:
Thanks for the explanation. I'll add the change-id.

BobbyRBruce (commented):
I may be wrong but could any of this code be replaced by that already provided by the `AddressRangeMap` class? [src/base/addr_range_map.hh](https://github.com/gem5/gem5/blob/b1a44b89c7bae73fae2dc547bc1f871452075b85/src/base/addr_range_map.hh#L140)

FJShen:
@BobbyRBruce I just gave it a look. In principle `class AddressRangeMap` should be able to do the job as well, although some changes need to be made. For example, I feel the exit condition for the do...while loop in the private method `find` should be relaxed for correct behavior if we use this class for `_vmaList`: `mergesWith` is too strict, we really want to find any `AddrRange` that has an overlap.

BobbyRBruce:
> @BobbyRBruce I just gave it a look. In principle `class AddressRangeMap` should be able to do the job as well, although some changes need to be made. For example, I feel the exit condition for the do...while loop in the private method `find` should be relaxed for correct behavior if we use this class for `_vmaList`: `mergesWith` is too strict, we really want to find any `AddrRange` that has an overlap.

I think it would be really useful and beneficial to use the `AddressRangeMap` as much as possible. If that class needs extended then that's fine but this class should be the go-to to help with address ranges.

FJShen:
Thanks for the suggestion! I will look into this on another day.

erin-le:
@FJShen Hello, could you provide an update on what the status of this PR is? Thank you!

FJShen:
> @FJShen Hello, could you provide an update on what the status of this PR is? Thank you!

I am not actively working on this PR at this moment. The PR in its current form works indeed, and I have been benefiting from its timing-saving capability in my work.  I am just waiting for a leisure time to refactor and merge the code into the existing `AddressRangeMap` class as suggested by Bobby.

erin-le:
@FJShen Hello, this PR is being closed due to inactivity, but please feel free to reopen it once you have time to continue working on it. Thank you!

PR #1478, opened 2024-08-19
arch-arm: Add arm demo board

Harshil2107:
This demo board is a preset arm board, that can be used to run example gem5 simulations. This board doesnt simulate any known hardware.

The board will be used to run benchmarks such as gapbs and npb to collect stats. The plan is to show these stats on the gem5 resources website to provide more details about the resources.

### DIFF
--- a/configs/example/gem5_library/arm-demo-ubuntu-run.py
+++ b/configs/example/gem5_library/arm-demo-ubuntu-run.py
@@ -1,0 +1,72 @@
+# Copyright (c) 2024 The Regents of the University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
+This script further shows an example of booting an ARM based full system Ubuntu
+disk image. This simulation boots the disk image using the ArmDemoBoard.
+
+Usage
+-----
+
+```
+scons build/ARM/gem5.opt -j<NUM_CPUS>
+./build/ARM/gem5.opt configs/example/gem5_library/arm-demo-ubuntu-run.py
+
+"""
+import argparse
+
+from gem5.isas import ISA
+from gem5.prebuilt.demo.arm_demo_board import ArmDemoBoard
+from gem5.resources.resource import obtain_resource
+from gem5.simulate.simulator import Simulator
+from gem5.utils.requires import requires
+
+# This runs a check to ensure the gem5 binary is compiled for ARM.
+requires(isa_required=ISA.ARM)
+
+parser = argparse.ArgumentParser(
+    description="An example configuration script to run the arm demo mode."
+)
+
+parser.add_argument(
+    "--use-kvm",
+    action="store_true",
+    help="Use KVM for full system simulation.",
+)
+args = parser.parse_args()
+
+board = ArmDemoBoard(use_kvm=args.use_kvm)
+
+
+board.set_workload(
+    obtain_resource("arm64-ubuntu-20.04-boot", resource_version="2.0.0")
+)
+# We define the system with the aforementioned system defined.
+simulator = Simulator(
+    board=board,
+)
+
+simulator.run()
--- a/configs/example/gem5_library/arm-demo-ubuntu-run.py
+++ b/configs/example/gem5_library/arm-demo-ubuntu-run.py
@@ -41,6 +41,7 @@ import argparse
 from gem5.isas import ISA
 from gem5.prebuilt.demo.arm_demo_board import ArmDemoBoard
 from gem5.resources.resource import obtain_resource
+from gem5.simulate.exit_event import ExitEvent
 from gem5.simulate.simulator import Simulator
 from gem5.utils.requires import requires
 
@@ -60,13 +61,32 @@ args = parser.parse_args()
 
 board = ArmDemoBoard(use_kvm=args.use_kvm)
 
-
 board.set_workload(
-    obtain_resource("arm64-ubuntu-20.04-boot", resource_version="2.0.0")
+    obtain_resource(
+        "arm-ubuntu-24.04-boot-with-systemd", resource_version="2.0.0"
+    )
 )
+
+
+def exit_event_handler():
+    print("First exit: kernel booted")
+    yield False  # gem5 is now executing systemd startup
+    print("Second exit: Started `after_boot.sh` script")
+    # The after_boot.sh script is executed after the kernel and systemd have
+    # booted.
+    yield False  # gem5 is now executing the `after_boot.sh` script
+    print("Third exit: Finished `after_boot.sh` script")
+    # The after_boot.sh script will run a script if it is passed via
+    # m5 readfile. This is the last exit event before the simulation exits.
+    yield True
+
+
 # We define the system with the aforementioned system defined.
 simulator = Simulator(
     board=board,
+    on_exit_event={
+        ExitEvent.EXIT: exit_event_handler(),
+    },
 )
 
 simulator.run()
--- a/configs/example/gem5_library/arm-demo-ubuntu-run.py
+++ b/configs/example/gem5_library/arm-demo-ubuntu-run.py
@@ -55,7 +55,7 @@ parser = argparse.ArgumentParser(
 parser.add_argument(
     "--use-kvm",
     action="store_true",
-    help="Use KVM for full system simulation.",
+    help="Use KVM cores instead of Timing.",
 )
 args = parser.parse_args()
 
--- a/configs/example/gem5_library/arm-demo-ubuntu-run.py
+++ b/configs/example/gem5_library/arm-demo-ubuntu-run.py
@@ -31,10 +31,10 @@ disk image. This simulation boots the disk image using the ArmDemoBoard.
 Usage
 -----
 
-```
-scons build/ARM/gem5.opt -j<NUM_CPUS>
+```bash
+scons build/ARM/gem5.opt -j $(nproc)
 ./build/ARM/gem5.opt configs/example/gem5_library/arm-demo-ubuntu-run.py
-
+```
 """
 import argparse
 
@@ -49,7 +49,7 @@ from gem5.utils.requires import requires
 requires(isa_required=ISA.ARM)
 
 parser = argparse.ArgumentParser(
-    description="An example configuration script to run the arm demo mode."
+    description="An example configuration script to run the ArmDemoBoard."
 )
 
 parser.add_argument(
--- a/configs/example/gem5_library/arm-demo-ubuntu-run.py
+++ b/configs/example/gem5_library/arm-demo-ubuntu-run.py
@@ -45,7 +45,7 @@ from gem5.simulate.exit_event import ExitEvent
 from gem5.simulate.simulator import Simulator
 from gem5.utils.requires import requires
 
-# This runs a check to ensure the gem5 binary is compiled for ARM.
+# This runs a check to ensure the gem5 binary interpreting this file is compiled to include the ARM ISA.
 requires(isa_required=ISA.ARM)
 
 parser = argparse.ArgumentParser(
--- a/src/python/gem5/prebuilt/demo/arm_demo_board.py
+++ b/src/python/gem5/prebuilt/demo/arm_demo_board.py
@@ -94,7 +94,7 @@ class ArmDemoBoard(ArmBoard):
 
         else:
             processor = SimpleProcessor(
-                cpu_type=CPUTypes.TIMING, num_cores=4, isa=ISA.ARM
+                cpu_type=CPUTypes.TIMING, num_cores=2, isa=ISA.ARM
             )
             release = ArmDefaultRelease()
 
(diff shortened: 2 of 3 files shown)


giactra:
Thanks @Harshil2107, can I ask you to provide some context in the PR on how the demo board is supposed to be used and how it differs with the current ArmBoard?

Harshil2107:
I would describe this arm board as a preset version of the ArmBoard that has everything from memory, process, etc set up. It does not represent an actual system but acts as a demo board. This board is just a demo board that can be used for examples or tutorials.

The reason for making this board is that we want to run workloads like npb, gapbs that are in gem5 resources to get some stats. Future plan is to show these stats on the resources website, to provide more information about the resources.

powerjg:
I had a comment here, I thought, but it's gone now.

Anyway, I need to give this more context at the next developer's meeting. Sorry for not doing this earlier.

Having a set of "prebuilt" boards which have *no parameters* is needed to start providing gem5 simulated performance data on the gem5-resources webpage. It plays into plans to have "known-good" configurations which model real systems. 

For now, the goal is just to have *something* even if it's not modeling anything real, just to show that the infrastructure works. Then, we (and others in the community) can start working on developing boards that match specific hardware.

erin-le (dismissed):
Lgtm

giactra:
> I had a comment here, I thought, but it's gone now.
> 
> Anyway, I need to give this more context at the next developer's meeting. Sorry for not doing this earlier.
> 
> Having a set of "prebuilt" boards which have _no parameters_ is needed to start providing gem5 simulated performance data on the gem5-resources webpage. It plays into plans to have "known-good" configurations which model real systems.
> 
> For now, the goal is just to have _something_ even if it's not modeling anything real, just to show that the infrastructure works. Then, we (and others in the community) can start working on developing boards that match specific hardware.

Thanks Jason, I am looking forward to discuss this at the next gem5 dev meeting. In the meantime, I would like the description of the purpose to be present in the PR message

configs/example/gem5_library/arm-demo-ubuntu-run.py
```diff
@@ -0,0 +1,92 @@
...
+disk image. This simulation boots the disk image using the ArmDemoBoard.
+
+Usage
+-----
+
+```
+scons build/ARM/gem5.opt -j<NUM_CPUS>
+./build/ARM/gem5.opt configs/example/gem5_library/arm-demo-ubuntu-run.py
+
+"""
```
BobbyRBruce: The verbatim block isn't closed.

I'd also use ` ```shell ` or ` ```bash ` to start so it formats nicely.

configs/example/gem5_library/arm-demo-ubuntu-run.py
```diff
@@ -0,0 +1,92 @@
...
+This script further shows an example of booting an ARM based full system Ubuntu
+disk image. This simulation boots the disk image using the ArmDemoBoard.
+
+Usage
+-----
+
+```
+scons build/ARM/gem5.opt -j<NUM_CPUS>
```
BobbyRBruce: I don't like sayting `NUM_CPUS` for three reasons:

1. It might confuse people into thinking this value has something to do with what we want to simulate. We usually ask for number of CPUs to simulate. Best just avoid this.
2. It's not the number of CPUs you have, or the number of CPUs you want to use concurrently. It's the number of _threads_. So the name is misleading.
3. Where possible i like scripts to be copy-and-pastable(-and-executable) without error or edits needed.

I opt for just `scons build/ARM/gem5.opt -j $(nproc)`. It's clear enough.

configs/example/gem5_library/arm-demo-ubuntu-run.py
```diff
@@ -0,0 +1,92 @@
+# Copyright (c) 2024 The Regents of the University of California
```
BobbyRBruce, line 1: If we're going with the "arm-demo-ubuntu-run.py" naming conveniton can we update the "x86-ubuntu-run.py" name to match?

configs/example/gem5_library/arm-demo-ubuntu-run.py
```diff
@@ -0,0 +1,92 @@
...
+from gem5.resources.resource import obtain_resource
+from gem5.simulate.exit_event import ExitEvent
+from gem5.simulate.simulator import Simulator
+from gem5.utils.requires import requires
+
+# This runs a check to ensure the gem5 binary is compiled for ARM.
+requires(isa_required=ISA.ARM)
+
+parser = argparse.ArgumentParser(
+    description="An example configuration script to run the arm demo mode."
```
BobbyRBruce: "arm" -> "ARM".

BobbyRBruce: Is this the "arm demo mode" or the "ARM Demo Board"? You should be consistent in what you're telling people your running.

Also if the "Demo Board" is in the name then it should be capitalized.

configs/example/gem5_library/arm-demo-ubuntu-run.py
```diff
@@ -0,0 +1,92 @@
...
+)
+
+parser.add_argument(
+    "--use-kvm",
+    action="store_true",
+    help="Use KVM for full system simulation.",
```
BobbyRBruce: This comment suggests this script/board supports SE mode. I don't see any evidence of that here.

Didn't we need to have this working wit SE mode?

configs/example/gem5_library/arm-demo-ubuntu-run.py
```diff
@@ -0,0 +1,92 @@
...
+)
+args = parser.parse_args()
+
+board = ArmDemoBoard(use_kvm=args.use_kvm)
+
+board.set_workload(
+    obtain_resource(
+        "arm-ubuntu-24.04-boot-with-systemd", resource_version="2.0.0"
```
BobbyRBruce, line 66: Why is this `with-systemd` qualifier needed? If I run `arm-ubuntu-24.04-boot` i'd expect systemd to run. It's `without-sysyem` that should be specifically stated.

This isn't a deal breaker, I'm just asking for a bit more create over namings this. This may confuse people a bit.

src/python/gem5/prebuilt/demo/arm_demo_board.py
```diff
@@ -0,0 +1,109 @@
...
+        warn(
+            "The ARMDemoBoard is solely for demonstration purposes. "
+            "This board is not known to be be representative of any "
+            "real-world system. Use with caution."
+        )
+        cache_hierarchy = PrivateL1SharedL2CacheHierarchy(
+            l1d_size="16KiB", l1i_size="16KiB", l2_size="8MiB"
+        )
+
+        memory = DualChannelDDR4_2400(size="8GB")
```
BobbyRBruce: This should be "8GiB"

src/python/gem5/prebuilt/demo/arm_demo_board.py
```diff
@@ -0,0 +1,109 @@
...
+            isa_required=ISA.ARM,
+        )
+
+        warn(
+            "The ARMDemoBoard is solely for demonstration purposes. "
+            "This board is not known to be be representative of any "
+            "real-world system. Use with caution."
+        )
+        cache_hierarchy = PrivateL1SharedL2CacheHierarchy(
+            l1d_size="16KiB", l1i_size="16KiB", l2_size="8MiB"
```
BobbyRBruce: I can't really see these numbers making any sense with a system that has "8GB" memory (should be "GiB") of memory and running at an very high 3GHz clock frewuency. 

Here's some data I got by reading the description from an Apple A7 processor:

* Clock frequency of 1.4GHz
* Using ARM v8.0-A
* 2 cores
* L1 Private cache: 64KiB inst cache & 64KiB data cache 
* L2 Shared cache: 1MiB
* L3 Shared cache: 4MiB

Can we just use these numbers?

BobbyRBruce: **Note**: This would normally only get 1GiB of memory but since this may block us running some bench marks we can probably up this 4GB and not spoint things too much.

src/python/gem5/prebuilt/demo/arm_demo_board.py
```diff
@@ -0,0 +1,109 @@
...
+class ArmDemoBoard(ArmBoard):
+    """
+    This prebuilt ARM board is used for demonstration purposes. It simulates an
+    ARM 3GHz dual-core system with a 2GB DDR4_2400 memory system. A
+    PrivateL1PrivateL2 cache hierarchy is set with an l1 data and instruction
+    cache, each 16kB, and a single bank l2 cache of 256kB.
+
+    **DISCLAIMER**: This board is solely for demonstration purposes. This board
+    is not known to be representative of any real-world system or produce
+    reliable statistical results.
```
BobbyRBruce: Start again with this. Almost everything written is wrong as far as I can see. Just a few errors I can see immiediately:

1. You're not using the base-2 SI Units.
2.  We're not seeing 2GB (or 2GiB) of memory.
3. The cache sizes are wrong.

src/python/gem5/prebuilt/demo/arm_demo_board.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+
+    def __init__(self, use_kvm=False):
```
BobbyRBruce: Please remember to use typing:

```python
    def __init__(self, use_kvm : bool = False) -> None:
```

BobbyRBruce (changes requested):

configs/example/gem5_library/arm-demo-ubuntu-run.py
Harshil2107: it should be board, I will update it

configs/example/gem5_library/arm-demo-ubuntu-run.py
Harshil2107: This was just to make the workloads explicit, one boots with systemd and other without.

src/python/gem5/prebuilt/demo/arm_demo_board.py
Harshil2107: so I will update the cache to be 64KiB for l1d and l1i and l2 shared to be 1MiB. I am not to clear on the second statement? do you mean update l2 cache to 4MiB instead of 1MiB?

configs/example/gem5_library/arm-demo-ubuntu-run.py
Harshil2107: I think the plan is to get this in and then later make the board compatible with SE mode.

configs/example/gem5_library/arm-demo-ubuntu-run.py
Harshil2107: I think we should rename a few other examples as well to make it more consistent, like `riscv-fs`, `riscvmatched-fs`, etc. I can make a separate PR for the name changes.

src/python/gem5/prebuilt/demo/arm_demo_board.py
```diff
@@ -0,0 +1,112 @@
...
+            release = ArmDefaultRelease.for_kvm()
+
+            # The platform sets up the memory ranges of all the on-chip and off-chip
+            # devices present on the ARM system. ARM KVM only works with VExpress_GEM5_V1
+            # on the ArmBoard at the moment.
+            platform = VExpress_GEM5_V1()
+
+        else:
+            processor = SimpleProcessor(
+                cpu_type=CPUTypes.TIMING, num_cores=4, isa=ISA.ARM
```
BobbyRBruce: In one setting the `num_cores` is 2, in the other it is 4. This should be consistent.

configs/example/gem5_library/arm-demo-ubuntu-run.py
BobbyRBruce: In which case can we say "Use KVM Cores instead of Timing." ?

configs/example/gem5_library/arm-demo-ubuntu-run.py
```diff
@@ -0,0 +1,92 @@
...
+    print("First exit: kernel booted")
+    yield False  # gem5 is now executing systemd startup
+    print("Second exit: Started `after_boot.sh` script")
+    # The after_boot.sh script is executed after the kernel and systemd have
+    # booted.
+    yield False  # gem5 is now executing the `after_boot.sh` script
+    print("Third exit: Finished `after_boot.sh` script")
+    # The after_boot.sh script will run a script if it is passed via
+    # m5 readfile. This is the last exit event before the simulation exits.
+    yield True
```
BobbyRBruce, line 81: I do really dislike having this in the codebase. I think it's very confusing to users. When I get the exit events done please remember to refactor this away.

src/python/gem5/prebuilt/demo/arm_demo_board.py
```diff
@@ -0,0 +1,112 @@
...
+
+from gem5.components.boards.arm_board import ArmBoard
+from gem5.components.cachehierarchies.classic.private_l1_shared_l2_cache_hierarchy import (
+    PrivateL1SharedL2CacheHierarchy,
+)
+from gem5.components.memory import DualChannelDDR4_2400
+from gem5.components.processors.cpu_types import CPUTypes
+from gem5.components.processors.simple_processor import SimpleProcessor
+from gem5.isas import ISA
+from gem5.utils.requires import requires
```
BobbyRBruce: When within the stdlib we typically use relative import paths So it should be

```py
from ...components.memory import DualChannelDDR4_2400
```

instead of `from gem5.components.memory import DualCHannelDDR4_2400`,for example.

I remmeber when I started the stdlib I had a good reason for doing this which I've since forgetting what the reason was. Either way, it's since become the convention.

src/python/gem5/prebuilt/demo/arm_demo_board.py
```diff
@@ -0,0 +1,112 @@
...
+            isa_required=ISA.ARM,
+        )
+
+        warn(
+            "The ARMDemoBoard is solely for demonstration purposes. "
+            "This board is not known to be be representative of any "
+            "real-world system. Use with caution."
+        )
+        cache_hierarchy = PrivateL1SharedL2CacheHierarchy(
+            l1d_size="64KiB", l1i_size="64KiB", l2_size="1MiB"
```
BobbyRBruce: I did base these numbers on real a real chip but Jason wants the L2 to be 8MiB. Can you update it? 

Remember to update the comment in this file when you do this.

configs/example/gem5_library/arm-demo-ubuntu-run.py
```diff
@@ -0,0 +1,92 @@
...
+import argparse
+
+from gem5.isas import ISA
+from gem5.prebuilt.demo.arm_demo_board import ArmDemoBoard
+from gem5.resources.resource import obtain_resource
+from gem5.simulate.exit_event import ExitEvent
+from gem5.simulate.simulator import Simulator
+from gem5.utils.requires import requires
+
+# This runs a check to ensure the gem5 binary is compiled for ARM.
```
BobbyRBruce:
```suggestion
# This runs a check to ensure the gem5 binary interpreting this file is compiled to include the ARM ISA.
```

BobbyRBruce (changes requested):
It's really pretty much there but I do have a few, small , annoying things to change.

src/python/gem5/prebuilt/demo/arm_demo_board.py
```diff
@@ -0,0 +1,112 @@
...
+                cpu_type=CPUTypes.TIMING, num_cores=2, isa=ISA.ARM
+            )
+            release = ArmDefaultRelease()
+
+            # The platform sets up the memory ranges of all the on-chip and off-chip
+            # devices present on the ARM system.
+            platform = VExpress_GEM5_Foundation()
+
+        super().__init__(
+            clk_freq="1.4GHz",
```
BobbyRBruce: Increase the frequency to 3GHz. Jason didn't like this being so and it does increase simulation time. Remember to update the docsrings and comments

src/python/gem5/prebuilt/demo/arm_demo_board.py
```diff
@@ -0,0 +1,112 @@
...
+from ...isas import ISA
+from ...utils.requires import requires
+
+
+class ArmDemoBoard(ArmBoard):
+    """
+    This prebuilt ARM board is used for demonstration purposes. It simulates an
+    ARM 1.4GHz dual-core system with a 4GiB DDR4_2400 memory system. It uses
+    a PrivateL1SharedL2CacheHierarchy with l1d and l1i caches set to 64KiB and
+    l2 shared cache set to 1MiB
```
BobbyRBruce: This is inaccurate.

```suggestion
    l2 shared cache set to 8MiB
```

BobbyRBruce (dismissed):
Almost there! Just a couple of small things.

BobbyRBruce (approved):

PR #1479, opened 2024-08-19
stdlib, python: Add warning message and clarify binary vs metric units

erin-le:
This PR changes memory and cache sizes in various parts of the gem5 codebase to use binary units (e.g. KiB) instead of metric units (e.g. kB). This makes the codebase more consistent, as gem5 automatically converts memory and cache sizes that are in metric units to binary units. 

This PR also adds a warning message to let users know when an auto-conversion from base 10 to base 2 units occurs.

There were a few places in configs and in the comments of various files where I didn't change the metric units, as I couldn't figure out where the parameters with those units were being used.


src/python/m5/util/convert.py
```diff
@@ -258,6 +264,16 @@ def toMemoryBandwidth(value):
...
+    if (
+        type(value) is str
+        and len(value) > 1
+        and value[-2] in binary_prefixes.keys()
+        and not "i" in value
+    ):
+        print(
+            f"warn: Base 10 memory/cache size {value} will be cast to base 2"
+            + f" size {value[0:-2]}{value[-2].upper()}i{value[-1]}"
+        )
```
BobbyRBruce: Here's a cheap solution to your circular importing. 

You can import anywhere so just do it when execute this function, not when the module is loaded.

```suggestion
        from m5.util import warn
        warn(
            f"Base 10 memory/cache size {value} will be cast to base 2"
            + f" size {value[0:-2]}{value[-2].upper()}i{value[-1]}"
        )
```

src/python/m5/util/convert.py
```diff
@@ -258,6 +264,18 @@ def toMemoryBandwidth(value):
...
+        and len(value) > 1
+        and value[-2] in binary_prefixes.keys()
+        and not "i" in value
+    ):
+        from m5.util import warn
+
+        warn(
+            f"Base 10 memory/cache size {value} will be cast to base 2"
+            + f" size {value[0:-2]}{value[-2].upper()}i{value[-1]}."
+        )
```
BobbyRBruce: I'm not going to blog it on this but i think you could save yourself a lot of effort here by just having a prefix lookup in "src/python/m5/utility/convert.py". Something like:

```py
si_base_10_to_2 = {
    "kB" : "KiB",
    "MB" : "MiB",
     #...  etc.
}
```

then just:

```py
if get_prefix(value) in si_base_10_to_2:
   correct_prefix = si_base_10_to_2
   # ... etc.
```

I also think a function for extractin the units from the string would be handy as demoed in `get_prefix` above.

BobbyRBruce (dismissed):

src/python/m5/util/convert.py
```diff
@@ -37,6 +37,12 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+# Attempting to import `warn` results in an error about a circular import:
+# ImportError: cannot import name 'warn' from partially initialized module 'm5.util' (most likely due to a circular import) (/home/bees/gem5-3rd-worktree/src/python/m5/util/__init__.py):
+
+# from m5.util import warn
+# from . import warn
+
```
BobbyRBruce: This comment is no longer needed right? We fixed this?

src/python/m5/util/convert.py
```diff
@@ -118,6 +133,32 @@ def _split_suffix(value, suffixes):
     return (value[: -len(matches[0])], matches[0]) if matches else (value, "")
 
 
+def _split_suffix_by_prefix(value):
```
BobbyRBruce: I think there should be a way of just having one `_split_suffix` function but i'm getting very confused looking at this files and how it splits these string so am fine with having two functions for now.

src/python/m5/util/convert.py
```diff
@@ -118,6 +133,32 @@ def _split_suffix(value, suffixes):
...
+    :returns: A tuple of (value, suffix). Suffix is the empty string
+              if there is no match.
+
+    """
+    matches = [sfx for sfx in binary_prefixes.keys() if sfx in value]
+
+    if len(matches) == 2:  # e.g. matches both M and Mi
+        matches.sort(key=len, reverse=True)
+        matches.pop()
+        assert len(matches) <= 1
```
BobbyRBruce: This is a weird assert for two reasons:

1. If `len(matches) == 2` and we pop once then `matches` is always equal to `1`. I can't see anything here that'll ever result in this assert failing.
2. Why `<=` and not just `==`? This login implies there might be a state where `matches == 0` and, in this case, we want to continue in which case the `return` function will fail when it tries to access `matches[0'`.

BobbyRBruce (changes requested):
I've added a PR here demonstrating how to add tests to this: https://github.com/BobbyRBruce/gem5/commit/1ced85887a263966fc78965f198f8eb0a69fe5fe.

My requested changes to this PR is to add tests to ensure this works as intended.

src/python/m5/util/convert.py
BobbyRBruce: In fact, doesn't `_split_suffix(value, suffixes=binary_prefixes)` do this for you?

erin-le: _split_suffix only matches the ending, so there would need to be two additional lists for the B and the B/s endings, e.g. [kB, MB, ...] and [kB/s, MB/s,...]. This seemed worse to me than adding a helper function, but would it actually be preferable?

src/python/m5/util/convert.py
erin-le: Would it be best just to remove it? It was carried over from _split_suffix, so it looks like it's useful in _split_suffix but not here.

src/python/m5/util/convert.py
erin-le: It occurred to me that I could put values through _split_suffix twice instead of using _split_suffix_by_prefix. I think I am going to refactor the code so it does this.

erin-le: It occurred to me that I could put values through _split_suffix twice instead of using _split_suffix_by_prefix. I think I am going to refactor the code so it does this.

powerjg (approved):
lgtm. After merging this change, you should probably run grep again and look for any more mistakes in case they sneak in between you creating this change and when it's merged.

src/python/m5/util/convert.py
BobbyRBruce: This still needs addressed.

src/python/m5/util/convert.py
BobbyRBruce: Remove it or change it to `assert len(matches) == 1`. The latter is a pretty meaningless assert but can do no real harm so I'd go with that.

BobbyRBruce (changes requested):

BobbyRBruce (approved):

PR #1480, opened 2024-08-19
gem5 Bridge Driver

nkrim:
Adds a Linux kernel module for a character device that can facilitate address-type `gem5-bridge` commands with user permissions. This is an improvement over the current system as it no longer requires `sudo` permissions from the simulation as we replace accesses to `/dev/mem/` with accesses to `/dev/gem5_bridge` for which we can control permissions.

Relates to https://github.com/gem5/gem5-resources/pull/48

Adds `util/gem5_bridge` which includes the C source for the driver and a Makefile to build it from within the disk image -- building and launching the module are done here https://github.com/gem5/gem5-resources/pull/48. The Makefile can take arguments to modify the base address for the gem5ops MMIO range by putting `GEM5OPS_BASE=<base_phys_addr>` at the end of the make command.

Modifies `util/m5/src/m5_mmap.c` to mmap from our device `/dev/gem5_bridge` rather than `/dev/mem`. 

**Current Issues**
- Requires the module to be built with the exact same version as the running kernel during simulation. This is most easily achieved by running the simulation with the `vmlinux` kernel extracted from the built disk image. However, need to find a proper solution to support multiple kernel versions. If the kernel headers are available on `apt`, then this has been shown to work, but efforts to acquire the headers for an arbitrary kernel version have so far failed.

### DIFF
--- a/util/gem5_bridge/gem5_bridge.c
+++ b/util/gem5_bridge/gem5_bridge.c
@@ -1,0 +1,158 @@
+// SPDX-License-Identifier: BSD-3-Clause OR GPL-2.0-Only
+/*
+ * Copyright (c) 2024 The Regents of the University of California
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <linux/err.h>
+#include <linux/init.h>
+#include <linux/io.h>
+#include <linux/kernel.h>
+#include <linux/miscdevice.h>
+#include <linux/mm.h>
+#include <linux/module.h>
+#include <linux/moduleparam.h>
+#include <linux/uaccess.h>
+
+/* ========================================================================= *
+ * Constants
+ * ========================================================================= */
+
+#define DEV_NAME "gem5_bridge"
+#define DEV_MODE ((umode_t)0666) /* All users have RW access to this device */
+
+
+/* ========================================================================= *
+ * Device Parameters
+ * ========================================================================= */
+
+unsigned long gem5_bridge_baseaddr;
+module_param(gem5_bridge_baseaddr, long, 0644);
+MODULE_PARM_DESC(gem5_bridge_baseaddr,
+    "Physical base addr for gem5ops MMIO range");
+
+unsigned long gem5_bridge_rangesize;
+module_param(gem5_bridge_rangesize, long, 0644);
+MODULE_PARM_DESC(gem5_bridge_rangesize,
+    "Size of gem5ops MMIO range");
+
+
+/* ========================================================================= *
+ * File Operations
+ * ========================================================================= */
+
+static int gem5_bridge_open(struct inode *inode, struct file *file)
+{
+    pr_info("gem5_bridge_open: SUCCESS!\n");
+    return 0;
+}
+
+static int gem5_bridge_release(struct inode *inode, struct file *file)
+{
+    pr_info("gem5_bridge_release: SUCCESS!\n");
+    return 0;
+}
+
+static int gem5_bridge_mmap(struct file *file, struct vm_area_struct *vma)
+{
+    unsigned long vm_size = vma->vm_end - vma->vm_start + vma->vm_pgoff;
+    unsigned long gem5_bridge_pfn = gem5_bridge_baseaddr >> PAGE_SHIFT;
+
+    if (vm_size > gem5_bridge_rangesize) {
+        pr_err("%s: requested memory range is too large\n", __func__);
+        return -EINVAL;
+    }
+
+    /* Map the virtual memory area to our gem5 ops MMIO range */
+    if (io_remap_pfn_range(
+            vma, vma->vm_start, gem5_bridge_pfn, vm_size, vma->vm_page_prot
+    )) {
+        pr_info("gem5_bridge_mmap: failed to remap vm area to phys addr\n");
+        return -EAGAIN;
+    }
+
+    pr_info("gem5_bridge_mmap: SUCCESS!\n");
+    return 0;
+}
+
+static struct file_operations gem5_bridge_fops = {
+    .owner      = THIS_MODULE,
+    .open       = gem5_bridge_open,
+    .release    = gem5_bridge_release,
+    .mmap       = gem5_bridge_mmap,
+};
+
+/* ========================================================================= *
+ * Module Initialization + Destruction
+ * ========================================================================= */
+
+static struct miscdevice gem5_bridge_miscdevice = {
+    .minor = MISC_DYNAMIC_MINOR,
+    .name = DEV_NAME,
+    .mode = DEV_MODE,
+    .fops = &gem5_bridge_fops,
+};
+
+static int __init gem5_bridge_init(void)
+{
+    /* Ensure params were defined */
+    if (!gem5_bridge_baseaddr || !gem5_bridge_rangesize) {
+        pr_err("%s: some parameters undefined or invalid, must give non-zero "
+               "values for gem5_bridge_baseaddr and gem5_bridge_rangesize\n",
+               __func__);
+        return -1;
+    }
+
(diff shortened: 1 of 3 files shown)


util/gem5_bridge/src/gem5_bridge.h
```diff
@@ -0,0 +1,46 @@
+/* SPDX-License-Identifier: BSD AND GPL-2.0+ */
```
andysan: These licenses seem a bit strange to me. It seems like it should be BSD-3-Clause (there isn't a plain BSD license in the SPDX license list). The GPL license should probably be GPL-2.0-only to match Linux's license (you'll sometimes see GPL-2.0 as well, but that's deprecated).

It's also unclear to me if the intention is that the user has the choice to apply the GPL license or the BSD license, or if both licenses always apply. The SPDX 'AND' operator specifies the latter. If the intention is to allow either license, the correct identifier would be 'BSD-3-Clause OR GPL-2.0-Only'.

util/gem5_bridge/src/gem5_bridge.h
```diff
@@ -0,0 +1,46 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __GEM5_BRIDGE_H__
+#define __GEM5_BRIDGE_H__
+
+/* Physical address for base of gem5ops MMIO range */
+#ifndef GEM5OPS_BASE
+#define GEM5OPS_BASE 0xffff0000 /* default, also used for x86 */
```
andysan: This really ought to be discovered from the device tree or an ACPI table to make it a bit more robust. There is a good example here: https://elixir.bootlin.com/linux/v6.10.8/source/drivers/virtio/virtio_mmio.c

Given that gem5's ACPI implementation is very basic, it might be better and easier to just emulate a PCI device instead. That would be reasonably easy to implement on the gem5 side and would just work.

util/gem5_bridge/src/gem5_bridge.c
```diff
@@ -0,0 +1,344 @@
...
+    kvfree(kbuff);
+    return result;
+}
+
+static struct file_operations gem5_bridge_fops = {
+    .owner      = THIS_MODULE,
+    .open       = gem5_bridge_open,
+    .release    = gem5_bridge_release,
+    .mmap       = gem5_bridge_mmap,
+    .read       = gem5_bridge_read,
```
andysan: I would strongly suggest that we don't support read and write. It doesn't really make sense to use this interface since it forces a transition into the kernel which seems undesirable. It also adds unnecessary (including some pretty nasty inline assembly) complexity in the driver. The only reason I can see for this interface is to enable call directly from shell scripts that write to the individual op files, but that doesn't seem like justification enough to replicate this part of the m5 tool.

util/gem5_bridge/src/gem5_bridge.h
powerjg: I agree that using the devicetree/acpi would be cool, but I think that can be a second step.

util/gem5_bridge/src/gem5_bridge.h
nkrim: Fixed, thanks!

util/gem5_bridge/src/gem5_bridge.c
nkrim: While I agree with the sentiment, I feel that it is important to keep these read/write implementations here to have functionality that matches the current set of operations. The m5 utility will remain, especially for the purposes of linking against it to truly prevent overhead during workloads, but the driver will be a useful platform for moving forward with some of our ideas for new gem5ops interfaces without dealing with the complexity of the m5 utility -- which, despite the complexity here, is far more daunting to modify. Because of this, it is useful to exemplify how to replicate these other ops in the driver, so that future changes will have a basis to work from.

As far as the inline assembly, it is unavoidable even in the m5 utility with the current interface. Honestly, that is part of the reason for progressing the bridge interface as the current method of abusing the calling convention to pass all of the arguments is a bit untenable.

util/gem5_bridge/src/gem5_bridge.h
nkrim: This has been changed as well. The driver now accepts arguments for these upon insertion, which supports both default values as defined by the disk image init script (where the module is inserted) as well as kernel argument overloads that come from the sim config, if needed. Those supporting changes found here: https://github.com/gem5/gem5-resources/pull/48/files#diff-a8234537f3d3cb86d8b2ce71ff0f838fbbbda123548340a26a3a4593e454f260

In the future though, you are right that it would be nice for the driver to be given a range by the machine and communicate this range back to gem5.

andysan: I'd strongly suggest using the platform driver interface instead of ad-hoc parameters here. That interface is designed for devices that are discovered at boot time using ACPI, Device Tree, the platform boot code, or by passing the parameters as a module argument. The VirtIO MMIO driver is a very clean example and there is a comment describing how to pass the base address as a module parameter. Using this interface shouldn't add any complexity, and gives you DeviceTree and ACPI support for free. We would likely need to switch to this interface to support DT anyway, so doing it now avoids changing the module parameters at a later stage.

powerjg: Can you please explain why this request would be a blocker to getting this initial implementation merged? From my perspective, it's not an insignificant undertaking to learn how ACPI and DT works. This change currently adds a nice new feature to gem5, and it would be shame to block it due to it not being "nice enough." If I'm mistaken, and there is something that is truly blocking here, please explain in more detail.

I don't disagree at all that it would be nice to have device tree/ACPI support. But it seems to me that this can be done after the initial change is merged.

andysan: This [platform API](https://docs.kernel.org/driver-api/driver-model/platform.html) is designed for simple, typically memory mapped, devices and provides a standardised way of getting to the associated MMIOs and IRQs. If you use this API, you get more convenient  managed (see the [devres manual](https://docs.kernel.org/driver-api/driver-model/devres.html)) IO ranges and interrupts, this simplifies initialisation and clean up. The other big benefit is that Device Tree can both be supported by simply adding a table with the compatible strings. I.e., there is no need to actually understand any device tree APIs in the basic case, things will just work automatically since MMIO and interrupt assignment is standardised.

Contrary to my previous comment, the only drawback in this case seems to be that the non-DT/ACPI  will need custom parameter parsing. The [VirtIO MMIO](https://elixir.bootlin.com/linux/v6.10.8/source/drivers/virtio/virtio_mmio.c) has a very good example of that though.

In practice, I believe that this rewrite would practically be needed to support device tree anyway, so we might as well do it now.

util/gem5_bridge/src/gem5_bridge.c
andysan: If the M5 utility remains for low-overhead operation, what's the point of the read/write endpoints? With very few exceptions, I can't see how a transition to kernel space would have an acceptably low impact on simulation results for most calls. Effectively, we just end up with two implementations of the same thing. If you really want this functionality, is there any reason you can't just link with `libgem5` to avoid code duplication?

I'm also worried that putting a lot of complexity in the kernel driver is going to be a maintenance nightmare, especially since this will never be upstream. Historically, all of the gem5-specific kernel extensions have been poorly maintained. My feeling is that the more complexity we add here, the more likely it is that this module will end up with maintenance issues (more complexity means a higher likelihood that it won't compile on older / newer kernels).

util/gem5_bridge/src/gem5_bridge.h
andysan: I'd like to point out that this isn't a blocker. It's very much a nice to have that would make it much easier to support device tree properly in the future.

util/gem5_bridge/src/gem5_bridge.c
nkrim: Ok yeah, after discussing I think you're correct that, for an initial implementation, it should be as simple as possible until we know we need any other features. I am moving the prior code into a draft/reference PR here https://github.com/gem5/gem5/pull/1555 so that we still have the driver interface to experiment with.

andysan (commented):
Thanks for the changes to this PR, the updated PR looks good. There is something weird with the PR at the moment and there seem to be some unrelated changes that have been included for some reason that you might need to fix before this is merged.

There are some minor changes you could make if you feel like cleaning up the commit history (I don't think this is strictly necessary though):
- Squash the license update into 'util: Add gem5-bridge driver'
- Factor out the changes to 'm5_mmap.c' from 'util: Add gem5-bridge driver' to a separate change, potentially merging that with the fallback change.
- I'd probably merge the param change with the main bridge driver.

These changes would make it easier for people to cherry-pick your changes if they are working on private branches (e.g., experiment-specific branches), but they are really just nice-to-have changes.

I'm currently out of office so it's unlikely that I'll have time to review this again after it goes out of draft state. I'm happy with the change as is, so consider the above comments as nice to have features.

Harshil2107:
I tested this driver code with arm disk iamges and it works. I waas able to run npb benchmarks in gem5 without needing sudo 

https://github.com/gem5/gem5-resources/pull/60

nkrim:
> I tested this driver code with arm disk iamges and it works. I waas able to run npb benchmarks in gem5 without needing sudo
> 
> [gem5/gem5-resources#60](https://github.com/gem5/gem5-resources/pull/60)

That's great news!

@powerjg @BobbyRBruce I remember being pointed towards some existing tests to look at for reference to test the driver, but I don't remember which tests those are exactly so if you have any ideas please let me know. One worry I have is that, right now, the only way I can think of testing this with certainty is to have a custom disk image that includes a binary to call `m5_exit_addr`. Is that possible / worthwhile to set up just for this test, or is there a simpler way to go about this? Would it be enough to ensure that entering and exiting any KVM workload works fine (akin to just ensuring the `m5` utility hasn't failed). In the latter case, would the existing KVM tests suffice?

Also, I feel it's probably time to finally take this out of draft.

Harshil2107:
I think a good test would be to run a workload that uses the driver, so as we will be updating the disks anyway to include this, we can have the workloads run benchmarks without using sudo (like npb for example) and I feel that would be a good enough test.

I had a clarifying question:

- From the changes, it looks to me that kvm is not necessary. that is If we were to use the address mode exits in non KVM, it would still work?

Harshil2107 (commented):
Overall, these changes look good to me, I have tested them by adding the driver to base disk images and running a benchmark without using sudo in gem5.

nkrim:
> I think a good test would be to run a workload that uses the driver, so as we will be updating the disks anyway to include this, we can have the workloads run benchmarks without using sudo (like npb for example) and I feel that would be a good enough test.
> 
> I had a clarifying question:
> 
> * From the changes, it looks to me that kvm is not necessary. that is If we were to use the address mode exits in non KVM, it would still work?

Oh yes, I think this would be true, since I recall the KVM trap going through some address normalization process to understand the MMIO request. I'm not 100% sure, but I remember that it ends up falling into the same pathway that a non-KVM memory request would follow.

powerjg (approved):

util/gem5_bridge/gem5_bridge.c
```diff
@@ -0,0 +1,158 @@
+// SPDX-License-Identifier: BSD-3-Clause OR GPL-2.0-Only
```
BobbyRBruce, line 1: What's the reason for having two license here? Why GPL-2.0?

We use permissive licensing in gem5. BSD-3-Clause is typically what we use.

powerjg: I believe what Noah has here is correct. We have to have GPL because of the kernel. See https://www.kernel.org/doc/html/v6.13-rc6/process/license-rules.html#id1

PR #1481, opened 2024-08-20
dev-amdgpu: Implement UNMAP_QUEUES queue_sel==2

abmerop:
Unmap queues with queue_sel of 2 unmaps all queues while queue_sel of 3 unmaps all non-static queues. The implementation of 3 was actually correct for 2. Static queues are queues which were mapped using a map queues packet with a queue_type of 1 or 2.

This commit adds ability to mark a queue as static. When unmap queues with queue_sel of 2 is sent, the existing code is now executed. With a value of 3, we now check if the queue was marked static and do not unmap it if marked.

Change-Id: I87d7cf78a0600c7baa516c01f42c294d3c4e90c5

### DIFF
--- a/src/dev/amdgpu/pm4_packet_processor.cc
+++ b/src/dev/amdgpu/pm4_packet_processor.cc
@@ -518,8 +518,11 @@ PM4PacketProcessor::processSDMAMQD(PM4MapQueues *pkt, PM4Queue *q, Addr addr,
     assert(pkt->engineSel == 2 || pkt->engineSel == 3);
     SDMAEngine *sdma_eng = gpuDevice->getSDMAById(pkt->engineSel - 2);
 
+    // Queue type 1 and 2 are "static" queues
+    bool is_static = (pkt->queueType == 2) || (pkt->queueType == 3);
+
     // Register RLC queue with SDMA
-    sdma_eng->registerRLCQueue(pkt->doorbellOffset << 2, addr, mqd);
+    sdma_eng->registerRLCQueue(pkt->doorbellOffset << 2, addr, mqd, is_static);
 
     // Register doorbell with GPU device
     gpuDevice->setSDMAEngine(pkt->doorbellOffset << 2, sdma_eng);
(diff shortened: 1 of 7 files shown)


abmerop:
@v-ramadas See if this solves your issue

src/dev/amdgpu/pm4_packet_processor.cc
```diff
@@ -664,7 +664,37 @@ PM4PacketProcessor::unmapQueues(PM4Queue *q, PM4UnmapQueues *pkt)
...
+        gpuDevice->deallocateAllQueues(true);
+        } break;
+      case 3: {
+        auto &hsa_pp = gpuDevice->CP()->hsaPacketProc();
+        for (auto iter : gpuDevice->getUsedVMIDs()) {
+            for (auto id : iter.second) {
+                assert(queues.count(id));
+
+                // Do not unmap KMD queues or static queues
+                if (queues[id]->privileged() || queues[id]->isStatic()) {
```
mattsinc: It feels like a better solution to have a helper function both call instead of duplicating this code twice with one minor difference, but it also isn't wrong this way.

src/dev/amdgpu/pm4_packet_processor.cc
```diff
@@ -664,7 +664,37 @@ PM4PacketProcessor::unmapQueues(PM4Queue *q, PM4UnmapQueues *pkt)
...
+                    continue;
+                }
+                QueueDesc *mqd = queues[id]->getMQD();
+                DPRINTF(PM4PacketProcessor, "Unmapping queue %d with read "
+                        "index %ld\n", id, mqd->mqdReadIndex);
+                // Partially writing the mqd with an offset of 96 dwords
```
mattsinc: I know this isn't immediately from this commit, but it would help if we explained why 96 dwords

mattsinc (dismissed):

src/dev/amdgpu/pm4_packet_processor.cc
abmerop: Good idea. I think the only difference is the bool value sent to the device. Will have to get to this on Friday

abmerop: done

src/dev/amdgpu/pm4_packet_processor.cc
abmerop: done

src/dev/amdgpu/pm4_packet_processor.cc
```diff
@@ -586,6 +589,47 @@ PM4PacketProcessor::updateReadIndex(Addr offset, uint64_t rd_idx)
...
+    for (auto iter : gpuDevice->getUsedVMIDs()) {
+        for (auto id : iter.second) {
+            assert(queues.count(id));
+
+            // Do not unmap KMD queues.
+            if (queues[id]->privileged()) {
+                continue;
+            }
+
+            // Do not unmap static queues if requested.
```
mattsinc, line 605: @v-ramadas : it has nothing to do with this patch, but is this the code you've been looking for?

mattsinc (approved):

PR #1482, opened 2024-08-20
sim-se, arch: Fix syscall parametre sizes for 32-bit OSs

TiberiuBucur:
A bug was uncovered in that for various syscalls that used 64bit parametres, the ABI for 32bit operating systems was passing the wrong values to the syscalls, due to discrepancies between the target and guest OS. This commit fixes that by replacing 64-bit types, or types that are platform specific in size, with the exact correspondent for the guest OS, thus producing the correct signature for the respective syscalls. On top of this, the --param argument is added to the starter_se script, in order to support attachment of remote debuggers.


configs/example/arm/starter_se.py
```diff
@@ -203,6 +203,19 @@ def main():
         default="stdoutput",
         help="Destination for the Tarmac trace output. [Default: stdoutput]",
     )
+    parser.add_argument(
```
powerjg, line 206: I won't block this PR because of this change, but generally I am against adding more command line parameters. I also don't like the idea of being able to set any parameter via the command line. While I understand that it is useful for debugging, it makes reproducible science/experiments essentially impossible. 

I would prefer this not be exposed in any of the "starter" or "example" scripts.

src/arch/mips/linux/linux.hh
```diff
@@ -192,6 +203,8 @@ class MipsLinux : public Linux, public OpenFlagTable<MipsLinux>
        uint32_t mem_unit;  /* Memory unit size in bytes */
     };
 
+    typedef uint32_t size_t;
+    typedef int32_t off_t;
```
powerjg, line 207: Why this change? Has it been tested? If so, how was it tested?

src/arch/sparc/linux/linux.hh
```diff
@@ -242,6 +253,9 @@ class SparcLinux : public Linux, public OpenFlagTable<SparcLinux>
         ptc->setReg(SparcISA::SyscallPseudoReturnReg, (RegVal)0);
         ctc->setReg(SparcISA::SyscallPseudoReturnReg, 1);
     }
+
+    typedef uint64_t size_t;
+    typedef int64_t off_t;
```
powerjg, line 258: Ditto: How has this change been tested?

src/arch/sparc/linux/linux.hh
```diff
@@ -288,6 +302,9 @@ class Sparc32Linux : public SparcLinux
         uint32_t mem_unit;  /* Memory unit size in bytes */
     };
 
+    typedef uint32_t size_t;
+    typedef int32_t off_t;
```
powerjg, line 306: Ditto: testing.

powerjg (commented):
I am slightly concerned about `typedef`ing `size_t` and `off_t` unless it has been very carefully tested.

Have you run binaries in SE mode that use these system calls in all of the ISAs that you modified? If not, I don't feel very comfortable merging the change. 

Can you provide some evidence that your change won't break things for SPARC, POWER, and other ISAs?

powerjg:
Thanks for the contribution, by the way!

configs/example/arm/starter_se.py
TiberiuBucur: I understand your point. However, it's common practice for other configuration scripts (see fs_bigLITTLE.py) to have this option, and it would simplify our lives.

src/arch/mips/linux/linux.hh
TiberiuBucur: This set of changes is needed because the `lseekFunc` and `getcwdFunc` have been templatised with the OS type parameter, and take an `off_t` or `size_t` type parameter. This is common practice for all syscalls that require parameters of this type (`mmapFunc` is a good example). As a result, all the ABIs built with syscall tables that include these functions must pass as a template parameter an OS that defines `size_t` and `off_t` types (this is, as I said, alreadt common practice - see `gem5::X86Linux64` for an example). `lseek` and `getcwd` were quite ubiquitous, and it would be unfortunate to remove them from the table of supported syscalls for any of the other architectures, hence the need for the typedefs.

powerjg: I'm still concerned that it isn't tested. I agree that we don't want to remove support for `lseek` and `getcwd`, but I also don't want to have any regressions in support.

Is there any way you can test this?

configs/example/arm/starter_se.py
powerjg: As I said, I won't block this, because, as you said, it's pervasive. However, I do want you to know where I'm coming from.

The question to me is "who is this script for?" Given that it is called "starter_se.py" that would lead me to believe that it is for new gem5 users to "get started" using Arm. The name would imply that if you are using gem5 for research *you should not use the `starter_se.py` script*.

So, unless I'm misinterpreting the purpose of this script, the goal of this script would not be to make our (the developers') lives easier, but to make it easy to understand and get started with gem5. I think that exposing every possible parameter on the command line is not the right way to expose new users to gem5. Exposing parameters this way encourages people to use gem5 in an unreproducible and error-prone way.

src/arch/mips/linux/linux.hh
giactra: I think @powerjg concerns for testing are valid. However IMHO the MIPS/POWER/SPARC change  is relatively straightforward; it is simply replacing an hardcoded(uint64_t) or host dependent (unsigned long) argument with the appropriate ISA size. 

That sizes are properly chosen can be verified in godbolt, choosing different sparc/mips/power cross compiler
```

#include <sys/types.h>
#include <iostream>
#include <stdint.h>

struct CurrISA
{
    using expected_off_t = uint32_t;
    using expected_size_t = uint32_t;
};

template <class ISA>
void checker()
{
    static_assert(
        sizeof(off_t) == sizeof(typename ISA::expected_off_t));
    static_assert(
        sizeof(size_t) == sizeof(typename ISA::expected_size_t));
}

int main()
{
    checker<CurrISA>();
}
```

Let me know if this is enough for you. We can try to come up with lseek testing programs but as you can imagine cross-compiling these not so used ISAs (I am sure our gem5 testing of MIPS SE is close to zero) might not be the simplest thing to do :)

powerjg: Yeah, this is great. I agree that it's a pain to test. I wish I could convince people to get rid of some of these rarer ISAs...

The other thing that I would like to make sure is that all of the code compiles. I *think* if you compile `ALL` it should compile all of this code, but it would be good to double check.

TiberiuBucur: The code compiles when building the ALL target. I have tried using this test program on PowerPC, using the cross compiler provided in the Docker image in tests/test-progs/hello/src and the cross compiler available for Ubuntu 22.01 in the apt package manager (a.k.a. powerpc64le-linux-gnu-gcc):
```
#include <stdio.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/syscall.h>
#include <errno.h>

int main() {  
    int fd = open("example.txt", O_RDONLY | O_CREAT, 0644);  
    if (fd == -1) {  
        perror("open");  
        return 1;  
    }
    off_t result;
    syscall(SYS__llseek, fd, 0, 10, &result, SEEK_SET);
    if (result == -1) {
        perror("lseek");  
        close(fd);  
        return 1;  
    }  
    char buffer[10];  
    ssize_t bytesRead = read(fd, buffer, sizeof(buffer));  
    if (bytesRead == -1) {  
        perror("read");  
        close(fd);  
        return 1;  
    }  
    printf("Read %zd bytes: %.*s\n", bytesRead, (int)bytesRead, buffer);  
    close(fd);  
    return 0;
}
```
When run under gem5, both versions of the produced binary encounter unimplemented instructions which are unrelated to syscall emulation and their implementation is outside the scope of this bugfix.
If you don't feel comfortable committing the patch in this form, an alternative would be to keep around an old implementation of the affected syscalls and revert the type definition changes for these rarer ISAs. This would ensure the exact same state prior to the patches is preserved.

powerjg: I guess this is OK as the patch currently is. If something ends up breaking, we will at least have this conversation to come back to. I just feel uncomfortable committing untested code.

powerjg (approved):

PR #1489, opened 2024-08-21
mem-ruby, sim-se: Add support for Maybe_Stale blocks in functional reads

hexengraf:
Functional reads can be satisfied by one of the following, in order:
1. Main memory (when the data is not present in the cache hierarchy);
2. Valid data block in cache;
3. Valid data block in coherence message;
4. Valid data block marked as `Maybe_Stale`;

Number 4 is not handled by the current implementation. A `Maybe_Stale` block can be either truly stale or actually valid. When it is stale, the memory read will be satisfied by either number 2 or number 3. When it is valid, there will be no coherence message with valid data inside, and the `Maybe_Stale` block will transition to a valid state after receiving some kind of acknowledgement.

The main challenge to handle number 4 is how to know from which `Maybe_Stale` block the data should be read from. For instance, in a two level cache hierarchy, we might have a block marked as `Maybe_Stale` in both L1 and L2. In this case, we should prioritize the cache controller that is closest to the CPU. To define this priority, a new virtual function `functionalReadPriority` was added to the `AbstractController` class.

This PR also fixes `MESI_Three_Level` and `MESI_Two_Level` protocols so that they make use of the new support for `Maybe_Stale` blocks, and additionally fixes two more issues related to functional reads:

1) Fix incorrect access permissions in `MESI_Three_Level-L1cache`:
* `S_IL0` is `Read_Only`, it is waiting for L0 to acknowledge the invalidation request before moving to `SS`, also a `Read_Only` state.
* `E_IL0` is `Maybe_Stale`, its contents might be valid, since there is a transition `(E_IL0, L0_Ack, EE)` with no writeback data.
* `M_IL0` is `Maybe_Stale`, its contents might be valid, since there is a transition `(M_IL0, L0_Ack, MM)` with no writeback data.

2) Add missing message types carrying valid data in functional reads:
* `INV_DATA` is a writeback from L0 to L1.
* `DATA` is a response to `GET_S`, but there are scenarios where it might be the only place with valid data (e.g. during L2 replacement).

### DIFF
--- a/src/mem/ruby/protocol/MESI_Two_Level-L2cache.sm
+++ b/src/mem/ruby/protocol/MESI_Two_Level-L2cache.sm
@@ -238,6 +238,10 @@ machine(MachineType:L2Cache, "MESI Directory L2 Cache CMP")
     }
   }
 
+  int functionalReadPriority() {
+    return 20;
+  }
+
   int functionalWrite(Addr addr, Packet *pkt) {
     int num_functional_writes := 0;
 
--- a/src/mem/ruby/slicc_interface/AbstractController.hh
+++ b/src/mem/ruby/slicc_interface/AbstractController.hh
@@ -134,6 +134,14 @@ class AbstractController : public ClockedObject, public Consumer
     virtual bool functionalReadBuffers(PacketPtr&) = 0;
     virtual void functionalRead(const Addr &addr, PacketPtr)
     { panic("functionalRead(Addr,PacketPtr) not implemented"); }
+    /**
+     * Returns the priority used by functional reads when deciding from which
+     * controller to read a Maybe_Stale data block.
+     * Lower positive values have higher priority, negative values are ignored.
+     *
+     * @return the controller's priority
+     */
+    virtual int functionalReadPriority() { return -1; }
 
     //! Functional read that reads only blocks not present in the mask.
     //! Return number of bytes read.
--- a/src/mem/ruby/protocol/MESI_Three_Level-L1cache.sm
+++ b/src/mem/ruby/protocol/MESI_Three_Level-L1cache.sm
@@ -90,9 +90,9 @@ machine(MachineType:L1Cache, "MESI Directory L1 Cache CMP")
     // For all of the following states, invalidate
     // message has been sent to L0 cache. The response
     // from the L0 cache has not been seen yet.
-    S_IL0, AccessPermission:Busy, desc="Shared in L1, invalidation sent to L0, have not seen response yet";
-    E_IL0, AccessPermission:Busy, desc="Exclusive in L1, invalidation sent to L0, have not seen response yet";
-    M_IL0, AccessPermission:Busy, desc="Modified in L1, invalidation sent to L0, have not seen response yet";
+    S_IL0, AccessPermission:Read_Only, desc="Shared in L1, invalidation sent to L0, have not seen response yet";
+    E_IL0, AccessPermission:Maybe_Stale, desc="Exclusive in L1, invalidation sent to L0, have not seen response yet";
+    M_IL0, AccessPermission:Maybe_Stale, desc="Modified in L1, invalidation sent to L0, have not seen response yet";
     MM_IL0, AccessPermission:Read_Write, desc="Invalidation sent to L0, have not seen response yet";
     SM_IL0, AccessPermission:Busy, desc="Invalidation sent to L0, have not seen response yet";
   }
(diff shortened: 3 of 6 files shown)


hexengraf:
In #1045, it was decided to be cautious and not add `DATA` messages to functional reads, but `DATA` is required as mentioned above. During L2 replacement, the L2 cache controller will be in `S_I` state with `AccessPermission:Busy`. It can enter this state while a `DATA` message is still circulating, and no other L1 cache has this block valid. Another valid solution is to change `S_I` to `AccessPermission:Maybe_Stale` and let the new mechanism take care of functional reads in this specific scenario. The caveat is that there might be other scenarios where `DATA` messages are the only valid place to access.

Allowing functional reads to read from `DATA` messages seems to be the most solid approach, and it is already done in other protocols (e.g. `MOESI_CMP_directory`).

Even if `DATA` is actually stale data, it would be fine in ISAs that allow other-multi-copy atomic writes (is it the case for all of the ISAs supported in gem5? I'm not sure about x86).

hexengraf:
I'm not super happy with a virtual function to define a simple integer representing a priority for each type of cache controller, but it was the simplest way I found to interface with SLICC. If there are any alternative suggestions that are not overly complicated to implement, I can try updating this commit to use them.

Also, considering other-multi-copy atomic writes, I'm pretty sure O3CPU allows read forwarding from the store queue. Is the current implementation of functional reads taking this into account?

vspetrov:
@hexengraf what is the status of this fix? Is it under review?

we've seen similar issue in FS. Functional read comes at the moment when the line is invalidated from L1 and the data should searched in INV_DATA packets.

hexengraf:
Hi @vspetrov,
I'm not a member of the project, so I can only see the information that is publicly available here. So far, it doesn't look like anyone has started reviewing this PR.

You may consider keeping a local change while we wait for a review. If you extract only the part relevant to `INV_DATA` packets, it should be a very safe change.

src/mem/ruby/slicc_interface/AbstractController.hh
```diff
@@ -133,6 +133,7 @@ class AbstractController : public ClockedObject, public Consumer
     virtual bool functionalReadBuffers(PacketPtr&) = 0;
     virtual void functionalRead(const Addr &addr, PacketPtr)
     { panic("functionalRead(Addr,PacketPtr) not implemented"); }
+    virtual int functionalReadPriority() { return -1; }
```
powerjg, line 144: I know it's not standard in this file, but could you add a detailed comment on the purpose of this virtual function?

I.e., in 5 years when we come back to this, I'd like to be able to remember why we need controllers to specify their functional read priority.

src/mem/ruby/protocol/MESI_Two_Level-L2cache.sm
```diff
@@ -238,6 +238,10 @@ machine(MachineType:L2Cache, "MESI Directory L2 Cache CMP")
     }
   }
 
+  int functionalReadPriority() {
+    return 2;
```
powerjg: Since this machine is used as both an L3 cache (in MESI_Three_Level) and L2 cache (in MESI_Two_Level), maybe it makes sense to multiply this value by 10 (e.g., 20) so that if other levels get put in between it will work out. Same for the other values.

I'm thinking specifically of when using MESI_Three_Level and you have the L1/L2 cache. Right now, it would be difficult specify their priority since the L2 is priority 2.

powerjg (commented):
Generally, this change looks good to me. Apologies for taking so long to review it. Could you please take a look at the small comments below?

ivanaamit:
Hi @hexengraf, could you please address Jason's comments? Thank you.

src/mem/ruby/slicc_interface/AbstractController.hh
hexengraf: Let me know if it is ok now or more explanation is needed.
Also I wasn't sure about which format to follow, this file has comments in multiple formats.

hexengraf:
Hi @powerjg, I did the requested changes. Let me know if anything else is needed. Thanks!

hexengraf:
Any chances of getting this PR merged before V24.1 staging? I would greatly appreciate it.

powerjg (approved):
Sorry it took me so long to get back to this. I appreciate the updates!

powerjg:
@BobbyRBruce can you merge this into the staging branch?

PR #1490, opened 2024-08-21
stdlib, configs: Add RiscvDemoBoard

erin-le:
This PR adds a RiscvDemoBoard that can be used with both SE and FS mode.This was tested using the workloads riscv-matrix-multiply-run for SE and riscv-ubuntu-20.04-boot for FS. Two example config scripts have also been added.


configs/example/gem5_library/riscv-demo-board-se.py
```diff
@@ -0,0 +1,66 @@
...
+parser.add_argument("--version")
+
+args = parser.parse_args()
+
+# Here we set up the board. The prebuilt RiscvDemoBoard allows for FS
+# or SE simulation.
+board = RiscvDemoBoard(is_fs=False)
+
+# Here we set a workload.
+board.set_workload(obtain_resource(args.workload, args.version))
```
Harshil2107: May be making the version an optional argument would be better, as if the use wants the latest version they can just pass the id.

configs/example/gem5_library/riscv-demo-board-se.py
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2021 The Regents of the University of California
```
BobbyRBruce: Update the copyright here to 2024

configs/example/gem5_library/riscv-demo-board-ubuntu-run-24.04.py
```diff
@@ -0,0 +1,102 @@
+# Copyright (c) 2021 The Regents of the University of California
```
BobbyRBruce: Update the copyright year

configs/example/gem5_library/riscv-demo-board-ubuntu-run.py
```diff
@@ -0,0 +1,72 @@
+# Copyright (c) 2021 The Regents of the University of California
```
BobbyRBruce: Update the copyright year

configs/example/gem5_library/riscv-demo-board-ubuntu-run-24.04.py
```diff
@@ -0,0 +1,102 @@
...
+board = RiscvDemoBoard(is_fs=True)
+
+
+def handle_workend():
+    print("Dump stats at the end of the ROI!")
+
+    m5.stats.dump()
+    print("Reset stats at the end of the ROI!")
+    m5.stats.reset()
+    yield False
```
BobbyRBruce: Isn't the `workend` exit event event the last exit event where we exit to the Python script? If true, shouldn't this be `yield True`?

configs/example/gem5_library/riscv-demo-board-ubuntu-run-24.04.py
```diff
@@ -0,0 +1,102 @@
...
+
+board = RiscvDemoBoard(is_fs=True)
+
+
+def handle_workend():
+    print("Dump stats at the end of the ROI!")
+
+    m5.stats.dump()
+    print("Reset stats at the end of the ROI!")
+    m5.stats.reset()
```
BobbyRBruce: Why do we reset stats at the end or the ROI? Isn't this the end of the simulation?

configs/example/gem5_library/riscv-demo-board-ubuntu-run-24.04.py
```diff
@@ -0,0 +1,102 @@
...
+    yield False
+
+
+def handle_workbegin():
+    print("Done booting Linux")
+    print("Dumping stats at beginning of ROI!")
+    m5.stats.dump()
```
BobbyRBruce: Why do we want to dump stats here? If we're at the start of the workbegin then isn't  all the stats before just useless OS stuff?

configs/example/gem5_library/riscv-demo-board-ubuntu-run-24.04.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def exit_event_handler():
+    print("first exit event: Kernel booted")
+    yield False
+    print("second exit event: In after boot")
+    yield False
+    print("third exit event: After run script")
+    print("Dump stats at end of workload!")
+    yield True
```
BobbyRBruce: So there's a final "Exit" type exit event after the workend has been called? I'm very confused.

src/python/gem5/prebuilt/demo/riscv_demo_board.py
```diff
@@ -0,0 +1,204 @@
+# Copyright (c) 2021 The Regents of the University of California
```
BobbyRBruce: Copyright year needs fixed here.

configs/example/gem5_library/riscv-demo-board-se.py
```diff
@@ -0,0 +1,69 @@
...
+
+# Here we set up the board. The prebuilt RiscvDemoBoard allows for FS
+# or SE simulation.
+board = RiscvDemoBoard(is_fs=False)
+
+# Here we set a workload. If nothing is passed for resource_version, the
+# latest resource version compatible with the gem5 version will be obtained.
+board.set_workload(
+    obtain_resource(resource_id=args.workload, resource_version=args.version)
+)
```
BobbyRBruce: why do you need to declare `is_fs` in the `RiscvDemoBoard` constructor? Can't this just be infered by what workload is passed? I.e., if a FS is one is called then setup teh FS board, if not, setup the SE workload.

If you did that you could just merge these three demos board script together into one

BobbyRBruce (changes requested):

BobbyRBruce (dismissed):

BobbyRBruce:
I can't push to this PR but here's the patch that fixes this:

```patch
index 555f723df1..f5c536599e 100644
--- a/src/python/gem5/components/boards/riscv_board.py
+++ b/src/python/gem5/components/boards/riscv_board.py
@@ -26,7 +26,10 @@
 
 import os
-from typing import List
+from typing import (
+    List,
+    Optional,
+)
 
 import m5
 from m5.objects import (
@@ -498,16 +501,26 @@ class RiscvBoard(AbstractSystemBoard, KernelDiskWorkload):
         return "/dev/vda"
 
-    def _pre_instantiate(self):
-        if len(self._bootloader) > 0:
-            self.workload.bootloader_addr = 0x0
-            self.workload.bootloader_filename = self._bootloader[0]
-            self.workload.kernel_addr = 0x80200000
-            self.workload.entry_point = 0x80000000  # Bootloader starting point
-        else:
-            self.workload.kernel_addr = 0x0
-            self.workload.entry_point = 0x80000000
-        self._connect_things()
+    def _pre_instantiate(self, full_system: Optional[bool] = None):
+        # This is a bit of a hack necessary to get the RiscDemoBoard working
+        # At the time of writing the RiscvBoard does not support SE mode so
+        # this branch looks pointless. However, the RiscvDemoBoard does and
+        # needs this logic in place.
+        #
+        # This should be refactored in the future as part of a chance to have
+        # all boards support both FS and SE modes.
+        if self._is_fs:
+            if len(self._bootloader) > 0:
+                self.workload.bootloader_addr = 0x0
+                self.workload.bootloader_filename = self._bootloader[0]
+                self.workload.kernel_addr = 0x80200000
+                self.workload.entry_point = (
+                    0x80000000  # Bootloader starting point
+                )
+            else:
+                self.workload.kernel_addr = 0x0
+                self.workload.entry_point = 0x80000000
+        super()._pre_instantiate(full_system=full_system)
 
     def _add_disk_to_board(self, disk_image: AbstractResource):
diff --git a/src/python/gem5/prebuilt/demo/riscv_demo_board.py b/src/python/gem5/prebuilt/demo/riscv_demo_board.py
index e4ef627895..cd73e8efe6 100644
--- a/src/python/gem5/prebuilt/demo/riscv_demo_board.py
+++ b/src/python/gem5/prebuilt/demo/riscv_demo_board.py
@@ -107,24 +107,6 @@ class RiscvDemoBoard(RiscvBoard, SEBinaryWorkload):
         )
 
-    # Taken from Riscv Matched board. Below are functions that are needed to get
-    # SE mode to work.
-    @overrides(RiscvBoard)
-    def _pre_instantiate(self):
-        if self._is_fs:
-            if len(self._bootloader) > 0:
-                self.workload.bootloader_addr = 0x0
-                self.workload.bootloader_filename = self._bootloader[0]
-                self.workload.kernel_addr = 0x80200000
-                self.workload.entry_point = (
-                    0x80000000  # Bootloader starting point
-                )
-            else:
-                self.workload.kernel_addr = 0x0
-                self.workload.entry_point = 0x80000000
-
-        self._connect_things()
-
     @overrides(RiscvBoard)
     def _setup_board(self) -> None:
         if self._is_fs:
```

BobbyRBruce:
The commit message I was going to have going with this patch (if it helps):

```commit
 stdlib,arch-riscv: Fix RISCV DemoBoard
    
    1. RiscvBoard did not have the correct signature for _pre_instantiate
       (missing `full_system` param).
    2. RiscvDemoBoard required some `is_fs` logic moved down to the
       RiscvBoard.
```

src/python/gem5/components/boards/riscv_board.py
```diff
@@ -502,14 +502,24 @@ def get_disk_device(self):
...
-        else:
-            self.workload.kernel_addr = 0x0
-            self.workload.entry_point = 0x80000000
+        # This is a bit of a hack necessary to get the RiscDemoBoard working
+        # At the time of writing the RiscvBoard does not support SE mode so
+        # this branch looks pointless. However, the RiscvDemoBoard does and
+        # needs this logic in place.
+        #
+        # This should be refactored in the future as part of a chance to have
+        # all boards support both FS and SE modes.
```
powerjg, line 511: This is OK, but couldn't you override this in `RiscvDemoBoard` and have something like:

```python
def _pre_instantiate(self...):
     if self._is_fs:
        super()._pre_instantiate()
    else:
        self.workload.kernel_addr = 0x0
...
        AbstractBoard._pre_instantiate(self, full_system)
```
```

erin-le:
I tested this change using the workloads riscv-matrix-multiply-run and riscv-ubuntu-24.04-boot. Both simulations run, so it looks like this fixed the problems that the RiscvDemoBoard was having with the pre-instantiate change.

src/python/gem5/components/boards/riscv_board.py
BobbyRBruce: The problem i have with that is you'd need to copy and paste the `AbstractBoard`'s `_pre_instantiate` function into the `else` block which is a bit annoying. In short, for FS mode we want to run the `RiscvBoard`'s `_pre_instantiate` and for SE mide we just want to run the `AbstractBoard`'s. I tried `super().super()._pre_instantiate()` but Python didn't like it. Perhaps there's a way to do it but in the end i went with the solution I had working.

powerjg: I see. I *could* do `AbstractBoard._pre_instantiate(self, ...)` in the `else`, but that isn't ideal either. Feel free to ignore this comment.

powerjg (approved):

BobbyRBruce (approved):

PR #1491, opened 2024-08-21
misc: update fs examples to use ubuntu 24.04 boot workloads

Harshil2107:
Change-Id: I7e16f69eff3a7ff0ab16c18e6d35e846d07ac829


configs/example/gem5_library/arm-ubuntu-run-with-kvm.py
```diff
@@ -105,39 +105,36 @@
...
+# Ubuntu 24.04.
+workload = obtain_resource("arm-ubuntu-24.04-boot-with-systemd")
+board.set_workload(workload)
+
+
+def exit_event_handler():
+    print("First exit: kernel booted")
+    yield False
+    print("Second exit: Started `after_boot.sh` script")
+    # The after_boot.sh script is executed after the kernel and systemd have\
```
powerjg:
```suggestion
    # The after_boot.sh script is executed after the kernel and systemd have
```

configs/example/gem5_library/arm-ubuntu-run-with-kvm.py
```diff
@@ -105,39 +105,36 @@
...
-# We define the system with the aforementioned system defined.
+# Here we set a full system workload. The "arm-ubuntu-24.04-boot-with-systemd" boots
+# Ubuntu 24.04.
+workload = obtain_resource("arm-ubuntu-24.04-boot-with-systemd")
+board.set_workload(workload)
+
+
+def exit_event_handler():
+    print("First exit: kernel booted")
+    yield False
```
powerjg:
```suggestion
    yield False # gem5 is now executing systemd startup
```

configs/example/gem5_library/arm-ubuntu-run-with-kvm.py
```diff
@@ -105,39 +105,36 @@
...
+    print("Second exit: Started `after_boot.sh` script")
+    # The after_boot.sh script is executed after the kernel and systemd have\
+    # booted.
+    # Here we switch the CPU type to Timing.
+    print("Switching to Timing CPU")
+    processor.switch()
+    yield False
```
powerjg:
```suggestion
    yield False # gem5 is now executing the `after_boot.sh` script
```

powerjg (commented):
Is there a riscv script to update?

configs/example/gem5_library/riscvmatched-fs.py
erin-le, line 1: I got an error when running this. 

```txt
src/cpu/minor/lsq.cc:1604: panic: Do not expect cross-cache-line atomic memory request
Memory Usage: 17910044 KBytes
Program aborted at tick 2908701793704
--- BEGIN LIBC BACKTRACE ---


```

Harshil2107: This seem to me like an error with the cpu code and not the resource, I will run the original and see if I get the same error

Harshil2107: This error occurs when we run riscvmatched with ubuntu 24.04 image. I ran the ubuntu 20.04 image and it works fine. I also ran the Ubuntu 24.04 image with the `RiscvBoard` and that also works.

Harshil2107: I have created an issue for it here: https://github.com/gem5/gem5/issues/1554

Harshil2107:
I reverted the riscvmatched-fs.py due to a bug that crashes the simulation when using the new 24.04 images. Link to issue #1554

powerjg (commented):
Let's discuss pausing this change for a little while. @BobbyRBruce is working on updating the exit events to be more flexible, which should eliminate the need for the verbose exit event handler. If Bobby is planning to get his change merged soon (for some definition of soon), it make sense to hold off on this change.

BobbyRBruce:
@Harshil2107 Just double checking here: this is needed now? I assume so, since the exit event changes aren't getting in?

PR #1497, opened 2024-08-22
stdlib: Extend `AbstractBoard` pre_instantiation functionality

BobbyRBruce:
* Deprecates the setting of FS/SE mode via the `Simulator` module.
* Moved the creation of the `Root` object from the `Simulator` to the board.
* Moved the setting of `sim_quantum` from the `Simulator` to the processor.
* Allows for easier development of boards which support both SE and FS mode simulation by moving board setup function calls to occur after the set_workload function is call which sets a boards stats `is_fs` status.

### DIFF
--- a/src/python/gem5/components/boards/abstract_board.py
+++ b/src/python/gem5/components/boards/abstract_board.py
@@ -406,11 +406,11 @@ class AbstractBoard:
                             occurs whether a simulation is to be run in FS or
                             SE mode will be determined by the board set."""
 
-        # Connect the memory, processor, and cache hierarchy.
+        # 1. Connect the memory, processor, and cache hierarchy.
         self._connect_things()
 
-        # Return the Root object.
-        return Root(
+        # 2. Create the root object
+        root = Root(
             full_system=(
                 full_system
                 if full_system is not None
@@ -418,6 +418,13 @@ class AbstractBoard:
             ),
             board=self,
         )
+
+        # 3. Call any of the components' `_pre_instantiate` functions.
+        # Right now, only the processor requires this.
+        self.get_processor()._pre_instantiate(root)
+
+        # 4. Return the root object.
+        return root
 
     def _connect_things_check(self):
         """
--- a/src/python/gem5/components/boards/abstract_board.py
+++ b/src/python/gem5/components/boards/abstract_board.py
@@ -41,6 +41,7 @@ from m5.objects import (
     ClockDomain,
     IOXBar,
     Port,
+    Root,
     SrcClockDomain,
     System,
     VoltageDomain,
@@ -391,12 +392,32 @@ class AbstractBoard:
             self.get_cache_hierarchy()._post_instantiate()
         self.get_memory()._post_instantiate()
 
-    def _pre_instantiate(self):
+    def _pre_instantiate(self, full_system: Optional[bool] = None) -> Root:
         """To be called immediately before ``m5.instantiate``. This is where
-        ``_connect_things`` is executed by default."""
+        ``_connect_things`` is executed by default and the root object is Root
+        object is created and returned.
+
+        :param full_system: Used to pass the full system flag to the board from
+                            the Simulator module. **Note**: This was
+                            implemented solely to maintain backawards
+                            compatibility with while the Simululator module's
+                            `full_system` flag is in state of deprecation. This
+                            parameter will be removed when it is. When this
+                            occurs whether a simulation is to be run in FS or
+                            SE mode will be determined by the board set."""
 
         # Connect the memory, processor, and cache hierarchy.
         self._connect_things()
+
+        # Return the Root object.
+        return Root(
+            full_system=(
+                full_system
+                if full_system is not None
+                else self.is_fullsystem()
+            ),
+            board=self,
+        )
 
     def _connect_things_check(self):
         """
--- a/src/python/gem5/components/boards/abstract_board.py
+++ b/src/python/gem5/components/boards/abstract_board.py
@@ -118,12 +118,6 @@ class AbstractBoard:
         # Simulator module.
         self._checkpoint = None
 
-        # Setup the board and memory system's memory ranges.
-        self._setup_memory_ranges()
-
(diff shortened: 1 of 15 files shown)


BobbyRBruce:
((Note: This is still a draft but I wanted to run the tests so have made it "ready for review" purely for this reason)).

BobbyRBruce:
@mahyarsamani : Is this ok with you?

src/python/gem5/components/boards/arm_board.py
```diff
@@ -327,8 +328,8 @@ def connect_system_port(self, port: Port) -> None:
         self.system_port = port
 
     @overrides(AbstractBoard)
-    def _pre_instantiate(self):
-        super()._pre_instantiate()
+    def _pre_instantiate(self, full_system: Optional[bool]) -> None:
```
powerjg: I don't think `full_system` is "optional" here since it doesn't have a default.

src/python/gem5/components/boards/riscv_board.py
```diff
@@ -498,7 +501,7 @@ def get_disk_device(self):
         return "/dev/vda"
 
     @overrides(AbstractSystemBoard)
-    def _pre_instantiate(self):
+    def _pre_instantiate(self, full_system: Optional[bool]):
```
powerjg: I don't think `full_system` is "optional" here since it doesn't have a default.

src/python/gem5/prebuilt/riscvmatched/riscvmatched_board.py
```diff
@@ -313,7 +313,7 @@ def _setup_memory_ranges(self):
             memory.set_memory_range(self.mem_ranges)
 
     @overrides(AbstractSystemBoard)
-    def _pre_instantiate(self):
+    def _pre_instantiate(self, full_system: Optional[bool]) -> None:
```
powerjg: I don't think `full_system` is "optional" here since it doesn't have a default.

src/python/gem5/components/boards/abstract_board.py
```diff
@@ -384,13 +385,40 @@ def _post_instantiate(self):
...
+                full_system
+                if full_system is not None
+                else self.is_fullsystem()
+            ),
+            board=self,
+        )
+
+        # 3. Call any of the components' `_pre_instantiate` functions.
+        # Right now, only the processor requires this.
+        self.get_processor()._pre_instantiate(root)
```
powerjg, line 431: Let's go ahead and add this for the `cache_hierarchy` and `memory_system`. I need the cache hierarchy to get the power model example in the bootcamp to work.

powerjg (commented):
A couple of small things below. I pushed a commit to fix these things.

powerjg:
Apologies for messing up anything with the force push. I was trying to make your life easier, but pre-commit is broken on my system after upgrading to 24.04.

src/python/gem5/components/boards/abstract_board.py
BobbyRBruce: Unless I'm very confused I believe I already have. Perhaps you were looking at an old commit? The next two lines call the cache hierarchy and memory system pre_instantiate functions

BobbyRBruce: Disregard, that... I was confused. I think I deleted some god reverting your changes. i'll add these

BobbyRBruce (commented):
@powerjg : I take it I can remove the last commit in this chain?  https://github.com/gem5/gem5/pull/1497/commits/14e12907346e4e9132c7bf969c0634f8115e5c2c. It's only purpose was to demo this stuff to Mayhar. I'm not sure it's actually useful

src/python/gem5/components/boards/abstract_board.py
BobbyRBruce: Done

powerjg:
> @powerjg : I take it I can remove the last commit in this chain? https://github.com/gem5/gem5/commit/14e12907346e4e9132c7bf969c0634f8115e5c2c. It's only purpose was to demo this stuff to Mayhar. I'm not sure it's actually useful

It's up to you. We can remove it and let Mahyar push a change if he thinks it's important.

src/python/gem5/components/memory/abstract_memory_system.py
```diff
@@ -50,6 +51,21 @@ class AbstractMemorySystem(SubSystem):
...
+
+        Subclasses should override this method to set up any connections.
+
+        At present there is no general task that must be specified here and is
+        default or applicable to all memory systems.
+        """
+        pass
+
+    def _pre_instantiate(self, root: Root) -> None:
+        """Called to set up anything needed before ``m5.instantiate``."""
```
powerjg: I think this should be removed.

powerjg (commented):
One little thing. Sorry if I made things harder for you by adding that commit.

src/python/gem5/components/memory/abstract_memory_system.py
BobbyRBruce: Yeah, that was a mistake. Removed.

powerjg:
I suggest dropping the branch predictor from this pr and creating another with an example.

powerjg (changes requested):
The branch predictor part of this (most likely) doesn't work. Let's drop that commit and then this can be merged. I would like this merged ASAP because we have other changes that depend on this.

BobbyRBruce:
> The branch predictor part of this (most likely) doesn't work. Let's drop that commit and then this can be merged. I would like this merged ASAP because we have other changes that depend on this.

I've removed it.

I've added one more patch to this chain. It moves board setup functions to occur after it's decided whether the board is in FS or SE (which is decided based on what set_workload). This allows for easier implementation of boards that support both. All setup functions can call `is_fullsystem` and modify the simulated system accordingly. Very little now happens before the workloads are specified.

powerjg (approved):

Harshil2107 (approved):
I tested this change on top of the arm demo board and the boot fs simulation ran successfully. Overall the other changes also look good to me.

PR #1498, opened 2024-08-22
configs,tests: Deprecate "configs/common"

BobbyRBruce:
Due to heavy dependencies between the tests declared in "tests/gem5/fs", and the "configs/common" directory, these tests have been removed.


giactra:
Hey @BobbyRBruce these are ***the most important Arm tests we have***.... They check Linux boot under all the configurations we claim to support. If we remove those tests we might as well delete the Arm ISA from gem5 ;)

If you want to deprecate common (understandable), the solution is to rewrite those tests without common.

giactra:
Also the configs under configs/common/cores/arm you are deprecating are actually used by existing scripts we rely upon, in [configs/example/arm](https://github.com/gem5/gem5/tree/stable/configs/example/arm) like [starter_fs.py](https://github.com/gem5/gem5/tree/stable/configs/example/arm/starter_fs.py) and [fs_bigLITTLE.py](https://github.com/gem5/gem5/tree/stable/configs/example/arm/fs_bigLITTLE.py)

powerjg:
Let's put this on the list of things to discuss at the next developers' meeting. @ivanaamit can you update/create that discussion post?

powerjg:
I agree that it is not an option to remove the Arm FS tests and they must be updated to work without configs/common. I am willing to invest some resources to updating these tests, but we will likely need some help from @giactra to make sure the coverage is the same.

For `starter_se.py` and other scripts in `configs/examples/arm`, I am *highly confident* that they do not need to have every single option in `Options.py` available. (I am 100% confident because I know that there are options in that file *that don't correspond to anything in gem5*.) So, the question that we should discuss in the developers meeting is "what features are *required* for these scripts?" and "how do we move users from these scripts to the standard library?" Again, I'm willing to invest some resources, but it is important to first get buy-in from the users and other developers.

giactra:
> I agree that it is not an option to remove the Arm FS tests and they must be updated to work without configs/common. I am willing to invest some resources to updating these tests, but we will likely need some help from @giactra to make sure the coverage is the same.
> 

I'll be happy to provide some help if needed

> For `starter_se.py` and other scripts in `configs/examples/arm`, I am _highly confident_ that they do not need to have every single option in `Options.py` available. (I am 100% confident because I know that there are options in that file _that don't correspond to anything in gem5_.)

100% agree

> So, the question that we should discuss in the developers meeting is "what features are _required_ for these scripts?" and "how do we move users from these scripts to the standard library?"

stdlib has its benefits but so does configs/example/arm. stdlib lets you run a image out of the box so it is quite convenient for someone interested on a functional run, but I find configs/example/arm a bit more flexible wrt system confguration. So while we should deprecate common, I don't think we should discourage people from using  configs/example/arm.

> Again, I'm willing to invest some resources, but it is important to first get buy-in from the users and other developers.

IMHO deprecating common has low priority. It was important to deprecate fs.py and se.py as those were the top-level scripts people were really using. If you advertise stdlib, people won't say: "ok I use stdlib but I will use common in the backend". In other words people don't choose common, it is a by-product of the top-level config.
Not saying this will never happen, but as this is still a critical part of gem5, I wouldn't want to rush things with the outcome of losing coverage / introducing bugs.

Another thing to mention about common is that it is not all bad
For example, while being an old CPU model, there is really nothing wrong in https://github.com/gem5/gem5/blob/fa2e7b374c7d0f4aed98d1a75c67626811034436/configs/deprecated/common/cores/arm/O3_ARM_v7a.py

So we should go file by file in common a discuss what is part of the old fs.py swiss army knife methodology, and what is a maintenable file that got just unlucky when people decided to create it under the common directory

PR #1499, opened 2024-08-22
stdlib: Issue warn if func is a gen for exit_event

### LINKED ISSUE #1492: Add warning in exit events when generator is passed instead of calling the generator.
When defining an exit event in gem5, it's important to call the generator function instead of passing the generator itself to `ExitEvent.EXIT`. If a generator is mistakenly passed instead of being called, gem5 should issue a warning.

### Example

Consider the following generator function `exit_event_handler`:

```python
def exit_event_handler():
    print("First exit: kernel booted")
    yield False
    print("Second exit: Started `after_boot.sh` script")
    # The after_boot.sh script is executed after the kernel and systemd have booted.
    # Here we switch the CPU type to Timing.
    print("Switching to Timing CPU")
    processor.switch()
    yield False
    print("Third exit: Finished `after_boot.sh` script")
    # The after_boot.sh script will run a script if it is passed via m5 readfile.
    # This is the last exit event before the simulation exits.
    yield True
```

To use this generator for handling exit events, you should call the generator when adding it to the `Simulator`'s `on_exit_event` argument:

```python
on_exit_event={
    # Override the default behavior for the first m5 exit event.
    ExitEvent.EXIT: exit_event_handler()
}
```

However, if you mistakenly pass the generator without calling it:

```python
on_exit_event={
    # This will cause an issue as the generator is not called.
    ExitEvent.EXIT: exit_event_handler
}
```

gem5 should throw a warning to alert you that the generator function needs to be called, not just referenced.

BobbyRBruce:
Addresses Issue #1492

### DIFF
--- a/src/python/gem5/simulate/simulator.py
+++ b/src/python/gem5/simulate/simulator.py
@@ -357,6 +357,20 @@ class Simulator:
                     # In instances where the user passes a lone function, the
                     # function is called on every exit event of that type. Here
                     # we convert the function into an infinite generator.
+
+                    # We check if the function is a generator. If it is we
+                    # throw a warning as this is likely a mistake.
+                    import inspect
+
+                    if inspect.isgeneratorfunction(value):
+                        warn(
+                            f"Function passed for '{key.value}' exit event "
+                            "is not a generator but a function that returns "
+                            "a generator. Did you mean to do this? (e.g., "
+                            "did you mean `ExitEvent.EVENT : gen()` instead "
+                            "of `ExitEvent.EVENT : gen`)"
+                        )
+
                     def function_generator(func: Callable):
                         while True:
                             yield func()


ivanaamit (dismissed):

Harshil2107 (dismissed):

src/python/gem5/simulate/simulator.py
```diff
@@ -353,10 +353,22 @@ def print_hello() -> bool:
...
                     self._on_exit_event[key] = (func() for func in value)
-                elif isinstance(value, Callable):
+                elif isinstance(value, function):
                     # In instances where the user passes a lone function, the
                     # function is called on every exit event of that type. Here
                     # we convert the function into an infinite generator.
+
+                    # We check if the function is a generator. If it is we
+                    # throw a warning as this is likely a mistake.
+                    if isinstance(value(), Generator):
```
powerjg: This will break when you pass in a function since it calls the function.

You should be able to use https://docs.python.org/3/library/inspect.html#inspect.isgeneratorfunction to check if the thing is a pointer to a function that is a generator.

So, the code should be something like

```python
import inspect
if inspect.isgeneratorfunction(value):
```

powerjg (changes requested):

src/python/gem5/simulate/simulator.py
powerjg: Impressively, our tests caught this! Good job writing tests!!!

https://github.com/gem5/gem5/actions/runs/10508905940/job/29134080005?pr=1499#step:7:89

BobbyRBruce: `Test: simulator-exit-event-handler-with-lone-function-ALL-x86_64-opt Failed`

Wow. Nice and precise too. I'm quite proud of that.

BobbyRBruce: > This will break when you pass in a function since it calls the function.
> 
> You should be able to use https://docs.python.org/3/library/inspect.html#inspect.isgeneratorfunction to check if the thing is a pointer to a function that is a generator.
> 
> So, the code should be something like
> 
> ```python
> import inspect
> if inspect.isgeneratorfunction(value):
> ```

This worked perfectly. i've incorporated this into the PR.

powerjg (approved):

PR #1501, opened 2024-08-22
ext,tests,misc: Suppress incorrect GCC 12 error in Pybind

BobbyRBruce:
There is a compiler error with GCC 12 discussed here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=115824

This Pybind code triggers the bug and was causing our compiler tests to fail.

To fix gem5 compilation for gcc 12 these warnings/errors have been suppressed for this code.


Harshil2107 (dismissed):
lgtm

powerjg (commented):
Instead, should we try updating pybind?

ivanaamit (approved):

PR #1509, opened 2024-08-23
Use board get_mem_ports consistently

powerjg:
Previously, whether the board object or the memory_system returned the memory ports was not consistent in the cache_hierarchies

This commit makes it consistently use the board. Note: the board is a better place so it can customize the ports (e.g., add I/O components or other things

Change-Id: Ibbd5291ae8e0fee10aefdbd8890a653b8bd737df

### DIFF
--- a/src/python/gem5/components/boards/arm_board.py
+++ b/src/python/gem5/components/boards/arm_board.py
@@ -274,11 +274,15 @@ class ArmBoard(ArmSystem, AbstractBoard, KernelDiskWorkload):
 
     @overrides(AbstractBoard)
     def get_mem_ports(self) -> Sequence[Tuple[AddrRange, Port]]:
-        all_ports = [
-            (self.realview.bootmem.range, self.realview.bootmem.port),
-        ] + self.get_memory().get_mem_ports()
+        # Note: Ruby needs to create a directory for the realview bootmem
+        if self.get_cache_hierarchy().is_ruby():
+            all_ports = [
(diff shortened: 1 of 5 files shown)


BobbyRBruce:
`AbstractMemorySystem` and `AbstractBoard` both have `get_mem_ports` functions, with the latter's function calling the former's. I think it'd be helpful to have different function names here to lessen the likelihood of this happening again . Maybe just have the memory  system's function to just `get_ports`?

(Note: if you do this, please deprecate the old function. It's definitely an API.)

powerjg:
Are there any actions that I need to take? This commit makes it consistent between all of the different cache hierarchies which function they are using. It also deletes a function that was never used/called.

BobbyRBruce:
> Are there any actions that I need to take? This commit makes it consistent between all of the different cache hierarchies which function they are using. It also deletes a function that was never used/called.

It's not a blocking issue for me but if you wanted to improve things it was just a suggestion that we should maybe rename some either the Memory System 'get_mem_ports' funciton the board's. If you don't feel it's worth your time I'm ok with this getting in as is. It's still an improvement as is.

Somehow this caused tests to fail... that's odd...

powerjg:
Annoyingly, there's another ruby-specific thing :/. Anyway, the tests should pass now.

I looked into your suggestion about renaming `MemorySystem.get_mem_ports` to `get_ports`. Given that we would have to deprecate the old API, I feel that it's a non-trivial change and should be done in another PR.

BobbyRBruce (approved):

PR #1510, opened 2024-08-23
arch-arm: when programming an invalid PMU ID detach the counter

arichardson:
I'm not entirely sure what the mandated behaviour is according the the ARM ARM, but I was very confused by the counters continuing to increment with the old event even when programmed to an event ID that is not currently supported by GEM5. Disconnecting the counter if the event is not supported is less surprising behaviour IMO.

Change-Id: I927d9339c138dafa1484db1515c2aa09b0a9a0a9

### DIFF
--- a/src/arch/arm/pmu.cc
+++ b/src/arch/arm/pmu.cc
@@ -601,6 +601,7 @@ PMU::updateCounter(CounterState &ctr)
         if (sourceEvent == eventMap.end()) {
             warn("Can't enable PMU counter of type '0x%x': "
                  "No such event type.\n", ctr.eventId);
+            ctr.detach();
         } else {
             ctr.attach(sourceEvent->second);
         }


giactra (approved):
The Arm arm says:

```
Otherwise, if PMEVTYPER<n>_EL0.evtCount is programmed to an event that is reserved or not supported by the PE, the behavior depends on the value written:

    For the range 0x0000 to 0x003F, no events are counted and the value returned by a direct or external read of the PMEVTYPER<n>_EL0.evtCount field is the value written to the field.
    If FEAT_PMUv3p1 is implemented, for the range 0x4000 to 0x403F, no events are counted and the value returned by a direct or external read of the PMEVTYPER<n>_EL0.evtCount field is the value written to the field.
    For other values, it is UNPREDICTABLE what event, if any, is counted and the value returned by a direct or external read of the PMEVTYPER<n>_EL0.evtCount field is UNKNOWN
```

So I believe it is safe to disable the counter to stop event counting

PR #1514, opened 2024-08-24
base: Allow DPRINTF debugging of AssociativeCache

giactra:
The AssociativeCache is used by different caching agents.
This PR will allow to pass the appropriate flag to the cache so that we can meaningfully debug
its internals. For instance, when used to model a prefetcher table, will will pass the
HWPrefetch flag; when used to model a TLB, we will pass the TLB flag.


odanrc (approved):

powerjg (approved):
Clever!

PR #1516, opened 2024-08-26
cpu-o3: Panic if no FU exists for an instruction needing to issue

aperais:
At present, if an instruction requires a functional unit that is not present in the O3CPU config, O3CPU treats it as a 1-cycle operation that does not consume an FU. This seems like a silent failure : if I forgot to add a FU for a new operation type I added, then I don't want it to silently work "for free".

The problem is that the code treats the FU allocator returning `NoCapableFU` for a given DynInst as equivalent to the case where the DynInst obtained an FU, with default latency of 1. This is because there is a single if statement that checks whether the FU allocator returned `NoFreeFU` or not, and `NoCapableFU` happens to be different. The change is to introduce `NoNeedFU` and to panic if the FU allocator returns `NoCapableFU`

An improvement would be to use a strongly typed enum rather than integer constants. Thoughts ?

In addition to unit tests, I have tested this with `main.py run` and get panics if I remove support for `IntMul` type in `O3CPU.py` in:

```
./SuiteUID-asm-riscv-rv32um-ps-mul-o3-ALL-x86_64-opt/TestUID-asm-riscv-rv32um-ps-mul-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-asm-riscv-rv32um-ps-mulh-o3-ALL-x86_64-opt/TestUID-asm-riscv-rv32um-ps-mulh-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-asm-riscv-rv32um-ps-mulhsu-o3-ALL-x86_64-opt/TestUID-asm-riscv-rv32um-ps-mulhsu-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-asm-riscv-rv32um-ps-mulhu-o3-ALL-x86_64-opt/TestUID-asm-riscv-rv32um-ps-mulhu-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-asm-riscv-rv64um-ps-mul-o3-ALL-x86_64-opt/TestUID-asm-riscv-rv64um-ps-mul-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-asm-riscv-rv64um-ps-mulh-o3-ALL-x86_64-opt/TestUID-asm-riscv-rv64um-ps-mulh-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-asm-riscv-rv64um-ps-mulhsu-o3-ALL-x86_64-opt/TestUID-asm-riscv-rv64um-ps-mulhsu-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-asm-riscv-rv64um-ps-mulhu-o3-ALL-x86_64-opt/TestUID-asm-riscv-rv64um-ps-mulhu-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-asm-riscv-rv64um-ps-mulw-o3-ALL-x86_64-opt/TestUID-asm-riscv-rv64um-ps-mulw-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-BaseCPUProcessor-arm-hello-ALL-x86_64-opt/TestUID-BaseCPUProcessor-arm-hello-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-cpu_test_ArmDerivO3CPU_Bubblesort-ALL-x86_64-opt/TestUID-cpu_test_ArmDerivO3CPU_Bubblesort-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-cpu_test_ArmDerivO3CPU_FloatMM-ALL-x86_64-opt/TestUID-cpu_test_ArmDerivO3CPU_FloatMM-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-cpu_test_RiscvDerivO3CPU_Bubblesort-ALL-x86_64-opt/TestUID-cpu_test_RiscvDerivO3CPU_Bubblesort-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-cpu_test_RiscvDerivO3CPU_FloatMM-ALL-x86_64-opt/TestUID-cpu_test_RiscvDerivO3CPU_FloatMM-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-o3-cpu_1-cores_classic_DualChannelDDR3_1600_arm_boot_test_to-tick-ALL-x86_64-opt/TestUID-o3-cpu_1-cores_classic_DualChannelDDR3_1600_arm_boot_test_to-tick-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-o3-cpu_1-cores_classic_DualChannelDDR3_1600_riscv-boot-test_to-tick-ALL-x86_64-opt/TestUID-o3-cpu_1-cores_classic_DualChannelDDR3_1600_riscv-boot-test_to-tick-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-test-arm-hello32-static-o3-ALL-x86_64-opt/TestUID-test-arm-hello32-static-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-test-arm-hello64-static-o3-ALL-x86_64-opt/TestUID-test-arm-hello64-static-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-test-mips-hello-o3-ALL-x86_64-opt/TestUID-test-mips-hello-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-test-riscv-hello-o3-ALL-x86_64-opt/TestUID-test-riscv-hello-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
./SuiteUID-test-riscv-print-this-o3-ALL-x86_64-opt/TestUID-test-riscv-print-this-o3-ALL-x86_64-opt/simerr.txt:src/cpu/o3/inst_queue.cc:905: panic: Processor cannot execute opclass:2
```

### DIFF
--- a/src/cpu/o3/commit.cc
+++ b/src/cpu/o3/commit.cc
@@ -964,6 +964,17 @@ Commit::commitInsts()
 
             // Record that the number of ROB entries has changed.
             changedROBNumEntries[tid] = true;
+        // Inst at head of ROB cannot execute because the CPU
+        // does not know how to (lack of FU). This is a misconfiguration,
+        // so panic.
+        } else if (head_inst->noCapableFU() &&
+            head_inst->getFault() == NoFault)  {
+            panic("CPU cannot execute [sn:%llu] op_class: %u but"
+              " did not trigger a fault. Do you need to update"
+              " the configuration and add a functional unit for"
+              " that op class?\n",
+              head_inst->seqNum,
+              head_inst->opClass());
         } else {
             set(pc[tid], head_inst->pcState());
 
--- a/src/cpu/o3/dyn_inst.hh
+++ b/src/cpu/o3/dyn_inst.hh
@@ -187,6 +187,8 @@ class DynInst : public ExecContext, public RefCounted
         ReqMade,
         MemOpDone,
         HtmFromTransaction,
+        NoCapableFU,           /// Processor does not have capability to
+                               /// execute the instruction
         MaxFlags
     };
 
@@ -840,6 +842,16 @@ class DynInst : public ExecContext, public RefCounted
 
     /** Returns whether or not this instruction is squashed in the ROB. */
     bool isSquashedInROB() const { return status[SquashedInROB]; }
+
+    /** Mark this instruction as having attempted to execute
+     * but CPU did not have a capable functional unit.
+     */
+    void setNoCapableFU() { instFlags.set(NoCapableFU); }
+
+    /** Returns whether or not this instruction attempted
+     * to execute and found not capable FU.
+     */
+    bool noCapableFU() const { return instFlags[NoCapableFU]; }
 
     /** Returns whether pinned registers are renamed */
     bool isPinnedRegsRenamed() const { return status[PinnedRegsRenamed]; }
--- a/src/cpu/o3/inst_queue.cc
+++ b/src/cpu/o3/inst_queue.cc
@@ -812,7 +812,7 @@ InstructionQueue::scheduleReadyInsts()
             continue;
         }
 
-        int idx = FUPool::NoCapableFU;
+        int idx = FUPool::NoNeedFU;
         Cycles op_latency = Cycles(1);
         ThreadID tid = issuing_inst->threadNumber;
 
@@ -832,7 +832,8 @@ InstructionQueue::scheduleReadyInsts()
 
         // If we have an instruction that doesn't require a FU, or a
         // valid FU, then schedule for execution.
-        if (idx != FUPool::NoFreeFU) {
+        if (idx > FUPool::NoFreeFU || idx == FUPool::NoNeedFU ||
+            idx == FUPool::NoCapableFU) {
             if (op_latency == Cycles(1)) {
                 i2e_info->size++;
                 instsToExecute.push_back(issuing_inst);
@@ -841,7 +842,17 @@ InstructionQueue::scheduleReadyInsts()
                 // cycle if we used one.
                 if (idx >= 0)
                     fuPool->freeUnitNextCycle(idx);
+
+                // CPU has no capable FU for the instruction
+                // but this may be OK if the instruction gets
+                // squashed. Remember this and give IEW
+                // the opportunity to trigger a fault
+                // if the instruction is unsupported.
+                // Otherwise, commit will panic.
+                if (idx == FUPool::NoCapableFU)
+                  issuing_inst->setNoCapableFU();
             } else {
+                assert(idx != FUPool::NoCapableFU);
                 bool pipelined = fuPool->isPipelined(op_class);
                 // Generate completion event for the FU
                 ++wbOutstanding;
@@ -898,6 +909,7 @@ InstructionQueue::scheduleReadyInsts()
             listOrder.erase(order_it++);
             iqStats.statIssuedInstType[tid][op_class]++;
         } else {
+            assert(idx == FUPool::NoFreeFU);
             iqStats.statFuBusy[op_class]++;
             iqStats.fuBusy[tid]++;
             ++order_it;
--- a/src/cpu/o3/fu_pool.hh
+++ b/src/cpu/o3/fu_pool.hh
@@ -139,8 +139,31 @@ class FUPool : public SimObject
     FUPool(const Params &p);
     ~FUPool();
 
+    /**
+     * Named constants to differentiate cases where an
+     * instruction asked the FUPool for a free FU
+     * but did not get one
+     */
+
+    /**
+     * Instruction asked for a FU but does not actually
+     * need any (e.g., NOP)
+     */
+    static constexpr auto NoNeedFU = -3;
+
+    /**
+     * Instruction asked for a FU but this FUPool does
+     * not have a FU for this instruction op type
+     */
     static constexpr auto NoCapableFU = -2;
+
+    /**
+     * Instruction asked for a FU but all FU for
+     * this op type have already been allocated to
+     * other instructions this cycle
+     */
     static constexpr auto NoFreeFU = -1;
+
     /**
      * Gets a FU providing the requested capability. Will mark the
      * unit as busy, but leaves the freeing of the unit up to the IEW


giactra (changes requested):
Wow @aperais thanks for spotting what I believe is a serious problem in gem5. As you mentioned, using a typed enum would be better. But we can do this with a later PR. 

Can I just ask you to document 

    static constexpr auto NoNeedFU = -3;
    static constexpr auto NoCapableFU = -2;
    static constexpr auto NoFreeFU = -1;
    
With a comment each describing what they mean? Other than that it lgtm

aperais:
Good idea, done !

aperais:
By the way, the bug was discovered by one of my student (Chandana S. Deshpande) and if possible I would like to have her mentioned somewhere in the detailed description of the commit when/if the PR goes through. @giactra Is there a preferred way to achieve that ?

giactra:
> By the way, the bug was discovered by one of my student (Chandana S. Deshpande) and if possible I would like to have her mentioned somewhere in the detailed description of the commit when/if the PR goes through. @giactra Is there a preferred way to achieve that ?

In the past I have just added

> Kudos to “name of the person” for “reason” 

at the end of the commit message (here you can add it to the PR). But let’s see if @powerjg has some other preference

powerjg:
Anything is fine with me. You can also say "Co-author: email" if they co authored the patch

giactra:
Hi @aperais what have you decided to do? Let me know as I will approve the PR once changes have been made

aperais:
@giactra  Hi, I'm not sure I got it : I have two commits (latest to address your comment to add a description of the constants), so I was assuming I would have the opportunity to enter a modified message once the PR is approved (via commit squashing), but I guess that's not how this work.

So, probably a dumb question but which of the 2 commit messages should I amend ? Oldest ?

Edit : This was dumb, I'll do my own squash commit after discussion below is resolved.

aperais:
@giactra I am also getting second thoughts about the way this is fixed. The issue is that this can trigger a spurious panic if the CPU decodes an instruction that it is not supposed to handle on the wrong path. In theory, the decoder would raise an illegal instruction exception, but I don't think we configure the decoder with specific ISA support (like disabling SVE for Aarch64), so we might end up in this case. Actually, my own gcc binary compiled for Aarch64 does decode SVE instructions on the wrong path.

The options are:
1. What is in the PR, which risks causing spurious panics that are annoying to debug. However I can add a comment in the IQ to warn about this issue, but then the quick and dirty fix for any user seeing a spurious `panic` would be to comment out the `panic` which is counter productive.
2. Do `warn` and treat `NoCapableFU` as `NoFreeFU` in the IQ. If the unsupported instruction is in the correct path, this will loop forever, but if it is on the wrong path, it will work as intended.
3. Mark the unsupported instruction with a flag and `panic` if it reaches the head of the ROB. This works well but requires adding state to DynInst just to track this issue and deal with it later, rather than dealing it "in place".

Do you have a preference ? Option 3 seems the cleanest to me at this point.

giactra:
> @giactra I am also getting second thoughts about the way this is fixed. The issue is that this can trigger a spurious panic if the CPU decodes an instruction that it is not supposed to handle on the wrong path. In theory, the decoder would raise an illegal instruction exception, but I don't think we configure the decoder with specific ISA support (like disabling SVE for Aarch64), so we might end up in this case. Actually, my own gcc binary compiled for Aarch64 does decode SVE instructions on the wrong path.

Yeah good catch; this is a good point I didn't think of...

> 
> The options are:
> 
>     1. What is in the PR, which risks causing spurious panics that are annoying to debug. However I can add a comment in the IQ to warn about this issue, but then the quick and dirty fix for any user seeing a spurious `panic` would be to comment out the `panic` which is counter productive.

IMHO this is not really an option
> 
>     2. Do `warn` and treat `NoCapableFU` as `NoFreeFU` in the IQ. If the unsupported instruction is in the correct path, this will loop forever, but if it is on the wrong path, it will work as intended.
> 

The problem is that most instructions produce the undefined instruction fault only once they get executed. At the head of the execute logic they would check for the feature presence and return a fault if feature is not supported/disabled. If we prevent those instructions from being issued, we will never produce a fault and simulation will be stuck..

>     3. Mark the unsupported instruction with a flag and `panic` if it reaches the head of the ROB. This works well but requires adding state to DynInst just to track this issue and deal with it later, rather than dealing it "in place".
> 
> 
> Do you have a preference ? Option 3 seems the cleanest to me at this point.

Yes option 3 seems the most viable solution. Just to be clear the panicking condition should be HEAD of ROB and committing (not producing a fault)

giactra:
> @giactra Hi, I'm not sure I got it : I have two commits (latest to address your comment to add a description of the constants), so I was assuming I would have the opportunity to enter a modified message once the PR is approved (via commit squashing), but I guess that's not how this work.
> 
> So, probably a dumb question but which of the 2 commit messages should I amend ? Oldest ?
> 
> Edit : This was dumb, I'll do my own squash commit after discussion below is resolved.

Not dumb at all, but yes I would suggest you to squash and force push manually, especially as only mantainers can change the PR message at squash time...

aperais:
@giactra I have updated the PR and the commit message, plus done the same testing (removing IntMul capability from baseline O3CPU yields the same list of failures in the `main.py` tests).

As mentioned, I mark an instruction for which there is no existing FU as potentially problematic by adding a flag in the existing `DynInst` flags structure. I also set the `CanCommit` flag so that commit processes it if it is at the head of the ROB, at which point I panic because it has to be on the correct path. I panic before `commitHead` is called, which I believe is fine:

1. Any instruction that got a fault from execution will have done so by executing, so it go a functional unit, so it cannot cause the panic.
2. Any instruction that got a fault from fetch carries the fault with a nop (afaik), which does not need an FU so won't panic either.

aperais:
@giactra From my perspective this one is good to go, unless you have remaining concerns of course.

src/cpu/o3/dyn_inst.hh
```diff
@@ -841,6 +843,16 @@ class DynInst : public ExecContext, public RefCounted
...
 
+    /** Mark this instruction as having attempted to execute
+     * but CPU did not have a capable functional unit.
+     */
+    void setCannotExecute() { status.set(CannotExecute); }
+
+    /** Returns whether or not this instruction attempted
+     * to execute and found not capable FU.
+     */
+    bool cannotExecute() { return status[CannotExecute]; }
```
giactra: Nit: This should be marked as const

src/cpu/o3/inst_queue.cc
```diff
@@ -897,10 +897,22 @@ InstructionQueue::scheduleReadyInsts()
 
             listOrder.erase(order_it++);
             iqStats.statIssuedInstType[tid][op_class]++;
-        } else {
+        } else if (idx != FUPool::NoCapableFU) {
```
giactra: IMHO it would be more intuitive to write this as

`idx == FUPool::NoFreeFU`

as it is easier to understand under which condition the code will be executed

src/cpu/o3/inst_queue.cc
```diff
@@ -897,10 +897,22 @@ InstructionQueue::scheduleReadyInsts()
...
             ++order_it;
+        } else {
+            // CPU has no capable FU for the instruction
+            // but this may be OK if the instruction gets
+            // squashed. Remember this and act if the
+            // instruction is identified as being part of
+            // the correct path (reaches head of ROB)
+            // Treat it as having to wait (NoFreeFU) in the
+            // meantime, but mark it as ready to commit
+            // so commit processes it
```
giactra: I would consider adding an

`assert(idx ==  FUPool::NoCapableFU)`

src/cpu/o3/commit.cc
```diff
@@ -964,6 +964,14 @@ Commit::commitInsts()
 
             // Record that the number of ROB entries has changed.
             changedROBNumEntries[tid] = true;
+        // Inst at head of ROB cannot execute because the CPU
+        // does not know how to (lack of FU). This is a misconfiguration,
+        // so panic.
+        } else if(head_inst->cannotExecute())  {
```
giactra: Nit: keep a space between if and "("

src/cpu/o3/dyn_inst.hh
```diff
@@ -158,6 +158,8 @@ class DynInst : public ExecContext, public RefCounted
         SquashedInIQ,            /// Instruction is squashed in the IQ
         SquashedInLSQ,           /// Instruction is squashed in the LSQ
         SquashedInROB,           /// Instruction is squashed in the ROB
+        CannotExecute,           /// Processor does not have capability to
```
giactra: It is probably not important but I can't figure out if this should reside in the Status enum or in the Flags enum.... Looking at the enum cases it looks like the initial idea was to track the progress/status of the instruction in the pipeline. However later cases are using this to carry instruction behaviour (looking at the Serialize options for example) regardless of the position in the pipeline.

I believe in the future we should clean this up. Anyway I am not going to push strongly on this as it needs refactoring anyway. One reason to prefer Flags is that the CannotExecute is sort of misleading if checked against CanIssue description... I would honestly consider removing any reference to the exectution bit as the instruction will actually execute in gem5 (it will call StaticInst::execute), it will only abort if the cpu tries to commit it on a later stage. So maybe using NoCapableFU (similarly to what we are already using in the IQ)?

giactra (changes requested):

src/cpu/o3/dyn_inst.hh
aperais: Agree with you on the Status vs. Flags. 

However I do not follow you with the `StaticInst::execute`. With the patch, if O3CPU does not have a capable functional unit, the instruction does not issue and so `inst->execute();` in iew.cc will not be called. If you prefer I change it that is fine, but unless I'm mistaken in my understanding of how issue and iew work in O3CPU the instruction cannot and will not execute (with the patch).

giactra: > However I do not follow you with the `StaticInst::execute`. With the patch, if O3CPU does not have a capable functional unit, the instruction does not issue and so `inst->execute();` in iew.cc will not be called.

Yes you are right I was somehow convinced you were retaining old behaviour (instruction would issue), only you were panicking when trying to commit the instruction.

I would still try to issue for the reason mentioned above:

> The problem is that most instructions produce the undefined instruction fault only once they get executed. At the head of the execute logic they would check for the feature presence and return a fault if feature is not supported/disabled. If we prevent those instructions from being issued, we will never produce a fault and simulation will be stuck..

Now what happens if you run a simulation with a CPU without SIMD FUs and therefore SIMD disabled in the ISA?

* Real HW: The instruction won't be decoded and undefined instruction will be triggered
* Issue even if no FU: matching the HW; we issue/execute even with no FU and if the ISA logic is correct we trigger undefined instruction. If the ISA does not generate an exception, we abort at commit as a safety net.
* Not issueing (your implementation): You just abort and do not give a chance to the instruction to generate a fault during execute. So this forces the user to provision the O3CPU with SIMD FUs even if they are not using it...

So I would treat NoCapableFU as NoNeedFU, only I would flag the instruction to panic if we try to commit it.
Please let me know your thoughts

giactra: I know it is a niche case, but sometimes arch correctness means correctly handling wrong behaviour ;)

aperais: I see your point, but then I can just call inst->execute() functionally when it fails to find a capable FU. I assume on real HW it would not take up an issue slot (maybe not even an IQ entry) ?
 
Something like (inst_queue.cc):

```
} else if (idx == FUPool::NoFreeFU) {
  iqStats.statFuBusy[op_class]++;
  iqStats.fuBusy[tid]++;
  ++order_it;
} else {
  assert(idx == FUPool::NoCapableFU);
  // CPU has no capable FU for the instruction
  // but this may be OK if the instruction gets
  // squashed. 
  // First, execute the instruction to give a chance to
  // the CPU to behave as real hardware would and
  // trigger an illegal instruction fault that
  // will be carried to commit.
  // If no fault is triggered, remember there is no FU
  // and act if the instruction is identified as being part of
  // the correct path (reaches head of ROB)
  // In both cases, keep it in the IQ:
  // Fault will do an inclusive squash and
  // panic won't care
            
  issuing_inst->execute();

  // ISA did not trigger the fault
  // Remember it, commit will panic()
  if (issuing_inst->getFault() == NoFault)
  {
    issuing_inst->setCannotExecute();
  }

  issuing_inst->setCanCommit();
  // Do not try re-issuing it again
  listOrder.erase(order_it++);
}
```

giactra: I just don't particularly like the idea of fast-tracking the faulting instruction. This method is supposed to simply schedule the ready instructions; adding this specific case will complicate the pipeline model IMHO. Also, even assuming the instruction takes an issue slot, it does not seem seem important to me as either simulation will panic or future instructions will be squashed anyway (by the exception).

aperais: By "fast tracking", do you mean the case where the instruction is at the head of the ROB and the fault would trigger one cycle earlier than if the instruction had issued and gone to IEW to do the call to `execute()`? Because it will not fault until it reaches head of commit regardless of when it issues here (just so we are on the same page).

By letting the instruction be issued rather than treated here, the code here will just have to be split between inst_queue.cc (remember it did not get an FU, even if it issues) and iew.cc (execute(), and if fault, nothing, if no fault, tell commit to panic when instruction reaches commit). 

I agree that it complexifies the pipeline model, but I feel it's a win to modify two files instead of three. I'll go with what you decide but wanted to discuss it before.

giactra: > By "fast tracking", do you mean the case where the instruction is at the head of the ROB and the fault would trigger one cycle earlier than if the instruction had issued and gone to IEW to do the call to `execute()`? Because it will not fault until it reaches head of commit regardless of when it issues here (just so we are on the same page).

No by fast-tracking I don't mean I am concerned about the early execution;  I am concerned about the software engineering aspect of providing two different routes that go from point A (issue) to point B (execute).

Basically I don't like the extra addition of the execute( call in the instruction queue)

```
} else {
  assert(idx == FUPool::NoCapableFU);
  // CPU has no capable FU for the instruction
  // but this may be OK if the instruction gets
  // squashed. 
  // First, execute the instruction to give a chance to
  // the CPU to behave as real hardware would and
  // trigger an illegal instruction fault that
  // will be carried to commit.
  // If no fault is triggered, remember there is no FU
  // and act if the instruction is identified as being part of
  // the correct path (reaches head of ROB)
  // In both cases, keep it in the IQ:
  // Fault will do an inclusive squash and
  // panic won't care
            
  issuing_inst->execute();

  // ISA did not trigger the fault
  // Remember it, commit will panic()
  if (issuing_inst->getFault() == NoFault)
  {
    issuing_inst->setCannotExecute();
  }
```

It basically as if we are forking the pipeline at the IQ point (from a SW engineering POV) and let the faulting instruction flow in this "fast-track". I think the faulting instruction should follow the existin flow.

> 
> By letting the instruction be issued rather than treated here, the code here will just have to be split between inst_queue.cc (remember it did not get an FU, even if it issues) and iew.cc (execute(), and if fault, nothing, if no fault, tell commit to panic when instruction reaches commit).
> 

Sorry I don't understand what modification we need to make to iew.cc
What I thought would suffice is to treat NoCapableFU like NoFreeFU (they would reuse the same code as they have been doing until now), only in the first case we would annotate the instruction to panic at commit. 
Let me know if there's something I am missing

aperais: > Sorry I don't understand what modification we need to make to iew.cc
What I thought would suffice is to treat NoCapableFU like NoFreeFU (they would reuse the same code as they have been doing until now), only in the first case we would annotate the instruction to panic at commit.
Let me know if there's something I am missing

No I think you're right, I was thinking I would also need to change the iew stage to mark that the instruction was marked by IQ and did not fault, but commit can do it. I'll go with this, thanks for the feedback.

aperais: Alright, so I have changed the flag name in dyn_inst.hh. That flag is set in inst_queue.cc if there is no capable FU but the instruction will still issue. At commit, the instruction will either have faulted, which will go through the usual fault handling code, or will not, in which case commit will panic.

powerjg (commented):
Thanks for this fix @aperais, and thanks for taking all of the feedback into account. Thanks also to @giactra for the careful reviews!

I've been following the conversation, and this solution looks good to me. I'll let @giactra sign off, though.

powerjg (approved):
(I can approve since Giacomo requested changes)

giactra (approved):
Thanks @aperais for the patience in dealing with my review. I cannot stress enough how valuable this contribution is

aperais:
> Not dumb at all, but yes I would suggest you to squash and force push manually, especially as only mantainers can change the PR message at squash time...

@giactra Well this is a bit awkward, it turns out the commit message in the gem5/develop after the branch merged is not the commit message of the commit (it has no commit ID and no mention of my student :(), but whatever is in the opening PR. Should this have happened? Did I do something wrong?

giactra:
Urgh, this is because the PR message differs from the commit msg. I forgot that in a squash merge the PR gets selected instead of the commit. The commit Id is not relevant I just want to make sure your student gets the right recognition.

Would you be up for creating a PR which is changing the hardcoded -2 into the enum value?

https://github.com/gem5/gem5/blob/stable/src/cpu/o3/fu_pool.cc#L167

Then we can mention your student in that PR

PR #1519, opened 2024-08-27
Optimized riscv simulation performance

Kingfish404:
This PR is in progress and is ready for merging, aiming at improving the page table walk and TLB's performance. There will be more commits in the near future after https://github.com/gem5/gem5/pull/1387/ is merged.

The first commit https://github.com/gem5/gem5/pull/1519/commits/e1fd222aca8b6b0f0cb370701ec45ebf8ee7c9bc is aimed at solving the TODO below, when the RISC-V version is done, it could also be ported to other ISA(x86, etc).
https://github.com/gem5/gem5/blob/b1a44b89c7bae73fae2dc547bc1f871452075b85/src/arch/riscv/pagetable_walker.cc#L76-L78


powerjg:
Thanks for the contribution, @Kingfish404 ! 

Could you give some more detail in the commit message about the purpose and the design of this commit? It is difficult to review code without the context of why the code was modified and how the code was modified. 

Also, "feat" is not an official tag. "arch-riscv" would be better for this change. See https://github.com/gem5/gem5/blob/stable/MAINTAINERS.yaml for a list of tags.

ivanaamit:
Hi @Kingfish404,

```
f=.git/hooks/commit-msg
mkdir -p $f
chmod +x $f
```

aarmejach:
@saul44203 can probably take a look at this.

I though we were already coalescing request.

saul44203:
I think this looks good. The only thing is that we merged #672 a while ago with a similar idea, in this case squashing pending walks that already hit in the TLB before starting them. Maybe with this patch that's no longer necessary.

src/arch/riscv/pagetable_walker.cc
```diff
@@ -195,6 +223,41 @@ Walker::WalkerState::initState(ThreadContext * _tc,
...
+        "Coalescing walk for %#lx(pc=%#lx) into %#lx(pc=%#lx)\n",
+        req->getVaddr(), _req->getPC(), req->getVaddr(), req->getPC());
+        // add to list of requestors
+        Fault new_fault = NoFault;
+        if (timingFault != NoFault) {
+            // recreate fault for this txn, we don't have pmp yet
+            // TODO: also consider pmp's addr fault
+            DPRINTF(PageTableWalker, "FAULT in tryCoalesce\n");
+        }
+        return std::make_pair(true, new_fault);
```
powerjg, line 256: I'm a bit confused with this code. By my understanding, the coalesced walk will *always* return `NoFault`, even when `timingFault` is not `NoFault`. Is this the intended behavior?

src/arch/riscv/pagetable_walker.cc
```diff
@@ -195,6 +223,41 @@ Walker::WalkerState::initState(ThreadContext * _tc,
...
+
+    addr_match = ((pre_match_num >> PageShift) << PageShift) ==
+                 ((addr_match_num >> PageShift) << PageShift);
+
+    if (priv_match && addr_match) {
+        // coalesce
+        DPRINTF(PageTableWalker,
+        "Coalescing walk for %#lx(pc=%#lx) into %#lx(pc=%#lx)\n",
+        req->getVaddr(), _req->getPC(), req->getVaddr(), req->getPC());
+        // add to list of requestors
```
powerjg, line 249: Where is the list of requestors?

Is there a maximum size for this list? In real hardware, there would be a maximum number of transactions that could be coalesced. Are we modeling this? If so, how?

powerjg (commented):
I have some questions below. I could be missing something obvious, but I currently don't understand how this implementation works.

ivanaamit:
Hi @Kingfish404, could you please address Jason's comments/questions? Thanks!

src/arch/riscv/pagetable_walker.cc
Kingfish404: If the coalesced walk returns `Fault`, the previous walk should have already returned a `Fault`, but maybe this info that comes from a later coalesced walk could also be returned. I will fix this to make sure that this duplicate `Fault` will not be used later.

src/arch/riscv/pagetable_walker.cc
Kingfish404: The list of requestors is `currStates`.

In hypervisor mode (RISC-V H Extension), the TLB request will be more frequent, while multiple in-flight TLB requests could co-exist. This should be orthogonal work to squashing already hit TLB https://github.com/gem5/gem5/pull/672.

This pr aims to improve the in-flight TLB request performance, as mentioned above wait for H Extension pr, expect to work like Miss Status Handling Registers (MSHR) for TLB (Still not yet finished).

Yes, the maximum size should be considered, but not yet been decided.

hnpl: @powerjg I think the real problem here is that the PTW can hold an infinite number of requests.

For coalescing, we can do this at either when we enqueue a request, or when a walk returns, and we can definitely limit the number of requests that can return (to the core, etc.) per cycle.

src/arch/riscv/pagetable_walker.cc
hnpl: My understanding is that you're coalescing a request with other requests when the new request is enqueued, and you keep track of the tag of each request.
I think we can handle the Fault as follows when a walk returns,
- In the case of NoFault, we can return NoFault to all of the coalesced requests.
- In the case of Fault, we can return Fault to one of the coalesced requests, and let the rest of the requests to try walking again.

src/arch/riscv/pagetable_walker.cc
saul44203: Hey @Kingfish404, have you tested this and got better performance? I'm interested in this because I can see a possible scenario where there's an uncoalesced request which benefits from the squash in #672, but the inverse is trickier (maybe only when the maximum number of squashed requests is reached?).

erin-le:
@Kingfish404 Hello, thank you for your contribution. Could you give an update on the status of this PR?

Kingfish404:
Unfortunately, the benchmarks I've run so far in the current branch don't show a noticeable performance improvement (e.g., > 1% to boot Linux).

I'll close this Pr first, wait for the H extension to be merged, and then modify and verify it again before opening it with the update if it works.

erin-le:
@Kingfish404 Hello, in case you wanted to keep working on this PR, the RISC-V H extension was recently merged into gem5 v25.0. Thanks for your contribution!

PR #1366, opened 2024-07-19
arch-vega: Multiple SOPC fixes

abmerop:
Make S_CMP_LT_U32 use < instead of <=. Change types of EQ / LG for U64 to be U64.

Change-Id: Ib0b3b7a46ba1aff16a6d439302ca087d988d6417

PR #1520, opened 2024-08-27
Revert "arch-vega: Multiple SOPC fixes"

abmerop:
Reverts gem5/gem5#1366.

This is not correct.

### DIFF
--- a/src/arch/amdgpu/vega/insts/sopc.cc
+++ b/src/arch/amdgpu/vega/insts/sopc.cc
@@ -368,7 +368,7 @@ namespace VegaISA
         src0.read();
         src1.read();
 
-        scc = (src0.rawData() < src1.rawData()) ? 1 : 0;
+        scc = (src0.rawData() <= src1.rawData()) ? 1 : 0;
 
         scc.write();
     } // execute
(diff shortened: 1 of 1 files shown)


src/arch/amdgpu/vega/insts/sopc.cc
```diff
@@ -556,8 +556,8 @@ namespace VegaISA
     void
     Inst_SOPC__S_CMP_EQ_U64::execute(GPUDynInstPtr gpuDynInst)
     {
-        ConstScalarOperandU64 src0(gpuDynInst, instData.SSRC0);
-        ConstScalarOperandU64 src1(gpuDynInst, instData.SSRC1);
+        ConstScalarOperandI64 src0(gpuDynInst, instData.SSRC0);
```
mattsinc, line 559: Why is this a bug?  The instruction is U64, not I64 ...

src/arch/amdgpu/vega/insts/sopc.cc
```diff
@@ -584,8 +584,8 @@ namespace VegaISA
     void
     Inst_SOPC__S_CMP_LG_U64::execute(GPUDynInstPtr gpuDynInst)
     {
-        ConstScalarOperandU64 src0(gpuDynInst, instData.SSRC0);
-        ConstScalarOperandU64 src1(gpuDynInst, instData.SSRC1);
+        ConstScalarOperandI64 src0(gpuDynInst, instData.SSRC0);
```
mattsinc, line 587: Why is this a bug?  The instruction is U64, not I64 ...

abmerop: This *should* still work? Unless the compiler sign extends. Probably safer to make a new PR just removing the real problem.

PR #1525, opened 2024-08-30
arch-riscv: Add support for riscv hardware probing syscall

5265325:
This PR adds the support for riscv hardware probing syscall described in [this](https://docs.kernel.org/arch/riscv/hwprobe.html). The implementation logic refers to [linux kernel](https://github.com/torvalds/linux/blob/master/arch/riscv/kernel/sys_hwprobe.c) and [qemu](https://github.com/qemu/qemu/blob/master/linux-user/syscall.c). And passed the [RISC-V hwprobe exmaple](https://github.com/cyyself/hwprobe) test.

Hope to be merged. Thanks.

Change-Id: Iab714974f0551fc451e0d6846c75a7153809a308

### DIFF
--- a/src/arch/riscv/linux/linux.hh
+++ b/src/arch/riscv/linux/linux.hh
@@ -34,6 +34,7 @@
 #include "arch/riscv/utility.hh"
 #include "kern/linux/flag_tables.hh"
 #include "kern/linux/linux.hh"
+#include "base/bitfield.hh"
 
 namespace gem5
 {
@@ -42,6 +43,101 @@ class RiscvLinux : public Linux
 {
   public:
     static const ByteOrder byteOrder = ByteOrder::little;
+
+    enum RiscvHwprobeKey
+    {
+        Mvendorid,
+        Marchid,
+        Mimpid,
+        BaseBehavior,
+        IMAExt0,
+        Cpuperf0,
+        ZicbozBlockSize,
+        HighestVirtAddress,
+        TimeCsrFreq,
+        MisalignedScalarPerf
+    };
+
+    /* Increase RISCV_HWPROBE_MAX_KEY when adding items. */
+    #define RISCV_HWPROBE_MAX_KEY 9
+
+    BitUnion64(key_base_behavior_t)
+    Bitfield<0> ima;
+    EndBitUnion(key_base_behavior_t)
+
+    BitUnion64(key_ima_ext_0_t)
+        Bitfield<49> ZAWRS;
+        Bitfield<48> ZCMOP;
+        Bitfield<47> ZCF;
+        Bitfield<46> ZCD;
+        Bitfield<45> ZCB;
+        Bitfield<44> ZCA;
+        Bitfield<43> ZIMOP;
+        Bitfield<42> ZVE64D;
+        Bitfield<41> ZVE64F;
+        Bitfield<40> ZVE64X;
+        Bitfield<39> ZVE32F;
+        Bitfield<38> ZVE32X;
+        Bitfield<37> ZIHINTPAUSE;
+        Bitfield<36> ZICOND;
+        Bitfield<35> ZACAS;
+        Bitfield<34> ZTSO;
+        Bitfield<33> ZFA;
+        Bitfield<32> ZVFHMIN;
+        Bitfield<31> ZVFH;
+        Bitfield<30> ZIHINTNTL;
+        Bitfield<29> ZFHMIN;
+        Bitfield<28> ZFH;
+        Bitfield<27> ZVKT;
+        Bitfield<26> ZVKSH;
+        Bitfield<25> ZVKSED;
+        Bitfield<24> ZVKNHB;
+        Bitfield<22> ZVKNHA;
+        Bitfield<21> ZVKNED;
+        Bitfield<20> ZVKG;
+        Bitfield<19> ZVKB;
+        Bitfield<18> ZVBC;
+        Bitfield<17> ZVBB;
+        Bitfield<16> ZKT;
+        Bitfield<15> ZKSH;
+        Bitfield<14> ZKSED;
+        Bitfield<13> ZKNH;
+        Bitfield<12> ZKNE;
+        Bitfield<11> ZKND;
+        Bitfield<10> ZBKX;
+        Bitfield<9>  ZBKC;
+        Bitfield<8>  ZBKB;
+        Bitfield<7>  ZBC;
+        Bitfield<6>  ZICBOZ;
+        Bitfield<5>  ZBS;
+        Bitfield<4>  ZBB;
+        Bitfield<3>  ZBA;
+        Bitfield<2>  V;
+        Bitfield<1>  C;
+        Bitfield<0>  FD;
+    EndBitUnion(key_ima_ext_0_t)
+
+    enum MisalignedScalarPerf
+    {
+        Unknown,
+        Emulated,
+        Slow,
+        Fast,
+        Unsupported
+    };
+
+    /* Flags */
+    #define RISCV_HWPROBE_WHICH_CPUS	(1 << 0)
+
+    struct riscv_hwprobe {
+        int64_t  key;
+        uint64_t value;
+    };
+
+    typedef struct cpumask {
+        size_t size;
+        uint64_t bits[];
+    } cpumask_t;
 };
 
 class RiscvLinux64 : public RiscvLinux, public OpenFlagTable<RiscvLinux64>
--- a/src/arch/riscv/linux/se_workload.cc
+++ b/src/arch/riscv/linux/se_workload.cc
@@ -44,6 +44,8 @@
 #include <sys/syscall.h>
 
 #include "arch/riscv/process.hh"
+#include "arch/riscv/insts/static_inst.hh"
+#include "arch/riscv/regs/misc.hh"
 #include "base/loader/object_file.hh"
 #include "base/trace.hh"
 #include "cpu/thread_context.hh"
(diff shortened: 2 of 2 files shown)


src/arch/riscv/linux/linux.hh
```diff
@@ -42,6 +42,86 @@ class RiscvLinux : public Linux
 {
   public:
     static const ByteOrder byteOrder = ByteOrder::little;
+
+    #define RISCV_HWPROBE_KEY_MVENDORID	0
```
powerjg: I think this would be better implemented as a `BitUnion`

See for instance the `types.hh` file.

src/arch/riscv/linux/se_workload.cc
```diff
@@ -123,6 +125,179 @@ unameFunc32(SyscallDesc *desc, ThreadContext *tc, VPtr<Linux::utsname> name)
...
+    BufferArg buf(cpus, target_size);
+    buf.copyIn(SETranslatingPortProxy(tc));
+
+    for (i = 0 ; i < target_size; i++) {
+        unsigned bit = i * target_bits;
+        uint8_t val = *((uint8_t *)buf.bufferPtr() + i);
+
+        for (j = 0; j < target_bits; j++, bit++) {
+            if (val & (1UL << j)) {
+                host_mask[bit / host_bits] |= 1UL << (bit % host_bits);
```
powerjg: Please use the functions in `bitfield.hh` instead of manually reimplementing them. This will make bugs less likely.

src/arch/riscv/linux/se_workload.cc
```diff
@@ -123,6 +125,179 @@ unameFunc32(SyscallDesc *desc, ThreadContext *tc, VPtr<Linux::utsname> name)
...
+              VPtr<> cpus)
+{
+    int ret, i, tmp;
+    size_t host_mask_size, target_mask_size;
+    unsigned long *host_mask;
+
+    tmp = (8 * sizeof(cpusetsize));
+    target_mask_size = ((cpusetsize + tmp - 1) / tmp) * sizeof(cpusetsize);
+    host_mask_size = (target_mask_size + (sizeof(*host_mask) - 1)) &
+                     ~(sizeof(*host_mask) - 1);
```
powerjg: Can you use functions in `bitfield.hh` to make this more readable?

If that's impossible, then adding comments to explain the logic is necessary.

src/arch/riscv/linux/se_workload.cc
```diff
@@ -123,6 +125,179 @@ unameFunc32(SyscallDesc *desc, ThreadContext *tc, VPtr<Linux::utsname> name)
...
+template <class OS>
+static SyscallReturn
+riscvHWProbeFunc(SyscallDesc *desc, ThreadContext *tc,
+                 VPtr<> pairs,
+                 typename OS::size_t pair_count,
+                 typename OS::size_t cpusetsize,
+                 VPtr<> cpus,
+                 unsigned int flags)
+{
+    int ret;
```
powerjg: This should be on line 284

src/arch/riscv/linux/se_workload.cc
```diff
@@ -123,6 +125,179 @@ unameFunc32(SyscallDesc *desc, ThreadContext *tc, VPtr<Linux::utsname> name)
...
+
+    /* flags must be 0 */
+    if (flags != 0) {
+        return -EINVAL;
+    }
+
+    /* check cpu set */
+    if (cpusetsize != 0) {
+        ret = cpu_set_valid<OS>(tc, cpusetsize, cpus);
```
powerjg:
```suggestion
        int ret = cpu_set_valid<OS>(tc, cpusetsize, cpus);
```

powerjg (dismissed):
Thanks for the contribution! I would be happy to see support for hardware probing merged. The current code is pretty hard to follow and I believe it can be simplified by using some of gem5's built in functions/APIs. See below for details.

5265325:
@powerjg I rewrote the code. The code is now closer to the linux kernel than to qemu to improve readability.

ivanaamit:
@5265325, Could you please mark the change requests as resolved if you've addressed them? This will make it easier for others to review or re-review. Thank you.

5265325:
> marked this conversation as resolved.

I have marked all conversations as resolved.  Is that right?

src/arch/riscv/linux/se_workload.cc
```diff
@@ -371,6 +735,8 @@ SyscallDescTable<SEWorkload::SyscallABI64> EmuLinux::syscallDescs64 = {
     { 241,  "perf_event_open" },
     { 242,  "accept4" },
     { 243,  "recvmmsg" },
+    { 244,  "specific_syscall" },
```
hnpl: Can you remove this line? I don't think RISC-V has the system call 244.

https://github.com/torvalds/linux/blob/88264981f2082248e892a706b2c5004650faac54/scripts/syscall.tbl#L292

src/arch/riscv/linux/se_workload.cc
```diff
@@ -710,6 +1076,8 @@ SyscallDescTable<SEWorkload::SyscallABI32> EmuLinux::syscallDescs32 = {
...
```
hnpl: Can you remove this line? I don't think RISC-V has the system call 244.

https://github.com/torvalds/linux/blob/88264981f2082248e892a706b2c5004650faac54/scripts/syscall.tbl#L292

src/arch/riscv/linux/se_workload.cc
```diff
@@ -123,6 +125,368 @@ unameFunc32(SyscallDesc *desc, ThreadContext *tc, VPtr<Linux::utsname> name)
...
+    case RiscvLinux::BaseBehavior:
+        {
+            MISA misa = tc->readMiscRegNoEffect(MISCREG_ISA);
+            RiscvLinux::key_base_behavior_t *base_behavior = (RiscvLinux::key_base_behavior_t *)&pair->value;
+            if (misa.rvi && misa.rvm && misa.rva) {
+                base_behavior->ima = 1;
+            }
+        }
+        break;
+    case RiscvLinux::IMAExt0:
```
hnpl, line 291: You don't need to make any change here.

Though, it'll be great if we have a single entity managing all riscv extensions. Right now, we have two places that do this, here and https://github.com/gem5/gem5/blob/ba02266260a1babc6c05d44ee9f24c87b83d5fd7/src/arch/riscv/RiscvISA.py

src/arch/riscv/linux/se_workload.cc
```diff
@@ -123,6 +125,368 @@ unameFunc32(SyscallDesc *desc, ThreadContext *tc, VPtr<Linux::utsname> name)
...
+    return pair->value == other_pair->value;
+}
+
+static inline int
+get_cpu_online_mask(ThreadContext *tc,
+                   uint64_t **cpu_online_mask,
+                   size_t *cpu_mask_size)
+{
+    /* 8-bytes up-boundary alignment */
+    *cpu_mask_size = (tc->getSystemPtr()->threads.size() + sizeof(**cpu_online_mask) - 1) /
```
hnpl: Note: this is rounding the `number_of_threads` to the next multiple of `size_of_each_mask`.

src/arch/riscv/linux/se_workload.cc
```diff
@@ -123,6 +125,368 @@ unameFunc32(SyscallDesc *desc, ThreadContext *tc, VPtr<Linux::utsname> name)
     return 0;
 }
 
+static inline void
+cpumask_set_cpu(unsigned int cpu,
+               uint64_t *dstp)
```
hnpl: style-nit: please fix the line break here and below. If the params fit within 79 characters, you don't need to break the line.

```cpp
cpumask_set_cpu(unsigned int cpu, uint64_t *dstp)
```

src/arch/riscv/linux/se_workload.cc
```diff
@@ -123,6 +125,368 @@ unameFunc32(SyscallDesc *desc, ThreadContext *tc, VPtr<Linux::utsname> name)
...
+    case RiscvLinux::Cpuperf0:
+        return true;
+    }
+
+    return false;
+}
+
+static inline bool
+riscv_hwprobe_pair_cmp(RiscvLinux::riscv_hwprobe *pair,
+                      RiscvLinux::riscv_hwprobe *other_pair)
```
hnpl: style-nit: please fix the params alignment here and below.

```cpp
riscv_hwprobe_pair_cmp(RiscvLinux::riscv_hwprobe *pair,
                       RiscvLinux::riscv_hwprobe *other_pair)
```

hnpl (commented):
I left a few comments, most are notes for future improvement, but there are a few style nits that need to be fixed.

The code logic looks okay to me as it follows the kernel implementation.

hnpl:
It would be nice if the struct for the cpu_mask to simplify the code.

5265325:
> It would be nice if the struct for the cpu_mask to simplify the code.

I have fixed the style problem you mentioned. In addition, I added the cpumask_t struct to simplify the code.

hnpl (approved):
Thanks!

Harshil2107 (dismissed):
Not an expert but it ltgm. I also looks like all the requested changes were addressed.

src/arch/riscv/linux/se_workload.cc
```diff
@@ -134,6 +136,381 @@ unameFunc32(SyscallDesc *desc, ThreadContext *tc, VPtr<Linux::utsname> name)
...
+
+    int ret = 0;
+    RiscvLinux::riscv_hwprobe *pair;
+    RiscvLinux::cpumask_t *cpu_online_mask = get_cpu_online_mask(tc);
+    RiscvLinux::cpumask_t *cpus = cpumask_malloc(tc);
+    BufferArg pairs_buf(pairs, sizeof(RiscvLinux::riscv_hwprobe) * pair_count);
+
+    if (cpu_online_mask == nullptr || cpus == nullptr) {
+        ret = -ENOMEM;
+        goto exit;
```
powerjg: I would strongly prefer not having "goto" and instead duplicating the code. It's only used in one or two places, so I don't think that it's worth the downsides.

Also, I strongly prefer *always* using `{ }` around if/while statements. I have had to fix multiple very difficult to find bugs because people want to save 2 characters 😉

powerjg: Re: `goto`: There's only three places in the entire gem5 codebase this is used. Let's not add another, if we can possibly avoid it.

5265325: I've removed all `goto` and added `{ }`.

powerjg (approved):

Issue #885, opened 2024-02-21
arch-riscv: remove branching from `vset{i}vl{i}` instruction and allow registers `vl` and `vtype` to be renamed


saul44203:
Currently the `vset{i}vl{i}` family of instructions is branch-like. This works well for some applications, but for others, especially those where the `vset{i}vl{i}` instruction is inside a loop which changes the `vl`/`vtype` state on every iteration, do not perform well due to constant squashing.

The proposal aims to eliminate the branch-like behavior of the `vset{i}vl{i}` instructions and, instead, set the `vl` and `vtype` registers as renameable, allowing for normal data dependencies to be created with subsequent vector instructions in the execution flow.

@adriaarmejach and I discussed two approaches:
* One where both registers are treated as integer registers.
* One where both registers pertain to a new class of renameable misc registers.

The later approach seems cleaner (new register class addition would be similar to the changes in commit fed81f34084ad46fc663ec236ce0e700881cc3c2) but it involves more overall code compared with the first one. 

Open to discussion.

powerjg:
I'm going to push back on this solution because the spec says that `vset{i}vl{i}` is a control dependency, not a data dependency. See https://github.com/riscv/riscv-v-spec/blob/v1.0/v-spec.adoc#9-vector-memory-consistency-model

Correct me if I'm not interpreting the spec correctly :)

aarmejach:
Yes, the spec says `vl` dependencies are treated as control, which allow speculation to happen as long as `vl` remains unchanged. However, this is not the case in many applications, specially sparse kernels that typically change `vl` often, which end up performing worse than scalar versions.

Having `vl` as a data dependency is more restrictive, does not break spec, and would serialize execution always. So some commercial proposals are actually implementing `vl` dependencies as data dependencies and allow both `vtype` and `vl` CSRs to be renamed. Which is bound to perform well in cases where `vl` stays constant as well as in sparse applications.

I understand that gem5 should be as close to spec as possible, so if this is not something we want to have upstream we will keep it internal.

powerjg:
It's an intriguing option. I also think looking into speculating on `vl` would also be an interesting direction.

I think I know the answer to this question, but I'll ask anyway. Do you think it's feasible to make it a configuration-time parameter whether `vl` is treated as a control or data dependence? It would be quite interesting if it was parameterized like whether the LSQs enforce TSO.

[...]

PR #1526, opened 2024-08-30
arch-riscv: Fix the bug of vsetivli frequently flushing the pipeline

5265325:
This PR fix the bug of vsetivli frequently flushing the pipeline.

Here are two pictures of the pipeline illustrate this phenomenon.

The vsetivli(0x00013334.0) instruction in the first picture flushes the pipeline every time it is executed. This is due to vsetivli being incorrectly flagged as a 'DirectControl' instruction. The branch predictor cannot predict it correctly.

The second picture is the pipeline after fixing the bug.

Change-Id: I5bede47919c06cea86fa23a81624b502fbdc1159

### DIFF
--- a/src/arch/riscv/isa/formats/vector_conf.isa
+++ b/src/arch/riscv/isa/formats/vector_conf.isa
@@ -202,6 +202,7 @@ def template VSetiVliBranchTarget {{
             rpc.vl(), requested_vl, vlmax, rd_bits, rs1_bits);
 
         std::unique_ptr<PCState> npc(dynamic_cast<PCState*>(rpc.clone()));
+        npc->set(rvSext(npc->pc() + 4));
         npc->vtype(new_vtype);
         npc->vl(new_vl);
         return npc;


aarmejach:
IIRC this is marked as DirectControl as we need to ensure the vsetvl instruction is committed before fetching/decoding the next instructions. The pipeline is not flushed, its stalled.

This is necessary since the vsetvl instruction may change `sew`, `vl`, `lmul`, etc. And these need to be known when fetching and decoding subsequent instructions.

5265325:
> DirectControl
In current gem5, vset{i}vl{i} is treated as an indirect branch. The BTB/Indirect Predictor is able to predict vl and vtype included in [PCState](https://github.com/gem5/gem5/blob/b1a44b89c7bae73fae2dc547bc1f871452075b85/src/arch/riscv/pcstate.hh#L69). And speculatively execute vset{i}vl{i}. Just like [vsetvl](https://github.com/gem5/gem5/blob/b1a44b89c7bae73fae2dc547bc1f871452075b85/src/arch/riscv/isa/decoder.isa#L4803) and [vsetvli](https://github.com/gem5/gem5/blob/b1a44b89c7bae73fae2dc547bc1f871452075b85/src/arch/riscv/isa/decoder.isa#L4787) are set to IsIndirectControl. In contrast, vsetivli is incorrectly set to DirectControl. As a result, the BTB/Indirect Predictor cannot accurately predict PCState that contain vl and vtype. Blocking or flushing pipeline helps with correctness. But it affected performance.
If there is a mistake in my understanding, please correct me.

powerjg:
cc: @hnpl

powerjg:
I am curious about two things:
1. Can you give us an IPC difference between the two implementations for a simple benchmark?
2. What tests have you run to ensure the change doesn't break anything else?

I know that we discussed predicting the outcome of vsetvli with the branch predictor, but I don't know if that was ever implemented in a way that made sense.

hnpl:
The way vtype prediction works is that, the vector instruction followed has the same type configuration as the previous instruction.

This approach has at least two downsides,
1. There's almost always a `vtype` misspeculation during the first iteration of the loop. This is a small downside if the number of elements is large.
2. If you have more than 1 vset{i}vl{i} in your loop, it's very likely that you'll encounter a lot of misspeculation for every 
iteration. This is the biggest downside of this approach.

Having said that, I believe `vsetivli` encodes almost all necessary all information about vtype, so the decoder should be able to figure out almost all information of the `vtype` (except for, maybe, `vill`, which the prediction is likely to be correct). So, there should not be a lot of misspeculation when it comes to speculating after a `vsetivli` instruction.

Can you post the disassemble code of the code block corresponding to the above pictures? It's hard to tell if the above case falls within the two known downsides, or it is a new problem.

5265325:
@powerjg @hnpl [Here](https://github.com/user-attachments/files/16837891/vsetivli.tar.gz) is a simple test.

You can compile and test with the following command:
```
# Compile it. Or directly use the compiled elf in the gz file.
make -C path_to_vsetivli
./build/RISCV/gem5.opt --debug-flags=O3PipeView --debug-start=60000000 --debug-file=trace.out configs/deprecated/example/se.py --cpu-type=O3CPU --caches --bp-type TAGE_SC_L_64KB --indirect-bp-type SimpleIndirectPredictor --cmd path_to_vsetivli/vsetivli
./util/o3-pipeview.py -c 500 -o pipeview.out --color m5out/trace.out
less -r pipeview.out
# Search 'vsetivli' to locate where it first appeared
```

This is a simple 100 times loop. Vsetivli flushes the pipeline every time after each execution. After modifying the flag to ’IsIndirectControl‘, vsetivli only flushed the pipeline in the first three loops. And you can observe the number of vadd_vv_micro to confirm whether vetivli is executed correctly. (Default vlen=256)

The two pictures below are part of the loop. The first picture is before the bug is fixed, and the second picture is after the bug is fixed.
截屏2024-09-02 下午9 22 58
截屏2024-09-02 下午9 23 35

hnpl:
(The first vsetvli instruction is at `0x10520`, and the second vsetvli instruction is at `0x10528`.)

The first picture is actually what I expected from the instructions trace, as the second vsetvli instruction modifies vtype reg to a different value compared to the vtype reg set by the first vsetvli instruction. The way the current out-of-order vector instruction works is that, the current vector instruction is speculated to share the same vtype as the previous vector instruction. In this case, the second vsetvli modifies vtype reg, so the second vsetvli causes a misprediction.

I'm curious about the correctness of the second picture. When the second vsetvli instruction is at the decode time, the first vsetvli instruction has yet to update the vtype reg, and I wonder if the vector instructions followed by the second vsetvli instruction has the correct element width.

5265325:
> (The first vsetvli instruction is at `0x10520`, and the second vsetvli instruction is at `0x10528`.)
> 
> The first picture is actually what I expected from the instructions trace, as the second vsetvli instruction modifies vtype reg to a different value compared to the vtype reg set by the first vsetvli instruction. The way the current out-of-order vector instruction works is that, the current vector instruction is speculated to share the same vtype as the previous vector instruction. In this case, the second vsetvli modifies vtype reg, so the second vsetvli causes a misprediction.
> 
> I'm curious about the correctness of the second picture. When the second vsetvli instruction is at the decode time, the first vsetvli instruction has yet to update the vtype reg, and I wonder if the vector instructions followed by the second vsetvli instruction has the correct element width.

The current decoding uses the vl and vtype predicted by the BPU instead of directly using the results of the previous vset{i}vl{i} instruction. Just like value prediction, the vl and vtype dependency have been eliminated by prediction.

[Here](https://github.com/user-attachments/files/17341100/vsetvli.tar.gz) is a similar test, the main difference is that the vsetivli instruction is replaced with vsetvli. Testing is conducted on the gem5 develop branch. You can see that the pipeline looks similar to picture 2 above. Perhaps this can help prove that there is a bug in the implementation of the vsetivli instruction.

截屏2024-10-11 下午5 54 40

Additionally, new changes have been tested using [this](https://github.com/5265325/rvv-intrinsic-doc/tree/gem5/examples). These test sets are mentioned in #1542. Perhaps this can help prove the correctness of the code.

BobbyRBruce:
@hnpl : Can you give this another review? 💅🏻

hnpl:
@5265325 You're right. `vsetivli` does not necessarily need to speculate `vtype` reg as it has the content of the reg in the instruction, and thus we know `vtype` at decode time. I'll do some checking and review the code.

5265325:
> @5265325 You're right. `vsetivli` does not necessarily need to speculate `vtype` reg as it has the content of the reg in the instruction, and thus we know `vtype` at decode time. I'll do some checking and review the code.

I will prepare new optimization in the future. include:
1. Cancel speculation for vsetivli. It means vsetivli will no longer be a control flow instruction.
2. vl and vtype are predicted only for vsetvl and vsetvli. The decoder will maintain temporary vl and vtype. When decoding instructions, vl and vtype will be used from the decoder instead of directly from the bpu. The vl and vtype predicted by bpu will be used to update the vl and vtype of the decoder. This means that other control flow instructions will no longer flush the pipeline due to mispredictions of vl and vtype.

I will submit a patch and examples to illustrate these issues after development is completed. But I'm not sure exactly when. This patch is currently a temporary fix. It can be merged to temporarily fix performance bug of vsetivli if the review passes.

saul44203:
Hey, just giving my two cents here.

I gave a look at this and the provided example and I think the current problem with `vsetivli` stems from this code:
https://github.com/gem5/gem5/blob/3fc6cc7763e67ebab76358845b06b94060dd19be/src/cpu/o3/decode.cc#L709-L735

Since `vsetivli` has the flag `IsDirectControl` it'll go inside this and check whether the branch target is the same as the branch predicted pc. However, although the predictor does a good job with the vtype and vl speculation, this seems to never be the case, as the branch predictor returns pc+4 but the instruction's `branchTarget` always gives pc+0:
https://github.com/gem5/gem5/blob/3fc6cc7763e67ebab76358845b06b94060dd19be/src/arch/riscv/isa/formats/vector_conf.isa#L188-L209

Because of this the instruction is squashed on decode and it's `predPC` is overwritten with the pc+0 from `branchTarget` (see first code snippet).  

Finally, when this instruction arrives at the execution stage, misprediction is checked again, and because the current `predPC` address is pc+0 it doesn't match the expected next pc and squashes again:
https://github.com/gem5/gem5/blob/fce42880b945edde627cae42cdab41a84b91105b/src/cpu/o3/iew.cc#L1259-L1272
https://github.com/gem5/gem5/blob/b1a44b89c7bae73fae2dc547bc1f871452075b85/src/cpu/o3/dyn_inst.hh#L527-L534

So, in summary, `branchTarget` is overwriting a correct predicted pc with an incorrect one. Changing `vsetivli` to `IsIndirectControl` works because `branchTarget` is never executed (which is also dead code for `vsetvl` and `vsetvli`), but I think we can keep the `IsDirectControl` flag and fix the `branchTarget` (i.e. return pc+4) to be able to squash earlier on decode if we want to keep speculation for `vsetivli`. If not we can discuss what to do.

EDIT: Just as an aside, last comment of #885 might be of interest if speculation for `vsetivli` is to be removed.

hnpl:
@saul44203 Thanks for taking a look!
@5265325 I checked the execution trace when vsetivli is marked as IsIndirectControl and didn't find any error in the trace. Though, can you try implement @saul44203's approach?

5265325:
@saul44203 Thanks for taking a look!
@hnpl I have modified the code according to @saul44203's approach.

powerjg (approved):
Thanks for the great conversation and the elegant solution!

aarmejach (approved):
LGTM

PR #1530, opened 2024-08-31
gpu-compute: Reuse RP list in GPU_VIPER

odanrc:
It is safer to reuse the dynamic list than manually listing all possible replacement policies.

### DIFF
--- a/configs/example/apu_se.py
+++ b/configs/example/apu_se.py
@@ -50,6 +50,7 @@ from common import (
     FileSystemConfig,
     GPUTLBConfig,
     GPUTLBOptions,
+    ObjectList,
     Options,
     Simulation,
 )
@@ -393,6 +394,7 @@ parser.add_argument(
     "--tcp-rp",
     type=str,
     default="TreePLRURP",
+    choices=ObjectList.rp_list.get_names(),
     help="cache replacement policy" "policy for tcp",
 )
 
@@ -400,6 +402,7 @@ parser.add_argument(
     "--tcc-rp",
     type=str,
     default="TreePLRURP",
+    choices=ObjectList.rp_list.get_names(),
     help="cache replacement policy" "policy for tcc",
 )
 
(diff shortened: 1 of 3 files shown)


mattsinc:
@yuxiaojia I think you were also working on this ... not sure what your version looks like currently, but please give this a try and let me know what it looks like ... seems ok to me but you should test out.

abmerop:
I am guessing these should also be added here? https://github.com/gem5/gem5/blob/b1a44b89c7bae73fae2dc547bc1f871452075b85/configs/example/gpufs/runfs.py#L198-L218

#1230 was created as a follow up to #1213. As it was also missing there.  I could do this if needed...

mattsinc:
> I am guessing these should also be added here?
> 
> https://github.com/gem5/gem5/blob/b1a44b89c7bae73fae2dc547bc1f871452075b85/configs/example/gpufs/runfs.py#L198-L218
> 
> #1230 was created as a follow up to #1213. As it was also missing there. I could do this if needed...

Yes, we'll need to do both ... ideally as part of this same patch if possible.

yuxiaojia:
> @yuxiaojia I think you were also working on this ... not sure what your version looks like currently, but please give this a try and let me know what it looks like ... seems ok to me but you should test out.

I have tested on my end, and it works for me now.

mattsinc:
> > @yuxiaojia I think you were also working on this ... not sure what your version looks like currently, but please give this a try and let me know what it looks like ... seems ok to me but you should test out.
> 
> I have tested on my end, and it works for me now.

Great!  @odanrc do you want to extend per @abmerop 's request for GPUFS too, or do you need help with this?

odanrc:
> > > @yuxiaojia I think you were also working on this ... not sure what your version looks like currently, but please give this a try and let me know what it looks like ... seems ok to me but you should test out.
> > 
> > 
> > I have tested on my end, and it works for me now.
> 
> Great! @odanrc do you want to extend per @abmerop 's request for GPUFS too, or do you need help with this?

This should be trivial enough to apply. I did a couple of greps, and it seems that the latest commit should do the trick.

abmerop (approved):

mattsinc (approved):
Thanks @odanrc !

PR #1534, opened 2024-09-02
misc: Do not share the random number generator across components

aperais:
Component that require randomness should not share their randomness
source with other components to avoid simulation noise. For instance,
the branch predictor of one core should not impact the random
cache replacement policy of the cache of another core. This currently
happens as all components share a single random number generator.
    
This PR provides their own generators to relevant components, although
a couple components still use rand().
    
Change-Id: I3fb7226111c9194ee457af0f0f2b83f8c7b69d1e

### DIFF
--- a/src/base/random.cc
+++ b/src/base/random.cc
@@ -45,12 +45,6 @@
 namespace gem5
 {
 
-Random::Random()
-{
-    // default random seed
-    init(5489);
-}
-
 Random::Random(uint32_t s)
 {
     init(s);
@@ -58,6 +52,26 @@ Random::Random(uint32_t s)
 
 Random::~Random()
 {
+    assert(instances);
+
+    int removed = 0;
+
+    instances->erase(std::remove_if(instances->begin(), instances->end(),
+    [&](const auto& s_ptr)
+    {
+        removed += s_ptr.expired();
+        return s_ptr.expired();
+    }));
+
+    // Can only remove one pointer
+    // since we are destroying one
+    // object
+    assert(removed == 1);
+
+    if (instances->empty()) {
+        delete instances;
+        instances = nullptr;
+    }
 }
 
 void
@@ -66,6 +80,7 @@ Random::init(uint32_t s)
     gen.seed(s);
 }
 
-Random random_mt;
+uint64_t Random::globalSeed = 5489;
+Random::Instances* Random::instances = nullptr;
 
 } // namespace gem5
--- a/src/base/random.hh
+++ b/src/base/random.hh
@@ -48,6 +48,7 @@
 #include <random>
 #include <string>
 #include <type_traits>
+#include <vector>
 
 #include "base/compiler.hh"
 #include "base/logging.hh"
@@ -58,24 +59,90 @@ namespace gem5
 
 class Random
 {
+  friend class RandomTest;
 
   public:
+    using RandomPtr = std::shared_ptr<Random>;
+    using Instances = std::vector<std::weak_ptr<Random>>;
+
+    static RandomPtr genRandom()
+    {
+      if (!instances)
+        instances = new Instances();
+
+      auto ptr = std::shared_ptr<Random>(new Random(globalSeed));
+      instances->emplace_back(ptr);
+      return ptr;
+    }
+
+    static RandomPtr genRandom(uint32_t s)
+    {
+      if (!instances)
+        instances = new Instances();
+
+      auto ptr = std::shared_ptr<Random>(new Random(s));
+      instances->emplace_back(ptr);
+      return ptr;
+    }
+
+    static uint64_t globalSeed;
 
     /**
      * @ingroup api_base_utils
      */
     std::mt19937_64 gen;
 
+  private:
+    /**
+     * Collection of all live instances
+     * of Random to enable global
+     * reseeding. We use a pointer
+     * because the loader will initialize
+     * it to 0x0 (it is in .bss), allowing us to avoid
+     * Static Initialization Order Fiasco
+     * if static Random instances are inialized
+     * before the vector by having the constructors
+     * of Random allocate memory for the pointer.
+     * This requires that nullptr matches how
+     * the loader initializes memory
+     */
+    static_assert(nullptr == 0x0, "nullptr is not 0x0, Random instance tracking will fail");
+    static Instances* instances;
+
     /**
      * @ingroup api_base_utils
      * @{
      */
-    Random();
+    Random() = delete;
     Random(uint32_t s);
+
+    Random(const Random& rng) = delete;
+    Random& operator=(const Random& rng) = delete;
+
+    Random(Random&& rng) = delete;
+    Random& operator=(Random&& rng) = delete;
+
+  public:
     /** @} */ // end of api_base_utils
     ~Random();
 
     void init(uint32_t s);
+
+    /**
+     * Facility to reseed all live instances
+     * and ensure future default constructed
+     * instances also use the new see
+     */
+    static void reseedAll(uint64_t seed)
+    {
+        globalSeed = seed;
+
+        if (instances == nullptr)
+          return;
+
+        for (auto rng_ptr : *instances)
+            rng_ptr.lock()->init(seed);
+    }
 
     /**
      * Use the SFINAE idiom to choose an implementation based on
@@ -117,11 +184,6 @@ class Random
         return r;
     }
 };
-
-/**
- * @ingroup api_base_utils
- */
-extern Random random_mt;
 
 } // namespace gem5
 
--- a/src/base/random.test.cc
+++ b/src/base/random.test.cc
@@ -30,11 +30,38 @@
 #include <gtest/gtest.h>
 
 #include <unordered_map>
+#include <utility>
+#include <set>
+#include <algorithm>
 
 #include "base/gtest/logging.hh"
 #include "base/random.hh"
 
+namespace gem5
+{
+
+/**
+ * Helper class to access private members of
+ * Random
+ */
+class RandomTest : public ::testing::Test
+{
+  public:
+    static Random::Instances* getInstances()
+    {
+        return Random::instances;
+    }
+
+    static uint64_t getGlobalSeed()
+    {
+        return Random::globalSeed;
+    }
+};
+
+}
+
 using namespace gem5;
+using RandomPtr = gem5::Random::RandomPtr;
 
 /**
  * Checking that default construction uses the default
@@ -42,31 +69,32 @@ using namespace gem5;
  */
 TEST(RandomCtor, UInt64DefaultConstruct)
 {
+    ASSERT_EQ(RandomTest::getGlobalSeed(), std::mt19937_64::default_seed);
     // Init seed was default seed 5489
-    Random dut;
+    RandomPtr dut = Random::genRandom();
+    ASSERT_EQ(RandomTest::getGlobalSeed(), std::mt19937_64::default_seed);
     // First uint64_t corresponds to the default seed
-    ASSERT_EQ(dut.random<uint64_t>(), 14514284786278117030llu);
+    ASSERT_EQ(dut->random<uint64_t>(), 14514284786278117030llu);
 }
 
 TEST(RandomCtor, DoubleDefaultConstruct)
 {
     // Init seed was default seed 5489
-    Random dut;
+    RandomPtr dut = Random::genRandom();
     // First double corresponds to the default seed
     // 14514284786278117030llu / std::numeric_limits<uint64_t>::max()
-    ASSERT_EQ(dut.random<double>(),
+    ASSERT_EQ(dut->random<double>(),
       0.7868209548678020137657540544751100242137908935546875d);
 }
 
 TEST(RandomCtor, FloatDefaultConstruct)
 {
     // Init seed was default seed 5489
-    Random dut;
+    RandomPtr dut = Random::genRandom();
     // First float corresponds to the default seed
     // 14514284786278117030llu / std::numeric_limits<uint64_t>::max()
-    ASSERT_EQ(dut.random<float>(), 0.786820948123931884765625f);
+    ASSERT_EQ(dut->random<float>(), 0.786820948123931884765625f);
 }
-
 
 /**
  * Checking that default construction uses the default
@@ -74,8 +102,8 @@ TEST(RandomCtor, FloatDefaultConstruct)
  */
 TEST(RandomCtor, ConstructUserSpecifiedSeed)
 {
-    Random dut{42};
-    ASSERT_EQ(dut.random<uint64_t>(), 13930160852258120406llu);
+    RandomPtr dut = Random::genRandom(42);
+    ASSERT_EQ(dut->random<uint64_t>(), 13930160852258120406llu);
 }
 
 /**
@@ -84,10 +112,10 @@ TEST(RandomCtor, ConstructUserSpecifiedSeed)
  */
 TEST(RandomCtor, ConstructThenReseed)
 {
-    Random dut{};
-    ASSERT_EQ(dut.random<uint64_t>(), 14514284786278117030llu);
-    dut.init(42);
-    ASSERT_EQ(dut.random<uint64_t>(), 13930160852258120406llu);
+    RandomPtr dut = Random::genRandom();
+    ASSERT_EQ(dut->random<uint64_t>(), 14514284786278117030llu);
+    dut->init(42);
+    ASSERT_EQ(dut->random<uint64_t>(), 13930160852258120406llu);
 }
 
 /**
@@ -96,12 +124,12 @@ TEST(RandomCtor, ConstructThenReseed)
  */
 TEST(RandomRange, MinEqualsMax)
 {
-    Random dut;
+    RandomPtr dut = Random::genRandom();
 
     for (int i = 0; i < 10; i++) {
-        ASSERT_EQ(dut.random<int>(0, 0), 0);
-        ASSERT_EQ(dut.random<uint64_t>(1, 1), 1);
-        ASSERT_EQ(dut.random<int>(-1, -1), -1);
+        ASSERT_EQ(dut->random<int>(0, 0), 0);
+        ASSERT_EQ(dut->random<uint64_t>(1, 1), 1);
+        ASSERT_EQ(dut->random<int>(-1, -1), -1);
     }
 }
 
@@ -127,14 +155,14 @@ bool withinFreqRange(int count)
  */
 TEST(RandomRange, Coverage)
 {
-    Random dut;
+    RandomPtr dut = Random::genRandom();
 
     // Count occurences if we want to check
     // frequencies in the future
     std::unordered_map<int, int> values;
 
     for (int i = 0; i < loopCount; i++) {
-        values[dut.random<int>(4,6)]++;
+        values[dut->random<int>(4,6)]++;
     }
 
     ASSERT_EQ(values.count(4), 1);
@@ -147,7 +175,7 @@ TEST(RandomRange, Coverage)
     values.clear();
 
     for (int i = 0; i < loopCount; i++) {
-        values[dut.random<int>(-1,1)]++;
+        values[dut->random<int>(-1,1)]++;
     }
 
     ASSERT_EQ(values.count(-1), 1);
@@ -161,7 +189,7 @@ TEST(RandomRange, Coverage)
     values.clear();
 
     for (int i = 0; i < loopCount; i++) {
-        values[dut.random<int>(-6,-4)]++;
+        values[dut->random<int>(-6,-4)]++;
     }
 
     ASSERT_EQ(values.count(-6), 1);
@@ -173,6 +201,88 @@ TEST(RandomRange, Coverage)
     ASSERT_EQ(withinFreqRange(values[-4]), true);
 }
 
+
+
+/**
+ * Test that all constructed objects get
+ * added to the live instance list
+ */
+TEST(RandomConstruct, LiveInstancesAdd)
+{
+    RandomPtr base_rng = Random::genRandom();
+    RandomPtr my_rng = Random::genRandom();
+
+    // All pointers have been added
+    ASSERT_EQ(RandomTest::getInstances()->size(), 2);
+
+    std::set<Random*> ptrs = { base_rng.get(), my_rng.get() };
+
+    // The correct pointers have been added
+    ASSERT_EQ(std::count_if(
+        RandomTest::getInstances()->begin(),
+        RandomTest::getInstances()->end(),
+        [&](const auto & rng) { return ptrs.count(rng.lock().get()) != 0;}), 2);
+}
+
+/**
+ * Test that all destructed objects get
+ * removed from the live instance list
+ */
+TEST(RandomConstruct, LiveInstancesRemove)
+{
+    {
+        RandomPtr base_rng = Random::genRandom();
+        {
+            RandomPtr my_rng = Random::genRandom(42);
+            ASSERT_EQ(RandomTest::getInstances()->size(), 2);
+        }
+        ASSERT_EQ(RandomTest::getInstances()->size(), 1);
+    }
+    ASSERT_EQ(RandomTest::getInstances(), nullptr);
+}
+
+/**
+ * Test that reseeding after construction with
+ * an explicit seed works
+ */
+TEST(RandomReseed, ConstructThenReseed)
+{
+    RandomPtr rng = Random::genRandom();
+    ASSERT_EQ(rng->random<uint64_t>(), 14514284786278117030llu);
+    rng->init(42);
+    ASSERT_EQ(rng->random<uint64_t>(), 13930160852258120406llu);
+    ASSERT_EQ(RandomTest::getGlobalSeed(), std::mt19937_64::default_seed);
+}
+
+/**
+ * Test if global reseeding updates seed
+ * of live instances
+ */
+TEST(RandomReseed, GlobalReseedLive)
+{
+    RandomPtr base_rng  = Random::genRandom();
+    RandomPtr my_rng = Random::genRandom(1337);
+
+    ASSERT_EQ(base_rng->random<uint64_t>(), 14514284786278117030llu);
+    ASSERT_EQ(my_rng->random<uint64_t>(), 12913197394697896830llu);
+
+    Random::reseedAll(42);
+
+    ASSERT_EQ(base_rng->random<uint64_t>(), 13930160852258120406llu);
+    ASSERT_EQ(my_rng->random<uint64_t>(), 13930160852258120406llu);
+}
+
+/**
+ * Test if global reseeding updates seed
+ * of future instances
+ */
+TEST(RandomReseed, GlobalReseedFuture)
+{
+    Random::reseedAll(42);
+    RandomPtr base_rng = Random::genRandom();
+    ASSERT_EQ(base_rng->random<uint64_t>(), 13930160852258120406llu);
+}
+
 /** Test that the range provided for random
  *  number generation is valid
  */
@@ -182,6 +292,6 @@ TEST(RandomDeathTest, InvalidRange)
     GTEST_SKIP() << "Skipping as assertions are "
         "stripped out of fast builds";
 #endif
-    Random dut;
-    ASSERT_DEATH(dut.random<int>(4, 2), "");
+    RandomPtr dut = Random::genRandom();
+    ASSERT_DEATH(dut->random<int>(4, 2), "");
 }
--- a/src/cpu/pred/multiperspective_perceptron_tage.hh
+++ b/src/cpu/pred/multiperspective_perceptron_tage.hh
@@ -51,6 +51,7 @@
 #ifndef __CPU_PRED_MULTIPERSPECTIVE_PERCEPTRON_TAGE_HH__
 #define __CPU_PRED_MULTIPERSPECTIVE_PERCEPTRON_TAGE_HH__
 
+#include "base/random.hh"
 #include "cpu/pred/loop_predictor.hh"
 #include "cpu/pred/multiperspective_perceptron.hh"
 #include "cpu/pred/statistical_corrector.hh"
@@ -69,6 +70,9 @@ namespace branch_prediction
 class MPP_TAGE : public TAGEBase
 {
     std::vector<unsigned int> tunedHistoryLengths;
+
+    Random::RandomPtr rng = Random::genRandom();
+
   public:
     struct BranchInfo : public TAGEBase::BranchInfo
     {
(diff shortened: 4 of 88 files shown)


src/sim/syscall_emul.hh
```diff
@@ -119,6 +119,11 @@
 namespace gem5
 {
 
+namespace
+{
+static Random se_prng = {};
```
giactra: static is unnecessary in an anonymous namespace (which is meant to replace static)

src/cpu/pred/simple_indirect.cc
```diff
@@ -342,7 +342,7 @@ SimpleIndirectPredictor::recordTarget(ThreadID tid, InstSeqNum sn,
             sn, history->pcAddr, history->set_index, target);
 
     // Did not find entry, random replacement
-    auto &way = iset[rand() % numWays];
```
giactra: This does not uses random_mt and IMHO it deserves a separate commit

src/mem/ruby/network/garnet/RoutingUnit.cc
```diff
@@ -138,7 +138,7 @@ RoutingUnit::lookupRoutingTable(int vnet, NetDest msg_destination)
     // Randomly select any candidate output link
     int candidate = 0;
     if (!(m_router->get_net_ptr())->isVNetOrdered(vnet))
-        candidate = rand() % num_candidates;
```
giactra: Ditto

src/python/pybind11/core.cc
```diff
@@ -310,7 +310,7 @@ pybind_init_core(py::module_ &m_native)
         .def("disableAllListeners", &ListenSocket::disableAll)
         .def("listenersDisabled", &ListenSocket::allDisabled)
         .def("listenersLoopbackOnly", &ListenSocket::loopbackOnly)
-        .def("seedRandom", [](uint64_t seed) { random_mt.init(seed); })
+        //.def("seedRandom", [](uint64_t seed) { random_mt.init(seed); })
```
giactra: This is a problem. As we don't have a single random generator anymore, we need to provide a way to seed all independent generators now. My suggestion is to keep the seedRandom (have a single seeder). Only it will store the seed in a global variable that will be used by the Random object (constructor) to init itself.

giactra (changes requested):
Thanks @aperais for this PR. I understand the problem; however I am not very keen on adding a new base class (like UsesRandom) to simply store an object. As we are not interested on defining a new interface, I would prefer [composition over inherithance](https://en.wikipedia.org/wiki/Composition_over_inheritance). In other words I would just add a Random object to every rng client.

I also have an important comment over seeding (see individual comment)

Let me know what you think about this

giactra:
Another comment: when you have so many subsystem being touched, it is fine to used the "misc" tag in the commit title instead of listing all of them ;)

src/python/pybind11/core.cc
aperais: I think I forgot to mention that I commented this for now in the PR but yes this sounded like an issue when I did it. Assuming I understood your suggestion, isn't there an assumption that seedRandom would be called before all Random objects are created ? Is that going to happen ?

If the goal is to have something callable from Python to reseed the random generator at any given time, we could keep a static (to the Random class) container of all existing Random objects and walk the container to reseed all of them on call to seedRandom.

src/cpu/pred/simple_indirect.cc
aperais: I can do that. I modified this here because it was essentially the same problem (simple_indirect.cc is sharing state with RoutingUnit.cc).

src/sim/syscall_emul.hh
aperais: Good catch, thanks !

src/python/pybind11/core.cc
giactra: > I think I forgot to mention that I commented this for now in the PR but yes this sounded like an issue when I did it. Assuming I understood your suggestion, isn't there an assumption that seedRandom would be called before all Random objects are created ? Is that going to happen ?
> 

This is a good point. With SimObjects this assumption holds as they are dynamically created after m5.instantiate, but since the random generator can be used by any object we cannot rely on this assumption being true in all circumstances

> If the goal is to have something callable from Python to reseed the random generator at any given time, we could keep a static (to the Random class) container of all existing Random objects and walk the container to reseed all of them on call to seedRandom.

Yep, that's a good solution!

src/cpu/pred/simple_indirect.cc
aperais: @giactra Done

src/mem/ruby/network/garnet/RoutingUnit.cc
aperais: @giactra Done

src/python/pybind11/core.cc
aperais: @giactra Done, but I do not have a good idea of how to test it. Do you have a recommendation for an existing test that could call seedRandom?

src/python/pybind11/core.cc
```diff
@@ -310,7 +310,10 @@ pybind_init_core(py::module_ &m_native)
         .def("disableAllListeners", &ListenSocket::disableAll)
         .def("listenersDisabled", &ListenSocket::allDisabled)
         .def("listenersLoopbackOnly", &ListenSocket::loopbackOnly)
-        .def("seedRandom", [](uint64_t seed) { random_mt.init(seed); })
+        .def("seedRandom", [](uint64_t seed) {
+            for (auto rng_ptr : Random::instances)
+                rng_ptr->init(seed);
```
giactra: This I believe will only catch cases where the Random is constructed before the seedRandom call. What happens if we call seedRandom before constructing a Random object? This might happen for example with SimObjects.

I believe the logic should be something like

         .def("seedRandom", [](uint64_t seed) {
            globalSeed = seed;
            // Reseed previously constructed instances
            for (auto rng_ptr : Random::instances)
                rng_ptr->init(seed);
          }

With globalSeed defined instead of mt_random as

`uint64_t globalSeed = 5489;`

And the Random constructor being:
```

Random::Random()
{
    init(globalSeed);
}
```

giactra (changes requested):

src/sim/syscall_emul.hh
aperais: Done

src/python/pybind11/core.cc
giactra: Good question. I believe you should take a simple config.

Then call

`m5.seedRandom(first_value)`

before m5.instantiate

Then call again 

`m5.seedRandom(second_value)`

And start simulation with m5.simulate

After simulation (m5.simulate), you should check for all Random objects to make sure they have the second seed. Now it looks like std::mt19937_64 does not have a getter to retrieve the seed (only a setter). As a workaround we could store the seed in the Random object after seeding has happened so we can retrieve it from there.

It's a bit of a hack and it requires adding debug only logic so let me know if you have better options in mind

src/python/pybind11/core.cc
giactra: I correct myself; rather than a global variable globalSeed could be a static member of Random

src/base/random.hh
```diff
@@ -62,12 +62,20 @@ class Random : public Serializable
 {
 
   public:
-
     /**
      * @ingroup api_base_utils
      */
     std::mt19937_64 gen;
 
+    static inline uint64_t globalSeed = 5489;
```
giactra: Nit: you don't need the inline kw

src/base/random.hh
```diff
@@ -62,12 +62,20 @@ class Random : public Serializable
...
+
+    /**
+     * Collection of all live instances
+     * of Random to enable global
+     * reseeding
+     */
+    static inline std::vector<Random*> instances = {};
```
giactra: Ditto

src/cpu/pred/multiperspective_perceptron_tage.hh
```diff
@@ -104,6 +108,8 @@ class MPP_TAGE : public TAGEBase
 
 class MPP_LoopPredictor : public LoopPredictor
 {
+
+
```
giactra: Nit: remove newlines

src/mem/cache/replacement_policies/bip_rp.hh
```diff
@@ -75,6 +78,7 @@ class BIP : public LRU
      */
     void reset(const std::shared_ptr<ReplacementData>& replacement_data) const
                                                                      override;
+
```
giactra: Nit: remove newline

giactra (commented):
Thanks @aperais, I have some minor comments and then if you tell me this has been tested correctly we are ready to merge this

powerjg:
Would it be possible to add some tests (unittests?) for this change in the behavior? I support the idea of having a precise behavior for random, but to ensure that we have implemented our goals, it would be useful to have some tests.

src/python/pybind11/core.cc
aperais: @giactra So I think ensuring all Random have the correct seed can be done with unittests rather than an actual simulation, unless I am missing some aspect.

I can add a static function of `Random`, let's call it `reseedAll(seed)`, that is what would be called from Python `seedRandom(seed)`. 

**Existing `Random` instances:** The expectation is that all `Random` objects will have the same seed after the call to `reseedAll(seed)` , hence will produce the same first value, which we can hardcode and check against since `std::mt19937_64` is standard. The only quirk with this one is that several seeds may produce the same first random value (I don't know if this is possible with the Mersenne Twister RNG), so the test coverage would not be 100%. The alternative would be to have a debug only `seed` member in `Random` as you suggested.

**Future `Random` instances:** The unit test could also check that after calling `reseedAll(seed)`, all future instances of Random are initialized with the seed passed by that call, by checking what number they provide first or a debug only member (as above).

What do you think ? This would also address @powerjg that we need some unittests for it, although those unittests would not look at interactions between components of which `Random` is a member (but I am not sure how to check this other than having your private or protected `Random` instance)

src/base/random.hh
aperais: With gcc 10.2.1 it complains if I do not have it. : `ISO C++ forbids in-class initialization of non-const static member ‘gem5::Random::globalSeed’`

src/base/random.hh
aperais: Ditto :) : `temporary of non-literal type ‘std::vector<gem5::Random*>’ in a constant expression`

src/base/random.hh
powerjg: We don't support gcc 10: https://github.com/gem5/gem5/blob/stable/RELEASE-NOTES.md#compiler-and-os-support

giactra: > With gcc 10.2.1 it complains if I do not have it. : ISO C++ forbids in-class initialization of non-const static member ‘gem5::Random::globalSeed’

Even if we don't support gcc10, I wouldn't disregard this. And it makes sense I forgot this only works for const variables. As inline is simply saying the compiler to not enforce the ODR, my fear is every translation units ends up with a different version of globalSeed. So the solution is to define/initialize in source file (random.cc). I think that is the safest option that makes everyone happy

aperais: I can go back to the old way but I thought C++17 (which is the standard used to compile) introduced static inline variables notably for this purpose.

From https://en.cppreference.com/w/cpp/language/inline, it suggests that since C++17, an inline variable without external linkage (i.e., `static inline`) may not have several definitions (in different translation units), because it may have several definitions only if it has external linkage (i.e., not `static`)

I'll roll back to keep it safe though.

aperais: Also, not to nitpick but the link to the release notes shows `As of this release gem5 support Clang version 6 to 16 and GCC version **10** to 13` for gem5 24.0.0. Maybe the plan is to drop it for the next version though (?).

powerjg: Oh, thanks for pointing that out! That was my confusion 🙂. I get all sorts of confused between clang and gcc versions anymore...

src/cpu/pred/multiperspective_perceptron_tage.hh
aperais: Done

src/python/pybind11/core.cc
aperais: I implemented the unit tests, but I had to break down `Random` as it was pulling a bunch of files from `/sim` because it needed `Serializable`. It's a bit ugly, I don't know if I should include all needed files from `/sim` in the SConscript or `/base`, or if we should give up on `Random` being `Serializable` (it's currently unimplemented anyway), or if having another class just for that purpose works.

src/base/random.cc
```diff
@@ -50,52 +50,45 @@ namespace gem5
...
-    if (optParamIn(cp, "mt_state", state)) {
-        std::istringstream iss(state);
-        iss >> gen;
-    }
+    gen.seed(s);
 }
 
-Random random_mt;
+uint64_t Random::globalSeed = 5489;
+std::vector<Random*> Random::instances = {};
```
powerjg: Could this be `std::vector<unique_ptr<Random>>`?

I *think* this would work... or, at least, I know that this is the intention of the change. It would be nice if we could enforce that the pointer doesn't leak out anywhere else and we only pass around references to the object (not copies, which would break the semantics, I think).

powerjg (commented):
Sorry I didn't review this sooner. Overall, the change makes sense to me, just one small thing, for which I could be mistaken.

src/base/random.cc
aperais: I see the idea, but I think it's tricky to achieve. Right now, the way to use  `Random` is like this:

```
class A
{
  Random rng = {};
}
```

So the storage for the `Random` object is allocated when instance of `A` is created and freed when instance of `A` is deleted. If you store a `unique_ptr<Random>` in the `instances` vector, I guess you are opening yourself to a double free because the storage for `Random` will be freed when `instances` is deleted (at exit I suppose) and when the instance of `A` gets deleted. Am I getting this right?

The only way I can imagine unique_ptr management working is to disallow having a `Random` member and force to have a `Random*` member but I don't think that's enforceable statically and you would get the double free anytime someone would have a `Random` member.

I don't have much experience with it, but unique_ptr takes a `Deleter` template parameter. If we provide a Deleter that does nothing, then storage will be deleted by the destructor of `A` and `instances` going away will not do anything?

giactra: IMHO we shouldn't be relying on unique_ptrs. They are used to model exclusive ownership of the pointer holder towards the pointed data. Here the static vector does not own the random generators which are instead owned by the objects instantiating them. The vector just stores a non-owning pointer so that it can pass them the new seed once it gets generated,

Also as @aperais pointed out, we cannot make any assumption about the lifetime of the Random object.
You can literally instantiate the Random object on the stack if you wanted to and to destroy it automatically when it goes out of scope. This is not compatible with unique_ptr lifetime management

powerjg: Thanks for the info. This makes sense to me.

I do wonder if it is a bad code smell to "instantiate the Random object on the stack if you wanted to and to destroy it automatically when it goes out of scope." It seems that it may be a good idea to require things that use randomness in gem5 to follow a more restricted model. But I'm happy to say that's out of scope.

giactra: > Thanks for the info. This makes sense to me.
> 
> I do wonder if it is a bad code smell to "instantiate the Random object on the stack if you wanted to and to destroy it automatically when it goes out of scope." It seems that it may be a good idea to require things that use randomness in gem5 to follow a more restricted model. But I'm happy to say that's out of scope.

Agreed and it was just an hypothetical example. @aperais gives a better one which is actually how the random generator is mainly used throughout the code

> class A
> {
>   Random rng = {};
> }
> 
> So the storage for the Random object is allocated when instance of A is created and freed when instance of A is deleted. If you store a unique_ptr<Random> in the instances vector, I guess you are opening yourself to a double free because the storage for Random will be freed when instances is deleted (at exit I suppose) and when the instance of A gets deleted

src/base/random.hh
```diff
@@ -58,22 +58,47 @@ namespace gem5
 
 class Checkpoint;
 
-class Random : public Serializable
```
giactra: I believe making Random not serializable makes sense but I would add a separate commit for this

giactra (commented):
Thanks @aperais for your changes. I am wondering what would be the performance impact of adding a mutex to guard the instances vector/global seed. At the moment we are mainly single threaded but I don't want to make the life impossible for the kind soul willing to parallelize gem5. I am sure lots of gem5 needs to be sanitized for that but maybe it is better to do things properly when we have the chance

src/base/random.test.cc
```diff
@@ -0,0 +1,217 @@
...
+ */
+TEST(RandomReseed, ConstructThenReseed)
+{
+#ifdef NDEBUG
+    GTEST_SKIP() << "Skipping as assertions are "
+        "stripped out of fast builds";
+#endif
+
+    Random rng{};
+    ASSERT_EQ(rng.random<uint64_t>(), 14514284786278117030llu);
```
giactra: Dumb question: are we sure the first value is not host dependent somehow?

aperais: Not dumb at all : it is dependent !

Actually I should have known given I started the discussion here : https://github.com/orgs/gem5/discussions/1517

It appears for the test to be compiler independent (STL implementation independent), either the test should not check for the value, so check the seed and carry the seed as a debug member, or I should first propose a PR that makes calls to `Random::random<>` compiler independent.

src/base/random.hh
aperais: I will create a PR doing that to enable this PR then.

src/base/random.hh
aperais: I'm answering this comment because actually tried the old style and it fails some tests, like :

`/fast-scratch/peraisa/gem5_fix/build/VEGA_X86/gem5.debug -d /tmp/gem5outrefaxyp2 -re --silent-redirect /fast-scratch/peraisa/gem5_fix/configs/example/apu_se.py --download-resource square-gpu-test --download-dir /fast-scratch/peraisa/gem5_fix/tests/gem5/resources --reg-alloc-policy=dynamic -n3 -c /fast-scratch/peraisa/gem5_fix/tests/gem5/resources/square-gpu-test`

I'm debugging it but the problem is the instance of `Random` in an anonymous namespace in `syscall_emul.hh` (`se_prng`). The static initialization of the libc initaliazes multiples objects through static initialization (which is already strange), and deletes one object two times, so the assert where it should be in the `instances` vector if it is deleted fires.

One way to make it work is to make `instances` a `static inline` member, but it's not obvious why changing `instances` is going to fix the behavior of multiples instances of `se_prng` being instantiated, unless there are also multiple versions of `instances` being created. I have seen this kind of mixups in the past where several dynamic libraries have their own version of a static variable (`static inline` is also supposed to solve that...I think...At least we used it for that purpose). I'll update once I get a clearer picture.

aperais: My ongoing best guess is this is a link order issue. The problem is that the `Random` instance in `syscall_emul.hh` is initialized before `instances` is initalized, so the pointer gets pushed in a unitialized vector, which gets constructed later and this causes issues. See the image below which corresponds to `se_prng` being constructed (see the size of the vector).

image

I suspect using `static inline` fixes the behavior by making `instances` construct first, so assuming this is a link order issues, the solutions are `static inline` (hypothesis to be verified), or fixing the link order. The latter is quite brittle because the compiler cannot really tell this is going to happen, and people may get surprises in the future.

@giactra @powerjg Any thoughts ?

powerjg: `static inline` is fine with me. You may want to include a pointer to this PR in a comment for the future. Or, we can rely on git blame :)

powerjg:
What's the status of this PR?

aperais:
It needs to be rebased on the recent changes #1580 and #1567. I'll probably get to it next week.

aperais:
@powerjg @giactra PR is mostly good to go but I would like your opinion about the following.

With the previous implementation where Random had a static std::vector<Random*> member called `instances`, we ran into the [Static Initialization Order Fiasco](https://en.cppreference.com/w/cpp/language/siof) because the Random instance in syscall_emul.hh called `se_prng` declared and defined in an anonymous namespace got initialized before the vector, so it got pushed back into an uninitialized vector and that eventually messed things up.

I found several ways to address this, none of which seemed fully satisfactory :

1. Move declaration and definition of `se_prng` inside the body of the only function that uses it. Since there are no other static declaration of Random objects (or in anonymous namespaces), that fixes it. **Pro: Simple Cons: Anyone deciding they want a static Random object outside of a function body will run into the issue in the future**
2. Make `Random::instances` `static inline` instead of just `static`. This causes it to be initialized before "regular" `static Random` objects. **Pro: Simple Cons: Not sure the construction is guaranteed by the standard of I'm just being lucky**
3. Make `Random::instances` a pointer to a vector and initialize it to nullptr. On platforms where nullptr is 0x0, this will go to .bss, which, on Linux and Windows, is zero-initialized. This allows the constructors of `Random` to check for nullptr and it so, to allocate memory for the vector. It's a bit heavy handed and seems to be platform dependent (but is there anyone running gem5 outside of Linux/Windows/MacOS?) **Pro: Should prevent future issues, and deals with destruction order as well. Cons: More involved.** => This is what is in the PR
4. Encapsulate `static std::vector<Random*> instances` in a function so it gets created the first time the function is called. **Pro: Relatively clean Cons: I think it does not deal with destruction order, unless we use a pointer to vector and not a vector**

giactra:
I see another option (which is similar to 4):

We mark the Random constructor private and we allow Random object construction only via a static factory method which returns a shared_ptr<Random>. This prevents from someone declaring a new Random in an anonymous namespace

Something like

```
class Random
{
    public:
      using RadomPtr = std::shared_ptr<Random>;
      using Instances = std::vector<RandomPtr>*;

      static RandomPtr
      genRandom(...)
      {
              if (!instances)
                  instances = new Instances();

              auto random = std::make_shared<Random>(...);
              instances->push_back(random);
              return random;
      }

    private:
      Random(...); 
      static Instances instances;
}
```

Ia there anything I am missing?

aperais:
The only thing I see is the deletion of `instances`, but doing it the same way I do it in the destructor would handle that.

aperais:
Well, I tried going this way and I think there is a problem if the objective is to release all the memory we allocate.

Assume I create a Random object using `genRandom` and assign it somewhere. This looks like :

```
struct NeedsRandomObject
{
  RandomPtr my_random = Random::genRandom();
};

void foo()
{
  NeedsRandomObject component{}; // Adds a shared_ptr<Random> to Random::instances
  // Do something
 return; 
 // component is destroyed, but `my_random` is still pointed to in Random::instances by a shared_ptr, 
 // so its destructor is not called, and it's going to stay in Random::instances forever.
}
```

One - questionable - way to handle this would be to have `genRandom()` return a unique_ptr<Random> and push the raw pointer in `Random::instances`. This way, when `component` goes out of scope, the destructor of `my_random` is called, which just has to remove the raw pointer from `Random::instances`. 

Alternatively, we can do it this way and leak memory until the end of simulation. It is unlikely to be a problem because `Random` objects are supposed to be tied to long lifetime components (not going to be created as a temporary in a function), but this is not very satisfying.

giactra:
> Well, I tried going this way and I think there is a problem if the objective is to release all the memory we allocate.
> 
> Assume I create a Random object using `genRandom` and assign it somewhere. This looks like :
> 
> ```
> struct NeedsRandomObject
> {
>   RandomPtr my_random = Random::genRandom();
> };
> 
> void foo()
> {
>   NeedsRandomObject component{}; // Adds a shared_ptr<Random> to Random::instances
>   // Do something
>  return; 
>  // component is destroyed, but `my_random` is still pointed to in Random::instances by a shared_ptr, 
>  // so its destructor is not called, and it's going to stay in Random::instances forever.
> }
> ```
> 
> One - questionable - way to handle this would be to have `genRandom()` return a unique_ptr and push the raw pointer in `Random::instances`. This way, when `component` goes out of scope, the destructor of `my_random` is called, which just has to remove the raw pointer from `Random::instances`.

I think using raw ptr + unique_ptr defeats the purpose of using smart pointers (having said that we do it sometimes in gem5 and I have my fair share of blame).
I think your algorithm plays well with weak_ptrs

Basically we push weak_ptrs in the instances vector and genRandom returns the owning pointer:


      static RandomPtr
      genRandom(...)
      {
              if (!instances)
                  instances = new Instances();

              std::weak_ptr<Random> instance(new Random(...));
              instances->push_back(instance);
              return instance.lock();
      }


> 
> Alternatively, we can do it this way and leak memory until the end of simulation. It is unlikely to be a problem because `Random` objects are supposed to be tied to long lifetime components (not going to be created as a temporary in a function), but this is not very satisfying.

Yeah and as an example we basically don't call the destructors of SimObjects, so every Random instance in a SimObject won't be deallocated until process termination. Anyway I believe we should do things properly if we can (and I think we can)

aperais:
@giactra The way to use `weak_ptr` is slightly more tricky than I was expecting because they expire when the last `shared_ptr` goes out of scope, so in the destructor I cannot compare raw addresses anymore. I get away with it by removing the only expired `weak_ptr` from `Random::instances`.

This seems to work, the only suboptimality is that the Random instance is not inline with its encapsulating class in memory but at least there should not be any issue with initialization of static variables.

aperais:
Btw, the test that fails in CI passes on my machine (using latest develop branch). Is there a way to get the output file of CI ?

Test is `suite-riscv-vertical-microbenchmarks_1.0.0_riscv_suite_test_to-tick-ALL-x86_64-opt`, command `./main.py run -vv -j$(nproc) gem5/suite_tests` (with --skip-build if relevant)

Edit: problem seems to have solved itself. @powerjg @giactra I think this is good to go.

aperais:
@giactra Gentle reminder that this is ready (at least I think so :)). Unless something else has to get in first ?

erin-le:
@giactra Hello, could you take another look at this PR when you get a chance? Thank you!

src/base/random.hh
```diff
@@ -58,25 +59,91 @@ namespace gem5
...
+
+    static RandomPtr genRandom()
+    {
+      if (!instances)
+        instances = new Instances();
+
+      auto ptr = std::shared_ptr<Random>(new Random(globalSeed));
+      instances->emplace_back(ptr);
+      return ptr;
+    }
```
giactra, line 76: Nit: I would move this to the source file if you don't mind. Also there should be 4 whitespaces indentation between braces and code

src/base/random.hh
```diff
@@ -58,25 +59,91 @@ namespace gem5
...
+
+    static RandomPtr genRandom(uint32_t s)
+    {
+      if (!instances)
+        instances = new Instances();
+
+      auto ptr = std::shared_ptr<Random>(new Random(s));
+      instances->emplace_back(ptr);
+      return ptr;
+    }
```
giactra, line 86: Ditto. Can't we keep only one version and have a default value for the method param?

    static RandomPtr genRandom(uint32_t seed = globalSeed)

src/base/random.hh
```diff
@@ -58,25 +59,91 @@ namespace gem5
...
+     * because the loader will initialize
+     * it to 0x0 (it is in .bss), allowing us to avoid
+     * Static Initialization Order Fiasco
+     * if static Random instances are inialized
+     * before the vector by having the constructors
+     * of Random allocate memory for the pointer.
+     * This requires that nullptr matches how
+     * the loader initializes memory
+     */
+    static_assert(nullptr == 0x0, "nullptr is not 0x0, Random instance tracking will fail");
```
giactra, line 109: What is this static assertion? I don't understand it. Wouldn't this always evaluate to true? (at compile time)

src/base/random.hh
```diff
@@ -58,25 +59,91 @@ namespace gem5
...
+     */
+    static void reseedAll(uint64_t seed)
+    {
+        globalSeed = seed;
+
+        if (instances == nullptr)
+          return;
+
+        for (auto rng_ptr : *instances)
+            rng_ptr.lock()->init(seed);
```
giactra, line 144: Ditto

BobbyRBruce:
@giactra could you re-review this when you have a moment? I think this is a Good contribution to get into v24.1 as long as you take no issue to it

giactra:
> @giactra could you re-review this when you have a moment? I think this is a Good contribution to get into v24.1 as long as you take no issue to it

Hi @BobbyRBruce I was actually about to comment on it.

I feel a bit guilty with @aperais as he spent lots of time iterating over my suggestion. And now I start to second guess it. What concerns me most of all is the performance impact of requiring every Random object to be allocated with a shared_ptr. While this contribution might be trivial in a long FS simulation, it could become non negligible for short SE runs.

So I would probably happier with solution number 3:

> Make Random::instances a pointer to a vector and initialize it to nullptr. On platforms where nullptr is 0x0, this will go to .bss, which, on Linux and Windows, is zero-initialized. This allows the constructors of Random to check for nullptr and it so, to allocate memory for the vector. It's a bit heavy handed and seems to be platform dependent (but is there anyone running gem5 outside of Linux/Windows/MacOS?) Pro: Should prevent future issues, and deals with destruction order as well. Cons: More involved. => This is what is in the PR

This could be a temporary solution... Maybe in the long term we should adopt a RandomDomain abstraction and require a RandomObject to be constructed with a reference to a RandomDomain (similar to what we do with ClockDomains). In this way we can also reseed individual domains and leave other domains unchanged.

Please @aperais let me know your thoughts and apologies again

powerjg:
I believe we should move forward with merging this PR. We've been asking a lot from @aperais, and while the feedback provided has been valuable, the additional work required does not seem to justify the marginal benefits. The core issue has been addressed effectively, and it's time to integrate these changes.

Personally, I would have chosen option 1 above (change the **1** place that needed to be changed to make it local to all functions).

> Maybe in the long term we should adopt a RandomDomain abstraction and require a RandomObject to be constructed with a reference to a RandomDomain (similar to what we do with ClockDomains). In this way we can also reseed individual domains and leave other domains unchanged.

This sounds like a significant amount of over engineering to me. I don't see why the simplest solution isn't the one we should take.

Please let me know if I'm missing something here. I'm afraid that this change is going to be like many others that get stuck in an optimization loop and will never be merged, which would be the worst possible outcome.

aperais:
@giactra I understand your wish to avoid shared_ptr, since gem5 speed is probably hindered by "death by a thousand cuts" rather than one big costly thing, so it's definitely legitimate.

However I kind of side with @powerjg regarding building even more abstraction for those Random objects, which I think boils down to "You ain't gonna need it", since either my initial proposal 3 or the current version of 4 with weak_ptr does it.

Since the ISCA deadline is coming up, I honestly don't have a lot of time to dedicate to this, even if to me this is a major flaw of gem5 in general (my opinion). Would you be ok with merging this one and, if there is a clear path to removing the shared_ptr to avoid the cost, open an issue and tag it "First contribution" to fix it ?

giactra (approved):

PR #1535, opened 2024-09-03
python: Ignore *args and **kwargs when generating cxxMethod pybinding script

handsomeliu-google:
When generating a `@cxxMethod` decorated Python method, a method

```python
  @cxxMethod 
  def f(arg1, arg2):
    pass
```
will be generated into

```cpp
  ...
  .def("f", &f, py::arg("arg1"), py::arg("arg2"))
  ...
```

This align with the general pybind usage.

However, for function signatures including `*args` and `**kwargs`, these shouldn't be in the pybinding `def` arguments. ([reference](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#accepting-args-and-kwargs))

This commit attempts to fix this and thus to support *args and **kwargs.

### DIFF
--- a/src/python/m5/SimObject.py
+++ b/src/python/m5/SimObject.py
@@ -489,6 +489,12 @@ def cxxMethod(*args, **kwargs):
                 # We don't cound 'self' as an argument in this case.
                 continue
             param = sig.parameters[param_name]
+            if param.kind in [
+                inspect.Parameter.VAR_POSITIONAL,
+                inspect.Parameter.VAR_KEYWORD,
+            ]:
+                # *args and **kwargs shouldn't be in generated parameters
+                continue
             if param.default is param.empty:
                 args.append(param_name)
             else:


src/python/m5/SimObject.py
```diff
@@ -489,6 +489,12 @@ def decorate(func):
                 # We don't cound 'self' as an argument in this case.
                 continue
             param = sig.parameters[param_name]
+            if param.kind in [
+                inspect._ParameterKind.VAR_POSITIONAL,
```
andysan: According to the documentation, this should be `inspect.Parameter.VAR_POSITIONAL`.

andysan (changes requested):
This looks like a useful change. The way the `inspect` module is used here is relying on an undocumented interface, but the change looks good other than that.

src/python/m5/SimObject.py
handsomeliu-google: Updated the commit. Thanks for providing the correct way to achieve this!

andysan (approved):

handsomeliu-google:
@ivanaamit It seems like the CI process is taking a while. It's my first time submitting a pull request, so I'm wondering whether this is normal, or is there something I should be aware of? Thanks!

ivanaamit:
Do you mean the following?

Currently, gem5 will do the following:
```
  @cxxMethod 
  def f(arg1, arg2,  *args, **kwargs):
    pass
```

gem5 will generate:
```
  .def("f", &f, py::arg("arg1"), py::arg("arg2"), py::arg("*args"), py::arg("**kwargs"))
```

And this is wrong.

If so, please explain how your change solves this problem. What will now happen when someone writes `def f(arg1, arg2,  *args, **kwargs)`?

handsomeliu-google:
The generated binding script will become

`.def("f", &f, py::arg("arg1"), py::arg("arg2"))`, which aligns with the pybind documentation.

powerjg:
@handsomeliu-google (Apologies, I asked @ivanaamit to write the message above and it is my confusion)

I am still unclear as to what is the *correct* result and what is the *incorrect* result. Can you please describe what the problem is that you're solving and what the correct output should be?

powerjg:
This is a guess, but I think a better description of your PR would be the following:

According to the [pybind documentation](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#accepting-args-and-kwargs), "When combining *args or **kwargs with [Keyword arguments](https://pybind11.readthedocs.io/en/stable/basics.html#keyword-args) you should not include py::arg tags for the py::args and py::kwargs arguments."

In the current implementation of gem5, if you use the `cxxMethod` decorator on a function that has `*args` or `**kwargs`, gem5 will incorrectly add these variables to the pybind generated declaration.

I.e., `def f(arg1, arg2,  *args, **kwargs):` -> `.def("f", &f, py::arg("arg1"), py::arg("arg2"), py::arg("*args"), py::arg("**kwargs"))` which is incorrect pybind code.

To fix this problem, we should ignore variables in the generator if they are `*args` or `**kwargs`. This change skips these variables when creating the pybind declaration.

handsomeliu-google:
Thanks @powerjg and @ivanaamit for the clarification and suggestion!

The description from @powerjg is so informative. I'll just copy them to this PR. Thanks a lot 😄

powerjg (approved):
Thanks for the clarification. This change makes sense now

PR #1537, opened 2024-09-04
Implement BTB using the cache library

pranith:
This enables the BTB to be associative and use various replacement policies.

### DIFF
--- a/configs/common/cores/arm/HPI.py
+++ b/configs/common/cores/arm/HPI.py
@@ -1683,6 +1683,15 @@ class HPI_MMU(ArmMMU):
 class HPI_BTB(SimpleBTB):
     numEntries = 128
     tagBits = 18
+    associativity = 1
+    instShiftAmt = 2
+    btbReplPolicy = LRURP()
+    btbIndexingPolicy = BTBSetAssociative(
+        num_entries=Parent.numEntries,
+        set_shift=Parent.instShiftAmt,
+        assoc=Parent.associativity,
+        tag_bits=Parent.tagBits,
+    )
 
 
 class HPI_BP(TournamentBP):
--- a/src/cpu/pred/BranchPredictor.py
+++ b/src/cpu/pred/BranchPredictor.py
@@ -38,6 +38,8 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 from m5.objects.ClockedObject import ClockedObject
+from m5.objects.IndexingPolicies import *
+from m5.objects.ReplacementPolicies import *
 from m5.params import *
 from m5.proxy import *
 from m5.SimObject import *
@@ -83,6 +85,38 @@ class BranchTargetBuffer(ClockedObject):
     numThreads = Param.Unsigned(Parent.numThreads, "Number of threads")
 
 
+class BTBIndexingPolicy(SimObject):
+    type = "BTBIndexingPolicy"
+    abstract = True
+    cxx_class = "gem5::IndexingPolicyTemplate<gem5::BTBTagType>"
+    cxx_header = "cpu/pred/btb_entry.hh"
+    cxx_template_params = ["class Types"]
+
+    # Get the associativity
+    assoc = Param.Int(Parent.assoc, "associativity")
+
+
+class BTBSetAssociative(BTBIndexingPolicy):
+    type = "BTBSetAssociative"
+    cxx_class = "gem5::BTBSetAssociative"
+    cxx_header = "cpu/pred/btb_entry.hh"
+
+    # Get the number of entries in the BTB from the parent
+    num_entries = Param.Unsigned(
+        Parent.numEntries, "Number of entries in the BTB"
+    )
+
+    # Set shift for the index. Ignore lower 2 bits for a 4 byte instruction.
+    set_shift = Param.Unsigned(2, "Number of bits to shift PC to get index")
+
+    # Total number of bits in the tag.
+    # This is above the index and offset bit
+    tag_bits = Param.Unsigned(64, "number of bits in the tag")
+
+    # Number of threads sharing the BTB
+    numThreads = Param.Unsigned(Parent.numThreads, "Number of threads")
+
+
 class SimpleBTB(BranchTargetBuffer):
     type = "SimpleBTB"
     cxx_class = "gem5::branch_prediction::SimpleBTB"
@@ -92,6 +126,19 @@ class SimpleBTB(BranchTargetBuffer):
     tagBits = Param.Unsigned(16, "Size of the BTB tags, in bits")
     instShiftAmt = Param.Unsigned(
         Parent.instShiftAmt, "Number of bits to shift instructions by"
+    )
+    associativity = Param.Unsigned(1, "BTB associativity")
+    btbReplPolicy = Param.BaseReplacementPolicy(
+        LRURP(), "BTB replacement policy"
+    )
+    btbIndexingPolicy = Param.BTBIndexingPolicy(
+        BTBSetAssociative(
+            assoc=Parent.associativity,
+            num_entries=Parent.numEntries,
+            set_shift=Parent.instShiftAmt,
+            numThreads=1,
+        ),
+        "BTB indexing policy",
     )
 
 
--- a/src/cpu/pred/btb_entry.hh
+++ b/src/cpu/pred/btb_entry.hh
@@ -1,0 +1,288 @@
+/*
+ * Copyright (c) 2024 Pranith Kumar
+ * All rights reserved.
+ *
+ * The license below extends only to copyright in the software and shall
+ * not be construed as granting a license to any other intellectual
+ * property including but not limited to intellectual property relating
+ * to a hardware implementation of the functionality of the software
+ * licensed hereunder.  You may use the software subject to the license
+ * terms below provided that you ensure that this notice is replicated
+ * unmodified and in its entirety in all distributions of the software,
+ * modified or unmodified, in source code or in binary form.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ * Declaration of a BTB entry and BTB indexing policy.
+ */
+
+#ifndef __CPU_PRED_BTB_ENTRY_HH__
+#define __CPU_PRED_BTB_ENTRY_HH__
+
+#include <vector>
+
+#include "arch/generic/pcstate.hh"
+#include "base/intmath.hh"
+#include "base/types.hh"
+#include "cpu/static_inst.hh"
+#include "mem/cache/replacement_policies/replaceable_entry.hh"
+#include "mem/cache/tags/indexing_policies/base.hh"
+#include "params/BTBIndexingPolicy.hh"
+#include "params/BTBSetAssociative.hh"
+
+namespace gem5 {
+
+class BTBTagType
+{
+  public:
+    struct KeyType
+    {
+        Addr address;
+        ThreadID tid;
+    };
+    using Params = BTBIndexingPolicyParams;
+};
+
+using BTBIndexingPolicy = IndexingPolicyTemplate<BTBTagType>;
+template class IndexingPolicyTemplate<BTBTagType>;
+
+class BTBSetAssociative : public BTBIndexingPolicy
+{
+  public:
+    PARAMS(BTBSetAssociative);
+    using KeyType = BTBTagType::KeyType;
+
+    BTBSetAssociative(const Params &p)
+        : BTBIndexingPolicy(p, p.num_entries, p.set_shift),
+          tagMask(mask(p.tag_bits))
+    {
+        setNumThreads(p.numThreads);
+    }
+
+  protected:
+    /**
+     * Extract the set index for the instruction PC based on tid.
+     */
+    uint32_t
+    extractSet(const KeyType &key) const
+    {
+        return ((key.address >> setShift)
+                ^ (key.tid << (tagShift - setShift - log2NumThreads)))
+            & setMask;
+    }
+
+  public:
+    /**
+     * Find all possible entries for insertion and replacement of an address.
+     */
+    std::vector<ReplaceableEntry*>
+    getPossibleEntries(const KeyType &key) const override
+    {
+        auto set_idx = extractSet(key);
+
+        assert(set_idx < sets.size());
+
+        return sets[set_idx];
+    }
+
+    /**
+     * Set number of threads sharing the BTB
+     */
+    void
+    setNumThreads(unsigned num_threads)
+    {
+        log2NumThreads = log2i(num_threads);
+    }
+
+    /**
+     * Generate the tag from the given address.
+     */
+    Addr
+    extractTag(const Addr addr) const override
+    {
+        return (addr >> tagShift) & tagMask;
+    }
+
+    Addr regenerateAddr(const KeyType &key,
+                        const ReplaceableEntry* entry) const override
+    {
+        panic("Not implemented!");
+        return 0;
+    }
+
+  private:
+    const uint64_t tagMask;
+    unsigned log2NumThreads;
+};
+
+namespace branch_prediction
+{
+
+class BTBEntry : public ReplaceableEntry
+{
+  public:
+    using IndexingPolicy = gem5::BTBIndexingPolicy;
+    using KeyType = gem5::BTBTagType::KeyType;
+    using TagExtractor = std::function<Addr(Addr)>;
+
+    /** Default constructor */
+    BTBEntry(TagExtractor ext)
+        : inst(nullptr), extractTag(ext), valid(false), tag({MaxAddr, -1})
+    {}
+
+    /** Update the target and instruction in the BTB entry.
+     *  During insertion, only the tag (key) is updated.
+     */
+    void
+    update(const PCStateBase &_target,
+           StaticInstPtr _inst)
+    {
+        set(target, _target);
+        inst = _inst;
+    }
+
+    /**
+     * Checks if the given tag information corresponds to this entry's.
+     */
+    bool
+    match(const KeyType &key) const
+    {
+        return isValid() && (tag.address == extractTag(key.address))
+            && (tag.tid == key.tid);
+    }
+
+    /**
+     * Insert the block by assigning it a tag and marking it valid. Touches
+     * block if it hadn't been touched previously.
+     */
+    void
+    insert(const KeyType &key)
+    {
+        setValid();
+        setTag({extractTag(key.address), key.tid});
+    }
+
+    /** Copy constructor */
+    BTBEntry(const BTBEntry &other)
+    {
+        valid      = other.valid;
+        tag        = other.tag;
+        inst       = other.inst;
+        extractTag = other.extractTag;
+        set(target, other.target);
+    }
+
+    /** Assignment operator */
+    BTBEntry& operator=(const BTBEntry &other)
+    {
+        valid      = other.valid;
+        tag        = other.tag;
+        inst       = other.inst;
+        extractTag = other.extractTag;
+        set(target, other.target);
+
+        return *this;
+    }
+
+    /**
+     * Checks if the entry is valid.
+     */
+    bool isValid() const { return valid; }
+
+    /**
+     * Get tag associated to this block.
+     */
+    KeyType getTag() const { return tag; }
+
+    /** Invalidate the block. Its contents are no longer valid. */
+    void
+    invalidate()
+    {
+        valid = false;
+        setTag({MaxAddr, -1});
+    }
+
+    /** The entry's target. */
+    std::unique_ptr<PCStateBase> target;
+
+    /** Pointer to the static branch inst at this address */
+    StaticInstPtr inst;
+
+    std::string
+    print() const override
+    {
+        return csprintf("tag: %#x tid: %d valid: %d | %s", tag.address, tag.tid,
+                        isValid(), ReplaceableEntry::print());
+    }
+
+  protected:
+    /**
+     * Set tag associated to this block.
+     */
+    void setTag(KeyType _tag) { tag = _tag; }
+
+    /** Set valid bit. The block must be invalid beforehand. */
+    void
+    setValid()
+    {
+        assert(!isValid());
+        valid = true;
+    }
+
+  private:
+    /** Callback used to extract the tag from the entry */
+    TagExtractor extractTag;
+
+    /**
+     * Valid bit. The contents of this entry are only valid if this bit is set.
+     * @sa invalidate()
+     * @sa insert()
+     */
+    bool valid;
+
+    /** The entry's tag. */
+    KeyType tag;
+};
+
+} // namespace gem5::branch_prediction
+/**
+ * This helper generates a tag extractor function object
+ * which will be typically used by Replaceable entries indexed
+ * with the BaseIndexingPolicy.
+ * It allows to "decouple" indexing from tagging. Those entries
+ * would call the functor without directly holding a pointer
+ * to the indexing policy which should reside in the cache.
+ */
+static constexpr auto
+genTagExtractor(BTBIndexingPolicy *ip)
+{
+    return [ip] (Addr addr) { return ip->extractTag(addr); };
+}
+
+}
+
+#endif //__CPU_PRED_BTB_ENTRY_HH__
--- a/src/cpu/pred/simple_btb.cc
+++ b/src/cpu/pred/simple_btb.cc
@@ -44,84 +44,38 @@
 #include "base/trace.hh"
 #include "debug/BTB.hh"
 
-namespace gem5
-{
-
-namespace branch_prediction
+namespace gem5::branch_prediction
 {
 
 SimpleBTB::SimpleBTB(const SimpleBTBParams &p)
     : BranchTargetBuffer(p),
-        numEntries(p.numEntries),
-        tagBits(p.tagBits),
-        instShiftAmt(p.instShiftAmt),
-        log2NumThreads(floorLog2(p.numThreads))
+      btb("simpleBTB", p.numEntries, p.associativity,
+          p.btbReplPolicy, p.btbIndexingPolicy,
+          BTBEntry(genTagExtractor(p.btbIndexingPolicy)))
 {
     DPRINTF(BTB, "BTB: Creating BTB object.\n");
 
-    if (!isPowerOf2(numEntries)) {
+    if (!isPowerOf2(p.numEntries)) {
         fatal("BTB entries is not a power of 2!");
     }
-
-    btb.resize(numEntries);
-
-    for (unsigned i = 0; i < numEntries; ++i) {
-        btb[i].valid = false;
-    }
-
-    idxMask = numEntries - 1;
-
-    tagMask = (1 << tagBits) - 1;
-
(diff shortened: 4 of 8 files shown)


pranith:
@giactra I tried using the templatized version of the cache library as an exercise. Can you please review this change? Thank you!

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,207 @@
...
+  public:
+    struct KeyType
+    {
+        Addr address;
+        ThreadID tid;
+    };
+    using Params = BaseIndexingPolicyParams;
+};
+
+class BTBEntry : public CacheEntry
```
giactra: I wouldn't inherit from CacheEntry, as this is only adding complexity IMHO with no gain. It forces you to overload and hide some methods. The BTB entry should inherit from ReplaceableEntry

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,207 @@
...
+
+
+class BTBIndexingPolicy : public SetAssociative
+{
+  public:
+    using KeyType = BTBTagTypes::KeyType;
+    /**
+     * Convenience typedef.
+     */
+    typedef BTBIndexingPolicyParams Params;
```
giactra: Nit: prefer using for consistency for what you are doing above with KeyType

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * Mark the following as private so as not to call
+     * them directly without tid
+     */
+    using CacheEntry::match;
+    using CacheEntry::insert;
+};
+
+
+class BTBIndexingPolicy : public SetAssociative
```
giactra: Similarly to the BTBEntry discussion above. We need to pay the price and not inherit from SetAssociative. SetAssociative and the parent BaseIndexingPolicy are assuming an Addr key, whereas here our key is the Addr + Tid pair. I suggest you to have a look at what I did with TaggedIndexingPolicy... This is the hard bit I recon, as it requires you to define a templated SimObject. Happy to provide further assistance if needed

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,207 @@
...
+    {
+        log2NumThreads = log2i(num_threads);
+    }
+
+  private:
+    unsigned log2NumThreads;
+
+    /* Mark these private as we need tid  */
+    using SetAssociative::extractSet;
+    using SetAssociative::getPossibleEntries;
```
giactra: Ditto this is a workaround

src/cpu/pred/simple_btb.cc
```diff
@@ -44,84 +44,37 @@
...
-    unsigned btb_idx = getIndex(instPC, tid);
-    Addr inst_tag = getTag(instPC);
-
-    assert(btb_idx < numEntries);
-
-    if (btb[btb_idx].valid
-        && inst_tag == btb[btb_idx].tag
-        && btb[btb_idx].tid == tid) {
-        return &btb[btb_idx];
-    }
```
giactra, line 116: With your change we are not checking if the btb entry is valid anymore. You need to check for validity in the match method

giactra (changes requested):
Thanks @pranith for this change! As this is a purely refactoring change, could you run some workload with O3 and see if there are any stats difference between pre and after?

src/cpu/pred/simple_btb.cc
pranith: The CacheEntry's 'match' checks for the valid flag.

src/cpu/pred/btb_entry.hh
pranith: I tried inheriting it from ReplaceableEntry and that just seems to be doing most of what CacheEntry is already doing for valid and other methods. We are duplicating most of the code here again.

Is there any reason why you prefer inheriting from ReplaceableEntry instead of CacheEntry and avoid duplication?

pranith:
@giactra Thanks for the review! I tried inheriting from the base classes (instead of CacheEntry and IndexingPolicy) but it seems like we need to copy most of what these classes provide. I feel it becomes needlessly more labor intensive and error prone to copy the code and duplicate it in multiple places. 

Can you please provide the rationale for using the base classes? Only reason I can think of is that we are overriding base class methods with methods of a different signature, but I don't see what is wrong in doing this.

src/cpu/pred/btb_entry.hh
giactra: There are several reasons IMO

1) Design concern: Public virtual methods define the interface of a class. If you inherit from a class but you end up overloading such interface, then you are basically saying the BtbEntry is not really a CacheEntry, it is just reusing some of its code.

2) Possible bugs: this hiding can lead to potential bugs, see https://github.com/gem5/gem5/pull/1374

3) Tag ambiguity: the idea of defining a tag as the single address is just an artifact of past design. In this case the tag type is really the pair of addr and tid. But by using the CacheEntry you are basically defining two different tag types.

4) Performance: most of the BTBEntry logic is calling virtual functions. So you are needlessly adding entries to a VTABLE for a polymorphism that is never used, and you are calling them paying the vtable indirection penalty.

I understand that some code will look the same between the CacheEntry and the BTBEntry but IMHO the overhead is minimal and we would benefit a lot in decoupling the two entries, simplifying the design.

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,207 @@
...
+  public:
+    /**
+     * Find all possible entries for insertion and replacement of an address.
+     *
+     * @param addr The addr to a find possible entries for.
+     * @param tid The thread id
+     * @return The possible entries.
+     */
+    std::vector<ReplaceableEntry*>
+    getPossibleEntries(const KeyType &key) const
```
giactra: Please mark this as override

src/cpu/pred/simple_btb.cc
giactra: Ah right

src/cpu/pred/simple_btb.hh
```diff
@@ -60,77 +84,31 @@ class SimpleBTB : public BranchTargetBuffer
...
-        /** Pointer to the static branch instruction at this address */
-        StaticInstPtr inst = nullptr;
-    };
-
-
     /** Returns the index into the BTB, based on the branch's PC.
      *  @param inst_PC The branch to look up.
      *  @return Returns the index into the BTB.
      */
     inline unsigned getIndex(Addr instPC, ThreadID tid);
```
giactra: It looks like this has been removed

src/cpu/pred/simple_btb.hh
```diff
@@ -60,77 +84,31 @@ class SimpleBTB : public BranchTargetBuffer
...
-    /** Log2 NumThreads used for hashing threadid */
-    unsigned log2NumThreads;
+    AssociativeCache<BTBEntry> btb;
 };
 
-} // namespace branch_prediction
-} // namespace gem5
+} // namespace gem5::branch_prediction
+
+}
```
odanrc: nit: re-add `// namespace gem5`

src/cpu/pred/simple_btb.hh
```diff
@@ -41,17 +41,41 @@
...
+#include "cpu/pred/btb_entry.hh"
+#include "mem/cache/replacement_policies/replaceable_entry.hh"
+#include "mem/cache/tags/indexing_policies/base.hh"
 #include "params/SimpleBTB.hh"
+#include "params/BTBIndexingPolicy.hh"
 
 namespace gem5
 {
 
+using BTBIndexingPolicy = IndexingPolicyTemplate<branch_prediction::BTBTagTypes>;
```
odanrc: I think this line exceeds the character limit (Lines must be a maximum of 79 characters long).

src/cpu/pred/simple_btb.hh
```diff
@@ -41,17 +41,41 @@
...
+template class IndexingPolicyTemplate<branch_prediction::BTBTagTypes>;
+
 namespace branch_prediction
 {
 
+/**
+ * This helper generates an a tag extractor function object
```
odanrc: nit: ´an a´

src/cpu/pred/simple_btb.hh
```diff
@@ -41,17 +41,41 @@
 #ifndef __CPU_PRED_SIMPLE_BTB_HH__
 #define __CPU_PRED_SIMPLE_BTB_HH__
 
+#include <memory>
```
odanrc: <memory> is not used in this file

src/cpu/pred/simple_btb.cc
```diff
@@ -44,84 +44,37 @@
...
 {
 
 SimpleBTB::SimpleBTB(const SimpleBTBParams &p)
     : BranchTargetBuffer(p),
-        numEntries(p.numEntries),
-        tagBits(p.tagBits),
-        instShiftAmt(p.instShiftAmt),
-        log2NumThreads(floorLog2(p.numThreads))
+      btb("simpleBTB", p.numEntries, p.associativity,
+          p.btbReplPolicy, p.btbIndexingPolicy, BTBEntry(genTagExtractor(p.btbIndexingPolicy)))
```
odanrc: Ditto line length

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,207 @@
...
+    using KeyType = BTBTagTypes::KeyType;
+    using TagExtractor = std::function<Addr(Addr)>;
+
+    /** Default constructor */
+    BTBEntry(TagExtractor ext)
+        : CacheEntry(ext), inst(nullptr), tid(0)
+    {}
+
+    /** Update the entry in the BTB */
+    void
```
odanrc, line 160: I think this function deserves an improvement in the documentation. Something to explain its difference from the insert() function. It seems to be a workaround due to the fact that the insertion only uses the key and the target and inst are not part of the key.

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,207 @@
...
+    update(ThreadID _tid,
```
odanrc: Since it seems that we only update() right after calling insert, the _tid assignment is redundant and should probably be dropped.

configs/common/cores/arm/HPI.py
```diff
@@ -1683,6 +1683,11 @@ class HPI_MMU(ArmMMU):
 class HPI_BTB(SimpleBTB):
     numEntries = 128
     tagBits = 18
+    associativity = 1
+    btbReplPolicy = LRURP()
+    btbIndexingPolicy = BTBIndexingPolicy(
+        size="384B", entry_size=3, assoc=associativity
```
odanrc: Curiosity: where do the config numbers come from? (I am not too familiar with BTBs)

src/cpu/pred/btb_entry.hh
pranith: @giactra I tried your suggestion and created a BTBIndexingPolicy/BTBSetAssociative policy derived from SimObject. However, I seem to be running into a linking error. Can you please help?

configs/common/cores/arm/HPI.py
pranith: It is approximately 'numEntires * entry_size' so that indexing is as before. I will re-work this part once before merging the final changes.

src/cpu/pred/btb_entry.hh
pranith: Ah, looks like I missed implementing the regenerateAddr method. It works now.

src/cpu/pred/simple_btb.hh
pranith: Done.

src/cpu/pred/simple_btb.hh
pranith: Fixed.

src/cpu/pred/simple_btb.hh
pranith: Fixed.

src/cpu/pred/simple_btb.hh
pranith: Removed.

src/cpu/pred/simple_btb.cc
pranith: Fixed.

src/cpu/pred/btb_entry.hh
pranith: Thanks for the explanation @giactra. BtbEntry is derived from CacheEntry means it has some properties of CacheEntry, but not all. This is I think a textbook example of OOP (mammals -> humans).

But, I am not really strong against implementing it from scratch. Just that, there will be code duplication.

I pushed a patch addressing all the review comments. Please take a look!

src/cpu/pred/btb_entry.hh
pranith: Done.

src/cpu/pred/btb_entry.hh
pranith: Removed.

src/cpu/pred/btb_entry.hh
pranith: Done.

src/cpu/pred/btb_entry.hh
pranith: Done.

src/cpu/pred/btb_entry.hh
pranith: Removed tid argument.

pranith:
> Thanks @pranith for this change! As this is a purely refactoring change, could you run some workload with O3 and see if there are any stats difference between pre and after?

I reran with the patch and there were no stat differences in BTB. I updated the BTB params to keep the same behavior as before.

src/cpu/pred/btb_entry.hh
odanrc: I don´t see the change. Have you pushed it?

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,262 @@
...
+    /** Pointer to the static branch inst at this address */
+    StaticInstPtr inst;
+
+    /** The entry's thread id. */
+    ThreadID tid;
+
+    std::string
+    print() const override
+    {
+        return csprintf("tag: %#x valid: %d | %s", getTag(),
```
odanrc: The tid should also be printed.

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,262 @@
...
+    void
+    insert(const KeyType &key)
+    {
+        setValid();
+        setTag(extractTag(key.address));
+        tid = key.tid;
+    }
+
+    /** Copy constructor */
+    BTBEntry(const BTBEntry &other)
```
odanrc, line 190: The rule of 3 dictates that if we need to define copy constructor then we should also need to define the copy assignment operator.

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * Checks if the entry is valid.
+     */
+    virtual bool isValid() const { return valid; }
+
+    /**
+     * Get tag associated to this block.
+     */
+    virtual Addr getTag() const { return tag; }
```
odanrc: I think that it should instead return KeyType, since the tid *is* part of the tag. We need to make sure that what happened with the prefetchers (some use the secure bit, and others don´t) doesn´t happen in this class. We can already see an instance of the bug happening in this class´ print() and invalidate(), where the tid was forgotten.

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * Valid bit. The contents of this entry are only valid if this bit is set.
+     * @sa invalidate()
+     * @sa insert()
+     */
+    bool valid;
+
+    /** The entry's tag. */
+    Addr tag;
```
odanrc: Instead of storing it as an address, we should store it as a KeyType. This class should not use Addr.

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    /** The entry's target. */
+    std::unique_ptr<PCStateBase> target;
+
+    /** Pointer to the static branch inst at this address */
+    StaticInstPtr inst;
+
+    /** The entry's thread id. */
+    ThreadID tid;
```
odanrc: As per the other comments, this should be the same visibility as the ´tag´ variable.

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,262 @@
...
+    {
+        return csprintf("tag: %#x valid: %d | %s", getTag(),
+                        isValid(), ReplaceableEntry::print());
+    }
+
+  protected:
+    /**
+     * Set tag associated to this block.
+     */
+    virtual void setTag(Addr _tag) { tag = _tag; }
```
odanrc: It should have used KeyType, because otherwise we end up forgetting about tid (see invalidate() using this function and not assigning a ´max tid´).

odanrc (changes requested):

src/cpu/pred/btb_entry.hh
pranith: Sorry! missed pushing it earlier... pushed it in the latest commit.

src/cpu/pred/btb_entry.hh
pranith: Added a copy assignment operator.

src/cpu/pred/btb_entry.hh
pranith: Changed the tag type to be KeyType througout.

src/cpu/pred/btb_entry.hh
pranith: Removed the independent tid member.

src/cpu/pred/btb_entry.hh
pranith: Updated.

src/cpu/pred/btb_entry.hh
pranith: Done.

src/cpu/pred/btb_entry.hh
pranith: Done.

odanrc (approved):

pranith:
@odanrc Thaks for the review. I squashed all the commits into one. Not sure if this is can be done during merge?

powerjg:
> I reran with the patch and there were no stat differences in BTB. I updated the BTB params to keep the same behavior as before.

I just want to double check that even though you saw there were no stats differences, did you make sure the BTB is being used in the tests? Our test coverage of many models like the BTB is pretty bad. You would likely have had to write your own test for this.

pranith:
> 
> I just want to double check that even though you saw there were no stats differences, did you make sure the BTB is being used in the tests? Our test coverage of many models like the BTB is pretty bad. You would likely have had to write your own test for this.

The BTB sees the same number of accesses, hits, and, misses with the change for a small test program. So yes, I can confirm that it is being exercised and behaves as before the patch.

odanrc:
> @odanrc Thaks for the review. I squashed all the commits into one. Not sure if this is can be done during merge?

I think for this small PR there will be no issue, but since there are other reviewers still reviewing I think it would have been better to wait for all approvals to be in order, in case someone was using the commits as part of their review methodology. IIRC, the merge does not trigger a merge squash, but I may be wrong.

pranith:
@powerjg Can you please point me to docs on how to run the CI tests locally? It seems some quick boot tests are failing.

powerjg:
See https://github.com/gem5/gem5/blob/stable/TESTING.md#running-tests-in-batch

pranith:
@giactra Can you please review? This now passes CI tests and is ready for merge. Please let me know if you want me to squash the two commits together or you can do it during merge. Thanks!

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,275 @@
...
+     */
+    virtual KeyType getTag() const { return tag; }
+
+    /** Invalidate the block. Its contents are no longer valid. */
+    virtual void
+    invalidate()
+    {
+        valid = false;
+        setTag({MaxAddr, -1});
+    }
```
giactra: Are you planning on extending this class? Otherwise I would consider devirtualizing these methods

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,275 @@
...
+     */
+    virtual void setTag(KeyType _tag) { tag = _tag; }
+
+    /** Set valid bit. The block must be invalid beforehand. */
+    virtual void
+    setValid()
+    {
+        assert(!isValid());
+        valid = true;
+    }
```
giactra: Ditto

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,275 @@
...
+    /**
+     * Find all possible entries for insertion and replacement of an address.
+     */
+    std::vector<ReplaceableEntry*>
+    getPossibleEntries(const KeyType &key) const override
+    {
+        auto set_idx = extractSet(key);
+
+        fatal_if(set_idx >= std::pow(2, tagShift - setShift),
+            "Set index outside range");
```
giactra: This looks expensive. can't we just assert (meaning we can skip this in the fast build) and simply check if set_idx is bigger or equal than sets.size()

This is what we were doing in the old version:

`    assert(btb_idx < numEntries);`

configs/common/cores/arm/HPI.py
```diff
@@ -1683,6 +1683,11 @@ class HPI_MMU(ArmMMU):
 class HPI_BTB(SimpleBTB):
     numEntries = 128
     tagBits = 18
+    associativity = 1
+    btbReplPolicy = LRURP()
+    btbIndexingPolicy = BTBSetAssociative(
+        size="512B", entry_size=4, assoc=associativity
```
giactra: I feel your pain here. I understand you are carefully choosing size and entry_size to account for numEntries and the instShift.

It is 100% not your fault but I believe this is a bit fragile. What happens if someone changes numEntries in the BTB?

To overcome this problem I just created:

https://github.com/gem5/gem5/pull/1557

It will allow you to do

    btbIndexingPolicy = BTBSetAssociative(
        num_entries=numEntries, set_shift=instShiftAmt, assoc=associativity)

Would you be up for waiting for my PR to be merged?

giactra (commented):
Hi @pranith I have some minor comments and a more important one that relates to my new PR: https://github.com/gem5/gem5/pull/1557

(I developed the PR once I saw this one here and its problems)

configs/common/cores/arm/HPI.py
pranith: Thanks for the PR @giactra. It definitely makes it much easier to instantiate the structures. I will rebase on top of your pull request and update this patch.

src/cpu/pred/btb_entry.hh
pranith: Updated.

src/cpu/pred/btb_entry.hh
pranith: Removed virtual.

src/cpu/pred/btb_entry.hh
pranith: Updated as suggested.

configs/common/cores/arm/HPI.py
pranith: Updated accordingly.

pranith:
@andysan Now that we have the tag based addressing in the cache, can you please review the 'tagMask' commit of this PR? We need tagMask for the BTB (which I just realized).Thanks!

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,275 @@
...
+#include "base/types.hh"
+#include "cpu/static_inst.hh"
+#include "mem/cache/replacement_policies/replaceable_entry.hh"
+#include "mem/cache/tags/indexing_policies/base.hh"
+#include "params/BTBIndexingPolicy.hh"
+#include "params/BTBSetAssociative.hh"
+
+namespace gem5 {
+
+class BTBTagTypes
```
odanrc: Nit: this is a tag type - i.e., not plural

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,275 @@
...
+using BTBIndexingPolicy = IndexingPolicyTemplate<BTBTagTypes>;
+template class IndexingPolicyTemplate<BTBTagTypes>;
+
+class BTBSetAssociative : public BTBIndexingPolicy
+{
+  public:
+    PARAMS(BTBSetAssociative);
+    using KeyType = BTBTagTypes::KeyType;
+
+    BTBSetAssociative(const Params &p) : BTBIndexingPolicy(p, p.num_entries, p.set_shift)
```
odanrc: Fix to fit in line length

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,275 @@
...
+    void
+    setNumThreads(unsigned num_threads)
+    {
+        log2NumThreads = log2i(num_threads);
+    }
+
+    Addr regenerateAddr(const KeyType &key,
+                        const ReplaceableEntry* entry) const override
+    {
+        return (key.address << tagShift) | (entry->getSet() << setShift);
```
odanrc: Is this function being called and its output verified anywhere? At a glance this does not seem to be correct because it lacks the xor on the tid

src/cpu/pred/btb_entry.hh
```diff
@@ -0,0 +1,275 @@
...
+    {
+        set(target, _target);
+        inst = _inst;
+    }
+
+    /**
+     * Checks if the given tag information corresponds to this entry's.
+     */
+    bool
+    match(const KeyType &key)
```
odanrc: This should probably be a const function - I don´t think we ever expect to change the class when checking for a match

src/cpu/pred/btb_entry.hh
pranith: Updated.

src/cpu/pred/btb_entry.hh
pranith: Fixed.

src/cpu/pred/btb_entry.hh
pranith: I don't think we use this function. I replaced the body with an 'assert(false)'. It would be good to remove it from the base class and only define it for mem-cache related policies, but it didn't seem straight forward.

src/cpu/pred/btb_entry.hh
pranith: Updated.

src/cpu/pred/btb_entry.hh
odanrc: Regarding not having it, if you are sure it is not going to be needed, then sure we don´t need to implement it. One alternative is to make the base class no longer pure virtual and panic there. Speaking of which, it would be better to use something in the lines of ´panic(not implemented)´ than the assertion, as it is not a user error, but a lack of feature.

pranith: I don't know what to provide in the default non-pure virtual function, so I just updated the assert() to a panic.

andysan:
> @andysan Now that we have the tag based addressing in the cache, can you please review the 'tagMask' commit of this PR? We need tagMask for the BTB (which I just realized).Thanks!

My main concern with the tag bits param is that there is no way for users to reason about this. Does setting it to something other than 64 truncate the TID or the address? What if we add even more information (we'll need to do this if we want to be representative of real hardware) in the tags does it truncate the same fields as before? I think realistically, we can't expect users to be able to use this param.

A better solution would be to deal with truncation in the btb indexing policy instead where the semantic information is available. That'd let users decide how many bits of address to include, whether to include the TID, VMID, etc., which would be much more meaningful than just truncating some bits.

I'm currently out of office, so I won't have time to look at this in more detail. @giactra could you have a look and drive this from our side?

pranith:
> A better solution would be to deal with truncation in the btb indexing policy instead where the semantic information is available. That'd let users decide how many bits of address to include, whether to include the TID, VMID, etc., which would be much more meaningful than just truncating some bits.

I updated BTB indexing policy to use tag mask independent of the base policy. There are other users of cache library that need similar tag mask like bop prefetcher, but I left it alone for now.

> 
> I'm currently out of office, so I won't have time to look at this in more detail. @giactra could you have a look and drive this from our side?

@giactra Can you please review? Thanks!

pranith:
@giactra @odanrc Can you please review this updated patch? It would be good to have it merged if there are no new comments.

src/cpu/pred/BranchPredictor.py
```diff
@@ -83,6 +85,38 @@ class BranchTargetBuffer(ClockedObject):
     numThreads = Param.Unsigned(Parent.numThreads, "Number of threads")
 
 
+class BTBIndexingPolicy(SimObject):
+    type = "BTBIndexingPolicy"
+    abstract = True
+    cxx_class = "gem5::IndexingPolicyTemplate<gem5::BTBTagType>"
+    cxx_header = "cpu/pred/btb_entry.hh"
+    cxx_template_params = ["class Types"]
```
odanrc, line 93: How does this line work? Is this string a description of the template param?

odanrc:
@pranith I cannot spot the differences between the previous version and the new version. This is probably because the commits were squashed together. Could you point to where we should focus the review efforts? As a side note, in the future I recommend avoiding squashing new commits added to a change that was already under review, as it makes spotting the changes much harder.

pranith:
> @pranith I cannot spot the differences between the previous version and the new version. This is probably because the commits were squashed together. Could you point to where we should focus the review efforts? As a side note, in the future I recommend avoiding squashing new commits added to a change that was already under review, as it makes spotting the changes much harder.

I apologize for this. There were some conflicts in the merge and I had to rebase to fix the conflicts. I will add commits addressing review commits only in the future.

The main change from your last review is the addition of the 'tagMask' field in the BTBIndexingPolicy. Rest of it remains the same.

src/cpu/pred/BranchPredictor.py
pranith: I copy+pasted this from the 'BaseIndexingPolicy' from 'mem/cache/tags/indexing_policies/IndexingPolicies.py' since I am trying to reuse the IndexingPolicyTemplate just like it does. Please let me know if this needs to be updated.

src/cpu/pred/BranchPredictor.py
```diff
@@ -93,6 +127,20 @@ class SimpleBTB(BranchTargetBuffer):
     instShiftAmt = Param.Unsigned(
         Parent.instShiftAmt, "Number of bits to shift instructions by"
     )
+    associativity = Param.Unsigned(1, "BTB associativity")
+    size = Param.MemorySize("16384", "Size of BTB")
```
giactra: Is this a leftover from the initial version?

giactra (changes requested):
Hi @pranith apologies for the late review. I only have one comment left. Then if @odanrc is also happy I will approve the PR

src/cpu/pred/BranchPredictor.py
pranith: Indeed, removed the line. I will squash the commits if you think this is OK. Thanks!

giactra: Yes feel free to squash

pranith: Done.

odanrc (approved):

giactra (approved):

Issue #1207, opened 2024-06-07
arch-riscv: remove default `vlen=256` parameters


saul44203:
While doing changes in relation to #1187 there seems to be some micro instruction related classes (e.g. `VectorMicroInst`) that have the `vlen` parameter with a default value in the constructor. Some macro instruction related classes have this as well. To remove them some changes need to be made.

I also noticed that the way micro instructions get the current `vlen` is not consistent. Some get it through their macro instruction, that gets it from the decoder, and some just get it from the pc state directly.

 Before doing this we should maybe choose a consistent way to pass `vlen` and refactor accordingly.

saul44203:
My opinion is that since vlen is fixed instructions should get it once decoded, removing it from the pc state altogether. The decoder already gets the vlen through the isa and passes it as a constructor parameter to some if not all vector instructions, so we only need to propagate this parameter correctly to all the child classes to homogenize all the instructions.

I'm starting with this but feel free to discuss.

PR #1538, opened 2024-09-04
arch-riscv: add VLEN/ELEN as class attributes for all vec insts

saul44203:
This refactor attempts to homogenize all riscv's vector (macro/micro) instruction classes so that ELEN and VLEN are guaranteed to be a class attribute. Since both are constant, all instructions will get it on the decoding process passed through to their vector base class.

This allows the removal of VLEN in the PC state and also in some constructor default parameters (solves issue #1207).

Change-Id: I6f0471004335f49b00b015c37e95dc7f9569e303

### DIFF
--- a/src/arch/riscv/insts/vector.cc
+++ b/src/arch/riscv/insts/vector.cc
@@ -412,10 +412,10 @@ VMvWholeMicroInst::generateDisassembly(Addr pc,
 }
 
 VMaskMergeMicroInst::VMaskMergeMicroInst(ExtMachInst extMachInst,
-    uint8_t _dstReg, uint8_t _numSrcs, uint32_t _vlen, size_t _elemSize)
+    uint8_t _dstReg, uint8_t _numSrcs, uint32_t _elen, uint32_t _vlen,
+    size_t _elemSize)
     : VectorArithMicroInst("vmask_mv_micro", extMachInst,
-                            SimdAddOp, 0, 0),
-      vlen(_vlen),
+                            SimdAddOp, 0, 0, _elen, _vlen),
       elemSize(_elemSize)
 {
     setRegIdxArrays(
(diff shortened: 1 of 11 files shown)


powerjg:
I have a minor issue with moving the elen/vlen to the static instruction... I believe that the current "API" (I put it in quotes because it's not an official API but it's more of the current status quo) assumes that every PC will generate exactly one static instruction. In other words, the assumption is that every time you decode a particular PC you get the exact same static instruction. (Others can correct me if I'm wrong.)

This assumption shows up most obviously in all other ISAs where they use a "decode cache" (see arch/generic/decode_cache.hh) and the decoders of each ISA that isn't RISC-V. The decode cache holds the static instructions for each PC and returns the *old* static instruction if the same PC is decoded again.

I have to admit, I don't know why RISC-V doesn't use this... I did a simple search of the git history, but didn't find anything.

Correct me if I'm wrong, but I believe this change violates the assumption that every time you decode a particular address it will return with the same static instruction. With this change, the static instruction depends on the address and the current state of the decoder.

Given gem5's architecture, I still feel that it's more natural to have the elen/vlen as part of the `PCState`. My apologies that I missed the discussion on #1207. Please let me know what you think.

saul44203:
Hi @powerjg, 

I think I see your point but ELEN and VLEN are both implementation defined constants in RVV. In gem5 the decoder gets both from the ISA on instantiation, and they don't change anymore, so I think there's no state in that sense (or at least it's always the same).

Apart from this I'm not doing something new, some instructions already had VLEN/ELEN as a class parameter (that has to be passed by the decoder if I'm not mistaken), so this patch just attempts to homogenize things to prevent using multiple paths to get VLEN/ELEN (e.g. to mitigate errors like #1187).

Maybe there's some other reasons for keeping it in the `PCState`, but I don't see them. Let me know what you think.

aarmejach:
Just to try to move the needle in this one. As @saul44203 explained, this patch homogenizes how we access the VLEN and ELEN architectural constants. Currently, access to these constants is done in two different ways: (i) some instructions have them defined as members, (ii) others access via the PCState. Since these are constants, having them in the PCState does not provide much benefit, and local access in the micro instruction object is much more convenient.

hnpl (approved):
The change looks good to me!

Since ELEN and VLEN are constants for the entire simulation, we can optimize a bit more by having something similar to a singleton for those values. Though I don't think we save that much of memory usage as these values are store in StaticInst objects, and StaticInst objects are cached.

powerjg (approved):
Apologies for taking so long to get back to you. Thanks for correcting my misunderstanding.

PR #1542, opened 2024-09-06
SE script and tests for risc-v's vector extension

saul44203:
This two commits add the SE config and test script, respectively, to run the rvv tests mentioned in #1246.

Marking it as a draft as the path for the binaries (and maybe the path/name for the SE config) will change.

### DIFF
--- a/configs/example/gem5_library/riscv-rvv-example.py
+++ b/configs/example/gem5_library/riscv-rvv-example.py
@@ -1,0 +1,120 @@
+# Copyright (c) 2024 Barcelona Supercomputing Center
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+# list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors
+# may be used to endorse or promote products derived from this software without
+# specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
+This script demonstrates how to run RISC-V vector-enabled binaries in SE mode
+with gem5. It accepts the number of CORES, VLEN, and ELEN as optional
+parameters, as well as the resource name to run. If no resource name is
+provided, a list of available resources will be displayed. If one is given the
+simulation will then execute the specified resource binary with the selected
+parameters until completion.
+
+
+Usage
+-----
+
+# Compile gem5 for RISC-V
+scons build/RISCV/gem5.opt
+
+# Run the simulation
+./build/RISCV/gem5.opt configs/example/gem5_library/riscv-rvv-example.py \
+    [-c CORES] [-v VLEN] [-e ELEN] <resource>
+
+"""
+
+import argparse
+
+from m5.objects import RiscvO3CPU
+
+from gem5.components.boards.simple_board import SimpleBoard
+from gem5.components.cachehierarchies.classic.private_l1_private_l2_cache_hierarchy import (
+    PrivateL1PrivateL2CacheHierarchy,
+)
+from gem5.components.memory import SingleChannelDDR3_1600
+from gem5.components.processors.base_cpu_core import BaseCPUCore
+from gem5.components.processors.base_cpu_processor import BaseCPUProcessor
+from gem5.isas import ISA
+from gem5.resources.resource import obtain_resource
+from gem5.simulate.simulator import Simulator
+from gem5.utils.requires import requires
+
+
+class RVVCore(BaseCPUCore):
+    def __init__(self, elen, vlen, cpu_id):
+        super().__init__(core=RiscvO3CPU(cpu_id=cpu_id), isa=ISA.RISCV)
+        self.core.isa[0].elen = elen
+        self.core.isa[0].vlen = vlen
+
+
+requires(isa_required=ISA.RISCV)
+
+resources = [
+    "rvv-branch",
+    "rvv-index",
+    "rvv-matmul",
+    "rvv-memcpy",
+    "rvv-reduce",
+    "rvv-saxpy",
+    "rvv-sgemm",
+    "rvv-strcmp",
+    "rvv-strcpy",
+    "rvv-strlen",
+    "rvv-strlen-fault",
+    "rvv-strncpy",
+]
+
+parser = argparse.ArgumentParser()
+parser.add_argument("resource", type=str, choices=resources)
+parser.add_argument("-c", "--cores", required=False, type=int, default=1)
+parser.add_argument("-v", "--vlen", required=False, type=int, default=256)
+parser.add_argument("-e", "--elen", required=False, type=int, default=64)
+
+args = parser.parse_args()
+
+cache_hierarchy = PrivateL1PrivateL2CacheHierarchy(
+    l1d_size="32KiB", l1i_size="32KiB", l2_size="512KiB"
+)
+
+memory = SingleChannelDDR3_1600()
+
+processor = BaseCPUProcessor(
+    cores=[RVVCore(args.elen, args.vlen, i) for i in range(args.cores)]
+)
+
+board = SimpleBoard(
+    clk_freq="1GHz",
+    processor=processor,
+    memory=memory,
+    cache_hierarchy=cache_hierarchy,
+)
+
+binary = obtain_resource(args.resource)
+board.set_se_binary_workload(binary)
+
+simulator = Simulator(board=board, full_system=False)
+print("Beginning simulation!")
+simulator.run()
--- a/tests/gem5/se_mode/rvv_intrinsic_tests/test.py
+++ b/tests/gem5/se_mode/rvv_intrinsic_tests/test.py
@@ -1,0 +1,63 @@
+# Copyright (c) 2024 Barcelona Supercomputing Center
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+# list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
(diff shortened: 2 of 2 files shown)


configs/example/riscv-se.py
```diff
@@ -0,0 +1,71 @@
+import argparse
```
powerjg: This file needs a copyright statement

configs/example/riscv-se.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+        for i in range(num_cores)
+    ]
+
+    for i in range(num_cores):
+        for j in range(cores[i].core.numThreads):
+            cores[i].core.isa[j].vlen = vlen
+            cores[i].core.isa[j].elen = elen
+
+    return cores
```
powerjg: Instead of this, can you extend the `BaseCPUCore`? We would prefer to show people this style as "examples" since it's more extensible if you have more customizations.

E.g.,

```python
class RVVCore(BaseCPUCore):
    def __init__(self, vlen, elen):
        super.__init__(core=RiscvO3CPU(cpu_id=i), isa=ISA.RISCV)
        self.core.isa.vlen = vlen
        self.core.isa.elen = elen
```

tests/gem5/se_mode/rvv_intrinsic_tests/test.py
```diff
@@ -0,0 +1,45 @@
...
+]
+
+vlens = [2**x for x in range(7, 15)]
+
+if not (src_dir := os.environ.get("SRCDIR")):
+    print("SRCDIR env var not defined. Exiting.", file=sys.stderr)
+    exit()
+
+for binary in binaries:
+    bin_path = f"{src_dir}/apps/rvv-intrinsic-tests/bin/{binary}"
```
powerjg: These should be on https://resources.gem5.org/ and we should use `obtain_resource`. Please work with @Harshil2107 to get him the source and explain to him how to build the binaries so he can upload them.

tests/gem5/se_mode/rvv_intrinsic_tests/test.py
```diff
@@ -0,0 +1,45 @@
...
+        print(f"Binary {bin_path} does not exist. Skipping.", file=sys.stderr)
+        continue
+
+    out_verifier = verifier.MatchRegex(re.compile(f"^{bin_path}: pass$"))
+    for vlen in vlens:
+        gem5_verify_config(
+            name=f"test-riscv-{binary}-vlen_{vlen}-O3-se-mode",
+            fixtures=(),
+            verifiers=(out_verifier,),
+            config=f"{config.base_dir}/configs/example/riscv-se.py",
```
powerjg: Please put the config file in the tests repository. I would prefer to not have tests depend on scripts outside of the tests directory (unless the test is *testing the script*).

configs/example/riscv-se.py
powerjg, line 1: This name is misleading. I would call the file `riscv-rvv-example.py` and put it in `configs/example/gem5_library/` since it is using the stdlib. I would call out the difference between this file and the `gpu_se.py` file in this directory :)

powerjg (changes requested):

tests/gem5/se_mode/rvv_intrinsic_tests/test.py
saul44203: Hi @Harshil2107,

I have opened a repo with the source and a simple Makefile you can use to compile them [here](https://github.com/saul44203/rvv-intrinsic-tests/tree/main). If you have any problem please tell me.

configs/example/gem5_library/riscv-rvv-example.py
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) 2024 Barcelona Supercomputing Center
```
powerjg, line 1: Can you copy the full BSD block like the other files. There's some argument that it's not necessary, but we don't want to start that argument again ;). To make everyone happy, please copy the whole block.

tests/gem5/se_mode/rvv_intrinsic_tests/test.py
powerjg, line 1: Needs copyright

tests/gem5/se_mode/rvv_intrinsic_tests/test.py
powerjg: We'll get on this :)

powerjg (changes requested):

tests/gem5/se_mode/rvv_intrinsic_tests/test.py
Harshil2107: Hi @saul44203,

When I try to make the binaries, I get the following error

```shell
src/rvv_branch.c:2:10: fatal error: riscv_vector.h: No such file or directory
    2 | #include <riscv_vector.h>
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:18: rvv_branch] Error 1
harshilp@ad3.ucdavis.edu@amarillo:~/misc-repos
```

Also, does the gcc version make a difference in this case? I am using `riscv64-linux-gnu-gcc-12`.

It would be great if we can add the source of the binaries to [gem5-resources-repo](https://github.com/gem5/gem5-resources), as we can point to the source of the binaries.

saul44203: This is an internal compiler header file used for the rvv intrinsics, so I think your gcc version might be the culprit. Try `riscv64-linux-gnu-gcc-13` if you can.

Harshil2107: I was able to build the binaries by using `riscv64-linux-gnu-gcc-13` in a docker container. I will upload them to resources. Is it ok if I copy the foler from your repo to gem5-resources as a source for these resources?

saul44203: Great. Yes, you can copy it.

Harshil2107: I uploaded the binaries and made JSON objects so that we can now use `obtain_resource` to get the binary.

I need a few things:

- What name should I use as author?
- Is there a specific license that I should mention for the resource?
- What would be a good one line description for the resource?

Here are the links to the binaries:
[rvv-strncpy](https://resources.gem5.org/resources/rvv-strncpy?version=1.0.0)  
[rvv-strlen](https://resources.gem5.org/resources/rvv-strlen?version=1.0.0)  
[rvv-memcpy](https://resources.gem5.org/resources/rvv-memcpy?version=1.0.0)  
[rvv-sgemm](https://resources.gem5.org/resources/rvv-sgemm?version=1.0.0)  
[rvv-strcmp](https://resources.gem5.org/resources/rvv-strcmp?version=1.0.0)  
[rvv-index](https://resources.gem5.org/resources/rvv-index?version=1.0.0)  
[rvv-branch](https://resources.gem5.org/resources/rvv-branch?version=1.0.0)  
[rvv-matmul](https://resources.gem5.org/resources/rvv-matmul?version=1.0.0)  
[rvv-reduce](https://resources.gem5.org/resources/rvv-reduce?version=1.0.0)  
[rvv-saxpy](https://resources.gem5.org/resources/rvv-saxpy?version=1.0.0)  
[rvv-strcpy](https://resources.gem5.org/resources/rvv-strcpy?version=1.0.0)  
[rvv-strlen-fault](https://resources.gem5.org/resources/rvv-strlen-fault?version=1.0.0)

Please let me know if I need to update or remake the binaries.

saul44203: Thanks @Harshil2107. I updated the config with `obtain_resource` and all seems to work well.

As for the questions:
- This tests are just minor adaptations of the [riscv-non-isa/rvv-intrinsic-doc examples](https://github.com/riscv-non-isa/rvv-intrinsic-doc/tree/main/examples), so maybe we should reference that.
- I'm not sure about the license. Their repo is under BSD-3 and CC-BY-4.0, so maybe both?
- We could use something like "RISC-V vector intrinsic {name} example program", where name is "strncpy", "strlen", etc.

powerjg (dismissed):

tests/gem5/se_mode/rvv_intrinsic_tests/test.py
```diff
@@ -0,0 +1,63 @@
...
+    out_verifier = verifier.MatchRegex(re.compile(f"^.*{resource}: pass$"))
+
+    for vlen in vlens:
+        gem5_verify_config(
+            name=f"test-riscv-{resource}-vlen_{vlen}-O3-se-mode",
+            fixtures=(),
+            verifiers=(out_verifier,),
+            config=f"{config.base_dir}/tests/gem5/se_mode/rvv_intrinsic_tests/config.py",
+            config_args=[resource, f"--vlen={vlen}"],
+            valid_isas=(constants.riscv_tag,),
```
Harshil2107: The CI tests are failing because they can not find the `build/RISCV/gem5.opt` binary. The CI tests only build ALL so I think changing this line to `valid_isas=(constants.all_compiled_tag,)` should get the tests to run.

Harshil2107 (changes requested):

tests/gem5/se_mode/rvv_intrinsic_tests/test.py
saul44203: Done

configs/example/gem5_library/riscv-rvv-example.py
```diff
@@ -0,0 +1,85 @@
...
+        super().__init__(core=RiscvO3CPU(cpu_id=cpu_id), isa=ISA.RISCV)
+        self.core.isa[0].elen = elen
+        self.core.isa[0].vlen = vlen
+
+
+requires(isa_required=ISA.RISCV)
+
+parser = argparse.ArgumentParser()
+parser.add_argument("binary", type=str)
+parser.add_argument("arguments", nargs="*", type=str, default=[])
```
BobbyRBruce: I think requiring the user to provide a binary with RVV instructions here may be a big ask for something that's supposed to be an example. Could we do two things here:

1. Change this to use gem5 resources (i.e., we'd set the workload with `board.set_se_binary_workload(obtain_resource("<ID>"), arguments=args.arguments)`). This means they can easily get pre-built binaries to run the example.
2. Have this argument be a list of resources to choose from, so they don't have to go figure out what resources work with this. I think you do choices in argparse with
`parser.add_argument("workloads", choices=("rvv-branch", "rvv-index", "etc")`

In terms of arguments: it looks like most of the binaries we have available just take in the vlen as an argument, so we don't need the user to specify it here.



I realize this example script was probably made so users could play with the example and try different binaries, etc., but I'm always very hesitant to make the examples too flexible as I've found many users stop seeing them as examples and instead just "the way you interface with gem5". I don't want people to see this and think "The way you run RVV binaries in gem5 is with 'configs/example/gem5_library/riscv-rvv-example'".

configs/example/gem5_library/riscv-rvv-example.py
```diff
@@ -0,0 +1,85 @@
...
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
```
BobbyRBruce, line 27: Can you put a docstting here explaining what this example does and a quick example of usage? Like this: https://github.com/gem5/gem5/blob/b1a44b89c7bae73fae2dc547bc1f871452075b85/configs/example/gem5_library/arm-ubuntu-run-with-kvm.py#L27

BobbyRBruce (changes requested):
Really love that this is being provided as an example. It'll really help as a reference. Testing for RVV is also fantastic.

I have a couple comments to address but I'm thankful for this

saul44203:
Hi @BobbyRBruce. I think I have addressed all the comments. Tell me if I missed something or we have to change anything else.

BobbyRBruce (changes requested):
It seems that the example script is identical to the test script. Can you just delete the "test/gem5/se_mode/rvv_instrict_tests/config.py" and just modigy tests.py to load the example. With this we get to have a test that both tests RVV but also ensures the examples we provide still function.

saul44203:
> It seems that the example script is identical to the test script. Can you just delete the "test/gem5/se_mode/rvv_instrict_tests/config.py" and just modigy tests.py to load the example. With this we get to have a test that both tests RVV but also ensures the examples we provide still function.

Right. Should be done now.

Harshil2107 (approved):
lgtm

BobbyRBruce (approved):

PR #1548, opened 2024-09-08
util-docker: Move `LABEL` to after image import

BobbyRBruce:
A Dockerfile must start with the importation of a docker base image. It is only after this point that `LABEL` be provided. Having `LABEL` at the top of the Dockerfiles resulted in the Docker images failing to build.

Change-Id: I7213fb1cf01dc004ec91e58c21e4fc17f0b68aa8


erin-le (approved):
Looks good to me

PR #1553, opened 2024-09-11
port: Make binding-related functions virtual

handsomeliu-google:
It would be useful if developer can have logs, or put wrapped monitors in their derived ports when they are bound. This commit makes the binding-related functions of Request/Response Port and SignalPort virtual.

Change-Id: I55c65267eba14793f1c102ab048e2e3044974d0f


giactra:
Hi @handsomeliu-google this is not needed. Once the method is marked as virtual in the base class it stays virtual across the child classes

handsomeliu-google:
Ah yes you're right. Let me close this PR.

PR #1555, opened 2024-09-12
DRAFT/REFERENCE: gem5 Bridge Driver with Read/Write Op Devices

nkrim:
This PR should not be merged as-is and is primarily included as a reference for how to augment the `gem5_bridge` driver introduced in PR:https://github.com/gem5/gem5/pull/1480

The code in this PR is a copy of a former implementation of the above PR, where many of these features were omitted for the sake of a simpler and more direct implementation. This implementation, on the other hand, offers a suite of devices under `/dev/gem5`. The mappable device used by the `m5` utility is `/dev/gem5/bridge` and then there are additional devices for each gem5op that produce the same behavior as calls to the `m5` utility.

Besides `readfile` and `writefile`, all ops can be triggered by writing a string of arguments to them. `readfile` can be read from directly, and `writefile` can be written to directly after defining a destination target through `/dev/gem5/writefiledest`. Here are a few usage examples:

```
echo "0" > /dev/gem5/exit
echo "1 2 3 4 5 6" > /dev/gem5/sum && cat /dev/gem5/retval
cat /dev/gem5/readfile > file.txt
echo "path/to/target.txt" > /dev/gem5/writefiledest && cat source.txt > /dev/gem5/writefile
```

Though this is replicating the behavior of the `m5` utility, it offers a new platform for experimenting with gem5ops interfaces. The implementation of these ops is done mostly as a proof of concept.

Beyond the complexity and thus maintenance concerns of this implementation, calls to this driver are less desirable during workloads as they perform extra transitions into the kernel. The `m5` utility, when linked to the workload program, can incur just a single transition to mmap to the gem5_bridge device at the start of the program, and then all pokes that trigger the ops will happen in user space, reducing the overhead of the ops that would appear in the workload statistics. Additionally, introducing magic-instruction ops into the driver would be another layer of complexity but would be a requirement for replacing the `m5` utility.

### DIFF
--- a/util/gem5_bridge_all_ops/Makefile
+++ b/util/gem5_bridge_all_ops/Makefile
@@ -1,5 +1,5 @@
 DRIVER  := gem5_bridge
-SOURCES := gem5_bridge gem5_ops
+SOURCES := gem5_bridge gem5_ops gem5_poke
 GROUP   := gem5
 
 ifneq ($(KERNELRELEASE),)
--- a/util/gem5_bridge_all_ops/Makefile
+++ b/util/gem5_bridge_all_ops/Makefile
@@ -1,8 +1,10 @@
-DRIVER := gem5_bridge
-GROUP  := gem5
+DRIVER  := gem5_bridge
+SOURCES := gem5_bridge gem5_ops
+GROUP   := gem5
 
 ifneq ($(KERNELRELEASE),)
-obj-m += $(DRIVER).o
+obj-m := $(DRIVER).o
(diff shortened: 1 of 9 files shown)


erin-le:
Hey @nkrim, since https://github.com/gem5/gem5/pull/1480 has been merged, would you mind if I closed this PR? It could be reopened at a later date if you wanted to work on adding the features in this PR.

nkrim:
> Hey @nkrim, since #1480 has been merged, would you mind if I closed this PR? It could be reopened at a later date if you wanted to work on adding the features in this PR.

Sounds good!

PR #1557, opened 2024-09-12
 mem-cache: Do not require p.size and p.entry_size in IP template 

giactra:
This PR is adjusting the constructor to relax template
requirements. In this way child classes are free to provide
their own way of calculating the number of entries and the
shifting required to extract the set

Why do we need this?
Up to this patch we have been configuring the indexing policy
by setting up the cache/table size (in bytes) and the entry size.
Those parameters make a lot of sense in caching structures
where:

a) We want to configure the caching structure using
the amount of storage (in bytes) provided (e.g. 4kB of Cache)
b) the content of a single entry is addressable therefore
we need the entry size to know how many bits in the indexing
process we need to shift to extract the set

In those cases the number of cache entries is derived from the formula

num_entries = size / entry_size

The adoption of the IndexingPolicy for different kinds
of caching structures (e.g. prefetcher tables) make this
way of configuring the IP a bit quirky.

For some tables directly setting the number of entries is a far more
intuitive way of configuring the IP, instead of allocating the desired
number of entries by working things out with the formula above

### DIFF
--- a/src/mem/cache/tags/Tags.py
+++ b/src/mem/cache/tags/Tags.py
@@ -46,12 +46,6 @@ class TaggedIndexingPolicy(SimObject):
     cxx_header = "mem/cache/tags/tagged_entry.hh"
     cxx_template_params = ["class Types"]
 
-    # Get the size from the parent (cache)
-    size = Param.MemorySize(Parent.size, "capacity in bytes")
-
-    # Get the entry size from the parent (tags)
-    entry_size = Param.Int(Parent.entry_size, "entry size in bytes")
-
     # Get the associativity
     assoc = Param.Int(Parent.assoc, "associativity")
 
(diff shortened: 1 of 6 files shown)


giactra:
@odanrc @pranith

powerjg (commented):
lgtm. I'll wait for others to approve since they've been working on this code.

pranith (approved):

odanrc:
I am not a fan of this PR, but the current way to instantiate indexing policy is indeed very confusing, so I will approve it.

odanrc (approved):

giactra:
> I am not a fan of this PR, but the current way to instantiate indexing policy is indeed very confusing, so I will approve it.

Yeah I couldn't find a better way.. For a moment I thought about using enable_if and provide different constructors in base depending on whether p.size or p.num_entries were defined.. But it looked a bit over-engineered

giactra:
Hey @powerjg are you happy to approve this PR or do you need more time? :) 
Thanks!

powerjg (approved):
Sorry, just missed it :)

PR #1560, opened 2024-09-12
arch-arm: Move generateTrap from MiscRegOp to ArmStaticInst

giactra:
System(Misc) register accesses are not the only trappable instructions.
We move the exception generation logic (generateTrap) from the
MiscRegOp64 to the base ArmStaticInst


Harshil2107 (approved):
Not an expert but this looks good to me.

PR #1564, opened 2024-09-15
mem-ruby: Fix replacement policy in GPU_VIPER

yuxiaojia:
The current GPU_VIPER protocol's TCC cache update the MRU information
twice with calling a_allocateBlock and ut_updateTag which affects the
LIP and RRIP replacement polies. Remove ut_updateTag fixes the LIP and
RRIP replacement polies.

Change-Id: I79ad9392593e00425a7fe8828048465b2c2c2e1f

### DIFF
--- a/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
@@ -1162,7 +1162,6 @@ machine(MachineType:TCC, "TCC Cache")
   transition(I, WrVicBlkBack, W) {TagArrayRead, TagArrayWrite, DataArrayWrite} {
     p_profileMiss;
     a_allocateBlock;
-    ut_updateTag;
     swb_sendWBAck;
     wdb_writeDirtyBytes;
     p_popRequestQueue;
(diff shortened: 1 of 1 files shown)


mattsinc (approved):
Thanks again!  This makes sense to me, but I'll let others chime in before merging.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -1162,7 +1162,6 @@ machine(MachineType:TCC, "TCC Cache")
   transition(I, WrVicBlkBack, W) {TagArrayRead, TagArrayWrite, DataArrayWrite} {
```
powerjg, line 1162: Does this still need to do a tag array write?

powerjg (approved):
lgtm. I will go ahead and approve because I think the answer to my question is "yes"

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
mattsinc: In my opinion yes -- because a_allocateBlock is updating the tag.

abmerop (approved):

mattsinc:
Thanks again, @yuxiaojia !

PR #1567, opened 2024-09-15
misc: Remove Serialize-related code in Random

aperais:
The Random ser/des support has been non-existent since 2014.
Removing it will enable the Random class to be unit tested
without having a dependency on the src/sim code.

### DIFF
--- a/src/base/random.cc
+++ b/src/base/random.cc
@@ -42,8 +42,6 @@
 
 #include <sstream>
 
-#include "base/logging.hh"
-
 namespace gem5
 {
 
--- a/src/base/random.cc
+++ b/src/base/random.cc
@@ -43,7 +43,6 @@
 #include <sstream>
 
 #include "base/logging.hh"
-#include "sim/serialize.hh"
 
 namespace gem5
 {
@@ -67,33 +66,6 @@ void
 Random::init(uint32_t s)
 {
     gen.seed(s);
-}
-
-void
-Random::serialize(CheckpointOut &cp) const
-{
-    panic("Currently not used anywhere.\n");
-
-    // get the state from the generator
-    std::ostringstream oss;
-    oss << gen;
-    std::string state = oss.str();
-    paramOut(cp, "mt_state", state);
-}
-
-void
-Random::unserialize(CheckpointIn &cp)
-{
-    panic("Currently not used anywhere.\n");
-
-    // the random generator state did not use to be part of the
-    // checkpoint state, so be forgiving in the unserialization and
-    // keep on going if the parameter is not there
-    std::string state;
-    if (optParamIn(cp, "mt_state", state)) {
-        std::istringstream iss(state);
-        iss >> gen;
-    }
 }
 
 Random random_mt;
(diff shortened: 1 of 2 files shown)


aperais:
For context, following the discussion in #1534, I am proposing we move the ser/des part of the `Random` class into its own class and file, to allow unit tests to be added for `Random` without needing to pull all the dependency on `Serializable` when compiling.

Also note that the ser/des code currently panics so another way would be to get rid of it altogether.

odanrc:
From git blame:
83a46bfc09c src/base/random.cc (Andreas Hansson  2014-09-03 07:42:55 -0400  75)     panic("Currently not used anywhere.\n");

2014... I believe we can safely remove it instead :)

src/base/serializable_random.hh
```diff
@@ -0,0 +1,72 @@
...
+
+class Checkpoint;
+
+/**
+ * This class is a wrapper around Random for anyone
+ * needing random generators to be serialized/unserialized.
+ * This enables Random to be unit-tested without pulling
+ * all files from src/sim
+ */
+class SerializableRandom : Serializable, Random
```
giactra: Use public keyword

src/base/serializable_random.hh
```diff
@@ -0,0 +1,72 @@
...
+{
+
```
giactra: Nit: whitespace

src/base/serializable_random.cc
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2024 CNRS
+ * All rights reserved
```
giactra: You removed Arm from the header

src/base/serializable_random.hh
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright (c) 2024 CNRS
+ * All rights reserved
```
giactra: Ditto

giactra (commented):
Thanks @aperais . This serialization is not really implemented so we should consider removing it at a certain point. However it is fine for now to move it aside.

src/base/serializable_random.cc
aperais: Got trigger happy when creating a new file, sorry. It's gone now anyway :)

src/base/serializable_random.hh
aperais: Ditto

aperais:
> Thanks @aperais . This serialization is not really implemented so we should consider removing it at a certain point. However it is fine for now to move it aside.

Following @odanrc investigation into git blame, removing it will not bother anyone and I suppose it's better not to add new files. Let's just remove it.

src/base/random.cc
```diff
@@ -43,7 +43,6 @@
 #include <sstream>
 
 #include "base/logging.hh"
```
giactra: I suspect you don't need these headers anymore

(logging was to use the panic)

giactra (changes requested):
LGTM, one minor detail

src/base/random.cc
aperais: Fixed !

giactra (approved):

powerjg (approved):
I love it when code is removed :)

PR #1571, opened 2024-09-17
python: Add warning message for set_workload being called twice

erin-le:
This commit adds a warning message for when set_workload is called twice, as users typically do not mean to do this.

### DIFF
--- a/src/python/gem5/components/boards/kernel_disk_workload.py
+++ b/src/python/gem5/components/boards/kernel_disk_workload.py
@@ -75,8 +75,6 @@ class KernelDiskWorkload:
         * This assumes the Linux kernel is used.
     """
 
-    _is_workload_set = False
-
     @abstractmethod
     def get_default_kernel_args(self) -> List[str]:
         """
@@ -183,9 +181,9 @@ class KernelDiskWorkload:
         # Abstract board. This function will not work otherwise.
         assert isinstance(self, AbstractBoard)
 
-        if self._is_workload_set:
+        if self.is_workload_set():
             warn("Workload has been set more than once!")
-        self._is_workload_set = True
+        self.set_is_workload_set(True)
 
         # Set the disk device
         self._disk_device = disk_device
--- a/src/python/gem5/components/boards/kernel_disk_workload.py
+++ b/src/python/gem5/components/boards/kernel_disk_workload.py
@@ -33,6 +33,7 @@ from typing import (
     Union,
 )
 
+import m5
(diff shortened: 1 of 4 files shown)


src/python/gem5/components/processors/base_cpu_core.py
```diff
@@ -135,6 +135,11 @@ def connect_walker_ports(self, port1: Port, port2: Port) -> None:
     def set_workload(self, process: Process) -> None:
         self.core.workload = process
 
+    def is_workload_set(self) -> bool:
```
Harshil2107: It might be useful to move this function to `AbstractCore` so that anyone who makes their own cores would also have access to this function.

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -187,6 +188,9 @@ def set_kernel_disk_workload(
         # full-system simulation.
         self._set_fullsystem(True)
 
+        if self.workload.object_file:
```
Harshil2107: You might be able to get access to `is_workload_set()` function if it is moved to `AbstractCore` as the `AbstractBoard` has an `AbstractProcessor` which has the `AbstractCore`, so it might be possible to use the same method to do this check as done in the `set_se_binary_workload` function.

Harshil2107 (changes requested):

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -74,6 +75,8 @@ def get_default_kernel_args(self) -> List[str]:
         * This assumes the Linux kernel is used.
     """
 
+    _is_workload_set = False
```
Harshil2107: I think it would be better to move this to `abstract_board` with functions to get the value and set it like we do with the `_is_fs` function. This will move the logic to one single place instead of having to write the logic in every set function that sets workloads.

Harshil2107 (changes requested):

Harshil2107 (dismissed):
lgtm

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -141,6 +145,7 @@ def set_se_binary_workload(
             #
             # A better API for this which avoids `isinstance` checks would be
             # welcome.
+
```
powerjg: Spurious whitespace change

Harshil2107 (approved):

PR #1574, opened 2024-09-17
arch-x86,stdlib: added MADT entries on the X86Board

kaustav-goswami:
This change adds MADT entries to the X86Board. Previously, the kernel in full-system mode was complaining about a `ACPI BIOS Error (bug): Invalid table length 0x24 in RSDT/XSDT (20190816/tbutils-291)`. This patch fixes the invalid length and initializes all the tables correctly.

### DIFF
--- a/src/python/gem5/components/boards/x86_board.py
+++ b/src/python/gem5/components/boards/x86_board.py
@@ -41,6 +41,10 @@ from m5.objects import (
     Pc,
     Port,
     RawDiskImage,
+    X86ACPIMadt,
+    X86ACPIMadtIntSourceOverride,
+    X86ACPIMadtIOAPIC,
+    X86ACPIMadtLAPIC,
     X86E820Entry,
     X86FsLinux,
     X86IntelMPBus,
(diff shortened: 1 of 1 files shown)


src/python/gem5/components/boards/x86_board.py
```diff
@@ -236,6 +262,14 @@ def assignISAInt(irq, apicPin):
         self.workload.intel_mp_table.base_entries = base_entries
         self.workload.intel_mp_table.ext_entries = ext_entries
 
+        madt = X86ACPIMadt(
+            local_apic_address=0, records=madt_entries, oem_id="madt"
+        )
+        self.workload.acpi_description_table_pointer.rsdt.entries.append(madt)
+        self.workload.acpi_description_table_pointer.xsdt.entries.append(madt)
```
powerjg, line 269: I am a bit skeptical that this works. Usually, `append` does not work with `VectorSimObject`.

Have you verified that this `madt` shows up in the config.ini file?

kaustav-goswami: Yes, `X86ACPIMadt` and `X86ACPIXSDT` shows up on config.ini.

Harshil2107 (approved):
Not an expert but this looks good to me.

PR #1575, opened 2024-09-18
ext,util-docker: updated SST to v.14.0.0

kaustav-goswami:
This change updates SST from v.13.0.0 to v.14.0.0. It also adds an updated docker file to test the new version.


powerjg:
How has this been tested?

kaustav-goswami:
> How has this been tested?



> How has this been tested?

I manually tested whether `ext/sst/sst/example.py` works. I did not see a dedicated test for the same. The only test file that I could find was in `tests/deprecated/nightly.sh`, which uses a docker image, which can be created by `util/dockerfiles/sst/Dockerfile`. I am unsure if I need to compile and push the docker image.

BobbyRBruce (approved):
Tested this on my end too. It works fine.

PR #1580, opened 2024-09-19
misc: Make random gen portable across compilers.

aperais:
Replace std::uniform_*_distribution by custom code
to make random number generation in gem5 portable across
compilers.

Of note, FP random number generation was not uniformly
distributed, and this PR does not fix that issue.

Thanks to Chandana S. Deshpande (deshpande.s.chandana@gmail.com)
for uncovering the issue.


aperais:
The current issue with the random number generation method is that it uses `std::uniform_int_distribution` or `std::uniform_real_distribution`  which are implementation dependent. The STL of clang and gcc do not implement it the same way, so simulation results of the same gem5 commit are different if compiled differently (see https://github.com/orgs/gem5/discussions/1517). The mersenne twister used to provide random numbers is not implementation dependent.

The PR replaces the distributions with custom code. I ran an experiment with the old code and the new code to see what the distributions looked like.

**This PR breaks a lot of tests which match some stats against an existing JSON file, because behavior will change slightly. => Actually does not**

## 10M uint16_t random number distribution

Perfect uniform distribution would be for each number to have frequency 1.52.10-5 (each number should appear 152.6 times in the 10M).

Before (avg value: 32762.01, stdev: 18918.84)

image
 
Now (avg value: 32759.92, stdev: 18918.94)

image

## 10M uint16_t between 256 and 10000

Perfect uniform distribution would be for each number to have frequency 1.02.10e-4 (each number should appear 1026 times).

Before (avg value: 5127.31, stdev: 2813.17)

image

After (avg value: 5128.20, stdev: 2813.60)

image


## 10M float

I'm not too well versed in FP arithmetic, so I won't pretend I can compute the expected frequency, however before and after give exactly the same result, suggesting gcc implements `std::uniform_real_distribution` the way I did it. In any case, even the current code seems to have trouble staying uniformly distributed as it will not produce values between 0 and 0.1 very often compared to other values. 

**If someone has a solution to make it "more uniform", feel free to chime in.**

Before (avg value: 0.50, stdev: 0.289)

image

After (Exact same with "Before")

image

Code to generate the numbers is attached (
[randomtest.zip](https://github.com/user-attachments/files/17075036/randomtest.zip)
)  : `g++ -o randomtest randomtest.cpp -O3 --std=c++17 && ./randomtest && python plot.py` (takes a couple minutes).

**Edit: Up to date figures, I had not rerun with 10M (was 50M, but same story)**
**Edit 2: Updated random range code to handle `min == max`**

src/base/random.hh
```diff
@@ -102,18 +101,20 @@ class Random : public Serializable
...
      * @ingroup api_base_utils
      */
     template <typename T>
     typename std::enable_if_t<std::is_integral_v<T>, T>
     random(T min, T max)
     {
-        std::uniform_int_distribution<T> dist(min, max);
-        return dist(gen);
+        assert(min < max);
+        T r = (gen() - min) % (max - min) + min;
```
powerjg: Maybe it already exists, but it would be good to test this with small numbers to make sure the inclusive/exclusive on the limits works. I.e., `random(0, 2)` should return 0, 1, and 2.

src/base/random.hh
```diff
@@ -102,18 +101,20 @@ class Random : public Serializable
     random()
     {
         // [0, 1) for real types
-        std::uniform_real_distribution<T> dist;
-        return dist(gen);
+        return (T) random<uint64_t>() /
```
powerjg: There are limits to this implementation. I.e., the resulting distribution is not perfectly uniform. But I think that's OK in our use case.

powerjg (commented):
Thanks for this change. I'm strongly in favor of finding ways to make gem5 more deterministic.

src/base/random.hh
aperais: That's correct, but also true for the current code. Maybe I can add a `warn` statement.

src/base/random.hh
aperais: Yes, planning to add some of the unit tests I've added for #1534

aperais: Done (see random.test.cc, last test). Btw, is there a guideline regarding files copyrighting?  Since I added some code for testing I put my institution as copyright owner but that may not be how it's supposed to work.

src/base/random.hh
aperais: Done

src/base/random.hh
powerjg: For copyright, you'll have to ask your institution. For instance, here is my institution's policy: https://copyright.universityofcalifornia.edu/resources/copyright-ownership-faqs.html

src/base/random.hh
powerjg: Sorry for not being clear... I was just making a generic comment, not asking you to make any changes. I would vote against having a `warn` statement here so that gem5 is less verbose. I feel like a lot of our warnings are ignored because we have so many already. I would suggest adding a comment. If you think a warning is more appropriate, then it should be `warn_once`.

powerjg (commented):
Other than the `warn` being overly verbose, the change looks good to me!

src/base/random.test.cc
```diff
@@ -0,0 +1,112 @@
...
+
+using namespace gem5;
+
+/**
+ * Checking that default construction uses the default
+ * seed as specified by the standard
+ */
+TEST(Random, DefaultConstruct)
+{
+#ifdef NDEBUG
```
odanrc: These skips should not be added: these are not death tests, so no assertions will be triggered.

src/base/random.test.cc
```diff
@@ -0,0 +1,112 @@
...
+        "stripped out of fast builds";
+#endif
+    Random dut;
+
+    // Count occurences if we want to check
+    // frequencies in the future
+    std::unordered_map<int, int> values;
+
+    for (int i = 0; i < 500; i++) {
+      values[dut.random<int>(0,2)]++;
```
odanrc: Fix indentation

src/base/random.test.cc
```diff
@@ -0,0 +1,112 @@
...
+TEST(Random, RangeCoverage)
+{
+#ifdef NDEBUG
+    GTEST_SKIP() << "Skipping as assertions are "
+        "stripped out of fast builds";
+#endif
+    Random dut;
+
+    // Count occurences if we want to check
+    // frequencies in the future
```
odanrc, line 133: Out of curiosity, how well did the values get distributed? Can we add checks for that, or is that too limiting?

src/base/random.test.cc
```diff
@@ -0,0 +1,112 @@
...
+
+    for (int i = 0; i < 500; i++) {
+      values[dut.random<int>(0,2)]++;
+    }
+
+    ASSERT_EQ(values.count(0), 1);
+    ASSERT_EQ(values.count(1), 1);
+    ASSERT_EQ(values.count(2), 1);
+    ASSERT_EQ(values.size(), 3);
+}
```
odanrc, line 187: Add death test for min > max (this one will need the skip statement)

odanrc, line 187: Could you add a test for min == max?

src/base/random.test.cc
```diff
@@ -0,0 +1,112 @@
...
+#include <gtest/gtest-spi.h>
+#include <gtest/gtest.h>
+
+#include <unordered_map>
+
+#include "base/gtest/logging.hh"
+#include "base/random.hh"
+
+using namespace gem5;
+
```
odanrc, line 38: Could you add equivalent tests for floating points?

src/base/random.hh
```diff
@@ -99,18 +99,22 @@ class Random
...
      */
     template <typename T>
     typename std::enable_if_t<std::is_integral_v<T>, T>
     random(T min, T max)
     {
-        std::uniform_int_distribution<T> dist(min, max);
-        return dist(gen);
+        assert(min <= max);
+        // + 1 to handle cases where min == max
+        T r = (gen() - min) % (max - min + 1) + min;
```
odanrc: Why `- min`? I think this should be:
        T r = gen() % (max - min + 1) + min;

src/base/random.hh
```diff
@@ -99,18 +99,22 @@ class Random
     random()
     {
         // [0, 1) for real types
-        std::uniform_real_distribution<T> dist;
-        return dist(gen);
+        warn("FP random numbers are not uniformly distributed.");
+        return (T) random<uint64_t>() /
```
odanrc: The call to random<uint64_t> adds an unnecessary modulo operation: (gen() % max<uint64_t>) / max<uint64_t>. Any reason not to use gen() directly? Also, could you add parenthesis to where the T casts are being applied to make reading easier?

odanrc (changes requested):

src/base/random.test.cc
aperais: FP API already provides numbers between 0.0 and 1.0, but currently the API does not let the user set another range. I think I did not get what exactly you wanted to test.

src/base/random.hh
aperais: Yes, that works too. Probably I tried to balance it (-min then +min) but it does not make much sense here.

src/base/random.test.cc
aperais: Looping 500 times it's 31%/38.8%/30.2%, looping 20k times it's 34%/33.3%/32.7% so within 1%. Looping 20k is fast enough, I added the frequency test.

Another tidbit is that range random can give you negative numbers if you want, whereas the scalar random is limited to positive integers.

src/base/random.hh
aperais: It's the same as Inria which has some copyright mentions in headers e.g. src/cpu/pred/tage_base.hh, so I think it's as it should be.

src/base/random.test.cc
odanrc: I meant adding something like the following:
```
TEST(RandomCtor, DoubleDefaultConstruct)
{
    // Init seed was default seed 5489
    Random dut;
    // First number corresponds to the default seed
    ASSERT_EQ(dut.random<double>(), 0,837237944); // 14514284786278117030llu / max<uint64_t>
}

TEST(RandomCtor, FloatDefaultConstruct)
{
    // Init seed was default seed 5489
    Random dut;
    // First number corresponds to the default seed
    ASSERT_EQ(dut.random<float>(), the_expected_float_value);
}
```

src/base/random.test.cc
```diff
@@ -0,0 +1,161 @@
...
+    for (int i = 0; i < 10; i++) {
+        ASSERT_EQ(dut.random<int>(0, 0), 0);
+        ASSERT_EQ(dut.random<uint64_t>(1, 1), 1);
+        ASSERT_EQ(dut.random<int>(-1, -1), -1);
+    }
+}
+
+constexpr int loopCount = 30000;
+
+bool within_freq_range(int count)
```
odanrc: The coding style dictates that functions should be in camel case.

Also, could you add documentation to this function? It is clear to me what it does, but it may not be to a future reader

src/base/random.test.cc
aperais: I added the tests, but they seem redundant to me : since float/double are currently based on integer random number generation, the fact that the first integer is the expected one should imply the first float and double are the expected one. Maybe I'm missing out the reason for adding this test : is this about ensuring FP behavior is consistent across different potential host ISA for random number generation ?

odanrc: The way that I see it, we can´t guarantee that the current behavior will be kept forever, so the test helps make sure that if it is ever changed (for example, to provide uniformity in the distribution), it is done consciously, not carelessly.

odanrc (approved):
Thanks for the changes, Arthur. LGTM

aperais:
@powerjg gentle reminder since this PR should get in before the last PR on RNG #1534

powerjg (approved):

PR #1581, opened 2024-09-19
base: Adding modulo intlv option to AddrRange

Will-Shaddix:
Added a new constructor to AddrRange which enables interleaving across any number of channels using the modulo operation instead of the traditional hashing. At this point the functionality is there but I still need to clean it up and add comments as well as fix some of the tests that were added. These changes shouldn't affect functionality of any existing work.


Will-Shaddix:
Basic idea is as follows:

- User specifies granularity by passing in "lowest_modulo_bit", so if the user wanted to interleave every 64 bytes this number should be 6
- The user also passes in how many channels we are interleaving in the "modulo_by" parameter
- From here to check if an incoming address is in the addr range we ignore the first "lowest_modulo_bit" number of bits. Then modulo by the value passed in to "modulo_by" and check if this matches intlvMatch.

- The main difficulties came in trying to removeIntlvBits and addIntlvBits because all (64 - lowest_modulo_bit) bits are used for interleaving
- So instead of just using a mask to move around the intlv bits we divide. Whatever bits are lost in the divide we know wont matter as they were already used to find the right channel.
- By using this divide we end up with a sequential address range with the size of the total AddrRange/number of channels

Will-Shaddix:
Added support for holes in AddrRange using the new modulo based interleaving constructor. A list of AddrRanges can be passed in as "holes" which this AddrRange does not contain.

Will-Shaddix:
I haven't had the time to add tests yet, I've done some basic testing with some traffic generators and it works. Would like to get some feedback before I take the time to write and annotate my tests. 

The modulo interleave tests I wrote before still need comments and some tuning.

abmerop:
I can take a closer look and give feedback soon.  I have two high level questions out of curiosity to see what this can enable.

- Does this allow for odd number of memory controllers? (e.g., could I have 6GiB of memory with 3 memory controllers?)
- Does the "holes" support allow for something like an address range from 0GiB - 5GiB with a 1GiB hole from 3GiB to 4GiB (i.e., X86's memory layout)?

Will-Shaddix:
- Yes, theoretically this should support any number of memory controllers with any size AddrRange, of course it wont divide perfectly even if the user gives it a Range of 8 GiB between 3 channels. But yes, odd number included!
- Yes, that was the exact goal for enabling these holes.

As an additional note apologies for the random meaningless changes, I tried using clang format to help with formatting but it ended up changing some other stuff which I have not been able to go through and change back yet, but I will hopefully be able to work on that some tomorrow.

abmerop:
I wasn't able to get to this last week, but I want to let you know I haven't forgot about this PR and I hope to test it out this week.

@powerjg @BobbyRBruce I think this can help finally fix the x86 "hole" problem in stdlib.  (Formerly known as gem5-11 on the old JIRA).

powerjg (commented):
I could be convinced this is a bad idea, but I feel like we should have an interface for address ranges and then multiple different implementations of that interface. I.e., an `BaseAddrRange` class with pure virtual methods and then a `BitInterleavedAddrRange` (probably typedef to `AddrRange` for backwards compatibility) and a `ModuloAddrRange`. Then, we could have other kinds of address ranges as well like `NonContiguousAddrRange` which has a hole in it (or is made up of multiple other `BaseAddrRange` instances. 

What do you think @Will-Shaddix?

Will-Shaddix:
I think ideally this sounds like a great idea considering right now I just have an if statement to check if it ModuloAddr or not. But I'm unsure of how much time this would take to implement

src/python/m5/params.py
```diff
@@ -976,7 +998,10 @@ def __str__(self):
 
     def size(self):
         # Divide the size by the size of the interleaving slice
-        return (int(self.end) - int(self.start)) >> self.intlvBits
+        if self.modulo_by == 0:
+            return int(self.end) - int(self.start) >> self.intlvBits
```
abmerop: Nit: I'd keep the parenthesis here. I had to actually look up python operator precedence to find out +,- is before <<,>>

src/python/m5/params.py
```diff
@@ -976,7 +998,10 @@ def __str__(self):
...
+
+        return (int(self.end) - int(self.start)) / self.modulo_by
```
abmerop: Do we need to subtract out the size of holes here? I don't know if that would cause any problems. However, that would allow this to interoperate nicely with pre-existing code.

For example the following sanity check in stdlib would succeed: https://github.com/gem5/gem5/blob/develop/src/python/gem5/components/memory/memory.py#L197

src/base/addr_range.hh
```diff
@@ -628,6 +837,7 @@ class AddrRange
     Addr
     getOffset(const Addr& a) const
     {
+        // potentially subtract by hole sizes / modulo_by
```
abmerop: Possibly a stray comment? Did you intend to add code here?

abmerop (commented):
First off, this is a great contribution and should help with a very long outstanding issue. Also kudos for writing gtests.  I am still trying to get something using holes to work with stdlib in full system but have not figured it out yet.  Do you have any example configs using this?

Until then I have a few minor comments below

Will-Shaddix:
I am currently trying to test it with the X86 board by adding support for this constructor in ChanneledMemory. My initial test was just in SE mode using traffic generators and not using the standard library. I can forward that script along if you'd like, or I can forward along the X86Board script once I finish testing it

src/base/addr_range.hh
Will-Shaddix: I think this was a stray comment, because we are using modulo, it shouldn't make too much of a difference i think if i were to do this.

src/python/m5/params.py
Will-Shaddix: This was something I wasn't too sure about. I think you're right and it isn't necessary.

abmerop:
> I am currently trying to test it with the X86 board by adding support for this constructor in ChanneledMemory. My initial test was just in SE mode using traffic generators and not using the standard library. I can forward that script along if you'd like, or I can forward along the X86Board script once I finish testing it

I can wait until you finish testing it.  Thanks again,

Will-Shaddix:
I haven't forgotten about this, still trying to get this to work. I'm working on changing some of the parts of the X86 board to work with this address range. Will update when I've found the solution.

Will-Shaddix:
With this commit you should be able to run things in FS mode and with kvm and in the standard library. I added a standard library component called DIMM_DDR5_4400_x86_Holes that can be used for testing. I also removed the panic from the x86 board and still need to add a warning there.  I ended up needing to make some changes to the kvm cpu to get it to work. 
    
 I left my changes on the example x86-ubuntu-run-with-kvm.py as a test for it.
 
 My next steps are to over haul the AddrRanges to have a BaseAddrRange class that ModuloAddrRange can inherit.

Will-Shaddix:
Hello, I ended up having a lot of troubles implementing inheritance with AddrRange because its used in som many places. Because it was expressed to me that we didn't want to have these if's all over the place I discussed with Jason and decided to remove all of my modulo based interleaving support. But i did leave the support for the IO hole. This should be all of the changes that I need to make! 

Theoretically this should work with multiple holes should that be needed in the future (with the exception of KVM which would require a slight change to work with multiple holes). I have run 1 workload that claims to use more than 3 GiB of RAM that ran through a few iterations. I will try to do some more testing/aask some others to do some testing of other workloads.

src/python/gem5/components/boards/x86_board.py
```diff
@@ -284,11 +304,15 @@ def _setup_memory_ranges(self):
...
+            # add a warning here about using holes
+            data_range = AddrRange(
+                start=0,
+                end=memory.get_size(),
+                holes=[AddrRange(start="3GiB", end="4GiB")],
             )
-        data_range = AddrRange(memory.get_size())
+        else:
+            data_range = AddrRange(memory.get_size())
+
```
BobbyRBruce: Why do we need have a special case for for < 3GiB? Can't we just have 

```py
            data_range = AddrRange(
                start=0,
                end=memory.get_size(),
                holes=[AddrRange(start="3GiB", end="4GiB")],
            )
```

run for both cases, only that when <3GiB the hole is just ignored as you can't have a whole in a range that doesn't contain it?

src/python/gem5/components/memory/memory.py
```diff
@@ -79,6 +79,8 @@ def __init__(
         interleaving_size: Union[int, str],
         size: Optional[str] = None,
         addr_mapping: Optional[str] = None,
+        holes: Optional[List[AddrRange]] = None,
+        # should holes be inserted here, or in the set_memory_range method?
```
BobbyRBruce: Instead of doing it this way, why don't you add a function to `AbstractMemorySystem` called `add_hole` which sets whether the memory has a hole and over what ranges. This can be something called when `connect_things` is run and the current ISA being run is X86 (which can be determined by inspecting the processor and seeing what cores it's using). 

This means all current implemented memory systems can have holes added automatically when needed and does not require the creation of new memory systems to handle this new case.

giactra:
Hi @Will-Shaddix ,

I am quickly going through the PR but I am not particularly sure I understand the full extent of the changes. Changing the way we do interleaving is one thing; I see here most of the code allows AddrRange to have several holes within itself.

Are we sure we want to over-complicate such a delicate class like AddrRange? If we want to support non contiguous ranges (and we do for sure), isn't it simple to amend memory controllers by accepting a vector/list of ranges instead of one?

Let me know if there is anything I am missing

BobbyRBruce (changes requested):
I also have the same question as @giactra on this one. Aren't we just splitting 1 range up into several when adding holes? In which case, can't we just pass a list of ranges to represent this?

E.g.: You have a system with 5GB of memory and one hole at 1GB and another at 3GB, each 1GB. Can't we pass something like:

`ranges=[AddrRange(start="0GB", end="1GB"), AddrRange(start=2GB, end=3GB, AddrRange(start=4GB, end=7GB)]`

abmerop:
@giactra @BobbyRBruce  The main benefit with the holes is you can interleave a memory range that is not contiguous. It is not as simple as having multiple memory controllers because then you could not use memory hashing. For example, 4GB of memory with the standard X86 holes is 0GB...5GB physical address range with a hole at 3GB...4GB. That could be interleaved and make use of channel/bank hashing. But if you have two ranges (0GB...3GB and 4GB...5GB) you are forced to either give up hashing or you get restricted to very specific sizes, e.g., channels with interleaving can be no more than 1GB each so they divide 3GB.

Having multiple ranges is also not supported by stdlib. Here is a quote from @powerjg on my PR when proposing this exact idea for stdlib:

> This isn't possible in gem5 in general. You can't interleave with multiple memory ranges. The only way I know to hack around this is to create a special memory controller which maps the other range. Otherwise, you would end up with * controllers, which is (also) unrealistic.

Will-Shaddix:
I think the AbstractMemory class is what holds the AddrRange of the memory, but I'm not sure if its possible to add support for multiple ranges there, and don't think I have the expertise to do that in time for the release. I think if we want to look at it again after the release I'd be happy to take a quick look then and see if it seems possible/if I think i can do it.

src/python/gem5/components/boards/x86_board.py
Will-Shaddix: Yes, That would still work, I guess I was just thinking that the hole would be within the AddrRange but it shouldnt need to be to work. I've changed this in my most recent push.

src/python/gem5/components/memory/memory.py
Will-Shaddix: I've made this change as well, and have the x86 board just add the hole itself in the stdlib, so the user will be able to do more than 3 GB with the caveat that they are losing out on 1 GB because of the hole. So if they wanted 8GB of RAM they need to pass in a 9GB size.

giactra:
> But if you have two ranges (0GB...3GB and 4GB...5GB) you are forced to either give up hashing or you get restricted to very specific sizes, e.g., channels with interleaving can be no more than 1GB each so they divide 3GB.
> 

How come? I am not familiar with the interleaving code and I am only picturing this as an image. Where is the 1GB limit coming from?

> Having multiple ranges is also not supported by stdlib. Here is a quote from @powerjg on my PR when proposing this exact idea for stdlib:
> 
> > This isn't possible in gem5 in general. You can't interleave with multiple memory ranges. The only way I know to hack around this is to create a special memory controller which maps the other range. Otherwise, you would end up with * controllers, which is (also) unrealistic.

Well also having AddrRanges with holes is not supported :) 
It's just a matter of implementing things.

I am just scared of complicating such an important class like AddrRange which IMHO it should be as simple as possible. We should (but I understand I might not be getting the full picture) either use a list/vector of ranges to model holes, or to push the complexity into a new class

Will-Shaddix:
This comment is to document why work on this has stopped in case I or someone else down the line tries to take a crack at this problem.
   
Issues with the proposed solution:
- While this allows it to simulate without issue, there is wasted memory capacity in the hole because the host still saves memory for the hole. For example, if we have an address range of 0 - 8GiB with a hole from the 3GiB - 4GiB range, it would still reserve a 8GiB range on the host, despite only needing 7GiB. 
- Some places within gem5 assume there isn't a hole and will just use the start and size of an AddrRange. One example of this is restoring from a checkpoint, this makes restoring from a checkpoint unsupported with these address holes, furthermore this gets more complicated when we think of how we would store the memory into a memory image file
- At least for the checkpointing situation, there are potential workarounds such as removing the hole when taking the checkpoint as in create the memory image with the hole filled in with 0's then add the hole back after restoring the image. However we deemed this too "hacky" for such an integral part of gem5.
- TLDR; The AddrRange is used for a wide variety of things making it extremely difficult to find a fix all cure which  doesn't break other parts of the code where AddrRange is used for a different purpose.

BobbyRBruce:
Closing this PR as work has stopped for now.

PR #1583, opened 2024-09-20
arch-arm,util-m5: Change arm64's default m5 call type to addr

nkrim:
As of PR #977, gem5 has a defined M5OPS_ADDR for arm64, even if it is constrained to certain conditions. With that change and the arm64 board KVM support (#725), it seems like interaction with the m5 utility under arm64 will most commonly occur in KVM -- where instruction-mode does not work -- and thus address-mode becomes more desirable as the default.

This also makes m5's behavior in arm64 consistent with x86, the only other architecture that supports address-mode operations.

### DIFF
--- a/util/m5/src/abi/arm64/SConsopts
+++ b/util/m5/src/abi/arm64/SConsopts
@@ -29,8 +29,8 @@ env['ABI'] = 'arm64'
 get_abi_opt('CROSS_COMPILE', 'aarch64-linux-gnu-')
 get_abi_opt('QEMU_ARCH', 'aarch64')
 
-env['CALL_TYPE']['inst'].impl('m5op.S', 'verify_inst.cc', default=True)
-env['CALL_TYPE']['addr'].impl('m5op_addr.S')
+env['CALL_TYPE']['inst'].impl('m5op.S', 'verify_inst.cc')
+env['CALL_TYPE']['addr'].impl('m5op_addr.S', default=True)
 env['CALL_TYPE']['semi'].impl('m5op_semi.S', 'verify_semi.cc')
 
 env.Append(CXXFLAGS='-DM5OP_ADDR=0x10010000')


BobbyRBruce (dismissed):
Makes sense to me! Thanks for the contribution.

hnpl:
@BobbyRBruce The current way we do m5 ops via address requires opening /dev/mem, which requires root access. Is it true that in all gem5 disk images, all commands are called as the root user? I just want to make sure that this won't break any gem5 disk images.

nkrim:
> @BobbyRBruce The current way we do m5 ops via address requires opening /dev/mem, which requires root access. Is it true that in all gem5 disk images, all commands are called as the root user? I just want to make sure that this won't break any gem5 disk images.

nkrim:
> Is it true that in all gem5 disk images, all commands are called as the root user? I just want to make sure that this won't break any gem5 disk images.

Yes, from what I've seen the m5 executable has setuid permissions to run as root, so this shouldn't affect the use of m5 unless the configuration just does not work with address mode for whatever reason.

powerjg:
> The current way we do m5 ops via address requires opening /dev/mem, which requires root access. Is it true that in all gem5 disk images, all commands are called as the root user? I just want to make sure that this won't break any gem5 disk images.

This is a good point. However, this should only affect newly created images. We are working to improve and make the images more consistent.

powerjg:
@giactra do you have any objection to this change?

giactra:
> @giactra do you have any objection to this change?

In theory it looks like a good improvement. My only concern is that  IIRC we only setup m5ops_base in FS mode, and not in SE... Whereas instruction based is always on.. Unless I am missing something. Can we test this in SE mode as well?

Note: referring to Arm but it might be true to other ISAs

powerjg:
> Can we test this in SE mode as well?

Yes, great point! We should look into this in SE mode.

BobbyRBruce:
> > Can we test this in SE mode as well?
> 
> Yes, great point! We should look into this in SE mode.

Would a simple binary calling m5exit be sufficient to prove this works? I could make that.

nkrim:
> > > Can we test this in SE mode as well?
> > 
> > 
> > Yes, great point! We should look into this in SE mode.
> 
> Would a simple binary calling m5exit be sufficient to prove this works? I could make that.

There is an x86 test here https://github.com/gem5/gem5/tree/stable/tests/gem5/m5_util that does exactly that, so can probably be easily repurposed for ARM.

At the same time, this change should only affect the behavior of the executable (which I believe only makes sense to use in FS), not the linked functions themselves since those are explicit about the mode being used. That said, probably better safe than sorry.

nkrim:
@BobbyRBruce I was looking into reworking the x86 test linked above for ARM, but I am a bit confused about how the binary for that test is created. I see this test program here https://github.com/gem5/gem5/tree/stable/tests/test-progs/m5-exit but these Makefiles are out of date and don't produce the exact name `x86-m5-exit`.

I see that this binary is on gem5 resources though but not an ARM compilation, so can an ARM binary be added or is there a way to have a local resource as a part of the test?

Harshil2107:
> @BobbyRBruce I was looking into reworking the x86 test linked above for ARM, but I am a bit confused about how the binary for that test is created. I see this test program here https://github.com/gem5/gem5/tree/stable/tests/test-progs/m5-exit but these Makefiles are out of date and don't produce the exact name `x86-m5-exit`.
> 
> I see that this binary is on gem5 resources though but not an ARM compilation, so can an ARM binary be added or is there a way to have a local resource as a part of the test?

The m5-exit tests in the test-progs are outdated. The latest binary code is at https://github.com/gem5/gem5-resources/blob/stable/src/deprecated/simple/m5_exit_addr.c and https://github.com/gem5/gem5-resources/blob/stable/src/deprecated/simple/m5_exit.c

We can upload an arm m5 exit binary to gem5 resources and thoat should work for the tests.

BobbyRBruce:
I've had a go at creating tests for this, but I don't know much about the m5-exit utility and how this works. I created some m5-exit binaries and some tests, which you can find in #1705, but when incorporated into tests and run, I run into a couple of failures.

The tests are failing whether the change is applied to the develop branch or this PR, so it's probably not anything to do with this change, though without fixing the tests I don't think I can verify this PR.  I could use some help with this; I'm not entirely sure how I put these tests together to verify this functionality. If anyone could look over what I've made and give suggestions, it'd be much appreciated.

BobbyRBruce:
Ok, my previous comments can be discarded. Tt took me a while to get my bearings on this but, in short, this change doesn't break anything. This default applies only to the compiled m5util binary which you don't use in SE mode. It's for executing in a FS simulated OS shell. In the SE mode case everything remains the same as before. When you compile an m5op in a binary  you have to explicitly state wither you're doing an address or "magic instruction" m5opt, there's no option to just go with a default option between the two.

BobbyRBruce (approved):

PR #1587, opened 2024-09-21
scons: Fix scons 'readCommand' non-zero exits

BobbyRBruce:
There appears to have been an assumption here that `Popen` would raise an exception if the command run returned non-zero. This is not the case. This commit fixes this by obtaining the return code and throwing an exception if it is non-zero.

This bug caused some minor issues as Exception handling code to handle the non-zero case elsewhere in Scons was never executed.


Harshil2107 (approved):
lgtm

PR #1594, opened 2024-09-23
gpu-compute: Fix '64kB' to '64KiB' in gpu-compute

BobbyRBruce:
64kB, in these cases, will cast to 64KiB regardless. To improve readability and understanding of these objects, this patch changes there SI Prefix (kB -> KiB).


mattsinc (approved):
Thanks @BobbyRBruce , I did not realize there was a distinction between them.

PR #1600, opened 2024-09-24
stdlib: Modify X86DemoBoard

erin-le:
This commit modifies X86DemoBoard so it has numbers more similar to that of RiscvDemoBoard and ArmDemoBoard. It also adds SE mode to X86DemoBoard. Note that the changes here depend on the changes in PR 1579.


src/python/gem5/prebuilt/demo/x86_demo_board.py
```diff
@@ -77,23 +83,68 @@ def __init__(self):
...
-            l2_assoc=16,
-            num_l2_banks=1,
+
+        cache_hierarchy = PrivateL1SharedL2CacheHierarchy(
+            l1d_size="64KiB", l1i_size="64KiB", l2_size="1MiB"
         )
 
         super().__init__(
-            clk_freq="3GHz",
+            clk_freq="1.4GHz",
```
powerjg, line 98: This frequency seems low and it's going to make simulations take longer. 3GHz is pretty reasonable.

src/python/gem5/prebuilt/demo/x86_demo_board.py
```diff
@@ -77,23 +83,68 @@ def __init__(self):
...
             processor=processor,
             memory=memory,
             cache_hierarchy=cache_hierarchy,
         )
+
+    # Copied over and adapted from RISCVMatchedBoard and X86Board. We need to
+    # override these functions to get SE mode to work.
```
powerjg, line 105: This seems like a good comment to have the commit message, but won't make sense in the context of the code in the future.

src/python/gem5/prebuilt/demo/x86_demo_board.py
BobbyRBruce: This is mostly to keep it similar the riscv and arm demo boards so they are somewhat comparable. a 3GHz RISC-V, for example, seems unrealistic

powerjg: The frequency depends on the microarchitecture, not the architecture (ISA). 1.4 GHz is too slow for modern devices.

BobbyRBruce: I meant more that i felt the boards demo boards should be similar rather than individually true-ish representations of what an X86/RISCV/ARM system is. I thought this would make it easier so we can compare their performance fairly.

I'm not against making this more realistic I just wanted you to know this makes the demoboards less alike. I'll make a patch putting this at 3GHz

BobbyRBruce:
@erin-le : I can't add commits to your PRs because your working on a fork of one of Zhantong's repos (think this is the reason why i can't add commits anyway). Could you try to work on a fork from gem5/gem5?

It's not a big problem but I i had to make changes to this in a separate PR: #1618 . I'll close this #PR to just work on #1618.

PR #1605, opened 2024-09-24
tests, configs, util, mem, python, systemc: Change base 10 units to base 2

erin-le:
This commit changes metric units (e.g. kB, MB, and GB) to binary units (KiB, MiB, GiB) in various files. This PR covers files that were missed by a previous PR that also made these changes.


tests/gem5/memory/test.py
```diff
@@ -41,10 +41,10 @@
 )
 
 simple_mem_params = [
-    ("inf-bandwidth", {"bandwidth": "0GB/s"}),
+    ("inf-bandwidth", {"bandwidth": "0GiB/s"}),
```
powerjg, line 44: I'm not certain these bandwidths should be changed. I know the reported bandwidth is in MB/s (power of 10).

The question is: Does the change you are making here have a functional difference?

tests/gem5/traffic_gen/configs/simple_traffic_run.py
```diff
@@ -53,7 +53,7 @@ def generator_factory(
 
         return LinearGenerator(
             duration="250us",
-            rate="40GB/s",
+            rate="40GiB/s",
```
powerjg, line 56: Ditto here for bandwidth

util/gem5img.py
powerjg, line 1: We should delete this file (in another PR)

util/plot_dram/dram_lat_mem_rd_plot.py
```diff
@@ -136,9 +136,9 @@ def is_pow2(num):
     xticks_labels = []
     for x in xticks_locations:
         if x < 1024:
-            xticks_labels.append("%d kB" % x)
+            xticks_labels.append("%d KiB" % x)
         else:
-            xticks_labels.append("%d MB" % (x / 1024))
+            xticks_labels.append("%d MiB" % (x / 1024))
```
powerjg, line 141: This should be further verified with the stats generation code in gem5. It might be right, but it could also be wrong.

powerjg (commented):
Thanks for doing this!

tests/gem5/memory/test.py
erin-le: I looked into this and I think it should be fine to change. `("inf-bandwidth", {"bandwidth": "0GiB/s"}),` and `("low-bandwidth", {"bandwidth": "1MiB/s"}),` are passed into `tests/gem5/memory/simple-run.py` as bandwidth arguments, which are then used in `class MyMem(SimpleMemory)`. In SimpleMemory.py, bandwidth uses `Param.MemoryBandwidth()`, which calls `convert.toMemoryBandwidth()`. 

`convert.toMemoryBandwidth` uses `toBinaryFloat()`, which uses `binary_prefixes` in `convert.py`, which means that it converts base 10 prefixes to base 2 numbers.

tests/gem5/traffic_gen/configs/simple_traffic_run.py
erin-le: I think this should also be okay to change, since LinearGeneratorCore uses `toMemoryBandwidth()`, which converts base 10 to base 2. RandomGeneratorCore also uses `toMemoryBandwidth()`.

util/plot_dram/dram_lat_mem_rd_plot.py
erin-le: I looked further into this one and I think it should be okay to change. This script uses outputs from `configs/dram/lat_mem_rd.py`, which (among other things) generates base 2 memory range values. These values are printed into simout.txt, read in by `dram_lat_mem_rd_plot.py`, and ultimately used in xticks_labels.

powerjg (approved):
Thanks for looking into my questions!

PR #1609, opened 2024-09-26
mem-cache: Helper functions to allow dynamic configuration of partitioning policies

giactra:
This PR is doing a simple refactoring of some partitioning policies. It moves existing functionalities
within PP methods so that they can be called multiple times throughout the simulation.
Therefore allowing a dynamic adjustment of the partitioning scheme

### DIFF
--- a/src/mem/cache/tags/partitioning_policies/max_capacity_pp.cc
+++ b/src/mem/cache/tags/partitioning_policies/max_capacity_pp.cc
@@ -52,8 +52,9 @@ namespace partitioning_policy
 
 MaxCapacityPartitioningPolicy::MaxCapacityPartitioningPolicy
     (const MaxCapacityPartitioningPolicyParams &params):
-    BasePartitioningPolicy(params), cacheSize(params.cache_size),
-    blkSize(params.blk_size), partitionIDs(params.partition_ids),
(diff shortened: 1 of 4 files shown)


src/mem/cache/tags/partitioning_policies/max_capacity_pp.cc
```diff
@@ -62,32 +63,35 @@ MaxCapacityPartitioningPolicy::MaxCapacityPartitioningPolicy
...
+        configurePartition(partition_id, cap_frac);
     }
 }
 
+void
+MaxCapacityPartitioningPolicy::configurePartition(uint64_t partition_id,
+                                                  double cap_frac)
+{
+    // check Capacity Fraction (cap_frac) is actually a fraction in [0,1]
+    panic_if(!(cap_frac >= 0 && cap_frac <= 1),
```
powerjg, line 81: I think this should be `fatal_if` since the user provides the `cap_frac`

giactra: So far you are right, but the purpose behind this PR is actually to allow guest SW to change the cap_frac at runtime (by programming a partitioning controller)

powerjg (approved):

Issue #1443, opened 2024-08-09
Stats Resetting


studyztp:
Currently, there does not exist a documentation or clear guidance about what stats in the stats.txt file that can be reset by the `m5.stats.reset()` function and what stats cannot.
For example, the stats `simInsts` does not reset when calling `m5.stats.reset()`.

BobbyRBruce:
Shouldn't it reset?

I'm wondering  if you're suggesting we document which stats are, and which are not, wiped, or if there are stats that should be reset but aren't.

My feeling is `m5.stats.reset()` should reset _all_ stats... no?

studyztp:
Sadly, `m5.stats.reset()` is not resetting all the stats in the stats file. 
I attached a simple run script that can show that (it is modified from [arm-hello.py](https://github.com/gem5/gem5/blob/stable/configs/example/gem5_library/arm-hello.py)). This script exits the simulation after simulating the workload for 143293729 Ticks, then dump and reset the stats using `m5.stats.reset()` and `m5.stats.dump()`. It then falls back to the simulation for only 1 Ticks then dump the stats before exit. The stats file should suggest that `simInsts` did not get reset.

We've been noticing `simInsts` is not being reset, but maybe there are more. 
I suggest testing out what stats in the stats file that can or cannot be reset first, then depending on what stats are actually having issue, we can decide if we should document them, wipe them, or fix them.
If it's only `simInsts` that is being an issue, I think we can fix it or add the warning in its stat description. 

```
from gem5.components.boards.simple_board import SimpleBoard
from gem5.components.cachehierarchies.classic.no_cache import NoCache
from gem5.components.memory import SingleChannelDDR3_1600
from gem5.components.processors.cpu_types import CPUTypes
from gem5.components.processors.simple_processor import SimpleProcessor
from gem5.isas import ISA
from gem5.resources.resource import obtain_resource
from gem5.simulate.simulator import Simulator
from gem5.utils.requires import requires
import m5

# This check ensures the gem5 binary is compiled to the ARM ISA target. If not,
# an exception will be thrown.
requires(isa_required=ISA.ARM)

# In this setup we don't have a cache. `NoCache` can be used for such setups.
cache_hierarchy = NoCache()

# We use a single channel DDR3_1600 memory system
memory = SingleChannelDDR3_1600(size="32MB")

# We use a simple Timing processor with one core.
[...]

PR #1615, opened 2024-09-27
cpu: fix simInsts and simOps not resetting

erin-le:
This PR fixes the bug where simInsts and simOps don't reset when m5.stats.reset() is called. The stats hostInstRate and hostOpRate are affected by this change as well, as they depend on simInsts and simOps respectively.

This is related to issue 1443 linked [here](https://github.com/gem5/gem5/issues/1443).

### DIFF
--- a/src/cpu/base.hh
+++ b/src/cpu/base.hh
@@ -156,6 +156,30 @@ class BaseCPU : public ClockedObject
 
         statistics::Formula hostInstRate;
         statistics::Formula hostOpRate;
+
+        Counter previousInsts = 0;
+        Counter previousOps = 0;
+
+        static Counter
+        numSimulatedInsts()
+        {
+            return totalNumSimulatedInsts() - (globalStats->previousInsts);
+        }
+
+        static Counter
+        numSimulatedOps()
+        {
+            return totalNumSimulatedOps() - (globalStats->previousOps);
+        }
+
+        void
+        resetStats() override
+        {
+            previousInsts = totalNumSimulatedInsts();
+            previousOps = totalNumSimulatedOps();
+
+            statistics::Group::resetStats();
+        }
     };
 
     /**
@@ -609,7 +633,7 @@ class BaseCPU : public ClockedObject
 
     static int numSimulatedCPUs() { return cpuList.size(); }
     static Counter
-    numSimulatedInsts()
+    totalNumSimulatedInsts()
     {
         Counter total = 0;
 
@@ -621,7 +645,7 @@ class BaseCPU : public ClockedObject
     }
 
     static Counter
-    numSimulatedOps()
+    totalNumSimulatedOps()
     {
         Counter total = 0;
 
--- a/src/cpu/base.cc
+++ b/src/cpu/base.cc
@@ -855,13 +855,13 @@ BaseCPU::GlobalStats::GlobalStats(statistics::Group *parent)
              "Simulator op (including micro ops) rate (op/s)")
 {
     simInsts
-        .functor(BaseCPU::numSimulatedInsts)
+        .functor(BaseCPU::GlobalStats::numSimulatedInsts)
         .precision(0)
         .prereq(simInsts)
         ;
 
     simOps
-        .functor(BaseCPU::numSimulatedOps)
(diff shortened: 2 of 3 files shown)


src/cpu/base.hh
```diff
@@ -149,13 +149,24 @@ class BaseCPU : public ClockedObject
     /** Global CPU statistics that are merged into the Root object. */
     struct GlobalStats : public statistics::Group
     {
+        void resetStats() override
```
giactra: Nit:

```
void
resetStats() override
```

src/cpu/base.hh
```diff
@@ -149,13 +149,24 @@ class BaseCPU : public ClockedObject
...
+        {
+          previousInsts = totalNumSimulatedInsts();
+          previousOps = totalNumSimulatedOps();
+
+          statistics::Group::resetStats();
```
giactra: requires 4 whitespaces for indentation

src/cpu/base.hh
```diff
@@ -621,7 +632,13 @@ class BaseCPU : public ClockedObject
     }
 
     static Counter
-    numSimulatedOps()
+    numSimulatedInsts()
+    {
+      return totalNumSimulatedInsts() - (globalStats -> previousInsts);
+    }
```
giactra: Shouldn't we make this a method of GlobalStats? In this way we avoid extra-stat usage of it

src/cpu/base.hh
```diff
@@ -632,6 +649,12 @@ class BaseCPU : public ClockedObject
         return total;
     }
 
+    static Counter
+    numSimulatedOps()
+    {
+      return totalNumSimulatedOps() - (globalStats -> previousOps);
+    }
```
giactra: Ditto

giactra (changes requested):
Hi @erin-le thanks for this fix... I am not a huge fan of caching the previous reset value, as caching lends itself to staleness. However I realise the alternative approach would be to duplicate data structures for the stats/non-stats case. So I am happy to fix it this way. Anyway I have some comments; feel free to review them when you have some time

src/cpu/base.hh
```diff
@@ -149,13 +149,37 @@ class BaseCPU : public ClockedObject
     /** Global CPU statistics that are merged into the Root object. */
     struct GlobalStats : public statistics::Group
     {
+        void
+        resetStats() override
```
giactra: Nit: would you mind grouping all methods together?

src/cpu/base.hh
```diff
@@ -149,13 +149,37 @@ class BaseCPU : public ClockedObject
...
         statistics::Formula hostInstRate;
         statistics::Formula hostOpRate;
+
+        Counter previousInsts = 0;
+        Counter previousOps = 0;
+
+        static Counter
+        numSimulatedInsts()
+        {
+            return totalNumSimulatedInsts() - (globalStats -> previousInsts);
```
giactra: There shouldn't be whitepaces around the -> operator

src/cpu/base.hh
```diff
@@ -149,13 +149,37 @@ class BaseCPU : public ClockedObject
...
+        }
+
+        static Counter
+        numSimulatedOps()
+        {
+            return totalNumSimulatedOps() - (globalStats -> previousOps);
```
giactra: Ditto

giactra (changes requested):
Thanks for your changes, I have some other minor comments (mainly coding style related). Would you mind squashing all commits together and force pushing the PR as a single commit?

powerjg (commented):
lgtm, after addressing Giacomo's requests

mattsinc:
While it contributes nothing to the solution discussion, I just wanted to add that some students in my class this semester seemingly had run into the same problem -- so I'm glad @erin-le was already fixing it :)

giactra (approved):
LGTM, thanks!

powerjg (approved):

powerjg:
@erin-le Can you make sure to add something big and bold in the release notes about this change? While this is *fixing* an incorrect behavior, it is making a major change to behavior that people need to be aware of in the next release.

PR #1617, opened 2024-10-01
arch-riscv: Implement CLINT reset feature

rogerchang23424:
When reset, registers are change
msip: cleared to zero
mtimecmp: unknown state, cloud be origin values or change to any values mtime: cleared to zero

Spec:
https://github.com/riscv/riscv-aclint/blob/main/riscv-aclint.adoc https://sifive.cdn.prismic.io/sifive/1a82e600-1f93-4f41-b2d8-86ed8b16acba_fu740-c000-manual-v1p6.pdf

Change-Id: I3c50b41eb765ad9cd7a8a03c427bd0011195de5c

### DIFF
--- a/src/dev/riscv/Clint.py
+++ b/src/dev/riscv/Clint.py
@@ -35,6 +35,7 @@
 
 from m5.objects.Device import BasicPioDevice
 from m5.objects.IntPin import IntSinkPin
+from m5.objects.ResetPort import ResetResponsePort
 from m5.params import *
 from m5.proxy import *
 from m5.util.fdthelper import *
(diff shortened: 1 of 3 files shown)


robhau:
Hi @rogerchang23424,

instead of manually reset every register, I think we should use the "Reset API" of `RegisterBank`. It enables us to register reset functions for each register.

https://github.com/gem5/gem5/blob/0792d94b6f2caa0dc3605d38eb5da385ae2aca14/src/dev/reg_bank.hh#L762-L780

A reset function receives a reference to the register that it has to reset.
https://github.com/gem5/gem5/blob/0792d94b6f2caa0dc3605d38eb5da385ae2aca14/src/dev/reg_bank.hh#L587-L587

The default reset function changes a register back to its initial value.
https://github.com/gem5/gem5/blob/0792d94b6f2caa0dc3605d38eb5da385ae2aca14/src/dev/reg_bank.hh#L622-L626


All registers can then be reset by calling `registers.reset()`.

https://github.com/gem5/gem5/blob/0792d94b6f2caa0dc3605d38eb5da385ae2aca14/src/dev/reg_bank.hh#L1120-L1126

rogerchang23424:
Updated the to reset API

Harshil2107 (approved):
Not an expert but it looks like all the requested changes were addressed.

powerjg (approved):

PR #1579, opened 2024-09-18
python: infer fs/se mode from workload for boards with both

erin-le:
This commit modifies abstract_board.py so that boards that support both SE and FS mode (have both set_se_binary_workload and set_kernel_disk_workload) can infer which to use based on the workload that is set.

PR #1618, opened 2024-10-01
stdlib,arch-x86: Update X86Demoboard

BobbyRBruce:
This commit modifies X86DemoBoard so it has numbers more similar to that of RiscvDemoBoard and ArmDemoBoard. It also adds SE mode to X86DemoBoard. Note that the changes here depend on the changes in PR 1579.

**Note**: This PR was created so @BobbyRBruce could add his commits to #1600

### DIFF
--- a/src/python/gem5/prebuilt/demo/x86_demo_board.py
+++ b/src/python/gem5/prebuilt/demo/x86_demo_board.py
@@ -48,10 +48,9 @@ from ...utils.requires import requires
 class X86DemoBoard(X86Board, SEBinaryWorkload):
     """
     This prebuilt X86 board is used for demonstration purposes. It simulates
-    an X86 3GHz quad-core system with a 2GiB DDR3_1600 memory system. A
-    MESI_Two_Level cache hierarchy is set with an l1 data and instruction
-    cache, each 32KiB with an associativity of 8, and a single bank l2 cache of
-    1MiB with an associativity of 16.
+    an X86 3GHz dual-core system with a 3GiB DDR4_2400 memory system. The
+    cache hierarchy consists of per-core private L1 instruction and data
+    caches (64KiB each) connected to a shared 1MiB L2 cache.
 
     **DISCLAIMER**: This board is solely for demonstration purposes. This board
     is not known to be representative of any real-world system or produce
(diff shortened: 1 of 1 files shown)


src/python/gem5/prebuilt/demo/x86_demo_board.py
```diff
@@ -77,18 +82,15 @@ def __init__(self):
...
-            l1d_size="32KiB",
-            l1d_assoc=8,
-            l1i_size="32KiB",
-            l1i_assoc=8,
-            l2_size="1MiB",
-            l2_assoc=16,
-            num_l2_banks=1,
+
+        cache_hierarchy = PrivateL1SharedL2CacheHierarchy(
+            l1d_size="64KiB", l1i_size="64KiB", l2_size="1MiB"
```
powerjg: Can we make this shared caches 8 MiB? 1 MiB is very small by today's standards. (This will also make the simulations take less time)

src/python/gem5/prebuilt/demo/x86_demo_board.py
```diff
@@ -24,27 +24,33 @@
...
+    X86FsLinux,
+)
 from m5.util import warn
 
-from ...coherence_protocol import CoherenceProtocol
+from ...components.boards.se_binary_workload import SEBinaryWorkload
 from ...components.boards.x86_board import X86Board
-from ...components.cachehierarchies.ruby.mesi_two_level_cache_hierarchy import (
-    MESITwoLevelCacheHierarchy,
+from ...components.cachehierarchies.classic.private_l1_shared_l2_cache_hierarchy import (
```
powerjg, line 37: I'm a bit concerned about classic caches + x86 FS not working. If we can boot all of our linux distros, then it's probably OK

BobbyRBruce: I don't think I understand this comment. As far as I know classic caches work fine with X86 FS. Is this not the case?

powerjg: Historically, classic caches and x86 have not worked because classic caches didn't implement x86-style atomics. There have been improvements, but we should test booting to ensure that it works with multiple CPUs.

erin-le: I ran the Ubuntu 22.04 and 24.04 no systemd workloads with X86DemoBoard and they finished successfully.

erin-le (approved):
LGTM. However, AbstractBoard still needs to be changed (discussed [here](https://github.com/gem5/gem5/pull/1579)) for this board to work.

PR #1619, opened 2024-10-01
configs: Deprecate Vega10

abmerop:
Vega10 is no longer officially supported by ROCm and ROCm is starting to use some packet types not supported. These were originally kept to allow users to use older disk images with newer gem5. Going forward the gem5 version and gem5-resources releases will be required to be the same to prevent lingering old configs.

As a replacement for vega10*.py, mi300.py or mi200.py should be used. HIP examples, cookbook, and rodinia configs can be replaced with the standard flow of building / obtaining the GPU application and running using mi300.py or mi200.py as they do not require any input options and therefore do not require changes to the disk image.


mattsinc (approved):
End of an era

Issue #1204, opened 2024-06-06
RISCV SE mode: pthread_create() leads to a problem with clone3()


DiaZixX:
**Describe the bug**
I want to do multi-threading in SE mode on RISCV. To isolate the problem, I'm executing a statically compiled binary that creates a thread (I've disassembled the binary to check that all the labels are present). Using Exec debug-flag, I discover that when clone3() calls ecall, a fatal is raised: `src/sim/syscall_desc.hh:209: fatal: Syscall 435 out of range` and the thread cannot be created.

**Affects version**
I'm on version 23.1.0.0 (The latest)

**gem5 Modifications**
No modification.

**To Reproduce**
Steps to reproduce the behavior. Please assume starting from a clean repository:

1. Compile gem5 with command : scons build/RISCV/gem5.opt
2. Execute the simulation with this python script (using option --nCPU 2):
```python

import m5

m5.util.addToPath("../../")
from m5.objects import *
from caches import *

import argparse

parser = argparse.ArgumentParser(
    description="A simple system with 2-level cache to execute binary riscv code."
)
parser.add_argument(
    "binary",
    default="",
    nargs="?",
    type=str,
    help="Path to the binary to execute."
)

#########################################################################################
#                                        CACHES                                         #
#########################################################################################

# L1 Caches
parser.add_argument(
    "--l1i-size", help="L1 inscrution cache size. Default value: 16kB."
)
parser.add_argument(
    "--l1i-assoc", help="L1 instruction cache associativity. Default value: 8."
)
parser.add_argument(
    "--l1d-size", help="L1 data cache size. Default value: 16kB."
)
parser.add_argument(
    "--l1d-assoc", help="L1 data cache associativity. Default value: 8."
)

# L2 Caches
parser.add_argument(
    "--l2-cache", action="store_true", help="Enable L2 cache"
)
parser.add_argument(
    "--l2-size", help="L2 cache size. Default value: 64kB."
)
parser.add_argument(
    "--l2-assoc", help="L2 cache associativity. Default value: 16"
)
parser.add_argument(
    "--l2-num-cache",
    default=1,
    type=int,
    help="Number of L2 cache. Need to divide the number of CPU. Default value: 1"
)
parser.add_argument(
    "--l2-clusivity",
    help="Clusivity of L2 caches. Default mode: mostly inclusive.",
    choices=["mostly_incl", "mostly_excl"]
)
# parser.add_argument(
#     "--l2-writeback-clean",
[...]

PR #1620, opened 2024-10-02
arch-riscv: Enable clone3 syscall in riscv64

tommarin:
The clone3 syscall, implemented in commit 87e774c, is currently only handled for x86-64 in gem5 SE mode. Clone3 is employed by modern glibc versions instead of clone for processes/threads generation (e.g. issue #1204). This commit enables the clone3 syscall in riscv64 by adding the corresponding handler call, as well as its arguments struct.


powerjg (dismissed):

powerjg (approved):
I'll re-run the failing test when the others finish. It looks like it was a random error.

tommarin:
clang-fast-compilation failed again. I think it hits a 90 mins time limit...

BobbyRBruce (approved):

PR #1621, opened 2024-10-02
dev-amdgpu: Use GPU specific cache line size

abmerop:
Invalidate requests align to system cache line size. This causes problems if the GPU cache hierarchy's cache line size is different than the system as the unlaigned requests never return, leading to deadlock on deferred dispatch.

This commit uses the cache line size from the GPU memory manager and makes the cache line size there non-optional.

Tested with multiple RubySystems where CPU side was 64B and GPU side was 128B cache lines.

### DIFF
--- a/configs/example/gpufs/system/system.py
+++ b/configs/example/gpufs/system/system.py
@@ -199,7 +199,7 @@ def makeGpuFSSystem(args):
     system.pc.south_bridge.gpu.pm4_pkt_procs = pm4_procs
 
     # GPU data path
-    gpu_mem_mgr = AMDGPUMemoryManager()
+    gpu_mem_mgr = AMDGPUMemoryManager(cache_line_size=args.cacheline_size)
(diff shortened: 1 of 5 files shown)


mattsinc (approved):
I'm not sure I understand how this interacts with Ruby, where the cache lines actually are, but I don't see anything wrong in this commit either ...

powerjg (commented):
Seems OK to me. I wonder if there's any way to check and `fatal` if the cache line sizes don't match.

abmerop:
I don't want them to match.  I want to have the CPU RubySystem be 64B and GPU RubySystem be 128B

powerjg:
Apologies for the confusion (or maybe I'm missing something). That's not what I meant. I mean that the GPU RubySystem's cache line size and the cache line size in the `AMDGPUMemoryManager`

abmerop:
I see.  Right now I guess we assume the user's config set them to be the same.  Not sure how you would verify it is the same everywhere

PR #1625, opened 2024-10-03
python: clarify SimObject error message

erin-le:
This adds more detail to the error message that is thrown when an orphan node is instantiated.


powerjg (approved):
Better error messages are always good!

giactra (approved):

BobbyRBruce (approved):

PR #1627, opened 2024-10-04
util-docker: Fix gpu dpcker images

BobbyRBruce:
Two faults:

1. You can't give description the docker-bake file for single platform builds. They must be in the Dockerfile..
2. The gpu docker image def in docker-bake.hcl was not overriding the "common" setttings as previously thought. This was causing builds to something build the wrong platform and vairous other weird bugs. This has been fixed in this patch.


mattsinc:
Just adding myself and @abmerop for future reference

PR #1628, opened 2024-10-04
misc,tests: Increase Weekly and Daily GPU test timeout

BobbyRBruce:
The Weekly GPU tests are failing due to a timeout, but I found the testing timeout was set to 5 hours, and  we have been frequently close to reaching this but have recently changed the test enough to consistently go over.

 The main two things that appear to have caused this are:

~~1. Moving the X86_VEGA compilation into the same step as the running of the tests.~~ (I take this back, the timeout is per-job, it shouldn't matter how stuff is deivided among steps in the job. However, keeping it separate does no harm and merging the two steps did coincide with failures occurring. I'll play it safe for now_.
2. Reducing the number of threads per GitHub Actions runner, thus slowing job execution.

In addition, we've added more tests to this weekly GPU suite, though I don't believe we have got to running these tests yet. The timeout appears to always have been triggered before this.

This PR increases the timeout to 3 days and moves the compilation into a separate step.

**Update: Same changes done for Daily tests too as it appears to be the same problem.

### DIFF
--- a/.github/workflows/daily-tests.yaml
+++ b/.github/workflows/daily-tests.yaml
@@ -112,7 +112,7 @@ jobs:
     gpu-tests:
         runs-on: [self-hosted, linux, x64]
         container: ghcr.io/gem5/gcn-gpu:latest
-        timeout-minutes: 300
+        timeout-minutes: 720 # 12 hours
 
         steps:
             - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


BobbyRBruce:
Sorry for the force merge but I just want a quick turn around on this and running the tests is the only way to really know. Also, the CI tests here don't touch this code, so a waste of time.

mattsinc (commented):
This looks good to me.

PR #1584, opened 2024-09-20
tests: Add Pannotia GPU Tests

erin-le:
This PR adds the Pannotia GPU tests. The input graphs to the binaries were too large to be committed, so they need to be obtained separately. Instructions for how to do so are in README.md.

weekly-tests.yaml was modified to download the input files, but I am not sure if it will work.

PR #1631, opened 2024-10-04
tests: update input sizes for pannotia tests

Harshil2107:
This PR addresses comments from #1584 

- removed tests using the same binary multiple times. Each binary is tested once with one graph
- Updated the input sizes as per the comments in the above mentioned PR


Harshil2107:
I am still running these tests locally.

mattsinc (dismissed):
As far as I can tell this integrates all my suggestions -- thanks!  Let me know if the MIS failure persists.

powerjg:
I'm re-running these jobs... @Harshil2107 have you looked into why there's an error?

Harshil2107:
I will look into why the tests are failing.

Harshil2107:
@mattsinc The mis tests seem to be failing still. 

This is the output from simerr.txt
```
warn: system.ruby.network adopting orphan SimObject param 'ext_links'
warn: system.ruby.network adopting orphan SimObject param 'int_links'
src/mem/dram_interface.cc:690: warn: DRAM device capacity (8192 Mbytes) does not match the address range assigned (65536 Mbytes)
mmap: Cannot allocate memory
src/mem/physical.cc:247: fatal: Could not mmap 68719476736 bytes for range [0:0x1000000000]!
Memory Usage: 151096 KBytes
```

This is the output from simout.txt

```

gem5 version DEVELOP-FOR-24.1
gem5 compiled Oct  6 2024 20:01:04
gem5 started Oct  8 2024 18:57:14
gem5 executing on b2650aeccb0d, pid 42
command line: /gem5/build/VEGA_X86/gem5.opt -d /tmp/gem5out8c12lgvi -re --silent-redirect /gem5/configs/example/apu_se.py -n3 --mem-size=64GiB -c /gem5/tests/gem5/gpu/../gpu-pannotia-resources/pannotia-bins/mis_hip.gem5 --options '/gem5/tests/gem5/gpu/../gpu-pannotia-resources/pannotia-datasets/1k_128k.gr 0'

Num SQC =  1 Num scalar caches =  1 Num CU =  4
```

BobbyRBruce:
Is this waiting on something?

mattsinc:
> @mattsinc The mis tests seem to be failing still.
> 
> This is the output from simerr.txt
> 
> ```
> warn: system.ruby.network adopting orphan SimObject param 'ext_links'
> warn: system.ruby.network adopting orphan SimObject param 'int_links'
> src/mem/dram_interface.cc:690: warn: DRAM device capacity (8192 Mbytes) does not match the address range assigned (65536 Mbytes)
> mmap: Cannot allocate memory
> src/mem/physical.cc:247: fatal: Could not mmap 68719476736 bytes for range [0:0x1000000000]!
> Memory Usage: 151096 KBytes
> ```
> 
> This is the output from simout.txt
> 
> ```
> gem5 Simulator System.  https://www.gem5.org
> gem5 is copyrighted software; use the --copyright option for details.
> 
> gem5 version DEVELOP-FOR-24.1
> gem5 compiled Oct  6 2024 20:01:04
> gem5 started Oct  8 2024 18:57:14
> gem5 executing on b2650aeccb0d, pid 42
> command line: /gem5/build/VEGA_X86/gem5.opt -d /tmp/gem5out8c12lgvi -re --silent-redirect /gem5/configs/example/apu_se.py -n3 --mem-size=64GiB -c /gem5/tests/gem5/gpu/../gpu-pannotia-resources/pannotia-bins/mis_hip.gem5 --options '/gem5/tests/gem5/gpu/../gpu-pannotia-resources/pannotia-datasets/1k_128k.gr 0'
> 
> Num SQC =  1 Num scalar caches =  1 Num CU =  4
> Global frequency set at 1000000000000 ticks per second
> ```

@Harshil2107 : apologies for my very delayed response (and the other 12 things waiting on me ...) ... I am slammed this week.  I also think this is a new error for MIS relative to what you said was happening before?

Regardless I checked the MIS code in gem5-resources and I do not see the version in gem5-resources using mmap at all.  Neither the parsing code (https://github.com/gem5/gem5-resources/blob/stable/src/gpu/pannotia/graph_parser/parse.cpp) nor the host code (https://github.com/gem5/gem5-resources/blob/stable/src/gpu/pannotia/mis/mis.cpp).  I do see on your command line you have a pretty large memory size -- much larger than we used to use with MIS (https://github.com/gem5/gem5/blob/stable/tests/deprecated/weekly.sh#L328).  I assume you made this change intentionally to fix something?

So assuming changing the memory size back to 8 GiB does not resolve the problem, is there a way you can get a backtrace to help me understand where the mmap is coming from?

Right now I am just not able to keep pace with everything you all are doing, sorry, and if that is blocking something per @BobbyRBruce 's message than you should proceed and create a list for me of what is not working...and I'll get to it at some point when I have more time.

Harshil2107:
Hi @mattsinc,

I will re run the test with 8 GB memory. I commented a lot of tests in this PR as they are taking a very long time to run. I started the runs last Tuesday and they are still not complete on my local machine. This will more than likely cause the runners to time out so I only left 3 tests to run during the weekly tests.

mattsinc:
> Hi @mattsinc,
> 
> I will re run the test with 8 GB memory. I commented a lot of tests in this PR as they are taking a very long time to run. I started the runs last Tuesday and they are still not complete on my local machine. This will more than likely cause the runners to time out so I only left 3 tests to run during the weekly tests.

Yes, I strongly agree you should comment everything else out and just rerun this.  Thanks for looking into this, regardless.

tests/gem5/gpu/test_gpu_pannotia.py
BobbyRBruce, line 1: Can we update this file to not download any tests/datasets we aren't using?

BobbyRBruce (changes requested):

BobbyRBruce (approved):

mattsinc:
> > Hi @mattsinc,
> > I will re run the test with 8 GB memory. I commented a lot of tests in this PR as they are taking a very long time to run. I started the runs last Tuesday and they are still not complete on my local machine. This will more than likely cause the runners to time out so I only left 3 tests to run during the weekly tests.
> 
> Yes, I strongly agree you should comment everything else out and just rerun this. Thanks for looking into this, regardless.

@Harshil2107 did you find a fix for this?  I see it was merged in

BobbyRBruce:
Yeah. The fix was just selecting a sunset of tests that ran in a few hours. We dropped the rest. The weekly tests should pass this weekend (instead of timing out as before)

mattsinc:
> Yeah. The fix was just selecting a sunset of tests that ran in a few hours. We dropped the rest. The weekly tests should pass this weekend (instead of timing out as before)

I'm not sure how that fixed the bug @Harshil2107 was seeing?  It sounds like it just hides it ... maybe I'm missing something?

BobbyRBruce:
> > Yeah. The fix was just selecting a sunset of tests that ran in a few hours. We dropped the rest. The weekly tests should pass this weekend (instead of timing out as before)
> 
> I'm not sure how that fixed the bug @Harshil2107 was seeing? It sounds like it just hides it ... maybe I'm missing something?

Maybe there's some confusion here (i.e., I could be wrong), but my understanding is the error @Harshil2107 was getting was just related to the tests timing out for taking too long to run. I don't think there was any evidence of gem5 stalling or there being any bug causing this. The issue, as far as I was made aware, was that they are big tests that take a long time to complete on our runners. So our "fix" is simply to not run the biggest Pannotia tests and just run the smaller ones that can complete in a matter of hours. 


@Harshil2107 : Is what I'm saying here true?


I will also say that implementing these tests has been a bit more of a pain for us than we have anticipated, so just dropping these large tests is also an attempt just to get this work item off our plates so we can move onto other things. I don't think we gain enough from doing more work on Pannotia tests at this point, though if there is a bug herd, like you suggest, we'll look into fixing it, or at least reporting it on the Issue's page.

Harshil2107:
> > Yeah. The fix was just selecting a sunset of tests that ran in a few hours. We dropped the rest. The weekly tests should pass this weekend (instead of timing out as before)
> 
> I'm not sure how that fixed the bug @Harshil2107 was seeing? It sounds like it just hides it ... maybe I'm missing something?

I ran the hip test with 8gb memory and that passed locally. But I still left the note that it fails with 64gb memory. I can remove the note if it is not required.

The reason for commenting the other tests and only running a few for weekly's is because the tests took a long time to run locally. I ran them for a week and they were still not finished running. So to stop the runners from timing out I commented the rest of the tests.

Edit: Pinging @mattsinc, in case notification wasnt sent.

mattsinc:
> > > Yeah. The fix was just selecting a sunset of tests that ran in a few hours. We dropped the rest. The weekly tests should pass this weekend (instead of timing out as before)
> > 
> > 
> > I'm not sure how that fixed the bug @Harshil2107 was seeing? It sounds like it just hides it ... maybe I'm missing something?
> 
> I ran the hip test with 8gb memory and that passed locally. But I still left the note that it fails with 64gb memory. I can remove the note if it is not required.
> 
> The reason for commenting the other tests and only running a few for weekly's is because the tests took a long time to run locally. I ran them for a week and they were still not finished running. So to stop the runners from timing out I commented the rest of the tests.
> 
> Edit: Pinging @mattsinc, in case notification wasnt sent.

Thanks @Harshil2107 -- it sounds like we need to make sure it uses 8 GB of memory then in what is committed for weekly tests.  Maybe I'm missing it, but it seems like that is not part of this merged patch?

Harshil2107:
I updated the comment and the test memory in this PR #1698. It also moves the gpu tests to have their separate jobs in the workflow files.

PR #1633, opened 2024-10-05
dev-amdgpu: Deprecate rom and mmio trace params

abmerop:
The ROM field was originally intended as a future alternate way to load VBIOS without the ROM being on the disk image. This code path is never taken for the devices gem5 supports and there is no gem5 implementation. Deprecate the rom_binary field for this reason.

Similarly, MMIO traces were only used for Vega10. Deprecate this as Vega10 is now deprecated. The MMIO trace reader is kept as it may still be useful in the future. It is still the primary way to handle devies which have graphics capability. None of the devices supported by gem5 have graphics now that Vega10 is deprecated.

### DIFF
--- a/src/dev/amdgpu/AMDGPU.py
+++ b/src/dev/amdgpu/AMDGPU.py
@@ -81,8 +81,6 @@ class AMDGPUDevice(PciDevice):
     InterruptPin = 2
     ExpansionROM = 0
 
-    rom_binary = Param.String("ROM binary dumped from hardware")
-    trace_file = Param.String("MMIO trace collected on hardware")
     checkpoint_before_mmios = Param.Bool(
         False, "Take a checkpoint before the device begins sending MMIOs"
     )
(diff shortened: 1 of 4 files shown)


src/dev/amdgpu/AMDGPU.py
```diff
@@ -81,8 +81,8 @@ class AMDGPUDevice(PciDevice):
     InterruptPin = 2
     ExpansionROM = 0
 
-    rom_binary = Param.String("ROM binary dumped from hardware")
-    trace_file = Param.String("MMIO trace collected on hardware")
+    rom_binary = Param.String("", "Deprecated - This is ignored")
```
mattsinc: Should we just remove these lines completely instead of removing the contents?

abmerop: Yea, Let me change it to do that.

mattsinc (approved):

PR #1636, opened 2024-10-06
stdlib: Add viper board, viper cache, and gpu components

abmerop:
Adds GPU_VIPER protocol related caches to stdlib components: CorePair cache, TCP, SQC, TCC, Directory, and DMA controllers. Adds GPU related components in a new components/devices/gpus/ directory. Adds prebuilt GPU and CPU cache hierarchies, GPU and CPU network classes, and a board overriding the X86Board to provide helper methods for disk image root, the complex kernel parameter list, and method to provide functionality to the current GPUFS scripts to load in applications and handle loading the GPU driver.

The new GPU components can be used as follows:
 - Create a GPU device *before* the CPU cache hierarchy is created.
 - Add the GPU's CPU-side DMA controllers to the list of CPU cache controllers.
 - Use GPU device method to connect to an AbstractBoard.

Each GPU components has it's own RubySystem, PCI device ID, and address ranges for VBIOS and legacy PCI BARs. Therefore, in theory, multiple GPUs can be created. This requires PR #1453 .

An example of using this board is added to configs/example/gem5_library under x86-mi300x-gpu.py. It is designed to work with the disk image, kernel, and applications provided in the gem5-resources repository.

Change-Id: Ie65ffcfee5e311d9492de935d6d0631260645cd3

### DIFF
--- a/configs/example/gem5_library/x86-mi300x-gpu.py
+++ b/configs/example/gem5_library/x86-mi300x-gpu.py
@@ -1,0 +1,155 @@
+# Copyright (c) 2024 Advanced Micro Devices, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its
+# contributors may be used to endorse or promote products derived from this
+# software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+"""
+Script to run a full system GPU simulation.
+
+Usage:
+------
+```
+scons build/VEGA_X86/gem5.opt
+./build/VEGA_X86/gem5.opt
+    configs/example/gem5_library/x86-viper-gpu.py
+    --image <disk image>
+    --kernel <kernel>
+    --app <gpu application>
+```
+
+Example:
+--------
+```
+./build/VEGA_X86/gem5.opt
+    configs/example/gem5_library/x86-viper-gpu.py
+    --image ./gem5-resources/src/x86-ubuntu-gpu-ml/disk-image/x86-ubuntu-gpu-ml
+    --kernel ./gem5-resources/src/x86-ubuntu-gpu-ml/vmlinux-gpu-ml
+    --app ./gem5-resources/src/gpu/square/bin.default/square.default
+```
+"""
+
+import argparse
+
+from gem5.coherence_protocol import CoherenceProtocol
+from gem5.components.devices.gpus.amdgpu import MI300X
+from gem5.components.memory.single_channel import SingleChannelDDR4_2400
+from gem5.components.processors.cpu_types import CPUTypes
+from gem5.components.processors.simple_processor import SimpleProcessor
+from gem5.isas import ISA
+from gem5.prebuilt.viper.board import ViperBoard
+from gem5.prebuilt.viper.cpu_cache_hierarchy import ViperCPUCacheHierarchy
+from gem5.resources.resource import (
+    DiskImageResource,
+    FileResource,
+)
+from gem5.simulate.simulator import Simulator
+from gem5.utils.requires import requires
+
+requires(
+    isa_required=ISA.X86,
+    coherence_protocol_required=CoherenceProtocol.GPU_VIPER,
+)
+
+# Kernel, disk, and applications are obtained locally.
+parser = argparse.ArgumentParser()
+
+parser.add_argument(
+    "--image",
+    type=str,
+    required=True,
+    help="Full path to the gem5-resources x86-ubuntu-gpu-ml disk-image.",
+)
+
+parser.add_argument(
+    "--kernel",
+    type=str,
+    required=True,
+    help="Full path to the gem5-resources vmlinux-gpu-ml kernel.",
+)
+
+parser.add_argument(
+    "--app",
+    type=str,
+    required=True,
+    help="Path to GPU application, python script, or bash script to run",
+)
+
+parser.add_argument(
+    "--kvm-perf",
+    default=False,
+    action="store_true",
+    help="Use KVM perf counters to give accurate GPU insts/cycles with KVM",
+)
+
+args = parser.parse_args()
+
+# stdlib only supports up to 3GiB currently. This will need to be expanded in
+# the future.
+memory = SingleChannelDDR4_2400(size="3GiB")
+
+# Note: Only KVM and ATOMIC work due to buggy MOESI_AMD_Base protocol.
+processor = SimpleProcessor(cpu_type=CPUTypes.KVM, isa=ISA.X86, num_cores=2)
+
+for core in processor.cores:
+    if core.is_kvm_core():
+        core.get_simobject().usePerf = args.kvm_perf
+
+# The GPU must be created first so we can assign CPU-side DMA ports to the
+# CPU cache hierarchy.
+gpu0 = MI300X()
+
+cache_hierarchy = ViperCPUCacheHierarchy(
+    l1d_size="32KiB",
+    l1d_assoc=8,
+    l1i_size="32KiB",
+    l1i_assoc=8,
+    l2_size="1MiB",
+    l2_assoc=16,
+    l3_size="16MiB",
+    l3_assoc=16,
+)
+
+board = ViperBoard(
+    clk_freq="3GHz",
+    processor=processor,
+    memory=memory,
+    cache_hierarchy=cache_hierarchy,
+    gpus=[gpu0],
+)
+
+# Example of using a local disk image resource
+disk = DiskImageResource(local_path=args.image, root_partition="1")
+kernel = FileResource(local_path=args.kernel)
+
+board.set_kernel_disk_workload(
+    kernel=kernel,
+    disk_image=disk,
+    readfile_contents=board.make_gpu_app(gpu0, args.app),
+)
+
+simulator = Simulator(board=board)
+simulator.run()
--- a/configs/example/gem5_library/x86-mi300x-gpu.py
+++ b/configs/example/gem5_library/x86-mi300x-gpu.py
@@ -56,6 +56,7 @@ import argparse
 
 from gem5.coherence_protocol import CoherenceProtocol
 from gem5.components.devices.gpus.amdgpu import MI300X
+from gem5.components.memory import HBM2Stack
 from gem5.components.memory.single_channel import SingleChannelDDR4_2400
 from gem5.components.processors.cpu_types import CPUTypes
 from gem5.components.processors.simple_processor import SimpleProcessor
@@ -120,7 +121,7 @@ for core in processor.cores:
 
 # The GPU must be created first so we can assign CPU-side DMA ports to the
 # CPU cache hierarchy.
-gpu0 = MI300X()
+gpu0 = MI300X(gpu_memory=HBM2Stack(size="16GiB"))
 
 cache_hierarchy = ViperCPUCacheHierarchy(
     l1d_size="32KiB",
--- a/src/python/SConscript
+++ b/src/python/SConscript
@@ -175,10 +175,34 @@ PySource('gem5.components.cachehierarchies.ruby.caches.mi_example',
     'dma_controller.py')
 PySource('gem5.components.cachehierarchies.ruby.caches.mi_example',
     'gem5/components/cachehierarchies/ruby/caches/mi_example/l1_cache.py')
+PySource('gem5.components.cachehierarchies.ruby.caches.viper',
+    'gem5/components/cachehierarchies/ruby/caches/viper/__init__.py')
+PySource('gem5.components.cachehierarchies.ruby.caches.viper',
+    'gem5/components/cachehierarchies/ruby/caches/viper/corepair_cache.py')
+PySource('gem5.components.cachehierarchies.ruby.caches.viper',
+    'gem5/components/cachehierarchies/ruby/caches/viper/directory.py')
+PySource('gem5.components.cachehierarchies.ruby.caches.viper',
+    'gem5/components/cachehierarchies/ruby/caches/viper/dma_controller.py')
+PySource('gem5.components.cachehierarchies.ruby.caches.viper',
+    'gem5/components/cachehierarchies/ruby/caches/viper/tcp.py')
+PySource('gem5.components.cachehierarchies.ruby.caches.viper',
+    'gem5/components/cachehierarchies/ruby/caches/viper/sqc.py')
+PySource('gem5.components.cachehierarchies.ruby.caches.viper',
+    'gem5/components/cachehierarchies/ruby/caches/viper/tcc.py')
 PySource('gem5.components.cachehierarchies.ruby.topologies',
     'gem5/components/cachehierarchies/ruby/topologies/__init__.py')
 PySource('gem5.components.cachehierarchies.ruby.topologies',
     'gem5/components/cachehierarchies/ruby/topologies/simple_pt2pt.py')
+
+PySource('gem5.components.devices',
+    'gem5/components/devices/__init__.py')
+PySource('gem5.components.devices.gpus',
+    'gem5/components/devices/gpus/__init__.py')
+PySource('gem5.components.devices.gpus',
+    'gem5/components/devices/gpus/amdgpu.py')
+PySource('gem5.components.devices.gpus',
+    'gem5/components/devices/gpus/viper_shader.py')
+
 PySource('gem5.components.memory', 'gem5/components/memory/__init__.py')
 PySource('gem5.components.memory', 'gem5/components/memory/abstract_memory_system.py')
 PySource('gem5.components.memory', 'gem5/components/memory/dramsim_3.py')
@@ -289,6 +313,14 @@ PySource('gem5.prebuilt.riscvmatched',
     'gem5/prebuilt/riscvmatched/riscvmatched_processor.py')
 PySource('gem5.prebuilt.riscvmatched',
     'gem5/prebuilt/riscvmatched/riscvmatched_core.py')
+PySource('gem5.prebuilt.viper', 'gem5/prebuilt/viper/__init__.py')
+PySource('gem5.prebuilt.viper', 'gem5/prebuilt/viper/board.py')
+PySource('gem5.prebuilt.viper',
+    'gem5/prebuilt/viper/cpu_cache_hierarchy.py')
+PySource('gem5.prebuilt.viper',
+    'gem5/prebuilt/viper/gpu_cache_hierarchy.py')
+PySource('gem5.prebuilt.viper',
+    'gem5/prebuilt/viper/viper_network.py')
 PySource('gem5.resources', 'gem5/resources/__init__.py')
 PySource('gem5.resources', 'gem5/resources/client.py')
 PySource('gem5.resources', 'gem5/resources/downloader.py')
--- a/src/python/gem5/components/cachehierarchies/ruby/caches/viper/corepair_cache.py
+++ b/src/python/gem5/components/cachehierarchies/ruby/caches/viper/corepair_cache.py
@@ -1,0 +1,124 @@
+# Copyright (c) 2024 Advanced Micro Devices, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its
+# contributors may be used to endorse or promote products derived from this
+# software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import math
+
+from m5.objects import (
+    CorePair_Controller,
+    MessageBuffer,
+    RubyCache,
+    TreePLRURP,
+)
+
+from gem5.components.processors.abstract_core import AbstractCore
+
+
+class CorePairCache(CorePair_Controller):
+    def __init__(
+        self,
+        l1i_size: str,
+        l1i_assoc: int,
+        l1d_size: str,
+        l1d_assoc: int,
+        l2_size: str,
+        l2_assoc: int,
+        network,
+        cache_line_size,
+        core: AbstractCore,
+    ):
+        """Creating CorePair cache controller. Consist of both instruction
+        and data cache for a pair of L1s and a single L2 cache shared between
+        them.
+        """
+        super().__init__()
+
+        self.send_evictions = core.requires_send_evicts()
+
+        self.L1Icache = RubyCache(
+            size=l1i_size,
+            assoc=l1i_assoc,
+            replacement_policy=TreePLRURP(),
+            resourceStalls=False,
+            dataArrayBanks=2,
+            tagArrayBanks=2,
+            dataAccessLatency=1,
+            tagAccessLatency=1,
+        )
+
+        self.L1D0cache = RubyCache(
+            size=l1d_size,
+            assoc=l1d_assoc,
+            replacement_policy=TreePLRURP(),
+            resourceStalls=False,
+            dataArrayBanks=2,
+            tagArrayBanks=2,
+            dataAccessLatency=1,
+            tagAccessLatency=1,
+        )
+
+        self.L1D1cache = RubyCache(
+            size=l1d_size,
+            assoc=l1d_assoc,
+            replacement_policy=TreePLRURP(),
+            resourceStalls=False,
+            dataArrayBanks=2,
+            tagArrayBanks=2,
+            dataAccessLatency=1,
+            tagAccessLatency=1,
+        )
+
+        self.L2cache = RubyCache(
+            size=l2_size,
+            assoc=l2_assoc,
+            replacement_policy=TreePLRURP(),
+            resourceStalls=False,
+            dataArrayBanks=16,
+            tagArrayBanks=16,
+        )
+
+        self.connectQueues(network)
+
+    def connectQueues(self, network):
+        self.requestFromCore = MessageBuffer()
+        self.requestFromCore.out_port = network.in_port
+
+        self.responseFromCore = MessageBuffer()
+        self.responseFromCore.out_port = network.in_port
+
+        self.unblockFromCore = MessageBuffer()
+        self.unblockFromCore.out_port = network.in_port
+
+        self.probeToCore = MessageBuffer()
+        self.probeToCore.in_port = network.out_port
+
+        self.responseToCore = MessageBuffer()
+        self.responseToCore.in_port = network.out_port
+
+        self.mandatoryQueue = MessageBuffer()
+        self.triggerQueue = MessageBuffer(ordered=True)
--- a/src/python/gem5/components/cachehierarchies/ruby/caches/viper/corepair_cache.py
+++ b/src/python/gem5/components/cachehierarchies/ruby/caches/viper/corepair_cache.py
@@ -30,7 +30,7 @@
 import math
 
 from m5.objects import (
-    CorePair_Controller,
+    GPU_VIPER_CorePair_Controller,
     MessageBuffer,
     RubyCache,
     TreePLRURP,
@@ -39,7 +39,7 @@ from m5.objects import (
 from gem5.components.processors.abstract_core import AbstractCore
 
 
-class CorePairCache(CorePair_Controller):
+class CorePairCache(GPU_VIPER_CorePair_Controller):
     def __init__(
         self,
         l1i_size: str,
--- a/src/python/gem5/components/cachehierarchies/ruby/caches/viper/sqc.py
+++ b/src/python/gem5/components/cachehierarchies/ruby/caches/viper/sqc.py
@@ -1,0 +1,73 @@
+# Copyright (c) 2024 Advanced Micro Devices, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its
+# contributors may be used to endorse or promote products derived from this
+# software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+from m5.objects import (
+    MessageBuffer,
+    RubyCache,
+    SQC_Controller,
+    TreePLRURP,
+)
+
+
+class SQCCache(SQC_Controller):
+    def __init__(
+        self,
+        sqc_size: str,
+        sqc_assoc: int,
+        network,
+        cache_line_size,
+    ):
+        """Creating SQC cache controller. This is the Icache for GPU devices."""
+
+        super().__init__()
+
+        self.L1cache = RubyCache(
+            size=sqc_size,
+            assoc=sqc_assoc,
+            dataArrayBanks=8,
+            tagArrayBanks=8,
+            dataAccessLatency=1,
+            tagAccessLatency=1,
+            resourceStalls=True,
+            replacement_policy=TreePLRURP(),
+        )
+
+        self.connectQueues(network)
+
+    def connectQueues(self, network):
+        self.requestFromSQC = MessageBuffer(ordered=True)
+        self.requestFromSQC.out_port = network.in_port
+
+        self.probeToSQC = MessageBuffer(ordered=True)
+        self.probeToSQC.in_port = network.out_port
+
+        self.responseToSQC = MessageBuffer(ordered=True)
+        self.responseToSQC.in_port = network.out_port
+
+        self.mandatoryQueue = MessageBuffer()
--- a/src/python/gem5/components/cachehierarchies/ruby/caches/viper/sqc.py
+++ b/src/python/gem5/components/cachehierarchies/ruby/caches/viper/sqc.py
@@ -28,14 +28,14 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 from m5.objects import (
+    GPU_VIPER_SQC_Controller,
     MessageBuffer,
     RubyCache,
-    SQC_Controller,
     TreePLRURP,
 )
 
 
-class SQCCache(SQC_Controller):
+class SQCCache(GPU_VIPER_SQC_Controller):
     def __init__(
         self,
         sqc_size: str,
--- a/src/python/gem5/components/cachehierarchies/ruby/caches/viper/tcc.py
+++ b/src/python/gem5/components/cachehierarchies/ruby/caches/viper/tcc.py
@@ -1,0 +1,87 @@
+# Copyright (c) 2024 Advanced Micro Devices, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its
+# contributors may be used to endorse or promote products derived from this
+# software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+from m5.objects import (
+    MessageBuffer,
+    RubyCache,
+    TCC_Controller,
+    TreePLRURP,
+)
+
+
+class TCCCache(TCC_Controller):
+    def __init__(
+        self,
+        tcc_size: str,
+        tcc_assoc: int,
+        network,
+        cache_line_size,
+    ):
+        """Creating TCC cache controller. This is the L2 cache for GPU devices."""
+
+        super().__init__()
+
+        self.L2cache = RubyCache(
+            size=tcc_size,
+            assoc=tcc_assoc,
+            dataArrayBanks=256,
+            tagArrayBanks=256,
+            dataAccessLatency=8,
+            tagAccessLatency=2,
+            resourceStalls=True,
+            replacement_policy=TreePLRURP(),
+            atomicLatency=0,
+            atomicALUs=64,
+        )
+
+        self.connectQueues(network)
+
+    def connectQueues(self, network):
+        self.requestFromTCP = MessageBuffer(ordered=True)
+        self.requestFromTCP.in_port = network.out_port
+
+        self.responseToCore = MessageBuffer(ordered=True)
+        self.responseToCore.out_port = network.in_port
+
+        self.probeFromNB = MessageBuffer()
+        self.probeFromNB.in_port = network.out_port
+
+        self.responseFromNB = MessageBuffer()
+        self.responseFromNB.in_port = network.out_port
+
+        self.requestToNB = MessageBuffer(ordered=True)
+        self.requestToNB.out_port = network.in_port
+
+        self.responseToNB = MessageBuffer()
+        self.responseToNB.out_port = network.in_port
+
+        self.unblockToNB = MessageBuffer()
+        self.unblockToNB.out_port = network.in_port
+
+        self.triggerQueue = MessageBuffer(ordered=True)
--- a/src/python/gem5/components/cachehierarchies/ruby/caches/viper/tcc.py
+++ b/src/python/gem5/components/cachehierarchies/ruby/caches/viper/tcc.py
@@ -28,14 +28,14 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 from m5.objects import (
+    GPU_VIPER_TCC_Controller,
     MessageBuffer,
     RubyCache,
-    TCC_Controller,
     TreePLRURP,
 )
 
 
-class TCCCache(TCC_Controller):
+class TCCCache(GPU_VIPER_TCC_Controller):
     def __init__(
         self,
         tcc_size: str,
--- a/src/python/gem5/components/cachehierarchies/ruby/caches/viper/tcp.py
+++ b/src/python/gem5/components/cachehierarchies/ruby/caches/viper/tcp.py
@@ -1,0 +1,79 @@
+# Copyright (c) 2024 Advanced Micro Devices, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its
+# contributors may be used to endorse or promote products derived from this
+# software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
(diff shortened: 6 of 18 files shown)


abmerop:
@powerjg @BobbyRBruce  Is there an example of an x86 board using >3GB?  This will be a problem because the GPU driver reserves ~2GB of memory, which doesn't leave much for the application.  I believe that is the only TODO I have for this PR.

abmerop:
There are several problems with the CPU protocol (MOESI_AMD_Base).  It requires Ruby backing store, which it seems no other protocol does anymore.  It also have several invalid transitions when trying to boot Linux with TimingSimpleCPU. 

The protocol can be fixed, but with multiple RubySystems it wouldn't be too difficult to make a .slicc file with MESI_Two_Level (for example) on the CPU side and VIPER on the GPU side. Mostly a single DMA needs to be used. @mattsinc maybe someone will take this up in your class :)

BobbyRBruce:
> @powerjg @BobbyRBruce Is there an example of an x86 board using >3GB? This will be a problem because the GPU driver reserves ~2GB of memory, which doesn't leave much for the application. I believe that is the only TODO I have for this PR.

As far as I know we don't have an example of a >3GB system. I've always considered this a restriction in all our X86 setups in gem5. I'm not sure how hard/easy it'd be to remove this restriction but If you could it'd be a welcome contribution.

powerjg:
> Is there an example of an x86 board using >3GB? This will be a problem because the GPU driver reserves ~2GB of memory, which doesn't leave much for the application. I believe that is the only TODO I have for this PR.

I have an example somewhere. I'll ask @mbabaie to put a link here tomorrow morning.

The main idea is to have a small memory range ~100MiB starting at 0 and having a memory controller dedicated to that memory range. Then, have your "real" memory range starting at 4GiB and being as large as you want (and using whatever interleaving you want). However, this is a hack. You'll have strange performance if your application happens to be assigned to physical memory in the bottom ~100MiB range. In practice, for reasonably-sized applications, this doesn't happen.

What we *need* is a way to have a hole in an interleaved address range and to backfill that hole. From what I can tell from looking at the E820 entries on a real system, this is what it does.

powerjg:
> but with multiple RubySystems it wouldn't be too difficult to make a .slicc file with MESI_Two_Level (for example) on the CPU side and VIPER on the GPU side. 

Would this really be easy?!?! If so, amazing! However, I think that it's going to be difficult since the machines, messages, etc. will have conflicting names.

abmerop:
I actually know how to set the e820 table correctly.  The current GPUFS scripts already do this. 

The issue I run into is that both SingleChannelDDR4_2400 and DualChannelDDR4_2400 in the stdlib error out saying they only accept one memory range.  Is there no stdlib memory component taking multiple memory ranges?

powerjg:
> Is there no stdlib memory component taking multiple memory ranges?

This isn't possible in gem5 in general. You can't interleave with multiple memory ranges. The only way I know to hack around this is to create a special memory controller which maps the other range. Otherwise, you would end up with <ranges> * <interleaving> controllers, which is (also) unrealistic.

mbabaie:
Hey @abmerop , here's the link to the example Jason is referring to:
  
https://github.com/darchr/dramCacheController/blob/isca-submission/configs-npb-gapbs-chkpt-restore/system/ruby_system_8channel.py

We also have it working for single channel in the same directory. Please feel free to brows the directory as needed.

Pay attention to the lines 117-123 where the address space is divided between kernel, IO, and data for workloads.
At line 249, function `_createMemoryControllers` creates the memory controllers for these ranges. `PolicyManager` is basically a DRAM cache controller acting as a drop-in replacement for memory controllers we generally have. 

Please let me know if you have any questions.

configs/example/gem5_library/x86-mi300x-gpu.py
```diff
@@ -0,0 +1,150 @@
...
+
+Usage:
+------
+```
+scons build/VEGA_X86/gem5.opt
+./build/VEGA_X86/gem5.opt
+    configs/example/gem5_library/x86-viper-gpu.py
+    --image <disk image>
+    --kernel <kernel>
+    --app <gpu application>
```
powerjg, line 41: I would prefer to hard-code these and not have command line parameters. We should use things available from gem5-resource (i.e., with `obtain_resource`). I like to think of these files as **examples** for people to model their own scripts off of. I don't think people should use these files directly.

configs/example/gem5_library/x86-mi300x-gpu.py
```diff
@@ -0,0 +1,150 @@
...
+
+# TODO: This memory size is going to be too small to do anything useful
+memory = SingleChannelDDR4_2400(size="3GiB")
+
+# Note: Only KVM and ATOMIC work due to buggy MOESI_AMD_Base protocol.
+processor = SimpleProcessor(cpu_type=CPUTypes.KVM, isa=ISA.X86, num_cores=4)
+
+for core in processor.cores:
+    if core.is_kvm_core():
+        core.get_simobject().usePerf = False
```
powerjg: Actually *this* could be a command line parameter 😆. Most systems support perf, so we could make it optional to disable it.

configs/example/gem5_library/x86-mi300x-gpu.py
```diff
@@ -0,0 +1,150 @@
...
+    l3_assoc=16,
+    # Add our GPU device's CPU-side DMA ports to this hierarchy
+    device_dmas=gpu0.get_cpu_dma_ports(),
+)
+
+board = ViperBoard(
+    clk_freq="3GHz",
+    processor=processor,
+    memory=memory,
+    cache_hierarchy=cache_hierarchy,
```
powerjg, line 141: I suggest adding a (optional) `gpu=` parameter here and connecting it in `connect_things` like everything else.

configs/example/gem5_library/x86-mi300x-gpu.py
```diff
@@ -0,0 +1,150 @@
...
+)
+
+# Connect GPU to board
+gpu0.connectGPU(board)
```
powerjg: See above. I don't like having an extra (special) call.

src/python/gem5/components/devices/__init__.py
powerjg, line 1: I like this new directory 😄

src/python/gem5/components/devices/gpus/amdgpu.py
```diff
@@ -0,0 +1,276 @@
...
+# POSSIBILITY OF SUCH DAMAGE.
+
+from m5.objects import AMDGPUDevice
+
+from ....components.boards.abstract_board import AbstractBoard
+from ....prebuilt.viper.gpu_cache_hierarchy import ViperGPUCacheHierarchy
+from .viper_shader import ViperShader
+
+
+class BaseViperGPU:
```
powerjg: Why doesn't this need to be a SimObject? E.g., I would expect this to inherit from some SimObject (SubSystem if nothing else).

src/python/gem5/prebuilt/viper/board.py
```diff
@@ -0,0 +1,99 @@
...
+    GPU device. Provides all the functionality of the X86Board with helper
+    methods specific to booting a disk with GPU libraries installed.
+    """
+
+    def __init__(
+        self,
+        clk_freq: str,
+        processor: AbstractProcessor,
+        memory: AbstractMemorySystem,
+        cache_hierarchy: AbstractCacheHierarchy,
```
powerjg, line 63: I suggest adding a `gpu` parameter here. I doubt the board makes sense without a GPU, but if it does, you can make the GPU optional (default to `None`)

src/python/gem5/prebuilt/viper/board.py
```diff
@@ -0,0 +1,99 @@
...
+        return "/dev/sda"
+
+    @overrides(KernelDiskWorkload)
+    def get_default_kernel_args(self) -> List[str]:
+        return [
+            "earlyprintk=ttyS0",
+            "console=ttyS0",
+            "lpj=7999923",
+            "root={root_value}",
+            "disk_device={disk_device}",
```
powerjg: `disk_device` isn't a valid kernel parameter. It's somehow sneaked in to some of our code, but it should be removed.

src/python/gem5/prebuilt/viper/board.py
```diff
@@ -0,0 +1,99 @@
...
+            encodedBin = base64.b64encode(binfile.read()).decode()
+
+        application_command = (
+            f'echo "{encodedBin}" | base64 -d > myapp\n'
+            "chmod +x myapp\n"
+            "./myapp {}\n"
+            "/sbin/m5 exit\n"
+        )
+
+        return driver_load_command + application_command
```
powerjg, line 191: I'm not sure about line 87, but the rest of this should be instead in a `workload` object. We want to keep "workloads" (i.e., the thing you run on the board) separate from the "hardware" of the board.

Alternatively, you could add a *mixin* like `set_kernel_disk_workload` for the GPU applications. However, it is important that we keep it compatible with the `Workload` objects. Happy to provide more detail if needed.

src/python/gem5/prebuilt/viper/cpu_cache_hierarchy.py
powerjg, line 1: I'm mixed on the way we've separated out "components" from "prebuilt".

I feel like if this file is in `prebuilt/` then that means the GPUs defined in `components/devices/gpus/amdgpu.py` should be in `prebuilt/`. However, I also recognize that we have at least one "prebuilt" cache hierarchy in the `prebuilt` directory... So, I don't know what the best solution is. I suppose there's no needed changes, but it may be worth discussing more.

src/python/gem5/prebuilt/viper/gpu_cache_hierarchy.py
```diff
@@ -0,0 +1,351 @@
...
+        self._tcc_size = tcc_size
+        self._tcc_assoc = tcc_assoc
+        self._cache_line_size = cache_line_size
+
+        self.ruby_gpu = RubySystem()
+        self.ruby_gpu.block_size_bytes = cache_line_size
+
+        # We don't want incorporate_cache which assigns to board, so the rest
+        # of this constructor implements the functionality of incorporate_cache
+        # seen in other stdlib cache hierarchies.
```
powerjg: Can you explain this comment more? I would think we *do* want the board to call `incorporate_cache` with the GPU.

I.e., the `ViperBoard` should override `connect_things()` in that class and call `gpu.incorporate_caches(self)`

powerjg (commented):
I can't be more excited about a gem5 PR! Thank you so much for all of the effort you have put into this!!

My comments below are all pretty small. There's a couple of points that I worry about integration into the stdlib that may not have been (read: definitely were not) documented well. However, if you want to push back, I'm happy to have a conversation.

We'll be testing this locally as well, and we'll report back if there are any issues. 

Again, I want to express a huge amount of appreciation for doing this work. On a personal note, this change is going to accelerate the research of my group, so I can't thank you enough!

abmerop:
> Hey @abmerop , here's the link to the example Jason is referring to:
> 
> https://github.com/darchr/dramCacheController/blob/isca-submission/configs-npb-gapbs-chkpt-restore/system/ruby_system_8channel.py
> 
> We also have it working for single channel in the same directory. Please feel free to brows the directory as needed.
> 
> Pay attention to the lines 117-123 where the address space is divided between kernel, IO, and data for workloads. At line 249, function `_createMemoryControllers` creates the memory controllers for these ranges. `PolicyManager` is basically a DRAM cache controller acting as a drop-in replacement for memory controllers we generally have.
> 
> Please let me know if you have any questions.

Thank you, this is super helpful!  I was able to get it working.  It's a bit messy and there is a draft PR which might enable holes in an AddrRange so I will wait for that before implementing support for >3GiB of memory.

configs/example/gem5_library/x86-mi300x-gpu.py
abmerop: What is the alternative since no (non-ancient) GPUFS resources are available online?  I needed a way to be able to test this and gave up trying to make a local json config (got a bunch of unhelpful python errors reading the json). Are the users expected to have to create a gem5-config.json for their local disk images to use obtain_resource?

configs/example/gem5_library/x86-mi300x-gpu.py
abmerop: I prefer to have it off by default. As far as I know this only impacts one (two?) stats which are the instruction count and maybe cpu cycle count. For GPU simulation we typically ignore CPU stats. I've also observed the opposite where most systems do _not_ support perf. From what I remember the Ubuntu default is a very high paranoid level that needs to be manually changed. I know our internal batch servers do not allow perf. We also turned it off by default in the current GPUFS configs because quite a few users on the mailing list, ORNL, and others ended up needing to add --no-kvm-perf to the command line (see #1391 ).  I'd much rather have one less command line option in the common case and have the user add a --kvm-perf if they really want it turned on

configs/example/gem5_library/x86-mi300x-gpu.py
abmerop: Good suggestion, it's even more clean now.

src/python/gem5/prebuilt/viper/board.py
abmerop: Should I then create something like GpuWorkload (in src/python/gem5/resources)?  There needs to be different driver load commands for different devices so that would be a base class for something like an MI300XWorkload, MI200Workload, etc.

src/python/gem5/prebuilt/viper/board.py
abmerop: Without a GPU it would ostensibly be a MOESI_AMD_Base board, so i've made it Optional.

src/python/gem5/prebuilt/viper/board.py
abmerop: Removed this line

src/python/gem5/prebuilt/viper/gpu_cache_hierarchy.py
abmerop: I changed the comment, but basically there is no reason to delay setting up the device. I haven't looked into the purpose of incorporate_cache but I think it need to wait until the cpu and memory are created first.  This is not the case for the GPU device

src/python/gem5/components/devices/gpus/amdgpu.py
abmerop: Made this SubSystem. It does not have to be a SimObject because it's not something that is connected to the board but rather a helper class that sets up the device, shader, and cache hierarchy and then connects those individual pieces to a board

configs/example/gem5_library/x86-mi300x-gpu.py
abmerop: Done.

src/python/gem5/prebuilt/viper/cpu_cache_hierarchy.py
```diff
@@ -0,0 +1,273 @@
...
+                ctrl.dma_sequencer = DMASequencer(version=i, in_ports=port)
+
+                ctrl.ruby_system = self.ruby_system
+                ctrl.dma_sequencer.ruby_system = self.ruby_system
+
+                self._dma_controllers.append(ctrl)
+
+        # Create DMA Controllers requires for any devices in the system.
+        device_dmas = []
+        if board.get_devices() is not None:
```
abmerop, line 230: I've changed the type passed to incorporate_cache in this file so that this function is available (not sure if you have to do this python).  In the future it would be nice if something like get_devices or get_pci_devices was part of an abstract board.

configs/example/gem5_library/x86-mi300x-gpu.py
powerjg: `--kvm-perf` (with default disabled) is a good compromise :)

configs/example/gem5_library/x86-mi300x-gpu.py
powerjg: I recognize this problem. I am pushing for us to make it easier and have better documentation. Would you be OK merging what you have and letting us update it to use resources?

src/python/gem5/prebuilt/viper/board.py
powerjg: Yeah, I think that makes sense. It could also make sense to have a new mixin class. So, in addition to `set_kernel_disk_workload` we could also have a `set_gpufs_workload` if you need more parameters than what's available in `set_kernel_disk_workload`. It may be more efficient to discuss this at the dev meeting on Thursday.

I'm OK with merging this change without full support for using resources and workloads, and asking my team to implement that, but we need to make sure that we don't have to do a major refactoring.

src/python/gem5/prebuilt/viper/cpu_cache_hierarchy.py
powerjg: Yes, we should probably have that interface. The problem is that devices behave very differently (at least, that's been my experience... maybe things are better now?). 

That said, I think that making a prebuilt cache hierarchy only compatible with a specific type of board is OK (and absolutely required with how complex gem5 is). In fact, I really like that this is explicitly only compatible with certain types of boards. This way we aren't advertising that any board can be used with any hierarchy.

BobbyRBruce:
@abmerop : Would you like to present/demo/get feedback on this at the gem5 Developer meeting tomorrow? I'd be more than happy to allocate you the time.

configs/example/gem5_library/x86-mi300x-gpu.py
abmerop: That sounds good. I made the KVM perf change below so if this looks good to you feel free to merge.

configs/example/gem5_library/x86-mi300x-gpu.py
abmerop: Done

configs/example/gem5_library/x86-mi300x-gpu.py
```diff
@@ -0,0 +1,155 @@
...
+    default=False,
+    action="store_true",
+    help="Use KVM perf counters to give accurate GPU insts/cycles with KVM",
+)
+
+args = parser.parse_args()
+
+# stdlib only supports up to 3GiB currently. This will need to be expanded in
+# the future.
+memory = SingleChannelDDR4_2400(size="3GiB")
```
mattsinc, line 113: I recognize the prior comments on this chain about the memory size, but workloads we support like Pannotia require 8 GB+ -- how have they been running as is?  Or is the 3 GiB constraint "only" for stdlib (it didn't sound like it, but maybe I don't understand ...)?

src/python/gem5/components/cachehierarchies/ruby/caches/viper/corepair_cache.py
```diff
@@ -0,0 +1,124 @@
...
+        them.
+        """
+        super().__init__()
+
+        self.send_evictions = core.requires_send_evicts()
+
+        self.L1Icache = RubyCache(
+            size=l1i_size,
+            assoc=l1i_assoc,
+            replacement_policy=TreePLRURP(),
```
mattsinc, line 66: Ideally this would integrate with the support we added recently for different replacement policies: https://github.com/gem5/gem5/pull/1530 (and what @abmerop did in https://github.com/gem5/gem5/pull/1230), but if it's going to be a pain in the butt to integrate in this patch, fine to let it be for now.

src/python/gem5/components/cachehierarchies/ruby/caches/viper/sqc.py
```diff
@@ -0,0 +1,73 @@
...
+
+        self.L1cache = RubyCache(
+            size=sqc_size,
+            assoc=sqc_assoc,
+            dataArrayBanks=8,
+            tagArrayBanks=8,
+            dataAccessLatency=1,
+            tagAccessLatency=1,
+            resourceStalls=True,
+            replacement_policy=TreePLRURP(),
```
mattsinc, line 58: same thing here

src/python/gem5/components/cachehierarchies/ruby/caches/viper/tcc.py
```diff
@@ -0,0 +1,87 @@
...
+
+        self.L2cache = RubyCache(
+            size=tcc_size,
+            assoc=tcc_assoc,
+            dataArrayBanks=256,
+            tagArrayBanks=256,
+            dataAccessLatency=8,
+            tagAccessLatency=2,
+            resourceStalls=True,
+            replacement_policy=TreePLRURP(),
```
mattsinc, line 58: same thing here

src/python/gem5/components/cachehierarchies/ruby/caches/viper/tcp.py
```diff
@@ -0,0 +1,79 @@
...
+
+        self.L1cache = RubyCache(
+            size=tcp_size,
+            assoc=tcp_assoc,
+            dataArrayBanks=16,
+            tagArrayBanks=16,
+            dataAccessLatency=4,
+            tagAccessLatency=1,
+            resourceStalls=True,
+            replacement_policy=TreePLRURP(),
```
mattsinc, line 58: same here

src/python/gem5/components/devices/gpus/amdgpu.py
```diff
@@ -0,0 +1,279 @@
...
+        tcp_assoc: int = 16,
+        sqc_size: str = "32KiB",
+        sqc_assoc: int = 8,
+        scalar_size: str = "32KiB",
+        scalar_assoc: int = 8,
+        tcc_size: str = "256KiB",
+        tcc_assoc: int = 16,
+        tcc_count: int = 8,
+        num_memory_channels: int = 8,
+        cache_line_size: int = 64,
```
mattsinc, line 131: As far as I understand it, Ruby only supports 64 as an option.  I'm guessing you intend this to be for the separate AMD GPU cache line part you pushed a patch for recently?

src/python/gem5/components/devices/gpus/viper_shader.py
```diff
@@ -0,0 +1,377 @@
...
+
+        self.cu_id = cu_id
+
+        # Use in multiple places. Define variables to change once.
+        self._vreg_file_size = 2048
+        self._sreg_file_size = 2048
+
+        # Latencies, etc. use defaults in src/gpu-compute/GPU.py.
+        self.num_SIMDs = 4
+        self.n_wf = 8
```
mattsinc, line 75: how would a user configure this now?  Or is it hardcoded?

src/python/gem5/components/devices/gpus/viper_shader.py
```diff
@@ -0,0 +1,377 @@
...
+            ScalarRegisterFile(simd_id=i, num_regs=self._sreg_file_size)
+            for i in range(self.num_SIMDs)
+        ]
+
+        self.register_file_cache = [
+            RegisterFileCache(simd_id=i) for i in range(self.num_SIMDs)
+        ]
+
+        self.register_manager = RegisterManager(
+            policy="static",
```
mattsinc, line 102: I am not an expert with the stdlib, but this seems like it sets the static policy but then instantiates the other register allocator, no?

src/python/gem5/components/devices/gpus/viper_shader.py
```diff
@@ -0,0 +1,377 @@
...
+        self.ldsPort = self.ldsBus.cpu_side_port
+        self.ldsBus.mem_side_port = self.localDataStore.cuPort
+
+        self._create_tlbs()
+
+    def _create_tlbs(self):
+        # Vector memory TLB
+        self.l1_tlb = VegaGPUTLB(
+            gpu_device=self._device,
+            size=64,
```
mattsinc, line 122: same here -- previously in the GPUTLBOptions we had arguments we could use to change these sizes.  Are those completely gone and we have to manually change them all here, now?  If so, that is going to make updating these (e.g., by launching experiments in parallel with different latencies, associativities, etc.) extremely challenging...

src/python/gem5/prebuilt/viper/board.py
```diff
@@ -0,0 +1,123 @@
...
+        print("Viper board __init__ was called here now")
+
+        self._gpus = gpus
+
+    def get_devices(self):
+        return self._gpus
+
+    @overrides(AbstractBoard)
+    def _connect_things(self) -> None:
+        print("Viper board connect things was called here now")
```
mattsinc: Not sure if this was a debug print you meant to remove :)

src/python/gem5/prebuilt/viper/board.py
mattsinc: I don't fully understand the implications of this yet, but will need to.

src/python/gem5/prebuilt/viper/cpu_cache_hierarchy.py
```diff
@@ -0,0 +1,273 @@
...
+        for i in range(0, num_cores, 2):
+            cache = CorePairCache(
+                l1d_size=self._l1d_size,
+                l1d_assoc=self._l1d_assoc,
+                l1i_size=self._l1i_size,
+                l1i_assoc=self._l1i_assoc,
+                l2_size=self._l2_size,
+                l2_assoc=self._l2_assoc,
+                network=self.ruby_system.network,
+                cache_line_size=board.get_cache_line_size(),
```
mattsinc, line 127: Doesn't Ruby only support 64?  If so, not sure if we need an assert here?

src/python/gem5/prebuilt/viper/gpu_cache_hierarchy.py
```diff
@@ -0,0 +1,351 @@
...
+            )
+            ctrl.dma_sequencer = DMASequencer(version=i, in_ports=port)
+
+            ctrl.ruby_system = self.ruby_gpu
+            ctrl.dma_sequencer.ruby_system = self.ruby_gpu
+
+            self._dma_controllers.append(ctrl)
+
+        # Create GPU memories. Currently fixed to HBM2.
+        mem_type_cls = HBM_2000_4H_1x64
```
mattsinc: We are about to push the updated HBM support to public (hopefully today) ... will need to figure out how that interacts with this ...

src/python/gem5/prebuilt/viper/viper_network.py
```diff
@@ -0,0 +1,165 @@
+# Copyright (c) 2021 The Regents of the University of California.
```
mattsinc, line 1: Probably you want this to be AMD instead of Davis, @abmerop ?

mattsinc (commented):
Thanks for all your hard work on this, @abmerop .

src/python/gem5/components/devices/gpus/amdgpu.py
```diff
@@ -0,0 +1,279 @@
...
+            num_memory_channels=self._num_memory_channels,
+            cache_line_size=self._cache_line_size,
+            shader=self._shader,
+        )
+
+        # Collect GPU memory controllers created in the GPU cache hierarchy.
+        # First assign them as a child to the device so the SimObject unproxy.
+        # The device requires the memories parameter to be set as the system
+        # pointer required by the AbstractMemory class is set by AMDGPUDevice.
+        self._device.mem_ctrls = self._device.gpu_caches.get_mem_ctrls()
```
powerjg: The memory controllers should not be part of the cache. They should be a `MemorySystem` just like on the CPU side.

powerjg (commented):
I have one thing below that I don't want to forget. 

I need to give a few more parts of this a thorough review, but it's going to be at least next week before I have the time. 

The comment below can be ignored for now (unless you want to pick it up :)). I am also willing to put in the leg work on it (but not before next week!).

configs/example/gem5_library/x86-mi300x-gpu.py
abmerop: I haven't tried Pannotia.  The 3GiB constraint is specific to stdlib I think because interleaved memory is required. It has worked for about 15 years in the deprecated configs, which get around the problem but adding a non-interleaved memory range.

src/python/gem5/components/cachehierarchies/ruby/caches/viper/corepair_cache.py
abmerop: I don't know what the right balance is between configurability and needing to write your own. It seems the goal of stdlib is to get everyone to explicitly change these values by writing their own script. As far as I know you can change these values at any point before calling simulate (which calls instantiate).  This applies to your other comments below on register file type/size.

src/python/gem5/prebuilt/viper/cpu_cache_hierarchy.py
abmerop: I have successfully used this to run 64B cache lines on CPU side and 128B cache line size on GPU side. I know some folks have used up to 256B as well.

src/python/gem5/prebuilt/viper/viper_network.py
abmerop: Jason wrote the Pt2Pt class in this file so I kept the copyright

src/python/gem5/components/devices/gpus/amdgpu.py
abmerop: The memory controllers* are part of the device though. I'd have to look at what MemorySystem is and see if that causes problem with what one of the goals here is (to have the gpu be a standalone device you can create in one single line with some defaults assumed)

src/python/gem5/components/cachehierarchies/ruby/caches/viper/corepair_cache.py
mattsinc: Would like to hear from @powerjg on this then ... at minimum so I can understand as it seems like we'd want these things in the base file but perhaps not?

abmerop: I am thinking a good middle ground might be to add `kwargs` and do a check so that only specific values can be set. This would act similar to the current configs where we pass around `args` or `options` from `argparse`. Except with a "checked" kwargs dict we could: 
 - Have a default set of key/values so that:
    - It is not *required* to be passed to the stdlib class if you just want the defaults.
    - Only contains keys that are actually used which eliminates the problem with current command line arguments where you don't even know if the command line option does anything without comparing `config.ini`s.
 - Would not require having argparse at the top of every python script which clutters the system setup.

Or maybe just import some argparser from stdlib with all the accepted command line arguments and passing around `args` could accomplish the same goal with less effort.

mbabaie:
Hi @abmerop , thank you very much for sharing this work.
I have tested it with changes suggested from PR #[1453](https://github.com/gem5/gem5/pull/1453).
I created the disk image from the gem5-resources/stable, [x86-ubuntu-gpu-ml](https://github.com/gem5/gem5-resources/tree/stable/src/x86-ubuntu-gpu-ml) and followed basically what the example command in the run script suggests. It starts simulation, boots linux, and finishes the simulation. However, nothing is executed and as a result, all the stats related to gpu is 0, e.g., `ipc`'s are either 0 or NAN and `numKernelLaunched` is 0. Are you able to confirm that the actual program is executed?

abmerop:
> Hi @abmerop , thank you very much for sharing this work. I have tested it with changes suggested from PR #[1453](https://github.com/gem5/gem5/pull/1453). I created the disk image from the gem5-resources/stable, [x86-ubuntu-gpu-ml](https://github.com/gem5/gem5-resources/tree/stable/src/x86-ubuntu-gpu-ml) and followed basically what the example command in the run script suggests. It starts simulation, boots linux, and finishes the simulation. However, nothing is executed and as a result, all the stats related to gpu is 0, e.g., `ipc`'s are either 0 or NAN and `numKernelLaunched` is 0. Are you able to confirm that the actual program is executed?

My guess is that the runtime did not find a code object for the GPU that matched.  May I ask how you compiled your test application?  You can check there is an object for the GPU using `roc-obj-ls` (let me know if you need more information about that).

If the runtime does not find matching code for the GPU it will just exit without doing anything or printing a message (I don't know why the dev decided that...).

mbabaie:
> My guess is that the runtime did not find a code object for the GPU that matched. May I ask how you compiled your test application? You can check there is an object for the GPU using `roc-obj-ls` (let me know if you need more information about that).
> 
Thanks for your prompt reply.

For square I followed the [Compiling Square](https://github.com/gem5/gem5-resources/blob/stable/src/gpu/square/README.md), basically used the command written in the instructions.

I also tried the very simple pytorch program written under [The contents of pytorch_test.py](https://github.com/gem5/gem5-resources/blob/stable/src/x86-ubuntu-gpu-ml/README.md) in the disk image resources.

abmerop:
> > My guess is that the runtime did not find a code object for the GPU that matched. May I ask how you compiled your test application? You can check there is an object for the GPU using `roc-obj-ls` (let me know if you need more information about that).
> 
> Thanks for your prompt reply.
> 
> For square I followed the [Compiling Square](https://github.com/gem5/gem5-resources/blob/stable/src/gpu/square/README.md), basically used the command written in the instructions.
> 
> I also tried the very simple pytorch program written under [The contents of pytorch_test.py](https://github.com/gem5/gem5-resources/blob/stable/src/x86-ubuntu-gpu-ml/README.md) in the disk image resources.

I see, that is likely the issue because it would be a different runtime version.  The instructions at https://www.gem5.org/documentation/general_docs/gpu_models/gpufs under "Building GPU applications" are hopefully easier to follow.

How did the pytorch test fare? Since that is not being compiled it should be easier to get going.

mbabaie:
> I see, that is likely the issue because it would be a different runtime version. The instructions at https://www.gem5.org/documentation/general_docs/gpu_models/gpufs under "Building GPU applications" are hopefully easier to follow.
> 
> How did the pytorch test fare? Since that is not being compiled it should be easier to get going.

Sure. I'll try the gpufs version.

The pytorch example also did not execute anything and the stats were 0. This is the guest terminal output after booted linux (from the `board.pc.com_1.device` file):

```
Ubuntu 22.04.2 LTS gem5 ttyS0

gem5 login: root (automatic login)



The programs included with the Ubuntu system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.

Ubuntu comes with ABSOLUTELY NO WARRANTY, to the extent permitted by
applicable law.

11+0 records in
11+0 records out
11264 bytes (11 kB, 11 KiB) copied, 0.00400025 s, 2.8 MB/s
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/video/fbdev/core/sysimgblt.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/video/fbdev/core/sysfillrect.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/video/fbdev/core/syscopyarea.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/video/fbdev/core/fb_sys_fops.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/i2c/algos/i2c-algo-bit.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/media/rc/rc-core.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/media/cec/core/cec.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/gpu/drm/drm_kms_helper.ko fbdev_emulation=0 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amdkcl.ko 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amd-sched.ko 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amddrm_buddy.ko 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amdxcp.ko 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amdttm.ko 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amddrm_ttm_helper.ko 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amdgpu.ko ip_block_mask=0x6f ppfeaturemask=0 dpm=0 audio=0 ras_enable=0
./scri
```

mbabaie:
> I see, that is likely the issue because it would be a different runtime version. The instructions at https://www.gem5.org/documentation/general_docs/gpu_models/gpufs under "Building GPU applications" are hopefully easier to follow.

I tried the instruction from the link written above and still couldn't get the workload to execute.
This is the guest output file (`board.pc.com_1.device`) from the simulation. I don't know if the last line that says pthread has failed is important or not:

```
Ubuntu 22.04.2 LTS gem5 ttyS0

gem5 login: root (automatic login)



The programs included with the Ubuntu system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.

Ubuntu comes with ABSOLUTELY NO WARRANTY, to the extent permitted by
applicable law.

11+0 records in
11+0 records out
11264 bytes (11 kB, 11 KiB) copied, 0 s, Infinity B/s
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/video/fbdev/core/sysimgblt.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/video/fbdev/core/sysfillrect.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/video/fbdev/core/syscopyarea.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/video/fbdev/core/fb_sys_fops.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/i2c/algos/i2c-algo-bit.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/media/rc/rc-core.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/media/cec/core/cec.ko 
insmod /lib/modules/5.15.0-124-generic/kernel/drivers/gpu/drm/drm_kms_helper.ko fbdev_emulation=0 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amdkcl.ko 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amd-sched.ko 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amddrm_buddy.ko 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amdxcp.ko 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amdttm.ko 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amddrm_ttm_helper.ko 
insmod /lib/modules/5.15.0-124-generic/updates/dkms/amdgpu.ko ip_block_mask=0x6f ppfeaturemask=0 dpm=0 audio=0 ras_enable=0
pthread_attr_setaffinity_np failed: No such process
ma
```

abmerop:
> > I see, that is likely the issue because it would be a different runtime version. The instructions at https://www.gem5.org/documentation/general_docs/gpu_models/gpufs under "Building GPU applications" are hopefully easier to follow.
> 
> I tried the instruction from the link written above and still couldn't get the workload to execute. This is the guest output file (`board.pc.com_1.device`) from the simulation. I don't know if the last line that says pthread has failed is important or not:
> 
> ```
> Ubuntu 22.04.2 LTS gem5 ttyS0
> 
> gem5 login: root (automatic login)
> 
> 
> 
> The programs included with the Ubuntu system are free software;
> the exact distribution terms for each program are described in the
> individual files in /usr/share/doc/*/copyright.
> 
> Ubuntu comes with ABSOLUTELY NO WARRANTY, to the extent permitted by
> applicable law.
> 
> 11+0 records in
> 11+0 records out
> 11264 bytes (11 kB, 11 KiB) copied, 0 s, Infinity B/s
> insmod /lib/modules/5.15.0-124-generic/kernel/drivers/video/fbdev/core/sysimgblt.ko 
> insmod /lib/modules/5.15.0-124-generic/kernel/drivers/video/fbdev/core/sysfillrect.ko 
> insmod /lib/modules/5.15.0-124-generic/kernel/drivers/video/fbdev/core/syscopyarea.ko 
> insmod /lib/modules/5.15.0-124-generic/kernel/drivers/video/fbdev/core/fb_sys_fops.ko 
> insmod /lib/modules/5.15.0-124-generic/kernel/drivers/i2c/algos/i2c-algo-bit.ko 
> insmod /lib/modules/5.15.0-124-generic/kernel/drivers/media/rc/rc-core.ko 
> insmod /lib/modules/5.15.0-124-generic/kernel/drivers/media/cec/core/cec.ko 
> insmod /lib/modules/5.15.0-124-generic/kernel/drivers/gpu/drm/drm_kms_helper.ko fbdev_emulation=0 
> insmod /lib/modules/5.15.0-124-generic/updates/dkms/amdkcl.ko 
> insmod /lib/modules/5.15.0-124-generic/updates/dkms/amd-sched.ko 
> insmod /lib/modules/5.15.0-124-generic/updates/dkms/amddrm_buddy.ko 
> insmod /lib/modules/5.15.0-124-generic/updates/dkms/amdxcp.ko 
> insmod /lib/modules/5.15.0-124-generic/updates/dkms/amdttm.ko 
> insmod /lib/modules/5.15.0-124-generic/updates/dkms/amddrm_ttm_helper.ko 
> insmod /lib/modules/5.15.0-124-generic/updates/dkms/amdgpu.ko ip_block_mask=0x6f ppfeaturemask=0 dpm=0 audio=0 ras_enable=0
> pthread_attr_setaffinity_np failed: No such process
> ma
> ```

Thanks, I am trying myself now and it seems to this PR has atrophied already in the past 2 weeks after I rebased.  Unfortunately m5 exit is interrupting stdout being dumped so it is hard to see what is going on.

I plan to at least temporarily add a `sleep 1;` to the run command to see if that helps

abmerop:
I've identified the issue - which is PyTorch needs AVX turned on. I forgot to copy that from the old configs. This will take some time, but thank you @mbabaie for testing this PR out!

mbabaie:
> I've identified the issue - which is PyTorch needs AVX turned on. I forgot to copy that from the old configs. This will take some time, but thank you @mbabaie for testing this PR out!

Sounds good. Glad that you figured it out.

powerjg:
> I've identified the issue - which is PyTorch needs AVX turned on. I forgot to copy that from the old configs. This will take some time

If you just have 1 minute @abmerop, could you point us to where in the old config this was set? There's a couple people in my group (Maryam included) who are playing with this and may be able to help out.

abmerop:
> > I've identified the issue - which is PyTorch needs AVX turned on. I forgot to copy that from the old configs. This will take some time
> 
> If you just have 1 minute @abmerop, could you point us to where in the old config this was set? There's a couple people in my group (Maryam included) who are playing with this and may be able to help out.

Absolutely! You will need all of this: https://github.com/gem5/gem5/blob/develop/configs/example/gpufs/system/system.py#L253-L304

Keep in mind that this will probably only work for KVM because of gem5's current x86 AVX support. It *might* work, but KVM is definitely a better start. I attempted writing a little bit before getting distracted, so thank you if you can pick this up!

abmerop:
> > > I've identified the issue - which is PyTorch needs AVX turned on. I forgot to copy that from the old configs. This will take some time
> > 
> > 
> > If you just have 1 minute @abmerop, could you point us to where in the old config this was set? There's a couple people in my group (Maryam included) who are playing with this and may be able to help out.
> 
> Absolutely! You will need all of this: https://github.com/gem5/gem5/blob/develop/configs/example/gpufs/system/system.py#L253-L304
> 
> Keep in mind that this will probably only work for KVM because of gem5's current x86 AVX support. It _might_ work, but KVM is definitely a better start. I attempted writing a little bit before getting distracted, so thank you if you can pick this up!

BTW, we did a deep dive on what gem5 was missing from modern libraries and why AVX is need and found that the rocBLAS.so lib 
which PyTorch uses is using *one* AVX instruction to clear high bits of SSE insts. So this is why I underscored _might_ work.  Props to @mattsinc , his students, and @v-ramadas who helped figure this out.

configs/example/gem5_library/x86-mi300x-gpu.py
BobbyRBruce: @Harshil2107: can you put it on your todo list to create and upload these resources ?

configs/example/gem5_library/x86-mi300x-gpu.py
BobbyRBruce: FYI: we are working on fixing this here at UC Davis

configs/example/gem5_library/x86-mi300x-gpu.py
```diff
@@ -0,0 +1,155 @@
...
+    --app <gpu application>
+```
+
+Example:
+--------
+```
+./build/VEGA_X86/gem5.opt
+    configs/example/gem5_library/x86-viper-gpu.py
+    --image ./gem5-resources/src/x86-ubuntu-gpu-ml/disk-image/x86-ubuntu-gpu-ml
+    --kernel ./gem5-resources/src/x86-ubuntu-gpu-ml/vmlinux-gpu-ml
```
BobbyRBruce, line 50: do you have the sources for these so we can build them and upload them to gem5-resources?

src/python/SConscript
```diff
@@ -178,10 +178,34 @@ PySource('gem5.components.cachehierarchies.ruby.caches.mi_example',
...
+PySource('gem5.components.cachehierarchies.ruby.caches.viper',
+    'gem5/components/cachehierarchies/ruby/caches/viper/sqc.py')
+PySource('gem5.components.cachehierarchies.ruby.caches.viper',
+    'gem5/components/cachehierarchies/ruby/caches/viper/tcc.py')
 PySource('gem5.components.cachehierarchies.ruby.topologies',
     'gem5/components/cachehierarchies/ruby/topologies/__init__.py')
 PySource('gem5.components.cachehierarchies.ruby.topologies',
     'gem5/components/cachehierarchies/ruby/topologies/simple_pt2pt.py')
+
+PySource('gem5.components.devices',
```
BobbyRBruce, line 197: You're using the word "device"/"devices" in this PR to describe the GPU you’re adding to the board. Is this because you intend for this change to be generic and used to add any number of different "devices" to a board, not just GPUs?

I'm just wondering if this is flexible for or we should have 'gem5.components.gpus' and not try to generalize.

src/python/gem5/prebuilt/viper/cpu_cache_hierarchy.py
BobbyRBruce: I think the ViperBoard should be under 'gem5.components.board' -- it's really pre-built. The user still has to declare the processor, memory, cache hierarchy, etc. The specific caches created, I believe, should be considered prebuilt because they provide a cache with everything configured to reasonably sensible defaults.

BobbyRBruce (commented):
I've got some high-level questions that I couldn't figure out scrolling through the source:

1. Can you run a non-Viper protocol cache hierarchy on the board with a VIPER GPU? The board's constructor suggests I can put any cache I want in there, and if that's the case, great, but if not, we should change that to be more explicit.

2. Why does the ViperBoard constructor accept a list of GPU devices? Does this actually support a board with multiple? What's the reason for this?

3. I know we've discussed this somewhere, but I can't quite find out where so sorry for having you explain again: what do you need to run workloads on this board? I'm wondering what "set_workload" function we need to create/extend to get this working? Is it just OS, kernel, GPU kernel?

configs/example/gem5_library/x86-mi300x-gpu.py
abmerop: You might want to coordinate with #1581 , who I believe I trying to accomplish the same goal with X86Board

configs/example/gem5_library/x86-mi300x-gpu.py
abmerop: Yes, these were built running `./build.sh` here: https://github.com/gem5/gem5-resources/tree/stable/src/x86-ubuntu-gpu-ml

src/python/SConscript
abmerop: Either is fine.  I can think of at least one other device at the moment which might have it's own cache hierarchy and that would be a smart NIC. Potentially more if gem5 gets CXL support. Until then either devices or having a gpu directory is fine with me.

src/python/gem5/components/devices/gpus/amdgpu.py
abmerop: Right, I have tested 128B and it works.

src/python/gem5/prebuilt/viper/cpu_cache_hierarchy.py
abmerop: Feel free to reorganize.  I am not sure if I should touch these files if your team is actively working on improving it

src/python/gem5/prebuilt/viper/board.py
abmerop: Yea, these should be removed.

src/python/SConscript
powerjg: I agree that other devices may be attached some day. However, I think we shouldn't be shy of specializing things for a specific type of device (e.g., GPU).

On the CXL front, at least with the way we're doing it, it won't be attached as a device in this way. We're taking the approach of creating a new kind of board (`<ISA>CXLHostBoard`) which takes as a parameter an optional remote memory device. There's just too many special things we need to do on that board to use the same board as what the GPU is using.

In general, since there will be a lot of "device"-specific code, I think it's OK if we continue to specialize and create different boards rather than try to make "one board to rule them all."

I don't have any strong opinion on the layout here. It's up to Matt and Bobby. However, I did want to give a bit of context on what else we're working on.

abmerop:
> I've got some high-level questions that I couldn't figure out scrolling through the source:
> 
> 1. Can you run a non-Viper protocol cache hierarchy on the board with a VIPER GPU? The board's constructor suggests I can put any cache I want in there, and if that's the case, great, but if not, we should change that to be more explicit.
> 2. Why does the ViperBoard constructor accept a list of GPU devices? Does this actually support a board with multiple? What's the reason for this?
> 3. I know we've discussed this somewhere, but I can't quite find out where so sorry for having you explain again: what do you need to run workloads on this board? I'm wondering what "set_workload" function we need to create/extend to get this working? Is it just OS, kernel, GPU kernel?

1. Yes and No. Currently, no, you need GPU_VIPER protocol and X86. There is a `requires` somewhere for this. If #117 is merged you could use a different protocol for the CPU and GPU_VIPER for the GPU.  GPU_VIPER has GPU specific memory flags which are required for applications to run correctly, so using a CPU protocol on the GPU side would not work for many applications
2. I intend to be able to have multiple GPUs, but it is not quite working yet. This was one of the major reasons for making this board as it is much easier to create multiple GPUs when everything is inside one python object
3. You need a disk, kernel, and an application.  The application is the different part, potentially. The current GPU scripts let you use a binary located on the host system, so it does not need to be on the disk image. I'd like to keep this support so the function would need to bundle some shell scripting and the application into one shell script which becomes the readfile for the simulation. It is `make_gpu_app` in board.py:110 in this PR

mbabaie:
> > > I've identified the issue - which is PyTorch needs AVX turned on. I forgot to copy that from the old configs. This will take some time
> > 
> > 
> > If you just have 1 minute @abmerop, could you point us to where in the old config this was set? There's a couple people in my group (Maryam included) who are playing with this and may be able to help out.
> 
> Absolutely! You will need all of this: https://github.com/gem5/gem5/blob/develop/configs/example/gpufs/system/system.py#L253-L304
> 
> Keep in mind that this will probably only work for KVM because of gem5's current x86 AVX support. It _might_ work, but KVM is definitely a better start. I attempted writing a little bit before getting distracted, so thank you if you can pick this up!


Hi @abmerop , I have have added the changes you mentioned here to the code base. Now, it runs both pytorch and simple programs. However, for pytorch programs the gpu's stats are all 0, even though the results of the program is printed in the board.terminal. Have you seen this issue before? The gpu's stats for simple programs (e.g., square) are not 0.

abmerop:
> > > > I've identified the issue - which is PyTorch needs AVX turned on. I forgot to copy that from the old configs. This will take some time
> > > 
> > > 
> > > If you just have 1 minute @abmerop, could you point us to where in the old config this was set? There's a couple people in my group (Maryam included) who are playing with this and may be able to help out.
> > 
> > 
> > Absolutely! You will need all of this: https://github.com/gem5/gem5/blob/develop/configs/example/gpufs/system/system.py#L253-L304
> > Keep in mind that this will probably only work for KVM because of gem5's current x86 AVX support. It _might_ work, but KVM is definitely a better start. I attempted writing a little bit before getting distracted, so thank you if you can pick this up!
> 
> Hi @abmerop , I have have added the changes you mentioned here to the code base. Now, it runs both pytorch and simple programs. However, for pytorch programs the gpu's stats are all 0, even though the results of the program is printed in the board.terminal. Have you seen this issue before? The gpu's stats for simple programs (e.g., square) are not 0.

Hmm, It might be a good idea to make sure pytorch is not falling back to the CPU. On the gem5 side, you can use `--debug-flag=GPUCommandProc` to see a print whenever a GPU task (generic term for kernels + misc. packets) is started.

On the pytorch side you can print `torch.cuda.is_available()` and make sure it returns True.

If it's not the best way to figure out why is to add `dmesg -n8` to the runscript. You can do that in src/python/gem5/prebuilt/viper/board.py in the `make_gpu_app` function and possibly other places

mbabaie:
> > > > > I've identified the issue - which is PyTorch needs AVX turned on. I forgot to copy that from the old configs. This will take some time
> > > > 
> > > > 
> > > > If you just have 1 minute @abmerop, could you point us to where in the old config this was set? There's a couple people in my group (Maryam included) who are playing with this and may be able to help out.
> > > 
> > > 
> > > Absolutely! You will need all of this: https://github.com/gem5/gem5/blob/develop/configs/example/gpufs/system/system.py#L253-L304
> > > Keep in mind that this will probably only work for KVM because of gem5's current x86 AVX support. It _might_ work, but KVM is definitely a better start. I attempted writing a little bit before getting distracted, so thank you if you can pick this up!
> > 
> > 
> > Hi @abmerop , I have have added the changes you mentioned here to the code base. Now, it runs both pytorch and simple programs. However, for pytorch programs the gpu's stats are all 0, even though the results of the program is printed in the board.terminal. Have you seen this issue before? The gpu's stats for simple programs (e.g., square) are not 0.
> 
> Hmm, It might be a good idea to make sure pytorch is not falling back to the CPU. On the gem5 side, you can use `--debug-flag=GPUCommandProc` to see a print whenever a GPU task (generic term for kernels + misc. packets) is started.
> 
> On the pytorch side you can print `torch.cuda.is_available()` and make sure it returns True.
> 
> If it's not the best way to figure out why is to add `dmesg -n8` to the runscript. You can do that in src/python/gem5/prebuilt/viper/board.py in the `make_gpu_app` function and possibly other places

Sounds good. I'll go ahead and try the options you suggested.

I already know the answer to one of them. Once running the sample pytorch program from gem5-resources:
``` 
#!/usr/bin/env python3

import torch
print("GPU available!") if torch.cuda.is_available() else print("No GPU available.")
x = torch.rand(5, 3)
print(x)
```

It was able to see pytorch/gpu is available:
```
GPU available!
tensor([[0.5121, 0.8710, 0.8262],
        [0.4231, 0.5720, 0.4033],
        [0.1641, 0.6688, 0.5657],
        [0.5556, 0.1401, 0.0556],
        [0.8286, 0.3764, 0.4575]])
```

I'll update here with further tests.

powerjg:
There were some things that the UCD folks (primarily @mahyarsamani, @mbabaie, and @Harshil2107) saw that could be improved. Rather than going back and forth, we just took a stab at it. Here's a link to a PR I created in your repo: https://github.com/abmerop/gem5/pull/6 I didn't want to tread on any toes by directly pushing to this PR.

abmerop:
> There were some things that the UCD folks (primarily @mahyarsamani, @mbabaie, and @Harshil2107) saw that could be improved. Rather than going back and forth, we just took a stab at it. Here's a link to a PR I created in your repo: [abmerop#6](https://github.com/abmerop/gem5/pull/6) I didn't want to tread on any toes by directly pushing to this PR.

You can push here directly, I have been holding off making changes while your group was making changes. I can make comments in either place

abmerop:
Rebase with the commit from https://github.com/abmerop/gem5/pull/6 .

If #117 merges first I have another commit to add.  Otherwise, will make another PR or add to #117 for the changes needed there.

configs/example/gem5_library/x86-mi300x-gpu.py
powerjg: @Harshil2107  what is the status of this? We can merge this change as-is, but it's a requirement for the release that we update this to use resources instead of this command-line interface.

src/python/gem5/components/boards/kernel_disk_workload.py
powerjg, line 1: This looks like a rebase error

src/python/gem5/components/boards/kernel_disk_workload.py
```diff
@@ -259,32 +259,3 @@ def _set_readfile_contents(self, readfile_contents: str) -> None:
         file = open(self.readfile, "w+")
         file.write(readfile_contents)
         file.close()
-
```
powerjg: This code generalized your (great) idea. I assume it was a rebase error to delete it.

src/python/gem5/components/cachehierarchies/ruby/caches/viper/corepair_cache.py
powerjg: Just to follow up on this (a month later, sorry...). We should make this have a reasonable default, but not allow "parameterization." If your research is interested in changing the replacement policy, then you should create a new class that enables that parameter. *We cannot have a parameter for everything, everywhere, all the time.* It is not possible to make code like that maintainable.

The purpose of the stdlib is twofold:
1. Have components with reasonable defaults. (This is a work in progress, but the prebuilt boards like this one is a good example.)
2. Enable people to more explicitly and with less room for error specialize the components they are using.

I admit this is a change from the way many people "grew up" with gem5, but from a software engineering perspective, it's going to be good for the project in the long run.

src/python/gem5/components/devices/gpus/amdgpu.py
```diff
@@ -0,0 +1,267 @@
...
+    @classmethod
+    def next_pci_dev(cls):
+        cls._gpu_count += 1  # Use count for this particular type
+        return cls._base_pci_dev + cls._gpu_count - 1
+
+    @classmethod
+    def get_gpu_count(cls):
+        return cls._gpu_count
+
+    def __init__(self, gpu_memory: AbstractMemorySystem):
```
powerjg, line 54: Some day (definitely not before this release), we should make this constructor take a list of shaders, a cache hierarchy, and a memory system. This way the GPU is just like a board.

src/python/gem5/prebuilt/viper/board.py
```diff
@@ -0,0 +1,191 @@
...
+
+        driver_load_command = self.get_devices()[0].get_driver_command()
+        contents = driver_load_command + "\n" + readfile_contents
+
+        return super()._set_readfile_contents(contents)
+
+    # Replicate the capability of the old GPUFS config, which embed a binary
+    # application or script into a bash script setting up the environment and
+    # loading the GPU driver.
+    def make_gpu_app(self, gpu: BaseViperGPU, app: str, debug: bool = False):
```
powerjg, line 178: Here, can you override `set_binary_to_run`? Also, `self` should have a reference to the `gpu`, right? Maybe use `gpu[0]` and warn or document that GPUs must be the same GPU architecture.

powerjg (commented):
A few small things that I think were minor mistakes in rebasing.

src/python/gem5/components/boards/kernel_disk_workload.py
abmerop: Fixed.

src/python/gem5/prebuilt/viper/board.py
abmerop: I am not sure how to do that since this function takes `str` not `BinaryResource`.  As far as I can tell only your team can create resources, right?  The GPU resources currently available are about 3 years old.  They don't know what an MI300X is and in GPU development time they may as well be a century old.  So I don't have a way to test this either

PR #1638, opened 2024-10-07
arch-riscv: Fix CLINT mtime reset handling

rogerchang23424:
The previous https://github.com/gem5/gem5/pull/1617 introduce the CLINT reset feature. When reset, we changed the mtime to 0 and keep mtimecmp unchanged by default, we also need to check mtime & mtimecmp regiter to update the MTI signal. However, the mtime register will be incremented to 1 by `raiseInterruptPin`.

In the PR, we introduced the interrupt ID for CLINT, the mtime will be incremented only if received the RTC signal

### DIFF
--- a/src/dev/riscv/clint.cc
+++ b/src/dev/riscv/clint.cc
@@ -72,7 +72,7 @@ Clint::raiseInterruptPin(int id)
     // Increment mtime when received RTC signal
     uint64_t& mtime = registers.mtime.get();
     if (id == INT_RTC) {
-      mtime++;
+        mtime++;
     }
 
     for (int context_id = 0; context_id < nThread; context_id++) {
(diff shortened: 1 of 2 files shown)


src/dev/riscv/clint.cc
```diff
@@ -69,9 +69,11 @@ Clint::Clint(const Params &params) :
 void
 Clint::raiseInterruptPin(int id)
 {
-    // Increment mtime
+    // Increment mtime when received RTC signal
     uint64_t& mtime = registers.mtime.get();
-    mtime++;
+    if (id == INT_RTC) {
+      mtime++;
+    }
```
powerjg:
```suggestion
    if (id == INT_RTC) {
        mtime++;
    }
```

powerjg (commented):
One minor style thing below... I'll commit it and then squash when we merge.

powerjg (approved):

PR #1643, opened 2024-10-08
arch-x86,arch-arm: Remove static variables in decoders

powerjg:
There were a number of variables in the arm and x86 decoders that are static (e.g., the decode cache). It's a bit interesting that this doesn't cause problems with multiple cores since each core has its own decoder.

However, this causes segfaults if you run different cores on different *host* threads. We are experimenting with running gem5 with multiple host thread (i.e., in parallel), and removing these static variables resolves the segfault.

This change also adds const to any other static variables to ensure that they cannot be modified.

### DIFF
--- a/src/arch/arm/decoder.cc
+++ b/src/arch/arm/decoder.cc
@@ -53,8 +53,6 @@ namespace gem5
 namespace ArmISA
 {
 
-GenericISA::BasicDecodeCache<Decoder, ExtMachInst> Decoder::defaultCache;
-
 Decoder::Decoder(const ArmDecoderParams &params)
     : InstDecoder(params, &data),
       dvmEnabled(params.dvm_enabled),
--- a/src/arch/arm/decoder.hh
+++ b/src/arch/arm/decoder.hh
@@ -94,7 +94,7 @@ class Decoder : public InstDecoder
     enums::DecoderFlavor decoderFlavor;
 
     /// A cache of decoded instruction objects.
-    static GenericISA::BasicDecodeCache<Decoder, ExtMachInst> defaultCache;
+    GenericISA::BasicDecodeCache<Decoder, ExtMachInst> defaultCache;
     friend class GenericISA::BasicDecodeCache<Decoder, ExtMachInst>;
 
     /**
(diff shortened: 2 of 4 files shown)


powerjg:
I'd love some feedback on this. Does any know why these variables were `static`? It seems like `static` was not a good idea.

giactra:
My hunch is that assuming an homogeneous multicore system, the same MachInst would translate into the same instruction object regardless of the CPU id. So not only we would save some storage; in theory one CPU could reuse the caching work done by another (which might be running ahead).
Having said that I never looked into the decoder cache in detail so I might be wrong.

It would be interesting to see if we suffer a host performance penalty when simulating a multicore system when removing the static and making the decoder cache a per-CPU structure...

I understand we should pay the penalty anyway in a parallel implementation of gem5

BobbyRBruce (approved):

powerjg:
> My hunch is that assuming an homogeneous multicore system, the same MachInst would translate into the same instruction object regardless of the CPU id. So not only we would save some storage; in theory one CPU could reuse the caching work done by another (which might be running ahead).

That was my guess too. I also wonder if this was some premature optimization ;)

> It would be interesting to see if we suffer a host performance penalty when simulating a multicore system when removing the static and making the decoder cache a per-CPU structure...

I'll run some tests. Given this change touches structures that are highly utilized in the core of gem5, I think we should go through some extra testing. However, I wanted to get the code out the ASAP so others could weigh in.

powerjg:
Here's the results from our tests. Thanks @erin-le for running this!

We tested Linux boot (no systemd) in *atomic* mode. My hypothesis is that this change will affect atomic more than any other mode. We ran with 8 cores and with 1 core. The 8 core data is below, since that should be more interesting. There was little change the Arm performance, and a ~5% slowdown for x86. However, we only ran these once, so it's possible the that 5% is in the noise.

I *think* this is the worst case, so 5% is probably livable. I'm open to hearing any feedback!

### Arm
before:
```
hostSeconds                                    573.02                       # Real time elapsed on the host (Second)
```

after:
```
hostSeconds                                    574.73                       # Real time elapsed on the host (Second)
```

### x86
before:
```
hostSeconds                                   1198.61                       # Real time elapsed on the host (Second)
```
after:
```
hostSeconds                                   1245.15                       # Real time elapsed on the host (Second)
```

giactra (approved):
LGTM

Issue #1630, opened 2024-10-04
unable to compile gem5


MGThePro:
**Describe the bug**
Errors out during compilation

**Affects version**
Latest commit on both stable branch (b1a44b89c7bae73fae2dc547bc1f871452075b85) and develop (6a24b69a972684544385ef7abd1e96c3644c1bb4)

**gem5 Modifications**
none

**To Reproduce**

1. Compile gem5 on an up to date linux distro (in my case Arch Linux) with command `scons build/X86/gem5.opt`

**Terminal Output**

```shell
 [     CXX] src/arch/x86/cpuid.cc -> X86/arch/x86/cpuid.o
Generating LALR tables
In file included from src/arch/x86/regs/int.hh:44,
                 from src/arch/x86/emulenv.hh:41,
                 from src/arch/x86/emulenv.cc:38:
src/cpu/reg_class.hh:168:7: error: ‘class gem5::RegClassOps’ has virtual functions and accessible non-virtual destructor [-Werror=non-virtual-dtor]
  168 | class RegClassOps
      |       ^~~~~~~~~~~
src/arch/x86/regs/int.hh:105:7: error: base class ‘class gem5::RegClassOps’ has accessible non-virtual destructor [-Werror=non-virtual-dtor]
  105 | class FlatIntRegClassOps : public RegClassOps
      |
```
With a lot more affected classes, but always the same error message (Probably too long for github, but I could also send the whole thing if necessary).

**Host Operating System**
Arch Linux

**Host ISA**
X86-64

**Compiler used**
gcc 14.2.1. Also tried with gcc 13.3.0 as well as clang 18.1.8 with the same issue

erin-le:
Hello, could you send the entire error message?

MGThePro:
```
In file included from src/base/stats/group.hh:46,
                 from src/sim/sim_object.hh:53,
                 from src/arch/generic/decoder.hh:37,
                 from src/arch/generic/decoder.cc:28:
src/base/stats/units.hh:120:7: error: ‘class gem5::statistics::units::Base’ has virtual functions and accessible non-virtual destructor [-Werror=non-virtual-dtor]
  120 | class Base
      |       ^~~~
src/base/stats/units.hh:126:7: error: base class ‘class gem5::statistics::units::Base’ has accessible non-virtual destructor [-Werror=non-virtual-dtor]
  126 | class Cycle : public Base
      |       ^~~~~
[...]

PR #1646, opened 2024-10-08
util-docker,tests: Add compiler tests & Dockerfiles for GCC 14

erin-le:
This commit adds gcc 14 to the compiler tests and Dockerfiles.

### DIFF
--- a/.github/workflows/compiler-tests.yaml
+++ b/.github/workflows/compiler-tests.yaml
@@ -13,8 +13,8 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                image: [gcc-version-13, gcc-version-12, gcc-version-11, gcc-version-10, clang-version-18, clang-version-17, clang-version-16, clang-version-15,
-                    clang-version-14, ubuntu-22.04_all-dependencies, ubuntu-24.04_all-dependencies, ubuntu-24.04_min-dependencies]
+                image: [gcc-version-14, gcc-version-13, gcc-version-12, gcc-version-11, gcc-version-10, clang-version-18, clang-version-17, clang-version-16,
+                    clang-version-15, clang-version-14, ubuntu-22.04_all-dependencies, ubuntu-24.04_all-dependencies, ubuntu-24.04_min-dependencies]
                 opts: [.opt, .fast]
         runs-on: [self-hosted, linux, x64]
         timeout-minutes: 2880 # 48 hours
@@ -32,7 +32,7 @@ jobs:
             matrix:
                 gem5-compilation: [ARM, ARM_MESI_Three_Level, ARM_MESI_Three_Level_HTM, ARM_MOESI_hammer, Garnet_standalone, MIPS, 'NULL', NULL_MESI_Two_Level,
                     NULL_MOESI_CMP_directory, NULL_MOESI_CMP_token, NULL_MOESI_hammer, POWER, RISCV, SPARC, X86, X86_MI_example, X86_MOESI_AMD_Base, VEGA_X86]
-                image: [gcc-version-13, clang-version-18]
+                image: [gcc-version-14, clang-version-18]
                 opts: [.opt]
         runs-on: [self-hosted, linux, x64]
         timeout-minutes: 2880 # 48 hours
--- a/util/dockerfiles/docker-bake.hcl
+++ b/util/dockerfiles/docker-bake.hcl
@@ -125,7 +125,8 @@ group "gcc-compilers" {
     "gcc-version-10",
     "gcc-version-11",
     "gcc-version-12",
-    "gcc-version-13"
+    "gcc-version-13",
+    "gcc-version-14"
   ]
 }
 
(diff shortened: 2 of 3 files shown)


.github/workflows/compiler-tests.yaml
```diff
@@ -13,8 +13,8 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                image: [gcc-version-13, gcc-version-12, gcc-version-11, gcc-version-10, clang-version-18, clang-version-17, clang-version-16, clang-version-15,
```
BobbyRBruce, line 16: You've accidentally removed `clang-version-15` from the image list.

util/dockerfiles/docker-bake.hcl
```diff
@@ -169,6 +170,16 @@ target "gcc-version-13" {
...
+target "gcc-version-14" {
+  inherits = ["common"]
+  annotations = ["index,manifest:org.opencontainers.image.description=An image with all dependencies for building gem5 with a GCC v14 compiler."]
+  args = {
+    version = "14"
+  }
+  context = "gcc-compiler"
+  tags = ["${IMAGE_URI}/gcc-version-14:${TAG}"]
+}
+
```
BobbyRBruce, line 182: This all looks good to me but did you build the image locally to mature sure this all works? 

To build this image you'd execute `docker buildx bake gcc-version-14` in the "util/dockerfiles" directory. This does require a machine with docker and some other dependencies installed. If you don't have one already get Mahyar to set you up to user docker on amarillo.

If you had permission to push to the GitHub Container registry you'd run `docker build xbake gcc-version-14 --push`, though setting this up a little annoying so I'll do this right now to save you the hassle. This is more just an "FYI" that these steps are necessary.


**Note:** as i had to build this to upload the image, I've verified it builds just fine :)

util/dockerfiles/gcc-compiler/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM  ghcr.io/gem5/ubuntu-24.04_all-dependencies:latest
 LABEL org.opencontainers.image.source=https://github.com/gem5/gem5
 LABEL org.opencontainers.image.licenses=BSD-3-Clause
 
-ARG version=13 # Version of GCC to install in this image. Default is 13.
+ARG version=14 # Version of GCC to install in this image. Default is 14.
```
BobbyRBruce, line 6: This is fine, but  just so you know, it's not a requirement that the default is the highest supported. I honestly dosn't matter what the default value is. There probably shouldn't be a defaulk as any user wanting to create image from this should know what version they and be blocked by an error when they don't specify it explicitly... but this isn't particularly important right now.

.github/workflows/compiler-tests.yaml
BobbyRBruce, line 1: In a separate commit or PR, can you can look into doing the same for `clang-version-19`? GCC v14 is the latest gcc version and Clang v19 is the last clang version. It's always nice to save we support the latest if possible.

BobbyRBruce (changes requested):
I've uploaded the docker image (ghcr.io/gem5/gcc-version-14:latest). 

I have one fix suggestion in this review that should be addressed.

.github/workflows/compiler-tests.yaml
erin-le: It's still there, it just got moved to the next line by the YAML formatter

BobbyRBruce: oh i see! Then this PR is perfect then. ! Thanks @erin-le

BobbyRBruce (approved):

erin-le:
I looked into adding clang-version-19, but since Ubuntu doesn't have a package available for clang-19 yet, I wasn't able to build the image.

powerjg:
When this is committed, I suggest changing the message to "Add dockerfiles for GCC 14" The current commit message sounds like gem5 is getting support for gcc 14, but IIRC, that's not the case right now.

powerjg (approved):

BobbyRBruce:
> I looked into adding clang-version-19, but since Ubuntu doesn't have a package available for clang-19 yet, I wasn't able to build the image.

That's fine, this is often the case. If you can't install it via apt in 24.04 then let's not bother supporting it.

BobbyRBruce:
> When this is committed, I suggest changing the message to "Add dockerfiles for GCC 14" The current commit message sounds like gem5 is getting support for gcc 14, but IIRC, that's not the case right now.

This PR also adds gcc-14 tests. Doesn't that mean we support GCC 14?

Harshil2107 (approved):
lgtm

powerjg:
> This PR also adds gcc-14 tests. Doesn't that mean we support GCC 14?

I'd say "no" #1630 

It adds tests, but do they pass?

BobbyRBruce:
> > This PR also adds gcc-14 tests. Doesn't that mean we support GCC 14?
> 
> I'd say "no" #1630
> 
> It adds tests, but do they pass?

Unknown until we merge this PR. 

To stop a chicken and problem arising from this conversation, do you want to rename this PR "util-docker: Add compiler tests & dockerfiles for GCC 14"?

PR #1649, opened 2024-10-09
stdlib,ruby: Enable resetting version numbers

powerjg:
Ruby requires each machine type to have a continuous set of version numbers starting at 0. We were hiding this from users/developers by using a Python class variable in the stdlib. Unfortunately, with multiple ruby systems this doesn't work anymore.

As a stop-gap this change adds "resetting" these versions to the beginning of `incorporate_caches`. It would be better to fix this in the C++ code (and assign these numbers in C++ probably via the RubySystem), but that's a bigger change than is needed right now.


powerjg:
Note: With this change I can run multiple systems/nodes with their own RubySystem objects

BobbyRBruce (approved):

abmerop (commented):
This works, although my approach when writing the VIPER stuff was to not use these static methods (or whatever python calls them). In the case of VIPER the number of cache controllers is known when they are created, so the version could just be the index value in a for loop from range() or enumerate(). 

I don't know if that is universally true but I considered the versionCount() etc. to be static variables which I did not want to reintroduce in brand new code

powerjg:
I'm going to merge this for now. I think there's a better long-term solution, but I don't think this change hurts anything, either.

PR #1650, opened 2024-10-10
mem-ruby: Remove unused variables/mark [maybe unused]

BobbyRBruce:
PR https://github.com/gem5/gem5/pull/1453 left some unused variables in the ruby code that triggered
"unused variable" warnings found compiling ALL/gem5.opt to use the CHI
protocol. These have been removed.

`p_ruby_system` in the `Message` class also triggered an unused
variable warning but is, in fact, used. The `[[maybe_unused]]`
annotation is not suitable here as though I found changing it from
a private to a protected class variable solved the issue.

### DIFF
--- a/src/mem/ruby/slicc_interface/Message.hh
+++ b/src/mem/ruby/slicc_interface/Message.hh
@@ -66,8 +66,7 @@ class Message
         : m_block_size(block_size),
           m_time(curTime),
           m_LastEnqueueTime(curTime),
-          m_DelayedTicks(0), m_msg_counter(0),
-          p_ruby_system(rs)
+          m_DelayedTicks(0), m_msg_counter(0)
     { }
 
     Message(const Message &other) = default;
@@ -135,9 +134,6 @@ class Message
     // Variables for required network traversal
     int incoming_link;
     int vnet;
-
-    // Needed to call MacheinType_base_count/level
-    const RubySystem *p_ruby_system = nullptr;
 };
 
 inline bool
--- a/src/mem/ruby/common/DataBlock.cc
+++ b/src/mem/ruby/common/DataBlock.cc
@@ -60,7 +60,6 @@ DataBlock::DataBlock(const DataBlock &cp)
     m_data = new uint8_t[m_block_size];
     memcpy(m_data, cp.m_data, m_block_size);
     m_alloc = true;
-    m_block_size = m_block_size;
     // If this data block is involved in an atomic operation, the effect
     // of applying the atomic operations on the data block are recorded in
     // m_atomicLog. If so, we must copy over every entry in the change log
--- a/src/mem/ruby/structures/BankedArray.hh
+++ b/src/mem/ruby/structures/BankedArray.hh
@@ -51,7 +51,6 @@ class BankedArray
     Tick clockPeriod = 0;
     unsigned int bankBits;
     unsigned int startIndexBit;
-    RubySystem *m_ruby_system;
 
     class AccessRecord
     {


BobbyRBruce:
I really have no idea why the VEGA build is failing because of this. If anyone has any ideas let me know.

The compilation error:

```txt
[     CXX] VEGA_X86/cpu/testers/gpu_ruby_test/GpuWavefront.py.cc -> .o
 [     CXX] VEGA_X86/python/_m5/param_DmaThread.cc -> .o
 [     CXX] VEGA_X86/python/_m5/param_ProtocolTester.cc -> .o
In file included from src/mem/ruby/network/MessageBuffer.hh:64,
                 from src/mem/ruby/system/RubyPort.hh:49,
                 from src/cpu/testers/gpu_ruby_test/protocol_tester.hh:58,
                 from src/cpu/testers/gpu_ruby_test/tester_thread.hh:41,
                 from src/cpu/testers/gpu_ruby_test/dma_thread.hh:35,
                 from build/VEGA_X86/python/_m5/param_DmaThread.cc:17:
src/mem/ruby/slicc_interface/Message.hh:140:56: error: ‘maybe_unused’ attribute ignored [-Werror=attributes]
  140 |     [[maybe_unused]] const RubySystem *p_ruby_system = nullptr;
      |                                                        ^~~~~~~
```

abmerop:
I think you need to remove the [[maybe_unused]].  I believe it does not make sense to use that attribute with a const member variable because it can only / must be assigned by the constructor.  The compiler is just warning you about that.

The other two changes look good though

BobbyRBruce:
> I think you need to remove the [[maybe_unused]]. I believe it does not make sense to use that attribute with a const member variable because it can only / must be assigned by the constructor. The compiler is just warning you about that.
> 
> The other two changes look good though

The only way I could get the unused variable private member variables warning to stop on my machine and not use `[[maybe_unused]]` was to change the `p_ruby_system` from being private to being protected. I _assume_ this is relatively harmless?

src/mem/ruby/slicc_interface/Message.hh
```diff
@@ -126,6 +127,9 @@ class Message
   protected:
     int m_block_size = 0;
 
+    // Needed to call MacheinType_base_count/level
+    const RubySystem *p_ruby_system = nullptr;
```
powerjg: As far as I can tell, this is *unused*. I grepped for it and I could only find the following occurences:

```
src/mem/ruby/slicc_interface/Message.hh:          p_ruby_system(rs)
src/mem/ruby/slicc_interface/Message.hh:    const RubySystem *p_ruby_system = nullptr;
```

Can we just delete it?

powerjg (approved):

src/mem/ruby/slicc_interface/Message.hh
BobbyRBruce: I'm pretty sure it's used. Delete that line and it won't compile. I'm certain I checked this.

powerjg: It works for me assuming I remove both lines. I imagine it worked to move it to "protected" because the compiler can't guarantee that a subclass doesn't use it.

powerjg (changes requested):

src/mem/ruby/slicc_interface/Message.hh
BobbyRBruce: I stand corrected. I don't know why I had that assumption originally. The latest push to the PR has this line removed.

powerjg (commented):
LGTM.

Let's wait for @abmerop to review before committing. I want to make sure that this isn't going to conflict with any other changes in the pipeline.

PR #1655, opened 2024-10-10
arch-arm: Add support of AArch32 VRINTN/X/A/Z/M/P instructions.

giactra:
Add decoder and function of AArch32 VRINTN, VRINTX, VRINTA, VRINTZ, VRINTM, and VRINTP (Advanced SIMD) instructions. Support both 16-bit and 32-bit variants.

Add vfpFPRint in vfp.hh to perform the behavior of round-to-integer.

Only support A32 encoding.

Change-Id: Icb9b6f71edf16ea14a439e15c480351cd8e1eb88

### DIFF
--- a/src/arch/arm/insts/vfp.hh
+++ b/src/arch/arm/insts/vfp.hh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010-2013, 2019 ARM Limited
+ * Copyright (c) 2010-2013, 2019, 2024 Arm Limited
  * All rights reserved
  *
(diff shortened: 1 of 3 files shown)


powerjg (approved):
Any chance we can get some tests for these? ;)

giactra:
> Any chance we can get some tests for these? ;)

I wish! I think the only viable solution is to either reuse some existing tests of some compute libraries ported to neon, or to write some simple tests ourselves with C intrinsics https://github.com/ARM-software/acle/tree/main.

PR #1656, opened 2024-10-10
Change debug flag generation to more portable code

heshpdx:
Modifies union construction in the debug directory so output is more amenable to alternative compilers.

Verified that this change produces code that builds with clang, gcc, msvc, nvhpc, aocc, icc, openxl, and cray hpc.

### DIFF
--- a/build_tools/debugflaghh.py
+++ b/build_tools/debugflaghh.py
@@ -100,13 +100,16 @@ if components:
 inline union ${{args.name}}
 {
     ~${{args.name}}() {}
-    CompoundFlag ${{args.name}} = {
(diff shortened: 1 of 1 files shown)


heshpdx:
These were the kinds of errors seen in MSVC, which this patch fixes.

```
debug/Decoder.hh(24): error C2461: 'gem5::debug::unions::Decoder': constructor syntax missing formal parameters
debug/Decoder.hh(31): error C7624: Type name 'gem5::debug::unions::Decoder' cannot appear on the right side of a class member access expression
```

giactra:
I haven't reviewed the change yet though I noticed this PR is targeting stable branch.. You should close this and target develop instead

PR #1659, opened 2024-10-10
cpu-o3: Use the generic cache library to build store sets

pranith:
This allows us to build an associative store set structure and also experiment with various replacement policies for the store set.

Change-Id: I18ddad2346b82423c01ef9117a5d3d85f7039be7

### DIFF
--- a/src/cpu/o3/BaseO3CPU.py
+++ b/src/cpu/o3/BaseO3CPU.py
@@ -42,8 +42,11 @@ from m5.objects.BaseCPU import BaseCPU
 # from m5.objects.O3Checker import O3Checker
 from m5.objects.BranchPredictor import *
 from m5.objects.FUPool import *
+from m5.objects.IndexingPolicies import *
+from m5.objects.ReplacementPolicies import *
 from m5.params import *
 from m5.proxy import *
+from m5.SimObject import *
 
 
 class SMTFetchPolicy(ScopedEnum):
@@ -153,7 +156,19 @@ class BaseO3CPU(BaseCPU):
         "should be invalidated",
     )
     LFSTSize = Param.Unsigned(1024, "Last fetched store table size")
-    SSITSize = Param.Unsigned(1024, "Store set ID table size")
+    SSITSize = Param.MemorySize("1024", "Store set ID table size")
+    SSITAssoc = Param.Unsigned(1, "SSIT table associativity")
+    SSITReplPolicy = Param.BaseReplacementPolicy(
+        LRURP(), "SSIT replacement policy"
+    )
+    SSITIndexingPolicy = Param.BaseIndexingPolicy(
+        SetAssociative(
+            size=Parent.SSITSize * 4,
+            assoc=Parent.SSITAssoc,
+            entry_size=4,
+        ),
+        "SSIT indexing policy",
+    )
 
     numRobs = Param.Unsigned(1, "Number of Reorder Buffers")
 
--- a/src/cpu/o3/store_set.hh
+++ b/src/cpu/o3/store_set.hh
@@ -34,8 +34,17 @@
 #include <utility>
 #include <vector>
 
+#include "base/cache/associative_cache.hh"
+#include "base/cache/cache_entry.hh"
+#include "base/named.hh"
 #include "base/types.hh"
 #include "cpu/inst_seq.hh"
+
+class BaseIndexingPolicy;
+
+namespace replacement_policy {
+class Base;
+}
 
 namespace gem5
 {
@@ -52,6 +61,7 @@ struct ltseqnum
     }
 };
 
+
 /**
  * Implements a store set predictor for determining if memory
  * instructions are dependent upon each other.  See paper "Memory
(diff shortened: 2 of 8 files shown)


src/cpu/o3/store_set.hh
```diff
@@ -61,21 +123,52 @@ struct ltseqnum
...
 {
+  protected:
+  std::string _name;
+
   public:
-    typedef unsigned SSID;
+    typedef Addr SSID;
 
   public:
+    std::string name() const { return _name; }
```
odanrc: Make StoreSet inherit from Named instead

src/cpu/o3/store_set.hh
```diff
@@ -61,21 +123,52 @@ struct ltseqnum
...
+
+    class SSITEntry : public CacheEntry
+    {
+        SSID _ssid;
+      public:
+        using IndexingPolicy = gem5::SSITIndexingPolicy;
```
odanrc: I don´t think we need all this overhead, as both classes behave the same. I think this suffices:
```
class SSITEntry : public CacheEntry
{
    SSID _ssid;
public:
    void setSSID(SSID id) { _ssid = id; }
    SSID getSSID(void) const { return _ssid; }
};
```

src/cpu/o3/store_set.hh
```diff
@@ -34,12 +34,73 @@
...
 #include "base/types.hh"
 #include "cpu/inst_seq.hh"
+#include "mem/cache/tags/indexing_policies/base.hh"
+#include "params/SSITIndexingPolicy.hh"
+#include "params/SSITSetAssociative.hh"
 
 namespace gem5
 {
 
+class SSITTagTypes
```
odanrc: There is no need for this any of these new classes, as they are entirely duplicated code. You should use SetAssociative (src/mem/cache/tags/indexing_policies/set_associative.hh) and Addr instead.

odanrc (changes requested):

src/cpu/o3/store_set.hh
pranith: I cannot use SetAssociative from mem-cache as that is organized based on cache size whereas store set is mainly organized by the number of entries. And because of this, I cannot use AddrTypes. I need to define my own SSITTagType to use the right indexing policy param.

I tried adding another constructor to SetAssociative that takes 'num_entries' but couldn't make it work.

Please let me know if you think we should still be able to use the existing classes.

odanrc: I see the following in the python file:
```
SSITSetAssociative(
            num_entries=Parent.SSITSize,
```

Can´t we just do:
```
SetAssociative(
            size=Parent.SSITSize,
            entry_size=1,
```
?

src/cpu/o3/BaseO3CPU.py
```diff
@@ -58,6 +61,31 @@ class CommitPolicy(ScopedEnum):
...
+
+
+class SSITSetAssociative(SSITIndexingPolicy):
+    type = "SSITSetAssociative"
+    cxx_class = "gem5::SSITSetAssociative"
+    cxx_header = "cpu/o3/store_set.hh"
+
+    # Get the number of entries in the SSIT from the parent
+    num_entries = Param.Unsigned(
+        Parent.numEntries, "Number of entries in the SSIT"
```
odanrc: I grepped numEntries in python files in src/ and I could not find anything. Does it exist somewhere?

src/cpu/o3/store_set.hh
pranith: We need to shift the PC to ignore the lower 2 bits so as to index the SSIT using instruction granularity (esp. for ARM). If we set entry_size to 1, we don't do this.

src/cpu/o3/BaseO3CPU.py
pranith: Not really, I am manually setting num_entries to SSIT entries. I can change the default to that if you think so.

src/cpu/o3/store_set.hh
pranith: Done.

src/cpu/o3/BaseO3CPU.py
odanrc: Yeah, we should not use nonexistent variables

src/cpu/o3/store_set.hh
odanrc: Then we adjust the parameters accordingly, as that is exactly what those params are for:
SetAssociative(
            size=Parent.SSITSize * 4,
            entry_size=4,

pranith: Here 'size' is of MemoryType. We have to declare SSITSize also as a memory type. If we do this, we cannot multiply 'SSITSize * 4' as they are of different types. Do you know how to get this to work?

odanrc: I have tried locally the following, and it compiles:

```
    SSITSize = Param.MemorySize("1024", "Store set ID table size")
    SSITAssoc = Param.Unsigned(1, "SSIT table associativity")
    SSITIndexingPolicy = Param.BaseIndexingPolicy(
        SetAssociative(
            size=Parent.SSITSize * 4,
            assoc=Parent.SSITAssoc,
            entry_size=1,
        ),
        "SSIT indexing policy",
    )
```

What is your error message?

A quick possible fix would be to revert the order to 4 * SSITSize. I remember running into errors in the past because multiplication of parameters was not implemented for all orders of a combination of parameters I wanted to use.

src/cpu/o3/BaseO3CPU.py
pranith: This is updated.

src/cpu/o3/store_set.hh
pranith: @odanrc Do you have any further comments on this patch? Please let me know since I want to get this merged for the next imminent release. Thank you!

odanrc (commented):
Sorry, I forgot to click to send

src/cpu/o3/store_set.hh
pranith: Done.

pranith:
I updated SSITSize to be of MemorySize and am using SetAssociative instead of the custom indexing policy. I think we should improve this in a separate PR that MemorySize should be able to take Integer instead of only string. Please let me know if there are any other comments. Thanks for the review!

odanrc:
> I updated SSITSize to be of MemorySize and am using SetAssociative instead of the custom indexing policy. I think we should improve this in a separate PR that MemorySize should be able to take Integer instead of only string. Please let me know if there are any other comments. Thanks for the review!

I don´t see the change removing the custom indexing policy. Only SSITTagTypes was removed. Did you by any chance forget to push some commits?

Regarding MemorySize, the python param types are declared in src/python/m5/params.py, IIRC. I think allowing values besides strings makes sense, but it seems to be too much effort for little benefit.

pranith:
Oops...
> 
> I don´t see the change removing the custom indexing policy. Only SSITTagTypes was removed. Did you by any chance forget to push some commits?

Looks like I pushed an intermediate patch instead of the final one. Updated the branch with the final changes...

> 
> Regarding MemorySize, the python param types are declared in src/python/m5/params.py, IIRC. I think allowing values besides strings makes sense, but it seems to be too much effort for little benefit.

Yes, but it will make it more consistent. Here LFSTSize is an integer and SSTISize is a string and it seems a bit off that size is different type for each.

src/cpu/o3/store_set.hh
```diff
@@ -34,8 +34,11 @@
 #include <utility>
 #include <vector>
 
+#include "base/cache/associative_cache.hh"
+#include "base/named.hh"
 #include "base/types.hh"
 #include "cpu/inst_seq.hh"
+#include "mem/cache/tags/indexing_policies/base.hh"
```
odanrc: Nit: this include, as well as mem/cache/tags/replacement_policies/base.hh can be moved to the .cc by having a forward declaration here instead

src/cpu/o3/store_set.hh
```diff
@@ -34,8 +34,11 @@
 #include <utility>
 #include <vector>
 
+#include "base/cache/associative_cache.hh"
```
odanrc, line 37: We are missing the include for base/cache/cache_entry.hh

odanrc (approved):

src/cpu/o3/store_set.hh
pranith: Done.

src/cpu/o3/store_set.hh
pranith: Added.

odanrc (approved):

pranith:
@powerjg @BobbyRBruce Can you please merge this PR?

BobbyRBruce (approved):

PR #1661, opened 2024-10-11
arch-arm: Rewrite the ArmTLB storage to use an AssociativeCache

giactra:
With this PR we replace the TlbEntry storage within the TLB from an
array of entries with a custom hardcoded FA indexing policy and LRU
replacement policy, into the flexible SetAssociative cache.

### DIFF
--- a/src/arch/arm/ArmTLB.py
+++ b/src/arch/arm/ArmTLB.py
@@ -36,6 +36,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 from m5.objects.BaseTLB import BaseTLB
+from m5.objects.ReplacementPolicies import LRURP
 from m5.params import *
 from m5.proxy import *
 from m5.SimObject import SimObject
@@ -71,6 +72,16 @@ class ArmTLB(BaseTLB):
     cxx_header = "arch/arm/tlb.hh"
     sys = Param.System(Parent.any, "system object parameter")
     size = Param.Int(64, "TLB size")
+    assoc = Param.Int(
+        Self.size, "Associativity of the TLB. Fully Associative by default"
+    )
+    indexing_policy = Param.TLBIndexingPolicy(
+        TLBSetAssociative(assoc=Parent.assoc, num_entries=Parent.size),
+        "Indexing policy of the TLB",
+    )
+    replacement_policy = Param.BaseReplacementPolicy(
+        LRURP(), "Replacement policy of the TLB"
+    )
     is_stage2 = Param.Bool(False, "Is this a stage 2 TLB?")
 
     partial_levels = VectorParam.ArmLookupLevel(
--- a/src/arch/arm/ArmTLB.py
+++ b/src/arch/arm/ArmTLB.py
@@ -1,6 +1,6 @@
 # -*- mode:python -*-
 
-# Copyright (c) 2009, 2013, 2015, 2021 Arm Limited
+# Copyright (c) 2009, 2013, 2015, 2021, 2024 Arm Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
@@ -43,6 +43,26 @@ from m5.SimObject import SimObject
 
 class ArmLookupLevel(Enum):
     vals = ["L0", "L1", "L2", "L3"]
+
+
+class TLBIndexingPolicy(SimObject):
+    type = "TLBIndexingPolicy"
+    abstract = True
+    cxx_class = "gem5::IndexingPolicyTemplate<gem5::ArmISA::TLBTypes>"
+    cxx_header = "arch/arm/pagetable.hh"
+    cxx_template_params = ["class Types"]
+
+    # Get the size from the parent (cache)
+    num_entries = Param.Int(Parent.size, "number of TLB entries")
+
+    # Get the associativity
+    assoc = Param.Int(Parent.assoc, "associativity")
+
+
+class TLBSetAssociative(TLBIndexingPolicy):
+    type = "TLBSetAssociative"
+    cxx_class = "gem5::ArmISA::TLBSetAssociative"
+    cxx_header = "arch/arm/pagetable.hh"
 
 
 class ArmTLB(BaseTLB):
--- a/src/arch/arm/pagetable.cc
+++ b/src/arch/arm/pagetable.cc
@@ -489,5 +489,14 @@ getPageTableOps(GrainSize trans_granule)
     }
 }
 
+TLBTypes::KeyType::KeyType(const TlbEntry &entry)
+  : va(entry.vpn << entry.N), pageSize(entry.N), size(0),
+    asn(entry.asid), ignoreAsn(false),
+    vmid(entry.vmid), ss(entry.ss),
+    functional(false),
+    targetRegime(entry.regime),
+    mode(BaseMMU::Read)
+{}
+
 } // namespace ArmISA
 } // namespace gem5
--- a/src/arch/arm/pagetable.hh
+++ b/src/arch/arm/pagetable.hh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2012-2013, 2021, 2023-2024 Arm Limited
+ * Copyright (c) 2010, 2012-2013, 2021, 2023-2024 ARM Limited
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
@@ -49,6 +49,7 @@
 #include "arch/generic/mmu.hh"
 #include "enums/TypeTLB.hh"
 #include "enums/ArmLookupLevel.hh"
+#include "mem/cache/replacement_policies/replaceable_entry.hh"
 #include "sim/serialize.hh"
 
 namespace gem5
@@ -162,7 +163,7 @@ struct V8PageTableOps64k : public PageTableOps
 };
 
 // ITB/DTB table entry
-struct TlbEntry : public Serializable
+struct TlbEntry : public ReplaceableEntry, Serializable
 {
   public:
     typedef enums::ArmLookupLevel LookupLevel;
@@ -296,6 +297,46 @@ struct TlbEntry : public Serializable
         // no restrictions by default, hap = 0x3
 
         // @todo Check the memory type
+    }
+    TlbEntry(const TlbEntry &rhs) = default;
+    TlbEntry& operator=(TlbEntry rhs)
+    {
+        swap(rhs);
+        return *this;
+    }
+
+    void
+    swap(TlbEntry &rhs)
+    {
+        std::swap(pfn, rhs.pfn);
+        std::swap(size, rhs.size);
+        std::swap(vpn, rhs.vpn);
+        std::swap(attributes, rhs.attributes);
+        std::swap(lookupLevel, rhs.lookupLevel);
+        std::swap(asid, rhs.asid);
+        std::swap(vmid, rhs.vmid);
+        std::swap(tg, rhs.tg);
+        std::swap(N, rhs.N);
+        std::swap(innerAttrs, rhs.innerAttrs);
+        std::swap(outerAttrs, rhs.outerAttrs);
+        std::swap(ap, rhs.ap);
+        std::swap(hap, rhs.hap);
+        std::swap(domain, rhs.domain);
+        std::swap(mtype, rhs.mtype);
+        std::swap(longDescFormat, rhs.longDescFormat);
+        std::swap(global, rhs.global);
+        std::swap(valid, rhs.valid);
+        std::swap(ns, rhs.ns);
+        std::swap(ss, rhs.ss);
+        std::swap(regime, rhs.regime);
+        std::swap(type, rhs.type);
+        std::swap(partial, rhs.partial);
+        std::swap(nonCacheable, rhs.nonCacheable);
+        std::swap(shareable, rhs.shareable);
+        std::swap(outerShareable, rhs.outerShareable);
+        std::swap(xn, rhs.xn);
+        std::swap(pxn, rhs.pxn);
+        std::swap(xs, rhs.xs);
     }
 
     void
--- a/src/arch/arm/pagetable.hh
+++ b/src/arch/arm/pagetable.hh
@@ -177,6 +177,16 @@ struct TlbEntry : public ReplaceableEntry, Serializable
 
     struct Lookup
     {
+        Lookup() = default;
+        explicit Lookup(const TlbEntry &entry)
+          : va(entry.vpn << entry.N), pageSize(entry.N), size(0),
+            asn(entry.asid), ignoreAsn(false),
+            vmid(entry.vmid), ss(entry.ss),
+            functional(false),
+            targetRegime(entry.regime),
+            mode(BaseMMU::Read)
+        {}
+
         // virtual address
         Addr va = 0;
         // page size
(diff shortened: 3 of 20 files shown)


src/arch/arm/ArmTLB.py
```diff
@@ -45,12 +46,42 @@ class ArmLookupLevel(Enum):
     vals = ["L0", "L1", "L2", "L3"]
 
 
+class TlbIndexingPolicy(SimObject):
```
odanrc: Consistency: TLB instead of Tlb

src/arch/arm/ArmTLB.py
```diff
@@ -45,12 +46,42 @@ class ArmLookupLevel(Enum):
...
+    cxx_template_params = ["class Types"]
+
+    # Get the size from the parent (cache)
+    size = Param.Int(Parent.size, "number of TLB entries")
+
+    # Get the associativity
+    assoc = Param.Int(Parent.assoc, "associativity")
+
+
+class TlbSetAssociative(TlbIndexingPolicy):
```
odanrc: Ditto

src/arch/arm/pagetable.hh
```diff
@@ -161,31 +165,19 @@ struct V8PageTableOps64k : public PageTableOps
     LookupLevel lastLevel() const override;
 };
 
-// ITB/DTB table entry
-struct TlbEntry : public Serializable
+struct TlbEntry;
+
+class TlbTypes
```
odanrc: Ditto TLB

src/arch/arm/pagetable.hh
```diff
@@ -161,31 +165,19 @@ struct V8PageTableOps64k : public PageTableOps
...
-        Manager
-    };
+        KeyType() = default;
+        explicit KeyType(const TlbEntry &entry);
 
-    struct Lookup
-    {
         // virtual address
         Addr va = 0;
+        Addr pagesize = Grain4KB;
```
odanrc: Should have been camel case

src/arch/arm/tlb.cc
```diff
@@ -103,62 +134,19 @@ TLB::setTableWalker(TableWalker *table_walker)
...
     const auto mode = lookup_data.mode;
 
-    TlbEntry *retval = match(lookup_data);
+    TlbEntry *retval = nullptr;
+    // We iterate over all stored sizes, starting from the
+    // smallest until the biggest. In this way we prioritize
+    // complete translations over partial translations
+    for (auto pagesize : observedPageSizes) {
+        lookup_data.pagesize = pagesize;
+        if (retval = table.accessEntry(lookup_data); retval)
```
odanrc, line 150: IIRC, the value of an assignment is the assigned value, so is the ´; retval´ really needed?

src/arch/arm/tlb.cc
```diff
@@ -103,62 +134,19 @@ TLB::setTableWalker(TableWalker *table_walker)
...
+TLB::lookup(Lookup lookup_data)
 {
     const auto mode = lookup_data.mode;
 
-    TlbEntry *retval = match(lookup_data);
+    TlbEntry *retval = nullptr;
+    // We iterate over all stored sizes, starting from the
+    // smallest until the biggest. In this way we prioritize
+    // complete translations over partial translations
+    for (auto pagesize : observedPageSizes) {
```
odanrc: Should have been page_size (also, it seems that it can be const auto &)

src/arch/arm/tlb.hh
```diff
@@ -167,7 +185,7 @@ class TLB : public BaseTLB
     /** Lookup an entry in the TLB
      * @return pointer to TLB entry if it exists
      */
-    TlbEntry *lookup(const Lookup &lookup_data);
+    TlbEntry *lookup(Lookup lookup_data);
```
odanrc, line 188: Any particular reason to remove const &?

src/arch/arm/tlb.hh
```diff
@@ -156,9 +164,19 @@ class TLB : public BaseTLB
...
+     * no matching TLB entry (e.g. it has been invalidated),
+     * but if the page size is not in the set, we are certain
+     * there is no associated TLB with that size
+     */
+    std::set<Addr> observedPageSizes;
+
   public:
     using Params = ArmTLBParams;
-    using Lookup = TlbEntry::Lookup;
+    using Lookup = TlbEntry::KeyType;
```
odanrc, line 179: Nit: maybe keeping the Lookup alias increases the difficulty of learning the class

src/arch/arm/pagetable.hh
```diff
@@ -208,6 +200,56 @@ struct TlbEntry : public Serializable
...
+
+    enum class DomainType : std::uint8_t
+    {
+        NoAccess = 0,
+        Client,
+        Reserved,
+        Manager
+    };
+
     // Matching variables
```
odanrc, line 246: Could you replace the corresponding variables by a KeyType variable?

src/arch/arm/tlb.cc
```diff
@@ -240,98 +228,75 @@ TLB::checkPromotion(TlbEntry *entry, BaseMMU::Mode mode)
 
 // insert a new TLB entry
 void
-TLB::insert(TlbEntry &entry)
+TLB::insert(const Lookup &lookup_data, TlbEntry &entry)
```
odanrc, line 234: Why do we add another parameter instead of using entry->keyType?

src/arch/arm/tlb.cc
```diff
@@ -58,8 +58,38 @@ namespace gem5
...
+}
+
+TlbEntry*
+TLB::Table::findEntry(const KeyType &key) const
+{
+    auto candidates = indexingPolicy->getPossibleEntries(key);
+
+    for (auto candidate : candidates) {
+        auto entry = static_cast<TlbEntry*>(candidate);
+        if (entry->N == key.pagesize && entry->match(key)) {
```
odanrc: Why is the entry->N == key.pagesize not part of the match check?

src/arch/arm/tlb.cc
```diff
@@ -58,8 +58,38 @@ namespace gem5
 
 using namespace ArmISA;
 
+TlbEntry*
+TLB::Table::accessEntry(const KeyType &key)
```
odanrc, line 62: I am going to dump my thoughts, but I am unfamiliar with the code, so there may be no action point:

I think we should override the other access function instead. It would be safer, as it guarantees that no one bypasses this functional check. That said, it seems that there is no guarantee that key.functional == entry.key.functional, right?

src/arch/arm/pagetable.hh
giactra: Ack

src/arch/arm/tlb.cc
giactra: It is not but I don't like treating assignment as expressions... I prefer being explicit

src/arch/arm/tlb.cc
giactra: Ack

src/arch/arm/tlb.hh
giactra: Just because we want to make a copy of the data anyway since we iterate over different page sizes withing the lookup method and we need a non cont parameter

src/arch/arm/tlb.cc
giactra: I understand your rationale. However entry.key.functional does not mean anything. The functional boolean is a property of the access, not of the entry. It means: this is not a real workload lookup/access; I just want to get the entry to be able to do a functional VA/PA translation. Therefore

1) Do not update stats
2) Do not affect replacement policy/data

Another solution which lets us avoid overriding access Entry would be to not use accessEntry for functional accesses and handle things differently in the TLB::lookup method. I am not sure it will make things better

src/arch/arm/tlb.cc
giactra: Interesting I will consider this option...

src/arch/arm/pagetable.hh
giactra: It has to be noted that we still use TlbEntry though... What shall we do about it?

src/arch/arm/tlb.cc
giactra: It probably can be done but bear in mind there is more to it:

Basically at the moment the Entry::match method is not considering the page size and the reason for it is that this is quite convenient to be able to match a TLB entry not knowing the size of the page that will be matched (because I cannot tell a priori).
For example I might have a 0x8000_0000 VA and even if I know I use 4K pages, I might match a 4K page, or I could match a Block (a Huge Page in Arm, like 2MB) or I could match a partial translation (Basically an intermediate walk). Those could all be valid matches. Adding pageSize to the Lookup I am able to prioritise them; for example I could start looking for 4K pages first, and only then check for bigger sizes (a partial translation has lower priority).

There are cases where this is not needed like in TLB invalidation where I want to invalidate any potential match. So if add pageSize matching in the entry::match, I basically have to iterate over all observed page sizes when doing a TLB invalidation...

src/arch/arm/tlb.hh
giactra: This PR is already big and I wanted to make it less intrusive and not so hard to review so that I didn't have to change all the Lookup instances (IIRC it is used a lot in the TLBI code). I can add an extra commit, alternatively I will open a new PR later on. Your choice

src/arch/arm/pagetable.hh
odanrc: That one too. Basically, my request was to change any Tlb to TLB, to standardize the naming scheme. I understand if you would rather leave TlbEntry for the future, as it is a class that already existed before this PR, so it may generate some merge conflicts.

src/arch/arm/tlb.hh
odanrc: It can be left for a future PR, no worries

src/arch/arm/tlb.hh
odanrc: Oh, ok. I thought it would be something related to the fact that const references may generate bugged code (https://stackoverflow.com/a/4705871)

src/arch/arm/tlb.cc
odanrc: Could you add some of this info to the code? It seems relevant to help avoid bugs in the future.

src/arch/arm/pagetable.hh
giactra: Amended the TlbTypes. Will update TlbEntry with a separate PR

src/arch/arm/ArmTLB.py
```diff
@@ -45,6 +45,26 @@ class ArmLookupLevel(Enum):
...
 
+class TLBIndexingPolicy(SimObject):
+    type = "TLBIndexingPolicy"
+    abstract = True
+    cxx_class = "gem5::IndexingPolicyTemplate<gem5::ArmISA::TLBTypes>"
+    cxx_header = "arch/arm/pagetable.hh"
+    cxx_template_params = ["class Types"]
+
+    # Get the size from the parent (cache)
+    size = Param.Int(Parent.size, "number of TLB entries")
```
pranith: In other contexts of indexing policies, size is defined to be MemoryType("8KB", "16KB" etc.,). Can we rename this to 'numEntries' to avoid confusion?

src/arch/arm/tlb.cc
```diff
@@ -58,8 +58,38 @@ namespace gem5
...
+{
+    auto entry = findEntry(key);
```
pranith, line 64: This could be:
`
if (!key.functional) {
    entry = accessEntry(key);
} else {
    entry = findEntry(key);
}
`

src/arch/arm/pagetable.hh
```diff
@@ -162,7 +163,7 @@ struct V8PageTableOps64k : public PageTableOps
 };
 
 // ITB/DTB table entry
-struct TlbEntry : public Serializable
+struct TlbEntry : public ReplaceableEntry, Serializable
```
pranith: Minor nit: can we please rename 'N' to be a bit more descriptive like 'pageSizeBits' or similar?

pranith (changes requested):

src/arch/arm/ArmTLB.py
giactra: Makes sense. I was using size for consistency with the Tlb field but I guess there's nothing wrong in doing

`num_entries = Param.Int(Parent.size, "number of TLB entries")`

src/arch/arm/pagetable.hh
giactra: As other comments. This makes a lot of sense but it is something historical and I would open a new PR for that, so that this one keeps being "easy" to review

src/arch/arm/tlb.cc
giactra: Unless I am missing something the first case would imply an endless recursion

src/arch/arm/tlb.cc
giactra: Done!

src/arch/arm/tlb.cc
pranith: Ah, right. Missed that we are in the same function. 😂

pranith (approved):

src/arch/arm/tlb.cc
giactra: So I looked into it and I am not fully convinced.. The issue is that KeyAccess does not convey the tag information only, but it contains access specific variables, like the aforementioned functional mentioned above. This means the TlbEntry might end up with garbage data within it

giactra:
As an update, I added some refactoring commits to make unit-tests compile (sigh)

src/arch/arm/tlb.cc
odanrc: This, for me, is a hint that the KeyType may contain more stuff than it should. Nonetheless, I will disregard this request for now

src/arch/arm/pagetable.cc
```diff
@@ -489,5 +489,14 @@ getPageTableOps(GrainSize trans_granule)
     }
 }
 
+TLBTypes::KeyType::KeyType(const TlbEntry &entry)
```
odanrc, line 492: For a future PR I would recommend fixing this. The TLB entry contains a key type which we want to copy. The proper solution would be to have a KeyType var in TlbEntry and to KeyType(entry.keyType).

odanrc (approved):

src/arch/arm/tlb.cc
giactra: Thanks @odanrc !

powerjg (approved):
lgtm. Does this affect the performance of the simulator at all?

giactra:
> lgtm. Does this affect the performance of the simulator at all?

Yes I am afraid.  I am currently profiling Linux boot trying to fix the performance regression (as you can see from https://github.com/gem5/gem5/pull/1739).
 
To give you an idea Linux boot with the m5exit.squashfs.arm64 disk takes 2mins and 30 seconds prior to this PR and 2mins 50 with the current PR. From flamegraph is clear the lookup logic is responsible for the regression (especially findEntry).

But more specifically I found the following motivations (there are probably more)

1) We are anyway transforming the L2 shared from being fully associative to 5-way set associative. This means we are having more conflicts in the l2shared with a partial entry potentially kicking out a valid complete translation. So we are walking more (and walking increases both guest and host time)
2) Current logic is unoptimized for fully associative TLBs (ITLB and DTLB). The indexing policy returns a set of candidates under the form of std::vector<TlbEntry*>. We are therefore copying a vector (e.g. with 64 entries) in the critical path (instruction fetch). We might need to reconsider the IP APIs to make it more FA friendly @odanrc 
3) To adhere with older design, we are copying a new TlbEntry from the table walker into the TLB while inserting a new entry. We should either use move semantics or eliminate the copy

Overall I think it is normal to introduce a small regression considered the increased complexity but I believe it shouldn't be as big as it is now...

PR #1662, opened 2024-10-11
systemc: Disable 'overloaded-virtual' warn for clang

### LINKED ISSUE #1658: Clang fails to compile latest develop branch
I am seeing the following build errors with clang:

```
In file included from src/mem/ruby/network/garnet/VirtualChannel.cc:32:
In file included from src/mem/ruby/network/garnet/VirtualChannel.hh:38:
In file included from src/mem/ruby/network/garnet/flitBuffer.hh:39:
In file included from src/mem/ruby/network/garnet/flit.hh:39:
src/mem/ruby/slicc_interface/Message.hh:140:23: error: private field 'p_ruby_system' is not used [-Werror,-Wunused-private-field]
    const RubySystem *p_ruby_system = nullptr;
                      ^
1 error generated.
 [     CXX] src/mem/cache/replacement_policies/random_rp.cc -> ARM/mem/cache/replacement_policies/random_rp.do
 [     CXX] src/arch/arm/htm.cc -> ARM/arch/arm/htm.do
 [     CXX] src/mem/cache/replacement_policies/mru_rp.cc -> ARM/mem/cache/replacement_policies/mru_rp.do
 [     CXX] src/mem/tport.cc -> ARM/mem/tport.do
scons: *** [build/ARM/mem/ruby/network/garnet/VirtualChannel.do] Error 1
src/mem/ruby/common/DataBlock.cc:63:18: error: assigning field to itself [-Werror,-Wself-assign-field]
    m_block_size = m_block_size;
                 ^
1 error generated.

```
There also seem to be some overloaded virtual function being hidden errors.

pranith:
We need to extend the warning disable even for clang compiler.

Fixes #1658

### DIFF
--- a/src/systemc/ext/core/sc_export.hh
+++ b/src/systemc/ext/core/sc_export.hh
@@ -78,7 +78,7 @@ class sc_export : public sc_export_base
  * code is correct).
  * Please check section 9.3 of SystemC 2.3.1 release note for more details.
  */
-#if defined(__GNUC__) && (__GNUC__ >= 13)
+#if defined(__clang__) || (defined(__GNUC__) && (__GNUC__ >= 13))
 #pragma GCC diagnostic ignored "-Woverloaded-virtual"
 #endif
     void operator () (IF &i) { bind(i); }
--- a/src/systemc/ext/core/sc_port.hh
+++ b/src/systemc/ext/core/sc_port.hh
@@ -126,7 +126,7 @@ class sc_port_b : public sc_port_base
  * code is correct).
  * Please check section 9.3 of SystemC 2.3.1 release note for more details.
  */
-#if defined(__GNUC__) && (__GNUC__ >= 13)
+#if defined(__clang__) || (defined(__GNUC__) && (__GNUC__ >= 13))
 #pragma GCC diagnostic ignored "-Woverloaded-virtual"
 #endif
     void operator () (IF &i) { bind(i); }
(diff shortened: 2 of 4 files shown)


src/systemc/ext/core/sc_export.hh
```diff
@@ -78,7 +78,7 @@ class sc_export : public sc_export_base
  * code is correct).
  * Please check section 9.3 of SystemC 2.3.1 release note for more details.
  */
-#if defined(__GNUC__) && (__GNUC__ >= 13)
+#if defined(__clang__) || (defined(__GNUC__) && (__GNUC__ >= 13))
```
powerjg, line 81: Just to double check, this `-Woverloaded-virtual` is present in all versions of clang? IIRC, we check `__GNUC__ >= 13` because in versions less than 13 this fails due to the warning not existing.

src/systemc/ext/core/sc_port.hh
```diff
@@ -126,7 +126,7 @@ class sc_port_b : public sc_port_base
  * code is correct).
  * Please check section 9.3 of SystemC 2.3.1 release note for more details.
  */
-#if defined(__GNUC__) && (__GNUC__ >= 13)
+#if defined(__clang__) || (defined(__GNUC__) && (__GNUC__ >= 13))
 #pragma GCC diagnostic ignored "-Woverloaded-virtual"
```
BobbyRBruce, line 130: Does clang understand this pragma? I had assumed `GCC diagnostic ignored` was only understood by GCC?

src/systemc/ext/core/sc_export.hh
pranith: It is present as far back as clang-6, so I think so.

src/systemc/ext/core/sc_port.hh
pranith: Yes, clang tries to stay compatible with GCC pragrams, so it implements most of them.

BobbyRBruce (approved):

PR #1666, opened 2024-10-11
scons,misc: Portable debug flag generation

heshpdx:
Modifies union construction in the debug directory so output is more amenable to alternative compilers. Verified that this change produces code that builds with clang, gcc, msvc, nvhpc, aocc, icc, openxl, and cray hpc.

These were the kinds of errors seen in MSVC, which this patch fixes.
```
debug/Decoder.hh(24): error C2461: 'gem5::debug::unions::Decoder': constructor syntax missing formal parameters
debug/Decoder.hh(31): error C7624: Type name 'gem5::debug::unions::Decoder' cannot appear on the right side of a class member access expression
```

### DIFF
--- a/build_tools/debugflaghh.py
+++ b/build_tools/debugflaghh.py
@@ -100,13 +100,16 @@ if components:
 inline union ${{args.name}}
 {
     ~${{args.name}}() {}
-    CompoundFlag ${{args.name}} = {
-        "${{args.name}}", "${{args.desc}}", {
+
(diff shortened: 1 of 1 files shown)


erin-le (approved):
LGTM

I tested by building gem5 using gcc 13, gcc 10, clang 18, and clang 14, then ran simulations using `arm-hello.py` with `--debug-flags=Arm` (a base flag) and `--debug-flags=MemoryAccess` (a compound flag).

PR #1679, opened 2024-10-17
sim: Make SignalSinkPort::set virtual

handsomeliu-google:
We are implementing derived classes of SignalSinkPort that does some additional logic after it's triggered (set() invoked by SignalSourcePort peer), and before executing the callback that a device provides (in onChange_). The logic is like additional logging, or providing debugging features. However, set() itself directly calls the onChange_ callback.

Making the set() virtual could provide the flexibility to achieve this feature.

Change-Id: Iaa3fd6bd76f0b60c17b651561a7567517a364360

### DIFF
--- a/src/sim/signal.hh
+++ b/src/sim/signal.hh
@@ -51,12 +51,11 @@ class SignalSinkPort : public Port
     SignalSourcePort<State> *_source = nullptr;
 
     State _state = {};
-    OnChangeFunc _onChange;
 
   protected:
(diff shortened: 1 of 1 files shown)


powerjg (dismissed):

handsomeliu-google:
I realized that the on change callback needs to be accessible to derived class as well to achieve this.

BobbyRBruce (approved):

PR #1681, opened 2024-10-17
cpu-o3: Clear thread state in time buffers on thread exit

### LINKED ISSUE #1049: cpu-o3: executing MFENCE in two parallel SMT threads causes hang

nmosier:
Fix #1049. Clear only thread-specific state from the O3 CPU time buffers, rather than clearing *all* state for *all* threads.

Specifically, this patch adds a `clearStates()` method for all O3 time buffer structs in src/cpu/o3/comm.hh. Upon thread exit, `CPU::removeThread()` now invokes this method for all structs in each time buffer, rather than flushing out the time buffers (which nukes the states for all threads, not just the exiting one).

### DIFF
--- a/src/cpu/o3/comm.hh
+++ b/src/cpu/o3/comm.hh
@@ -225,6 +225,25 @@ struct TimeStruct
     bool iewUnblock[MaxThreads];
 };
 
+/**
+ * Remove instructions belonging to given thread from the
+ * given comm struct's instruction array. Automatically
+ * updates the array size.
+ */
+template <class CommStruct>
+void
+removeCommThreadInsts(ThreadID tid, CommStruct& comm_struct)
+{
+    auto has_tid = [tid] (const auto &inst) -> bool {
+        return inst && inst->threadNumber == tid;
+    };
+    DynInstPtr *last = std::remove_if(comm_struct.insts,
+                                      comm_struct.insts + comm_struct.size,
+                                      has_tid);
+    std::fill(last, comm_struct.insts + comm_struct.size, nullptr);
+    comm_struct.size = last - comm_struct.insts;
+}
+
 } // namespace o3
 } // namespace gem5
 
--- a/src/cpu/o3/comm.hh
+++ b/src/cpu/o3/comm.hh
@@ -116,14 +116,14 @@ struct TimeStruct
         std::unique_ptr<PCStateBase> nextPC;
         DynInstPtr mispredictInst;
         DynInstPtr squashInst;
-        InstSeqNum doneSeqNum;
-        Addr mispredPC;
-        uint64_t branchAddr;
-        unsigned branchCount;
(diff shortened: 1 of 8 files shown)


powerjg (commented):
This is a great solution. Thanks for fixing this issue!

I would prefer to see the functions in `comm.hh` to be in their respective places with the other member functions of each class. I do not believe we have any examples of this kind of style in gem5, and it will make future maintenance easier if we don't add new styles. Let me know if I'm wrong and we do this in other places, though.

Also, it would be good to have the change which modifies all of the initial values in cpu.hh to be in a different commit (no need to make a different PR, though).

erin-le:
@nmosier Hello, thank you for your contribution. Could you address Jason's comments when you have time? Thank you!

nmosier:
@powerjg Thanks for the feedback! I tried to address it.

powerjg (approved):

powerjg:
Apologies for the slow review!

PR #1684, opened 2024-10-18
dev: move dprint of reg name before register read/write

harry900831:
Originally, the debug print for read/write to specific register name will happen after reg.read() and reg.write(). However, there might be other debug print or warning inside reg.read(), reg.write() which would be confusing if this debug log happen after all other debug print inside reg.read(), reg.write().

Creating this commit to change the order.


giactra (approved):

giactra:
Giving some time to @hungweihsuG to review this; I will merge mid-week if there's no stopper

hungweihsuG (approved):
The change looks good to me. Thanks!

PR #1651, opened 2024-10-10
mem-ruby,tests: Add CHI with ISA tests

BobbyRBruce:
These tests ensure CHI is sitll working with ARM, X86, and RISCV ISA targets. Additionally whether it still functions in atomic mode and with varying number of cores.

PR #1689, opened 2024-10-18
mem-ruby: Fix errors in CHI protocol due to recent update in slicc

hnpl:
Discussion: https://github.com/gem5/gem5/pull/1651


hnpl:
See #1690.

PR #1690, opened 2024-10-18
mem-ruby: Fix issues in protocols due to multi-RubySystem

abmerop:
Starting with https://github.com/gem5/gem5/pull/1453 , some Ruby structures require a block size be set
and other require a pointer to the Ruby system. This fixes some cases
which were not covered by the per-checkin tests but seen in daily+
tests. In particular:

 - WriteMasks and PerfectCacheMemory must explicitly set a block size.
 - NetDest and RubyProxyPort require RubySystem pointer.
 - Classes inheriting Message now have a setRubySystem collecting all
   objects that need a RubySystem pointer and this should be called in
   the constructor of the Message.

This commit makes sure all of these happen. This should fix daily
arm_boot_tests and daily learning_gem5 tests.

### DIFF
--- a/configs/learning_gem5/part3/msi_caches.py
+++ b/configs/learning_gem5/part3/msi_caches.py
@@ -103,7 +103,7 @@ class MyCacheSystem(RubySystem):
 
         # Set up a proxy port for the system_port. Used for load binaries and
         # other functional-only things.
-        self.sys_port_proxy = RubyPortProxy()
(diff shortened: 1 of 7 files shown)


hnpl (approved):

BobbyRBruce (approved):

BobbyRBruce:
Thanks @abmerop !

powerjg:
FYI, @giactra

giactra:
Thanks @powerjg for the tag

PR #1692, opened 2024-10-20
dev-amdgpu, gpu-compute, mem-ruby: Add support for writeback L2 in GPU

v-ramadas:
Previously, GPU L2 caches could be configured in either writeback or writethrough mode when used in an APU. However, in a CPU+dGPU system, only writethrough worked. This is mainly because in CPU+dGPU system, the CPU sends either PCI or SDMA requests to transfer data from the GPU memory to CPU. When L2 cache is configured to be writeback, the dirty data resides in L2 when CPU transfers data from GPU memory. This leads to the wrong version being transferred. A similar issue also crops up when the GPU command processor reads kernel information before kernel dispatch, only to incorrect data. This PR contains a set of commits that fix both these issues.

### DIFF
--- a/src/gpu-compute/compute_unit.cc
+++ b/src/gpu-compute/compute_unit.cc
@@ -1062,7 +1062,18 @@ ComputeUnit::SQCPort::recvTimingResp(PacketPtr pkt)
      * and doesn't have a wavefront or instruction associated with it.
      */
     if (sender_state->wavefront != nullptr) {
-        computeUnit->handleSQCReturn(pkt);
+        RequestPtr req = pkt->req;
+        // If the sender state's isKernDispath is set, then the request came
+        // from the gpu command processor. The request fetches information
+        // that will be used in the kernel dispatch process. It should be
+        // handled in the gpu command processor. If the flag isn't set,
+        // then the request is an instruction fetch and can be handled in
+        // the compute unit
+        if (sender_state->isKernDispatch) {
+          computeUnit->shader->gpuCmdProc.completeTimingRead();
+        } else {
+          computeUnit->handleSQCReturn(pkt);
+        }
     } else {
         delete pkt->senderState;
         delete pkt;
--- a/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
@@ -832,17 +832,14 @@ machine(MachineType:TCC, "TCC Cache")
       out_msg.Type := CoherenceResponseType:CPUPrbResp;  // TCC, L3  respond in same way to probes
       out_msg.Sender := machineID;
       out_msg.Destination.add(mapAddressToMachine(address, MachineType:Directory));
-      if (getState(tbe, cache_entry, address) == State:V || getState(tbe, cache_entry, address) == State:I) {
-        out_msg.Dirty := false;
-      } else {
-        out_msg.Dirty := true;
-      }
       if (getState(tbe, cache_entry, address) == State:V || getState(tbe, cache_entry, address) == State:M || getState(tbe, cache_entry, address) == State:W) {
         out_msg.Hit := true;
+        out_msg.Dirty := true;
         out_msg.DataBlk := cache_entry.DataBlk;
         out_msg.MessageSize := MessageSizeType:Response_Data;
       } else {
 	    out_msg.Hit := false;
+        out_msg.Dirty := false;
         out_msg.MessageSize := MessageSizeType:Response_Control;
       }
       out_msg.Ntsl := true;
--- a/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
@@ -73,6 +73,7 @@ machine(MachineType:TCC, "TCC Cache")
     // Probes
     PrbInv,                 desc="Invalidating probe";
     InvCache,               desc="Invalidating probe from TCP";
+    PrbDowngrade,	        desc="Downgrading probe";
     // Coming from Memory Controller
     WBAck,                  desc="writethrough ack from memory";
     Bypass,                 desc="Bypass the entire L2 cache";
@@ -348,7 +349,14 @@ machine(MachineType:TCC, "TCC Cache")
         DPRINTF(RubySlicc, "%s\n", in_msg);
         Entry cache_entry := getCacheEntry(in_msg.addr);
         TBE tbe := TBEs.lookup(in_msg.addr);
-        trigger(Event:PrbInv, in_msg.addr, cache_entry, tbe);
+        if (in_msg.Type == ProbeRequestType:PrbInv) {
+          // Invalidate data and send it downstream
+          trigger(Event:PrbInv, in_msg.addr, cache_entry, tbe);
+        } else {
+          // If data present in cache, then downgrade it and send it
+          // downstream
+          trigger(Event:PrbDowngrade, in_msg.addr, cache_entry, tbe);
+        }
       }
     }
   }
@@ -818,6 +826,31 @@ machine(MachineType:TCC, "TCC Cache")
     }
   }
 
+  action(pd_sendProbeResponseDowngrade, "pd", desc="send probe downgrade") {
+    enqueue(responseToNB_out, ResponseMsg, 1) {
+      out_msg.addr := address;
+      out_msg.Type := CoherenceResponseType:CPUPrbResp;  // TCC, L3  respond in same way to probes
+      out_msg.Sender := machineID;
+      out_msg.Destination.add(mapAddressToMachine(address, MachineType:Directory));
+      if (getState(tbe, cache_entry, address) == State:V || getState(tbe, cache_entry, address) == State:I) {
+        out_msg.Dirty := false;
+      } else {
+        out_msg.Dirty := true;
+      }
+      if (getState(tbe, cache_entry, address) == State:V || getState(tbe, cache_entry, address) == State:M || getState(tbe, cache_entry, address) == State:W) {
+        out_msg.Hit := true;
+        out_msg.DataBlk := cache_entry.DataBlk;
+        out_msg.MessageSize := MessageSizeType:Response_Data;
+      } else {
+	    out_msg.Hit := false;
+        out_msg.MessageSize := MessageSizeType:Response_Control;
+      }
+      out_msg.Ntsl := true;
+      out_msg.State := CoherenceState:NA;
+    }
+  }
+
+
   action(pi_sendProbeResponseInv, "pi", desc="send probe ack inv, no data") {
     enqueue(responseToNB_out, ResponseMsg, 1) {
       out_msg.addr := address;
@@ -1213,6 +1246,25 @@ machine(MachineType:TCC, "TCC Cache")
     i_invL2;
     ir_invL2Resp;
     p_popRequestQueue;
+  }
+
+  transition(I, PrbDowngrade) {TagArrayRead} {
+    pd_sendProbeResponseDowngrade;
+    pp_popProbeQueue;
+  }
+
+  transition(V, PrbDowngrade) {TagArrayRead} {
+    pd_sendProbeResponseDowngrade;
+    pp_popProbeQueue;
+  }
+
+  transition({M, W}, PrbDowngrade, V) {TagArrayRead, TagArrayWrite} {
+    pd_sendProbeResponseDowngrade;
+    pp_popProbeQueue;
+  }
+
+  transition({A, IV, WI, WIB}, PrbDowngrade) {TagArrayRead, TagArrayWrite} {
+    st_stallAndWaitRequest;
   }
 
   transition({I, V}, PrbInv, I) {TagArrayRead, TagArrayWrite} {
--- a/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
@@ -180,8 +180,11 @@ machine(MachineType:TCC, "TCC Cache")
 
   void functionalRead(Addr addr, Packet *pkt) {
     TBE tbe := TBEs.lookup(addr);
+    Entry cache_entry := getCacheEntry(addr);
     if(is_valid(tbe)) {
       testAndRead(addr, tbe.DataBlk, pkt);
+    } else if (is_valid(cache_entry)) {
+      testAndRead(addr, cache_entry.DataBlk, pkt);
     } else {
       functionalMemoryRead(pkt);
     }
--- a/src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
+++ b/src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
@@ -107,7 +107,7 @@ machine(MachineType:Directory, "AMD Baseline protocol")
 
     // probe responses
     CPUPrbResp,            desc="Probe Response Msg";
-    CPUPrbRespHit,	       desc="Probe Response Msg and Data";
+    CPUPrbRespWB,	       desc="Probe Response Msg and Data";
 
     ProbeAcksComplete,  desc="Probe Acks Complete";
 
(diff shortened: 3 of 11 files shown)


src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -815,6 +827,28 @@ machine(MachineType:TCC, "TCC Cache")
     }
   }
 
+  action(pd_sendProbeResponseDowngrade, "pd", desc="send probe downgrade") {
+    enqueue(responseToNB_out, ResponseMsg, 1) {
+      out_msg.addr := address;
+      out_msg.Type := CoherenceResponseType:CPUPrbResp;  // TCC, L3  respond in same way to probes
+      out_msg.Sender := machineID;
+      out_msg.Destination.add(mapAddressToMachine(address, MachineType:Directory));
+      if (getState(tbe, cache_entry, address) == State:V || getState(tbe, cache_entry, address) == State:M || getState(tbe, cache_entry, address) == State:W) {
```
mattsinc, line 835: this is not my favorite format, but I guess in this case checking isDirty or something like that is not sufficient?  If so, then this is fine.

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
```diff
@@ -300,7 +304,11 @@ machine(MachineType:Directory, "AMD Baseline protocol")
         TBE tbe := TBEs.lookup(in_msg.LineAddress);
         CacheEntry entry := static_cast(CacheEntry, "pointer", L3CacheMemory.lookup(in_msg.LineAddress));
         if (in_msg.Type == DMARequestType:READ) {
-          trigger(Event:DmaRead, in_msg.LineAddress, entry, tbe);
+          if (L2isWB) {
+            trigger(Event:DmaReadWB, in_msg.LineAddress, entry, tbe);
+          } else {
+            trigger(Event:DmaRead, in_msg.LineAddress, entry, tbe);
+	      }
```
mattsinc: tabs/spacing off on this line?

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
```diff
@@ -825,6 +837,14 @@ machine(MachineType:Directory, "AMD Baseline protocol")
                                             TCC_select_low_bit,
                                             TCC_select_num_bits));
         }
+
+        if (GPUonly && L2isWB) {
+          probe_dests.add(mapAddressToRange(address, MachineType:TCC,
+                                            TCC_select_low_bit,
+                                            TCC_select_num_bits));
+        }
+
```
mattsinc: remove excess lines

mattsinc (dismissed):
A few minor changes but from my end this mostly looks like what we discussed.

@abmerop should look at this before we merge though

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
v-ramadas: Done

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
v-ramadas: I/Nagendra added an additional if() condition to check if L2 is WB. The additional spaces are from that, I think.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
v-ramadas: The TBE/cache entry's dirty flag doesn't seem to be set anywhere. I would have to set that to true/false in transitions before I can use it here. Would you prefer I do that, or is this sufficient?

src/gpu-compute/compute_unit.cc
```diff
@@ -1062,7 +1062,19 @@ ComputeUnit::SQCPort::recvTimingResp(PacketPtr pkt)
...
-        computeUnit->handleSQCReturn(pkt);
+        RequestPtr req = pkt->req;
+        Flags flags = req->getFlags();
+        // If the PHYSICAL flag is set, then the request came from the
+        // gpu command processor. The request fetches information that
+        // will be used in the kernel dispatch process. It should be
+        // handled in the gpu command processor. If the flag isn't set,
+        // then the request is an instruction fetch and can be handled in
+        // the compute unit
+        if (flags == Request::PHYSICAL) {
```
abmerop: This is fine, however it might be cleaner to add something to the SQCPort::SenderState instead indicating it came from the command processor. I am not sure the origins of this flag, but my assumption has been it indicates that address translation should be bypassed. The current way of reading the kernel code structure uses this flag, so I can see why you did this.

src/mem/ruby/system/RubySystem.cc
```diff
@@ -560,7 +560,11 @@ RubySystem::functionalRead(PacketPtr pkt)
     // it only if it's not in the cache hierarchy at all.
     int num_controllers = netCntrls[request_net_id].size();
     if (num_invalid == (num_controllers - 1) && num_backing_store == 1) {
-        DPRINTF(RubySystem, "only copy in Backing_Store memory, read from it\n");
+        /*if (pkt->getGPUFuncAccess()) {
+            return false;
+        }*/
+        DPRINTF(RubySystem,
+                "only copy in Backing_Store memory, read from it\n");
```
abmerop: This can be dropped, it looks like no there are changes to this file.

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
```diff
@@ -1567,13 +1607,75 @@ machine(MachineType:Directory, "AMD Baseline protocol")
...
+  }
+
+  transition(BL2, MemData, BL2_Pm) {
+    mt_writeMemDataToTBE;
+    pm_popMemQueue;
+  }
+
+  transition({BL2_Pm, U}, MemData) {
+    pm_popMemQueue;
+  }
```
abmerop, line 1667: This seems suspect. Is there some case where duplicate memory requests get sent and we ignore their responses?

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -345,7 +349,15 @@ machine(MachineType:TCC, "TCC Cache")
...
-        trigger(Event:PrbInv, in_msg.addr, cache_entry, tbe);
+        if (in_msg.Type == ProbeRequestType:PrbInv) {
+          // Cache is write-through. Invalidate data and send it
+          // downstream
+          trigger(Event:PrbInv, in_msg.addr, cache_entry, tbe);
+        } else {
+          // Cache is writeback. If data present in cache, then downgrade
+          // it and send it downstream
+          trigger(Event:PrbDowngrade, in_msg.addr, cache_entry, tbe);
+        }
```
abmerop: Could you explain this part? Do PrbInv requests only happen in write-through cache? I see elsewhere that `L2isWB` is used instead to check this, so does that imply PrbInv can happen in write-back cache too and the comment needs to be updated?

src/mem/packet.hh
```diff
@@ -419,6 +419,8 @@ class Packet : public Printable, public Extensible<Packet>
      */
     uint64_t htmTransactionUid;
 
+    bool _isGPUFuncAccess;
+
```
abmerop: Historically these have gone in the request class.  Does it need to be part of Packet class (i.e., can you have a Packet with a nullptr Request?)

abmerop (commented):
I really like adding the timing read for the kernel code object!  I believe this is the last place the GPU model is using a functional access in a place where it is not intentional (e.g., a performance optimization).

I think I mostly follow the protocol changes but not completely so I could use some clarification there.

src/mem/packet.hh
powerjg: To add on here, I would *strongly* prefer not changing Packet *or* Request unless absolutely necessary. There should be a very high bar for changing these widely used objects. Why do we need a special functional access flag for the GPU?

abmerop: I can try to answer based on the comment in RubyPort.cc... This is related to the multiple Networks in one RubySystem which uses the requestor ID to know if requests should go to the GPU or CPU network. It defaults to 0 which is an unmatched network = CPU side. To know the network at the RubyPort, some misc. object accessible by the RubyPort must know which network to pick. Packet/Request have been the best interface so far.

This is mainly because of the way *I* implemented that multiple network support. The multiple-RubySystem supersedes this but would require moving to that...  In the long term everything related to the multiple Networks will need a refactor, I think.  I am aware of some folks using it, but I don't know if they can use multiple-RubySystem. Right now this is the common way to deal with multiple networks in Ruby.

powerjg: I suppose my question is whether this is duplicate information. E.g., can we infer that it's a GPU functional access from (1) it's functional, and (2) it is in the 1 network?

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
v-ramadas: I've updated the comments to remove the write-through/writeback part. It is more generic now. Thanks for pointing this out!

src/mem/ruby/system/RubySystem.cc
v-ramadas: Removed the commented out section. I'm unable to revert the change to the DPRINTF. The gem5 style checker complains about the old DPRINTF() having an invalid line length and wants me to put the line break in. Is there a way to get past this and revert it?

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
v-ramadas: The protocol changes are intended to work as follows when the cache line receives a DMAReadWB event trigger (only triggered if L2 is WB):

- The cache line moves from U to BL2. It checks L3 cache (if size is non-zero), probes L2 caches, and sends out a request to memory simultaneously.
- If it is a hit at L3 (line 1625) or if L2 sends back data (line 1630), the line transitions to BL2_Pm (got memory but waiting for probes to complete). Any data from memory is now ignored (line 1666)
- If it is not a hit at L3 and if L2 sends probe ack without data (line 1610), the cache waits for data from memory. If data from memory arrives before ProbeAcksComplete, the transition at line 1660 is called first and the transition at line 1650 is called when all probes are received
- If it is not a hit at L3 and if none of the L2s send data but send ProbeAcksComplete before memory request returns, the transition at line 1643 is called and the line moves to BL2_M state (waiting for mem). Then, when data from memory arrives, the transition at line 1669 is called.

There are four possible states for the cacheline when mem data returns: if it is in BL2 or BL2_M, the data is written to TBE and any dirty data in L2/L3 overwrites the TBE later. If it already received dirty data (in BL2_Pm) the mem data is ignored. If the line is in U (already received dirty data and all acks) the mem data is ignored.

I am unable to think of any other case. Please let me know if you can think of any, and I will add that too!

src/mem/ruby/system/RubySystem.cc
abmerop: There *is* however I think it's fine to keep this.  Someday it will be caught by clang-format anyway :).

src/gpu-compute/compute_unit.cc
v-ramadas: That sounds like a good idea! I'm currently working on it and will update the commit soon

v-ramadas: I've pushed this change. I added a isKernDispatch field to SQCPort::SenderState that tracks whether the request is a kernel dispatch or not

src/mem/packet.hh
v-ramadas: The issue I faced when making this fix was that the GPU functional accesses go through network 0 (CPU network) by default. I use this flag to make packets use one of the GPU requestor IDs so that it is injected into network 1 when it reaches RubyPort instead of network 0. I couldn't find any other structure that is shared between Ruby and the GPU compute units that would help me with this.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
mattsinc: let it be then

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
mattsinc: I'm not sure how that can be for line 311?  But if it's an optical illusion, ok

src/mem/ruby/system/RubySystem.cc
mattsinc: Going to mark this resolved

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
v-ramadas: Surprisingly, I still see the right indentation in my github, vim views. But when I open the file in vim, I see that it uses a tab instead of spaces. Maybe a different tab/space configurations cause it to be formatted differently in different environments? I have now changed the tab to space. Hope that fixes it!

src/mem/packet.hh
v-ramadas: > Historically these have gone in the request class. Does it need to be part of Packet class (i.e., can you have a Packet with a nullptr Request?)

Moved it to the request class.

mattsinc: My sense is that @v-ramadas is saying the answer to your question is no, @powerjg .  Or at least absent the complete working @abmerop is proposing, the answer is no.  So unless you all have an alternative, I think this is the best option we could think of.  Let us know ...

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
mattsinc: @abmerop just checking if this addresses your concern?

mattsinc (approved):
Approving so testing can happen in parallel with @abmerop and @powerjg re-reviewing

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
abmerop: Thanks for the explanation!  It should then be impossible to get MemData in U state, correct?  This is the transition that concerns me.  MemData in BL2_Pm state makes sense.  If that is true, then the change would be `{BL2_Pm, U}` -> `BL2_Pm`.

The concern is it might be hiding some other bugs (potentially entirely unrelated to this PR) which go to U state when it should be waiting for memory.

powerjg (approved):
I still have reservations about modifying the request for this one small use case. My gut says there's a better way to do this. However, I don't understand the problem well enough to suggest anything specific; so, this change is OK with me.

abmerop:
> I still have reservations about modifying the request for this one small use case. My gut says there's a better way to do this. However, I don't understand the problem well enough to suggest anything specific; so, this change is OK with me.

If it makes it better, this won't be needed anymore after the VIPER board.  That is using multiple RubySystems instead of multiple RubyNetworks (aka disjoint network).  In the multiple RubySystems case, there is no ambiguity which network functional reads go to.

powerjg:
So should we not merge this?

abmerop:
We should merge this. The old configs are not yet deprecated and still using multiple RubyNetworks.

src/mem/ruby/protocol/MOESI_AMD_Base-dir.sm
v-ramadas: It would be possible to get MemData in U state if the memory request returns after the cache probes complete. A slow memory request, in this case, would then come to a cache state in U. I now wonder if this would introduce a scenario where a new request comes in before MemData returns, and lead to an inconsistency. I will update the protocol to handle this scenario too.

abmerop: Sorry to belabor this, but I am still missing something.  In the case you describe wouldn't be in `BL2_M`? I.e., the state which is "got probes, waiting for memory"?

v-ramadas: The line would be in BL2_M only if none of the probes return dirty data. If a probe did, then the line moves to BL2_Pm.

There are two paths the sequence of transitions could take. When the state is in BL2, it is waiting for both probes from caches and response from memory. If the cache probes return dirty data, the line transitions to BL2_Pm. From here, either all the cache probes could return first (BL2_Pm->U) or memory could return first (remains in BL2_PM and moves to U once other probes return). If none of the cache probes returns dirty data, the line could move to BL2_M once all probes return but MemData hasn't or move to BL2_Pm if MemData comes first (BL2_Pm->U once probes return).

I will add more comments, change the state descriptions, and use a different name for the CPUPrbRespWB event (change it to CPUPrbRespDirty)

PR #1693, opened 2024-10-20
configs,scons: Update scripts and build_opts to make GPU-FS simulations more configurable

v-ramadas:
This PR adds support for command line arguments in GPU-FS runs to allow the user to configure several parts of the GPU. It also increases the bits per set in the build_opts/VEGA_X86 file to enable GPU-FS simulations to use 64 directories or more.


configs/example/gpufs/amd/AmdGPUOptions.py
```diff
@@ -247,16 +247,94 @@ def addAmdGPUOptions(parser):
...
         help="register allocation policy (simple/dynamic)",
     )
+
+    parser.add_argument(
+        "--SegFaultDebug",
+        action="store_true",
+        help="checks for GPU seg fault before TLB access",
+    )
+
+    parser.add_argument(
```
mattsinc: Why do we need to include this?

mattsinc (commented):
Just the one question.

No idea how this will integrate with #1636 though...

build_opts/VEGA_X86
```diff
@@ -1,4 +1,5 @@
 RUBY=y
+NUMBER_BITS_PER_SET=128
```
abmerop, line 2: Unrelated to this PR, but there really ought to be a way to configure the bitset at runtime so this isn't necessary.

I don't really like needing to delete my build directory and build from scratch so that scons propogates this variable.

configs/example/gpufs/amd/AmdGPUOptions.py
```diff
@@ -247,16 +247,88 @@ def addAmdGPUOptions(parser):
         default="simple",
         help="register allocation policy (simple/dynamic)",
     )
+
+    parser.add_argument(
+        "--FunctionalTLB",
+        action="store_true",
+        help="Assumes TLB has no latency",
+    )
```
abmerop: Has functionalTLB been implemented in GPUFS??  I wasn't aware anyone was implementing this, but it would be another method to put the GPU model in a "fast-forwarding" mode.

configs/example/gpufs/amd/AmdGPUOptions.py
```diff
@@ -247,16 +247,88 @@ def addAmdGPUOptions(parser):
...
+    parser.add_argument(
+        "--dgpu",
+        action="store_true",
+        default=False,
+        help="Configure the system as a dGPU instead of an APU."
+        "The dGPU config has its own local memory pool and is not "
+        "coherent with the host through hardware.  Data is "
+        "transfered from host to device memory using runtime calls "
+        "that copy data over a PCIe-like IO bus.",
+    )
```
abmerop: Is this not implied? Are you working on APU support for GPUFS?

configs/example/gpufs/amd/AmdGPUOptions.py
```diff
@@ -247,16 +247,82 @@ def addAmdGPUOptions(parser):
...
+
+    parser.add_argument(
+        "--gfx-version",
+        type=str,
+        default="gfx902",
+        choices=["gfx900", "gfx902", "gfx908", "gfx90a"],
```
abmerop: Missing `gfx942` (MI300).  Also unrelated to this patch, but the AMDGPUDevice uses a "device_name" argument which has the string value for these versions.  We should pick one or the other.  The `gfx*` choices are good in the sense that multiple devices can have the same gfx version (e.g., MI300X and MI300A are both gfx942).  On the other hand, more people would recognize MI300X vs. gfx942.

configs/example/gpufs/amd/AmdGPUOptions.py
v-ramadas: Removed the seg fault debug flag.

configs/example/gpufs/amd/AmdGPUOptions.py
abmerop: See also: https://github.com/gem5/gem5/blob/develop/configs/example/gpufs/runfs.py#L127-L136 and https://github.com/gem5/gem5/blob/develop/configs/example/gpufs/mi300.py#L158

configs/example/gpufs/amd/AmdGPUOptions.py
v-ramadas: I removed this now. I am pushing a commit by someone else and this skipped my eye. Sorry about that!

configs/example/gpufs/amd/AmdGPUOptions.py
v-ramadas: Removed this now. Same as above, I missed catching this when I reviewed the commit internally

configs/example/gpufs/amd/AmdGPUOptions.py
v-ramadas: Removed now

build_opts/VEGA_X86
v-ramadas: That is a good suggestion! I will look into this

powerjg: This would be a great improvement. I do worry that it's something like `typedef bitset uint8_t[NUMBER_OF_BITS_PER_SET/8];` or something like that which will be difficult to change dynamically.

abmerop: There are dynamic_bitset kind of things in external libs (e.g., boost). But that seems like a lot of overhead for one thing.  I wonder if we can get away with `vector<bool>`..

powerjg: Just FYI, this is defined in `src/mem/ruby/common/Set.hh`. It uses a bitset which requires the size to be known at compile time. https://en.cppreference.com/w/cpp/utility/bitset

We could *possibly* use `vector<bool>` (https://en.cppreference.com/w/cpp/container/vector_bool), but I don't know how we would know dynamically what the max entries per set will be (probably the number of machines used for `NetDest`, but I'm not certain).

mattsinc:
@v-ramadas is this ready for review again?  Or do you need to resolve the bitset issue first (not sure if that is a blocker @abmerop?)?

abmerop:
> @v-ramadas is this ready for review again? Or do you need to resolve the bitset issue first (not sure if that is a blocker @abmerop?)?

Definitely *not* a blocker. That should be a separate PR in fact

build_opts/VEGA_X86
abmerop: What I did (temporarily) is just move this to a #define in Set.hh.  Not the best solution but as you can tell I don't like recompiling from scratch :)

v-ramadas:
Would it be better if I remove the NUM_BITS_PER_SET=128 change from this PR and make it separate (after looking into the suggestions made)?

mattsinc:
> Would it be better if I remove the NUM_BITS_PER_SET=128 change from this PR and make it separate (after looking into the suggestions made)?

No, my sense from @abmerop 's reply above is we should leave it and fix it separately / after.

mattsinc (approved):
@abmerop or @powerjg can chime in if they disagree, but it appears all requested changes for this specific patch have been made.

abmerop (approved):

PR #1695, opened 2024-10-21
tests: Fix replacement_policies tests' refs

BobbyRBruce:
At some point 'system' -> 'board' in the stdlib code the replacement policy tests used. Due to this the output is slightly different meaning the refs need updated.

This was causing the Daily Tests to fail.


odanrc (approved):

mattsinc (approved):
Thanks @BobbyRBruce!

CC @yuxiaojia for reference -- just an FYI, nothing you need to do

PR #1698, opened 2024-10-21
tests: move weekly gpu tests to have separate jobs

Harshil2107:


tests/gem5/gpu/test_gpu_pannotia.py
```diff
@@ -168,75 +169,75 @@
...
-        "-c",
-        joinpath(binary_path, "pagerank_spmv.gem5"),
-        "--options",
-        f'{joinpath(dataset_path, "coAuthorsDBLP.graph")} 0',
-    ],
-    valid_isas=(constants.vega_x86_tag,),
-    valid_hosts=(constants.host_gcn_gpu_tag,),
-    length=constants.very_long_tag,
-)
+# gem5_verify_config(
```
mattsinc, line 172: Is this failing?

mattsinc (commented):
I don't fully understand the new test format -- especially where the input args for each of these are coming from when test_gpu_pannotia seems to have everything commented out.  But if @BobbyRBruce thinks this is fine and that information is available somewhere, then this seems ok to me...

tests/gem5/gpu/test_gpu_pannotia.py
powerjg: Harshil can correct me if I'm wrong, but I think this is causing timeouts (i.e., it takes more than a few hours). Let's get a few tests working and then expand it.

BobbyRBruce (approved):
This seems fine.

As with all GitHub actions it's too hard to test locally so we'll just need to see this works once merged.

tests/gem5/gpu/test_gpu_pannotia.py
mattsinc: I could be completely wrong here, but I thought the tests are now running on different runners in parallel despite @Harshil2107 setting the timeout to 12 hours ...maybe it's still timing out despite that?

BobbyRBruce: I believe @Harshil2107 has commented these out as they take days to complete. Even with parallelization, we can't support tests that take that long.

Harshil2107: I set the timeout to 12 hours as the tests that are running ran in 12 hours locally. They were timing out before the change to each test having its own runner.

mattsinc: In the old weekly tests guise I thought I had sized everything to complete within 3 days (without parallelization).  It sounds like something has changed that made them run much longer?  I know on a separate patch @Harshil2107 and I were talking about the much larger datasets he was now using -- not sure if that is the ultimate issue?

BobbyRBruce:
I'm going to merge this because I want the Weekly tests to get back into a passing state and at least say we run some Pannotia tests as part of our test suite. I think the discussion about the tests timing out and why that might be is worth having but if a subset of tests is all we can do right now while we figure that out, that's fine. t

PR #1699, opened 2024-10-22
tests: Add tests for checking that stats reset

### LINKED ISSUE #1644: Add tests to ensure that stats reset when m5.stats.reset() is called
**Describe the bug**
Currently, we do not have tests to validate that stats reset when `m5.stats.reset()` is called. As a result, we are uncertain about which stats correctly reset and which do not.  This was recently a problem as it was discovered that `simInsts` was not resetting properly, discussed in [issue 1443](https://github.com/gem5/gem5/issues/1443) and a fix implemented in [PR 1615](https://github.com/gem5/gem5/pull/1615). 
This is also an issue for gem5 users at large as inconsistent stats resetting behavior could cause confusion in interpreting simulation results.

**Implementation**

The following implementation is from [Issue 1443, Stats Resetting](https://github.com/gem5/gem5/issues/1443).
One way that we could implement tests is to run various configurations for some number of ticks, dump stats and reset, run for one tick, then dump again.

With this setup, stats with the same value (or greater, on the off chance that something incremented in one tick) in both stats dumps are not being reset correctly. Exceptions to this are if stats have values of 0 in the first stats dump, are meant to measure stats throughout the simulation, such as finalTick, or are simulation constants, such as simFreq. 

A few considerations:
- We will need to take care to identify the stats that should be constant/always increasing throughout the simulation.
- We need to select which configurations and workloads to run that would exercise a wide range of stats
- We need to 1. identify the frequency stats, and 2. validate the frequency stats against the values they are generated from, not by comparing stats from different dumps.
- Some stats do not show up in stats.txt if their values are zero. If we wanted to ensure that these stats appear, we could run the simulation for a large number of ticks for the first dump, then reset and run a smaller number of ticks. This would also test that stats continue to work correctly after a reset, although this doesn't seem to have been a problem in the past, and we can simply assume that the stats that don't appear have values of zero.

**2/26/2025 update:**

Right now, it wouldn't be possible to write good tests for stats resetting due to the way tests run on the runners. The best approach would be to run something to generate a reference stats file before the actual stats reset tests run. However, the tests don't run in a guaranteed order on the runners, so the reference stats file might not be generated before the tests run. 

The next option would be to upload a reference stats file, but this file would need to be updated very frequently. Thus, the best solution would be to repurpose the TestLib scripts into a utility that can be run separately from the tests.

erin-le:
This PR adds tests to check that stats reset correctly. The first of these tests resets the stats before dumping them and checks that the stats have values of 0 or nan. 

The second test runs a simulation for 5 billion ticks, dumps the stats, then resets and runs the simulation for another 10 billion ticks and dumps the stats. (Most of) the values from these runs are added together and compared against a run that runs for 15 billion ticks without resetting. 

The stats whose values are not added are stats that remain constant throughout the simulation (such as simFreq) or are intended to increase without resetting (such as finalTick). Stats that are averages or rates are also not added together, but they are currently not being checked in the tests. This is because there does not appear to be a generalizable way to calculate the overall average from the averages from the 5 billion and 10 billion tick parts of the simulation.

There are a few statistics that do not appear to reset, but I am not sure what the expected behavior for these is. For now, they are not checked when the tests are run and there are comments indicating which stats these are.

In general, I excluded a number of stats that may or may not behave correctly in order to get these tests to pass.


tests/gem5/stats-reset/configs/x86-ubuntu-run-end-reset.py
```diff
@@ -0,0 +1,195 @@
...
+def read_stats_files(filepath, stats_dict):
+    with open(filepath) as stats:
+        for line in stats:
+            tmp = line.split()
+            if len(tmp) > 1:
+                stats_dict[tmp[0]] = tmp[1]
+    stats_dict.pop("----------", None)
+
+
+read_stats_files("./gem5/stats-reset/base-case-stats.txt", no_reset_dict)
```
BobbyRBruce, line 120: I think you forgot to add this to the PR?

erin-le: That stats file should be automatically generated when the first test case, which uses `x86-ubuntu-run-base.py`, is run. The first test case isn't actually a test; its sole purpose is to generate a stats file that can be used as a point of comparison for the second and third test cases. 

I set it up like this because I thought it would be more efficient than having to update the reference stats file every time a change to gem5 caused the output of the stats file to change.

BobbyRBruce: i'll give this a full review later but just to say here: you can't guarantee order of tests in testlib or declare dependencies between tests in testlib. For example, if you can just run the 2nd test here alone and it wouldn't work without running the first.

I'm guessing having this file submitted to the repo would be a bad idea as these stats may change a lot over time and break the comparison. 

Do you think what we want to do here is run two gem5 simulations from one config script. Like MultSim but just on a single thread with a guaranteed order... I wonder if such a thing is possible? We  an discuss this later but if you have time try looking into it.

BobbyRBruce: i've been thinking about this. From my understanding you want:

1. A stat file dumped at X ticks without resetting.
2. A stat file dumped at X ticks after resetting.

In which case I believe this can all be done from one simulation.

why don't you just have one exit at X ticks then in the code handling the exit you just dump the stats, move those stats somewhere safe, reset the stats and dump them again.


it wouldn't surprise me if there is some good reason you can't do this so please tell me, me if I'm wrong.

erin-le: I think that will work for the first test, but it may cause issues for the second test, which compares the results of one simulation that runs without resetting, and another that resets in the middle. Getting both sets of stats for the second test in the same simulation may cause some problems, but I can try it.

erin-le: Update: This will not work for either test. For the first test, m5.stats.dump() does not allow for multiple dumps in the same tick. For the second test, different regions of simulation (ie 0 to 15 billion ticks, then 15 to 30 billion ticks) result in different statistics, so the results can't be compared.

tests/gem5/stats-reset/configs/x86-ubuntu-run-base.py
```diff
@@ -0,0 +1,125 @@
...
+    board=board,
+    on_exit_event={
+        # Here we want override the default behavior for the first m5 exit
+        # exit event.
+        ExitEvent.EXIT: exit_event_handler(),
+        ExitEvent.MAX_TICK: max_tick_handler_no_reset(),
+    },
+)
+
+simulator.set_max_ticks(15000000000)  # 15,000,000,000
```
powerjg: FYI, You can use underscores in python to make big numbers more readable.
```suggestion
simulator.set_max_ticks(15_000_000_000)  # 15 ms
```

tests/gem5/stats-reset/configs/x86-ubuntu-run-end-reset.py
```diff
@@ -0,0 +1,176 @@
...
+    "simFreq",
+    "hostSeconds",
+    "hostMemory",
+    "UNDEFINED",
+    "clk_domain.clock",
+    "voltage_domain.voltage",
+    "peakBW",
+]
+
+# These are the stats that don't reset when m5.stats.reset() is called, and which I don't know if they are supposed to reset
```
powerjg: We should talk about these. (After submission of the patch is OK, but let's not forget.)

Also, I think these comments are >80 characters on a line

tests/gem5/stats-reset/configs/x86-ubuntu-run-base.py
```diff
@@ -0,0 +1,126 @@
+# Copyright (c) 2023 The Regents of the University of California
```
BobbyRBruce: (Here and elsewhere in this PR where it applies)

These year is wrong. It should be 2025, or if you lifted this from elsewhere and made modifications: "2023 - 2025"

tests/gem5/stats-reset/configs/x86-ubuntu-run-end-reset.py
```diff
@@ -0,0 +1,196 @@
...
+missing_keys = {}
+
+# If a stat is present when you reset but isn't there if you don't reset,
+# something is wrong
+for key, value in end_reset_dict.items():
+    if key not in no_reset_dict and not (
+        math.isnan(float(value)) or float(value) == 0.0
+    ):
+        # commented out to let tests pass
+        # not_reset_properly[key] = [value, "missing from no_reset_dict"]
```
BobbyRBruce: Should this still be here?

tests/gem5/stats-reset/configs/x86-ubuntu-run-end-reset.py
```diff
@@ -0,0 +1,196 @@
...
+for item in missing_keys.items():
+    print(item)
+
+for key, value in end_reset_dict.items():
+    # Get stats that weren't reset properly.
+    if (
+        value != "nan"
+        and float(value) != 0.0
+        and not is_match_key(key, exclude_from_check)
+        and not is_match_key(key, uncertain_exclude_from_check)
```
BobbyRBruce: If I understand this `is_match_key` thing correctly, I think it could be simplified into a one-liner with:

```
not bool(set(key.split(".")).intersection(set(exclude_from_check + uncertain_exclude_from_check)))
```

We're basically splitting the key up, then we put it in the set, we then create a set containing all the values from both the "exclude" list. We then get the intersection. When cast to a boolean an empty set is False.

tests/gem5/stats-reset/configs/x86-ubuntu-run-end-reset.py
```diff
@@ -0,0 +1,196 @@
...
+    if key not in no_reset_dict and not (
+        math.isnan(float(value)) or float(value) == 0.0
+    ):
+        # commented out to let tests pass
+        # not_reset_properly[key] = [value, "missing from no_reset_dict"]
+        missing_keys[key] = [value, "missing from no_reset_dict"]
+
+print("Missing keys:")
+for item in missing_keys.items():
+    print(item)
```
BobbyRBruce: Why are we storing this "missing from no_reset_dict" string in the `missing_keys` dictionary? Doesn't the fact a key is in the missing key output imply it's missing?

Why not just have `missing_keys[key] = value`?

The when you print I think you could just do:

```py
print("Missing keys")
for key in missing_keys:
    print(key)
```

tests/gem5/stats-reset/configs/x86-ubuntu-run-end-reset.py
```diff
@@ -0,0 +1,196 @@
...
+    ):
+        not_reset_properly[key] = value
+
+print("Incorrectly reset stats:")
+for item in not_reset_properly.items():
+    print(item)
+
+if len(not_reset_properly) > 0:
+    print(f"{len(not_reset_properly)} stats did not reset properly!")
+    sys.exit(1)
```
BobbyRBruce:
```suggestion
if not_reset_properly: 
    print(f"{len(not_reset_properly)} Incorrectly reset stats:")
    for key in not_reset_properly:
        print(key)
    sys.exit(1)
```

tests/gem5/stats-reset/configs/x86-ubuntu-run-middle-reset.py
```diff
@@ -0,0 +1,344 @@
...
+
+def read_stats_files(filepath: str, stats_dict: defaultdict[list]) -> None:
+    with open(filepath) as stats:
+        for line in stats:
+            tmp = line.split()
+            if len(tmp) > 1:
+                stats_dict[tmp[0]].append(
+                    tmp[1]
+                )  # The key is the name of the stat, the value is a list of
+                # numerical stat values in string format, e.g. "12".
```
BobbyRBruce, line 122: Why is the value here a list? Doesn't this mean two or more stats with the same name exist?

BobbyRBruce (changes requested):

tests/gem5/stats-reset/configs/x86-ubuntu-run-middle-reset.py
erin-le: Yes. There are two stats dumps in the test case that resets after 5 billion ticks and runs for an additional 10 billion, so we have to collect two values for the same stat/key. 
We then add the two values together and compare the sum to the value in the reference file.

tests/gem5/stats-reset/configs/x86-ubuntu-run-end-reset.py
```diff
@@ -140,24 +139,23 @@ def read_stats_files(filepath: str, stats_dict: dict) -> None:
...
 
 
-# checks to see if any items in a list match with a key/ part of a key.
-# This is used to filter for stats that appear for several components
-def is_match_key(key: str, match_list: list[str]) -> bool:
-    if len([item for item in match_list if item in key]) == 0:
-        return False
-    else:
-        return True
+def is_excluded(key: str) -> bool:
```
erin-le, line 153: This works, but makes selecting which stats to filter out less flexible. I had to change a few items in exclude_from_check and uncertain_exclude_from_check to filter out everything to exclude. For example, we can't select larger parts of keys like `clk_domain.clock` anymore since the keys are split by `.`. Everything works for now, but it could make things more difficult in the future.

tests/gem5/stats-reset/configs/x86-ubuntu-run-end-reset.py
```diff
@@ -0,0 +1,193 @@
...
+    "pwrStateTime",
+    "pwrStateResidencyTicks",
+    "idleFraction",
+    "notIdleFraction",
+]
+
+
+def is_excluded(key: str) -> bool:
+    return bool(
+        set(key.replace("::", ".").split(".")).intersection(
```
erin-le, line 155: This works, but makes selecting which stats to filter out less flexible. I had to change a few items in exclude_from_check and uncertain_exclude_from_check to filter out everything to exclude. For example, we can't select larger parts of keys like `clk_domain.clock` anymore since the keys are split by `.`. Everything works for now, but it could make things more difficult in the future.

BobbyRBruce (approved):
Thanks for making my suggested changes. Just for future: It's helpful to mark comments you've addressed as "resolved". It helps me understand what you've done rather than having to check for myself.

giactra:
Hi all, sorry to jump in only now but I somehow missed this PR. I didn't review it carefully but from what I have seen I have the following concerns:

- It looks like we are uploading a reference stat file. How is that used? We did rely on them in the old regression system and we ended up discarding them as any reasonable uarch change required generating a new stat file.
- Do we really need an integration test to verify stat reset? Shouldn't this be addressed with unit tests instead? I think they generalize better as this is only tests stat reset for a particular platform and a particular ISA and a particular workload...

erin-le:
@giactra: I brought up your concerns with @powerjg and @BobbyRBruce and they agreed that the reference stats file would cause problems. I initially set up these tests so that one test would run before the other two to generate the reference stats file. However, there isn't a way to guarantee that the tests will run in a certain order on the runners. I will close this PR and convert the code to a script that can be placed under `util` to test that the stats are resetting properly.

PR #1700, opened 2024-10-22
Mfma matrix core timing

mjkpolo:
A lookup table is used with values taken from the mi200 and mi300 ISA
specifying how long a SIMD's matrix core unit will be consumed by MFMAs.

This time can be adjusted for all MFMAs using the --mfma-scale parameter.


mattsinc:
Needs to be merged in after #1693

mattsinc (changes requested):
Since Vishnu already integrated Nagendra's patches in #1693, please remove those from this patch before we review.

configs/example/apu_se.py
```diff
@@ -303,6 +303,13 @@
     default=41,
     help="Latency for memtimes in scalar memory pipeline.",
 )
+parser.add_argument(
+    "--mfma-scale",
+    type=float,
+    # Set to a default of 1 to not scale MFMA cycles
+    default=1,
+    help="Scale how long an mfma consumes the matrix core unit",
```
mattsinc: Minor nit: consumes --> holds onto (?)

src/gpu-compute/compute_unit.cc
```diff
@@ -106,6 +107,61 @@ ComputeUnit::ComputeUnit(const Params &p) : ClockedObject(p),
     scalar_resp_tick_latency(
             p.scalar_mem_resp_latency * p.clk_domain->clockPeriod()),
     memtime_latency(p.memtime_latency * p.clk_domain->clockPeriod()),
+    mfma_scale(p.mfma_scale),
+    mfma_cycles({
+        {GfxVersion::gfx90a, {
```
mattsinc, line 118: please add a comment here identifying GFX90A as MI200 and gfx942 as MI300

src/gpu-compute/scoreboard_check_stage.cc
```diff
@@ -196,6 +196,21 @@ ScoreboardCheckStage::ready(Wavefront *w, nonrdytype_e *rdyStatus,
...
         }
     }
+
+    if (ii->isMFMA()) {
+        if (computeUnit.matrix_core_ready[w->simdId] <= curTick()) {
+            computeUnit.matrix_core_ready[w->simdId] =
+                curTick() +
+                computeUnit.cyclesToTicks(Cycles(
+                    computeUnit.mfma_scale *
+                    computeUnit.mfma_cycles[w->gfxVersion]
```
mattsinc, line 209: Probably we need an assert somewhere in this code block (between lines 201 and 200?) that fails if MFMAs are not supported in the GFX version being used?

mattsinc (changes requested):
Few small things, but overall nice job!

src/gpu-compute/compute_unit.cc
abmerop: I would additionally point to the ISA specs and mention that is where all of these numbers came from.  The table for MI300 and the "passes" in the descriptions for MI200

abmerop (commented):
This is mostly good, thanks!  I only have one additional comment on top of Matt's comments.

erin-le:
@mjkpolo Hello, thank you for your contribution! Could you give an update on the status of this PR? Also, if you are able to make the changes requested within the next day or so/by the end of this week, we may be able to get this PR into the staging branch.

mattsinc:
> @mjkpolo Hello, thank you for your contribution! Could you give an update on the status of this PR? Also, if you are able to make the changes requested within the next day or so/by the end of this week, we may be able to get this PR into the staging branch.

@mjkpolo is working towards an ISCA submission, so my guess is he won't be able to address these comments in the timeline you provided.

abmerop:
Should I just add a commit with the changes? It seems like we just want comments and an assert?

src/gpu-compute/compute_unit.cc
abmerop: done

src/gpu-compute/scoreboard_check_stage.cc
abmerop: done

mattsinc (approved):

mattsinc:
> Should I just add a commit with the changes? It seems like we just want comments and an assert?

Thanks @abmerop ! I also fixed the print info.

mattsinc:
> > Should I just add a commit with the changes? It seems like we just want comments and an assert?
> 
> Thanks @abmerop ! I also fixed the print info.

@abmerop since I pushed the print fix it won't let me approve ... can you?

abmerop (approved):

PR #1702, opened 2024-10-22
Add SE mode to X86Board and RiscvBoard

erin-le:
This PR adds SE mode to X86Board and RiscvBoard. X86DemoBoard and RiscvDemoBoard have also been modified, as the functions that were previously needed to add SE mode to them were removed.

### DIFF
--- a/src/python/gem5/components/boards/riscv_board.py
+++ b/src/python/gem5/components/boards/riscv_board.py
@@ -60,6 +60,7 @@ from m5.util.fdthelper import (
     FdtState,
 )
 
+from ...components.boards.se_binary_workload import SEBinaryWorkload
 from ...isas import ISA
 from ...resources.resource import AbstractResource
 from ...utils.override import overrides
@@ -70,7 +71,7 @@ from .abstract_system_board import AbstractSystemBoard
 from .kernel_disk_workload import KernelDiskWorkload
 
 
-class RiscvBoard(AbstractSystemBoard, KernelDiskWorkload):
+class RiscvBoard(AbstractSystemBoard, KernelDiskWorkload, SEBinaryWorkload):
     """
     A board capable of full system simulation for RISC-V.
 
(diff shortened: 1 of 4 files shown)


src/python/gem5/components/boards/riscv_board.py
```diff
@@ -100,47 +101,53 @@ def __init__(
...
+            # isn't general enough.
+            self._on_chip_devices = [self.platform.clint, self.platform.plic]
+            self._off_chip_devices = [self.platform.uart, self.disk, self.rng]
 
-        # Note: This overrides the platform's code because the platform isn't
-        # general enough.
-        self._on_chip_devices = [self.platform.clint, self.platform.plic]
-        self._off_chip_devices = [self.platform.uart, self.disk, self.rng]
+        else:
+            pass
```
BobbyRBruce, line 151: .Maybe just to make it very to obvious to anyone wanting to make changes, adds comment such as:

```suggestion
            # SE mode board setup.
            pass
```

src/python/gem5/components/boards/riscv_board.py
```diff
@@ -213,19 +220,31 @@ def get_dma_ports(self) -> List[Port]:
...
-        return True
+        return self.is_fullsystem()
 
     @overrides(AbstractSystemBoard)
     def get_io_bus(self) -> IOXBar:
-        return self.iobus
+        if self.has_io_bus():
+            return self.iobus
+        else:
+            raise NotImplementedError(
```
BobbyRBruce: I don't think this exception type is the right fit. It's well worth looking up the Python docs to see what an Exception is designed to do. Here is entry for `NotImplementedError` : <https://docs.python.org/3/library/exceptions.html#NotImplementedError>

It's usage is explained as:


> In user-defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.


So, this is just an exception designed to be raised in the case where someone inherits from a class but doesn't implement all the abstract methods they were supposed to.

To be honest, I've rarely found specific Exceptions in Python that really work for the kind of Exceptions we throw in gem5, so in the stdlib I've usually just used `raise Exception`, the most basic and general type. I'd recommend that.

It's been on my TODO list for a long time now to create a set of gem5 exceptions to cover this stuff (e.g., something like "InvalidGem5SimulationSetup" would be useful) but `Exception` is fine for now.

**((Do this update for all instances of this issue in this PR))**

BobbyRBruce: Do this for all the other case in the PR

src/python/gem5/components/boards/riscv_board.py
```diff
@@ -213,19 +220,31 @@ def get_dma_ports(self) -> List[Port]:
...
+                "Board was not configured for FS mode and does not have an "
+                "I/O bus. Use `has_io_bus()` to check this."
```
BobbyRBruce: I’d say this exception string is very specific to this board. You’ve got to remember that anyone can create a subclass of `RISCVBoard` and override `has_io_bus` to have a completely different logic. In such a situation when `get_io_bus` is called and `has_io_bus` returns `False` the exception thrown here won’t be correct.

The real, core reason this exception is raised is because we can't get the I/O bus from a board that doesn't have one. Calling this on a board without an I/O bus is simply invalid.

I’d just have the text read something like “Cannot execute `get_io_bus` : Board does not have an I/O bus to return. The board’s `has_io_bus` method can be used to check this.”

**((Make this change for other, similar exception text in this PR))**

BobbyRBruce (changes requested):

BobbyRBruce (approved):

Harshil2107 (approved):
lgtm

BobbyRBruce (approved):

PR #1703, opened 2024-10-23
cpu: Add FastMemSimpleCPU model

h0lyalg0rithm:
FastMemSimpleCPU doesnt use caches for memory access but instead accesses the memory directly.
This speeds up checkpointing considerably where accuracy of the simulation is not required.

In this PR, i have added the cpu model for all of the architectures.

@aarmejach 

Change-Id: Ia21757dad5a13b741b46d541f074fb6569098809

### DIFF
--- a/src/arch/arm/ArmCPU.py
+++ b/src/arch/arm/ArmCPU.py
@@ -28,6 +28,7 @@ from m5.objects.ArmInterrupts import ArmInterrupts
 from m5.objects.ArmISA import ArmISA
 from m5.objects.ArmMMU import ArmMMU
 from m5.objects.BaseAtomicSimpleCPU import BaseAtomicSimpleCPU
+from m5.objects.BaseFastMemSimpleCPU import BaseFastMemSimpleCPU
 from m5.objects.BaseMinorCPU import BaseMinorCPU
 from m5.objects.BaseNonCachingSimpleCPU import BaseNonCachingSimpleCPU
 from m5.objects.BaseO3Checker import BaseO3Checker
@@ -48,6 +49,10 @@ class ArmAtomicSimpleCPU(BaseAtomicSimpleCPU, ArmCPU):
 
 
 class ArmNonCachingSimpleCPU(BaseNonCachingSimpleCPU, ArmCPU):
+    mmu = ArmMMU()
+
+
+class ArmFastMemSimpleCPU(BaseFastMemSimpleCPU, ArmCPU):
     mmu = ArmMMU()
 
 
--- a/src/arch/mips/MipsCPU.py
+++ b/src/arch/mips/MipsCPU.py
@@ -24,6 +24,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 from m5.objects.BaseAtomicSimpleCPU import BaseAtomicSimpleCPU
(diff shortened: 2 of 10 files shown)


powerjg (commented):
This looks great to me! There are a couple of things which we should do. I am mixed on whether these are blockers or not.

1. Update documentation. With this new CPU type we will need to update the website with documentation describing when it is appropriate to use.
2. Add tests. We don't want this code to rot, so we should extend some of the general gem5 tests (e.g., boot test, simple SE tests, etc.) to use this CPU type as well as the timing, atomic, o3, and minor CPUs.
3. We need to make sure this is compatible with the standard library. There are a few places in the stdlib where the CPU types are hard coded that will need to be updated.

Generally, I'm ok if these are done in other PRs, but it is important to get them done before the release in a few weeks.  Unfortunately, with the release coming up, I can't allocate resources myself to doing these things.

giactra:
What is the difference between this and the NonCachingSimpleCPU?

h0lyalg0rithm:
> What is the difference between this and the NonCachingSimpleCPU?

Hi @giactra Now that you mentioned it, it similar to NonCachingSimpleCPU, without the memory backdoor but it access the memory directly without going through the memory controller

aarmejach:
Yes, `NonCachingSimpleCPU` still goes through the memory controller. On large-ish systems that use the ruby models this is still a large overhead for checkpointing, as you have to go through the MachineID selection (which takes most of the simulation time). IIRC, on our checkpoints, the use of `FastMemSimpleCPU` speeds up checkpointing by more than 2x.

giactra:
> Hi @giactra Now that you mentioned it, it similar to NonCachingSimpleCPU, without the memory backdoor but it access the memory directly without going through the memory controller

Interestingly this was the behaviour of the NonCachingSimpleCPU before the backdoor introduction [1].
I didn't go through the CR [1] but it's clear from code + commit message that the intended behaviour in the long term was to enable backdoors for data accesses, whereas at the moment it is only used by fetch. And this is why you see memory controller accesses.

I think before we merge this we should evaluate

1) Should we rework the NonCachingSimpleCPU to use backdoors for data accesses as well? What the performance will be wrt to FastMemSimpleCPU?
2) Should we introduce a new FastMemSimpleCPU or should we have a single parametrizable one?
3) If we add FastMemSimpleCPU, what the design should look like?
    - Should NonCaching and FastMemSimpleCPU be related somehow?

[1]: https://gem5-review.googlesource.com/c/public/gem5/+/36979/6/src/cpu/simple/noncaching.cc

giactra:
@andysan

giactra:
In other words (and apologies if I reiterate over the same concepts) it seems to me the FastMem is reimplementing how the NonCachingSimpleCPU was before the backdoor mechanism. 
I worry about having two cpus that are almost identical; firstly because it will be difficult for a user to grasp the actual differences; secondly because these difference might be removed in the future with the reworking of NonCachingSimpleCPU data accesses to use the backdoor

powerjg:
Thanks @giactra! I totally forgot about the `NonCachingSimpleCPU`. Goes to show how important documentation, tests, and compatibility with how others use gem5 is 😄

erin-le:
@h0lyalg0rithm : Closing this due to inactivity, but feel free to reopen this if you want to make changes or work on it further.

PR #1705, opened 2024-10-23
tests,util: Extend SE mode m5-exit tests

BobbyRBruce:
These tests expand the m5-exit tests to target other ISAs using magic instructions and address approaches.

The motivating factor in creating these tests was to verify https://github.com/gem5/gem5/pull/1583

### DIFF
--- a/tests/gem5/m5_util/configs/se-m5-exit.py
+++ b/tests/gem5/m5_util/configs/se-m5-exit.py
@@ -39,7 +39,7 @@ Usage
 `<isas>` can be one of `x86`, `arm`, or `riscv`.
 
 
-The `--resource-directory` argument is optional, and can be passed with a 
+The `--resource-directory` argument is optional, and can be passed with a
 <path> in order to specify where the downloaded resources should be cached.
 If not set, the downloaded resources are cached to the default location for
 gem5 binaries (typically "~/.cache/gem5/resources").
--- a/tests/gem5/m5_util/configs/se-m5-exit.py
+++ b/tests/gem5/m5_util/configs/se-m5-exit.py
@@ -39,10 +39,10 @@ Usage
 `<isas>` can be one of `x86`, `arm`, or `riscv`.
 
 
-The `--resource-directory` argument can be passed and along with path to the a
-directory the downloaded resources are to cached. If not set the downloaded
-resources are cached by the gem5 binary's default cache location defaults
-(typically "~/.cache/gem5/resources").
+The `--resource-directory` argument is optional, and can be passed with a 
+<path> in order to specify where the downloaded resources should be cached.
+If not set, the downloaded resources are cached to the default location for
+gem5 binaries (typically "~/.cache/gem5/resources").
 """
 
 import argparse
--- a/tests/gem5/m5_util/configs/se-m5-exit.py
+++ b/tests/gem5/m5_util/configs/se-m5-exit.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2021 The Regents of the University of California
+# Copyright (c) 2021-2026 The Regents of the University of California
 # Copyright (c) 2022 Google Inc
 # All rights reserved.
 #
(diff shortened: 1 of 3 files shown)


BobbyRBruce:
This PR is still in a "draft" state but I hope it's clear what's going on. I build the m5-exit binaries locally (in the "tests/gem5/m5_util/m5_exit" dir. They can be build with `make`), then run in the "tests" directory with: `./main.py run -vvv --skip-build gem5/m5_util`.

The status of the tests is as follows:

```sh
Test: m5_exit_test_x86_addr-ALL-x86_64-opt Passed
Test: m5_exit_test_x86_addr-ALL-x86_64-opt-MatchRegex Failed
Test: m5_exit_test_x86_inst-ALL-x86_64-opt Passed
Test: m5_exit_test_x86_inst-ALL-x86_64-opt-MatchRegex Passed
Test: m5_exit_test_arm_addr-ALL-x86_64-opt Passed
Test: m5_exit_test_arm_addr-ALL-x86_64-opt-MatchRegex Failed
Test: m5_exit_test_arm_inst-ALL-x86_64-opt Passed
Test: m5_exit_test_arm_inst-ALL-x86_64-opt-MatchRegex Passed
Test: m5_exit_test_riscv_inst-ALL-x86_64-opt Passed
Test: m5_exit_test_riscv_inst-ALL-x86_64-opt-MatchRegex Passed
```

Failing tests aren't returning a non-zero exit code, but failing because they don't return what we expect from the stdout. The stdout is:

```sh
src/sim/syscall_emul.cc:86: warn: ignoring syscall set_robust_list(...)
src/sim/syscall_emul.cc:98: warn: ignoring syscall rseq(...)
      (further warnings will be suppressed)
src/sim/syscall_emul.hh:1119: warn: readlink() called on '/proc/self/exe' may yield unexpected results in various settings.
      Returning '/Users/bobbyrbruce/Workspaces/gem5/tests/gem5/m5_util/m5_exit/bin/arm-addr-m5-exit'
src/arch/arm/insts/pseudo.cc:175: warn:         instruction 'bti' unimplemented
src/sim/syscall_emul.cc:86: warn: ignoring syscall mprotect(...)
Can't open /dev/mem: No such file or directory
Exiting @ tick 31972329 because exiting with last active thread context.
```

I don't understand the `Can't open /dev/mem` error and how to resolve it. I also don't understand why this error doesn't result in a non-zero exit.

I'm very new to the address-based approach to m5-exits and perhaps i'm doing something silly here.

Harshil2107:
@BobbyRBruce  The error is happening because m5ops need to access `/dev/mem` to map the addresses. You can see it in https://github.com/gem5/gem5/blob/b1a44b89c7bae73fae2dc547bc1f871452075b85/util/m5/src/m5_mmap.c#L60

Running the binary with `sudo` solves the error as that grants it the necessary permissions.

BobbyRBruce:
> @BobbyRBruce  The error is happening because m5ops need to access `/dev/mem` to map the addresses. You can see it in https://github.com/gem5/gem5/blob/b1a44b89c7bae73fae2dc547bc1f871452075b85/util/m5/src/m5_mmap.c#L60
> 
> 
> 
> Running the binary with `sudo` solves the error as that grants it the necessary permissions.


Sorry I feel a little confused by this.

This is an SE Mode simulation. You can't run the binary with sudo, right? 

If what. that's true then there's no way for the address based approach to work in SE mode.

BobbyRBruce:
> > @BobbyRBruce  The error is happening because m5ops need to access `/dev/mem` to map the addresses. You can see it in https://github.com/gem5/gem5/blob/b1a44b89c7bae73fae2dc547bc1f871452075b85/util/m5/src/m5_mmap.c#L60
> > 
> > Running the binary with `sudo` solves the error as that grants it the necessary permissions.
> 
> Sorry I feel a little confused by this.
> 
> This is an SE Mode simulation. You can't run the binary with sudo, right?
> 
> If what. that's true then there's no way for the address based approach to work in SE mode.

My understanding of all this has improved. I have realized this is true. Keeping these tests here as they are still of some use.

BobbyRBruce:
The first commit in this PR contains sources to build resources and run them locally. The second commit removes the sources and tries to obtain them from gem5-resources. At the time of writing, this 2nd commit causes the tests to fail as these resources have not been uploaded to gem5 resources and therefore there's nothing to obtain.

TODO: Move the sources in the first commit to the gem5-resources src repo and upload the binaries these sources produce to gem5-resources. When done, the these should pass and this PR squash-merged.

BobbyRBruce:
> The first commit in this PR contains sources to build resources and run them locally. The second commit removes the sources and tries to obtain them from gem5-resources. At the time of writing, this 2nd commit causes the tests to fail as these resources have not been uploaded to gem5 resources and therefore there's nothing to obtain.
> 
> TODO: Move the sources in the first commit to the gem5-resources src repo and upload the binaries these sources produce to gem5-resources. When done, the these should pass and this PR squash-merged.

@Harshil2107 : It seems like this is all that needs done to complete this PR. When you have time can you do this?

BobbyRBruce:
@Harshil2107 : I can't recreate the error you've been having. All I needed to do was checkout [a053b8g](https://github.com/gem5/gem5/pull/1705/commits/a05eb8f5b5955f88873df6366891c1972e5c455b) then:

```shell
cd tests/gem5/m5_util/m5_exit
make
```

It requires you to have docker installed but should work.

Attached are the binaries but you'll need to migrate this code to the gem5-resources repository.

[bin.zip](https://github.com/user-attachments/files/19695159/bin.zip)

Harshil2107:
I have added the binaries to resoruces. Note that the binary already existed for x86 so I added a new version.

Here are the links for them on resources website:

https://resources.gem5.org/resources/arm-m5-exit?version=1.0.0
https://resources.gem5.org/resources/riscv-m5-exit?database=gem5-resources&version=1.0.0
https://resources.gem5.org/resources/x86-m5-exit?version=2.0.0

BobbyRBruce:
@erin-le : I don't think there is nothing blocking this PR. I think it's done... it's just waiting for a review. The failing test looked like a runner issue. I've rebased and I am rerunning the test.

Not sure why I assumed there was work to do here. From the looks of it this is complete. So, please review and merge if you think it's good enough.

tests/gem5/m5_util/configs/se-m5-exit.py
```diff
@@ -26,61 +26,99 @@
...
-The system has no cache heirarchy and is as "bare-bones" as you can get in
-gem5 while still being functinal.
+A script used to test the m5_exit magic instruction events function correctly
+for a given ISA.
+
+Usage
+-----
+
+```sh
+<gem5> test/gem5/m5_util/configs/se-m5-exit.py <isa>
```
erin-le:
```suggestion
<gem5> test/gem5/m5_util/configs/se-m5-exit.py <isa> --resource-directory <path>
```

tests/gem5/m5_util/configs/se-m5-exit.py
```diff
@@ -26,61 +26,99 @@
...
+```
+
+`<isas>` can be one of `x86`, `arm`, or `riscv`.
+
+`<path-to-resources>` is the path to the directory where the resources are
+stored.
```
erin-le:
```suggestion
```

tests/gem5/m5_util/configs/se-m5-exit.py
```diff
@@ -26,61 +26,99 @@
...
+
+The `--resource-directory` argument can be passed and along with path to the a
+directory the downloaded resources are to cached. If not set the downloaded
+resources are cached by the gem5 binary's default cache location defaults
+(typically "~/.cache/gem5/resources").
```
erin-le:
```suggestion
The `--resource-directory` argument is optional, and can be passed with a 
<path> in order to specify where the downloaded resources should be cached.
If not set, the downloaded resources are cached to the default location for
gem5 binaries (typically "~/.cache/gem5/resources").
```

tests/gem5/m5_util/test_exit.py
```diff
@@ -37,39 +38,47 @@
...
+        config=str(
+            Path(
+                Path(__file__).parent,
+                "configs",
+                "se-m5-exit.py",
+            )
+        ),
+        config_args=[
+            isa,
+            f"--resource-directory='{resource_directory}'",
```
erin-le:
```suggestion
            f"--resource-directory={resource_directory}",
```
Without this change, the resources are put into nested subdirectories in `tests`,  with the format `tests/'/home/.../gem5/tests/gem5/resources'`, because the first `'` is counted as a directory name.

erin-le (changes requested):
Hey @BobbyRBruce, I took a look at this PR and it mostly looks good. I made some suggestions to make comments clearer, and also suggested a minor fix to save the resources to the intended directory.

erin-le (dismissed):

erin-le (approved):

PR #1709, opened 2024-10-24
arch-riscv: Fix misprediction of control flow instruction caused by vset{i}vl{i}

### LINKED ISSUE #1708: Control flow instructions unexpectedly flush the pipeline after a vset instruction
**Describe the bug**
The current BPU predicts pc, vl, and vtype for all RISCV instructions. However, control flow instructions other than vset{i}vl{i} do not need to predict vl and vtype. This can lead to unexpected mispredictions.

The pipeline picture below shows this problem. In a simple `for` loop, the vset instruction modifies vl and vtype, causing the subsequent control flow instructions to flush the pipeline. This is not what we expect.

截屏2024-10-24 下午4 57 04

The reason for mispredictions of function return instructions is:
1. The function call instruction pushes pcstate (pc+4, vl, and vtype) into RAS.
2. The vset instruction modifies vl and vtype.
3. The function return instruction pops pcstate (pc+4, vl, and vtype) from RAS and checks the correctness of pcstate, finding that vl and vtype are predicted incorrectly. Then the pipeline is flushed.

I didn't look for the reason of the unexpected misprediction of conditional branches. But I think it has similar reason.

**Affects version**
Develop branch. Commit [709f2c769534936e0b154fdbb78bc535de9ae04e](https://github.com/gem5/gem5/tree/709f2c769534936e0b154fdbb78bc535de9ae04e).

**gem5 Modifications**
#1709

**To Reproduce**
[Here](https://github.com/user-attachments/files/17504208/vset_ret.tar.gz) is a simple test. Use the `vset_ret` elf file in gz package file and `configs/deprecated/example/se.py` script to simulate. And dump the `O3PipeView`.
```shell
./build/RISCV/gem5.opt --debug-flags=O3PipeView --debug-start=67978500 --debug-file=trace.out  configs/deprecated/example/se.py --cpu-type RiscvO3CPU --caches --l1d_size 64kB --l1d_assoc 4 --l1i_size 64kB --l1i_assoc 4 --l2cache --l2_size 1024kB --l2_assoc 16 --cacheline_size 64 --mem-size 1GB --mem-type DDR4_2400_8x8 --mem-channels 2 --bp-type TAGE_SC_L_64KB --indirect-bp-type SimpleIndirectPredictor --cmd vset_ret
./util/o3-pipeview.py -c 500 -o pipeview.out --color m5out/trace.out
less -r pipeview.out
# Search ‘vset’ to locate the vsetivli instruction.
```

**Expected behavior**
After warmup, there is no flushing in the `for` loop.

**Host Operating System**
Debian 12.

Host ISA
X86

Compiler used
gcc 14

5265325:
This PR fixed the bug mentioned in #1708.

In the new implementation, the BPU only predicts vl&vtype for vset{i}vl{i} instructions. Other control flow instructions will not be predicted by the BPU for vl and vtype, because it is meaningless. The new implementation adds `new_conf` flag in pcstate and `vl/vtype` registers in decoder. This is used to filter the prediction of `vl&vtype` by other control flow instructions. When a flush occurs, the `vl/vtype` registers in the decoder are restored.

When checking pcstate at IEW stage, `vl&vtype` are only checked for instructions with new_vconf = 1. It means that vl and vtype are only checked for vset{i}vl{i} instructions.

Here is a simple picture describing the new changes：
绘图1

The following picture is the new pipeline of the test case from #1708. No more flush after warmup.
截屏2024-10-24 下午4 08 02

Change-Id: I7d3df0328d5757f6b3506743200d5b154dee367f

### DIFF
--- a/src/arch/riscv/isa/decoder.isa
+++ b/src/arch/riscv/isa/decoder.isa
@@ -4839,6 +4839,7 @@ decode QUADRANT default Unknown::unknown() {
                         Rd_ud = new_vl;
                         VL = new_vl;
                         Vtype = new_vtype;
+                        IsVset = true;
                     }}, VSetVlDeclare, VSetVliBranchTarget
                       , SimdConfigOp, IsUncondControl
                       , IsIndirectControl);
@@ -4854,6 +4855,7 @@ decode QUADRANT default Unknown::unknown() {
                             Rd_ud = new_vl;
                             VL = new_vl;
                             Vtype = new_vtype;
+                            IsVset = true;
                         }}, VSetVlDeclare, VSetVlBranchTarget
                           , SimdConfigOp, IsUncondControl
                           , IsIndirectControl);
@@ -4868,6 +4870,7 @@ decode QUADRANT default Unknown::unknown() {
                             Rd_ud = new_vl;
                             VL = new_vl;
                             Vtype = new_vtype;
+                            IsVset = true;
                         }}, VSetiVliDeclare, VSetiVliBranchTarget
                           , SimdConfigOp, IsUncondControl
                           , IsDirectControl);
--- a/src/arch/riscv/pcstate.hh
+++ b/src/arch/riscv/pcstate.hh
@@ -65,6 +65,7 @@ class PCState : public GenericISA::UPCState<4>
 
     bool _compressed = false;
     RiscvType _rvType = RV64;
+    bool _new_vconf = false;
     VTYPE _vtype = (1ULL << 63); // vtype.vill = 1 at initial;
     uint32_t _vl = 0;
     bool _zcmtSecondFetch = false;
@@ -73,8 +74,8 @@ class PCState : public GenericISA::UPCState<4>
   public:
     PCState(const PCState &other) : Base(other),
         _compressed(other._compressed),
-        _rvType(other._rvType), _vtype(other._vtype), _vl(other._vl),
-        _zcmtSecondFetch(other._zcmtSecondFetch), _zcmtPc(other._zcmtPc)
+        _rvType(other._rvType), _new_vconf(other._new_vconf), _vtype(other._vtype),
+        _vl(other._vl), _zcmtSecondFetch(other._zcmtSecondFetch), _zcmtPc(other._zcmtPc)
     {}
     PCState &operator=(const PCState &other) = default;
     PCState() = default;
@@ -94,6 +95,7 @@ class PCState : public GenericISA::UPCState<4>
         auto &pcstate = other.as<PCState>();
         _compressed = pcstate._compressed;
         _rvType = pcstate._rvType;
+        _new_vconf = pcstate._new_vconf;
         _vtype = pcstate._vtype;
         _vl = pcstate._vl;
         _zcmtSecondFetch = pcstate._zcmtSecondFetch;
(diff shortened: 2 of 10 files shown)


saul44203 (commented):
I think we can remove `BranchTarget` for both `vsetvli` and `vsetvl` since they are `IsIndirectControl` and will not execute this anyways (only `vsetivli` really needs it). Apart from this it looks good to me.

erin-le:
@5265325 Hello, thanks for your contribution! Could you respond to the comment on this change?

saul44203:
Hi @5265325. Assuming this is done and tested, LGTM.

src/arch/riscv/pcstate.hh
```diff
@@ -120,15 +126,16 @@ class PCState : public GenericISA::UPCState<4>
...
+            (_new_vconf == opc._new_vconf) &&
+            (!_new_vconf || (_vtype == opc._vtype && _vl == opc._vl));
     }
 
     void
     serialize(CheckpointOut &cp) const override
     {
         Base::serialize(cp);
         SERIALIZE_SCALAR(_rvType);
+        SERIALIZE_SCALAR(_new_vconf);
```
hnpl, line 150: @Harshil2107 Does this require updating the checkpointing upgrade script?

hnpl (approved):
The code looks good to me.

We probably need update the checkpoint upgrader but it can be done in a different commit.

src/arch/riscv/isa/decoder.isa
```diff
@@ -4845,6 +4846,7 @@ decode QUADRANT default Unknown::unknown() {
                             Rd_ud = new_vl;
                             VL = new_vl;
                             Vtype = new_vtype;
+                            IsVset = true;
                         }}, VSetVlDeclare, VSetVlBranchTarget
                           , SimdConfigOp, IsUncondControl
                           , IsIndirectControl);
```
rogerchang23424, line 4861: I noticed you removed the `branchTarget` for `vsetvl` and `vsetvli`. Not sure if we need to remove the instruction flag `IsIndirectControl` and `IsUncondControl` and mark them `IsSerializeAfter` and `IsNonSpeculative`. Not sure if there will be the issue if BPU support indirect jump prediction in the future.

5265325: I had similar thoughts. At least vsetivli instructions don't need to be speculated on. I think we can create a new pr after the current pr is merged to modify the code and discuss it.

rogerchang23424 (approved):

src/arch/riscv/pcstate.hh
5265325: Currently the iscv-hello-example-checkpoint test uses checkpoint from http://dist.gem5.org/dist/develop/checkpoints/riscv-hello-example-checkpoint.tar. But the checkpoint does not contain `_new_vconf` entry. So it cannot pass the test in CI.

saul44203 (approved):

powerjg:
@Harshil2107 can you work with @5265325 to update the checkpoints for the CI tests?

(Note: checkpoint updates should probably be included in the release notes)

Harshil2107:
I tested the checkpoint upgrader, and it works fine. However, the updated section needs to be in a new file because modifying an existing upgrader file won’t reapply changes to checkpoints that already have the tag from the previous version. Is it okay if I update the upgrader and push a commit?

5265325:
> I tested the checkpoint upgrader, and it works fine. However, the updated section needs to be in a new file because modifying an existing upgrader file won’t reapply changes to checkpoints that already have the tag from the previous version. Is it okay if I update the upgrader and push a commit?

Thank you for testing the checkpoint upgrader and providing feedback! Feel free to do it if you need to.

erin-le (approved):

PR #1711, opened 2024-10-24
arch-riscv: Fix vector instruction assertion caused by speculative execution

### LINKED ISSUE #1645: riscv simulation raises panic condition
**Describe the bug**
running vlseg4e32 instruction causes panic condition

**Affects version**
commit b1a44b89c

**gem5 Modifications**

building the vec-fft benchmark from https://github.com/ucb-bar/saturn-vectors/tree/master/benchmarks/vec-fft

1. Compile gem5 with command ...  scons build/RISCV/gem5.opt
2. Execute the simulation with...
riscv64-unknown-elf-gcc -DPREALLOCATE=1 -march=rv64imafdcv -mabi=lp64d -O3 -ftree-vectorize  -ffast-math -fno-common fft2_main.c vec-fft2.c  -o FFT

build/RISCV/gem5.opt ../Test/riscv.py

```python
import m5
from m5.objects import *


class L1Cache(Cache):
    assoc = 2
    tag_latency = 1
    data_latency = 1
    response_latency = 1
    mshrs = 4
    tgts_per_mshr = 20

    def connectCPU(self, cpu):
        raise NotImplementedError

    def connectBus(self, bus):
        self.mem_side = bus.cpu_side_ports


class L1ICache(L1Cache):
    size = '512kB'

    def connectCPU(self, cpu):
        self.cpu_side = cpu.icache_port


class L1DCache(L1Cache):
    size = '1024kB'

    def connectCPU(self, cpu):
        self.cpu_side = cpu.dcache_port


class L2Cache(Cache):
    size = '1024kB'
    assoc = 8
    tag_latency = 1
    data_latency = 1
    response_latency = 1
    mshrs = 20
    tgts_per_mshr = 12

    def connectCPUSideBus(self, bus):
        self.cpu_side = bus.mem_side_ports

    def connectMemSideBus(self, bus):
        self.mem_side = bus.cpu_side_ports


system = System()

system.clk_domain = SrcClockDomain()
system.clk_domain.clock = "800MHz"
system.clk_domain.voltage_domain = VoltageDomain()

system.mem_mode = "timing"
system.mem_ranges = [AddrRange("1024MB")]
system.cpu = RiscvO3CPU()
#system.cpu = RiscvMinorCPU()
system.cpu.isa = RiscvISA(vlen=256, elen=32)

# Create the L1 caches
system.cpu.icache = L1ICache()
system.cpu.dcache = L1DCache()

# Connect the caches to the CPU
system.cpu.icache.connectCPU(system.cpu)
system.cpu.dcache.connectCPU(system.cpu)

# Connect the CPU to the L2 bus
system.l2bus = L2XBar()

# Connect the L1 caches to the L2 bus
system.cpu.icache.connectBus(system.l2bus)
system.cpu.dcache.connectBus(system.l2bus)

# Connect the L2 cache to the CPU side bus
system.l2cache = L2Cache()
system.l2cache.connectCPUSideBus(system.l2bus)

# Connect the L2 cache to the memory bus
system.membus = SystemXBar()
system.l2cache.connectMemSideBus(system.membus)

# Connect the CPU to the memory bus
system.cpu.createInterruptController()


system.mem_ctrl = MemCtrl()
system.mem_ctrl.dram = DDR3_1600_8x8()
system.mem_ctrl.dram.range = system.mem_ranges[0]
system.mem_ctrl.port = system.membus.mem_side_ports

thispath = os.path.dirname(os.path.realpath(__file__))
binary = '/vec-fft/FFT'

system.workload = SEWorkload.init_compatible(binary)

process = Process()
process.cmd = [binary]
system.cpu.workload = process
system.cpu.createThreads()
root = Root(full_system=False, system=system)
m5.instantiate()

print(f"Beginning simulation!")
exit_event = m5.simulate()
  

```

**Terminal Output**
Beginning simulation!
build/RISCV/arch/riscv/generated/exec-ns.cc.inc:10499: panic: panic condition (pow(2, vlmul) * this->numFields) > 8 occurred: LMUL value is illegal for vlseg inst

**Expected behavior**
Spike simulation goes through without any issues. 

**Host Operating System**
Ubuntu 22.04 on WSL.

**Host ISA**
RISC-V

5265325:
Tihs PR fixed vector instruction assertion caused by speculative execution. #1645

The approach refers to the spike simulator.
https://github.com/riscv-software-src/riscv-isa-sim/blob/88fc84ded155a9e01987c4dfb7a77800e69b232b/riscv/v_ext_macros.h#L77

When lmul is found to be illegal, an instruction exception is triggered.

Change-Id: If4a8782e65ba80633df74aec8ad287f7d92b2dca

### DIFF
--- a/src/arch/riscv/isa/templates/vector_arith.isa
+++ b/src/arch/riscv/isa/templates/vector_arith.isa
@@ -367,8 +367,6 @@ template<typename ElemType>
     %(set_reg_idx_arr)s;
     %(constructor)s;
     const int64_t vlmul = vtype_vlmul(_machInst.vtype8);
-    // Todo: move to Decode template
-    panic_if(vlmul == 3, "LMUL=8 is illegal for widening inst");
     // when LMUL setted as m1, need to split to 2 micro insts
     const uint32_t num_microops = 1 << std::max<int64_t>(0, vlmul + 1);
 
@@ -462,6 +460,8 @@ Fault
 
     if (machInst.vill)
         return std::make_shared<IllegalInstFault>("VILL is set", machInst);
+    if (vtype_vlmul(machInst.vtype8) == 3)
+        return std::make_shared<IllegalInstFault>("LMUL=8 is illegal for widening inst", machInst);
 
     status.vs = VPUStatus::DIRTY;
     xc->setMiscReg(MISCREG_STATUS, status);
@@ -747,6 +747,8 @@ Fault
 
     if (machInst.vill)
         return std::make_shared<IllegalInstFault>("VILL is set", machInst);
+    if (vtype_vlmul(machInst.vtype8) == 3)
+        return std::make_shared<IllegalInstFault>("LMUL=8 is illegal for widening inst", machInst);
 
     status.vs = VPUStatus::DIRTY;
     xc->setMiscReg(MISCREG_STATUS, status);
@@ -1884,6 +1886,8 @@ Fault
 
     if (machInst.vill)
         return std::make_shared<IllegalInstFault>("VILL is set", machInst);
+    if (vtype_vlmul(machInst.vtype8) == 3)
+        return std::make_shared<IllegalInstFault>("LMUL=8 is illegal for widening inst", machInst);
 
     status.vs = VPUStatus::DIRTY;
     xc->setMiscReg(MISCREG_STATUS, status);
--- a/src/arch/riscv/isa/templates/vector_mem.isa
+++ b/src/arch/riscv/isa/templates/vector_mem.isa
@@ -1792,9 +1792,8 @@ Fault
         return std::make_shared<IllegalInstFault>("VILL is set", machInst);
 
     const int64_t vlmul = vtype_vlmul(machInst.vtype8);
-
-    panic_if(((1 << vlmul) * this->numFields) > 8,
-        "LMUL value is illegal for vlseg inst");
+    if (((1 << vlmul) * this->numFields) > 8)
+        return std::make_shared<IllegalInstFault>("LMUL value is illegal for vlseg inst", machInst);
 
     status.vs = VPUStatus::DIRTY;
     xc->setMiscReg(MISCREG_STATUS, status);
@@ -1851,9 +1850,8 @@ Fault
         return std::make_shared<IllegalInstFault>("VILL is set", machInst);
 
     const int64_t vlmul = vtype_vlmul(machInst.vtype8);
-
-    panic_if(((1 << vlmul) * this->numFields) > 8,
-        "LMUL value is illegal for vlseg inst");
+    if (((1 << vlmul) * this->numFields) > 8)
+        return std::make_shared<IllegalInstFault>("LMUL value is illegal for vlseg inst", machInst);
 
     const std::vector<bool> byte_enable(mem_size, true);
     Fault fault = initiateMemRead(xc, EA, mem_size, memAccessFlags,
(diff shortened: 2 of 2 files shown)


powerjg (commented):
This looks good to me. @h0lyalg0rithm or @aarmejach may be good people to take a look as well.

saul44203:
Hey @5265325, I think you mixed in some other issue. This does not fix #1645. The panic that's triggered is from a segmented load.

Apart from this, if you would do the same with that panic, how would that solve the issue? The illegal instruction fault would also halt the execution, right?

5265325:
> Hey @5265325, I think you mixed in some other issue. This does not fix #1645. The panic that's triggered is from a segmented load.
> 
> Apart from this, if you would do the same with that panic, how would that solve the issue? The illegal instruction fault would also halt the execution, right?

I think I did mix it up. But there is indeed a same panic in widening instruction. These panics have the same triggering cause.

Illegal instructions can indeed halt execution. But it must be when the instruction is committed, not when it is decoded or executed. When an instruction in the wrong path is decoded, the assertion may be triggered. But the instruction will flushed later in the pipeline. We don't want the panic to be triggered in the wrong path.

aarmejach:
I think the problem will persist. I do not see how you can execute an illegal instruction, the fault will trigger and stop program execution if the instruction is not squashed on time.

There needs to be another mechanism to handle illegal decoding due to vtype speculation to identify when it is an actual illegal instruction or just a mispeculation event.

I do not have an answer for this with the way things are implemented in develop. We have an internal implementation of vset{i}vl{i} that momentarily stalls decode until the in-flight vset{i}vl{i} has been executed (not committed), so we always decode non-speculatively from the vset{i}vl{i} point of view.

5265325:
> I think the problem will persist. I do not see how you can execute an illegal instruction, the fault will trigger and stop program execution if the instruction is not squashed on time.
> 
> There needs to be another mechanism to handle illegal decoding due to vtype speculation to identify when it is an actual illegal instruction or just a mispeculation event.
> 
> I do not have an answer for this with the way things are implemented in develop. We have an internal implementation of vset{i}vl{i} that momentarily stalls decode until the in-flight vset{i}vl{i} has been executed (not committed), so we always decode non-speculatively from the vset{i}vl{i} point of view.

@aarmejach @saul44203

I did not execute any illegal instructions. Illegal instruction detection just moves from the decoding function to the beginning of the execution function. If the detection fails, the fault reason is returned instead of triggering panic. The exception is triggered when the instruction is finally committed by rob. This will filter out unexpected exceptions on the wrong path.

Here are 2 simple tests for this patch.

[Test1](https://github.com/user-attachments/files/17647069/vector_assertion_test1.tar.gz) is a test for `legal` widening instruction. The `vector_assertion_test` function contains only a few test instructions. It should not trigger exception or panic in any model. In the o3 model of the current patch, it works fine. However, it triggered a panic in the o3 model of the develop branch. This is not as expected.

[Test2](https://github.com/user-attachments/files/17647070/vector_assertion_test2.tar.gz)  is a test for `illegal` widening instruction. It triggers an exception in the o3 model of the current patch. And it triggers a panic in the develop branch. This is as expected.

Simulation Command:
```shell
./build/RISCV/gem5.opt configs/deprecated/example/se.py --cpu-type RiscvO3CPU --caches --cmd vector_assertion_test1/vector_assertion
```

The segmented load instruction has the same problem, which I have fixed in the patch.

aarmejach:
Thanks for the clarifications, tests helped and working on our end. This lgtm

aarmejach (approved):

erin-le (approved):
LGTM

PR #1712, opened 2024-10-24
arch-riscv: Fix incorrect vector slide instructions and statically filter redundant uops

### LINKED ISSUE #1667: Incorrect riscv vslide instruction implementation
**Describe the bug**
The execution result of the riscv vector slide instruction is not as expected. The picture of trace here shows the bug. (vlen=128)

The results in trace do not conform to the description in the [specification](https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#vector-slide-instructions).

截屏2024-10-12 下午4 50 01

The correct result of vslideup(0x106b4) shoule be:
```
vslideup_vi_micro v0 : SimdMisc :  D=[00000000_00000000_00000000_00000000]
vslideup_vi_micro v1 : SimdMisc :  D=[00010101_01010101_01010101_01010101]
vslideup_vi_micro v2 : SimdMisc :  D=[01020202_02020202_02020202_02020202]
vslideup_vi_micro v3 : SimdMisc :  D=[02030303_03030303_03030303_03030303]
```

 The correct result of vslidedown(0x106dc) shoule be:
```
vslidedown_vi_micro v0 : SimdMisc :  D=[02020202_02020202_02020202_02020203]
vslidedown_vi_micro v1 : SimdMisc :  D=[03030303_03030303_03030303_03030304]
vslidedown_vi_micro v2 : SimdMisc :  D=[04040404_04040404_04040404_04040400]
vslidedown_vi_micro v3 : SimdMisc :  D=[00000000_00000000_00000000_00000000]
```

In addition, the slide instruction whose offset is an immediate value can statically filter out redundant uops during decoding to improve performance.  For example, lines 3551, 3552, 3554, 3555, 3556, and 3558 in the picture above.

**Affects version**
Develop branch. Commit [a8f88abfb12f772f496aad2b9866476f7ebb575c](https://github.com/gem5/gem5/tree/a8f88abfb12f772f496aad2b9866476f7ebb575c).

**gem5 Modifications**
#1712

**To Reproduce**
[Here](https://github.com/user-attachments/files/17349445/slide.tar.gz) is a simple test. Use the `slide` elf file and `se.py` script in `gz` package file to simulate. And dump the Exec trace.

**Terminal Output**
```shell

gem5 version DEVELOP-FOR-24.1
gem5 compiled Oct 12 2024 16:38:05
gem5 started Oct 12 2024 16:42:42
gem5 executing on n37-094-005, pid 126645
command line: ./build/RISCV/gem5.opt --debug-flags=Exec --debug-start=70593000 --debug-file=trace.out configs/bytedance/orca/se.py --cmd /data00/home/hongzhibo/share/benchmark/slide/slide

warn: Base 10 memory/cache size 1kB will be cast to base 2 size 1KiB.
warn: Base 10 memory/cache size 1kB will be cast to base 2 size 1KiB.
warn: Base 10 memory/cache size 8GB will be cast to base 2 size 8GiB.
warn: Base 10 memory/cache size 1MB will be cast to base 2 size 1MiB.
warn: Base 10 memory/cache size 64MB will be cast to base 2 size 64MiB.
warn: Base 10 memory/cache size 64kB will be cast to base 2 size 64KiB.
warn: Base 10 memory/cache size 64kB will be cast to base 2 size 64KiB.
warn: Base 10 memory/cache size 1MB will be cast to base 2 size 1MiB.
warn: system.cpu.dcache.prefetcher adopting orphan SimObject param 'prefetchers'
src/arch/riscv/isa.cc:278: info: RVV enabled, VLEN = 128 bits, ELEN = 64 bits
**** REAL SIMULATION ****
src/sim/syscall_emul.cc:97: warn: ignoring syscall set_robust_list(...)
      (further warnings will be suppressed)
src/sim/syscall_emul.cc:86: warn: ignoring syscall mprotect(...)
Exiting @ tick 78593000 because exiting with last active thread context
```

**Expected behavior**
You can see the above picture from the trace file. The results of the vslideup and vslidedown instructions do not meet the riscv vector specifications.

**Host Operating System**
Debian 12.

**Host ISA**
X86

**Compiler used**
gcc 14

5265325:
This PR fixed incorrect vector slide instructions mentioned in #1667.

And statically filter redundant uops for vslideup.vi/vslidedown.vi/vslide1up.vx /vsilde1down.vx/vfslide1up.vf/vfslide1down.vf instructions to improve performance.

Change-Id: I1ba36f216f0bb55372206a94b7f8108acc674f5a

### DIFF
--- a/src/arch/riscv/insts/vector.cc
+++ b/src/arch/riscv/insts/vector.cc
@@ -192,9 +192,11 @@ std::string VectorSlideMicroInst::generateDisassembly(Addr pc,
     std::stringstream ss;
     ss << mnemonic << ' ' << registerName(destRegIdx(0)) <<  ", ";
     if (machInst.funct3 == 0x3) {
-      ss  << registerName(srcRegIdx(0)) << ", " << machInst.vecimm;
+      ss  << registerName(srcRegIdx(0)) << ", "
+        << registerName(srcRegIdx(1)) << ", " << machInst.vecimm;
     } else {
-      ss  << registerName(srcRegIdx(1)) << ", " << registerName(srcRegIdx(0));
+      ss  << registerName(srcRegIdx(1)) << ", "
+        << registerName(srcRegIdx(2)) << ", " << registerName(srcRegIdx(0));
     }
     if (machInst.vm == 0) ss << ", v0.t";
     return ss.str();
@@ -860,10 +862,12 @@ VCpyVsMicroInst::generateDisassembly(Addr pc,
 
 VPinVdMicroInst::VPinVdMicroInst(ExtMachInst _machInst, uint32_t _microIdx,
                                  uint32_t _numVdPins, uint32_t _elen,
-                                 uint32_t _vlen, bool _hasVdOffset)
+                                 uint32_t _vlen, bool _hasVdOffset,
+                                 bool _copyVs, uint32_t _vsIdx)
     : VectorArithMicroInst("vpinvd_v_micro", _machInst, SimdMiscOp, 0,
                            _microIdx, _elen, _vlen)
     , hasVdOffset(_hasVdOffset)
+    , copyVs(_copyVs)
 {
     setRegIdxArrays(
         reinterpret_cast<RegIdArrayPtr>(
@@ -882,6 +886,10 @@ VPinVdMicroInst::VPinVdMicroInst(ExtMachInst _machInst, uint32_t _microIdx,
     RegId Vd = destRegIdx(0);
     Vd.setNumPinnedWrites(_numVdPins);
     setDestRegIdx(0, Vd);
+
+    if (copyVs) {
+        setSrcRegIdx(_numSrcRegs++, vecRegClass[_vsIdx]);
+    }
 }
 
 Fault
@@ -901,7 +909,7 @@ VPinVdMicroInst::execute(ExecContext* xc, trace::InstRecord* traceData) const
     // tail/mask policy: both undisturbed if one is, 1s if none
     vreg_t& vd = *(vreg_t *)xc->getWritableRegOperand(this, 0);
     if (!machInst.vtype8.vta || (!machInst.vm && !machInst.vtype8.vma)
-                             || hasVdOffset) {
+                            || hasVdOffset) {
         vreg_t old_vd;
         xc->getRegOperand(this, 0, &old_vd);
         vd = old_vd;
@@ -910,7 +918,14 @@ VPinVdMicroInst::execute(ExecContext* xc, trace::InstRecord* traceData) const
     }
 
     if (traceData) {
-        traceData->setData(vecRegClass, &vd);
+        traceData->setData(vecRegClass, xc->getWritableRegOperand(this, 0));
+    }
+
+    if (copyVs) {
+        vreg_t& vs = *(vreg_t *)xc->getWritableRegOperand(this, 1);
+        vreg_t old_vs;
+        xc->getRegOperand(this, 1, &old_vs);
+        vs = old_vs;
     }
 
     return NoFault;
--- a/src/arch/riscv/isa/decoder.isa
+++ b/src/arch/riscv/isa/decoder.isa
@@ -642,43 +642,50 @@ decode QUADRANT default Unknown::unknown() {
                         }}, inst_flags=SimdUnitStrideLoadOp);
                         format VlSegOp {
                             0x01: vlseg2e8_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 2)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 2)) &&
                                     i < this->microVl) {
                                     Vd_ub[i] = Mem_vc.as<uint8_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x02: vlseg3e8_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 3)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 3)) &&
                                     i < this->microVl) {
                                     Vd_ub[i] = Mem_vc.as<uint8_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x03: vlseg4e8_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 4)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 4)) &&
                                     i < this->microVl) {
                                     Vd_ub[i] = Mem_vc.as<uint8_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x04: vlseg5e8_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 5)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 5)) &&
                                     i < this->microVl) {
                                     Vd_ub[i] = Mem_vc.as<uint8_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x05: vlseg6e8_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 6)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 6)) &&
                                     i < this->microVl) {
                                     Vd_ub[i] = Mem_vc.as<uint8_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x06: vlseg7e8_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 7)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 7)) &&
                                     i < this->microVl) {
                                     Vd_ub[i] = Mem_vc.as<uint8_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x07: vlseg8e8_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 8)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 8)) &&
                                     i < this->microVl) {
                                     Vd_ub[i] = Mem_vc.as<uint8_t>()[i];
                                 }
@@ -736,43 +743,50 @@ decode QUADRANT default Unknown::unknown() {
                         }}, inst_flags=SimdUnitStrideLoadOp);
                         format VlSegOp {
                             0x01: vlseg2e16_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 2)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 2)) &&
                                     i < this->microVl) {
                                     Vd_uh[i] = Mem_vc.as<uint16_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x02: vlseg3e16_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 3)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 3)) &&
                                     i < this->microVl) {
                                     Vd_uh[i] = Mem_vc.as<uint16_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x03: vlseg4e16_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 4)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 4)) &&
                                     i < this->microVl) {
                                     Vd_uh[i] = Mem_vc.as<uint16_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x04: vlseg5e16_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 5)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 5)) &&
                                     i < this->microVl) {
                                     Vd_uh[i] = Mem_vc.as<uint16_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x05: vlseg6e16_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 6)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 6)) &&
                                     i < this->microVl) {
                                     Vd_uh[i] = Mem_vc.as<uint16_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x06: vlseg7e16_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 7)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 7)) &&
                                     i < this->microVl) {
                                     Vd_uh[i] = Mem_vc.as<uint16_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x07: vlseg8e16_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 8)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 8)) &&
                                     i < this->microVl) {
                                     Vd_uh[i] = Mem_vc.as<uint16_t>()[i];
                                 }
@@ -827,43 +841,50 @@ decode QUADRANT default Unknown::unknown() {
                         }}, inst_flags=SimdUnitStrideLoadOp);
                         format VlSegOp {
                             0x01: vlseg2e32_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 2)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 2)) &&
                                     i < this->microVl) {
                                     Vd_uw[i] = Mem_vc.as<uint32_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x02: vlseg3e32_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 3)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 3)) &&
                                     i < this->microVl) {
                                     Vd_uw[i] = Mem_vc.as<uint32_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x03: vlseg4e32_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 4)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 4)) &&
                                     i < this->microVl) {
                                     Vd_uw[i] = Mem_vc.as<uint32_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x04: vlseg5e32_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 5)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 5)) &&
                                     i < this->microVl) {
                                     Vd_uw[i] = Mem_vc.as<uint32_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x05: vlseg6e32_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 6)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 6)) &&
                                     i < this->microVl) {
                                     Vd_uw[i] = Mem_vc.as<uint32_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x06: vlseg7e32_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 7)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 7)) &&
                                     i < this->microVl) {
                                     Vd_uw[i] = Mem_vc.as<uint32_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x07: vlseg8e32_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 8)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 8)) &&
                                     i < this->microVl) {
                                     Vd_uw[i] = Mem_vc.as<uint32_t>()[i];
                                 }
@@ -918,43 +939,50 @@ decode QUADRANT default Unknown::unknown() {
                         }}, inst_flags=SimdUnitStrideLoadOp);
                         format VlSegOp {
                             0x01: vlseg2e64_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 2)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 2)) &&
                                     i < this->microVl) {
                                     Vd_ud[i] = Mem_vc.as<uint64_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x02: vlseg3e64_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 3)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 3)) &&
                                     i < this->microVl) {
                                     Vd_ud[i] = Mem_vc.as<uint64_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x03: vlseg4e64_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 4)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 4)) &&
                                     i < this->microVl) {
                                     Vd_ud[i] = Mem_vc.as<uint64_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x04: vlseg5e64_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 5)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 5)) &&
                                     i < this->microVl) {
                                     Vd_ud[i] = Mem_vc.as<uint64_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x05: vlseg6e64_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 6)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 6)) &&
                                     i < this->microVl) {
                                     Vd_ud[i] = Mem_vc.as<uint64_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x06: vlseg7e64_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 7)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 7)) &&
                                     i < this->microVl) {
                                     Vd_ud[i] = Mem_vc.as<uint64_t>()[i];
                                 }
                             }}, inst_flags=SimdUnitStrideSegmentedLoadOp);
                             0x07: vlseg8e64_v({{
-                                if ((machInst.vm || elem_mask_vseg(v0, ei + (field * micro_elems), 8)) &&
+                                if ((machInst.vm || elem_mask_vseg(v0,
+                                    ei + (field * micro_elems), 8)) &&
                                     i < this->microVl) {
                                     Vd_ud[i] = Mem_vc.as<uint64_t>()[i];
                                 }
@@ -1194,13 +1222,16 @@ decode QUADRANT default Unknown::unknown() {
                 format ROp {
                     0x0c: decode RS2 {
                         0x00: clz({{
-                            Rd = (machInst.rv_type == RV32) ? clz32(Rs1) : clz64(Rs1);
+                            Rd = (machInst.rv_type == RV32)
+                                ? clz32(Rs1) : clz64(Rs1);
                         }});
                         0x01: ctz({{
-                            Rd = (machInst.rv_type == RV32) ? ctz32(Rs1) : ctz64(Rs1);
+                            Rd = (machInst.rv_type == RV32)
+                                ? ctz32(Rs1) : ctz64(Rs1);
                         }});
                         0x02: cpop({{
-                            Rd = (machInst.rv_type == RV32) ? popCount(Rs1<31:0>) : popCount(Rs1);
+                            Rd = (machInst.rv_type == RV32)
+                                ? popCount(Rs1<31:0>) : popCount(Rs1);
                         }});
                         0x04: sext_b({{
                             Rd = sext<8>(Rs1_ub);
@@ -1346,10 +1377,12 @@ decode QUADRANT default Unknown::unknown() {
                     0x1: decode FS3 {
                         0x0: slliw({{
                             Rd_sd = Rs1_sw << imm;
-                        }}, imm_type = uint64_t, imm_code = {{ imm = SHAMT5; }});
+                        }}, imm_type = uint64_t,
+                        imm_code = {{ imm = SHAMT5; }});
                         0x1: slli_uw({{
                             Rd = ((uint64_t)(Rs1_uw)) << imm;
-                        }}, imm_type = uint64_t, imm_code = {{ imm = SHAMT6; }});
+                        }}, imm_type = uint64_t,
+                        imm_code = {{ imm = SHAMT6; }});
                         0xc: decode FS2 {
                             0x0: clzw({{
                                 Rd = clz32(Rs1);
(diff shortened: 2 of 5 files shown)


powerjg:
@saul44203 could you take a look at this? I think you're our resident RVV expert :)

src/arch/riscv/insts/vector.cc
```diff
@@ -881,6 +882,9 @@ VPinVdMicroInst::VPinVdMicroInst(ExtMachInst _machInst, uint32_t _microIdx,
     RegId Vd = destRegIdx(0);
     Vd.setNumPinnedWrites(_numVdPins);
     setDestRegIdx(0, Vd);
+    if (_hasTempVd) {
+        setDestRegIdx(_numDestRegs++, vecRegClass[VecMemInternalReg0 + _microIdx]);
+    }
```
saul44203, line 884: I wanted to get to this myself, but I think we can address this here.

The idea was to use an internal register as a dummy destination for the pin micro to create a data dependency with the actual micro instructions that write to the destination register vd. This is to be sure that the mask/tail policy (i.e. keep unchanged or filled with 1s) in the pinned vd is always done before anything is written. The `setDestRegIdx` was missing, so it was actually not creating any dependency at all, but I think we should always create this dependency, so there's no need to do it under a flag.

We should check that all pinned instructions have this register as source too, but this can be done in a separate pr maybe.

src/arch/riscv/insts/vector.cc
```diff
@@ -898,18 +902,20 @@ VPinVdMicroInst::execute(ExecContext* xc, trace::InstRecord* traceData) const
...
-    vreg_t& vd = *(vreg_t *)xc->getWritableRegOperand(this, 0);
-    if (!machInst.vtype8.vta || (!machInst.vm && !machInst.vtype8.vma)
-                             || hasVdOffset) {
-        vreg_t old_vd;
-        xc->getRegOperand(this, 0, &old_vd);
-        vd = old_vd;
-    } else {
-        vd.set(0xff);
-    }
+    for (int i = 0; i < _numDestRegs; i++) {
```
saul44203: There's no need to apply the policy to the internal reg, so I think we can remove the for loop.

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -1490,9 +1504,18 @@ def VectorSlideBase(name, Name, category, code, flags, macro_construtor,
         num_src_regs += 1
     set_src_reg_idx += setSrcWrapper(src2_reg_id)
     num_src_regs += 1
+    set_src_reg_idx += setSrcWrapper(src3_reg_id)
+    num_src_regs += 1
+    if is_vi:
+        set_src_reg_idx += """
+            oldDstIdx = _numSrcRegs;
+            setSrcRegIdx(_numSrcRegs++, destRegIdxArr[0]);
```
saul44203: I think only slideup instructions should have the old register as source without the tail/mask wrapper (as it always needed), but the rest should use it.

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2353,19 +2367,72 @@ template<typename ElemType>
         this->microops.push_back(microop);
     }
 
-    for (uint32_t i = 0; i < ceil((float) this->vl / micro_vlmax); i++) {
-        microop = new VPinVdMicroInst(machInst, i, num_microops-i, elen, vlen,
-                                      false);
+    %(offset_code)s;
+
+    int micro_idx = 0;
+    int microVlmax = vtype_VLMAX(machInst.vtype8, vlen, true);
```
saul44203: Minor thing. You already have `micro_vlmax` here.

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2379,6 +2446,61 @@ template<typename ElemType>
...
+
+    if (micro_vl == 0) {
+        microop = new VectorNopMicroInst(_machInst);
+        this->microops.push_back(microop);
+    }
+
+    %(offset_code)s;
+
+    int micro_idx = 0;
+    int microVlmax = vtype_VLMAX(machInst.vtype8, vlen, true);
```
saul44203: Same here.

src/arch/riscv/isa/decoder.isa
```diff
@@ -3693,67 +3693,101 @@ decode QUADRANT default Unknown::unknown() {
...
+                                Vd_vu[vdOffset] = copyBackFromVs3 ? Vs3_vu[i] : 0;
+                            }
+                        }
+                        for (; vdOffset < microVl; vdOffset++) {
+                            if (this->vm || elem_mask(v0, vdOffset + vdWindowLeft)) {
+                                Vd_vu[vdOffset] = 0;
+                            }
+                        }
+                    } else if (copyFrontFromVs3) {
+                        int frontLen = vs3WindowRightSlide - vdWindowLeft;
```
saul44203: I think this code (and all the similar ones) has some issues. Giving you an example where it fails:

Just changing your example uploaded in #1667 from `vl=64` to `vl=63`:
```
vslidedown_vi v0, v4, 17
vslidedown_vi_micro v0, v5, 17 : SimdMisc :  D=[02020202_02020202_02020202_02020203]
vslidedown_vi_micro v1, v6, 17 : SimdMisc :  D=[03030303_03030303_03030303_03030300]
vslidedown_vi_micro v2, v6, 17 : SimdMisc :  D=[04040404_04040404_04040404_04040000]
vslidedown_vi_micro v3, v7, 17 : SimdMisc :  D=[00000000_00000000_00000000_000000ff]
```
where it should be:
```
vslidedown_vi v0, v4, 17
vslidedown_vi_micro v0, v5, 17 : SimdMisc :  D=[02020202_02020202_02020202_02020203]
vslidedown_vi_micro v1, v6, 17 : SimdMisc :  D=[03030303_03030303_03030303_03030304]
vslidedown_vi_micro v2, v6, 17 : SimdMisc :  D=[04040404_04040404_04040404_04040400]
vslidedown_vi_micro v3, v7, 17 : SimdMisc :  D=[00000000_00000000_00000000_000000ff]
```

Main issue here is the checks that you do with `vl` in the for loops, as this is provoking to not read certain values. I think you can remove them, as we don't really need to check if the reading side surpasses `vl`. This should solve some things but there might be other issues with this 2 source approach (which should be represented in the assembly, as the second source is hidden) and the calculation of the branch which will be taken.

saul44203 (changes requested):

erin-le:
@5265325 Hello, thank you for your contribution. Could you give an update on the status of this PR on your end?

powerjg:
@saul44203 when you get a chance can you re-review this?

src/arch/riscv/isa/formats/vector_arith.isa
saul44203: Hi @5265325. Here I meant the use of `tailMaskCondSetSrcWrapper` to create the dependency only when the register tail/mask policy calls for it, but we should only do this for instructions that do not require pinning, If they have pinning the `VPinVdMicroInst` already handles the copying according to the tail/mask policy (that can be overridden with the `hasVdOffset` flag), and we should not put the destination register as source in those cases, as it will cause gem5 to crash when building the dependency graph. 

This is the case with `vslideup.vx`, where the slide micro will be a consumer of the pinning micro as a cause of this code, but then it will also try to put itself as a producer of the same physical register.

So, in summary, I think we have to remove this for the pinning cases, and the others, since they copy directly, should do this using `tailMaskCondSetSrcWrapper`.

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2361,23 +2387,104 @@ template<typename ElemType>
...
-        microop = new VCpyVsMicroInst(machInst, i, machInst.vs2, elen, vlen);
-        microop->setFlag(IsDelayedCommit);
+    %(offset_code)s;
+
+    uint32_t vs2_end = machInst.vs2 + num_microops - 1;
+    uint32_t vd_end = machInst.vd + num_microops - 1;
+    bool copyVs = (machInst.vs2 >= machInst.vd && machInst.vs2 <= vd_end)
+        || (vs2_end >= machInst.vd && vs2_end <= vd_end);
+
+    if (copyVs) {
```
saul44203, line 2397: Minor thing, but this isn't necessary in the instructions that don't pin, right? Even if the source and destination registers are the same arch reg, they will rename to different physical registers.

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2361,23 +2387,104 @@ template<typename ElemType>
...
+        for (int i = 0; i < ceil((float) this->vl / micro_vlmax); i++) {
+            microop = new VCpyVsMicroInst(machInst, i, machInst.vs2, elen, vlen);
+            microop->setFlag(IsDelayedCommit);
+            this->microops.push_back(microop);
+        }
+    }
```
saul44203: Minor thing, but this isn't necessary in the instructions that don't pin, right? Even if the source and destination registers are the same arch reg, they will rename to different physical registers.

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -1490,9 +1497,22 @@ def VectorSlideBase(name, Name, category, code, flags, macro_construtor,
...
+    set_src_reg_idx += setSrcWrapper(src3_reg_id)
+    num_src_regs += 1
+
+    is_slideup = "up" in name
+    if is_slideup:
+        set_src_reg_idx += """
+            oldDstIdx = _numSrcRegs;
+            setSrcRegIdx(_numSrcRegs++, destRegIdxArr[0]);
+        """
+        num_src_regs += 1
```
saul44203: Hi @5265325. Here I meant the use of `tailMaskCondSetSrcWrapper` to create the dependency only when the register tail/mask policy calls for it, but we should only do this for instructions that do not require pinning, If they have pinning the `VPinVdMicroInst` already handles the copying according to the tail/mask policy (that can be overridden with the `hasVdOffset` flag), and we should not put the destination register as source in those cases, as it will cause gem5 to crash when building the dependency graph. 

This is the case with `vslideup.vx`, where the slide micro will be a consumer of the pinning micro as a cause of this code, but then it will also try to put itself as a producer of the same physical register.

So, in summary, I think we have to remove this for the pinning cases, and the others, since they copy directly, should do this using `tailMaskCondSetSrcWrapper`.

saul44203 (changes requested):

src/arch/riscv/isa/templates/vector_arith.isa
5265325: I just want to reduce the redundant uop, so that the uop split closer to the real cpu. Excess uop can affect performance results.

saul44203: Yeah, I see that, but in this case I meant that we do not need the `VCpyVsMicroInst` at all.

5265325: There is still a vslideup scene that needs VCpyVsMicroInst. 
```shell
vsetivli zero, 8, e64, m2, ta, ma
vslideup.vi v1, v0, 2
```
In gem5's renaming logic, the destination register of the first uop is v1, but the source register of the second uop uses the result of the first uop, which is incorrect. So we need to copy v1 to a temporary register before uop split.

saul44203: I see that would be the case. However, one thing that we missed is that per spec the source and destination register groups of the [slideup instructions cannot overlap](https://github.com/riscvarchive/riscv-v-spec/blob/master/v-spec.adoc#1631-vector-slideup-instructions) , so this should not happen (and maybe we should add this encodings as explicit illegal instruction faults in the future). Even further, this should mean that we do not even need it in the pinning cases after all.

For slidedown it does not say this but spike gives illegal instruction when there's overlap and LMUL>1. So, if we follow this, we only need it to cover the LMUL=1 case when pinning, but otherwise it's not necessary.

5265325: Now the vslide instruction that overlaps will raise an illegal exception.

src/arch/riscv/isa/formats/vector_arith.isa
5265325: > VPinVdMicroInst

I'm not quite sure what you mean. It seems to me that VPinVdMicroInst is responsible for deciding whether to keep the old value or write 0xff. And slide micro is responsible for deciding whether to write the new value. Both are necessary.

saul44203: Yes, it decides whether or not to copy based on the current tail/mask policy. But the thing is that if you use the `VPinVdMicroInst` there's no need to use the `copyOldVd` method (with the `oldDstIdx` as src reg and such) on the actual slide micro instruction. Doing so will produce the crash that I mentioned with `vslideup.vx`.

src/arch/riscv/isa/templates/vector_arith.isa
saul44203: Sorry for reopening this but what about slidedown instructions? I think we should allow overlap at least when LMUL=1 like spike does. We could also make them work for LMUL>1 if you want, as spec does not say otherwise.

5265325: Yes. I check overlap only when LMUL > 1.

5265325:
@saul44203 The rvv-matmul and rvv-reduce test failed in CI. However, they passed in my local testing. when you get a chance could you take a look?

Here are my test commands.
```shell
./build/RISCV/gem5.opt  ./configs/example/gem5_library/riscv-rvv-example.py rvv-matmul --vlen=256
./build/RISCV/gem5.opt  ./configs/example/gem5_library/riscv-rvv-example.py rvv-reduce --vlen=16384
```

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2452,6 +2589,18 @@ Fault
...
+        uint32_t num_microops = vtype_regs_per_group(vtype);
+        uint32_t vs2_end = machInst.vs2 + num_microops - 1;
+        uint32_t vd_end = machInst.vd + num_microops - 1;
+        bool overlap = (machInst.vs2 >= machInst.vd && machInst.vs2 <= vd_end)
+            || (vs2_end >= machInst.vd && vs2_end <= vd_end);
+        if (overlap) {
+            return std::make_shared<IllegalInstFault>("Regs overlap is not allowed", machInst);
+        }
+    }
+
```
saul44203: Sorry for repeating myself, but this will allow slide up instructions to overlap when LMUL<=1, and spec does not allow this. Only slide down instructions should be allowed to overlap as we talked previously.

Maybe we can filter here with a flag or something?

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2361,23 +2376,80 @@ template<typename ElemType>
...
+    if (micro_vl == 0) {
+        microop = new VectorNopMicroInst(_machInst);
         this->microops.push_back(microop);
+    }
 
-        microop = new VPinVdMicroInst(machInst, i, num_microops-i, elen, vlen,
-                                      false);
+    for (int i = 0; i < ceil((float) this->vl / micro_vlmax); i++) {
+        microop = new VPinVdMicroInst(machInst, i, std::max(num_microops-i-1, 1), elen, vlen,
+                                      true);
```
saul44203: Here we will also need the `VCpyVsMicroInst` when there's overlap to avoid the dependency graph crash.

saul44203 (changes requested):

saul44203:
> @saul44203 The rvv-matmul and rvv-reduce test failed in CI. However, they passed in my local testing. when you get a chance could you take a look?
> 
> Here are my test commands.
> 
> ```shell
> ./build/RISCV/gem5.opt  ./configs/example/gem5_library/riscv-rvv-example.py rvv-matmul --vlen=256
> ./build/RISCV/gem5.opt  ./configs/example/gem5_library/riscv-rvv-example.py rvv-reduce --vlen=16384
> ```

This I don't understand. They are working well on my side too.

Harshil2107:
I have also tested running the CI tests locally in the docker container and they pass. I will take a look into the failing tests but it might be a resource issue with the runners not having enough resources to run the tests.

src/arch/riscv/isa/templates/vector_arith.isa
5265325: Now only num_microops>2 cases require pin micro. VCpyVsMicroInst was moved to VPinVdMicroInst to reduce the number of uops.

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2384,23 +2406,86 @@ template<typename ElemType>
...
-                                      false);
+    uint32_t vs2_end = machInst.vs2 + num_microops - 1;
+    uint32_t vd_end = machInst.vd + num_microops - 1;
+    bool overlap = (machInst.vs2 >= machInst.vd && machInst.vs2 <= vd_end)
+        || (vs2_end >= machInst.vd && vs2_end <= vd_end);
+    bool need_pin = overlap || num_microops > 2;
+
+    for (int i = 0; need_pin && i < ceil((float) this->vl / micro_vlmax); i++) {
+        microop = new VPinVdMicroInst(machInst, i, std::max(num_microops-i-1, 1), elen, vlen,
+                                      true, overlap, machInst.vs2 + i);
```
saul44203: I think here we should use `hasVdOffset=false`.

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2475,6 +2627,19 @@ Fault
     if (machInst.vill)
         return std::make_shared<IllegalInstFault>("VILL is set", machInst);
 
+    if (slideUp || vtype_vlmul(machInst.vtype8) > 1) {
```
saul44203: I think this should be `> 0` if using the `vlmul` (not `LMUL`) since this will sign extend to 1 when `LMUL=2`.

saul44203 (changes requested):
Only a couple of small things. Otherwise lgtm.

saul44203 (approved):

powerjg (dismissed):

powerjg:
Quick update on this... I just downloaded this PR and tried to run the commands. When I did, they both resulted in "fail"

```
./build/RISCV/gem5.opt  ./configs/example/gem5_library/riscv-rvv-example.py rvv-matmul --vlen=256
./build/RISCV/gem5.opt  ./configs/example/gem5_library/riscv-rvv-example.py rvv-reduce --vlen=16384
```

```
command line: ./build/RISCV/gem5.opt ./configs/example/gem5_library/riscv-rvv-example.py rvv-reduce --vlen=16384

info: Using default config
warn: Setting the full_system parameter via the Simulator constructor is deprecated and will be removed in future releases of gem5. The board determines if the simulation is full system or not via it's `is_fullsystem` method.
Beginning simulation!
src/arch/riscv/isa.cc:286: info: RVV enabled, VLEN = 16384 bits, ELEN = 64 bits
src/sim/simulate.cc:200: info: Entering event queue @ 0.  Starting simulation...
src/sim/syscall_emul.cc:97: warn: ignoring syscall set_robust_list(...)
      (further warnings will be suppressed)
src/sim/syscall_emul.hh:1121: warn: readlink() called on '/proc/self/exe' may yield unexpected results in various settings.
      Returning '/home/jlp/.cache/gem5/rvv-reduce'
src/sim/syscall_emul.cc:86: warn: ignoring syscall mprotect(...)
/home/jlp/.cache/gem5/rvv-reduce: fail
```

src/arch/riscv/insts/vector.cc
```diff
@@ -881,6 +883,11 @@ VPinVdMicroInst::VPinVdMicroInst(ExtMachInst _machInst, uint32_t _microIdx,
     RegId Vd = destRegIdx(0);
     Vd.setNumPinnedWrites(_numVdPins);
     setDestRegIdx(0, Vd);
+    setDestRegIdx(_numDestRegs++, vecRegClass[VecMemInternalReg0 + _microIdx]);
```
saul44203: This interferes with `vcpyvs_v_micro` when both are renaming the same internal reg. Since the pin micro comes after the copy one, the actual instruction's micro that has it as a source will get the renamed reg from this one instead of the copy, which is incorrect.

@powerjg I think we didn't see this happen before because one of the rebases included #1955 and since both `rvv-reduce` and `rvv-matmul` have a reduction those will now fail with the added pinning.

I think we should remove this so the tests pass normally, as this was added as way of enforcing a dependency between pinning and regular micro instructions, but it was preemptive and we could revise this if needed on another pr.

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -1474,29 +1474,43 @@ def VectorSlideBase(name, Name, category, code, flags, macro_construtor,
...
+    set_src_reg_idx = ""
+
+    src2_reg_id = "vecRegClass[(_copyVs ? VecMemInternalReg0 : _machInst.vs2) + vs2Idx]"
+    src3_reg_id = "vecRegClass[(_copyVs ? VecMemInternalReg0 : _machInst.vs2) + vs3Idx]"
     src1_ireg_id = "intRegClass[_machInst.rs1]"
     src1_freg_id = "floatRegClass[_machInst.rs1]"
 
+    # vtmp0 as dummy src reg to create dependency with pin vd micro
+    pin_reg_id = "vecRegClass[VecMemInternalReg0 + vdIdx]"
+
```
saul44203: This too.

saul44203 (changes requested):

src/arch/riscv/insts/vector.cc
powerjg: Sounds like a good solution to me.

erin-le:
@5265325 Hello, could you respond to the comments on this PR when you get a chance? Thank you!

saul44203 (approved):
LGTM

powerjg:
@erin-le can you re-run the tests above to make sure they pass now?

erin-le (changes requested):
Running `riscv-rvv-example.py` with `rvv-reduce --vlen=16384` and `rvv-matmul --vlen=256` both pass. However, there are numerous lines in this PR that are longer than 79 characters.

5265325:
> Running `riscv-rvv-example.py` with `rvv-reduce --vlen=16384` and `rvv-matmul --vlen=256` both pass. However, there are numerous lines in this PR that are longer than 79 characters.

Now the 80 character limit has been satisfied.

src/arch/riscv/insts/vector.cc
```diff
@@ -192,9 +192,9 @@ std::string VectorSlideMicroInst::generateDisassembly(Addr pc,
     std::stringstream ss;
     ss << mnemonic << ' ' << registerName(destRegIdx(0)) <<  ", ";
     if (machInst.funct3 == 0x3) {
-      ss  << registerName(srcRegIdx(0)) << ", " << machInst.vecimm;
+      ss  << registerName(srcRegIdx(0)) << ", " << registerName(srcRegIdx(1)) << ", " << machInst.vecimm;
```
powerjg: Over 80 char

src/arch/riscv/insts/vector.cc
```diff
@@ -192,9 +192,9 @@ std::string VectorSlideMicroInst::generateDisassembly(Addr pc,
...
     } else {
-      ss  << registerName(srcRegIdx(1)) << ", " << registerName(srcRegIdx(0));
+      ss  << registerName(srcRegIdx(1)) << ", " << registerName(srcRegIdx(2)) << ", " << registerName(srcRegIdx(0));
```
powerjg: Over 80 char

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -36,10 +36,10 @@ let {{
     def setSrcWrapper(srcRegId):
         return "setSrcRegIdx(_numSrcRegs++, " + srcRegId + ");\n"
     def tailMaskCondSetSrcWrapper(setSrcRegCode):
-        return f'''
-        if (!_machInst.vtype8.vta || (!_machInst.vm && !_machInst.vtype8.vma))
-            {setSrcRegCode}
-        '''
+        return "if (!_machInst.vtype8.vta || (!_machInst.vm && !_machInst.vtype8.vma)) {\n" \
```
powerjg: Over 80 char

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -1472,29 +1472,31 @@ def VectorSlideBase(name, Name, category, code, flags, macro_construtor,
...
+        'offset_code': offset_code},
         flags
     )
     inst_name, inst_suffix = name.split("_", maxsplit=1)
     dest_reg_id = "vecRegClass[_machInst.vd + vdIdx]"
-    src2_reg_id = "vecRegClass[VecMemInternalReg0 + vs2Idx]"
+
+    set_src_reg_idx = ""
+
+    src2_reg_id = "vecRegClass[(_copyVs ? VecMemInternalReg0 : _machInst.vs2) + vs2Idx]"
```
powerjg: Over 80 char

powerjg (changes requested):
Still has lines >80 char. I know it's annoying..

src/arch/riscv/insts/vector.cc
powerjg: This still is over 80 characters. Maybe there is a force-push/rebase issue here?

powerjg:
@BobbyRBruce we should get this on the staging branch if at all possible. Apologies for missing this on the milestone.

BobbyRBruce:
> @BobbyRBruce we should get this on the staging branch if at all possible. Apologies for missing this on the milestone.

Ack. Will cherry-pick to `staging` when merged into `develop`

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2400,22 +2414,30 @@ template<typename ElemType>
...
-        this->microops.push_back(microop);
+    uint32_t vs2_end = machInst.vs2 + num_microops - 1;
+    uint32_t vd_end = machInst.vd + num_microops - 1;
+    bool overlap = (machInst.vs2 >= machInst.vd && machInst.vs2 <= vd_end)
+        || (vs2_end >= machInst.vd && vs2_end <= vd_end);
+    bool need_pin = overlap || num_microops > 2;
 
-        microop = new VPinVdMicroInst(machInst, i, i+1, elen, vlen, true);
+    for (int i = 0; need_pin && i < ceil((float) this->vl/micro_vlmax); i++) {
+        microop = new VPinVdMicroInst(machInst, i, std::max(i, 1), elen, vlen, true,
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2400,22 +2414,30 @@ template<typename ElemType>
...
+                                      overlap, machInst.vs2 + i);
         microop->setFlag(IsDelayedCommit);
         this->microops.push_back(microop);
     }
 
-    // Todo static filter out useless uop
     int micro_idx = 0;
+    uint32_t vsIdxMax = (1 << std::max<int64_t>(0, vtype_vlmul(_machInst.vtype8))) - 1;
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2400,22 +2414,30 @@ template<typename ElemType>
...
+
     for (int i = 0; i < num_microops && micro_vl > 0; ++i) {
-        for (int j = 0; j <= i; ++j) {
+        for (int j = 0; j <= std::max(i - 1, 0); ++j) {
+            uint32_t vs2Idx = j;
+            uint32_t vs3Idx = (j == vsIdxMax) ? j : j + 1;
+
             microop = new %(class_name)sMicro<ElemType>(
-                _machInst, micro_vl, micro_idx++, i, j, _elen, _vlen);
+                _machInst, micro_vl, micro_idx++, i, vs2Idx, vs3Idx, _elen, _vlen,
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2449,23 +2471,86 @@ template<typename ElemType>
         this->microops.push_back(microop);
     }
 
-    for (uint32_t i = 0; i < ceil((float) this->vl / micro_vlmax); i++) {
-        microop = new VCpyVsMicroInst(machInst, i, machInst.vs2, elen, vlen);
-        microop->setFlag(IsDelayedCommit);
+    %(offset_code)s;
+
+    int micro_idx = 0;
+    uint32_t vsIdxMax = (1 << std::max<int64_t>(0, vtype_vlmul(_machInst.vtype8))) - 1;
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2449,23 +2471,86 @@ template<typename ElemType>
...
+    for (int i = 0; i < num_microops && micro_vl > 0; ++i) {
+        for (int j = 0; j <= std::max(i - 1, 0); ++j) {
+            uint32_t vs2Idx = j;
+            uint32_t vs3Idx = (j == vsIdxMax) ? j : j + 1;
+
+            int vdWindowLeft = micro_vlmax * i;
+            int vdWindowRight = micro_vlmax * (i + 1);
+            int vs2WindowLeft = micro_vlmax * j + offset;
+            int vs3WindowLeft = micro_vlmax * (j + 1) + offset;
+            bool copyAll = vs3WindowLeft >= vdWindowLeft && vs3WindowLeft < vdWindowRight;
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2449,23 +2471,86 @@ template<typename ElemType>
...
+            bool copyBack = j == 0 && vs2WindowLeft >= vdWindowLeft && vs2WindowLeft < vdWindowRight;
+            // Static filter out useless uop
+            if (copyAll || copyBack || j == std::max(i - 1, 0)) {
+                microop = new %(class_name)sMicro<ElemType>(
+                    _machInst, micro_vl, micro_idx++, i, vs2Idx, vs3Idx, _elen, _vlen,
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2449,23 +2471,86 @@ template<typename ElemType>
...
-        microop = new VPinVdMicroInst(machInst, i, num_microops-i, elen, vlen,
-                                      false);
+    uint32_t vs2_end = machInst.vs2 + num_microops - 1;
+    uint32_t vd_end = machInst.vd + num_microops - 1;
+    bool overlap = (machInst.vs2 >= machInst.vd && machInst.vs2 <= vd_end)
+        || (vs2_end >= machInst.vd && vs2_end <= vd_end);
+    bool need_pin = overlap || num_microops > 2;
+
+    for (int i = 0; need_pin && i < ceil((float) this->vl / micro_vlmax); i++) {
+        microop = new VPinVdMicroInst(machInst, i, std::max(num_microops-i-1, 1), elen, vlen,
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2449,23 +2471,86 @@ template<typename ElemType>
...
+                                      false, overlap, machInst.vs2 + i);
         microop->setFlag(IsDelayedCommit);
         this->microops.push_back(microop);
     }
 
-    // Todo static filter out useless uop
     int micro_idx = 0;
+    uint32_t vsIdxMax = (1 << std::max<int64_t>(0, vtype_vlmul(_machInst.vtype8))) - 1;
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2449,23 +2471,86 @@ template<typename ElemType>
...
+
     for (int i = 0; i < num_microops && micro_vl > 0; ++i) {
-        for (int j = i; j < num_microops; ++j) {
+        for (int j = i; j < std::max(num_microops-1, 1) || i == j; ++j) {
+            uint32_t vs2Idx = j;
+            uint32_t vs3Idx = (j == vsIdxMax) ? j : j + 1;
+
             microop = new %(class_name)sMicro<ElemType>(
-                _machInst, micro_vl, micro_idx++, i, j, _elen, _vlen);
+                _machInst, micro_vl, micro_idx++, i, vs2Idx, vs3Idx, _elen, _vlen,
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2479,6 +2564,64 @@ template<typename ElemType>
...
+
+    if (micro_vl == 0) {
+        microop = new VectorNopMicroInst(_machInst);
+        this->microops.push_back(microop);
+    }
+
+    %(offset_code)s;
+
+    int micro_idx = 0;
+    uint32_t vsIdxMax = (1 << std::max<int64_t>(0, vtype_vlmul(_machInst.vtype8))) - 1;
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2479,6 +2564,64 @@ template<typename ElemType>
...
+    for (int i = 0; i < num_microops && micro_vl > 0; ++i) {
+        for (int j = i; j < std::max(num_microops-1, 1) || i == j; ++j) {
+            uint32_t vs2Idx = j;
+            uint32_t vs3Idx = (j == vsIdxMax) ? j : j + 1;
+
+            int vdWindowLeft = micro_vlmax * i;
+            int vdWindowRight = micro_vlmax * (i + 1);
+            int vs3WindowLeftSlide = micro_vlmax * vs3Idx - offset;
+            int vs3WindowRightSlide = micro_vlmax * (vs3Idx + 1) - offset;
+            bool copyFrontFromVs2 = vs3WindowLeftSlide > vdWindowLeft && vs3WindowLeftSlide <= vdWindowRight;
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2479,6 +2564,64 @@ template<typename ElemType>
...
+            bool lastUopForVd = j >= (num_microops - 2);
+            bool copyFrontFromVs3 = lastUopForVd && vs3WindowRightSlide > vdWindowLeft && vs3WindowRightSlide <= vdWindowRight;
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2479,6 +2564,64 @@ template<typename ElemType>
...
+            bool needZeroTail = lastUopForVd && vs3WindowRightSlide <= vdWindowLeft;
+            // Static filter out useless uop
+            if (copyFrontFromVs2 || copyFrontFromVs3 || needZeroTail) {
+                microop = new %(class_name)sMicro<ElemType>(
+                    _machInst, micro_vl, micro_idx++, i, vs2Idx, vs3Idx, _elen, _vlen,
+                    false, lastUopForVd, i == (num_microops - 1) || micro_vl < micro_vlmax);
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2479,6 +2564,64 @@ template<typename ElemType>
...
+            int vs3WindowLeftSlide = micro_vlmax * vs3Idx - offset;
+            int vs3WindowRightSlide = micro_vlmax * (vs3Idx + 1) - offset;
+            bool copyFrontFromVs2 = vs3WindowLeftSlide > vdWindowLeft && vs3WindowLeftSlide <= vdWindowRight;
+            bool lastUopForVd = j >= (num_microops - 2);
+            bool copyFrontFromVs3 = lastUopForVd && vs3WindowRightSlide > vdWindowLeft && vs3WindowRightSlide <= vdWindowRight;
+            bool needZeroTail = lastUopForVd && vs3WindowRightSlide <= vdWindowLeft;
+            // Static filter out useless uop
+            if (copyFrontFromVs2 || copyFrontFromVs3 || needZeroTail) {
+                microop = new %(class_name)sMicro<ElemType>(
+                    _machInst, micro_vl, micro_idx++, i, vs2Idx, vs3Idx, _elen, _vlen,
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2479,6 +2564,64 @@ template<typename ElemType>
...
+            uint32_t vs3Idx = (j == vsIdxMax) ? j : j + 1;
+
+            int vdWindowLeft = micro_vlmax * i;
+            int vdWindowRight = micro_vlmax * (i + 1);
+            int vs3WindowLeftSlide = micro_vlmax * vs3Idx - offset;
+            int vs3WindowRightSlide = micro_vlmax * (vs3Idx + 1) - offset;
+            bool copyFrontFromVs2 = vs3WindowLeftSlide > vdWindowLeft && vs3WindowLeftSlide <= vdWindowRight;
+            bool lastUopForVd = j >= (num_microops - 2);
+            bool copyFrontFromVs3 = lastUopForVd && vs3WindowRightSlide > vdWindowLeft && vs3WindowRightSlide <= vdWindowRight;
+            bool needZeroTail = lastUopForVd && vs3WindowRightSlide <= vdWindowLeft;
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2540,6 +2692,19 @@ Fault
...
         return std::make_shared<IllegalInstFault>("VILL is set", machInst);
 
+    uint32_t num_microops = vtype_regs_per_group(vtype);
+    if (slideUp || num_microops > 1) {
+        uint32_t vs2_end = machInst.vs2 + num_microops - 1;
+        uint32_t vd_end = machInst.vd + num_microops - 1;
+        bool overlap = (machInst.vs2 >= machInst.vd && machInst.vs2 <= vd_end)
+            || (vs2_end >= machInst.vd && vs2_end <= vd_end);
+        if (overlap && slideUp) {
+            return std::make_shared<IllegalInstFault>("Regs overlap is not allowed for vslideup", machInst);
```
erin-le: Over 80 chars

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2540,6 +2692,19 @@ Fault
...
+        } else if (overlap && !slideUp) {
+            return std::make_shared<IllegalInstFault>("Regs overlap is not allowed for vslidedown when lmul > 1", machInst);
```
erin-le: Over 80 chars

erin-le (changes requested):
There are a few more lines that are longer than 80 characters. Thanks for your patience with the review process!

5265325:
@powerjg @erin-le Thanks. I have used precommit to check the character length. The length requirement should now be met.

BobbyRBruce:
You had a merge conflict in "src/arch/riscv/isa/templates/vector_arith.isa" due to changes introduced in PR #1387 (it was just merged yesterday morning so a new addition). I took the liberty of fixing the conflict by simply preserving the changes introduced in #1387 along side yours but _please_ double check to ensure I haven't done anything stupid here.

5265325:
> You had a merge conflict in "src/arch/riscv/isa/templates/vector_arith.isa" due to changes introduced in PR #1387 (it was just merged yesterday morning so a new addition). I took the liberty of fixing the conflict by simply preserving the changes introduced in #1387 along side yours but _please_ double check to ensure I haven't done anything stupid here.

LGTM

powerjg (approved):
Thanks for the updates!

PR #1718, opened 2024-10-26
mem, tests: SimObject tester for replacement policies

erin-le:
This PR adds a SimObject that tests the FIFO replacement policy. It initializes several entries, evicts an entry, and checks that the evicted entry was the first to be created, indicating that the replacement policy behaves correctly.

I am still working on getting this change to work with several replacement policies.

### DIFF
--- a/src/mem/cache/replacement_policies/sim_object_unit_tester.cc
+++ b/src/mem/cache/replacement_policies/sim_object_unit_tester.cc
@@ -1,0 +1,99 @@
+#include "mem/cache/replacement_policies/sim_object_unit_tester.hh"
+
+#include <iostream>
+
+#include "base/trace.hh"
+#include "mem/cache/replacement_policies/fifo_rp.hh"
+#include "mem/cache/replacement_policies/replaceable_entry.hh"
+
+namespace gem5
+{
+namespace replacement_policy
+{
+
+SimObjectUnitTester::SimObjectUnitTester(
+  const SimObjectUnitTesterParams& params):
+  SimObject(params),
+  nextEvent([this](){ processNextEvent(); }, name() + "nextEvent"),
+  replacement_policy(params.replacement_policy),
+  numEntries(params.num_entries)
+{}
+
+void
+SimObjectUnitTester::processNextEvent()
+{
+  std::cout << "tick: " << curTick() <<
+  ", Hello from SimObjectUnitTester::processNextEvent!" << std::endl;
+
+  // create a new entry and put it in candidates
+  ReplaceableEntry* temp = new ReplaceableEntry();
+  temp->replacementData = replacement_policy->instantiateEntry();
+  replacement_policy->reset(temp->replacementData); // record insertion tick
+  // std::cout<< "Inserted tick: " <<
+  //std::static_pointer_cast<FIFO::FIFOReplData>(temp->replacementData)
+  //->tickInserted <<std::endl;
+  candidates.push_back(temp);
+
+  if (numEntries > 0){
+    schedule(nextEvent, curTick() + 500);
+    numEntries--;
+  }
+  else {
+    // FIFO* tmp;
+    bool correct = checkCorrectness(/*tmp*/);
+    // free memory
+    for (const auto& candidate : candidates) {
+      delete candidate;
+    }
+    if (correct){
+      exit(0);
+    }
+    exit(1);
+  }
+}
+
+bool
+SimObjectUnitTester::checkCorrectness(/*FIFO* */){
+  ReplaceableEntry* victim = replacement_policy->getVictim(candidates);
+  if (std::static_pointer_cast<FIFO::FIFOReplData>(
+    victim ->replacementData)->tickInserted != 1){
+
+    std::cout<<"tick: "<<std::static_pointer_cast<FIFO::FIFOReplData>(
+      victim ->replacementData)->tickInserted <<std::endl;
+    return false;
+  }
+  return true;
+}
+
+  // template <>
+  // bool
+  // SimObjectUnitTester::checkCorrectness<FIFO>(){
+  //   ReplaceableEntry* victim = replacement_policy->getVictim(candidates);
+  //   if (std::static_pointer_cast<FIFO::FIFOReplData>(victim
+  //->replacementData)->tickInserted != 1){
+  //     std::cout<<"tick: "<<std::static_pointer_cast<FIFO::FIFOReplData>
+  //(victim ->replacementData)->tickInserted <<std::endl;
+  //     return false;
+  //   }
+  //   return true;
+  // }
+
+  // template <>
+  // bool
+  // SimObjectUnitTester::checkCorrectness<Base>(){
+  //   return false;
+  // }
+
+
+void
+SimObjectUnitTester::startup()
+{
+  panic_if(curTick() != 0, "startup() called at a tick other than 0");
+  panic_if(nextEvent.scheduled(),
+  "nextEvent scheduled before startup() called! ");
+  schedule(nextEvent, curTick() + 500);
+}
+
+} //namespace replacement_policy
+
+} // namespace gem5
--- a/src/mem/cache/replacement_policies/sim_object_unit_tester.cc
+++ b/src/mem/cache/replacement_policies/sim_object_unit_tester.cc
@@ -15,8 +15,9 @@ SimObjectUnitTester::SimObjectUnitTester(
   const SimObjectUnitTesterParams& params):
   SimObject(params),
   nextEvent([this](){ processNextEvent(); }, name() + "nextEvent"),
-  replacement_policy(params.replacement_policy),
-  numEntries(params.num_entries)
+  replacementPolicy(params.replacement_policy),
+  numEntries(params.num_entries),
+  replacementPolicyName(params.name)
 {}
 
 void
@@ -27,8 +28,8 @@ SimObjectUnitTester::processNextEvent()
 
   // create a new entry and put it in candidates
   ReplaceableEntry* temp = new ReplaceableEntry();
-  temp->replacementData = replacement_policy->instantiateEntry();
-  replacement_policy->reset(temp->replacementData); // record insertion tick
+  temp->replacementData = replacementPolicy->instantiateEntry();
+  replacementPolicy->reset(temp->replacementData); // record insertion tick
   // std::cout<< "Inserted tick: " <<
   //std::static_pointer_cast<FIFO::FIFOReplData>(temp->replacementData)
   //->tickInserted <<std::endl;
@@ -39,12 +40,9 @@ SimObjectUnitTester::processNextEvent()
     numEntries--;
   }
   else {
-    // FIFO* tmp;
-    bool correct = checkCorrectness(/*tmp*/);
+    bool correct = checkCorrectness(replacementPolicy);
     // free memory
-    for (const auto& candidate : candidates) {
-      delete candidate;
-    }
+    freeCandidates();
     if (correct){
       exit(0);
     }
@@ -52,9 +50,24 @@ SimObjectUnitTester::processNextEvent()
   }
 }
 
+void
+SimObjectUnitTester::freeCandidates(void){
+  for (const auto& candidate : candidates) {
+    delete candidate;
+  }
+}
+
+// bool
+// SimObjectUnitTester::checkCorrectness(void){
+//   return false;
+// }
+
+
 bool
-SimObjectUnitTester::checkCorrectness(/*FIFO* */){
-  ReplaceableEntry* victim = replacement_policy->getVictim(candidates);
+SimObjectUnitTester::checkCorrectness(FIFO*){
+// SimObjectUnitTester::checkCorrectnessFIFO(){
+
+  ReplaceableEntry* victim = replacementPolicy->getVictim(candidates);
   if (std::static_pointer_cast<FIFO::FIFOReplData>(
     victim ->replacementData)->tickInserted != 1){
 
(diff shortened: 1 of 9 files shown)


src/mem/cache/replacement_policies/sim_object_unit_tester.cc
```diff
@@ -0,0 +1,193 @@
...
+  if (evictedLastTick != expectedEvictedLastTick){
+    return false;
+  }
+  return true;
+}
+
+
+
+bool
+SimObjectUnitTester::checkCorrectness(Base*){
```
erin-le, line 108: There's definietly a better way of calling different functions based on the replacement policy type; I'm just not sure what that would be. I would really appreciate help/feedback on this @BobbyRBruce!

src/mem/cache/replacement_policies/sim_object_unit_tester.cc
```diff
@@ -0,0 +1,193 @@
+#include "mem/cache/replacement_policies/sim_object_unit_tester.hh"
```
odanrc, line 1: Missing copyright header

odanrc, line 1: The coding style does not match gem5´s: https://www.gem5.org/documentation/general_docs/development/coding_style/

It should be fixed accordingly. Things I notice from a quick glance: the number of spaces in an indentation should be 4; an else should be in the same line as the closing brace of the if; the functions´ opening braces should be on the following line

src/mem/cache/replacement_policies/sim_object_unit_tester.cc
```diff
@@ -0,0 +1,193 @@
...
+{}
+
+void
+SimObjectUnitTester::processNextEvent()
+{
+  std::cout << "tick: " << curTick() <<
+  ", Hello from SimObjectUnitTester::processNextEvent!" << std::endl;
+
+  // create a new entry and put it in candidates
+  ReplaceableEntry* temp = new ReplaceableEntry();
```
odanrc, line 30: It is better to use smart pointers instead of new/delete

src/mem/cache/replacement_policies/sim_object_unit_tester.cc
```diff
@@ -0,0 +1,193 @@
...
+  nextEvent([this](){ processNextEvent(); }, name() + "nextEvent"),
+  replacementPolicy(params.replacement_policy),
+  numEntries(params.num_entries),
+  replacementPolicyName(params.name)
+{}
+
+void
+SimObjectUnitTester::processNextEvent()
+{
+  std::cout << "tick: " << curTick() <<
```
odanrc, line 26: We typically use the DEBUG macros to print debug information. That way when you are done debugging you will be able to consciously decide which DEBUGs to keep, and remove the ones that didn´t help much

src/mem/cache/replacement_policies/sim_object_unit_tester.cc
odanrc: dynamic_cast, maybe?

src/mem/cache/replacement_policies/sim_object_unit_tester.hh
```diff
@@ -0,0 +1,63 @@
+#ifndef __MEM_CACHE_REPLACEMENT_POLICIES_SIM_OBJECT_UNIT_TESTER_HH__
```
odanrc, line 1: Ditto copyright header

src/mem/cache/replacement_policies/sim_object_unit_tester.hh
```diff
@@ -0,0 +1,63 @@
...
+//     virtual bool checkCorrectness() override;
+// };
+
+
+
+} //namespace replacement_policy
+
+} //namespace gem5
+
+#endif // __SIM_OBJECT_UNIT_TESTER_SIM_OBJECT_UNIT_TESTER_HH__
```
odanrc, line 63: The commented-out macro name does not match the one defined earlier

odanrc (changes requested):
I am definitely in favor of unit tests for RPs, but I don´t think this PR is going the right direction.

Here the proposal is to add unit tests, but for that I would expect a lot of small tests using the same testing framework we use elsewhere: gtest. We already have tests that verify that the replacement policies behave as expected (I don´t recall where they are, but I recall reviewing the tests, and that they are not in C++), so I don´t think we desperately need to add unit tests to RPs - i.e., we have the time to do it the gtest-way.

When we were using gerrit I managed to make SimObjects unit-testable (https://gem5-review.googlesource.com/c/public/gem5/+/44111/13), but things got stuck in review limbo and I stopped pushing for it. I think that if we are to unit-test the RP, then we should revive those changes to use gtest instead.

BobbyRBruce:
> I am definitely in favor of unit tests for RPs, but I don´t think this PR is going the right direction.
> 
> Here the proposal is to add unit tests, but for that I would expect a lot of small tests using the same testing framework we use elsewhere: gtest. We already have tests that verify that the replacement policies behave as expected (I don´t recall where they are, but I recall reviewing the tests, and that they are not in C++), so I don´t think we desperately need to add unit tests to RPs - i.e., we have the time to do it the gtest-way.
> 
> When we were using gerrit I managed to make SimObjects unit-testable (https://gem5-review.googlesource.com/c/public/gem5/+/44111/13), but things got stuck in review limbo and I stopped pushing for it. I think that if we are to unit-test the RP, then we should revive those changes to use gtest instead.

The issue here is it's very hard to decouple SimObjects from the rest of gem5 in a way in which you can verify their behavior in a gem5 simulation. If there's anyone who's found a way to do this it'd be a great contribution but I've found the SimObjects are just too embedded in the logic of the gem5 simulator for real unit-tests at this level.

src/mem/cache/replacement_policies/sim_object_unit_tester.cc
BobbyRBruce: I jsut want to point out that this is a draft and not officially ready for review. Erin submitted it to get some feedback on the general direction of this development task.

So anything here not quite right might just a placeholder or some debugging code.

src/mem/cache/replacement_policies/sim_object_unit_tester.cc
BobbyRBruce: I think the way to do this is have the `ReplacementPolicyTester` be a base class with an abstract `checkCorrectness` function. Then for each replacement policy you have a subclass which implements this function (e.g., you'd have a `FIFOReplacementPolicyTester` with its own `checkCorrectness` function that overrides that in `ReplacementPolicyTester`). Everything general to all replacement policies in the ReplacementPolicyTester and anything specific to each implementation in their respective subclasses.

mattsinc:
> > I am definitely in favor of unit tests for RPs, but I don´t think this PR is going the right direction.
> > Here the proposal is to add unit tests, but for that I would expect a lot of small tests using the same testing framework we use elsewhere: gtest. We already have tests that verify that the replacement policies behave as expected (I don´t recall where they are, but I recall reviewing the tests, and that they are not in C++), so I don´t think we desperately need to add unit tests to RPs - i.e., we have the time to do it the gtest-way.
> > When we were using gerrit I managed to make SimObjects unit-testable (https://gem5-review.googlesource.com/c/public/gem5/+/44111/13), but things got stuck in review limbo and I stopped pushing for it. I think that if we are to unit-test the RP, then we should revive those changes to use gtest instead.
> 
> The issue here is it's very hard to decouple SimObjects from the rest of gem5 in a way in which you can verify their behavior in a gem5 simulation. If there's anyone who's found a way to do this it'd be a great contribution but I've found the SimObjects are just too embedded in the logic of the gem5 simulator for real unit-tests at this level.

@odanrc :  the tests are all here: https://github.com/gem5/gem5/tree/stable/tests/gem5/replacement_policies

I thought those tests were alrady functioning as unit tests, but maybe I'm not quite understanding @BobbyRBruce 's reply about the difference ...

src/mem/cache/replacement_policies/sim_object_unit_tester.cc
```diff
@@ -0,0 +1,193 @@
...
+    numEntries--;
+  }
+  else {
+    bool correct = checkCorrectness(replacementPolicy);
+    // free memory
+    freeCandidates();
+    if (correct){
+      exit(0);
+    }
+    exit(1);
```
BobbyRBruce, line 49: Id remove this. I think the best thing to do here is just to have the SimObject print something is to throw a `warn` if there's unexpected behavior(e.g. "Replacement policy 'FOFI''s victim was not first in!'). Then, in our testlib, we just check the output and report an report a test failure if there's anything printed we weren't expected.

BobbyRBruce (commented):
I think most of this belongs in "src/test_objects"

BobbyRBruce:
> > > I am definitely in favor of unit tests for RPs, but I don´t think this PR is going the right direction.
> > > Here the proposal is to add unit tests, but for that I would expect a lot of small tests using the same testing framework we use elsewhere: gtest. We already have tests that verify that the replacement policies behave as expected (I don´t recall where they are, but I recall reviewing the tests, and that they are not in C++), so I don´t think we desperately need to add unit tests to RPs - i.e., we have the time to do it the gtest-way.
> > > When we were using gerrit I managed to make SimObjects unit-testable (https://gem5-review.googlesource.com/c/public/gem5/+/44111/13), but things got stuck in review limbo and I stopped pushing for it. I think that if we are to unit-test the RP, then we should revive those changes to use gtest instead.
> > 
> > 
> > The issue here is it's very hard to decouple SimObjects from the rest of gem5 in a way in which you can verify their behavior in a gem5 simulation. If there's anyone who's found a way to do this it'd be a great contribution but I've found the SimObjects are just too embedded in the logic of the gem5 simulator for real unit-tests at this level.
> 
> @odanrc : the tests are all here: https://github.com/gem5/gem5/tree/stable/tests/gem5/replacement_policies
> 
> I thought those tests were alrady functioning as unit tests, but maybe I'm not quite understanding @BobbyRBruce 's reply about the difference ...

My argument was more as to why we're not using gtest or something more like what you'd expect from unit tests. It's because the simobjects can't be compiled and run independently of the simulator.

What Erin is doing here, which differs from that in "tests/gem5/replacement_policies", is that with it you can specify a simulation with just  the replacement policy's tester and nothing else. No memory, processor, ISA, anything. Just replacement policy tester. Then you run it to test the replacement policy in  and of itself, independent of anything else in the project. Not only is this a lot less overhead (not requiring simulation of the rest of the system, the booting of the OS or even an input binary to work), it means that the tests are only testing that one SimObject. If the test fails, then it's undoubtedly a failure with the SimObject. We don't need to spend any time determining that.

One of the big problems with most of our tests is that they require complex interactions between other parts of gem5 to work. Not only is this costly, it also makes the cause of failures harder to find. Right now, the GPU tests can fail when there's an issue with X86, or the cache hierarchy they use, or one of the many SimObjects required to stitch together a system capable of executing something. When failures like this happen, it becomes a pain to track down the cause. With this approach, if the FIFO replacement policy test fails, it's because of an issue with the FIFO replacement policy.

Your tests are great, but I'd argue instead of being unit tests, they are integration tests. They test the behavior of the replacement policy when integrated into a larger system. It's a coarser level of granularity. It's definitely needed, but both approaches are.

mattsinc:
> > > > I am definitely in favor of unit tests for RPs, but I don´t think this PR is going the right direction.
> > > > Here the proposal is to add unit tests, but for that I would expect a lot of small tests using the same testing framework we use elsewhere: gtest. We already have tests that verify that the replacement policies behave as expected (I don´t recall where they are, but I recall reviewing the tests, and that they are not in C++), so I don´t think we desperately need to add unit tests to RPs - i.e., we have the time to do it the gtest-way.
> > > > When we were using gerrit I managed to make SimObjects unit-testable (https://gem5-review.googlesource.com/c/public/gem5/+/44111/13), but things got stuck in review limbo and I stopped pushing for it. I think that if we are to unit-test the RP, then we should revive those changes to use gtest instead.
> > > 
> > > 
> > > The issue here is it's very hard to decouple SimObjects from the rest of gem5 in a way in which you can verify their behavior in a gem5 simulation. If there's anyone who's found a way to do this it'd be a great contribution but I've found the SimObjects are just too embedded in the logic of the gem5 simulator for real unit-tests at this level.
> > 
> > 
> > @odanrc : the tests are all here: https://github.com/gem5/gem5/tree/stable/tests/gem5/replacement_policies
> > I thought those tests were alrady functioning as unit tests, but maybe I'm not quite understanding @BobbyRBruce 's reply about the difference ...
> 
> My argument was more as to why we're not using gtest or something more like what you'd expect from unit tests. It's because the simobjects can't be compiled and run independently of the simulator.
> 
> What Erin is doing here, which differs from that in "tests/gem5/replacement_policies", is that with it you can specify a simulation with just the replacement policy's tester and nothing else. No memory, processor, ISA, anything. Just replacement policy tester. Then you run it to test the replacement policy in and of itself, independent of anything else in the project. Not only is this a lot less overhead (not requiring simulation of the rest of the system, the booting of the OS or even an input binary to work), it means that the tests are only testing that one SimObject. If the test fails, then it's undoubtedly a failure with the SimObject. We don't need to spend any time determining that.
> 
> One of the big problems with most of our tests is that they require complex interactions between other parts of gem5 to work. Not only is this costly, it also makes the cause of failures harder to find. Right now, the GPU tests can fail when there's an issue with X86, or the cache hierarchy they use, or one of the many SimObjects required to stitch together a system capable of executing something. When failures like this happen, it becomes a pain to track down the cause. With this approach, if the FIFO replacement policy test fails, it's because of an issue with the FIFO replacement policy.
> 
> Your tests are great, but I'd argue instead of being unit tests, they are integration tests. They test the behavior of the replacement policy when integrated into a larger system. It's a coarser level of granularity. It's definitely needed, but both approaches are.

I wasn't really trying to start an argument, sorry.  Sure unit vs. integration makes more sense now.  In terms of those replacement policy tests I think they just connect the traffic generator to the caches so it eliminates much of the "other" potential sources of error you were listing.  But I can see how something even simpler is useful.  Hopefully @erin-le can use what is there to avoid duplicating too much work though.

odanrc:
I am going to double down here: the POC for unit testing of SimObjects with gtest was already written in the past. I had even written a couple of unit tests for existing components as examples:

https://gem5-review.googlesource.com/c/public/gem5/+/44111/13
https://gem5-review.googlesource.com/c/public/gem5/+/44113/13
https://gem5-review.googlesource.com/c/public/gem5/+/44114/13

They all behave as unit tests: if they fail, and the dependencies' tests do not, then it is clear that the respective Bloom filter broke.

The unit-testable-SimObject change got stuck only 10 commits away from being merged into upstream. Now that we have more review power I think we can and should push for it again. I will try to bring a couple of those commits to GitHub this week.

I understand that there is the decoupling concern, but I don´t think the current solution will solve this concern given that the UnitTester class is itself a SimObject.

BobbyRBruce:
> I am going to double down here: the POC for unit testing of SimObjects with gtest was already written in the past. I had even written a couple of unit tests for existing components as examples:
> 
> https://gem5-review.googlesource.com/c/public/gem5/+/44111/13 https://gem5-review.googlesource.com/c/public/gem5/+/44113/13 https://gem5-review.googlesource.com/c/public/gem5/+/44114/13
> 
> They all behave as unit tests: if they fail, and the dependencies' tests do not, then it is clear that the respective Bloom filter broke.
> 
> The unit-testable-SimObject change got stuck only 10 commits away from being merged into upstream. Now that we have more review power I think we can and should push for it again. I will try to bring a couple of those commits to GitHub this week.
> 
> I understand that there is the decoupling concern, but I don´t think the current solution will solve this concern given that the UnitTester class is itself a SimObject.

By all means, please get this change in. It's not that I don't want unit testable SImObjects it's that, right now, we can't do it easily. If you can get that change in I'd more than happily reject tests like this in favor of the GTest approach.

BobbyRBruce:
I've moved this Gerrit relation-chain here: #1726. I'm happy to help get this in.

@erin-le You're free to check this out and play with it to see if it works for you.

erin-le:
Closing since https://github.com/gem5/gem5/pull/1786 was opened.

PR #1732, opened 2024-10-29
arch-x86: Update MTRR defType register

powerjg:


powerjg:
Note: This doesn't work with KVM, but I don't think this is a supported MSR.

abmerop (commented):
Seems fine. Why is it needed though? What is the behavior if you do not do this?

powerjg:
> Seems fine. Why is it needed though? What is the behavior if you do not do this?

Good question. Without this change, you get the error "MTRRs disabled by BIOS." As far as I can tell, we support MTRRs, and this was not the intended behavior. 

[Link to kernel code](https://elixir.bootlin.com/linux/v6.8/source/arch/x86/kernel/cpu/mtrr/mtrr.c#L600). That line was being triggered.

The *problem* this is solving is when we use a "legacy pmem" device (i.e., type 12), the device is set to uncachable mode and all accesses are to uncachable pages (as marked in the page table). With this change, we can now use the (processor) caches when accessing pmem devices.

(As a side note, as a bit of hack to avoid implementing all of the CXL.io protocol, we are using the legacy pmem device when working with CXL remote memory.)

abmerop (approved):

PR #1734, opened 2024-10-30
base: Make BaseGdbRegCache::data() non constant

ylldummy:
The method is defined as const but the caller will actually modify the content of the structure directly with the pointer in BaseRemoteGDB::cmdRegW. The member access in the const method are actually treated as const and will cause error if we use reinterpret_cast instead.

Remove the const tag to align the expectation of the virtual method.

### DIFF
--- a/src/arch/arm/remote_gdb.hh
+++ b/src/arch/arm/remote_gdb.hh
@@ -77,7 +77,7 @@ class RemoteGDB : public BaseRemoteGDB
           uint32_t fpscr;
         } r;
       public:
-        char *data() const override { return (char *)&r; }
+        char *data() override { return (char *)&r; }
         size_t size() const override { return sizeof(r); }
         void getRegs(ThreadContext*) override;
         void setRegs(ThreadContext*) const override;
@@ -103,7 +103,7 @@ class RemoteGDB : public BaseRemoteGDB
           uint32_t fpcr;
         } r;
       public:
-        char *data() const override { return (char *)&r; }
+        char *data() override { return (char *)&r; }
         size_t size() const override { return sizeof(r); }
         void getRegs(ThreadContext*) override;
         void setRegs(ThreadContext*) const override;
--- a/src/arch/mips/remote_gdb.hh
+++ b/src/arch/mips/remote_gdb.hh
@@ -66,7 +66,7 @@ class RemoteGDB : public BaseRemoteGDB
             uint32_t fir;
         } r;
       public:
-        char *data() const { return (char *)&r; }
+        char *data() { return (char *)&r; }
         size_t size() const { return sizeof(r); }
         void getRegs(ThreadContext*);
         void setRegs(ThreadContext*) const;
--- a/src/arch/power/remote_gdb.hh
+++ b/src/arch/power/remote_gdb.hh
@@ -68,7 +68,7 @@ class RemoteGDB : public BaseRemoteGDB
         } r;
 
       public:
-        char *data() const { return (char *)&r; }
+        char *data() { return (char *)&r; }
         size_t size() const { return sizeof(r); }
         void getRegs(ThreadContext*);
         void setRegs(ThreadContext*) const;
@@ -97,7 +97,7 @@ class RemoteGDB : public BaseRemoteGDB
         } r;
 
       public:
-        char *data() const { return (char *)&r; }
+        char *data() { return (char *)&r; }
         size_t size() const { return sizeof(r); }
         void getRegs(ThreadContext*);
         void setRegs(ThreadContext*) const;
--- a/src/arch/riscv/remote_gdb.hh
+++ b/src/arch/riscv/remote_gdb.hh
@@ -137,7 +137,7 @@ class RemoteGDB : public BaseRemoteGDB
             uint32_t hip;
         } r;
       public:
-        char *data() const { return (char *)&r; }
+        char *data() { return (char *)&r; }
         size_t size() const { return sizeof(r); }
         void getRegs(ThreadContext*);
         void setRegs(ThreadContext*) const;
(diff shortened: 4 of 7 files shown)


powerjg (commented):
I don't follow this change. From what I understand, the `const` is saying that the function doesn't modify the data in the `RemoteGDB` class (which is true). It returns a non-const pointer, which the callee can then use to modify data. But this doesn't break the const-ness of the function. 

I don't see where we need a reinterpret cast or where there's an error. Could you correct me where I'm wrong or give some more information?

ylldummy:
Hi, thanks for the review!

I have a toy example here to demonstrate the error I got if I use reinterpret_cast instead: https://godbolt.org/z/8E9PsnTe6

(I did this in my child class to comply with coding style: https://google.github.io/styleguide/cppguide.html#Casting)

In the toy example, compiler might give the following error:

error: 'reinterpret_cast' from type 'const Foo::Bar*' to type 'char*' casts away qualifiers
   11 |         return reinterpret_cast<char*>(&bar);

My understanding of the error is that the method marked as const will require access to its member also as constants so potentially casting point to non-cast to caller of member isn't expected in a const function ?

powerjg:
I must be missing some important context. Is there a bug in gem5 that this is fixing? If so, can you describe this bug?

Otherwise, is this a problem for some proprietary extension? If so, can you describe why you need to change gem5 for the extension? I.e., what are you extending and why do you need to change it?

ylldummy:
Hi, sorry, I probably didn't explain the problem well. It has nothing to do with my extension and it's not fixing any existing bug.

The main issue is the interface design. i.e. BaseGdbRegCache::data is a const function but allows caller to modify its data through the returned pointer.

From the reinterpret_cast toy example, we can see that when accessing the member `bar` in the `data() const` method, it is treated as a constant type. When we try to cast its pointer to `char*` with `reinterpret_cast`, compiler will have error about discarding the const qualifier.

I'm not actually sure why compiler doesn't complain about the current casting though, maybe compiler thinks `const_cast` can make sense here, but `const_cast` has some pitfalls that can cause undefined behavior, although it's never the case in related use cases of the change. See Note section in [ref](https://en.cppreference.com/w/cpp/language/const_cast)

OTOH, as an example, std::vector::data() interface is declared as:

```
T* data();
const T* data() const;
```
ref: https://en.cppreference.com/w/cpp/container/vector/data

When the one with const qualifier is used, the const pointer is returned and caller is not expected to modify the content.

---

I'm not very familiar with C++ standard, but I tried to collect some related reference from standard [draft](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/n4928.pdf). Hope this might help.

```
7.5.2.3
If a declaration declares a member function or member function template of a class X, the expression this
is a prvalue of type “pointer to cv-qualifier-seq X” wherever X is the current class between the optional
cv-qualifier-seq and the end of the function-definition, member-declarator, or declarator.
```
From this statement, we can know `this` pointer in the const member function has const qualifier.

```
6.8.5.1.1
A const object is an object of type const T or a non-mutable subobject of a const object.
```
From this statement, the reference to the subobject from const typed `this` will also have `const` qualifier.

ylldummy:
To better demonstrate that const member function means "the access to the object is with const qualifier" instead of only "the function itself doesn't change the object", I made an example of how casting away const qualifier in the const member function can go wrong: https://godbolt.org/z/n188s3rfq.

That program can get SIGSEGV. (I think you can directly see the result from the output tab in the godbolt link)

powerjg:
So, are you saying there *is* a bug in gem5? Are you saying that currently, everyone is using the pointer from data as read-only so we don't run into a segfault?

If that's the case, then why don't we add const to the return type?

Here's where I'm coming from: We shouldn't make changes unless the change is needed. If this change doesn't have any effect, then I don't think it should be merged. If this change fixes a bug or enables a new use case, then we should definitely merge it!

Here's another question that may help me understand better: How did you find this issue? (You saw a seg fault, you ran a static analyzer, etc.?)

giactra:
Hi all, I don't have a strong opinion on this PR but here's my 2 cents:

I believe this overall makes sense. The issue @ylldummy is trying to solve is one of "bad" practice/software design. So there is probably no bug as of now, but basically the fact that we allow a const method to return a non const pointer opens up the following scenario:

Say I design my API (returnData) to accept a non-mutable object:

```
class MyObject
{
  public:
    char* getData() const { return data; }
  private:
    char* data;
}
char returnData(const MyObject& obj);
```

Someone implementing returnData might be able to modify my const object as getData effectively works as a backdoor:

```
char
returnData(const MyObject &obj)
{
    auto data = obj.getData();
    // I can modify MyObject (by mistake?) and this really breaks the
    // API assumption obj is non-modifiable within the returnData
    data[0] = ..
    return data[0];
}
```

This wouldn't happen if getData was marked as non-const as the following snippet wouldn't compile... A better design would be:

```
class MyObject
{
  public:
    char* getData() { return data; } -> to be used if we need to modify MyObject
    const char* getData() const { return data; } -> to be used in returnData
  private:
    char* data;
}
```

I believe constness can be misleading in C++ and while it is true you usually interpret a const method as saying "the object won't be modified by this method", it is also true that returning a non-const pointer to member can break constness somewhere down the line...

ylldummy:
Hi all,

> Are you saying that currently, everyone is using the pointer from data as read-only so we don't run into a segfault? If that's the case, then why don't we add const to the return type?

No, `BaseRemoteGDB::cmdRegW` will write via those pointers. However, that `cmdRegW` is non-const so it actually has a mutable reference to the object so it won't be in read-only range and the memory access won't raise segfault.

Also, as we can see that things still build after dropping const qualifier, all callers in the codebase have a non-const reference.

> If this change doesn't have any effect, then I don't think it should be merged.

As pointed out by Giacomo, this is trying to fix a interface design that can potentially allow incorrect access, instead of a real bug anywhere.

> Here's another question that may help me understand better: How did you find this issue? (You saw a seg fault, you ran a static analyzer, etc.?)

I originally noticed this issue because I used `reinterpret_cast` in my local experimental implementation of the interface, which was simply for compliance with the coding style as mentioned in another comment. `reinterpret_cast` has a stronger limitation that doesn't allow removing the const qualifier so compiler errors out.

If you prefer having this const-ness check available in current `data()` member function as well, I can also add that.

powerjg (approved):

odanrc (approved):

PR #1741, opened 2024-10-31
arch-riscv: Implement a unified L2 TLB

ncppd:
This commit adds an inclusive unified L2 TLB that services both ITLB & DTLB. I have followed a style similar to the ARM implementation of the L2 TLB.

Specifically:

- Since RISC-V TLB inherits from BaseTLB, I am taking advantage of the `BaseTLB *_nextLevel;` pointer in order to setup an L2 TLB
- The `_nextLevel` pointer is used from both ITLB & DTLB and points to the second level TLB 
- In order to handle inserts & lookups of the L2 TLB I took inspiration from the ARM L2 TLB implementation and adopted a similar approach
    - **Inserts:** I implemented a `multiInsert` function that inserts a new entry to the L1 TLB, checks if `nextLevel` is valid, and if yes also updates the L2 TLB (inclusive L2 TLB, with little changes it can be re-designed as exclusive)
    - **Lookups:** I implemented a `multiLookup` function that performs an L1 lookup and in case of a miss the `nextLevel` pointer is checked and if it is valid an L2 lookup is performed. If the L2 lookup results in a hit the L1 TLB is updated (again in an inclusive design).
- Standard `lookup` & `insert` functions remain unchanged
- I changed the function calls from `lookup` to `multiLookup` & `insert` to `multiInsert`  in oder to take advantage of the L2 TLB
- Finally, I updated the `RiscvTLB.py` script to correctly setup the L1 & L2 TLBs. In case and implementation does not require an L2 TLB, the `next_level = l2_shared` parameter can simply be omitted in order to remove the L2 TLB from the design.

Some notes:

- TLB flushes (sfence.vma) will be counted twice in this initial implementation (use locks to handle this behavior?)
- This implementation does not follow the usual L1 TLB->L2 TLB->PTW since both the DTLB & ITLB are directly connected to the PTW

For extra details you might be interested in reviewing an earlier [paper](https://cgi.di.uoa.gr/~vkarakos/papers/fpl20_configurable_tlb_riscv.pdf) of mine. I recently started using gem5 for my research and I intend to implement  features such as set-associative TLBs, the SVNAPOT extension etc.


rogerchang23424:
Do you have any reference? Thanks

ncppd:
Could you elaborate on that a little further? Are you referring to the notes or the L2 implementation?

rogerchang23424:
I mean could you leave any L2 TLB spec link in the commit message or in the code section? So that it's benefit for gem5 community

ncppd:
I updated the PR message with detailed info.

aarmejach:
Thanks for this contribution, will try to get someone to review it asap.

> This implementation does not follow the usual L1 TLB->L2 TLB->PTW since both the DTLB & ITLB are directly connected to the PTW

Could you elaborate a bit more on this point? what happens on L1TLB miss? can it lead to stale/wrong translations?

ncppd:
Hi @aarmejach, this implementation assumed that on an L1 TLB miss (1) the L2 TLB is looked-up (2), and if the L2 TLB also reports a miss (3), it informs the L1 TLB (4) to initiate a PTW (5). 

In common/real systems, on an L1 TLB miss (1) the L2 TLB lookup happens in parallel with the PTW (2), and if the L2 TLB hits the corresponding PTW is killed (3-i). Otherwise the PTW commences until a leaf PTE is found (3-ii) and the L1/L2 TLBs are updated.

I ran several benchmarks and the number of total (ITLB & DTLB) 4K PTWs reported is reduced -- meaning the TLB reach is larger due to the newly added L2 TLB. The L2 TLB reports 0 PTWs since only the DTLB/ITLB have ports connected to the PTW.

For any other clarification feel free to ask!

h0lyalg0rithm:
@ncppd I had implemented something similar in our fork of gem5.
The only difference in my change is  that in the `TLB::flushAll()` we also flush the entries in the l2 tlb

ncppd:
Hi @h0lyalg0rithm, in my implementation L2 TLB flushing is triggered in `TLB::demapPage()`, since `sfence.vma rs1, rs2` accepts two arguments it might not trigger a full TLB flush, but a flush based on VPN or/and ASID. Flushing all L2 TLB entries using `TLB::flushAll()` might hurt performance if `sfence.vma` targets single/group entries.

h0lyalg0rithm:
@ncppd Makes sense, with that approach.

h0lyalg0rithm:
Looks good to me. @ncppd Thanks for the PR

powerjg:
Thanks for the contribution! However, I am currently not for incorporating this change into gem5. The reason is that this change implements a very specific microarchitecture which is not part of the generic ISA.

My view on gem5's balance between architecture and microarchitecture is that we should have a full *architectural* implementation upstream. If it is in the spec, then there is no doubt that we should implement it in gem5. (We can also implement things that are out of the spec if it's an optional component, but it should be default off.)

For microarchitectural models, we should err on the side of *configurability* and *generality* and avoid committing upstream specific implementation details (when not in the spec).

To take this PR as an example, if the RISC-V spec defined a multi-level TLB structure (e.g., as done in Arm), then we should have this upstream. However, the spec doesn't specify this. The other way to modify this PR to meet these expectations would be to provide a *configurable* TLB structure and allow the user to choose 1-level, 2-level, or whatever TLB/page walk cache architecture they want. 

This said, I recognize that the current address translation microarchitecture in gem5 (and specifically RISC-V) is lacking and limiting the kinds of research and development we can do. It would be a wonderful contribution to improve the generic MMU models in gem5. I don't see why we can't have a much more flexible and robust model for the MMU that can be used across many of our ISAs.

**Potential modifications to this PR**

One way to modify this PR to address these concerns would be to make the L2 TLB *optional* and support the use case that someone doesn't want to have a multi-level TLB structure.

*A separate concern* that I have with the PR is I don't see how the ports for the page table walkers have been updated. I would have expected this code to need to be updated with the unified L2 TLB.

hnpl:
The code looks good to me, however this is a very specific way of implementing the 2-level TLB.

@ncppd Since this is a very specific microarchitecture detail, can you make this optional?

ncppd:
@powerjg 
> To take this PR as an example, if the RISC-V spec defined a multi-level TLB structure (e.g., as done in Arm), then we should have this upstream. However, the spec doesn't specify this. The other way to modify this PR to meet these expectations would be to provide a _configurable_ TLB structure and allow the user to choose 1-level, 2-level, or whatever TLB/page walk cache architecture they want.

Yes, I agree with you; configurability should be the best way to go. I implemented this simple (and hacky!) TLB hierarchy with respect to workloads that stress the MMU and may skew results due to L1 TLB thrashing, which could be alleviated with a larger L2 TLB.

> This said, I recognize that the current address translation microarchitecture in gem5 (and specifically RISC-V) is lacking and limiting the kinds of research and development we can do. It would be a wonderful contribution to improve the generic MMU models in gem5. I don't see why we can't have a much more flexible and robust model for the MMU that can be used across many of our ISAs.

I would be delighted to contribute to this; however, my free time is limited, so I may need to work on it at a slow pace.

@hnpl + @powerjg 
> One way to modify this PR to address these concerns would be to make the L2 TLB _optional_ and support the use case that someone doesn't want to have a multi-level TLB structure.

This should be straightforward to address. `RiscvMMU.py` can remain unchanged, and for those interested in using the L2 TLB, I could include a note either to the documentation or within `RiscvMMU.py`.

> _A separate concern_ that I have with the PR is I don't see how the ports for the page table walkers have been updated. I would have expected this code to need to be updated with the unified L2 TLB.

In my implementation, the L1 TLB remains responsible for sending requests to the PTW, rather than following the standard L1 → L2 → PTW flow, and that is why I have not updated the PTW ports.

powerjg:
> In my implementation, the L1 TLB remains responsible for sending requests to the PTW, rather than following the standard L1 → L2 → PTW flow, and that is why I have not updated the PTW ports.

Yeah, this seems kinda hacky. My gut says that the PTW would be part of the unified L2 TLB (or part of the PWC), not two separate state machines. I could be wrong though as I haven't studied real-world PTW designs in a number of years. 

We need to be careful what we upstream in gem5. We don't want to add a model that people start to use if that model is unrealistic. It's a (hard) tradeoff. *You* may know that the intention of the L2 TLB is to alleviate thrashing in the L1 and that it doesn't model "realistic" designs, but it's unlikely there will be a good way to communicate that to the 1000s of people that use gem5 in the future. Of course, the flip side is that the current single-level TLB design is also not realistic! So... It's a tough decision.

hnpl:
@ncppd Thanks for the explanation. I have another question: is it correct that this implementation does not update L1 TLB when there's a L2 TLB hit?

>  this implementation assumed that on an L1 TLB miss (1) the L2 TLB is looked-up (2), and if the L2 TLB also reports a miss (3), it informs the L1 TLB (4) to initiate a PTW (5).

The stats might skew towards having more L2 TLB accesses than it should because useful TLB entries might be stuck in L2 TLB.

ncppd:
Making the L2 TLB optional requires only a few edits in `RiscvTLB.py`, I plan to send a patch in the coming days.

ncppd:
@powerjg I have made the L2 TLB optional and added a comment in case someone wants to use an L2 TLB.

PR #1743, opened 2024-10-31
stdlib: Add interface to set binary in fs mode

powerjg:
This change adds a `set_binary_to_run` function in the `KernelDiskWorkload` mixin. This function allows you to pass a binary that will be executed after boot when booting in FS mode. With this change, you can write code something like below to run a binary from the host on the simulated guest.

```python
board.set_workload(obtain_resource("x86-ubuntu-24.04-boot-no-systemd"))
board.set_binary_to_run(BinaryResource("/bin/ls"), "-l")
```

This will copy the file `/bin/ls` from the host to the gem5 guest and execute it with the `-l` argument.

Note: This function should only be used for testing. **This is not intended for use in experiments as it will make the experiments very difficult to reproduce!**

Inspiration/code from Matt Poremba.
See https://github.com/gem5/gem5/pull/1636

### DIFF
--- a/src/python/gem5/components/boards/kernel_disk_workload.py
+++ b/src/python/gem5/components/boards/kernel_disk_workload.py
@@ -37,6 +37,7 @@ import m5
 from m5.util import warn
 
 from ...resources.resource import (
+    BinaryResource,
     BootloaderResource,
     CheckpointResource,
     DiskImageResource,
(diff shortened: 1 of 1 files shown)


powerjg:
cc: @abmerop, your code was so cool that I thought we should generalize it for all boards!

abmerop:
The readfile command is very powerful and underused IMO.  The next problem I ran into after this is you cannot use this for applications which have input files.  I am using a short term solution which is a copy of readfile that takes host path as a string from within the simulation and functions the same as readfile to copy in multiple files from the host: https://github.com/abmerop/gem5/commits/loadfile/ 

Longer term we are looking at better ways to do this and leaning towards using ethernet. The other option would be something like mounting the host filesystem but ethernet in the end would be more flexible for other uses.

powerjg:
You probably know my feelings here, but this should be used *very, very, very* carefully. It really should only be used for debugging. 

Creating new disk images with custom files loaded is now much easier. That is the path I would strongly encourage anyone to take.

Harshil2107 (approved):

BobbyRBruce (approved):

PR #1752, opened 2024-11-03
arch-arm: Cache a pointer to previously matched TLB entry

giactra:
One of the perks of the previous TLB storage implementation [1] is that its custom implementation of LRU exploited temporal locality to speed up simulation performance

            TlbEntry tmp_entry = *entry;
            for (int i = idx; i > 0; i--)
                table[i] = table[i - 1];
            table[0] = tmp_entry;
            return &table[0];

In other words the matching entry was placed as the first entry of the TLB table (table[0], top of LRU stack). In this way a following lookup would encounter it as the first entry while looping over the TLB table, therefore massively reducing simulation time when temporal locality is present
(most of TLB table loops would find a match in the first iteration).

   int x = 0;
    while (x < size) {
        if (table[x].match(lookup_data)) {

With the new implementation we decouple TLB storage from the replacement policy. The result is a more flexible implementation but with the drawback of a slower lookup/search.  We therefore we need to find another way to exploit temporal locality. This patch addresses it by caching a previously matched entry in the TLB table

[1]: https://github.com/gem5/gem5/blob/v24.0.0.0/src/arch/arm/tlb.cc

Change-Id: Id7dedf5411ea6f6724d1e4bdb51635417a6d5363

### DIFF
--- a/src/arch/arm/tlb.cc
+++ b/src/arch/arm/tlb.cc
@@ -73,19 +73,29 @@ TLB::Table::accessEntry(const KeyType &key)
 TlbEntry*
 TLB::Table::findEntry(const KeyType &key) const
 {
-    auto candidates = indexingPolicy->getPossibleEntries(key);
+    // Exploiting locality to maximaize simulator performance
+    if (prev && prev->N == key.pageSize && prev->match(key)) {
+        return prev;
+    }
 
-    for (auto candidate : candidates) {
+    for (auto candidate : indexingPolicy->getPossibleEntries(key)) {
         auto entry = static_cast<TlbEntry*>(candidate);
         // We check for pageSize match outside of the Entry::match
         // as the latter is also used to match entries in TLBI invalidation
         // where we don't care about the pageSize
         if (entry->N == key.pageSize && entry->match(key)) {
-            return entry;
+            return prev = entry;
         }
     }
 
     return nullptr;
+}
+
+void
+TLB::Table::invalidatePrev(const TlbEntry *invalid)
+{
+    if (!invalid || invalid == prev)
+        prev = nullptr;
 }
 
 TLB::TLB(const ArmTLBParams &p)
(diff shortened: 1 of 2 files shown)


src/arch/arm/tlb.hh
```diff
@@ -110,6 +110,10 @@ class TLB : public BaseTLB
         using AssociativeCache<TlbEntry>::accessEntry;
         TlbEntry* accessEntry(const KeyType &key) override;
         TlbEntry* findEntry(const KeyType &key) const override;
+
+      private:
+        /** Last matched entry */
+        mutable TlbEntry *prev = nullptr;
```
powerjg, line 127: If you flush the TLB should you set this back to nullptr? Also, if you evict this entry, should you set it to nullptr?

powerjg (commented):
Seems fine to me. I have one question below, but I imagine I am just missing something :)

src/arch/arm/tlb.hh
giactra: That is a valid question. Currrent logic handle those cases as well:

- TLB flush: you invalidate the entry meaning the valid boolean will be set to false. So prev->match() will return false
- Eviction: more or less the same: a new tag will be stored in the entry so entry->match will also return false

Having said that, I believe that while this minimizes the code being written, it can be a bit obscure. Having prev pointing to nullptr instead of stale data is cleaner and also faster (entry->match is expensive and it would be faster if we would skip it). I will evaluate a more explicit design which is flushing the cached pointer upon invalidation...

powerjg (approved):
Thank for the update!

src/arch/arm/tlb.cc
```diff
@@ -239,6 +249,8 @@ TLB::insert(const Lookup &lookup_data, TlbEntry &entry)
 
     table.insertEntry(lookup_data, victim);
 
+    table.invalidatePrev(victim);
```
odanrc, line 252: No action points for this PR: I just noticed that AssociativeCache::findVictim() calls invalidate(). I think the API should be simpler and more in line with the name of the functions: findVictim() should only find a victim, and invalidate() should do the invalidation.

odanrc (approved):

PR #1758, opened 2024-11-05
arch, cpu: Add generic getValidAddr to correct exetrace symbol table

rogerchang23424:
The getValidAddr is the method get virtual address with valid bits. It is useful to get the correct symbol table via valid virtual address.

For ARM, we have `purifyTaggedAddr` to get the right virtual address. For RISC-V, we only get lower 32 bits in RV32 mode to get the right symbol table.

Change-Id: I33ad7bec6e7ea4ec82cb1b3a7f521432c6d735b6

### DIFF
--- a/src/arch/arm/mmu.hh
+++ b/src/arch/arm/mmu.hh
@@ -276,6 +276,8 @@ class MMU : public BaseMMU
             Translation *translation, bool &delay, bool timing,
             CachedState &state);
 
+    Addr getValidAddr(Addr vaddr, ThreadContext *tc, Mode mode) override;
+
     Fault translateComplete(const RequestPtr &req, ThreadContext *tc,
             Translation *translation, Mode mode, ArmTranslationType tran_type,
             bool call_from_s2);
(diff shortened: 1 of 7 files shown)


src/arch/arm/mmu.hh
```diff
@@ -276,6 +276,9 @@ class MMU : public BaseMMU
             Translation *translation, bool &delay, bool timing,
             CachedState &state);
 
+    Addr
+    getValidAddr(Addr vaddr, ThreadContext *tc, Mode mode) override;
```
giactra: Nit, in method declaration return type and name should be in same line

src/arch/generic/mmu.hh
```diff
@@ -125,6 +125,9 @@ class BaseMMU : public SimObject
     translateFunctional(const RequestPtr &req, ThreadContext *tc,
                         Mode mode);
 
+    virtual Addr
+    getValidAddr(Addr vaddr, ThreadContext *tc, Mode mode);
```
giactra: Ditto

giactra (commented):
Makes sense to me; I have minor comments

giactra (approved):

PR #1761, opened 2024-11-06
arch-riscv: Implement Zcmt

rogerchang23424:
Implement table jump (cm.jt and cm.jalt) instructions

According to spec, when jump to the JVT entry address. We need to take two fetch steps. The first is fetching the zcmt instruction. The second is fetching the JVT entry. The JVT entries is considered the extension of table jump instructions. The JVT entries read from icache_port and enable get them in I-cache, so we can't follow the generic way to load JVT entry from memory helper.

In gem5 implementation we will get two StaticInst for cm.jalt. The first is move the PC to address of JVT entry, and mark the pc_state to zcmt fetch mode. When in zcmt mode, we will get the StaticInst that jumps to the JVT entry, and leaves the zcmt fetch mode.

If there is fault during second fetching, xEPC is set to instruction address, xCAUSE is set to expect type of fault, xTVAL is set to address of JVT entry

Spec: https://github.com/riscv/riscv-isa-manual/blob/main/src/zc.adoc#zcmt

Change-Id: I8e0caca35a22a51386677d70fa09c3dc97e4610a


powerjg:
~Looks like the checkpoint tests are failing. Should this affect checkpointing?~

Edit: I see now that you included a checkpoint upgrader. My bad.

@Harshil2107, can you update the checkpoints used in the tests?

src/arch/riscv/isa/decoder.isa
```diff
@@ -488,6 +488,15 @@ decode QUADRANT default Unknown::unknown() {
         format CompressedStore {
             0x5: decode ENABLE_ZCD {
                 0x0: decode CFUNCT6LOW3 {
+                    0x0: decode JTINDEX5TO7 {
```
powerjg, line 491: How are we going to deal with the fact that "This extension reuses some encodings from c.fsdsp"? 

Do the changes still work with original compressed extension?

powerjg (commented):
Do we need to do something (e.g., in the `RiscvISA` object) to make sure that Zcmt and Zcd are not both enabled at the same time since they are incompatible?

Harshil2107:
I updated the checkpoint https://resources.gem5.org/resources/riscv-hello-example-checkpoint?database=gem5-resources&version=4.0.0, you would need to update the following line to use version `4.0.0` and then the CI test will use the new checkpoint 

https://github.com/gem5/gem5/blob/63d25922a2db14643f132c0f62a224c7f7991c49/configs/example/gem5_library/checkpoints/riscv-hello-restore-checkpoint.py#L94

src/arch/riscv/isa/decoder.isa
rogerchang23424: Hi @powerjg,

Yes.

We have provided the option `enable_Zcd` in `RiscvISA` to allow user to switch `c.fsdsp` and `Zcm*` instructions

https://github.com/gem5/gem5/blob/63d25922a2db14643f132c0f62a224c7f7991c49/src/arch/riscv/RiscvISA.py#L117-L123

We have extra bit `enable_zcd` in `ExtMachInst`

https://github.com/gem5/gem5/blob/63d25922a2db14643f132c0f62a224c7f7991c49/src/arch/riscv/types.hh#L57-L64

We have `enable_zcd` checking in `decoder.isa`

https://github.com/gem5/gem5/blob/63d25922a2db14643f132c0f62a224c7f7991c49/src/arch/riscv/isa/decoder.isa#L489-L517

If `enable_zcd = 1`, we will decode it as `c.fsdsp`, otherwise decode it as `Zcm*` instructions.

powerjg: Got it. Thanks! I do think we should probably provide some error checking in the initialization of the RISC-V ISA object. I'm pretty sure that it's possible to configure it in a bad state right now. This can be done in another PR, though.

powerjg:
@Harshil2107, can you look into or explain why the checkpoint test is still failing? Please do what you can to get this change merged.

Harshil2107:
It looks like the issue is in the upgrader, the restore checkpoint fails with the following error

```
src/sim/serialize.hh:379: fatal: fatal condition !paramInImpl(cp, name, param) occurred: Can't unserialize 'board.processor.cores.core.xc.0:_zcmtSecondFetch'
```

I took a fresh checkpoint with this change and it added the 

```
_zcmtSecondFetch=false
_zcmtPc=0
```

fields to the `board.processor.cores.core.xc.0` section but the upgrader adds it to `board.processor.cores.core.isa` section
moving the above lines to `board.processor.cores.core.xc.0` makes the checkpoint restore properly.

rogerchang23424:
Hi @Harshil2107, I have updated the CLs to ToT but the `gem5_library_example_tests` still failed. Could you help me fix the checkpoint? Thanks

Harshil2107:
I will update the checkpoint and upload the updated one, the resource version will be the same. I will resun the tests after I have updated the checkpoint.

src/arch/riscv/isa/decoder.isa
hnpl: @rogerchang23424 Nit: Since Zcmt is not part of RVA profiles, I think we should have a separate flag for the Zcmt extension; i.e., turning Zcd off should not mean Zcmt is on. Though, this is not a blocking issue.

src/arch/riscv/isa/formats/zcmt.isa
```diff
@@ -0,0 +1,124 @@
...
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+////////////////////////////////////////////////////////////////////
+//
+// Unknown instructions
```
hnpl: This comment should be updated.

hnpl (approved):
lgtm

powerjg (dismissed):

util/cpt_upgraders/riscv-jvt.py
```diff
@@ -0,0 +1,72 @@
...
+                # Add dummy value for MISCREG_JVT, where idx is
+                # NUM_PHYS_MISCREGS - 1
+                mr.insert(len(mr) - 1, 0)
+                cpt.set(sec, "miscRegFile", " ".join(str(x) for x in mr))
+
+            if cpt.get(sec, "_zcmtSecondFetch", fallback="") == "":
+                cpt.set(sec, "_zcmtSecondFetch", "false")
+
+            if cpt.get(sec, "_zcmtPc", fallback="") == "":
+                cpt.set(sec, "_zcmtPc", "0")
```
Harshil2107:
```suggestion
    is_riscv = False
    for sec in cpt.sections():
        # Search for all ISA sections
        if (
            re.search(r".*processor.*\.core.*\.isa$", sec)
            and cpt.get(sec, "isaName") == "riscv"
        ):
            is_riscv = True
            # Updating Snmrnm misc registers (dummy values)
            mr = cpt.get(sec, "miscRegFile").split()
            if len(mr) >= 170:
                print(
                    "MISCREG_* Zcmt registers already seem " "to be inserted."
                )
            else:
                # Add dummy value for MISCREG_JVT, where idx is
                # NUM_PHYS_MISCREGS - 1
                mr.insert(len(mr) - 1, 0)
                cpt.set(sec, "miscRegFile", " ".join(str(x) for x in mr))
    
    for sec in cpt.sections():
        #board.processor.cores.core.xc.0
        if is_riscv and re.search(r".*processor.*\.core.*\.xc.*", sec):


            if cpt.get(sec, "_zcmtSecondFetch", fallback="") == "":
                cpt.set(sec, "_zcmtSecondFetch", "false")

            if cpt.get(sec, "_zcmtPc", fallback="") == "":
                cpt.set(sec, "_zcmtPc", "0")

```
It seems that the checkpoint tests are still failing even after updating the checkpoints. Upon investigation, it looks like the changes made by the checkpoint upgrader are not correctly aligning with the checkpoint when it is taken directly at the head of this PR.  

The issue appears to be in the upgrader, as restoring the checkpoint fails with the following error:  
```
src/sim/serialize.hh:379: fatal: fatal condition !paramInImpl(cp, name, param) occurred: Can't unserialize 'board.processor.cores.core.xc.0:_zcmtSecondFetch'
```  

When I inspected the checkpoint, gem5 expects the following lines:  
```
_zcmtSecondFetch=false
_zcmtPc=0
```  
to be in the `board.processor.cores.core.xc.0` section. However, the upgrader is incorrectly adding these lines to the `board.processor.cores.core.isa` section.  

The suggested change I highlighted above allows the checkpoint upgrader to update the checkpoints properly. However, I’m uncertain about the correctness of this solution.

rogerchang23424:
I have fixed the riscv-jvt.py upgrader issue and it should be able to pass the test

Harshil2107:
I will update the checkpoint with the current head of the PR and re run the test.

Harshil2107 (approved):
I am not an expert but it looks like all the requested changes were addressed and the changes look good to me.

powerjg (approved):

PR #1762, opened 2024-11-06
base: Introduce registerExtraLog() in Logger

yp155136:
We may want to register extra logs when logging in gem5. For example, log CPU PC when gem5 panic / gem5 fatal.

To achieve this, I introduce extraLog and registerExtraLog() in Logger class. We can attach custom logs after the log message. For example:

Logger::getPanic().registerExtraLog([]() { return getCPUPC(); });

In the existing ExitLogger, we can also put the memory usage printing message in extraLog

Change-Id: I4daa175c6e4eeeee1234ccccffffddddeeeeeeee

### DIFF
--- a/src/base/logging.hh
+++ b/src/base/logging.hh
@@ -42,8 +42,11 @@
 #define __BASE_LOGGING_HH__
 
 #include <cassert>
+#include <functional>
 #include <sstream>
 #include <utility>
+#include <string>
+#include <vector>
 
 #include "base/compiler.hh"
 #include "base/cprintf.hh"
@@ -123,19 +126,48 @@ class Logger
      */
     [[noreturn]] void exit_helper() { exit(); ::abort(); }
 
+    /** Register callback to generate extra log. */
+    void
+    registerExtraLog(std::function<std::string()> cb)
+    {
+        extraLogs.emplace_back(std::move(cb));
+    }
+
   protected:
     bool enabled;
+
+    /**
+     * Generate the formatted extra log string. Developers can register extra
+     * logs when printing logs, such as PC of each CPUs, .etc. This function
+     * ensures that there's at least a new line between each extra log.
+     * */
+    std::string
+    getFormattedExtraLog() const
+    {
+        std::stringstream ss;
+        for (const auto &cb : extraLogs) {
+            std::string str = cb();
+            ss << str;
+            if (str.length() && str.back() != '\n' && str.back() != '\r')
+                ss << std::endl;
+        }
+        return ss.str();
+    }
 
     /** Generates the log message. By default it is sent to cerr. */
     virtual void
     log(const Loc &loc, std::string s)
     {
-        std::cerr << loc.file << ":" << loc.line << ": " << s;
+        std::cerr << loc.file << ":" << loc.line << ": " << s
+                  << getFormattedExtraLog();
     }
 
     virtual void exit() { /* Fall through to the abort in exit_helper. */ }
 
     const char *prefix;
+
+    /** Data structure to store extra log callbacks. */
+    std::vector<std::function<std::string()>> extraLogs;
 };
 
 #define base_message(logger, ...)                                       \
--- a/src/base/logging.cc
+++ b/src/base/logging.cc
@@ -52,15 +52,11 @@ namespace {
 class ExitLogger : public Logger
 {
   public:
-    using Logger::Logger;
-
-  protected:
-    void
-    log(const Loc &loc, std::string s) override
+    ExitLogger(const char *prefix) : Logger(prefix)
     {
-        std::stringstream ss;
-        ccprintf(ss, "Memory Usage: %ld KBytes\n", memUsage());
-        Logger::log(loc, s + ss.str());
+        registerExtraLog([]() {
+            return csprintf("Memory Usage: %ld KBytes\n", memUsage());
+        });
     }
 };
 
--- a/src/base/logging.test.cc
+++ b/src/base/logging.test.cc
@@ -157,6 +157,27 @@ TEST_F(LoggingFixture, DisabledPrint)
     ASSERT_EQ(gtestLogOutput.str(), "");
 }
 
+/** Test print with extra logs. */
+TEST_F(LoggingFixture, PrintWithExtraLogs)
+{
+    Logger logger("test: ");
+
+    logger.registerExtraLog([]() { return "Extra log 1\n"; });
+
+    gtestLogOutput.str("");
+    logger.print(Logger::Loc("File", 10), "message");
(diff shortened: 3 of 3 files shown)


src/base/logging.hh
```diff
@@ -123,19 +126,39 @@ class Logger
      */
     [[noreturn]] void exit_helper() { exit(); ::abort(); }
 
+    void registerExtraLog(std::function<std::string()> cb)
```
giactra: Nit:

```
void
registerExtraLog
```

src/base/logging.hh
```diff
@@ -123,19 +126,39 @@ class Logger
...
+    {
+        extraLog.emplace_back(std::move(cb));
+    }
+
   protected:
     bool enabled;
 
+    std::string getFormattedExtraLog()
```
giactra: Ditto

src/base/logging.hh
```diff
@@ -123,19 +126,39 @@ class Logger
...
+        }
+        return ss.str();
+    }
+
     /** Generates the log message. By default it is sent to cerr. */
     virtual void
     log(const Loc &loc, std::string s)
     {
+        s += getFormattedExtraLog();
         std::cerr << loc.file << ":" << loc.line << ": " << s;
```
giactra: Nit: I find this more elegant:

` std::cerr << loc.file << ":" << loc.line << ": " << s << getFormattedExtraLog();`

giactra (commented):
Hi, I only have very minor comments

src/base/logging.hh
yp155136: Thanks for catching! Fixed

src/base/logging.hh
yp155136: Thanks for catching! Fixed

src/base/logging.hh
yp155136: Thanks for the suggestion! Fixed!

yp155136:
Thanks for the review! I've applied your suggestion in the latest commit :)

src/base/logging.hh
```diff
@@ -123,19 +126,41 @@ class Logger
...
-        std::cerr << loc.file << ":" << loc.line << ": " << s;
+        std::cerr << loc.file << ":" << loc.line << ": " << s
+                  << getFormattedExtraLog();
     }
 
     virtual void exit() { /* Fall through to the abort in exit_helper. */ }
 
     const char *prefix;
+
+    std::vector<std::function<std::string()>> extraLog;
```
odanrc: Could you add doxygen documentation for this variable?

src/base/logging.hh
```diff
@@ -123,19 +126,41 @@ class Logger
...
+    void
+    registerExtraLog(std::function<std::string()> cb)
+    {
+        extraLog.emplace_back(std::move(cb));
+    }
+
   protected:
     bool enabled;
 
+    std::string
```
odanrc, line 144: Ditto add documentation

src/base/logging.hh
```diff
@@ -123,19 +126,41 @@ class Logger
      */
     [[noreturn]] void exit_helper() { exit(); ::abort(); }
 
+    void
```
odanrc, line 130: Ditto add documentation

src/base/logging.hh
```diff
@@ -123,19 +126,41 @@ class Logger
...
+    registerExtraLog(std::function<std::string()> cb)
```
odanrc, line 131: This function was added to the public API, and as such respective unit tests should be added (src/base/logging.test.cc)

src/base/logging.hh
```diff
@@ -123,19 +126,41 @@ class Logger
...
+            std::string str = cb();
+            ss << str;
+            if (str.length() && str.back() != '\n' && str.back() != '\r')
+                ss << std::endl;
+        }
+        return ss.str();
+    }
+
     /** Generates the log message. By default it is sent to cerr. */
     virtual void
```
odanrc, line 158: Does this function still need to be virtual?

odanrc (changes requested):

src/base/logging.hh
yp155136: Sure. documentation added!

src/base/logging.hh
yp155136: Added!

src/base/logging.hh
yp155136: Added!

src/base/logging.hh
yp155136: Test added in logging.test.cc. Please check :)

src/base/logging.hh
yp155136: After the refactor in this PR, I think this function no longer need to a virtual function. Fixed!

src/base/logging.hh
odanrc: nit: I know it is like this in the other functions (and I was the one that wrote the poor documentation in log()), but could you be more verbose? When adding documentation one should try to focus on information that is not acquired from the variable/function. For example, by reading the function name I know that this function will return the formatted extra log. But what is an extra log? What is the format?

src/base/logging.hh
```diff
@@ -123,19 +126,44 @@ class Logger
...
+    {
+        extraLogs.emplace_back(std::move(cb));
+    }
+
   protected:
     bool enabled;
 
+    /** Generate the formatted extra log string. */
+    std::string
+    getFormattedExtraLog()
```
odanrc: Sorry, just noticed this. This should be a const function, as it is not expected to modify the class.

src/base/logging.hh
```diff
@@ -123,19 +126,44 @@ class Logger
...
+    {
+        std::stringstream ss;
+        for (auto &cb : extraLogs) {
```
odanrc: Ditto const auto &, as we do not want to modify cb

odanrc (changes requested):

src/base/logging.hh
yp155136: Added const. Thanks for catching!

src/base/logging.hh
yp155136: Added const. Thanks for catching!

src/base/logging.hh
yp155136: Thanks for the suggestion! I've added more description of this function. Please check :)

odanrc (approved):
Thanks for the changes! LGTM

src/base/logging.hh
yp155136: https://github.com/gem5/gem5/blob/stable/src/base/gtest/logging_mock.cc#L39

I find out that GTestLogger overrides the log function. So I change the function back to virtual in the latest commit

giactra (approved):

PR #1766, opened 2024-11-07
stdlib: fix LoopPoint module not getting relative counts

studyztp:
When restoring a checkpoint for a LoopPoint, we need to use the relative count for the marker instead of the global count because the checkpoint does not record the PC count map at where the checkpoint was taken.
However, the current LoopPoint stdlib module does not have a method to access the relative count information for the markers. This patch adds a method to use the relative count for the target PC count pairs when we have a target region id. We also offer a param to overwrite this default behavior as `use_global_count`.

Beside the above fix, this patch also ensure the region id is an integer to prevent unable to find the region id due to type difference. This patch also make sure the PcCountPair object return from the backend has the same type as the Python PcCountPair object to prevent unable to find the PcCountPair due to type difference.


studyztp:
@BobbyRBruce this is the bug fix that I was mentioning. This should not change the user interface of the current module since it is hidden under the `set_se_looppoint_workload`.

BobbyRBruce (approved):

BobbyRBruce:
You have some failing pyunit tests @studyztp . You'll need to fix them.

erin-le:
@studyztp: This PR is being closed for inactivity. Please feel free to reopen it if you have time to fix the pyunit tests at a later point. Thank you!

PR #1768, opened 2024-11-08
arch-riscv: Use getValidAddr to get zero-extend address in RV32 mode

rogerchang23424:
Previous PR #1758 implements the generic getValidAddr to get pure vaddr without any tags or sign-extend bits.

In RISC-V implementation, the getValidAddr will zero-extend address in RV32 mode and use it to do TLB translation. Use getValidAddr to get zero-extend vaddr can reduce zero-extend repetition

Change-Id: I2273ce48bccb873790103ba0fcdb0b48de9ced4c

### DIFF
--- a/src/arch/riscv/tlb.cc
+++ b/src/arch/riscv/tlb.cc
@@ -386,15 +386,8 @@ TLB::translate(const RequestPtr &req, ThreadContext *tc,
             if (req->getFlags() & Request::PHYSICAL) {
                 /**
                  * we simply set the virtual address to physical address.
-                 *
-                 * For RV32, we follow what the specification said:
-                 * When mapping between narrower and wider addresses,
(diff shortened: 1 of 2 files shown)


powerjg (approved):
lgtm. The tests are failing randomly... We'll get them re-run.

Do you think we need better CI tests for rv32?

rogerchang23424:
Hi @powerjg,

Sorry for reply lately, we have included the RV32 SE riscv-tests in asmtest before [\[1\]](https://gem5-review.googlesource.com/c/public/gem5-resources/+/65551). The tests contain frequently used instructions, so I think it is enough for me.

powerjg:
@BobbyRBruce this should probably be on staging

powerjg (approved):

PR #1769, opened 2024-11-08
gpu-compute: Reverting L1 TLB entries and L1, L2, L3 TLB assoc back to 32 from 64

shreyamukh:
Previously a commit changed the GPU L1 TLB entries to 64 from 32, and changed all GPU TLB associativities to 64 from 32.  Further studying has shown this change does not accurately model the behavior of the underlying GPU, so we reverted it to 32, which is more representative.

### DIFF
--- a/configs/common/GPUTLBOptions.py
+++ b/configs/common/GPUTLBOptions.py
@@ -44,8 +44,8 @@ def tlb_options(parser):
     #   L1 TLB Options (D-TLB, I-TLB, Dispatcher-TLB)
     # ===================================================================
 
-    parser.add_argument("--L1TLBentries", type=int, default="64")
-    parser.add_argument("--L1TLBassoc", type=int, default="64")
+    parser.add_argument("--L1TLBentries", type=int, default="32")
+    parser.add_argument("--L1TLBassoc", type=int, default="32")
     parser.add_argument(
         "--L1AccessLatency",
         type=int,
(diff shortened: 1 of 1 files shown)


shreyamukh:
Reverting L1 Bentries and L1, L2, L3 Basic back to 32 from 64.

mattsinc:
@shreyamukh it looks like there is a typo in title "Bentries" --> "TLBentries"

mattsinc (approved):
@abmerop this is the patch reverting the prior change we discussed previously -- the change I put in actually hurts correlation with the equivalent real GPU part for the L1 and L2 TLBs.

mattsinc:
@shreyamukh you'll need to fix the title before this can be merged

erin-le:
@shreyamukh: Thank you for your contribution! In addition to fixing the typos, the title also needs a tag, which in this case should be `gpu-compute`.
With the typo fixes, I believe the title should be `gpu-compute: Reverting L1 TLBentries and L1, L2, L3 TLBassoc back to 32 from 64.`

mattsinc:
> @shreyamukh: Thank you for your contribution! In addition to fixing the typos, the title also needs a tag, which in this case should be `gpu-compute`. With the typo fixes, I believe the title should be `gpu-compute: Reverting L1 TLBentries and L1, L2, L3 TLBassoc back to 32 from 64.`

@shreyamukh can you please make these changes?

mattsinc:
> > @shreyamukh: Thank you for your contribution! In addition to fixing the typos, the title also needs a tag, which in this case should be `gpu-compute`. With the typo fixes, I believe the title should be `gpu-compute: Reverting L1 TLBentries and L1, L2, L3 TLBassoc back to 32 from 64.`
> 
> @shreyamukh can you please make these changes?

Just checking on this again, let me know if you need help.

mattsinc:
@shreyamukh since we are trying to release the next version of gem5 tonight, I have made the fix myself.  Next week let's discuss

CC @powerjg , @BobbyRBruce for FYI

PR #1779, opened 2024-11-13
Update exit event to send in integer from simulation to host

Harshil2107:


mattsinc:
@abmerop : will we need to update the GPUSE/GPUFS exit events to work with this change?

Harshil2107:
I am going to add a new event type instead of updating m5 exit so this change should not affect any of the current exit events.

PR #1780, opened 2024-11-14
tests: update gpu weekly test to test mis hip instead of fw hip

Harshil2107:
I am moving this test because the fw tests fails on runners due to the following error 

```
File "/__w/gem5/gem5/tests/../ext/testlib/helper.py", line 207, in log_call
    raise subprocess.CalledProcessError(retval, cmdstr)
subprocess.CalledProcessError: Command '/__w/gem5/gem5/build/VEGA_X86/gem5.opt -d /tmp/gem5outesz2d05s -re --silent-redirect /__w/gem5/gem5/configs/example/apu_se.py -n3 --mem-size=8GB -c /__w/gem5/gem5/tests/gem5/gpu/../resources/gpu-pannotia/pannotia-bins/fw_hip.gem5 --options -f /__w/gem5/gem5/tests/gem5/gpu/../resources/gpu-pannotia/pannotia-datasets/1k_128k.gr -m default' died with &lt;Signals.SIGKILL: 9&gt;.
```

This looks like runners are killing the test due to lack of resources. This test did past locally so I changed the test with mis_hip.

### DIFF
--- a/.github/workflows/weekly-tests.yaml
+++ b/.github/workflows/weekly-tests.yaml
@@ -211,7 +211,7 @@ jobs:
                   path: tests/testing-results
                   retention-days: 7
 
-    gpu-test-pannotia-fw-hip:
+    gpu-test-pannotia-mis-hip:
         runs-on: [self-hosted, linux, x64]
         container: ghcr.io/gem5/gcn-gpu:latest
         timeout-minutes: 2160 # 36 hours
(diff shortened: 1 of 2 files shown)


mattsinc:
@Harshil2107 : maybe I'm missing something obvious, but since MIS uses the same input file and is a similar type of test to FW, do you have any intuition about why one passes and one fails?  The error is generic enough that it's hard for me to say...

Harshil2107:
I am not sure, but when I ran the test locally it did run, if I recall correclty. This test was killed before the timeout so I am not sure. I might run it locally to see if I can get any more details, the local run last time took a day or so. I couldnt see anything on simerr.txt or simout.txt on the runner output. Also all the pannotia test use the same input file. I updated this so that we can have our weekly test to run to completion.

BobbyRBruce (approved):

BobbyRBruce:
I'm going to merge this just because I want the tests passing but @Harshil2107 : Can you try running this test and keep track of memory usage (I don't know how to do this tbh beyond just checking with `top` manually every so often), but I agree this error seems like the process was killed by "something" and the only reason I can think that'd be is because it took up too many resources... but I'd at least like to verify that this is a hog... because I don't see why it should be.

mattsinc:
> I am not sure, but when I ran the test locally it did run, if I recall correclty. This test was killed before the timeout so I am not sure. I might run it locally to see if I can get any more details, the local run last time took a day or so. I couldnt see anything on simerr.txt or simout.txt on the runner output. Also all the pannotia test use the same input file. I updated this so that we can have our weekly test to run to completion.

Agree with @BobbyRBruce this is very suspicious ... please let me know what you find

powerjg:
> (I don't know how to do this tbh beyond just checking with top manually every so often)

Use `/usr/bin/time` that will give you the memory information as well as the time.

Issue #1722, opened 2024-10-28
Build gem5 successfully with no errors, but no .opt file is generated.


ivanaamit:
It appears others have encountered this issue, though we haven’t been able to reproduce it ourselves.

#1536 
#1706 

@liuyuxivapor @soniab Please let us know if you're still experiencing this, or if you have found a solution.

If it’s still an issue, could you provide details about your host system? This will help us to create a docker environment and to try to reproduce the issue.

If you have resolved it, sharing your solution would be very helpful for others in the future. Thank you!

JianYugen:
Hi. I also recently got this problem when setting up gem5. I'm posting compilation message and my system info here.

Compilation: 
```
scons: Reading SConscript files ...

You're missing the pre-commit/commit-msg hooks. These hook help to ensure your
code follows gem5's style rules on git commit and your commit messages follow
our commit message requirements. This script will now install these hooks in
your .git/hooks/ directory.
Press enter to continue, or ctrl-c to abort:
 Cannot find 'pre-commit'. Please ensure all Python requirements are
installed. This can be done via 'pip install -r requirements.txt'.
It is strongly recommended you install the pre-commit hooks before working with gem5. Do you want to continue compilation (y/n)?
y
Mkdir("/home/jianyugen/gem5/build/X86/gem5.build")
Checking for linker -Wl,--as-needed support... (cached) yes
Checking for compiler -gz support... (cached) yes
Checking for linker -gz support... (cached) yes
Info: Using Python config: python3-config
Checking for C header file Python.h... (cached) yes
Checking Python version... (cached) 3.8.10
Checking for accept(0,0,0) in C++ library None... (cached) yes
Checking for zlibVersion() in C++ library z... (cached) yes
Checking for C library tcmalloc_minimal... (cached) yes
Building in /home/jianyugen/gem5/build/X86
"build_tools/kconfig_base.py" "/home/jianyugen/gem5/build/X86/gem5.build/Kconfig" "/home/jianyugen/gem5/src/Kconfig"
Checking size of struct kvm_xsave ... (cached) yes
Checking for C header file fenv.h... (cached) yes
Checking for C header file png.h... (cached) no
Warning: Header file <png.h> not found.
         This host has no libpng library.
         Disabling support for PNG framebuffers.
Checking for clock_nanosleep(0,0,NULL,NULL) in C library None... (cached) yes
Checking for C header file valgrind/valgrind.h... (cached) yes
[...]

Issue #1706, opened 2024-10-23
gem5 is not building for RISCV


soniab:
version used: gem5-24.0.0.0.tar.gz
I have also tried with git clone https://github.com/gem5/gem5.git

**modules used:** 
module load python/3.12.3 
 module load llvm/18.1.4
module load scons/4.2.0 
module load gcc/12.2.0 

command used:
scons  build/RISCV/gem5.opt -j4 > build_log.txt

**Issue:** gem5.opt has not been created inside build/RISCV
I can see only the following folders:
soniar@login1:/cfs/klemming/projects/snic/naiss2023-23-49/sonia/gem5-new/gem5-24.0.0.0> ls build/RISCV/
arch  base  config  cpu  debug  dev  enums  ext  gem5.build  gem5py  gem5py_m5  kern  learning_gem5  mem  params  python  sim  sst  systemc

I checked available targets:
The predefined targets currently available are:

          ALL
          ARM
          ARM_MESI_Three_Level
          ARM_MESI_Three_Level_HTM
          ARM_MOESI_hammer
          Garnet_standalone
          MIPS
          NULL
          NULL_MESI_Two_Level
          NULL_MOESI_CMP_directory
          NULL_MOESI_CMP_token
          NULL_MOESI_hammer
          POWER
          RISCV
          RISCV_old
          SPARC
          VEGA_X86
          X86
          X86_MESI_Two_Level
          X86_MI_example
          X86_MOESI_AMD_Base

**I can see only one error:** 
Error: A PrgEnv-* modulefile must be loaded.
readelf: Error: 'nefZx5pux.o': No such file
ext/libelf/elf.c:34:46: error: expected expression before ',' token
   34 |         .libelf_arch            = LIBELF_ARCH,
But I am not sure that it should impact anything

[build_log.txt](https://github.com/user-attachments/files/17494199/build_log.txt)

erin-le:
Duplicate

Issue #1536, opened 2024-09-04
No error reported but fail to link or generate an .opt file.


liuyuxivapor:
**Describe the bug**
No error reported but fail to link or generate an .opt file.

**Affects version**
24.0

**gem5 Modifications**
Nothing changed compared to the repo.

**To Reproduce**
Steps to reproduce the behavior. Please assume starting from a clean repository:

1. Compile gem5 with command `scons build/X86/gem5.opt -j10`

**Terminal Output**

```shell
<lots of outputs>
 [SO Param] m5.objects.Sequencer, RubyPort -> X86/python/_m5/param_RubyPort.cc
 [     CXX] src/base/date.cc -> X86/base/date.o
scons: done building targets.
*** Summary of Warnings ***
Warning: Header file <capstone/capstone.h> not found.
         This host has no capstone library installed.
Warning: Couldn't find HDF5 C++ libraries. Disabling HDF5 support.
```

**Expected behavior**

**Host Operating System**
Ubuntu 20.04

**Host ISA**
X86

**Additional information**
It seems that I have installed all requirements, but it still appears no error!

Harshil2107:
It looks like you are building gem5 with just 1 thread so it will take a long time. I would recommend trying to build gem5 with more than one thread using the `-j` command. you can use `nproc` to get the total number of threads you have on your system. To build gem5 you can use `scons build/X86/gem5.opt -j16` for example.

liuyuxivapor:
> It looks like you are building gem5 with just 1 thread so it will take a long time. I would recommend trying to build gem5 with more than one thread using the `-j` command. you can use `nproc` to get the total number of threads you have on your system. To build gem5 you can use `scons build/X86/gem5.opt -j16` for example.

I have tried this with `-j10`, but got the same problem. :( Thanks a lot anyway.

giactra:
Have you tried to redirect stderr to file to check if there is a linkage error line you have missed in the terminal?

`scons build/X86/gem5.opt -j10 2> error.log`

Mr-Cong:
The same error. I have solved it! Use the tar.gz version or unzip the zip version in Linux, not in Windows.

Harshil2107:
It looks like the issue was solved. I am closing the issue as resolved.

liuyuxivapor:
> Have you tried to redirect stderr to file to check if there is a linkage error line you have missed in the terminal?
> 
> `scons build/X86/gem5.opt -j10 2> error.log`

[...]

PR #1785, opened 2024-11-17
misc: Fix scons tags

odanrc:
Fixed a bug in Scons tag resolution, which was making files that were not required be added to the compilation unit.

Basically, apply_filter() is used to filter the list of all sources to figure out if it contains a certain tag. We should not resolve the tags of the source list being filtered, otherwise we do not filter anything. In other words, when parsing a file F, which requires tag T, if another file F2 contains tag T2, which implies tag T, then F2 is should not be a dependency of F.

As a side effect of the debugging some classes in that file had their printing functionality improved.

After this bug was fixed, it highlighted that the tags API were not being used correctly - i.e., gem5 stopped compiling. We have tags(), which overrides all tags and replaces them with the provided tags, and we have add_tags(), which adds tags to a compilation unit. To simplify the API, the add_tags() function was removed.

When using tags() one does not need to add gem5 lib as a tag, as it is added by default. It is still possible not to automatically add this tag by setting tag_gem5_lib to false. This, however, would rarely ever be desired (one use case are the mocks for logging code in gtest).


src/arch/arm/SConscript
```diff
@@ -63,104 +63,110 @@ if env["CONF"]["USE_ARM_ISA"]:
...
-Source("insts/pred_inst.cc", tags="arm isa")
-Source("insts/pseudo.cc", tags="arm isa")
-Source("insts/sme.cc", tags="arm isa")
-Source("insts/static_inst.cc", tags="arm isa")
-Source("insts/sve.cc", tags="arm isa")
-Source("insts/sve_mem.cc", tags="arm isa")
-Source("insts/vfp.cc", tags="arm isa")
-Source("insts/crypto.cc", tags="arm isa")
-Source("insts/tme64.cc", tags="arm isa")
+Source("decoder.cc", add_tags="arm isa")
```
giactra: Hi @odanrc; there's something I am missing:

Let's take decoder.cc as en example. What's the difference between

`Source("decoder.cc", add_tags="arm isa")`

and

`Source("decoder.cc", tags="arm isa")`

It looks like in both cases arm_isa is the only tag specified... Or is there another utility that adds tags from somewhere else in Scons? I am not particularly keen on the idea of always appending the tag as it seems to me opening a path of overtagging what is needed. We should restrict tagging as a base only when needed and make appending the exception.

Again, I might be missing something here

giactra (commented):
Hey @odanrc , thanks for handling this. Can I ask you to use commits tags for the PR title as well? At the end the PR title a merge commit title and I'd like it to be equally informative on the areas being touched

src/arch/arm/SConscript
odanrc: If you check SourceItem, in site_scons/gem5_scons/sources.py, you will see that there is a default tag: gem5 lib. When we assign tags=, we end up without this tag, and this causes a lot of places to break.

After some reflection, I agree that appending tags seems to be more prone to overtagging, but fixing it by removing the default tag is significantly more time consuming, as it would require trial-and-error/manually fixing every single tag (we have used the tagging API wrong all these years). In other words, I would rather leave it for future work, and focus on making SimObjects unit-testable now.

By the way, we were already overtagging in a way: we were always compiling most of the isa-specific files, even if they were not being used. The latest commit in this PR fixes that. I didn´t notice any difference in compilation speed, though, although I didn´t spend much time on that.

giactra: Thanks daniel, makes sense now. Another solution would be:

1) We do not remove the default tag clause. We keep it there so that if there is some source which is relying on it, it will still use it
2) In all other instances (where you replace tags with add_tags), we keep tags and we explicitly use "gem5 lib"
3) In the future, we could remove the default tag behaviour. It's tedious but not error prone: it would suffice to assert if tags is None (and maybe if add_tags is also None) and try to compile to spot places where this default behaviour is used. One by one we will catch all instances

I believe I haven't yet seen the case in gem5 where there is a compelling reason for having the add_tags parameter. I believe it is useful if it was a method and you had something like:

```
source = Source(my_source.cc, tags="gem5 lib")
if (...):
   source.add_tags("conditional tag")
```

But as add_tags is just a parameter, the only motivation behind having it is to just exploit the default "gem5 lib" (and avoid code repetition). Why doing

```
source = Source(my_source.cc, tags="tag A", add_tags["tag B", "tag C"])
```

if you could do

```
source = Source(my_source.cc, tags=["tag A", "tag B", "tag C"])
```

So basically I still believe it is a weird behaviour to obscurely rely on a default value and to append tags in the common case (I can't tell what's a Source tag without checking site_scons/gem5_scons/sources.py ). But I don't want to push back too much on this as you could make the case we avoid code duplication in this way... Let me know what's your thought. I will approve if you prefer going this way

odanrc:
Hey, @mattsinc , this PR is fixing compilation in a way that isa-specific files are no longer compiled if their respective isas are not being used. The problem with this is that although the amdgpu is configured for x86 it was using arm's fplib (and that is why the Source("insts/fplib.cc") line was before the isa check in src/arch/arm/Sconscript).

To get this PR merged I will just keep that line as before, but that means that we would be compiling it in other isas too, which is likely bad. How do we proceed here? Would you be able to add the missing implementation to amdgpu so that we can hide fplib in the future?

src/arch/arm/SConscript
odanrc: >     1. We do not remove the default tag clause. We keep it there so that if there is some source which is relying on it, it will still use it
> 
>     2. In all other instances (where you replace tags with add_tags), we keep tags and we explicitly use "gem5 lib"
> 
>     3. In the future, we could remove the default tag behaviour. It's tedious but not error prone: it would suffice to assert if tags is None (and maybe if add_tags is also None) and try to compile to spot places where this default behaviour is used. One by one we will catch all instances

I agree with everything that you said, but unfortunately the trial-and-error labor-intensive step is 2. I will not be able to proceed with this PR until the first week of December, but maybe when I get back I will give it a try.

giactra (dismissed):
I am approving this, even if I don't believe this should be the final state of the tagging system, it is at least fixing an existing problem

Please @odanrc could you open an issue ticket or a discussion summarizing what we discussed here? I will merge this PR once this is done. Thanks for your patience

odanrc:
> I am approving this, even if I don't believe this should be the final state of the tagging system, it is at least fixing an existing problem
> 
> Please @odanrc could you open an issue ticket or a discussion summarizing what we discussed here? I will merge this PR once this is done. Thanks for your patience

I plan to try to tackle what you suggested still in this PR. I will probably be able to do so this weekend, so we can hold off merging for one more week so that I can give it a spin.

odanrc:
@giactra , what do you think is better? The way it is implemented now (the user has to declare all tags, including ´gem5 lib´) or tags always implicitly declares ´gem5 lib´? The second option seems to be much more convenient and less error-prone, but it also removes flexibility. I am not sure we need that flexibility, though.

giactra:
> @giactra , what do you think is better? The way it is implemented now (the user has to declare all tags, including ´gem5 lib´) or tags always implicitly declares ´gem5 lib´? The second option seems to be much more convenient and less error-prone, but it also removes flexibility. I am not sure we need that flexibility, though.

Yeah I get your point. I am in favour of being explicit, but let's get some opinion from other maintainers

powerjg:
I have three thoughts on this change:
1. I don't understand scons, our tag system, or why it's so complicated. Therefore, I'm happy to defer to others.
2. My biggest concern is teaching new people how to use and develop for gem5. So, if we are changing the way people have to write their scons files, we should make sure the change is worth it. Remember, 100s or 1000s of people use gem5, maybe 10x fewer develop new models for gem5, but we are only ~10 maintainers. Sometimes it's worth a little more effort from us to make it easier on the "masses." In other words, I would be for whatever is the *minimum* user-facing change.
3. Whatever is changed here, the tag system in scons needs to be better documented. @odanrc now that you seem to understand what the purpose is and how to use it, I would greatly appreciate you writing it down so that others can learn from your work. As I mentioned in (1), I truly have no idea what tags are or how to use them. So, even just for me, it would be a big help.

My last thought is that with the ALL ISA and ALL RUBY changes, we are *very close* to being able to distribute gem5 binaries. (We have one big missing piece which is a plug-in system to extend gem5...) I am hopeful that within the medium term we won't need to be able to customize the build so much. So, I don't think we need a perfect solution if tags are meant to customize the build. We just need something that fixes the bug, IMO.

odanrc:
I have changed the code so that the gem5 lib tag is always added to every source, as it is backwards compatible, simpler and less error prone. I don´t think this will be an issue, as all tests pass (but I think we should keep an eye open for any possible issues that rise in src/base/gtest).

Regarding scons, I am not a Python guy; I made it work, but it does not mean that my understanding is correct (and there is a lot of black magic happening due to lack of documentation). In any case, here is what I wanted to achieve with the fix (please place this documentation in the most suitable place):

Scons will associate files so that they are compiled using only their dependencies. To do so we could either manually add all files that a file depends on, but that would be quite cumbersome. The alternative is to tag groups of common files, so that they can be easily added as dependencies. We can use the predicates defined in site_scons/gem5_scons/sources.py to select how we want to associate a dependency.

The predicates are:
- with_any_tags(tag_list): declares all files tagged with any of the tags in the list as dependencies for this source
- with_all_tags(tag_list): declares all files simultaneously tagged with all of the tags in the list as dependencies for this source
- without_tags(tag_list): declares all files that are not tagged with any of the tags in the list as dependencies for this source
- without_tag(tag): declares all files that are not tagged with the tag as dependencies for this source
- with_tag(tag): declares all files tagged with the tag as dependencies for this source

For example, let's assume that we have the files pair x_0.cc/hh, and x_1.cc/hh. Both these files do not have any dependencies, and they are supposed to be used in many places in the codebase. Let's also assume that file Y uses the functionalities of x_0 and x_1.

If we didn't have tags, we would have to declare these files as:
Source('x_0.cc')
Source('x_1.cc')
Source('y.cc', 'x_0.cc', 'x_1.cc')

But with tags we can declare them as:
Source('x_0.cc', tags=['tag x'])
Source('x_1.cc', tags=['tag x'])
Source('y.cc', with_tag('tag x'))

Keep in mind that by default all files have the 'gem5 lib' tag

powerjg:
Thanks, Daniel. I have no problems with this change, but I don't have enough confidence to approve. If Giacomo doesn't get back to this in ~1 week, ping me and I'll approve.

abmerop:
I've only parsed the commit message so far, but I thought maybe this could fix some of the issues users were seeing where gem5 was not linking, maybe due to scons thinking there was not a change when there should have been due to tags?  https://github.com/gem5/gem5/issues/1722 https://github.com/gem5/gem5/issues/1536 https://github.com/gem5/gem5/issues/1706

odanrc:
> I've only parsed the commit message so far, but I thought maybe this could fix some of the issues users were seeing where gem5 was not linking, maybe due to scons thinking there was not a change when there should have been due to tags? #1722 #1536 #1706

I think these are probably unrelated issues; before this PR we were overlinking, not underlinking.

powerjg (approved):
No one has objected for 3 weeks. I'm good with this as-is. It may need to be rebased and re-tested, though.

odanrc:
@powerjg , is it possible to rerun just the failing (download-related) tests?

giactra:
```
In file included from src/base/stats/hdf5.cc:38:                                                                                                                                                                                                                                                                                                                                           
src/base/stats/hdf5.hh:41:10: fatal error: H5Cpp.h: No such file or directory                 
   41 | #include <H5Cpp.h>  
```

It looks like now we are building HDF5 code even if there is no library installed

abmerop:
> ```
> In file included from src/base/stats/hdf5.cc:38:                                                                                                                                                                                                                                                                                                                                           
> src/base/stats/hdf5.hh:41:10: fatal error: H5Cpp.h: No such file or directory                 
>    41 | #include <H5Cpp.h>  
> ```
> 
> It looks like now we are building HDF5 code even if there is no library installed

Also seeing this happen recently

odanrc:
@giactra @abmerop does this change fix it? https://github.com/gem5/gem5/pull/1911

I might have some hidden hdf5 cached somewhere, so I couldn´t reproduce the error, but I know it exists; at some point I installed hdf5 because of this error, but I thought it was before creating this PR, in develop - I guess not.

I also added protobuf to that PR because the old code did not include gem5 lib either - I am not sure it was necessary, because of the following implied tag, but it was the old behavior anyway:
src/proto/SConsopts:74:    main.TagImplies('protobuf', 'gem5 lib')

PR #1439, opened 2024-08-08
sim,mem,cpu,arch-arm: Add support for cache PMU events

arichardson:
See individual commits for details.

This was rather tricky to figure out - I noticed I could not check the dynamic type of the connected probes since we were calling virtual functions on partially constructed objects.

PR #1787, opened 2024-11-18
sim: Remove cyclic dependency of SimObjects and probes

odanrc:
Remove SimObject pointer from ProbeManager, as it is only used to get the SimObject's name for improved debugging messages. This adds an unnecessary dependency, and since the name should not change throughout execution, it can be stored on construction instead by inheriting from the Named class.

Finally, we move the ProbeListenerObject class to its own file, as it was adding a cyclic dependency that complicated creating unit tests for SimObjects.

### DIFF
--- a/src/cpu/simple/probes/looppoint_analysis.hh
+++ b/src/cpu/simple/probes/looppoint_analysis.hh
@@ -36,8 +36,8 @@
 
 // m5 includes
 #include "arch/generic/pcstate.hh"
-#include "cpu/probes/pc_count_pair.hh"
 #include "cpu/simple_thread.hh"
+#include "sim/probe/probe_listener_object.hh"
 #include "sim/sim_exit.hh"
 
 // class includes
--- a/src/cpu/o3/probe/elastic_trace.hh
+++ b/src/cpu/o3/probe/elastic_trace.hh
@@ -59,7 +59,7 @@
 #include "proto/packet.pb.h"
 #include "proto/protoio.hh"
 #include "sim/eventq.hh"
-#include "sim/probe/probe.hh"
+#include "sim/probe/probe_listener_object.hh"
 
 namespace gem5
 {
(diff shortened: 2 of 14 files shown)


giactra:
Just wanted to flag this PR is conflicting with https://github.com/gem5/gem5/pull/1439

It would be nice if we could harmonize between them

BobbyRBruce:
@studyztp : You're my expert on probes. Can you review this change and #1439?

studyztp:
> @studyztp : You're my expert on probes. Can you review this change and #1439?

Sure, I'll review them. For #1439, I was waiting for the update on the PR.

odanrc:
Dropped the commit removing the probe listener object and instead moved it to a separate file. This removed the dependency, and made the PR backwards compatible and trivial. Please review when possible, as this and #1785 are the last steps required to allow unit testing SimObjects.

powerjg (approved):

erin-le:
@studyztp: If you have the time, could you give this PR a review? Thank you!

src/cpu/simple/probes/looppoint_analysis.hh
```diff
@@ -36,8 +36,8 @@
 
 // m5 includes
 #include "arch/generic/pcstate.hh"
-#include "cpu/probes/pc_count_pair.hh"
 #include "cpu/simple_thread.hh"
```
studyztp, line 39: The ```#include "cpu/probes/pc_count_pair.hh"``` is still needed for the PcCountPair type.

studyztp (changes requested):
Thank you for your contribution! I carefully reviewed the changed files, and they look good to me. There is just one small issue with one of the changes involving the include file.

src/cpu/simple/probes/looppoint_analysis.hh
odanrc: There is no instance of PcCountPair in this file

studyztp: You're right. Sorry about this.

studyztp (approved):

BobbyRBruce (approved):

PR #1789, opened 2024-11-18
arch-arm,misc: Fix build errors

BobbyRBruce:
1. Add missing override to `print` function.
2. Change `TlbEntry` to struct in `ArmISA` class.

I honestly don't know what introduced these build errors but this fixes them.

This was found attempting to compile gem5 on MacOS (Apple Silicon) with clang v19.


giactra (approved):
Thanks for fixing this!

PR #1793, opened 2024-11-20
base: Fix failing compiler tests

ivanaamit:
This PR adds `#include <algorithm>` to `src/base/random.cc`. This should fix some of the failing compiler tests.

### DIFF
--- a/src/base/random.cc
+++ b/src/base/random.cc
@@ -42,6 +42,8 @@
 
 #include <sstream>
 
+#include <algorithm>
+
 namespace gem5
 {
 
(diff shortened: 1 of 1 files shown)


src/base/random.cc
```diff
@@ -42,6 +42,8 @@
 
 #include <sstream>
 
+#include <algorithm>
```
powerjg, line 43: I think this should go above `sstream` (in alphabetical order) and shouldn't be separated by spaces.

powerjg (commented):
Tiny, tiny nit

powerjg (approved):

PR #1794, opened 2024-11-20
util: Adding TargetNamedBreakpoint class for gdb.

mahyarsamani:
This change adds a utility script that adds the definition of the TargetNamedBreakpoint class. This class inherits from gdb.Breakpoint and extends the class to allow stopping the execution only if it can extract symbol `_name` from the context of the breakpoint and if `_name` matches a predefined set of exact strings or regular expressions.

### DIFF
--- a/util/gdb_funcs.py
+++ b/util/gdb_funcs.py
@@ -124,7 +124,7 @@ class TargetNamedBreakpoint(gdb.Breakpoint):
         Add a regex match to the list of potential names.
 
         :param regex_match: The pattern to regex_match against upon hitting the breakpoint.
-        E.g. "system.cache_hierarchy.core_clusters(0-9+).dcache" if you want to stop
+        E.g. "system.cache_hierarchy.core_clusters[0-9]+.dcache" if you want to stop
         at all dcache objects for all core_clusters.
         """
         self._regex_matches.add(re.compile(regex_match))
--- a/util/gdb_funcs.py
+++ b/util/gdb_funcs.py
@@ -1,0 +1,167 @@
+# Copyright (c) 2024 The Regents of The University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


mahyarsamani:
@powerjg @BobbyRBruce @studyztp

powerjg (dismissed):
Super cool and great documentation

studyztp (approved):
This will come in really handy! Thanks for upstreaming this.

The only nitpick I see is that maybe it could use sets instead of lists to store the strings. However, I also don't see any significant harm in having duplicates.

studyztp (approved):
Awesome!

util/gdb_funcs.py
```diff
@@ -0,0 +1,167 @@
...
+        only for dcache for core_clusters0.
+        """
+        self._exact_matches.add(exact_match)
+
+    def add_match_regex(self, regex_match: str) -> None:
+        """
+        Add a regex match to the list of potential names.
+
+        :param regex_match: The pattern to regex_match against upon hitting the breakpoint.
+        E.g. "system.cache_hierarchy.core_clusters(0-9+).dcache" if you want to stop
```
erin-le:
```suggestion
        E.g. "system.cache_hierarchy.core_clusters[0-9]+.dcache" if you want to stop
```
Hey Mahyar, sorry about the nitpick. Should this be "system.cache_hierarchy.core_clusters[0-9]+.dcache" instead? I tried testing this change with `python test.add_match_regex("board.processor.cores(0-9+).core.mmu")` (i.e. modifying the example at the top of the file) and didn't hit a breakpoint.

erin-le (commented):
Looks good! There's just have a small nitpick about a documentation comment

util/gdb_funcs.py
powerjg: @mahyarsamani can you take a look at this feedback?

erin-le (approved):
LGTM

PR #1797, opened 2024-11-21
fix delayed execution bug when using classic cache model

2channelkrt:
### 1. what is this bug?

some instructions have unintended 1 cpu cycle delay when being ready for execution when instrucitons are waiting on data load from l1d cache

and for a bonus, fixes a bug where you couldn't run ruby-CHI cache model in develop branch due to wrong class name

### 2. when does it happen?

when multiple instructions are dependent in single cache line load. Only first instructions gets executed while others have unintended 1 cycle delay when cache line is ready.

This happens only when you're using classic cache model. This doesn't happend when using Ruby

### 3. how can i reproduce?

Use `configs/deprecated/example/se.py` and dummy assembly code (aka code) to demonstrate this issue. Running the code with ref cache setting and ruby-CHI setting will show that this issue shows only on ref cache setting, and after applying this PR, issue disappears on ref cache model.

here is the [link](https://github.com/2channelkrt/gem5-bug-report) for the code and script to run ref cache model and ruby-CHI cache model. clone that repo anywhere you like.

`binary/` dir has sample assembly code and compiled binary to be run on simulation to produce the bug. If you look at the assembly code, it's simple: lots of nops, and 3 `ldr` instruction starting at line 3210.

    ldr x1, =DataStart
    ldr x2, =DataStart
    ldr x3, =DataStart

these three loads requests line read on same cache line; and put on hold until data is available. By executing this code and looking at the log, you can see that on ref cache model, only `ldr x1` gets executed and latter two is executed on next cycle while all three `ldr` is executed on same cycle when using ruby-CHI cache model.

`runscript/` dir has the script to run ref cache model (`ref.sh`) and ruby-CHI cache model (`chi.sh`). just change value of `gem5_dir` of each script to match your gem5 directory and run each script once.

When you run each script, you will get the result directory `m5out-ref` and `m5out-chi`, and in each dir, you'll get debug log file `ref-out.txt` and `chi-out.txt` for each run. Open up these debug log, and look for the timestamp below. following is how to read the timestamp

example timestamp:
[sn:105786] 53110400 53157600 118 **// first ldr instruction**
[sn:105787] 53110400 53158000 119 **// second ldr instruction**
[sn:105788] 53110400 53158000 119 **// third ldr instruction**

[sn:105788] : instruction sequence number
first number (53157600) : tick when instruction calls mem read
second number (53157600) : tick when mem read complete and instruction gets executed
third number (118) : number of cpu cycle elapsed (assuming 2.5GHz CPU core clock)

**If you're running on `stable` branch:**

**ref before fix**

[sn:105786] 53110400 53157600 118
[sn:105787] 53110400 53158000 119
[sn:105788] 53110400 53158000 119

**ref after fix**

[sn:105859] 52748000 52795200 118
[sn:105860] 52748000 52795200 118
[sn:105861] 52748000 52795200 118

**chi before, after fix**

[sn:105333] 67030800 67090400 149
[sn:105334] 67030800 67090400 149
[sn:105335] 67030800 67090400 149

**If you're running on `develop` branch:**

**ref before fix**

[sn:105786] 53110400 53157600 118
[sn:105787] 53110400 53158000 119
[sn:105788] 53110400 53158000 119

**ref after fix**

[sn:105859] 52748000 52795200 118
[sn:105860] 52748000 52795200 118
[sn:105861] 52748000 52795200 118

chi before fix(doesn't run because of class naming scheme)

**chi after fix**

[sn:105333] 66974000 67033600 149
[sn:105334] 66974000 67033600 149
[sn:105335] 66974000 67033600 149

as you can see, before fix, first `ldr` instruction from ref-cache model takes 118 cycle while latter two takes 119 cycles. after fix, this bug disappears

### 4. Ok. what is going on here?

following is the simplified timeline of what's happening in the simulation. Let's say 3 `ldr` instructions are instruction a, b, and c.

**before fix**

tick 300

- cache load done. resp latency : 1
- inst a,b,c queued to release at next cpu cycle  + resp latency  = (tick 800)

tick 400 (cpu clock edge)
- cpu executes

tick 800(cpu clock edge)
- inst a released, checking for any more instruction... yes. scheduling inst b at tick 801
- cpu executes, **finds inst a released.**

tick 801
- inst b released, checking for any more instruction... yes. scheduling inst c at tick 802

tick 802
- inst c released. checking for any more instruction... no

tick 1200(cpu clock edge)
- cpu executes, **finds inst b, c released**

==============

**after fix**

tick 300
- cache load done. resp latency : 1
- inst a,b,c queued to release at curTime + resp latency = (tick 700)

tick 400 (cpu clock edge)
- cpu executes

tick 700
- inst a released, checking for any more instruction... yes. scheduling inst b at tick 701

tick 701
- inst b released, checking for any more instruction... yes. scheduling inst c at tick 702

tick 702
- inst c released, checking for any more instruction... no

tick 800(cpu clock edge)
- cpu executes, **finds inst a,b,c released.**

==============

**ruby-cache (resp latency isn't implemented in code, though parameter exists)**

tick 300
- cache load done. resp latency : 1
- inst a,b,c queued to release at next clock edge (tick 400)

tick 400 (cpu clock edge)
- cpu executes
- inst a released, checking for any more instruction... yes. scheduling inst b at tick 401

tick 401
- inst b released, checking for any more instruction... yes. scheduling inst c at tick 402

tick 402
- inst c released, checking for any more instruction... no

tick 800(cpu clock edge)
- cpu executes, **finds inst a,b,c released.**

so basically, order of `releasing the ready instruction` and `cpu execution` is the problem. So I changed the tick timing of the release of instructions, so instructions are released on time regardless of the order. This solution might seem like a XY problem, but all tests passed, and my benchmarks run without problem with meaningful perfomrance difference.

### DIFF
--- a/configs/ruby/CHI_config.py
+++ b/configs/ruby/CHI_config.py
@@ -359,7 +359,7 @@ class CHI_HNFController(Base_CHI_Cache_Controller):
         self.unify_repl_TBEs = False
 
 
-class CHI_MNController(Base_CHI_MiscNode_Controller):
+class CHI_MNController(CHI_MiscNode_Controller):
     """
     Default parameters for a Misc Node
     """
--- a/configs/ruby/CHI_config.py
+++ b/configs/ruby/CHI_config.py
@@ -359,7 +359,7 @@ class CHI_HNFController(Base_CHI_Cache_Controller):
         self.unify_repl_TBEs = False
 
 
-class CHI_MNController(CHI_MiscNode_Controller):
+class CHI_MNController(Base_CHI_MiscNode_Controller):
     """
     Default parameters for a Misc Node
     """
--- a/src/mem/cache/cache.cc
+++ b/src/mem/cache/cache.cc
@@ -807,7 +807,7 @@ Cache::serviceMSHRTargets(MSHR *mshr, const PacketPtr pkt, CacheBlk *blk)
                 // responseLatency is the latency of the return path
                 // from lower level caches/memory to an upper level cache or
                 // the core.
-                completion_time += clockEdge(responseLatency) +
+                completion_time += curTick() + responseLatency +
                     (transfer_offset ? pkt->payloadDelay : 0);
 
                 assert(!tgt_pkt->req->isUncacheable());
--- a/src/mem/cache/cache.cc
+++ b/src/mem/cache/cache.cc
@@ -807,7 +807,7 @@ Cache::serviceMSHRTargets(MSHR *mshr, const PacketPtr pkt, CacheBlk *blk)
                 // responseLatency is the latency of the return path
                 // from lower level caches/memory to an upper level cache or
                 // the core.
-                completion_time += curTick() + responseLatency +
+                completion_time += curTick() + responseLatency * clockPeriod() +
                     (transfer_offset ? pkt->payloadDelay : 0);
 
                 assert(!tgt_pkt->req->isUncacheable());


BobbyRBruce (dismissed):
I think this makes sense but it's a cautious approval as I'm not an expert on the caches.

giactra:
Thanks @2channelkrt for this fix. Can I ask you to split the changes in two separate PRs?

The CHI change definitely needs to go as soon as possible in next release

BobbyRBruce:
> Thanks @2channelkrt for this fix. Can I ask you to split the changes in two separate PRs?
> 
> The CHI change definitely needs to go as soon as possible in next release

It's two commits. Is that ok? 

I can cherry-pick the CHI one when merged

giactra:
> > Thanks @2channelkrt for this fix. Can I ask you to split the changes in two separate PRs?
> > The CHI change definitely needs to go as soon as possible in next release
> 
> It's two commits. Is that ok?
> 
> I can cherry-pick the CHI one when merged

I haven't read the problem description in full, but the classic cache change looks wrong to me.

`completion_time += curTick() + responseLatency`

Here we are adding Tick to cycles. This is wrong. clockEdge was rightfully transforming the latency value (cycles) into the corresponding tick number. So that's why I believe we should split this in two; the CHI change is straightforward and fixes CHI in the staging branch.

We need extra time to review the problem description for the classic bit

2channelkrt:
Hi @BobbyRBruce and @giactra! thanks for the review.

> Here we are adding Tick to cycles. This is wrong. clockEdge was rightfully transforming the latency value (cycles) into the corresponding tick number.

Original code was transforming latency into clockEdge and queueing data-ready instructions to mark them ready, and this can be seen as a right way to go, but 'marking the instruction ready' and 'executing CPU tick' was done in different order between 'classic cache' and 'ruby cache' caused the delay. So queueing data-ready instrucitons at tick rather than clockEdge solves this issue for most case.

For the case where cache line is ready on the cpu clockEdge, issue would still occur, as they will be ready on clockEdge even if we use tick() to queue instructions to mark them ready.

2channelkrt:
i will add a new commit which reverts ruby-CHI naming and make new PR just for that.

2channelkrt:
now this pr is solely about delayed execution.

PR #1817 was open for wrong ruby-CHI class name

giactra:
> Hi @BobbyRBruce and @giactra! thanks for the review.
> 
> > Here we are adding Tick to cycles. This is wrong. clockEdge was rightfully transforming the latency value (cycles) into the corresponding tick number.
> 
> Original code was transforming latency into clockEdge and queueing data-ready instructions to mark them ready, and this can be seen as a right way to go, but 'marking the instruction ready' and 'executing CPU tick' was done in different order between 'classic cache' and 'ruby cache' caused the delay. So queueing data-ready instrucitons at tick rather than clockEdge solves this issue for most case.
> 
> For the case where cache line is ready on the cpu clockEdge, issue would still occur, as they will be ready on clockEdge even if we use tick() to queue instructions to mark them ready.

The fact that ruby and classic caches' behaviour differ does not surprise me. Ultimately it's a matter of understanding which model resembles the operation of a real world cache. To this I have little knowledge but it does not look unreasonable to me to respond to multiple memory operation targeting the same line in the same cycle

However it looks to me you are just tampering the event scheduling to make things work. But effectively it is just a coincidence and you are doing so adding two values which should not be added together.

Look at the snipper I posted before:

`completion_time += curTick() + responseLatency`

responseLatency is defined as being number of Cycles. You cannot add Cycles to Ticks (the output of curTick()); that 's like adding apples and pears.

Tick is the atomic unit of measure of gem5 time. Everything is scheduled according to ticks. Cycles are made of a number of ticks (>1) and that number depends on the frequency of the object (you can have multiple clock domains)

So for example

[ _ _ _ _ _ _ _ _ _ _] = 10 Ticks

[ _ _ ] = 1 Cycle = 2 Ticks for a 2 GHz CPU

[ _ _ _ _ ] = 1 Cycle = 4 Ticks for a 1 GHz CPU

Ok, so the moment a user defines responseLatency as number of Cycles, you cannot directly add that value to a Tick number because it will not give the expected results. In the above case, say you model an operation that will take 2 cycles.

[ _ _ _ ] = 1 Cycle = 3 Ticks

x = curTickTime();

Simulation time: [ _ _ x _ _ _ _ _ _ _ _ _ _ _ _ _]

If you want the operation to happen in 2 cycles, you cannot do

y = x + 2
Simulation time: [ _ _ x _ y _ _ _ _ _ _ _ _ _ _ _]

You need to do

y = x + clockEdge(2) = x + 3 * 2 = x + 6
Simulation time: [ _ _ x _ _ _ _ _ y _ _ _ _ _ _ _]

Hope this helps.

2channelkrt:
I got the point. When i tested it with resp latency 1 it worked as intended (1 cpu clock cycle, 400ticks for 2.5Ghz) but i will double check with other values. 

Even if other value works as intended, adding tick and latency without any explicit conversion seem absurd to me too, this i will add some conversion method for resp latency for better readability and make sure the value is synched to the clock.

2channelkrt:
i confirmed the code is wrong.
when curtick() is 500 and resplatency is 5 while cpu clock is 2.5GHz, my intended comp_time is 500 + (5 * 400) = 2500 but now it's printing 505. I will push fix for this.

2channelkrt:
fixed. this is what I intended to do.

giactra:
> fixed. this is what I intended to do.

The new version is effectively similar to [clockEdge](https://github.com/gem5/gem5/blob/stable/src/sim/clocked_object.hh#L183)

    Tick
    clockEdge(Cycles cycles=Cycles(0)) const
    {
        // align tick to the next clock edge
        update();

        // figure out when this future cycle is
        return tick + clockPeriod() * cycles;
    }

The only difference is the update method which is aligning the computation starting point to the next clock edge, if we are not (curTick) on a clock edge. This makes sense as it assumes changes (triggering an elaboration) will be made visible to the logic only on a clock edge (and that should be the starting count on the latency).

I fail to understand the logical connection between this line change and what you are effectively trying to do. Unless somehow calling ClockedObject::update over different MSHR targets yields to different completion times.
But as far as I can see from the code, this shouldn't be the case.

I believe you should make sure (debugging is a solution) the PR is not fixing your problem coincidentally/indirectly. The starting point shouldn't be: I want the CPU to get the data at this tick time and I am going to scram the timing of the cache model so that this condition is satisfied. The starting point should rather be: I want to be able to service multiple targets at the same cycle. Why is this not happening?

Also, no need to act on this now, but it would be nice if you could squash all your commits into a single one and force push the PR ;)

2channelkrt:
the motivation for this fix was that only first instruction were applied correct latency while other instructions were having unintended +1 cycle latency on ref cache mode. Since the number of register gets free in single cycle isn't parametized, and behaves differently between classic cache and ruby cache, I decided to fix it since this caused up to 10% ipc difference on the benchmarks I tested.

I found this issue while debugging instruction execution timing between classic cache and ruby cache, where they had different timing result despite having same latency config.

I provided simulation setting to reproduce the issue, and while I didn't provide the log itself, I wrote what was happening in the first comment. This issue is mainly related to the line [here](https://github.com/gem5/gem5/blob/ce4033e36fa854b41fa566c27de3c24bd5e7ff5f/src/mem/packet_queue.cc#L167)

```
        when = std::max(when, curTick() + 1);
```
when going over mshr slot, gem5 adds extra tick per slot to free the register dependant on it.

So i came up with the fix to desync between the timing of the MSHR release and cpu clock edge, so adding +1 tick on the MSHR release timing doesn't happen crossing over cpu clock edge on most part.

erin-le:
This PR is being closed for inactivity, but feel free to reopen it if you would like to work on it further!

PR #1801, opened 2024-11-22
tests: add non KVM CPU type switch tests

### LINKED ISSUE #483: Add tests to test CPU type switching for each ISA
I am submitting this issue in response to the observation made in #451, which was subsequently resolved in #479. The problem pertains to the full system RISC-V simulation experiencing stalls when transitioning the CPU type from ATOMIC to TIMING.

It will be beneficial to incorporate tests for CPU type transitions across all ISAs, as this will help identify and address any existing or potential issues in similar scenarios.

erin-le:
This commit adds tests for non KVM type switching. The CPU switches are atomic to timing, atomic to minor, atomic to o3, timing to minor, and timing to o3. There are at least 3 tests for each type of CPU switch, and the original plan was to only have 3 tests for each type, to avoid adding too many tests. Each set has one test for each ISA out of X86, ARM, and RISCV, and uses 1, 2, or 4 cores. Two tests of each set use the classic memory system and the remaining test uses MESI two level. I tried to shuffle around the combinations of ISAs, cores, and memory system to maximize the range of what was tested without using too many tests, but I'm not sure if the current set of tests does this well. Feedback on this would be helpful @BobbyRBruce!

### DIFF
--- a/tests/gem5/non_kvm_switch_tests/README.md
+++ b/tests/gem5/non_kvm_switch_tests/README.md
@@ -1,0 +1,8 @@
+# Non-KVM Switch Tests
+
+These tests ensure that gem5 can switch processors during simulation.
+To run these tests by themselves, you can run the following command in the tests directory:
+
+```bash
+./main.py run gem5/non_kvm_switch_tests
+```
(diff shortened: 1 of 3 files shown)


erin-le:
Closing in favor of https://github.com/gem5/gem5/pull/2301

PR #1804, opened 2024-11-24
arch-arm: Fix bug in VQRSHL.

giactra:
If shiftAmt is 0, bits raise assert, causing core dump.

Change-Id: Ic4285f51a866ffc017645655e98674ca69a15a40


powerjg (approved):
@BobbyRBruce this should be in 24.1

PR #1817, opened 2024-11-28
fix wrong ruby-CHI base class name

2channelkrt:
fix ruby-CHI base class name so it actually runs

previously was combined with PR #1797


powerjg (approved):

giactra:
This seems to be failing some tests?

powerjg:
Looks like an issue with the runners. I just re-ran it. Can you access the button "re-run failing tests"? when you go to the "details" for the test?

PR #1823, opened 2024-11-29
arch-arm: Fix inconsistency of rint().

giactra:
The behavior of rint() is different due to compiler optimizations. In vfp.hh, the output of rint() is correct in -O0, but error in -O3.

To fix this, vfpFpToFixed and vfpFpRint is forced to use -O0 by pragma.

Change-Id: I753aeb3aa61213ecbcc1009bb58d353a3863bf6e


powerjg (approved):
Lgtm. I hope that all of our supported compilers behave the same way.

PR #1829, opened 2024-12-02
mem-ruby, scons: Add ProtocolInfo.hh files in build targets

Harshil2107:
- In the new MultiRuby system, the generated ProtocolInfo header files were not being correctly added to the build targets in SCons.

- As a result, when building gem5 with the --duplicate-sources option, these files were mistakenly deleted by SCons. This happened because SCons treated them as source files instead of generated build targets.

- This commit ensures that the ProtocolInfo header files are explicitly included in the build targets, preventing their unintended removal and fixing the build issue.

### DIFF
--- a/src/mem/ruby/protocol/SConscript
+++ b/src/mem/ruby/protocol/SConscript
@@ -91,6 +91,14 @@ def slicc_emitter(target, source, env):
 
         files.update([output_dir.File(f) for f in sorted(slicc.files())])
 
+        # Dynamically determine protocol and add ProtocolInfo.hh to the list of
+        # files to be built
+        protocol_name = os.path.splitext(os.path.basename(filepath))[0]
+        protocol_file = output_dir.File(
+            f"{protocol_name}/{protocol_name}ProtocolInfo.hh"
+        )
+        files.update([protocol_file])
+
     return list(files), source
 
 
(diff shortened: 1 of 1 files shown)


Harshil2107:
This change should fix the failing sst and systemc daily tests.

src/mem/ruby/protocol/SConscript
```diff
@@ -91,6 +91,14 @@ def slicc_emitter(target, source, env):
 
         files.update([output_dir.File(f) for f in sorted(slicc.files())])
 
+        # Dynamically determine protocol and add ProtocolInfo.hh to the list of
+        # files to be built
+        protocol_name = os.path.splitext(os.path.basename(filepath))[0]
+        protocol_file = output_dir.File(
+            f"{protocol_name}/{protocol_name}ProtocolInfo.hh"
+        )
+        files.update([protocol_file])
```
BobbyRBruce, line 99: are you sure there isn't a parameter or something you can access in scons that can tell you what these particles are instead of doing this string parsing of the paths?

BobbyRBruce (changes requested):

src/mem/ruby/protocol/SConscript
Harshil2107: I updated the code to use `slicc.protocol` variable.

BobbyRBruce (approved):

PR #1831, opened 2024-12-02
arch-riscv: Add senvcfg CSR

### LINKED ISSUE #1674: Illegal instruction with 6.11.3 Kernel
**Describe the bug**
When attempting to create a RISCV version of the Linux Kernel v6.11.3, the kernel panics due to an illegal instruction.
 
**Affects version**
gem5 24.0.0.1

**gem5 Modifications**
No modifications have been made.

**To Reproduce**
1. Follow instructions in the gem5-resources repository for RISCV kernel creation, found here: https://github.com/gem5/gem5-resources/tree/stable/src/riscv-fs/riscv-opensbi.
2. Instead of using kernel v6.5.5 use kernel v6.11.3, i.e. `wget https://cdn.kernel.org/pub/linux/kernel/v6.x/linux-6.11.3.tar.xz`
3. Modify `riscv-ubuntu-run.py` to use our built kernel and both disk image and bootloader provided by gem5 resources:
```python
    kernel=KernelResource("/home/darchr-admin/gem5-resources/src/riscv-fs/riscv-opensbi/linux-6.11.3/vmlinux"),
    bootloader=obtain_resource(resource_id="riscv-bootloader-opensbi-1.3.1"),
    disk_image=obtain_resource(resource_id="riscv-ubuntu-24.04-img"),
)
```
4. Compile gem5 (`scons build/ALL/gem5.opt`) and run the script (`build/ALL/gem5.opt ./configs/example/gem5_library/riscv-ubuntu-run.py`)

**Terminal Output**
```shell
[0.000000] Linux version 6.11.3 (@bc2bd15b0338) (riscv64-linux-gnu-gcc (Ubuntu 12.3.0-1ubuntu1~22.04) 12.3.0, GNU ld (GNU Binutils for Ubuntu) 2.38) #1 SMP Mon Oct 14 23:42:10 UTC 2024
[0.000000] Machine model: riscv-virtio
[0.000000] SBI specification v1.0 detected
[0.000000] SBI implementation ID=0x1 Version=0x10003
[0.000000] SBI TIME extension detected
[0.000000] SBI IPI extension detected
[0.000000] SBI RFENCE extension detected
[0.000000] efi: UEFI not found.
[0.000000] OF: reserved mem: 0x0000000080000000..0x000000008001ffff (128 KiB) nomap non-reusable mmode_resv0@80000000
[0.000000] OF: reserved mem: 0x0000000080020000..0x000000008003ffff (128 KiB) nomap non-reusable mmode_resv1@80020000
[0.000000] Zone ranges:
[0.000000]   DMA32    [mem 0x0000000080000000-0x00000000ffffffff]
[0.000000]   Normal   [mem 0x0000000100000000-0x000000013fffffff]
[0.000000] Movable zone start for each node
[0.000000] Early memory node ranges
[0.000000]   node   0: [mem 0x0000000080000000-0x000000008003ffff]
[0.000000]   node   0: [mem 0x0000000080040000-0x000000013fffffff]
[0.000000] Initmem setup node 0 [mem 0x0000000080000000-0x000000013fffffff]
[0.000000] SBI HSM extension detected
[0.000000] Falling back to deprecated "riscv,isa"
[0.000000] riscv: base ISA extensions acdfimv
[0.000000] riscv: ELF capabilities acdfimv
[0.000000] Oops - illegal instruction [#1]
[0.000000] Modules linked in:
[0.000000] CPU: 0 UID: 0 PID: 0 Comm: swapper Not tainted 6.11.3 #1
[0.000000] Hardware name: riscv-virtio (DT)
[0.000000] epc : riscv_user_isa_enable+0x26/0x30
[0.000000]  ra : setup_arch+0x5e4/0x602
[0.000000] epc : ffffffff800031a2 ra : ffffffff80a04566 sp : ffffffff81403df0
[0.000000]  gp : ffffffff81513508 tp : ffffffff8140cc80 t0 : 0000000000000061
[0.000000]  t1 : 0000000000000073 t2 : 2073656974696c69 s0 : ffffffff81403e00
[0.000000]  s1 : 0000000000000006 a0 : 0000000000000001 a1 : 000000000000001f
[0.000000]  a2 : 0000000000000be8 a3 : 0000000000000be8 a4 : 0000000000000001
[0.000000]  a5 : 0000000000000080 a6 : 0000000000406000 a7 : 0000000000000003
[0.000000]  s2 : ffffffff81491808 s3 : ffffffd6beff4dc0 s4 : 0000000000000240
[0.000000]  s5 : ffffffff8154f188 s6 : 0000000081000200 s7 : 0000000000000fff
[0.000000]  s8 : 0000000008000200 s9 : ffffffff80ff0690 s10: 0000000000000018
[0.000000]  s11: 0000000000000007 t3 : ffffffff81528cb7 t4 : ffffffff81528cb7
[0.000000]  t5 : ffffffff81528c98 t6 : ffffffff814039e8
[0.000000] status: 0000000200000100 badaddr: 0000000010a7a073 cause: 0000000000000002
[0.000000] [<ffffffff800031a2>] riscv_user_isa_enable+0x26/0x30
[0.000000] [<ffffffff80a04566>] setup_arch+0x5e4/0x602
[0.000000] [<ffffffff80a007ba>] start_kernel+0x86/0x7dc
[0.000000] Code: 0151 0713 e847 97ba 639c 9389 8b85 c789 0793 0800 (a073) 10a7
[0.000000] ---[ end trace 0000000000000000 ]---
```

**Expected behavior**
A complete boot of Ubuntu.

**Host Operating System**
Ubuntu 24.04

**Host ISA**
X86

**Compiler used**
gcc 13.2.0

erin-le:
This commit adds the senvcfg control status register, which fixes the 6.11.3 kernel crash documented in issue 1674. I have not added a bitfield and its implementation in isa.cc only uses setMiscRegNoEffect, so this implementation is missing some components.

### DIFF
--- a/src/arch/riscv/isa.cc
+++ b/src/arch/riscv/isa.cc
@@ -185,6 +185,7 @@ namespace RiscvISA
     [MISCREG_SCAUSE]        = "SCAUSE",
     [MISCREG_STVAL]         = "STVAL",
     [MISCREG_SATP]          = "SATP",
+    [MISCREG_SENVCFG]       = "SENVCFG",
 
     [MISCREG_UTVEC]         = "UTVEC",
     [MISCREG_USCRATCH]      = "USCRATCH",
@@ -777,6 +778,11 @@ ISA::setMiscReg(RegIndex idx, RegVal val)
                     new_val.mode != AddrXlateMode::SV39)
                     new_val.mode = cur_val.mode;
                 setMiscRegNoEffect(idx, new_val);
+            }
+            break;
+          case MISCREG_SENVCFG:
+            {
+                setMiscRegNoEffect(idx, val + 1);
             }
             break;
           case MISCREG_TSELECT:
--- a/src/arch/riscv/isa.cc
+++ b/src/arch/riscv/isa.cc
@@ -797,12 +797,7 @@ ISA::setMiscReg(RegIndex idx, RegVal val)
                     val);
                 }
 
-                if ((wpri_mask & val) != 0) {
-                    warn("Ignoring write to WPRI bit(s) in senvcfg CSR.\n"
-                    "The attempted write was:\n %" PRIu64 "\n", val);
-                } else {
-                    setMiscRegNoEffect(idx, val);
-                }
+                setMiscRegNoEffect(idx, val & ~wpri_mask);
             }
             break;
           case MISCREG_TSELECT:
--- a/src/arch/riscv/isa.cc
+++ b/src/arch/riscv/isa.cc
@@ -782,7 +782,27 @@ ISA::setMiscReg(RegIndex idx, RegVal val)
             break;
           case MISCREG_SENVCFG:
             {
-                setMiscRegNoEffect(idx, val + 1);
+                // panic on write to bitfields that aren't implemented in gem5
+                SENVCFG panic_mask = 0;
+                panic_mask.pmm = 3;
+
+                SENVCFG wpri_mask = 0;
+                wpri_mask.wpri_1 = ~wpri_mask.wpri_1;
+                wpri_mask.wpri_2 = ~wpri_mask.wpri_2;
+                wpri_mask.wpri_3 = ~wpri_mask.wpri_3;
+
+                if ((panic_mask & val) != 0) {
+                    panic("Tried to write to an unimplemented bitfield in the "
+                    "senvcfg CSR!\nThe attempted write was:\n %" PRIu64 "\n",
+                    val);
+                }
+
+                if ((wpri_mask & val) != 0) {
+                    warn("Ignoring write to WPRI bit(s) in senvcfg CSR.\n"
+                    "The attempted write was:\n %" PRIu64 "\n", val);
+                } else {
+                    setMiscRegNoEffect(idx, val);
+                }
             }
             break;
           case MISCREG_TSELECT:
--- a/src/arch/riscv/regs/misc.hh
+++ b/src/arch/riscv/regs/misc.hh
@@ -1265,6 +1265,20 @@ BitUnion64(INTERRUPT)
     Bitfield<0> usi;
 EndBitUnion(INTERRUPT)
 
+
+// From the RISCV specification version 20240411, volume 2,
+// section 10.1.10, page 98
+BitUnion64(SENVCFG)
+    Bitfield<63,34> wpri_1;
+    Bitfield<33,32> pmm;
+    Bitfield<31,8> wpri_2;
+    Bitfield<7> cbze;
+    Bitfield<6> cbcfe;
+    Bitfield<5,4> cbie;
+    Bitfield<3,1> wpri_3;
+    Bitfield<0> fiom;
+EndBitUnion(SENVCFG)
+
 const off_t MXL_OFFSETS[enums::Num_RiscvType] = {
     [RV32] = (sizeof(uint32_t) * 8 - 2),
     [RV64] = (sizeof(uint64_t) * 8 - 2),
--- a/src/arch/riscv/regs/misc.hh
+++ b/src/arch/riscv/regs/misc.hh
@@ -183,6 +183,7 @@ enum MiscRegIndex
     MISCREG_SCAUSE,
     MISCREG_STVAL,
     MISCREG_SATP,
+    MISCREG_SENVCFG,
 
     MISCREG_UTVEC,
     MISCREG_USCRATCH,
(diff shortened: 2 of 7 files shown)


BobbyRBruce (dismissed):
I'm very impressed you figured this out.

src/arch/riscv/isa.cc
```diff
@@ -779,6 +780,11 @@ ISA::setMiscReg(RegIndex idx, RegVal val)
                 setMiscRegNoEffect(idx, new_val);
             }
             break;
+          case MISCREG_SENVCFG:
+            {
+                setMiscRegNoEffect(idx, val + 1);
```
powerjg: We should probably make this behave correctly before committing. I'll see if I have time today to look at it.

powerjg (changes requested):

powerjg:
Specific things that should change before merging:
1. We should only support writing FIOM, not the other bitfileds. If the other bitfileds (i.e., the bitfields that the manual says "will be furnished by the forthcoming extension...") are written we should panic with an appropriate error so we can fix it later. We want to panic here so that in the future if the kernel (or some other application) writes these bits we know how to implement it. Or, maybe we'll chose to ignore it, but let's have our future selves make that decision.
2. This should probably not be a `setMiscRegNoEffect` but we should use the normal `setMiscReg` function (I believe). 
3. We should make sure to ignore the writes to any of the bits marked `WPRI`. Don't panic in this case, but ignore the writes.

I don't believe we need to change the implementation of atomics or fence instructions because I believe they already have the same ordering for memory and I/O space implied. I.e., gem5 takes an conservative approach. However, we *might* need to read the `senvcfg` in the fence and atomic instruction implementations. I think that we can skip that for now, but it is a true dependence that may need to be enforced.

Thanks for picking this up Erin, and great job getting started on it! Everything that I said above you couldn't figure out without a lot more background on RISC-V and gem5's implementation. Again, great job!

Edit: I was wrong about what to do with `wpri` fields.

erin-le:
@rogerchang23424 Feedback on this would be greatly appreciated!

erin-le:
@powerjg: I implemented the bitfields according to the RISCV spec and added panics/warnings based on what you said above, but it looks like the instruction that accesses senvcfg tries to write to the CBZE bitfield, which causes a panic. Should the panic be a warning instead?

powerjg:
> it looks like the instruction that accesses senvcfg tries to write to the CBZE bitfield

Then the panic worked as expected! Let's figure out how to implement that :)

Here's the definition: https://github.com/riscv/riscv-CMOs/blob/8e8a8e01e1e2761579eaaa0f1c0f0c1efac9eddb/cmobase/csr_state.adoc#L51

src/arch/riscv/isa.cc
```diff
@@ -779,6 +780,31 @@ ISA::setMiscReg(RegIndex idx, RegVal val)
...
+                wpri_mask.wpri_3 = ~wpri_mask.wpri_3;
+
+                if ((panic_mask & val) != 0) {
+                    panic("Tried to write to an unimplemented bitfield in the "
+                    "senvcfg CSR!\nThe attempted write was:\n %" PRIu64 "\n",
+                    val);
+                }
+
+                if ((wpri_mask & val) != 0) {
+                    warn("Ignoring write to WPRI bit(s) in senvcfg CSR.\n"
```
powerjg: Do we print a warning in other cases of wrpi writes? If so, we should keep this. If not, we should not have this warning.

src/arch/riscv/isa.cc
```diff
@@ -779,6 +780,31 @@ ISA::setMiscReg(RegIndex idx, RegVal val)
...
+                    "The attempted write was:\n %" PRIu64 "\n", val);
+                } else {
+                    setMiscRegNoEffect(idx, val);
```
powerjg: Can you simply call `setMiscReg`? I think we want the effect.

powerjg (commented):
One more small thing. Great job!!

src/arch/riscv/isa.cc
erin-le: It doesn't look like there's a warning in other cases, so I've removed this

src/arch/riscv/isa.cc
erin-le: I've left this call as setMiscRegNoEffect, as this part of the code is already in setMiscReg. It's just used as a setter function here with the code above being the effect. I think that's how the other registers also use setMiscRegNoEffect.

I've also changed the implementation so `val` excludes the WPRI bits.

powerjg (approved):

src/arch/riscv/regs/misc.hh
```diff
@@ -183,6 +183,7 @@ enum MiscRegIndex
     MISCREG_SCAUSE,
     MISCREG_STVAL,
     MISCREG_SATP,
+    MISCREG_SENVCFG,
```
rogerchang23424, line 186: We forgot adding extra 0 for `MISCREG_SENVCFG` in `util/cpt_upgrader`. We should upgrade the checkpoint to avoid inconsistent miscreg size between old and new checkpoint version.

PR #1832, opened 2024-12-03
arch-riscv: fix tlb stats in timming mode

clemdiep:
The previous #484 issue reported a bug where the TLB stats on RISC-V were incremented twice on misses by calling the `lookup` function twice with hidden argument set to `false`. The fix is only applied on atomic mode as the `translation` argument of `doTranslate` will not be `nullptr` in timing mode.

In that case, if the TLB lookup miss, the `doTranslate` function will start the walker and then return without doing anything more. Then later, when the pagetable walker found the corresponding PTE, it will insert it and call `translateWithTLB`. This function then call `lookup` again which will hit in any case (and crash if not due to the following
assert), but the hit count is incremented here too. 

This commit fix by setting the `hidden` argument of `lookup` to true.

### DIFF
--- a/src/arch/riscv/pagetable_walker.cc
+++ b/src/arch/riscv/pagetable_walker.cc
@@ -574,8 +574,8 @@ Walker::WalkerState::recvPacket(PacketPtr pkt)
              */
             Addr vaddr = req->getVaddr();
             vaddr = Addr(sext<VADDR_BITS>(vaddr));
-            Addr paddr = walker->tlb->translateWithTLB(vaddr, satp.asid,
-                                                       satp.mode, mode);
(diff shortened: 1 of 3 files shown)


powerjg (changes requested):
Thanks for the contribution!

I think this looks like a good fix. However, now the name `translateWithTLB` is misleading. I think we should call it `hiddenTranslateWithTLB` if the access is always hidden.

If you can make this change by tomorrow, we will merge this into the upcoming 24.1 release. Thanks!

clemdiep:
Thanks for you review!

I did the change at all necessary places and rebased on top of develop, so it should be good to go.

powerjg (approved):
Thanks for the updates!

BobbyRBruce (approved):

PR #1834, opened 2024-12-03
base,arch-arm: Add GEM5_NO_OPTIMIZE; use in ARM's vfp.hh

BobbyRBruce:
GCC and CLANG have different annotations for declaring code should not
be optimized. Adding GEM5_NO_OPTIMZE provides gem5 developers a MACRO
that works in both cases.

This change replaces the GCC pragmas in vfp.hh with GEM5_NO_OPTIMIZE
as this solution didn't work with clang.

### DIFF
--- a/src/arch/arm/insts/vfp.hh
+++ b/src/arch/arm/insts/vfp.hh
@@ -264,11 +264,9 @@ setFPExceptions(int exceptions) {
     feraiseexcept(exceptions);
 }
 
-#pragma GCC push_options
-#pragma GCC optimize ("O0")
-
 template <typename T>
 uint64_t
+GEM5_NO_OPTIMIZE
 vfpFpToFixed(T val, bool isSigned, uint8_t width, uint8_t imm, bool
              useRmode = true, VfpRoundingMode roundMode = VfpRoundZero,
              bool aarch64 = false)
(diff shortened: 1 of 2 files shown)


giactra:
Thanks Bobby for fixing this issue! Oddly speaking the GCC pragma is compiling on my machine with clang.. But I presume it depends on the compiler version

As disabling this for clang will mean gcc and clang will lead to different simulated outcomes, may I propose to implement the same logic in clang instead?

https://releases.llvm.org/12.0.0/tools/clang/docs/LanguageExtensions.html#extensions-for-selectively-disabling-optimization

We could either add the file/range based clang pragma; alternatively we could use the gcc and clang per function attributes. For example we could define a GEM5_NO_OPTIMIZE in compiler.hh and use

`__attribute__((optimize("O0")))` for gcc
`[[clang::optnone]]` for clang

Please let me know what you think of this

BobbyRBruce:
> Thanks Bobby for fixing this issue! Oddly speaking the GCC pragma is compiling on my machine with clang.. But I presume it depends on the compiler version
> 
> As disabling this for clang will mean gcc and clang will lead to different simulated outcomes, may I propose to implement the same logic in clang instead?
> 
> https://releases.llvm.org/12.0.0/tools/clang/docs/LanguageExtensions.html#extensions-for-selectively-disabling-optimization
> 
> We could either add the file/range based clang pragma; alternatively we could use the gcc and clang per function attributes. For example we could define a GEM5_NO_OPTIMIZE in compiler.hh and use
> 
> `__attribute__((optimize("O0")))` for gcc `[[clang::optnone]]` for clang
> 
> Please let me know what you think of this

I think I've updated it using your approach. Let me know if it's suitable.

giactra (approved):
Thanks Bobby; it looks perfect to me

PR #1842, opened 2024-12-06
mem-ruby,misc: Fix RNG range

BobbyRBruce:
This upper range must be `UINT_MAX - 1`. This was previously fixed but reverted back. Without this the RNG crashes.


powerjg (approved):

abmerop (approved):
Thanks! I didn't notice the reversion.

mattsinc (approved):

PR #1845, opened 2024-12-07
mem-ruby: Fix Atomic transitions in VIPER protocol

v-ramadas:
This PR fixes two bugs in the GPU VIPER protocol when globally coherent atomic requests are issued. First, a segfault happens when the cache line in TCC moves from V/I to a transient state but later runs transitions back to I. The segfault happens because the pa_performAtomic action tries to access a request queue that has already dequeued the message. The fix changes the action to use the TBE entry instead of peeking into the queue. If the cacheline does not allocate a TBE or move to a transient state, it performs the peek instead. The second fix completes the cache line's transition to I in the TCP upon receiving an ack from the TCC when it is configured to be in write-back mode

### DIFF
--- a/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-TCC.sm
@@ -933,14 +933,18 @@ machine(MachineType:TCC, "TCC Cache")
   }
 
   action(pa_performAtomic, "pa", desc="Perform atomic") {
-    peek(coreRequestNetwork_in, CPURequestMsg) {
-      if ((is_valid(tbe) && tbe.atomicDataReturn) || in_msg.Type == CoherenceRequestType:AtomicReturn) {
-        cache_entry.DataBlk.atomicPartial(cache_entry.DataBlk, cache_entry.writeMask, false);
-      } else {
-        // Set the isAtomicNoReturn flag to ensure that logs are not
-        // generated erroneously
-        assert((is_valid(tbe) && tbe.atomicDataNoReturn) || in_msg.Type == CoherenceRequestType:AtomicNoReturn);
-        cache_entry.DataBlk.atomicPartial(cache_entry.DataBlk, cache_entry.writeMask, true);
+    if (is_valid(tbe) && tbe.atomicDataReturn) {
+       cache_entry.DataBlk.atomicPartial(cache_entry.DataBlk, cache_entry.writeMask, false);
+    } else if (is_valid(tbe) && tbe.atomicDataNoReturn) {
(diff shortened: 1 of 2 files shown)


BobbyRBruce (dismissed):

powerjg (dismissed):
I'll approve, as I assume this has been carefully tested. However, it looks like this now is missing the condition for then the action is executed on the request path. Have you seen the request transitions executed correctly?

v-ramadas:
@powerjg Thanks for catching this! The test I used to verify this executed the transitions that used to have the segfault. I fixed the action to work for both conditions now and tested that it works.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -933,14 +933,18 @@ machine(MachineType:TCC, "TCC Cache")
...
   action(pa_performAtomic, "pa", desc="Perform atomic") {
-    peek(coreRequestNetwork_in, CPURequestMsg) {
-      if ((is_valid(tbe) && tbe.atomicDataReturn) || in_msg.Type == CoherenceRequestType:AtomicReturn) {
-        cache_entry.DataBlk.atomicPartial(cache_entry.DataBlk, cache_entry.writeMask, false);
-      } else {
-        // Set the isAtomicNoReturn flag to ensure that logs are not
-        // generated erroneously
-        assert((is_valid(tbe) && tbe.atomicDataNoReturn) || in_msg.Type == CoherenceRequestType:AtomicNoReturn);
-        cache_entry.DataBlk.atomicPartial(cache_entry.DataBlk, cache_entry.writeMask, true);
+    if (is_valid(tbe) && tbe.atomicDataReturn) {
```
mattsinc, line 936: It would be better if we added comments explaining each of these cases, but I'm fine having that as a separate patch if it means this goes into the stable release as is

mattsinc (approved):

src/mem/ruby/protocol/GPU_VIPER-TCP.sm
abmerop, line 1: I believe this was last changed to allow pipelining atomics (i.e., don't stall an atomic because one is in flight, allow multiple in flight). Does this change break that?

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -933,14 +933,18 @@ machine(MachineType:TCC, "TCC Cache")
   }
 
   action(pa_performAtomic, "pa", desc="Perform atomic") {
```
abmerop, line 935: Which transition were you seeing this being called after the queue was popped? I don't see on in my code. If it's for some research code that's fine. Generally the pop is the last action though.

PR #1849, opened 2024-12-09
sim: Update multisim to use Process instead of Pool

### LINKED ISSUE #1738: Stats.txt file being wiped/overriden when using the MultiSim module when the # of jobs >4x the number of processes available.
**Describe the bug**

In instances where the number of jobs exceeds four times the number of available processes, the MultiSim module overwrites/wipes the stats.txt of some jobs.

**Affects version**

At the time of writing, this bug affects MultiSim and has done since its incorporation into the simulator.

**To Reproduce**

Attached is a zip archive containing two files that you can use to reproduce the tests I ran to investigate this issue. Simply extract the two files and place them in the root directory of the gem5 repository. Then execute the "run_tests.sh" script.

[multisim-empty-stat-repro-package.zip](https://github.com/user-attachments/files/17583397/multisim-empty-stat-repro-package.zip)

These tests generate a cross-product of simulations differing in the number of available processes and jobs for gem5 MultiSim runs: 1 to 10 processes, each with simulations for 1 to 10 jobs, and then , for each job,verifying the existence of the stat.txt file and that it contained data. In other words, the tests execute the following commands "pseudocode":

```txt
for p in 1..10:
    for j in 1..10:
    1) Run multsim with j jobs using p processes, up to the maximum limit.
    2) Verify that all stat.txt files exist and are not empty.
```

The output of these tests shows that there are always failures in cases where the number of jobs is greater than 4 times the number of processes and always pass otherwise.

To directly observe this error and "play around" with it, you can use the "multisim-issue-reproduce-stats-override.py" script. It can be executed directly with a gem5 binary. Just modify the number of processes and jobs directly in the configuration file, then  execute the script and inspect the output directories for each job and their corresponding stats.txt files.

**Host ISA**
ARM (Apple Silicon machine)

**Compiler used**
Clang 16.0.0.0

Harshil2107:
- Using `Multiprocessing.Pool` in multisim was causing errors when number of jobs were > 4x the number pf processes as a worker thread was being used to run a gem5 simulation after already running one.

- This change updates the multisim code to manually spawn the processes and handle them.

### DIFF
--- a/src/python/gem5/utils/multisim/multisim.py
+++ b/src/python/gem5/utils/multisim/multisim.py
@@ -53,8 +53,8 @@ import importlib
 import multiprocessing
 import signal
 import time
+from multiprocessing import Lock
 from pathlib import Path
-from threading import Lock
 from typing import (
     Optional,
     Set,
--- a/src/python/gem5/utils/multisim/multisim.py
+++ b/src/python/gem5/utils/multisim/multisim.py
@@ -209,7 +209,7 @@ def run(module_path: Path, processes: Optional[int] = None) -> None:
     active_processes = []
     remaining_ids = list(ids).copy()
     process_lock = Lock()
-    from gem5.utils.multiprocessing import Process
+    from ..multiprocessing import Process
 
     def handle_exit(signum, frame):
         """Signal handler to clean up processes on termination."""
@@ -246,6 +246,9 @@ def run(module_path: Path, processes: Optional[int] = None) -> None:
             # Wait for active processes to finish
             time.sleep(1)  # Avoid busy-waiting
             with process_lock:
+                # Using list comprehension to remove finished processes
+                # as using `remove` in a loop over the list will cause
+                # the list to be modified during iteration.
                 active_processes = [
                     process
                     for process in active_processes
--- a/src/python/gem5/utils/multisim/multisim.py
+++ b/src/python/gem5/utils/multisim/multisim.py
@@ -210,7 +210,13 @@ def run(module_path: Path, processes: Optional[int] = None) -> None:
     # module path (the config script specifying all simulations using MultiSim)
     # but a different ID. The ID is used to select the correct simulator to
     # run.
-    pool.starmap(_run, zip([module_path for _ in range(len(ids))], tuple(ids)))
+    try:
+        pool.starmap(
+            _run, zip([module_path for _ in range(len(ids))], tuple(ids))
+        )
+    finally:
+        pool.close()
+        pool.join()
 
 
 def set_num_processes(num_processes: int) -> None:
--- a/src/python/gem5/utils/multisim/multisim.py
+++ b/src/python/gem5/utils/multisim/multisim.py
@@ -207,7 +207,6 @@ def run(module_path: Path, processes: Optional[int] = None) -> None:
     while remaining_ids or active_processes:
         while remaining_ids and len(active_processes) < max_num_processes:
             id_to_run = remaining_ids.pop()
-            print(f"Running {id_to_run}")
             process = Process(
                 target=_run, args=(module_path, id_to_run), name=id_to_run
             )
@@ -216,7 +215,6 @@ def run(module_path: Path, processes: Optional[int] = None) -> None:
         for process in active_processes:
             if not process.is_alive():
                 active_processes.remove(process)
-    print("All simulations complete.")
 
 
 def set_num_processes(num_processes: int) -> None:
--- a/src/python/gem5/utils/multisim/multisim.py
+++ b/src/python/gem5/utils/multisim/multisim.py
@@ -51,13 +51,17 @@ This script is then passed to the child processes to load.
 
 import importlib
 import multiprocessing
+import signal
+import time
 from pathlib import Path
+from threading import Lock
 from typing import (
     Optional,
     Set,
 )
 
 from m5.core import override_re_outdir
+from m5.util import inform
 
 # A global variable which __main__.py flips to `True` when multisim is run as
 # an executable module.
@@ -172,8 +176,10 @@ def _run(module_path: Path, id: str) -> None:
     sim_list[0].override_outdir(subdir)
     # This doesn't do anything if none of the redirect options are passed
     override_re_outdir(subdir)
-
-    sim_list[0].run()
+    try:
+        sim_list[0].run()
+    except Exception as e:
+        inform(f"Error running simulator {id}: {e}")
 
 
 def run(module_path: Path, processes: Optional[int] = None) -> None:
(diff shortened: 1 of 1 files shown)


Harshil2107:
This commit fixes the error that occurs when the number of jobs are <= 4x number of processes. I am looking into the error that occurs when number of jobs > 4x number of processes.

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -172,8 +176,10 @@ def _run(module_path: Path, id: str) -> None:
     sim_list[0].override_outdir(subdir)
     # This doesn't do anything if none of the redirect options are passed
     override_re_outdir(subdir)
-
-    sim_list[0].run()
+    try:
+        sim_list[0].run()
```
powerjg, line 180: Did you test this? Try putting an error in the simulator (e.g., maybe in the exit event handler?) and see what happens.

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -200,17 +206,53 @@ def run(module_path: Path, processes: Optional[int] = None) -> None:
...
-
-    # Use the starmap function to create N child processes each with same
-    # module path (the config script specifying all simulations using MultiSim)
-    # but a different ID. The ID is used to select the correct simulator to
-    # run.
-    pool.starmap(_run, zip([module_path for _ in range(len(ids))], tuple(ids)))
+    active_processes = []
+    remaining_ids = list(ids).copy()
+    process_lock = Lock()
+    from gem5.utils.multiprocessing import Process
```
powerjg: The old code used relative imports... Why change it?

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -200,17 +206,53 @@ def run(module_path: Path, processes: Optional[int] = None) -> None:
...
+                    process = Process(
+                        target=_run,
+                        args=(module_path, id_to_run),
+                        name=id_to_run,
+                    )
+                    process.start()
+                    with process_lock:
+                        active_processes.append(process)
+                except Exception as e:
+                    inform(f"Error starting process for {id_to_run}: {e}")
```
powerjg, line 245: Do we want to move on, exit, or re-raise the exception?

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -200,17 +206,53 @@ def run(module_path: Path, processes: Optional[int] = None) -> None:
...
+            # Wait for active processes to finish
+            time.sleep(1)  # Avoid busy-waiting
+            with process_lock:
+                active_processes = [
```
powerjg, line 252: Why the "copy" here? I have a gut feeling why you chose to do this, but it would be good to document in a comment why you made that decision.

powerjg (commented):
From copilot (the second point is important to address)

### Review of Changes in `multisim.py`

The changes in the `multisim.py` file are aimed at improving the process handling for the gem5 MultiSim module. Here are the key modifications and their review:

1. **Imports and Global Variables**:
   - Added `signal`, `time`, and `Lock` from `threading`.
   - Added `inform` from `m5.util`.

2. **Error Handling in `_run` Function**:
   - Wrapped the simulator run call in a try-except block to log errors using `inform`.

3. **Process Management in `run` Function**:
   - Replaced the use of `multiprocessing.Pool` with manual process management using `Process` from `gem5.utils.multiprocessing`.
   - Introduced a lock (`process_lock`) for thread-safe modifications to the `active_processes` list.
   - Added signal handlers (`SIGINT` and `SIGTERM`) to ensure processes are cleaned up on termination.

4. **Main Process Loop**:
   - Managed starting new processes and handling active ones within a loop.
   - Added a sleep interval to avoid busy-waiting.
   - Ensured proper cleanup of processes in a finally block.

### Concerns and Suggestions

1. **Busy-Waiting**:
   - The use of `time.sleep(1)` helps mitigate busy-waiting but might need tuning based on performance needs.

2. **Thread-Safe Process List**:
   - Using `threading.Lock` is not suitable for multiprocessing. Instead, consider using `multiprocessing.Lock` or other synchronization mechanisms provided by the multiprocessing module.

3. **Signal Handling**:
   - Ensure that the signal handlers are sufficient for all environments where the code will run, as some platforms may handle signals differently.

4. **Exception Handling**:
   - The try-except blocks provide basic error logging. Consider adding more granular error handling if specific error types need different handling logic.

5. **Resource Management**:
   - Verify that all resources are properly managed and released, especially if the processes open files or network connections.

### Conclusion
The changes address the initial issues with `multiprocessing.Pool` by implementing a more controlled process management system. Ensure to test thoroughly to handle edge cases and platform-specific behaviors.

erin-le (commented):
I don't have a thorough understanding of the code itself, but I tested this PR using the multisim testing scripts that Bobby provided in the [issue](https://github.com/gem5/gem5/issues/1738) and all of the cases passed. These scripts test combinations of 1 process and 1 job up to 10 processes and 10 jobs, for a total of 100 cases.

src/python/gem5/utils/multisim/multisim.py
Harshil2107: I think we can throw the exeption for the simulator but then move on to other simulators.

src/python/gem5/utils/multisim/multisim.py
Harshil2107: Just to clarify, you mean why I used `.copy` to initialize the variable?

If yes that was so that the original list is not modified.

powerjg: No, I mean that you use a list comprehension instead of directly modifying `active_processes`

powerjg:
> I don't have a thorough understanding of the code itself, but I tested this PR using the multisim testing scripts that Bobby provided in the [issue](https://github.com/gem5/gem5/issues/1738) and all of the cases passed. These scripts test combinations of 1 process and 1 job up to 10 processes and 10 jobs, for a total of 100 cases.

@erin-le did you test thinks like 10 processes and 100 jobs? 10 processes 3 jobs?

src/python/gem5/utils/multisim/multisim.py
Harshil2107: I did that, as if we did it during a while loop we are editing the list we are iterating over and that might cause errors. I will add it as a comment.

Harshil2107:
I locally tested this change using Bobby's script with 10 processes and 50 jobs, It works properly but it looks like its slower in spawning processes as the jobs increase, which I assume is due to running the config script again and again for 50 simulator objects for every simulation.

src/python/gem5/utils/multisim/multisim.py
Harshil2107: Yes, I raised an exeption in exit event handler and it was caught and the error message was printed.

```
info: Error running simulator fs-sim: Exit event handler not implemented
```

The other simulations were not affected by this. I tested by running this fullsystem simulation with other SE mode simulation. The SE mode simulation ran as expected irrespective if the FS simulation was run first or last.

src/python/gem5/utils/multisim/multisim.py
```diff
@@ -51,13 +51,17 @@
 
 import importlib
 import multiprocessing
+import signal
+import time
 from pathlib import Path
+from threading import Lock
```
powerjg: This should be
```suggestion
from multiprocessing import Lock
```

The threading lock is not compatible with multiprocessing.

powerjg (commented):
One more comment, then this looks good to me. We should let Bobby give it a review before merging, though

BobbyRBruce:
> * Using `Multiprocessing.Pool` in multisim was causing errors when number of jobs were > 4x the number pf processes as a worker thread was being used to run a gem5 simulation after already running one.

Can you explain a bit more what exactly `Pool` was doing to cause this bug? I don't really understand this explanation. Why did 4x cause the issues?

powerjg:
> Can you explain a bit more what exactly Pool was doing to cause this bug? I don't really understand this explanation. Why did 4x cause the issues?

It simply didn't work. There were many different combinations of number of processes and the number of jobs such that it didn't work. We have to override the way multiprocessing works to get things to work with the gem5 binary, and the default `multiprocessing.Pool` is incompatible with this.

Let's merge this change since it makes multisim work better.

BobbyRBruce (approved):

PR #1851, opened 2024-12-11
configs: Generalize class types in CHI RNF/MN generators

tommarin:
Classes CHI_RNF and CHI_MN can be specialized to override base class/subclass attributes, like it happens in CustomMesh with router_list (see configs/example/noc_config/2x4.py). To avoid missing these attributes, it is needed to generalize the class types when instantiating the objects in the recently added generators.


giactra (approved):
@BobbyRBruce this should be added as a bug fix for 24.1.0.1 :)

PR #1855, opened 2024-12-12
Adds taken-only history

dhschall:
This PR adds taken-only history as proposed by https://ieeexplore.ieee.org/document/9246215
It is needed for the decoupled front-end


powerjg (approved):
With a quick scan this lgtm

PR #1861, opened 2024-12-14
cpu: Add EpisodeCount debug flag

Harshil2107:
Updating the print statement to only print when the debug flag is specified.

Adding this change as the current gpu0random test is failing due to being killed by the runner. The test runs as expected when tested locally.

The test migh be getting killed due to having a very large output.


powerjg:
@mattsinc let us know if this is a major problem for you. The test is failing in our runners, but it runs fine locally. So far, our only guess is that the 100+MiB to stdout is the problem.

mattsinc:
I am in the middle of trying to get final grades in, so it will be a few days before I can get to this.

mattsinc (approved):
I don't see anything wrong with this, although I agree with @powerjg that it seems unlikely this is the root problem.

PR #1864, opened 2024-12-16
mem-ruby: Fix missing RubySystem in PerfectCacheMemory's entries

hexengraf:
MOESI_CMP_directory protocol crashes with one of the several assertions in NetDest.cc. It happens because the entry type used to instantiate a PerfectCacheMemory object in MOESI_CMP_directory-L2cache.sm contains a NetDest object, so it requires a RubySystem object to be manually set for it.

Instead of just receiving the block size, change PerfectCacheMemory to receive a RubySystem object and use it to set the block size and call ENTRY::setRubySystem if the entries require it.

### DIFF
--- a/src/mem/ruby/structures/PerfectCacheMemory.hh
+++ b/src/mem/ruby/structures/PerfectCacheMemory.hh
@@ -41,11 +41,13 @@
 #ifndef __MEM_RUBY_STRUCTURES_PERFECTCACHEMEMORY_HH__
 #define __MEM_RUBY_STRUCTURES_PERFECTCACHEMEMORY_HH__
 
+#include <type_traits>
 #include <unordered_map>
 
 #include "base/compiler.hh"
 #include "mem/ruby/common/Address.hh"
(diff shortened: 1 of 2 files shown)


hexengraf:
This probably should be a hotfix?

I only tested using the deprecated config scripts, so let me know if I'm missing something and this crash is actually avoidable using different config scripts.

abmerop (approved):
Looks good to me. I might have missed this in the original commit or it's possible users have non-public changes where RubySystem ptr is needed.

@BobbyRBruce @powerjg would be able to respond about the hotfix recommendation.

aarmejach (approved):

aarmejach:
Our simulations with CHI were also failing due to empty NetDest on PerfectCacheMemory objects. This fix solves the issue on our end as well.

+1 on the request to make this a hotfix

powerjg (approved):
LGTM. The hotfix should be out before the holidays.

mattsinc (approved):

BobbyRBruce (approved):

PR #1399, opened 2024-07-30
mem-ruby: Prevent LL/SC livelock in MESI protocols (#1384)

hexengraf:
Fix #1384.

MESI_Two_Level and MESI_Three_Level protocols are susceptible to LL/SC livelocks when simulating boards with high core count.

This fix is based on MOESI_CMP_directory's implementation of locked states, but tailors the solution to only apply it when a Load-Linked is initiated.

There are two new states to act as locked states and stall any messages leading to eviction:
* LLSC_E: equivalent to E state, go to E after timeout.
* LLSC_M: equivalent to M state, go to M after timeout.

The main new event is Load_Linked, which is very similar (in behavior) to a Store, reusing several transient states. When a controller receives the exclusive data, it differentiates a Load_Linked from a Store by checking a new field added to the TBE: 'isLoadLinked'. It triggers a different event when it is a Load_Linked, which in turn causes the transition to one of the locked states.

The entire mechanism can be turned off by setting 'use_llsc_lock' to false, and the amount of time to keep locked is defined by 'llsc_lock_timeout_latency'.

Change-Id: I13f415b6b7890d51d01f23001047d2363467a814

PR #1865, opened 2024-12-16
mem-ruby: Add missing option in ProtocolInfo

hexengraf:
After the support for multiple ruby protocols was added, the macros PROTOCOL_MESI_Two_Level and PROTOCOL_MESI_Three_Level were removed. These macros are still being used to determine if Load_Linked requests are sent to the protocol, an information required by the fix that addresses LL/SC livelock.
Replace the macros with a new option: useSecondaryLoadLinked.


hexengraf:
It'd be great if this PR can be added as hotfix, since it relates to a fix (#1399) that was broken by later changes to allow multiple ruby protocols.

BobbyRBruce (approved):

PR #1869, opened 2024-12-17
mem: fix warning message for when memory size > 4GiB

### LINKED ISSUE #1860: warning: DRAM > 4GB and/or capacity is 0
**Describe the bug**
When running gem5 with memory > 4GiB, a warning said "DRAM > 4GiB and/or capacity 0)"
(It should not affect simulation results.)

**Affects version**
gem5 23.x, 24.x

**To Reproduce**
set DRAM > 4GiB

**Additional information**
fix: 

(24.1.0.0)
src/mem/mem_interface.hh (line: 134)
const uint32_t deviceSize;  => const uint64_t deviceSize;

(please help with commit)

erin-le:
This PR fixes the bug described in issue 1860, where a warning occurs when memory size is set to greater than 4GiB. It does this by changing the type of the variable deviceSize in src/mem/mem_interface.hh from uint32_t to uint64_t.

I was not able to reproduce the bug using configuration scripts adapted from those in `configs/example/gem5_library`, but I am still creating this PR so it is here if we need it.

### DIFF
--- a/src/mem/mem_interface.hh
+++ b/src/mem/mem_interface.hh
@@ -131,7 +131,7 @@ class MemInterface : public AbstractMemory
      * ranks and banks, the burst size, and the row buffer size.
      */
     const uint32_t burstSize;
-    const uint32_t deviceSize;
+    const uint64_t deviceSize;
     const uint32_t deviceRowBufferSize;
     const uint32_t devicesPerRank;
     const uint32_t rowBufferSize;


src/mem/mem_interface.hh
```diff
@@ -131,7 +131,7 @@ class MemInterface : public AbstractMemory
      * ranks and banks, the burst size, and the row buffer size.
      */
     const uint32_t burstSize;
-    const uint32_t deviceSize;
+    const uint64_t deviceSize;
```
powerjg, line 134: Just a note: deviceSize should be the size of a memory chip. The whole DIMM (or MemoryController in gem5) is `deviceSize * devicesPerRank * ranksPerChannel*. 

It's possible that this is above 4 GiB, so it should be a `uint64_t`. Some devices have just 1 devices per rank and 1 ranks per channel (e.g., HBM).

powerjg (approved):
Even if you can't reproduce, this seems like a bug, so let's fix it. It's not a very important bug though, because `devicesSize is only used for sanity checking.

PR #1870, opened 2024-12-18
resources: Update error handling for JSONClient

Harshil2107:
- This change updates error handling for the path of the JSON file in JSONClient class

### DIFF
--- a/src/python/gem5/resources/client_api/abstract_client.py
+++ b/src/python/gem5/resources/client_api/abstract_client.py
@@ -51,6 +51,10 @@ class AbstractClient(ABC):
         """
         try:
             result = urllib.parse.urlparse(url)
+            # Check if the URL is a file path
+            # we check path != "" because default path is "" and not None
+            if result.scheme == "file" and result.path != "":
+                return True
             return all([result.scheme, result.netloc, result.path])
         except:
             return False
--- a/src/python/gem5/resources/client_api/jsonclient.py
+++ b/src/python/gem5/resources/client_api/jsonclient.py
@@ -25,6 +25,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import json
+import urllib.parse
 from pathlib import Path
 from typing import (
     Any,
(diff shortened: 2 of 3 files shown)


src/python/gem5/resources/client_api/abstract_client.py
```diff
@@ -51,6 +51,8 @@ def _url_validator(self, url: str) -> bool:
         """
         try:
             result = urllib.parse.urlparse(url)
+            if result.scheme == "file":
```
powerjg: Don't you also want to make sure that `result.path` isn't empty?

src/python/gem5/resources/client_api/jsonclient.py
```diff
@@ -54,21 +54,59 @@ def __init__(self, path: str):
...
+
+            # Check if it's a valid URL
+            elif not self._url_validator(self.path):
+                raise ValueError(
+                    f"'{self.path}' is not a valid file path or URL."
                 )
-            self.resources = json.loads(response.read().decode("utf-8"))
+
+            # Handle 'file://' URIs
+            elif self.path.startswith("file://"):
```
powerjg: This is not a good solution. Use the urlparse to find out if it's a file url. Also, there's now way too much nesting in this code.

Please simplify the logic in this whole codeblock.

powerjg (changes requested):

src/python/gem5/resources/client_api/jsonclient.py
powerjg: Please use `urllib.parse.urlparse` here. Don't write your own code when there's code that exists in libraries to do what you're doing.

src/python/gem5/resources/client_api/jsonclient.py
```diff
@@ -54,21 +54,59 @@ def __init__(self, path: str):
...
+                    )
+                except json.JSONDecodeError as e:
+                    raise ValueError(
+                        f"Invalid JSON content at URL '{self.path}': {e}"
+                    )
+
+        except FileNotFoundError as e:
+            print(f"File Error: {e}")
+        except ConnectionError as e:
+            print(f"URL Error: {e}")
```
powerjg: Why is this not under the `else` "handle standard URLs"? Is there any way a connection error could happen in other cases?

Also, I generally don't like having nested try/except clauses. Could you give me an explanation for why this is necessary?

In researching this, it looks like what's bothering me about this whole code block is that it is a combination of the ["LBYL" style](https://docs.python.org/3/glossary.html#term-LBYL) and the ["EAFP" style](https://docs.python.org/3/glossary.html#term-EAFP). While the code you have here may work, I think the combination of these styles makes it difficult to understand and increases the chance of mistakes.

Harshil2107:
I pushed the commits with resolved changes yesterday, but it looks like they are still processing and not showing on the PR, I did remove the starts with and used url parse and reduce nesting.

powerjg:
Odd. Maybe try pushing again?

Harshil2107:
I will try force pushing and see if it works, my current guess is that the processing is stuck as the PR is stuck at `Checking for the ability to merge automatically...`

powerjg:
If it doesn't resolve, you may need to close the PR and open a new one.

Harshil2107:
Force pushing worked the changes show up now

src/python/gem5/resources/client_api/jsonclient.py
```diff
@@ -54,21 +55,47 @@ def __init__(self, path: str):
...
+                    with local_path.open("r", encoding="utf-8") as f:
+                        self.resources = json.load(f)
+                    return
+
+                # Handle HTTP/HTTPS URLs
+                req = request.Request(path)
+                with request.urlopen(req) as response:
+                    self.resources = json.loads(
+                        response.read().decode("utf-8")
+                    )
```
powerjg, line 88: This may be something you want to punt to another PR, but is this going to break for people who are using a proxy?

powerjg (commented):
This looks good. Thanks for the updates. One more small thing below.

powerjg:
I reran the failing jobs, but I think the pyunit test failure is real. You should look into it.

src/python/gem5/resources/client_api/jsonclient.py
Harshil2107: This will probably fail for people using proxy with an actual url as a JSON. This would still fail currently (as this part of the code is essentially the same). I will update it in another PR and add proxy context to the `urlopen` like it is done in `atlasclient.py` https://github.com/gem5/gem5/blob/c9625ce9cc5b5a90a38327de5ac0e1870974af5e/src/python/gem5/resources/client_api/atlasclient.py#L138

powerjg: Sounds good. Just make sure to minimize the duplicated code when you do that :)

powerjg (approved):

PR #1872, opened 2024-12-18
Stats Added to O3 CPU

KNS3092:
The following stats have been added:

1. Integer Free List Writes
2. Floating Point Free List Writes
3. Load Store Queue Writes

### DIFF
--- a/src/cpu/o3/lsq_unit.hh
+++ b/src/cpu/o3/lsq_unit.hh
@@ -539,7 +539,7 @@ class LSQUnit
          * is issued and its completion */
         statistics::Distribution loadToUse;
 
-        /** Total number of squashed stores. */
+        /** Total number of loads and stores written to the load store queue */
         statistics::Scalar addedLoadsAndStores;
     } stats;
 
--- a/src/cpu/o3/lsq_unit.hh
+++ b/src/cpu/o3/lsq_unit.hh
@@ -538,6 +538,9 @@ class LSQUnit
         /** Distribution of cycle latency between the first time a load
          * is issued and its completion */
         statistics::Distribution loadToUse;
+        
+        /** Total number of squashed stores. */
+        statistics::Scalar addedLoadsAndStores;
     } stats;
 
   public:
--- a/src/cpu/o3/lsq_unit.hh
+++ b/src/cpu/o3/lsq_unit.hh
@@ -538,7 +538,7 @@ class LSQUnit
         /** Distribution of cycle latency between the first time a load
          * is issued and its completion */
         statistics::Distribution loadToUse;
-        
+
         /** Total number of squashed stores. */
         statistics::Scalar addedLoadsAndStores;
     } stats;
--- a/src/cpu/o3/rename.cc
+++ b/src/cpu/o3/rename.cc
@@ -150,7 +150,7 @@ Rename::RenameStats::RenameStats(statistics::Group *parent)
                "count of registers freed and written back to integer free list"),
       ADD_STAT(fpWrites, statistics::units::Count::get(),
                "count of registers freed and written back to floating point free list")
-           
+
 {
     squashCycles.prereq(squashCycles);
     idleCycles.prereq(idleCycles);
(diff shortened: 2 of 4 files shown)


giactra:
Hi @KNS3092 . I know this is a WIP, but just so you know you shouldn't be opening a PR against stable (you should target develop instead)

src/cpu/o3/lsq_unit.hh
```diff
@@ -538,6 +538,9 @@ class LSQUnit
         /** Distribution of cycle latency between the first time a load
          * is issued and its completion */
         statistics::Distribution loadToUse;
+
+        /** Total number of squashed stores. */
```
mattsinc: At minimum I think this comment is wrong.  First, the above changes are seeming to add stats when new loads or stores are added to the LSQ, but this comments says we're counting number of squashes.  Second, comment says squashed stores but counter is for loads and stores.

mattsinc (dismissed):
Thanks @KNS3092 !  There is at least one change to make.  Plus, as @giactra highlighted, we'll need to change this from WIP to not WIP.

Lastly, @powerjg should chime in too with any opinions on overall stats across models.

src/cpu/o3/rename.cc
```diff
@@ -145,7 +145,12 @@ Rename::RenameStats::RenameStats(statistics::Group *parent)
       ADD_STAT(tempSerializing, statistics::units::Count::get(),
                "count of temporary serializing insts renamed"),
       ADD_STAT(skidInsts, statistics::units::Count::get(),
-               "count of insts added to the skid buffer")
+               "count of insts added to the skid buffer"),
+      ADD_STAT(intWrites, statistics::units::Count::get(),
+               "count of registers freed and written back to integer free list"),
+      ADD_STAT(fpWrites, statistics::units::Count::get(),
+               "count of registers freed and written back to floating point free list")
```
powerjg: Either the stat names are confusing or the stat description is wrong. When I see "intWrites" I would assume that this means the number of times the integer register file was written. 

Maybe I could help more if you could explain the use of this statistic. Why would people want to know the number of times we've written the free list?

Could a better name be "freeListWritesInt" and "freeListWritesFP"?

powerjg (commented):
Overall this seems good. One small thing below to try to help future people when they see these stats.

src/cpu/o3/rename.cc
mriankamp: For a bit of context, @KNS3092 and I were interested in the number of times integer registers and float registers were freed and written back to the free list because we were developing a power model and found that the integer and float free lists have different power consumption associated with them (Even though there's just one free list that mixes integers and floats in gem5).
I believe the names you suggested describe what the stats are more fully.
We chose the original names because we wanted to be consistent with the "intLookups" and "fpLookups" stats that we used to track the number of reads for integers and floats from the free list.
We can definitely make that name change if you think it'll be easier for people to understand, or if we want to be more consistent with the above stats, maybe "intReturned" and "fpReturned" are good for describing the number of registers 'returned' to the free list?
Here's what they look like as they are now in the stats file:
```
system\cpu\rename\intLookups
system\cpu\rename\fpLookups
...
system\cpu\rename\intWrites
system\cpu\rename\fpWrites
```

powerjg: Got it. In that case, all of these names are confusing to me 😆 

I like "intReturned" an "fpReturned" better than simply "writes". That said, it's not a strong preference, so I'll leave it up to you.

powerjg:
After you make Matt's change, remove "WIP" and make sure the commit has the right format, we can run the tests and merge the PR. Thanks again for your contribution!

src/cpu/o3/rename.cc
mattsinc: I agree with @powerjg about the naming updates.

src/cpu/o3/lsq_unit.hh
KNS3092: Comment has been corrected

src/cpu/o3/rename.cc
KNS3092: @powerjg we have changed the stats names to intReturned and fpReturned

powerjg (approved):
Whoever merges this needs to squash and fix the commit message.

PR #1874, opened 2024-12-20
Fix bug in SysBridge + XBar and Add option for external memory addresses

leoredivo:
Jason made these commits to enable disaggregated memory in gem5. The first commit fixes a panic thrown out when using a system bridge after a coherent crossbar. The second commit adds a new parameter to the System object. This allows physical addresses not managed by the System to go through the normal memory system. See commit descriptions for more details.

### DIFF
--- a/src/sim/System.py
+++ b/src/sim/System.py
@@ -86,6 +86,15 @@ class System(SimObject):
         [], "Ranges that constitute main memory"
     )
 
+    external_memory_ranges = VectorParam.AddrRange(
+        [],
+        "Ranges that are valid physical address but not part of physmem. "
+        "These are considered to be coherent addresses, not for I/O or "
+        "devices. This is used for external memory controllers which are "
+        "owned by a different instance of a `System` object (e.g., remote) "
+        "memory.",
+    )
+
     # The ranges backed by a shadowed ROM
     shadow_rom_ranges = VectorParam.AddrRange(
         [], "Ranges  backed by a shadowed ROM"
--- a/src/sim/system.cc
+++ b/src/sim/system.cc
@@ -166,6 +166,8 @@ int System::numSystemsRunning = 0;
 
 System::System(const Params &p)
     : SimObject(p), _systemPort("system_port"),
+      externalMemRanges(p.external_memory_ranges.begin(),
+                         p.external_memory_ranges.end()),
       multiThread(p.multi_thread),
       init_param(p.init_param),
       physProxy(_systemPort, p.cache_line_size),
@@ -287,7 +289,18 @@ System::memSize() const
 bool
 System::isMemAddr(Addr addr) const
 {
-    return physmem.isMemAddr(addr);
+    if (physmem.isMemAddr(addr)) {
+        return true;
+    } else {
+        // Check the external memory ranges as well
+        for (const auto& range : externalMemRanges) {
+            if (range.contains(addr)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
 }
 
 void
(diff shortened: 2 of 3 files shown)


powerjg:
I originally authored this code so someone else should review it.

powerjg:
@BobbyRBruce  can we get a review on this?

giactra:
Hi, thanks for the contribution; 

> Jason made these commits to enable disaggregated memory in gem5.

IMHO while I understand the two commits aim at the same purpose, I believe this PR should be split in two; one is fixing an existing routing bug, the other is adding an extra way for us to define memory.

I have reviewed the SysBridge changes and it looks ready to go; whereas I would like some more documentation on the external mem change as it is not clear to me what's its use is. I suspect the "not backed by physmem" won't mean a lot to a normal user trying to configure their setup neither

BobbyRBruce:
Apologies, I was playing around with this PR and accidentally pushed the fix straight to the develop branch via my elevated permissions: https://github.com/gem5/gem5/commit/30e2838c742ba5186b66bc2bcdd88a6b5aed3abe

I could revert it, but this change seemed close done anyway with only one request to break this into two PRs, suggesting there was no complaint there. I'll therefore keep it on develop unless anyone complains.

I believe the second commit has already been merged via another PR. In which case, am  I safe to close this?

giactra:
> Apologies, I was playing around with this PR and accidentally pushed the fix straight to the develop branch via my elevated permissions: [30e2838](https://github.com/gem5/gem5/commit/30e2838c742ba5186b66bc2bcdd88a6b5aed3abe)
> 
> I could revert it, but this change seemed close done anyway with only one request to break this into two PRs, suggesting there was no complaint there. I'll therefore keep it on develop unless anyone complains.
> 
> I believe the second commit has already been merged via another PR. In which case, am I safe to close this?

Hi Bobby,

> whereas I would like some more documentation on the external mem change as it is not clear to me what's its use is. I suspect the "not backed by physmem" won't mean a lot to a normal user trying to configure their setup neither

I actually don't know what the direct purpose of the patch is... I don't think we should revert it as of now, but I'd like a clarification on that front...

BobbyRBruce:
> entation on the external mem change as it is not clear to me what's its use is. I suspect the "not backed by phy

Thanks for clarifying. I'll keep it open. @leoredivo : Is it possible you address @giactra 's comments?

powerjg:
This was a while ago... let's see what I can remember. I agree that this should have been split into two patches. My apologies.

The problem I was trying to solve with the external memory is that we are trying to model CXL 3.0 devices where there is a *shared* memory between two systems. So, we need a way to specify that some addresses are *valid* addresses, but they *should not* appear in *this system's* physical memory backing store. (Instead, the addresses will appear in a *different* system's physical memory backing store.)

Note that while most gem5 simulations assume that there is exactly one `System` object, this is not a true strict requirement.

Said another way, we want to allow a `System` to have valid physical addresses to memory for which it does not own.

The only change that is needed is to modify the `isMemAddr` function in `System` to understand these new address ranges. This shouldn't cause any performance overhead in the default (normal) use case as they are only checked if the current check fails (i.e., on I/O accesses, which should be rare compared to regular memory accesses).

I'm happy to answer any other questions if you have them, @giactra

powerjg:
@giactra does my comment above address your concerns?

giactra:
Thanks @powerjg , apologies for the late reply. It makes a lot of sense now. IMHO it would be nice if we could document this use case in the System param description. Especially if its example usage resides in a fork, someone might accidentaly delete this down the line if its usage is not fully understood

powerjg (commented):
Let me know if the change doesn't address your concern, @giactra.

powerjg:
@BobbyRBruce @giactra can I have a quick review on this? Just some small documentation updates :)

giactra (approved):

PR #1877, opened 2024-12-22
arch-x86: Add YMM Registers to X86KvmCPU and Floating Point Register File

elise-song:
Co-authored with @xuann6.

When KvmCPU executes AVX instructions and is interrupted by a checkpoint, YMM registers are not correctly saved and restored. 

This PR adds YMM registers to the x86 Floating Point Register File and helper functions for accessing YMM to automatically serialize YMM registers during checkpointing. YMM registers follow the same format as existing XMM registers in the Floating Point Register File. 

This PR adds a struct called AVXState containing a 16x16 byte array to represent the 16 128-bit YMM registers in KvmCPU. Along with existing Floating Point Unit (FPU) update functions, thread context is updated to save the AVXState to the x86 YMM registers, and KVM state is updated to restore AVXState from x86 YMM registers. 

The KVM YMM registers are accessed according to the Intel Developer's Manual Chapter 13.5.3 AVX State. They are located in the XSAVE region below the XSAVE Header and XMM registers. 

Overall, this provides checkpointing support for YMM registers of AVX State in KvmCPU, described in issue #958.

### DIFF
--- a/src/arch/x86/kvm/x86_cpu.cc
+++ b/src/arch/x86/kvm/x86_cpu.cc
@@ -116,7 +116,7 @@ struct GEM5_PACKED FXSave
     uint64_t reserved[12];
 };
 
-// for the upper 128 bits of 16 ymm registers 
+// for the upper 128 bits of 16 ymm registers
 struct GEM5_PACKED AVXState {
     uint8_t ymmh[16][16];
 };
@@ -880,10 +880,10 @@ X86KvmCPU::updateKvmStateSRegs()
 void
 X86KvmCPU::updateKvmStateAVXCommon(ThreadContext *tc, const kvm_xsave &kxsave)
 {
-    // Like updateKvmStateFPUCommon, updateKvmStateAVXCommon is called by 
-    // updateKvmStateFPU function 
+    // Like updateKvmStateFPUCommon, updateKvmStateAVXCommon is called by
+    // updateKvmStateFPU function
     // Setting Kvm's ymm reg values to gem5's ymm reg values
-    const AVXState *avx = (const AVXState*)((const char*)kxsave.region 
+    const AVXState *avx = (const AVXState*)((const char*)kxsave.region
                             + sizeof(FXSave) + sizeof(XSaveHeader));
 
     for (int i = 0; i < 16; i++) {
@@ -1148,16 +1148,16 @@ X86KvmCPU::updateThreadContextSRegs(const struct kvm_sregs &sregs)
 #undef APPLY_DTABLE
 }
 
-void 
-X86KvmCPU::updateThreadContextAVXCommon(ThreadContext *tc, 
-                                        const struct kvm_xsave &kxsave) 
+void
+X86KvmCPU::updateThreadContextAVXCommon(ThreadContext *tc,
+                                        const struct kvm_xsave &kxsave)
 {
-    // Setting gem5's ymm reg values to Kvm's ymm reg values 
-    XSaveHeader& xsave_hdr = * (XSaveHeader *) ((char *) &kxsave 
+    // Setting gem5's ymm reg values to Kvm's ymm reg values
+    XSaveHeader& xsave_hdr = * (XSaveHeader *) ((char *) &kxsave
                                 + sizeof(FXSave));
-    
+
     if (xsave_hdr.xstate_bv.avx) {
-        const AVXState *avx = (const AVXState*)((const char*)kxsave.region 
+        const AVXState *avx = (const AVXState*)((const char*)kxsave.region
                                 + sizeof(FXSave) + sizeof(XSaveHeader));
 
         for (int i = 0; i < 16; i++) {
--- a/src/arch/x86/kvm/x86_cpu.cc
+++ b/src/arch/x86/kvm/x86_cpu.cc
@@ -116,6 +116,13 @@ struct GEM5_PACKED FXSave
     uint64_t reserved[12];
 };
 
+// for the upper 128 bits of 16 ymm registers 
+struct GEM5_PACKED AVXState {
+    uint8_t ymmh[16][16];
+};
+
+// update according checker for AVX ymm reg
+static_assert(sizeof(AVXState) == 256, "Unexpected size of AVXState");
 static_assert(sizeof(FXSave) == 512, "Unexpected size of FXSave");
 
 BitUnion64(XStateBV)
@@ -869,6 +876,22 @@ X86KvmCPU::updateKvmStateSRegs()
     setSpecialRegisters(sregs);
 }
 
+
+void
+X86KvmCPU::updateKvmStateAVXCommon(ThreadContext *tc, const kvm_xsave &kxsave)
+{
+    // Like updateKvmStateFPUCommon, updateKvmStateAVXCommon is called by 
+    // updateKvmStateFPU function 
+    // Setting Kvm's ymm reg values to gem5's ymm reg values
+    const AVXState *avx = (const AVXState*)((const char*)kxsave.region 
+                            + sizeof(FXSave) + sizeof(XSaveHeader));
+
+    for (int i = 0; i < 16; i++) {
+        *(uint64_t *)&avx->ymmh[i][0] = tc->getReg(float_reg::ymmHigh(i));
+        *(uint64_t *)&avx->ymmh[i][8] = tc->getReg(float_reg::ymmLow(i));
+    }
+}
+
 template <typename T>
 static void
 updateKvmStateFPUCommon(ThreadContext *tc, T &fpu)
@@ -952,6 +975,10 @@ X86KvmCPU::updateKvmStateFPUXSave()
       * (XSaveHeader *) ((char *) &kxsave + sizeof(FXSave));
     xsave_hdr.xstate_bv.fpu = 1;
     xsave_hdr.xstate_bv.sse = 1;
+
+    // enable avx bitmap and run the AVX state update function
+    xsave_hdr.xstate_bv.avx = 1;
+    updateKvmStateAVXCommon(tc, kxsave);
 
     if (tc->readMiscRegNoEffect(misc_reg::Fiseg))
         warn_once("misc_reg::Fiseg is non-zero.\n");
@@ -1121,6 +1148,25 @@ X86KvmCPU::updateThreadContextSRegs(const struct kvm_sregs &sregs)
 #undef APPLY_DTABLE
 }
 
+void 
+X86KvmCPU::updateThreadContextAVXCommon(ThreadContext *tc, 
+                                        const struct kvm_xsave &kxsave) 
+{
+    // Setting gem5's ymm reg values to Kvm's ymm reg values 
+    XSaveHeader& xsave_hdr = * (XSaveHeader *) ((char *) &kxsave 
+                                + sizeof(FXSave));
+    
+    if (xsave_hdr.xstate_bv.avx) {
+        const AVXState *avx = (const AVXState*)((const char*)kxsave.region 
+                                + sizeof(FXSave) + sizeof(XSaveHeader));
+
+        for (int i = 0; i < 16; i++) {
+            tc->setReg(float_reg::ymmLow(i), *(uint64_t *)&avx->ymmh[i][0]);
+            tc->setReg(float_reg::ymmHigh(i), *(uint64_t *)&avx->ymmh[i][8]);
+        }
+    }
+}
+
 template<typename T>
 static void
 updateThreadContextFPUCommon(ThreadContext *tc, const T &fpu)
@@ -1172,6 +1218,8 @@ X86KvmCPU::updateThreadContextXSave(const struct kvm_xsave &kxsave)
     const FXSave &xsave(*(const FXSave *)kxsave.region);
 
     updateThreadContextFPUCommon(tc, xsave);
+
+    updateThreadContextAVXCommon(tc, kxsave);
 
     tc->setMiscRegNoEffect(misc_reg::Fiseg, 0);
     tc->setMiscRegNoEffect(misc_reg::Fioff, xsave.ctrl64.fpu_ip);
--- a/src/arch/x86/kvm/x86_cpu.cc
+++ b/src/arch/x86/kvm/x86_cpu.cc
@@ -112,7 +112,7 @@ struct GEM5_PACKED FXSave
 
     uint8_t fpr[8][16];
     uint8_t xmm[16][16];
-
+    
     uint64_t reserved[12];
 };
 
@@ -882,9 +882,10 @@ X86KvmCPU::updateKvmStateAVXCommon(ThreadContext *tc, const kvm_xsave &kxsave)
 {
     // Like updateKvmStateFPUCommon, updateKvmStateAVXCommon is called by
     // updateKvmStateFPU function
-    // Setting Kvm's ymm reg values to gem5's ymm reg values
-    const AVXState *avx = (const AVXState*)((const char*)kxsave.region
-                            + sizeof(FXSave) + sizeof(XSaveHeader));
+    // Setting KVM's YMM reg values to gem5's YMM reg values
+    const AVXState *avx = (const AVXState*)((const char*)
+                            kxsave.region[sizeof(FXSave) 
+                            + sizeof(XSaveHeader)]);
 
     for (int i = 0; i < 16; i++) {
         *(uint64_t *)&avx->ymmh[i][0] = tc->getReg(float_reg::ymmHigh(i));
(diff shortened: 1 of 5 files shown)


mattsinc:
@v-ramadas : just an FYI so you are on this thread

src/arch/x86/kvm/x86_cpu.cc
```diff
@@ -869,6 +876,22 @@ X86KvmCPU::updateKvmStateSRegs()
     setSpecialRegisters(sregs);
 }
 
+
+void
+X86KvmCPU::updateKvmStateAVXCommon(ThreadContext *tc, const kvm_xsave &kxsave)
+{
+    // Like updateKvmStateFPUCommon, updateKvmStateAVXCommon is called by
+    // updateKvmStateFPU function
+    // Setting Kvm's ymm reg values to gem5's ymm reg values
```
mattsinc: Minor: Kvm --> KVM, ymm --> YMM

src/arch/x86/kvm/x86_cpu.cc
```diff
@@ -953,6 +976,10 @@ X86KvmCPU::updateKvmStateFPUXSave()
     xsave_hdr.xstate_bv.fpu = 1;
     xsave_hdr.xstate_bv.sse = 1;
 
+    // enable avx bitmap and run the AVX state update function
```
mattsinc: can you please update your comment here to explain why this is needed? / what it is needed for?

src/arch/x86/kvm/x86_cpu.cc
```diff
@@ -1121,6 +1148,25 @@ X86KvmCPU::updateThreadContextSRegs(const struct kvm_sregs &sregs)
 #undef APPLY_DTABLE
 }
 
+void
+X86KvmCPU::updateThreadContextAVXCommon(ThreadContext *tc,
+                                        const struct kvm_xsave &kxsave)
+{
+    // Setting gem5's ymm reg values to Kvm's ymm reg values
```
mattsinc: same capitalization comment here

mattsinc (changes requested):
In addition to the minor changes here, you'll need to redo your commit message to conform to the gem5 commit guidelines (https://www.gem5.org/contributing) -- you'll need to start with something like arch-x86: ...

src/arch/x86/kvm/x86_cpu.cc
```diff
@@ -116,6 +116,13 @@ struct GEM5_PACKED FXSave
     uint64_t reserved[12];
 };
 
+// for the upper 128 bits of 16 ymm registers
+struct GEM5_PACKED AVXState {
+    uint8_t ymmh[16][16];
+};
```
powerjg, line 122: Why not add these to the FXSave struct?

src/arch/x86/kvm/x86_cpu.cc
```diff
@@ -869,6 +876,22 @@ X86KvmCPU::updateKvmStateSRegs()
...
 
+
+void
+X86KvmCPU::updateKvmStateAVXCommon(ThreadContext *tc, const kvm_xsave &kxsave)
+{
+    // Like updateKvmStateFPUCommon, updateKvmStateAVXCommon is called by
+    // updateKvmStateFPU function
+    // Setting Kvm's ymm reg values to gem5's ymm reg values
+    const AVXState *avx = (const AVXState*)((const char*)kxsave.region
+                            + sizeof(FXSave) + sizeof(XSaveHeader));
```
powerjg: This code makes me very uncomfortable. At a minimum, you should use `uint8_t` instead of `char`. However, I don't like doing this kind of pointer arithmetic at all. It's error prone and hard to debug. If it's at all possible to use a struct to get the offset to the AVXState it would be much better.

src/arch/x86/regs/float.hh
```diff
@@ -179,6 +213,24 @@ xmmHigh(int index)
     return floatRegClass[XmmBase + 2 * index + 1];
 }
 
+static inline RegId
+ymm(int index)
+{
+    return floatRegClass[YmmBase + index];
+}
```
powerjg: This function is unused. Unless there is a good reason to keep it, it should be removed.

src/arch/x86/regs/float.cc
```diff
@@ -63,6 +63,12 @@ FlatFloatRegClassOps::regName(const RegId &id) const
         return ss.str();
     }
     reg_idx -= NumXMMRegs * 2;
+    if (reg_idx < NumXMMRegs * 2) {
+        ccprintf(ss, "%%ymm%d_%s", reg_idx / 2,
+                (reg_idx % 2) ? "high": "low");
+        return ss.str();
+    }
+    reg_idx -= NumXMMRegs * 2;
```
powerjg: You should add a new member which is `NumYMMRegs` instead of reusing `NumXMMRegs`.

powerjg (changes requested):
Thanks for the contribution! This will be good to get merged, but there are a few things to improve first.

Another question: Can we go ahead and bite the bullet and do the zmm registers, too? I assume the code will look similar, so maybe it will be easy for you to implement and test?

src/arch/x86/kvm/x86_cpu.cc
abmerop: I think `kxsave.region[sizeof(FXSave) + sizeof(XSaveHeader)]` would resolve this

abmerop:
> Another question: Can we go ahead and bite the bullet and do the zmm registers, too? I assume the code will look similar, so maybe it will be easy for you to implement and test?

Pushing back a bit because I don't think the code would look similar, except maybe for maybe the AVX512_HI feature (something like that).  Some other additions with AVX-512 are the mask registers and the increase from 16 vgprs to 32 which have additional xsave information, so the change would involve adding all of those as well most likely. I don't know if there is a real CPU doesn't enable all 3 at once. I think also the way Linux stores xsave doesn't actually have enough space for 32x 512b registers and does something special, but I'd have to check again. I'd be happy seeing this feature standalone and the AVX-512 parts can come when (if) gem5 adds AVX-512 support.

Otherwise I am interested in seeing a before/after, i.e., what tests did you run to confirm it works?

powerjg:
> Pushing back a bit because I don't think the code would look similar, except maybe for maybe the AVX512_HI feature (something like that). Some other additions with AVX-512 are the mask registers and the increase from 16 vgprs to 32 which have additional xsave information, so the change would involve adding all of those as well most likely.

Happy to be wrong :). Thanks for the information, Matt.

mattsinc:
> > Pushing back a bit because I don't think the code would look similar, except maybe for maybe the AVX512_HI feature (something like that). Some other additions with AVX-512 are the mask registers and the increase from 16 vgprs to 32 which have additional xsave information, so the change would involve adding all of those as well most likely.
> 
> Happy to be wrong :). Thanks for the information, Matt.

We should definitely make sure they have tests in place, regardless -- we'll need those no matter what.  If @abmerop feels adding YMM is too hard, then I'm ok with that too.

mattsinc:
@eys29 let us/me know if you need help making the requested changes.

erin-le:
@eys29 Hello, thank you for your contribution! Could you give an update on the status of this PR?

elise-song:
Hi, thank you everyone for the comments and feedback. @xuann6 and I will be working on changes in the next weeks. We are both traveling at the moment, sorry for the delay!

erin-le:
Hello @eys29, could you give another update on the status of this PR? Thank you!

erin-le:
@eys29, @xuann6: Could someone provide an update on the current status of this PR? Thank you!

erin-le:
@mattsinc, @eys29, @xuann6: Hello, could someone provide an update on the status of this PR? Thank you!

mattsinc:
> @mattsinc, @eys29, @xuann6: Hello, could someone provide an update on the status of this PR? Thank you!

I am assuming since the students are no longer in my class they don't intend to update further ... I will try to get to this myself after the semester ends.

erin-le:
@mattsinc: Hello, are you still planning to get back to this PR? If not, do you mind if I close this PR for inactivity?

abmerop:
> @mattsinc: Hello, are you still planning to get back to this PR? If not, do you mind if I close this PR for inactivity?

Hi Erin, I'll wait for Matt to respond but I am also willing to sheppard this forward as it is ~95% of the way there and the change requests look minor.

mattsinc:
@eys29 let us/me know if you need help making the requested changes.

> > @mattsinc: Hello, are you still planning to get back to this PR? If not, do you mind if I close this PR for inactivity?
> 
> Hi Erin, I'll wait for Matt to respond but I am also willing to sheppard this forward as it is ~95% of the way there and the change requests look minor.

sorry for the delay.  I definitely will not get to this in the next two weeks. So @abmerop if you want to get it across the line please go ahead.  Else, I'll try to get to it afterwards.  It would definitely be a shame to lose it.

elise-song:
Sorry for the delay on this. The most recent push addresses most of the reviewers' feedback. Thanks for everyone's help. 

> Otherwise I am interested in seeing a before/after, i.e., what tests did you run to confirm it works?

We implemented a simple unit test that invoked an AVX instruction (e.g., VMOVUPS/_mm256_set_ps), followed by a checkpoint.  The checkpoint file was verified to contain the expected YMM register values. After restoring from the checkpoint, the YMM register was also verified to match the checkpoint file's values (using _mm256_storeu_ps). We also ran short-running programs with a checkpoint to further verify.

mattsinc (approved):
as far as I can tell, @eys29 has made all of @abmerop and my requested changes -- except for having a test we can integrate to test it.

@eys29 in your reply you mentioned a test you used -- perhaps we can integrate that into gem5-resources?

mattsinc:
> as far as I can tell, @eys29 has made all of @abmerop and my requested changes -- except for having a test we can integrate to test it.
> 
> @eys29 in your reply you mentioned a test you used -- perhaps we can integrate that into gem5-resources?

@powerjg when you get a chance, let us know if this addresses your requested changes

src/arch/x86/kvm/x86_cpu.cc
powerjg: Could you address this question? (Apologies if I missed it in another place.)

src/arch/x86/kvm/x86_cpu.cc
```diff
@@ -869,6 +876,23 @@ X86KvmCPU::updateKvmStateSRegs()
...
 }
 
+
+void
+X86KvmCPU::updateKvmStateAVXCommon(ThreadContext *tc, const kvm_xsave &kxsave)
+{
+    // Like updateKvmStateFPUCommon, updateKvmStateAVXCommon is called by
+    // updateKvmStateFPU function
+    // Setting KVM's YMM reg values to gem5's YMM reg values
+    const AVXState *avx = (const AVXState*)((const char*)
```
powerjg, line 886: Is this dependent on the host at all? I.e., is it possible that there's a host without avx that tries to run this code an causes a fault?

powerjg (commented):
A couple of small questions below. Also, it looks like clang in the CI tests caught a bug.

mattsinc:
thanks @powerjg , @eys29 let us know!

src/arch/x86/kvm/x86_cpu.cc
elise-song: @xuann6 would you mind addressing these?

elise-song:
Here is the test we used:
```
#include <immintrin.h>
#include <stdio.h>

void m5_checkpoint()
{
    asm volatile(".byte 0x0F, 0x04;"
                 ".word 0x40;"
                 ".byte 0x00;");
}

int main() {
    // Create a YMM register with different values in upper/lower halves
    __m256 a = _mm256_set_ps(8.0, 7.0, 6.0, 5.0,  // upper 128 bits
                            4.0, 3.0, 2.0, 1.0);   // lower 128 bits
    
    // Store it in memory before checkpoint
    float before[8];
    _mm256_storeu_ps(before, a);
    
    printf("Before checkpoint:\n");
    for(int i = 0; i < 8; i++) {
        printf("value[%d]: %f\n", i, before[i]);
    }

    // Create checkpoint using m5 operation
    m5_checkpoint();
    
    // Try to use the same YMM register without reinitializing
    float after[8];
    _mm256_storeu_ps(after, a);
    
    printf("\nAfter checkpoint:\n");
    for(int i = 0; i < 8; i++) {
        printf("value[%d]: %f\n", i, after[i]);
    }
    
    return 0;
}
```

Could we get some guidance on how to write a proper test for gem5-resources? Are there checkpointing tests that we can reference?

erin-le:
@eys29, @xuann6, @mattsinc: Hello, I mentioned this question

> Could we get some guidance on how to write a proper test for gem5-resources? Are there checkpointing tests that we can reference?

to Jason and Bobby, and they said that the test doesn't need to be added.

However, could someone address Jason's (powerjg's) questions, and also fix the failing CI tests? Thank you!

mattsinc:
> @eys29, @xuann6, @mattsinc: Hello, I mentioned this question
> 
> > Could we get some guidance on how to write a proper test for gem5-resources? Are there checkpointing tests that we can reference?
> 
> to Jason and Bobby, and they said that the test doesn't need to be added.
> 
> However, could someone address Jason's (powerjg's) questions, and also fix the failing CI tests? Thank you!

Thanks @eys29 , @erin-le .

@eys29 : this is definitely useful, and @erin-le could probably guide you through the process of adding this test (although like @erin-le said, not necessarily mandatory for this commit).  @xuann6 : I think the main issue is waiting on your feedback now.  Any chance you can chime in?

erin-le:
@xuann6: Hello, could you address powerjg's comments? Thank you!

src/arch/x86/kvm/x86_cpu.cc
xuann6: The reasoning behind separating FXSave region and our AVXState region came from the Intel 64 and IA32 Architectures Manual shown above. The XSAVE header (13.4) is 64 bytes large and contains the XMM registers. The extended region AVXState (13.5.3) contains YMM registers. If someone wanted to extend to ZMM, they would go in the AVX-512 state (13.5.5). 

We wanted to mirror the manual as close as possible but if it makes sense to contain XMM, YMM (and maybe ZMM) in the same struct, we can make those changes. Let us know what you think!

image
image
image

src/arch/x86/kvm/x86_cpu.cc
xuann6: We made the above change, and get a casting error in the CI tests. @abmerop what do you suggest? We are bad at cpp lol.

image

src/arch/x86/kvm/x86_cpu.cc
powerjg: Got it! This makes sense. I didn't realize xmm was lumped in with x87 but ymm were separated out. I appreciate the explanation.

xuann6: if there's a host without avx, it would not enter the `updateKvmStateAVXCommon` function. Separately, the check for AVX support would happen elsewhere (we don't know where). Currently, we assume the machine has AVX support by setting the .avx bit to 1:
```
xsave_hdr.xstate_bv.fpu = 1;
xsave_hdr.xstate_bv.sse = 1;
xsave_hdr.xstate_bv.avx = 1; // here
```

src/arch/x86/kvm/x86_cpu.cc
abmerop: I guess since it is using uint64_t below you could do:

```
    const uint64_t *avx = reinterpret_cast<const uint64_t *>(&kxsave.region[sizeof(FXSave) + sizeof(XSaveHeader)]);

    for (int i = 0; i < 16; i++) {
        avx->ymmh[i][0] = tc->getReg(float_reg::ymmLow(i));
        avx->ymmh[i][8] = tc->getReg(float_reg::ymmHigh(i));
    }
```

Note: I did not test this though

erin-le:
@xuann6, @eys29, @mattsinc: Thanks for responding to Jason's (powerjg's) questions! I think the next step to move this forward should be fixing the CI test failures. Do you know when you'll have time to work on it?

mattsinc:
> @xuann6, @eys29, @mattsinc: Thanks for responding to Jason's (powerjg's) questions! I think the next step to move this forward should be fixing the CI test failures. Do you know when you'll have time to work on it?

@xuann6 , @eys29 : do you think you can look at what the CI failure is and fix it?  In theory it should be something minor, but it's been long enough now that we probably need to rerun the test to see what the failure is?

mattsinc:
> > @xuann6, @eys29, @mattsinc: Thanks for responding to Jason's (powerjg's) questions! I think the next step to move this forward should be fixing the CI test failures. Do you know when you'll have time to work on it?
> 
> @xuann6 , @eys29 : do you think you can look at what the CI failure is and fix it? In theory it should be something minor, but it's been long enough now that we probably need to rerun the test to see what the failure is?

@erin-le is there a way to relaunch the tests so I can see what failed?

erin-le:
@mattsinc: I've rebased the PR to be on top of current `develop` to trigger a re-run of the tests. I didn't know this before, but it seems like the option to re-run a workflow goes away 30 days after the original run: https://docs.github.com/en/actions/how-tos/manage-workflow-runs/re-run-workflows-and-jobs#about-re-running-workflows-and-jobs.

erin-le:
@mattsinc: Would you be okay with it if I closed this PR, as it has been inactive for some time? It can be reopened later if someone wants to pick it up again.

mattsinc:
> @mattsinc: Would you be okay with it if I closed this PR, as it has been inactive for some time? It can be reopened later if someone wants to pick it up again.

@erin-le I'm not sure what the value in closing patches is, but yes I don't think @elise-song will be updating this any time soon, so if you really need to close it, it's fine to do so.

PR #1891, opened 2025-01-03
Devcontainer fixes

powerjg:


util/dockerfiles/README.md
```diff
@@ -33,19 +33,19 @@ It is worth consulting these files and noting the "targets" and "groups", these
...
 ```sh
-docker buildx bake gcc-compiler
+docker buildx bake --platform linux/amd64 gcc-compiler
 ```
 
 If no target is specified all the images will be built.
 
 ```sh
-docker buildx bake
+docker buildx bake --platform linux/amd64
```
BobbyRBruce, line 48: `---platform linux/amd64` will only build the images to the linux/amd64 platform. These are multiplatform images, so you just build them with `docker buildx bake`, and they will run on `linux/amd64` and `linux/aarch64`. Your local Docker will handle which of the multiple images to use depending on your host machine.

If you do this, people who run on ARM machines (like me) won't be able to use these images (outside of emulation of X86).

powerjg: No. That does not work on any platform I have access to. The following is the output.

```
[+] Building 0.3s (1/1) FINISHED                                                                         docker:default  
=> [internal] load local bake definitions                                                                         0.0s  
=> => reading docker-bake.hcl 8.78kB / 8.78kB                                                                     0.0s 
ERROR: Multi-platform build is not supported for the docker driver.                                                     
Switch to a different driver, or turn on the containerd image store, and try again.                                     
Learn more at https://docs.docker.com/go/build-multi-platform/ 
```

I am happy to update the readme, but we *must* have directions with correctly work on the platforms that people use. Specifically, I have needed to build these docker containers many times and wasted 15-30 minutes each time trying to figure out what the right "bake" command is to get them to build. I don't want to waste this time again, so we need to update the readme.

Edit: I just tested on *both* x86 and Arm Linux and I get the same error. So, we should change the readme to a command that doesn't give an error.

BobbyRBruce (approved):

PR #1896, opened 2025-01-05
include algorithm in random.hh

MGThePro:
This was seemingly left out in #1534 . The way it is right now it builds on some but not all distributions. Seems like at least Arch Linux is one where these things need to be explicitly included.

Without this it fails on line 59 of the file `src/base/random.cc` with the error message
`error: ‘remove_if’ is not a member of ‘std’; did you mean ‘remove_cv’?`
After adding this it at least built the ARM debug build successfully, haven't tested anything else

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ cscope.out
 .*.swo
 m5out
 /src/doxygen/html
+/docs/_build
 /ext/dramsim2/DRAMSim2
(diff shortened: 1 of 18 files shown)


MGThePro:
Oops, made it merge into stable by accident and changing it to develop made it worse

will close this and reopen properly

PR #1897, opened 2025-01-05
include algorithm in random.hh

MGThePro:
This was seemingly left out in #1534 . The way it is right now it builds on some but not all distributions. Seems like at least Arch Linux is one where these things need to be explicitly included.

Without this it fails on line 59 of the file `src/base/random.cc` with the error message
`error: ‘remove_if’ is not a member of ‘std’; did you mean ‘remove_cv’?`
After adding this it at least built the ARM debug build successfully, haven't tested anything else

### DIFF
--- a/src/base/random.hh
+++ b/src/base/random.hh
@@ -45,6 +45,7 @@
 #ifndef __BASE_RANDOM_HH__
 #define __BASE_RANDOM_HH__
 
+#include <algorithm>
 #include <random>
 #include <string>
 #include <type_traits>


powerjg (approved):
Thanks for the contribution!

src/base/random.hh
```diff
@@ -45,6 +45,7 @@
 #ifndef __BASE_RANDOM_HH__
 #define __BASE_RANDOM_HH__
 
+#include <algorithm>
```
odanrc, line 48: On the other hand, `<memory>` is missing here, since shared_ptr is being used

odanrc (changes requested):
Are you sure your local develop branch is up to date? In the current develop branch, random.hh does not use any algorithm function, and random.cc already includes `<algorithm>`, so that error would not be possible.

MGThePro:
Oh you're right, that include was already in develop, but not on stable. I must've tested it on stable then, my bad

PR #1898, opened 2025-01-05
Add an API to set the exit handler for the simulator

pranith:


src/python/gem5/simulate/simulator.py
```diff
@@ -463,6 +463,23 @@ def get_id(self) -> Optional[str]:
...
+            ],
+        ]
+    ) -> None:
+        """Set the on_exit_event handler for the simulator.
+
+        Set the exit event handler that specifies what to execute on each exit
+        event."""
+
+        self._on_exit_event = on_exit_event
+
```
BobbyRBruce, line 482: You'll need to do some translation of the dictionary like we do in the constructor (here: https://github.com/gem5/gem5/blob/186a913a48f13bdd484fcdef17ac3e28d2b8b4c9/src/python/gem5/simulate/simulator.py#L363).

I'd recommend having the constructor call this new function and moving the translation code here.

BobbyRBruce (changes requested):
I take it the purpose of this is to change the exit event handlers after the simulation initialization? I think that'll work but I want to ensure my assumption here is correct.

erin-le:
@pranith Thank you for your contribution! Could you respond to the comments on this PR when you have time?

erin-le:
@pranith Hello, this PR is being closed for inactivity. If you would like to continue working on this PR, please feel free to reopen it. Thank you!

pranith:
I will update the commit as suggested.

PR #1902, opened 2025-01-06
mem-cache: Fix use-after-free in MSHR handling

powerjg:
Found the error while running valgrind. In the code, when popping the MSHR target it will delete the data pointed to by `initial_tgt`. This change just swaps the order.

Valgrind output (just the first line):
==1267085== Invalid read of size 8
==1267085==    at 0x256BF8D: gem5::Cache::serviceMSHRTargets(gem5::MSHR*, gem5::Packet*, gem5::CacheBlk*) (cache.cc:725)


giactra (approved):
Nice!

PR #1904, opened 2025-01-06
dev: Fix memory leaks

### LINKED ISSUE #1903: Potential memory leaks
In trying to track down the memory usage increase on x86 boot, I ran valgrind. I didn't find the culprit to our memory usage increase, but I did find some potential memory leaks. Noting them here.

I've also attached the full valgrind output.

### This one shows up a couple of times. Seems like a real leak.

Shows up both on the read and the write path

```
==1267085== 72 bytes in 1 blocks are possibly lost in loss record 16,205 of 29,787
==1267085==    at 0x72BFFA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==1267085==    by 0x34760D0: gem5::TimingSimpleCPU::handleReadPacket(gem5::Packet*) (timing.cc:282)
```

```
==1267085== 42,192 bytes in 586 blocks are definitely lost in loss record 29,253 of 29,787
==1267085==    at 0x72BFFA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==1267085==    by 0x3476975: gem5::TimingSimpleCPU::handleWritePacket() (timing.cc:511)
==1267085==    by 0x347ACB5: gem5::TimingSimpleCPU::sendData(std::shared_ptr<gem5::Request> const&, unsigned char*, unsigned long*, bool) (timing.cc:336)
```

https://github.com/gem5/gem5/blob/develop/src/cpu/simple/timing.cc#L282
https://github.com/gem5/gem5/blob/develop/src/cpu/simple/timing.cc#L511

Note: There are multiple paths to get to these `new` calls.

### These are likely missing a delete at the same place

```
==1267085== 472 (176 direct, 296 indirect) bytes in 1 blocks are definitely lost in loss record 25,224 of 29,787
==1267085==    at 0x72BFFA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==1267085==    by 0x403DE31: gem5::X86ISA::Walker::WalkerState::setupWalk(unsigned long) (pagetable_walker.cc:607)
==1267085==    by 0x404030F: gem5::X86ISA::Walker::WalkerState::startWalk() (pagetable_walker.cc:235)
```

```
==1267085== 472 (176 direct, 296 indirect) bytes in 1 blocks are definitely lost in loss record 25,225 of 29,787
==1267085==    at 0x72BFFA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==1267085==    by 0x403EBE1: gem5::X86ISA::Walker::WalkerState::stepWalk(gem5::Packet*&) (pagetable_walker.cc:531)
==1267085==    by 0x404128A: gem5::X86ISA::Walker::WalkerState::recvPacket(gem5::Packet*) (pagetable_walker.cc:634)
```

https://github.com/gem5/gem5/blob/develop/src/arch/x86/pagetable_walker.cc#L607
https://github.com/gem5/gem5/blob/develop/src/arch/x86/pagetable_walker.cc#L531

### In the IDE controller

This *could* be our culprit since we leaked 128KiB in one leak. Maybe when we're reading more off of the disk during systemd boot this is leaking more? It only happens on resetting the device state, but that happens in `updateState` on line 780. I wonder if something changed so that resetting is happening more often now?

```
==1267085== 131,072 bytes in 1 blocks are possibly lost in loss record 29,548 of 29,787
==1267085==    at 0x72C15C3: operator new[](unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==1267085==    by 0x29D7D89: gem5::IdeDisk::reset(int) (ide_disk.cc:148)
==1267085==    by 0x29DC02C: gem5::IdeDisk::IdeDisk(gem5::IdeDiskParams const&) (ide_disk.cc:76)
```

https://github.com/gem5/gem5/blob/develop/src/dev/storage/ide_disk.cc#L148

### These are just done once at initialization or a small number of times during simulation

```
==1267085== 80 bytes in 2 blocks are definitely lost in loss record 16,893 of 29,787
==1267085==    at 0x72BFFA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==1267085==    by 0x4012B66: gem5::X86ISA::ISA::newPCState(unsigned long) const (isa.hh:69)
==1267085==    by 0x329EFCC: clearArchRegs (simple_thread.hh:247)
```

```
==1267085== 360 (64 direct, 296 indirect) bytes in 1 blocks are definitely lost in loss record 24,739 of 29,787
==1267085==    at 0x72BFFA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==1267085==    by 0x315E2A9: BaseGlobalEventTemplate (global_event.hh:163)
==1267085==    by 0x315E2A9: GlobalEvent (global_event.hh:197)
==1267085==    by 0x315E2A9: gem5::GlobalSimLoopExitEvent::GlobalSimLoopExitEvent(unsigned long, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, unsigned long) (sim_events.cc:59)
==1267085==    by 0x315E9D7: gem5::exitSimLoop(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, unsigned long, unsigned long, bool) (sim_events.cc:95)
==1267085==    by 0x31CB065: gem5::pseudo_inst::m5exit(gem5::ThreadContext*, unsigned long) (pseudo_inst.cc:183)
```

[valgrind.txt](https://github.com/user-attachments/files/18322326/valgrind.txt)

Harshil2107:
This PR fixes the memory leaks highlighted in #1903

### DIFF
--- a/src/arch/x86/pagetable_walker.cc
+++ b/src/arch/x86/pagetable_walker.cc
@@ -246,8 +246,11 @@ Walker::WalkerState::startWalk()
             assert(fault == NoFault || read == NULL);
             state = nextState;
             nextState = Ready;
-            if (write)
+            if (write) {
                 walker->port.sendAtomic(write);
+                // delete the write packet as it is not needed anymore
+                delete write;
+            }
         } while (read);
         state = Ready;
         nextState = Waiting;
@@ -270,6 +273,10 @@ Walker::WalkerState::startFunctional(Addr &addr, unsigned &logBytes)
         PacketPtr write = NULL;
         fault = stepWalk(write);
         assert(fault == NoFault || read == NULL);
+        // delete the write packet if it exists
+        if (write) {
+            delete write;
+        }
         state = nextState;
         nextState = Ready;
     } while (read);
--- a/src/dev/storage/ide_disk.cc
+++ b/src/dev/storage/ide_disk.cc
@@ -64,7 +64,8 @@ namespace gem5
 {
 
 IdeDisk::IdeDisk(const Params &p)
-    : SimObject(p), image(p.image), diskDelay(p.delay), ideDiskStats(this),
+    : SimObject(p), image(p.image), diskDelay(p.delay), dataBuffer(nullptr),
+      ideDiskStats(this),
       dmaTransferEvent([this]{ doDmaTransfer(); }, name()),
       dmaReadWaitEvent([this]{ doDmaRead(); }, name()),
       dmaWriteWaitEvent([this]{ doDmaWrite(); }, name()),
@@ -145,7 +146,9 @@ void
 IdeDisk::reset(int id)
 {
     // initialize the data buffer and shadow registers
-    dataBuffer = new uint8_t[MAX_DMA_SIZE];
+    // if the data buffer has not been allocated yet
+    if (!dataBuffer)
+        dataBuffer = new uint8_t[MAX_DMA_SIZE];
 
     memset(dataBuffer, 0, MAX_DMA_SIZE);
     memset(&cmdReg, 0, sizeof(CommandReg_t));


src/dev/storage/ide_disk.cc
```diff
@@ -144,6 +144,10 @@ IdeDisk::~IdeDisk()
 void
 IdeDisk::reset(int id)
 {
+    // free the data buffer if it exists
+    if (dataBuffer)
+        delete [] dataBuffer;
+
     // initialize the data buffer and shadow registers
     dataBuffer = new uint8_t[MAX_DMA_SIZE];
```
powerjg: A better solution would be to set `dataBuffer` to `nullptr` in initialization and then only allocate `if (dataBuffer == nullptr)`

Harshil2107: I initialized the `dataBuffer` to `nullptr`. I noticed that the `reset` function is being called multiple times, so wouldnt it be better to initialize `databuffer` as `nullptr` and then free before using new?

powerjg: > I noticed that the reset function is being called multiple times

How did you notice this? I tried running this test and I didn't see that happening.

> wouldnt it be better to initialize databuffer as nullptr and then free before using new?

You should only call `new` once in the entire simulation. You'll notice that just below this `new` call is a `memset` call, which is all that's needed if `reset` is called multiple times. So, the logic should be: the first time `reset` is called you allocate the data. All other times, you don't call new. Then, finally, you `delete` in the constructor.

src/arch/x86/pagetable_walker.cc
```diff
@@ -270,6 +273,9 @@ Walker::WalkerState::startFunctional(Addr &addr, unsigned &logBytes)
         PacketPtr write = NULL;
         fault = stepWalk(write);
         assert(fault == NoFault || read == NULL);
+        // delete the write packet if it exists
+        if (write)
+            delete write;
```
powerjg: Curly braces 😄

src/dev/storage/ide_disk.hh
```diff
@@ -231,7 +231,7 @@ class IdeDisk : public SimObject
     /** Drive identification structure for this disk */
     struct ataparams driveID;
     /** Data buffer for transfers */
-    uint8_t *dataBuffer;
+    uint8_t *dataBuffer=nullptr;
```
powerjg: 1. Needs spaces around the `=`.
2. Probably should be in the intializer list in the constructor since that's where all the other initializations are.

src/dev/storage/ide_disk.cc
Harshil2107: > How did you notice this? I tried running this test and I didn't see that happening.

I mean that it is called twice in the same file, line https://github.com/gem5/gem5/blob/c66e4ca40c7c3d71703fb67ef1f4485d0637b8a1/src/dev/storage/ide_disk.cc#L76 and https://github.com/gem5/gem5/blob/c66e4ca40c7c3d71703fb67ef1f4485d0637b8a1/src/dev/storage/ide_disk.cc#L784. I think a better way to frame my comment would be that the reset function can be called more than once.

I will update the initialization code.

powerjg (commented):
lgtm, but the commits are messy. Since this is addressing two unrealated memory leaks, the changes should be in two commits. This could be done with either two PRs that have squashed commits or a single PR with two commits. So, you can either rebase this PR or you can create separate PRs.

Harshil2107:
I rebased the PR to have 2 commits, each having changes to fix a leak.

BobbyRBruce (approved):

PR #1905, opened 2025-01-06
arch-arm: setMiscRegs: Show CSPR.d in debug log on writes to CPSR

bpfoley:
The interrupt mask bits DAIF are typically all accessed together, so show all 4 of them on update. See, eg https://developer.arm.com/documentation/ddi0595/2021-06/AArch64-Registers/DAIF--Interrupt-Mask-Bits

Change-Id: I122b1e9c1804b6077c1e91f35ee6fdd92a06f292


powerjg (approved):
Thanks for the contribution!

giactra (approved):

PR #1910, opened 2025-01-08
arch-x86: Fix typo for readlinkat syscall

taoky:
readlinkat shall use readlinkatFunc instead of readlinkFunc, otherwise it would always throw -EFAULT as arg regs are incorrect.

### DIFF
--- a/src/arch/x86/linux/syscall_tbl32.cc
+++ b/src/arch/x86/linux/syscall_tbl32.cc
@@ -363,7 +363,7 @@ SyscallDescTable<EmuLinux::SyscallABI32> EmuLinux::syscallDescs32 = {
     { 302, "renameat" },
     { 303, "linkat" },
     { 304, "symlinkat" },
-    { 305, "readlinkat", readlinkFunc<X86Linux32> },
+    { 305, "readlinkat", readlinkatFunc<X86Linux32> },
     { 306, "fchmodat" },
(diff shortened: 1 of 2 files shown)


powerjg (commented):
This seems right to me, but I'm super surprised that we haven't noticed until now. Any idea why this wouldn't have been fixed prior to now? I.e., what changed to make you find this bug?

taoky:
> This seems right to me, but I'm super surprised that we haven't noticed until now. Any idea why this wouldn't have been fixed prior to now? I.e., what changed to make you find this bug?

I'm comparing the syscall trace between gem5 and qemu-user, and found that readlinkat() returns differently between the two.

powerjg:
> I'm comparing the syscall trace between gem5 and qemu-user, and found that readlinkat() returns differently between the two.

Cool approach! Thanks for finding and fixing a bug!

powerjg (approved):

mattsinc:
> > I'm comparing the syscall trace between gem5 and qemu-user, and found that readlinkat() returns differently between the two.
> 
> Cool approach! Thanks for finding and fixing a bug!

I think I implemented this a while back.  I certainly did not compare to QEMU, although I am surprised about the bug as in theory the daily/weekly tests that needed this support would have failed without this tweak.  But I guess we can just let the tests run and go from there?

PR #1915, opened 2025-01-09
arch-vega: Add SDWA to v_cmp_ne_u32

abmerop:


powerjg (approved):

src/arch/amdgpu/vega/insts/vopc.cc
```diff
@@ -5223,12 +5223,17 @@ namespace VegaISA
...
-        panic_if(isSDWAInst(), "SDWA not supported for %s", _opcode);
-        panic_if(isDPPInst(), "DPP not supported for %s", _opcode);
+        auto cmpImpl = [](uint32_t a, uint32_t b) { return a != b ? 1 : 0; };
 
-        for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
-            if (wf->execMask(lane)) {
-                vcc.setBit(lane, src0[lane] != src1[lane] ? 1 : 0);
+        if (isSDWAInst()) {
+            sdwabHelper<uint32_t>(gpuDynInst, cmpImpl);
+        } else if (isDPPInst()) {
```
mattsinc, line 5230: I'm not sure this else if is really needed versus the panic_if above, but it's also fine as it sets up the code nicely if we do add later.

src/arch/amdgpu/vega/insts/vopc.cc
```diff
@@ -5223,12 +5223,17 @@ namespace VegaISA
...
 
-        panic_if(isSDWAInst(), "SDWA not supported for %s", _opcode);
-        panic_if(isDPPInst(), "DPP not supported for %s", _opcode);
+        auto cmpImpl = [](uint32_t a, uint32_t b) { return a != b ? 1 : 0; };
 
-        for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
-            if (wf->execMask(lane)) {
-                vcc.setBit(lane, src0[lane] != src1[lane] ? 1 : 0);
+        if (isSDWAInst()) {
+            sdwabHelper<uint32_t>(gpuDynInst, cmpImpl);
```
mattsinc, line 5229: I'm assuming you tested that the bHelper function has the needed functionality for this?

mattsinc (approved):

src/arch/amdgpu/vega/insts/vopc.cc
abmerop: Yes, it's already used elsewhere too

src/arch/amdgpu/vega/insts/vopc.cc
abmerop: Eventually all VOPC execute codes could look like this where there are three cases for sdwa, dpp, and neither. In fact, this file in particular could probably just be a big table of lambda functions which handles the three cases implicitly.  I am not planning to do any kind of refactor like that any time soon though

PR #1916, opened 2025-01-09
arch-vega: Fix LDS/buffer load/store x2,x3,x4

abmerop:
These instructions have the wrong index values when writing to the GPUDynInst d_data field, which expects data to be contiguous in the d_data buffer. This results in loads/stores being mis-aligned with zeros padding the empty space.

Set the correct index values (see flat.cc for example) so that the data values are in the correct locations. This should help fix issues with nanoGPT and MNIST spotted during the ISCA tutorial.

### DIFF
--- a/src/arch/amdgpu/vega/insts/ds.cc
+++ b/src/arch/amdgpu/vega/insts/ds.cc
@@ -4557,11 +4557,11 @@ namespace VegaISA
         for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
             if (gpuDynInst->exec_mask[lane]) {
                 (reinterpret_cast<VecElemU32*>(
-                    gpuDynInst->d_data))[lane * 4] = data0[lane];
+                    gpuDynInst->d_data))[lane * 3] = data0[lane];
                 (reinterpret_cast<VecElemU32*>(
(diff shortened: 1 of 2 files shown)


abmerop:
This should fix the issue with larger batch sizes in MNIST / nanoGPT observed at the ISCA tutorial (finally!), where larger batches were getting zero values in the hidden layers causing accuracy to be way off / nanoGPT outaf ao bnt woti.

powerjg (approved):

mattsinc (approved):
@abmerop while I have no reason to doubt you tested and validated these, where do I look in the ISA guide to validate the updated indices?  They are basically magic numbers

abmerop:
> @abmerop while I have no reason to doubt you tested and validated these, where do I look in the ISA guide to validate the updated indices? They are basically magic numbers

You can look at buffer_store_dwordx3 for example in section 12 of any of the ISA guides. However, that only shows the data placement for each lane. What is more important is how gem5 is placing the data into packets which go to memory: https://github.com/gem5/gem5/blob/stable/src/arch/amdgpu/vega/gpu_mem_helpers.hh#L122 . You can see the data should be contiguous (lane * N, where N = number of dwords)

PR #1911, opened 2025-01-08
misc: Fixup optional libs being required

odanrc:
hdf5, protobuf, libpng and kvm should not be required in the gem5 lib. Fix this by checking in the scons files if they are present before declaring the source files that use them.

PR #1917, opened 2025-01-09
scons: Unbreak builds if kvm, libpng, or libhdf5 aren't found

bpfoley:
(The kvm headers aren't present on macOS, and libpng and libhdf5 aren't mandatory on Linux builds.)

The recent add_tags exposed this for some reason I don't understand.

Change-Id: I0548bb9f7ced9a470e569df8d23e5dd72f33fda0


powerjg:
Is this a duplicate of #1911, a better solution to the same problem, or a different problem?

PS: Our runners are having issues. We're working to fix this ASAP.

bpfoley:
> Is this a duplicate of #1911, a better solution to the same problem, or a different problem?
> 
> PS: Our runners are having issues. We're working to fix this ASAP.

Ah! It looks like a partial dupe of #1911. Although #1911 doesn't seem to handle the !HAVE_KVM case.

odanrc:
Thanks for catching that! I have just updated that PR to also fix kvm and libpng

PR #1926, opened 2025-01-13
cpu-o3: add retry resp to LSQ with throttling params

saul44203:
This patch aims to add some latency modeling to the O3's LSQ dcache port side. Specifically it adds two config parameters named `recvRespMaxCachelines` and `recvRespBufferSize` to limit the the maximum number of cachelines and bytes (respectively) per cycle that can be processed. When one of this limits is exceeded the current packet is throttled and scheduled for retry on the next cycle (potentially spanning multiple retries), until it's fully processed.

Change-Id: I41e71b92357a828a8695e95f53c170342d6c0293

### DIFF
--- a/src/cpu/o3/BaseO3CPU.py
+++ b/src/cpu/o3/BaseO3CPU.py
@@ -206,3 +206,14 @@ class BaseO3CPU(BaseCPU):
         TournamentBP(numThreads=Parent.numThreads), "Branch Predictor"
     )
     needsTSO = Param.Bool(False, "Enable TSO Memory model")
+
+    recvRespThrottling = Param.Bool(
+        False, "Enable load receive response throttling in the LSQ"
+    )
+    recvRespMaxCachelines = Param.Unsigned(
+        1,
+        "Maximum number of different receive response cachelines per cycle",
+    )
+    recvRespBufferSize = Param.Unsigned(
+        64, "Maximum number of receive response bytes per cycle"
+    )
--- a/src/cpu/o3/lsq.cc
+++ b/src/cpu/o3/lsq.cc
@@ -65,8 +65,47 @@ namespace o3
 {
 
 LSQ::DcachePort::DcachePort(LSQ *_lsq, CPU *_cpu) :
-    RequestPort(_cpu->name() + ".dcache_port"), lsq(_lsq), cpu(_cpu)
+    RequestPort(_cpu->name() + ".dcache_port"), lsq(_lsq), cpu(_cpu),
+    dcachePortStats(_cpu)
 {}
+
+LSQ::DcachePort::DcachePortStats::DcachePortStats(CPU* _cpu)
+    : statistics::Group(_cpu),
+      ADD_STAT(numRecvResp, statistics::units::Count::get(),
+              "Number of received responses"),
+      ADD_STAT(numRecvRespBytes, statistics::units::Byte::get(),
+              "Number of received response bytes"),
+      ADD_STAT(recvRespAvgBW,
+               statistics::units::Rate<statistics::units::Byte,
+                                       statistics::units::Cycle>::get(),
+               "Average bandwidth of received responses"),
+      ADD_STAT(recvRespAvgSize,
+               statistics::units::Rate<statistics::units::Byte,
+                                       statistics::units::Count>::get(),
+               "Average packet size per received response"),
+      ADD_STAT(recvRespAvgRate,
+               statistics::units::Rate<statistics::units::Count,
+                                       statistics::units::Cycle>::get(),
+               "Average rate of received responses per cycle"),
+      ADD_STAT(recvRespAvgRetryRate,
+               statistics::units::Rate<statistics::units::Count,
+                                       statistics::units::Count>::get(),
+               "Average retry rate per received response"),
+      ADD_STAT(numSendRetryResp, statistics::units::Count::get(),
+               "Number of retry responses sent")
+{
+    recvRespAvgBW.precision(2);
+    recvRespAvgBW = numRecvRespBytes / _cpu->baseStats.numCycles;
+
+    recvRespAvgSize.precision(2);
+    recvRespAvgSize = numRecvRespBytes / numRecvResp;
+
+    recvRespAvgRate.precision(2);
+    recvRespAvgRate = numRecvResp / _cpu->baseStats.numCycles;
+
+    recvRespAvgRetryRate.precision(2);
+    recvRespAvgRetryRate = numSendRetryResp / numRecvResp;
+}
 
 LSQ::LSQ(CPU *cpu_ptr, IEW *iew_ptr, const BaseO3CPUParams &params)
     : cpu(cpu_ptr), iewStage(iew_ptr),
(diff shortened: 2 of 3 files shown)


src/cpu/o3/lsq.cc
```diff
@@ -1411,6 +1465,92 @@ LSQ::SplitDataRequest::isCacheBlockHit(Addr blockAddr, Addr blockMask)
 bool
 LSQ::DcachePort::recvTimingResp(PacketPtr pkt)
 {
+    // Throttling enabled
+    if (lsq->recvRespThrottling) {
+        // Reset counters/flags on new cycle
+        if (curTick() > lsq->recvRespLastActiveTick + 1) {
```
powerjg: Why not `curTick() > lsq->recvRespLastActiveTick` or  `curTick() >= lsq->recvRespLastActiveTick + 1`?

It seems that you're checking that the `curTick()` is 2 *ticks* (not cycles) away from the last active tick.

src/cpu/o3/lsq.cc
```diff
@@ -1411,6 +1465,92 @@ LSQ::SplitDataRequest::isCacheBlockHit(Addr blockAddr, Addr blockMask)
...
+        if (isNewCacheline) {
+            lsq->recvRespCachelines++;
+            lsq->recvRespLastCachelineAddr = pktCachelineAddr;
+        }
+
+        // Throttling this cycle
+        if (lsq->recvRespThrottled) {
+            // Check if tick is aligned to clock edge to correct 1 cycle delay
+            bool isAlignedTick = curTick() % cpu->clockPeriod() == 0;
+            Tick nextCycle = cpu->clockEdge(Cycles(isAlignedTick));
```
powerjg: Just commenting for others if they are confused by this code like I was. The gem5 clocked object APIs don't fit this use case. From [the documentation](https://doxygen.gem5.org/develop/classClocked.html#a6917375aeb29de3471027a549c7508fb):

> Tick Clocked::nextCycle
> Based on the clock of the object, determine the start tick of the first cycle that is at least one cycle in the future.
> When curTick() is at the current cycle edge, this returns the next clock edge. When calling this during the middle of a cycle, this returns 2 clock edges in the future.

> Cycles Clocked::curCycle
> Determine the current cycle, corresponding to a tick aligned to a clock edge.
> Returns
> When curTick() is on a clock edge, return the Cycle corresponding to that clock edge. When curTick() is not on a clock edge, return the Cycle corresponding to the next clock edge.

So, neither of these functions do what the author is trying to do here which is to get either the cycle that is exactly 1 cycle away, or the next closest cycle.

powerjg (commented):
Just one small question, but overall this looks good to me.

src/cpu/o3/lsq.cc
saul44203: Yeah, it's a bit confusing. Here `recvRespLastActiveTick` has the latest tick that was processed, so If the current tick differs from it by more than 1 tick I assume we are on another cycle. From what I know and tested this is safe to assume, but let me know if I should change this.

src/cpu/o3/lsq.cc
```diff
@@ -1411,6 +1465,92 @@ LSQ::SplitDataRequest::isCacheBlockHit(Addr blockAddr, Addr blockMask)
 bool
 LSQ::DcachePort::recvTimingResp(PacketPtr pkt)
 {
+    // Throttling enabled
+    if (lsq->recvRespThrottling) {
```
giactra: For readibility purposes, it would be nice if we could move all this logic on a separate method

src/cpu/o3/lsq.cc
```diff
@@ -1411,6 +1465,92 @@ LSQ::SplitDataRequest::isCacheBlockHit(Addr blockAddr, Addr blockMask)
...
+        Addr pktCachelineAddr = addrBlockAlign(pkt->getAddr(),
+                                               cpu->cacheLineSize());
+
+        // Check limits
+        bool isNewCacheline = pktCachelineAddr
+                              != lsq->recvRespLastCachelineAddr;
+        bool maxCachelinesReached = (lsq->recvRespCachelines + isNewCacheline)
+                                    > lsq->recvRespMaxCachelines;
+        int freeBufferSize = lsq->recvRespBufferSize - lsq->recvRespBytes;
+        bool maxBytesReached = pkt->getSize() > freeBufferSize;
```
giactra: Local vars should use snake_case

src/cpu/o3/lsq.cc
```diff
@@ -1411,6 +1465,92 @@ LSQ::SplitDataRequest::isCacheBlockHit(Addr blockAddr, Addr blockMask)
...
 LSQ::DcachePort::recvTimingResp(PacketPtr pkt)
 {
+    // Throttling enabled
+    if (lsq->recvRespThrottling) {
+        // Reset counters/flags on new cycle
+        if (curTick() > lsq->recvRespLastActiveTick + 1) {
+            lsq->recvRespBytes = 0;
+            lsq->recvRespCachelines = 0;
+            lsq->recvRespLastCachelineAddr = 0;
+            lsq->recvRespThrottled = false;
```
giactra: Is it necessary to make recvRespThrottled a member variable? It looks like it could be made a local one, to be calculated from the others.

src/cpu/o3/BaseO3CPU.py
```diff
@@ -206,3 +206,14 @@ def support_take_over(cls):
         TournamentBP(numThreads=Parent.numThreads), "Branch Predictor"
     )
     needsTSO = Param.Bool(False, "Enable TSO Memory model")
+
+    recvRespThrottling = Param.Bool(
+        False, "Enable load receive response throttling in the LSQ"
```
giactra, line 211: From this comment it looks like the throttling should work for loads only.. However I don't see the type of the Packet response being checked... I believe stores also get a WriteResp (with no data), and we unfortunately pass non-load/store instructions through the LSQ like the HTM and TLBIs commands:

https://github.com/gem5/gem5/blob/stable/src/cpu/o3/lsq.cc#L814C1-L814C31

These have a non-zero size.

All this to say we need to be careful as there are several request types

saul44203: You're right, I missed this. Now I'm filtering only for `ReadResp`.

src/cpu/o3/lsq.cc
saul44203: In the end I changed this because it was not a safe assumption. Hope it's better now.

powerjg (dismissed):

src/cpu/o3/lsq.cc
```diff
@@ -1408,9 +1461,106 @@ LSQ::SplitDataRequest::isCacheBlockHit(Addr blockAddr, Addr blockMask)
     return is_hit;
 }
 
+bool
+LSQ::DcachePort::throttleReadResp(PacketPtr pkt) {
+    // Check if tick is unaligned to correct +1 curCycle delta
```
erin-le:
```suggestion
bool
LSQ::DcachePort::throttleReadResp(PacketPtr pkt)
{
    // Check if tick is unaligned to correct +1 curCycle delta
```

erin-le (commented):
Hello, there's a minor issue with whitespace for one of the functions. It would be nice if it could be fixed before merging, but it isn't a major problem. Thank you for your contribution!

src/cpu/o3/lsq.cc
saul44203: No problem. Should be done.

powerjg (approved):

PR #1927, opened 2025-01-13
sim,tests: Add a tag for SimObjects

odanrc:
This tag can be used to determine which files are needed when sim/sim_object.hh is included in a header file. For example, when declaring a unit test, this tag makes the SConscript declaration much simpler.


erin-le (commented):
I'm not too familiar with how SCons works, but overall this looks fine and doesn't result in any build errors for me. 

Just for my own knowledge, I was wondering how the dependencies are determined? I was comparing the include directives in sim_object.hh and sim_object.cc with the files tagged with `'gem5 simobject'` and noticed that most of the newly tagged files had a corresponding header file included in sim_object.hh or sim_object.cc, with the exception of tags.cc and globals.cc.

odanrc:
> I'm not too familiar with how SCons works, but overall this looks fine and doesn't result in any build errors for me.
> 
> Just for my own knowledge, I was wondering how the dependencies are determined? I was comparing the include directives in sim_object.hh and sim_object.cc with the files tagged with `'gem5 simobject'` and noticed that most of the newly tagged files had a corresponding header file included in sim_object.hh or sim_object.cc, with the exception of tags.cc and globals.cc.

Good catch! The dependencies are determined by being included, as you say.

I wrote this commit 4 years ago, and maybe back then these files were included in sim_object; however, they do not seem to be needed by sim obj anymore, so I have removed the sim obj tag from them.

erin-le:
In that case, should `port.cc` (sim/port.hh included in sim_object.hh) and `logging.cc` (base/logging.hh included in sim_object.cc) also be tagged with `'gem5 simobject'`? These seem to be the only files that aren't either tagged with `'gem5 simobject'` or through env.TagImplies, and also aren't autogenerated files.

odanrc:
Regarding base/logging.cc, I don't think we want it - we probably want to keep using the mock (src/base/gtest/logging.*), which should already be included when testing due to its gtest tag.

For the port, it may be missing, but my knowledge of compilers is not enough to be sure. If you check the sim_object files, there is no Port being instantiated/used - the function is virtual, and fatals() as soon as it starts. Since there was no compiler complaint, I am assuming it is alright to keep it as is (i.e., the compiler does not need to read the implementation code). On the one hand, if you prefer to be on the safe side and make it future proof, let me know and I can add the tag to ports; on the other hand, keep in mind that this will possibly mean one more compilation unit for every sim object that has unit tests.

erin-le (approved):
Thanks for your explanation! If it's probably fine, then no need to change it.

PR #1936, opened 2025-01-16
Fix-append-kernel-arg-risv

powerjg:
Previously, the device tree was being generated when the workload was
initially set in `set_kernel_disk_workload` (when the disk image was
set). However, this meant that if you changed the kernel boot parameters
after the initial call they would not show up in the dtb file. The
RISC-V kernel picks up the parameters from the dtb file, not what is
passed in via the gem5 workload.

The solution is to generate the dtb file in `_pre_instantiate` instead
of in `_add_disk_to_board`

### DIFF
--- a/src/python/gem5/components/boards/riscv_board.py
+++ b/src/python/gem5/components/boards/riscv_board.py
@@ -541,6 +541,17 @@ class RiscvBoard(AbstractSystemBoard, KernelDiskWorkload, SEBinaryWorkload):
             else:
                 self.workload.kernel_addr = 0x0
                 self.workload.entry_point = 0x80000000
+
+            # Set up the device tree. We need to wait until pre-instantiate to
+            # do this since the workload could change until this point.
+            # Default DTB address if bbl is built with --with-dts option
+            self.workload.dtb_addr = 0x87E00000
+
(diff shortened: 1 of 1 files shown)


src/python/gem5/components/boards/riscv_board.py
```diff
@@ -541,6 +541,17 @@ def _pre_instantiate(self, full_system: Optional[bool] = None):
...
+
+            # Set up the device tree. We need to wait until pre-instantiate to
+            # do this since the workload could change until this point.
+            # Default DTB address if bbl is built with --with-dts option
+            self.workload.dtb_addr = 0x87E00000
+
+            self.generate_device_tree(m5.options.outdir)
+            self.workload.dtb_filename = os.path.join(
+                m5.options.outdir, "device.dtb"
+            )
```
BobbyRBruce, line 553: Shouldn't this only ben run for FS mode?

powerjg: Isn't it?

BobbyRBruce: Yip! I misread the change.

BobbyRBruce (approved):

PR #1942, opened 2025-01-21
scons:  Add a target to generate a compilation database

yalavrinenko:
This update introduces a target for generating a `compile_commands.json` file, which is useful for improving IDE support and compatibility with tools such as `clangd`, `clang-tidy`, and others.

**Target Pattern**

The target follows this usage pattern:

```bash
scons build/{ISA}/compile_commands.json
```

This command generates the `compile_commands.json` file without triggering the main build process.

**Requirements**

The target is only available when the scons version is greater or equal to 4.0.

**Benefits**

Adding this target enhances integration with development environments and tooling, streamlining workflows for developers.

Jira Issue: https://gem5.atlassian.net/browse/GEM5-1318

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -85,6 +85,7 @@ import SCons
 import SCons.Node
 import SCons.Node.FS
 import SCons.Tool
+import SCons.Errors
 
 if getattr(SCons, '__version__', None) in ('3.0.0', '3.0.1'):
     # Monkey patch a fix which appears in version 3.0.2, since we only
@@ -264,6 +265,15 @@ Targets:
 
         scons build/SPARC/base/bitunion.test.opt
         build/SPARC/base/bitunion.test.opt
+
+        To generate the compile_commands.json, you can use a target:
+
+        scons build/{{ISA}}/compile_commands.json
+
+        The {{ISA}} is a target Instruction Set Architecture (X86, ARM,
+        RISCV, etc.). This command creates a compile_commands.json in the
+        respective build directory. You can generate a compile_commands.json
+        only with scons version 4.0+.
 """, append=True)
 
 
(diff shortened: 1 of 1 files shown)


SConstruct
```diff
@@ -536,6 +545,17 @@ for variant_path in variant_paths:
...
+    cdb_tool = SCons.Tool.FindTool(['compilation_db'], env)
+
+    if cdb_tool:
+        env['COMPILATIONDB_USE_ABSPATH'] = True
+        env.Tool(cdb_tool)
+
+        cdb_path = f"{variant_path}/compile_commands.json"
+        cdb_build_target = env.CompilationDatabase(cdb_path)
+
+        Alias(f"{variant_path}/compile_commands", cdb_build_target)
```
powerjg: Is the alias required? Can we have the target be `compile_commands.json` instead of `compile_commands`?

I ask this because all other targets are explicit files. (e.g., `gem5.opt` or the path to a `.o` file)

powerjg (commented):
Seems good to me. One small question below.

SConstruct
yalavrinenko: > Can we have the target be compile_commands.json instead of compile_commands?

We can change the target to `build/{ISA}/compile_commands.json` instead of `build/{ISA}/compile_commands`

Moreover, we no need to use `Alias` in that case. 
Currently,  `Alias` is used to force the target name to `build/{ISA}/compile_commands`.

I will reflect this change is PR.

powerjg (dismissed):
lgtm. I don't see any harm in this. However, I tried to play around with it in VSCode and I wasn't able to understand the use case. That said, if it's useful for someone else, it seems good to me to merge.

yalavrinenko:
I updated a PR with a possible fix for CI.  For some reason `FindTool` raises an exception before `if-exist`  check. The `try-except` block was introduced as a possible solution. If it is not ok with code guideline, I can make a more complicated checks that `compilation_db` module exists.

yalavrinenko:
> lgtm. I don't see any harm in this. However, I tried to play around with it in VSCode and I wasn't able to understand the use case. That said, if it's useful for someone else, it seems good to me to merge.

I use it in VSCode with `clangd` instead of `intellisense` and it looks pretty good. It can index all code and makes static analyzer works pretty well. 

In `settings.json`:
```json
"C_Cpp.intelliSenseEngine": "disabled",
"clangd.arguments": [
    "--compile-commands-dir=${workspaceFolder}/build/ALL"
],
```

Anyway, I made this only because I didn't find any solution on google/stack owerflow/etc, so I do not insist on merging it.

ipocentro87:
@powerjg This change helped me setting up Emacs lsp-mode successfully, and I expect it to work with any other LSP integration on other editors.
I'd be ok for a merge.

erin-le (approved):

SConstruct
```diff
@@ -536,6 +546,23 @@ for variant_path in variant_paths:
...
+        # that tool exists and makes FindTool function useless in some way.
+        cdb_tool = SCons.Tool.FindTool(['compilation_db'], env)
+
+        if cdb_tool:
+            env['COMPILATIONDB_USE_ABSPATH'] = True
+            env.Tool(cdb_tool)
+
+            cdb_path = f"{variant_path}/compile_commands.json"
+            env.CompilationDatabase(cdb_path)
+    except (SCons.Errors.SConsEnvironmentError, SCons.Errors.UserError):
```
giactra, line 562: It has to be noted that gem5 does not build anymore with scons < 4.0.0 because SCons.Errors.SConsEnvironmentError is undefined. Either we fix this or we update gem5 requirements:

https://www.gem5.org/documentation/learning_gem5/part1/building/

giactra: https://github.com/gem5/gem5/pull/2141

PR #1943, opened 2025-01-22
arch-riscv: Implement SVNAPOT Extension

ncppd:
This commit introduces support for the RISC-V SVNAPOT Extension, which enables the use of contiguous 4K pages by combining them into a 64KB page, thereby increasing the reach of the TLB.

I tested this on Linux 6.9 with transparent huge pages enabled. To support THP for 64KB pages, I applied this [patch](https://lore.kernel.org/kvm-riscv/20240508191931.46060-5-alexghiti@rivosinc.com/T/). Additionally, I verified its functionality using glibc tunables, specifically `glibc.malloc.hugetlb=$(( 64 << 10 ))`.


powerjg (approved):
Thanks for the contribution! This looks good to me. I'm (happily) surprised the code was so simple!

powerjg:
If someone else merges this, it looks like it should be squashed, no merged.

PR #1944, opened 2025-01-23
mem-ruby: remove incorrect actions in MOESI_CMP_directory

hexengraf:
The states MI, OI, and MII in L2 don't have a DirEntry allocated, since they're preceded by states that own the block and therefore use CacheEntry instead of DirEntry. In fact, MI, OI, and MII don't even have a CacheEntry allocated, at this point everything is already deallocated except for the TBE entry.

Remove 'removeFromDir' action from a few transitions coming from these states. After the changes to enable support for multiple Ruby protocols, these spurious actions cause assertion failure in NetDest.cc.


powerjg (approved):
I'm surprised this didn't cause errors before.

PR #1945, opened 2025-01-23
base: Refactor base/random to avoid memory leak

BobbyRBruce:
There was a memory leak occurring prior to this fix which can be reproduced with:

```sh
scons build/VEGA_X86/gem5.opt -j `nproc`

/usr/bin/time -l ./build/VEGA_X86/gem5.opt configs/example/ruby_gpu_random_test.py --WB_L2 --test-length 5000000 --num-dmas 0
```

The `/usr/bin/time -l` will output the maximum memory consumption of this process. On my machines It was getting to 27GB.

With this change gem5 now peaks at around 6GB of memory usage.

This incorporates two changes:

1) Instances of `static RandomPtr` were not creating singular RNG
   instances like assumed. Ergo in casees where this was depended
   upon to stop the over creation of RNGs, multiple were being
   created. The `RandomPtr` object was static but the RNG it pointed
   towards was always changing on every use. This has been fixed.
2) The RNG instance list has been refactored to avoid accidental
   inclusion of the same RNG more than once and better checks have
   been added as to explicitly remove RNGs from this instance list
   when they are deleted.

Tests updated accordingly.

### DIFF
--- a/src/base/random.cc
+++ b/src/base/random.cc
@@ -46,32 +46,30 @@
 namespace gem5
 {
 
+Random::Random(): Random(globalSeed){}
 Random::Random(uint32_t s)
 {
     init(s);
 }
 
+
+
 Random::~Random()
 {
-    assert(instances);
+    if (instances) {
+        // Remove expired weak pointers
+        instances->erase(
+            std::remove_if(instances->begin(), instances->end(),
+                [](const std::weak_ptr<Random>& ptr) {
+                    return ptr.expired();
+                }),
+            instances->end());
 
-    int removed = 0;
-
-    instances->erase(std::remove_if(instances->begin(), instances->end(),
-    [&](const auto& s_ptr)
-    {
-        removed += s_ptr.expired();
-        return s_ptr.expired();
-    }));
-
-    // Can only remove one pointer
-    // since we are destroying one
-    // object
-    assert(removed == 1);
-
-    if (instances->empty()) {
-        delete instances;
-        instances = nullptr;
+        // Clean up instances if empty
+        if (instances->empty()) {
+            delete instances;
+            instances = nullptr;
+        }
     }
 }
 
--- a/src/base/random.cc
+++ b/src/base/random.cc
@@ -47,12 +47,11 @@ namespace gem5
 {
 
 Random::Random(): Random(globalSeed){}
+
 Random::Random(uint32_t s)
 {
     init(s);
 }
-
-
 
 Random::~Random()
 {
--- a/src/base/random.cc
+++ b/src/base/random.cc
@@ -46,8 +46,6 @@
 namespace gem5
 {
 
-Random::Random(): Random(globalSeed){}
-
 Random::Random(uint32_t s)
 {
     init(s);
--- a/src/base/random.hh
+++ b/src/base/random.hh
@@ -65,24 +65,26 @@ class Random
     using RandomPtr = std::shared_ptr<Random>;
     using Instances = std::vector<std::weak_ptr<Random>>;
 
-    static RandomPtr genRandom()
+    static RandomPtr genRandom(Random* r = nullptr)
     {
-      if (!instances)
-        instances = new Instances();
+        if (instances == nullptr)
+            instances = new Instances();
 
-      auto ptr = std::shared_ptr<Random>(new Random(globalSeed));
-      instances->emplace_back(ptr);
-      return ptr;
-    }
+        auto randpoint = r ? std::shared_ptr<Random>(r, [](Random*){})
+                         : std::shared_ptr<Random>(new Random());
+        // Check if randpoint is not already in the vector
+        bool exists = false;
+        for (const auto& weak_ptr : *instances) {
+            if (!weak_ptr.expired() && weak_ptr.lock() == randpoint) {
+            exists = true;
+            break;
+            }
+        }
+        if (!exists) {
+            instances->emplace_back(randpoint);
+        }
 
-    static RandomPtr genRandom(uint32_t s)
-    {
-      if (!instances)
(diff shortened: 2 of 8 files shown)


src/base/random.hh
```diff
@@ -65,24 +65,26 @@ class Random
...
+            instances = new Instances();
+
+        auto randpoint = r ? std::shared_ptr<Random>(r, [](Random*){})
+                         : std::shared_ptr<Random>(new Random());
+        // Check if randpoint is not already in the vector
+        bool exists = false;
+        for (const auto& weak_ptr : *instances) {
+            if (!weak_ptr.expired() && weak_ptr.lock() == randpoint) {
+            exists = true;
+            break;
```
powerjg: whitespace

src/base/random.cc
```diff
@@ -46,32 +46,30 @@
 namespace gem5
 {
 
+Random::Random(): Random(globalSeed){}
```
powerjg: Whitespace? I *think* this is a function definition but given that it's all on one line and just above the other definition, I can't tell if that was then intention.

src/base/random.test.cc
```diff
@@ -84,7 +84,7 @@ TEST(RandomCtor, DoubleDefaultConstruct)
     // First double corresponds to the default seed
     // 14514284786278117030llu / std::numeric_limits<uint64_t>::max()
     ASSERT_EQ(dut->random<double>(),
-      0.7868209548678020137657540544751100242137908935546875d);
+      0.7868209548678020137657540544751100242137908935546875);
```
powerjg, line 87: May still need the trailing "d" so that the compiler doesn't truncate.

src/mem/ruby/network/MessageBuffer.cc
```diff
@@ -205,7 +205,7 @@ MessageBuffer::peek() const
 Tick
 random_time()
 {
-    static Random::RandomPtr rng = Random::genRandom();
+    Random::RandomPtr rng = Random::genRandom();
```
powerjg: is this not going to generate a new heap instance of `Random::Random` each time `random_time` is called? We should create one instance and reuse it (e.g., with `static`). At least that is my understanding of the original code.

powerjg:
Two things:
1. This wasn't a "leak" (in that it wasn't detected by valgrind --leak-check) because pointers to the random numbers were never out of scope. However, it was a "leak" in the sense that the heap was growing indefinitely when running the `ruby_gpu_random_test`. This was found using the massif tool in `valgrind`.
2. tagging the people involved in the previous refactor: @aperais @odanrc. Tagging @giactra for visibility.

mattsinc (commented):
As far as I can tell this does not specifically involve a problem with the GPU, but rather VEGA is just being used to demonstrate a specific problem.  So I'll let @powerjg lead the review, but let me know if you need something from me.

src/base/random.hh
```diff
@@ -65,24 +65,26 @@ class Random
...
-      if (!instances)
-        instances = new Instances();
-
-      auto ptr = std::shared_ptr<Random>(new Random(s));
-      instances->emplace_back(ptr);
-      return ptr;
+        if (instances == nullptr)
+            instances = new Instances();
+
+        auto randpoint = r ? std::shared_ptr<Random>(r, [](Random*){})
```
odanrc, line 82: Is this a call to the constructor with a deleter? I think this line deserves some documentation of why we provide a nop deleter.

src/base/random.hh
```diff
@@ -87,6 +87,27 @@ class Random
...
+        if (instances == nullptr)
+            instances = new Instances();
+
+        auto randpoint = std::shared_ptr<Random>(new Random(s));
+        // Check if randpoint is not already in the vector
+        bool exists = false;
+        for (const auto& weak_ptr : *instances) {
+            if (!weak_ptr.expired() && weak_ptr.lock() == randpoint) {
+            exists = true;
+            break;
```
powerjg: whitespace

powerjg:
@BobbyRBruce can we get this in soon?

src/base/random.test.cc
BobbyRBruce: In CPP literal floating-points are double. The "F" is used to force float and "L" is used to force a long double.

"d" isn't an official CPP convention though GCC permits it. Clang does not which is how I found this bug.

src/cpu/testers/gpu_ruby_test/episode.cc
```diff
@@ -41,6 +41,8 @@
 namespace gem5
 {
 
+//static Random::RandomPtr rng(gem5::Random::genRandom());
```
powerjg: Probably shouldn't leave comments in the final PR

src/base/random.hh
powerjg: @BobbyRBruce make sure you address this concern :)

BobbyRBruce: The PR as it currently stands doesn't provide a nop deleter (I did in a previous iteration but I simplified it). However I have added a comment explaining this if-else setting of `randpoint` to alleviate any confusion.

BobbyRBruce:
I think i addressed all the comments and consider this PR ready for final review and merging. This should fix the Daily Tests.

src/base/random.hh
powerjg: What is purpose of the second argument to the constructor of `shared_ptr` here? It looks like an empty deleter, but maybe I'm not understanding correctly.

BobbyRBruce: It's a lambda function which servers as the deletion operator pointer. The purpose here is for the case where the deletion of the pointer is managed elsewhere, which in this case it is as this is.

powerjg (approved):

PR #1946, opened 2025-01-23
arch-x86: add LRU cache eviction in `decoder.hh`

Harshil2107:
In `decoder.hh` the `addrCacheMap` and the `instCacheMap` can grow unbound as there are no limits on their sizes. This change adds a size limit and implements an LRU cache eviction for the `addrCacheMap` and the `instCacheMap`.


odanrc:
I am not familiar with this code at all, but is there any reason that we are re-implementing LRU instead of re-using the one in src/mem/cache/replacement_policies/lru_rp.hh? The direction over the last years was to get rid of all untested LRU implementations in favor of that standardized implementation.

Harshil2107:
I am closing this PR as the solution in This PR does not seem to solve the issue correctly.

powerjg:
See #1957 for the new solution

PR #1947, opened 2025-01-24
arch-vega,gpu-compute: Fix architected flat scratch

abmerop:
There are several issues with architected flat scratch. Since this was written, new documentation in the RDNA3 ISA manual explains the address calculations better. This commit updates the following:

1) The memory request helper for scratch accesses of >1 DWORD has incorrect request size causing a misalignment assert.
2) Scratch base should be different per-wave and currently is per-shader and clobbering data due to having the same addresses for every wave.
3) Offsets in calcAddr can be negative. Use signed types.
4) Element size should always be 4. Based on RDNA3 ISA manual. No need to calculate payload size every time.
5) Add some asserts based on the text in RDNA3 ISA manual, Section 11.2.

This fixes PyTorch examples from the ISCA'24 tutorial for MI300X (recall that MI200 was fixed in #1916).

### DIFF
--- a/src/arch/amdgpu/vega/insts/op_encodings.hh
+++ b/src/arch/amdgpu/vega/insts/op_encodings.hh
@@ -1477,22 +1477,36 @@ namespace VegaISA
                 }
 
                 Addr flat_scratch_addr = readFlatScratch(gpuDynInst);
+                VecElemI32 swizzleOffset = soffset.rawData() + offset;
 
-                int elemSize;
-                auto staticInst = gpuDynInst->staticInstruction();
-                if (gpuDynInst->isLoad()) {
-                    elemSize = staticInst->getOperandSize(2);
-                } else {
-                    assert(gpuDynInst->isStore());
-                    elemSize = staticInst->getOperandSize(1);
+                // These are the same as RDNA3. From RDNA3 ISA manual:
+                // In Scratch SS mode (saddr != NULL (0x7f) SVE==0), the
+                // inst_offset must be aligned to the payload size: 4 byte
+                // aligned for 1-DWORD, 16-byte aligned for 4-DWORD.
+                //
+                // Also (SADDR + INST_OFFSET) must be at least DWORD-aligned.
+                if (!instData.SVE) {
+                    [[maybe_unused]] int elemSize;
+                    [[maybe_unused]] auto staticInst =
+                        gpuDynInst->staticInstruction();
+                    if (gpuDynInst->isLoad()) {
+                        elemSize = staticInst->getOperandSize(2);
+                    } else {
+                        assert(gpuDynInst->isStore());
+                        elemSize = staticInst->getOperandSize(1);
+                    }
+
+                    // Check offset aligned to payload and saddr+offset is
+                    // dword aligned.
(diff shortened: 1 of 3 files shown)


src/arch/amdgpu/vega/insts/op_encodings.hh
```diff
@@ -1477,22 +1477,36 @@ namespace VegaISA
...
+                VecElemI32 swizzleOffset = soffset.rawData() + offset;
+
+                // These are the same as RDNA3. From RDNA3 ISA manual:
+                // In Scratch SS mode (saddr != NULL (0x7f) SVE==0), the
+                // inst_offset must be aligned to the payload size: 4 byte
+                // aligned for 1-DWORD, 16-byte aligned for 4-DWORD.
+                //
+                // Also (SADDR + INST_OFFSET) must be at least DWORD-aligned.
+                if (!instData.SVE) {
+                    [[maybe_unused]] int elemSize;
```
mattsinc, line 1489: I'm not sure elemSize can be unused since either the if or else would be taken, but this is minor.

src/gpu-compute/wavefront.cc
```diff
@@ -384,8 +384,12 @@ Wavefront::initRegState(HSAQueueEntry *task, int wgSizeInWorkItems)
                 // the FLAT_SCRATCH register pair to the scratch backing
                 // memory: https://llvm.org/docs/AMDGPUUsage.html#flat-scratch
                 if (task->gfxVersion() == GfxVersion::gfx942) {
+                    uint32_t scratchPerWI =
+                        task->amdQueue.scratch_workitem_byte_size;
+
                     archFlatScratchAddr =
-                        task->amdQueue.scratch_backing_memory_location;
+                        task->amdQueue.scratch_backing_memory_location
```
mattsinc, line 391: should we explain what 64 is?  If wave size ever changes, then this probably wouldn't work?

mattsinc (approved):
Thanks @abmerop , nice job!

src/arch/amdgpu/vega/insts/op_encodings.hh
abmerop: Since i'm using assert here, those would be removed in gem5.fast and these would be unused

src/gpu-compute/wavefront.cc
abmerop: Yea, it could be NumVecElemPerVecReg... but I was using the style of the code below

mattsinc: we could change all of them in a separate commit?  That seems like a good idea ...

src/arch/amdgpu/vega/insts/op_encodings.hh
mattsinc: Ahh, I assumed maybe unused cared about if the variable was never read or written when asserts weren't used.  Sounds like it is only if it's never read -- makes sense then.

src/gpu-compute/wavefront.cc
abmerop: Sure.  I have something already queued up for other parts, so I will add this

PR #1948, opened 2025-01-25
util-docker: Update README.md and fix bugs in devcontainer Dockerfile

BobbyRBruce:


powerjg (dismissed):
I'm good with merging this.

However, I think the most common use case is to run this on x86 to build the x86 image (followed, maybe, by running this on Arm to build Arm). Can we add some lines about how to build locally *without* the cross platform build?

Said another way, if I wanted to build the docker images for personal use or testing and I didn't want to wait for all versions of gem5 to build, how would I do that?

powerjg:
One other thing... we should make this a github action workflow and update these images automatically every time the dockerfiles are modified.

BobbyRBruce:
> One other thing... we should make this a github action workflow and update these images automatically every time the dockerfiles are modified.

You can see in ".github/workflows/docker-build.yaml" that I made a start on this but it's still not working right (I can't remember the exact issue but anyone wanting to try this can run it and see for themselves). 

I'm pretty sure almost over the line, i just never found time to fix the bugs.

PR #1952, opened 2025-01-28
arch-x86: fix pack micro-op implementation

### LINKED ISSUE #1951: arch-x86: `PACKUSWB` produces incorrect results
**Describe the bug**
gem5' implementation of the `PACKUSWB` instruction does not produce the correct output (when compared to what real x86 hardware produces).

Here is a proof-of-concept program to reproduce the bug:
```c
#include <stdio.h>
#include <stdint.h>

static uint64_t test(uint64_t a, uint64_t b) {
  uint64_t c;
  asm volatile ("movq %1, %%mm0\n"
                "movq %2, %%mm1\n"
                "packuswb %%mm1, %%mm0\n"
                "movq %%mm0, %0\n"
                : "=r"(c)
                : "r"(a), "r"(b));
  return c;
}

static void do_test(uint64_t a, uint64_t b) {
  const uint64_t c = test(a, b);
  printf("%016lx\n"
         "%016lx\n"
         "%016lx\n",
         a, b, c);
}

int main() {
  do_test(0x80, 0);
  do_test(0xF000, 0);
  do_test(0xFFFF, 0);
}
```

On the host, it prints:
```
0000000000000080
0000000000000000
0000000000000080
000000000000f000
0000000000000000
0000000000000000
000000000000ffff
0000000000000000
0000000000000000
```

In gem5, it prints:
```
0000000000000080
0000000000000000
00000000000000ff
000000000000f000
0000000000000000
0000000000000000
000000000000ffff
0000000000000000
00000000000000ff
```

I.e., the output of the first and last tests are incorrect.

**Affects version**
develop @ [86e08e8](https://github.com/gem5/gem5/commit/86e08e87d3917602db2fef0b8052df61237cc2ab)

**gem5 Modifications**
None

**To Reproduce**
1. Compile gem: `scons build/X86/gem5.opt`
2. Compile the test program: `cc -o test test.c`
3. Run the test program on the host: `./test > host.txt`
4. Run the test program in gem5: `./build/X86/gem5.opt configs/deprecated/example/se.py --cmd ./test --output=$PWD/sim.txt`
5. Observe the differences in output: `diff host.txt sim.txt`

**Terminal Output**
```
$ diff host.txt sim.txt
3c3
< 0000000000000080
---
> 00000000000000ff
9c9
< 0000000000000000
---
> 00000000000000ff
```

**Expected behavior**
The program should print the same output in gem5 as on the host.

**Host Operating System**
Ubuntu 22.04

**Host ISA**
X86

**Compiler used**
GCC 11.4.0

**Additional information**

nmosier:
Fix #1951. This patch fixes two bugs in the pack x86 micro-op implementation. The pack micro-op packs signed 2n-bit inputs into signed or unsigned n-bit outputs with saturation, i.e., if the 2n-bit value is out of the range that can be represented by the signed or unsigned n-bit value, it is replaced with -128/127 or 0/255.

Bug 1: a signed 2n-bit input that fits into one signed byte is not clamped to 0 before being copied to an unsigned n-bit output. This is because the checks when the input is negative were performed out of order.

Bug 2: a signed 2n-bit input in the range 128 to 255 (i.e., it fits in one unsigned byte and the 7th bit is set) is "clamped" to 255 despite fitting in one unsigned byte.

This patch fixes the logic errors resulting in both bugs.

Change-Id: I2c1c59e84463c8b4def808f34bd9be4f24989b5f

### DIFF
--- a/src/arch/x86/isa/microops/mediaop.isa
+++ b/src/arch/x86/isa/microops/mediaop.isa
@@ -389,17 +389,20 @@ let {{
                     bits(FpSrcReg1_uqw, (i + 1) * srcBits - 1,
                                         (i + 0) * srcBits);
                 unsigned signBit = bits(picked, srcBits - 1);
-                uint64_t overflow = bits(picked, srcBits - 1, destBits - 1);
 
                 // Handle saturation.
                 if (signBit) {
-                    if (overflow != mask(srcBits - destBits + 1)) {
-                        if (signedOp())
+                    if (signedOp()) {
+                        uint64_t overflow = bits(picked, srcBits - 1,
+                                                 destBits - 1);
+                        if (overflow != mask(srcBits - destBits + 1)) {
                             picked = (1ULL << (destBits - 1));
-                        else
-                            picked = 0;
+                        }
+                    } else {
+                        picked = 0;
                     }
                 } else {
+                    uint64_t overflow = bits(picked, srcBits - 1, destBits);
                     if (overflow != 0) {
                         if (signedOp())
                             picked = mask(destBits - 1);


powerjg (approved):
Thanks for the fix!

PR #1957, opened 2025-01-30
arch-x86: Remove decodePages decode cache

powerjg:
The `decodePage/addrCacheMap` datastructure was meant as a *software* cache to improve decode performance. The main problem is that the `addrCacheMap` structure would grow without bounds as different applications were run in the guest system.

To solve this problem, this change removes the cache. The result is a more than 10x memory capacity savings for the simulator when booting Ubuntu 24.04 and no performance change.

More details because the original code was difficult to understand:
- A new cache object (mapping physical addresses to data) was created *every time* the m5reg was updated with a unique value. This meant that there was a *new* cache created frequently.
- A new entry was created in the cache for every new address accessed. In other words, the "decodePages" cache (of which there was a copy for each unique value of the m5reg) would cache *all physical memory* which was accessed during fetch.

When running Ubuntu 24.04 boot with massif, the `addrCacheMap` was over 95% of all memory allocated by gem5. This was *not* a memory *leak* because the data was always referenced in the map, but it was not a good use of this memory. The memory footprint of this workload was over 18 GiB. (As a side note, this is why our daily tests were failing. They were running out of memory.)

After this change, the time to complete the boot is basically unchanged (maybe slightly faster), and the memory footprint is now under 2 GiB.

### DIFF
--- a/src/arch/x86/decoder.cc
+++ b/src/arch/x86/decoder.cc
@@ -46,7 +46,7 @@ Decoder::doResetState()
 {
     origPC = basePC + offset;
     DPRINTF(Decoder, "Setting origPC to %#x\n", origPC);
-    instBytes = &decodePages->lookup(origPC);
+    instBytes.reset();
     chunkIdx = 0;
 
     emi.rex = 0;
@@ -64,12 +64,7 @@ Decoder::doResetState()
     emi.modRM = 0;
     emi.sib = 0;
 
-    if (instBytes->si) {
-        return FromCacheState;
-    } else {
-        instBytes->chunks.clear();
-        return PrefixState;
-    }
+    return PrefixState;
 }
 
 void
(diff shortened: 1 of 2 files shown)


powerjg:
I am running some more tests now (the boot tests take a long time on my machine), but I wanted to get this out there for feedback ASAP.

powerjg:
Running x86-ubuntu-boot.py with atomic CPU I get the following results:

Original code:

```
10849.69user 9.28system 3:07:09elapsed 96%CPU (0avgtext+0avgdata 17745408maxresident)k
```

This patch

```
10612.90user 3.54system 2:58:02elapsed 99%CPU (0avgtext+0avgdata 1179056maxresident)k
```

| | time | memory footprint |
|---|---|---|
|Original | 3 hr 7 minutes | 16.9 GiB |
|This patch | 2 hour 58 minutes  | 1.1 GiB |

powerjg:
This should fix the failing daily test as it's running out of memory. See https://github.com/gem5/gem5/actions/runs/13038986331/job/36384035674 for the latest failure.

Harshil2107 (approved):
lgtm

powerjg:
@abmerop Could you take a brief look at this code? I don't think anyone is an *expert* here, and I would feel more comfortable with merging something that's so core if it had another set of eyes on it.

abmerop (approved):
This seems fine. I'd probably also add this is not very realistic either. To be fair the original commit is 13 years old. This looks like it is intended to function similar to what modern x86 processors would call a uop cache, but that is not part of the decoder, so that would be better modeled elsewhere.

powerjg:
Thanks for looking into it! I don't think this was modeling *any* timing behavior. My reading of the code is that it was purely a software (simulator) "optimization".

PR #1961, opened 2025-01-31
stdlib: Add SE mode support to multi-program workloads

### LINKED ISSUE #1960: stdlib: add SE mode support to multi-program workloads
#### Describe the enhancement
The current stdlib does not support multi-program workloads, which were previously supported by the now-deprecated se.py. In memory architecture research, inducing memory congestion is crucial, necessitating support for multi-program workloads

#### Implementation
New function `set_se_multi_binary_workload()` is added. Here is a full script If stdlib multi-prgoram is used.
```python
# Importing the required python packages here
import os
# We need to first determine which ISA that we want to use. Then we have to
# make sure that we are using the correct ISA while executing this script.
from gem5.isas import ISA
from gem5.resources.resource import BinaryResource 
# We import various parameters of the machine.
from gem5.components.cachehierarchies.classic\
    .private_l1_private_l2_cache_hierarchy import (
)
from gem5.components.processors.simple_switchable_processor import SimpleSwitchableProcessor
# We will use the new simulator module to simulate this task.
from gem5.simulate.exit_event import ExitEvent
from gem5.resources.workload import Workload
from m5.util import fatal

    l1d_size="16kB",
    l1i_size="16kB",
    l2_size="256kB",
)

memory = SingleChannelDDR3_1600(size="32MiB")

processor = SimpleProcessor(
    cpu_type=CPUTypes.TIMING,
    isa=ISA.X86,
    num_cores=4
)

board = SimpleBoard(
    clk_freq = "3GHz",
    processor = processor,
    memory = memory,
    cache_hierarchy = cache_hierarchy,
)

# Set SE mode binary workload
board.set_se_multi_binary_workload(
    binaries=[Workload("x86-print-this"),
            Workload("x86-print-this"),
            Workload("x86-matrix-multiply"),
            Workload("x86-matrix-multiply"),
            ],
    arguments=[[f"Hello World!! 1", 1],
               [f"Hello World!! 2", 1],
               [],
               [],
               ],
)

# Lastly we instantiate the simulator module and simulate the program.
simulator.run()

print(
    "Exiting @ tick {} because {}.".format(
        simulator.get_current_tick(),
        simulator.get_last_exit_event_cause(),
    )
)

```

There is a few consideration too:
  - The number of programs to be run shoubled match the number of cores (`se.py` was same before)
  - Mixing multi-thread program can produce unexpected result.

sangjae4309:
This PR implements adding multi-prgoram support of stdlib SE mode(#1960).

A new function `set_se_multi_binary_workload()` is added, It distributes each binary and its corresponding arguments to cores.


powerjg (commented):
See https://github.com/gem5/gem5/issues/1960#issuecomment-2627744564 for some feedback.

sangjae4309:
@powerjg 
Thank you for your feedback. I’ve taken your suggestion and implemented the separation of the multi-program function.

However, I’m a bit confused. Could you please confirm if you were referring to `Workload` in `src/python/gem5/resources/workload.py`? I have tested it, and it seems to be running correctly. Here is the script I used for verification: [simple-se-workload-resource.zip](https://github.com/user-attachments/files/18628126/simple-se-workload-resource.zip). And also uploaded its raw code at #1960.

I appreciate your feedback. Looking forward to your response!

powerjg:
That looks good. Yes, we need to make sure that workloads (e.g., https://resources.gem5.org/resources/[x86-npb-bt-size-s-run](https://resources.gem5.org/resources/x86-npb-bt-size-s-run/raw?database=gem5-resources&version=1.0.0)/raw?database=gem5-resources&version=1.0.0) will work correctly with `set_se_multi_binary_workload`.

This can be done in another PR. @Harshil2107 making sure `set_se_multi_binary_workload` works as a workload is required for gem5 v25.0. I also think it would be a good idea to factor out the common code between that function and `set_se_binary_workload`.

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -61,13 +61,116 @@ class SEBinaryWorkload:
...
+        # If we are setting a workload of this type, we need to run as a
+        # SE-mode simulation.
+        self._set_fullsystem(False)
+
+        multiprocesses = []
+        for i, binary in enumerate(binaries):
+            process = Process(pid=100 + i)
+            binary_path = binary.get_local_path()
+            process.executable = binary_path
+            process.cmd = [binary_path] + arguments[i]
```
Harshil2107: This line throws an index out of bounds error when running binaries like:

```python
board.set_se_multi_binary_workload(
    binaries=[
        obtain_resource("x86-hello64-static"),
        obtain_resource("x86-npb-is-size-s"),
    ],
)
```

The issue arises because we expect the `arguments` list to have the same size as the `binaries` list. I suggest adding a check similar to the one ensuring the number of binaries matches the number of cores.

To improve flexibility, I would suggest making `arguments` optional:  
- If `arguments` is an empty list, default to `process.cmd = [binary_path]`.  
- If `arguments` is non-empty, enforce that its size matches the number of binaries.  

This would allow users to omit `arguments` when not needed.

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -61,13 +61,116 @@ class SEBinaryWorkload:
...
+        :param arguments: The list of input arguments for the binary
+        :param checkpoint: The checkpoint directory. Used to restore the
+                           simulation to that checkpoint.
+        """
+
+        # We assume this this is in a multiple-inheritance setup with an
+        # Abstract board. This function will not work otherwise.
+        assert isinstance(self, AbstractBoard)
+        # The number of workload should match the nubmer of cores
+        assert len(binaries) == self.get_processor().get_num_cores()
```
Harshil2107: It would be helpful for the user to understand why the assertion is failing. Right now, the program only prints `AssertionError: <EMPTY MESSAGE>`. Adding a descriptive error message would make it clearer what went wrong and how to fix it.

Harshil2107 (changes requested):

sangjae4309:
@powerjg
Ah, I got it!!. I will run another workloads and explore improvement in a future PR.

@Harshil2107 
Thank you for your feedback. I have incorporated your suggestions into my code and also refactored the common logic with `set_se_binary_workload`.

Looking forward to your response!

sangjae4309 (commented):
.

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -61,13 +61,145 @@ class SEBinaryWorkload:
...
+            warn("Workload has been set more than once!")
+        self.set_is_workload_set(True)
+
+        # If we are setting a workload of this type, we need to run as a
+        # SE-mode simulation.
+        self._set_fullsystem(False)
+
+        multiprocesses = []
+        for i, binary in enumerate(binaries):
+            process = self._create_process(
```
Harshil2107, line 188: I think `arguments` should also be passed here as a parameter, as otherwise we would not be able to run the binaries that require arguments.

I would also recommend something along the lines of 

```python
binary_arguments=[]
if arguments:
    binary_arguments=arguments[i]
process = self._create_process(
                binary=binary,
                pid=100 + i,
                stdin_file=stdin_file,
                stdout_file=stdout_file,
                stderr_file=stderr_file,
                env_list=env_list,
                arguments=binary_arguments
            )
```

This way if the user is running binaries that do not require arguments, they can just call

```
board.set_se_multi_binary_workload(binaries=[b1,b2])
```

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -61,13 +61,145 @@ class SEBinaryWorkload:
...
+
+        # We assume this this is in a multiple-inheritance setup with an
+        # Abstract board. This function will not work otherwise.
+        assert isinstance(self, AbstractBoard)
+
+        num_cores = self.get_processor().get_num_cores()
+        assert len(binaries) == num_cores == len(arguments), (
+            f"Number of binaries({len(binaries)}), "
+            f"arguments({len(arguments)}) and cores({num_cores}) does not match."
+        )
```
Harshil2107:
```suggestion
        assert len(binaries) == num_cores, (
            f"Number of binaries({len(binaries)}) "
            f"and cores({num_cores}) should be the same."
        )
        if arguments:
            assert len(binaries) == len(arguments), (
                f"Mismatch between binaries and arguments: "
                f"{len(binaries)} binaries but {len(arguments)} argument lists provided. "
                "The 'arguments' list is optional if none of the binaries require arguments. "
                "However, if any binary requires arguments, a list must be provided for all binaries. "
                "For binaries that do not require arguments, an empty list should be used."
)
        
```

I would suggest updating the assert to something similar to what I mentioned above. As the `arguments` list might be empty if none of the binaries need any arguments. 

This is based on my understanding of the functionality we want. Below is an example of how I think the use case is:

Lets say we have 3 binaries: `b1`, `b2` and `b3`. Only binary `b3` needs an argument `arg1`.

we can call `board.set_se_multi_binary_workload(binaries=[b1,b2])` and `board.set_se_multi_binary_workload(binaries=[b1,b2,b3], arguments=[[],[],[arg1]])` but we cannot call `board.set_se_multi_binary_workload(binaries=[b1,b2,b3], arguements=[[arg1]])`.

Please let me know if I am misunderstanding something.

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -109,29 +241,17 @@ def set_se_binary_workload(
...
-        if env_list is not None:
-            process.env = env_list
+        process = self._create_process(
+            binary=binary,
+            pid=100,
+            stdin_file=stdin_file,
+            stdout_file=stdout_file,
+            stderr_file=stderr_file,
+            env_list=env_list,
+        )
```
Harshil2107:
```suggestion
        process = self._create_process(
            binary=binary,
            pid=100,
            stdin_file=stdin_file,
            stdout_file=stdout_file,
            stderr_file=stderr_file,
            env_list=env_list,
            arguments=arguments,
        )
```

The parameter `arguments` should also be passed otherwise we wont we able to run binaries that require arguments.

Harshil2107 (changes requested):

Harshil2107:
Thank you for the changes, The `_create_process` function looks good to me. I have requested a few  more small changes, mainly related to the `arguments` parameter.

src/python/gem5/components/boards/se_binary_workload.py
sangjae4309: Oh, I overlooked the case where an argument is not required. Thanks for pointing it out!

src/python/gem5/components/boards/se_binary_workload.py
sangjae4309: Right, that is exactly what I intended. If there is at least one argument, the sequence count must be properly aligned.

sangjae4309: Thanks for adding the detailed descriptions. That is  very helpful to understand.

src/python/gem5/components/boards/se_binary_workload.py
sangjae4309: Oh, I missed the parameter. Thanks for pointing it out.

sangjae4309:
@Harshil2107 
Thanks for pointing several bugs and adding helpful suggestion. I have made the requested changes.

Harshil2107 (dismissed):
lgtm

src/python/gem5/components/boards/se_binary_workload.py
```diff
@@ -61,13 +61,157 @@ class SEBinaryWorkload:
...
+                    "AbstractResource."
+                )
+
+    def set_se_multi_binary_workload(
+        self,
+        binaries: List[BinaryResource],
+        exit_on_work_items: bool = True,
+        stdin_file: Optional[FileResource] = None,
+        stdout_file: Optional[Path] = None,
+        stderr_file: Optional[Path] = None,
```
powerjg: If you're running multiple binaries, I believe you will need to have all of these parameters also be a `List`. Especially for `stdin` you're going to want one per binary. Generally, I think you want a different file descriptor for all `stdin`,  `stdout`, and `stderr` for each binary.

sangjae4309: Thank you for pointing out!! I've implemented your suggestion.

sangjae4309:
Thanks for the review! I've identified some mistakes in the code and have implemented the necessary corrections. I’d appreciate any further review.

powerjg (commented):
Thanks for all of the updates! This looks good to me, but I'll let @Harshil2107 sign off.

Harshil2107 (approved):
lgtm

PR #1965, opened 2025-02-02
Update se.py

Mounishayarava:
updated se.py


powerjg (commented):
Could you explain the purpose of this change? It looks like these may be personal changes not meant for upstream, but if this is fixing a bug, we would be happy to accept the change with appropriate explanation.

You may also be interested in using non-deprecated interfaces with gem5. See the [gem5 bootcamp](https://bootcamp.gem5.org/) for lots of learning materials.

erin-le:
Hello, this PR is being closed as it modifies a deprecated script and appears to be inactive. Feel free to reopen this if this is fixing a bug and you would like to keep working on it, however!

PR #1968, opened 2025-02-03
scons: Improve codegen for build/../params/*.hh

bpfoley:
Headers such as cstddef, string, vector etc were being #include-d once per SimObject configuration param if the param type depended on it. This lead to headers such as params/PciDevice.hh having 65 copies of '#include <cstddef>'.

Most system headers have include guards to stop them being included more than once, and modern GCC and Clang have special case code to avoid the cost of running the entire header through the C preprocessor to do this, but still, we may as well not create unnecessary work for our toolchain/language server/code analysis tools if we can easily avoid it.

Change-Id: Ib3313f515177c71ff2a77f7c7d6967f11585623e

### DIFF
--- a/build_tools/code_formatter.py
+++ b/build_tools/code_formatter.py
@@ -329,7 +329,8 @@ class code_formatter(metaclass=code_formatter_meta):
                 )
 
         d = code_formatter.pattern.sub(convert, format)
-        self._append(d)
+        if not kwargs.get("add_once") or d not in self._data:
+            self._append(d)
 
 
 __all__ = ["code_formatter"]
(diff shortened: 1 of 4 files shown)


powerjg (commented):
Out of curiosity, is there ever a time that we *don't* want to use `add_once`? In other words, should we just make that behavior the default?

bpfoley:
> Out of curiosity, is there ever a time that we _don't_ want to use `add_once`? In other words, should we just make that behavior the default?

For `#include`s, it's probably a reasonable default. But the `code` function is just a generic 'emit generated line of code' function that has no particular knowledge of any programming language and is used for all sorts of purposes -- generating .dot files, C++, HTML, etc.  See, e.g. https://github.com/gem5/gem5/blob/stable/src/python/m5/params.py#L997-L1023

powerjg:
Got it. Thanks for the explanation!

powerjg (approved):

PR #1975, opened 2025-02-05
scons: Use relative paths in codegen to improve build caching

bpfoley:
The gem5 codegen for params, python wrappers, ruby classes etc embeds the absolute path of the codegen script as a comment in the generated code.

This is bad news if you happen to have multiple similar checkouts of gem5 in diferent directories and are using ccache as a frontend to your C++ compiler to provide an extra layer of caching across workspaces. (This is something that could easily occur in a CI system too). Even though they're only comments, the differing strings in the generated code will dramatically reduce the hit rate on this shared build cache.

Use workspace-relative paths for all the header files to stop everything that transitively #include-s them from missing the build cache.

This changes, eg
/**
 * DO NOT EDIT THIS FILE!
 * File automatically generated by
 *   /home/gem5-dev/gem5/build_tools/sim_object_param_struct_cc.py:295
 */

to

/**
 * DO NOT EDIT THIS FILE!
 * File automatically generated by
 *   build_tools/sim_object_param_struct_cc.py:295
 */

Unfortunately, the EmbeddedPython data in the .py.cc bindings require absolute path strings in them, so we can't remove those, but those paths affect a more limited set of object files so this still a big improvement for cross-workspace cacheability.

Change-Id: I516adb19a0792c10943b3e410a1835a01ec7589a


powerjg (approved):
Cool! Thanks for the improvement!

PR #1976, opened 2025-02-06
gpu-compute: GPU progress prints and debug tracing

abmerop:
This PR contains two commits of changes which have been extremely useful in the past to debug problems with the GPU execution pipeline. It includes a progress printer, off by default, and a new debug flag with minimal prints with only useful information for debugging. These have been used by us to find issues in GPU applications caused by bugs in gem5.

The GPU progress print can be used as a basic progress print but is also useful for visually identify live locks. It has been used in the past to identify a region of code with an instruction using/producing incorrect data and leading to fixes.  It prints each CUs wave status, why scheduling is blocked if applicable, and what memory instructions are inflight.

The GPU debug trace prints minimal information per-wave about what instruction is being executed and the register values read/written.  For memory instructions it also prints the virtual to physical translations and what data is placed in registers upon loads, both with sequence number to tie back to the execute stage of the instruction.


abmerop:
@mattsinc This is something that has been on my backlog for several months.  I am attempting to reduce the backlog :)

powerjg:
Personally, I love more debug options!

Could you give a sample of the output from these traces? It may also be good to find a place to document them a little. Specifically, these debug flags are unlikely to be very easy to *find* for new (or current) gem5 users/developers. Doing something to make these options findable would be good.

abmerop:
> Personally, I love more debug options!
> 
> Could you give a sample of the output from these traces? It may also be good to find a place to document them a little. Specifically, these debug flags are unlikely to be very easy to _find_ for new (or current) gem5 users/developers. Doing something to make these options findable would be good.

Sure, maybe a website page within https://www.gem5.org/documentation/general_docs/gpu_models/ ?

mattsinc (commented):
Are the untrack functions there to track addresses that may have deadlocked?  Rest looks good.

abmerop:
> Are the untrack functions there to track addresses that may have deadlocked? Rest looks good.

Correct. I've also used this in the past to find instructions missing corresponding waitcnt decrements (e.g., when exec==0)

abmerop:
Updated to not track anything when progress interval is 0 (perf/mem savings)

mattsinc (dismissed):

abmerop:
Fixed a typo causing a crash in global memory pipeline, ready for re-review (one character change m -> mp)

mattsinc (approved):

PR #1977, opened 2025-02-06
mem-cache: Unit test FIFO RP

odanrc:
Add unit tests for the FIFO replacement policy.

As side effects:
    - Fixed a bug where invalidated entries would not be prioritized during the victimization process

### DIFF
--- a/src/mem/cache/replacement_policies/fifo_rp.test.cc
+++ b/src/mem/cache/replacement_policies/fifo_rp.test.cc
@@ -1,0 +1,229 @@
+/**
+ * Copyright (c) 2025 Daniel R. Carvalho
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <gtest/gtest-spi.h>
+#include <gtest/gtest.h>
+
+#include <cassert>
+
+#include "mem/cache/replacement_policies/fifo_rp.hh"
+#include "params/FIFORP.hh"
+
+/// Common fixture that initializes the replacement policy
+class FIFORPTestF : public ::testing::Test
+{
+  public:
+    std::shared_ptr<gem5::replacement_policy::FIFO> rp;
+
+    FIFORPTestF()
+    {
+        gem5::FIFORPParams params;
+        params.eventq_index = 0;
+        rp = std::make_shared<gem5::replacement_policy::FIFO>(params);
+    }
+};
+
+/// Test that instantiating an entry generates a non-null replacement data
+TEST_F(FIFORPTestF, InstantiatedEntry)
+{
+    const auto repl_data = rp->instantiateEntry();
+
+    // instantiateEntry must return a valid pointer
+    ASSERT_NE(repl_data, nullptr);
+}
+
+/// Test that if there is one candidate, then it will always be the victim,
+/// regardless of its replacement data
+TEST_F(FIFORPTestF, GetVictim1Candidate)
+{
+    gem5::ReplaceableEntry entry;
+    entry.replacementData = rp->instantiateEntry();
+    gem5::ReplacementCandidates candidates;
+    candidates.push_back(&entry);
+    ASSERT_EQ(rp->getVictim(candidates), &entry);
+
+    rp->invalidate(entry.replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), &entry);
+
+    rp->reset(entry.replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), &entry);
+
+    rp->touch(entry.replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), &entry);
+}
+
+/// Fixture that tests victimization
+class FIFORPVictimizationTestF : public FIFORPTestF
+{
+  protected:
+    // The entries being victimized
+    std::vector<gem5::ReplaceableEntry> entries;
+
+    // The entries, in candidate form
+    gem5::ReplacementCandidates candidates;
+
+  public:
+    // The number of entries is arbitrary. It does not need to be high, since
+    // having more entries is not expected to increase coverage
+    FIFORPVictimizationTestF() : FIFORPTestF(), entries(4) {
+        for (auto &entry : entries) {
+            entry.replacementData = rp->instantiateEntry();
+            candidates.push_back(&entry);
+        }
+    }
+};
+
+/// Test that when all entries are invalid the first candidate will always be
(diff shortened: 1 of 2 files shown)


odanrc:
@BobbyRBruce locally I have added tests for MRU, FIFO, LFU, LRU and 2nd chance. So far only FIFO had a bug.

powerjg:
So cool that you found a bug! The failing test is just a runner being flakey. We can re-run once the other tests pass.

odanrc:
> So cool that you found a bug!

Yeah! Now that we can easily add tests to sim objects we should be able to spot more (or hopefully none :)).

> The failing test is just a runner being flakey. We can re-run once the other tests pass.

In this case it was because I reordered my commits and forgot to pick one of them. It should pass now.

src/mem/cache/replacement_policies/fifo_rp.test.cc
```diff
@@ -0,0 +1,233 @@
...
+    rp->touch(entry.replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), &entry);
+}
+
+/// Fixture that tests victimization
+class FIFORPVictimizationTestF : public FIFORPTestF
+{
+  protected:
+    // The entries being victimized
+    std::vector<gem5::ReplaceableEntry> entries;
```
giactra, line 85: Do we really need the entries vector? Can't we just use the candidates one?

src/mem/cache/replacement_policies/fifo_rp.test.cc
```diff
@@ -0,0 +1,233 @@
...
+            candidates.push_back(&entry);
+        }
+    }
+};
+
+/// Test that when all entries are invalid a single entry will always be
+/// selected, regardless of the order of the invalidations
+TEST_F(FIFORPVictimizationTestF, GetVictimAllInvalid)
+{
+    auto expected_victim = &entries[0];
```
giactra: Nit: "front()" seems like a more descriptive way of referencing the first entry

src/mem/cache/replacement_policies/fifo_rp.test.cc
```diff
@@ -0,0 +1,233 @@
...
+    // having more entries is not expected to increase coverage
+    FIFORPVictimizationTestF() : FIFORPTestF(), entries(4) {
+        for (auto &entry : entries) {
+            entry.replacementData = rp->instantiateEntry();
+            candidates.push_back(&entry);
+        }
+    }
+};
+
+/// Test that when all entries are invalid a single entry will always be
```
giactra: What do you mean by a single entry?
Do you mean the first one to be instantiated? It's not clear from the test description which candidate will be selected in that scenario

src/mem/cache/replacement_policies/fifo_rp.test.cc
```diff
@@ -0,0 +1,233 @@
...
+    // Since all candidates are already invalid, nothing changes if we
+    // invalidate all of them again
+    for (auto &entry : entries) {
+        rp->invalidate(entry.replacementData);
+    }
+    ASSERT_EQ(rp->getVictim(candidates), expected_victim);
+
+    // Even if we invalidate the entry being selected for victimization last
+    rp->invalidate(expected_victim->replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), expected_victim);
```
giactra: Considering the scope of the test is to check:

```
/// Test that when all entries are invalid a single entry will always be
/// selected, regardless of the order of the invalidations
```

I believe a simpler fixture generation would construct the test with all entries in a VALID state. So you can have a simpler test case:

    // Reverse invalidation order
    for (auto it = candidates.rbegin(); it != candidates.rend(); ++it) {
        rp->invalidate(it->replacementData);
    }
    ASSERT_EQ(rp->getVictim(candidates), expected_victim);

giactra (commented):
That is amazing @odanrc !
Thanks for this valuable contribution! Looking forward to see some tests for LRU and RRIP! I have some comments wrt the unit tests

src/mem/cache/replacement_policies/fifo_rp.test.cc
odanrc: Candidates is a vector of pointers, so we need to instantiate them somewhere, which is what entries is being used for.

src/mem/cache/replacement_policies/fifo_rp.test.cc
odanrc: I will try to clarify the comment, but I mean that it always picks the first entry in the candidate list.

src/mem/cache/replacement_policies/fifo_rp.test.cc
odanrc: For context, although I think you understood that, the name of this test is GetVictimAllInvalid, so its goal is to determine which entry is picked when all entries are invalid. The following test, GetVictimOneInvalid, determines the behavior when all entries but one are valid. Finally, GetVictim determines the behavior when all entries are valid.

Regarding your suggestion, I actually added lines 143 and 142 on purpose to reduce the likelihood of some kind of bias happening inside the invalidation function. For example, under your proposal, what guarantees that the victim is front() because it was the last to be invalidated, and not because it is the first candidate? I wanted to 'make sure' that no matter the order at which an invalidation happens, the victim is deterministic (front()), if all candidates are invalid.

odanrc:
I only changed 2 lines in the last upload, so I squashed into the first commit. Here are the changes:

```
-/// Test that when all entries are invalid a single entry will always be
+/// Test that when all entries are invalid the first candidate will always be
 /// selected, regardless of the order of the invalidations
 TEST_F(FIFORPVictimizationTestF, GetVictimAllInvalid)
 {
-    auto expected_victim = &entries[0];
+    auto expected_victim = &entries.front();
```

odanrc:
@giactra , have you got some time to review this again?

src/mem/cache/replacement_policies/fifo_rp.test.cc
giactra: > For example, under your proposal, what guarantees that the victim is front() because it was the last to be invalidated, and not because it is the first candidate?

I am not sure I understand. I could ask the same about the current version?

There's really no reason why invalidating outside of a loop is better than within a loop. Just doing

```
// Reverse invalidation order
for (auto it = candidates.rbegin(); it != candidates.rend(); ++it) {
    rp->invalidate(it->replacementData);
}
```

Will guarantee the expected victim will be invalidated last, Similarly to what you do by adding an out of the loop invalidation call.
I believe we should try to make unit-tests as simple as possible and avoid calling too many methods. If you fear there is a risk of an entry being selected because it is the first of the candidates, we should add a separate test that solely checks for that

src/mem/cache/replacement_policies/fifo_rp.cc
```diff
@@ -48,20 +48,27 @@ FIFO::FIFO(const Params &p)
...
 FIFO::invalidate(const std::shared_ptr<ReplacementData>& replacement_data)
 {
-    // Reset insertion tick
+    assert(replacement_data);
+
+    // To prioritize these entries to be selected during victimization
+    // we set their insertion tick as 0
     std::static_pointer_cast<FIFOReplData>(
-        replacement_data)->tickInserted = ++timeTicks;
+        replacement_data)->tickInserted = 0;
```
giactra: This is the bug fix. Shall we push this as a separate commit?

giactra (changes requested):
Hi @odanrc apologies for the late review. 

Wrt to the first commit which is making the ReplacementData polymorphic: 

> This struct is supposed to be used polymorphically, so make
> it possible to do so.

IMHO this is not true. ReplacementData is simply an empty struct whose sole purpose is to serve as an interface for polymophic ReplacementPolicy. There is really no virtual method defined in ReplacementData so I fail to understand how it can be considered polymorphic.

I understand adding a vtable is useful for dynamic type checking in your unit-tests, but I don't consider this to be a sufficient condition for adding a virtual method to the class. Let me know if you disagree. If there is consensus on paying this penalty, I believe it should be at least explicit.
I fear though that transforming ReplacementData into a poly type might make the design of replacement framework much more complicated (but you are the expert here so please correct me if I am wrong)

src/mem/cache/replacement_policies/fifo_rp.cc
odanrc: Done: https://github.com/gem5/gem5/pull/2412

odanrc:
> Hi @odanrc apologies for the late review.
> 
> Wrt to the first commit which is making the ReplacementData polymorphic:
> 
> > This struct is supposed to be used polymorphically, so make
> > it possible to do so.
> 
> IMHO this is not true. ReplacementData is simply an empty struct whose sole purpose is to serve as an interface for polymophic ReplacementPolicy. There is really no virtual method defined in ReplacementData so I fail to understand how it can be considered polymorphic.
> 
> I understand adding a vtable is useful for dynamic type checking in your unit-tests, but I don't consider this to be a sufficient condition for adding a virtual method to the class. Let me know if you disagree. If there is consensus on paying this penalty, I believe it should be at least explicit. I fear though that transforming ReplacementData into a poly type might make the design of replacement framework much more complicated (but you are the expert here so please correct me if I am wrong)

Turns out I made the destructor virtual because of a compilation error (using dynamic cast) from the test. I fixed that by reducing the test coverage, as we shouldn´t be modifying the source to accommodate for a test.

src/mem/cache/replacement_policies/fifo_rp.test.cc
powerjg: I don't think it hurts to "double check" this. At this point, I don't think we want to stall this PR if we don't have to.

powerjg (commented):
@odanrc is this ready for review?

src/mem/cache/replacement_policies/fifo_rp.test.cc
odanrc: I seem to have not seen Giacomo's latest comment. I have just updated the code as he suggested (last commit).

powerjg (dismissed):

powerjg:
@odanrc It looks like the test is failing. Other than that, the code looks good to me.

odanrc:
> @odanrc It looks like the test is failing. Other than that, the code looks good to me.

Solved it. It was a case of forgetting to dereference an iterator.

odanrc:
@powerjg @giactra @erin-le @BobbyRBruce  , could someone merge this PR? I have been getting 1-2 automated e-mails about it *every single day* since its inception. I really don't know why - they claim that this PR fails the CI, but... it doesn't?

BobbyRBruce (approved):

PR #1978, opened 2025-02-07
arch: Add BaseISA::getIntegerLength interface

rogerchang23424:
Get integer register length based on the current mode of context. The Implementation is based on the checking arch bits methods of RemoteGDB::gdbRegs

Implemented the methods for each ISA:
ARM: Chech inAArch64
MIPS: Only 32 bits mode implemented
POWER: Check Msr.sf bit
RISC-V: Check ISA::_rvType
SPARC: Check PSTATE.am bit
X86: Check m5reg.submode

### DIFF
--- a/src/arch/arm/fastmodel/iris/isa.hh
+++ b/src/arch/arm/fastmodel/iris/isa.hh
@@ -82,6 +82,11 @@ class ISA : public BaseISA
     {
         panic("setMiscReg not implemented.");
     }
+
+    int64_t getIntegerLength() const override
+    {
+        panic("getIntegerLength not implemented.");
+    }
 };
 
 } // namespace Iris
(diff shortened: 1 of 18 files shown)


giactra:
Hi @rogerchang23424, I think I need more context to understand how this will be used.
Usually when a virtual method gets added to the GenericISA it means it will be used by the ISA agnostic code (like the cpu-03,cpu-minor etc). This PR seems to lack such thing and instead it employs it in the ISA code instead

Is there a plan on adding extra PRs iin the future to motivate this addition?

I also have an extra comment: getIntegerLength name can be a bit misleading (for Arm) **depending on the context**
For example the way it is defined works if it refers to the maximum length of an integer register. However, referenced in the context of a single instruction, could lead to incorrect behaviour as in AArch64 mode you can execute instructions with 32bit registers as operands.

This is why I am annoyingly asking to give some extra context as I cannot review the API without understanding which will be its application. Thanks!

rogerchang23424:
Thanks @giactra comments. I just want to add a generic methods to get the arch integer register bits. I will close the PR.

PR #1980, opened 2025-02-07
arch-arm: arm misc reg for gic pfr0_el1.gic

### LINKED ISSUE #1941: [bug] [arm]： pfr0_el1.gic  will never be true because of the value is used before it can be initiated
the  value is initialized in [initializeMiscRegMetadata()](https://github.com/gem5/gem5/blob/c9625ce9cc5b5a90a38327de5ac0e1870974af5e/src/arch/arm/isa.cc#L131)  in class ISA  constructor
[pfr0_el1.gic = FullSystem && getGICv3CPUInterface(tc) ? 0x1 : 0;](https://github.com/gem5/gem5/blob/c9625ce9cc5b5a90a38327de5ac0e1870974af5e/src/arch/arm/regs/misc.cc#L4906)

the question is when ISA is created and constructor is called  the [ISA::getGICv3CPUInterface(ThreadContext *tc)](https://github.com/gem5/gem5/blob/c9625ce9cc5b5a90a38327de5ac0e1870974af5e/src/arch/arm/isa.cc#L1385) **will always return nullptr**, because at this time the gic has NOT been created yet !! 
the ArmSystem.gic is set in [BaseGic::init()](https://github.com/gem5/gem5/blob/c9625ce9cc5b5a90a38327de5ac0e1870974af5e/src/dev/arm/base_gic.cc#L71C1-L71C16) which is behind the SimObject creation

```
BaseGic::init()
{
    PioDevice::init();
    getSystem()->setGIC(this);
}
```
one solution for this might be put the [initializeMiscRegMetadata()](https://github.com/gem5/gem5/blob/c9625ce9cc5b5a90a38327de5ac0e1870974af5e/src/arch/arm/isa.cc#L131) in  ISA::initState which will be called after all the simobject is created

### Code snippet from [simulate.py :: instantiate](https://github.com/gem5/gem5/blob/c9625ce9cc5b5a90a38327de5ac0e1870974af5e/src/python/m5/simulate.py#L80)

```
    # Create the C++ sim objects and connect ports
    for obj in root.descendants():
        obj.createCCObject()                        //  **simobject created here**
    for obj in root.descendants():
        obj.connectPorts()

    # Do a second pass to finish initializing the sim objects
    for obj in root.descendants():
        obj.init()                                        //  **init()  is called here**
....
    # Restore checkpoint (if any)
    if ckpt_dir:
        _drain_manager.preCheckpointRestore()
        ckpt = _m5.core.getCheckpoint(ckpt_dir)
        for obj in root.descendants():
            obj.loadState(ckpt)
    else:
        for obj in root.descendants():
            obj.initState()                //   **call preUnflattenMiscReg()  to init arm register here after  all simobjects created !**
```

gxflying:
the value of pfr0_el1.gic is set before the gic pointer is initialized, so the vaule will always be false.
the solution is to set pfr0_el1.gic in ISA::initState() which is called after the  gic pointer's initialization (in BaseGic::init())

more details please refer to : https://github.com/gem5/gem5/issues/1941

### DIFF
--- a/src/arch/arm/isa.cc
+++ b/src/arch/arm/isa.cc
@@ -128,10 +128,15 @@ ISA::ISA(const Params &p) : BaseISA(p, "arm"), system(NULL),
     }
 
     selfDebug = new SelfDebug();
-    initializeMiscRegMetadata();
-    preUnflattenMiscReg();
 
     clear();
+}
+
+void
+ISA::initState()
+{
+    initializeMiscRegMetadata();
+    preUnflattenMiscReg();
 }
 
 void
--- a/src/arch/arm/isa.hh
+++ b/src/arch/arm/isa.hh
@@ -201,6 +201,8 @@ namespace ArmISA
         RegVal readMiscRegReset(RegIndex) const;
         void setMiscRegReset(RegIndex, RegVal val);
 
+        void initState() override;
+
         int
         flattenMiscIndex(int reg) const
         {


erin-le:
@gxflying Edit: I replied here before seeing your comment on the issue. For tagging your PR, all you have to do is change the PR title and commit message title to `arch-arm : arm misc reg for gic pfr0_el1.gic`. Someone else will take care of assigning labels.

It is also highly recommended to install pre-commit, which will tell you about issues with commit message formatting and will also check the formatting of Python files. Instructions on installing pre-commit can be found [here](https://bootcamp.gem5.org/#06-Contributing/01-contributing), on pages 29-30. I also pointed you to the wrong list of tags in my comment on the issue; the correct list is https://github.com/gem5/gem5/blob/stable/MAINTAINERS.yaml.

src/arch/arm/isa.cc
```diff
@@ -128,12 +128,17 @@ ISA::ISA(const Params &p) : BaseISA(p, "arm"), system(NULL),
     }
 
     selfDebug = new SelfDebug();
-    initializeMiscRegMetadata();
-    preUnflattenMiscReg();
 
     clear();
```
giactra, line 132: If you leave the clear call here, it will be called before initializeMiscRegMetadata and you would wrongly initialize the ISA registers.

I am talking specifically about the code below:


    for (auto idx = 0; idx < NUM_MISCREGS; idx++) {
        miscRegs[idx] = lookUpMiscReg[idx].reset();
    }

src/arch/arm/isa.cc
```diff
@@ -128,12 +128,17 @@ ISA::ISA(const Params &p) : BaseISA(p, "arm"), system(NULL),
...
 }
 
+void
+ISA::initState()
+{
+    initializeMiscRegMetadata();
+    preUnflattenMiscReg();
+}
```
giactra, line 140: Deferring to the initState is a potential solution. What concerns me though is other classes looking into the ISA at construction/init time and reading invalid data... I will run some tests and come back to you on this.

Ideally (as the ISA is one of the most important classes) we should fix more on the GIC side..

giactra (commented):
Many thanks for your contribution! Please check my comments below. Also, you should also amend the commit title to start with "arch-arm" (and no "bug-fix")

src/arch/arm/isa.cc
gxflying: hi, @giactra thanks for your review,  
you are right that this issue  probably is better to be fixed on gic side,
and the solution might be set gic pointer in the constructor of basic_gic, (as the generic_timer has done which also need a pointer to initialize the micsReg)

but one of my concerns is that  **Are there any mechanisms** to ensure the order in which the constructors of these simobjects are called (eg. the ISA constructor should be called before GIC/GenericTimer to fix this issue)?  

the simobjects are created in a recursive manner in simulate.py by  [obj.createCCObject()](https://github.com/gem5/gem5/blob/c9625ce9cc5b5a90a38327de5ac0e1870974af5e/src/python/m5/simulate.py#L135) 
```
    for obj in root.descendants():
        obj.createCCObject()
```
but the [SimObject::descendants](https://github.com/gem5/gem5/blob/c9625ce9cc5b5a90a38327de5ac0e1870974af5e/src/python/m5/SimObject.py#L1277) can provide limited orderring functionality ( **_ordered by name_** as said in the comment which is not intuitive I think), except that the parent will definitely be created before its children.
but the ArmISA and GIC/ genericTimer is not a parent-child relationship
```
    def descendants(self):
        yield self
        # The order of the dict is implementation dependent, so sort
        # it based on the key (name) to ensure the order is the same
        # on all hosts
        for name, child in sorted(self._children.items()):
            yield from child.descendants()
```
and the constructor,  init,  initState  act as a good way to enfore ordering when a deterministic order is required

could you please comment on this to let me know what to do next ?

gxflying: by the way,   the [miscreg.MISCREG_ID_PFR1](https://github.com/gem5/gem5/blob/c9625ce9cc5b5a90a38327de5ac0e1870974af5e/src/arch/arm/regs/misc.cc#L3508) will be right value  only if the ArmISA is constructed before generic timer 
```
  InitReg(MISCREG_ID_PFR1)
    .reset([release=release,system=system](){
        // Timer | Virti | !M Profile | TrustZone | ARMv4
        //  generic timer in system  is set in genricTomer constructor ******
        bool have_timer = (system && system->getGenericTimer() != nullptr);   // ATTENTION  here
        return 0x00000001 |
            (release->has(ArmExtension::SECURITY) ?
                0x00000010 : 0x0) |
            (release->has(ArmExtension::VIRTUALIZATION) ?
                0x00001000 : 0x0) |
            (have_timer ? 0x00010000 : 0x0);
    }())
```

src/arch/arm/isa.cc
gxflying: hi @giactra you are right,  the clear() should be moved with the oher 2 functions to InitState,  after your confirmation on the ordering of the simobject creation mentioned below,  I will handle it.

src/arch/arm/isa.cc
gxflying: for my best knowlege there are several ways to enfore ordering :
1.  parent will always constructed before its children
2. api order :   simobject::constructor(1st) which is before simobject::init()(2nd) which is before simobject::initState()(3rd)
3. objects with the same parent :  ordering depends on their name   (a not good way)

When we encounter scenarios that depend on ordering, we might consider the methods above.   
one exemple is the gic pinter in system should be set before setting the value of miscReg.pfr0.gic 
is it good to  add this information to the gem5 document?

powerjg: > is it good to add this information to the gem5 document?

You can find that information here: https://doxygen.gem5.org/release/current/classgem5_1_1SimObject.html#details

If you think there's a better place on the website, feel free to make a PR on the website repo. More/better documentation is always appreciated!

erin-le:
@giactra: Hello, just a reminder about this PR. It looks like the author still has some questions, so it'd be great if you could answer them. Thank you!

erin-le:
@giactra Hello, could you take another look at this PR when you have time? The author had some questions, and you also previously said that you would [run some tests](https://github.com/gem5/gem5/pull/1980#discussion_r1949584827). Thank you!

giactra:
Hi @gxflying , apologies for the super late reply.

> hi, @giactra thanks for your review,
> you are right that this issue probably is better to be fixed on gic side,
> and the solution might be set gic pointer in the constructor of basic_gic, (as the generic_timer has done which also need a pointer to initialize the micsReg)

Yes either that or the GICv3 CPU interface manually patching the ID register during initialization to set the bit to 1.

> but one of my concerns is that Are there any mechanisms to ensure the order in which the constructors of these simobjects are called (eg. the ISA constructor should be called before GIC/GenericTimer to fix this issue)?

Only those you mentioned below. Mainly parent before child.

> but the ArmISA and GIC/ genericTimer is not a parent-child relationship

>     def descendants(self):
>         yield self
>         # The order of the dict is implementation dependent, so sort
>         # it based on the key (name) to ensure the order is the same
>         # on all hosts
>         for name, child in sorted(self._children.items()):
>             yield from child.descendants()

We do the tree traversal with DFS. I feel doing a BFS would allow us to know almost deterministically the order. For example system.gic will be initialized before system.cpu.isa.... BFS requires a FIFO though so it might make the traversal a bit uglier...

Honestly I would be happy to accept a contribution where we patch the ID register from the Gicv3CPUInterface at init time (as I mentioned above)

erin-le:
@gxflying: Hello, could you respond to the comment above when you have time? Thank you!

erin-le:
@gxflying: Hello, just another reminder about this PR. Could you respond to giactra's comment or give an update on the status of this PR on your end? Thank you!

erin-le:
@gxflying: Hello, just another reminder to respond to this PR. Thank you!

erin-le:
@gxflying Hello, are you still interested in working on this PR? If not, would you be okay with me closing it? If you decide to come back to this PR later, it can be reopened at that point.

gxflying:
sorry for long time no updating, and I have no time for this PR.  
yes, this issue can be closed now
thanks for your time


------------------------------------------------------------------
发件人：Erin (Jianghua) ***@***.***>
日　期：2026年01月21日 05:04:01
***@***.***>
***@***.***>; ***@***.***>
主　题：Re: [gem5/gem5] arch-arm: arm misc reg for gic pfr0_el1.gic (PR #1980)

erin-le left a comment (gem5/gem5#1980)
@gxflying Hello, are you still interested in working on this PR? If not, would you be okay with me closing it? If you decide to come back to this PR later, it can be reopened at that point.

PR #1982, opened 2025-02-08
sim,stdlib: Fixes for external signal

powerjg:
Many fixes for external signal added in PR #947

- Fixed it so that if an invalid id number is sent there is no longer a panic/assert but a warning
- Many fixes in external signal code to correctly handle json input and gracefully deal with bad formats
- Many fixes in the transmitter program to correctly send data, log information, and gracefully handle errors

You can test this code with the following config script

```python
from gem5.prebuilt.demo.x86_demo_board import X86DemoBoard
from gem5.simulate.exit_handler import ExitHandler
from gem5.utils.override import overrides

# Add this class to exit_handler.py
class PrintPayloadExitHandler(ExitHandler):
    @overrides(ExitHandler)
    def _process(self, simulator: Simulator) -> None:
        print(f"Exit payload: {self._payload}")
        print(f"Tick: {simulator.get_current_tick()}")
        # I want to get the board in the simulator and get the workload and print the workload name

    @overrides(ExitHandler)
        return False

# Here we setup the board. The prebuilt X86DemoBoard allows for Full-System X86
# simulation.
board = X86DemoBoard()

workload = obtain_resource("x86-ubuntu-24.04-boot-with-systemd")
board.set_workload(workload)

simulator.update_exit_handler_id_map({
    500: PrintPayloadExitHandler 
})
simulator.run()
```

Then, you can use util/hypercall_external_signal/transmitter.py like:

```bash
./transmitter.py <gem5 PID> 500 '{"test": 10, "yay":[1, 2, 3]}'
```

And the exit handler will print the information.

Next step is to make it so that gem5 can send *back* information! For instance, the number of instructions and how many more instructions there are. Then we can build a progress bar 😄


src/python/gem5/simulate/simulator.py
```diff
@@ -812,12 +812,17 @@ def run(self, max_ticks: Optional[int] = None) -> None:
...
+            if (
                 exit_event_hypercall_id
-                in self.get_exit_handler_id_map().keys()
-            ), f"Exit event type ID {self._last_exit_event.getTypeID()} in exit handler ID map"
+                not in self.get_exit_handler_id_map().keys()
+            ):
+                warn(
+                    f"Warning: Exit event type ID "
+                    f"{self._last_exit_event.getHypercallId()} "
+                    f"not in exit handler ID map"
```
BobbyRBruce: I mentioned this in the other PR but this should tell the user what the default behavior here is (i.e., that it's going reenter the simulation loop having done nothing).

erin-le:
I tested this fix with the staging branch under the following conditions:

## Valid input 
- with tabs, excess spaces, no spaces, and newline characters
## Invalid input
### id
- no id
- a string that could be cast to an integer ("123")
- and a string that could not ("dog")
### payload
- payload formatted as a list that was either empty or had valid content
- with an int as one of the keys/values in the payload 
- with a key/value without quotation marks, or both without quotation marks
- with an empty string as one of the keys/values, or both as empty strings
- omitting a key or value, or omitting both (only a colon in the middle)
- omitting a comma between key-value pairs

Aside from a few cases where something I thought was invalid input was actually valid, everything ran as expected.

powerjg:
@BobbyRBruce I have updated this PR with your requested changes from #1995. Please approve when you have a chance so I can rebase my other PRs.

BobbyRBruce (commented):
I'm happy with the changes and just got one outstanding comment (which isn't a big issue).

src/python/gem5/simulate/simulator.py
powerjg: Good find... I *did* do this, but it got lost in a rebase :(. I'll redo it.

BobbyRBruce (approved):

PR #1984, opened 2025-02-08
mem: Fix IMP panics

hnpl:
The memory accesses induced by x86 IN instruction (an I/O instruction) was picked up by the prefetcher. I believe the memory accesses induced by an IN instruction are supposed to be uncacheable, but somehow it got past observeAccess.

This change checks if the request's response has data before letting the prefetcher inspect the packet.

### DIFF
--- a/src/mem/cache/prefetch/indirect_memory.cc
+++ b/src/mem/cache/prefetch/indirect_memory.cc
@@ -103,6 +103,11 @@ IndirectMemory::calculatePrefetch(const PrefetchInfo &pfi,
                 pt_entry->address = addr;
                 pt_entry->secure = is_secure;
 
+                // If the miss does not have any data, we don't need to
+                // inspect response data
+                if (!pfi.requestHasData()) {
+                    return;
+                }
 
                 // if this is a read, read the data from the cache and assume
                 // it is an index (this is only possible if the data is already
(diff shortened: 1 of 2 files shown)


src/mem/cache/prefetch/indirect_memory.cc
```diff
@@ -103,6 +103,11 @@ IndirectMemory::calculatePrefetch(const PrefetchInfo &pfi,
                 pt_entry->address = addr;
                 pt_entry->secure = is_secure;
 
+                // If the miss does not have any data, we don't need to
+                // inspect response data
+                if (!pfi.requestHasData()) {
```
odanrc, line 108: We typically add restrictions to accesses being observed by the prefetcher through Base::ObserveAccess. Is it possible to achieve the same results by using one of the existing parameters there?

nit: If I were to add this check here, I would add it as part of the check of line 117 instead to avoid adding this early return statement, and because it directly correlates with that block

hnpl: Thanks for the review. I believe you're right observeAccess is the place for filtering out the access, and I see that the cache maintenance accesses are already taken into account.

I looked closer to the reason why I got the panic when using IMP, I found that it was actually the memory accesses induced by x86 IN instruction (an I/O instruction) was picked up by the prefetcher. I believe the memory accesses induced by an IN instruction are supposed to be uncacheable, but somehow it got past observeAccess.

powerjg:
@hnpl should we close this PR?

PR #1988, opened 2025-02-10
Add a bi-directional communication orchestrator to gem5's stdlib

powerjg:
Note: This depends on #1982 (#1982 is included in this PR, but should be removed after it is merged.)

The gem5 orchestrator is a starting point and example of how the new
external process handler interface can be used. The orchestrator is a
bi-directional communication framework so that an external process can
both send commands to a running gem5 instance and receive responses.

The orchestrator sends commands using the externalProcessHandler in
gem5. The commands are simple json packages using the hypercall number
1000 (now reserved for this purpose). In this json message the
orchestrator can send a string function to run and a socket path to
communicate back (note that this only supports unix socket communication
right now, so likely no Windows support. This has not been tested on
MacOS.)

Two functions are initially supported:
- status: Returns the number of instructions and ticks executed so far
  and the workload that is running.
- get_stats: Returns a json object with all of the current statistic
  values.

This patch also includes a simple script that leverages the transmitter
script to run these two functions.

Note: The orchestrator will only support connecting to gem5 when its
running using the standard library. fs/se.py or other fully custom
scripts will not be supported.

Example output

image

### DIFF
--- a/src/python/gem5/components/boards/abstract_board.py
+++ b/src/python/gem5/components/boards/abstract_board.py
@@ -112,6 +112,7 @@ class AbstractBoard:
         # determined by which kind of workload is set.
         self._is_fs = None
         self._is_workload_set = False
+        self._workload: WorkloadResource = None
 
         # This variable is used to record the checkpoint directory which is
         # set when declaring the board's workload and then used by the
@@ -235,6 +236,7 @@ class AbstractBoard:
 
         :param workload: The workload to be set to this board.
         """
+        self._workload = workload
 
         try:
             func = getattr(self, workload.get_function_str())
@@ -255,6 +257,9 @@ class AbstractBoard:
                 )
 
         func(**workload.get_parameters())
+
+    def get_workload(self) -> Optional[WorkloadResource]:
+        return self._workload
 
     @abstractmethod
     def _setup_board(self) -> None:
--- a/util/hypercall_external_signal/orchestrator-request.py
+++ b/util/hypercall_external_signal/orchestrator-request.py
@@ -1,0 +1,269 @@
+#!/usr/bin/env python3
+# Copyright (c) Jason Lowe-Power
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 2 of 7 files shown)


src/python/gem5/components/boards/abstract_board.py
```diff
@@ -256,6 +258,9 @@ def set_workload(self, workload: WorkloadResource) -> None:
 
         func(**workload.get_parameters())
 
+    def get_workload(self) -> WorkloadResource:
```
BobbyRBruce:
```suggestion
    def get_workload(self) -> Optional[WorkloadResource]:
```

This could be `None` in the case a workload is not set.

util/hypercall_external_signal/transmitter.py
BobbyRBruce, line 1: I think I reviewed these changes in PR #1995

src/sim/init_signals.cc
BobbyRBruce, line 1: I think i reviewed all this code in PR #1995

util/hypercall_external_signal/orchestrator-request.py
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+Error Handling:
+    - Raises ValueError if no gem5 process found or multiple processes found
+    - Raises TimeoutError if gem5 doesn't respond within timeout period
+    - Handles SIGINT (Ctrl+C) gracefully with proper cleanup
+
+Note:
+    Requires a running gem5 simulation with hypercall support enabled
+    (gem5-v25.0 or later).
```
BobbyRBruce, line 76: This may be `gem5-v24.1.1 or later` depending on what we want to do.

util/hypercall_external_signal/orchestrator-request.py
```diff
@@ -0,0 +1,272 @@
...
+    Args:
+        conn: Connected socket to read from
+        timeout: Maximum time to wait for data in seconds (default: 30.0)
+
+    Returns:
+        str: Complete decoded message
+
+    Raises:
+        TimeoutError: If no data received within timeout period
+    """
```
BobbyRBruce: If this correct Sphinx? I thought it was in the format of:

```py
:param conn: Connect socket to read from
:param timeout: Max time to wait for data in seconds (default: 30.0)
:returns: Complete decoded message
```

util/hypercall_external_signal/orchestrator-request.py
```diff
@@ -0,0 +1,272 @@
...
+        "--pid",
+        type=int,
+        help="Process ID to send hypercall to "
+        "(auto-detected if not specified)",
+    )
+    parser.add_argument(
+        "function", choices=["status", "get_stats"], help="Function to execute"
+    )
+    parser.add_argument(
+        "--output", type=str, help="Write response to specified file"
```
BobbyRBruce, line 247: Is there much use for this? I would have assumed piping the output would work just as well.

BobbyRBruce (changes requested):
I think I reviewed this and PR #1995 in the wrong order so I skipped over the code I had reviewed in #1995 and tried to only review the code unique to this.

util/hypercall_external_signal/orchestrator-request.py
powerjg: Yeah, I think so. For big files like stats it would be a bit of a pain. It doesn't hurt much to have it.

powerjg:
Made the updates to this PR. It is now rebased on top of both #1982 and #1995. Once those two PRs are merged, we can rebase this PR and merge it.

powerjg:
Due to rebasing, it's possible that some of the intermediary changes aren't quite right. We'll have to be very careful when doing the final rebase after #1995 and #1982 are merged.

powerjg:
@BobbyRBruce this is now cleanly rebased. Can I get another review?

BobbyRBruce (approved):

PR #1991, opened 2025-02-11
sim: Add `switch_processor` function in `simulator.py`

Harshil2107:
- Added a helper function in `simulator.py` that calls `processor.switch`. This function is needed as with the updated exit event handlers and hypercalls could not guarantee access to the `processor` variable.

- Added a `switch` function to `AbstractProcessor` as a default for when `switch_processor` is called for non switchable processors.


src/python/gem5/components/processors/abstract_processor.py
```diff
@@ -37,6 +37,9 @@
     Root,
     SubSystem,
 )
+from m5.util import (
+    warn,
+)
```
powerjg: Do you have to have this across multiple lines?

src/python/gem5/components/processors/abstract_processor.py
```diff
@@ -95,3 +98,11 @@ def _pre_instantiate(self, root: Root) -> None:
         Subclasses should override this method to set up any connections.
         """
         pass
+
+    def switch(self) -> None:
```
powerjg, line 107: Do you need to add an override in `SwitchableProcessor` now?

powerjg (changes requested):
Your commit message doesn't really explain why you need this function. Presumably, you're going to use it in the future. Could you explain what you are going to use it for?

src/python/gem5/components/processors/abstract_processor.py
Harshil2107: I have added an override in `SwitchableProcessor`.

src/python/gem5/components/processors/abstract_processor.py
Harshil2107: I updated the import to a single line.

Harshil2107:
I have also updated the commit message and added why we need the switch function in the simulator class.

erin-le:
I tested this by running several simulations using the new NPB disk images and switching the processor from KVM to atomic at hypercalls 1, 2, and 4. I dumped and reset the stats at each exit, and observed that in each simulation, the ratio of hostSeconds to simSeconds significantly increased after the processor was switched, indicating that this function works as intended.

src/python/gem5/components/processors/switchable_processor.py
```diff
@@ -157,6 +157,10 @@ def switch_to_processor(self, switchable_core_key: str):
         # Ensure the current processor is updated.
         self._current_cores = to_switch
 
+    @overrides(AbstractProcessor)
+    def switch(self) -> None:
+        super().switch()
```
powerjg: I don't think this code is needed if you just call the super class code. the `SimpleSwitchableProcessor` can override `AbstractProcessor`.

erin-le (dismissed):

erin-le (approved):

powerjg (approved):

PR #1993, opened 2025-02-11
arch-arm: Fix off-by-one when initializing PMU counters

arichardson:
When configured to have e.g. 6 counters we were only adding 5 which resulted in errors when trying to configure PMU[5]. It's possible the subtraction of 1 was there to handle the cycle counters, but that is now always a separate event.

Change-Id: Idbc3e98d8aea7307929143755a03c11f9f2ec9b4


giactra (approved):
Makes sense!

PR #1994, opened 2025-02-12
sim: Add option to print exit event information

Harshil2107:
- Introduces the `--show-exit-event-messages` option in `m5.options`. When enabled, gem5 prints exit event details along with the tick at which they occur to stdout.

- Adds `simulator.show_exit_event_messages()` which has the same behaviour as running gem5 with `--show-exit-event-messages`.

- Exit event messages are now logged regardless of the `--show-exit-event-messages` setting or `simulator.show_exit_event_messages()` being called. Users can retrieve them using `simulator.get_exit_id_event_log()`.

With these changes, when running an Ubuntu boot workload with a disk image that invokes the new hypercalls and running gem5 with  `--show-exit-event-messages`, the output will look like this:

```
src/arch/x86/kvm/x86_cpu.cc:563: warn: cs limit (0xffffffff) and g (0) combination is illegal.
src/arch/x86/interrupts.cc:530: hack: Assuming logical destinations are 1 << id.
src/dev/pci/host.cc:171: warn: 00:1f.1: Write to config space on non-existent PCI device
src/dev/pci/host.cc:171: warn: 00:1f.1: Write to config space on non-existent PCI device
src/dev/x86/pc.cc:117: warn: Don't know what interrupt to clear for console.
Exit event called at Tick 1866696258178: Kernel booted.
src/sim/simulate.cc:200: info: Entering event queue @ 1866696258178.  Starting simulation...
src/dev/ps2/mouse.cc:153: warn: Unknown mouse command 0xea.
src/dev/ps2/mouse.cc:153: warn: Unknown mouse command 0xea.
src/dev/ps2/mouse.cc:153: warn: Unknown mouse command 0xe1.
src/dev/ps2/mouse.cc:153: warn: Unknown mouse command 0xe1.
src/dev/ps2/mouse.cc:153: warn: Unknown mouse command 0xa.
src/dev/ps2/mouse.cc:153: warn: Unknown mouse command 0xa.
src/dev/ps2/mouse.cc:153: warn: Unknown mouse command 0x88.
src/dev/ps2/mouse.cc:153: warn: Unknown mouse command 0x0.
Exit event called at Tick 32787037575568: Started `after_boot.sh` script.
src/sim/simulate.cc:200: info: Entering event queue @ 32787037575568.  Starting simulation...
Exit event called at Tick 32801838224437: Finished `after_boot.sh` script.
```

When calling simulator.get_exit_event_log(), the output is:

```
['Exit event called at Tick 1866696258178: Kernel booted.', 'Exit event called at Tick 32787037575568: Started `after_boot.sh` script.', 'Exit event called at Tick 32801838224437: Finished `after_boot.sh` script.']
```


powerjg:
Can you show example output on the PR description?

Harshil2107:
I have updated the PR description with an example output.

src/python/gem5/simulate/exit_handler.py
```diff
@@ -48,6 +48,9 @@ def handle(self, simulator: "Simulator") -> bool:
         self._process(simulator)
         return self._exit_simulation()
 
+    def get_exit_info(self) -> str:
+        return f"Exit Handler {self.__class__.__name__} called."
```
BobbyRBruce, line 110: 1. You can't call classes. 

2. You don't know when `get_exit_info` is going to be called so how do you know this info is being delivered at the right time?

src/python/gem5/simulate/simulator.py
```diff
@@ -821,11 +825,18 @@ def run(self, max_ticks: Optional[int] = None) -> None:
...
+                f"Exit event called at Tick {self.get_current_tick()}: "
+                f"{exit_handler.get_exit_info()}"
             )
 
+            if m5.options.show_exit_event_messages:
+                print(exit_event_info)
+
+            exit_on_completion = exit_handler.handle(self)
+            self._exit_event_log.append(exit_event_info)
+
```
BobbyRBruce: I think all the information you need to output here much more simply and doesn't require modifying the Exit Handlers:

```py
if m5.options.show_exit_event_messages:
    print(f"Hypercall ID {self._last_exit_event.getHypercallId()} called at Tick {self.get_current_tick()} "
          f"Registered Handler: {type(exit_handler)}. "
          f"Payload to pass: {self._last_exit_event.getPayload()}"
    )
```

BobbyRBruce (commented):
This is prbably more of a meta point but we need a better way to enable and disable logging gem5's Python code. Particularly the standard library.  Maybe we should look into some better logging libraries for this kind of thing? `print` statements in `if m5.do_logging:`-like branches is going to make things messy

BobbyRBruce:
Actually, I've just been thinking about this I'm going to say this shouldn't be merged. The reason I have is if someone wants logging they should be able to add it easily via the config script and we shouldn't be making our default exit events more complicated than they need to be.

The Exit Event Handlers we provide are supposed to be for default usage and if the user wants something in addition (i.e., to print things like "Kernel Booted") then they should do so via the config script:

```py
class NoisyKernelBoot(KernelBootedExitHandler):
        print("Kernel Booted")
        super._process(simulator)

simulator.update_exit_handler_id_map({1: NoisyKernelBoot})
```

In short, there's no need for new classes or making our defaults more complicated.

powerjg:
I disagree. I think this is quite useful!

src/python/gem5/simulate/simulator.py
powerjg: I disagree with this. I would like each exit event handler to have a way to pretty-print for itself. It can print the payload if that makes sense, but it doesn't have to. This is really useful for the default handlers.

src/python/gem5/simulate/exit_handler.py
powerjg: I don't understand these comments, could you provide more information? Since all exit handlers are defined statically (i.e., the `process` function is defined once per class), this code will print the name of the class (i.e., the *description* of the exit handler) by default. I'm not sure by what you mean "the right time". This function will be called when the user (or when gem5) wants to call it to get a description of the exit event handler.

Maybe a better name would be `get_handler_description`.

src/python/gem5/simulate/exit_handler.py
```diff
@@ -239,6 +262,11 @@ def _exit_simulation(self) -> bool:
 
 
 class WorkEndExitHandler(ExitHandler):
+
+    @overrides(ExitHandler)
+    def get_exit_info(self):
+        return "Finished executing Region of Interest (ROI)."
```
powerjg, line 384: For ROI begin/end... do we have the ROI number or other information about the ROI in the payload? If so, it could be useful to include it here. E.g., "Finished executing Region of Interest (ROI) labeled 1."

src/python/gem5/simulate/simulator.py
```diff
@@ -717,6 +715,12 @@ def get_roi_ticks(self) -> List[int]:
 
         return to_return
 
+    def get_exit_event_log(self) -> List[str]:
+        """
+        Returns a list of exit event IDs and the ticks at which they occurred.
+        """
+        return self._exit_event_log
```
powerjg: I don't like this function. A list of strings is not very useful. 

At one point we had an exit event log that had a map of ticks to event instances. I think that would still be useful. Then, if someone wanted to generated this string list, they could write a simple list comprehension to do that.

BobbyRBruce (commented):
@Harshil2107 : Can you respond to @powerjg requests?

i'll give a sign off after but don't have much else to add to @powerjg 's comments.

src/python/gem5/simulate/exit_handler.py
Harshil2107: I dont think there is any other information being passed, as the hypercall event only expects an integer. I am pretty sure that we can not pass a payload from the disk to the simulation(C++ side of gem5) when calling a hypercall.

src/python/gem5/simulate/simulator.py
Harshil2107: I will update the log to be a dictionary with ticks and the exit events like before.

BobbyRBruce (approved):

PR #1995, opened 2025-02-12
Improvements to exit events

powerjg:
Depends on #1982 

This updates how the new hypercall exit events are declared and also moves the code for the classic exit events out of the simulator to reduce the complexity in that object.

Example of how to use the new API:

```python
from gem5.prebuilt.demo.x86_demo_board import X86DemoBoard
from gem5.simulate.exit_handler import (
    ExitHandler,
    register_exit_handler,
    ScheduledExitEventHandler,
)

from m5.simulate import scheduleTickExitAbsolute

# You can test this file by using the transmitter program in
# util/hypercall_external_signal. To test you may need to disable
# the `scheduleTickExitAbsolute` call below.
#
# ```
# ./transmitter.py <PID of gem5> 100 '{"hello":2}'
# ./transmitter.py <PID of gem5> 101 '{"hello":5}'
# ./transmitter.py <PID of gem5> 1101 '{"hello":12}'
# ```

# This is an example of how to create a new exit handler for a bespoke
# hypercall
class MyExitHandler(ExitHandler, hypercall_num=100):
    num_hypercalls: int = 0

    def __init__(self, payload):
        super().__init__(payload)
        self.should_exit = False

    # Override the _process method to define what happens when the exit
    def _process(self, simulator) -> None:
        # Note: You have to use a class variable if you want your handler
        # to have state since each hypercall will create a new instance
        MyExitHandler.num_hypercalls += 1

        print("Got this hypercall")
        print(f"got {self.num_hypercalls} hypercalls")

        # The payload is available for use in the instance. This is different
        # for each instance of the exit.
        print(f"payload is {self._payload}")

        # The simulator is also available for use in the instance.
        print(f"the current tick is {simulator.get_current_tick()}")

        # You can use dynamic information to set the exit condition for this
        # instance.
        if simulator.get_current_tick() > 1000000000:
            self.should_exit = True
        else:
            self.should_exit = False

    def _exit_simulation(self):
        return self.should_exit


# ------------------------------------------------------------------------


# Here is an alternative way to specify the hypercall behavior.
# You can create a function that takes the simulator and payload as
# arguments and returns a boolean indicating whether to exit the
# simulation. This function can be used for many different hypercalls.
def do_a_thing(simulator, payload):
    print("Doing a thing")
    print(f"payload is {payload}")
    print(f"the current tick is {simulator.get_current_tick()}")
    return False


# here is how you register a function to handler a hyperall.
# The last parameter is a human understandable string for the hypercall
register_exit_handler(101, do_a_thing, "my handler for 101")
register_exit_handler(1101, do_a_thing, "my handler for 1101")

# ------------------------------------------------------------------------


# Here is an example of overriding an exit handler that already exists
# Note that you don't need to know the hypercall number. It is inferred
# from the parent class. The new handler will replace the old one.
class MySchedHandler(ScheduledExitEventHandler):
    def _process(self, simulator):
        super()._process(simulator)
        print("Got this scheduled event")
        print(f"the current tick is {simulator.get_current_tick()}")
        print(f"justification is {self.justification()}")

    def _exit_simulation(self):
        return False


board = X86DemoBoard()
workload = obtain_resource("x86-ubuntu-24.04-boot-with-systemd")
board.set_workload(workload)

scheduleTickExitAbsolute(10000000, "hello!")

simulator.run()
```

### DIFF
--- a/src/python/gem5/simulate/simulator.py
+++ b/src/python/gem5/simulate/simulator.py
@@ -58,13 +58,7 @@ from .exit_event_generators import (
     warn_default_decorator,
 )
 from .exit_handler import (
-    AfterBootExitHandler,
-    AfterBootScriptExitHandler,
-    ClassicGeneratorExitHandler,
-    KernelBootedExitHandler,
-    ScheduledExitEventHandler,
-    WorkBeginExitHandler,
-    WorkEndExitHandler,
+    ExitHandler,
 )
 
 
@@ -99,37 +93,6 @@ class Simulator:
         "compatible with the gem5 standard library.",
     }
 
-    def get_default_exit_handler_id_map(cls) -> Dict[int, Type[ExitEvent]]:
-        default_map = {
-            # The default exit handlers for the older "generated-based" exit
-            # events.
-            0: ClassicGeneratorExitHandler,
-            # The default exit handler for when the the kernel in a FS
-            # simulation has completed booting.
-            1: KernelBootedExitHandler,
-            # The default exit handler for when the `after_boot.sh` script has
-            # completed execution in FS simulations which utilize this.
-            2: AfterBootExitHandler,
-            # The default exit handler triggered when the post-boot script has
-            # completed execution in FS simulation. Note: it is common for this
-            # to be the the defacto end of the simulation as the post-boot
-            # script is used to execute the user's workload.
-            3: AfterBootScriptExitHandler,
-            # The default exit handler triggered when the WorkBegin and WorkEnd
-            # exit events are encountered. These are used to reset and dump
-            # stats respectively.
-            4: WorkBeginExitHandler,
-            5: WorkEndExitHandler,
-            # The default exit handler for scheduled exit event, such as those
-            # scheduled by the user via `scheduleTickExitAbsolute` or
-            # `scheduleTickExitFromCurrent`.
-            6: ScheduledExitEventHandler,
-        }
-        assert all(
-            i >= 0 for i in default_map.keys()
-        ), "Exit handler mapped to ID <0 in default map."
-        return default_map
-
     def __init__(
         self,
         board: AbstractBoard,
@@ -148,7 +111,6 @@ class Simulator:
         checkpoint_path: Optional[Path] = None,
         max_ticks: Optional[int] = m5.MaxTick,
         id: Optional[int] = None,
-        exit_event_handler_id_map: Dict[int, Type[ExitEvent]] = {},
     ) -> None:
         """
         :param board: The board to be simulated.
@@ -207,7 +169,7 @@ class Simulator:
         for handling them. The Simulator provides sensible defaults for stdlib
         exit events, but this parameter allows the user to override these
         or add handlers for custom exit events. Use
-        `Simulator.get_default_exit_handler_id_map` to see the default mapping.
+        `ExitHandler.get_handler_map` to see the mapping.
 
 
         ``on_exit_event`` usage notes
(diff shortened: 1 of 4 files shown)


powerjg:
Note: There's annoying dependencies between #1982, #1988, and this PR. Both this PR and #1988 should not be merged until after #1982 and then they will need to be rebased. This PR was built on #1988, but I tried to pull it out so it can be reviewed on its own. I may have made rebasing mistakes.

BobbyRBruce (commented):
I'm not completely sold on this redesign. I think it complicates some things and takes away some  flexibility as well as some functionality we're trying to get into Gem5.

First, this necessitates that every hypercall IDS be coupled with its own `ExitHandler` class in a 1-to-1 relationship. The reason I avoided this is the hypercall ID is the signal as to what exit type has been triggered and the Handler is the behavior expected. It deliverably decouples type and the expected response to it. This is much cleaner as many types require the same response. For example, the default behavior for a lot of events will be "Do nothing, and don't exit the simloop". In which case  we only need to define one class (say, `DoNothingNoExitHandler`) and simply map the and any all IDs to that. Taking a checkpoint is another example. My goal was for us to provide a library of exit_handlers which cover most cases and the users can just map the ID to the behavior designed. I'm thinking in particular this design may be frustrating if a workload has many exit events which, say, leave the simulation loop and the user wants to turn them all off. From my understanding, with this design, they'd need to go through and write a new function for each one. Maybe I'm missing something but I don't think we can utilize OO design here in any capacity. How can we take an existing `ExitHandler` class and just override the `_exit_simulation` function? So, for example, we still want to dump stats but we just want to have it exit the simulation after. I don't think that's possible with this design.

This design also gets rid of the ideas of an exit event carrying information to the Handler via a payload which I feel is a must as this is a really powerful feature and one of the key reasons we went ahead with refactoring the exit events. For example, I wanted to create an exit event to dump stats but only stats as specified in the payload. So the idea would be that we have many exits but at each point the payload is passed as to tell the exit handler exactly what stats should be dumped at that point. This would also allow us to do so via the signal transmitter. I think this is quite powerful: just give the hypercall ID and specify the stats needed in the payload and immediately get what you're looking for.

powerjg:
Maybe we can find a way to have the best of both worlds? The *defaults* could be specified as I have it, and we could have an API to do what you're suggesting. We could also make it so that the `hypercall_num` parameter to optionally be a list if you want one class to handle many different hypercalls.

I actually think that the 1-1 thing is a good feature. It makes things *explicit* instead of *implicit*, which is one of the main goal of the stdlib, IMO. We can have behaviors that are similar (e.g., "do nothing") but then you can inherit from that for the specific event. A good example of this is what is already done with the kernel boot, begin after_boot, etc.

> This design also gets rid of the ideas of an exit event carrying information to the Handler via a payload which I feel is a must as this is a really powerful feature and one of the key reasons we went ahead with refactoring the exit events

I don't think this is right, but maybe I'm misunderstanding you. You can still access the payload in the same way as before. See https://github.com/gem5/gem5/pull/1988/files#diff-c0c74f3318993c87996b2de533b3a2e228ca4d9cfbaf7d796df2acc2269c2064R252. I did implement this on top of this PR, but then I tried to decouple it

src/python/gem5/simulate/simulator.py
```diff
@@ -812,12 +525,17 @@ def run(self, max_ticks: Optional[int] = None) -> None:
...
                 exit_event_hypercall_id
-                in self.get_exit_handler_id_map().keys()
-            ), f"Exit event type ID {self._last_exit_event.getTypeID()} in exit handler ID map"
+                not in self.get_exit_handler_id_map().keys()
+            ):
+                warn(
+                    f"Warning: Exit event type ID "
+                    f"{self._last_exit_event.getHypercallId()} "
+                    f"not in exit handler ID map"
+                )
```
BobbyRBruce: I think it'd be good for the warning just to mention the default behavior. That is, re-entering the simulation loop having done nothing.

src/sim/init_signals.cc
```diff
@@ -213,29 +257,117 @@ externalProcessHandler(int sigtype)
...
+    std::size_t id_pos = full_payload_str.find("\"id\":");
+    if (id_pos == std::string::npos) {
+        munmap(shm_ptr, shared_mem_size);
+        close(shm_fd);
+        DPRINTF(ExternalSignal, "Error: No message ID found in payload\n");
+        return;
+    }
+
+    // Skip past "id":
+    id_pos += 5;
```
BobbyRBruce:
```suggestion
    id_pos += strlen(id_field);
```

src/sim/init_signals.cc
```diff
@@ -213,29 +257,117 @@ externalProcessHandler(int sigtype)
...
     std::string full_payload_str = full_payload;
     std::map<std::string, std::string> payload_map;
 
-    //get the hypercall id
-    std::size_t payload_pos = 0;
-    std::string hypercall_id_str = extractStringFromJSON(full_payload_str,
-        "\"id\":", ",", payload_pos);
-    uint64_t hypercall_id = std::stoi(hypercall_id_str);
+    // Find ID field
+    std::size_t id_pos = full_payload_str.find("\"id\":");
```
BobbyRBruce:
```suggestion
    std::string id_field("\"id\":\"");
    std::size_t id_pos = full_payload_str.find(id_field);
```

src/sim/init_signals.cc
```diff
@@ -213,29 +257,117 @@ externalProcessHandler(int sigtype)
...
+    while (!message_id_str.empty() && isspace(message_id_str.back())) {
+        message_id_str.pop_back();
+    }
+
+    long long id_value = std::stoll(message_id_str);
+    if (id_value < 0) {
+        munmap(shm_ptr, shared_mem_size);
+        close(shm_fd);
+        panic("Error: Invalid ID format "
+              "- must be a valid non-negative 64-bit integer\n");
```
BobbyRBruce: I think a panic here is too much. This would kill the gem5 process which would be a harsh in the case someone accidentally types an invalid id".

I wouldn't want someone 10 days into a spec simulation accidentally killing their gem5 instance.

util/hypercall_external_signal/transmitter.py
```diff
@@ -26,16 +26,65 @@
...
+    """
+    Sends a signal with payload to a gem5 process via shared memory.
+
+    Args:
+        pid: Process ID of the target gem5 process
+        id: Message ID for the signal
+        payload: String payload to send
+
+    Raises:
+        ProcessLookupError: If the specified PID does not exist
```
BobbyRBruce: Does it? I think this is caught by the `except Exception as e:` in line 125.

util/hypercall_external_signal/transmitter.py
```diff
@@ -60,46 +109,112 @@ def send_signal(pid: int, id: int, payload: str) -> None:
...
+        sys.exit(1)
+
+    logger.debug(sys.argv)
+
+    try:
+        if len(sys.argv) == 4:
+            send_signal(int(sys.argv[1]), int(sys.argv[2]), sys.argv[3])
+        else:
+            send_signal(int(sys.argv[1]), int(sys.argv[2]), "{}")
+    except ValueError as e:
```
BobbyRBruce: Under what case is a ValueError raised in `send_signal`?

BobbyRBruce (changes requested):

src/python/gem5/simulate/simulator.py
powerjg: Note to self: This is actually a comment on #1982

powerjg:
This PR is updated now. It still has #1982 in this PR, but when #1982 is merged, I'll drop that commit.

This rebase is a big pain :(

powerjg:
@BobbyRBruce, this has been rebased. Please re-review

BobbyRBruce (dismissed):

powerjg:
Fixed CI errors.

BobbyRBruce (approved):

PR #1997, opened 2025-02-12
sim: Fix MemoryError at system call

### LINKED ISSUE #1996: Inconsistent Data Type Conversion in read System Call Causes Allocation Issues in SE Mode (MemoryError: std::bad_alloc)
**Hi**
When running large Kronecker graphs in SE mode, I encountered an error (_MemoryError: std::bad_alloc_) that occurs specifically when a system call is invoked by the CPU. After investigating the issue for several days, I identified an inconsistency in data type conversion within the `read` system call implementation.

In `syscall_emul_buf.hh`, the size parameter is defined as an `int`, whereas in the call to `buf_arg(buf_ptr, nbytes)`, `nbytes` is of type `size_t`. This type mismatch can lead to incorrect memory allocations or overflows when handling large buffer sizes.

**Proposed Fix:**
To ensure correctness and consistency, the type of `size` in `syscall_emul_buf.hh` should be changed from int to `size_t`. This modification will align it with the standard definition of read system calls and prevent potential issues when handling large data sizes.

Let me know if you need additional details or test cases to reproduce the issue.

RashidAGP:
This fixes an issue with inconsistent data type conversion in the `read` system call.  
Fixes #1996

### DIFF
--- a/src/sim/syscall_emul_buf.hh
+++ b/src/sim/syscall_emul_buf.hh
@@ -63,7 +63,7 @@ class BaseBufferArg
      * Allocate a buffer of size 'size' representing the memory at
      * target address 'addr'.
      */
-    BaseBufferArg(Addr _addr, int _size)
+    BaseBufferArg(Addr _addr, size_t _size)
         : addr(_addr), size(_size), bufPtr(new uint8_t[size])
     {
         // clear out buffer: in case we only partially populate this,
@@ -96,7 +96,7 @@ class BaseBufferArg
 
   protected:
     const Addr addr;        ///< address of buffer in target address space
-    const int size;         ///< buffer size
+    const size_t size;         ///< buffer size
     uint8_t * const bufPtr; ///< pointer to buffer in simulator space
 };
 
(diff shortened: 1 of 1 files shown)


erin-le:
@RashidAGP: Thank you for your contribution! Could you update your PR and commit titles to have the format specified on pages 32 and 33 of the following guide? https://bootcamp.gem5.org/#06-Contributing/01-contributing. It is also recommended to install pre-commit, which is discussed on pages 29-31.

Essentially, you need to have one or more tags, a colon, and then the title of your PR. A list of tags can be found here: https://github.com/gem5/gem5/blob/stable/MAINTAINERS.yaml. For this PR, the PR and commit titles should be something like `sim: MemoryError at systemcall`.

Also, could you change your target branch to `develop` instead of `stable`?

erin-le:
@RashidAGP: Hello, just a reminder about this PR. If you have time, could you change the target branch of this PR to `develop` and change the PR and commit titles to something along the lines of `sim: MemoryError at systemcall`? You should be able to edit the PR title and target branch by clicking `Edit` to the right of the PR title.
Screenshot from 2025-03-05 12-05-03

erin-le:
@RashidAGP Just a heads up, I am going to push the requested changes to this PR so we can get this into develop. Thanks again for your contribution!

giactra:
One of the commits in the PR (the original) does not have a valid commit title as it does not contain a tag.
I understand @erin-le you didn't want to modify the original author commit. It makes sense.

@powerjg what is your take here? Shall we merge this as is or shall @erin-le create a new PR?

powerjg:
We can squash and merge and fix the commit message there. There's no reason for this to be two commits anyway.

BobbyRBruce (approved):

PR #1998, opened 2025-02-12
configs: Add BadAddr responder to the L2Cache XBar

arichardson:
I was running a a test binary that had a load after a conditional branch to address 0-immediate, i.e. 0xfff... which was working fine until I turned on an l2 cache. Then I was hitting`fatal("Unable to find destination for ...` which turns out to be because the L2XBar does not have a default destination.

Change-Id: I7ff624b0ef058789bd9b0ea31dbd46189769bce4

### DIFF
--- a/configs/common/CacheConfig.py
+++ b/configs/common/CacheConfig.py
@@ -136,6 +136,8 @@ def config_cache(options, system):
         )
 
         system.tol2bus = L2XBar(clk_domain=system.cpu_clk_domain)
+        system.tol2bus.badaddr_responder = BadAddr()
+        system.tol2bus.default = system.tol2bus.badaddr_responder.pio
         system.l2.cpu_side = system.tol2bus.mem_side_ports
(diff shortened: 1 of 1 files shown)


giactra:
Hi @arichardson.

There's something peculiar about your configuration.
This error shouldn't arise in a L2Xbar as by default L2 cache should cover the entire address space.

https://github.com/gem5/gem5/blob/stable/src/mem/cache/Cache.py#L139

In general we defer address resolution to the SystemXBar which should already use a BadAddr responder.
I don't think there's a harm in adding the badaddr to L2, but it would be a bit of a workaround IMHO. Maybe you fix it by properly setting the cache addr ranges

arichardson:
Yes I am very confused by this one, I will have to look at my configuration in more detail but this seemed like the simplest fix.

erin-le:
@arichardson: This PR is being closed for inactivity. If you would like to continue working on it, please feel to reopen. Thank you!

PR #1999, opened 2025-02-13
dev-amdgpu: Add RLC queues to checkpoint

v-ramadas:
This commit adds RLC queues to the checkpoint as some apps such as Lulesh use these queues after running for several iterations

### DIFF
--- a/src/dev/amdgpu/sdma_engine.cc
+++ b/src/dev/amdgpu/sdma_engine.cc
@@ -1264,6 +1264,62 @@ SDMAEngine::serialize(CheckpointOut &cp) const
     SERIALIZE_ARRAY(wptr, num_queues);
     SERIALIZE_ARRAY(size, num_queues);
     SERIALIZE_ARRAY(processing, num_queues);
+
+    // Capture RLC queue information in checkpoint
+    // Only two RLC queues are supported right now
+    const int num_rlc_queues = 2;
+    std::vector<SDMAQueue *> rlc_queues;
+    rlc_queues.push_back((SDMAQueue *)&rlc0);
+    rlc_queues.push_back((SDMAQueue *)&rlc1);
+
+    Addr rlc_info[num_rlc_queues];
+    bool rlc_valid[num_rlc_queues];
+    Addr rlc_base[num_rlc_queues];
+    Addr rlc_rptr[num_rlc_queues];
+    Addr rlc_global_rptr[num_rlc_queues];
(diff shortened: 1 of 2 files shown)


mattsinc (dismissed):
Is it possible to add some comments explaining what these do?  Otherwise looks good, thanks!

v-ramadas:
@mattsinc, I've added a few comments now. Please let me know if you'd like them to be more detailed.

mattsinc (dismissed):
thanks!  I don't think these comments really explain what all the pieces are, but I think it's sufficient.

mattsinc:
I have re-kicked off the one job that failed -- it seems the test machine crashed, nothing to do with this commit

src/dev/amdgpu/sdma_engine.cc
```diff
@@ -1264,6 +1264,56 @@ SDMAEngine::serialize(CheckpointOut &cp) const
...
+    SERIALIZE_ARRAY(rlc_info, num_rlc_queues);
+    SERIALIZE_ARRAY(rlc_valid, num_rlc_queues);
+    SERIALIZE_ARRAY(rlc_base, num_rlc_queues);
+    SERIALIZE_ARRAY(rlc_rptr, num_rlc_queues);
+    SERIALIZE_ARRAY(rlc_global_rptr, num_rlc_queues);
+    SERIALIZE_ARRAY(rlc_wptr, num_rlc_queues);
+    SERIALIZE_ARRAY(rlc_size, num_rlc_queues);
+    SERIALIZE_ARRAY(rlc_rptr_wb_addr, num_rlc_queues);
+    SERIALIZE_ARRAY(rlc_processing, num_rlc_queues);
+    SERIALIZE_ARRAY(rlc_mqd_addr, num_rlc_queues);
```
abmerop, line 1319: I believe you need to also checkpoint the MQD struct itself as it is written back to host memory when an RLC queue is unmapped. Is that happening somewhere else I might have missed?

If not, it might be easier than it looks. You should be able to get the pointer and memcpy to a temporary `uint32_t mqd[128]` array and serialize that. Reverse that to unserialize

abmerop (changes requested):

v-ramadas:
@abmerop, I have now updated the patch to checkpoint the mqd structure too.

mattsinc (approved):
This looks like it will address @abmerop 's feedback, but I'll wait for him to approve

abmerop (approved):

Issue #1487, opened 2024-08-21
crash with multiprefetcher


yichungchen84:
**Describe the bug**
multiprefecter failed when including 2+ prefetchers in the prefetcher list

**Affects version**
24.0.0.1

**gem5 Modifications**
None

**To Reproduce**

add a multiprefetcher
```python
    pf_list = [STeMSPrefetcher(prefetch_on_access=True), IndirectMemoryPrefetcher(prefetch_on_access=True)]
    prefetcher = MultiPrefetcher(prefetchers=pf_list)
```

**Terminal Output**
src/mem/cache/prefetch/base.hh
panic("PrefetchInfo::get called with a request with no data.");

```shell
#Terminal output here#
```

**Expected behavior**
running SE mode.

**Host Operating System**
Ubuntu 22.04

**Host ISA**
X86

erin-le:
Hello, could you provide a more thorough explanation of how to reproduce the issue? For instance, could you provide the configuration script you used, along with anything else that you think might be relevant?

yichungchen84:
Hi @erin-le, 

new update:

when you set up an simple example with multiprefetcher, you can easily reproduce the issue. Sorry that I cannot share the configuration with you due to IT restrictions. 

I found the problem, but I haven't dive into why this is the issue. 

Here is from 24.1.0.0
src/mem/cache/prefetch/base.cc line 68 "if ((!write && miss) || !pkt->hasData()) {"

in 23.1.0.0, it was "if (!write && miss) {"

if you remove the  "|| !pkt->hasData()", then multiprefetcher works. 

Hope this could help you.

erin-le:
I have reproduced the error and am in the process of testing the suggested fix. 

### Steps to reproduce

Add the following as `l2cache_multiprefetcher.py` in `src/python/gem5/components/cachehierarchies/classic/caches`

```python
# Copyright (c) 2021 The Regents of the University of California
# All rights reserved.
#
# Redistribution and use in source and binary forms, with or without
# modification, are permitted provided that the following conditions are
# met: redistributions of source code must retain the above copyright
# notice, this list of conditions and the following disclaimer;
# redistributions in binary form must reproduce the above copyright
# notice, this list of conditions and the following disclaimer in the
# documentation and/or other materials provided with the distribution;
# neither the name of the copyright holders nor the names of its
# contributors may be used to endorse or promote products derived from
# this software without specific prior written permission.
#
[...]

PR #2003, opened 2025-02-13
mem: fix multiprefetcher crash

erin-le:
Fixes issue 1487. This effectively reverts PR 871, which previously fixed a crash with riscv-ubuntu-run.py in the daily tests. Due to updates to the disk images, the fix in PR 871 is no longer necessary, so it shouldn't cause any problems to revert it to fix the multiprefetcher crash.

I tested this locally by running `riscv-ubuntu-run.py` and the long testlib tests.

### DIFF
--- a/src/mem/cache/prefetch/base.cc
+++ b/src/mem/cache/prefetch/base.cc
@@ -65,7 +65,7 @@ Base::PrefetchInfo::PrefetchInfo(PacketPtr pkt, Addr addr, bool miss)
     paddress(pkt->req->getPaddr()), cacheMiss(miss)
 {
     unsigned int req_size = pkt->req->getSize();
-    if ((!write && miss) || !pkt->hasData()) {
+    if (!write && miss) {
         data = nullptr;
     } else {
         data = new uint8_t[req_size];


src/mem/cache/prefetch/base.cc
```diff
@@ -65,7 +65,7 @@ Base::PrefetchInfo::PrefetchInfo(PacketPtr pkt, Addr addr, bool miss)
     paddress(pkt->req->getPaddr()), cacheMiss(miss)
 {
     unsigned int req_size = pkt->req->getSize();
-    if ((!write && miss) || !pkt->hasData()) {
```
odanrc, line 68: The check for hasData was to avoid reading data from a packet that does not have data (line 73). If we remove this check then we are likely inserting a bug. BTW, can req_size ever be 0?

odanrc (commented):
I would be more inclined to say that the bug is either in the configuration or code of the MultiPrefetcher (the sub prefetchers are both prefetch_on_access, but the multi isn't - shouldn't it?)

Also, I suggest adding to the review the original author of this change being reversed, if you haven't already.

erin-le:
@yichungchen84: Hello, this is the PR with the fix you suggested for https://github.com/gem5/gem5/issues/1487.

hnpl:
This looks very similar to what I saw in #1984 with the IMP. There is a request with the request size of 2 but does not have data.

erin-le:
I tried to reproduce the error again more recently and wasn't able to reproduce it. It may have been fixed by another change, so I am closing this PR.

PR #2006, opened 2025-02-14
Fix Stats::Vecotr::zero()

ffk0716:
Before:
nozero will print the stat if any value is zero

After:
nozero will print the stat if any value is not zero

### DIFF
--- a/src/base/statistics.hh
+++ b/src/base/statistics.hh
@@ -1004,7 +1004,7 @@ class VectorBase : public DataWrapVec<Derived, VectorInfoProxy>
     zero() const
     {
         for (off_type i = 0; i < size(); ++i)
-            if (data(i)->zero())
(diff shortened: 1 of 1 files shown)


powerjg (dismissed):

erin-le:
@ffk0716: Hello, thank you for your contribution! Could you change your target branch for this PR from `gem5:stable` to `gem5:develop`? 
Also, could add a tag to the start of your commit and PR titles? A list of tags can be found here: https://github.com/gem5/gem5/blob/stable/MAINTAINERS.yaml, and instructions for commit formatting can be found on pages 32-33 of the following tutorial: https://bootcamp.gem5.org/#06-Contributing/01-contributing. For this PR specifically, the title could be `base: Fix Stats::Vector::zero()`.

giactra (approved):

PR #2012, opened 2025-02-17
cpu: Fix postInterrupt wakeup

rogerchang23424:
The futexMap.is_waiting should only be checked in SE mode. After the #1641, the CPU will be waken by futexMap.is_waiting when isWakeUp is false in FullSystem.

Change-Id: I47f9f2d533428adb57f28b67fc0b075fb382c737


giactra (approved):
@BobbyRBruce could we backport this into the new release?

PR #2033, opened 2025-02-24
base: Add print when accessing an AssociativeCache entry

jonathanhaley:
Adds a debug print message when an AssociativeCache entry is accessed. This is useful when investigating cache entry modification or verifying replacement policies are functioning correctly.
This builds on https://github.com/gem5/gem5/pull/1514

Change-Id: I73bae68c6ce2135f21988de83703a4985275d557

### DIFF
--- a/src/base/cache/associative_cache.hh
+++ b/src/base/cache/associative_cache.hh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 Arm Limited
+ * Copyright (c) 2024-2025 Arm Limited
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
@@ -196,6 +196,13 @@ class AssociativeCache : public Named
     virtual void
     accessEntry(Entry *entry)
     {
+         if (debugFlag && debugFlag->tracing()) {
+            ::gem5::trace::getDebugLogger()->dprintf_flag(
+                curTick(), name(), debugFlag->name(),
+                "Accessing entry: %s\n", entry->print());
+        }
+
+
         replPolicy->touch(entry->replacementData);
     }
 


powerjg (dismissed):

powerjg:
Thanks for the contribution!

mattsinc (commented):
Just wanted to add my two cents that this is very useful -- thanks!

In my class, I teach students how to add prints to all the "entry" and "eviction" points in the caches.  Not asking you to do the same here, but maybe it would be useful for adding?

jonathanhaley:
Hi,
Nice to know that the change might have some utility. This change which prints on access builds on a commit (#1514 ) that prints when inserting or replacing cache entries. Which I believe would already do as you have requested? If not I must be misunderstanding. You just need to link the cache to a debug flag by calling:
`cacheName.setDebugFlag(debugFlagName)`

src/base/cache/associative_cache.hh
```diff
@@ -196,6 +196,13 @@ class AssociativeCache : public Named
     virtual void
     accessEntry(Entry *entry)
     {
+         if (debugFlag && debugFlag->tracing() && entry) {
```
odanrc: There is no need to check if entry is not null, because it is assumed in line 206 that it is not null.

(Obviously, it would have been better to have an assertion/exception be thrown if entry is null in any of these functions that make this assumption, but it is out of the scope of this PR :))

odanrc (approved):

powerjg (approved):

odanrc (approved):

mattsinc:
> Hi, Nice to know that the change might have some utility. This change which prints on access builds on a commit (#1514 ) that prints when inserting or replacing cache entries. Which I believe would already do as you have requested? If not I must be misunderstanding. You just need to link the cache to a debug flag by calling: `cacheName.setDebugFlag(debugFlagName)`

Very interesting, I'll have to test this out.  The changes I made are more directly to the replacement policies themselves (e.g., getVictim and instantiateEntry in fifo_rp.cc).  Does #1514 do the same/enable the same?

PR #2034, opened 2025-02-24
configs: Dump and reset m5 stats at kernel end in FS mode

v-ramadas:
This commit dumps and resets m5 stats whenever simulation receives kernel end exit events in GPU-FS mode. With this patch, the stats.txt file at the end of a GPU-FS simulation should contain a copy of each stat for every blit and normal kernel launched

### DIFF
--- a/configs/example/gpufs/runfs.py
+++ b/configs/example/gpufs/runfs.py
@@ -291,8 +291,12 @@ def runGpuFSSystem(args):
                 break
             kernels_completed += 1
             tasks_completed += 1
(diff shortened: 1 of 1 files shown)


powerjg (changes requested):
I'm going to push back on this. I think it could be very likely that someone using gem5 would want to have stats that are *not* reset across multiple kernels.

Note that in the stdlib, we are addressing this issue in #1995 That PR allows the user to specify what to do at different exit events.

PR #2035, opened 2025-02-24
mem-ruby: Move GPU L1 cache MSHR to the coalescer

v-ramadas:
This commit undoes PR https://github.com/gem5/gem5/pull/540 which added MSHRs to the L1 SQC/TCP to handle requests to the same line. The commit moves the MSHR instead to the coalescer where the coalescedTable maintains a list of all requests that go to the same line. These pending requests are completed as soon as the first request that went to cache returns. If any of these pending requests have GLC/SLC flags set, then the request is not marked complete. It is instead sent to cache. The commit also adds stats to track coalescer mshr misses and hits under misses

### DIFF
--- a/src/mem/ruby/protocol/GPU_VIPER-SQC.sm
+++ b/src/mem/ruby/protocol/GPU_VIPER-SQC.sm
@@ -48,9 +48,6 @@ machine(MachineType:SQC, "GPU SQC (L1 I Cache)")
 {
   state_declaration(State, desc="SQC Cache States", default="SQC_State_I") {
     I, AccessPermission:Invalid, desc="Invalid";
-    // Note: currently IV in the TCP is only for pending loads to a given cache
-    // line.  Since the SQC is read only, there are no stores.
-    IV, AccessPermission:Invalid, desc="Going from I to V, waiting on TCC data";
     V, AccessPermission:Read_Only, desc="Valid";
   }
 
@@ -103,7 +100,6 @@ machine(MachineType:SQC, "GPU SQC (L1 I Cache)")
   void unset_tbe();
   void wakeUpAllBuffers();
   void wakeUpBuffers(Addr a);
-  void wakeUpAllBuffers(Addr a);
   Cycles curCycle();
 
   // Internal functions
@@ -286,31 +282,12 @@ machine(MachineType:SQC, "GPU SQC (L1 I Cache)")
     }
   }
 
-  action(t_allocateTBE, "t", desc="allocate TBE Entry") {
-    check_allocate(TBEs);
-    TBEs.allocate(address);
-    set_tbe(TBEs.lookup(address));
-  }
-
-  action(d_deallocateTBE, "d", desc="Deallocate TBE") {
-    TBEs.deallocate(address);
-    unset_tbe();
-  }
-
-  action(st_stallAndWaitRequest, "st", desc="Stall and wait on the address") {
-    stall_and_wait(mandatoryQueue_in, address);
-  }
-
   action(p_popMandatoryQueue, "pm", desc="Pop Mandatory Queue") {
     mandatoryQueue_in.dequeue(clockEdge());
   }
 
   action(pr_popResponseQueue, "pr", desc="Pop Response Queue") {
     responseToSQC_in.dequeue(clockEdge());
-  }
-
-  action(wada_wakeUpAllDependentsAddr, "wada", desc="Wake up any requests waiting for this address") {
-    wakeUpAllBuffers(address);
   }
 
   action(l_loadDoneHit, "ldh", desc="local load done (hits in SQC)") {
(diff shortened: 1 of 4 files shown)


configs/example/gpufs/runfs.py
```diff
@@ -291,8 +291,12 @@ def runGpuFSSystem(args):
                 break
             kernels_completed += 1
             tasks_completed += 1
+            m5.stats.dump()
```
mattsinc: I don't think these should be part of this commit

src/mem/ruby/protocol/GPU_VIPER-SQC.sm
```diff
@@ -48,9 +48,6 @@ machine(MachineType:SQC, "GPU SQC (L1 I Cache)")
 {
   state_declaration(State, desc="SQC Cache States", default="SQC_State_I") {
     I, AccessPermission:Invalid, desc="Invalid";
-    // Note: currently IV in the TCP is only for pending loads to a given cache
```
mattsinc, line 51: Does all of this code need to be removed to change how the MSHRs work?  I guess your point is the states are no longer reachable?

src/mem/ruby/protocol/GPU_VIPER-TCP.sm
```diff
@@ -396,7 +392,7 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
       if (use_seq_not_coal) {
         sequencer.writeCallback(address, tmp, false, MachineType:L1Cache);
       } else {
-        coalescer.writeCallback(address, MachineType:L1Cache, tmp);
+        coalescer.writeCallback(address, MachineType:L1Cache, tmp, false);
```
mattsinc: is this change part of the MSHR fix?

src/mem/ruby/system/GPUCoalescer.cc
```diff
@@ -211,6 +211,7 @@ GPUCoalescer::GPUCoalescer(const Params &p)
                  false, Event::Progress_Event_Pri),
       uncoalescedTable(this),
       deadlockCheckEvent([this]{ wakeup(); }, "GPUCoalescer deadlock check"),
+      stats(this),
```
mattsinc, line 214: Stats should be a separate commit?

mattsinc (changes requested):

configs/example/gpufs/runfs.py
v-ramadas: Oops. I think they got added by mistake; removed now. The PR I created earlier for the stats dump and reset got rejected and I guess the commit somehow got stuck to this PR...

src/mem/ruby/protocol/GPU_VIPER-SQC.sm
v-ramadas: I removed these lines because the IV state is no longer reachable.

src/mem/ruby/protocol/GPU_VIPER-TCP.sm
v-ramadas: It's part of the MSHR stats fix. Will make that a separate commit tonight

src/mem/ruby/system/GPUCoalescer.cc
v-ramadas: Will make it a separate commit

src/mem/ruby/protocol/GPU_VIPER-TCP.sm
mattsinc: ok, let me know when these are separate.

v-ramadas: Sorry, I got sidetracked yesterday. Just updated the PR after making these separate

src/mem/ruby/system/GPUCoalescer.cc
```diff
@@ -1025,12 +1050,52 @@ GPUCoalescer::completeHitCallback(std::vector<PacketPtr> & mylist)
...
 
 void
-GPUCoalescer::recordMissLatency(CoalescedRequest* crequest,
+GPUCoalescer::recordStats(CoalescedRequest* crequest,
                                 MachineType mach,
                                 Cycles initialRequestTime,
                                 Cycles forwardRequestTime,
                                 Cycles firstResponseTime,
-                                bool success, bool isRegion)
+                                bool isRegion, bool mshrHitUnderMiss)
```
abmerop, line 1058: Nit: Are all these parameters needed?

src/mem/ruby/system/GPUCoalescer.cc
```diff
@@ -530,16 +531,38 @@ GPUCoalescer::readCallback(Addr address,
...
+    bool mshr_hit_under_miss = false;
+    // Iterate over the coalesced requests to respond to as many loads as
+    // possible until another request type is seen. Models MSHR for
+    // Coalescer. Do not respond to pending loads that have SLC/GLC flags
+    // set; issue them instead
+    while (crequest->getRubyType() == RubyRequestType_LD) {
+    hitCallback(crequest, mach, data, true,
+            crequest->getIssueTime(), forwardRequestTime, firstResponseTime,
+            isRegion, externalHit, mshr_hit_under_miss);
+
```
abmerop, line 543: Nit: Alignment

abmerop (commented):
I believe we did something similar in the past. The issue was the hit/miss stats printed by Ruby were not accurate. Is the intention to combine the new stats here + Ruby stats or are you planning a way to update the Ruby stats too?

mattsinc:
> I believe we did something similar in the past. The issue was the hit/miss stats printed by Ruby were not accurate. Is the intention to combine the new stats here + Ruby stats or are you planning a way to update the Ruby stats too?

@v-ramadas : I believe the fixes between this and the stats here, you have validated they give correct results now?

mattsinc (commented):
This looks ok to me now, but will wait for @abmerop's comments to be addresed

src/mem/ruby/system/GPUCoalescer.cc
v-ramadas: isRegion isn't used but was there in the older version. Same with the Cycles parameters; they were used to calculate a few stats long ago before the current stats class was introduced. But they can't be used anymore since some readCallback() calls set them to zero. I didn't remove them assuming they might be used in the future. Do you think its better to remove them now, and add them again if needed later?

v-ramadas:
@abmerop, @mattsinc, the new stats I added track the MSHR's effectiveness. They also update the TCP hit rate when there is an MSHR hit-under-miss. I validated them for a small test that does three loads to the same address; I see that one MSHR miss and two MSHR hits-under-misses. I also see that the TCP cache hits are updated with the two MSHR hits under misses. Additionally, if the GLC or SLC bit is set, requests go to the TCP and do not hit in the MSHR

mattsinc (approved):
Ok on my end, but will wait for @abmerop before pushing

mattsinc:
> Ok on my end, but will wait for @abmerop before pushing

@abmerop just checking on this

abmerop (approved):

src/mem/ruby/system/GPUCoalescer.cc
abmerop: Probably fine to keep until we know if they are used in other protocols but I assume they are not (publicly).

PR #2039, opened 2025-02-26
gpu-compute: Added MFMA insts, check if inst exists

mjkpolo:
A few mfmas were not added in the initial commit.
Additionally, we were only checking if the gfx version existed, and not if the instruction also existed. Now, scoreboard will panic if it cannot find the MFMA instruction in the cycle lookuptable: `mfma_cycles`


mattsinc (approved):
thanks @mjkpolo !

abmerop (approved):
Looks good - Thanks!  I was debating if missing an instruction timing is "panic-worthy" but these are easy enough to add for new ISAs that it's not a big deal.

PR #2044, opened 2025-02-28
base, arch-arm: New version of decoding for AdvSIMD

giactra:
Rewriting the decodeFpAdvSIMD function
using the new bit pattern matcher utility.
The purpose is to have a more tabular format for the
decode tree to more closely match the Arm pseudocode
specification.

As we are trying to match the machInst with all the instruction
group matchers until one is found, the time complexity is now
O(n). This means the flatter the instruction group
is, the more it might potentially take to decode an instruction.
Luckily the checking is as simple as an AND & COMPARE so even
with a big instruction pool like the SIMD one, the overhead is
negligible (and IMHO because of the sometimes complex decoding
logic we have been using, it might even make it better)

This has been tested with the following benchmark, which is
executing an instruction FMADD from the last entry in the decoding
table Fp data processing 3-source (hence testing the
"worst case scenario"):

```
int main()
{
    uint32_t prog[PROG_SIZE] = {0};

    std::fill_n(prog, PROG_SIZE, FMAD);
    prog[PROG_SIZE - 1] = RET;

    typedef double (*fmadd_func)(void);
    fmadd_func func = (fmadd_func)&prog[0];

    for (int idx = 0; idx < 10000; idx++)
        func();

    return 0;
}
```

The benchmark (run with atomic in SE mode) runs PROG_SIZE * 10000 = 1bn
FMADDs and shows no difference between introduced by the current
patch (runtime is equally ~14minutes in my machine)

### DIFF
--- a/src/arch/arm/isa/formats/aarch64.isa
+++ b/src/arch/arm/isa/formats/aarch64.isa
@@ -43,10 +43,6 @@ namespace Aarch64
 
     template <typename DecoderFeatures>
     StaticInstPtr decodeFpAdvSIMD(ExtMachInst machInst);
-    StaticInstPtr decodeFp(ExtMachInst machInst);
-    template <typename DecoderFeatures>
-    StaticInstPtr decodeAdvSIMD(ExtMachInst machInst);
-    StaticInstPtr decodeAdvSIMDScalar(ExtMachInst machInst);
 
     StaticInstPtr decodeSveInt(ExtMachInst machInst);
     StaticInstPtr decodeSveFp(ExtMachInst machInst);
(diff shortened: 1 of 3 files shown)


powerjg (approved):
Overall, I like the new approach! 

My only concern is that this approach makes it very hard to reason about whether you're covering all of the formats. As long as we have tests for all of the instructions, I think it covers this concern, but without tests for each instruction format it could lead to unexpected mistakes.

I'm not sure I would suggest all decoders for all ISAs move to this approach, but I'm happy to see it used here (and in other similar places).

giactra:
> Overall, I like the new approach!
> 

Thanks :)

> My only concern is that this approach makes it very hard to reason about whether you're covering all of the formats. As long as we have tests for all of the instructions, I think it covers this concern, but without tests for each instruction format it could lead to unexpected mistakes.

Yes, in fact it is mainly meant to be used in complex scenarios where the allocation of don't care bits makes the decoding via nested switch extremely complicated and error-prone. Where it is equally difficult to check if all patterns are covered

> 
> I'm not sure I would suggest all decoders for all ISAs move to this approach, but I'm happy to see it used here (and in other similar places).

I agree! I don't even intend to use it globally in the ArmISA: writing a decoder with this utility has O(n) time complexity (where n is the number of instructions in the pool) so it is not an optimal solution.

I believe the ideal would be to autogen a decoder from the ISA parser... But there is no active plan to implement this as of now

PR #2046, opened 2025-03-01
Analyzing Statistics in Multisim

Sir-NoChill:
# Summary

This pull request implements a simple way to pass statistics from the simulations spawned by multisim back to the parent process and addresses intermittent failure to generate statistics files.

# Motivation

In my workflow, I generally create my own `driver.py` file that is a modified version of `src/python/gem5/utils/multisim/__main__.py` and run my simulations using this file instead of the standard multisim file. I also have a caching mechanism implemented to decide when to rerun the simulations that I did not include in this PR so far (see feedback requests).

I believe this functionality would be useful to expose to users of gem5 to facilitate easier data analysis while using the simulator.

# Feedback Requests

- What would the optimal way to expose this functionality to the user be?  Ideally, I think the user would supply an additional python script that performs the analysis using the gem5 python statistics API, but I am not sure how that would look/how to implement that.
- Should the multisim module be responsible for caching executions and checking them against the existing configurations? (ie. check the .ini files and see if they change between runs, run simulations only if they do)


powerjg:
@Sir-NoChill I see you closed this PR. I want to let you know that it looked interesting to me. I haven't had a chance to dig into the code, but we are very aware that analyzing stats is a place where gem5 needs significant improvement. Contributions in this space, or even starting a conversation about features/requirements is greatly appreciated!

Sir-NoChill:
@powerjg I did start this, but my changes are incompatible with the development version of gem5 multisim so I'm trying to get them fixed before I reopen this (likely in a new pr)

PR #2080, opened 2025-03-14
scons: Explicitly define call arg in `CheckLibWithHeader`

### LINKED ISSUE #2072: [Compilation error] zlib.h not found -> flags error in zlib.h compilation sentence
Hello,

I have a problem compiling and I get the error that the zlib.h is not found. 

I think the problem comes from the fact that the zlib.h is not in a standard folder of the operating system, since I have compiled this library in my Home. 

Normally what I do to pass the headers to a compiler is:
export CFLAGS=/home/....
export CPPFLAGS=/home/...
export CXXFLAGS=/home/...

The problem is that I think the scons config file ignores these shell definitions and redefines these variables. 

Any idea how to tell the compiler where my zlib.h is?

Thank you very much,
David

Harshil2107:
This PR fixes gem5's build compatibility with SCons 4.9.0.

With the release of SCons 4.9.0, a new optional `extra_libs` argument was introduced in the `CheckLibWithHeader` function, modifying its signature from:

```python
def CheckLibWithHeader(context, libs, header, language,
                       call=None, autoadd=True, append=True, unique=False) -> bool:
```
to:
```python
def CheckLibWithHeader(context, libs, header, language,
                       extra_libs=None, call=None, autoadd=True,
                       append=True, unique=False) -> bool:
```

Since gem5 calls `CheckLibWithHeader` using positional arguments—for example:
```python
conf.CheckLibWithHeader('z', 'zlib.h', 'C++', 'zlibVersion();')
```
—in SCons 4.9.0, `'zlibVersion();'` is mistakenly assigned to `extra_libs` instead of `call`. This leads to SCons incorrectly trying to link with `lz -ll -li -lb -lV -le -lr -ls -li -lo -ln -l( -l) -l;` instead of calling `zlibversion();`, causing the build to fail.

To resolve this, this PR explicitly specifies the `call` argument:
```python
conf.CheckLibWithHeader('z', 'zlib.h', 'C++', call='zlibVersion();')
```
This ensures compatibility with both SCons 4.9.0 and older versions.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -798,10 +798,12 @@ for variant_path in variant_paths:
     with gem5_scons.Configure(env) as conf:
         # On Solaris you need to use libsocket for socket ops
         if not conf.CheckLibWithHeader(
-                [None, 'socket'], 'sys/socket.h', 'C++', 'accept(0,0,0);'):
+                [None, 'socket'], 'sys/socket.h', 'C++',
+                call='accept(0,0,0);'):
            error("Can't find library with socket calls (e.g. accept()).")
 
-        if not conf.CheckLibWithHeader('z', 'zlib.h', 'C++','zlibVersion();'):
+        if not conf.CheckLibWithHeader('z', 'zlib.h', 'C++',
+                                       call='zlibVersion();'):
             error('Did not find needed zlib compression library '
                   'and/or zlib.h header file.\n'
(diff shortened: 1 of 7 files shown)


powerjg (approved):
Thanks for getting this done so quickly!

erin-le (approved):

PR #2082, opened 2025-03-14
tests: add tests for restoring from checkpoints using multisim

erin-le:
This commit adds a test for restoring from `riscv-hello` checkpoints taken at 500,000, 1 million, and 1.5 million ticks. It also adds tests for restoring from `hello world` binaries of the Arm, X86 and RISC-V ISAs. The riscv-hello tests are currently commented out, as there are issues with obtaining the checkpoint via `obtain_resource`.


erin-le:
@BobbyRBruce: Could you take a look at this PR? Thanks!

BobbyRBruce (approved):

PR #2083, opened 2025-03-14
tests: Add tests for processor switch with MultiSim

erin-le:
This PR adds a test that checks if switching processors works when using MultiSim. It tests switching between combinations of Atomic, Timing, and O3 core types by running an X86 kernel boot on one core type, then switching to another core type and running for another 1 billion ticks.


tests/gem5/multisim/configs/x86-processor-switch.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class KernelBootProcessorSwitch(KernelBootedExitHandler):
+    def _process(self, simulator: "Simulator") -> None:
+        m5.stats.dump()
+        m5.stats.reset()
+        print("Dumping and resetting stats at kernel boot! Hypercall 1")
+        print("Switching processors at kernel boot! Hypercall 1")
+        simulator.switch_processor()
+        simulator.set_hypercall_absolute_max_ticks(1_000_000)
```
powerjg: Can we run a little longer? Maybe `1_000_000_000`?

powerjg (commented):
One tiny thing. Otherwise, this is great.

Harshil2107 (dismissed):

Harshil2107 (approved):

PR #2084, opened 2025-03-16
gpu-compute: Implement kernarg preload

abmerop:
Kernarg preloading is a way to pre-populate a wavefront's user SGPRs when the wave is created from device memory, avoiding some s_load_* instructions:

https://llvm.org/docs/AMDGPUUsage.html#preloaded-kernel-arguments

Implement this feature as it is used in newer versions of ROCm and is assumed by the compiler to be supported by gfx942.

Tested using hipblaslt-bench which uses the feature. Should fix any misbehaving application which printed "warn: Kernarg preload not implemented."

### DIFF
--- a/src/gpu-compute/gpu_command_processor.cc
+++ b/src/gpu-compute/gpu_command_processor.cc
@@ -102,34 +102,47 @@ GPUCommandProcessor::translate(Addr vaddr, Addr size)
 }
 
 void
-GPUCommandProcessor::performTimingRead(PacketPtr pkt)
+GPUCommandProcessor::performTimingRead(PacketPtr pkt, int dispType)
 {
-        // Use the shader to access the CUs and call the read request from
-        // the SQC port. Call submit kernel dispatch in the timing response
-        // function in receive timing response of SQC port. Schedule this
-        // timing read when...just currTick
-        ComputeUnit *cu = shader()->cuList[0];
-        pkt->senderState = new ComputeUnit::SQCPort::SenderState(
-                cu->wfList[0][0], true);
-        ComputeUnit::SQCPort::SenderState *sender_state =
-            safe_cast<ComputeUnit::SQCPort::SenderState*>(pkt->senderState);
-        ComputeUnit::SQCPort sqc_port = cu->sqcPort;
-        if (!sqc_port.sendTimingReq(pkt)) {
-                sqc_port.retries.push_back(
-                        std::pair<PacketPtr, Wavefront*>(pkt,
-                            sender_state->wavefront));
-        }
+    // Use the shader to access the CUs and call the read request from
+    // the SQC port. Call submit kernel dispatch in the timing response
+    // function in receive timing response of SQC port. Schedule this
+    // timing read when...just currTick
+    ComputeUnit *cu = shader()->cuList[0];
+    pkt->senderState = new ComputeUnit::SQCPort::SenderState(
+            cu->wfList[0][0], true);
+    ComputeUnit::SQCPort::SenderState *sender_state =
+        safe_cast<ComputeUnit::SQCPort::SenderState*>(pkt->senderState);
+    sender_state->dispatchType = dispType;
+    ComputeUnit::SQCPort sqc_port = cu->sqcPort;
+
+    if (!sqc_port.sendTimingReq(pkt)) {
+        sqc_port.retries.push_back(
+            std::pair<PacketPtr, Wavefront*>(pkt, sender_state->wavefront)
+        );
+    }
 }
 
 void
-GPUCommandProcessor::completeTimingRead()
+GPUCommandProcessor::completeTimingRead(int dispType)
 {
-        struct KernelDispatchData dispatchData = kernelDispatchList.front();
-        kernelDispatchList.pop_front();
-        delete dispatchData.readPkt;
-        if (kernelDispatchList.size() == 0)
-                dispatchKernelObject(dispatchData.akc, dispatchData.raw_pkt,
-                        dispatchData.queue_id, dispatchData.host_pkt_addr);
+    struct KernelDispatchData dispatchData = kernelDispatchList.front();
+    kernelDispatchList.pop_front();
+    delete dispatchData.readPkt;
+
+    // Only one of the following can happen at any time from one CP. Figure
+    // out what performed the timing read and call to appropriate function.
+    if (kernelDispatchList.size() == 0) {
+        switch (dispType) {
+          case ComputeUnit::SQCPort::SenderState::DISPATCH_KERNEL_OBJECT:
+            dispatchKernelObject(dispatchData.akc, dispatchData.raw_pkt,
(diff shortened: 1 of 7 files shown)


abmerop:
A few things to make review easier:

- The new `GPUCommandProc::readPreload` is nearly the same as `GPUCommandProc::submitDispatchPkt`, so it is easier to compare with that.
- `GPUCommandProcessor::performTimingRead` and `GPUCommandProcessor::completeTimingRead` are mostly style fixes (indentation) plus setting/reading the new `dispatchType` member in `SQCPort::SenderState`. There is less to review there than it looks like.

src/gpu-compute/gpu_command_processor.cc
```diff
@@ -102,34 +102,47 @@ GPUCommandProcessor::translate(Addr vaddr, Addr size)
...
-                cu->wfList[0][0], true);
-        ComputeUnit::SQCPort::SenderState *sender_state =
-            safe_cast<ComputeUnit::SQCPort::SenderState*>(pkt->senderState);
-        ComputeUnit::SQCPort sqc_port = cu->sqcPort;
-        if (!sqc_port.sendTimingReq(pkt)) {
-                sqc_port.retries.push_back(
-                        std::pair<PacketPtr, Wavefront*>(pkt,
-                            sender_state->wavefront));
-        }
+GPUCommandProcessor::performTimingRead(PacketPtr pkt, int dispType)
```
mattsinc, line 105: As far as I can tell, this is just spacing changed -- fine

src/gpu-compute/gpu_command_processor.cc
```diff
@@ -102,34 +102,47 @@ GPUCommandProcessor::translate(Addr vaddr, Addr size)
...
 {
-        struct KernelDispatchData dispatchData = kernelDispatchList.front();
-        kernelDispatchList.pop_front();
-        delete dispatchData.readPkt;
-        if (kernelDispatchList.size() == 0)
-                dispatchKernelObject(dispatchData.akc, dispatchData.raw_pkt,
-                        dispatchData.queue_id, dispatchData.host_pkt_addr);
+    struct KernelDispatchData dispatchData = kernelDispatchList.front();
+    kernelDispatchList.pop_front();
+    delete dispatchData.readPkt;
```
mattsinc, line 131: Should this be deleted after the if/switch to ensure no data corruption?

I guess it's ok because different fields are accessed?

src/gpu-compute/gpu_command_processor.cc
```diff
@@ -692,6 +721,116 @@ GPUCommandProcessor::signalWakeupEvent(uint32_t event_id)
...
+            PacketPtr readPkt = new Packet(request, MemCmd::ReadReq);
+            readPkt->dataStatic((uint8_t *)task->preloadArgs()
+                                 + gen.complete());
+
+            struct KernelDispatchData dispatchData;
+            dispatchData.akc = akc;
+            dispatchData.task = task;
+            dispatchData.readPkt = readPkt;
+            kernelDispatchList.push_back(dispatchData);
+            performTimingRead(readPkt,
```
mattsinc, line 815: can you add a comment about why we do a functional and then timing read?

src/gpu-compute/gpu_command_processor.cc
```diff
@@ -292,6 +306,16 @@ GPUCommandProcessor::dispatchKernelObject(AMDKernelCode *akc, void *raw_pkt,
...
     _hsa_dispatch_packet_t *disp_pkt = (_hsa_dispatch_packet_t*)raw_pkt;
 
+    /**
+     * If the kernarg_preload_spec_length is non-zero, the CP firmware will
+     * append an additional 256 bytes to the kernel_code_entry_byte_offset.
+     *
+     * https://llvm.org/docs/AMDGPUUsage.html#amdgpu-amdhsa-kernarg-preload
+     */
+    if (akc->kernarg_preload_spec_length != 0) {
+        akc->kernel_code_entry_byte_offset += 256;
```
mattsinc: see comment below

src/gpu-compute/gpu_command_processor.cc
```diff
@@ -692,6 +721,116 @@ GPUCommandProcessor::signalWakeupEvent(uint32_t event_id)
...
 }
 
+void
+GPUCommandProcessor::readPreload(AMDKernelCode *akc, HSAQueueEntry *task)
+{
+    _hsa_dispatch_packet_t *disp_pkt =
+        (_hsa_dispatch_packet_t*)task->dispPktPtr();
+
+    Addr preload_addr = (Addr)disp_pkt->kernel_object
+        + akc->kernel_code_entry_byte_offset - 256;
```
mattsinc: see comment below about magic number

src/gpu-compute/hsa_queue_entry.hh
```diff
@@ -504,6 +522,11 @@ class HSAQueueEntry
     std::bitset<NumScalarInitFields> initialSgprState;
 
     unsigned _accumOffset;
+
+    // For preloading args there is 256 bytes of space so we can preload
+    // 256 / sizeof(uint32_t) = 64 user SGPRs. The length is the exact amount.
+    unsigned _preloadLength = 0;
+    uint32_t _preloadArgs[256 / sizeof(uint32_t)];
```
mattsinc: Can we add a #define for this (256) instead of having a magic number used?

(this is true throughout)

mattsinc (changes requested):
some minor things, but otherwise good, thanks!

src/gpu-compute/hsa_queue_entry.hh
abmerop: Done

src/gpu-compute/gpu_command_processor.cc
abmerop: This does a functional translation, same as the `submitDispatchPkt` function. I think the idea was we don't really care about simulating the timing here and doing the translation functionally makes the code way easier.

src/gpu-compute/gpu_command_processor.cc
abmerop: Done

src/gpu-compute/gpu_command_processor.cc
abmerop: Done

src/gpu-compute/gpu_command_processor.cc
abmerop: Yes, we don't need the readPkt field anymore. Since it is not used in this function the delete location doesn't really matter functionally.

src/gpu-compute/gpu_command_processor.cc
abmerop: Correct

abmerop (commented):
Fixed a problem with the original implementation with what the starting SGPR is. Actually ready now.

mattsinc (approved):

PR #2087, opened 2025-03-17
Add HiFive Unleashed/Unmatched DMA engine model

bombanya:
Add model of the SiFive Platform DMA Engine (PDMA). This DMA engine is a part of FU540 and FU740 SoCs (used by HiFive Unleashed and Unmatched boards). This is a basic DMA memcpy controller which can be used to transfer data between devices and main memory or copy data between two locations in memory.

PDMA is decribed in SiFive FU540 Manual, Chapter 12:
https://static.dev.sifive.com/FU540-C000-v1.0.pdf

PDMA gem5 model is compatible with sf-pdma Linux driver and can be used as a generator of the DMA-requests.

Also a small demo with Linux dmatest module is available: https://drive.google.com/drive/folders/10m4tfQVQPKWBl1Gt_kMV2JktqFn2Uwsm?usp=sharing
You can use it to see how the new model works.

**_Pay attention:_**
For convenience and usage demonstration (and also because physical FU540 has a PDMA on-board) new model was added to the existent gem5 HiFive platform. But this platform did not support devices with DMA ports that required additional changes to it and also led to changes to the boards that use this platform (`RiscvBoard` and `RISCVMatchedBoard`). The tests are passing and it looks like the existing scripts using these boards are working, but I'm not an expert in the gem5 stdlib at all, so please double check these changes.

### DIFF
--- a/src/dev/riscv/HiFive.py
+++ b/src/dev/riscv/HiFive.py
@@ -35,18 +35,29 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+from typing import (
+    Generator,
+    List,
+)
+
 from m5.objects.Clint import Clint
+from m5.objects.ClockedObject import ClockedObject
 from m5.objects.PciHost import GenericPciHost
 from m5.objects.PciUpstream import PciBus
 from m5.objects.Platform import Platform
 from m5.objects.Plic import Plic
 from m5.objects.PMAChecker import PMAChecker
 from m5.objects.RTC import RiscvRTC
+from m5.objects.Sf_PDMA import SfPDMA
 from m5.objects.Terminal import Terminal
 from m5.objects.Uart import RiscvUart8250
+from m5.objects.XBar import BaseXBar
 from m5.params import *
+from m5.params.port_params import PortRef
 from m5.proxy import *
 from m5.util.fdthelper import *
+
+from gem5.utils.override import overrides
 
 
 class GenericRiscvPciHost(GenericPciHost):
@@ -104,15 +115,31 @@ class HiFiveBase(Platform):
     # Set to 0 if using a pci interrupt for Uart instead
     uart_int_id = Param.Int(0, "PLIC Uart interrupt ID")
 
-    def _on_chip_devices(self):
+    def _on_chip_devices(self) -> List[ClockedObject]:
         """Returns a list of on-chip peripherals"""
         return []
 
-    def _off_chip_devices(self):
+    def _off_chip_devices(self) -> List[ClockedObject]:
         """Returns a list of off-chip peripherals"""
         return []
 
-    def _on_chip_ranges(self):
+    def _on_chip_ports_resp(self) -> List[PortRef]:
+        """Returns a list of on-chip peripherals response ports"""
+        return []
+
+    def _off_chip_ports_resp(self) -> List[PortRef]:
+        """Returns a list of off-chip peripherals response ports"""
+        return []
+
+    def _on_chip_ports_req(self) -> List[PortRef]:
+        """Returns a list of on-chip peripherals request/dma ports"""
+        return []
+
+    def _off_chip_ports_req(self) -> List[PortRef]:
+        """Returns a list of off-chip peripherals request/dma ports"""
+        return []
+
+    def _on_chip_ranges(self) -> List[AddrRange]:
         """Returns a list of on-chip peripherals
         address range
         """
@@ -121,7 +148,7 @@ class HiFiveBase(Platform):
             for dev in self._on_chip_devices()
         ]
 
-    def _off_chip_ranges(self):
+    def _off_chip_ranges(self) -> List[AddrRange]:
         """Returns a list of off-chip peripherals
         address range
         """
@@ -130,19 +157,19 @@ class HiFiveBase(Platform):
             for dev in self._off_chip_devices()
         ]
 
-    def attachOnChipIO(self, bus):
-        """Attach on-chip IO devices, needs modification
-        to support DMA
-        """
-        for device in self._on_chip_devices():
-            device.pio = bus.mem_side_ports
+    def attachOnChipIO(self, bus: BaseXBar, ruby: bool = False) -> None:
+        for port in self._on_chip_ports_resp():
+            port.connect(bus.mem_side_ports)
+        if not ruby:
+            for port in self._on_chip_ports_req():
+                port.connect(bus.cpu_side_ports)
 
-    def attachOffChipIO(self, bus):
-        """Attach off-chip IO devices, needs modification
-        to support DMA
-        """
-        for device in self._off_chip_devices():
-            device.pio = bus.mem_side_ports
+    def attachOffChipIO(self, bus: BaseXBar, ruby: bool = False) -> None:
+        for port in self._off_chip_ports_resp():
+            port.connect(bus.mem_side_ports)
+        if not ruby:
+            for port in self._off_chip_ports_req():
+                port.connect(bus.cpu_side_ports)
 
 
 class HiFive(HiFiveBase):
@@ -198,38 +225,83 @@ class HiFive(HiFiveBase):
     uart_int_id = 0xA
     terminal = Terminal()
 
-    def _on_chip_devices(self):
+    # DMA-controller
+    pdma = SfPDMA(
+        pio_addr=0x3000000,
+        done_irq=[i for i in range(23, 30, 2)],
+        error_irq=[i for i in range(24, 31, 2)],
+        _dma_coherent=True,
+    )
+
+    _off_chip_ports_extra = []
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+        for i in range(self.pdma.chan_cnt):
+            self._off_chip_ports_extra.append(self.pdma.dma[i])
+
+    @overrides(HiFiveBase)
+    def _on_chip_devices(self) -> List[ClockedObject]:
         """Returns a list of on-chip peripherals"""
         return [self.clint, self.plic]
 
-    def _off_chip_devices(self):
+    @overrides(HiFiveBase)
+    def _off_chip_devices(self) -> List[ClockedObject]:
         """Returns a list of off-chip peripherals"""
-        devices = [self.uart]
+        devices = [self.uart, self.pdma]
         if hasattr(self, "disk"):
             devices.append(self.disk)
         if hasattr(self, "rng"):
             devices.append(self.rng)
         return devices
 
-    def attachPlic(self):
+    @overrides(HiFiveBase)
+    def _on_chip_ports_resp(self) -> List[PortRef]:
+        ports = []
+        for dev in self._on_chip_devices():
+            if hasattr(dev, "pio"):
+                ports.append(dev.pio)
+        return [i for i in ports if i.role == "GEM5 RESPONDER"]
+
+    @overrides(HiFiveBase)
+    def _off_chip_ports_resp(self) -> List[PortRef]:
+        ports = []
+        for dev in self._off_chip_devices():
+            if hasattr(dev, "pio"):
+                ports.append(dev.pio)
+        ports += self._off_chip_ports_extra
+        return [i for i in ports if i.role == "GEM5 RESPONDER"]
+
+    @overrides(HiFiveBase)
+    def _off_chip_ports_req(self) -> List[PortRef]:
+        ports = self._off_chip_ports_extra
+        return [i for i in ports if i.role == "GEM5 REQUESTOR"]
+
+    def attachPlic(self) -> None:
         """Count and set number of PLIC interrupt sources"""
         plic_srcs = [
             self.uart_int_id,
             self.pci_host.int_base + self.pci_host.int_count,
+            *self.pdma.done_irq,
+            *self.pdma.error_irq,
         ]
         for device in self._off_chip_devices():
             if hasattr(device, "interrupt_id"):
                 plic_srcs.append(device.interrupt_id)
         self.plic.n_src = max(plic_srcs) + 1
 
-    def setNumCores(self, num_cpu):
+    def setNumCores(self, num_cpu: int) -> None:
         """Sets the CLINT to number of threads and the PLIC hartID/pmode for
         each contexts. Assumes that the cores have a single hardware thread.
         """
         self.plic.hart_config = ",".join(["MS" for _ in range(num_cpu)])
         self.clint.num_threads = num_cpu
 
-    def generateDeviceTree(self, state):
+    @overrides(SimObject)
+    def generateDeviceTree(
+        self, state: FdtState
+    ) -> Generator[FdtNode, None, None]:
         cpus_node = FdtNode("cpus")
         cpus_node.append(FdtPropertyWords("timebase-frequency", [10000000]))
         yield cpus_node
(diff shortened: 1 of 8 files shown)


bombanya:
@powerjg @erin-le
Gently pinging. Any requests/comments for me, please?

Thanks.

BobbyRBruce:
> @powerjg @erin-le Gently pinging. Any requests/comments for me, please?
> 
> Thanks.

Apologies, this has been on my TODO list for the past week. Will review today!

src/python/gem5/prebuilt/riscvmatched/riscvmatched_board.py
BobbyRBruce, line 1: I can see that this was born out of a desire to integrate the DMA engine into the RISCVMatched board but I think this change needs reverted. The purpose of this board was to be stable and separate with preset sensible values to emulate this boards behavior. These values in here are fixed and were decided upon as values thought best to represent this board in function. I see a lot of it here has had to be removed to make this a subclass of the RISC-V board. I think it's best to keep these separate. it may seem redundant, and it is, but I'd rather get rid of this entirely than make it a subclass and change it's functionality.

src/python/gem5/components/boards/riscv_board.py
```diff
@@ -210,14 +208,17 @@ def _setup_pma(self) -> None:
...
-            "Cannot execute `get_dma_ports()`: Board does not have DMA ports "
-            "to return. Use `has_dma_ports()` to check this."
-        )
+        # The DMA ports differ depending upon the cache hierarchy.
+        if self.get_cache_hierarchy().is_ruby():
+            if self._dma_ports is None:
+                self._dma_ports = []
+
+        # _dma_ports should always be empty for classic caches.
+        return self._dma_ports
```
BobbyRBruce: What's the expected behavior here if the cache hierarchy is classic? Is it possible for this function to return none?

src/python/gem5/components/boards/riscv_board.py
```diff
@@ -162,17 +159,17 @@ def _setup_io_devices(self) -> None:
...
+            self.platform.attachOnChipIO(self.iobus, self.get_dma_ports())
 
         else:
-            for device in self._off_chip_devices:
-                device.pio = self.iobus.mem_side_ports
-            for device in self._on_chip_devices:
-                device.pio = self.get_cache_hierarchy().get_mem_side_port()
+            self.platform.attachOnChipIO(
+                self.get_cache_hierarchy().membus, self.get_dma_ports()
+            )
```
BobbyRBruce: I'd avoid directly accessing the membus here. The classic caches have two functions `get_mem_side_port` and `get_cpu_side_port` which can be used to get the the objects I think you need here.

src/dev/riscv/HiFive.py
```diff
@@ -129,19 +138,25 @@ def _off_chip_ranges(self):
...
-
-    def attachOffChipIO(self, bus):
-        """Attach off-chip IO devices, needs modification
-        to support DMA
-        """
-        for device in self._off_chip_devices():
-            device.pio = bus.mem_side_ports
+    def attachOnChipIO(self, bus, dma_ports=None):
+        for port in self._on_chip_ports():
+            if port.role == "GEM5 RESPONDER":
```
BobbyRBruce:
```suggestion
            if isinstance(port, ResponsePort):
```

src/dev/riscv/HiFive.py
```diff
@@ -129,19 +138,25 @@ def _off_chip_ranges(self):
...
+                port.connect(bus.mem_side_ports)
+            else:
+                if dma_ports is None:
+                    port.connect(bus.cpu_side_ports)
+                else:
+                    dma_ports.append(port)
+
+    def attachOffChipIO(self, bus, dma_ports=None):
+        for port in self._off_chip_ports():
+            if port.role == "GEM5 RESPONDER":
```
BobbyRBruce:
```suggestion
            if isinstance(port, ResponsePort):
```

BobbyRBruce (changes requested):
I broadly support this change, though I do have some things noted in my review that i'd appreciate being addressed before merging.

src/dev/riscv/HiFive.py
bombanya: Are you sure that it is better to use `isinstance()` here? 
As I understand it, `attachOnChipIO()/attachOffChipIO()` are called with ports which are instances of the `class PortRef` and to get back from them to the original `Request/ResponsePort` we would need to use something like:
```suggestion
            real_port = port.simobj._ports[port.name]
            if isinstance(real_port, ResponsePort) or isinstance(
                real_port, VectorResponsePort
            ):
```
In my opinion, version with just `port.role` looks better.

src/python/gem5/components/boards/riscv_board.py
bombanya: The expected behavior for classic caches is to return None. `_dma_ports` is initialized [here](https://github.com/gem5/gem5/pull/2087/files#diff-7cce437d81e063378e32f58b48e55b95a06f835420d36b39b46eeb72b3567a6cR103) in the constructor and, for classic caches, remains unchanged for the whole lifetime. Because of that `attachOnChipIO()/attachOffChipIO()` in HiFive platform are called with None for `dma_ports` and these functions connect ports as we expect.

bombanya (commented):
Thanks for the review, @BobbyRBruce !

src/python/gem5/components/boards/riscv_board.py
bombanya: Fixed it [this way](https://github.com/gem5/gem5/pull/2087/files#diff-7cce437d81e063378e32f58b48e55b95a06f835420d36b39b46eeb72b3567a6cR171-R178)

src/python/gem5/prebuilt/riscvmatched/riscvmatched_board.py
bombanya: The problem here is that I added a new device to the platform that RISCVMatchedBoard uses, so we cannot keep it unchanged, because this new DMA device remains unconnected and simulation fails. I can offer several options on how to fix it:
- It is possible to make sf-pdma addition to the platform optional. For example, as it was done with SMMU in the ARM `VExpress_GEM5_Base`;
- If HiFive remains as it is made now in my PR:
   - It is possible to connect sf-pdma properly, but that requires changes identical to those for `riscv_board` and looks like you want to keep RISCVMatchedBoard with minimal changes;
   - It is possible to just remove sf-pdma from the platform especially for this board. For example, as I implemented it [here](https://github.com/gem5/gem5/pull/2087/files#diff-68a1ed1c9e88c6e3412deb7cb4f5a21086d7ac903531601cc85a21a3acbbe86eR157)

src/python/gem5/components/boards/riscv_board.py
BobbyRBruce: but you state above in `has_dma_ports` that the system has DMA ports if your in FS mode. SO I'm in FS mode and use classic caches, this returns `None`. What does that mean?

src/dev/riscv/HiFive.py
BobbyRBruce: You can leave it this way if you want but I'm always quite cautious at inferring types via comparison of strings. it only takes one typo to ruin your day

src/python/gem5/components/boards/riscv_board.py
bombanya: Ah, now I understand what you meant. Yes, the logic turned out to be strange, but I copied it from [arm_board](https://github.com/gem5/gem5/blob/stable/src/python/gem5/components/boards/arm_board.py#L319) so I had the impression that I was doing it right.

I've [updated this function](https://github.com/gem5/gem5/pull/2087/files#diff-7cce437d81e063378e32f58b48e55b95a06f835420d36b39b46eeb72b3567a6cR222-R226): now it returns the complete list of the dma ports when in FS mode for both caches. That required some [changes for port connecting functions](https://github.com/gem5/gem5/pull/2087/files#diff-ac1ee32f2b73ad1a3c9faadab59817e09670fb22a3ae2a080bfba892035a6a3fR141-R159) from HiFive platform.

src/dev/riscv/HiFive.py
```diff
@@ -111,6 +112,14 @@ def _off_chip_devices(self):
...
         return []
 
+    def _on_chip_ports(self):
+        """Returns a list of on-chip peripherals request/response ports"""
+        return []
+
+    def _off_chip_ports(self):
+        """Returns a list of off-chip peripherals request/response ports"""
+        return []
+
```
BobbyRBruce: Not a hard requirement for this PR as we didn't have these in the HiFive code prior to this PR, but it would be nice if the functions included typing hints for return types.

e.g.:

```suggestion
    def _on_chip_ports(self) -> List[Port]:
        """Returns a list of on-chip peripherals request/response ports"""
        return []

    def _off_chip_ports(self) -> List[Port]:
        """Returns a list of off-chip peripherals request/response ports"""
        return []

```

src/python/gem5/components/boards/riscv_board.py
```diff
@@ -210,14 +216,14 @@ def _setup_pma(self) -> None:
...
     @overrides(AbstractSystemBoard)
     def get_dma_ports(self) -> List[Port]:
-        raise Exception(
-            "Cannot execute `get_dma_ports()`: Board does not have DMA ports "
-            "to return. Use `has_dma_ports()` to check this."
-        )
+        if self.has_dma_ports():
+            return self._dma_ports
+        else:
+            return None
```
BobbyRBruce: We should preserve the original functionality. By returning `None` in the case we don't have DMA ports we're breaking that (it's also not specified by the return type that this function may return a `None` value. Id just suggest:


```suggestion
        if not self.has_dma_ports():
           raise Exception(
            "Cannot execute `get_dma_ports()`: Board does not have DMA ports "
            "to return. Use `has_dma_ports()` to check this."
        )
        return self._dma_ports
```

src/python/gem5/components/boards/riscv_board.py
```diff
@@ -162,17 +160,24 @@ def _setup_io_devices(self) -> None:
...
-            for device in self._on_chip_devices:
-                device.pio = self.get_cache_hierarchy().get_mem_side_port()
+            membus = attrdict()
+            membus["mem_side_ports"] = (
+                self.get_cache_hierarchy().get_mem_side_port()
+            )
+            membus["cpu_side_ports"] = (
+                self.get_cache_hierarchy().get_cpu_side_port()
+            )
+            self.platform.attachOnChipIO(membus, False, self._dma_ports)
```
BobbyRBruce: I'm confused by this 'attrdict' work around. it seems to me that the most robust solution would be to create a `SystemXBar` SimObject for this connection. Is there any reason not to do that?

src/dev/riscv/HiFive.py
```diff
@@ -196,24 +211,57 @@ class HiFive(HiFiveBase):
...
 
+    # DMA-controller
+    pdma = SfPDMA(
+        pio_addr=0x3000000,
+        done_irq=[i for i in range(23, 30, 2)],
+        error_irq=[i for i in range(24, 31, 2)],
+        _dma_coherent=True,
+    )
+
+    _off_chip_dev_extra = []
```
BobbyRBruce: As far as I can see this is never used. What's it here for?

src/dev/riscv/HiFive.py
```diff
@@ -129,19 +138,25 @@ def _off_chip_ranges(self):
...
-            device.pio = bus.mem_side_ports
+    def attachOnChipIO(self, bus, ruby=False, dma_ports=None):
+        for port in self._on_chip_ports():
+            if port.role == "GEM5 RESPONDER":
+                port.connect(bus.mem_side_ports)
+            else:
+                if dma_ports is not None:
+                    dma_ports.append(port)
+                if not ruby:
+                    port.connect(bus.cpu_side_ports)
```
BobbyRBruce: I'm struggling to understand this logic. If the dma_ports is not none, then add the ports to the dma_ports list. Under what case do we pass `None` here and in in what cases do we pass a non-None value? And why if we pass a non-None value do we assume all on-chip ports are dma ports? 

Can we not split this function into two (well 4: 2 new functions or both the "on chip" and "off chip" cases). One funcioton that attaches the bus correctly to the devices, and another that simply returns the dma_ports. I don't think it's necessary to create a list of DMA ports while you're attaching them to the bus.

src/dev/riscv/HiFive.py
```diff
@@ -129,19 +138,25 @@ def _off_chip_ranges(self):
...
-        for device in self._on_chip_devices():
-            device.pio = bus.mem_side_ports
-
-    def attachOffChipIO(self, bus):
-        """Attach off-chip IO devices, needs modification
-        to support DMA
-        """
-        for device in self._off_chip_devices():
-            device.pio = bus.mem_side_ports
+    def attachOnChipIO(self, bus, ruby=False, dma_ports=None):
```
BobbyRBruce: Likewise, it would be nice to start proving typing hints some of these functions as they grow in complexity (again, not a hard requirement for this PR, but it would be nice to have):

```suggestion
    def attachOnChipIO(self, bus : SystemXBar, ruby : bool=False, dma_ports : Optional[Port]=None ) -> None:
```

BobbyRBruce (commented):
My review is somewhat partial as I want to test this locally. Recently #2298 was merged and there may be some refactoring needed to this PR as a result. I'll spend a little time investigating this and report back.

erin-le:
@bombanya: Hello, could you respond to the comments on this PR when you get a chance? Could you also take a look at the conflicts in this PR? Thank you!

bombanya:
I'm sorry for the delay, guys. I will respond to comments and resolve conflicts this weekend.

src/dev/riscv/HiFive.py
bombanya: [Added](https://github.com/gem5/gem5/pull/2087/files#diff-ac1ee32f2b73ad1a3c9faadab59817e09670fb22a3ae2a080bfba892035a6a3f)

src/python/gem5/components/boards/riscv_board.py
bombanya: [Done](https://github.com/gem5/gem5/pull/2087/files#diff-7cce437d81e063378e32f58b48e55b95a06f835420d36b39b46eeb72b3567a6cR242-R249)

src/python/gem5/components/boards/riscv_board.py
bombanya: Apparently, I need a little more help here, I don't really understand how to do this.

In fact, we already have an XBar which ports we want to use. We can do something like this: 
```
self.membus = SystemXBar(
                mem_side_ports=self.get_cache_hierarchy().get_cpu_side_port(),
                cpu_side_ports=self.get_cache_hierarchy().get_mem_side_port(),
            )
self.platform.attachOnChipIO(self.membus, False)
```

but it will add another bus to the hierarchy, it seems that this is not what we need.

src/dev/riscv/HiFive.py
bombanya: Used it for experiments, forgot to remove.

src/dev/riscv/HiFive.py
bombanya: [Separated functions that return ports into those that return requestors and responders.](https://github.com/gem5/gem5/pull/2087/files#diff-ac1ee32f2b73ad1a3c9faadab59817e09670fb22a3ae2a080bfba892035a6a3fR126-R140)

['attach' functions are now only concerned with connecting ports.](https://github.com/gem5/gem5/pull/2087/files#diff-ac1ee32f2b73ad1a3c9faadab59817e09670fb22a3ae2a080bfba892035a6a3fR160-R165)

[has_dma_ports() from riscv_board now simply calls dma ports getters.](https://github.com/gem5/gem5/pull/2087/files#diff-7cce437d81e063378e32f58b48e55b95a06f835420d36b39b46eeb72b3567a6cR247-R249)

src/dev/riscv/HiFive.py
bombanya: Done

erin-le:
@bombanya: Hello, sorry for how long it's taken for the gem5 developers to get back to this PR. Are you still interested in working on it? If so, could you take a look at the Copilot comments, see if they make sense, and make changes if necessary? Thank you!

erin-le:
@bombanya: Hello, just another ping to ask about this PR. Could you let us know if you're still interested in working on this, and respond to the Copilot comments if you are? If you're not planning to work on this further, would you be okay with closing this PR? If this PR is closed, it can be still be reopened if you have more time to work on it later. Thanks!

PR #2088, opened 2025-03-17
tests: add tests for taking checkpoints with multisim

erin-le:
This PR adds two tests for taking checkpoints with MultiSim. 

- `test-multisim-hello-save-checkpoint` runs three simulations using the Arm, X86, and RISCV "hello world"
binaries and takes a checkpoint for each simulation after 1 million ticks.

- `test-multisim-riscv-hello-save-checkpoint` runs three simulations which use the "riscv-hello" binary and
takes checkpoints at 500,000, 1 million, and 1.5 million ticks.

Taking checkpoints using MultiSim is only possible when using hypercall exit events. If you try to take checkpoints with the "classic" style of exit events while using MultiSim, i.e. calling `simulator.save_checkpoint()` after `multisim.add_simulator(...)`, you will encounter the error  `TypeError: Checkpoint must be called on a root object.`


Harshil2107 (dismissed):
lgtm

Harshil2107 (approved):
lgtm

PR #2089, opened 2025-03-17
tests: add test for workloads of varying length in multisim

erin-le:
This PR adds a test that runs workloads of varying length using MultiSim. The workloads are Arm and X86 "hello world" binaries, which are added 6 times each, and RISCV boots with and without systemd.

The "hello world" binaries are added multiple times so that the test also checks if MultiSim works when the number of jobs exceeds the number of MultiSim processes.


tests/gem5/multisim/configs/varying-length-10-proc.py
```diff
@@ -0,0 +1,187 @@
...
+from gem5.components.processors.simple_processor import SimpleProcessor
+from gem5.isas import ISA
+from gem5.resources.resource import (
+    DiskImageResource,
+    KernelResource,
+    obtain_resource,
+)
+from gem5.simulate.simulator import Simulator
+
+NUM_PROCESSES = 10
```
erin-le, line 41: I also tested with 5, 9, and 100 processes

Harshil2107 (dismissed):
lgtm

Harshil2107 (dismissed):
lgtm

Harshil2107 (approved):

PR #2091, opened 2025-03-18
gpu-compute: GPU protocol tester bug

mysoreanoop:
Recent commits that added multiple RubySystems to gem5 caused a bug in the GPU protocol tester where the parameter needs to use floorLog2 of the cache line size to properly set the number of bits in a cache line for the tester.

### DIFF
--- a/src/cpu/testers/gpu_ruby_test/tester_thread.cc
+++ b/src/cpu/testers/gpu_ruby_test/tester_thread.cc
@@ -491,7 +491,7 @@ TesterThread::printAllOutstandingReqs(std::stringstream& ss) const
 std::string
 TesterThread::printAddress(Addr addr) const
 {
-    return ruby::printAddress(addr, cacheLineSize * 8);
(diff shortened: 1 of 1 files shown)


mattsinc:
@mysoreanoop : thanks for the contribution!  Can you please update your commit message to better explain what the problem is?  It *seems* like since gem5 uses 64B cache lines that both versions should be the same, but perhaps I'm missing something without context about what the issue is.

abmerop:
This is related to multiple RubySystem changes.  The parameter should be the number of bits in a cache line. It should be floorLog2(cacheLineSize).

mattsinc:
> This is related to multiple RubySystem changes. The parameter should be the number of bits in a cache line. It should be floorLog2(cacheLineSize).

Great, then this change is correct.  @mysoreanoop please update the commit message and then we can proceed

powerjg (approved):

mattsinc:
@mysoreanoop : I updated the commit message on GitHub too.  Going to merge.

PR #2103, opened 2025-03-21
stdlib: Add warning message for multisim processor switching

erin-le:
This commit adds an additional message to the assert that occurs when processor.switch() is used with multisim. When processor.switch() is used, only the last simulation to be run will have its processors switched correctly.


powerjg:
~~This seems like a deeper problem. I guess it's OK to at least give a warning... but it seems that if it's not possible to do switching for a particular experiment in multisim that there's something deeper that needs to be fixed.~~

Edit: Apologies, I didn't read the PR first 😆  This looks good to me.

powerjg (approved):

PR #2107, opened 2025-03-24
scons: add option --debug-fission

harry900831:
This CL introduces the `--debug-fission` flag, which allows the compiler to generate split DWARF debug information. This separates the main debug information into a `.dwo` file, reducing the size of the main object file.


giactra (approved):
Nice! We should probably document this on the gem5 website

harry900831:
Let's let others help on updating the gem5 website documentation with information about the --debug-fission flag and merge this first.

Thanks!

PR #2119, opened 2025-03-26
cpu-o3: Replace C++03 boilerplate with range-based for loops

bpfoley:
No functional change.

Removing all these explict std::list<T>::iterator decls will make it easier to change activeThreads to a denser and more cache friendly data structure such as std::array<T> should we want to too.

Change-Id: Ie365e876e387fdcd5eef8f8e3117812fb7056e55

### DIFF
--- a/src/cpu/o3/commit.cc
+++ b/src/cpu/o3/commit.cc
@@ -371,8 +371,8 @@ Commit::takeOverFrom()
 void
 Commit::deactivateThread(ThreadID tid)
 {
-    std::list<ThreadID>::iterator thread_it = std::find(priority_list.begin(),
-            priority_list.end(), tid);
+    auto thread_it = std::find(
+            priority_list.begin(), priority_list.end(), tid);
 
     if (thread_it != priority_list.end()) {
         priority_list.erase(thread_it);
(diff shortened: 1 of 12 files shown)


src/cpu/o3/commit.cc
```diff
@@ -371,11 +371,10 @@ Commit::takeOverFrom()
 void
 Commit::deactivateThread(ThreadID tid)
 {
-    std::list<ThreadID>::iterator thread_it = std::find(priority_list.begin(),
-            priority_list.end(), tid);
+    auto it = std::find(priority_list.begin(), priority_list.end(), tid);
```
giactra: I would keep thread_it which is more meaningful

src/cpu/o3/cpu.cc
```diff
@@ -450,13 +450,12 @@ CPU::startup()
 void
 CPU::activateThread(ThreadID tid)
 {
-    std::list<ThreadID>::iterator isActive =
-        std::find(activeThreads.begin(), activeThreads.end(), tid);
+    auto it = std::find(activeThreads.begin(), activeThreads.end(), tid);
```
giactra: I would call it active_it

src/cpu/o3/cpu.cc
```diff
@@ -471,16 +470,15 @@ CPU::deactivateThread(ThreadID tid)
     assert(!commit.executingHtmTransaction(tid));
 
     //Remove From Active List, if Active
-    std::list<ThreadID>::iterator thread_it =
-        std::find(activeThreads.begin(), activeThreads.end(), tid);
+    auto it = std::find(activeThreads.begin(), activeThreads.end(), tid);
```
giactra: Ditto

giactra (changes requested):
Thanks for this contribution! I have some minor comments regarding the new iterator/variable names. As we are changing them it would make sense we choose something more meaningful

bpfoley:
Thanks. Updated with all your suggestions.

giactra (approved):

PR #2120, opened 2025-03-27
arch-riscv: Fix SVNAPOT PPN encoding

ncppd:
Set the last 4 bits of the SVNAPOT PPN to 0000
instead of 1000 before inserting it into the TLB

### DIFF
--- a/src/arch/riscv/pagetable_walker.cc
+++ b/src/arch/riscv/pagetable_walker.cc
@@ -415,7 +415,8 @@ Walker::WalkerState::stepWalk(PacketPtr &write)
                         assert(!(pte.n) || level == 0);
                         entry.logBytes = (pte.n) ? PageShift + NapotShift :
                                             PageShift + (level * LEVEL_BITS);
-                        entry.paddr = pte.ppn;
+                        entry.paddr = (!pte.n) ? pte.ppn :
+                                        pte.ppn & ~((1UL << NapotShift) - 1);
                         entry.vaddr &= ~((1 << entry.logBytes) - 1);
                         entry.pte = pte;
                         // put it non-writable into the TLB to detect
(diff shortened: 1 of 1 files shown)


src/arch/riscv/pagetable_walker.cc
```diff
@@ -415,7 +415,8 @@ Walker::WalkerState::stepWalk(PacketPtr &write)
                         assert(!(pte.n) || level == 0);
                         entry.logBytes = (pte.n) ? PageShift + NapotShift :
                                             PageShift + (level * LEVEL_BITS);
-                        entry.paddr = pte.ppn;
+                        entry.paddr = (!pte.n) ? pte.ppn :
+                                        pte.ppn & ~((1UL << NapotShift) - 1);
```
powerjg: Would it be possible to use `bits()` here instead of shifting and masking? Using `bits` makes it easier to maintain in the long run.

powerjg (commented):
I assume this is the documentation from RISC-V on this change: https://riscv-software-src.github.io/riscv-unified-db/manual/html/isa/isa_20240411/exts/Svnapot.html

Let me know if I'm wrong.

src/arch/riscv/pagetable_walker.cc
ncppd: Yes you are correct, I will use `mask()` to make it easier to read and maintain.

ncppd:
In my previous PR, I was wrong to omit zeroing the last four PPN bits, which caused weird behavior in some workloads due to address calculation in the TLB, for example:
`Addr paddr = e->paddr << PageShift | (vaddr & mask(e->logBytes));` (line 347 in `tlb.cc`)
Even though `(vaddr & mask(e->logBytes))` will produce the correct 16bit mask, `e->paddr << PageShift` only shifts by 12bits and the end result may be an erroneous address if  `1000` are the LSBs . By zeroing the 4 last PPN bits this problem is resolved, I did not want to change the TLB logic so I decided this would be a cleaner solution.

powerjg (approved):

powerjg:
Thanks for the updates!

PR #2123, opened 2025-03-28
sim, stdlib: Add progress bar for simulation progress

erin-le:
This commit adds a progress bar to show how many instructions have executed compared to the total number of instructions for the workload. If -r is passed, the progress bar will be outputted to progress_bar.log in the m5out directory. Otherwise, it will be outputted to the terminal. This progress bar is only shown if simulation_run_results is part of the workload's JSON. This commit also converts an inform statement to a debug flag.


erin-le:
Closing in favor of a different implementation.

PR #2124, opened 2025-03-28
use mkdtemp for tmpfile in SerializationFixture

harry900831:
SerializationFixture has been used in:
* src/sim/globals.test.cc
* src/sim/serialize.test.cc
* src/base/loader/symtab.test.cc

The original implementation may cause those tests to operate on the same files.
Hence, use mkdtemp to prevent multiple gtest executable use the same
temporary directory name.

generateTempDirName() is remained for the test that needs non-exists
directory name.

### DIFF
--- a/src/base/gtest/serialization_fixture.hh
+++ b/src/base/gtest/serialization_fixture.hh
@@ -29,6 +29,7 @@
 #include <gtest/gtest.h>
 
 #include <cassert>
+#include <cstdlib>
 #include <cerrno>
 #include <fstream>
 #include <string>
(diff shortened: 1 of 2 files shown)


src/base/gtest/serialization_fixture.hh
```diff
@@ -98,9 +98,8 @@ class SerializationFixture : public ::testing::Test
     SetUp() override
     {
         // Create the directory
-        dirName = generateTempDirName();
-        [[maybe_unused]] int success = mkdir(dirName.c_str(), 0775);
-        assert(!(success == -1 && errno != EEXIST));
+        [[maybe_unused]] char* ret = mkdtemp(dirName.data());
```
harry900831, line 89: The reason for not use `dirName = /tmp/ + generateTempDirName()` here is because that it is possible that a file with the name returned by std::tmpnam is created by another process between the moment std::tmpnam returns and the moment this program attempts to use the returned name to create a file.

harry900831: I've removed the generateTempDirName() anyway so this is no longer an issue :)

giactra (approved):

PR #2126, opened 2025-03-29
sim-se: Add ArmLinux32 support for clone3 syscall

### LINKED ISSUE #2066: arch-arm: clone3 syscall fails on arm64
**Describe the bug**
ARM syscall emulation fails due to an unimplemented clone3 syscall (435) when running programs that use threading like the [thread test](https://github.com/gem5/gem5/blob/stable/tests/test-progs/threads/src/threads.cpp).

**Affects version**
Found in v24.1.0.2

**gem5 Modifications**
No modifications

**To Reproduce**
Compile a program using std::thread using newer glibc, mine was arm64 Ubuntu 24.04.

**Host ISA**
arm64

**Compiler used**
gcc version 13.3.0

bpfoley:
Fix for issue #2066

Change-Id: I578f9113d6a149beacb6adb46503aea6a55c05c7

### DIFF
--- a/src/arch/arm/linux/linux.hh
+++ b/src/arch/arm/linux/linux.hh
@@ -281,6 +281,22 @@ class ArmLinux32 : public ArmLinux, public OpenFlagTable<ArmLinux32>
         int32_t tms_cstime;     //!< system time of children
     };
 
+    // For clone(3)
+    struct tgt_clone_args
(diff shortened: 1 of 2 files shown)


giactra (approved):
Thanks for hooking this to arm!

PR #2128, opened 2025-03-31
scons,arch-arm: Fix clang 18 build issues

bpfoley:
Ubuntu 24.04 LTS ships with clang 18, and Ubuntu 24.10 ships with clang 19.

Clang 18 switches on some new `-W` warnings by default which cause gem5 build failures thanks to `-Werror`.

This patchset and #2127 fixes these.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -706,6 +706,8 @@ for variant_path in variant_paths:
         with gem5_scons.Configure(env) as conf:
             conf.CheckCxxFlag('-Wno-c99-designator')
             conf.CheckCxxFlag('-Wno-defaulted-function-deleted')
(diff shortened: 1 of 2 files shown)


powerjg (approved):

powerjg:
Out of curiosity... how many VLA warnings do we have? It wouldn't be bad to go through and fix them at some point if it's manageable.

giactra:
Nit, I would add the scons,arch-arm tag to the PR title

bpfoley:
> Out of curiosity... how many VLA warnings do we have? It wouldn't be bad to go through and fix them at some point if it's manageable.

There's ~10-20 of them. It would definitely be worth fixing them, but I didn't want to get too distracted. I'll add it to the todo list...

PR #499, opened 2023-10-23
cpu: BPU support for decoupled front-end

dhschall:
To support the decoupled front-end the BPU must have the ability to add a placeholder for branches that are not in
the BTB (surprise branches). In this case the front-end cannot detect the branch and make a prediction. However,
to rollback in case of a misprediction the BPU needs a history element. This function allows generation of such elements to add it to the the speculative history buffer. It does NOT make a prediction or alter the BPU state.

PR #2129, opened 2025-03-31
cpu: Fix Clang compilation error for TAGE-SC-L

rjc-arch:
PR #499 introduced a shadowing of virtual method
`TAGEBase::updatePathAndGlobalHistory` in `TAGE_SC_L_TAGE` that should have been an overload. This shadow method works in the context that it is currently used, but causes a warning-as-error in Clang, which is more proactive about warning of potentially unwanted shadowing.

This patch matches the prototype to overload the existing virtual method, avoiding the error in Clang and fixing the build.

Change-Id: If4296f9b6d6154066567a58dc9cbdcd4fc44f425

### DIFF
--- a/src/cpu/pred/tage_sc_l.cc
+++ b/src/cpu/pred/tage_sc_l.cc
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2022-2023 The University of Edinburgh
+ * Copyright (c) 2025 Arm Limited
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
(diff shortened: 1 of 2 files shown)


bpfoley:
This and #2127 are duplicates

rjc-arch:
@bpfoley Looks like you got there first!

bpfoley:
> @bpfoley Looks like you got there first!

Note that in #2127 I added `bi->modified = true` to `TAGE_SC_L_TAGE::updatePathAndGlobalHistory`. However I don't actually know this code, and I just did it by pattern matching. Do you know if this is the correct thing to do?

BobbyRBruce (approved):
Thanks for the fix. I came up with the same solution locally before realizing this was already pushed. it works

Edit: Woops, appears even this is a duplicate. Is this PR ok to close?

rjc-arch:
@bpfoley, I did run my proposal past @dhschall before submitting. @dhschall, could you take a quick look at #2127 to check that the addition of `bi->modified = true` to `TAGE_SC_L_TAGE::updatePathAndGlobalHistory` is correct?

@BobbyRBruce, yes it looks like Brian and I submitted the same fix. I'm happy to defer to Brian's fix and close this PR.

erin-le:
@rjc-arch Hello, since PR 2127 has been merged, I'm closing this PR as a duplicate. If you want to continue working on this PR, however, feel free to reopen it. Thank you for your contribution!

PR #2131, opened 2025-03-31
mem: Make PortTerminator create resp_ports as well as req_ports

bpfoley:
The bug in the ctor caused fatal errors if a Python config tried to use pt.resp_ports.connect(...).

Change-Id: Ie38ba553a04afce23095f790193a82f1784370f0

### DIFF
--- a/src/mem/port_terminator.hh
+++ b/src/mem/port_terminator.hh
@@ -100,6 +100,31 @@ class PortTerminator : public SimObject
         RespPort(const std::string &name):
             ResponsePort(name)
         {}
+
+        Tick recvAtomic(PacketPtr) override
+        {
+            panic("PortTerminator recvAtomic should never be called");
+        }
+
+        bool recvTimingReq(PacketPtr) override
+        {
+            panic("PortTerminator recvTimingReq should never be called");
+        }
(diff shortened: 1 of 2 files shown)


src/mem/port_terminator.hh
```diff
@@ -100,6 +100,22 @@ class PortTerminator : public SimObject
...
+        }
+        void recvRespRetry() override {
+          panic("PortTerminator recvRespRetry should never be called");
+        }
+        void recvFunctional(PacketPtr) override {
+          panic("PortTerminator recvFunctional should never be called");
+        }
+        AddrRangeList getAddrRanges() const override {
+          panic("PortTerminator getAddrRanges should never be called");
+        }
```
giactra: Could you align these with ReqPort coding style for consistency?
(braces on newline and 4 whitespaces indentation)

giactra (changes requested):
Thanks, I only have a coding style comment

src/mem/port_terminator.hh
bpfoley: Done.

giactra (approved):

PR #2139, opened 2025-04-02
 mem-ruby: Make SLICC enum FooType_to_string handle the _NUM case

bpfoley:
FooType_NUM is used as a sentinel in a lot of Ruby code, so it's
    useful to be able to see a string representation of it rather than
    having to special case printing of FooType_NUM in log messages.

    While here, make the panics for out-of-range enum values more
    informative.

    Change-Id: I6633183a384ad52551a1239601560d5e541732c6


giactra:
This is reasonable, however I believe a better solution would be to add the _NUM case in MachineType_to_string and return "invalid" instead...

bpfoley:
> This is reasonable, however I believe a better solution would be to add the _NUM case in MachineType_to_string and return "invalid" instead...

Good point. I like your solution better.

powerjg (approved):

PR #2143, opened 2025-04-04
arch-riscv: Fix atomic ops on big endian hosts

### LINKED ISSUE #2148: Potential bug on big endian hosts for "far" atomics
The current implementation atomic functors does not convert the guest memory data into host endianness before operating on the data.

See #2143 for a fix for RISC-V.

There may be issues in other places. Here are some potential links:

### In Arm code

I'm hesitant to change this code with all of the templating.

* https://github.com/gem5/gem5/blob/develop/src/arch/arm/isa/insts/amo64.isa#L208
* https://github.com/gem5/gem5/blob/develop/src/arch/arm/isa/insts/amo64.isa#L329
* https://github.com/gem5/gem5/blob/develop/src/arch/arm/isa/insts/amo64.isa#L348
* https://github.com/gem5/gem5/blob/develop/src/arch/arm/isa/insts/amo64.isa#L409

### In GPU code

The GPU uses specialized classes for each operation. I think I can update them correctly. I'll push a patch soon.

See https://github.com/gem5/gem5/blob/develop/src/base/amo.hh#L129

See https://github.com/gem5/gem5/pull/2149

powerjg:
RISC-V atomic operations pass a functor which operates on the data. Previously, this functor ignored the endianness of the host and just directly operated on the data. This change ensures that when reading data from the guest memory into host variables that we set the endianness correctly.

A small number of these operations were tested on a BE host by ensuring the execution trace was the same on both the BE host and a LE host.

### DIFF
--- a/src/arch/riscv/isa/decoder.isa
+++ b/src/arch/riscv/isa/decoder.isa
@@ -1636,7 +1636,11 @@ decode QUADRANT default Unknown::unknown() {
                 }}, {{
                     TypedAtomicOpFunctor<int32_t> *amo_op =
                           new AtomicGenericOp<int32_t>(Rs2_sw,
-                                  [](int32_t* b, int32_t a){ *b += a; });
+                                  [](int32_t* b, int32_t a){
+                                      int32_t mem_data = letoh(*b);
+                                      int32_t new_data = mem_data + a;
+                                      *b = htole(new_data);
+                                  });
                 }}, mem_flags=ATOMIC_RETURN_OP);
                 0x1: AtomicMemOp::amoswap_w({{
                     Rd_sd = Mem_sw;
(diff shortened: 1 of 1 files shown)


abmerop (approved):
This looks like it was "fun" to track down.  Any ideas if it applies to the functors in src/base/amo.hh? I think GPU and ARM code share the implementations there.

powerjg:
> This looks like it was "fun" to track down.  Any ideas if it applies to the functors in src/base/amo.hh? I think GPU and ARM code share the implementations there.

Probably. I'll create an issue for it tomorrow. 

Clearly not many gem5 users are in big endian systems. But for those one or two...

mattsinc:
> > This looks like it was "fun" to track down.  Any ideas if it applies to the functors in src/base/amo.hh? I think GPU and ARM code share the implementations there.
> 
> Probably. I'll create an issue for it tomorrow.
> 
> Clearly not many gem5 users are in big endian systems. But for those one or two...

I agree I've never checked endian for GPU atomics, and until this patch hadn't been worried about it :)

PR #2147, opened 2025-04-04
cpu: Free memory allocated by indirect predictor lookup

### LINKED ISSUE #2058: Memory consumption gradually increases over time when running RISC-V FS with the O3 model
I am benchmarking gem5 using the RISC-V O3 model and have observed a significant increase in memory usage during benchmark execution. This memory growth eventually crashes the server when multiple instances or long-running benchmarks are run. A long running simulation with 16GB of DRAM may end up consuming more than 60GB of physical memory. I have not observed such behavior using ATOMIC/TIMING models.

I'm not sure if this behavior is intentional (perhaps for checkpointing purposes?) -- I discovered that the memory increase is caused by a `clone()` function call from the `SimpleIndirectPredictor`, as revealed by the Massif profiler. I ran the profiler for about two days, and the memory usage showed a clearly linear increase (the first 1/3 of the simulation is run using the ATOMIC model and then switched to O3). The workload under test is `605.mcf` from the spec2017 benchmark suite.

I have attached the profiler output for anyone interested in reviewing it.

[massif.log](https://github.com/user-attachments/files/19090364/massif.log)

ncppd:
A memory leak occurred during indirect branch prediction lookups. When `SimpleIndirectPredictor::lookup` resulted in a hit, it would allocate a new `PCStateBase` object via `PCState::clone()` (using `new`). The pointer to this object (`itarget`) was returned to the caller.

However, the caller used the object's data but did not `delete` the allocated memory before the `itarget` pointer went out of scope. This led to memory accumulating on each predictor hit.

This commit fixes the leak by adding `delete itarget;` after the object pointed to by `itarget` is no longer needed within the relevant code block.

This PR impacts long-running MINOR/O3 simulations and significantly lowers memory usage.

This resolves #2058, MINOR/O3 memory consumption is now pretty close to ATOMIC/TIMING models and does not creep up with time.

### DIFF
--- a/src/cpu/pred/bpred_unit.cc
+++ b/src/cpu/pred/bpred_unit.cc
@@ -271,7 +271,7 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
             ++stats.indirectHits;
             hist->targetProvider = TargetProvider::Indirect;
             set(hist->target, *itarget);
-
+            delete itarget;
             DPRINTF(Branch,
                     "[tid:%i, sn:%llu] Instruction %s predicted "
                     "indirect target is %s\n",


powerjg (commented):
Looks like this was tough to track down... good job!

I have a couple of questions.

The "new" code in question is `pcstate.hh:228`

```
inline void
set(PCStateBase *&dest, const PCStateBase &src)
{
    if (GEM5_LIKELY(dest)) {
        // Update dest with the contents of src.
        dest->update(src);
    } else {
        // Clone src over to dest.
        dest = src.clone();
    }
}
```

This is called in `simple_indirect.cc:162`

```
            set(target, *way->target);
```

What happens in your new code if `itarget` is `way->target`?

TBH, it seems that this code really should be using  `unique_ptr` or  `shared_ptr`. That said, if you're confident in the fix, I can take your work for it :)

ncppd:
From my understanding `itarget` has been already cloned in `set(hist->target, *itarget);` in `bpred_unit.cc: 273`, so it is not lost. 

What confuses me is the fact that in `bpred_unit.cc: 186` --> `const PCStateBase * btb_target = btb->lookup` the `btb_target` is also not deleted after `hit->target` is updated, but does not show any heap blowups in massif. 
Also `std::unique_ptr<PCStateBase> target;` in `bpred_unit.hh` is correctly setup as a unique pointer which means that is should free up memory if it goes out-of-scope.

I will try to find some time to further dig into this and get back to you.

dhschall:
Hi @ncppd, thanks a lot for digging into that!
The whole PCStateBase is quite confusing at some point, and when I refactored some of the code, I was also not always sure what the correct way should be.
I think the main difference between the two "lookup" methods is that in the BTB, we use.
```
return entry->target.get()
```
and in the indirect predictor we use
```
set(target, *way->target);
```

The BTB method returns a pointer to the target in the BTB entry, whereas for the indirect predictor we create a pointer first
`PCStateBase* target = nullptr;` and then **clone** the target.

@powerjg I think @ncppd's code should work because we pass a null pointer to the lookup function, so it should always copy.

That said, I think it's **not** the right way to fix it, and we should make both methods consistent. I personally think it should be done using the BTB method, returning a pointer to the target PC. This is more intuitive, creates less allocation/deletion, thus more performant, and in my understanding, the same approach used in other parts of gem5 (e.g., caches). 
On the other hand, the `.get` method is a bit weird because it creates a raw pointer to a unique_ptr object, so it kind of defeats its purpose.

@powerjg, any comment on what should be the right approach? Is having `std::unique_ptr<PCStateBase>` for the target is the right way in the first place? During my refactoring, I just copied that pre-existing code.

dhschall:
I just saw there was a fix two days ago. https://github.com/gem5/gem5/commit/bfb6c54e45de98deb02f78336ad6422640e2cf6d
It fixes the leak but still does not fix the underlying problem

ncppd:
#2220 also solves #2058 btw, I think using a unique pointer looks better than calling `delete` on a raw pointer.

powerjg:
> Is having std::unique_ptr<PCStateBase> for the target is the right way in the first place?

I *think* that would be the best approach. If what we're doing is essentially passing around a reference, then passing it as a `unique_ptr` with `std::move` is likely the best approach. This way, we don't have to worry about calling `delete` at the right time.

dhschall:
Yes, I agree that this PR and #2058 are fixed and can be closed.

Though I'm also a bit confused now, but maybe my explanation was also not good.

The problem is that both the BTB and indirect predictor hold the target PC as `std::unique_ptr<PCStateBase>` in their entries. If we do a lookup we return a reference/pointer to the `PCStateBase`. We **don't** want to move it otherwise it would delete the `PCStateBase` in the BTB/Indirect entry.

The way it is done now works but it is inconsistent. Btw. the RAS uses the same method as the BTB. I was wondering if we (me ;)) should make it consistent and what is the preferred way.

PR #2149, opened 2025-04-04
base: Update AMO code to be little endian specific

### LINKED ISSUE #2148: Potential bug on big endian hosts for "far" atomics

powerjg:
Make the generic AMOs use endian switching so that they will work if the host endianness is BE.

Note: This change should only affect the GPU code, so it assumes the guest is little endian.

### DIFF
--- a/src/base/amo.hh
+++ b/src/base/amo.hh
@@ -126,13 +126,19 @@ class AtomicGenericPair3Op : public TypedAtomicOpFunctor<T>
     std::function<void(T*, std::array<T, 2>&, std::array<T, 2>)> op;
 };
 
+// IMPORTANT: The following operations assume the guest is little-endian.
+
 template<typename T>
 class AtomicOpAnd : public TypedAtomicOpFunctor<T>
 {
     // Bitwise operations are only legal on integral types
     template<typename B>
     typename std::enable_if<std::is_integral<B>::value, void>::type
-    executeImpl(B *b) { *b &= a; }
+    executeImpl(B *b) {
+        B mem_data = letoh(*b);
+        B new_data = mem_data & a;
+        *b = htole(new_data);
+    }
 
     template<typename B>
     typename std::enable_if<!std::is_integral<B>::value, void>::type
@@ -151,7 +157,11 @@ class AtomicOpOr : public TypedAtomicOpFunctor<T>
     // Bitwise operations are only legal on integral types
     template<typename B>
     typename std::enable_if<std::is_integral<B>::value, void>::type
-    executeImpl(B *b) { *b |= a; }
+    executeImpl(B *b) {
+        B mem_data = letoh(*b);
+        B new_data = mem_data | a;
+        *b = htole(new_data);
+    }
 
     template<typename B>
     typename std::enable_if<!std::is_integral<B>::value, void>::type
@@ -170,7 +180,11 @@ class AtomicOpXor : public TypedAtomicOpFunctor<T>
     // Bitwise operations are only legal on integral types
     template<typename B>
     typename std::enable_if<std::is_integral<B>::value, void>::type
-    executeImpl(B *b) { *b ^= a; }
+    executeImpl(B *b) {
+        B mem_data = letoh(*b);
+        B new_data = mem_data ^ a;
+        *b = htole(new_data);
+    }
 
     template<typename B>
     typename std::enable_if<!std::is_integral<B>::value, void>::type
@@ -199,7 +213,11 @@ class AtomicOpAdd : public TypedAtomicOpFunctor<T>
   public:
     T a;
     AtomicOpAdd(T _a) : a(_a) { }
-    void execute(T *b) { *b += a; }
+    void execute(T *b) {
+        T mem_data = letoh(*b);
+        T new_data = mem_data + a;
+        *b = htole(new_data);
+    }
     AtomicOpFunctor* clone () { return new AtomicOpAdd(a); }
 };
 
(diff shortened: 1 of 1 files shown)


powerjg:
After making the change I'm less confident... I don't really think we need both BE and LE versions of this code since it's only used by the GPU.

Let me know what you think.

src/base/amo.hh
```diff
@@ -126,13 +126,19 @@ class AtomicGenericPair3Op : public TypedAtomicOpFunctor<T>
     std::function<void(T*, std::array<T, 2>&, std::array<T, 2>)> op;
 };
 
+// IMPORTANT: The following operations assume the guest is little-endian.
```
mattsinc, line 129: even if we decide to drop everything else, we should probably keep this

src/base/amo.hh
```diff
@@ -126,13 +126,19 @@ class AtomicGenericPair3Op : public TypedAtomicOpFunctor<T>
...
+
 template<typename T>
 class AtomicOpAnd : public TypedAtomicOpFunctor<T>
 {
     // Bitwise operations are only legal on integral types
     template<typename B>
     typename std::enable_if<std::is_integral<B>::value, void>::type
-    executeImpl(B *b) { *b &= a; }
+    executeImpl(B *b) {
+        B mem_data = letoh(*b);
```
mattsinc, line 138: maybe I'm missing something obvious, but since a given ISA is (in theory) using the same endianness, wouldn't we want letoh or htole but not both?

mattsinc:
> After making the change I'm less confident... I don't really think we need both BE and LE versions of this code since it's only used by the GPU.
> 
> Let me know what you think.

I do agree if we only use it for GPU, putting both in is probably not necessary.  But we should put some sort of assert or comment, etc. to define the expectations.

giactra:
Endianness always confuses me: isn't it sufficient to adjust the endianness of the instruction operand?
For example in AtomicOpAnd:

B = (B & A)

As B is in memory, assuming it is stored to match the ISA (guest) endianness, shouldn't we just convert A with htog before performing the bitwise operation?
One of the reasons why I am asking this is that I don't think it is clean to encode guest/host endianness in certain atomic operations which should be agnostic to this issue...

powerjg:
> Endianness always confuses me: isn't it sufficient to adjust the endianness of the instruction operand? For example in AtomicOpAnd:
> 
> B = (B & A)
> 
> As B is in memory, assuming it is stored to match the ISA (guest) endianness, shouldn't we just convert A with htog before performing the bitwise operation? One of the reasons why I am asking this is that I don't think it is clean to encode guest/host endianness in certain atomic operations which should be agnostic to this issue...

I agree it's confusing! It took me a while to figure out #2143. For that patch, I was able to test it on a BE system, so I'm pretty confident that fix correct.

To answer your question... the way I reasoned about it is that we have to (1) read from the guest memory in guest endianness into a host register in host endianness. Then, (2) we do the computation in host endianness. Finally, (3) we need to write the result from the host register (in host endianness) into the guest memory (in guest endianness). So, we need to convert both when we read and when we write.

There are likely some operations that are the same whether you convert the endianness or not (e.g., bitwise operations?). But it seems safest just to do it everywhere.

Finally, I am not completely happy with this change. This AMO code is supposed to be "generic," but it's only used for the GPU. I also don't like encoding endianness in operations that should be agnostic, but we *have* to convert from guest to host (and back) if the guest and host endianness doesn't match.

It may be a better change to create an both LE and BE versions of all of this code so that it can be used with LE and BE guests. I hesitate to do that since no BE guest uses this code, though.

src/base/amo.hh
powerjg: We first have to read the data from memory (which is in *guest* endianness, LE specifically in the case of the GPU) and convert it to the host endianness. Then, the data is loaded in the host register, and the host does the operation. After the host has the answer (in host endianness), we then need to write the data back in to guest memory *in guest endianness*. So, there's another conversion from host to LE (because we're assuming the guest is LE)

src/base/amo.hh
powerjg: If we leave this (and drop the other changes), then the comment should read `IMPORTANT: The following code assumes the guest and host have the same endianness. If this code is used for a BE guest or if the host is BE then there may be difficult to track down bugs in the execution!`

powerjg:
@mattsinc and @giactra any other feedback here?

I'm thinking of simply dropping this change. It's not a good solution, and the "problem" doesn't *really* exist.

If I drop this, then the GPU model *will not work on a big-endian host*. 

The main downside of merging this as-is is that the "base" atomic implementations (currently and as modified by this PR) *only support little-endian guests.* Although, they are also only used for the GPU which is little-endian.

Please let me know your thoughts. I would be happy to drop this and simply keep it around as "documentation" of the issue.

powerjg:
Closing since people don't seem to see the need for this.

mattsinc:
> @mattsinc and @giactra any other feedback here?
> 
> I'm thinking of simply dropping this change. It's not a good solution, and the "problem" doesn't _really_ exist.
> 
> If I drop this, then the GPU model _will not work on a big-endian host_.
> 
> The main downside of merging this as-is is that the "base" atomic implementations (currently and as modified by this PR) _only support little-endian guests._ Although, they are also only used for the GPU which is little-endian.
> 
> Please let me know your thoughts. I would be happy to drop this and simply keep it around as "documentation" of the issue.

sorry somehow I missed this ... I think dropping is ok

PR #2155, opened 2025-04-07
ext,tests: Fix weekly DRAMSys tests failing

derchr:
Due to recent changes, the default option for using FetchContent was not
    properly set in the default DRAMSys build configuration. Because of that
    the weekly tests failed as the build environment does not provide all
    dependencies.
    This fix enables FetchContent in DRAMSys to satisfy the internal
    dependencies using CMake.


powerjg (approved):
Thank you!

PR #2156, opened 2025-04-07
stdlib: Add a simulator API to set the exit event handler

pranith:
This API is useful in situations where we need the simulator object to construct the exit event handler. One such situation is with multisim where we need to call the appropriate exit handler for each checkpoint run.

### DIFF
--- a/src/python/gem5/simulate/simulator.py
+++ b/src/python/gem5/simulate/simulator.py
@@ -231,7 +231,7 @@ class Simulator:
     def set_id(self, id: str) -> None:
         """Set the ID of the simulator.
 
-        As, in the caae of multisim, this ID will be used to create an
+        As, in the case of multisim, this ID will be used to create an
         output subdirectory, there needs to be rules on what an ID can be.
         For now, this function encoures that IDs can only be alphanumeric
         characters with underscores  and dashes. Uunderscores and dashes cannot
@@ -272,6 +272,26 @@ class Simulator:
             return self._id
 
         return None
+
+    def set_on_exit_handler(
+        on_exit_event: Dict[
+            ExitEvent,
+            Union[
+                Generator[Optional[bool], None, None],
+                List[Callable],
+                Callable,
+            ],
+        ],
+        expected_execution_order: Optional[List[ExitEvent]] = None,
+    ) -> None:
+        """Set the on_exit_event handler for the simulator.
+
+        Set the exit event handler that specifies what to execute on each exit
+        event."""
+
+        ClassicGeneratorExitHandler.set_exit_event_map(
+            on_exit_event, expected_execution_order
+        )
 
     def set_max_ticks(self, max_tick: int) -> None:
         """Set the absolute (not relative) maximum number of ticks to run the


erin-le (changes requested):
@pranith Hello, thank you for your contribution. This PR has some CI test failures, which will have to be fixed before the PR can be merged. The specific error message for a few of the tests is as follows:
```
AttributeError: type object 'ClassicGeneratorExitHandler' has no attribute 'set_on_exit_handler'
```
I was able to reproduce the error by attempting to run `configs/example/gem5_library/x86-ubuntu-run-with-kvm-no-perf.py`. It looks like the issue is that `set_on_exit_handler` isn't accessible from where it is called in `exit_handler.py`.

erin-le:
Additionally, this change may overlap in functionality with hypercalls, which are a new way to set exit handlers in gem5. There isn't much documentation on them yet, but there is an example in the top comment of https://github.com/gem5/gem5/pull/1995. It may also be helpful to look at https://github.com/gem5/gem5/pull/2088, which has an example config script for taking checkpoints using multisim and the new hypercall handlers. Does this help to solve any of the problems you've encountered?

pranith:
I fixed the issue @erin-le , please review.

BobbyRBruce (commented):
My understanding is this is for setting the exit event handlers after the Simulator construction (where we currently allow the setting of the exit_event handlers). Why would we want to do this?

I realize you stated MultiSim but can you give me an example script just so i can get my head around the usecase a bit?

pranith:
@BobbyRBruce please find an example use in multisim script here: https://github.com/pranith/gem5/blob/e58b3594a91ea570622f8ab02fed0c4f209d7e6b/configs/example/gem5_library/checkpoints/simpoints-se-restore_arm_specrate2017.py#L331

BobbyRBruce:
> @BobbyRBruce please find an example use in multisim script here: https://github.com/pranith/gem5/blob/e58b3594a91ea570622f8ab02fed0c4f209d7e6b/configs/example/gem5_library/checkpoints/simpoints-se-restore_arm_specrate2017.py#L331

I'm not sure I understand. Doesn't the following work for you?

```py

board.set_se_simpoint_workload(
    workload_resource,
    arguments=argv[1:],
    simpoint=SimpointResource(
        simpoint_interval=200000000,
        # simpoint_interval=20000,
        simpoint_list=simpts_list,
        weight_list=weights_list,
        warmup_interval=50000000,
        # warmup_interval=5000,
    ),
    checkpoint=CheckpointResource(local_path=chkpt),
)


def simpoint_warmup_end(self):
    print(
        "\n\n\n################## end of warmup, starting to simulate SimPoint"
    )
    simulator.schedule_max_insts(
        board.get_simpoint().get_simpoint_interval()
    )
    # dump()
    reset()

    return False

def simpoint_interval_end(self):
    print("\n\n\n################## end of SimPoint interval\n\n\n")
    dump()

    return True


simulator = Simulator (self, board=board, on_exit_event = {
            ExitEvent.MAX_INSTS: [simpoint_warmup_end ,simpoint_interval_end]
    })

```

erin-le:
@pranith: Hello, could you respond to the previous comment when you have time? Thank you!

erin-le:
@pranith: Hello, this PR is being closed for inactivity. If you would like to continue working on it, please feel free to reopen. Thank you!

pranith:
@BobbyRBruce I tried the method you mentioned, and it is setting the interval end callback only on the last simulator constructed in the multisim flow. All the other simulators continue running as we do not set the interval end based on instruction count. Can you please check?

pranith:
@erin-le Can you please reopen this PR? I am unable to do it.

PR #2157, opened 2025-04-08
mem-ruby: Fix Ruby MemCtrl functionalRead

hnpl:
The makeLineAddress() function takes the number of bits to offset the block as a parameter, not the size of the block.


hnpl:
CC: @abmerop

powerjg (approved):

Issue #2159, opened 2025-04-08
Ruby protocol MOESI_CMP_token broken after #1453


powerjg:
#1453 made a change where the `RubySystem` pointer has to be explicitly passed around. This does not happen for the `PersistentTable` object that is used in MOESI_CMP_token.

You can reproduce this issue with the following: (after applying #2160)

```sh
 build/ALL/gem5.opt configs/example/ruby_random_test.py -n 4 --protocol=MOESI_CMP_token
```

The expected behavior is that the test runs to completion. The actual result is an assert failure.

The error is that the `NetDest` in the `PersistentTableEntry` is not initialized with a size.
https://github.com/gem5/gem5/blob/develop/src/mem/ruby/structures/PersistentTable.hh#L52

Unfortunately, with a quick skim I can't find a simple fix. We may want to simply delete this protocol.

I don't want to call this a "bug" per-se. It *is* a bug, but it's in code that we have not been supporting for probably decades.

powerjg:
@abmerop do you happen to see an easy way to fix this? If not, I advocate for dropping this protocol.

Here's the code that needs to change: https://github.com/gem5/gem5/blob/develop/src/mem/ruby/structures/PersistentTable.hh#L52

abmerop:
I thought I had handled this case. I think something like this would work:

1) Add `"PersistentTable"` to the list here https://github.com/gem5/gem5/blob/develop/src/mem/slicc/symbols/StateMachine.py#L836 (or below if it is a _ptr)
2) Define `setRubySystem` in `PersistentTable`
3) Call `setRubySystem` for each `NetDest` in `PersistentTableEntry` when it is created.

Thinking about it more, I believe I couldn't figure out why `PersistentTableEntry` is `static`: https://github.com/gem5/gem5/blob/develop/src/mem/ruby/structures/PersistentTable.cc#L52 and the protocol looked unused so I may not have finished a commit for it.

powerjg:
That will work for `PersistentTable`, but since the entries are in a map it's going to take a bit deeper change to fix. We'll have to `emplace` or provide a default implementation for the creator for the map as well.

This is low priority, I think.

PR #2160, opened 2025-04-08
configs,mem-ruby: Update ruby configs for ALL target

powerjg:
Use the full names for all of the controllers in the ruby scripts.
Now the `ruby_random_test.py` script will work with all of the ruby
protocols except for CHI.

### DIFF
--- a/configs/ruby/CHI_config.py
+++ b/configs/ruby/CHI_config.py
@@ -368,7 +368,7 @@ class CHI_MNController(CHI_MiscNode_Controller):
         self, ruby_system, addr_range, l1d_caches, early_nonsync_comp
     ):
         super().__init__(
-            version=Versions.getVersion(MiscNode_Controller),
+            version=Versions.getVersion(CHI_MiscNode_Controller),
             ruby_system=ruby_system,
             mandatoryQueue=MessageBuffer(),
             triggerQueue=TriggerMessageBuffer(),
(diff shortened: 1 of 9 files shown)


powerjg:
Note: I found #2159 with this change.

BobbyRBruce (approved):

giactra:
Hi Jason, is there a reason why it does not work with CHI?

powerjg:
Here's the error

```
> build/ALL/gem5.opt configs/example/ruby_random_test.py -n 2 --protocol=CHI
...
src/mem/ruby/slicc_interface/AbstractController.cc:453: fatal: system.ruby.network.ext_links1.ext_node: couldn't find mapping for address 3c0 mtype=NUM (invalid)
```

The problem is that CHI expects that every "CPU" has two ports one for instruction and one for data. This is deeply incompatible with the way the random tester works. See https://github.com/gem5/gem5/blob/develop/configs/example/ruby_random_test.py#L118

The random tester is just one object with vector ports. It assumes there is exactly one port per CPU.

I tried to mess around with the CHI config files and added something like "if CPU is type(RubyTester)"... but I couldn't get it to work in 15-30 minutes. It's not at all straightforward.

Just to be clear, what doesn't work is using the `--protocol=CHI` parameter when using the `ruby_random_test.py` script. The underlying problem is the mess that is `Ruby.py`.

PR #2161, opened 2025-04-09
util: Add feature to update debug flags in external hypercall util

Harshil2107:
**Add utility to update debug flags during simulation**

This PR introduces a utility that enables users to dynamically update debug flags during a running simulation using the `hypercall_external_signal` utility.

The utility is executed via `orchestrator-request.py` located in `util/hypercall_external_signal`. To use it, run the following command:

```
python3 ./orchestrator-request.py update_debug_flags --debug-flags-to-update=Flag1,Flag2,...
```

If you want to disable an already enabled debug flag, simply prepend the flag with a `-`. For example, passing `-Exec` will disable the `Exec` debug flag.

### Example Usage

1. **Start a simulation:**
   ```bash
   ./build/ALL/gem5.opt configs/example/gem5_library/x86-ubuntu-run.py
   ```

2. **To enable the `Exec` debug flag mid-simulation:**
   ```bash
   python3 ./orchestrator-request.py update_debug_flags --debug-flags-to-update=Exec
   ```

3. **To disable the `Exec` flag and enable the `Fetch` debug flag:**
   ```bash
   python3 ./orchestrator-request.py update_debug_flags --debug-flags-to-update=-Exec,Fetch
   ```

This utility makes it easy to modify debug flags without the need to restart the simulation.

### DIFF
--- a/src/python/gem5/simulate/exit_handler.py
+++ b/src/python/gem5/simulate/exit_handler.py
@@ -392,12 +392,40 @@ class OrchestratorExitHandler(ExitHandler, hypercall_num=1000):
             "instruction_count": simulator.get_instruction_count(),
         }
 
+    def _add_debug_flags(self, debug_flags: [List]) -> Dict[str, str]:
+        from m5 import debug
+
+        flags_disabled = []
+        flags_enabled = []
+        invalid_flags = []
+        for flag in debug_flags:
+            off = False
+            if flag.startswith("-"):
+                flag = flag[1:]
+                off = True
+
+            if flag not in debug.flags:
+                invalid_flags.append(flag)
+                continue
+
+            if off:
+                flags_disabled.append(flag)
+                debug.flags[flag].disable()
+            else:
+                flags_enabled.append(flag)
+                debug.flags[flag].enable()
+        return {
+            "flags_enabled": str(flags_enabled),
+            "flags_disabled": str(flags_disabled),
+            "invalid_flags": str(invalid_flags),
+        }
+
     @overrides(ExitHandler)
     def _process(self, simulator: "Simulator") -> None:
         try:
             socket_path = self._payload.get("response_socket")
             function = self._payload.get("function", "status")
-
+            debug_flags_to_add = self._payload.get("debug_flags_to_add")
             if socket_path:
                 sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
                 sock.connect(socket_path)
@@ -407,6 +435,9 @@ class OrchestratorExitHandler(ExitHandler, hypercall_num=1000):
                 elif function == "get_stats":
                     stats = simulator.get_stats()
                     response = json.dumps(stats)
+                elif function == "add_debug_flags":
+                    debug_flags = debug_flags_to_add.split(",")
+                    response = json.dumps(self._add_debug_flags(debug_flags))
                 else:
                     response = json.dumps(
                         {"error": f"Unknown function: {function}"}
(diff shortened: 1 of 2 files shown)


src/python/gem5/simulate/exit_handler.py
```diff
@@ -392,12 +392,40 @@ def _get_status(self, simulator: "Simulator") -> Dict[str, str]:
             "instruction_count": simulator.get_instruction_count(),
         }
 
+    def _add_debug_flags(self, debug_flags: [List]) -> Dict[str, str]:
```
powerjg: I think you mean

```suggestion
    def _add_debug_flags(self, debug_flags: List[str]) -> Dict[str, str]:
```

util/hypercall_external_signal/orchestrator-request.py
```diff
@@ -171,7 +171,9 @@ def receive_full_message(conn: socket.socket, timeout: float = 30.0) -> str:
     return b"".join(chunks).decode()
 
 
-def send_and_receive_hypercall(pid: int, function: str) -> str:
+def send_and_receive_hypercall(
+    pid: int, function: str, debug_flags_to_add: str = None
```
powerjg: I don't think it makes sense to have an argument specific to one function in the generic function. This can be solved by either (1) making it a generic "arguments" or (2) making a new function to send/recv "add_debug_flags".

Also, why is it called "add_debug_flags" if you can both enable and disable flags? Maybe it should be "update_debug_flags" or something like that.

util/hypercall_external_signal/orchestrator-request.py
```diff
@@ -180,8 +182,10 @@ def send_and_receive_hypercall(pid: int, function: str) -> str:
 
     :param pid: Process ID of gem5 instance
     :type pid: int
-    :param function: Hypercall function to execute ('status'|'get_stats')
+    :param function: Hypercall function to execute ('status'|'get_stats'|'add_debug_flags')
     :type function: str
+    :param debug_flags_to_add: Contains a ',' separated string of debug flags to add to the simulation. This field will be only work if the value of the 'function' argument is 'add_debug_flags'.
```
powerjg: Make it clear that `-` at the beginning of the flag means to remove it.

util/hypercall_external_signal/orchestrator-request.py
```diff
@@ -241,7 +249,14 @@ def main():
...
     parser.add_argument(
-        "function", choices=["status", "get_stats"], help="Function to execute"
+        "function",
+        choices=["status", "get_stats", "add_debug_flags"],
+        help="Function to execute",
+    )
+    parser.add_argument(
+        "--debug-flags-to-add",
+        metavar="FLAG1,FLAG2,..",
+        help="Sets the flags for debug output. This flag will only work if positional argument 'add_debug_flags' is passed",
```
powerjg: Again, make it clear that `-` means to remove it. Also, specify what happens if a flag is "invalid"

powerjg:
This looks good. Can you extend the description (at least on GitHub) with an example of how you can use this? Also, an explanation of what you can do and why it's useful would be good to have in the description.

Finally, do you mean for this to have 2 commits, or do you want a single commit? If a single commit, then you need to update the commit message.

powerjg:
@Harshil2107 can you make these small changes so we can get this merged?

Harshil2107:
I have updated the PR description with more details and an example. We can squash this PR since we don't need two commits. This will allow the PR description to become the commit message, providing more context in the commit history.

powerjg (approved):

PR #2162, opened 2025-04-09
arch-vega: Add pagetable walker buffer

TomXia:
Add a 64-entry PTE buffer to the pagetable walker. It significantly reduces memory accesses, improving address translation performance.

### DIFF
--- a/src/arch/amdgpu/vega/VegaGPUTLB.py
+++ b/src/arch/amdgpu/vega/VegaGPUTLB.py
@@ -41,6 +41,8 @@ class VegaPagetableWalker(ClockedObject):
     cxx_header = "arch/amdgpu/vega/pagetable_walker.hh"
     port = RequestPort("Port for the hardware table walker")
     system = Param.System(Parent.any, "system object")
+    b_size = Param.Int(64, "Page walk cache size")
+    enable_pte_buffer = Param.Bool(True, "Enable page walk cache")
 
 
 class VegaGPUTLB(ClockedObject):
--- a/src/arch/amdgpu/vega/VegaGPUTLB.py
+++ b/src/arch/amdgpu/vega/VegaGPUTLB.py
@@ -32,12 +32,27 @@ from m5.objects.AMDGPU import AMDGPUDevice
 from m5.objects.ClockedObject import ClockedObject
 from m5.objects.IndexingPolicies import (
     BaseIndexingPolicy,
-    SetAssociative,
 )
 from m5.objects.ReplacementPolicies import LRURP
 from m5.params import *
 from m5.proxy import *
 from m5.SimObject import SimObject
+
+
+class VegaPWCIndexingPolicy(BaseIndexingPolicy):
+    """
+    This class implements the GPU page walk cache indexing policy.
+    This policy is specific to the GPU page table entries.
+    """
+
+    type = "VegaPWCIndexingPolicy"
+    cxx_class = "gem5::VegaISA::VegaPWCIndexingPolicy"
+    cxx_header = "arch/amdgpu/vega/page_walk_cache.hh"
+
+    entries = Param.Int(
+        Parent.page_walk_cache_entries,
+        "Number of entries in the page walk cache",
+    )
 
 
 class VegaPagetableWalker(ClockedObject):
@@ -47,15 +62,13 @@ class VegaPagetableWalker(ClockedObject):
     port = RequestPort("Port for the hardware table walker")
     system = Param.System(Parent.any, "system object")
 
-    page_walk_cache_entires = Param.Int(64, "Page walk cache entries")
+    page_walk_cache_entries = Param.Int(64, "Page walk cache entries")
     pwc_replacement_policy = Param.BaseReplacementPolicy(
         LRURP(), "Replacement policy of the PWC"
     )
-    pwc_indexing_policy = Param.SetAssociative(
-        SetAssociative(
-            assoc=Parent.page_walk_cache_entires, size="512", entry_size=8
-        ),
-        "Indexing policy of the PWC, should be SetAssociative.",
+    pwc_indexing_policy = Param.VegaPWCIndexingPolicy(
+        VegaPWCIndexingPolicy(assoc=Parent.page_walk_cache_entries),
+        "Indexing policy of the PWC. Must be GPU PWC specific",
     )
     enable_pwc = Param.Bool(True, "Enable page walk cache")
 
--- a/src/arch/amdgpu/vega/VegaGPUTLB.py
+++ b/src/arch/amdgpu/vega/VegaGPUTLB.py
@@ -30,6 +30,11 @@
 from m5.defines import buildEnv
 from m5.objects.AMDGPU import AMDGPUDevice
 from m5.objects.ClockedObject import ClockedObject
+from m5.objects.IndexingPolicies import (
+    BaseIndexingPolicy,
+    SetAssociative,
+)
+from m5.objects.ReplacementPolicies import LRURP
 from m5.params import *
 from m5.proxy import *
 from m5.SimObject import SimObject
@@ -41,7 +46,17 @@ class VegaPagetableWalker(ClockedObject):
     cxx_header = "arch/amdgpu/vega/pagetable_walker.hh"
     port = RequestPort("Port for the hardware table walker")
     system = Param.System(Parent.any, "system object")
-    page_walk_cache_size = Param.Int(64, "Page walk cache size")
+
+    page_walk_cache_entires = Param.Int(64, "Page walk cache entries")
+    pwc_replacement_policy = Param.BaseReplacementPolicy(
+        LRURP(), "Replacement policy of the PWC"
+    )
+    pwc_indexing_policy = Param.SetAssociative(
+        SetAssociative(
+            assoc=Parent.page_walk_cache_entires, size="512", entry_size=8
+        ),
+        "Indexing policy of the PWC, should be SetAssociative.",
+    )
     enable_pwc = Param.Bool(True, "Enable page walk cache")
 
 
--- a/src/arch/amdgpu/vega/VegaGPUTLB.py
+++ b/src/arch/amdgpu/vega/VegaGPUTLB.py
@@ -41,8 +41,8 @@ class VegaPagetableWalker(ClockedObject):
     cxx_header = "arch/amdgpu/vega/pagetable_walker.hh"
     port = RequestPort("Port for the hardware table walker")
     system = Param.System(Parent.any, "system object")
-    b_size = Param.Int(64, "Page walk cache size")
-    enable_pte_buffer = Param.Bool(True, "Enable page walk cache")
+    page_walk_cache_size = Param.Int(64, "Page walk cache size")
+    enable_pwc = Param.Bool(True, "Enable page walk cache")
 
 
 class VegaGPUTLB(ClockedObject):
--- a/src/arch/amdgpu/vega/page_walk_cache.hh
+++ b/src/arch/amdgpu/vega/page_walk_cache.hh
@@ -1,0 +1,159 @@
+/*
+ * Copyright (c) 2025 Jason Lowe-Power
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __ARCH_AMDGPU_VEGA_PAGE_WALK_CACHE_HH__
+#define __ARCH_AMDGPU_VEGA_PAGE_WALK_CACHE_HH__
+
+
+#include "arch/amdgpu/vega/pagetable.hh"
+#include "base/cache/associative_cache.hh"
+#include "mem/cache/replacement_policies/replaceable_entry.hh"
+#include "mem/cache/tags/indexing_policies/set_associative.hh"
+#include "params/VegaPWCIndexingPolicy.hh"
+
+namespace gem5
+{
+
+
+namespace VegaISA
+{
+
+// Page walk cache entry
+struct PWCEntry : public ReplaceableEntry
+{
+  public:
+    using IndexingPolicy = VegaPWCIndexingPolicy;
+    using KeyType = Addr;
+
+    // The data stored in PTE
+    PageTableEntry pteEntry;
+
+    // The physical address of the page table entry
+    Addr paddr;
+
+    bool valid;
+
+    void
+    invalidate()
+    {
+        valid = false;
+    }
+
+    void insert(const KeyType &key) {}
+    bool isValid() const { return valid; }
+
+    bool
+    match(const KeyType &key) const
+    {
+        return valid && paddr == key;
+    }
+};
+
+/**
+ * Set associative indexing policy for the page walk cache.
+ * This policy differs from the standard set associative policy in that it
+ * hard codes the size of the entry to be 8 bytes (1 PTE).
+ * Note that other PWC designs may need to extend or modify this policy.
+ */
+class VegaPWCIndexingPolicy : public BaseIndexingPolicy
+{
+  protected:
+    virtual uint32_t extractSet(const Addr addr) const
+    {
+        // Extract the set bits from the address
+        return (addr >> setShift) & setMask;
+    }
+
+  public:
+    /**
+     * Convenience typedef.
+     */
+    typedef VegaPWCIndexingPolicyParams Params;
+
+    /**
+     * Construct and initialize this policy.
+     * Assume that all PTEs are 8 bytes so the set shift is 3 bits.
+     */
+    VegaPWCIndexingPolicy(const Params &p) :
+        BaseIndexingPolicy(p, p.entries, 3)
+    {}
+
+    /**
+     * Destructor.
+     */
+    ~VegaPWCIndexingPolicy() {};
+
+    std::vector<ReplaceableEntry*> getPossibleEntries(const Addr &addr) const
+                                                                     override
+    {
+        return sets[extractSet(addr)];
+    }
+
+    Addr regenerateAddr(const Addr &tag,
+                        const ReplaceableEntry* entry) const override
+    {
+        panic("PWCSetAssociative::regenerateAddr() not implemented");
+        return 0;
+    }
+};
+
+// Page walk cache
+class PageWalkCache : public AssociativeCache<PWCEntry>
+{
+  public:
+    using AssociativeCache<PWCEntry>::AssociativeCache;
+    using AssociativeCache<PWCEntry>::accessEntry;
+
+    PWCEntry* accessEntry(const KeyType &key) override
+    {
+        auto entry = findEntry(key);
+        accessEntry(entry);
+        return entry;
+    }
+    PWCEntry* findEntry(const KeyType &key) const override
+    {
+        for (auto candidate : indexingPolicy->getPossibleEntries(key)) {
+        auto entry = static_cast<PWCEntry*>(candidate);
+        if (entry->match(key))
+            return entry;
+        }
+        return nullptr;
+    }
+    void insert(const KeyType &key, const PageTableEntry &pte_entry) {
+        PWCEntry *vict = findVictim(key);
+        vict->pteEntry = pte_entry;
+        vict->paddr = key;
+        vict->valid = true;
+        insertEntry(key, vict);
+    }
+};
+
+} // namespace VegaISA
+} // namespace gem5
+
+#endif // __ARCH_AMDGPU_VEGA_PAGE_WALK_CACHE_HH__
--- a/src/arch/amdgpu/vega/pagetable_walker.cc
+++ b/src/arch/amdgpu/vega/pagetable_walker.cc
@@ -100,47 +100,6 @@ Walker::WalkerState::startFunctional(Addr base, Addr vaddr,
     return fault;
 }
 
-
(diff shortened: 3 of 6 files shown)


src/arch/amdgpu/vega/pagetable_walker.cc
```diff
@@ -101,6 +101,45 @@ Walker::WalkerState::startFunctional(Addr base, Addr vaddr,
...
+    DPRINTF(GPUPTWalker, "Insert pte for paddr %#x. Current length: %d\n",
+            paddr, entryList.size());
+    *newEntry = std::make_pair(paddr, entry);
+    entryList.push_front(newEntry);
+}
+PageTableEntry* Walker::lookup(Addr paddr)
+{
+    auto entry = entryList.begin();
+
+    for ( ; entry != entryList.end(); ++entry)
```
mattsinc: please add a comment explaining what this is doing

src/arch/amdgpu/vega/pagetable_walker.cc
```diff
@@ -165,8 +204,8 @@ Walker::WalkerState::startWalk()
         DPRINTF(GPUPTWalker, "Sending timing read to %#lx\n",
                 read->getAddr());
 
-        sendPackets();
         started = true;
+        sendPackets();
```
mattsinc, line 168: why did you invert the order of these two lines?

src/arch/amdgpu/vega/pagetable_walker.cc
```diff
@@ -366,23 +405,36 @@ Walker::WalkerState::sendPackets()
...
         DPRINTF(GPUPTWalker, "Timing request for %#lx failed\n",
-                read->getAddr());
+                addr);
 
         retrying = true;
-    } else {
-        DPRINTF(GPUPTWalker, "Timing request for %#lx successful\n",
-                read->getAddr());
     }
+    //No lines after sendTiming because the state might be freed
```
mattsinc: if this isn't used anymore then we should remove these lines instead of commenting them out

src/arch/amdgpu/vega/pagetable_walker.cc
```diff
@@ -427,8 +492,9 @@ void
 Walker::recvReqRetry()
 {
     std::list<WalkerState *>::iterator iter;
-    for (iter = currStates.begin(); iter != currStates.end(); iter++) {
+    for (iter = currStates.begin(); iter != currStates.end(); ) {
         WalkerState * walkerState = *(iter);
+        iter++;
```
mattsinc, line 461: why are you incrementing iter here vs. at the bottom of the loop?  Also seems a little weird you had to move it at all...

src/arch/amdgpu/vega/pagetable_walker.hh
```diff
@@ -54,6 +54,11 @@ namespace VegaISA
 class Walker : public ClockedObject
 {
   protected:
+  typedef std::pair<Addr, PageTableEntry> paddr_pte_t;
```
mattsinc: please add comments explaining what each of these are used for

src/arch/amdgpu/vega/pagetable_walker.hh
```diff
@@ -199,6 +215,12 @@ class Walker : public ClockedObject
         deviceRequestorId(999), system(p.system)
     {
         DPRINTF(GPUPTWalker, "Walker::Walker %p\n", this);
+        pte_buffer = false;
+        int num_sets = 64;
```
mattsinc: probably we want to make this a configurable thing eventually, but probably fine for now

src/arch/amdgpu/vega/tlb.cc
```diff
@@ -84,6 +84,7 @@ GpuTLB::GpuTLB(const VegaGPUTLBParams &p)
     missLatency1 = p.missLatency1;
     missLatency2 = p.missLatency2;
 
+
```
mattsinc: remove this

src/arch/amdgpu/vega/tlb.cc
```diff
@@ -104,6 +105,8 @@ GpuTLB::GpuTLB(const VegaGPUTLBParams &p)
     if (gpuDevice) {
         gpuDevice->getVM().registerTLB(this);
     }
+
+    walker->setEnablePTEBuffer(true);
```
mattsinc: probably we want the walker to be configurable eventually, but ok for now.

mattsinc (changes requested):
I'm not sure the commit message is right -- I think what you've done here is add a page walk cache (PWC) to the page table walker (PTW).  We should probably update the commit message to be very explicit about this.

In general this looks good though, thanks!

powerjg (commented):
If you're adding a new cache to gem5, you really should use the generic cache objects so that you can use replacement policies, tagging policies, etc.

See https://github.com/gem5/gem5/tree/stable/src/base/cache and how that code is used all over the codebase.

src/arch/amdgpu/vega/pagetable_walker.cc
powerjg:
```suggestion
    for (auto entry : entryList) {
```
Something like that would be better and more standard modern C++

src/arch/amdgpu/vega/tlb.cc
abmerop: This should be moved into the walker itself.  You could technically not have TLBs, then the this feature would not be used. I would configure this right away you can add a parameter here with the default value of true: https://github.com/gem5/gem5/blob/develop/src/arch/amdgpu/vega/VegaGPUTLB.py#L38 .

src/arch/amdgpu/vega/pagetable_walker.cc
abmerop: Agree, move this back

src/arch/amdgpu/vega/pagetable_walker.cc
abmerop: I am inclined to keep this code. It's something I look for when debugging translation issues so you can see which level of the page table you received a "weird" page table entry

src/arch/amdgpu/vega/pagetable_walker.hh
abmerop: I would also make this configurable immediately. You can add a parameter here https://github.com/gem5/gem5/blob/develop/src/arch/amdgpu/vega/VegaGPUTLB.py#L38 e.g., page_walk_cache_size and then replace num_sets with p.page_walk_cache_size

src/arch/amdgpu/vega/pagetable_walker.cc
abmerop: You can drop this change. Generally want to touch as few lines of code as possible

abmerop (changes requested):
I marked few changes that shouldn't take too long or need retesting.

Optionally I would recommend some more descriptive variable/function names for things like buffer, insert, lookup that make it clear they are for the page walk cache and avoid confusion that you are looking up/inserting to the page table itself

mattsinc:
@TomXia how are the changes coming along that @abmerop and I requested?

@powerjg : we had discussed using that instead, and decided against it because this isn't a true cache (which is what @TomXia referred to is as a buffer instead).  @abmerop do you remember why we opted not to use that cache format?

abmerop:
> @powerjg : we had discussed using that instead, and decided against it because this isn't a true cache (which is what @TomXia referred to is as a buffer instead). @abmerop do you remember why we opted not to use that cache format?

This is so small it would not be implemented as a cache in real hardware

src/arch/amdgpu/vega/pagetable_walker.cc
TomXia: It must be reordered here. Because, sendPackets() might immediately have the response due to a buffer hit, which causes the WalkerState to change before sendPackets() returns. Therefore, the state (started) needs to be updated before sendPackets().

src/arch/amdgpu/vega/pagetable_walker.cc
TomXia: Similarly, after retry(), the current iterator might be erased, because the request is immediately returned and finishes the walk. We need to move to the next iterator before the current one becomes invalid.

src/arch/amdgpu/vega/pagetable_walker.cc
mattsinc: @TomXia : still need to make change @powerjg requested here

src/arch/amdgpu/vega/pagetable_walker.cc
```diff
@@ -366,23 +405,36 @@ Walker::WalkerState::sendPackets()
     // If we're already waiting for the port to become available, just return.
     if (retrying)
         return;
-
+    auto addr = read->getAddr();
```
mattsinc: Since this variable is only used a DPRINTF for LLVM compilation I think we need to use the __maybe__unused__ declaration?

src/arch/amdgpu/vega/pagetable_walker.cc
mattsinc: I'm not sure why moving it here accomplishes that goal?

src/arch/amdgpu/vega/pagetable_walker.hh
mattsinc: this remains

src/arch/amdgpu/vega/tlb.cc
mattsinc: this remains

mattsinc (changes requested):
Getting closer!  Some questions, as well as some changes that still are needed

src/arch/amdgpu/vega/pagetable_walker.cc
TomXia: Thanks for pointing out.
To track entries in LRU order, the container might be modified. Range based for might not be a direct way to implement it. Shall I still proceed to change?

TomXia:
Thanks for the comments. A new commit has been pushed, with a new commit message.

src/arch/amdgpu/vega/pagetable_walker.cc
mattsinc: How does the change impact LRU order?

src/arch/amdgpu/vega/pagetable_walker.cc
```diff
@@ -101,6 +101,44 @@ Walker::WalkerState::startFunctional(Addr base, Addr vaddr,
...
+    } else {
+        newEntry = pwcEntryList.back();
+        pwcEntryList.pop_back();
+    }
+    DPRINTF(GPUPTWalker, "Insert pte for paddr %#x. Current length: %d\n",
+            paddr, pwcEntryList.size());
+    *newEntry = std::make_pair(paddr, entry);
+    pwcEntryList.push_front(newEntry);
+}
+PageTableEntry* Walker::pwcLookup(Addr paddr)
```
mattsinc: There should be a line of space between these functions

src/arch/amdgpu/vega/pagetable_walker.cc
```diff
@@ -427,8 +491,9 @@ void
 Walker::recvReqRetry()
 {
     std::list<WalkerState *>::iterator iter;
-    for (iter = currStates.begin(); iter != currStates.end(); iter++) {
+    for (iter = currStates.begin(); iter != currStates.end(); ) {
         WalkerState * walkerState = *(iter);
+        iter++;//increment before it becomes invalid
```
mattsinc: Please put a space between semicolon and comment

mattsinc: Also probably need a more descriptive comment -- you need to explain that the retry() function may update the current iterator element, hence updating iter ahead of time is needed (or something like this)

src/arch/amdgpu/vega/VegaGPUTLB.py
```diff
@@ -41,6 +41,8 @@ class VegaPagetableWalker(ClockedObject):
     cxx_header = "arch/amdgpu/vega/pagetable_walker.hh"
     port = RequestPort("Port for the hardware table walker")
     system = Param.System(Parent.any, "system object")
+    b_size = Param.Int(64, "Page walk cache size")
```
mattsinc: I'm not sure b_size is a sufficiently descriptive name so that it's clear.  I would prefer pwcSize or something like that

mattsinc: likewise @abmerop recommended page_walk_cache_size

src/arch/amdgpu/vega/pagetable_walker.hh
```diff
@@ -199,6 +214,14 @@ class Walker : public ClockedObject
         deviceRequestorId(999), system(p.system)
     {
         DPRINTF(GPUPTWalker, "Walker::Walker %p\n", this);
+        pte_buffer = p.enable_pte_buffer;
```
mattsinc: probably we want a more descriptive name than pte_buffer -- pte_buffer sounds more like the name of a pointer to the pteBuffer.

src/arch/amdgpu/vega/pagetable_walker.cc
mattsinc: I think this still needs to be done?

mattsinc (changes requested):
Getting closer!

src/arch/amdgpu/vega/pagetable_walker.cc
TomXia: entryList (now renamed to pwcEntryList) here maintains the LRU-ordered list of current cached items. Thus, the list needs to be updated in this lookup method. Using regular for loop, it is simper to reorder the list within the loop because the iterators automatically come with it.

Although, there might be a different and efficient way to achieve it with range-based for that I am not aware of.

TomXia:
New commit pushed.

powerjg:
> > @powerjg : we had discussed using that instead, and decided against it because this isn't a true cache (which is what @TomXia referred to is as a buffer instead). @abmerop do you remember why we opted not to use that cache format?
> 
> This is so small it would not be implemented as a cache in real hardware

That's fine. I'm happy for this to be merged without this change.

*However*, even if this isn't a cache in hardware, that doesn't mean that the `AssociativeCache` (found in `src/base/cache`) class isn't the right way to model it in gem5. There's a couple of benefits to using the `AssociativeCache`. (1) Right now, this is LRU, but what if you wanted to change that to random or NMRU or anything else? If you use `AssociativeCache` all of the replacement policies are implemented *for you*. *You have no work to do!* (2) Maybe in the future, someone wants to investigate a more complex structure (e.g., not fully associative). Again, if you use `AssociativeCache` then there's literally no code that needs to be changed to do that. You get the features for free.

The downside is low (very little complexity is added to the code), but the potential upside is high. It was just a suggestion, so no need to take action, but I would encourage you to consider these tradeoffs when submitting future gem5 changes.

src/arch/amdgpu/vega/pagetable_walker.cc
mattsinc: Maybe I'm just missing something, but I'm asking for why this:

for ( ; entry != entryList.end(); ++entry)

and this:

for (auto entry : entryList) {

would lead to a different update order.  To me, they seem like they would both result in the same order ... I'm pretty sure the second option is the same as the first but using newer C++ primitives?

mattsinc (changes requested):
I think the main thing that remains is the auto change @powerjg asked for, which I don't totally follow.  It looks like everything else is resolved!

src/arch/amdgpu/vega/pagetable_walker.hh
```diff
@@ -199,6 +214,14 @@ class Walker : public ClockedObject
         deviceRequestorId(999), system(p.system)
     {
         DPRINTF(GPUPTWalker, "Walker::Walker %p\n", this);
+        enable_pwc = p.enable_pwc;
+        int buf_size = p.page_walk_cache_size;
+        pwcBuffer.assign(buf_size, std::make_pair((Addr)0, (PageTableEntry)0));
+        for (int buf_i = 0; buf_i < buf_size; ++buf_i) {
+            pwcFreeList.push_back(&pwcBuffer.at(buf_i));
+        if (enable_pwc)
+          DPRINTF(GPUPTWalker, "PWC enabled, size %d\n", pwcBuffer.size());
```
powerjg:
```suggestion
          if (enable_pwc) {
               DPRINTF(GPUPTWalker, "PWC enabled, size %d\n", pwcBuffer.size());
          }
```
Indentation is wrong too

src/arch/amdgpu/vega/pagetable_walker.hh
```diff
@@ -54,6 +54,15 @@ namespace VegaISA
...
   protected:
+  //PTE indexed by paddr
+  typedef std::pair<Addr, PageTableEntry> paddr_pte_t;
+  //the PTE buffer
+  std::vector<paddr_pte_t> pwcBuffer;
+  //List of valid entries in the buffer, LRU ordered
+  std::list<paddr_pte_t*> pwcEntryList;
+  //List of free entries in the buffer
+  std::list<paddr_pte_t*> pwcFreeList;
+
```
powerjg: Indentation is wrong

src/arch/amdgpu/vega/pagetable_walker.cc
```diff
@@ -412,11 +464,24 @@ Walker::recvTimingResp(PacketPtr pkt)
 
     DPRINTF(GPUPTWalker, "Got response for %#lx from walker %p -- %#lx\n",
             pkt->getAddr(), senderState->senderWalk, pkt->getLE<uint64_t>());
+    if (enable_pwc && pwcLookup(pkt->getAddr()) == nullptr) //not found, push
+        pwcInsert(pkt->getAddr(), pkt->getLE<uint64_t>());
```
powerjg: Please use curly braces... I know it seems silly, but until you spend hours tracking down a bug because someone else thought they would save themselves 2 characters...

src/arch/amdgpu/vega/pagetable_walker.cc
```diff
@@ -366,23 +405,36 @@ Walker::WalkerState::sendPackets()
...
 }
 
 bool Walker::sendTiming(WalkerState* sending_walker, PacketPtr pkt)
 {
     auto walker_state = new WalkerSenderState(sending_walker);
     pkt->pushSenderState(walker_state);
 
+    PageTableEntry *entry = pwcLookup(pkt->getAddr());
+    if (entry != nullptr)
+    {
```
powerjg: Style
```suggestion
    if (entry != nullptr) {
```

src/arch/amdgpu/vega/pagetable_walker.cc
powerjg: Here's the problem with this code: The comment isn't correct. It's not "searching the buffer for the paddr." It's updating the LRU position of an entry. If the code did what the comment said, then it would be read-only and a range-based for loop would work. What you're actually doing is manipulating the list inside an iteration (which is error prone, but looks OK in this case). As you said @TomXia you probably can't use the range-based for given you're changing the list.

(I will also go back to my first comment which is if you used the `AssociativeCache` from base/ you wouldn't have to worry about any of this.)

powerjg (commented):
There are many style issues still.

mattsinc:
@TomXia just checking if you are working on addressing @powerjg 's feedback?

TomXia:
Thanks @powerjg . I have rewrite with `AssociativeCache`. Also fixed the styling issues. I have taken notes on styling rules and will ensure they are strictly followed.

powerjg (commented):
First of all, thanks for taking the time to use the base indexing policy and more generic cache. There's a lot less code in the change now!

Looking at this implementation, I got to thinking... The base cache/indexing policy doesn't really work correctly for what you're trying to do. A better approach would be to create your own indexing policy instead of using the set associative one.

This isn't super straightforward, so I took a crack at it for you. See my changes here: https://github.com/powerjg/gem5/tree/pr/TomXia/2162. Specifically https://github.com/gem5/gem5/commit/c08e54da988479f8b7bd721d18d1f4c48301a22a.

Feel free to take that code and squash it all into a single commit authored by you.

However, as I was re-writing the code, I had a question about the intention/implementation. Was the intention to cache *only a single PTE* in an entry in the page walk cache? I.e., is this comment correct? https://github.com/gem5/gem5/commit/cd1c7fc7c867524eb50d31ea059575c180855ff5#diff-e45ce01b5e603be741b8372fbfe1969965c7cb939464c5910cfba1e123f64f99R80

If that was the intention, then how does this structure differ from the TLB?

Finally, **I did not test this code**. The system I'm on doesn't have an easy way for me to test it. It compiles, but if you think this is a good way forward then make sure you test it :).

The main problem I had with the current implementation was the way the indexing policy was hard coded in the SimObject description file. It is bad practice to *instantiate* a SimObject in the "SimObject description" file. This file is for specifying parameters (and some defaults). 
The reason behind this is that it is *incredibly difficult* for a user to know where the default value for the SimObject (in this case the number of entries in the "cache") comes from if there are defaults in the command line parameters, SimObject file, and overridden in scripts.

With my new implementation, you only set the number of entries and nothing is "hard coded."
You can also have an associativity that is not fully associative and it should still work correctly.

src/arch/amdgpu/vega/pagetable_walker.hh
```diff
@@ -193,12 +258,16 @@ class Walker : public ClockedObject
...
+            p.pwc_indexing_policy),
         port(name() + ".port", this),
-        funcState(this, nullptr, true), tlb(nullptr),
+        funcState(this, nullptr, true),
+        enable_pwc(p.enable_pwc), tlb(nullptr),
         requestorId(p.system->getRequestorId(this)),
         deviceRequestorId(999), system(p.system)
     {
-        DPRINTF(GPUPTWalker, "Walker::Walker %p\n", this);
+      DPRINTF(GPUPTWalker, "Walker::Walker %p\n", this);
```
mattsinc: pretty sure indentation is wrong here?

mattsinc (changes requested):
There is one minor change with this code directly, but I'm honestly now sure how to review this anymore given the large changes separately on @powerjg 's branch that seem mostly orthogonal to the original goal of this patch ... as it seems to mostly be about refactoring the what the GPU PTW is working to follow a different format.  I'm not opposed to that, but I also am not sure squashing that and this together makes sense as this commit is about adding a PWC, whereas that commit is about using a different style for modeling the PWC.

I don't really know what to do anymore so if @powerjg is happy and the code actually works with all these changes (which I find difficult to follow as they are now scattered across at least two different places, in two different styles), then we should just merge ...

mattsinc:
@TomXia just checking if you have made at least the one minor change I highlighted?

src/arch/amdgpu/vega/pagetable_walker.hh
TomXia: Sorry, missed this. Fixed in the new commit.

TomXia:
Thanks.

I was indeed struggling passing the parameter to the indexing policy. Now everything looks better. Thanks again @powerjg  for making the change. I tested the implementation and fixed a small issue. Pushed in a new commit.

And Yes, the PWC is similar to TLB in some way. But it acts more like a cache and helps to reduce page table walk cost.

src/arch/amdgpu/vega/page_walk_cache.hh
```diff
@@ -0,0 +1,159 @@
...
+
+namespace gem5
+{
+
+
+namespace VegaISA
+{
+
+// Page walk cache entry
+struct PWCEntry : public ReplaceableEntry
```
powerjg, line 47: I just want to triple check, the intended implementation here is that each entry in the "page walk cache" is a *single* PTE. I.e., the page walk cache is *exactly the same* as a second level TLB except that it is tagged by the address of the PTE instead of the virtual address.

I have to say that I don't understand why the hardware would be that way, but if that's the way it is... 🤷

mattsinc: Yes, each entry is a single PTE.  I don't think the PWC is the same as the L3 GPU TLB though (unless the L3 GPU TLB is the same number of entries) -- because the PWC is (I believe) only going to hold pages that recently missed in the L3 GPU TLB and are going to be sent there (so eventually they'll be cached again in the L3 GPU TLB, yes).  Conversely, the L3 GPU TLB would likely hold more pages if it is larger (and potentially evict them in a different order depending on reuse).  In that sense, I've been thinking of the PWC as being more like a victim buffer for the L3 GPU TLB, but maybe I'm wrong.

@abmerop I believe this is what you had recommended though, so you can chime in if I'm misrepresented anything.

abmerop: PWC is caching intermediate nodes of the page table tree structure.  TLBs cache the leaf nodes.  The intention of the PWC is not to directly cache a translation but to reduce the number of round trips to memory (one for each level in the page table at most, possibly as low as 1 with PWC).

abmerop (approved):

mattsinc (approved):

src/arch/amdgpu/vega/page_walk_cache.hh
mattsinc: I'll wait to make sure @powerjg is ok with @abmerop 's reply before merging

powerjg: I think I understand the difference now. Thanks!

powerjg (approved):

PR #2163, opened 2025-04-09
arch-vega: Improved dispatch scheduler

TomXia:
Now, the wave with the oldest instruction that has been ready is selected for dispatch, rather than choosing the wave that has been in the schList the longest (FIFO). It prioritizes the instructions that have waited the longest. It better mimics the hardware.

### DIFF
--- a/src/gpu-compute/schedule_stage.cc
+++ b/src/gpu-compute/schedule_stage.cc
@@ -574,23 +574,21 @@ ScheduleStage::fillDispatchList()
 
         //find the earliest ready wave according to seqNum
         for (auto iter = schList.at(j).begin();
-            iter != schList.at(j).end(); iter++)
-        {
-            if (iter->second == RFREADY && dispatchReady(iter->first))
-            {
-                if (selected_iter == schList.at(j).end())
+            iter != schList.at(j).end(); iter++) {
+            if (iter->second == RFREADY && dispatchReady(iter->first)) {
+                if (selected_iter == schList.at(j).end()) {
                     selected_iter = iter;
-                else if
-                     (selected_iter->first->seqNum() > iter->first->seqNum())
+                } else if
+                     (selected_iter->first->seqNum() > iter->first->seqNum()) {
                     selected_iter = iter;
+                }
             }
         }
         while (schIter != schList.at(j).end()) {
             // only attempt to dispatch if status is RFREADY
             if (schIter->second == RFREADY) {
                 //oldest wave selected for dispatch
-                if (schIter == selected_iter)
-                {
+                if (schIter == selected_iter) {
                     GPUDynInstPtr mp = schIter->first;
                     if (!mp->isMemSync() && !mp->isScalar() &&
                         mp->needsToken()) {
@@ -607,8 +605,7 @@ ScheduleStage::fillDispatchList()
                             "EMPTY->EXREADY\n", j);
                     schIter->first = nullptr;
                     schIter++;
-                }
-                else {
+                } else {
                     // Either another wave has been selected for dispatch,
                     // or this wave was not ready, so it is stalled this cycle
                     schIter->first->wavefront()->stats.schStalls++;
--- a/src/gpu-compute/schedule_stage.cc
+++ b/src/gpu-compute/schedule_stage.cc
@@ -570,20 +570,27 @@ ScheduleStage::fillDispatchList()
 
         // iterate waves in schList to pick one for dispatch
         auto schIter = schList.at(j).begin();
-        bool dispatched = false;
+        auto selected_iter = schList.at(j).end();
+
+        //find the earliest ready wave according to seqNum
+        for (auto iter = schList.at(j).begin();
+            iter != schList.at(j).end(); iter++)
+        {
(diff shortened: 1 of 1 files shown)


mattsinc (commented):
I'll approve to kick off the tests.  But @abmerop I think the key question is if we should make this dispatcher variant an option instead of wholesale replacing the existing one?  I.e., do we want to keep the old version around as something they could be selected by the configuration if someone wants to study the behavior of different dispatchers?

mattsinc (dismissed):

powerjg:
> I'll approve to kick off the tests.

You don't need to approve for tests anymore @mattsinc. They will run as long as the submitter has had a commit to gem5 in the past. Even if not, there's just a button to push to allow the tests to run.

abmerop:
> I'll approve to kick off the tests. But @abmerop I think the key question is if we should make this dispatcher variant an option instead of wholesale replacing the existing one? I.e., do we want to keep the old version around as something they could be selected by the configuration if someone wants to study the behavior of different dispatchers?

If it is measured to be closer to hardware then I think completely replacing is fine.  If you want to keep both I would suggest to duplicate the fillDispatchList method since it is fairly small. Then you could use a parameter to choose which function to call.

src/gpu-compute/schedule_stage.cc
```diff
@@ -570,20 +570,27 @@ ScheduleStage::fillDispatchList()
...
+        for (auto iter = schList.at(j).begin();
+            iter != schList.at(j).end(); iter++)
+        {
+            if (iter->second == RFREADY && dispatchReady(iter->first))
+            {
+                if (selected_iter == schList.at(j).end())
+                    selected_iter = iter;
+                else if
+                     (selected_iter->first->seqNum() > iter->first->seqNum())
+                    selected_iter = iter;
```
abmerop: Add curly braces here. It helps avoid accidentally adding a line of code in the future that ends up being outside the conditional. 

At first glance with the else if statement on a different line, I thought line 585 was intended to be inside the else statement what wasn't (it is)

src/gpu-compute/schedule_stage.cc
```diff
@@ -599,13 +606,13 @@ ScheduleStage::fillDispatchList()
...
                     schIter->first = nullptr;
-                    schIter = schList.at(j).erase(schIter);
-                    dispatched = true;
-                } else {
-                    // Either another wave has been dispatched, or this wave
-                    // was not ready, so it is stalled this cycle
+                    schIter++;
+                }
+                else {
+                    // Either another wave has been selected for dispatch,
```
abmerop: style 
```suggestion
                } else {
                    // Either another wave has been selected for dispatch,
```

src/gpu-compute/schedule_stage.cc
```diff
@@ -570,20 +570,27 @@ ScheduleStage::fillDispatchList()
...
-                    // No other wave has been dispatched for this exe
-                    // resource, and this wave is ready. Place this wave
-                    // on dispatchList and make it ready for execution
-                    // next cycle.
-
-                    // Acquire a coalescer token if it is a global mem
-                    // operation.
+                //oldest wave selected for dispatch
+                if (schIter == selected_iter)
+                {
```
abmerop: style 

```suggestion
                if (schIter == selected_iter) {
```

abmerop (commented):
There are a few style issues here. Nothing major and should be quick to change.

mattsinc:
> > I'll approve to kick off the tests. But @abmerop I think the key question is if we should make this dispatcher variant an option instead of wholesale replacing the existing one? I.e., do we want to keep the old version around as something they could be selected by the configuration if someone wants to study the behavior of different dispatchers?
> 
> If it is measured to be closer to hardware then I think completely replacing is fine. If you want to keep both I would suggest to duplicate the fillDispatchList method since it is fairly small. Then you could use a parameter to choose which function to call.

Yes, @TomXia made the change to improve fidelity relative to real hardware ... so ok we can just replace then

mattsinc:
@TomXia how are the changes that @abmerop requested coming?

TomXia:
Thanks for pointing out @abmerop. Changes have been made to the new commit.

src/gpu-compute/schedule_stage.cc
```diff
@@ -570,20 +570,25 @@ ScheduleStage::fillDispatchList()
...
-        bool dispatched = false;
+        auto selected_iter = schList.at(j).end();
+
+        //find the earliest ready wave according to seqNum
+        for (auto iter = schList.at(j).begin();
+            iter != schList.at(j).end(); iter++) {
+            if (iter->second == RFREADY && dispatchReady(iter->first)) {
+                if (selected_iter == schList.at(j).end()) {
+                    selected_iter = iter;
+                } else if
```
mattsinc, line 581: I am not sure if this confirms to the official gem5 style by moving the entire () onto the next line, but it's probably fine.

mattsinc (approved):
@abmerop please also take a look before we merge

abmerop (approved):

mattsinc:
Nice job @TomXia !

PR #2165, opened 2025-04-09
gpu-compute: Fix kernarg preload address

abmerop:
Currently preloaded kernel arguments are being read from the added padding to the code object which is not correct. The arguments should be read from the kernarg segment. This commit fixes the preload address.

This was retested with hipblaslt. The previous testing appeared to work because one preloaded argument (a loop boundary) happened to work. However after detailed tracing it was clear the loop boundary was not correct. The application under test now loops the correct number of times.


abmerop:
@mattsinc I posted during one of your many busy weeks, but if you get a chance, this is a pretty important fix for next release

mattsinc (approved):

PR #2169, opened 2025-04-09
base,systemc,python: Replace use of VLAs with heap allocs/strings

bpfoley:


src/base/cprintf_formats.hh
```diff
@@ -230,10 +230,7 @@ _formatString(std::ostream &out, const T &data, Format &fmt)
         int flen = foo.str().size();
 
         if (fmt.width > flen) {
-            char spaces[fmt.width - flen + 1];
-            std::memset(spaces, ' ', fmt.width - flen);
-            spaces[fmt.width - flen] = 0;
-
+            std::string spaces(fmt.width - flen + 1, ' ');
```
odanrc: Shouldn´t the +1 have been removed?

bpfoley: Good catch, thankyou!

odanrc (approved):

erin-le (approved):

PR #2171, opened 2025-04-09
sim: Use heap allocs instead of VLAs

bpfoley:


src/sim/process.cc
```diff
@@ -390,14 +390,15 @@ Process::replicatePage(Addr vaddr, Addr new_paddr, ThreadContext *old_tc,
         new_paddr = seWorkload->allocPhysPages(1);
 
     // Read from old physical page.
-    uint8_t buf_p[pTable->pageSize()];
-    SETranslatingPortProxy(old_tc).readBlob(vaddr, buf_p, sizeof(buf_p));
+    const size_t buf_size = pTable->pageSize();
+    auto buf_p = std::make_unique<uint8_t[]>(buf_size);
```
giactra, line 394: I am honestly considering whether it would make sense to have a global constexpr maxPageSize in order to statically allocate the array.. Then the EmulationPageTable constructor could assert if PageSize > maxPageSize... Maybe it is not worth it

bpfoley: Well... there are much much worse offenders in gem5 when it comes to heap allocation. For example the allocations made to decode and execute every single instruction, not to mention the multiple allocations that will occur for every load/store when cache modelling is enabled.

powerjg: I tend to agree with Brian here. This code isn't executed very often either, so it shouldn't have a big impact on performance in the end. Given that the performance impact is low, it's slightly more readable this way, so I would advocate for keeping this code as-is.

odanrc (approved):

erin-le (approved):

Issue #709, opened 2023-12-22
Transition to Bazel build system


rw1nkler:
Some time ago, a [discussion](https://www.mail-archive.com/gem5-dev@gem5.org/msg39686.html) was started on the gem5-dev mailing channel regarding simplifying the project's build system. This discussion mentioned a desire to replace [SCons](https://scons.org/) with [Bazel](https://bazel.build/) at some point in the future:

> One thing I feel is worth noting is that there is a desire to move away
from SCons at some point (likely towards bazel: https://bazel.build/). This
is definitely not something we could accomplish in the coming months, as I
believe, at the very least, it would require breaking gem5 further down
into separate modules; but it's something you should be aware is on some
peoples' minds.

Is this approach still desirable and has any work been started to modify the build system? I noticed that some attempts to provide Bazel build rules were reviewed on [Gerrit #58949](https://gem5-review.googlesource.com/c/public/gem5/+/58949). Will this effort be continued?

powerjg:
As far as I know, no one is actively working on this.

rw1nkler:
@powerjg Could you please provide more information on whether Bazel support is still desired? Do you know if any contributions in that area will be accepted, or if development will focus solely on the SCons build system in the future?

powerjg:
Oof, sorry that it's been so long. I just noticed this dropped through the cracks.

Let me try to brain dump here.

In my view, we want a build system that has the following "requirements". These are somewhat in order of importance (in my opinion.
- **Is "standard."** I.e., when someone googles for a problem they can find some help that we don't have to write. I think that scons, especially how we use it, absolutely fails at this requirement. The documentation for scons is hard to come by, and it usually doesn't apply to gem5 because we've customized it so much. It would be great to use the "most common" build system for C++/python. Unfortunately, I'm not sure there is a "most common" build system or if the common build systems (e.g., autoconf and cmake) actually support the other features we need.
- **Can deal with code generation.** We have a lot of code generation in gem5. E.g., the `isa_parser` generates the ISA files, SLICC generates Ruby protocol files, and scons generates a bunch of files including C++ wrappers for SimObjects, enums, and possibly other thing.
[...]

PR #2172, opened 2025-04-10
misc: PoC Bazel integration

hchsiao:
Context: #709 (later became a [discussion](https://github.com/orgs/gem5/discussions/917))
Because the issue was closed, I outlined my plan to approach this in the discussion.

I acknowledge that not everyone perceives Bazel's merit the same way. I also understand that Bazel is not currently as popular as many alternatives (large awareness but low actual/active usage). One foreseeable implication of introducing this Bazel support is that gem5 may begin to receive Bazel-related questions from users. However, the existing SCons workflow remains fully supported and unchanged, meaning that addressing Bazel-specific questions will initially be a lower priority. This change serves primarily as a proof-of-concept rather than a fully functional alternative intended to replace SCons.

The primary goal of this pull request is to gauge the gem5 project's interest in pursuing Bazel integration by presenting this prototype implementation. Internally, our team has already begun adopting Bazel for our own fork of gem5, currently focusing on the specific configurations we utilize. If this initial proof-of-concept is well-received and deemed worth pursuing for upstream gem5, we are prepared to contribute further, leveraging our internal efforts to potentially extend Bazel support to cover a wider range of use cases.

Note that this PR adds more than 60k lines, but most of them are generated BUILD files which are verbose and uniform in nature. Technical details and some more high-level information can be found in `build_tools/bazel/README.md`.


powerjg:
I have a few of quick comments:

1. Thanks for the contribution! I've been thinking about bazel integration for at least 5 years, probably more. It's great to see some movement in that direction!
2. I think we should *not* support kconfig going forward. Now that we can build a single binary with *all* of gem5's features, the need for different binary specializations should go away. IMO, we should focus our development efforts on a single universal gem5 binary and actually distributing that binary. The main reason I advocate for this position is that maintaining support for the cross product of different build-time configuration options consumes many developer resources. It would also do many of our users a service if we distributed a gem5 binary.
3. When I initially was thinking about bazel integration, one thing that I hoped for is that it would be a forcing function to make gem5 more modular. Specifically, I was hoping that we could, at the same time as re-writing the build system, also make gem5 into a set of shared objects (e.g., o3cpu.so, arm_isa.so, etc.) that all link to a single (small) gem5 binary. This would solve one of the major problems with my suggestion (2) above since (2) would cause us to have a very large binary.

As a separate thing, if we can get modularity working, users won't have to compile gem5 from source. Instead, we can build a plugin-like system that allows people to just build a new SimObject into a shared library that gem5 can find and use. This would make it so that users don't have to build gem5 from source but just have to build their own new objects. This long-term vision would mean the build system matters a lot less to our users.

Finally, I think that in retrospect using SCons was a bad decision. (I'm sure at the time it was a good decision!) It was the "cool new thing" at the time, and in the long run (e.g., next 20 years) it did not mature like the alternatives. I always tell people to choose the *most popular* tool, not the "coolest." Going with that idea, we really should be migrating to cmake. If I were a betting person, I would bet that in 2045 cmake will have a larger community than bazel. (Ok, now that I write that I guess I would really bet that no one does any programming in 2045, so maybe the build system doesn't matter? Let's just ask Claude or Gemini what they want to use :D.)

That said... I *do* like the idea of bazel. I recognize that Bazel has strong benefits over CMake, particularly in terms of scalability, incremental builds, and dependency management. However, as gem5 is currently implemented, I don't see the broader community using these advantages of bazel. Going forward, if we can move toward this more modular vision of gem5, then the build system matters less, because it only needs to work for *developers* not for all possible *users*. Thus, it could make sense to choose something more developer focused like bazel.

As a quick gut feeling:
- I do not like SCons. Getting rid of scons would be great! Similarly, kconfig has been quite painful to maintain. In general, gem5's build system is (to me) one of the most painful parts of gem5 to maintain.
- I don't like the idea of having to maintain two parallel build systems. That seems like just extra work.
- I don't like the idea of replacing one non-standard build system with another non-standard build system.
- I *love the idea* of making gem5 more modular. If adding bazel support can move us toward being more modular in the long run, then I would be open to it.

To summarize: I don't want to be too negative. I see a lot of positives in bazel. However, I think the bar for making such a big change to a component of gem5 which is already difficult to maintain, use, extend, and understand should be very high.

hchsiao:
Thanks @powerjg for your informative feedback!

One of my original thought was to minimize the impact to the gem5 community. This led me to the idea of keeping both SCons and Bazel and make incremental changes before the bazel build becomes fully functional, which is fairly conservative. Making gem5 more modular seems to be more aggressive to me, and theoretically should be orthogonal to changing the build system. But I actually think this is a better thing to do especially when we can utilize bazel's portfolio during refactoring.

In fact, the current status of this PR IMO is in a good stance to approach this. Instead of adding *all* the remaining features to the current bazel build (like my `README.md`, the old plan, says) and make a monolithic binary, I can instead **"removing unnecessary files from the bazel build"** and just keeping the minimum set to build a core binary. And the core binary should be free of configuration switches. For example, we don't need `learning_gem5` stuff to be included with the core binary. This is one example which can be decoupled easily, but I can also imagine there are more stuff that requires some efforts to decouple from the core. The beauty of bazel is that it can help us here to identify what needs to be decoupled and how to decouple. On top of the core binary, extensions can be added given the dynamically linked binaries built by users' own favorite build tools.

As a side-effect, this still falls into the category of "replacing a non-standard build system with another non-standard build system". But I bet moving from Bazel to cmake shouldn't be harder than moving from SCons to cmake if it's desired :)).

In summary, I am also desired to see that gem5 becomes more modularized. I believe this will benefit the community, including our team. The current bazel build in this PR is actually pretty close to the "core" binary in that not so many SCons transforms are currently ported to bazel. So I think now is the opportunity for me to switch gear to **"stepping toward a more modularized gem5 utilizing bazel"**. This implies several actions required further for this PR:
- Keep the core binary minimal. For example, `learning_gem5` can be removed.
- Remove implementations deviating from this goal. For example, bazel configuration should be removed.
- Come up with an approach to dynamically link non-core binaries. For example, `learning_gem5` will be added back this way

Does this aligns with your vision? Any input would be appreciated. Thanks!

build_tools/BUILD.bazel
```diff
@@ -0,0 +1,52 @@
...
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# Not using py_binary because python runtime is gem5py.
```
powerjg, line 27: For these scripts, we can use py_binary. The reason they were being executed with gem5py is to get around an issue with scons where there are multiple `python` binaries and `libpythonX.Y.so` files. gem5py ensures that whatever `libpythonX.Y.so` file we'll use to build gem5 will be the same one running these files. With bazel this should be unnecessary.

build_tools/bazel/README.md
```diff
@@ -0,0 +1,362 @@
...
+| `[CXXCPRCC]`    | Not yet implemented                                     |
+| `[ENUMDECL]`    | `macro_sim_object()`                                    |
+| `[ENUM STR]`    | `macro_sim_object()`                                    |
+| `[PROTOC]`      | Temporarily using `genrule` and system `protoc`. Ideally `proto_library()` |
+| `[STRIP]`       | Not yet implemented                                     |
+| `[TRACING]`     | Not yet implemented                                     |
+| `[CFG ISA]`     | Not yet implemented                                     |
+| `[DEFINES]`     | `build_tools/bazel/hermetic_wrapper/stub-defines-py.sh` |
+| `[INFO]`        | Ad-hoc `genrule`, see `src/python/m5/BUILD.bazel`       |
+| `[VER TAGS]`    | `build_tools/bazel/hermetic_wrapper/stub-tags-cc.sh`    |
```
powerjg, line 133: Let's just drop this feature. It's not that useful.

build_tools/bazel/README.md
```diff
@@ -0,0 +1,362 @@
...
+1.  **Simplified Maintenance:** Ensure BUILD file maintenance is as
+    straightforward as maintaining SConscripts.
+1.  **Generic Rules Preference:** Prioritize standard Bazel rules over
+    gem5-specific rules to facilitate knowledge transfer and maintainability.
+
+## Current Status
+
+Bazel can build an executable capable of running
+`configs/learning_gem5/part2/run_simple.py`, indicating a core gem5 simulation
+framework is functioning. Build configuration is currently fixed and hardcoded.
```
powerjg, line 20: I suggest we target the ALL and *only* ALL with bazel build. Similarly, let's not allow configuring any "extras" (or EXTRAS). We support one combination of things with this build system.

powerjg (commented):
A few comments below. Also, I'm going to cherry-pick the first commit into its own PR to help get things going.

PR #2174, opened 2025-04-10
mem-ruby: return early response for software prefetches

### LINKED ISSUE #768: Support non-block prefetch inst in Ruby
* Problem
By default, the prefetch instruction in x86 is implemented as normal load instruction (with prefetch flag in the request), and blocks the O3 cpu to wait for the response when we are using Ruby. This causes performance problem as the latency is exposed to the pipeline.

* Possible Solution
Currently, the classic cache returns early for the prefetch request. We could also implement this behavior in the RubySequencer. This would fix this issue for all CPU models and Ruby protocols.

saul44203:
This patch adds support for early response of software prefetches for ruby in a similar way as it's done in the classic cache, preventing these instructions from stalling on the pipeline. This is, sending an early response back to the core just as the request arrives, and pass through a copy of the request whose response will be catched and deleted before it's sent back to the core.

### DIFF
--- a/src/mem/ruby/system/RubyPort.cc
+++ b/src/mem/ruby/system/RubyPort.cc
@@ -258,6 +258,25 @@ RubyPort::MemResponsePort::recvTimingReq(PacketPtr pkt)
         panic("RubyPort should never see request with the "
               "cacheResponding flag set\n");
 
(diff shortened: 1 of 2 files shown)


saul44203:
Related issue #768

powerjg (approved):

PR #2176, opened 2025-04-10
util: Add a progress bar utility for gem5 simulation progress

erin-le:
This PR adds a utility that shows the progress of running gem5 simulations based on the current number of instructions executed compared to the total for a workload. It currently does not work with multisim and is not ready for review.

### DIFF
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -880,10 +880,6 @@ class WorkloadResource(AbstractResource):
             )["total_instructions"]
         except:
             self._estimated_instructions = None
-            # warn(
-            #     "Workload doesn't have simulation_run_results field, "
-            #     "util/hypercall_external_signal/gem5-dashboard.py won't print a progress bar"
-            # )
 
     def get_estimated_instructions(self) -> Optional[str]:
         return self._estimated_instructions
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -880,10 +880,10 @@ class WorkloadResource(AbstractResource):
             )["total_instructions"]
         except:
             self._estimated_instructions = None
-            warn(
-                "Workload doesn't have simulation_run_results field,"
-                "util/hypercall_external_signal/gem5-dashboard.py won't print a progress bar"
-            )
+            # warn(
+            #     "Workload doesn't have simulation_run_results field, "
+            #     "util/hypercall_external_signal/gem5-dashboard.py won't print a progress bar"
+            # )
 
     def get_estimated_instructions(self) -> Optional[str]:
         return self._estimated_instructions
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -874,6 +874,19 @@ class WorkloadResource(AbstractResource):
         self._id = id
         self._func = function
         self._params = parameters
+        try:
+            self._estimated_instructions = kwargs.get(
+                "simulation_run_results"
+            )["total_instructions"]
+        except:
+            self._estimated_instructions = None
+            warn(
+                "Workload doesn't have simulation_run_results field,"
+                "util/hypercall_external_signal/gem5-dashboard.py won't print a progress bar"
+            )
+
+    def get_estimated_instructions(self) -> Optional[str]:
+        return self._estimated_instructions
 
     def get_id(self) -> str:
         """Returns the ID of the workload."""
--- a/src/python/gem5/simulate/exit_handler.py
+++ b/src/python/gem5/simulate/exit_handler.py
@@ -410,7 +410,7 @@ class OrchestratorExitHandler(ExitHandler, hypercall_num=1000):
                 elif function == "get_progress_bar_init_stats":
                     response = json.dumps(
                         {
-                            "total_insts": simulator.get_total_insts(),
+                            "total_insts": simulator.get_workload().get_estimated_instructions(),
                             "sim_id": simulator.get_id(),
                             "workload": simulator.get_workload().get_id(),
                         }
--- a/src/python/gem5/simulate/exit_handler.py
+++ b/src/python/gem5/simulate/exit_handler.py
@@ -407,6 +407,18 @@ class OrchestratorExitHandler(ExitHandler, hypercall_num=1000):
                 elif function == "get_stats":
                     stats = simulator.get_stats()
                     response = json.dumps(stats)
+                elif function == "get_progress_bar_init_stats":
+                    response = json.dumps(
+                        {
+                            "total_insts": simulator.get_total_insts(),
+                            "sim_id": simulator.get_id(),
+                            "workload": simulator.get_workload().get_id(),
+                        }
+                    )
+                elif function == "get_progress_bar_inst_count":
+                    response = json.dumps(
+                        self._get_status(simulator)["instruction_count"]
+                    )
                 else:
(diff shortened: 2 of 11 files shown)


src/python/gem5/simulate/simulator.py
```diff
@@ -191,6 +191,18 @@ def __init__(
         self._last_exit_event = None
         self._exit_event_count = 0
 
+        # self._total_insts = int(self._board.get_workload()._estimated_instructions)
```
powerjg: This doesn't need to be in the simulator. You can get the workload id in the simulator, then we can get information from there.

src/sim/init_signals.cc
```diff
@@ -451,6 +451,7 @@ initSignals()
     // Install a SIGIO handler to handle asynchronous file IO. See the
     // PollQueue class.
     installSignalHandler(SIGIO, ioHandler);
+    initSigRtmin();
```
powerjg: Why this change?

src/sim/init_signals.cc
```diff
@@ -453,9 +473,9 @@ initSignals()
     installSignalHandler(SIGIO, ioHandler);
 }
 
-void initSigRtmin()
+void initSigCont()
 {
-    installSignalHandler(SIGHUP, externalProcessHandler);
+    installSignalHandler(SIGCONT, externalProcessHandler);
 }
```
powerjg, line 470: Can you explain this change?

src/sim/simulate.cc
```diff
@@ -327,6 +330,10 @@ doSimLoop(EventQueue *eventq)
                 async_exception = false;
                 return NULL;
             }
+            if (async_hypercall){
+                async_hypercall = false;
+                processExternalSignal();
+            }
```
powerjg: Can you explain this change?

src/sim/simulate.cc
```diff
@@ -191,13 +193,14 @@ simulate(Tick num_cycles)
...
-    initSigRtmin();
+    initSigCont();
 
     if (global_exit_event)//cleaning last global exit event
         global_exit_event->clean();
     std::unique_ptr<GlobalSyncEvent, DescheduleDeleter> quantum_event;
 
-    inform("Entering event queue @ %d.  Starting simulation...\n", curTick());
+    DPRINTF(EnteringEventQueue, "Entering event queue @ %d. Starting "
+        "simulation...\n", curTick());
```
powerjg: This is a good change, IMO, but it must be in another PR. This is a significant user-facing change that should have many eyes on it. Changes like this shouldn't be in the middle of a "util" PR.

util/hypercall_external_signal/gem5-dashboard.py
```diff
@@ -0,0 +1,131 @@
+import argparse
```
powerjg, line 1: COPYRIGHT

util/hypercall_external_signal/transmitter.py
```diff
@@ -73,7 +73,7 @@
 from time import sleep
 from typing import Optional
 
-logging.basicConfig(level=logging.INFO)
+logging.basicConfig(level=logging.DEBUG)
```
powerjg: You can change this on the command line instead of changing the hard-coded value.

src/sim/init_signals.cc
erin-le: If the dashboard sends SIGHUP to a newly launched gem5 process that hasn't registered signal handlers yet, the process will use the default action for receiving SIGHUP and terminate. I changed the signal to SIGCONT, which is ignored by default.

This is mostly needed to add progress bars to the dashboard for simulations launched while the dashboard is running, though it could also prevent issues with launching the dashboard too soon after launching simulations.

src/sim/simulate.cc
erin-le: This is part of a change to the external signal handler that fixes segmentation faults caused by the dashboard. Previously, the external signal handler wasn't set up in the same way as the other signal handlers, so this changes the format to be the same. I'm not completely sure what was causing the seg faults, but it looks like fixing the format also fixed the seg faults.

src/sim/simulate.cc
erin-le: Separate PR for this change opened here: https://github.com/gem5/gem5/pull/2215

util/hypercall_external_signal/transmitter.py
```diff
@@ -98,23 +101,32 @@ def send_signal(pid: int, id: int, payload: str) -> None:
...
+
+        os.kill(pid, signal.SIGCONT)
     except ProcessLookupError:
-        logger.error(
-            "Process does not exist! Check that you are using the correct PID."
-        )
-        shm.close()
-        shm.unlink()
-        sys.exit(1)
+        if "get_progress_bar_inst_count" in payload:
```
powerjg, line 116: Rather than make this change, it would be better to factor out this function to send the exceptions to the caller. This script was not meant to be used by other scripts. It's not good practice to have sys.exit() in a function this way.

util/hypercall_external_signal/orchestrator_request.py
```diff
@@ -105,22 +108,63 @@ def find_gem5_pid() -> int:
...
+                            # exclude multisim processes that aren't simulations
+                            # using both comm and cmdline. If only cmdline is
+                            # used, we also pick up processes that have "gem5"
+                            # in their names but aren't gem5 simulations.
+                            if (
+                                "gem5.utils.multisim" not in cmdline
+                                and "multiprocessing.resource_tracker"
+                                not in cmdline
+                            ):
+                                gem5_pids.append(int(pid))
```
powerjg: This should be a function since it's the same for linux and mac

Also, you should avoid adding PIDs that are from other users since you can't send signals. Test each PID with a SIGCONT signal and see if you get a `PermissionError`

Finally, add some debug statements in here to facilitate debugging.

src/python/gem5/simulate/exit_handler.py
```diff
@@ -407,6 +407,18 @@ def _process(self, simulator: "Simulator") -> None:
                 elif function == "get_stats":
                     stats = simulator.get_stats()
```
powerjg, line 408: I don't understand why this new code and new function is needed. Could you explain why you can't use `get_stats`? It has all of the information that it needed.

src/python/gem5/resources/resource.py
```diff
@@ -874,6 +874,15 @@ def __init__(
...
         self._func = function
         self._params = parameters
+        try:
+            self._estimated_instructions = kwargs.get(
+                "simulation_run_results"
+            )["total_instructions"]
+        except:
+            self._estimated_instructions = None
+
+    def get_estimated_instructions(self) -> Optional[str]:
```
powerjg, line 884: Shouldn't this return an int?

src/python/gem5/simulate/exit_handler.py
erin-le: Originally, I wasn't sure if I should modify the existing functions, so I added `get_progress_bar_init_stats` to send the number of estimated total instructions + the other stats to initialize a progress bar. I was also planning to send signals to gem5 much more frequently, and thought that sending only the instruction count would improve performance, which isn't a significant concern anymore. I also ended up adding different timeout behavior for functions related to the dashboard. If it's better to use `get_stats`, I can try to refactor the code to use it and remove the new functions.

erin-le:
Closing in favor of https://github.com/gem5/gem5/pull/2250.

PR #2187, opened 2025-04-16
mem-ruby: Revert "mem-ruby: return early response for software prefetches"

erin-le:
Reverts gem5/gem5#2174

@saul44203 Thank you for your contribution! Unfortunately, PR 2174 is causing failures in the daily tests. The details are as follows:

For `testlib-long-tests (fs)`,  `realview64-o3-dual-ruby-ARM_X86-x86_64-opt` and its MatchRegex are failing. The following is printed in simerr:
```
src/mem/ruby/system/RubyPort.cc:283: warn: Cache maintenance operations are not supported in Ruby.
gem5.opt: src/cpu/o3/lsq.cc:1164: virtual bool gem5::o3::LSQ::SingleDataRequest::recvTimingResp(gem5::PacketPtr): Assertion `_numOutstandingPackets == 1' failed.
Program aborted at tick 4489520001
--- BEGIN LIBC BACKTRACE ---
```
For `testlib-long-tests (x86_boot_tests)`, `timing-cpu_1-cores_mesi_two_level_DualChannelDDR3_1600_systemd_x86-boot-test-ALL-x86_64-opt` is failing. The following is printed in simerr:

```
src/arch/generic/debugfaults.hh:145: warn: MOVNTDQ: Ignoring non-temporal hint, modeling as cacheable!
build/ALL/mem/ruby/protocol/MESI_Two_Level/Directory_Transitions.cc:268: panic: Invalid transition
board.cache_hierarchy.ruby_system.directory_controllers0 time: 15159272260 addr: 0xba174e00 event: CleanReplacement state: M_DRD
Memory Usage: 4323588 KBytes
Program aborted at tick 5048037662580
--- BEGIN LIBC BACKTRACE ---
```

### DIFF
--- a/src/mem/ruby/system/RubyPort.cc
+++ b/src/mem/ruby/system/RubyPort.cc
@@ -258,25 +258,6 @@ RubyPort::MemResponsePort::recvTimingReq(PacketPtr pkt)
         panic("RubyPort should never see request with the "
               "cacheResponding flag set\n");
 
-    // For software prefetches, immediately respond to avoid stalling on the
-    // core while still processing the prefetch by passing a copy of the
-    // request through
-    if (pkt->cmd.isSWPrefetch()) {
-        RequestPtr req = std::make_shared<Request>(pkt->req->getPaddr(),
-                                                   pkt->req->getSize(),
-                                                   pkt->req->getFlags(),
-                                                   pkt->req->requestorId());
-        PacketPtr pf = new Packet(req, pkt->cmd);
-        pf->allocate();
-        assert(pf->matchAddr(pkt));
-        assert(pf->getSize() == pkt->getSize());
-
-        pkt->makeResponse();
-        schedTimingResp(pkt, curTick());
-
-        pkt = pf;
-    }
-
     // ruby doesn't support cache maintenance operations at the
     // moment, as a workaround, we respond right away
     if (pkt->req->isCacheMaintenance()) {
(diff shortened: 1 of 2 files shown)


erin-le:
An issue has been created to record this information here: https://github.com/gem5/gem5/issues/2188

powerjg (approved):
Unless we can fix this in ~12 hours, let's revert for now so our tests will pass and re-merge the PR later.

src/mem/ruby/system/RubyPort.cc
```diff
@@ -258,25 +258,6 @@ RubyPort::MemResponsePort::recvTimingReq(PacketPtr pkt)
...
-                                                   pkt->req->getSize(),
-                                                   pkt->req->getFlags(),
-                                                   pkt->req->requestorId());
-        PacketPtr pf = new Packet(req, pkt->cmd);
-        pf->allocate();
-        assert(pf->matchAddr(pkt));
-        assert(pf->getSize() == pkt->getSize());
-
-        pkt->makeResponse();
-        schedTimingResp(pkt, curTick());
```
hnpl, line 275: This is a wild guess but I think this line is the problem. It should be curTick() + 1 or something.

giactra:
Thanks. Just wanted to note, SW prefetch handling is very tricky and requires some coordination with the arch+cpu code... Will link this with some comments I made on a separate PR

src/mem/ruby/system/RubyPort.cc
powerjg: That might help for one of the errors, but for the MESI_Two_Level case, there's a new transition being triggered that hasn't happened before. I'm guessing that doesn't have anything to do with the timing.

PR #2189, opened 2025-04-16
sim: Add warning for non-default create function

powerjg:
It is very difficult to track down the problem when the empty dummy shunt is chosen. It turns out, this is usually because your SimObject has pure virtual functions, but the error message from the compiler simply says the create function can't be found. This commit adds a warning and explanation for this case.

Note: there are a number of places in the codebase where we use non-standard create functions. Mostly it's in the SystemC and TLM integrations. 

This set of changes eliminates false positives by allowing the developer to specify that the SimObject has an explicit create function instead of the default of an implicit (auto generated) create function.

Since there are so few places that we use an explicit create function, with this change we can discuss eliminating the (brittle and difficult to debug) template magic that we currently use. One possible future implementation could eliminate the templates and auto-generate the create function for SimObjects that don't override and not auto-generate only when the developer specifies they want to override it.

I have personally wasted at least 4-8 hours (and you can triple that for my students) on this relatively common error. I think it's best for the community to merge something like this change to give a better warning.

### Old error

```
/usr/bin/ld: build/CXL/python/_m5/param_MSI_L1Cache_Controller.o: in function `void pybind11::cpp_function::initialize<pybind11::cpp_function::initialize<gem5::ruby::MSI::L1Cache_Controller*, gem5::MSI_L1Cache_ControllerParams, , pybind11::name, pybind11::is_method, pybind11::sibling>(gem5::ruby::MSI::L1Cache_Controller* (gem5::MSI_L1Cache_ControllerParams::*)() const, pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&)::{lambda(gem5::MSI_L1Cache_ControllerParams const*)#1}, gem5::ruby::MSI::L1Cache_Controller*, gem5::MSI_L1Cache_ControllerParams const*, pybind11::name, pybind11::is_method, pybind11::sibling>(pybind11::cpp_function::initialize<gem5::ruby::MSI::L1Cache_Controller*, gem5::MSI_L1Cache_ControllerParams, , pybind11::name, pybind11::is_method, pybind11::sibling>(gem5::ruby::MSI::L1Cache_Controller* (gem5::MSI_L1Cache_ControllerParams::*)() const, pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&)::{lambda(gem5::MSI_L1Cache_ControllerParams const*)#1}&&, gem5::ruby::MSI::L1Cache_Controller* (*)(gem5::MSI_L1Cache_ControllerParams const*), pybind11::name const&, pybind11::is_method const&, pybind11::sibling const&)':
/home/jlp/Code/gem5/gem5/ext/pybind11/include/pybind11/pybind11.h:223:(.text+0x20c6): undefined reference to `gem5::MSI_L1Cache_ControllerParams::create() const'
collect2: error: ld returned 1 exit status
```

### New warning (error is unchanged)

```
build/CXL/python/_m5/param_MSI_L1Cache_Controller.cc:145:69: warning: ‘NonDefaultCreate’ is deprecated: Warning: MSI_L1Cache_Controller is not constructible from MSI_L1Cache_ControllerParams. It is deprecated to use non-standard `create()` methods. If you see this warning followed by a linking error, the most likely problem is MSI_L1Cache_Controller has a parent class with pure virtual functions. [-Wdeprecated-declarations]
  145 |                              const MSI_L1Cache_ControllerParams &>> warning_instance;
      |                                                                     ^~~~~~~~~~~~~~~~
build/CXL/python/_m5/param_MSI_L1Cache_Controller.cc:141:38: note: declared here
  141 |         "pure virtual functions.")]] NonDefaultCreate<true> {};
      |
```

### DIFF
--- a/build_tools/sim_object_param_struct_cc.py
+++ b/build_tools/sim_object_param_struct_cc.py
@@ -214,6 +214,11 @@ py::module_ m = m_internal.def_submodule("param_${sim_object}");
 
 # include the create() methods whether or not python is enabled.
 if not hasattr(sim_object, "abstract") or not sim_object.abstract:
+    sim_object._unique_namespace = sim_object.cxx_class.replace("::", "_")
+    sim_object._unique_namespace = sim_object._unique_namespace.replace(
+        "<", "_"
+    ).replace(">", "_")
+    sim_object._unique_namespace += "_create"
     if "type" in sim_object.__dict__:
         code(
             """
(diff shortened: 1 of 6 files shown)


BobbyRBruce (approved):
I'm always in favor of more helpful errors!

PR #2193, opened 2025-04-17
Fix, improve, refactor gpu-tests to improve runner testing stability

BobbyRBruce:
This PR does the following:

* Removes the necessity to use the `gcn-gpu` Docker containers for tests which do not require it (it is still required for SE mode tests).
* Reduces the Random Ruby Test's input size to 1.5 (ensures the nightly runners complete promptly).
* Adds the `VIPER` protocol and `VEGA` GPU ISA to the "build_opts/ALL", allowing for GPU tests to be run from an "ALL/gem5.opt" build. (test simplified as a result of this).
* Moves to more logical places in CI/NIGHTLY/WEEKLY breakdown based on time needed to execute.
* Cleanup GPU tests' use of cached "VEGA" build.
* Merged the separate Weekly GPU tests (1 test per job), into a single job.

Note: To _only_ run tests which require a `gcn-gpu` Docker container "main.py" should be run with `--host gcn_gpu`. To _exclude any tests which require the container_ "main.py" should be run with `--exclude-tags gcn_gpu`.


powerjg (dismissed):
lgtm. @mattsinc should also approve before merging.

@mattsinc the only functional change is that the GPU random Ruby tester is running 1.5M loads instead of 5M loads. I don't think this is a big problem, though.

.github/workflows/daily-tests.yaml
```diff
@@ -38,28 +38,6 @@ jobs:
             - name: Build ALL/gem5.opt
               run: scons --no-compress-debug build/ALL/gem5.opt -j $(nproc)
 
-
```
mattsinc, line 41: I'm assuming this is removed because ALL was changed?

.github/workflows/daily-tests.yaml
```diff
@@ -102,7 +80,7 @@ jobs:
 
             - name: long ${{ matrix.test-type }} tests
               working-directory: ${{ github.workspace }}/tests
-              run: ./main.py run gem5/${{ matrix.test-type }} -j$(nproc) --length=long -vv -t $(nproc)
+              run: ./main.py run gem5/${{ matrix.test-type }} -j$(nproc) --length=long -vv -t $(nproc) --exclude-tags gcn_gpu
```
mattsinc, line 83: Why would we want to exclude GPU tests here?

.github/workflows/daily-tests.yaml
```diff
@@ -145,7 +123,7 @@ jobs:
             - name: long ${{ matrix.test-type }} gem5_library_example_tests
               working-directory: ${{ github.workspace }}/tests
               run: ./main.py run --uid SuiteUID:tests/gem5/gem5_library_example_tests/test_gem5_library_examples.py:test-${{ matrix.test-type }} -j $(nproc)
-                  --length=long -vv
+                  --exclude-tags gcn-gpu --length=long -vv
```
mattsinc: Why would we want to exclude GPU tests here?

.github/workflows/weekly-tests.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
 
             - name: very-long ${{ matrix.test-type }}
               working-directory: ${{ github.workspace }}/tests
-              run: ./main.py run gem5/${{ matrix.test-type }} --length very-long -j$(nproc) -vv
+              run: ./main.py run gem5/${{ matrix.test-type }} --length very-long -j$(nproc) -vv --exclude-tags gcn_gpu
```
mattsinc, line 46: Why would we want to exclude GPU tests here?

tests/gem5/gpu/README.md
```diff
@@ -1,7 +1,12 @@
 # GPU
 
-These tests do random checks to the Ruby GPU protocol within gem5. Additionally, these tests run SE mode GPU tests.
-To run these tests by themselves, you can run the following command in the **gem5** directory:
+These tests:
+
+1. Carry out randomly generated checks to the Ruby GPU protocol within gem5.
+2. Run a set SE mode GPU simulations.
+
+To run the SE mode by themselves, you must run within a `gcn-gpu` Docker
```
mattsinc, line 8: I'm assuming this is still required outside testlib?  Because @BobbyRBruce has a commit in this chain that seems to remove the docker for this test ...

tests/gem5/gpu/test_gpu_ruby_random.py
```diff
@@ -77,8 +85,15 @@
     config=joinpath(
         config.base_dir, "configs", "example", "ruby_gpu_random_test.py"
     ),
-    config_args=["--test-length", "5000000", "--num-dmas", "0"],
-    valid_isas=(constants.vega_x86_tag,),
-    valid_hosts=(constants.host_gcn_gpu_tag,),
+    config_args=[
```
mattsinc, line 88: I don't have a strong sense about the reduction in coverage with this change -- @abmerop it is probably ok?

abmerop: Yes should be fine.  Error with the protocol tester occur almost immediately.

.github/workflows/daily-tests.yaml
```diff
@@ -181,7 +158,7 @@ jobs:
 
             - name: Run Testlib GPU Tests
               working-directory: ${{ github.workspace }}/tests
-              run: ./main.py run --length=long -vvv -t $(nproc) --host gcn_gpu  gem5/gpu
+              run: ./main.py run --length=long -vvv -j $(nproc) --host gcn_gpu
```
abmerop, line 161: Wouldn't this change serialize the tests?

build_opts/ALL
```diff
@@ -12,6 +12,9 @@ RUBY_PROTOCOL_MOESI_hammer=y
 RUBY_PROTOCOL_Garnet_standalone=y
 RUBY_PROTOCOL_CHI=y
 RUBY_PROTOCOL_MSI=y
+RUBY_PROTOCOL_GPU_VIPER=y
+BUILD_GPU=y
+VEGA_GPU_ISA=y
```
abmerop, line 17: I definitely like this change!

abmerop (commented):
Well this is way more advanced github workflow-ing than I have ever done, so I reviewed what I could

mattsinc:
@BobbyRBruce as far as I can tell your additional commit after my comments are about separate things -- let me know if I'm wrong though.  Else it seems like my questions partially remain.

.github/workflows/daily-tests.yaml
BobbyRBruce: Yes, but I think `-t $(nproc)` was a mistake here. Theese daily GPU tests are quite hefty and think running multiple in parallel on our weedy little runners is asking for trouble. So I dropped it. I think this was originally meant to be `-j` which is compilation threads. It makes sense to compile with `$nproc` threads but run each GPU simulation sequentially.

build_opts/ALL
BobbyRBruce: It appears to work. I don't know why we didn't include it sooner!

tests/gem5/gpu/README.md
BobbyRBruce: All the GPU SE mode tests in testlib still must be run within the docker container (e.g., those defined in "test_gpu_apu_se.py"). What I removed were the other tests which do not require this (of note: The GPU Ruby Random tests).

.github/workflows/daily-tests.yaml
BobbyRBruce: This code was removed because it's unnecessary to have a separate pre-build step for "VEGA_X86". This was useful when we have lots of jobs running this binary but now it's all down to one job, "gpu-tests". It's just inefficient to do it this way now so it was refactored out

.github/workflows/daily-tests.yaml
BobbyRBruce: We're really starting to push the limits of our testing framework, so like a lot of it, this is a bit of a hack, but essentially this tag says "this test requires the `gcn-gpu` Docker container to run correctly." So here we're saying "Run all the tests except those that require gcn-gpu." 

When we are running with the Docker container, you can use the `--host gcn_gpu` flag to _only_ run tests which require the container.

This is essentially how I switch between tests that require it and don't.

.github/workflows/weekly-tests.yaml
BobbyRBruce: The GPU Tests which require the `gcn-gpu` Docker container (which this tag signfies) are run below in `gpu-tests`

BobbyRBruce:
> @BobbyRBruce as far as I can tell your additional commit after my comments are about separate things -- let me know if I'm wrong though. Else it seems like my questions partially remain.

You're mostly alright. I think the main thing to keep in mind (which may not have been clear from the PR in its original state) is the Docker container is still a requirement for the SE mode GPU tests so it's still around and still requires some fancy hacks to get integrated into the testing framework. Most of the work here and which you commented on is separating them most more clearly. In addition, now that this PR compiles VEGA and VIPER into the ALL build, we need less bespoke testing for the GPU. They can be run with the ALL/gem5.opt build. That also cleans this up somewhat.

I've updated the PR description with a high-level summary of things.

powerjg (approved):

powerjg:
@BobbyRBruce can you go through the comments and mark "resolved" for anything you're confident are resolved? @mattsinc  can you re-review?

powerjg:
@mattsinc or @abmerop can we have an approval on this? Our daily tests have been failing for weeks, and this change should enable them to pass again. 

To reiterate, most of this PR is just fixing issues in the github actions and has *no functional changes to gem5*. The only functional change is that the GPU random Ruby tester is running 1.5M loads instead of 5M loads. I don't think this is a big problem, though.

BobbyRBruce:
I'm going to be cheeky and merge this, as I'd really like to ensure this makes it in for tonights rounds of tests. Though please feel free to make comments. I can address them in another PR if you decide there's something here you don't like. I think I answered most of the questions you asked. From what I could see you wanted to understand rather than had real criticism. any

mattsinc:
> I'm going to be cheeky and merge this, as I'd really like to ensure this makes it in for tonights rounds of tests. Though please feel free to make comments. I can address them in another PR if you decide there's something here you don't like. I think I answered most of the questions you asked. From what I could see you wanted to understand rather than had real criticism. any

Sorry I had this on my list to review this week, but I guess didn't get to it fast enough.

It's totally on me for being busy, but I'm a bit concerned about the precedent set by just being able to merge things in if I don't respond on the same day ...

.github/workflows/daily-tests.yaml
mattsinc: I think I was asking why the gcn-gpu tests are being excluded (why "run all these tests except those that require gcn-gpu"), not what the flag does.  Maybe unittests-fast-debug is too fast for GPU tests to be included or something?

.github/workflows/daily-tests.yaml
mattsinc: I'm not opposed to running the tests serially, but @abmerop can chime in if he disagrees

mattsinc (commented):
I guess this seems fine

PR #2198, opened 2025-04-18
mem: cache bank implementation

amin-mamandi:
**Summary:**
Cache bank-level modeling to simulate bank contention and conflicts, enabling more accurate architectural analysis of multicore systems.

Key features:
  - Per-bank MSHR entries with configurable limits
  - Conflict detection and tracking per cache bank
  - Detailed stats for utilization and conflicts

**Validation:**
Simulation results produced using this model are in line with real hardware observations, as documented in the following publication:

Michael Garrett Bechtel and Heechul Yun. Cache Bank-Aware Denial-of-Service Attacks on Multicore ARM Processors. IEEE Intl. Conference on Real-Time and Embedded Technology and Applications Symposium (RTAS), May 2023. [[paper](https://www.ittc.ku.edu/~heechul/papers/cachebank-rtas2023-camera.pdf)]

The figure below shows slowdown behavior under cache bank pressure, closely matching trends reported on real ARM hardware:

slowdown_combined_vs_solo-read

### DIFF
--- a/src/mem/cache/Cache.py
+++ b/src/mem/cache/Cache.py
@@ -90,6 +90,24 @@ class BaseCache(ClockedObject):
     data_latency = Param.Cycles("Data access latency")
     response_latency = Param.Cycles("Latency for the return path on a miss")
 
+    enable_banks = Param.Bool(
+        False, "Knob to control if the bank model is used"
+    )
+
+    num_banks = Param.Int(4, "Number of cache banks")
+
+    bank_intlv_high_bit = Param.Int(
+        0,
+        "Cache bank interleave highest bit "
+        "(0 = automatically aligned to cache line granularity)",
+    )
+
+    unlocked_tags = Param.Bool(
+        True, "Allow to check tags even if the bank is blocked"
+    )
+
+    write_latency = Param.Cycles(0, "Data write latency")
+
     warmup_percentage = Param.Percent(
         0, "Percentage of tags to be touched to warm up the cache"
     )
--- a/src/mem/cache/SConscript
+++ b/src/mem/cache/SConscript
@@ -52,6 +52,7 @@ Source('noncoherent_cache.cc')
 Source('write_queue.cc')
 Source('write_queue_entry.cc')
 
+DebugFlag('CacheBank')
 DebugFlag('Cache')
 DebugFlag('CacheComp')
 DebugFlag('CachePort')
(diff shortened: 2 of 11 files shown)


powerjg:
I haven't reviewed the code in detail, yet. However, I have one very high-level question. How is this different (and better) than just having N copies of the `Cache` object which have interleaved address ranges assigned?

amin-mamandi:
Hi Jason,

I would say this differs from having N instances of the cache object by centralizing metadata and statistics, allowing unified management of the shared caches (which also reduces redundancy). The main benefit is that it simplifies simulation and avoids the complexity of managing multiple instances.

powerjg:
Thank you for your contribution to gem5. What we need to understand better is the benefits of this change before we can review it and possibly merge it. The change is large (more than 1000 lines of code) and it changes a "core" part of gem5 (the cache models) that are used by a large fraction of the gem5 users. For a change that is this large and that touches a "popular" part of gem5, the benefits need to be large as well. The reason is that (1) the change could have unintended side effects and (2) the change will require maintenance *forever*.

Specifically, the benefit for this change should be something like "without this change, it is impossible to model <some behavior> in hardware accurately." 

> centralizing metadata and statistics, allowing unified management of the shared caches (which also reduces redundancy).

I don't fully understand this point... maybe you can further clarify. To me, centralizing the metadata is moving us away from a more realistic (to hardware) implementation which would *not* have centralized metadata in a banked cache. Furthermore, the stats are something that we should be able to deal with in a different way. If there's something about the current models that makes it impossible to model the correct timing of banking, then it's a different argument.

> The main benefit is that it simplifies simulation and avoids the complexity of managing multiple instances.

Personally, I would much prefer to keep the complexity in the Python configuration and make the underlying C++ models as simple (and parameterizable) as possible. However, if we must add complexity to the C++ models to get timing correctness, then that would be worth it.

All of this said, there is definitely one "bug" if you use multiple interleaved caches to model banking right now. The current system does not deal with the associative caches correctly for interleaved address ranges. See https://github.com/gem5/gem5/blob/stable/src/mem/cache/tags/indexing_policies/set_associative.cc#L55

Would it be possible to try to use your from-scratch banking implementation to test the performance of multiple interleaved caches to fix the (likely) bugs that we currently have?

amin-mamandi:
Hi Jason,

Thank you for your detailed feedback and for pointing out the key issues. I understand your concerns about the impact of this change on such a core part of gem5.

I’ll look into the issue you mentioned and test my implementation using interleaved addresses to see if it resolves the current bugs. I’ll share an update once I have results.

erin-le:
@amin-mamandi: Hello, could you provide an update on the status of this PR? Thank you!

amin-mamandi:
> @amin-mamandi: Hello, could you provide an update on the status of this PR? Thank you!
Hi, Thanks for checking in. I haven’t had a chance to work on it yet. Once I finalize the other approach, I’ll submit a new pull request.

erin-le:
@amin-mamandi: Hello, if you're planning to open another PR, would you be alright with closing this one?

amin-mamandi:
@erin-le yes! Please go ahead and close it.

Issue #1862, opened 2024-12-16
Implement PCIe architecture model


clemdiep:
The current implementation of PCI and PCIe in gem5 isn't accurately modeling the actual architecture. Only the configuration registers are implemented, and PCI devices are directly connected to the IO bus like any other devices, or through intermediate crossbars.

Some works have been made to implement a more accurate model of the PCIe architecture into gem5, but it has never been brought to gem5 mainline. Here are links to the original works:

  - Master thesis: https://www.ideals.illinois.edu/items/107666
  - Related paper: https://ieeexplore.ieee.org/abstract/document/8573496
  - Source code: https://github.com/Krishnaps/Gem5-PCI-Express
    - Most pertinent files are:
      - `src/mem/PciBridge.[py/cc/hh]`
      - `src/mem/pcie_link.[cc/hh]`

In summary, it adds a `PciBridge` class acting either as a Host-PCI or PCI-PCI bridge. It is then used to create the PCIe `Root_Complex` and `PciESwitch` Python class, which enables the creation of the PCIe architecture. It also adds the class `PCIELink` which acts as a PCIe endpoint and forward all packets to the actual device.

However, the implementation requires more work to bring the support into gem5 mainline. The main problems are:

  - The `PciBridge` class isn't scalable and require to have three PCIe devices connected to root complex and each switch.
  - Some declarations could be placed in already existing file.
  - Variables, structures, functions, etc. names can be revised to be more understandable.
  - There is code duplication that can be reduced.
  - Global code architecture can be revised.

We, the REDS team from HEIG-VD/HES-SO, are currently working on improving those points and aim to bring it into mainline when ready.

You can follow our progress on the following repo/branch: https://github.com/reds-heig/gem5/tree/pcie-architecture

We are open to any suggestion or requirement for this implementation.

The team:

  - Basile Cuneo @BasileCuneo
  - Clément Dieperink @clemdiep
  - Marina Zapater Sancho @marinazapater
  - Karan Pathak @karnapathak
  - Arzur Gabriel Catel Torres @EDGEMTech-GabrielC
  - Liu Qunyou @LIUQyou

abmerop:
This would be a great contribution! I will keep my eye on it.

I have to dedicate time to other work, but at the very least, I could help test by seeing if the multiple-GPU code I've been attempting on the side work with the changes.

PR #2204, opened 2025-04-24
dev: reworks PCI to add a PCI host bridge

clemdiep:
This commit reworks the PCI architecture to:

  - Add a PCI host bridge
  - Add the possibility to have different upstream between multiple `PciDevice` (for future PCI-PCI bridge implementation)
  - Moves the responsibility to response a PCI configuration request in `PciDevice` instead of `PciHost`.

For that, the `PciUpstream` class is added and works like the old `PciHost`, which now inherit from it. `PciDevice` will now need to register to a `PciUpstream` instead of `PciHost` to get an instance of `DeviceInterface`, this allows for the future implementation of a PCI-to-PCI bridge. Also, bus number is now given to the device by its upstream and not by the gem5 configuration which allow it to be changed by the OS on boot for PCI-PCI bridge.

The new `PciHost` implementation now has a member of the new class `PciHostBridge`. That class acts like a two-way bridge between the system bus and the PCI bus. It exposes all PCI address to the system side and should act as the default destination on the PCI side. The implementation is mostly taken from the already existing memory bridge, with some additions.

The python `PciXBar` class is also added, which for now simply inherit from `NoncoherentXBar`.

So, the new PCI architecture is composed of a `PciHost` with internal `PciHostBridge`, connected to a `PciXBar` on PCI side to which all `PciDevice` are connected.

PCI configuration request are now handled by each `PciDevice` instead of `PciHost` to mimic real PCI configuration.

This is another big step related to #1862.

It has an impact on existing gem5 configuration as now all PCI devices need to be connected behind a `PciHostBridge`. All configurations on the repo have been updated to account that.
However, I'm not sure to understand correctly systems using Ruby and how to connect on it, like with the GPU. For now, I connected PCI device's indirectly through the bridge to  Ruby, which works fine.

If needed, it would be possible to *revert* and also have a PCI host that doesn't require the bridge and the PCI bus, and so work like the current one. I haven't implemented it to have only one way of configuring the PCI, but let me know if you think otherwise.


abmerop (approved):
Great work!  I'm probably not qualified to do a detailed review on the PCI specifics but my tests are all passing and changes to the GPU code look minimal enough

giactra:
This is a big change I'd need some time to review it if you don't mind

clemdiep:
Due to mishandling on our original fork, which led us to delete it to start over a clean one, this PR has been repost to #2298 as it cannot be reopen. Sorry for the inconvenience.

PR #2207, opened 2025-04-24
misc: Increase Clang support to v19; Drop GCC v10 support

BobbyRBruce:
- Increases the supported Clang version to 19.
- Removes support for GCC v10, making the oldest supported version v11.
- Converts compiler version errors to warnings, thus allowing use of unsupported compilers with just a warning thrown.

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -101,7 +101,7 @@ jobs:
         # for compilation error to be exposed.
         runs-on: [self-hosted, linux, x64]
         if: github.event.pull_request.draft == false
-        container: ghcr.io/gem5/clang-version-18:latest
+        container: ghcr.io/gem5/clang-version-19:latest
         needs: [pre-commit]
         timeout-minutes: 90
         steps:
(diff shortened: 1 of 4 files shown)


BobbyRBruce:
I would've liked to increase to v20 but v19 is the highest you can get on Ubuntu 24.04 with out-of-the-box APT. Supporting the latest Ubuntu LTS release APT package library is normally what we aim for.

abmerop:
Do we have a list of the _minimum_ compiler versions gem5 supports?  If we do, I could not find it.  I'd love to use some features from newer compilers but it requires gcc 12+ and clang 15+ for example.

abmerop (dismissed):

BobbyRBruce:
I had a go at removing GCC 10 as it's quite old now, but gcn-gpu container requires gcc 10 so I'll keep it for now.

BobbyRBruce:
I've decided to drop v10 support. In addition to this change being added I also changed the errors thrown is using unsupported compiler versions to warnings. I think this is more apt (an older/newer compiler may well work,  it's ust that we don't support them)

PR #2210, opened 2025-04-25
configs,mem-ruby: Fix Directory_Controller fallback logic

BobbyRBruce:
This change corrects the instantiation of the Directory_Controller when
using Ruby protocols. The new behavior attempts to use the
protocol-specific controller class (e.g.,
`MI_example_Directory_Controller`) as required when running with
multiple protocols compiled (typically when using the ALL gem5 binary).
If the protocol-specific controller is not found (the case when working
with a binary compiled to a single protocol), the code falls back to
using the generic Directory_Controller.

This fix resolves an issue that caused the gem5 nightly tests to fail,
as seen in GitHub Actions Run #14650118937.

### DIFF
--- a/configs/ruby/Ruby.py
+++ b/configs/ruby/Ruby.py
@@ -310,7 +310,7 @@ def create_directories(options, bootmem, ruby_system, system):
     try:
         # The supported way to use Ruby is now to use the protocol name as
         # part of the names for all of the controllers. This is *required*
-        # when using `MULTIPLE` as the protocl and the `ALL` target.
+        # when using `MULTIPLE` as the protocol and the `ALL` target.
         Directory_Controller = getattr(
             importlib.import_module("m5.objects"),
             f"{options.protocol}_Directory_Controller",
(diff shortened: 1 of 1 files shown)


configs/ruby/Ruby.py
```diff
@@ -307,9 +307,13 @@ def create_system(
...
     import importlib
 
+    if "PROTOCOL" not in buildEnv:
+        raise RuntimeError(
+            "The 'PROTOCOL' key is missing in buildEnv. Please ensure it is set correctly."
+        )
     Directory_Controller = getattr(
         importlib.import_module("m5.objects"),
-        f"{options.protocol}_Directory_Controller",
+        f"{buildEnv['PROTOCOL']}_Directory_Controller",
```
giactra: Isn't this against the multi-ruby support? What happens to the PROTOCOL env var if gem5 is built with all protocols?

IMHO it's better to  check for options.protocol and only if this is not set rely on buildEnv...

BobbyRBruce: I've updated this with Jason's help. The new code now supports both the new way and the "old way" as a fallback when running a single-protocol build. So with `ALL` you must use `--protocol` but it's not necessary for cases (like in the case of the failing tests) where the binary contained a single protocol.

powerjg (approved):
I tested this will `ALL` and `NULL` and it worked for me. We should wait for Giacomo to sign off, though.

PR #2215, opened 2025-04-25
sim: convert entering event queue message to debug flag

erin-le:
This commit converts an inform(...) message for entering event queue and starting simulation to a debug flag. This prevents the message from cluttering simerr when the simulation is frequently started and stopped.


powerjg:
I'm supportive of this change, but others should take a look too.

@giactra @abmerop @mattsinc

BobbyRBruce (approved):
I don't see any issue with this. This is fine with me.

PR #2217, opened 2025-04-26
scons: Remove -Wno-vla-cxx-extension. gem5 is now VLA-clean

bpfoley:
Change-Id: I142b6dddbd88d48659fc22b1ee3f135d8f3d04f2

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -726,8 +726,6 @@ for variant_path in variant_paths:
         with gem5_scons.Configure(env) as conf:
             conf.CheckCxxFlag('-Wno-c99-designator')
             conf.CheckCxxFlag('-Wno-defaulted-function-deleted')
-        if compareVersions(env['CXXVERSION'], '18') > 0:
-            env.Append(CCFLAGS=['-Wno-vla-cxx-extension'])
 
         env.Append(TCMALLOC_CCFLAGS=['-fno-builtin'])
 


abmerop (changes requested):
Have you tried to build build/ALL and build/VEGA_X86? I believe there are still some VLAs in the GPU model's checkpointing code and that is not yet part of the ALL build:

https://github.com/gem5/gem5/blob/develop/src/dev/amdgpu/pm4_packet_processor.cc#L1132
https://github.com/gem5/gem5/blob/develop/src/dev/amdgpu/amdgpu_device.cc#L769

I do not have an easy way to test clang at the moment so I am not able to test this myself.

bpfoley:
Good point! No, I haven't. You mean there's more to the world than ARM? 😄  I'll try those and see if I missed anything.

abmerop:
Much appreciated!  I am traveling until Thursday but I could help out removing them after that (git blame says I added them 😨).

powerjg:
Thanks for taking this on, Brian! What's the current status of getting ALL to build?

bpfoley:
> Thanks for taking this on, Brian! What's the current status of getting ALL to build?

I think when the current batch of PRs land, ALL and VEGA_X86 will be clean. I'll rebase and update this PR when that happens.

bpfoley:
I've rebased against develop, and the build configs for ALL and X86_VEGA build without `-Wno-vla-cxx-extension`, so it looks like this is good to be committed now. Thanks for the feedback.

powerjg (approved):

abmerop (approved):

PR #2225, opened 2025-04-29
sim, util: changes to the external signal handler

erin-le:
This PR changes the signal handler for external gem5 orchestrator signals so that it matches the format of the other signal handlers. It was previously set up incorrectly, which occasionally caused seg faults when the hypercall_external_signal utility was used to communicate with gem5. This PR also changes the signal used for external signals from SIGHUP to SIGCONT to prevent gem5 processes from being killed if they receive a signal before the signal handlers are registered.

### DIFF
--- a/src/sim/init_signals.cc
+++ b/src/sim/init_signals.cc
@@ -233,7 +233,15 @@ static void
 externalProcessHandler(int sigtype)
 {
     async_event = true;
+    async_hypercall = true;
+    /* Wake up some event queue to handle event */
+    getEventQueue(0)->wakeup();
 
+}
+
+void
+processExternalSignal(void)
+{
     std::string shared_mem_name_str = "shared_gem5_signal_mem_" +
         std::to_string(getpid());
     const char* shared_mem_name = shared_mem_name_str.c_str();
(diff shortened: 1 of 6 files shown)


src/sim/init_signals.cc
```diff
@@ -381,7 +389,9 @@ externalProcessHandler(int sigtype)
     munmap(shm_ptr, shared_mem_size);
     close(shm_fd);
 
-    exitSimulationLoopNow(hypercall_id, payload_map);
+    // exitSimulationLoopNow(hypercall_id, payload_map);
+    exitSimLoopWithHypercall("Handling external signal!", 0, curTick(), 0,
+    payload_map, hypercall_id, false);
```
powerjg:
```suggestion
    exitSimLoopWithHypercall("Handling external signal!", 0, curTick(), 0,
                             payload_map, hypercall_id, false);
```

src/sim/simulate.cc
```diff
@@ -327,6 +327,10 @@ doSimLoop(EventQueue *eventq)
                 async_exception = false;
                 return NULL;
             }
+            if (async_hypercall){
```
powerjg:
```suggestion
            if (async_hypercall) {
```

powerjg (commented):
A couple of small style things (e.g., don't leave commented code). Otherwise this looks great. Thanks for finding these fixes!

powerjg (approved):

PR #2231, opened 2025-04-30
configs: Remove riscv/fs_linux.py deplicate --bootloader option

rogerchang23424:
Duplicate in `addFSOptions`

https://github.com/gem5/gem5/blob/13d116f3925e0437db29753918a4d87dedab1c96/configs/common/Options.py#L831-L835

### DIFF
--- a/configs/example/riscv/fs_linux.py
+++ b/configs/example/riscv/fs_linux.py
@@ -150,12 +150,6 @@ parser = argparse.ArgumentParser()
 Options.addCommonOptions(parser, ISA.RISCV)
 Options.addFSOptions(parser)
 parser.add_argument(
-    "--bootloader",
-    action="store",
-    type=str,
-    help="File that contains the bootloader",
-)
-parser.add_argument(
     "--virtio-rng", action="store_true", help="Enable VirtIORng device"
 )
 parser.add_argument(


powerjg (approved):

Navyera:
Hi! I encountered the following due to this change:

Since `bootloader` in Options.py is now a list (instead of a string) I am getting this behaviour with the `fs_linux.py` script.

gem5 command: 
`./build/RISCV/gem5.opt configs/example/riscv/fs_linux.py 
--bootloader=/home/nick/CAL/GEM5_RESOURCES/riscv/bootloaders/fw_jump.elf 
--kernel=/home/nick/CAL/GEM5_RESOURCES/riscv/kernels/linux6.7-vmlinux 
--disk=/home/nick/CAL/GEM5_RESOURCES/riscv/disks/riscv-disk.img`

Output:
...
src/base/loader/image_file_data.cc:105: fatal: fatal condition fd < 0 occurred: Failed to open file ['/home/nick/GEM5_RESOURCES/riscv/bootloaders/fw_jump.elf'].
This error typically occurs when the file path specified is incorrect.
Memory Usage: 170592 KBytes

You can see that the filename variable in C++ code has now taken the value of the string representation of the Python list which of course will never be the intended path.

Am I misusing something here?

PR #2234, opened 2025-04-30
dev-amdgpu: Don't use VLA in device (de)serialization

bpfoley:
Change-Id: I7378779498d8565948549adab10761fe0477b6af


powerjg (commented):
lgtm, but I'll let Matt P. approve.

abmerop (dismissed):
Thanks for putting this together!  It looks a lot cleaner in addition to removing the VLAs.

mattsinc (approved):
This looks fine but @v-ramadas before we merge can you please confirm this doesn't break any of your GPUFS support?

v-ramadas:
@bpfoley, @mattsinc, This fails at my end during restoration. I tried running [square](https://github.com/gem5/gem5-resources/blob/develop/src/gpu/square/square.cpp) (modified to create checkpoint after kernel execution). While the checkpoint gets created successfully, I see the following failure during restoration:
`src/sim/serialize.hh:446: fatal: fatal condition !cp.find(section, name, str) occurred: Can't unserialize 'system.pc.south_bridge.gpu:vmid_array.get()'.`

Inspecting the checkpoint file, I noticed that it contains `vmid_array=10 11` instead of `vmid_array.get()=10 11`

I've added comments on the lines that I believe are causing this issue.

src/dev/amdgpu/amdgpu_device.cc
```diff
@@ -804,20 +804,15 @@ AMDGPUDevice::serialize(CheckpointOut &cp) const
...
+    SERIALIZE_ARRAY(sdma_engs.get(), sdma_engs_size);
     // Save the vmids used in an array
-    SERIALIZE_ARRAY(used_vmids, sizeof(used_vmids)/sizeof(used_vmids[0]));
+    SERIALIZE_ARRAY(used_vmids.get(), used_vmid_map_size);
     // Save the size of the set of queue ids mapped to each vmid
-    SERIALIZE_ARRAY(used_queue_id_sizes,
-            sizeof(used_queue_id_sizes)/sizeof(used_queue_id_sizes[0]));
+    SERIALIZE_ARRAY(used_queue_id_sizes.get(), used_vmid_map_size);
     // Save all the queue ids used for all the vmids
     SERIALIZE_ARRAY(vmid_array, num_queue_id);
```
v-ramadas: vmid_array gets serialized here but vmid_array.get() is used in unserialize(). Should this also be changed to vmid_array.get()?

mattsinc:
> @bpfoley, @mattsinc, This fails at my end during restoration. I tried running [square](https://github.com/gem5/gem5-resources/blob/develop/src/gpu/square/square.cpp) (modified to create checkpoint after kernel execution). While the checkpoint gets created successfully, I see the following failure during restoration: `src/sim/serialize.hh:446: fatal: fatal condition !cp.find(section, name, str) occurred: Can't unserialize 'system.pc.south_bridge.gpu:vmid_array.get()'.`
> 
> Inspecting the checkpoint file, I noticed that it contains `vmid_array=10 11` instead of `vmid_array.get()=10 11`
> 
> I've added comments on the lines that I believe are causing this issue.

Thanks @v-ramadas !  @bpfoley can you please make this change?  Also, would it help if we point you to the tests we run for GPUFS (they aren't currently in the daily or weekly runs in gem5)?

mattsinc (changes requested):
Please make the update @v-ramadas asked for before we merge

bpfoley:
> > @bpfoley, @mattsinc, This fails at my end during restoration. I tried running [square](https://github.com/gem5/gem5-resources/blob/develop/src/gpu/square/square.cpp) (modified to create checkpoint after kernel execution). While the checkpoint gets created successfully, I see the following failure during restoration: `src/sim/serialize.hh:446: fatal: fatal condition !cp.find(section, name, str) occurred: Can't unserialize 'system.pc.south_bridge.gpu:vmid_array.get()'.`
> > Inspecting the checkpoint file, I noticed that it contains `vmid_array=10 11` instead of `vmid_array.get()=10 11`
> > I've added comments on the lines that I believe are causing this issue.
> 
> Thanks @v-ramadas ! @bpfoley can you please make this change? Also, would it help if we point you to the tests we run for GPUFS (they aren't currently in the daily or weekly runs in gem5)?

Absolutely it would help. I'm not yet at all familiar with the gem5 testing infrastructure, sadly.

I need to look closer at the `SERIALIZE_ARRAY` macros, because if they're stringifying the first argument and using that in the serialized output, then I've caused a backwards compatibility break by changing key names from `foo` to `foo.get()`. I'll write a macro to avoid this.

mattsinc:
> > > @bpfoley, @mattsinc, This fails at my end during restoration. I tried running [square](https://github.com/gem5/gem5-resources/blob/develop/src/gpu/square/square.cpp) (modified to create checkpoint after kernel execution). While the checkpoint gets created successfully, I see the following failure during restoration: `src/sim/serialize.hh:446: fatal: fatal condition !cp.find(section, name, str) occurred: Can't unserialize 'system.pc.south_bridge.gpu:vmid_array.get()'.`
> > > Inspecting the checkpoint file, I noticed that it contains `vmid_array=10 11` instead of `vmid_array.get()=10 11`
> > > I've added comments on the lines that I believe are causing this issue.
> > 
> > 
> > Thanks @v-ramadas ! @bpfoley can you please make this change? Also, would it help if we point you to the tests we run for GPUFS (they aren't currently in the daily or weekly runs in gem5)?
> 
> Absolutely it would help. I'm not yet at all familiar with the gem5 testing infrastructure, sadly.
> 
> I need to look closer at the `SERIALIZE_ARRAY` macros, because if they're stringifying the first argument and using that in the serialized output, then I've caused a backwards compatibility break by changing key names from `foo` to `foo.get()`. I'll write a macro to avoid this.

Testing framework is not your fault here -- this test is not yet in the tests that automatically run.

@abmerop , @v-ramadas which one of you has the GPUFS test to point @bpfoley to?

v-ramadas:
@bpfoley, [This](https://github.com/gem5-ISCA24-tutorial/gem5-bootcamp-env/blob/main/materials/isca24/10-gpufs/square-cpt/square.cpp) is a version of square modified to create a checkpoint after Linux boot but before the square kernel is launched. [This directory](https://github.com/gem5-ISCA24-tutorial/gem5-bootcamp-env/tree/main/materials/isca24/10-gpufs) contains a Makefile and commands needed to checkpoint/restore the application

bpfoley:
Yeah, as it turns out the `SERIALIZE_ARRAY` macros do stringify the array argument. I've added a new `SERIALIZE_UNIQUE_PTR_ARRAY` and used it consistently everywhere. This means the on-disk checkpoints should be exactly the same as they were before this patch (i.e., no extra `.get()` in the key names).

I wasn't able to easily build the GPUFS test, because I'm developing on an arm64 box at the moment, and I couldn't figure out exactly what's needed to get all of square to cross-compile. @v-ramadas  Would you be able to retest this patch for me please?

v-ramadas:
I am now able to checkpoint and restore successfully with the latest fixes. Thanks, @bpfoley.

src/dev/amdgpu/pm4_packet_processor.cc
```diff
@@ -1222,53 +1222,54 @@ PM4PacketProcessor::unserialize(CheckpointIn &cp)
...
+    auto queue = std::make_unique<uint32_t[]>(num_queues);
+    auto privileged = std::make_unique<bool[]>(num_queues);
+    auto queue_type = std::make_unique<uint32_t[]>(num_queues);
+    auto hqd_active = std::make_unique<uint32_t[]>(num_queues);
+    auto hqd_vmid = std::make_unique<uint32_t[]>(num_queues);
+    auto aql_rptr = std::make_unique<Addr[]>(num_queues);
+    auto aql = std::make_unique<uint32_t[]>(num_queues);
+    auto doorbell = std::make_unique<uint32_t[]>(num_queues);
+    auto hqd_pq_control = std::make_unique<uint32_t[]>(num_queues);
+
```
mattsinc, line 1248: extra line of space should be removed

src/dev/amdgpu/amdgpu_device.cc
```diff
@@ -846,16 +839,13 @@ AMDGPUDevice::unserialize(CheckpointIn &cp)
...
+        auto doorbells_queues = std::make_unique<QueueType[]>(doorbells_size);
+        auto doorbells_ip_ids = std::make_unique<int[]>(doorbells_size);
 
-        UNSERIALIZE_ARRAY(doorbells_offset, sizeof(doorbells_offset)/
-                sizeof(doorbells_offset[0]));
-        UNSERIALIZE_ARRAY(doorbells_queues, sizeof(doorbells_queues)/
-                sizeof(doorbells_queues[0]));
-        UNSERIALIZE_ARRAY(doorbells_ip_ids, sizeof(doorbells_ip_ids)/
-                sizeof(doorbells_ip_ids[0]));
+        UNSERIALIZE_UNIQUE_PTR_ARRAY(doorbells_offset, doorbells_size);
```
mattsinc, line 846: I am not sure I've totally convinced myself why doorbells_size is guaranteed to be the same as doorbells_offset / sizeof(doorbells_offset[0]) in all cases ... am I missing something?

(this same question applies in other places too, but just marking here)

mattsinc (changes requested):
Mostly minor things, thanks!

src/dev/amdgpu/amdgpu_device.cc
bpfoley: In C/C++, `sizeof(a)/sizeof(a[0])` is an idiom to get the number of elements in an array when its size is known by the compiler. In fact it's often defined as a macro. `sizeof(a)` evaluates to the size of the whole array in bytes, and `sizeof(a[0])` evaluates to the size of the first element in bytes.

On lines 842..845 we've allocated 3 `doorbells_foo` arrays of size `doorbells_size`.

mattsinc (approved):
Thanks!

PR #2235, opened 2025-04-30
arch-x86: Don't use non ISO C++ variable length arrays

bpfoley:
Remove uses of (stack allocated) variable length arrays. This isn't portable ISO C++ code, and clang 18 complains about it unless you switch off the warning.

### DIFF
--- a/src/arch/x86/bios/intelmp.cc
+++ b/src/arch/x86/bios/intelmp.cc
@@ -85,22 +85,17 @@ writeOutField(PortProxy& proxy, Addr addr, T val)
 uint8_t
 writeOutString(PortProxy& proxy, Addr addr, std::string str, int length)
 {
-    char cleanedString[length + 1];
-    cleanedString[length] = 0;
-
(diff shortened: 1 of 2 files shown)


src/arch/x86/bios/intelmp.cc
```diff
@@ -85,7 +85,7 @@ writeOutField(PortProxy& proxy, Addr addr, T val)
 uint8_t
 writeOutString(PortProxy& proxy, Addr addr, std::string str, int length)
 {
-    char cleanedString[length + 1];
+    auto cleanedString = static_cast<char *>(malloc(length + 1));
```
powerjg: This is a small thing, but could we use a string? Or, maybe better, a `std::vector<char>`?

I'd rather use C++ things and not malloc/free, but it's not a big deal either way.

bpfoley: ACK. Done.

powerjg (approved):

PR #2241, opened 2025-05-01
mem-ruby: Fix DMA sequencer request size above 64

### LINKED ISSUE #2218: Uncacheable request is split into several blocks, and assert failed except for the first block
**Describe the bug**
We want to write or read the same address to simulate our special hardware design, but assert failed if the size of packet is larger than 64; as far as we know, the default block size of ruby system is 64 bytes which is defined in src/mem/ruby/system/RubySystem.py as block_size_bytes, and requests size larger than this will split into several blocks. 
When the request reached DMA sequencer, before processing, it will be recorded in a unordered map(m_RequestTable), the key is address; according to gem5 tutorial, once hit wanted data, the function dataCallback in raised, and at the end of function dataCallback, is issueNext function, things becomes wired, the issueNext function set the second block's address to 0 + 64, which can not be found in m_RequestTable, result an assert error, all of this happens in dataCallback and issueNext function of src/mem/ruby/system/DMASequencer.cc.

**Affects version**
The version of gem5 is 24.1.0.2 and 4.8.1 of scons.

**gem5 Modifications**
We modified the behavior of DMA sequencer to support several request for the same address, which will not affect reproduction of this error.

**To Reproduce**
This error can be reproduced using our script. The command is 
```
./build/RISCV/gem5.opt ./src/reproduce/reproduce_dma.py
```

The error is 
```shell
Assertion failed: (i != m_RequestTable.end()), function dataCallback, file DMASequencer.cc, line 252.
```

**Host Operating System**
Mac OS X

Use following files to reproduce this error

[reproduce.zip](https://github.com/user-attachments/files/19929980/reproduce.zip)

powerjg:
Fixes #2218

Previously, when the DMA sequencer had a request for more than a block size it would split up the request, but didn't track outstanding requests correctly. This change adds a new map to track just those requests. If the original request is within 64 bytes, there are no changes.

### DIFF
--- a/src/mem/ruby/system/DMASequencer.cc
+++ b/src/mem/ruby/system/DMASequencer.cc
@@ -108,7 +108,7 @@ DMASequencer::makeRequest(PacketPtr pkt)
             return RequestStatus_Aliased;
     }
 
-    DPRINTF(RubyDma, "DMA req created: addr %p, len %d\n", line_addr, len);
+    DPRINTF(RubyDma, "DMA req created: addr %#llx, len %d\n", line_addr, len);
 
     int blk_size = m_ruby_system->getBlockSizeBytes();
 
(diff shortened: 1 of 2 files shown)


powerjg:
Disclosure: this was written by Claude. Below is the prompt I used. I take responsibility for the code (I reviewed and made some minor changes).

```
I'm seeing an error. Here's a debug trace and the error. The problem is the m_requestTable is monitoring the original base address not the block-level addresses. I think the intention is to have many outstanding requests at once, so we can't replace the original request in the table.

10000: system.ruby.network.ext_links0.ext_node.dma_sequencer: DMA req created: addr 0, len 1024
63000: system.ruby.network.ext_links0.ext_node.dma_sequencer: DMA data callback: addr 0
63000: system.ruby.network.ext_links0.ext_node.dma_sequencer: DMA request bytes issued 128, bytes completed 64, total len 1024 addr 0x40
116000: system.ruby.network.ext_links0.ext_node.dma_sequencer: DMA data callback: addr 0x40
gem5.opt: src/mem/ruby/system/DMASequencer.cc:231: void gem5::ruby::DMASequencer::dataCallback(const gem5::ruby::DataBlock&, const gem5::Addr&): Assertion `i != m_RequestTable.end()' failed.
```

mattsinc (approved):
Seems ok to me ...

PR #2242, opened 2025-05-02
dev-amdgpu: Remove SDMA header heap allocations

abmerop:
Some SDMA packets require headers. For those that do, a header struct is allocated on the heap and assigned to the header dword read during packet decoding. The heap allocation is unnecessary and has a second issue of assigning to a local variable that goes out of scope by the time the callback is called.

This commit simply passes the header as a dword to the callback methods which can then reinterpret the dword as a struct to access the fields needed. Fixes possibly bug related to the memory pointed to on the heap being recycled.


abmerop:
@v-ramadas FYI

mattsinc (approved):

PR #2243, opened 2025-05-02
arch-vega: Ignore EXEC SGPR dest for VOP3 V_CMPX_*

abmerop:
These instructions write *both* EXEC and either VCC or SGPR. Currently the special EXEC register is not being ignored, causing gem5 to think s126/s127 are user SGPRs that are out of range. This causes a panic.

The commit checks for VOP3 instructions writing EXEC (only V_CMPX_* instructions do this) and decrements the number of destinations that need to be added to the user GPR destinations for virtual to physical mapping.

Note that the actually write to EXEC happens by writing the wavefront class directly and is not done through SGPRs. VCC on the other hand is handled using SGPRs.


mattsinc (approved):
Wow, what a catch!

PR #2246, opened 2025-05-04
tests: Reduce the gpu_ruby_rand Daily tests to 1h

BobbyRBruce:
These tests were taking about 12 hours to run on our GitHub Action runners. This commit reduces this to around an hour by dividing the input size by 10.

.


mattsinc (approved):
I believe @abmerop ok'd this with your other commit, so I'll approve here too

PR #2249, opened 2025-05-05
resources: return a copy of resource JSON to avoid mutations

Harshil2107:
- Updated `get_resources()` in `JSONClient` to return a deep copy of the resource dictionary instead of the original reference.
- This prevents `obtain_resources()` from unintentionally modifying the original resource data.


powerjg (commented):
lgtm. @erin-le can you test this with your code and approve if it works?

erin-le (approved):

PR #2250, opened 2025-05-05
util: add dashboard for viewing gem5 simulation progress

erin-le:
This PR adds a utility for viewing the progress of gem5 simulations, based on the number of instructions executed compared to the total number of instructions for a workload, which is included in the workload's JSON. It also shows the simulated time based on the tick of the simulation.

The base code for the utility was generated by powerjg using Claude, then tested and modified.
This PR requires https://github.com/gem5/gem5/pull/2249 to run.

### DIFF
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -874,6 +874,15 @@ class WorkloadResource(AbstractResource):
         self._id = id
         self._func = function
         self._params = parameters
+        try:
+            self._estimated_instructions = int(
+                kwargs.get("simulation_run_results")["total_instructions"]
+            )
+        except:
+            self._estimated_instructions = None
+
+    def get_estimated_instructions(self) -> Optional[int]:
+        return self._estimated_instructions
 
     def get_id(self) -> str:
         """Returns the ID of the workload."""
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -875,14 +875,14 @@ class WorkloadResource(AbstractResource):
         self._func = function
         self._params = parameters
         try:
-            self._estimated_instructions = int(
+            self._estimated_total_instructions = int(
                 kwargs.get("simulation_run_results")["total_instructions"]
             )
         except:
-            self._estimated_instructions = None
+            self._estimated_total_instructions = None
 
-    def get_estimated_instructions(self) -> Optional[int]:
-        return self._estimated_instructions
+    def get_estimated_total_instructions(self) -> Optional[int]:
+        return self._estimated_total_instructions
 
     def get_id(self) -> str:
         """Returns the ID of the workload."""
--- a/util/hypercall_external_signal/gem5-dashboard.py
+++ b/util/hypercall_external_signal/gem5-dashboard.py
@@ -202,12 +202,14 @@ class Gem5ProgressBar:
             )
 
             workload = obtain_resource(curr_info["workload"])
-            self.max_insts = workload.get_estimated_instructions() or 0
+            self.max_insts = workload.get_estimated_total_instructions() or 0
             logger.info(
                 f"Max instructions for workload {workload} is "
                 f"{self.max_insts}"
             )
-            self.current_insts = curr_info["instruction_count"]
+            self.current_insts_executed = curr_info[
+                "curr_instructions_executed"
+            ]
             self.sim_id = curr_info["sim_id"]
             self.workload_id = curr_info["workload"]
             self.current_ticks = curr_info["tick"]
@@ -216,7 +218,7 @@ class Gem5ProgressBar:
             desc_format = f"{self.pid}|{self.sim_id}|{self.workload_id}"
             self.bar = tqdm.tqdm(
                 total=self.max_insts,
-                initial=self.current_insts,
+                initial=self.current_insts_executed,
                 desc=desc_format,
                 unit="insts",
                 position=self.position,
@@ -279,11 +281,13 @@ class Gem5ProgressBar:
             )
 
             # Update values
-            self.current_insts = curr_info["instruction_count"]
+            self.current_insts_executed = curr_info[
+                "curr_instructions_executed"
+            ]
             self.current_ticks = curr_info["tick"]
 
             # Update the progress bar
-            self.bar.n = self.current_insts
+            self.bar.n = self.current_insts_executed
             self.bar.set_postfix(
                 {"SimTime": format_ticks_time(self.current_ticks)},
                 refresh=True,
--- a/util/hypercall_external_signal/gem5-dashboard.py
+++ b/util/hypercall_external_signal/gem5-dashboard.py
@@ -1,3 +1,29 @@
+# Copyright (c) 2025 The Regents of the University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
 import argparse
 import json
 import logging
@@ -18,7 +44,10 @@ from orchestrator_request import (
     send_and_receive_hypercall,
 )
 
-from gem5.resources.resource import obtain_resource
+try:
+    from gem5.resources.resource import obtain_resource
+except:
+    pass
 
 logger = logging.getLogger(__name__)
 
@@ -96,13 +125,8 @@ def find_gem5_pids() -> List[int]:
                 # Skip processes we can't read
                 continue
 
-    # Mac OS doesn't have /proc, so we need a different approach
+    # macOS doesn't have /proc, so we need a different approach
     elif platform.system() == "Darwin":
-        # Get processes for current user
-        user_processes = subprocess.run(
-            ["ps", "-U", str(current_user)], capture_output=True
-        ).stdout.decode("UTF-8")
-
         # Filter for gem5 processes
         gem5_pids_proc = subprocess.run(["pgrep", "gem5"], capture_output=True)
 
@@ -118,7 +142,6 @@ def find_gem5_pids() -> List[int]:
                 if is_valid_gem5_process(pid, cmdline):
                     gem5_pids.append(int(pid))
 
-    # Add support for other platforms here if needed
     else:
         logger.error(f"Unsupported platform: {platform.system()}")
 
@@ -129,9 +152,10 @@ def find_gem5_pids() -> List[int]:
 
 
 def format_ticks_time(ticks):
-    """Convert tick count to human-readable time format"""
-    # Define tick to time unit conversions (example values - adjust based on gem5's tick rate)
-    # Assuming 1 tick = 1 picosecond for gem5
+    """
+    Convert tick count to human-readable time format.
+    1 tick = 1 picosecond.
+    """
     if ticks < 1_000:
         return f"{ticks}ps"
     elif ticks < 1_000_000:
@@ -142,20 +166,6 @@ def format_ticks_time(ticks):
         return f"{ticks/1_000_000_000:.2f}ms"
     else:
         return f"{ticks/1_000_000_000_000:.2f}s"
-
-
-def format_instruction_count(count):
-    """Format instruction count with appropriate suffix (K, M, B, T)"""
-    if count < 1_000:
-        return f"{count}"
-    elif count < 1_000_000:
-        return f"{count/1_000:.4g}K"
-    elif count < 1_000_000_000:
-        return f"{count/1_000_000:.4g}M"
-    elif count < 1_000_000_000_000:
-        return f"{count/1_000_000_000:.4g}B"
-    else:
-        return f"{count/1_000_000_000_000:.4g}T"
 
 
 class Gem5ProgressBar:
@@ -187,18 +197,19 @@ class Gem5ProgressBar:
                 send_and_receive_hypercall(self.pid, "status")
             )
 
-            # Get workload information
             workload = obtain_resource(curr_info["workload"])
             self.max_insts = workload.get_estimated_instructions() or 0
-
-            # Store simulation properties
+            logger.info(
+                f"Max instructions for workload {workload} is "
+                f"{self.max_insts}"
+            )
             self.current_insts = curr_info["instruction_count"]
             self.sim_id = curr_info["sim_id"]
             self.workload_id = curr_info["workload"]
             self.current_ticks = curr_info["tick"]
 
             # Create the actual tqdm progress bar
-            desc_format = f"{self.sim_id}|{self.workload_id}"
+            desc_format = f"{self.pid}|{self.sim_id}|{self.workload_id}"
             self.bar = tqdm.tqdm(
                 total=self.max_insts,
                 initial=self.current_insts,
@@ -206,6 +217,7 @@ class Gem5ProgressBar:
                 unit="insts",
                 position=self.position,
                 leave=True,
+                dynamic_ncols=True,
             )
 
             # Initial postfix update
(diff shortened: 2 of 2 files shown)


src/python/gem5/resources/resource.py
```diff
@@ -874,6 +874,15 @@ def __init__(
         self._id = id
         self._func = function
         self._params = parameters
+        try:
+            self._estimated_instructions = int(
+                kwargs.get("simulation_run_results")["total_instructions"]
+            )
+        except:
+            self._estimated_instructions = None
```
BobbyRBruce: We shouldn't use kwargs to add new stuff like this. It can get out of hand. 

Given the necessity for backwards compatibly and that this data is hard to get, I propose adding this as an optional int (unsigned) property with a default value of `None`, which can be set through the constructor or after (in case for  the user wishes to amend/alter this resource record with their own information which is a necessity when we don't have the instruction exec count in the gem5-resource record and they do) .

Something like this: 
````py

__init__(
  self,
  ...
  inst_exec_count : Optional[int] = None
):
  ...
  self.inst_exec_count = inst_exec_count

@property 
def inst_exec_count (self):
  """The total number of instructions that this workload will execute prior to completion. May be used as an indicator of progress towards simulation completion in circumstances where the simulation's instruction count from start is known and the instruction execution rate across the simulation is understood"""
  return self._inst_exec_count

@property.setter
def inst_exec_count(self, value):
   if not isinstance(value,Optional[int])
      raise ...
   if value is not None and value < 0:
      raise ...
   return self._inst_exec_count =value
````

File: util/hypercall_external_signal/gem5-dashboard.py [hunk 2]
```diff
@@ -0,0 +1,488 @@
...
+    - Are part of gem5's multisim infrastructure
+    - Cannot receive signals (permission error)
+
+    :return: List of valid gem5 process IDs
+    :raises ValueError: If no gem5 process found
+    """
+    gem5_pids = []
+    current_user = os.getuid()
+
+    def is_valid_gem5_process(pid_str, cmdline=""):
```
BobbyRBruce, line 75: I have a different way to do this:

1. Set the process name to be 'gem5'. This is possible in the C++ code.  At the start of the program, in the main method, use `prctl(PR_SET_NAME,"gem5\0",NULL,NULL,NULL); ` This allows us a very basic way to filter processes down to ones that are just instances of gem5.
2. Similar to the PR name, set an environment variable method to specify the gem5 version. This can be done with `setenv`. E.g.," `setenv(GEM5_VERSION, "v23.1.0.1")`. This allows us a quick way to inspect a gem5 process for its version (` /proc/{PID}/environ` is used to see a process's env vars. In our use case, this would allow us to know whether it is a gem5 version which can support our monitoring.
3. Set another environment variable called `ALLOW_EXTERN_HYPERCALLS` which is True or False depending on whether the gem5 process is open to processing signals or not. This step is the final check to ensure a gem5 process should be sent signals. In the case of multisim, this can be set to "False" for the parent process and "True" for the children.

util/hypercall_external_signal/gem5-dashboard.py
```diff
@@ -0,0 +1,488 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        "--pid",
+        help="Enter the pid(s) of the gem5 process(es) to display. "
+        "If entering multiple pids, please separate pids using spaces, "
+        "e.g. --pid 12345 12346",
```
BobbyRBruce: This help message should explain that if no PIDs are passed we will automatically discover the gem5 processes

util/hypercall_external_signal/transmitter.py
```diff
@@ -1,6 +1,4 @@
-#!/usr/bin/env python3
```
BobbyRBruce: Keep this, it allows for directly inning the side as I like an executable

util/hypercall_external_signal/transmitter.py
```diff
@@ -1,6 +1,4 @@
...
-
-# Copyright (c) 2023 The Regents of the University of California
+# Copyright (c) 2025 The Regents of the University of California
```
BobbyRBruce: Shouldn't this be 2024-2025?

util/hypercall_external_signal/gem5-dashboard.py
```diff
@@ -0,0 +1,488 @@
...
+
+    def _initialize(self):
+        """Initialize the progress bar with simulation data."""
+        try:
+            # Get initial simulation status
+            curr_info = json.loads(
+                send_and_receive_hypercall(self.pid, "status")
+            )
+
+            workload = obtain_resource(curr_info["workload"])
```
BobbyRBruce, line 209: This is unnecessary. The gem5 process already has this information and, as such, the workload information can be obtained via a hypercall request. It is also potentially problematic that the workload may not be obtainable via `obtain_resource` as in the case a user creates a workload locally in a config script.

With this removal, we can decouple ourselves in the gem5 binary, which I would prefer at this stage as it makes building considerably easier and allows for this to be run as an independent tool.

util/hypercall_external_signal/gem5-dashboard.py
```diff
@@ -0,0 +1,488 @@
...
+import subprocess
+import time
+from typing import (
+    Any,
+    Dict,
+    List,
+    Optional,
+)
+
+import tqdm
```
BobbyRBruce, line 46: Having thought about this, I think we should move away from having progress bars. 

First of all, this PR has become very progress bar-centric, and I think we've lost track of what we originally wanted. The dashboard was to be a place you could get key pieces of Gem5 information and use them to quickly diagnose progress, failure, etc. Over time, it could grow to be more sophisticated (like monitoring processes for crashes and restarting them from "saves" of simulation, for instance), but for a v1, just a basic readout of information is fine.


Progress bars, I fear, are going to be misleading to users and turn the dashboard into a "completion board" which is just used to see how long things are going to take. It's important to keep in mind that {num instructions executed}/{total instructions to be executed} can be very misleading. 50% of instructions executed doesn't mean you're halfway through a simulation in terms of execution time. It may not even be close. Many simulations will move between detailed and non-detailed simulation moves (e.g., KVM to Timing CPU cores), meaning we could easily have progress bars jump rapidly to certain points(when running with low-detailed simulations, only to halt for long periods when running in detailed simulations).




I think here we should just have a dynamic terminal table that lists these data per simulation in a raw manner: 

* the number of ticks executed
* The simulated seconds.
* Time since simulation started
* the number of instructions executed
* the number of instructions to be executed
* the instruction rate (which we infer by taking the number of instructions since I last took and the time between now and then)
* The stats output directory
* config file used and parameters passed 


I think this works like the tool `top`. A big table that gives you the current readout of the running Gem5 processes with  key data (and marking ones that completed as "Done" ). I don't think we should see it as a time-to-completion tool. It's a tool to monitor the essential high-level simulation information.

From this, we can add some things like warnings if a process has spent a long time without processing instructions (possible stall?).

util/hypercall_external_signal/gem5-dashboard.py
```diff
@@ -0,0 +1,488 @@
...
+                send_and_receive_hypercall(self.pid, "status")
+            )
+
+            workload = obtain_resource(curr_info["workload"])
+            self.max_insts = workload.get_estimated_instructions() or 0
+            logger.info(
+                f"Max instructions for workload {workload} is "
+                f"{self.max_insts}"
+            )
+            self.current_insts = curr_info["instruction_count"]
```
BobbyRBruce: [Please have this comment apply to all instances in which it is acceptable.]

I personally find phrases like 'instruction count' to be ambiguous, especially referring to workloads. Perhaps it is just me, but I get easily confused about whether we are referring to the number of instructions in a binary, the number of instructions that are executed for a particular input, or the number of instructions executed up to  this point in time. I would appreciate some more precise naming terminology here.

util/hypercall_external_signal/gem5-dashboard.py
```diff
@@ -0,0 +1,488 @@
...
+            self.sim_id = curr_info["sim_id"]
+            self.workload_id = curr_info["workload"]
+            self.current_ticks = curr_info["tick"]
+
+            # Create the actual tqdm progress bar
+            desc_format = f"{self.pid}|{self.sim_id}|{self.workload_id}"
```
BobbyRBruce, line 224: Related to my previous comment about using process names as useful process identifiers, we could just use Process names instead of introducing SimIDs as a separate thing. When creating processes, we can set a str to postfix . In multisim we could have the process name for the parent gem5 process could be "gem5 (parent)" and the process ID for the children is "gem5 child (<user-specified str>)". 

Process names do have to be under 15 characters total, which is tight, but I do prefer using something standard over making up our own thing here.  The process name is also obtainable through traditional process inspection tooling and thus wouldn't require passing and interpreting through our signal-sharedmemory-hypercall setup.

util/hypercall_external_signal/gem5-dashboard.py
```diff
@@ -0,0 +1,488 @@
...
+        # If bar isn't initialized yet, retry initialization
+        if self.bar is None:
+            self._initialize()
+            # Still initializing, return True to keep trying
+            return self.active
+
+        try:
+            # Get current simulation status
+            curr_info = json.loads(
+                send_and_receive_hypercall(self.pid, "status")
```
BobbyRBruce, line 290: This can be somewhat of a separate conversation as this isn't related to this PR exactly, but I don't know whether I want all this wrapped up in one hypercall. It feels bloated and inflexible. It feels like this will be open to abuse with people expanding it to add more data and, in the end, turning it into a bloated mess. I'm somewhat intrigued by the idea that maybe we can just have a hypercall for each of these properties we want to obtain, or maybe a flexible hypercall in which we state what data we want to obtain via the in its payload.

util/hypercall_external_signal/gem5-dashboard.py
```diff
@@ -0,0 +1,488 @@
...
+                    if not bar.update():
+                        # If update failed, bar is now inactive
+                        # Clear the dashboard so remaining bars can be
+                        # rearranged and printed cleanly
+                        self._print_dashboard_label()
+
+                # fill in gaps from closed bars if necessary
+                self._rearrange_positions()
+
+        except KeyboardInterrupt:
```
BobbyRBruce: Can we also process a ESC key press to exit the program?

util/hypercall_external_signal/gem5-dashboard.py
```diff
@@ -0,0 +1,488 @@
...
+        return f"{ticks/1000:.2f}ns"
+    elif ticks < 1_000_000_000:
+        return f"{ticks/1_000_000:.2f}μs"
+    elif ticks < 1_000_000_000_000:
+        return f"{ticks/1_000_000_000:.2f}ms"
+    else:
+        return f"{ticks/1_000_000_000_000:.2f}s"
+
+
+class Gem5ProgressBar:
```
BobbyRBruce, line 180: I think we should envision this as a more genetic holder of gem5 simulation information than _just_ a progress bar.

util/hypercall_external_signal/gem5-dashboard.py
```diff
@@ -0,0 +1,488 @@
...
+    if not gem5_pids:
+        raise ValueError("No valid gem5 process found")
+
+    return gem5_pids
+
+
+def format_ticks_time(ticks):
+    """
+    Convert tick count to human-readable time format.
+    1 tick = 1 picosecond.
```
BobbyRBruce: This is a nitpick as it's rarely ever changed, but you shouldn't assume 1 tick to 1 picosecond is fixed and true for all Gem5 processes. It is a configurable part of Gem5. You should implement some way of obtaining this information from a running process (or just have Gem5 give you this in seconds and leave it the responsibility of conversion).

src/python/gem5/resources/resource.py
```diff
@@ -874,6 +874,15 @@ def __init__(
...
+
+    def get_estimated_instructions(self) -> Optional[str]:
```
BobbyRBruce: Why does this return a string?

src/python/gem5/resources/resource.py
powerjg: I'm not sure I understand how this suggestion will work when instantiating this from the database. Could you explain how this would work with the client API?

util/hypercall_external_signal/gem5-dashboard.py
powerjg: I like the idea of 2 & 3 (though it may not work on Mac?)

For idea 1, isn't that going to screw up all sorts of other Linux things (e.g., top).

In general, I think that automatically finding the gem5 processes is not the most important aspect of this change right now. If we were able to merge something that worked for *1* gem5 process, that would be a good first step.

util/hypercall_external_signal/gem5-dashboard.py
powerjg: I am pushing back *very* hard here. This was my idea, and I'm going to die on this hill.

This allows us to not have to change the gem5 binary if we want different information about the workload. I want to keep the hypercall interface as thin as possible. I don't want to have to change the hypercall interface every time we think of something new to add (e.g., cycles, queries, iterations).

util/hypercall_external_signal/gem5-dashboard.py
powerjg: Maybe "q" since that is how top works?

util/hypercall_external_signal/gem5-dashboard.py
powerjg: I fear you and I (Bobby and Jason) are making Erin run in circles here. You clearly have a different vision than I do. I won't make a judgement call now, but whatever we do *it should be in consultation with the people who are actually using gem5*. We should not build something that isn't tied to how it's going to be used.

src/python/gem5/resources/resource.py
BobbyRBruce: The client retrieves the gem5 resource's entry from the database and passes the key-value pairs as parameters to the right class constructor. There are some special corner cases for this, but broadly we try to make a database entry map to map cleanly to the object we construct from it.

So an entry like:

```
id: example
version: 1.0.0
category: BinaryResource
url: https://where.to.download.org/file
md5sum: aaaaabbbccc
Architecture: X86
...
```

is downloaded. It's just a dictionary, let's call it `resource_dict`.  We then do `resource_class = _get_resource_json_type_map[(resource_dict[category])(**resource_dict)`.

In this example, `_get_resource_json_type_map` would return the `BinaryResource` class, which looks like this:

```py
class BinaryResource(FileResource):
    def __init__(self, url, md5sum, architecture, ..., **kwargs):
        self.url = url
        self.md5sum = md5sum
        self.architecture = architecture
       ...
```

The gem5 library would then call `resource = resource_class(**resource_dict)` and use it as an encapsulation of the gem5-resource data from then onwards.


If we add a new field resources in the database, nothing goes wrong for older versions of gem5. This data is simply captured by the **kwargs and sits there. No complaints. In newer versions of gem5, this data will be set if available or reset to the default None value if not.

The use of `@property` is something I wish I thought of sooner, as it provides a superior way of having getters and setters, but here it's mostly used to ensure the data is valid. I would hope for the database this is always true; these error checking here is mostly to warn the user if they choose to override this or create their own Workload.

powerjg: Adding the `@property` functions are orthogonal to this change though, right? Suggestion you're making is more that it should be an optional parameter to the constructor instead of using a try/catch

util/hypercall_external_signal/gem5-dashboard.py
BobbyRBruce: Actually you're right about 1. I have done it before but I realize it may need elevated permissions in some cases so best not risk that.

However I think using the `setenv` (like in 2) could be a way for us to flag processes as being gem5.

src/python/gem5/resources/resource.py
BobbyRBruce: yes. Sorry, it was probably confusing to use it in this example.

util/hypercall_external_signal/transmitter.py
erin-le: It looks like transmitter.py was committed and merged into develop in February 2025, but it was part of PR https://github.com/gem5/gem5/pull/947, which was opened in 2024. What time/point in development does the copyright message start from?

util/hypercall_external_signal/gem5-dashboard.py
BobbyRBruce: I think they are various ways to future proof this. For example we could have a hypercall that that just returns us a json serialization of the workload. And we can  use that and if we need any more workload data we've got all of it. I think we should also have one that gets the config.json, and one that we can use to get status on the fly. With that I think we can get all the info we'll ever need.

util/hypercall_external_signal/gem5-dashboard.py
erin-le: I think it could be more useful to keep `sim_id`, as this is the id that is passed into Simulator objects and used in multisim to determine directory names within `m5out`. It gives the user more information, as users can separately access the process name using the PID, and is more configurable, especially if point (1) from one of the previous comments isn't feasible.

util/hypercall_external_signal/gem5-dashboard.py
erin-le: I added an implementation for this in the latest commit (https://github.com/gem5/gem5/pull/2250/commits/36b84422a3c16e8bc7b1a86c7da996d5e15552d3), though it does take longer to exit compared to only CTRL-C and KeyboardInterrupt. It might not be the best solution, so please let me know your feedback on this.

util/hypercall_external_signal/gem5-dashboard.py
erin-le: The following commit should address this: https://github.com/gem5/gem5/pull/2250/commits/bb5c00dd95069dd80b127d930f2de94f342f4f66. If everything looks good, I'll rebase the changes into the previous commits.

util/hypercall_external_signal/gem5-dashboard.py
erin-le: The following commit should address this: https://github.com/gem5/gem5/pull/2250/commits/d92226fb32737de7299d15958e0b561cba41f550. If everything looks good, I'll rebase this commit into the previous commits.

src/python/gem5/resources/resource.py
```diff
@@ -874,6 +874,15 @@ def __init__(
...
         self._params = parameters
+        try:
+            self._estimated_total_instructions = int(
+                kwargs.get("simulation_run_results")["total_instructions"]
+            )
+        except:
+            self._estimated_total_instructions = None
+
+    def get_estimated_total_instructions(self) -> Optional[int]:
+        return self._estimated_total_instructions
```
powerjg, line 885: I think this is the right solution. While I have some mixed ideas on the "simulation_run_results" in the `Workload` resource, I think that *the number of instructions* should be intrinsic to the workload and is appropriate to have in the resource.

util/hypercall_external_signal/gem5-dashboard.py
powerjg: Coming back to this... having a progress-bar based dashboard hurts nothing. It's self-contained code. Let's merge this and if @BobbyRBruce has different/better ideas we can merge those ideas too.

util/hypercall_external_signal/gem5-dashboard.py
powerjg: Let's do that in another script.

util/hypercall_external_signal/gem5-dashboard.py
erin-le: Okay, I can extract the changes to files other than `gem5-dashboard.py` into different PRs, as discussed previously.

util/hypercall_external_signal/orchestrator_request.py
erin-le, line 1: Changes in this file have been copied over to https://github.com/gem5/gem5/pull/2470.

util/hypercall_external_signal/transmitter.py
erin-le, line 1: Changes in this file have been copied over to https://github.com/gem5/gem5/pull/2470.

src/python/gem5/resources/resource.py
erin-le, line 1: Changes to this file have been copied to https://github.com/gem5/gem5/pull/2471.

src/python/gem5/simulate/exit_handler.py
erin-le, line 1: Changes in this file have been copied over to https://github.com/gem5/gem5/pull/2472.

erin-le:
Closing in favor of https://github.com/gem5/gem5/pull/2900.

PR #2252, opened 2025-05-06
python: Add OptionalParam to support std::optional.

dl8sd11:
This commit introduces `OptionalParam` to generate `std::optional` parameters. This allows gem5 developer to define an optional parameter without using magic default numbers.

For example, if we want a parameter to overwrite sid. We used to have two parameters for this:

```
sid = Param.UInt32(0, 'SID')
set_ssid = Param.Bool(False, 'Whether to set the SSID')
```

This caused invalid states to handle, for example what if sid is set but set_ssid remains default False.

After this commit, we are able to solve this problem:

```
sid = OptionalParam.UInt32('SID')
```

On the C++ side, the parameter map to a `std::optional<uint32_t>` member. We eliminate the undesired states.

Limitation:

OptionalParam only supports primitive types currently. We use nullptr to express the emptiness for SimObjects.

Change-Id: I7e13231448862af8c03eb06e0c9e53e0a82b2efd


powerjg (commented):
Cool change! I like it. I'll let Giacomo have the sign-off, though.

giactra (approved):
Nice! I actually had an identical contribution in my local machine (forgot to push it).
This is a great improvement for non SimObject parameters. I believe we should start to cleanup gem5 from reserved values (e.g. -1 and so on)

> Limitation:
> OptionalParam only supports primitive types currently. We use nullptr to express the emptiness for SimObjects.

Thanks for reporting this. I believe it is a "fake" limitation as for a SimObject we would use NULL anyway

powerjg:
> I believe we should start to cleanup gem5 from reserved values (e.g. -1 and so on)

Agreed!

dl8sd11:
Thanks for the feedback. I will help migrate the reserved values to adopt the optional parameter if I run into them.

PR #2255, opened 2025-05-06
mem-ruby: fix wired deadlocks when running ruby_random_test with -n > 18

### LINKED ISSUE #2164: Weird deadlocks with -n > 18 in ruby_random_test.py.
**Describe the bug**
I've been using ruby_random_test.py to test my own protocol. Everything works fine with -n <= 16 and --maxloads=50,000,000, but when I increase -n to 32, it hits a deadlock.

At first, I suspected a bug in my protocol and enabled ProtocolTrace and RubySlicc, but found nothing wrong. Then, with RubyQueue, I noticed one processor/node becomes idle right after initialization and is marked as deadlocked after 50,000 Cycles— but it hasn't done anything yet.

This seems to be a bug in RubyTester. I confirmed the same issue using build/X86/gem5.opt — the deadlock appears when -n is greater than 18.

**Affects version**
My gem5 version is 24.1.0.1

**gem5 Modifications**
No changes except my own protocol sm files.

**To Reproduce**
1. Compile gem5 with command: scons -j$(nproc) build/X86/gem5.opt 
2. Execute the simulation: build/X86/gem5.opt --debug-flags=RubyQueue --debug-file=trace_q.log configs/example/ruby_random_test.py -n19 --maxloads=100

**Terminal Output**

```shell
src/cpu/testers/rubytest/RubyTester.cc:271: panic: Deadlock detected: current_time: 50001 last_progress_time: 0 difference:  50001 processor: 2
```

**Host Operating System**
Ubuntu 22.04

**Additional information**
If you search for l1_cntrl2 in the trace_q.log file, you will find that it has only 7 entries related to initialization.

Mayyygic:
Fix bug described in issue: #2164

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,5 +3,9 @@
         "src/python",
         "ext",
         "tests"
-    ]
+    ],
+    "files.associations": {
+        "*.cc": "cpp",
+        "iosfwd": "cpp"
+    }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,9 +3,5 @@
         "src/python",
         "ext",
         "tests"
-    ],
-    "files.associations": {
-        "*.cc": "cpp",
-        "iosfwd": "cpp"
-    }
+    ]
 }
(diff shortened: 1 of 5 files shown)


powerjg (approved):
This lgtm. Thanks for fixing this!!

powerjg:
When merging, this needs to be squashed and the commit message should be updated.

PR #2256, opened 2025-05-06
configs: Enable bootloader argument for RISCV simulations

tommarin:
A recent patch (PR #2231) removed the bootloader argument from `configs/example/riscv/fs_linux.py` due to the entry already being present in `Options.py`. However, the entry is under an `if buildEnv["USE_ARM_ISA"]` condition, making it unavailable for RISCV. This patch moves the argument to a different position in the `Options.py` file to enable it for RISCV simulations too.

### DIFF
--- a/configs/common/Options.py
+++ b/configs/common/Options.py
@@ -808,6 +808,11 @@ def addFSOptions(parser):
             help="Specifies device tree blob file to use with device-tree-"
             "enabled kernels",
         )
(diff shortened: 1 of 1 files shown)


rogerchang23424 (approved):
Thanks for fixing.

powerjg (approved):
What a mess....  We need to deprecate/delete Options.py 😢

PR #2263, opened 2025-05-08
arch-vega,gpu-compute: Opcode overrides based on gfx version

abmerop:
Some newer gfx9 ISAs reuse opcodes from previous generations to implement completely different instructions. In order to support this, make the decoder object take the gfx version from the shader and override specific opcodes rather than duplicate the entire decoder table.

Implement the new V_FMAC_F64 instruction which replaces the old V_MUL_LEGACY_F32.

### DIFF
--- a/src/arch/amdgpu/vega/gpu_decoder.cc
+++ b/src/arch/amdgpu/vega/gpu_decoder.cc
@@ -50,6 +50,26 @@ namespace VegaISA
     {
     } // ~Decoder
 
+    void Decoder::fixupGfx90a()
+    {
+        tableDecodePrimary[16] = &Decoder::decode_OP_VOP2__V_FMAC_F64;
+        tableDecodePrimary[17] = &Decoder::decode_OP_VOP2__V_FMAC_F64;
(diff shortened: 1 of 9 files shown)


src/arch/amdgpu/vega/gpu_decoder.cc
```diff
@@ -50,6 +50,35 @@ namespace VegaISA
     {
     } // ~Decoder
 
+    void Decoder::setGfx90a()
```
mattsinc: the next two functions below appear to be identical unless I'm missing something small?  If so, what is the value in having them?

abmerop: ~~Yes they are the same.  Gfx950 might be different though.  Do you have a suggested name?~~  Actually it may make more sense to progressively apply the table changes per ISA

mattsinc: I feel slightly guilty that my change made you make such a big modification, but at the same time if they are identical I do think the updated code makes sense if these two are the same.

mattsinc (approved):

PR #2264, opened 2025-05-09
misc: Introduce dictionary parameters (DictParam) in gem5

giactra:
Until now only two types of SimObject parameters existed in gem5:
Params and VectorParams.

```
class MySimObject:
    [...]
    s_param = Param.Int()
    v_param = VectorParam.Int()
```

If users wanted to have a parameter which associated one piece
of data to another (potentially of different type), they would
have resorted to one of the following solutions:

* bind data together with python methods
* bind data together with 2 VectorParam
* bind data together with VectorParam of SimObject

With this commit we add DictParam, which allows to have dictionaries
as parameters.

A DictParam will have the following syntax:

`v_param = VectorParam.KeyType.ValType()`

This will instruct the build system to add an unordered_map member to the
SimObjectParam struct with matching key/value types:

`std::unordered_map<key.ptype, val.ptype> d_param;`

This effectively means using an integral key type comes for free.
If willing to use a custom type, a hash function has to be provided.
This hasn't been currently tested

### DIFF
--- a/build_tools/cxx_config_cc.py
+++ b/build_tools/cxx_config_cc.py
@@ -1,7 +1,7 @@
 # Copyright 2004-2006 The Regents of The University of Michigan
 # Copyright 2010-20013 Advanced Micro Devices, Inc.
 # Copyright 2013 Mark D. Hill and David A. Wood
-# Copyright 2017-2020 ARM Limited
+# Copyright 2017-2020, 2025 Arm Limited
 # Copyright 2021 Google, Inc.
 # Copyright 2023 COSEDA Technologies GmbH
 #
@@ -74,11 +74,12 @@ def cxx_bool(b):
 code('#include "params/%s.hh"' % sim_object_name)
 
 for param in sim_object._params.values():
-    if isSimObjectClass(param.ptype):
-        code('#include "%s"' % param.ptype._value_dict["cxx_header"])
-        code('#include "params/%s.hh"' % param.ptype.__name__)
-    else:
-        param.ptype.cxx_ini_predecls(code)
+    for ptype in param.ptypes:
+        if isSimObjectClass(ptype):
+            code('#include "%s"' % ptype._value_dict["cxx_header"])
+            code('#include "params/%s.hh"' % ptype.__name__)
+        else:
+            ptype.cxx_ini_predecls(code)
 
 code(
     """#include "${{sim_object._value_dict['cxx_header']}}"
@@ -94,12 +95,22 @@ ${param_class}::DirectoryEntry::DirectoryEntry()
 )
 code.indent()
 for param in sim_object._params.values():
+    is_dict = isinstance(param, m5.params.DictParamDesc)
     is_vector = isinstance(param, m5.params.VectorParamDesc)
-    is_simobj = issubclass(param.ptype, m5.SimObject.SimObject)
-
+    is_simobj = (
+        issubclass(param.ptype, m5.SimObject.SimObject)
+        if not is_dict
+        else False
+    )
     code(
-        'parameters["%s"] = new ParamDesc("%s", %s, %s);'
-        % (param.name, param.name, cxx_bool(is_vector), cxx_bool(is_simobj))
+        'parameters["%s"] = new ParamDesc("%s", %s, %s, %s);'
+        % (
+            param.name,
+            param.name,
+            cxx_bool(is_vector),
+            cxx_bool(is_dict),
+            cxx_bool(is_simobj),
+        )
     )
 
 for port in sim_object._ports.values():
@@ -126,19 +137,21 @@ ${param_class}::setSimObject(const std::string &name, SimObject *simObject)
 
 code.indent()
 for param in sim_object._params.values():
-    is_vector = isinstance(param, m5.params.VectorParamDesc)
-    is_simobj = issubclass(param.ptype, m5.SimObject.SimObject)
+    is_dict = isinstance(param, m5.params.DictParamDesc)
+    if not is_dict:
+        is_vector = isinstance(param, m5.params.VectorParamDesc)
+        is_simobj = issubclass(param.ptype, m5.SimObject.SimObject)
 
-    if is_simobj and not is_vector:
-        code('} else if (name == "${{param.name}}") {')
-        code.indent()
-        code(
-            "this->${{param.name}} = "
-            "dynamic_cast<${{param.ptype.cxx_type}}>(simObject);"
-        )
-        code("if (simObject && !this->${{param.name}})")
-        code("   ret = false;")
-        code.dedent()
+        if is_simobj and not is_vector:
+            code('} else if (name == "${{param.name}}") {')
+            code.indent()
+            code(
+                "this->${{param.name}} = "
+                "dynamic_cast<${{param.ptype.cxx_type}}>(simObject);"
+            )
+            code("if (simObject && !this->${{param.name}})")
+            code("   ret = false;")
+            code.dedent()
 code.dedent()
 
 code(
@@ -162,30 +175,32 @@ ${param_class}::setSimObjectVector(const std::string &name,
 
 code.indent()
 for param in sim_object._params.values():
-    is_vector = isinstance(param, m5.params.VectorParamDesc)
-    is_simobj = issubclass(param.ptype, m5.SimObject.SimObject)
+    is_dict = isinstance(param, m5.params.DictParamDesc)
+    if not is_dict:
+        is_vector = isinstance(param, m5.params.VectorParamDesc)
+        is_simobj = issubclass(param.ptype, m5.SimObject.SimObject)
 
-    if is_simobj and is_vector:
-        code('} else if (name == "${{param.name}}") {')
-        code.indent()
-        code("this->${{param.name}}.clear();")
-        code(
-            "for (auto i = simObjects.begin(); "
-            "ret && i != simObjects.end(); i ++)"
-        )
-        code("{")
-        code.indent()
-        code(
-            "${{param.ptype.cxx_type}} object = "
-            "dynamic_cast<${{param.ptype.cxx_type}}>(*i);"
-        )
-        code("if (*i && !object)")
-        code("    ret = false;")
-        code("else")
-        code("    this->${{param.name}}.push_back(object);")
-        code.dedent()
-        code("}")
-        code.dedent()
+        if is_simobj and is_vector:
+            code('} else if (name == "${{param.name}}") {')
+            code.indent()
+            code("this->${{param.name}}.clear();")
+            code(
+                "for (auto i = simObjects.begin(); "
+                "ret && i != simObjects.end(); i ++)"
+            )
+            code("{")
+            code.indent()
+            code(
+                "${{param.ptype.cxx_type}} object = "
+                "dynamic_cast<${{param.ptype.cxx_type}}>(*i);"
+            )
+            code("if (*i && !object)")
+            code("    ret = false;")
+            code("else")
+            code("    this->${{param.name}}.push_back(object);")
+            code.dedent()
+            code("}")
+            code.dedent()
 code.dedent()
 
 code(
@@ -215,16 +230,18 @@ ${param_class}::setParam(const std::string &name,
 
 code.indent()
 for param in sim_object._params.values():
-    is_vector = isinstance(param, m5.params.VectorParamDesc)
-    is_simobj = issubclass(param.ptype, m5.SimObject.SimObject)
+    is_dict = isinstance(param, m5.params.DictParamDesc)
+    if not is_dict:
+        is_vector = isinstance(param, m5.params.VectorParamDesc)
+        is_simobj = issubclass(param.ptype, m5.SimObject.SimObject)
 
-    if not is_simobj and not is_vector:
-        code('} else if (name == "${{param.name}}") {')
-        code.indent()
-        param.ptype.cxx_ini_parse(
-            code, "value", "this->%s" % param.name, "ret ="
-        )
-        code.dedent()
+        if not is_simobj and not is_vector:
+            code('} else if (name == "${{param.name}}") {')
+            code.indent()
+            param.ptype.cxx_ini_parse(
+                code, "value", "this->%s" % param.name, "ret ="
+            )
+            code.dedent()
 code.dedent()
 
 code(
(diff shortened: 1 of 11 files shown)


src/sim/cxx_manager.hh
```diff
@@ -210,6 +210,11 @@ class CxxConfigManager
     SimObject *findObject(const std::string &object_name,
         bool visit_children = false);
 
+    void populateParams(const std::string &object_name,
```
odanrc, line 215: Could you document this function?

src/sim/cxx_manager.cc
```diff
@@ -128,6 +128,103 @@ std::string formatParamList(const std::vector<std::string> &param_values)
     return params.str();
 }
 
+static std::string
```
odanrc, line 134: Could you document this function?

src/sim/cxx_config_ini.hh
```diff
@@ -69,6 +69,10 @@ class CxxIniFile : public CxxConfigFileBase
         const std::string &param_name,
         std::vector<std::string> &values) const;
 
+    bool getParamDict(const std::string &object_name,
```
odanrc, line 81: Ditto documentation

src/sim/cxx_manager.cc
```diff
@@ -128,6 +128,103 @@ std::string formatParamList(const std::vector<std::string> &param_values)
...
+            throw Exception(instance_name, csprintf(
+                "Bad parameter value: .%s=X=\"%s\"",
+                param->name, formatParamList(param_values)));
+        }
+
+        DPRINTF(CxxConfig, "Setting parameter"
+            " %s.%s=%s\n", instance_name, param->name,
+            formatParamList(param_values));
+    } else if (param->isDict) {
+        std::unordered_map<std::string, std::string> param_values;
```
odanrc, line 186: include unordered_map

src/sim/cxx_config_ini.hh
```diff
@@ -69,6 +69,10 @@ class CxxIniFile : public CxxConfigFileBase
...
+        const std::string &param_name,
+        std::unordered_map<std::string, std::string> &values) const;
```
odanrc, line 83: include unordered_map, vector, and string

src/sim/cxx_config_ini.cc
```diff
@@ -67,6 +67,29 @@ CxxIniFile::getParamVector(const std::string &object_name,
...
+CxxIniFile::getParamDict(const std::string &object_name,
+    const std::string &param_name,
+    std::unordered_map<std::string, std::string> &values) const
+{
+    std::vector<std::string> vec_values;
+    auto ret = getParamVector(object_name, param_name, vec_values);
+
+    // Fail if number of vector entries is odd as it means we are
+    // missing a key or a value
+    assert(vec_values.size() % 2 == 0);
```
odanrc, line 82: include cassert

src/sim/cxx_config_ini.cc
```diff
@@ -67,6 +67,29 @@ CxxIniFile::getParamVector(const std::string &object_name,
     return ret;
 }
 
+bool
+CxxIniFile::getParamDict(const std::string &object_name,
+    const std::string &param_name,
+    std::unordered_map<std::string, std::string> &values) const
+{
+    std::vector<std::string> vec_values;
```
odanrc, line 77: Document that we expect the map to be provided as a vector, where each map entry is stored as consecutive values in the vector. The first entry being the key, and the second the value corresponding to that key.

src/sim/cxx_config_ini.cc
```diff
@@ -67,6 +67,29 @@ CxxIniFile::getParamVector(const std::string &object_name,
...
+
+    // Fail if number of vector entries is odd as it means we are
+    // missing a key or a value
+    assert(vec_values.size() % 2 == 0);
+
+    if (ret) {
+        for (auto idx = 0; idx < vec_values.size(); idx+=2) {
+            const std::string &key = vec_values[idx];
+            const std::string &val = vec_values[idx + 1];
+            values[key] = val;
```
odanrc, line 91: Should we verify that the key does not appear twice, or is this checked elsewhere?

src/sim/cxx_config.hh
```diff
@@ -185,6 +189,13 @@ class CxxConfigParams
         const std::vector<std::string> &values, const Flags flags)
     { return false; }
 
+    /** As setParamVector but for parameters given as dictionaries
```
odanrc, line 192: nit: this references setParamVector, which references itself. Could you fix setParamVector's self reference?

src/sim/cxx_config.hh
```diff
@@ -79,13 +79,17 @@ class CxxConfigDirectoryEntry
         /* Is this a vector or singleton parameters/SimObject */
         const bool isVector;
 
+        /* Is this a dictionary or singleton parameters/SimObject */
```
odanrc: Why do we have two booleans that can represent the same thing (the singleton part)? This is very confusing.

src/python/m5/params.py
```diff
@@ -469,9 +604,56 @@ def __call__(self, *args, **kwargs):
         return self.param_desc_class(self.ptype_str, ptype, *args, **kwargs)
 
 
-Param = ParamFactory(ParamDesc)
+class DictParamFactory:
```
odanrc, line 607: Ditto documentation

src/python/m5/params.py
```diff
@@ -469,9 +604,56 @@ def __call__(self, *args, **kwargs):
...
+            self.key_ptype_str,
+            key_ptype,
+            self.val_ptype_str,
+            val_ptype,
+            *args,
+            **kwargs,
+        )
+
+
+class DictParamValueFactory:
```
odanrc, line 643: Ditto documentation

src/python/m5/params.py
```diff
@@ -469,9 +604,56 @@ def __call__(self, *args, **kwargs):
...
+        self.param_desc_class = param_desc_class
+        self.key_ptype_str = key_ptype_str
+
+    def __getattr__(self, val_ptype_str):
+        return DictParamFactory(
+            self.param_desc_class, self.key_ptype_str, val_ptype_str
+        )
+
+
+class DictParamKeyFactory:
```
odanrc, line 660: Ditto documentation

src/python/m5/params.py
```diff
@@ -447,6 +487,101 @@ def cxx_decl(self, code):
         code("std::optional< ${{self.ptype.cxx_type}} > ${{self.name}};")
 
 
+class DictParamDesc(ParamDesc):
```
odanrc, line 498: Ditto documentation

src/python/m5/params.py
```diff
@@ -270,6 +279,34 @@ def unproxy(self, base):
             return [v.unproxy(base) for v in self]
 
 
+class DictParamValue(dict, metaclass=MetaParamValue):
```
odanrc, line 290: Ditto documentation, including inner functions

src/python/m5/params.py
```diff
@@ -183,6 +179,15 @@ def __init__(self, ptype_str, ptype, *args, **kwargs):
         if not hasattr(self, "desc"):
             raise TypeError("desc attribute missing")
 
+
+class SingleTypeParamDesc(ParamDesc):
```
odanrc, line 183: Ditto documentation

build_tools/cxx_config_cc.py
```diff
@@ -248,20 +265,64 @@ def cxx_bool(b):
...
+    return ret;
+}
+
+bool
+${param_class}::setParamDict(const std::string &name,
+    const std::unordered_map<std::string, std::string> &values, const Flags flags)
+{
+    bool ret = true;
+
+    if (false) {
```
odanrc, line 306: I am guessing this if (false) is to align with the previous setters. Could you add this info as a comment?

configs/example/read_config.py
```diff
@@ -467,6 +488,15 @@ def get_param(self, object_name, param_name):
     def get_param_vector(self, object_name, param_name):
         return self.parser.get(object_name, param_name).split()
 
+    def get_param_dict(self, object_name, param_name):
```
odanrc, line 491: Ditto documentation

src/python/m5/SimObject.py
```diff
@@ -1052,6 +1056,10 @@ def find_any(self, ptype):
                 found_obj = child
         # search param space
         for pname, pdesc in self._params.items():
+            if isinstance(pdesc, DictParamDesc):
+                # DictParams are not supported
```
odanrc, line 1060: GitHub is making this review much harder than it should be by bugging when I click to expand the code, so could you explain why dict params are not supported here?

src/python/m5/SimObject.py
```diff
@@ -1236,6 +1248,14 @@ def getCCParams(self):
                 else:
                     for v in value:
                         getattr(cc_params, param).append(v)
+            elif isinstance(self._params[param], DictParamDesc):
```
odanrc, line 1251: nit: I have no clue of what is happening here, so adding comments would be nice

odanrc (commented):
I understand that there is almost no documentation in this portion of the codebase, so you just went with the flow, but given that you know probably understands a bit better what is going on, could you add documentation to at least the new code? Reviewing this was quite hard.

src/sim/cxx_manager.hh
giactra: Ack

src/sim/cxx_manager.cc
giactra: Ack

src/sim/cxx_config_ini.hh
giactra: Ack

src/sim/cxx_config_ini.cc
giactra: Ack

src/sim/cxx_config_ini.cc
giactra: Ack

src/sim/cxx_config_ini.cc
giactra: Ack

src/sim/cxx_config.hh
giactra: I don't understand what you mean by that. But I agree it is confusing: if isDict is false, the parameter could still be a vector. Will amend the comment

src/python/m5/params.py
giactra: Ack

src/python/m5/params.py
giactra: Ack

src/python/m5/params.py
giactra: Ack

src/python/m5/params.py
giactra: Ack

src/sim/cxx_config.hh
giactra: Ack

build_tools/cxx_config_cc.py
giactra: If I do this here I have to change it in all other instances for consistency, which makes the PR harder to review

src/python/m5/SimObject.py
giactra: Hey Daniel, if you factor out the weird way gem5 does parameter assignment, code is pretty self-explanatory IMHO. Unless you want me to comment that I am assigning a dictionary to a dictionary parameter

src/python/m5/SimObject.py
giactra: This is the find_any(self, ptype) method which is called for example when you use proxies like Parent.Any and Parent.All.

It assumes a single ptype as a parameter. This is why DictParams are not supported because they require two. In other words you cannot use .Any and .All with DictParams

configs/example/read_config.py
giactra: Ack

src/python/m5/params.py
```diff
@@ -447,6 +494,101 @@ def cxx_decl(self, code):
...
+
+        tmp_dict = {
+            self.key_desc.pretty_print(key): self.val_desc.pretty_print(val)
+            for key, val in value.items()
+        }
+
+        return pprint.pformat(tmp_dict)
+
+    # This is a helper function for the new config system
+    def __call__(self, value):
```
powerjg, line 566: Should you instead use the `convert` function? The code is exactly the same.

src/python/m5/SimObject.py
```diff
@@ -1236,6 +1248,14 @@ def getCCParams(self):
...
+                assert isinstance(value, dict)
+                dic = getattr(cc_params, param)
+                assert not len(dic)
```
powerjg: `dic` should be a `dict`. So, I think that `assert not dic, "Dictionary param should not be empty"` would be more clear.

src/python/m5/SimObject.py
```diff
@@ -1236,6 +1248,14 @@ def getCCParams(self):
...
+                if isinstance(dic, dict):
+                    setattr(cc_params, param, dict(value))
+                else:
+                    raise TypeError("Shouldn't happen")
```
powerjg: Can you explain to the user what shouldn't happen? Maybe `TypeError("Must provide dictionary for dictionary param {name()}")` or something like that.

powerjg (commented):
Generally, I'm happy with this code. The only kinda big thing I see that needs to be done is to put something about this in the release notes. It's a big new feature, so we should make sure to advertise it!

Out of curiosity, how much is the C++ config code used? If it's often used, it may be worth writing some good tests for it. It's something we should put on our medium importance list of to do items, if this code is used heavily.

src/python/m5/SimObject.py
odanrc: I am guessing someone that uses Python more frequently would have a clue, so I will not insist on this

odanrc (approved):

src/python/m5/params.py
giactra: Even if they are not part of the same hierarchy (we rely on duck typing) these are already documented in ParamValue. Let me know if you want me to copy the comment here as well. This would effectively mean we should copy in VectorParamValue as well

src/python/m5/SimObject.py
giactra: This logic is actually checking the opposite

`assert not len(dic)`

will fail if dic is NOT empty.

I took this from the Vector handling code.
This code does not get called at parameter assignment but rather at parameter construction (only once) and therefore assumes the parameter contains a empty container

powerjg: Ah, sorry. I see now. A more descriptive error could potentially be useful for future developers, but since it's not user-facing this is just a nit.

giactra: Done!

powerjg (approved):
Thanks! I'm excited about this change. Let's just not forget to mention it in the release notes.

giactra:
Thanks @odanrc @powerjg for your review!

PR #2269, opened 2025-05-14
python: improve error message in SimObject __setattr__

harry900831:
This CL improves the error message when the attr doesn't match the cases above in __setattr__.

Example:
One possible scenario is that User might forget to assign the type_signature for SignalPort:

- correct: port = SignalSourcePort("bool")("port description")
- wrong: port = SignalSourcePort("port description")

And the gem5 can sill be built and error only happens on gem5 runtime when it is actually accessing that port, however the original error message is quite misleading. Hence creating this CL to improve the error message so the user is able to identify the issue easily.

### DIFF
--- a/src/python/m5/SimObject.py
+++ b/src/python/m5/SimObject.py
@@ -418,7 +418,9 @@ class MetaSimObject(type):
             return
 
         # no valid assignment... raise exception
-        raise AttributeError(f"Class {cls.__name__} has no parameter '{attr}'")
+        raise AttributeError(
+            f"Invalid assignment for Class {cls.__name__} with parameter {attr}"
+        )
 
     def __getattr__(cls, attr):
         if attr == "cxx_class_path":
@@ -935,7 +937,8 @@ class SimObject(metaclass=MetaSimObject):
 
         # no valid assignment... raise exception
         raise AttributeError(
-            f"Class {self.__class__.__name__} has no parameter {attr}"
+            f"Invalid assignment for Class {self.__class__.__name__} with"
+            f" parameter {attr}"
         )
 
     # this hack allows tacking a '[0]' onto parameters that may or may


giactra:
> This CL improves the error message when the attr doesn't match the cases above in **setattr**.
> 
> Example: One possible scenario is that User might forget to assign the type_signature for SignalPort:
> 
>     * correct: port = SignalSourcePort("bool")("port description")
> 
>     * wrong: port = SignalSourcePort("port description")
> 

I am all for better debug messages, though can I ask you to provide a different example? Or to better detail the general case where this print would be triggered?

In your example the error (arising from the fact the factory function is named identically to the class it is supposed to return [1]) is that you are assigning a class to a parameter, rather than its instance

[1]: https://github.com/gem5/gem5/blob/stable/src/sim/SignalPort.py#L35

harry900831:
Why the original debug message is misleading: In the case that SimObject actually has that attribute`hasattr(self, attr)`, the  debug message `"Class {self.__class__.__name__} has no parameter {attr}"` would be confusing. Hence I specifically deal with the case `hasattr(self, attr)` and output a more informative message.

The SignalSourcePort case is the objective for this PR because we've encounter this problem several times and for people that are not familiar with gem5, the original error message would make them hard to identify the issue.

I haven't come up with another real example, but I believe the `hasattr(self, attr)` check is reasonable.

Thanks!

giactra:
> Why the original debug message is misleading: In the case that SimObject actually has that attribute`hasattr(self, attr)`, the debug message `"Class {self.__class__.__name__} has no parameter {attr}"` would be confusing. Hence I specifically deal with the case `hasattr(self, attr)` and output a more informative message.
> 
> The SignalSourcePort case is the objective for this PR because we've encounter this problem several times and for people that are not familiar with gem5, the original error message would make them hard to identify the issue.
> 
> I haven't come up with another real example, but I believe the `hasattr(self, attr)` check is reasonable.
> 
> Thanks!

Yes that makes sense. However the reason why I asked you about the general case is because of the following statement

```
{existing_attr_type_name}) exists, but setting it directly"
                " is not permitted by this class's logic, or the provided"
                " value is not of an expected type for this operation. Please"
                f" check the type or the instantiation of '{attr}'."
```

If the reason is type incompatibility between the parameter and the assigned value, that should be automatically handled few lines before [here](https://github.com/gem5/gem5/blob/stable/src/python/m5/SimObject.py#L885). So the question is: why is that exception not being triggered there? Are we sure we aren't just adding a safety net here to cover a bug in your Param.convert method?

If current version of the code assumes the attr does not exist at the end of the method, it means the original writers thought all type inconsistencies should be handled before by existing logic.

So I'd rather understand better which error types are currently not handled and for what reason, instead of breaking the existing implicit assumption within __setattr__

Please let me know if it makes sense to you

harry900831:
> If the reason is type incompatibility between the parameter and the assigned value, that should be automatically handled few lines before [here](https://github.com/gem5/gem5/blob/stable/src/python/m5/SimObject.py#L885).

IIUC, the code you refer to is specifically for "ParamDesc" which is not related to the original problem we encounter.

> So I'd rather understand better which error types are currently not handled and for what reason

Objective:  Catch the incorrect definition of SignalPort in a SimObject:
```py
class FooObject(SimObject):
  port = SignalSourcePort("bool")("port description") # correct
  port = SignalSourcePort("port description") # incorrect
```
For the incorrect one, FooObject.port would actually be a `type` instead of `port`.

Next, let's take a look into the SimObject.\_\_setattr\_\_, which should be called when we write `FooObject.port = AnotherObject.port` in gem5 config.
:
```py
    def __setattr__(self, attr, value):
        # normal processing for private attributes
        if attr.startswith("_"):
            object.__setattr__(self, attr, value)
            return

        if attr in self._deprecated_params:
            ...

        if attr in self._ports:
            # set up port connection
            self._get_port_ref(attr).connect(value)
            return

        param = self._params.get(attr)
        if param:
            ...

        # if RHS is a SimObject, it's an implicit child assignment
        if isSimObjectOrSequence(value):
            ...

        # no valid assignment... raise exception
        raise AttributeError(
            f"Class {self.__class__.__name__} has no parameter {attr}"
        )

```
Because `port` is a `type` instead of `Port`, so it won't fall into the `if attr in self._ports:` cases and do the port connection. Therefore, it will raise the `Class FooObject has no parameter port` error.

> If current version of the code assumes the attr does not exist at the end of the method, it means the original writers thought all type inconsistencies should be handled before by existing logic.


I think the assumption is wrong. The current implementation of \_\_setaddr\_\_ only assumes the "valid attr" for a SimObject is:
- startswith("_") and we won't do anything special:
- _deprecated_params
- port
- Param
- SimObject or Sequence

All other types of attr is "invalid" and we should raise an error because the user is clearly doing something wrong. Like in the SignalPort error cases above, we're setting an attr that is a 'type' which is invalid, hence we should raise an error.

And in this PR, we're just distinguishing the cases for "invalid attr". It could be either "attr not exist" or "attr does not falled in to any valid attr cases".

Above is the reason why I think the PR is reasonable and it doesn't break any assumption, WDYT? Any suggestions regarding the error message or other approach to resolve this would be helpful.

Thanks!

giactra:
Apologies for the late reply.

You are right a parameter wouldn't fall into that code. I was just concerned about being overly generic when it comes to catching errors.

For example it's clear the error you are trying to cover is when you try to set(attr) a SimObject attribute which is a type. IMHO a more informative debug message would explicitly look into it:


    def __setattr__(self, attr, value):
        # normal processing for private attributes
        if attr.startswith("_"):
            object.__setattr__(self, attr, value)
            return

        if attr in self._deprecated_params:
            ...

        if attr in self._ports:
            # set up port connection
            self._get_port_ref(attr).connect(value)
            return

        param = self._params.get(attr)
        if param:
            ...

        # if RHS is a SimObject, it's an implicit child assignment
        if isSimObjectOrSequence(value):
            ...

       if hasattr(self, attr) and isinstance(getattr(self, attr), type):
           raise AttributeError(f"Cannot assign to {attr} which is a type attribute")

        # no valid assignment... raise exception
        raise AttributeError(
            f"Class {self.__class__.__name__} has no parameter {attr}"
        )

This IMHO makes it more explicit that that type of operation is forbidden/not supported and probably gives more information to the developer.

Please let me know what you think of this

harry900831:
I'm fine with explicitly having a check to type.

However, `hasattr(self, attr)` but not a "type" is still contradicting the  "Class {self.__class__.__name__} has no parameter {attr}" error message. 

I propose the below two options, WDYT?

1.
```py
    if hasattr(self, attr):
       if isinstance(getattr(self, attr), type):
         raise AttributeError(f"Cannot assign to {attr} which is a type attribute, you might forget to instantiate the object!")
       else:
         existing_attr_type_name = type(getattr(self, attr))
         raise AttributeError(
                f"Error in '{self.__class__.__name__}': Cannot assign to"
                f" existing attribute '{attr}'. The attribute '{attr}' (type:"
                f" {existing_attr_type_name}) exists, but setting it directly"
                " is not permitted by this class's logic, or the provided"
                " value is not of an expected type for this operation. Please"
                f" check the type or the instantiation of '{attr}'."
         )

    # no valid assignment... raise exception
    raise AttributeError(
        f"Class {self.__class__.__name__} has no parameter {attr}"
    )
```

2.
```py
    if hasattr(self, attr) and isinstance(getattr(self, attr), type):
       raise AttributeError(f"Cannot assign to {attr} which is a type attribute, you might forget to instantiate the object!")

    # no valid assignment... raise exception
    raise AttributeError(
        f"Invalid assignment for Class {self.__class__.__name__} with parameter {attr}"
    )
```

giactra:
> I'm fine with explicitly having a check to type.
> 
> However, `hasattr(self, attr)` but not a "type" is still contradicting the "Class {self.**class**.**name**} has no parameter {attr}" error message.

I would enter into the same line of questions I started with my initial comments. After our discussion it is clear the generic error we were not covering (of which the SignalSinkPort is an instance) was that of trying to set a type attribute.

When you say that the following case is still not covered:

> However, hasattr(self, attr) but not a "type"

I would ask you: "when is this happening?" (more specifically, like you replied previously with "when attr is a type"). In other words, which attributes can be declared within a SimObject [other than Param, Port, private attributes (starting with _), type, which we explicitly handle in the SimObject.setattr] ?

Apologies if you feel like I am nitpicking. But since this PR is aimed at providing a more descriptive insight to our users, I want to make sure this comes not only from the explicit log, but also from the structure of our code.

So I am concerned of adding a generic fallback error message, which potentially will never be executed and which will let some users wonder: "upon which condition would this be triggered?".

I realise this might come from personal preference, so I will add @powerjg as a tie breaker in case you were not convinced about my arguments and wanted a second opinion. I'd be happy to relax my views if the rest of the community disagrees with my review :). Have a nice weekend

harry900831:
The reason why `type` can be an attribute of SimObject is because it satisfies the public_value condition.
https://github.com/gem5/gem5/blob/b9da2bfe1e21f6efa2730edd4e80e58c5d965458/src/python/m5/SimObject.py#L125-L128
So it won't panic at the MetaSimObject's  __setattr__() implementation because it's not in value_dict.
https://github.com/gem5/gem5/blob/b9da2bfe1e21f6efa2730edd4e80e58c5d965458/src/python/m5/SimObject.py#L157-L176


> I would ask you: "when is this happening?" (more specifically, like you replied previously with "when attr is a type"). In other words, which attributes can be declared within a SimObject [other than Param, Port, private attributes (starting with _), type, which we explicitly handle in the SimObject.setattr] ?

Based on the above implementation, we can create a case that defines a normal function in a SimOjbect will have the same, `f"Class {self.__class__.__name__} has no parameter {attr}"` error message.


```py
def meow():
  print("Meow")

class FooSimObject(SimObject):
  sound = meow
```
The above code can successfully build gem5.opt but failed at running gem5 with the below code just like the signal port issue. (Although this case is not realistic, I just want to showcase that it's possible)

```py
def woof():
  print('woof')

FooSimObject(sound = woof)
# AttributeError: Class FooSimObject has no parameter sound
```

The part I care most is the misleading `f"Class {self.__class__.__name__} has no parameter {attr}"`, even if we just simply change the error message to invalid assignment without any extra condition, it would solve my concern too.

```py
raise AttributeError(
        f"Invalid assignment for Class {self.__class__.__name__} with parameter {attr}"
    )
```

I find out that we should also change the `has no parameter` message in MetaSimObject's __setattr__(), it has a similar issue too.

https://github.com/gem5/gem5/blob/b9da2bfe1e21f6efa2730edd4e80e58c5d965458/src/python/m5/SimObject.py#L416

The current message would make the below case return a misleading error message when building gem5:

```py
def FooSimObject(SimObject):
  foo = 3
# terminate called after throwing an instance of 'pybind11::error_already_set'
#  what():  AttributeError: Class FooSimObject has no parameter 'foo'
```

Thank you for the thoughtful feedback, and I appreciate you taking the time to explain your perspective on the  error message. I believe we both agree that avoiding potential user confusion is important. If we still have no conclusion after this comment, let's take a second opinion.

Thanks again!

powerjg:
Apologies for reviewing this late... can I have a brief overview of where the PR is now? Specifically, I'd like to the following:

1. Can you provide a short python script that I can run in gem5 which exercises this code?
2. Show the output before this PR
3. Show the output after this PR

Thanks!

erin-le:
@harry900831: Hello, could you respond to Jason's (powerjg's) comment when you have time? Thank you!

erin-le:
@harry900831: Hello, just another reminder about this PR. Could you respond to Jason's (powerjg's) comment? Thank you!

erin-le:
@harry900831: Hello, if you're still interested in working on this PR, could you respond to powerjg's comment? Thank you!

harry900831:
Sorry for the late reply, I've updated the commit, I decided to make things simpler. PTAL.

## Objective: 
The "has no parameter" is misleading for people who are new to gem5.

## How to trigger this "has no parameter" message?

Define a signal port in the wrong way on whatever SimObject, and this can successfully build gem5, but encounter "has no  parameter" in runtime.
correct: port = SignalSourcePort("bool")("port description")
wrong: port = SignalSourcePort("port description")

```
class FooSimObject(SimObject):
  port = SignalSourcePort("port description")
```

Or just define whatever parameter that is not expected in SimObject.

harry900831:
@erin-le @powerjg Do we have any updates on this? Thanks!

powerjg (approved):
Seems fine to me. Thanks for improving the error messages!

PR #2271, opened 2025-05-14
arch-arm: Add stats to track PMU events

giactra:
In gem5 Arm there are two ways for extracting performance
metrics from a simulation: via the generic gem5 statistics
and via the architected PMU object.

The biggest difference between the two is that, while the PMU is visible
from guest application, gem5 statistics are transparent to the simulated
system.

This effectively means:

1) While guest application is in charge of enabling/setting the
recording activity with PMU counters, event recording for gem5 stats
happens without any intervention from the guest software.
Moreover, since the number of available PMU counters is smaller than the
number of architected events, only a fraction of those events will
be recorded at the same time.

2) Extracting PMU numbers requires the guest application to read them
and write them somewhere; either to the terminal, or to a file.
Whereas gem5 stats will be automatically dumped to a human readable file
(stats.txt) at the end of simulation, without explicit programming
from the user (as point 1)

This patch is amending the ArmPMU object to dump PMU counters as
if they were common statistics. In this way we can observe PMU
events without requiring explicit programming. We can also
observe all of them without being limited by the number of
performance counters

The way we do it is by defining non-architectural counters, called
Stat Counters. Every instantiated event will have its own Stat counter
(and will hold a pointer to it).
Those counters will listen on the associated event so that if the
PMU is actually programmed by the guest application, there will
be two counters recording the same event and bearing the same
value.

### DIFF
--- a/src/arch/arm/ArmPMU.py
+++ b/src/arch/arm/ArmPMU.py
@@ -276,3 +276,25 @@ class ArmPMU(SimObject):
         False,
         "Choose whether to use 64-bit or 32-bit PMEVCNTR<n>_EL0 registers.",
     )
+
+    statCounters = DictParam.Int.String(
+        {},
+        """
+        Dictionary of PMU events to be merged into the stats framework.
+            Key = ID of the PMU event
+            Value = Name of the gem5 stat
+
+        For example the following param value:
+
+        my_stat_counters = {
+            int(EventTypeId("INST_RETIRED")): "INST_RETIRED",
+            int(EventTypeId("CPU_CYCLES")): "CPU_CYCLES",
+            0xC0C0 : "MY_IMPDEF_EVENT"
+        }
+
+        Will produce three stats under the pmu section:
+            root.<>.pmu.INST_RETIRED=<val>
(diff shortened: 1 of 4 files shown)


powerjg:
Seems like a good idea! I'll dig into this in more detail soon... I may have lots of questions for you.

giactra:
> Seems like a good idea! I'll dig into this in more detail soon... I may have lots of questions for you.

Sure! Happy to provide further details!

andysan (approved):

giactra:
Thanks @andysan! I will wait for @powerjg as he had some questions though it would be nice if we could merge this before the staging branch gets created.

Note: I have forced pushed a simpler version of the PR after @andysan spotted a stat reset issue in the original implementation

powerjg:
It will get in for staging... I'll take a look today or tomorrow.

BobbyRBruce (approved):
The code looks fine to me, definitely like enums over passing unsigned ints

powerjg (approved):
Sorry I didn't look at this before now. LGTM.

giactra:
Thanks @BobbyRBruce @powerjg !

PR #2272, opened 2025-05-14
gpu-compute,arch-vega: Included two new DS instructions

german-v-t-um-es:
While testing benchmarks on the FS gfx90a model, I needed to add support for two new ds instructions to ensure correct execution. Additionally, since the benchmarks made use of both v_mbcnt_lo_u32_b32 and v_mbcnt_hi_u32_b32, I added REG_EXEC_HI to the relevant register switch-case to properly support these instructions.

### DIFF
--- a/src/arch/amdgpu/vega/insts/ds.cc
+++ b/src/arch/amdgpu/vega/insts/ds.cc
@@ -986,6 +986,9 @@ namespace VegaISA
     Inst_DS__DS_ADD_RTN_U32::Inst_DS__DS_ADD_RTN_U32(InFmt_DS *iFmt)
         : Inst_DS(iFmt, "ds_add_rtn_u32")
     {
+        setFlag(MemoryRef);
+        setFlag(AtomicAdd);
+        setFlag(AtomicReturn);
     } // Inst_DS__DS_ADD_RTN_U32
 
     Inst_DS__DS_ADD_RTN_U32::~Inst_DS__DS_ADD_RTN_U32()
@@ -1000,8 +1003,60 @@ namespace VegaISA
     void
     Inst_DS__DS_ADD_RTN_U32::execute(GPUDynInstPtr gpuDynInst)
     {
-        panicUnimplemented();
+        Wavefront *wf = gpuDynInst->wavefront();
+
+        if (gpuDynInst->exec_mask.none()) {
+            wf->decLGKMInstsIssued();
+            return;
+        }
+
+        gpuDynInst->execUnitId = wf->execUnitId;
+        gpuDynInst->latency.init(gpuDynInst->computeUnit());
+        gpuDynInst->latency.set(
+                gpuDynInst->computeUnit()->cyclesToTicks(Cycles(24)));
+        ConstVecOperandU32 addr(gpuDynInst, extData.ADDR);
+        ConstVecOperandU32 data(gpuDynInst, extData.DATA0);
+
+        addr.read();
+        data.read();
+
+        calcAddr(gpuDynInst, addr);
+
+        for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
+            if (gpuDynInst->exec_mask[lane]) {
+                (reinterpret_cast<VecElemU32*>(gpuDynInst->a_data))[lane]
+                    = data[lane];
+            }
+        }
+
+        gpuDynInst->computeUnit()->localMemoryPipe.issueRequest(gpuDynInst);
     } // execute
+
+    void
+    Inst_DS__DS_ADD_RTN_U32::initiateAcc(GPUDynInstPtr gpuDynInst)
+    {
+        Addr offset0 = instData.OFFSET0;
+        Addr offset1 = instData.OFFSET1;
+        Addr offset = (offset1 << 8) | offset0;
+
+        initAtomicAccess<VecElemU32>(gpuDynInst, offset);
+    } // initiateAcc
+
+    void
+    Inst_DS__DS_ADD_RTN_U32::completeAcc(GPUDynInstPtr gpuDynInst)
+    {
+        VecOperandU32 vdst(gpuDynInst, extData.VDST);
+
+        for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
+            if (gpuDynInst->exec_mask[lane]) {
+                // vdst[lane] = 10;
+                vdst[lane] = (reinterpret_cast<VecElemU32*>(
+                    gpuDynInst->d_data))[lane];
+            }
+        }
+
+        vdst.write();
+    } // completeAcc
     // --- Inst_DS__DS_SUB_RTN_U32 class methods ---
 
     Inst_DS__DS_SUB_RTN_U32::Inst_DS__DS_SUB_RTN_U32(InFmt_DS *iFmt)
@@ -1319,6 +1374,9 @@ namespace VegaISA
     Inst_DS__DS_CMPST_RTN_B32::Inst_DS__DS_CMPST_RTN_B32(InFmt_DS *iFmt)
         : Inst_DS(iFmt, "ds_cmpst_rtn_b32")
     {
+        setFlag(MemoryRef);
+        setFlag(AtomicCAS);
+        setFlag(AtomicReturn);
     } // Inst_DS__DS_CMPST_RTN_B32
 
     Inst_DS__DS_CMPST_RTN_B32::~Inst_DS__DS_CMPST_RTN_B32()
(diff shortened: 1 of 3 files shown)


src/arch/amdgpu/vega/insts/ds.cc
```diff
@@ -1000,8 +1003,60 @@ namespace VegaISA
...
+
+        if (gpuDynInst->exec_mask.none()) {
+            wf->decLGKMInstsIssued();
+            return;
+        }
+
+        gpuDynInst->execUnitId = wf->execUnitId;
+        gpuDynInst->latency.init(gpuDynInst->computeUnit());
+        gpuDynInst->latency.set(
+                gpuDynInst->computeUnit()->cyclesToTicks(Cycles(24)));
```
mattsinc, line 1016: it would be better if we weren't using magic numbers like 24 here, but at the same time I realize this isn't the only instruction to do so, so I won't fight too hard against it.

src/arch/amdgpu/vega/insts/ds.cc
```diff
@@ -1000,8 +1003,60 @@ namespace VegaISA
...
+    } // initiateAcc
+
+    void
+    Inst_DS__DS_ADD_RTN_U32::completeAcc(GPUDynInstPtr gpuDynInst)
+    {
+        VecOperandU32 vdst(gpuDynInst, extData.VDST);
+
+        for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
+            if (gpuDynInst->exec_mask[lane]) {
+                // vdst[lane] = 10;
```
mattsinc: This line (1052) should be removed

src/arch/amdgpu/vega/insts/ds.cc
```diff
@@ -1319,6 +1374,9 @@ namespace VegaISA
     Inst_DS__DS_CMPST_RTN_B32::Inst_DS__DS_CMPST_RTN_B32(InFmt_DS *iFmt)
         : Inst_DS(iFmt, "ds_cmpst_rtn_b32")
     {
+        setFlag(MemoryRef);
+        setFlag(AtomicCAS);
```
mattsinc, line 1377: @abmerop can chime in here, but technically this is not a CAS, it's an atomicAdd.  I'm not sure if we are distinguishing between them for the flags though

src/arch/amdgpu/vega/insts/ds.cc
```diff
@@ -1338,10 +1396,68 @@ namespace VegaISA
...
+
     } // execute
     // --- Inst_DS__DS_CMPST_RTN_F32 class methods ---
 
+    void
+    Inst_DS__DS_CMPST_RTN_B32::initiateAcc(GPUDynInstPtr gpuDynInst)
+    {
+        Addr offset0 = instData.OFFSET0;
+        Addr offset1 = instData.OFFSET1;
+        Addr offset = (offset1 << 8) | offset0;
```
mattsinc, line 1439: I realize you probably copied this from another atomic instruction, but any idea why you are shifting by 8?  This sort of thing seems worth documenting.

mattsinc (changes requested):
thanks @german-v-t-um-es !  A couple small things

mattsinc:
Also @german-v-t-um-es I did want to know that we technically don't directly support gfx90a because the MI100 generation of GPUs added a special register file that was immediately removed in the MI200/300 generations.  So, we decided it wasn't worth adding support for a generation of GPU that was already not the latest generation and would have required special support beyond that.  Nevertheless, I think these instructions are worth adding -- do you have a test that validates them?

src/arch/amdgpu/vega/insts/ds.cc
```diff
@@ -1338,10 +1396,68 @@ namespace VegaISA
...
+        src.read();
+        cmp.read();
+        calcAddr(gpuDynInst, addr);
+
+        for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
+            if (gpuDynInst->exec_mask[lane]) {
+                (reinterpret_cast<VecElemU32*>(gpuDynInst->x_data))[lane]
+                    = src[lane];
+                (reinterpret_cast<VecElemU32*>(gpuDynInst->a_data))[lane]
+                    = cmp[lane];
```
abmerop, line 1425: In the instruction description it says: "Compare and store. Caution, the order of src and cmp are the
*opposite* of the BUFFER_ATOMIC_CMPSWAP opcode." I think in order to use the existing AtomicCAS operator these values would have to be reversed.  I.e., x_data is cmp and a_data is src

src/arch/amdgpu/vega/insts/ds.cc
abmerop: This should work, but I believe the order needs to be switched in the execute method below

src/arch/amdgpu/vega/insts/ds.cc
german-v-t-um-es: Sure, sorry for that line. I included it while I was debugging and forgot to erase it when I finished debugging.

src/arch/amdgpu/vega/insts/ds.cc
german-v-t-um-es: So, we can leave it like that I guess?

src/arch/amdgpu/vega/insts/ds.cc
german-v-t-um-es: Thanks for the note! I believe the values you're referring to are already inverted in lines 1413–1414.
From what I see, `src` is assigned from `extData.DATA1` and `cmp` from `extData.DATA0`, which matches the expected order for this opcode, if I'm understanding it correctly. Can you verify it?

src/arch/amdgpu/vega/insts/ds.cc
german-v-t-um-es: You're right, I took this pattern from the existing DS_ADD_U32 implementation. From what I understand, we shift OFFSET1 by 8 bits to build a 16-bit offset, where OFFSET1 provides the upper 8 bits and OFFSET0 the lower 8 bits. This offset is needed to compute the full virtual address used in initAtomicAccess.

I can add this comment if you'd like and @abmerop also finds it correct: 
`// Build a 16-bit offset by combining two 8-bit fields:                                             
//   OFFSET1 → high byte, OFFSET0 → low byte.`

german-v-t-um-es:
> Also @german-v-t-um-es I did want to know that we technically don't directly support gfx90a because the MI100 generation of GPUs added a special register file that was immediately removed in the MI200/300 generations. So, we decided it wasn't worth adding support for a generation of GPU that was already not the latest generation and would have required special support beyond that. Nevertheless, I think these instructions are worth adding -- do you have a test that validates them?

I understand what you are saying, but isn't gfx90a the MI200 generation and gfx908 the MI100 generation, according to [Accelerator and GPU hardware specifications](https://rocm.docs.amd.com/en/latest/reference/gpu-arch-specs.html). 

I am very happy that you find these instructions useful and I do have a small test that validates them. I'm attaching here both the Makefile that compiles the test and the source code, which is compiled inside the docker container with ROCm 6.1 used for the GPUFS models. To make sure the code generates the instructions I have implemented, I obtained the assembly code and checked their presence.
[test.zip](https://github.com/user-attachments/files/20242706/test.zip)

src/arch/amdgpu/vega/insts/ds.cc
abmerop: I see now, it does look correct then.

src/arch/amdgpu/vega/insts/ds.cc
abmerop: That's correct.  Here is the quote from the MI200 spec: "Immediate offset, in bytes. Instructions with one address combine the offset fields into a single 16-bit unsigned offset: {offset1, offset0}."

mattsinc:
> > Also @german-v-t-um-es I did want to know that we technically don't directly support gfx90a because the MI100 generation of GPUs added a special register file that was immediately removed in the MI200/300 generations. So, we decided it wasn't worth adding support for a generation of GPU that was already not the latest generation and would have required special support beyond that. Nevertheless, I think these instructions are worth adding -- do you have a test that validates them?
> 
> I understand what you are saying, but isn't gfx90a the MI200 generation and gfx908 the MI100 generation, according to [Accelerator and GPU hardware specifications](https://rocm.docs.amd.com/en/latest/reference/gpu-arch-specs.html).
> 
> I am very happy that you find these instructions useful and I do have a small test that validates them. I'm attaching here both the Makefile that compiles the test and the source code, which is compiled inside the docker container with ROCm 6.1 used for the GPUFS models. To make sure the code generates the instructions I have implemented, I obtained the assembly code and checked their presence. [test.zip](https://github.com/user-attachments/files/20242706/test.zip)

You're right, I got the ISA numbers wrong

mattsinc (approved):
since @abmerop had questions about the order of the CAS lines, he should also approve before we merge

abmerop (approved):

german-v-t-um-es:
Hi @abmerop and @mattsinc, just wanted to check if there's anything else needed to finalize the pull request. The tests are passing, but since develop is being updated frequently, it keeps requiring a merge with the latest develop, which then triggers the tests to rerun each time.

abmerop:
You shouldn't have to keep updating it, once the tests pass GitHub should be smart enough to rebase automatically

mattsinc:
> You shouldn't have to keep updating it, once the tests pass GitHub should be smart enough to rebase automatically

thanks @abmerop for merging before I got to this

PR #2274, opened 2025-05-15
stdlib: correct memory range check in dramsim3

### LINKED ISSUE #2273: Incorrect condition in set_memory_range in dramsim_3.py
**Describe the bug**
There are two syntax/logic errors in the conditional statement within the `set_memory_range` method in `src/python/gem5/components/memory/dramsim_3.py`. These errors prevent the memory range validation from working as intended.

1.  The `len()` function is incorrectly used with a boolean expression (`ranges != 1`) as its argument. The `len()` function expects an iterable.
2.  The size of the `AddrRange` object (`ranges[0]`) is accessed as an attribute (`ranges[0].size`) instead of calling its method (`ranges[0].size()`).

**Affects version**
This bug was identified in stable version.

**gem5 Modifications**
No modifications were made to the gem5 source code that would cause this bug. The issue is present in the stock `dramsim_3.py` file when attempting to integrate DRAMsim3.

**To Reproduce**
This bug is a static code issue and can be identified through code review or static analysis. It will cause an error or incorrect behavior when the `set_memory_range` method of the `DRAMsim3` class (in `src/python/gem5/components/memory/dramsim_3.py`) is called.

This typically happens during the setup of a simulation using DRAMsim3 as the memory system.

For example, if the code path is executed as is, Python would raise:
1.  A `TypeError` for `len(ranges != 1)`, as `len()` cannot be called on a boolean.
2.  An `AttributeError` for `ranges[0].size` if `AddrRange` does not have a property named `size` (and expects `size()` to be called as a method).

The problematic code is:
```python
    @overrides(AbstractMemorySystem)
    def set_memory_range(self, ranges: List[AddrRange]) -> None:
        if len(ranges != 1) or ranges[0].size != self._size:
            fatal(
                "DRAMsim3 only supports a single contiguous address range "
                "and the range must match the DRAMsim3 size."
            )
        # ... rest of the code
```

**Expected behavior**
The set_memory_range method should correctly validate the provided memory ranges. The conditional statement should be:

```python
if len(ranges) != 1 or ranges[0].size() != self._size:
```

**Host Operating System**
Ubuntu 22.04.

**Host ISA**
X86.

**Compiler used**
gcc 12.3.0

**Additional information**
This bug could lead to incorrect behavior or runtime errors when configuring gem5 with DRAMsim3, specifically when the system is attempting to set and validate the memory address ranges for the DRAMsim3 component. The proposed fix in the "Expected behavior" section should resolve these issues.

h1demasa:
This pull request addresses a bug in the `set_memory_range` method within `src/python/gem5/components/memory/dramsim_3.py`.

**Description of the issues fixed:**

1.  **Incorrect argument type for `len()` function:**
    The original code `len(ranges != 1)` passed a boolean expression to `len()`. This has been corrected to `len(ranges) != 1` to properly check the number of elements in the `ranges` list.

2.  **Incorrect access to `AddrRange` size:**
    The original code `ranges[0].size` attempted to access `size` as an attribute of an `AddrRange` object. This has been corrected to `ranges[0].size()` to call the appropriate method for retrieving the size.

**Reference to Issue:**
#2273

**Code changes:**

The following line in `src/python/gem5/components/memory/dramsim_3.py`:
```python
if len(ranges != 1) or ranges[0].size != self._size:
```

has been changed to:

```python
if len(ranges) != 1 or ranges[0].size() != self._size:
```

This ensures that the memory range validation logic operates as intended.

### DIFF
--- a/src/python/gem5/components/memory/dramsim_3.py
+++ b/src/python/gem5/components/memory/dramsim_3.py
@@ -131,7 +131,7 @@ class SingleChannel(AbstractMemorySystem):
 
     @overrides(AbstractMemorySystem)
     def set_memory_range(self, ranges: List[AddrRange]) -> None:
-        if len(ranges != 1) or ranges[0].size != self._size:
+        if len(ranges) != 1 or ranges[0].size() != self._size:
             raise Exception(
                 "Single channel DRAMSim memory controller requires a single "
                 "range which matches the memory's size."


powerjg:
This change looks good to me. However, the commit message needs to follow our format: E.g., "stdlib: correct memory range check in dramsim3"

erin-le (approved):

PR #2270, opened 2025-05-14
dev: Revert 'Adopt OptionalParam for DmaDevice. (#2260)'

BobbyRBruce:
This reverts commit a06dac9560aa0471ba9d6590294c3a7a5421f435 (#2260).

This was causing the Daily SystemC tests to fail (https://github.com/gem5/gem5/actions/runs/15005141307/job/42161741700).

To reproduce:

```sh
docker run -u $UID:$GID -v $(pwd):/workdir -w /workdir -it --rm ghcr.io/gem5/systemc-env:latest scons build/ARM/gem5.opt -j12
docker run -u $UID:$GID -v $(pwd):/workdir -w /workdir -it --rm ghcr.io/gem5/systemc-env:latest scons setconfig build/ARM --ignore-style USE_SYSTEMC=n
docker run -u $UID:$GID -v $(pwd):/workdir -w /workdir -it --rm ghcr.io/gem5/systemc-env:latest scons build/ARM/libgem5_opt.so --with-cxx-config --without-python --without-tcmalloc -j$(nproc) --duplicate-sources
```

Causes error:

```shell
 [   SHCXX] ARM/cxx_config/DmaVirtDevice.cc -> .os
build/ARM/cxx_config/DmaVirtDevice.cc: In member function ‘virtual bool gem5::DmaVirtDeviceCxxConfigParams::setParam(const std::string&, const std::string&, gem5::CxxConfigParams::Flags)’:
build/ARM/cxx_config/DmaVirtDevice.cc:98:24: error: no matching function for call to ‘to_number(const std::string&, std::optional<unsigned int>&)’
   98 |         ret = to_number(value, this->sid);
      |               ~~~~~~~~~^~~~~~~~~~~~~~~~~~
In file included from build/ARM/sim/serialize_handlers.hh:53,
                 from build/ARM/sim/serialize.hh:61,
                 from build/ARM/base/loader/symtab.hh:54,
                 from build/ARM/sim/system.hh:52,
                 from build/ARM/cxx_config/DmaVirtDevice.cc:8:
build/ARM/base/str.hh:174:1: note: candidate: ‘template<class T> std::enable_if_t<(((is_integral_v<T> || is_floating_point_v<T>) || is_enum_v<T>) && (! is_same_v<bool, T>)), bool> gem5::to_number(const std::string&, T&)’
  174 | to_number(const std::string &value, T &retval)
      | ^~~~~~~~~
build/ARM/base/str.hh:174:1: note:   template argument deduction/substitution failed:
[...]

PR #2275, opened 2025-05-15
scons: Add cxx_config support for OptionalParams

giactra:
The initial implementation of OptionalParams [1] does not support cxx_config builds and therefore breaks some of our CI tests, as reported by a recent PR [2].

The issue is trying to unserialize a std::optional<int> from a config file (ini). The to_number does not allow a std::optional parameter.

While at the beginning I thought about specializing to_number, I realised this is not generic enough as it covers the std::optional<int/unsigned etc> only cases and it does not cover when the parameter is not a number.

[1]: https://github.com/gem5/gem5/pull/2252
[2]: https://github.com/gem5/gem5/pull/2270

Change-Id: Ibb03e26b9390fbdc5b7a369a1c00d5ff625923ce

### DIFF
--- a/build_tools/cxx_config_cc.py
+++ b/build_tools/cxx_config_cc.py
@@ -238,9 +238,24 @@ for param in sim_object._params.values():
         if not is_simobj and not is_vector:
             code('} else if (name == "${{param.name}}") {')
             code.indent()
-            param.ptype.cxx_ini_parse(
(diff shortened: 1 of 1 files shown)


BobbyRBruce (approved):
Thanks @giactra , I was tearing my hair out trying to figure out how to modify cxx_config. This looks great.

powerjg (approved):

PR #2278, opened 2025-05-16
stdlib, tests: Fix ruby cache init signature

shengjiex98:
The {`l1i`/`l1d`/`l2`/`l3`}`_assoc` arguments in `__init__()` of MESI cache classes have type `str`, which is not consistent with other cache classes in the stdlib (e.g., `AbstractTwoLevelCacheHierarchy`). I have changed the type hints of these arguments to `int`.

Also changed 2 files in `tests` to use `int` instead of `str` to be consistent with the change above.


powerjg (approved):
Thanks for the fix!

PR #2280, opened 2025-05-16
dev: add value check for debug flag

yctg:
This commit updates the debug logging condition in register bank, which enable runtime changes of debug flag to be effected on logging.

Test: run unittest of dev/reg_bank.test.xml

Change-Id: Ic4a8e919489b671ee85afe60d779dc0618fb022f

### DIFF
--- a/src/dev/reg_bank.hh
+++ b/src/dev/reg_bank.hh
@@ -1049,7 +1049,7 @@ class RegisterBank : public RegisterBankBase
           const Addr reg_bytes = std::min(reg_size, bytes - done);
 
           if (reg_bytes != reg.size()) {
-              if (_debug_flag) {
+              if (_debug_flag && _debug_flag->tracing()) {
                   ::gem5::trace::getDebugLogger()->dprintf_flag(
                       curTick(), name(), _debug_flag->name(),
                       "Read register %s, byte offset %d, size %d\n",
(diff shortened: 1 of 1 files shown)


src/dev/reg_bank.hh
```diff
@@ -1049,15 +1049,15 @@ class RegisterBank : public RegisterBankBase
           const Addr reg_bytes = std::min(reg_size, bytes - done);
 
           if (reg_bytes != reg.size()) {
-              if (_debug_flag) {
+              if (_debug_flag && *_debug_flag) {
```
powerjg: Could you describe why we're checking the boolean cast for both the pointer and dereferencing?

I think a comment would be useful. Alternatively, using more explicit checks (e.g. `_debug_flag != nullptr` and something like `_debug_flag->enabled()`) would be much clearer.

giactra: I am honestly surprised we got to this point without this PR... 

In the AssociativeCache we use the tracing method (which is effectively called by the overloaded bool operator)

https://github.com/gem5/gem5/blob/stable/src/base/cache/associative_cache.hh#L235

powerjg: Just so I understand better, are you suggesting the change
```suggestion
              if (_debug_flag && _debug_flag->tracing()) {
```

Or, are you saying that `*_debug_flag`is the best way?

giactra: No I was just pointing out this change makes a lot of sense. Dereferencing and calling the tracing method is effectively the same. I have a slight preference towards "debug_flag->tracing" as it seems more explicit (some people would disagree). It's just a matter of "taste" though in general it would be nice to be consistent about it

powerjg: Gotcha. I have a preference for `_debug_flag->tracing()` as well. :)

yctg: Hello all, I'm doing this for checking the enabled status of flags. Thanks for the suggestion! Changed it to `_debug_flag->tracing()`.

giactra (approved):
Thanks for your changes

PR #2282, opened 2025-05-17
arch-riscv: Add RVV support for riscv32 ISA

zhongchengyong:
Previous version configure the mstatus and misa only in riscv64, illegal instruction exception occurs when simulate workload which compiled as riscv-32 target, this commit set the CSR once 'enable_rvv' is true

Test: run scons build/RISCV/unittests.opt


powerjg (commented):
Has this been tested? I worry that there is some implementation details that hardcode rv64 assumptions.

cc: @saul44203

zhongchengyong:
> Has this been tested? I worry that there is some implementation details that hardcode rv64 assumptions.
> 
> cc: @saul44203

How do we test the RVV implementation now? We are testing the vector extension with RV32 by riscv-vector-tests(see: https://github.com/chipsalliance/riscv-vector-tests), it seems that some tests failed on both RV64 and RV32.

powerjg:
> How do we test the RVV implementation now? We are testing the vector extension with RV32 by riscv-vector-tests(see: https://github.com/chipsalliance/riscv-vector-tests), it seems that some tests failed on both RV64 and RV32.

I don't know how it's been tested. I don't think there are any tests in our test suite right now, unfortunately.

I guess I'm simply asking: Have you compiled a program and tested to see if it runs with RVV extensions in 32 bit mode. If so, that would pass the bar for "testing" for right now.

zhongchengyong:
> > How do we test the RVV implementation now? We are testing the vector extension with RV32 by riscv-vector-tests(see: https://github.com/chipsalliance/riscv-vector-tests), it seems that some tests failed on both RV64 and RV32.
> 
> I don't know how it's been tested. I don't think there are any tests in our test suite right now, unfortunately.
> 
> I guess I'm simply asking: Have you compiled a program and tested to see if it runs with RVV extensions in 32 bit mode. If so, that would pass the bar for "testing" for right now.

Of course I do. I have tested with a bunch of RVV testcase which generated by riscv-vector-tests(https://github.com/chipsalliance/riscv-vector-tests), as I said, they are passed or failed on both rv64 and rv32.

powerjg:
The change looks good to me. Can you rebase to resolve the conflicts? After that, I'll approve so we can merge.

PR #2292, opened 2025-05-21
mem,cpu: Make load-to-use constant when L1D hit

pierav:
**Problem**

We have observed that L1D hit cache latency is not always the same when several requests are made during the same cycle. The first response will be executed before the pipeline `tick()` and the others will be executed with a tick increment.

**Exemple**

```
# Cycle 1054000 : cache schedule 3 resps at 1057000 tick
1054000: system.cpu.dcache: access for ReadReq [800000c8:800000cb] hit state: 6 (E) writable: 1 readable: 1 dirty: 0 prefetched: 0 | tag: 0x40000 secure: 0 valid: 1 | set: 0x3 way: 0
1054000: system.cpu.dcache: handleTimingReqHit: schedule resps : 1057000 tick [REQ A]
1054000: system.cpu.dcache: access for ReadReq [800000c8:800000cb] hit state: 6 (E) writable: 1 readable: 1 dirty: 0 prefetched: 0 | tag: 0x40000 secure: 0 valid: 1 | set: 0x3 way: 0
1054000: system.cpu.dcache: handleTimingReqHit: schedule resps : 1057000 tick [REQ B]
1054000: system.cpu.dcache: access for ReadReq [800000c8:800000cb] hit state: 6 (E) writable: 1 readable: 1 dirty: 0 prefetched: 0 | tag: 0x40000 secure: 0 valid: 1 | set: 0x3 way: 0
1054000: system.cpu.dcache: handleTimingReqHit: schedule resps : 1057000 tick [REQ C]


# Cycle 1057000 Before pipeline evaluate: the first req is received
1057000: system.cpu: Dreq(_____InMemory,Vx800000c8,#x4,[R],Dx1000): Received load/store response [REQ A]
1057000: system.cpu.dcache.cpu_side_port-CpuSidePort: schedule : 1057001

# Cycle 1057000: Evaluate the pipeline (WB loads)
1057000: global: -------------- PIPELINE EVALUATE ---------------

# Cycle 1057000 After pipeline evaluate: the second and third req are received
1057001: system.cpu: Dreq(_____InMemory,Vx800000c8,#x4,[R],Dx1000): Received load/store response [REQ B]
1057001: system.cpu.dcache.cpu_side_port-CpuSidePort: schedule : 1057002

1057002: system.cpu: Dreq(_____InMemory,Vx800000c8,#x4,[R],Dx1000): Received load/store response [REQ C]
1057002: system.cpu.dcache.cpu_side_port-CpuSidePort: schedule : 1057003
```

**Solution**

To fix it, we've increased by one tick the scheduling of the response.

Before applying the patch, the distribution of the load-to-use (O3 model) is as follow (for a loop of load hit):
```
system.cpu.lsq0.loadToUse::4                      739     35.94%     35.94% # Distribution of cycle latency between the first time a load is issued and its completion (Unspecified)
system.cpu.lsq0.loadToUse::5                     1285     62.50%     98.44% # Distribution of cycle latency between the first time a load is issued and its completion (Unspecified)
```

After patch:
```
system.cpu.lsq0.loadToUse::5                     2024     98.44%     98.44% # Distribution of cycle latency between the first time a load is issued and its completion (Unspecified)
```

This patch ensures a constant time between send and recv in case of L1D hit.
Note that this patch increases the latency of a single load by 1.

Another implementation might be to unlock all the ready packets (in `transmitList` in `PacketQueue::sendDeferredPacket()`) in one tick without `schedSendEvent`.

### DIFF
--- a/src/cpu/o3/lsq_unit.cc
+++ b/src/cpu/o3/lsq_unit.cc
@@ -275,7 +275,7 @@ LSQUnit::LSQUnitStats::LSQUnitStats(statistics::Group *parent)
                "Number of loads and stores written to the Load Store Queue")
 {
     loadToUse
-        .init(0, 299, 10)
+        .init(0, 299, 1)
         .flags(statistics::nozero);
 }
 
--- a/src/mem/cache/base.hh
+++ b/src/mem/cache/base.hh
@@ -266,6 +266,13 @@ class BaseCache : public ClockedObject
 
         bool isBlocked() const { return blocked; }
 
+        void schedTimingResp(PacketPtr pkt, Tick when) override {
+            // Do when + 1 to avoid misschedule.
+            // This is the case when several schedules have to be carried out
+            // in the same cycle
+            QueuedResponsePort::schedTimingResp(pkt, when + 1);
+        }
+
       protected:
 
         CacheResponsePort(const std::string &_name, BaseCache& _cache,
--- a/src/mem/qport.hh
+++ b/src/mem/qport.hh
@@ -91,7 +91,7 @@ class QueuedResponsePort : public ResponsePort
      * @param pkt Packet to send
      * @param when Absolute time (in ticks) to send packet
      */
-    void schedTimingResp(PacketPtr pkt, Tick when)
+    virtual void schedTimingResp(PacketPtr pkt, Tick when)
     { respQueue.schedSendTiming(pkt, when); }
 
     /** Check the list of buffered packets against the supplied


src/cpu/o3/lsq_unit.cc
```diff
@@ -275,7 +275,7 @@ LSQUnit::LSQUnitStats::LSQUnitStats(statistics::Group *parent)
                "Number of loads and stores written to the Load Store Queue")
 {
     loadToUse
-        .init(0, 299, 10)
+        .init(0, 299, 1)
```
giactra, line 278: I understand you want to get a finer granularity in the L1 hit case, however this makes the stat relatively difficult to read/understand when your load misses L1/L2 and potentially L3.

src/mem/cache/base.hh
```diff
@@ -266,6 +266,13 @@ class BaseCache : public ClockedObject
 
         bool isBlocked() const { return blocked; }
 
+        void schedTimingResp(PacketPtr pkt, Tick when) override {
+            // Do when + 1 to avoid misschedule.
+            // This is the case when several schedules have to be carried out
+            // in the same cycle
+            QueuedResponsePort::schedTimingResp(pkt, when + 1);
+        }
```
giactra, line 274: I wouldn't make the method virtual simply to add +1 to one of the parameters. If we really want to do this, we should just call schedTimingResp with when+1

giactra (commented):
Hi, thanks for your change. I think I've already seen some time ago a PR trying to fix the same "problem".

I have to say I am not particularly keen on further increasing the response time of the dcache. In gem5 a Load/Store will reside in the LSQ/IQ until the data access completes so this means an extra cycle occupying a valuable? slot.

I personally don't think it is so unreasonable to have a cache responding to a limited amount of reqs per cycle, though I understand in our case this behaviour might be un-engineered and not explicitly defined.

The question I have for you is: why  do we evaluate the pipeline after the first response? Can't we just respond at the same time without introducing the extra delay?

powerjg:
What effect does this have if the L1D is 1 cycle away? Does this mean all loads are now 2 cycles?

src/cpu/o3/lsq_unit.cc
pierav: Maybe we should add new loadToUse stats for latencies < 20. Load-to-use is a critical metric.

pierav:
Hi, thanks for the feedback.

Yes, it's reasonable to have a maximum number of requests per cycle, but the problem here is that the requests are spread over 2 cycles.

The first request will be made one cycle before all the others (potentially a lot if the processor's read ports are not limited).

The scheduling priority of the cache may be lower than that of the processor, which is why the first packet is scheduled first.

To not add a cycle delay, all the responses have to be handled in the same tick.

Like this:
```
# Cycle 1054000 : cache schedule 3 resps at 1057000 tick
1054000: system.cpu.dcache: handleTimingReqHit: schedule resps : 1057000 tick [REQ A]
1054000: system.cpu.dcache: handleTimingReqHit: schedule resps : 1057000 tick [REQ B]
1054000: system.cpu.dcache: handleTimingReqHit: schedule resps : 1057000 tick [REQ C]


# Cycle 1057000 Before pipeline evaluate: evalue all reqs
1057000: system.cpu: Dreq(_____InMemory,Vx800000c8,#x4,[R],Dx1000): Received load/store response [REQ A]
1057000: system.cpu: Dreq(_____InMemory,Vx800000c8,#x4,[R],Dx1000): Received load/store response [REQ B]
1057000: system.cpu: Dreq(_____InMemory,Vx800000c8,#x4,[R],Dx1000): Received load/store response [REQ C]

# Cycle 1057000: Evaluate the pipeline (WB loads)
1057000: global: -------------- PIPELINE EVALUATE ---------------
```

To do this we can simply remove the +1 here:
https://github.com/gem5/gem5/blob/82dd7efa8fd9bbee74fba490b77a914a7fbb0ef9/src/mem/packet_queue.cc#L165C9-L168C32

Maybe this adds side effects.

To answer you Jason, with the +1 and a 1 cycle cache latency, the cpu will see the response in the next next cycle.
```
Tick 1000: (New cycle) : Send packed to 1 cycle + 1 tick = 2001
Tick 2000: (New cycle) : Load Stall
Tick 2001:                      : Recv pkt
Tick 3000: (New cycle) : WB load
```

With this patch and a 0 cycle cache latency, the cpu will see the response in the next cycle.  We cannot have 0 cycle load-to-use.
```
Tick 1000: (New cycle) : Send packed to 1001
Tick 1001:             : Recv pkt
Tick 2000: (New cycle) : WB load
```

src/cpu/o3/lsq_unit.cc
powerjg: Adding two stats is OK with me.

src/mem/cache/base.hh
powerjg: I agree.

powerjg (commented):
> To answer you Jason, with the +1 and a 1 cycle cache latency, the cpu will see the response in the next next cycle.

So, does this mean that if the latency is N cycles the load-to-use will always be N+1 cycles on a hit? If so, we should either update the parameter or always subtract 1.

pierav:
Yes, if we have the same definition of the load-to-use. For me, a load-to-use of 0 means that the cache response is obtained in the same cycle. A load-to-use of 1 means that the result is obtained in the next cycle.

So, yes, if the latency is N cycles the load-to-use will always be N+1 cycles on a hit.

Subtracting 1 is tricky because the latency can no longer be set to 0.
But setting a latency of 0 in the cache to obtain the response in the next cycle (1 cycle load-to-use) seems reasonable.

powerjg:
> But setting a latency of 0 in the cache to obtain the response in the next cycle (1 cycle load-to-use) seems reasonable.

I disagree if you increase N by 1. If you set the latency to 1, you would **NOT** expect there to be a 2 cycle load-to-use.

So, if we're going to make this change (and @giactra agrees), we should change the parameter in the cache. Note that this is a difficult change to make because everywhere that the parameter is set we'll need to subtract 1 (this means everywhere in the codebase including all configs, all tests, and the stdlib), **and** any user that has private cache implementations would also have to update.

I am concerned about how this is going to affect not only L1 caches but L2 and L3 caches by 1 cycle as well.

You're right that 0 would now be invalid, but that's an easy fix. We can just have a `fatal` if it's 0 and explain to the user it must be at least 1.

However, I'm concerned about all of the side effects this change is going to have on the broad user community.

pierav:
I think this patch should not involve modifying the API.

Perhaps the simplest solution is not to add a cycle delay. This can be done by deleting the +1 here:
https://github.com/gem5/gem5/blob/82dd7efa8fd9bbee74fba490b77a914a7fbb0ef9/src/mem/packet_queue.cc#L165C9-L168C32
 In this way a latency of 4 always means a load-to-use of 4 on a hit.

Note that previously, for a latency of 4, the load-to-use is distributed between 4 and 5 cycles on a hit.

powerjg:
Sorry for the slow response. If @giactra has feedback it would be good to hear from him too.

I don't think we can change `packet_queue` behavior. This will affect everything in the codebase that uses `packet_queue` (e.g., the memory controller, the caches, etc.). I'm afraid that will lead to unintended side effects.

I fully acknowledge the problem you're trying to solve, but we have two constraints that can't be ignored: 
1. Support single cycle load-to-use
2. We can't cause performance changes in other parts of the code. 

If this patch solves the problem under these constraints, then I'll be happy to merge it!

erin-le:
@pierav: Hello, could you respond to Jason's (powerjg's) comment when you have time? Thank you!

pierav:
Sorry for the late reply.

Fixing this issue requires modifying some code, but we first need to define the scope of the change.

* At a minimum, we apply the fix only to caches (which is what this patch currently does).
* At a maximum, we apply the change to the base PacketQueue class.

We also need to decide whether we want to preserve 0-cycle load-to-use latency.

 * If yes, we need to change this method one way or another:
 https://github.com/gem5/gem5/blob/82dd7efa8fd9bbee74fba490b77a914a7fbb0ef9/src/mem/packet_queue.cc#L167

* If not, adding a 1-tick delay ensures that all responses arrive in the same cycle (which is what this patch does).

To preserve 0-cycle load-to-use in a way that stays consistent with the configuration, while avoiding impact on other parts of the codebase, we could add an option to PacketQueue to control the tick increment (set to 0 for caches, for example)?

Note that a commit from ten years ago suggested revisiting this +1, without explaining what should be done.

https://github.com/gem5/gem5/blob/82dd7efa8fd9bbee74fba490b77a914a7fbb0ef9/src/mem/packet_queue.cc#L168

PR #2294, opened 2025-05-22
mem: Add port name in port_wrapper panic message

yp155136:
This commit adds the name of the port in port_wrapper's panic message to make debugging easier

Change-Id: Ie2f14df88b485889cd3a708140b52dc4eeeeeeee

### DIFF
--- a/src/mem/port_wrapper.cc
+++ b/src/mem/port_wrapper.cc
@@ -48,14 +48,16 @@ RequestPortWrapper::recvRangeChange()
 bool
 RequestPortWrapper::recvTimingResp(PacketPtr packet)
 {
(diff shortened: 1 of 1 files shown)


giactra (approved):
Happy to approve and merge this at the time being.

I believe checking for the callback everytime a request flows through the PortWrapper is a big overhead. We really shouldn't be doing this. We should instead assign a dummy callback at construction time that panics and replace it with the proper one once it gets registered.

Anyway we could change this with a later patch

yp155136:
> We should instead assign a dummy callback at construction time that panics and replace it with the proper one once it gets registered.

This idea SGTM :)

I'll create another PR to implement this!

PR #2298, opened 2025-05-23
dev: reworks PCI to add a PCI host bridge

clemdiep:
Due to mishandling on our original fork, which led us to delete it to start over a clean one, this PR is a repost of #2204. Sorry for the inconvenience.

This commit reworks the PCI architecture to:

  - Add a PCI host bridge
  - Add the possibility to have different upstream between multiple
    `PciDevice` (for future PCI-PCI bridge implementation)
  - Moves the responsibility to response a PCI configuration request in
    `PciDevice` instead of `PciHost`.

For that, the `PciUpstream` class is added and works like the old
`PciHost`, which now inherit from it. `PciDevice` will now need to
register to a `PciUpstream` instead of `PciHost` to get an instance of
`DeviceInterface`, this allows for the future implementation of a
PCI-to-PCI bridge. Also, bus number is now given to the device by its
upstream and not by the gem5 configuration which allow it to be changed
by the OS on boot for PCI-PCI bridge.

`PciUpstream` implementation has two member of the new class
`PciOneWayBridge` to implement the two way bridge two buses (e.g. system
and PCI bus). That class is a bridge with dynamic address ranges
depending on the ones from the connected buses.

The python `PciBus` class is also added, which simply inherit
from `NoncoherentXBar` and set the default port to an `IsaFake` to
answer to configuration request where no device respond.

So, the new PCI architecture is composed of a `PciHost`, connected to a
`PciBus` on PCI side to which all `PciDevice` are connected.

PCI configuration request are now handled by each `PciDevice` instead
of `PciHost` to mimic real PCI configuration.

This is another big step related to #1862.

It has an impact on existing gem5 configuration as now all PCI devices need to be connected behind the `PciHost`. All configurations on the repo have been updated to account that.
However, I'm not sure to understand correctly systems using Ruby and how to connect on it, like with the GPU. For now, I connected PCI device's indirectly through the bridge to Ruby, which works fine.

### DIFF
--- a/src/dev/arm/RealView.py
+++ b/src/dev/arm/RealView.py
@@ -75,6 +75,7 @@ from m5.objects.PciDevice import (
     PciLegacyIoBar,
 )
 from m5.objects.PciHost import *
+from m5.objects.PciUpstream import PciBus
 from m5.objects.Platform import Platform
 from m5.objects.PS2 import *
 from m5.objects.Scmi import *
@@ -847,6 +848,9 @@ class RealView(Platform):
     def _off_chip_memory(self):
         return []
 
+    def _platform_pci_devices(self):
+        return []
+
     _off_chip_ranges = []
 
     def _attach_memory(self, mem, bus, mem_ports=None):
@@ -865,6 +869,11 @@ class RealView(Platform):
             else:
                 dma_ports.append(device.dma)
 
+    def _attach_pci_device(self, device, upstream, bus):
+        device.pio = bus.mem_side_ports
+        device.dma = bus.cpu_side_ports
+        device.upstream = upstream
+
     def _attach_io(self, devices, *args, **kwargs):
         for d in devices:
             self._attach_device(d, *args, **kwargs)
@@ -873,12 +882,18 @@ class RealView(Platform):
         for mem in memories:
             self._attach_memory(mem, *args, **kwargs)
 
+    def _attach_pci(self, devices, bus, dma_ports=None):
+        pass
+
     def _attach_clk(self, devices, clkdomain):
         for d in devices:
             if hasattr(d, "clk_domain"):
                 d.clk_domain = clkdomain
 
-    def attachPciDevices(self):
+    def attachPciDevice(self, device):
+        pass
+
+    def attachPlatformPciDevices(self):
         pass
 
     def enableMSIX(self):
@@ -899,6 +914,7 @@ class RealView(Platform):
     def attachIO(self, bus, dma_ports=None, mem_ports=None):
         self._attach_mem(self._off_chip_memory(), bus, mem_ports)
         self._attach_io(self._off_chip_devices(), bus, dma_ports)
+        self._attach_pci(self._platform_pci_devices(), bus, dma_ports)
 
     def setupBootLoader(self, cur_sys, boot_loader, dtb_addr, load_offset):
         cur_sys.workload.boot_loader = boot_loader
@@ -996,6 +1012,7 @@ class VExpress_EMM(RealView):
         conf_device_bits=16,
         pci_pio_base=0,
     )
+    pci_bus = PciBus()
 
     sys_counter = SystemCounter()
     generic_timer = GenericTimer(
@@ -1033,7 +1050,6 @@ class VExpress_EMM(RealView):
         disks=[],
         pci_func=0,
         pci_dev=0,
-        pci_bus=2,
         io_shift=2,
         ctrl_offset=2,
         Command=0x1,
@@ -1063,13 +1079,11 @@ class VExpress_EMM(RealView):
         devices = [
             self.uart,
             self.realview_io,
-            self.pci_host,
             self.timer0,
             self.timer1,
             self.clcd,
             self.kmi0,
             self.kmi1,
-            self.cf_ctrl,
             self.rtc,
             self.vram,
             self.l2x0_fake,
@@ -1084,6 +1098,12 @@ class VExpress_EMM(RealView):
             self.mmc_fake,
             self.energy_ctrl,
         ]
+        return devices
+
+    def _platform_pci_devices(self):
+        devices = [
+            self.cf_ctrl,
+        ]
         # Try to attach the I/O if it exists
         if hasattr(self, "ide"):
             devices.append(self.ide)
@@ -1092,18 +1112,33 @@ class VExpress_EMM(RealView):
         return devices
 
     # Attach any PCI devices that are supported
-    def attachPciDevices(self):
+    def attachPlatformPciDevices(self):
         self.ethernet = IGbE_e1000(
-            pci_bus=0, pci_dev=0, pci_func=0, InterruptLine=1, InterruptPin=1
+            pci_dev=0, pci_func=0, InterruptLine=1, InterruptPin=1
         )
         self.ide = IdeController(
             disks=[],
-            pci_bus=0,
             pci_dev=1,
             pci_func=0,
             InterruptLine=2,
             InterruptPin=2,
         )
+
+    def attachPciDevice(self, device):
+        self._attach_pci_device(device, self.pci_host, self.pci_bus)
+
+    def _attach_pci(self, devices, bus, dma_ports=None):
+        self.pci_bus.cpu_side_ports = self.pci_host.down_request_port()
+        self.pci_bus.mem_side_ports = self.pci_host.down_response_port()
+
+        bus.mem_side_ports = self.pci_host.up_response_port()
+        if dma_ports is None:
+            bus.cpu_side_ports = self.pci_host.up_request_port()
+        else:
+            dma_ports.append(self.pci_host.up_request_port())
+
+        for d in devices:
+            self._attach_pci_device(d, self.pci_host, self.pci_bus)
 
     def enableMSIX(self):
         self.gic = Gic400(
@@ -1433,6 +1468,7 @@ class VExpress_GEM5_Base(RealView):
         int_base=100,
         int_count=4,
     )
+    pci_bus = PciBus()
 
     energy_ctrl = EnergyCtrl(pio_addr=0x10000000)
 
@@ -1464,7 +1500,6 @@ class VExpress_GEM5_Base(RealView):
             self.kmi1,
             self.watchdog,
             self.rtc,
-            self.pci_host,
             self.energy_ctrl,
             self.pwr_ctrl,
             self.clock32KHz,
@@ -1483,13 +1518,24 @@ class VExpress_GEM5_Base(RealView):
         self.system_watchdog.clk_domain = self.dcc.osc_sys
         self.watchdog.clk_domain = self.clock32KHz
 
-    def attachPciDevice(self, device, *args, **kwargs):
-        device.host = self.pci_host
+    def attachPciDevice(self, device):
         self._num_pci_dev += 1
-        device.pci_bus = 0
         device.pci_dev = self._num_pci_dev
         device.pci_func = 0
-        self._attach_device(device, *args, **kwargs)
+        self._attach_pci_device(device, self.pci_host, self.pci_bus)
+
+    def _attach_pci(self, devices, bus, dma_ports=None):
+        self.pci_bus.cpu_side_ports = self.pci_host.down_request_port()
+        self.pci_bus.mem_side_ports = self.pci_host.down_response_port()
+
+        bus.mem_side_ports = self.pci_host.up_response_port()
+        if dma_ports is None:
+            bus.cpu_side_ports = self.pci_host.up_request_port()
+        else:
+            dma_ports.append(self.pci_host.up_request_port())
+
+        for d in devices:
+            self._attach_pci_device(d, self.pci_host, self.pci_bus)
 
     def attachSmmu(self, devices, bus):
         """
--- a/src/dev/pci/one_way_bridge.cc
+++ b/src/dev/pci/one_way_bridge.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014 ARM Limited
+ * Copyright (c) 2025 REDS institute of the HEIG-VD
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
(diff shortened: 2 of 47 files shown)


erin-le:
@abmerop, @giactra: Hello, just a reminder about this PR. This is a reposted version of https://github.com/gem5/gem5/pull/2204, which was closed and couldn't be reopened due to the authors' management of their branch. It would be helpful if you could carry over your previous reviews to this PR. Thank you!

erin-le:
@giactra: Hello, would you still be interested in reviewing this PR? For context, this PR is a reposted version of https://github.com/gem5/gem5/pull/2204, where you commented that you'd like some time to review it.

erin-le:
Note for future reference: This PR is a repost of https://github.com/gem5/gem5/pull/2204, which abmerop approved with the following comment:

```
Great work! I'm probably not qualified to do a detailed review on the PCI specifics but my tests are all passing and changes to the GPU code look minimal enough
```

src/dev/pci/PciHostBridge.py
```diff
@@ -0,0 +1,57 @@
...
+    cxx_class = "gem5::PciHostBridge"
+    cxx_header = "dev/pci/host_bridge.hh"
+
+    # Bridge mem -> pci
+    mem_response_port = ResponsePort("Response port on memory side")
+    pci_request_port = RequestPort("Request port on PCI side")
+
+    # Bridge pci -> mem
+    pci_response_port = ResponsePort("Response port on PCI side")
+    mem_request_port = RequestPort("Request port on memory side")
```
giactra: Would you consider implementing this with composition of two Bridges?

Something like:

```
class PciHostBridge(ClockedObject):
    type = "PciHostBridge"
    cxx_class = "gem5::PciHostBridge"
    cxx_header = "dev/pci/host_bridge.hh"

    mem_to_pci = Bridge("Bridge mem -> pci")
    pci_to_mem = Bridge("Bridge pci -> mem")

```

This will allow you to avoid copy pasting most of the bridge logic into the PciHostBridge class

src/dev/arm/RealView.py
```diff
@@ -847,6 +848,9 @@ def _on_chip_memory(self):
     def _off_chip_memory(self):
         return []
 
+    def _pci_devices(self):
+        return []
```
giactra: I think this can be confusing as a user might expect calling the method will return a list of instantiated PCI devices, whereas it will only work for the VExpress_EMM platform.

Needless to say this is not your fault at all. Platforms in this module handle PCI devices differently.

giactra (commented):
Hi, this overall looks good to me, though I will wait for your reply on my PciHostBridge comment before doing a thorough review

src/dev/pci/PciHostBridge.py
clemdiep: I considered it at first, but ended up going for the copy pasting.

The main reason was to prepare for the implementation of PCI-to-PCI bridge which can be dynamically reconfigured by the OS, impacting address ranges passing through the bridge. Bridge doesn't allow this and all ranges are defined by Python configuration. The idea was to reuse PciHostBridge for it (well, the name isn't good for that through).

It also seems that addresses alocated to PCI BAR can be split into multiple ranges with other devices in between, making it easier to have a bridge with dynamic range. I'm not entirely sure of this.

Also, PCI configuration requests are no longer handled by a dedicated object (old PciHost) and, in case of access to an invalid configuration (= no device) the error code should be return (all bits set to 1) which I implemented directly in the bridge.

I will reconsider this, now that I have started the implementation of PCI-to-PCI bridge and so have a better understanding of the needs. It will probably require to modify Bridge as well, maybe splitting it into BridgeBase and Bridge? So it can be more easily overriden.

I comeback to you when it's done.

giactra: Thanks @clemdiep, I feel some of the work of defining a bridge with dynamic address ranges is something that could be useful outside of the PCI domain as well.

Really excited to see this contribution in.

> Also, PCI configuration requests are no longer handled by a dedicated object (old PciHost) and, in case of access to an invalid configuration (= no device) the error code should be return (all bits set to 1) which I implemented directly in the bridge.

Isn't this conceptually the same? My understanding has always been that the PciHost was nothing more than a simple version of a PCI-Host bridge. In fact I even wonder whether it makes sense to merge the two classes into a single one (The PciHostBridge). What do we need the PciHost for? Anyway I am not going to add this requirement for now. If you agree with my interpretation we could work it out in a future PR

abmerop (commented):
Similar to my last "review" I will just comment that all my GPUFS tests appear to be working which is good enough for me as I am not a PCIe expert

src/dev/pci/PciHostBridge.py
clemdiep: > Isn't this conceptually the same?

Yes, I just wanted to point this out as it is another thing done by the host bridge which isn't done by Bridge, which only let packets pass through it and never directly responds to them.

> In fact I even wonder whether it makes sense to merge the two classes into a single one. What do we need the PciHost for?

This was to keep the bridge logic (letting packets pass between the two buses) separated from the host (where PCI devices register to a host to get *physical address* for BAR) and avoid a big class making two things. But in the end, the two are used as PciHost as a reference to a PciHostBridge.

But, by using the composition of two Bridges, it will make more sense to merge PciHostBridge and PciHost together. This should not be a big work to do, so I will do it for this PR, so this can reduce code review work (done one time instead of two).

Anyway, I was able to come up with something to have a composition of two Bridges for the host bridge and allows to derive from a BridgeBase to have either fixed ranges or dynamic ones, removing the code duplication. As this affect the Bridge class which outside of PCI and the scope of this PR, I will create a new PR specific for it and update this one to use it. However, I won't be able to finalize it until August and so this PR will be postpone.

clemdiep:
This PR is now in draft as some modification from another PR are necessary (see https://github.com/gem5/gem5/pull/2298#discussion_r2201189918)

clemdiep:
This PR has been updated to use the new BridgeBase merge in #2507 

Main differences with previous change:

  - Added the class PciOneWayBridge inheriting from BridgeBase to have dynamic ranges
  - Replaced PciHostBridge with two instance directly into PciUpstream
  - Replaced the Python class PciXBar by PciBus which connect the XBar default port to a IsaFake to reply to bad configuration request (inspired by MemBus in configs/example/arm/devices.py). The class BadAddr isn't used as it is expected to get request to those address for the PCI discovery.

src/dev/arm/RealView.py
clemdiep: I renamed this to _platform_pci_devices to hopefully remove the confusion.

Another solution would be to remove it as it is only call for _attach_pci, and so VExpress_EMM could directly connect the device in that function.

However, I prefer the current solution, as it mimic how non-pci devices are attached.

src/dev/arm/RealView.py
```diff
@@ -1483,13 +1518,24 @@ def __init__(self, **kwargs):
...
+        self.pci_bus.mem_side_ports = self.pci_host.down_response_port()
+
+        bus.mem_side_ports = self.pci_host.up_response_port()
+        if dma_ports is None:
+            bus.cpu_side_ports = self.pci_host.up_request_port()
+        else:
+            dma_ports.append(self.pci_host.up_request_port())
+
+        for d in devices:
+            self._attach_pci_device(d, self.pci_host, self.pci_bus)
```
giactra, line 1538: Unless I am mistaken this (_attach_pci) doesn't seem to be called for VExpress_GEM5_Base platforms.

These rely on the attachPciDevice method to attach devices https://github.com/gem5/gem5/blob/stable/configs/example/arm/devices.py#L495. So while the device will be rightfully connected, this won't happen for the Host

src/dev/pci/upstream.hh
giactra, line 1: As I presume part of this file has been copied from the existing host.hh file, you should retain the original copyright header and extend it with your contribution (as you correctly did)

src/dev/pci/upstream.cc
giactra, line 1: Same as upstream.hh

giactra (commented):
Hi @clemdiep, thanks for the update. Things look almost ready to go, I just have two comments

src/dev/arm/RealView.py
giactra: Nevermind, I realized it gets called by attachIO

src/dev/pci/upstream.hh
clemdiep: Thanks for your review. I added them to both file.

clemdiep:
I will correct the code format, but I have question regarding some of the change that clang-format is reporting.

  - As of my understanding, openning braces after namespace and struct should be placed on a new line, like for classes?
For example, the current version of the .clang-format force the following change
```diff
-namespace gem5
-{
+namespace gem5 {
```
  - Are single line function authorized in class definition? For example, the following change is made by clang-format
```diff
class PciDevice
{

-    void intrPost() { upstreamInterface.postInt(); }
+    void
+    intrPost()
+    {
+        upstreamInterface.postInt();
+    }

}
```
  - What's the desired formatting for empty function (e.g. for constructor):
```
ClassConstructor::ClassConstructor() {}

ClassConstructor::ClassConstructor()
{}

ClassConstructor::ClassConstructor()
{
}
```
Currently, the thrid one is done by clang-format

As clang-format support is new, I can make the require change in .clang-format to follow the desired formatting.

src/dev/arm/RealView.py
giactra: That is fine. At a certain point I might refactor it myself. Especially as 

1) We want to remove FSConfig
2) We want to remove VExpress_EMM

giactra:
> I will correct the code format, but I have question regarding some of the change that clang-format is reporting.
> 
>     * As of my understanding, openning braces after namespace and struct should be placed on a new line, like for classes?
>       For example, the current version of the .clang-format force the following change
> 
> 
> ```diff
> -namespace gem5
> -{
> +namespace gem5 {
> ```


I have seen both solutions in gem5. I agree we should go for the brace in a newline

> 
>     * Are single line function authorized in class definition? For example, the following change is made by clang-format
> 
> 
> ```diff
> class PciDevice
> {
> 
> -    void intrPost() { upstreamInterface.postInt(); }
> +    void
> +    intrPost()
> +    {
> +        upstreamInterface.postInt();
> +    }
> 
> }
> ```

This is surprising to me as short functions should be allowed on a single line whenever they are inlined:
https://github.com/gem5/gem5/blob/develop/.clang-format#L69C1-L69C33

Maybe that method is not considered as being short?


> 
>     * What's the desired formatting for empty function (e.g. for constructor):
> 
> 
> ```
> ClassConstructor::ClassConstructor() {}
> 
> ClassConstructor::ClassConstructor()
> {}
> 
> ClassConstructor::ClassConstructor()
> {
> }
> ```
> 

Pretty vague. I personally prefer the second option but it is just personal.


> Currently, the thrid one is done by clang-format
> 
> As clang-format support is new, I can make the require change in .clang-format to follow the desired formatting.


Yes please, the idea with using git-clang-format is to iteratively refine the format file to match as closely as possible existing coding style (being aware gem5 is full of inconsistency)

clemdiep:
> > * Are single line function authorized in class definition? For example, the following change is made by clang-format  
> >   
> > ```diff
> > class PciDevice
> > {
> > 
> > -    void intrPost() { upstreamInterface.postInt(); }
> > +    void
> > +    intrPost()
> > +    {
> > +        upstreamInterface.postInt();
> > +    }
> > 
> > }
> > ```
> 
> This is surprising to me as short functions should be allowed on a single line whenever they are inlined: https://github.com/gem5/gem5/blob/develop/.clang-format#L69C1-L69C33
> 
> Maybe that method is not considered as being short?

I found out that it is *ignored* due to the option AlwaysBreakAfterDefinitionReturnType being set to ALL. If that last one is set to TopLevel (= function not in class/struct definition), intrPost is then on a single line, but now function that must be on multiline in class definition have no break after the return type...

> > As clang-format support is new, I can make the require change in .clang-format to follow the desired formatting.
> 
> Yes please, the idea with using git-clang-format is to iteratively refine the format file to match as closely as possible existing coding style (being aware gem5 is full of inconsistency)

I will open a new PR with changes on .clang-format (directly related or not to this PR)

src/dev/pci/one_way_bridge.cc
```diff
@@ -0,0 +1,100 @@
...
+void
+PciOneWayBridge::init()
+{
+    fatal_if(!reverseBridge, "No reverse bridge given for this bridge.\n");
+
+    BridgeBase::init();
+}
+
+void
+PciOneWayBridge::setConfigRange(AddrRange configRange)
```
giactra: config_range

src/dev/pci/one_way_bridge.hh
```diff
@@ -0,0 +1,113 @@
...
+
+  public:
+    void init() override;
+
+    /**
+     * Set the bridge handlig packet for the reverse way.
+     * This shoudl be called before the init phase.
+     */
+    void
+    setReverseBridge(PciOneWayBridge *reverseBridge)
```
giactra: reverse_bridge

src/dev/pci/one_way_bridge.hh
```diff
@@ -0,0 +1,113 @@
...
+    {
+        this->reverseBridge = reverseBridge;
+    }
+
+    /**
+     * Set the PCI configuration range that is behind the bridge.
+     */
+    void setConfigRange(AddrRange configRange);
```
giactra: config_range

src/dev/pci/upstream.hh
```diff
@@ -0,0 +1,317 @@
...
+
+    /**
+     * Calculate the physical address of an IO location on the PCI
+     * bus.
+     *
+     * @param dev_addr The device's position on the PCI bus
+     * @param pci_addr Address in the PCI IO address space
+     * @return Address in the system's physical address space.
+     */
+    virtual Addr interfacePioAddr(const PciDevAddr &dev_addr,
```
giactra, line 257: Minor comment: what does "interface" stand for? As this is returning the upstream view of a downstream address, wouldn't it be more intuitive to call this something like

upstreamPioAddr?

Not a strong opinion; wanna hear your thought. Happy to go with whatever you prefer

giactra (commented):
Minor coding style issues. Otherwise this is good to go from my side.

We should not couple this PR with clang-format issues IMHO. No harm in treating them separately

src/dev/pci/upstream.hh
clemdiep: interface come from the class DeviceInterface, which call those interface* callbacks. I named them this way, so it indicates they are for DeviceInterface and, as we are already in a PciUpstream, I think it's ok to not repeat it

giactra: Ack. the name IMHO is not particularly intuitive but I consider this a minor thing.

giactra:
Hi @clemdiep. In my opinion this PR is ready to be merged. We just need to pass the git-clang-format test. Please rebase your PR and reformat it accordingly :)

giactra (approved):

giragon6:
Hi @clemdiep! It seems like this PR breaks the x86-spec-cpu2017-benchmarks config in configs/example/gem5_library/. 

Here's the error I get when I try to run the benchmarks after this commit:
```
gem5 version DEVELOP-FOR-25.1
gem5 compiled Sep 11 2025 14:39:13
gem5 started Sep 11 2025 14:44:45
command line: gem5/build/ALL/gem5.opt gem5/configs/example/gem5_library/x86-spec-cpu2017-benchmarks.py --image /path/to/spec2017 --benchmark 548.exchange2_r --size test

info: Using default config
AttributeError: object 'PcPciHost' has no attribute 'mem_request_port'
  (C++ object is not yet constructed, so wrapped C++ methods are unavailable.)

At:
  src/python/m5/SimObject.py(876): __getattr__
  src/python/gem5/components/boards/x86_board.py(133): _setup_io_devices
  src/python/gem5/components/boards/x86_board.py(111): _setup_board
  src/python/gem5/components/boards/abstract_board.py(204): _set_fullsystem
  src/python/gem5/components/boards/kernel_disk_workload.py(194): set_kernel_disk_workload
  gem5/configs/example/gem5_library/x86-spec-cpu2017-benchmarks.py(278): <module>
  src/python/m5/main.py(682): main
```

This error isn't present on the previous commit, 531fc25.

clemdiep:
> Hi @clemdiep! It seems like this PR breaks the x86-spec-cpu2017-benchmarks config in configs/example/gem5_library/. 
> 
> Here's the error I get when I try to run the benchmarks after this commit:
> ```
> gem5 version DEVELOP-FOR-25.1
> gem5 compiled Sep 11 2025 14:39:13
> gem5 started Sep 11 2025 14:44:45
> command line: gem5/build/ALL/gem5.opt gem5/configs/example/gem5_library/x86-spec-cpu2017-benchmarks.py --image /path/to/spec2017 --benchmark 548.exchange2_r --size test
> 
> info: Using default config
> AttributeError: object 'PcPciHost' has no attribute 'mem_request_port'
>   (C++ object is not yet constructed, so wrapped C++ methods are unavailable.)
> 
> At:
>   src/python/m5/SimObject.py(876): __getattr__
>   src/python/gem5/components/boards/x86_board.py(133): _setup_io_devices
>   src/python/gem5/components/boards/x86_board.py(111): _setup_board
>   src/python/gem5/components/boards/abstract_board.py(204): _set_fullsystem
>   src/python/gem5/components/boards/kernel_disk_workload.py(194): set_kernel_disk_workload
>   gem5/configs/example/gem5_library/x86-spec-cpu2017-benchmarks.py(278): <module>
>   src/python/m5/main.py(682): main
> ```
> 
> This error isn't present on the previous commit, 531fc25.

Hi @giragon6 

Thanks for your report. This should be fixed by #2570. I am unable to test the benchmark now to validate, but I will try it when I can!
Sorry for the inconvenience.

giragon6:
Hi @clemdiep, that seemed to fix the issue -- thank you!

PR #2307, opened 2025-05-26
arch-riscv: fix vcpyvs µops for vector indexed loads

### LINKED ISSUE #2248: Gem5 RVV Potential Bug: Incorrect Matrix Multiplication Result
**Describe the bug**
When performing a matrix multiplication `C = A_transposed * B` using RISC-V vector extensions (RVV), gem5 produces an incorrect result under specific conditions. The issue occurs when the matrices are allocated with a larger dimension (`computeDepth`) than the dimension used for initialization (`initDepth`). The multiplication loop iterates up to `computeDepth`, potentially reading uninitialized memory regions within the allocated vectors. While scalar execution might handle this by chance (e.g., reading zeros), the RVV implementation seems to produce a result consistent with reading non-zero values (potentially ones) from the uninitialized parts of the input matrices.

In the provided example (`rvv_bug.cpp`), `A_T` and `B` are allocated to hold `computeDepth = 8` elements but initialized only for `initDepth = 4` elements with the value `1`. The expected result for `C[0][0]` is the sum of `A_T[k] * B[k]` for `k` from 0 to 3, which is `1*1 + 1*1 + 1*1 + 1*1 = 4`. However, when simulated with gem5 using RVV, the result is `8`, suggesting the computation included the uninitialized elements (`k=4` to `k=7`) as if they were also `1`.

**To Reproduce**
Steps to reproduce the behavior starting from a clean repository:

1.  Compile the provided C++ code (`rvv_bug.cpp`) using the RISC-V GCC toolchain with RVV enabled:
    ```bash
    riscv64-unknown-elf-g++ -march=rv64gc_zvl256b_v -mabi=lp64d -O2 -Wall -std=c++17 rvv_bug.cpp -o bug.o
    ```

    **`rvv_bug.cpp`:**
    ```cpp
    #include <iostream>
    #include <vector>
    #include <numeric>
    #include <stdexcept>
    #include <sstream>
    #include <cstdint>
    #include <cstddef>

    void multiplyMatrices(
        const std::vector<int16_t>& matrixA_T,
        const std::vector<int16_t>& matrixB,
        std::vector<int32_t>& matrixC,
        const int rowsA,
        const int colsB,
        const int computeDepth)
    {
        std::fill(matrixC.begin(), matrixC.end(), 0);

        for (int i = 0; i < rowsA; ++i) {
            for (int j = 0; j < colsB; ++j) {
                int32_t sum = 0;
                for (int k = 0; k < computeDepth; ++k) {
                    size_t indexA = static_cast<size_t>(k * rowsA + i);
                    size_t indexB = static_cast<size_t>(k * colsB + j);
                    sum += static_cast<int32_t>(matrixA_T[indexA]) *
                           static_cast<int32_t>(matrixB[indexB]);
                }
                matrixC[static_cast<size_t>(i * colsB + j)] = sum;
            }
        }
    }

    template<typename T>
    std::string formatMatrix(const std::vector<T>& matrix, int rows, int cols) {
        if (matrix.size() != static_cast<size_t>(rows * cols))
            return "[Error: Matrix size does not match dimensions]";

        std::stringstream ss;
        for (int i = 0; i < rows; ++i) {
            for (int j = 0; j < cols; ++j)
                ss << matrix[static_cast<size_t>(i * cols + j)] << " ";
            ss << "\n";
        }
        return ss.str();
    }

    int main() {
        const int rowsA = 1;
        const int colsB = 1;
        const int initDepth = 4;
        const int computeDepth = 8;

        std::vector<int16_t> matrixA_T(static_cast<size_t>(computeDepth * rowsA));
        std::vector<int16_t> matrixB(static_cast<size_t>(computeDepth * colsB));
        std::vector<int32_t> matrixC(static_cast<size_t>(rowsA * colsB));

        for (int k = 0; k < initDepth; ++k) {
            for (int i = 0; i < rowsA; ++i)
                matrixA_T[static_cast<size_t>(k * rowsA + i)] = 1;
        }

        for (int k = 0; k < initDepth; ++k) {
            for (int j = 0; j < colsB; ++j)
                matrixB[static_cast<size_t>(k * colsB + j)] = 1;
        }

        std::cout << "Matrix A (Transposed, " << computeDepth << "x" << rowsA << "):\n";
        std::cout << formatMatrix(matrixA_T, computeDepth, rowsA);

        std::cout << "Matrix B (" << computeDepth << "x" << colsB << "):\n";
        std::cout << formatMatrix(matrixB, computeDepth, colsB);

        multiplyMatrices(matrixA_T, matrixB, matrixC, rowsA, colsB, computeDepth);

        std::cout << "Result Matrix C (" << rowsA << "x" << colsB << "):\n";
        std::cout << formatMatrix(matrixC, rowsA, colsB);

        return 0;
    }
    ```

2.  Compile gem5 for RISC-V:
    ```bash
    scons build/RISCV/gem5.opt -j$(nproc)
    ```
    *(Adjust based on your specific build configuration if different)*

3.  Execute the simulation using the provided gem5 configuration script (`simple-riscv.py`) which loads the compiled binary:
    ```bash
    ./build/RISCV/gem5.opt configs/learning_gem5/part1/simple-riscv.py
    ```

    **`configs/learning_gem5/part1/simple-riscv.py`:**
    ```python
    # Copyright (c) 2015 Jason Power
    # All rights reserved.
    #
    # Redistribution and use in source and binary forms, with or without
    # modification, are permitted provided that the following conditions are
    # met: redistributions of source code must retain the above copyright
    # notice, this list of conditions and the following disclaimer;
    # redistributions in binary form must reproduce the above copyright
    # notice, this list of conditions and the following disclaimer in the
    # documentation and/or other materials provided with the distribution;
    # neither the name of the copyright holders nor the names of its
    # contributors may be used to endorse or promote products derived from
    #
    # "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
    # LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
    # A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
    # OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
    # SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
    # LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
    # DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
    # THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT

    import m5
    from m5.objects import *
    import os # Need to import os

    system = System()

    system.clk_domain = SrcClockDomain()
    system.clk_domain.clock = "1GHz"
    system.clk_domain.voltage_domain = VoltageDomain()

    system.mem_mode = "timing"
    system.mem_ranges = [AddrRange("512MiB")]
    system.cpu = RiscvTimingSimpleCPU() # Make sure this CPU model supports RVV if needed, or use one that does explicitly

    system.membus = SystemXBar()

    system.cpu.icache_port = system.membus.cpu_side_ports
    system.cpu.dcache_port = system.membus.cpu_side_ports

    system.cpu.createInterruptController()

    system.mem_ctrl = MemCtrl()
    system.mem_ctrl.dram = DDR3_1600_8x8()
    system.mem_ctrl.dram.range = system.mem_ranges[0]
    system.mem_ctrl.port = system.membus.mem_side_ports

    # --- Correctly locate the binary ---
    # Assumes the script is run from the gem5 root directory
    # and the binary is in 'benchmarks/bin/bug.o' relative to gem5 root
    thispath = os.path.dirname(os.path.realpath(__file__)) # Path to dir containing this script
    gem5_root = os.path.abspath(os.path.join(thispath, "../../../")) # Navigate up to gem5 root
    # binary = os.path.join(gem5_root, "benchmarks/bin/bug.o") # Original path might be incorrect depending on CWD
    binary = "benchmarks/bin/bug.o" # Path relative to gem5 root where binary is placed

    # Check if binary exists at the specified path relative to gem5 root
    binary_full_path = os.path.join(gem5_root, binary)
    if not os.path.exists(binary_full_path):
        # Fallback or alternative path if needed
        alt_binary_path = os.path.join(thispath, "../../../", "bug.o") # Path used in original script logic
        if os.path.exists(alt_binary_path):
             binary = alt_binary_path # Use this if found
        else:
             print(f"Error: Binary '{binary}' not found at expected locations:")
             print(f"- Relative to gem5 root: {binary_full_path}")
             print(f"- Relative to script path: {alt_binary_path}")
             exit(1)
    # -----------------------------------


    system.workload = SEWorkload.init_compatible(binary)

    process = Process()
    process.cmd = [binary]
    system.cpu.workload = process
    system.cpu.createThreads()

    root = Root(full_system=False, system=system)
    m5.instantiate()

    print(f"Beginning simulation!")
    exit_event = m5.simulate()
    ```
    *(Note: Added `import os` and adjusted binary path logic slightly in the script above for robustness)*

**Terminal Output**
The relevant output from the simulation run (specifically the output from the C++ program) shows the incorrect result:

```shell
# Sample gem5 preamble output ...
info: Entering event queue @ 0. Starting simulation...
Beginning simulation!
Result Matrix C (1x1):
8
Exiting @ tick 6304044000 because exiting with last active thread context
# Sample gem5 postamble output ...
```

**Expected behavior**
The C++ program, when executed, should compute the matrix product based on the initialized values. Since only the first `initDepth = 4` elements are initialized to 1, the correct sum for `C[0][0]` should be 4.

The expected terminal output from the C++ program within the gem5 simulation is:

```shell
Result Matrix C (1x1):
4
```

**Additional information**
- The bug seems specific to the interaction between the loop iterating beyond initialized data and the RVV implementation in gem5. Running the same C++ code compiled without `-march=rv64gc_zvl256b_v` (e.g., just `-march=rv64gc`) or natively on an x86 machine produces the expected result of 4.
- This suggests that the vector instructions might be reading memory past the initialized section (`initDepth`) up to the allocated size (`computeDepth`) and interpreting the uninitialized data in a way that contributes to the sum (e.g., treating it as 1), whereas scalar execution or native execution might treat uninitialized stack/heap data as 0 or optimize differently.
- The `RiscvTimingSimpleCPU` was used in the config script. The behavior might differ with other CPU models.

**Disassembly of the problematic inner loop**

Below is the disassembly of the inner loop using RISC-V Vector instructions that is likely causing the problem:

```assembly
112e6: 06805363          blez    s0,1134c <_Z16multiplyMatricesRKSt6vectorIsSaIsEES3_RS_IiSaIiEEiii+0xf4>
112ea: 00093603          ld      a2,0(s2)
112ee: 0009b683          ld      a3,0(s3)
112f2: 0d007ed7          vsetvli t4,zero,e32,m1,ta,ma  # Setting vector mode for 32-bit elements
112f6: 5e0030d7          vmv.v.i v1,0                  # Initialize accumulator vector to zero
112fa: 9672               add     a2,a2,t3
112fc: 96c6               add     a3,a3,a7
112fe: 8722               mv      a4,s0
11300: 08f777d7          vsetvli a5,a4,e16,mf2,tu,ma   # Vector-length for 16-bit elements
11304: 02f805b3          mul     a1,a6,a5
11308: 8f1d               sub     a4,a4,a5
1130a: 06667107          vluxei64.v v2,(a2),v6         # Load from matrixA_T (indexed)
1130e: 0646f187          vluxei64.v v3,(a3),v4         # Load from matrixB (indexed)
11312: f621a0d7          vwmacc.vv v1,v3,v2            # Widening multiply-accumulate
11316: 02f507b3          mul     a5,a0,a5
1131a: 962e               add     a2,a2,a1
1131c: 96be               add     a3,a3,a5
1131e: f36d               bnez    a4,11300 <_Z16multiplyMatricesRKSt6vectorIsSaIsEES3_RS_IiSaIiEEiii+0xa8> # Loop back if not done
11320: 0d0077d7          vsetvli a5,zero,e32,m1,ta,ma  # Setting vector mode for reduction
11324: 42076157          vmv.s.x v2,a4
11328: 021120d7          vredsum.vs v1,v1,v2           # Vector reduction sum
1132c: 42102757          vmv.x.s a4,v1                 # Move result to scalar register
```

saul44203:
The number of `VCpyVsMicroInst` needed for vector indexed loads was being calculated incorrectly by using the destination's EEW instead of the source index's one. This could cause execution bugs in certain situations (#2248).

This patch fixes the issue by using the source index's EEW when calculating the number of µops needed.

### DIFF
--- a/src/arch/riscv/isa/templates/vector_mem.isa
+++ b/src/arch/riscv/isa/templates/vector_mem.isa
@@ -1228,14 +1228,18 @@ template<typename ElemType>
         this->microops.push_back(microop);
     }
 
-    const uint32_t vd_vlmax = vlenb / vd_eewb;
-    const uint8_t num_pinvdcpyvs_microops = ceil((float) this->vl/vd_vlmax);
-    for (uint32_t i = 0; i < num_pinvdcpyvs_microops; i++) {
-        uint32_t vdNumElems = (vl >= vd_vlmax*(i+1)) ? vd_vlmax:vl-vd_vlmax*i;
-
+    const uint32_t vs2_vlmax = vlenb / vs2_eewb;
+    const uint8_t num_cpyvs_microops = ceil((float) this->vl/vs2_vlmax);
+    for (uint32_t i = 0; i < num_cpyvs_microops; i++) {
         microop = new VCpyVsMicroInst(machInst, i, machInst.vs2, elen, vlen);
         microop->setFlag(IsDelayedCommit);
         this->microops.push_back(microop);
+    }
+
+    const uint32_t vd_vlmax = vlenb / vd_eewb;
+    const uint8_t num_pinvd_microops = ceil((float) this->vl/vd_vlmax);
+    for (uint32_t i = 0; i < num_pinvd_microops; i++) {
+        uint32_t vdNumElems = (vl >= vd_vlmax*(i+1)) ? vd_vlmax:vl-vd_vlmax*i;
 
         microop = new VPinVdMicroInst(machInst, i, vdNumElems, elen, vlen);
         microop->setFlag(IsDelayedCommit);


powerjg (approved):
lgtm. I would wait for a reply on #2248 before committing, though.

BobbyRBruce (approved):

BobbyRBruce:
It doesn't look like we are getting a prompt response in #2248 so will merge

PR #2308, opened 2025-05-26
cpu: Add user-mode stats

cmolder:
One major concern when running simulations, especially in FS mode, is whether a simulation's ROI actually executes the benchmark of interest. One major problem here is _kernel foo_: instructions executed by the kernel that don't relate to the benchmark of interest.

This PR tries to make kernel foo more transparent by exposing the number of instructions and ops committed in user and kernel mode. In most cases, a simulation should ideally have few kernel-mode instructions (say, <5%). This PR uses an existing method, `inUserMode()` in gem5's ISA class, to determine if a thread is in user-mode every time it commits an op. From this, we can track the number of user- and kernel-mode instructions and ops committed by each thread.

**Some considerations:**
- Is `inUserMode()` reliable and correct?
- Do the calls to `getIsaPtr()` and `inUserMode()` harm performance? This code is run on every op commit, so ideally we don't want to slow it down. Alternatively, we could track each thread's mode as a boolean variable and flip it when its mode changes.
- It would be nice to have an event similar to MAX_INSTS, but for user-mode instructions only.

Thanks for taking a look at this, I hope it's helpful!

### DIFF
--- a/src/cpu/base.cc
+++ b/src/cpu/base.cc
@@ -219,6 +219,10 @@ BaseCPU::BaseCPU(const Params &p, bool is_checker)
         CommitCPUStats* commitStatptr = new CommitCPUStats(this, i);
         commitStatptr->ipc = commitStatptr->numInsts / baseStats.numCycles;
         commitStatptr->cpi = baseStats.numCycles / commitStatptr->numInsts;
+        commitStatptr->ratioUserInsts = commitStatptr->numUserInsts /
+            commitStatptr->numInsts;
+        commitStatptr->ratioUserOps = commitStatptr->numUserOps /
+            commitStatptr->numOps;
         commitStats.emplace_back(commitStatptr);
     }
 }
@@ -1018,6 +1022,14 @@ CommitCPUStats::CommitCPUStats(statistics::Group *parent, int thread_id)
              "Number of instructions committed excluding NOPs or prefetches"),
     ADD_STAT(numOpsNotNOP, statistics::units::Count::get(),
              "Number of Ops (including micro ops) Simulated"),
+    ADD_STAT(numUserInsts, statistics::units::Count::get(),
+             "Numbrer of instructions committed in user mode"),
+    ADD_STAT(numUserOps, statistics::units::Count::get(),
+            "Number of ops committed in user mode"),
+    ADD_STAT(ratioUserInsts, statistics::units::Ratio::get(),
+             "Ratio of instructions committed in user mode"),
+    ADD_STAT(ratioUserOps, statistics::units::Ratio::get(),
+             "Ratio of ops committed in user mode"),
     ADD_STAT(cpi, statistics::units::Rate<
                 statistics::units::Cycle, statistics::units::Count>::get(),
              "CPI: cycles per instruction (thread level)"),
--- a/src/cpu/base.hh
+++ b/src/cpu/base.hh
@@ -812,6 +812,14 @@ class BaseCPU : public ClockedObject
         statistics::Scalar numInstsNotNOP;
         statistics::Scalar numOpsNotNOP;
 
+        /* Number of instructions committed in user mode */
+        statistics::Scalar numUserInsts;
+        statistics::Scalar numUserOps;
+
+        /* Ratio of instructions committed in user mode */
+        statistics::Formula ratioUserInsts;
+        statistics::Formula ratioUserOps;
+
         /* CPI/IPC for total cycle counts and macro insts */
         statistics::Formula cpi;
         statistics::Formula ipc;
--- a/src/cpu/kvm/base.cc
+++ b/src/cpu/kvm/base.cc
@@ -806,10 +806,16 @@ BaseKvmCPU::kvmRun(Tick ticks)
         ticksExecuted = runTimer->ticksFromHostCycles(hostCyclesExecuted);
 
         /* Update statistics */
-        baseStats.numCycles += simCyclesExecuted;;
-        commitStats[thread->threadId()]->numInsts += instsExecuted;
+        baseStats.numCycles += simCyclesExecuted;
         baseStats.numInsts += instsExecuted;
         ctrInsts += instsExecuted;
+
+        const ThreadID tid = thread->threadId();
+        const bool in_user_mode = thread->getIsaPtr()->inUserMode();
+        commitStats[tid]->numInsts += instsExecuted;
+        if (in_user_mode) {
+            commitStats[tid]->numUserInsts += instsExecuted;
+        }
 
         DPRINTF(KvmRun,
                 "KVM: Executed %i instructions in %i cycles "
--- a/src/cpu/minor/execute.cc
+++ b/src/cpu/minor/execute.cc
@@ -856,8 +856,9 @@ void
 Execute::doInstCommitAccounting(MinorDynInstPtr inst)
 {
     assert(!inst->isFault());
-
-    MinorThread *thread = cpu.threads[inst->id.threadId];
+    const ThreadID tid = inst->id.threadId;
+    MinorThread *thread = cpu.threads[tid];
+    const bool in_user_mode = thread->getIsaPtr()->inUserMode();
 
     /* Increment the many and various inst and op counts in the
      *  thread and system */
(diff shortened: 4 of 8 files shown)


giactra:
Hi @cmolder, thanks for this contribution.

> One major concern when running simulations, especially in FS mode, is whether a simulation's ROI actually executes the benchmark of interest. One major problem here is _kernel foo_: instructions executed by the kernel that don't relate to the benchmark of interest.

I completely understand the requirement, and this will no doubt give a better insight into the userspace/kernelspace instruction ratio for FS simulations.
It's however worth noting that this PR does not necessarily translate into a precise profiler for the benchmark of interest. Not only this won't disambiguate between multiple processes running on the same CPU. Even with a single process there is no guarantee the kernel won't drop to userspace to execute some code (I am thinking of userspace drivers).

If you really wanted to gather data for a specific process/benchmark, you could run with perf-events within gem5 which will ultimately use hw perf counters and other event sources. Moreover, this will also allow you to get other process specific information which this PR does not currently provide (for example, number of benchmark load/stores..)

This is just food for thoughts, I am not opposing to merging this


>     * Do the calls to `getIsaPtr()` and `inUserMode()` harm performance? This code is run on every op commit, so ideally we don't want to slow it down. Alternatively, we could track each thread's mode as a boolean variable and flip it when its mode changes.

That was my concern as well. Did you try benchmarking this change?

powerjg:
> Do the calls to `getIsaPtr()` and `inUserMode()` harm performance? This code is run on every op commit, so ideally we don't want to slow it down. Alternatively, we could track each thread's mode as a boolean variable and flip it when its mode changes.

I highly doubt this will make any difference. There is already a huge amount of code in `commit` so I don't think that one more simple call will cause any performance difference.

Another thing to consider is that some ISAs have more modes than just "user" and "kernel." E.g., RISC-V has "machine" mode as well. Furthermore, hypervisors throw another wrench into the classification.

I suggest two approaches:
1. Count instructions committed while the CPU is in user mode. This seems like a useful statistic, but as Giacomo said, this won't necessarily count the instructions committed by a particular process's code. So, you would have "user mode instructions" and "total instructions."
2. Classify instructions with the mode they were committed in (e.g., ring 0, 1, 2, 3, etc.). The problem with this approach is that it's going to be ISA-specific, unless we just go with ring numbers. Even if we do just go with ring numbers, there is going to have to be an ISA-specific conversion (e.g., instead of calling `inUserMode` we'll have to add a function `getCurrentRing`)

My preference is to go with (1) above. I am also supportive of enabling an exit event on a specific number of user mode instructions. (I have implemented that at least once myself, so there's an existence proof that it's useful.)

giactra:
> > Do the calls to `getIsaPtr()` and `inUserMode()` harm performance? This code is run on every op commit, so ideally we don't want to slow it down. Alternatively, we could track each thread's mode as a boolean variable and flip it when its mode changes.
> 
> I highly doubt this will make any difference. There is already a huge amount of code in `commit` so I don't think that one more simple call will cause any performance difference.

Probably won't make a difference with o3/minor, but what about lean CPU models like Atomic/KVM/NonCaching? It's worth benchmarking it IMHO

> 
> Another thing to consider is that some ISAs have more modes than just "user" and "kernel." E.g., RISC-V has "machine" mode as well. Furthermore, hypervisors throw another wrench into the classification.
> 
> I suggest two approaches:
> 
>     1. Count instructions committed while the CPU is in user mode. This seems like a useful statistic, but as Giacomo said, this won't necessarily count the instructions committed by a particular process's code. So, you would have "user mode instructions" and "total instructions."
> 
>     2. Classify instructions with the mode they were committed in (e.g., ring 0, 1, 2, 3, etc.). The problem with this approach is that it's going to be ISA-specific, unless we just go with ring numbers. Even if we do just go with ring numbers, there is going to have to be an ISA-specific conversion (e.g., instead of calling `inUserMode` we'll have to add a function `getCurrentRing`)
> 
> 
> My preference is to go with (1) above. I am also supportive of enabling an exit event on a specific number of user mode instructions. (I have implemented that at least once myself, so there's an existence proof that it's useful.)

Can I propose a 2b?

Haven't thought this carefully, but what about adding a probe listener to ISA objects? We are already firing the probe when we commit an instruction. In this way the ISA could be notified and every ISA could classify the instruction committed accordingly (ring/EL, etc).
It would even allow to classify per ASID; for example in Arm we could read [TPIDR_EL0](https://developer.arm.com/documentation/ddi0601/2025-03/AArch64-Registers/TPIDR-EL0--EL0-Read-Write-Software-Thread-ID-Register)

So it will appear as an ISA stat. It could also be reused for events different from instruction commit

powerjg:
> Probably won't make a difference with o3/minor, but what about lean CPU models like Atomic/KVM/NonCaching? It's worth benchmarking it IMHO

Getting data instead of going off of my intuition is always a good idea 😄 

> Can I propose a 2b?

This is a great idea, but maybe more than is necessary for @cmolder to do? I worry that it's going to take a lot of engineering on the ISA stats side (e.g., we don't want to use a stats vector for all possible ASIDs). It is also going to take significant work just to get this right in the "common" ISAs (x86, Arm, RISC-V).

I think that a general "user mode or not" is useful in addition to the idea of classifying instructions based on ASID and ring. But, if this doesn't solve a particular problem, then there's no point in adding it.

@cmolder, given the new information from me and Giacomo, do you think that simply having "user mode" vs "not user mode" will solve your problem? Or do you think it's worth diving into Giacomo's suggestion?

cmolder:
Thanks for the feedback. Jason's suggestions make sense to me: at least for my purposes, I only care about whether an instruction is executed in user mode vs. *not* in user mode. Additionally, it aligns nicely with the `inUserMode()` semantics and simplifies the stats a little. So, I've changed the PR to only track user mode instructions and ops.

To generalize this, I could see ISA classes defining an ISA-specific enum which lists all the possible privilege levels/rings supported by that ISA, and stats could be tracked for each privilege level.

For benchmarking, I'll get some wall clock numbers on atomic and kvm using one of the public disk images. If there isn't a slowdown, the code probably fine as-is. But if there is, I can try having each thread store a boolean variable which tracks whether it is in user mode, and change this variable only when the thread's privilege level changes. That would also provide a natural starting point if someone wanted to generalize this to multiple privilege levels.

powerjg:
> I'll get some wall clock numbers on atomic and kvm

I would just check atomic. (1) That's where you'll see the biggest possible difference and (2) kvm doesn't track stats in the same way, so it won't exercise your new code.

Otherwise, the change looks good to me.

BobbyRBruce (approved):

giactra:
This looks like a good compromise to me. Will approve as soon as I get some numbers on reported slowdown (if any)

cmolder:
Here's my `hostSeconds` results from running gem5 on the Atomic CPU while booting the Ubuntu 24.04 kernel. It looks like the change adds about ~4% to the wall clock time on average.

<details>
<summary>Command</summary>

`./build/X86/gem5.fast -- ./configs/example/gem5_library/x86-ubuntu-run-with-kvm.py`

(I replaced `CPUTypes.KVM` to `CPUTypes.ATOMIC` when instantiating the processor)
</details>

<details>
<summary>System</summary>

- **OS**: Ubuntu 20.04.6 LTS x86_64
- **Kernel**: Linux 5.15.0-130-generic
- **CPU**: 2 x Intel(R) Xeon(R) Gold 6226R (64) @ 3.90 GHz
- **Memory**: 93.02 GiB

</details>

  | Old | New | Change
-- | -- | -- | --
Run 1 | 9621.1 | 10033.27 | 4.28%
Run 2 | 9549.32 | 10134.84 | 6.13%
Run 3 | 9514.77 | 9695.1 | 1.90%
Average | 9561.73 | 9954.40333 | 4.11%

erin-le:
@giactra: Hello, could you get back to this PR when you have time? Thank you!

powerjg:
Ping @giactra 

The data looks fine to me. It seems to be only a slight signal above the noise that there's a slowdown.

giactra (approved):
I honestly don't consider 4% to be negligible

1) especially if it is added to CPU models which are meant to be fast
2) especially as it gets executed all the time on SE mode as well where we know for a fact we always execute user mode instructions

I am going to approve this, as I believe the logic is anyway useful; I am still however left with the aftertaste of something than could be engineered better, more efficiently.

As a side note, we should **start benchmarking FS and SE performance on a daily basis**.

powerjg:
>  we should start benchmarking FS and SE performance on a daily basis.

@giactra Agreed! Do you want to take this on? We can find a way to get you access to the server that runs our github actions (and you should have access to all of the data on github).

PR #2311, opened 2025-05-27
mem-ruby: return early response for software prefetches

### LINKED ISSUE #768: Support non-block prefetch inst in Ruby

### LINKED ISSUE #2188: Documentation of why PR 2174 was reverted
This issue has been opened to record the reason for reverting PR 2174. PR 2174 was reverted due to the following failures with the daily tests:

For `testlib-long-tests (fs)`,  `realview64-o3-dual-ruby-ARM_X86-x86_64-opt` and its MatchRegex failed. The following is printed in simerr:
```
src/mem/ruby/system/RubyPort.cc:283: warn: Cache maintenance operations are not supported in Ruby.
gem5.opt: src/cpu/o3/lsq.cc:1164: virtual bool gem5::o3::LSQ::SingleDataRequest::recvTimingResp(gem5::PacketPtr): Assertion `_numOutstandingPackets == 1' failed.
Program aborted at tick 4489520001
--- BEGIN LIBC BACKTRACE ---
```
For `testlib-long-tests (x86_boot_tests)`, `timing-cpu_1-cores_mesi_two_level_DualChannelDDR3_1600_systemd_x86-boot-test-ALL-x86_64-opt` failed. The following is printed in simerr:

```
src/arch/generic/debugfaults.hh:145: warn: MOVNTDQ: Ignoring non-temporal hint, modeling as cacheable!
build/ALL/mem/ruby/protocol/MESI_Two_Level/Directory_Transitions.cc:268: panic: Invalid transition
board.cache_hierarchy.ruby_system.directory_controllers0 time: 15159272260 addr: 0xba174e00 event: CleanReplacement state: M_DRD
Memory Usage: 4323588 KBytes
Program aborted at tick 5048037662580
--- BEGIN LIBC BACKTRACE ---
```

saul44203:
This patch adds support for early response of software prefetches for ruby in a similar way as it's done in the classic cache, preventing these instructions from stalling on the pipeline until the memory request completes. This is done by sending an early response back to the core just as the request arrives (leaving the instruction freedom to commit), and pass through a copy whose response will be catched and deleted before it's sent back to the core.

### DIFF
--- a/src/mem/ruby/system/RubyPort.cc
+++ b/src/mem/ruby/system/RubyPort.cc
@@ -258,6 +258,22 @@ RubyPort::MemResponsePort::recvTimingReq(PacketPtr pkt)
         panic("RubyPort should never see request with the "
               "cacheResponding flag set\n");
 
+    // For software prefetches, immediately respond to avoid stalling on the
+    // core while still processing the prefetch by passing a copy of the
+    // request through
+    PacketPtr pf = nullptr;
+    if (pkt->cmd.isSWPrefetch()) {
+        RequestPtr req = std::make_shared<Request>(pkt->req->getPaddr(),
+                                                   pkt->req->getSize(),
+                                                   pkt->req->getFlags(),
+                                                   pkt->req->requestorId());
+        pf = pkt;
+        pkt = new Packet(req, pf->cmd);
+        pkt->allocate();
+        assert(pkt->matchAddr(pf));
+        assert(pkt->getSize() == pf->getSize());
+    }
+
     // ruby doesn't support cache maintenance operations at the
     // moment, as a workaround, we respond right away
     if (pkt->req->isCacheMaintenance()) {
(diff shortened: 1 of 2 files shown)


src/mem/ruby/system/RubyPort.cc
```diff
@@ -297,6 +313,14 @@ RubyPort::MemResponsePort::recvTimingReq(PacketPtr pkt)
...
     // and return false.
     if (requestStatus == RequestStatus_Issued) {
+        // For software prefetches, only schedule early return if it was
+        // actually issued/aliased by the sequencer
+        if (pkt->cmd.isSWPrefetch()) {
+            assert(pf != nullptr);
+            pf->makeResponse();
+            schedTimingResp(pf, curTick());
+        }
+
```
saul44203, line 323: This is the main difference with #2174.

powerjg (commented):
lgtm. @erin-le is going to re-run the tests and then we can merge.

erin-le (approved):
I reran the daily tests from `fs/linux/arm` and `x86_boot_tests`, which were failing for PR 2174, and they all passed.

PR #2313, opened 2025-05-27
Introduce git-clang-format to gem5 (Simple, Option 1 of 2)

rjc-arch:
This PR introduces `git-clang-format` to gem5, as discussed in [1].

This PR is a simple option 1 of 2, which integrates `git-clang-format` directly into `pre-commit`. I will also submit a PR for a second option that uses a wrapper script to afford better control of the formatting.

The advantage of this version is its simplicity. The disadvantage is that it will block a commit if the `clang-format` tools are not installed, and it offers no control over which files are re-formatted. Only one of these two options is intended to be merged.

I have also included a placeholder `.clang-format` configuration file, but I suggest we instead use the `.clang-format` proposed in PR #362 [2] by @trollet91 and other gem5 contributors. Since this PR does not propose including a large, breaking change that re-formats all the gem5 C/C++ source files at once, the sticking points over this configuration become less important, and we can tweak the configuration incrementally as necessary as time goes by.

[1] https://github.com/orgs/gem5/discussions/333#discussioncomment-8872728
[2] https://github.com/gem5/gem5/pull/362


powerjg:
My preference would be to block commit if clang-format isn't installed. The point is to make the reviewing easier. IMO we should force people into it.

As a side note, I am also all for one giant change to get us up to date on the formatting. The problem with doing it a little along is that reviewing each PR is going to be a pain since it will have both the PR changes and the style changes. (I am fine with a mismatch in style, though.)

Overall, I just want clang-format committed. I will accept *any* PR that does that.

rjc-arch:
Hi Jason. 

Those are all fair points, but after multiple rounds of discussion over the 'ideal' clang-format configuration, git-clang-format seems like a reasonable compromise that would allow us to introduce at least some level of automatic formatting now. Plus, accepting git-clang-format now would not preclude switching to a full clang-format in the future.

I agree it's a bit of a pain to review formatting changes alongside functional changes; however, in effect we are already in that situation, as contributors should already be fixing formatting of the areas of code that they modify, and reviewers do tend to ask for that reformatting. Using git-clang-format just automates the process to avoid requests for reformatting during review, and establishes a reasonable baseline format.

Having said that, personally I don't think the pain of a 'global reformat' should be underestimated. For the core gem5 project it would not be a significant event, but I suspect many organisations maintain either long-running or temporary branches of gem5 (e.g. research projects that have not been published yet) that would be significantly impacted by a global reformat.

rjc-arch:
Regarding this PR and #2314, I would be happy if either were accepted. I could also modify #2314 to block commits without reformatting, if the community sees value in allowing formatting exclusions to specified files. While clang format does allow fine-grained control using comments, a file-level exclusion could be a useful feature during development, or for specific files containing formatted (e.g. tabular) declarative code.

giactra:
My personal preference would be to merge either option 1 or option2.

I agree with most of what @rjc-arch pointed out:

1) Allowing this in won't preclude the possibility of a global formatting later on; if anything it will make it less intrusive and therefore more likely to happen.
2) It will be making reviewing easier for everyone
3) We will start having a checked in clang-format file. People would be able to open PR towards the upstream project to propose modifications to it. At the moment it does require users to snoop inside Tom's fork.
4) We will practically see from our commits if there are some refinements to be made to the clang-format file. Effectively using our commits as a sample, instead of embarking on the daunting task of reviewing the entire codebase.

abmerop:
Am I correct in thinking that Option 1 vs. Option 2 is essentially that Option 1 requires clang-format be installed?  I don't have a strong preference but since Jason has a strong vote I'm leaning towards this option.

I know part of the point here is to ease into clang-format, but I am a little concerned PRs might have a lot of formatting changes. Once nice thing about applying clang-format globally is we could add the commit to `.git-blame-ignore-revs`. We can discuss more next Thursday.

rjc-arch:
The main advantage of Option 2 (with wrapper script) is to allow clang-format to be disabled globally or for individual files using environment variables, for users who don't want to use it on particular files (or for CI, etc). Of course that could also be considered a disadvantage if the project prefers to enforce clang-format.

giactra:
> I know part of the point here is to ease into clang-format, but I am a little concerned PRs might have a lot of formatting changes. Once nice thing about applying clang-format globally is we could add the commit to `.git-blame-ignore-revs`. We can discuss more next Thursday.

Just to be clear, the pre-commit clang-format hook will run automatically on the local machine whenever a user tries to commit code, before the commit gets created. The commit will fail if style inconsistencies are encountered and source code will be modified accordingly.
AFAIU this is even better to what we already do in python with black. In black, the entire file will be re-formatted. In git-clang-format only affected lines.

So PRs won't generally have formatting changes on top of functional ones: whenever a user tries to commit functional changes, the pre-commit hook wil automatically transform the snippet to be clang-format compliant. It will be invisible on the github side.

rjc-arch:
> AFAIU this is even better to what we already do in python with black. In black, the entire file will be re-formatted. In git-clang-format only affected lines.

This is exactly the case. Only lines modified by the commit (and potentially the few surrounding lines, i.e. the git diff context) will be re-formatted by git-clang-format. 

I believe style inconsistencies in these lines would be flagged during manual code review under our current procedures, so I would say the introduction of git-clang-format is simply automating what we have already.

BobbyRBruce:
So, I created a third option: https://github.com/gem5/gem5/pull/2370. WPerhaps I'm missing some context but is there a reason simply using the https://github.com/pre-commit/mirrors-clang-format? It was mentioned in the discussion (#333) and don't see anyone protesting it.

giactra:
> So, I created a third option: #2370. WPerhaps I'm missing some context but is there a reason simply using the https://github.com/pre-commit/mirrors-clang-format? It was mentioned in the discussion (#333) and don't see anyone protesting it.

Hi @BobbyRBruce , Option3 is effectively the same as the [original PR](https://github.com/gem5/gem5/pull/362/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9), minus the changes to the existing codebase. The difference with Option1 and Option2 is that those will only run clang-format over the modified lines (in the commit), whereas your solution will run it on the entire file, therefore it suffers the issue raised by @abmerop in this PR:

> I know part of the point here is to ease into clang-format, but I am a little concerned PRs might have a lot of formatting changes. Once nice thing about applying clang-format globally is we could add the commit to .git-blame-ignore-revs

So a commit adding a single line of code will trigger pre-commit on the entire file, making it hard for someone to review PRs (and confusing blame history)

giactra (dismissed):
Tentateively approving this PR based on our developer's meeting discussion.

Waiting for possible objections from existing reviewers

I believe once we merge this, we should open a PR on top with https://github.com/gem5/gem5/pull/362 clang-format file cherry-picked. Will ping Tom on the mentioned PR

powerjg (dismissed):
🎉

giactra:
This PR is failing since the CI container does not have clang-format installed: https://github.com/gem5/gem5/actions/runs/15274968491/job/43005880182?pr=2313

In order to merge this a container update is required

@BobbyRBruce

BobbyRBruce:
I don't believe this approach will be feasible. The GitHub pre-commit action runs checks on all files modified in a PR, not just the lines changed. Furthermore, a PR may contain numerous commits, requiring the formatter to run on all lines modified across all commits in the PR. I can't figure out how this can be done, at least not without some very bespoke scripting in our GitHub Action Workflows.

Implementing this locally is acceptable (ignoring the requirement of having clang-format installed on the local machine), allowing us to format lines locally on a per-commit basis. However, I don't see how we can achieve this on GitHub Actions. Without this, we can't effectively enforce the use of clang-format.

Unless someone is willing to dedicate time to finding a solution, I believe we may need to go back to our original plan to formatting the entire repository in one go. I'm happy to be proved wrong on this but I'd been sitting for a while trying to figure this out and simply can't find a way.

giactra:
> I don't believe this approach will be feasible. The GitHub pre-commit action runs checks on all files modified in a PR, not just the lines changed. Furthermore, a PR may contain numerous commits, requiring the formatter to run on all lines modified across all commits in the PR. I can't figure out how this can be done, at least not without some very bespoke scripting in our GitHub Action Workflows.

While the action will run pre-commit an all files, the hook under scrutiny executes git-clang-format, and not clang-format.
But because there is no commit being committed, hence no changes, git-clang-format will simply pass on the server side with no disruption.

Feel free to test it yourself with:

`pre-commit run --show-diff-on-failure --color=always --all-files`

So AFAIU, simply installing clang-format in our docker images (which is useful anyway for our users) should fix the GitHub action.


> 
> Unless someone is willing to dedicate time to finding a solution, I believe we may need to go back to our original plan to formatting the entire repository in one go. I'm happy to be proved wrong on this but I'd been sitting for a while trying to figure this out and simply can't find a way.

Even if for some reason Option1 was not feasible for issues on the GitHub side, the closest alternative would be anyway Option2, which allows the CI to bypass the git-clang-format hook.

BobbyRBruce:
> > I don't believe this approach will be feasible. The GitHub pre-commit action runs checks on all files modified in a PR, not just the lines changed. Furthermore, a PR may contain numerous commits, requiring the formatter to run on all lines modified across all commits in the PR. I can't figure out how this can be done, at least not without some very bespoke scripting in our GitHub Action Workflows.
> 
> While the action will run pre-commit an all files, the hook under scrutiny executes git-clang-format, and not clang-format. But because there is no commit being committed, hence no changes, git-clang-format will simply pass on the server side with no disruption.

> 
> Feel free to test it yourself with:
> 
> `pre-commit run --show-diff-on-failure --color=always --all-files`
> 
> So AFAIU, simply installing clang-format in our docker images (which is useful anyway for our users) should fix the GitHub action.

I've pushed a "fix" to this PR to achieve this, but it's not working. The code passes, but it will never run git-clang-commit on a PR successfully. It won't catch anything.

My question, which may provide some clarity, is: what is 'git-clang-format' supposed to run on in this case? It works locally on the staging code before and during a commit, inspecting and formatting only those lines checked. That's fine, but a PR isn't a commit; it's a branch that can contain multiple commits relative to the target branch. If I push a PR with 10 commits, what can our CI tests run to verify that all the lines changed in the PR conform to the style? It could run the check on the last commit in the branch, perhaps, but that's not complete, and there's no real clean interface to interact with the commits in a PR. Even if you have a complete diff that represents the PRs changes to the target branch, I don't know how you can run git-clang-format on that diff. All GitHub Actions does is checkout the PR branch, run tests on it, and the only clever part is that it knows which files have been changed relative to the target branch and therefore passes them to the pre-commit tool for inspection. 


> 
> > Unless someone is willing to dedicate time to finding a solution, I believe we may need to go back to our original plan to formatting the entire repository in one go. I'm happy to be proved wrong on this but I'd been sitting for a while trying to figure this out and simply can't find a way.
>

> Even if for some reason Option1 was not feasible for issues on the GitHub side, the closest alternative would be anyway Option2, which allows the CI to bypass the git-clang-format hook.

I don't believe Option 2 works with GitHub Actions either. The problem isn't with bypassing the hook; we don't want people to bypass hooks too easily (and you can always do it with clang-formats "ignore style" annotations anyway). The issue is that the hook doesn't work in CI tests, so we don't have a style checker to check code with.

rjc-arch:
Hi @BobbyRBruce,

Thanks for looking into this!

I don't think the CI side needs to do anything in this case except 'not fail', as `git-clang-format` should have run on the client side already, and whether or not `git-clang-format` was used, the PR shouldn't be approved and merged unless the code formatting is acceptable. 

So either of the options @giactra mentioned above (installing `git-clang-format` in the container [preferred], or using the second option with the wrapper script and environment variable) should be fine to make sure the CI succeeds.

In other words, I would consider the use of clang-format to be strongly encouraged but optional, and the checking of code formatting to be part of the 'human code review', not part of the automated checks.

rjc-arch:
As an aside, we probably /could/ achieve what you are asking for (automated checking of the code formatting in CI), but it would require a different `git-clang-format` command.

If the GitHub actions can provide you with the hash `BASE_HASH` of the commit the PR is being merged/rebased onto, then you can use the following `git-clang-format` command to re-format all the commits in the PR from that hash to `HEAD`.

```
git-clang-format --quiet --style=file --commit "${BASE_HASH}"
```

However, this would require a different `pre-commit` configuration to the one for running `git-clang-format` locally. We could achieve this by modifying the wrapper script in Option 2, and using an environment variable set in the CI environment to pass in `BASE_HASH` and activate the alternative behaviour. 

Let me know if you'd like me to mock that up, though personally I don't think it is necessary.

.github/workflows/ci-tests.yaml
```diff
@@ -19,6 +19,10 @@ jobs:
         steps:
             - uses: actions/checkout@v4
             - uses: actions/setup-python@v5
+            - run: |
+                  apt update && \
+                  apt install -y clang-format git && \
+                  alias git-clang-format='git clang-format'
```
rjc-arch, line 27: I don't think this `alias` is required. 

`git clang-format` just calls through to `git-clang-format` anyway, and `.pre-commit-config.yaml` calls `git-pre-commit` directly.

rjc-arch (commented):
Adding `clang-format` to the CI container looks good - just one comment about the likely unnecessary alias.

BobbyRBruce:
I deliberately pushed a commit that adds bad formatting to demonstrate how the clang formatting isn't properly executed. This formatting check only runs on the last change in the commit chain. In this instance, the last change doesn't contain bad formatting, so the pre-commit passes even though the PR is clearly incorrect.

giactra:
> > > I don't believe this approach will be feasible. The GitHub pre-commit action runs checks on all files modified in a PR, not just the lines changed. Furthermore, a PR may contain numerous commits, requiring the formatter to run on all lines modified across all commits in the PR. I can't figure out how this can be done, at least not without some very bespoke scripting in our GitHub Action Workflows.
> > 
> > 
> > While the action will run pre-commit an all files, the hook under scrutiny executes git-clang-format, and not clang-format. But because there is no commit being committed, hence no changes, git-clang-format will simply pass on the server side with no disruption.
> 
> > Feel free to test it yourself with:
> > `pre-commit run --show-diff-on-failure --color=always --all-files`
> > So AFAIU, simply installing clang-format in our docker images (which is useful anyway for our users) should fix the GitHub action.
> 
> I've pushed a "fix" to this PR to achieve this, but it's not working. The code passes, but it will never run git-clang-commit on a PR successfully. It won't catch anything.
> 

Yes, I believe the source of misunderstanding in our case is philosophical.
I believe @rjc-arch and myself see the git-clang-format as a best-effort attempt to ease the work of a reviewer and introduce a bit of standardization within the project. The best-effort approach is reflected not only on the fact we format code bit by bit, but also from the idea of building up a clang-format file iteratively via successive refinements.

If you look at it this way, you can see why I don't consider running the hook on GitHub as a necessity, and the "passing" is simply for git-clang-format to not interfere on the server side.

Whereas on your side you want automatic enforcement of the hook for every PR and don't want any PR to be exempted from passing through the git-clang-format filter; you consequently consider the latter case a failure rather than a pass.

I can see your point, but as Richard mentioned above, if you really want such automatic enforcement, it could be possible to do so from git-clang-format directly

> 
> If the GitHub actions can provide you with the hash BASE_HASH of the commit the PR is being merged/rebased onto, then you can use the following git-clang-format command to re-format all the commits in the PR from that hash to HEAD.
> 
> git-clang-format --quiet --style=file --commit "${BASE_HASH}"

rjc-arch:
Hi @BobbyRBruce.

Sorry for all the confusion. As @giactra  mentioned above, I think we are probably aiming for slightly different goals here.

While I am in agreement with Giacomo's comments above, I have just pushed an update to #2314 which adds a `--ci-pr-base-commit` option and `GEM5_CI_PR_BASE_COMMIT` environment variable to the `util/run-git-clang-format.py` wrapper script. 

Using the wrapper, you can now switch the behaviour to "CI mode" by providing a base-commit ID for the PR. I think this should allow you to achieve what you are trying for.

I'm not familiar with GitHub Actions, but something along the lines of the following command should work:

```
GEM5_CI_PR_BASE_COMMIT="${{ github.event.pull_request.base.sha }}" pre-commit run --show-diff-on-failure --color=always --all-files
```

erin-le:
This PR is being closed as Option 2 of `git-clang-format`(https://github.com/gem5/gem5/pull/2314) has been merged.

PR #2314, opened 2025-05-27
Introduce git-clang-format to gem5 with wrapper script (Option 2 of 2)

rjc-arch:
This PR introduces `git-clang-format` to gem5 with a wrapper script, as discussed in [1].

This PR is a more complex option 2 of 2, which integrates `git-clang-format` into `pre-commit` using a wrapper script. I will also submit a PR for a first option that integrates `git-clang-format` directly into `pre-commit`.

The advantage of this version is that it will not block a commit if the `clang-format` tools are not installed, and it offers more control over which files are re-formatted. The disadvantage is that the wrapper script introduces more complexity. Only one of these two options is intended to be merged.

I have also included a placeholder `.clang-format` configuration file, but I suggest we instead use the `.clang-format` proposed in PR #362 [2] by @trollet91 and other gem5 contributors. Since this PR does not propose including a large, breaking change that re-formats all the gem5 C/C++ source files at once, the sticking points over this configuration become less important, and we can tweak the configuration incrementally as necessary as time goes by.

[1] https://github.com/orgs/gem5/discussions/333#discussioncomment-8872728
[2] https://github.com/gem5/gem5/pull/362


util/run-git-clang-format.py
```diff
@@ -0,0 +1,367 @@
...
+    logging.basicConfig(
+        level=logging.INFO,
+        format="%(levelname)7s: %(message)s",
+        stream=sys.stderr,
+    )
+    logger = logging.getLogger(__name__)
+    if args.pre_commit:
+        logger.setLevel(logging.CRITICAL + 1)
+    elif args.verbose:
+        logger.setLevel(logging.DEBUG)
```
giactra: Nit: wouldn't it be better to invert the order? Making sure a verbose option always means a verbose output, even if it gets called from pre-commit?

So instead of current version, doing:
```
    if args.verbose:
        logger.setLevel(logging.DEBUG)
    elif args.pre_commit:
        logger.setLevel(logging.CRITICAL + 1)
        
```

util/run-git-clang-format.py
```diff
@@ -0,0 +1,367 @@
...
+
+    """
+
+    res = run(*command, cwd=cwd, check=True)
+    if res is not None:
+        return res
+    elif silent_exit:
+        sys.exit(0)
+    else:
+        return None
```
giactra: Nit: this code is intuitive and readable as every case is treated on a separate branch . A shorter version (not necessarily more readable though) could be:

```
if res is None and silent_exit:
     sys.exit(0)
else:
    return res
```

Let me know what you prefer. Happy to keep things as they are

util/run-git-clang-format.py
```diff
@@ -0,0 +1,367 @@
...
+    # Get the repository root directory
+    #
+    res = run_silent_exit(
+        "git", "rev-parse", "--show-toplevel", silent_exit=args.pre_commit
+    )
+    if res is None:
+        logger.error("Not a git repository, or git tools not available.")
+        sys.exit(2)
+    repo_root = Path(res.stdout.strip()).resolve()
+    logger.debug("Repository root: '%s'", repo_root)
```
giactra, line 285: Nit: using f-strings?

util/run-git-clang-format.py
```diff
@@ -0,0 +1,367 @@
...
+    # The list combines ignored files specified on the command line
+    # and via the environment variable. Missing files are removed.
+    # Finally, `git` is queried to get a list of staged files, and
+    # ignored files are removed to generate a git pathspec for the
+    # operation. This is necessary because `git-clang-format` does not
+    # support an ignored file list directly.
+    #
+    # Also check the environment variable for a global disable.
+    #
+    if DISABLE_ENV_VAR_NAME in os.environ:
```
giactra, line 301: What happens if I set

`export GEM5_NO_GIT_CLANG_FORMAT=0`

An user might intuitively believe git-clang-format would be enabled. Shall we check if it is a non zero value or would that be a bit too tricky?

util/run-git-clang-format.py
```diff
@@ -0,0 +1,367 @@
...
+    if cwd is None:
+        cwd = Path.cwd()
+
+    logger = logging.getLogger(__name__)
+
+    try:
+        if not command:
+            raise Exception("No command specified.")
+
+        logger.debug("Running command: '%s'", shlex.join(command))
```
giactra, line 110: Nit: using f-strings?

util/run-git-clang-format.py
```diff
@@ -0,0 +1,367 @@
...
+        res = subprocess.run(
+            command,
+            capture_output=True,
+            text=True,
+            check=check,
+            cwd=cwd,
+        )
+
+        logger.debug(
+            "    Command '%s' returned %d", shlex.join(command), res.returncode
```
giactra, line 121: Nit: using f-strings?

util/run-git-clang-format.py
```diff
@@ -0,0 +1,367 @@
...
+        )
+        return res
+    except subprocess.CalledProcessError as ex:
+        logger.error(
+            "Command '%s' failed with retcode %d.", ex.cmd, ex.returncode
```
giactra, line 126: Nit: using f-strings?

util/run-git-clang-format.py
```diff
@@ -0,0 +1,367 @@
...
+        )
+        logger.debug(
+            "Command '%s' stdout:\n%s", shlex.join(command), ex.output
```
giactra, line 129: Nit: using f-strings?

util/run-git-clang-format.py
```diff
@@ -0,0 +1,367 @@
...
+        )
+        logger.debug(
+            "Command '%s' stderr:\n%s", shlex.join(command), ex.stderr
```
giactra, line 132: Nit: using f-strings?

util/run-git-clang-format.py
```diff
@@ -0,0 +1,367 @@
...
+        )
+    except FileNotFoundError as ex:
+        logger.error("Command '%s' not found", command[0])
```
giactra, line 135: Nit: using f-strings?

util/run-git-clang-format.py
```diff
@@ -0,0 +1,367 @@
...
+    except PermissionError as ex:
+        logger.error(
+            "Insufficient permissions to run command '%s'",
+            shlex.join(command),
+        )
+    except Exception as ex:
+        logger.error(
+            "Error running command '%s': %s", shlex.join(command), str(ex)
```
giactra, line 143: Nit: using f-strings?

util/run-git-clang-format.py
```diff
@@ -0,0 +1,367 @@
...
+            "Command '%s' stdout:\n%s", shlex.join(command), ex.output
+        )
+        logger.debug(
+            "Command '%s' stderr:\n%s", shlex.join(command), ex.stderr
+        )
+    except FileNotFoundError as ex:
+        logger.error("Command '%s' not found", command[0])
+    except PermissionError as ex:
+        logger.error(
+            "Insufficient permissions to run command '%s'",
```
giactra, line 138: Nit: using f-strings?

giactra (commented):
Hi @rjc-arch, this looks good to me, I only have some minor comments (mainly style related)

util/run-git-clang-format.py
rjc-arch: That makes sense - I'll make the change.

util/run-git-clang-format.py
rjc-arch: I think both options read just as well, so I'll change to the simpler version.

util/run-git-clang-format.py
rjc-arch: This is a difficult one, and I did consider choosing a specific value to indicate 'disabled'. The problem is that if we go down the path of trying to spot values that mean 'false' then we should probably try to cover other cases such as 'False', 'false', etc. The list would never be exhaustive.

I will update the argparse description string to make it clear that any value for this variable will disable the checks. Do you think that would be enough?

If not, then I think choosing a single true value to check for to activate the feature would be better than trying to spot a set of implicitly false values.

Which option do you prefer?

util/run-git-clang-format.py
rjc-arch: If I remember correctly, the recommendation for the `logging` library is to use string interpolation ('%-strings' or optionally '{}'/'$' interpolation), since the library will defer the interpolation if the logging level is not met. I believe some linters also check for this by default and complain if you use f-strings.

Having said that, I do also prefer the f-string syntax, and for a short script like this I don't think the performance difference is significant.

Considering the above comments, would you prefer me to make the change to f-strings, or should I leave it as is for now?

giactra: Ok, that makes lot of sense and I was not aware of it. Happy to keep things as they are

util/run-git-clang-format.py
giactra: Makes sense! I believe updating the argparse description would be enough. Afterall, if the script does not get disabled and a style inconsistency is found, it will be promptly signalled to the committer. It shouldn't take long for a user to debug the issue

rjc-arch:
Thanks for the review, @giactra! I've just pushed the changes we discussed above.

giactra (approved):
Hi @BobbyRBruce,

This PR enables git-clang-format in gem5 though it avoids perusing the commits under review in GitHub, and automatically "passes" the pre-commit action, which AFAIR is something you don't particularly like.

However, as a follow-up from @rjc-arch latest update in the [Option 1 PR](https://github.com/gem5/gem5/pull/2313#issuecomment-3000745539), the new script version allows you to enable git-clang-format to run as a github action (with --ci-pr-base-commit). This means a following pull request can enable it and therefore enforce its usage on the server side.

I see no harm at the moment on approving this, to signal my belief that the PR is in a mergeable state. I will however wait for you (and any other maintainer) to chime in and approve it before hitting the merge button, as this is something that deserves unanimous approval from the maintainers.

powerjg:
Edit: Nevermind... see below.

~~I was playing around with this today, and I don't know if it's helping us much. Every time I tried to commit something with formatting errors either (1) the gem5 style checker found the error or (2) clang-format didn't catch the error. Here's an example that passes clang-format, but clearly has bad formatting (thanks to claude for the examples).~~

~~What is the expected result here?~~

```c++
#include <cstring>
#include <iostream>
#include <vector>

namespace test {

class BadlyFormatted
{
  public:
    BadlyFormatted( int value  )   {
        this->value=value;
    }

    int getValue()const { return value;}

    void conditionalTest(  ) {
        if (value>10)  {
            std::cout<<"Value is greater than 10"<<std::endl;
        } else{
            std::cout<<"Value is less than or equal to 10"<<std::endl;
        }
    }

    void* getPointerExample()
    {
        int* ptr = nullptr;
        if (value > 0)   { ptr = &value; }
        return (void*)ptr;
    }

  private:
    int value;
};

}  // namespace test

int main(int argc,char**argv) {
    test::BadlyFormatted obj(15);
    obj.conditionalTest();
    return 0;
}
```

powerjg:
I'm going to leave my above comment... however, it looks like the problem was on my end. Since I didn't have `clang-format` installed, pre-commit told me it passed. 

I think we should flip that check and *fail* if clang-format isn't installed. At a minimum, it should say "skipped" not "passed" if it didn't run clang-format. However, I strongly prefer that it is a requirement to be installed to run pre-commit.

rjc-arch:
Hi @powerjg,

In addition to your comment above, the `.clang-format` configuration file included in this PR is just a placeholder, and likely does not fully reflect the gem5 Style guidelines.
The plan was also to merge the `.clang-format` commit from #362. As far as I know, the `.clang-format` from that PR is the best we have so far.

I'd be happy to flip the check on the wrapper script and treat a missing git-clang-format as a failure if you prefer. I chose the current behaviour to make things as frictionless as possible, but it's open to debate. I can do it next week, or am happy for you to make the change if you prefer.

PR #2326, opened 2025-05-30
mem-garnet,mem-ruby: Fixed Garnet_standalone simulation

AlishKanani:
The current version of gem5 encounters some assertions because of recent changes related to DataBlock.

To reproduce the current issue, compile with Garnet_standalone as:

```python3 `which scons` build/Garnet_standalone/gem5.opt -j12```

To reproduce and verify the bug fix, run the following command for a 4 × 4 mesh:

```
./build/Garnet_standalone/gem5.opt configs/example/garnet_synth_traffic.py \
                --num-cpus=16 \
                --num-dirs=16 \
                --network=garnet \
                --topology=Mesh_XY \
                --mesh-rows=4 \
                --synthetic=uniform_random \
                --injectionrate=0.1 \
                --sim-cycles=50000
```

This issue and its fix stems from PR #1453, which removed static methods from RubySystem and inadvertently broke the Garnet_standalone SLICC protocol.

### DIFF
--- a/configs/ruby/Garnet_standalone.py
+++ b/configs/ruby/Garnet_standalone.py
@@ -78,7 +78,11 @@ def create_system(
         # Only one cache exists for this protocol, so by default use the L1D
         # config parameters.
         #
-        cache = L1Cache(size=options.l1d_size, assoc=options.l1d_assoc)
+        cache = L1Cache(
+            size=options.l1d_size,
+            assoc=options.l1d_assoc,
+            block_size=f"{options.cacheline_size}B",
+        )
 
         #
         # Only one unified L1 cache exists.  Can cache instructions and data.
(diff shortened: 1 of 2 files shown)


powerjg (commented):
The config change is ok with me, @abmerop can you take a look at the `StateMachine.py` change?

abmerop:
I believe the StateMachine.py change handles situations where there is a DataBlock defined outside of a structure? E.g., https://github.com/gem5/gem5/blob/stable/src/mem/ruby/protocol/Garnet_standalone-cache.sm#L58 . Is that correct?  It looks like this protocol is that only thing that does this.

AlishKanani:
@abmerop yes, precisely. Basically, Garnet_standalone protocol sends dummy data instead of actual cache lines, which is defined in the protocol file as DataBlock.

abmerop (approved):

mattsinc (commented):
Sorry, I'm way behind but just wanted to say this looks good to me too

PR #2331, opened 2025-05-30
arch-riscv: Allow load-reserved accesses to use the TLB

powerjg:
PR #1387 modified the TLB logic such that load-reserved memory access commands to skip the TLB. This caused problems after the page table walk. This change modifies the logic to allow them to use the TLB as before #1387.

This change also refactors `special_access` variable to be a function in the `MemAccessInfo` object to reduce code duplication.

### DIFF
--- a/src/arch/riscv/pagetable_walker.cc
+++ b/src/arch/riscv/pagetable_walker.cc
@@ -418,12 +418,6 @@ Walker::WalkerState::walkOneStage(Addr vaddr)
 
     curstage = XlateStage::FIRST_STAGE;
 
-    // TLB insert for one stage walks is done in stepwalk
-    // special access is used there!
-    // bool special_access = memaccess.force_virt ||
-    //                             memaccess.hlvx ||
-    //                             memaccess.lr;
-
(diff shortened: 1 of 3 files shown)


BobbyRBruce (approved):

Navyera:
Nice catch and refactor. Looks good to me.

(Perhaps we should look into whether it is always safe to use TLB entries for LR accesses as it is not clear to me right now)

powerjg:
I don't know why spike does this differently, but you have to translate LR/SC accesses when they are virtual addresses. I see no reason these can't go into the TLB. It's likely that spike is a pure functional model, so there's no need for them to have a TLB at all.

erin-le:
@BobbyRBruce: This should be cherry-picked onto staging

rogerchang23424:
Spike doesn't emulate actual guest's CPU TLB (which lookups the guest virtual address to physical address), the spike implement its own TLB translate the guest virtual address directly to host address. The TLB in spike is different with gem5 TLB. The gem5 TLB simulate guest's CPU TLB.

PR #2346, opened 2025-06-05
Add missing operands to disassembly for common arm instructions

bpfoley:
For a number of common instructions, gem5's built-in ARM disassembler doesn't print the instruction operands. This can be a bit of a headache when you don't want to use libcapstone, and are looking at instruction traces.

Reproducer:

test.S
```
    .text
    .globl _start
_start:
    ldr x30, [sp]
    ldr x30, [sp], #16
    str x30, [sp]
    str x30, [sp], #16

    ldp x10, x20, [x30]
    stp x10, x20, [x30]
    ldp w10, w20, [x30]
    stp w10, w20, [x30]
    ldp x10, x20, [x30], #-8  // post-index, negative
    stp x10, x20, [x30, #16]! // pre-index, write-back
    ldp w10, w20, [x30, #32]  // pre-index, no write-back
    stp w10, w20, [x30], #128 // post-index, positive

    dc zva, x17
    dc cvac, x17

    isb
    isb #10

    dsb nsh
    dsb sy
    dsb #8

    .long 0xff210110 // m5_op EXIT
```

```
# Build binary
aarch64-linux-gnu-gcc -c test.S -o test.o
aarch64-linux-gnu-ld -o test.arm64 test.o -N -Ttext 0x10 -static

# objdump disassembly
aarch64-linux-gnu-objdump -D test.arm64

# gem5 disassembly
./build/ARM/gem5.opt configs/example/arm/baremetal.py --tarmac-gen --kernel test.arm64 | grep ' IT '
```

### DIFF
--- a/src/arch/arm/insts/mem64.cc
+++ b/src/arch/arm/insts/mem64.cc
@@ -155,9 +155,9 @@ MemoryPostIndex64::generateDisassembly(
 {
     std::stringstream ss;
     startDisassembly(ss);
-    if (imm)
-        ccprintf(ss, "], #%d", imm);
     ccprintf(ss, "]");
+    if (imm)
+        ccprintf(ss, ", #%d", imm);
     return ss.str();
 }
 
--- a/src/arch/arm/insts/mem64.cc
+++ b/src/arch/arm/insts/mem64.cc
@@ -51,8 +51,7 @@ std::string
 SysDC64::generateDisassembly(Addr pc, const loader::SymbolTable *symtab) const
 {
     std::stringstream ss;
-    printMnemonic(ss, "", false);
-    ccprintf(ss, ", ");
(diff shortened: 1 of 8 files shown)


src/arch/arm/insts/mem64.cc
```diff
@@ -51,8 +51,7 @@ std::string
 SysDC64::generateDisassembly(Addr pc, const loader::SymbolTable *symtab) const
 {
     std::stringstream ss;
-    printMnemonic(ss, "", false);
-    ccprintf(ss, ", ");
+    ss << "  " << mnemonic << ", ";
```
giactra, line 54: This breaks the "assumption" we leave some whitespaces between the mnemonic and the operands... But ultimately it is not important

giactra (approved):

PR #2348, opened 2025-06-05
sim: remove cerr and improve macOS support in init_signals.cc

erin-le:
This commit modifies error messages that previously used cerr to instead use DPRINTF. It also fixes a bug with the external process handler on macOS. Previously, gem5 was unable to open shared memory on macOS when handling an external signal because the shared memory name didn't start with a '/'.


powerjg (approved):

erin-le:
@BobbyRBruce: This should be cherry-picked onto staging

PR #2350, opened 2025-06-05
arch-vega: Included modifiers support in vop3_cmp instructions

german-v-t-um-es:
Includes support for abs and neg modifiers in five F32 vop3_cmp instructions (GE, NGE, GT, NGT, EQ)

### DIFF
--- a/src/arch/amdgpu/vega/insts/vop3_cmp.cc
+++ b/src/arch/amdgpu/vega/insts/vop3_cmp.cc
@@ -1381,6 +1381,28 @@ namespace VegaISA
         src0.readSrc();
         src1.readSrc();
 
+        if (instData.ABS & 0x1) {
+            src0.absModifier();
+        }
+
+        if (instData.ABS & 0x2) {
+            src1.absModifier();
+        }
+
+        if (extData.NEG & 0x1) {
+            src0.negModifier();
+        }
+
+        if (extData.NEG & 0x2) {
+            src1.negModifier();
+        }
+
+        /**
+         * input modifiers are supported by FP operations only
+         */
+        assert(!(instData.ABS & 0x4));
+        assert(!(extData.NEG & 0x4));
+
         for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
             if (wf->execMask(lane)) {
                 sdst.setBit(lane, src0[lane] == src1[lane] ? 1 : 0);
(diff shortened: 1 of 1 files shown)


mattsinc (commented):
@german-v-t-um-es we use these flags in many instructions, but is there a specific reason you added them to these?

german-v-t-um-es:
> @german-v-t-um-es we use these flags in many instructions, but is there a specific reason you added them to these?

@mattsinc While porting some benchmarks into gem5, I found that the failures were caused by comparison instructions using abs and neg modifiers, which weren’t supported. So, I added support for those modifiers in the instructions that I needed, and I figured it would be helpful to share the changes with the simulator so others can benefit too.

mattsinc:
> > @german-v-t-um-es we use these flags in many instructions, but is there a specific reason you added them to these?
> 
> @mattsinc While porting some benchmarks into gem5, I found that the failures were caused by comparison instructions using abs and neg modifiers, which weren’t supported. So, I added support for those modifiers in the instructions that I needed, and I figured it would be helpful to share the changes with the simulator so others can benefit too.

Fair enough ... any tests we can use to validate the support?

german-v-t-um-es:
> > > @german-v-t-um-es we use these flags in many instructions, but is there a specific reason you added them to these?
> > 
> > 
> > @mattsinc While porting some benchmarks into gem5, I found that the failures were caused by comparison instructions using abs and neg modifiers, which weren’t supported. So, I added support for those modifiers in the instructions that I needed, and I figured it would be helpful to share the changes with the simulator so others can benefit too.
> 
> Fair enough ... any tests we can use to validate the support?

Sure, here is a simple HIP program that tests three out of the five instructions I modified for the abs modifier and the EQ instruction for the neg modifier. I wasn’t able to generate the remaining cases in this test due to compiler optimizations.
[cmpTest.zip](https://github.com/user-attachments/files/20619382/cmpTest.zip)

abmerop:
Good find! This is a bit of an oversight on our end.  I see that you added abs/neg to 5 instructions, would you be able to add to all V_CMP_*_F32 instructions? I can help out if needed. From what I can tell, the F32 compare should also support the other two of four modifers: OMOD and CLAMP. I am curious if your test sees these.

mattsinc:
> Good find! This is a bit of an oversight on our end. I see that you added abs/neg to 5 instructions, would you be able to add to all V_CMP_*_F32 instructions? I can help out if needed. From what I can tell, the F32 compare should also support the other two of four modifers: OMOD and CLAMP. I am curious if your test sees these.

+1 to @abmerop 's request

@german-v-t-um-es as long as you've tested them and not just randomly added, that's the most important thing -- thanks

german-v-t-um-es:
Hi @abmerop  and @mattsinc 

In my latest commit, I’ve added support for the abs and neg modifiers to the remaining V_CMP_*_F32 instructions, except for V_CMP_F_F32 and V_CMP_TRU_F32. As I understand it, these two don't require modifier support since they only produce a constant result (0 or 1).

I’ve also extended the test to cover more instructions, although some cases couldn’t be included due to compiler optimizations. [cmpTest_v2.zip](https://github.com/user-attachments/files/20633139/cmpTest_v2.zip)

Regarding the clamp and omod modifiers, I’m not sure how you'd prefer I approach them, since they aren’t currently used with any of the comparison instructions. Let me know how you'd like to proceed.

mattsinc (approved):
thanks!

abmerop (approved):

PR #2351, opened 2025-06-06
arch-x86: automatically exit X86 simulations on kernel panic

erin-le:
This PR attempts to make X86 simulations exit automatically upon kernel panic based on the RISCV and Arm implementations of the same thing. This will only work if the kernel used in the workload has symbols; if it doesn't have symbols, the event to exit simulation on kernel panic won't be added and simulations will continue to hang on kernel panic.

Link to the RISC-V implementation of this, which I heavily referenced: https://github.com/gem5/gem5/commit/cf087d4d11155a9d0a7bb33e397678087a6885a8

### DIFF
--- a/src/arch/x86/linux/fs_workload.hh
+++ b/src/arch/x86/linux/fs_workload.hh
@@ -55,8 +55,8 @@ class FsLinux : public X86ISA::FsWorkload
     PCEvent *kernelOopsPcEvent = nullptr;
     void addExitOnKernelPanicEvent();
     void addExitOnKernelOopsEvent();
-    bool exit_on_kernel_panic;
-    bool exit_on_kernel_oops;
+    // bool exit_on_kernel_panic;
+    // bool exit_on_kernel_oops;
   protected:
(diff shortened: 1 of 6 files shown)


src/arch/x86/linux/fs_workload.hh
```diff
@@ -50,14 +50,28 @@ namespace X86ISA
...
+    PCEvent *kernelPanicPcEvent = nullptr;
+    PCEvent *kernelOopsPcEvent = nullptr;
+    void addExitOnKernelPanicEvent();
+    void addExitOnKernelOopsEvent();
   protected:
     E820Table *e820Table;
 
   public:
-    typedef X86FsLinuxParams Params;
+    PARAMS(X86FsLinux);
```
erin-le, line 62: I don't think this will cause problems, but someone should double check

powerjg: lgtm

BobbyRBruce (approved):

PR #2353, opened 2025-06-06
sim: Add NumericSignalSourcePort.

### LINKED ISSUE #2352: Support type conversion for SignalPort.py
For some generic SimObject, we often want to connect the SignalPort to ports with different types.

```
class GerenicObject(SimObject):
  signal_port = SignalSinkPort('uint32_t')('Signal sink port.')
```

For example, we may want to connect it to different integer types of ports (bool, int, uint32_t).
Is it possible to support the generic port type?

Thanks

dl8sd11:
For some generic models, it's uncertain about
the type of the peer in C++.

Backward compatibility: V

Test: build gem5
Change-Id: I25a453281c0e00b0f9d2af165e78dbbf0dde43eb

### DIFF
--- a/configs/ruby/CHI_config.py
+++ b/configs/ruby/CHI_config.py
@@ -112,8 +112,13 @@ class NoC_Params:
     router_buffer_size = 4
     cntrl_msg_size = 8
     data_width = 32
-    cross_links = []
-    cross_link_latency = 0
+    # Map of (src,dst) -> weight,latency (cycles)
+    # Customizes the weight and latency for the link between the routers src
+    # and dst.
+    # The custom latency applies in both directions unless both (src,dst) and
+    # (dst,src) are present.
+    # A new link is created if no src->dst link exists.
+    custom_links = {}
 
 
 class CHI_Node(SubSystem):
--- a/configs/topologies/CustomMesh.py
+++ b/configs/topologies/CustomMesh.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2021,2022 ARM Limited
+# Copyright (c) 2021,2022,2026 ARM Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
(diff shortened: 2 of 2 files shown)


giactra:
Unless I am missing something, can't you just define your port as

`port = SignalSinkPort("std::variant<uint32_t, bool>")("Port description")`

Without introducing extra code?

dl8sd11:
> Unless I am missing something, can't you just define your port as
> 
> `port = SignalSinkPort("std::variant<uint32_t, bool>")("Port description")`
> 
> Without introducing extra code?


The problem is on the cpp side, we want the source port `SignalSourcePort<variant<uint32_t, bool>>` to bind to a `SignalSinkPort<uint32_t>`. The current implementation in `signal.hh` assumes the source and sink share the same type (also source<T> is a friend class of sink<T>). If we bind a variant source to an integer sink, the bind function fails on https://github.com/gem5/gem5/blob/d22064c1c05f5857431219da888de7e29a268c16/src/sim/signal.hh#L82.

dl8sd11:
Hi @giactra,

Our team is encountering similar type compatibility issues. We'd appreciate any suggestions you might have. Thanks!

giactra:
Hi @dl8sd11 , apologies for the late reply.

> > Unless I am missing something, can't you just define your port as
> > `port = SignalSinkPort("std::variant<uint32_t, bool>")("Port description")`
> > Without introducing extra code?
> 
> The problem is on the cpp side, we want the source port `SignalSourcePort<variant<uint32_t, bool>>` to bind to a `SignalSinkPort<uint32_t>`. The current implementation in `signal.hh` assumes the source and sink share the same type (also source is a friend class of sink). If we bind a variant source to an integer sink, the bind function fails on
> 
> https://github.com/gem5/gem5/blob/d22064c1c05f5857431219da888de7e29a268c16/src/sim/signal.hh#L82
> .

There are two options in front of you:

1) Declare the peer port also to be a variant. 
2) You could specialize the template class for a std::variant so that when you bind you iterate over all alternatives and you throw the fatal only if no option is available (all pointers are nullptr)

dl8sd11:
Hi @giactra, thanks for your suggestion.

I would like to lean to the second option - template specialization because the requirement is to connect to different ports (bool, int, uint) with a single port on a generic gem5 model. We prefer not to modify all the peers.

I draft another version with the template specialization approach which allows int64 source to connect the integral sinks.

https://github.com/gem5/gem5/pull/2528/files

PR #2356, opened 2025-06-06
tests: Update fs tests to use the 24.04 disk images

Harshil2107:


powerjg (approved):

erin-le:
@BobbyRBruce: This should probably be cherry-picked onto staging, but since it only affects the tests, it might also be okay if it stays on develop

PR #2357, opened 2025-06-06
mem: Skip redundant test in AddrRange::isSubset in 1-byte ranges.

adhilton:
AddrRange::isSubset checks if both the start and (end - 1) of one range are contained within another range.  In the case of single-byte ranges (start = end - 1) the second check is redundant and contains may be computationally intensive in some situations. Note that single byte ranges are common as other parts of the code generate them to convert one Addr into an AddrRange.


powerjg (approved):
lgtm

PR #2359, opened 2025-06-09
tests: Revert "Update fs tests to use the 24.04 disk images"

BobbyRBruce:
Reverts gem5/gem5#2356

gem5/gem5#2356 appears to have substantially increased the boot times which has resulted in Daily Test. You can see the ARM, RISK, and X86 boot tests
reaching the 24 hour timeout in our Daily Tests: https://github.com/gem5/gem5/actions/runs/15511028190.

Prior to this PR these tests completed in 4~ish hours (e.g.: https://github.com/gem5/gem5/actions/runs/15498288596).

A >5x execution time going from 20.04 to 24.04 is definitely something that needs fixed. I have not had time to look further into this issue but until there is a fix I think we should revert.

### DIFF
--- a/tests/gem5/arm_boot_tests/configs/arm_boot_exit_run.py
+++ b/tests/gem5/arm_boot_tests/configs/arm_boot_exit_run.py
@@ -201,11 +201,21 @@ board = ArmBoard(
 )
 
 # Set the Full System workload.
-board.set_workload(
-    obtain_resource(
-        "arm-ubuntu-24.04-boot-with-systemd",
+board.set_kernel_disk_workload(
+    kernel=obtain_resource(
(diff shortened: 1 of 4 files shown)


powerjg (changes requested):
Before we revert this, I would like to know where the time is going and why this change causes the slowdown. 

* Is it just that 24.04 does more and so we should expect it to take longer? 
* Did we switch from booting without systemd to with systemd? 
* Is there a new bug in gem5 that has cropped up? 
* Does linux boot hang?
* Are the disk images broken?

These are a few possibilities, but we need to understand things.

BobbyRBruce:
Think I've fixed this issue with #2364 and #2363. Revert not required.

PR #2365, opened 2025-06-11
tests: Reduce Nightly/Weekly X86/ARM Boot tests

BobbyRBruce:
Attemps to reduce the execution time of the X86/ARM boot tests for the Nightly (long) and Weekly (very-long) suites. The changes reduce the number and length of tests with the goal of achieving the same coverage.

The main mains to do this are:

- Where appropriate, exclude systemd from the boot, just boot to the `init` script then exit.
- Remove redundant tests (e.g.., no point in testing 2 cores sim of also testing 8 core sim with all else equal).

### DIFF
--- a/tests/gem5/arm_boot_tests/test_linux_boot.py
+++ b/tests/gem5/arm_boot_tests/test_linux_boot.py
@@ -41,6 +41,7 @@ def test_boot(
     mem_system: str,
     memory_class: str,
     length: str,
+    systemd: bool,
     to_tick: Optional[int] = None,
 ):
     name = f"{cpu}-cpu_{num_cpus}-cores_{mem_system}_{memory_class}_\
@@ -59,6 +60,7 @@ arm_boot_test"
         memory_class,
         "--resource-directory",
         resource_path,
+        "--systemd" if systemd else "--no-systemd",
     ]
 
     if to_tick:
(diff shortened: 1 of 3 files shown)


tests/gem5/arm_boot_tests/test_linux_boot.py
```diff
@@ -152,10 +157,11 @@ def test_boot(
 
 test_boot(
```
powerjg, line 161: This took me 7 hours (for 4 cores). Just FYI.

tests/gem5/arm_boot_tests/test_linux_boot.py
```diff
@@ -42,6 +42,7 @@ def test_boot(
     memory_class: str,
     length: str,
     to_tick: Optional[int] = None,
+    systemd: bool = False,
```
powerjg: I hate to ask, but... can you make this a required parameter so it is explicit in our tests. I think that improves readability.

tests/gem5/arm_boot_tests/test_linux_boot.py
```diff
@@ -152,10 +157,11 @@ def test_boot(
...
     cpu="atomic",
-    num_cpus=4,
+    num_cpus=1,
     mem_system="no_cache",
     memory_class="HBM2Stack",
     length=constants.long_tag,
+    systemd=True,
 )
 
 test_boot(
```
powerjg, line 170: Just to init doesn't take more than a couple of hours in my latest test.

powerjg (dismissed):

powerjg (approved):
Thanks for the change! I didn't realize it would need to be updated so many places, but this is definitely better.

PR #2368, opened 2025-06-11
stdlib: Append resource version to cached resource filename

BobbyRBruce:
Prior to this change only 1 version of a particular resource could be cached at anyone time due to setting the filename to the resource_id. When a different version of an already cached resource was requested the stdlib would delete the cached version to then download the requested version.

This change permits a cache per version of a resource by setting the filename to the resources_id + resource_version.


powerjg (dismissed):
Really, we should be using a uuid as a key, not id + version... but that ship sailed a long time ago.

powerjg (approved):

PR #2371, opened 2025-06-12
arch-arm: Automatically enable stat probe listeners

giactra:
Since a recent PR [1], PMU stats should be regarded as event listeners, so that they can get notified when the event happens and they can update their value accordingly.

In the aforementioned PR, we forgot to enable the listeners.

This patch is fixing this.

[1]: https://github.com/gem5/gem5/pull/2271

Change-Id: I90865f54489af68f510ede080ebf60e3c793a9ef


giactra:
@BobbyRBruce this is the PR I was mentioning, which should be merged into the release staging branch as a hotfix

powerjg (approved):

PR #2375, opened 2025-06-15
[Bug Fix] FIx the bug of MeshDirCorners_XY under MESI_Three_Level

sun123-cmd:
FIx the bug of topology MeshDirCorners_XY under MESI_Three_Level

### DIFF
--- a/configs/ruby/MESI_Three_Level.py
+++ b/configs/ruby/MESI_Three_Level.py
@@ -151,7 +151,7 @@ def create_system(
                 block_size=options.cacheline_size,
             )
 
-            l0_cntrl = L0Cache_Controller(
+            l0_cntrl = MESI_Three_Level_L0Cache_Controller(
                 version=i * num_cpus_per_cluster + j,
                 Icache=l0i_cache,
                 Dcache=l0d_cache,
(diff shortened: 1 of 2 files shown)


powerjg (commented):
The changes to `MESI_Three_Level.py` are OK (though, they should not be necessary if you compile only a single protocol or if you use the `--protocol` option).

For `MeshDirCorners_XY.py`... I am concerned. What is being modeled here? Is the L3 banked in each tile? How is it banked? Generally, that file is *incredibly* brittle since it depends on exactly the right number of L1 and L2 caches and the right number of directories and DMAs. Instead of trying to shoehorn things into this "generic" file, I would strongly suggest writing your own implementation that connects the cache controllers together as is done in the standard library. See https://github.com/gem5/gem5/blob/develop/src/python/gem5/components/cachehierarchies/ruby/mesi_three_level_cache_hierarchy.py for an example.

erin-le:
@sun123-cmd: Hello, could you respond to the previous comment when you get the chance? Thank you!

erin-le:
@sun123-cmd: Hello, just another reminder about this PR. Could you respond to the comment when you get a chance? Thank you!

erin-le:
@sun123-cmd: This PR is being closed for inactivity. If you would like to continue working on this PR, please feel free to reopen. Thank you!

PR #2376, opened 2025-06-16
arch-riscv: Fix PMP checking

rogerchang23424:
From spec:
```
If at least one PMP entry is implemented, but all PMP entries’ A
fields are set to OFF, then all S-mode and U-mode memory accesses
will fail.
```

The `hasLockEntry` will no longer required for checking so the field also removed in the PR

Change-Id: Ibf31fa8ca65cdcfc7ad6d530747d1be74e4139f0

### DIFF
--- a/src/arch/riscv/pmp.cc
+++ b/src/arch/riscv/pmp.cc
@@ -50,8 +50,7 @@ namespace RiscvISA
 PMP::PMP(const Params &params) :
     SimObject(params),
     pmpEntries(params.pmp_entries),
-    numRules(0),
-    hasLockEntry(false)
+    numRules(0)
 {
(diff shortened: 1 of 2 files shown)


powerjg (approved):
Is this a *bug*? (I.e., should we cherry pick this for the release?)

If it doesn't cause incorrect behavior, then we shouldn't cherry-pick, but if there is incorrect execution of a benchmark, then we should.

rogerchang23424:
The PR will cause access fault if the software jumps to lower privilege mode without initizlizing PMP CSRs. The RISC-V ubuntu won't be impacted because the PMP is config well. So I think it's not required.

PR #2377, opened 2025-06-16
sim: Add semihosting ExitEvent

iagocaran:
Semihosting is a common feature between gem5 and QEMU. While working with a coworker with both simulators, we found the necessity to act on semihosting exit, leading to the addition of this event to our configuration scripts.

This PR adds a SEMI_EXIT item to the ExitEvent enum allowing semihosting to exit without a NotImplementedError.

### DIFF
--- a/src/python/gem5/simulate/exit_event.py
+++ b/src/python/gem5/simulate/exit_event.py
@@ -56,6 +56,7 @@ class ExitEvent(Enum):
     PERF_COUNTER_INTERRUPT = "performance counter interrupt"
     KERNEL_PANIC = "kernel panic in simulated system"
     KERNEL_OOPS = "kernel oops in simulated system"
+    SEMI_EXIT = "semihosting exit"
 
     @classmethod
     def translate_exit_status(cls, exit_string: str) -> "ExitEvent":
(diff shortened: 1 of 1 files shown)


powerjg (commented):
I would prefer if you could use the new hypercall support instead of brittle string matching. https://github.com/gem5/gem5/blob/develop/src/sim/sim_exit.hh#L58

Let me know if there's some reason this won't work for your use case.

iagocaran:
I was not aware of the new hypercall function. Since you are recommending a m5 call, is semihosting undesirable in gem5?

To give you some context, I have a single application codebase for both simulators, and since both simulators have semihosting support, we just reused the same code, but the configuration scripts would return with an error because of the missing check on the ExitEvent switch-case.

powerjg:
I'm suggesting calling `exitSimLoopWithHypercall` instead of `exitSimLoop`. If you use `exitSimLoopWithHypercall`, you can provide a default handler in this file: https://github.com/gem5/gem5/lob/develop/src/python/gem5/simulate/exit_handler.py

You could even include rich information (via a python dictionary) if that makes sense for semihosting.

For example:

```c++
exitSimLoopWithHypercall(... 8);
```

```python
class SemiHostingExitHandler(ExitHandler, hypercall_num=8):

    @overrides(ExitHandler)
    def get_handler_description(self):
        return "Exiting because of Semihosting exit."

    @overrides(ExitHandler)
        pass

    @overrides(ExitHandler)
        return True
```

iagocaran:
Understood, I will apply the suggested modifications and adapt the semihosting code accordingly.  

I will mark this PR as a draft while I update the proposed solution.

powerjg:
If it doesn't work out, or it's not straightforward, let me know. If the case is that the new hypercall doesn't work for this exit type, we can make an issue and take it up ourselves.

erin-le:
This PR has been closed for inactivity, but can be reopened if you would like to continue working on it. Thanks for your interest in contributing to gem5!

PR #2378, opened 2025-06-16
arch-arm: Enable new FP extensions in SE mode

giactra:
The following FP&SIMD Arm extensions:

* FEAT_FP16
* FEAT_FRINTTS
* FEAT_FHM

Have been recently implemented in gem5.  While they had been automatically enabled in FS, they weren't added to the default ArmDefaultSE release object. This commit is adding them so that they can be used in SE as well.

Change-Id: Ib903760f5ef84b9b5c69461bc2e874605d0b7281


powerjg (dismissed):

giactra:
Thanks @powerjg for the review. I realised we were not populating the Arm_Frint hwcap so I forced pushed the PR accordingly

powerjg (approved):

PR #2379, opened 2025-06-16
arch-arm: Generate disable PMU ExitEvent when local disable

giactra:
With the ArmPMU it is possible to generate an exit event whenever the PMU gets enabled/disabled. The obvious way to check for it is to catch writes to the PMCR register and generate the exit event accordingly.

While this works well for enabling events, we have seen cases where userspace does not disable the PMU globally at the end of the region of interest, and instead simply disables all counters by writing 0xffff_ffff to the PMCNTENCLR register.

We now cover this case by generating an exit event provided all counters are disabled

Change-Id: I9c487f8dc8e423a3c86ac7bbd7af90b5d0fb9e4d

### DIFF
--- a/src/arch/arm/pmu.cc
+++ b/src/arch/arm/pmu.cc
@@ -212,12 +212,14 @@ PMU::setMiscReg(int misc_reg, RegVal val)
       case MISCREG_PMCNTENSET:
         reg_pmcnten |= val;
         updateAllCounters();
+        pmuExitEvent(false);
         return;
 
       case MISCREG_PMCNTENCLR_EL0:
       case MISCREG_PMCNTENCLR:
         reg_pmcnten &= ~val;
         updateAllCounters();
+        pmuExitEvent(false);
         return;
 
       case MISCREG_PMOVSCLR_EL0:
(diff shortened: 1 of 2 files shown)


src/arch/arm/pmu.cc
```diff
@@ -428,23 +430,36 @@ PMU::setControlReg(PMCR_t val)
...
+        } else {
             inform("Exiting simulation: PMU disable detected");
             exitSimLoop("performance counter disabled", 0);
-        } else if (val.p) {
-            inform("Exiting simulation: PMU reset detected");
-            exitSimLoop("performance counter reset", 0);
         }
+    } else if (check_reset && reg_pmcr.p) {
+        inform("Exiting simulation: PMU reset detected");
+        exitSimLoop("performance counter reset", 0);
```
powerjg, line 461: Would it make sense to use the new hypercall support to have a PMU exit type? This way, instead of having to do string matching to "performance counter reset" you can have a rich type. 

See https://github.com/gem5/gem5/blob/develop/src/sim/sim_exit.hh#L58

giactra: It probably does, though I admit I haven't looked into hypercall exit events; the only info I got comes from the brief introduction we got at the developers meeting.

As my understanding is that it would require changes in the config file, would it make sense to address this with a separate PR?

I am honestly curious to study the new framework especially as I have some ideas on how to make exit event definition/handling more powerful, which might be already covered by hypercalls

powerjg: > As my understanding is that it would require changes in the config file, would it make sense to address this with a separate PR?

No, just to extend this file: https://github.com/gem5/gem5/blob/develop/src/python/gem5/simulate/exit_handler.py The extension may be very simple if you want to just exit the sim loop by default. For instance

```python
class PMUExitHandler(ExitHandler, hypercall_num=<THE NEW HYPERCALL NUMBER>):

    @overrides(ExitHandler)
    def get_handler_description(self):
        return "<SOME DESCRIPTION>"

    @overrides(ExitHandler)
        pass

    @overrides(ExitHandler)
        return True

```

giactra: I have few questions which probably stem from the fact that this is a new thing (AFAIU has not been merged to stable yet).

1) Do we have some documentation / tutorial on how to generate a custom exit handler?
2) Would using exitSimLoopWithHypercall break compatibility with existing configs scripts? For instance https://github.com/gem5/gem5/blob/stable/configs/example/arm/baremetal.py#L219.

If that is the case, I cannot simply use Hypercalls as it would require me to change configs scripts as well, and that is way beyond the scope of this patch.

I believe this PR is already changing relevant PMU functionalities. This change will already increase(fix) the number of exit events when enabling/disabling the PMU. I don't want to mix it with something that breaks backwards compatibility and as such deserves a separate PR.

So I am pushing back against the Hypercall modification, and I'd like to add the following comments:

- Related to point one above, hypercalls are a new thing (not in stable yet), and I'd like to see more documentation and/or see their APIs stabilize before revamping our configs scripts.
- It looks like hypercalls code has been tightly coupled with stdlib (see the simulator parameter in the snippet you posted, which is the ExitHandler ifc AFAIU)

```
    @overrides(ExitHandler)
        pass
```


Does it mean C++ code making use of Hypercalls is forced to use stdlib? @erin-le

powerjg: > Do we have some documentation / tutorial on how to generate a custom exit handler?

This has been on my to do list for a while... It's pretty simple, though.

```python

class MyExitHandler(<The exit handler you want to override>):
    def __init__(self, payload: Dict[str, str]):
        """This is called from C++ with the payload passed into the c++ exit function"""
        # You should probably not every override this.

    def _process(self, simulator):
        """This has the simulator object passed in and is called by the simulator run loop on exit"""
        # Put your logic here. If you want stateful logic, you need to use class variables as a new instance is created each time there is an exit event.

    def _exit_simulation(self):
        """Decides whether to exit the simulation loop after this event"""
        # Return true if you want to stop the simulation, false otherwise
```

> Would using exitSimLoopWithHypercall break compatibility with existing configs scripts? 

Maybe... but if this is a requirement, then it's going to be very difficult, if not impossible, to make exits more powerful (as you mentioned that you're interested in above).

We *might* be able to hack something where we look at the exit event type and take an action in the non-stdlib scripts. But we would not be able to have the nice ExitHandler definitions.

The non-stdlib scripts do a lot of things that are brittle and difficult to maintain. The string matching of exit events is a good example of this. I think that, at least in the long term, we need to deprecate some of these scripts and/or re-write them using the stdlib. Without this, we really can't make forward progress.

powerjg (approved):

PR #2381, opened 2025-06-16
tests, configs: move configs/common into fs/linux/arm test dir

erin-le:
The directory configs/common should be deprecated for most users of gem5, but the Arm tests' reliance on configs/common prevents us from deprecating it. To get around this, we move configs/common to tests/gem5/fs/linux/arm/common to allow the Arm tests to continue using configs/common while preventing most gem5 users from seeing it.


erin-le:
I ran the `quick` and `long` fs/linux/arm tests locally to test this change.

giactra:
Hi @erin-le . This breaks all config scripts in configs/example/arm, which are also relying on the configs/common/cores/arm library.

As I mentioned in a previous PR, we shouldn't be treating all configs/common files in the same way. Mixed with Option.py related modules (deprecated) there are relevant files (like the core models), which, while a bit dated, are structurally sound and follow all best practices we should be enforcing in gem5.

Someone should carefully go through the directory and subdirectory and cathegorize modules accordingly

PR #2392, opened 2025-06-20
cpu: Add Top-Down Methodology

osmanyasar05:
This PR introduces the Top-Down Methodology based on [1], presented at the gem5 Workshop at ISCA 2025. 

[1] Yasin, Ahmad. (2014). A Top-Down method for performance analysis and counters architecture. 35-44. 10.1109/ISPASS.2014.6844459

### DIFF
--- a/src/cpu/o3/commit.cc
+++ b/src/cpu/o3/commit.cc
@@ -165,7 +165,9 @@ Commit::CommitStats::CommitStats(CPU *cpu, Commit *commit)
       ADD_STAT(committedInstType, statistics::units::Count::get(),
                "Class of committed instruction"),
       ADD_STAT(commitEligibleSamples, statistics::units::Cycle::get(),
-               "number cycles where commit BW limit reached")
+               "Number cycles where commit BW limit reached"),
+      ADD_STAT(committedInst, statistics::units::Count::get(),
+               "Number of committed instructions (For Top-Down)")
 {
     using namespace statistics;
 
@@ -1079,6 +1081,7 @@ Commit::commitInsts()
 
     DPRINTF(CommitRate, "%i\n", num_committed);
     stats.numCommittedDist.sample(num_committed);
+    stats.committedInst += num_committed;
 
     if (num_committed == commitWidth) {
         stats.commitEligibleSamples++;
--- a/src/cpu/o3/cpu.cc
+++ b/src/cpu/o3/cpu.cc
@@ -47,6 +47,7 @@
 #include "cpu/checker/thread_context.hh"
 #include "cpu/o3/dyn_inst.hh"
 #include "cpu/o3/limits.hh"
+#include "cpu/o3/tdm.hh"
 #include "cpu/o3/thread_context.hh"
 #include "cpu/simple_thread.hh"
 #include "cpu/thread_context.hh"
@@ -346,7 +347,9 @@ CPU::CPUStats::CPUStats(CPU *cpu)
                "to idling"),
       ADD_STAT(quiesceCycles, statistics::units::Cycle::get(),
                "Total number of cycles that CPU has spent quiesced or waiting "
-               "for an interrupt")
+               "for an interrupt"),
+      topDownStats(cpu, &cpu->fetch, &cpu->rename, &cpu->decode, &cpu->iew,
+                   &cpu->commit)
 {
     // Register any of the O3CPU's stats here.
     timesIdled
--- a/src/cpu/o3/cpu.cc
+++ b/src/cpu/o3/cpu.cc
@@ -115,6 +115,7 @@ CPU::CPU(const BaseO3CPUParams &params)
       globalSeqNum(1),
       system(params.system),
       lastRunningCycle(curCycle()),
+      subtractIdleCycles(params.subtractIdleCycles),
       cpuStats(this)
 {
     fatal_if(FullSystem && params.numThreads > 1,
@@ -1322,6 +1323,10 @@ CPU::wakeCPU()
         --cycles;
         cpuStats.idleCycles += cycles;
         baseStats.numCycles += cycles;
+        if (!subtractIdleCycles) {
+            cpuStats.tdaCycles += cycles;
+        }
+        rename.addIdleCycles(cycles);
     }
 
     schedule(tickEvent, clockEdge());
--- a/src/cpu/o3/cpu.hh
+++ b/src/cpu/o3/cpu.hh
@@ -582,6 +582,8 @@ class CPU : public BaseCPU
         return iew.ldstQueue.getDataPort();
     }
 
+    const bool subtractIdleCycles;
+
     struct CPUStats : public statistics::Group
     {
         CPUStats(CPU *cpu);
@@ -590,9 +592,20 @@ class CPU : public BaseCPU
         statistics::Scalar timesIdled;
         /** Stat for total number of cycles the CPU spends descheduled. */
         statistics::Scalar idleCycles;
+        statistics::Scalar tdaCycles;
         /** Stat for total number of cycles the CPU spends descheduled due to a
          * quiesce operation or waiting for an interrupt. */
         statistics::Scalar quiesceCycles;
+
+        statistics::Scalar fetched;
+        statistics::Scalar decoded;
+        statistics::Scalar renamed;
+        statistics::Scalar dispatched;
+        statistics::Scalar issued;
+        statistics::Scalar executed;
+        statistics::Scalar createdInst;
+        statistics::Scalar squashedInROB;
+
         /** Top Down Methodology stats */
         TopDownStats topDownStats;
 
--- a/src/cpu/o3/cpu.hh
+++ b/src/cpu/o3/cpu.hh
@@ -51,6 +51,8 @@
 
 #include "arch/generic/pcstate.hh"
 #include "base/statistics.hh"
+#include "cpu/activity.hh"
+#include "cpu/base.hh"
 #include "cpu/o3/comm.hh"
 #include "cpu/o3/commit.hh"
 #include "cpu/o3/decode.hh"
@@ -62,9 +64,8 @@
 #include "cpu/o3/rename.hh"
 #include "cpu/o3/rob.hh"
 #include "cpu/o3/scoreboard.hh"
+#include "cpu/o3/tdm.hh"
 #include "cpu/o3/thread_state.hh"
-#include "cpu/activity.hh"
-#include "cpu/base.hh"
 #include "cpu/simple_thread.hh"
 #include "cpu/timebuf.hh"
 #include "params/BaseO3CPU.hh"
@@ -592,6 +593,9 @@ class CPU : public BaseCPU
         /** Stat for total number of cycles the CPU spends descheduled due to a
          * quiesce operation or waiting for an interrupt. */
         statistics::Scalar quiesceCycles;
+        /** Top Down Methodology stats */
+        TopDownStats topDownStats;
+
     } cpuStats;
 
   public:
--- a/src/cpu/o3/decode.hh
+++ b/src/cpu/o3/decode.hh
@@ -316,6 +316,19 @@ class Decode
         /** Stat for total number of squashed instructions. */
         statistics::Scalar squashedInsts;
     } stats;
+
+  public:
+    const DecodeStats &
+    getStats() const
+    {
+        return stats;
+    }
+
+    Cycles
+    getFetchToDecodeDelay()
+    {
+        return fetchToDecodeDelay;
+    }
 };
 
 } // namespace o3
--- a/src/cpu/o3/fetch.cc
+++ b/src/cpu/o3/fetch.cc
@@ -157,43 +157,50 @@ Fetch::regProbePoints()
 
 Fetch::FetchStatGroup::FetchStatGroup(CPU *cpu, Fetch *fetch)
     : statistics::Group(cpu, "fetch"),
-    ADD_STAT(predictedBranches, statistics::units::Count::get(),
-             "Number of branches that fetch has predicted taken"),
-    ADD_STAT(cycles, statistics::units::Cycle::get(),
-             "Number of cycles fetch has run and was not squashing or "
-             "blocked"),
-    ADD_STAT(squashCycles, statistics::units::Cycle::get(),
-             "Number of cycles fetch has spent squashing"),
-    ADD_STAT(tlbCycles, statistics::units::Cycle::get(),
-             "Number of cycles fetch has spent waiting for tlb"),
-    ADD_STAT(idleCycles, statistics::units::Cycle::get(),
-             "Number of cycles fetch was idle"),
-    ADD_STAT(blockedCycles, statistics::units::Cycle::get(),
-             "Number of cycles fetch has spent blocked"),
-    ADD_STAT(miscStallCycles, statistics::units::Cycle::get(),
-             "Number of cycles fetch has spent waiting on interrupts, or bad "
-             "addresses, or out of MSHRs"),
-    ADD_STAT(pendingDrainCycles, statistics::units::Cycle::get(),
-             "Number of cycles fetch has spent waiting on pipes to drain"),
-    ADD_STAT(noActiveThreadStallCycles, statistics::units::Cycle::get(),
-             "Number of stall cycles due to no active thread to fetch from"),
-    ADD_STAT(pendingTrapStallCycles, statistics::units::Cycle::get(),
-             "Number of stall cycles due to pending traps"),
-    ADD_STAT(pendingQuiesceStallCycles, statistics::units::Cycle::get(),
-             "Number of stall cycles due to pending quiesce instructions"),
-    ADD_STAT(icacheWaitRetryStallCycles, statistics::units::Cycle::get(),
-             "Number of stall cycles due to full MSHR"),
-    ADD_STAT(cacheLines, statistics::units::Count::get(),
-             "Number of cache lines fetched"),
-    ADD_STAT(icacheSquashes, statistics::units::Count::get(),
-             "Number of outstanding Icache misses that were squashed"),
-    ADD_STAT(tlbSquashes, statistics::units::Count::get(),
-             "Number of outstanding ITLB misses that were squashed"),
-    ADD_STAT(nisnDist, statistics::units::Count::get(),
-             "Number of instructions fetched each cycle (Total)"),
-    ADD_STAT(idleRate, statistics::units::Ratio::get(),
-             "Ratio of cycles fetch was idle",
-             idleCycles / cpu->baseStats.numCycles)
+      ADD_STAT(predictedBranches, statistics::units::Count::get(),
+               "Number of branches that fetch has predicted taken"),
+      ADD_STAT(cycles, statistics::units::Cycle::get(),
+               "Number of cycles fetch has run and was not squashing or "
+               "blocked"),
+      ADD_STAT(squashCycles, statistics::units::Cycle::get(),
+               "Number of cycles fetch has spent squashing"),
+      ADD_STAT(tlbCycles, statistics::units::Cycle::get(),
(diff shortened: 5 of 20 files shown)


src/cpu/o3/cpu.cc
```diff
@@ -359,6 +360,197 @@ CPU::CPUStats::CPUStats(CPU *cpu)
...
+
+CPU::CPUStats::TopDownStats::TopDownBackendBoundL2::TopDownBackendBoundL2(
+    CPU *cpu)
+    : statistics::Group(cpu, "TopDownL2_BackendBound"),
+      ADD_STAT(memoryBound,
+               statistics::units::Rate<statistics::units::Count,
+                                       statistics::units::Count>::get(),
+               "Memory Bound, backend stalls due to memory subsystem"),
+      ADD_STAT(
+          coreBound,
```
dhschall: Move this in the line up.
Fix this in all the other places as well

src/cpu/o3/cpu.cc
```diff
@@ -359,6 +360,197 @@ CPU::CPUStats::CPUStats(CPU *cpu)
...
+  // Backend Bound / Memory Bound L3
+  auto l1BoundRaw = (cpu->iew.instQueue.getStats().loadStallCycles -
+                     cpu->iew.instQueue.getStats().L1miss) /
+                    (cpu->baseStats.numCycles);
+  auto l2BoundRaw = (cpu->iew.instQueue.getStats().L1miss -
+                     cpu->iew.instQueue.getStats().L2miss) /
+                    (cpu->baseStats.numCycles);
+  auto l3BoundRaw = (cpu->iew.instQueue.getStats().L2miss -
+                  cpu->iew.instQueue.getStats().L3miss) /
+                 (cpu->baseStats.numCycles);
```
dhschall: Wrong indentation

src/cpu/o3/cpu.hh
```diff
@@ -584,6 +584,48 @@ class CPU : public BaseCPU
         /** Stat for total number of cycles the CPU spends descheduled due to a
          * quiesce operation or waiting for an interrupt. */
         statistics::Scalar quiesceCycles;
+
+        struct TopDownStats : statistics::Group {
+          TopDownStats(CPU *cpu);
+
+          struct TopDownL1 : statistics::Group {
+            TopDownL1(CPU *cpu);
```
dhschall: wrong indentation. 4 spaces

src/cpu/o3/decode.cc
```diff
@@ -587,9 +587,9 @@ Decode::decode(bool &status_change, ThreadID tid)
     //     check if stall conditions have passed
 
     if (decodeStatus[tid] == Blocked) {
-        ++stats.blockedCycles;
+      ++stats.blockedCycles;
```
dhschall: wrong indentation

src/cpu/o3/fu_pool.hh
```diff
@@ -201,6 +201,16 @@ class FUPool : public SimObject
 
     /** Takes over from another CPU's thread. */
     void takeOverFrom() {};
+
+    /** Returns the number of free FUs */
+    int numBusyFUs() const {
+      int busy = 0;
```
dhschall: wrong indentation

src/cpu/o3/inst_queue.cc
```diff
@@ -923,6 +942,30 @@ InstructionQueue::scheduleReadyInsts()
     } else {
         DPRINTF(IQ, "Not able to schedule any instructions.\n");
     }
+
+    int numBusyFUs = fuPool->numBusyFUs();
+
+    if (numBusyFUs == 0)
+      iqStats.numInstsExec0++;
```
dhschall: wrong indentation

dhschall (changes requested):
First of all, thank you for adding this useful feature.

For now, just two very high-level comments:
1. Add more description in the PR/ commit message and code. With a link to the paper you used to implement.
2. Please correct the formatting. Check the code committed from the automatic hook. Also fix the indentation. Look at this PR https://github.com/gem5/gem5/pull/2313 which will check correct formatting. You can use the `.clang-format` file to auto-format your commit.
AFAIK you just need to download the file and add it to gem5 and then run `git clang-format --staged --style=file`

src/cpu/o3/cpu.cc
```diff
@@ -359,6 +360,197 @@ CPU::CPUStats::CPUStats(CPU *cpu)
         .prereq(quiesceCycles);
 }
 
+CPU::CPUStats::TopDownStats::TopDownStats(CPU *cpu)
```
dhschall: You can add some description here, including a link to the original paper.

src/cpu/o3/cpu.cc
```diff
@@ -359,6 +360,197 @@ CPU::CPUStats::CPUStats(CPU *cpu)
...
+    CPU *cpu)
+    : statistics::Group(cpu, "TopDownL2_BadSpeculation"),
+      ADD_STAT(branchMissPredicts,
+               statistics::units::Rate<statistics::units::Count,
+                                       statistics::units::Count>::get(),
+               "Branch Miss Predicts"),
+      ADD_STAT(machineClears,
+               statistics::units::Rate<statistics::units::Count,
+                                       statistics::units::Count>::get(),
+               "Memory Order Violations") {
```
dhschall: Start currly braces in new line

src/cpu/o3/cpu.cc
```diff
@@ -359,6 +360,197 @@ CPU::CPUStats::CPUStats(CPU *cpu)
...
+    // L1 Bad Speculation
+    // Recovery cycles for mispredictions detected at Decode
+    int recoveryCycleToDecode = cpu->decode.getFetchToDecodeDelay();
+
+    auto decodeBranchMispred =
+        (int)recoveryCycleToDecode * cpu->decode.getStats().branchMispred;
+
+    // Recovery cycles for mispredictions detected at IEW
+    int recoveryCycleToIEW = cpu->decode.getFetchToDecodeDelay() +
+                            cpu->rename.getDecodeToRenameDelay() +
```
dhschall: Indentation

dhschall: Indentation

src/cpu/o3/cpu.cc
```diff
@@ -359,6 +360,197 @@ CPU::CPUStats::CPUStats(CPU *cpu)
...
+               "Fetch Latency Bound, frontend stalls due to instruction cache "
+               "inefficiency"),
+      ADD_STAT(fetchBandwidth,
+               statistics::units::Rate<statistics::units::Count,
+                                       statistics::units::Count>::get(),
+               "Fetch Bandwidth Bound, frontend stalls due to decoder "
+               "inefficiency") {
+    // Frontend L2
+    fetchLatency =
+        cpu->fetch.getStats().fetchBubblesMax / (cpu->baseStats.numCycles);
```
dhschall: Braces not needed

giactra:
Thanks @cosmiclat05 for this contribution! I know this has been pushed in draft mode so I will hold most of my comments for the time being until you mark this as ready for review.

An initial comment I have though is that it would be really nice if you could push the TDA struct definition into a separate file (something like src/cpu/o3/tda.{hh.cc}), so that we avoid overloading the cpu files

src/cpu/o3/lsq.cc
```diff
@@ -442,6 +442,20 @@ LSQ::sendRetryResp()
     dcachePort.sendRetryResp();
 }
 
+bool
+LSQ::anyCacheLevelMisses(int level) {
+    for (LSQUnit &unit : thread) {
+        for (auto &entry : unit.loadQueue) {
+            if (entry.valid() && entry.hasRequest()) {
+                auto req = entry.request()->mainReq();
```
dhschall: This raises an assert in situations where there is a `LSQRequest` but no request to the memory `Request`
```
gem5::o3::LSQ::LSQRequest::mainReq(): Assertion `_reqs.size() == 1' failed
```
I think it happens when an LSQ entry is allocated but requests are not yet sent. Due to a squash or something else.
Also, I'm not sure I agree with the implementation. A request might already be completed, the request valid and the depth of the request is still set. Your logic will consider it as a miss, but it's actually not.

I fixed it by checking if there are outstanding memory requests.
Note that this also returns true if the request is currently being translated.
```
bool 
LSQ::anyCacheLevelMisses(int level) {
  for (LSQUnit &unit : thread) {
    for (auto &entry : unit.loadQueue) {
      if (entry.valid() && entry.hasRequest()) {
        auto req = entry.request();
        if (req->isAnyOutstandingRequest()) {
            if (req->mainReq()->getAccessDepth() == level)
                return true;
        }
      }
    }
  }
  return false;
}

File: src/cpu/o3/commit.cc [hunk 12]
```diff
@@ -165,7 +165,9 @@ Commit::CommitStats::CommitStats(CPU *cpu, Commit *commit)
       ADD_STAT(committedInstType, statistics::units::Count::get(),
                "Class of committed instruction"),
       ADD_STAT(commitEligibleSamples, statistics::units::Cycle::get(),
-               "number cycles where commit BW limit reached")
+               "number cycles where commit BW limit reached"),
```
dhschall:
```suggestion
               "Number cycles where commit BW limit reached"),
```

src/cpu/o3/commit.cc
```diff
@@ -165,7 +165,9 @@ Commit::CommitStats::CommitStats(CPU *cpu, Commit *commit)
...
+      ADD_STAT(committedInst, statistics::units::Count::get(),
+               "Required for Top-Down, number of committed instructions")
```
dhschall:
```suggestion
               "Number of committed instructions (Required for Top-Down)")
```

src/cpu/o3/decode.hh
```diff
@@ -294,6 +294,8 @@ class Decode
      */
     bool squashAfterDelaySlot[MaxThreads];
 
+    unsigned fetchBubbles = 0;
```
dhschall: This looks like its never referenced somewhere? Can it be removed?

src/cpu/o3/fetch.cc
```diff
@@ -193,7 +193,13 @@ Fetch::FetchStatGroup::FetchStatGroup(CPU *cpu, Fetch *fetch)
              "Number of instructions fetched each cycle (Total)"),
     ADD_STAT(idleRate, statistics::units::Ratio::get(),
              "Ratio of cycles fetch was idle",
-             idleCycles / cpu->baseStats.numCycles)
+             idleCycles / cpu->baseStats.numCycles),
+    ADD_STAT(fetchBubbles, statistics::units::Count::get(),
+             "Stat for Top-Down Methodology, number of instructions not "
```
dhschall: Could be repreased like "number of empty slots where no uop was delivered to decode"

src/cpu/o3/inst_queue.hh
```diff
@@ -544,6 +544,17 @@ class InstructionQueue
         statistics::Vector fuBusy;
         /** Number of times the FU was busy per instruction issued. */
         statistics::Formula fuBusyRate;
+
+        /* Top down, cycles in which few ops are executed */
+        statistics::Scalar numInstsExec0;
+        statistics::Scalar numInstsExec1;
+        statistics::Scalar numInstsExec2;
+
+        /*Top down, MemStalls.AnyLoad*/
```
dhschall:
```suggestion
        /* Top down, MemStalls.AnyLoad */
```

dhschall (changes requested):
Thank you @cosmiclat05,
this looks now really clean. A few nits and one bug I encountered recently

dhschall:
@giactra can you also have a look now at this please

src/cpu/o3/tdm.cc
```diff
@@ -0,0 +1,257 @@
...
+                                       statistics::units::Count>::get(),
+               "Memory Bound, backend stalls due to memory subsystem"),
+      ADD_STAT(coreBound,
+               statistics::units::Rate<statistics::units::Count,
+                                       statistics::units::Count>::get(),
+               "Core Bound, backend stalls due to functional unit constraints")
+{
+    // Backend L2
+    executionStalls = (iew->instQueue.getStats().numInstsExec0 -
+                       rename->getStats().idleCycles +
```
dhschall:
```suggestion
                       rename->getStats().status[Rename::ThreadStatus::Idle]
```

Due to the new changes in the rename stats this will not work anymore.
https://github.com/gem5/gem5/commit/0525337c970f2c0cc2e0b1d6a2e4d12a475da868
@giactra maybe you can confirm this but I solved it in my code with this.

You also need to remove the two `const` from the getStats() function in rename
```

```

src/cpu/o3/tdm.cc
```diff
@@ -0,0 +1,257 @@
...
+                       iew->instQueue.getStats().numInstsExec1 +
+                       iew->instQueue.getStats().numInstsExec2) /
+                      (cpu->baseStats.numCycles);
+    auto memoryBoundRaw = (iew->instQueue.getStats().loadStallCycles +
+                           rename->getStats().storeStalls) /
```
dhschall: Same here

src/cpu/o3/tdm.cc
giactra: Apologies I forgot my change would affect this PR which has higher priority. Yes that is the right solution.

giactra:
Hi @cosmiclat05, next week I will spend some time providing a proper review to the entire PR. In the meantime, can I ask you to rebase your PR to fix existing conflicts coding style issues (this can be done automatically with git-clang-format which we now support!)

src/cpu/o3/rename.hh
```diff
@@ -540,7 +540,16 @@ class Rename
         statistics::Scalar intReturned;
         /** Number of registers freed and written back to floating point free list*/
         statistics::Scalar fpReturned;
+        /** Top Down, IEW stall while there is an in flight load */
+        statistics::Scalar storeStalls;
     } stats;
+
+  public:
+    const RenameStats &getStats() const { return stats; }
```
dhschall: I think the `const` here breaks the compilation. I don't know if its possible to access the vector stats with const.
@giactra do you know that?

src/cpu/o3/rename.hh
```diff
@@ -533,7 +533,28 @@ class Rename
         statistics::Scalar intReturned;
         /** Number of registers freed and written back to floating point free list*/
         statistics::Scalar fpReturned;
+        /** Top Down, IEW stall while there is an in flight load */
+        statistics::Scalar storeStalls;
     } stats;
+
+  public:
+    const RenameStats &
```
dhschall:
```suggestion
    RenameStats &
    getStats()
```

Removing the two consts will fix the compilation failure

src/cpu/o3/tdm.cc
```diff
@@ -0,0 +1,260 @@
...
+          statistics::units::Rate<statistics::units::Count,
+                                  statistics::units::Count>::get(),
+          "Retiring, fraction of slots successfully retired by the backend")
+{
+    // Total Slots
+    statistics::Temp totalSlots =
+        rename->getWidth() * cpu->baseStats.numCycles;
+
+    // L1 Frontend Bound
+    frontendBound = fetch->getStats().fetchBubbles / (totalSlots);
```
dhschall: Tbh. I don't think this is correct. Why do you get the fetch bubbles from the front-end? Shouldn't you measure the fetchBubbles at the dispatch point (rename)? There should be a single point in the pipeline where you measure.

I ask this also because I get negative values for `backendBound`. This happens in particular when I increase the delays between the front-end stages (Decode, Rename,...). So either badSpeculations or frontendBound is counted wrong, such that the fraction is together more than 1, resulting in the negative backend fraction.
I believe actually both are calculated wrong. 94% badSpeculation sounds wrong:
image

src/cpu/o3/tdm.cc
```diff
@@ -0,0 +1,260 @@
...
+
+    auto iewBadSpec =
+        (int)recoveryCycleToIEW * (iew->getStats().branchMispredicts +
+                                   iew->getStats().memOrderViolationEvents);
+
+    // Number of wasted slots due to bad speculation
+    auto wastedSlots =
+        rename->getStats().renamedInsts - commit->getStats().committedInst;
+
+    badSpeculation = (wastedSlots + (decodeBranchMispred + iewBadSpec) *
```
dhschall: I don't think `decodeBranchMispred` count to badSpeculation. These are BTB misses and are part of the frontend-bound fraction.

src/cpu/o3/tdm.cc
```diff
@@ -0,0 +1,260 @@
...
+               " constraints."),
+      ADD_STAT(
+          retiring,
+          statistics::units::Rate<statistics::units::Count,
+                                  statistics::units::Count>::get(),
+          "Retiring, fraction of slots successfully retired by the backend")
+{
+    // Total Slots
+    statistics::Temp totalSlots =
+        rename->getWidth() * cpu->baseStats.numCycles;
```
dhschall: Using `numCycles` will also include `idleCycles`. I don't think we should include them.

dhschall:
So I think there are a few things that we need to discuss/fix before we have a fully working metric.

1. Idle cycles with no activity: If there is no activity in any of the stages, the core will go into idle and nothing is counted. This I found is a problem for a long-latency I-cache miss. The core will eventually go idle, and it will not continue counting fetch bubbles and so the fetchBound fraction becomes incorrect. This can be quite significant if the idle cycles are quite a lot. I assume the same can happen for the backend on a long-latency data cache miss.
2. Currently, the badSpeculation fraction adds for every misprediction a constant refill penalty (pipeline length * pipeline width). However, it can happen that two mispredictions overlap. Adding a constant penalty will overestimate the badSpeculation fraction in that case

I'll think about how we can address those and the other issues mentioned in the inline comments

powerjg:
> Idle cycles with no activity: If there is no activity in any of the stages, the core will go into idle and nothing is counted. This I found is a problem for a long-latency I-cache miss. The core will eventually go idle, and it will not continue counting fetch bubbles and so the fetchBound fraction becomes incorrect. This can be quite significant if the idle cycles are quite a lot. I assume the same can happen for the backend on a long-latency data cache miss.

This should be able to be addressed by tracking how long we idle and adding that to the metric (instead of doing `idle++` do `idle += now - last_active_time`).

src/cpu/o3/tdm.cc
```diff
@@ -0,0 +1,260 @@
...
+    : statistics::Group(cpu, "TopDownL1"),
+      ADD_STAT(frontendBound,
+               statistics::units::Rate<statistics::units::Count,
+                                       statistics::units::Count>::get(),
+               "Frontend Bound, fraction of slots lost due to frontend "
+               "undersupplying the backend"),
+      ADD_STAT(badSpeculation,
+               statistics::units::Rate<statistics::units::Count,
+                                       statistics::units::Count>::get(),
+               "Bad Speculation, fraction of slots lost due to mispeculation"),
```
heshpdx, line 73: I recommend changing this from "Bad Speculation" to just "Speculation". 

Not all wrong-path execution is bad; many times useful lines are brought into the cache, as is the case with 505.mcf which has a terrible branch prediction rate but doesn't suffer from a front-end bottleneck because the wrong-path fetched data is used soon after on the correct-path as the tree traversal progresses in the program. I've had discussions with Ahmad about this; he agreed but alas the "cat is out of the bag" with the taxonomy given by the paper.

erin-le:
@osmanyasar05: Hello, could you give an update on the status of this PR on your end and respond to the comments above? Thank you!

erin-le:
@osmanyasar05: Hello, just another reminder to respond to the comments on this PR. Also, it looks like some conflicts have arisen with this PR. Could you resolve them when you get a chance? Thank you!

src/cpu/o3/tdm.cc
heshpdx: At Ampere, we use the term "Lost" instead of "Bad Speculation".

https://github.com/AmpereComputing/ampere-pmu-profiler

PR #2394, opened 2025-06-20
arch-vega: Implement CDNA4 (MI350) instructions, part 1

abmerop:
- Add SDWA to a few instructions and fix one MI300 instruction.
- Add new BF16 DOT product instructions.
- Add packed BF16 atomics.
- Implement load VRAM-to-LDS.
- Implement LDS transpose read instructions.
- Add misc. AHSR_PK and BITOP3 instructions.

This is about 1/4th of new instructions described here: https://www.amd.com/content/dam/amd/en/documents/instinct-tech-docs/instruction-set-architectures/amd-instinct-cdna4-instruction-set-architecture.pdf

### DIFF
--- a/src/arch/amdgpu/vega/gpu_decoder.cc
+++ b/src/arch/amdgpu/vega/gpu_decoder.cc
@@ -3626,7 +3626,7 @@ namespace VegaISA
         &Decoder::decode_invalid,
         &Decoder::decode_invalid,
         &Decoder::decode_invalid,
-        &Decoder::decode_invalid,
+        &Decoder::decode_OP_VOP3P__V_DOT2_F32_BF16,
         &Decoder::decode_invalid,
         &Decoder::decode_invalid,
         &Decoder::decode_invalid,
@@ -13178,6 +13178,12 @@ namespace VegaISA
     Decoder::decode_OP_VOP3P__V_DOT2_F32_F16(MachInst iFmt)
     {
         return new Inst_VOP3P__V_DOT2_F32_F16(&iFmt->iFmt_VOP3P);
+    }
+
+    GPUStaticInst*
+    Decoder::decode_OP_VOP3P__V_DOT2_F32_BF16(MachInst iFmt)
+    {
+        return new Inst_VOP3P__V_DOT2_F32_BF16(&iFmt->iFmt_VOP3P);
     }
 
     GPUStaticInst*
--- a/src/arch/amdgpu/vega/gpu_decoder.cc
+++ b/src/arch/amdgpu/vega/gpu_decoder.cc
@@ -1703,7 +1703,7 @@ namespace VegaISA
         &Decoder::decode_OP_FLAT__FLAT_ATOMIC_ADD_F64,
         &Decoder::decode_OP_FLAT__FLAT_ATOMIC_MIN_F64,
         &Decoder::decode_OP_FLAT__FLAT_ATOMIC_MAX_F64,
-        &Decoder::decode_invalid,
+        &Decoder::decode_OP_FLAT__FLAT_ATOMIC_PK_ADD_BF16,
         &Decoder::decode_invalid,
         &Decoder::decode_invalid,
         &Decoder::decode_invalid,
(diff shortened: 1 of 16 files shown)


src/arch/amdgpu/vega/gpu_decoder.cc
```diff
@@ -2115,7 +2115,7 @@ namespace VegaISA
         &Decoder::decode_invalid,
```
mattsinc, line 2115: Not relevant to this patch, but one of these days we should add a comment for each invalid one so that it's easy for us to ID which one is the "right" one to add

src/arch/amdgpu/vega/insts/ds.cc
```diff
@@ -4925,5 +4925,492 @@ namespace VegaISA
...
+        : Inst_DS(iFmt, "ds_read_b64_tr_b4")
+    {
+        setFlag(Load);
+    } // Inst_DS__DS_READ_B64_TR_B4
+
+    Inst_DS__DS_READ_B64_TR_B4::~Inst_DS__DS_READ_B64_TR_B4()
+    {
+    } // ~Inst_DS__DS_READ_B64_TR_B4
+
+    void
```
mattsinc, line 4940: previously these instructions had short descriptions of what they are doing above.  Is it possible to bring that back?

src/arch/amdgpu/vega/insts/flat.cc
```diff
@@ -988,6 +988,246 @@ namespace VegaISA
...
+        Wavefront *wf = gpuDynInst->wavefront();
+
+        if (gpuDynInst->exec_mask.none()) {
+            wf->decVMemInstsIssued();
+            return;
+        }
+
+        gpuDynInst->execUnitId = wf->execUnitId;
+        gpuDynInst->latency.init(gpuDynInst->computeUnit());
+        gpuDynInst->latency.set(gpuDynInst->computeUnit()->clockPeriod());
```
mattsinc, line 1068: is this the latency we've been using for other LDS accesses?  Seems low...

src/arch/amdgpu/vega/insts/flat.cc
```diff
@@ -2227,5 +2467,133 @@ namespace VegaISA
...
+    // --- Inst_FLAT__FLAT_ATOMIC_PK_ADD_BF16 class methods ---
+
+    Inst_FLAT__FLAT_ATOMIC_PK_ADD_BF16::Inst_FLAT__FLAT_ATOMIC_PK_ADD_BF16(
+        InFmt_FLAT *iFmt)
+        : Inst_FLAT(iFmt, "flat_atomic_pk_add_bf16")
+    {
+        setFlag(AtomicPkAddBF16);
+
+        // MI300 spec: "Float atomics must set SC[0]=0 (no return value)."
+        panic_if(instData.GLC, "Saw float atomic with return set!");
```
mattsinc, line 2479: is there a way for us to transform GLC into SC[0]?  I guess not with backwards support?

mattsinc (approved):

src/arch/amdgpu/vega/gpu_decoder.cc
abmerop: Yea, I am not sure how to do that right now without adding stubs for everything missing.  Right now it prints out what the invalid opcode is.  That's good enough if you have access to a nice tool to disassemble it for you but otherwise requires some manual effort to figure out the instruction.

src/arch/amdgpu/vega/insts/ds.cc
abmerop: Those were auto generated comments from some mystery file no one seems to be able to find.  For instructions that already had them previously I have not been removing them.  For the new instructions I added some very detailed comments in the relevant method (completeAcc in this case).  I could add something along the lines of "see comment in completeAcc" though.

src/arch/amdgpu/vega/insts/flat.cc
abmerop: For VGPR -> LDS (ds.cc instructions) we use a fixed, hard-coded latency of 24 (which is another issue).  These instructions are from VRAM -> LDS and happen "in the background."  The latency specified here is just the instruction execute latency (address calculation).

src/arch/amdgpu/vega/insts/flat.cc
abmerop: We could make that bit a union such that you would be able to specify `instData.GLC` or `instData.SC0`.  For instructions pre-dating MI300 / MI350 it probably makes sense to remain GLC though.

src/arch/amdgpu/vega/insts/ds.cc
mattsinc: gotcha

src/arch/amdgpu/vega/insts/flat.cc
mattsinc: another problem to fix some day :)

PR #2396, opened 2025-06-23
configs: Fix bootloader argument parsing in legacy RISC-V script

### LINKED ISSUE #2354: fs_linux.py bootloader option does not work
**Describe the bug**

#2231 and  #2256  change the way the bootloader argument works.
I encountered the following due to this change:

Since bootloader in Options.py is now a list (instead of a string) I am getting this behaviour with the fs_linux.py script.

gem5 command:
`./build/RISCV/gem5.opt configs/example/riscv/fs_linux.py  --bootloader=/home/nick/CAL/GEM5_RESOURCES/riscv/bootloaders/fw_jump.elf  --kernel=/home/nick/CAL/GEM5_RESOURCES/riscv/kernels/linux6.7-vmlinux  --disk=/home/nick/CAL/GEM5_RESOURCES/riscv/disks/riscv-disk.img`

Output:
```
...
src/base/loader/image_file_data.cc:105: fatal: fatal condition fd < 0 occurred: Failed to open file ['/home/nick/GEM5_RESOURCES/riscv/bootloaders/fw_jump.elf'].
This error typically occurs when the file path specified is incorrect.
Memory Usage: 170592 KBytes
```

You can see that the filename variable in C++ code has now taken the value of the string representation of the Python list which of course will never be the intended path.

Am I misusing something here?

**Affects version**
v24.1.0.3 

**gem5 Modifications**
No modifications.

**To Reproduce**

1. Compile gem5 with command: `scons build/RISCV/gem5.opt -j <NUMCORES> --linker=gold`
2. Execute the simulation with: `./build/RISCV/gem5.opt configs/example/riscv/fs_linux.py  --bootloader=<your_bootloader>  --kernel=<your_kernel>  --disk=<your_disk>`

**Expected behavior**
Crash due to bad parsing of the bootloader argument.

**Host Operating System**
Distributor ID:	Ubuntu
Description:	Ubuntu 24.04.2 LTS
Release:	24.04
Codename:	noble

**Host ISA**
X86

**Compiler used**
g++ (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0

tommarin:
Bootloader argument is provided in form of list for ARM simulations, but a single string is expected by RISC-V bootloader_filename.

Fixes issue #2354

### DIFF
--- a/configs/example/riscv/fs_linux.py
+++ b/configs/example/riscv/fs_linux.py
@@ -205,8 +205,9 @@ elif not args.bootloader:
     system.workload = RiscvLinux(**workload_args)
     system.workload.object_file = args.kernel
 else:
+    assert len(args.bootloader) == 1, "Please specify only one bootloader"
     system.workload = RiscvBootloaderKernelWorkload(**workload_args)
-    system.workload.bootloader_filename = args.bootloader
+    system.workload.bootloader_filename = args.bootloader[0]
     system.workload.object_file = args.kernel
 
 system.iobus = IOXBar()


tommarin:
Adding a small note here: I understand that `Options.py` and the legacy example scripts will be deprecated soon. However, I believe it is good to have these scripts still working before their support is fully dropped.

configs/example/riscv/fs_linux.py
```diff
@@ -206,7 +206,7 @@ def generateDtb(system):
     system.workload.object_file = args.kernel
 else:
     system.workload = RiscvBootloaderKernelWorkload(**workload_args)
-    system.workload.bootloader_filename = args.bootloader
+    system.workload.bootloader_filename = args.bootloader[0]
```
powerjg, line 210: Can you add an assert/error message if the length is not 0?

powerjg (approved):

PR #2399, opened 2025-06-24
cpu: Fix looppoint analysis v25

### LINKED ISSUE #2395: LoopPoint Analysis probe needs to be updated for the new probe structure
**Describe the bug**
After updating the probe structure in gem5, the LoopPoint Analysis probe needs to be updated.

**Affects version**
gem5 v25 stable

**gem5 Modifications**
We need to update the way the LoopPoint Analysis probelistener is attached to the probe point.

**To Reproduce**
Will attach a script and a more detailed description of how to reproduce this soon.

studyztp:
This is a PR for the issue: https://github.com/gem5/gem5/issues/2395

The new probe structure uses ```dynamic_cast``` to check the```ProbeListener```, which caused a casting mismatch with the LoopPoint Analysis ProbeListener, because the LoopPoint Analysis ProbeListener uses ```StaticInstPtr``` instead of ```const StaticInstPtr``` as part of its Arg while the Probe Point that it is listening to, ```Commit```, uses ```const StaticInstPtr```.
This commit fixes the issue.

### DIFF
--- a/src/cpu/simple/probes/looppoint_analysis.cc
+++ b/src/cpu/simple/probes/looppoint_analysis.cc
@@ -73,7 +73,7 @@ LooppointAnalysis::updateLocalBBV(const Addr pc)
 
 void
 LooppointAnalysis::checkPc(
-    const std::pair<SimpleThread*,StaticInstPtr>& inst_pair
+    const std::pair<SimpleThread*,const StaticInstPtr>& inst_pair
 )
 {
     const StaticInstPtr &inst = inst_pair.second;
--- a/src/cpu/simple/probes/looppoint_analysis.hh
+++ b/src/cpu/simple/probes/looppoint_analysis.hh
@@ -62,7 +62,8 @@ class LooppointAnalysis : public ProbeListenerObject
      * @param inst_pair A pair that contains the SimpleThread pointer and the
      * StaticInstPtr of the commited instruction from the atomic CPU.
      */
-    void checkPc(const std::pair<SimpleThread*, StaticInstPtr>& inst_pair);
+    void checkPc(const std::pair<SimpleThread*,
+        const StaticInstPtr>& inst_pair);
 
     /**
      * When this function is called, it sets the class variable ifListening to
@@ -77,8 +78,8 @@ class LooppointAnalysis : public ProbeListenerObject
      */
     void stopListening();
 
-    typedef ProbeListenerArg<LooppointAnalysis,
-        std::pair<SimpleThread*, StaticInstPtr>> looppointAnalysisListener;
+    typedef ProbeListenerArg<LooppointAnalysis, std::pair<SimpleThread*,
+        const StaticInstPtr>> looppointAnalysisListener;
 
   private:
 


giactra:
Hi, thanks for your contribution. This PR is targeting stable branch but it should actually target the develop branch:

https://github.com/gem5/gem5/blob/stable/CONTRIBUTING.md

studyztp:
> Hi, thanks for your contribution. This PR is targeting stable branch but it should actually target the develop branch:
> 
> https://github.com/gem5/gem5/blob/stable/CONTRIBUTING.md

Thank you for pointing this out. I rebased my branch. I planned to have this as a hotfix for the stable branch, so I targeted the stable branch. Sorry about this.

test-script/test-looppoint-analysis-script.py
```diff
@@ -0,0 +1,500 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
+Usage
+-----
+
+```
+scons build/X86/gem5.opt
+./build/X86/gem5.opt \
```
powerjg: Does it have to be `X86`? Can we start making all examples use `ALL`?

test-script/test-looppoint-analysis-script.py
```diff
@@ -0,0 +1,500 @@
+# Copyright (c) 2025 The Regents of the University of California
```
powerjg: For this whole file... Can you ask copilot to put in comments? This file is going to be impossible for our future selves to be able to debug without any comments. 

You may also want to ask copilot to write the tests for you.

Finally, it may be a better idea to separate out the "tests" into their own file instead of mixing simulation details in with the test code.

powerjg (changes requested):

test-script/test-looppoint-analysis-script.py
studyztp: I'll use ALL for all the future examples.

test-script/test-looppoint-analysis-script.py
studyztp: I used copilot to put in comments, but I asked it to make only necessary comments to prevent polluting the file. I'll let it make more detailed comments for the future test. Thanks for the suggestion.
I think I should make a better test set for the LoopPoint analysis module and separate the tests from this commit. I removed this test script and left only the fix to this commit.

BobbyRBruce:
Just to clarify here: what's the status of these tests? The PR adds them, then removes them. Is the intent that I re-add them as testlib tests on this PR?

studyztp:
> Just to clarify here: what's the status of these tests? The PR adds them, then removes them. Is the intent that I re-add them as testlib tests on this PR?

I was intended to add them but now I think I should separate them completely from this PR. Now this PR only has the fix of the LoopPoint Analysis probe. Sorry for my mistake.

powerjg (approved):

PR #2400, opened 2025-06-26
arch-riscv: Fix bootloader & kernel load address

rogerchang23424:
The bootloader_addr should be expect where the bootloader address put. The bootloader_addr only place expect location only if the base address of firmware is zero. If the base address is not zero, we should subtrace the elf offset

Change-Id: I66025659d80ae2c34821d665e351f9e96f4384d6

### DIFF
--- a/src/arch/riscv/linux/fs_workload.cc
+++ b/src/arch/riscv/linux/fs_workload.cc
@@ -153,15 +153,15 @@ void
 BootloaderKernelWorkload::loadBootloader()
 {
     if (params().bootloader_filename != "") {
-        Addr bootloader_addr_offset = params().bootloader_addr;
-        bootloader->buildImage().offset(bootloader_addr_offset).write(
-            system->physProxy
-        );
+        loader::MemoryImage image = bootloader->buildImage();
+        Addr bootloader_addr_offset = \
+            params().bootloader_addr - image.minAddr();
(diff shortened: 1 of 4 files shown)


powerjg (dismissed):

src/arch/riscv/linux/fs_workload.cc
```diff
@@ -153,15 +153,14 @@ void
 BootloaderKernelWorkload::loadBootloader()
 {
     if (params().bootloader_filename != "") {
-        Addr bootloader_addr_offset = params().bootloader_addr;
-        bootloader->buildImage().offset(bootloader_addr_offset).write(
-            system->physProxy
-        );
+        loader::MemoryImage image = bootloader->buildImage();
+        Addr bootloader_addr_offset = params().bootloader_addr - image.minAddr();
```
erin-le: @rogerchang23424: Hello, sorry for the nitpick. Could you reduce the length of this line to 79 characters or less?

erin-le (changes requested):

src/arch/riscv/linux/fs_workload.cc
rogerchang23424: Done

powerjg (approved):

erin-le (approved):

PR #2403, opened 2025-06-27
misc: Update protobuf files to remove required

powerjg:
The "required" statement is considered bad practice now. So much so that proto3 removed "required" as an option. See the discussion: https://stackoverflow.com/questions/31801257/why-required-and-optional-is-removed-in-protocol-buffers-3

This change converts required to optional. This shouldn't affect any existing traces or gem5's code in any way.

### DIFF
--- a/src/proto/inst.proto
+++ b/src/proto/inst.proto
@@ -42,14 +42,14 @@ package ProtoMessage;
 // the trace, the version of this file format, and the tick frequency
 // for all the packet time stamps.
 message InstHeader {
-  required string obj_id = 1;
-  required uint32 ver = 2 [default = 0];
-  required uint64 tick_freq = 3;
-  required bool has_mem = 4;
+  optional string obj_id = 1;
+  optional uint32 ver = 2 [default = 0];
+  optional uint64 tick_freq = 3;
+  optional bool has_mem = 4;
 }
 
 message Inst {
-  required uint64 pc = 1;
+  optional uint64 pc = 1;
 
   // Either inst or inst_bytes must be used, but never both. That should be
   // enforced by the oneof keyword, but that's not supported in all versions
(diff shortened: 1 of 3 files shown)


giactra:
Hi Jason, interesting read.

It probably matters very little but I would propose to not explicitly mark everything as optional, but instead to simply remove the required specifier. So instead of:

```
message Packet {
  optional uint64 tick = 1;
  optional uint32 cmd = 2;
  optional uint64 addr = 3;
  optional uint32 size = 4;
  optional uint32 flags = 5;
  optional uint64 pkt_id = 6;
  optional uint64 pc = 7;
}
```

Prefer

```
message Packet {
  uint64 tick = 1;
  uint32 cmd = 2;
  uint64 addr = 3;
  uint32 size = 4;
  optional uint32 flags = 5;
  optional uint64 pkt_id = 6;
  optional uint64 pc = 7;
}
```

While it is the same from a protobuf perspective, it however retains some semantic information about which fields are effectively expected to be present in a gem5 Packet, even if protobuf per se does not require them.

What do you think?

powerjg:
My understanding is that you should always use optional (and that's my understanding of my requirements internally as well, which caused me to create this change).

https://protobuf.dev/programming-guides/proto2/#field-labels

My understanding is that every singular field should have a label, and that label should be "optional."

If you want to include human-understandable semantic information, I can add a comment that says something to the extent of "All packets are expected to have this field" on the fields that were required.

giactra:
> My understanding is that you should always use optional (and that's my understanding of my requirements internally as well, which caused me to create this change).
> 
> https://protobuf.dev/programming-guides/proto2/#field-labels
> 
> My understanding is that every singular field should have a label, and that label should be "optional."
> 

Understood!

> If you want to include human-understandable semantic information, I can add a comment that says something to the extent of "All packets are expected to have this field" on the fields that were required.

Yes I feel that would help!

BobbyRBruce (approved):

giactra:
Not a big problem but it would have been nice to have the PR merged once my comments were addressed...

PR #2404, opened 2025-06-27
arch,mem-ruby: Make parsers less verbose by default

powerjg:


giactra:
Hi @powerjg , how's the new build log now after this change? Could you post a before-and-after as PR message?

powerjg:
Annoyingly, it's hard to get good logs here since SCons doesn't reliably execute things in any order. Basically, this gets rid of "LALR Warning: ..." and removes some extra files that are generated (the "cached" tables that are never used).

This was something to make building easier for us. I would strongly prefer not to keep this as an internal patch. These warnings have been ignored upstream for a decade, so I can't believe they are useful.

src/arch/isa_parser/isa_parser.py
```diff
@@ -519,10 +519,13 @@ def padDestRegIdx(self, padding):
...
 
 class ISAParser(Grammar):
-    def __init__(self, output_dir, decoder_name="Decoder"):
+    def __init__(self, output_dir, decoder_name="Decoder", verbose=False):
         super().__init__()
         self.lex_kwargs["reflags"] = int(re.MULTILINE)
         self.output_dir = output_dir
+        self.verbose = verbose
+        self.yacc_kwargs["debug"] = self.verbose
+        self.yacc_kwargs["write_tables"] = False
```
giactra, line 528: Shouldn't this also be self.verbose?

giactra (commented):
Agreed I see no harm in merging this. Just wanted to make sure the PR had a bit of context for debug purposes (unlikely though it would be necessary)

src/arch/isa_parser/isa_parser.py
powerjg: I think dumping the tables in yacc is never useful, which is why I did it this way, but if you or someone else finds those generated files useful, I can change it.

giactra: Ack. One question: is this hardcoded to False as of now? If yes, I'll approve the PR. If not, it would be nice to allow consistent behaviour across releases (therefore binding it to self.verbose)

powerjg: > is this hardcoded to False as of now?

It's hardcoded `True` now.

powerjg: @giactra did the comment above answer your question?

giactra: Hi @powerjg , apologies. I thought it was implicit in my message a request to change it to verbose, for consistency with what we are doing now.

The default value for the verbose parameter is anyway False so we are anyway making parsers less verbose by default as the PR is suggesting to do. In this way we are just making sure that a user doesn't lose any debugging insight in case they decide to build with verbose.

At the end of the day it is not a big deal and I don't want to annoy you too much. I will approve this so that you can merge it if you disagree with me or force push otherwise :)

giactra (approved):

PR #2405, opened 2025-06-27
mem-ruby: Update slicc to use relative imports

powerjg:
I think that this code was from before python had good support for relative imports and packages. This PR updates slicc to use relative imports instead of starting everything from `slicc...` This will make it easier to make building slicc more flexible in the future.


giactra:
Hi @powerjg, in general absolute imports should be preferred to relative ones (https://peps.python.org/pep-0008/).

what's the bigger plan I am not seeing?

powerjg:
I want to be able to execute this script outside of SCons. In that context, we don't know how to set up the pythonpath. I disagree with absolute imports in this case because `slicc` should be a python package. 

That said, if you have a strong objection, I can find a different workaround. Do you have a specific reason that it would impact anyone?

erin-le:
@giactra: Hello, can you let us know if you're okay with Jason's changes? Thank you!

powerjg:
@giactra are you blocking this PR? I agree that absolute imports are better, but the problem is setting up the `PYTHONPATH` for these "absolute" imports to work. They are "absolute" only if `gem5/src/mem/slicc` is in the `PYTHONPATH`. Otherwise, they are broken relative imports as it stands. With this change. `slicc` is a true python package and can be used as an exported package without modifying the `PYTHONPATH`.

BobbyRBruce (dismissed):

PR #2406, opened 2025-06-27
arch: Update ISA parser to be moveable

powerjg:
This changes allows the ISA parser to run from a directory other than gem5/. This is useful if you want to run this code outside of scons.

The verbosity of x86 is also reduced in this commit. Note that this commit only affects x86 because it is the only ISA that has its own python files.

### DIFF
--- a/src/arch/isa_parser/isa_parser.py
+++ b/src/arch/isa_parser/isa_parser.py
@@ -1078,6 +1078,16 @@ del wrap
         # next split's #define from the parser and add it to the current
         # emission-in-progress.
         try:
+            # x86 requires more python files to be interpreted. To allow this
+            # script to be run from other locations for x86, force adding the
+            # x86 ISA directory to the Python path.
+            import os
+            import sys
+
+            current_file_path = os.path.realpath(__file__)
+            arch_dir = os.path.dirname(os.path.dirname(current_file_path))
+            sys.path.append(arch_dir)
+            sys.path.append(os.path.join(arch_dir, "x86/isa/"))
             exec(split_setup + fixPythonIndentation(t[2]), self.exportContext)
         except Exception as exc:
             traceback.print_exc(file=sys.stdout)
--- a/src/arch/micro_asm.py
+++ b/src/arch/micro_asm.py
@@ -575,7 +575,7 @@ def p_error(t):
 class MicroAssembler:
     def __init__(self, macro_type, microops, rom=None, rom_macroop_type=None):
         self.lexer = lex.lex()
-        self.parser = yacc.yacc(write_tables=False)
+        self.parser = yacc.yacc(write_tables=False, debug=False)
         self.parser.macro_type = macro_type
         self.parser.macroops = {}
         self.parser.microops = microops


giactra:
Hi Jason, this I believe is part of a series of PR which looks like it is aiming at executing some gem5 python code from a different repo. Would you mind providing more context? For instance documenting the final goal of this refactoring with a design document or a github discussion/issue? Thanks

powerjg:
Yeah, as part of my work, I need to be able to execute the `isa_parser` from outside of SCons. The final goal is simply making it possible to run this script without using SCons. Nothing else is going to change here.

To be honest, this code is really brittle and we've been lucky that it has worked so far. Making the code less brittle is not on my to do list in the near term.

src/arch/isa_parser/isa_parser.py
```diff
@@ -1078,6 +1078,16 @@ def split(sec):
...
+            # x86 requires more python files to be interpreted. To allow this
+            # script to be run from other locations for x86, force adding the
+            # x86 ISA directory to the Python path.
+            import os
+            import sys
+
+            current_file_path = os.path.realpath(__file__)
+            arch_dir = os.path.dirname(os.path.dirname(current_file_path))
+            sys.path.append(arch_dir)
+            sys.path.append(os.path.join(arch_dir, "x86/isa/"))
```
giactra, line 1090: Hi Jason, I don't particularly understand how split is used in the ISA parser, but I can see it being referred by Arm ISA files as well...

powerjg: I don't understand the question.

The code highlighted is x86-specific. The reason it's needed is the `exec(split_setup + fixPythonIndentation(t[2]), self.exportContext)` executes this code: `from insts import microcode` (https://github.com/gem5/gem5/blob/stable/src/arch/x86/isa/microasm.isa#L53).

So, to `exec` the `let` block in microasm.isa, we need to ensure that the pythonpath includes the `isa` directory.

*I do not like this code, and I think that all cases of `exec` should be purged from the codebase.* However, I tried a couple of other quick fixes, but they didn't work.

I see no reason the code highlighted won't work for Arm because I have tested building x86, Arm, and RISC-V outside of scons with this code in a fully hermetic environment (e.g., bazel).

BobbyRBruce (approved):

src/arch/isa_parser/isa_parser.py
giactra: Thanks for the explanation.

I feel the isa parser code is highly unreadable and convoluted (mainly due to the execs) as you correctly suggested.

I fear this will make it even worse. Adding x86/isa/ path for every ISA build is weird and could confuse users or potentially lead to horrible bugs (what happens if the same module is defined in both x86/isa and arm/isa ?)

I believe the issue should be handled on the ISA side. In other words, all "__init__.py" files like https://github.com/gem5/gem5/blob/stable/src/arch/x86/isa/insts/__init__.py and below should be purged of execs and have direct imports.

I don't know if this will solve the issue but it will make things easier to debug what is going wrong.

Alternatively you could consider adding the sys.path to the microasm.isa file and constrain the workaround to x86

giactra (changes requested):

PR #2408, opened 2025-06-27
misc: Add gitignore for ext/sst

powerjg:


giactra (approved):

giactra:
Naive question: shouldn't we copy the gitignore from the SST project and add gem5 specific files on top of it?

powerjg:
> Naive question: shouldn't we copy the gitignore from the SST project and add gem5 specific files on top of it?

I don't think so. The `ext/sst` directory just has our SST component, not the whole SST git repo.

PR #2409, opened 2025-06-27
scons: Update build_tools to enable importing

powerjg:
This updates the scripts in build_tools to use `if __name__=="__main__"`. It factors out the main functions to allow them to be used from other scripts. The purpose is to enable code sharing in the case we're not using the importer.

### DIFF
--- a/build_tools/cxx_config_cc.py
+++ b/build_tools/cxx_config_cc.py
@@ -41,48 +41,57 @@ import argparse
 import importlib
 import os.path
 import sys
+from typing import Type
 
-import importer
 from code_formatter import code_formatter
 
-parser = argparse.ArgumentParser()
-parser.add_argument("modpath", help="module the simobject belongs to")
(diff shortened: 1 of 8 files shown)


powerjg:
For reviewers: The only changes should be to indentation in these files. It's not super easy to see on the github website, but vscode makes it easier to see.

giactra (approved):

erin-le:
@powerjg: Just FYI, there are a few lines in this change that are over 80 characters. These files also had lines longer than 80 characters before your changes, however, so I'm not sure if the 80 character requirement applies here.

powerjg:
> @powerjg: Just FYI, there are a few lines in this change that are over 80 characters. These files also had lines longer than 80 characters before your changes, however, so I'm not sure if the 80 character requirement applies here.

I guess it's just comments or long quotes? IMO, if black doesn't catch it, lets not worry about it. However, if someone strongly disagrees I can manually fix it...

erin-le:
I'll go ahead and squash merge this PR then

PR #2413, opened 2025-07-01
tests: Add an IPC-based Regression Test

### LINKED ISSUE #702: Add automatic regression test to check for performance degradations
### Discussed in https://github.com/orgs/gem5/discussions/446

<div type='discussions-op-text'>

<sup>Originally posted by **dhschall** October 12, 2023</sup>
After today's developers meeting discussion, I was wondering if there is a simple way to add some sort of regression test to the existing testing infrastructure. Those are orthogonal to the normal unit tests and will report some common statistics like IPC, Branch MPKI, I-cache or D-cache MPKI, and simulated instructions per second (KIPS/MIPS).

In addition, the test can issue a warning if one of the statistics exceeds a certain range (for simulation time in particular).

I only thought about a few benchmarks for the most common ISA's and CPU cores. (I.e. x86, Arm, RISC-V, O3, atomic, timing)
 
This would have three main advantages.
1. For developers to quickly encounter performance degradations. Either simulated performance or simulation speed
2. Would help reviewers to quickly see if the change alters behavior or degrades performance.
3. For people new to gem5 it would give them a reference to see spot misconfigurations. (There are too many parameters you can mess with)

Also, it would be great to see those metrics evolve over time (in addition to stars and clones). It might help to speed up the simulation time of gem5 and improve the models.

</div>

erin-le:
This PR adds a regression test that runs an X86 kernel boot and checks the IPC against the IPC of a previous run. It also caches the `stats.txt` of the current run so it can be used for comparison in later runs of the test. The test is currently in its own yaml file for testing, but should be incorporated into the daily tests yaml once testing is complete.

### DIFF
--- a/tests/gem5/regression_tests/configs/kernel-boot.py
+++ b/tests/gem5/regression_tests/configs/kernel-boot.py
@@ -84,8 +84,8 @@ memory = SingleChannelDDR3_1600(size="3GiB")
 #     num_cores=1,
 # )
 processor = SimpleProcessor(
-    # cpu_type=CPUTypes.KVM,  # switch to ATOMIC for final
-    cpu_type=CPUTypes.ATOMIC,
+    cpu_type=CPUTypes.KVM,  # switch to ATOMIC for final
+    # cpu_type=CPUTypes.ATOMIC,
     isa=ISA.X86,
     num_cores=1,
 )
@@ -128,7 +128,10 @@ try:
     if (new_ipc / old_ipc) < 0.8:
         print(f"New IPC is less than 80% of the old IPC!")
         exit(1)
-
+except FileNotFoundError:
+    print("One of the stats files was not found!")
+except KeyError:
+    print("One of the stats files might be empty!")
 finally:
     shutil.move(
         f"{m5.options.outdir}/stats.txt",
--- a/tests/gem5/regression_tests/configs/kernel-boot.py
+++ b/tests/gem5/regression_tests/configs/kernel-boot.py
@@ -1,0 +1,133 @@
+# Copyright (c) 2025 The Regents of the University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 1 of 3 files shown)


tests/gem5/regression_tests/configs/kernel-boot.py
```diff
@@ -0,0 +1,139 @@
...
+        exit(1)
+except FileNotFoundError:
+    print("One of the stats files was not found!")
+except KeyError:
+    print("One of the stats files might be empty!")
+finally:
+    shutil.move(
+        f"{m5.options.outdir}/stats.txt",
+        "./gem5/regression_tests/cached_stats.txt",
+    )
```
erin-le, line 139: Right now, the new `stats.txt` file replaces the cached `stats.txt` even if the new IPC is less than the threshold. Should it be changed so that the cached `stats.txt` is only replaced if the new IPC isn't significantly smaller?

tests/gem5/regression_tests/configs/kernel-boot.py
```diff
@@ -0,0 +1,139 @@
...
+    read_stats_files("./gem5/regression_tests/cached_stats.txt", cached_stats)
+    read_stats_files(f"{m5.options.outdir}/stats.txt", curr_stats)
+
+    old_ipc = float(cached_stats["board.processor.cores.core.ipc"])
+    new_ipc = float(curr_stats["board.processor.cores.core.ipc"])
+
+    print(f"old IPC: {old_ipc}")
+    print(f"new IPC: {new_ipc}")
+
+    if (new_ipc / old_ipc) < 0.8:
```
erin-le, line 128: I chose this arbitrarily, so it can be replaced with something else if necessary.

powerjg: I'd choose like 2%.

giactra:
Hi @erin-le, thanks for this change. I have expressed multiple times the wish to include regression testing to our CI.

I have some ideas and I would like to find some time to them put on "paper" at a certain point (always too busy, sigh :( )
Even if I still need to form a definite idea in my head, I'd like to throw the following suggestions (related to the PR). I don't claim I have thought through them enough so feel free to share your opinion

### We should really extract stats from the python world, at the end of a simulation, and avoid relying on cached stats files as golden reference:

In this way:
* We don't have to deal with the stat file format specification (which is humand readable, but not machine friendly when it comes to parsing)
* It gives a user a better visibility over the golden reference value

There are probably multiple ways for reading stats in python. IIRC Bobby worked at something similar some time ago (I remember a presentation during a developer's meeting). In the past I extracted stats from a SimObject via the getStats pybind11 generated method: https://github.com/gem5/gem5/blob/stable/src/python/pybind11/stats.cc#L249

So for instance the IPC check could be rewritten as (pseudocode):

```
ipc_reference = 3.0
cpu_stats = system.cpu.getStats()
if cpu_stats["ipc"] < ipc_reference:
    raise(.....)

```

So a legal change in IPC would require a PR to amend the ipc_reference directly in python

### Rather than adding yet another kernel boot test, we should consider implementing stats checking verifiers
Instead of booting the same kernel multiple times, from separate config files, simply because we are interested on a different metric (e.g. IPC regression), we should consider using the existing testlib [verifier framework](https://github.com/gem5/gem5/blob/stable/tests/gem5/verifier.py).

A verifier gives you the condition upon which a test is considered as passing. Multilple verifiers can be used in a single test. We should consider adding IPC verifiers as a specific case of stat checking verifier, and instantiated those verifiers on existing kernel boot tests

Please let me know which are your thoughts on this

Giacomo

erin-le:
@giactra: Hello, sorry that it's taken a while to get back to you on this. I discussed your comments with Jason and Bobby and they agreed that we should use stats from Python instead of extracting them from `stats.txt`, but they also said that it would currently be too large of an undertaking to add verifiers to every test. For now, I've opened another draft PR that gets stats from Python instead of `stats.txt`: https://github.com/gem5/gem5/pull/2524

erin-le:
Closing as another PR has been opened.

PR #2415, opened 2025-07-01
stdlib: remove duplicate ClassicGeneratorExitHandler

erin-le:
Previously, there were two definitions of the ClassicGeneratorExitHandler, which handles hypercall 0, in `exit_handler.py`. This commit removes the definition located earlier in the file, as it isn't used.


powerjg (approved):
Thanks, I think this was my rebase error.

PR #2416, opened 2025-07-01
dev: Export reg bank's offsetMap

ylldummy:
This can be useful in some cases. For example, we have a bank of double buffered registers and on some condition, we need to iterate through all registers in the bank to sync the buffered state to active state.

Instead of maintaining another list of register references, having this field exported would help simplify the implementation.

### DIFF
--- a/src/dev/reg_bank.hh
+++ b/src/dev/reg_bank.hh
@@ -1026,6 +1026,8 @@ class RegisterBank : public RegisterBankBase
     Addr base() const { return _base; }
     Addr size() const { return _size; }
     const std::string &name() const { return _name; }
+    const std::map<Addr, std::reference_wrapper<RegisterBase>> &
+    offsetMap() const { return _offsetMap; }
 
     virtual void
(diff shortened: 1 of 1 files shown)


src/dev/reg_bank.hh
```diff
@@ -1026,6 +1026,7 @@ class RegisterBank : public RegisterBankBase
     Addr base() const { return _base; }
     Addr size() const { return _size; }
     const std::string &name() const { return _name; }
+    const auto &offsetMap() const { return _offsetMap; }
```
powerjg: Can you put the type here explicitly? I think that it harms readability to use `auto` in header files.

powerjg (commented):
One small thing.

src/dev/reg_bank.hh
ylldummy: Sure. Thanks for the suggestion. Updated.

powerjg (approved):

PR #2417, opened 2025-07-01
arch-riscv: Make PLIC output lantency configurable

rogerchang23424:
Change-Id: Ia935a89c3f5fe82b1ae949c93719555bbdbb5ed0

### DIFF
--- a/src/dev/riscv/Plic.py
+++ b/src/dev/riscv/Plic.py
@@ -82,6 +82,10 @@ class Plic(PlicBase):
         "Number of interrupt contexts. Usually the number "
         "of threads * 2. One for M mode, one for S mode",
     )
+    output_latency = Param.Int(
+        3,
+        "Cycle latency from received interrupt signal to output",
+    )
 
     def generateDeviceTree(self, state):
         node = self.generateBasicPioDeviceNode(
--- a/src/dev/riscv/plic.cc
+++ b/src/dev/riscv/plic.cc
@@ -57,6 +57,7 @@ Plic::Plic(const Params &params) :
     PlicBase(params),
     system(params.system),
     nSrc(params.n_src),
+    outputLatency(params.output_latency),
     registers(params.name, pioAddr, this),
     update([this]{updateOutput();}, name() + ".update")
 {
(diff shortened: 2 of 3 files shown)


src/dev/riscv/Plic.py
```diff
@@ -82,6 +82,10 @@ class Plic(PlicBase):
         "Number of interrupt contexts. Usually the number "
         "of threads * 2. One for M mode, one for S mode",
     )
+    output_latency = Param.Int(
+        3,
+        "Cycle lantency from received interrupt signal to output",
```
powerjg:
```suggestion
        "Cycle latency from received interrupt signal to output",
```

src/dev/riscv/plic.cc
```diff
@@ -415,7 +416,7 @@ Plic::propagateOutput()
     }
 
     // Add new output to outputQueue
-    Tick next_update = curTick() + cyclesToTicks(Cycles(3));
+    Tick next_update = curTick() + cyclesToTicks(Cycles(outputLatency));
```
powerjg: Could this break things if `outputLatency` is 0? If so, we should either have a fatal in the constructor.

I'd also suggest the following to make sure that you're getting a clock edge. The current code could return a tick that is not on a clock edge if curTick() is not on a clock edge.

```suggestion
    Tick next_update = clockEdge(Cycles(outputLatency));
```

src/dev/riscv/Plic.py
rogerchang23424: Done

src/dev/riscv/plic.cc
rogerchang23424: > Could this break things if `outputLatency` is 0? If so, we should either have a fatal in the constructor.

No, we can schedule the intterupt output greater eqaul than current tick
https://github.com/gem5/gem5/blob/f93c153f2c2d7d5b3266054fdf421749e9589783/src/sim/eventq.hh#L757-L760

powerjg: Sure, it makes sense that it's OK to do on the event queue. I was more worried about the PLIC logic of having two interrupts in the same tick. This could cause problems in other logic. Or, it could be no problem at all. I'll resolve this issue.

powerjg (approved):

PR #2418, opened 2025-07-01
arch-vega: MI350 microscaling updates

abmerop:
- Adds new data types (mxfp4, mxfp6 e3m2 and e2m3) and unit tests.
- Fixes several issues with denormal aka subnormal FP conversions.
- Helper methods for upcoming instructions.
- Tested by converting to F32 and back and ensure same bit values.

### DIFF
--- a/src/arch/amdgpu/common/dtype/mxfp.hh
+++ b/src/arch/amdgpu/common/dtype/mxfp.hh
@@ -145,6 +145,11 @@ class mxfp
         FMT in = getFmt();
         int exp = in.exp;
 
+        // Our value is zero, scaling by anything remains zero.
+        if (exp == 0 && in.mant == 0) {
+            return;
+        }
+
         if (exp + scale_val > max_exp<FMT>()) {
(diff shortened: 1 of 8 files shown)


src/arch/amdgpu/common/dtype/mxfp.hh
```diff
@@ -194,6 +204,41 @@ class mxfp
         data = in.storage;
     }
 
+    // Helper method specific to AMDGPU instructions.
+    void
+    omodModifier(unsigned omod)
+    {
+        assert(omod < 4);
+
+        if (omod == 1) scaleMul(2.0f);
```
mattsinc, line 222: perhaps a comment explaining these?

mattsinc (changes requested):
1 minor thing

src/arch/amdgpu/common/dtype/mxfp.hh
abmerop: done

mattsinc (approved):

PR #2420, opened 2025-07-02
dev-amdgpu: Update GPU SDMA model to support non-zero VMID packets, MMIOs, and fix SDMA poll livelock

v-ramadas:
This PR adds three fixes:

1. Fix a livelock in GPU SDMA where the SDMA polls memory locations continuously without allowing the Linux drivers to update the location
2. Add support for handling MMIO writes to non-zero VMID page tables
3. Add support for translation non-zero VMID SDMA packets

### DIFF
--- a/src/dev/amdgpu/amdgpu_nbio.cc
+++ b/src/dev/amdgpu/amdgpu_nbio.cc
@@ -204,14 +204,55 @@ AMDGPUNbio::writeMMIO(PacketPtr pkt, Addr offset)
     } else if (offset == AMDGPU_MP0_SMN_C2PMSG_71) {
         // PSP ring size
         psp_ring_size = pkt->getLE<uint32_t>();
+    } else if (is_MI200_regBM_PAGE_TABLE_BASE_ADDR(offset)) {
+        uint16_t context_id =
+            get_context_from_MI200_regBM_PAGE_TABLE_BASE_ADDR(offset);
+        regs[offset] = pkt->getLE<uint32_t>();
+        if ((offset % 8) == 0) {
+            // The register write is to ptBaseH
+            gpuDevice->getVM().setPageTableBaseH(context_id,
+                    pkt->getLE<uint32_t>());
+        } else {
+            // The register write is to ptBaseL
+            gpuDevice->getVM().setPageTableBaseL(context_id,
+                    pkt->getLE<uint32_t>());
+        }
+    } else if (is_MI200_regBM_PAGE_TABLE_START_ADDR(offset)) {
+        uint16_t context_id =
+            get_context_from_MI200_regBM_PAGE_TABLE_START_ADDR(offset);
+        regs[offset] = pkt->getLE<uint32_t>();
+        if ((offset % 8) == 0) {
+            // The register write is to ptBaseH
+            gpuDevice->getVM().setPageTableStartH(context_id,
+                    pkt->getLE<uint32_t>());
+        } else {
+            // The register write is to ptBaseL
+            gpuDevice->getVM().setPageTableStartL(context_id,
+                    pkt->getLE<uint32_t>());
+        }
+    } else if (is_MI200_regBM_PAGE_TABLE_END_ADDR(offset)) {
+        uint16_t context_id =
+            get_context_from_MI200_regBM_PAGE_TABLE_END_ADDR(offset);
+        regs[offset] = pkt->getLE<uint32_t>();
+        // MI200 page table addresses are 64 bits long. There are
+        // separate registers to handle the lower 32 bits and upper 32
+        // bits. Use the MMIO offset to figure out which part of the
+        // address is being written to
+        if ((offset % 8) == 0) {
+            // The register write is to ptBaseH
+            gpuDevice->getVM().setPageTableEndH(context_id,
+                    pkt->getLE<uint32_t>());
+        } else {
+            // The register write is to ptBaseL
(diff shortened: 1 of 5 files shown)


v-ramadas:
@abmerop, I've included a commit to add delays to SDMA read/write operations in this PR.

src/dev/amdgpu/amdgpu_nbio.hh
```diff
@@ -103,6 +103,20 @@ class AMDGPUDevice;
...
+#define GET_CONTEXT_FROM_regBM_PAGE_TABLE_BASE_ADDR(offset) \
+        (((offset - 0x6b0ac) >> 2)/2)
+#define IS_regBM_PAGE_TABLE_START_ADDR(offset) \
+        ((offset >= 0x6b12c && offset <=0x6b1a8) ? true : false)
+#define GET_CONTEXT_FROM_regBM_PAGE_TABLE_START_ADDR(offset) \
+        (((offset - 0x6b12c) >> 2)/2)
+#define IS_regBM_PAGE_TABLE_END_ADDR(offset) \
+        ((offset >= 0x6b1ac && offset <=0x6b1c8) ? true : false)
+#define GET_CONTEXT_FROM_regBM_PAGE_TABLE_END_ADDR(offset) \
+        (((offset - 0x6b1ac) >> 2)/2)
```
abmerop: These need to have MI200 somewhere in the name. The offsets here are different in MI300.

It might be better to have these be class methods. These can't be resolved at compile time anyway and the method would more easily allow for checking the gfx version and selecting the appropriate MMIO addresses.

src/dev/amdgpu/sdma_engine.cc
```diff
@@ -366,7 +366,7 @@ SDMAEngine::decodeNext(SDMAQueue *q)
         auto cb = new DmaVirtCallback<uint32_t>(
             [ = ] (const uint32_t &header)
                 { decodeHeader(q, header); });
-        dmaReadVirt(q->rptr(), sizeof(uint32_t), cb, &cb->dmaBuffer);
+        dmaReadVirt(q->rptr(), sizeof(uint32_t), cb, &cb->dmaBuffer, 1e9);
```
abmerop: `1e9` should probably be defined as a `static constexpr int` in some header file so we could play with the value more easily if it ends up being a latency bottleneck in the future. Same applies to the usages elsewhere.

src/dev/amdgpu/amdgpu_nbio.cc
```diff
@@ -204,14 +204,51 @@ AMDGPUNbio::writeMMIO(PacketPtr pkt, Addr offset)
     } else if (offset == AMDGPU_MP0_SMN_C2PMSG_71) {
         // PSP ring size
         psp_ring_size = pkt->getLE<uint32_t>();
+    } else if (IS_regBM_PAGE_TABLE_BASE_ADDR(offset)) {
+        uint16_t context_id =
+            GET_CONTEXT_FROM_regBM_PAGE_TABLE_BASE_ADDR(offset);
+        regs[offset] = pkt->getLE<uint32_t>();
+        if ((offset % 8) == 0) {
```
mattsinc, line 211: add comment explaining what 8 means

src/dev/amdgpu/sdma_engine.cc
```diff
@@ -461,7 +464,7 @@ SDMAEngine::decodeHeader(SDMAQueue *q, uint32_t header)
             cb = new DmaVirtCallback<uint64_t>(
                 [ = ] (const uint64_t &)
                     { write(q, (sdmaWrite *)dmaBuffer); });
-            dmaReadVirt(q->rptr(), sizeof(sdmaWrite), cb, dmaBuffer);
+            dmaReadVirt(q->rptr(), sizeof(sdmaWrite), cb, dmaBuffer, 1e9);
```
mattsinc: not sure if it's a GitHub oddity, but spacing looks wrong here (and on line 431 above)

mattsinc (changes requested):

src/dev/amdgpu/amdgpu_nbio.hh
v-ramadas: Done

src/dev/amdgpu/sdma_engine.cc
v-ramadas: Done

src/dev/amdgpu/amdgpu_nbio.cc
v-ramadas: Done

src/dev/amdgpu/sdma_engine.cc
v-ramadas: I don't see this in my editor. Must be a GitHub oddity? I remember this happening earlier too

abmerop (commented):
I think it would be a good idea still to either `#define` the MMIO addresses or `static constexpr int` them.

I know this is leading towards greater work, but this would fix a potential future style issue

mattsinc:
> I think it would be a good idea still to either `#define` the MMIO addresses or `static constexpr int` them.
> 
> I know this is leading towards greater work, but this would fix a potential future style issue

we should just fix it now before merging in then

v-ramadas:
Thanks for the suggestions! I've moved the register addresses to `#define` macros.

mattsinc (approved):
This appears to address @abmerop 's comments, but will let him chime in before merging

abmerop (approved):

PR #2422, opened 2025-07-03
arch-vega: Fix incorrect address translation caused by tlb

TomXia:
This commit fixes an issue where the tlb returns incorrect physical address if the pagesize is other than 4KB.

### DIFF
--- a/src/arch/amdgpu/vega/tlb.cc
+++ b/src/arch/amdgpu/vega/tlb.cc
@@ -361,10 +361,8 @@ GpuTLB::issueTLBLookup(PacketPtr pkt)
             // Set if this is a system request
             pkt->req->setSystemReq(entry->pte.s);
 
-            Addr alignedPaddr = pageAlign(entry->paddr);
             sender_state->tlbEntry =
-                new VegaTlbEntry(1 /* VMID */, virt_page_addr, alignedPaddr,
-                                 entry->logBytes, entry->pte);
+                new VegaTlbEntry(*entry);
         }
 
         if (update_stats) {
@@ -847,7 +845,7 @@ GpuTLB::CpuSidePort::recvFunctional(PacketPtr pkt)
                 DPRINTF(GPUTLB, "Mapping %#x to %#x\n", vaddr, paddr);
 
                 sender_state->tlbEntry =
-                    new VegaTlbEntry(1 /* VMID */, virt_page_addr,
+                    new VegaTlbEntry(1 /* VMID */, vaddr & (~mask(logBytes)),
                                  alignedPaddr, logBytes, pte);
             } else {
                 // If this was a prefetch, then do the normal thing if it
@@ -858,7 +856,8 @@ GpuTLB::CpuSidePort::recvFunctional(PacketPtr pkt)
                     DPRINTF(GPUTLB, "Mapping %#x to %#x\n", vaddr, paddr);
 
                     sender_state->tlbEntry =
-                        new VegaTlbEntry(1 /* VMID */, virt_page_addr,
+                        new VegaTlbEntry(1 /* VMID */,
+                                     vaddr & (~mask(logBytes)),
                                      alignedPaddr, logBytes, pte);
                 } else {
                     DPRINTF(GPUPrefetch, "Prefetch failed %#x\n", vaddr);
@@ -878,9 +877,7 @@ GpuTLB::CpuSidePort::recvFunctional(PacketPtr pkt)
                     entry->vaddr);
         }
 
-        sender_state->tlbEntry = new VegaTlbEntry(1 /* VMID */, entry->vaddr,
-                                                 entry->paddr, entry->logBytes,
-                                                 entry->pte);
+        sender_state->tlbEntry = new VegaTlbEntry(*entry);
     }
 
     // This is the function that would populate pkt->req with the paddr of


mattsinc:
@BobbyRBruce , @powerjg this is the hotfix we discussed via email

mattsinc (commented):
@TomXia : are you sure this is the right patch?  The commit message says it's about the hotfix for the wrong physical address, but the contents appear to be about adding explicit constructor and destructors in -- which is also needed, but not a hotfix...

TomXia:
@mattsinc Yes this is the patch for the tlb fix. Main changes are made in tlb.cc, which use copy constructor. So I also included explicit copy constructor in the patch.

src/arch/amdgpu/vega/tlb.cc
```diff
@@ -361,10 +361,8 @@ GpuTLB::issueTLBLookup(PacketPtr pkt)
             // Set if this is a system request
             pkt->req->setSystemReq(entry->pte.s);
 
-            Addr alignedPaddr = pageAlign(entry->paddr);
             sender_state->tlbEntry =
-                new VegaTlbEntry(1 /* VMID */, virt_page_addr, alignedPaddr,
-                                 entry->logBytes, entry->pte);
+                new VegaTlbEntry(*entry);
```
mattsinc, line 365: can you please explain why replacing this wholesale resolves the problem?  I guess because virt_page_addr is not used inside the copy constructor?

TomXia: Yes, virt_page_addr might be different from entry->vaddr for 2MB pages.

Using the copy constructor guarantees the returned new TLB entry is the same as the existing entry.

TomXia: I think the copy constructor isn't really necessary here. Reverted the changes on that part.

mattsinc: With the updated change you've made, the copy constructor is now implicit and we're relying on the fact that it doesn't use something like virt_page_addr internally.  Personally, I would feel better if we used the original version you uploaded, but @abmerop may feel differently.

TomXia: The implicit copy constructor does 1 to 1 copy for all member variables. virt_page_addr isn't a member, so it should be fine. But, maybe we should not be using the exact copy here.

mattsinc: I will wait for @abmerop to chime in then ... I'm ok with breaking those two commits apart and only having this as the hotfix, but Matt P wanted both (I think) together.

abmerop: > The implicit copy constructor does 1 to 1 copy for all member variables. virt_page_addr isn't a member, so it should be fine. But, maybe we should not be using the exact copy here.

This is correct here because the TLB entry is all POD types (we'd need to make sure the types had copy constructors otherwise to do a deeper copy)

abmerop (commented):
Looks good code-wise however you should mention in the commit message that the implicit copy constructor is using the post-translated page size instead of the "guessed" or "assumed" 4kb page align.  This is just to make it clear for a potential hotfix,

src/arch/amdgpu/vega/tlb.cc
mattsinc: sounds like you are fine merging as is then?

mattsinc (dismissed):

abmerop (approved):

mattsinc:
@BobbyRBruce , @powerjg: from @abmerop and my points of view, this is ready for a hotfix merge -- let us know

abmerop (changes requested):
One more thing:  Do we need this change also on line 847-849?  (GitHub highlighting isn't working for me right now).

mattsinc:
> One more thing: Do we need this change also on line 847-849? (GitHub highlighting isn't working for me right now).

in the latest version, I don't see these lines being changed -- which file?

abmerop:
> > One more thing: Do we need this change also on line 847-849? (GitHub highlighting isn't working for me right now).
> 
> in the latest version, I don't see these lines being changed -- which file?

Will just do this for now.  I only see one file being changed: 
{C37AB2E4-C7AB-4AC5-8D90-FE5614801FBE}

Edit: And
{DE1A0A1C-3817-4318-8900-AA917704A5FF}

Basically grep for all `new VegaTlbEntry` and they should all be replaced with the new (implicit) copy constructor

mattsinc:
> > > One more thing: Do we need this change also on line 847-849? (GitHub highlighting isn't working for me right now).
> > 
> > 
> > in the latest version, I don't see these lines being changed -- which file?
> 
> Will just do this for now. I only see one file being changed: {C37AB2E4-C7AB-4AC5-8D90-FE5614801FBE}
> 
> Edit: And {DE1A0A1C-3817-4318-8900-AA917704A5FF}
> 
> Basically grep for all `new VegaTlbEntry` and they should all be replaced with the new (implicit) copy constructor

sorry I misread -- I thought you said they were changed, but you clearly said they *should* be changed.  I agree they should also be changed -- @TomXia please update them too

TomXia:
Doing functional page walk from TLB behaves a bit differently from the normal page walk. I didn't notice it was also incorrect last time.

The updated change is what I think it is supposed to be. But I don't have a good way to actually test the functional operation. Please take a look. @mattsinc @abmerop 

Thanks.

mattsinc:
> Doing functional page walk from TLB behaves a bit differently from the normal page walk. I didn't notice it was also incorrect last time.
> 
> The updated change is what I think it is supposed to be. But I don't have a good way to actually test the functional operation. Please take a look. @mattsinc @abmerop
> 
> Thanks.

I think the way to test would be to use the functionalTLB configuration option, no?

src/arch/amdgpu/vega/tlb.cc
```diff
@@ -847,7 +845,7 @@ GpuTLB::CpuSidePort::recvFunctional(PacketPtr pkt)
                 DPRINTF(GPUTLB, "Mapping %#x to %#x\n", vaddr, paddr);
 
                 sender_state->tlbEntry =
-                    new VegaTlbEntry(1 /* VMID */, virt_page_addr,
+                    new VegaTlbEntry(1 /* VMID */, vaddr & (~mask(logBytes)),
```
mattsinc, line 848: shouldn't this just use the copy constructor like your other changes?

src/arch/amdgpu/vega/tlb.cc
```diff
@@ -858,7 +856,8 @@ GpuTLB::CpuSidePort::recvFunctional(PacketPtr pkt)
                     DPRINTF(GPUTLB, "Mapping %#x to %#x\n", vaddr, paddr);
 
                     sender_state->tlbEntry =
-                        new VegaTlbEntry(1 /* VMID */, virt_page_addr,
+                        new VegaTlbEntry(1 /* VMID */,
```
mattsinc, line 859: shouldn't this just use the copy constructor like your other changes?

mattsinc (changes requested):

src/arch/amdgpu/vega/tlb.cc
TomXia: Can't really use copy constructor here because functional page walk does not return a TLB Entry, so no TLB Entry to copy from. Need to construct a new TLB entry with all parameters instead.

src/arch/amdgpu/vega/tlb.cc
TomXia: Same as above.

src/arch/amdgpu/vega/tlb.cc
mattsinc: fair enough

mattsinc (approved):
I think this addresses @abmerop 's satisfactorily, but will wait for him to chime in about the functional path.

BobbyRBruce (approved):
This appears ok to me. Happy to make this hotfix once merged into develop. When that's done I can handle the rest.

abmerop (approved):

mattsinc:
> This appears ok to me. Happy to make this hotfix once merged into develop. When that's done I can handle the rest.

Done, thanks @BobbyRBruce !

mattsinc:
> This appears ok to me. Happy to make this hotfix once merged into develop. When that's done I can handle the rest.

@BobbyRBruce just wanted to follow up on this -- let us know if you need anything else for the hotfix.

PR #2424, opened 2025-07-04
dev: Make DmaPort can be inherited

wmin0:
DmaPort is a base class for modeling a port with DMA ability. It's possible that the port needing to support different kinds of protocol with different kinds of metadata. For example for modeling AXI, we need to add AxCache and AxDomain in the request.

To make the base class be able to extend, I made the private fields to protected. Secondly, I made DmaReqState::createPacket be able to override which allows creating a packet with extra information.

Change-Id: I32d4ab1b6388c460d80769969bbbee3b824d2c50

### DIFF
--- a/src/dev/dma_device.hh
+++ b/src/dev/dma_device.hh
@@ -61,7 +61,7 @@ class ClockedObject;
 
 class DmaPort : public RequestPort, public Drainable
 {
-  private:
+  protected:
     AddrRangeMap<MemBackdoorPtr, 1> memBackdoors;
 
     /**
@@ -80,6 +80,11 @@ class DmaPort : public RequestPort, public Drainable
      */
     void sendDma();
 
+    /**
+     * DmaReqState contains data of a DMA request and is responsible for
+     * creating gem5 requests accordingly. The class can be extended to
+     * generate different kinds of DMA traffic.
+     */
     struct DmaReqState : public Packet::SenderState
     {
         /** Event to call on the device when this transaction (all packets)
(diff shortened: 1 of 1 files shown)


src/dev/dma_device.hh
```diff
@@ -129,7 +129,7 @@ class DmaPort : public RequestPort, public Drainable
               sid(_sid), ssid(_ssid), cmd(_cmd)
         {}
 
-        PacketPtr createPacket();
+        virtual PacketPtr createPacket();
```
giactra, line 141: My sugestion is that unless you plan to upstream your child port, it would be better if you added a comment here. Otherwise people might see no class is extending the DmaPort and would decide to "unvirtualize" the createPacket method

wmin0: That's fair. Thanks for the suggestion. I can update this part by next week.

wmin0:
Hi @giactra,

I updated the code comment. Please check.

Thanks :)

erin-le:
@giactra: Hello, could you take another look at this PR if you have time? Thank you!

wmin0:
@giactra friendly ping :)

erin-le:
@giactra: Hello, could you take another look at this PR when you get a chance? Thank you!

powerjg (approved):

powerjg:
@giactra let us know if we should *not* merge this. Otherwise, we'll merge by the end of the week.

giactra (approved):
Apologies, please go ahead and merge

PR #2425, opened 2025-07-05
configs: Allow GPU-FS to use HBMCtrl memory controller

TomXia:
This commit adds a command line argument --hbm-ctrl to enable HBMCtrl in GPU-FS. HBM_2000_4H_1x64 is tested.

### DIFF
--- a/configs/example/gpufs/Disjoint_VIPER.py
+++ b/configs/example/gpufs/Disjoint_VIPER.py
@@ -107,8 +107,17 @@ class Disjoint_VIPER(RubySystem):
         system.memories = cpu_abstract_mems
 
         gpu_abstract_mems = []
+
         for mem_ctrl in gpu_mem_ctrls:
-            gpu_abstract_mems.append(mem_ctrl.dram)
+            # memctrl
+            if hasattr(mem_ctrl, "dram"):
+                gpu_abstract_mems.append(mem_ctrl.dram)
+            else:
+                gpu_abstract_mems.append(mem_ctrl)
+            # hbmctrl
+            if hasattr(mem_ctrl, "dram_2"):
+                gpu_abstract_mems.append(mem_ctrl.dram_2)
+
         system.pc.south_bridge.gpu.memories = gpu_abstract_mems
 
         # Setup DMA controllers
--- a/configs/ruby/GPU_VIPER.py
+++ b/configs/ruby/GPU_VIPER.py
@@ -669,23 +669,53 @@ def construct_gpudirs(options, system, ruby_system, network):
 
         # Create memory controllers too
         mem_type = ObjectList.mem_list.get(options.dgpu_mem_type)
-        dram_intf = MemConfig.create_mem_intf(
-            mem_type,
-            gpu_mem_range,
-            i,
-            int(math.log(options.dgpu_num_dirs, 2)),
-            options.cacheline_size,
-            xor_low_bit,
-        )
+
+        if options.hbm_ctrl:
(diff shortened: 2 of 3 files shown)


configs/example/gpufs/Disjoint_VIPER.py
```diff
@@ -107,8 +107,17 @@ def create(self, options, system, piobus, dma_devices):
         system.memories = cpu_abstract_mems
 
         gpu_abstract_mems = []
+
         for mem_ctrl in gpu_mem_ctrls:
-            gpu_abstract_mems.append(mem_ctrl.dram)
+            # memctrl
+            if hasattr(mem_ctrl, "dram"):
```
mattsinc, line 113: instead of adding a second memory controller for the new HBM, shouldn't we add 2 dram's if the new model exists?  I guess this code seems a little complex to me, but maybe there is a good reason.  But it seems like this:

```
if (hbm_ctrl)
  gpu_abstract_mems.append(mem_ctrl.dram)
  gpu_abstract_mems.append(mem_ctrl.dram_2)
else if (!hbm_ctrl but using a different hbm)
  gpu_abstract_mems.append(mem_ctrl.dram)
else
  gpu_abstract_mems.append(mem_ctrl)
```
  
is more straightforward, no?

configs/ruby/GPU_VIPER.py
```diff
@@ -669,23 +669,51 @@ def construct_gpudirs(options, system, ruby_system, network):
...
-        dram_intf = MemConfig.create_mem_intf(
-            mem_type,
-            gpu_mem_range,
-            i,
-            int(math.log(options.dgpu_num_dirs, 2)),
-            options.cacheline_size,
-            xor_low_bit,
-        )
+
+        if options.hbm_ctrl:
```
mattsinc, line 673: please add a comment about why we are adding 2 controllers here -- it's obvious to you, but less obvious to others trying to use this

mattsinc (changes requested):
A couple minor things, but otherwise looks reasonable -- thanks!

configs/example/gpufs/Disjoint_VIPER.py
TomXia: I was thinking about making it config-agnostic. Effectively it detects which controller is being used and adds the memory to system, without having to specify in args or change config (tested with MemCtrl, HBMCtrl and SimpleMemory). But, of course, I am happy to change it to a more direct way.

mattsinc: @abmerop thoughts?

abmerop: That is fine with me, although I believe we want to try to move to stdlib for most things (except GPU / X86 right now due to lack of >3GB memory)

abmerop (dismissed):
Is there a related change needed in stdlib files or does that already support this?

configs/example/gpufs/Disjoint_VIPER.py
mattsinc: ok, we can leave it be then

mattsinc:
@TomXia please address my other question and @abmerop stdlib question

TomXia:
@abmerop :The stdlib already supports this. All we need is to update GPU configs.
@mattsinc :Added comments explaining the HBM config.

Thanks for reviewing!

mattsinc (approved):
Will wait for @abmerop to ok before merging

abmerop (approved):

PR #2427, opened 2025-07-07
arch-vega: Implement all MI355X conversion instructions

abmerop:
There are 5 types of instructions included in this commit:
 - "Regular" conversion instructions: These are implemented one by one in VOP3 like any other instruction. Converts from one data type to another with no microscaling factor. May be packed or use stochastic rounding (SR) but not both.
 - Scaled, packed instructions: These are implemented as a template class which handles 18 instructions. Converts packed values usings an E08M scaling factor (which is placed in an F32 in the exponent bits).
 - Scaled, unpacked instructions: Implemented as a template handling 4 instructions. Same as above but are not packed.
 - Scaled, 32-pack instructions: These are implemented as a template handling 12 instructions. The 32-pack consists of 32 6-bit FPs packed into 6 VGPRs. Scaled the same as above. There is a more specialized template for F32 sources.
 - Scaled, SR instructions: Same as scaled, unpacked instructions but uses SR exclusively for downcasting. Handles 6 instructions.
 - Scaled, 32-pack, SR instructions: Similar to scaled, 32-pack but always downcasts and hence uses SR as well. Handles 6 instructions.
 - Scaled, packed, SR instructions: Handles downcasts to FP4 specially. Implements 3 instructions.

In total this patch adds 54 instructions.


abmerop:
This is a lot of code -- No rush here as we can't simulate this until ROCm 7.0 is out of alpha.  These instructions have been tested on a ROCm 7.0 pre-release however.

src/arch/amdgpu/vega/insts/vop1.cc
```diff
@@ -2554,6 +2554,38 @@ namespace VegaISA
...
+    {
+    } // ~Inst_VOP1__V_CVT_F32_BF16
+
+    void
+    Inst_VOP1__V_CVT_F32_BF16::execute(GPUDynInstPtr gpuDynInst)
+    {
+        // Using U16/U32 to avoid implicit conversions
+        Wavefront *wf = gpuDynInst->wavefront();
+        ConstVecOperandU16 src(gpuDynInst, instData.SRC0);
+        VecOperandU32 vdst(gpuDynInst, instData.VDST);
```
mattsinc: Shouldn't vdst be a f32-type register (if it exists) given the output type?

src/arch/amdgpu/vega/insts/vop1.cc
```diff
@@ -2587,6 +2619,270 @@ namespace VegaISA
...
+    Inst_VOP1__V_CVT_F32_FP8::~Inst_VOP1__V_CVT_F32_FP8()
+    {
+    } // ~Inst_VOP1__V_CVT_F32_FP8
+
+    void
+    Inst_VOP1__V_CVT_F32_FP8::execute(GPUDynInstPtr gpuDynInst)
+    {
+        Wavefront *wf = gpuDynInst->wavefront();
+
+        ConstVecOperandU32 src(gpuDynInst, instData.SRC0);
```
mattsinc, line 2642: should this be U8 instead given src input size?

src/arch/amdgpu/vega/insts/vop1.cc
```diff
@@ -2587,6 +2619,270 @@ namespace VegaISA
...
+        } else {
+            for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
+                srcData[lane] = src[lane];
+            }
+        }
+
+        for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
+            if (wf->execMask(lane)) {
+                AMDGPU::mxfloat8
+                    in(bits(srcData[lane], byte * 8 + 7, byte * 8));
```
mattsinc, line 2674: I can't remember if we removed it, but would the old SDWA code that tried to support all the different variants be better here?  Then we wouldn't need to worry so much about handling byte > 3 ...

src/arch/amdgpu/vega/insts/vop1.cc
```diff
@@ -2587,6 +2619,270 @@ namespace VegaISA
...
+    {
+    } // ~Inst_VOP1__V_CVT_PK_F32_FP8
+
+    void
+    Inst_VOP1__V_CVT_PK_F32_FP8::execute(GPUDynInstPtr gpuDynInst)
+    {
+        Wavefront *wf = gpuDynInst->wavefront();
+
+        ConstVecOperandU32 src(gpuDynInst, instData.SRC0);
+        VecOperandU64 vdst(gpuDynInst, instData.VDST);
```
mattsinc, line 2767: why u64 given the output size (f32)?

src/arch/amdgpu/vega/insts/vop1.cc
```diff
@@ -2587,6 +2619,270 @@ namespace VegaISA
...
+    {
+    } // ~Inst_VOP1__V_CVT_PK_F32_BF8
+
+    void
+    Inst_VOP1__V_CVT_PK_F32_BF8::execute(GPUDynInstPtr gpuDynInst)
+    {
+        Wavefront *wf = gpuDynInst->wavefront();
+
+        ConstVecOperandU32 src(gpuDynInst, instData.SRC0);
+        VecOperandU64 vdst(gpuDynInst, instData.VDST);
```
mattsinc, line 2837: same question here

src/arch/amdgpu/vega/insts/vop1.cc
abmerop: Yes, good catch! There is an implicit convert to U32 on line `vdst[lane] = float(tmp);` which will truncate the value... Will fix

src/arch/amdgpu/vega/insts/vop1.cc
abmerop: I am using U32 here so that it can be assigned to the `srcData` array which gets populated depending on whether or not this is SDWA. That way we populate based on SDWA, DPP, or vanilla and can use the same code that actually executes the instruction rather than copying the execution part multiple times and potentially have to fix it multiple times.

We could however use ConstVecOperandU8 on line 2666 (else statement) and implicitly convert to U32 there if that is less confusing

src/arch/amdgpu/vega/insts/vop1.cc
abmerop: This is a special case instruction it seems where no SDWA fields are used besides SRC0_SEL. This is why the dest register helper is missing too.

My other issue with using the SDWA helpers is that I would have to implement the actual instruction execution twice as it returns a reference to a new src0 VecOperand.  I'd again like to avoid that so that if there is some bug here it doesn't need to be fixed in two places.

Essentially what I am trying to do here is very similar to what vop2Helper in src/arch/amdgpu/vega/insts/op_encodings.hh does. I.e., if it is SDWA or DPP, use those helpers but define a lambda function to call on their outputs to avoid duplicate implementations of the actual execution

src/arch/amdgpu/vega/insts/vop1.cc
abmerop: The output is packed F32 in this case, so we want to convert to two F32 values and then bitwise copy those into untyped 64-bit registers

src/arch/amdgpu/vega/insts/vop1.cc
mattsinc: this is fine then as long as you truncate

mattsinc (approved):

PR #2432, opened 2025-07-09
cpu-o3: Specify unit of measurement for LSQ loadToUse stat

giactra:
This is to make it explicit we are referring to cycles and not to (micro/nano)seconds

Change-Id: Ic1a8e3dc804384f4d3de88513c2a5d317032712c


dhschall (approved):
Thanks! LGTM.

giactra:
Thanks @dhschall for you review! FYI when the PR is made of a single commit, we usually Merge with "Squash and Merge" option. This avoids the introduction from GitHub of a merge commit and we get linear history instead :)

dhschall:
Thank you @giactra for the clarification! I'll keep that in mind for the future.

PR #2435, opened 2025-07-09
arch-vega: Add Large Pagesize Support to GPU TLB Coalescer

TomXia:
GPU TLB coalescer now can coalesce address translation requests if they fall within the same larger (e.g. 2MB, configurable) region.

### DIFF
--- a/src/arch/amdgpu/vega/VegaGPUTLB.py
+++ b/src/arch/amdgpu/vega/VegaGPUTLB.py
@@ -69,6 +69,7 @@ class VegaTLBCoalescer(ClockedObject):
     cxx_class = "gem5::VegaTLBCoalescer"
     cxx_header = "arch/amdgpu/vega/tlb_coalescer.hh"
     tlb_level = Param.Int(64, "tlb level")
+    default_pgSize = Param.MemorySize("2MiB", "Default page size to coalesce")
     maxDownstream = Param.Int(64, "max downstream @ this level")
     probesPerCycle = Param.Int(2, "Number of TLB probes per cycle")
     coalescingWindow = Param.Int(1, "Permit coalescing across that many ticks")
--- a/src/arch/amdgpu/vega/tlb_coalescer.cc
+++ b/src/arch/amdgpu/vega/tlb_coalescer.cc
@@ -69,6 +69,9 @@ VegaTLBCoalescer::VegaTLBCoalescer(const VegaTLBCoalescerParams &p)
         memSidePort.push_back(new MemSidePort(csprintf("%s-port%d", name(), i),
                                               this, i));
     }
+
+    default_pgSize = p.default_pgSize;
+    potentialPagesize.insert(default_pgSize);
 }
 
 Port &
@@ -98,7 +101,9 @@ VegaTLBCoalescer::getPort(const std::string &if_name, PortID idx)
  * The rules can potentially be modified based on the TLB level.
  */
 bool
-VegaTLBCoalescer::canCoalesce(PacketPtr incoming_pkt, PacketPtr coalesced_pkt)
+VegaTLBCoalescer::canCoalesce
+(PacketPtr incoming_pkt, PacketPtr coalesced_pkt,
+ Addr pagebytes = VegaISA::PageBytes)
 {
     if (disableCoalescing)
         return false;
@@ -112,10 +117,10 @@ VegaTLBCoalescer::canCoalesce(PacketPtr incoming_pkt, PacketPtr coalesced_pkt)
     // Rule 1: Coalesce requests only if they
     // fall within the same virtual page
     Addr incoming_virt_page_addr = roundDown(incoming_pkt->req->getVaddr(),
-                                             VegaISA::PageBytes);
+                                             pagebytes);
 
     Addr coalesced_virt_page_addr = roundDown(coalesced_pkt->req->getVaddr(),
-                                              VegaISA::PageBytes);
+                                              pagebytes);
 
     if (incoming_virt_page_addr != coalesced_virt_page_addr)
         return false;
@@ -145,11 +150,6 @@ VegaTLBCoalescer::canCoalesce(PacketPtr incoming_pkt, PacketPtr coalesced_pkt)
 void
 VegaTLBCoalescer::updatePhysAddresses(PacketPtr pkt)
 {
-    Addr virt_page_addr = roundDown(pkt->req->getVaddr(), VegaISA::PageBytes);
-
-    DPRINTF(GPUTLB, "Update phys. addr. for %d coalesced reqs for page %#x\n",
-            issuedTranslationsTable[virt_page_addr].size(), virt_page_addr);
-
     GpuTranslationState *sender_state =
         safe_cast<GpuTranslationState*>(pkt->senderState);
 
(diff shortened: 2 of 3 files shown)


src/arch/amdgpu/vega/VegaGPUTLB.py
```diff
@@ -69,6 +69,7 @@ class VegaTLBCoalescer(ClockedObject):
     cxx_class = "gem5::VegaTLBCoalescer"
     cxx_header = "arch/amdgpu/vega/tlb_coalescer.hh"
     tlb_level = Param.Int(64, "tlb level")
+    starting_pgsize = Param.Int(2**21, "Default page size to coalesce")
```
mattsinc: Should we call this default_pgSize instead?

src/arch/amdgpu/vega/tlb_coalescer.cc
```diff
@@ -69,6 +69,9 @@ VegaTLBCoalescer::VegaTLBCoalescer(const VegaTLBCoalescerParams &p)
         memSidePort.push_back(new MemSidePort(csprintf("%s-port%d", name(), i),
                                               this, i));
     }
+
+    starting_pgsize = p.starting_pgsize;
+    observedPagesize.insert(starting_pgsize);
```
mattsinc: Why initialize observedPagesize here?

src/arch/amdgpu/vega/tlb_coalescer.cc
```diff
@@ -160,12 +160,42 @@ VegaTLBCoalescer::updatePhysAddresses(PacketPtr pkt)
...
+        virt_page_addr = roundDown(pkt->req->getVaddr(), pgsize_seen);
+        if (issuedTranslationsTable.count(virt_page_addr) != 0)
+            break;
+    }
+
+    DPRINTF(GPUTLB, "Update phys. addr. for %d \
+            coalesced reqs for page %#x\n",
+            issuedTranslationsTable[virt_page_addr].size(),
+            virt_page_addr);
+
```
mattsinc, line 179: extra line of space here, remove

src/arch/amdgpu/vega/tlb_coalescer.cc
```diff
@@ -160,12 +160,42 @@ VegaTLBCoalescer::updatePhysAddresses(PacketPtr pkt)
...
     bool is_system = pkt->req->systemReq();
 
-    for (int i = 0; i < issuedTranslationsTable[virt_page_addr].size(); ++i) {
+    for (int i = 0;
+         i < issuedTranslationsTable[virt_page_addr].size(); ++i) {
         PacketPtr local_pkt = issuedTranslationsTable[virt_page_addr][i];
+
+        Addr local_pkt_vaddr = local_pkt->req->getVaddr();
+
+        //check if the local packet vaddr belongs to the page translated,
```
mattsinc: I think "page translated" is a bit confusing here.  Maybe something like "check if the returned page translation's virtual address matches any of the pending requests.  If not, reissue pending request as a 4K page" would make more sense?

src/arch/amdgpu/vega/tlb_coalescer.cc
```diff
@@ -236,6 +266,76 @@ VegaTLBCoalescer::updatePhysAddresses(PacketPtr pkt)
...
+    // It is shared by all candidates that fall within the
+    // given coalescingWindow.
+    Tick tick_index = sender_state->issueTime / coalescingWindow;
+
+    if (coalescerFIFO.count(tick_index)) {
+        coalescedReq_cnt = coalescerFIFO[tick_index].size();
+    }
+
+    // see if we can coalesce the incoming pkt with another
+    // coalesced request with the same tick_index
```
mattsinc, line 295: Why does tick index matter for coalescing?  Wouldn't any tick within a window be ok?

src/arch/amdgpu/vega/tlb_coalescer.cc
```diff
@@ -236,6 +266,76 @@ VegaTLBCoalescer::updatePhysAddresses(PacketPtr pkt)
...
+            didCoalesce = true;
+            break;
+        }
+    }
+
+    // if this is the first request for this tick_index
+    // or we did not manage to coalesce, update stats
+    // and make necessary allocations.
+    if (!coalescedReq_cnt || !didCoalesce) {
+
```
mattsinc: remove extra line of space

src/arch/amdgpu/vega/tlb_coalescer.cc
```diff
@@ -463,12 +565,16 @@ VegaTLBCoalescer::processProbeTLBEvent()
             // coalesced for the same virtual page address
 
             // compute virtual page address for this request
+            // use the assumed page size
```
mattsinc: did you define somewhere what the assumed page size is?

mattsinc (changes requested):

src/arch/amdgpu/vega/VegaGPUTLB.py
abmerop: Optional:  It might be more readable to do something like the following and in the TLB coalescer code do a `starting_pgsize.size()` to get the integer size.

```
starting_pgsize = Param.AddrRange("2MiB", "Default page size to coalesce")
```

abmerop (commented):
Just to confirm, this will handle any page size, correct?  I.e., even if the starting page size is 2MB, if it detects an 8MB page, for example, it will try to coalesce any request in that 8MB region? It looks to be the case in the code but just want to confirm.

src/arch/amdgpu/vega/VegaGPUTLB.py
powerjg: I think it's `MemorySize` that you want here. https://github.com/gem5/gem5/blob/stable/src/python/m5/params.py#L1027C7-L1027C17

src/arch/amdgpu/vega/tlb_coalescer.cc
TomXia: This was copied from recvTimingReq(). But tick index is partially determine by tick window, so effectively it is saying within a window.

src/arch/amdgpu/vega/tlb_coalescer.cc
TomXia: It is stored in iter->second[vector_index].second, the next line, defined in CoalescingFIFO.

TomXia:
> Just to confirm, this will handle any page size, correct? I.e., even if the starting page size is 2MB, if it detects an 8MB page, for example, it will try to coalesce any request in that 8MB region? It looks to be the case in the code but just want to confirm.

It can handle any page size correctly. But It will always try to coalesce in 2MB first (starting pgsize=2MB), because it does not track vaddr and its page size. So for the maximum efficiency, starting pgszie has to be a good value

src/arch/amdgpu/vega/VegaGPUTLB.py
TomXia: Thanks, this is very handy.

src/arch/amdgpu/vega/tlb_coalescer.cc
TomXia: observedPagesize will be used to update all stalled translations after one translation returned, so it must not be empty before that. The name is indeed a bit confusing after it's purpose changed. Renamed to potentialPagesize.

TomXia:
Updated. Thanks!

mattsinc (approved):
at least the changes I asked for are resolved.  Will wait for @abmerop and @powerjg to chime in before merging though

abmerop (approved):

PR #2436, opened 2025-07-09
arch-vega: GPU TLB Multi Page Size Support

TomXia:
This patch modifies the GPU TLB implementation to properly account for page sizes during both insertion and lookup operations, enabling better handling of mixed page sizes. This increase TLB hit rate. List of supported pagesize can be simply expanded in tlb.hh

### DIFF
--- a/src/arch/amdgpu/vega/tlb.cc
+++ b/src/arch/amdgpu/vega/tlb.cc
@@ -153,12 +153,18 @@ GpuTLB::pageAlign(Addr vaddr)
     return (vaddr & ~pageMask);
 }
 
+int
+GpuTLB::getSet(Addr va, unsigned int page_shift)
+{
+    return (va >> page_shift) & setMask;
+}
+
 VegaTlbEntry*
 GpuTLB::insert(Addr vpn, VegaTlbEntry &entry)
 {
     VegaTlbEntry *newEntry = nullptr;
 
-    int set = (entry.vaddr >> VegaISA::PageShift) & setMask;
+    int set = getSet(entry.vaddr, entry.logBytes);
 
     if (!freeList[set].empty()) {
         newEntry = freeList[set].front();
@@ -178,9 +184,9 @@ GpuTLB::insert(Addr vpn, VegaTlbEntry &entry)
 }
 
 GpuTLB::EntryList::iterator
-GpuTLB::lookupIt(Addr va, bool update_lru)
+GpuTLB::lookupIt(Addr va, unsigned int ps, bool update_lru)
 {
-    int set = (va >> VegaISA::PageShift) & setMask;
+    int set = getSet(va, ps);
 
     if (FA) {
         assert(!set);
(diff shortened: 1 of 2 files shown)


mattsinc:
Note: #2435 should not be merged in until after this one is

src/arch/amdgpu/vega/tlb.hh
```diff
@@ -105,10 +105,17 @@ class GpuTLB : public ClockedObject
...
+    EntryList::iterator lookupIt(Addr va, unsigned int ps,
+                                 bool update_lru=true);
     Walker *walker;
     AMDGPUDevice *gpuDevice;
 
+    int getSet(Addr va, unsigned int page_shift);
+
+    //List of possible page size, 4k and 2m for now
+    const std::array<unsigned int, 2>
+    pageShiftList = {VegaISA::PageShift, 21};
```
mattsinc: I would personally prefer this be called "pageSizeList" or "pageSizeShiftList"

src/arch/amdgpu/vega/tlb.cc
```diff
@@ -478,11 +490,11 @@ GpuTLB::handleTranslationReturn(Addr virt_page_addr,
     }
 
     if (tlb_outcome == TLB_HIT) {
-        DPRINTF(GPUTLB, "Translation Done - TLB Hit for addr %#x\n",
+DPRINTF(GPUTLB, "Translation Done - TLB Hit for addr %#x\n",
             vaddr);
         local_entry = safe_cast<VegaTlbEntry *>(sender_state->tlbEntry);
-    } else {
-        DPRINTF(GPUTLB, "Translation Done - TLB Miss for addr %#x\n",
+            } else {
```
mattsinc: Spacing is incorrect on this line and next line

src/arch/amdgpu/vega/tlb.cc
```diff
@@ -190,7 +196,8 @@ GpuTLB::lookupIt(Addr va, bool update_lru)
     for (; entry != entryList[set].end(); ++entry) {
```
mattsinc, line 196: It seems you are having entryList hold pages of multiple sizes and then search through to see if a given page matches the desired size, right?

mattsinc (changes requested):

src/arch/amdgpu/vega/tlb.cc
TomXia: Yes, correct.

src/arch/amdgpu/vega/tlb.hh
TomXia: I changed it to logPageShiftList. And the spacing issue below was also fixed.

Thanks for pointing out

mattsinc (approved):
Will wait for @abmerop to check this too

src/arch/amdgpu/vega/tlb.cc
```diff
@@ -210,14 +217,17 @@ GpuTLB::lookupIt(Addr va, bool update_lru)
...
-    if (entry == entryList[set].end())
-        return nullptr;
-    else
-        return *entry;
+        if (entry == entryList[set].end())
+            continue;
+        else
+            return *entry;
+    }
+    return nullptr;
```
abmerop, line 230: Wouldn't the TLB miss for any page size not in `logPageShiftList`?  Do we need to add page sizes as we see them, e.g., in `insert`?

abmerop: I see, it looks like that won't happen. The page table walker ignores the "fragment" size by using logBytes = 12 + min(9, fragment).  Basically a 2MB page would be returned and any other 2MB pages in that fragment would miss but be returned as a new different entry to TLB.

abmerop (approved):

PR #2441, opened 2025-07-11
arch-arm: Add FEAT_FP16 FP instructions

giactra:
With PR 2071 [1] we introduced FEAT_FP16 into gem5.  While it covered most SIMD FP16 instructions, it left out most of the scalar ones.

This commit complements the original PR by providing the AArch64 scalar floating point implementation of FEAT_FP16

[1]: https://github.com/gem5/gem5/pull/2071

Change-Id: I60837911eedebc2b44590b17c163dee7b6819df3


giactra:
@BobbyRBruce this is the PR I would like to integrate as a fix to v25 as it adds missing instruction sin FEAT_FP16

powerjg (approved):

PR #2442, opened 2025-07-11
tests: fix kernel panic on kvm switch tests

clemdiep:
The resource used for kvm switch tests as kernel arguments set to force the "root=/dev/sda2" ([x86-ubuntu-24.04-boot-with-systemd](https://resources.gem5.org/resources/x86-ubuntu-24.04-boot-with-systemd/raw?database=gem5-resources&version=5.0.0)). The kvm switch test adds some kernel arguments, but use the X86Board default kernel arguments. Which leads to "root=/dev/hda1" and a kernel panic occurs as it cannot mount root fs.

This commits fix that by getting the kernel arguments from the workload instead of the default ones.

### DIFF
--- a/tests/gem5/kvm_switch_tests/configs/boot_kvm_switch_exit.py
+++ b/tests/gem5/kvm_switch_tests/configs/boot_kvm_switch_exit.py
@@ -163,13 +163,20 @@ motherboard = X86Board(
     cache_hierarchy=cache_hierarchy,
 )
 
-kernal_args = motherboard.get_default_kernel_args() + [args.kernel_args]
 workload = obtain_resource(
     "x86-ubuntu-24.04-boot-with-systemd",
     resource_directory=args.resource_directory,
     resource_version="5.0.0",
 )
-workload.set_parameter("kernel_args", kernal_args)
+
+if "kernel_args" in workload.get_parameters():
+    kernel_args = workload.get_parameters()["kernel_args"]
+else:
+    kernel_args = motherboard.get_default_kernel_args()
+
+kernel_args += [args.kernel_args]
+
+workload.set_parameter("kernel_args", kernel_args)
 # Set the Full System workload.
 motherboard.set_workload(workload)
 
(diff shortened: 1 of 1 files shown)


mattsinc:
@abmerop , @v-ramadas : just checking if we need to integrate this with GPUFS?

powerjg (commented):
I feel like this should be fixed in `set_kernel_disk_workload` if the workload's kernel arguments aren't getting passed to the board.

Maybe we only need to remove line 166?

cc @Harshil2107

tests/gem5/kvm_switch_tests/configs/boot_kvm_switch_exit.py
```diff
@@ -163,13 +163,20 @@
...
+if "kernel_args" in workload.get_parameters():
+    kernel_args = workload.get_parameters()["kernel_args"]
+else:
+    kernel_args = motherboard.get_default_kernel_args()
+
+kernel_args += [args.kernel_args]
+
+workload.set_parameter("kernel_args", kernel_args)
 # Set the Full System workload.
 motherboard.set_workload(workload)
```
Harshil2107: This change should fix the issue, but I believe we should be using the `append_kernel_arg` function instead. Boards that inherit from `KernelDiskWorkload` provide this function, which is intended to append kernel arguments to the existing ones. In our case, it should allow appending the arguments passed via argparse to the current `kernel_args` in the workload.

I recall there was a bug where `append_kernel_arg` wasn't actually appending the arguments. That issue was reported for RISC-V and appears to be fixed in #1936 . I'm currently double-checking whether it works as intended for x86 and Arm as well.

```suggestion
# Set the Full System workload.
motherboard.set_workload(workload)
motherboard.append_kernel_arg(args.kernel_args)
```

Harshil2107: The `append_kernel_arg` function seems to work as intended—it correctly adds the passed arguments to the `kernel_args` defined in the workload. I tested the change mentioned above by running the quick KVM switch tests, and they pass.

abmerop:
> @abmerop , @v-ramadas : just checking if we need to integrate this with GPUFS?

No, we don't use resources

erin-le:
@clemdiep: Hello, just letting you know that I'm planning to edit this PR to use Harshil's suggested changes.

erin-le (approved):

Harshil2107 (approved):
lgtm

PR #2443, opened 2025-07-11
misc: Update ci-tests with clang format

powerjg:
This should cause a failure if any commits fail clang-format

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -23,6 +23,39 @@ jobs:
                   python-version: '3.10'
             - uses: pre-commit/action@v3.0.1
 
+    # Runs on the github hosted runner
+    # Iterates through all of the commits in the the PR and runs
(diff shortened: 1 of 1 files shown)


powerjg:
I have tested the cases below. This is ready for review.

When I have a bad commit, followed by a good commit (the ci test update), things fail correctly.
When I have a good commit (the ci test update) followed by a bad commit, things fail correctly.
When I have a good commit (the ci test update) followed by another good commit, things work correctly.

giactra (approved):
Thank you so much @powerjg for looking into this. This LGTM though I will let @BobbyRBruce merge this as he's maintaining the CI

arichardson:
I wonder if it would make sense to reuse the LLVM workflow logic? Whenever I contribute there, I find the comment posted on the PR quite useful.

The code is here: https://github.com/llvm/llvm-project/blob/main/.github/workflows/pr-code-format.yml

It seems quite similar to your approach just with a few extra bits of logic.

PR #2445, opened 2025-07-14
Fix-orchestrator-workload-id

powerjg:
The `workload` member of `Simulator` can be None (e.g., when
someone manually calls `set_kernel_disk_workload`). This
change updates the logic so that it doesn't cause an error.


Harshil2107 (approved):
lgtm

PR #2447, opened 2025-07-16
systemc: Fix tlm::tlm_dmi usage

yp155136:
The address range presented by tlm::tlm_dmi is inclusive, so when converting to gem5::MemBackdoor, we should + 0x1 on end address.

Change-Id: Ie2f14df88b485889cd3a708140b5dddddddddddd

### DIFF
--- a/src/systemc/tlm_bridge/gem5_to_tlm.cc
+++ b/src/systemc/tlm_bridge/gem5_to_tlm.cc
@@ -284,7 +284,10 @@ Gem5ToTlmBridge<BITWIDTH>::getBackdoor(tlm::tlm_generic_payload &trans)
 
     // If the target gave us one, translate it to a gem5 MemBackdoor and
     // store it in our cache.
-    AddrRange dmi_r(dmi_data.get_start_address(), dmi_data.get_end_address());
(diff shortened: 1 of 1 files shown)


src/systemc/tlm_bridge/gem5_to_tlm.cc
```diff
@@ -284,7 +284,8 @@ Gem5ToTlmBridge<BITWIDTH>::getBackdoor(tlm::tlm_generic_payload &trans)
 
     // If the target gave us one, translate it to a gem5 MemBackdoor and
     // store it in our cache.
-    AddrRange dmi_r(dmi_data.get_start_address(), dmi_data.get_end_address());
+    AddrRange dmi_r(dmi_data.get_start_address(),
+                    dmi_data.get_end_address() + 0x1);
```
powerjg: This needs a comment. Can you simply copy in what you said in the commit message? Also, I would use base 10 instead of base 16 since it's really just "1", not an address or anything like that.

yp155136: Thanks for the suggestion! Updated this PR accordingly!

powerjg (approved):
Thanks for the update!

PR #2452, opened 2025-07-16
configs: deprecate configs/example/gem5_library/riscv-fs.py

erin-le:
This PR deprecates `configs/example/gem5_library/riscv-fs.py`, as we no longer support `riscv-disk-img` for gem5 v25.0, and updating the disk image would make `riscv-fs.py` nearly identical to `riscv-ubuntu-run.py`.


powerjg:
+1 

Others should chime in if they are using this script. If this script has something that `riscv-ubuntu-run.py` doesn't we should try to expand `riscv-ubuntu-run.py` or add another example that has those features.

30ZHANGjin:
1.  After this deprecation, what will be the recommended way to run a simulation with a custom kernel and disk image? Can `riscv-ubuntu-run.py` be easily modified for this purpose?
2.  How can we adapt `riscv-ubuntu-run.py` to run an interactive simulation that does not automatically exit, allowing for manual debugging with `m5term`, similar to the previous behavior of `riscv-fs.py`?

Harshil2107:
Hi @30ZHANGjin,

> 1. After this deprecation, what will be the recommended way to run a simulation with a custom kernel and disk image? Can `riscv-ubuntu-run.py` be easily modified for this purpose?

Yes, `riscv-ubuntu-run.py` can be easily modified to use custom kernel and disk images. There are two main approaches:

* **Using a local resources JSON file**:
  You can create a JSON file that defines your custom resources and use it with gem5 via `obtain_resource`. For details, refer to the documentation here:
  [https://www.gem5.org/documentation/gem5-stdlib/using-local-resources](https://www.gem5.org/documentation/gem5-stdlib/using-local-resources)

* **Directly using local file paths**:
  You can use the `DiskImageResource` and `KernelResource` classes directly in your config and provide the local paths. For example:

  ```python
  from gem5.resources.resource import KernelResource, DiskImageResource

      kernel=KernelResource("path/to/kernel"),
      disk_image=DiskImageResource("path/to/disk"),
      ...
  )
  ```

> 2. How can we adapt `riscv-ubuntu-run.py` to run an interactive simulation that does not automatically exit, allowing for manual debugging with `m5term`, similar to the previous behavior of `riscv-fs.py`?

If you're using any of the Ubuntu 22.04 or 24.04 disk images (including the NPB and GAPBS variants) provided via gem5 resources, you can add the kernel argument `interactive=true`. This prevents gem5 from exiting automatically after the run script completes, enabling manual interaction via `m5term`. You can set the flag as follows:

```python
board.append_kernel_arg("interactive=true")
```

Let me know if you have any other questions.

30ZHANGjin:
@Harshil2107 
Thank you  much for the clear and detailed solutions! This is exactly what I was looking for.

Harshil2107 (approved):

PR #2456, opened 2025-07-17
cpu,sim,switch: Fix assertion failure in switchable CPUs due to doClone

### LINKED ISSUE #2446: SE mode, multithread execution calling doClone before switchCpus causes assertion `ntc.getProcessPtr() == otc.getProcessPtr()' failure
**Describe the bug**
When simulating a simple multithreaded program using OpenMP with SimpleSwitchableProcessor and using m5ops in the simulated program to mark ROI causes assertion error:
```shell
gem5.opt: src/cpu/thread_context.cc:254: void gem5::takeOverFrom(gem5::ThreadContext&, gem5::ThreadContext&): Assertion `ntc.getProcessPtr() == otc.getProcessPtr()' failed.
```
Further investigation reveals that the error is caused when ROI starts, i.e. m5_work_begin is called, after a parallel region in the program. It looks like doClone function in src/sim/syscall_emul.hh changes the processPtr  (ctc->setProcessPtr(cp); in line 1813) of the starting cores but the switch cores still point to the old process.

**Affects version**
I had encountered the error on v24.0.0.1 but the problem persists in v25.0.0.0 stable branch.

**gem5 Modifications**
No modifications

**To Reproduce**
1. Compile gem5.opt and m5 on X86 architecture, i.e.  `scons build/X86/gem5.opt -j8; scons build/x86/out/m5 -j8;`  
2. Adjust the Makefile in the attached example program for GEM5_HOME directory, compile it using `make` command.
3. Change the binary_path in the attached clone_switch_bug.py configuration to point to the example program.
4. cd into gem5 directory.
5. Using the attached python configuration script, run the simulation with:
```
shell ./build/X86/gem5.opt  -re ./configs/bug/clone_switch_bug.py --num_cores 4
```

**Terminal Output**
I am putting the assertion failure here again:

```shell
gem5.opt: src/cpu/thread_context.cc:254: void gem5::takeOverFrom(gem5::ThreadContext&, gem5::ThreadContext&): Assertion `ntc.getProcessPtr() == otc.getProcessPtr()' failed.
```

**Expected behavior**
No assertion failure, and the program is simulated for its ROI without any issues, e.g. switch cores should get set processPtr when the starting cores are being set, and vice versa.

**Host Operating System**
CentOS Linux 7 (Core)

**Host ISA**
X86

**Compiler used**
gcc version 10.3.0 (GCC) 
scons v3.1.0

**Additional information**
Check the example.cc for some comments on exactly how error is reproduced, --debug-flags=SyscallVerbose is also useful. Attached file:
[clone_switch_bug_minimal_example.zip](https://github.com/user-attachments/files/21228775/clone_switch_bug_minimal_example.zip)

alisemi:
When using SimpleSwitchableProcessor with multithreaded programs an assertion failure occurs in takeOverFrom() due to mismatched process pointers between active and switched-out CPUs.

The issue arises because clone() syscall creates new threads on active CPUs (starting cores) but doesn't update corresponding switched-out CPUs (switch cores) with the new process pointers. Later CPU switching triggers the assertion failure:
`ntc.getProcessPtr() == otc.getProcessPtr()' failed`

This patch synchronizes process pointers during clone operations by:
- Adding public getter methods to BaseCPU for thread context access
- Updating doClone() to set process pointers on switched-out CPUs with matching CPU IDs

Github Issue: https://github.com/gem5/gem5/issues/2446

### DIFF
--- a/src/sim/syscall_emul.hh
+++ b/src/sim/syscall_emul.hh
@@ -1832,8 +1832,7 @@ doClone(SyscallDesc *desc, ThreadContext *tc, RegVal flags, RegVal newStack,
     for (BaseCPU *cpu : BaseCPU::getCpuList()) {
         // Skip the current CPU and only consider switched-out CPUs with
         // matching ID
-        if (cpu != current_cpu &&
-            cpu->switchedOut() &&
+        if (cpu != current_cpu && cpu->switchedOut() &&
             cpu->cpuId() == current_cpu->cpuId()) {
 
             // Find the corresponding thread context on the switched-out CPU
@@ -1843,7 +1842,8 @@ doClone(SyscallDesc *desc, ThreadContext *tc, RegVal flags, RegVal newStack,
 
                 // Update the process pointer to match the active CPU
                 if (switched_tc) {
-                    DPRINTF(SyscallVerbose, "doClone: Updating switched-out "
+                    DPRINTF(SyscallVerbose,
+                            "doClone: Updating switched-out "
                             "CPU %d thread %d process pointer from %p to %p\n",
                             cpu->cpuId(), current_thread_id,
                             switched_tc->getProcessPtr(), cp);
--- a/src/sim/syscall_emul.hh
+++ b/src/sim/syscall_emul.hh
@@ -1827,25 +1827,25 @@ doClone(SyscallDesc *desc, ThreadContext *tc, RegVal flags, RegVal newStack,
     // takeOverFrom().
     BaseCPU *current_cpu = ctc->getCpuPtr();
     ThreadID current_thread_id = ctc->threadId();
-    
+
     // Iterate through all CPUs in the system to find switchable partners
     for (BaseCPU *cpu : BaseCPU::getCpuList()) {
         // Skip the current CPU and only consider switched-out CPUs with
         // matching ID
-        if (cpu != current_cpu && 
-            cpu->switchedOut() && 
+        if (cpu != current_cpu &&
+            cpu->switchedOut() &&
             cpu->cpuId() == current_cpu->cpuId()) {
-            
+
             // Find the corresponding thread context on the switched-out CPU
             if (current_thread_id < cpu->numThreadContexts()) {
-                ThreadContext *switched_tc = 
+                ThreadContext *switched_tc =
                     cpu->getThreadContext(current_thread_id);
-                
+
                 // Update the process pointer to match the active CPU
                 if (switched_tc) {
                     DPRINTF(SyscallVerbose, "doClone: Updating switched-out "
                             "CPU %d thread %d process pointer from %p to %p\n",
-                            cpu->cpuId(), current_thread_id, 
+                            cpu->cpuId(), current_thread_id,
                             switched_tc->getProcessPtr(), cp);
                     switched_tc->setProcessPtr(cp);
                 }
--- a/src/sim/syscall_emul.hh
+++ b/src/sim/syscall_emul.hh
@@ -1820,6 +1820,39 @@ doClone(SyscallDesc *desc, ThreadContext *tc, RegVal flags, RegVal newStack,
     cp->assignThreadContext(ctc->contextId());
     owner->revokeThreadContext(ctc->contextId());
 
+    // For switchable CPU configurations, we need to also set the process
+    // pointer on any switched-out CPUs that have the same CPU ID and thread
+    // context ID. This ensures that when CPU switching occurs, both CPUs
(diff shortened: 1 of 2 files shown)


powerjg (commented):
I think something went wrong in merging/rebasing. All I see in the diff is whitespace changes.

alisemi:
> I think something went wrong in merging/rebasing. All I see in the diff is whitespace changes.

I used the github feature to automaically rebase the branch, did not really notice that it added  then removing those whitespaces, though they seem to be ineffective on the code anyway. The final changes on the does not show such changes.

alisemi:
> > I think something went wrong in merging/rebasing. All I see in the diff is whitespace changes.
> 
> I used the github feature to automaically rebase the branch, did not really notice that it added then removing those whitespaces, though they seem to be ineffective on the code anyway. The final changes on the does not show such changes.

No, I realized that the whitespace fixes are coming from the "[pre-commit.ci] auto fixes from pre-commit.com hooks" commit earlier.

src/sim/syscall_emul.hh
```diff
@@ -1814,6 +1814,39 @@ doClone(SyscallDesc *desc, ThreadContext *tc, RegVal flags, RegVal newStack,
...
+                ThreadContext *switched_tc =
+                    cpu->getThreadContext(current_thread_id);
+
+                // Update the process pointer to match the active CPU
+                if (switched_tc) {
+                    DPRINTF(SyscallVerbose, "doClone: Updating switched-out "
+                            "CPU %d thread %d process pointer from %p to %p\n",
+                            cpu->cpuId(), current_thread_id,
+                            switched_tc->getProcessPtr(), cp);
+                    switched_tc->setProcessPtr(cp);
```
powerjg, line 1850: Shouldn't this be done in takeOverFrom on the CPU side? Why do this in the syscall implementation?

erin-le:
@alisemi: Hello, could you respond to the comment above when you get the chance? Thank you!

src/sim/syscall_emul.hh
alisemi: I remember trying to solve the problem by only modifying the takeOverFrom function as you suggested but it has been a while now that I don't recall why I could not do it. I will take a look it again soon.

alisemi:
> @alisemi: Hello, could you respond to the comment above when you get the chance? Thank you!

Hi, thanks! I've just come back from vacation, I will take a look at it.

src/sim/syscall_emul.hh
alisemi: @powerjg 
Hi, looking at the issue and the solution again, the main problem arises because the other cores start as idle, then assigned Processes and Process pointers. Here is what happens on a multi-core execution with switcable CPUS.
1- The execution starts as sequential on a single thread, only one of the CPUs (starting core) gets a Process pointer. The other CPUs are idle and have no Process.
2- The program enters into a parallel region, it eventually calls the doClone() function, cloning the Process into the other cores (but not to the switchable cores).
3- The ROI starts at a later place after the parallel region started, the switchable cores still don't have any process pointers, though their starting core counterparts have. This triggers the assertion failure in the takeOverFrom function at the thread context.

We can possibly add a statement like switched_tc->setProcessPtr(cp); before calling the takeOverFrom() of the ThreadContext within the BaseCPU's takeOverFrom() but then that will make the assertion inside the ThreadContext redundant. Do you think we should proceed in that way?

powerjg: Thanks for the explanation. I think your current solution is OK as it is.

powerjg (dismissed):

erin-le:
@alisemi: Hello, thanks for this PR! Could you run the following command and push the changes to get the `clang-format-check` CI test to pass? It's just the same command as what the test runs. Thanks!

```bash
python util/run-git-clang-format.py --ci-pr-base-commit 22e4e8331d62601d1cb14ada7eab2ad9f77fd639
```

alisemi:
> @alisemi: Hello, thanks for this PR! Could you run the following command and push the changes to get the `clang-format-check` CI test to pass? It's just the same command as what the test runs. Thanks!
> 
> ```shell
> python util/run-git-clang-format.py --ci-pr-base-commit 22e4e8331d62601d1cb14ada7eab2ad9f77fd639
> ```

Done, formatted two files now!

powerjg (approved):

BobbyRBruce (approved):
I think the tests will pass now. If so I'll merge and cherry-pick to the staging branch

PR #2459, opened 2025-07-19
stdlib: add Simulator API to schedule hypercall exits

erin-le:
This commit adds functions to Simulator to allow for m5.scheduleTickExitAbsolute and m5.scheduleTickExitFromCurrent to be used in a config script without importing m5. By extension, this allows for hypercall 6 to be trigged and hypercall scheduled exit handlers to be used without having to import m5 into config scripts.

### DIFF
--- a/src/python/gem5/simulate/simulator.py
+++ b/src/python/gem5/simulate/simulator.py
@@ -338,23 +338,9 @@ class Simulator:
             tick before exiting.
             exit_str (str): The reason for the exit.
         """
-        if self._max_ticks and self._max_ticks != m5.MaxTick:
-            warn(
-                "A classic MAX_TICKS exit has already been scheduled for "
-                f"tick {self._max_ticks}. Setting hypercall and classic "
-                "exits in the same simulation is not well tested and the "
-                "simulation may not behave as expected."
-            )
-        if exit_str == "Tick exit reached":
-            warn(
-                f"The exit string '{exit_str}' will cause the simulation "
-                "to exit with a classic ExitEvent.SCHEDULED_TICK exit event "
-                "instead of a hypercall."
-            )
-        m5.scheduleTickExitFromCurrent(ticks_from_current, exit_str)
-        self._hypercall_max_ticks = (
-            self.get_current_tick() + ticks_from_current
-        )
+        max_tick = self.get_current_tick() + ticks_from_current
+        self.set_hypercall_absolute_max_ticks(max_tick, exit_str)
+        self._hypercall_max_ticks = max_tick
 
     def schedule_simpoint(self, simpoint_start_insts: List[int]) -> None:
         """
--- a/src/python/gem5/simulate/simulator.py
+++ b/src/python/gem5/simulate/simulator.py
@@ -178,6 +178,7 @@ class Simulator:
                 "via it's `is_fullsystem` method."
             )
 
+        self._hypercall_max_ticks = None
         self.set_max_ticks(max_ticks)
 
         if id:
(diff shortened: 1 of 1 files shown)


src/python/gem5/simulate/simulator.py
```diff
@@ -282,12 +283,71 @@ def set_max_ticks(self, max_tick: int) -> None:
...
+            )
+
         self._max_ticks = max_tick
 
     def get_max_ticks(self) -> int:
         assert hasattr(self, "_max_ticks"), "Max ticks not set"
         return self._max_ticks
 
+    def set_hypercall_absolute_max_ticks(
+        self, max_tick: int, exit_str: str
```
powerjg: How about 
```suggestion
        self, max_tick: int, exit_str: str="Max ticks reached"
```

And you can drop the "This must be provided to exit with hypercall 6 exit."

In the long run, I'd like to deprecate the old way of doing it, and I think having a default string will make it easier.

Maybe also `warn` if the string is empty.

src/python/gem5/simulate/simulator.py
```diff
@@ -282,12 +283,71 @@ def set_max_ticks(self, max_tick: int) -> None:
...
+                "A classic MAX_TICKS exit has already been scheduled for "
+                f"tick {self._max_ticks}. Setting hypercall and classic "
+                "exits in the same simulation is not well tested and the "
+                "simulation may not behave as expected."
+            )
+        m5.scheduleTickExitAbsolute(max_tick, exit_str)
+        self._hypercall_max_ticks = max_tick
+
+    def set_hypercall_relative_max_ticks(
+        self, ticks_from_current: int, exit_str: str
```
powerjg: Ditto above

src/python/gem5/simulate/simulator.py
erin-le: I made the suggested changes and tested. If the user passes "" for `exit_str`, the function will still trigger hypercall 6. However, if they use "Tick exit reached", which is the default string in m5.scheduleTickExitAbsolute and m5.scheduleTickExitCurrent, they will trigger `ExitEvent.SCHEDULED_TICK`. As such, I've added a warning for if users accidentally use "Tick exit reached".

src/python/gem5/simulate/simulator.py
```diff
@@ -282,12 +283,79 @@ def set_max_ticks(self, max_tick: int) -> None:
...
+        if exit_str == "Tick exit reached":
+            warn(
+                f"The exit string '{exit_str}' will cause the simulation "
+                "to exit with a classic ExitEvent.SCHEDULED_TICK exit event "
+                "instead of a hypercall."
+            )
+        m5.scheduleTickExitAbsolute(max_tick, exit_str)
+        self._hypercall_max_ticks = max_tick
+
+    def set_hypercall_relative_max_ticks(
```
powerjg, line 328: Sorry one other thing... why don't you just call `set_hypercall_absolute_max_ticks(curTick() + ticks_from_current)` and remove the duplicated warnings.

src/python/gem5/simulate/simulator.py
```diff
@@ -282,12 +283,65 @@ def set_max_ticks(self, max_tick: int) -> None:
...
+        exit_handler.py` for details.
+
+        Args:
+            max_tick (int): The number of ticks to simulate from the current
+            tick before exiting.
+            exit_str (str): The reason for the exit.
+        """
+        max_tick = self.get_current_tick() + ticks_from_current
+        self.set_hypercall_absolute_max_ticks(max_tick, exit_str)
+        self._hypercall_max_ticks = max_tick
```
powerjg: This line seems redundant with line 326. Let me know if I'm missing something.

powerjg (approved):

PR #2466, opened 2025-07-23
arch-riscv: set 0 PMP entries as default

saul44203:
With #2376 a PMP check will now fail if there's at least one entry but no rules are implemented. 

Right now the default number of entries is 16, so this means that if the PMP is not to be used, the number of entries will need to get 0'd explicitly. This patch puts 0 as the default value to effectively disable the PMP by default.

### DIFF
--- a/src/arch/riscv/PMP.py
+++ b/src/arch/riscv/PMP.py
@@ -34,4 +34,4 @@ class PMP(SimObject):
     cxx_header = "arch/riscv/pmp.hh"
     cxx_class = "gem5::RiscvISA::PMP"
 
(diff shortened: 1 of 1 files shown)


powerjg (commented):
Wouldn't a better solution be to make sure the bootloader/OpenSBI sets the pmp_entries correctly?

Most RISC-V hardware has PMP entires. I don't love the idea of removing the PMP by default.

What's the problem that this PR is fixing?

erin-le:
@saul44203: Hello, could you respond to Jason's comments when you have time? Thank you!

saul44203:
Hi @erin-le, sorry for the delay. I think @powerjg is right and this might not be the best solution. Closing this. Thanks.

PR #2467, opened 2025-07-23
misc: define default hypercall ids in new header

Harshil2107:


powerjg (approved):
This needs to be documented on the website under m5_ops

PR #2478, opened 2025-07-25
arch-vega: Improve MFMA precision to match MI300X hardware

zevorn:
When comparing the precision of CPU FMA and MI300X MFMA instructions, we found that gem5’s results had 3~4 more ulp errors than the hardware.Because showed that gem5’s MFMA used separate multiply and add steps,
leading to lower precision. Using std::fma() instead improved the precision to match real hardware.


abmerop (approved):
Great contribution!  Thanks for finding and fixing this.

abmerop:
@mattsinc @mjkpolo For your awareness, merging

mattsinc (commented):
@abmerop : sorry, many deadlines this past week and didn't get to this until now.

@zevorn : thanks for the contribution!  I'd be interested in understanding what test(s) you used to identify this problem.

Regardless, appreciate the contribution!

@v-ramadas : FYI

PR #2481, opened 2025-07-28
python: Prevent SimObject from sharing reference of class member.

dl8sd11:
The SimObject clones a member if it is a SimObject or a vector
of SimObjects. However, if the member is an empty list. The
current implementation does not clone this member, causing the
instance sharing the reference of the class member.

It causes the following behavior:

a = Obj1()
a.foo.append(Obj2)

b = Obj1()
len(b.foo) == 1 # True

This commit handles the empty list case. Note that, it invoke
the value's constructor because the member could be a derived
class of a list (e.g. VectorParamValue).

### DIFF
--- a/src/python/m5/SimObject.py
+++ b/src/python/m5/SimObject.py
@@ -777,6 +777,9 @@ class SimObject(metaclass=MetaSimObject):
         self._hr_values = multidict(ancestor._hr_values)
         # clone SimObject-valued parameters
         for key, val in ancestor._values.items():
+            if val == []:
+                self._values[key] = type(val)()
+                continue
             val = tryAsSimObjectOrVector(val)
             if val is not None:
                 self._values[key] = val(_memo=memo_dict)


powerjg:
This seems reasonable, but it's a bit intricate. Could you provide a full script that I can execute where the incorrect behavior happens on the current mainline and then this PR fixes the behavior. Please specify what the correct and incorrect behavior is.

dl8sd11:
Thanks for the suggestion, here is the reproducing script:
(I also comment the root cause of this issue in the comments)

```
from m5.objects import AddrRange
from m5.objects import System

s_1 = System()
s_2 = System()

s_1.mem_ranges.append(AddrRange(start=0, size=0)) 

print(s_1.mem_ranges) # s_1.mem_ranges is System.mem_ranges

s_2.mem_ranges.append(AddrRange(start=0, size=0)) # s_2.mem_ranges is cloned from System.mem_ranges which already has one element

print(s_2.mem_ranges)
```

Result:

```

gem5 version DEVELOP-FOR-25.1
gem5 compiled Jul 29 2025 03:43:00
gem5 started Jul 29 2025 03:47:09
gem5 executing on gcchen.c.googlers.com, pid 74568
command line: ./build/X86/gem5.opt break.py

[<m5.params.AddrRange object at 0x7f53444fd850>]
[<m5.params.AddrRange object at 0x7f53444fd850>, <m5.params.AddrRange object at 0x7f53444fd940>]
```

Expected:

`s_2.mem_ranges` should not contain two elements.

powerjg (commented):
I understand the problem you're describing now, and I see how this gets around that problem.

However, I don't know if this really fixes things. The issue is that `.append` doesn't work with SimObjects in general. In this case, the `.append` is actually modifying the *class* member (not the *instance* member of `s_1`. So, this solution would break down if the SimObject's parameter had a non-empty list as the default (which is allowed, but may not be used in practice). It also doesn't work for anything that's not a *parameter* since it wouldn't appear in the `ancestor.items()`. (This problem doesn't exactly exist for non-parameter members of a SimObject instance, but the `.append` function is also messed up there.)

I don't know what I would prefer around this. I think this PR changes the behavior, but it isn't going to be a python-consistent behavior with or without this PR. I don't really have anything actionable to change... are there other opinions in the community?

dl8sd11:
To clarify, this change would make the behavior of SimObject's VectorParam.SimObject consistent between empty list and non-empty list.


Before the change:

- empty VectorParam.SimObject: reuse the class member (the bug I mentioned earlier)
- non-empty VectorParam.SimObject: clone the class member

After  the change:

- empty VectorParam.SimObject: clone the class member
- non-empty VectorParam.SimObject: clone the class member

```
# Make I2CBus.devices has 1 value by default 
class I2CBus(BasicPioDevice):
    type = "I2CBus"
    cxx_header = "dev/i2c/bus.hh"
    cxx_class = "gem5::I2CBus"
    devices = VectorParam.I2CDevice([
      I2CDevice(),
    ], "Devices")


from m5.objects import I2CBus
from m5.objects import I2CDevice

s_1 = I2CBus()
s_2 = I2CBus()

s_1.devices.append(I2CDevice())

print(s_1.devices) # 2 element - before & after the change

s_2.devices.append(I2CDevice())

print(s_2.devices) # 2 element - before & after the change
```

However, as you point out, there are other cases with undesire behaviors (including, non-param, non-simojbect param) that this change does not handle. This change only handles the empty case which is what our team faced recently.

Another possibility is to copy the SimObject member when new SimObject is created.

powerjg (approved):
Thanks for answering my questions. This improves things, and even if it's not perfect, I think it's good to get it in.

PR #2484, opened 2025-07-28
misc: Update stats dump to have a optional message

Harshil2107:
In the current implementation, we call `stats.dump` as `m5.stats.dump()` and this call generated a stats.txt file that starts as the following:

```
---------- Begin Simulation Statistics ----------
simSeconds                                   9.915965                       # Number of seconds simulated (Second)
simTicks                                 9915965318458                       # Number of ticks simulated (Tick)
```

With this change the function call would look like `m5.stats.dump(message = "Dumping stats at end of ROI")` and the stats.txt output would then start with

```
---------- Begin Simulation Statistics : Dumping stats at end of ROI ----------
simSeconds                                   9.915965                       # Number of seconds simulated (Second)
simTicks                                 9915965318458                       # Number of ticks simulated (Tick)
```

### DIFF
--- a/src/base/stats/hdf5.cc
+++ b/src/base/stats/hdf5.cc
@@ -79,7 +79,7 @@ Hdf5::~Hdf5()
 }
 
 void
-Hdf5::begin(const std::string& message)
+Hdf5::begin(const std::string &message)
 {
     // HDF5 format ignores the message parameter
     // as it doesn't support text messages
--- a/src/base/stats/hdf5.cc
+++ b/src/base/stats/hdf5.cc
@@ -78,7 +78,6 @@ Hdf5::~Hdf5()
 {
 }
 
-
 void
 Hdf5::begin(const std::string& message)
 {
--- a/src/base/stats/hdf5.cc
+++ b/src/base/stats/hdf5.cc
@@ -80,8 +80,10 @@ Hdf5::~Hdf5()
 
 
 void
-Hdf5::begin()
+Hdf5::begin(const std::string& message)
 {
+    // HDF5 format ignores the message parameter
+    // as it doesn't support text messages
     h5File = H5::H5File(fname,
                         // Truncate the file if this is the first dump
                         dumpCount > 0 ? H5F_ACC_RDWR : H5F_ACC_TRUNC);
--- a/src/python/m5/stats/__init__.py
+++ b/src/python/m5/stats/__init__.py
@@ -388,8 +388,13 @@ lastDump = 0
 global_dump_roots = []
 
 
-def dump(roots=None):
-    """Dump all statistics data to the registered outputs"""
+def dump(roots=None, message=""):
+    """Dump all statistics data to the registered outputs
+
+    Args:
+        roots: Optional list of roots to dump
+        message: Optional message to include in text output headers
+    """
 
     all_roots = []
     if roots is not None:
@@ -425,13 +430,21 @@ def dump(roots=None):
                 output.dump(all_roots)
         else:
             if output.valid():
-                output.begin()
+                output.begin(message)
                 _dump_to_visitor(output, roots=all_roots)
                 output.end()
 
 
-def reset():
-    """Reset all statistics to the base state"""
+def reset(message=""):
+    """Reset all statistics to the base state
+
+    Args:
+        message: Optional message to output when resetting stats
+    """
+
+    for output in outputList:
+        if not isinstance(output, JsonOutputVistor):
+            output.printResetMessage(message)
 
     # call reset stats on all SimObjects
     root = Root.getInstance()
--- a/src/python/m5/stats/__init__.py
+++ b/src/python/m5/stats/__init__.py
@@ -435,16 +435,8 @@ def dump(roots=None, message=""):
                 output.end()
 
 
-def reset(message=""):
-    """Reset all statistics to the base state
-
-    Args:
-        message: Optional message to output when resetting stats
-    """
-
-    for output in outputList:
-        if not isinstance(output, JsonOutputVistor):
-            output.printResetMessage(message)
+def reset():
+    """Reset all statistics to the base state"""
 
     # call reset stats on all SimObjects
     root = Root.getInstance()
--- a/src/python/pybind11/stats.cc
+++ b/src/python/pybind11/stats.cc
@@ -134,8 +134,6 @@ pybind_init_stats(py::module_ &m_native)
         .def("valid", &statistics::Output::valid)
         .def("beginGroup", &statistics::Output::beginGroup)
         .def("endGroup", &statistics::Output::endGroup)
-        .def("printResetMessage",
-        &statistics::Output::printResetMessage, py::arg("message") = "")
         ;
 
     py::class_<statistics::Info,
--- a/src/python/pybind11/stats.cc
+++ b/src/python/pybind11/stats.cc
@@ -133,7 +133,8 @@ pybind_init_stats(py::module_ &m_native)
         .def("end", &statistics::Output::end)
         .def("valid", &statistics::Output::valid)
         .def("beginGroup", &statistics::Output::beginGroup)
-        .def("endGroup", &statistics::Output::endGroup);
+        .def("endGroup", &statistics::Output::endGroup)
+        ;
 
     py::class_<statistics::Info,
         std::unique_ptr<statistics::Info, py::nodelete>>(m, "Info")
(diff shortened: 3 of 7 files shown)


powerjg (dismissed):
Cool feature! It's going to break a bunch of external scripts if we ever have a nonempty message by default, though!

erin-le (dismissed):
Hey @Harshil2107, I tested this change with modified `arm-ubuntu-run-with-kvm.py` scripts and got this error message at the first call to m5.stats.dump():

```
src/dev/arm/rv_ctrl.cc:198: warn: Tried to write RVIO at offset 0xa8 (data 0) that doesn't exist
src/dev/arm/rv_ctrl.cc:198: warn: Tried to write RVIO at offset 0xa8 (data 0) that doesn't exist
src/dev/arm/rv_ctrl.cc:198: warn: Tried to write RVIO at offset 0xa8 (data 0) that doesn't exist
AttributeError: 'str' object has no attribute 'path_list'

At:
  src/python/m5/stats/__init__.py(372): _dump_to_visitor
  src/python/m5/stats/__init__.py(434): dump
  arm-ubuntu-run-with-kvm-hypercall.py(133): _process
  src/python/gem5/simulate/exit_handler.py(106): handle
  src/python/gem5/simulate/simulator.py(638): run
  arm-ubuntu-run-with-kvm-hypercall.py(169): <module>
  src/python/m5/main.py(682): main
```

Can you double check if you've pushed everything? Also, the config scripts I used to test this are attached below. Can you let me know if there's something wrong with the setup?
[arm-ubuntu-run-with-kvm-classic-exit.txt](https://github.com/user-attachments/files/21478804/arm-ubuntu-run-with-kvm-classic-exit.txt)
[arm-ubuntu-run-with-kvm-hypercall.txt](https://github.com/user-attachments/files/21478805/arm-ubuntu-run-with-kvm-hypercall.txt)

Harshil2107:
> Hey @Harshil2107, I tested this change with modified `arm-ubuntu-run-with-kvm.py` scripts and got this error message at the first call to m5.stats.dump():
> 
> ```
> src/dev/arm/rv_ctrl.cc:198: warn: Tried to write RVIO at offset 0xa8 (data 0) that doesn't exist
> src/dev/arm/rv_ctrl.cc:198: warn: Tried to write RVIO at offset 0xa8 (data 0) that doesn't exist
> src/dev/arm/rv_ctrl.cc:198: warn: Tried to write RVIO at offset 0xa8 (data 0) that doesn't exist
> AttributeError: 'str' object has no attribute 'path_list'
> 
> At:
>   src/python/m5/stats/__init__.py(372): _dump_to_visitor
>   src/python/m5/stats/__init__.py(434): dump
>   arm-ubuntu-run-with-kvm-hypercall.py(133): _process
>   src/python/gem5/simulate/exit_handler.py(106): handle
>   src/python/gem5/simulate/simulator.py(638): run
>   arm-ubuntu-run-with-kvm-hypercall.py(169): <module>
>   src/python/m5/main.py(682): main
> ```
> 
> Can you double check if you've pushed everything? Also, the config scripts I used to test this are attached below. Can you let me know if there's something wrong with the setup? [arm-ubuntu-run-with-kvm-classic-exit.txt](https://github.com/user-attachments/files/21478804/arm-ubuntu-run-with-kvm-classic-exit.txt) [arm-ubuntu-run-with-kvm-hypercall.txt](https://github.com/user-attachments/files/21478805/arm-ubuntu-run-with-kvm-hypercall.txt)

Thanks for testing this! For some reason, the attached scripts aren’t accessible when I try to open them, but based on the error, I suspect the issue is related to how `m5.stats.dump()` is being called.

If you're calling it like this:

```python
m5.stats.dump("message")
```

this will result in an error because the `dump()` function already has a parameter called `roots`, and by passing a string as the first argument, it's being incorrectly assigned to `roots`. That’s why you’re seeing an `AttributeError: 'str' object has no attribute 'path_list'`.

The correct way to pass a message is to use the named parameter:

```python
m5.stats.dump(message="your message here")
```

As shown in the diff, the function signature was changed from:

```python
def dump(roots=None):
```

to:

```python
def dump(roots=None, message=""):
```

So calling `dump()` with a string directly will not work. Let me know if updating the call fixes the issue!

src/python/m5/stats/__init__.py
```diff
@@ -425,13 +430,21 @@ def dump(roots=None):
                 output.dump(all_roots)
         else:
             if output.valid():
-                output.begin()
+                output.begin(message)
```
giactra, line 433: Can we avoid having the "being" method with an empty string as default param? If begin is not the direct API for stats processing, message here will already contain an empty string in case no message is forwarded (so being could require a mandatory message)

giactra (commented):
Hi @Harshil2107 , first of all many thanks for this feature. I have always found it confusing when the stat file contained multiple dumps, and you had to understand which dump relates to which segment of your guest code.

I was thinking a solution could be to dump start and end tick time, but this is a more general approach as you can either dump tick pairs or you could dump a more human readable message as you are doing. Nice!

I have a minor comment below (coding perspective) and a more philosophical one.

At the moment the underlying assumption has always been the stats.dump command prints to the stats file, and the stats.reset does not touch the file as it simply resets values in gem5.

If we introduce the message printing to the stats.reset command we inevitably break this assumption. I have no strong feeling against this; a part of me thinks this won't be a problem; another one thinks we could maybe have a cleaner API set by adding a different stats.message API which could be called before the stats.reset to explicitly print something, and leave the reset as it is....

What do you think?

powerjg:
> different stats.message API which could be called before the stats.reset to explicitly print something

This is a good idea. It needs to be compatible with more than just text output, though (e.g., I'm thinking of hdf5, etc.). Let me think about this a bit before we spend too much time on it.

There's three options
1. Drop the reset message for now and make a new PR with the message API (after some deeper thought)
2. Keep the reset message for now (we are talking about develop, not a release) and update it with the message API later
3. Hold off on committing this PR

giactra:
> 
> There's three options
> 
>     1. Drop the reset message for now and make a new PR with the message API (after some deeper thought)
> 
>     2. Keep the reset message for now (we are talking about develop, not a release) and update it with the message API later
> 
>     3. Hold off on committing this PR

I'd be happy to go for Option 1 which is the right option in the middle IMHO

src/python/m5/stats/__init__.py
Harshil2107: I have updated the begin function so that now the message arg is required.

Harshil2107:
I have removed the changes made to stats reset.

src/python/pybind11/stats.cc
```diff
@@ -129,7 +129,7 @@ pybind_init_stats(py::module_ &m_native)
         ;
 
     py::class_<statistics::Output>(m, "Output")
-        .def("begin", &statistics::Output::begin)
+        .def("begin", &statistics::Output::begin, py::arg("message") = "")
```
giactra: This does not need a default value anymore

src/python/pybind11/stats.cc
```diff
@@ -129,7 +129,7 @@ pybind_init_stats(py::module_ &m_native)
         ;
 
     py::class_<statistics::Output>(m, "Output")
-        .def("begin", &statistics::Output::begin)
+        .def("begin", &statistics::Output::begin, py::arg("message"))
```
giactra: I a not a pybind11 expert but I don't think py::arg is needed anymore now that begin does not have default arguments

giactra (commented):
Hi @Harshil2107 apologies for the late review. I only have a minor comment

erin-le:
I tested this after removing py::arg and gem5 built and ran correctly.

powerjg (dismissed):
Looks like there may be formatting errors. Also, it looks like now there's a file with only a style update. That should probably be dropped.

src/python/pybind11/stats.cc
```diff
@@ -133,8 +133,7 @@ pybind_init_stats(py::module_ &m_native)
         .def("end", &statistics::Output::end)
         .def("valid", &statistics::Output::valid)
         .def("beginGroup", &statistics::Output::beginGroup)
-        .def("endGroup", &statistics::Output::endGroup)
-        ;
+        .def("endGroup", &statistics::Output::endGroup);
```
erin-le:
```suggestion
        .def("endGroup", &statistics::Output::endGroup)
        ;
```

src/base/stats/hdf5.cc
```diff
@@ -80,8 +80,10 @@ Hdf5::~Hdf5()
```
erin-le:
```suggestion

```

powerjg (approved):
I believe we want to squash and merge. Thanks for the updates.

PR #2485, opened 2025-07-28
arch-arm: Fix inverted RAO check of HCPTR.TCP10

arichardson:
The documentation states this field is read-only ones if NSACR.CP10 is set to zero, not when it is set to 1. This logic error probably happened because The values of the CP10 fields in these registers are inverted: NSACR.CP10=1 means "Advanced SIMD and floating-point features can be accessed from both Security states." whereas HCPTR.TCP10=1 means "Any attempted access to Advanced SIMD and floating-point functionality from Non-secure state is trapped to Hyp mode"

### DIFF
--- a/src/arch/arm/insts/static_inst.cc
+++ b/src/arch/arm/insts/static_inst.cc
@@ -752,17 +752,17 @@ ArmStaticInst::checkAdvSIMDOrFPEnabled32(ThreadContext *tc,
 
     if (have_virtualization && !is_secure) {
         HCPTR hcptr = tc->readMiscReg(MISCREG_HCPTR);
-        bool hcptr_cp10 = hcptr.tcp10;
+        bool hcptr_tcp10 = hcptr.tcp10;
         bool hcptr_tase = hcptr.tase;
 
         if (have_security && !ELIs64(tc, EL3) && !is_secure) {
             if (nsacr.nsasedis)
                 hcptr_tase = true;
-            if (nsacr.cp10)
-                hcptr_cp10 = true;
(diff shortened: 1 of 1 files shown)


src/arch/arm/insts/static_inst.cc
```diff
@@ -752,17 +752,17 @@ ArmStaticInst::checkAdvSIMDOrFPEnabled32(ThreadContext *tc,
...
     if (have_virtualization && !is_secure) {
         HCPTR hcptr = tc->readMiscReg(MISCREG_HCPTR);
-        bool hcptr_cp10 = hcptr.tcp10;
+        bool hcptr_tcp10 = hcptr.tcp10;
         bool hcptr_tase = hcptr.tase;
 
         if (have_security && !ELIs64(tc, EL3) && !is_secure) {
             if (nsacr.nsasedis)
                 hcptr_tase = true;
-            if (nsacr.cp10)
```
arichardson, line 761: I feel like this RAO/RAZ handling should really be done inside readMiscReg() in the same way other registers handle it but for now I just submitted the minimal fix.

powerjg (approved):

src/arch/arm/insts/static_inst.cc
giactra: Understand your point, on the other hand, this logic is closely matching the pseudocode in the arm architectural reference manual.. Lots of time we have sacrificed the best implementation on gem5 for the sake of being amenable to debug/extend whenever extensions are added

giactra (approved):
Thanks for spotting this! That's awesome!

PR #2494, opened 2025-07-29
misc: split single line if statements and add braces in .clang-format

erin-le:
This commit modifies .clang-format to split single line if statements across multiple lines. It also modifies .clang-format to require curly braces around control statements (if, else, for, do, while) unless inside macro definitions or if braces would enclose preprocessor directives.

Before:

```cpp
if (true) std::cout << "Hello world!" << std::endl;
```

After:

```cpp
if (true) {
    std::cout << "Hello world!" << std::endl;
}
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -63,10 +63,15 @@ SpaceInEmptyParentheses: false               # No f( ) -> f() (sec. Spacing)
 SpacesInAngles: false                        # No < T > in templates
 
 # Single‐line constructs
-AllowShortIfStatementsOnASingleLine: true
+AllowShortIfStatementsOnASingleLine: false
 AllowShortBlocksOnASingleLine: true
 AllowShortLoopsOnASingleLine: true
 AllowShortFunctionsOnASingleLine: InlineOnly
+
+# Require curly braces after control statements (if, else, for, do, while)
+# unless inside macro definitions or if the braces would enclose preprocessor
+# directives
+InsertBraces: true
 
 # Return‐type / function‐name separation
 AlwaysBreakAfterReturnType: All              # Return on its own line (sec. Functions)


powerjg:
Is it possible to allow the following

```
if (true) std::cout << "Hello world!" << std::endl;
```

But disallow

```
if (true) 
    std::cout << "Hello world!" << std::endl;
```

.clang-format
```diff
@@ -63,11 +63,14 @@ SpaceInEmptyParentheses: false               # No f( ) -> f() (sec. Spacing)
 SpacesInAngles: false                        # No < T > in templates
 
 # Single‐line constructs
-AllowShortIfStatementsOnASingleLine: true
+AllowShortIfStatementsOnASingleLine: false
 AllowShortBlocksOnASingleLine: true
 AllowShortLoopsOnASingleLine: true
 AllowShortFunctionsOnASingleLine: InlineOnly
 
+# Require curly braces
```
powerjg: What does this require curly braces on? If/while? what about case statements?

erin-le:
> Is it possible to allow the following
> 
> ```
> if (true) std::cout << "Hello world!" << std::endl;
> ```
> 
> But disallow
> 
> ```
> if (true) 
>     std::cout << "Hello world!" << std::endl;
> ```

It looks like the clang format file on the `develop` branch already disallows
```
> if (true) 
>     std::cout << "Hello world!" << std::endl;
```
as it is reformatted to one line, `if (true) std::cout << "Hello world!" << std::endl;`. However, the `develop` clang format will also reformat 

```
if (true) {
    std::cout << "Hello world!" << std::endl;
}
```
to
```
if (true) { std::cout << "Hello world!" << std::endl; }
```
which may be undesirable. I looked through the documentation, asked ChatGPT and Claude, and tested the formatting with different options, and I'm not sure if it's possible to enforce
```
if (true) {
    std::cout << "Hello world!" << std::endl;
}
```
while also allowing single line if statements with no brackets, e.g. `if (true) std::cout << "Hello world!" << std::endl;`, though it's quite possible that I'm missing something.

.clang-format
erin-le: According to [the documentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html), this should require braces "after control statements (if, else, for, do, and while) in C++ unless the control statements are inside macro definitions or the braces would enclose preprocessor directives."

giactra:
It has to be noted that InsertBraces is considered a risky option by clang-format doc: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces


> 
> Warning
> 
> Setting this option to true could lead to incorrect code formatting due to clang-format’s lack of complete semantic information. As such, extra care should be taken to review code changes made by this option.

powerjg:
If I understand correctly, this warning is essentially because not having curly braces is unclear. I agree we would need to take a lot of care if/when we ever apply this to the whole codebase, but if we're only applying it for new code, this shouldn't be a big issue. Thoughts?

powerjg:
@erin-le I would be happy to always have multi-line statements with curly braces and disallow single line statements as long as that's good with others. (In fact, this is my preference.)

giactra:
> @erin-le I would be happy to always have multi-line statements with curly braces and disallow single line statements as long as that's good with others. (In fact, this is my preference.)

I was thinking the same. If someone really wants a compact one line if statement (to make a chain of ifs more tabular) can always disable clang format locally (with "clang-format off" annotations)

BobbyRBruce (approved):

rjc-arch:
Just to note that the `InsertBraces` directive requires `clang-format-15` or greater [1], a requirement which is not met by the default version on older releases of Ubuntu. Do we still support Ubuntu 22.04 and 20.04? If so, this PR doesn't work on some supported platforms.

Unfortunately, the resulting error also blocks `pre-commit` from passing. We could update `run-git-clang-format.py` to also ignore this type of error, but that would silently disable automatic formatting (which we have already discussed as undesirable in other threads). 

As a work-around, `git-clang-format-15` can be configured on Ubuntu 22.04 using `update-alternatives`:

```
# Install clang-format-15:
sudo apt install clang-format-15

# Configure clang-format-15 as the system default:
sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-15 150
sudo update-alternatives --install /usr/bin/git-clang-format git-clang-format /usr/bin/git-clang-format-15 150

# Optionally install other alternatives as above, and choose the default alternative:
sudo update-alternatives --config clang-format 
sudo update-alternatives --config git-clang-format 
```

Ubuntu 20.04 has a package for `clang-format-18` which can be configured in a similar manner.

[1] https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces

PR #2497, opened 2025-07-31
arch-riscv: Fix CSR writing mask

rogerchang23424:
arch-riscv: Fix CSR writing mask

Change in the PR:
1. Implmement RV32 CSR mask
2. Add LOCAL_MASK write mask for mip/sip
3. Remove STATUS_TVM mask in M, MU privilege mode set
4. Remove mip/sip/vsip/hip writing mask, they read/write mask for these CSRs are same

Change-Id: I805982eceea3bb4329a8161300db1a24aad94c69


powerjg (commented):
lgtm, and I'm happy to approve as-is.

However... if you want to take a stab at changing the .clang-format file to make this more legible (or even using `// clang format off` (or whatever it is)) that would be good too. the clang format is making this less readable, IMO

powerjg (approved):

PR #2499, opened 2025-07-31
cpu: Workaround missing IsUnconditional flag

### LINKED ISSUE #2463: BranchPredictor: predict a instruction as not taken, but may still provide a target and change the next pc
**Describe the bug**
When the branch predictior makes predictions and predicts the instruction "not taken", it will still access modules such as BTB, RAS and indirect predictor to obtain its target. 
In this case, branch predictior will give a wrong infomation to CPU: a instruction not taken but its next pc not is not consecutive.

Code in function _predict()_, link: [https://github.com/gem5/gem5/blob/develop/src/cpu/pred/bpred_unit.cc](url)

I have checked the function _predict()_ in version gem5 v23.0.1.0, the logic is different and correct.

I have also checked code of O3 CPU, when O3 CPU got wrong infomation from BPredUnit: _this_pc_ will may be changed to a discontinuous pc and _predictedBranch_ is also false.But this won't cause the CPU to crash, because of _fetchBufferBlockPC_ is changed according to _this_pc_. CPU thought PC has moved to next fetchBufferBlock, then stop fetch. This precisely masked the aforementioned problem, which led to it never being detected.

This piece of code is in fuction _fetch()_ of src/cpu/o3/fetch.cc, as follow:
```
// If buffer is no longer valid or fetchAddr has moved to point
// to the next cache block then start fetch from icache.
if (!fetchBufferValid[tid] ||
	fetchBufferBlockPC != fetchBufferPC[tid])
	break;
```

**Affects version**
All version after gem5 v23.1, commit hash code is bae34876780dfb2bc22b9151bfda1d39ee80cfb1

**gem5 Modifications**
I discovered this bug while using a custom CPU and custom workload.

**Expected behavior**
When instruction is predicted as not taken, branch predictor should no longer try to find target.

**Host Operating System**
Ubuntu 22.04

**Host ISA**
X86

**Additional information**
A simple solution in fuction _predict()_ of src/cpu/pred/bpred_unit.cc as follow:
```
    /** ----------------------------------------------
     * Fallthrough
     * -----------------------------------------------
     * All the target predictors did their job.
     * If there is no target its either not taken or
     * a BTB miss. In that case we just fallthrough.
     * */
    if (!hist->predTaken ||
    	hist->targetProvider == TargetProvider::NoTarget) {
    	set(hist->target, pc);
    	inst->advancePC(*hist->target);
    	hist->predTaken = false;
    }
```

but in this solution, we still looked for target. May a better solution as this:
```
    // The direction is done now get the target address
    // from BTB, RAS or indirect predictor.
    hist->targetProvider = TargetProvider::NoTarget;
	
    if (hist->predTaken) {
        ///////// Find target
    }

    /** ----------------------------------------------
     * Fallthrough
     * -----------------------------------------------
     * All the target predictors did their job.
     * If there is no target its either not taken or
     * a BTB miss. In that case we just fallthrough.
     * */
    if (hist->targetProvider == TargetProvider::NoTarget) {
        set(hist->target, pc);
        inst->advancePC(*hist->target);
        hist->predTaken = false;
    }
    stats.targetProvider[tid][hist->targetProvider]++;

    // The actual prediction is done.
    // For now the BPU assume its correct. The update
    // functions will correct the branch if needed.
    // If prediction and actual direction are the same
    // at commit the prediction was correct.
    hist->actuallyTaken = hist->predTaken;
    set(pc, *hist->target);

    DPRINTF(Branch, "%s(tid:%i, sn:%i, PC:%#x, %s) -> taken:%i, target:%s "
            "provider:%s\n", __func__, tid, seqNum, hist->pc,
            toString(brType), hist->predTaken, *hist->target,
            enums::TargetProviderStrings[hist->targetProvider]);
```

dhschall:
Some branch instructions do not set the `IsUnconditional` flag appropriately
when setting flags like `IsCall` or `IsReturn`. For example `sret` or `mret` in
RISC-V but also some instructions in ARM.
This causes the branch predictor to incorrectly think its an conditional branch
while still doing the RAS access. (See #2463)
We assume now all instructions are unconditional except the `IsConditional`
flag is explicity set.

Fixes #2463

### DIFF
--- a/src/cpu/pred/bpred_unit.cc
+++ b/src/cpu/pred/bpred_unit.cc
@@ -145,7 +145,7 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
      * the direction is always taken
      */
 
-    if (inst->isUncondCtrl()) {
+    if (hist->uncond) {
         // Unconditional branches -----
         hist->condPred = true;
     } else {
--- a/src/cpu/pred/bpred_unit.hh
+++ b/src/cpu/pred/bpred_unit.hh
@@ -292,7 +292,7 @@ class BPredUnit : public SimObject
                          const StaticInstPtr & inst)
             : seqNum(sn), tid(_tid), pc(_pc),
               inst(inst), type(getBranchType(inst)),
-              call(inst->isCall()), uncond(inst->isUncondCtrl()),
+              call(inst->isCall()), uncond(!inst->isCondCtrl()),
               predTaken(false), actuallyTaken(false), condPred(false),
               btbHit(false), targetProvider(TargetProvider::NoTarget),
               resteered(false), mispredict(false), target(nullptr),


powerjg (approved):

src/cpu/pred/bpred_unit.cc
```diff
@@ -145,7 +145,7 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
      * the direction is always taken
      */
 
-    if (inst->isUncondCtrl()) {
+    if (hist->uncond) {
```
giactra, line 148: This seems like a reasonable solution.

One comment though:

Could we add an assertion here to check the branch type? Something like

`assert(inst->isUncondCtrl || inst->isCall() || inst->isRet() ...etc);`

In this way not only we add a sanity check; it is also more explicit for a reader which kind of control transfer instruction fall inside this cathegory

PR #2500, opened 2025-07-31
cpu-o3: remove inactive tail bytes from memory request

saul44203:
Currently some LSQ request types in O3 like SplitDataRequest can have various block aligned (sub)requests for the full line size even though their enablement mask has an inactive (0'd) tail. This patch checks for this pattern to reduce the request size only to the first n contiguous (enabled or disabled) bytes until the last enabled one.

### DIFF
--- a/src/cpu/utils.hh
+++ b/src/cpu/utils.hh
@@ -94,6 +94,27 @@ isAnyActiveElement(const std::vector<bool>::const_iterator& it_start,
     return (it_tmp != it_end);
 }
 
+/**
+ * Get size of inactive tail in an enablement range (0 if none).
+ */
+inline unsigned
+inactiveTailSize(const std::vector<bool>::const_iterator& it_start,
+                 const std::vector<bool>::const_iterator& it_end)
+{
+    std::vector<bool>::const_reverse_iterator rit_start(it_end);
+    std::vector<bool>::const_reverse_iterator rit_end(it_start);
+
+    unsigned count = 0;
+    for (auto it = rit_start; it != rit_end; ++it) {
+        if (!(*it)) {
+            count++;
+        } else {
+            break;
+        }
+    }
+    return count;
+}
+
 } // namespace gem5
 
 #endif // __CPU_UTILS_HH__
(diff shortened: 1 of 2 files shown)


src/cpu/utils.hh
```diff
@@ -94,6 +94,22 @@ isAnyActiveElement(const std::vector<bool>::const_iterator& it_start,
...
+inline unsigned
+inactiveTailSize(const std::vector<bool>::const_iterator& it_start,
+                 const std::vector<bool>::const_iterator& it_end)
+{
+    std::vector<bool>::const_reverse_iterator rit_start(it_end);
+    std::vector<bool>::const_reverse_iterator rit_end(it_start);
+
+    auto it_tmp = rit_start;
+    unsigned count = 0;
+    for (;it_tmp != rit_end && !(*it_tmp); ++it_tmp, count++);
```
powerjg: haha, whoa. I've got to ask you to re-write this line to be readable as a human. This reminds me of the [obfuscated c competition](https://en.wikipedia.org/wiki/International_Obfuscated_C_Code_Contest)

powerjg (commented):
Seems like a good fix, one small thing below.

src/cpu/utils.hh
```diff
@@ -94,6 +94,22 @@ isAnyActiveElement(const std::vector<bool>::const_iterator& it_start,
...
+    return count;
+}
+
```
powerjg: Here's what gemini says will work


```suggestion
/**
 * Get size of inactive tail in an enablement range (0 if none).
 */
inline unsigned
inactiveTailSize(const std::vector<bool>::const_iterator& it_start,
                 const std::vector<bool>::const_iterator& it_end)
{
    // Create reverse iterators to traverse the range backwards.
    std::vector<bool>::const_reverse_iterator rit_start(it_end);
    std::vector<bool>::const_reverse_iterator rit_end(it_start);

    unsigned count = 0;
    // Iterate from the back of the range to the front.
    for (auto it = rit_start; it != rit_end; ++it) {
        // If the current element is 'false' (inactive), increment the count.
        if (!(*it)) {
            count++;
        } else {
            // If we find a 'true' (active) element, stop counting.
            break;
        }
    }

    return count;
}

```

saul44203: Thanks. Pushed this code instead.

powerjg (approved):
Thanks for the update

PR #2507, opened 2025-08-04
mem: refactor Bridge to allow derived class

clemdiep:
This commit splits Bridge into BridgeBase and Bridge. BridgeBase contains the implementation to handle packets passing through a bridge, like Bridge before. But instead of relying on a fixed set of address ranges, it allows other class to derived from it to implement the functions getAddrRanges() and recvRangeChange() and so the set of address ranges can be either fixed or dynamic depending on the derived class implementation.

The functions canPassThrough() and setRejectedData() can also be implemented by derived classes, allowing to block some ranges and directly respond to them from the bridge. This is usefull to let a bridge respond with an error code when no device is responsible for a given range behind the bridge.

The *new* Bridge class is a simple class derived from BridgeBase with a fixed address ranges and, in the end, work like the current one.

This PR is the result of modification made base on this thread https://github.com/gem5/gem5/pull/2298#discussion_r2190569552 as bridge with dynamic ranges is needed for PCI implementation.

Made in collaboration with @marinazapater

### DIFF
--- a/src/mem/bridge.cc
+++ b/src/mem/bridge.cc
@@ -53,23 +53,19 @@
 namespace gem5
 {
 
-Bridge::BridgeResponsePort::BridgeResponsePort(const std::string& _name,
-                                         Bridge& _bridge,
-                                         BridgeRequestPort& _memSidePort,
-                                         Cycles _delay, int _resp_limit,
-                                         std::vector<AddrRange> _ranges)
+BridgeBase::BridgeResponsePort::BridgeResponsePort(
+    const std::string& _name, BridgeBase& _bridge,
+    BridgeRequestPort& _memSidePort, Cycles _delay, int _resp_limit)
     : ResponsePort(_name), bridge(_bridge),
       memSidePort(_memSidePort), delay(_delay),
-      ranges(_ranges.begin(), _ranges.end()),
       outstandingResponses(0), retryReq(false), respQueueLimit(_resp_limit),
       sendEvent([this]{ trySendTiming(); }, _name)
 {
 }
 
-Bridge::BridgeRequestPort::BridgeRequestPort(const std::string& _name,
-                                           Bridge& _bridge,
-                                           BridgeResponsePort& _cpuSidePort,
-                                           Cycles _delay, int _req_limit)
+BridgeBase::BridgeRequestPort::BridgeRequestPort(
+    const std::string& _name, BridgeBase& _bridge,
+    BridgeResponsePort& _cpuSidePort, Cycles _delay, int _req_limit)
     : RequestPort(_name), bridge(_bridge),
       cpuSidePort(_cpuSidePort),
       delay(_delay), reqQueueLimit(_req_limit),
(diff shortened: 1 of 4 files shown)


src/mem/bridge.cc
```diff
@@ -161,8 +162,11 @@ Bridge::BridgeResponsePort::recvTimingReq(PacketPtr pkt)
     DPRINTF(Bridge, "Response queue size: %d outresp: %d\n",
             transmitList.size(), outstandingResponses);
 
-    // if the request queue is full then there is no hope
-    if (memSidePort.reqQueueFull()) {
+    bool canPassThrough = bridge.canPassThrough(pkt->getAddrRange());
```
giactra: Nit: this should not be camelCase being a local variable. Use can_pass_through or can_pass instead

src/mem/bridge.cc
```diff
@@ -327,36 +337,50 @@ Bridge::BridgeResponsePort::trySendTiming()
...
 
 Tick
-Bridge::BridgeResponsePort::recvAtomicBackdoor(
-    PacketPtr pkt, MemBackdoorPtr &backdoor)
+BridgeBase::BridgeResponsePort::recvAtomicBackdoor(PacketPtr pkt,
+                                                   MemBackdoorPtr& backdoor)
 {
-    return delay * bridge.clockPeriod() + memSidePort.sendAtomicBackdoor(
-        pkt, backdoor);
+    Tick additionnalDelay = 0;
```
giactra: Ditto: additional_delay

giactra (commented):
Hi @clemdiep thanks for this change; I find it very useful.

One comment I have is about the pass-through logic. I presume you want to handle downstream issues from the bridge. Is there a specific reason for this?

For example, if we reject purely on an address basis, can't we just instantiate an IsaFake device in the PCI bus to handle missing devices?
https://github.com/gem5/gem5/blob/stable/src/dev/isa_fake.hh#L52

We already use the IsaFake as a BadAddr responder.
https://github.com/gem5/gem5/blob/stable/src/dev/Device.py#L156

**Pros**: we won't be adding virtual methods to the BaseBridge class. These methods are actually called multiple times for every incoming request. Instead we would purely rely on the routing logic.

**Cons**: It might make more sense for a root complex implementation to directly handle this.

clemdiep:
Thanks for your review @giactra 

> One comment I have is about the pass-through logic. I presume you want to handle downstream issues from the bridge. Is there a specific reason for this?

No more specific reason, it's a remnant from my early implementation for PCI, where I couldn't easily use IsaFake for some reason. But, in fact, it will now work.

> **Cons**: It might make more sense for a root complex implementation to directly handle this.

This should not be a big deal, as invalid PCI configuration access should only be made during PCI device discovery and so during OS boot.

So, I will remove that pass-through logic as it will simplify the bridge class

src/mem/bridge.cc
clemdiep: Removed with pass-through logic removal

src/mem/bridge.cc
clemdiep: Removed with pass-through logic removal

giactra:
Thanks, looks great to me. I will approve this, but can I ask you first to squash the two commits together?
If a PR contains more than 1 commit, a merge commit will be generated. It is not a big deal but I'd like to avoid it if possible (also, I will be the author of the merge commit. I have never liked this as it looks like I am taking credit for someone else work :) )

clemdiep:
Done, the two commits are squashed !

giactra (approved):
Thanks for your changes!

PR #2510, opened 2025-08-05
cpu-o3: Bundle some RenameStats into a vector

giactra:
This commit is bundling the following Scalar stats into a single vector stat:

* squashCycles = "Number of cycles rename is squashing"
* idleCycles = "Number of cycles rename is idle"
* blockCycles = "Number of cycles rename is blocking"
* serializeStallCycles = "Number of cycles rename stalled for serializing inst"
* runCycles = "Number of cycles rename is running"
* unblockCycles = "Number of cycles rename is unblocking"

There are two reasons behind this change:

1) More compact definition of the stats
2) Allowing vector level analysis:
At the moment the stats are unrelated to each other so it is not immediate to assess the cycle distribution among those. For example my stat file could end up looking like

rename.squashCycles = 235400
rename.idleCycles = 35404
rename.blockCycles = 2399
...

While I can infer from a qualitative point of view the rename stage is spending most of its time squashing, it is not immediate to extrapolate how much of the time compared to the overall number of cycles.

gem5 statistics::Vector supports this natively with the Stats::pdf flag.

So with the new Vector stat this will look like:

rename.status::Squashing = 235400, 75%
rename.status::Idle = 35404, 23%
rename.status::Blocking = 2399, 5%
...

So I know we are squashing 75% of our time.

Change-Id: I9ca58744093945966e96d18402bdb44986cf23d3

### DIFF
--- a/src/cpu/o3/rename.cc
+++ b/src/cpu/o3/rename.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010-2012, 2014-2019 ARM Limited
+ * Copyright (c) 2010-2012, 2014-2019, 2025 Arm Limited
  * Copyright (c) 2013 Advanced Micro Devices, Inc.
  * All rights reserved.
  *
@@ -58,6 +58,28 @@ namespace gem5
 namespace o3
 {
 
+// clang-format off
+std::string Rename::RenameStats::statusStrings[ThreadStatusMax] = {
+    "Running",
+    "Idle",
+    "StartSquash",
+    "Squashing",
+    "Blocked",
+    "Unblocking",
+    "SerializeStall",
+};
+
+std::string Rename::RenameStats::statusDefinitions[ThreadStatusMax] = {
+    "Number of cycles rename is running",
+    "Number of cycles rename is idle",
+    "Not Used",
+    "Number of cycles rename is squashing",
+    "Number of cycles rename is blocking",
+    "Number of cycles rename is unblocking",
+    "Number of cycles rename stalled for serializing inst",
+};
+// clang-format on
+
 Rename::Rename(CPU *_cpu, const BaseO3CPUParams &params)
     : cpu(_cpu),
       iewToRenameDelay(params.iewToRenameDelay),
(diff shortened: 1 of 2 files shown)


src/cpu/o3/rename.cc
```diff
@@ -152,12 +164,12 @@ Rename::RenameStats::RenameStats(statistics::Group *parent)
...
 
 {
-    squashCycles.prereq(squashCycles);
-    idleCycles.prereq(idleCycles);
-    blockCycles.prereq(blockCycles);
-    serializeStallCycles.flags(statistics::total);
-    runCycles.prereq(idleCycles);
-    unblockCycles.prereq(unblockCycles);
+    status.init(ThreadStatusMax).flags(statistics::pdf | statistics::nozero);
+    ;
```
dhschall:
```suggestion
```

Or is that needed?

src/cpu/o3/rename.cc
```diff
@@ -58,6 +58,28 @@ namespace gem5
...
+    "Squashing",
+    "Blocked",
+    "Unblocking",
+    "SerializeStall",
+};
+
+std::string Rename::RenameStats::statusDefinitions[ThreadStatusMax] = {
+    "Number of cycles rename is running",
+    "Number of cycles rename is idle",
+    "Not Used",
```
dhschall, line 75: Shouldn't `StartSquash` be removed at all if its not used?

dhschall (commented):
Wow, this is a great idea and improvement.
There are only two small comments but otherwise it looks good.

dhschall:
Actually, I think we should do this for the other stages as well. Do you think it would also be possible to combine the status, including (`statusString` and `statusDescription`) for all stages?

src/cpu/o3/rename.cc
giactra: Yes, probably, but I am keeping it for now as I believe removing that available state requires a separate PR

src/cpu/o3/rename.cc
giactra: It doesn't! Thanks for spotting this

giactra:
> Actually, I think we should do this for the other stages as well. Do you think it would also be possible to combine the status, including (`statusString` and `statusDescription`) for all stages?

Yes, I agree we could be doing something similar for the other stages (decode for example)

I don't think we should be reusing the same string array for other pipeline stages... I think it can make things more confusing. At the end of the day, every stage has a different definition of enum ThreadStatus which is tightly coupled to the stat it is supposed to index.
The moment we merge all ThreadStatus definitions into a single one (and I don't see the point) we should probably have a single unified array of strings..

dhschall (approved):

PR #2514, opened 2025-08-06
sim, cpu: use hypercalls for Simulator.schedule_max_insts

erin-le:
This PR modifies Simulator.schedule_max_insts() to use hypercalls instead of classic exit events. The simulation will now handle reaching a maximum number of instructions using the hypercall 8 handler. Other types of exit events that use the same underlying C++ functions that schedule_max_insts() uses will continue to use classic exit events.

### DIFF
--- a/src/cpu/BaseCPU.py
+++ b/src/cpu/BaseCPU.py
@@ -139,6 +139,10 @@ class BaseCPU(ClockedObject):
     max_insts_any_thread = Param.Counter(
         0, "terminate when any thread reaches this inst count"
     )
+    hypercall_num_max_inst_any_thread = Param.Counter(
+        0,
+        "Hypercall number of the hypercall to call upon exiting with `max_insts_any_thread` param",
+    )
     simpoint_start_insts = VectorParam.Counter(
         [], "starting instruction counts of simpoints"
     )
--- a/src/python/gem5/components/processors/abstract_core.py
+++ b/src/python/gem5/components/processors/abstract_core.py
@@ -167,7 +167,7 @@ class AbstractCore(SubSystem):
 
     @abstractmethod
     def _set_inst_stop_any_thread(
-        self, inst: int, board_initialized: bool
+        self, inst: int, board_initialized: bool, hypercall_num: int = 0
     ) -> None:
         """Schedule an exit event when any thread in this core reaches the
         given number of instructions. This is called through the simulator
@@ -180,6 +180,7 @@ class AbstractCore(SubSystem):
                                   initialized, otherwise ``False``. This parameter is
                                   necessary as the instruction stop is setup
                                   differently dependent on this.
+        :param hypercall_num: The hypercall number of the exit event.
         """
         raise NotImplementedError("This core type does not support MAX_INSTS")
 
--- a/src/sim/sim_events.cc
+++ b/src/sim/sim_events.cc
@@ -179,7 +179,8 @@ LocalSimLoopExitEvent::process()
     } else {
         std::map<std::string, std::string> payload = {
             {"justification", cause}};
-        exitSimLoopWithHypercall(cause, 0, curTick(), 0, payload, hypercall_id, false);
+        exitSimLoopWithHypercall(cause, 0, curTick(), 0, payload, hypercall_id,
+                                 false);
     }
 }
 
--- a/src/sim/sim_events.cc
+++ b/src/sim/sim_events.cc
@@ -179,7 +179,7 @@ LocalSimLoopExitEvent::process()
     } else {
         std::map<std::string, std::string> payload = {
             {"justification", cause}};
-        exitSimLoopWithHypercall(cause, 0, curTick(), 0, payload, 8, false);
+        exitSimLoopWithHypercall(cause, 0, curTick(), 0, payload, hypercall_id, false);
     }
 }
 
--- a/src/sim/sim_events.cc
+++ b/src/sim/sim_events.cc
@@ -158,13 +158,27 @@ LocalSimLoopExitEvent::LocalSimLoopExitEvent(const std::string &_cause, int c,
 {
 }
 
+LocalSimLoopExitEvent::LocalSimLoopExitEvent(const std::string& _cause, int c,
+                                             Tick repeat,
+                                             uint64_t hypercall_id)
+    : Event(Sim_Exit_Pri, IsExitEvent), cause(_cause), code(c), repeat(repeat),
+      hypercall_id(hypercall_id)
+{
+}
+
 //
 // handle termination event
 //
 void
 LocalSimLoopExitEvent::process()
 {
-    exitSimLoop(cause, 0);
+    if (hypercall_id == 0) {
+        exitSimLoop(cause, 0);
+    } else {
+        std::map<std::string, std::string> payload = {
+            {"justification", cause}};
+        exitSimLoopWithHypercall(cause, 0, curTick(), 0, payload, 8, false);
+    }
 }
 
 
--- a/src/sim/sim_events.cc
+++ b/src/sim/sim_events.cc
@@ -158,13 +158,15 @@ LocalSimLoopExitEvent::LocalSimLoopExitEvent(const std::string &_cause, int c,
 {
 }
 
-LocalSimLoopExitEvent::LocalSimLoopExitEvent(const std::string& _cause, int c,
+LocalSimLoopExitEvent::LocalSimLoopExitEvent(const std::string &_cause, int c,
                                              Tick repeat,
                                              uint64_t hypercall_id)
-    : Event(Sim_Exit_Pri, IsExitEvent), cause(_cause), code(c), repeat(repeat),
+    : Event(Sim_Exit_Pri, IsExitEvent),
+      cause(_cause),
+      code(c),
+      repeat(repeat),
       hypercall_id(hypercall_id)
-{
-}
+{}
 
 //
 // handle termination event
--- a/src/cpu/base.cc
+++ b/src/cpu/base.cc
@@ -332,10 +332,8 @@ BaseCPU::init()
     // Set up instruction-count-based termination events, if any. This needs
     // to happen after threadContexts has been constructed.
     if (params().max_insts_any_thread != 0) {
-        scheduleInstStopAnyThread(
-            params().max_insts_any_thread,
-            params().hypercall_num_max_inst_any_thread
-        );
+        scheduleInstStopAnyThread(params().max_insts_any_thread,
+                                  params().hypercall_num_max_inst_any_thread);
     }
 
     // Set up instruction-count-based termination events for SimPoints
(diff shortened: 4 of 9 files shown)


powerjg:
I don't understand how this works with the "old" exit events. In `set_se_binary_workload` we schedule all of the instruction stops (via `BaseCPUCore.scheduleSimpointsInstStop`). So, at the beginning of the simulation, we have 4 stops scheduled. However, in your script, you schedule *another* event for every case. E.g., `simulator.schedule_max_insts(board.get_simpoint().get_warmup_list()[0])`. But that event has already been schedule by the SimPoint code. 

As to why this doesn't happen in the old code... The problem is that you should have used the exit event type "SIMPOINT_BEGIN" instead of "MAX_INSTS". I'll note that this confusion is part of why we're going to hypercalls!

erin-le:
Just an update on this: I tried testing this PR with a simpler script that only sets max insts and saw that it worked as expected. The situation with `simpoint_se_restore.py` is as follows: the classic script has separate handlers for ExitEvent.MAX_INSTS and ExitEvent.SIMPOINT_BEGIN. It doesn't register a handler for ExitEvent.SIMPOINT_BEGIN, so it only handles the MAX_INSTS events.

If the script is translated to use hypercalls and is run on a build of this PR, the equivalents of MAX_INSTS and SIMPOINT_BEGIN are handled using the same exit handler. This results in there being **more handled exits** from the hypercall script than the classic script.

The only way I could get the behaviors to match up (with the current implementation) is to only use/register handlers for SIMPOINT_BEGIN events, without calling simulator.schedule_max_insts. The behavior of this is different from the current behavior of the `simpoint_se_restore.py` script, however, in that there are more exits if the simulation is allowed to run all the way through. The solution would be to have separate handlers for SIMPOINT_BEGIN and MAX_INSTS in the hypercall implementation as well.

**TLDR**: The "classic" script + `develop` branch have separate handlers for MAX_INSTS and SIMPOINT_BEGIN. This PR handles these events with the same handler. This PR should also have separate handlers.

Other concerns:
  - The current implementation of hypercall max insts in this PR doesn't preserve backwards compatibility with classic exit events.
  - Which hypercall numbers should the new handlers use?

@Harshil2107: Hopefully this explanation and the attached scripts are helpful. Please let me know if you need more information to look into this PR. @powerjg: Please let me know if my understanding of what is happening is incorrect, and there are issues with my explanation.

Below are scripts that may be useful for investigating this further:
The two scripts for only testing MAX_INSTS:
[tester-7th-100M-classic-handler.py](https://github.com/user-attachments/files/22675592/tester-7th-100M-classic-handler.py) - run this on `develop`
[tester-7th-100M.py](https://github.com/user-attachments/files/22675597/tester-7th-100M.py) - run this on a build of this PR


The two simpoint_se_restore scripts that only use SIMPOINT_BEGIN, and which have matching behavior:
[rewrite-simpoints-se-restore-for-hypercall.py](https://github.com/user-attachments/files/22675620/rewrite-simpoints-se-restore-for-hypercall.py) - run this on a build of this PR
[simpoints-se-restore-no-schedule-max-insts.py](https://github.com/user-attachments/files/22675623/simpoints-se-restore-no-schedule-max-insts.py) - run this on a build of `develop`.

A minimally modified version of `simpoints_se_restore.py`. The handler yields `False` under both conditions, and prints some messages:
[simpoints-se-restore.py](https://github.com/user-attachments/files/22675646/simpoints-se-restore.py) - run this on `develop`.

A "directly translated" version of `simpoints_se_restore.py`, which results in more exits than the classic `simpoints_se_restore.py`.
[rewrite-simpoints-se-restore-for-hypercall-schedulemaxinsts.py](https://github.com/user-attachments/files/22675713/rewrite-simpoints-se-restore-for-hypercall-schedulemaxinsts.py) - Run this on a build of this PR.

src/python/gem5/components/processors/abstract_core.py
```diff
@@ -156,7 +156,7 @@ def _set_simpoint(
 
     @abstractmethod
     def _set_inst_stop_any_thread(
-        self, inst: int, board_initialized: bool
+        self, inst: int, board_initialized: bool, hypercall_num: int = 0
```
powerjg, line 170: why isn't this hardcoded to 8?

src/sim/sim_events.cc
```diff
@@ -158,13 +158,29 @@ LocalSimLoopExitEvent::LocalSimLoopExitEvent(const std::string &_cause, int c,
...
 LocalSimLoopExitEvent::process()
 {
-    exitSimLoop(cause, 0);
+    if (hypercall_id == 0) {
+        exitSimLoop(cause, 0);
+    } else {
+        std::map<std::string, std::string> payload = {
+            {"justification", cause}};
+        // exitSimulationLoopNow(8, payload);
+        // exitSimulationLoop(8, payload);
```
powerjg: Remove unnecessary comments.

powerjg (changes requested):
One question (same as comment below): Why is there a parameter for the hypercall if we are going to always use 8? I *think* I can guess the design decision, but I would like to see it written down.

One other requirement before merging (other than fixing the test), we need to clean up the commit messages.

src/python/gem5/components/processors/abstract_core.py
Harshil2107: I decided not to hardcode the default value to 8 for two reasons:

- In most cases, we default to using the classic exit events when a hypercall number is not specified. I want to keep this behavior consistent with the existing implementation, as shown here:
https://github.com/gem5/gem5/blob/ddd4ae35adb0a3df1f1ba11e9a973a5c2f8c2944/src/sim/sim_events.hh#L73

- Since the default is always 0, in the future if we or users add features that call `_set_inst_stop_any_thread` with a different hypercall number, won’t have to worry about the function accidentally invoking another hypercall by default.

erin-le:
Preserving the original PR message as a comment, as the actual PR message should be modified before merging.

This PR attempts to modify Simulator.schedule_max_insts() to use hypercalls instead of classic exit events. Specifically, hypercall 8 is used. Unfortunately, this does not work as intended, as there are more exits on the Python side when hypercalls are used compared to when classic exit events are used. Despite this, debugging with gdb shows that the function to process the exit on the C++ side (gem5::LocalSimLoopExitEvent::process) is called the same number of times for both hypercalls and classic exits. It seems that Python exits will occur every time LocalSimLoopExitEvent::process is called if hypercalls are used, and will only occur sometimes if classic exits are used.

I have included a comparison of the Python outputs below. I have also included the config scripts used to generate the Python outputs. Please let me know if you see anything in either the configs or the changes in this PR that may cause the differences in behavior between hypercalls and classic exits.

```txt
# hypercall
1721 print this
hypercall 8 handler: end of warmup, starting to simulate SimPoint
simulator.get_current_tick() is 1339404588
board.get_simpoint().get_simpoint_interval() is 1000000
hypercall 8 handler: end of SimPoint interval
simulator.get_current_tick() is 1339404588
1722 print this

# classic exit
1721 print this
develop reference: end of warmup, starting to simulate SimPoint
simulator.get_current_tick() is 1339404588
board.get_simpoint().get_simpoint_interval() is 1000000
1722 print this

```

```
# hypercall
2577 print this
hypercall 8 handler: end of SimPoint interval
simulator.get_current_tick() is 1907607150
hypercall 8 handler: end of SimPoint interval
simulator.get_current_tick() is 1907607150
2578 print this

# classic exit
2577 print this
develop reference: end of SimPoint interval
simulator.get_current_tick() is 1907607150
2578 print this

```

```
# hypercall
3433 print this
hypercall 8 handler: end of SimPoint interval
simulator.get_current_tick() is 2475819036
3434 print this

# classic exit
3433 print this
3434 print this
```

```
# hypercall
12801 print this
hypercall 8 handler: end of SimPoint interval
simulator.get_current_tick() is 8721667269
12802 print this

# classic exit
12801 print this
12802 print this
```

Configuration scripts used:
With hypercalls: 
[simpoints-se-restore-hypercall-py.txt](https://github.com/user-attachments/files/21609982/simpoints-se-restore-hypercall-py.txt)

With classic exits: 
[simpoints-se-restore-py.txt](https://github.com/user-attachments/files/21609986/simpoints-se-restore-py.txt)

Harshil2107 (dismissed):
lgtm

src/sim/sim_events.cc
```diff
@@ -158,13 +158,29 @@ LocalSimLoopExitEvent::LocalSimLoopExitEvent(const std::string &_cause, int c,
...
 void
 LocalSimLoopExitEvent::process()
 {
-    exitSimLoop(cause, 0);
+    if (hypercall_id == 0) {
+        exitSimLoop(cause, 0);
+    } else {
+        std::map<std::string, std::string> payload = {
+            {"justification", cause}};
+        exitSimLoopWithHypercall(cause, 0, curTick(), 0, payload, 8, false);
```
Harshil2107:
```suggestion
        exitSimLoopWithHypercall(cause, 0, curTick(), 0, payload, hypercall_id, false);
```

One small thing. Shouldnt we be calling `exitSimLoopWithHypercall` with `hypercall_id` instead of hard coding it to 8?

BobbyRBruce: i'm pretty sure you're correct @Harshil2107 , so I added the patch to the PR. Though I'll wait for confirmation on this before merging.

erin-le: @BobbyRBruce: Yes, that should be fine

src/cpu/BaseCPU.py
```diff
@@ -139,6 +139,10 @@ def takeOverFrom(self, old_cpu):
     max_insts_any_thread = Param.Counter(
         0, "terminate when any thread reaches this inst count"
     )
+    hypercall_num_max_inst_any_thread = Param.Counter(
+        0,
+        "Hypercall number of the hypercall to call upon exiting with `max_insts_any_thread` param",
+    )
```
BobbyRBruce, line 145: I'm confused by this... if the hypercall int returned is specified here, then why do we force it to be a value of 8 when calling `simulator.schedule_max_insts`?  What is the purpose of someone setting this parameter?

Harshil2107: This variable is set by the `_set_inst_stop_any_thread` (https://github.com/gem5/gem5/blob/a5dbe41ebc435cb6c0b6489eac3d33459ff38b97/src/python/gem5/components/processors/base_cpu_core.py#L177) function if it is called to set Max Insts before the board is initialized.

Harshil2107 (approved):
lgtm

src/cpu/BaseCPU.py
BobbyRBruce: I need a bit more information here because I'm still confused. Let's just talk about the pre-instantation case for now: This parameter, to my understanding, is only used to know what hypercall ID to return when max_inst is triggered. We schedule max_inst events by calling `simulator.schedule_max_inst` and in this function it overrides this parameter and sets it to 8. So if we set the parameter before calling the function, that setting doesn't matter due to the overwriting that happens.  We can change it back to something after calling `simulator.schedule_max_inst` but it’s weird to run `simulator.schedule_max_inst` first then configure what you actually want the scheduled event to do (i.e., which hypercall id to return) after. Overriding this parameter to 8 in `simulator.schedule_max_inst` also begs the question as to why 0 is the default at all. It doesn't matter what the default is, the default is never used because `simulator.schedule_max_inst` always overwrites the default with 8.

Harshil2107: 1) If the user wants to define custom behaviour for th max inst hypercall, they can override the hypercall class. We can hard code the max inst to 8 instead of creating a new param but I feel it is bettre to have one place that defines the hypercall we want to call instead of hard coding the numbers in the function call.

2) we default to using the classic exit events (0) when a hypercall number is not specified. I want to keep this behavior consistent with the existing implementation, as shown here:
https://github.com/gem5/gem5/blob/ddd4ae35adb0a3df1f1ba11e9a973a5c2f8c2944/src/sim/sim_events.hh#L73

src/python/gem5/components/processors/abstract_core.py
powerjg: @BobbyRBruce see this.

PR #2517, opened 2025-08-06
cpu-o3: Bundle some DecodeStats into a vector

giactra:
Same idea as the rename stage (PR 2510):
https://github.com/gem5/gem5/pull/2510

Change-Id: I42aecb899206a414b47ccc5e21d9aaabf3269d7f


dhschall (approved):
Great! Thanks for doing that so quickly after the other!

PR #2518, opened 2025-08-07
cpu-o3: Bundle some Fetch/IEW/Commit stats into a vector

dhschall:
Apply the same idea as the rename stage (PR 2510):
#2510, #2517 

Furthermore, remove many if-else to check the state in favour of simply indexing with the state

### DIFF
--- a/src/cpu/o3/commit.cc
+++ b/src/cpu/o3/commit.cc
@@ -72,6 +72,27 @@ namespace gem5
 namespace o3
 {
 
+// clang-format off
+std::string Commit::CommitStats::statusStrings[ThreadStatusMax] = {
+    "running",
+    "idle",
+    "robSquashing",
+    "trapPending",
+    "fetchTrapPending",
+    "squashAfterPending",
+};
+
+std::string Commit::CommitStats::statusDefinitions[ThreadStatusMax] = {
+    "Number of cycles commit is running",
+    "Number of cycles commit is idle",
+    "Number of cycles commit is squashing the ROB",
+    "Number of cycles commit is processing a trap",
+    "Number of cycles commit is processing a fetch trap",
+    "Number of cycles commit is squashing instructions after a pending "
+    "squash",
+};
+// clang-format on
+
 void
 Commit::processTrapEvent(ThreadID tid)
 {
@@ -146,6 +167,8 @@ Commit::regProbePoints()
 
 Commit::CommitStats::CommitStats(CPU *cpu, Commit *commit)
     : statistics::Group(cpu, "commit"),
+      ADD_STAT(status, statistics::units::Cycle::get(),
+               "Commit status cycles"),
       ADD_STAT(commitSquashedInsts, statistics::units::Count::get(),
                "The number of squashed insts skipped by commit"),
       ADD_STAT(commitNonSpecStalls, statistics::units::Count::get(),
@@ -166,6 +189,11 @@ Commit::CommitStats::CommitStats(CPU *cpu, Commit *commit)
 {
     using namespace statistics;
 
+    status.init(ThreadStatusMax).flags(statistics::pdf | statistics::nozero);
+    for (int i = 0; i < ThreadStatusMax; ++i) {
+        status.subname(i, statusStrings[i]);
+        status.subdesc(i, statusDefinitions[i]);
+    }
     commitSquashedInsts.prereq(commitSquashedInsts);
     commitNonSpecStalls.prereq(commitNonSpecStalls);
     branchMispredicts.prereq(branchMispredicts);
@@ -729,6 +757,7 @@ Commit::commit()
 
     int num_squashing_threads = 0;
     for (ThreadID tid : *activeThreads) {
+        stats.status[commitStatus[tid]]++;
         // Not sure which one takes priority.  I think if we have
         // both, that's a bad sign.
         if (trapSquash[tid]) {
--- a/src/cpu/o3/fetch.cc
+++ b/src/cpu/o3/fetch.cc
@@ -74,6 +74,40 @@ namespace gem5
 namespace o3
 {
 
+// clang-format off
+std::string Fetch::FetchStatGroup::statusStrings[ThreadStatusMax] = {
+    "running",
+    "idle",
+    "squashing",
+    "blocked",
+    "fetching",
+    "trapPending",
+    "quiescePending",
+    "itlbWait",
+    "icacheWaitResponse",
+    "icacheWaitRetry",
+    "icacheAccessComplete",
+    "ftqWait",
+    "noGoodAddr",
+};
+
+std::string Fetch::FetchStatGroup::statusDefinitions[ThreadStatusMax] = {
+    "Number of cycles fetch has run and was not squashing or blocked",
+    "Number of cycles fetch was idle",
+    "Number of cycles fetch has spent squashing",
+    "Number of cycles fetch has spent blocked",
+    "Not in use",
+    "Number of cycles fetch is waiting for a trap to be processed",
+    "Number of cycles fetch is waiting for a quiesce instr. to be processed",
+    "Number of cycles fetch is waiting for the ITLB to respond translation",
+    "Number of cycles fetch is waiting for the Icache to respond",
+    "Number of cycles fetch is waiting for the Icache to retry",
+    "Number of cycles fetch is resuming after Icache access",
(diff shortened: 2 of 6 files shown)


dhschall:
@giactra, I've tried to do the same for the other stages.
Before we merge two things:

- What do you think about removing all the if-else
- I also think I should properly clean up the stats in the fetch stage.

I thought about using the `descriptionString` also in the debug messages
```
DPRINTF(Fetch, "[tid:%i] Fetch is waiting cache response!\n",
```
to something like
```
DPRINTF(Fetch, "[tid:%i] Fetch is waiting %s\n", tid, description[status]);
```

src/cpu/o3/decode.cc
```diff
@@ -598,11 +598,7 @@ Decode::decode(bool &status_change, ThreadID tid)
     //     continue trying to empty skid buffer
     //     check if stall conditions have passed
 
-    if (decodeStatus[tid] == Blocked) {
-        ++stats.status[Blocked];
-    } else if (decodeStatus[tid] == Squashing) {
-        ++stats.status[Squashing];
-    }
+    stats.status[decodeStatus[tid]];
```
dhschall, line 601:
```suggestion
    stats.status[decodeStatus[tid]]++;
```

giactra:
HI @dhschall thanks for adding this PR!

> @giactra, I've tried to do the same for the other stages. Before we merge two things:
> 
>     * What do you think about removing all the if-else
> 
>     * I also think I should properly clean up the stats in the fetch stage.
> 
> 
> I thought about using the `descriptionString` also in the debug messages
> 
> ```
> DPRINTF(Fetch, "[tid:%i] Fetch is waiting cache response!\n",
> ```
> 
> to something like
> 
> ```
> DPRINTF(Fetch, "[tid:%i] Fetch is waiting %s\n", tid, description[status]);
> ```

Yes both points make lots of sense! Though IMHO they should go on separate PRs.

Personally I would have the current PR simply transform Scalar stats into the Vector counterpart as you did for Fetch/IEW/Commit stages.

I would open separate ones for the other changes you suggested.
I also want to stress these are lower priorities PR (IMHO) with respect to the more important ones like FDP and TDA. I realise I should have waited for example to get TDA merged before posting my rename/decode PRs. Now it is too late and anyway changes needed are minimal.

So my proposal is to get these transformation in, but wait for the aforementioned PRs to get merged before we create too many conflicts.. What do you think? (Afterall both PRs come from you and your team :) )

erin-le:
@dhschall: Hello, could you respond to the comment above when you get a chance? Thank you!

dhschall:
Thanks @erin-le for the reminder.
@giactra, I have now rebased this PR and made only the changes to replace the individual stats with the vector.
In the next PR, I can do the rest.

Since the TDM PR might need a bit more time we can do this one first I believe

erin-le:
@giactra: Could you take another look at this PR when you get a chance? Thank you!

src/cpu/o3/fetch.cc
```diff
@@ -75,6 +75,40 @@ namespace gem5
...
+    "ItlbWait",
+    "IcacheWaitResponse",
+    "IcacheWaitRetry",
+    "IcacheAccessComplete",
+    "FTQEmpty",
+    "NoGoodAddr",
+};
+
+std::string Fetch::FetchStatGroup::statusDefinitions[ThreadStatusMax] = {
+    "fetch is running",
```
giactra: For consistency with previous definition, this should be:

> Number of cycles fetch has run and was not squashing or blocked

src/cpu/o3/fetch.cc
```diff
@@ -75,6 +75,40 @@ namespace gem5
...
+    "fetch is idle",
```
giactra: Ditto:

> Number of cycles fetch was idle

src/cpu/o3/fetch.cc
```diff
@@ -75,6 +75,40 @@ namespace gem5
...
+    "fetch is squashing",
```
giactra: Ditto:

> Number of cycles fetch has spent squashing

src/cpu/o3/fetch.cc
```diff
@@ -75,6 +75,40 @@ namespace gem5
...
+    "fetch is blocking",
```
giactra: Ditto:

> Number of cycles fetch has spent blocked

src/cpu/o3/fetch.cc
```diff
@@ -75,6 +75,40 @@ namespace gem5
...
+    "Not in use",
+    "fetch is waiting for a trap to be processed",
+    "fetch is waiting for a quiesce instr. to be processed",
+    "fetch is waiting for the ITLB to respond translation",
+    "fetch is waiting for the Icache to respond",
+    "fetch is waiting for the Icache to retry",
+    "fetch is resuming after Icache access",
+    "fetch is waiting for the FTQ to become ready",
```
giactra: Ditto for these others

src/cpu/o3/fetch.cc
```diff
@@ -74,6 +74,40 @@ namespace gem5
...
+    "Blocked",
+    "Fetching",
+    "TrapPending",
+    "QuiescePending",
+    "ItlbWait",
+    "IcacheWaitResponse",
+    "IcacheWaitRetry",
+    "IcacheAccessComplete",
+    "FTQEmpty",
+    "NoGoodAddr",
```
giactra: Ditto

src/cpu/o3/commit.cc
```diff
@@ -72,6 +72,27 @@ namespace gem5
...
 {
 
+// clang-format off
+std::string Commit::CommitStats::statusStrings[ThreadStatusMax] = {
+    "Running",
+    "Idle",
+    "ROBSquashing",
+    "TrapPending",
+    "FetchTrapPending",
+    "SquashAfterPending",
```
giactra: This will turn out to be the name of the stats. I think in general we use camelCase so the first letter should be lowercase

src/cpu/o3/fetch.cc
```diff
@@ -1107,7 +1103,7 @@ Fetch::fetch(bool &status_change)
             DPRINTF(Fetch, "[tid:%i] PC:%#x not within fetch target: %s\n",
                     tid, this_pc, curFT->toString());
             bacResteer(this_pc, tid);
-            ++fetchStats.ftqStallCycles;
+            fetchStats.status[FTQEmpty]++;
```
giactra: Does this really means the FTQ is empty? It looks to me from the assertion above that the FTQ actually contains some entries

src/cpu/o3/fetch.cc
```diff
@@ -1137,11 +1133,12 @@ Fetch::fetch(bool &status_change)
 
             if (fetchStatus[tid] == IcacheWaitResponse) {
                 cpu->fetchStats[tid]->icacheStallCycles++;
```
giactra, line 1135: Do we still need this now?

src/cpu/o3/iew.cc
```diff
@@ -818,12 +840,12 @@ IEW::dispatch(ThreadID tid)
...
     //     check if stall conditions have passed
-
+    iewStats.dispatchStatus[dispatchStatus[tid]]++;
     if (dispatchStatus[tid] == Blocked) {
-        ++iewStats.blockCycles;
+        iewStats.dispatchStatus[Blocked]++;
 
     } else if (dispatchStatus[tid] == Squashing) {
-        ++iewStats.squashCycles;
+        iewStats.dispatchStatus[Squashing]++;
```
giactra: We are incrementing this twice

giactra (commented):
Hi @dhschall apologies for the late review, I have been relatively busy

erin-le:
@dhschall: Hello, similarly to the other PR, just a reminder to take another look at this and make changes before Monday if you'd like to get this in for v25.1. Thank you!

src/cpu/o3/fetch.cc
dhschall: What do you mean? the `icacheStallCycles`? I think yes because its a common stat that is used across CPU types.

src/cpu/o3/fetch.cc
dhschall: True. I guess the status should be renamed to FTQWait

powerjg:
This lgtm. It would be great if we could bundle this with the other stat changes in 25.1

BobbyRBruce (approved):

PR #2519, opened 2025-08-11
misc: improvement to clang-format

clemdiep:
This commit makes clang-format break before open braces for struct, union and namespace, like it does for class.

Result in

```cpp
namespace gem5
{
```

instead of

```cpp
namespace gem5 {
```

It will also regroup open and close brace for empty function on a single line (meanly for constructor), which seems to be the more common way used accross the project.

Result in

```cpp
MyClass::MyClass()
{}
```

instead of

```cpp
MyClass::MyClass()
{
}
```

Finally, it defines the macros BitUnion and SubBitUnion as block begin/end so they are indented correctly.

Result in

```cpp
BitUnion8(Example)
    Bitfield<8> field1;
    Bitfield<7> field2;
    ...
EndBitUnion(Example)
```

instead of

```cpp
BitUnion8(Example) Bitfield<8> field1;
bitfield<7> field2;
...
EndBitUnion(Example)
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -51,10 +51,15 @@ AccessModifierOffset: -2                     # Outdent public/protected/private 
 # Brace placement rules
 BreakBeforeBraces: Custom
 BraceWrapping:
-    AfterFunction: true                      # Function/class braces on newline (sec. Braces)
+    AfterFunction: true                      # Function/object braces on newline (sec. Braces)
     AfterClass: true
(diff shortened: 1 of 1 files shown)


giactra (approved):
Looks non-controvertial to me, waiting for @powerjg to approve as well

powerjg (approved):
No argument here!

PR #2524, opened 2025-08-12
tests: IPC Regression Tests

### LINKED ISSUE #702: Add automatic regression test to check for performance degradations

erin-le:
This commit adds a version of the IPC regression tests that uses a reference value instead of cached IPC values from the previous run of the tests (see https://github.com/gem5/gem5/pull/2413 for previous version). The tests run hello world binaries and matrix multiply workloads for the Arm, RISC-V, and X86 ISAs.

### DIFF
--- a/tests/gem5/regression_tests/configs/hello-world-binary.py
+++ b/tests/gem5/regression_tests/configs/hello-world-binary.py
@@ -1,0 +1,141 @@
+# Copyright (c) 2021-2025 The Regents of the University of California
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
+This gem5 configuation script creates a simple board to run a matrix multiply
+workload. The ISA of the board and workload depends on the argument passed
+into the config.
+
+This setup is close to the simplest setup possible using the gem5 library. It
+does not contain any kind of caching, IO, or any non-essential components.
+
+Usage
+-----
+
+```
+cd tests
+./main.py run gem5/regression_tests -t <number of threads to run tests with>
+
+```
+"""
+
+import argparse
+
+from gem5.components.boards.simple_board import SimpleBoard
+from gem5.components.cachehierarchies.classic.no_cache import NoCache
+from gem5.components.memory.simple import SingleChannelSimpleMemory
+from gem5.components.processors.cpu_types import CPUTypes
+from gem5.components.processors.simple_processor import SimpleProcessor
+from gem5.isas import get_isa_from_str
+from gem5.resources.resource import obtain_resource
+from gem5.simulate.simulator import Simulator
+from gem5.utils.requires import requires
+
+parser = argparse.ArgumentParser()
+
+parser.add_argument(
+    "--isa",
+    required=True,
+    choices=["arm", "x86", "riscv"],
+    help="Enter the isa to use.",
+)
+
+args = parser.parse_args()
+
+# This check ensures the gem5 binary contains the ISA target. If not,
+# an exception will be thrown.
+requires(isa_required=get_isa_from_str(args.isa))
+
+# In this setup we don't have a cache. `NoCache` can be used for such setups.
+cache_hierarchy = NoCache()
+
+# We use a simple memory system. Taken from
+# tests/gem5/replacement_policies/configs/run_replacement_policy.py
+memory = SingleChannelSimpleMemory(
+    latency="30ns",
+    latency_var="0ns",
+    bandwidth="12.8GiB/s",
+    size="512MiB",
+)
+
+# We use a simple Timing processor with one core.
+processor = SimpleProcessor(
+    cpu_type=CPUTypes.TIMING, isa=get_isa_from_str(args.isa), num_cores=1
+)
+
+# The gem5 library simple board which can be used to run simple SE-mode
+# simulations.
+board = SimpleBoard(
+    clk_freq="3GHz",
+    processor=processor,
+    memory=memory,
+    cache_hierarchy=cache_hierarchy,
+)
+
+# Here we set the workload. In this case we want to run a matrix multiply
+# workload.
+if args.isa == "riscv":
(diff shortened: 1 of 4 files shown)


tests/gem5/regression_tests/configs/hello-world-binary.py
```diff
@@ -0,0 +1,133 @@
...
+
+# This check ensures the gem5 binary contains the ISA target. If not,
+# an exception will be thrown.
+requires(isa_required=get_isa_from_str(args.isa))
+
+# In this setup we don't have a cache. `NoCache` can be used for such setups.
+cache_hierarchy = NoCache()
+
+# We use a single channel DDR3_1600 memory system
+memory = SingleChannelDDR3_1600(size="32MiB")
```
powerjg: Let's use simple memory so we have a more narrow test of the core.

tests/gem5/regression_tests/configs/hello-world-binary.py
```diff
@@ -0,0 +1,133 @@
...
+simulator = Simulator(board=board)
+simulator.run()
+
+stats = simulator.get_stats()
+
+# Compare values in the current run against reference IPC values calculated
+# using simInsts / numCycles. The reference IPC values were obtained by running
+# the tests on gem5 v25.0.0.0.
+
+if args.isa == "arm":
```
powerjg: Let's have a dictionary for this instead of if/else

tests/gem5/regression_tests/configs/hello-world-binary.py
```diff
@@ -0,0 +1,133 @@
...
+    / stats["board"]["processor"]["cores"]["value"][0]["core"]["numCycles"][
+        "value"
+    ]
+)
+
+print(f"Current IPC is: {curr_ipc}")
+print(f"Comparison IPC is: {reference_ipc}")
+print(f"Current IPC is {curr_ipc/reference_ipc:.2%} of reference")
+
+if curr_ipc < 0.98 * reference_ipc:
```
powerjg: We should also check if ipc goes up

tests/gem5/regression_tests/configs/hello-world-binary.py
```diff
@@ -0,0 +1,139 @@
...
+    clk_freq="3GHz",
+    processor=processor,
+    memory=memory,
+    cache_hierarchy=cache_hierarchy,
+)
+
+# Here we set the workload. In this case we want to run a matrix multiply
+# workload.
+if args.isa == "riscv":
+    workload_name = "riscv-hello"
```
powerjg:
```suggestion
    binary_name = "riscv-hello"
```

tests/gem5/regression_tests/configs/hello-world-binary.py
```diff
@@ -0,0 +1,139 @@
...
+print(f"Current IPC is: {curr_ipc}")
+print(f"Comparison IPC is: {reference_ipcs[args.isa]}")
+print(f"Current IPC is {curr_ipc/reference_ipcs[args.isa]:.2%} of reference")
+
+if curr_ipc < 0.98 * reference_ipcs[args.isa]:
+    print("The current IPC is less than 98 percent of the reference IPC!")
+    exit(1)
+
+if curr_ipc > reference_ipcs[args.isa]:
+    print("The current IPC is greater than the reference ISA!")
```
powerjg:
```suggestion
if curr_ipc < 0.98 * reference_ipcs[args.isa] or curr_ipc > reference_ipcs[args.isa]:
    print(f"The current IPC is {curr_ipc}. Expected {reference_ipcs[args.isa]}")
    exit(1)
```

tests/gem5/regression_tests/configs/matrix-multiply.py
```diff
@@ -0,0 +1,138 @@
...
+print(f"Current IPC is: {curr_ipc}")
+print(f"Comparison IPC is: {reference_ipcs[args.isa]}")
+print(f"Current IPC is {curr_ipc/reference_ipcs[args.isa]:.2%} of reference")
+
+if curr_ipc < 0.98 * reference_ipcs[args.isa]:
+    print("The current IPC is less than 98 percent of the reference IPC!")
+    exit(1)
+
+if curr_ipc > reference_ipcs[args.isa]:
+    print("The current IPC is greater than the reference ISA!")
```
powerjg: Ditto above

tests/gem5/regression_tests/configs/matrix-multiply.py
```diff
@@ -0,0 +1,138 @@
...
+stats = simulator.get_stats()
+
+# Compare values in the current run against reference IPC values calculated
+# using simInsts / numCycles. The reference IPC values were obtained by running
+# the tests on gem5 v25.0.0.0.
+
+reference_ipcs = {
+    "arm": 0.006344991461591237,
+    "riscv": 0.006273970221306115,
+    "x86": 0.005040373616806491,
```
powerjg, line 119: These seem too low.

tests/gem5/regression_tests/configs/hello-world-binary.py
```diff
@@ -0,0 +1,139 @@
...
+stats = simulator.get_stats()
+
+# Compare values in the current run against reference IPC values calculated
+# using simInsts / numCycles. The reference IPC values were obtained by running
+# the tests on gem5 v25.0.0.0.
+
+reference_ipcs = {
+    "arm": 0.007182533150689518,
+    "riscv": 0.0063129307138331215,
+    "x86": 0.00578978714017867,
```
powerjg, line 120: These seem low, but maybe it's OK.

tests/gem5/regression_tests/configs/matrix-multiply.py
erin-le: I obtained these stats by running the test configs on gem5 v25.0.0.0. I also reran the configs on v25.0.0.1 and got the same results. When running the tests on `develop` at commit 14992bd5ba6e24bf855dbd8ee96a4913ac86f868, the test values (for all 6 tests) were between 100.00% and 100.20% of these reference values. Should I get reference stats from `develop` and use those instead?

tests/gem5/regression_tests/configs/hello-world-binary.py
erin-le: Same comment as above; should I use reference values from `develop` instead?

tests/gem5/regression_tests/configs/matrix-multiply.py
powerjg: It doesn't really matter 25.0.0.0 is fine. Just document what the commit hash was in the file, please

erin-le:
@powerjg: FYI, this is ready for re-review

tests/gem5/regression_tests/configs/hello-world-binary.py
```diff
@@ -0,0 +1,141 @@
...
+    ]
+)
+
+print(f"Current IPC is: {curr_ipc}")
+print(f"Comparison IPC is: {reference_ipcs[args.isa]}")
+print(f"Current IPC is {curr_ipc/reference_ipcs[args.isa]:.2%} of reference")
+
+if (
+    curr_ipc < 0.98 * reference_ipcs[args.isa]
+    or curr_ipc > reference_ipcs[args.isa]
```
powerjg: Should this be 


```suggestion
    or curr_ipc > reference_ipcs[args.isa]*1.02
```

tests/gem5/regression_tests/configs/matrix-multiply.py
```diff
@@ -0,0 +1,140 @@
...
+    ]
+)
+
+print(f"Current IPC is: {curr_ipc}")
+print(f"Comparison IPC is: {reference_ipcs[args.isa]}")
+print(f"Current IPC is {curr_ipc/reference_ipcs[args.isa]:.2%} of reference")
+
+if (
+    curr_ipc < 0.98 * reference_ipcs[args.isa]
+    or curr_ipc > reference_ipcs[args.isa]
```
powerjg: Ditto

powerjg (commented):
I think your testing not if the ipc is within 2% but if it is between the correct value and 2% less. We want +/-2% I believe.

Another note... If these are going to be our IPC tests, which is good for now, we should also use the same configs for testing if the different CPU models have the same stat outputs.

tests/gem5/regression_tests/configs/hello-world-binary.py
erin-le: Sounds good, I'll make the change. Since this particular line was a change you suggested, I thought you wanted the tests to fail if the newer IPC was over exactly 100%.

powerjg: I see. No, I was just lazy/wrong.

powerjg (approved):

Issue #2391, opened 2025-06-20
SE Mode: Statically linked RISC-V executable fails due to unimplemented syscall 293 (rseq)


30ZHANGjin:
**Describe the bug**

When running a complex, statically-linked RISC-V application (like llama.cpp) in Syscall Emulation (SE) mode, the simulation fails with a fatal: Syscall 293 out of range error. This occurs because the application, although statically linked, was compiled with a modern glibc which invokes the rseq (Restartable Sequences) system call during its startup or runtime. The gem5 SE mode for RISC-V currently does not have an implementation for this system call.

**Affects version**
gem5 version 24.1.0.3

**gem5 Modifications**
No modifications to the C++ source code. The issue is observed with the vanilla stable branch of gem5.

**To Reproduce**

**1. Cross-compile llama.cpp as a static RISC-V executable.** Use a modern RISC-V GCC toolchain. The following CMake command was used:
```
cmake -B build-riscv -DCMAKE_BUILD_TYPE=Release \
  -DLLAMA_STATIC=ON \
  -DGGML_OPENMP=OFF \
  -DCMAKE_SYSTEM_NAME=Linux \
  -DCMAKE_SYSTEM_PROCESSOR=riscv64 \
  -DCMAKE_C_COMPILER=/opt/riscv/bin/riscv64-unknown-linux-gnu-gcc \
  -DCMAKE_CXX_COMPILER=/opt/riscv/bin/riscv64-unknown-linux-gnu-g++
```
The resulting executable is confirmed to be statically linked:
```
$ file ./build-riscv/bin/llama-cli
llama-cli: ELF 64-bit LSB executable, UCB RISC-V, RVC, double-float ABI, version 1 (GNU/Linux), statically linked...
```

**2. Execute the compiled binary in RISC-V SE mode.**

```
build/RISCV/gem5.opt configs/deprecated/example/se.py \
--cpu-type=AtomicSimpleCPU \
--cmd=/path/to/llama-cli \
--options="-m /path/to/model.gguf -p hello"
```

If possible, please include the Python configuration script used and state clearly any parameters passed.

**Terminal Output**

```shell
fatal: syscall 293 out of range
 @ tick 14614145500
[invoke: fatal]
```

**Expected behavior**
The simulation should not crash. A more graceful behavior would be for gem5's SE mode to provide a "dummy" implementation for the rseq system call (e.g., simply returning 0 to indicate success), which is often sufficient to satisfy glibc and allow the program to continue execution.
[...]

PR #2525, opened 2025-08-12
Ignore rseq syscall on riscv to fix SE workloads built with newer glibc

jiegec:
The rseq syscall is invoked by newer glibc on e.g. Debian Trixie. It
leads to syscall out of range failure in SE mode running riscv binaries.

Reproducible via:

```
pushd tests/test-progs/hello
riscv64-linux-gnu-gcc -static src/hello.c -o bin/riscv/linux/hello
popd
./build/RISCV/gem5.opt configs/learning_gem5/part1/simple-riscv.py
```

Fails with:

```
fatal: Syscall 293 out of range
```

We can safely ignore 293 to make it work.

Additionally, to make pre-commit hook happy, run clang-format in a separate commit.

Fixes https://github.com/gem5/gem5/issues/2391.


powerjg (commented):
From https://lwn.net/Articles/883104/ it sounds like glibc is using this to initialize threads now. This change may work for hello_world, but I worry that it's going to cause issues with any multithreaded applications. From the comments in that article, a better solution would be to return `ENOSYS`.

jiegec:
> From https://lwn.net/Articles/883104/ it sounds like glibc is using this to initialize threads now. This change may work for hello_world, but I worry that it's going to cause issues with any multithreaded applications. From the comments in that article, a better solution would be to return `ENOSYS`.

Added:

```
    sim: Add enosysFunc to syscall emulation

    The existing ignoreFunc always returns success, but in some cases we
    want to return -ENOSYS to signify that the syscall is missing
    explicitly. The new enosysFunc is added for this purpose.
```

And uses enosysFunc instead of ignoreFunc for syscall 293.

src/sim/syscall_emul.hh
```diff
@@ -141,6 +141,10 @@ SyscallReturn ignoreFunc(SyscallDesc *desc, ThreadContext *tc);
 SyscallReturn
 ignoreWarnOnceFunc(SyscallDesc *desc, ThreadContext *tc);
 
+/// Handler for unimplemented syscalls that return -ENOSYS to the target
+/// program.
+SyscallReturn enosysFunc(SyscallDesc *desc, ThreadContext *tc);
```
powerjg: I like this! Could you rename to ignoreWithEnosys? This way it will be clear in the syscall list that the syscall isn't enosys but that we are ignoring it with enosys.

powerjg (commented):
Thanks for the update. I like this solution a lot. One small feedback, though.

erin-le:
@jiegec: Hello, it looks like the entirety of `se_workload.cc` has been reformatted with `clang-format`. In order to avoid mixing style changes with unrelated changes, the CI tests only check the modified/added lines in a PR using `git-clang-format`.
Could you drop the current reformatting commit and run `git-clang-format` on your changes? To do so, please run `util/run-git-clang-format.py --ci-pr-base-commit <commit hash of commit before the first commit of the PR>` from the gem5 directory. You can also look at `CI Tests / clang-format-check` for an example. Thank you!

src/sim/syscall_emul.hh
jiegec: Done.

jiegec:
> @jiegec: Hello, it looks like the entirety of `se_workload.cc` has been reformatted with `clang-format`. In order to avoid mixing style changes with unrelated changes, the CI tests only check the modified/added lines in a PR using `git-clang-format`. Could you drop the current reformatting commit and run `git-clang-format` on your changes? To do so, please run `util/run-git-clang-format.py --ci-pr-base-commit <commit hash of commit before the first commit of the PR>` from the gem5 directory. You can also look at `CI Tests / clang-format-check` for an example. Thank you!

Done, however the reformatted range is extended to the whole syscallDescs64 definition by the tool. Not sure if it is intended, though.

powerjg:
> Done, however the reformatted range is extended to the whole syscallDescs64 definition by the tool. Not sure if it is intended, though.

I think that's a reasonable compromise.

powerjg (approved):
Thanks for the updates and apologies for the clang-format annoyance. We're going through some growing pains with it.

PR #2532, opened 2025-08-15
misc: clang-format update

dhschall:
While working on the decoupled front-end and formatting the whole files I found these clang-config updates to align with other files.

This commit makes clang-format break for each enum

Result in

```cpp
enum Status
{
    Invalid,
    Valid,
...
```

instead of

```cpp
enum Status { Invalid, Valid,...
```

It also breaks after each initializer in the constructor

Results in

```cpp
BAC::BAC(CPU *_cpu)
    : cpu(_cpu),
      bpu(params.branchPred),
```

instead of

```cpp

BAC::BAC(CPU *_cpu)
    : cpu(_cpu), bpu(params.branchPred),
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -55,6 +55,7 @@ BraceWrapping:
     AfterClass: true
     AfterStruct: true
     AfterUnion: true
+    AfterEnum: true
     AfterNamespace: true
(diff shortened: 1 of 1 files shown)


powerjg (approved):
lgtm. Thanks!

PR #2534, opened 2025-08-15
systemc: fix systemc daily test failure

erin-le:
The daily tests are failing due to the addition of https://github.com/gem5/gem5/pull/2484, which modified the arguments of the `begin()` function. A call to the `begin()` function in `util/systemc` was not updated for this, causing a build failure. This commit fixes this issue.

### DIFF
--- a/util/systemc/gem5_within_systemc/stats.cc
+++ b/util/systemc/gem5_within_systemc/stats.cc
@@ -87,7 +87,7 @@ void statsDump()
     statsEnable();
     statsPrepare();
 
-    output->begin();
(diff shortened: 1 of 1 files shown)


powerjg (approved):

BobbyRBruce:
Let's close this and merge #2535 instead. I pushed #2535 before seeing this PR had  but I think the solution is slightly better as it makes the argument optional. This was the original intention of PR #2484 and it doing it that way ensures backwards compatibility for this function if used elsewhere or extended by others elsewhere.

PR #2535, opened 2025-08-15
base: Make dump stats message optional

BobbyRBruce:
https://github.com/gem5/gem5/pull/2484 Updated the stats dump to include a message. This was intended to be optional but was implemented as a mandatory argument. This commit corrects this mistake.

This fixes the failing daily test: https://github.com/gem5/gem5/actions/runs/16974822600/job/48120913459

### DIFF
--- a/src/base/stats/hdf5.hh
+++ b/src/base/stats/hdf5.hh
@@ -67,7 +67,7 @@ class Hdf5 : public Output
     Hdf5(const Hdf5 &other) = delete;
 
   public: // Output interface
-    void begin(const std::string &message) override;
+    void begin(const std::string &message = "") override;
     void end() override;
     bool valid() const override;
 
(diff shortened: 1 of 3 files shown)


powerjg:
I don't think this is correct. I believe Giacomo requested that it be a mandatory parameter..I could be wrong... It would be good to go back and read the thread. I also think Erin submitted another change to fix this.

giactra:
> I don't think this is correct. I believe Giacomo requested that it be a mandatory parameter..I could be wrong... It would be good to go back and read the thread. I also think Erin submitted another change to fix this.

Thanks Jason. It was not a hard requirement. I thought the begin() method was not used outside of the dump method and was instead private to the class. I didn't consider checking in the systemc code. I think Bobby makes a good point about backwards compatibility which probably trumps my aversion to default arguments in both caller and callee

erin-le (approved):
I tested this by running the commands in the SystemC daily test; this fixes the daily test failure.

PR #2542, opened 2025-08-20
misc: split single line loop

clemdiep:
PR #2494 forced braces around control statement (if/else, for, do/while) and short if to be on multiple line.

This commit also force short loops to be on multiple line.

Before:

```c
for (auto it: vec) { it.doSomething(); }
```

After:

```c
for (auto it: vec) {
    it.doSomething();
}
```


powerjg (approved):
Thanks!

PR #2546, opened 2025-08-21
Toward removing main.py as a dependence

powerjg:
I am working toward enabling a more modular version of gem5 where we can link to python modules instead of using the gem5 binary with the marshalled python code. This is one of the many hurdles to overcome: specifically that main.py populates m5.options. This PR is the first step toward decoupling the stdlib from main.py and m5.options. Generally, this is cleaning up code and there should be no functional changes.

### DIFF
--- a/src/python/gem5/simulate/simulator.py
+++ b/src/python/gem5/simulate/simulator.py
@@ -653,3 +653,6 @@ class Simulator:
                                will be saved.
         """
         m5.checkpoint(str(checkpoint_dir))
+
+    def get_checkpoint_dir(self) -> Optional[Path]:
+        return self._board.get_checkpoint_dir()
--- a/src/python/gem5/simulate/simulator.py
+++ b/src/python/gem5/simulate/simulator.py
@@ -547,6 +547,7 @@ class Simulator:
             self._root = self._board._pre_instantiate(
                 full_system=self._full_system
             )
+            assert self._root is not None
 
             # m5.instantiate() takes a parameter specifying the path to the
             # checkpoint directory. If the parameter is None, no checkpoint
--- a/src/python/gem5/simulate/simulator.py
+++ b/src/python/gem5/simulate/simulator.py
@@ -101,7 +101,6 @@ class Simulator:
             ]
         ] = None,
         expected_execution_order: Optional[List[ExitEvent]] = None,
-        checkpoint_path: Optional[Path] = None,
         max_ticks: Optional[int] = m5.MaxTick,
         id: Optional[int] = None,
     ) -> None:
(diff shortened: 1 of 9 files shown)


src/python/m5/simulate.py
```diff
@@ -103,32 +90,75 @@ def instantiate(ckpt_dir=None):
...
+    dot_config: Optional[str] = None,
+):
+    # Use a slightly convoluted way to set these variables for backwards
+    # compatibility. Now, this function is no longer dependent on main.py and
+    # options
+    if outdir is None:
+        from m5 import options
+
+        outdir = options.outdir
+    if dump_config is None:
```
BobbyRBruce: Any particular reason we couldn't take this opportunity to rename this something like `ini_config`? Just to be a bit more explicit? From my understanding we dump 3 different versions of the configs. This is just one of them

src/python/gem5/simulate/simulator.py
```diff
@@ -101,9 +101,9 @@ def __init__(
             ]
         ] = None,
         expected_execution_order: Optional[List[ExitEvent]] = None,
-        checkpoint_path: Optional[Path] = None,
         max_ticks: Optional[int] = m5.MaxTick,
         id: Optional[int] = None,
+        outdir: Optional[str | Path] = None,
     ) -> None:
```
BobbyRBruce, line 107: I take it the intention here is that the user specifies the checkpoint path when you take the checkpoint. If so, I'd be careful here: I believes some exit events/hyper calls expect to take a checkpoint with the path already set. I believe this why this is in the constructor: It  enables someone to set up a default checkpoint path for these exit event handlers to dump stuff in.

BobbyRBruce (commented):
In addition to my comments, I'd also strongly suggest running some MultiSim runs to verify this doesn't break anything. I can't remember exactly how it all fits together, but I remember having to hack quite a lot of the `m5.outdir` and `main.py` code to get all the various gem5 subprocesses to output the right files to the right directories. Given you're moving this code around a bit, it'd be worthwhile checking that all still works. There are some examples in "configs/example/gem5_librabry".

src/python/gem5/simulate/simulator.py
powerjg: This is what's causing the CI tests to fail. I'm confused as to why we're dependent on this if it's deprecated.

BobbyRBruce: I think while this was in a deprecated state, we started using it for Multisim. In Multisim, you can't declare anything to be set up after instantiation for a given gem5 child process system, -- you have to do everything beforehand. So, this became a necessity. We should have taken it out of a deprecated state (or figured out a way around this problem we require it for), but I guess we lost track of it.

src/python/m5/simulate.py
powerjg: Done.

src/python/gem5/simulate/simulator.py
powerjg: I updated this a bit. The checkpoint_path seems to be part of the board now. I've added the needed APIs to expose it.

BobbyRBruce (approved):

PR #2548, opened 2025-08-21
mem, tests: Add SimObject unit tests for the TreePLRU replacement policy

erin-le:
This PR adds SimObject unit tests for the TreePLRU replacement policy. These tests are based on the SimObject unit tests for the FIFO replacement policy in https://github.com/gem5/gem5/pull/1977, and were mainly written to see how easy or difficult it would be to write SimObject unit tests based on PR 1977.

This PR also modifies the TreePLRU replacement policy itself to avoid the following error after running a certain number of unit tests:

```bash
src/tcmalloc.cc:1460] Object was not in-use
Aborted (core dumped)
```


erin-le:
@odanrc: Hello, I wrote these unit tests at @BobbyRBruce's behest to see how the framework in https://github.com/gem5/gem5/pull/1977 could be used to create new unit tests for other SimObjects. Could you give this PR a review when you have time? Thank you!

odanrc:
@erin-le Sure, I will try to do so over the weekend. FYI, I have local commits ready with tests for LRURP, SecondChanceRP, LFURP, and MRURP - they are just waiting for the FIFORP PR to be merged.

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <gtest/gtest.h>
+
+#include <cassert>
```
odanrc: Unused

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+    int num_leaves = 8;
+
+    TreePLRUTestF()
+    {
+        gem5::TreePLRURPParams params;
+        params.eventq_index = 0;
+        params.num_leaves = num_leaves;
+        rp = std::make_shared<gem5::replacement_policy::TreePLRU>(params);
+    }
+    TreePLRUTestF(int num_entries)
```
odanrc: Could you merge these 2 constructors with a default num_entries of 8?

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+#include <cassert>
+
+#include "mem/cache/replacement_policies/tree_plru_rp.hh"
+#include "params/TreePLRURP.hh"
+
+class TreePLRUTestF : public ::testing::Test
+{
+  public:
+    std::shared_ptr<gem5::replacement_policy::TreePLRU> rp;
+    int num_leaves = 8;
```
odanrc: Member variables should use camel case

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+TEST_F(TreePLRUTestF, InstantiatedEntry)
+{
+    const auto repl_data = rp->instantiateEntry();
+    ASSERT_NE(repl_data, nullptr);
+}
+
+/// Test that if there is one candidate and it is invalid, it will be
+/// the victim
+
+// I'm not sure that it makes sense to test this for TreePLRU.
```
odanrc: It makes sense, because it is a behavior we expect from all RP.

We also expect the behavior ´when all entries are invalid the first candidate will always be selected, regardless of the order of the invalidations´, and there is no test for that. Could you add it (it is in the FIFO tests too)?

Finally, could you remove this comment?

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+    void
+    instantiateAllEntries(void)
+    {
+        for (auto &entry : entries) {
+            entry.replacementData = rp->instantiateEntry();
+            candidates.push_back(&entry);
+        }
+    }
+    // We use 8 entries here
+    TreePLRUVictimizationTestF() : TreePLRUTestF(), entries(num_leaves)
```
odanrc: Ditto merge constructors

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+
+  public:
+    void
+    instantiateAllEntries(void)
+    {
+        for (auto &entry : entries) {
+            entry.replacementData = rp->instantiateEntry();
+            candidates.push_back(&entry);
+        }
+    }
```
odanrc, line 88: Strong nit: there should be space between functions to minimize the likelihood of merge conflicts

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+
+    rp->invalidate(entry.replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), &entry);
+}
+
+/// Fixture that tests victimization
+class TreePLRUVictimizationTestF : public TreePLRUTestF
+{
+  protected:
+    // The entries being victimized
```
odanrc: Nit: use 3 slashes to make it Doxygen compliant

Although since this is a test I am not sure anyone will even see the Doxygen doc...

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+        instantiateAllEntries();
+    }
+    TreePLRUVictimizationTestF(int num_entries)
+        : TreePLRUTestF(num_entries), entries(num_leaves)
+    {
+        instantiateAllEntries();
+    }
+};
+
+// Test only resetting one entry
```
odanrc: nit: Ditto Doxygen to all tests' docs

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+TEST_F(TreePLRUVictimizationTestF, GetVictimSingleResetLeftmost)
+{
+    rp->reset(entries[0].replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), &entries[4]);
+}
+
+// Reset entry H and entry A will be victimized.
+TEST_F(TreePLRUVictimizationTestF, GetVictimSingleResetRightmost)
+{
+    rp->reset(entries[7].replacementData);
```
odanrc, line 151: Could you add here the tree state after the reset?

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+    ASSERT_EQ(rp->getVictim(candidates), &entries[0]);
+}
+
+// Reset entry B and entry E will be victimized.
+TEST_F(TreePLRUVictimizationTestF, GetVictimSingleResetMiddle)
+{
+    rp->reset(entries[1].replacementData);
```
odanrc, line 164: Ditto tree state

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+    ASSERT_EQ(rp->getVictim(candidates), &entries[4]);
+}
+
+// Test resetting no entries. The tree's nodes should all have values of 0,
+// pointing toward entry A at index 0.
+
+TEST_F(TreePLRUVictimizationTestF, GetVictimNoReset)
```
odanrc: Nit: I would have put these tests before the 3 tests before, because it is simpler, and makes it easier to understand them

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+{
+    ASSERT_EQ(rp->getVictim(candidates), &entries[0]);
+}
+
+// Entries A, B, E, and F are reset, in that order. The victimized entry
+// should be C at index 2.
+
+TEST_F(TreePLRUVictimizationTestF, GetVictimHalfReset)
+{
+    rp->reset(entries[0].replacementData);
```
odanrc, line 196: nit: Could you add the tree states after each reset?

Strong nit: Could you add the tree state after the final reset?

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+    rp->reset(entries[1].replacementData);
+    rp->reset(entries[4].replacementData);
+    rp->reset(entries[5].replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), &entries[2]);
+}
+
+// Reset all entries once, starting from the leftmost side of the tree. The
+// victimized entry should be A, as it was the least recently used.
```
odanrc: Could you remove the 'as it was the least recently used' part? This is not causal. Counter example: ACDEFGHB victimizes E, despite A being the LRU.

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+    for (auto entry = entries.rbegin(); entry != entries.rend(); entry++) {
+        rp->reset((*entry).replacementData);
+    }
+
+    ASSERT_EQ(rp->getVictim(candidates), &entries[7]);
+}
+
+// Test that when there is at least a single invalid entry, it will be
+// selected during the victimization
+// Taken directly from the FIFO SimObject unit tests, introduced by
```
odanrc: I don't think this FIFO reference is needed. Could you remove it?

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+        }
+
+        // Set one of the entries as invalid
+        rp->invalidate(entry.replacementData);
+
+        ASSERT_EQ(rp->getVictim(candidates), &entry);
+    }
+}
+
+TEST_F(TreePLRUVictimizationTestF, TestTwoTrees)
```
odanrc, line 274: Could you document the goal of the test? Also, what is meant by *two* trees?

Note to myself: I need to review this test later

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+    ASSERT_EQ(rp->getVictim(second_candidates), &entries[12]);
+    ASSERT_NE(rp->getVictim(second_candidates), &entries[7]);
+}
+
+TEST_F(TreePLRUVictimizationTestF, TestMixedResetInvalidate)
+{
+    // If the entry is correctly invalidated, the entry at index 5 will be
+    // selected.
+    // If nothing happens when invalidate is called, the entry at index 4 will
+    // be selected.
```
odanrc: Nit: I would remove this comment, because we cannot 100% predict how bugged code will behave

I would also either remove the NE assertion against 4, or add this assertion against all non-5 entries.

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+    rp->reset(entries[0].replacementData);
+    rp->invalidate(entries[5].replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), &entries[5]);
+    ASSERT_NE(rp->getVictim(candidates), &entries[4]);
+
+    rp->reset(entries[1].replacementData);
```
odanrc, line 317: What is the goal of this part of the test?

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+}
+
+// Test the smallest tree size (2 leaves) and a large tree size (512 leaves)
+class SmallTreePLRUVictimizationTestF : public TreePLRUVictimizationTestF
+{
+  public:
+    SmallTreePLRUVictimizationTestF() : TreePLRUVictimizationTestF(2) {}
+};
+
+class LargeTreePLRUVictimizationTestF : public TreePLRUVictimizationTestF
```
odanrc, line 328: Nit: I am not sure this is useful. What is the goal here?

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+    rp->reset(entries[0].replacementData);
+    rp->invalidate(entries[5].replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), &entries[5]);
+    ASSERT_NE(rp->getVictim(candidates), &entries[4]);
+
+    rp->reset(entries[1].replacementData);
+    rp->invalidate(entries[1].replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), &entries[1]);
+}
+
```
odanrc, line 321: So far we have tested reset, invalidate and getVictim. There are no tests covering touch. Could you add them?

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+// Test the smallest tree size (2 leaves) and a large tree size (512 leaves)
+class SmallTreePLRUVictimizationTestF : public TreePLRUVictimizationTestF
```
odanrc, line 322: The smallest supported tree size is 1. Could you add a small test for it, like you did for 2?

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+    ASSERT_EQ(candidates.size(), 512);
+
+    rp->reset(entries[0].replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), &entries[256]);
+
+    rp->invalidate(entries[511].replacementData);
+    ASSERT_EQ(rp->getVictim(candidates), &entries[511]);
+}
+
+// Taken directly from the FIFO SimObject unit tests, introduced by
```
odanrc: Ditto remove

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,338 @@
...
+TEST_F(TreePLRUDeathTest, TouchNull)
+{
+    ASSERT_DEATH(rp->touch(nullptr), "");
+}
+
+TEST_F(TreePLRUDeathTest, NoCandidates)
+{
+    gem5::ReplacementCandidates candidates;
+    ASSERT_DEATH(rp->getVictim(candidates), "");
+}
```
odanrc, line 395: Add death test for num leaves <= 0.

odanrc (changes requested):

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
erin-le: The goal was to check a larger tree size, in case there were issues that wouldn't be detected with tree sizes of 2 and 8. I can remove the test for the larger tree if you don't think it's useful.

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
erin-le: I looked into this and I'm not sure it's currently supported. I tried implementing a test and found that it failed with an exception due to the PLRUTree size being 0, as the PLRUTree is instantiated with `numLeaves - 1` as its size. Is this something that should be fixed?

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
erin-le: The purpose was to check that there weren't any issues with calling `reset` and `invalidate` on the same entry. If you think this isn't useful or is redundant, please let me know.

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
erin-le: I might be misunderstanding the implementation of the TreePLRU replacement policy, but it looks like a new and separate PLRUTree is created each time enough entries have been created to fill up the previous tree. The "two trees" part of the test comes from the number of leaves per tree being 8, and initializing 16 entries in the test. The purpose of the test is to check that interacting with the entries of one tree doesn't affect the entries in the other tree. If you don't think this test is useful, or if there's something I'm misunderstanding, please let me know.

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
erin-le: I added one test, `CheckTouchResetSame`, to check that `touch()` and `reset()` have the same functionality. Could you let me know if this test is sufficient, or if I should add more tests for `touch()`?

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,402 @@
...
+    TreePLRUVictimizationTestF(int num_entries = 8)
+        : TreePLRUTestF(num_entries), entries(numLeaves)
+    {
+        instantiateAllEntries();
+    }
+};
+
+/// Test that when all entries are invalid the first candidate will always be
+/// selected, regardless of the order of the invalidations. The order of
+/// invalidations of this test is reversed in comparison to the other test.
```
odanrc: Which other test? I think this is an artifact from a copy paste

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,402 @@
...
+{
+    auto expected_victim = &entries.front();
+
+    // At this point all tree nodes should have values of 0, as no entries have
+    // ever been reset. The tree points toward the first entry.
+    ///    ____0____
+    ///  __0__   __0__
+    /// _0_ _0_ _0_ _0_
+    /// A B C D E F G H
+    ASSERT_EQ(rp->getVictim(candidates), expected_victim);
```
odanrc: This line is a duplicate of test GetVictimNoReset, so I suggest reordering these 2 tests and removing this line

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    for (auto entry = entries.rbegin(); entry != entries.rend(); entry++) {
+        rp->reset((*entry).replacementData);
+    }
+
+    ASSERT_EQ(rp->getVictim(candidates), &entries[7]);
+}
+
+/// In the definition of the TreePLRU replacement policy, `reset()` has the
```
odanrc: nit: In tests one shouldn't refer to code implementations, because code changes; one should write tests based on the expected behavior. For example, here someone could change the code so that touch calls reset instead of the other way around, and this comment would become outdated. I would simply say that touching and resetting should behave the same for this RP.

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
odanrc: Gotcha. Could you mark it as resolved?

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
odanrc: It should be good enough. You can mark this as resolved

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
odanrc: I don't really think it is important to fix (what is the point of a tree PLRU with a single entry?), but we should probably create a death test to show that it is not an unknown bug, but a feature

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
odanrc: I think that it is likely not useful (the tests with a size of 8 should be enough), but it is also not useless so you can leave it here if you want.

src/mem/cache/replacement_policies/tree_plru_rp.test.cc
odanrc: I see what you mean. You can mark it as resolved

odanrc (approved):

giactra:
Hi all, thanks for the test! I just wanted to provide this feedback. I believe I once encountered an error in the TrePLRU if the associativity of the cache was set to 1 (direct mapped). Would it be possible to add this test as well?

erin-le:
@giactra: Sorry for the delayed response. At one point, I did try to add a (normal) test for num_leaves = 1, but ran into some issues with it (see [this comment chain](https://github.com/gem5/gem5/pull/2548#discussion_r2295847454)) and added a death test instead (see [here](https://github.com/gem5/gem5/pull/2548/files#diff-c25e502afe6ff2d85469985ccf115c215edef02188666ed47f1480733425b9dcR408), at the bottom of the file).

BobbyRBruce (approved):

PR #2552, opened 2025-08-25
arch-arm: Make fplub.cc unused attribtue 'maybe_unused'

BobbyRBruce:
These functions appear to be unused and this is triggering a unused attribute error.


BobbyRBruce:
@giactra  These were added in #2490 but they appear to be unused. Is there a reason for this? I can remove them if they were included in error.

giactra (approved):
Approving this for now to properly compile. We should clean this up at a certain point

PR #2561, opened 2025-09-01
dev: add PCI to PCI bridge

### LINKED ISSUE #1862: Implement PCIe architecture model

clemdiep:
This commit brings PCI to PCI bridge into gem5.
    
A P2P bridge acts like an upstream to one or many PCI device (or other P2P bridge), like a PCI host bridge do. But unlike those, a P2P bridge must be connected to another upstream (host bridge or P2P bridge). They also have a type 1 PCI configuration header which gives bus number and addresses behind the bridge.

In this implementation, the class PciToPciBridge is added. It inherit from both PciType1Device and PciUpstream, implementing required functions. This is a simple implementation which redirect all request from downstream device to upper bridge. For this purpose, DeviceInterface class in PciUpstream has been extended for the function to be called with the PCI device and its bus number in arguments.

The PciToPciBridge Python configuration is similar to the one of PciHost.

### DIFF
--- a/src/dev/pci/device.cc
+++ b/src/dev/pci/device.cc
@@ -301,7 +301,14 @@ PciDevice::getAddrRanges() const
             ranges.push_back(bar->range());
     }
 
-    ranges.push_back(upstreamInterface->configRange());
+    // An invalid range can be returned if the upstream PCI to PCI bridge
+    // doesn't have a bus number yet. So check that to avoid adding it to the
+    // range list.
+    auto config_range = upstreamInterface->configRange();
+
+    if (config_range.valid()) {
+        ranges.push_back(config_range);
+    }
 
     return ranges;
 }
@@ -794,102 +801,135 @@ PciType1Device::writeConfig(PacketPtr pkt)
     }
 
     switch (pkt->getSize()) {
-      case sizeof(uint8_t):
-        switch (offset) {
-          case PCI1_PRI_BUS_NUM:
-            _config.type1.primaryBusNum = pkt->getLE<uint8_t>();
-            break;
-          case PCI1_SEC_BUS_NUM:
-            _config.type1.secondaryBusNum = pkt->getLE<uint8_t>();
-            break;
-          case PCI1_SUB_BUS_NUM:
-            _config.type1.subClassCode = pkt->getLE<uint8_t>();
-            break;
-          case PCI1_SEC_LAT_TIMER:
-            _config.type1.secondaryLatencyTimer = pkt->getLE<uint8_t>();
-            break;
-          case PCI1_IO_BASE:
-            _config.type1.ioBase = pkt->getLE<uint8_t>();
-            break;
-          case PCI1_IO_LIMIT:
-            _config.type1.ioLimit = pkt->getLE<uint8_t>();
-            break;
-          default:
-              panic("writing to a read only register");
-        }
-        DPRINTF(PciDevice,
+        case sizeof(uint8_t):
+            switch (offset) {
+                case PCI1_PRI_BUS_NUM:
+                    _config.type1.primaryBusNum = pkt->getLE<uint8_t>();
+                    break;
+                case PCI1_SEC_BUS_NUM:
+                    _config.type1.secondaryBusNum = pkt->getLE<uint8_t>();
+                    break;
+                case PCI1_SUB_BUS_NUM:
+                    _config.type1.subordinateBusNum = pkt->getLE<uint8_t>();
+                    break;
+                case PCI1_SEC_LAT_TIMER:
+                    _config.type1.secondaryLatencyTimer =
+                        pkt->getLE<uint8_t>();
+                    break;
+                case PCI1_IO_BASE:
+                    _config.type1.ioBase = pkt->getLE<uint8_t>();
+                    break;
+                case PCI1_IO_LIMIT:
+                    _config.type1.ioLimit = pkt->getLE<uint8_t>();
+                    break;
+                default:
+                    panic("writing to a read only register");
+            }
+            DPRINTF(
+                PciDevice,
                 "writeConfig: dev %#x func %#x reg %#x 1 bytes: data = %#x\n",
                 _devAddr.dev, _devAddr.func, offset,
                 (uint32_t)pkt->getLE<uint8_t>());
-        break;
-      case sizeof(uint16_t):
-        switch (offset) {
-          case PCI1_SECONDARY_STATUS:
-            _config.type1.secondaryStatus = pkt->getLE<uint16_t>();
             break;
-          case PCI1_MEM_BASE:
-            _config.type1.memBase = pkt->getLE<uint16_t>();
-            break;
-          case PCI1_MEM_LIMIT:
-            _config.type1.memLimit = pkt->getLE<uint16_t>();
-            break;
-          case PCI1_PRF_MEM_BASE:
-            _config.type1.prefetchMemBase = pkt->getLE<uint16_t>();
-            break;
-          case PCI1_PRF_MEM_LIMIT:
-            _config.type1.prefetchMemLimit = pkt->getLE<uint16_t>();
-            break;
-          case PCI1_IO_BASE_UPPER:
-            _config.type1.ioBaseUpper = pkt->getLE<uint16_t>();
-            break;
-          case PCI1_IO_LIMIT_UPPER:
-            _config.type1.ioLimitUpper = pkt->getLE<uint16_t>();
-            break;
-          case PCI1_BRIDGE_CTRL:
-            _config.type1.bridgeControl = pkt->getLE<uint16_t>();
-            break;
-          default:
-            panic("writing to a read only register");
-        }
-        DPRINTF(PciDevice,
+        case sizeof(uint16_t):
+            switch (offset) {
+                case PCI1_IO_BASE: {
+                    uint16_t ioConfig = pkt->getLE<uint16_t>();
+                    _config.type1.ioBase = ioConfig & mask(8);
+                    _config.type1.ioLimit = ioConfig >> 8;
+                } break;
+                case PCI1_SECONDARY_STATUS:
+                    _config.type1.secondaryStatus = pkt->getLE<uint16_t>();
+                    break;
+                case PCI1_MEM_BASE:
+                    _config.type1.memBase = pkt->getLE<uint16_t>();
+                    break;
+                case PCI1_MEM_LIMIT:
+                    _config.type1.memLimit = pkt->getLE<uint16_t>();
+                    break;
+                case PCI1_PRF_MEM_BASE:
+                    _config.type1.prefetchMemBase = pkt->getLE<uint16_t>();
+                    break;
+                case PCI1_PRF_MEM_LIMIT:
+                    _config.type1.prefetchMemLimit = pkt->getLE<uint16_t>();
+                    break;
+                case PCI1_IO_BASE_UPPER:
+                    _config.type1.ioBaseUpper = pkt->getLE<uint16_t>();
+                    break;
+                case PCI1_IO_LIMIT_UPPER:
+                    _config.type1.ioLimitUpper = pkt->getLE<uint16_t>();
+                    break;
+                case PCI1_BRIDGE_CTRL:
+                    _config.type1.bridgeControl = pkt->getLE<uint16_t>();
+                    break;
+                default:
+                    panic("writing to a read only register");
+            }
+            DPRINTF(
+                PciDevice,
                 "writeConfig: dev %#x func %#x reg %#x 2 bytes: data = %#x\n",
                 _devAddr.dev, _devAddr.func, offset,
                 (uint32_t)pkt->getLE<uint16_t>());
-        break;
-      case sizeof(uint32_t):
-        switch (offset) {
-          case PCI1_BASE_ADDR0:
-          case PCI1_BASE_ADDR1:
-            {
-              int num = PCI1_BAR_NUMBER(offset);
-              auto *bar = BARs[num];
-              _config.type1.baseAddr[num] = htole(
-                  bar->write(*upstreamInterface, pkt->getLE<uint32_t>()));
-              pioPort.sendRangeChange();
+            break;
+        case sizeof(uint32_t):
+            switch (offset) {
+                case PCI1_BASE_ADDR0:
+                case PCI1_BASE_ADDR1: {
+                    int num = PCI1_BAR_NUMBER(offset);
+                    auto *bar = BARs[num];
+                    _config.type1.baseAddr[num] = htole(bar->write(
+                        *upstreamInterface, pkt->getLE<uint32_t>()));
+                    pioPort.sendRangeChange();
+                } break;
+                case PCI1_PRI_BUS_NUM: {
+                    uint32_t busConfig = pkt->getLE<uint32_t>();
+                    _config.type1.primaryBusNum = busConfig & mask(8);
+                    _config.type1.secondaryBusNum = (busConfig >> 8) & mask(8);
+                    _config.type1.subordinateBusNum =
+                        (busConfig >> 16) & mask(8);
+                    _config.type1.secondaryLatencyTimer =
+                        (busConfig >> 24) & mask(8);
+                } break;
+                case PCI1_MEM_BASE: {
+                    uint32_t memConfig = pkt->getLE<uint32_t>();
+                    _config.type1.memBase = memConfig & mask(16);
+                    _config.type1.memLimit = memConfig >> 16;
+                } break;
+                case PCI1_PRF_MEM_BASE: {
+                    uint32_t prefConfig = pkt->getLE<uint32_t>();
+                    _config.type1.prefetchMemBase = prefConfig & mask(16);
+                    _config.type1.prefetchMemLimit = prefConfig >> 16;
+                } break;
+                case PCI1_PRF_BASE_UPPER:
+                    _config.type1.prefetchBaseUpper = pkt->getLE<uint32_t>();
+                    break;
+                case PCI1_PRF_LIMIT_UPPER:
+                    _config.type1.prefetchLimitUpper = pkt->getLE<uint32_t>();
+                    break;
+                case PCI1_IO_BASE_UPPER: {
+                    uint32_t ioUpperConfig = pkt->getLE<uint32_t>();
+                    _config.type1.ioBaseUpper = ioUpperConfig & mask(16);
+                    _config.type1.ioLimitUpper = ioUpperConfig >> 16;
+                } break;
+                case PCI1_ROM_BASE_ADDR:
+                    if (letoh(pkt->getLE<uint32_t>()) == 0xfffffffe) {
+                        _config.type1.expansionROM =
+                            htole((uint32_t)0xffffffff);
+                    } else {
+                        _config.type1.expansionROM = pkt->getLE<uint32_t>();
+                    }
+                    break;
+                default:
+                    panic("writing to a read only register");
             }
-            break;
-          case PCI1_PRF_BASE_UPPER:
-            _config.type1.prefetchBaseUpper = pkt->getLE<uint32_t>();
-            break;
-          case PCI1_PRF_LIMIT_UPPER:
-            _config.type1.prefetchLimitUpper = pkt->getLE<uint32_t>();
-            break;
-          case PCI1_ROM_BASE_ADDR:
-            if (letoh(pkt->getLE<uint32_t>()) == 0xfffffffe)
-                _config.type1.expansionROM = htole((uint32_t)0xffffffff);
-            else
-                _config.type1.expansionROM = pkt->getLE<uint32_t>();
-            break;
-          default:
-            panic("writing to a read only register");
-        }
-        DPRINTF(PciDevice,
+            DPRINTF(
+                PciDevice,
                 "writeConfig: dev %#x func %#x reg %#x 4 bytes: data = %#x\n",
                 _devAddr.dev, _devAddr.func, offset,
                 (uint32_t)pkt->getLE<uint32_t>());
-        break;
-      default:
-        panic("invalid access size(?) for PCI configspace!\n");
+            break;
+        default:
+            panic("invalid access size(?) for PCI configspace!\n");
     }
     pkt->makeAtomicResponse();
     return configDelay;
--- a/src/dev/pci/device.cc
+++ b/src/dev/pci/device.cc
@@ -293,13 +293,6 @@ AddrRangeList
 PciDevice::getAddrRanges() const
 {
     AddrRangeList ranges;
-    PciCommandRegister command = letoh(_config.common.command);
-    for (auto *bar: BARs) {
-        if (command.ioSpace && bar->isIo())
-            ranges.push_back(bar->range());
-        if (command.memorySpace && bar->isMem())
-            ranges.push_back(bar->range());
-    }
 
     // An invalid range can be returned if the upstream PCI to PCI bridge
     // doesn't have a bus number yet. So check that to avoid adding it to the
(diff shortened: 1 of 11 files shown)


src/dev/pci/p2p_bridge.hh
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+#ifndef __DEV_PCI_P2P_bridge_HH__
+#define __DEV_PCI_P2P_bridge_HH__
+
+#include "base/addr_range.hh"
+#include "dev/pci/p2p_device.hh"
+#include "dev/pci/types.hh"
+#include "dev/pci/upstream.hh"
+#include "params/PciToPciBridge.hh"
```
giactra: Nit: prefer forward declaration

giactra (commented):
Hi @clemdiep, thanks for iterating fast over the changes. I am looking forward to get the entirety of the PCIe implementation in.

As an initial comment, I would like to say I am bit concerned about the naming provided by the PR. I feel there is something unintuitive about having a

PciToPciBridge which **is-a** PciUpstream

and a

PciToPciDevice which **is-a** PciBridge

In my humble opinion, there should only be a single class, which is a PCI Bridge. This is consistent with the naming I found in the [Linux Device Driver book](https://static.lwn.net/images/pdf/LDD3/ch12.pdf).

Is there a reason for having a PciToPciDevice as a separate SimObject? Are we expecting users to provide a different one?

clemdiep:
> Is there a reason for having a PciToPciDevice as a separate SimObject? Are we expecting users to provide a different one?

As PciBridge inherits from PciDevice, that gives implementation for PCI configuration header, which share common parts with PciEndpoint (e.g. GPU). And then PciUpstream implements the bridge between the two PCI busses, like for PCI host bridge. Reuse of both avoid code duplication in their own ways.

So those two classes are childs of SimObject and can be independent, but PciToPciBridge needs the two to work but we cannot inherit from two SimObject class at the same time, or at least I don't know how.

For the naming, the idea was that PciToPciBridge is the main class used to add a bridge in Python configuration and it *automatically* adds the PciToPciDevice for the configuration header. Most of the time, users wouldn't need to use another PciToPciDevice, but should set vendor/device ID and the upstream bridge (host or another P2P). But in some rare(?) case, users would want a custom device, e.g., to add PCI capability.

I opened for renaming those classes or refactoring.

BobbyRBruce:
I'm having several issues with #2298 which is causing a large amount of our Daily (and one weekly) test to fail(daily tests can be found here: https://github.com/gem5/gem5/actions/workflows/daily-tests.yaml, weekly tests here: https://github.com/gem5/gem5/actions/workflows/weekly-tests.yaml). I've fixed a few of these issues with #2564 but I'm still encountering problems. Most notably I get the following:

```
src/dev/x86/pc.cc:117: warn: Don't know what interrupt to clear for console.
src/arch/x86/interrupts.cc:530: hack: Assuming logical destinations are 1 << id.
Assertion failed: (cmdBytesLeft == 0), function dmaWriteDone, file ide_disk.cc, line 547.
```

I believe the following should reproduce this issue:

```
/build/ALL/gem5.opt tests/gem5/x86_boot_tests/configs/x86_boot_exit_run.py --cpu timing --num-cpus 1 --mem-system mesi_two_level --dram-class DualChannelDDR3_1600 --boot-type init
```

I may have to revert #2298 if i can't fix these issues. Could you look into this?

clemdiep:
> I'm having several issues with #2298 which is causing a large amount of our Daily (and one weekly) test to fail(daily tests can be found here: https://github.com/gem5/gem5/actions/workflows/daily-tests.yaml, weekly tests here: https://github.com/gem5/gem5/actions/workflows/weekly-tests.yaml). I've fixed a few of these issues with #2564 but I'm still encountering problems. Most notably I get the following:
> 
> ```
> src/dev/x86/pc.cc:117: warn: Don't know what interrupt to clear for console.
> src/arch/x86/interrupts.cc:530: hack: Assuming logical destinations are 1 << id.
> Assertion failed: (cmdBytesLeft == 0), function dmaWriteDone, file ide_disk.cc, line 547.
> ```
> 
> I believe the following should reproduce this issue:
> 
> ```
> /build/ALL/gem5.opt tests/gem5/x86_boot_tests/configs/x86_boot_exit_run.py --cpu timing --num-cpus 1 --mem-system mesi_two_level --dram-class DualChannelDDR3_1600 --boot-type init
> ```
> 
> I may have to revert #2298 if i can't fix these issues. Could you look into this?

I will have a look to it yes!

src/dev/x86/Pc.py
```diff
@@ -92,7 +92,7 @@ class Pc(Platform):
     bad_addr = BadAddr(pio=default_bus.default)
 
     def attachIO(self, bus, dma_ports=[]):
-        self.south_bridge.attachIO(bus, self.pci_bus, dma_ports)
+        self.south_bridge.attachIO(bus, self.pci_bus)
```
powerjg, line 95: I believe this will break gem5 if you use Ruby.

clemdiep: This arguments isn't used anymore in SouthBridge.attachIO after #2298 (as only PCI device were connected to DMA) and wasn't removed.

But if you mean that this would break customs configurations, I can revert this change.

karnapathak: What was meant was if this breaks gem5 - X86 with Ruby Caches.
Having this
if dma_ports.count(self.pci_host.up_request_port()) == 0:
            bus.cpu_side_ports = self.pci_host.up_request_port()
it's fine. Tested it with x86. Bridge Works fine.

giactra:
> > Is there a reason for having a PciToPciDevice as a separate SimObject? Are we expecting users to provide a different one?
> 
> As PciBridge inherits from PciDevice, that gives implementation for PCI configuration header, which share common parts with PciEndpoint (e.g. GPU). And then PciUpstream implements the bridge between the two PCI busses, like for PCI host bridge. Reuse of both avoid code duplication in their own ways.
> 

Right. If I understood correctly, the main issue is in the PciBridge name.
Wouldn't something like PciType1Device a more intuitive name to avoid further confusion with the proliferation of PCI bridge names?

If you agree I would propose you to change the name as a preparatory PR before this one. I would also document the 
old PciBridge class to make sure it's clear even for non PCI experts what the class really tries to model.

The youn could rebase this PR on top of the new one

Thanks!

BobbyRBruce:
@clemdiep: The PCI changes introduced in [https://github.com/gem5/gem5/pull/2570](https://github.com/gem5/gem5/pull/2570) seem to be causing one of the daily tests to fail. The change causes the test to stall and timeout (You can see this issue in the following action run: https://github.com/gem5/gem5/actions/runs/18507902345, for example).

To reproduce this failure, you can run the following command on the `develop` branch:

```console
./build/ALL/gem5.opt configs/example/gem5_library/x86-ubuntu-run.py
```

After the system boots, it hangs. If you look at the `m5out/board.pc.com_1.device` file (which contains the simulated system's terminal output), you'll see it stall after the following error:

```console
[    1.098094] systemd[1]: Hostname set to <gem5>.
[    1.146765] netplan[103]: segfault at 0 ip 00007995216943e8 sp 00007ffe3c940640 error 4 in libglib-2.0.so.0.8000.0[79952166e000+a0000] likely on CPU 0 (core 0, socket 0)
[    1.146884] Code: fa 01 48 19 f6 83 e0 01 89 c2 0f b6 47 18 48 83 e6 e0 8d 0c 12 48 83 c6 40 83 e0 fc 09 d0 09 c8 88 47 18 31 ff e8 48 f4 01 00 <48> 8b 3b be 04 00 00 00 48 89 43 20 48 89 43 30 e8 e3 f4 01 00 48
[    1.160627] (sd-exec-[101]: /usr/lib/systemd/system-generators/netplan terminated by signal SEGV.
```

It seems like there's a segfault preventing the system from proceeding with initialization, which appears to have only appeared after the introduction of #2570.

Could you please take a look at this issue and come up with a fix?

Here's the full `board.pc.com_1.device` log for reference: [board.pc.com_1.device.txt](https://github.com/user-attachments/files/22949015/board.pc.com_1.device.txt)

clemdiep:
> > > Is there a reason for having a PciToPciDevice as a separate SimObject? Are we expecting users to provide a different one?
> > 
> > 
> > As PciBridge inherits from PciDevice, that gives implementation for PCI configuration header, which share common parts with PciEndpoint (e.g. GPU). And then PciUpstream implements the bridge between the two PCI busses, like for PCI host bridge. Reuse of both avoid code duplication in their own ways.
> 
> Right. If I understood correctly, the main issue is in the PciBridge name. Wouldn't something like PciType1Device a more intuitive name to avoid further confusion with the proliferation of PCI bridge names?

That's make sense yes. I will do that in a new PR and also rename `PciEndpoint` to `PciType0Device` for consistency !

clemdiep:
> @clemdiep: The PCI changes introduced in #2570 seem to be causing one of the daily tests to fail. The change causes the test to stall and timeout (You can see this issue in the following action run: https://github.com/gem5/gem5/actions/runs/18507902345, for example).
> 
> To reproduce this failure, you can run the following command on the `develop` branch:
> 
> ```
> ./build/ALL/gem5.opt configs/example/gem5_library/x86-ubuntu-run.py
> ```
> 
> After the system boots, it hangs. If you look at the `m5out/board.pc.com_1.device` file (which contains the simulated system's terminal output), you'll see it stall after the following error:
> 
> ```
> [    1.098094] systemd[1]: Hostname set to <gem5>.
> [    1.146765] netplan[103]: segfault at 0 ip 00007995216943e8 sp 00007ffe3c940640 error 4 in libglib-2.0.so.0.8000.0[79952166e000+a0000] likely on CPU 0 (core 0, socket 0)
> [    1.146884] Code: fa 01 48 19 f6 83 e0 01 89 c2 0f b6 47 18 48 83 e6 e0 8d 0c 12 48 83 c6 40 83 e0 fc 09 d0 09 c8 88 47 18 31 ff e8 48 f4 01 00 <48> 8b 3b be 04 00 00 00 48 89 43 20 48 89 43 30 e8 e3 f4 01 00 48
> [    1.160627] (sd-exec-[101]: /usr/lib/systemd/system-generators/netplan terminated by signal SEGV.
> ```
> 
> It seems like there's a segfault preventing the system from proceeding with initialization, which appears to have only appeared after the introduction of #2570.
> 
> Could you please take a look at this issue and come up with a fix?
> 
> Here's the full `board.pc.com_1.device` log for reference: [board.pc.com_1.device.txt](https://github.com/user-attachments/files/22949015/board.pc.com_1.device.txt)

I will have a look yes!

BobbyRBruce:
@clemdiep Is the latest patch in this PR meant to fix the bug I mentioned?

clemdiep:
> @clemdiep Is the latest patch in this PR meant to fix the bug I mentioned?

@BobbyRBruce, no I'm still trying to figure out what is going on exactly. I found out a temporary fix, that I will submit in a new PR and keep looking for the correct fix.

giactra:
@clemdiep What's the status of this PR? Does it need rebasing? Is it ready for a new round of review?

clemdiep:
> @clemdiep What's the status of this PR? Does it need rebasing? Is it ready for a new round of review?

I rebased to be sure I didn't forgot something. It is ready for a new round of review now!

src/dev/pci/device.cc
```diff
@@ -282,7 +282,13 @@ PciDevice::getAddrRanges() const
             ranges.push_back(bar->range());
     }
 
-    ranges.push_back(upstreamInterface.configRange());
+    // An invalid range can be returned if the upstream PCI to PCI bridge
+    // hasn't a bus number yet. So check that to avoid adding them to the range
+    // list.
+    auto configRange = upstreamInterface.configRange();
```
giactra, line 288: Nit: config_range

src/dev/pci/host.hh
```diff
@@ -43,6 +43,10 @@
 #include "dev/pci/upstream.hh"
 #include "params/PciHost.hh"
 
+#define PCI_FUNCTION_OFFSET 0
+#define PCI_DEVICE_OFFSET 3
+#define PCI_BUS_OFFSET 8
```
giactra, line 48: Nit: replace with constexpr

src/dev/pci/p2p_bridge.hh
```diff
@@ -0,0 +1,109 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __DEV_PCI_P2P_bridge_HH__
+#define __DEV_PCI_P2P_bridge_HH__
```
giactra, line 39: Nit: bridge should be uppercase

src/dev/pci/p2p_device.hh
```diff
@@ -0,0 +1,86 @@
...
+
+namespace gem5
+{
+
+struct PciToPciDeviceParams;
+
+/**
+ * PCI to PCI bridge device implementation for the config space.
+ */
+class PciToPciDevice : public PciType1Device
```
giactra, line 54: What is the need to add an extra class? What is the difference between a PciToPciDevice and a PciType1Device?

giactra (commented):
Hi @clemdiep, here an initial round of review. In general I am very concerned about the proliferation of abstractions within the code. I see new classes being added to the point that we have 4 or 5 classes which are covering the functionality of what should be a single bridge.

Is there a way to make things more intuitive and not overly complicate the overall design?

clemdiep:
> Hi @clemdiep, here an initial round of review. In general I am very concerned about the proliferation of abstractions within the code. I see new classes being added to the point that we have 4 or 5 classes which are covering the functionality of what should be a single bridge.
> 
> Is there a way to make things more intuitive and not overly complicate the overall design?

Hi @giactra,

I understand this worry. The reason for having multiple classes is because each PciDevice needs a PciUpstream (upstream attribute in Python config) to which it will register. This was inherited from the old PciHost before #2298. So the PciToPciBridge implements that PciUpstream, but it also needs to represent a PCI type 1 device which is implemented by PciType1Device.

PciToPciDevice inherits from it, I made this that way so that PciType1Device doesn't have more logic in it and allows for other type of inheritance (a bit like it's done with PciEndpoint) as I am not sure if other type of bridge could exist. But in fact, PciToPciDevice could be merge into PciType1Device to simplify.

Overall, we need to have two distincts classes (PciToPciBridge and PciType1Device) and can't merge them into one as they both inherit from SimObject.

But, as I kept working on some following steps, I got some problems with PciDevice having a reference to its related PciUpstream (in Python configuration) and had to remove the upstream attribute and add a devices attribute to PciUpstream which is a vector of PciDevice. This doesn't requires a lot of changes in the code as the PciDevice use the DeviceInterface and not the PciUpstream directly.

With that said, I now get the idea that we can simplify all this by totaly removing PciUpstream in Python and move their params directly into PciHost and PciType1Device. Then in C++, we can keep the PciUpstream and remove all SimObject related implementation in it and use it as a second inheritance for PciHost and PciType1Device. This would lead to minimal code duplication and remove the needs to have multiple classes for the PCI-PCI bridge.

Here is a small example of what it could look like:

```python
class PciHost(ClockedObject):
    up_to_down = Param.PciUpDownBridge(...)
    down_to_up = Param.BridgeBase(...)
    config_error = Param.PciConfigError(...)

    devices = VectorParam.PciDevice(...)

# Or PciToPciBridge
class PciType1Device(PciDevice):
    up_to_down = Param.PciUpDownBridge(...)
    down_to_up = Param.BridgeBase(...)
    config_error = Param.PciConfigError(...)

    devices = VectorParam.PciDevice(...)
```

```c++
class PciUpstream
{
    PciUpstream(devices)
    { // Register devices to give them their interface }

    // Common functions, same as what is now in PciUpstream
};

class PciHost : ClockedObject, PciUpstream
{
    PciHost(Param &p): ClockedObject(p), PciUpstream(p.devices)
    {}

    // Override needed function from PciUpstream
};

class PciType1Device : PciDevice, PciUpstream
{
    PciType1Device(Params &p): PciDeviec(p), PciUpstream(p.devices)
    {}

    // Override needed function from PciUpstream
};
```

Overall that change should not affect directly the code of classes that inherit from PciDevice, but will require to ensure that all upstream attribute are move to the devices attribute.

Let me know your through about it. I can make a more complete example in a new branch if you want.

Also I planned to add some functions in the Python classes to simplify connecting device to an upstream, which would connect the ports and add the device in the vector param. This would needs to be duplicated in bot PciHost and PciType1Device in this case.

clemdiep:
Hi @giactra,

What is your thought on my last message? Did you had time to consider it?

Let me know if you need some more explanation.

giactra:
> Hi @giactra,
> 
> What is your thought on my last message? Did you had time to consider it?
> 
> Let me know if you need some more explanation.

Hey, this message is just to push myself into reviewing it this week

giactra:
Hi @clemdiep. Apologies for the late reply. I have reviewed your proposed design and it looks like a great improvement. Looking forward to see a PR. I will make sure it gets a fast review

clemdiep:
Hello @giactra, I updated this PR to rebase on develop with the refactor of PciUpstream from #3208. You can review it!

src/dev/pci/upstream.hh
```diff
@@ -147,11 +149,31 @@ class PciUpstream
...
 
+        /**
+         * Post a PCI interrupt to the CPU.
+         * This function should be used by bridges to post interrupts
+         * from devices downstream of them.
+         *
+         * @param bus_num Bus number to which the device is connected.
+         * @param device PCI device posting the interrupts.
+         */
+        void postInt(PciBusNum bus_num, const PciDevice &device);
```
giactra: What's the reason for passing the bus_num? Isn't it already present in the upstream device?

giactra (commented):
Thanks @clemdiep for rebasing this PR over your latest changes. Do you happen to have a way to test this new feature? It would be nice if we could have a simple programmable PCI device

src/dev/pci/upstream.hh
clemdiep: Those functions are to receive requests from the devices behind a downstream bridge. This allows to forwards request to the host bridge. For `configRange` the actual bus number to which the device is connected, is used to get the correct addresses.

Let's say we have this simple hierarchy:

```
Host - Bus 1 -> P2P - Bus 2 -> Device
```

To get the config range, the device will ask P2P using `configRange` with no parameters. P2P will then forwards the request to Host using `configRange` with the two parameters.

This avoid to have the configuration address calcul at diffrent places.

For the other function, I'm not entirely sure if can be useful or not if other version of PCI host will be implemented. So I just added those to parameters for the function, so they have the same interface.

clemdiep:
> Do you happen to have a way to test this new feature? It would be nice if we could have a simple programmable PCI device

I added the preconfigured `IPcieRootPort` (which are PCI-to-PCI bridge) inspired by one on my computer as well as a example configuration that use it. That's part of what I used to test it that I can share. Is it ok for you, or do you want something else?

src/dev/pci/upstream.hh
giactra: > To get the config range, the device will ask P2P using configRange with no parameters. P2P will then forwards the request to Host using configRange with the two parameters.
For the other function, I'm not entirely sure if can be useful or not if other version of PCI host will be implemented. So I just added those to parameters for the function, so they have the same interface.

In my opinion we should try to minimize the IFC changes for other methods. I understand striving for symmetry but we should also aim at providing clear APIs that are intuitive to user/developers. If there's no plan to use the new parameter in the future, we should defer the change to a later time. Otherwise the motivation of the change will simply be forgotten.

In addition I suspect we don't have to change the APIs for configRange either. If all this is to relay the bus number all the way upstream to the PciHost:

```
AddrRange
PciToPciBridge::interfaceConfigRange(PciBusNum bus_num,
                                     const PciDevice &device) const
{
    if (!isDownstreamBus(bus_num)) {
        return AddrRange();
    }

    return upstreamInterface->configRange(bus_num, device);
}
```

Can't we just have the PciHost extract the PciBusNum from the device itself?
At the moment the PciBusNumber is only available to the Upstream class, but I see no reason why a device shouldn't be able to know its PciBusNumber itself.

Something like

```
class PciDevice
{

    void
    getBusNum()
    {
        upstreamInterface.getBusNum();
    }
```

The reason behind my proposal is not only to minimize changes. There are also the following design considerations:
1) Passing back to the Host both the handle to Device and its attached pci bus number feels a bit like duplication.
2) In this way we don't amend the interface *when/where it is not needed*. Effectively the API change is unused/meaningless in a flat single BUS topology. By allowing the PciBusNum to be extracted from the PciDevice, we make sure this gets used only within the P2P device

Please let me know if there's something I am missing

clemdiep: Thanks for the suggestion. I haven't considered this while implementing the bridge, and I find that it is a better option too.

I will update the code for this design!

erin-le:
@clemdiep: Hello, just wanted to ask what the status of this PR is on your end. This PR is currently on the milestone for the next gem5 release (v26.0), with the creation of the staging branch coming up next week. Could you let me know if you're aiming to get this PR into the release, and make changes soon if you are? Thanks!

clemdiep:
> @clemdiep: Hello, just wanted to ask what the status of this PR is on your end. This PR is currently on the milestone for the next gem5 release (v26.0), with the creation of the staging branch coming up next week. Could you let me know if you're aiming to get this PR into the release, and make changes soon if you are? Thanks!

Hey, I was waiting for @giactra to review my last changes based on his comments. Sorry, I forgot to mention it after making them 😬 

For me, the PR is ready. The failing test is due to it beeing unable to download the artifact of built gem5, and it doesn't seem that I can rerun it to ensure it pass.

giactra (approved):
Thanks for your changes

PR #2568, opened 2025-09-03
arch-arm, dev-arm: Replace params inclusion with forward declaration

giactra:
This PR is trying to avoid including the params header in the SimObject header file to minimize
header dependency and therefore minimize recompilation time. This is done via forward declaration.

Unless the situation requires it (e.g. in Gicv3 we have a templated method), there's no reason why
we shouldn't be moving any Param logic to the source file


powerjg (approved):

powerjg:
Seems like something an LLM should be able to do for the rest of the code base... Maybe I'll try to unleash gemini CLI on this over the weekend.

PR #2569, opened 2025-09-04
arch-arm,base: Fix ramdisk loading

rogerchang23424:
We encounter the issue when loading ramdisk in cpio.gz format. Because the data includes the gzip magic flag, the workload will try to decompress it before loading. However, it will lose some [cpio.gz](http://cpio.gz/) contents because cpio.gz might contain several files in it.
    
When loading ramdisk to memory by workload, we should disable data  decompression. The kernel can decompress then during execution.

Change-Id: Ic02add10c91ae7e3548819a870ff97e6385c7348

### DIFF
--- a/src/base/loader/image_file_data.hh
+++ b/src/base/loader/image_file_data.hh
@@ -53,7 +53,7 @@ class ImageFileData
     uint8_t const *data() const { return _data; }
     size_t len() const { return _len; }
 
-    ImageFileData(const std::string &f_name);
+    ImageFileData(const std::string &f_name, bool try_decompress = true);
     virtual ~ImageFileData();
 };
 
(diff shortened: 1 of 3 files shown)


src/base/loader/image_file_data.hh
```diff
@@ -53,7 +53,8 @@ class ImageFileData
     uint8_t const *data() const { return _data; }
     size_t len() const { return _len; }
 
-    ImageFileData(const std::string &f_name);
+    ImageFileData(const std::string &f_name) : ImageFileData(f_name, true) {}
+    ImageFileData(const std::string &f_name, bool try_decompress);
```
giactra: You could have a single constructor with default argument

    ImageFileData(const std::string &f_name, bool try_decompress=true);

giactra (changes requested):
Thanks! Could you provide a bit more of context on the commit/PR on why it is better to not decompress it?

giactra (approved):

PR #2570, opened 2025-09-04
dev: Fix PCI host bridge with no range from up

clemdiep:
PCI host bridge was intended to be connected to ports which returns a correct set of address ranges when calling getAddrRanges(). But RubyPort::MemResponsePort does return an empty list, making use of that function when connected to it, which is the case for system using Ruby
hierarchy.

This commit change the downstream to upstream bridge to be connected to the default port of PciBus. But, as that port was connected to an IsaFake to handle configuration access error, that one requires to be moved. So PciBus now contains a new port for that which will act like a second default port but only for the configuration ranges.

Also, as down to up bridge is now connected to default, a simple Bridge class can be used. And so PciOneWayBridge can be renamed to PciUpDownBridge and simplified as it doesn't require to remove undesired ranges responded by the down to up bridge.

### DIFF
--- a/src/dev/arm/RealView.py
+++ b/src/dev/arm/RealView.py
@@ -1129,7 +1129,8 @@ class VExpress_EMM(RealView):
 
     def _attach_pci(self, devices, bus, dma_ports=None):
         self.pci_bus.cpu_side_ports = self.pci_host.down_request_port()
-        self.pci_bus.mem_side_ports = self.pci_host.down_response_port()
+        self.pci_bus.default = self.pci_host.down_response_port()
+        self.pci_bus.config_error_port = self.pci_host.config_error.pio
 
         bus.mem_side_ports = self.pci_host.up_response_port()
         if dma_ports is None:
@@ -1526,7 +1527,8 @@ class VExpress_GEM5_Base(RealView):
 
     def _attach_pci(self, devices, bus, dma_ports=None):
         self.pci_bus.cpu_side_ports = self.pci_host.down_request_port()
-        self.pci_bus.mem_side_ports = self.pci_host.down_response_port()
+        self.pci_bus.default = self.pci_host.down_response_port()
+        self.pci_bus.config_error_port = self.pci_host.config_error.pio
 
         bus.mem_side_ports = self.pci_host.up_response_port()
         if dma_ports is None:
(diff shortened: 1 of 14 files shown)


clemdiep:
Hi @BobbyRBruce,

That PR fix the daily test errors you reported here https://github.com/gem5/gem5/pull/2561#issuecomment-3246802333.
I was able to run the test locally. It also include changes from your draft #2564.

abmerop (changes requested):
Any tips on how to fix this error with the GPU model?

```
src/mem/xbar.cc:436: fatal: board.pc.pci_bus has two ports responding within range [0xe0000:0xc000000000000000]:
        board.pc.pci_host.down_to_up.cpu_side_port
        board.gpus.device.pio
```

Since the GPU model is the only public PCI device besides IDE this should be fixed too.

clemdiep:
> Any tips on how to fix this error with the GPU model?
> 
> ```
> src/mem/xbar.cc:436: fatal: board.pc.pci_bus has two ports responding within range [0xe0000:0xc000000000000000]:
>         board.pc.pci_host.down_to_up.cpu_side_port
>         board.gpus.device.pio
> ```
> 
> Since the GPU model is the only public PCI device besides IDE this should be fixed too.

Well, it was *impossible* to fix that, so I completely changed my PR (actually prefere it now) and it should be good. I couldn't build the x86-ubuntu-gpu-ml to run the GPU model, but, I could reproduce that error with other images and now it is ok.

The PR description is updated to match the new implementation.

abmerop:
> > Any tips on how to fix this error with the GPU model?
> > ```
> > src/mem/xbar.cc:436: fatal: board.pc.pci_bus has two ports responding within range [0xe0000:0xc000000000000000]:
> >         board.pc.pci_host.down_to_up.cpu_side_port
> >         board.gpus.device.pio
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Since the GPU model is the only public PCI device besides IDE this should be fixed too.
> 
> Well, it was _impossible_ to fix that, so I completely changed my PR (actually prefere it now) and it should be good. I couldn't build the x86-ubuntu-gpu-ml to run the GPU model, but, I could reproduce that error with other images and now it is ok.
> 
> The PR description is updated to match the new implementation.

Ah ok, explains why I wasn't able to figure much out maybe :).  My smoke test is working now, so i'll unblock the change at least and give a closer look at some other time.

Sorry about the disk, you can let me know on the side what the issue was. I have been having error with subiquity (ubuntu auto installer) not working since yesterday.

powerjg (approved):
lgtm

PR #2574, opened 2025-09-06
Cleaning up m5.params

powerjg:
This PR does the following:
1. Creates an m5.params package and splits up the massive params.py file into multiple files.
2. Does some minor housekeeping on the params python files. These could use significantly more cleanup.
3. It moves the IntSource/SinkPins and ResetRequest/Response ports into params from src/dev. IMO, this is a more logical place, and it enables future tooling more easily

While working on my task to make gem5's targets more consistent, I noticed the IntSource/SinkPins and ResetRequest/ResponsePorts in src/dev. These were in a weird place, since they are not SimObjects, and the fact that they worked there was following the rest of gem5's codebase. To be honest, I couldn't follow how it was working. I wanted to move these to m5.params (where I think they fit more naturally), but it would have made an already bloated and difficult to maintain file worse. Thus, the refactoring of m5.params.


powerjg:
This is now ready for review. The CI tests should pass this time. Sorry for the spam.

odanrc (approved):

BobbyRBruce (approved):

BobbyRBruce (approved):

PR #2580, opened 2025-09-09
misc: Re-enable Address Sanitizer for GCC

rogerchang23424:
To enable Address Sanitizer for GCC, the PR changes:

1. Change some `RegClassOps` variables definition from `static inline` to `inline` to ensure they are not discarded
2. Fix some maybe-uninitalized warning issue

The PR build successfully with
```
scons build/ALL/gem5.opt --with-asan --without-tcmalloc
```

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -759,26 +759,20 @@ for variant_path in variant_paths:
     if GetOption('with_ubsan'):
         sanitizers.append('undefined')
     if GetOption('with_asan'):
-        if env['GCC']:
-            # Address sanitizer is not supported with GCC. Please see Github
-            # Issue https://github.com/gem5/gem5/issues/916 for more details.
-            warning("Address Sanitizer is not supported with GCC. "
-                    "This option will be ignored.")
-        else:
-            # Available for llvm >= 3.1. A requirement by the build system.
-            sanitizers.append('address')
-            suppressions_file = Dir('util').File('lsan-suppressions')\
-                                .get_abspath()
-            suppressions_opt = 'suppressions=%s' % suppressions_file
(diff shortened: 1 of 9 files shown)


abmerop (commented):
I would like to see this again for sure as I have a local commit with just the SConstruct changes and it works fine.

I believe most users are not compiling with asan and it is more of a developer feature, so I don't foresee this causing that many issues

powerjg (dismissed):

rogerchang23424:
Updated the PR according to gemini suggestion:

- This error occurs because `vecRegClassOps` is defined as a `static inline` variable in a header file (`src/arch/riscv/regs/vector.hh`). This creates a separate, private copy of the variable for each source file that includes the header.

- The compiler sees that in some source files, this variable is only used to initialize a `constexpr` variable (`vecRegClass`) at compile time. It then optimizes the `vecRegClassOps` variable away by placing it in a  section that the linker discards.

- However, another part of the code (in `build/ALL/arch/riscv/insts/vector.o`) needs to refer to `vecRegClassOps` at runtime. When the linker tries to resolve this reference, it finds that the symbol has been  discarded, leading to the relocation error.

- The solution is to change the variable definition from `static inline` to just `inline`. This is a C++17 feature that tells the linker to merge all instances of the variable into a single one in the final binary,  ensuring it is not discarded if it is used anywhere.

BobbyRBruce (approved):

rogerchang23424:
Hi @BobbyRBruce , @powerjg 

Do we need to add a compiler test for Address Sanitizer?

powerjg:
+1 on my side!

PR #2591, opened 2025-09-16
sim,cpu: Make SimObject imports consistent in SimObject files

powerjg:
In SimObject description files, you often need to import other SimObjects to use as parameter values. Note that these are only used as parameters, so we don't need the C++ backend at this point.
The problem is that when there is a SimObject description file with multiple SimObjects declared, it's hard to know what file you should import *from*. This change begins to make a consistent decision to import from m5.objects.<SimObjectName> instead of from the file that it was declared in. This works because all SimObjects end up in m5.objects.<SimObjectName> modules.

### DIFF
--- a/src/cpu/BaseCPU.py
+++ b/src/cpu/BaseCPU.py
@@ -43,13 +43,13 @@ import sys
 from m5.defines import buildEnv
 from m5.objects.ClockDomain import *
 from m5.objects.ClockedObject import ClockedObject
-from m5.objects.CPUTracers import ExeTracer
+from m5.objects.ExeTracer import ExeTracer
 from m5.objects.InstTracer import InstTracer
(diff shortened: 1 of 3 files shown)


powerjg:
There will probably be more of these as I go through the rest of the codebase.

BobbyRBruce:
Unittests failing with:

```shell
scons: *** [build/ALL/enums/ArmExtension.hh] Error 134
terminate called after throwing an instance of 'pybind11::error_already_set'
  what():  ModuleNotFoundError: No module named 'm5.objects.StubWorkload'
```

powerjg:
I need to revisit this implementation.

powerjg:
Closing because this was wrong. I misunderstood how to import things. You should import from the original module name.

PR #2592, opened 2025-09-16
dev: Fix PCI build

powerjg:
Clang 17.0 is very pedantic about class vs. struct. This change uses the header files instead of forward declarations. This change also removes an unused variable.


BobbyRBruce (approved):
I'll make sure the tests look good before merging but seems fine to me

PR #2595, opened 2025-09-17
Fdp-library

dhschall:
Two new components are added to the standard
library that makes integrating the decoupled
front-end easier:
- A processor that preconfigures a number or
  cores with the correct branch predictors
  and parameters for decoupling
- A cache hierarchy that appends the fetch-
  directed prefetcher to the L1I caches.

Additionally, a config file was added that
leverages the two new components

### DIFF
--- a/src/python/gem5/components/processors/decoupled_processor.py
+++ b/src/python/gem5/components/processors/decoupled_processor.py
@@ -1,0 +1,109 @@
+# Copyright (c) 2025 Technical University of Munich
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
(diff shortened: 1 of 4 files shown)


src/python/gem5/components/processors/decoupled_processor.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        super().__init__(
+            cores=[
+                SimpleCore(cpu_type=CPUTypes.O3, core_id=i, isa=isa)
+                for i in range(num_cores)
+            ]
+        )
+
+        for c in self.get_cores():
+            c.core
```
powerjg: I think this is a typo

powerjg (commented):
One small thing below

powerjg (approved):
Thanks!

PR #2599, opened 2025-09-18
arch-vega: Implement remaining non-MFMA CDNA4 instructions

abmerop:
Implement V_PRNG_B32, V_PERMLANE16_SWAP_B32, V_PERMLANE32_SWAP_B32, and "V_MFMA_LD_SCALE_B32." The last is a workaround for the 4x dword MFMA instructions as gem5 support only up to 2 dwords.

### DIFF
--- a/src/arch/amdgpu/vega/insts/vop1.cc
+++ b/src/arch/amdgpu/vega/insts/vop1.cc
@@ -2930,5 +2930,83 @@ namespace VegaISA
 
         vdst.write();
     } // execute
+    // --- Inst_VOP1__V_PERMLANE16_SWAP_B32 class methods ---
+
+    Inst_VOP1__V_PERMLANE16_SWAP_B32::Inst_VOP1__V_PERMLANE16_SWAP_B32(
+        InFmt_VOP1 *iFmt)
+        : Inst_VOP1(iFmt, "v_permlane16_swap_b32")
+    {
+        setFlag(ALU);
+    } // Inst_VOP1__V_PERMLANE16_SWAP_B32
+
+    Inst_VOP1__V_PERMLANE16_SWAP_B32::~Inst_VOP1__V_PERMLANE16_SWAP_B32()
+    {} // ~Inst_VOP1__V_PERMLANE16_SWAP_B32
+
+    // Swap data between two vector registers. Odd rows of the first operand
+    // are swapped with even rows of the second operand (one row is 16 lanes).
+    //
+    // Notes: ABS, NEG and OMOD modifiers should all be zeroed for this
+    // instruction. This instruction is useful for BFP data conversions.
+    void
+    Inst_VOP1__V_PERMLANE16_SWAP_B32::execute(GPUDynInstPtr gpuDynInst)
+    {
+        VecOperandU32 src(gpuDynInst, instData.SRC0);
+        VecOperandU32 vdst(gpuDynInst, instData.VDST);
+
+        src.read();
+        vdst.read();
+
+        // Ignores EXEC MASK
+        for (int pass = 0; pass < 2; ++pass) {
+            for (int lane = 0; lane < 16; ++lane) {
+                int dlane = pass * 32 + lane + 16;
+                int slane = pass * 32 + lane;
+
+                VecElemU32 tmp = src[slane];
+                src[slane] = vdst[dlane];
+                vdst[dlane] = tmp;
+            }
+        }
+
+        src.write();
+        vdst.write();
+    } // execute
+    // --- Inst_VOP1__V_PERMLANE32_SWAP_B32 class methods ---
+
+    Inst_VOP1__V_PERMLANE32_SWAP_B32::Inst_VOP1__V_PERMLANE32_SWAP_B32(
+        InFmt_VOP1 *iFmt)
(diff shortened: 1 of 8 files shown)


src/arch/amdgpu/vega/insts/vop1.cc
```diff
@@ -2885,5 +2885,110 @@ namespace VegaISA
...
+        : Inst_VOP1(iFmt, "v_prng_b32")
+    {
+        setFlag(ALU);
+    } // Inst_VOP1__V_PRNG_B32
+
+    Inst_VOP1__V_PRNG_B32::~Inst_VOP1__V_PRNG_B32()
+    {
+    } // ~Inst_VOP1__V_PRNG_B32
+
+    void
```
mattsinc, line 2909: Is it possible to put the comment above these (like Tony did previously) to indicate what the instruction is doing?

src/arch/amdgpu/vega/insts/vop1.cc
```diff
@@ -2885,5 +2885,110 @@ namespace VegaISA
...
+        Wavefront *wf = gpuDynInst->wavefront();
+        ConstVecOperandU32 src(gpuDynInst, instData.SRC0);
+        VecOperandU32 vdst(gpuDynInst, instData.VDST);
+
+        src.readSrc();
+
+        panic_if(isSDWAInst(), "SDWA not implemented for %s", _opcode);
+        panic_if(isDPPInst(), "DPP not implemented for %s", _opcode);
+
+        auto randFunc = [](VecElemU32 in) {
```
mattsinc, line 2921: please add a comment on this.

src/arch/amdgpu/vega/insts/vop1.cc
```diff
@@ -2885,5 +2885,110 @@ namespace VegaISA
...
+        VecOperandU32 src(gpuDynInst, instData.SRC0);
+        VecOperandU32 vdst(gpuDynInst, instData.VDST);
+
+        src.read();
+        vdst.read();
+
+        // Ignores EXEC MASK
+        for (int pass = 0; pass < 2; ++pass) {
+            for (int lane = 0; lane < 16; ++lane) {
+                int dlane = pass * 32 + lane + 16;
```
mattsinc, line 2962: comment here would be helpful too

mattsinc (dismissed):
In general this is fine, but some comments would help a lot

src/arch/amdgpu/vega/insts/vop1.cc
abmerop: The comment were not added by Tony as far as I know, they were auto-generated by whatever source is used to create the public ISA PDFs, I assume so they could just fill in the stubs manually without needing to go back and forth been the code and the manual. My issue with adding these comments is they provide no value. They are just pseudo-code for the instruction which can be seen in the C++ code.  I did go ahead and add the English text descriptions in the ISA manual though.

mattsinc: I'm not sure I understand -- I guess your point is someone can just go read the manual?

src/arch/amdgpu/vega/insts/vop1.cc
mattsinc: looks like this remains

src/arch/amdgpu/vega/insts/vop1.cc
mattsinc: and here

src/arch/amdgpu/vega/insts/vop1.cc
abmerop: No, my point is you can just read the C++ code which is no more complicated than a comment with pseudo-code.

```
    // --- description from .arch file ---
    // D.u = S0.u[23:0] * S1.u[23:0].
    void
    Inst_VOP2__V_MUL_U32_U24::execute(GPUDynInstPtr gpuDynInst)
    {
        auto opImpl = [](VecOperandU32& src0, VecOperandU32& src1,
                         VecOperandU32& vdst, Wavefront* wf) {
            for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
                if (wf->execMask(lane)) {
                    vdst[lane] = bits(src0[lane], 23, 0) *
                                 bits(src1[lane], 23, 0);
                }
            }
        };

        vop2Helper<ConstVecOperandU32, VecOperandU32>(gpuDynInst, opImpl);
    } // execute
```

I don't think we need a comment saying `// D.u = S0.u[23:0] * S1.u[23:0].` to know that `vdst[lane] = bits(src0[lane], 23, 0) * bits(src1[lane], 23, 0);` is doing `vdst = src0[23:0] * src1[23:0]` for each lane.

src/arch/amdgpu/vega/insts/vop1.cc
abmerop: This is described in the comment above the method

src/arch/amdgpu/vega/insts/vop1.cc
abmerop: This is described in the comment above the method

mattsinc (approved):

PR #2600, opened 2025-09-18
mem-cache: Prefetch for all cache blocks in a Fetch Target

rjc-arch:
This PR updates the Fetch Directed Prefetcher to prefetch for all cache 
blocks in each Fetch Target. The previous version only prefetched the 
first cache block in the fetch target. In many cases the full Fetch 
Target resides in a single cache line. However, for some Fetch Target 
sizes and alignments, the Fetch Target may span multiple consecutive 
cache lines, resulting in under-prefetching.

This update provides a small additional performance uplift for I-cache
bound workloads (up to ~2% for the workloads tested).

This PR also adds code to squash the Fetch Target associated with an
in-flight translation when the `notifyFTQRemove` probe is triggered
in the `FetchDirectedPrefetcher`. Previously, if the translation was
in progress the FT would not be squashed.

Since the translation mechanism is asynchronous in timing mode, the
FTs are marked as canceled and then dropped before progressing to the
prefetch queue in the `translationComplete` handler. Stats accounting
is performed in `notifyFTQRemove`.

### DIFF
--- a/src/mem/cache/prefetch/fdp.cc
+++ b/src/mem/cache/prefetch/fdp.cc
@@ -67,41 +67,47 @@ FetchDirectedPrefetcher::FetchDirectedPrefetcher(
 void
 FetchDirectedPrefetcher::notifyFTQInsert(const o3::FetchTargetPtr &ft)
 {
-    Addr blk_addr = blockAddress(ft->startAddress());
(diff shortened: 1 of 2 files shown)


dhschall (approved):
Thank you @rjc-arch! This looks good to me. 
Also, thanks for making the async squash of translations.

PR #2601, opened 2025-09-18
python: Add missing escape in param_types cxx_ini-parse func

BobbyRBruce:
'\\0' became '\0' in #2574. This caused daily tests to fail: https://github.com/gem5/gem5/actions/runs/17808599651/job/50626492040


powerjg (approved):
Sorry! Thanks for the quick fix!

PR #2608, opened 2025-09-22
 mem-cache: Cleanup headers and python imports in mem/cache

powerjg:


giactra (approved):

powerjg:
@BobbyRBruce any idea why "clang-fast-compilation" is getting cancelled after 30 minutes each time I run it? Maybe it's running out of memory?

BobbyRBruce:
@powerjg  : @powerjg : I honestly have no idea. I think I recreated this issue locally a couple of times, but it suddenly started working again. My guess is there's some flakiness. I re-ran the test here, and now it's passing.

BobbyRBruce:
@erin-le pointed out that we had a similar issue reported last year: #1722. It appears to be a (likely flaky) bug where gem5 stops compiling and no binary is generated. At the time of writing, I’m still not sure what the issue is.

PR #2609, opened 2025-09-23
stdlib: Define a CHI PrivateL1PrivateL2CacheHierarchy

giactra:
This is basically taking the existing CHI PrivateL1Hierarchy and extending it with a second level which is also private
per core.

The PR also amend the existing arm_boot_exit test to use the two level hierarchy

### DIFF
--- a/src/python/gem5/components/cachehierarchies/chi/nodes/l1_cache.py
+++ b/src/python/gem5/components/cachehierarchies/chi/nodes/l1_cache.py
@@ -37,7 +37,7 @@ from gem5.isas import ISA
 from .abstract_node import AbstractNode
 
 
-class PrivateL1MOESICache(AbstractNode):
+class L1CacheController(AbstractNode):
     def __init__(
         self,
         size: str,
--- a/src/python/gem5/components/cachehierarchies/chi/nodes/l2_cache.py
+++ b/src/python/gem5/components/cachehierarchies/chi/nodes/l2_cache.py
@@ -1,0 +1,113 @@
+# Copyright (c) 2021-2025 Arm Limited
+# All rights reserved.
+#
+# The license below extends only to copyright in the software and shall
+# not be construed as granting a license to any other intellectual
+# property including but not limited to intellectual property relating
+# to a hardware implementation of the functionality of the software
+# licensed hereunder.  You may use the software subject to the license
+# terms below provided that you ensure that this notice is replicated
+# unmodified and in its entirety in all distributions of the software,
+# modified or unmodified, in source code or in binary form.
+#
+# Copyright (c) 2021 The Regents of the University of California
+# All Rights Reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
+# this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from m5.objects import (
+    NULL,
+    ClockDomain,
+    RubyCache,
+    RubyNetwork,
+)
+
+from gem5.components.processors.abstract_core import AbstractCore
+from gem5.isas import ISA
+
+from .abstract_node import AbstractNode
+
+
+class L2Cache(RubyCache):
+    is_icache = False
+    dataAccessLatency = 6
+    tagAccessLatency = 2
+
+
+class L2CacheController(AbstractNode):
+    """
+    Strictly inclusive MOESI L2 controller.
+    Strictly inclusive:
+        - alloc_on_* = True
+        - dealloc_on* = False
+    MOESI:
+        - allow_SD = True
+    """
+
+    def __init__(
+        self,
+        size: str,
+        assoc: int,
+        network: RubyNetwork,
+        cache_line_size,
+        clk_domain: ClockDomain,
+    ):
+        super().__init__(network, cache_line_size)
+
+        self.sequencer = NULL
+        self.cache = L2Cache(
+            size=size, assoc=assoc, start_index_bit=self.getBlockSizeBits()
+        )
+
+        self.clk_domain = clk_domain
+        self.send_evictions = False
+        self.use_prefetcher = False
+        self.prefetcher = NULL
+
+        # Only applies to home nodes
+        self.is_HN = False
+        self.enable_DMT = False
+        self.enable_DCT = False
+
+        # Strict inclusive MOESI
+        self.allow_SD = True
+        self.alloc_on_seq_acc = False
+        self.alloc_on_seq_line_write = False
+        self.alloc_on_readshared = True
+        self.alloc_on_readunique = True
+        self.alloc_on_readonce = True
+        self.alloc_on_writeback = True
+        self.alloc_on_atomic = False
+        self.dealloc_on_unique = False
+        self.dealloc_on_shared = False
+        self.dealloc_backinv_unique = True
+        self.dealloc_backinv_shared = True
+
+        # Some reasonable default TBE params
+        self.number_of_TBEs = 32
+        self.number_of_repl_TBEs = 32
+        self.number_of_snoop_TBEs = 16
+        self.number_of_DVM_TBEs = 1  # should not receive any dvm
+        self.number_of_DVM_snoop_TBEs = 1  # should not receive any dvm
+        self.unify_repl_TBEs = False
--- a/src/python/gem5/components/cachehierarchies/chi/private_l1_cache_hierarchy.py
+++ b/src/python/gem5/components/cachehierarchies/chi/private_l1_cache_hierarchy.py
@@ -56,8 +56,8 @@ from gem5.utils.override import overrides
 
 from .nodes.directory import SimpleDirectory
 from .nodes.dma_requestor import DMARequestor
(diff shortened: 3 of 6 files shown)


src/python/gem5/components/cachehierarchies/chi/nodes/l1_cache.py
```diff
@@ -37,7 +37,7 @@
 from .abstract_node import AbstractNode
 
 
-class PrivateL1MOESICache(AbstractNode):
+class L1CacheController(AbstractNode):
```
powerjg, line 40: Why this change? I would think that it's better to define what states the cache supports in the name. I'm not opposed, and maybe the name wasn't great before, but it seems this class is more specific than just an L1 cache. I.e., you could imagine another implementation of an L1 that has different states supported.

src/python/gem5/components/cachehierarchies/chi/nodes/l2_cache.py
```diff
@@ -0,0 +1,104 @@
...
+from gem5.components.processors.abstract_core import AbstractCore
+from gem5.isas import ISA
+
+from .abstract_node import AbstractNode
+
+
+class L2Cache(RubyCache):
+    is_icache = False
+    dataAccessLatency = 6
+    tagAccessLatency = 2
```
powerjg, line 55: Is this needed? Why not set the parameter values below? If this class will be reused, it seems like a good idea. If it's only ever used on line 70, then we might as well set these values there.

src/python/gem5/components/cachehierarchies/chi/nodes/l2_cache.py
```diff
@@ -0,0 +1,104 @@
...
+        self.send_evictions = False
+        self.use_prefetcher = False
+        self.prefetcher = NULL
+
+        # Only applies to home nodes
+        self.is_HN = False
+        self.enable_DMT = False
+        self.enable_DCT = False
+
+        # Strict inclusive MOESI
```
powerjg, line 93: The fact this is a fully inclusive MOESI cache should either be in the name or in a docstring. This isn't a generic L2, it's one specific embodiment.

src/python/gem5/components/cachehierarchies/chi/private_l1_cache_hierarchy.py
```diff
@@ -154,7 +154,7 @@ def _create_core_cluster(
         for the core with a split I/D cache.
         """
         cluster = SubSystem()
-        cluster.dcache = PrivateL1MOESICache(
+        cluster.dcache = L1CacheController(
```
powerjg, line 157: I'm pretty surprised that the same CHI cache can act as both an L1 without any L2 and also act as an L1 with a fully inclusive private L2. But I guess if it works 🤷

powerjg (commented):
Generally, this change looks good to me. Just a couple of small naming/style comments, but nothing that's blocking.

One question: Other than being slightly more realistic, what is the need for this change?

src/python/gem5/components/cachehierarchies/chi/nodes/l1_cache.py
giactra: That's a great question. I addressed the motivation in the following commit message:

https://github.com/gem5/gem5/pull/2609/commits/8e1e3d7a1c83d232af22716c160d063c45716e70

Basically:

```
Two keywords were removed:

1) "Private": This is somehow aligning with other protocols
which are using Private only to describe the cache hierarchy
as a whole (for instance: private L1, shared L2).
Generally speaking it should not be a property of a cache
object per se; the same cache object could be instantiated
once per core (private) or shared amongst cores (shared)

2) "MOESI": This is proably more controversial. Effectively
the difference between MOESI and MESI is purely dictated
by the value of the allow_SD parameter.
There is no strong reason to rename it, but it is weird to
have a separate class/module for something which differs
from the other option by a single boolean value
```

But I am not a strong proponent for the renaming so let me know what you think

src/python/gem5/components/cachehierarchies/chi/nodes/l2_cache.py
giactra: I just thought it would be cleaner and modular this way, better than having everything embedded in a single class.

If later on we define a different L2 controller in the same module (a MESI?), we will reference the same L2Cache class (if we want).

src/python/gem5/components/cachehierarchies/chi/private_l1_cache_hierarchy.py
giactra: Really? We do stack L1+L2 in the default CHI implementation (in configs/ruby/CHI.py) and I am pretty sure the L1 controller https://github.com/gem5/gem5/blob/stable/configs/ruby/CHI_config.py#L249 is configured indentically as the one in stdlib (Will double check for curiosity just in case)

src/python/gem5/components/cachehierarchies/chi/nodes/l2_cache.py
giactra: Agree on the docstring; wrt the name, let's move the discussion over the renaming comment: https://github.com/gem5/gem5/pull/2609#discussion_r2372892905

giactra:
> Generally, this change looks good to me. Just a couple of small naming/style comments, but nothing that's blocking.
> 

Many thanks for the review!

> One question: Other than being slightly more realistic, what is the need for this change?

The main need is to define a L2 CacheController for CHI with the idea of dropping Ruby.py usage in this config script:
https://github.com/gem5/gem5/blob/stable/configs/example/arm/ruby_fs.py

so that we can finally deprecate and remove it

The plan would be to use gem5 components as much as possible. I will definitely be able to reuse nodes defined in the stdlib (Directory, L1Controller,L2Controller etc). I would love to use cache hierarchies in general which are IMHO the best part of stdlib.

However their reliance on the Board object might force me to avoid this option and write my own hierarchy in the config file. Until the point when the hierarchy object becomes a bit more modular and decoupled from other components

Hope this helps and happy to have a chat about it if you want

src/python/gem5/components/cachehierarchies/chi/nodes/l2_cache.py
powerjg: I'm OK either way.

src/python/gem5/components/cachehierarchies/chi/nodes/l1_cache.py
powerjg: Sorry, I didn't read the commit messages :).

So, I'm mixed on this. On one hand, it doesn't really matter. On the other hand, I think we're now mixing up "storage" details, "protocol" details, and "hierarchy" details.

In my mind, the behavior of a private L1 cache attached to a private is totally different than the behavior of a private L1 that is backed by a shared cache or directory. I would say that an L1 without an private L2 is more like the L2 in the privateL1/privateL2 hierarchy!

Storage wise, these are all just "caches" that have a size and an associativity. The storage is already mostly decoupled in the current design (though you could correctly argue that it's a pain if you want to change storage details like tag policy...).

"Protocol" wise, there are a huge number of different varieties of caches. They could send clean WBs (or not), allow modified data (or not), be part of cache-to-cache transfers, require back invalidates, be private or shared, etc. What is really cool about the CHI implementation that we have is that we can construct a cache to implement almost any reasonable protocol.

Finally, there's the "hierarchy" which is how we define where in the CPU to Memory hierarchy we want to position each kind of cache with different protocols. E.g., where to put private caches, where to make the point of coherence, whether to have inclusion or exclusion for different levels of caches, etc.

So, I don't like calling one particular protocol-hierarchy pair "L1CacheController" because you could imagine that same cache being used in different hierarchies. I feel like we could go in two directions:
1. We could try to make generic "protocol" caches (e.g., L1s that are MOESI or MESI that are inclusive or exclusive and the cross product thereof). Of course, this would get very unwieldy, and I'm not even sure it's totally possible to do.
2. We could couple the cache protocol specification with the entire cache hierarchy. Instead of trying to make "generic" cache controllers that could be used with multiple hierarchies, we could have some code duplication and make bespoke cache controllers for each hierarchy.

When I created `PrivateL1CacheHierarchy`, the latter (number 2) is what I was thinking. I didn't really think that we would be able to reuse cache protocol definitions for different hierarchies. I'll admit that for these super simple L1-only and private L1-private L2 we may be able to get away with the same parameter values, but I don't think that's going to work more generally.

Maybe I'm regretting putting a `nodes` directory :).

In general, my strongly held conviction is that we shouldn't be trying to be generic in the standard library, and we shouldn't try to build "configurable" things. Instead we should be building *extensible* things that we (and users) can use to create *specific* model instances. In other words, fewer parameters and more inheritance.

All this said... it's all personal preference. Changing this one object's name has little impact on the general design, and I don't really have very strong feelings (despite writing the novel).

powerjg:
> However their reliance on the Board object might force me to avoid this option and write my own hierarchy in the config file. Until the point when the hierarchy object becomes a bit more modular and decoupled from other components

Would it be possible to modify/extend the boards to work for your use case? If they don't work for you, I think that means we should improve or change the design of the boards rather than have multiple "competing" ways to set up systems.

src/python/gem5/components/cachehierarchies/chi/nodes/l1_cache.py
giactra: > Maybe I'm regretting putting a nodes directory :).

In my opinion you shouldn't! See comment below

> In general, my strongly held conviction is that we shouldn't be trying to be generic in the standard library, and we shouldn't try to build "configurable" things. Instead we should be building extensible things that we (and users) can use to create specific model instances. In other words, fewer parameters and more inheritance.

Paradoxically the renaming is better suited for what you are actually proposing. In a way, it feels like doing this (extending the component)

```
MyL1MOESI(L1CacheController):
    allow_SD = True

MyMESI(L1CacheController):
    allow_SD = False
```

Is more intuitive than

```
MyMESI(PrivateL1MOESI):
    allow_SD = False
```

This is even more true when it comes to the node directory. It is useful that the stdlib exposes the building blocks to the user as that is the best way to provide extendability. Let's imagine we don't do that and we only provide the hierarchy box to the user.
There's a lot of things going on inside the hierarchy object. How is the user supposed to extend the class to amend for example simply the behaviour of the L2 controller?

It would be unreasonable for a user to define a parent class and work their way through inherithance. It is more likely they would simply take the hierarchy and configure the parameter 

`board.hierarchy.l2.allow_SD = False`

I believe the stdlib should provide hierarchical modularity in the sense that it should provide to users the chance to pick individual bricks (for higher configurability) and collection of bricks (for simpler configuration scripts). That said, this could be a hard task and we shouldn't be too naive about how extensible the entire code would be.

powerjg: > Paradoxically the renaming is better suited for what you are actually proposing. In a way, it feels like doing this (extending the component)

Ah, yes. This makes sense. I agree with your point here.

> It would be unreasonable for a user to define a parent class and work their way through inherithance. It is more likely they would simply take the hierarchy and configure the parameter

If it were me, I would inherit, but I can understand why people coming from other backgrounds may choose to set the parameter directly. (It's better than a command line parameter, though!)

> I believe the stdlib should provide hierarchical modularity in the sense that it should provide to users the chance to pick individual bricks (for higher configurability) and collection of bricks (for simpler configuration scripts). That said, this could be a hard task and we shouldn't be too naive about how extensible the entire code would be.

Hard agree here! I think it's impossible to actually provide all possible needed bricks, so we need to be careful to make sure that users know how to "build their own bricks" in some sense.

I'm happy to drop my comment on the name change here :)

src/python/gem5/components/cachehierarchies/chi/nodes/l2_cache.py
```diff
@@ -0,0 +1,104 @@
...
+from .abstract_node import AbstractNode
+
+
+class L2Cache(RubyCache):
+    is_icache = False
+    dataAccessLatency = 6
+    tagAccessLatency = 2
+
+
+class L2CacheController(AbstractNode):
```
powerjg, line 58: If you add a docstring here describing what this object is meant to model (i.e., MOESI), I'll approve.

It would be great to also mention how it's intended (and not intended) to be used as well.

powerjg (commented):
My "last word" on the philosophical point here is that I don't think we should be obsessed with eliminating all code duplication. It is a bit of a pain when we change things to find all of the different places we need to update sometimes, but I think that is outweighed by the benefit of a simpler design in many cases.

src/python/gem5/components/cachehierarchies/chi/nodes/l2_cache.py
giactra: Done!

powerjg (approved):

PR #2614, opened 2025-09-26
stdlib: Improvements for VIPER board

abmerop:
Going forward it would be nice to only have to maintain one set of configs for GPU simulations. The current MI355X model will be the first device supported only by standard library. To enable more simulation however, some improvements to the VIPER board are required. First, we need more than 3GiB of memory for many if not all interesting workloads. Second, the configs can be written quicker by removing some required parameters that are not useful for GPU simulation such as CPU cache sizes. Last, add a command line option for command line options to pass to the GPU application in the example config as most applications use command line options. This series of commits addresses all of these issues.

### DIFF
--- a/src/python/gem5/prebuilt/viper/board.py
+++ b/src/python/gem5/prebuilt/viper/board.py
@@ -32,11 +32,19 @@ import os
 from typing import (
     List,
     Optional,
+    Sequence,
+    Tuple,
 )
 
+from m5.objects import X86E820Entry
+from m5.params import (
+    AddrRange,
+    Port,
+)
 from m5.util import warn
 
 from ...components.boards.abstract_board import AbstractBoard
+from ...components.boards.abstract_system_board import AbstractSystemBoard
 from ...components.boards.kernel_disk_workload import KernelDiskWorkload
 from ...components.boards.x86_board import X86Board
 from ...components.cachehierarchies.abstract_cache_hierarchy import (
@@ -44,6 +52,7 @@ from ...components.cachehierarchies.abstract_cache_hierarchy import (
 )
 from ...components.devices.gpus.amdgpu import BaseViperGPU
 from ...components.memory.abstract_memory_system import AbstractMemorySystem
+from ...components.memory.single_channel import SingleChannelDDR4_2400
 from ...components.processors.abstract_processor import AbstractProcessor
 from ...utils.override import overrides
 
@@ -52,7 +61,16 @@ class ViperBoard(X86Board):
     """
     A derivative of X86Board capable of full system simulation for X86 with a
     GPU device. Provides all the functionality of the X86Board with helper
-    methods specific to booting a disk with GPU libraries installed.
+    methods specific to booting a disk with GPU libraries installed, enables
+    AVX on supported CPU models (KVM only right now) for PyTorch, a workaround
+    for X86Board 3GiB memory limitation, and helper methods to create a
+    temporary readfile which reads in a binary blob into the simulator.
+
+    **Limitations**
+    * Only KVM and ATOMIC CPU models can be used due to issue CPU Ruby protocol
+    * 3GiB limit workaround is a VIPER specific hack which silently creates an
+      additional memory controller with 128MiB of additional memory.
+    * This board should not be used to measure CPU performance.
     """
 
     def __init__(
@@ -68,6 +86,15 @@ class ViperBoard(X86Board):
             processor=processor,
             memory=memory,
             cache_hierarchy=cache_hierarchy,
+        )
+
+        # Low memory for CPU side which allows the OS to boot. The rest of
+        # main memory is then mapped above 4GiB. The memory type does not
+        # matter as it is on the CPU side and should not be used much.
+        low_mem_size = "128MiB"
+        self.low_mem = SingleChannelDDR4_2400(size=low_mem_size)
+        self.low_mem.mem_ctrl[0].dram.range = AddrRange(
+            start=0x0, size=low_mem_size
         )
 
         self.gpus = gpus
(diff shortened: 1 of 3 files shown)


src/python/gem5/prebuilt/viper/board.py
```diff
@@ -32,18 +32,27 @@
 from typing import (
     List,
     Optional,
+    Sequence,
+    Tuple,
 )
 
+from m5.objects import (
+    AddrRange,
+    Port,
```
powerjg, line 42: These are not in m5.objects, they are in m5.params. Can you update? I have been going through and fixing all of the other files.

src/python/gem5/prebuilt/viper/board.py
```diff
@@ -70,6 +79,15 @@ def __init__(
             cache_hierarchy=cache_hierarchy,
         )
 
+        # Low memory for CPU side which allows the OS to boot. The rest of
+        # main memory is then mapped above 4GiB. The memory type does not
+        # matter as it is on the CPU side and should not be used much.
```
powerjg, line 93: This is a reasonable *hack*. However, since it's a hack, we should document it clearly in the docstring for the board. This works OK for the GPU board since we don't really care about the CPU/memory performance. But, if anyone really cared about the CPU performance, then great care would need to be taken with this extra memory controller.

src/python/gem5/prebuilt/viper/board.py
```diff
@@ -155,6 +176,58 @@ def get_default_kernel_args(self) -> List[str]:
...
+        return [(ctrl.dram.range, ctrl.port) for ctrl in self.low_mem.mem_ctrl]
+
+    @overrides(AbstractSystemBoard)
+    def _setup_memory_ranges(self):
+        # This is exactly the same code as x86_board.py but with the memory
+        # size >3GiB exception removed.
+        memory = self.get_memory()
+
+        data_range = AddrRange(start=0x100000000, size=memory.get_size())
+        memory.set_memory_range([data_range])
```
powerjg, line 198: Just to be clear, this means that the board has `memory.get_size() + 128MiB` of capacity?

powerjg (commented):
Just a few small things below. Otherwise, it looks good to me.

src/python/gem5/prebuilt/viper/board.py
abmerop: Done

src/python/gem5/prebuilt/viper/board.py
abmerop: Yea hopefully we can revisit #1581 but break it up into two changes: 1) Support for holes 2) Non-power two memory channel support.

For now changed the docstring.

src/python/gem5/prebuilt/viper/board.py
abmerop: The board does, yes, and `free -m` will report something like 4.2 GiB for 4 GiB in the top level config.  The AbstractMemorySystem (I think this is the base) is completely unaware of the extra memory controller. Meaning, when `get_memory()` / `set_memory_range()` are called they only know about the user specified amount in the top level config

src/python/gem5/prebuilt/viper/board.py
powerjg: Perfect. Thanks!

src/python/gem5/prebuilt/viper/board.py
powerjg: Makse sense to me. Thanks!

powerjg (approved):

michaelwcl:
So there won't be any changes to allow Timing CPUS to use more than 3GB RAM for X86 boards? I am asking because I have CPU workloads Im looking to measure the performance that demand more than 3GB of RAM.

abmerop:
> So there won't be any changes to allow Timing CPUS to use more than 3GB RAM for X86 boards? I am asking because I have CPU workloads Im looking to measure the performance that demand more than 3GB of RAM.

The issue with timing CPUs is that the MOESI_AMD_Base protocol is broken and requires backing store to function.  The best way to get timing CPUs to work in the GPU model would be one of the following:
1) New build which build one of the other CPU protocols + VIPER GPU protocol (e.g., GFX9_X86 instead of VEGA_X86).  This would allow stdlib boards to use MESI protocol on the the CPU side, for example, and the GPU side would continue to use VIPER.  Probably the easiest route but non-stdlib configs would not work.
2) Same as above, but wait for a PR which allows for all Ruby protocols to be build in the ALL build.  This was proposed a long time ago but probably no longer being worked on.  It is a lot of work and has the same limitations as above regarding only stdlib boards working.
3) Integrate a different CPU protocol with VIPER and remove MOESI_AMD_Base.  This would be a lot of work but both stdlib and non-stdlib configs will work.  It would require a near complete rewrite of the directory SM which is shared between CPU and GPU.

abmerop:
> So there won't be any changes to allow Timing CPUS to use more than 3GB RAM for X86 boards? I am asking because I have CPU workloads Im looking to measure the performance that demand more than 3GB of RAM.

For non-GPU simulations you could probably apply the changes in 9563e9f4a6f6d296c37e292c807e98679a6ab705 to x86_board.py and one of the cache protocols to get a CPU-only board running with >3GiB.

michaelwcl:
> > So there won't be any changes to allow Timing CPUS to use more than 3GB RAM for X86 boards? I am asking because I have CPU workloads Im looking to measure the performance that demand more than 3GB of RAM.
> 
> For non-GPU simulations you could probably apply the changes in [9563e9f](https://github.com/gem5/gem5/commit/9563e9f4a6f6d296c37e292c807e98679a6ab705) to x86_board.py and one of the cache protocols to get a CPU-only board running with >3GiB.

I appreciate the advice. I tried this, and now I seem to be running into issues with memory deadlocking (I assume thats why the viper board will only allow KVM/atomic CPU). I'll keep looking into it however.

Edit: I assume the deadlocking issues stem from the MOESI protocol being broken?

PR #2623, opened 2025-10-01
arch-arm: Decouple insts from generated decoder

powerjg:
To compile the insts/ files separately from the generated files there was a circular dependence between the generated decoder.hh and sve_macromem.hh that had to be broken. decoder.hh defines the classes SveGatherLoadCpySrcVecMicroop and SveGatherLoadCpySrcVecMicroop which were used with `new` in the header file. Also, decoder.hh depends on all insts/*.hh files.

To break this dependence, we need to move the constructor implementations into a cc file. However, the classes are template classes so we also need to specify all of the uses of the templates. I believe these are fixed, so they are manually added to sve_macromem.cc.


powerjg:
@giactra let me know what you think on this. I don't *love* this solution, but I believe it's the best way to break this circular dependence.

Please review at least semi-carefully. Gemini did (almost) all the work.

(Note: More changes will be coming for circular dependences between arch/arm and arch/dev/arm once I get things to compile...)

erin-le:
@giactra: Hello, could you take a look at this when you get a chance? Thank you!

powerjg:
@giactra Before I take the time to rebase this on your recent changes, could you let me know if you believe this is a reasonable direction?

giactra:
> @giactra Before I take the time to rebase this on your recent changes, could you let me know if you believe this is a reasonable direction?

Apologies @powerjg.

Other than the impractical idea of disallowing ISA micro-ops (basically moving them to C++ world), the other solution could be to break down the gigantic decoder files into multiple header/source files. In this way for instance

generated-sve.hh -> includes macromem_sve.hh -> includes generated-micro-ops.hh (unlikely this includes the generated-sve.hh back)

Needless to say, while I believe this would be the best long term strategy (compilation time would also improve), I'd be happy to accept a solution like yours... Let me know your thoughts. I honestly don't know how easy/hard it would be to split the decoder files.. Last time I look into the ISA parser it seemed like there was already some logic inside so it should be possible...

powerjg:
> The other solution could be to break down the gigantic decoder files into multiple header/source files.

Yeah, this would solve part of the problem, but it isn't very practical right now.

The goal here is to simply remove `#include "decoder.hh"` from any header files. This way, the decoder can depend on the instruction headers, but they don't depend on the decoder.

This is the only file that has that dependence, so breaking it isn't so bad. The templates made it a bit of a pain, but I think I figured it out.

It's ready for review now. It's been rebased, and I fixed the formatting errors.

powerjg:
@giactra can I have a review on this PR?

powerjg:
@giactra ping :)

giactra (approved):
> The goal here is to simply remove #include "decoder.hh" from any header files. This way, the decoder can depend on the instruction headers, but they don't depend on the decoder.

Yeah, I am going to approve it for now as it is relatively minimal. I feel though we should consider breaking the decoder into multiple files is the real modular solution in the long run...

BobbyRBruce:
> ## Pull Request Overview
> This PR successfully decouples the ARM instruction definitions in `insts/` from the generated decoder by breaking a circular dependency between `decoder.hh` and `sve_macromem.hh`. The key changes involve moving template class constructor implementations from the header to a new source file and providing explicit template instantiations.
> 
> * Removed `#include "arch/arm/generated/decoder.hh"` from `sve_macromem.hh` and moved it to the new `sve_macromem.cc` file
> * Converted three template class constructors from inline definitions to declarations in the header, with implementations in the new source file
> * Added explicit template instantiations for all required type combinations in the new source file
> 
> ### Reviewed Changes
> Copilot reviewed 3 out of 3 changed files in this pull request and generated 4 comments.
> 
> File	Description
> src/arch/arm/insts/sve_macromem.hh	Removed decoder.hh include dependency and converted constructor implementations to declarations
> src/arch/arm/insts/sve_macromem.cc	New file containing moved constructor implementations and explicit template instantiations
> src/arch/arm/SConscript	Added new source file to the build configuration
> 💡 [Add Copilot custom instructions](/gem5/gem5/new/stable/.github/instructions?filename=*.instructions.md) for smarter, more guided reviews. [Learn how to get started](https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot).

Just a couple of typos that aren't worth re-running the tests for. All looks good to me

PR #2630, opened 2025-10-02
mem-cache: Add unit test for MRU RP

odanrc:
Add unit tests for the MRU replacement policy.

### DIFF
--- a/src/mem/cache/replacement_policies/mru_rp.test.cc
+++ b/src/mem/cache/replacement_policies/mru_rp.test.cc
@@ -1,0 +1,257 @@
+/**
+ * Copyright (c) 2024 Daniel R. Carvalho
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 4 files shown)


src/mem/cache/replacement_policies/mru_rp.test.cc
```diff
@@ -0,0 +1,256 @@
+/**
+ * Copyright (c) 2024 Daniel R. Carvalho
```
erin-le: I don't think it matters too much, but did you want to update the copyright year?

erin-le (dismissed):
@odanrc: LGTM. For the failing `clang-format-check` test, could you run the following command and push the changes? It's just what the CI test itself uses to check the formatting.

```bash
python util/run-git-clang-format.py --ci-pr-base-commit d24ee0d969f41d343177c5f44d3204d1ee651ff1
```

src/mem/cache/replacement_policies/mru_rp.test.cc
odanrc: I forgot to update this, but I haven't touched this code since last year anyway.

src/mem/cache/replacement_policies/mru_rp.test.cc
```diff
@@ -0,0 +1,257 @@
...
+    std::vector<gem5::ReplaceableEntry> entries;
+
+    // The entries, in candidate form
+    gem5::ReplacementCandidates candidates;
+
+  public:
+    // The number of entries is arbitrary. It does not need to be high, since
+    // having more entries is not expected to increase coverage
+    MRURPVictimizationTestF() : MRURPTestF(), entries(4)
+    {
```
odanrc, line 105: @erin-le , I squashed the clang format fix, as it was just moving this brace to the next line

src/sim/serialize.test.cc
```diff
@@ -811,7 +811,7 @@ TEST_F(SerializeFixture, OptParamOutIn)
 /** Check for death when there is no section and paramIn is requested. */
 TEST_F(SerializableFixtureDeathTest, NoSectionParamIn)
 {
-#ifndef NDEBUG
+#ifdef NDEBUG
```
odanrc, line 814: I remembered that the FIFO tests had issues in fast mode, so I added the GTEST_SKIP to the MRU tests and ran in fast mode. This issue popped up while running, so I fixed it in this PR already, since it was small.

erin-le: I tried running this test before and after this change, and it looks like this test succeeds for fast builds with or without the guard. I think it's still good to fix the guard, though.

erin-le (approved):

PR #2632, opened 2025-10-04
arch-arm: Add support for LRCPC instructions

pranith:
ARMv8.4 adds LRCPC2 instructions to enable RCpc memory model.

This patch adds support for them.

### DIFF
--- a/src/mem/request.hh
+++ b/src/mem/request.hh
@@ -103,6 +103,7 @@ class Request : public Extensible<Request>
 
     enum : FlagsType
     {
+        // clang-format off
         /**
          * Architecture specific flags.
          *
@@ -168,6 +169,8 @@ class Request : public Extensible<Request>
         EVICT_NEXT                  = 0x04000000,
         /** The request should be marked with ACQUIRE. */
         ACQUIRE                     = 0x00020000,
+        /** The request should be marked with ACQUIRE_PC. */
+        ACQUIRE_PC                  = 0x00002000,
         /** The request should be marked with RELEASE. */
         RELEASE                     = 0x00040000,
 
@@ -259,6 +262,7 @@ class Request : public Extensible<Request>
         /** TLBI_EXT_SYNC_COMP seems to be the largest value
             of FlagsType, so HAS_NO_ADDR's value is that << 1 */
         HAS_NO_ADDR                = 0x0001000000000000,
+        // clang-format on
     };
     static const FlagsType STORE_NO_DATA = CACHE_BLOCK_ZERO |
         CLEAN | INVALIDATE;
@@ -1086,7 +1090,11 @@ class Request : public Extensible<Request>
     Flags getDest() const { return _flags & DST_BITS; }
 
     bool isAcquire() const { return _cacheCoherenceFlags.isSet(ACQUIRE); }
-
+    bool
+    isAcquirePC() const
+    {
+        return _cacheCoherenceFlags.isSet(ACQUIRE_PC);
+    }
 
     /**
      * Accessor functions for the cache bypass flags. The cache bypass
--- a/src/arch/arm/ArmISA.py
+++ b/src/arch/arm/ArmISA.py
@@ -72,9 +72,11 @@ class ArmDefaultSERelease(ArmRelease):
         "FEAT_FCMA",
         "FEAT_JSCVT",
         "FEAT_PAuth",
+        "FEAT_LRCPC",
         # Armv8.4
         "FEAT_FLAGM",
         "FEAT_FRINTTS",
+        "FEAT_LRCPC2",
         # Armv8.5
         "FEAT_FLAGM2",
         # Armv9.2
@@ -155,9 +157,9 @@ class ArmISA(BaseISA):
     )
 
     # !I8MM | !BF16 | SPECRES = 0 | !SB |
-    # GPI = 0x0 | GPA = 0x1 | API=0x0 | FCMA | JSCVT | APA=0x1
+    # GPI = 0x0 | GPA = 0x1 | LRCPC=0x2 | FCMA | JSCVT | APA=0x1
     id_aa64isar1_el1 = Param.UInt64(
-        0x0000000001011010, "AArch64 Instruction Set Attribute Register 1"
+        0x0000000001211010, "AArch64 Instruction Set Attribute Register 1"
     )
 
     # 4K | 64K | !16K | !BigEndEL0 | !SNSMem | !BigEnd | 8b ASID | 40b PA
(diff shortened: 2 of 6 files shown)


pranith:
@giactra Can you please review this patch? Thanks!

giactra:
Hi @pranith, thanks for your contribution.

It looks like you implemented the extension on the architectural side, but there is no logic in neither the CPU nor the MEM code to handle the different consistency model of the acquire-release pair. Unless I am missing something I presume the new LD Acquire RCPc will still follow the same SC ordering of the generic LDAR...

pranith:
@giactra Yes, for now they are treated with the stronger RCsc semantics, which is technically right, although not performant. This is more to handle binaries that have these RCpc instructions.

src/mem/request.hh
```diff
@@ -110,19 +110,19 @@ class Request : public Extensible<Request>
          * architecture-specific code. For example, SPARC uses them to
          * represent ASIs.
          */
-        ARCH_BITS                   = 0x000000FF,
+        ARCH_BITS = 0x000000FF,
         /** The request was an instruction fetch. */
-        INST_FETCH                  = 0x00000100,
+        INST_FETCH = 0x00000100,
```
giactra: Could you please revert this change and add a clang-format off comment here? I think there is a value in having them aligned (plus it makes reviewing easier)

giactra:
> @giactra Yes, for now they are treated with the stronger RCsc semantics, which is technically right, although not performant. This is more to handle binaries that have these RCpc instructions.

Yes I understand (though I hate binaries not checking ID registers /aux vectors before using a specific instruction).

I think adding what is supposed to be a performance optimization without really providing such perf gain is something very misleading for a user especially as it is not documented anyware that this is a functional implementation only.
Someone might play with the release object and come to the conclusion there is no perf gain between  RCpc and RCsc.

As I believe this is still a valuable contribution, I would ask you to do the following things:

1) Document in the PR/commit this is a functional only implementation and the consistency model is actually unaltered
2) Don't advertise the feature to guest software: this means the ID register should not be affected/changed by the extension being present in the release. Instead if the extension is present we should issue a warning. In this way code which is not looking into an ID register will still make use of the feature and not trigger and undefined instruction

Let me know what you think about this solution

pranith:
I will document in the PR that these instructions are functionally equivalent to the RCsc variants in our current implementation.

We must be cautious with any implied performance guarantees. We should not advertise this feature with the expectation of a significant performance uplift over the RCsc instructions. Given the limitations of our current microarchitectural models(esp., O3), we cannot guarantee RCpc will be more performant than RCsc. The initial benefit is functionality and instruction set completeness, not performance.

We can always improve our implementation to improve the performance. But that shouldn't be a prerequisite to using these instructions. I think we should advertise the instructions as supported by updating the ID register.

giactra:
> I will document in the PR that these instructions are functionally equivalent to the RCsc variants in our current implementation.

Thanks @pranith!

> 
> We must be cautious with any implied performance guarantees. We should not advertise this feature with the expectation of a significant performance uplift over the RCsc instructions. Given the limitations of our current microarchitectural models(esp., O3), we cannot guarantee RCpc will be more performant than RCsc. The initial benefit is functionality and instruction set completeness, not performance.
> 
> We can always improve our implementation to improve the performance. But that shouldn't be a prerequisite to using these instructions. I think we should advertise the instructions as supported by updating the ID register.

Let me push back on this. Regardless of the perf-gain (which is dictated by multiple factors), the idea behind these new instructions is that they allow a more aggressive reordering with the release instruction. Reordering which we don't exploit with the current PR. Their only _raison d-etre_ is exactly to allow for this more relaxed consistency model. We could stick otherwise with traditional load-acquire and be happy about it. Now I understand of course there is the compatibility problem and these instructions are now mandatory. For this, I am extremely grateful for your PR!

Ultimately, I think I am ok with advertising this in the ID registers, with the caveat that we should warn about current implementation if the extension is part of the release.

As a side note: we should not use the state of the O3 model as an excuse for extending the gap with a real modern core even further; I agree we shouldn't make performance gain statements, but it's not unreasonable to assume that with a balanced core configuration, there should be a bit of an uplift from increased MLP, unless it's empirically proven that across a series of reference benchmarks there is no net gain.

If we cannot even make such simple assumptions ( > MLP implies > perf) we might as well throw gem5 in the bin

powerjg:
> Ultimately, I think I am ok with advertising this in the ID registers, with the caveat that we should warn about current implementation if the extension is part of the release.

> As a side note: we should not use the state of the O3 model as an excuse for extending the gap with a real modern core even further; I agree we shouldn't make performance gain statements, but it's not unreasonable to assume that with a balanced core configuration, there should be a bit of an uplift from increased MLP, unless it's empirically proven that across a series of reference benchmarks there is no net gain.

> If we cannot even make such simple assumptions ( > MLP implies > perf) we might as well throw gem5 in the bin

+1

pranith:
I updated the commit message. @giactra Can you please review? Thanks!

giactra (commented):
Thanks.

I can see you updated the commit only. Would you mind reflecting the change in the PR message as well. Whenever we squash merge, the PR message will effectively become your only commit so the commit message will be lost.

Also

```

The instructions are treated as RCsc instructions which is functionally
correct but not better performant
```

I actually agree with your initial statement. Rather than talking about performance, we should actually phrase the existing limitation with something like:

```
The instructions are treated as RCsc instructions which is functionally
correct but does not exploit the more relaxed memory model of RCpc
```

Also in addition to the commit/PR message, it would be cool if we would warn in gem5 whenever the extension gets used. You could for example warn once in the LoadAcquirePC

`warn_once("LoadAcquirePC implemented with SC\n")`

As a last note, I don't think it should be hard to implement PC for the new load acquire instructions in the CPU model... But we can try that in the future in a separate PR

src/mem/request.hh
pranith: Done.

pranith:
@giactra I updated the commit as requested. Please review. Thanks!

src/mem/request.hh
```diff
@@ -167,7 +168,9 @@ class Request : public Extensible<Request>
         /** The request should be marked as LRU. */
         EVICT_NEXT                  = 0x04000000,
         /** The request should be marked with ACQUIRE. */
-        ACQUIRE                     = 0x00020000,
+        ACQUIRE                     = 0x000A0000,
```
giactra: Here you are displacing the original ACQUIRE code to assign it to the ACQUIRE_PC. I would do the opposite for consistency. Also the new code:

0x000A0000

conflicts with the existing NO_ACCESS code:

0x00080000.

In other words with this change any Load Acquire is treated as a ACQUIRE_PC + NO_ACCESS.

0x00080000 +
0x00020000 =
0x000A0000

No access basically prevents a memory access from happening therefore this breaks load acquires in gem5.
You should try to find a new free encoding for ACQUIRE_PC (BTW this flag system does not scale and we should move to something different/ easier to extend in the long run)

pranith: @giactra I updated the patch so that ACQUIRE_PC is using the lowest available encoding. Please review! Thanks.

giactra (approved):

PR #2603, opened 2025-09-19
configs,gpu-compute,dev: MI355X and ROCm 7 support

abmerop:
This adds the CDNA4 ISA to the GfxVersion enum and handles conditionals in configs, gpu-compute, and dev-amdgpu which enable certain features for specific GFX versions. This is mostly the same as gfx942.

A new disk image with ROCm 7.0+ installed is required to use the MI355X model. A preview disk image can be found at:

https://github.com/abmerop/gem5-resources/tree/rocm70

To use the model you can start with
configs/example/gem5_library/x86-mi300x-gpu.py and replace MI300X with MI355X.

Note that most new instructions/features are available with the exception of the new MFMA instructions.

PR #2633, opened 2025-10-04
dev-amdgpu,stdlib: Allow for multiple GPUs in stdlib

abmerop:
This is based on #2603 and will have to be rebased when that is merged.

- Allows for user specified GPU memory size. Previously this was hard coded to 16GiB and the user specified size in stdlib was ignored. This is more necessary for multiple GPU simulation to reduce memory required for a single simulation.
- Add support for additional GPU's ROM region. These read from the PCI ROM address instead of legacy x86 VGA ROM region. It works by copying the VBIOS from VGA region. Implicitly this means all GPUs must be the same model (i.e., all MI300X or all MI355X).
- Support for miscellaneous MMIOs only used when multiple GPUs are detected.

Example top level config snippet:

```
...
# HBM2Stack best available
gpu0 = MI355X(gpu_memory=HBM2Stack(size="4GiB"))
gpu1 = MI355X(gpu_memory=HBM2Stack(size="4GiB"))
gpu2 = MI355X(gpu_memory=HBM2Stack(size="4GiB"))
gpu3 = MI355X(gpu_memory=HBM2Stack(size="4GiB"))

board = ViperBoard(
    ...,
    gpus=[gpu0, gpu1, gpu2, gpu3],
)
...
```

Limitation: The GPUs cannot yet communicate device-to-device. XGMI support is needed for this.

### DIFF
--- a/src/dev/amdgpu/amdgpu_device.cc
+++ b/src/dev/amdgpu/amdgpu_device.cc
@@ -56,7 +56,8 @@ AMDGPUDevice::AMDGPUDevice(const AMDGPUDeviceParams &p)
     : PciEndpoint(p), gpuMemMgr(p.memory_manager), deviceIH(p.device_ih),
       cp(p.cp), checkpoint_before_mmios(p.checkpoint_before_mmios),
       init_interrupt_count(0), _lastVMID(0),
-      deviceMem(name() + ".deviceMem", p.memories, false, "", false)
+      deviceMem(name() + ".deviceMem", p.memories, false, "", false),
+      system(p.system)
 {
     uint64_t vram_size = 0;
 
@@ -271,13 +272,23 @@ void
 AMDGPUDevice::readROM(PacketPtr pkt)
 {
     Addr rom_offset = pkt->getAddr() & (ROM_SIZE - 1);
-    uint64_t rom_data = 0;
 
-    memcpy(&rom_data, rom.data() + rom_offset, pkt->getSize());
-    pkt->setUintX(rom_data, ByteOrder::little);
+    // Read directly from the VGA ROM region. For multiple GPUs, this means
+    // every GPU must be the same type. However, this allows for one less
+    // input file as the GPU VBIOS is already part of the gem5 resources disk
+    // image and loaded at the VGA_ROM_DEFAULT address as part of readfile.
+    RequestPtr request = std::make_shared<Request>(
+        VGA_ROM_DEFAULT + rom_offset, pkt->getSize(), 0, vramRequestorId());
 
-    DPRINTF(AMDGPUDevice, "Read from addr %#x on ROM offset %#x data: %#x\n",
-            pkt->getAddr(), rom_offset, rom_data);
+    auto readPkt = new Packet(request, MemCmd::ReadReq);
+    readPkt->allocate();
+
+    system->getPhysMem().access(readPkt);
+
+    DPRINTF(AMDGPUDevice, "Read from VGA ROM offset %#x returned %#x\n",
+            rom_offset, readPkt->getUintX(ByteOrder::little));
+
+    pkt->setUintX(readPkt->getUintX(ByteOrder::little), ByteOrder::little);
 }
 
 void
@@ -285,13 +296,21 @@ AMDGPUDevice::writeROM(PacketPtr pkt)
 {
     assert(isROM(pkt->getAddr()));
 
+    // Read directly from the VGA ROM region at VGA_ROM_DEFAULT address.
     Addr rom_offset = pkt->getAddr() - romRange.start();
     uint64_t rom_data = pkt->getUintX(ByteOrder::little);
 
-    memcpy(rom.data() + rom_offset, &rom_data, pkt->getSize());
+    RequestPtr request = std::make_shared<Request>(
+        VGA_ROM_DEFAULT + rom_offset, pkt->getSize(), 0, vramRequestorId());
 
-    DPRINTF(AMDGPUDevice, "Write to addr %#x on ROM offset %#x data: %#x\n",
-            pkt->getAddr(), rom_offset, rom_data);
+    auto writePkt = new Packet(request, MemCmd::WriteReq);
+    writePkt->allocate();
+    writePkt->setUintX(rom_data, ByteOrder::little);
+
+    system->getPhysMem().access(writePkt);
+
+    DPRINTF(AMDGPUDevice, "Wrote to VGA ROM offset %#x value %#x\n",
+            rom_offset, writePkt->getUintX(ByteOrder::little));
 }
 
 AddrRangeList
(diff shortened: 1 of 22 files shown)


mattsinc:
@v-ramadas , @ronpower7: FYI

src/dev/amdgpu/amdgpu_device.cc
```diff
@@ -66,8 +69,12 @@ AMDGPUDevice::AMDGPUDevice(const AMDGPUDeviceParams &p)
 
         // Add to system's device memory map.
         p.system->addDeviceMemory(gpuMemMgr->getRequestorID(), m);
+
+        vram_size += m->getAddrRange().size();
     }
 
+    vramSize = vram_size;
+
```
mattsinc, line 82: not sure the line of space is needed here?  But this is minor.

src/dev/amdgpu/amdgpu_device.cc
```diff
@@ -148,12 +157,17 @@ AMDGPUDevice::AMDGPUDevice(const AMDGPUDeviceParams &p)
...
     // Address aperture for device memory. We tell this to the driver and
     // could possibly be anything, but these are the values used by hardware.
     uint64_t mmhubBase = 0x8000ULL << 24;
     uint64_t mmhubTop = 0x83ffULL << 24;
-    uint64_t mem_size = 0x3ff0; // 16 GB of memory
+    uint64_t mmio_mem_size = vram_size / 0x100000;
+
+    // For the max memory size, this would overflow. Subtracting one prevents
+    // this and is still "close enough" to the BAR size to enable "large BAR."
+    mmio_mem_size -= 0x1;
```
mattsinc, line 175: I'm not sure I follow the logic about why this would overflow, as line 166 would have already overflowed with the larger number by the time you get to this line, no?

src/dev/amdgpu/amdgpu_device.cc
```diff
@@ -169,11 +183,13 @@ AMDGPUDevice::AMDGPUDevice(const AMDGPUDeviceParams &p)
...
     gpuvm.setMMIOAperture(GRBM_MMIO_RANGE, AddrRange(0x8000, 0xC000));
     gpuvm.setMMIOAperture(GFX_MMIO_RANGE,  AddrRange(0x28000, 0x3F000));
-    if (getGfxVersion() == GfxVersion::gfx942) {
+    if (getGfxVersion() == GfxVersion::gfx942 ||
+        getGfxVersion() == GfxVersion::gfx950) {
         gpuvm.setMMIOAperture(MMHUB_MMIO_RANGE,  AddrRange(0x60D00, 0x62E20));
     } else {
         gpuvm.setMMIOAperture(MMHUB_MMIO_RANGE,  AddrRange(0x68000, 0x6A120));
     }
+    gpuvm.setMMIOAperture(SMU_MMIO_RANGE,  AddrRange(0x5A000, 0x5ACE4));
```
mattsinc: is it possible to add a comment explaining what this address range represents?

src/dev/amdgpu/amdgpu_device.cc
```diff
@@ -186,27 +202,49 @@ AMDGPUDevice::AMDGPUDevice(const AMDGPUDeviceParams &p)
     } else if (p.device_name == "MI100") {
         setRegVal(MI100_FB_LOCATION_BASE, mmhubBase >> 24);
         setRegVal(MI100_FB_LOCATION_TOP, mmhubTop >> 24);
-        setRegVal(MI100_MEM_SIZE_REG, mem_size);
+        setRegVal(MI100_MEM_SIZE_REG, mmio_mem_size);
```
mattsinc, line 215: this fix (and line 211 below) seem unrelated to the main support you are adding in this patch, but I think it's ok to leave.

src/dev/amdgpu/amdgpu_device.cc
```diff
@@ -186,27 +202,49 @@ AMDGPUDevice::AMDGPUDevice(const AMDGPUDeviceParams &p)
...
         setRegVal(MI200_FB_LOCATION_TOP, mmhubTop >> 24);
-        setRegVal(MI200_MEM_SIZE_REG, mem_size);
+        setRegVal(MI200_MEM_SIZE_REG, mmio_mem_size);
     } else if (p.device_name == "MI300X") {
         // VRAM size in MB (shifted right by 20 bits)
         setRegVal(MI300X_FB_LOCATION_BASE, mmhubBase >> 24);
         setRegVal(MI300X_FB_LOCATION_TOP, mmhubTop >> 24);
-        setRegVal(MI300X_MEM_SIZE_REG, mem_size);
+        setRegVal(MI300X_MEM_SIZE_REG, mmio_mem_size);
+    } else if (p.device_name == "MI355X") {
```
mattsinc: if everything is the same with MI300X, do we want to make the else if on line 212 just check for both?  I guess this way enables some additional future flexibility?

src/dev/amdgpu/amdgpu_smu.hh
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright (c) 2024 Advanced Micro Devices, Inc.
```
mattsinc: should be 2025?

src/dev/amdgpu/amdgpu_smu.cc
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright (c) 2024 Advanced Micro Devices, Inc.
```
mattsinc: 2025?

src/python/gem5/components/devices/gpus/amdgpu.py
```diff
@@ -295,6 +303,60 @@ def get_driver_command(self, debug: bool = False):
             "export HCC_AMDGPU_TARGET=gfx942\n"
```
mattsinc, line 303: I guess this means the default is MI300X?

src/python/gem5/components/devices/gpus/viper_shader.py
```diff
@@ -213,7 +215,7 @@ def __init__(
         self._create_tlbs(device)
 
         # This arbitrary address is something in the X86 I/O hole
-        hsapp_gpu_map_paddr = 0xE00000000
+        hsapp_gpu_map_paddr = 0xE00000000 + 0x1000 * shader_id
```
mattsinc, line 218: This is the same thing my group had done with our "old" chiplet support, and it ended up having strange problems with addresses slightly off the end ... but if it's working for you and in newer ROCm versions, then this is fine with me.

mattsinc (changes requested):
few minor things, but largely looks good!

src/dev/amdgpu/amdgpu_device.cc
abmerop: The C++ variable will not overflow, but the the MMIO register is only a few bits wide.  Take for example 1 GiB = 0x40000000. This requires 31 bits. But if we write it as 0x3FFFFFFF and assume an implicit +1 we only need 30 bits. IIRC this MMIO is 16 bits so if we don't subtract 1 the max memory size would be read as 0 bytes in size since the driver will only read 16 bits

Updated comment to explain this better.

src/dev/amdgpu/amdgpu_device.cc
abmerop: I guess this one is since MI100 isn't actually supported... But we do support MI200 and being able to change memory size is still useful there.

src/python/gem5/components/devices/gpus/amdgpu.py
abmerop: No, this is in the MI300X class.  These exports are used in case the user wants to build code in the gem5 simulation but probably are not necessary anymore

src/python/gem5/components/devices/gpus/viper_shader.py
abmerop: It is not clear to me how this is used in FS mode, but they at least need to be different to avoid AddrRange error of two ports mapping the same address range

src/dev/amdgpu/amdgpu_device.cc
abmerop: Added to the comment block above all the MMIO ranges

src/dev/amdgpu/amdgpu_device.cc
abmerop: That's fair, done

src/dev/amdgpu/amdgpu_smu.cc
abmerop: Updated. Reveals how old this code is and how long it takes to pipeline out :)

src/dev/amdgpu/amdgpu_device.cc
mattsinc: I think it's fine

src/python/gem5/components/devices/gpus/amdgpu.py
mattsinc: but even so wouldn't it mean they are building for MI300X?

mattsinc (dismissed):

src/python/gem5/components/devices/gpus/amdgpu.py
abmerop: Yes, it will build for MI300X within gem5 when using the MI300X model.  I am not sure how this interacts with --offload-arch in LLVM.  The GPU device is properly detected though (for all gem5 supported models) if you are suggesting we remove these in the future

powerjg (commented):
The stdlib changes lgtm. This is *super cool*, Matt! Thanks!

src/python/gem5/components/devices/gpus/amdgpu.py
mattsinc: I think this is fine as is, I just wanted to understand it

mattsinc (approved):

PR #2634, opened 2025-10-06
mem: Remove deprecated port names in mem

giactra:
We tagged them as deprecated quite some time ago.
Pretty confident we can get rid of them

Change-Id: I6f1f6f760fd6719d09ec01d83dd449076d1af78a


powerjg (approved):
🥳

PR #2640, opened 2025-10-06
Small cleanup in arch-arm and dev-arm

powerjg:
Add missing fallthrough, python imports, and virtual destructors

### DIFF
--- a/src/arch/arm/ArmMMU.py
+++ b/src/arch/arm/ArmMMU.py
@@ -42,6 +42,7 @@ from m5.objects.ArmTLB import (
 )
 from m5.objects.BaseMMU import BaseMMU
 from m5.objects.ClockedObject import ClockedObject
+from m5.objects.System import System
 from m5.params import *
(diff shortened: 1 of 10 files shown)


powerjg:
@giactra can you take careful look at the missing fallthough in `isa.cc`. That *might* be a bug and it could need to be a `break`. https://github.com/gem5/gem5/pull/2640/commits/733fd05d6668a70b8e3c81661f02efd112216ed6

giactra:
> @giactra can you take careful look at the missing fallthough in `isa.cc`. That _might_ be a bug and it could need to be a `break`. [733fd05](https://github.com/gem5/gem5/commit/733fd05d6668a70b8e3c81661f02efd112216ed6)

Thanks jason, it's not a bug. The break should only happen to go at the end of the switch and perform the setMiscRegNoEffect. Here we are already doing the SCTLR write within the switch case

giactra (approved):

PR #2628, opened 2025-10-02
Fix compiler warnings arm isa

powerjg:
This PR fixes some self-assignment warnings and one missing break statement. The self-assignments are replaced with `[[maybe_unused]]` when needed.

PR #2645, opened 2025-10-07
arch-arm: Add missing break in decoder

powerjg:
My compiler complained about this missing breakpoint in the decoder
switch/case.


powerjg (commented):
From #2628

giactra (approved):

PR #2646, opened 2025-10-07
tests: fix failing FIFO RP daily tests

erin-le:
The unittests-fast-debug (fast) tests in the daily tests are failing due to the FIFO replacement policy TouchNull death test. The assert that is supposed to be thrown in the test doesn't occur, as the gem5.fast build optimizes out assertions. This commit makes it so that the test is skipped on fast builds.

### DIFF
--- a/src/mem/cache/replacement_policies/fifo_rp.test.cc
+++ b/src/mem/cache/replacement_policies/fifo_rp.test.cc
@@ -218,6 +218,10 @@ TEST_F(FIFORPFDeathTest, ResetNull)
 
 TEST_F(FIFORPFDeathTest, TouchNull)
 {
+#ifdef NDEBUG
+    GTEST_SKIP() << "Skipping as assertions are "
+                    "stripped out of fast builds";
+#endif
     ASSERT_DEATH(rp->touch(nullptr), "");
 }
 


powerjg:
@odanrc I think this was your change... does this fix for the daily tests look good to you? I believe the other DEATH tests actually work because they end up with segfaults.

odanrc:
> @odanrc I think this was your change... does this fix for the daily tests look good to you? I believe the other DEATH tests actually work because they end up with segfaults.

Yes, this is the usual fix, but a segfault is not good - wouldn't a better fix be replacing assert by fatal_if, gem5_assert, or exceptions?

odanrc (approved):

powerjg:
> wouldn't a better fix be replacing assert by fatal_if, gem5_assert, or exceptions?

Maybe so, but we don't have much consistency throughout the codebase. And, let's not use exceptions 😉

powerjg (approved):

BobbyRBruce (approved):
Looks good to me. @erin-le : We were talking of making these unit tests part of the CI test suite. Are you going to do that in this PR or a separate PR?

If that latter I will merge.

erin-le:
@BobbyRBruce: I'll do it in a separate PR.

PR #2647, opened 2025-10-07
stdlib: Define a new get_mem_ranges method

giactra:


src/python/gem5/components/boards/abstract_board.py
```diff
@@ -141,6 +153,15 @@ def get_memory(self) -> "AbstractMemory":
         """
         return self.memory
 
+    def get_mem_ranges(self) -> Sequence[AddrRange]:
+        """Get all the mem ranges in the board, This
+        tries to account for boards shipping memories other
```
powerjg:
```suggestion
        tries to account for boards skipping memories other
```
I think this is what you meant ;)

src/python/gem5/components/boards/abstract_board.py
```diff
@@ -141,6 +153,15 @@ def get_memory(self) -> "AbstractMemory":
         """
         return self.memory
 
+    def get_mem_ranges(self) -> Sequence[AddrRange]:
```
powerjg, line 156: Could you instead implement this like:

```python
return [rng for rng, _ in get_mem_ports()]
```

I am curious as to what this function is needed for that hasn't been used already. That said, I can definitely believe that we missed some important APIs.

powerjg (commented):
Where/how will `get_mem_ranges` be used?

src/python/gem5/components/boards/abstract_board.py
giactra: Nono I meant shipping, but it's probably bad translation from italian. Will reword it

src/python/gem5/components/boards/abstract_board.py
giactra: Hey Jason, you will see with the following PR, but basically it's linked with the capability of generating multiple directories which you already reviewed. As you need to pass a list of address ranges to the classmethod, we need to find a way of passing all the memory ranges (dram + bootmem + any sram).

The problem with this:

`return [rng for rng, _ in get_mem_ports()]`

I initially tried to do it this way. The problem is that if you have a dual channel DDR in your platform you actually create two memory controllers with two interleaved address ranges and two ports.

So the get_mem_ports returns two ranges which are effectively equal and the directory controller is confused by the fact that same range is registered twice and panics.

I could hack it by filtering equal ranges, but it would be a hack. Also if you think about it, the upstream directory should not depend upon the range generation of a downstream component like the memory controller, whose addressing scheme should be completely invisible. So in my view there should be available an abstract idea of memory ranges which is decoupled from the memory implementation

giactra:
> Where/how will `get_mem_ranges` be used?

Good question, I think I addressed it here: https://github.com/gem5/gem5/pull/2647#discussion_r2415969129

I thought about posting a single PR (with its usage) but I thought it would make reviewing time a bit harder

Next PR is adding custom mesh capability to stdlib CHI and I need to find a way to instantiate multiple directories in the mesh

powerjg (approved):

PR #2648, opened 2025-10-08
stdlib: Support generation of multiple directories in CHI

giactra:
Up until this point stdlib was only supporting a single directory system (with no cache), therefore it didn't require to be passed a list of address ranges, relying on the default Param value (set to the AllMemory alias :from 0 to MaxAddr) [1]

If we want to support multiple directories hierarchies, we need to provide address ranges so that a request gets dispatched to the proper directory

We add a classmethod, taken from the ruby CHI_config.py example [2] and we adapt it to stdlib

    src/mem/ruby/slicc_interface/Controller.py#L51
[2]: https://github.com/gem5/gem5/blob/stable/\
    configs/ruby/CHI_config.py#L636

Change-Id: I24341926da3a950554c2251bc74fff001d69fd7b

### DIFF
--- a/src/python/gem5/components/cachehierarchies/chi/nodes/directory.py
+++ b/src/python/gem5/components/cachehierarchies/chi/nodes/directory.py
@@ -72,7 +72,7 @@ class BaseDirectory(AbstractNode):
         dir_idx: int,
         mem_ranges: List[AddrRange],
         cache_line_size,
-    ):
+    ) -> List[AddrRange]:
         block_size_bits = int(math.log(cache_line_size, 2))
         llc_bits = int(math.log(num_directories, 2))
         numa_bit = block_size_bits + llc_bits - 1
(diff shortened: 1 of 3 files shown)


powerjg (dismissed):
This makes sense to me. 

Here's an example of what we did to "solve" this problem: https://github.com/darchr/novoverse/blob/main/components/saga/cache_hierarchy/__init__.py#L166

In that code, we created the directory, then set the address range. Making it more explicit, like this PR, is a better approach. Thanks!

src/python/gem5/components/cachehierarchies/chi/nodes/directory.py
```diff
@@ -24,17 +36,61 @@
...
+        super().__init__(network, cache_line_size)
+
+    @classmethod
+    def create_addr_ranges(
+        cls,
+        num_directories: int,
+        dir_idx: int,
+        mem_ranges: List[AddrRange],
+        cache_line_size,
+    ):
```
powerjg: Minor nit: 
```suggestion
    ) -> List[AddrRange]:
```

BobbyRBruce (dismissed):
Outside of @powerjg 's nitpick change I'm ok with this. Consider it approved.

BobbyRBruce (approved):

PR #2651, opened 2025-10-09
cpu-o3: Make space for a pipeline tracer

giactra:
At the moment the internal O3 pipeline tracer is bolted inside the O3 CPU model and the dynamic instruction
class. Timings are annotated on every O3 instruction whenever the O3PipeViewer option is selected and all
the annotations are dumped when the instruction gets destructed.

With this PR we aim at making space for other pipeline viewers, to be implemented as CPU tracers:

1) We decouple timing annotations from the O3PipeViewer option
2) We extend the InstTracer APIs to accept a DynInst / ExecContext in general, which contains dynamic information
about the execution context.

**Caveats and limitations:**

We don't move the O3PipeViewer to use the new logic for the following reasons:

1) We will lose the capability of generating an O3PipeView while using another tracer (gem5 only supports one tracer at a time) -> Note: a multi-tracer implementation is actually possible, it was my first CR for gem5 (got rejected but I was too junior and didn't push back)

2) The O3PipeViewer dumps every instruction (even instructions squashed due to misprediction), but the tracer only dumps committed instructions. So the pipe viewer won't have an easy way of dumping speculative instructions. This solution might require an extension to the tracer APIs (which are easy but pretty restrictive)

### DIFF
--- a/src/cpu/o3/fetch.cc
+++ b/src/cpu/o3/fetch.cc
@@ -60,7 +60,6 @@
 #include "debug/Drain.hh"
 #include "debug/Fetch.hh"
 #include "debug/O3CPU.hh"
-#include "debug/O3PipeView.hh"
 #include "mem/packet.hh"
 #include "params/BaseO3CPU.hh"
 #include "sim/byteswap.hh"
@@ -1283,11 +1282,7 @@ Fetch::fetch(bool &status_change)
             ppFetch->notify(instruction);
             numInst++;
 
-#if TRACING_ON
-            if (debug::O3PipeView) {
-                instruction->fetchTick = curTick();
-            }
-#endif
+            instruction->fetchTick = curTick();
 
             set(next_pc, this_pc);
 
--- a/src/cpu/o3/fetch.cc
+++ b/src/cpu/o3/fetch.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010-2014 ARM Limited
+ * Copyright (c) 2010-2014, 2025 Arm Limited
  * Copyright (c) 2012-2013 AMD
  * Copyright (c) 2022-2023 The University of Edinburgh
  * All rights reserved.
@@ -1032,9 +1032,9 @@ Fetch::buildInst(ThreadID tid, StaticInstPtr staticInst,
 
 #if TRACING_ON
     if (trace) {
-        instruction->traceData =
-            cpu->getTracer()->getInstRecord(curTick(), cpu->tcBase(tid),
-                    instruction->staticInst, this_pc, curMacroop);
+        instruction->traceData = cpu->getTracer()->getInstRecord(
+            curTick(), cpu->tcBase(tid), instruction.get(),
+            instruction->staticInst, this_pc, curMacroop);
     }
 #else
     instruction->traceData = NULL;
--- a/src/cpu/o3/rename.cc
+++ b/src/cpu/o3/rename.cc
@@ -48,7 +48,6 @@
 #include "cpu/o3/limits.hh"
 #include "cpu/reg_class.hh"
 #include "debug/Activity.hh"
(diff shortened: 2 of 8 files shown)


src/cpu/o3/fetch.cc
```diff
@@ -1033,9 +1032,9 @@ Fetch::buildInst(ThreadID tid, StaticInstPtr staticInst,
 
 #if TRACING_ON
     if (trace) {
-        instruction->traceData =
-            cpu->getTracer()->getInstRecord(curTick(), cpu->tcBase(tid),
-                    instruction->staticInst, this_pc, curMacroop);
+        instruction->traceData = cpu->getTracer()->getInstRecord(
+            curTick(), cpu->tcBase(tid), instruction.get(),
```
dhschall, line 1036: Shouldn't that be the `staticInst`? The function `getInstRecord` takes a `StaticInstPtr` but you pass the in `DynInstPtr`.

Edit: Oh I see its moved to the next line. You pass a new argument: `ExecContext`. But then the question is the same will: wont the `get()` function pass a raw pointer to the dynamic instruction instead of the ExecContext?

src/cpu/o3/rename.cc
```diff
@@ -751,6 +750,8 @@ Rename::renameInsts(ThreadID tid)
         // this instruction have been renamed.
         ppRename->notify(inst);
 
+        inst->renameEndTick = curTick() - inst->fetchTick;
```
dhschall: Do you know if that will cause any backward compatibility issues? I.e. will the Pipeline viewer and Konata still work?
I assume that the viewers simply ignore the new tick.

dhschall (commented):
Thanks for this. It makes sense and is much cleaner. One question which looks like a bug but I'm not sure.

You may wanna add the new tick to the trace dumping.
https://github.com/giactra/gem5/blob/a3e21c2b6e2a98ff20f3f019cd61f4857302b7ed/src/cpu/o3/dyn_inst.cc#L242
However I don't know if that will cause any backward compatibility issues? I.e. will the Pipeline viewer and Konata still work if you dump it? I assume that the viewers simply ignore the new tick.

src/cpu/o3/fetch.cc
giactra: > Edit: Oh I see its moved to the next line. You pass a new argument: ExecContext. But then the question is the same will: wont the get() function pass a raw pointer to the dynamic instruction instead of the ExecContext?

Yes it will. The DynInst is actually an ExecContext: https://github.com/gem5/gem5/blob/stable/src/cpu/o3/dyn_inst.hh#L75

Passing down a raw pointer is less preferrable than passing a smart one, however I wanted this to be generic across CPU models and other CPUs don't wrap their ExecContext around smart pointers

giactra:
> Thanks for this. It makes sense and is much cleaner. One question which looks like a bug but I'm not sure.
> 

Thanks @dhschall 

> You may wanna add the new tick to the trace dumping. https://github.com/giactra/gem5/blob/a3e21c2b6e2a98ff20f3f019cd61f4857302b7ed/src/cpu/o3/dyn_inst.cc#L242 However I don't know if that will cause any backward compatibility issues? I.e. will the Pipeline viewer and Konata still work if you dump it? I assume that the viewers simply ignore the new tick.

I could add it! However the viewers would definitely disregard it as of now (in the good case) or crash the generation (in the bad case) so I don't see a motivation to add the data to the dump for now!

src/cpu/o3/fetch.cc
dhschall: Great! Thanks for the clarification.

dhschall (approved):
Thanks @giactra for clarifying. Yes, I agree that there is no need to risk a bug. If people want to have the new data in the viewer, they can add it themself.
I'm good with it.

PR #2652, opened 2025-10-09
configs, cpu-o3: Implement a distributed InstructionQueue

giactra:
This commit is implementing a distributed InstructionQueue.

Prior to this patch it was only possible to have a unified instruction queue for the O3 CPU: every renamed/dispatched instruction was allocated in a single queue.

With this patch we define a IQUnit SimObject (which is basically a single IQ) and we allow the O3CPU to have a vector of IQs to reflect the presence of multiple instruction queues.

A IQUnit is tighlty coupled with its own FUPool: it can only store instructions executable by one of the FU in the pool. This also means we don't have a single fuPool anymore, but a collection of pools (fuPool is a IQUnit Param)

By default, the unified design is in place. To adopt the multi-IQs design, one simply has to provide a different vector to the O3CPU

Change-Id: Ic3923ab779bdb2e957fc5e7a4c5881f9e685aa81

### DIFF
--- a/src/cpu/o3/IQUnit.py
+++ b/src/cpu/o3/IQUnit.py
@@ -1,5 +1,3 @@
-# -*- mode:python -*-
-
 # Copyright (c) 2025 Arm Limited
 # All rights reserved.
 #
@@ -11,9 +9,6 @@
 # terms below provided that you ensure that this notice is replicated
 # unmodified and in its entirety in all distributions of the software,
 # modified or unmodified, in source code or in binary form.
-#
-# Copyright (c) 2006 The Regents of The University of Michigan
-# All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions are
@@ -38,64 +33,28 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-import sys
+from m5.objects.FUPool import (
+    DefaultFUPool,
+    FUPool,
+)
+from m5.objects.SMT import SMTQueuePolicy
+from m5.params import *
+from m5.proxy import Parent
+from m5.SimObject import SimObject
 
-Import('*')
 
-if env['CONF']['BUILD_ISA']:
-    SimObject('FUPool.py', sim_objects=['FUPool'])
-    SimObject('FuncUnitConfig.py', sim_objects=[])
-    SimObject('BaseO3CPU.py', sim_objects=['BaseO3CPU'])
-    SimObject('IQUnit.py', sim_objects=['IQUnit'])
-    SimObject('SMT.py',
-        enums=['SMTFetchPolicy', 'SMTQueuePolicy', 'CommitPolicy'])
+class IQUnit(SimObject):
+    type = "IQUnit"
+    cxx_class = "gem5::o3::IQUnit"
+    cxx_header = "cpu/o3/inst_queue.hh"
 
-    Source('bac.cc')
-    Source('commit.cc')
-    Source('cpu.cc')
-    Source('decode.cc')
-    Source('dyn_inst.cc')
-    Source('fetch.cc')
-    Source('free_list.cc')
-    Source('ftq.cc')
-    Source('fu_pool.cc')
-    Source('iew.cc')
-    Source('inst_queue.cc')
-    Source('lsq.cc')
-    Source('lsq_unit.cc')
-    Source('mem_dep_unit.cc')
-    Source('regfile.cc')
-    Source('rename.cc')
-    Source('rename_map.cc')
-    Source('rob.cc')
-    Source('scoreboard.cc')
-    Source('store_set.cc')
-    Source('thread_context.cc')
-    Source('thread_state.cc')
+    numEntries = Param.Unsigned(64, "Number of instruction queue entries")
 
-    DebugFlag('BAC')
-    DebugFlag('CommitRate')
-    DebugFlag('FTQ')
-    DebugFlag('IEW')
-    DebugFlag('IQ')
-    DebugFlag('LSQ')
-    DebugFlag('LSQUnit')
-    DebugFlag('MemDepUnit')
-    DebugFlag('O3CPU')
-    DebugFlag('ROB')
-    DebugFlag('Rename')
-    DebugFlag('Scoreboard')
-    DebugFlag('StoreSet')
-    DebugFlag('Writeback')
+    fuPool = Param.FUPool(DefaultFUPool(), "Functional Unit pool")
 
-    CompoundFlag('O3CPUAll', [ 'BAC', 'FTQ', 'Fetch', 'Decode', 'Rename',
-        'IEW', 'Commit',
-        'IQ', 'ROB', 'FreeList', 'LSQ', 'LSQUnit', 'StoreSet', 'MemDepUnit',
-        'DynInst', 'O3CPU', 'Activity', 'Scoreboard', 'Writeback' ])
+    numThreads = Param.Unsigned(
+        Parent.numThreads, "number of HW thread contexts"
+    )
 
-    SimObject('BaseO3Checker.py', sim_objects=['BaseO3Checker'])
-    Source('checker.cc')
-
-    # For backwards compatibility
-    SimObject('O3CPU.py', sim_objects=[], tags=['isa'])
-    SimObject('O3Checker.py', sim_objects=[], tags=['isa'])
+    smtIQPolicy = Param.SMTQueuePolicy("Partitioned", "SMT IQ Sharing Policy")
+    smtIQThreshold = Param.Int(100, "SMT IQ Threshold Sharing Parameter")
--- a/src/cpu/o3/SMT.py
+++ b/src/cpu/o3/SMT.py
@@ -1,6 +1,4 @@
-# -*- mode:python -*-
-
-# Copyright (c) 2025 Arm Limited
+# Copyright (c) 2016, 2019 ARM Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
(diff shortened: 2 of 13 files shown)


pranith:
@giactra Nice to see this support added! Can you please also add a config that uses multiple IQs?

src/cpu/o3/inst_queue.hh
```diff
@@ -78,6 +79,85 @@ class FUPool;
...
+    fuPool()
+    {
+        return _fuPool;
+    }
+
+  private:
+    /** IQ sharing policy for SMT. */
+    SMTQueuePolicy iqPolicy;
+
+    /** Number of Total Threads*/
```
dhschall: Nitpick
```suggestion
    /** Number of Total Threads */
```

src/cpu/o3/inst_queue.cc
```diff
@@ -63,17 +64,152 @@ namespace gem5
...
+
+    inst->setInIQ(this);
+
+    count[inst->threadNumber]++;
+}
+
+void
+IQUnit::remove(const DynInstPtr &inst)
+{
+    _freeEntries++;
```
dhschall, line 132: Why is here the `clearInIQ` not called? Is it intentially?

src/cpu/o3/inst_queue.cc
```diff
@@ -63,17 +64,152 @@ namespace gem5
...
+void
+IQUnit::setActiveThreads(list<ThreadID> *at_ptr)
+{
+    activeThreads = at_ptr;
+}
+
+void
+IQUnit::resetState()
+{
+    _freeEntries = _numEntries;
```
dhschall, line 147: Same here. Only the count is updated but the instructions iqPointer still points to this queue. I guess its ok or not?

dhschall (commented):
Thanks for the contribution @giactra! It looks quite good to me. I only had a few minor comments

src/cpu/o3/inst_queue.cc
giactra: It would be a circular call. IQUnit::remove is actually called by clearInIQ itself:

https://github.com/gem5/gem5/pull/2652/files/574985143937a6734558fbcf929442fce54edfe0

    void
    clearInIQ()
    {
        assert(iq);
        status.reset(IqEntry);
        iq->remove(this);
        iq = nullptr;
    }

src/cpu/o3/inst_queue.hh
giactra: Ack

src/cpu/o3/inst_queue.cc
giactra: Considering the resetState marks the IQ as being empty, I would say it means a CPU pipeline would drain only when the instructions have fully retired from the pipeline so the IQ pointer should already be zero... But it is a good point. Maybe we should consider adding an assertion...

giactra:
> @giactra Nice to see this support added! Can you please also add a config that uses multiple IQs?

Thanks @pranith. Makes sense.

I am happy to post something, but I think we need to answer the following questions first:

1) Where should it reside?
2) How do we size the IQs? (Either in absolute or relative terms)
3) Would this require a testing config?

Please let me know what you think about these questions. I'd be happy to post something when there is consensus about the bullet points above

pranith:
> 
>     1. Where should it reside?

How about configs/example/arm?

> 
>     2. How do we size the IQs? (Either in absolute or relative terms)

To compare against unified IQ, we should have similar total number of entries across all IQs.

> 
>     3. Would this require a testing config?

Yes, it would be good to have a test that exercises this code path so that it doesn't break.

src/cpu/o3/inst_queue.cc
dhschall: Oh I see! I thought this function can be called from anywhere.

erin-le:
@giactra: Hello, could you resolve the conflicts on this PR when you get a chance? Also, it looks like Pranith responded to your questions; did you have any thoughts or opinions about the responses? Thank you!

giactra:
> > ```
> > 1. Where should it reside?
> > ```
> 
> How about configs/example/arm?
> 

This is a bit complex for mainly two reasons

1) I believe an example should be ISA agnostic
2) There are political reasons why it is hard for me to push an example distribution in configs/example/arm as people
might think it is a statement on how IQs are split for different functional units in an Arm core.


> > ```
> > 2. How do we size the IQs? (Either in absolute or relative terms)
> > ```
> 
> To compare against unified IQ, we should have similar total number of entries across all IQs.
> 

That the total should be the same I agree. The question is more about how we split the total between different IQs. For the same reason as the point above (2), it could be a bit problematic for me come up with the. numbers myslef..

IMHO, we should consider providing an example in the default O3 python file, and basically have different IQs between INT,SIMD and MEM FUs. If you or someone else wants to come up with a reasonable distribution, I will add a new commit to the PR accordingly

> > ```
> > 3. Would this require a testing config?
> > ```
> 
> Yes, it would be good to have a test that exercises this code path so that it doesn't break.

Yes. probably worth adding a test with a separate PR once the example distributed IQ is in place

powerjg:
> Can you please also add a config that uses multiple IQs?

Let's not block this PR on the lack of a config file. It would be great if we could get an example from a neutral party. @dhschall any chance that you or one of your students could put together an example with a distributed inst queue, decoupled front end, and FDIP? That would be good to have for the release.

dhschall:
I agree @powerjg this should not block this PR and should be added in a later PR. I'll see what I can do. For now I've created a draft PR with a dummy config #2724 so it will not get lost.
I'll try to work on alligning the parameters ASAP.

powerjg:
> I agree @powerjg this should not block this PR and should be added in a later PR. I'll see what I can do. For now I've created a draft PR with a dummy config #2724 so it will not get lost. I'll try to work on alligning the parameters ASAP.

Thanks!! I wouldn't worry *too* much about parameter alignment. I think at this point having something that shows how to use the parameters would be good. That said, if we had publicly available configs with parameters like real hardware, that would be amazing!

dhschall (dismissed):
I've updated #2724 to use the new distributed instruction queue. 
IMO this looks good and I think we should go ahead and merge this PR so we can build a test using the new example config.

giactra:
> using

Thanks! I have rebased the PR and fixed the merge conflict

src/cpu/o3/IQUnit.py
```diff
@@ -0,0 +1,57 @@
...
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from m5.objects.FUPool import *
+from m5.objects.SMT import *
```
powerjg: Nit:

```suggestion
from m5.objects.FUPool import FUPool
from m5.objects.SMT import SMTQueuePolicy
```

src/cpu/o3/IQUnit.py
```diff
@@ -0,0 +1,57 @@
...
+from m5.params import *
+from m5.proxy import *
+from m5.SimObject import *
```
powerjg: Nit:

```suggestion
from m5.SimObject import SimObject
```

powerjg (dismissed):
lgtm. I am a bit concerned that all of the SMT code is totally untested, but I suppose it's no different than before.

src/cpu/o3/IQUnit.py
```diff
@@ -0,0 +1,60 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from m5.objects.FUPool import (
+    DefaultFUPool,
+    FUPool,
+)
+from m5.objects.SMT import SMTQueuePolicy
+from m5.params import *
+from m5.proxy import *
```
powerjg: Unused.

src/cpu/o3/SMT.py
```diff
@@ -0,0 +1,51 @@
...
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from m5.params import *
```
powerjg: Nit
```suggestion
from m5.params import ScopedEnum
```

powerjg (commented):
A couple of tiny things, but it lgtm.

src/cpu/o3/IQUnit.py
giactra: Used by numThread. Reverting to fix compilation

BobbyRBruce (approved):

Issue #2575, opened 2025-09-06
arch-riscv: Register overlap in `vslidedown` with `lmul > 1`


Joao-Pedro-Cabral:
**Describe the bug**
When running `vslidedown v2, v2, 0` with `lmul > 1` , I got a `Illegal Instruction` in `gem5`.

**Affects version**
I use `v25.0.0.0` version of `gem5`.

**gem5 Modifications**
I have added some new instructions to the Decoder.isa, but they aren't used in this code.

**To Reproduce**
I don't think it's necessary to show how to reproduce this issue.

I believe that I've already found which code line generates this illegal instruction, in file `src/arch/riscv/isa/templates/vector_arith.isa`:

```
def template VectorSlideMicroExecute {{

template<typename ElemType>
Fault
%(class_name)s<ElemType>::execute(ExecContext* xc,
                                  trace::InstRecord* traceData) const
{
    using vu [[maybe_unused]] = std::make_unsigned_t<ElemType>;
    using vi [[maybe_unused]] = std::make_signed_t<ElemType>;

    uint32_t num_microops = vtype_regs_per_group(vtype);
    if (slideUp || num_microops > 1) {
        uint32_t vs2_end = machInst.vs2 + num_microops - 1;
        uint32_t vd_end = machInst.vd + num_microops - 1;
        bool overlap = (machInst.vs2 >= machInst.vd && machInst.vs2 <= vd_end)
            || (vs2_end >= machInst.vd && vs2_end <= vd_end);
        if (overlap && slideUp) {
            return std::make_shared<IllegalInstFault>(
                "Regs overlap is not allowed for vslideup", machInst);
        } else if (overlap && !slideUp) {
            return std::make_shared<IllegalInstFault>(
                "Regs overlap is not allowed for vslidedown when lmul > 1",
                machInst);
        }
    }

    bool set_dirty = true;
    bool check_vill = true;
    Fault update_fault = updateVPUStatus(xc, machInst, set_dirty, check_vill);
    if (update_fault != NoFault) { return update_fault; }

    %(op_decl)s;
    %(op_rd)s;
    %(set_vlenb)s;

    [[maybe_unused]]const uint32_t vlmax = vtype_VLMAX(vtype, vlen);

    VM_REQUIRED();
    %(code)s;
    %(op_wb)s;

    return NoFault;
};

%(declare_varith_template)s;

}};
```

The message is generated by this `else if`:

```
else if (overlap && !slideUp) {
[...]

Issue #2621, opened 2025-10-01
arch-riscv: `O3CPU` cannot handle the dependecy graph of my code


Joao-Pedro-Cabral:
**Describe the bug**
I have a RVV Intrinsics code in which `O3CPU` aborts the program with this error: "src/cpu/o3/inst_queue.cc:1417: panic: Dependency graph 139 (vector) (flat: 651) not empty". When I ran the same code, but change one line in my gem5 config .py file to use `MinorCPU` instead of `O3CPU` my code works.

**Affects version**
I'm using `gem5/develop` branch in commit `22e4e8331d62601d1cb14ada7eab2ad9f77fd639`.

**gem5 Modifications**
My local code also has the commits from these two PRs: [#2604](https://github.com/gem5/gem5/pull/2604) and [#2619](https://github.com/gem5/gem5/pull/2619), but I believe that they aren't relevent in this situation.

**To Reproduce**
I'm using `gem5.fast` compiled with `scons build/RISCV/gem5.fast -j20`.

My code was compiled with this command:
```
riscv64-unknown-linux-musl-gcc -O2 -static -Wall -Wextra -o build/app_bin main.c compress.c
```

`main.c` file:

```

#include "main.h"

int main() {
  uint8_t r[KYBER_POLYVECCOMPRESSEDBYTES];
  polyvec a;

  for(int i = 0; i < KYBER_K; i++) {
    for(int j = 0; j < KYBER_N; j++) {
      a.vec[i].coeffs[j] = (i + 37*j)%KYBER_Q;
    }
  }

  polyvec_compress(r, &a);

  for(int i = 0; i < KYBER_POLYVECCOMPRESSEDBYTES; i++) {
    printf("r[%d]: %d\n", i, r[i]);
  }

  return 0;
}

```

`main.h` file:

```
#include <stdint.h>
#include <stdio.h>
#include <riscv_vector.h>

#define KYBER_N 256
#define KYBER_Q 3329
#define KYBER_K 2
#define KYBER_POLYVECCOMPRESSEDBYTES (KYBER_K * 320)

typedef struct {
  int16_t coeffs[KYBER_N];
} poly;

typedef struct {
  poly vec[KYBER_K];
} polyvec;

void polyvec_compress(uint8_t r[KYBER_POLYVECCOMPRESSEDBYTES], const polyvec *a);
```

`compress.c` file:
```

#include "main.h"

[...]

PR #2657, opened 2025-10-13
cpu-o3: incorrect `vtype` and `vl` generates `assert` fail in  `O3 IQ`

Joao-Pedro-Cabral:
As discussed here: [#2575](https://github.com/gem5/gem5/issues/2575), I got some unexpected exception in my code with the `IQ` dependency graph.

The bug happens because in the same tick the `IQ` have a `vsetvli` instruction, which will reduce the current `LMUL`, and another instruction (in this case a `vnsrl`), which depends on the new `vsetvli` configuration. However, depending on the `vs1/vs2` and the previous `LMUL` it is possible to have one of the source registers of the micro-instructions depending on a register which has never been written, causing an exception in the dependency graph, which terminates the program.  

The `O3` considers `vsetvl` instructions as control instructions, in which the result can only be inferred in the `Execute` Stage, since some variants of this instruction requires a source register. In special, `O3` always supposes that the new `vsetvl` instruction will not change the current configuration and when the prediction is wrong, the following instructions are flushed, which makes sense.

So in this case, I believe that the correct behavior is allowing speculative instructions to depend in any register, since for this situation they'll be flushed and then after executed with the correct vector configuration.

I know that removing this `assert` can allow some incorrect programs assembly-made programs to execute without problems (and maybe hide some new `gem5` bugs), but I don't know another feasible solution for this problem.

What do you think about it?

### DIFF
--- a/src/cpu/o3/inst_queue.cc
+++ b/src/cpu/o3/inst_queue.cc
@@ -1329,8 +1329,9 @@ InstructionQueue::doSquash(ThreadID tid)
             if (dest_reg->isFixedMapping()){
                 continue;
             }
-            assert(dependGraph.empty(dest_reg->flatIndex()));
-            dependGraph.clearInst(dest_reg->flatIndex());
+            if (!dependGraph.empty(dest_reg->flatIndex())) {
+                dependGraph.clearInst(dest_reg->flatIndex());
+            }
         }
         instList[tid].erase(squash_it--);
         ++iqStats.squashedInstsExamined;
@@ -1412,7 +1413,8 @@ InstructionQueue::addToProducers(const DynInstPtr &new_inst)
             continue;
         }
 
-        if (!dependGraph.empty(dest_reg->flatIndex())) {
+        if (!dependGraph.empty(dest_reg->flatIndex()) &&
+            new_inst->isNonSpeculative()) {
             dependGraph.dump();
             panic("Dependency graph %i (%s) (flat: %i) not empty!",
                   dest_reg->index(), dest_reg->className(),


giactra:
Hi @Joao-Pedro-Cabral, this PR should be tagged as "cpu-o3" as it is effectively touching cpu-o3 code. I have no familiarity with the riscv ISA, but from a quick look at the problem description it looks to me you need to establish explicit relationship between producer (vsetvli) and consumer (vnsrl) and not relax the dependency graph

Joao-Pedro-Cabral:
Hello @giactra , I changed the PR name.
In this case, `vsetvl` isn't the producer of `vnsrl` source registers, but the problem here isn't correlated only to `vnsrl` instruction, it can happens with any instructions that are processed by the `IQ` in the same tick as its vector configuration instruction `vsetvl`.

I personally, don't have any better idea about how to fix it (since the configuration can only be determined in the Execute stage, so after the others micro-instructions have been created).

giactra:
> Hello @giactra , I changed the PR name. In this case, `vsetvl` isn't the producer of `vnsrl` source registers, but the problem here isn't correlated only to `vnsrl` instruction, it can happens with any instructions that are processed by the `IQ` in the same tick as its vector configuration instruction `vsetvl`.
> 

Sorry I should have been clearer. When I say vsetvl is the producer I mean that it produces information upon which next instructions in program order depend upon. In this case it seems to me the "produced" data is the new vector length. Not necessarily a register value.

So either you generate an explicit dependency or you could alternatively mark the vsetvl as "IsSerializeAfter". Note: the latter is a heavyweight solution.

Let me reiterate again I know very little about the ISA; I am coming more from a generic O3CPU knowledge so take things with a pinch of salt :)

Joao-Pedro-Cabral:
> > Hello @giactra , I changed the PR name. In this case, `vsetvl` isn't the producer of `vnsrl` source registers, but the problem here isn't correlated only to `vnsrl` instruction, it can happens with any instructions that are processed by the `IQ` in the same tick as its vector configuration instruction `vsetvl`.
> 
> Sorry I should have been clearer. When I say vsetvl is the producer I mean that it produces information upon which next instructions in program order depend upon. In this case it seems to me the "produced" data is the new vector length. Not necessarily a register value.
> 
> So either you generate an explicit dependency or you could alternatively mark the vsetvl as "IsSerializeAfter". Note: the latter is a heavyweight solution.
> 
> Let me reiterate again I know very little about the ISA; I am coming more from a generic O3CPU knowledge so take things with a pinch of salt :)

For me is the opposite :). I think I understand reasonably well the RISC-V part, but idk basically nothing about the `gem5` `O3` implementation.

I understand at high level your idea, but I don't know how can I code this dependency for the instructions after the `vsetvl` :( .

powerjg:
We made a decision a while ago to make `vsetvli` a *control* instruction. So, any instruction that depends on the vector length, has a *control* dependency on `vsetvli`. We decided that we would allow the branch predictor to "predict" the vector length for younger instructions until the `vsetvli` commits. If the vector length changes, then we squash the in-flight instructions.

I think this change is reasonable since now we have speculative instructions that could be accessing registers that "don't exist" (if I understand the code correctly).

saul44203:
Assuming you are talking about #2621 (correct me if it's not)  from what I see the problem is that the instruction is constructed with a mispredicted LMUL=4 and one of the µops has an overlap on the source and destination register that's not taken into account with a vcpy, and that causes the dependency graph error, since the destination is pinned and both (src/dest) are pointing to the same physical register.

The issue here is that a vcpy µop is not being created for that case, as it only happens once when the lowest part of the source register group matches the destination: https://github.com/gem5/gem5/blob/ddd4ae35adb0a3df1f1ba11e9a973a5c2f8c2944/src/arch/riscv/isa/templates/vector_arith.isa#L516-L522

I think this is correct per spec for narrowing instructions (from section 5.2 of RISC-V's vector extension):
> The destination EEW is smaller than the source EEW and the overlap is in the lowest-numbered part of the source
register group (e.g., when LMUL=1, vnsrl.wi v0, v0, 3 is legal, but a destination of v1 is not).

but since misprediction can happen, illegal overlaps can occur like the one here.

So maybe the solution would be to allow this overlap to happen correctly (i.e. with it's corresponding vcpy) since the branch will be squashed eventually.

Joao-Pedro-Cabral:
@saul44203 I liked your idea, do you want that I try it on my code and then modify this PR (removing the modifications in the `O3` CPU code)?

Joao-Pedro-Cabral:
@saul44203 but there are others instructions in the ISA that isn't legal to have source and destiny registers overlapping, so we need to allow the overlap for every instruction, no?

saul44203:
> @saul44203 I liked your idea, do you want that I try it on my code and then modify this PR (removing the modifications in the `O3` CPU code)?

@Joao-Pedro-Cabral Sure, feel free to implement this solution. 

> @saul44203 but there are others instructions in the ISA that isn't legal to have source and destiny registers overlapping, so we need to allow the overlap for every instruction, no?

Yes, there might be some other instructions with the same problem. Those that pin the destination and can overlap with a misspeculated LMUL, but do not scale the number of vcpys with it, can be subject to this error.

erin-le:
@Joao-Pedro-Cabral: Hello, just pinging to let you know that the staging branch for gem5 v25.1 will be created this Friday. If you're finished with modifying this PR on your end, and if this PR gets approved and merged into `develop` before then, it will make it into the next release. Thanks for your contribution to gem5!

whisperity:
I can not comment on the implementation details here because I am not intimately familiar with Gem5, but I have to report that applying this patch locally was instrumental to allow us running the TSVC benchmark ("Test Suite for Vectorising Compilers", https://github.com/UoB-HPC/TSVC_2, https://github.com/llvm/llvm-test-suite/tree/main/MultiSource/Benchmarks/TSVC) when compiled to RISC–V+RVV (both in the case of GCC and Clang) in Gem5. Prior to the application of this patch, the test suite crashed in:

```
src/cpu/o3/inst_queue.cc:1416: panic: Dependency graph 95 (vector) (flat: 607) not empty!
```

So this seems to be an issue you can trigger in the wild with seemingly innocent loops that happen to get vectorised.

powerjg (approved):

giactra:
Can we avoid fast merging sensitive changes (like to the o3 CPU) and wait for people who made comments and proposed alternative solutions to chime in?

powerjg:
@giactra Apologies. I thought that your comment was addressed back in October.

It looks like this PR does cause regressions, so you were right to be concerned. See https://github.com/gem5/gem5/pull/2914

@Joao-Pedro-Cabral I don't know what the right way to solve this problem is. I think that something deeper is going wrong.

saul44203:
Hey all,

I can try to implement the solution I proposed in my previous response, no ETA though.

PR #2665, opened 2025-10-14
cpu-minor: explicitly check if instruction in ROM during decoding

giragon6:
Check if current micro-op is in ROM, and, if so, fetch the instruction from ROM when decoding. This fixes the `invalid micro-op!` panic that occurs when running the SPEC2017 benchmarks with the Minor CPU and [matches the corresponding behavior in the O3 CPU](https://github.com/gem5/gem5/blob/ddd4ae35adb0a3df1f1ba11e9a973a5c2f8c2944/src/cpu/o3/fetch.cc#L1238).

### DIFF
--- a/src/cpu/minor/decode.cc
+++ b/src/cpu/minor/decode.cc
@@ -164,9 +164,8 @@ Decode::evaluate()
                 StaticInstPtr parent_static_inst = NULL;
                 MinorDynInstPtr output_inst = inst;
 
-                auto* dec_ptr = cpu.getContext(
-                                               inst->id.threadId)
-                                               ->getDecoderPtr();
+                auto *dec_ptr =
+                    cpu.getContext(inst->id.threadId)->getDecoderPtr();
 
                 if (inst->isFault()) {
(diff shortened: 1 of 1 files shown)


powerjg (commented):
lgtm... I have to say, though, if this is how clang-format chose to format these function calls, I do not like it.

Just one question: Did clang-format make this choice?

giragon6:
Oops -- that's my bad. My editor did something weird with it, and I forgot to run clang-format before committing. Should be fixed now, sorry about that!

powerjg (dismissed):

erin-le:
@giragon6: Hello, it looks like the `clang-format-check` CI test is still failing. Could you run the following command (from the CI test) and push the changes? It will run `git-clang-format` on all of your commits. Thank you!

```bash
python util/run-git-clang-format.py --ci-pr-base-commit 3c4dedb088e22d2a62dd4aa4d056d281a1310082
```

erin-le (approved):

PR #2668, opened 2025-10-16
cpu-kvm: added support for hosts with larger page size

callumgran:
On systems using large base pages (e.g. my given testbench with a Neoverse V2), KVM rejects user memory regions that are not aligned to the host page size. This causes failures when loading small regions such as theDTB, which was in my case smaller than 64 KiB.

This change pads any KVM memory region to the host page size before mapping it. The padded buffer is tracked and freed when the slot is disabled. This prevents startup failures on hosts with large page sizes without affecting behavior on 4 KiB hosts.

The current ownership tracking uses an unordered_map and an ugly struct to manage temporary padded buffers. This can probably be improved.

Here is the error that occurs without the given code:

```
src/arch/arm/fs_workload.cc:121: info: Using bootloader at address 0x10 
src/arch/arm/fs_workload.cc:139: info: Using kernel entry physical address at 0x80080000 
src/arch/arm/linux/fs_workload.cc:111: info: Loading DTB file: m5out/device.dtb at address 0x88000000 
src/cpu/kvm/vm.cc:470: panic: Failed to setup KVM memory region: 
Host Address: 0x Size: %ll 
Guest Address: 0x Flags: 0x%x 

<bad format><bad format><bad format><bad format> 
```

NOTE: This is my first real open source PR, so I hope I haven't forgotten anything or that I'm breaking rules of practice.

### DIFF
--- a/.github/workflows/compiler-tests.yaml
+++ b/.github/workflows/compiler-tests.yaml
@@ -59,4 +59,4 @@ jobs:
               # The address sanitizer is quite memory intensive; triggering OOM errors on smaller runners.
               # To mitigate this, we limit scons to a single thread (`-j1`)
               run: /usr/bin/env python3 /usr/bin/scons --ignore-style build/ALL/gem5${{ matrix.opts }} -j1 --with-${{ matrix.sanitizer }}
-              timeout-minutes: 600 # 10 hours
+              timeout-minutes: 1200 # 20 hours


Harshil2107:
Hi @callumgran,

I am not an expert but the changes look fine to me. Please fix the clang format errors in the changes to get the CI tests to pass.

I also tested the change on an X86 and ARM machine with KVM. This change doesnot affect the original behaviour in machines with page size of 4096 bytes and fullsystem simulation run as expected.

callumgran:
@Harshil2107 The formatting issue should be fixed now, is there anything more I need to do?

I really don't understand why it fails the clang-format-check. It passes all tests locally and `python util/run-git-clang-format.py` works just fine...

src/cpu/kvm/vm.cc
```diff
@@ -457,23 +458,61 @@ KvmVM::setUserMemoryRegion(uint32_t slot,
                            void *host_addr, Addr guest_addr,
                            uint64_t len, uint32_t flags)
 {
-    struct kvm_userspace_memory_region m;
+    const long page_sysconf = sysconf(_SC_PAGESIZE);
+
+    // 4096 is the most common page size
+    const size_t page_size = (page_sysconf > 0) ? (size_t)page_sysconf : 4096;
```
powerjg: Nit: It may be a good idea to add the following here. I'll let you make the call.
```suggestion
    warn_if(page_sysconf <= 0, "Defaulting to 4 KiB page size for KVM alignment");
    const size_t page_size = (page_sysconf > 0) ? (size_t)page_sysconf : 4096;
```

powerjg (dismissed):
One nit below. Generally, this looks good to me.

powerjg:
> @Harshil2107 The formatting issue should be fixed now, is there anything more I need to do?
> 
> I really don't understand why it fails the clang-format-check. It passes all tests locally and `python util/run-git-clang-format.py` works just fine...

I think the problem is that `util/run-git-clang-format.py` only checks what is staged, not what is committed. It makes it a big PITA to reformat things once you've committed it. 

I asked gemini how to do it...

---

The process of "unstaging" the last commit, running the formatter, and then recreating the commit to force push involves using specific `git` commands to manage your history and staging area.

Here are the step-by-step commands to achieve this:

## ⚙️ Steps to Unstage, Reformat, and Amend

This sequence assumes you are on the branch with the faulty commit and that the commit is the most recent one (`HEAD`).

### 1\. Unstage the Last Commit (HEAD)

Use `git reset` with the **mixed** mode (which is the default) to uncommit the changes, but keep them in your working directory as unstaged changes.

  * **Command:**
    ```bash
    git reset HEAD^
    ```
  * **Explanation:**
      * `git reset` moves the current branch pointer (`HEAD`) back one commit.
      * `HEAD^` (or `HEAD~1`) refers to the parent of the current commit.
      * The **default** mode (`--mixed`) unstages the changes, meaning all files from the original commit are now in your working directory but are *not* staged. This is exactly what is needed for `util/run-git-clang-format.py` to pick up and reformat them before staging.

### 2\. Run the Formatter

Now that the changes are in your working directory and not yet staged, run the formatting script. Since `util/run-git-clang-format.py` only reformats what is staged, you first need to stage the changes. However, typically, **local formatting scripts** (like a custom wrapper around `clang-format -i`) are designed to be run against the *unstaged* files in your working directory.

Since you are specifically referencing a script that checks **staged** content, you should stage the changes *first*, let the tool re-stage the reformatted content, or simply run the tool against the unstaged files if it supports in-place modifications (`-i`).

Assuming your utility script **only reformats files that are staged** and then *diffs* against the commit you are on (which is now `HEAD^`), or simply **formats the files in your working directory** that are different from `HEAD^`, the general step is:

  * **Run the Formatter:**
    ```bash
    # Run the utility script
    util/run-git-clang-format.py
    ```
  * **Result:** This script should modify the files in your **working directory** (the uncommitted changes) to comply with the code style.

### 3\. Restage the Changes

After the formatter runs, your files are modified in the working directory and are ready to be staged.

  * **Command:**
    ```bash
    git add .
    ```
  * **Explanation:** This stages all the modified files, including the formatting changes.

### 4\. Amend the Original Commit

The staged changes are now the clean, formatted version of the original commit. Use `git commit --amend` to replace the old, unformatted commit with the new, formatted one.

  * **Command:**
    ```bash
    git commit --amend --no-edit
    ```
  * **Explanation:**
      * `git commit --amend` takes the current staged changes and combines them with the previous commit (`HEAD` before the reset), replacing it entirely.
      * `--no-edit` keeps the original commit message, so you don't have to open the text editor. If you need to change the message, just omit this flag.

### 5\. Force Push to the PR Branch

Because you have re-written the commit history (the original commit's hash has changed), you must use a force push to update the remote branch and the PR.

  * **Command (Use this one if safe):**
    ```bash
    git push --force-with-lease
    ```
  * **Recommended Explanation:** **Always use** `git push --force-with-lease` instead of `git push --force`. The `--force-with-lease` option is safer because it prevents you from accidentally overwriting work on the remote branch that you haven't pulled locally. It's a "safe force-push."

-----

## 💡 Summary of Commands

For easy copy/pasting:

```bash
# 1. Unstage the last commit, keeping changes in the working directory
git reset HEAD^

# 2. Run the formatting utility (it should modify files in place)
util/run-git-clang-format.py

# 3. Stage the newly formatted changes
git add .

# 4. Replace the old commit with the new, formatted changes, keeping the message
git commit --amend --no-edit

# 5. Force push the rewritten history to the remote branch
git push --force-with-lease
```

erin-le:
@callumgran: Hello, for the `clang-format-check` CI Test, could you try running the following command? It should run git-clang-format on all of the commits in this PR. For future reference, this command was obtained by clicking onto the `clang-format-check` CI Test, and I believe it takes the last commit before the first commit of the PR as the argument.

```
python util/run-git-clang-format.py --ci-pr-base-commit e7185b9975dad3d2b17b23c113ce2fab1f8f1234
```

CC: @powerjg 
Running this does make changes that still need to be committed (rather than inserting the changes into the previous commits), but I think that should be okay for this PR, since it looks like this PR should be fine to squash-merge.

src/cpu/kvm/vm.cc
callumgran: I agree, it's probably best to give users a hint in case sysconf fails and they have a different page size. I would hate to trying to debug that without any hints.

callumgran:
@Harshil2107 @powerjg @erin-le The commit history is now fubar, so I'm regrettably creating a new branch with a singular commit being my changes with the formatting correct.

https://github.com/gem5/gem5/pull/2737

PR #2670, opened 2025-10-17
cpu-o3: Instant ROB squash

dhschall:
The squashWidth limits the number of instructions squashed
within on cycle. In case of a full, large ROB this can incure tens
of cycles during which the rename stage is blocked. E.g. with
160 instructions in the ROB and squash width of 8 rename will be
blocked 20 cycles.
With this change users can set the squash width to a negative number
and then all instructions will be squashed immediatly.
In a real implementation it is more likely the case where this can
simply be done with updating a pointer.

### DIFF
--- a/src/cpu/o3/BaseO3CPU.py
+++ b/src/cpu/o3/BaseO3CPU.py
@@ -136,7 +136,10 @@ class BaseO3CPU(BaseCPU):
     )
     renameToROBDelay = Param.Cycles(1, "Rename to reorder buffer delay")
     commitWidth = Param.Unsigned(8, "Commit width")
-    squashWidth = Param.Unsigned(8, "Squash width")
+    squashWidth = OptionalParam.Unsigned(
+        "Squash width. If unspecified all instructions are "
+        "squashed instantly within one cycle.",
+    )
     trapLatency = Param.Cycles(13, "Trap latency")
     fetchTrapLatency = Param.Cycles(1, "Fetch trap latency")
 
--- a/src/cpu/o3/rob.cc
+++ b/src/cpu/o3/rob.cc
@@ -64,6 +64,7 @@ ROB::ROB(CPU *_cpu, const BaseO3CPUParams &params)
       numThreads(params.numThreads),
       stats(_cpu)
 {
+    assert(!squashWidth.has_value() || (squashWidth > 0));
     //Figure out rob policy
     if (robPolicy == SMTQueuePolicy::Dynamic) {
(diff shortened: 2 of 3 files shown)


src/cpu/o3/BaseO3CPU.py
```diff
@@ -136,7 +136,11 @@ def support_take_over(cls):
     )
     renameToROBDelay = Param.Cycles(1, "Rename to reorder buffer delay")
     commitWidth = Param.Unsigned(8, "Commit width")
-    squashWidth = Param.Unsigned(8, "Squash width")
+    squashWidth = Param.Int(
```
BobbyRBruce: A few months ago, someone added "OptionalParam", which allowed us to have optional parameters. I think it'd be good to use them here instead of the utilization of negatives. I.e., if the squashWidth is not specified, it's assumed to be infinite, and all instructions are squashed within one cycle.

src/cpu/o3/BaseO3CPU.py
```diff
@@ -136,7 +136,11 @@ def support_take_over(cls):
...
+        -1,
+        "Squash width. A negative number means all instructions are "
+        "squashed instantly within on cycle.",
```
BobbyRBruce:
```suggestion
        "squashed instantly within one cycle.",
```

src/cpu/o3/rob.cc
```diff
@@ -316,7 +316,7 @@ ROB::doSquash(ThreadID tid)
 
     bool robTailUpdate = false;
 
-    unsigned int numInstsToSquash = squashWidth;
+    auto numInstsToSquash = (squashWidth < 0) ? numEntries : squashWidth;
```
BobbyRBruce: What would it mean if the squashWidth was 0?

BobbyRBruce (changes requested):

src/cpu/o3/BaseO3CPU.py
dhschall: Thanks for pointing this out! That's cool, didn't know about that.

src/cpu/o3/rob.cc
dhschall: It would not squash anything. Thats true. While this behaviour existed already before I'll add an assert in the constructor

dhschall:
@BobbyRBruce I've made the `squashWidth` an optional parameter now. Let me know if that's the correct way to do.

giactra (approved):

BobbyRBruce (approved):

PR #2671, opened 2025-10-17
dev: rename PCI devices base classes

clemdiep:
This commit rename PciEndpoint and PciBridge to PciType0Device and PciType1Device to avoid confusion with proliferation of PCI bridges names, as mentionned in https://github.com/gem5/gem5/pull/2561#issuecomment-3381337325

It also add documentation to those classes.

### DIFF
--- a/src/dev/pci/PciDevice.py
+++ b/src/dev/pci/PciDevice.py
@@ -216,9 +216,9 @@ class PciEndpoint(PciDevice):
     MinimumGrant = Param.UInt8(0x00, "Minimum Grant")
 
 
-class PciBridge(PciDevice):
-    type = "PciBridge"
-    cxx_class = "gem5::PciBridge"
+class PciType1Device(PciDevice):
+    type = "PciType1Device"
+    cxx_class = "gem5::PciType1Device"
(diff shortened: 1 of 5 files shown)


giactra:
Thanks @clemdiep for this PR.

Apologies for the request, but would you consider keeping the PciEndpoint name unchanged? I understand your solution is better for symmetry but it would be nice if we could keep disruption to the minimum. You could add a PciType0 as a rename of PciEndpoint if you really wanted to:

```
class PciType0(PciEndpoint):
    pass
```

clemdiep:
> Thanks @clemdiep for this PR.
> 
> Apologies for the request, but would you consider keeping the PciEndpoint name unchanged? I understand your solution is better for symmetry but it would be nice if we could keep disruption to the minimum. You could add a PciType0 as a rename of PciEndpoint if you really wanted to:
> 
> ```
> class PciType0(PciEndpoint):
>     pass
> ```

Okay, I don't care that much about it. Will revert this!

abmerop (approved):
lgtm

giactra (approved):

PR #2672, opened 2025-10-17
dev: refactor PCI configuration to use RegisterBank

clemdiep:
This commit refactors PCI configuration space to use RegisterBank. For that, the configuration header and capabilities of a device are now child classes of RegisterBank and stored in a common map with their base offset as key. A PCI device can define their own RegisterBank if needed and add them to the configuration map, like IdeCtrl do.

This remove the need of extending readConfig and writeConfig function in PciDevice subclasses. Actions that are specific to a register (e.g. sending range change on BAR writes) can be implemented using writer/reader from Register class. And if needed, readConfig and writeConfig can still be overriden for more complex implementation.

Capabilities are also now their own classes, instead of being parameter in PciDevice which weren't always used.

Finally, there is a checkpoint upgrader to correctly set the checkpoint for the new serilization, which would need to be apply for the test to pass.

Made in collaboration with @marinazapater and @karnapathak


clemdiep:
Just to mention that this PR will enter in conflict with #2561. So depending on which of the two get merged first, I will solve those conflicts on the other

PR #2675, opened 2025-10-20
mem-ruby: Add support for CLFLUSH type instructions in MESI Three Level protocol

### LINKED ISSUE #2642: clflush in Ruby Cache
`src/mem/packet.hh` says that clflush instruction leads to `CleanInvalidReq` flag being set in the packet. 
```cpp
/**
 * Is this packet a clean invalidate request, e.g., clflush/clflushopt?
 */
bool
isCleanInvalidateRequest() const
{
    return cmd == MemCmd::CleanInvalidReq;
}
```
However, when the packet is translated to a `RubyRequest` in `sequencer.cc`, it checks for `pkt->isFlush()` and not `pkt->isCleanInvalidateRequest()`.
```cpp
} else if (pkt->isFlush()) {
    primary_type = secondary_type = RubyRequestType_FLUSH;
} else if (pkt->cmd == MemCmd::MemSyncReq) {
```
 Is this a potential mistake in the implementation, or am I interpreting it incorrectly? When running my workload I am observing that clflush instruction does not clear the cache indeed. I am running full system linux kernel on x86.

AmoghBhagwat:
Aims to fix the issue referenced [here](https://github.com/gem5/gem5/issues/2642). Correctness tests done:

- Verified that cache line is actually being flushed by measuring time of access
- Tested correctness of writing dirty data. Cache lines shared among cores are also flushed correctly. Example program:
```c
#include <stdio.h>
#include <stdlib.h>
#include <sys/mman.h>
#include <stdint.h>
#include <string.h>
#include <unistd.h>
#include <sched.h>

void clflush(void *ptr) {
    asm volatile(
        "clflush (%0)"
        :
        : "r"(ptr)
        : "memory"
    );
}

void mfence() {
    asm volatile("mfence" ::: "memory");
}

int main() {
    int data = 42;
    printf("Before clflush: data = %d\n", data);
    clflush(&data);
    printf("After clflush: data = %d\n", data);

    volatile char* x = (char*) mmap(NULL, 256*sizeof(char), PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
    x[255] = 1;
    int pid = fork();
    if (pid == 0) {
        for (int i = 0; i < 128; i++) {
            x[i] = 1;
        }
        mfence();
        x[255] = 0;
        mfence();
        return 0;
    }

    while (x[255] == 1);
    clflush(&x[0]);
    clflush(&x[64]);
    int sum = 0;
    for (int i = 0; i < 128; i++) sum += x[i];

    printf("sum = %d\n", sum);

    return 0;
}

```
sum is correctly printed as 128.

Note that for this to work, these lines (279-284) need to be removed from `RubyPort.cc`. I have not included this in the PR as it would break other protocols.
```cpp
if (pkt->req->isCacheMaintenance()) {
    warn_once("Cache maintenance operations are not supported in Ruby.\n");
    pkt->makeResponse();
    schedTimingResp(pkt, curTick());
    return true;
}
```

### DIFF
--- a/src/mem/ruby/protocol/MESI_Three_Level-L0cache.sm
+++ b/src/mem/ruby/protocol/MESI_Three_Level-L0cache.sm
@@ -109,6 +109,10 @@ machine(MachineType:L0Cache, "MESI Directory L0 Cache")
     // LL/SC states
     LLSC_E, AccessPermission:Read_Only, desc="a L1 cache entry Exclusive (LL/SC locked)";
     LLSC_M, AccessPermission:Read_Write, desc="a L1 cache entry Modified (LL/SC locked)";
+
+    // Flush states
+    I_F, AccessPermission:Busy, desc="Sent flush to L1, was in invalid";
+    V_F, AccessPermission:Busy, desc="Sent flush to L1, was valid";
   }
 
   // EVENTS
@@ -152,6 +156,11 @@ machine(MachineType:L0Cache, "MESI Directory L0 Cache")
     Load_Linked,         desc="Load-Linked request from the home processor";
     LLSC_Data_Exclusive, desc="Data for processor (start LL/SC  lock timer)";
     LLSC_Timeout,        desc="LL/SC lock timeout";
+
+    // Flush transitions
+    Flush, desc="Receive flush from processor";
+    Flush_Ack, desc="Flush ack from L1";
+    External_Flush_Data, desc="Send data to L1 during flush";
   }
 
   // TYPES
@@ -303,6 +312,8 @@ machine(MachineType:L0Cache, "MESI Directory L0 Cache")
     } else if ((type == RubyRequestType:ST) || (type == RubyRequestType:ATOMIC)
                || (type == RubyRequestType:Store_Conditional)) {
       return Event:Store;
+    } else if (type == RubyRequestType:FLUSH) {
+      return Event:Flush;
     } else {
       error("Invalid RubyRequestType");
     }
@@ -481,6 +492,10 @@ machine(MachineType:L0Cache, "MESI Directory L0 Cache")
           trigger(Event:Fwd_GETS, in_msg.addr, cache_entry, tbe);
         } else if (in_msg.Class == CoherenceClass:GET_INSTR) {
           trigger(Event:Fwd_GET_INSTR, in_msg.addr, cache_entry, tbe);
+        } else if (in_msg.Class == CoherenceClass:FLUSH_ACK) {
+          trigger(Event:Flush_Ack, in_msg.addr, cache_entry, tbe);
+        } else if (in_msg.Class == CoherenceClass:FLUSH_DATA) {
+          trigger(Event:External_Flush_Data, in_msg.addr, cache_entry, tbe);
         } else {
           error("Invalid forwarded request type");
         }
@@ -550,6 +565,9 @@ machine(MachineType:L0Cache, "MESI Directory L0 Cache")
             if (in_msg.Type == RubyRequestType:Store_Conditional) {
                 // if the line is not valid, it can't be locked
                 trigger(Event:Failed_SC, in_msg.LineAddress,
+                        Dcache_entry, TBEs[in_msg.LineAddress]);
+            } else if (in_msg.Type == RubyRequestType:FLUSH) {
+                trigger(Event:Flush, in_msg.LineAddress,
                         Dcache_entry, TBEs[in_msg.LineAddress]);
             } else {
               // Check to see if it is in the OTHER L0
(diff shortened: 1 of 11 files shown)


AmoghBhagwat:
@erin-le any comments about the PR? The clang format test is failing because of the .sm files, perhaps a separate PR is needed for fixing it.

powerjg:
> erin-le any comments about the PR? The clang format test is failing because of the .sm files, perhaps a separate PR is needed for fixing it.

We shouldn't have clang-format run on .sm files. If that's what's happening here, we should fix the pre-commit and the CI test.

powerjg (commented):
lgtm. Thanks for the contribution!!!

One (hopefully small) request: If it's possible, can you run the ruby_random_test with flushes enabled with this PR? It may be that that the random tester is broken, so this may not be easily feasible. If so, just let me know.

AmoghBhagwat:
Updated the ruby random tester, also I have changed a few things, earlier I was writing the flushed block to memory even if it was not modified. Now it only writes back if it is dirty.

src/mem/ruby/protocol/MESI_Three_Level-L0cache.sm
```diff
@@ -152,6 +156,11 @@ machine(MachineType:L0Cache, "MESI Directory L0 Cache")
     Load_Linked,         desc="Load-Linked request from the home processor";
     LLSC_Data_Exclusive, desc="Data for processor (start LL/SC  lock timer)";
     LLSC_Timeout,        desc="LL/SC lock timeout";
+
+    // Flush transitions
+    Flush, desc="Receive flush from processor";
+    Flush_Ack, desc="Flush ack from L1";
+    Flush_Data, desc="Send data to L1 during flush";
```
powerjg: Nit: Could you call this something like "External_Flush_data"? From what I understand (correct me if I'm wrong). This event happens when a higher level cache receives a flush request for which this cache owns the data.

powerjg (commented):
Just one tiny nit. 

Thanks for doing this! Have you tested taking checkpoints with this protocol?

src/mem/ruby/protocol/MESI_Three_Level-L0cache.sm
AmoghBhagwat: Yes you are right, did the changes.

AmoghBhagwat:
> Have you tested taking checkpoints with this protocol?

I had not tested it before, seems to be working fine for `tests/gem5/checkpoint_tests/configs/x86-fs-save-checkpoint.py` and ``tests/gem5/checkpoint_tests/configs/x86-fs-restore-checkpoint.py`` by using MESI_Three_Level hierarchy, and restoring the appropriate checkpoint. However, it is not working for `x86-hello-save-checkpoint.py`, I am not sure why. I am getting the following error while creating the checkpoint. To reproduce, the lines in `RubyPort.cc` need to be removed as mentioned in the PR description.
```
src/mem/ruby/common/Consumer.cc:89: void gem5::ruby::Consumer::processCurrentEvent(): Assertion `em->clockEdge() == *curr' failed.
```

powerjg:
Great! Thanks. A couple of other things:

> However, it is not working for x86-hello-save-checkpoint.py, I am not sure why. I am getting the following error while creating the checkpoint. To reproduce, the lines in RubyPort.cc need to be removed as mentioned in the PR description.

I think this is OK. I have a feeling that checkpointing with Ruby doesn't work generally in SE mode. If it's works in FS mode. I'm happy!

> Note that for this to work, these lines (279-284) need to be removed from RubyPort.cc. I have not included this in the PR as it would break other protocols.

There's a new(-ish) interface in Ruby/SLICC to add metadata to protocols. See https://github.com/gem5/gem5/blob/develop/src/mem/ruby/slicc_interface/ProtocolInfo.hh and https://github.com/gem5/gem5/blob/develop/src/mem/ruby/protocol/chi/CHI.slicc#L1 for an example use case.

I think you can extend the `ProtocolInfo` object with "supports flushes" and restructure the if statement in RubyPort.cc.

powerjg (dismissed):
Thanks, this is great!

erin-le:
@AmoghBhagwat: Hello, thanks for this PR! Could you run the following command in your `gem5` directory and push the changes? This is just to get the `clang-format-check` CI test to pass. Thank you!

```
python util/run-git-clang-format.py --ci-pr-base-commit d65257daf3f60d000cb33e4b12e79bdbeb02b878
```

AmoghBhagwat:
Not sure why the CI test is failing, running
`python util/run-git-clang-format.py --ci-pr-base-commit 52af36db4090ce41f5a02d207afa0cb4ae84c394`
locally does not change anything.

erin-le:
@AmoghBhagwat: I was able to see a change when I ran the command, but used `python3` instead of `python`, which could have made a difference. I'll push the change as a commit and see if that causes the `clang-format-check` to pass.

I also tried running the command using the commit before the first commit of the PR (d65257daf3f60d000cb33e4b12e79bdbeb02b878), but this caused `git-clang-format` to modify a file that this PR hadn't changed. I'm not sure why this happened.

erin-le (approved):

powerjg (approved):

PR #2679, opened 2025-10-21
dev: Temporary fix for X86 board hanging forever

clemdiep:
The example x86-ubuntu-run.py is currently hanging out due to the new PCI host bridge architecture (#2298). It's seems to be related to some timing added by the PCI bus. This cause the daily tests to fail due to a timeout.

This fix removes the timing added by the PCI bus, which means that all timing are back to (nearly) the same as before when there was no real PCI hierarchy.

As PCI hierarchy should add some delay at some point, this is a temporary fix and a futur PR will be made with a better fix.

### DIFF
--- a/src/dev/pci/PciUpstream.py
+++ b/src/dev/pci/PciUpstream.py
@@ -51,11 +51,13 @@ class PciBus(NoncoherentXBar):
 
     config_error_port = RequestPort("Port to send config errors to")
 
-    # Set some default values bases on IOXBar
+    # FIXME: There should have some latency added to transfert packets within
+    # the PCI bus/hierarchy, but adding them breaks X86 Board which will hang
+    # forever.
     width = 16
-    frontend_latency = 2
-    forward_latency = 1
-    response_latency = 2
+    frontend_latency = 0
+    forward_latency = 0
+    response_latency = 0
 
 
 class PciConfigError(IsaFake):


clemdiep:
@BobbyRBruce This is a temporary fix for the bug mentionned in https://github.com/gem5/gem5/pull/2561#issuecomment-3410645490.

I will keep looking for a better fix as those latency shouldn't be an issue, but at least this will make daily test pass again.

Also, side note, this segfault, mentionned in your comment, was already present before #2298 was merge and the system can boot with it. So I'm not sure a which point this is an issue?

```
[    1.098094] systemd[1]: Hostname set to <gem5>.
[    1.146765] netplan[103]: segfault at 0 ip 00007995216943e8 sp 00007ffe3c940640 error 4 in libglib-2.0.so.0.8000.0[79952166e000+a0000] likely on CPU 0 (core 0, socket 0)
[    1.146884] Code: fa 01 48 19 f6 83 e0 01 89 c2 0f b6 47 18 48 83 e6 e0 8d 0c 12 48 83 c6 40 83 e0 fc 09 d0 09 c8 88 47 18 31 ff e8 48 f4 01 00 <48> 8b 3b be 04 00 00 00 48 89 43 20 48 89 43 30 e8 e3 f4 01 00 48
[    1.160627] (sd-exec-[101]: /usr/lib/systemd/system-generators/netplan terminated by signal SEGV.
```

giactra:
Thanks @clemdiep.

I have a comment here:

If changing the latency fixes the problem, are we sure we are not simply (un)exposing a different problem somewhere else? In other words, what happens if we go back before your PCI change and artificially add a delay? Would we experience the same segfault?

clemdiep:
> If changing the latency fixes the problem, are we sure we are not simply (un)exposing a different problem somewhere else?

Well, obviously there is a different problem, that's why I will keep searching what it is. For X86Board, only the IDE controller is connected as a PCI Device. Before #2298 (new PCI architecture), it was directly connected to the IO bus, but now it's connected to it through the PCI bus, and so it added latency. The point of this PR is to revert that latency until we find out the real problem.

> In other words, what happens if we go back before your PCI change and artificially add a delay?

I didn't think to test, but it is a good idea. I will do it!

> Would we experience the same segfault?

Well this segfault isn't related to the latency as it was already present before the PCI changes (so without delay), but it's maybe with the real problem.

For more information, as I forgot to include it in my original message, the real problem caused by that latency, is that during boot the TTY isn't properly discovered by systemd even if it is correctly used to print boot messages. This means that when the shell is launched after the systemd boot, it will wait for a TTY to be available (at least I think so, as we loose any log message at this point). My first thought is that this is related to the interrupts as it's the only things shared between com_1 (TTY) and the IDE controller.

src/dev/pci/PciUpstream.py
```diff
@@ -51,11 +51,13 @@ class PciBus(NoncoherentXBar):
 
     config_error_port = RequestPort("Port to send config errors to")
 
-    # Set some default values bases on IOXBar
+    # FIXME: There should have some latency added to transfert packets within
```
powerjg, line 54: Rather than putting this in the definition of the *model*, let's move it to the stdlib where we define the parameters to the models. (As a side note, I really don't like that we set so many default parameters like this where we define the models. I would much rather have a nice separation of concerns between the definitions of how the models function and the parameters that we use.)

So, if we could move this to the X86Board, that would be a better place, IMO. Especially since it seems that only X86 is affected.

clemdiep:
Closing this as it was added by #2704

PR #2684, opened 2025-10-21
cpu: Enable PcCountTracker in NULL builds

powerjg:
The `AbstractBoard` (which is needed for tests) depends on this object
So, include this object even in NULL builds

### DIFF
--- a/src/cpu/probes/SConscript
+++ b/src/cpu/probes/SConscript
@@ -26,21 +26,19 @@
 
 Import("*")
 
-if env['CONF']['BUILD_ISA']:
+SimObject(
+    "PcCountTracker.py",
+    sim_objects=["PcCountTracker", "PcCountTrackerManager"],
+)
+Source("pc_count_tracker.cc")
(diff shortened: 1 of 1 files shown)


BobbyRBruce (approved):
Though I'm not entirely sure what `PcCountTracker` is meant to do when compiling to `NULL`, this simplifies the build and I don't see why it shouldn't be compiled in for all cases.

powerjg:
> Though I'm not entirely sure what `PcCountTracker` is meant to do when compiling to `NULL`, this simplifies the build and I don't see why it shouldn't be compiled in for all cases.

It doesn't do anything in NULL builds, but the AbstractBoard imports the SimObject. The TestBoard depends on AbstractBoard, so even if you're using NULL with the TestBoard, this is required in the binary. 

The underlying issue is that the typing in AbstractBoard causes us to need a lot of potentially unused SimObjects (PCIBus is another example). We should revisit this at some point.

PR #2685, opened 2025-10-21
mem-cache: register TagExtractor for SectorTag

### LINKED ISSUE #2586: RuntimeError: bad_function_call when cache compressor is used
Hi, I am running a very simple simulation on a cache hierarchy with a compressor attached to L3. However, the simulation reports `RuntimeError: bad_function_call` and the simulation ends after it. I believe this issue also exists when I use the same configuration to run full-system simulations. One example of fs simulation I run is based on [this](https://github.com/gem5/gem5/blob/v24.1.0.3/configs/example/gem5_library/x86-parsec-benchmarks.py), but its cache hierarchy is changed. Because how to use compressor is not well documented, and any help from you is much appreciated!

**Configuration script** is based on [this template](https://github.com/gem5bootcamp/2024/blob/main/materials/02-Using-gem5/05-cache-hierarchies/completed/three_level.py). Then, compressors are imported and L3 cache is configured accordingly:
```
from m5.objects import (
    BadAddr,
    Cache,
    L2XBar,
    SystemXBar,
    SubSystem,
    CompressedTags,  # newly added
    CPack,  # newly added
    BDI,  # newly added
)
```
```
class L3Cache(Cache):
    def __init__(self, size, assoc):
        super().__init__()
        self.size = size
        self.assoc = assoc
        self.tag_latency = 20
        self.data_latency = 20
        self.response_latency = 1
        self.mshrs = 20
        self.tgts_per_mshr = 12
        self.writeback_clean = False
        self.clusivity = "mostly_incl"
        self.compressor = BDI()        # newly added
        self.tags = CompressedTags()   # newly added

```

**Simulation Script**:
```
from three_level import PrivateL1PrivateL2SharedL3CacheHierarchy

from gem5.components.boards.test_board import TestBoard
from gem5.components.memory.multi_channel import DualChannelDDR4_2400
from gem5.components.processors.linear_generator import LinearGenerator

board = TestBoard(
    generator=LinearGenerator(num_cores=4, max_addr=2**22, rd_perc=75),
    cache_hierarchy=PrivateL1PrivateL2SharedL3CacheHierarchy(
        l1d_size="32KiB",
        l1d_assoc=8,
        l1i_size="32KiB",
        l1i_assoc=8,
        l2_size="256KiB",
        l2_assoc=16,
        l3_size="2MiB",
        l3_assoc=32,
    ),
    memory=DualChannelDDR4_2400(size="2GB"),
    clk_freq="3GHz",
)

sim = Simulator(board)
sim.run()
```

Error:
```

gem5 version 24.1.0.3
gem5 compiled Sep 14 2025 17:04:48
gem5 started Sep 14 2025 17:55:21
gem5 executing on vm-instunix-07, pid 3069435
command line: ./build/X86/gem5.opt test-cache.py

warn: Base 10 memory/cache size 2GB will be cast to base 2 size 2GiB.
src/mem/dram_interface.cc:690: warn: DRAM device capacity (16384 Mbytes) does not match the address range assigned (1024 Mbytes)
src/mem/dram_interface.cc:690: warn: DRAM device capacity (16384 Mbytes) does not match the address range assigned (1024 Mbytes)
RuntimeError: bad_function_call

At:
  src/python/m5/simulate.py(214): simulate
  src/python/gem5/simulate/simulator.py(720): run
  test-cache.py(33): <module>
  src/python/m5/main.py(675): main
```

studyztp:
SectorTag didn’t register TagExtractor, causing errors when calling blk->match() or sec_blk->match() since match() uses the extractor internally. 
Because SectorTag has both a sector block and a sector sub-block—and both directly or indirectly inherit TaggedEntry—each must register TagExtractor.
Related issue: https://github.com/gem5/gem5/issues/2586

### DIFF
--- a/src/mem/cache/tags/sector_tags.cc
+++ b/src/mem/cache/tags/sector_tags.cc
@@ -120,12 +120,18 @@ SectorTags::tagsInit()
             // Set its index and sector offset
             blk->setSectorOffset(k);
 
+            // Register TagExtractor for SubBlk
+            blk->registerTagExtractor(genTagExtractor(indexingPolicy));
+
             // Update block index
             ++blk_index;
         }
 
         // Link block to indexing policy
         indexingPolicy->setEntry(sec_blk, sec_blk_index);
+
+        // Register TagExtractor for SecBlk
+        sec_blk->registerTagExtractor(genTagExtractor(indexingPolicy));
     }
 }
 


src/mem/cache/tags/sector_tags.cc
```diff
@@ -120,12 +120,18 @@ SectorTags::tagsInit()
             // Set its index and sector offset
             blk->setSectorOffset(k);
 
+            // Register TagExtractor for SubBlk
+            blk->registerTagExtractor(genTagExtractor(indexingPolicy));
```
odanrc, line 124: I believe this line should not be added, because indexingPolicy refers to the sector blocks (see line 128/131). If we were to add this, we would have to have a sub-indexing policy for every sector block, I think.

studyztp: I agree with you that logically it shouldn't be added but this needs to be added for SectorSubBlock to use the `match()` mechanism to check if it's matched (https://github.com/gem5/gem5/blob/develop/src/mem/cache/tags/sector_tags.cc#L277).
Registering the TagExtractor using IndexingPolicy for the SectorSubBlock should be functionally correct because [registerTagExtractor](https://github.com/gem5/gem5/blob/develop/src/mem/cache/tags/tagged_entry.hh#L250) creates a function call object [extractTag](https://github.com/gem5/gem5/blob/develop/src/mem/cache/tags/tagged_entry.hh#L125) to trigger indexingPolicy's [extractTag(addr) function](https://github.com/gem5/gem5/blob/develop/src/mem/cache/tags/indexing_policies/base.hh#L182), which returns `(addr >> tagShift);`.
The `tagShift` should be the same for SectorBlock and SectorSubBlock since it's sector tagged. Therefore, it is functionally correct.
As far as I checked, the only place SectorSubBlock uses extractTag is in the `findBlock` function.
The SectorBlock also handles all the placement so SectorSubBlock doesn't need to have a new indexingPolicy or entry.

studyztp: I went double checked. The `extractTag` is also called in `insert()` for SectorSubBlock but it is also a functionally operation which doesn't do anything with the entry in indexingPolicy.

odanrc: I *think* it should be ok then, but I haven't used these changes, so I am not familiar with them. @giactra , would you like to chime in?

powerjg: I think the code should work whether the subblock has a tag extractor or not. As far as I can tell, the subblock tag extractor won't be accessed. @studyztp, do you find that there's a panic or some other issue if this line is removed? In that case, we should keep the line. However, it things work without this line, let's remove it.

The change on line 134 seems correct to me. I belive that is the tag extractor that is used even when accessing the subblocks (e.g., in `sector_blk.cc` it says `_sectorBlk->match(tag)`). Let me know if I'm wrong.

studyztp: > I think the code should work whether the subblock has a tag extractor or not. As far as I can tell, the subblock tag extractor won't be accessed. @studyztp, do you find that there's a panic or some other issue if this line is removed? In that case, we should keep the line. However, it things work without this line, let's remove it.
> 
> The change on line 134 seems correct to me. I belive that is the tag extractor that is used even when accessing the subblocks (e.g., in `sector_blk.cc` it says `_sectorBlk->match(tag)`). Let me know if I'm wrong.

We need to keep the line because subblock also calls ```match``` (e.g., [here](https://github.com/gem5/gem5/blob/develop/src/mem/cache/tags/sector_tags.cc#L277)). There are panics if the subblock doesn't have a Tag Extractor.

studyztp: > I think the code should work whether the subblock has a tag extractor or not. As far as I can tell, the subblock tag extractor won't be accessed. @studyztp, do you find that there's a panic or some other issue if this line is removed? In that case, we should keep the line. However, it things work without this line, let's remove it.
> 
> The change on line 134 seems correct to me. I belive that is the tag extractor that is used even when accessing the subblocks (e.g., in `sector_blk.cc` it says `_sectorBlk->match(tag)`). Let me know if I'm wrong.

You're right that TagExtractor is also used by the sectorblock so both subblock and sectorblock need to register the TagExtractor.

powerjg (approved):
Thanks for the clarifications!

studyztp:
> Thanks for the clarifications!

Thanks for the review!

PR #2689, opened 2025-10-22
sim,mem-ruby: Define a CHI-TLM port

giactra:
Instead of connecting CHI-TLM components directly to the CHI controller via pointer/param, we
define CHI-TLM ports therefore generalizing the interface and allowing objects like the generator
to be connected to other CHI-TLM capable devices


src/mem/ruby/protocol/chi/tlm/python/port.py
powerjg, line 1: This file isn't a SimObject description file. It doesn't need to be parsed to generate any new SimObjects. Furthermore, `TlmSourcePort` and `TlmSinkPort` have no corresponding C++ classes (e.g., `TlmSinkPortParams`). I am working to make all of this more consistent throughout the codebase.

Since this is pure python code and not something that needs to be parsed into C++ code, can you move it to src/python/params like was done for the signal ports?

src/mem/ruby/protocol/chi/tlm/SConscript
```diff
@@ -69,13 +69,19 @@ if env["CONF"]["BUILD_TLM"]:
         sim_objects=['TlmGenerator'],
         tags=['arm isa']
     )
+    SimObject(
+        "TlmPort.py",
+        sim_objects=[],
+        tags=['arm isa']
```
powerjg: This doesn't need to be tagged Arm. It's pure python and has no dependencies on Arm ISA.

(Actually, that's probably true of most of the code in here. But it's OK to keep the other code as is. See below on this file anyway.)

powerjg (commented):
Sorry I missed this. It was on my queue of "important PRs" but then it timed out :)

src/mem/ruby/protocol/chi/tlm/python/port.py
giactra: > Since this is pure python code and not something that needs to be parsed into C++ code, can you move it to src/python/params like was done for the signal ports?

Right, I missed reviewing this PR: https://github.com/gem5/gem5/pull/2574

I think it makes a lot of sense to remove this from the SimObject description file. On the other hand I am not so fond of moving everything within a centralized param directory.... Effectively these ports should only be defined if the AMBA CHI-TLM library is linked to gem5.

Also IMHO a port definition should stay close to the definition of the data type exchanged between source and sink.

I can surely amend the SConscript in python/m5/params to check for the TLM variable... But I'd rather explore the option of defining:

PySource('m5.params', 'chi/tlm/TlmPort.py') from here.

What's your feeling about it?

src/mem/ruby/protocol/chi/tlm/SConscript
giactra: Sure, CHI is supposed to be ISA agnostic. On the other hand the  external CHI-TLM library explicitly uses the ARM namespace (check the C++ code). I'd vote for keeping this as is for now. Happy to accept a PR which decouples this from the Arm ISA in the future

powerjg: +1

src/mem/ruby/protocol/chi/tlm/python/port.py
powerjg: Unfortunately, it totally breaks my bazel stuff. Pure python needs to be in the python directory so that it can be built as a python package. In my bazel implementation, we no longer need the gem5_py binary, and we no longer have C++ python marshaling. So, there's no way to "fake" python code being in a different place than it actually is. 

I would be all for not having everything params, but that's not how gem5 works at all right now. E.g., I would be very happy if instead you would say `from mem.ruby.protocol.chi.tlm.python.port import TlmSourcePort`, but right now what we do is say `from params import *`.

I'm going to push hard for us to move all "pure python" into the python directory so that we can make true python packages and no longer depend on the brittle and confusing marshaling. 

Some day, we're going to have to reckon with the fact that `m5.objects` is both a module with all objects and also a package with many submodules with objects. But I'm putting that off for a later time.

giactra: How would the end design look like with the optional libraries like the CHI-TLM one? Where is the python file supposed to be moved? How do we add an import guard?

More importantly, how is this gonna work with EXTRAS external gem5 libraries? I believe several users have external repositories built alongside the main gem5 repo. They probably contain pure python code as well...

I feel I'd like to see a design document to properly understand where we are going and how we can simplify gem5 but without affecting current gem5 maintainability / modularity which allows to combine building blocks from different repos

powerjg: There's a problem with gem5's design which makes these things really, really difficult: these catch all python "modules" like m5.params and m5.objects. For m5.objects, because it is so embedded throughout the codebase, I have a special workaround. For m5.params, I would really like to keep things pythonic and not spread code around in many different directories.

For the guards, it's no big deal for the `TlmSourcePort` and `TlmSinkPort` to be in every gem5 build. We don't add any new dependencies because they are pure python. If these were SimObjects, that depended on something in src/arch/arm, then I would understand the need for a guard on them. However, they don't depend on anything except for m5.params. So, we shouldn't have a guard. This is just like, say, `EthAddr` where most people will never use it.

For `EXTRAS`... I would argue that we shouldn't be adding things to any python module/package in an `EXTRAS` repo except for `m5.objects`. That said, I do not envision keeping support for `EXTRAS` as we know it with bazel. There is no need for `EXTRAS` with bazel because you could just build your own version with a dependency on whatever you need from the gem5 repo. With bazel, there is no "gem5" binary, just a bunch of dynamic (or static if you prefer) libraries that are linked when they are needed.

The alternative design, if you insist on keeping the python code in src/mem/ruby/protocol/chi/tlm/python, would be to use something like `import mem.ruby.protocol.chi.tlm.python.port`. I would discourage this, though, because nothing like this is anywhere else in gem5's codebase.

Ignoring the constraints I'm under with bazel integration, I would *strongly* prefer having python code in src/python. Spreading around pure python (i.e., python that is not interpreted by the build system to generated C++ code) makes maintainance harder, IMO.

giactra: > There's a problem with gem5's design which makes these things really, really difficult: these catch all python "modules" like m5.params and m5.objects. For m5.objects, because it is so embedded throughout the codebase, I have a special workaround. For m5.params, I would really like to keep things pythonic and not spread code around in many different directories.

I agree, but this is a bit different, no?
I am effectively not "polluting" the m5.params module and the new ports are instead defined in a smaller submodule (tlm_chi)

`    PySource("m5.tlm_chi", "python/port.py")`



> 
> For `EXTRAS`... I would argue that we shouldn't be adding things to any python module/package in an `EXTRAS` repo except for `m5.objects`.

With this I disagree (in the EXTRAS based design of course)!
One of the main use-cases for EXTRAS is to link with repositories containing confidential code. In this way we don't
have a big fork of gem5 and the maintenance burden is relatively minimal.

We do have (and we should be allowing it in general) pure python in EXTRAS. This mainly refers to private code we cannot upstream. Would a user/us be forced to move that logic into m5.objects ?


> Ignoring the constraints I'm under with bazel integration, I would _strongly_ prefer having python code in src/python. Spreading around pure python (i.e., python that is not interpreted by the build system to generated C++ code) makes maintainance harder, IMO.

If you take the Bazel variable aside, I don't think this is necessarily true. There is equal chance of things rotting either in src/python than having it somewhere else. The real differences are

1) Whether a python module gets included by the default build or whether it stays optional as it depends on an external library.
2) How frequently the module is used

Take configs/common as an example. Someone in the past might have thought that putting a collection of modules into a common directory was a good way to ensure maintainability. The outcome was: a gigantic library with lots of untested-rotting code inside. Of course I don't mean the centralization was the problem, rather than it's not a sufficient condition for ensuring maintainability.

As a counter-argument (but I am sure there are some arguments in favour of having everything centralized as well), I would say that if I place some CHI-TLM (as an example) modules within src/python, I might forget about updating them whenever there is a change in the C++ code (in src/mem/ruby/protocol/chi/tlm).

In some instances having a separate directory also provides a natural namespace and it's less likely for a user to mistake similarly named modules. I believe CHI-TLM is really a great example. I believe at this point we have 
3 different flavours of TLM based logic within gem5: the original gem5-systemc transactor (gem5_within_systemc), Gabe's systemc implementation in gem5 (systemc_within_gem5) and the AMBA CHI-TLM module I added.

How confusing it has to be if the TlmPorts get embedded in src/python/m5 with no compilation guard?

I am not trying to say the centralization approach is necessarily bad, but I'd like developers/maintainers to get a full picture before we rework the entire codebase, which leads me to my last point below

> That said, I do not envision keeping support for `EXTRAS` as we know it with bazel. There is no need for `EXTRAS` with bazel because you could just build your own version with a dependency on whatever you need from the gem5 repo. With bazel, there is no "gem5" binary, just a bunch of dynamic (or static if you prefer) libraries that are linked when they are needed.

This is a massive change that asks for more visibility to the rest of the users/maintainers, whether it is in the form of a new branch, a design document (similarly to what we required for Kconfig), a PR, or a link to your own personal bazel fork.

We have integrated some great contributions from you which were making gem5 objectively better and were therefore independent of the build system.
However I feel we are getting to a point (at least this change request is) where changes have pros and cons (as natural) and my feeling is that they are somehow linked to restrictions with the build system. It is hard for us to make a decision if we don't know what the end design will be. This is partly cause we don't have a written document and partly cause we don't know Bazel. Hopefully having some visibility over what the end structure will be, we can provide some prompt feedback to make sure everyone's use-case (or at least the most important ones) are preserved.

Personally I'd be up for having a gem5/bazel branch; it would give us some time to assess how we can make things work with EXTRAS

Tagging other maintainers here @abmerop @mattsinc @dhschall @BobbyRBruce 

To go back to the original topic, I'd push to get this merged as-is (unless there are other comments) until there is consensus over where python-only logic should be placed. If the centralization argument wins, I believe moving it would take me very little time

powerjg: > To go back to the original topic, I'd push to get this merged as-is (unless there are other comments) until there is consensus over where python-only logic should be placed. If the centralization argument wins, I believe moving it would take me very little time

I'm OK with this. 

I agree that it should be a longer conversation about how to do the migration. I need to spend some time and get things upstream-ready so we can have a productive conversation.

I also agree with a lot of the points you're making, and I want to give it the time and focus it deserves. I am out next week for Thanksgiving, but after we get the 25.1 release done, I'll put together a design doc and post what I have. (Hopefully before the Christmas holiday!)

I appreciate the technical discussion! Looking forward to coming back to it soon 🙂

powerjg (approved):

src/mem/ruby/protocol/chi/tlm/python/port.py
giactra: Thanks! Looking forward to it as well!

PR #2690, opened 2025-10-22
stdlib: Remove GPU ip_discovery.bin if exists

abmerop:
This will prevent the gem5 device specific IP discovery from being symlinked which will cause driver loading to fail and subsequently the application and simulation.

The ip_discovery.bin may or may not exist depending on what the user has installed on the disk image, so use the safe option and remove whatever is there and replace with known working gem5 provided discovery files.

### DIFF
--- a/configs/example/gpufs/mi300.py
+++ b/configs/example/gpufs/mi300.py
@@ -62,6 +62,7 @@ dd if=/root/roms/mi300.rom of=/dev/mem bs=1k seek=768 count=128
 
 # Check if exists (backwards compat with ROCm <7.0)
 if [ -e /usr/lib/firmware/amdgpu/mi300_discovery ]; then
+    rm -f /usr/lib/firmware/amdgpu/ip_discovery.bin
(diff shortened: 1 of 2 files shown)


configs/example/gpufs/mi300.py
```diff
@@ -62,6 +62,7 @@
 
 # Check if exists (backwards compat with ROCm <7.0)
 if [ -e /usr/lib/firmware/amdgpu/mi300_discovery ]; then
+    rm /usr/lib/firmware/amdgpu/ip_discovery.bin
```
powerjg: Do you need to do `rm -f` in case the file isn't present? (Just asking, I don't remember the behavior if the file doesn't exist)

mattsinc (commented):
other than @powerjg 's question, I don't have any additional ones -- but will wait for an answer before approving.

configs/example/gpufs/mi300.py
abmerop: It's not completely necessary as it won't fail as `ln` will, but I noticed it does not print a message if the file does not exist, so I have changed it to `rm -f`

powerjg (approved):

PR #2698, opened 2025-10-24
mem-ruby: Update SLICC to generate only shared files

powerjg:
After the "all protocols" change, there are two types of generated files
in SLICC. There are the normal protocol files and the "shared" protocol
files. The shared files are files that are generated *exactly the same*
across all protocols. Currently, we generate these files for each protocol
that is included in the gem5 binary. (The reason we have to is because
SLICC needs to process these files every time.) This change makes it
possible out output *only* these shared files by allowing no protocol
slicc file.

These "shared" files are the only files that should be referenced by
the generic Ruby files. An exception is the CHI protocol which has
some C++ files that directly reference CHI protocol generated files.

Note: this is not currently used in SCons.

### DIFF
--- a/src/mem/slicc/parser.py
+++ b/src/mem/slicc/parser.py
@@ -92,7 +92,8 @@ class SLICC(Grammar):
             # set all of the types parsed so far as shared
             self.decl_list.setShared()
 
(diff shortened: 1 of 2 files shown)


BobbyRBruce (approved):
Looks good to me.

PR #2700, opened 2025-10-24
misc: Allow non-serializing behaviour for MiscRegs

giactra:
We traditionally enforce System register read/writes consistency by marking writes as:

`IsNonSpeculative && IsSerializeAfter`

and marking reads as:

`IsSerializeBefore`

For example in Arm this means the instructions following the MSR (write) will block in the rename stage
until the MSR gets executed and committed, while the MRS (read) will block itself in rename until
the ROB is empty

This is somehow necessary because MiscReg writes are buffered in the execution context (the DynInst in o3)
and the architectural regfile (in the ISA object) only gets updated when the MSR gets committed

While this could make sense for particular system registers with side effects, some registers can
safely be updated speculatively. Imposing a serializing behaviour for every misc reg read/write is a big
performance penalty especially for cores with a big instruction window.
Being able to read and write without serializing behaviour means allowing the CPU to recognise
producer-consumer dependencies between system register writes/reads

This PR is adding the capability in the O3 CPU of handling non-serializing misc reg updates and it does
so by starting to track aforementioned dependencies through the scoreboard. Any ISA could use this
logic by marking their MiscRegs as non-serializing.

**Note1**: This still requires the system register write to be non-speculative as we currently don't implement rollback
of system register write in case of mispeculation

**Note2**: This PR does not implement renaming for MiscRegs. This means WAW dependencies are handled in
the rename stage by effectively tagging a non-serializing MiscReg write as serializing in case there is already a
pending producer.


src/arch/x86/X86CPU.py
```diff
@@ -80,7 +80,6 @@ class DefaultX86FUPool(FUPool):
         PredALU(),
         WritePort(),
         RdWrPort(),
-        IprPort(),
```
dhschall, line 83: I think this should be replaced with `System_Unit`

src/cpu/o3/dyn_inst.hh
```diff
@@ -1066,6 +1066,23 @@ class DynInst : public ExecContext, public RefCounted
...
+        /** If the MiscReg allows non-serializing updates, we
+         * can update the value immediately without waiting
+         * for commit, but only if the instruction is non
+         * speculative
+         */
+        if (!reg.regClass().isSerializing(reg)) {
+            assert(isNonSpeculative());
+
+            bool no_squash_from_TC = thread->noSquashFromTC;
+            thread->noSquashFromTC = true;
```
dhschall, line 1079: Why is that needed?
Will the `setMisc` function do something weird?

powerjg:
At a high-level this all sounds good to me. I don't know if I'll get a chance to review carefully in the next ~week or so, though. Let's just make sure that Linux boots with RISC-V and x86 before merging.

src/arch/arm/regs/misc.cc
```diff
@@ -4826,81 +4842,97 @@ ISA::initializeMiscRegMetadata()
       .allPrivileges().exceptUserMode().writes(0);
     InitReg(MISCREG_ID_PFR0_EL1)
       .allPrivileges().exceptUserMode().writes(0)
+      .serializing(false)
```
dhschall, line 5573: What determines an instruction that does not need to be serialized?

dhschall (commented):
Thank you @giactra !
This is a huge help for us. Overall it looks good to me and while testing it we saw a significant reduction in serialization stalls without any bug.

My main question is, in the `misc.cc` file you specify manually which instructions can be serialized or not. How do you know which are does not need to be serialized?
Is it possible that other instructions can be marked as non-serializing which are not yet marked?


### Our Analysis

@yongjiehuang, a student of mine did extensive tests with our suite of server workloads which have a lot of accesses to MISC registers.

Here are the results. 
image

Your patch indeed reduced the stalls for most benchmarks. However, there are still a few that have a lot. He analyzed which instructions caused the stalls for the two outlier benchmarks ([The patch to trace serializing instructions](https://github.com/dhschall/gem5-fdp/commit/b475d23b7719c10d994656835bec3f4ed2ad0839))

Below the results for `otmetrics` (the other benchmark looks the same)
image

Most of the stalls are still initiated by MRS/MSR instructions. Before it was mostly reads from special registers now its writes.
In particular, writes to `daif` and `spsr_el1`. So the question is how do we know if these registers can also be marked as non-serializing without trying?

Another question is why do we still see `sp_el0` in the trace? The stack pointer is now mapped to an integer register and should not need to be serialized or not?

powerjg:
I believe at least one of the CI test failures are real. It looks like `build/VEGA_X86/gem5.opt` is failing to build. (I could be wrong, the tests have been a little flaky lately.)

The failure on the fast unittests also seems real. It ended with the same error twice. Plus, it looks like it's related to the changed code: 

```
/usr/bin/ld: build/ALL/arch/arm/aapcs64.test.tfo: in function `gem5::ArmISA::MiscRegClassOps::serializing(gem5::RegId const&) const':
aapcs64.test.cc:(.text._ZNK4gem56ArmISA15MiscRegClassOps11serializingERKNS_5RegIdE[_ZNK4gem56ArmISA15MiscRegClassOps11serializingERKNS_5RegIdE]+0x12): undefined reference to `gem5::ArmISA::lookUpMiscReg'
```

giactra:
> I believe at least one of the CI test failures are real. It looks like `build/VEGA_X86/gem5.opt` is failing to build. (I could be wrong, the tests have been a little flaky lately.)
> 
> The failure on the fast unittests also seems real. It ended with the same error twice. Plus, it looks like it's related to the changed code:
> 
> ```
> /usr/bin/ld: build/ALL/arch/arm/aapcs64.test.tfo: in function `gem5::ArmISA::MiscRegClassOps::serializing(gem5::RegId const&) const':
> aapcs64.test.cc:(.text._ZNK4gem56ArmISA15MiscRegClassOps11serializingERKNS_5RegIdE[_ZNK4gem56ArmISA15MiscRegClassOps11serializingERKNS_5RegIdE]+0x12): undefined reference to `gem5::ArmISA::lookUpMiscReg'
> ```

Yes I am trying to fix that. The issue is the unit-tests try to include as little as possible. By referencing the lookUpMiscReg in the misc.hh header file I am forced to add misc.cc to the source list

Which I'd rather avoid as it will bring lots of code inside the binary.. So I am trying to restructure code so that we don't end up including misc.cc

powerjg:
> Which I'd rather avoid as it will bring lots of code inside the binary.. So I am trying to restructure code so that we don't end up including misc.cc

Yes, please! As you know, I've been on the crusade recently as well 😄

giactra:
> Thank you @giactra ! This is a huge help for us. Overall it looks good to me and while testing it we saw a significant reduction in serialization stalls without any bug.
> 
> My main question is, in the `misc.cc` file you specify manually which instructions can be serialized or not. How do you know which are does not need to be serialized? Is it possible that other instructions can be marked as non-serializing which are not yet marked?

100%

It is not specified in the PR message per se but It is in one of the commits: the current labelling of system register is not exhaustive:
```


By default a MiscReg is marked as serializing.
 
We expect to slowly transition most of the sysregs into the
non-serializing subset

```

There are many other register which could be marked as non-serializing. For instance, the CONTEXTIDR register.
In general every register with no side-effect can be easily transitioned into the non-serializing bunch. For registers with side effects, it is a bit more complex.

A register with side effects can only be transitioned if we are able to establish dependencies (in the o3) between the register write and any instruction which depends on its value

giactra:
> Another question is why do we still see `sp_el0` in the trace? The stack pointer is now mapped to an integer register and should not need to be serialized or not?

Thanks for the report. I don't know for sure, but looking at your patch, I believe the issue might not be the MRS SP_EL0 per se, but rather the preceding instruction.

In fact SerializeAfter does nothing more than labelling the next instruction in PO as SerializeBefore.

So if you have

MSR SCTLR_EL1 x0 -> SerializeAfter
MRS x1, SP_EL0 

The SP read will be tagged as SerializeBefore by the preceding instruction. So in your evaluation you need to distinguish between static and temporary SerializeBefore. If you find some time to debug your experiments, let me know your finding

dhschall:
> In fact SerializeAfter does nothing more than labelling the next instruction in PO as SerializeBefore.

Thanks! that might be indeed correct. I'll ask the student if he can distinguish between instructions that are marked by previous instructions

src/cpu/o3/rename.cc
```diff
@@ -1155,6 +1170,38 @@ Rename::renameDestRegs(const DynInstPtr &inst, ThreadID tid)
     }
 }
 
+void
+Rename::setSerializeBefore(DynInstPtr &inst, ThreadID tid)
```
dhschall: Tbh. I found this a bit misleading. The function suggests you set serialization regardless. However, what the function actually does it check for WaW. If one exists, resolve it with serialization.
I think `checkMiscWaWConflict` or something similar would be more appropriate

dhschall:
Just to be clear what do you mean by _"side-effect"_
For example we are now trying `SPSR_EL1` and `ELR_EL1`.
SPSR_EL1: _Holds the saved process state when an exception is taken to EL1._
Would you call it side-effect free?

src/cpu/o3/rename.cc
giactra: I agree, I thought about it. I will rename accordingly

giactra:
> Just to be clear what do you mean by _"side-effect"_ For example we are now trying `SPSR_EL1` and `ELR_EL1`. SPSR_EL1: _Holds the saved process state when an exception is taken to EL1._ Would you call it side-effect free?

One **initial** definition of side effects (in our case, because for the Arm architecture is a bit different) could be:

> A register whose value will be consumed by an instruction different from an explicit read

For instance a MSR TPIDR_EL0 does not have side effects because apart for a MRS TPIDR_EL0 no other instruction will try to read that value. So as long as you keep track of the dependency via the scoreboard, the scheduler/O3 engine will ensure the RAW dependency is preserved.

The SPSR_EL1 value is actually consumed in several circumstances, namely when you ERET and when you throw an exception.

However, considering those are serializing-non speculative events, I suppose you should be able to mark the SPSR as non serializing. Unless I am missing something.

src/proto/inst.proto
```diff
@@ -95,7 +95,6 @@ message Inst {
     SIMDFloatSqrt = 29;
     MemRead = 30;
     MemWrite = 31;
-    IprAccess = 32;
```
powerjg, line 98: To ensure we don't accidentally reuse the field number 32 in the future, it's best practice in Protobuf to explicitly reserve it after deletion. This prevents potential bugs with older clients that might still be sending or expecting the old IprAccess field.

Could we please add the following reserved keywords?

```Protocol Buffers
    reserved 32;
    reserved "IprAccess";
```

This will safeguard our protocol's backward compatibility. Thanks!

powerjg (commented):
lgtm. Please take a look at David's suggestions/questions as well. One small thing on the protobuf change.

src/proto/inst.proto
giactra: Ack

src/arch/x86/X86CPU.py
giactra: Done

src/cpu/o3/dyn_inst.hh
giactra: Copy pasted from the updateMiscRegs logic:

```
1092         // @todo: Pretty convoluted way to avoid squashing from happening when
1093         // using the TC during an instruction's execution (specifically for
1094         // instructions that have side-effects that use the TC).  Fix this.
1095         // See cpu/o3/dyn_inst_impl.hh.
```

IIRC there are cases where calling

thread->setMiscReg would cause a squash in the O3 model due to the following:

https://github.com/gem5/gem5/blob/stable/src/cpu/o3/thread_context.cc#L207C16-L207C26

The idea is that if you are changing anything through the TC interface, it means you are forcing a change in the execution context and therefore you need to squash the pipeline.

powerjg (approved):

PR #2703, opened 2025-10-27
 arch-riscv: Fix RISC-V `Decoder::reset()` not calling base class reset

### LINKED ISSUE #2681: RISC-V decoder `reset()` does not call `InstDecoder::reset()` (unlike ARM/x86)
### Description

While studying the RISC-V front-end behavior, I noticed that
`src/arch/riscv/decoder.cc` defines its own `Decoder::reset()` implementation but **does not call** the base `InstDecoder::reset()`.

In contrast, both the ARM and x86 decoders explicitly invoke the base reset, which initializes two internal state variables:

```cpp
InstDecoder::reset() {
    outOfBytes = true;
    instDone = false;
}
```

Because the RISC-V implementation omits this call, `outOfBytes` and `instDone` may retain stale values across resets.

---

### Context / Repro

I ran into this while working on a custom out-of-order core that reuses the RISC-V decode logic.
After a squash, the decoder did not fully reset — it occasionally resumed mid-instruction, depending on what was in its internal buffer.

Unfortunately, I cannot share the proprietary core model, but the problem disappears if I add:

```cpp
InstDecoder::reset();
```

at the beginning of `RiscvISA::Decoder::reset()`.

All the stock gem5 CPU models seem unaffected, so it’s unclear if leaving these fields untouched was intentional (perhaps RISC-V doesn’t rely on `instDone`/`outOfBytes`), or if this is simply an oversight.

---

### Questions

1. Is there a reason `RiscvISA::Decoder::reset()` doesn’t call its base implementation?
2. Would adding `InstDecoder::reset();` there be a safe and consistent change across ISAs?

JerryAZR:
**Description:**
This patch adds a call to the base class `InstDecoder::reset()` inside `Decoder::reset()` in `src/arch/riscv/decoder.cc`.

While this change does not affect existing RISC-V CPU models in gem5, it ensures consistency across ISA decoders and correct behavior of functions such as `instReady()` and `needMoreBytes()` after a reset.

This also prevents potential issues when developing custom RISC-V CPU models that rely on decoder reset semantics consistent with other ISAs.

**Files changed:**

* `src/arch/riscv/decoder.cc`

**Related issue:**
Fixes discussion in https://github.com/gem5/gem5/issues/2681)

### DIFF
--- a/src/arch/riscv/decoder.cc
+++ b/src/arch/riscv/decoder.cc
@@ -51,6 +51,7 @@ Decoder::Decoder(const RiscvDecoderParams &p) : InstDecoder(p, &machInst)
 
 void Decoder::reset()
 {
+    InstDecoder::reset();
     aligned = true;
     mid = false;
     machInst = 0;


powerjg:
@rogerchang23424 Does this look OK to you? It looks good to me, but I think you've been working with this code most recently.

rogerchang23424 (approved):
LGTM

powerjg (approved):

PR #2707, opened 2025-10-27
stdlib: Clean up stdlib imports in CHI

powerjg:
- NULL should come from m5.params
- Use relative paths for gem5 library imports


Harshil2107 (approved):
LGTM

PR #2708, opened 2025-10-28
resources: make workloads work with set_se_multi_binary_workload

Harshil2107:
The  `set_se_multi_binary_workload ` function expects parameter which accepts a list of resources. This change updates the `resources` field in workload json to work with that requiremet.

- With this change:
  - Each Key in the resources dictionary represents a parameter name in the function that is called by the workload (No change here)
  - Each value in the resources dictionary can be a single resource or a list of resources that the above mentioned parameter can take
  
 To test this change with `set_se_multi_binary_workload`, you can create a JSON file with the following test resource:
 
 ```json
 [
    {
    "category": "workload",
    "id": "x86-multi-hello",
    "description": "",
    "function": "set_se_multi_binary_workload",
    "resources": {
        "binaries": [
            {
                "id": "x86-hello64-static",
                "resource_version": "1.0.0"
            },
            {
                "id": "x86-hello64-static",
                "resource_version": "1.0.0"
            }
        ]
    },
    "additional_params": {
        "arguments": []
    },
    "architecture": "X86",
    "size": 0,
    "tags": [],
    "license": "",
    "author": [
    ],
    "source_url": "",
    "resource_version": "1.0.0",
    "gem5_versions": [
        "25.0"
    ],
    "example_usage": ""
}
]
 ```
 
 and call with workload with the following test config
 
 ```python
from gem5.components.cachehierarchies.classic.no_cache import NoCache
from gem5.isas import ISA

requires(isa_required=ISA.X86)

cache_hierarchy = NoCache()

memory = SingleChannelDDR3_1600(size="32MiB")

processor = SimpleProcessor(cpu_type=CPUTypes.TIMING, isa=ISA.X86, num_cores=2)

board = SimpleBoard(
    clk_freq="3GHz",
    processor=processor,
    memory=memory,
)

board.set_workload(
    obtain_resource("x86-multi-hello", resource_version="1.0.0")
)

simulator.run()
```

### DIFF
--- a/src/python/gem5/resources/resource.py
+++ b/src/python/gem5/resources/resource.py
@@ -1270,19 +1270,34 @@ def _get_workload(
     if "parameters" not in workload:
         workload["parameters"] = {}
 
-    for resource in workload["resources"].values():
-        db_query.append(
-            ClientQuery(
-                resource_id=resource["id"],
-                resource_version=resource["resource_version"],
-                gem5_version=gem5_version,
+    for resource_param in workload["resources"].values():
+        # Each Key in the resources dictionary represents a parameter name in the function that is called by the workload
+        # Each value in the resources dictionary can be a single resource or a list of resources that the parameter can take
+
+        if isinstance(resource_param, list):
+            for res in resource_param:
+                db_query.append(
+                    ClientQuery(
+                        resource_id=res["id"],
+                        resource_version=res["resource_version"],
+                        gem5_version=gem5_version,
+                    )
+                )
+        else:
+            db_query.append(
+                ClientQuery(
+                    resource_id=resource_param["id"],
+                    resource_version=resource_param["resource_version"],
+                    gem5_version=gem5_version,
+                )
             )
-        )
     # Fetching resources as a list of dicts
     resource_details_list = get_multiple_resource_json_obj(db_query, clients)
 
-    # Creating the resource objects for each resource
-    for param_name, param_resource in workload["resources"].items():
+    def create_resource_object(
+        param_resource: Dict[str, str],
+        resource_details_list: List[Dict[str, str]],
+    ) -> AbstractResource:
         resource_match = None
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/python/gem5/resources/resource.py [hunk 1]
```diff
@@ -1270,19 +1270,34 @@ def _get_workload(
...
+        if isinstance(resource_param, list):
+            for res in resource_param:
+                db_query.append(
+                    ClientQuery(
+                        resource_id=res["id"],
+                        resource_version=res["resource_version"],
+                        gem5_version=gem5_version,
+                    )
+                )
+        else:
```
powerjg: What type do you expect if it's not a list? It may be good to put a check here and raise an exception if it's not. This should also have unittests (ask AI to write them).

src/python/gem5/resources/resource.py
```diff
@@ -1270,19 +1270,34 @@ def _get_workload(
...
+                    gem5_version=gem5_version,
+                )
             )
-        )
     # Fetching resources as a list of dicts
     resource_details_list = get_multiple_resource_json_obj(db_query, clients)
 
-    # Creating the resource objects for each resource
-    for param_name, param_resource in workload["resources"].items():
+    def create_resource_object(
```
powerjg, line 1303: I don't like that this is a function defined in the function. I appreciate that it's only used here, so you don't want to make it more public (and I'm all for that). But it was pretty confusing when reviewing that there was a return and a function here.

The problem this is solving is that resource.py has gotten completely out of hand. A 1600 line file is way too big. This should probably be a package with a bunch of submodules.

Actionable item: Nothing right now. This is OK for this PR. However, I would suggest asking AI to review this file and ask for suggestions on how to make it more maintainable. I hope that it gives 2 suggestions: (1) break into submodules and (2) write unittests. It should be able to do both of these will a little guidance.

powerjg (approved):

PR #2714, opened 2025-10-29
arch-riscv: add initrd option to FsWorkload

Nik-Sch:
Similarly to the ArmFsWorkload the riscv fs workload also gets an initrd.

### DIFF
--- a/src/arch/riscv/linux/fs_workload.cc
+++ b/src/arch/riscv/linux/fs_workload.cc
@@ -184,12 +184,36 @@ BootloaderKernelWorkload::loadKernel()
     }
 }
 
+void
+BootloaderKernelWorkload::loadInitrd()
+{
+    if (params().initrd_filename != "") {
+        inform("Loading initrd file: %s at address %#x\n",
+               params().initrd_filename, params().initrd_addr);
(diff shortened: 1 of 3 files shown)


src/arch/riscv/linux/fs_workload.cc
```diff
@@ -184,13 +184,37 @@ BootloaderKernelWorkload::loadKernel()
...
 
+void
+BootloaderKernelWorkload::loadInitrd()
+{
+    if (params().initrd_filename != "") {
+        inform("Loading initrd file: %s at address %#x\n",
+               params().initrd_filename, params().initrd_addr);
+
+        loader::ImageFileDataPtr initrd_file_data(
+            new loader::ImageFileData(params().initrd_filename));
```
rogerchang23424: Initrd should not be decompressed https://github.com/gem5/gem5/pull/2569

```suggestion
        loader::ImageFileDataPtr initrd_file_data(
            new loader::ImageFileData(params().initrd_filename, false));
```

rogerchang23424 (approved):

erin-le (approved):

PR #2716, opened 2025-10-29
python: Match walker caches to available MMU ports

BobbyRBruce:
Some CPUs expose zero or one MMU walker port, but the classic cache hierarchies always instantiated both ITLB and DTLB walk caches. When a CPU lacked an instruction walker, the ITLB cache was left dangling and gem5 aborted with “Cache ports on board.cache_hierarchy.iptw_caches are not connected” which was causing our Daily tests to fail (specifically, "configs/example/gem5_library/arm-hello-dramsys.py" and other DRAMSys tests).

The fix here is to only create walk caches when the CPU reports walker ports, and reuse the D-side cache when there is a single shared walker. The prebuilt RISC-V hierarchy and the private L1/L2 walk-cache setup follow the same pattern.

~~The Arm table walker’s cached port pointer is marked as `[[ maybe_unused]]` to avoid a new compiler warning when no I-side walker exists.~~

### DIFF
--- a/src/python/gem5/components/cachehierarchies/classic/private_l1_cache_hierarchy.py
+++ b/src/python/gem5/components/cachehierarchies/classic/private_l1_cache_hierarchy.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2021 The Regents of the University of California
+# Copyright (c) 2021-2025 The Regents of the University of California
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
--- a/src/python/gem5/components/cachehierarchies/classic/private_l1_cache_hierarchy.py
+++ b/src/python/gem5/components/cachehierarchies/classic/private_l1_cache_hierarchy.py
@@ -28,6 +28,7 @@ from typing import Optional
 
 from m5.objects import (
     BadAddr,
+    BaseCPU,
     BaseXBar,
     Cache,
     SystemXBar,
@@ -41,12 +42,13 @@ from ..abstract_cache_hierarchy import AbstractCacheHierarchy
 from .abstract_classic_cache_hierarchy import AbstractClassicCacheHierarchy
 from .caches.l1dcache import L1DCache
 from .caches.l1icache import L1ICache
-from .caches.mmu_cache import MMUCache
 
 
 class PrivateL1CacheHierarchy(AbstractClassicCacheHierarchy):
     """
     A cache setup where each core has a private L1 data and instruction Cache.
+    Table walker ports connect directly to the membus by default so walk caches
+    are only instantiated via subclasses such as PrivateL1WalkCacheHierarchy.
     """
 
     def _get_default_membus(self) -> SystemXBar:
@@ -108,8 +110,6 @@ class PrivateL1CacheHierarchy(AbstractClassicCacheHierarchy):
             L1DCache(size=self._l1d_size)
             for i in range(board.get_processor().get_num_cores())
         ]
-        iptw_caches = []
-        dptw_caches = []
 
         if board.has_coherent_io():
             self._setup_io_cache(board)
@@ -120,37 +120,7 @@ class PrivateL1CacheHierarchy(AbstractClassicCacheHierarchy):
 
             self.l1icaches[i].mem_side = self.membus.cpu_side_ports
             self.l1dcaches[i].mem_side = self.membus.cpu_side_ports
-
-            walker_ports = cpu.get_mmu().walkerPorts()
-            if len(walker_ports) > 2:
-                raise RuntimeError(
-                    "Unexpected number of walker ports "
-                    f"from CPU {i}: {len(walker_ports)}.\n"
-                    "Expected 0, 1, or 2"
-                )
-            if len(walker_ports) == 0:
-                continue
-
-            dptw_cache = MMUCache(size="8KiB")
-            dptw_cache.mem_side = self.membus.cpu_side_ports
-
-            if len(walker_ports) == 2:
-                iptw_cache = MMUCache(size="8KiB")
-                iptw_cache.mem_side = self.membus.cpu_side_ports
-                cpu.connect_walker_ports(
-                    iptw_cache.cpu_side, dptw_cache.cpu_side
-                )
-                iptw_caches.append(iptw_cache)
-            else:
-                assert len(walker_ports) == 1, (
-                    f"This branch expects 1 walker_port, got "
-                    f"{len(walker_ports)}."
-                )
-                cpu.connect_walker_ports(
-                    dptw_cache.cpu_side, dptw_cache.cpu_side
-                )
-
-            dptw_caches.append(dptw_cache)
+            self._connect_table_walker(i, cpu)
 
             if board.get_processor().get_isa() == ISA.X86:
                 int_req_port = self.membus.mem_side_ports
@@ -159,9 +129,22 @@ class PrivateL1CacheHierarchy(AbstractClassicCacheHierarchy):
             else:
                 cpu.connect_interrupt()
 
-        if iptw_caches:
-            self.iptw_caches = iptw_caches
-        self.dptw_caches = dptw_caches
+    def _connect_table_walker(self, cpu_id: int, cpu: BaseCPU) -> None:
+        walker_ports = cpu.get_mmu().walkerPorts()
+        if len(walker_ports) > 2:
+            raise RuntimeError(
+                "Unexpected number of walker ports "
+                f"from CPU {cpu_id}: {len(walker_ports)}.\n"
+                "Expected 0, 1, or 2"
+            )
+
+        if len(walker_ports) == 0:
+            return
+
+        cpu.connect_walker_ports(
+            self.membus.cpu_side_ports,
+            self.membus.cpu_side_ports,
+        )
 
     def _setup_io_cache(self, board: AbstractBoard) -> None:
         """Create a cache for coherent I/O connections"""
--- a/src/python/gem5/components/cachehierarchies/classic/private_l1_cache_hierarchy.py
+++ b/src/python/gem5/components/cachehierarchies/classic/private_l1_cache_hierarchy.py
@@ -122,13 +122,19 @@ class PrivateL1CacheHierarchy(AbstractClassicCacheHierarchy):
             self.l1dcaches[i].mem_side = self.membus.cpu_side_ports
 
             walker_ports = cpu.get_mmu().walkerPorts()
+            if len(walker_ports) > 2:
+                raise RuntimeError(
+                    "Unexpected number of walker ports "
+                    f"from CPU {i}: {len(walker_ports)}.\n"
+                    "Expected 0, 1, or 2"
+                )
             if len(walker_ports) == 0:
                 continue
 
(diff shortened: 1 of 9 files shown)


src/arch/arm/table_walker.hh
```diff
@@ -1196,7 +1196,7 @@ class WalkUnit : public ClockedObject
         BaseMMU::Mode mode;
         MMU::ArmTranslationType tranType;
 
-        TableWalker::Port *port;
+        [[maybe_unused]] TableWalker::Port *port;
```
giactra: What's the error/warning you are getting here?

I am surprised you would mark this as being maybe unused considering it is definitely being used

BobbyRBruce: I was mistaken in my original reasoning for this. I observed the code didn't compile on my Mac without this `[[maybe_unused]]` so assumed it was to do with the other changes I made in this PR. It turns out it's a separate issue independent of this change. I've therefore removed it and will open up another PR.

You are right that it's definitely used but my Mac complains regardless. It’s using "Apple clang version 17.0.0".

BobbyRBruce: Actually, i think this unused warning/error is correct. I've create #2718 which removes the pointer and explains why it's unused.

giactra:
Hi Bobby, I haven't explored the change in detail, but my first thought is:

It seems to me the issue comes from Arm case having a single table walker port and its conflict with the two table walker caches.

Now, Arm configs shouldn't be using table walker caches, and that's the reason why I defined an explicit walk cache hierarchy some time ago in stdlib:

https://github.com/gem5/gem5/blob/stable/src/python/gem5/components/cachehierarchies/classic/private_l1_private_l2_walk_cache_hierarchy.py

The reason for Arm not using the TBW cache is because Arm already caches partial translations in the L2 TLB. So I would say any Arm test should not be using configs with a MMU cache. When I added the private_l1_private_l2_walk_cache_hierarchy module, I only covered priv L1 and priv L2. I believe we should be applying the same logic for the other configs (priv L1 shared L2) as well.

This will probably make things cleaner on the python front and improve the transparency for a user (which is not getting an hidden cache somewhere in their config).

Let me know your thought

src/python/gem5/components/cachehierarchies/classic/private_l1_cache_hierarchy.py
```diff
@@ -129,12 +121,26 @@ def incorporate_cache(self, board: AbstractBoard) -> None:
...
+            if len(walker_ports) == 0:
+                continue
+
+            dptw_cache = MMUCache(size="8KiB")
+            dptw_cache.mem_side = self.membus.cpu_side_ports
 
-            cpu.connect_walker_ports(
-                self.iptw_caches[i].cpu_side, self.dptw_caches[i].cpu_side
-            )
+            if len(walker_ports) > 1:
```
powerjg:
```suggestion
            if len(walker_ports) == 2:
```

src/python/gem5/components/cachehierarchies/classic/private_l1_private_l2_walk_cache_hierarchy.py
```diff
@@ -57,25 +57,28 @@ def __init__(self, *args, **kwargs) -> None:
...
+        self._tmp_dptw_caches = []
         super().incorporate_cache(board)
-
-        for i, cpu in enumerate(board.get_processor().get_cores()):
-            self.iptw_caches[i].mem_side = self.l2buses[i].cpu_side_ports
-            self.dptw_caches[i].mem_side = self.l2buses[i].cpu_side_ports
+        self.iptw_caches = self._tmp_iptw_caches
+        self.dptw_caches = self._tmp_dptw_caches
+        del self._tmp_iptw_caches
+        del self._tmp_dptw_caches
```
powerjg: Does this work? deleting simobjects is... problematic.

src/python/gem5/components/cachehierarchies/classic/private_l1_private_l2_walk_cache_hierarchy.py
```diff
@@ -57,25 +57,28 @@ def __init__(self, *args, **kwargs) -> None:
...
-            self.dptw_caches[cpu_id].cpu_side,
-        )
+        walker_ports = cpu.get_mmu().walkerPorts()
+        if len(walker_ports) == 0:
+            return
+
+        dptw_cache = MMUCache(size="8KiB")
+        dptw_cache.mem_side = self.l2buses[cpu_id].cpu_side_ports
+
+        if len(walker_ports) > 1:
```
powerjg:
```suggestion
        if len(walker_ports) == 2:
```

I forgot to mention above, but you should probably panic if it's not 0, 1, or 2.

src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy.py
```diff
@@ -156,12 +148,26 @@ def incorporate_cache(self, board: AbstractBoard) -> None:
...
-            self.iptw_caches[i].mem_side = self.l2bus.cpu_side_ports
-            self.dptw_caches[i].mem_side = self.l2bus.cpu_side_ports
+            walker_ports = cpu.get_mmu().walkerPorts()
+            if len(walker_ports) == 0:
+                continue
+
+            dptw_cache = MMUCache(size="8KiB", writeback_clean=False)
+            dptw_cache.mem_side = self.l2bus.cpu_side_ports
+
+            if len(walker_ports) > 1:
```
powerjg: Ditto

powerjg (commented):
This is an OK fix for now, but generally we need to improve the API to the MMU. The MMU should probably say if ports are instruction or data side and it should also say if a cache can be used.

These caches are not very realistic. TBH, I don't remember why we have them. I believe it's because the classic caches *require* a cache at the front of any CPU-side requestor port. Or, it could have been something that I decided in my PhD was a more reasonable design than no cache at all. (I could have been wrong.)

giactra:
Hi @powerjg . Please review my comment above. It is my opinion the cleanest solution consists on defining a walker_cache variant (similarly to [private_l1_private_l2_walk_cache_hierarchy.py](https://github.com/gem5/gem5/pull/2716/files/e2abbbcda9216d9ed95b3c28729e5d792ffb3c17#diff-5d3a35ca6bd46793721f4910ed13855a86cc1fca24bfa29e66928bb069bdc8c7)) to simplify things and make the cache hierarchy more explicit (at the moment users are not fully aware they get a MMU cache in front of the table waker).

> These caches are not very realistic. TBH, I don't remember why we have them. I believe it's because the classic caches require a cache at the front of any CPU-side requestor port.

I don't thinkg that is the case, considering we can safely attach an Arm table walker to the L2 bus.

> Or, it could have been something that I decided in my PhD was a more reasonable design than no cache at all. (I could have been wrong.

So when I looked into MMU caches some time ago, I found this paper (https://dl.acm.org/doi/10.1145/1816038.1815970) reporting how some x86 based system are using a MMU cache to store walk descriptors. I don't know whether this is still relevant considering how old the paper is (2010).

In Arm we don't store a descriptor in a cacheline, instead we devolve part of a TLB to store partial translations, meaning it does not store a full VA-PA translation, but it stores VA-PA of an intermediate descriptor. So we don't need a walk cache powered hierarchy at all.

This is why I would avoid making any modification to the stdlib config to accomodate for the Arm case. I believe any Arm setup should be using a cacheless config in the first place. This will simplify code IMHO

BobbyRBruce:
I think I've addressed all the suggestions/comments raised here. This fixes the failing Daily Tests so would like to get this merged soon.

Harshil2107 (dismissed):
Not an expert but the change LGTM

src/python/gem5/components/cachehierarchies/classic/private_l1_cache_hierarchy.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2021 The Regents of the University of California
+# Copyright (c) 2021-2025 The Regents of the University of California
```
giactra: Shouldn't this be 2021, 2025?

src/python/gem5/prebuilt/demo/x86_demo_board.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2021 The Regents of the University of California
+# Copyright (c) 2021-2025 The Regents of the University of California
```
giactra: Ditto

src/python/gem5/components/cachehierarchies/classic/private_l1_cache_hierarchy.py
```diff
@@ -143,6 +129,23 @@ def incorporate_cache(self, board: AbstractBoard) -> None:
...
+        walker_ports = cpu.get_mmu().walkerPorts() if cpu.has_mmu() else []
+        if len(walker_ports) > 2:
+            raise RuntimeError(
+                "Unexpected number of walker ports "
+                f"from CPU {cpu_id}: {len(walker_ports)}.\n"
+                "Expected 0, 1, or 2"
+            )
+
+        if len(walker_ports) == 0:
+            return
```
giactra, line 142: Can't we just move this in the connect_walker_ports?

A generator CPU will return immediately without any binding

src/python/gem5/components/cachehierarchies/classic/private_l1_private_l2_walk_cache_hierarchy.py
```diff
@@ -57,25 +58,36 @@ def __init__(self, *args, **kwargs) -> None:
...
+            iptw_cache.mem_side = self.l2buses[cpu_id].cpu_side_ports
+            cpu.connect_walker_ports(iptw_cache.cpu_side, dptw_cache.cpu_side)
+            self._tmp_iptw_caches.append(iptw_cache)
+        else:
+            assert (
+                len(walker_ports) == 1
+            ), f"This branch expects 1 walker_port, got {len(walker_ports)}."
+            cpu.connect_walker_ports(dptw_cache.cpu_side, dptw_cache.cpu_side)
+
+        self._tmp_dptw_caches.append(dptw_cache)
```
giactra, line 93: This does the job, so if it fixes the daily test we should merge this.
The table walker interface should probably be reworked in the future though to make code a bit cleaner

giactra (commented):
Hi @BobbyRBruce few minor comments and then I think it is ready to be merged

src/python/gem5/components/cachehierarchies/classic/private_l1_cache_hierarchy.py
BobbyRBruce: Even though right now all the cores have `connect_walker_ports` implemented, I'd argue in the case there are no walker ports, the `connect_walker_ports` doesn't need to be implemented and therefore an exception would be raised if called if so. The intent of the function is to connect the input ports to the CPU's walker ports. If it is called i'd assume that is what is happening. So, if these ports don't exist I think that's an exception case as it's not doable. This probably should be the case for the generator,  but i'll leave that as-is for now.

At worst this return is redundant but does cover future cases where a new core may not have walker ports.

PR #2717, opened 2025-10-29
mem-ruby: Fix CHI TBE allocation

BobbyRBruce:
A bug was introduced in 0f476aabf3 (PR #2706) and observed when building VEGA_X86 in a ghcr.io/gem5/gcn-gpu:latest container. It causes incomplete-type errors for `MiscNode_TBE`. It fails because the container still uses GCC 10, which instantiates `TBETable<MiscNode_TBE>` while the type is incomplete. As per-standard, this is ill-formed. GCC 11+ (and Clang) happen to instantiate the template later, after the full definition is visble, so the issue still exists even if newer compilers don't flag it.

To preserve the dependency break introduced in 0f476aabf3, this fix stores entries in `TBETable` via unique_ptr and specializes the factory in "MN_TBETable.cc" so the full `MiscNode_TBE` definition is only needed in the "MN_TBETABLE.cc" file

### DIFF
--- a/src/mem/ruby/structures/MN_TBETable.cc
+++ b/src/mem/ruby/structures/MN_TBETable.cc
@@ -49,6 +49,18 @@ namespace gem5
 namespace ruby
 {
 
+CHI::MiscNode_TBE *
+EntryFactory<CHI::MiscNode_TBE>::create(int block_size)
+{
+    return new CHI::MiscNode_TBE(block_size);
+}
+
+void
+EntryFactory<CHI::MiscNode_TBE>::destroy(CHI::MiscNode_TBE *entry)
+{
+    delete entry;
+}
+
 namespace CHI
 {
 
@@ -67,7 +79,7 @@ MN_TBETable::chooseNewDistributor()
     bool has_waiting_sync = false;
     int waiting_count = 0;
     for (auto& keyValuePair : m_map) {
-        MiscNode_TBE& tbe = keyValuePair.second;
+        MiscNode_TBE &tbe = *keyValuePair.second;
 
         switch (tbe.getstate()) {
             case MiscNode_State_DvmSync_Distributing:
--- a/src/mem/ruby/structures/MN_TBETable.hh
+++ b/src/mem/ruby/structures/MN_TBETable.hh
@@ -51,9 +51,8 @@ namespace CHI
 
 class MiscNode_TBE;
 
-// Custom class only used for the CHI protocol Misc Node
-// Includes the definition of the MiscNode_TBE, because it
-// includes functions that rely on fields in the structure
+// Custom class only used for the CHI protocol Misc Node.
+// Implementation relies on MiscNode_TBE fields defined in the protocol.
 class MN_TBETable : public TBETable<MiscNode_TBE>
 {
   public:
--- a/src/mem/ruby/structures/TBETable.hh
+++ b/src/mem/ruby/structures/TBETable.hh
@@ -42,6 +42,7 @@
 #define __MEM_RUBY_STRUCTURES_TBETABLE_HH__
 
 #include <iostream>
+#include <memory>
 #include <unordered_map>
 
 #include "mem/ruby/common/Address.hh"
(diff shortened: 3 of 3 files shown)


abmerop:
We should probably discuss in one of the dev meetings what to do with the GPU SE mode docker if it continues to cause issues (I guess it was partially help here).  I don't believe there is any on going work to get GPU SE mode up to date and hence it is stuck on an old software stack, old OS, and old compiler.

giactra (commented):
Thanks Bobby for fixing this. I believe this is a reasonable solution for the current failure, but IMHO we are probably making things

1) more complicated from a SW perspective
2) slower because key data structures like the TBE are now doing heap allocation/deallocation

1 = I lost track of what you can do with metaprogramming and forward declaration, but as a noob I am not confident your template specialization with a forward declare MiscNode_TBE will compile seamlessly in the future. I wonder if we are not solving this now to get a new failure later on
 
2 = I believe this is a huge price to fix something which is could be achieved with a bit of restructuring. As painful as it is to @powerjg I vote for reverting his PR.

There is no real cyclic dependency here among headers, if you consider generated and non-generated files at the same level:

CHIController.hh -> includes TBE Table.hh
CHIController.hh ->includes MiscNode_TBE entry.hh
TBE Table.hh -> includes MiscNode_TBE entry.hh

There is a cyclic dependency only if you consider SLICC generated files as a block and not SLICC generated files as another block. I agree it is not ideal, but it is not as bad as to justify the cost we are paying in my opinion

Please let me your thought

powerjg:
I thought I tried that, and it was much worse than just the dependency you listed. I'll try again and see what I can do.

@BobbyRBruce is my fix here https://github.com/gem5/gem5/pull/2623 going to cause the same problem?

powerjg:
The problem with @giactra's solution is that if I declare each generated header as its own `cc_library`, then I have to declare **separately** each header's dependences. This is completely impossible to do in a "generic" way with the SLICC generated headers.

Also, it's not as simple as the dependencies you listed. Here are all of the headers that are referenced from `MiscNode_TBE.hh`

```
#include "mem/ruby/protocol/CHI/CHIRequestType.hh"
#include "mem/ruby/protocol/MachineID.hh"
#include "mem/ruby/protocol/CHI/MiscNode_State.hh"
#include "mem/ruby/protocol/NetDest.hh"
#include "mem/ruby/protocol/NetDest.hh"
#include "mem/ruby/protocol/NetDest.hh"
#include "mem/ruby/protocol/TriggerQueue.hh"
#include "mem/ruby/protocol/CHI/MiscNode_Event.hh"
#include "mem/ruby/protocol/CHI/MiscNode_State.hh"
#include "mem/ruby/protocol/ExpectedMap.hh"
#include "mem/ruby/protocol/ExpectedMap.hh"
#include "mem/ruby/protocol/CHI/CHIResponseType.hh"
#include "mem/ruby/protocol/CHI/CHIDataType.hh"
```

So, I would need to declare all of these files as dependencies of `MiscNode_TBE.hh`. Then, I would also need to create new separate header `cc_libraries` for each of these headers (and they can't depend on `CHIController.hh` because then we would have a circular dependence.

If this was all just C++ code, I would write a separate header `cc_library` for each of these (this is what I've been doing). However, since it's all generated, I would essentially have to hard code all of the details of the generated files. Honestly, this was my initial plan, but I've been able to make the code generation work in bazel 🎉 , so I'd like to keep it that way.

I want to come back to this statement:
> There is no real cyclic dependency here among headers, if you consider generated and non-generated files at the same level:

I don't think it's a true statement in practice (maybe in theory it's true). In practice, in the generated files, they all include almost all of the other generated headers (transitively at least). Fixing this in the SLICC compiler, I think we can all agree, is absolutely not worth the effort.

powerjg:
Let me add one more thing:

I don't think it's a bad idea at all to move the implementation of the constructor to a cc file. That's all https://github.com/gem5/gem5/pull/2706/files does. 

Surely there's a way to have this behavior be correct for this one instance of a template parameter.

powerjg:
@BobbyRBruce Does this diff not work for you? Can you simply add an explicit template instantiation?

```
diff --git a/src/mem/ruby/structures/MN_TBETable.cc b/src/mem/ruby/structures/MN_TBETable.cc
--- a/src/mem/ruby/structures/MN_TBETable.cc
+++ b/src/mem/ruby/structures/MN_TBETable.cc
@@ -160,6 +160,8 @@ MN_TBETable::chooseNewDistributor()
 
 } // namespace CHI
 
+template class TBETable<CHI::MiscNode_TBE>;
+
 } // namespace ruby
 
 } // namespace gem5
```

powerjg:
Ok, after spending too many hours on this, I was able to concoct a solution that reverts #2706. @giactra you're going to have to promise that when I push my bazel stuff you're not going to complain about how I had to bend over backwards to get CHI to work, though!

powerjg:
Closing in favor of #2727

giactra:
> Ok, after spending too many hours on this, I was able to concoct a solution that reverts #2706. @giactra you're going to have to promise that when I push my bazel stuff you're not going to complain about how I had to bend over backwards to get CHI to work, though!

😆 😆 😆

PR #2718, opened 2025-10-30
arch-arm: Ensure Stage2Walk port pointer is used

BobbyRBruce:
Clang’s -Wunused-private-field is trigged when building because
`WalkUnit::Stage2Walk::port`, a private member declared in
src/arch/arm/table_walker.hh,  is written in the constructor at
src/arch/arm/table_walker.cc but never read anywhere else.
`fetchDescriptor()` still constructs `Stage2Walk` objects, but
subsequent access goes through `parent.getTableWalkerPort()`.

This patch ensures this port pointer is used by replacing the access
via `parent.getTableWalkerPort()` with the port.

### DIFF
--- a/src/arch/arm/table_walker.cc
+++ b/src/arch/arm/table_walker.cc
@@ -2721,8 +2721,7 @@ WalkUnit::Stage2Walk::finish(const Fault &_fault, const RequestPtr &req,
     }
 
     if (_fault == NoFault && !req->getFlags().isSet(Request::NO_ACCESS)) {
-        parent.getTableWalkerPort().sendTimingReq(req, data,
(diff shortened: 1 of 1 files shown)


giactra:
Thanks, makes sense!

Would you mind doing the opposite though? Which means using the cached port pointer? It's probably faster than calling the getTableWalker method

BobbyRBruce:
> Thanks, makes sense!
> 
> Would you mind doing the opposite though? Which means using the cached port pointer? It's probably faster than calling the getTableWalker method

Done. 

That actually worked out to be a much cleaner solution :)

giactra (approved):
Thanks for your changes!

PR #2725, opened 2025-10-31
cpu-o3: Add LQ and SQ average occupancy stat

giactra:
Current LSQ statistics report several info, but they lack one of the key ones, which is the average
occupancy of LD/ST in the LSQ.
This helps on understanding whether your workload is memory bound and your IPC is somehow limited by your LSQ (a full LSQ will stall the entire pipeline at rename).

We therefore add a lq and sq occupancy stat which will provide the average occupancy (from 0 to 1, where a number close to zero means there is plenty of space available and a number close to 1 means we are running at full capacity and the memsys is bottlenecking your performance)

Change-Id: I38c0b93e23b4b26fddb430fb3649fcdaf022393c


powerjg (approved):

dhschall (approved):
Thank you for the new addition.
Its perfect 👌

PR #2727, opened 2025-10-31
Revert "mem-ruby: Break dependence in CHI protocol"

powerjg:
Reverts gem5/gem5#2706


powerjg:
See discussion in #2717

BobbyRBruce (approved):
I've confirmed that this fixes the error we were getting. It was reproducible with `docker-run ghcr.io/gem5/gcc-version-10:latest scons build/ALL/gem5.opt -j$(nproc)` and can no longer be observed  when this revert is included.

I'm happy to close #2717 and accept this PR in its place.

powerjg:
Getting this "revert" commit to pass our bespoke formatting tests is a nightmare. This is a revert. Let's force it through.

PR #2731, opened 2025-11-02
arch-x86: raise #DE on division overflow

### LINKED ISSUE #2730: x86 DIV overflow does not raise fault
**Describe the bug**
On x86, division overflows should raise a #DE (divide error) fault. However, it doesn't in gem5.

Example program (test.asm):
```
    global _start

_start:
    mov rax, 0
    mov rdx, 1
    mov rcx, 1
    div rcx

    mov eax, 60
    mov edi, 0
    syscall
```

**Affects version**
develop @ [d65257d](https://github.com/gem5/gem5/commit/d65257daf3f60d000cb33e4b12e79bdbeb02b878)

**gem5 Modifications**
None

**To Reproduce**

1. Compile gem5: scons build/X86/gem5.opt
2. Compile the test program: `nasm -felf64 -o test.o test.asm && ld -o test test.o`
3. Run the test on the host and observe a #DE error: `./test`
4. Run the test on gem5, observe no #DE error: `./build/X86/gem5.opt configs/deprecated/example/se.py --cmd ./test`

**Terminal Output**
On the host:
```
$ ./div
Floating point exception
```

```
$ ./build/X86/gem5.opt configs/deprecated/example/se.py --cmd ./div

gem5 version DEVELOP-FOR-25.1
gem5 compiled Aug  5 2025 09:31:58
gem5 started Nov  2 2025 10:05:58
gem5 executing on cafe-cet, pid 154085
command line: ./build/X86/gem5.opt configs/deprecated/example/se.py --cmd ./div

system.remote_gdb: Listening for connections on port 7023
**** REAL SIMULATION ****
Exiting @ tick 9000 because exiting with last active thread context
```

**Expected behavior**
gem5 should raise a divide error.

**Host Operating System**
Ubuntu 22.04

**Host ISA**
ARM, X86, RISC-V, etc.

**Compiler used**
State which compiler was used to compile gem5. Please include the compiler version.

**Additional information**

nmosier:
Fix #2730. Check for overflow when performing division. If the quotient does not fit in 8/16/32/64 bits for a 8-/16-/32-/64-bit div2 micro-op, then raise a divide error (#DE).

Change-Id: Ic06714ef9546c3542316ec3a3a74042fb613da64

### DIFF
--- a/src/arch/x86/insts/static_inst.cc
+++ b/src/arch/x86/insts/static_inst.cc
@@ -38,6 +38,7 @@
 
 #include "arch/x86/insts/static_inst.hh"
 
+#include "arch/x86/faults.hh"
 #include "arch/x86/regs/segment.hh"
 #include "cpu/reg_class.hh"
 
@@ -110,7 +111,8 @@ void X86StaticInst::printSegment(std::ostream &os, int segment)
 
 void
 X86StaticInst::divideStep(uint64_t dividend, uint64_t divisor,
-        uint64_t &quotient, uint64_t &remainder)
+                          uint64_t &quotient, uint64_t &remainder,
+                          Fault &fault)
 {
     // Check for divide by zero.
     assert(divisor != 0);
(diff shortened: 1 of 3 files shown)


powerjg (dismissed):
Thanks for fixing this!

powerjg (approved):

PR #2738, opened 2025-11-07
cpu-o3: Increase MaxWidth to 16

giactra:
The MaxWidth controls the maximum width of the pipeline. With cores getting bigger and bigger, there's really no harm in increasing it other than midly expanding some arrays. At the moment affected data structures are in src/cpu/o3/comm.hh and this patch increases their size by 4 * sizeof(pointer) so around 32bytes in 64bit machines

Change-Id: I90fbcf15d79658d02e708c213a93a65de89a8c65

### DIFF
--- a/src/cpu/o3/limits.hh
+++ b/src/cpu/o3/limits.hh
@@ -1,4 +1,16 @@
 /*
+ * Copyright (c) 2025 Arm Limited
+ * All rights reserved
+ *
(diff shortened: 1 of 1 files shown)


dhschall (approved):
LGTM. I actually bumped this up to >1k to do some exploration and its still worked. (was needed without the instant squash PR to flush the entire ROB at once)

PR #2740, opened 2025-11-07
mem-ruby: Update Ruby Network to use new-style stats

powerjg:
Ruby is one of the last places in the codebase that are still using the old way of initializing stats. This change updates the network and simple network with the new style stats.

This change also fixes a bug where the Switch stats were not reset.

There's one wrinkle. The stats can't be initialized until after ports are connected and all of the internal/external links have been created. To get around this, we override the regStats functions in the stats objects.

*User-facing changes*
- All of the modified stats now have descriptions and units

### DIFF
--- a/src/mem/ruby/network/simple/PerfectSwitch.cc
+++ b/src/mem/ruby/network/simple/PerfectSwitch.cc
@@ -306,16 +306,6 @@ PerfectSwitch::storeEventInfo(int info)
 }
 
 void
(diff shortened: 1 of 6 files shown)


powerjg:
I should add that I tested this by running `configs/example/gem5_library/caches/octopi-cache-example.py` for a few ms and compared the stats before and after. They were exactly the same (except for the descriptions).

BobbyRBruce (approved):

PR #2741, opened 2025-11-07
arch-arm,sim-se: Implement sigreturn for Arm64

powerjg:
The sigreturn system call takes CPU context data from the (usermode) stack and applies it to the processor. This is usually used for returning to the original program location after a signal, but it can also be used to restore arbitrary program state (e.g., for SE-mode "checkpoints").

This patch implements a subset of the system call to enable usermode checkpoint restore using sigreturn.

Currently missing: Any floating point data


src/arch/arm/linux/se_workload.cc
powerjg, line 1: I know it's a bunch of changes, but it's from clang-format. I think it's OK, but if someone would like me to try to break this up into multiple commits, I can do that.

src/arch/arm/linux/linux.hh
```diff
@@ -568,6 +569,45 @@ class ArmLinux64 : public ArmLinux, public OpenFlagTable<ArmLinux64>
...
+                int ss_flags;
+            } uc_stack;
+            unsigned char uc_sigset_ex[136];
+            struct
+            {
+                unsigned long long int fault_address;
+                unsigned long long int regs[31];
+                unsigned long long int sp;
+                unsigned long long int pc;
+                unsigned long long int pstate;
```
BobbyRBruce: Any reason not to use `uint64_t` here? I believe that the C standard states `long long int` guarantees a value of _at least_ 64 bits while `uint64_t` guarantees _exactly_ 64 bits. Ergo, the latter is more portable.

I also think `uint64_t` is easier to read. I always need a minute to remind myself what `long long` means.

giactra: That's a good catch @BobbyRBruce .

I will add this is not just for the sake of readability. If we run gem5 on a 32bit system, we would get the wrong tgt_struct size when trying to emulate an Arm64 SE system

powerjg: Thanks, great suggestion. Done.

BobbyRBruce (approved):

giactra (approved):

PR #2756, opened 2025-11-15
mem-cache: Add unit test for LFU RP

odanrc:
As side effects:

### DIFF
--- a/src/mem/cache/replacement_policies/lfu_rp.test.cc
+++ b/src/mem/cache/replacement_policies/lfu_rp.test.cc
@@ -34,51 +34,27 @@
 #include "mem/cache/replacement_policies/lfu_rp.hh"
 #include "params/LFURP.hh"
 
-// Create a dummy RP class so that we can verify if the replacement data is
-// of the expected type for this RP
-class DummyLFU : public gem5::replacement_policy::LFU
-{
-  public:
-    using gem5::replacement_policy::LFU::LFU;
-
-    /// Verify that the data is of the expected type
-    bool
-    validateType(
-        const std::shared_ptr<gem5::replacement_policy::ReplacementData>
-            repl_data) const
-    {
-        return nullptr !=
-               std::dynamic_pointer_cast<
-                   gem5::replacement_policy::LFU::LFUReplData>(repl_data);
-    }
-};
-
 /// Common fixture that initializes the replacement policy
 class LFURPTestF : public ::testing::Test
 {
   public:
-    std::shared_ptr<DummyLFU> rp;
+    std::shared_ptr<gem5::replacement_policy::LFU> rp;
 
     LFURPTestF()
     {
         gem5::LFURPParams params;
         params.eventq_index = 0;
-        rp = std::make_shared<DummyLFU>(params);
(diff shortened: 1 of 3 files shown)


src/mem/cache/replacement_policies/lfu_rp.test.cc
```diff
@@ -0,0 +1,286 @@
...
+        // Set one of the entries as invalid
+        rp->invalidate(entry.replacementData);
+
+        ASSERT_EQ(rp->getVictim(candidates), &entry);
+    }
+}
+
+/// Test that when all entries have been reset the same entry will always
+/// be the one victimized
+TEST_F(LFURPVictimizationTestF, GetVictimAllReset)
```
erin-le, line 142: Is there supposed to be more to this test? The comment mentions checking that the same entry is "always" victimized, so I was wondering if there were other conditions you also wanted to check.

erin-le (commented):
@odanrc: Hello, thank you for this PR. The tests generally look good to me, but it looks like this PR is failing the CI tests because of the dynamic_pointer_cast you have on lines 51 and 52. Could you take a look? Thank you!

src/mem/cache/replacement_policies/lfu_rp.test.cc
odanrc: Not really; I reused the tests/documentation from the other test, making the necessary changes. In theory it would make sense to run other scenarios (e.g., reset entry[0] last to make sure it is still picked), but at this point it feels like a guessing game of which bugs could occur.

src/mem/cache/replacement_policies/lfu_rp.test.cc
```diff
@@ -0,0 +1,286 @@
...
+  public:
+    using gem5::replacement_policy::LFU::LFU;
+
+    /// Verify that the data is of the expected type
+    bool
+    validateType(
+        const std::shared_ptr<gem5::replacement_policy::ReplacementData>
+            repl_data) const
+    {
+        return nullptr !=
```
odanrc: The compiler complains: ´source type is not polymorphic´

The issue is that the empty struct/base type ReplacementData is not polymorphic, and I would have to at least declare its destructor virtual. I don´t think it is sane to do this change in the source because of a test, so I have removed this validation from the test.

odanrc:
@erin-le , are you able to see what is the clang-format error? I run it locally, and it complains about nothing. I am using clang-format 15.0.7. What is the version being used for the verification?

odanrc:
> @erin-le , are you able to see what is the clang-format error? I run it locally, and it complains about nothing. I am using clang-format 15.0.7. What is the version being used for the verification?

I rebased on top of develop and uploaded again, and the error is gone. I didn't make any changes to the code...

erin-le (approved):
LGTM. I'm not 100% sure about this, but I think the clang-format test sometimes fails if the `.clang-format` file is updated on `develop` after the PR branch has been created, and the `develop` branch and PR end up checking for different formatting. The last update to the `.clang-format` file was 4 months ago, however, so I don't know if that's the case here.

PR #2757, opened 2025-11-15
mem-cache: Add unit test for LRURP

odanrc:
As side effects:

### DIFF
--- a/src/mem/cache/replacement_policies/lru_rp.test.cc
+++ b/src/mem/cache/replacement_policies/lru_rp.test.cc
@@ -35,7 +35,7 @@
 #include "params/LRURP.hh"
 
 // We need to initialize the event queue to be able to count ticks
-gem5::EventQueue eventQueue("MRURPTest Queue");
+gem5::EventQueue eventQueue("LRURPTest Queue");
 
 /// Common fixture that initializes the replacement policy
 class LRURPTestF : public ::testing::Test
--- a/src/mem/cache/replacement_policies/lru_rp.test.cc
+++ b/src/mem/cache/replacement_policies/lru_rp.test.cc
@@ -1,0 +1,334 @@
+/**
+ * Copyright (c) 2025 Daniel R. Carvalho
+ * All rights reserved.
+ *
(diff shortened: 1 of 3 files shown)


erin-le (commented):
@odanrc: Hello, the tests for LRU RP generally look good to me. I tried running these tests locally, and it looks like the build fails due to the dynamic_pointer_cast on line 54 of lru_rp.test.cc. Could you take a look? Could you also remove the changes for LFU RP from this PR? Thank you!

odanrc:
> @odanrc: Hello, the tests for LRU RP generally look good to me. I tried running these tests locally, and it looks like the build fails due to the dynamic_pointer_cast on line 54 of lru_rp.test.cc. Could you take a look? Could you also remove the changes for LFU RP from this PR? Thank you!

I wish I could remove the changes for LFU, but I am completely unable to change the base branch to the unit_test_lfu_rp branch, so the changes from the LFU PR, which I want to merge first, ends up showing in this PR too. Otherwise, I have removed the dynamic cast, like in the LFU PR.

src/mem/cache/replacement_policies/lru_rp.test.cc
```diff
@@ -0,0 +1,310 @@
...
+#include <gtest/gtest-spi.h>
+#include <gtest/gtest.h>
+
+#include <cassert>
+
+#include "mem/cache/replacement_policies/lru_rp.hh"
+#include "params/LRURP.hh"
+
+// We need to initialize the event queue to be able to count ticks
+gem5::EventQueue eventQueue("MRURPTest Queue");
```
erin-le: Should this be:

```suggestion
gem5::EventQueue eventQueue("LRURPTest Queue");
```

erin-le (dismissed):
LGTM

src/mem/cache/replacement_policies/lru_rp.test.cc
odanrc: Now you know where I got my template from :)

Fixed

erin-le (approved):

PR #2759, opened 2025-11-18
arch-riscv: Only show RVV message when we enable it

### LINKED ISSUE #2712: "RVV enabled" is printed with RVV disabled
**Describe the bug**
The info message "RVV enabled, VLEN = %d bits, ELEN = %d bits" is printed unconditionally even if you set enable_rvv=false. 

**Affects version**
Found in 25.0.0.1 (stable branch), also present on the develop branch. 

**gem5 Modifications**

**To Reproduce**
Modify the `configs/learning_gem5/part1/simple-riscv.py` by adding the following in line 44:

```Python
...
system.cpu = RiscvTimingSimpleCPU()
system.cpu.isa = RiscvISA(enable_rvv=False)

system.membus = SystemXBar()
...
```
Run with `gem5 configs/learning_gem5/part1/simple-riscv.py`

**Terminal Output**

```shell
src/arch/riscv/isa.cc:279: info: RVV enabled, VLEN = 256 bits, ELEN = 64 bits
Beginning simulation!
src/sim/syscall_emul.hh:1117: warn: readlink() called on '/proc/self/exe' may yield unexpected results in various settings.
      Returning '/workspaces/gem5/tests/test-progs/hello/bin/riscv/linux/hello'
Hello world!
Exiting @ tick 488366000 because exiting with last active thread context
```

**Expected behavior**
Instead of RVV enabled and the VLEN, ELEN information it should print RVV disabled.

**Additional information**
Could be easily fixed in `src/arch/riscv/isa.cc`. The constructor always prints "RVV enabled". Just add an if branch around that which checks `p.enable_rvv`.

TommyWu-fdgkhdkgh:
Onriginally, the info message "RVV enabled, VLEN = %d bits, ELEN = %d bits" is printed unconditionally even if you set enable_rvv=false. This patch try to fix this issue.

Related issue : #2712

### DIFF
--- a/src/arch/riscv/isa.cc
+++ b/src/arch/riscv/isa.cc
@@ -314,12 +314,12 @@ ISA::ISA(const Params &p) : BaseISA(p, "riscv"),
     _regClasses.push_back(&ccRegClass);
     _regClasses.push_back(&miscRegClass);
 
-    fatal_if( p.vlen < p.elen,
-    "VLEN should be greater or equal",
-        "than ELEN. Ch. 2RISC-V vector spec.");
+    if (enableRvv) {
+        fatal_if(p.vlen < p.elen, "VLEN should be greater or equal",
+                 "than ELEN. Ch. 2RISC-V vector spec.");
 
-    inform("RVV enabled, VLEN = %d bits, ELEN = %d bits",
-            p.vlen, p.elen);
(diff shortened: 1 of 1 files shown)


BobbyRBruce:
Thanks! The solution makes sense to me!

BobbyRBruce (approved):

PR #2761, opened 2025-11-19
base-stats: Add m5_stats.Group to stats dump

jonathanhaley:
m5_stats.Group objects were being dropped when stats were being processed. As a result 
stats group objects were valueless. This commit adds processing for that stat type.

Change-Id: Ia50530bb6243a10e362ba0395e6e14139803df33


BobbyRBruce (approved):
Thanks for the fix! LGTM

PR #2765, opened 2025-11-21
arch-arm: Implement crypto instructions for FEAT_SVE/2

giactra:


powerjg (approved):

powerjg:
@giactra this breaks the build on MacOS

```
In file included from build/ALL/arch/arm/generated/generic_cpu_exec_6.cc:17:
build/ALL/arch/arm/generated/exec-ns.cc.inc:312150:27: error: variable length arrays in C++ are a Clang extension [-Werror,-Wvla-cxx-extension]
 312150 |         Element tempElem1[eltspersegment];
        |                           ^~~~~~~~~~~~~~
build/ALL/arch/arm/generated/exec-ns.cc.inc:312150:27: note: read of non-const variable 'eltspersegment' is not allowed in a constant expression
build/ALL/arch/arm/generated/exec-ns.cc.inc:312148:18: note: declared here
 312148 |         uint32_t eltspersegment = 16 / sizeof(Element);
 ```

I don't know if I'll get around to posting a PR today or not.

PR #2766, opened 2025-11-21
arch-vega: Support MUBUF to LDS instructions

abmerop:
The MUBUF instruction format allows reading data from a memory buffer
directly into LDS without passing through VGPRs. Currently gem5 does not
support this correctly as it marks instructions with LDS dests as group
coherent. This causes them to increment a local memory pipe counter, but
are issued to the global memory pipeline. This causes an error such as:

src/gpu-compute/wavefront.cc:831: panic: panic condition wrGmReqsInPipe
< 0 || rdGmReqsInPipe < 0 || wrLmReqsInPipe < 0 || rdLmReqsInPipe < 0 ||
outstandingReqs < 0 occurred: ...

### DIFF
--- a/src/arch/amdgpu/vega/insts/mubuf.cc
+++ b/src/arch/amdgpu/vega/insts/mubuf.cc
@@ -577,11 +577,7 @@ namespace VegaISA
     {
         setFlag(MemoryRef);
         setFlag(Load);
-        if (instData.LDS) {
-            setFlag(GroupSegment);
-        } else {
-            setFlag(GlobalSegment);
-        }
+        setFlag(GlobalSegment);
     } // Inst_MUBUF__BUFFER_LOAD_UBYTE
 
     Inst_MUBUF__BUFFER_LOAD_UBYTE::~Inst_MUBUF__BUFFER_LOAD_UBYTE()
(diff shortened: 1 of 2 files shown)


src/arch/amdgpu/vega/insts/mubuf.cc
```diff
@@ -784,6 +793,25 @@ namespace VegaISA
...
+                reinterpret_cast<VecElemU32 *>(gpuDynInst->d_data);
+            VecElemU16 *as_u16 =
+                reinterpret_cast<VecElemU16 *>(gpuDynInst->d_data);
+
+            for (int lane = NumVecElemPerVecReg - 1; lane >= 0; --lane) {
+                // Reverse copy from largest to smallest index as the pointers
+                // are the same. Avoids clobbering.
+                if (gpuDynInst->exec_mask[lane]) {
+                    if (!oobMask[lane]) {
+                        as_u32[lane] = as_u16[lane];
```
mattsinc, line 808: Does this ever go out of bounds?  Or is it just datatype, not array size, that differs?

mattsinc (approved):

src/arch/amdgpu/vega/insts/mubuf.cc
abmerop: It won't in this case. Here we know d_data allocates 4x dwords per lane as that is the maximum load/store size per instruction. Here we have a maximum of one dword

PR #2769, opened 2025-11-24
dev-amdgpu: Update mmhub size

abmerop:
The mmhub is an aliased range for vram and should be the same size.


abmerop:
Adding v25.1 as this fixes a bug with mi200 config where mmhub is too small.

mattsinc (approved):

PR #2772, opened 2025-11-25
arch-x86: Fix assertion due to clflush in DataTranslation::finish

odxa20:
CLFLUSHOPT/WB/FLUSH are treated as reads to avoid page faults when done on write-protected pages. This results in an assertion when `translateTiming `calls `translation->finish` as it expects the `mode `to be consistent with `state->mode`. To fix this keep original mode and pass that to finish function

related PR: [#1080](https://github.com/gem5/gem5/pull/1080)

### DIFF
--- a/src/arch/x86/tlb.cc
+++ b/src/arch/x86/tlb.cc
@@ -578,12 +578,13 @@ TLB::translateTiming(const RequestPtr &req, ThreadContext *tc,
     bool delayedResponse;
     assert(translation);
     // CLFLUSHOPT/WB/FLUSH should be treated as read for protection checks
+    BaseMMU::Mode orig_mode = mode;
     if (req->isCacheClean())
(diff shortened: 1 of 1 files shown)


powerjg (dismissed):
From a hardware security perspective this change scares me (we're checking for read permission instead of write), but since this is just a model, I think the change is OK.

powerjg:
@odxa20 can you rebase and clean up the commit messages? Also, the main (only?) commit message should read the same as your PR message.

powerjg:
@BobbyRBruce This can go onto staging if it doesn't make it in before that.

odxa20:
Done!

BobbyRBruce (approved):

BobbyRBruce:
> @BobbyRBruce This can go onto staging if it doesn't make it in before that.

I'll make sure this gets in.

PR #2773, opened 2025-11-25
stats: Add lastCommitPC to ThreadStateStats

cyyself:
This commit adds a new statistic, lastCommitPC, to the ThreadStateStats structure in thread_state.hh. This statistic tracks the program counter (PC) of the last instruction that was committed by the CPU. We have also changed simple, O3, and minor CPU models to update this statistic whenever an instruction is committed.

This addition provides valuable insight into the execution flow of the simulated CPU, allowing users to monitor the last executed instruction's address. This can be particularly useful when used with `periodicStatDump` to trace sampled execution points over time.

### DIFF
--- a/src/cpu/minor/execute.cc
+++ b/src/cpu/minor/execute.cc
@@ -881,6 +881,7 @@ Execute::doInstCommitAccounting(MinorDynInstPtr inst)
     {
         thread->numInst++;
         thread->threadStats.numInsts++;
+        thread->threadStats.lastCommitPC = inst->pc->instAddr();
         cpu.commitStats[tid]->numInsts++;
         cpu.executeStats[tid]->numInsts++;
         cpu.baseStats.numInsts++;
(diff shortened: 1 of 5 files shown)


cyyself:
In addition, I'm writing a tool to convert `stats.txt` to Linux-perf's `perf.data` to help us analyze the execution from gem5 `stats.txt` with `periodicStatDump`.

powerjg:
Quick question: why not use a commit probe point for this?

erin-le:
@cyyself: Hello, could you change `base-stats` in the PR and commit titles to just `stats` instead? Thank you!

erin-le:
@cyyself: Hello, thanks for modifying the PR title. Could you also modify the commit title to also use `stats` instead of `base-stats`? Furthermore, could you respond to powerjg's question? Thank you!

cyyself:
> @cyyself: Hello, thanks for modifying the PR title. Could you also modify the commit title to also use `stats` instead of `base-stats`? Furthermore, could you respond to powerjg's question? Thank you!

OK. I will check the commit probe point soon. If I still need this for some reason, I will update the commit title.

erin-le:
@cyyself: Hello, have you had the chance to look into commit probe points?

PR #2782, opened 2025-11-27
arch-arm: Do some OpClass retagging for some SIMD instructions

giactra:
With this PR we are

1. Retagging some existing instructions
2. Defining new OpClasses ("SimdSha3", "SimdCrc", "SimdDotProd")

The aim is to ensure some SIMD instructions get a more accurate depiction of the operation being performed.

### DIFF
--- a/src/cpu/o3/FuncUnitConfig.py
+++ b/src/cpu/o3/FuncUnitConfig.py
@@ -121,6 +121,7 @@ class SIMD_Unit(FUDesc):
         OpDesc(opClass="SimdShaSigma3"),
         OpDesc(opClass="SimdSha3"),
         OpDesc(opClass="SimdSm4e"),
+        OpDesc(opClass="SimdCrc"),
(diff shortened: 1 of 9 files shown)


src/cpu/o3/FuncUnitConfig.py
```diff
@@ -126,6 +127,7 @@ class SIMD_Unit(FUDesc):
         OpDesc(opClass="SimdBf16MatMultAcc"),
         OpDesc(opClass="SimdBf16Mult"),
         OpDesc(opClass="SimdBf16MultAcc"),
+        OpDesc(opClass="SimdSha3"),
```
powerjg: Is this missing `SimdCrc`?

giactra: Right! Nice catch! I might have missed it during rebase

BobbyRBruce (approved):

PR #2789, opened 2025-12-01
base: Add destructors to stats objects

powerjg:
This fixes some ASAN issues when running stats tests

### DIFF
--- a/src/base/statistics.hh
+++ b/src/base/statistics.hh
@@ -279,6 +279,8 @@ class DataWrap : public InfoAccess
             warn_once("One of the stats is a legacy stat. " + common_message);
     }
 
+    virtual ~DataWrap() { delete info(); }
+
     /**
      * Set the name and marks this stat to print at the end of simulation.
      * @param name The new name.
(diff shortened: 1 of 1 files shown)


src/base/statistics.hh
```diff
@@ -1317,6 +1319,13 @@ class DistBase : public DataWrap<Derived, DistInfoProxy>
     {
     }
 
+    virtual ~DistBase()
+    {
+        if (this->info()) {
+            data()->~Storage();
```
odanrc, line 1327: I think it would be safer to add a comment in the code explaining why we are doing a direct destructor call, instead of a deletion. I am assuming it is because of the 'placement new'.

src/base/statistics.hh
```diff
@@ -279,6 +279,8 @@ class DataWrap : public InfoAccess
             warn_once("One of the stats is a legacy stat. " + common_message);
     }
 
+    virtual ~DataWrap() { delete info(); }
```
odanrc, line 282: nit: making InfoAccess's destructor virtual does not seem to be required, as we don't seem to have `InfoAccess *info = new DataWrap` (or akin), but it would be less bug-prone in case things change in the future

odanrc (approved):

src/base/statistics.hh
powerjg: Yeah, this was an AI suggestion. It doesn't hurt, though.

src/base/statistics.hh
powerjg: Done. Great point.

BobbyRBruce (approved):

powerjg:
@BobbyRBruce can we cherry pick this onto staging? I thought that it was going to get in

PR #2794, opened 2025-12-02
mem: Skip unused chunks when restoring checkpoints

henrick-bonda:
Problem:
When restoring from a checkpoint previously the loader performed gzread() directly into the destination memory (pmem) in fixed-size chunks. This meant we always allocated RAM pages, even when the compressed data represented all-zero blocks. On large memory images, this resulted in wasting physical memory on the host system.

What the changes do:
Reads each chunk into a temporary buffer instead of writing directly to the target memory.
Scans the chunk to determine whether it contains only zero bytes.
Writes the chunk to pmem only if it contains non-zero data.
Still advances pmem and curr_size to preserve the correct memory layout.

### DIFF
--- a/src/mem/physical.cc
+++ b/src/mem/physical.cc
@@ -465,22 +465,21 @@ PhysicalMemory::unserializeStore(CheckpointIn &cp)
     uint64_t curr_size = 0;
     uint32_t bytes_read;
     uint8_t buffer[chunk_size];
+    uint8_t zeros[chunk_size] = {0};
     while (curr_size < range.size()) {
         bytes_read = gzread(compressed_mem, buffer, chunk_size);
         if (bytes_read == 0)
             break;
 
-        bool all_zero = true;
-        for (uint32_t i = 0; i < bytes_read; i++) {
-            if (buffer[i] != 0) {
-                all_zero = false;
-                break;
-            }
-        }
+        bool all_zero = (memcmp(buffer, zeros, bytes_read) == 0);
 
         if (!all_zero) {
             memcpy(pmem, buffer, bytes_read);
         }
+
+        curr_size += bytes_read;
+        pmem += bytes_read;
+    }
 
     if (gzclose(compressed_mem))
         fatal("Close failed on physical memory checkpoint file '%s'\n",
--- a/src/mem/physical.cc
+++ b/src/mem/physical.cc
@@ -464,13 +464,23 @@ PhysicalMemory::unserializeStore(CheckpointIn &cp)
 
     uint64_t curr_size = 0;
     uint32_t bytes_read;
+    uint8_t buffer[chunk_size];
     while (curr_size < range.size()) {
-        bytes_read = gzread(compressed_mem, pmem, chunk_size);
+        bytes_read = gzread(compressed_mem, buffer, chunk_size);
         if (bytes_read == 0)
             break;
-        curr_size += bytes_read;
-        pmem += bytes_read;
-    }
+
+        bool all_zero = true;
+        for (uint32_t i = 0; i < bytes_read; i++) {
+            if (buffer[i] != 0) {
+                all_zero = false;
+                break;
+            }
+        }
+
+        if (!all_zero) {
+            memcpy(pmem, buffer, bytes_read);
+        }
 
     if (gzclose(compressed_mem))
         fatal("Close failed on physical memory checkpoint file '%s'\n",
--- a/src/mem/physical.cc
+++ b/src/mem/physical.cc
@@ -76,14 +76,19 @@ namespace gem5
 namespace memory
 {
 
-PhysicalMemory::PhysicalMemory(const std::string& _name,
-                               const std::vector<AbstractMemory*>& _memories,
+PhysicalMemory::PhysicalMemory(const std::string &_name,
+                               const std::vector<AbstractMemory *> &_memories,
                                bool mmap_using_noreserve,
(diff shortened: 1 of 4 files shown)


powerjg (commented):
I understand the problem this is solving. However, I'm concerned about the performance impact in two ways:
1. we are now doing an extra memcpy for each chunk
2. the for loop to check for 0s is slow. Could we use `std::all_of` or something else from `<algorithm>`?

Can you speak to the change in performance when loading checkpoints with large physical memory footprints?

We may want to go further to support sparse memories in checkpoints. We may want to change the format in which the memory is saved instead of just leveraging gzip.

giactra:
Thanks @powerjg I was thinking along the same lines. To add another alternative: If we have zero pages, can't we just reduce the physical footprint with existing optimization techniques on Linux (for instance like KSM: https://en.wikipedia.org/wiki/Kernel_same-page_merging) ?

powerjg:
I don't think KSM is going to kick in fast enough to enable us to say, run multiple large guests on a single host (e.g., two 64 GiB guests on a host with 64 GiB of memory).

I think this is just a checkpoint restore problem, right @henrick-bonda? When running from scratch, the virtual (host) pages are never touched so the mmap'ed memory is never physically allocated. I think this is hard to replicate after a checkpoint unless we somehow track dirtiness on the guest memory in gem5.

Your solution definitely works. I am just concerned about the performance. Maybe this could be a parameter to the `System` SimObject?

henrick-bonda:
@powerjg yes exactly this is only a problem on restoring from a checkpoint.

To your concerns:
1. The loop I addressed in the new commit with ```memcmp```.
2. The impact on performance of ```memcpy``` probably needs to be investigated. However, I want to point out that the impact largely dependents on how many zero pages are present.

powerjg:
> The impact on performance of memcpy probably needs to be investigated. However, I want to point out that the impact largely dependents on how many zero pages are present.

Right. And if you take the case of no zero pages (the common case?), this change will be much slower than what we have today. If there are a significant fraction of zero pages, then this code is much faster. 

I'm worrying that we're optimizing for the uncommon case and making the common case slower. That's why i think we need this information embedded in the checkpoint output somehow.

henrick-bonda:
I understand your concern and I see the following 4 strategies:

1. Save some information when checkpointing how much RAM was utilized. If there are more than e.g. 50% zero pages then take the new  approach, otherwise take the old approach. This is what you suggested in the last comment, right?

2. Lets say we have two cases, both cases with 50% RAM utilization.
Case 1: a system configured with 2 GB RAM
Case 2: a system configured with 16 GB RAM
In case 1 the gain would be only 1 GB, however the gain in case 2 would be 8 GB.
So you could also optimize for the size of the configured system.

3. A combination of strategy 1 and 2 to benefit from both advantages and mitigate disadvantages.

4. Leave the choice to the user, since the user knows best how much RAM is allocated and if he would benefit from the new approach. So add an optional flag which approach to choose, standard case is the old approach and on demand the user can enable the new approach.

I would prefer option 4, because it is the most flexible. Strategy 1 and 2 are in my opinion a bad tradeoff and strategy 3 is not flexible.

powerjg:
Sorry for the slow response. Option 4 sounds good to me.

odanrc:
Just throwing my 2 cents: another option would be to add metadata informing whether each page/group of pages is null (a classic in memory compression). Contrary to compression, though, we would be making the checkpoints larger.

erin-le:
@henrick-bonda: Hello, could you give an update on the status of this PR on your end? Thank you!

henrick-bonda:
Hi, currently I am not working on this PR. Probably later in March I will have some time to work on it.

src/mem/physical.cc
```diff
@@ -464,12 +469,33 @@ PhysicalMemory::unserializeStore(CheckpointIn &cp)
...
     uint64_t curr_size = 0;
     uint32_t bytes_read;
-    while (curr_size < range.size()) {
-        bytes_read = gzread(compressed_mem, pmem, chunk_size);
-        if (bytes_read == 0)
-            break;
-        curr_size += bytes_read;
-        pmem += bytes_read;
+    if (isSparseRestore) {
+        uint8_t buffer[chunk_size];
```
powerjg, line 479: Do you want to add a (static) assert here that chunk size is something reasonable? E.g., if someone changes the chunk size to `1000` it would probably break the performance improvement of this code.

src/sim/System.py
```diff
@@ -114,6 +114,10 @@ class System(SimObject):
         "shared_backstore is non-empty.",
     )
 
+    is_sparse_restore = Param.Bool(
+        False, "Skip zero-filled pages during restore"
```
powerjg: Can you add a little more info here. E.g., if someone with no context read this option they should know what it does.

Maybe something like... "Skip zero-filled pages during restore. For guest memories that have many zero pages, this improves restore performance."

powerjg (commented):
This looks great to me! Just a couple of small things below.

powerjg (dismissed):

erin-le:
@henrick-bonda: Hello, it looks like the CI tests for this PR are failing due to build errors. Could you look into what's causing them? Thank you!

henrick-bonda:
The issue is that the constructor of PhysicalMemory has been changed, and all the places where it is called are now missing the argument I added.

powerjg (approved):

PR #2797, opened 2025-12-03
arch-riscv: Stop leaking snoop state into A/D writes

BobbyRBruce:
The RISC‑V walker reuses its read Packet when it needs to write back the Accessed/Dirty bits. That Packet already went through the coherent fabric, so it may have `cacheResponding`/express‑snoop flags set. When the walker reused it for the write, the SystemXBar saw a normal request with those snoop flags and fired the `is_express_snoop == cacheResponding` assert.

Fix by allocating a fresh Packet before issuing the write and only copying the minimal metadata we still need (`hasSharers`). This lets the walker keep the line‑sharing information so downstream caches don’t panic about keeping a dirty block, while ensuring the write request no longer carries stale snoop responder state.

This fixes the failing daily tests: https://github.com/gem5/gem5/actions/runs/19871109055/job/56958444742

### DIFF
--- a/src/arch/riscv/pagetable_walker.cc
+++ b/src/arch/riscv/pagetable_walker.cc
@@ -837,17 +837,23 @@ Walker::WalkerState::stepWalk(PacketPtr &write)
 
     if (stepWalkFlags.doEndWalk) {
         // If we need to write, adjust the read packet to write the modified
-        // value back to memory.
+        // value back to memory. Use a fresh packet so any responder flags set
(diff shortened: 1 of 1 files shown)


BobbyRBruce:
Disclaimer: This patch is largely AI generated, but i did confirm it fixes the bug I observed.

powerjg (approved):
lgtm. Seems like a real bug and a good solution.

PR #2800, opened 2025-12-03
configs: Fix tuple unpacking error in Simulation.py with fast-forward

ivin1:
**Description**
This PR fixes a bug where `se.py` crashes with an `AttributeError: 'tuple' object has no attribute 'numThreads'` when running with the `--fast-forward` option.

**Root Cause**
In `configs/common/Simulation.py`, the `setCPUClass` function returns a tuple of `(AtomicSimpleCPU, O3CPU)` when fast-forwarding is enabled.
However, inside the function, the return value from `getCPUClass` (which is `(TmpClass, test_mem_mode)`) was assigned to a single variable `TmpClass` without unpacking. This caused `TmpClass` to become a nested tuple, leading to the crash downstream in `se.py`.

**Fix**
I corrected the logic in `Simulation.py` to properly unpack the returned values from `getCPUClass` into `TmpClass` and `test_mem_mode`.

**Verification**
Tested with `se.py` and `--fast-forward` option. The simulation now runs without the AttributeError.


powerjg (approved):

powerjg (changes requested):
Oops. This looks fine to me. But it the target should be "develop" not stable. Can you hit the "edit" button and change the target branch?

ivin1:
Changed the target branch to develop. Thanks!

powerjg (approved):

BobbyRBruce (approved):

PR #2802, opened 2025-12-03
gpu-compute: Add missing MFMA timings

abmerop:
There are some MFMA timings for MI300X missing and all CDNA4 timings. The simulator issues a panic if the timing is not found, therefore this fixes some applications which prefer the MFMAs that were missing.


src/gpu-compute/compute_unit.cc
```diff
@@ -65,130 +65,197 @@
...
+      mfma_cycles(
+          {// gfx90a is MI200 series (MI210, MI250X). The latency values are
+           // the
+           // "passes" in the MI200 Instruction Set Architecture reference
+           // listed
+           // for each instruction in section 12.10:
+           // https://www.amd.com/content/dam/amd/en/documents/
+           //         instinct-tech-docs/instruction-set-architectures/
+           //         instinct-mi200-cdna2-instruction-set-architecture.pdf
+           {GfxVersion::gfx90a,
```
powerjg, line 126: Would it make sense to move this out of the constructor into a static variable? It's pretty hard to read as it is.

powerjg (approved):
It's fine as-is. I guess this is following the style that's there. I would prefer that the constructor isn't so massive, though.

mattsinc (approved):

src/gpu-compute/compute_unit.cc
mattsinc: We had discussed this (maybe not with @powerjg ?) when @mjkpolo integrated the original support for this.  I can't recall why we went this way, do you @abmerop ?

abmerop: I'm not sure. It does not have to be in the constructor since the type is not const.

PR #2812, opened 2025-12-05
util: Bump GPU build docker to ROCm 7.0

abmerop:
This docker image is used to build applications to be run within gem5. It is not a disk image.


abmerop:
I am debating a different approach where users could use the disk image itself to compile GPU applications using something like `systemd-nspawn` so this does not need to keep being updated when the disk image is updated.  My main concern is that tool requires sudo afaik and I believe it to be more likely a user can run docker without sudo. May be something to discuss.

mattsinc (approved):

BobbyRBruce (approved):
Thanks. I'll rebuild the Docker images at some point this week and let you know when it's done

PR #2817, opened 2025-12-07
cpu: Branch predictor latency and overriding model for the decoupled frontend

shadowcpy:
This change enhances the realism of the branch predictor, providing a model for access latency to the conditional BP and an accompanying implementation of overriding prediction (https://www.cs.utexas.edu/~skeckler/pubs/micro00.pdf)

It builds on top of the decoupled frontend, and applies an amount of latency which can be statically configured but then also dynamically overridden by implementors of `conditional.cc` to allow more complex behaviour.

The decoupled frontend is stalled for the amount of cycles specified after each prediction.
Users can optionally specify a secondary overriding predictor, with a different latency.

By default, overriding is disabled and the latency is statically configured to 0 cycles.
To choose different settings, the following configuration can be used:

```python
from m5.objects import BranchPredictor, SimpleBTB, LocalBP, TAGE_SC_L_64KB 

class BPU(BranchPredictor):
    btb = SimpleBTB()
    conditionalBranchPred = LocalBP(
        defaultLatency = 1
    )
    overridingBranchPred = TAGE_SC_L_64KB(
        defaultLatency = 3
    )

cpu.branchPred = BPU()
```

In this case, `LocalBP` is overridden by `TAGE_SC_L_64KB`, producing the following behaviour:
- When a conditional branch is reached by the decoupled frontend, both predictors are queried simultaneously
- The results are compared:
    - If both predictors agree, 1 cycle of latency is applied to the `BAC` stage
    - On a disagreement, the result from `TAGE_SC_L_64KB` is used, and 3 cycles of latency are applied to the `BAC` stage

### DIFF
--- a/src/cpu/o3/bac.cc
+++ b/src/cpu/o3/bac.cc
@@ -113,6 +113,7 @@ BAC::BAC(CPU *_cpu, const BaseO3CPUParams &params)
     for (int i = 0; i < MaxThreads; i++) {
         bacPC[i].reset(params.isa[0]->newPCState());
         stalls[i] = {false, false, false};
+        branchPredictRemaining[i] = Cycles(0);
     }
 }
 
@@ -227,6 +228,9 @@ BAC::drainStall(ThreadID tid)
     assert(cpu->isDraining());
     assert(!stalls[tid].drain);
     DPRINTF(Drain, "%i: Thread drained.\n", tid);
+    DPRINTF(BAC, "%i: Clearing remaining %i cycles of predictor latency.\n",
+            tid, branchPredictRemaining[tid]);
+    branchPredictRemaining[tid] = Cycles(0);
     stalls[tid].drain = true;
 }
 
@@ -403,6 +407,16 @@ BAC::checkSignalsAndUpdate(ThreadID tid)
         return true;
     }
 
+    if (branchPredictRemaining[tid] > Cycles(0)) {
+        DPRINTF(BAC,
+                "[tid:%i] Stalling for Branch Predictor for %i more cycles.\n",
+                tid, branchPredictRemaining[tid]);
+        --branchPredictRemaining[tid];
+        stalls[tid].bpu = true;
+    } else {
+        stalls[tid].bpu = false;
+    }
+
     if (checkStall(tid)) {
         bacStatus[tid] = Blocked;
         return true;
@@ -572,7 +586,7 @@ BAC::newFetchTarget(ThreadID tid, const PCStateBase &start_pc)
     return ft;
 }
 
-bool
+Prediction
 BAC::predict(ThreadID tid, const StaticInstPtr &inst, const FetchTargetPtr &ft,
              PCStateBase &pc)
 {
@@ -586,10 +600,11 @@ BAC::predict(ThreadID tid, const StaticInstPtr &inst, const FetchTargetPtr &ft,
      * main history of the BPU and insert these missing histories.
      */
     assert(ft->bpuHistory == nullptr);
-    bool taken = bpu->predict(inst, ft->ftNum(), pc, tid, ft->bpuHistory);
+
+    Prediction pred = bpu->predict(inst, ft->ftNum(), pc, tid, ft->bpuHistory);
 
     DPRINTF(Branch, "[tid:%i, ftn:%llu] History added.\n", tid, ft->ftNum());
-    return taken;
+    return pred;
 }
 
 void
@@ -686,7 +701,9 @@ BAC::generateFetchTargets(ThreadID tid, bool &status_change)
 
             // Now make the actual prediction. Note the BPU will advance
             // the PC to the next instruction.
-            predict_taken = predict(tid, staticInst, curFT, *next_pc);
+            Prediction pred = predict(tid, staticInst, curFT, *next_pc);
+            predict_taken = pred.taken;
+            branchPredictRemaining[tid] = Cycles(pred.latency);
 
             DPRINTF(BAC,
                     "[tid:%i, ftn:%llu] Branch found at PC %#x "
@@ -881,8 +898,7 @@ BAC::updatePreDecode(ThreadID tid, const InstSeqNum seqNum,
 
         hist =
             new BPredUnit::PredictorHistory(tid, seqNum, pc.instAddr(), inst);
-        bpu->branchPlaceholder(tid, pc.instAddr(), inst->isUncondCtrl(),
-                               hist->bpHistory);
+        bpu->branchPlaceholder(tid, pc.instAddr(), inst->isUncondCtrl(), hist);
 
         set(hist->target, std::unique_ptr<PCStateBase>(pc.clone()));
         inst->advancePC(*hist->target);
@@ -933,8 +949,11 @@ BAC::updatePC(const DynInstPtr &inst, PCStateBase &fetch_pc,
         } else {
             // With a coupled front-end we need to make the branch prediction
             // here.
-            predict_taken =
+            //
+            // Latency is ignored in coupled mode
+            Prediction pred =
                 bpu->predict(inst->staticInst, inst->seqNum, fetch_pc, tid);
+            predict_taken = pred.taken;
         }
 
         DPRINTF(BAC,
--- a/src/cpu/o3/bac.hh
+++ b/src/cpu/o3/bac.hh
@@ -52,6 +52,7 @@ namespace gem5
 {
 
 struct BaseO3CPUParams;
+using branch_prediction::Prediction;
 
 namespace o3
 {
@@ -247,10 +248,10 @@ class BAC
      * @param inst The branch instruction.
      * @param ft The fetch target that is currently processed.
      * @param PC The predicted PC is passed back through this parameter.
-     * @return Returns if the branch is taken or not.
+     * @return Returns the prediction result from the BPU.
      */
-    bool predict(ThreadID tid, const StaticInstPtr &inst,
-                 const FetchTargetPtr &ft, PCStateBase &pc);
+    Prediction predict(ThreadID tid, const StaticInstPtr &inst,
+                       const FetchTargetPtr &ft, PCStateBase &pc);
 
     /**
      * Main function that feeds the FTQ with new fetch targets.
@@ -364,6 +365,9 @@ class BAC
      * cycle. Used to tell CPU if there is activity this cycle.
      */
     bool wroteToTimeBuffer;
+
+    /** Tracks remaining cycles that the branch predictor stalls BAC */
+    Cycles branchPredictRemaining[MaxThreads];
 
     /** Source of possible stalls. */
     struct Stalls
--- a/src/cpu/pred/2bit_local.cc
+++ b/src/cpu/pred/2bit_local.cc
@@ -92,9 +92,8 @@ LocalBP::updateHistories(ThreadID tid, Addr pc, bool uncond, bool taken,
 // Place holder for a function that is called to update predictor history
 }
 
-
-bool
-LocalBP::lookup(ThreadID tid, Addr branch_addr, void * &bp_history)
+Prediction
+LocalBP::lookup(ThreadID tid, Addr branch_addr, void *&bp_history)
 {
     bool taken;
     unsigned local_predictor_idx = getLocalIndex(branch_addr);
@@ -109,7 +108,7 @@ LocalBP::lookup(ThreadID tid, Addr branch_addr, void * &bp_history)
 
     taken = getPrediction(counter_val);
 
-    return taken;
+    return predictWithDefaultLatency(taken);
 }
 
 void
--- a/src/cpu/pred/BranchPredictor.py
+++ b/src/cpu/pred/BranchPredictor.py
@@ -145,7 +145,7 @@ class SimpleBTB(BranchTargetBuffer):
     )
 
 
-class ConditionalPredictor(SimObject):
+class ConditionalPredictor(ClockedObject):
     type = "ConditionalPredictor"
     cxx_class = "gem5::branch_prediction::ConditionalPredictor"
     cxx_header = "cpu/pred/conditional.hh"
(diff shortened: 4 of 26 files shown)


src/cpu/o3/bac.cc
```diff
@@ -469,6 +480,11 @@ BAC::squashBpuHistories(ThreadID tid)
 
     DPRINTF(BAC, "%s(tid:%i): FTQ sz: %i\n", __func__, tid, ftq->size(tid));
 
+    unsigned n_fts = ftq->size(tid);
```
dhschall: Can you remove that. Its has nothing to do with the latency and the `forAllBackward` will just immediatly return so it redundant.

src/cpu/o3/bac.cc
```diff
@@ -970,10 +995,48 @@ BAC::updatePC(const DynInstPtr &inst, PCStateBase &fetch_pc,
     return predict_taken;
 }
 
+void
```
dhschall: Please remove that. There is now a `state` stats vector that replaces the old individual counting
`stats.status[bacStatus[tid]]++;`

src/cpu/o3/bac.cc
```diff
@@ -970,10 +995,48 @@ BAC::updatePC(const DynInstPtr &inst, PCStateBase &fetch_pc,
...
+        default:
+            break;
+    }
+}
+
 BAC::BACStats::BACStats(o3::CPU *cpu, BAC *bac)
     : statistics::Group(cpu, "bac"),
       ADD_STAT(status, statistics::units::Cycle::get(),
                "Number of cycles BAC in state"),
+
```
dhschall: Same here

src/cpu/o3/bac.hh
```diff
@@ -91,6 +92,7 @@ typedef std::shared_ptr<FetchTarget> FetchTargetPtr;
 class BAC
 {
     typedef branch_prediction::BranchType BranchType;
+    typedef branch_prediction::BPredUnit BPredUnit;
```
dhschall: Still needed?

src/cpu/o3/bac.hh
```diff
@@ -184,6 +186,9 @@ class BAC
     /** Process all input signals and create the next fetch target. */
     void tick();
 
+    /** Register stats for the current cycle */
```
dhschall: Remove

src/cpu/o3/bac.hh
```diff
@@ -391,6 +399,9 @@ class BAC
     /** BAC to fetch delay. */
     const Cycles bacToFetchDelay;
 
+    /** BAC branch predict delay. */
```
dhschall: Still needed?

src/cpu/o3/bac.hh
```diff
@@ -432,6 +443,16 @@ class BAC
 
         /** Stat for total number of cycles spent in each BAC state */
         statistics::Vector status;
+        /** Stat for total number of idle cycles. */
```
dhschall: Remove

src/cpu/pred/bpred_unit.hh
```diff
@@ -181,24 +185,6 @@ class BPredUnit : public SimObject
         return btb->update(tid, pc, target);
     }
 
-    /**
```
dhschall, line 184: Can you keep the function here. I was just moved

src/cpu/pred/conditional.hh
```diff
@@ -46,22 +46,31 @@
...
     typedef ConditionalPredictorParams Params;
 
     ConditionalPredictor(const Params &params);
 
+    /**
+     * Returns the configured prediction latency in cycles
+     * @return The prediction latency in cycles
+     */
+    Cycles
+    getStaticLatency() const
```
dhschall: What does "static" latency mean? Can't we call it predictionLatency or predLatency?

src/cpu/pred/multiperspective_perceptron.hh
```diff
@@ -55,6 +55,7 @@
 #include <vector>
 
 #include "base/random.hh"
+#include "cpu/pred/branch_type.hh"
```
dhschall: The branch type is already included in the `conditional.hh` file. You don't need it everywhere as its rather part of the ConditionalPred base class

src/cpu/pred/conditional.hh
```diff
@@ -141,9 +150,19 @@ class ConditionalPredictor : public SimObject
 
     /** Number of bits to shift instructions by for predictor addresses. */
     const unsigned instShiftAmt;
+
+    /** Static latency of the predictor in cycles */
+    const Cycles staticLatency;
+
+    /** Return a prediction with only static latency */
+    Prediction
+    staticPrediction(bool taken) const
```
dhschall: Same here. I found static prediction a bit weird. How about `makePrediction` ?

dhschall (changes requested):
Just a few cosmetic comments for now. You can mark it as ready for review afterwards. This should also trigger the tests.

dhschall:
Thanks @shadowcpy for the contribution! One first, comment can you remove ITTAGE from the example? Unfortunately, this is not yet integrated.

src/cpu/o3/bac.hh
```diff
@@ -52,6 +52,7 @@ namespace gem5
 {
 
 struct BaseO3CPUParams;
+typedef branch_prediction::Prediction Prediction;
```
powerjg: I think the more C++ way to do this is `using branch_prediction::Prediction;`

src/cpu/pred/BranchPredictor.py
```diff
@@ -155,6 +155,9 @@ class ConditionalPredictor(SimObject):
     instShiftAmt = Param.Unsigned(
         Parent.instShiftAmt, "Number of bits to shift instructions by"
     )
+    latency = Param.Cycles(
+        0, "Static (flat) latency of the predictor (in cycles)"
```
powerjg: What does the default latency of 0 cycles mean? Does this mean no added latency or that the branch predictor finishes without any latency?

src/cpu/pred/conditional.hh
powerjg: Or `staticLatencyPrediction`? However, if it's going to be overridden by a dynamic latency, then this is a bad name too.

powerjg (commented):
Just a couple of quick comments. Generally, I'm supportive of enabling dynamic latency this way.

src/cpu/pred/BranchPredictor.py
shadowcpy: Currently the static and dynamic latencies are not additive. 
The configured (static) latency from the configuration file is made available to the (runtime) predictor, which can use it in any possible way (including ignoring it, or adding it to another value).

For all current predictors, this means that the default is to predict with no latency at all.

src/cpu/pred/conditional.hh
shadowcpy: Maybe `predictWithDefaultLatency` is a good name? 
Then the parameter inside the configuration could be renamed to `defaultLatency`, which can make it clearer that the predictor may use a calculated value instead, and also provide its own parameters for configuring this value.

powerjg: `predictWithDefaultLatency` sounds OK to me.

shadowcpy:
Sorry for the delay, I implemented the changes discussed before.

src/cpu/pred/bpred_unit.hh
```diff
@@ -429,22 +456,29 @@ class BPredUnit : public SimObject
...
 
     /** Statistics */
     struct BPredUnitStats : public statistics::Group
     {
         BPredUnitStats(BPredUnit *bp);
 
+        std::unordered_set<Addr> uniqueBranches;
+
+        void preDumpStats() override;
+
```
dhschall, line 482: I think you should also implement the reset function. In case you call stats resets it will not clear the uniqueBranches: https://github.com/gem5/gem5/blob/7a2b0e413d06c5ce7097104abef3b1d9eaabca91/src/base/stats/group.hh#L146

dhschall (changes requested):
Thanks @shadowcpy. Only one small comment. Apart from that everything looks all good!

src/cpu/o3/bac.cc
```diff
@@ -403,6 +404,16 @@ BAC::checkSignalsAndUpdate(ThreadID tid)
...
+    if (branchPredictRemaining[tid] > Cycles(0)) {
+        --branchPredictRemaining[tid];
+        DPRINTF(BAC,
+                "[global] Stalling for Branch Predictor for %i more cycles.\n",
+                branchPredictRemaining[tid]);
+        stalls[tid].bpu = true;
+    } else {
+        stalls[tid].bpu = false;
+    }
+
```
powerjg, line 419: It is possible to make this an event instead of decrementing each cycle?

src/cpu/pred/2bit_local.cc
powerjg, line 1: It seems like most of these changes are from clang-format and not from your changes. Clang-format shouldn't modify lines that you didn't modify.

src/cpu/pred/bi_mode.cc
powerjg, line 1: It seems like most of these changes are from clang-format and not from your changes. Clang-format shouldn't modify lines that you didn't modify.

src/cpu/pred/ltage.cc
powerjg, line 1: It seems like most of these changes are from clang-format and not from your changes. Clang-format shouldn't modify lines that you didn't modify.

src/cpu/pred/multiperspective_perceptron.cc
powerjg, line 1: It seems like most of these changes are from clang-format and not from your changes. Clang-format shouldn't modify lines that you didn't modify.

src/cpu/pred/multiperspective_perceptron.hh
powerjg, line 1: It seems like most of these changes are from clang-format and not from your changes. Clang-format shouldn't modify lines that you didn't modify.

src/cpu/pred/multiperspective_perceptron_tage.cc
powerjg, line 1: It seems like most of these changes are from clang-format and not from your changes. Clang-format shouldn't modify lines that you didn't modify.

src/cpu/pred/tage.cc
powerjg, line 1: It seems like most of these changes are from clang-format and not from your changes. Clang-format shouldn't modify lines that you didn't modify.

src/cpu/pred/tage_sc_l.cc
powerjg, line 1: It seems like most of these changes are from clang-format and not from your changes. Clang-format shouldn't modify lines that you didn't modify.

src/cpu/pred/tournament.cc
powerjg, line 1: It seems like most of these changes are from clang-format and not from your changes. Clang-format shouldn't modify lines that you didn't modify.

powerjg (commented):
Looks like there's a bunch of extra clang-format changes. If at all possible, it would be great to separate formatting changes from your changes. In the most part, there should just be 1 or 2 lines changes in each file.

src/cpu/o3/bac.cc
shadowcpy: From my understanding, this would not really fit into the current model of the `BAC` class.
As it is not a `SimObject` itself, and is directly driven by function calls of its parent `Fetch`, it cannot schedule events currently.
If `BAC` is made a `SimObject` (or I add a child object just for the latency reset which is a `SimObject`), the latency logic would no longer be in lockstep with the rest of the `BAC` pipeline. 
This could IMO lead to incorrect behavior, as the scheduled event ticks do not necessarily have to correspond to the cycles seen by `BAC`.
Does this seem roughly correct @dhschall?

powerjg: Maybe this is a dumb question, but how do you know that this function will be called once per cycle?

powerjg (dismissed):

src/cpu/o3/bac.cc
shadowcpy: As far as I understand, that is guaranteed to be the case due to `tick()` being called here: https://github.com/gem5/gem5/blob/7a2b0e413d06c5ce7097104abef3b1d9eaabca91/src/cpu/o3/cpu.cc#L380

Which then in turn calls `checkSignalsAndUpdate` for each thread here: https://github.com/gem5/gem5/blob/7a2b0e413d06c5ce7097104abef3b1d9eaabca91/src/cpu/o3/bac.cc#L516

This finally calls the latency logic here: 
https://github.com/dhschall/gem5-fdp/blob/8ff8b836a59e8022b8108763dca4b42c16d0d96a/src/cpu/o3/bac.cc#L407

However, in the case of a drain stall being already active, or during a squash operation, this would in fact not trigger, therefore the latency would not "advance" until the squash or drain stall has cleared.

This degrades the realism of the model a bit. Do you think this should be addressed before proceeding to merge?
Then I would look at ways to track the latency further up the chain, such that it always makes progress.

powerjg: Sounds good about the tick. Thanks for the clarification.

Good catch on the drain state.

Maybe to make sure things drain correctly you could add the following to `BAC::drainStall`

```c++
branchPredictRemaining[tid] = Cycles(0);
```

I'm not worried about the timing realism. I'm more concerned about making sure that everything is drained properly and we don't have a case that we get stuck in an infinite loop waiting for something to drain.

To test draining, you should be able to run for a small amount of time and take a checkpoint. The checkpoint should trigger a drain (or, actually, you could just call `m5.simulate.drain()` after warming up the system).

shadowcpy: I added the suggested line and tested the draining with the `gem5_library/fdp-hello.py` script, adding a drain every 1000 instructions while applying 2 cycles of latency, and it seems to work as expected:

```
10510146: board.processor.cores.core.bac: [tid:0, ftn:6628] Branch found at PC 0x41340f taken?:1, target:0x4134a0
10510146: board.processor.cores.core.bac: [tid:0] [fn:6628] 7 addresses searched. Branch found:1. Continue with PC:(0x4134a0=>0x4134a8).(0=>1) in next cycle
10510479: board.processor.cores.core.bac: 0: Stalling for Branch Predictor for 2 more cycles.
10510479: board.processor.cores.core.bac: [tid:0] BPU stall detected.
10510479: board.processor.cores.core.bac: 0: Clearing remaining 1 cycles of predictor latency.
10510812: board.processor.cores.core.bac: [tid:0] Drain stall detected.
10510812: board.processor.cores.core.bac: squashBpuHistories(tid:0): FTQ sz: 3
10511145: board.processor.cores.core.bac: [tid:0] Squashing from commit. PC = (0x451e6c=>0x451e74).(0=>1)
```

powerjg (approved):

dhschall (approved):

PR #2820, opened 2025-12-08
configs, test: Fix Neoverse SystemOp handling & mv test to CI with ALL/gem5 com

BobbyRBruce:
- Restore the Neoverse V2 complex integer FU’s ability to execute SystemOp instructions so MSR/MRS sequences (e.g., TLS setup) no longer deadlock SE workloads.
- Update the Neoverse example tests to run with the ALL/gem5.opt binary, matching how we teset other configs. This is more efficient at ALL/gem5.opt builds are cached. ARM/gem5.opt needs compiled from scratch each time.
- Move the test_neoverse_v2 suite into the quick CI workflow so the fixed configuration is exercised on every PR.


powerjg (approved):
lgtm, but it would be best if @dhschall could sign off before merging

dhschall (approved):
Yes perfect!
I didn't know that ALL is faster so thanks for the hint

PR #2822, opened 2025-12-09
cpu: BTB latency and multi-level BTB hierarchy model for the decoupled frontend

yongjiehuang:
This PR extends gem5's decoupled frontend infrastructure in branch prediction with:

1. A configurable BTB lookup latency, integrated with the decoupled frontend.
2. A configurable multi-level BTB implementation that models hierarchical BTB organizations seen in modern [Branch Target Buffer Organizations](https://hal.science/hal-04234792v1/file/MICRO23_CameraReady-1.pdf).
Many recent cores deploy hierarchical BTB designs (e.g., a small, low-latency first-level BTB on top of larger and slower BTBs), so zero-bubble branching is feasible when taken branches hit in the first-level BTB, but incur extra bubbles when deeper levels override the L1-BTB miss in serveral cycles. This patch adds enough structure to gem5's BTB model to capture that behavior while remaining configurable inclusiveness.

The PR builds on top of the branch predictor latency / overriding model introduced in [PR#2817](https://github.com/gem5/gem5/pull/2817).
Given that BTB lookup and direction prediction are conducted in parallel, the decoupled frontend is stalled for the maximum cycles of the branch predictor latency and BTB latency.

For backward compatibility, you can still choose to configure SimplelBTB but now its latency is configurable. For example:
```
from m5.objects import SimpleBTB
class BPU(BranchPredictor):
     btb = SimpleBTB(
          numEntries = 16 * 1024
          latency = 4
     )
     conditionalBranchPred = TAGE_SC_L_64KB(
         latency = 4
     )
cpu.branchPred = BPU()
```
Instead, you directly use MultiLevelBTB, which is a 3-level hierarchy by default:
```
from m5.objects import MultiLevelBTB
class BPU(BranchPredictor):
     btb = MultiLevelBTB()
     conditionalBranchPred = TAGE_SC_L_64KB(
         latency = 4
     )
cpu.branchPred = BPU()

```
Also, you can also configure any levels by inheriting MultiLevelBTB, a TwoLevelBTB is as below:
```
from m5.objects import MultiLevelBTB
from m5.objects import BTBLevel

class TwoLevelBTB(MultiLevelBTB):
    levels = [ BTBLevel(numEntries=256, associativity=8, latency=0, inclusive=True),
               BTBLevel(numEntries=16384, associativity=8, latency=3, inclusive=False) ]

class BPU(BranchPredictor):
     btb = TwoLevelBTB()
     conditionalBranchPred = TAGE_SC_L_64KB(
         latency = 4
     )
cpu.branchPred = BPU()

```


src/cpu/o3/bac.cc
```diff
@@ -782,6 +782,9 @@ BAC::generateFetchTargets(ThreadID tid, bool &status_change)
             status_change = true;
             break;
         }
+        // Stall the front-end if L2-BTB or branch predictor overrides.
+        if (branchPredictRemaining[tid] != 0)
```
dhschall: This change should moved to another PR that fixes the branchPredictLatnency.
I've added the change now in this PR: https://github.com/gem5/gem5/pull/3266
Including the other fix

dhschall (commented):
Thanks for the contribution.
Overall, it looks already quite good. However, you make it now a victim BTB. How about if you allow it to configure the allocation policy? It should be trivial to switch between inclusive (allocate L1 and L2 together) or victim mostly exclusive.

Also, why not make it a multi-level BTB right away?
Then you could have two Python templates that override the MultiLevelBTB for two and three levels.

src/cpu/pred/btb.cc
```diff
@@ -45,10 +45,11 @@ namespace gem5
 
 namespace branch_prediction
 {
-
+
```
dhschall:
```suggestion
```

src/cpu/pred/bpred_unit.cc
```diff
@@ -205,9 +206,20 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
...
     stats.BTBLookups++;
-    const PCStateBase * btb_target = btb->lookup(tid, pc.instAddr(), brType);
+    BTBLookupResult btb_res = btb->lookup(tid, pc.instAddr(), brType);
+    const PCStateBase * btb_target = btb_res.target;
+    // The BTB and CBP are accessed in parallel.
+    // The total latency should be the maximum of the bpred latency and the BTB latency.
+    totalLatency = std::max(totalLatency, btb_res.latency);
     if (btb_target) {
         stats.BTBHits++;
+        if (isTwoLevelBTB) {
```
dhschall: I would actually prefer to have none of the BTB multi-level statistics in the BPU.
That's why we have the statistics in the BTB itself and should do the accounting there. 
It just adds complexity and creates dependencies and most people will likely never use the multi-level BTB because they use the default one

src/cpu/pred/bpred_unit.cc
```diff
@@ -69,6 +69,7 @@ BPredUnit::BPredUnit(const Params &params)
       iPred(params.indirectBranchPred),
       stats(this)
 {
+    isTwoLevelBTB = (dynamic_cast<const TwoLevelBTB*>(btb) != nullptr);
```
dhschall: See below

src/cpu/o3/bac.cc
yongjiehuang: Thanks for the review!
All are great ideas!  In response:
1:  I refactored the code to make it support multi-level BTB with the added the "threeLevel" parameter.
2: The clusivity is now configurable: if the "inclusive" parameter is set, lower-level BTB is mostly-inclusive(no back-invalidation) of upper-level BTB, otherwise lower-level BTB acts as a victim buffer of the upper-level. 
3: I removed the bug fix to make this PR independent of PR: #3266.

src/cpu/pred/bpred_unit.cc
yongjiehuang: Thanks for pointing this out - I removed the related code in bpred_unit.cc.

src/cpu/pred/simple_btb.cc
```diff
@@ -51,7 +51,7 @@ SimpleBTB::SimpleBTB(const SimpleBTBParams &p)
     : BranchTargetBuffer(p),
       btb("simpleBTB", p.numEntries, p.associativity,
           p.btbReplPolicy, p.btbIndexingPolicy,
-          BTBEntry(genTagExtractor(p.btbIndexingPolicy)))
+          BTBEntry(genTagExtractor(p.btbIndexingPolicy))), latency(p.latency)
```
dhschall:
```suggestion
          BTBEntry(genTagExtractor(p.btbIndexingPolicy)))
```
The latency paramenter belongs to the `BranchTargetBuffer` class not to the `SimpleBTB`

src/cpu/pred/simple_btb.hh
```diff
@@ -77,6 +77,9 @@ class SimpleBTB : public BranchTargetBuffer
 
     /** The actual BTB. */
     AssociativeCache<BTBEntry> btb;
+
```
dhschall: I don't think this should be here. The latency is already in the parent class.
clang might even fail because it is more conservative with shadowing of member variables

src/cpu/pred/multi_level_btb.hh
```diff
@@ -0,0 +1,58 @@
...
+class MultiLevelBTB : public BranchTargetBuffer
+{
+  public:
+    MultiLevelBTB(const MultiLevelBTBParams &params);
+
+    void memInvalidate() override;
+    bool valid(ThreadID tid, Addr instPC) override;
+
+    const BTBLookupResult lookup(ThreadID tid, Addr instPC,
+                                      BranchType type = BranchType::NoBranch) override;
```
dhschall:
```suggestion
                                  BranchType type = BranchType::NoBranch) override;
```
The autoformating should allign this AFAIK

src/cpu/pred/multi_level_btb.hh
```diff
@@ -0,0 +1,58 @@
...
+
+    void update(ThreadID tid, Addr instPC, const PCStateBase &target_pc,
+                BranchType type = BranchType::NoBranch,
+                StaticInstPtr inst = nullptr) override;
+
+    const StaticInstPtr getInst(ThreadID tid, Addr instPC) override;
+
+
+  private:
+    BTBEntry *handleEviction(ThreadID tid, Addr instPC, AssociativeCache<BTBEntry> &upper_btb,
```
dhschall: Does that fit within the column limit?

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,240 @@
...
+            }
+            return BTBLookupResult(l1_victim->target.get(), l3Latency);
+        }
+    }
+    // Miss all levels
+    stats.misses[type]++;
+    DPRINTF(BTB, "BTB miss for PC %#x\n", instPC);
+
+    // Execution will never get here, otherwise there is bug.
+    panic("bpu->BTBValid(tid, br_addr) from bac.cc should have filtered btb miss");
```
dhschall: The BTB can be used with another core without the decoupled front-end.
You might not simulate the latency in this case, but you can still understand how many L1, L2, and L3 misses you have.
So this panic can be found

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,240 @@
...
+                l3_entry = l3btb.findVictim({instPC, tid});
+                l3btb.insertEntry({instPC, tid}, l3_entry);
+            }
+            l3_entry->update(target, inst);
+        }
+    }
+}
+
+BTBEntry*
+MultiLevelBTB::handleEviction(ThreadID tid, Addr instPC, AssociativeCache<BTBEntry> &upper_btb, AssociativeCache<BTBEntry> &lower_btb, bool lowerIsL3)
```
dhschall: This all will certainly increase the line limit (IIRC 72)
Please run the format checker

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,240 @@
...
+{
+    // Get victim from the upper-level BTB
+    BTBEntry *upper_victim = upper_btb.findVictim({instPC, tid}, false);
```
dhschall: The `false` value will prevent the entry from being invalidated.
Is that a problem for the replacement policy? if you never explicity call invalidate?
The insert will reset the entry and replacement policy but is that the same?

dhschall (changes requested):
Overall, looks ok. A few changes and formatting issues, but the logic looks reasonable.

src/cpu/pred/multi_level_btb.cc
yongjiehuang: This should not influence the replacement, but I found it will cause the assertion in setValid() fails in btb_entry.hh.
Thanks for the catch and I added the manual invalidation.

src/cpu/pred/multi_level_btb.cc
yongjiehuang: Ah, I totally forgot that decoupled front-end is not the only choice. Thanks and it's fixed now.

src/cpu/pred/simple_btb.hh
yongjiehuang: Thanks. I enabled the checker and fixed them all.

src/cpu/pred/multi_level_btb.cc
yongjiehuang: Fixed.

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,244 @@
...
+
+    // L1miss, lookup l2 btb
+    BTBEntry *l2_entry = l2btb.accessEntry({instPC, tid});
+    if (l2_entry != nullptr) {
+        BTBEntry *l1_victim = handleEviction(tid, instPC, l1btb, l2btb);
+        l1_victim->update(*l2_entry->target, l2_entry->inst);
+        multilevelstats.l2Hits++;
+        DPRINTF(BTB, "L2 BTB hit for PC %#x, latency=%d cycles, insert in L1\n", instPC, l2Latency);
+        return BTBLookupResult(l1_victim->target.get(), l2Latency);
+    }
```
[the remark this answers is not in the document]
yongjiehuang: Good catch. This policy indeed wastes overall capacity. However, as discussed in  [Two Level Bulk Preload Branch Prediction](https://class.ece.iastate.edu/tyagi/cpre581/papers/HPCA13BulkPreloadBranch.pdf) (sec. 3.3), maintaining a truly exclusive victim buffer is costly, because it requires two write operations upon L2 hit if the L1-victim and the L2 hit map to different ways in L2-BTB.
So at this point, I'd prefer to implement 'semi-exclusive' as the IBM paper does - demote the hit slot to LRU-like  position instead of invalidating it right away. Please check my lastest commit to see if you are OK on it.

dhschall (commented):
Thanks for making the changes. Looks good to me now.
~@powerjg do you know why the formater is failing for an unmodified file (riscv/procees.c)?~
Ok got it. Was addressed by #3260. Now it works

To me, it looks good. @giactra you may want to have a look as well, especially regarding change in the AssociativeCache template if thats all right

dhschall (dismissed):

giactra (commented):
Thanks for this PR! Finally we will have multilevel-BTBs in gem5! I have a major comment about the current design. I believe we can make things more modular/configurable (see below). Let me know what you think of it.

dhschall:
@giactra I cannot see your comment/suggestion?

src/cpu/pred/BranchPredictor.py
```diff
@@ -145,6 +146,80 @@ class SimpleBTB(BranchTargetBuffer):
...
+        LRURP(), "L1 BTB replacement policy"
+    )
+    l1IndexingPolicy = Param.BTBIndexingPolicy(
+        BTBSetAssociative(
+            assoc=Parent.l1Associativity,
+            num_entries=Parent.l1NumEntries,
+            set_shift=Parent.instShiftAmt,
+            numThreads=1,
+        ),
+        "L1-BTB indexing policy",
```
giactra: My feeling is that we can make things more modular-cleaner by adding another SimObject. Something like
BTBTable/BTBLevel, and then make the number of levels of BTBs completely configurable/unbounded, instead of forcing a user to configure or 3 level of parameters. I am talking about something like:

```
class MultiLevelBTB(BranchTargetBuffer):
    type = "MultiLevelBTB"
    cxx_class = "gem5::branch_prediction::MultiLevelBTB"
    cxx_header = "cpu/pred/multi_level_btb.hh"

    tables = VectorParam.BTBLevel([], "Vector of branch target buffers")

[...]

class BTBLevel(SimObject):
    numEntries = Param.Unsigned(4096, "Number of BTB entries")
    associativity = Param.Unsigned(8, "BTB associativity")
    latency = Param.Cycles(1, "BTB override latency in cycles")
    replPolicy = Param.BaseReplacementPolicy(
        LRURP(), "BTB replacement policy"
    )
    indexingPolicy = Param.BTBIndexingPolicy(
[...]
        ),
        "BTB indexing policy",
    )
```

yongjiehuang: That's a great idea - thanks! I push the refacor of the code to make it more configurable. Now also the inclusive parameter works for each BTB level.  Let me know any concerns so that I can iterate on it. 
Basically, a new entry tackled by update() will try to insert into all levels which are inclusive . And a evicted victim from a level will writeback only when its immediately lower neighbor is semi-exclusive (inclusive == false).

src/cpu/pred/multi_level_btb.hh
```diff
@@ -0,0 +1,62 @@
...
+namespace gem5::branch_prediction
+{
+
+class BTBLevel : public SimObject
+{
+  public:
+    BTBLevel(const BTBLevelParams &params);
+
+  private:
+    friend class MultiLevelBTB;
```
dhschall, line 51: Why is the friend needed? Is it to check the `inclusive` ?

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,223 @@
...
+{
+    for (auto *level : levels) {
+        level->btb.clear();
+    }
+}
+
+bool
+MultiLevelBTB::valid(ThreadID tid, Addr instPC)
+{
+    unsigned level_num = 1;
```
dhschall, line 168: Is the lowest level 0 or 1?
Right now it seems inconsistent.

Above the lowest level is 0. Here it will print "L1 BTB valid..."

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,223 @@
...
+      multilevelstats(this, levels.size())
+{
+    fatal_if(levels.empty(), "%s must contain at least one BTB level", name());
+    DPRINTF(BTB, "%s: creating a %d-level BTB hierarchy\n", name(),
+            levels.size());
+
+    unsigned level_num = 1;
+    for (auto *level : levels) {
+        fatal_if(!level, "%s: BTB level %d is null", name(), level_num);
+        multilevelstats.levelHits.subname(level_num - 1,
```
dhschall: Is the lowest level 0 or 1?
Right now it seems inconsistent

src/cpu/pred/BranchPredictor.py
adhilton: A colleague called my attention to this discussion since I have a multi-level BTB implementation in my local workspace.  Mine is designed as a variant of the Decorator pattern: i.e., a multi-level BTB is-A BTB and has-many BTBs.     It keeps the same interface as the original BTB.   This means that you configure it by giving it a list of its BTBs and their respective latencies, and it supports any number of levels/combination of BTBs, e.g., 

```python
class MyParticularBTB(MultiLevelBTB):
  """A 3-Level Branch Target Buffer Hierarchy."""

  btbs = [
      SimpleBTB(
          numEntries=20,
          associativity=4,
          tagBits=16,
          instShiftAmt=2,
          btbReplPolicy=LRURP(),
          btbIndexingPolicy=BTBSetAssociative(
              num_entries=16,
              assoc=4,
              set_shift=2,
              tag_bits=20,
          ),
      ),
      SimpleBTB(
          numEntries=512,
          associativity=4,
          tagBits=20,
          instShiftAmt=2,
          btbReplPolicy=LRURP(),
          btbIndexingPolicy=BTBSetAssociative(
              num_entries=512,
              assoc=4,
              set_shift=2,
              tag_bits=20,
          ),
      ),
      SimpleBTB(
          numEntries=16 * 1024,
          associativity=8,
          tagBits=20,
          instShiftAmt=2,
          btbReplPolicy=LRURP(),
          btbIndexingPolicy=BTBSetAssociative(
              num_entries=16 * 1024,
              assoc=8,
              set_shift=2,
              tag_bits=20,
          ),
      ),
  ]
  hitLatencies = [1, 2, 4]
```

dhschall (commented):
Looks actually quite good to me.
Just a minor question/comment.

Also, have you tested it with different hierarchies? E.g. 2,3 and 4 levels?

dhschall:
@giactra can you also have a look if that's what you had in mind?

src/cpu/pred/BranchPredictor.py
yongjiehuang: Thanks @adhilton for sharing this! I also noticed that the BTBLevel I added (see my latest push) is quite similar to the existing simpleBTB in BranchPredictor.py, except for the inclusive/semi-exclusive parameter.
However, to me, it's mainly the configurable inclusive property that makes it necessary to modify the simpleBTB implementation, which I am not sure if @dhschall @giactra would think it's acceptable or have other comments. 
The current simpleBTB interface only exposes valid, lookup, update, getInst. Particularly, SimpleBTB::update() overwrites its selected victim without returning it. Therefore, a wrapper containing std::vector<SimpleBTB *> might be not sufficient to implement the cascading victim movement required by a hierarchical BTB where there can be semi-exclusive BTB level.

src/cpu/pred/multi_level_btb.hh
yongjiehuang: Will test it more thoroughly with heavy workloads tomorrow. In case there are any major issues which needs modification of certain logics or refactor so I pushed first for discussion.

I just wanted to explicit note only MultiLevelBTB uses this class. I can also set everything to public if that's a more preferable style to the upstream.

src/cpu/pred/multi_level_btb.cc
yongjiehuang: The statistics vector is zero-indexed, so I use level_num - 1 here.

src/cpu/pred/multi_level_btb.cc
yongjiehuang: For the display purpose, I wanted the lowest level to be L1.

src/cpu/pred/BranchPredictor.py
adhilton: To me, it is nice to have extensibility that doesn't lock someone into just multiple simple BTBs (with duplication of that code as BTBLevels).  What if someone has a FancyBTB and a SuperFancyBTB  [the specifics don't matter for this discussion, but assume they do something other than just increased capacity], and wants L0 to be a SimpleBTB, L1 to be a FancyBTB and L2 to be a SuperFancyBTB?

src/cpu/pred/btb_entry.hh
```diff
@@ -266,6 +280,9 @@ class BTBEntry : public ReplaceableEntry
 
     /** The entry's tag. */
     KeyType tag;
+
+    /** The entry's branch address. */
+    Addr branchAddr;
```
giactra: This is fine but I believe calling this branchPC will make it clearer we are not talking about the target address of the branch

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,223 @@
...
+                entry = handleEviction(tid, instPC, level);
+            }
+            entry->update(target, inst);
+        }
+        first_level = false;
+    }
+}
+
+BTBEntry *
+MultiLevelBTB::handleEviction(ThreadID tid, Addr instPC,
```
giactra: I find all the logic in handleEviction (but I would say in lookup as well) to be overly-complicated and a bit hard to follow.

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,223 @@
...
+    victim_entry->invalidate();
+    insertion_btb.insertEntry({instPC, tid}, victim_entry);
+
+    bool reached_lower_level = false;
+    for (auto *level : levels) {
+        // Possible writeback only happens at lower levels.
+        if (!reached_lower_level) {
+            reached_lower_level = level == insertionLevel;
+            continue;
+        }
```
giactra: Range-based for loops are great, but not if we are skipping initial levels. At that point you might as well work with good old indexes. Something like

```
for (auto level = start_level; level < levels.size(); level++) {
}
```

There's nothing wrong with that

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,223 @@
...
+                        allocated_entry->update(*hit_backup.target,
+                                                hit_backup.inst);
+
+                        if (first_level) {
+                            result = allocated_entry;
+                        }
+                    }
+                    first_level = false;
+                }
+            }
```
giactra: The lookup logic should be pretty intuitive but I find this one a bit hard to follow.
IMHO this code can be simplified if we moved some of the logic within the BTBLevel, similarly to what we do for example for the ArmTLB (Check how we handle lookup/insertion/promotion over there and think if that approach is applicable here)

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,223 @@
...
+                bool first_level = true;
+                // Insert the hit entry into the L1 BTB and other
+                // upper levels which are inclusive.
+                for (auto *promotion_level : levels) {
+                    if (promotion_level == lookup_level) {
+                        break;
+                    }
+                    if (first_level || promotion_level->inclusive) {
+                        auto *allocated_entry =
+                            handleEviction(tid, instPC, promotion_level);
```
giactra: I know what this is doing, but I find the handleEviction name a bit confusing. Shouldn't something like "iinsertEntry" be more meaningful?

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,223 @@
...
+                    lookup_level->btb.demoteEntry(hit_entry);
+                }
+
+                bool first_level = true;
+                // Insert the hit entry into the L1 BTB and other
+                // upper levels which are inclusive.
+                for (auto *promotion_level : levels) {
+                    if (promotion_level == lookup_level) {
+                        break;
+                    }
```
giactra: This is another case where using indexes is much better:

```
for (auto level = 1; level < level_num; level++) {
}
```

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,223 @@
...
+                    if (first_level || promotion_level->inclusive) {
```
giactra: You really don't need to check for first_level here if you FORCE (fatal at construction) the first_level BTB to be inclusive (even if it is meaningless)

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,223 @@
...
+
+void
+MultiLevelBTB::update(ThreadID tid, Addr instPC, const PCStateBase &target,
+                      BranchType type, StaticInstPtr inst)
+{
+    stats.updates[type]++;
+
+    bool first_level = true;
+    for (auto *level : levels) {
+        if (first_level || level->inclusive) {
```
giactra: No need explicit first level handling if we force L1 BTB to be inclusive

yongjiehuang:
@giactra Thanks for the detailed review! My latest push refined the code based on your comment/pointer.
Now the lookup looks like the style of ArmTLB, and the logic summary: 

- Each BTBLevel has pointer to its immediate lower BTBLevel.  MultiLevelBTB starts the lookup with L1 calling the multiLookup(). 

- If a level is not inclusive, the hit slot will be invalidated. (exclusive)

- Hit from lower btb levels will refill into all inclusive upper levels via calling insertEntry on recursive multiLookup return. 

- InsertEntry on a level will writeback the victim to immediately lower level if not inclusive. 

-L1 is forced to set inclusive via fatal_if, and the update function insert a new BTB entry into all inclusive levels while the semi-exclusive level can only get entry via acommodating the victims from upper levels.

src/cpu/pred/BranchPredictor.py
yongjiehuang: Yes, that is a nice idea- Thanks! Now the BTBLevel class doesn't inherit the BranchTargetBuffer class, and thus not quite as flexible as you pointed out. But I think maybe someone can create a FancyBTB class inheriting BTBLevel to do whatever they want, otherwise it's a bit intrusive to modify existing simpleBTB related implementation merely for implementing the inclusive/semi-exclusive of levels in multilevelBTB of this PR.

src/base/cache/associative_cache.hh
```diff
@@ -233,8 +233,8 @@ class AssociativeCache : public Named
      * @param key key to select the possible victim
      * @result entry to be victimized
      */
-    virtual Entry*
-    findVictim(const KeyType &key)
+    virtual Entry *
+    findVictim(const KeyType &key, bool do_invalidate = true)
```
powerjg: Apologies if this has been previous answered, but why do we need to change this API? Why can't you call `findVictim` followed by `invalidate` in the branch predictor code? 

An explanation for the change in the API would be good. Or, if it's not needed, reverting the API change would work too.

src/base/cache/associative_cache.hh
```diff
@@ -263,6 +265,17 @@ class AssociativeCache : public Named
...
 
+    /**
+     * Demote replacement data of an entry without invalidating the entry.
+     *
+     * @param entry Entry to be demoted for replacement.
+     */
+    virtual void
+    demoteEntry(Entry *entry)
+    {
+        replPolicy->invalidate(entry->replacementData);
```
powerjg: I don't understand the use case for this API. Why is it safe to invalidate the replacement data but not the entry itself. Wouldn't this result in the entry and the replacement data being incoherent?

src/cpu/pred/BranchPredictor.py
```diff
@@ -145,6 +146,80 @@ class SimpleBTB(BranchTargetBuffer):
...
+    cxx_class = "gem5::branch_prediction::MultiLevelBTB"
+    cxx_header = "cpu/pred/multi_level_btb.hh"
+
+    levels = VectorParam.BTBLevel(
+        [],
+        "BTB levels ordered from the uppermost to the lowermost level",
+    )
+
+
+class TwoLevelBTB(MultiLevelBTB):
```
powerjg: I would prefer (and this is just a preference) that all concrete implementations of a SimObject are in the stdlib (src/python/gem5) not in the SimObject Description File. Since this (and three level) don't have a `type`/`cxx_class`/etc. and don't have C++ objects associated with them, they shouldn't be in these python files. 

That said, this is an aspiration since there is a lot of code that goes against this. It is also not compatible with the deprecated Options.py and se/fs.py if that is a requirement.

src/base/cache/associative_cache.hh
yongjiehuang: The reason is that a valid victim in a level need to be written back to the immediately lower level who is semi-exclusive. If a victim slot was originally invalid, there is no need to writeback the entry. However, the original findVictim API anyway invalidate a slot, which makes it unable to distinguish if the victim slot was valid or not, and hard to decide whether it needs writeback.

src/base/cache/associative_cache.hh
yongjiehuang: Thanks for the catch! Well, it indeed cause incoherent valid status between entry payload and its replacement data, except for LRURP which I use to implement this PR. 
For LRURP's code, there is not a specific boolean to flag valid. After calling replPolicy->invalidate, a slot becomes the priority to be evicted. If it got accessed again before eviction, LRURP's touch() will correclty maintain it by updating its lastTouchTick to curTick().
In this PR, saying an entry in the semi-exclusive L2-BTB is hit, then it refills into L1-BTB with its slot in L2 demoted by calling replPolicy->invalidate. If the L2 slot survives until the refilled entry gets evicted from L1 and thus written back into L2 again, the calling chain doWriteck() -> insertEntry() -> below, calls accessEntry -> LRU::touch() and promote the slot. 
> 
TBLevel::insertEntry(...)
{
     BTBEntry *entry = btb.findEntry({instPC, tid});    // saying btb here is L2
    if (entry) {
        btb.accessEntry(entry);             
        
However, for others like RandomRP,BRRIP, they have specific boolean for valid, demoteEntry() -> replPolicy->invalidate() causes valid set false, but their touch() do not set valid back to true. Thus, in the scenario above, even btb.accessEntry calls RP::touch(), this slot is still the next to be evicted due to its invalid replament data.

Seems there are two choices:

1.  Modify touch() for every replacement policy, set valid once touch() is called.
2. Specifically implement a "demote with valid set" API of all replacement policy.

src/cpu/pred/BranchPredictor.py
yongjiehuang: Thanks for the comment. How about I create a .py in src/python/gem5/components/processors/ for only containing the TwoLevelBTB/ThreeLevelBTB ? 
Or deleting those two examples class but only reamin the MultilevelBTB class in the BranchPredictor.py. Let use to choose what they want to configurate.

src/base/cache/associative_cache.hh
powerjg: Apologies, I still don't understand.

> The reason is that a valid victim in a level need to be written back to the immediately lower level who is semi-exclusive.

The normal caches (e.g., `src/mem/cache`) use this associative structure and support "semi-exclusive" protocols. I still don't see why this change is necessary.

> makes it unable to distinguish if the victim slot was valid or not

I understand the need for this, but I must be missing something in the function `findVictim(const KeyType &key, bool do_invalidate = true)`. You are optionally invalidating the victim, not returning if the victim is valid or invalid. It seems to me (again, I could be missing something) that you could have code like

```cpp
Entry *entry = findVictim(key);
if (entry.valid()) {
  entry.invalidate();
}
```

Why does the "if the entry is valid, you should invalidate" code go *inside* the **FIND**entry function? That function is about finding entries, not managing them.

src/base/cache/associative_cache.hh
powerjg: Both of these options are non-ideal.

It seems to me that there may be a better use of the `AssociativeCache` API in the BTB more generally. The generic `AssociativeCache` should be able to support any replacement policy and any insertion/demotion policy without fundamental changes to it. This generic code is used throughout the codebase for multilevel structures. 

This statement may be a *little* too strong. It's understandable if there is some new kind of feature that isn't needed for data caches that are needed for new kinds of caches like BTBs or TLBs, but my gut tells me that we aren't going to need to make fundamental changes like these proposed here.

src/cpu/pred/BranchPredictor.py
powerjg: That sounds good to me!

src/base/cache/associative_cache.hh
dhschall: The problem is that `findVictim` unconditionally invalidates the entry. So `entry.valid()` will always be false.

That said I agree with @powerjg that I also don't like these API changes. Especially the new `demoteEntry()` call.
@yongjiehuang can't we just unconditionally move the entry to the lower level. E.g., don't check if the entry is valid or not. In practice, I would assume the BTB has no valid bit anyway. It is a predictive structure and the valid bit is just waste of storage.
Will it crash? It might be because of the target pointer being `nullptr`.
Another option is to initialize the `branchAddr` field to 0 or MaxAddr, and then at eviction you check whether branchAddress has a valid address. Its a slight hack, but not IMO still better than the API changes.

src/cpu/pred/multi_level_btb.cc
powerjg, line 1: Needs copyright

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,211 @@
...
+        btb.accessEntry(entry);
+    } else {
+        entry = btb.findVictim({instPC, tid}, false);
+
+        std::optional<BTBEntry> victim;
+        if (entry->isValid()) {
+            victim = *entry;
+        }
+
+        entry->invalidate();
```
powerjg: See, you unconditionally invalidate just like the original `findVictim` code. The only different is if it *was* valid before you found a victim you save the data in the entry. I think this means we should return a `std::optional<Entry*>` exactly as you're using it here.

src/base/cache/associative_cache.hh
powerjg: Ah, I see my confusion. 

> The problem is that findVictim unconditionally invalidates the entry. So entry.valid() will always be false.

But I think this is correct. The **ENTRY** in this particular structure is now invalid. The `Entry` represents the SRAM in a particular associate structure. If you want to move the data in the entry to another structure you can do that.

Maybe a better API would be `std::optional<Entry*>` as the return value. Then, in `findVictim` if the entry was valid, we return the entry. Otherwise, we don't return anything. I would be for this API change.

@pranith was the original designer of this API and @giactra and @odanrc had a lot of comments when it was added (https://github.com/gem5/gem5/pull/745). I wonder if they have any inputs on how to deal with using this API in a multi-level structure.

@yongjiehuang sorry that this is going a bit in circles. I want to make sure that we get the `AssociativeCache` API right as it is a good interface that I hope to see used more throughout the codebase.

pranith: @powerjg We need to have std::optional\<Entry\> (instead of Entry*) as the return value so as to copy the entry that is going to be invalidated.

I am neutral on this API change. Its either this, or returning optional entry or a new API like (findValidVictim()) that does not invalidate the victim.

yongjiehuang: @powerjg @pranith @dhschall Thanks a lot for the suggestions!  Changing the return type of the existing findVictim API involves lots of changes in terms of a various of callers of this API. I am not sure if it's proper to do so in this multi-level BTB PR.  Thus, I'd personally favor a new API findValidVictim here.  But I am still open to it.

src/cpu/pred/multi_level_btb.cc
yongjiehuang: Added.

src/cpu/pred/BranchPredictor.py
yongjiehuang: Now I just removed the two concrete implementation, instead leaving MultiLevelBTB with default levels like ThreeLevelBTB. My intention is let user to override levels themself, or use MultiLevelBTB as ThreeLevelBTB by default.

src/base/cache/associative_cache.hh
yongjiehuang: My latest push remove the demoteEntry API, as well as the semi-exclusive semantics. Instead, as a general implementation,  just calling existing invalidate API to make a level acting more as a victim buffer if inclusive is false.

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,249 @@
...
+
+void
+BTBLevel::doWriteback(ThreadID tid, const BTBEntry &upper_victim)
+{
+    // An inclusive level already contains entries inserted in its upper
+    // level. Only a victim-buffer level needs to accept the writeback.
+    if (inclusive) {
+        return;
+    }
+
```
dhschall, line 118: @yongjiehuang can we simply add here a return if the target address is nullptr?

src/base/cache/associative_cache.hh
dhschall: I prefer not changing the API at all in this PR because I think it would compromise the simplicity of this interface.  
And I think all can be done within the BTB itself. The only problem is the nullptr for uninitialized targets at the beginning of the simulation. 
We can simply check if its null and skip the move to lower level.

If we ever need the optional return for invalid data it should be done in another PR

src/cpu/pred/multi_level_btb.cc
yongjiehuang: It's a great idea. Yes it should work. For a slot in BTB, only if it hasn't accommodated an entry (cold), it's invalid - target address is nullptr.  Thanks for the advice. Now there is no change on the cache API.

src/cpu/pred/multi_level_btb.hh
dhschall: No friend is better than everything public

src/cpu/pred/btb_entry.hh
```diff
@@ -193,6 +198,7 @@ class BTBEntry : public ReplaceableEntry
         tag        = other.tag;
         inst       = other.inst;
         extractTag = other.extractTag;
+        branchPC = other.branchPC;
```
dhschall, line 201:
```suggestion
        branchPC   = other.branchPC;
```

src/cpu/pred/btb_entry.hh
```diff
@@ -203,6 +209,7 @@ class BTBEntry : public ReplaceableEntry
...
```
dhschall, line 212:
```suggestion
        branchPC   = other.branchPC;
```

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,241 @@
...
+    entry->update(target, inst);
+    return entry;
+}
+
+void
+BTBLevel::doWriteback(ThreadID tid, const BTBEntry &upper_victim)
+{
+    // An inclusive level already contains entries inserted in its upper
+    // level. Only a victim-buffer level needs to accept the writeback.
+    if (inclusive || !upper_victim.target) {
```
dhschall, line 115: Can you also add a comment why we check for the target. Maybe something like this
```suggestion
    // Also, ignore writeback if the target address is null. I.e., invalid entry.
    if (inclusive || !upper_victim.target) {
```

dhschall (commented):
Ok this looks now better. Just some minor things. Otherwise I'm ok.
@powerjg what do you think now given Jongjie removed the API changes all together?

dhschall (dismissed):
Looks good to me now

dhschall (dismissed):

src/cpu/pred/BranchPredictor.py
```diff
@@ -145,6 +146,60 @@ class SimpleBTB(BranchTargetBuffer):
...
+    cxx_header = "cpu/pred/multi_level_btb.hh"
+
+    tagBits = Param.Unsigned(64, "Size of the BTB tags, in bits")
+    numEntries = Param.Unsigned(4096, "Number of BTB entries")
+    associativity = Param.Unsigned(8, "BTB associativity")
+    latency = Param.Cycles(1, "BTB access latency in cycles")
+    inclusive = Param.Bool(
+        False,
+        "Whether this level is inclusive of the immediately upper level; "
+        "otherwise, this level is a victim buffer of the upper levels ",
```
powerjg: "this level is a victim buffer" -> "this level is an **exclusive** victim buffer"

It's important to make it clear that if not inclusive means strictly exclusive since you invalidate the entry on lookup.

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,242 @@
...
+    : BranchTargetBuffer(p),
+      levels(p.levels),
+      multilevelstats(this, levels.size())
+{
+    fatal_if(levels.empty(), "%s must contain at least one BTB level", name());
+    DPRINTF(BTB, "%s: creating a %d-level BTB hierarchy\n", name(),
+            levels.size());
+
+    unsigned level_num = 1;
+    for (auto *level : levels) {
```
powerjg: I am pretty sure this code should be in the `MultiLevelBTBStats` constructor. Maybe a better way would be to just pass the `levels` instance to the stats constructor.

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,242 @@
...
+
+void
+MultiLevelBTB::update(ThreadID tid, Addr instPC, const PCStateBase &target,
+                      BranchType type, StaticInstPtr inst)
+{
+    stats.updates[type]++;
+
+    for (unsigned level = 0; level < levels.size(); ++level) {
+        auto *update_level = levels[level];
+        if (update_level->inclusive) {
```
powerjg, line 223: Just to double check my understanding: This does not update the exclusive (non L1 BTB) because the L1 BTB is always inclusive and so we only need to update the L1 BTB entry (and others if other levels are inclusive).

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,242 @@
...
+    for (auto *level : levels) {
+        fatal_if(!level, "%s: BTB level %d is null", name(), level_num);
+        multilevelstats.levelHits.subname(level_num - 1,
+                                          csprintf("L%d", level_num));
+        ++level_num;
+    }
+
+    for (unsigned level = 0; level < levels.size(); ++level) {
+        levels[level]->level = level;
+        if (level + 1 < levels.size()) {
```
powerjg, line 148: I don't think it's checked anywhere, but I believe it's also required that if level N is inclusive than all levels < N must be inclusive as well. 

This makes logical sense, but I believe right now a users could set L1 inclusive, L2 not inclusive (exclusive), and L3 inclusive, which would break the logic.

src/cpu/pred/multi_level_btb.hh
```diff
@@ -0,0 +1,81 @@
+#ifndef __CPU_PRED_MULTI_LEVEL_BTB_HH__
```
powerjg, line 30: Need copyright

src/cpu/pred/multi_level_btb.cc
```diff
@@ -0,0 +1,242 @@
+/*
+ * Copyright (c) 2026 The University of Edinburgh
+ * Copyright (c) 2026 Technical University of Munich
+ * All rights reserved
+ *
+ * The license below extends only to copyright in the software and shall
```
powerjg: Remove this paragraph unless absolutely required by the copyright holder. Use the version found in [LICENSE](https://github.com/gem5/gem5/blob/stable/LICENSE)

powerjg (changes requested):
Sorry, meant to hit the "request changes" button on the review.

src/cpu/pred/multi_level_btb.cc
yongjiehuang: Yes, that's the logic.

src/cpu/pred/multi_level_btb.cc
yongjiehuang: In case I miss some microarchitectural stuff - why is this compulsory? From my understanding, if L2 is exclusive, it's a vicitm buffer of L1, while an inclusive L3 contains all entries in both L1/L2.

powerjg: I see. I was confused. :)

powerjg (approved):

PR #2826, opened 2025-12-10
stdlib, resources: Fix `list_resources` by calling a new endpoint

Harshil2107:
This PR fixes the `list_resources` function in `client.py` by adding a function that points to a new endpoint on azure functions.

This PR is dependant on https://github.com/gem5/resources-azure-api/pull/7 as this PR adds the endpoint that is called by the updated `list_resources` function.

### DIFF
--- a/src/python/gem5/resources/client.py
+++ b/src/python/gem5/resources/client.py
@@ -238,7 +238,9 @@ def _list_all_resources(
             raise Exception(f"Client: {client} does not exist")
         try:
             resources.extend(
-                clientwrapper[client].get_resources(gem5_version=gem5_version)
+                clientwrapper[client].get_all_resources(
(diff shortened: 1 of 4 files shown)


Harshil2107:
The following script can be used to test the function, once the endpoint is deployed

```
from gem5.resources.downloader import list_resources

resources = list_resources(gem5_version="24.1")
assert len(resources) > 0, "len(list_resources()) == 0"

print("The following resources are compatible with this version of gem5:")
for resource_id, resource_version in resources.items():
    print(f"{resource_id} version {resource_version}")
```

I have tested the new function by running azure functions locally, modifying  gem5 resources config to point to local functions and then running the above script

BobbyRBruce (approved):

PR #2828, opened 2025-12-11
arch-arm: Implement STP with only one store Micro-Op

giactra:
Current STP is cracked in multiple micro-ops depending
on the source register size.

With this commit we only generate one store MicroOp
regardless of the store size. This alleviates
some of the pressure on the SQ

This issue has been raised on upstream some time ago [1]

[1]: https://github.com/orgs/gem5/discussions/1545


andysan (dismissed):

powerjg:
The "regression" test failure makes sense. You're changing the modeled IPC here. Can you update the results for the test in the PR? @erin-le can you give some pointers on how to update the "known good" values?

The x86 test failed due to a download error. Rerunning now.

erin-le:
The "known good"/reference values are located in `tests/gem5/regression_tests/configs/hello-world-binary.py` and `tests/gem5/regression_tests/configs/matrix-multiply.py`, in dictionaries named `reference_ipcs`. To get the updated reference values, you could run the tests locally with the following command, after `cd`ing into the `tests` directory:
```
./main.py run gem5/regression_tests --build-dir=/path/to/build --skip-build -vvv -t 10
```
and inspect the outputs in `tests/testing-results`.

Alternatively, you could download the results from the CI tests themselves. I did this, and the following is the relevant output from the Arm `hello-world` test, which is the only one that's failing.

```
Hello world!
Current IPC is: 0.00745645746683752
Comparison IPC is: 0.007182533150689518
Current IPC is 103.81% of reference
The current IPC is 0.00745645746683752. Expected 0.007182533150689518
```

powerjg (approved):

PR #2829, opened 2025-12-11
mem: Fix asan error in backdoor test

powerjg:
The error is nullptr_with_nonzero_offset. Fixed by using
dummy address instead of nullptr.

### DIFF
--- a/src/mem/backdoor_manager.test.cc
+++ b/src/mem/backdoor_manager.test.cc
@@ -37,6 +37,7 @@ namespace gem5
 {
 namespace backdoor_manager_test
 {
+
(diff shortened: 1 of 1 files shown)


erin-le (approved):
I wasn't sure how to reproduce the error that this PR refers to, but I tested this PR with and without `--with-asan` for the `fast`, `opt`, and `debug` builds and saw that the tests pass in all cases.

powerjg:
> I wasn't sure how to reproduce the error that this PR refers to, but I tested this PR with and without `--with-asan` for the `fast`, `opt`, and `debug` builds and saw that the tests pass in all cases.

Thanks! I'm not sure how to reproduce with public tools either.

PR #2830, opened 2025-12-11
Fix trace data overwrite

dbear496:
This fixes an issue in the exec trace whereby the printed instruction result (data) value would be incorrect.

For example, the `and` instruction in the X86 architecture should print the result value after `D=` like this:
```
6466462000: system.cpu: T0 : 0x4235d4 @bn_mod_exp_mont_fixed_top+1743    : and edx
6466462000: system.cpu: T0 : 0x4235d4 @bn_mod_exp_mont_fixed_top+1743. 0 :   AND_R_R : and   eax, eax, edx : IntAlu :  D=0x0000000000000007
```
However, (without this patch), the trace was reporting erroneous result values for X86 ALU instructions. I found upon further inspection that the trace was actually reporting the result of one of the CC flags instead of the primary result value that is written to the destination register.

This PR addresses the issue by checking whether the InstRecord already has a data (result) value stored before assigning a new value -- this avoids the "correct" value from being subsequently overwritten when CC flags are handled.

I am not certain this is the correct way to solve this issue, but it seems to work for me. In particular, this could break instructions (potentially in other architectures) that write to the destination register after (rather than before) writing CC flags.

### DIFF
--- a/src/arch/isa_parser/operand_types.py
+++ b/src/arch/isa_parser/operand_types.py
@@ -238,7 +238,7 @@ class RegValOperand(RegOperand):
         {{
             RegVal final_val = {reg_val};
             xc->setRegOperand(this, {self.dest_reg_idx}, final_val);
-            if (traceData) {{
+            if (traceData && !traceData->getDataStatus()) {{
                 traceData->setData({self.reg_class}, final_val);
             }}
         }}"""
@@ -371,7 +371,7 @@ class VecRegOperand(RegOperand):
 
     def makeWrite(self):
         return f"""
-        if (traceData) {{
+        if (traceData && !traceData->getDataStatus()) {{
             traceData->setData({self.reg_class}, &tmp_d{self.dest_reg_idx});
         }}
         """
@@ -425,7 +425,7 @@ class VecPredRegOperand(RegOperand):
 
     def makeWrite(self):
         return f"""
-        if (traceData) {{
+        if (traceData && !traceData->getDataStatus()) {{
             traceData->setData({self.reg_class}, &tmp_d{self.dest_reg_idx});
         }}
         """
@@ -531,7 +531,7 @@ class ControlRegOperand(RegOperand):
             f"{self.dest_reg_idx}, {self.base_name});\n"
         )
         wb += f"""
-        if (traceData) {{
+        if (traceData && !traceData->getDataStatus()) {{
             traceData->setData({self.reg_class}, {self.base_name});
         }}
         """


giactra:
Hi @dbear496, thanks for this PR.

Firstly let me state this is a known issue in the tracer.
As you mentioned in your PR message, this is probably not the correct way to solve the issue. Effectively, why should the first destination be more important than the second one?

To really solve the problem, we should be recording and potentially printing *all* destination values.

Two good news here:

1) It is not hard to implement this: one simply needs to amend the logic in the tracer https://github.com/gem5/gem5/blob/stable/src/sim/insttracer.hh#L107 to use a vector of Data instead of a single data. Then we should probably change the behaviour/semantics of setData to append instead of assign.

2) It's a feature everyone has been waiting for and if you do this you will have eternal gratitude from the community.

What do you think? Are you convinced? ;)

dbear496:
@giactra , Thanks for the response.

> Effectively, why should the first destination be more important than the second one?

The destinations are ordered by the operand sort priority, which is defined in [src/arch/&lt;arch&gt;/isa/operands.isa](https://github.com/gem5/gem5/blob/stable/src/arch/x86/isa/operands.isa). According to the [gem5 documentation](https://www.gem5.org/documentation/general_docs/architecture_support/isa_parser/#instruction-operands), "a sort priority used to control the order of operands in disassembly." So I believe it actually does make sense to prefer printing the first destination over the last. The only concern I have is this change is rather unilateral, and while I do believe the patched behavior is better in general, I do not know for certain if it is best in all cases (as I am not that familiar with the codebase holistically).

You recommend printing _all_ destination values. However, my concern is this could potentially clutter the trace line. The X86 `and` instruction, for instance, writes to 5 different CC flag registers in addition to the primary destination register. Do you think this is a valid concern?

giactra:
> @giactra , Thanks for the response.
> 
> > Effectively, why should the first destination be more important than the second one?
> 
> The destinations are ordered by the operand sort priority, which is defined in [src/arch/<arch>/isa/operands.isa](https://github.com/gem5/gem5/blob/stable/src/arch/x86/isa/operands.isa). According to the [gem5 documentation](https://www.gem5.org/documentation/general_docs/architecture_support/isa_parser/#instruction-operands), "a sort priority used to control the order of operands in disassembly." So I believe it actually does make sense to prefer printing the first destination over the last. The only concern I have is this change is rather unilateral, and while I do believe the patched behavior is better in general, I do not know for certain if it is best in all cases (as I am not that familiar with the codebase holistically).
> 

Thanks, I actually didn't know this.

My feeling is that it is probably quite hard to have this rudimentary sorting behaviour work properly for all ISAs.
Instruction disassembly is really instruction specific... We cannot expect the operand keywork itself do all the magic as it
might require us to redefine different operands (targeting the same register) for different instructions

> You recommend printing _all_ destination values. However, my concern is this could potentially clutter the trace line. The X86 `and` instruction, for instance, writes to 5 different CC flag registers in addition to the primary destination register. Do you think this is a valid concern?

That is a valid concern indeed.
What do you think about amending the parser to include a suffix for an operand that should *NOT* be printed? That will spare us from those instructions writing to multiple registers

erin-le:
@dbear496: Hello, could you respond to the previous comment on this PR when you get a chance? Thank you!

dbear496:
Apologies for the delay.

I dislike the idea of using a special suffix for operands that should not be printed because this violates separation between names and semantics.

Perhaps it would be better to somehow piggyback on the existing disassembly. That is, the exec trace already prints the disassembled instruction with the register names of relevant operands, and we could prefer printing the value of a result operand that is also printed by name in the disassembled instruction. If no destination registers are printed by name, then we could fall back to existing behavior.

I have not had an opportunity to look into the logic of which operand registers are chosen to be printed by name, but I highly suspect that it uses the operand sort priority. If this is the case, then my current implementation is a very good approximation of this behavior and merely diverges in the fallback case.

dbear496:
If the disassembly logic chooses to print destination register names based on which operand(s) have the _highest_ sort priority, then I think it is a little bit silly to print the result value of the operand with the _lowest_ sort priority. But alas, this is existing behavior. Do we want to keep the old behavior simply because fixing it might break _something_ when we are not even sure if _something_ actually exists?

erin-le:
@giactra: Hello, could you respond to the author on this PR when you get a chance? Thank you!

PR #2841, opened 2025-12-16
cpu-o3: Update O3 CPU modules to reduce coupling.

udaimuhammed:
Refactors the O3 CPU model and related components to resolve circular dependencies and satisfy explicit dependency requirements for the Bazel build system.

Key changes include:
*   **cpu/o3**: Moved inline method implementations (e.g., `DynInst`, `LSQUnit`, `InstructionQueue`) from header files to source files to break dependency cycles (specifically `DynInst` <-> `LSQUnit` and `DynInst` <-> `CPU`).
*   **cpu/o3**: Refactored `LSQRequest` and `LSQ` class definitions to improve header modularity.
*   **cpu/o3**: Updated `BaseO3CPU.py` to point to the correct C++ header (`cpu/o3/cpu.hh`).
*   **cpu/checker**: Cleaned up includes in `cpu_impl.hh` and `cpu.hh` to reduce dependency coupling.
*   **cpu/exetrace**: Moved `getInstRecord` implementation to `exetrace.cc`.
*   **General**: Added missing `#include` directives across various source files to fix compilation errors arising from stricter dependency sandboxing.

### DIFF
--- a/src/cpu/o3/lsq.hh
+++ b/src/cpu/o3/lsq.hh
@@ -45,8 +45,6 @@
 #include <cassert>
 #include <cstdint>
 #include <list>
-#include <map>
-#include <queue>
 #include <vector>
 
 #include "arch/generic/mmu.hh"
@@ -55,8 +53,8 @@
 #include "base/statistics.hh"
 #include "base/types.hh"
 #include "cpu/inst_seq.hh"
-#include "cpu/o3/dyn_inst_ptr.hh"
 #include "cpu/utils.hh"
+#include "dyn_inst_ptr.hh"
 #include "enums/SMTQueuePolicy.hh"
 #include "mem/port.hh"
 #include "sim/sim_object.hh"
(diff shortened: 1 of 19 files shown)


src/cpu/o3/lsq.hh
```diff
@@ -984,7 +969,7 @@ class LSQ
     DcachePort dcachePort;
 
     /** The LSQ units for individual threads. */
-    std::vector<LSQUnit> thread;
+    std::vector<std::unique_ptr<LSQUnit>> thread;
```
dhschall, line 972: Just out of curiosity. Was there a reason for this change?

dhschall (commented):
Thank you @udaimuhammed for this effort of refactoring and breaking the dependencies.
Where did all the new headers come from that you now have to include in all cc files? From the `cpu/o3/cpu.hh` in the dynamic instruction?

The clang compilation still seems to fail for the `exectracer.cc`. Also please run the formater.

Also a general comment. It seems that you not only removed the cyclic dependencies but also moved out subclasses from the LSQ unit and you also changed classes to pointer to classes. It would make it easier to have separate PRs.

Anyway, overall it looks good once you have fixed the compilation errors

src/cpu/o3/lsq.hh
udaimuhammed: Using values instead of pointers would require full definition of LSQUnit (i.e.  we need to include lsq_unit.hh). this caused Bazel compilation to fail with another cyclic dependency.

powerjg:
> Where did all the new headers come from that you now have to include in all cc files? From the cpu/o3/cpu.hh in the dynamic instruction?

We use clang-tidy, clang-include-fixer, and some internal tools for this. We should really run these throughout the codebase. The result is that we check to see what headers are actually used and imports all of the directly. We've been getting "lucky" in that we often use declarations from transitively included headers. It would be better practice to make sure that all of the declarations that are used in each file have a direct include in that file.

udaimuhammed:
>  Where did all the new headers come from that you now have to include in all cc files? From the `cpu/o3/cpu.hh` in the dynamic instruction?

that's a requirement enforced by Bazel. if your file depend on it, you should explicitly include it. 

> Also a general comment. It seems that you not only removed the cyclic dependencies but also moved out subclasses from the LSQ unit and you also changed classes to pointer to classes. It would make it easier to have separate PRs.

Sorry for the large change, but that's still necessary to break the cyclic dependencies. The yanking of LSQRequest subclass simplified Bazel build targets auto-dependency generation. if you think a separate PR is absolutely needed, I am happy to do it.

udaimuhammed:
I don't understand `clang-format-check` failure.
A)  pre-commit should have caught any formatting issues before the push. 
B) I ran this command 
```
(.venv) ~/g/gem5 > git diff --name-only develop...HEAD | xargs python util/style.py -f
invalid line length in src/cpu/checker/cpu_impl.hh:154
Warning: cannot automatically fix overly long lines.
invalid line length in src/cpu/o3/BaseO3CPU.py:255
Warning: cannot automatically fix overly long lines.
invalid line length in src/cpu/o3/probe/elastic_trace.cc:843
Warning: cannot automatically fix overly long lines.
```
all these lines not touched by my change.

dhschall:
> if you think a separate PR is absolutely needed, I am happy to do it.

I think its ok. I can understand that breaking these dependencies is not easy and it's worth to make a bigger change
As long as there is no chance in functionality I'm.
I couldn't spot any problem.

powerjg:
@BobbyRBruce can I get a thumbs up on merging this without passing the clang-format-check. We've tried a number of things to get it to pass, but the CI test is still failing. It passes locally...

BobbyRBruce (approved):

PR #2844, opened 2025-12-17
util: Fix decode_packet_trace.py

### LINKED ISSUE #2831: script `util/decode_packet_trace.py` has not been maintained for a long time
`util/decode_packet_trace.py:91-107`: 
```
    # Decode the packet messages until we hit the end of the file
    while protolib.decodeMessage(proto_in, packet):
        num_packets += 1
        # ReadReq is 1 and WriteReq is 4 in src/mem/packet.hh Command enum
        cmd = "r" if packet.cmd == 1 else ("w" if packet.cmd == 4 else "u")
        if packet.HasField("pkt_id"):
            ascii_out.write(f"{packet.pkt_id},")
        if packet.HasField("flags"):
            ascii_out.write(
                f"{cmd},{packet.addr},{packet.size},{packet.flags},{packet.tick}"
            )
        else:
            ascii_out.write(f"{cmd},{packet.addr},{packet.size},{packet.tick}")
        if packet.HasField("pc"):
            ascii_out.write(f",{packet.pc}\n")
        else:
            ascii_out.write("\n")
```

This line  `cmd = "r" if packet.cmd == 1 else ("w" if packet.cmd == 4 else "u")`.
It can only process cmd: `ReadReq` and `WriteReq`.

I wander whether it's good to bearly print the cmd id here

HUJIYONG:
I tried an implementaiton, which gives the format:
```
# <<< protobuf header start
# .....
# >>> protobuf header end
# column header
line1
line2
.........
```
It will:

1. print the protobuf header for better understanding
2. print the column header (automatically extract field names from protocol defination)
3. print all columns in the packet with original format

for example:
image


powerjg (approved):
Thanks for the contribution!

PR #2845, opened 2025-12-18
tests: Update PR CI clang-format check to use verbose mode.

udaimuhammed:
when clang-format check fails, we should get info about which files are failing.

### DIFF
--- a/util/run-git-clang-format.py
+++ b/util/run-git-clang-format.py
@@ -376,7 +376,6 @@ if __name__ == "__main__":
         # Run git-clang-format in 'CI' context
         command = [
             "git-clang-format",
-            "--quiet",
             "--style=file",
             f"--commit={ci_pr_base_commit}",
         ]
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -50,7 +50,7 @@ jobs:
 
             - name: Run clang-format check
               run: |
-                  python util/run-git-clang-format.py --ci-pr-base-commit ${{ steps.get-base-commit.outputs.base_commit }}
+                  python util/run-git-clang-format.py --verbose --ci-pr-base-commit ${{ steps.get-base-commit.outputs.base_commit }}
                   if [ $? -ne 0 ]; then
                     echo "::error::Code formatting issues detected! Please run clang-format on your changes."
                     exit 1


powerjg (dismissed):
Thanks @udaimuhammed! Hopefully this will help contributors when the CI tests fail.

util/run-git-clang-format.py
```diff
@@ -376,7 +376,6 @@ def make_ignore_path_rel(path_str: str, repo_root: Path) -> Optional[Path]:
         # Run git-clang-format in 'CI' context
         command = [
             "git-clang-format",
-            "--quiet",
```
BobbyRBruce, line 379: There's still another `--quiet` when running in he normal staged context. Line 387/388. Should it be removed too?

BobbyRBruce (commented):
Can you include some clang-breaking format changes in this PR just so we can see if it's suitably verbose prior to merging?

util/run-git-clang-format.py
powerjg: I think we want this "loud" since we should be able to see it in the CI output on github. For the other one, it's run when committing. I think that even if it's not quiet, the user won't see it due to how precommit works. I'm not 100% sure on the latter, but it's kind of a different issue than the one here.

powerjg:
> Can you include some clang-breaking format changes in this PR just so we can see if it's suitably verbose prior to merging?

Precommit error is useful

```
If you are seeing this message in CI, reproduce locally with: `pre-commit run --all-files`.
To run `pre-commit` as part of git workflow, use `pre-commit install`.
All changes made by hooks:
diff --git a/src/mem/port_wrapper.hh b/src/mem/port_wrapper.hh
index 2c4f955..0491734 100644
--- a/src/mem/port_wrapper.hh
+++ b/src/mem/port_wrapper.hh
@@ -87,7 +87,7 @@ class RequestPortWrapper : public RequestPort
     // TimingRequestProtocol
     bool recvTimingResp(PacketPtr) override;
     void recvReqRetry() override;
-    
+
     void setRangeChangeCallback(RecvReqRetryCallback);
     void setTimingCallbacks(RecvTimingRespCallback, RecvReqRetryCallback);
```

git-clang-format isn't terribly useful, though

```
  DEBUG: Running command: 'git rev-parse --show-toplevel'
  DEBUG:     Command 'git rev-parse --show-toplevel' returned 0
  DEBUG: Repository root: '/home/runner/work/gem5/gem5'
  DEBUG: Running in CI mode with base commit id '4729aba1ccb8f73972449d17981c36f9ff5866c4'
  DEBUG: No disabled files.
  DEBUG: Running command: 'git-clang-format --style=file --commit=4729aba1ccb8f73972449d17981c36f9ff5866c4'
  DEBUG:     Command 'git-clang-format --style=file --commit=4729aba1ccb8f73972449d17981c36f9ff5866c4' returned 1
```

powerjg:
@BobbyRBruce, if you have suggestions on how to improve the errors, please let Udai know what you would like to see changed. Otherwise, can you review this? (I can't approve anymore since I just pushed.)

BobbyRBruce (approved):

BobbyRBruce:
Approved. Merged (bipassing the CI tests as this change cannot affect any CI tests beyond the clang-format check, which is functioning correctly)

PR #2846, opened 2025-12-19
mem-cache: fix memory leakage in fetch directed prefetcher

ChoungJX:
`notifyFTQRemove()` is triggered by certain O3CPU events (e.g., squash events), which results in flushing the `pfq`. At this point, the pfq may contain `PrefetchRequests` that have already generated a `pkt` but have not yet sent it to the cache (i.e., their `pkt` was never retrieved via `getPacket()`). For these requests, the `pkt` must be manually deleted before the requests are evicted from the `pfq`.


dhschall (approved):
Thanks for spotting this leakage.
Looks good to me.

BobbyRBruce (approved):

PR #2852, opened 2025-12-22
misc: Add a PyPort to write to physmem from python

giactra:
The PyPort is the pybind11 interface around the PortProxy class.  It is meant to be used to read and write binary blobs to the guest memory from python.

As an example, a PyPort can be returned from the System:

port = system.physPort

Its main APIs are the python based methods defined in src/python/pybind11/port.cc. Those are simply:

def read(addr, size) -> bytes
def write(addr, data) -> None

To write an arbitrary binary blob to memory: at the moment we have an heterogeneous set of solutions to do so. One is through the extras parameters:

extras = VectorParam.String([], "Additional object files to load") extras_addrs = VectorParam.Addr(
    [], "Load addresses for additional object files"
)

Though for some obscure reason those are only defined for KernelWorkloads. "Simpler" environments (e.g. a traffic generator platform with no binary) would not be able to use this static interface. It also requires a path to a file in the the host FS for things to work and does not allow inline slamming of binary data (sure, a python platform can always write a binary blob to /tmp/ disk but it is not ideal)

Another alternative is via the AbstractMemory.image_file parameter. The caveat (on top of the requiring a file as solution above) is that

a) there is no way of specifying the offset of the image_file b) I cannot specify multiple image files at the same time, so the only workaround to this is to define multiple memories

This python port solution fixes all aforementioned shortcomings and therefore has the following perks:

1) It's a generall solution which works with every configuration

2) We can use it to dynamically patch (write) the content of memory during simulation from python

3) We can use it to dynamicall read the content of memory during simulation from python for debug purposes

Change-Id: I99b274fb64504d922044c0b8cf24e6ad75060886

### DIFF
--- a/src/python/pybind11/port.cc
+++ b/src/python/pybind11/port.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2025 Arm Limited
+ * Copyright (c) 2025 Arm Limited
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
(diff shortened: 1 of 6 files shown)


src/python/pybind11/port.cc
```diff
@@ -0,0 +1,77 @@
...
+                                 size);
+            },
+            py::arg("addr"), py::arg("size"),
+            "Read size bytes from addr and return as Python bytes")
+        .def(
+            "write",
+            [](gem5::PortProxy &self, Addr phys_addr, py::buffer src_buf) {
+                py::buffer_info info = src_buf.request();
+
+                self.writeBlob(phys_addr,
```
powerjg: Optional: It may be better to use `tryWriteBlob`/`tryReadBlob` and raise a python exception on failure. However, this is OK too. By using simply `writeBlob`/`readBlob` if the user does something wrong it will result in a `fatal` and this could be quite annoying if something goes wrong *after* running the simulation since you would lose all of your data (no stats dump, etc.). `fatal` also doesn't do an abort, IIRC, so it's difficult to debug after the fact. I don't have a strong preference, though.

powerjg (dismissed):
Super cool how straightforward the code is! I'll approve as-is, but consider the comment below.

src/python/pybind11/port.cc
giactra: Thanks that's a good point! Let me know what you think about this new version:

**Note**: in PortProxy tryReadBlob/tryWriteProxy always returns true

powerjg: lol. Well, if that's ever changed we won't get fatals in this code. Thanks for the update!

powerjg (approved):

PR #2859, opened 2025-12-29
sim: multisim: Fix stats filename mangling

pranith:
Commit 91ee67ce8950 added an option to allow multiple output stat files. However, in multisim, if we don't use this argument, the previously named 'stats.txt' file is now renamed to a literal list as '['stats.txt']'.

Fix it by checking of the stats file argument is passed.


pranith:
This has been fixed in PR 2835. Closing.

PR #2861, opened 2025-12-29
base, configs, mem, python, stdlib: Refactor addr range

### LINKED ISSUE #2855: Refactoring `AddrRange` for Flexible Interleaving and Sparse Ranges
# Design Doc: Refactoring `AddrRange` for Flexible Interleaving and Sparse Ranges

## 1. Background & Motivation

The `AddrRange` class in gem5 is a fundamental primitive used throughout the memory system to represent address ranges. It currently supports contiguous ranges and power-of-two bit-masked interleaving.

However, modern memory systems and specific architecture requirements (e.g., x86 legacy I/O holes) necessitate more flexibility:

1. **Non-power-of-two interleaving**: Memory controllers often interface with 3, 6, or 12 channels, requiring modulo-based arithmetic rather than simple bit-masking.
2. **Sparse Ranges ("Holes")**: Some address ranges must exclude specific sub-ranges (e.g., the local APIC/IO hole at 3GB-4GB in x86). Currently, `AbstractMemory` assumes a single contiguous range (after interleaving), requiring all consumers to manually handle these holes or forcing checking logic into every memory object.

This design proposes a refactor of `AddrRange` to support these features with minimal performance impact (~0% regression) and backward compatibility for the ~400 internal usage sites.

## 2. Requirements

* **Arbitrary Interleaving**: Support modulo-based and potentially hash-based interleaving.
* **Sparse Address Spaces**: Support "holes" where a range `[A, D)` is valid but `[B, C)` inside it is excluded. `getOffset` must account for these holes (i.e., addresses after the hole are shifted down in offset).
* **Performance**: `contains` and `getOffset` are on the hot path. Overheads must be negligible compared to the current implementation.
* **Backward Compatibility**: Minimal changes to `AbstractMemory` and other clients. `AddrRange` is passed by value extensively; changing this to pointer semantics would be a massive refactor.

## 3. Investigated Designs

We evaluated four designs. We recommend **Design A** (Strategy Pattern) and explicitly reject the others based on maintainability, object size, and refactoring costs.

### Design A: Strategy Pattern (Polymorphic Matcher) - *Recommended*

Introduce a lightweight mapping policy class held by a `std::shared_ptr` within `AddrRange`.

**Structure:**

```cpp
class AddrRange {
  private:
    Addr _start;
    Addr _end;
    // nullptr implies simple contiguous range (Fast Path)
    std::shared_ptr<AddrMapPolicy> policy; 

  public:
    // API delegates to policy if present
    bool contains(Addr a) const;
    Addr getOffset(Addr a) const; 
};

class AddrMapPolicy {
  public:
    virtual bool contains(Addr rangeStart, Addr rangeEnd, Addr a) const = 0;
    virtual Addr getOffset(Addr rangeStart, Addr rangeEnd, Addr a) const = 0;
    virtual void serialize(CheckpointOut &cp) const = 0;
};

```

**Pros:**

* **Flexibility**: Can implement any mapping logic without bloating `AddrRange`.
* **Size**: `AddrRange` remains small (~24 bytes).
* **Performance**: Common case is fast; complex cases incur one virtual call.

### Design B: `std::variant` (Static Polymorphism)

Use `std::variant<BitMaskPolicy, ModuloPolicy, SparsePolicy>` inside `AddrRange`.

**Pros:**

* **Allocation**: No heap allocation required (uses inline storage).
* **Overhead**: No virtual function overhead (dispatch handles via `std::visit` or switch).

**Cons:**

* **Size**: `AddrRange` size grows to `sizeof(LargestPolicy)` + constant overhead. For example, `BitMaskPolicy` (containing vectors) is at least 24 bytes; `SparsePolicy` (vector of ranges) could be significantly larger.
* **Fixed Extensibility**: Adding a new policy requires changing the `AddrRange` type definition and recompiling the core of gem5.
* **Copy Cost**: Copying `AddrRange` implies deep-copying the variant (including internal vectors). This is significantly more expensive than copying a `shared_ptr`.

### Design C: Hybrid / Fixed Fields

Add specific fields to `AddrRange` for the new features (e.g., adding `std::vector<AddrRange> holes`, `uint64_t modulo`, `uint64_t match`).

**Pros:**

* **Simplicity**: Simple to understand if the requirements are strictly fixed and never change.
* **Performance**: Fastest for the specific case (completely inline logic).

**Cons:**

* **Bloat**: The `AddrRange` object grows indefinitely as new features are added.
* **Complexity**: The `contains` method becomes a monolithic "spaghetti" function checking bitmasks, modulos, holes, and ranges all in one block.
* **Maintenance**: Harder to test isolated features or add new interleaving schemes without touching critical path code.

### Design D: Virtual Base Class (`virtual AddrRange`) - *Rejected*

Making `AddrRange` itself a virtual base class with subclasses like `InterleavedAddrRange`.

**Cons:**

* **Object Slicing**: `AddrRange` is passed by **value** in hundreds of functions. Converting it to a polymorphic base class would cause object slicing unless the entire codebase was refactored to pass by pointer.
* **Refactoring Scope**: Changing usage to pointers/references across ~400 files is high-risk/high-effort.

### 4.1. The Policy Hierarchy

* `gem5::AddrMapPolicy` (Abstract Base)
* `gem5::MaskedInterleavingPolicy`: Encapsulates current XOR/Bitmask logic.
* `gem5::ModuloInterleavingPolicy`: Implements `addr % N == match`.
* `gem5::SparsePolicy`: Manages a list of invalid sub-ranges ("holes").

### 4.2. Handling "Holes" (Sparse Usage)

This is critical for the x86 3-4GiB I/O hole.

* **`contains(a)`**: Checks if `a` is in the main range AND not in any defined hole.
* **`getOffset(a)`**: Calculates offset by subtracting `_start` and *also* subtracting the size of any holes that come before `a`.
* *Example:* Range 0-5GB, Hole 3-4GB.
* `getOffset(2GB)` = 2GB.
* `getOffset(4.5GB)` = 4.5GB - 0 - 1GB (hole size) = 3.5GB.
* This ensures the backing store in `AbstractMemory` is packed contiguously (4GB size total), avoiding wasted allocation for the hole.

### 4.3. Serialization

* If `policy` is null, serialize a flag indicating "Simple/Contiguous".
* If `policy` exists, serialize a type ID, then call `policy->serialize()`.
* `unserialize` reads the ID and instantiates the correct Policy subclass.

### 4.4. Python Integration

The `AddrRange` class is exposed to Python via pybind11 in `src/python/pybind11/core.cc`. Since `AddrRange` is passed by value, the Python integration is straightforward:

*   **Existing Bindings**: The existing bindings for `start`, `end`, `size`, etc., remain unchanged. The internal delegation to `policy` is hidden from Python.
*   **New Constructors**: We will add new constructors (or static factory methods) to `AddrRange` in `core.cc` to allow creating complex ranges from Python configuration scripts.
    *   `AddrRange(start, end, modulo, match)` for modulo-based interleaving.
    *   `AddrRange(start, end, holes)` for sparse ranges.
*   The `AddrMapPolicy` class hierarchy does **not** need to be exposed to Python. Python objects will simply create `AddrRange` objects, which internally construct the appropriate C++ policy.

## 5. Implementation Plan

1. **Refactor (Phase 1)**: Extract current bitmask logic into `MaskedInterleavingPolicy`. Update `AddrRange` to use the shared policy pointer. Ensure `policy == nullptr` optimization is in place.
2. **Verify**: Run all regressions to ensure no performance regression on the hot path.
3. **Extend (Phase 2)**: Implement `ModuloInterleavingPolicy` for 3/6 channel memory.
4. **Extend (Phase 3)**: Implement `SparsePolicy` to handle address holes.
5. **Test**: Add unit tests for non-power-of-two interleaving and ranges with holes (validating correct offset calculations).

powerjg:
See #2855 for more details. Detailed commit messages in each commit.

**Important note**: Gemini wrote most of this code. I supervised, tested, and carefully reviewed.

### DIFF
--- a/src/base/addr_range.hh
+++ b/src/base/addr_range.hh
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2012, 2014, 2017-2019, 2021 Arm Limited
+ * Copyright (c) 2026 Google Inc.
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
@@ -44,8 +45,10 @@
 #include <algorithm>
 #include <iterator>
 #include <list>
+#include <utility>
 #include <vector>
 
+#include "base/addr_range_policy.hh"
 #include "base/bitfield.hh"
 #include "base/cprintf.hh"
 #include "base/logging.hh"
@@ -89,69 +92,65 @@ class AddrRange
     Addr _start;
     Addr _end;
 
-    /**
-     * Each mask determines the bits we need to xor to get one bit of
-     * sel. The first (0) mask is used to get the LSB and the last for
-     * the MSB of sel.
-     */
-    std::vector<Addr> masks;
+    /// Valid sub-ranges (hunks) for sparse ranges.
+    /// If empty, the range is considered contiguous [_start, _end).
+    std::vector<std::pair<Addr, Addr>> _chunks;
 
-    /** The value to compare sel with. */
-    uint8_t intlvMatch;
+    std::shared_ptr<AddrMapPolicy> _policy;
 
   protected:
     struct Dummy {};
 
     // The dummy parameter Dummy distinguishes this from the other two argument
     // constructor which takes two Addrs.
+    // This constructor takes multiple address ranges and merges them into a
+    // single address range, if possible. If not possible, there is a fatal
+    // error.
     template <class Iterator>
     AddrRange(Dummy, Iterator begin_it, Iterator end_it)
-        : _start(1), _end(0), intlvMatch(0)
+        : _start(1), _end(0), _policy(nullptr)
     {
         if (begin_it != end_it) {
             // get the values from the first one and check the others
             _start = begin_it->_start;
             _end = begin_it->_end;
-            masks = begin_it->masks;
-            intlvMatch = begin_it->intlvMatch;
+            _chunks = begin_it->_chunks;
+            _policy = begin_it->_policy;
         }
 
         auto count = std::distance(begin_it, end_it);
         // either merge if got all ranges or keep this equal to the single
         // interleaved range
         if (count > 1) {
-            fatal_if(count != (1ULL << masks.size()),
-                    "Got %d ranges spanning %d interleaving bits.",
-                    count, masks.size());
+            fatal_if(!interleaved(), "Merging non-interleaved ranges?");
 
-            uint8_t match = 0;
+            std::vector<std::shared_ptr<AddrMapPolicy>> policies;
+            policies.reserve(count);
+
             for (auto it = begin_it; it != end_it; it++) {
-                fatal_if(!mergesWith(*it),
-                        "Can only merge ranges with the same start, end "
-                        "and interleaving bits, %s %s.", to_string(),
-                        it->to_string());
-
-                fatal_if(it->intlvMatch != match,
-                        "Expected interleave match %d but got %d when "
-                        "merging.", match, it->intlvMatch);
-                ++match;
+                fatal_if(it->_start != _start || it->_end != _end,
+                         "Can only merge ranges with the same start and end");
+                fatal_if(it->_chunks != _chunks,
+                         "Can only merge ranges with the same sparse chunks");
+                fatal_if(!it->_policy, "Cannot merge flat ranges");
+                policies.push_back(it->_policy);
             }
-            masks.clear();
-            intlvMatch = 0;
+
+            // Attempt to merge policies
+            _policy = _policy->createMerged(policies);
         }
     }
 
   public:
-
     /**
+     * Default constructor, creates an invalid address range.
+     *
      * @ingroup api_addr_range
      */
-    AddrRange()
-        : _start(1), _end(0), intlvMatch(0)
-    {}
+    AddrRange() : _start(1), _end(0), _policy(nullptr) {}
 
     /**
-     * Construct an address range
+     * Construct a masked interleaved address range.
      *
      * If the user provides a non empty vector of masks then the
      * address range is interleaved. Each mask determines a set of
@@ -182,13 +181,168 @@ class AddrRange
      */
     AddrRange(Addr _start, Addr _end, const std::vector<Addr> &_masks,
               uint8_t _intlv_match)
-        : _start(_start), _end(_end), masks(_masks),
-          intlvMatch(_intlv_match)
+        : _start(_start), _end(_end), _policy(nullptr)
     {
-        // sanity checks
-        fatal_if(!masks.empty() && _intlv_match >= 1ULL << masks.size(),
-                 "Match value %d does not fit in %d interleaving bits\n",
-                 _intlv_match, masks.size());
+        if (!_masks.empty()) {
+            // sanity checks
+            fatal_if(_intlv_match >= 1ULL << _masks.size(),
+                     "Match value %d does not fit in %d interleaving bits\n",
+                     _intlv_match, _masks.size());
+            _policy = std::make_shared<MaskedInterleavingPolicy>(_masks,
+                                                                 _intlv_match);
+        }
+    }
+
+    /**
+     * Create an address range with modulo-based interleaving.
+     *
+     * If the user provides a stripe count of 1, the address range is not
+     * interleaved.
+     * Using modulo instead of interleaving allows for non-power
+     * of 2 stripe counts.
+     *
+     * @param start The start address of the range.
+     * @param end The end address of the range.
+     * @param stripes The number of stripes (channels).
+     * @param intlv_match The stripe index for this range.
+     * @param intlv_low_bit The bit position of the interleaving granularity
+     *                      (log2).
+     *
+     * @ingroup api_addr_range
+     */
+    AddrRange(Addr start, Addr end, uint32_t stripes, uint32_t intlv_match,
+              uint32_t intlv_low_bit = 0)
+        : _start(start), _end(end), _policy(nullptr)
+    {
+        fatal_if(stripes == 0, "Modulo stripes must be greater than 0");
+        fatal_if(intlv_match >= stripes,
+                 "Modulo match value %d must be less than stripes %d",
+                 intlv_match, stripes);
+        if (stripes > 1) {
+            Addr range_size = end - start;
+            Addr granularity = 1ULL << intlv_low_bit;
+            fatal_if(
+                range_size % granularity != 0,
+                "Modulo range size %d must be a multiple of granularity %d",
+                range_size, granularity);
+            Addr num_blocks = range_size / granularity;
+            fatal_if(num_blocks % stripes != 0,
+                     "Modulo range size %d (in blocks) must be a multiple of "
+                     "stripes %d",
+                     num_blocks, stripes);
+            _policy = std::make_shared<ModuloInterleavingPolicy>(
+                stripes, intlv_match, intlv_low_bit);
+        }
+    }
+
+    /**
+     * Create an address range with sparse sub-ranges (holes).
+     *
+     * This constructor is useful for creating address ranges that are not
+     * contiguous, such as the I/O hole in x86.
+     * This constructor only enables "simple" sparse ranges, i.e. no
+     * interleaving.
+     *
+     * @param ranges Vector of valid address chunks (start, end).
+     *               Must be sorted and non-overlapping.
+     *
+     * @ingroup api_addr_range
+     */
+    AddrRange(const std::vector<std::pair<Addr, Addr>> &ranges)
+        : _policy(nullptr)
+    {
+        _chunks = ranges;
+        std::sort(_chunks.begin(), _chunks.end());
+
+        if (_chunks.empty()) {
+            _start = 1;
+            _end = 0;
+        } else {
+            for (size_t i = 0; i < _chunks.size(); ++i) {
+                fatal_if(_chunks[i].first >= _chunks[i].second,
+                         "Chunk start (%#llx) >= end (%#llx)",
+                         _chunks[i].first, _chunks[i].second);
+                if (i > 0) {
+                    fatal_if(_chunks[i - 1].second > _chunks[i].first,
+                             "Sparse chunks must be non-overlapping: "
+                             "chunk %d end (%#llx) > chunk %d start (%#llx)",
+                             i - 1, _chunks[i - 1].second, i,
+                             _chunks[i].first);
+                }
+            }
+            _start = _chunks.front().first;
+            _end = _chunks.back().second;
+        }
+    }
+
+    /**
+     * Create an address range with sparse sub-ranges and masked interleaving.
+     *
+     * This constructor will create an interleaved address range in the same
+     * way as the masked interleaving constructor, but with sparse sub-ranges.
+     * If the masks are empty, the address range will be sparse only.
+     *
+     * @param ranges Vector of valid address chunks (start, end).
+     *               Must be sorted and non-overlapping.
+     * @param masks The input vector of masks.
+     * @param intlv_match The matching value of the xor operations.
+     *
+     * @ingroup api_addr_range
+     */
+    AddrRange(const std::vector<std::pair<Addr, Addr>> &ranges,
+              const std::vector<Addr> &masks, uint8_t intlv_match)
+        : AddrRange(ranges) // Delegate setup of chunks/start/end
+    {
+        if (!masks.empty()) {
+            fatal_if(intlv_match >= 1ULL << masks.size(),
+                     "Match value %d does not fit in %d interleaving bits\n",
+                     intlv_match, masks.size());
+            _policy =
+                std::make_shared<MaskedInterleavingPolicy>(masks, intlv_match);
+        }
+    }
+
+    /**
+     * Create an address range with sparse sub-ranges and modulo interleaving.
+     *
+     * This constructor will create an interleaved address range in the same
+     * way as the modulo interleaving constructor, but with sparse sub-ranges.
+     * If the stripes are 1, the address range will be sparse only.
+     *
+     * @param ranges Vector of valid address chunks (start, end).
+     *               Must be sorted and non-overlapping.
+     * @param stripes The number of stripes (channels).
+     * @param intlv_match The matching value of the modulo operation.
+     * @param intlv_low_bit The lowest bit to use for the modulo operation.
+     *
+     * @ingroup api_addr_range
+     */
+    AddrRange(const std::vector<std::pair<Addr, Addr>> &ranges,
+              uint32_t stripes, uint32_t intlv_match,
+              uint32_t intlv_low_bit = 0)
+        : AddrRange(ranges) // Delegate setup of chunks/start/end
+    {
+        fatal_if(stripes == 0, "Modulo stripes must be greater than 0");
+        fatal_if(intlv_match >= stripes,
+                 "Modulo match value %d must be less than stripes %d",
+                 intlv_match, stripes);
+        if (stripes > 1) {
+            for (const auto &chunk : ranges) {
+                Addr chunk_size = chunk.second - chunk.first;
+                Addr granularity = 1ULL << intlv_low_bit;
+                fatal_if(chunk_size % granularity != 0,
+                         "Modulo chunk size %d must be a multiple of "
+                         "granularity %d",
+                         chunk_size, granularity);
+                Addr num_blocks = chunk_size / granularity;
+                fatal_if(num_blocks % stripes != 0,
+                         "Modulo chunk size %d (in blocks) must be a multiple "
+                         "of stripes %d",
+                         num_blocks, stripes);
+            }
+            _policy = std::make_shared<ModuloInterleavingPolicy>(
+                stripes, intlv_match, intlv_low_bit);
+        }
     }
 
     /**
@@ -218,45 +372,58 @@ class AddrRange
      * @ingroup api_addr_range
      */
     AddrRange(Addr _start, Addr _end, uint8_t _intlv_high_bit,
-              uint8_t _xor_high_bit, uint8_t _intlv_bits,
-              uint8_t _intlv_match)
-        : _start(_start), _end(_end), masks(_intlv_bits),
-          intlvMatch(_intlv_match)
+              uint8_t _xor_high_bit, uint8_t _intlv_bits, uint8_t _intlv_match)
+        : _start(_start), _end(_end), _policy(nullptr)
     {
-        // sanity checks
-        fatal_if(_intlv_bits && _intlv_match >= 1ULL << _intlv_bits,
-                 "Match value %d does not fit in %d interleaving bits\n",
-                 _intlv_match, _intlv_bits);
+        if (_intlv_bits) {
+            fatal_if(_intlv_match >= 1ULL << _intlv_bits,
+                     "Match value %d does not fit in %d interleaving bits\n",
+                     _intlv_match, _intlv_bits);
 
-        // ignore the XOR bits if not interleaving
-        if (_intlv_bits && _xor_high_bit) {
-            if (_xor_high_bit == _intlv_high_bit) {
-                fatal("XOR and interleave high bit must be different\n");
-            }  else if (_xor_high_bit > _intlv_high_bit) {
-                if ((_xor_high_bit - _intlv_high_bit) < _intlv_bits)
-                    fatal("XOR and interleave high bit must be at least "
-                          "%d bits apart\n", _intlv_bits);
-            } else {
-                if ((_intlv_high_bit - _xor_high_bit) < _intlv_bits) {
-                    fatal("Interleave and XOR high bit must be at least "
-                          "%d bits apart\n", _intlv_bits);
+            // ignore the XOR bits if not interleaving
+            if (_xor_high_bit) {
+                if (_xor_high_bit == _intlv_high_bit) {
+                    fatal("XOR and interleave high bit must be different\n");
+                } else if (_xor_high_bit > _intlv_high_bit) {
+                    if ((_xor_high_bit - _intlv_high_bit) < _intlv_bits) {
+                        fatal("XOR and interleave high bit must be at least "
+                              "%d bits apart\n",
+                              _intlv_bits);
+                    }
+                } else {
+                    if ((_intlv_high_bit - _xor_high_bit) < _intlv_bits) {
+                        fatal("Interleave and XOR high bit must be at least "
+                              "%d bits apart\n",
+                              _intlv_bits);
+                    }
                 }
             }
-        }
 
-        for (auto i = 0; i < _intlv_bits; i++) {
-            uint8_t bit1 = _intlv_high_bit - i;
-            Addr mask = (1ULL << bit1);
-            if (_xor_high_bit) {
-                uint8_t bit2 = _xor_high_bit - i;
-                mask |= (1ULL << bit2);
+            std::vector<Addr> masks(_intlv_bits);
+            for (auto i = 0; i < _intlv_bits; i++) {
+                uint8_t bit1 = _intlv_high_bit - i;
+                Addr mask = (1ULL << bit1);
+                if (_xor_high_bit) {
+                    uint8_t bit2 = _xor_high_bit - i;
+                    mask |= (1ULL << bit2);
+                }
+                masks[_intlv_bits - i - 1] = mask;
             }
-            masks[_intlv_bits - i - 1] = mask;
+            _policy = std::make_shared<MaskedInterleavingPolicy>(masks,
+                                                                 _intlv_match);
         }
     }
 
+    /**
+     * Create a simple address range without interleaving.
+     *
+     * @param _start Start address of the range.
+     * @param _end End address of the range.
+     *
+     * @ingroup api_addr_range
+     */
     AddrRange(Addr _start, Addr _end)
-        : _start(_start), _end(_end), intlvMatch(0)
+        : _start(_start), _end(_end), _policy(nullptr)
     {}
 
     /**
@@ -281,7 +448,86 @@ class AddrRange
      *
      * @ingroup api_addr_range
      */
-    bool interleaved() const { return masks.size() > 0; }
+    bool
+    interleaved() const
+    {
+        return _policy != nullptr;
+    }
+
+    /**
+     * Determine if this range is a sparse range.
+     *
+     * @return true if sparse
+     *
+     * @ingroup api_addr_range
+     */
+    bool
+    isSparse() const
+    {
+        return !_chunks.empty();
+    }
+
+    /**
+     * Get the sub-ranges of this address range.
+     *
+     * If the address range is not sparse, it will return a vector
+     * containing a single element which is the address range itself.
+     *
+     * @return Vector of valid address chunks (start, end).
+     *
+     * @ingroup api_addr_range
+     */
+    std::vector<std::pair<Addr, Addr>>
+    subRanges() const
+    {
+        if (isSparse()) {
+            return _chunks;
+        }
+        return std::vector<std::pair<Addr, Addr>>{{_start, _end}};
+    }
+
+    /**
+     * Decompose the address range into a vector of simple address ranges.
+     *
+     * If the address range is sparse, it will return a vector where each
+     * element corresponds to a contiguous valid chunk of the sparse range.
+     * If the range has an interleaving policy, each decomposed range will
+     * strictly inherit the policy configuration.
+     *
+     * @return Vector of decomposed address ranges.
+     *
+     * @ingroup api_addr_range
+     */
+    std::vector<AddrRange>
+    decompose() const
+    {
+        if (_chunks.empty()) {
+            return {*this};
+        }
+
+        std::vector<AddrRange> decomposed;
+        for (const auto &chunk : _chunks) {
+            if (!_policy) {
+                decomposed.emplace_back(chunk.first, chunk.second);
+            } else {
+                if (auto masked =
+                        std::dynamic_pointer_cast<MaskedInterleavingPolicy>(
+                            _policy)) {
+                    decomposed.emplace_back(chunk.first, chunk.second,
+                                            masked->getMasks(),
+                                            masked->getMatch());
+                } else if (auto modulo = std::dynamic_pointer_cast<
+                               ModuloInterleavingPolicy>(_policy)) {
+                    decomposed.emplace_back(
+                        chunk.first, chunk.second, modulo->getStripes(),
+                        modulo->getMatch(), modulo->getLowBit());
+                } else {
+                    panic("Unknown policy type in AddrRange::decompose");
+                }
+            }
+        }
+        return decomposed;
+    }
 
     /**
      * Determing the interleaving granularity of the range.
@@ -293,13 +539,8 @@ class AddrRange
     uint64_t
     granularity() const
     {
-        if (interleaved()) {
-            auto combined_mask = 0;
-            for (auto mask: masks) {
-                combined_mask |= mask;
-            }
-            const uint8_t lowest_bit = ctz64(combined_mask);
-            return 1ULL << lowest_bit;
+        if (_policy) {
+            return _policy->granularity();
         } else {
             return size();
         }
@@ -313,7 +554,14 @@ class AddrRange
      *
      * @ingroup api_addr_range
      */
-    uint32_t stripes() const { return 1ULL << masks.size(); }
+    uint32_t
+    stripes() const
+    {
+        if (_policy) {
+            return _policy->stripes();
+        }
+        return 1;
+    }
 
     /**
      * Get the size of the address range. For a case where
@@ -325,7 +573,22 @@ class AddrRange
     Addr
     size() const
     {
-        return (_end - _start) >> masks.size();
+        if (!isSparse()) {
+            if (_policy) {
+                return _policy->size(_start, _end);
+            }
+            return _end - _start;
+        }
+
+        Addr total = 0;
+        for (const auto &chunk : _chunks) {
+            if (_policy) {
+                total += _policy->size(chunk.first, chunk.second);
+            } else {
+                total += (chunk.second - chunk.first);
+            }
+        }
+        return total;
     }
 
     /**
@@ -345,36 +608,36 @@ class AddrRange
     /**
      * Get the end address of the range.
      *
+     * If the address range is sparse, it will return the end address
+     * of the last sub-range. If the address range is interleaved, it will
+     * return the end address of the last interleaved chunk.
+     *
      * @ingroup api_addr_range
      */
     Addr end() const { return _end; }
 
     /**
-     * Get a string representation of the range. This could
-     * alternatively be implemented as a operator<<, but at the moment
-     * that seems like overkill.
+     * Get a string representation of the range.
      *
      * @ingroup api_addr_range
      */
     std::string
     to_string() const
     {
-        if (interleaved()) {
-            std::string str;
-            for (unsigned int i = 0; i < masks.size(); i++) {
-                str += " ";
-                Addr mask = masks[i];
-                while (mask) {
-                    auto bit = ctz64(mask);
-                    mask &= ~(1ULL << bit);
-                    str += csprintf("a[%d]^", bit);
-                }
-                str += csprintf("\b=%d", bits(intlvMatch, i));
+        std::string s;
+        if (isSparse()) {
+            s = csprintf("Sparse[%#llx:%#llx]", _start, _end);
+            for (const auto &r : _chunks) {
+                s += csprintf(":[%#llx:%#llx]", r.first, r.second);
             }
-            return csprintf("[%#llx:%#llx]%s", _start, _end, str);
         } else {
-            return csprintf("[%#llx:%#llx]", _start, _end);
+            s = csprintf("[%#llx:%#llx]", _start, _end);
         }
+
+        if (_policy) {
+            s += _policy->to_string();
+        }
+        return s;
     }
 
     /**
@@ -390,8 +653,21 @@ class AddrRange
     bool
     mergesWith(const AddrRange& r) const
     {
-        return r._start == _start && r._end == _end &&
-            r.masks == masks;
+        bool same_policy = false;
+        if (!_policy && !r._policy) {
+            same_policy = true;
+        } else if (_policy && r._policy) {
+            same_policy = _policy->canMerge(r._policy);
+        }
+
+        if (isSparse() != r.isSparse()) {
+            return false;
+        }
+        if (isSparse() && _chunks != r._chunks) {
+            return false;
+        }
+
+        return r._start == _start && r._end == _end && same_policy;
     }
 
     /**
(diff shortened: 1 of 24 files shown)


powerjg:
Output from `/proc/meminfo` and early printk:

```
[    0.000000] BIOS-provided physical RAM map:
[    0.000000] BIOS-e820: [mem 0x0000000000000000-0x000000000009fbff] usable
[    0.000000] BIOS-e820: [mem 0x000000000009fc00-0x00000000000fffff] reserved
[    0.000000] BIOS-e820: [mem 0x0000000000100000-0x00000000bfffffff] usable
[    0.000000] BIOS-e820: [mem 0x00000000ffff0000-0x00000000ffffffff] reserved
[    0.000000] BIOS-e820: [mem 0x0000000100000000-0x000000083fffffff] usable

...

In after_boot.sh...
Interactive mode: false
Starting gem5 init... trying to read run script file via readfile.
Running script from gem5-bridge stored in /tmp/script
MemTotal:       32868444 kB
MemFree:        32626468 kB
MemAvailable:   32429096 kB
Buffers:           11316 kB
Cached:           114844 kB
SwapCached:            0 kB
Active:           141272 kB
Inactive:          24028 kB
Active(anon):      39660 kB
Inactive(anon):        0 kB
Active(file):     101612 kB
Inactive(file):    24028 kB
Unevictable:           0 kB
Mlocked:               0 kB
SwapTotal:             0 kB
SwapFree:              0 kB
Zswap:                 0 kB
Zswapped:              0 kB
Dirty:              1712 kB
Writeback:           148 kB
AnonPages:         39188 kB
Mapped:            53504 kB
Shmem:               368 kB
KReclaimable:       9876 kB
Slab:              32444 kB
SReclaimable:       9876 kB
SUnreclaim:        22568 kB
KernelStack:        1760 kB
PageTables:         1780 kB
SecPageTables:         0 kB
NFS_Unstable:          0 kB
Bounce:                0 kB
WritebackTmp:          0 kB
CommitLimit:    16434220 kB
Committed_AS:     176872 kB
VmallocTotal:   34359738367 kB
VmallocUsed:        8736 kB
VmallocChunk:          0 kB
Percpu:              864 kB
HardwareCorrupted:     0 kB
AnonHugePages:         0 kB
ShmemHugePages:        0 kB
ShmemPmdMapped:        0 kB
FileHugePages:         0 kB
FilePmdMapped:         0 kB
Unaccepted:            0 kB
HugePages_Total:       0
HugePages_Free:        0
HugePages_Rsvd:        0
HugePages_Surp:        0
Hugepagesize:       2048 kB
Hugetlb:               0 kB
DirectMap4k:       40960 kB
DirectMap2M:    33513472 kB
```

Tests to see if the address mapping still works. Here are some of the stats. Each channel has about the same page hit rate, and nearly the same read/write requests. Also, it looks like the accesses are mostly distributed evenly across the banks. This test was run with kvm and using the timing CPU after boot. I compared to the old version and things look *similar*, but not exactly the same.

```
board.memory.mem_ctrl0.dram.readRowHitRate        60.72                       # Row buffer hit rate for reads (Ratio)
board.memory.mem_ctrl1.dram.readRowHitRate        61.34                       # Row buffer hit rate for reads (Ratio)

board.memory.mem_ctrl0.readReqs                274240                       # Number of read requests accepted (Count)
board.memory.mem_ctrl1.readReqs                263593                       # Number of read requests accepted (Count)
board.memory.mem_ctrl0.writeReqs               102066                       # Number of write requests accepted (Count)
board.memory.mem_ctrl1.writeReqs                98626                       # Number of write requests accepted (Count)

board.memory.mem_ctrl0.dram.perBankRdBursts::0         7183                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::1         7022                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::2         9854                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::3         7266                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::4         9030                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::5         6440                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::6         9289                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::7         7044                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::8        10356                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::9         8481                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::10         7521                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::11         7805                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::12         9397                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::13         9463                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::14         6371                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::15         9020                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::16         9224                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::17         7284                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::18         8310                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::19         8686                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::20         9741                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::21         9078                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::22         7825                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::23        10640                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::24         8991                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::25         8475                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::26         8044                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::27         9164                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::28         9005                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::29         8571                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::30         8445                       # Per bank write bursts (Count)
board.memory.mem_ctrl0.dram.perBankRdBursts::31         9304                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::0         6679                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::1         7050                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::2         9494                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::3         7271                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::4         8881                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::5         6021                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::6         8711                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::7         6788                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::8        10239                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::9         8221                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::10         7152                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::11         7612                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::12         8562                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::13         8361                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::14         6155                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::15         8700                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::16         8889                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::17         6783                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::18         7659                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::19         8377                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::20         9186                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::21         8646                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::22         7709                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::23        10212                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::24         8902                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::25         8328                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::26         7776                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::27         9053                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::28         8856                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::29         8055                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::30         8337                       # Per bank write bursts (Count)
board.memory.mem_ctrl1.dram.perBankRdBursts::31         9161                       # Per bank write bursts (Count)
```

powerjg:
I've tested booting x86 with dual channel + sparse ranges on both KVM and timing CPUs. I've also tested booting Arm with dual channel (not sparse). Everything seems to be working. I know it's a big change. I still would like to cleanup the commits. Please provide any feedback before I do, though!

powerjg:
@abmerop FYI, I did not update the ViperBoard.

powerjg:
@abmerop or @mattsinc any idea why this would cause the SE GPU script to fail? I would love a pointer... It's quite difficult for me to run those tests locally. I don't have access to docker easily.

The failure is

```
gem5.opt: src/mem/ruby/common/NetDest.cc:55: void gem5::ruby::NetDest::add(gem5::ruby::MachineID): Assertion `bitIndex(newElement.num) < m_bits[vecIndex(newElement)].getSize()' failed.
```

Given this is SE mode, I don't know what would have changed in this PR. All ranges are simple uninterleaved ranges, from what I can tell.

abmerop:
> @abmerop or @mattsinc any idea why this would cause the SE GPU script to fail? I would love a pointer... It's quite difficult for me to run those tests locally. I don't have access to docker easily.
> 
> The failure is
> 
> ```
> gem5.opt: src/mem/ruby/common/NetDest.cc:55: void gem5::ruby::NetDest::add(gem5::ruby::MachineID): Assertion `bitIndex(newElement.num) < m_bits[vecIndex(newElement)].getSize()' failed.
> ```
> 
> Given this is SE mode, I don't know what would have changed in this PR. All ranges are simple uninterleaved ranges, from what I can tell.

Not sure off the top of my head.  If Matt has no idea either I can take a look Sunday at the latest.  Have you tested if GPUFS has the same issue?  I kind of expect it would given they use the same Ruby protocol

mattsinc:
> > @abmerop or @mattsinc any idea why this would cause the SE GPU script to fail? I would love a pointer... It's quite difficult for me to run those tests locally. I don't have access to docker easily.
> > The failure is
> > ```
> > gem5.opt: src/mem/ruby/common/NetDest.cc:55: void gem5::ruby::NetDest::add(gem5::ruby::MachineID): Assertion `bitIndex(newElement.num) < m_bits[vecIndex(newElement)].getSize()' failed.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Given this is SE mode, I don't know what would have changed in this PR. All ranges are simple uninterleaved ranges, from what I can tell.
> 
> Not sure off the top of my head. If Matt has no idea either I can take a look Sunday at the latest. Have you tested if GPUFS has the same issue? I kind of expect it would given they use the same Ruby protocol

@v-ramadas : I think we've seen this before when we were changing the HBM addressing scheme for @TomXia 's patches?  But I can't recall what we changed.

powerjg:
> Have you tested if GPUFS has the same issue?

No, but I'd like to. Do we have prebuilt disks/kernels/workloads? I can't seem to find them on resources.gem5.org.

powerjg:
I fixed the issue with the HBMStack. I hope that fixes the GPU issue, but I doubt it.

mattsinc:
> > > @abmerop or @mattsinc any idea why this would cause the SE GPU script to fail? I would love a pointer... It's quite difficult for me to run those tests locally. I don't have access to docker easily.
> > > The failure is
> > > ```
> > > gem5.opt: src/mem/ruby/common/NetDest.cc:55: void gem5::ruby::NetDest::add(gem5::ruby::MachineID): Assertion `bitIndex(newElement.num) < m_bits[vecIndex(newElement)].getSize()' failed.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Given this is SE mode, I don't know what would have changed in this PR. All ranges are simple uninterleaved ranges, from what I can tell.
> > 
> > 
> > Not sure off the top of my head. If Matt has no idea either I can take a look Sunday at the latest. Have you tested if GPUFS has the same issue? I kind of expect it would given they use the same Ruby protocol
> 
> @v-ramadas : I think we've seen this before when we were changing the HBM addressing scheme for @TomXia 's patches? But I can't recall what we changed.

@v-ramadas , @TomXia reminder to please chime in here

abmerop:
> I fixed the issue with the HBMStack. I hope that fixes the GPU issue, but I doubt it.

Yea, I am now seeing an issue on the first PCI access. I haven't done too much digging yet, but how does this PR handle address ranges like PCI BARs that are unknown at gem5 config time? These would be set when the simulated OS sets them at runtime. I recall a mention somewhere about moving away from Ruby[Port]'s assumption that anything outside the main memory range is assumed to be I/O. Is that part of this PR or am I confusing with a different PR?

TomXia:
> > > > @abmerop or @mattsinc any idea why this would cause the SE GPU script to fail? I would love a pointer... It's quite difficult for me to run those tests locally. I don't have access to docker easily.
> > > > The failure is
> > > > ```
> > > > gem5.opt: src/mem/ruby/common/NetDest.cc:55: void gem5::ruby::NetDest::add(gem5::ruby::MachineID): Assertion `bitIndex(newElement.num) < m_bits[vecIndex(newElement)].getSize()' failed.
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Given this is SE mode, I don't know what would have changed in this PR. All ranges are simple uninterleaved ranges, from what I can tell.
> > > 
> > > 
> > > Not sure off the top of my head. If Matt has no idea either I can take a look Sunday at the latest. Have you tested if GPUFS has the same issue? I kind of expect it would given they use the same Ruby protocol
> > 
> > 
> > @v-ramadas : I think we've seen this before when we were changing the HBM addressing scheme for @TomXia 's patches? But I can't recall what we changed.
> 
> @v-ramadas , @TomXia reminder to please chime in here

Apologies for the delay. I do recall encountering errors previously due to an incorrectly configured interleaving scheme. However, this specific assertion appears to be new and different from the ones we’ve seen before.

powerjg:
> Yea, I am now seeing an issue on the first PCI access.

I *thought* I had tested GPU FS and things seemed to work. I'll double check this.

> I haven't done too much digging yet, but how does this PR handle address ranges like PCI BARs that are unknown at gem5 config time? These would be set when the simulated OS sets them at runtime.

This should be completely unchanged. I don't think the simulator creates any address ranges on the fly for this, but I could be mistaken. If I'm wrong, do you know where the code is for this?

> I recall a mention somewhere about moving away from Ruby[Port]'s assumption that anything outside the main memory range is assumed to be I/O. Is that part of this PR or am I confusing with a different PR?

I optimized this path in the Ruby port, but the logic is unchanged.

abmerop:
> > Yea, I am now seeing an issue on the first PCI access.
> 
> I _thought_ I had tested GPU FS and things seemed to work. I'll double check this.

I currently run ~70 tests and they all fail the same way. It seems to not be finding a peer port (based on invalid this pointers):

```
#0  0x000055555943f286 in gem5::RequestPort::sendAtomic (this=0x348, pkt=0x55555f387c30) at src/mem/port.hh:556
#1  0x000055555b8ad4bb in gem5::ruby::AbstractController::recvAtomic (this=0x0, pkt=0x55555f387c30) at src/mem/ruby/slicc_interface/AbstractController.cc:422
#2  0x000055555b511db5 in gem5::ruby::RubyPort::MemResponsePort::recvAtomic (this=0x55555ea4ad80, pkt=0x55555f387c30) at src/mem/ruby/system/RubyPort.cc:393
#3  0x000055555b8bb040 in gem5::AtomicRequestProtocol::send (this=0x55555d969ad8, peer=0x55555ea4adc0, pkt=0x55555f387c30) at src/mem/protocol/atomic.cc:54
#4  0x000055555943f2b1 in gem5::RequestPort::sendAtomic (this=0x55555d969a98, pkt=0x55555f387c30) at src/mem/port.hh:556
#5  0x0000555559d70128 in gem5::BaseKvmCPU::KVMCpuPort::submitIO (this=0x55555d969a98, pkt=0x55555f387c30) at src/cpu/kvm/base.cc:194
...
```

powerjg:
I am able to reproduce the error with GPU FS. So, I'll work on debugging it.

However, it looks like something is misconfigured on the disk image. I'm seeing this output.

```
Can't open /dev/gem5_bridge: No such file or directory
--> Make sure the gem5_bridge device driver has been properly inserted into the kernel. Otherwise, sudo access required to perform address-mode ops when linking against m5 library.
3+0 records in
3+0 records out
3072 bytes (3.1 kB, 3.0 KiB) copied, 0.00599909 s, 512 kB/s
[   14.075923] amdgpu 0000:00:08.0: amdgpu: Unhandled NPS info table 0.0
No GPU available.
tensor([[0.5402, 0.0949, 0.8380],
        [0.6795, 0.6509, 0.5943],
        [0.0825, 0.1875, 0.7475],
        [0.2452, 0.8151, 0.7634],
        [0.5808, 0.5483, 0.3453]])
Can't open /dev/gem5_bridge: No such file or directory
--> Make sure the gem5_bridge device driver has been properly inserted into the kernel. Otherwise, sudo access required to perform address-mode ops when linking against m5 library.
```

powerjg:
Thanks for the help! I found the issue in GPUFS. I hope that fixes the SE GPU issue too. I now get the same behavior on the develop branch and on this branch.

powerjg:
🎉  everything is passing. 

@giactra, @abmerop, @BobbyRBruce let me know if there's anything I can do to make reviewing easier or if you have any questions.

src/base/addr_range.hh
```diff
@@ -89,69 +92,65 @@ class AddrRange
...
   public:
-
     /**
+     * Default constructor, creates an invalid address range.
+     *
      * @ingroup api_addr_range
      */
-    AddrRange()
-        : _start(1), _end(0), intlvMatch(0)
-    {}
```
BobbyRBruce, line 151: Is there a reason we'd ever want  to use this constructor? Shouldn't the default constructor just be disabled?

src/base/addr_range.hh
```diff
@@ -182,13 +181,124 @@ class AddrRange
...
+     * contiguous, such as the I/O hole in x86.
+     * This constructor only enables "simple" sparse ranges, i.e. no
+     * interleaving.
+     *
+     * @param ranges Vector of valid address chunks (start, end).
+     *               Must be sorted and non-overlapping.
+     *
+     * @ingroup api_addr_range
+     */
+    AddrRange(const std::vector<std::pair<Addr, Addr>> &ranges)
```
BobbyRBruce, line 236: It's fine this way, so not a hard requirement, but just as an alternative implementation: Would it not be better here to have this be `std::vector<std::pair<Addr, size_t>> &ranges`?

It saves any issues of people putting the beginning of the range before the end in the pairs, and makes figuring out the size of the of the logical  memory simple (sum all the sizes). I think you could also create a very simple algorithm for translating physical to logical addresses this way too. 

You'd only need to worry about the case where two ranges overlap (in which case  they should be one singular range), though this is still an issue with the current pair<start, end> design.

src/python/m5/params/param_types.py
```diff
@@ -658,6 +658,124 @@ def is_subset(self, addr_range):
...
+            int(self.start),
+            int(self.end),
+            int(self.stripes),
+            int(self.stripeMatch),
+            int(self.intlvLowBit),
+        )
+
+
+class SparseModuloAddrRange(ModuloAddrRange):
+    def __init__(self, ranges, **kwargs):
```
BobbyRBruce, line 726: Typing here would be nice. It's not clear what data type should be used for "ranges"

src/base/addr_range.hh
```diff
@@ -281,7 +404,86 @@ class AddrRange
...
+     * @ingroup api_addr_range
+     */
+    std::vector<std::pair<Addr, Addr>>
+    subRanges() const
+    {
+        if (isSparse()) {
+            return _chunks;
+        }
+        return std::vector<std::pair<Addr, Addr>>{{_start, _end}};
+    }
```
BobbyRBruce, line 443: So, why is the addr_range being sparse or not so important; you to design everything so differently dependent on that. This code here just models a non-sparse address range as a sparse one with just one chunk. Can't we just have "chunks" be the central data structure here with just non-sparse case just having `Len(chunks) == 1`. It seems simpler to me

src/base/addr_range.hh
```diff
@@ -281,7 +404,86 @@ class AddrRange
...
+     * Determine if this range is a sparse range.
+     *
+     * @return true if sparse
+     *
+     * @ingroup api_addr_range
+     */
+    bool
+    isSparse() const
+    {
+        return !_chunks.empty();
```
BobbyRBruce, line 423: what would it mean here if `len(_chunks) == 1`?

BobbyRBruce (commented):
No major complaints here. I have some comments/questions/suggestions but everything seems, on the whole, ok.

src/base/addr_range.hh
BobbyRBruce: I'll expand on this to say: buy that I mean, can we have a just one chunk?

src/base/addr_range.hh
powerjg: It's used in a few places in the codebase. I believe it's also used in the AddrRangeMap class (but I could be mistaken there).

src/base/addr_range.hh
powerjg: Yeah, except that performance is quite important here. I don't feel comfortable making the common path through the sparse logic.

src/base/addr_range.hh
powerjg: I think your proposal is fine, but I believe this design was chosen because it matches how AddrRanges were (and are) defined. They have always been defined with a begin and an end. The begin + size is a specialization. I'd prefer to keep the logic as it is.

src/base/addr_range.hh
BobbyRBruce: Fair enough.

src/base/addr_range.hh
BobbyRBruce: Assumed this was the case. Thank you for the response. I'm ok with beginning and end since it's the de-facto historic way of doing it

src/base/addr_range.hh
BobbyRBruce: I took a look into this. It’s only used in two places in the codebase: ["src/mem/backdoor.hh" at line 96](https://github.com/gem5/gem5/blob/7a2b0e413d06c5ce7097104abef3b1d9eaabca91/src/mem/backdoor.hh#L96) and ["src/dev/amdgpu/amdgpu_vm.cc" at line 54](https://github.com/gem5/gem5/blob/7a2b0e413d06c5ce7097104abef3b1d9eaabca91/src/dev/amdgpu/amdgpu_vm.cc). In both cases, it’s used as a "Not set" option. `contains()` and `valid` always return `false`, which is what you’d expect from such an option. It would probably be better to have a special constructor for the "Not set" case to make it safer and force the user to state explicitly it's what they want. If we were to refactor this I'd suggest a factory like `AddrRange::not_set()` with this default constructor disabled. 

However, the constructors’ usage in "backdoor.hh" dates back 7 years, and it's technically an API function we'd need to deprecate before removing, so I won't push for it in this PR.

abmerop (commented):
No complaints from me.  All our GPU tests pass.

powerjg:
I have rebased these commits and squashed/reordered to clean things up. It is READY FOR REVIEW!

The old branch can be found here: https://github.com/powerjg/gem5/tree/old-refactor-addr-range

powerjg:
@giactra I would like to get this merged this week, if possible.

powerjg:
clang-format is passing locally. I am not sure what the problem is.

src/base/addr_range.hh
```diff
@@ -814,51 +1190,61 @@ operator-(const AddrRangeList &base, const AddrRangeList &to_exclude)
...
  * @ingroup api_addr_range
  */
-inline AddrRange
-RangeEx(Addr start, Addr end)
-{
-    return AddrRange(start, end);
-}
+static inline AddrRange
+RangeSize(Addr start, Addr size)
+{ return AddrRange(start, start + size); }
```
abmerop: example

abmerop (commented):
Maybe run clang-format directly? The clang-format hook runs for me if I soft reset to HEAD~1 and readd everything. It is mostly changing the one line method definitions

giactra (commented):
Hi @powerjg, thank you so much for this long-awaited change. I am really looking forward for this to be merged and I plan to dedicate more time in the reviewing process to fast track this.

I didn't look into the details of the patch, just into the overall design. These are my impressions:

1) It looks in good shape (other than some clang-format quirks)
2) I approve the interleaving policy and IMO once it gets a quick functionality check it is ready to be merged
3) About sparse ranges, this is also strongly needed in gem5. However as I believe I mentioned some time ago, I don't believe having sparse ranges implemented within the single AddrRange is the way to go.

To me it is much more maintainable if we "simply" tweaked AddrRange components to support vectors of AddrRange instead of encoding vectors of AddrRange within the AddrRange iteself.

In this way we keep the crucial AddrRange class simple, and rely more on composition of easy to reason building blocks. I believe this will make things much more maintenable. What do you think? Am I missing something?

powerjg:
Hi @giactra,

Thanks for the high-level architectural feedback! I completely understand the appeal of keeping `AddrRange` as a simple, singular building block and using composition (`std::vector<AddrRange>`) at the component level.

During the early stages of this design, I actually prototyped a composition approach, but ran into a few hard engineering constraints that forced the current embedded structure:

1. **Interleaving / Sparsity Interaction:** The implementation introduces complex sparse-interleaved types (like `SparseModuloAddrRange`). The interleaving policies need to evaluate a contiguous logical address space across all valid chunks while gracefully ignoring the holes. If we decouple these into a vector of independent `AddrRange` objects, the interleaving math becomes incredibly fragmented and forces individual components (like crossbars or controllers) to manually stitch the chunks together to compute channel matches.
2. **Interface Stability:** Moving to a vector approach would require modifying core simulator interfaces—such as changing packet routing targets and port range queries to handle collections of ranges rather than a unified pass-by-value token.

Interestingly, the current implementation actually does use composition where it matters most. For instance, in `AddrRangeMap::insert` and `PhysicalMemory::createBackingStore`, we use `r.decompose()` and `r.subRanges()` to unpack the sparse chunks and treat them as independent contiguous blocks at the structural layer. The single `AddrRange` class essentially acts as a clean interface wrapper around this behavior.

Given that this design satisfies both the encapsulation of the interleaving math and has successfully undergone rigorous functional and performance testing, changing the top-level architecture now would be incredibly costly.

What are your thoughts on merging this implementation as-is to unblock the community? If we identify areas down the road where decoupling the container makes sense, we can absolutely tackle that as an isolated, post-merge refactor.

src/base/addr_range.test.cc
```diff
@@ -1594,8 +1595,335 @@ TEST(AddrRangeDeathTest, ExcludeInterleavingRanges)
...
+    AddrRange r1(0, 300, 3, 1);
+    AddrRange r2(0, 300, 3, 2);
+
+    EXPECT_TRUE(r0.contains(0));
+    EXPECT_TRUE(r0.contains(3));
+    EXPECT_FALSE(r0.contains(1));
+    EXPECT_FALSE(r0.contains(2));
+
+    EXPECT_TRUE(r1.contains(1));
+    EXPECT_TRUE(r1.contains(4));
```
giactra, line 1629: Nit: for simmetry you should be checking all values from 0 to 5

src/base/addr_range_policy.hh
```diff
@@ -0,0 +1,544 @@
...
+    /**
+     * Get a string representation of the policy.
+     *
+     * This is just the policy information (e.g., interleaving bits). The base,
+     * and bounds is printed by the AddrRange class.
+     */
+    virtual std::string to_string() const = 0;
+};
+
+class MaskedInterleavingPolicy : public AddrMapPolicy
```
giactra, line 168: Could you add some documentation / example here on how the match works?

src/base/addr_range_policy.hh
```diff
@@ -0,0 +1,544 @@
...
+     */
+    virtual bool
+    lessThan(const std::shared_ptr<AddrMapPolicy> &other) const
+    {
+        auto ptr = other.get();
+        if (typeid(*this) != typeid(*ptr)) {
+            return typeid(*this).before(typeid(*ptr));
+        }
+        return false;
+    }
```
giactra, line 114: This seems very obscure. Could you guide me through its intended behaviour? Is it simply a way to tie-break ordering between ranges with different interleaving policies?

src/mem/abstract_mem.cc
```diff
@@ -112,7 +122,20 @@ AbstractMemory::setBackingStore(uint8_t* pmem_addr)
     // The back door can't handle interleaved memory.
     backdoor.ptr(range.interleaved() ? nullptr : pmem_addr);
```
giactra: Shouldn't we invalidate the backdoor for sparse ranges as well?

src/mem/abstract_mem.hh
```diff
@@ -117,6 +118,18 @@ class AbstractMemory : public ClockedObject
     // Pointer to host memory used to implement this memory
     uint8_t* pmemAddr;
 
+    // Map of address ranges to host memory pointers. Used when this memory has
+    // a sparse address range.
+    // Note: Either pmemAddr or pmemMap should be used, but not both.
+    AddrRangeMap<uint8_t *> pmemMap;
+
+    // Whether the memory is sparse
+    const bool isSparse;
```
giactra: We could really collapse this into

pmemMap only.

I don't see the reason to keep pmemAddr (other than for some sort of compatibility), considering it increases the chances of using the wrong pointer. isSparse could be inferred from the fact that pmemMap.size() != 1. (I am fine on keeping it anyway if it is on the critical path)

src/mem/physical.cc
```diff
@@ -113,6 +113,17 @@ PhysicalMemory::PhysicalMemory(const std::string &_name,
...
+            if (m->getAddrRange().isSparse()) {
+                const auto &sub_ranges = m->getAddrRange().subRanges();
+                panic_if(sub_ranges.empty(),
+                         "Sparse memory range has no subranges\n");
+                validAddrMap.insert(validAddrMap.end(), sub_ranges.begin(),
+                                    sub_ranges.end());
+            } else {
+                validAddrMap.emplace_back(m->getAddrRange().start(),
+                                          m->getAddrRange().end());
+            }
```
giactra: Shouldn't subRanges work for a non sparse address range as well?

giactra:
> Given that this design satisfies both the encapsulation of the interleaving math and has successfully undergone rigorous functional and performance testing, changing the top-level architecture now would be incredibly costly.
> 
> What are your thoughts on merging this implementation as-is to unblock the community? If we identify areas down the road where decoupling the container makes sense, we can absolutely tackle that as an isolated, post-merge refactor.

Sounds good to me. To me a possible (I haven't given too much thought to it) future development could be to define a smaller AddrRange type which is the contiguous version of the new AddrRange. To be used for simple components where we don't expect interleaving

src/base/addr_range.test.cc
powerjg: Great suggestion. I have updated TEST(AddrRangeTest, ModuloSimple) so that r0, r1, and r2 all verify values 0 through 5 for full symmetry.

src/base/addr_range_policy.hh
powerjg: You hit the nail on the head. For STL containers like AddrRangeMap (and sets/maps using AddrRange as a key), strict weak ordering (operator<) requires a deterministic tie-breaker when comparing two ranges that have identical start and end addresses but use different policy classes. I have added a clarifying C++ comment to lessThan explaining this behavior.

src/base/addr_range_policy.hh
powerjg: Oops. This was accidentally dropped. Agreed. I have restored and adapted the detailed Doxygen documentation from upstream gem5's legacy AddrRange constructor, which provides a concrete multi-bit mask example explaining how the XOR parity calculations form sel to compare against intlvMatch.

src/mem/abstract_mem.cc
powerjg: Yes, you are completely correct. PhysicalMemory::createBackingStore allocates separate host memory buffers for each subrange of a sparse range and calls setBackingStore multiple times. Previously, backdoor.ptr() got overwritten with the host buffer of the last subrange, which breaks MemBackdoor's contiguous linear offset calculation (paddr - range.start()). I have updated setBackingStore to invalidate the backdoor for both interleaved and sparse ranges: backdoor.ptr((range.interleaved() || range.isSparse()) ? nullptr : pmem_addr);.

src/mem/abstract_mem.hh
powerjg: You are completely right on the intuition, but pmemAddr is indeed on the absolute critical path of simulation memory accesses (functional, atomic, and packet accesses). Direct pointer calculation pmemAddr + addr - range.start() is O(1), whereas searching an AddrRangeMap (an interval tree) adds measurable latency during standard contiguous memory simulations. I have added a comment in the header documenting this critical path optimization.

[Jason, not AI] IIRC, pmemAddr is directly accessed in places other than `access`, but I could be mistaken. This is performance critical, so I was hesitant to to put the map access on the critical path.

src/mem/physical.cc
powerjg: Excellent observation! AddrRange::subRanges() already returns {{_start, _end}} for non-sparse ranges, so the isSparse() check here was entirely redundant. I have removed the if/else check entirely and simplified it to unconditionally use subRanges().

powerjg (commented):
> Sounds good to me. To me a possible (I haven't given too much thought to it) future development could be to define a smaller AddrRange type which is the contiguous version of the new AddrRange. To be used for simple components where we don't expect interleaving

I thought about this, but I didn't come up with a simple way to do it. Somewhere we need polymorphism. Either we have it in the AddrRange (what I did), we make AddrRange an abstract class (this would require cascading changes all over the codebase), or we put the polymorphism at every AddrRange call site. I felt like all 3 were imperfect choices.

Now that we have a existence proof of new kinds of address ranges, refactoring this (e.g., with AI), should be straightforward. However, before we do that we need to make sure we have good test coverage, of course.

Thanks for you review. You caught a bunch of good stuff! Note that most of the responses are AI. I carefully reviewed the output though, so blame me if anything doesn't meet your expectations.

src/mem/abstract_mem.hh
giactra: Ok, I understand the critical path optimization. However I have two comments:

1) I wonder if the slowdown has been actually measured (I am not skeptical, just curious)
2) I would shift the perspective and consider the contiguous range pmemAddr usage as a sort of optimization over the common case of sparse range. This effectively means that pmemMap is **ALWAYS** valid and initialized (even in the contiguous range case). This makes debugging easier as we rely on one data structure only to gather abstract memory range info.
We simply **cache** the single pointer into pmemAddr in order for it to be used in the critical path.

This seems more robust to me. I am really scared of having a potentially uninitialized pmemMap

powerjg: > I wonder if the slowdown has been actually measured (I am not skeptical, just curious)

I did not directly test it. However, as an indirect test, https://github.com/gem5/gem5/issues/2855#issuecomment-3710977555 shows the performance with dual channel memory with sparse and not sparse address ranges. When using sparse ranges, the time spent in `AddrRangeMap` goes up by 3x. 

The `AddrRangeMap` code is not trivial, and putting it on the critical path would definitely cause a performance hit. If it's 10% or 5x, I have no idea.

> We simply cache the single pointer into pmemAddr in order for it to be used in the critical path.

This is a good approach. Let me investigate the cost of the implementation (likely low).

powerjg:
~~On the clang format thing... here is what I did... I am at a total loss on what it's issue is. Every single commit said that the presubmit clang format passed.~~

I FINALLY FIGURED IT OUT! Locally, I have a different version of clang-format. The solution is to use `pip install clang-format==18.1.8` and then run the following.

`git rebase -i develop` and set all commits to `edit`. Then,

1. **Undo the current commit** (keeping all changes in the working tree):
   ```bash
   git reset HEAD~1
   ```

2. **Stage only the files that were part of the original commit**:
   ```bash
   git diff --name-only HEAD ORIG_HEAD | xargs git add
   ```
   *(This dynamically gets the list of files changed between the parent commit `HEAD` and the original commit `ORIG_HEAD`, and stages only those files).*

3. **Re-commit the changes** (triggering the hooks):
   ```bash
   git commit -c ORIG_HEAD
   ```

4. **Handle pre-commit hook modifications** (if any):
   If the clang-format hook automatically formats files, stage those formatting updates and commit:
   ```bash
   git add -u
   git commit -c ORIG_HEAD
   ```

5. **Continue the rebase**:
   ```bash
   git rebase --continue
   ```
   
Furthermore, the error says:
 
```
Command 'git-clang-format --style=file --commit=c2c505f97bfae67f71769505adea56bfe104f0c7' returned 1
```
 
This is what I see locally. It completes with exit code 0.
```
 % git-clang-format --style=file --commit=c2c505f97bfae67f71769505adea56bfe104f0c7
clang-format did not modify any files
```

powerjg:
@giactra thanks again for the diligent reviews. I believe I have addressed all of the concerns (and clang-format is finally passing!!!). I tried to mark things resolved where appropriate, and my intention is that everything is addressed. Let me know otherwise.

src/base/addr_range.hh
```diff
@@ -325,7 +549,21 @@ class AddrRange
...
+        } else {
+            total = _end - _start;
+        }
+
+        if (_policy) {
+            // Policy assumes a contiguous logic space of 'total' size
+            // returning 'size' usually just divides by stripes
+            return _policy->size(0, total);
+        }
+        return total;
```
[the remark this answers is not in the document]
powerjg: Done.

src/python/m5/params/param_types.py
```diff
@@ -658,6 +659,129 @@ def is_subset(self, addr_range):
...
+        )
+
+    def size(self):
+        def to_compact(a):
+            return (
+                (a >> self.intlvLowBit) // self.stripes
+            ) << self.intlvLowBit | (a & ((1 << self.intlvLowBit) - 1))
+
+        return to_compact(int(self.end)) - to_compact(int(self.start))
+
```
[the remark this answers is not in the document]
powerjg: This and the following comments are addressed by disallowing ranges that are not evenly divisible.

src/base/addr_range.hh
```diff
@@ -182,13 +181,168 @@ class AddrRange
...
+     * @param end The end address of the range.
+     * @param stripes The number of stripes (channels).
+     * @param intlv_match The stripe index for this range.
+     * @param intlv_low_bit The bit position of the interleaving granularity
+     *                      (log2).
+     *
+     * @ingroup api_addr_range
+     */
+    AddrRange(Addr start, Addr end, uint32_t stripes, uint32_t intlv_match,
+              uint32_t intlv_low_bit = 0)
```
giactra, line 214: Proposal: to avoid confusion between XOR based and Module based interleaving, 

example (numbers are random:

AddrRange(0, 30, 1, 2, 3) vs -> MODULO

AddrRange(0, 30, 1, 2, 3, 4) -> XOR

would relying on a factory method make things more readable? I mean making the module constructor protected/private and have somthing like

range = AddrRange::with_modulo_intlv(...)

We could have a symmetric version for the xor interleaving, but I would keep the constructor public for backwards compatibility

With the pybind interface (python) we could have something similar or we could rely on the py::kw_only option to make sure modulo data can only be passed as a KW argument (named) therefore avoding any ambiguity

https://pybind11.readthedocs.io/en/stable/advanced/functions.html

giactra (commented):
Hi @powerjg, thanks for addressing my comments. It looks in a good state now. I am happy to approve this great change; I only have a last minute proposal which I believe will make code much more readable

src/base/addr_range.hh
powerjg: This is a great proposal. I am not going to be able to get to it for a bit (at least a week). We can wait to merge this until I get around to it, or we can merge it and I will create a new PR with this implementation. It's up to you. 🙂

giactra (approved):
I am going to approve this considering the importance of the PR. However I feel we should try to move quickly to the factory based API before users start adopting this temporary interface

PR #2862, opened 2025-12-29
arch-x86: Revert "arch-x86: implement FSGSBASE"

BobbyRBruce:
Reverts gem5/gem5#2749


BobbyRBruce:
@nmosier : Sorry for abruptly reverting this change but we found it was breaking the X86 boots (they were stalling). All the Daily X86 Boot tests timed out as a result. This should be fairly easy to reproduce locally. There are example x86 full system boots in "configs/example/gem5_library"

powerjg:
@nmosier wanted to make sure you saw this since it happened over the holiday.

nmosier:
Thanks for the reminder. I'll look into it over the next few days.

PR #2869, opened 2026-01-02
misc: Remove redundant importer.py PySource and gem5py_m5_env

harry900831:
importer.py will be dealt with the Blob, so it doesn't have to go throught the PySource and EmbeddedPython path.

https://github.com/gem5/gem5/blob/7a2b0e413d06c5ce7097104abef3b1d9eaabca91/src/python/SConscript#L445C1-L445C51


powerjg (approved):
I have to admit, the scons code here is mostly magic to me. I suppose that since the tests pass that this is OK. The only possible problem could be with using gem5 as a library (without python), but we shouldn't be using importer.py in that case. So, I guess this looks good to me.

PR #2883, opened 2026-01-11
misc: custom routing

amoghsgk:
adding parameter to network and routing logic  to routing unit with custom routing table


amoghsgk:
passed all the gem5 style requirements and tested working

PR #2891, opened 2026-01-16
arch-x86: implement FSGSBASE

nmosier:
This patch series implements the FSGSBASE x86-64 extension, fixing a critical issue with the previous patch set (#2749, reverted by #2862). This extension allows userspace to directly read/write the bases of the FS and GS segment registers via rdfsbase, rdgsbase, wrfsbase, and wrgsbase.

The first commit implements the decode logic and microcode for the new instructions. The second commit enables the extension in syscall emulation mode. The third commit exposes the CPUID feature flag for FSGSBASE, so that full system kernels can detect presence of the feature and automatically enable it.

The issue causing the previous patch set was that the implementation of the `wrcr` micro-op was not updated to allow setting CR4.FSGSBASE, resulting in a #GP fault and kernel panic when the kernel tried to do so.

### DIFF
--- a/src/arch/x86/X86ISA.py
+++ b/src/arch/x86/X86ISA.py
@@ -83,7 +83,7 @@ class X86ISA(BaseISA):
     )
     # 0000_0007h
     ExtendedFeatures = VectorParam.UInt32(
-        [0x00000000, 0x01800000, 0x00000000, 0x00000000], "feature flags"
+        [0x00000000, 0x01800001, 0x00000000, 0x00000000], "feature flags"
     )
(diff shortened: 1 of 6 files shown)


nmosier:
This time, I verified that `build/ALL/gem5.opt configs/example/gem5_library/x86-ubuntu-run.py` successfully ran to completion. For the previous buggy patch set, I had only run the default quick tests (`./tests/main.py run`) and a KVM boot + switch (configs/examples/gem5_library/x86-ubuntu-run-with-kvm.py).

powerjg (approved):

PR #2892, opened 2026-01-18
mem: Fixing logic data pointer for PrefetchInfo.

mahyarsamani:
This change fixes an error when deciding whether an object of PrefetchInfo should have a valid data pointer. This bug impacts those prefetchers that need the data stored in the cache blocks such as IndirectMemoryPrefetcher.

### DIFF
--- a/src/mem/cache/prefetch/base.cc
+++ b/src/mem/cache/prefetch/base.cc
@@ -64,7 +64,7 @@ Base::PrefetchInfo::PrefetchInfo(PacketPtr pkt, Addr addr, bool miss)
     paddress(pkt->req->getPaddr()), cacheMiss(miss)
 {
     unsigned int req_size = pkt->req->getSize();
-    if ((!write && miss) || !pkt->hasData()) {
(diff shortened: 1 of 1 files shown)


src/mem/cache/prefetch/base.cc
```diff
@@ -64,7 +64,7 @@ Base::PrefetchInfo::PrefetchInfo(PacketPtr pkt, Addr addr, bool miss)
     paddress(pkt->req->getPaddr()), cacheMiss(miss)
 {
     unsigned int req_size = pkt->req->getSize();
-    if ((!write && miss) || !pkt->hasData()) {
+    if ((!write && miss) && !pkt->hasData()) {
```
giactra, line 67: I am sure this solves some problems, but there is something weird in it and I cannot help but wonder if we aren't exposing other problems by doing this.

What happens if we don't have a valid data pointer but the prefetch is not a miss? We would still try to copy invalid data with memcpy... Can you detail better in the commit message how the IMP is affected by this?

PR #2900, opened 2026-01-20
util: Add gem5 dashboard tui

Harshil2107:
- added a gem5 dashboard TUI as a utility. This dashboard allows users to view all their gem5 processes and perform certain actions on the processes via the TUI

- Added an action to kill the selected process


util/gem5-dashboard/actions/dashboard_action.py
```diff
@@ -0,0 +1,62 @@
...
+    @property
+    @abstractmethod
+    def id(self) -> str:
+        """Unique ID for the button"""
+        pass
+
+    @property
+    def variant(self) -> str:
+        """Button style: 'primary', 'default', 'success', 'warning', 'error'"""
+        return "default"
```
BobbyRBruce, line 53: What is this for?

util/gem5-dashboard/hypercall_actions/dashboard_hypercall_request.py
```diff
@@ -0,0 +1,130 @@
...
+sock = None
+
+
+def _make_hypercall(pid: int, signal_num: int, extra_payload: dict) -> dict:
+    """
+    Core hypercall helper: opens a Unix socket, sends signal_num to pid
+    with extra_payload merged into the payload, and waits for gem5's response.
+
+    :param pid: Process ID of the target gem5 process
+    :param signal_num: Signal number to send to gem5
```
BobbyRBruce, line 37: Are we not using the name `hypercall_id`? `signal_num` implies hypercalls are always signals which isn't the case

util/gem5-dashboard/hypercall_actions/dashboard_hypercall_request.py
```diff
@@ -0,0 +1,130 @@
...
+    :param extra_payload: Additional fields to include in the JSON payload
+    :return: Dictionary containing gem5's response
+    :rtype: dict
+    """
+    global sock, socket_path
+
+    socket_path = f"/tmp/hypercall_{pid}.sock"
```
BobbyRBruce, line 44: Use `MKTMP` to make this random every time. I think using this could introduce socket collisions.

util/gem5-dashboard/hypercall_actions/dashboard_hypercall_request.py
```diff
@@ -0,0 +1,130 @@
...
+
+def get_gem5_data(pid: int) -> dict:
+    """
+    Fetch data from gem5 used by the dashboard via hypercall (signal 999).
+
+    :param pid: Process ID of the target gem5 process
+    :return: Dictionary containing gem5 data
+    :rtype: dict
+    """
+    return _make_hypercall(pid, 999, {})
```
BobbyRBruce, line 85: How is this extendable if we never pass any payload? This surely means we always get the same data upon each request?

util/gem5-dashboard/hypercall_actions/dashboard_hypercall_request.py
```diff
@@ -0,0 +1,130 @@
...
+    Send an action hypercall to gem5 (signal 998).
+
+    :param pid: Process ID of the target gem5 process
+    :param action: Action identifier for gem5 to dispatch (e.g. 'checkpoint')
+    :param arguments: Optional key-value arguments passed alongside the action
+    :return: Dictionary containing gem5's response
+    :rtype: dict
+    """
+    return _make_hypercall(
+        pid, 998, {"action": action, "arguments": arguments or {}}
```
BobbyRBruce, line 99: Why not have this by just one Hypercall (999). Isn't "get_data" essentially an action?

util/gem5-dashboard/hypercall_actions/dashboard_hypercall_request.py
```diff
@@ -0,0 +1,130 @@
...
+        sys.exit(1)
+    except OSError as e:
+        logger.error(f"File error: {str(e)}")
+        sys.exit(1)
+    except KeyboardInterrupt:
+        cleanup()
+
+
+if __name__ == "__main__":
+    main()
```
BobbyRBruce, line 130: What is this for?

util/gem5-dashboard/hypercall_actions/gem5_data_manager.py
```diff
@@ -0,0 +1,170 @@
...
+            )
+            log(f"Fetching gem5 data for PID {pid}")
+
+            # Call dashboard_hypercall_request.py as a subprocess
+            result = subprocess.run(
+                ["python3", str(script_path), "--pid", pid],
+                capture_output=True,
+                text=True,
+                timeout=5,
+            )
```
BobbyRBruce, line 135: I don't understand: why are you running each request to a gem5 process as a subprocess?

util/gem5-dashboard/hypercall_actions/gem5_data_manager.py
```diff
@@ -0,0 +1,170 @@
...
+        except subprocess.TimeoutExpired:
+            log.error(f"Timeout fetching gem5 data for PID {pid}")
+            return {}
+        except json.JSONDecodeError as e:
+            log.error(
+                f"JSON decode error for PID {pid}: {e}, response: {response}"
+            )
+            return {}
+        except FileNotFoundError:
+            log.error(f"orchestrator-request.py not found at {script_path}")
```
BobbyRBruce, line 163: There is no `orchestrator-request.py`

util/gem5-dashboard/hypercall_actions/gem5_hypercall_action.py
BobbyRBruce, line 1: This file is missing a copyright notice (same goes for the other `table_column` python files).

util/gem5-dashboard/actions/dashboard_action.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @abstractmethod
+    async def execute(self, pid: str, console: Static) -> None:
```
BobbyRBruce, line 56: I see this throughout: Why is the `pid` a string? This is always an int, isn't it?

util/gem5-dashboard/table_columns/instructions_column.py
BobbyRBruce, line 1: This file is lacking a copyright notice

util/gem5-dashboard/table_columns/instructions_column.py
```diff
@@ -0,0 +1,20 @@
...
+        Formatted string of instructions executed, or "N/A" if unavailable
+    """
+    if gem5_data is None:
+        gem5_data = {}
+
+    instructions = gem5_data.get("curr_instructions_executed", None)
+
+    if instructions is None:
+        return "N/A"
+
```
BobbyRBruce, line 19: Why is the default `None`, and if it's None it just sets it to an empty set/map to which it returns "N/A"? At the very least this is inefficient, but also, why is `gem5_data` ever `None`?

util/gem5-dashboard/table_columns/workload_id_column.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    if gem5_data is None:
+        gem5_data = {}
+
+    workload_id = gem5_data.get("workload", None)
+
+    if workload_id is None:
+        return "N/A"
+
+    return str(workload_id)
```
BobbyRBruce, line 19: Same argument here as with the instructions column: this is weird

util/gem5-dashboard/action_registry.py
```diff
@@ -0,0 +1,31 @@
...
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from actions.kill_process_action import KillAction
+
+ENABLED_ACTIONS = [
+    KillAction(),
```
BobbyRBruce, line 30: isn't this calling `KillAction`?

util/gem5-dashboard/gem5_dashboard.py
```diff
@@ -0,0 +1,216 @@
...
+                else:
+                    table.add_row(*row_data, key=pid)
+
+            except (
+                psutil.NoSuchProcess,
+                psutil.AccessDenied,
+                psutil.ZombieProcess,
+            ):
+                # Process died or we lost permission while iterating
+                continue
```
BobbyRBruce, line 199: Maybe we should throw a warning when this happens, or something to tell the user it failed

util/gem5-dashboard/actions/dashboard_action.py
```diff
@@ -0,0 +1,62 @@
...
+    @abstractmethod
+    def label(self) -> str:
+        """Text to display on the button"""
+        pass
+
+    @property
+    @abstractmethod
+    def id(self) -> str:
+        """Unique ID for the button"""
+        pass
```
BobbyRBruce, line 48: Is there any purpose to this ID?

util/gem5-dashboard/requirements.txt
```diff
@@ -0,0 +1,11 @@
...
+markdown-it-py==4.0.0
+mdit-py-plugins==0.5.0
+mdurl==0.1.2
+platformdirs==4.5.1
+psutil==7.2.1
+Pygments==2.19.2
+rich==14.2.0
+textual==7.3.0
+typing_extensions==4.15.0
+uc-micro-py==1.0.3
```
BobbyRBruce, line 11: Do we really need all these requirements? I thought all we needed with textual?

src/python/gem5/simulate/exit_handler.py
```diff
@@ -318,6 +318,81 @@ def _exit_simulation(self) -> bool:
...
+    def _process(self, simulator: "Simulator") -> None:
+        try:
+            socket_path = self._payload.get("response_socket")
+            action = self._payload.get("action", "")
+            arguments = self._payload.get("arguments")
+            if socket_path:
+                sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+                sock.connect(socket_path)
+
+                response = json.dumps({"error": f"Unknown action: {action}"})
```
BobbyRBruce, line 339: So this always return an error? Where is this action executed?

src/python/gem5/simulate/exit_handler.py
```diff
@@ -318,6 +318,81 @@ def _exit_simulation(self) -> bool:
         return False
 
 
+class gem5DashboardActionExitHandler(ExitHandler, hypercall_num=998):
+    """A handler designed to be the default for the gem5 dashboard action
+    hypercall exit event. This exit event is triggered by the dashboard's
+    hypercalls to send actions to gem5.
+
+    It will not exit the simulation loop by default.
```
BobbyRBruce, line 326: It does exit the simulation loop.

src/python/gem5/simulate/exit_handler.py
BobbyRBruce, line 1: My biggest problem with the current design is the hardcoding on the gem5 side. The current design would just have the user modify the Exit Handler every time they want to get something different from gem5.

I propose we have some kind of central registry of metrics that users can register functions. Something like:


```py
# src/python/gem5/simulate/dashboard_metrics.py
from typing import Callable, Dict, Any

_METRICS: Dict[str, Callable] = {}

def register_dashboard_metric(name: str, fn: Callable) -> None:
    _METRICS[name] = fn

def collect_dashboard_metrics(simulator, requested=None) -> Dict[str, Any]:
    names = requested or list(_METRICS.keys())
    out = {}
    for n in names:
        fn = _METRICS.get(n)
        if fn:
            try:
                out[n] = fn(simulator)
            except Exception as e:
                out[n] = {"error": str(e)}
    return out
```

and in `gem5DashboardExitHandler._process(...)` we'd use:

```py
requested = self._payload.get("metrics")  # e.g. ["workload", "host_rss_mb"]
response = json.dumps(collect_dashboard_metrics(simulator, requested))
```

This means users can extend the framework to return more data just by using the `register_dashboard_metric` in gem5. Assuming we want to check something like, the amount of memroy gem5 is using, we can add something like the following:

```py
import os
from gem5.simulate.dashboard_metrics import register_dashboard_metric

def host_rss_mb(simulator):
    # memory used by the running gem5 process
    with open(f"/proc/{os.getpid()}/status") as f:
        for line in f:
            if line.startswith("VmRSS:"):
                kb = int(line.split()[1])
                return round(kb / 1024.0, 2)
    return None

register_dashboard_metric("host_rss_mb", host_rss_mb)
```

I also proposer that the Hypercall has you pass which metrics you want as the payload, so the users/app developers can pick and choose which data it's worth gem5 getting. So a hypercall payload would be something like:

```json
{ "response_socket": "...", "metrics": ["workload", "curr_instructions_executed", "host_rss_mb"] }
```

You could also have a dashboard features that calls gem5 and have it return all the current registered metrics. We could have some sensible defaults shipped with gem5 but make it easy for users/devs to add more locally if they want to.

util/gem5-dashboard/table_column_map.py
```diff
@@ -0,0 +1,27 @@
...
+        "func": get_current_instructions,
+    },
+    {
+        "name": "Workload ID",
+        "key": "Workload ID",
+        "width": 40,
+        "func": get_workload_id,
+    },
+    {"name": "Command", "key": "Command", "func": get_command},
+]
```
BobbyRBruce, line 27: I suggest just having the dashboard represented as JSON file that is then loaded. I think requiring anyone to modify python to add a column is unnecessary. This is already pretty much JSON so why not just have a configuration file for this tool which specifies what the columns are and what data they get.

I think you could also get rid of the table_columns as you could just specify what data you want for each column here then get it as needed (from what I see these table_column files are just getters).

BobbyRBruce (changes requested):

util/gem5-dashboard/actions/dashboard_action.py
Harshil2107: This is just for styling the buttons, their primary color, etc.

util/gem5-dashboard/hypercall_actions/dashboard_hypercall_request.py
Harshil2107: I will update the argument to say hypercall num.

util/gem5-dashboard/hypercall_actions/dashboard_hypercall_request.py
Harshil2107: Because I separted the requests to two hypercalls:

1. One (the one highlighted here) just fetches all the data needed by the dash board at regular time intervals. We dont need a payload as only the ExitHandler needs to be updated to retrun the new requested value.
2. The hypercall an action will send which will have a specific payload argument to determine which action it is that is calling the hypercall.

util/gem5-dashboard/hypercall_actions/dashboard_hypercall_request.py
Harshil2107: I wanted to seperate two hypercalls as the hypercall 999 is being called every x seconds and this one is only called when an action happens. That way we have a clear seperation on the hypercall that is being periodically called and one that is action specific.

util/gem5-dashboard/hypercall_actions/gem5_data_manager.py
Harshil2107: Thanks for pointing this out, I will update this statement, it is outdated. Initially, I was using the same `orchestrator-request.py` that is already present in `util/hypercall_external_signal` to call the hypercalls.

util/gem5-dashboard/table_columns/instructions_column.py
Harshil2107: It can be `None` when the simulation is starting (downloading the resources, etc) as at that point the simulation will not return anything on the hypercall signal as it has not started yet.

util/gem5-dashboard/table_columns/workload_id_column.py
Harshil2107: It can be None when the simulation is starting (downloading the resources, etc) as at that point the simulation will not return anything on the hypercall signal as it has not started yet.

util/gem5-dashboard/requirements.txt
Harshil2107: I will double check but all be need is `psutil` and `textual` these others requirements might be used by these two libraries. I will check and reduce them to minimun needed requirements.

util/gem5-dashboard/hypercall_actions/gem5_data_manager.py
Harshil2107: I am using a subprocess because the dashboard needs to send a hypercall signal to each running gem5 instance every X seconds to receive updates. If I ran this in the same thread as the main dashboard process, it would cause the dashboard to freeze and become unresponsive, since every few seconds I would be triggering an hypercall-signal call for each gem5 instance simultaneously.

With a subprocess I can call the hypercalls and easily parse the output without needing to coordinate across multiple processes or ensure the right PID receives the correct data, something that would get complicated quickly with threading or similar approaches.

util/gem5-dashboard/hypercall_actions/dashboard_hypercall_request.py
Harshil2107: This is needed as I am calling this file via suprocess using the command `python3`, this block acts as a guard to ensure main() is called.

src/python/gem5/simulate/exit_handler.py
Harshil2107: I like this idea, but my concern is that fetching stats requires calling specific functions on the simulator (e.g. `simulator.get_instruction_count()`) or _m5 (e.g. `_m5.core.getClockFrequency()`), so a list of metric IDs alone is not sufficient. We could have users pass the function to call alongside the metric, but since we are not importing gem5 on the dashboard side the function would have to be passed as a string, meaning we would need to use something like `eval()` or `exec()` to convert it into an actual callable, which feels hacky. In any case, we would essentially be asking users to understand gem5, write the correct function to call, and register it in a specific format, at that point it seems simpler to just have them update the handler directly.

util/gem5-dashboard/hypercall_actions/gem5_data_manager.py
Harshil2107: I updated the dashboard to use async instead of subprocess

util/gem5-dashboard/actions/dashboard_action.py
Harshil2107: Updated pid to be int

Harshil2107:
A note, when we discussed these changes @BobbyRBruce you wanted to take a look at the new implementation of the way we get stats without needed to update/rebuild gem5 before I make the same changes to the dashboard action hypercall.

PR #2904, opened 2026-01-21
python: Add re to imports in gem5stats

powerjg:


powerjg:
Note: The reason this worked before is that some of the `import *` could have `re` in their namespace.

BobbyRBruce (approved):

PR #2905, opened 2026-01-21
arch-x86: build return PC correctly for calls

### LINKED ISSUE #2901: arch-x86: decoupled frontend mispredicts the majority of returns
**Describe the bug**
When the O3 CPU's decoupled frontend is disabled when running x86 code, the branch predictor mispredicts the majority of returns. This appears to be because return addresses are not calculated correctly (they are calculated as  `PC + 8`, where `PC` is the current call's PC, since the decoupled frontend doesn't initialize the instruction length (defaulted to 8 in x86's PCState, [here](https://github.com/gem5/gem5/blob/3bfb5c21e0446b9790e553877132a3a9970ca8bb/src/arch/x86/pcstate.hh#L50)).

Here's a test program to reproduce the performance issue (`test.asm`):
```asm
    global _start

foo:
    ret

_start:
    call foo
    nop
    nop
    nop
    nop
    nop
    nop
    nop
    jmp _start
```

**Affects version**
develop @ [3bfb5c2](https://github.com/gem5/gem5/commit/3bfb5c21e0446b9790e553877132a3a9970ca8bb)

**gem5 Modifications**
None

**To Reproduce**
1. Compile gem5 for x86: `scons build/X86/gem5.opt`
2. Assemble the test program: `nasm -felf64 -o test.o test.asm && ld -o test test.o`
3. Save the attached script (`test.py`)
4. Run gem5 with the attached script: `./build/X86/gem5.opt test.py --num-insts=100_000`
5. Find the number of mispredicts: `grep -F 'board.processor.cores.core.branchPred.mispredicted_0::Return' m5out/stats.txt`

**Terminal Output**
```
board.processor.cores.core.branchPred.mispredicted_0::Return        10000     99.98%     99.98% # Number of committed branches that were mispredicted. (Count)
```

**Expected behavior**
Very few, if any, returns should be mispredicted.

**Host Operating System**
Ubuntu 24.04

**Host ISA**
X86

**Compiler used**
GCC 13.3

**Additional information**
This appears to fix the issue:
```
diff --git a/src/cpu/pred/bpred_unit.cc b/src/cpu/pred/bpred_unit.cc
index db485a7409..46a9a991ac 100644
--- a/src/cpu/pred/bpred_unit.cc
+++ b/src/cpu/pred/bpred_unit.cc
@@ -213,6 +213,7 @@ BPredUnit::predict(const StaticInstPtr &inst, const InstSeqNum &seqNum,
             // In case of a call build the return address and
             // push it to the RAS.
             auto return_addr = inst->buildRetPC(pc, pc);
+            return_addr->set(pc.instAddr() + inst->size());
             ras->push(tid, *return_addr, hist->rasHistory);
 
             DPRINTF(Branch, "[tid:%i] [sn:%llu] Instr. %s was "
```

nmosier:
Fix #2901. Set a call's return PC to be the call's PC plus its (variable) instruction length. Previous behavior simply added a default +8 to the call PC, which is rarely correct for x86, resulting in consistently incorrect return address predictions.

Here are some performance stats on the SPEC CPU2017 benchmarks, collected by running the O3 CPU (using the default parameters aside from the following config) on the most representative SimPoint for each intspeed benchmark under the "train" inputs for 10/50 million instructions (most benchmarks crashed early, due to an apparent bug in the O3's decoupled frontend.

Config:
```
o3.fetchWidth = 16
o3.fetchTargetWidth = 128
o3.decoupledFrontEnd = True
o3.branchPred.conditionalBranchPred = TAGE_SC_L_64KB()
o3.branchPred.requiresBTBHit = True
```

Results:
| Benchmark | Instructions (warmup) | Return mispredict rate before fix | Reeturn mispredict rate after fix | Speedup after fix |
|---|---|---|---|---|
| 600.perlbench_s | crash | - | - | - |
| 602.gcc_s | crash | - | - | - |
| 605.mcf_s | 50M (10M) | 60.8% | 0.0% | 1.08x | 
| 620.omnetpp_s | 10M (-) | 72.5% | 27.8% | 1.10x | 
| 623.xalancbmk_s | 50M (10M) | 87.0% | 20.9% | 1.35x | 
| 625.x264_s | crash | - | - | - |
| 631.deepsjeng_s | 50M (10M) | 75.1% | 5.3% | 1.11x | 
| 641.leela_s | crash | - | - | - |
| 648.exchange2_s | 10M (-) | 60.3% | 0.6% | 1.06x | 
| 657.xz_s | 50M (10M) | 13.0% | 0.0% | 1.02x |
| average | - | 61.4% | 9.1% | 1.11x |
---
O3 cores with a better configuration would experience a better improvement from this fix. Anecdotally, an O3 core I configured to resemble an Intel Lion Cove P-core got >1.5x speedup (at least on one workload), if I remember correctly.

Change-Id: Id39d101de23e51d8ec2acffc02ef8d2222e6ca6f

### DIFF
--- a/src/arch/x86/insts/static_inst.hh
+++ b/src/arch/x86/insts/static_inst.hh
@@ -224,6 +224,7 @@ class X86StaticInst : public StaticInst
     {
         PCStateBase *ret_pc_ptr = call_pc.clone();
         ret_pc_ptr->as<PCState>().uEnd();
+        ret_pc_ptr->set(call_pc.instAddr() + size());
         return std::unique_ptr<PCStateBase>{ret_pc_ptr};
     }
 };


powerjg (approved):
I assume that you have data showing this fix improves the prediction for x86. It would be good to include that as a note on the PR (or issue). 

Also, do you think this should be backported to 25.1 as a hotfix? This seems pretty important. 

Finally, thanks for catching this and getting a fix so quickly! We need more stats/performance-based CI tests.

nmosier:
I've done some small-scale performance tests, but I can benchmark it before/after on the SPEC2017 benchmarks over the next couple days. I've converted the PR to a draft for now, and will mark it ready once I've updated the PR with the performance results.

nmosier:
I updated the PR description with some performance stats.

powerjg:
Thanks! Looks great!

PR #2906, opened 2026-01-21
cpu: Implement ITTAGE

dhschall:
Implementation of the indirect target TAGE (ITTAGE) after "A 64-Kbytes ITTAGE indirect branch predictor" from Andre Seznec. As compared to the work above this implementation does not use shared tables. Furthermore, instead of using a decicated base predictor this implementation uses the BTB as fallback. That means only indirect branches that have a changing target will use the indirect predictor

Signed-off-by: David Schall <david.schall@tum.de>


powerjg (approved):
I assume the implementation is correct ;) It looks good to me.

PR #2914, opened 2026-01-28
cpu-o3: Revert "incorrect `vtype` and `vl` generates `assert` fail in  `O3 IQ`"

BobbyRBruce:
Reverts gem5/gem5#2657

I believe gem5/gem5#2657 is causing a failure in the o3 CPU which is triggering our weekly tests to fail (see: https://github.com/gem5/gem5/actions/runs/21298731348/job/61310974117). I had previously created PR #2909 to fix this issue but discussions with @powerjg led me to the conclusion it was safer to revert this PR for now.

This specific failure in the weekly tests is due to the following assertion being raised when running the o3 CPU:

`gem5::o3::DynInst::DynInst(...) Assertion 'cpu->instcount <= 1500' failed.`

To reproduce the bug (warning, this takes a while):

```
../build/ALL/gem5.opt tests/gem5/x86_boot_tests/configs/ x86_boot_exit_run.py --cpu o3 --num-cpus 8 --mem-system classic --dram-class DualChannelDDR4_2400 --boot-type init
```

### DIFF
--- a/src/cpu/o3/inst_queue.cc
+++ b/src/cpu/o3/inst_queue.cc
@@ -1452,9 +1452,8 @@ InstructionQueue::doSquash(ThreadID tid)
             if (dest_reg->isAlwaysReady()) {
                 continue;
             }
-            if (!dependGraph.empty(dest_reg->flatIndex())) {
-                dependGraph.clearInst(dest_reg->flatIndex());
-            }
+            assert(dependGraph.empty(dest_reg->flatIndex()));
+            dependGraph.clearInst(dest_reg->flatIndex());
         }
         instList[tid].erase(squash_it--);
         ++iqStats.squashedInstsExamined;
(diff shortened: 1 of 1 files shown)


powerjg:
For documentation's sake... The `dependGraph.empty` function checks to see if `next` is empty or not. The `dependGraph.clearInst` sets `dependGraph.inst` to `NULL` (essentially just making sure `shared_ptr` reference count is decremented). 

So, the changed code:
```c++
            if (!dependGraph.empty(dest_reg->flatIndex())) {
                dependGraph.clearInst(dest_reg->flatIndex());
            }
```

Doesn't quite do what I thought it did. I believe what's happening is that we're not decrementing the `shared_ptr` reference count when we need to. However, I think it's a symptom of something else wrong if the assert is triggered (so I don't think we should simply remove the assert).

powerjg (approved):

PR #2920, opened 2026-01-30
arch-riscv: add isSerializeAfter on 1st unordered reduction µop

saul44203:
This patch adds the `isSerializeAfter` flag to the first unordered reduction µop (that's not also the last) to guarantee correctness as it initializes the reduction (`vd[0]`) with `vs1[0]` apart from accumulating `vs2`.

### DIFF
--- a/src/arch/riscv/isa/templates/vector_arith.isa
+++ b/src/arch/riscv/isa/templates/vector_arith.isa
@@ -1737,7 +1737,10 @@ template<typename ElemType>
     for (int i = 0; i < num_microops && micro_vl > 0; ++i) {
         microop = new %(class_name)sMicro<ElemType>(_machInst, micro_vl, i,
                                                     _elen, _vlen);
-        if (flags[IsSerializeAfter] && i < num_microops-1) { // ordered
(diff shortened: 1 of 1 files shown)


BobbyRBruce (approved):

powerjg (approved):
Is serialization the right flag here or should we have some other kind of dependence? Serialization seems like it's going to hurt performance.

saul44203:
Hi @powerjg. I think it´s the only mechanism right now to do this kind of inter-µop dependencies apart from creating fake data dependencies, which has it's own issues as well. I agree serializing is not the best for performance though. Is there any other mechanism that I'm missing?

powerjg:
> Hi @powerjg. I think it´s the only mechanism right now to do this kind of inter-µop dependencies apart from creating fake data dependencies, which has it's own issues as well. I agree serializing is not the best for performance though. Is there any other mechanism that I'm missing?

No, I don't think there's anything you're missing. This is strictly improving the codebase, so we might as well merge it. We can tackle the performance issue another day. We may need to add a new kind of dependency tracking, which is going to be a big undertaking.

PR #2921, opened 2026-01-30
mem-ruby: Improve performance of NetDest::resize

powerjg:
This function is on the critical path to many, many Ruby functions. It is one of the hottest functions in Ruby.

To optimize the code, we add a cache to keep a copy of a recent vector that has been allocated. Since most of the NetDest use is creating a NetDest for all machines in the RubySystem, the vector is almost always exactly the same size. Essentially, What we're doing is a simple vector copy instead of the for loop.

This works with multiple RubySystems (e.g., with the GPU) by using the RubySystem as the key to the map.

With this code, when running a simple traffic generator code the time goes from 28 seconds in NetDest::resize to less than 1 second (out of about 188 seconds in the baseline).

---

More performance details:
- When running Arm Linux boot with CHI (when testing the AddrRange change impacts #2861) I noticed that out of 2000 seconds, NetDest::resize was taking about 12% of the total time.
- In a more constrained environment, I ran a simple traffic generator test with the CHI protocol with a 3 level hierarchy where the data fit in the L3. I ran 10ms of simulated time. On the baseline NetDest::resize took 28 seconds out of 188 seconds. With this PR, it took less than 1 second. See the pprof diff below (colors represent differences). I've zoomed in just on this function.

image

Basically, with this change we see a 1.16x speedup.

### DIFF
--- a/src/mem/ruby/common/NetDest.cc
+++ b/src/mem/ruby/common/NetDest.cc
@@ -28,8 +28,10 @@
 
 #include "mem/ruby/common/NetDest.hh"
 
-#include <algorithm>
+#include <map>
+#include <vector>
 
+#include "mem/ruby/common/Set.hh"
 #include "mem/ruby/system/RubySystem.hh"
 
 namespace gem5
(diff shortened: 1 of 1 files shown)


hnpl:
Assume that the number of RubySystems is typically small (e.g., less than 10), I'm curious if using std::vector and linear search is better

powerjg:
It's typically 1, sometimes 2 if you're using the FS GPU. Do we think it will make a big difference? We're already down to basically no time in this function.

hnpl:
I think the performance gain from using vector is pretty much negligible in this case then

src/mem/ruby/common/NetDest.cc
```diff
@@ -285,14 +287,22 @@ NetDest::isElement(MachineID element) const
 void
 NetDest::resize()
 {
-    assert(m_ruby_system != nullptr);
+    static std::map<RubySystem *, std::vector<Set>> resize_cache;
```
mattsinc, line 296: Would it make more sense for this variable to be a variable declared in the class as opposed to statically declared inside this function?

mattsinc (commented):
Nice job!  Just one question

src/mem/ruby/common/NetDest.cc
powerjg: I think it's better as-is since the cache is private to this function. Nothing else in the class should reference this variable.

mattsinc: fair enough

mattsinc (approved):

PR #2922, opened 2026-02-01
mem-ruby: Allow non-power-of-two cache capacity

powerjg:
Previously, Ruby only allowed power of two sizes for the cache capacity by using bitwise operations. This changeset enables non power of two capacity by using modulo only when necessary.

### DIFF
--- a/src/mem/ruby/structures/CacheMemory.cc
+++ b/src/mem/ruby/structures/CacheMemory.cc
@@ -49,6 +49,7 @@
 #include "debug/RubyResourceStalls.hh"
 #include "debug/RubyStats.hh"
 #include "mem/cache/replacement_policies/weighted_lru_rp.hh"
+#include "mem/ruby/common/Address.hh"
 #include "mem/ruby/protocol/AccessPermission.hh"
 #include "mem/ruby/system/RubySystem.hh"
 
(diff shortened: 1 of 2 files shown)


mattsinc:
@v-ramadas : I believe this is similar to what we did internally?

mattsinc (dismissed):
In general this looks reasonable to me.  Before merging, it would be good to know what testing you did with the GPU though -- since that uses Ruby heavily.

powerjg:
> In general this looks reasonable to me.  Before merging, it would be good to know what testing you did with the GPU though -- since that uses Ruby heavily.

Honestly, no tests of the GPU. However, if the size of the cache is a power of two there should be no difference.

powerjg:
> In general this looks reasonable to me.  Before merging, it would be good to know what testing you did with the GPU though -- since that uses Ruby heavily.

Honestly, no tests of the GPU. However, if the size of the cache is a power of two there should be no difference.

powerjg:
> In general this looks reasonable to me.  Before merging, it would be good to know what testing you did with the GPU though -- since that uses Ruby heavily.

Honestly, no tests of the GPU. However, if the size of the cache is a power of two there should be no difference.

mattsinc:
> > In general this looks reasonable to me.  Before merging, it would be good to know what testing you did with the GPU though -- since that uses Ruby heavily.
> 
> Honestly, no tests of the GPU. However, if the size of the cache is a power of two there should be no difference.

I must have misunderstood your commit message then.  I thought you were saying you had tested it with the GPU.  I don't see that part of the message there anymore though, so maybe I'm just imagining things.  In any event, eventually we'll need to make sure it doesn't break anything for the GPU but I agree in theory it shouldn't.

BobbyRBruce (approved):

PR #2928, opened 2026-02-04
dev-amdgpu, mem-ruby: Adding/modifying GPU stats to track different inst types

poal023:
This PR adds in/improves several stats for the GPU (and flags to characterize instructions for these stats). Specifically, it adds:

- Characterizing 24 and 32 bit integer multiplies/operations, single and double precision FLOPs, special FP operations (log, exp, sin), and texture insts.
- Adding stat increments for each of these characterized instructions in `compute_unit.cc` as well as the total instructions (scalar or vector)
- Added shaderActiveCycles to increment more frequently as the periodic updates in the head of dev are insufficient if you want to sample at a fine granularity like every 500 GPU clock cycles. I think the stat I have right now in [this file](https://github.com/user-attachments/files/25077780/aw-stats.zip) is off by a little bit (assume 3GHz from the stdlib, (3e9 / 1e12) * 212016102 = 636048.306, this is from running square in gem5-resources)
- Modified the GPU coherence protocols to have stat increments if something was considered a read hit/miss or a write H/M
- Adds support for using Garnet networks for GPUs (see: `src/python/gem5/prebuilt/viper/viper_garnet_network.py`)

Hopefully these changes are mostly sufficient, but I have a couple of conerns/questions I'm unsure about:

1. As far as I know, the L2/TCC caches don't have the ability to check if a request was a load or store, so at the moment, accesses are categorized as both a "read and write hit/miss". The PR which has the Accelwattch Power Model (#blah) requires these changes in order to model the L2/TCC's power.
2. I'm not sure if there's also a better way to increment "read hits/misses" and "write hits/misses", as power models (e.g., McPAT and Accelwattch) require these for their cache modeling. 
3. I'm not sure if I've covered all/most instructions and have correctly categorized them into the buckets in `GPUStaticInst.py`. There could be some things missing?

EDIT: Typos, forgot to attach stats file, and tagging @mattsinc @abmerop


src/mem/ruby/structures/CacheMemory.hh
```diff
@@ -246,6 +246,14 @@ class CacheMemory : public SimObject
           statistics::Scalar m_demand_misses;
           statistics::Formula m_demand_accesses;
 
+          statistics::Scalar read_hits;
+          statistics::Scalar read_misses;
+          statistics::Formula read_accesses;
+
+          statistics::Scalar write_hits;
+          statistics::Scalar write_misses;
+          statistics::Formula write_accesses;
```
powerjg, line 255: Why not use the `tagArrayReads`/`Writes and `dataArrayReads`/`Writes`. These stats were originally designed (in part) to be used with mcpat and other similar power tools.

I don't like the idea of adding new stats, especially ones that overlap with many other stats (e.g., `m_demand_*`). I also *really* don't like having only a single protocol that uses the new stats. We need to move towards unifying the stats across protocols, not making them more different.

src/mem/ruby/structures/CacheMemory.hh
```diff
@@ -258,6 +266,10 @@ class CacheMemory : public SimObject
       // each time they are called
       void profileDemandHit();
       void profileDemandMiss();
+      void profileReadHit();
+      void profileReadMiss();
+      void profileWriteHit();
+      void profileWriteMiss();
```
powerjg, line 272: Why do you need to add these new functions?

powerjg (commented):
Given the changes to shared ruby structures the PR name message should make it clear that there are mem-ruby changes.

I have comments on that code. I didn't look at the GPU code.

src/arch/amdgpu/vega/insts/mimg.cc
```diff
@@ -41,6 +41,7 @@ namespace VegaISA
     Inst_MIMG__IMAGE_LOAD::Inst_MIMG__IMAGE_LOAD(InFmt_MIMG *iFmt)
         : Inst_MIMG(iFmt, "image_load")
     {
+        setFlag(Tex);
```
mattsinc, line 44: To my knowledge, gem5 does not support texture instructions.  Are you seeing these instructions in programs?

src/arch/amdgpu/vega/insts/vop2.cc
```diff
@@ -223,6 +223,7 @@ namespace VegaISA
         : Inst_VOP2(iFmt, "v_mul_legacy_f32")
     {
         setFlag(ALU);
+        setFlag(FpMul);
```
mattsinc, line 226: I guess the idea is adding the FPMul flag allows you to separately track stats for them?

src/gpu-compute/compute_unit.cc
```diff
@@ -927,6 +927,7 @@ ComputeUnit::exec()
 
     stats.totalCycles++;
     execCycles++;
+    shader->notifyCuActive();
```
mattsinc, line 930: what does this have to do with the rest of the commit?

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -665,11 +665,17 @@ machine(MachineType:TCC, "TCC Cache")
   }
 
   action(p_profileMiss, "pm", desc="Profile cache miss") {
+      // TODO: Fix this so all RM != WM
```
mattsinc: either fix or remove comment

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
```diff
@@ -665,11 +665,17 @@ machine(MachineType:TCC, "TCC Cache")
...
+      L2cache.profileReadMiss();
+      L2cache.profileWriteMiss();
       L2cache.profileDemandMiss();
       DPRINTF(RubyHitMiss, "in TCC miss at %#lx\n", address);
   }
 
   action(p_profileHit, "ph", desc="Profile cache hit") {
+      // TODO: Fix this so all RH != WH
```
mattsinc: same here

src/mem/ruby/structures/CacheMemory.hh
poal023: Can you elaborate on how I would use the dataArray and writeArray stats to find the number of RH/RM/WH/WM? I can imagine you use the number of times you've written to the tag array as finding the write misses (since you need to write the tag for the first time into a cache), but how would you find the other numbers? Aren't the number of DataArrayWrites just how many times you've written to the cache, irrespective if they are hits or misses? 

I implemented these stats specifically to make it explicit what they are tracking, rather than inferring them from the stats you mentioned. If there's something I'm missing/misunderstanding, please let me know.

src/mem/ruby/structures/CacheMemory.hh
poal023: Basically so that these stats can individually be incremented if there is a read request or write request. If the stats above aren't necessary, I can remove these.

src/arch/amdgpu/vega/insts/mimg.cc
poal023: While they're not in the programs I've ran, I think it's still a good idea to label them explicitly anyway.l

src/arch/amdgpu/vega/insts/vop2.cc
poal023: Yes, that's pretty much right. I want to explicitly label these instructions so when in `wavefront.cc` we check which instructions maps to which stat, having these flags make it easy to do so.

src/gpu-compute/compute_unit.cc
poal023: The function basically tries to increment `shaderActiveCycles` often, rather than the periodic updates (see `shaderActiveTicks`), so if we need to sample at some granularity (e.g., every 250, 400, 500 cycles), it is easier to do that with a stat that updates often.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
poal023: done

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
poal023: same as above!

src/mem/ruby/structures/CacheMemory.hh
powerjg: What affects power/energy is the number of times that the tags and data are read a written (and, maybe, MSHRs and other structures). In slicc, you can specify different "resources" (like tag/data read/write, but you can also specify others) and track the number of times those happen in transitions. I would think this is the best way to model cache power. The cache does a lot more than just "read/write hit/miss." There are also probes and replacements and other things that cause activity and need to be accounted for in power models. The resources model is meant to be able to be flexible enough to model these behaviors.

My question is what is your new stats doing that cannot be done with the current "resource" model?

The problem with adding new stats is that we are further bifurcating the protocols. Some are supporting "resources" and now some have "hit/miss" stats. We need to be making things more homogenous, not more different.

poal023: While I'm not disagreeing with anything that you said here and it's true that cache power models are more complex than what I've outlined, Accelwattch only really cares about the items I've listed. The new stats basically allow you to track what Accelwattch considers as part of its' cache power models. 

Like I've said, if there is some way to derive the metrics using what exists today, please let me know, but it's not clear to me? If there is an answer-- that's even better since we don't need to add the new Ruby stats. I know this can be done in the classic caches (since they explicitly tell you these counters), but I'm still unsure how to do this in Ruby.

mattsinc: @powerjg : let us know if you think what we need can be derived from existing stats or not.

src/arch/amdgpu/vega/insts/mimg.cc
mattsinc: @abmerop : are you ok with adding a flag for textures even if none are implemented currently?  I also wonder if this should be done separately from this patch ...

src/gpu-compute/shader.cc
```diff
@@ -537,6 +538,16 @@ Shader::sampleLineRoundTrip(const std::map<Addr, std::vector<Tick>>& lineMap)
     }
 }
 
+void
+Shader::notifyCuActive() {
+    Cycles cur = curCycle();
+    if (cur > _lastActiveCycle) {
+        _lastActiveCycle = cur;
+        stats.shaderActiveCycles++;
+
```
mattsinc: unnecessary extra line of space

mattsinc (changes requested):
1 minor change.

Beyond that, I guess it would be good for @abmerop to chime in about your adding all the flags for the GPU code.  I get why you're doing it (to make tracking different instruction types easy for power purposes), but I just want to make sure it doesn't set off any concerns from @abmerop about subtle impacts those flags might have (e.g., on instruction latency).  I am not aware of any ...

src/mem/ruby/structures/CacheMemory.hh
powerjg: I still don't like the idea of adding new stats to CacheMemory *that are only used in one protocol*. I think it will confuse users when they use another protocol (e.g., CHI) and see that read misses, write misses, etc. are all 0s. 

I also don't like that we're adding these stats just for accelwatch when, IMO, that model doesn't use the best stats that are available. Just because GPGPU-Sim didn't have the ability to track stats at a finer granularity (e.g., data array reads and writes and tag array reads and writes) doesn't mean that gem5 should track things at a coarser granularity.

I believe "read hit/miss" and "write hit/miss" can be derived based on the count of transitions from one state to another based on the event (e.g., S->S on Load is a "read hit"). I understand this would be a big pain.

I think we're trading off "doing things to make accelwatch work" and "doing things the right way for gem5's future" here. I would much prefer we err toward doing things we think are right for gem5.

abmerop: I agree with Jason here.  The way I would recommend doing this is to utilize the global `simout` OutputDirectory class to create a new file in the user specified m5out directory which contains all the information needed for input to AccelWattch  (https://github.com/gem5/gem5/blob/stable/src/base/output.hh).

This will also let you create a file in any format you want which can sometimes be easier to parse than the stats.txt file. There are a handful of references in gem5 to refer to: https://github.com/gem5/gem5/blob/stable/src/cpu/testers/gpu_ruby_test/protocol_tester.cc#L162, https://github.com/gem5/gem5/blob/stable/src/cpu/simple/probes/simpoint.cc#L54.

Basically you would want to `simout.create` a file in some class that all AccelWattch-related stats can access (e.g., RubySystem, System). Then you can create variables and increment the values somewhere, and use `registerExitCallback` to register a function to write the stats when simulation ends. _Note: the examples I linked write output and/or close the file in the destructor but I prefer exit callback in case destructors are not called._

This would allow you to do what you are trying to achieve while avoiding the issue of potentially confusing stats Jason pointed out above.

src/arch/amdgpu/vega/insts/mimg.cc
abmerop: I am completely fine with overly verbose flags. In my opinion there should be more flags than the hardware profilers have available as it adds benefit to using simulation.

src/mem/ruby/structures/CacheMemory.hh
abmerop: May a more simplistic approach is to simply rename the stats to be AccelWattch specific (i.e., accelwattch_read_misses, etc.) and set the statsitics::nozero flags so they do not show up in other protocols. But i'll see what others think of the approach I suggested

powerjg: > May a more simplistic approach is to simply rename the stats to be AccelWattch specific (i.e., accelwattch_read_misses, etc.) and set the statsitics::nozero flags so they do not show up in other protocols. But i'll see what others think of the approach I suggested

I wouldn't fight this approach. I still don't love it, but I won't fight it.

poal023: For the time being, I've added the `statistics::nozero` flag.

src/arch/amdgpu/vega/insts/vop3p.cc
```diff
@@ -758,6 +758,7 @@ Inst_VOP3P__V_PK_MUL_F32::Inst_VOP3P__V_PK_MUL_F32(InFmt_VOP3P *iFmt)
     : Inst_VOP3P(iFmt, "v_pk_mul_f32")
 {
     setFlag(ALU);
+    setFlag(IntMul);
 } // Inst_VOP3P__V_PK_MUL_F32
```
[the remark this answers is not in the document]
poal023: Fixed in latest commit.

src/python/gem5/prebuilt/viper/viper_garnet_network.py
```diff
@@ -0,0 +1,137 @@
+from m5.objects import (
+    GarnetExtLink,
+    GarnetIntLink,
+    GarnetNetwork,
+    GarnetNetworkInterface,
+    GarnetRouter,
+)
```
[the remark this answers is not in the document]
poal023: Addressed in latest commit.

src/python/gem5/prebuilt/viper/gpu_cache_hierarchy.py
```diff
@@ -112,7 +119,10 @@ def __init__(
         self.ruby_gpu.block_size_bytes = cache_line_size
 
         # Ruby network for this GPU
-        self.ruby_gpu.network = SimpleDoubleCrossbar(self.ruby_gpu)
+        if self._use_garnet:
+            self.ruby_gpu.network = GarnetDoubleCrossbar(self.ruby_gpu)
+        else:
+            self.ruby_gpu.network = SimpleDoubleCrossbar(self.ruby_gpu)
```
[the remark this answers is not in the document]
poal023: Addressed in latest commit.

src/gpu-compute/compute_unit.hh
```diff
@@ -1060,6 +1060,26 @@ class ComputeUnit : public ClockedObject
         statistics::Scalar vALUInsts;
         statistics::Formula vALUInstsPerWF;
         statistics::Scalar sALUInsts;
+        statistics::Scalar totalInsts;
+        statistics::Scalar decodedInsts;
+        statistics::Scalar decodedIntInsts;
+        statistics::Scalar decodedFpInsts;
+        statistics::Scalar intInsts;
```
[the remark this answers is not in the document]
poal023: Addressed in latest commit.

src/gpu-compute/GPUStaticInstFlags.py
```diff
@@ -38,6 +38,20 @@ class GPUStaticInstFlags(Enum):
...
+        "FpMul",  # Single-Precision Muls
+        "FpDiv",  # Single-Precision Divs
+        "FpSqrt",  # Fp Square Roots
+        "FpLg",  # Fp Logarithms
+        "FpSin",  # Fp Sin
+        "FpExp",  # Fp Exp
+        "DpMul",  # Double-Precision Muls
+        "DpDiv",  # Double-Precision Divs
+        "Tensor",  # Tensor Insts
+        "Tex",  # Texture Insts
```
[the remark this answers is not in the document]
poal023: Removed the Tensor/Dp flags as we already have MFMA/FMA flags.

src/gpu-compute/compute_unit.cc
```diff
@@ -2505,6 +2510,42 @@ ComputeUnit::ComputeUnitStats::ComputeUnitStats(statistics::Group *parent,
...
                "per-wavefront."),
       ADD_STAT(sALUInsts, "Number of scalar ALU insts issued."),
+      ADD_STAT(totalInsts, "Number of total insts issued."),
+      ADD_STAT(decodedInsts, "Number of total decoded instructions."),
+      ADD_STAT(decodedIntInsts, "Number of total decoded\
+              integer instructions."),
+      ADD_STAT(decodedFpInsts, "Number of total decoded FP instructions."),
+      ADD_STAT(intInsts, "Number of total integer instructions\
+              executed (vector only)."),
+      ADD_STAT(fpInsts, "Number of total fp instructions\
```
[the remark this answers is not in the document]
poal023: Addressed in latest commit.

src/mem/ruby/protocol/GPU_VIPER-TCP.sm
```diff
@@ -719,10 +719,26 @@ machine(MachineType:TCP, "GPU TCP (L1 Data Cache)")
...
 
   action(uu_profileDataHit, "\udh", desc="Profile the demand hit"){
+      peek(mandatoryQueue_in, RubyRequest){
+          if (in_msg.Type == RubyRequestType:LD) {
+              L1cache.profileReadHit();
+          }
+          if (in_msg.Type == RubyRequestType:ST) {
+              L1cache.profileWriteHit();
+          }
+      }
```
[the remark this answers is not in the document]
poal023: @mattsinc @abmerop do I need to change these stat increments to consider other request types like copilot mentions above?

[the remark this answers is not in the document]
abmerop: Probably no for cache maintenance (e.g., FLUSH). For atomics sometimes these are counted as a read+write in which case you might want to.  I find it more useful to explicitly count atomics, however, so I would also say no for that.

src/gpu-compute/fetch_unit.cc
```diff
@@ -639,6 +650,18 @@ FetchUnit::FetchBufDesc::decodeSplitInst()
                                        wavefront->computeUnit->
                                            getAndIncSeqNum());
     wavefront->instructionBuffer.push_back(gpu_dyn_inst);
+    // TODO: stats below *only* apply to vector insts.
```
BobbyRBruce: Is this TODO not important to implement?

src/python/gem5/prebuilt/viper/viper_garnet_network.py
BobbyRBruce, line 1: Is there a reason we're adding a Viper Garnet Network as part of this PR? What has it got to do with adding GPU status to track different instruction types?

src/gpu-compute/shader.cc
```diff
@@ -537,6 +538,15 @@ Shader::sampleLineRoundTrip(const std::map<Addr, std::vector<Tick>>& lineMap)
     }
 }
 
+void
+Shader::notifyCuActive() {
+    Cycles cur = curCycle();
+    if (cur > _lastActiveCycle) {
+        _lastActiveCycle = cur;
+        stats.shaderActiveCycles++;
+    }
```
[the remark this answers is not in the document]
BobbyRBruce: This still needs addressed

src/mem/ruby/structures/CacheMemory.cc
```diff
@@ -567,6 +567,14 @@ CacheMemoryStats::CacheMemoryStats(statistics::Group *parent)
...
       ADD_STAT(m_demand_accesses, "Number of cache demand accesses",
                m_demand_hits + m_demand_misses),
+      ADD_STAT(read_hits, "Number of cache read hits"),
+      ADD_STAT(read_misses, "Number of cache read misses"),
+      ADD_STAT(read_accesses, "Number of cache read accesses",
+               read_hits + read_misses),
+      ADD_STAT(write_hits, "Number of cache write hits"),
+      ADD_STAT(write_misses, "Number of cache write misses"),
+      ADD_STAT(write_accesses, "Number of cache write accesses",
+               write_hits + write_misses),
```
BobbyRBruce: Is there any really any sense in having all three hits, misses, and accesses? With two you can always derive the other?

src/mem/ruby/structures/CacheMemory.cc
```diff
@@ -806,6 +821,30 @@ CacheMemory::profileDemandMiss()
...
+CacheMemory::profileWriteHit()
+{
+    cacheMemoryStats.write_hits++;
+}
+
+void
+CacheMemory::profileWriteMiss()
+{
+    cacheMemoryStats.write_misses++;
+}
```
BobbyRBruce, line 900: Do you not need to set `write_accesses`/`read_accesses` here?

BobbyRBruce (changes requested):

src/mem/ruby/structures/CacheMemory.cc
poal023: the `accesses` stats sort of follow similar behavior to how `m_demand_accesses` works, where the `hits`/`misses` are separate stats and the `accesses` is just a formula (i.e., `accesses = hits + misses). Basically, I'm trying to replicate the same behavior for my stats.

src/mem/ruby/structures/CacheMemory.cc
poal023: This should be addressed by my reply of the `accesses` being a formula of hits/misses, but if you need me to explain more, please let me know.

src/python/gem5/prebuilt/viper/viper_garnet_network.py
poal023: While it's not necessarily adding stats, it will make it easier if someone wants to look at stats which show behavior of a network (e.g., how many flits were injected).

src/gpu-compute/fetch_unit.cc
poal023: It is important, however, I'm only considering the vector instructions because of the Accelwattch power model in another PR only cares about vector insts (at least, the architecture modeled by the one in the PR). I could add separate stats for vector and scalar insts.

mattsinc:
> Given the changes to shared ruby structures the PR name message should make it clear that there are mem-ruby changes.
> 
> I have comments on that code. I didn't look at the GPU code.

@poal023 you will need to updat

> Given the changes to shared ruby structures the PR name message should make it clear that there are mem-ruby changes.
> 
> I have comments on that code. I didn't look at the GPU code.

I have updated the pull request message to resolve this

src/mem/ruby/structures/CacheMemory.hh
mattsinc: @powerjg : as part of the power models we're adding, the power models need to distinguish reads and writes in order to properly attribute energy costs.  That's why we're adding these counters.  If we were to remove them, we wouldn't be able to properly attribute this.

src/mem/ruby/structures/CacheMemory.hh
mattsinc: @poal023 has updated this in his most recent push request to align with @abmerop's suggestion, but since @abmerop and @powerjg are the ones requesting changes, they should ok.

I just realized my reply to the other thread ignored this thread.  If we have to count the way @powerjg wants, then that's what we'll have to do but it will be extremely painful.  It feels like the better long-term strategy would be to update all the protocols to count reads and writes -- I'm not sure why that is considered bad here, but perhaps I don't have the vision to fully understand this.  Maybe it's just too much work?  It seems like the right thing to do to me ... unless there is a straightforward way to do the counting @powerjg wants.

Just let us know what you want.

src/arch/amdgpu/vega/insts/vop2.cc
```diff
@@ -2287,6 +2298,7 @@ namespace VegaISA
         : Inst_VOP2(iFmt, "v_fmac_f32")
     {
         setFlag(ALU);
+        setFlag(FpMul);
     } // Inst_VOP2__V_FMAC_F32
```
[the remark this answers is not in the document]
mattsinc: @poal023 this remains to be done

src/gpu-compute/fetch_unit.cc
mattsinc: @poal023 I think we should just add scalar instructions here too even if AccelWattch doesn't need/want them.  It's not that hard to add, and we'll need it anyways for other power models we add.

src/mem/ruby/protocol/GPU_VIPER-TCP.sm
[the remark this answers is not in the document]
mattsinc: it looks like @poal023 's updates have resolved this and I agree with @abmerop

src/mem/ruby/structures/CacheMemory.cc
mattsinc: I believe @poal023 is correct, but @BobbyRBruce should resolve this

mattsinc (changes requested):
@poal023 from my POV there are still a few things that need to be done.  Please do these.  However, even once these are done there is ultimately the larger question from @powerjg of if the new stats you have proposed to add here should all be ripped out and done another way.  I cannot answer that question, but certainly you have made the changes along the lines of what @abmerop proposed that seem like they meet the minimum of what @powerjg wants ... so he'll have to chime in.

src/mem/ruby/structures/CacheMemory.hh
abmerop: I am fine with the nozero change

abmerop:
> 1 minor change.
> 
> Beyond that, I guess it would be good for @abmerop to chime in about your adding all the flags for the GPU code. I get why you're doing it (to make tracking different instruction types easy for power purposes), but I just want to make sure it doesn't set off any concerns from @abmerop about subtle impacts those flags might have (e.g., on instruction latency). I am not aware of any ...

New flags are fine with me

src/gpu-compute/fetch_unit.cc
poal023: should be changed as of the latest commit

src/arch/amdgpu/vega/insts/vop2.cc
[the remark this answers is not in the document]
poal023: should be addressed

src/gpu-compute/shader.cc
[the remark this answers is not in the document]
poal023: changed the way it has been handled, instead there's an event that is scheduled which will increment this stat versus the way it is done now.

powerjg (commented):
I still don't like having the profile read/write hit/miss stats. I think it's supporting a bad implementation of a power model that we shouldn't be enabling. However, I will accept the current design as long as you make a comment in `CacheMemory.hh` saying that the `Profile[Read/Write][Hit/Miss]` is only used in the GPU protocol and only used for accel watch.

poal023:
> I still don't like having the profile read/write hit/miss stats. I think it's supporting a bad implementation of a power model that we shouldn't be enabling. However, I will accept the current design as long as you make a comment in `CacheMemory.hh` saying that the `Profile[Read/Write][Hit/Miss]` is only used in the GPU protocol and only used for accel watch.

This has been done. Although, as @mattsinc I agree, it might be better to add these counters to other coherence protocols so this sort of counting (not necessarily just for the power model I have) can be done.

To further clarify, the counters are needed because `write_hits` and `write_misses` would resolve identically in the TCC protocol and `write/read_misses` would also resolve identically in the TCP protocol. The number of transitions in the protocols do not tell you this information nor do the existing [Data/Tag]Array[Reads/Writes].

mattsinc:
> > I still don't like having the profile read/write hit/miss stats. I think it's supporting a bad implementation of a power model that we shouldn't be enabling. However, I will accept the current design as long as you make a comment in `CacheMemory.hh` saying that the `Profile[Read/Write][Hit/Miss]` is only used in the GPU protocol and only used for accel watch.
> 
> This has been done. Although, as @mattsinc I agree, it might be better to add these counters to other coherence protocols so this sort of counting (not necessarily just for the power model I have) can be done.
> 
> To further clarify, the counters are needed because `write_hits` and `write_misses` would resolve identically in the TCC protocol and `write/read_misses` would also resolve identically in the TCP protocol. The number of transitions in the protocols do not tell you this information nor do the existing [Data/Tag]Array[Reads/Writes].

@powerjg I just wanted to check in on this again.  @poal023 tried your approach instead, and it does not seem to give the desired results.  Do you want us to try something else, or is what @abmerop proposed (and what @poal023 implemented) sufficient?  Ultimately, I think this patch is waiting on you now.

mattsinc:
@poal023 once you rebase this onto head of develop again, @BobbyRBruce and I can re-review our parts of this

powerjg:
This is fine with me

poal023:
@BobbyRBruce @mattsinc @abmerop @powerjg please have a look, I have rebased on top of `develop`!

src/gpu-compute/exec_stage.cc
```diff
@@ -195,6 +195,31 @@ ExecStage::exec()
...
+            break;
+        }
+    }
+    for (int j = 0; !anyResourceBusy && j < computeUnit.numScalarALUs; ++j) {
+        if (!computeUnit.scalarALUs[j].rdy()) {
+            anyResourceBusy = true;
+        }
+    }
+    if (!anyResourceBusy && !computeUnit.vectorGlobalMemUnit.rdy()) {
+        anyResourceBusy = true;
```
abmerop, line 211: could add break here as well

src/gpu-compute/dispatcher.cc
```diff
@@ -339,7 +349,13 @@ GPUDispatcher::notifyWgCompl(Wavefront *wf)
         DPRINTF(GPUKernelInfo, "Completed kernel %d\n", kern_id);
 
         if (kernelExitEvents) {
-            shader->requestKernelExitEvent(task->completionSignal());
+            // Use this task's own isBlitKernel(), not the shared shader
```
abmerop, line 352: Is this a wider issue? I.e., should other places in the code be checking the task field?

src/gpu-compute/gpu_static_inst.hh
```diff
@@ -141,51 +141,31 @@ class GPUStaticInst : public GPUStaticInstFlags
...
-    isSDWAInst() const
-    {
-        return _flags[IsSDWA];
-    }
-    bool
-    isDPPInst() const
-    {
-        return _flags[IsDPP];
-    }
+    bool isALU() const { return _flags[ALU]; }
```
abmerop, line 144: Was this changed by clang-format? It does not collapse these to one line on my end and the full GPU code was run through clang-format to prevent clutter in PRs. I also want to avoid ping-ponging formatting. Could you double check this file with `clang-format -i src/gpu-compute/gpu_static_inst.hh`

src/gpu-compute/shader.cc
```diff
@@ -555,24 +562,57 @@ Shader::sampleLineRoundTrip(const std::map<Addr, std::vector<Tick>> &lineMap)
...
+        schedule(shaderActiveCycleEvent, clockEdge());
+    }
+}
+
+void
+Shader::shaderActiveCycleTick()
 {
+    if (_activeCus > 0) {
+        stats.shaderActiveCycles++;
+        schedule(shaderActiveCycleEvent, clockEdge(Cycles(1)));
```
abmerop, line 577: Self-scheduling events one cycle away can degrade simulation performance by a lot. Have you checked if this impacts simulation time? Is so, I would try to think about another way to count this.

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
abmerop, line 668: Is this fixed?

src/mem/ruby/protocol/GPU_VIPER-TCC.sm
abmerop, line 676: Is this fixed?

src/python/gem5/components/devices/gpus/viper_shader.py
```diff
@@ -217,7 +217,7 @@ def __init__(
 
         # This arbitrary address is something in the X86 I/O hole
         hsapp_gpu_map_paddr = 0xE00000000 + 0x1000 * shader_id
-        self.dispatcher = GPUDispatcher()
+        self.dispatcher = GPUDispatcher(kernel_exit_events=True)
```
abmerop, line 220: Is this going to cause problems? IIRC, if an stdlib script does not handle an exit event it will exit simulation or something, basically making handling required in all scripts. Or does the change in exit_handler.py prevent this?

Might be better to have the top-level script (i.e., in configs/example/gem5_library/ or elsewhere) set this explicitly. For example, `gpu0.shader.dispatcher.kernel_exit_events = True` in the stdlib script.

abmerop (commented):
I'm fine with the new stats. Some others might want to be aware of the Ruby changes, but moving stats to statistics::Group should be a welcome change.

I have two main concerns (1) Check for simulation performance degradation with the per-cycle added event and (2) Make sure setting dispatcher exit events to always be true does not cause problems with pre-existing stdlib scripts in configs/example/gem5_library/.

erin-le:
@poal023: Hello, could you address the comments above when you get a chance? Also, could you resolve the conflicts that have arisen on this PR? Thanks!

PR #2929, opened 2026-02-04
stdlib: Added support for power models

poal023:
This PR adds in power modeling into the stdlib, particularly the following:

1. A `PowerModelPyFunc` SimObject which takes in Python functions for dynamic and static power. These functions should return doubles. Using functions instead of strings like `MathExprPowerModel` does allows for higher expressivty (e.g., you can load in ML weights for an ML-based power model, you can perform linear algebra operations significantly easier, and you can compose classes in a more coherent manner)
2. An example power model, which tries to remodel McPAT [1] (and by extension, CACTI), using our Python interface. We opted not to just call the C libraries because expressing the model in Python allows for greater flexibility (e.g., if you want to change how a particular component scales by the number of threads, cores, or another parameter, it is significantly easier to do that on a PowerModelPyFunc versus the old `MathExprPowerModel`). This reimplementation of McPAT has been evaluated for accuracy [2], and the gem5 stats and configuration used by this implementation of the power model try to closely match that of McPAT
3. A config script which attaches a power model to the CPU, memory, and caches and runs hello world by replicating an ARM-A9 core. 

There are a couple of caveats, however:

1. This PR does not support power sampling (e.g., if you want to take the power every 500 CPU clock cycles, or 1us, you can't do that at the minute. 
2. The implemented McPAT power model makes a couple of assumptions about your configuration and assumes fixed activation energies (i.e., the activation energies are correct for only one particular configuration). You can obtain these activation energies from McPAT and then put them into a Python dictionary to be passed into the power model itself. A future PR would be add a tool which will estimate these for a particular gem5 configuration.
3. Support for Garnet and Ruby NoCs needs to be finalized, however, as far as I know Garnet is not officially supported in the stdlib? 
4. The stats used for the example power model are close to what McPAT expects (and uses) for their power calculations
5. For CPU cores in the stdlib, you cannot [at the minute? unless I'm wrong?] you cannot specify the number of threads per core, which McPAT requires if you want to do power modeling for multithreaded/SMT cores.
6. McPAT itself doesn't support more modern features (e.g., future technology nodes, DDR4+, HBM), but these are **not a limitation of our PR, but the power model itself.**

I also have some concerns about implementing in the stdlib:
1. I'm not really sure how I can enforce the list of activation energies to be of a type? Is Dict even a good idea? There should be some way to identify which activation energies are what and that the power model [should] need these to work correctly.
2. Is `simulate` the best place for the power model to be in? I thought this is probably okay since the original power models are in `src/sim/power/`, so it would make sense that the gem5 stdlib mirrors this? Power Models aren't really components, and they should be opt-in.
3. I'm not really sure what the "best way" to switch between power models that are already defined and custom power models. One thing I can think of is having an enum in the stdlib, and if a user passes a parameter like `power_model_type=PowerModels.Custom`, they have to pass in their own `PowerModelPyFunc`. Where should I even put this switch, should this be a per-component thing? I've put an example of a cache hierarchy which adds in a power model to all the caches, but perhaps this should be in pre-built? 
4. is there a cleaner way to pass in a power model to some of the objects I've specified? Something like the memory controller (see config script) requires a `MemCtrl` m5.objects (because all the relevant stats live there), but this seems a little cumbersome?

[1] - Li, Sheng, et al. "McPAT: An integrated power, area, and timing modeling framework for multicore and manycore architectures." Proceedings of the 42nd annual ieee/acm international symposium on microarchitecture. 2009.
[2] - https://pages.cs.wisc.edu/~sinclair/papers/asmith-gem5workshop25-power.pdf

EDIT: Typos, formatting, and tagging @mattsinc @BobbyRBruce


src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy_with_power_model.py
```diff
@@ -0,0 +1,134 @@
...
+        for cache in self.l1dcaches:
+            cache.power_model = McPATClassicL1DPowerModel(
+                cache, cache.writeback_clean, self._act_energies
+            )
+        for cache in self.l2cache:
+            cache.power_model = McPATClassicL2PowerModel(
+                cache, cache.writeback_clean, self._act_energies
+            )
+
+    def init_mcpat_act_energies(self) -> None:
```
mattsinc: please add a comment here explaining how you got these values, etc.

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l1d_power_model.py
```diff
@@ -0,0 +1,129 @@
...
+        """Returns dynamic power in Watts"""
+        total_energy = self.dcache_energy()
+        total_energy += self.miss_buffer_energy()
+        total_energy += self.inst_fill_buffer_energy()
+        total_energy += self.prefetch_buffer_energy()
+        total_energy += self.writeback_buffer_energy()
+        return self.convert_to_watts(total_energy)
+
+    def dcache_energy(self):
+        read_accesses = self.get_stat("ReadReq.accesses").total
```
mattsinc, line 69: add comment here explaining what you are doing and how you are accessing the stats

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l1d_power_model.py
```diff
@@ -0,0 +1,129 @@
...
+            + write_misses * self._act_energies["DataCacheTag"]
+            + write_accesses * self._act_energies["DataCache"]["Write"]
+        )
+        if self._writeback:
+            # extra write for write misses since wb policy
+            energy += write_misses * self._act_energies["DataCache"]["Write"]
+        return energy
+
+    def miss_buffer_energy(self):
+        # by default, there is WB in caches (to my knowledge)
```
mattsinc: I'm not sure what this comment means?

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l2_power_model.py
```diff
@@ -0,0 +1,115 @@
...
+        # The SimObject that contains the stats we need.
+        self._simobj = l1dcache
+        self._act_energies = act_energies
+        self._writeback = writeback
+
+        self.dyn = lambda: self.dynamic_power()
+        self.st = lambda: self.static_power()
+
+    def static_power(self):
+        """Returns static power in Watts"""
```
mattsinc: explain in a comment why you are returning 1 here

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l2_power_model.py
```diff
@@ -0,0 +1,115 @@
...
+        write_hits = write_accesses - write_misses
+        return (
+            read_hits * self._act_energies["L2Cache"]["Read"]
+            + read_misses * self._act_energies["L2CacheTag"]["Read"]
+            + write_misses * self._act_energies["L2CacheTag"]["Write"]
+            + write_accesses * self._act_energies["L2Cache"]["Write"]
+        )
+
+    def miss_buffer_energy(self):
+        # by default, there is WB in caches (to my knowledge)
```
mattsinc: I don't think this comment is relevant to the code here?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_decode_power_model.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+        print(" " * indent + f"Instruction Decoder")
+        print(
+            " " * (indent + 2)
+            + f"Runtime Dynamic = {self.convert_to_watts(id_energy)} W\n"
+        )
+
+    def static_power(self) -> float:
+        """Returns static power in Watts"""
+        return 1.0
```
mattsinc, line 61: probably the comment about IF/ID above should be here to explain why you return 1?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_decode_power_model.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def dynamic_power(self) -> float:
+        energy = self.inst_buffer_energy()
+        energy += self.inst_decode_energy()
+        return self.convert_to_watts(energy)
+
+    def inst_buffer_energy(self) -> float:
+        decoded_insts = self.get_stat("fetchStats0.numInsts").total
```
mattsinc, line 69: I suspect you are assuming there is only 1 CPU?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_exec_power_model.py
```diff
@@ -0,0 +1,117 @@
...
+    ):
+        super().__init__(cpu, act_energies)
+        self.name = "McPATCpuExecutePowerModel"
+        self._alu = McPATCpuAluPowerModel(cpu=cpu, act_energies=act_energies)
+        self._rf = McPATCpuRfPowerModel(cpu=cpu, act_energies=act_energies)
+        self._inst_scheduler = McPATCpuInstructionSchedulerPowerModel(
+            cpu=cpu, act_energies=act_energies
+        )
+
+        """ The Activity Factor of the Execution Unit (default: 0.76): """
```
mattsinc: where is default set to 0.76?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_exec_power_model.py
```diff
@@ -0,0 +1,117 @@
...
+        self._exu_act_factor = exu_act_factor
+
+        """ The Activity Factor of the Pipeline itself (default: 1.0): """
+        self._pipeline_act_factor = pipeline_act_factor
+
+        """ Number of Pipeline Stages for any CPUs in McPAT: """
+        self._num_units = 4.0
```
mattsinc, line 82: Add comment explaining why 4/5

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_exec_power_model.py
```diff
@@ -0,0 +1,117 @@
...
+        if isinstance(cpu, BaseO3CPU):
+            self._num_units = 5.0
+
+        """ The number of pipelines our CPU has (assume 1): """
```
mattsinc: doesn't this assumption clash with the above 4/5?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_exec_power_model.py
```diff
@@ -0,0 +1,117 @@
...
+            + self._rf.dynamic_power()
+            + self.convert_to_watts(cdb_energy)
+            + self.convert_to_watts(self.pipeline_energy())
+        )
+        if isinstance(self._simobj, BaseO3CPU):
+            total_energy += self._inst_scheduler.dynamic_power()
+
+        print(" " * indent + f"Execution Unit:")
+        print(" " * (indent + 2) + f"Runtime Dynamic = {total_energy} W\n")
+        self._rf.print_mcpat(indent + 4)
```
mattsinc, line 102: The number "4" (and "5") appears in so many places and is going to be hard to follow.  Can we create a single variable set to 4 and use that everywhere?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_exec_power_model.py
```diff
@@ -0,0 +1,117 @@
...
+            self._alu.dynamic_power()
+            + self._rf.dynamic_power()
+            + self.convert_to_watts(self.bypass_energy())
+            + self.convert_to_watts(self.pipeline_energy())
+        )
+        if isinstance(self._simobj, BaseO3CPU):
+            energy += self._inst_scheduler.dynamic_power()
+        return energy
+
+    def static_power(self) -> float:
```
mattsinc, line 123: why 1?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_fetch_power_model.py
```diff
@@ -0,0 +1,98 @@
...
+        self._btb = McPATCpuBtbPowerModel(
+            cpu=cpu,
+            act_energies=act_energies,
+            has_predictor=self._has_predictor,
+        )
+        self._decode = McPATCpuDecodePowerModel(
+            cpu=cpu, act_energies=act_energies
+        )
+
+        """ The Activity Factor of the Inst. Fetch Unit (default: 0.9): """
```
mattsinc: where are these defaults set?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_lsu_power_model.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class McPATCpuLsuPowerModel(BaseMcPATPowerModel):
+    # avoid the use of default values
+    def __init__(
+        self, cpu: BaseCPU, act_energies, pipeline_act_factor, lsu_act_factor
+    ):
+        super().__init__(cpu, act_energies)
+        self.name = "McPATCpuLsuPower"
+        """ The Activity Factor of the LSU (default: 0.71): """
```
mattsinc: where is default set?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_lsu_power_model.py
```diff
@@ -0,0 +1,99 @@
...
+        self._lsu_act_factor = lsu_act_factor
+
+        """ The Activity Factor of the Pipeline itself (default: 1.0): """
+        self._pipeline_act_factor = pipeline_act_factor
+
+        """ Number of Pipeline Stages for any Inorder CPU in McPAT: """
```
mattsinc: given that this code is effectively identical to that in some of your other files, is there a way to declare it once and use it everywhere else?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_power_model.py
```diff
@@ -0,0 +1,96 @@
...
+            self._rnu = McPATCpuRenamingUnitPowerModel(
+                cpu=cpu, act_energies=act_energies, pipeline_act_factor=1.0
+            )
+        self._lsu = McPATCpuLsuPowerModel(
+            cpu=cpu,
+            act_energies=act_energies,
+            pipeline_act_factor=1.0,
+            lsu_act_factor=0.71,
+        )
+        self._mmu = McPATCpuMmuPowerModel(
```
mattsinc, line 82: seems like this sets defaults?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_renaming_unit_power_model.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(self, cpu: BaseO3CPU, act_energies, pipeline_act_factor):
+        super().__init__(cpu, act_energies)
+        self.name = "McPATCpuRenamingUnitPower"
+
+        """
+         McPAT supports PRF and RS based renaming, but gem5 only renames using
+         a PRF combined with a free list.
+        """
+        """ The Activity Factor of the Pipeline itself (default: 1.0): """
+        self._pipeline_act_factor = pipeline_act_factor
```
mattsinc, line 59: I'm not sure I understand the implications of the comment here about PRF and RS -- why does it mean pipeline_act_factor should be used?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_tournament_bp_power_model.py
```diff
@@ -0,0 +1,129 @@
...
+        )
+
+        return self.convert_to_watts(total_energy)
+
+    def total_glob_pred_energy(self) -> float:
+        if not self._has_predictor:
+            return 0
+        predicts = self.get_stat("branchPred.condPredicted").total
+        mispredicts = (
+            self.get_stat("branchPred.condIncorrect").total + 0.1 * predicts
```
mattsinc, line 115: Why 0.1?

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_backend_power_model.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def set_mem_vars(self):
+        self._llc_block_size = ceil(
+            int(self._simobj._parent.get_cache_line_size()) / 8.0
+        ) + int(self._simobj._parent.get_cache_line_size())
+        self._data_bus_width = ceil(
+            self._simobj.dram.device_bus_width / 8.0
+        ) + int(self._simobj.dram.device_bus_width)
+        clk = 10**6 / self._simobj.clk_domain.clock.getValue()[0]
+        self._clock_rate = clk * 2 * 1e6
```
mattsinc, line 63: why this clock rate?

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_backend_power_model.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def calc_backend_dyn(self):
+        low_power = True
+        if low_power:
+            peakBW = self.get_stat("dram.peakBW").total
+            Vdd = 0.661538
```
mattsinc, line 78: lots of magic numbers here without any comments explaining them

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_frontend_power_model.py
```diff
@@ -0,0 +1,110 @@
...
+        """Returns static power in Watts"""
+        return 1.0
+
+    def dynamic_power(self) -> float:
+        self.set_mem_vars()
+        sched_result_modifier = self._llc_block_size * 8 / self._data_bus_width
+        reads = self.get_stat("readReqs").total
+        writes = self.get_stat("writeReqs").total
+        frontendBufReads = (
+            reads * sched_result_modifier * self._data_bus_width / 72
```
mattsinc: why 72?  Make it a parameter or variable instead of having a magic number?

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_phy_power_model.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def set_mem_vars(self):
+        self._llc_block_size = ceil(
+            int(self._simobj._parent.get_cache_line_size()) / 8.0
+        ) + int(self._simobj._parent.get_cache_line_size())
+        self._data_bus_width = ceil(
+            self._simobj.dram.device_bus_width / 8.0
+        ) + int(self._simobj.dram.device_bus_width)
+        clk = 10**6 / self._simobj.clk_domain.clock.getValue()[0]
+        self._clock_rate = clk * 2 * 1e6
```
mattsinc, line 61: Same comment here about why this clock rate, but also we use it in multiple places, probably should have a variable and use everywhere?

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_phy_power_model.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+        energy += self.mem_refresh_overhead()
+        return self.convert_to_watts(energy)
+
+    def calc_tdp(self):
+        # obj's path: board.memory.mem_ctrl
+        peakBW = self.get_stat("dram.peakBW").total
+        energy = (
+            self._curve_fitted_constant
+            * (peakBW * 8 * 1e6 / 1e9)
```
mattsinc: Need to explain magic numbers

mattsinc (commented):
Thanks @poal023 for pushing this!  I've added a number of comments for you to address.

I also wonder if the best way to integrate this is to first have a patch without McPAT, but rather just the code creating the interface, then a subsequent commit with the McPAT support to make the commits more modular?  Not sure, but @BobbyRBruce or @powerjg can chime in.

src/python/gem5/simulate/power_models/abstract_power_model.py
```diff
@@ -0,0 +1,61 @@
...
+from m5.objects import Root
+from m5.util import panic
+
+
+class AbstractPowerModel:
+    def __init__(self, simobj):
+        self._simobj = simobj
+        self.name = "AbstractPowerModel"
+
+    def get_stat(self, stat):
```
powerjg, line 58: Should this return a float or whatever a "stat" is? (is a "stat" a float?)

src/python/gem5/simulate/power_models/abstract_power_model.py
```diff
@@ -0,0 +1,61 @@
...
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+from m5.objects import Root
+from m5.util import panic
+
+
+class AbstractPowerModel:
```
powerjg, line 42: Probably should have the ABC metaclass.

src/python/gem5/simulate/power_models/mcpat_power_model/base_mcpat_power_model.py
```diff
@@ -0,0 +1,49 @@
...
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+from m5.objects import Root
+
+from ..abstract_power_model import AbstractPowerModel
+
+
+class BaseMcPATPowerModel(AbstractPowerModel):
+    def __init__(self, simobj, act_energies):
```
powerjg: Can you add types here? What is "act_energies?" Also, a docstring would be nice.

src/python/gem5/simulate/power_models/mcpat_power_model/base_mcpat_power_model.py
```diff
@@ -0,0 +1,49 @@
...
+        super().__init__(simobj)
+        self.name = "BaseMcPATPowerModel"
+        self._act_energies = act_energies
+
+    def getExecutionTime(self):
+        """Note that this you are running 1 workload and *only*
```
powerjg: This is not a docstring that describes how to use this function this is a comment.

src/python/gem5/simulate/power_models/mcpat_power_model/base_mcpat_power_model.py
```diff
@@ -0,0 +1,49 @@
...
+        want power values after your ROI/simulation"""
+        return Root.getInstance().resolveStat("simSeconds").total
+
+    def convert_to_watts(self, value: float) -> float:
+        """Note that McPAT AEs are already in terms of J,
```
powerjg: Ditto, this is a comment.

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_phy_power_model.py
powerjg, line 1: Needs copyright

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_frontend_power_model.py
```diff
@@ -0,0 +1,110 @@
+from math import ceil
```
powerjg, line 30: Copyright

src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy_with_power_model.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        Same the parent, however, adds a McPATPowerModel
+        """
+        # Initialize the parent class with the standard arguments
+        super().__init__(
+            l1d_size=l1d_size,
+            l1i_size=l1i_size,
+            l2_size=l2_size,
+        )
+        self._act_energies = {}
```
powerjg: A type here would be nice.

src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy_with_power_model.py
powerjg: Why is this a function and not just an assignment in the contructor?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_alu_power_model.py
```diff
@@ -0,0 +1,79 @@
...
+class McPATCpuAluPowerModel(BaseMcPATPowerModel):
+    def __init__(self, cpu: BaseCPU, act_energies):
+        super().__init__(cpu, act_energies)
+        self.name = "McPATCpuAluPowerModel"
+        """
+        Note that for vec access energy I'm using
+        McPAT's access energy for Mul/Div ops.
+        I suspect this is actually larger, but this
+        is a placeholder. Surprisingly, McPAT doesn't
+        actually account for SIMD insts.
```
powerjg: Don't use triple quotes for comments.

src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy_with_power_model.py
```diff
@@ -0,0 +1,134 @@
...
+)
+
+
+class PrivateL1SharedL2CacheHierarchyWithPowerModel(
+    PrivateL1SharedL2CacheHierarchy
+):
+    def __init__(self, l1d_size: str, l1i_size: str, l2_size: str) -> None:
+        """
+        Same the parent, however, adds a McPATPowerModel
+        """
```
powerjg: Docstrings aren't comments. They are used to document the use of the function. Write comments if you want to comment.

src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy_with_power_model.py
powerjg: Why are these energies in this file and not in `cache_power_model/`? 

Also, these numbers are for some specific size. They are invalid for other cache capacities and associativities. This should probably be part of some concrete instance of this class with hard-coded sizes/associativities.

powerjg (changes requested):
Below, I've left a few comments. The review isn't complete, though.

A comment on the overall design: I *like* that the power model for the cache is using composition over inheritance. This is a good object-oriented design choice. Using a member variable instead of subclassing follows the 'Has-A' vs. 'Is-A' rule perfectly here, making the relationship between these objects much clearer.

However, I don't know if I like separating the files so much. The file `mcpat_classic_l1dcache_power_model.py` can only possibly be used right here. There is no other uses of that file anywhere in the codebase. I think it would be better to put the power models with the things they will be used with.

Please add copyright to all of the files. Similarly, where I made comments about "comments" vs "docstrings" please apply these comments throughout the PR.

src/python/gem5/simulate/power_models/mcpat_power_model/base_mcpat_power_model.py
poal023: Should be addressed in latest commits.

src/python/gem5/simulate/power_models/mcpat_power_model/base_mcpat_power_model.py
poal023: Should be addressed in latest commits.

src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy_with_power_model.py
poal023: I've explicitly added an ActEnergyType which is just a dict with keys as either a float or a nested dictionary (particularly useful if you want an operation and then if you want a read/write/search like act_energies["ICache"]["Read"]).

src/python/gem5/simulate/power_models/abstract_power_model.py
poal023: Done, AbstractPowerModel should have the metaclass attribute updated.

src/python/gem5/simulate/power_models/abstract_power_model.py
poal023: Not necessarily, you can use `resolveStat(stat).total` to get a float, but there are instsances where you might not want this (e.g., a vector stat, you might want to only grab an index rather than the total, see the ALU power model for an example).

src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy_with_power_model.py
poal023: Should be changed, the constants used by this example power model are put into another file.

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_alu_power_model.py
poal023: Fixed, only comments are used/DocStrings were used.

src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy_with_power_model.py
poal023: Done, this is removed.

src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy_with_power_model.py
poal023: Should be now in a separate file, the constants are put into another file. I've also updated the params to explicitly be of a certain value.

src/sim/power/power_model_pyfunc.hh
```diff
@@ -0,0 +1,35 @@
...
+     PARAMS(PowerModelPyFunc);
+     PowerModelPyFunc(const Params &p);
+     double getDynamicPower() const override {
+             pybind11::object result_py = dyn_func();
+             return result_py.cast<double>();
+     }
+     double getStaticPower() const override {
+             pybind11::object result_py = st_func();
+             return result_py.cast<double>();
+     }
```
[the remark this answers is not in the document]
poal023: Should be resolved, added an assert statement to check for this.

src/sim/power/power_model_pyfunc.cc
```diff
@@ -0,0 +1,18 @@
+#include "sim/power/power_model_pyfunc.hh"
+
+#include "base/trace.hh"
```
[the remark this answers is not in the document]
poal023: Done.

src/sim/power/PowerModelPyFunc.py
```diff
@@ -0,0 +1,20 @@
...
+class PowerModelPyFunc(PowerModelState):
+    type = "PowerModelPyFunc"
+    cxx_header = "sim/power/power_model_pyfunc.hh"
+    cxx_class = "gem5::PowerModelPyFunc"
+
+    # Equations for dynamic and static power in Watts
+    # Equations may use gem5 stats ie. "1.1*ipc + 2.3*l2_cache.overall_misses"
+    # It is possible to use automatic variables such as "temp"
+    # You may also use stat names (relative path to the simobject)
+    dyn = Param.PyFunc("Function to call for Dynamic Power")
```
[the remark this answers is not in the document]
poal023: Should be corrected.

SConstruct
```diff
@@ -657,6 +657,7 @@ for variant_path in variant_paths:
         env.Append(CCFLAGS=['-Werror',
                              '-Wno-error=deprecated-declarations',
                              '-Wno-error=deprecated',
+                             '-Wno-error=attributes'
                             ])
```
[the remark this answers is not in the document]
powerjg: Following up on this, whatever warning is being thrown should be fixed unless it absolutely can't be. Please let us know what the warning is so we can decide if it's OK to ignore it.

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l2_power_model.py
```diff
@@ -0,0 +1,115 @@
...
+    PowerModel,
+    PowerModelPyFunc,
+)
+
+from ...base_mcpat_power_model import BaseMcPATPowerModel
+
+
+class L2PowerOn(PowerModelPyFunc, BaseMcPATPowerModel):
+    def __init__(self, l1dcache: Cache, writeback: bool, act_energies):
+        super().__init__()
```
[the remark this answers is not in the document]
poal023: The L2 model calls all the functions and I already set `_simobj` in the L2 PM class, so it should work fine?

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l1d_power_model.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def dcache_energy(self):
+        read_accesses = self.get_stat("ReadReq.accesses").total
+        write_accesses = self.get_stat("WriteReq.accesses").total
+
+        read_misses = self.get_stat("ReadReq.misses").total
+        write_misses = self.get_stat("WriteReq.misses").total
+
+        read_hits = read_accesses - read_misses
+        write_hits = write_accesses - write_misses
```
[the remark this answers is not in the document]
poal023: Should be done.

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l1i_power_model.py
```diff
@@ -0,0 +1,87 @@
...
+
+        return (
+            read_hits * self._act_energies["InstCache"]["Read"]
+            + read_misses * self._act_energies["InstCache"]["Read"]
+            + read_misses * self._act_energies["InstCache"]["Write"]
+        )
+
+    def miss_buffer_energy(self):
+        read_accesses = write_accesses = self.get_stat("ReadReq.misses").total
+        read_misses = self.get_stat("ReadReq.misses").total
```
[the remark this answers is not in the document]
poal023: Done.

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l2_power_model.py
```diff
@@ -0,0 +1,115 @@
...
+        write_accesses = (
+            self.get_stat("overallAccesses").total
+            + self.get_stat("WritebackClean.accesses").total
+        )
+
+        read_misses = self.get_stat("ReadExReq.misses").total
+        write_misses = self.get_stat("overallMisses").total - read_misses
+
+        read_hits = read_accesses - read_misses
+        write_hits = write_accesses - write_misses
```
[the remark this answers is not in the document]
poal023: Done.

src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy_with_power_model.py
```diff
@@ -0,0 +1,134 @@
...
+                cache, cache.writeback_clean, self._act_energies
+            )
+        for cache in self.l1dcaches:
+            cache.power_model = McPATClassicL1DPowerModel(
+                cache, cache.writeback_clean, self._act_energies
+            )
+        for cache in self.l2cache:
+            cache.power_model = McPATClassicL2PowerModel(
+                cache, cache.writeback_clean, self._act_energies
+            )
```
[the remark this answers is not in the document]
poal023: This should be changed in the latest commits?

src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy_with_power_model.py
```diff
@@ -0,0 +1,134 @@
...
+    McPATClassicL2PowerModel,
+)
+
+
+class PrivateL1SharedL2CacheHierarchyWithPowerModel(
+    PrivateL1SharedL2CacheHierarchy
+):
+    def __init__(self, l1d_size: str, l1i_size: str, l2_size: str) -> None:
+        """
+        Same the parent, however, adds a McPATPowerModel
```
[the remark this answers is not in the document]
poal023: Removed comment.

SConstruct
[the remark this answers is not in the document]
poal023: Seems like I get these warnings:
```src/sim/power/power_model_pyfunc.hh:43:7: warning: ‘gem5::PowerModelPyFunc’ declared with greater visibility than the type of its field ‘gem5::PowerModelPyFunc::st’ [-Wattributes]
src/sim/power/power_model_pyfunc.hh:43:7: warning: ‘gem5::PowerModelPyFunc’ declared with greater visibility than the type of its field ‘gem5::PowerModelPyFunc::st_func’ [-Wattributes]
src/sim/power/power_model_pyfunc.hh:43:7: warning: ‘gem5::PowerModelPyFunc’ declared with greater visibility than the type of its field ‘gem5::PowerModelPyFunc::dyn_func’ [-Wattributes]
In file included from src/sim/power/power_model_pyfunc.cc:32:
src/sim/power/power_model_pyfunc.hh:43:7: warning: ‘gem5::PowerModelPyFunc’ declared with greater visibility than the type of its field ‘gem5::PowerModelPyFunc::dyn’ [-Wattributes]
```

I don't know if there's a good way to fix this?

poal023:
> Below, I've left a few comments. The review isn't complete, though.
> 
> A comment on the overall design: I _like_ that the power model for the cache is using composition over inheritance. This is a good object-oriented design choice. Using a member variable instead of subclassing follows the 'Has-A' vs. 'Is-A' rule perfectly here, making the relationship between these objects much clearer.
> 
> However, I don't know if I like separating the files so much. The file `mcpat_classic_l1dcache_power_model.py` can only possibly be used right here. There is no other uses of that file anywhere in the codebase. I think it would be better to put the power models with the things they will be used with.
> 
> Please add copyright to all of the files. Similarly, where I made comments about "comments" vs "docstrings" please apply these comments throughout the PR.

I think I've added the majority of the changes you've suggested (feel free to push back). 

I'm not really sure what you mean about separating the files, do you mean like in my example cache hierarchy with power model, the `PowerModelPyFunc` classes that I use to attach a PM should be in the same place as the cache hierarchy with a power model?

As far as the composition-like design, is this something I should also do with the CPU/Memory? Just provide an example of an ARM core which has been validated through McPAT and have something like `ArmA9WithPowerModel` which inherits from `SimpleCore` or something and specify the ISA to be ARM and align the parameters correctly?

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_phy_power_model.py
poal023: Some of these I'm not actually too sure how they get them, but I've added a couple of comments to explain where they come from?

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_phy_power_model.py
poal023: Maybe I should make something like a `McPATBaseMemoryPowerModel` class so things like the clock rate are defined there?

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_frontend_power_model.py
poal023: I've changed this in my commits-- there should be a `dimm_data_width` parameter which is set to 72.

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_backend_power_model.py
poal023: This should be addressed in the latest commits?

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_backend_power_model.py
poal023: Probably need to add a comment for this, but basically you convert the clock rate into MHz (e.g., 2000 is passed into gem5/the McPAT XML file, this needs to be 2000e6) then because it's double-pumped, you need to scale it by 2x.

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l2_power_model.py
poal023: Removed this comment in my commit.

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l2_power_model.py
poal023: It is a placeholder, many comments should've been added to explain this.

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l1d_power_model.py
poal023: This should be removed, but basically this is always assumed to be true, but in case the caches are specified to not be writeback, there are some energy costs which are not added.

src/python/gem5/components/cachehierarchies/classic/private_l1_shared_l2_cache_hierarchy_with_power_model.py
poal023: This should be addressed in my latest commits?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_fetch_power_model.py
poal023: These are based on the McPAT XML for the ARM A9 @ 2GHz, the duty cycle/activity factors for the stage are defined per XML file. Should probably add a comment explaining so.

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_lsu_power_model.py
poal023: I should probably add this in somewhere, but it should be set in the `McPATCpuPowerModel` class.

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_power_model.py
poal023: yes, that is true for this case.

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_renaming_unit_power_model.py
poal023: This should be corrected in my latest commit, but this is just a comment about the class itself. It doesn't have anything to do with the pipeline activity factor.

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_tournament_bp_power_model.py
poal023: I think the assumption is that 10% of all branches will flip internal bits?

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_decode_power_model.py
poal023: This should be explained in my latest commit.

src/python/gem5/simulate/power_models/mcpat_power_model/cpu_power_model/mcpat_cpu_decode_power_model.py
poal023: Yes, this assumes there's only 1 thread in a core, since all the stdlib (least as far as I know) cores also assume 1 thread. Since the power model is per-core, this should work fine (you can just apply the power model to each core).

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_phy_power_model.py
poal023: Should be addressed in latest commit?

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l2_power_model.py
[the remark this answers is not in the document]
mattsinc: @powerjg maybe an issue Copilot cannot catch properly?

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l1d_power_model.py
mattsinc: @poal023 : as far as I can tell, you did not add a comment here.

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_frontend_power_model.py
```diff
@@ -0,0 +1,143 @@
...
+    def set_mem_vars(self):
+        self._llc_block_size = ceil(
+            int(self._simobj._parent.get_cache_line_size()) / 8.0
+        ) + int(self._simobj._parent.get_cache_line_size())
+        self._data_bus_width = ceil(
+            self._simobj.dram.device_bus_width / 8.0
+        ) + int(self._simobj.dram.device_bus_width)
+        self._num_rws_ports = self._simobj._parent._num_channels
+        clk = 10**6 / self._simobj.clk_domain.clock.getValue()[0]
+        self._clock_rate = clk * 2 * 1e6
```
mattsinc, line 66: seems like the comment you added to mcpat_mem_backend_power_model.py about the clock rate also needs to be added here.

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_backend_power_model.py
mattsinc: seems like you still need to add this comment?

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_backend_power_model.py
```diff
@@ -0,0 +1,127 @@
...
+            ones (e.g., 32 <= 40 <= 45), then the Vdd is obtained by
+            linearly interpolating between the two known technology nodes.
+        (3) the peak bandwidth of a MC.
+        """
+        low_power = True
+        if low_power:
+            peakBW = self.get_stat("dram.peakBW").total
+            Vdd = 0.661538
+            tech_node = 40
+            # Some of the constants below (e.g., 0.9e-9, 12800)
```
mattsinc: I think we discussed how you need to update this comment

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_backend_power_model.py
mattsinc: I think the comment is still missing the context that these numbers come from McPAT?

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_backend_power_model.py
```diff
@@ -0,0 +1,82 @@
...
+                * self._data_bus_width
+                / 72
+                * Vdd
+                / 1.1
+                * Vdd
+                / 1.1
+                * (tech_node / 65)
+            )
+        self._backend_dyn = backend_dyn
+
```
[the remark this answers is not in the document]
mattsinc: @poal023 : does this need to be fixed?

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_phy_power_model.py
mattsinc: that seems reasonable to me, but if it's a massive pain in the butt to do, then just adding a comment in places like this is fine with me.

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_phy_power_model.py
mattsinc: where did you add these comments?

mattsinc (changes requested):
Seems like at least a couple of my prior requests still need to be done -- let me know if I'm missing them though.

src/python/gem5/simulate/power_models/mcpat_power_model/cache_power_model/classic/mcpat_classic_l2_power_model.py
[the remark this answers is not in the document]
powerjg: Copilot is correct. This is a great example of why you shouldn't use multiple inheritance. It creates ambiguity in which class is responsible for what.

You should probably call `BaseMcPATPowerModel.__init__(self, act_energies)`.

Just because it works today doesn't mean it's "right" or "good" code. What happens when, in the future, someone modifies `BaseMcPATPowerModel.__init__`? Then, it would likely break this code!

src/python/gem5/simulate/power_models/mcpat_power_model/mem_ctrl_power_model/mcpat_mem_phy_power_model.py
poal023: Should be added with latest commit.

PR #2930, opened 2026-02-04
cpu-o3: restore PR #2657 and fix dependency graph cleanup on squash

BobbyRBruce:
This PR fixes the Weekly x86 failure by addressing an O3 Instruction Queue dependency-graph invariant violation triggered by squash ordering.

The Weekly test failure first appeared after merging PR #2891. This PR alters the guest-visible CPU feature set which changes the Linux boot instruction mix/timing in a way that triggers a buggy corner case in O3 IQ dependency-graph cleanup. As such, this PR does not undo #2891, but fixes the bug #2891 exposes. During squash, destination physical registers can still have dependent nodes present transiently, and leaving those nodes behind can later trip the “dependency graph not empty” sanity check when a new producer is installed for the same register.

The first commit reverts PR #2914, which had reverted PR #2657 under the false assumption it caused the Weekly failures. The Weekly failures persist with that revert in place, so this PR restores the behavior from PR #2657 and then applies the actual fix.

The second commit updates the squash path to drain any remaining dependents for each destination physical register (by popping the list) and clears the producer pointer unconditionally. This ensures the dependency graph does not retain stale nodes/producer pointers for squashed instructions and preserves the intended producer-side invariant check in addToProducers().

### DIFF
--- a/src/cpu/o3/inst_queue.cc
+++ b/src/cpu/o3/inst_queue.cc
@@ -1452,8 +1452,9 @@ InstructionQueue::doSquash(ThreadID tid)
             if (dest_reg->isAlwaysReady()) {
                 continue;
             }
-            assert(dependGraph.empty(dest_reg->flatIndex()));
-            dependGraph.clearInst(dest_reg->flatIndex());
+            if (!dependGraph.empty(dest_reg->flatIndex())) {
+                dependGraph.clearInst(dest_reg->flatIndex());
+            }
         }
         instList[tid].erase(squash_it--);
         ++iqStats.squashedInstsExamined;
@@ -1535,7 +1536,8 @@ InstructionQueue::addToProducers(const DynInstPtr &new_inst)
             continue;
         }
 
-        if (!dependGraph.empty(dest_reg->flatIndex())) {
+        if (!dependGraph.empty(dest_reg->flatIndex()) &&
+            new_inst->isNonSpeculative()) {
             dependGraph.dump();
             panic("Dependency graph %i (%s) (flat: %i) not empty!",
                   dest_reg->index(), dest_reg->className(),
--- a/src/cpu/o3/inst_queue.cc
+++ b/src/cpu/o3/inst_queue.cc
@@ -1452,9 +1452,14 @@ InstructionQueue::doSquash(ThreadID tid)
             if (dest_reg->isAlwaysReady()) {
                 continue;
             }
-            if (!dependGraph.empty(dest_reg->flatIndex())) {
-                dependGraph.clearInst(dest_reg->flatIndex());
+            // When squashing, remove any remaining dependents and clear the
+            // producer pointer. Squash ordering can leave dependents
+            // temporarily present, and keeping them around can trip later
+            // dependency graph sanity checks.
+            while (!dependGraph.empty(dest_reg->flatIndex())) {
+                dependGraph.pop(dest_reg->flatIndex());
             }
+            dependGraph.clearInst(dest_reg->flatIndex());
         }
         instList[tid].erase(squash_it--);
         ++iqStats.squashedInstsExamined;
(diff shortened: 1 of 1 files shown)


powerjg:
What was the root cause of this regression? The test worked at some point, then stopped working. What change happened that caused this test to fail?

BobbyRBruce:
> What was the root cause of this regression? The test worked at some point, then stopped working. What change happened that caused this test to fail?

Oh yes, sorry for forgetting to mention that. I edited the PR description to include an explanation. In short though, it is PR #2891 that triggers the Weekly test failures but #2891 isn't the true cause of the bug. PR #2891 exposes a corner case in the dependency graph in that weekly test. It is this corner case which needed fixed.

powerjg:
I am still not convinced here... How long does the error take to happen? I would like to trace back the instruction and its dependents that are still held at this point. I *really* think the bug is somewhere else. Here's my reasoning (from gemini of course). I prompted it to ask if the assert was wrong (i.e., I prompted it from your side). Here's the output

Of course. It's a complex interaction, and you're right to be thorough. The lifetime of the DynInstPtr is the core issue. Let's break it down using both files.

### **1\. The DependencyGraph Structure (dep\_graph.hh)**

First, let's clarify the data structure from dep\_graph.hh. For each physical register, there's a DependencyEntry that acts as a "head" node.

* dependGraph\[idx\].inst: This holds a DynInstPtr to the instruction that **PRODUCES** the register's value.  
* dependGraph\[idx\].next: This is a pointer to a linked list of other DependencyEntry nodes. Each of these nodes holds a DynInstPtr to an instruction that **CONSUMES** the register's value (i.e., a dependent instruction).

So, for a given register, we have one producer and a list of consumers.

### **2\. empty() vs. clearInst()**

Now let's look at those two functions in this context:

* **bool empty(RegIndex idx) const**: The implementation is return \!dependGraph\[idx\].next;. This checks if the pointer to the **consumer list** is null. It returns true only if there are **no dependent instructions**. It completely ignores the producer.  
* **void clearInst(RegIndex idx)**: The implementation is dependGraph\[idx\].inst \= NULL;. This nullifies the DynInstPtr for the **producer instruction**.

So, the code in inst\_queue.cc:

```

assert(dependGraph.empty(dest_reg->flatIndex()));
dependGraph.clearInst(dest_reg->flatIndex());

```

is doing two very different things:

1. **Assert** that there are no consumers waiting for this register.  
2. **Clear** the pointer to the producer instruction (the one being squashed).

### **3\. Why Not Just Walk the Graph?**

You're still not convinced you shouldn't just walk the consumer list and clear the pointers. Let's trace the implications of that compared to the current code.

The fundamental question is: **Whose responsibility is it to remove an instruction from a dependency list?**

* **Current Design:** When an instruction (inst\_B) is squashed, it is responsible for removing itself from the consumer lists of all the instructions it depends on (inst\_A). This happens in the first for loop in doSquash that iterates over src\_reg\_idx.  
* **Your Proposed Design:** When a producer (inst\_A) is squashed, it would be responsible for clearing its entire list of consumers.

The current design is more robust for catching bugs. Consider a scenario where a dependent instruction, inst\_B, is for some reason *not* marked for squashing when it should have been.

* **With the Current Code:** The main doSquash loop would finish. inst\_B would still be in the system and still on inst\_A's consumer list. When the loop processes inst\_A, the assert(dependGraph.empty(...)) **will fail**. This immediately tells you there's a bug: a dependent instruction was not correctly squashed.  
* **With Your Proposed Change:** When inst\_A is squashed, it would walk its consumer list and remove the pointer to inst\_B. The assertion would now pass. However, inst\_B is still alive in the instList\! It has not been properly squashed. It will now wait forever for a register that has no producer, and the explicit dependency link is gone, making this a much harder bug to track down.

The assertion isn't just checking for an empty list; it's verifying a critical invariant of the squash process: **an instruction should not be squashed while other instructions still depend on it.** The existing logic ensures all dependents are squashed *first*, because they are younger. The assertion confirms this logic holds.

You are absolutely correct that the DynInstPtr held by the dependency graph for the producer (dependGraph\[idx\].inst) must be cleared to prevent a memory leak. That is precisely what clearInst() does. But it should only be done *after* verifying that no dependent instructions were missed by the squash logic, which is what assert(dependGraph.empty()) accomplishes.

Edit: tried to fix the formatting

PR #2935, opened 2026-02-06
mem-ruby: Add missing txnIds in CHI protocol

powerjg:

### DIFF
--- a/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
+++ b/src/mem/ruby/protocol/chi/CHI-cache-actions.sm
@@ -3077,6 +3077,7 @@ action(Send_FwdSnpResp, desc="") {
     out_msg.addr := address;
     out_msg.responder := machineID;
     out_msg.Destination.add(tbe.requestor);
+    out_msg.txnId := tbe.txnId;
 
     // We have the data (locally or upstream) or are dropping it
     bool keepData := (tbe.dir_sharers.count() > 0) ||
@@ -3162,6 +3163,8 @@ action(Send_RespSepData, desc="") {
     out_msg.type := CHIResponseType:RespSepData;
     out_msg.responder := machineID;
     out_msg.Destination.add(tbe.requestor);
+    out_msg.txnId := tbe.txnId;
+    out_msg.dbid := tbe.txnId;
   }
 }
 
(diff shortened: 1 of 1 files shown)


powerjg:
Note: txnId is used for the RN->HN path and dbid is used for HN->RN (when the HN needs to initiate the transaction). See B2.4.2 in the CHI specification.

Here's a detailed explanation from gemini

## Gem5 SLICC Implementation
In `CHI-cache-actions.sm`, the usage structurally adheres to this specification workflow, with a minor simplification specific to Gem5's address-based routing.

### 1. Request Phase
When a request is initiated (e.g. at the L1 cache), a TBE is allocated. The `txnId` is usually derived from the TBE's slot or inherited along the protocol path.
```ruby
// In CHI-cache-funcs.sm
tbe.txnId := static_cast(Addr, "value", tbe.storSlot);
```

### 2. Completer Response Phase
When the Completer (e.g., HN or downstream directory) responds with `DBIDResp` or `CompDBIDResp`, it sets the IDs as follows:
```ruby
action(Send_CompDBIDResp, desc="") {
  // ...
  out_msg.txnId := tbe.txnId; // Echoes the Requester's TxnID
  out_msg.dbid := tbe.txnId;  // ALSO echoes the Requester's TxnID as the DBID
}
```
**Observation:** The Completer echoes the Requester's `TxnID` into *both* the `TxnID` and `DBID` fields. Since `TxnID` is already guaranteed to be unique for that Requester, the Completer safely uses the exact same ID value rather than explicitly allocating a distinct set of IDs for its internal buffers. Gem5's Ruby routing fundamentally uses line addresses for lookup regardless of the `DBID`, but mathematically the constraint is maintained.

### 3. Requester Handles the Response
When the Requester receives the response, it obeys the specification by updating its internal state to capture the generic `DBID` from the response for future messages:
```ruby
action(Receive_ReqResp_CopyDBID, desc="Copy the rsp DBID into the TBE") {
  // ...
  peek(rspInPort, CHIResponseMsg) {
    tbe.txnId := in_msg.dbid; // Captures the DBID
  }
}
```

### 4. Data/Ack Phase
Finally, when the Requester sends Write Data or a Completion Ack, it uses that newly captured `TxnID` (which is functionally the `DBID` received in the response):
```ruby
action(Send_Data, desc="") {
  enqueue(datOutPort, CHIDataMsg, data_latency) {
    // ...
    out_msg.txnId := tbe.txnId; // Now uses the copied DBID
  }
}
```

giactra (approved):
Thanks!

PR #2939, opened 2026-02-11
CI: always upload artifacts even when test steps fail

BobbyRBruce:
This PR ensures result/artifact upload steps still run when earlier test/compile steps fail (so failures are still reported, but outputs are preserved for debugging).

Changes:
- Replace if: success() || failure() with if: always() for artifact-related steps in CI/Daily/Weekly workflows.
- Verified no continue-on-error: true exists in these workflows on stable.


BobbyRBruce:
Closing: this PR mistakenly targeted stable. Superseded by draft PR #2940 which targets develop.

PR #2944, opened 2026-02-12
RISC-V: Refine CSR Serialization Handling in O3 CPU

hossein-mokhtarnia:
**Summary**

This PR addresses an incompatibility between the changes proposed in [#2700](https://github.com/gem5/gem5/pull/2700) and the RISC-V ISA. The original implementation is not directly adaptable to RISC-V due to differences in CSR semantics and serialization requirements.

To resolve this, we introduce RISCV-specific handling for CSR operations in the O3 CPU pipeline.

**Motivation**

In RISC-V, not all CSR operations require serialization. While CSR writes must remain serialized to preserve architectural correctness, CSR reads do not require full pipeline serialization in user mode.

**Changes Introduced**

1. New Instruction Flag for RISC-V CSR Operations

- A dedicated instruction flag is introduced for CSR-related operations in the RISC-V ISA.

2. Refined Serialization Handling in O3 Rename Stage

- During renaming, CSR operations are examined:

- If the operation is a CSR write, it remains serialized.

- If the operation is a CSR read, the serialized flag is cleared, since serialization is not required.

- This avoids unnecessary pipeline serialization for CSR reads.

3.User-Mode Restriction

- The optimization is applied only in user mode.

In kernel mode, all CSR operations remain serialized to preserve correct RISC-V semantics, particularly for first-use and privileged CSR handling.

### DIFF
--- a/src/arch/riscv/isa/decoder.isa
+++ b/src/arch/riscv/isa/decoder.isa
@@ -6383,32 +6383,32 @@ decode QUADRANT default Unknown::unknown() {
                     Rd = rvSext(data);
                     data = rvZext(Rs1);
                 }}, 'RD != 0', 'true'
-                  , IsSerializeAfter, IsNonSpeculative, No_OpClass);
+                  , IsCsrInst, IsSerializeAfter, IsNonSpeculative, No_OpClass);
                 0x2: csrrs({{
                     Rd = rvSext(data);
                     data = rvZext(data | Rs1);
                 }}, 'true', 'RS1 != 0'
-                  , IsSerializeAfter, IsNonSpeculative, No_OpClass);
+                  , IsCsrInst, IsSerializeAfter, IsNonSpeculative, No_OpClass);
                 0x3: csrrc({{
                     Rd = rvSext(data);
                     data = rvZext(data & ~Rs1);
                 }}, 'true', 'RS1 != 0'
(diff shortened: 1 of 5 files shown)


giactra:
Thanks @hossein-mokhtarnia for your contribution. I don't think though that's they way to implement it for other ISAs like RISCV. In theory there shouldn't be any need to embed ISA specific information in the generic cpu-o3 code.

What you should be doing is to override the isSerializing method for RISCV MiscRegClass, similarly to what we do in Arm:

https://github.com/gem5/gem5/pull/2700/changes#diff-df2f9b3a8253c833924f363a9ee40057000702ac4366f206df8bad0086963cf5R596:
```

class MiscRegClassOps : public RegClassOps
{
  public:
    std::string
    regName(const RegId &id) const override
    {
        return miscRegName[id.index()];
    }

    bool
    serializing(const RegId &id) const override
    {
        return lookUpMiscReg[id.index()].info[MISCREG_SERIALIZING];
    }
};
```

You need to have a metadata structure which tags which CSR can be serialized and which doesn't

erin-le:
@hossein-mokhtarnia: Hello, could you provide an update on the status of this on your end? Thank you!

hossein-mokhtarnia:
Hi, there will be no further updates to this PR, so we can close it. We will open a new PR once the changes are aligned with PR https://github.com/gem5/gem5/pull/2700.
Thank you!

PR #2960, opened 2026-02-24
arch-arm, sim-se: Implement arm linux syscall 403 clock_gettime

dbear496:
This PR implements syscall 403 for arm. This is essentially the same as the existing clock_gettime syscall except the timespec is 64-bit even on 32-bit arm. I also implemented the monotonic clock so it is different from the wall clock.

### DIFF
--- a/src/arch/arm/linux/se_workload.cc
+++ b/src/arch/arm/linux/se_workload.cc
@@ -493,6 +493,7 @@ class SyscallTable32 : public SyscallDescTable<EmuLinux::SyscallABI32>
               {base + 384, "getrandom", getrandomFunc<ArmLinux32>},
               {base + 397, "sys_statx", ignoreFunc},
               {base + 398, "sys_rseq", ignoreFunc},
+              {base + 403, "clock_gettime", clock_gettimeFunc<ArmLinux32, ArmLinux64::timespec>},
               {base + 435, "clone3", clone3Func<ArmLinux32>},
           })
     {}
@@ -788,6 +789,7 @@ class SyscallTable64 : public SyscallDescTable<EmuLinux::SyscallABI64>
                {base + 292, "io_pgetevents"},
                {base + 293, "rseq", rseqFunc<ArmLinux64>},
                {base + 294, "kexec_file_load"},
+               {base + 403, "clock_gettime", clock_gettimeFunc<ArmLinux64>},
                {base + 435, "clone3", clone3Func<ArmLinux64>},
                {base + 1024, "open", openFunc<ArmLinux64>},
                {base + 1025, "link"},
--- a/src/arch/arm/linux/se_workload.cc
+++ b/src/arch/arm/linux/se_workload.cc
@@ -493,7 +493,8 @@ class SyscallTable32 : public SyscallDescTable<EmuLinux::SyscallABI32>
               {base + 384, "getrandom", getrandomFunc<ArmLinux32>},
               {base + 397, "sys_statx", ignoreFunc},
               {base + 398, "sys_rseq", ignoreFunc},
-              {base + 403, "clock_gettime", clock_gettimeFunc<ArmLinux32, ArmLinux64::timespec>},
+              {base + 403, "clock_gettime",
+               clock_gettimeFunc<ArmLinux32, ArmLinux64::timespec>},
               {base + 435, "clone3", clone3Func<ArmLinux32>},
           })
     {}
--- a/src/kern/linux/linux.hh
+++ b/src/kern/linux/linux.hh
@@ -336,6 +336,19 @@ class Linux : public OperatingSystem
     static const unsigned TGT_WEXITED               = 0x00000004;
     static const unsigned TGT_WCONTINUED            = 0x00000008;
     static const unsigned TGT_WNOWAIT               = 0x01000000;
+
+    static const unsigned TGT_CLOCK_REALTIME        = 0x00000000;
+    static const unsigned TGT_CLOCK_MONOTONIC       = 0x00000001;
+    static const unsigned TGT_CLOCK_PROCESS_CPUTIME = 0x00000002;
+    static const unsigned TGT_CLOCK_THREAD_CPUTIME  = 0x00000003;
+    static const unsigned TGT_CLOCK_MONOTONIC_RAW   = 0x00000004;
+    static const unsigned TGT_CLOCK_REALTIME_COARSE = 0x00000005;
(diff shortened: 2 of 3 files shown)


powerjg (commented):
This looks fine to me, thanks for the contribution! I believe the PR needs to be rebased before we can merge, though.

dbear496:
@powerjg I rebased onto the latest develop. I'm not sure why GH was complaining since there were no conflicts.

powerjg (dismissed):

giactra:
> This PR implements syscall 403 for arm. This is essentially the same as the existing clock_gettime syscall except the timespec is 64-bit even on 32-bit arm. I also implemented the monotonic clock so it is different from the wall clock.

Thanks for this PR. In order for CI to pass, you need to run:

`$git clang-format HEAD~1`

over your commit.

I have one question: is there a reason for using 64-bit timespec on arm32 systems? That might not be working I am afraid with the guest ABIs. The VPtr will read 128bit from memoy and if the caller ABI is different the data structures won't match

powerjg:
The CI errors look like real problems, not flakiness in the runners. @dbear496 can you take a look? It's likely related to Giacomo's comment.

dbear496:
I ran git-clang-format, and it reformatted some things -- honestly not to my taste, but whatever.

The default for glibc on 32-bit ARM is to use syscall 403 for clock_gettime, and it interprets both tv_sec and tv_nsec as 64-bit values. Yes, on 32-bit ARM! I observed that even with `_TIME_BITS=32`, glibc would still use syscall 403, and interpret each value as 64-bits, and then it would cast the 64-bit values to 32-bit values. If the ABI wants a 32-bit version of clock_gettime, then it should use syscall 263, which is purely 32-bit as before.

giactra:
> I ran git-clang-format, and it reformatted some things -- honestly not to my taste, but whatever.
> 

Yeah, understandable, though you cannot put a price on consistency

> The default for glibc on 32-bit ARM is to use syscall 403 for clock_gettime, and it interprets both tv_sec and tv_nsec as 64-bit values. Yes, on 32-bit ARM! I observed that even with `_TIME_BITS=32`, glibc would still use syscall 403, and interpret each value as 64-bits, and then it would cast the 64-bit values to 32-bit values. If the ABI wants a 32-bit version of clock_gettime, then it should use syscall 263, which is purely 32-bit as before.

Gotcha.
I did a bit of digging and it looks like the timespec is now using 64bit:

https://elixir.bootlin.com/linux/v6.19.6/source/include/uapi/linux/time_types.h#L7

And previous timespec has been renamed into old_timespec (using long instead of 64bit variable)

https://elixir.bootlin.com/linux/v6.19.6/source/include/uapi/linux/time_types.h#L31

I think we could merge this as is once you rebase and we restart the CI.
Ideally we should consider following kernel convention and renaming timespec into old_timespec and add a 64-bit based timespec. And have clock_gettime32 use old_timespec.

But's more like an excercise for the future

erin-le:
@dbear496: Hello, could you give an update on the status of this PR on your end? Thank you!

dbear496:
Oh. I just forgot to rebase. I'll do that once I'm at my computer.

dbear496:
I'm running into build issues from #3009, so I am unable to test may changes for this PR

powerjg:
Does this fix your build problem? https://github.com/gem5/gem5/pull/3028/changes#diff-d78c5c377c03dd251edc72cdb0ae7d9986fa08361eb392187218f8c670c0bc4eR696

dbear496:
> Does this fix your build problem? https://github.com/gem5/gem5/pull/3028/changes#diff-d78c5c377c03dd251edc72cdb0ae7d9986fa08361eb392187218f8c670c0bc4eR696

I still get errors when building #3028.

erin-le:
@dbear496: Could you try testing this PR from within gem5's Ubuntu 24.04 Docker container? It uses GCC 13.2, so you shouldn't run into the build issues from #3009. The command to launch the Docker container is below. It will mount your current working directory to `/gem5` in the Docker container. 
```bash
docker run -u $UID:$GID --volume ./:/gem5 --rm -it  ghcr.io/gem5/ubuntu-24.04_all-dependencies:latest
```

There's some more information on the Docker containers on the following page: https://www.gem5.org/documentation/general_docs/building

dbear496:
I tried building with the docker container like you mentioned, but I get the following error:

```
In file included from build/ARM/proto/inst_dep_record.pb.cc:6:
build/ARM/proto/inst_dep_record.pb.h:14:10: fatal error: google/protobuf/runtime_version.h: No such file or directory
   14 | #include "google/protobuf/runtime_version.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I get this error when checked out on the latest commit on the develop branch.

erin-le:
@dbear496: Hello, sorry for the delay on this. I'm having trouble reproducing the error you encountered above. I used the following commands, and used the latest gem5 `develop` as of April 17th (235c858447053a64abcd0b182bde8e60010284e0):

To launch the Docker container (same command as above):
`docker run -u $UID:$GID --volume ./:/gem5 --rm -it  ghcr.io/gem5/ubuntu-24.04_all-dependencies:latest`

To build gem5 after launching the Docker container:
```
cd gem5
scons build/ARM/gem5.opt -j 20
```

On my end, Docker finishes building `ARM/gem5.opt`. I also tried building gem5 `ALL` in the Docker container with this PR checked out and saw the same errors as those in the CI tests. Could you provide more detail on how you're building? Also, could you check your Protobuf version? Here's the command and output that I got:

Command:
```
protoc --version
```

Output:
```
libprotoc 3.21.12
```
If you have a different protobuf version, could you try using the one above? Thanks!

Additionally, you could try making a new worktree, checking out the latest develop, and building in the Docker container with the new worktree mounted, in case there's anything in your current worktree interfering with your build.

dbear496:
Alright, I got it working after re-cloning the repository. I rebased onto the latest develop and made sure that it compiles.

> Ideally we should consider following kernel convention and renaming timespec into old_timespec and add a 64-bit based timespec. And have clock_gettime32 use old_timespec.

I agree, but it is not trivial. The `timespec` struct is also used elsewhere in syscalls that I didn't bother to touch, and there is also a `timeval` struct that perhaps should use 64-bit ints too.

BobbyRBruce (approved):
This all seems to have been addressed. Seems fine to me.

erin-le:
@dbear496: Hello, sorry to block this before merging, but could you modify the title of your clang-format commit to include `misc: ` at the beginning? Thank you!

PR #2977, opened 2026-03-03
mem-ruby: Exit simulation when TlmGenerator is done

giactra:
The TlmGenerator currently doesn't exit the simulator when the packet
has been processed. This is not a big problem if using SimpleMemory
since we just skip to the end of time. For more complex systems, we
end up simulating for a very long time since other components continue
to schedule events even after we stop injecting packets.

Add support a method to check if the component is active
(TlmGenerator::isActive) which is used to check if we should exit
after when receiving a reply to an injected packet. The isActive
method is exported to Python to simplify cases where we have multiple
TlmGenerators and we need to wait for all of them to exit.

### DIFF
--- a/src/mem/ruby/protocol/chi/tlm/generator.cc
+++ b/src/mem/ruby/protocol/chi/tlm/generator.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024-2025 Arm Limited
+ * Copyright (c) 2024-2026 Arm Limited
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
@@ -47,7 +47,8 @@ namespace tlm::chi {
 bool
 TlmGenerator::Transaction::Expectation::run(Transaction *tran)
 {
-    auto res_print = csprintf("Checking %s...", name());
+    auto res_print =
+        csprintf("%u: Checking %s...", tran->phase().txn_id, name());
     if (cb(tran)) {
         inform("%s\n", res_print + " Success ");
         return true;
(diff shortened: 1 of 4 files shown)


src/mem/ruby/protocol/chi/tlm/generator.cc
```diff
@@ -300,7 +300,11 @@ TlmGenerator::recv(ARM::CHI::Payload *payload, ARM::CHI::Phase *phase)
         // Check existing expectations
         it->second->runCallbacks();
     } else {
-        warn("Transaction untested\n");
+        warn("%u: Transaction untested\n", phase->txn_id);
+    }
+
+    if (!isActive()) {
+        exitSimLoop("TlmGenerator done");
```
powerjg, line 307: No action required, but FYI, there is a new API for exiting: https://github.com/gem5/gem5/blob/stable/src/sim/sim_exit.hh#L64

You can specify a "type" for the exit (`type_id`) instead of just a string. These types are automatically matched with exit handlers (example found here: https://github.com/gem5/gem5/blob/stable/src/python/gem5/simulate/exit_handler.py#L231) and the handlers is invoked when the exit happens.

This allows us to not have massive switch/case style statements that do string matching for exit reasons. You can also add any information you would like to return to python (in a string:string dictionary/map), but that is unlikely to be needed in cases like this one.

powerjg (approved):

src/mem/ruby/protocol/chi/tlm/generator.cc
giactra: Ack, I will add to my TODO list to study the hypercalls and see whether we can port some of the existing exits

PR #2980, opened 2026-03-03
scons: Set default NUMBER_BITS_PER_SET for ARM target

giactra:
We frequently simulate interconnects with nodes. This means that we run into the default snoopfilter limits in Ruby. Set the NUMBER_BITS_PER_SET to 128 to make sure we can track more nodes.

Change-Id: I03c9278c5fb6953c2c3d500baa1ec607dd2542f8


powerjg (approved):

yichungchen84:
This has to be set before "RUBY_PROTOCOL_CHI=y" in the previous versions (24?). Not sure if the rule still applies to 25.

giactra:
> This has to be set before "RUBY_PROTOCOL_CHI=y" in the previous versions (24?). Not sure if the rule still applies to 25.

It seems to work for me..
When I do

`scons defconfig build_dir build_opts/ARM`

I can see the following snippet in build_dir/gem5.build/config:

```
RUBY_PROTOCOL_CHI=y
# RUBY_PROTOCOL_MSI is not set
# end of Ruby protocols

# SLICC_HTML is not set
NUMBER_BITS_PER_SET=128
# end of Ruby

```

PR #2983, opened 2026-03-03
python,sim: Unify simulation-exit handling through hypercalls

BobbyRBruce:
## Summary

This PR establishes hypercall ID and payload dispatch as the primary model for
handling simulation exits in the gem5 Python standard library. It centralizes
gem5's built-in simulation-exit hypercall IDs, exposes the same identifiers to
guest code, C++, and Python, and migrates internal callers to clearly defined
exit APIs.

Legacy string-based exits remain supported through an explicit compatibility
path that preserves their cause, code, scheduling, priority, and repetition
semantics.

## Motivation

Classic gem5 exits are identified primarily by human-readable cause strings,
such as `"checkpoint"` or `"Finished drain"`, with an optional integer code.
Python translates those strings into `ExitEvent` values and selects the
corresponding behavior. As a result:

- Exact message spelling forms part of the behavioral interface.
- A diagnostic string can be mistaken for a stable event identity.
- Similar events may be conflated, while unrelated events may be assigned an
  incorrect cause through copy-and-modify errors.
- The classic cause/code interface cannot naturally carry structured event
  metadata.
- Guest code, simulator code, and Python may represent the same event using
  separate numeric constants, strings, or registration rules.

Hypercall exits provide a distinct event identity and an optional key/value
payload. Python can dispatch the event to a registered handler, which performs
the required action and determines whether simulation should continue.
Before this change, however, built-in IDs remained distributed across several
layers, the intended use of the overlapping exit APIs was unclear, and many
internal callers still depended on classic behavior.

## Background

- #947 introduced hypercall exit events and demonstrated that exits could
  originate from guest software, simulator components, or an external control
  signal and be handled through Python using an ID and payload.
- #1995 moved hypercall registration and default behavior into `ExitHandler`
  classes, reducing the `Simulator` object's dependency on individual event
  types and allowing users to register custom handlers.
- #2514 attempted to migrate maximum-instruction exits to a new hypercall. That
  work showed that a partial migration could conflate semantically distinct
  events such as `MAX_INSTS` and `SIMPOINT_BEGIN`, expose additional exits to
  Python, and change legacy behavior. It also highlighted the need for a
  central allocation and definition of built-in hypercall IDs.

This PR provides the common identifier, dispatch, documentation, and
compatibility infrastructure required for subsequent exit-event migrations.

### Canonical hypercall identifiers

`include/gem5/hypercall_ids.h` is the authoritative definition of gem5's
built-in simulation-exit hypercall IDs and their descriptions. The same list
generates:

- Guest-facing `M5_HYPERCALL_*` constants.
- The C++ `gem5::ExitHypercall` enum.
- Descriptors used by the Python bindings.
- The Python `_m5.event.ExitHypercall` enum.

This prevents the guest, simulator, and Python layers from independently
assigning different meanings to the same numeric value.

Python `ExitHandler` registration accepts an `ExitHypercall` enum member for a
built-in event or an explicit integer for a custom event. String selectors are
intentionally rejected so handler registration does not depend on name parsing
or spelling conventions.

### Hypercall-based dispatch

The preferred simulator API is `exitSimulationLoop`, which accepts an
`ExitHypercall` or custom integer ID and an optional string key/value payload.
The resulting `GlobalSimLoopExitEvent` carries both values to the Python
`Simulator` run loop. The run loop selects an `ExitHandler` by ID, invokes it
with the payload, and uses its result to decide whether to continue or leave
the simulation loop.

The event may originate inside the simulator through `exitSimulationLoop`, in
guest software through `m5_hypercall`, or through the external hypercall signal
interface. All sources use the same Python dispatch model.

### Classic compatibility boundary

Existing configurations and simulator code may observe legacy
`getCause()`/`getCode()` fields, match exact cause strings, or depend on the
scheduling behavior of `exitSimLoop` and `exitSimLoopNow`. Converting those
callers directly to an ordinary hypercall exit would change observable
behavior.

`exitSimulationLoopClassic` and `exitSimulationLoopClassicNow` therefore carry
the legacy cause and code through an internal `CLASSIC_GENERATOR` hypercall
payload. `ClassicGeneratorExitHandler` translates the cause into the existing
`ExitEvent` path. This preserves:

- Legacy cause and code reporting.
- `simQuantum` scheduling for global exits.
- Immediate minimum-priority exit behavior.
- Repeating exit scheduling.
- Existing `on_exit_event` generator and function behavior.

`CLASSIC_GENERATOR` is not exported as a guest-facing `M5_HYPERCALL_*`
constant because guest `m5_hypercall` calls cannot supply its required cause
and code payload. Generic callers that use this internal ID must provide a
non-empty cause and a valid base-10 integer code.

### API consolidation and migration

- Deprecate the C++ and Python `exitSimLoop` compatibility APIs.
- Direct new hypercall-based code to `exitSimulationLoop` and
  `exitSimulationLoopNow`.
- Direct code that requires legacy cause/code behavior to
  `exitSimulationLoopClassic` or `exitSimulationLoopClassicNow`.
- Remove the transitional `exitSimLoopWithHypercall` and
  `exitSimLoopWithClassicGenerator` APIs.
- Migrate internal callers across CPU models, Ruby, GPU components, devices,
  SystemC, SST, utilities, examples, and tests to the appropriate API.

## Architectural benefits

- Replaces distributed magic numbers with named, shared identifiers.
- Separates event identity from human-readable diagnostics.
- Separates event identity, event metadata, handler behavior, and the decision
  to continue or stop simulation.
- Allows payload-bearing events without expanding the classic cause/code
  interface.
- Detects invalid or incomplete compatibility payloads at the C++ boundary.
- Provides a documented migration path without changing established classic
  exit behavior.
- Gives future exit types a central, reviewable identifier allocation point.

## Enabled capabilities

The unified dispatch model supports:

- Guest-reported kernel and after-boot milestones.
- Region-of-interest begin and end handling.
- Simulator-scheduled exits with justification metadata.
- Checkpoint requests.
- External status, statistics, and debug-control requests.
- Custom user-defined IDs and handlers.
- Handler specialization through subclassing or function registration.
- Future separation of currently conflated events, including distinct
  maximum-instruction and SimPoint identities.

## Compatibility and scope

This change does not remove the classic `ExitEvent` interface. Classic events
continue to be handled through the compatibility handler while new code can use
the hypercall-based path directly.

This PR does not complete the maximum-instruction migration proposed in #2514.
It establishes the identifier and compatibility infrastructure needed to
implement that migration without combining distinct event meanings or changing
legacy behavior.

Delayed and repeating classic exits are supported, but pending global exit
events are not saved in checkpoints and therefore are not automatically
rescheduled when a checkpoint is restored. The legacy `serialize` argument is
retained for API compatibility but has no effect.

## Validation

- Changed-file `pre-commit` checks.
- `scons -Q build/ALL/gem5.opt -j10`.
- `build/ALL/gem5.opt tests/run_pyunit.py --directory tests/pyunit/stdlib`
  (109 tests passed).
- Compiled runtime smoke tests for classic cause/code preservation and
  malformed `CLASSIC_GENERATOR` payload rejection.

### DIFF
--- a/include/gem5/exit_hypercalls.hh
+++ b/include/gem5/exit_hypercalls.hh
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2026
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met: redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer;
+ * redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution;
+ * neither the name of the copyright holders nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __GEM5_EXIT_HYPERCALLS_HH__
+#define __GEM5_EXIT_HYPERCALLS_HH__
+
+/**
+ * Central list of hypercall-driven exit handlers. The list is expressed as a
+ * macro so multiple translation units (C and C++) can derive enums, constants,
+ * and documentation without duplicating the mapping.
+ *
+ * OP arguments: (EnumName, MacroName, Value, Description)
+ */
+#define GEM5_FOREACH_EXIT_HYPERCALL(OP)                                       \
+    OP(ClassicGenerator, CLASSIC_GENERATOR, 0,                                \
+       "Legacy generator-based exit handling (ExitEvent translation)")        \
+    OP(KernelBooted, KERNEL_BOOTED, 1, "Guest kernel reported it has booted") \
+    OP(AfterBoot, AFTER_BOOT, 2, "Guest entered the after_boot hook")         \
+    OP(AfterBootScript, AFTER_BOOT_SCRIPT, 3,                                 \
+       "Guest completed after_boot.sh")                                       \
+    OP(WorkBegin, WORK_BEGIN, 4, "Entered a region-of-interest (workbegin)")  \
+    OP(WorkEnd, WORK_END, 5, "Exited a region-of-interest (workend)")         \
+    OP(ScheduledExit, SCHEDULED_EXIT, 6,                                      \
+       "Simulator scheduled tick/max-tick exit")                              \
+    OP(Checkpoint, CHECKPOINT, 7, "Take a checkpoint and continue running")   \
+    OP(Orchestrator, ORCHESTRATOR, 1000,                                      \
+       "Orchestrator control/status hypercall")
+
+#endif // __GEM5_EXIT_HYPERCALLS_HH__
--- a/src/arch/arm/pmu.cc
+++ b/src/arch/arm/pmu.cc
@@ -459,14 +459,14 @@ PMU::pmuExitEvent(bool check_reset)
         pmuEnabled = any_counter_enabled;
         if (pmuEnabled) {
             inform("Exiting simulation: PMU enable detected");
-            exitSimLoop("performance counter enabled", 0);
+            exitSimulationLoopClassic("performance counter enabled");
         } else {
             inform("Exiting simulation: PMU disable detected");
-            exitSimLoop("performance counter disabled", 0);
+            exitSimulationLoopClassic("performance counter disabled");
         }
     } else if (check_reset && reg_pmcr.p) {
         inform("Exiting simulation: PMU reset detected");
-        exitSimLoop("performance counter reset", 0);
+        exitSimulationLoopClassic("performance counter reset");
     }
 }
 
@@ -738,7 +738,7 @@ PMU::raiseInterrupt()
 {
     if (exitOnPMUInterrupt) {
         inform("Exiting simulation: PMU interrupt detected");
-        exitSimLoop("performance counter interrupt", 0);
+        exitSimulationLoopClassic("performance counter interrupt");
     }
     if (interrupt) {
         DPRINTF(PMUVerbose, "Delivering PMU interrupt.\n");
--- a/src/python/gem5/simulate/exit_handler.py
+++ b/src/python/gem5/simulate/exit_handler.py
@@ -46,6 +46,8 @@ import m5
 from m5 import options
 from m5.util import warn
 
+from _m5 import event as _m5_event
+
 from ..utils.override import overrides
 from .exit_event import ExitEvent
 from .exit_event_generators import (
@@ -59,33 +61,86 @@ from .exit_event_generators import (
     warn_default_decorator,
 )
 
+HypercallSelector = Union[int, str, _m5_event.ExitHypercall]
+
+
+def _hypercall_name_candidates(raw: str) -> List[str]:
+    cleaned = raw.strip()
+    if not cleaned:
+        return []
+    snake = cleaned.replace("-", "_").replace(" ", "_")
+    pascal = "".join(part.capitalize() for part in snake.split("_") if part)
+    candidates = [cleaned, snake, snake.upper(), cleaned.upper(), pascal]
+    deduped = []
+    for candidate in candidates:
+        if candidate and candidate not in deduped:
+            deduped.append(candidate)
+    return deduped
+
+
+def _resolve_hypercall_id(selector: HypercallSelector) -> int:
+    if isinstance(selector, _m5_event.ExitHypercall):
+        return int(selector.value)
+    if isinstance(selector, str):
+        for candidate in _hypercall_name_candidates(selector):
+            if hasattr(_m5_event.ExitHypercall, candidate):
+                return int(getattr(_m5_event.ExitHypercall, candidate).value)
+        raise ValueError(
+            f"Unknown hypercall '{selector}'. Known values: "
+            f"{', '.join(member.name for member in _m5_event.ExitHypercall)}"
+        )
+    try:
+        return int(selector)
+    except (TypeError, ValueError) as exc:
+        raise ValueError(
+            f"Unable to convert '{selector}' into a hypercall ID"
+        ) from exc
+
 
 class ExitHandlerMeta(ABCMeta):
     """Metaclass for ExitHandler that automatically registers subclasses"""
 
-    def __new__(mcs, name, bases, attrs, hypercall_num: int = None) -> Any:
+    def __new__(
+        mcs,
+        name,
+        bases,
+        attrs,
+        hypercall_num: Optional[HypercallSelector] = None,
+        hypercall: Optional[HypercallSelector] = None,
+    ) -> Any:
+        if hypercall is not None and hypercall_num is not None:
+            raise TypeError(
+                "Specify only one of 'hypercall' or 'hypercall_num'."
+            )
+        selector = hypercall if hypercall is not None else hypercall_num
+
         cls = super().__new__(mcs, name, bases, attrs)
         # Don't register the base ExitHandler class itself
         if name != "ExitHandler":
-            # If hypercall_num is not provided, and the class is a subclass
-            # of another ExitHandler, use the hypercall_num of the base class
+            hypercall_id = (
+                _resolve_hypercall_id(selector)
+                if selector is not None
+                else None
+            )
+            # If no hypercall is provided, and the class is a subclass of another
+            # ExitHandler, reuse the hypercall of the base class.
             for base in bases:
                 if base in ExitHandler.get_handler_map().values():
-                    hypercall_num = base.get_handler_id()
+                    hypercall_id = base.get_handler_id()
                     break
-            assert hypercall_num is not None, (
-                f"Hypercall number must be provided for {name}. Use "
-                "`class {name}(ExitHandler, hypercall_num={hypercall_num}):`"
+            assert hypercall_id is not None, (
+                f"Hypercall identifier must be provided for {name}. Use "
+                "`class {name}(ExitHandler, hypercall=<hypercall>):`"
             )
-            ExitHandler._handler_map[hypercall_num] = cls
-            cls._handler_id = hypercall_num
+            ExitHandler._handler_map[hypercall_id] = cls
+            cls._handler_id = hypercall_id
 
         return cls
 
 
 class ExitHandler(metaclass=ExitHandlerMeta):
 
-    _handler_map: Dict[str, Type["ExitHandler"]] = {}
+    _handler_map: Dict[int, Type["ExitHandler"]] = {}
     _handler_id: int = None
 
     @classmethod
@@ -94,7 +149,7 @@ class ExitHandler(metaclass=ExitHandlerMeta):
         return cls._handler_id
 
     @classmethod
-    def get_handler_map(cls) -> Dict[str, Type["ExitHandler"]]:
+    def get_handler_map(cls) -> Dict[int, Type["ExitHandler"]]:
         """Returns the mapping of exit handler IDs to handler classes"""
         return cls._handler_map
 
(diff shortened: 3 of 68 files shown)


powerjg:
Other places: https://github.com/gem5/gem5/blob/7a2b0e413d06c5ce7097104abef3b1d9eaabca91/src/python/m5/simulate.py#L345

It would be even better to go through all of the places we call `exitSimLoop` and update it to `exitSimulationLoop`: https://github.com/search?q=repo%3Agem5/gem5%20exitSimLoop&type=code

Also, related: https://github.com/gem5/gem5/pull/2514. This PR should be revived or replaced with this one.

include/gem5/exit_hypercalls.hh
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright (c) 2026
```
powerjg: Copyright who?

src/sim/sim_exit.hh
```diff
@@ -32,45 +32,152 @@
...
 
 #include "base/types.hh"
+#include "gem5/exit_hypercalls.hh"
 
 namespace gem5
 {
 
 Tick curTick();
 
+/// Hypercall identifiers used when scheduling exits with
```
powerjg: We should make it clear which one of these many, many, similar functions are preferred and which are deprecated. I would be for adding `[[deprecated]]` in another PR as ell.

BobbyRBruce: Yeah, you're right. I'll spin this deprecation out into a separate PR

include/gem5/exit_hypercalls.hh
BobbyRBruce: Fixed.

src/python/m5/event.py
```diff
@@ -96,4 +96,29 @@ def create(func, priority=Event.Default_Pri):
     "SimExit",
     "mainq",
     "create",
+    "exitSimLoop",
 ]
+
+
+def exitSimLoop(message, exit_code=0, when=None, repeat=0, serialize=False):
```
powerjg, line 110: I would prefer if this was above `__all__`

util/disk-image-validator/config_tester.py
```diff
@@ -76,7 +76,7 @@ def MESI_cache_test(workload_id, resource_version):
         obtain_resource(workload_id, resource_version=resource_version)
     )
 
-    class KernelBootedExit(ExitHandler, hypercall_num=1):
+    class KernelBootedExit(ExitHandler, hypercall="KernelBooted"):
```
powerjg: Why the change from `hypercall_num` to `hypercall`?

Why is the `hypercall` parameter a string and not a (enum) object?

util/disk-image-validator/gem5-bridge-driver-validate.py
```diff
@@ -117,7 +117,7 @@
 from gem5.simulate.exit_handler import ExitHandler
 
 
-class Gem5BridgeDriverTestExitHandler(ExitHandler, hypercall_num=1234):
+class Gem5BridgeDriverTestExitHandler(ExitHandler, hypercall=1234):
```
powerjg: What is 1234? we should at least document it.

src/python/gem5/simulate/exit_handler.py
```diff
@@ -59,42 +71,95 @@
     warn_default_decorator,
 )
 
+HypercallSelector = Union[int, str, _m5_event.ExitHypercall]
+
+
+def _hypercall_name_candidates(raw: str) -> List[str]:
```
powerjg: Please add documentation. Even though it's internal, I as a reviewer don't understand what the purpose is without thinking really hard.

src/python/gem5/simulate/exit_handler.py
```diff
@@ -59,42 +71,95 @@
...
+    pascal = "".join(part.capitalize() for part in snake.split("_") if part)
+    candidates = [cleaned, snake, snake.upper(), cleaned.upper(), pascal]
+    deduped = []
+    for candidate in candidates:
+        if candidate and candidate not in deduped:
+            deduped.append(candidate)
+    return deduped
+
+
+def _resolve_hypercall_id(selector: HypercallSelector) -> int:
```
powerjg: Please add documentation.

src/python/gem5/simulate/exit_handler.py
```diff
@@ -59,42 +71,95 @@
...
     @classmethod
-    def get_handler_id(cls) -> int:
-        """Returns the ID of the exit handler"""
+    def get_handler_id(cls) -> Optional[int]:
+        """Returns the ID of the exit handler (or None for the base class)."""
         return cls._handler_id
 
     @classmethod
-    def get_handler_map(cls) -> Dict[str, Type["ExitHandler"]]:
+    def get_handler_map(cls) -> Dict[int, Type["ExitHandler"]]:
```
powerjg, line 282: I wrote this code, but I'm now confused (which means I should have documented better)

Can we add documentation for `_handler_map` and `_handler_id`?

I think that `_handler_id` is the ID that the class is supposed to handle. And then the class is instantiated each time the exit happens.

I think that `_handler_map` is a map shared across all classes (in the `ExitHandler` class) to know what class to instantiate when there is an exit.

I am confused how both of these are `classmethod`s in the `ExitHandler`. It seems that one is specific to a subclass and one is shared across all subclasses. Clearly I'm missing something. Can you clarify?

src/python/gem5/simulate/exit_handler.py
```diff
@@ -122,23 +187,16 @@ def _exit_simulation(self) -> bool:
 
 
 def _ExitHandlerFactory(
-    hypercall_num: int, func: Callable[["Simulator"], bool], description: str
+    hypercall: HypercallSelector,
+    func: Callable[["Simulator", Dict[str, str]], bool],
+    description: str,
 ) -> Type[ExitHandler]:
-    """
-    Factory function to create a new ExitHandler class with a specific hypercall number,
```
powerjg, line 128: Why delete the args/return?

src/python/gem5/simulate/exit_handler.py
```diff
@@ -154,32 +212,66 @@ def _exit_simulation(self) -> bool:
     return NewExitHandler
 
 
+@overload
```
powerjg: This is fancy python that is not obvious to everyone. Can you add a comment to explain what this code is doing.

src/python/gem5/simulate/exit_handler.py
```diff
@@ -59,42 +71,95 @@
...
             )
-            ExitHandler._handler_map[hypercall_num] = cls
-            cls._handler_id = hypercall_num
+            ExitHandler._handler_map[hypercall_id] = cls
+            cls._handler_id = hypercall_id
 
         return cls
 
 
 class ExitHandler(metaclass=ExitHandlerMeta):
```
powerjg: This class is where we should add a lot more documentation. We should specify how to subclass it and how to specify the hypercall that it responds to. We should also discuss how when exits happen we create a new instance of the class, etc.

powerjg (changes requested):
Some of my comments go a little beyond the scope of what you were doing, but I think it's important we get the documentation done now.

src/python/gem5/simulate/exit_handler.py
BobbyRBruce: This was a mistake. I got AI to write some better descriptions  for these but it seems to think the arg docstrings can, or should, be removed. I'll re-add these

src/python/gem5/simulate/exit_handler.py
```diff
@@ -24,10 +24,60 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+"""Exit handler helpers for the gem5 Python stdlib.
```
powerjg, line 27: Can you add here something like "hypercalls can originate in either the simulator (e.g., by your model calling `exit...(<hypercall num>)` or from the *guest* application running on the simulator by linking to the m5 library and the guest calling the `m5_hypercall` function."

src/python/gem5/simulate/exit_handler.py
```diff
@@ -643,10 +925,13 @@ def function_generator(func: Callable):
 
     @overrides(ExitHandler)
     def _process(self, simulator: "Simulator") -> None:
-        #  # Translate the exit event cause to the exit event enum.
-        exit_enum = ExitEvent.translate_exit_status(
-            simulator.get_last_exit_event_cause()
-        )
+        # Translate the exit event cause to the exit event enum. New-style
```
powerjg, line 949: Did we delete the old `exit` function? If so, this is OK. If not, I think this will break if someone has a model that uses the old exit function.

src/sim/sim_exit.hh
```diff
@@ -32,44 +32,148 @@
...
-/// and exit_code parameters are saved in the SimLoopExitEvent to
-/// indicate why the exit occurred.
+/**
+ * Legacy “classic” exit path that predates hypercalls. This schedules a
+ * generator-based exit event that reports only a human-readable message and an
+ * optional integer code. Modern stdlib simulations should prefer the
+ * hypercall-based APIs below so Python ExitHandlers can dispatch on a specific
+ * ID and payload.
+ */
+/// DEPRECATED: Prefer exitSimulationLoop (hypercall-based).
```
powerjg: Do you want to use [[deprecated]]

src/systemc/core/scheduler.cc
```diff
@@ -396,7 +398,9 @@ Scheduler::stop()
         if (scMainFiber.finished())
             fatal("Stopping systemc after sc_main completed.");
         else
-            gem5::exitSimLoopNow("systemc stop");
+            exitSimulationLoopNow(ExitHypercall::CLASSIC_GENERATOR,
+                                  std::map<std::string, std::string>{
+                                      {"cause", "systemc pause"}});
```
giactra: This was systemc stop

src/sim/sim_events.cc
```diff
@@ -127,29 +127,30 @@ exitSimLoopNow(const std::string &message, int exit_code, Tick repeat,
...
  */
-void exitSimulationLoop(uint64_t type_id,
-    std::map<std::string, std::string> payload, Tick when)
-{
-    new GlobalSimLoopExitEvent(when, type_id, payload);
-}
+void
+exitSimulationLoop(uint64_t hypercall_id,
+                   std::map<std::string, std::string> payload, Tick when)
+{ new GlobalSimLoopExitEvent(when, hypercall_id, payload); }
```
giactra: Nit and unrelated: is clang-format really pushiing for this?

src/arch/arm/pmu.cc
```diff
@@ -459,14 +459,26 @@ PMU::pmuExitEvent(bool check_reset)
...
+                std::map<std::string, std::string>{
+                    {"cause", "performance counter enabled"}, {"code", "0"}},
+                curTick());
         } else {
             inform("Exiting simulation: PMU disable detected");
-            exitSimLoop("performance counter disabled", 0);
+            exitSimulationLoop(
+                ExitHypercall::CLASSIC_GENERATOR,
+                std::map<std::string, std::string>{
+                    {"cause", "performance counter enabled"}, {"code", "0"}},
```
giactra: performance counter disabled

src/arch/arm/pmu.cc
```diff
@@ -459,14 +459,26 @@ PMU::pmuExitEvent(bool check_reset)
...
+                curTick());
         }
     } else if (check_reset && reg_pmcr.p) {
         inform("Exiting simulation: PMU reset detected");
-        exitSimLoop("performance counter reset", 0);
+        exitSimulationLoop(
+            ExitHypercall::CLASSIC_GENERATOR,
+            std::map<std::string, std::string>{
+                {"cause", "performance counter enabled"}, {"code", "0"}},
```
giactra: performance counter reset

src/arch/arm/pmu.cc
```diff
@@ -738,7 +750,11 @@ PMU::raiseInterrupt()
 {
     if (exitOnPMUInterrupt) {
         inform("Exiting simulation: PMU interrupt detected");
-        exitSimLoop("performance counter interrupt", 0);
+        exitSimulationLoop(
+            ExitHypercall::CLASSIC_GENERATOR,
+            std::map<std::string, std::string>{
+                {"cause", "performance counter enabled"}, {"code", "0"}},
```
giactra: Performance counter interrupt

giactra (changes requested):

src/sim/sim_events.cc
BobbyRBruce: Yip, this is clang-format's work.

src/python/gem5/simulate/exit_handler.py
BobbyRBruce: I'm not sure I get what you mean here. This function still runs `get_last_exit_event_cause()` if that's what you mean

src/python/gem5/simulate/exit_handler.py
```diff
@@ -59,86 +116,232 @@
...
+        A newly defined ``ExitHandler`` subclass registered for ``hypercall``.
     """
 
-    class NewExitHandler(ExitHandler, hypercall_num=hypercall_num):
-        def __init__(self, payload):
+    class NewExitHandler(ExitHandler, hypercall=hypercall):
+        def __init__(self, payload: HypercallPayload) -> None:
             super().__init__(payload)
             self.should_exit = False
             self.description = description
```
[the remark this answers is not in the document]
BobbyRBruce: @copilot apply changes based on [this feedback](https://github.com/gem5/gem5/pull/2983#discussion_r3191109887)

src/sim/pseudo_inst.cc
```diff
@@ -612,8 +614,7 @@ void
 m5Hypercall(ThreadContext *tc, uint64_t hypercall_id)
 {
     DPRINTF(PseudoInst, "pseudo_inst::m5Hypercall(%i)\n", hypercall_id);
-    exitSimLoopWithHypercall("m5_hypercall instruction encountered", 0,
-    curTick(),0, std::map<std::string, std::string>(), hypercall_id, true);
+    exitSimulationLoop(hypercall_id);
```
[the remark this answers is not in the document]
BobbyRBruce: @copilot apply changes based on [this feedback](https://github.com/gem5/gem5/pull/2983#discussion_r3191109808)

src/sim/init_signals.cc
```diff
@@ -391,8 +393,7 @@ processExternalSignal(void)
     munmap(shm_ptr, shared_mem_size);
     close(shm_fd);
 
-    exitSimLoopWithHypercall("Handling external signal!", 0, curTick(), 0,
-                             payload_map, hypercall_id, false);
+    exitSimulationLoop(hypercall_id, std::move(payload_map));
```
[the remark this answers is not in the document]
BobbyRBruce: @copilot apply changes based on [this feedback](https://github.com/gem5/gem5/pull/2983#discussion_r3191109867)

src/python/gem5/simulate/exit_handler.py
powerjg: I see what you mean. It was my confusion.

powerjg:
@BobbyRBruce what is the status of this? let's put this on the list for the 26.0 release

BobbyRBruce:
Status update: this PR has been rebased onto current `develop` (`219748c124`), including a semantic resolution of the newer `MemTest` StashOnce revert, and the updated branch is now at `be7bfb302a`. All review threads have been addressed and resolved, and the PR description has been refreshed. Local validation passed the changed-file pre-commit checks, a full `build/ALL/gem5.opt` build, 109 stdlib pyunit tests, and compiled runtime checks for classic exit payload preservation and malformed-payload rejection. GitHub now reports the branch as mergeable; the remaining CI jobs are running. Ready for re-review once they complete.

powerjg (commented):
The design, etc. look good to me. I spot checked some of the bigger changes, and they lgtm. Once we get the tests passing, I'll take one more quick pass through, but I don't see anything that is likely to be an issue. Thanks for all of the the changes!

BobbyRBruce:
I investigated the failing stdlib pyunit shards. They all reached `CLASSIC_GENERATOR` with an empty payload because the payload could be moved before the cause and code were read, depending on C++ argument-evaluation order. Commit bdabf921f3 extracts both fields before moving the payload. The targeted `sim_events.o` build and changed-file pre-commit checks pass. Commit 9e713d3f95 also addresses the remaining review-body feedback by validating raw IDs as unsigned 64-bit values, rejecting booleans, and exporting the replacement APIs through `m5.event`. Fresh CI is now running.

BobbyRBruce:
> The design, etc. look good to me. I spot checked some of the bigger changes, and they lgtm. Once we get the tests passing, I'll take one more quick pass through, but I don't see anything that is likely to be an issue. Thanks for all of the the changes!

The tests are now passing. You are free to give another pass.

powerjg (approved):

PR #2987, opened 2026-03-04
mem-ruby: Handle separate Comp/DBIDResp in the CHI Controller

giactra:
With this patch we start handling separate Comp+DBIDResp RSP in the tlm::chi::Controller. We were previously handling the joint CompDBIDResp response only

Change-Id: I018991eecaa313c16789638ddfc3e4ba27cf3235

### DIFF
--- a/src/mem/ruby/protocol/chi/tlm/utils.cc
+++ b/src/mem/ruby/protocol/chi/tlm/utils.cc
@@ -378,9 +378,12 @@ rspOpcode(CHIResponseType rsp)
       case CHIResponseType_Comp_UD_PD:
       case CHIResponseType_Comp_UC:
       case CHIResponseType_Comp_I:
-        return RSP_OPCODE_COMP;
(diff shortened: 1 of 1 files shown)


powerjg (dismissed):

giactra:
Hey @powerjg this lost the approval after the rebase

sjain70:
@giactra I wasn’t able to generate Comp + DBIDResp RSP (vs combined CompDBIDResp) in TLM + Gem5 RUBY. Would it be possible for you to share an example or test case that produces this response so I can observe it?

giactra:
> @giactra I wasn’t able to generate Comp + DBIDResp RSP (vs combined CompDBIDResp) in TLM + Gem5 RUBY. Would it be possible for you to share an example or test case that produces this response so I can observe it?

IIRC if comp_wu is set to True, Comp and DBIDResp are sent separately

powerjg (approved):

PR #2990, opened 2026-03-05
arch-arm: Fix crc32cx instruction

powerjg:
Previously, when executing the crc32cx instruction the next PC would be 0.

Because Dest used the legacy IntReg mapping, gem5 incorrectly intercepted the write to register 15 and forcefully treated the crc32cx output as a branch target, thereby writing 0 into the PC. The ARM Architecture Reference Manual dictates that crc32cx writes to a 32-bit accumulator register Wd (which zero-extends into Xd). So, instead of Dest, we must use WDest.

Testing: The code that executed the crc32cx instruction now runs past that instruction.


giactra (approved):
Wow thanks for this big fix

PR #2992, opened 2026-03-09
arch-arm: Implement FEAT_SME2/FEAT_SME2p1

giactra:


powerjg (dismissed):
Great addition! We should definitely highlight in the release notes!

powerjg:
Looks like the failure is real. I believe you need to implement a checkpoint upgrader. Here's the output:

```
warn: Setting the full_system parameter via the Simulator constructor is deprecated and will be removed in future releases of gem5. The board determines if the simulation is full system or not via it's `is_fullsystem` method.
src/mem/dram_interface.cc:691: warn: DRAM device capacity (8192 Mbytes) does not match the address range assigned (32 Mbytes)
src/base/remote_gdb.cc:418: warn: Sockets disabled, not accepting gdb connections
src/sim/globals.cc:91: warn: **********************************************************
src/sim/globals.cc:92: warn: !!! Checkpoint is missing the following version tags:
src/sim/globals.cc:96: warn:   pci-config-pxcap
src/sim/globals.cc:96: warn:   riscv-dyn-vlen
src/sim/globals.cc:96: warn:   riscv-hext
src/sim/globals.cc:96: warn:   riscv-jvt
src/sim/globals.cc:96: warn:   riscv-pcstate
src/sim/globals.cc:96: warn:   riscv-senvcfg
src/sim/globals.cc:96: warn:   riscv-smrnmi
src/sim/globals.cc:96: warn:   riscv-vconf
src/sim/globals.cc:96: warn:   riscv-vext
src/sim/globals.cc:96: warn:   x86-add-xcr0
src/sim/globals.cc:100: warn: You might experience some issues when restoring and should run the checkpoint upgrader (util/cpt_upgrader.py) on your checkpoint
src/sim/globals.cc:103: warn: **********************************************************
src/sim/serialize.hh:452: fatal: fatal condition fixed_size >= 0 && tokens.size() != fixed_size occurred: Array size mismatch on board.processor.cores0.core.xc.0:regs.vector (Got 11264, expected 11520)'
Memory Usage: 333984 KBytes
```

giactra:
Hi @powerjg. I have added a new cpt_updgrader and while doing so I noticed the following:

1) cpt_upgrading is broken for ISAs other than X86 (for older checkpoints of course) after the PR mentioned in https://github.com/gem5/gem5/pull/3002

2) The arm-hello-test-checkpoint https://resources.gem5.org/resources/arm-hello-test-checkpoint?version=1.0.0 needs to be upgraded not just with this contribution; it also misses necessary data for the upgrading to happen which is to have the isaName=arm string on it. My suggestion is to just create a new checkpoint from scratch through the arm-hello-save-checkpoint test. Sure we would miss the testing of the current cpt_upgrader but I wonder if it still makes sense to test checkpoint restore with a checkpoint which is rather old and misses most recent changes. @Harshil2107 what do you think?

powerjg (dismissed):

Harshil2107:
> 2. The arm-hello-test-checkpoint https://resources.gem5.org/resources/arm-hello-test-checkpoint?version=1.0.0 needs to be upgraded not just with this contribution; it also misses necessary data for the upgrading to happen which is to have the isaName=arm string on it. My suggestion is to just create a new checkpoint from scratch through the arm-hello-save-checkpoint test. Sure we would miss the testing of the current cpt_upgrader but I wonder if it still makes sense to test checkpoint restore with a checkpoint which is rather old and misses most recent changes. @Harshil2107 what do you think?

I agree that it is better to take a new checkpoint and update the version because in my opinion manually updating an exisiting (old) checkpoint to add isaName and then running the cpt upgrader doesnot seem right way to do it.

I will upload the new checkpoint. Is it ok if I were to push a commit to this PR updating the test file for the checkpoint test once I have uploaded the new checkpoint?

giactra:
> 
> I will upload the new checkpoint. Is it ok if I were to push a commit to this PR updating the test file for the checkpoint test once I have uploaded the new checkpoint?

Sure works for me! Thanks

Harshil2107:
> > I will upload the new checkpoint. Is it ok if I were to push a commit to this PR updating the test file for the checkpoint test once I have uploaded the new checkpoint?
> 
> Sure works for me! Thanks

I have uploaded the new checkpoint https://resources.gem5.org/resources/arm-hello-test-checkpoint?database=gem5-resources but I dont have permission to push to this PR.
Can you update the following line to make the checkpoint tests use the new checkpoint.

https://github.com/gem5/gem5/blob/7a2b0e413d06c5ce7097104abef3b1d9eaabca91/tests/gem5/checkpoint_tests/configs/arm-hello-restore-checkpoint.py#L72
to 
```
"arm-hello-test-checkpoint", resource_version="2.0.0"
```

giactra:
> > > I will upload the new checkpoint. Is it ok if I were to push a commit to this PR updating the test file for the checkpoint test once I have uploaded the new checkpoint?
> > 
> > 
> > Sure works for me! Thanks
> 
> I have uploaded the new checkpoint https://resources.gem5.org/resources/arm-hello-test-checkpoint?database=gem5-resources but I dont have permission to push to this PR. Can you update the following line to make the checkpoint tests use the new checkpoint.
> 
> https://github.com/gem5/gem5/blob/7a2b0e413d06c5ce7097104abef3b1d9eaabca91/tests/gem5/checkpoint_tests/configs/arm-hello-restore-checkpoint.py#L72
> 
> to
> 
> ```
> "arm-hello-test-checkpoint", resource_version="2.0.0"
> ```

Done, thanks!

giactra:
@powerjg could you re-approve this?

powerjg (approved):

PR #2994, opened 2026-03-09
arch-riscv: Implement several RVV integer instructions

amatabsc:
This patch completes and fixes several RVV (RISC-V Vector Extension) integer instructions in the decoder.

Main changes:
- Integrate the correct vector rounding mode (VXRM) into integer rounding operations.
- Improve saturation handling by setting the VXSAT flag when saturation occurs, following the RVV specification.

Instructions affected:
- vsmul
- vnclip / vnclipu
- vaadd / vaaddu
- vasub / vasubu
- vssrl

### DIFF
--- a/src/arch/riscv/isa/decoder.isa
+++ b/src/arch/riscv/isa/decoder.isa
@@ -3817,16 +3817,19 @@ decode QUADRANT default Unknown::unknown() {
                     0x27: vsmul_vv({{
                         vi max = std::numeric_limits<vi>::max();
                         vi min = std::numeric_limits<vi>::min();
-                        bool overflow = Vs1_vi[i] == Vs2_vi[i] &&
-                                        Vs1_vi[i] == min;
+                        bool overflow = (Vs1_vi[i] == Vs2_vi[i] &&
+                                        Vs1_vi[i] == min);
                         __int128_t result = (__int128_t)Vs1_vi[i] *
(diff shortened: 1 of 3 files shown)


powerjg:
@saul44203 could you take a look at this PR?

src/arch/riscv/isa/decoder.isa
```diff
@@ -3817,16 +3817,19 @@ decode QUADRANT default Unknown::unknown() {
...
-                            result, 0 /* TODO */, sew - 1);
+                            result,
+                            xc->readMiscReg(MISCREG_VXRM),
+                            sew - 1
+                        );
                         result = result >> (sew - 1);
                         if (overflow) {
                             result = max;
-                            *vxsatptr = true;
+                            xc->setMiscReg(MISCREG_VXSAT, 1);
```
saul44203: Maybe we should keep using `vxsatptr`, as from what I understand it's used by `VxsatMicroInst` (the last µop for instructions with saturation) to set VXSAT and mirror it to VCSR.

saul44203 (changes requested):
Only one minor thing, otherwise LGTM.

src/arch/riscv/isa/decoder.isa
amatabsc: Sure!

saul44203 (approved):

powerjg (approved):

PR #2998, opened 2026-03-10
Updates to SE mode to run more Arm/x86/RISC-V binaries

powerjg:


src/arch/arm/isa.cc
```diff
@@ -691,10 +694,21 @@ ISA::readMiscReg(RegIndex idx)
...
               return getGenericTimer().readMiscReg(idx);
           } else {
-              warn("Call to %s attempts to access a system timer which is "
-                   "inaccessible within SE mode. Divergent behaviour is "
-                   "possible.",
-                   miscRegName[idx]);
+              // Intercept specific timer reads for SE mode
+              if (idx == MISCREG_CNTFRQ_EL0 || idx == MISCREG_CNTFRQ) {
+                  // Return the simulator's global frequency (ticks per second)
+                  return sim_clock::Frequency / 1000;
```
giactra: We are actually "lucky". The system counter frequency used to be an implementation defined number between 1MHz and and 50MHz (I believe we are using 24MHz in GenericTimer.py).

Arm architecture now mandates it to be fixed at 1GHz

https://developer.arm.com/documentation/102379/0104/What-is-the-Generic-Timer-

I suppose we can safely fix this as 1GHz.

giactra (commented):
Hey Jason, thanks for this. It looks like though this does not interest Arm only. Could you change the PR title accordingly?

powerjg:
> Hey Jason, thanks for this. It looks like though this does not interest Arm only. Could you change the PR title accordingly?

Done. The commits were tagged correctly :)

powerjg:
Slight update on one syscall I missed earlier (keeping track of 64 vs 32 bit was difficult while rebasing).

giactra:
> > Hey Jason, thanks for this. It looks like though this does not interest Arm only. Could you change the PR title accordingly?
> 
> Done. The commits were tagged correctly :)

Thanks! Please review my other comment here: https://github.com/gem5/gem5/pull/2998#discussion_r2933264632

src/arch/arm/isa.cc
powerjg: Done. Thanks for the suggestion.

giactra (approved):

PR #3006, opened 2026-03-12
arch-riscv: Improve error for incompatible ISA in RISC-V

powerjg:
Previously, there would be a segfault if the wrong ISA was provided. Now, there is an understandable error message


Harshil2107 (approved):
lgtm

PR #3019, opened 2026-03-18
cpu: Use value-initialization for TimeBuffer entries

BobbyRBruce:
Replace the memset-plus-placement-new pattern with `new (ptr) T()`. This avoids relying on all-bits-zero as a valid initial state for arbitrary `T` and uses the language-defined initialization semantics instead.

`fetchAddr` in `ForwardLineData` was dependent on the zeroing behavior. Given this was not an explicit contract of `TimeBuffer`, this change explicitly zero's `fetchAddr` in default construction to avoid undefined behavior.

### DIFF
--- a/src/cpu/minor/pipe_data.hh
+++ b/src/cpu/minor/pipe_data.hh
@@ -200,7 +200,7 @@ class ForwardLineData /* : public ReportIF, public BubbleIF */
     std::unique_ptr<PCStateBase> pc;
 
     /** Address of this line of data */
-    Addr fetchAddr;
+    Addr fetchAddr = 0;
 
     /** Explicit line width, don't rely on data.size */
     unsigned int lineWidth = 0;
--- a/src/cpu/timebuf.hh
+++ b/src/cpu/timebuf.hh
@@ -30,6 +30,7 @@
 #define __BASE_TIMEBUF_HH__
 
 #include <cassert>
+#include <type_traits>
 #include <vector>
 
 namespace gem5
@@ -140,6 +141,8 @@ class TimeBuffer
         : past(p), future(f), size(past + future + 1),
           data(new char[size * sizeof(T)]), index(size), base(0)
     {
+        static_assert(std::is_default_constructible_v<T>,
+                      "T must be default constructible.");
         assert(past >= 0 && future >= 0);
         char *ptr = data;
         for (unsigned i = 0; i < size; i++) {
--- a/src/cpu/timebuf.hh
+++ b/src/cpu/timebuf.hh
@@ -30,7 +30,6 @@
 #define __BASE_TIMEBUF_HH__
 
 #include <cassert>
-#include <cstring>
 #include <vector>
 
 namespace gem5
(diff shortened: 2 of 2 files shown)


powerjg (changes requested):
Correct me if I'm wrong, but I'm worried that `T` will have uninitialized values. I believe that `T` is one of the structs here: `src/cpu/o3/comm.hh` or `src/cpu/minor/pipe_data.hh`. Some of those structs have uninitialized members.

You may be able to add something like the following to check this at compile time.

```
static_assert(std::is_default_constructible_v<T>, 
                  "T must be default constructible.");
```

BobbyRBruce:
> Correct me if I'm wrong, but I'm worried that `T` will have uninitialized values. I believe that `T` is one of the structs here: `src/cpu/o3/comm.hh` or `src/cpu/minor/pipe_data.hh`. Some of those structs have uninitialized members.
> 
> You may be able to add something like the following to check this at compile time.
> 
> ```
> static_assert(std::is_default_constructible_v<T>, 
>                   "T must be default constructible.");
> ```

You are right though i did an audit of the codebase and found `fetchAddr` in `ForwardLineData` (in `src/cpu/minor/pipe_data.hh) to be the only case where we didn't have explicit default instantiation. 

I've updated this PR to run on the following assumptions/views:

1. TimeBuffer zero'ing the memory was never a contract it guaranteed and is not something we should be guaranteeing as it incentivizes bad design. It should therefore be removed.
2. We should always being avoiding undefined behavior and never excuse it via zero'ing of memory done elsewhere. It's therefore good to update `ForwardLineData` regardless.

I've added the static_assert. It wouldn't have caught this exact instance but it would help catch incorrect use of `TimeBuffer`. It solidifies the contract.

powerjg (approved):
Thanks! This lgtm. Looks like the test failure is unrelated. Can we pin the version of xcode/etc in our CI tests?

BobbyRBruce:
> Thanks! This lgtm. Looks like the test failure is unrelated. Can we pin the version of xcode/etc in our CI tests?

I can't think of an easy way or one I'd like to depend on. I use brew and brew deliberately make it hard to fix versions or get older versions of packages. So, if depending on brew, the environment our cached Mac build dir was built in is never guaranteed to be the same env as we use in our CI tests. I can only think of the following alternatives:

1. Avoid use of `brew` and install using a script with fixed versions (likely, for each dependency: download the tarball from Git, configure and install the old-fashioned way) then have this script setup the env for both caching and the CI building. --- a lot of scripting and may run into dependency hell pretty quickly. It also wouldn't make it easy to change versions.
2. Have it so a failed CI building on top of a cached build dir simply wipes the build dir and starts again. I.e., every CI run get's a 2nd chance at Mac compilation from scratch --- Simple but means we do a lot of unnecessary compilation. If the compilation bug is genuinely introduced by the PR it means we build using the cache, then build again to just run into the same compilation error. It also means when this env mismatch scenario occurs we have a whole day where we do a "run-fail-wipe-rerun" for every CI run until the cached build dir is rebuilt in the newer env. This scares me because Mac Action runners are very expensive. It's only cheap now due to the heavy use of caching.
3. Protobuf appears to be the main culprit here. I've not seen this happen with the other dependencies. As it's optional we could drop it -- Easiest solution but reduces the scope of our testing and means we never test if protobuf compiles on Mac hardware. It also doesn't stop the other dependencies causing this issue.

powerjg:
Makes sense. I was hoping there would be a way like docker where you can fix the base, but I see that wouldn't work with the way brew works.

PR #3022, opened 2026-03-20
 cpu-o3: Release no-access LSQ requests safely

BobbyRBruce:
This change fixes a `DynInst` lifetime leak observed during long O3 x86
boot runs (failed with assert
`DynInst count on board.processor.cores7.core exceeded 1500`). When the MMU
indicates a memory access requires no access (i.e. the access is
short-circuited and no packet is sent), the `LSQ` can retain the
corresponding `LSQRequest` pointer for the `lq`/`sq` entry. Because no
response ever arrives, the normal cleanup path that would release the
`DynInst` is not invoked, allowing the `LSQRequest` to keep the
`DynInstPtr` alive and causing `DynInst` accumulation.

See the short-circuit detection in `LSQ::pushRequest()`:
https://github.com/gem5/gem5/blob/a403abb67942e92b5c51229935ce0fb0fbcc1e6f/src/cpu/o3/lsq.cc#L838-L842

and the load execute handoff in `LSQUnit::executeLoad()`:
https://github.com/gem5/gem5/blob/a403abb67942e92b5c51229935ce0fb0fbcc1e6f/src/cpu/o3/lsq_unit.cc#L636-L642

This commit changes:

- For `store` short-circuited accesses we clear the store queue entry
  (`thread[tid]->storeQueue[inst->sqIdx].setRequest(nullptr)`) and call
  `request->discard()` in `LSQ::pushRequest()`. Stores do not require an
  execute-stage load completion handoff, so they are safe to release
  immediately.

- For `load` short-circuited accesses we retain the `LSQRequest` until
  `LSQUnit::executeLoad()` completes the execute-stage handoff (mark the
  instruction executed, call `completeAcc(nullptr)`, forward the
  instruction to commit, and record activity). After that handoff
  completes we clear the load queue's request pointer
  (`loadQueue[inst->lqIdx].setRequest(nullptr)`) and call
  `request->discard()`. That ensures the `LSQRequest` does not keep the
  `DynInst` alive after the instruction is done.

This works because clearing the LSQ's pointer and discarding the
`LSQRequest` removes the last reference keeping the `DynInst` live.
Immediate discard is safe for stores because there is no later
execute-stage work. Loads require the execute-stage handoff, so
deferring discard until after `LSQUnit::executeLoad()` ensures the
instruction's normal completion path runs first, and then the request is
released.

### DIFF
--- a/src/cpu/o3/lsq.cc
+++ b/src/cpu/o3/lsq.cc
@@ -835,11 +835,16 @@ LSQ::pushRequest(const DynInstPtr& inst, bool isLoad, uint8_t *data,
             // (e.g. re-exec).
             if (fault != NoFault)
                 inst->getFault() = fault;
-        } else if (isLoad) {
+        } else {
             inst->setMemAccPredicate(false);
-            // Commit will have to clean up whatever happened.  Set this
-            // instruction as executed.
-            inst->setExecuted();
+            if (isLoad) {
+                inst->setExecuted();
+            } else {
+                if (inst->sqIdx >= 0) {
+                    thread[tid]->storeQueue[inst->sqIdx].setRequest(nullptr);
+                }
+                request->discard();
+            }
         }
     }
 
--- a/src/cpu/o3/lsq_unit.cc
+++ b/src/cpu/o3/lsq_unit.cc
@@ -639,6 +639,17 @@ LSQUnit::executeLoad(const DynInstPtr &inst)
         inst->completeAcc(nullptr);
         iewStage->instToCommit(inst);
         iewStage->activityThisCycle();
+        if (inst->lqIdx >= 0) {
+            LSQRequest *request = loadQueue[inst->lqIdx].request();
+            if (request) {
+                // No packet or translation response will arrive for this
+                // short-circuited access.  Drop the LSQ's request pointer here
+                // after the instruction has completed its execute-stage work,
+                // so the request no longer keeps the DynInst alive.
+                loadQueue[inst->lqIdx].setRequest(nullptr);
+                request->discard();
+            }
+        }
         return NoFault;
     }
 


giactra:
Thanks @BobbyRBruce I will have a look at this. Is crazy that we allowed https://github.com/gem5/gem5/commit/f7cb4d8d1fb10eb0c1df4b221379cb2e32aac356 to be merged considering it sneaked functional changes within a "cosmetic" refactoring.

Could you point me in the original PR which were the changes that changed ownership?

powerjg:
I don't know about all of this.

I will take the responsibility for https://github.com/gem5/gem5/commit/f7cb4d8d1fb10eb0c1df4b221379cb2e32aac356, but if there was a bug introduced there, it was certainly accidental. I just looked through all of the code again, and by my eye it looks OK.

The **problem** here is the lack of consistent testing. We get into these states where the daily or weekly tests start failing and we never actually fix the root cause. Then, new changes can cause additional bugs (which might have been what happened here).

The root cause of the O3 bugs is way before https://github.com/gem5/gem5/commit/f7cb4d8d1fb10eb0c1df4b221379cb2e32aac356. There was a change to get the RISC-V vector instructions to work, and we haven't had passing tests since then.

This PR doesn't fix things either, unfortunately. 

See #2930, #2657, #2891, and #2914 for more context. This whole thing is a big mess. If we want to put blame on merging anything, I think it should probably be the initial issue which was PR #2657 ([line of code that's the problem](https://github.com/gem5/gem5/pull/2657/changes#diff-1a693d4c25673fc17e8477344df90185a81a7ba96e3acd01a8ba30f8c2cf2f86R1332))

src/cpu/o3/lsq.cc
```diff
@@ -835,11 +852,20 @@ LSQ::pushRequest(const DynInstPtr& inst, bool isLoad, uint8_t *data,
...
+                if (inst->lqIdx >= 0) {
+                    thread[tid]->loadQueue[inst->lqIdx].setRequest(nullptr);
+                }
+            } else {
+                if (inst->sqIdx >= 0) {
+                    thread[tid]->storeQueue[inst->sqIdx].setRequest(nullptr);
+                }
+            }
+            request->discard();
+            discarded = true;
```
powerjg: This line of code does nothing, right?

BobbyRBruce: Correct. It's pointless. I have removed it.

giactra:
> See #2930, #2657, #2891, and #2914 for more context. This whole thing is a big mess. If we want to put blame on merging anything, I think it should probably be the initial issue which was PR #2657 ([line of code that's the problem](https://github.com/gem5/gem5/pull/2657/changes#diff-1a693d4c25673fc17e8477344df90185a81a7ba96e3acd01a8ba30f8c2cf2f86R1332))

I thought that PR was reverted... IMHO the only way to make it work is to create explicit data dependency via register and stop modelling that instruction as a branch.

erin-le:
@giactra, @powerjg: Hello, just a reminder to review this PR when possible, as it should fix the [Weekly Test failures](https://github.com/gem5/gem5/actions/runs/26306722286/job/77445429492). Thank you!

powerjg (approved):

erin-le (approved):
I don't know enough about the O3 CPU to give this a proper review, but style LGTM

PR #3026, opened 2026-03-22
arch-x86: x87 FPU movfp/lfpimm micro-ops use wrong dataSize

hakan-demirli:
x87 FPU macroop definitions (constant loads, register loads/stores, exchanges, trig functions) do not specify `dataSize=8` for their `movfp` and `lfpimm` micro-ops. These micro-ops inherit `env.dataSize`, which defaults to 4 (32-bit operand size) in 64-bit mode for x87 instructions. When `dataSize=4`, the `movfp` micro-op does a 32-bit partial write (read-modify-write preserving the upper 32 bits of the destination), which silently drops the upper 32 bits of the source value.

This means instructions like `FLD1`, `FLDPI`, `FLD`, `FST`, `FXCH`, etc. produce incorrect results on any CPU model that goes through `SimpleThread::setReg` for register writes (AtomicSimpleCPU, TimingSimpleCPU, MinorCPU).

Test to reproduce:
```asm
global _start
section .text
_start:
    finit
    fld1                    ; push 1.0
    sub rsp, 16
    fstp qword [rsp]        ; pop st(0) to [rsp]
    mov rax, [rsp]          ; rax = the double value
    add rsp, 16
    ; rax should be 0x3FF0000000000000
    ; compare with expected
    mov rbx, 0x3FF0000000000000
    cmp rax, rbx
    jne .fail
    ; pass: exit 0
    mov eax, 60
    xor edi, edi
    syscall
.fail:
    ; fail: exit 1
    mov eax, 60
    mov edi, 1
    syscall
```

----------

Kinda related to https://github.com/gem5/gem5/pull/2779

### DIFF
--- a/src/arch/x86/isa/insts/x87/data_transfer_and_conversion/convert_and_load_or_store_integer.py
+++ b/src/arch/x86/isa/insts/x87/data_transfer_and_conversion/convert_and_load_or_store_integer.py
@@ -39,14 +39,14 @@ microcode = """
 # fild common case
 def macroop FILD_M {
     ldifp87 ufp1, seg, sib, disp
-    movfp st(-1), ufp1, spm=-1
+    movfp st(-1), ufp1, spm=-1, dataSize=8
 };
 
 # fild with RIP-relative addressing
 def macroop FILD_P {
     rdip t7
     ldifp87 ufp1, seg, riprel, disp
-    movfp st(-1), ufp1, spm=-1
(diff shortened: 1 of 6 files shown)


powerjg (approved):

erin-le:
Note: The gem5 devs are going to hold off on merging this until the X86 very-long Weekly tests are passing again, which shouldn't take too long.

PR #3028, opened 2026-03-24
misc: Update to use c++20

ylldummy:
GCC will soon enable C++20 as default [ref](https://gcc.gnu.org/projects/cxx-status.html) and some style guide (e.g. the Google one [ref](https://google.github.io/styleguide/cppguide.html)) suggest using C++20 already.

This is a trial of enabling it in gem5. Would like to check if this makes sense or any concern.

Tested locally with clang-14 and GCC-12

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -603,13 +603,6 @@ for variant_path in variant_paths:
         with gem5_scons.Configure(env) as conf:
             conf.CheckCxxFlag('-Wno-volatile')
 
-        # Arithmetic operations between differen enums are deprecated.
-        # https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1120r0.html
-        # However, this is common used in code under arch/.
-        # e.g. ArmFault::AddressSizeLL + start_lookup_level
-        env.Append(CXXFLAGS=['-Wno-deprecated-enum-enum-conversion'])
-        env.Append(CXXFLAGS=['-Wno-deprecated-anon-enum-enum-conversion'])
-
         if sys.platform.startswith('freebsd'):
             env.Append(CCFLAGS=['-I/usr/local/include'])
             env.Append(CXXFLAGS=['-I/usr/local/include'])
--- a/SConstruct
+++ b/SConstruct
@@ -594,8 +594,21 @@ for variant_path in variant_paths:
         env.Append(CCFLAGS=['-Wall', '-Wundef', '-Wextra',
                             '-Wno-sign-compare', '-Wno-unused-parameter'])
 
-        # We always compile using C++17
-        env.Append(CXXFLAGS=['-std=c++17'])
+        # We always compile using C++20
+        env.Append(CXXFLAGS=['-std=c++20'])
+        # Left operand of volatile is deprecated in C++20 and then
+        # de-deprecaetd. This skip is a workaround to avoid warning on
+        # intermediate compiler versions. Ref:
+        # https://cplusplus.github.io/CWG/issues/2654.html
+        with gem5_scons.Configure(env) as conf:
+            conf.CheckCxxFlag('-Wno-volatile')
+
+        # Arithmetic operations between differen enums are deprecated.
+        # https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1120r0.html
+        # However, this is common used in code under arch/.
+        # e.g. ArmFault::AddressSizeLL + start_lookup_level
+        env.Append(CXXFLAGS=['-Wno-deprecated-enum-enum-conversion'])
+        env.Append(CXXFLAGS=['-Wno-deprecated-anon-enum-enum-conversion'])
 
         if sys.platform.startswith('freebsd'):
             env.Append(CCFLAGS=['-I/usr/local/include'])
@@ -687,6 +700,9 @@ for variant_path in variant_paths:
                 f'to v{gcc_max_version}.\n'
             )
 
+        # Workaround https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105651
+        if compareVersions(gcc_version, "13") < 0:
+            env.Append(CXXFLAGS=['-Wno-restrict'])
 
         # Add the appropriate Link-Time Optimization (LTO) flags if
         # `--with-lto` is set.
--- a/ext/softfloat/SConscript
+++ b/ext/softfloat/SConscript
@@ -47,6 +47,7 @@ if sf_env['GCC']:
 elif sf_env['CLANG']:
     sf_env.Append(CCFLAGS=['-Wno-unused-variable',
                            '-Wno-unused-label',
+                           '-Wno-implicit-fallthrough',
                            '-Wno-implicit-function-declaration',
                            '-g'])
 
--- a/ext/softfloat/SConscript
+++ b/ext/softfloat/SConscript
@@ -40,14 +40,12 @@ sf_env = env.Clone()
 if sf_env['GCC']:
     sf_env.Append(CCFLAGS=['-Wno-unused-variable',
                            '-Wno-unused-label',
-                           '-Wno-implicit-fallthrough',
                            '-Wno-implicit-function-declaration',
                            '-g'])
 
 elif sf_env['CLANG']:
     sf_env.Append(CCFLAGS=['-Wno-unused-variable',
                            '-Wno-unused-label',
-                           '-Wno-implicit-fallthrough',
                            '-Wno-implicit-function-declaration',
                            '-g'])
 
--- a/src/arch/amdgpu/common/dtype/binary32.hh
+++ b/src/arch/amdgpu/common/dtype/binary32.hh
@@ -42,32 +42,29 @@ namespace AMDGPU
 // this format by default. For now as there do not seem to be any MI300
 // instructions operating directly on the types (i.e., they all cast to FP32
 // first and then perform arithmetic operations).
-typedef union binary32_u
+union binary32
 {
-    enum bitSizes
-    {
-        ebits = 8,
-        mbits = 23,
-        sbits = 1,
-        bias = 127,
+    static constexpr size_t ebits = 8;
+    static constexpr size_t mbits = 23;
+    static constexpr size_t sbits = 1;
+    static constexpr size_t bias = 127;
 
-        inf = 0x7f800000,
-        nan = 0x7f800100,
-        max = 0x7f7fffff
-    };
+    static constexpr uint32_t inf = 0x7f800000;
+    static constexpr uint32_t nan = 0x7f800100;
+    static constexpr uint32_t max = 0x7f7fffff;
 
     uint32_t storage;
     float fp32;
     struct
     {
-        unsigned mant : 23;
-        unsigned exp : 8;
-        unsigned sign : 1;
+        unsigned mant : mbits;
+        unsigned exp : ebits;
+        unsigned sign : sbits;
     };
 
     // To help with stdlib functions with T = float.
     operator float() const { return fp32; }
-} binary32;
+};
 static_assert(sizeof(binary32) == 4);
 
 } // namespace AMDGPU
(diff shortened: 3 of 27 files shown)


ext/softfloat/SConscript
```diff
@@ -47,6 +47,7 @@ if sf_env['GCC']:
 elif sf_env['CLANG']:
     sf_env.Append(CCFLAGS=['-Wno-unused-variable',
                            '-Wno-unused-label',
+                           '-Wno-implicit-fallthrough',
```
powerjg: We should fix the implicit fallthroughs instead of removing the warning.

src/dev/arm/css/scmi_protocols.cc
```diff
@@ -267,8 +267,7 @@ BaseProtocol::discoverAgent(Message &msg)
 
         agent_size = agent_name.length();
 
-        strncpy((char *)&payload.name,
-                agent_name.c_str(), agent_size);
+        strncpy((char *)&payload.name, agent_name.c_str(), agent_size + 1);
```
powerjg, line 270: Is this changing the the behavior? Is `agent_size + 1` one character too long?

SConstruct
```diff
@@ -594,8 +594,14 @@ for variant_path in variant_paths:
...
 
-        # We always compile using C++17
-        env.Append(CXXFLAGS=['-std=c++17'])
+        # We always compile using C++20
+        env.Append(CXXFLAGS=['-std=c++20'])
+        # Arithmetic operations between differen enums are deprecated.
+        # https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1120r0.html
+        # However, this is common used in code under arch/.
+        # e.g. ArmFault::AddressSizeLL + start_lookup_level
+        env.Append(CXXFLAGS=['-Wno-deprecated-enum-enum-conversion'])
```
powerjg: Again, we should fix this not remove the warnings.

src/arch/arm/insts/vfp.hh
```diff
@@ -378,7 +378,7 @@ vfpFpToFixed(T val, bool isSigned, uint8_t width, uint8_t imm, bool
                 // -(val - origVal). This is required to get the corruct bit
                 // exact behaviour when very close to the 0.5 threshold.
                 volatile T error = val;
-                error -= origVal;
+                error = error - origVal;
```
powerjg: Can you explain why this change is needed?

powerjg (changes requested):

src/dev/arm/css/scmi_protocols.cc
ylldummy: The returned length from std::string would not include the null terminator. However, strncpy's count expects null terminator to be included, that's why a +1 should be included.

The code might work correctly because strncpy:
```
If count is reached before the entire string src was copied, the resulting character array is not null-terminated.
```
And if the buffer is zero-ed initially, that's likely equivalent.

OTOH, I see in the payload name's definition:
```
uint8_t name[Protocol::MAX_STRING_SIZE + 1];
``` 

It does have the +1  size for null terminator, so I think +1 should be the correct behavior.

src/arch/arm/insts/vfp.hh
ylldummy: Sorry, had it in commitmsg but not in the PR messge.

Left operand of volatile is deprecated in C++20 so compiler would complain about the "-=" usage.

SConstruct
ylldummy: Have follow up commits fixing these. PTAL

ext/softfloat/SConscript
ylldummy: Have followup commits fixing this. PTAL

src/arch/arm/insts/vfp.hh
powerjg: Thanks!

src/arch/arm/mmu.cc
```diff
@@ -411,10 +411,10 @@ MMU::checkPermissions(TlbEntry *te, const RequestPtr &req, Mode mode,
     // as a device or strongly ordered.
     if (state.isStage2 && req->isPTWalk() && state.hcr.ptw &&
         (te->mtype != TlbEntry::MemoryType::Normal)) {
-        return std::make_shared<DataAbort>(
-            vaddr, te->domain, is_write,
-            ArmFault::PermissionLL + te->lookupLevel,
-            state.isStage2, tran_method);
+        return std::make_shared<DataAbort>(vaddr, te->domain, is_write,
+                                           (int)ArmFault::PermissionLL +
```
powerjg: @giactra can you take a look at this? 

I understand what the code is doing, but maybe we shouldn't be using the enum values as `int`s here. 

Here's the copilot suggestion. I like this, but Giacomo, I'll let you make a call. (It gave a couple of other suggestions, but I think this is best. The other suggestion that was good was to make FaultSource an enum class. I don't think that's necessary.)


### Option 1 (best fit here): add a helper that returns a `FaultSource`
Centralize the “LL + level” computation in one function (ideally next to the enum, e.g., in the same header as `FaultSource`).

```cpp
#include <type_traits>
#include <cassert>

static inline ArmFault::FaultSource
llFaultSource(ArmFault::FaultSource baseLL, unsigned level)
{
    assert(level < 4);

    using U = std::underlying_type_t<ArmFault::FaultSource>;
    const auto base = static_cast<U>(baseLL);
    return static_cast<ArmFault::FaultSource>(base + static_cast<U>(level));
}
```

Usage at call sites becomes:

```cpp
return std::make_shared<DataAbort>(
    vaddr, te->domain, is_write,
    llFaultSource(ArmFault::PermissionLL, te->lookupLevel),
    state.isStage2, tran_method);
```

Benefits:
- No enum/int arithmetic at call sites (fixes the C++20 warning in the “right” place).
- Makes the “this is an LL base” assumption explicit.
- Gives you a single place to validate `level < 4`.

If you want it even more explicit, provide *named* helpers:

```cpp
static inline ArmFault::FaultSource permissionFault(unsigned level) {
    return llFaultSource(ArmFault::PermissionLL, level);
}
static inline ArmFault::FaultSource domainFault(unsigned level) {
    return llFaultSource(ArmFault::DomainLL, level);
}
```

src/arch/amdgpu/common/dtype/mxfp_convert.hh
```diff
@@ -137,7 +137,7 @@ dFMT convertMXFP(sFMT in, mxfpRoundingMode mode = roundTiesToEven,
             // Save the value for rounding so we don't need to recompute it.
             uint32_t saved_mant = mant;
 
-            mant >>= (sFMT::mbits - dFMT::mbits);
+            mant >>= ((int)sFMT::mbits - (int)dFMT::mbits);
```
powerjg: @abmerop can you take a look at my comment below on the Arm code to see if it applies here?

src/arch/mips/utility.cc
```diff
@@ -132,11 +132,11 @@ uint32_t
 genInvalidVector(uint32_t fcsr_bits)
 {
     //Set FCSR invalid in "flag" field
-    int invalid_offset = Invalid + Flag_Field;
+    int invalid_offset = (int)Invalid + (int)Flag_Field;
```
powerjg, line 135: I think this is OK here. The MIPS code is not used heavily and not well maintained anyway. It's not worth putting much effort in here.

powerjg (commented):
thanks for the updates! I think we're close. Thanks for digging in to the changes needed. I think we should take this opportunity to improve the code instead of just working around the warnings.

src/arch/amdgpu/common/dtype/mxfp_convert.hh
abmerop: I would have to check, but maybe typed enum can avoid the cast. Changing to static constexpr int might work too, I'd have to also check that there is no identifier name collision that way.

src/arch/arm/mmu.cc
ylldummy: Thanks for the suggestion. Applied to the latest push.

src/arch/amdgpu/common/dtype/mxfp_convert.hh
ylldummy: I checked a bit more into the usage here. It seems these "*bits" fields shouldn't be enum actually. It's used for bit length of the related types.

I have a commit refining it to be static constexpr under the types. Hope that makes sense. PTAL.

src/arch/amdgpu/common/dtype/binary32.hh
```diff
@@ -42,35 +42,32 @@ namespace AMDGPU
 // this format by default. For now as there do not seem to be any MI300
 // instructions operating directly on the types (i.e., they all cast to FP32
 // first and then perform arithmetic operations).
-typedef union binary32_u
+union binary32
```
powerjg, line 45: @abmerop can you PTAL?

powerjg (commented):
lgtm. I'll save some time for others to review.

src/arch/amdgpu/common/dtype/binary32.hh
ylldummy: friendly ping :)

src/arch/arm/insts/vfp.hh
lone-wolf-akela: FYI, C++ standard committee later revert the decision of deprecating left operand of volatile.
see https://cplusplus.github.io/CWG/issues/2654.html

ylldummy: Thanks for the reference. Yeah, I did see the de-deprecation, but IIUC, it's effective in C++23 so I feel it might still be safer to change it here so we don't get warning/error regardless of C++/toolchain version used.

lone-wolf-akela: > Thanks for the reference. Yeah, I did see the de-deprecation, but IIUC, it's effective in C++23 so I feel it might still be safer to change it here so we don't get warning/error regardless of C++/toolchain version used.

As it shows in that page, this change is accepted as a DR(Defect Report), so technically it should also apply to older c++ version including c++20. Though in reality, this only means compiler of **newer versions** will not output that warning when you set it to c++20 mode (as we cannot change the behaviour of any already released compiler of older versions).

If it were me, I would add a `-Wno` flag to the compilation options to silence this warning for those older compilers instead of changing the code. But I'm not saying changing the code is wrong, it's just personal preference。

powerjg: +1 for reverting this change and adding the `-Wno<whatever>` to the global flags **with a comment on why we are are ignoring this warning**.

erin-le:
@ylldummy: Hello, thanks for this PR! Could you modify some of the commit titles and the PR title to better fit with gem5's style guide?

For the commit titles, could you change `arch: riscv:` to `arch-riscv:`, `arch: arm:` to `arch-arm:`, and `arch: amdgpu:` to `arch-vega:`?

For the PR title, could you change `RFC` to one or more of the labels in the [Maintainers file](https://github.com/gem5/gem5/blob/stable/MAINTAINERS.yaml)?

Thank you!

src/arch/arm/insts/vfp.hh
ylldummy: Updated.

ylldummy:
Updated the commit titles

src/arch/amdgpu/common/dtype/binary32.hh
ylldummy: @abmerop friendly ping again :)

abmerop: Should be fine as long as the static_assert did not trigger

ylldummy:
Rebase to resolve conflict.

powerjg (approved):

powerjg:
@ylldummy thanks for your effort on this! I re-ran the failing job (I think it was just a flake), so we should be good to merge soon. Thanks again!

PR #3031, opened 2026-03-25
configs,stdlib: Disable ASLR for GPU

abmerop:
This change disables address space layout randomization (ASLR) and has two benefits:
 - For debugging, GPU virtual addresses should be the same from run to run which makes greping based on virtual address easier
 - The simulation is more deterministic as units like TLB and page table walker do not vary due to different virtual page numbers each run.


powerjg (approved):

mattsinc (approved):
@v-ramadas : FYI

PR #3034, opened 2026-03-26
gpu-compute: Remove unncessary `virtual` specifier

BobbyRBruce:
Clang 21 added `-Wunnecessary-virtual-specifier` (a warning that is thrown as an error when compiling on the `develop` branch). It is catches this instance in "gpu_compute_driver.hh" where the `virtual` specifier is unnecessary. This patch remove this.


mattsinc (approved):
This seems reasonable to me, but @abmerop any reason you want to keep this?

abmerop (approved):

PR #3042, opened 2026-03-27
cpu-o3: fix renameSrcRegs crash on InvalidRegClass src reg

### LINKED ISSUE #3041: cpu-o3: segfault in Rename::renameSrcRegs on InvalidRegClass src reg
## Description

gem5 O3 CPU crashes with a segmentation fault in `Rename::renameSrcRegs`
when an instruction with an `InvalidRegClass` source register reaches the
scoreboard check.

Crash is deterministic and reproducible in RISC-V FS mode with 8 O3
cores.

## Backtrace

```
  gem5 has encountered a segmentation fault!
  --- BEGIN LIBC BACKTRACE ---
  gem5.opt(print_backtrace+0x19)
  gem5.opt(+0x1c7adef)
  libc.so.6(+0x3c050)
  gem5.opt(Rename::renameSrcRegs+0x117)
  gem5.opt(Rename::renameInsts+0x48a)
  gem5.opt(Rename::rename+0x60)
  gem5.opt(Rename::tick+0x87)
  gem5.opt(CPU::tick+0x13d)
  gem5.opt(EventQueue::serviceOne+0xa5)
  gem5.opt(doSimLoop+0x58)
  gem5.opt(simulate+0x23e)
  --- END LIBC BACKTRACE ---
```
## Root Cause

In `renameSrcRegs`, the switch on `flat_reg.classValue()` has:

```
      case InvalidRegClass:
          break;
```

The break exits the switch but NOT the for loop, so control falls
through to `scoreboard->getReg(renamed_reg)`. 
Calling `scoreboard->getReg(renamed_reg)` on `invalidPhysRegId` causes a crash.

  ## Proposed Fix

In the `InvalidRegClass` case, map the `src` slot to `invalidPhysRegId`
and mark it ready directly, then continue past the scoreboard:

```
    case InvalidRegClass:
        inst->renameSrcReg(src_idx, renamed_reg);
        inst->markSrcRegReady(src_idx);
        continue;
```

A PR with the fix will follow.

polpetras:
cpu-o3: fix renameSrcRegs crash on InvalidRegClass src reg

case InvalidRegClass: break exits the switch but not the for loop,
causing scoreboard->getReg to be called on invalidPhysRegId.
Fix by mapping the src slot to invalidPhysRegId and marking it
ready directly, bypassing the scoreboard.

GitHub Issue: https://github.com/gem5/gem5/issues/3041


giactra:
I remember years ago complaining about the introduction of the InvalidRegClass, claiming it would only add confusion and it would be abused. Is there a reason why you want an instruction renaming an invalid register to *not* crash? I think the real solution is: why an instruction is marking a destination register as invalid?

polpetras:
## On "is there a reason why you want an instruction renaming an invalid register to not crash"

Yes. `InvalidRegClass` source operands can appear intentionally (see below). The issue fixed in this patch is a control-flow bug in `renameSrcRegs`: `case InvalidRegClass: break` exits only the `switch`, not the enclosing `for` loop, so execution falls through into the normal scoreboard-readiness path.

For this case, `map->lookup()` returns `&invalidPhysRegId`, which is a sentinel (not a real physical register). Source-readiness scoreboard lookup should be bypassed for that sentinel. The fix does this by using `continue` in the `InvalidRegClass` case, after renaming the source and marking it ready.

This patch removes the crash on our reproducer (O3 + 8 cores + RVV + stream); this is targeted validation, not a full regression proof.

---

## On "why is an instruction marking a destination register as invalid"

This is by design in the RISC-V decoder. In `src/arch/riscv/isa/operands.isa:54-66`, `IntReg.regId()` emits `RegId()` when the integer operand index is zero:

```python
def regId(self):
    return f'(({self.reg_spec}) == 0) ? RegId() : {self.reg_class}[{self.reg_spec}]'
```

`RegId()` default-constructs to `invalidRegClass`, so an operand resolving to `x0` can legitimately become `InvalidRegClass` on either side:

- destination: writes to `x0` are discarded
- source: reads from `x0` have no producer and are immediately ready

`renameDstRegs` already handles this correctly via `map->rename()`/`isRenameable()`. The bug was only in the `renameSrcRegs` fall-through path.

I understand the general concern about InvalidRegClass misuse; here we're only fixing a specific renameSrcRegs control-flow bug for ISA-defined InvalidRegClass operands.

polpetras:
I think 4-cores setup is still crashing with almost identical error, so let me check in more detail and will get back to you.

polpetras:
Reopening as draft.  I could not reproduce the crash on latest develop, so keeping as draft for visibility and further review.

polpetras:
Could not reproduce it, So I close this.

PR #3044, opened 2026-03-28
cpu: Incorrect eviction in SimpleBTB

dhschall:
The BTB is updated without checking if the current entry already exists. However, the `findVictim` function will not check if it already exists. This creates duplication.
In some benchmarks, we observed more than 30% increase in misses.

### DIFF
--- a/src/cpu/pred/simple_btb.cc
+++ b/src/cpu/pred/simple_btb.cc
@@ -117,10 +117,16 @@ SimpleBTB::update(ThreadID tid, Addr instPC,
 {
     stats.updates[type]++;
 
-    BTBEntry *victim = btb.findVictim({instPC, tid});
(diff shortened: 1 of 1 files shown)


pranith:
I think the flow is 'lookup()' followed by 'update()' if the lookup returns a wrong prediction. Are we installing a target even if it is a hit/correct?

dhschall:
In the O3 core, update() can be 10s of cycles after the lookup(). The BTB is only updated on misprediction thats true. But that's not the point; you can have a conditional branch that is constantly mispredicted and hits in the BTB. Without checking, every update will create a duplicate entry in the BTB, unless the replacement policy is lucky enough to select the existing entry.

pranith (approved):

giactra (approved):
This is awesome, thanks for spotting this.
Side note, we should really start benchmarking our BTB performance

PR #3058, opened 2026-04-02
arch-riscv: Add RELEASE flag to all CBO instructions

zhongchengyong:
Add Request::RELEASE flag to all four RISC-V Zicbom extension cache block operation instructions (cbo.inval, cbo.clean, cbo.flush, and cbo.zero) to ensure proper store queue ordering and prevent cache race conditions.

Cache maintenance operations must execute after all prior memory accesses complete to ensure correctness:
For cbo.clean and cbo.flush, without the RELEASE flag, when prior store instructions experience cache misses, the cache does not immediately allocate dirty cache lines. **As a result, the WriteClean operation would have no effect since there are no dirty lines to clean**. 
The RELEASE flag ensures all pending store operations complete before the cache maintenance instruction executes. Similarly, the RELEASE flag has been added to cbo.inval and cbo.zero to prevent race conditions between these cache maintenance operations and pending store queue entries.

### DIFF
--- a/src/arch/riscv/isa/decoder.isa
+++ b/src/arch/riscv/isa/decoder.isa
@@ -1373,7 +1373,7 @@ decode QUADRANT default Unknown::unknown() {
                                 // on menvcfg (01 and 11 respectively)
                                 Mem = 0;
                             }
-                        }}, mem_flags = [INVALIDATE, DST_POC]);
+                        }}, mem_flags=[INVALIDATE, DST_POC, RELEASE]);
                             // mem_flags might need a conditional to add CLEAN
                             // in certain cases; unsure of syntax for that
 
@@ -1393,7 +1393,7 @@ decode QUADRANT default Unknown::unknown() {
                             } else {
                                 Mem = 0;
                             }
-                        }}, mem_flags=[CLEAN, DST_POC]);
+                        }}, mem_flags=[CLEAN, DST_POC, RELEASE]);
                         0x2: cbo_flush({{
                             SENVCFG senvcfg = xc->readMiscReg(MISCREG_SENVCFG);
                             auto pm = (PrivilegeMode)xc->readMiscReg(
@@ -1410,7 +1410,7 @@ decode QUADRANT default Unknown::unknown() {
                             } else {
                                 Mem = 0;
                             }
-                        }}, mem_flags=[CLEAN, INVALIDATE, DST_POC]);
+                        }}, mem_flags=[CLEAN, INVALIDATE, DST_POC, RELEASE]);
                         0x4: cbo_zero({{
                             SENVCFG senvcfg = xc->readMiscReg(MISCREG_SENVCFG);
                             auto pm = (PrivilegeMode)xc->readMiscReg(
@@ -1426,7 +1426,7 @@ decode QUADRANT default Unknown::unknown() {
                             } else {
                                 Mem = 0;
                             }
-                        }}, mem_flags=[CACHE_BLOCK_ZERO]);
+                        }}, mem_flags=[CACHE_BLOCK_ZERO, RELEASE]);
                     }
                 }
             }


zhongchengyong:
testcase to reproduce:
```
/* 
 * Use inline assembly to make store and cbo.clean adjacent.
 * First access to this address will MISS in L1 cache.
 */
asm volatile(
    /* Store 0xAB - will MISS and trigger cache fill */
    "sb %1, %0\n\t"
    /* Immediately execute cbo.clean on the same address */
    "cbo.clean %0\n\t"
    : "+m" (*(volatile uint8_t *)addr)
    : "r" (test_val)
    : "memory"
);

```

powerjg:
The code change looks fine to me. But could you provide a permalink to the RISC-V spec where it says that these should be ordered after all prior memory accesses just for record keeping?

Note: The failure was just the test being flaky. I'm re running now.

zhongchengyong:
> The code change looks fine to me. But could you provide a permalink to the RISC-V spec where it says that these should be ordered after all prior memory accesses just for record keeping?
> 
> Note: The failure was just the test being flaky. I'm re running now.

For example, 'cbo.clean' instruction definition in spec: 
Link: https://docs.riscv.org/reference/isa/unpriv/cmo.html#insns-cbo_clean
> A cbo.clean instruction performs a clean operation on the cache block whose effective address is the base address specified in rs1.   

If the following cbo.clean can't writeback the prior store cacheline,  it does not comply with the instruction behavior/functionality defined in the spec.

powerjg (approved):

powerjg (changes requested):
Sorry, I changed my mind on the review. I don't see why this should have release set on it. I don't see anywhere in the spec where there are any guarantees on the ordering of cbo instructions. Here's what gemini told me. Please let me know how it is wrong.

The short answer is **no**. The `cbo.clean` instruction does not inherently imply a release barrier, nor does it provide any implicit ordering guarantees relative to other memory accesses.

In the RISC-V architecture, cache-block management instructions (part of the **Zicbom** extension) are designed to be as "weak" as the rest of the memory model to allow for high-performance implementations.

### Why it doesn’t imply a Release
In the RISC-V Weak Memory Ordering (RVWMO) model, `cbo.clean` is treated similarly to a **store** instruction. This means:

*   **No Implicit Fencing:** It does not act as a barrier for preceding or succeeding instructions.
*   **Pipeline Reordering:** The processor is free to reorder `cbo.clean` relative to other loads and stores unless you explicitly tell it not to.
*   **Visibility:** While it ensures a dirty line is written back to the next level in the memory hierarchy, it doesn't guarantee *when* that happens relative to other memory operations in the pipeline.

---

### How to enforce ordering
If you need `cbo.clean` to behave like a release (ensuring all prior stores are visible to the cache before the "clean" operation occurs), you must use a `FENCE` instruction.

| Desired Behavior | Required Sequence |
| :--- | :--- |
| **Ensure prior writes are cleaned** | `FENCE O, O` followed by `cbo.clean` |
| **Ensure clean completes before next task** | `cbo.clean` followed by `FENCE O, O` |
| **Full Release/Acquire Semantics** | Use `FENCE RW, RW` or appropriate `.rl`/`.aq` bits on atomic instructions surrounding the CBO. |

### Technical Nuance: The "Store" Classification
The RISC-V specification explicitly categorizes `cbo.clean`, `cbo.flush`, and `cbo.inval` as operations that perform a **write** to the cache state. Because they are treated as writes, they follow the standard $P_{o}$ (Program Order) rules for stores. 

If you are using `cbo.clean` to prepare a buffer for a non-coherent DMA engine, you almost certainly need a `FENCE W, W` (or `FENCE O, O`) before the instruction to ensure the data you just wrote to the buffer actually reaches the cache before the clean operation tries to push it to main memory.

> [!TIP]
> If you're looking for instructions with built-in ordering, you're usually looking for the **.aq (acquire)** and **.rl (release)** suffixes, but these are currently only available on Atomic (A extension) instructions, not on Zicbom cache management instructions.

zhongchengyong:
@powerjg Thanks for your reply. I think you’re right — according to the spec, there really is no memory-ordering constraint. In fact, this issue was discovered while debugging our in-house instructions, and our implementation does guarantee the ordering of cbo instructions.   
I’ll go ahead and close this PR.

PR #3061, opened 2026-04-02
stdlib: Use atomic_noncaching switching to atomic CPU with Ruby

### LINKED ISSUE #3030: stdlib: Switching to CPUTypes.ATOMIC with a Ruby cache fails with wrong MemMode
**Describe the bug**
When using `SimpleSwitchableProcessor()` in stdlib where the `switch_core_type` is `CPUTypes.ATOMIC` and the cache hierarchy is Ruby, `m5.switchCpus` changes memory mode to `atomic` instead of `atomic_noncaching`.

**Affects version**
develop (6c2a4507d5965c0feb5951715f78f9a38bd63ca3)

**gem5 Modifications**
Changed `configs/example/gem5_library/x86-ubuntu-run-with-kvm.py`:
```
-    switch_core_type=CPUTypes.TIMING,
+    switch_core_type=CPUTypes.ATOMIC,
```

**To Reproduce**

0. Make modification above
1. Compile gem5 with command `scons build/ALL/gem5.opt` + any -j, etc.
2. Execute the simulation with `build/ALL/gem5.opt configs/example/gem5_library/x86-ubuntu-run-with-kvm.py`

**Terminal Output**

Relevant panic:

```shell
...
src/mem/ruby/system/RubyPort.cc:348: panic: Ruby supports atomic accesses only in noncaching mode
...
```

**Expected behavior**
Either stdlib or m5.simulate should check for ruby and change the system memory mode to atomic_noncaching.

**Host Operating System**
Ubuntu 24.04

**Host ISA**
X86

**Compiler used**
gcc 13.3.0

**Additional information**
I am not sure if the fix can be done in stdlib by setting the CPU simobj memory mode or if the older `src/python/m5/simulate.py` needs to be changed for this.
I think stdlib can maybe set the simobj such that this line returns atomic_noncaching: https://github.com/gem5/gem5/blob/stable/src/python/m5/simulate.py#L457

abmerop:
Atomic CPU with Ruby requires the memory mode to be atomic_noncaching. Currently the m5 python library does not consider this, but stdlib allows for switching to an atomic CPU with a Ruby cache hierarchy.

This commit forces the memory mode to be noncaching if it was atomic and ruby is enabled in the m5 python library. Proposed fix for #3030.

### DIFF
--- a/src/python/gem5/components/processors/switchable_processor.py
+++ b/src/python/gem5/components/processors/switchable_processor.py
@@ -155,7 +155,9 @@ class SwitchableProcessor(AbstractProcessor):
 
         # Switch the CPUs
         m5.switchCpus(
-            self._board, list(zip(current_core_simobj, to_switch_simobj))
+            self._board,
+            list(zip(current_core_simobj, to_switch_simobj)),
+            is_ruby=self._board.get_cache_hierarchy().is_ruby(),
         )
 
         # Ensure the current processor is updated.
--- a/src/python/m5/simulate.py
+++ b/src/python/m5/simulate.py
@@ -427,7 +427,7 @@ def _changeMemoryMode(system, mode):
         print("System already in target mode. Memory mode unchanged.")
 
 
-def switchCpus(system, cpuList, verbose=True):
+def switchCpus(system, cpuList, verbose=True, is_ruby=False):
     """Switch CPUs in a system.
 
     .. note::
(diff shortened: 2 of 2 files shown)


mattsinc:
@poal023 , @v-ramadas , @TomXia : FYI

powerjg (approved):

PR #3067, opened 2026-04-07
arch-riscv: Fix incorrect old_vd_idx for OPIVI vector instructions

amatabsc:
The widening format was hardcoded to look for the old destination register at index 2. This is correct for OPIVV, but for OPIVI the index should be 1 as there is no vs1 register. This caused the mask register to be incorrectly used as the old VD.


saul44203 (approved):
LGTM

erin-le (approved):

PR #3074, opened 2026-04-10
sim: Add current tick to panic output

powerjg:
Register an extra log callback in the Root constructor to print the current simulation tick whenever a panic occurs. This provides useful context for debugging simulation failures without needing to manually add the tick to every panic message.

The callback safely checks if the event queue pointer (`_curTickPtr`) is valid before calling `curTick()` to avoid segmentation faults if a panic occurs before the simulation event queues are fully initialized.

### DIFF
--- a/src/sim/root.cc
+++ b/src/sim/root.cc
@@ -193,6 +193,16 @@ Root::Root(const RootParams &p, int)
     // having a single global stat group for global stats. Merge that
     // group into the root object here.
     mergeStatGroup(&Root::RootStats::instance);
+
+    // Print the tick if we panic.
+    Logger::getPanic().registerExtraLog([]() {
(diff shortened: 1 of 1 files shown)


src/sim/root.cc
```diff
@@ -193,6 +193,19 @@ Root::Root(const RootParams &p, int)
     // having a single global stat group for global stats. Merge that
     // group into the root object here.
     mergeStatGroup(&Root::RootStats::instance);
+
+    static bool registered = false;
```
giactra: Why is this needed? Don't we always have one Root only?

powerjg: Good point! I'll fix that.

giactra (approved):

PR #3080, opened 2026-04-14
mem: Add enable_backdoor param to BaseXBar

rogerchang23424:
If the enable_backdoor is ture, the XBar will send packet to mem_side_ports with backdoor. Otherwise, send packet without backdoor

Change-Id: I48e2ae353a72adaa81fd10f45b0475b2490952e3

### DIFF
--- a/src/mem/XBar.py
+++ b/src/mem/XBar.py
@@ -93,6 +93,11 @@ class BaseXBar(ClockedObject):
     use_default_range = Param.Bool(
         False, "Perform address mapping for the default port"
     )
+    enable_backdoor = Param.Bool(
+        True,
+        "If true, allow receiving and forwarding of backdoor requests. "
+        "If false, the XBar will send packets without a backdoor.",
+    )
 
 
 class NoncoherentXBar(BaseXBar):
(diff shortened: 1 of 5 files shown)


powerjg (commented):
LGTM, but can you explain why one would use `enable_backdoor = False`? Why wouldn't it be up to the downstream controller to decide what to do with the backdoor?

ylldummy:
One use case we've recently encountered is that some 3rd party model (closed source) has incorrect backdoor behavior and it'd be easier to isolate/workaround the problem if we could simply disable it on the XBar before it.

src/mem/XBar.py
```diff
@@ -93,6 +93,7 @@ class BaseXBar(ClockedObject):
     use_default_range = Param.Bool(
         False, "Perform address mapping for the default port"
     )
+    enable_backdoor = Param.Bool(True, "Enable receiving backdoor")
```
powerjg: One tiny request: Can re replicate the comment in `xbar.hh` here? "If true, allow receiving and forwarding of backdoor requests. If false, the XBar will send packets without a backdoor."

powerjg (changes requested):

src/mem/XBar.py
rogerchang23424: Done

powerjg (approved):

PR #3084, opened 2026-04-14
mem-ruby: add transition_no_stall

giactra:
transition_no_stall defines a transition that executes without checking if the necessary resources are available.

Change-Id: Ibddab0cfafc74e7861512f6f0d42d8147adb368e

### DIFF
--- a/src/mem/slicc/symbols/Transition.py
+++ b/src/mem/slicc/symbols/Transition.py
@@ -1,3 +1,15 @@
+# Copyright (c) 2026 ARM Limited
+# All rights reserved.
+#
+# The license below extends only to copyright in the software and shall
+# not be construed as granting a license to any other intellectual
+# property including but not limited to intellectual property relating
+# to a hardware implementation of the functionality of the software
+# licensed hereunder.  You may use the software subject to the license
+# terms below provided that you ensure that this notice is replicated
+# unmodified and in its entirety in all distributions of the software,
+# modified or unmodified, in source code or in binary form.
+#
 # Copyright (c) 1999-2008 Mark D. Hill and David A. Wood
 # Copyright (c) 2009 The Hewlett-Packard Development Company
 # All rights reserved.
@@ -40,6 +52,7 @@ class Transition(Symbol):
         actions,
         request_types,
         location,
+        ignore_resources,
     ):
         ident = f"{state}|{event}"
         super().__init__(table, ident, location)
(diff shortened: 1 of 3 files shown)


src/mem/slicc/symbols/Transition.py
```diff
@@ -64,12 +77,15 @@ def __init__(
...
         self.resources = {}
 
-        for action in self.actions:
-            for var, value in action.resources.items():
-                num = int(value)
-                if var in self.resources:
-                    num += int(value)
-                self.resources[var] = str(num)
+        if ignore_resources:
+            print("%s: transition ignores resource allocation" % location)
```
powerjg, line 81: Do you mean to leave this print statement?

powerjg (commented):
This looks good to me... can you explain when/why you would use transistion_no_stall?

All of these keywords are poorly documented (mostly just in the learning_gem5 chapter). We should at least write down in the commit message what the use case for the keyword is.

erin-le:
@giactra: Hello, could you respond to Jason's comments when you get a chance? Thank you!

erin-le:
@giactra: Hello, just wanted to send a reminder about this PR. Could you respond to Jason's comments and take a look at the conflicts when you have time? Thank you!

src/mem/slicc/symbols/Transition.py
giactra: Hey Jason, I think @tiagormk added it for debug purposes. Happy to remove it if you wish

giactra:
> This looks good to me... can you explain when/why you would use transistion_no_stall?
> 

It's basically a way to execute a transition (and connected actions) even if there are no available resources. This is something that is useful if for example I have a decision flow like the following:

- Try to do something next cycle
- If resource available do A
- If resource not-available do B

> All of these keywords are poorly documented (mostly just in the learning_gem5 chapter). We should at least write down in the commit message what the use case for the keyword is.

Ack, if this explanation satisfies you, I will update the PR/commit msg accordingly

src/mem/slicc/symbols/Transition.py
powerjg: If we compile all protocols, does this get triggered? If not, it's OK to keep (i.e., it's only printed if something is wrong). If there are protocols where this spams during build, we shouldn't include it.

powerjg:
That makes sense to me. It is something that could be abused to make bad models if you aren't careful or don't understand what you're doing. That's most of Ruby, though :)

erin-le:
@giactra: Hello, just wanted to let you know that some conflicts have arisen on this PR. Also, in case you missed it, Jason responded to your comment on this PR. Could you take a look at both of these when you get a chance? Thank you!

BobbyRBruce (approved):
This appears fine to me. Merging.

BobbyRBruce:
> This appears fine to me. Merging.

Sorry, scratch that, there's merge conflicts here. @giactra : Could you resolve this conflicts (I don't have permissions to push to your branch)

BobbyRBruce:
> > This appears fine to me. Merging.
> 
> Sorry, scratch that, there's merge conflicts here. @giactra : Could you resolve this conflicts (I don't have permissions to push to your branch)

I've created #3428 which is just this PR with the merge conflict resolved. We can merge if we get to the point we wish to create the staging branch but @giactra has yet to fix this PR. #3428 should be closed without merge if this (#3084) is merged, and vice-versa.

PR #3087, opened 2026-04-14
misc: clang-format GPU files

abmerop:
The intent of this is to solve the following:
 1) Running clang-format per-commit clutters PRs
 2) When doing development and trying to break up changes into multiple
 logical commits, the reformat can be a PITA when using `git add -p` for
 example to re-add the the reformated files. Sometimes the reformant
 happens multiple times.
 3) These codes are largely touched by few people who would like the
 reformant done all at once rather than reformatting more widely
 developed file (e.g., the whole repo).


powerjg (approved):
Wow. Lots of changes!

PR #3088, opened 2026-04-15
mem: addd SimpleMemory latency based on norm dist 

giactra:
If latency_stdev is set, latency_var is ignored, and the latency for
each accesss is computed using a normal distribution with avg=latancy
and stdev=latency_stdev.

### DIFF
--- a/src/mem/SimpleMemory.py
+++ b/src/mem/SimpleMemory.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2012-2013, 2021 Arm Limited
+# Copyright (c) 2012-2013, 2021, 2026 Arm Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
@@ -40,6 +40,19 @@ from m5.objects.AbstractMemory import *
 from m5.params import *
 
 
+class LatencyDistribution(ScopedEnum):
+    vals = [
+        # Uniform distribution between [0, latency_dist_param] which
+        # gets added to the latency Param
+        "Uniform",
+        # Normal distribution norm(latency, latency_dist_param)
+        # latency_dist_param is the standard deviation.
+        # Distribution gets clipped after 3 * stdev from the mean
+        "Normal",
+        "None",
+    ]
+
+
 class SimpleMemory(AbstractMemory):
     type = "SimpleMemory"
     cxx_header = "mem/simple_mem.hh"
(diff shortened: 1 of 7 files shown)


src/mem/SimpleMemory.py
```diff
@@ -47,7 +47,9 @@ class SimpleMemory(AbstractMemory):
 
     port = ResponsePort("This port sends responses and receives requests")
     latency = Param.Latency("30ns", "Request to response latency")
+    # use either one or the other, stdev used a normal distribution
```
powerjg, line 50: I would like the names/documentation to be more clear. It is quite confusing for users of this feature to have both variance and stddev and one used for (I assume) a uniform distribution and the other used for normal distribution. My suggestion is:
- Deprecate `latency_var`
- Add a new parameter which is an enum for "random distribution" which is either "normal" or "uniform" (you may want to add "none" to turn off randomization).
- Add a parameter "random_variance" or something like that which is documented such that "controls the parameter for the random function. For uniform random is used as the width of the distribution (latency +/- variance) and for normal is used as the standard deviation. Normal distribution cannot generate latencies more than 3 stdev away from the mean."
- Throw a fatal if `random_variance` (or whatever you choose) is nonzero when the randomness is off.

src/mem/simple_mem.cc
```diff
@@ -236,8 +243,15 @@ SimpleMemory::dequeue()
 Tick
 SimpleMemory::getLatency() const
 {
-    return latency +
-        (latency_var ? rng->random<Tick>(0, latency_var) : 0);
+    if (latency_stdev) {
+        Tick lat = std::min(rng->random_norm(latency, latency_stdev),
+                            latency_stdev_max);
```
powerjg, line 248: This code can produce latencies that are more than 3 stddev away from the mean in the negative direction. This is an unexpected (but maybe simply undocumented) behavior.

powerjg (changes requested):
It's a "small" comment, but I can see this parameter being used frequently so I think it's important to get the interface and documentation clear.

src/mem/SimpleMemory.py
giactra: > Deprecate latency_var
Add a new parameter which is an enum for "random distribution" which is either "normal" or "uniform" (you may want to add "none" to turn off randomization).

Yes agreed

> Add a parameter "random_variance" or something like that which is documented such that "controls the parameter for the random function. For uniform random is used as the width of the distribution (latency +/- variance) and for normal is used as the standard deviation. Normal distribution cannot generate latencies more than 3 stdev away from the mean."

Pardon my ignorance; where does the 3stdev constraint comes from?

powerjg: > Pardon my ignorance; where does the 3stdev constraint comes from?

Line 60 of `src/mem/simple_mem.cc`: ` latency_stdev_max(p.latency + (p.latency_stdev * 3)),`

I'm suggesting documenting the behavior. I don't have an opinion on what the behavior should be.

giactra: I missed that. Looks reasonable

giactra: Done, this is now documented

src/mem/simple_mem.cc
giactra: Yeah this asymmetry is a bit weird. I added the clipping on the negative tail as well.

powerjg (approved):

PR #3095, opened 2026-04-17
misc: add GitHub agentic workflow for labeling issues

erin-le:
This PR adds a GitHub agentic workflow for automatically labeling issues when they are opened. Two workflows are needed because each workflow is limited to only 50 labels, and gem5 has more than that. The files ending with `.lock` are generated using the command `gh aw compile`, and are necessary for the workflows to run.

I tested this by opening issues in my own fork of gem5 and seeing which labels were applied.


.github/workflows/issue-labeler-second-half.lock.yml
BobbyRBruce, line 1: I assume these `.lock.yml` files were accidentally added?

lhartung:
That is interesting work! I noticed that most of the PR labels seem to be path-based rules, e.g. if files in `src/arch/x86` were modified, then apply label `arch-x86`. I wonder if you had considered the [labeler action](https://github.com/actions/labeler) and found it to be insufficient in some way. It seems like a set of path-based rules might be easier to maintain long-term.

erin-le:
@lhartung: Hello, thanks for your comment! I mostly made this PR as a way for the gem5 dev team to explore ways of increasing automation using AI, and wasn't aware of the `labeler` action. 

I briefly looked into the `labeler` action, and it looks like one of the restrictions is that it only works on PRs. The approach in this PR (using GitHub Agentic Workflows) can label issues as well, so it's somewhat more comprehensive. In terms of long term maintenance, it looks like the `labeler` action is more stable than GitHub Agentic Workflows, so it's a bit of a tradeoff. I'd like to get @BobbyRBruce's opinion on what would be better to use as well.

lhartung:
@erin-le: Thanks for your thoughts! Yes, you are absolutely right that automatically labeling issues is a distinct and interesting capability enabled by LLMs.

I have also been experimenting with some of these ideas, both conventional CI approaches and doing some work with Gemini. I found that for labeling PRs based on files changed, the `labeler` action is quite simple to use. [Here](https://github.com/lhartung/gem5-ci-work/blob/develop/.github/labeler.yml) is an example. It is by no means comprehensive, and I think I missed some of the intricacies like do not apply `cpu` tag if `cpu-o3` is also applied. I think it would be pretty easy to finish if there is interest, though.

erin-le:
@lhartung: I brought this up with @BobbyRBruce and he said that it would be worthwhile for me to look into using the `labeler` action for PRs while keeping the agentic workflows for issues. Would you be okay with it if I built off of your commit?

lhartung:
@erin-le: Yes, of course! Let me know if there is any way I can help. I wish I had had time to finish writing all of the rules.

BobbyRBruce (approved):

PR #3101, opened 2026-04-21
scons: Raise minimum protoc version to 3.0.0

MrJaxK:
The proto files in src/proto/ use the 'reserved' keyword (e.g., inst.proto), which was introduced in protobuf 3.0. The previous minimum of 2.1.0 allowed protoc 2.x to pass the configure check but then fail during compilation with a confusing parse error:

  inst.proto:98:14: Missing numeric value for enum constant.

### DIFF
--- a/src/proto/SConsopts
+++ b/src/proto/SConsopts
@@ -39,8 +39,8 @@ with gem5_scons.Configure(main) as conf:
     except Exception as e:
         warning('While checking protoc version:', str(e))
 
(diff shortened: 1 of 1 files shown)


BobbyRBruce:
Wow, Protobuf's latest stable release is v34! Version 3 was released 10 years ago.

So, yeah, as conservative as I can be about such things, this is fine. I can take the heat from the legacy Protobuf v2 users when they complain we broke their build.

BobbyRBruce (approved):

PR #2619, opened 2025-09-30
arch-riscv: new vector instructions

Joao-Pedro-Cabral:
As discussed here [#2566](https://github.com/gem5/gem5/issues/2566). I noted that there are some missing vector instructions, in this PR I implemented the following instructions:
- vandn
- vwsll
- vror
- vrol
- vcompress
- vclmul
- vclmulh
I tested the instructions locally and looks fine.
Can you do a special check  in my implementation of vcompress? Since I relies in a internal buffer and I spends 2*micro_vl instructions for doing it.

PR #3104, opened 2026-04-22
arch-riscv: Reimplement vcompress to correctly handle overlapping source and destination registers

amatabsc:
The previous vcompress.vm implementation relied on static buffers (vcompress_buf, vcompress_cnt) shared across all micro-ops and across independent dynamic instances of the instruction. This was unsafe: squash/replay of micro-ops, or multiple in-flight vcompress.vm instructions, could corrupt the shared static state and produce incorrect results, especially when source and destination registers overlapped.

Now, vcompress.vm has been reimplemented so each micro-op is self-contained and stateless:
- Each micro-op now sources the full vs2 register group it needs directly, instead of relying on a single shared source register.
- Each micro-op independently recomputes the global count of active elements from the start of vs2, writing only the elements that fall in its own destination range.
- This removes all shared mutable state between micro-ops/instances, so the instruction is correct under speculative execution, squash, and replay.
- vector_arith.isa format and templates updated accordingly.


saul44203 (changes requested):
I think we have a deeper issue here with the `vcompress.vm` implementation (#2619) as a whole.

It makes use of a shared buffer that requires in-order execution of µops (both in the same instruction and others) to maintain correctness, which is not always the case, and it also doubles the µop count to divide the gather and the packing functionalities with no ordering mechanism between them, which again can be problematic with OoO execution.

We should re-implement it in my opinion. One way would be to have ⌈vl/µvlmax⌉ µops doing the gather independently, each into a different internal reg, and a final µop, that has a data dependency with all of them, which does the final packing into vd. To get the element count of each gather µop when packing you could just use the mask in vs1.

UPDATE:
I think there's no need to use internal registers, neither a final packing µop. We could implement a version that still does one µop per vd (so ⌈vl/µvlmax⌉), and each of them has the (same) whole vs2 vector register group as source (so LMUL vector register sources).  This way each µop can do the gather and the packing directly to vd, with respect to the mask in vs1. On those µops where vs1 active elements have been consumed (i.e. there's no gather) but µvl still isn't, we treat vd elements as part of the tail per spec.

amatabsc:
> I think we have a deeper issue here with the `vcompress.vm` implementation (#2619) as a whole.
> 
> It makes use of a shared buffer that requires in-order execution of µops (both in the same instruction and others) to maintain correctness, which is not always the case, and it also doubles the µop count to divide the gather and the packing functionalities with no ordering mechanism between them, which again can be problematic with OoO execution.
> 
> We should re-implement it in my opinion. One way would be to have ⌈vl/µvlmax⌉ µops doing the gather independently, each into a different internal reg, and a final µop, that has a data dependency with all of them, which does the final packing into vd. To get the element count of each gather µop when packing you could just use the mask in vs1.
> 
> UPDATE: I think there's no need to use internal registers, neither a final packing µop. We could implement a version that still does one µop per vd (so ⌈vl/µvlmax⌉), and each of them has the (same) whole vs2 vector register group as source (so LMUL vector register sources). This way each µop can do the gather and the packing directly to vd, with respect to the mask in vs1. On those µops where vs1 active elements have been consumed (i.e. there's no gather) but µvl still isn't, we treat vd elements as part of the tail per spec.

I've updated it. Can you give it a look?

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -298,51 +297,32 @@ def format VectorCompressFormat(code, category, *flags) {{
...
-    set_src_reg_idx += setSrcWrapper(src1_reg_id)
-    set_src_reg_idx += setSrcWrapper(src2_reg_id)
-
-    dest_set_src_reg_idx = setSrcWrapper(dest_reg_id)
-
-    set_src_reg_idx += dest_set_src_reg_idx
+    set_src_reg_idx += setSrcWrapper("vecRegClass[_machInst.vs1]")
+    for g in range(8):
+        set_src_reg_idx += setSrcWrapper(
+            f"vecRegClass[(_machInst.vs2 + {g}) % NumVecRegs]")
```
saul44203: Let's use LMUL here and not hardcode it to 8.

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -298,51 +297,32 @@ def format VectorCompressFormat(code, category, *flags) {{
...
+    old_vd_src_idx = 1 + 8
+    set_src_reg_idx += setSrcWrapper(
+        "vecRegClass[_machInst.vd + _microIdx]")
     set_src_reg_idx += setSrcVm()
```
saul44203: Actually we don't need this I think, since the masked encoding is reserved.

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -642,7 +622,7 @@ def format VectorIntMaskFormat(code, category, *flags) {{
 def format VectorGatherFormat(code, category, *flags) {{
     inst_name, inst_suffix = name.split("_", maxsplit=1)
     if inst_name == "vrgatherei16":
-        idx_type = "uint16_t"
+        idx_type = "uint16_t"s
```
saul44203: Typo?

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2335,9 +2335,25 @@ Fault
...
+def template VectorCompressMicroDeclare {{
 
-def template VectorCompressMacroConstructor {{
+template<typename ElemType>
+class %(class_name)s : public %(base_class)s
+{
+private:
+    RegId srcRegIdxArr[12];
+    RegId destRegIdxArr[1];
+    bool vm;
```
saul44203: Same here, no need.

src/arch/riscv/isa/decoder.isa
```diff
@@ -4573,25 +4573,43 @@ decode QUADRANT default Unknown::unknown() {
...
+                        for (uint64_t ei = 0; ei < total_vl; ei++) {
+                            if (!elem_mask(Vs1_vu, ei))
+                                continue;
+                            if (dst_idx >= dst_start && dst_idx < dst_end) {
+                                Vd_vu[dst_idx - dst_start] = vs2_elem(ei);
+                                filled++;
                             }
+                            dst_idx++;
+                            if (dst_idx >= dst_end)
+                                break;
```
saul44203: Let's avoid breaks, we can just linearly scan.

src/arch/riscv/isa/decoder.isa
```diff
@@ -4573,25 +4573,43 @@ decode QUADRANT default Unknown::unknown() {
...
+                        };
+
+                        uint64_t dst_idx = 0;
+                        uint32_t filled  = 0;
+
+                        for (uint64_t ei = 0; ei < total_vl; ei++) {
+                            if (!elem_mask(Vs1_vu, ei))
+                                continue;
+                            if (dst_idx >= dst_start && dst_idx < dst_end) {
+                                Vd_vu[dst_idx - dst_start] = vs2_elem(ei);
```
saul44203: Is this indexing right? You are offsetting in vd like it's is more than one destination register. This will overflow in µops with `microIdx`>0.

We have 1 µop per vd, so we only need to fill it linearly with µvl elements as we scan the vs2 group, no need for offsetting vd. Whe should only need to offset in the vs2 register group.

src/arch/riscv/isa/decoder.isa
```diff
@@ -4573,25 +4573,43 @@ decode QUADRANT default Unknown::unknown() {
...
+                                filled++;
                             }
+                            dst_idx++;
+                            if (dst_idx >= dst_end)
+                                break;
+                        }
+
+                        if (!bits(vtype, 6)) {
```
saul44203: Use something more descriptive like `machInst.vtype8.vta` instead.

src/arch/riscv/isa/decoder.isa
```diff
@@ -4573,25 +4573,43 @@ decode QUADRANT default Unknown::unknown() {
...
+                        const uint64_t dst_start = (uint64_t)this->microIdx * elems_per_reg;
+                        const uint64_t dst_end   = dst_start + elems_per_reg;
+
+                        auto vs2_elem = [&](uint64_t ei) -> ElemType {
+                            const uint32_t src_reg     = (uint32_t)(ei / elems_per_reg);
+                            const uint32_t elem_in_reg = (uint32_t)(ei % elems_per_reg);
+                            gem5::RiscvISA::VecRegContainer tmp;
+                            xc->getRegOperand(this, 1 + src_reg, &tmp);
+                            return tmp.template as<ElemType>()[elem_in_reg];
+                        };
```
saul44203: I think this is really inefficient. Let's remove the lambda and use a nested loop to scan each source register instead.

saul44203 (changes requested):
Major changes to be done.

Citing the implementation again:
> We could implement a version that still does one µop per vd (so ⌈vl/µvlmax⌉), and each of them has the (same) whole vs2 vector register group as source (so LMUL vector register sources). This way each µop can do the gather and the packing directly to vd, with respect to the mask in vs1. On those µops where vs1 active elements have been consumed (i.e. there's no gather) but µvl still isn't, we treat vd elements as part of the tail per spec.

Ask any questions if unclear.

src/arch/riscv/isa/formats/vector_arith.isa
amatabsc: As far as i know, vtype_vlmul and vtype_regs_per_group are C++ functions and are not available at format generation time in the ISA Python layer. The hardcoded 8 is the maximum possible (LMUL8), which is the pattern used throughout the codebase. LMUL is only evaluable in the generated C++ constructor. Tell my what you mean in case that i didn't understood you correctly.

src/arch/riscv/isa/templates/vector_arith.isa
amatabsc: Do you mean we should just reduce the array size from 12 to 10 (vs1 + 8 vs2 + vm, no old_vd), or drop the dedicated VectorCompressMicroDeclare entirely and reuse an existing template?

amatabsc:
> Major changes to be done.
> 
> Citing the implementation again:
> 
> > We could implement a version that still does one µop per vd (so ⌈vl/µvlmax⌉), and each of them has the (same) whole vs2 vector register group as source (so LMUL vector register sources). This way each µop can do the gather and the packing directly to vd, with respect to the mask in vs1. On those µops where vs1 active elements have been consumed (i.e. there's no gather) but µvl still isn't, we treat vd elements as part of the tail per spec.
> 
> Ask any questions if unclear.

The current implementation already does one µop per vd register. Each µop has the whole vs2 register group as source (8 slots hardcoded to max LMUL). The key issue is tracking how many active vs1 elements have been consumed by previous µops, so each µop knows where to start filling its vd. Should we use a shared counter pointer (similar to how viota uses `int* cnt`) passed from the macro constructor to each micro constructor. Each µop would then scan the full vs2 group, skip elements until it reaches the counter offset, fill its vd linearly, and update the counter. Elements beyond the active vs1 count within µvl would be treated as tail. Does this approach make sense for you?

UPDATE: As i've said maybe we can implement the shared counter approach similar to how viota handles cross-µop state. The macro constructor would hold a uint32_t dst_cnt = 0 and pass a pointer to each micro-op. The first micro-op resets it via isFirstMicroop(). Each micro-op then scans the full vs2 register group (all LMUL registers), counting active vs1 elements to determine which ones fall in its vd range [dst_cnt_start, dst_cnt_start + elems_per_reg), fills its vd linearly, and updates the counter. Elements beyond the active vs1 count within µvl are treated as tail per spec.
One concern: each µop scans the full vs2 group (O(vl) work per µop vs O(µvlmax)), which is less efficient but correct. Does this approach look good to you, or do you see a better way to track the cross-µop active element count?

src/arch/riscv/isa/formats/vector_arith.isa
saul44203: Check #3223's `setVecRegGroupSrcWrapper`. You can reuse it. It's not technically LMUL but ⌈vl/µvlmax⌉ registers instead to avoid sourcing unused registers, but the idea is the same.

src/arch/riscv/isa/templates/vector_arith.isa
saul44203: This is in reference of the comment about masking above:

> Actually we don't need this I think, since the masked encoding is reserved.

The instruction does not support it, so remove the vm bool and set the size to 10  (vs1, vs2 (up to 8) and old vd).

saul44203:
> > Major changes to be done.
> > Citing the implementation again:
> > > We could implement a version that still does one µop per vd (so ⌈vl/µvlmax⌉), and each of them has the (same) whole vs2 vector register group as source (so LMUL vector register sources). This way each µop can do the gather and the packing directly to vd, with respect to the mask in vs1. On those µops where vs1 active elements have been consumed (i.e. there's no gather) but µvl still isn't, we treat vd elements as part of the tail per spec.
> > 
> > 
> > Ask any questions if unclear.
> 
> The current implementation already does one µop per vd register. Each µop has the whole vs2 register group as source (8 slots hardcoded to max LMUL). The key issue is tracking how many active vs1 elements have been consumed by previous µops, so each µop knows where to start filling its vd. Should we use a shared counter pointer (similar to how viota uses `int* cnt`) passed from the macro constructor to each micro constructor. Each µop would then scan the full vs2 group, skip elements until it reaches the counter offset, fill its vd linearly, and update the counter. Elements beyond the active vs1 count within µvl would be treated as tail. Does this approach make sense for you?
> 
> UPDATE: As i've said maybe we can implement the shared counter approach similar to how viota handles cross-µop state. The macro constructor would hold a uint32_t dst_cnt = 0 and pass a pointer to each micro-op. The first micro-op resets it via isFirstMicroop(). Each micro-op then scans the full vs2 register group (all LMUL registers), counting active vs1 elements to determine which ones fall in its vd range [dst_cnt_start, dst_cnt_start + elems_per_reg), fills its vd linearly, and updates the counter. Elements beyond the active vs1 count within µvl are treated as tail per spec. One concern: each µop scans the full vs2 group (O(vl) work per µop vs O(µvlmax)), which is less efficient but correct. Does this approach look good to you, or do you see a better way to track the cross-µop active element count?

Yes, the current macro constructor instantiates one µop per vd but your code is accessing it with an offset like vd points to the whole destination group.

Regarding the shared counter I think you don't need it. It might be more complex but each µop can calculate the number of elements prior µops have consumed in vs1. 

From the spec:

> The vector mask register specified by vs1 indicates which of the first vl elements of vector register group
> vs2 should be extracted and packed into contiguous elements at the beginning of vector register vd. The
> remaining elements of vd are treated as tail elements according to the current tail policy (Section 31.3.4.3)

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -298,54 +297,37 @@ def format VectorCompressFormat(code, category, *flags) {{
...
-    # code
-
+    set_src_reg_idx += setSrcWrapper("vecRegClass[_machInst.vs1]")
+    set_src_reg_idx += """
+        {
+            const uint32_t _num_vs2_regs = vtype_regs_per_group(vtype);
+            for (uint32_t _g = 0; _g < _num_vs2_regs; _g++)
+                setSrcRegIdx(_numSrcRegs++, vecRegClass[(_machInst.vs2 + _g) % NumVecRegs]);
+        }
+    """
```
saul44203, line 342: Do ⌈vl/µvlmax⌉ instead of LMUL, this should be enough since we have a bound on the first vl elements. We avoid sourcing unnecessary registers this way.

Modulo with `NumVecRegs` shouldn't be needed.

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -2380,6 +2382,27 @@ template<typename ElemType>
...
+                                         uint32_t _elen, uint32_t _vlen)
+    : %(base_class)s("%(mnemonic)s", _machInst, %(op_class)s, _microVl,
+                     _microIdx, _elen, _vlen)
+{
+    %(set_reg_idx_arr)s;
+    _numSrcRegs = 0;
+    _numDestRegs = 0;
+    %(set_dest_reg_idx)s;
+    %(set_src_reg_idx)s;
+    %(fast_move_code)s
```
saul44203: This doesn't exist.

src/arch/riscv/isa/decoder.isa
```diff
@@ -4573,26 +4573,47 @@ decode QUADRANT default Unknown::unknown() {
...
-                                Vd_vu[i] = vcompress_buf[ei];
+                        const uint32_t elems_per_reg = vlenb / sizeof(ElemType);
+                        const uint64_t total_vl = (uint64_t)machInst.vl;
+
+                        const uint64_t dst_start = (uint64_t)this->microIdx * elems_per_reg;
+                        const uint64_t dst_end = dst_start + elems_per_reg;
+
+                        uint64_t global_active_count = 0;
+                        uint32_t filled = 0;
+                        const uint32_t num_vs2_regs = vtype_regs_per_group(vtype);
```
saul44203: Related to the comment above for source registers. Will need to be changed.

saul44203 (changes requested):
Some minor details remaining but implementation seems correct. 

LGTM assuming this is tested.

src/arch/riscv/isa/formats/vector_arith.isa
saul44203: This is still the same, you only changed the variable names.

src/arch/riscv/isa/decoder.isa
```diff
@@ -4575,22 +4575,33 @@ decode QUADRANT default Unknown::unknown() {
                     0x17: vcompress_vm({{
                         const uint32_t elems_per_reg = vlenb / sizeof(ElemType);
                         const uint64_t total_vl = (uint64_t)machInst.vl;
+<<<<<<< HEAD
 
+=======
+                       
+>>>>>>> ea7b61a49c (Fix vtype reference in vcompress instruction)
```
saul44203: Remove artifacts from merge conflict

src/arch/riscv/isa/decoder.isa
```diff
@@ -4575,22 +4575,33 @@ decode QUADRANT default Unknown::unknown() {
...
 
                             for (uint32_t elem_in_reg = 0; elem_in_reg < elems_per_reg; elem_in_reg++) {
                                 uint64_t ei = (uint64_t)src_reg * elems_per_reg + elem_in_reg;
 
+=======
+                           
+                            for (uint32_t elem_in_reg = 0; elem_in_reg < elems_per_reg; elem_in_reg++) {
+                                uint64_t ei = (uint64_t)src_reg * elems_per_reg + elem_in_reg;
+                               
+>>>>>>> ea7b61a49c (Fix vtype reference in vcompress instruction)
```
saul44203: Same here

src/arch/riscv/isa/decoder.isa
```diff
@@ -4575,22 +4575,33 @@ decode QUADRANT default Unknown::unknown() {
...
+=======
+                       
+>>>>>>> ea7b61a49c (Fix vtype reference in vcompress instruction)
                         const uint64_t dst_start = (uint64_t)this->microIdx * elems_per_reg;
                         const uint64_t dst_end = dst_start + elems_per_reg;
 
                         uint64_t global_active_count = 0;
                         uint32_t filled = 0;
-                        const uint32_t num_vs2_regs = vtype_regs_per_group(vtype);
+                        const uint32_t num_vs2_regs = vtype_regs_per_group(machInst.vtype8);
```
saul44203: This is still the same

saul44203 (changes requested):

src/arch/riscv/isa/formats/vector_arith.isa
amatabsc: I don't know why my vscode is doing weird stuff, probably i have a problem with my branches.

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -298,54 +297,41 @@ def format VectorCompressFormat(code, category, *flags) {{
...
+    set_src_reg_idx += setSrcWrapper("vecRegClass[_machInst.vs1]")
+    set_src_reg_idx += """
+        {
+            const uint32_t _elems_per_reg = (vlen / 8) / sizeof(ElemType);
+            const uint32_t _num_vs2_regs = (_machInst.vl + _elems_per_reg - 1) / _elems_per_reg;
+            for (uint32_t _g = 0; _g < _num_vs2_regs; _g++)
+                setSrcRegIdx(_numSrcRegs++, vecRegClass[_machInst.vs2 + _g]);
+        }
+    """
+    set_src_reg_idx += setSrcWrapper("vecRegClass[_machInst.vd + _microIdx]")
```
saul44203: Use tailMaskCondSetSrcWrapper here. This way we prevent old vd from being a dependency when using a tail agnostic policy. Can also use `dest_reg_id` var above.

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -298,54 +297,41 @@ def format VectorCompressFormat(code, category, *flags) {{
...
     vm_decl_rd = vmDeclAndReadData()
```
saul44203: We don't need this. Masking is reserved for this instruction.

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -298,54 +297,41 @@ def format VectorCompressFormat(code, category, *flags) {{
...
+    microiop = InstObjParams(
+        name + "_micro",
         Name + "Micro",
         microop_class_name,
         {'code': code,
          'set_dest_reg_idx': set_dest_reg_idx,
          'set_src_reg_idx': set_src_reg_idx,
-         'set_vlenb' : set_vlenb,
+         'set_vlenb': set_vlenb,
          'vm_decl_rd': vm_decl_rd,
```
saul44203: Don't need.

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -298,54 +297,41 @@ def format VectorCompressFormat(code, category, *flags) {{
...
-
-    # code
-
+    set_src_reg_idx += setSrcWrapper("vecRegClass[_machInst.vs1]")
+    set_src_reg_idx += """
+        {
+            const uint32_t _elems_per_reg = (vlen / 8) / sizeof(ElemType);
+            const uint32_t _num_vs2_regs = (_machInst.vl + _elems_per_reg - 1) / _elems_per_reg;
+            for (uint32_t _g = 0; _g < _num_vs2_regs; _g++)
+                setSrcRegIdx(_numSrcRegs++, vecRegClass[_machInst.vs2 + _g]);
```
saul44203: Remove starting underscore from local vars.

src/arch/riscv/isa/decoder.isa
```diff
@@ -4573,26 +4573,45 @@ decode QUADRANT default Unknown::unknown() {
...
+                                    global_active_count++;
+                                }
                             }
                         }
+
+                        if (!machInst.vtype8.vta) {
+                            gem5::RiscvISA::VecRegContainer old_vd_tmp;
+                            xc->getRegOperand(this, 1 + num_vs2_regs, &old_vd_tmp);
+                            auto old_elems = old_vd_tmp.template as<ElemType>();
+                            for (uint32_t i = filled; i < elems_per_reg; i++)
```
saul44203: Curly brace here.

src/arch/riscv/isa/decoder.isa
```diff
@@ -4573,26 +4573,45 @@ decode QUADRANT default Unknown::unknown() {
...
+                                Vd_vu[i] = old_elems[i];
+                        } else {
+                            for (uint32_t i = filled; i < elems_per_reg; i++)
```
saul44203: Here too.

saul44203 (changes requested):
Some minor things I initially missed plus some code style changes I saw (should pass pre-commit hooks too).

src/arch/riscv/isa/formats/vector_arith.isa
```diff
@@ -321,8 +320,8 @@ def format VectorCompressFormat(code, category, *flags) {{
          'set_dest_reg_idx': set_dest_reg_idx,
          'set_src_reg_idx': set_src_reg_idx,
          'set_vlenb': set_vlenb,
-         'vm_decl_rd': vm_decl_rd,
-         'copy_old_vd': copyOldVd(old_vd_idx),
+         'vm_decl_rd': '',
+         'copy_old_vd': '',
```
saul44203: We want to keep this.

src/arch/riscv/isa/decoder.isa
```diff
@@ -4606,11 +4606,13 @@ decode QUADRANT default Unknown::unknown() {
                             gem5::RiscvISA::VecRegContainer old_vd_tmp;
                             xc->getRegOperand(this, 1 + num_vs2_regs, &old_vd_tmp);
                             auto old_elems = old_vd_tmp.template as<ElemType>();
-                            for (uint32_t i = filled; i < elems_per_reg; i++)
+                            for (uint32_t i = filled; i < elems_per_reg; i++){
```
saul44203: Space between parenthesis and curly brace.

src/arch/riscv/isa/decoder.isa
```diff
@@ -4606,11 +4606,13 @@ decode QUADRANT default Unknown::unknown() {
...
                                 Vd_vu[i] = old_elems[i];
+                            }
                         } else {
-                            for (uint32_t i = filled; i < elems_per_reg; i++)
+                            for (uint32_t i = filled; i < elems_per_reg; i++){
```
saul44203: Same.

saul44203 (changes requested):

src/arch/riscv/isa/decoder.isa
```diff
@@ -4602,15 +4602,8 @@ decode QUADRANT default Unknown::unknown() {
...
-                            for (uint32_t i = filled; i < elems_per_reg; i++){
-                                Vd_vu[i] = old_elems[i];
-                            }
-                        } else {
-                            for (uint32_t i = filled; i < elems_per_reg; i++){
+                        if (machInst.vtype8.vta) {
+                            for (uint32_t i = filled; i < elems_per_reg; i++) {
                                 Vd_vu[i] = static_cast<ElemType>(-1);
                             }
                         }
```
saul44203: We don't need this loop either now, right?

saul44203 (changes requested):

src/arch/riscv/isa/decoder.isa
amatabsc: done

saul44203 (approved):

erin-le:
@amatabsc: Hello, it looks like a conflict has arisen on this PR. Could you take a look? Also, once this PR is ready to be merged, would you be okay with a squash merge? This will compress all of the commits on this PR down to one, and will use the PR description as the commit message. If you're fine with it, please double check that the PR description is up to date; otherwise, please modify your commits so each of the commit titles starts with `arch-riscv: `. Thank you!

amatabsc:
> @amatabsc: Hello, it looks like a conflict has arisen on this PR. Could you take a look? Also, once this PR is ready to be merged, would you be okay with a squash merge? This will compress all of the commits on this PR down to one, and will use the PR description as the commit message. If you're fine with it, please double check that the PR description is up to date; otherwise, please modify your commits so each of the commit titles starts with `arch-riscv: `. Thank you!

it is already solved!

erin-le (dismissed):
@amatabsc: Could you double check the PR description and update it if necessary? It looks like the last time it was updated was in April, but there seem to have been significant changes since then.

The PR description will become the commit message once this PR is merged.

amatabsc:
> @amatabsc: Could you double check the PR description and update it if necessary? It looks like the last time it was updated was in April, but there seem to have been significant changes since then.
> 
> The PR description will become the commit message once this PR is merged.

done

erin-le (approved):

PR #3105, opened 2026-04-22
arch-riscv: fix incorrect Rs1 write position in v/vfslide1down.vx when vl < VMAX

amatabsc:
The implementation of vslide1down.vx/vfslide1down.vf incorrectly relied on lastUop to write the final active element (vl - 1). This assumption only holds when vl == VLMAX, where the last element naturally falls within the final micro-op. When vl < VLMAX, the element vl - 1 may belong to an earlier micro-op, causing the write to be skipped entirely.
By replacing the lastUop check with a condition that detects whether vl - 1 lies within the current micro-op window, the write is performed in the correct micro-op regardless of vl. This ensures the final element is always written to the proper position in Vd_vu.
This fix corrects the behavior for all vl values and ensures consistency across both integer and floating-point variants of the instruction.

### DIFF
--- a/src/arch/riscv/isa/decoder.isa
+++ b/src/arch/riscv/isa/decoder.isa
@@ -5552,11 +5552,13 @@ decode QUADRANT default Unknown::unknown() {
                             }
                         }
                     }
-                    if (lastUop) {
-                        if (this->vm || elem_mask(v0,
-                            vdWindowLeft + microVl - 1)) {
-                            auto fd = ftype_freg<et>(freg(Fs1_bits));
-                            Vd_vu[microVl - 1] = fd.v;
+                    if (machInst.vl > 0 && vdWindowLeft <= (int)(machInst.vl - 1) && (int)(machInst.vl - 1) < vdWindowRight) {
+                        int posInMicro = (int)(machInst.vl - 1) - vdWindowLeft;
+                        if (posInMicro < microVl) {
+                            if (this->vm || elem_mask(v0, (int)(machInst.vl - 1))) {
+                                auto fd = ftype_freg<et>(freg(Fs1_bits));
+                                Vd_vu[posInMicro] = fd.v;
+                            }
                         }
                     }
                 }}, OPFVF, SimdMiscOp);
@@ -5903,10 +5905,13 @@ decode QUADRANT default Unknown::unknown() {
                             }
                         }
                     }
-                    if (lastUop) {
-                        if (this->vm || elem_mask(v0,
-                            vdWindowLeft + microVl - 1)) {
-                            Vd_vu[microVl - 1] = Rs1_vu;
+                    if (machInst.vl > 0 && vdWindowLeft <= (int)(machInst.vl - 1)
+                        && (int)(machInst.vl - 1) < vdWindowRight) {
+                        int posInMicro = (int)(machInst.vl - 1) - vdWindowLeft;
+                        if (posInMicro < microVl) {
+                            if (this->vm || elem_mask(v0, (int)(machInst.vl - 1))) {
+                                Vd_vu[posInMicro] = Rs1_vu;
+                            }
                         }
                     }
                 }}, OPIVX, SimdMiscOp);
--- a/src/arch/riscv/isa/decoder.isa
+++ b/src/arch/riscv/isa/decoder.isa
@@ -5552,12 +5552,15 @@ decode QUADRANT default Unknown::unknown() {
                             }
                         }
                     }
-                    if (machInst.vl > 0 && vdWindowLeft <= (int)(machInst.vl - 1) && (int)(machInst.vl - 1) < vdWindowRight) {
-                        int posInMicro = (int)(machInst.vl - 1) - vdWindowLeft;
-                        if (posInMicro < microVl) {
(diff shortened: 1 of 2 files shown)


saul44203 (commented):
I think I need more info of the failing case.

The macro constructor already instantiates µops to consume vl, not vlmax, so element vl-1 should always be written by the last µop implicitly.

amatabsc:
> I think I need more info of the failing case.
> 
> The macro constructor already instantiates µops to consume vl, not vlmax, so element vl-1 should always be written by the last µop implicitly.

ah yes, you're right that the loop only generates µops while micro_vl > 0, so no µops are emitted beyond vl. However, lastUop is not equivalent to "the µop that handles element vl-1", it's a separate boolean set by i == (num_microops - 1) || micro_vl < micro_vlmax.
In a failing case (vl=6, LMUL=4, SEW=64, micro_vlmax=2), num_microops = vtype_regs_per_group(vtype) = 4. The loop generates 3 µops (i=0,1,2) and stops when micro_vl hits 0 at i=3. For each of those µops: i == (num_microops - 1) is i == 3 which is never true for any generated µop, and micro_vl < micro_vlmax is 2 < 2 which is also false throughout. So lastUop ends up false for all generated µops and Rs1 never gets written anywhere, which is exactly what the cosim trace shows.
The window-based condition in this fix (vl-1 ∈ [vdWindowLeft, vdWindowRight)) correctly identifies which µop owns element vl-1 regardless of how lastUop resolves, and is also more explicit about the intent.

saul44203:
> > I think I need more info of the failing case.
> > The macro constructor already instantiates µops to consume vl, not vlmax, so element vl-1 should always be written by the last µop implicitly.
> 
> ah yes, you're right that the loop only generates µops while micro_vl > 0, so no µops are emitted beyond vl. However, lastUop is not equivalent to "the µop that handles element vl-1", it's a separate boolean set by i == (num_microops - 1) || micro_vl < micro_vlmax. In a failing case (vl=6, LMUL=4, SEW=64, micro_vlmax=2), num_microops = vtype_regs_per_group(vtype) = 4. The loop generates 3 µops (i=0,1,2) and stops when micro_vl hits 0 at i=3. For each of those µops: i == (num_microops - 1) is i == 3 which is never true for any generated µop, and micro_vl < micro_vlmax is 2 < 2 which is also false throughout. So lastUop ends up false for all generated µops and Rs1 never gets written anywhere, which is exactly what the cosim trace shows. The window-based condition in this fix (vl-1 ∈ [vdWindowLeft, vdWindowRight)) correctly identifies which µop owns element vl-1 regardless of how lastUop resolves, and is also more explicit about the intent.

Thanks @amatabsc. 

So basically the error is here:
https://github.com/gem5/gem5/blob/99ae06b64bf54e528975de9ca0fa54db0e935bc1/src/arch/riscv/isa/templates/vector_arith.isa#L2622

It fails to mark the last µop correctly as we have less µops than LMUL and the VL happens to be a multiple of µVLMAX.

We could fix it two ways. Either fix the comparison to also cover this case, or just use the `IsLastMicroop` flag in `StaticInst`. I think the later case is easier and this way we can remove `lastUop` completely. We just replace it with `isLastMicroop()` instead and no additional logic is needed, Both are valid, up to you.

amatabsc:
> > > I think I need more info of the failing case.
> > > The macro constructor already instantiates µops to consume vl, not vlmax, so element vl-1 should always be written by the last µop implicitly.
> > 
> > 
> > ah yes, you're right that the loop only generates µops while micro_vl > 0, so no µops are emitted beyond vl. However, lastUop is not equivalent to "the µop that handles element vl-1", it's a separate boolean set by i == (num_microops - 1) || micro_vl < micro_vlmax. In a failing case (vl=6, LMUL=4, SEW=64, micro_vlmax=2), num_microops = vtype_regs_per_group(vtype) = 4. The loop generates 3 µops (i=0,1,2) and stops when micro_vl hits 0 at i=3. For each of those µops: i == (num_microops - 1) is i == 3 which is never true for any generated µop, and micro_vl < micro_vlmax is 2 < 2 which is also false throughout. So lastUop ends up false for all generated µops and Rs1 never gets written anywhere, which is exactly what the cosim trace shows. The window-based condition in this fix (vl-1 ∈ [vdWindowLeft, vdWindowRight)) correctly identifies which µop owns element vl-1 regardless of how lastUop resolves, and is also more explicit about the intent.
> 
> Thanks @amatabsc.
> 
> So basically the error is here:
> 
> https://github.com/gem5/gem5/blob/99ae06b64bf54e528975de9ca0fa54db0e935bc1/src/arch/riscv/isa/templates/vector_arith.isa#L2622
> 
> It fails to mark the last µop correctly as we have less µops than LMUL and the VL happens to be a multiple of µVLMAX.
> 
> We could fix it two ways. Either fix the comparison to also cover this case, or just use the `IsLastMicroop` flag in `StaticInst`. I think the later case is easier and this way we can remove `lastUop` completely. We just replace it with `isLastMicroop()` instead and no additional logic is needed, Both are valid, up to you.

Done, can you give it a look?

saul44203 (approved):

powerjg (approved):

PR #3107, opened 2026-04-22
stdlib,gpu-compute: GPU exit events for stdlib

abmerop:
This adds exit events which were available in classic configs and imports to stdlib. It adds a new exit event at kernel start. These can be used, for example, to reset/dump stats per-kernel with stdlib.

### DIFF
--- a/src/gpu-compute/dispatcher.cc
+++ b/src/gpu-compute/dispatcher.cc
@@ -120,6 +120,10 @@ GPUDispatcher::dispatch(HSAQueueEntry *task)
 {
     ++stats.numKernelLaunched;
 
+    if (kernelExitEvents) {
+        exitSimLoopNow("GPU Kernel Started");
+    }
+
     DPRINTF(GPUDisp, "launching kernel: %s, dispatch ID: %d\n",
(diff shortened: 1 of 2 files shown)


powerjg (approved):
This is OK to replicate the current support. However, I want to deprecate this string matching code. It's hard to maintain and not best practice.

I know it's not document well, but I *really* would prefer if we could start using the hypercall exits. See https://github.com/gem5/gem5/blob/stable/src/sim/sim_exit.hh#L64

You should reserve a new number for each of these events.

@BobbyRBruce please get #2983 updates ASAP so we can start improving the documentation around this.

abmerop:
> This is OK to replicate the current support. However, I want to deprecate this string matching code. It's hard to maintain and not best practice.
> 
> I know it's not document well, but I _really_ would prefer if we could start using the hypercall exits. See https://github.com/gem5/gem5/blob/stable/src/sim/sim_exit.hh#L64
> 
> You should reserve a new number for each of these events.
> 
> @BobbyRBruce please get #2983 updates ASAP so we can start improving the documentation around this.

Can do.  I believe the disk image will require `gem5-bridge` for that, correct?

PR #3110, opened 2026-04-22
arch-vega: More instruction helpers for gfx9

abmerop:
Implements the following instruction helpers:
- SDWA helpers for floating point types. Essentially avoids bitwise operations in the current SDWA helpers which are illegal on FP types.
- VOP1 helper to handle vanilla, SDWA, and DPP variants with only one functional instruction definition.
- Adds missing VGPR index to string names for disassembly / human-readable instruction.


src/arch/amdgpu/vega/insts/inst_util.hh
```diff
@@ -787,6 +787,29 @@ sdwaInstDstImpl(T &dstOper, T &origDstOper, const bool clamp,
...
+    // Only can select dword. At this size, there are no unused bits
+    // so unusedBits_format does not matter. We might still want to
+    // clamp though.
+    assert(sel == SDWA_DWORD);
+
+    if (clamp) {
+        for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
+            dstOper[lane] = std::clamp(dstOper[lane], 0.0f, 1.0f);
+        }
+    }
```
mattsinc, line 810: should we put a WARN_ONCE in an else here?  Right now this function only has a clamp case ...

mattsinc (dismissed):

src/arch/amdgpu/vega/insts/inst_util.hh
abmerop: Clamp == 0 (the default / common case) is essentially a NOP. It's not incorrect or anything which is why I am not checking for it. I believe for FP instructions clamp is the only output modifier for SDWA so checking for clamp should be all we have to do here.

mattsinc: All right ... I guess we can try it this way and see how it goes.

mattsinc (approved):

PR #3113, opened 2026-04-22
gpu-compute: Emulate copy aligned BLIT kernels

abmerop:
Detect and emulate blit kernels. Currently only supports the Copy Aligned BLIT kernel. Unaligned copy and Fill BLIT kernels have not been regularly observed in practice, so in those cases the kernels are simulated.

This is intended as a performance improvement. Ranges from nice simulation reduction (static apps) to massive improvements (apps using hipModuleLoad for example). Note that this does NOT do timing stores to device memory which will cause problems for writeback cache configurations. Since the default is write-through this is left for future work. Device writes also happen one byte at a time. The improvement could be slightly better writing pages at a time.

This can be turned on using the --emulate-blit-kernels command line argument for classic GPUFS configs (configs in configs/) or enabled in stdlib by setting emulate_blits = True in each GPUCommandProcessor.

### DIFF
--- a/configs/example/gpufs/runfs.py
+++ b/configs/example/gpufs/runfs.py
@@ -229,6 +229,14 @@ def addRunFSOptions(parser):
         help="Frequency in exec cycles of GPU progress prints",
     )
 
+    parser.add_argument(
+        "--emulate-blit-kernels",
+        default=False,
+        action="store_true",
+        help="Emulate blit kernels as functional copies instead of running"
+        "GPU kernel. Intended to improve simulation performance.",
+    )
+
 
 def runGpuFSSystem(args):
     """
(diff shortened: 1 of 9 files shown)


abmerop:
@mattsinc Some context: We originally wrote this when attempting to simulate LLaMA3 in gem5. This made it possible to do it in reasonable time. Recently PyTorch has been seemingly adding preloading of code objects which is using blit kernels and it speeds that up as well. Should be used sparingly though and does not work with write back caches.

mattsinc:
@abmerop just curious what the issue with this one was?

abmerop:
> @abmerop just curious what the issue with this one was?

I found an issue with BLIT kernels that aren't supported + kernarg preload I need to fix.  I'd rather do that before pushing anything.  I think it will be a minor change if you wanted to look over this in the mean time.  I'll reopen this once I have the fix tested.

abmerop:
Issue fixed. Another thing that would be a lot easier if we knew the kernel name :)

mattsinc:
> Issue fixed. Another thing that would be a lot easier if we knew the kernel name :)

Yes ... I don't think Tony T ever replied to my email about that?

PR #3114, opened 2026-04-22
mem-ruby: Extend slicc to ease interaction with SimObjects

giactra:
One of the biggest shortcomings of slicc is the difficulty in integrating SimObjects within the language.
SimObjects are the building blocks of gem5 and allow modularity and specialization at the same time.

It would be nice if we could offload some of the non protocol logic (outside of coherence handling) from slicc
files to C++ files.
While we can already integrate SimObjects with the slicc Cache_Controller, there is no way for the SimObject to
gain access to its internal state.

This is mainly cause the autogenerated methods (examples are in CHI-cache-funcs.sm) are defined as private within
the controller. They are in fact meant to be internal methods only.

With this patch we add the 2 following improvements:

1) We extend the language with the "PUBLIC" keyword, which describes a method as being part of the public interface.

So if I have the following method:
```

int tbeSize() {
    return storTBEs.size()
}

// autogenerated file:
class Cache_Controller
{
  private:
    int tbeSize();
};

```
Marking it as public will transform the code as follows

```
public int tbeSize() {
    return storTBEs.size()
}

// autogenerated file:
class Cache_Controller
{
  public:
    int tbeSize();
  private:
    [..]
};
```

2) To be able to use the aforementioned public methods, a SimObject needs to get hold of a pointer of the
Coherence_Controller. To do so, we define the "this" keyword, so that a parent controller can pass its
pointer to the child SimObject

### DIFF
--- a/src/mem/slicc/ast/FuncDeclAST.py
+++ b/src/mem/slicc/ast/FuncDeclAST.py
@@ -1,3 +1,16 @@
+# -*- mode:python -*-
+# Copyright (c) 2026 Arm Limited
+# All rights reserved.
+#
+# The license below extends only to copyright in the software and shall
+# not be construed as granting a license to any other intellectual
+# property including but not limited to intellectual property relating
+# to a hardware implementation of the functionality of the software
+# licensed hereunder.  You may use the software subject to the license
+# terms below provided that you ensure that this notice is replicated
+# unmodified and in its entirety in all distributions of the software,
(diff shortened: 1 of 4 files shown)


abmerop (approved):
LGTM

giactra:
@powerjg Waiting for your feedback before merging

powerjg (approved):

PR #3115, opened 2026-04-23
cpu-o3: add LeastLoaded IQ insertion policy

jiegec:
For early CPU models with a unified IQ, the existing behavior works fine. However, newer models like Neoverse-V2 have multiple IQs for the same OpClass (e.g. IntAlu appears in 4 IQs). With the previous FirstMatch behavior (which this commit names), some FUs can be underutilized because they wait for another IQ to fill up first. For example, if IQ0 and IQ1 both have ALU slots, the ALU in IQ1 may only be utilized after IQ0 is full.

This commit adds a LeastLoaded insertion policy that distributes instructions evenly across IQs, and names the previous behavior FirstMatch. On a microbenchmark, this improves IPC from 3.72 to 4.00, matching the theoretical ceiling determined by the number of FUs.

### DIFF
--- a/src/cpu/o3/BaseO3CPU.py
+++ b/src/cpu/o3/BaseO3CPU.py
@@ -185,6 +185,9 @@ class BaseO3CPU(BaseCPU):
     # most ISAs don't use condition-code regs, so default is 0
     numPhysCCRegs = Param.Unsigned(0, "Number of physical cc registers")
     instQueues = VectorParam.IQUnit(IQUnit(), "Vector of IQs")
+    iqInsertionPolicy = Param.IQInsertionPolicy(
+        "FirstMatch", "Policy for inserting instructions to IQs"
+    )
     numROBEntries = Param.Unsigned(192, "Number of reorder buffer entries")
 
     smtNumFetchingThreads = Param.Unsigned(1, "SMT Number of Fetching Threads")
--- a/src/cpu/o3/inst_queue.cc
+++ b/src/cpu/o3/inst_queue.cc
@@ -248,6 +248,7 @@ InstructionQueue::InstructionQueue(CPU *cpu_ptr, IEW *iew_ptr,
     : cpu(cpu_ptr),
       iewStage(iew_ptr),
       iqs(params.instQueues),
+      insertionPolicy(params.iqInsertionPolicy),
       numThreads(params.numThreads),
       totalWidth(params.issueWidth),
       commitToIEWDelay(params.commitToIEWDelay),
@@ -692,13 +693,43 @@ IQUnit *
 InstructionQueue::findIQ(const DynInstPtr &inst)
 {
     bool any_capable = false;
-    for (auto iq : iqs) {
-        if (iq->isCapable(inst->opClass())) {
-            any_capable = true;
-            if (iq->numFreeEntries(inst) > 0) {
-                return iq;
+    if (insertionPolicy == IQInsertionPolicy::LeastLoaded) {
+        // LeastLoaded insertion policy
+        // Find the least loaded IQ to insert
+
+        IQUnit *best_iq = nullptr;
+        unsigned best_load = 0;
+
+        for (auto iq : iqs) {
+            if (iq->isCapable(inst->opClass())) {
+                any_capable = true;
+                if (iq->numFreeEntries(inst) > 0) {
+                    unsigned load = iq->numEntries() - iq->numFreeEntries();
+                    if (!best_iq || load < best_load) {
+                        best_iq = iq;
+                        best_load = load;
+                    }
+                }
             }
         }
+
+        if (best_iq) {
+            return best_iq;
+        }
+    } else if (insertionPolicy == IQInsertionPolicy::FirstMatch) {
+        // FirstMatch insertion policy (default)
+        // Find the first non-full IQ to insert
+
+        for (auto iq : iqs) {
+            if (iq->isCapable(inst->opClass())) {
+                any_capable = true;
+                if (iq->numFreeEntries(inst) > 0) {
+                    return iq;
+                }
+            }
+        }
+    } else {
+        fatal("Unknown IQ insertion policy %d\n", (int)insertionPolicy);
     }
 
     if (any_capable) {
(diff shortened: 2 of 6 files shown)


src/cpu/o3/inst_queue.cc
```diff
@@ -672,13 +673,36 @@ InstructionQueue::hasReadyInsts()
...
+
+        IQUnit *best_iq = nullptr;
+        unsigned best_load = 0;
+
+        for (auto iq : iqs) {
+            if (iq->numFreeEntries(inst) > 0) {
+                unsigned used = iq->numEntries() - iq->numFreeEntries();
+                if (!best_iq || used < best_load) {
+                    best_iq = iq;
+                    best_load = used;
```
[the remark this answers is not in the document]
jiegec: Good idea, but I would like to hear from maintainers, which policy is preferred.

jiegec:
Currently, when counting the number of occupied entries, SMT is not taken into account; instead, the total occupied count of the IQ is used. If the maintainer wants, I can change it to be based on count[tid], i.e., the number actually used by the current SMT thread, but then the overall load balancing may be less effective. I'm not sure which implementation aligns better with the hardware.

giactra:
Thanks, this makes lots of sense. When I added the distributed IQ I realised it was with a very naif insertion policy. My concern now is that this PR is slightly conflicting with

https://github.com/gem5/gem5/pull/3073

Would you mind if we merge 3073 first and ask you to rebase on top of it?

jiegec:
> Thanks, this makes lots of sense. When I added the distributed IQ I realised it was with a very naif insertion policy. My concern now is that this PR is slightly conflicting with
> 
> #3073
> 
> Would you mind if we merge 3073 first and ask you to rebase on top of it?

Sure! I'm actually using #3073 in my internal repo as well; it's very useful. In addition, there will be a follow‑up PR that changes the ready queue from unified (current) to per‑IQ. This avoids head‑of‑line blocking when one IQ’s functional units are busy: without this change, instructions in other IQs could fail to be scheduled due to the single shared ready queue.

Some initial figures for this PR and a follow‑up one:

- Current develop branch: achieves 44.5% peak IPC  
- With this PR (LeastLoaded): achieves 52.0% peak IPC  
- With a follow‑up PR (PerIQ): achieves 93.9% peak IPC  
- With both this PR (LeastLoaded) and the follow‑up PR (PerIQ): achieves 100.0% peak IPC  

Testing methodology: Microbenchmark measures throughput for different instructions. The observed IPC should match the number of functional units that can process the given OpClass.

src/cpu/o3/BaseO3CPU.py
```diff
@@ -185,6 +185,9 @@ def support_take_over(cls):
     # most ISAs don't use condition-code regs, so default is 0
     numPhysCCRegs = Param.Unsigned(0, "Number of physical cc registers")
     instQueues = VectorParam.IQUnit(IQUnit(), "Vector of IQs")
+    iqInsertionPolicy = Param.IQInsertionPolicy(
+        "FirstMatch", "Policy for inserting instructions to IQs"
```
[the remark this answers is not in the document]
jiegec: If requested, I can change the default policy for neoverse v2 to the new one.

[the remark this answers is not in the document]
dhschall: The NeoverseV2 was never correlated with a real core. It was just configured based on publicly available data.
I would say that ARM has probably a better policy so IMO its ok to change it to the new policy

src/cpu/o3/inst_queue.cc
dhschall, line 735: I know it doesn't belong to this PR, but @giactra, why do we have this? Isn't it somehow useless?

src/cpu/o3/inst_queue.cc
```diff
@@ -692,11 +693,39 @@ IQUnit *
...
+                        best_load = load;
+                    }
+                }
+            }
+        }
+
+        if (best_iq) {
+            return best_iq;
+        }
+    } else {
```
dhschall:
```suggestion
    } else if (insertionPolicy == IQInsertionPolicy::FirstMatch ) { 
```
I would make the first match policy explicit. and an `fatal`  in the else branch. As of now any other configuration than `LeastLoaded` will use `FirstMatch` and we should catch any misconfiguration

dhschall (changes requested):
Thanks for the contribution! I only have a small request. IMO, you should change the default of the NeoverseV2 config.

src/cpu/o3/inst_queue.cc
jiegec: Done!

jiegec:
> Thanks for the contribution! I only have a small request. IMO, you should change the default of the NeoverseV2 config.

Done! The commit message is changed accordingly.

src/cpu/o3/inst_queue.cc
giactra: Yeah, I think we should add a comment here because this logic is a bit complex to understand if you don't know the history of this code. Basically if we get to this line of code it is because of one of the two options

1) We don't have available IQUnit slots because all IQUnits are FULL
2) We don't have available IQUnit slots because the instruction has an unrecognised OpClass. Which means there is no IQUnit which can host the instruction (but the IQUnits could even be empty!)

So the condition here says:

"If there are capable IQUnits it means that we are in option 1: just return nullptr to signal to earlier stages the backend is FULL and they should stall"

If there are not capable IQUnits, we don't want to return nullptr. Instead we simply return the number of free SLOTs *as if the unrecognised instruction could be hosted by any IQUnit*.
Why do we do that?

Because we only panic for unrecognised instruction once we execute it. But if we return nullptr we never execute it. Because rename believes that nullptr means there is no IQUnit slot and therefore stalls indefinitely.

Could we panic earlier than execute? No because it does not allow the case of guest software trapping the instruction (this happens at execute in gem5) like in the case where we do trap and emulate on embedded to workaround the absence of a FPU...

erin-le:
@dhschall: Hello, could you take another look at this when you get a chance? Thank you!

src/cpu/o3/inst_queue.cc
dhschall: Ok, thanks for the clarification!

dhschall (approved):
Great it looks good to me!
Thanks for adding this.

PR #3116, opened 2026-04-23
stdlib: Avoid `resource_tracker` spawn from tqdm

BobbyRBruce:
When gem5 uses tqdm for resource download/md5 progress, tqdm creates a multiprocessing `RLock` by default. Under gem5, this launches an extra process:

```shell
<gem5.opt> -c 'from multiprocessing.resource_tracker import main;...'
```

This caused duplicate gem5 banner output during normal single-run workloads and made logs noisy/confusing.

This change fixes this by setting tqdm's global lock to a thread-only `RLock` in `src/python/gem5/utils/progress_bar.py`. This preserves progress bar synchronization without creating multiprocessing semaphores, so `resource_tracker` is not started.

The fix applies the lock override to both `tqdm.auto.tqdm` and `tqdm.std.tqdm` to cover the class used at runtime.

### DIFF
--- a/src/python/gem5/utils/progress_bar.py
+++ b/src/python/gem5/utils/progress_bar.py
@@ -48,9 +48,18 @@ class FakeTQDM:
 
 
 try:
+    from threading import RLock
+
     from tqdm.auto import tqdm
+    from tqdm.std import tqdm as std_tqdm
 
     _have_tqdm = True
+
+    # tqdm's default global lock includes a multiprocessing RLock, which may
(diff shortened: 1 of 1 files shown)


Harshil2107 (approved):
lgtm

erin-le (dismissed):
I am having trouble reproducing the issue that this PR is trying to fix, so it's now on Bobby to investigate.

powerjg (approved):

PR #3117, opened 2026-04-23
arch-vega: Fix calcAddr bug with scratch_ instructions

abmerop:
An offset of 0 is valid. The comment is correct and this should only check the value is not negative.


erin-le:
@mattsinc: Hello, just a light reminder to review this PR, https://github.com/gem5/gem5/pull/3110, and https://github.com/gem5/gem5/pull/3106 when you get a chance! Thank you!

mattsinc (approved):

PR #1084, opened 2024-04-30
mem-ruby, configs: Add a CHI-TLM controller + testing

giactra:
This PR is implement a CacheController which acts as a bridge between the
AMBA TLM 2.0 implementation of CHI [[1]](https://developer.arm.com/documentation/101459/latest) [[2](https://developer.arm.com/Architectures/AMBA#Downloads)] with the gem5 (ruby) one.

In other words it translates AMBA CHI transactions into ruby
messages (which are then forwarded to the MessageQueues)
and viceversa.

```
ARM::CHI::Payload,         CHIRequestMsg
                     <-->  CHIDataMsg
ARM::CHI::Phase            CHIResponseMsg
                           CHIDataMsg
```

To be able to do so, it first adds a CHIGenericController class to model a coherence controller entirely
in C++ to work with ruby. Adding a CHIGenericController removes the limitation of using the gem5 Sequencer
and associated data structures, therefore allowing the aforementioned CHI to CHI bridge/controller
and potentially a hybrid setup with classic caches + ruby.

The PR also adds a CHI tester, which is a transaction generator directly connected to the CHI-TLM controller. It will
be possible to write small unit tests in python (the library has been wrapped with pybind) for CHI transactions and unlock unit-testing for ruby

PR #3121, opened 2026-04-24
configs: Add a TLM-generator testing platform

giactra:
This commit is adding a new TLM-generator based testing platform. This is somewhat similar to the platform introduced in the original CHI-TLM PR [1].

However, while the original platform in tests/gem5/chi_tlm_tests was focusing on unit-testing the CHI implementation in ruby, this platform focuses on testing the system as a whole while using the TlmGenerator as a stimulus injector.

This can be seen in the way we structure the suites-config files:

'''
for verifier in suite.test_all(root.system, system.cpu[0]):
    while True:
        exit_event = m5.simulate(args.abs_max_tick)

        if exit_event.getCause() != "TlmGenerator done" or not any(
            c.isActive() for c in system.cpu
        ):
            break

    # Call the suite verifier
    verifier(root.system)
'''

The suite.test_all is called now in a for loop and for every iteration it generates a sets of transactions and returns a verifier, which is a callback supposed to check if simulation lead to the correct results.

As an example, we provide a simple memory tester, which compares for instance the number of R/W bytes written in memory by extracting those from the SimpleMemory gem5 stats. Working with python stats allows us to verify lots of microarchitectural state. We envision this kind of platform could/should be used to write tests for any major component and encourages developers to define a rich stat system.
Stats could be used now not only to assess performance, but also to make sure a component is functionally correct at a uarch level. We could double check number of hits/misses in a cache, or the mpki, or the outgoing BW of a component...

[1]: https://github.com/gem5/gem5/pull/1084

Change-Id: I4086b1a9645a53f6e634df0a23dce6cdee595d2c

### DIFF
--- a/configs/example/chi_tlm/tlm_generator_test.py
+++ b/configs/example/chi_tlm/tlm_generator_test.py
@@ -1,0 +1,286 @@
+# Copyright (c) 2024-2025 Arm Limited
+# All rights reserved.
+#
+# The license below extends only to copyright in the software and shall
+# not be construed as granting a license to any other intellectual
+# property including but not limited to intellectual property relating
+# to a hardware implementation of the functionality of the software
+# licensed hereunder.  You may use the software subject to the license
+# terms below provided that you ensure that this notice is replicated
+# unmodified and in its entirety in all distributions of the software,
+# modified or unmodified, in source code or in binary form.
+#
+# Copyright (c) 2006-2007 The Regents of The University of Michigan
+# Copyright (c) 2009 Advanced Micro Devices, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met: redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer;
+# redistributions in binary form must reproduce the above copyright
+# notice, this list of conditions and the following disclaimer in the
+# documentation and/or other materials provided with the distribution;
+# neither the name of the copyright holders nor the names of its
+# contributors may be used to endorse or promote products derived from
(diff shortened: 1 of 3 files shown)


configs/example/chi_tlm/tlm_generator_test.py
```diff
@@ -0,0 +1,256 @@
...
+import sys
+
+import m5
+from m5.defines import buildEnv
+from m5.objects import *
+from m5.tlm_chi.utils import *
+from m5.util import addToPath
+
+m5.util.addToPath(os.path.join(m5.util.repoPath(), "configs"))
+from common import Options
```
powerjg: If at all possible, I would like to not have a new dependence on `common.Options`. Is it possible to refactor this to not use this module?

powerjg (commented):
Would it be possible to refactor this to use the stdlib instead of Options.py? Maybe you can ask AI to do it?

configs/example/chi_tlm/tlm_generator_test.py
giactra: Done, Options has been removed

giactra:
> Would it be possible to refactor this to use the stdlib instead of Options.py? Maybe you can ask AI to do it?

Hi @powerjg, thanks for the review.
I removed Options usage but I am afraid we cannot fully transition to stdlib because stdlib does not currently support CustomMesh which is one of the main usecases when it comes to topology. Luckily I am working on integrating it into stdlib so we will be able to fully transition sometime down the line

powerjg (approved):
Thanks for the update. Ruby.py and the other scripts that have been used with se/fs.py are such a mess. It makes me sad to see them used.

I appreciate the effort it takes to update to using the stdlib.

PR #3125, opened 2026-04-27
scons: Support for GCC 15.2

shuyand:
**SConstruct**: Bumped `gcc_max_version` to `15.2` to enable GCC 15 support.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -683,7 +683,7 @@ for variant_path in variant_paths:
 
     if env['GCC']:
         gcc_min_version = "11"
-        gcc_max_version = "14.2"
+        gcc_max_version = "15.2"
         gcc_version = env['CXXVERSION']
         if compareVersions(gcc_version, gcc_min_version) < 0 or \
               compareVersions(gcc_version, gcc_max_version) > 0:


powerjg (approved):

powerjg (approved):

src/base/statistics.hh
```diff
@@ -956,9 +956,13 @@ class VectorBase : public DataWrapVec<Derived, VectorInfoProxy>
...
 
-        storage.reserve(s);
+        // Use a local temporary vector to avoid false positive -Warray-bounds
+        // warnings from both GCC 15 and Clang.
+        std::vector<Storage *> temp;
+        temp.reserve(s);
         for (size_type i = 0; i < s; ++i)
-            storage.push_back(new Storage(this->info()->getStorageParams()));
+            temp.push_back(new Storage(this->info()->getStorageParams()));
+        storage = std::move(temp);
```
rogerchang23424: The issue is fixed in https://github.com/gem5/gem5/pull/3145. should we need to keep the change?

src/base/statistics.hh
```diff
@@ -1406,9 +1410,13 @@ class VectorDistBase : public DataWrapVec<Derived, VectorDistInfoProxy>
...
```
rogerchang23424: ditto

src/base/statistics.hh
powerjg: Good find!

powerjg (changes requested):

src/base/statistics.hh
shuyand: Thanks a lot! Can verify that gcc15 now builds the develop branch without error. I have therefore removed my workaround for [-Werror=array-bounds=] so this PR will only contain the supported gcc version bump to 15.2.

powerjg (approved):

PR #3133, opened 2026-04-29
mem-ruby: Implementation of CHI Cbusy

giactra:
This PR implements the Completer Busy field in CHI. According to the CHI spec:

```
The Completer Busy indication is a mechanism for the Completer of a transaction to indicate its current level of
activity. This signaling provides additional information to a Requester on how aggressive it can be in generating
speculative activity to improve performance
```

The patch implements CBusy generation by defining a generic BackpressureGen class and providing a
CBusy generator implementation which matches the example contained in the CHI spec.
This uses the fullness of TBETables as a proxy for fullness.
More specifically:

     * cbusy:0 = occupancy < 50%
     * cbusy:1 = 50%-75%
     * cbusy:2 = 75%-90%
     * cbusy:3 = >90%

It has to be noted this PR does not consume the received CBusy value to curb aggressiveness of speculative activity. This will be provided by a future PR.

Also because of the following CHI spec suggestion:
```
It is IMPLEMENTATION DEFINED how a CBusy indication is set by the Completer and how it should be interpreted
by a Requester. However, it is recommended that implementations consider mechanisms to avoid a few busy
Completers among many skewing the results.
```

We add a BackpressureTracker class in addition the BackpressureGen, which is responsible on aggregating results to provide the aforementioned stability and avoid skewing results

### DIFF
--- a/configs/ruby/CHI_config.py
+++ b/configs/ruby/CHI_config.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2021-2024 Arm Limited
+# Copyright (c) 2021-2024, 2026 Arm Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
@@ -257,6 +257,8 @@ class CHI_L1Controller(Base_CHI_Cache_Controller):
         self.cache = cache
         self.prefetcher = prefetcher
         self.use_prefetcher = prefetcher != NULL
+        self.cbusy_generator = NULL
+        self.cbusy_tracker = CBusyTracker()
         self.send_evictions = True
         self.is_HN = False
         self.enable_DMT = False
@@ -295,6 +297,8 @@ class CHI_L2Controller(Base_CHI_Cache_Controller):
         self.cache = cache
         self.prefetcher = prefetcher
         self.use_prefetcher = prefetcher != NULL
+        self.cbusy_generator = CBusy()
+        self.cbusy_tracker = CBusyTracker()
         self.allow_SD = True
         self.is_HN = False
         self.enable_DMT = False
@@ -357,6 +361,8 @@ class CHI_HNFController(Base_CHI_Cache_Controller):
         self.number_of_DVM_TBEs = 1  # should not receive any dvm
         self.number_of_DVM_snoop_TBEs = 1  # should not receive any dvm
         self.unify_repl_TBEs = False
+        self.cbusy_generator = CBusy()
+        self.cbusy_tracker = CBusyTracker()
 
 
 class CHI_MNController(CHI_MiscNode_Controller):
@@ -409,6 +415,8 @@ class CHI_DMAController(Base_CHI_Cache_Controller):
 
         self.prefetcher = NULL
         self.use_prefetcher = False
+        self.cbusy_generator = NULL
+        self.cbusy_tracker = CBusyTracker()
         self.cache = DummyCache()
         self.sequencer.dcache = NULL
         # All allocations are false
@@ -745,6 +753,7 @@ class CHI_SNF_Base(CHI_Node):
             requestToMemory=MemCtrlMessageBuffer(),
             reqRdy=TriggerMessageBuffer(),
             transitions_per_cycle=1024,
+            cbusy_generator=SnfCBusy(),
         )
 
         # The Memory_Controller implementation deallocates the TBE for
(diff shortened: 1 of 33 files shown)


src/mem/ruby/network/MessageBuffer.cc
```diff
@@ -344,6 +344,18 @@ MessageBuffer::dequeue(Tick current_time, bool decrement_messages)
     return delay;
 }
 
+Tick
+MessageBuffer::oldestMsgStallTicks(Tick current_time)
```
powerjg: How/where is this function used?

powerjg (commented):
Generally, this lgtm. One small question below. Also, it looks like there are some real CI tests failures.

powerjg:
A couple of things after doing a deep review with an LLM:

* **DMT (Direct Memory Transfer) Tracker Blindspot:** In `CHI-cache-actions.sm`, the `UpdateBPTracker_FromResp` action explicitly tracks DMT responses into a separate `dmtEntries` bin (`cbusy_tracker.updateDMT(...)`). However, looking at `CBusyTracker::report()`, it currently completely ignores `dmtEntries` and only returns the max from non-DMT responders. Review if this omission is intentional or a bug.

*   **Missed `cbusy` injection**: In `src/mem/ruby/protocol/chi/CHI-cache-actions.sm`, the action `Send_CompUD_PD` (lines 3303-3313) does **not** set `out_msg.cbusy := getCurrCBusy(out_msg);`. It defaults to `-1` (as defined in `CHI-msg.sm`).
    *   **Crash on Assert**: Several transitions in `CHI-cache-actions.sm` (e.g., `UpdateBPTracker_FromDatalessResp` and others) contain the assertion `assert(in_msg.cbusy >= 0);`. If a message from `Send_CompUD_PD` triggers one of these transitions, the assertion will fail and crash the simulation.

*   **Broken Override**: In `src/mem/ruby/system/Sequencer.hh`, the PR changes the signature of `hitCallback` (adding `const int cBusy`) and removes the `virtual` keyword.
    *   **Failure to Update Derived Class**: In `src/mem/ruby/system/VIPERSequencer.hh`, the declaration of `hitCallback` (lines 54-60) retains the old signature (without `cBusy`). Because the signature no longer matches and the base method is no longer virtual, `VIPERSequencer::hitCallback` no longer overrides the base class method, breaking polymorphism for VIPER-specific requests.

I did not verify these myself, so feel free to ignore if they are wrong.

src/mem/ruby/network/MessageBuffer.cc
giactra: This will be used by a later patch. I will remove it for now and have a dedicated PR for it

giactra:
> A couple of things after doing a deep review with an LLM:
> 
>     * **DMT (Direct Memory Transfer) Tracker Blindspot:** In `CHI-cache-actions.sm`, the `UpdateBPTracker_FromResp` action explicitly tracks DMT responses into a separate `dmtEntries` bin (`cbusy_tracker.updateDMT(...)`). However, looking at `CBusyTracker::report()`, it currently completely ignores `dmtEntries` and only returns the max from non-DMT responders. Review if this omission is intentional or a bug.

Yeah that is the basic tracker implementation. You could define more elaborate trackers which are taking DMT data into consideration instead.

> 
>     * **Missed `cbusy` injection**: In `src/mem/ruby/protocol/chi/CHI-cache-actions.sm`, the action `Send_CompUD_PD` (lines 3303-3313) does **not** set `out_msg.cbusy := getCurrCBusy(out_msg);`. It defaults to `-1` (as defined in `CHI-msg.sm`).
> 

Thanks that was a bug. I fixed it

>       * **Crash on Assert**: Several transitions in `CHI-cache-actions.sm` (e.g., `UpdateBPTracker_FromDatalessResp` and others) contain the assertion `assert(in_msg.cbusy >= 0);`. If a message from `Send_CompUD_PD` triggers one of these transitions, the assertion will fail and crash the simulation.
> 
>     * **Broken Override**: In `src/mem/ruby/system/Sequencer.hh`, the PR changes the signature of `hitCallback` (adding `const int cBusy`) and removes the `virtual` keyword.
>       
>       * **Failure to Update Derived Class**: In `src/mem/ruby/system/VIPERSequencer.hh`, the declaration of `hitCallback` (lines 54-60) retains the old signature (without `cBusy`). Because the signature no longer matches and the base method is no longer virtual, `VIPERSequencer::hitCallback` no longer overrides the base class method, breaking polymorphism for VIPER-specific requests.
> 

Same, we were effectively hiding the VIPER sequencer hitCallback by replacing overriding with overloading. I think in the future the class should be amended to use override specifiers @abmerop but it is beyond the scope of this PR

powerjg (approved):

PR #3136, opened 2026-04-30
python: Fix unhashable SimObjectVector bug in SimObject find_all()

harry900831:
When executing m5.instantiate(), gem5 traverses the simulation tree using find_all() to discover objects of specific types (e.g., AbstractMemory). During this traversal, the function checks if a parameter's inner type (pdesc.ptype) matches the target search type.

When a VectorParam is used (e.g., VectorParam.AbstractMemory), pdesc.ptype matches the target type, but its runtime value is a SimObjectVector container. In original implementation, find_all() erroneously used this container directly as a dictionary key. Since SimObjectVector lacks a __hash__ method, this caused a crash: TypeError: unhashable type: 'SimObjectVector'

This patch modifies find_all() to unpack SimObjectVector values and add their individual SimObjects to the deduplication dictionary, resolving the crash while preserving type safety.

### DIFF
--- a/src/python/m5/SimObject.py
+++ b/src/python/m5/SimObject.py
@@ -1106,7 +1106,12 @@ class SimObject(metaclass=MetaSimObject):
             if issubclass(pdesc.ptype, ptype):
                 match_obj = self._values[pname]
                 if not isproxy(match_obj) and not isNullPointer(match_obj):
-                    all[match_obj] = True
+                    if isSimObjectVector(match_obj):
+                        for item in match_obj:
+                            if not isNullPointer(item):
+                                all[item] = True
+                    else:
+                        all[match_obj] = True
         # Also make sure to sort the keys based on the objects' path to
         # ensure that the order is the same on all hosts
         return sorted(all.keys(), key=lambda o: o.path()), True


powerjg:
This looks reasonable to me. Could you elaborate a bit more on what the incorrect behavior is? E.g., what is going on in instantiate when find_all is called. 

I'm asking because there is no documentation at all in this code, and I really don't know how or why it's used.

harry900831:
I encounter the crash while I was developing a SimObject with below parameter:
```
  memories = VectorParam.AbstractMemory([], "Memories")
```
The call trace:

1. m5.instantiate() initializes the simulation:
  (Located in src/python/m5/simulate.py)
  It first calls a helper function called _fix_all_objects() to prepare the Python
  object tree before passing it to C++.

  2. _fix_all_objects() unproxies all parameters:
  (Located in src/python/m5/simulate.py)
  It loops through all the SimObjects and tells them to resolve their proxy
  parameters (like Parent.any) by calling unproxyParams().

  3. unproxyParams() resolves individual proxy values:
  (Located in src/python/m5/SimObject.py)
  It loops through the object's parameters and calls .unproxy() on any parameter
  that is currently a proxy.

  4. The Proxy object searches the tree:
  (Located in src/python/m5/proxy.py)
  When a proxy (like Parent.any or Parent.any.memories) is asked to resolve
  itself, it needs to look at the surrounding system to find a matching object. To
  do this, the Proxy object calls its own find() method.

  5. find() triggers find_all():
  (Located in src/python/m5/proxy.py)
  To resolve a proxy like system.memories, the proxy searches the system by
  invoking find_all() to gather a list of every matching parameter in the
  hierarchy.

  6. find_all() recursively traverses the tree:
  (Located in src/python/m5/SimObject.py)
  find_all() recursively calls itself on every child SimObject to build the full
  deduplicated dictionary of matching objects (where our crash occurred).

  Summary Traceback:

   1 m5.instantiate()
   2 └── _fix_all_objects()
   3     └── SimObject.unproxyParams()
   4         └── ParamDesc.unproxy()
   5             └── Proxy.unproxy()
   6                 └── Proxy.find()
   7                     └── SimObject.find_all() # <-- The crash!

powerjg (approved):

PR #3149, opened 2026-05-07
gpu-compute: Emulate copy aligned BLIT kernels

abmerop:
Detect and emulate blit kernels. Currently only supports the Copy Aligned BLIT kernel. Unaligned copy and Fill BLIT kernels have not been regularly observed in practice, so in those cases the kernels are simulated.

This is intended as a performance improvement. Ranges from nice simulation reduction (static apps) to massive improvements (apps using hipModuleLoad for example). Note that this does NOT do timing stores to device memory which will cause problems for writeback cache configurations. Since the default is write-through this is left for future work. Device writes also happen one byte at a time. The improvement could be slightly better writing pages at a time.

This can be turned on using the --emulate-blit-kernels command line argument for classic GPUFS configs (configs in configs/) or enabled in stdlib by setting emulate_blits = True in each GPUCommandProcessor.

### DIFF
--- a/src/gpu-compute/blit_emulation.cc
+++ b/src/gpu-compute/blit_emulation.cc
@@ -1,0 +1,317 @@
+/*
+ * Copyright (c) 2026 Advanced Micro Devices, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ * this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following disclaimer in the documentation
+ * and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ * contributors may be used to endorse or promote products derived from this
+ * software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "gpu-compute/blit_emulation.hh"
+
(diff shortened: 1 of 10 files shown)


abmerop:
Same as #3113 .. It seems you need to reopen before force pushing.

src/gpu-compute/blit_emulation.cc
```diff
@@ -0,0 +1,314 @@
+/*
+ * Copyright (c) 2025 Advanced Micro Devices, Inc.
```
mattsinc: 2026?

src/gpu-compute/blit_emulation.cc
```diff
@@ -0,0 +1,314 @@
...
+     * time. Because of this, for now we know the VMID is always 1. Later
+     * the VMID would have to be passed on to the command processor.
+     */
+    int vmid = 1;
+    unsigned tmp_bytes;
+    walker->startFunctional(gpuDevice->getVM().getPageTableBase(vmid),
+                            phys_addr, tmp_bytes, BaseMMU::Mode::Read,
+                            is_system_page);
+
+    DPRINTF(GPUCommandProc, "Blit kernarg data is in %s memory\n",
```
mattsinc, line 70: I can't find the file off the top of my head, but all this code seems eerily similar to code from other files -- am I imaging it?

src/gpu-compute/blit_emulation.cc
```diff
@@ -0,0 +1,314 @@
...
+            DPRINTF(GPUCommandProc, "Blit kernarg[%d] = %#x\n", i, args[i]);
+        }
+    }
+
+    // There are 3 types of blit kernels: Copy Aligned, Copy Misaligned, and
+    // Fill. Typically we see Copy Aligned so implement that first. If we
+    // see the other types then we should run the blit kernel. Copy Misaligned
+    // has 13 SGPRs and Fill has 8 SGPRs. We use the first kernarg that is
+    // zero to detect the type.
+    if (args[8] == 0) {
```
mattsinc: it would be nice if we had a #define or something for 8 so if it changed, we would just be able to change that #define ... but not the end of the world

src/gpu-compute/blit_emulation.cc
```diff
@@ -0,0 +1,314 @@
...
+
+        delete args;
+        if (task->akc()->kernarg_preload_spec_length == 0) {
+            initABI(task);
+        } else {
+            readPreload(task);
+        }
+
+        return;
+    } else if (args[13] == 0) {
```
mattsinc: same with 13

erin-le:
@mattsinc: Hello, could you take a look at this PR when you get a chance? Thank you!

mattsinc:
> @mattsinc: Hello, could you take a look at this PR when you get a chance? Thank you!

@abmerop did you update this patch after my feedback last month?

abmerop:
> > @mattsinc: Hello, could you take a look at this PR when you get a chance? Thank you!
> 
> @abmerop did you update this patch after my feedback last month?

Could you remind me what that was? I don't see any comments here or on the old PR (#3113).

Just updated copyright date and fixed incorrect scalar delete

src/gpu-compute/blit_emulation.cc
mattsinc: Resolved

mattsinc (changes requested):
Ok so apparently my review was left in the preview state for months, sorry.  Posting it now then re-reviewing.

src/gpu-compute/blit_emulation.cc
mattsinc: This still remains (granted you didn't see it until now).

src/gpu-compute/blit_emulation.cc
mattsinc: This remains too

mattsinc (changes requested):
Just a couple minor things

src/gpu-compute/blit_emulation.cc
abmerop: Probably this? https://github.com/gem5/gem5/blob/stable/src/gpu-compute/gpu_command_processor.cc#L239

abmerop:
> Just a couple minor things

Done

src/gpu-compute/blit_emulation.cc
mattsinc: yes, exactly -- I guess my question is if we need both of these, or if we should consolidate into a single implementation that both call?

mattsinc (approved):

src/gpu-compute/blit_emulation.cc
abmerop: I'll add something as part of another PR in the future. I want to revamp how system bit (and VMID) are handled more properly

PR #3155, opened 2026-05-08
mem-ruby: Add python APIs for flushing AbstractController state

giactra:
By exposing the flushing to python, we add a mean for
tearing down the state of the cache (controller). This is very
useful for debug platforms/purposes.

For instance, we could write multiple tests which test
replacement policies for RubyCache. By allowing the cache
to be flushed in between tests, we make sure
tests are properly isolated

This is another step towards making gem5 more testable

### DIFF
--- a/src/mem/ruby/slicc_interface/AbstractController.hh
+++ b/src/mem/ruby/slicc_interface/AbstractController.hh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017,2019-2023 ARM Limited
+ * Copyright (c) 2017,2019-2023, 2026 Arm Limited
  * All rights reserved.
  *
  * The license below extends only to copyright in the software and shall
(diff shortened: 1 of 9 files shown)


src/mem/ruby/slicc_interface/AbstractController.hh
```diff
@@ -180,6 +180,13 @@ class AbstractController : public ClockedObject, public Consumer
     //! Initialize the message buffers.
     virtual void initNetQueues() = 0;
 
+    //! Flush the internal state of the coherence controller
+    virtual void
+    flushController()
+    {
+        fatal("flushController not supported");
```
powerjg: Can you make this print the name. It will help debugging. (I think the code below is correct)

```suggestion
        fatal("flushController not supported in %s", name());
```

src/mem/ruby/structures/CacheMemory.hh
```diff
@@ -108,6 +108,13 @@ class CacheMemory : public SimObject
     // Explicitly free up this address
     void deallocate(Addr address);
 
+    // Explicitly free up this entry
+    void deallocate(AbstractCacheEntry *entry);
+
+    // Flush all entries within the cache.
+    // Mainly used for debug purposes
```
powerjg, line 115: Can you add here " Does not write back dirty data" just to make it explicit.

powerjg (dismissed):
lgtm, just a couple of small documentation requests.

powerjg (approved):

PR #3162, opened 2026-05-11
arch-riscv: Fix segfault in masked vector strided loads

amatabsc:
VlElementMicroConstructor did not register v0 as a source operand when vm=0, despite execute() calling getRegOperand(this, _numSrcRegs-1) to read the mask. This out-of-bounds access caused a segfault.
Affected instructions: vlse8.v, vlse16.v, vlse32.v, vlse64.v and all vlsseg* variants when used with a mask (vm=0).
Fix is one line: add SET_VM_SRC() at the end of the constructor, matching the pattern already used in VlSegMicroConstructor and VleMicroConstructor.


saul44203 (approved):
LGTM

erin-le (approved):

PR #3165, opened 2026-05-11
dev, gpu-compute: fix C++20 implicit this capture in lambdas

Harshil2107:
After C++20 change, capturing 'this' implicitly via '[=]' is deprecated. Added explicit 'this' to lambda capture lists to fix the resulting - Wdeprecated warnings.


abmerop (dismissed):

mattsinc (dismissed):

abmerop:
The three copilot suggestions should be committed too. It seems Apple Clang does not like the unused captures.

Harshil2107:
I addressed the copilot comments and those should fix the mac build errors as well.

abmerop (approved):

mattsinc (approved):

PR #3172, opened 2026-05-14
arch-arm: Fix ESR ISS2 bit offset

Mark-Gibson-44:
The dirtyBit subBitUnion field was using a relative offset instead of an absolute offset which would lead to incorrect behaviour when set.


giactra (approved):
Thanks for fixing this!

PR #3174, opened 2026-05-15
sim,mem: Add zstd/raw as backing store compression options

hnpl:
I recently found myself taking a checkpoint of a system with 192GiB of RAM, and I found that the zlib was too slow for compressing the backing store (it was faster to forge this solution and take a checkpoint than waiting for zlib to finish).

This change adds a hard dependency on zstd, and an option to use zstd as the compression library for compressing the backing store.
By default, the checkpoint still uses gzip for backward compatibility.

|Compression Method| Checkpoint size | Saving checkpoint time (on Ampere Altra) | Restoring checkpoint time (on Zen 4) |
|-|-|-|-|
|Raw 48G|48G|?|?|
|Raw 192G|192G|?|?|
|gzip 48G|?|?|?|
|gzip 192G|?|?|?|
|zstd 48G|16G|4m|less than 2m|
|zstd 192G|136G|17m|4m27s|

This change is assisted by Claude with the zstd API.

TODOs:
- [ ] Support sparse restore

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -171,7 +171,8 @@ jobs:
                     pkg-config \
                     protobuf \
                     grpc \
-                    boost
+                    boost \
+                    zstd
 
 
             - name: Restore macOS build/ALL cache
--- a/SConstruct
+++ b/SConstruct
@@ -886,6 +886,13 @@ for variant_path in variant_paths:
                   'and/or zlib.h header file.\n'
                   'Please install zlib and try again.')
 
+        if not conf.CheckLibWithHeader(
+            'zstd', 'zstd.h', 'C++', 'ZSTD_versionNumber();'
+        ):
+            error('Did not find needed zstd compression library '
+                  'and/or zstd.h header file.\n'
+                  'Please install zstd and try again.')
+
     if not GetOption('without_tcmalloc'):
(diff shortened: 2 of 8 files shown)


powerjg:
Can you comment on whether #2794 makes any difference for you? I am not opposed to this change, but I want to make sure you're aware of #2794

hnpl:
I think #2794 and this change address different problems.

#2794 saves the host's memory based on the host's lazy allocation behavior. This does not address the problem where the working set is large/dense (e.g., ~150GiB).

This change is about what compression algorithm to use when dumping the guest memory to a file, which addresses the speed and the compression ratio of the checkpoint. I find that compression/decompression are the bottleneck of saving/restoring checkpoint so using zstd/raw actually saves a lot of time.

hnpl:
What do you think about introducing a new dependency on zstd?

erin-le:
@hnpl: Hello, a number of the CI tests are failing on this PR. Could you take a look? Thank you!

erin-le:
@hnpl: Hello, just wanted to ask if you were still planning to continue working on this PR. If you are, could you take a look at the Copilot comments and the failing CI tests? If not, would you be okay with closing it? Thank you!

erin-le:
This PR is being closed for inactivity. However, it can be reopened later if/when you have more time to continue working on it. Thanks for your understanding!

PR #3176, opened 2026-05-15
dev-virtio,configs-arm: Add 9P directory rootfs support

lionkov:
Adds an optional npfs/ufs-backed VirtIO 9P backend. The backend is enabled at build time by setting NPFS_PATH to an npfs source tree and serializes npfs state into gem5 checkpoints, allowing a 9P-mounted directory filesystem to work across checkpoint and restore.

npfs is available from:

    https://github.com/lionkov/npfs

Updates the ARM fs_bigLITTLE example with --rootfs-dir, --vio-9p-root, --vio-9p-tag, and --vio-9p-debug options. --rootfs-dir exports a host directory through VirtIO 9P, uses /dev/root as the default mount tag, and adds the kernel arguments required to mount the directory as the guest root filesystem.

Documents the optional npfs build flow and rootfs directory usage under src/dev/virtio.

### DIFF
--- a/configs/common/FSConfig.py
+++ b/configs/common/FSConfig.py
@@ -83,15 +83,46 @@ class MemBus(SystemXBar):
     default = Self.badaddr_responder.pio
 
 
-def attach_9p(parent):
+def attach_9p(parent, root=None, tag=None):
+    """Attach a VirtIO 9P device to parent.
+
+    Without a root directory, export a scratch directory under the
+    output directory through the diod server. This is the backend used
+    by --vio-9p and is always available.
+
+    With a root directory, export that host directory through the
+    npfs/ufs backend, which is only present when gem5 was built with
+    NPFS_PATH set. Unlike diod, it serializes its state into
+    checkpoints, so a 9P-mounted root filesystem survives a restore.
+
+    A tag of None leaves the device's default mount tag in place.
+    """
     viopci = PciVirtIO()
-    viopci.vio = VirtIO9PDiod()
-    viodir = os.path.realpath(os.path.join(m5.options.outdir, "9p"))
-    viopci.vio.root = os.path.join(viodir, "share")
-    viopci.vio.socketPath = os.path.join(viodir, "socket")
-    os.makedirs(viopci.vio.root, exist_ok=True)
-    if os.path.exists(viopci.vio.socketPath):
-        os.remove(viopci.vio.socketPath)
+
+    if root is None:
+        viodir = os.path.realpath(os.path.join(m5.options.outdir, "9p"))
+        viopci.vio = VirtIO9PDiod()
+        viopci.vio.root = os.path.join(viodir, "share")
+        viopci.vio.socketPath = os.path.join(viodir, "socket")
+        os.makedirs(viopci.vio.root, exist_ok=True)
+        if os.path.exists(viopci.vio.socketPath):
+            os.remove(viopci.vio.socketPath)
+    else:
+        # VirtIO9PUfs is only registered as a SimObject in npfs-enabled
+        # builds, so this check has to precede the first reference to it.
+        if not m5.defines.buildEnv.get("BUILD_9PUFS", False):
+            fatal(
+                "Exporting a host directory through VirtIO 9P requires the "
+                "npfs-backed device, which is not available in this gem5 "
+                "binary. Rebuild gem5 with NPFS_PATH set to the npfs source "
+                "directory."
+            )
+        viopci.vio = VirtIO9PUfs()
+        viopci.vio.root = os.path.realpath(root)
+
+    if tag is not None:
+        viopci.vio.tag = tag
+
     parent.viopci = viopci
     parent.attachPciDevice(viopci)
 
--- a/src/dev/virtio/fs9pufs.cc
+++ b/src/dev/virtio/fs9pufs.cc
@@ -1,0 +1,180 @@
+/*
+ * Copyright (c) 2014-2025 ARM Limited
+ * All rights reserved
+ *
+ * The license below extends only to copyright in the software and shall
+ * not be construed as granting a license to any other intellectual
+ * property including but not limited to intellectual property relating
+ * to a hardware implementation of the functionality of the software
+ * licensed hereunder.  You may use the software subject to the license
+ * terms below provided that you ensure that this notice is replicated
+ * unmodified and in its entirety in all distributions of the software,
+ * modified or unmodified, in source code or in binary form.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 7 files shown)


erin-le:
@giactra: Hello, would you be able to take a look at this PR when you get a chance? Thank you!

erin-le:
@lionkov: Hello, could you try rebasing this branch on current `develop` and see if the CI tests pass after the rebase? Thank you!

lionkov:
Rebased on current `develop` — the branch is now `bf7ae311c3` on top of `fe89a41723`. The rebase was clean, no conflicts, and GitHub now reports the PR as mergeable.

While rebasing I also fixed a regression this PR had introduced: `attach_9p` made plain `--vio-9p` fatal on any build without `NPFS_PATH`. It now selects the backend by whether a host directory was actually requested. `--vio-9p` on its own keeps using diod and its default mount tag, so builds without npfs behave exactly as before; only `--vio-9p-root` and `--rootfs-dir` need the npfs-backed device. `configs/common/Options.py` reverted to its upstream contents as a result, so the diff is down to 8 files.

Locally, `pre-commit` passes on all changed files and `git-clang-format` reports no changes against the merge base (clang-format 18, matching the ubuntu-24.04 CI image). pre-commit.ci is green on the new head.

The `CI Tests` workflow is sitting at `action_required` and hasn't run — it looks like the same was true of the previous head, so I don't think it has ever run on this PR. Could you approve the workflow run? I don't have permission to do that myself. Thanks!

src/dev/virtio/fs9pufs.cc
```diff
@@ -0,0 +1,169 @@
...
+    : VirtIO9PProxy(params), fd_to_ufs(-1), fd_from_ufs(-1), srv(nullptr)
+{}
+
+VirtIO9PUfs::~VirtIO9PUfs()
+{}
+
+void
+VirtIO9PUfs::startup()
+{
+    inform("VirtIO9PUfs: startup");
```
powerjg: Can we replace `inform` with `DPRINTF`? I would prefer to not spam more messages to the user unless they are necessary.

src/dev/virtio/fs9pufs.hh
```diff
@@ -0,0 +1,112 @@
...
+#include <map>
+#include <memory>
+#include <string>
+
+extern "C" {
+#include <npfs.h>
+#include <ufs.h>
+}
+
+// #include "base/compiler.hh"
```
powerjg: please remove commented out lines if they are not needed.

src/dev/virtio/VirtIO9P.py
```diff
@@ -74,3 +74,12 @@ class VirtIO9PSocket(VirtIO9PProxy):
...
     port = Param.String("564", "9P server port")
+
+
+class VirtIO9PUfs(VirtIO9PProxy):
+    type = "VirtIO9PUfs"
+    cxx_header = "dev/virtio/fs9pufs.hh"
+    cxx_class = "gem5::VirtIO9PUfs"
+
+    root = Param.String("Host directory to export through npfs/ufs")
+    debug = Param.Bool(False, "Enable verbose npfs/ufs output")
```
powerjg: If feasible, please use `DPRINTF` and `DebugFlags` instead of other parameters in your sim objects.

lionkov:
All three addressed:

- `fs9pufs.cc`: both `inform` calls are now `DPRINTF(VIO9P, ...)`.
- `fs9pufs.hh`: commented-out includes removed.
- `VirtIO9P.py`: the `debug` parameter is replaced by a `VIO9PUfs` debug flag, registered only in npfs-enabled builds. This drops `--vio-9p-debug`; `--debug-flags=VIO9PUfs` replaces it.

On the earlier CI run, 30 of 32 checks passed. Both failures were download flakes rather than this change: `x86_boot_tests` hit a truncated `x86-ubuntu-24.04-img` download (`ContentTooShortError`), and `multisim` failed in "Download gem5 artifacts" before any test ran. This push should start a fresh run.

configs/common/FSConfig.py
```diff
@@ -83,15 +83,46 @@ class MemBus(SystemXBar):
     default = Self.badaddr_responder.pio
 
 
-def attach_9p(parent):
+def attach_9p(parent, root=None, tag=None):
```
powerjg: No action needed, just FYI: This file is basically deprecated. We are moving everything to the stdlib. This would go on a "board" in the stdlib.

src/dev/virtio/fs9pufs.cc
```diff
@@ -0,0 +1,170 @@
...
+VirtIO9PUfs::startup()
+{
+    DPRINTF(VIO9P, "VirtIO9PUfs: startup\n");
+    start();
+    dataEvent.reset(new UfsDataEvent(*this, fd_from_ufs, POLLIN));
+    pollQueue.schedule(dataEvent.get());
+}
+
+void
+VirtIO9PUfs::start()
```
powerjg: I think this code should just be in `startup`.

See "Detailed Description" on https://doxygen.gem5.org/release/current/classgem5_1_1SimObject.html

If you want something called after initialization and after loading a checkpoint, you put it in `startup`. Right now, this code is executed twice when loading a checkpoint (did you test loading a checkpoint?)

src/dev/virtio/fs9pufs.cc
```diff
@@ -0,0 +1,170 @@
...
+}
+
+void
+VirtIO9PUfs::unserialize(CheckpointIn &cp)
+{
+    int datalen;
+    char *data;
+    char err[256];
+
+    start();
```
powerjg: This will be a called twice if you call the function here and in `startup`

src/dev/virtio/fs9pufs.cc
```diff
@@ -0,0 +1,170 @@
...
+#include "dev/virtio/fs9pufs.hh"
+
+namespace gem5
+{
+
+VirtIO9PUfs::VirtIO9PUfs(const Params &params)
+    : VirtIO9PProxy(params), fd_to_ufs(-1), fd_from_ufs(-1), srv(nullptr)
+{}
+
+VirtIO9PUfs::~VirtIO9PUfs()
```
powerjg: Delete empty destructor.

src/dev/virtio/fs9pufs.hh
```diff
@@ -0,0 +1,109 @@
...
+
+  protected:
+    /**
+     * Start the embedded ufs server and setup the communication pipes.
+     */
+    void start();
+
+    ssize_t read(uint8_t *data, size_t len);
+    ssize_t write(const uint8_t *data, size_t len);
+    void terminate();
```
powerjg: Docstrings would be good

PR #3181, opened 2026-05-19
sim: runtime debug flags toggle via SIGRTMIN+1

polpetras:
Follows up on discussion #1598.

## Summary

This patch adds the ability to enable and disable gem5 debug flags in a
running simulation without restarting it.

gem5 installs a handler on `SIGRTMIN+1` at simulation start. A companion
shell script (`util/gem5_debug_ctl.sh`) writes commands to a temporary file
and sends the signal. Commands are processed on the main event loop at the
next sim tick after delivery — no mid-instruction races.

Supported commands:

| Command        | Effect                                          |
|----------------|-------------------------------------------------|
| `+FLAG`        | Enable flag                                     |
| `+FLAG:Nt`     | Enable flag, auto-disable after N ticks         |
| `-FLAG`        | Disable flag                                    |
| `trace:on`     | Global output enable (individual flags unchanged)  |
| `trace:off`    | Global output silence (individual flags unchanged) |

Example usage:

```bash
# Enable Rename for 100000 ticks on a running sim (PID e.g. 12345)
util/gem5_debug_ctl.sh <PID> +Rename:100000t

# Enable multiple flags at once for a bounded window, with output enabled
util/gem5_debug_ctl.sh <PID> +Fetch:100000t +Rename:100000t +IEW:100000t trace:on

# Disable Fetch and silence all output
util/gem5_debug_ctl.sh <PID> -Fetch trace:off
```

Confirmation lines (`gem5_debug_ctl: ...`) appear in the sim's stdout/log.

## Design notes

- `SIGRTMIN+1` is used to avoid `SIGRTMIN` which is reserved for KVM
  (`KVM_KICK_SIGNAL`).
- The signal handler follows the existing gem5 async pattern: sets
  `async_event` + `async_debug_cmd`, wakes the primary event queue, returns
  immediately. All file I/O and flag changes happen on the main thread.
- `Flag::_tracing` is changed from `bool` to `std::atomic<bool>` (relaxed
  ordering) to eliminate the data race that arises when `changeFlag()` is
  called while other host threads are executing `DPRINTF`.
- Auto-disable is implemented as a self-deleting `Event` (`AutoDelete`)
  scheduled at `curTick() + duration`.

---

*This implementation was developed with AI assistance.*

### DIFF
--- a/src/base/debug.hh
+++ b/src/base/debug.hh
@@ -42,6 +42,7 @@
 #ifndef __BASE_DEBUG_HH__
 #define __BASE_DEBUG_HH__
 
+#include <atomic>
 #include <initializer_list>
 #include <iostream>
 #include <map>
@@ -63,7 +64,10 @@ class Flag
   protected:
     static bool _globalEnable; // whether debug tracings are enabled
 
-    bool _tracing = false; // tracing is enabled and flag is on
+    // Read by every DPRINTF on every thread; atomic to avoid data races
+    // when changeFlag() is called at runtime. Relaxed ordering suffices:
+    // a stale read causes at most one missed or extra debug line.
+    std::atomic<bool> _tracing{false};
 
     const char *_name;
     const char *_desc;
@@ -77,7 +81,11 @@ class Flag
     std::string name() const { return _name; }
     std::string desc() const { return _desc; }
 
-    bool tracing() const { return TRACING_ON && _tracing; }
+    bool
+    tracing() const
+    {
+        return TRACING_ON && _tracing.load(std::memory_order_relaxed);
+    }
 
     virtual void enable() = 0;
     virtual void disable() = 0;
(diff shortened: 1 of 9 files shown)


Harshil2107:
Hi @polpetras, thank you for the contribution!

I just wanted to note that toggling debug flags at runtime is already supported via the `hypercall_external_signal` utility (#2161). Not all features in this PR have equivalents there, but I think it would be better to have a single place for this kind of functionality rather than two parallel approaches. Would you consider adding the missing features (like auto-disable after N ticks) to the existing utility instead?

polpetras:
Hi @Harshil2107 , and thank you for your feedback. Yes I can work towards adding the extra functionality in `hypercall_external_signal ` utility.

polpetras:
Superseded by #3259, which consolidates the runtime debug control into the existing `hypercall_external_signal` orchestrator infrastructure per @Harshil2107's review. The `Flag::_tracing` thread-safety concern raised
in this thread is a separate pre-existing C++ issue and will be addressed in a follow-up PR.

PR #3184, opened 2026-05-19
arch-vega: Implement several instructions for Triton

abmerop:
Adding a bunch of instructions that triton likes to use. We now have advanced instructions like the ability to add and subtract signed numbers. Tested with single instruction tests found here: https://github.com/abmerop/gem5-resources/tree/inst-tests/src/gpu/instruction-tests.

### DIFF
--- a/src/arch/amdgpu/vega/insts/vop1.cc
+++ b/src/arch/amdgpu/vega/insts/vop1.cc
@@ -441,25 +441,18 @@ Inst_VOP1__V_CVT_F16_F32::~Inst_VOP1__V_CVT_F16_F32()
 void
 Inst_VOP1__V_CVT_F16_F32::execute(GPUDynInstPtr gpuDynInst)
 {
-    Wavefront *wf = gpuDynInst->wavefront();
-    ConstVecOperandF32 src(gpuDynInst, instData.SRC0);
-    VecOperandU32 vdst(gpuDynInst, instData.VDST);
-
-    src.readSrc();
-
-    panic_if(isSDWAInst(), "SDWA not implemented for %s", _opcode);
-    panic_if(isDPPInst(), "DPP not implemented for %s", _opcode);
-
-    for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
-        if (wf->execMask(lane)) {
-            float tmp = src[lane];
-            AMDGPU::mxfloat16 out(tmp);
-
-            vdst[lane] = (out.data >> 16);
+    auto opImpl = [](VecOperandU32 &src, VecOperandU32 &vdst, Wavefront *wf) {
+        for (int lane = 0; lane < NumVecElemPerVecReg; ++lane) {
+            if (wf->execMask(lane)) {
+                uint32_t raw_src = src[lane];
+                float tmp = *reinterpret_cast<float *>(&raw_src);
+                AMDGPU::mxfloat16 out(tmp);
+                vdst[lane] = (out.data >> 16);
+            }
         }
-    }
+    };
 
-    vdst.write();
+    vop1Helper<ConstVecOperandU32, VecOperandU32>(gpuDynInst, opImpl);
 } // execute
 // --- Inst_VOP1__V_CVT_F32_F16 class methods ---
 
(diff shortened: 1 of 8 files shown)


src/arch/amdgpu/vega/insts/vop1.cc
```diff
@@ -441,25 +441,18 @@ Inst_VOP1__V_CVT_F16_F32::~Inst_VOP1__V_CVT_F16_F32()
 void
 Inst_VOP1__V_CVT_F16_F32::execute(GPUDynInstPtr gpuDynInst)
 {
-    Wavefront *wf = gpuDynInst->wavefront();
-    ConstVecOperandF32 src(gpuDynInst, instData.SRC0);
-    VecOperandU32 vdst(gpuDynInst, instData.VDST);
-
-    src.readSrc();
-
-    panic_if(isSDWAInst(), "SDWA not implemented for %s", _opcode);
```
mattsinc, line 450: why was this removed?  It doesn't seem like the updated code handles it?  Unless vop1Helper does?

src/arch/amdgpu/vega/insts/vop3p.cc
```diff
@@ -728,10 +728,10 @@ Inst_VOP3P__V_PK_FMA_F32::execute(GPUDynInstPtr gpuDynInst)
             if (neg & 1) {
                 s0lf = -s0lf;
             }
-            if (neg & 1) {
+            if (neg & 2) {
```
mattsinc, line 731: was this a bug previously?  If so should it be a separate patch?

src/arch/amdgpu/vega/insts/vop3p.cc
```diff
@@ -1033,5 +1033,220 @@ Inst_VOP3P__V_MFMA_LOAD_SCALE::execute(GPUDynInstPtr gpuDynInst)
...
+
+    src0.readSrc();
+    src1.readSrc();
+    src2.readSrc();
+
+    int opsel = instData.OPSEL;
+    int opsel_hi = extData.OPSEL_HI | (instData.OPSEL_HI2 << 2);
+    int neg_hi = instData.NEG_HI;
+
+    ArmISA::FPSCR fpscr;
```
mattsinc, line 1062: Should this be AMDIsa instead of ArmISA?

mattsinc (approved):
A few questions but I'm sure it's fine

src/arch/amdgpu/vega/insts/vop1.cc
abmerop: Yep, the new vop1Helper from #3110 will handle SDWA and DPP implicitly so we only need to define the actually instruction logic once

src/arch/amdgpu/vega/insts/vop3p.cc
abmerop: Yea, it is a previous bug. I lumped it into this PR because it's a small change. I could create a new PR if you prefer

src/arch/amdgpu/vega/insts/vop3p.cc
abmerop: No, we are intentionally using ARM's soft float library for FP16 because there is currently no native type. C++ does not support fixed-width std::float16_t until C++23 and _Float16 is not available until gcc 12+/clang 15+ (I guess we could be using that now).

We need to use ARM soft float / _Float16 / std::float16_t for all F16 arithmetic instructions. Pointing this out because the MXFP types in gem5 (mxfloat16) are not to be used for arithmetic. The only instructions using those types do implicit conversion to float, perform the operation, and convert back which causes precision loss for some arithmetic operations

src/arch/amdgpu/vega/insts/vop3p.cc
mattsinc: Certainly it would be cleaner to have a separate patch, but I'll defer to you here

Basemism:
Thanks for the work on this, this is really useful for me.
Quick follow-up: is BF16 instruction support on the roadmap too (BF16 MFMA/SMFMAC), or would that likely be a separate implementation effort?

abmerop:
> Thanks for the work on this, this is really useful for me. Quick follow-up: is BF16 instruction support on the roadmap too (BF16 MFMA/SMFMAC), or would that likely be a separate implementation effort?

I have those implemented somewhere. I can post them soon.

Basemism:
That would be great, thanks! Very keen on using those instructions.

PR #3186, opened 2026-05-19
stdlib: Use VIPERSequencer for GPU SQC/Scalar cache

abmerop:
The VIPERSequencer handles special reads which are not associated with an address. For example, s_memtime will issue a read request to SQC with no address set. These caches therefore need to use the special VIPER sequencer instead of the standard sequencer.


mattsinc (approved):
Good catch, shocked we didn't see this issue previously :)

powerjg (approved):

PR #3191, opened 2026-05-21
cpu-o3: O3 CPU inter-stage unblock signal causes 1-cycle bubble due to propagation delay

### LINKED ISSUE #3190: BUG: O3 CPU inter-stage unblock signal causes 1-cycle bubble due to propagation delay
**Describe the bug**

In the gem5 O3 CPU pipeline, when a downstream stage (e.g., decode, rename, IEW) blocks its upstream stage, it only sends the unblock signal after its skid buffer is **completely empty**. Due to the signal propagation delay between stages (e.g., `decodeToFetchDelay = 1`), this causes an unnecessary 1-cycle bubble before the upstream stage can resume delivering instructions.

Example timeline (fetch → decode, with `decodeToFetchDelay = 1` and `fetchToDecodeDelay = 1`):

| Cycle | Event |
|-------|-------|
| T     | decode's skid buffer becomes empty, sends `decodeUnblock` |
| T+1   | fetch receives unblock, resumes sending instructions; **decode is idle (bubble)** |
| T+2   | decode receives the new instructions and resumes processing |

This 1-cycle bubble occurs on every block→unblock transition across all adjacent stage pairs: decode→fetch, rename→decode, and IEW→rename.

**To Reproduce**

1. Compile gem5:

```shell
scons build/RISCV/gem5.opt -j$(nproc)
```

2. Run a workload with O3CPU and debug tracing:

```shell
build/RISCV/gem5.opt --debug-flags=Decode,Fetch \
  --debug-file=debug_trace.txt \
  configs/deprecated/example/se.py \
  --cpu-type=O3CPU --caches \
  --cmd=tests/test-progs/hello/bin/riscv/linux/hello
```

3. In `m5out/debug_trace.txt`, search for the pattern where decode prints `Done unblocking` followed by `Nothing to do, breaking out early` on the next cycle.

**Terminal Output**

```shell
4228500: system.cpu.decode: [tid:0] Done unblocking.
4229000: system.cpu.fetch: [tid:0] [sn:3179] Sending instruction to decode from fetch queue. Fetch queue size: 23.
4229000: system.cpu.decode: [tid:0] Not blocked, so attempting to run stage.
4229000: system.cpu.decode: [tid:0] Nothing to do, breaking out early.
4229500: system.cpu.decode: [tid:0] Sending instruction to rename.
4229500: system.cpu.decode: [tid:0] Processing instruction [sn:3179] with PC (0x1c244=>0x1c246).(0=>1)
```

At cycle 4229000, decode has transitioned to Running but has no instructions to process — fetch sent instructions at the same cycle but they won't arrive until the next cycle due to `fetchToDecodeDelay`. This is the 1-cycle bubble.

**Expected behavior**

The downstream stage should send the unblock signal **early** — before its skid buffer is fully empty — to compensate for the signal propagation delay. Specifically, unblock should be sent when `skidBuffer.size() <= stageDelay × stageWidth`, so the upstream stage resumes in time for the downstream stage to have instructions immediately when the skid buffer drains.

**Affects version**

gem5 develop branch (commit `679832b342`)

**Host Operating System**

macOS (Apple Silicon)

**Host ISA**

RISC-V (simulated)

**Compiler used**

Apple Clang 15.0 (Xcode)

zhongchengyong:
Fixes #3190 

When a downstream stage (decode/rename/IEW) blocks the upstream stage, it previously only sent the unblock signal after its skid buffer was completely drained. Due to signal propagation delay (e.g., decodeToFetchDelay), this caused a 1-cycle bubble before the upstream stage could resume sending instructions.

Fix: send unblock early when skidBuffer.size() <= stageDelay * stageWidth, i.e., when the remaining entries can be consumed during the signal propagation time. The stage only transitions to Running state when the skid buffer is truly empty, maintaining correctness.

Applied uniformly to decode->fetch, rename->decode, and IEW->rename paths. Remove stale assertions that assumed unblock is only received once.

### DIFF
--- a/src/cpu/o3/BaseO3CPU.py
+++ b/src/cpu/o3/BaseO3CPU.py
@@ -100,6 +100,16 @@ class BaseO3CPU(BaseCPU):
     fetchToDecodeDelay = Param.Cycles(1, "Fetch to decode delay")
     decodeWidth = Param.Unsigned(8, "Decode width")
 
+    earlyUnblock = Param.Bool(
+        True,
+        "Signal an upstream stage to unblock as soon as the entries left in "
+        "the skid buffer can keep this stage busy for the block/unblock "
+        "signal round trip, instead of waiting for the skid buffer to drain "
+        "completely. Waiting for a fully drained skid buffer leaves a bubble "
+        "as wide as the round trip. Set to False to restore the legacy "
+        "behaviour.",
+    )
+
     iewToRenameDelay = Param.Cycles(
         1, "Issue/Execute/Writeback to rename delay"
     )
--- a/src/cpu/o3/decode.cc
+++ b/src/cpu/o3/decode.cc
@@ -97,6 +97,7 @@ Decode::Decode(CPU *_cpu, const BaseO3CPUParams &params)
       iewToDecodeDelay(params.iewToDecodeDelay),
       commitToDecodeDelay(params.commitToDecodeDelay),
       fetchToDecodeDelay(params.fetchToDecodeDelay),
+      decodeToFetchDelay(params.decodeToFetchDelay),
       decodeWidth(params.decodeWidth),
       numThreads(params.numThreads),
       stats(_cpu)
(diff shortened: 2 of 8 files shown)


erin-le:
@zhongchengyong: Hello, could you investigate the CI test failures on this PR when you get a chance? Thank you!

sebastian-kim:
Hi,
    I was checking this PR, and I think that while it fixes the Fetch-Decode delay, it will not work for the Decode-Rename stages, because the threshold (skidBuffer.size() <= renameToDecodeDelay * renameWidth) is not considering how much can be actually renamed in that cycle.

In other words, I think that even if rename can process (for example) 8 instructions, if there are only 3 free ROB entries, it will send the unblock signal to decode too early, overflowing the skidbuffer.

zhongchengyong:
> Hi, I was checking this PR, and I think that while it fixes the Fetch-Decode delay, it will not work for the Decode-Rename stages, because the threshold (skidBuffer.size() <= renameToDecodeDelay * renameWidth) is not considering how much can be actually renamed in that cycle.
> 
> In other words, I think that even if rename can process (for example) 8 instructions, if there are only 3 free ROB entries, it will send the unblock signal to decode too early, overflowing the skidbuffer.

Got it.  I'll try to find a proper solution.  
On the other hand, maybe the xxx_stage.tick() called in a reversed order just like MinorCPU will be a better way.

sebastian-kim:
> > Hi, I was checking this PR, and I think that while it fixes the Fetch-Decode delay, it will not work for the Decode-Rename stages, because the threshold (skidBuffer.size() <= renameToDecodeDelay * renameWidth) is not considering how much can be actually renamed in that cycle.
> > In other words, I think that even if rename can process (for example) 8 instructions, if there are only 3 free ROB entries, it will send the unblock signal to decode too early, overflowing the skidbuffer.
> 
> Got it. I'll try to find a proper solution. On the other hand, maybe the xxx_stage.tick() called in a reversed order just like MinorCPU will be a better way.

Yes. I think this could also be simplified by using shared inter-stage queues: a the fetch queue between fetch and decode, and a decode/rename queue (decoded queue or allocation queue). Rename would consume as many instructions as its available resources allow, while decode would only produce instructions when there is sufficient space in the allocation queue. Backpressure would then follow directly from queue occupancy, avoiding the need to predict when an unblock signal should be sent.

erin-le:
@zhongchengyong: Hello, could you give an update on the status of this PR on your end? Thank you!

zhongchengyong:
> @zhongchengyong: Hello, could you give an update on the status of this PR on your end? Thank you!

I’m trying to implement the new solution.

PR #3199, opened 2026-06-01
base, mem: Add new Request::DEBUG to GDB read/write request

rogerchang23424:
The Request::DEBUG explicitly indicates that request is from debug model (GDB, etc). Which is useful to bypass access check and probe data in device model


rogerchang23424:
Not required now

PR #3200, opened 2026-06-01
mem-ruby: more readable NetDest print

jonathanhaley:
Prints only the masks for machines with a destination set and explicitly prints the machine names. This makes debug traces smaller and easier to read.

Change-Id: I9cf1ee9616f917679090cfabaa355e4a8e2d05e8


abmerop (approved):
This will be super useful for large topologies

PR #3202, opened 2026-06-02
mem-ruby: Add Cbusy handling logic to the TlmGenerator

giactra:
This PR adds CBusy handling logic to the TLMGenerator

1) Incoming values are recorded in the cbusy tracker child
2) Incoming values are recorded as cbusy10 and cbusy2 stats

The PR also adds a CHI-TLM <-> ruby srcId converstion logic

### DIFF
--- a/src/mem/ruby/protocol/chi/tlm/controller.hh
+++ b/src/mem/ruby/protocol/chi/tlm/controller.hh
@@ -92,6 +92,9 @@ namespace tlm::chi {
 class CacheController : public ruby::CHIGenericController
 {
   public:
+    /** We mainly use this when we flatten MachineID into src_id/tgt_id */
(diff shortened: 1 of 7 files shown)


src/mem/ruby/protocol/chi/tlm/controller.hh
```diff
@@ -92,6 +92,9 @@ namespace tlm::chi {
 class CacheController : public ruby::CHIGenericController
 {
   public:
+    /** We mainly use this when we flatten MachineID into src_id/tgt_id */
+    static constexpr uint64_t MAX_NODES = 1024;
```
powerjg, line 96: Is this related to `NUM_BITS_PER_SET`? Should we use that number instead here?

powerjg (approved):
One optional comment below.

PR #3203, opened 2026-06-02
mem-cache: Fix cache block movement logic

### LINKED ISSUE #2955: Mismatched tag after moving Cache Blocks
**Describe the bug**
The move assignment operator of `CacheBlk` behaves in invalid ways. Specifically, the call to `CacheBlk::insert` results in a modified tag that does not match with the original tag of the other CacheBlk instance.

https://github.com/gem5/gem5/blob/7a2b0e413d06c5ce7097104abef3b1d9eaabca91/src/mem/cache/cache_blk.hh#L172-L196

The cause of this is the addition of a call to `extractTag` in `TaggedEntry::insert`. https://github.com/gem5/gem5/blob/7a2b0e413d06c5ce7097104abef3b1d9eaabca91/src/mem/cache/tags/tagged_entry.hh#L175

While previously, the insert function only took the tag as a parameter, starting from ee9814499d492da607e12559cb52980b19a7ec39 the entire address is expected. This change is not mirrored in the move assignment operation (where the full address is also not available as far as I can see). When `TaggedEntry::insert` is called with only the tag, extractTag is called on the tag instead of the address. This likely performs shifts on the tag, leading to truncation and an invalid tag in the target block.

**Affects version**
Starting from [Version 24.1.0.0](https://github.com/gem5/gem5/releases/tag/v24.1.0.0),  ee9814499d492da607e12559cb52980b19a7ec39

**gem5 Modifications**
None necessary.

**To Reproduce**
Any use of the move assignment operator of `CacheBlk`, for example through `BaseTags::moveBlock` (or `BaseSetAssoc::moveBlock`), will trigger this issue. While the mismatched tag might not become apparent immediately, an eviction will trigger failed assertions in the snoop filter (and an inconsistent state of the cache in general).

https://github.com/gem5/gem5/blob/7a2b0e413d06c5ce7097104abef3b1d9eaabca91/src/mem/cache/tags/base.cc#L137 

**Expected behavior**
Using the move assignment operation of `CacheBlk` should copy the metadata of the source block to the target block. The tag of the target block should be identical to the tag of the source block from before.

**Host Operating System**
Any

**Host ISA**
Any

**Compiler used**
gcc version 15.2.1 20260103 (GCC)

**Additional information**
I'm not sure what the best way to fix this would be. The full address is not available to the `CacheBlk` during the move operation and we don't have any way to call `regenerateAddr` from this point on. It's also not possible to just shift the tag, as the structure of the tag and specific implementation of `extractTag` depend on the implementation of the indexing policy.

xcf-t:
Moving metadata between classic cache blocks currently leads to inconsistent tags and coherency issues (gem5/gem5#2955).

To fix that, a dedicated move assignment operator is added to the `TaggedEntry` class so that `insert` is no longer needed for move operations. Subclass move operations are updated to reflect this change and its side-effects.

Fixes gem5/gem5#2955.

### DIFF
--- a/src/mem/cache/cache_blk.hh
+++ b/src/mem/cache/cache_blk.hh
@@ -168,6 +168,15 @@ class CacheBlk : public TaggedEntry
      * will become invalid, and the invalid, valid. All location related
      * variables will remain the same, that is, an entry cannot move its
      * data, just its metadata contents.
+     *
+     * This implementation has to be virtual, because move assignments are
+     * generally performed on the CacheBlk object. If the CacheBlk is backed
+     * by an inheriting class like CompressionBlk, we need the inheritor to
+     * handle the move operation, even if it's called on a CacheBlk type.
+     *
+     * @param other A valid CacheBlk whose metadata should be moved into the
+     *              current entry
+     * @return *this
      */
     virtual CacheBlk&
     operator=(CacheBlk&& other)
@@ -177,10 +186,10 @@ class CacheBlk : public TaggedEntry
         assert(!isValid());
         assert(other.isValid());
 
-        insert({other.getTag(), other.isSecure()});
-
         if (other.wasPrefetched()) {
             setPrefetched();
+        } else {
+            clearPrefetched();
         }
         setCoherenceBits(other.coherence);
         setTaskId(other.getTaskId());
@@ -190,7 +199,13 @@ class CacheBlk : public TaggedEntry
         setSrcRequestorId(other.getSrcRequestorId());
         std::swap(lockList, other.lockList);
 
-        other.invalidate();
+        /*
+         * We assume that the TaggedEntry move assignment operator invalidates
+         * the source block. We cannot call invalidate again, as this would
+         * cause side effects on some CacheBlk implementations.
+         */
+        TaggedEntry::operator=(std::move(other));
+        assert(!other.isValid());
 
         return *this;
     }
@@ -532,8 +547,55 @@ class TempCacheBlk final : public CacheBlk
         registerTagExtractor(ext);
     }
     TempCacheBlk(const TempCacheBlk&) = delete;
-    using CacheBlk::operator=;
+    TempCacheBlk(TempCacheBlk &&) = delete;
     TempCacheBlk& operator=(const TempCacheBlk&) = delete;
+    /**
+     * Move assignment operator.
+     * This should only be used to move an existing valid entry into an
+     * invalid one, not to create a new entry. In the end the valid entry
+     * will become invalid, and the invalid, valid. All location related
+     * variables will remain the same, that is, an entry cannot move its
+     * data, just its metadata contents. Because of this, we do not interact
+     * with any stored ReplaceableEntry data.
+     *
+     * This operator must only be used with other instances of TempCacheBlk. A
+     * similar operation cannot be done using any other CacheBlk instance,
+     * because the tempBlock's address needs to be stored.
+     *
+     * @param other A valid TempCacheBlk whose metadata should be moved into
+     *              the current entry
+     * @return *this
+     */
+    TempCacheBlk &
+    operator=(TempCacheBlk &&other)
+    {
+        // Copying an entry into a valid one would imply in skipping all
+        // replacement steps, so it cannot be allowed
+        assert(!isValid());
+        assert(other.isValid());
+
+        _addr = other._addr;
+
+        CacheBlk::operator=(std::move(other));
+        assert(!other.isValid());
+
+        return *this;
+    }
+
+    CacheBlk &
+    operator=(CacheBlk &&other) override
+    {
+        auto *temp_other = dynamic_cast<TempCacheBlk *>(&other);
+
+        if (!temp_other) {
+            panic("A CacheBlk instance must not be moved into a TempCacheBlk, "
+                  "because we cannot restore the address.");
+        }
+
+        TempCacheBlk::operator=(std::move(*temp_other));
+        return *this;
+    }
+
     ~TempCacheBlk() { delete [] data; };
 
     /**
--- a/src/mem/cache/tags/sector_blk.hh
+++ b/src/mem/cache/tags/sector_blk.hh
@@ -63,8 +63,7 @@ class SectorSubBlk : public CacheBlk
 
   public:
     SectorSubBlk() : CacheBlk(), _sectorBlk(nullptr), _sectorOffset(0) {}
-    SectorSubBlk(const SectorSubBlk&) = delete;
-    using CacheBlk::operator=;
+    SectorSubBlk(const SectorSubBlk &) = delete;
     SectorSubBlk& operator=(const SectorSubBlk&) = delete;
     SectorSubBlk(SectorSubBlk&&) = delete;
     /**
@@ -72,10 +71,13 @@ class SectorSubBlk : public CacheBlk
      * This should only be used to move an existing valid entry into an
      * invalid one, not to create a new entry. In the end the valid entry
      * will become invalid, and the invalid, valid. All location related
-     * variables will remain the same, that is, an entry cannot change
-     * its sector block nor its offset.
+     * variables will remain the same, that is, an entry cannot move its
+     * data, just its metadata contents. As SectorSubBlk only holds additional
+     * location data, we can use the unmodified CacheBlk move assignment
+     * operator.
      */
-    SectorSubBlk& operator=(SectorSubBlk&& other) = default;
+    using CacheBlk::operator=;
+    SectorSubBlk &operator=(SectorSubBlk &&other);
     ~SectorSubBlk() = default;
 
     /**
@@ -144,6 +146,8 @@ class SectorBlk : public TaggedEntry
   public:
     SectorBlk();
     SectorBlk(const SectorBlk&) = delete;
+    SectorBlk(SectorBlk &&) = delete;
+    SectorBlk &operator=(SectorBlk &&) = delete;
     SectorBlk& operator=(const SectorBlk&) = delete;
     ~SectorBlk() {};
 
--- a/src/mem/cache/tags/super_blk.hh
+++ b/src/mem/cache/tags/super_blk.hh
@@ -183,7 +183,9 @@ class SuperBlk : public SectorBlk
   public:
     SuperBlk();
     SuperBlk(const SuperBlk&) = delete;
+    SuperBlk(SuperBlk &&) = delete;
     SuperBlk& operator=(const SuperBlk&) = delete;
+    SuperBlk &operator=(SuperBlk &&) = delete;
     ~SuperBlk() {};
 
     /**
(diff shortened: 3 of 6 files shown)


erin-le:
@xcf-t: Hello, just wanted to ask about the status of this PR. If it's ready for review, or if you're looking for comments and feedback, could you take it out of draft? Thanks!

erin-le:
@xcf-t: Hello, just wanted to ask about this PR again. Could you give a brief update on the status of this on your end? If you aren't planning to work on it further, would you be okay with closing it, and potentially reopening at a later date? Thank you!

xcf-t:
@erin-le: Sorry for the late reply; GitHub didn't send me any notifications for this. The PR is in a working state and fixes the mentioned issue. I'd appreciate some feedback on the structure, as C++ inheritance is not exactly my strong suit.

src/mem/cache/tags/sector_blk.hh
```diff
@@ -63,20 +63,33 @@ class SectorSubBlk : public CacheBlk
...
+     * variables will remain the same, that is, an entry cannot move its
+     * data, just its metadata contents. As SectorSubBlk only holds additional
+     * location data, we can use the unmodified CacheBlk move assignment
+     * operator.
      */
-    SectorSubBlk& operator=(SectorSubBlk&& other) = default;
-    ~SectorSubBlk() = default;
+    using CacheBlk::operator=;
+
+    SectorSubBlk &
```
odanrc: Please move the implementation to the .cc file.

src/mem/cache/tags/sector_blk.hh
```diff
@@ -63,20 +63,33 @@ class SectorSubBlk : public CacheBlk
...
+    SectorSubBlk &operator=(const SectorSubBlk &) = delete;
+
+    /*
      * Move assignment operator.
      * This should only be used to move an existing valid entry into an
      * invalid one, not to create a new entry. In the end the valid entry
      * will become invalid, and the invalid, valid. All location related
-     * variables will remain the same, that is, an entry cannot change
-     * its sector block nor its offset.
+     * variables will remain the same, that is, an entry cannot move its
```
odanrc, line 74: The part of the documentation that was added is an implementation detail. Could you either move it to the implementation itself, as a regular comment, instead of adding to the function documentation, or remove it?

src/mem/cache/tags/sector_blk.hh
```diff
@@ -63,20 +63,33 @@ class SectorSubBlk : public CacheBlk
 
   public:
     SectorSubBlk() : CacheBlk(), _sectorBlk(nullptr), _sectorOffset(0) {}
-    SectorSubBlk(const SectorSubBlk&) = delete;
-    using CacheBlk::operator=;
-    SectorSubBlk& operator=(const SectorSubBlk&) = delete;
+
+    SectorSubBlk(const SectorSubBlk &) = delete;
```
odanrc, line 66: Hint: whenever making changes to repos with many users, avoid making unrelated/unnecessary changes, as every line changed risks generating a merge conflict to multiple people.

src/mem/cache/tags/super_blk.hh
```diff
@@ -82,18 +82,65 @@ class CompressionBlk : public SectorSubBlk
...
+     *
+     * @param other A valid CacheBlk whose metadata should be moved into the
+     * current entry. THIS MUST BE AN INSTANCE OF CompressionBlk! While mixing
+     * CompressionBlk with other types of CacheBlk should generally not occur,
+     * this would lead to soundness issues
+     * @return *this
      */
-    CompressionBlk& operator=(CompressionBlk&& other);
-    CacheBlk& operator=(CacheBlk&& other) override;
+    CompressionBlk &
```
odanrc: Could you move the implementation back to the implementation file?

src/mem/cache/tags/tagged_entry.hh
```diff
@@ -119,6 +119,46 @@ class TaggedEntry : public ReplaceableEntry
...
+     * variables will remain the same, that is, an entry cannot move its
+     * data, just its metadata contents. Because of this, we do not interact
+     * with any stored ReplaceableEntry data.
+     *
+     * @param other A valid TaggedEntry whose metadata should be moved into the
+     *              current entry
+     * @return *this
+     */
+    TaggedEntry &
+    operator=(TaggedEntry &&other) noexcept
```
odanrc: The functions called by this function are not tagged as noexcept. Since it is out of scope of this PR to mark the other functions as noexcept, it is better to remove the noexcept tag from this move operator, and leave adding noexcept to a future PR.

src/mem/cache/cache_blk.hh
```diff
@@ -159,41 +159,28 @@ class CacheBlk : public TaggedEntry
...
+     * This implementation has to be virtual, because move assignments are
+     * generally performed on the CacheBlk object. If the CacheBlk is backed
+     * by an inheriting class like CompressionBlk, we need the inheritor to
+     * handle the move operation, even if it's called on a CacheBlk type.
+     *
+     * @param other A valid CacheBlk whose metadata should be moved into the
+     *              current entry
+     * @return *this
      */
-    virtual CacheBlk&
```
odanrc, line 172: Hint: whenever making changes, avoid moving code around, as this complicates review. To move code I recommend not doing it in the same PR as a change that modifies it, but if that needs to be done, it should be done in its own isolate commit, to make it easier to isolate the changes during review.

src/mem/cache/cache_blk.hh
```diff
@@ -531,9 +518,52 @@ class TempCacheBlk final : public CacheBlk
...
+        CacheBlk::operator=(std::move(other));
+        assert(!other.isValid());
+
+        return *this;
+    }
+
+    TempCacheBlk &
+    operator=(CacheBlk &&) override
+    {
+        panic("A CacheBlk instance must not be moved into a TempCacheBlk, "
```
odanrc: Can we do
    TempCacheBlk & operator=(CacheBlk &&) override = delete;
instead to catch the error at compile time?

src/mem/cache/cache_blk.hh
```diff
@@ -531,9 +518,52 @@ class TempCacheBlk final : public CacheBlk
...
+              "because we cannot restore the address.");
+    }
+
```
odanrc, line 598: Shouldn't we also delete moving from TaggedEntry to TempCacheBlk?

odanrc (commented):
I think you can remove the "draft" tag. Most of my comments are nitpicks or hints, except for the "noexcept" and the TempCacheBlk ones.

src/mem/cache/cache_blk.hh
xcf-t: We actually cannot, because the parent implementation is virtual, so it cannot be deleted. At the same time, it must be virtual so that subclasses can add moves for custom metadata. Otherwise, major changes to the structure would be necessary. We could instead use a `dynamic_cast` and panic when the operator is used with another CacheBlk that is not a TempCacheBlk.

src/mem/cache/cache_blk.hh
xcf-t: We could. On the other hand, such a move would require that both the `TempCacheBlk` and the movement source are stored as `TaggedEntry*`. This will always result in weird partial assignment bugs, regardless of which subclass we are using. Moving directly from `TaggedEntry` to a `TempCacheBlk` shouldn't be possible, while the other case would be difficult to prevent without structural changes to the parent classes.

src/mem/cache/tags/sector_blk.hh
xcf-t: Thanks for the info. I'll update the PR to minimize unnecessary changes.

src/mem/cache/tags/sector_blk.hh
xcf-t: I'd argue that it is not an implementation detail but an explanation for why this implementation makes sense. When I first looked into this TaggedEntry inheritance chain, I found such choices somewhat opaque.

Also, I'm not sure what you mean by moving it to the implementation, as `using CacheBlk::operator=;` should just take the implementation from the parent class. Moving it there wouldn't make much sense either, as the comment is purely an explanation for why the parent operation can be used as is.

src/mem/cache/cache_blk.hh
```diff
@@ -532,8 +547,55 @@ class TempCacheBlk final : public CacheBlk
...
+        _addr = other._addr;
+
+        CacheBlk::operator=(std::move(other));
+        assert(!other.isValid());
+
+        return *this;
+    }
+
+    CacheBlk &
+    operator=(CacheBlk &&other) override
```
odanrc: I see, the compiler is doing some trickery that relies on this. Then I think it is better to keep it as you originally implemented, without the dynamic_cast (i.e., just the panic).

src/mem/cache/cache_blk.hh
odanrc: I am not sure I understand the answer. Are you saying that one would really want to break the code in order to use this move from TaggedEntry to TempCacheBlk?

In any case, if there is no drawback in deleting this operator, I see no reason not to do it as a safeguard measuer; however, if it suffers from the same issue of the move assignment (i.e., having to change a lot of the API due to the implicit compiler trickery), then you can drop it and consider this comment solved.

src/mem/cache/tags/sector_blk.hh
odanrc: Fair enough.

odanrc (approved):
Thanks for the changes. It looks like this change would fix the issue at hand. By the way, how did you verify that it behaves as intended? Did you manage to test the compressed and sector caches?

You should probably edit the merge request's description to remove the uncertainty and place it as a merge request comment (I don't recall if it goes into the git history).

src/mem/cache/cache_blk.hh
xcf-t: I think I might've written my explanation a bit ambiguously. 

Moving from TaggedEntry to TempCacheBlk would need some real effort. We cannot really prevent such a move from occurring here using delete, as this would have no real effect in the first place. TaggedEntry cannot be downcasted to a CacheBlk, so the defined move assignment operator doesn't apply. As far as I understand, the only way we could move from a TaggedEntry to another non-TaggedEntry type is if both of them are stored in TaggedEntry variables. This is possible but shouldn't actually happen. If it does, it causes an invalid state, regardless of which type of CacheBlk we mix.

If we want to really prevent this, we would need to make the operator virtual in the TaggedEntry so that we can override it in all subclasses.

xcf-t:
> Thanks for the changes. It looks like this change would fix the issue at hand. By the way, how did you verify that it behaves as intended? Did you manage to test the compressed and sector caches?
> 
> You should probably edit the merge request's description to remove the uncertainty and place it as a merge request comment (I don't recall if it goes into the git history).

I have verified the base implementation for standard `CacheBlk`s by re-running the same benchmarks that have caused this movement issue for me in the first place. Not a perfect strategy, but any invalid movement should be detected by the snoop filter. Unfortunately I don't have a method for testing compressed or sector caches at the moment. If you have an idea how to verify these or maybe a better method for standard blocks as well, I can do some more testing.

PR #3204, opened 2026-06-02
misc,util-m5,ext-testlib: add color to panics/warnings/infos

blokyk:
This PR colors the `panic`/`fatal`/`warn`/`info`/`hack` log prefixes with some color to make them easier to spot:
    - bold red for `panic` & `fatal`
    - bold yellow for `warn`
    - bold cyan for `info`
    - non-bold yellow for `hack`

This is also implemented for the equivalent python functions in src/python/m5/utils and ext/testlib.
    
It only uses those colors when stderr is a tty and [`NO_COLOR`](https://no-color.org) is not set, or when [`FORCE_COLOR`](https://force-color.org) is set. (In other words, it will only activate when the output is a terminal (thus avoiding weird sequences in log files), and it's possible to turn it off by setting `NO_COLOR=true`, and possible to forcing when piping input by setting `FORCE_COLOR=true`.) This logic has also been integrated into the `m5.util.terminal` module, so that it better respects those environment variables.

### DIFF
--- a/ext/testlib/terminal.py
+++ b/ext/testlib/terminal.py
@@ -107,7 +107,7 @@ def should_use_colors(out: TextIO):
     should_force = (
         "CLICOLOR_FORCE" in os.environ or "FORCE_COLOR" in os.environ
     )
-    istty = out.isatty() and os.environ["TERM"] not in [None, "dumb"]
+    istty = out.isatty() and os.getenv("TERM", default="dumb") != "dumb"
     return (not must_not_colour) and (should_force or istty)
 
 
@@ -117,10 +117,10 @@ else:
     tty_termcap = no_termcap
 
 
-def get_termcap(use_colors=None):
+def get_termcap(stream=sys.stdout, use_colors=None):
     if use_colors is None:
         # option unspecified; default behavior is to use colors iff isatty && NO_COLOR is not set
-        use_colors = should_use_colors(sys.stdout)
+        use_colors = should_use_colors(stream)
 
     if use_colors:
         return termcap
--- a/ext/testlib/terminal.py
+++ b/ext/testlib/terminal.py
@@ -27,9 +27,11 @@
 # Author: Steve Reinhardt
 
 import fcntl
+import os
 import struct
 import sys
 import termios
+from typing import TextIO
 
 # Intended usage example:
 #
@@ -99,18 +101,29 @@ class ColorStrings:
 termcap = ColorStrings(cap_string)
 no_termcap = ColorStrings(null_cap_string)
 
-if sys.stdout.isatty():
+
+def should_use_colors(out: TextIO):
+    must_not_colour = "NO_COLOR" in os.environ or "NOCOLOR" in os.environ
+    should_force = (
+        "CLICOLOR_FORCE" in os.environ or "FORCE_COLOR" in os.environ
+    )
+    istty = out.isatty() and os.environ["TERM"] not in [None, "dumb"]
+    return (not must_not_colour) and (should_force or istty)
+
+
+if should_use_colors(sys.stdout):
     tty_termcap = termcap
 else:
     tty_termcap = no_termcap
 
 
 def get_termcap(use_colors=None):
+    if use_colors is None:
+        # option unspecified; default behavior is to use colors iff isatty && NO_COLOR is not set
+        use_colors = should_use_colors(sys.stdout)
+
     if use_colors:
         return termcap
-    elif use_colors is None:
-        # option unspecified; default behavior is to use colors iff isatty
-        return tty_termcap
     else:
         return no_termcap
 
--- a/ext/testlib/terminal.py
+++ b/ext/testlib/terminal.py
@@ -103,12 +103,17 @@ no_termcap = ColorStrings(null_cap_string)
 
 
 def should_use_colors(out: TextIO):
-    must_not_colour = "NO_COLOR" in os.environ or "NOCOLOR" in os.environ
+    must_not_colour = (
+        os.getenv("NO_COLOR", default=os.getenv("NOCOLOR", default="")) != ""
+    )
     should_force = (
-        "CLICOLOR_FORCE" in os.environ or "FORCE_COLOR" in os.environ
+        os.getenv(
+            "CLICOLOR_FORCE", default=os.getenv("FORCE_COLOR", default="")
+        )
+        != ""
     )
     istty = out.isatty() and os.getenv("TERM", default="dumb") != "dumb"
-    return (not must_not_colour) and (should_force or istty)
+    return should_force or ((not must_not_colour) and istty)
 
 
 if should_use_colors(sys.stdout):
--- a/ext/testlib/terminal.py
+++ b/ext/testlib/terminal.py
@@ -104,13 +104,12 @@ no_termcap = ColorStrings(null_cap_string)
 
 def should_use_colors(out: TextIO):
     must_not_colour = (
-        os.getenv("NO_COLOR", default=os.getenv("NOCOLOR", default="")) != ""
+        os.getenv("NO_COLOR", default="") != ""
+        or os.getenv("NOCOLOR", default="") != ""
     )
     should_force = (
-        os.getenv(
-            "CLICOLOR_FORCE", default=os.getenv("FORCE_COLOR", default="")
-        )
-        != ""
+        os.getenv("CLICOLOR_FORCE", default="") != ""
+        or os.getenv("FORCE_COLOR", default="") != ""
     )
     istty = out.isatty() and os.getenv("TERM", default="dumb") != "dumb"
     return should_force or ((not must_not_colour) and istty)
--- a/src/base/logging.cc
+++ b/src/base/logging.cc
@@ -75,10 +75,13 @@ static bool
 shouldANSI()
 {
     bool mustNotColour =
-        getenv("NO_COLOR") != NULL || getenv("NOCOLOR") != NULL;
+        (getenv("NO_COLOR") != NULL && getenv("NO_COLOR")[0] != '\0') ||
+        (getenv("NOCOLOR") != NULL && getenv("NOCOLOR")[0] != '\0');
 
     bool shouldForce =
-        getenv("CLICOLOR_FORCE") != NULL || getenv("FORCE_COLOR") != NULL;
+        (getenv("CLICOLOR_FORCE") != NULL &&
+         getenv("CLICOLOR_FORCE")[0] != '\0') ||
+        (getenv("FORCE_COLOR") != NULL && getenv("FORCE_COLOR")[0] != '\0');
 
     char *term = getenv("TERM");
     bool isTerminal =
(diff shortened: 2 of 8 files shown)


blokyk:
(since there are no reviews yet, i'm pushing again to rebase and rerun the tests)

powerjg (commented):
Can you double check that when you redirect stdout/stderr (i.e., use `gem5.opt -re`) that the files do *not* have terminal coloring?

blokyk:
yep, i've checked again and `xxd` doesn't show any non-ascii characters in the resulting files, neither with `-re` nor with manual/shell redirection

powerjg:
I like the idea in this change. However, it's a relatively big UI change. I put this on the agenda for the developer meeting this Thursday: https://github.com/orgs/gem5/discussions/3201#discussioncomment-17242971

erin-le:
@blokyk: Hello, just wanted to let you know that the community doesn't seem opposed to this PR, so the gem5 developers will start reviewing it. For now, could you take a look at the review comments from Copilot? Thank you!

blokyk:
@erin-le fixed the robot comments, although the one about `std::cerr.rdbuf` isn't really "properly fixable" and the proposed fix would not work as expected since the redirection of `cerr` can be done after `shouldANSI()` has been called and the loggers created, so they would still be colored despite the cerr now being redirected. instead of adding a basically useless check, i opted for simply setting `TERM=dummy` in the logger tests, which has the same effect (and is more "accurate" if anything else uses `$TERM` in the future) but is a lot less intrusive. (also, the CI tests succeeded on the previous version, so i'm pretty sure it was a non-issue to begin with? not sure if it's just because of the CI environment, since i can't run the tests locally...)

i don't have any easy/obvious fix for this, as it's just a consequence of logger's lifetimes and how colors are injected. if there's anything i missed and you have an idea, i'm all ears.

blokyk:
the clang-format check is failing in files i didn't touch (and the local check runs fine on my machine), so i have no idea what to do here except maybe rebase on latest `develop`? but that would invalidate the robot comments, and some devs don't like rebases, so i want to get your ok-go first

erin-le:
@blokyk: No problem with rebasing; you could also try the steps listed here: https://github.com/orgs/gem5/discussions/3201#discussioncomment-17242898

blokyk:
rebased on latest develop, tried the steps for clang-format (i did have a too-recent version (21.1.8)), but nothing changed, so we'll see i guess ¯\\\_(ツ)_/¯

blokyk:
hi, is there anything more i can do to get this pr over the finish line?

erin-le:
Hey @blokyk, there isn't anything left for you to do right now; the gem5 developers just have to find some time to review this PR. Thanks for your patience with the review process!

ext/testlib/terminal.py
```diff
@@ -99,18 +101,29 @@ def __init__(self, cap_string):
...
 
-if sys.stdout.isatty():
+
+def should_use_colors(out: TextIO):
+    must_not_colour = "NO_COLOR" in os.environ or "NOCOLOR" in os.environ
+    should_force = (
+        "CLICOLOR_FORCE" in os.environ or "FORCE_COLOR" in os.environ
+    )
+    istty = out.isatty() and os.getenv("TERM", default="dumb") != "dumb"
+    return (not must_not_colour) and (should_force or istty)
```
erin-le: The logic here is different from the logic in `src/python/m5/util/terminal.py`, and NO_COLOR takes precedence over FORCE_COLOR when both are set. It might be better to use the same logic so things are consistent:

```suggestion
    return should_force or ((not must_not_colour) and istty)
```

ext/testlib/terminal.py
```diff
@@ -99,18 +101,29 @@ def __init__(self, cap_string):
 termcap = ColorStrings(cap_string)
 no_termcap = ColorStrings(null_cap_string)
 
-if sys.stdout.isatty():
+
+def should_use_colors(out: TextIO):
+    must_not_colour = "NO_COLOR" in os.environ or "NOCOLOR" in os.environ
```
erin-le: Nit: The documentation for NO_COLOR says that it shouldn't be considered "set" when it is an empty string:
```suggestion
    must_not_colour = (
        "NO_COLOR" in os.environ and os.getenv("NO_COLOR") != ""
    ) or ("NOCOLOR" in os.environ and os.getenv("NOCOLOR") != "")
```

ext/testlib/terminal.py
```diff
@@ -99,18 +101,29 @@ def __init__(self, cap_string):
...
+    should_force = (
+        "CLICOLOR_FORCE" in os.environ or "FORCE_COLOR" in os.environ
+    )
```
erin-le, line 113: Nit: Similarly, the documentation for FORCE_COLOR says it forces color when set to values other than an empty string:

```suggestion
    should_force = (
        "CLICOLOR_FORCE" in os.environ and os.getenv("CLICOLOR_FORCE") != ""
    ) or ("FORCE_COLOR" in os.environ and os.getenv("FORCE_COLOR") != "")
```

src/python/m5/util/terminal.py
```diff
@@ -95,18 +97,29 @@ def __init__(self, cap_string):
 termcap = ColorStrings(cap_string)
 no_termcap = ColorStrings(null_cap_string)
 
-if sys.stdout.isatty():
+
+def should_use_colors(out: TextIO):
+    must_not_colour = "NO_COLOR" in os.environ or "NOCOLOR" in os.environ
```
erin-le: Nit: Similar comment as above:
```suggestion
    must_not_colour = (
        "NO_COLOR" in os.environ and os.getenv("NO_COLOR") != ""
    ) or ("NOCOLOR" in os.environ and os.getenv("NOCOLOR") != "")
```

src/python/m5/util/terminal.py
```diff
@@ -95,18 +97,29 @@ def __init__(self, cap_string):
...
+    should_force = (
+        "CLICOLOR_FORCE" in os.environ or "FORCE_COLOR" in os.environ
+    )
```
erin-le, line 111: Nit: Similar comment as above:

```suggestion
    should_force = (
        "CLICOLOR_FORCE" in os.environ and os.getenv("CLICOLOR_FORCE") != ""
    ) or ("FORCE_COLOR" in os.environ and os.getenv("FORCE_COLOR") != "")
```

src/base/logging.test.cc
```diff
@@ -38,12 +40,28 @@ using namespace gem5;
...
 {
   public:
-    void SetUp() override { old = std::cerr.rdbuf(gtestLogOutput.rdbuf()); }
-    void TearDown() override { std::cerr.rdbuf(old); }
+    void
+    SetUp() override
+    {
+        old = std::cerr.rdbuf(gtestLogOutput.rdbuf());
+        oldTerm = getenv("TERM");
+        setenv("TERM", "dummy", /* overwrite */ true);
```
erin-le: I ran this test locally, where I had terminal colors enabled, and got a number of test failures. Making the following change fixed it for me though. Can you double check if this is a proper fix?
```suggestion
        setenv("TERM", "dumb", /* overwrite */ true);
```

src/base/logging.cc
```diff
@@ -69,6 +71,22 @@ class FatalLogger : public ExitLogger
     void exit() override { ::exit(1); }
 };
 
+static bool
+shouldANSI()
+{
+    bool mustNotColour =
+        getenv("NO_COLOR") != NULL || getenv("NOCOLOR") != NULL;
```
erin-le: Nit: Similar to the comments above, check if NO_COLOR and FORCE_COLOR are empty strings, and don't apply their effects if they are:
```suggestion
    bool mustNotColour =
        (getenv("NO_COLOR") != NULL && getenv("NO_COLOR")[0] != '\0') ||
        (getenv("NOCOLOR") != NULL && getenv("NOCOLOR")[0] != '\0');
```

src/base/logging.cc
```diff
@@ -69,6 +71,22 @@ class FatalLogger : public ExitLogger
...
+
+    bool shouldForce =
+        getenv("CLICOLOR_FORCE") != NULL || getenv("FORCE_COLOR") != NULL;
```
erin-le:
```suggestion
    bool shouldForce =
        (getenv("CLICOLOR_FORCE") != NULL &&
         getenv("CLICOLOR_FORCE")[0] != '\0') ||
        (getenv("FORCE_COLOR") != NULL && getenv("FORCE_COLOR")[0] != '\0');
```

erin-le (commented):
@blokyk: This mostly looks good and works as expected; I left a few comments with suggested code, but feel free to use something else if it would work better, and also let me know if there's something I'm mistaken about. Thanks!

blokyk:
hi @erin-le, addressed the comments (sorry for the dumb inconsistency between the python scripts...), although i (ab)used `os.getenv`'s `default` parameter instead of checking inside `os.environ` first. hope that's okay!

ext/testlib/terminal.py
```diff
@@ -99,18 +101,34 @@ def __init__(self, cap_string):
...
+def should_use_colors(out: TextIO):
+    must_not_colour = (
+        os.getenv("NO_COLOR", default=os.getenv("NOCOLOR", default="")) != ""
+    )
+    should_force = (
+        os.getenv(
+            "CLICOLOR_FORCE", default=os.getenv("FORCE_COLOR", default="")
+        )
+        != ""
+    )
```
erin-le, line 113: Here, I would prefer something like the suggestion below. Currently, when NOCOLOR=true but NO_COLOR="", colors still appear in the terminal, and when FORCE_COLOR=true but CLICOLOR_FORCE="", colors aren't forced.
```suggestion
    # is there a *non-empty* variable named NO_COLOR/NOCOLOR in the env?
    must_not_colour = (
        os.getenv("NO_COLOR", default="") != ""
        or os.getenv("NOCOLOR", default="") != ""
    )
    # is there a *non-empty* variable named CLICOLOR_FORCE/FORCE_COLOR in the env?
    should_force = (
        os.getenv("CLICOLOR_FORCE", default="") != ""
        or os.getenv("FORCE_COLOR", default="") != ""
    )
```

src/python/m5/util/terminal.py
```diff
@@ -95,18 +97,36 @@ def __init__(self, cap_string):
...
+    must_not_colour = (
+        os.getenv("NO_COLOR", default=os.getenv("NOCOLOR", default="")) != ""
+    )
+    # is there a *non-empty* variable named CLICOLOR_FORCE/FORCE_COLOR in the env?
+    should_force = (
+        os.getenv(
+            "CLICOLOR_FORCE", default=os.getenv("FORCE_COLOR", default="")
+        )
+        != ""
+    )
```
erin-le, line 111: Same comment as above

erin-le (changes requested):
Hey @blokyk, thanks for the changes! Just a couple more minor comments:

erin-le (approved):

PR #3208, opened 2026-06-04
dev: refactor PciUpstream to be a C++ interface

clemdiep:
This PR refactor PciUpstream to be a C++ interface instead of a full SimObject. As discussed in #2561, this allows to keep common code between PCI host bridge and (futur) PCI to PCI bridges, without having distinct SimObject for the PCI to PCI bridge.

Also, this move PciBus (the XBar to which are connected PciDevice) to be a child of PciHost, as it will always be needed and adds some helper functions to connect ports related to PciHost, PciBus and device, reducing possible point of mistake. This represent most of the Python changes and may be put in another PR if required.

Made in collaboration with @marinazapater

### DIFF
--- a/src/dev/arm/RealView.py
+++ b/src/dev/arm/RealView.py
@@ -75,7 +75,6 @@ from m5.objects.PciDevice import (
     PciLegacyIoBar,
 )
 from m5.objects.PciHost import *
-from m5.objects.PciUpstream import PciBus
 from m5.objects.Platform import Platform
 from m5.objects.PS2 import *
 from m5.objects.Scmi import *
@@ -869,11 +868,6 @@ class RealView(Platform):
             else:
                 dma_ports.append(device.dma)
 
-    def _attach_pci_device(self, device, upstream, bus):
-        device.pio = bus.mem_side_ports
-        device.dma = bus.cpu_side_ports
-        upstream.devices.append(device)
-
     def _attach_io(self, devices, *args, **kwargs):
         for d in devices:
             self._attach_device(d, *args, **kwargs)
@@ -1012,7 +1006,6 @@ class VExpress_EMM(RealView):
         conf_device_bits=16,
         pci_pio_base=0,
     )
-    pci_bus = PciBus()
 
     sys_counter = SystemCounter()
     generic_timer = GenericTimer(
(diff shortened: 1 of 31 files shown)


clemdiep:
Hey @giactra, here is a new PR for my proposed design on #2561

src/dev/pci/upstream.hh
```diff
@@ -232,63 +200,60 @@ class PciUpstream : public ClockedObject
     /**
      * Post an interrupt to the CPU.
      *
-     * @param dev_addr The device's position on the PCI bus
+     * @param device The requesting PCI device
      * @param pin PCI interrupt pin
```
giactra: Remove the pin reference here

src/dev/pci/upstream.hh
```diff
@@ -232,63 +200,60 @@ class PciUpstream : public ClockedObject
...
-                                  PciIntPin pin) = 0;
+    virtual void interfacePostInt(const PciDevice &device) = 0;
 
     /**
-     * Post an interrupt to the CPU.
+     * Clear an interrupt to the CPU.
      *
-     * @param dev_addr The device's position on the PCI bus
+     * @param device The requesting PCI device
      * @param pin PCI interrupt pin
```
giactra: Ditto

src/dev/pci/upstream.hh
```diff
@@ -323,6 +288,8 @@ class PciUpstream : public ClockedObject
     PciUpDownBridge *upToDown;
 
     PciConfigError *configErrorDevice;
+
+    SimObject &upstreamObject;
```
giactra: Is there a second expected use for upstreamObject (maybe in the future)? At the moment it looks like the only reason for passing a SimObject reference is to use the name method... This effectively ties the PciUpstream to always be used with double inheritance with a SimObject.

If you want to keep it this way, this should be documented within the class. Alternatively you can pass a string in the constructor and decouple completely PciUpstream from inheritance requirements

giactra (commented):
Thanks @clemdiep . I have some minor comments. LGTM overall

src/dev/arm/RealView.py
```diff
@@ -1523,21 +1513,19 @@ def attachPciDevice(self, device):
...
-            bus.cpu_side_ports = self.pci_host.up_request_port()
+            self.pci_host.connect_upper_bus(bus, True)
         else:
-            dma_ports.append(self.pci_host.up_request_port())
+            pci_dma = self.pci_host.connect_upper_bus(bus, False)
+            dma_ports.append(pci_dma)
 
         for d in devices:
-            self._attach_pci_device(d, self.pci_host, self.pci_bus)
+            self.pci_host.connect_device(device)
```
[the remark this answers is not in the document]
giactra: @clemdiep please review this (it's a real bug)

giactra (commented):
Thanks @clemdiep for your fast changes.
I underestimated the DPRINTF problem which requires a name method to be defined.

I think we can workaround the double definition by defining a new DPRINTF flavour (see all the flavours available in src/base/trace.hh). Effectively we need a new DPRINTF where we pass the upstreamName directly to the macro.

Anyway I am not requiring this to be fixed now as I don't want to stall this further. We can add this capability later on.

However, there are remaining bugs in this PR (flagged by Co-pilot). Please review them. I will approve once they are handled

src/dev/arm/RealView.py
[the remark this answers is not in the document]
clemdiep: I fixed this !

clemdiep:
Thank you for the review @giactra !

> I underestimated the DPRINTF problem which requires a name method to be defined.

In fact, the problem with the name method was already there. DPRINTF was using the default one defined in src/base/trace.hh for the upstream class and I did pay attention until I firstly named upstreamName to name and got compilation errors...
 
> However, there are remaining bugs in this PR (flagged by Co-pilot). Please review them. I will approve once they are handled

I fixed them, it detected another bug on the amdgpu were the PCI device was connected twice, which would lead to errors. I have revert to not use the new PciHost connect_device function for the amdgpu, as ViperShader is already connecting the dma to Ruby and connect_device would also connect this port and raise an error. I tested to run the GPU-FS configuration and this seems to be ok now.

giactra:
Hi @abmerop, adding you as a reviewer since this touches GPU platforms and could be relevant to you. Please let me know if you need some time to review this change. If not, I will approve and merge this on Tuesday

abmerop:
> Hi @abmerop, adding you as a reviewer since this touches GPU platforms and could be relevant to you. Please let me know if you need some time to review this change. If not, I will approve and merge this on Tuesday

Confirmed working on my end

abmerop (approved):

PR #3209, opened 2026-06-05
mem-cache: Add profiling statistics to replacement policy

HirunaVishwamith:
The cache replacement policies did not expose any statistics, so their behavior could not be profiled. This change adds profiling counters that are shared by every replacement policy.

The public interface (invalidate, touch, reset and getVictim) is turned into a Non-Virtual Interface: the public methods are now small wrappers in replacement_policy::Base that gather statistics and then forward to protected *Impl methods implemented by each policy. Keeping the counters in one place means no policy can accidentally bypass them, and the callers in the cache tags are left unchanged.

The new statistics record the number of victimizations, hits, insertions and invalidations, plus a formula for the average number of hits an entry receives before being replaced. WeightedLRU keeps its occupancy-aware touch() overload, which is now routed through the common interface so the access is also counted.

All policies are updated to implement the *Impl methods. A unit test for the new counters is added. The existing replacement policy unit tests are updated to link the statistics implementation, which is now required to instantiate any policy.

### DIFF
--- a/src/mem/cache/replacement_policies/SConscript
+++ b/src/mem/cache/replacement_policies/SConscript
@@ -46,13 +46,35 @@ Source('ship_rp.cc')
 Source('tree_plru_rp.cc')
 Source('weighted_lru_rp.cc')
 
+# The common replacement policy base class (replacement_policy::Base) now
+# gathers profiling statistics. Instantiating any policy therefore pulls in
+# the statistics implementation and its dependencies, none of which are part
+# of the 'gem5 simobject' tag. The unit tests below must link them explicitly,
+# and the 'socket_test' tag is reused to pull in the zlib (-lz) dependency
+# required by base/output.cc.
+stats_srcs = [
+    '../../../base/statistics.cc',
+    '../../../base/stats/info.cc',
+    '../../../base/stats/storage.cc',
+    '../../../base/time.cc',
+    '../../../base/hostinfo.cc',
+    '../../../base/output.cc',
+    '../../../sim/core.cc',
+    '../../../sim/root.cc',
+    '../../../sim/globals.cc',
+    '../../../sim/tags.cc',
+]
+stats_tags = [with_tag('gem5 simobject'), with_tag('socket_test')]
+
 GTest('replaceable_entry.test', 'replaceable_entry.test.cc')
+GTest('base_rp.test', 'lru_rp.cc', 'base_rp.test.cc', *stats_srcs,
+    *stats_tags)
 GTest('tree_plru_rp.test', 'tree_plru_rp.cc', 'tree_plru_rp.test.cc',
-    with_tag('gem5 simobject'))
-GTest('fifo_rp.test', 'fifo_rp.cc', 'fifo_rp.test.cc',
-    with_tag('gem5 simobject'))
-GTest('lfu_rp.test', 'lfu_rp.cc', 'lfu_rp.test.cc', with_tag('gem5 simobject'))
-GTest('lru_rp.test', 'lru_rp.cc', 'lru_rp.test.cc', with_tag('gem5 simobject'))
-GTest('mru_rp.test', 'mru_rp.cc', 'mru_rp.test.cc', with_tag('gem5 simobject'))
+    *stats_srcs, *stats_tags)
+GTest('fifo_rp.test', 'fifo_rp.cc', 'fifo_rp.test.cc', *stats_srcs,
+    *stats_tags)
+GTest('lfu_rp.test', 'lfu_rp.cc', 'lfu_rp.test.cc', *stats_srcs, *stats_tags)
+GTest('lru_rp.test', 'lru_rp.cc', 'lru_rp.test.cc', *stats_srcs, *stats_tags)
+GTest('mru_rp.test', 'mru_rp.cc', 'mru_rp.test.cc', *stats_srcs, *stats_tags)
 GTest('second_chance_rp.test', 'fifo_rp.cc', 'second_chance_rp.cc',
-    'second_chance_rp.test.cc', with_tag('gem5 simobject'))
+    'second_chance_rp.test.cc', *stats_srcs, *stats_tags)
--- a/src/mem/cache/replacement_policies/base.hh
+++ b/src/mem/cache/replacement_policies/base.hh
@@ -32,6 +32,7 @@
 #include <memory>
 
 #include "base/compiler.hh"
+#include "base/statistics.hh"
 #include "mem/cache/replacement_policies/replaceable_entry.hh"
 #include "mem/packet.hh"
 #include "params/BaseReplacementPolicy.hh"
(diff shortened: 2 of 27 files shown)


src/mem/cache/replacement_policies/base_rp.test.cc
```diff
@@ -0,0 +1,146 @@
...
+#include "sim/eventq.hh"
+
+using namespace gem5;
+
+namespace
+{
+
+// LRU calls curTick(), so a current event queue must exist for the policy to
+// be exercised. See the fixture below.
+EventQueue eventQueue("BaseRPStatsTest Event Queue");
```
odanrc, line 53: If we used LFU we wouldn't need EventQueue's dependency, making the test slightly simpler. This change could also be used to improve the fixture, not requiring makeReplacementPolicy() anymore, allowing us to simply have:

```
class BaseRPStatsTest : public ::testing::Test
{
  protected:
    TestableLFU rp;
    BaseRPStatsTest() : rp(LFURPParams()) {}
};
```

src/mem/cache/replacement_policies/base.hh
```diff
@@ -50,65 +51,157 @@ namespace replacement_policy
...
+ * by this base class. These wrappers gather profiling statistics common to
+ * every policy and then forward to the policy-specific implementation. This
+ * Non-Virtual Interface keeps statistics collection in a single place, so no
+ * policy can accidentally bypass it.
  */
 class Base : public SimObject
 {
+  protected:
+    /** Profiling information shared by all replacement policies. */
+    struct ReplacementStats : public statistics::Group
```
odanrc, line 67: I see 3 possible approaches to take:
- **the "gem5 pattern"** (i.e., this proposal, and what we currently do everywhere)
PROS: It is simple, works; and people are used to it, so adopting it is a no-brainer.
CONS: Breaks the single responsibility principle (SRP): now RPs take care of RP AND stats.

- **decorator pattern**
```
class StatisticsDecorator : public ReplacementPolicy {
private:
    std::shared_ptr<ReplacementPolicy> wrappedPolicy;
    Stats stats;
    
public:
    void reset(const std::shared_ptr<ReplacementData>& data) override {
        stats.resets++;
        wrapped_policy->reset(data);
    }
    // ... similar for other methods
};
```
PROS: Simple, maintains SRP, and allow enabling/disabling stats easily. Decorators can even be added one on top of the other to selectively add more stats. We can even add decorators to add tracing! IMO it is the ideal solution.
CONS: Additional learning curve. Has an ugly instantiation (rp = replacement_policy::StatisticsDecorator(LFU()), or something like that), but that drawback can be reduced with a Builder pattern, I think.

- **observer pattern**
```
class ReplacementPolicy {
    std::vector<ReplacementPolicyListener*> listeners;
    
    void touch(const std::shared_ptr<ReplacementData>& data) override {
        for (auto listener : listeners) listener->onTouch();
        touchImpl(data);
    }
};

class StatisticsListener : public ReplacementPolicyListener {
    void onTouch() override { stats.touches++; }
    // ...
};
```
PROS: allows enabling/disabling stats easily, and we can add listeners for things other than stats, like tracing, and whatever else.
CONS: it still breaks SRP, but it is more complex (yeah, I don't think this is a good solution) ;)

================================================================
All that said, _I am ok with the proposed solution_ - so why do I bring the alternatives up, you may ask? At some point we were discussing enabling/disabling stats (maybe with @giactra ?), and these alternative patterns would be able to cover that. We have the opportunity here to start leaning into the change - but do we want to take it? I think the answer is "no, if we do this we should apply this everywhere, not only here", but at the same time I personally see the RPs as a starting point for change (the first SimObject's unit tests are RP's).

Thoughts?

src/mem/cache/replacement_policies/base.hh
```diff
@@ -50,65 +51,157 @@ namespace replacement_policy
...
      * Invalidate replacement data to set it as the next probable victim.
      *
      * @param replacement_data Replacement data to be invalidated.
      */
-    virtual void invalidate(const std::shared_ptr<ReplacementData>&
-        replacement_data) = 0;
+    void
+    invalidate(const std::shared_ptr<ReplacementData> &replacement_data)
+    {
+        stats.invalidations++;
```
odanrc, line 115: Nit: we can't invalidate more than we have inserted, so we could add an assertion for that here

src/mem/cache/replacement_policies/base_rp.test.cc
```diff
@@ -0,0 +1,146 @@
...
+
+    rp->touch(data);
+    rp->touch(data);
+    EXPECT_EQ(rp->hits(), 2);
+
+    rp->invalidate(data);
+    EXPECT_EQ(rp->invalidations(), 1);
+
+    // The other counters must be unaffected by unrelated operations.
+    EXPECT_EQ(rp->victimizations(), 0);
```
odanrc, line 125: We should probably do the full round of checks for each of the 3 spots here. i.e./e.g., where we currently check that insertions are updated when reset is called, we should also check that the other stats remain unchanged.

src/mem/cache/replacement_policies/SConscript
```diff
@@ -46,13 +46,35 @@ Source('ship_rp.cc')
 Source('tree_plru_rp.cc')
 Source('weighted_lru_rp.cc')
 
+# The common replacement policy base class (replacement_policy::Base) now
+# gathers profiling statistics. Instantiating any policy therefore pulls in
+# the statistics implementation and its dependencies, none of which are part
+# of the 'gem5 simobject' tag. The unit tests below must link them explicitly,
+# and the 'socket_test' tag is reused to pull in the zlib (-lz) dependency
+# required by base/output.cc.
+stats_srcs = [
```
odanrc, line 55: This is not the way to tag files; the proper way to do is to go to each of these files and add ", tags=['gem5 stats']". Check src/base/SConscript for multiple examples. Although from the huge list of files here it looks like we haven't cleaned up the dependencies for the stats yet, so you may have some trouble doing so - I am not sure.

src/mem/cache/replacement_policies/SConscript
```diff
@@ -46,13 +46,35 @@ Source('ship_rp.cc')
...
+    '../../../base/stats/storage.cc',
+    '../../../base/time.cc',
+    '../../../base/hostinfo.cc',
+    '../../../base/output.cc',
+    '../../../sim/core.cc',
+    '../../../sim/root.cc',
+    '../../../sim/globals.cc',
+    '../../../sim/tags.cc',
+]
+stats_tags = [with_tag('gem5 simobject'), with_tag('socket_test')]
```
odanrc, line 67: Nit: I had the impression that we had with_tags, but I can't recall where these are declared to look it up.

odanrc (commented):
Thanks for the contribution, and extra thanks for adding documentation and a unit test!

I brought up the design pattern issue, but it is not that big of a deal - it can be solved in the future, so you can consider it a nitpick if you want.

src/mem/cache/replacement_policies/base.hh
odanrc: Yeah... after giving it some thought, I don't think the decorator pattern would work without significant rework of the whole codebase, since it only works well for the boundaries of a function. Since we only care about enabling/disabling stats, this can be done easily through other means, like what we currently do for tracing.

odanrc (commented):
I forgot to ask: could you re-title this PR to include the "mem-cache:" prefix? You can check other PRs for examples.

erin-le:
@HirunaVishwamith: Hello, could you respond to the comments on this PR when you get a chance? Thank you!

erin-le:
@HirunaVishwamith: Hello, could you give an update on the status of this PR on your end? If you aren't planning to work on it further, would you be alright with closing? It can be reopened later if you want to come back to it. Thank you!

HirunaVishwamith:
> @HirunaVishwamith: Hello, could you give an update on the status of this PR on your end? If you aren't planning to work on it further, would you be alright with closing? It can be reopened later if you want to come back to it. Thank you!

Since the changes in this PR aren't important, I'll close it for now.

PR #3210, opened 2026-06-06
misc: Add GitHub Agentic Workflow for diagnosing test failures

erin-le:
This commit adds a GitHub Agentic Workflow for diagnosing test failures in the Weekly, Compiler, Daily, and CI tests. The workflow will rerun failed jobs if it determines that the jobs failed due to flakiness/intermittent failures in the tests or runners. Otherwise, it will open an issue for failures in the Weekly, Compiler, and Daily tests, and leave a comment on the relevant PR for failures in CI tests.


.github/workflows/test-failure-doctor.md
```diff
@@ -0,0 +1,180 @@
...
+due to runner instability or other reasons unrelated to the code/code changes.
+**ONLY proceed if the workflow conclusion is 'failure' or 'cancelled'**.
+
+### Phase 1: Initial Triage
+
+1. **Verify Failure**: Check that the workflow status is `failure` or `cancelled`
+   - **If the workflow was successful**: Do not proceed with any further analysis on the current test workflow, and immediately start looking at the next test workflow that finished within the last day.
+   - **If the workflow failed**: Proceed with the investigation steps below.
+
+   - **If the workflow failed and the latest run was a rerun triggered by a maintainer or by the Test Failure Doctor**: Make an issue about the failure, and *do not* rerun the failing workflow, even if the failure category was "Flaky Tests" or "Infrastructure".
```
powerjg: The "make an issue" command here... should you add something to check to see if an issue already exists? We also *don't* want to make issues on CI failures, we just want to report it to the submitter.

.github/workflows/test-failure-doctor.md
```diff
@@ -0,0 +1,180 @@
...
+   - **If the workflow was intentionally cancelled by a maintainer**: Most of the tests are automatically launched by the `github-actions` bot, although they will occasionally be launched by a maintainer. Disregard who the test was launched by, and only pay attention to who **cancelled** the test. If the test was cancelled by a maintainer, leave the message "Test workflow cancelled intentionally - no investigation needed" and **stop immediately**. Do not proceed with any further analysis, and start looking at the next test workflow that finished within the last day.
+   - **If the workflow was cancelled under other circumstances**: proceed with the investigation steps below.
+2. **Get Workflow Details**: Use `get_workflow_run` to get full details of the failed run
+3. **List Jobs**: Use `list_workflow_jobs` to identify which specific jobs failed
+4. **Quick Assessment**: Determine if this is a new type of failure or a recurring pattern
+
+### Phase 2: Deep Log Analysis
+
+1. **Retrieve Logs**: Use `get_job_logs` with `failed_only=true` to get logs from all failed jobs
+2. **Pattern Recognition**: Analyze logs for:
```
powerjg, line 85: Do you want to tell it to look in tests-output/.../.../.../.../simout & simerr.txt?

.github/workflows/test-failure-doctor.md
```diff
@@ -0,0 +1,180 @@
...
+
+2. **Deep Dive Analysis**:
+   - For test failures: Identify specific test methods and assertions
+   - For build failures: Analyze compilation errors and missing dependencies
+   - For infrastructure issues: Check runner logs and resource usage
+   - For timeout issues: Identify slow operations and bottlenecks
+
+### Phase 4: Rerun workflow if necessary
+
+1. If the failure type from the previous step was **Infrastructure** or **Flaky Tests**, rerun the failed tests in the **workflow run that triggered this Test Failure Doctor run** using the rerun-failed-jobs tool.
```
powerjg: I think you should define flaky tests. E.g., if the failure test has nothing to do with the PR or if there is no error in the error logs.

.github/workflows/test-failure-doctor.md
```diff
@@ -0,0 +1,180 @@
...
+
+- Don't run this step if the failure type was **Flaky Tests**, unless the latest run of the triggering workflow was a rerun by the Test Failure Doctor or by a maintainer.
+
+1. **Create Investigation Report**: Generate a comprehensive analysis including:
+   - **Executive Summary**: Quick overview of the failure
+   - **Root Cause**: Detailed explanation of what went wrong
+   - **Reproduction Steps**: How to reproduce the issue locally
+   - **Recommended Actions**: Specific steps to fix the issue. Suggest code changes or configuration updates, and provide specific file locations and line numbers for fixes.
+
+2. **Actionable Deliverables**:
```
powerjg, line 144: I would update the "create an issue" above with this information.

.github/workflows/test-failure-doctor.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Recommended Actions
+- [ ] [Specific actionable steps]
+
+```
+
+## Important Guidelines
+
+- **Be Thorough**: Don't just report the error - investigate the underlying cause
+- **Use Memory**: Always check for similar past failures and learn from them
```
powerjg: Is there memory in this service?

.github/workflows/test-failure-doctor.md
```diff
@@ -0,0 +1,180 @@
...
+- **Be Specific**: Provide exact file paths, line numbers, and error messages
+- **Be Concise**: Write issues and comments with concise language
+- **Action-Oriented**: Focus on actionable recommendations, not just analysis
+- **Resource Efficient**: Use caching to avoid re-downloading large logs
```
powerjg: Is this possible with this service?

.github/workflows/test-failure-doctor.md
erin-le: I'll make the description in line 108 more specific, and also add some examples of words and phrases to look out for.

.github/workflows/test-failure-doctor.md
erin-le: This line is just something that I forgot to delete. It is possible to have memory using the `cache-memory` tool, which is what the [original CI Doctor workflow](https://github.com/githubnext/agentics/blob/main/workflows/ci-doctor.md?plain=1) uses, but I removed most of the instructions to use memory for two reasons:
1. From what I've seen, test failures in gem5 tend to be unrelated to each other, so it might not be useful to cache past failures
2. I was worried that it would consume additional tokens if the agent had to look through the cache of past failures

However, if you think it would be useful to have memory, I can modify the workflow to add it back in.

.github/workflows/test-failure-doctor.md
erin-le: Same situation as above; a line that I forgot to delete

.github/workflows/test-failure-doctor.md
erin-le: I took a look into this and I don't think `get_job_logs` actually returns the folders that contain `simout.txt` and `simerr.txt` files. It looks like it just returns the same log that you can see when you click the settings button for a job, then "View raw logs". This only includes the terminal outputs from running the jobs, i.e. the gem5 build output, and the messages from running TestLib.

BobbyRBruce (approved):

PR #3213, opened 2026-06-10
arch-vega: Instructions related to 16-bit data types

abmerop:
Adds several instructions related to 16-bit data types (bf16, fp16):
- Scalar packing instructions
- Flat and derivative segment D16 loads/stores
- LDS D16 reads/writes

### DIFF
--- a/src/arch/amdgpu/vega/insts/flat.cc
+++ b/src/arch/amdgpu/vega/insts/flat.cc
@@ -93,7 +93,130 @@ Inst_FLAT__FLAT_LOAD_UBYTE::completeAcc(GPUDynInstPtr gpuDynInst)
     }
     vdst.write();
 } // execute
-// --- Inst_FLAT__FLAT_LOAD_SBYTE class methods ---
+// --- Inst_FLAT__FLAT_LOAD_UBYTE_D16 class methods ---
+
+Inst_FLAT__FLAT_LOAD_UBYTE_D16::Inst_FLAT__FLAT_LOAD_UBYTE_D16(
+    InFmt_FLAT *iFmt)
+    : Inst_FLAT(iFmt, "flat_load_ubyte_d16")
+{
+    setFlag(MemoryRef);
+    setFlag(Load);
+} // Inst_FLAT__FLAT_LOAD_UBYTE_D16
+
+Inst_FLAT__FLAT_LOAD_UBYTE_D16::~Inst_FLAT__FLAT_LOAD_UBYTE_D16()
+{} // ~Inst_FLAT__FLAT_LOAD_UBYTE_D16
+
+// VDATA[15 : 0].u16 = 16'U({ 8'0U, MEM[addr].u8 });
+// VDATA[31:16] is preserved.
+void
+Inst_FLAT__FLAT_LOAD_UBYTE_D16::execute(GPUDynInstPtr gpuDynInst)
+{
+    Wavefront *wf = gpuDynInst->wavefront();
+
+    if (gpuDynInst->exec_mask.none()) {
+        wf->decVMemInstsIssued();
+        wf->untrackVMemInst(gpuDynInst);
+        if (isFlat()) {
+            wf->decLGKMInstsIssued();
+            wf->untrackLGKMInst(gpuDynInst);
+        }
+        return;
+    }
+
+    gpuDynInst->execUnitId = wf->execUnitId;
+    gpuDynInst->latency.init(gpuDynInst->computeUnit());
+    gpuDynInst->latency.set(gpuDynInst->computeUnit()->clockPeriod());
+
+    calcAddr(gpuDynInst, extData.ADDR, extData.SADDR, instData.OFFSET);
+
+    issueRequestHelper(gpuDynInst);
+} // execute
+
+void
+Inst_FLAT__FLAT_LOAD_UBYTE_D16::initiateAcc(GPUDynInstPtr gpuDynInst)
+{
+    initMemRead<VecElemU8>(gpuDynInst);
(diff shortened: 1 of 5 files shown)


src/arch/amdgpu/vega/insts/flat.cc
```diff
@@ -93,7 +93,124 @@ Inst_FLAT__FLAT_LOAD_UBYTE::completeAcc(GPUDynInstPtr gpuDynInst)
...
+
+    calcAddr(gpuDynInst, extData.ADDR, extData.SADDR, instData.OFFSET);
+
+    issueRequestHelper(gpuDynInst);
+} // execute
+
+void
+Inst_FLAT__FLAT_LOAD_UBYTE_D16::initiateAcc(GPUDynInstPtr gpuDynInst)
+{
+    initMemRead<VecElemU8>(gpuDynInst);
```
mattsinc, line 138: Should this be a U16 instead of U8?

src/arch/amdgpu/vega/insts/flat.cc
```diff
@@ -93,7 +93,124 @@ Inst_FLAT__FLAT_LOAD_UBYTE::completeAcc(GPUDynInstPtr gpuDynInst)
...
+} // initiateAcc
+
+void
+Inst_FLAT__FLAT_LOAD_UBYTE_D16::completeAcc(GPUDynInstPtr gpuDynInst)
+{
+    VecOperandU32 vdst(gpuDynInst, extData.VDST);
```
mattsinc, line 144: Shouldn't this be a U16 instead of a U32?

src/arch/amdgpu/vega/insts/flat.cc
```diff
@@ -93,7 +93,124 @@ Inst_FLAT__FLAT_LOAD_UBYTE::completeAcc(GPUDynInstPtr gpuDynInst)
...
+void
+Inst_FLAT__FLAT_LOAD_UBYTE_D16_HI::initiateAcc(GPUDynInstPtr gpuDynInst)
+{
+    initMemRead<VecElemU8>(gpuDynInst);
+} // initiateAcc
+
+void
+Inst_FLAT__FLAT_LOAD_UBYTE_D16_HI::completeAcc(GPUDynInstPtr gpuDynInst)
+{
+    VecOperandU32 vdst(gpuDynInst, extData.VDST);
```
mattsinc, line 206: same here

src/arch/amdgpu/vega/insts/flat.cc
```diff
@@ -150,6 +267,124 @@ Inst_FLAT__FLAT_LOAD_SBYTE::completeAcc(GPUDynInstPtr gpuDynInst)
...
+void
+Inst_FLAT__FLAT_LOAD_SBYTE_D16::initiateAcc(GPUDynInstPtr gpuDynInst)
+{
+    initMemRead<VecElemI8>(gpuDynInst);
+} // initiateAcc
+
+void
+Inst_FLAT__FLAT_LOAD_SBYTE_D16::completeAcc(GPUDynInstPtr gpuDynInst)
+{
+    VecOperandU32 vdst(gpuDynInst, extData.VDST);
```
mattsinc, line 324: same here

src/arch/amdgpu/vega/insts/flat.cc
```diff
@@ -150,6 +267,124 @@ Inst_FLAT__FLAT_LOAD_SBYTE::completeAcc(GPUDynInstPtr gpuDynInst)
...
+void
+Inst_FLAT__FLAT_LOAD_SBYTE_D16_HI::initiateAcc(GPUDynInstPtr gpuDynInst)
+{
+    initMemRead<VecElemI8>(gpuDynInst);
+} // initiateAcc
+
+void
+Inst_FLAT__FLAT_LOAD_SBYTE_D16_HI::completeAcc(GPUDynInstPtr gpuDynInst)
+{
+    VecOperandU32 vdst(gpuDynInst, extData.VDST);
```
mattsinc, line 386: same here

src/arch/amdgpu/vega/insts/flat.cc
```diff
@@ -207,7 +442,124 @@ Inst_FLAT__FLAT_LOAD_USHORT::completeAcc(GPUDynInstPtr gpuDynInst)
...
+void
+Inst_FLAT__FLAT_LOAD_SHORT_D16::initiateAcc(GPUDynInstPtr gpuDynInst)
+{
+    initMemRead<VecElemU16>(gpuDynInst);
+} // initiateAcc
+
+void
+Inst_FLAT__FLAT_LOAD_SHORT_D16::completeAcc(GPUDynInstPtr gpuDynInst)
+{
+    VecOperandU32 vdst(gpuDynInst, extData.VDST);
```
mattsinc, line 505: same here

mattsinc (commented):
I'm sure I'm missing something, but it seems like the data types are wrong in a few places?

src/arch/amdgpu/vega/insts/flat.cc
abmerop: I added some comments from the isa manual for clarity above execute. For this one we read 8-bits, explicitly zero out bits 15:8, and preserve bits 31:16. So we just need to read 8-bits from memory here.

src/arch/amdgpu/vega/insts/flat.cc
abmerop: We need to read the whole register to preserve bits. Same in all other cases. Added comment above execute() with description from isa manual

mattsinc (approved):

PR #3215, opened 2026-06-11
misc: Implement CustomMesh based cache hierarchies in stdlib

giactra:
This PR adds the following CustomMesh based cache hierarchies in stdlib:

- PrivateL1PrivateL2MeshCacheHierarchy
- PrivateL1PrivateL2SharedL3MeshCacheHIerarchy

To do so the CustomMesh model in configs/topologies/CustomMesh has been ported
to the stdlib environment.
The philosophy behind these new hierarchies is the same behind the original
CustomMesh topology in configs/ruby: the user will provide a topology file
which details the placement and configuration of relevant CHI nodes in the
mesh. See for example the old 2x4.py chi-config file:

https://github.com/gem5/gem5/blob/stable/configs/example/noc_config/2x4.py

With the PR we provide the porting of the same example into the new stdlib format:

https://github.com/gem5/gem5/compare/develop...giactra:gem5:stdlib-mesh?expand=1#diff-ee2a51d9c1acecb76233b7fcfeb05dcf6a9a9adc20ceef288cde72e33a9d48af

The structure is more or less similar with the following distinctions:

1) Distinct Node_Params (CHI node specific) and NoC_Params (MESH specific) as
opposed to global NoC_Params and Node::NoC_Params
2) configuration is not based on subclassing anymore and Node_Params are
objects which get initialized with different values
3) The example file has to respect a particular interface which is
assumed by consumer scripts; the following symbols have to be defined/exposed:
**get_node_params** (to retrieve node parameters) and **cache_hierarchy_type**
(to instantiate the cache hierarchy)

The PR also contains a testlib config to be added to our CI, and a hybrid configs/example/arm
script (ruby_fs_mesh.py) which uses the new cache hierarchy (without using the ArmBoard)

### DIFF
--- a/src/mem/ruby/protocol/chi/generic/CHIGeneric.py
+++ b/src/mem/ruby/protocol/chi/generic/CHIGeneric.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2023 ARM Limited
+# Copyright (c) 2023, 2026 Arm Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
@@ -54,3 +54,24 @@ class CHIGenericController(RubyController):
     snpIn = Param.MessageBuffer("")
     rspIn = Param.MessageBuffer("")
     datIn = Param.MessageBuffer("")
+
+    def connectQueues(self, network):
+        """Connect all of the queues for this controller.
+        This may be extended in subclasses.
+        """
+        self.reqOut = MessageBuffer()
+        self.rspOut = MessageBuffer()
+        self.snpOut = MessageBuffer()
+        self.datOut = MessageBuffer()
+        self.reqIn = MessageBuffer()
+        self.rspIn = MessageBuffer()
+        self.snpIn = MessageBuffer()
+        self.datIn = MessageBuffer()
+        self.reqOut.out_port = network.in_port
+        self.rspOut.out_port = network.in_port
+        self.snpOut.out_port = network.in_port
+        self.datOut.out_port = network.in_port
+        self.reqIn.in_port = network.out_port
+        self.rspIn.in_port = network.out_port
+        self.snpIn.in_port = network.out_port
+        self.datIn.in_port = network.out_port
--- a/src/python/gem5/components/cachehierarchies/chi/nodes/abstract_node.py
+++ b/src/python/gem5/components/cachehierarchies/chi/nodes/abstract_node.py
@@ -1,3 +1,15 @@
+# Copyright (c) 2021-2024,2026 Arm Limited
+# All rights reserved.
+#
+# The license below extends only to copyright in the software and shall
+# not be construed as granting a license to any other intellectual
+# property including but not limited to intellectual property relating
+# to a hardware implementation of the functionality of the software
+# licensed hereunder.  You may use the software subject to the license
+# terms below provided that you ensure that this notice is replicated
+# unmodified and in its entirety in all distributions of the software,
+# modified or unmodified, in source code or in binary form.
+#
 # Copyright (c) 2021 The Regents of the University of California
 # All Rights Reserved.
 #
@@ -26,11 +38,18 @@
 
 import math
 from abc import abstractmethod
+from dataclasses import (
+    dataclass,
+    field,
+)
+from enum import Enum
+from typing import List
 
 from m5.objects import (
     CHI_Cache_Controller,
     MessageBuffer,
     RubyNetwork,
+    SubSystem,
 )
 
 from .....isas import ISA
@@ -53,10 +72,10 @@ class OrderedTriggerMessageBuffer(TriggerMessageBuffer):
     ordered = True
 
 
-class AbstractNode(CHI_Cache_Controller):
+class CacheController(CHI_Cache_Controller):
     """A node is the abstract unit for caches in the CHI protocol.
 
-    You can extend the AbstractNode to create caches (private or shared) and
+    You can extend the CacheController to create caches (private or shared) and
     directories with or without data caches.
     """
 
@@ -74,7 +93,7 @@ class AbstractNode(CHI_Cache_Controller):
 
         # Note: Need to call versionCount method on *this* class, not the
         # potentially derived class
-        self.version = AbstractNode.versionCount()
+        self.version = CacheController.versionCount()
         self._cache_line_size = cache_line_size
 
         # Set somewhat large number since we really a lot on internal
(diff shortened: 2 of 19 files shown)


src/mem/ruby/protocol/chi/generic/CHIGeneric.py
```diff
@@ -54,3 +54,24 @@ class CHIGenericController(RubyController):
     snpIn = Param.MessageBuffer("")
     rspIn = Param.MessageBuffer("")
     datIn = Param.MessageBuffer("")
+
+    def connectQueues(self, network):
```
powerjg, line 58: I would strongly prefer this code to be in the stdlib instead of in the SimObject declaration file. Is there a reason that it is required to be in this file?

The reasoning behind this request is that it is difficult to track the location of code and defaults when it is spread out between many different files. It is better if we have only parameters defined in the SimObject declaration files and no code.

src/python/gem5/components/cachehierarchies/chi/nodes/abstract_node.py
```diff
@@ -53,10 +72,10 @@ class OrderedTriggerMessageBuffer(TriggerMessageBuffer):
     ordered = True
 
 
-class AbstractNode(CHI_Cache_Controller):
+class AbstractController(CHI_Cache_Controller):
```
powerjg: I would much rather a different class name (e.g., `AbstractCHINode` or anything else) since `AbstractController` is already in the gem5 python namespace.

src/python/gem5/components/cachehierarchies/chi/nodes/abstract_node.py
```diff
@@ -129,3 +148,83 @@ def connectQueues(self, network: RubyNetwork):
...
+
+    def num_nodes(self) -> int:
+        """Derive node count from Node_Params placement metadata."""
+
+        if self.num_nodes_per_router < 0:
+            raise ValueError("num_nodes_per_router must be >= 0.")
+        return len(self.router_list) * self.num_nodes_per_router
+
+
+class CHI_Node(SubSystem):
```
powerjg, line 189: Why is this a `SubSystem`? Alternatively, why can't this code be in the "`AbstractController`" above?

src/python/gem5/components/cachehierarchies/chi/nodes/hnf.py
```diff
@@ -0,0 +1,73 @@
...
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+import m5
+from m5.objects import *
```
powerjg, line 38: This line is the problem with `AbstractController` Now, in this file `AbstractController` is a different class than above.

src/python/gem5/components/cachehierarchies/chi/private_l1_private_l2_mesh_cache_hierarchy.py
```diff
@@ -0,0 +1,440 @@
...
+    def incorporate_cache(self, board: AbstractBoard) -> None:
+        super().incorporate_cache(board)
+        self.ruby_system = RubySystem()
+
+        # Ruby's global network.
+        self.ruby_system.network = CustomMesh(
+            ruby_system=self.ruby_system,
+            noc_params=self._noc_params,
+        )
+        self.ruby_system.network.physical_vnets_bandwidth = [8, 8, 8, 40]
```
powerjg: Comment for this line? These magic numbers could use an explanation.

src/python/gem5/components/cachehierarchies/chi/private_l1_private_l2_mesh_cache_hierarchy.py
```diff
@@ -0,0 +1,440 @@
...
+
+            rni = CHI_RNI_DMA(self.ruby_system, ctrl)
+            rni.setDownstreamNodes(self.hnf_nodes)
+
+            rnis.append(rni)
+
+        return rnis
+
+    @overrides(AbstractRubyCacheHierarchy)
+    def _reset_version_numbers(self):
```
powerjg, line 442: Where is this function used?

powerjg (changes requested):
Generally, this is great! Thanks for putting it together, Giacomo! Just one thing and a couple of minor things below.

src/mem/ruby/protocol/chi/generic/CHIGeneric.py
giactra: The issue is that there is no CHIGenericController defined in stdlib. In stdlib this method is defined in the AbstractNode:

https://github.com/gem5/gem5/blob/stable/src/python/gem5/components/cachehierarchies/chi/nodes/abstract_node.py#L103

Either we move it out of the controller class, or we define an extra base interface and live with double inherithance...

src/python/gem5/components/cachehierarchies/chi/nodes/abstract_node.py
giactra: Ack I will think of something else

src/python/gem5/components/cachehierarchies/chi/nodes/abstract_node.py
giactra: Because a Node can have more than one controller... For example a RN can be thought as being a collection of L1D, L1I, L2 coherence controllers.

src/python/gem5/components/cachehierarchies/chi/private_l1_private_l2_mesh_cache_hierarchy.py
giactra: Here: https://github.com/gem5/gem5/blob/stable/src/python/gem5/components/cachehierarchies/ruby/abstract_ruby_cache_hierarchy.py#L50

src/python/gem5/components/cachehierarchies/chi/private_l1_private_l2_mesh_cache_hierarchy.py
giactra: Yes. 

These are [ctrl_msg_size, ..., ctrl_msg_size + data_width].

These are parametrizable once

https://github.com/gem5/gem5/pull/3146/changes

gets merged. Meaning I will port this configurability to the stdlib version as well. I will add a comment detailing it, however my hope is that we change these hardcoded values ASAP

giactra: I have actually updated the PR with params, and avoided any hardcoding

src/python/gem5/components/cachehierarchies/chi/nodes/abstract_node.py
powerjg: I see. Thanks.

src/mem/ruby/protocol/chi/generic/CHIGeneric.py
powerjg: Ok. I don't like it, but I'll accept it :)

powerjg (approved):

PR #3218, opened 2026-06-12
base: Fix ldMin/ldMax not updated in handleLoadableSegment

### LINKED ISSUE #3217: [SE mode] ldMin/ldMax never updated in ElfObject::handleLoadableSegment, causing mmap_end not to advance for dynamic linker
## Summary

In SE mode, when loading a dynamically-linked binary, `mmap_end` is not advanced
to reserve space for the dynamic linker (interpreter). This can cause subsequent
`mmap` allocations to overlap the interpreter's address region.

## Root Cause

**Files:** `src/base/loader/elf_object.cc` / `src/base/loader/elf_object.hh`

`ElfObject::handleLoadableSegment()` never updates `ldMin` and `ldMax`, even
though `mapSize()` depends on them:

```cpp
// elf_object.hh
Addr ldMin = MaxAddr;  // never updated
Addr ldMax = MaxAddr;  // never updated (also wrong initial value; should be 0)

Addr mapSize() const override { return ldMax - ldMin; }  // always returns 0
```

```cpp
// elf_object.cc
ElfObject::handleLoadableSegment(GElf_Phdr phdr, int seg_num)
{
    image.addSegment(...);
    // ... program header table handling ...

    // BUG: these two lines are missing:
    // ldMin = std::min(ldMin, phdr.p_vaddr);
    // ldMax = std::max(ldMax, phdr.p_vaddr + phdr.p_memsz);
}
```

Because `mapSize()` returns 0, `Process::updateBias()` (`src/sim/process.cc:506`)
computes `interp_mapsize = 0`, so `mmap_end` is never advanced:

```cpp
Addr interp_mapsize = roundUp(interp->mapSize(), pTable->pageSize()); // → 0
Addr mmap_end = memState->getMmapEnd();
Addr ld_bias = mmapGrowsDown() ? mmap_end - 0 : mmap_end;  // no movement
mmap_end     = mmapGrowsDown() ? ld_bias : mmap_end + 0;   // unchanged
memState->setMmapEnd(mmap_end);  // mmap_end never advances
```

## Regression Introduced By

Commit **`d856e5ce71`** — *"arch,base: Restructure the object file loaders"*
(2019-10-03, author: Gabe Black)

Before this refactoring, the constructor's PT_LOAD loop explicitly updated
`ldMin`/`ldMax`:

```cpp
// OLD constructor (correct)
ldMin(std::numeric_limits<Addr>::max()),
ldMax(std::numeric_limits<Addr>::min())
...
ldMin = std::min(ldMin, phdr.p_vaddr);
ldMax = std::max(ldMax, phdr.p_vaddr + phdr.p_memsz);
```

When the loop body was extracted into `handleLoadableSegment()`, these two
lines were accidentally dropped.

## Proposed Fix

In `src/base/loader/elf_object.cc`, add to `handleLoadableSegment()`:

```cpp
ElfObject::handleLoadableSegment(GElf_Phdr phdr, int seg_num)
{
    if (phdr.p_memsz == 0) {
        warn("Ignoring empty loadable segment %s", name);
        return;
    }

    image.addSegment(...);
    // ...

    // Add these two lines:
    ldMin = std::min(ldMin, phdr.p_vaddr);
    ldMax = std::max(ldMax, phdr.p_vaddr + phdr.p_memsz);
}
```

And fix the initial value in `src/base/loader/elf_object.hh`:

```cpp
Addr ldMin = MaxAddr;  // OK
Addr ldMax = 0;        // was MaxAddr — should be 0 for correct max-tracking
```

## Impact

SE mode dynamic binary execution: interpreter (`ld-linux.so`) may be placed at
an address that overlaps with subsequent `mmap` allocations.

nahj135:
## Summary

In SE mode, `ElfObject::handleLoadableSegment()` never updates `ldMin`
and `ldMax`, causing `mapSize()` to always return 0. As a result,
`Process::updateBias()` computes `interp_mapsize=0` and `mmap_end` is
never advanced to reserve space for the dynamic linker. This can cause
subsequent `mmap` allocations to overlap the interpreter's address region.

Fixes #3217

## Root Cause

The bug was introduced as a regression in `d856e5ce71` ("arch,base:
Restructure the object file loaders"), which extracted the PT_LOAD
loop body into `handleLoadableSegment()` but accidentally dropped the
two lines that tracked `ldMin`/`ldMax`.

Before the refactoring, the constructor explicitly set these:
```cpp
ldMin(std::numeric_limits<Addr>::max()),
ldMax(std::numeric_limits<Addr>::min())
...
ldMin = std::min(ldMin, phdr.p_vaddr);
ldMax = std::max(ldMax, phdr.p_vaddr + phdr.p_memsz);
```

## Changes

**`src/base/loader/elf_object.cc`** — add the missing two lines to
`handleLoadableSegment()`:
```cpp
ldMin = std::min(ldMin, phdr.p_vaddr);
ldMax = std::max(ldMax, phdr.p_vaddr + phdr.p_memsz);
```

**`src/base/loader/elf_object.hh`** — fix the initial value of `ldMax`:
```cpp
Addr ldMax = 0;  // was MaxAddr; should be 0 for correct max-tracking
```

## Test Plan

- [ ] SE mode: run a dynamically-linked binary and confirm `mmap_end`
      advances by the interpreter's actual size after `updateBias()`
- [ ] SE mode: confirm interpreter load address does not overlap with
      subsequent `mmap` allocations

### DIFF
--- a/src/base/loader/elf_object.hh
+++ b/src/base/loader/elf_object.hh
@@ -94,7 +94,7 @@ class ElfObject : public ObjectFile
     // The ldMin and ldMax fields are required to know how large of an
     // area is required to map the interpreter.
     Addr ldMin = MaxAddr;
-    Addr ldMax = MaxAddr;
+    Addr ldMax = 0;
 
     /// Helper functions for loadGlobalSymbols() and loadLocalSymbols().
     bool loadSomeSymbols(SymbolTable *symtab, int binding, Addr mask,
@@ -116,7 +116,14 @@ class ElfObject : public ObjectFile
 
     Addr bias() const override { return ldBias; }
     bool relocatable() const override { return relocate; }
-    Addr mapSize() const override { return ldMax - ldMin; }
+    Addr
+    mapSize() const override
+    {
+        // ldMin/ldMax stay at their sentinel values if no non-empty
+        // PT_LOAD segment was seen. Returning ldMax - ldMin in that case
+        // would underflow, so report an empty mapping instead.
+        return (ldMin <= ldMax) ? (ldMax - ldMin) : 0;
+    }
     void updateBias(Addr bias_addr) override;
 
     bool hasTLS() override { return sectionExists(".tbss"); }
--- a/src/base/loader/elf_object.cc
+++ b/src/base/loader/elf_object.cc
@@ -46,6 +46,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#include <algorithm>
 #include <cassert>
 #include <string>
 
@@ -391,6 +392,9 @@ ElfObject::handleLoadableSegment(GElf_Phdr phdr, int seg_num)
         image.addSegment({ name + "(uninitialized)",
                            phdr.p_paddr + phdr.p_filesz, uninitialized });
     }
+
+    ldMin = std::min(ldMin, phdr.p_vaddr);
+    ldMax = std::max(ldMax, phdr.p_vaddr + phdr.p_memsz);
 
     const Addr file_start = phdr.p_offset;
     const Addr file_end = file_start + phdr.p_filesz;


powerjg (dismissed):

erin-le:
@nahj135: Hello, could you address the Copilot comment and add the include for `<algorithm>`? Thank you!

nahj135:
@erin-le Thanks for the review! Both points are addressed in the latest push (`be529a9`):

1. **`#include <algorithm>`** — added explicitly in `elf_object.cc` for `std::min`/`std::max`.
2. **Copilot comment (`mapSize()` underflow)** — `mapSize()` now guards against the sentinel case. If no non-empty `PT_LOAD` segment is seen, `ldMin`/`ldMax` stay at their initial values and `ldMax - ldMin` would underflow; it now returns 0 instead:

```cpp
Addr mapSize() const override
{
    // ldMin/ldMax stay at their sentinel values if no non-empty
    // PT_LOAD segment was seen. Returning ldMax - ldMin in that case
    // would underflow, so report an empty mapping instead.
    return (ldMin <= ldMax) ? (ldMax - ldMin) : 0;
}
```

Verified `elf_object.cc` compiles cleanly. PTAL, thanks!

erin-le (dismissed):

src/base/loader/elf_object.hh
```diff
@@ -116,7 +116,13 @@ class ElfObject : public ObjectFile
 
     Addr bias() const override { return ldBias; }
     bool relocatable() const override { return relocate; }
-    Addr mapSize() const override { return ldMax - ldMin; }
+    Addr mapSize() const override
```
erin-le: @nahj135: It looks like the clang format check wants this change:
```suggestion
    Addr
    mapSize() const override
```

erin-le (approved):

PR #3227, opened 2026-06-16
mem: Fix Formula calculation for DRAM interface

giactra:
The following stats are reported as showing the bandwidth in mebibytes per second (MiB/s). However the really report (MB/s) instead.

        ADD_STAT(avgRdBW, statistics::units::Rate<
                 "Average DRAM read bandwidth in MiBytes/s"),
        ADD_STAT(avgWrBW, statistics::units::Rate<
                 "Average DRAM write bandwidth in MiBytes/s"),
        ADD_STAT(peakBW,  statistics::units::Rate<
                 "Theoretical peak bandwidth in MiByte/s"),

Change-Id: Iced3268de329d508a1a245d2dfc323c0c533bb6a

### DIFF
--- a/src/mem/dram_interface.cc
+++ b/src/mem/dram_interface.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010-2020 ARM Limited
+ * Copyright (c) 2010-2020, 2026 Arm Limited
  * All rights reserved
  *
  * The license below extends only to copyright in the software and shall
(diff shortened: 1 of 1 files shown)


odanrc (approved):
We don't have a unit called MebiByte, so isn't the actual unit clashing with the unit displayed by the corresponding str dump (statistics::units::Byte::getUnitString())?

powerjg (changes requested):
I don't think we want this change. It has been known that the bandwidth is in MB not MiB. This would break many users' understanding of this data.

I believe it would have been better if this was MiB from the beginning of time, but it's been MB for too long to change now.

We should change the documentation instead of the data.

giactra:
> We don't have a unit called MebiByte, so isn't the actual unit clashing with the unit displayed by the corresponding str dump (statistics::units::Byte::getUnitString())?

Yes sort of. The output is something like:

`system.memory.mem_ctrl1.dram.peakBW          19207.00                       # Theoretical peak bandwidth in MiByte/s ((Byte/Second)) `

So one one hand the unit is byte/second, on the other the comment states the value is actually MiB/s

> We should change the documentation instead of the data.

Makes sense, I have simply updated the BW stats descriptions to remove the "i"

abmerop:
If I understand the change, this would now print this:

`system.memory.mem_ctrl1.dram.peakBW          19207.00                       # Theoretical peak bandwidth in MByte/s ((Byte/Second))`

The `Byte/Second` is still conflicting/confusing. It looks like there is no MB unit... But perhaps while changing this we could fix that as well? Or start another PR which adds some new common units and updates the whole repo?

powerjg:
IMO, all formula should be abolished. 

My vote for the most straightforward path is to update the documentation here to say "10^6 B/s". Whoever added this stat (@ahansson-arm? I am confident that handle isn't used anymore 😉) has caused us a lot of confusion!  

I think the "Bytes / second" unit is OK. It *is* a unit of bytes divided by seconds, it just also has a (large) scalar factor.

powerjg (approved):

powerjg:
@abmerop is this OK with you?

PR #3234, opened 2026-06-17
arch-x86,stdlib: More x86 ACPI support + Updates for ROCm 7.2

abmerop:
This PR adds support needed to boot ROCm 7.2 with Linux 6.17. Current GPU disk images use a workaround to load amdgpu driver due to a hard requirement of basic ACPI support in the WMI module. This workaround no longer is possible (something to do with how Module.symver is calculated with Linux' new rust support). Therefore, we need *actual* minimal ACPI support. For the WMI module to be happy and not disable ACPI, we need at least DSDT, which gets pointed to by an FADT table. This PR adds the following:

- ACPI FADT and DSDT support with very minimal features
- Removal of load_amdgpu.sh workaround script
- Change stdlib files to check for load_amdgpu.sh anyway to support ROCm 7.0 disk images for now.
- Remove check for amdgpu.ko, *deprecating ROCm 6.4 and earlier disk images*
- Modprobe amdgpu driver directly in stdlib instead of load_amdgpu.sh
- Fix debug option in amdgpu.py which was not exposed anywhere from user configs
- Add option for user provided IP discovery file while I am modifying the modprobe lines
- New disks will use gem5-bridge, so remove calls to /sbin/m5 and use m5

ROCm 7.2 MI300X and MI355X now boot in gem5 and can run apps.


src/python/gem5/components/devices/gpus/amdgpu.py
```diff
@@ -190,7 +210,7 @@ def get_driver_command(self, debug: bool = False):
             "    sh /home/gem5/load_amdgpu.sh\n"
             "elif [ ! -f /lib/modules/`uname -r`/updates/dkms/amdgpu.ko ]; then\n"
             '    echo "ERROR: Missing DKMS package for kernel `uname -r`. Exiting gem5."\n'
-            "    /sbin/m5 exit\n"
```
mattsinc, line 193: does this change have anything to do with the commit?  Seems unrelated ...

mattsinc (approved):
I don't feel like I deeply understand this code as I'm not an x86 expert, but it seemed ok to me, minus one question.

mattsinc:
@xsu-757, @v-ramadas : FYI

src/python/gem5/components/devices/gpus/amdgpu.py
abmerop: Part of the "getting ROCm 7.2 working" is the new disk image. I added the gem5-bridge stuff (I think) to the disk image and /sbin/m5 no longer exists. I think it is now at /usr/local/bin/m5.

I will post that soon, running more test right now, but it's staged here: https://github.com/abmerop/gem5-resources/commit/fc8cb8cbaf31526c5678f562a61e5479cb0c511b

PR #3240, opened 2026-06-19
arch-riscv: Add AT_BASE to the SE-mode auxiliary vector

### LINKED ISSUE #3239: arch-riscv: SE-mode auxv is missing AT_BASE, breaking dynamic linker self-relocation
**Affected version:** v25.1.0.1 (also present since the feature was introduced)
**ISA:** RISC-V, SE (syscall-emulation) mode

### Summary
In RISC-V SE mode, the auxiliary vector built for a dynamically-linked process omits `AT_BASE`. As a result the dynamic linker (`ld.so`) cannot determine its own load bias and performs self-/relative relocations with a base of `0`, so relocated pointers hold raw link-time offsets instead of `base + addend`. Dereferencing such a pointer (e.g. data reached via a `dlsym`'d symbol) faults.

### Location
`src/arch/riscv/process.cc`, the auxv population block (`RiscvProcess::argsInit`). The vector contains `Entry / Phnum / Phent / Phdr / Pagesz / Secure / Random / Null` but no `Base`.

### Evidence — RISC-V is the only ISA missing it
`grep auxv::Base src/arch/*/process.cc`:

| ISA | `auxv::Base` (AT_BASE) |
|---|---|
| x86 | yes |
| arm | yes |
| sparc | yes |
| power | yes |
| mips | yes |
| **riscv** | **no** |

All other ISAs emit `auxv.emplace_back(gem5::auxv::Base, getBias());` right after `Phdr`.

### Symptom
A dynamically-linked RISC-V binary that loads a shared object segfaults; under a debugger, a relocated data pointer points at a raw section offset rather than `base + offset`, consistent with `ld.so` self-relocating with base `0`.

### Suggested fix
Mirror the other ISAs by adding, after the `Phdr` entry:
```cpp
auxv.emplace_back(gem5::auxv::Base, getBias());
```
(Note: gem5 generally recommends `-static` for SE mode, so dynamic linking may hit further gaps such as unimplemented syscalls afterwards — but the missing `AT_BASE` is a clear, isolated inconsistency vs. the other ISAs.)

nahj135:
### Summary
In RISC-V SE mode, the auxiliary vector built by `RiscvProcess::argsInit`
(`src/arch/riscv/process.cc`) omitted `AT_BASE`. Without it the dynamic linker
(`ld.so`) cannot determine its own load bias and performs relative/self
relocations against a base of `0`, so relocated pointers end up holding raw
link-time offsets instead of `base + addend`. Dereferencing such a pointer
(e.g. data reached via a `dlsym`'d symbol in a shared object) faults.

### Why
RISC-V was the only ISA missing this entry — `x86`, `arm`, `sparc`, `power`
and `mips` all emit `auxv::Base` with `getBias()` right after `Phdr`:

```
$ grep -L auxv::Base src/arch/*/process.cc
src/arch/riscv/process.cc      # the only one
```

### Change
Add `auxv.emplace_back(gem5::auxv::Base, getBias());` after the `Phdr` entry,
mirroring the other ISAs, and register `Base` in the `Stack` debug name map.

### Notes
gem5 generally recommends `-static` for SE mode, so dynamic linking may still
hit further gaps afterwards (e.g. unimplemented syscalls). This patch only
fixes the isolated, clear inconsistency vs. the other ISAs.

Closes #3239

### DIFF
--- a/src/arch/riscv/process.cc
+++ b/src/arch/riscv/process.cc
@@ -156,6 +156,7 @@ RiscvProcess::argsInit(int pageSize)
         auxv.emplace_back(gem5::auxv::Phnum, elfObject->programHeaderCount());
         auxv.emplace_back(gem5::auxv::Phent, elfObject->programHeaderSize());
         auxv.emplace_back(gem5::auxv::Phdr, elfObject->programHeaderTable());
+        auxv.emplace_back(gem5::auxv::Base, getBias());
         auxv.emplace_back(gem5::auxv::Pagesz, PageBytes);
         auxv.emplace_back(gem5::auxv::Secure, 0);
         auxv.emplace_back(gem5::auxv::Random, stack_top);
@@ -242,6 +243,7 @@ RiscvProcess::argsInit(int pageSize)
         {gem5::auxv::Phnum, "gem5::auxv::Phnum"},
         {gem5::auxv::Phent, "gem5::auxv::Phent"},
         {gem5::auxv::Phdr, "gem5::auxv::Phdr"},
+        {gem5::auxv::Base, "gem5::auxv::Base"},
         {gem5::auxv::Pagesz, "gem5::auxv::Pagesz"},
         {gem5::auxv::Secure, "gem5::auxv::Secure"},
         {gem5::auxv::Random, "gem5::auxv::Random"},


powerjg (approved):

pranith:
@powerjg @erin-le There was a clang format check fail in the checks here which we missed. This is causing fails for the other PRs. Can we please make sure not to merge changes that fail these checks? Thanks!

erin-le:
Hey @pranith, sorry about that. I've gone ahead and approved your clang-format PR. Tagging @BobbyRBruce here as well, since he told me to merge this with the failing check, as the change that clang-format wanted was fairly minor and unconnected to the actual change.

Theoretically, the clang-format check should only check commits on the current PR, but there have been numerous instances where that hasn't been the case. In case it's useful for you in the future, here's a [link](https://github.com/orgs/gem5/discussions/3201#discussioncomment-17242898) to Jason's guide on how to get the checks to pass.

PR #3241, opened 2026-06-19
stats: add runtime lookup of single values from multi-value stats

### LINKED ISSUE #3235: Reading gem5 stats during simulation, for live power modeling
# Reading gem5 stats during simulation, for live power modeling

Hello, 

I'm writing a power model that needs to read gem5 statistics *while the simulation is running*, instead of waiting for the stats file at the end. To do this, I need to look up a stat by its name and get its current value during the run.

The values I need to read are:

- Simple stats, like `ipc` or `simSeconds`
- A single value out of a stat that holds many values, written as `name::subname` (for example `overallAccesses::cpu.data`)
- A few extra values that help with power calculations: `clock_period` (the chip's speed), and `temp` / `voltage` (placeholders for now)

## What I built so far

gem5 currently has no way to read a single value out of a multi-value stat at runtime, so I implemented one using a `name::subname` syntax. It works like this: you give it an expression, it finds the stat names inside it, looks each one up with `statistics::resolve()`, and returns the current value whenever asked. If a stat holds many values and you ask for `name::subname`, it returns just the matching one.

It's only about 20 lines of code.

I've attached the file I changed.

## Questions

1. Would gem5 want this built in, or is it better kept as a separate add-on?

I have a working prototype and I'm happy to clean it up into a patch if this sounds like a good direction.

HugoSipearl:
Add a way to read gem5 statistics by name while the simulation is running, for use in live power and thermal models that need stat values during the run rather than from the stats file at the end. Github Issue : #3235

### DIFF
--- a/src/sim/mathexpr.cc
+++ b/src/sim/mathexpr.cc
@@ -136,7 +136,7 @@ MathExpr::parse(std::string expr) {
                 !( (c >= 'a' && c <= 'z') ||
                    (c >= 'A' && c <= 'Z') ||
                    (c >= '0' && c <= '9') ||
-                   c == '$' || c == '\\' || c == '.' || c == '_');
+                   c == '$' || c == '\\' || c == '.' || c == '_' || c == ':');
 
         if (!contains_non_alpha) {
             Node * n = new Node();
--- a/src/sim/mathexpr.cc
+++ b/src/sim/mathexpr.cc
@@ -132,11 +132,11 @@ MathExpr::parse(std::string expr) {
     {
         bool contains_non_alpha = false;
         for (auto & c: expr)
-            contains_non_alpha = contains_non_alpha or
-                !( (c >= 'a' && c <= 'z') ||
-                   (c >= 'A' && c <= 'Z') ||
-                   (c >= '0' && c <= '9') ||
-                   c == '$' || c == '\\' || c == '.' || c == '_' || c == ':');
+            contains_non_alpha =
+                contains_non_alpha or
+                !((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') ||
+                  (c >= '0' && c <= '9') || c == '$' || c == '\\' ||
+                  c == '.' || c == '_' || c == ':');
 
         if (!contains_non_alpha) {
             Node * n = new Node();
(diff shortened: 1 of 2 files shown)


skku970412:
I took a look from the testing/edge-case side. The PR is small, and the `name::subname` direction matches the live power-modeling use case in #3235, but I think a few cases need tests before this is safe to merge:

- If a requested `name::subname` is not found, `getStatValue()` currently falls back to `total()`. That means a typo in a subname can silently evaluate to an aggregate value instead of failing.
- A scalar stat requested as `scalar_stat::subname` currently returns the scalar value. I think that should fail clearly, since `::subname` was explicit.
- The implementation only casts multi-value stats to `FormulaInfo`, while `VectorInfo` also has `subnames`, `result()`, and `total()`. If FormulaInfo-only support is intentional, it would be good to document/test that limitation; otherwise this probably wants to handle `VectorInfo` generally.
- `fi->subnames.size() >= 2` skips lookup for one-element stats even if a valid subname was explicitly requested.
- `MathExpr` now accepts any `:` in variable names, but only `::` has semantics downstream. Malformed expressions like `foo:bar`, `foo::`, or `foo:::bar` should probably fail with a clear message.

Suggested focused tests:

- scalar lookup succeeds
- `name::subname` lookup succeeds for a multi-value stat
- missing stat name fails clearly
- missing subname fails clearly
- `scalar_stat::subname` fails clearly
- malformed colon syntax fails clearly

Also, the current CI has `clang-format-check` failing, which looks consistent with the formatting around the new branches. I could not reproduce the exact clang-format diff locally because this checkout does not have `git-clang-format`, but the CI failure is the only failing check I see.

powerjg (dismissed):

erin-le:
@HugoSipearl: Hello, for the failing clang-format check, would you mind if I pushed a formatting commit to this PR? If you'd like to make the change yourself, you can use the following command, which is what the clang-format check runs:
```
python3 util/run-git-clang-format.py --verbose --ci-pr-base-commit 79800f90a1c93139f102cd0098bcd150cd0af6bb
```

HugoSipearl:
Hello @erin-le,
You can push it. There’s no problem. :)

erin-le (approved):

BobbyRBruce (approved):

PR #3242, opened 2026-06-20
stdlib: Re-enable x86-64-v2 in VIPER board

abmerop:
Latest pytorch builds are failing with an error saying something about a module being imported twice. The real issue is that numpy fails to load because the x86 microarchitecture level is too low. Numpy is required to load pytorch, which we claim to support. This commit bumps the level to v2. Only the SSE 4.1 and SSE 4.2 bits needed to be changed. There is a nice table of the requirements for each micro-architectural level at https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels. VIPER board now has everything for v1 and v2.

### DIFF
--- a/src/python/gem5/prebuilt/viper/board.py
+++ b/src/python/gem5/prebuilt/viper/board.py
@@ -144,8 +144,12 @@ class ViperBoard(X86Board):
         # This modifies the default value for ECX only (4th in this array).
         # See: https://sandpile.org/x86/cpuid.htm#level_0000_0001h
         # Enables AVX, OSXSAVE, XSAVE, POPCNT, SSE4.2, SSE4.1, CMPXCHG16B,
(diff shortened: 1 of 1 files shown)


powerjg (approved):
Are you saying we implement all instructions for v2 now? Is this true if the timing models? If so, we should apply this to the SE mode configs as it would simplify working with glibc!

abmerop:
> Are you saying we implement all instructions for v2 now? Is this true if the timing models? If so, we should apply this to the SE mode configs as it would simplify working with glibc!

This still only applies to KVM CPUs. I've have to check what gem5 instructions are missing for this support. I'm considering punting the work to an agent soon-ish

mattsinc (commented):
sorry I missed this one before it was merged

PR #3243, opened 2026-06-21
cpu-o3: Delay misc reg dependent wake until commit

pranith:
O3 already models non-serializing misc registers with fixed PhysRegIds so
that same-register readers can wait on older writers in the IQ dependency
graph. However, the old wakeup point was still too early for misc
registers.

A misc register write does not update the architectural misc register file
when the instruction executes. DynInst::setMiscReg() records the write in
the dynamic instruction, and commit later makes the value visible through
updateMiscRegs(). If IEW writeback wakes dependents for that misc
destination, a younger same-register read can issue after the writer
executed but before the writer committed. The read then goes through
readMiscRegOperand(), which reads the still-stale architectural misc
register file.

This allowed sequences such as msr tpidr_el0 followed by mrs tpidr_el0 to
observe the old thread-pointer value under O3. The instructions still
retired in program order, but the younger read had already captured the
stale value and dependent TLS address calculations could fault before
retirement repaired architectural state.

Keep normal destination wakeups at IEW writeback, but skip misc-register
destinations there. After commit applies updateMiscRegs(), wake the IQ
dependents for committed misc-register destinations and update the rename
scoreboard. This preserves per-misc-register RAW ordering without making
every system register globally serializing.

### DIFF
--- a/src/cpu/o3/iew.cc
+++ b/src/cpu/o3/iew.cc
@@ -578,6 +578,7 @@ void
 IEW::wakeCommittedMiscRegDependents(const DynInstPtr &inst)
 {
     int dependents = 0;
+    const bool update_producer = inst->lastWakeDependents == -1;
 
     DPRINTF(IEW,
             "Waking misc register dependents of committed instruction.\n");
@@ -599,7 +600,9 @@ IEW::wakeCommittedMiscRegDependents(const DynInstPtr &inst)
 
     if (dependents) {
         ThreadID tid = inst->threadNumber;
-        iewStats.producerInst[tid]++;
+        if (update_producer) {
+            iewStats.producerInst[tid]++;
+        }
         iewStats.consumerInst[tid] += dependents;
     }
 }
--- a/src/cpu/o3/iew.cc
+++ b/src/cpu/o3/iew.cc
@@ -575,6 +575,36 @@ IEW::wakeDependents(const DynInstPtr& inst)
 }
 
 void
+IEW::wakeCommittedMiscRegDependents(const DynInstPtr &inst)
+{
+    int dependents = 0;
+
+    DPRINTF(IEW,
+            "Waking misc register dependents of committed instruction.\n");
+
+    assert(!inst->isSquashed());
+
+    for (int i = 0; i < inst->numDestRegs(); i++) {
+        PhysRegIdPtr dest_reg = inst->renamedDestIdx(i);
+        if (!dest_reg->is(MiscRegClass)) {
+            continue;
+        }
+
+        dependents += instQueue.wakeDependents(dest_reg);
+
+        DPRINTF(IEW, "Setting committed misc register %i (%s)\n",
+                dest_reg->index(), dest_reg->className());
+        scoreboard->setReg(dest_reg);
+    }
+
+    if (dependents) {
+        ThreadID tid = inst->threadNumber;
+        iewStats.producerInst[tid]++;
+        iewStats.consumerInst[tid] += dependents;
+    }
+}
+
+void
 IEW::rescheduleMemInst(const DynInstPtr& inst)
 {
     instQueue.rescheduleMemInst(inst);
@@ -1422,20 +1452,27 @@ IEW::writebackInsts()
             int dependents = instQueue.wakeDependents(inst);
 
             for (int i = 0; i < inst->numDestRegs(); i++) {
+                PhysRegIdPtr dest_reg = inst->renamedDestIdx(i);
+                if (dest_reg->is(MiscRegClass)) {
+                    continue;
+                }
+
                 // Mark register as ready if not pinned
-                if (inst->renamedDestIdx(i)->
-                        getNumPinnedWritesToComplete() == 0) {
-                    DPRINTF(IEW,"Setting Destination Register %i (%s)\n",
-                            inst->renamedDestIdx(i)->index(),
-                            inst->renamedDestIdx(i)->className());
-                    scoreboard->setReg(inst->renamedDestIdx(i));
+                if (dest_reg->getNumPinnedWritesToComplete() == 0) {
+                    DPRINTF(IEW, "Setting Destination Register %i (%s)\n",
+                            dest_reg->index(), dest_reg->className());
+                    scoreboard->setReg(dest_reg);
                 }
             }
 
+            // Misc-register dependents are excluded here. They are woken and
+            // accounted for after the producer commits its misc-register
+            // updates.
             if (dependents) {
                 iewStats.producerInst[tid]++;
-                iewStats.consumerInst[tid]+= dependents;
+                iewStats.consumerInst[tid] += dependents;
             }
+
             iewStats.writebackCount[tid]++;
         }
     }
--- a/src/cpu/o3/inst_queue.cc
+++ b/src/cpu/o3/inst_queue.cc
@@ -1209,8 +1209,6 @@ InstructionQueue::wakeDependents(const DynInstPtr &completed_inst)
         iqIOStats.intInstQueueWakeupAccesses++;
     }
 
-    completed_inst->lastWakeDependents = curTick();
-
     DPRINTF(IQ, "Waking dependents of completed instruction.\n");
 
     assert(!completed_inst->isSquashed());
@@ -1266,6 +1264,10 @@ InstructionQueue::wakeDependents(const DynInstPtr &completed_inst)
         }
 
         dependents += wakeDependents(dest_reg);
+    }
+
+    if (dependents) {
+        completed_inst->lastWakeDependents = curTick();
     }
 
     return dependents;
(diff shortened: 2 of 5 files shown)


src/cpu/o3/inst_queue.cc
```diff
@@ -1219,6 +1199,63 @@ InstructionQueue::wakeDependents(const DynInstPtr &completed_inst)
...
+int
+InstructionQueue::wakeCommittedMiscRegDependents(
+    const DynInstPtr &completed_inst)
+{
+    DPRINTF(IQ, "Waking misc register dependents of committed instruction.\n");
+
+    assert(!completed_inst->isSquashed());
+
+    return completeDestRegs(completed_inst, true, false, true);
+}
```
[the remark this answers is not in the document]
pranith: Fixed it in the latest push.

src/cpu/o3/inst_queue.cc
```diff
@@ -1120,69 +1120,49 @@ InstructionQueue::commit(const InstSeqNum &inst, ThreadID tid)
 }
 
 int
-InstructionQueue::wakeDependents(const DynInstPtr &completed_inst)
+InstructionQueue::completeDestRegs(const DynInstPtr &completed_inst,
+                                   bool misc_regs, bool complete_pinned_writes,
```
zhzhqian: can we replace these bool parameter to flags, for better understanding

src/cpu/o3/inst_queue.cc
```diff
@@ -1219,6 +1199,72 @@ InstructionQueue::wakeDependents(const DynInstPtr &completed_inst)
...
+InstructionQueue::wakeDependents(const DynInstPtr &completed_inst)
+{
+    // The instruction queue here takes care of both floating and int ops
+    if (completed_inst->isFloating()) {
+        iqIOStats.fpInstQueueWakeupAccesses++;
+    } else if (completed_inst->isVector()) {
+        iqIOStats.vecInstQueueWakeupAccesses++;
+    } else {
+        iqIOStats.intInstQueueWakeupAccesses++;
+    }
```
zhzhqian, line 1210: Would the statistics calculated here duplicated with those in the function wakeCommittedMiscRegDependents?

zhzhqian:
It's impressive to see this bug fixed with just over 100 lines of changes—really nice work! I've added a few comments on parts where I could use some clarification,  Happy to discuss further!

src/cpu/o3/inst_queue.cc
pranith: Can you explain how the flags arg should be structured?

zhzhqian: something like this
```c++
 enum CompleteFlags {
      MiscRegs = 1,
      CompletePinnedWrites = 2,
      WakeDependents = 4
  };
  completeDestRegs(completed_inst, CompletePinnedWrites | WakeDependents);
```

pranith: Thanks for the suggestion @zhzhqian. Since we only have limited set of options, I am not sure its an improvement over the current design.

src/cpu/o3/inst_queue.cc
pranith: Yes, this would duplicate it for misc register wakeups. I removed this change so that we count it only once at wakeup for both misc and non-misc register wakeups.

erin-le:
@giactra: Hello, just in case you missed it, the author of this PR pinged you from https://github.com/gem5/gem5/pull/3158, saying that this PR implements your suggestion. Could you take a look at this PR when you get a chance? Thank you!

pranith:
@giactra @powerjg @BobbyRBruce Ping! Please include this in the next release as the SE mode is broken for ARM64.

src/cpu/o3/inst_queue.cc
```diff
@@ -1219,6 +1199,63 @@ InstructionQueue::wakeDependents(const DynInstPtr &completed_inst)
...
+        // Completes a non mem ref barrier
+        memDepUnit[tid].completeInst(completed_inst);
+    }
+
+    int dependents = completeDestRegs(completed_inst, false, true, true);
+
+    // Misc register writes are deferred until commit updates architectural
+    // state. Complete their pinned-write accounting at writeback, but do not
+    // wake same-register readers until the commit-time path below.
+    completeDestRegs(completed_inst, true, true, false);
```
giactra: I find this a bit hard to read (the sequence of boolean values + the completeDestRegs implementation) and potentially less efficient than my proposal below.

I would consider changing completeDestRegs to bring the loop out to the caller. So rather than doing this:

```
 for (int dest_reg_idx = 0;
         dest_reg_idx < completed_inst->numDestRegs();
         dest_reg_idx++)
    {
        PhysRegIdPtr dest_reg =
            completed_inst->renamedDestIdx(dest_reg_idx);

        if (dest_reg->is(MiscRegClass) != misc_regs) {
            continue;
        }

        // Serializing fixed-mapping registers do not use the IQ dependency
        // graph. Non-serializing misc registers use the graph, but their
        // consumer wakeup is delayed until commit makes the write visible.
        if (dest_reg->isAlwaysReady()) {
            DPRINTF(IQ, "Reg %d [%s] is part of a fix mapping, skipping\n",
                    dest_reg->index(), dest_reg->className());
            continue;
        }
```

where we loop over the same destination register twice, firstly to check for non-misc regs, and then to check misc regs, and just skipping loop iterations

We do instead:

```

 for (int dest_reg_idx = 0;
         dest_reg_idx < completed_inst->numDestRegs();
         dest_reg_idx++)
    {
        PhysRegIdPtr dest_reg =
            completed_inst->renamedDestIdx(dest_reg_idx);

        if (!dest_reg->is(MiscRegClass)) {
            completeDestRegs(completed_inst, dest_reg)
        }
```

In this way we only call completeDestRegs once for registers which really need to be "completed".
On that note, I am tempted to say that we should strive to rename completeDestRegs into wakeDependents, and let overloading make it work. I don't have a strong bias against completeDestRegs, but effectively it is not clear what it means to "complete" a register. Instead if we had

wakeDependents(DynInstruction inst)
wakeDependents(PhysRegId dest_reg)

It is IMHO more intuitive that what we are doing in the first method is to wake up dependent instructions, and in the second one we are focusing on dependent registers.

There is one issue left behind: pinned registers. Well, register pinning is a feature that only makes sense for renameable registers. it's there to prevent un-necessary renaming. MiscRegs cannot be renamed. So there is no way to pin a MiscReg and current code only uses register pinning for vector instructions.

I don't think we are enforcing it explicitly at the moment in the code but I am pretty sure calling completeDesRegs to handle register pinning for MIscRegs won't do anything.

src/cpu/o3/iew.cc
giactra, line 1479: Aren't we calling this twice for MiscRegs? Once here and once within wakeCommittedMiscRegDependents ?

giactra (commented):
Hi @pranith many thanks for this PR and apologies for the late review (to my defense I was on vacation for some of the time). I definitely was planning to include this in v26.0.
I find it an intuitive implementation so kudo to that. Having said that, I believe we can restructure things to make them even more intuitive. See my comments below

src/cpu/o3/inst_queue.cc
pranith: Thanks for the review @giactra. I updated it as you suggested. Looks much cleaner!

src/cpu/o3/iew.cc
pranith: The writebackCount is not updated for misc-register wakeup. Added a comment for the producer/consumer case clarifying that it isn't duplicate.

src/cpu/o3/inst_queue.cc
giactra: Thanks @pranith looks great!

src/cpu/o3/iew.cc
giactra: It's not the writebackCount that concerns me, is the producerInst update.


            if (dependents) {
                iewStats.producerInst[tid]++;
                iewStats.consumerInst[tid]+= dependents;
            }


Effectively in the (unlikely I reckon) case of an instruction writing to both GPR and MiscReg, the producerInst stat is incremented twice

1. writebackInsts
2. wakeCommittedMiscRegDependents

pranith: It looks like we need to add instruction metadata to track if the instruction has woken its dependents. We set this flag once so that the second time we can skip the stat update. What do you think?

    if (!inst->hasWokenDependents()) {
        iewStats.producerInst[tid]++;
        inst->setWokeDependents();
    }

pranith: As you mentioned, it unlikely that an instruction writes to both GPR and MiscReg. Maybe just add a comment mentioning this?

giactra: There is a smalll ugly trick we could use. In wakeCommittedMiscRegDependents, before we call wakeDependents, we check if lastWakeDependents == - 1. If it doesn't, the instruction already produced values for consumers and we don't update the producer stat

pranith: @giactra added the change as suggested. Please take a look!

giactra (approved):
Thanks @pranith !

PR #3255, opened 2026-06-24
mem-ruby: Fix MII→I assert when local directory entry still exists

Deconsts:
As noted in the prior issue, state MII previously incorrectly invoked removeFromDir when no local directory entry was present (https://github.com/gem5/gem5/pull/1944).

However, the existence of the directory entry for MII is non-deterministic, as this state can be reached via different paths:
Directory entry exists: OLSX → OLSI → ILSI → MII
Directory entry does not exist: M → MI → MII
Previously removing removeFromDir may cause _setState(I)_ asserts when MII goes back to I from the first path.

Splitting the MII state might be a better fix, which aligns with SLICC conventions.

### DIFF
--- a/src/mem/ruby/protocol/MOESI_CMP_directory-L2cache.sm
+++ b/src/mem/ruby/protocol/MOESI_CMP_directory-L2cache.sm
@@ -1586,6 +1586,13 @@ machine(MachineType:L2Cache, "Token protocol")
     localDirectory.deallocate(address);
   }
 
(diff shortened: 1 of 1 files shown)


erin-le:
@Deconsts: Could you change the target branch of this PR from `gem5:stable` to `gem5:develop`, and also rebase or cherry-pick your change onto current `develop`? Thanks!

powerjg (approved):

erin-le:
pre-commit.ci run

PR #3261, opened 2026-06-24
stdlib: fix Addr comparison by casting to int

erin-le:
Previously, this comparison was failing because it was comparing an Addr and an int. This commit casts the Addr to an int to fix the comparison, and fix the test failures in the Daily `readfile_tests/x86` and `example_configs/x86` tests.


Harshil2107 (approved):
lgtm

PR #3266, opened 2026-06-25
cpu: Fix BPU overriding at squash

dhschall:
In case the BAC stage was blocked due to override
it also did so in case of a squash.
Furthermore, the generation of fetch targets should be stopped once the latency is greater than zero.


powerjg (dismissed):

dhschall:
@powerjg thanks for checking the changes. Sorry, I had a formatting issue and had to amend the commit

powerjg (approved):

PR #3270, opened 2026-06-28
mem-dram: Add HBM3 6.4 Gbps preset (HBM_6400_8H_1x64)

### LINKED ISSUE #3269: RFE: Add HBM3 (JESD238B.01) DRAMInterface preset and HBMCtrl extension
## Context

gem5's current HBM modelling tops out at HBM2:

- `src/mem/DRAMInterface.py` HBM presets (L1046, L1136, L1182):
  - `HBM_1000_4H_1x128` — HBM gen1, 8 channels × 128-bit
  - `HBM_1000_4H_1x64` — HBM gen2 pseudo-channel mode, 16 channels × 64-bit
  - `HBM_2000_4H_1x64` — HBM2, 16 PCs × 64-bit, BL4
- `src/mem/HBMCtrl.py` — explicit comment "HBMCtrl manages two pseudo channels of HBM2"

A `grep` for `HBM3` across the codebase returns no hits. With HBM3 now in mass production for AI/HPC (Nvidia H100/H200, AMD MI300) and HBM3E / HBM4 samples shipping from Samsung and SK hynix, this is becoming a meaningful gap for memory-system research.

## Why it matters

The OSS DRAM-simulator landscape covers HBM3 only via:
- **Ramulator2** (CMU SAFARI) — HBM3 in v2.1 main
- **DRAMSys** (Fraunhofer / RPTU) — HBM3 supported (read-only mirror, PRs not merged)

gem5 is the only **full-system** simulator of the three, so anyone doing OS-aware or GPU-FS work hits this gap.

## Scope (proposal)

In priority order:

1. **`DRAMInterface.py` — HBM3 preset(s)** anchored to JESD238B.01 (April 2025):
   - `HBM_6400_8H_1x64` (6.4 Gbps base rate, 16 channels × 2 PC, BL8) — minimum viable preset
   - Bank organisation: 16 banks per PC (`BA[3:0]`), 4 bank groups (§3.2.1 Table 5)
   - `tCCD_L`, `tRRD_L`, `tFAW`, `tRFC`, `tREFI`, `tRREFD` per JESD238B.01 AC Timings
   - Follow-up presets for HBM3E speed bins (9.6 Gbps)

2. **`HBMCtrl.py` / `hbm_ctrl.cc/hh` — HBM3 extensions**:
   - Per-bank refresh (`REFpb`) handling
   - Per-PC Refresh Management (`RFM`) hint
   - Existing two-pseudo-channel pattern still applies

3. (Follow-up) **HBM4** (JESD270-4A, Nov 2025) — 32 channels, 2048-bit, DRFM — separate effort.

## Background

I have read JESD238B.01 (HBM3) and JESD270-4A (HBM4) end-to-end and have a working pyuvm reference modelling the same command set + RFM/ARFM/DRFM lifecycle (verification side rather than performance side, but the spec coverage overlaps). Recent OSS contribution context: pyuvm core IEEE 1800.2 fix merged (pyuvm#375), cocotb memory example PR in review (cocotb#5658), ramulator2 cross-spec timing audit in active follow-up with maintainers.

## Per CONTRIBUTING.md — may I take this on?

If yes, I'd propose splitting into small reviewable PRs:

1. First PR: just the `HBM_6400_8H_1x64` preset (compile + a smoke test against `DRAMSimpleTest`), with timing values quoted against JESD238B.01 section/table numbers in code comments so reviewers can spot-check accuracy without owning the spec.
2. Once that lands, follow-up PR for `HBMCtrl` extensions.
3. Optional third PR for HBM3E speed bins.

## Questions

- Is there an existing internal plan or Jira issue for HBM3 that I'd be duplicating? (couldn't find one)
- Do you prefer the HBM3 preset to subclass `HBM_2000_4H_1x64` or be a standalone class? (Lean standalone — HBM3 changes bank count and command interface enough that inheritance might mislead.)
- Format preference for JEDEC citations in comments? (e.g. `# tRRD_L = 4: JESD238B.01 §6.3.2 Table 70`)

Happy to start with whichever piece you'd find most useful first.

seam0814:
## Summary

Adds the first HBM3 `DRAMInterface` preset (`HBM_6400_8H_1x64`) to gem5, anchored to **JEDEC JESD238B.01 (April 2025)**. Existing HBM presets stop at HBM2 — see #3269 for the broader context and @powerjg's green-light on this plan.

This is the first of three PRs discussed in #3269:
1. **(this PR)** — `DRAMInterface.py` HBM3 preset
2. `HBMCtrl` extension for per-bank refresh / RFM (follow-up)
3. HBM3E speed bins (optional follow-up)

## What's in this PR

A single new class right after `HBM_2000_4H_1x64`:

- **Standalone** (not subclassed from HBM2) per @powerjg's reply in #3269 — HBM3 changes the burst length (BL=4 → BL=8), data rate (2.0 → 6.4 Gbps/pin), and refresh model enough that inheritance would mislead.
- **Every timing carries an inline JESD238B.01 §/Table citation** (format also confirmed in #3269) so reviewers can spot-check accuracy against the spec without owning a copy.
- Where JESD238B.01 leaves a MIN value to the vendor datasheet, the comment says so and uses a starting value taken from public HBM3 6.4 Gbps part datasheets (SK hynix, Samsung).

## Why 6.4 Gbps as the baseline

JESD238B.01 Table 92 speed bins start at 6.8 Gbps, but the 6.4 Gbps point is the most widely deployed bin in shipping HBM3 silicon (NVIDIA H100 HBM3 stacks). Picking it as the minimum-viable preset means anyone running gem5 against an H100-class profile can use this preset directly.

`tCK = 0.625 ns` (1.6 GHz CK; WDQS at 3.2 GHz DDR).

## Test plan

- [x] Python parses cleanly (`scons --help` loads the config without error)
- [x] Attribute parity with `HBM_2000_4H_1x64` — every Param assigned, no typos (`diff` of attribute sets is empty)
- [ ] Run `tests/test-progs/hello/bin/x86/linux/hello` with `--mem-type=HBM_6400_8H_1x64` once a reviewer can rebuild — I don't have a working full-build environment locally, would appreciate a CI verification
- [ ] Spot-check the JESD citations against the spec

## Notes for reviewers

- `tRFCpb` and `tRREFD` are intentionally **not** included in this preset — they plug into REFpb dispatch in `HBMCtrl`, which is the follow-up PR (also in #3269). Kept this change as small as I could.
- The values not directly fixed by the JEDEC spec (most row timings — JESD238B.01 leaves MIN to the vendor datasheet) are easy to tune in a follow-up if you have a specific target part in mind.

Issue: #3269

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/mem/DRAMInterface.py
+++ b/src/mem/DRAMInterface.py
@@ -1258,6 +1258,131 @@ class HBM_2000_4H_1x64(DRAMInterface):
     two_cycle_activate = True
 
 
+# A single HBM3 x64 pseudo-channel interface, anchored to
+# JEDEC JESD238B.01 (April 2025). HBM3 introduces several
+# changes over HBM2 that this preset captures:
+#
+#   * Native burst length BL=8 (JESD238B.01 Figure 32 onwards);
+#     HBM2 pseudo-channel defaults to BL=4.
+#   * 6.4 Gbps/pin base data rate (used by current production
+#     silicon — e.g. NVIDIA H100 HBM3 stacks). The JESD238B.01
+#     speed bins (Table 92) start at 6.8 Gbps; the 6.4 Gbps
+#     point is the most widely deployed bin in shipping HBM3
+#     parts, hence chosen here as the minimum-viable baseline.
+#     tCK = 0.625 ns (1.6 GHz CK; WDQS at 3.2 GHz DDR).
+#   * Per-bank Refresh (REFpb) is a first-class command; refresh
+#     handling differs from HBM2 but the DRAMInterface side just
+#     needs the matching tRFCpb / tRREFD timings. Controller-side
+#     wiring is left to a follow-up HBMCtrl change.
+#
+# Total stack assumed: 16 channels x 2 pseudo-channels x 8H x
+# 16 Gb/die = 16 GiB per stack -> 512 MiB per pseudo-channel.
+# Tune ``device_size`` for other stack heights / die densities.
+#
+# Where JESD238B.01 leaves a parameter to the vendor datasheet,
+# values are taken from public HBM3 6.4 Gbps part datasheets and
+# marked accordingly so reviewers can refine.
+class HBM_6400_8H_1x64(DRAMInterface):
+    # 64-bit interface for a single pseudo-channel (DQ[31:0] on
+    # PC0, DQ[63:32] on PC1; JESD238B.01 Table 94).
+    device_bus_width = 64
+
+    # HBM3 native burst length is 8 (JESD238B.01 Figure 32, 33,
+    # 40, 41 — every read/write burst diagram uses BL=8).
+    burst_length = 8
+
+    # 8H x 16 Gb/die stack -> 8 Gb per channel -> 512 MiB per
+    # pseudo-channel (16 channels x 2 PCs per stack).
+    device_size = "512MiB"
+
+    # Row buffer size: HBM3 retains the 1 KiB row (JESD238B.01
+    # §3.2 — 32-bit DQ x BL=8 x 1024 columns per row segment).
+    device_rowbuffer_size = "1KiB"
+
+    devices_per_rank = 1
+
+    ranks_per_channel = 1
+
+    # 16 banks per pseudo-channel, organised as 4 bank groups
+    # (JESD238B.01 §3.2 Table 5).
+    banks_per_rank = 16
+    bank_groups_per_rank = 4
+
+    # 1.6 GHz CK for 6.4 Gbps/pin DDR data rate; tCK = 0.625 ns
+    # (just outside the JESD238B.01 Table 92 6.8 Gbps speed bin
+    # — chosen to match shipping HBM3 silicon as noted above).
+    tCK = "0.625ns"
+
+    # Row-access timings (JESD238B.01 Table 93). The spec leaves
+    # MIN values for most row timings to the vendor datasheet;
+    # values below are taken from publicly available HBM3
+    # 6.4 Gbps datasheets (SK hynix, Samsung) and rounded to gem5
+    # quantum granularity. Refine for a specific part.
+    tRP = "14ns"
+    tRCD = "14ns"
+    tRCD_WR = "8ns"
+    tCL = "18ns"
+    tCWL = "8ns"
+    tRAS = "29ns"
+
+    # tCCD_L = max(4 nCK, 2.5 ns) per JESD238B.01 Table 93.
+    # At tCK = 0.625 ns -> 4 nCK = 2.5 ns -> floor.
+    tCCD_L = "2.5ns"
+
+    # BL=8 at DDR -> 8/2 = 4 tCK on the data bus -> 2.5 ns.
+    tBURST = "2.5ns"
+
+    # tRFCab for 8H x 16 Gb (8 Gb / channel) -> 350 ns
+    # (JESD238B.01 Table 93, Refresh Timings row).
+    #
+    # JESD238B.01 also defines tRFCpb (per-bank refresh duration,
+    # 200 ns at 16 Gb/die) and tRREFD (REFpb-to-REFpb spacing,
+    # MAX(3 tCK, 8 ns)). Those plug into REFpb dispatch in
+    # HBMCtrl rather than DRAMInterface, so they are deferred to
+    # the follow-up HBMCtrl HBM3 change tracked in gem5/gem5#3269.
+    tRFC = "350ns"
+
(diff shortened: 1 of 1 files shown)


seam0814:
The failing `quick-tests (gem5/x86_boot_tests)` job is a transient infrastructure issue, not a regression from this PR. From the attempt-2 artifact:

```
ContentTooShortError: <urlopen error retrieval incomplete:
  got only 1413955584 out of 1930823675 bytes>
At:
  src/python/gem5/resources/downloader.py(160): _download
  src/python/gem5/components/boards/x86_board.py(408): _add_disk_to_board
```

The boot test was trying to fetch the ~1.9 GB kernel/disk image and the download truncated at ~1.4 GB. The test exercises `SingleChannelDDR3_1600` / `DDR4_2400` / `LPDDR3_1600` — nothing in the HBM path this PR touches.

Cross-check: PR #3271 stacks on top of this one and its `x86_boot_tests` passed cleanly, so the failure does not reproduce on a fresh download attempt. Could a maintainer kick **Re-run failed jobs** when convenient?

Local checks that do apply to this PR's diff all pass:
- `pre-commit` ✅
- `clang-format-check` ✅
- `unittests-all-fast-opt` ✅
- `quick-tests (gem5/dram_lowp)` ✅
- `quick-tests (gem5/traffic_gen)` ✅

Thanks for the patience!

abmerop (approved):
Seems reasonable to me.  I am only able to really verify the sizing parameters as I only have a JEDEC account but not access to specific vendor timings. The timings are within range though. Any idea what vendor was used here and if the datasheet is public?

seam0814:
Thanks for the review, Matt!

For the values JESD238B.01 leaves MIN to the vendor datasheet, the starting numbers came from the two publicly available HBM3 6.4 Gbps product briefs / newsroom material — **SK hynix HBM3** and **Samsung Icebolt HBM3** (both stack the 6.4 Gbps bin as their entry HBM3 SKU). Where the two vendors overlap I took the more conservative side; where JEDEC pins the value I dropped the vendor entirely and cited the § / Table instead (that's what the inline comments are marking with "JESD238B.01 §…" vs. "vendor-brief starting value").

Neither vendor publishes a full DDR-style datasheet publicly (both are gated behind NDAs for detailed AC characterization), so the preset is deliberately anchored to whatever is quotable without an NDA — spec numbers first, product-brief numbers where the spec is silent. That's why the PR body says "within range" rather than "matches part X exactly".

If it would help future reviewers, I'm happy to add a short header docstring on the preset class that names the specific product-brief pages the vendor-side values were taken from, so anyone auditing later can retrace the exact source without digging into the commit description. Would that be useful, or is the current per-line inline citation format enough?

Same citation pattern will carry over to #3271 (HBM3E 9.6 Gbps) and to the HBMCtrl / RFM follow-up in the trilogy.

abmerop:
> Thanks for the review, Matt!
> 
> For the values JESD238B.01 leaves MIN to the vendor datasheet, the starting numbers came from the two publicly available HBM3 6.4 Gbps product briefs / newsroom material — **SK hynix HBM3** and **Samsung Icebolt HBM3** (both stack the 6.4 Gbps bin as their entry HBM3 SKU). Where the two vendors overlap I took the more conservative side; where JEDEC pins the value I dropped the vendor entirely and cited the § / Table instead (that's what the inline comments are marking with "JESD238B.01 §…" vs. "vendor-brief starting value").
> 
> Neither vendor publishes a full DDR-style datasheet publicly (both are gated behind NDAs for detailed AC characterization), so the preset is deliberately anchored to whatever is quotable without an NDA — spec numbers first, product-brief numbers where the spec is silent. That's why the PR body says "within range" rather than "matches part X exactly".
> 
> If it would help future reviewers, I'm happy to add a short header docstring on the preset class that names the specific product-brief pages the vendor-side values were taken from, so anyone auditing later can retrace the exact source without digging into the commit description. Would that be useful, or is the current per-line inline citation format enough?
> 
> Same citation pattern will carry over to #3271 (HBM3E 9.6 Gbps) and to the HBMCtrl / RFM follow-up in the trilogy.

Thanks. As long as this gets the quoted bandwidth I'm alright with the timings as they are in-line with the ranges. Not sure we have a test for that though so will approve for now

seam0814:
Closing this in favor of #3271 per @abmerop's suggestion — the `HBM_9600_8H_1x64` subclass in #3271 carries the `HBM_6400_8H_1x64` base class in the same diff, so keeping both PRs open is redundant. The 6.4 Gbps preset still lands upstream as the parent class of the HBM3E preset; nothing is lost.

Thanks again for the careful review, Matt!

PR #3272, opened 2026-06-29
mem-cache: fix WriteClean snoop response in write buffer

zhongchengyong:
When a snoop request (e.g., ReadExReq) hits a WriteClean packet in the write buffer, the cache should supply the dirty data carried by the WriteClean to the snooping requestor. Previously, only WritebackDirty packets in the write buffer would trigger a data response; WriteClean packets were silently skipped.

This caused a coherence violation in multi-core scenarios involving cache maintenance operations (e.g., RISC-V cbo.flush). The race condition occurs when:

1. Core A executes a store followed by cbo.flush
2. The cbo.flush generates a WriteClean (carrying the store data) that enters the L1 write buffer
3. Core B issues a ReadExReq for the same line, which snoops Core A L1 and hits the WriteClean in the write buffer
4. Because WriteClean was not handled, Core B receives no data and fetches stale data from lower memory
5. The stale data is later written back, overwriting the correct value

Fix by extending the respond condition to also match WriteClean packets, ensuring their dirty data is supplied to snooping requests.


pranith (approved):

pranith (approved):

zhongchengyong:
@erin-le Can you check why the CI was failed, I click the 'Update branch' by mistaken.

powerjg (approved):

PR #3275, opened 2026-07-01
arch-riscv: refactor vector register gather instructions

saul44203:
This patch refactors RISC-V's vector register gather instructions to produce 1 µop per vd instead of LMUL (or 2*LMUL in the worst case with `vrgatherei16.vv` and SEW=8). With this change there's also no need for destination register pinning, further reducing the µop count. 

In other words, total µop count is now O(LMUL) instead of O(LMUL^2) assuming vl = vlmax.

### DIFF
--- a/src/arch/riscv/insts/vector.cc
+++ b/src/arch/riscv/insts/vector.cc
@@ -216,6 +216,47 @@ std::string VectorSlideMacroInst::generateDisassembly(Addr pc,
     return ss.str();
 }
 
+std::string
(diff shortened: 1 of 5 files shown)


erin-le:
@saul44203: Hello, just wanted to let you know that a conflict has arisen on this PR. Could you take a look when you get a chance? Thanks!

aarmejach (approved):
LGTM and tested internally with Spike co-simulation.

erin-le (approved):

PR #3291, opened 2026-07-07
tests: add stats.txt parser utility

skku970412:
Summary:
- Add a small Python parser for text `stats.txt` files with one or more stats dumps.
- Preserve stat names as emitted, including names with `::`, brackets, hyphens, dots, and digits.
- Add fixture-based pyunit coverage for scalar stats, multi-dump files, edge numeric values, malformed input, duplicate names, and empty/no-dump files.

Motivation:
- This is a non-invasive first step toward #1644.
- A follow-up reset validation test can generate before/reset/after dumps in one invocation and compare selected resettable stats without relying on fixed reference stats files.

Implementation:
- Adds `tests/pyunit/stats/stats_txt.py`.
- Represents each dump as a `StatsDump` with ordered stat values and optional dump message text.
- Parses integer, float, scientific notation, `nan`, and `inf` values using Python `float`.
- Ignores text outside dump delimiters and skips valid `statistics::oneline` rows instead of synthesizing stat names.
- Raises `StatsParseError` for malformed stat values, duplicate stat names within one dump, unterminated dumps, and end-before-begin delimiters.

Testing:
- [x] `python3 -m unittest discover -s tests/pyunit/stats -p 'pyunit*.py' -v` passed, 10 tests.
- [x] `python3 -m unittest discover -s tests/pyunit -p 'pyunit_stats_txt.py' -v` passed, 10 tests.
- [x] Python 3.8 grammar check with `ast.parse(..., feature_version=(3, 8))` passed for the new parser/test files.
- [x] `git diff --check origin/develop...HEAD` passed.
- [x] `pre-commit.ci` passed on the PR after applying its automatic isort/black fix commit.
- [x] `pre-commit run --files $(git diff --name-only origin/develop...HEAD)` passed.
- [x] `scons build/ALL/gem5.opt -j4` passed.
- [x] `./build/ALL/gem5.opt tests/run_pyunit.py --directory tests/pyunit/stats` passed, 10 tests.
- [x] Real-output smoke with `configs/learning_gem5/part1/simple.py` parsed
      one generated `stats.txt` dump with 609 stats, including `simTicks`,
      `simInsts`, and `system.cpu.numCycles`.

Compatibility:
- Does not change simulator behavior or text stats output.
- Does not add external dependencies.
- Keeps parser behavior test-local for now.

Follow-up:
- Add a `m5.stats.reset()` validation test that generates before/after dumps in the same gem5 invocation and uses this parser to compare a narrow set of resettable stats.

### DIFF
--- a/tests/pyunit/stats/pyunit_stats_txt.py
+++ b/tests/pyunit/stats/pyunit_stats_txt.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2026 The Regents of the University of California
+# Copyright (c) 2026 Sungkyunkwan University
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
@@ -55,8 +55,8 @@ class StatsTxtParserTestCase(unittest.TestCase):
             list(dumps[0].stats.keys()),
         )
         self.assertEqual(0.000001, dumps[0]["simSeconds"])
-        self.assertEqual(1000.0, dumps[0]["simTicks"])
-        self.assertEqual(500.0, dumps[0]["system.cpu.numCycles"])
+        self.assertEqual(1000, dumps[0]["simTicks"])
+        self.assertEqual(500, dumps[0]["system.cpu.numCycles"])
 
     def test_multiple_dumps_preserve_order_and_messages(self):
         dumps = parse_stats_file(_FIXTURES / "multiple_dumps.txt")
@@ -64,8 +64,8 @@ class StatsTxtParserTestCase(unittest.TestCase):
         self.assertEqual(2, len(dumps))
         self.assertEqual("before reset", dumps[0].message)
         self.assertEqual("after reset", dumps[1].message)
-        self.assertEqual(100.0, dumps[0]["simInsts"])
-        self.assertEqual(2.0, dumps[1]["simInsts"])
+        self.assertEqual(100, dumps[0]["simInsts"])
+        self.assertEqual(2, dumps[1]["simInsts"])
 
     def test_subnames_and_numeric_edge_values(self):
         dumps = parse_stats_file(_FIXTURES / "edge_values.txt")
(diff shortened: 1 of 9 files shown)


skku970412:
CI note: the current `macos-compilation (opt)` failure appears unrelated to this parser PR. This PR only changes files under `tests/pyunit/stats`, while the failing job stops while compiling `param_CHIGenericController.cc` with `-Wnonportable-include-path` errors for CHI Ruby protocol includes such as:

```text
src/mem/ruby/protocol/chi/generic/CHIGenericController.hh:51:10: error: non-portable path to file '"mem/ruby/protocol/chi/CHIDataMsg.hh"'; specified path differs in case from file name on disk [-Werror,-Wnonportable-include-path]
    51 | #include "mem/ruby/protocol/CHI/CHIDataMsg.hh"
```

I am leaving this PR scoped to the stats parser/test utility rather than mixing in an unrelated CHI include-path fix.

skku970412:
Follow-up on the unrelated macOS opt failure: I opened #3292 as a separate focused PR for the CHI directory case mismatch. Keeping this parser PR unchanged so the stats parser diff stays small.

tests/pyunit/stats/pyunit_stats_txt.py
```diff
@@ -0,0 +1,139 @@
+# Copyright (c) 2026 The Regents of the University of California
```
erin-le: Could you double check the copyright holder here? Typically, it's either the school you attend, or the company you work for.

tests/pyunit/stats/stats_txt.py
```diff
@@ -0,0 +1,156 @@
+# Copyright (c) 2026 The Regents of the University of California
```
erin-le: Same as above

erin-le (commented):
A few comments about copyright:

PR #3292, opened 2026-07-07
mem-ruby: match CHI protocol directory case

skku970412:
Summary:
- Rename `src/mem/ruby/protocol/chi` to `src/mem/ruby/protocol/CHI`.
- Update source includes, SimObject `cxx_header` paths, and the protocol Kconfig include to match the uppercase directory.
- Keep the SLICC protocol name, generated namespace, and simulator behavior unchanged.

Motivation:
- The SLICC protocol is named `CHI`, and generated files are emitted under `build/.../mem/ruby/protocol/CHI`.
- The source directory was lower-case `chi`.
- On case-insensitive filesystems such as the default macOS filesystem, the source mirror path and generated protocol path can collapse to the same physical directory with lower-case canonical spelling.
- Clang then reports `-Wnonportable-include-path` for includes such as `mem/ruby/protocol/CHI/CHIDataMsg.hh`, and the warning is promoted to an error in CI.

Implementation:
- Rename the CHI source directory to match the SLICC protocol name.
- Update references from `mem/ruby/protocol/chi/...` to `mem/ruby/protocol/CHI/...`.
- Update `src/mem/ruby/protocol/Kconfig` from `rsource "chi/Kconfig"` to `rsource "CHI/Kconfig"`.
- Apply minimal style fixes needed because the renamed files are checked by pre-commit.

Testing:
- [x] `pre-commit run --files $(git diff --cached --name-only)` passed before commit.
- [x] `pre-commit run --files $(git diff --name-only origin/develop...HEAD)` passed on 2026-07-07.
- [x] `git diff --check origin/develop...HEAD` passed.
- [x] `scons build/ALL/python/_m5/param_CHIGenericController.o -j4` passed.
- [x] `scons build/ALL/mem/ruby/protocol/CHI/generic/CHIGenericController.o build/ALL/mem/ruby/protocol/CHI/generic/CBusy.o build/ALL/python/_m5/param_CHIGenericController.o -j4` passed.

Compatibility:
- No simulator behavior change intended.
- No SLICC protocol name change.
- No generated namespace change.

Notes:
- This is a clean, focused version of the same root-cause fix attempted in closed PR #3286.
- This also matches the unrelated macOS CI failure observed on PR #3291.

### DIFF
--- a/src/mem/ruby/protocol/chi/generic/CBusy.cc
+++ b/src/mem/ruby/protocol/chi/generic/CBusy.cc
@@ -37,8 +37,8 @@
 
 #include "mem/ruby/protocol/chi/generic/CBusy.hh"
 
-#include "mem/ruby/protocol/CHI/Cache_Controller.hh"
-#include "mem/ruby/protocol/CHI/Memory_Controller.hh"
+#include "Cache_Controller.hh"
+#include "Memory_Controller.hh"
 
 namespace gem5
 {
--- a/src/mem/ruby/protocol/chi/generic/CHIGenericController.cc
+++ b/src/mem/ruby/protocol/chi/generic/CHIGenericController.cc
@@ -45,7 +45,7 @@
 
 #include "debug/RubyCHIGeneric.hh"
 #include "mem/ruby/network/Network.hh"
-#include "mem/ruby/protocol/CHI/CHIProtocolInfo.hh"
+#include "CHIProtocolInfo.hh"
 #include "mem/ruby/protocol/MemoryMsg.hh"
 #include "mem/ruby/system/RubySystem.hh"
 #include "mem/ruby/system/Sequencer.hh"
--- a/src/mem/ruby/protocol/chi/generic/CHIGenericController.hh
+++ b/src/mem/ruby/protocol/chi/generic/CHIGenericController.hh
@@ -48,9 +48,9 @@
 #include "params/CHIGenericController.hh"
 
 // Generated from SLICC
-#include "mem/ruby/protocol/CHI/CHIDataMsg.hh"
-#include "mem/ruby/protocol/CHI/CHIRequestMsg.hh"
-#include "mem/ruby/protocol/CHI/CHIResponseMsg.hh"
+#include "CHIDataMsg.hh"
+#include "CHIRequestMsg.hh"
+#include "CHIResponseMsg.hh"
 
 namespace gem5
 {
--- a/src/mem/ruby/protocol/chi/generic/SConscript
+++ b/src/mem/ruby/protocol/chi/generic/SConscript
@@ -40,6 +40,10 @@ Import('*')
 if not env['CONF']['RUBY_PROTOCOL_CHI']:
     Return()
 
+# SLICC writes generated CHI headers under the protocol name, while the
+# hand-written CHI sources live in a lower-case source directory.
+env.Append(CPPPATH=Dir(env['BUILDDIR']).Dir('mem/ruby/protocol/CHI'))
+
 SimObject('CHIGeneric.py', sim_objects=['CHIGenericController'])
 SimObject(
     'CBusy.py',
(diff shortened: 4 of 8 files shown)


skku970412:
CI note: GitHub Actions `CI Tests` for this PR is currently `action_required` and no jobs have started on the run, so I do not have a failing CI log to fix from this branch.

For the branch-side verification, I ran the targeted local checks listed in the PR body: `git diff --check`, pre-commit on the changed files, and targeted SCons builds for the CHI controller/CBusy/Python param objects. I can re-check or adjust the patch once the workflow is approved or if maintainers prefer a different way to handle the directory/include casing.

powerjg:
Thank you for the contribution. However, I don't see any benefit to this change, and it is going to cause pain to anyone who has private changes. Thus, I would be against merging this PR.

@skku970412 if you can enumerate the benefits and explain how they are much larger than the potential downsides, I am open to reconsidering.

giactra:
> Thank you for the contribution. However, I don't see any benefit to this change, and it is going to cause pain to anyone who has private changes. Thus, I would be against merging this PR.
> 
> @skku970412 if you can enumerate the benefits and explain how they are much larger than the potential downsides, I am open to reconsidering.

I agree this is invasive and a PITA for users, on the other hand I believe @skku970412 addressed their motivation here:

> On case-insensitive filesystems such as the default macOS filesystem, the source mirror path and generated protocol path can collapse to the same physical directory with lower-case canonical spelling.
Clang then reports -Wnonportable-include-path for includes such as mem/ruby/protocol/CHI/CHIDataMsg.hh, and the warning is promoted to an error in CI.

To me this is a valid issue, even if this is going to be painful for us in particular. I thought at the changes with the following questions in mind:

1) In this new scheme generated and non-generated files will live under the same CHI
directory. Is this the final structure we want to achieve? (This is different from what we are doing now)

2) Is there another way to keep disruption at a minimum?

IMHO @skku970412 original implementation is a reasonable long-term solution. It effectively unifies CHI code within one directory, with auto-generated in the top-level and handwritten code in its subdirectories.

mem/ruby/protocol/
mem/ruby/protocol/CHI
mem/ruby/protocol/CHI/tlm/
mem/ruby/protocol/CHI/generic/

I know we are used to it but it is better than having the confusing:

mem/ruby/protocol/
mem/ruby/protocol/CHI
mem/ruby/protocol/chi

Note: having CHI/generated/ would be even better but I am not keen on making slicc harder

About not making slicc harder, another solution could be to leave handwritten code where it is, and simply move autogenerated files to a different folder so that there is no overlap, though this STILL requires us to change some imports, AND it forces us to specialize the CHI protocol so that when we compile it we cannot directly use the protocol name as a subdirectory...

@powerjg what do you think?

powerjg:
Short term, @BobbyRBruce can we revert any changes to the CI infrastructure so this is using an older clang so we can get the CI tests to pass?

I now understand the problem here. I am going to need a few days to investigate the proposed solutions to see what is going to work for us at Google. We have a lot of internal changes to how the compilation works, so moving SLICC code around isn't straightforward.

BobbyRBruce:
> Short term, @BobbyRBruce can we revert any changes to the CI infrastructure so this is using an older clang so we can get the CI tests to pass?
> 
> I now understand the problem here. I am going to need a few days to investigate the proposed solutions to see what is going to work for us at Google. We have a lot of internal changes to how the compilation works, so moving SLICC code around isn't straightforward.

No. This one is due to GitHub's `macos-26` image. On June 30th they updated it so it uses clang 21.0.0 (previously 20.1.8). Clang 21.0.0 has this new/more sensitive warning.

FYI: The reason "fast" is passing is simply because it's building on top of a cache. The warning doesn't trigger because the problematic code is cached and not property parsed by the newer clang. The simplest fix would be a quick fix to the SCons to treat this warning as a warning, though addressing the warning, as this PR is doing, also works.

BobbyRBruce:
I prefered https://github.com/gem5/gem5/commit/ff8edc4333d01e3b18bf15cabb0a41767fb89485 over this current implementation. Any chance we could go back to that? @giactra : I believe you preferred this?

PR #3302, opened 2026-07-09
scons: no error on 'nonportable-include-path'

BobbyRBruce:
This is a quick 'fix' to get the tests working until a better solution is developed.

Current PR under consideration: #3292 (This change should be reverted after this is merged, assuming it fixes warning).


powerjg (approved):

powerjg:
Let's force push while the GPU tests are failing.

PR #3308, opened 2026-07-13
arch-riscv: fix vlseg/vsseg reg-group check for fractional LMUL

pxk27:
The unit-stride segment load/store templates checked the number of vector registers occupied by the instruction with
`(1 << vlmul) * numFields`, where `vlmul` is the sign-extended vtype LMUL field returned by `vtype_vlmul()`.

For fractional LMUL settings (LMUL = 1/2, 1/4, 1/8) `vtype_vlmul()` returns a negative value (-1, -2, -3). Using it as a left-shift amount in `1 << vlmul` is undefined behaviour and in practice produces a huge value, so the `> 8` guard wrongly rejects perfectly legal segment accesses as illegal instructions.

Use the existing `vtype_regs_per_group()` helper instead, which clamps the effective register multiplier to a minimum of 1 for fractional LMUL, matching the RVV specification requirement that a segment instruction occupies at most 8 vector registers. This fixes spurious IllegalInstFault (previously panic) on code compiled with -march=rv64gcv, e.g. the RVV kernels emitted by llama.cpp/ggml.

### DIFF
--- a/src/arch/riscv/isa/templates/vector_mem.isa
+++ b/src/arch/riscv/isa/templates/vector_mem.isa
@@ -1527,9 +1527,9 @@ Fault
     if (update_fault != NoFault) { return update_fault; }
 
 
-    const int64_t vlmul = vtype_vlmul(machInst.vtype8);
-    if (((1 << vlmul) * this->numFields) > 8)
-        return std::make_shared<IllegalInstFault>("LMUL value is illegal for vlseg inst", machInst);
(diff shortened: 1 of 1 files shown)


erin-le:
@saul44203: Hello, could you take a look at this PR when you get a chance? Thank you!

aarmejach:
Thanks for the fix, LGTM.

aarmejach (approved):

powerjg (approved):

PR #3309, opened 2026-07-13
build: Add LoongArch host build support

cheungxi:
## Overview

  Enable gem5 to build and run natively on LoongArch64 hosts.

  This change only adds LoongArch host support. It does not add LoongArch as a simulated target ISA.

  ## Changes

  - Add `EM_LOONGARCH` to the bundled libelf definitions.
  - Detect LoongArch when generating the native libelf format header.
  - Ensure the generated native format header tracks the source-tree detection script when duplicate sources are
  disabled.
  - Recognize LoongArch as little endian in SystemC and HSA.
  - Use the appropriate SystemC 64-bit aliases on LoongArch64 to avoid LP64 overload conflicts.
  - Isolate the SystemC linker flag check from external libraries, avoiding SCons 4.3 failures caused by reused
  library build state.

  ## Testing

  Tested on:

  - Loongnix Server 23.2
  - LoongArch64
  - GCC/G++ 12.3.0
  - Python 3.11.6
  - SCons 4.3.0

  Built successfully with:

  ```bash
  scons --ignore-style --config=force build/ALL/gem5.opt -j32
 ```
 
 Verified that build/ALL/gem5.opt is a native LoongArch64 ELF executable.

  Ran the RISC-V SE hello-world workload:

```
  build/ALL/gem5.opt \
    --outdir=/tmp/gem5-loongarch-smoke \
    configs/deprecated/example/se.py \
    --cpu-type=RiscvAtomicSimpleCPU \
    --cmd=tests/test-progs/hello/bin/riscv/linux/hello
```

  Result:

```
  Hello world!
  Exiting @ tick 3306500 because exiting with last active thread context
```
  The simulator exited successfully with status 0 .

### DIFF
--- a/src/systemc/core/SConscript
+++ b/src/systemc/core/SConscript
@@ -74,6 +74,11 @@ if env['CONF']['USE_SYSTEMC']:
         Source('sc_time_python.cc', append=append)
 
     # Disable the false positive warning for the event members of the scheduler.
-    with gem5_scons.Configure(env) as conf:
-        flag = '-Wno-free-nonheap-object'
-        conf.CheckLinkFlag(flag)
+    flag = '-Wno-free-nonheap-object'
+    # Check whether the linker accepts the flag in an isolated environment so
+    # SCons does not build unrelated library targets during configuration.
+    flag_env = env.Clone()
+    flag_env.Replace(LIBS=[], LIBPATH=[])
+    with gem5_scons.Configure(flag_env) as conf:
+        if conf.CheckLinkFlag(flag, autoadd=False, set_for_shared=False):
+            env.Append(LINKFLAGS=[flag], SHLINKFLAGS=[flag])
--- a/ext/libelf/elfdefinitions.h
+++ b/ext/libelf/elfdefinitions.h
@@ -558,265 +558,227 @@ enum {
 /*
  * ELF Machine types: (EM_*).
  */
-#define	_ELF_DEFINE_ELF_MACHINES()					\
-_ELF_DEFINE_EM(EM_NONE,             0, "No machine")			\
-_ELF_DEFINE_EM(EM_M32,              1, "AT&T WE 32100")			\
-_ELF_DEFINE_EM(EM_SPARC,            2, "SPARC")				\
-_ELF_DEFINE_EM(EM_386,              3, "Intel 80386")			\
-_ELF_DEFINE_EM(EM_68K,              4, "Motorola 68000")		\
-_ELF_DEFINE_EM(EM_88K,              5, "Motorola 88000")		\
-_ELF_DEFINE_EM(EM_IAMCU,            6, "Intel MCU")			\
-_ELF_DEFINE_EM(EM_860,              7, "Intel 80860")			\
-_ELF_DEFINE_EM(EM_MIPS,             8, "MIPS I Architecture")		\
-_ELF_DEFINE_EM(EM_S370,             9, "IBM System/370 Processor")	\
-_ELF_DEFINE_EM(EM_MIPS_RS3_LE,      10, "MIPS RS3000 Little-endian")	\
-_ELF_DEFINE_EM(EM_SPARC64,          11, "SPARC64")				\
-_ELF_DEFINE_EM(EM_PARISC,           15, "Hewlett-Packard PA-RISC")	\
-_ELF_DEFINE_EM(EM_VPP500,           17, "Fujitsu VPP500")		\
-_ELF_DEFINE_EM(EM_SPARC32PLUS,      18,					\
-	"Enhanced instruction set SPARC")				\
-_ELF_DEFINE_EM(EM_960,              19, "Intel 80960")			\
-_ELF_DEFINE_EM(EM_PPC,              20, "PowerPC")			\
-_ELF_DEFINE_EM(EM_PPC64,            21, "64-bit PowerPC")		\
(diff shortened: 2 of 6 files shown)


src/systemc/core/SConscript
```diff
@@ -74,6 +74,9 @@ if env['CONF']['USE_SYSTEMC']:
...
     # Disable the false positive warning for the event members of the scheduler.
-    with gem5_scons.Configure(env) as conf:
-        flag = '-Wno-free-nonheap-object'
-        conf.CheckLinkFlag(flag)
+    flag = '-Wno-free-nonheap-object'
+    flag_env = env.Clone()
+    flag_env.Replace(LIBS=[], LIBPATH=[])
+    with gem5_scons.Configure(flag_env) as conf:
+        if conf.CheckLinkFlag(flag, autoadd=False, set_for_shared=False):
+            env.Append(LINKFLAGS=[flag], SHLINKFLAGS=[flag])
```
abmerop, line 84: Could you add a comment what this does? I think it is checking the flag exists, but I am not sure

ext/libelf/SConscript
```diff
@@ -146,7 +146,8 @@ m4env.Append(CPPPATH=[Dir('.'), Dir('.').srcnode()])
 m4env.Library('elf', [m4env.SharedObject(f) for f in elf_files])
 
 # Generate the native-elf-format header file based on the build system
-m4env.Command(File('native-elf-format.h'), File('native-elf-format'),
+m4env.Command(File('native-elf-format.h'),
+              File('native-elf-format').srcnode(),
```
abmerop: What does this do?

abmerop (commented):
Looks mostly good, I do want to understand some comments though. Thanks for making a separate style commit as well, it makes it easier to review.

src/systemc/core/SConscript
cheungxi: Yes, it checks whether the compiler/linker driver accepts the flag during a link. This isolation is necessary because, with SCons 4.3, the original check inherits gem5's build-time libraries and attempts to build unrelated library targets during the Configure test. That fails internally and incorrectly reports the flag as unsupported. Clearing LIBS and LIBPATH makes the test check only the flag.

ext/libelf/SConscript
cheungxi: After re-checking this, SCons already resolves the file to its source-tree node when sources are not duplicated, and changes to the script correctly trigger regeneration of the header. Therefore, the explicit srcnode() call is redundant, so I’ll remove this change.

abmerop (approved):

powerjg (commented):
For the changes in **ext**/libelf, are these pulling changes in from upstream? We should be very, very careful about introducing local changes as they will be overwritten next time we pull in from the upstream repo.

If they are changes from upstream, then we should pull the whole directory in from upstream.

cheungxi:
> For the changes in **ext**/libelf, are these pulling changes in from upstream? We should be very, very careful about introducing local changes as they will be overwritten next time we pull in from the upstream repo.
> 
> If they are changes from upstream, then we should pull the whole directory in from upstream.

No, the `ext/libelf` changes are not pulled from upstream.

gem5 currently vendors elftoolchain 0.7.1. In that version, `native-elf-format` does not recognize the `LoongArch` machine string, and `elfdefinitions.h` does not define `EM_LOONGARCH`. Without both changes, the bundled libelf cannot be built correctly on a LoongArch Linux host.

The intended semantic changes are only:

* add `EM_LOONGARCH`;
* recognize `LoongArch` in `native-elf-format`.

The large diff in `elfdefinitions.h` is formatting-only churn. My original minimal change failed the standalone `clang-format-check`, so I ran the formatter to satisfy CI. Since the new entry is inside the large `_ELF_DEFINE_ELF_MACHINES()` continuation macro, clang-format reformatted the entire macro.

There also appears to be a difference between the two style-checking paths: gem5's regular style checker excludes files under `ext/`, while the standalone clang-format CI invokes `run-git-clang-format.py` directly and still checks changed vendored files. This is why the minimal libelf change triggered formatting of the full macro.

I agree that we should be careful about carrying downstream changes in vendored code. Upstream elftoolchain later removed `native-elf-format` and changed its configuration mechanism, so pulling the current upstream directory would be a substantially larger vendor update rather than a direct import of these two changes.

Would you prefer that I:

1. restore the original libelf formatting and retain only the minimal LoongArch compatibility changes, then address the clang-format behavior separately if needed; or
2. prepare a complete libelf refresh from a pinned upstream revision as a separate prerequisite PR?

I do not want to change the CI policy or introduce a large vendor refresh in this PR without first agreeing on the preferred approach.

powerjg (approved):
It looks like we are way behind the upstream for libelf. I will approve this as it, but there is a chance that if/when we ever update libelf that this change could vanish.

erin-le:
@cheungxi: Could you change the title of your second commit from `style: Format LoongArch host changes` to `build: Format LoongArch host changes`? For reference, [MAINTAINERS.yaml](https://github.com/gem5/gem5/blob/stable/MAINTAINERS.yaml) contains a list of the tags this repo uses. Thank you!

cheungxi:
> @cheungxi: Could you change the title of your second commit from `style: Format LoongArch host changes` to `build: Format LoongArch host changes`? For reference, [MAINTAINERS.yaml](https://github.com/gem5/gem5/blob/stable/MAINTAINERS.yaml) contains a list of the tags this repo uses. Thank you!

Done. I changed the second commit title to build: Format LoongArch host changes. Thank you!

PR #3313, opened 2026-07-14
arch-riscv: refactor vector slide instructions

saul44203:
This patch refactors RISC-V's vector slide instructions implementation to reduce µop splitting from O(LMUL^2) (i.e. LMUL(LMUL-1)/2+1) to O(LMUL) (i.e. one per vd) assuming VL = VLMAX. This also gets rid of additional register pin and copy µops.


aarmejach (approved):
LGTM and tested internally with Spike co-simulation.

BobbyRBruce (approved):

PR #3314, opened 2026-07-16
misc: Don't run very-long x86_boot_tests with gcov

erin-le:
When the very-long x86_boot_tests are run with a gem5 binary compiled with flags for gcov, the 8-core O3 CPU test fails with a segmentation fault. This commit modifies the weekly test workflow file so the very-long x86_boot_tests don't use the --gcov=test-only flag.


erin-le:
Ready for review now

powerjg (approved):
Ironically, one reason we want gcov is to see where our tests are *not* adding any coverage value. I wonder if this test is adding coverage value or not...

PR #3319, opened 2026-07-20
misc: Hide modulo-based interleaving constructor

giactra:
A recent PR added modulo-based interleaving for
the AddrRange class in gem5 [1]

With this patch we mark the modulo-based interleaving constructor as private.
An address range with modulo-based interleaving can only be constructed via a factory method:

auto addr_range = AddrRange::withModuloInterleaving(..)

This is done to improve readibility and decrease chances of misconfiguring the interleaving policy (for example if someone mistakenly uses masked interleaving in lieu of modulo-based) and also to improve readability, so that it is obvious what's the interleaving policy being used.

For symmetry we add a withMaskInterleaving method though we don't mark the matching constructor as privare since it would be a compatibility nightmare

[1]: https://github.com/gem5/gem5/pull/2861

Change-Id: Ia579c6f9ee97846e7d754d15635925eb938f949a

### DIFF
--- a/src/base/addr_range.hh
+++ b/src/base/addr_range.hh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2014, 2017-2019, 2021 Arm Limited
+ * Copyright (c) 2012, 2014, 2017-2019, 2021, 2026 Arm Limited
  * Copyright (c) 2026 Google Inc.
  * All rights reserved
  *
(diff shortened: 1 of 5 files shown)


powerjg (dismissed):
Thanks for doing this. Apologies that I didn't get around to it. It looks good to me (without looking closely). Let's see if copilot has any suggestions :)

powerjg (approved):

PR #3330, opened 2026-07-21
misc: Add labeler workflow to label PRs based on files changed

erin-le:
This PR adds a workflow to automatically assign labels to PRs based on the files changed. The path rules in `labeler.yml` were written based on how labels have typically been assigned in the gem5 repository in the past. The discussion for this workflow can be seen in https://github.com/gem5/gem5/pull/3095, and the work for this workflow was started by @lhartung.

The commit this PR adds was originally merged into the `develop` branch in https://github.com/gem5/gem5/pull/3151, but using the `pull_request_target` trigger means that the workflow files need to be on the default branch (stable) for the workflow to run (see [here](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target)).

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,181 @@
+---
+# If this action would apply 5+ labels to a PR, don't apply any labels.
+changed-files-labels-limit: 4
+
+arch:
+    - changed-files:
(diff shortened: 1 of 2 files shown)


powerjg (approved):
This is OK, but I don't like that we now have two different places where we have the same information: `.github/labeler.yml` is basically the same as what is in `CONTRIBUTING.md`. Maybe it's OK since it's so rare that we update `CONTRIBUTING.md`

BobbyRBruce (approved):

PR #3331, opened 2026-07-23
sim: Add getGdbListenerOutput method to Workload

OmeletWithoutEgg:
Expose getListenerOutput on BaseRemoteGDB and getGdbListenerOutput on Workload to query the actual bound runtime socket address of the remote GDB listener.

### DIFF
--- a/src/base/remote_gdb.cc
+++ b/src/base/remote_gdb.cc
@@ -452,6 +452,17 @@ BaseRemoteGDB::hostSocket() const
     return *listener;
 }
 
+std::string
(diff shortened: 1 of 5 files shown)


giactra (commented):
HI, this looks good to me. Can I ask you what is the use-case?

OmeletWithoutEgg:
Similar to #3289 which exports serial terminal ports, we want to expose the remote GDB ones as well.

powerjg (approved):

PR #3332, opened 2026-07-23
dev-amdgpu: Device-backed MQD support

abmerop:
Latest ROCm release 7.14 now utilizes VRAM-backed memory queue descriptors (MQDs) for PM4 and SDMA queues. Add support in gem5 to be able to read and write back these MQDs.


src/dev/amdgpu/pm4_packet_processor.cc
```diff
@@ -467,6 +479,84 @@ PM4PacketProcessor::mapQueues(PM4Queue *q, PM4MapQueues *pkt)
     }
 }
 
+void
+PM4PacketProcessor::mapDeviceQueue(PM4Queue *q, PM4MapQueues *pkt)
+{
+    pkt->mqdAddr = pkt->mqdAddr - gpuDevice->getVM().getMMHUBBase();
+
+    // We are known to be in MMHUB. Subtract MMHUB base to get device address.
+    if (pkt->engineSel == 0 || pkt->engineSel == 1 || pkt->engineSel == 4) {
```
mattsinc: do we want to have a helper function for this -- since the code is repeated twice?

src/dev/amdgpu/pm4_packet_processor.cc
```diff
@@ -631,13 +723,33 @@ PM4PacketProcessor::unmapAllQueues(bool unmap_static)
...
+                for (; !gen.done(); gen.next()) {
+                    gpuDevice->getMemMgr()->writeRequest(
+                        chunk_addr, buf, gen.size(), 0,
+                        gen.last() ? cb : nullptr);
+                    buf += gen.size();
+                    chunk_addr += gen.size();
+                }
+            } else {
+                Addr addr =
+                    getGARTAddr(queues[id]->mqdBase() + 96 * sizeof(uint32_t));
```
mattsinc: perhaps do we want a #define for 96?  Because it is used a lot ...

erin-le:
@abmerop: Hello, could you respond to the comments above when you get a chance? Thank you!

src/dev/amdgpu/pm4_packet_processor.cc
abmerop: done

src/dev/amdgpu/pm4_packet_processor.cc
abmerop: done

abmerop:
@mattsinc Would be good to get this in next release so we can use new disk images. I believe I addressed all comments.

BobbyRBruce (changes requested):
The following values should be added to the existing deserialization:

- `PM4Queue::_deviceBacked` — for each PM4 queue.
- `PM4Queue::_pkt.mqdAddr` — for each PM4 queue.
- `SDMAEngine::SDMAQueue::_deviceBacked` — for each SDMA RLC queue (`rlc0 and `rlc1`).

Without this the flags are set to `false` and the address `0` after a checkpoint is restored. This would cause problems.

abmerop:
> The following values should be added to the existing deserialization:
> 
> * `PM4Queue::_deviceBacked` — for each PM4 queue.
> * `PM4Queue::_pkt.mqdAddr` — for each PM4 queue.
> * `SDMAEngine::SDMAQueue::_deviceBacked` — for each SDMA RLC queue (`rlc0 and `rlc1`).
> 
> Without this the flags are set to `false` and the address `0` after a checkpoint is restored. This would cause problems.

Done. Is this going to require a checkpoint upgrader? I think your new tests might be the only usage of GPU checkpoints non-privately.

abmerop:
Added a commit to make the new fields optionally unserialize. The possible values for older disk images, and therefore any possible existing checkpoint, are false/unused.

Would prefer to squash commits for this PR if the test passes with the optional fields change.

BobbyRBruce:
This seems ok to me. The checkpoint appears to be working with these changes and I agree there's no need to create a checkpoint updater for this. I'm almost certain i'm the only person who's created a checkpoint to this extent.

PR #3335, opened 2026-07-24
stdlib: MultiChannelSimpleMemory for bandwidth scaling

abmerop:
Add MultiChannelSimpleMemory to the gem5 stdlib simple memory module. The new class instantiates N SimpleMemory SimObjects (N must be a power of 2) and interleaves their address ranges using gem5's AddrRange bit-mask mechanism, mirroring the pattern used by ChanneledMemory for DRAM. Bandwidth is specified per channel so total aggregate bandwidth scales linearly with channel count, enabling straightforward bandwidth scaling experiments without modelling a specific memory technology.

### DIFF
--- a/src/python/gem5/components/memory/__init__.py
+++ b/src/python/gem5/components/memory/__init__.py
@@ -32,6 +32,10 @@ from .multi_channel import (
     DualChannelLPDDR3_1600,
     DualChannelLPDDR5_5500,
 )
+from .simple import (
+    MultiChannelSimpleMemory,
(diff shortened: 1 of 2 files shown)


mattsinc (approved):
I don't see anything obviously wrong here, and thus it seems reasonable to merge in.  But probably @powerjg should chime in to make sure this fits the vision for models we want to support.

powerjg (approved):
lgtm. No reason we can't support modulo addresses, though, if you wanted to have non powers of two. No need to add support to this PR.

PR #3339, opened 2026-07-27
ext, tests: Add AMBA CHI_TLM library to gem5

giactra:
This PR adds the public AMBA TLM 2.0 Library available
through the Arm Developer website [1] to the gem5
ext libraries.

From the webiste:

```
The AMBA Transaction-Level Modelling (TLM) 2.0 Library allows you to
model and simulate Approximately-Timed (AT) and Cycle-Accurate (CA) AXI,
ACE and CHI transactions and ports.

This library is intended for use with:

    Custom AT and CA models used for virtual prototype development.
    Third-party models that use supported AMBA interfaces.

Your AMBA TLM Library package includes the library, documentation, and
required licenses.
```

It has to be noted we removed the pdf from the tarball. It
can always be downloaded separately from the following
link [2]. We also removed the examples directory

The library will be statically compiled and linked with gem5, UNLESS a
TLM_PATH= option is provided while building the simulator.
Manually setting TLM_PATH will redirect the build system towards
an alternative implementation and will try to dynamically link
to it.
This will allow users to experiment with different flavours
of the library if needed, and fallback to ext/amba_tlm if no
override variable is specified. Simply put:

scons build/ALL/gem5.opt -> Compile and use ext/amba_tlm/
scons build/ALL/gem5.opt TLM_PATH=/path/to/tlm/lib -> Dynamically link
to custom AMBA TLM library in /path/to/tlm/lib

In the PR we also port the existing testing python config to stdlib and
we hook it to testlib, effectively starting unit-testing CHI in gem5

[1]: https://developer.arm.com/architectures/amba#Downloads
[2]: https://developer.arm.com/documentation/101459/latest


tests/gem5/chi_tlm_tests/configs/ruby_mem_test.py
```diff
@@ -164,70 +272,97 @@ def create_system(options, system):
...
-root = Root(full_system=False, system=system)
-root.system.mem_mode = "timing"
+board = TestBoard(
+    clk_freq=args.sys_clock,
+    generator=processor,
+    memory=memory,
+    cache_hierarchy=cache_hierarchy,
+)
+board.cache_line_size = args.cacheline_size
+board.clk_domain.voltage_domain = voltage_domain
```
powerjg: This line and the next shouldn't be needed.

tests/gem5/chi_tlm_tests/configs/ruby_mem_test.py
```diff
@@ -164,70 +272,97 @@ def create_system(options, system):
...
+board._set_fullsystem(False)
 
-# Not much point in this being higher than the L1 latency
-m5.ticks.setGlobalFrequency("1ns")
+for generator in processor.get_tlm_generators():
+    suite.test_all(generator)
 
-# instantiate configuration
+root = board._pre_instantiate()
```
powerjg, line 358: Why are you not using the `Simulator` object? You don't have to manually manage the root, `instantiate`, or `simulate` if you use `sim = Simulator(board=board); sim.run()` (or something close to that).

I appreciate that this was modified to use the stdlib. It would be great if we could take it just a bit further and finish the migration.

Let me know if there is something that is blocking, though.

powerjg (commented):
Just a couple of small things. If it's a big lift, they can be skipped.

tests/gem5/chi_tlm_tests/configs/ruby_mem_test.py
giactra: Yeah so the issue is that the "TlmGenerator done" exit reason is not handled by the translate_exit_status function which is called by the Simulator class.

https://github.com/gem5/gem5/blob/stable/src/python/gem5/simulate/exit_event.py#L61

I don't think we should be adding more custom exit events over that static if-else chain. I am not sure if they hypercall logic covers it, but we should be able to let the user specify custom exits as if they were custom exceptions in python... I have not fully though of it, but in the future I want to be able to specify custom exits from simple microarchitectural events (something like: exit simulation on a cache refill). It does not scale with the current logic

tests/gem5/chi_tlm_tests/configs/ruby_mem_test.py
giactra: Done

tests/gem5/chi_tlm_tests/configs/ruby_mem_test.py
powerjg: Sigh, https://github.com/gem5/gem5/pull/2983 should implement this. It should be merged soon. 

The idea is that you should be able to subclass `ExitHandler` (https://github.com/gem5/gem5/blob/develop/src/python/gem5/simulate/exit_handler.py#L86) to add your own user-defined exits and handlers. 

For example:

If in the gem5 code you have this:

```cpp
event = new GlobalSimLoopExitEvent(curTick(), 12345, {"my_metadata":"is_important"});
// note: the third argument is a map<string,string>
```

OR, in your binary that is executing you have this:

```cpp
#include <gem5/m5ops>
...
  m5_hypercall(12345);
  // note: we didn't implement a payload when calling from usermode guest
```

Then, in your python runscript you can catch this exit handler with the following code:

```python
class MyHandler(ExitHandler, hypercall_num=12345):
        print(self._payload)

  def _exit_simulation():
        return False
```

That's the *IDEA*, but it requires #2983 and it's not perfect. It's a huge step forward, though.

giactra: Hey @powerjg, thanks for the overview.

I forgot the other reason why we didn't use the higher level Simulator component. Effectively we wanted down the line to re-create the simulate -> verifier -> simulate loop I presented one or two months ago at the developer's meeting.
Rather than waiting for this to be merged and push another PR, I have added the changes here so that you can have a better picture on what we are trying to achieve. If this is fine with you, I'd rather go forward with things as they are (provided that everyone is ok with the rest of the PR of course :) ), and consider later if the scheduling/simulation loop I am implementing is achievable within existing stdlib classes.

Apologies if I repeat myself, but as I mentioned sometime ago, if the plan is to replace everything with stdlib, we should make sure it is able to flexibly cover every use case. The only way to do so is with modularity, so that users can, within the realm of stdlib, cherry pick components and build their own configuration platform.
Unfortunately at times stdlib seems to be providing a useful monolithic black box to the user, at the cost of configurability/modularity. This is probably valid for top-level complex classes.
I don't think we should be forcing users to use objects like Simulator. If the goal of stdlib is to define modular building blocks, a user should be allowed to replace one of those blocks with their own.
They can use the Simulator as an example for an easy solution, but they might want to come-up with something different if the Simulator does not fit the purpose. The alternative is to extend the Simulator to allow it to cover more niche use-cases. Do we want that complexity?

powerjg (approved):

tests/gem5/chi_tlm_tests/configs/ruby_mem_test.py
powerjg: I can't agree more with your points here. I hugely welcome any improvements you may have to the stdlib and its APIs!!!

>  if the plan is to replace everything with stdlib, we should make sure it is able to flexibly cover every use case

I can't agree more!

> Unfortunately at times stdlib seems to be providing a useful monolithic black box to the user, at the cost of configurability/modularity

This means the interfaces should be improved. 

My software engineering approach is to produce something that works for one use case, modify it for a second use case, and then generalize it when we find a third (or more use cases). I think that trying to design something for all possible use cases at the first commit will inevitably lead to poor outcomes. 

So, for the stdlib, if you have a "third" use case where we need to generalize, we should be fixing the stdlib interfaces. We should make the `Simulator` more extensible so that it fits the use case you are going for.

I would rather not *break* the abstractions (e.g., by manually calling `m5.instantiate()`). I would rather fix the abstractions that don't work for your use case.

> The alternative is to extend the Simulator to allow it to cover more niche use-cases. Do we want that complexity?

I think it depends. We *ABSOLUTELY DO NOT* want the `Simulator` to turn back into `se.py`. I would much rather see us add a `AbstractSimulator` (or something similar) that has a `run()` function that has the guarantee that it will manage things like `m5.instantiate()`, etc. and to have a `TestingSimulator` and a `Simulator` that do things differently (but share similar code).

---

All this said, for this PR, if the stdlib's interfaces right now don't fit the use case that you have, you should absolutely "go around them" and then we can refactor these scripts when we improve the stdlib interfaces later. We should create an issue or something so we can track the progress, though!

PR #3341, opened 2026-07-27
sim: Use `std::barrier` instead of custom implementation

sudhanshu2:
Replaces gem5's custom barrier implementation that uses conditional variables with C++20's `std::barrier`. The new implementation models the exact behavior of the custom barrier implementation and there are no changes expected to the correctness of simulation.

`std::barrier` scales substantially better than gem5's barrier implementation in parallel workload execution resulting in CPU time spent in barriers halving for a 16-core parallel simulation for SPEC workloads (shown in figure below).

barrier_percent


src/sim/global_event.cc
```diff
@@ -36,8 +36,13 @@ namespace gem5
 
 std::mutex BaseGlobalEvent::globalQMutex;
 
+// Defines the function that is called **once** per barrier sync event
+void
+ProcessBarrier::operator()() const
+{ event->process(); }
```
giactra: This is likely failing clang-format. Did you use clang-format by chance? I know some versions produce this kind of output by I believe CI is using something like 18.1. May I ask you to update clang-format to that version?

giactra (commented):
> Replaces gem5's custom barrier implementation that uses conditional variables with C++20's `std::barrier`. The new implementation models the exact behavior of the custom barrier implementation and there are no changes expected to the correctness of simulation.
> 

Thanks, this makes sense!

> `std::barrier` scales substantially better than gem5's barrier implementation in parallel workload execution resulting in CPU time spent in barriers halving for a 16-core parallel simulation for SPEC workloads (shown in figure below).

I am actually curious: are you executing a multi-threaded version of gem5? Would you mind sharing more details? Is there a plan to upstream this?

sudhanshu2:
Yes, I tested this on a multi-threaded gem5 execution! However, it is not a "generic" implementation and I will add an example of a multi-threaded gem5 board once I've finished testing it!

src/sim/global_event.cc
sudhanshu2: Yeah, I finally got that working and fixed this issue!

giactra (dismissed):

giactra:
> Yes, I tested this on a multi-threaded gem5 execution! However, it is not a "generic" implementation and I will add an example of a multi-threaded gem5 board once I've finished testing it!

Great, a non-generic config is better than no config. Would love to see this config tested on our CI framework. Some of your changes could be easily undone by mistake as most reviewers (I include myself in this pool of people) are not so familiar with a MT gem5.

powerjg:
> Would love to see this config tested on our CI framework.

Good point! Once we get an example working, we'll do this.

giactra:
The GPU tests are failing because of the old compiler/image being used:

```
                 from src/python/pybind11/event.cc:48:
src/sim/global_event.hh:33:10: fatal error: barrier: No such file or directory
   33 | #include <barrier>
```

@abmerop @BobbyRBruce

powerjg:
IIRC, we decided in last month's developer meeting to enable C++20 and drop support for ubuntu 20.04. Am I mistaken?

giactra:
> IIRC, we decided in last month's developer meeting to enable C++20 and drop support for ubuntu 20.04. Am I mistaken?

Yes but the gpu image update caused some segfaults in gpu-tests and IIRC the decision was for @BobbyRBruce to roll back to the older version, which might not support C++20.

BobbyRBruce:
Thanks for putting this together! I have merged the latest `develop` into the PR branch without rewriting the existing commit, then added a small follow-up to preserve the existing global-event behavior and cover it with a focused test. The local build and test checks pass.

This PR is still blocked on #3401 being merged. The old `gcn-gpu` code and CI environment rely on an older Ubuntu/compiler setup that is not compatible with C++20. #3401 removes that obsolete `gcn-gpu` dependency. Once #3401 lands, we can bring this branch up to date with `develop`, run the normal CI checks, and proceed with final review and merge.

src/sim/global_event.hh
```diff
@@ -57,6 +57,15 @@ namespace gem5
...
  */
 
+/*
+ * Global barrier used to synchronize threads
+ */
+struct ProcessBarrier
+{
+    BaseGlobalEvent *event;
+    void operator()() const;
+};
```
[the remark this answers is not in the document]
BobbyRBruce: Added.

BobbyRBruce (approved):

PR #3342, opened 2026-07-27
sim: Remove event queue asserts that cross thread boundaries

sudhanshu2:
`schedule` can be called from different event queues executing on different threads resulting in `getCurTick()` accessing data on a different thread that is not protected by locks/atomics. This PR moves the assert for checking if an event is not scheduled in the past to `insert()` prevent cross-thread data accesses. 

`reschedule` is also refactored to call `getCurTick()` after it has verified that the method is being called from the "home" event queue.

### DIFF
--- a/src/sim/eventq.cc
+++ b/src/sim/eventq.cc
@@ -133,6 +133,8 @@ Event::releaseImpl()
 void
 EventQueue::insert(Event *event)
 {
+    gem5_assert(event->when() >= getCurTick(),
+                "Event must be inserted at or after the current tick.");
     // Deal with the head case
     if (!head || *event <= *head) {
         head = Event::insertBefore(event, head);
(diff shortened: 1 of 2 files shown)


giactra:
I am not against this change, and I understand the motivation.
However it feels like we are slightly changing what is currently allowed, because we now allow in a parallel simulation to schedule events before the queue current tick (a parallel simulation won't use insert). What am I missing?

sudhanshu2:
I think #3351 addresses your concern about scheduling events in the past in `asyncInsert`. Or do you have something else in mind?

powerjg:
> However it feels like we are slightly changing what is currently allowed, because we now allow in a parallel simulation to schedule events before the queue current tick (a parallel simulation won't use insert). What am I missing?

We aren't changing the behavior. `insert` is only called from the owning thread and at the *sim quantum* when the owning thread drains the `asyncQueue`.  So, you can still add events into the `asyncQueue` from another thread even if the current thread is "ahead" in time.

At least, that's the idea.

giactra (approved):
Apologies, I didn't realize insert was called anyway at the end of the quantum during sync-up. This defers the assertion but we are still checking functional correctness

PR #3346, opened 2026-07-28
mem-ruby: enqueue message with timestamp

jonathanhaley:
Added an 'enqueue' which takes a timestamp as an additional optional parameter. This timestamp is used to order messages scheduled in the same cycle in the message buffer. Also add a test script to verify the functionality of this change.

### DIFF
--- a/src/mem/ruby/slicc_interface/Message.hh
+++ b/src/mem/ruby/slicc_interface/Message.hh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 ARM Limited
+ * Copyright (c) 2021, 2026 Arm Limited
  * All rights reserved.
  *
  * The license below extends only to copyright in the software and shall
@@ -66,7 +66,9 @@ class Message
         : m_block_size(block_size),
           m_time(curTime),
           m_LastEnqueueTime(curTime),
-          m_DelayedTicks(0), m_msg_counter(0)
+          m_DelayedTicks(0),
+          m_msg_counter(0),
+          m_transaction_time(MaxTick)
     { }
 
     Message(const Message &other) = default;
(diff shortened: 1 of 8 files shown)


src/mem/ruby/slicc_interface/Message.hh
```diff
@@ -66,7 +66,9 @@ class Message
         : m_block_size(block_size),
           m_time(curTime),
           m_LastEnqueueTime(curTime),
-          m_DelayedTicks(0), m_msg_counter(0)
+          m_DelayedTicks(0),
+          m_msg_counter(0),
+          m_transaction_time(MaxTick)
```
powerjg, line 71: I feel like this name is a bit confusing. As far as I can tell, this is a priority or an ordering, not a "time."

Can you let me know if I'm misinterpreting or update the variable name?

jonathanhaley: Yes it is being used to prioritize ordering however it is specifically designed to be used  to order events according to the timestamp of the transaction that initiated them. Admittedly though this was not clear without the third commit in this patch set (which has now been added). Personally as the use case is relating to the time at which the transaction began I believe the naming is fine but if you think it would be better generalized Id be ok with that.

src/mem/slicc/parser.py
```diff
@@ -696,15 +696,27 @@ def p_statement__assign(self, p):
...
     def p_statement__enqueue_latency_bypass_strict_fifo(self, p):
         "statement : ENQUEUE '(' var ',' type ',' expr ',' expr ')' statements"
-        p[0] = ast.EnqueueStatementAST(self, p[3], p[5], p[7], p[9], p[11])
+        p[0] = ast.EnqueueStatementAST(
+            self, p[3], p[5], p[7], p[9], None, p[11]
+        )
+
+    def p_statement__enqueue_latency_bypass_strict_fifo_timestamp(self, p):
+        "statement : ENQUEUE '(' var ',' type ',' expr ',' expr ',' expr ')' statements"
+        p[0] = ast.EnqueueStatementAST(
```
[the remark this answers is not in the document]
jonathanhaley: I kind of believe this is fine, semantically it is optional whether to also provide this option.

powerjg (approved):

PR #3348, opened 2026-07-28
arch-x86: Implement SGDT and SIDT macroops

seaver-olson:
## Overview

  - Replace warning-only SGDT and SIDT handling with real x86 macroops.
  - Store a 2-byte limit followed by an 8-byte base in 64-bit mode.
  - Store a 2-byte limit followed by a 4-byte base in legacy and
    compatibility modes.
  - For legacy operand-size-16 forms, retain base bits 23:0 and store
    bits 31:24 as zero.
  - Preserve memory-only decoding, existing group-7 aliases, and
    RIP-relative addressing.

  ## Testing

  - `pre-commit` checks passed.
  - Incremental X86 ISA generation and `build/X86/gem5.opt` build passed.
  - Self-checking x86-64 and i386 probes passed on AtomicSimpleCPU and
    O3CPU.
  - Probes checked exact operand bounds, RIP-relative addressing,
    32-bit and 16-bit layouts, and the 64-bit behavior with `0x66`.
  - A non-aliased register ModRM encoding reached `InvalidOpcode`.

  ## Known limitations

  - CR4.UMIP is not modeled.
  - VMX descriptor-table exiting is not modeled.
  - The repository currently has no focused x86 ISA execution-test
    harness, so the execution probes were not added to the test suite. 

AI-assisted development: OpenAI Codex was used to help inspect relevant gem5 conventions, implement the initial patch, and run validation. I reviewed the final diff, verified the architectural behavior against the Intel specification, and independently take responsibility for the correctness and maintenance of the change.


powerjg (approved):
LGTM. Let's see if copilot catches anything.

powerjg:
This is a good addition, and I assume that you will be adding more unimplemented x86 instructions (or fixing broken ones). It would be great if you could work with AI to build a better tester for x86 instructions as you're working on this.

seaver-olson:
@powerjg Yes, I will get on that right away, I was actually planning on building one anyways so the timing is perfect.

PR #3349, opened 2026-07-28
sim: Replace Event::instanceCounter and remove EVENTQ_DEBUG

sudhanshu2:
The Event::instanceCounter is used to generate unique IDs across all queues for Event instances when generating traces for debugging gem5. Multiple threads can write to the same counter at the same time, and so making it thread-safe is necessary. We replace it with an EventQueue local counter and the trace outputs the name of the queue where an Event is scheduled. 

Additionally, this PR removes the unused EVENTQ_DEBUG macro and associated member variables (whenCreated, whenScheduled) as they are not defined or used.

### DIFF
--- a/src/sim/eventq.hh
+++ b/src/sim/eventq.hh
@@ -276,9 +276,6 @@ class Event : public EventBase, public Serializable
     Flags flags;
 
 #ifndef NDEBUG
-    /// Global counter to generate unique IDs for Event instances
-    static Counter instanceCounter;
-
     /// This event's unique ID.  We can also use pointer values for
     /// this but they're not consistent across runs making debugging
     /// more difficult.  Thus we use a global counter value when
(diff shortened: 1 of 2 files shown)


src/sim/eventq.hh
```diff
@@ -288,11 +285,9 @@ class Event : public EventBase, public Serializable
     /// queue to which this event belongs (though it may or may not be
     /// scheduled on this queue yet)
     EventQueue *queue;
-#endif
 
-#ifdef EVENTQ_DEBUG
-    Tick whenCreated;   //!< time created
-    Tick whenScheduled; //!< time scheduled
+    // The queue from which this event was scheduled.
+    EventQueue *origin_queue;
```
giactra: This should be originQueue

src/sim/eventq.hh
```diff
@@ -961,6 +959,12 @@ class EventQueue
      */
     void checkpointReschedule(Event *event);
 
+#ifndef NDEBUG
+    Counter
+    incrementAndGetEventCounter()
+    { return ++eventCounter; }
+#endif
```
giactra, line 968: As I mentioned in another PR, you need to install clang-format 18.1 to get this formatting accepted by our CI

giactra (commented):
Hi @sudhanshu2 , thanks for this. Makes sense. I have some minor comments

src/sim/eventq.hh
sudhanshu2: Fixed!

src/sim/eventq.hh
sudhanshu2: Done!

giactra (approved):

PR #3351, opened 2026-07-29
sim, mem-ruby: Add assertions to prevent scheduling events in the past and fix data race in MessageBuffer

sudhanshu2:
This PR introduces tracking for the next simulation quantum in EventQueue, which is updated by GlobalSyncEvent. It adds assertions to ensure that asynchronous events are scheduled after the next simulation quantum, and that Ruby Consumer wakeups are not scheduled in the past. This PR also adds ScopedMigration to enqueue in MessageBuffer to prevent data races in the buffer.

### DIFF
--- a/src/mem/ruby/common/Consumer.cc
+++ b/src/mem/ruby/common/Consumer.cc
@@ -40,6 +40,10 @@
 
 #include "mem/ruby/common/Consumer.hh"
 
+#include <cassert>
+
+#include "base/logging.hh"
+
 namespace gem5
 {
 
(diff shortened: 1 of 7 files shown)


src/sim/eventq.hh
```diff
@@ -748,6 +749,10 @@ class EventQueue
     void name(const std::string &st) { objName = st; }
     /** @}*/ //end of api_eventq group
 
+    void
+    setNextSimQuantum(Tick when)
+    { _nextSimQuantum = when; }
```
giactra: Usual git-clang-format problem.

src/mem/ruby/common/Consumer.cc
```diff
@@ -55,15 +59,26 @@ Consumer::Consumer(ClockedObject *_em, Event::Priority ev_prio)
 void
 Consumer::scheduleEvent(Cycles timeDelta)
 {
-    m_wakeup_ticks.insert(em->clockEdge(timeDelta));
+    // m_wakeup_ticks is a shared variable whose "home" event queue is the
+    // one for the object associated with thie consumer.
+    assert(em->eventQueue() == curEventQueue());
+    Tick when = em->clockEdge(timeDelta);
```
giactra: Unrelated: this is our chance to rename timeDelta to time_delta for consistency with the coding style

erin-le:
@sudhanshu2: Hello, could you respond to/address the above comments when you get a chance? Thank you!

src/sim/eventq.hh
sudhanshu2: Fixed!

src/mem/ruby/common/Consumer.cc
sudhanshu2: Refactored it!

giactra (approved):

PR #3352, opened 2026-07-29
arch-riscv: Propagate ISA flags to vector microops

### LINKED ISSUE #2885: RISC-V Vector Instruction Micro Operations Ignore Flags
**Describe the bug**
When creating a new RISC-V Vector instruction, flags do not apply to the micro operations (e.g. "IsSerializeAfter", "IsNonSpeculative") 

**Affects version**
25.1.0.0

**gem5 Modifications**
I have simply added a new flag and a scalar and vector instruction, the flag applies to the scalar instruction, but not the micro ops the vector instruction produces.

**To Reproduce**
Open src/arch/riscv/isa/decoder.isa 
Add a flag to a vector instruction (e.g. IsSerializeAfter and IsNonSpeculative).
Perhaps print the flags of this opclass in the O3CPU inst_queue.cc when issuing.

**Terminal Output**
N/A

**Expected behavior**
When running a scalar instruction with specific flags, for example that allow serialization, the flag applies, when adding a flags to a vector instruction, the flags are expected to apply as well.
Flags assigned to a Vector instruction in the ISA description should be propagated to all generated micro-operations. For example, if a Vector instruction is marked IsSerializeAfter, the O3CPU's instruction queue should see that flag on every micro-op produced by that macro-op to ensure correct execution ordering.

**Host Operating System**
MX Linux 23.6

**Host ISA**
X86

**Compiler used**
GCC 12.2.0

**Additional information**
None

aphlta:
## Summary
- RISC-V vector `*MicroConstructor` templates intentionally omit `%(constructor)s` so they can wire per-micro register indices by hand. That also dropped `makeFlagConstructor()` output, so decoder flags such as `IsSerializeAfter` / `IsNonSpeculative` only applied to the macroop (which O3 does not execute).
- Split flag initialization into `InstObjParams.flag_constructor` and apply `%(flag_constructor)s` in all vector arith/mem MicroConstructors.

## Fixes
Fixes #2885

## Test plan
- [x] Regenerated `build/RISCV/arch/riscv/generated/decoder-ns.cc.inc`
- [x] Confirmed that adding `IsSerializeAfter, IsNonSpeculative` to `vadd_vv` causes those flags to appear in both the macro and `Vadd_vvMicro` constructors (test change not included in this PR)
- [ ] CI: RISCV build / existing RVV tests


aarmejach:
Thanks for this. 

I'd like to understand two things:
- Does this break/change current instruction implementation that assumes that not all micro-ops inherit the flags?
- Also, I do not think that conceptually we have to pass certain flags to all micros. For example, IsSerializeAfter may only be necessary for the last micro of a sequence.

@saul44203 can you take a look?

aphlta:
Thanks for the review — both points are fair.

**1) Impact on current instructions**
Today almost no `Vector*Format` entries in `decoder.isa` pass `IsSerializeAfter` / `IsNonSpeculative` (those `Is*` uses are mostly scalar CSR/fence/syscall paths). Existing vector code that needs serialize/nonspeculative already sets them manually on a specific helper, e.g. `VxsatMicroInst` in `VectorIntVxsatMacroConstructor` / narrowing. So for the current ISA description this PR should not change those hand-tuned sequences.

What *does* change for existing vector ops is mainly inheriting ordinary operand flags such as `IsVector` onto the micros (previously dropped with `%(constructor)s`).

**2) Not all flags belong on every micro**
Agreed. In particular `IsSerializeAfter` means “the next instruction is serialize-before” in O3 rename; putting it on every micro would drain the ROB between consecutive micros of the same vector instruction, which is wrong. `IsSerializeBefore` / `IsSquashAfter` are similarly boundary-oriented. `IsNonSpeculative` is more naturally whole-instruction.

I pushed an update that:
- uses `micro_flag_constructor` for non-position-sensitive flags on each `*Micro` ctor;
- after the micro sequence is built, copies `IsNonSpeculative` to **all** micros (including helpers like `VPinVd`), and applies `IsSerializeBefore` / `IsSerializeAfter` / `IsSquashAfter` only to the **first/last** micro.

Happy to adjust further if you’d rather keep serialize flags entirely manual in MacroConstructors.

aphlta:
@aarmejach Friendly ping — just checking if you had a chance to look at my earlier reply on the flag-propagation semantics / boundary microops.

Happy to adjust the patch (or add more comments/tests) if anything is still unclear. Thanks!

aphlta:
@aarmejach Gentle follow-up ping — just checking whether you had a chance to revisit the flag-propagation / boundary-microop points from my July 29 reply (and the Aug 5 note).

CI is still green on my side. Happy to rebase, add more comments/tests, or further narrow which flags land on which micros if that would help. Thanks!

aphlta:
@aarmejach Gentle follow-up ping — just checking whether you had a chance to revisit the flag-propagation / boundary-microop points from my July 29 reply (and the Aug 5 note).

CI is still green on my side. Happy to rebase, add more comments/tests, or further narrow which flags land on which micros if that would help. Thanks!

aphlta:
@aarmejach Gentle follow-up ping — just checking whether you had a chance to revisit the flag-propagation / boundary-microop points from my July 29 reply (and the Aug 5 note).

CI is still green on my side. Happy to rebase, add more comments/tests, or further narrow which flags land on which micros if that would help. Thanks!

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -131,6 +131,25 @@ template<typename ElemType>
...
     this->microops.back()->setLastMicroop();
+    // Propagate decoder StaticInst flags onto the microop sequence.
+    // IsNonSpeculative applies to every micro (including helpers).
+    // Serialize/squash flags only apply at the sequence boundary so O3
+    // does not drain the ROB between consecutive micros of one vector op.
+    if (flags[IsNonSpeculative]) {
+        for (auto &microop : this->microops) {
+            microop->setFlag(StaticInst::IsNonSpeculative);
+        }
+    }
```
saul44203: Why are we propagating this flag manually? `IsNonSpeculative` won't be filtered out in `micro_flag_constructor`, right?

src/arch/riscv/isa/templates/vector_arith.isa
```diff
@@ -395,6 +415,25 @@ template<typename ElemType>
...
+    if (flags[IsSerializeBefore]) {
+        this->microops.front()->setFlag(StaticInst::IsSerializeBefore);
+    }
+    if (flags[IsSerializeAfter]) {
+        this->microops.back()->setFlag(StaticInst::IsSerializeAfter);
+    }
+    if (flags[IsSquashAfter]) {
+        this->microops.back()->setFlag(StaticInst::IsSquashAfter);
+    }
+
```
saul44203: Can we avoid duplicating this code? Maybe inject it in the `micro_flag_constructor` (e.g. with a `makeFlagMicroConstructor`)

saul44203 (changes requested):
Hi @aphlta, thanks for the patch.

I have a couple of comments in the code.

src/arch/riscv/isa/templates/vector_arith.isa
aphlta: You're right that `IsNonSpeculative` is not filtered out of `micro_flag_constructor`, so this op's own `*Micro` ctor already sets it.

The extra stamp is only for **helper** micros that are a different `StaticInst` (`VectorNopMicroInst`, `VCpyVsMicroInst`, `VlFFTrimVlMicroOp`, …) and never run that snippet. I pulled the loop into `makeFlagMicroConstructor()` so it isn't pasted in every MacroConstructor. `setFlag` is idempotent on `*Micro`.

Hardcoded `VxsatMicroInst` flags are unchanged (instruction-internal, not decoder propagation).

src/arch/riscv/isa/templates/vector_arith.isa
aphlta: Agreed on the duplication. Added `makeFlagMicroConstructor()` / `%(flag_micro_constructor)s`, expanded in the MacroConstructor **after** `setFirstMicroop()` / `setLastMicroop()`.

I did not put serialize/squash into `micro_flag_constructor`: that runs in each micro ctor before first/last is known, and would serialize between micros of one vector op (the bug the previous commit fixed). `VseConstructor` now uses the same snippet.

aphlta:
Thanks @saul44203 — addressed in a follow-up (`f75dce91e9`):

- One generated `makeFlagMicroConstructor()` instead of the copied block.
- Serialize/squash still applied only at the sequence boundary, not in `micro_flag_constructor`.
- `IsNonSpeculative` on this op's `*Micro` remains in `micro_flag_constructor`; the generated loop is for helpers only.

Happy to drop the helper loop if you'd rather those stay speculative.

PR #3354, opened 2026-07-30
mem: Use std::vector for TracingExtension and add helpers

Harrisonust:
- Change port trace type from std::stack to std::vector for faster random access.
- Add getTraceSource helper to get the source port of a packet.
- Add dumpPortTrace helper to print the complete port trace.

### DIFF
--- a/src/mem/port.hh
+++ b/src/mem/port.hh
@@ -53,6 +53,8 @@
 #include <vector>
 
 #include "base/addr_range.hh"
+#include "base/trace.hh"
+#include "debug/PortTrace.hh"
 #include "mem/packet.hh"
 #include "mem/protocol/atomic.hh"
 #include "mem/protocol/functional.hh"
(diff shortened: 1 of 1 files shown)


src/mem/port.hh
```diff
@@ -71,54 +72,82 @@ class ResponsePort;
...
+        }
+        return trace_[0];
+    }
+
+    static void
+    dumpPortTrace(const gem5::PacketPtr pkt)
+    {
+        auto ext = pkt->getExtension<gem5::TracingExtension>();
+        if (ext) {
+            warn("%s", ext->getTraceInString());
```
powerjg: `warn` is probably not the right API to use here. 

What is the purpose of this function? If the purpose is to debug, then it should be `DPRINTF`. There are not many (if any) cases of function's printing to stdout throughout the gem5 codebase. `inform` is another option, but I don't think that is necessarily appropriate for a function.

Harrisonust: Hi @powerjg sorry for the late reply. The purpose of this function is to debug the packet. I have changed it to `DPRINTF`.

BobbyRBruce (approved):

PR #3358, opened 2026-07-31
mem-ruby: Implement draining for AbstractController

giactra:
This patch implements draining for the AsbtractController by relying on the existing map of outstanding transactions.

Ok so why are we doing this? Wasn't draining already supported in ruby?

At the moment it looks like the only logic making sure transient state is drained in ruby is in the RubyPort/Sequencer. [1] Which is effectively the input into the ruby memory sub-system. In a way it makes sense, assuming the system always reacts from stimuli passing through the RubyPort, if no outstanding request is allocatd at the RubyPort, we can infer the system is drained.

However this is not really an architected solution. There's always in fact a chance that there is some transient activity happening after a CPU/DMA initiate request has completed (and cleared the RubyPort).

Think for instance of a READ which initiates a writeback. A read might be completed and the RubyPort marks ruby as a whole as drained, while the cache controller is in the meantime in the process of writing back a potentially dirty line.

With this patch we mark a controller as draining if there are pending incoming transactions, and drained in the other case

Note 1:
I didn't come across this issue with the traditional hierarchy but by experimenting with a RubyPort-less system (with the TlmController) which as a consequence doesn't implement draining at all for the memory subsystem.

Note 2:
An initial implementation that works for CHI is provided here. Other sm controllers for other protocols have to either override the drain method according to their internal data structures or start using the incomingTransaction logic

    src/mem/ruby/system/RubyPort.cc#L608

Change-Id: I74fa32587cefae76ac716bdc9d7fd59e3489beb9

### DIFF
--- a/src/mem/ruby/slicc_interface/AbstractController.cc
+++ b/src/mem/ruby/slicc_interface/AbstractController.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017,2019-2022 ARM Limited
+ * Copyright (c) 2017,2019-2022, 2026 Arm Limited
  * All rights reserved.
  *
  * The license below extends only to copyright in the software and shall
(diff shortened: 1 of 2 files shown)


powerjg (approved):
I don't know if this will work with the original checkpointing with ruby where we "drain" by creating a new event queue, enqueue a flush for every address in the cache, run the event queue until it's done, and then switch back to the original event queue to drain.

That said, I don't think we have a ruby checkpointing test, and I don't know if anyone is depending on this behavior anymore. So, I guess this is OK even if the original checkpointing behavior isn't tested/verified.

PR #3363, opened 2026-08-03
dev: Allow a non zero sized IsaFake when ret_bad_addr=True

giactra:
For some reason whenever the ret_bad_addr Param was set
to true, the IsaFake device was programmed to have no size.
This does not allow us mark a specific portion of memory
as "do not access here and I will return a bad addr response
if you do".

I believe the motivation for this is that a classic
XBar will use a BadAddr device as a default responder whenever
it is not able to map an incoming address to a port

default = RequestPort("Port for connecting an optional default responder")
badaddr_responder = BadAddr()
default = badaddr_responder.pio

As such it does not require BadAddr to have a real address range
(pio_addr, pio_size).
Its range is actually: "whatever is not recognised".

I believe this is not a property per se of the IsaFake device,
so we shouldn't carry the BadAddr device usage with a XBar
into the initialization of a IsaFake. Therefore we now simplify
IsaFake to make it less obscure, and we simply manually
initialize the BadAddr device to have a pio_size=0 in python

### DIFF
--- a/src/dev/Device.py
+++ b/src/dev/Device.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2012-2016,2019-2020 ARM Limited
+# Copyright (c) 2012-2016,2019-2020,2026 Arm Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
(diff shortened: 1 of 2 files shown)


powerjg (approved):
I'm a little surprised the prior code worked. You shouldn't be able to define a new `Param` in a class that isn't a "SimObject Description Class" (i.e., doesn't have a `cxx_class`). 

Also, as always, I don't like having non "SimObject Description Classes" (i.e., Python classes without a `cxx_class` that are simply specialized instances of a SimObject) in src. There will be a day that we purge this anti-pattern. 🙂

dhschall (approved):

PR #3365, opened 2026-08-04
stdlib: Allow independent setting of CPU frequency

giactra:
One of the many things preventing the widespread adoption of stdlib is that it starts from the unreal assumption that all components in a simulator will be part of the same clock domain.

At the moment most stdlib platforms (with some exceptions) assume every component from CPU to caches and interconnect will use the board clock_domain.

With this commit we allow setting a separate clk_domain for the Processor/Cores.

We add a core_frequency parameter to the AbstractProcessor which is using to define a new Processor clk_domain. We delegate to parent classes to attach this clk_domain to the child cores. According to gem5 magic this might not be needed but relying on Parent.clk_domain assignment is too risky when there is such an important parameter to be set. We leve the board.clk_domain to clock the rest of the board, meaning interconnects,devices and so on.

Change-Id: Id44f183f8ba9106eb03c11a548a72f1a624295dd

### DIFF
--- a/configs/deprecated/example/gem5_library/riscv-fs.py
+++ b/configs/deprecated/example/gem5_library/riscv-fs.py
@@ -66,7 +66,10 @@ memory = SingleChannelDDR3_1600()
 
 # Setup a single core Processor.
 processor = SimpleProcessor(
-    cpu_type=CPUTypes.TIMING, isa=ISA.RISCV, num_cores=1
+    cpu_type=CPUTypes.TIMING,
+    isa=ISA.RISCV,
+    num_cores=1,
+    clk_freq="1GHz",
 )
 
 # Setup the board.
--- a/configs/example/arm/fdp_neoverse_v2.py
+++ b/configs/example/arm/fdp_neoverse_v2.py
@@ -223,7 +223,8 @@ cache_hierarchy = CacheHierarchy()
 # Create the processor with one core
 
 processor = BaseCPUProcessor(
-    cores=[BaseCPUCore(neoverse_v2.NeoverseV2(), isa=ISA.ARM)]
+    cores=[BaseCPUCore(neoverse_v2.NeoverseV2(), isa=ISA.ARM)],
+    clk_freq="3GHz",
 )
 
 for core in processor.cores:
--- a/configs/example/gem5_library/arm-hello.py
+++ b/configs/example/gem5_library/arm-hello.py
@@ -62,7 +62,9 @@ cache_hierarchy = NoCache()
 memory = SingleChannelDDR3_1600(size="32MiB")
 
 # We use a simple Timing processor with one core.
-processor = SimpleProcessor(cpu_type=CPUTypes.TIMING, isa=ISA.ARM, num_cores=1)
+processor = SimpleProcessor(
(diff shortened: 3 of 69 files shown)


powerjg (commented):
I agree with the need for this change and the main idea that a frequency can be set for the processor that is different from the board. However, I want to make sure that our users can't easily shoot themselves in the foot and create nonsensical systems under simulation.

1. Can we make the `clk_freq` optional to the processor and if it's not provided use the board's frequency? I feel like this is a common case (clearly it is currently in the codebase, and I think it will be in the future as well). It will also reduce this diff significantly.
2. I feel that the clock of processor should be a `DerivedClockDomain` of the board. Otherwise, there may be bad clock domain crossings that are not noticed by the author of the configuration file.
3. The other concern I have is with the core to L1 clock domain. The core and the L1/L2 should really always be in the same clock domain. However, I believe (correct me if I'm wrong), that with this implementation, since the `CacheHierachy` will get the board's clock, the L1 clock frequency could be very different from the core clock frequency.

giactra:
> I agree with the need for this change and the main idea that a frequency can be set for the processor that is different from the board. However, I want to make sure that our users can't easily shoot themselves in the foot and create nonsensical systems under simulation.
> 
>     1. Can we make the `clk_freq` optional to the processor and if it's not provided use the board's frequency? I feel like this is a common case (clearly it is currently in the codebase, and I think it will be in the future as well). It will also reduce this diff significantly.

Well I actually believe your second paragraph is going against the first one. By having default values with optional parameters and making it default to the board clock we are making it more likely for a user to create a non-sensical system. We should force them to explcitly choose a Board clock and a Processor clock, and fail if this is not happening.


> 
>     2. I feel that the clock of processor should be a `DerivedClockDomain` of the board. Otherwise, there may be bad clock domain crossings that are not noticed by the author of the configuration file.

I theoretically agree with this statement. However a DerivedClockDomain can only have sub-frequencies of the parent clock domain. This is not the case. Processors are usually clocked at a higher frequency than the rest of the system.

> 
>     3. The other concern I have is with the core to L1 clock domain. The core and the L1/L2 should really always be in the same clock domain. However, I believe (correct me if I'm wrong), that with this implementation, since the `CacheHierachy` will get the board's clock, the L1 clock frequency could be very different from the core clock frequency.

Yes that's absolutely right. I have another patch which is amending the stdlib cachehierarchies to tie the private caches to the processor.get_clock_domain. I didn't push it together to simplify the reviewing. Happy to do so if needed

powerjg:
>  We should force them to explcitly choose a Board clock and a Processor clock, and fail if this is not happening.

Ok. I'm on board with this.

> However a DerivedClockDomain can only have sub-frequencies of the parent clock domain.

Ah, that make sense. It's OK with me to have separate domains then. If we care about getting DVFS right, we'll have to go back and refactor a lot of things, I guess.

> Yes that's absolutely right. I have another patch which is amending the stdlib cachehierarchies to tie the private caches to the processor.get_clock_domain. I didn't push it together to simplify the reviewing. Happy to do so if needed

I'm OK with staging this, but with the release coming up, we need to make sure that the "feature" isn't half in the release.

powerjg (approved):

PR #3369, opened 2026-08-05
mem-ruby: Handle snoops in the TlmController

giactra:
With this commit we properly handle snoop requests at the TlmController:

1) We start translating relevant fields like:
* ret_to_src
* src_id
* fwd_nid

2) When responding to a snoop, we make sure we forward the message to the right controller. We cannot use the mapAddressToDownstreamMachine as usual because that would direct the message to the HN.
We use the tgt_id instead, assuming the RN implementation is properly initializing it to be the snoop request src_id (for responses to the HN) or the fwd_nid (for responses to the RN)

Change-Id: I4c30bc294757beb719cf58705ed8e91fb7aba919

### DIFF
--- a/src/mem/ruby/protocol/chi/tlm/utils.cc
+++ b/src/mem/ruby/protocol/chi/tlm/utils.cc
@@ -308,13 +308,26 @@ datOpcode(DatOpcode dat, Resp resp)
         }
       case DAT_OPCODE_SNP_RESP_DATA:
         RESP_CASE(CHIDataType_SnpRespData)
+      case DAT_OPCODE_COMP_DATA:
+          switch (resp) {
+              case RESP_I:
+                  return CHIDataType_CompData_I;
+              case RESP_UC:
+                  return CHIDataType_CompData_UC;
+              case RESP_SC:
+                  return CHIDataType_CompData_SC;
+              case RESP_SD_PD:
+                  return CHIDataType_CompData_SD_PD;
+              default:
+                  panic("");
+          }
       default:
         panic("Unsupported Translation: %s\n", datOpcodeToName(dat));
     }
 }
 
 CHIResponseType
-rspOpcode(RspOpcode opc, Resp resp)
+rspOpcode(RspOpcode opc, Resp resp, Resp fwd_state)
 {
     switch(opc) {
       case RSP_OPCODE_COMP_ACK: return CHIResponseType_CompAck;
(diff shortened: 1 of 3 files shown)


src/mem/ruby/protocol/chi/tlm/utils.cc
```diff
@@ -323,8 +336,57 @@ rspOpcode(RspOpcode opc, Resp resp)
       case RSP_OPCODE_SNP_RESP:
         switch (resp) {
           case RESP_I: return CHIResponseType_SnpResp_I;
+          case RESP_SC:
+              return CHIResponseType_SnpResp_SC;
           default: panic("Invalid resp %d for %d\n", resp, opc);
         }
+      case RSP_OPCODE_SNP_RESP_FWDED:
```
powerjg, line 343: No action required, but I wonder if this nested switch/case is getting a little unwieldy. AI suggests that a table would be an improved design. Below is a code snippet. Just something to think about as this grows with more states/transitions.

```
static constexpr auto Invalid = CHIResponseType_NULL_; // pick an actual invalid sentinel

static std::array<std::array<std::array<CHIResponseType, RespCount>,
                             RespCount>,
                  RspOpcodeCount> rsp_tbl = []{
    decltype(rsp_tbl) t{};
    for (auto &a : t) for (auto &b : a) for (auto &c : b) c = Invalid;

    t[RSP_OPCODE_SNP_RESP][RESP_I][RESP_I]   = CHIResponseType_SnpResp_I;
    t[RSP_OPCODE_SNP_RESP][RESP_SC][RESP_I]  = CHIResponseType_SnpResp_SC;

    t[RSP_OPCODE_SNP_RESP_FWDED][RESP_I][RESP_UC]    = CHIResponseType_SnpResp_I_Fwded_UC;
    t[RSP_OPCODE_SNP_RESP_FWDED][RESP_I][RESP_UD_PD] = CHIResponseType_SnpResp_I_Fwded_UD_PD;
    t[RSP_OPCODE_SNP_RESP_FWDED][RESP_SC][RESP_I]    = CHIResponseType_SnpResp_SC_Fwded_I;
    // ...
    return t;
}();

CHIResponseType rspOpcode(RspOpcode opc, Resp resp, Resp fwd_state)
{
    auto out = rsp_tbl[opc][resp][fwd_state];
    if (out != Invalid) return out;
    panic("Invalid rsp transition opc=%d resp=%d fwd=%d\n", opc, resp, fwd_state);
}
```

powerjg (approved):

src/mem/ruby/protocol/chi/tlm/utils.cc
giactra: Yeah I am not a big fan either.
The problem with a map/matrix approach is that most of the entries are actually a duplication; only forwarding snoops require the forwarding state.. al the others require resp only... I think another solution could be to separate forwarding snoops from the bunch and treat the rest as a map...

PR #3374, opened 2026-08-07
mem-ruby: Handle snoops in the TlmGenerator

giactra:
This commit adds a basic SnoopHandlder interface to the
TlmGenerator which is in charge of responding to any
snoop the TlmGenerator might receive.
    
This is not mandatory for a TlmGenerator: there are cases
where we don't expect a snoop and in such cases having
a snoop handler in un-necessary. Snoops can be handled
in many different ways.
Someone could potentially write a very simple SnoopHandler
which simply assumes there is no line present in the
modelled RN. Some others could for example model a rudimentary
cache which keeps track of filled lines and return data
accordingly.
    
In this commit we propose a SnoopHandler implementation (PySnoopHandler)
following the same principle behind the TlmGenerator; similar to
Transactions which are defined and injected from python, we define
SnoopResponses whose handling logic resides purely in python.
There is no internal state inside the PySnoopHandler.
We use the following API to generate SnoopResponse objects and return a
handle to the response.
    
    snoop = snoop_handler.add_snoop(address)
    
This will inform the SnoopHandler it is expected to receive a
SNP request at the provided address. A user should then attach
Transaction::Actions to the SnoopResponse which will entirely
dictate how the handler will react to such snoop.

### DIFF
--- a/src/mem/ruby/protocol/chi/tlm/SnoopHandler.py
+++ b/src/mem/ruby/protocol/chi/tlm/SnoopHandler.py
@@ -1,5 +1,5 @@
 # -*- mode:python -*-
-# Copyright (c) 2024-2026 Arm Limited
+# Copyright (c) 2026 Arm Limited
 # All rights reserved.
 #
 # The license below extends only to copyright in the software and shall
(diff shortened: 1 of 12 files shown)


src/mem/ruby/protocol/chi/tlm/SnoopHandler.py
```diff
@@ -0,0 +1,93 @@
...
+        phase.opcode = RspOpcode.SNP_RESP
+        phase.resp = Resp.RESP_I
+
+        transaction = SnoopResponse(phase)
+        self._snoop_transactions.append((address, transaction))
+        if self._cc_initialized:
+            self.getCCObject().addSnoop(address, transaction)
+        return transaction
+
+    def init(self):
```
powerjg: I believe this is going to conflict with #3290

powerjg (approved):
lgtm.

src/mem/ruby/protocol/chi/tlm/SnoopHandler.py
harry900831: Hi @giactra, can you help to check whether it's capable to override createCCObject instead of init? It seems to be fine. Thanks in advance!

giactra: Yeah I don't think there should be a problem. Let me try it out

powerjg (dismissed):
lgtm, just be careful with the ordering of this PR and the other PR that changes the `init` function in SimObject's

giactra:
I have rebased on top of the init change

powerjg (approved):

PR #3379, opened 2026-08-07
sim: add process restoration syscalls and debug support

adhilton:
Add system call emulation handlers for dup3 and certain prctl behaviors in gem5 SE mode.   This also increases the max PID, to allow running PID-sensitive workloads with realistic PIDs.  Note that there is one unrelated variable name change (to make the parameters names in the function implementation match them in the header file), and two additions of "const" which were done by code tidyness tools.


src/sim/syscall_emul.hh
```diff
@@ -375,6 +379,37 @@ atSyscallPath(ThreadContext *tc, int dirfd, std::string &path)
...
+SyscallReturn
+prctlFunc(SyscallDesc *desc, ThreadContext *tc, int option, uint64_t arg2,
+          uint64_t arg3, uint64_t arg4, uint64_t arg5)
+{
+    auto p = tc->getProcessPtr();
+    std::shared_ptr<MemState> mem_state = p->memState;
+
+    if (option == OS::TGT_PR_SET_MM) {
+        switch (arg2) {
+            case OS::TGT_PR_SET_MM_START_BRK:
```
powerjg, line 393: Should we have a warning that we are not actually setting the brk start location? Could this cause restored applications to fail if there is a pointer to the stack given the stack may not start at the same point?

adhilton: This changes the lower (starting) boundary of the heap, unlike the next case TGT_PR_SET_MM_BRK, which changes the higher (ending) boundary of the heap.  The later is used for the brk() system call.  As far as I can tell, the former is not used anywhere---it does not appear to be tracked in MemState.

powerjg: My question is whether if the guest user-mode code calls this system call, if it will break given that we ignore it. If there is a chance that the guest code is going to break, we should at least add a warning so the gem5 user is aware of the possible problem.

powerjg:
There are some errors in the CI test

```
In file included from src/arch/arm/linux/se_workload.cc:51:
src/sim/syscall_emul.hh: In instantiation of ‘gem5::SyscallReturn gem5::prctlFunc(SyscallDesc*, ThreadContext*, int, uint64_t, uint64_t, uint64_t, uint64_t) [with OS = ArmLinux32; uint64_t = long unsigned int]’:
src/arch/arm/linux/se_workload.cc:499:12:   required from here
src/sim/syscall_emul.hh:391:23: error: ‘TGT_PR_SET_MM’ is not a member of ‘gem5::ArmLinux32’
  391 |     if (option == OS::TGT_PR_SET_MM) {
      |                       ^~~~~~~~~~~~~
src/sim/syscall_emul.hh:393:22: error: ‘TGT_PR_SET_MM_START_BRK’ is not a member of ‘gem5::ArmLinux32’
  393 |             case OS::TGT_PR_SET_MM_START_BRK:
      |                      ^~~~~~~~~~~~~~~~~~~~~~~
src/sim/syscall_emul.hh:398:22: error: ‘TGT_PR_SET_MM_BRK’ is not a member of ‘gem5::ArmLinux32’
  398 |             case OS::TGT_PR_SET_MM_BRK:
      |                      ^~~~~~~~~~~~~~~~~
src/sim/syscall_emul.hh: In instantiation of ‘gem5::SyscallReturn gem5::prctlFunc(SyscallDesc*, ThreadContext*, int, uint64_t, uint64_t, uint64_t, uint64_t) [with OS = ArmLinux64; uint64_t = long unsigned int]’:
src/arch/arm/linux/se_workload.cc:854:38:   required from here
src/sim/syscall_emul.hh:391:23: error: ‘TGT_PR_SET_MM’ is not a member of ‘gem5::ArmLinux64’
  391 |     if (option == OS::TGT_PR_SET_MM) {
      |                       ^~~~~~~~~~~~~
src/sim/syscall_emul.hh:393:22: error: ‘TGT_PR_SET_MM_START_BRK’ is not a member of ‘gem5::ArmLinux64’
  393 |             case OS::TGT_PR_SET_MM_START_BRK:
      |                      ^~~~~~~~~~~~~~~~~~~~~~~
src/sim/syscall_emul.hh:398:22: error: ‘TGT_PR_SET_MM_BRK’ is not a member of ‘gem5::ArmLinux64’
  398 |             case OS::TGT_PR_SET_MM_BRK:
      |                      ^~~~~~~~~~~~~~~~~
 [     CXX] src/arch/amdgpu/vega/pagetable_walker.cc -> ALL/arch/amdgpu/vega/pagetable_walker.fo
 [     CXX] ALL/python/_m5/param_VGic.cc -> .fo
```

erin-le:
@adhilton: Hello, could you take a look at the comments above, and also look into the failing CI tests when you get a chance? Thank you!

adhilton:
@erin-le Yes, sorry---between start of semester (and teaching a new class) and a few other major things that have come up, this is on the back burner.  I am hoping to work on this again in the next week or two.

PR #3380, opened 2026-08-07
cpu-probe: add hot PC profiling to PcCountTracker

adhilton:
This PR extends PcCountTracker (which provides support for exiting simulation when a target PC is executed a specified number of times) to report the hottest PCs during the simulation. This new feature allows users to find regions of code to focus performance debugging/analysis work on. The new mode for the tracker supports filtering by PC region and bucketing PCs together at a specific granularity in case the experimenter is worried about the space required to store hot PC information when profiling a large region of code.

### DIFF
--- a/src/cpu/probes/PcCountTracker.py
+++ b/src/cpu/probes/PcCountTracker.py
@@ -59,6 +59,21 @@ class PcCountTracker(ProbeListenerObject):
     cxx_header = "cpu/probes/pc_count_tracker.hh"
     cxx_class = "gem5::PcCountTracker"
 
+    cxx_exports = [
+        PyBindMethod("getHottestPcs"),
+    ]
+
     targets = VectorParam.PcCountPair("the target PC Count pairs")
     core = Param.BaseCPU("the connected cpu")
-    ptmanager = Param.PcCountTrackerManager("the PcCountTracker manager")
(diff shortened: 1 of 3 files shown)


powerjg:
Looks like a real failure from the tests:

```
src/cpu/probes/pc_count_tracker.hh:53:18: error: 'regProbeListeners' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
 [     CXX] ALL/mem/ruby/protocol/CHI/Cache_RetryQueueEntry.cc -> .fo
   53 |     virtual void regProbeListeners();
      |                  ^
src/sim/sim_object.hh:232:18: note: overridden virtual function is here
  232 |     virtual void regProbeListeners();
      |                  ^
1 error generated.
```

powerjg:
tagging @studyztp as well

erin-le:
@adhilton: Hello, thanks for this PR. Could you take a look at the failing macOS compilation tests? Also, could you take a look at the comments from Copilot and address them if necessary? Thank you!

adhilton:
@erin-le Yes, sorry---between start of semester (and teaching a new class) and a few other major things that have come up, this is on the back burner.  I am hoping to work on this again in the next week or two.

PR #3381, opened 2026-08-07
arch-arm: Allow EL0 reads of ID registers in SE mode

adhilton:
When executing ARM64 binaries (such as SPEC CPU 2017 benchmarks or glibc startup routines) in gem5 System-call Emulation (SE) mode, some programs execute `mrs Xt, <sysreg>` instructions (e.g., `mrs x1, midr_el1`) to query CPU ID and capability registers.

Unlike real hardware or OS-managed environments where the Linux kernel in EL1 fields the trap and emulates these ID register reads for user-space processes (EL0), gem5 SE mode runs without an OS kernel and must directly emulate this behavior.

We update `faultIdst` in `src/arch/arm/regs/misc.cc` to immediately return `NoFault` when running in non-full system (`!FullSystem`) mode. This allows EL0 instructions in SE mode to read architectural ID registers (such as `MIDR_EL1`) without generating an architectural trap or undefined instruction fault.  Note that this follows the same pattern as many other instructions which trap in the same way on real hardware.

### DIFF
--- a/src/arch/arm/regs/misc.cc
+++ b/src/arch/arm/regs/misc.cc
@@ -3559,6 +3559,9 @@ Fault
 faultIdst(const MiscRegLUTEntry &entry,
     ThreadContext *tc, const MiscRegOp64 &inst)
 {
+    if (!FullSystem) {
+        return NoFault;
+    }
     if (HaveExt(tc, ArmExtension::FEAT_IDST)) {
         const HCR hcr = tc->readMiscReg(MISCREG_HCR_EL2);
         if (EL2Enabled(tc) && hcr.tge) {


BobbyRBruce (approved):

giactra:
Nice to see Arm changes merged quickly with no ping and without the review of the Arm maintainer...

Can I ask post-mortem some details on the compilation-toolchain used?
Also on the list of ID registers that are being accessed from EL0.
As far as I know the default way of a Linux system to expose ID registers to userspace is via HWCAPS. https://elixir.bootlin.com/linux/v7.1.7/source/Documentation/arch/arm64/elf_hwcaps.rst

We do populate them at init time in gem5 so a program should not be directly reading from the ID regs.
Not that this solution is unreasonable. The "risk" of exposing ID regs to userspace is pretty minimal but I would like to get to the bottom of what's happening

adhilton:
This was with clang, though it is worth noting that it was from a memory state captured from real hardware.  Accordingly, the auxvector had the values provided by the Linux kernel on realhardware, not constructed by gem5.

powerjg:
This is similar to https://github.com/gem5/gem5/pull/2998/changes/a18982c7904a3c9e9a6a254ceb487cfd8bede162

From my understand, and @adhilton can correct me if I'm wrong, in Arm there are a bunch of instructions (msr mostly) where usermode often uses the instruction (e.g., in glibc), but the instructions are actually EL0/privileged. When running in FS mode or on real systems, the kernel does a trap and emulate on those instructions.

Trap and emulate doesn't work in SE mode. So, we need some way, in SE mode, to make execute the instructions and not throw faults when they are executed.

This change makes it so the instructions execute as if EL0 and they don't throw faults. But this only happens in SE mode (not in FS mode). 

Does this make sense? Is there a better solution?

giactra:
> This is similar to [a18982c](https://github.com/gem5/gem5/commit/a18982c7904a3c9e9a6a254ceb487cfd8bede162)
> 
> From my understand, and @adhilton can correct me if I'm wrong, in Arm there are a bunch of instructions (msr mostly) where usermode often uses the instruction (e.g., in glibc), but the instructions are actually EL0/privileged. When running in FS mode or on real systems, the kernel does a trap and emulate on those instructions.
> 

Trap-and-emulate feels like a heavyweight solution for MOST ID registers (if not all).
As I mentioned in my message above, all ID regs like ID_AA64PFR0_EL1 should be accessible via auxiliary vectors populated by the kernel at startup time when loading an ELF.
Now MIDR_EL1/MPIDR_EL1 do not fall into that category so I can see the kernel trapping and emulating them.

So it is very likely that SOME regs require the EL0 relaxation, but I am sure not ALL of them. A cleaner solution would have been to factor a different trapping callback for those registers (and that's why I asked if @adhilton knew which were the culprits)

> Accordingly, the auxvector had the values provided by the Linux kernel on realhardware, not constructed by gem5.

This could make sense but in SE we will write the auxvector ourselves anyway when we load the image so unless you are doing some magic afterwards with the guest stack it shouldn't be the case. Also I am expecting HW to have a populated auxvector anyway; probably with different values, but still an auxvector. But not knowing what you are doing in your simulation it is hard for me to say

> This change makes it so the instructions execute as if EL0 and they don't throw faults. But this only happens in SE mode (not in FS mode).
> 
> Does this make sense? Is there a better solution?

IMHO the better solution is to allow from EL0 only registers which are not readable via HWCAPS. If a register is directly accessed, I'd be curious to know why. If the motivation is that we want to be able to run experimental bootcode or libc implementation which are not making use of HWCAPS, then I am fine with that as long as it is intentional AND documented.

adhilton:
It was midr_el1 in particular.  This happens in init_cpu_features in the glibc startup code.

PR #3389, opened 2026-08-11
arch-x86: fix IOPL extraction in the STI and CLI microcode

### LINKED ISSUE #2912: arch-x86: STI and CLI compute IOPL incorrectly
**Describe the bug**
The microcode for the `sti` and `cli` instructions compute the current IOPL from `rflags` incorrectly, due to an apparent typo. This means that, when the carry flag is set (e.g., with `stc`), userspace programs (with CPL=3) can execute `cli` without faulting in FS mode, which is architecturally incorrect.

Test program (`test.asm`) that doesn't crash when it should:
```asm
    global _start    
                     
_start:              
    stc              
    cli              
                     
    mov eax, 60      
    mov edi, 0       
    syscall
```

**Affects version**
develop @ [eacdc6e](https://github.com/gem5/gem5/commit/eacdc6e3fa6d76dacc1ee64b4c609c44dbd12da5)

**gem5 Modifications**
None

**To Reproduce**
2. Assemble the test program: `nasm -felf64 -o test.o test.asm && ld -o test test.o`
3. Run the test program with the attached script ([test.py](https://github.com/user-attachments/files/24897527/test.py)): `build/X86/gem5.opt test.py --binary ./test`

**Terminal Output**
```shell
...
In after_boot.sh...
Interactive mode: false
Starting gem5 init... trying to read run script file via readfile.
[    1.665642] gem5_bridge_open: SUCCESS!
[    1.667716] gem5_bridge_mmap: SUCCESS!
[    1.763248] gem5_bridge_release: SUCCESS!
Running script from gem5-bridge stored in /tmp/script
Done running script from gem5-bridge, exiting.
[    2.542842] gem5_bridge_open: SUCCESS!
[    2.544799] gem5_bridge_mmap: SUCCESS!
```

**Expected behavior**
The program should crash with a segfault (#GP fault).

**Host Operating System**
Ubuntu 24.04

**Host ISA**
X86

**Compiler used**
GCC 13

**Additional information**
Will submit fix in a moment.

phverg:
## Summary

Fixes #2912.

The `STI` and `CLI` macroops extract the current IOPL from RFLAGS to decide whether a `CPL > IOPL` access must raise `#GP`. The sequence shifts RFLAGS right by 12 to align the IOPL field (bits 13:12) into the low two bits, but then masks the *original* RFLAGS value instead of the shifted one:

```
srli t2, t1, 12, dataSize=8
andi t2, t1, 0x3, dataSize=8   # masks raw RFLAGS, not the shifted value
```

So the computed "IOPL" is actually `RFLAGS & 0x3` = reserved bit 1 (always set) OR the carry flag, i.e. 2 or 3, rather than the real IOPL. A CPL=3 program that sets CF (for example with `stc`) can therefore execute `STI`/`CLI` without taking the required `#GP` fault. This is exactly what breaks booting `configs/example/gem5_library/x86-ubuntu-run.py` as noted in the issue: a low-IOPL library that executes `cli` is wrongly permitted.

The fix masks the shifted value (`t2`) instead, matching the CPL extraction a few lines below which already does shift-then-mask correctly (`srli t3, t3, 4; andi t3, t3, 0x3`).

## Verification

Using the reporter's reproducer as a freestanding SE-mode binary (gem5 SE mode runs x86 user code at CPL=3):

```asm
_start:
    stc
    cli
    mov $60, %rax
    xor %rdi, %rdi
    syscall
```

- **Before (unpatched):** the program runs to completion and exits 0 — the `cli` is wrongly permitted.
- **After (patched):** gem5 reports `fault (General-Protection) detected @ PC (0x401001=>0x401002)` at the `cli` instruction — the architecturally correct `#GP`.

A control binary that performs only the `exit` syscall (no `cli`) exits cleanly on both builds, confirming the change only affects the privileged path.

### DIFF
--- a/src/arch/x86/isa/insts/general_purpose/flags/set_and_clear.py
+++ b/src/arch/x86/isa/insts/general_purpose/flags/set_and_clear.py
@@ -71,7 +71,7 @@ def macroop STI {
 
     # Extract the IOPL.
     srli t2, t1, 12, dataSize=8
-    andi t2, t1, 0x3, dataSize=8
+    andi t2, t2, 0x3, dataSize=8
 
     # Find the CPL.
     rdm5reg t3, dataSize=8
@@ -129,7 +129,7 @@ def macroop CLI {
 
     # Extract the IOPL.
     srli t2, t1, 12, dataSize=8
-    andi t2, t1, 0x3, dataSize=8
+    andi t2, t2, 0x3, dataSize=8
 
     # Find the CPL.
     rdm5reg t3, dataSize=8


powerjg:
Can you comment more on:

> This is exactly what breaks booting configs/example/gem5_library/x86-ubuntu-run.py

powerjg (approved):

phverg:
That line was paraphrasing @nmosier's comment on the issue. I verified the fix in SE mode (`stc; cli` at CPL=3 raises `#GP` with the patch and not without it), but the `x86-ubuntu-run.py` boot failure is his observation, not mine.

PR #3396, opened 2026-08-14
mem: Refactor TracingExtension and defer string formatting

Harrisonust:
1. Move TracingExtension implementation details from port.hh to port.cc.
2. Store TraceEntry structs (pointers and address) instead of formatted strings in TracingExtension. When PortTrace is enabled, numerous traces are added to the container. Deferring string formatting to getTraceInString avoids unnecessary string allocations and constructions for traces that are never printed.
3. Relax logic in RequestPort::removeTrace to issue a warning instead of panicking if TracingExtension is absent, allowing PortTrace to be enabled dynamically during simulation.

### DIFF
--- a/src/mem/port.cc
+++ b/src/mem/port.cc
@@ -59,12 +59,9 @@ TracingExtension::clone() const
 }
 
 void
-TracingExtension::add(const RequestPort *request_port,
-                      const ResponsePort *response_port, gem5::Addr addr)
+TracingExtension::add(const RequestPort *request_port, gem5::Addr addr)
 {
-    trace_.push_back(TraceEntry{.request_port = request_port,
-                                .response_port = response_port,
-                                .addr = addr});
+    trace_.push_back(TraceEntry{.request_port = request_port, .addr = addr});
 }
 
 void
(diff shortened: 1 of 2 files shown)


harry900831:
@erin-le 

Friendly ping for the review! Thanks.

src/mem/port.cc
```diff
@@ -52,6 +52,75 @@
...
+        }
+    }
+    port_trace << ")";
+    return port_trace.str();
+}
+
+std::optional<std::string>
+TracingExtension::getTraceSource() const
+{
+    if (trace_.empty() || !trace_[0].request_port) {
```
giactra, line 109: Is there a case where the request_port would point to nullptr?

giactra (commented):
Hi @harry900831 , looks good to me, I only have one comment

src/mem/port.cc
harry900831: I don't really come up with an example that request_port is nullptr. The check over here is just like a safe guard to make sure the below trace_[0].request_port->name(); won't dereference a nullptr.

PR #3400, opened 2026-08-17
base: Add support for Zstd compressed images

giactra:
When loading a binary into guest memory, we support ELFs
and GZIPPED ELFS. The latter is particularly useful as
it allows people to store huge binaries in compressed format.

With this commit we add support for ZSTD compressed images [1]
This is an optional dependency (differently from zlib)

[1]: https://github.com/facebook/zstd

### DIFF
--- a/src/base/loader/SConscript
+++ b/src/base/loader/SConscript
@@ -1,5 +1,17 @@
 # -*- mode:python -*-
 
+# Copyright (c) 2026 Arm Limited
+# All rights reserved
+#
+# The license below extends only to copyright in the software and shall
+# not be construed as granting a license to any other intellectual
+# property including but not limited to intellectual property relating
+# to a hardware implementation of the functionality of the software
+# licensed hereunder.  You may use the software subject to the license
+# terms below provided that you ensure that this notice is replicated
+# unmodified and in its entirety in all distributions of the software,
+# modified or unmodified, in source code or in binary form.
+#
 # Copyright (c) 2006 The Regents of The University of Michigan
 # All rights reserved.
 #
@@ -28,10 +40,18 @@
 
 Import('*')
 
+Source('compression_file_format.cc')
 Source('dtb_file.cc')
 Source('elf_object.cc')
+Source('gzip_compression.cc')
 Source('image_file_data.cc')
-GTest('image_file_data.test', 'image_file_data.test.cc', 'image_file_data.cc')
+image_file_data_test_sources = [
+    'image_file_data.test.cc',
+    'image_file_data.cc',
+    'compression_file_format.cc',
+    'gzip_compression.cc',
+]
+GTest('image_file_data.test', *image_file_data_test_sources)
 Source('memory_image.cc')
 Source('object_file.cc')
 Source('symtab.cc')
--- a/src/base/loader/SConscript
+++ b/src/base/loader/SConscript
@@ -51,6 +51,9 @@ image_file_data_test_sources = [
     'compression_file_format.cc',
     'gzip_compression.cc',
 ]
+if env['CONF']['HAVE_ZSTD']:
+    Source('zstd_compression.cc')
+    image_file_data_test_sources.append('zstd_compression.cc')
 GTest('image_file_data.test', *image_file_data_test_sources)
 Source('memory_image.cc')
 Source('object_file.cc')
--- a/src/base/loader/compression_file_format.cc
+++ b/src/base/loader/compression_file_format.cc
@@ -38,14 +38,16 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include "base/loader/image_file_data.hh"
+#include "base/loader/compression_file_format.hh"
 
-#include <fcntl.h>
-#include <sys/mman.h>
-#include <sys/types.h>
 #include <unistd.h>
 
-#include "base/loader/compression_file_format.hh"
+#include <algorithm>
+#include <cstdio>
+#include <cstring>
+#include <vector>
+
+#include "base/loader/gzip_compression.hh"
 #include "base/logging.hh"
 
 namespace gem5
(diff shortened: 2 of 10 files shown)


BobbyRBruce (changes requested):
I notice Zstd and Gzip libraries are included in Boost. We already have Boost as an optional dependency, so if we utilized Boost for gzip and zstd compressed image support we could avoid expanding our set of optional dependencies and just state that Boost is required for this.

Utilizing the Boost libraries would also avoid us rolling our own gzip and zstd comp/decomp like we do here.

powerjg:
>I notice Zstd and Gzip libraries are included in Boost. We already have Boost as an optional dependency, so if we utilized Boost for gzip and zstd compressed image support we could avoid expanding our set of optional dependencies and just state that Boost is required for this.
> 
> Utilizing the Boost libraries would also avoid us rolling our own gzip and zstd comp/decomp like we do here.

Is this right? I don't think we use Boost for anything? (Where did the AI come up with this?)

I do NOT want to add Boost as a required dependency for a number of reasons.

There are a couple of small things from copilot that seem reasonable. Otherwise, this PR looks good to me.

src/base/loader/gzip_compression.cc
```diff
@@ -0,0 +1,112 @@
...
+        if (!writeAll(fd, buf, r)) {
+            delete[] buf;
+            gzclose(fdz);
+            close(fd);
+            return -1;
+        }
+    }
+    delete[] buf;
+    gzclose(fdz);
+    if (r < 0) {
```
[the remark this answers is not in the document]
giactra: This makes sense, but I believe this deserves another PR (I simply copy-pasted gzip handling logic)

src/base/loader/SConscript
```diff
@@ -28,10 +40,21 @@
 
 Import('*')
 
+Source('compression_file_format.cc')
 Source('dtb_file.cc')
 Source('elf_object.cc')
+Source('gzip_compression.cc')
```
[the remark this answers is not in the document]
giactra: Smart, would have not thought about libgem5 in a million years

powerjg (dismissed):

giactra:
> I notice Zstd and Gzip libraries are included in Boost. We already have Boost as an optional dependency, so if we utilized Boost for gzip and zstd compressed image support we could avoid expanding our set of optional dependencies and just state that Boost is required for this.
> 
> Utilizing the Boost libraries would also avoid us rolling our own gzip and zstd comp/decomp like we do here.

Hi Bobby,

I agree with Jason here, we have historically tried to stay away from boost as a dependency. In general I feel less dependencies are better for the project. Using low level APIs might also be a requirement (but I don't know boost enough) the moment we decide we want to implement something more complex like seekable frames (with zstd).
With a bit of restructuring it might allow us to decompress and load portions of a guest binary on demand (when we see a memory access to it).

Please let me know what you think

src/base/loader/compression_file_format.hh
```diff
@@ -35,62 +35,41 @@
...
-
-    a->setLastEnqueueTime(100);
-    b->setLastEnqueueTime(100);
 
-    a->setTransactionTime(10);
-    b->setTransactionTime(20);
-
-    a->setMsgCounter(999);
-    b->setMsgCounter(1);
+class CompressionFileFormat
```
odanrc, line 67: Could you document this class and functions, specially writeAll, whose signature is not obvious (the boolean return)?

src/base/loader/compression_file_format.cc
```diff
@@ -0,0 +1,157 @@
...
+typedef std::vector<CompressionFileFormat *> CompressionFileFormatList;
+
+CompressionFileFormatList &
+compressionFileFormats()
+{
+    static CompressionFileFormatList formats;
+    return formats;
+}
+
+void
```
odanrc, line 75: nit: this is inconvenient/easy to forget to update.

Why not make it more automated and less centralized through a singleton registry class? Roughly:

```
class CompressionRegistry {
private:
    CompressionRegistry() = default;
    CompressionFileFormatList _formats;

public:
    static CompressionRegistry& instance() {
        static CompressionRegistry registry;
        return registry;
    }

    void register(std::unique_ptr<CompressionFormat> format) {
        _formats.push_back(std::move(format));
    }

    const CompressionFileFormatList& getFormats() const { return _formats; }
};

template <typename Format>
class CompressionFormatRegistrar {
public:
    CompressionFormatRegistrar() {
        CompressionRegistry::instance().register(
            std::make_unique<Format>());
    }
};

int decompressImageFile(int fd, const std::string& filename)
{
    const auto formats = CompressionRegistry::instance().getFormats();
    // ... Do decompress
}

// On gzip and zstd files:
namespace {
CompressionFormatRegistrar<ZstdCompressionFormat> zstdRegistrar;
}
```

Although still forget-prone, I think it is less risky because a new format would probably be added by copying the gzip or the zstd file, and the registrar is instantiated in those files. Besides, there would be no centralized location being touched to add formats.

src/base/loader/compression_file_format.hh
giactra: Ack will do that

src/base/loader/compression_file_format.cc
giactra: This is more or less the original design (you can check the first draft of the PR). I had to change the decentralized approach to the centralized one as I was breaking compatibility with gzipped binaries when compiling gem5 as a static library (libgem5).

This was flagged by Co-pilot in the PR. You can check the comment above. In synthesis, if we instantiate in gzip/zstd.cc files, in anonymous namespaces, we won't be adding the formats when statically linking as a client will only include compression_format.hh and the linker won't try to include zstd/gzip compilation units

odanrc: Oh, now I get that AI comment. That's unfortunate. We could make scons generate that centralized register function, but I don't think it is worth the effort. Maybe add a comment to the gzip/zstd files to remember to register the class (without actually mentioning the function name). If you decide not to add the comment, it is fine.

BobbyRBruce:
> > I notice Zstd and Gzip libraries are included in Boost. We already have Boost as an optional dependency, so if we utilized Boost for gzip and zstd compressed image support we could avoid expanding our set of optional dependencies and just state that Boost is required for this.
> > Utilizing the Boost libraries would also avoid us rolling our own gzip and zstd comp/decomp like we do here.
> 
> Hi Bobby,
> 
> I agree with Jason here, we have historically tried to stay away from boost as a dependency. In general I feel less dependencies are better for the project. Using low level APIs might also be a requirement (but I don't know boost enough) the moment we decide we want to implement something more complex like seekable frames (with zstd). With a bit of restructuring it might allow us to decompress and load portions of a guest binary on demand (when we see a memory access to it).
> 
> Please let me know what you think

Yeah, I retract my suggestion. I thought we used Boost for a bit more than we do (it's necessary if you want to use the SystemC implementation), and i agree it's just too large to use for little things like this. My suggestion can be ignored.

BobbyRBruce (approved):
I'm fine with this.

PR #3401, opened 2026-08-17
stdlib,tests: Add MI200 and MI355X GPUFS smoke tests

### LINKED ISSUE #456: There are no GPUFS tests

BobbyRBruce:
## Overview

This change adds dedicated standard-library GPU full-system coverage:

- Daily cold-boots an MI200-family (`MI210`, `gfx90a`) configuration and an
  MI355X (`gfx950`) configuration. Neither Daily test uses a checkpoint.
- Pull-request CI restores a warmed MI355X checkpoint, launches a second real
  `gfx950` kernel, verifies its result, and checks the guest serial marker.
- Both paths run through the public x86 GPU examples so those examples and the
  regression configurations cannot drift independently.

The checkpoint and its smoke program are published as immutable gem5
Resources:

- `x86-mi355x-gpu-fs-smoke-checkpoint`, version `1.0.0`
- `x86-mi355x-gpu-fs-smoke`, version `1.0.0`

Reproducible resource sources are in
[gem5-resources #84](https://github.com/gem5/gem5-resources/pull/84).

## Supporting changes

- Remove the deprecated `gcn-gpu` image, its active tests, workflow entries,
  host tag, and guidance. Archived tests under `tests/deprecated` are untouched.
- Stream disk-image downloads into sparse files by default, while retaining
  dense output through `sparse=False`.
- Fix drain handling for objects created during an active drain and include
  both HBM pseudo channels in checkpoint drain accounting.
- Add a shared x86 GPU example helper and an MI200 entry point. Disk, kernel,
  and checkpoint resource versions remain independently selectable.
- Remove the obsolete `HSA_OVERRIDE_GFX_VERSION` setting from the classic
  MI300X example. ROCr enumeration instead relies on the required guest
  `/proc`, `/sys`, and `/dev/pts` mounts.
- Run the `ALL` GPUFS suites through the standard CI and Daily TestLib
  matrices using the standard Ubuntu dependencies image.


tests/gem5/gpu/configs/gpu_fs_check.sh
```diff
@@ -0,0 +1,42 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+set -u
+
+expected_gpu="$1"
+
+# rocminfo returning successfully is not enough: verify that it reports the
+# architecture expected for the simulated GPU model.
+if rocminfo_output=$(rocminfo 2>&1) && \
+        grep -q "$expected_gpu" <<< "$rocminfo_output"; then
```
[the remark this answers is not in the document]
BobbyRBruce: Addressed. The checks are now embedded in the individual configs and use `grep -Fq --` for the expected `gfx90a` or `gfx942` architecture string.

tests/gem5/gpu/configs/gpu_fs_resources.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+
+    disk_is_set = "--disk-image" in sys.argv
+    kernel_is_set = "--kernel" in sys.argv
+    if disk_is_set != kernel_is_set:
+        raise ValueError(
+            "The GPUFS disk image and kernel must be set together"
+        )
+    if disk_is_set:
+        return None, None
```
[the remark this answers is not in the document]
BobbyRBruce: Addressed by removing the argv-detection helper entirely. Each self-contained GPUFS config now obtains its matched resources directly with `obtain_resource`, so neither `--disk-image` token form is parsed here.

tests/gem5/gpu/configs/gpu_fs_resources.py
```diff
@@ -0,0 +1,82 @@
...
+
+    gpu_directory = Path(__file__).resolve().parent.parent
+    resource_directory = gpu_directory / "resources"
+
```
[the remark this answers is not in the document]
BobbyRBruce: Addressed. Each self-contained config creates its selected resource directory with parents=True, exist_ok=True before calling obtain_resource; the directory also honors the pre-provisioned testlib --bin-path.

.github/workflows/ci-tests.yaml
```diff
@@ -107,8 +107,7 @@ jobs:
                   unset PREV
                   echo "test-dirs-matrix=$(\
                     ./main.py list gem5 \
-                    --suites \
-                    --exclude-tags gcn_gpu | \
+                    --suites | \
```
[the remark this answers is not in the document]
BobbyRBruce: Addressed in the final series, although the implementation has since changed. The GPUFS suites now use `ALL`, so the general matrices no longer discover a `VEGA_X86` suite whose binary they exclude. [`a0b7b0f446`](https://github.com/gem5/gem5/commit/a0b7b0f4466b1b95c234f30ffed176449711ccb2) registers the GPUFS suites in the standard CI and Daily paths, and [`24084284d0`](https://github.com/gem5/gem5/commit/24084284d04801ff209cda48f19b0a389a05109d) removes the obsolete GPU-specific jobs and host dependency. The final TestLib selection contains one quick checkpoint suite and two long cold-boot suites.

BobbyRBruce:
I also audited the earlier Copilot review-body comments that were not represented as active inline threads. The resource-publication note is outdated against the current branch because the resources are published. The MI200 classification concern is addressed: both the PR description and TestLib registration now identify its cold boot as a long/Daily test. The GPU SE/Ruby removal is intentional and documented in the PR scope.

The remaining valid matrix-duplication concern is addressed by the final series: [`a0b7b0f446`](https://github.com/gem5/gem5/commit/a0b7b0f4466b1b95c234f30ffed176449711ccb2) registers the `ALL` GPUFS suites in the standard CI and Daily paths, while [`24084284d0`](https://github.com/gem5/gem5/commit/24084284d04801ff209cda48f19b0a389a05109d) removes the obsolete GPU-specific jobs and host dependency.

src/sim/drain.cc
```diff
@@ -194,7 +194,8 @@ DrainManager::drainableCount() const
 
 Drainable::Drainable()
     : _drainManager(DrainManager::instance()),
-      _drainState(_drainManager.state())
+      _drainState(_drainManager.state() == DrainState::Draining ?
+                      DrainState::Running : _drainManager.state())
```
abmerop: Nice fix. I ran into this issue creating checkpoints as well but the fix is more complex and specific to the GPU memory manager. I have only seen this in the GPU model so probably fine here, but this might deserve its own PR if it fixes other ISAs.

src/python/gem5/resources/resource.py
```diff
@@ -953,6 +954,9 @@ def obtain_resource(
                     **Note**: Usage of this parameter will override the
                     ``resource_directory`` parameter.
     :param quiet: If ``True``, suppress output. ``False`` by default.
+    :param sparse: If ``True``, request a sparse local representation of a
+                   disk image resource. If ``False``, disable sparse
+                   downloading. ``True`` by default.
```
abmerop, line 959: This is an interesting solution to the disk size vs. tester VM space problem. Just an FYI, one of the major features of latest ROCm (7.14) is reduced size of installed packages. That disk image is <10GB (ROCm only, no other frameworks like PyTorch). Still very useful though as we typically need to leave space to copy in binaries.

configs/example/gpufs/mi300.py
```diff
@@ -77,35 +73,75 @@
...
 /sbin/m5 exit
 """
 
 demo_runscript_with_checkpoint = """\
 export LD_LIBRARY_PATH=/opt/rocm/lib:$LD_LIBRARY_PATH
 export HSA_ENABLE_INTERRUPT=0
-export HCC_AMDGPU_TARGET=gfx942
-export HSA_OVERRIDE_GFX_VERSION="9.4.2"
+export HCC_AMDGPU_TARGET={gpu_target}
+export HSA_OVERRIDE_GFX_VERSION="{gfx_override}"
```
abmerop: This override is actually no longer needed

abmerop (commented):
Thanks for putting this together, this will be useful for making sure the GPU model is not breaking with some unrelated PRs.

Any reason why you chose to use the "classic" configs (configs/example/gpufs/miXXX.py) instead of the stdlib configs (configs/example/gem5_library/x86-mi300x-gpu.py)? I wanted to converge on one set of scripts and eventually remove the other. It seemed like stdlib was the way forward there. Something to consider before putting too much more effort down the current path.

src/sim/drain.cc
BobbyRBruce: Thanks. The related changes remain split into self-contained commits:

- [`62fa98e834`](https://github.com/gem5/gem5/commit/62fa98e834bcd9e589f83695d5728d38701b0a21) contains the generic `Drainable` state fix, standalone DMA-test registration, and chained-DMA regression reproducing the callback order independently of the GPU model.
- [`8c0cb335d0`](https://github.com/gem5/gem5/commit/8c0cb335d03d7590333335fb9344de8095f20779) separately drains both HBM pseudo channels.
- [`a0b7b0f446`](https://github.com/gem5/gem5/commit/a0b7b0f4466b1b95c234f30ffed176449711ccb2) contains the dedicated GPUFS coverage that exposed both drain issues, including the warmed MI355X checkpoint flow. [`6f52ab3141`](https://github.com/gem5/gem5/commit/6f52ab3141e47bf537a9504b709a242d61e1c4ed) later routes that coverage through the public example scripts.

src/python/gem5/resources/resource.py
BobbyRBruce: Thanks for the note. The smaller ROCm 7.14 footprint should help future images, while sparse extraction still avoids allocating the image logical size and preserves space for injected binaries and checkpoint artifacts.

configs/example/gpufs/mi300.py
BobbyRBruce: Confirmed: the override is no longer needed and has been removed.

BobbyRBruce:
> Thanks for putting this together, this will be useful for making sure the GPU model is not breaking with some unrelated PRs.
> 
> Any reason why you chose to use the "classic" configs (configs/example/gpufs/miXXX.py) instead of the stdlib configs (configs/example/gem5_library/x86-mi300x-gpu.py)? I wanted to converge on one set of scripts and eventually remove the other. It seemed like stdlib was the way forward there. Something to consider before putting too much more effort down the current path.

The reason was I didn't know we had these stdlib configs! I therefore updated the PR to utilize them. At the time of writing the tests utilize the configs/example/gem5_library configs. These have also been modified but only for better flexibility and to make them more suitable for testing. They are essentially the same as before.

BobbyRBruce:
As far as I'm concerned this PR is complete, in a good state, and ready for review/approval.

configs/example/gem5_library/x86-mi355x-gpu-checkpoint.py
```diff
@@ -0,0 +1,158 @@
...
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+"""Create the two-stage MI355X checkpoint used by the GPUFS smoke test."""
```
powerjg: Can you add more docstring here? E.g., explain how to use this script. The flags are not obvious.

configs/example/gem5_library/x86_gpu.py
powerjg, line 1: Maybe this, and the other GPU scrips should be moved to a subdirectory? Just a thought, not an ask.

configs/example/gem5_library/x86_gpu.py
```diff
@@ -0,0 +1,233 @@
...
+        type=Path,
+        help="local checkpoint directory to restore",
+    )
+    checkpoint.add_argument(
+        "--checkpoint-resource",
+        help="CheckpointResource ID to obtain and restore",
+    )
+    parser.add_argument(
+        "--checkpoint-resource-version",
+        default=GPUFS_RESOURCE_VERSION,
```
powerjg: I don't think it makes sense to have the same version for the checkpoint and the disk. These are versioned independently.

src/sim/drain.cc
```diff
@@ -190,11 +190,11 @@ DrainManager::drainableCount() const
     return _allDrainable.size();
 }
 
-
-
 Drainable::Drainable()
     : _drainManager(DrainManager::instance()),
-      _drainState(_drainManager.state())
+      _drainState(_drainManager.state() == DrainState::Draining
```
powerjg, line 198: Can you explain the need for this change. It is suspicious. Why would you initialize the object to running if it's draining?

src/sim/drain.hh
```diff
@@ -282,6 +282,11 @@ class Drainable
      * until all objects return DrainState::Drained on the first call
      * to drain().
      *
+     * Objects created while the system is draining start in the Running
```
powerjg: How can this happen? I guess it answers my question above, but it's still suspicious. What is the trace of events that causes this issue?

powerjg (commented):
generally lgtm. Some minor comments.

configs/example/gem5_library/x86-mi200-gpu.py
```diff
@@ -0,0 +1,36 @@
...
+
+"""Run an x86 full-system simulation with an MI200-series GPU."""
+
+from x86_gpu import run_gpu_example
+
+from gem5.components.devices.gpus.amdgpu import MI210
+
+run_gpu_example(
+    gpu_class=MI210,
+    description="Run an x86 full-system simulation with an MI200-series GPU.",
```
erin-le: Would it be better to name this file `x86-mi210-gpu.py` instead, and change the description as well?
```suggestion
from gem5.components.devices.gpus.amdgpu import MI210

run_gpu_example(
    gpu_class=MI210,
    description="Run an x86 full-system simulation with an MI210-series GPU.",
```

configs/example/gem5_library/x86_gpu.py
BobbyRBruce: This grouping does make sense as a possible cleanup. I kept the files flat in this PR because `x86-mi300x-gpu.py` and `x86-mi355x-gpu.py` are existing public example paths; moving them would broaden this change and break established invocations. I would handle a directory reorganization separately so those compatibility implications can be considered explicitly.

src/sim/drain.cc
BobbyRBruce: The `DrainManager` is draining, but this newly constructed `Drainable` is not yet part of the active drain pass. An object only enters `Draining` when `dmDrain()` has been called on it.

`DrainManager::_count` is the number of objects that returned `Draining` during the current `tryDrain()` pass. The regression in [`62fa98e834`](https://github.com/gem5/gem5/commit/62fa98e834bcd9e589f83695d5728d38701b0a21) demonstrates the failure:

1. `tryDrain()` counts the first `DmaCallback`, making `_count == 1`.
2. That callback completes and creates a second callback after the drain pass established its count.
3. The first callback signals completion, reducing `_count` to zero.
4. With the old constructor, the uncounted second callback inherits `Draining`. When it completes, it calls `signalDrainDone()` and attempts to decrement the already-zero count, triggering `assert(_count > 0)`.

Initializing the new object as `Running` records that `dmDrain()` has not yet been called on it. Its completion therefore cannot decrement the active count. If it survives, the normal drain loop calls `tryDrain()` again and drains it in that subsequent pass.

configs/example/gem5_library/x86_gpu.py
BobbyRBruce: Agreed and fixed in [`f1ba58d525`](https://github.com/gem5/gem5/commit/f1ba58d525816eb3136f9afef8ebae50f3defe45). `--disk-resource-version` and `--kernel-resource-version` now control those resources independently, while `--checkpoint-resource-version` remains separate for checkpoints.

configs/example/gem5_library/x86-mi355x-gpu-checkpoint.py
BobbyRBruce: Added in [`6f52ab3141`](https://github.com/gem5/gem5/commit/6f52ab3141e47bf537a9504b709a242d61e1c4ed). The module docstring now shows both `create-loader` and `create-kernel` commands, identifies the input and output of each stage, and documents the Atomic four-CU topology and the fact that the warmed checkpoint is not a generic post-boot checkpoint.

src/sim/drain.hh
BobbyRBruce: The concrete event trace is exercised by `DmaCallbackTest.ChainedCallbackCompletesDuringDrain` in [`62fa98e834`](https://github.com/gem5/gem5/commit/62fa98e834bcd9e589f83695d5728d38701b0a21):

1. The first callback has a pending chunk, so `tryDrain()` calls its `dmDrain()`; it returns `Draining` and contributes one to `DrainManager::_count`.
2. Processing that callback creates a second callback while the manager is still `Draining`. The second callback was not visited by this `tryDrain()` pass and contributed nothing to `_count`.
3. The first callback then calls `signalDrainDone()`, reducing `_count` from one to zero.
4. Before this fix, the second callback inherited `Draining`. Completing it therefore called `signalDrainDone()` again and hit `assert(_count > 0)`.

Starting the second callback in `Running` preserves the invariant that only an object whose `dmDrain()` returned `Draining` may decrement the active count. If the object survives, the next `tryDrain()` pass calls `dmDrain()` and accounts for it normally. The revised constructor documentation now states this invariant directly.

BobbyRBruce:
@abmerop : Could you give this a review when you have the time?

powerjg (dismissed):

abmerop:
> @abmerop : Could you give this a review when you have the time?

I'll get to it soon. I see it is already merged but can check if there is something that could use another PR.

abmerop (commented):
Got a chance to take a look - Everything looks good

